PR #3240, opened 2016-01-14
devpollmplexer fixes + fix DS ucontext.h pollution once and for all

jeffpc:


jeffpc:
FWIW, the devpollmplexer changes are the same as what's on the 3.7.4 branch.

PR #3242, opened 2016-01-15
Add test for zone2sql --json-comments

### LINKED ISSUE #3181: zone2sql --json-comments doesn't have a test
It's never used in any test, so we don't know if it works.

pieterlexis:
Closes #3181

### DIFF
--- a/regression-tests.nobackend/zone2sql-json-comments/command
+++ b/regression-tests.nobackend/zone2sql-json-comments/command
@@ -1,0 +1,5 @@
+#!/bin/sh
+
+# We use faketime because the comments table has a modified_at field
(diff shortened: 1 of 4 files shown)


zeha:
```
-insert into comments (domain_id,name,type,modified_at, comment) select id, 'example.net', 'NS', 1451602800, 'Some NS record' from domains where name='example.net';
+insert into comments (domain_id,name,type,modified_at, comment) select id, 'example.net', 'NS', 1451606400, 'Some NS record' from domains where name='example.net';
```

Strip modified_at out of the result, somehow?

pieterlexis:
this is probably a timezone issue....

zeha:
Same test failure again :)

PR #3244, opened 2016-01-15
Fix the forward zones in the recursor

pieterlexis:
In the pre-DNSName era, when dns-native names were passed as strings, we
overloaded the NS-name for a forward or auth zone. e.g. an empty string
meant 'this is an auth zone' and '+203.0.113.1' meant 'forward to 203.0.113.1
with the RD bit set'. With DNSNames, this is impossible (yay!).

In this commit, the set of strings (and later DNSNames), is replaced by
a map where a DNSName is the key and the value is a pair of a
ComboAddress and a boolean.

A non-empty DNSName: This is a normal NS, recurse as usual (the pair is
ignored).

An empty DNSName and empty ComboAddress: We are auth for this zone,
check the auth store for an answer.

An empty DNSName and non-empty ComboAddress: The query must be forwarded
to the ComboAddress specified and the boolean in the pair tells us the
value of the RD bit in the query we need to send.

### DIFF
--- a/pdns/syncres.cc
+++ b/pdns/syncres.cc
@@ -432,7 +432,7 @@ int SyncRes::doResolve(const DNSName &qname, const QType &qtype, vector<DNSRecor
   DNSName subdomain(qname);
   if(qtype == QType::DS) subdomain.chopOff();
 
-  set<DNSName> nsset;
+  map< DNSName, pair< ComboAddress, bool > > nsset;
   bool flawedNSSet=false;
 
   // the two retries allow getBestNSNamesFromCache&co to reprime the root
@@ -618,7 +618,7 @@ SyncRes::domainmap_t::const_iterator SyncRes::getBestAuthZone(DNSName* qname)
 }
 
 /** doesn't actually do the work, leaves that to getBestNSFromCache */
-DNSName SyncRes::getBestNSNamesFromCache(const DNSName &qname, const QType& qtype, set<DNSName>& nsset, bool* flawedNSSet, int depth, set<GetBestNSAnswer>&beenthere)
+DNSName SyncRes::getBestNSNamesFromCache(const DNSName &qname, const QType& qtype, map<DNSName, pair<ComboAddress, bool> >& nsset, bool* flawedNSSet, int depth, set<GetBestNSAnswer>&beenthere)
 {
   DNSName subdomain(qname);
   DNSName authdomain(qname);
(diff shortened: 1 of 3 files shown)


pieterlexis:
ToDo:

- [ ] Check if `forward-zones-file` still works

pdns/syncres.cc
```diff
@@ -954,14 +951,14 @@ int SyncRes::doResolveAt(set<DNSName> nameservers, DNSName auth, bool flawedNSSe
       bool sendRDQuery=false;
       boost::optional<Netmask> ednsmask;
       LWResult lwr;
-      if(tns->empty()) {
+      if(tns->empty() && nameservers[*tns].first == ComboAddress() ) {
```
ahupowerdns: I'm not sure if this is idiomatic. We should maybe add a test for it if it is. It looks like we currently get away with it.

pdns/syncres.hh
```diff
@@ -460,7 +460,7 @@ public:
 
 private:
   struct GetBestNSAnswer;
-  int doResolveAt(set<DNSName> nameservers, DNSName auth, bool flawedNSSet, const DNSName &qname, const QType &qtype, vector<DNSRecord>&ret,
+  int doResolveAt(map<DNSName, pair<ComboAddress, bool> > nameservers, DNSName auth, bool flawedNSSet, const DNSName &qname, const QType &qtype, vector<DNSRecord>&ret,
```
ahupowerdns: it is pretty odd that this was passed by value. I'm sure there was a reason for it one day, might be good to see if we can get away with a const reference now.

pdns/syncres.cc
```diff
@@ -626,14 +626,11 @@ DNSName SyncRes::getBestNSNamesFromCache(const DNSName &qname, const QType& qtyp
   domainmap_t::const_iterator iter=getBestAuthZone(&authdomain);
   if(iter!=t_sstorage->domainmap->end()) {
     if( iter->second.d_servers.empty() )
-      nsset.insert(DNSName()); // this gets picked up in doResolveAt, if empty it means "we are auth", otherwise it denotes a forward
+      nsset.insert(make_pair(DNSName(), make_pair(ComboAddress(), false))); // this gets picked up in doResolveAt, the empty DNSName, combined with the empty ComboAddress means 'we are auth'
```
ahupowerdns: you could try {DNSName(), {ComboAddress(), false}} instead of make_pairs - might just work.

pdns/syncres.hh
pieterlexis: `nameservers` is modified in this function [https://github.com/PowerDNS/pdns/pull/3244/files#diff-9b3238d7a7cb2e313f1d8f1588971f93L1360](here)

pieterlexis: oh wait, completely mixed up `const` and pass-by-X in my head :)

pdns/syncres.cc
pieterlexis: I added a unit test for this.

PR #3251, opened 2016-01-17
dnsdist: Document toString() aliases. Add TCPRule. Make AnyTCRule set TC only over UDP

rgacogne:
- For maximum compatibility with PowerDNS Lua scripts, I have recently added toString() and toStringWithPort() aliases, but forgot to document them
- Add TCPRule() to be able to match queries based on the protocol they were received over
- Make addAnyTCRule set TC=1 for ANY queries received on UDP, but not on TCP, as suggested by Greg on the dnsdist mailing-list (thanks!)


ahupowerdns:
I was just pondering the need for all this! Good stuff!

PR #3255, opened 2016-01-18
Don't log authentication errors before sending HTTP basic auth challenge

### LINKED ISSUE #3254: dnsdist: Erroneous "Web Authentication failed" errors
**Steps to reproduce**

1. Clear browser cache
2. Open dnsdist web-ui in browser
3. Enter correct credentials in password prompt
4. Check dnsdist log/console

**Expected result**
No error is logged

**Actual result**
> HTTP Request "/" from 46.xxx.xxx.xxx:63904: Web Authentication failed

janeczku:
fixes #3254

### DIFF
--- a/pdns/dnsdist-web.cc
+++ b/pdns/dnsdist-web.cc
@@ -88,7 +88,9 @@ static void connectionThread(int sock, ComboAddress remote, string password)
       resp.status=200;
     }
     else if (!compareAuthorization(req, password)) {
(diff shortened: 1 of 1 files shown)


stbuehler:
You also change the order in which 405 (Method not allowed) and 401 (Unauthorized) are returned. Please restrict your changes to what the commit messages says :)

Maybe instead of duplicating the whole 401 block you could just make the `errlog` call conditional on `req->headers.find("authorization") != req->headers.end()`

PR #3259, opened 2016-01-18
[Do Not Merge] Fully autotoolize the recursor distribution

pieterlexis:
The commits need to be squashed, but please test.

The distribution of the recursor now works the same as dnsdist (`cd pdns/recursordist; autoreconf -i; ./configure; make dist`), please the the tarballs generated from `make dist`.


Habbie:
Apart from checking in both the `ext/` symlink forest and the `.py` making it, this LGTM.

PR #3274, opened 2016-01-20
redo packet cache to be simpler, have more features, testcases and still reduce lines of code!

ahupowerdns:


ahupowerdns:
with thanks to @Habbie who pushed me to do this.

PR #3279, opened 2016-01-21
Handle return codes from OpenSSL BIO_*

zeha:
Fixes warning from gcc.

### DIFF
--- a/pdns/base64.cc
+++ b/pdns/base64.cc
@@ -2,6 +2,7 @@
 #include "config.h"
 #endif
 #include "base64.hh"
+#include <stdexcept>
 #include <boost/scoped_array.hpp>
 #include <openssl/bio.h>
 #include <openssl/evp.h>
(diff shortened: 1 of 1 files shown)


pdns/base64.cc
```diff
@@ -58,8 +62,10 @@ std::string Base64Encode (const std::string& src)
     bio = BIO_new(BIO_s_mem());
     bio = BIO_push(b64, bio);
     BIO_set_flags(bio, BIO_FLAGS_BASE64_NO_NL);
-    BIO_write(bio, src.c_str(), src.length());
-    BIO_flush(bio);
+    if (BIO_write(bio, src.c_str(), src.length() != src.length())) {
```
rgacogne: I think you have misplaced a ')', here src.length() != src.length() is not going to work well :)

pdns/base64.cc
```diff
@@ -32,6 +33,9 @@ int B64Decode(const std::string& src, std::string& dst)
   bio = BIO_push(b64, bio);
   BIO_set_flags(bio, BIO_FLAGS_BASE64_NO_NL);
   olen = BIO_read(b64, d.get(), dlen);
+  if ((olen == 0 || olen == -1) && BIO_should_retry(bio)) {
+    throw std::runtime_error("BIO_read failed to read all data from memory buffer");
```
rgacogne: Doesn't that leak bio?

Habbie:
ping

pieterlexis:
ping!

zeha:
Rebased on top of current master to get rid of merge issues, and fixed the possible BIO leaks, plus actually compiles ;-)

PR #3281, opened 2016-01-21
Fully take out mbedTLS

pieterlexis:

PR #3283, opened 2016-01-21
Remove Crypto++ support

pieterlexis:


zeha:
Could you rebase this on top of master, so we get a working travis build (while xs is down). (Changes LGTM)

ahupowerdns:
is there a reason to remove this beyond "it is not better than the other dependencies"?

pieterlexis:
we now require OpenSSL (see #3281), so cryptopp primitives are never used. (discussed with @Habbie, @zeha and @mind04 on IRC)

PR #3293, opened 2016-01-23
make asynchronous UDP Lua queries work again in 4.x

### LINKED ISSUE #3276: Recursor: udpQueryResponse broken in 4.x Lua code
Setting any of
```Lua
dq.udpQueryDest
dq.udpCallback
dq.udpQuery
```
will result in the following error:
> STL error (www.example.com./A from 127.0.0.1): No setter found

Also, the code that handles `udpQueryResponse` results in an endless loop as it doesn't actually evaluate the callback function:
https://github.com/PowerDNS/pdns/blob/master/pdns/lua-recursor4.cc#L388-L406

ahupowerdns:
Closes #3276 when merged.

### DIFF
--- a/pdns/lua-recursor4.cc
+++ b/pdns/lua-recursor4.cc
@@ -247,6 +247,11 @@ RecursorLua4::RecursorLua4(const std::string& fname)
   d_lw->registerMember("followupFunction", &DNSQuestion::followupFunction);
   d_lw->registerMember("followupPrefix", &DNSQuestion::followupPrefix);
   d_lw->registerMember("followupName", &DNSQuestion::followupName);
+  d_lw->registerMember("data", &DNSQuestion::data);
+  d_lw->registerMember("udpQuery", &DNSQuestion::udpQuery);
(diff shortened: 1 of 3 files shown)


pdns/lua-recursor4.cc
```diff
@@ -386,26 +391,31 @@ bool RecursorLua4::genhook(luacall_t& func, const ComboAddress& remote,const Com
...
+        auto func = d_lw->readVariable<boost::optional<luacall_t>>(dq->udpCallback).get_value_or(0);
+        if(!func) {
+          theL()<<Logger::Error<<"Attempted callback for Lua UDP Query/Response which could not be found"<<endl;
+          return false;
+        }
+        bool res=func(dq);
+        if(variable) *variable |= dq->variable; // could still be set to indicate this *name* is variable
+        if(!res) {
+          return false;
+        }
```
janeczku: @ahupowerdns Thanks for fixing!
I think there is a `ret=dq->rcode;` missing here.
Otherwise setting `dq.rcode` in the callback will not take effect...

PR #3295, opened 2016-01-24
Correct YaHTTP license

cmouse:


zeha:
Please keep use of the Expat name, as per DEP5

cmouse:
what does Expat mean?

Habbie:
ping

cmouse:
There was some indication that MIT might be valid too. @zeha I know I asked this over IRC but could you comment here?

zeha:
DEP5 says "use Expat". Not sure which point wasn't clear.

cmouse:
Expat is not mit license. Why does it mandate different license?

PR #3299, opened 2016-01-25
ragel is also a build dependency

HLFH:


Habbie:
Uhm, don't we shop yahttp and json11 in our tree?

cmouse:
Depends what you understand with build dependencies =)

HLFH:
@Habbie 

> On Linux, make pdns_recursor in the pdns subdir may work. The portable, and supported, way to build the recursor is first running dist-recursor and compiling from the pdns-recursor-x.y subdirectory.

And you're right. Maybe the last commit is useless but it seems that when we build the recursor by running `./build-scripts/dist-recursor`, we  have to compile from the `pdns/recursordist` directory by using the commands `./configure`, `make` (not `make pdns_recursor`!) and then install with `make install`. Ragem seems the only dependency. Or, am I completely wrong about the process to build pdns-recursor?

Habbie:
The right way to build the recursor is from that dist dir, yes.

HLFH:
@Habbie So how is this last commit? Is that more clear to build pdns-recursor with these added infos? https://github.com/HLFH/pdns/commit/7dffd6946e62fc1ccc86f8541d18e37d7bb3bc62

pieterlexis:
hrm, considering the pdns/recursordist dir is now equal (when the symlinks are dereferenced) to the tarball, the dist-recursor step is no longer needed. I guess the best way to compile the recursor from git is:

1. `cd pdns/recursordist`
2. `./bootstrap`
3. `./configure`
4. `make`
5. `make install`

HLFH:
@pieterlexis Is that better now? Should I remove dist-recursor or shoudl I finally keep it?

pieterlexis:
@HLFH that file needs to stay, we use it to create tarballs for releases :-)

HLFH:
@pieterlexis Thanks! Is that now correct?

pieterlexis:
@HLFH if you change the filemode of dist-recursor back to 0755, I'm happy. With regards to your commits, you can [squash](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html) them and force-push this to your branch to update this PR.

PR #3317, opened 2016-01-28
add incfiles and src_js/ to dnsdist dist tarball

tjikkun:
This allows to regenerate the js files


rgacogne:
Hi Sander,
Thank you, could you just fix the indentation? Otherwise this looks fine to me.

tjikkun:
Heh that is some weird indentation :)
Should we go with tabs or spaces? "Let's do both"
Anyway, now mine is the same.

rgacogne:
Thank you! And we are indeed not always consistent, we need to work on that!

PR #3319, opened 2016-01-29
PDNSUTIL EDIT-ZONE

ahupowerdns:


mind04:
You cannot edit zones if there are ents present
```
$ ../pdns/pdnsutil --config-dir=. --config-name=gmysql edit-zone test.com
Error: Unknown record was stored incorrectly, need 3 fields, got 0: 
```

PR #3322, opened 2016-02-02
Clarify dnssec=process

paddg:
Hopefully I got it right.


pieterlexis:
hrm, it shouldn't work like this. `process` should just set DO and return DNSSEC records to clients that ask for it, but not validate. @ahupowerdns?

paddg:
@pieterlexis,  But then I miss the option "validate on client-request"

Habbie:
Will be corrected as part of #3682, thanks!

PR #3331, opened 2016-02-03
Link pdnsutil with lua, fixes #3330

### LINKED ISSUE #3330: pdnsutil needs to be linked with the same lua lib as pdns_server
pdnsutil currently is not linked with the $(LUA_LIBS), which causes the weird error of a libluabackend.so seeming to load correctly and then not being able to 'require' a lua-module that is a shared library itself.

adding $(LUA_LIBS) to pdnsutil_LDADD fixes that problem.

no patch/pr because i have no idea how to fix this in the automake world.

cmouse:


Habbie:
This fix looks wrong to me, `libluabackend.so` should be pulling this in by itself..

PR #3338, opened 2016-02-05
Add weight feature to GeoIP backend

cmouse:
This allows administrator to define weights on records so that they are returned approximately weight% times on queries.

### DIFF
--- a/docs/markdown/authoritative/backend-geoip.md
+++ b/docs/markdown/authoritative/backend-geoip.md
@@ -55,7 +55,7 @@ domains:
       - ns: ns2.example.com
       - mx: 10 mx.example.com
     fin.eu.service.geo.example.com:
-      - a: 198.51.100.221
+      - a: 192.0.2.1
       - txt: hello world
       - aaaa: 2001:DB8::12:34DE:3
   services:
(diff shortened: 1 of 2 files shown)


docs/markdown/authoritative/backend-geoip.md
```diff
@@ -71,13 +71,20 @@ domains:
...
       - mx: 10 mx.example.com
     fin.eu.service.geo.example.com:
       - a: 198.51.100.221
       - txt: hello world
       - aaaa: 2001:DB8::12:34DE:3
+    swe.eu.service.geo.example.com
+      - a:
+           content: 1.2.3.4
+           weight: 50
+      - a: 2.3.4.5
```
ahupowerdns: what is the weight of 2.3.4.5?

Habbie:
(Please use documentation IP space like 192.0.2.0/24)

modules/geoipbackend/geoipbackend.cc
```diff
@@ -123,9 +128,24 @@ void GeoIPBackend::initialize() {
...
+             string attr = iter->first.as<string>();
+             if (attr == "content") {
+	       string content = iter->second.as<string>();
+               rr.content = content;
+             } else if (attr == "weight") {
+               rr.weight = iter->second.as<int>();
+             } else if (attr == "ttl") {
+               rr.ttl = iter->second.as<int>();
+             } else {
+               L<<Logger::Warning<<"Unsupported record attribute " << attr << " for " << rr.qname << endl;
```
ahupowerdns: just kill it at this point. Will only lead to silently wrong config files.

modules/geoipbackend/geoipbackend.cc
```diff
@@ -208,6 +230,39 @@ void GeoIPBackend::initialize() {
...
 
+    // finally fix weights
+    for(auto &item: dom.records) {
+      int weight=-1;
+      int rem, count;
+      rem = count = 0;
+      // first we look for used weight
+      for(const auto &rr: item.second) {
+        if (rr.weight > -1) {
+          if (weight==-1)
```
ahupowerdns: > -1 means it won't be -1

modules/geoipbackend/geoipbackend.cc
cmouse: well, it does log it. =)

cmouse:
Rewrote to use weight "as usual". Please review again, sorry for rebase.

PR #3345, opened 2016-02-08
pdnsutil: warn if we can't read the config file

### LINKED ISSUE #3305: if pdnsutil can't read pdns.conf because of a permission issue, it is silent about that
leading to confusing behaviour, it says 'no backends configured'. And I'm like, DAMN RIGHT THEY ARE! Perhaps pdnsutil should report being unable to load any config at all.

pieterlexis:
Closes #3305

### DIFF
--- a/pdns/pdnsutil.cc
+++ b/pdns/pdnsutil.cc
@@ -99,7 +99,9 @@ void loadMainConfig(const std::string& configdir)
   L.toConsole(Logger::Error);   // so we print any errors
   BackendMakers().launch(::arg()["launch"]); // vrooooom!
   L.toConsole((Logger::Urgency)(::arg().asNum("loglevel")));  
-  ::arg().laxFile(configname.c_str());    
+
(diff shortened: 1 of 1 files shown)


zeha:
(lax)File already returns true/false and writes a message to L ... maybe use the returncode or hook up L instead?

pdns/pdnsutil.cc
```diff
@@ -99,6 +99,11 @@ void loadMainConfig(const std::string& configdir)
   L.toConsole(Logger::Error);   // so we print any errors
   BackendMakers().launch(::arg()["launch"]); // vrooooom!
   L.toConsole((Logger::Urgency)(::arg().asNum("loglevel")));  
+
+  struct stat buf;
+  if (stat(configname.c_str(), &buf) != 0)
+    cerr<<"Warning: unable to read configuration file '"<<configname<<"'."<<endl;
```
Habbie: Please catch errors instead of trying to predict them

PR #3346, opened 2016-02-08
dnsdist: Add reuseport support, maxOutstanding -> 10240, document perf tuning

rgacogne:
If set to true, this parameter sets SO_REUSEPORT on platforms
supporting for this option, allowing multiple servers to
bind to the same port.
The same parameter is also added to addDNSCryptBind().


zeha:
If you don't do this by default, add rationale to the docs when it'd be good to disable SO_REUSEPORT?

rgacogne:
Added a "Performance tuning" section to the README, bumped maxOutstanding to 10240 by default.

PR #3349, opened 2016-02-09
Fix tabs handling in TXT records.

ddbilik:
Update txtEscape() in PacketReader to omit also tabs, as they are printable
characters. In fact, tabs are put equivalent to spaces in RFC 1464.

### DIFF
--- a/pdns/dnsparser.cc
+++ b/pdns/dnsparser.cc
@@ -443,7 +443,7 @@ static string txtEscape(const string &name)
   char ebuf[5];
 
   for(string::const_iterator i=name.begin();i!=name.end();++i) {
-    if((unsigned char) *i > 127 || (unsigned char) *i < 32) {
+    if((unsigned char) *i > 127 || ((unsigned char) *i < 32 && (unsigned char) *i != 9)) {
       snprintf(ebuf, sizeof(ebuf), "\\%03u", (unsigned char)*i);
       ret += ebuf;
     }


Habbie:
I'm sorry, what problem does this solve? Printing tabs as single octets reduces zone file readability. Note that 1464 has no bearing on the handling of TXT in DNS software.

ddbilik:
> what problem does this solve?

To not confuse tools that expect printable characters in TXT records "as is", ie. in non-encoded form.

> Printing tabs as single octets reduces zone file readability.

Depends on who is reading. IMHO, "custom\009data" is less readable than "custom        data".

> Note that 1464 has no bearing on the handling of TXT in DNS software.

Yes, I've used it just as an example of how encoding content of TXT records too much (ie. where it's not needed) may influence functionality of some other tools.

cmouse:
If RFC says they are equivalent, shouldn't tabs be replaced with spaces?

ddbilik:
> If RFC says they are equivalent, shouldn't tabs be replaced with spaces?

Sorry, seems I was somewhat misleading here... RFC 1464 does not explicitely say that tabs are equal spaces. But in examples used there, spaces and tabs are handled the same. And that's the point: to handle spaces and tabs the same way, either encode them or leave them as-is, both.

Habbie:
But RFC1464 is not about the behaviour of PowerDNS, it is about a specific application of TXT, namely storing of key/attribute pairs. The equivalence of tabs and spaces (which is indeed not even in there) is not something for us to do anything with.

Habbie:
I only see added user confusion from this patch; closing this PR, sorry.

ddbilik:
No problem. Though I think it's more confusing as it works now, ie. let a user put _his_ data into TXT, and then mangle some printable characters in it, while others leaving alone.

Habbie:
User can put anything in TXT he wants, as long as it is valid. We do not *require* `\009` or forbid 'actual tab' there. If you have an actual problem it is unclear to me.

ddbilik:
Nevermind... I've checked behaviour in several pdns backends, and it seems to impact just specific master-slave setups, ie. "common user" will most probably not hit this. Thank you, and sorry for the noise.

PR #3352, opened 2016-02-10
recursor: when replacing an expired entry, move it to the back

rgacogne:
If we did a lookup earlier and found the entry to be expired,
we moved it to the front of the expunge queue. If we don't move
it to the back, and the next cache purging operation occurs before
a cache hit using this entry, it's very likely to get expunged.


Habbie:
I merged #3354, can you rebase?

rgacogne:
rebased and travis passed!

Habbie:
Currently being tested by user.

PR #3355, opened 2016-02-10
dnsdist: Add a simple Packet Cache

rgacogne:
Per-pool Packet Cache, using the whole query packet minus the id
has hashing key, to prevent issue related to:
* EDNS Payload size
* ECS
* DNSSEC

The packet cache is not enabled by default, and can be skipped
for specific queries using SkipCacheAction.
It's a per-pool cache, in case you have different responses, but
you can use the same cache for several pools if you want to.

We cache the whole response and age the TTLs when fetching the
response from the cache.

This commit also refactors a bit the way server pools are handled
to be able to have a per-pool cache, and to avoid scanning all
servers when looking for the ones in a given pool.

It is using a fixed-size unordered_map to prevent rehashing. It
is not very efficient with regard to cache cleaning, but I really
would like to use only a ReadLock on the fastpath, and using a
multi index container and moving cache entries to the back / front
on hit / miss would prevent that.

Health checks are moved to a different thread, to prevent them from
being impacted by the cache cleaning operation being slow.

### DIFF
--- a/pdns/README-dnsdist.md
+++ b/pdns/README-dnsdist.md
@@ -313,6 +313,7 @@ Current actions are:
  * Delay a response by n milliseconds (DelayAction), over UDP only
  * Modify query to clear the RD or CD bit
  * Add the source MAC address to the query (MacAddrAction)
+ * Skip the cache, if any
 
 Rules can be added via:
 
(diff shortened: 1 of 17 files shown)


rgacogne:
First issue reported by @Habbie, the cache lookup is done before adding ECS information, this is not right

rgacogne:
Moved the cache lookup after the ECS addition. Lookup is now case-insensitive.

rgacogne:
Rebased

Issue #3127, opened 2015-12-28
When system mbedtls is present, it is used even if --with-system-mbedtls is not used


cmouse:
This is because linkage is done with -L../ext/mbedtls -lmbedtls, when it should be done with ../ext/mbedtls/.libs/libmbedtls.a

rubenk:
Oh wow, that has been broken for a while :)

rubenk:
Can't reproduce it though, this is my linker line:

```
[...]

PR #3368, opened 2016-02-13
Link our internal libs directly

rubenk:
If we don't do this and we have an external version of the same
lib installed, we risk picking that one up instead.

See #3127 for background


Habbie:
LGTM

PR #3370, opened 2016-02-14
This adds --enable-re2 so we can benefit from libre2 if we want

ahupowerdns:
Plus documents it. RE2Rule()


ahupowerdns:
@rubenk did I get the autoconf stuff right?

pdns/dnsdistdist/m4/pdns_check_re2.m4
```diff
@@ -0,0 +1 @@
+../../../m4/pdns_check_re2.m4
```
Habbie: do you imagine needing/wanting re2 in auth later?

ahupowerdns: no but this still allows me to make dnsdist from the main directory if needed

pieterlexis:
autotools changes look good from here.

pdns/dnsdistdist/m4/pdns_check_re2.m4
Habbie: that's a feature I feel should go in general, it's a maintenance pain - and now our auth tarball knows about re2!

pieterlexis: I am inclined to agree with @Habbie on this (although I understand both viewpoints)

PR #3371, opened 2016-02-14
Save 4KiB memory allocation per-client in recursor.

nlyan:
The Socket class maintains an internal 4KB buffer which nothing uses except
Socket::recvFrom(), read() and recvFromAsync(), and these in turn only
then copy the content to the caller. Recursor doesn't even use these
functions, which means it allocates 4KiB it never even uses
(PacketID->sock) for every socket.

The following patch eliminates it, hopefully without altering any
behaviour...

### DIFF
--- a/pdns/sstuff.hh
+++ b/pdns/sstuff.hh
@@ -36,11 +36,11 @@ typedef int ProtocolType; //!< Supported protocol types
 //! Representation of a Socket and many of the Berkeley functions available
 class Socket : public boost::noncopyable
 {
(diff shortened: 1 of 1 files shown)


rgacogne:
Needs to be rebased (sorry about that) because the return value of `recv{,from}()` is now stored in a `ssize_t`, but otherwise looks good.

pieterlexis:
I think @ahupowerdns had some strong opinions about this PR, perhaps he can chime in as well

pieterlexis:
Closing for lack of response

PR #3389, opened 2016-02-16
pdnsutil: support chroot

### LINKED ISSUE #964: pdnssec does not chroot
`pdnssec` does not honor the `chroot` config directive from pdns.conf and hence fails to find files like databases or bind config files when pdns is run within chroot.

This is a bug, since I can either specify those paths in a way so that `pdns_server` can find them, or giving the absolute path, so that `pdnssec` finds them; never both.

My current workaround is to pass `pdnssec` a different config file with the absolute paths specified.

The problem occurred to me using pdns 3.0-1.1ubuntu1, however I was told on IRC that it probably wasn't fixed in later versions.

pieterlexis:
Closes #964


pieterlexis:
could use a review

PR #3373, opened 2016-02-15
Give each binary a single place for building

Habbie:

PR #3395, opened 2016-02-17
re-dist dnsrecords.hh

Habbie:
@mind04 reported this file was missing from our dist balls since #3373. This PR fixes that. I confirmed locally that after this, a dist tarball will build.

### DIFF
--- a/pdns/Makefile.am
+++ b/pdns/Makefile.am
@@ -142,7 +142,7 @@ pdns_server_SOURCES = \
 	dnspacket.cc dnspacket.hh \
 	dnsparser.cc \
 	dnsproxy.cc dnsproxy.hh \
-	dnsrecords.cc \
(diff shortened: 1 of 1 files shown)


pieterlexis:
maybe we should `make dist` and build the auth from the tarball in travis as well?

Habbie:
I wouldn't mind that.

cmouse:
This should happen if you make a build dir and build there
```
mkdir build
cd build
../sources/configure ....
```

Habbie:
@cmouse that symlinks in all the 'disted' files?

cmouse:
nope. it copies them.

mind04:
Depend on the amount of extra build time it will add to travis.

Habbie:
@cmouse I just tested this, it would not have detected the issue. There is no copying, no symlinking.

PR #3400, opened 2016-02-17
Fix building on OpenIndiana (and possibly Solaris)

### LINKED ISSUE #3369: Please revert a79ce884a004a423c203597749475e4d5fb004a7
I'm not sure what issue it fixed, but imho there are a few issues with this change:
- We lost the check for inet_aton in libresolv
- It adds -lnsl to LIBS if libsocket contains socket, regardless of if it is needed. It isn't if gethostname is in libc and not in libnsl.
- It first looks for socket in libsocket, and if it doesn't find it in libsocket, it looks for it in libsocket again.

pieterlexis:
This PR closes #3369 by fixing the issues noted therein.

auth, recursor and dnsdist now succesfully build on openindana hipster (tested).


m4/pdns_enable_reproducible.m4
```diff
@@ -12,7 +12,7 @@ AC_DEFUN([PDNS_ENABLE_REPRODUCIBLE], [
     AC_DEFINE([REPRODUCIBLE], [1], [Define to 1 for reproducible builds])
   ],[
     build_user=$(id -u -n)
-    if [ test x"$host_os" = "xSunOS" ]; then
+    if [ test x"$host_os" = "xSunOS" -o x"$host_os" = "xsolaris2.11" ]; then
```
pieterlexis: we could use `$have_solaris` here, but that would mean always calling `PDNS_CHECK_OS` before `PDNS_ENABLE_REPRODUCIBLE`

pieterlexis:
all 3 products still build on linux as well, FreeBSD and OpenBSD need to be tested.

m4/pdns_enable_reproducible.m4
rubenk: You could use AC_REQUIRE([PDNS_CHECK_OS])

Btw, I'm wondering when $host_os is ever SunOS? The systems I've seen always have something like solaris*

Otherwise LGTM.

Habbie:
I think OpenBSD didn't work before this PR either, so feel free to ignore that. For FreeBSD I also feel the community can pick that up.

m4/pdns_check_os.m4
```diff
@@ -2,7 +2,7 @@ AC_DEFUN([PDNS_CHECK_OS],[
   THREADFLAGS=""
 
   case "$host_os" in
-  solaris2.10)
+  solaris2.10 | solaris2.11)
```
Habbie: `solaris2.1*` ?

pieterlexis: "but maybe they will break it in 2.12"

pdns/recursordist/Makefile.am
```diff
@@ -216,3 +222,14 @@ $(MANPAGES):
 	exit 1
 endif
 endif
+
+if !HAVE_LUA_HPP
+# no lua.hpp on solaris :(
+lua.hpp:
```
Habbie: We have this code in two places now. How about just shipping our own lua.hpp and never relying on the distro?

m4/pdns_check_os.m4
Habbie: That is far less likely than 2.12 being exactly the same, in which case we need to edit. It is a gamble either way, but I like gambles that have a fair chance of not causing action :)

pdns/recursordist/Makefile.am
pieterlexis: In the odd case that the OS's (or user's with `CFLAGS=-Imyspeciallua`) lua.hpp is 'special' (patched from upstream for whatever reason), we don't break this setup

Habbie: Ok, then how about shipping compat-lua.hpp that we can symlink in to the right projects, to avoid the duplication?

pieterlexis: this file is created at build-time, not distribution time. We could have a `lua_hpp.mk` that is symlinked and included into the makefiles?

Habbie: :+1:

Habbie:
This branch has conflicts that must be resolved.

Habbie:
Other than that we are still duplicationg the lua C++ header (which I can overlook), this LGTM.

pieterlexis:
@Habbie fixed! I'll try another build on OpenSolaris soon to verify it all still works

pieterlexis:
Confirmed this allows building the Authoritative Server, Recursor and dnsdist on OpenIndiana Hipster.

PR #3407, opened 2016-02-19
Permit star entry

cmouse:


cmouse:
@zeha can you review please?

zeha:
NAK, this function is called for both zone names and full qnames. To fix this properly, please have a new function that only deals with qnames and splits those first to ensure that only the left-most part can be a (single) star.

cmouse:
@zeha is this better?

pdns/ws-api.cc
```diff
@@ -283,7 +283,23 @@ string apiZoneNameToId(const DNSName& dname) {
   return id;
 }
 
-void apiCheckNameAllowedCharacters(const string& label) {
-  if (label.find_first_not_of("abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ01234567890_/.-") != std::string::npos)
-    throw ApiException("Label '"+label+"' contains unsupported characters");
+void apiCheckNameAllowedCharacters(const string& name) {
+  if (name.find_first_not_of("abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ01234567890_/.-") != std::string::npos)
+    throw ApiException("Name '"+name+"' contains unsupported characters");
+}
```
cmouse: I'll remove this change when everything is OK

zeha:
Yes, but I guess I'd put the whitelist in a global const and again use `find_first_not_of` instead of hand-coding the loop.

cmouse:
the hand-coded loop also moves the ptr forward...

zeha:
I don't see why you actually need that. The check could probably be "if (leftmost2 == "*.") pos += 2;" + find_first_not_of and done?

zeha:
You've short-circuited the need for splitting with the if lefmost2 == "*.* anyway :)

cmouse:
I did think of just using this alternatively, you can choose
```
void apiCheckQNameAllowedCharacters(const string& qname) {
  string::size_type pos;
  if (qname.compare(0, 2, "*.") == 0) pos = 2;
  else pos = 0;
  apiCheckNameAllowedCharacters(qname.substr(pos));
}
```

zeha:
That probably works as well, but copies the string :-)

cmouse:
so we go with a1844c1 or 860e89c? I reduced the packet copying to occur with star only.

zeha:
LGTM

cmouse:
Ready for merge once CI passes.

zeha:
Also - please add a test for this...

cmouse:
test added.

PR #3408, opened 2016-02-19
dnsdist: Always create the default pool, clean related log messages

rgacogne:


rgacogne:
As noted by @pieterlexis I should instead use that opportunity to rename the default pool to "default".

PR #3412, opened 2016-02-19
Dnsdist check config

pieterlexis:
This allows testing of the configuration before one will try to restart
dnsdist with a broken config. Additionally, add tests to confirm the
config check still works.


pdns/dnsdist.cc
```diff
@@ -1303,6 +1308,13 @@ try
     g_ACL.setState(acl);
   }
 
+  if (g_cmdLine.checkConfig) {
+    setupLua(true, g_cmdLine.config);
+    // No exception was thrown
+    infolog("Configuration '%s' OK!", g_cmdLine.config);
```
rgacogne: Just one thing, setupLua() outputs a warning but does not throw an exception if the configuration file does not exist. I believe that's fine, but that means we will gladly validate a non-existent file, which might not be what you had in mind.

pieterlexis: An empty config is a valid config was my reasoning :)

PR #3415, opened 2016-02-20
pdnsutil clearzone allows operator to remove all records of a zone but leave everything else in place

ahupowerdns:
Used when perhaps pdns_control retrieve does not work but the records have to go away PDQ


zeha:
maybe document that such zones are no longer usable over httpapi

PR #3417, opened 2016-02-20
Secure and Rectify Zones When they are Created or Modified Through API Calls (POST/PATCH)

nils-wisiol:
This PR extends existing API functionality of POST and PATCH zones by optionally sign and rectify them. To do so, the patch honors the `dnssec`, `nsec3param` and `nsec3narrow` parameters of the API request on POST and stores them. If anything goes wrong with signing or rectifying, the API throws an APIException and aborts the transaction -- the zone will not be inserted. On a PATCH, the zone is rectified again.

Technical note: the code for signing and rectifying was taken from pdnsutil. Due to the amount of code needed for rectifying, it was moved to the DNSSECKeeper to use the same code in pdnsutil and in the API.

The PR does not cause any more gmysql test cases to fail than before. Failing test cases are:

```
0dyndns-prereq-all
0dyndns-prereq-nxrrset-full
1dyndns-big-package
1dyndns-check-soa-update
1dyndns-update-add-delete
1dyndns-update-add-delete-casesensative
1dyndns-update-add-delete-cname
1dyndns-update-add-delete-ds
1dyndns-update-add-delete-mx
1dyndns-update-add-delete-wildcard
1dyndns-update-add-invalid-record
1dyndns-update-add-soa
1dyndns-update-deep-add-delete
1dyndns-update-deep-delegate
1dyndns-update-delegate
1dyndns-update-delegate-in-between
1dyndns-update-delete-add-host
1dyndns-update-delete-multi-add-host
1dyndns-update-delete-mx-prio
1dyndns-update-delete-ns
1dyndns-update-delete-parent-delegate
1dyndns-update-delete-soa
1dyndns-update-in-between
1dyndns-update-nsec3params
1dyndns-update-nsec3params-with-others
1dyndns-update-replace-a-host
1dyndns-update-replace-cname
1dyndns-update-replace-mx
1dyndns-update-srv
1dyndns-update-update-ttl
2dyndns-update-replace-soa
tsig-axfr
tsig-axfr-algorithm-mismatch
tsig-axfr-key-mismatch
tsig-ixfr
tsig-ixfr-algorithm-mismatch
tsig-ixfr-key-mismatch
```

I can edit the commit messages if desired.

Please let me know if any more editing is needed in order to pass this PR.

### DIFF
--- a/pdns/pdnsutil.cc
+++ b/pdns/pdnsutil.cc
@@ -149,29 +149,7 @@ bool rectifyZone(DNSSECKeeper& dk, const DNSName& zone)
   SOAData sd;
 
   if(!B.getSOAUncached(zone, sd)) {
-    cerr<<"No SOA known for '"<<zone<<"', is such a zone in the database?"<<endl;
-    return false;
-  }
-  sd.db->list(zone, sd.domain_id);
-
-  DNSResourceRecord rr;
-  set<DNSName> qnames, nsset, dsnames, insnonterm, delnonterm;
-  map<DNSName,bool> nonterm;
-  vector<DNSResourceRecord> rrs;
-
-  while(sd.db->get(rr)) {
-    rr.qname.makeUsLowerCase();
-    if (rr.qtype.getCode())
-    {
-      rrs.push_back(rr);
-      qnames.insert(rr.qname);
-      if(rr.qtype.getCode() == QType::NS && rr.qname != zone)
-        nsset.insert(rr.qname);
-      if(rr.qtype.getCode() == QType::DS)
-        dsnames.insert(rr.qname);
-    }
-    else
-      delnonterm.insert(rr.qname);
+    cerr<<"No SOA known for '"<<zone.toString()<<"', is such a zone in the database?"<<endl;
   }
 
   NSEC3PARAMRecordContent ns3pr;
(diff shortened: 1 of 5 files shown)


pdns/pdnsutil.cc
```diff
@@ -139,145 +139,10 @@ bool rectifyZone(DNSSECKeeper& dk, const DNSName& zone)
...
-  bool narrow;
-  bool haveNSEC3=dk.getNSEC3PARAM(zone, &ns3pr, &narrow);
-  bool isOptOut=(haveNSEC3 && ns3pr.d_flags);
-  if(sd.db->doesDNSSEC())
-  {
-    if(!haveNSEC3)
-      cerr<<"Adding NSEC ordering information "<<endl;
-    else if(!narrow) {
-      if(!isOptOut)
-        cerr<<"Adding NSEC3 hashed ordering information for '"<<zone.toString()<<"'"<<endl;
```
Habbie: It looks like these messages are completely gone after your pull?

zeha:
Thank you for submitting this PR. It's a long-standing TODO and indeed we should be doing this sooner than later. As @Habbie noted elsewhere, we've been holding off doing this for various reasons, but we can work on them later after it works as imagined :-)

I have some comments on your changes; if you want/need clarifications or help feel free to drop by on IRC for a more real-time interaction.

- 550dd18 and 7a9dd50 should be a single commit, because those two just move code around.
- Creating slave zones via API appears to fail now (see travis)
- Please add at least one test in regression-tests.api with, say, dnssec=true and ensuring that the zone is indeed set up correctly afterwards. More tests, probably with different nsec3param, nsec3narrow settings would be good as well.
- Calling stringFromJson with an empty string as the default is equivalent to just calling `document["nsec3param"].string_value()`.
- In apiServerZones all checking and throwing needs to go before `B.createDomain`, as otherwise a broken domain is left over in the DB.
- The error messages for dk.addKey and dk.isSecuredZone failing could be more specific. With pdnsutil there is/was probably more output, and an admin might just look into config files or the DB for clues, where with the API the user might only get that single error message.
- Error checking on the rectifyZone call in patchZone is missing.

zeha:
Documentation also needs an update saying what's supported now.

Habbie:
By the way, all gmysql tests pass for us on master. If they don't pass for you, perhaps you are missing some tools (like `nsupdate`).

Habbie:
What @zeha is referring to that I said elsewhere is: we've always felt that the right approach is to do a surgically small rectify around the actual update (and I still feel that way). However, because of that idea, we basically forgot to consider your approach, i.e. a full rectify after each update. We have given this some thought today and indeed we want to go forward with your approach. For small zones it will be just fine, for big zones it will make updates slower, but most people have small zones and thus a full rectify is much better than no rectify. Great work!

nils-wisiol:
Thanks for the feedback, it is highly appreciated. I'll start working on the mentioned points as soon as possible and then get back to you.

nils-wisiol:
@zeha, I would like to move error checking in `ApiServerZones` before `B.createDomain`. Error checking though is partially done by `DNSSECKeeper::setNSEC3PARAM`, which in turn returns false if some database query fails. The query requires the nsec3data to be in the database, which I believe requires the domain to exist.

Why is `B.createDomain` called before the transaction is started? Can't we fix this by enclosing the creation of the domain in the transaction as well?

`DNSSECKeeper::setNSEC3PARAM` btw sometimes throws, sometimes returns false on failure. Do we need to handle them differently?

nils-wisiol:
@zeha, I did some research on the `dk.addKey` and `dk.isSecuredZone` errors. Essentially `addKey` fails, if DNSSEC is not supported by any backend (this is what the current error message says). `isSecuredZone` fails, if the zone was presigned (which we check before) or the zone has too many non-empty terminals (which I now included into the error msg).

I'd leave the `addKey` message unchanged, but I'm also open for concrete suggestions.

zeha:
> @zeha, I would like to move error checking in ApiServerZones before B.createDomain. Error checking though is partially done by DNSSECKeeper::setNSEC3PARAM, which in turn returns false if some database query fails. The query requires the nsec3data to be in the database, which I believe requires the domain to exist.
Okay, that's a bit unfortunate. I think the only part that we can and should care about is the max-nsec3-iterations check. Maybe we can split setNSEC3PARAM into one function for checking and one for setting.

> Why is B.createDomain called before the transaction is started? Can't we fix this by enclosing the creation of the domain in the transaction as well?
It's my understanding that beginTransaction already needs the zone Id, and we get that by calling createDomain.

> DNSSECKeeper::setNSEC3PARAM btw sometimes throws, sometimes returns false on failure. Do we need to handle them differently?

It appears to return false on catastrophic failure/misconfiguration. That's odd. Maybe it's behaviour will be a lot clearer if the function is split as suggested above.

Thanks!

Habbie:
This branch has conflicts

Habbie:
This branch still has conflicts.

PR #3422, opened 2016-02-21
Multispoof: make the dnsdist spoofing actions accept multiple IPv4 and IPv6 addresses, retaining old syntax + updated regression tests

ahupowerdns:


ahupowerdns:
after some discussion we'll leave it like it is - unless you specify IPv6 we won't overrule an AAAA query

PR #3426, opened 2016-02-21
Ensure consistent TTLs for RRsets

zeha:


zeha:
More an issue-PR, not sure I'll have time to work more on this. Also doesn't come with tests (for now?).

mind04:
I'm no fan of polishing bad data while answering queries. In the end it will only result in more and harder to find problems.
BTW what problem are we trying to fix here?

hlindqvist:
I would assume the intention is just to comply with https://tools.ietf.org/html/rfc2181#section-5.2

mind04:
I'm well aware of the rfc. The question is, is it just nice to follow the rfc or are we fixing a real life problem here?

- If it is just nice, it should be fixed db side (provide consistent data). Polish rrsets in the API before they enter the db and ....
- If there is a real life problem, I'm maybe in for a fix like this but it should be located in lookup()/get(), and not in a zillion places al over the code, and the resulting ttl  should be stable.

zeha:
@mind04 I agree that my implementation is shit. Also I can't remember what the issue prompting this was; will close this now, and somebody else can work on it if needed.

PR #3428, opened 2016-02-22
API: Fix zone/records design mistake

zeha:
Will break any existing 4.0-alpha api-v1 users.


zeha:
Note that this assumes the TTL merging logic from my `consistentttl` branch - i.e. lowest TTL wins.

PR #3437, opened 2016-02-23
add pdns.now timeval struct to recursor lua

Habbie:
Not sure this is the best way - perhaps adding a `pdns.now` just containing that int (or float) would be better?

UPDATE: have that now.

### DIFF
--- a/pdns/lua-recursor4.cc
+++ b/pdns/lua-recursor4.cc
@@ -318,7 +318,7 @@ RecursorLua4::RecursorLua4(const std::string& fname)
       theL() << (Logger::Urgency)loglevel << msg<<endl;
     });
   typedef vector<pair<string, int> > in_t;
-  vector<pair<string, boost::variant<int, in_t > > >  pd{
+  vector<pair<string, boost::variant<int, in_t, struct timeval* > > >  pd{
     {"PASS", (int)PolicyDecision::PASS}, {"DROP",  (int)PolicyDecision::DROP},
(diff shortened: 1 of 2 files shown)


Habbie:
Ready for review and merge.

pdns/lua-recursor4.cc
```diff
@@ -54,6 +54,8 @@ int RecursorLua4::gettag(const ComboAddress& remote, const ComboAddress& local,
 #undef L
 #include "ext/luawrapper/include/LuaContext.hpp"
 
+extern __thread struct timeval g_now; // timestamp, updated (too) frequently
```
ahupowerdns: this is asking for problems since it might get out of sync, perhaps put this in syncres.hh?

Habbie: will do

pdns/lua-recursor4.cc
```diff
@@ -336,6 +338,10 @@ RecursorLua4::RecursorLua4(const std::string& fname)
   
   for(const auto& n : QType::names)
     pd.push_back({n.first, n.second});
+  pd.push_back({"now", &g_now});
```
ahupowerdns: neat

pdns/lua-recursor4.cc
Habbie: done

Habbie:
Nit handled, squashed.

PR #3438, opened 2016-02-23
Add compatibility with selinux to the systemd unit file

mortenstevens:
PowerDNS auth fails to start with "NoNewPrivileges=true" if selinux is set to enforcing.
See: https://bugzilla.redhat.com/show_bug.cgi?id=1305522

ProtectSystem=full caused also selinux related warnings:
type=AVC msg=audit(1456267226.894:458): avc:  denied  { mounton } for  pid=3776 comm="(s_server)" path="/etc" dev="dm-0" ino=134217857 scontext=system_u:system_r:init_t:s0 tcontext=system_u:object_r:etc_t:s0 tclass=dir permissive=0

I suggest to disable both settings by default.

### DIFF
--- a/contrib/systemd-pdns.service
+++ b/contrib/systemd-pdns.service
@@ -12,10 +12,6 @@ StartLimitInterval=0
 PrivateTmp=true
 PrivateDevices=true
 CapabilityBoundingSet=CAP_NET_BIND_SERVICE CAP_SETGID CAP_SETUID
-NoNewPrivileges=true
-# ProtectSystem=full will disallow write access to /etc and /usr, possibly
-# not being able to write slaved-zones into sqlite3 or zonefiles.
-ProtectSystem=full
 ProtectHome=true
 RestrictAddressFamilies=AF_UNIX AF_INET AF_INET6
 


Habbie:
Disabling relevant systemd security features because they break with SELinux looks wrong to me. Are you sure this is the best approach?

mortenstevens:
@Habbie 
Sure, this is not the best approach and we need to investigate if it is a systemd or selinux issue.
It is enough to remove only "NoNewPrivileges=true". 

ProtectSystem=full is not a problem. We just need to modify the selinux policy.

rubenk:
I doubt this has anything to do with SELinux.

From systemd.exec(5):
```
NoNewPrivileges=
           Takes a boolean argument. If true, ensures that the service process and all its children can never gain new privileges. This option is more powerful than the respective secure bits flags (see above), as it also prohibits UID changes of any kind. This is the simplest, most effective way to ensure that a process and its children can never elevate privileges again.
```

Note the "it also prohibits UID changes of any kind." part. If the documentation is correct, perhaps this could be the issue?

rgacogne:
@rubenk It might not be correct, see https://bugs.freedesktop.org/show_bug.cgi?id=81959

rubenk:
@rgacogne thanks, and you're right:

With selinux=0 on the kernel command line:
```
ruben@rubenkerkhof: ~$ sudo systemctl start pdns
ruben@rubenkerkhof: ~$ sudo systemctl status pdns
● pdns.service - PowerDNS Authoritative Server
   Loaded: loaded (/usr/lib/systemd/system/pdns.service; disabled; vendor preset: disabled)
   Active: active (running) since Wed 2016-02-24 12:29:39 CET; 1s ago
     Docs: man:pdns_server(8)
           man:pdns_control(8)
  Process: 1032 ExecStart=/usr/sbin/pdns_server --daemon (code=exited, status=0/SUCCESS)
 Main PID: 1035 (pdns_server)
    Tasks: 8 (limit: 512)
   CGroup: /system.slice/pdns.service
           └─1035 /usr/sbin/pdns_server --daemon

Feb 24 12:29:39 rubenkerkhof.com pdns[1035]: Unable to enable timestamp reporting for socket
Feb 24 12:29:39 rubenkerkhof.com pdns[1035]: UDPv6 server bound to [::]:53
Feb 24 12:29:39 rubenkerkhof.com pdns[1035]: TCP server bound to 0.0.0.0:53
Feb 24 12:29:39 rubenkerkhof.com pdns[1035]: TCPv6 server bound to [::]:53
Feb 24 12:29:39 rubenkerkhof.com pdns[1035]: PowerDNS Authoritative Server 4.0.0-alpha1 (C) 2001-2015 PowerDNS.COM BV
Feb 24 12:29:39 rubenkerkhof.com pdns[1035]: Using 64-bits mode. Built using gcc 6.0.0 20160201 (Red Hat 6.0.0-0.9) on Feb  4 2016 18:06:23 by buildbot@7a068cdb837b.
Feb 24 12:29:39 rubenkerkhof.com pdns[1035]: PowerDNS comes with ABSOLUTELY NO WARRANTY. This is free software, and you are welcome to redistribute it according to the terms of the GPL version 2.
Feb 24 12:29:39 rubenkerkhof.com pdns[1035]: Creating backend connection for TCP
Feb 24 12:29:39 rubenkerkhof.com pdns[1035]: About to create 3 backend threads for UDP
Feb 24 12:29:39 rubenkerkhof.com pdns[1035]: Done launching threads, ready to distribute questions
```

rubenk:
I looked into this some more. SELinux does not handle ProtectSystem=full and NoNewPrivileges=true in our unit file properly, hence the AVCs. This is an upstream SELinux policy issue, which Morten already reported. FWIW I agree with @Habbie in that disabling these features is wrong and this should be solved in policy.

PR #3440, opened 2016-02-24
Client operation improvements for dnsdist

pieterlexis:
These commits allow the user to specify the dnsdist instance to connect to on the command line and what key to use.

### DIFF
--- a/pdns/dnsdist.cc
+++ b/pdns/dnsdist.cc
@@ -1199,6 +1199,7 @@ try
   }
   
 #endif
+  ComboAddress clientAddress = ComboAddress();
   g_cmdLine.config=SYSCONFDIR "/dnsdist.conf";
   struct option longopts[]={ 
     {"acl", required_argument, 0, 'a'},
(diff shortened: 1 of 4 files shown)


pieterlexis:
@rgacogne: ready for review

pdns/dnsdist.cc
```diff
@@ -1199,6 +1202,7 @@ try
   }
   
 #endif
+  ComboAddress clientAddress = ComboAddress();
```
rgacogne: Just a nit, perhaps `clientAddress` could be a member of `g_cmdLine`?

pieterlexis: not really, this is because I overload the argv arguments usually used for the backend servers [here](https://github.com/PowerDNS/pdns/pull/3440/files#diff-194369649d922d5a26ab47ac9c895d33R1322)

rgacogne: Sure, I had something like [1] in mind, but I'm ok either way.

[1]: https://github.com/pieterlexis/pdns/compare/dnsdist-client-improvements...rgacogne:dnsdist-client-improvements?expand=1

pieterlexis: I was pondering this last night and it might be cleaner to do it that way yes.

ahupowerdns:
so should this be merged or does @rgacogne have a better pull up his sleeve?

pieterlexis:
this works fine, we can always make this prettier later

PR #3453, opened 2016-02-25
Silence a BIO_write warning

pieterlexis:


pdns/base64.cc
```diff
@@ -44,7 +44,8 @@ std::string Base64Encode(const std::string& src)
     bio = BIO_new(BIO_s_mem());
     bio = BIO_push(b64, bio);
     BIO_set_flags(bio, BIO_FLAGS_BASE64_NO_NL);
-    if (BIO_write(bio, src.c_str(), src.length()) != src.length()) {
+    int bioWriteRet = BIO_write(bio, src.c_str(), src.length());
+    if (bioWriteRet < 0 || (uint) bioWriteRet != src.length()) {
```
zeha: cast to size_t instead of uint, afterall that's the .length() type AFAIK

PR #3466, opened 2016-02-27
dnsdist: add dq binding for skipping packet cache in LuaAction

janeczku:


rgacogne:
Thanks! Could you add this new binding to the README too?

janeczku:
Added to README

PR #3472, opened 2016-02-28
Improve compatibility for old compilers & libs in MTasker

nlyan:
Fixes MTasker (recursor) build against GCC 4.7 and Boost versions <= 1.51. With respect to the latter we fallback automagically to System V context switching.

Tested against the following:
- Debian Wheezy (GCC 4.7, Boost 1.49)
- Debian Jessie (GCC 4.9, Boost 1.55)
- Arch Linux (GCC 5.3, Boost 1.60)

### DIFF
--- a/pdns/recursordist/Makefile.am
+++ b/pdns/recursordist/Makefile.am
@@ -85,7 +85,7 @@ pdns_recursor_SOURCES = \
 	misc.hh misc.cc \
 	mplexer.hh \
 	mtasker.hh \
-	mtasker_context.hh mtasker_fcontext.cc \
+	mtasker_context.hh mtasker_context.cc \
 	namespaces.hh \
 	nsecrecords.cc \
 	opensslsigners.cc opensslsigners.hh \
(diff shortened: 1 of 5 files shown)


nlyan:
Damnit, Travis didn't like my #include hack

ahupowerdns:
@nlyan - I think I fixed the travis make dist issue. But could you rebase?

pdns/recursordist/Makefile.am
```diff
@@ -85,7 +86,7 @@ pdns_recursor_SOURCES = \
 	misc.hh misc.cc \
 	mplexer.hh \
 	mtasker.hh \
-	mtasker_context.hh mtasker_fcontext.cc \
+	mtasker_context.hh mtasker_context.cc \
```
rubenk: Not wanting to be a PITA, but could you swap these, so source file first, then header?

nlyan: Done

PR #3476, opened 2016-02-28
add missing bracket  in configure.ac

mind04:


rubenk:
LGTM, and fixes my build on EL6.

PR #3477, opened 2016-02-28
Fix pgsql backend compilation on centos 6

pieterlexis:


modules/gpgsqlbackend/spgsql.cc
```diff
@@ -96,10 +96,16 @@ class SPgSQLStatement: public SSqlStatement
     // by stored procedures. you can return more than one
     // if you return SETOF refcursor.
     if (PQftype(d_res_set, 0) == 1790) { // REFCURSOR
+#if PG_VERSION_NUM > 90000
+      // PQescapeIdentifier was added to libpq in postggresql 9.0
       char *val = PQgetvalue(d_res_set, d_cur_set++, 0);
       char *portal =  PQescapeIdentifier(d_db(), val, strlen(val));
       string cmd = string("FETCH ALL FROM \"") + string(portal) + string("\"");
       PQfreemem(portal);
```
pieterlexis: I _think_ this line should be after `#endif`

cmouse:
Hopefully the portal value won't contain escapable characters...

ahupowerdns:
is this ok for merge?

modules/gpgsqlbackend/spgsql.cc
rubenk: Pff, what I meant to say was, there comment should say PostgreSQL, not postggresql.

Issue #3506, opened 2016-03-04
api: Filter api-key value from /config view 


terrorobe:
We're already filtering sensitive information like the webserver-password, we should do the same for the api-key.

Example:

```
root@pdnstest:/etc/dnsdist# curl  -H 'X-API-Key: YOLIDDY' http://127.0.0.1:8082/api/v1/servers/localhost/config  | python -m json.tool | egrep -iA3 '(pass|key)'
        "name": "api-key",
        "type": "ConfigSetting",
        "value": "YOLIDDY"
    },
--
        "name": "webserver-password",
        "type": "ConfigSetting",
        "value": "***"
    },
root@pdnstest:/etc/dnsdist# 
```

PR #3507, opened 2016-03-04
API: Mask out api-key in settings list

zeha:


zeha:
Closes #3506.

Note that I don't think this is a documented feature.

PR #3514, opened 2016-03-05
propagate calculated packet cache tag to preresolve (the other hooks …

ahupowerdns:
…could get it with a typing exercise)

### DIFF
--- a/pdns/lua-recursor4.cc
+++ b/pdns/lua-recursor4.cc
@@ -30,7 +30,7 @@ bool RecursorLua4::postresolve(const ComboAddress& remote,const ComboAddress& lo
 }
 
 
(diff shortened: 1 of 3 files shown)


ahupowerdns:
```
filterips = newNMG()
filterips:addMask("192.168.1.244")

filterdomains = newDS()
filterdomains:add{"org", "li"}

function gettag(remote, subnet, loc, qname, qtype)
        print("Called for "..remote:toString()..", "..qname:toString())
        if(not subnet:empty())
        then
                remote=subnet:getNetwork()
                print("Got a subnet: "..subnet:toString()..", set remote to: "..remote:toString());
        end
        if(filterips:match(remote) and filterdomains:check(qname)) then
                magicMetric:inc()
                print("Returning tag 1")
                return 1
        end
        print("Returning tag 0")
        return 0;
end

function preresolve(dq)
        print("Tag is set as: "..dq.tag)
        if(dq.tag == 1) then
                dq:addAnswer(pdns.A, "1.2.3.4")
                return true;
        end

        return false;
end
```

PR #3522, opened 2016-03-07
dnsdist: Per-carbon server hostnames

rgacogne:
We used the last specified `ourname` for all servers. We still use
the first specified interval for all servers for now.


rgacogne:
Now with regression tests.

Habbie:
Merge conflicts.

PR #3538, opened 2016-03-10
dnsdist: Add API regression tests

rgacogne:
So we'll notice if we break the web server or the API.
This commit also cleans up the responses counter usage.
Because the counters are written to by the responder threads,
they are updated at the class level. We do not want to alter
them at the instance level, because that would create a new variable
at this level, disconnected from the other one, but we can still
read them from the instance level using 'self'.

### DIFF
--- a/regression-tests.dnsdist/requirements.txt
+++ b/regression-tests.dnsdist/requirements.txt
@@ -1,3 +1,4 @@
 dnspython>=1.11
 nose>=1.3.7
 libnacl>=1.4.3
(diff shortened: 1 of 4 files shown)


zeha:
Not sure you really want to test /jsonstat for now, (it probably should go away). OTOH, knowing that it breaks is useful too...

rgacogne:
I agree /jsonstat will probably go away, but yeah, it doesn't hurt to check that we did not break it accidentally while it's there :)

Issue #3531, opened 2016-03-09
Building pdns-recursor fails on arm64, s390x because of missing boost-context


zeha:
Debian stretch has (at the time of writing) boost 1.58 which has no implementation of Boost.Context for arm64 or s390x.

The configure check doesn't seem to work in that it just fails fatally and doesn't choose the fallback implementation.

Note that the Debian libboost-context1.58-dev packages exist but are empty on arm64, s390x.

Habbie:
Should be fixed by #3554

PR #3554, opened 2016-03-11
Fix Boost component library detection on ARM64 etc.

### LINKED ISSUE #3536: PowerDNS Recursor 4.0.0 Alpha 2: ./configure fails
```
...
checking boost/context/all.hpp usability... no
checking boost/context/all.hpp presence... yes
configure: WARNING: boost/context/all.hpp: present but cannot be compiled
configure: WARNING: boost/context/all.hpp:     check for missing prerequisite headers?
configure: WARNING: boost/context/all.hpp: see the Autoconf documentation
configure: WARNING: boost/context/all.hpp:     section "Present But Cannot Be Compiled"
configure: WARNING: boost/context/all.hpp: proceeding with the compiler's result
checking for boost/context/all.hpp... no
configure: error: cannot find boost/context/all.hpp
```
It seems only Boost<1.59 is affected.
[config_log.txt](https://github.com/PowerDNS/pdns/files/167303/config_log.txt)

nlyan:
On some platforms the Boost context headers are being shipped on architectures where the library can't be built. Here we patch boost.m4 to set HAVE_BOOST_<component> definitions based on whether the library actually exists.

This should also fix #3536 by avoiding the context/all.hpp header check.

### DIFF
--- a/m4/boost.m4
+++ b/m4/boost.m4
@@ -328,14 +328,14 @@ CPPFLAGS="$CPPFLAGS $BOOST_CPPFLAGS"
 AC_CACHE_CHECK([for the Boost $1 library], [Boost_lib],
                [_BOOST_FIND_LIBS($@)])
 case $Boost_lib in #(
-  (no) _AC_MSG_LOG_CONFTEST
-    AC_MSG_ERROR([cannot find the flags to link with Boost $1])
+  (yes) _AC_MSG_LOG_CONFTEST
+    AC_DEFINE(AS_TR_CPP([HAVE_BOOST_$1]), [1], [Defined if the Boost $1 library is available])dnl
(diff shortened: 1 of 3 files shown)


rubenk:
Thanks, LGTM.

nlyan:
Fixes #3531

Habbie:
@nlyan thanks! Want to upstream those boost.m4 changes?

nlyan:
@Habbie : yes, it's something I will look in to, otherwise it'll cause grief down the line

nlyan:
Upstreamed PRs relevent to this issue.

https://github.com/tsuna/boost.m4/pull/85

https://github.com/tsuna/boost.m4/pull/86

PR #3557, opened 2016-03-12
pdns: Fix build on OSX

rubenk:
Using OpenSSL from Homebrew, by doing
./configure --with-openssl=/Users/ruben/homebrew/Cellar/openssl/1.0.2g

Results in:

Making all in pdns
/Library/Developer/CommandLineTools/usr/bin/make  all-am
  CXX      base64.o
base64.cc:7:10: fatal error: 'openssl/bio.h' file not found
         ^
1 error generated.


Habbie:
Interestingly I recall this used to work. Thanks in any case :)

PR #3565, opened 2016-03-14
Small fixes reported by coverity

rgacogne:
I would appreciate any comment, especially regarding the NSEC3*RecordContent changes. The removed fields don't seem to be used, but I'm really not familiar with this code and thus might very well be missing something.

### DIFF
--- a/pdns/syncres.cc
+++ b/pdns/syncres.cc
@@ -283,6 +283,9 @@ bool SyncRes::doOOBResolve(const DNSName &qname, const QType &qtype, vector<DNSR
 void SyncRes::doEDNSDumpAndClose(int fd)
 {
   FILE* fp=fdopen(fd, "w");
+  if (!fp) {
+    return;
+  }
   fprintf(fp,"IP Address\tMode\tMode last updated at\n");
   for(const auto& eds : t_sstorage->ednsstatus) {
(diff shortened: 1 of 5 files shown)


pdns/syncres.cc
```diff
@@ -283,6 +283,9 @@ bool SyncRes::doOOBResolve(const DNSName &qname, const QType &qtype, vector<DNSR
 void SyncRes::doEDNSDumpAndClose(int fd)
 {
   FILE* fp=fdopen(fd, "w");
+  if (!fp) {
+    return;
```
cmouse: Should this cause some error/warning?

Habbie:
NSEC3 code looks good. We already use the 'int len' trick for NSEC3RecordContent and this syncs the PARAM code with that.

pdns/syncres.cc
rgacogne: You are right, but the error should get to the right channel, so we would need to return a error to the caller or move the `fdopen()` to the calling function. Similar functions (`MemRecursorCache::doDumpNSSpeeds()`, `MemRecursorCache::doDump()`, `RecursorPacketCache::doDump()`, ...) don't seem to care.

cmouse: :ok:

PR #3569, opened 2016-03-14
log name/qtype when sending out servfail

Habbie:
Should we check for 'empty qname' to avoid generating another exception? Should we log client? query ID?

### DIFF
--- a/pdns/dnsname.hh
+++ b/pdns/dnsname.hh
@@ -52,6 +52,7 @@ public:
   bool operator!=(const DNSName& other) const { return !(*this == other); }
 
   std::string toString(const std::string& separator=".", const bool trailing=true) const;              //!< Our human-friendly, escaped, representation
+  std::string toLogString() const; //!< like plain toString, but returns (empty) on empty names
   std::string toStringNoDot() const { return toString(".", false); }
   std::string toStringRootDot() const { if(isRoot()) return "."; else return toString(".", false); }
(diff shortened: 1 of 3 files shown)


pieterlexis:
Checking for empty might be a good idea

ahupowerdns:
checking for empty is mandatory when logging errors I'm afraid unless we make toString() do it for us.

Habbie:
Updated, please review.

pdns/dnsname.hh
```diff
@@ -52,6 +52,7 @@ public:
   bool operator!=(const DNSName& other) const { return !(*this == other); }
 
   std::string toString(const std::string& separator=".", const bool trailing=true) const;              //!< Our human-friendly, escaped, representation
+  std::string toLogString() const; //!< like plain toString, but returns [empty] on empty names
```
pieterlexis: "but returns [empty] on empty names" => "but returns (empty) on empty names"

pieterlexis:
We might need to review all lines that log DNSNames to have consistent output. Other than that, this looks nice

pdns/dnsname.hh
Habbie: fixed

Habbie:
Yes, review would be nice but is not a holdup for this PR :)

pieterlexis:
indeed, needs a ticket though :)

PR #3571, opened 2016-03-14
strip all types of whitespace from around content while parsing zones, fixes #3568

### LINKED ISSUE #3568: PowerDNS bind zone parsing
We use the Net::DNS perl module to format DNS records in a consistent way. This results for TXT records with long values in something like this:

```
bla       IN      TXT     (
          "v=spf1 include:spf.protection.outlook.com include:mail.foobar.nl -all"
          )
```

When Power DNS read a zonefile with these kind of records it doesn't produce an error (that I can see) but when an AXFR comes for a zone with such a record we see in the log:

```
Mar 14 12:54:12 dnssec1 pdns[2723]: Exception: Parsing record content: Data
field in DNS should start with quote (") at position 3 of '" "v=spf1
include:spf.protection.outlook.com include:mail.foobar.nl -all" "'
Mar 14 12:54:12 dnssec1 pdns[2723]: TCP Connection Thread died because of STL
error: Parsing record content: Data field in DNS should start with quote (") at
position 3 of '" "v=spf1 include:spf.protection.outlook.com
include:mail.foobar.nl -all" "'
```

It seems that the AXFR is not completely failing, only the records that are failing are missing after the AXFR, other records are present and updated, including the SOA (with the serial).

In some other cases the TXT record is read and returned, but after AXFR (from the hidden master) to our public masters we notice extra spaces inside the TXT records between the quote characters. For example, a string like

```
@   IN TXT (
   "LongBase64EncodedString"
   )
```

would become

```
@   IN TXT " LongBase64EncodedString "
```

We use pdns-static 3.4.7-1 amd64, downloaded from downloads.powerdns.com on Debian Wheezy.

Habbie:
Review welcome!

### DIFF
--- a/pdns/zoneparser-tng.cc
+++ b/pdns/zoneparser-tng.cc
@@ -401,6 +401,7 @@ bool ZoneParserTNG::get(DNSResourceRecord& rr, std::string* comment)
       }
     }
   }
+  trim_if(rr.content, is_any_of(" \r\n\x1a"));
 
   vector<string> recparts;
   switch(rr.qtype.getCode()) {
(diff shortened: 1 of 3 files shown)


pieterlexis:
do it!

Issue #3447, opened 2016-02-24
Rename shipped utils to pdns-xyz?


zeha:
Fedora renames zone2ldap to pdns-zone2ldap because the bind ldap backend also has a zone2ldap tool and the names clash.

Maybe a good idea to do this before 4.0.0.

pieterlexis:
I like the idea, but on the other hand I consider this more of a "downstream" issue. We could also install `zone2*` into `/usr/lib/powerdns/bin/` as they are libexec style binaries.

zeha:
It's not really a downstream issue as the public docs then are all wrong for some downstreams :-)

I don't think they are libexec style binaries at all - they are really meant to be called by users. (Also note that libexec is a GNU thing that doesn't -really- exist on most Linux distributions, it's merely being faked in a bad way.)

rubenk:
I'm all for it.

ahupowerdns:
so to clarify - this is about which tools exactly? And it does not cover pdns_control, pdnsutil, as confusing as the names may be?

rubenk:
Here's what we ship in Fedora now:
```
%{_bindir}/pdns_control
%{_bindir}/pdns-zone2ldap
%{_bindir}/zone2sql
%{_bindir}/zone2json
%{_bindir}/calidns
%{_bindir}/dnsgram
%{_bindir}/dumresp
%{_bindir}/ixplore
%{_bindir}/notify
%{_bindir}/pdnsutil
%{_bindir}/sdig
```

zeha:
I guess zone2* are the tools that are the most interesting? In Debian we ship those (+pdnsutil, pdns_control) as part of pdns-server, and the other tools just go into pdns-tools and there I don't really care about names. (Downstream Debian packaging layout is almost the same as your upstream Debian packaging layout.)

zeha:
`notify` got renamed to `pdns_notify` recently. It was kinda agreed that `zone2*` should keep their names - although rereading this ticket I see Fedora already renames `zone2ldap` ...

PR #3572, opened 2016-03-14
Ship the notify tool in the pdns-tools deb

pieterlexis:
This commit fixes our testsuite on buildbot. However, the `pdns-tools` package now conflicts with `ruby-notify`, as they both ship `/usr/bin/notify`. See also #3447


Habbie:
LGTM

PR #3582, opened 2016-03-16
dnsdist: Refactor query/response handling code (UDP/TCP)

rgacogne:
Merge the UDP/TCP code handling:

* dynamic blocks
* blockfilter
* rules
* response consistency check
* response fixup (removing EDNS if needed, restoring flags, restoring case)

This fixes an issue with DNSCrypt, where cached responses were not
being encrypted, and make the counters more consistent between UDP and TCP.


rgacogne:
Removed the ugly #ifdef'ed definitions, fixed restoring flags for generated responses.

PR #3584, opened 2016-03-16
Update Makefile.am: do not swallow curl errors

sjvs:
The versions of Makefile.am execute 'curl -s' when fetching public_suffix_list.dat. However, if an error occurs during the retrieval of the URL, curl will swallow that error and silently fail. This PR adds the '-S' option, which will report details of an error, should one occur.

(In my case, the problem was caused by a transparent HTTPS proxy which was MITM'ing my connections using certificates that were generated on-the-fly. However, the root of those automatically-generated certificates was not in my machine's certificate store, which made curl fail silently)


pieterlexis:
:+1:

PR #3587, opened 2016-03-17
Move the root server addresses to a header

### LINKED ISSUE #3533: root server IP list is in 3 files
```
$ git grep 199.7.83.42
packethandler.cc:                     "192.36.148.17","192.58.128.30", "193.0.14.129", "199.7.83.42", "202.12.27.33"};
reczones.cc:                             "192.36.148.17","192.58.128.30", "193.0.14.129", "199.7.83.42", "202.12.27.33"};
speedtest.cc:                     "192.36.148.17","192.58.128.30", "193.0.14.129", "199.7.83.42", "202.12.27.33"};
```

should probably centralize that

### LINKED ISSUE #3534: [dns-operations] L-Root IPv6 Address renumbering
https://lists.dns-oarc.net/pipermail/dns-operations/2016-March/014489.html

> the new address is already operational.

Should probably just stick it in ASAP to go with the next Recursor release.

pieterlexis:
And update the L-Root address

Closes #3533
Closes #3534

### DIFF
--- a/pdns/root-addresses.hh
+++ b/pdns/root-addresses.hh
@@ -1,0 +1,53 @@
+/*
+    PowerDNS Versatile Database Driven Nameserver
+    Copyright (C) 2002 - 2016 PowerDNS.COM BV
+
+    This program is free software; you can redistribute it and/or modify
+    it under the terms of the GNU General Public License version 2
+    as published by the Free Software Foundation
+
+    Additionally, the license of this program contains a special
+    exception which allows to distribute the program in binary form when
+    it is linked against OpenSSL.
+
+    This program is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+    along with this program; if not, write to the Free Software
+    Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+*/
+
+#pragma once
+
+static const char*rootIps4[]={"198.41.0.4",             // a.root-servers.net.
+                              "192.228.79.201",         // b.root-servers.net.
+                              "192.33.4.12",            // c.root-servers.net.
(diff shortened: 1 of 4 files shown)


pdns/root-addresses.hh
```diff
@@ -0,0 +1,53 @@
...
+                              NULL,                     // e.root-servers.net.
+                              "2001:500:2f::f",         // f.root-servers.net.
+                              NULL,                     // g.root-servers.net.
+                              "2001:500:1::53",         // h.root-servers.net.
+                              "2001:7fe::53",           // i.root-servers.net.
+                              "2001:503:c27::2:30",     // j.root-servers.net.
+                              "2001:7fd::1",            // k.root-servers.net.
+                              "2001:500:9f::42",        // l.root-servers.net.
+                              "2001:dc3::35"            // m.root-servers.net.
+                              };
```
Habbie: While okay in practice (this file is only included once per project, in fact we only have one consumer left), the header should not contain the content - it should just announce its availability while the content should live in a .cc file so it ends up in one .o file.

nlyan:
Why not ship them as a zone file like Bind does?

zeha:
Ship as zone, get all my <3

pdns/root-addresses.hh
Habbie: I just realised that because they are `static`, the worst that could happen is that they get duplicated. There is no actual risk of symbol conflict. Feel free to keep it this way.

Habbie:
We do support reading root hints as a zone. We just also have a builtin copy in case you don't have one. Debian should feel free to default to pointing to the root hints shipped and updated with the distro.

pdns/root-addresses.hh
pieterlexis: isn't the duplication avoided with the `#pragme once`?

zeha: `#pragma once` is a preprocessor thing. Each compiler invocation will still compile the file and you may end up with one `static` copy of the data in each compilation unit that includes the header.

tfarina:
Would be fine to make the following?

root-addresses.hh
extern const char* rootIps6[];

root-addresses.cc
const char* rootIps6[]  {
...
};

It won't hurt, right? And will be slightly better, no?

PR #3596, opened 2016-03-18
remove unnecessary botan signers (rsa and ecdsa)

mind04:
OpenSSL is preferred for these algorithms so they are never used


Habbie:
I tested this, with and without Botan. Code also looks good. Merging.

PR #3609, opened 2016-03-22
dnsdist: Display the ACL content on startup

### LINKED ISSUE #3588: dnsdist should make it clearer that there is a limited ACL by default
perhaps mention the ACL at startup, or provide periodic feedback on number of dropped packets because of ACL. Documentation reminder is nice and we should do it, but nobody will read it.

rgacogne:
Closes #3588.


dol:
Thx

Issue #3430, opened 2016-02-22
Document v6 scoped syntax for pipebackend et al


Habbie:
In (broken) networking setups, pipebackend (and remotebackend) scripts could see source addresses ending in `%eth0` or similar. We should document this as it may confuse script authors.

PR #3618, opened 2016-03-23
Document v6 scoped syntax

pieterlexis:
Closes: #3430


paddg:
Just a note, we saw it reproducable with link-local (fe80) addresses. Thanks!

Issue #3602, opened 2016-03-21
Slave pdns server does not change master upon termination


oridistor:
I have a slave pdns with bind backend listening to several different masters for SOA updates and notifications. If the server is on and the first IP in the list is terminated, my slave pdns can't connect to it to get SOA updates even though he gets notified by other masters that a change need to be made.

This obviously gets solved if I restart my pdns server, in which case it takes the next ip in list, but the connection is always only kept with the first master.

I would like for the slave to either:
- Ask the notifier itself for the SOA update
- Make the slave use the bind masters list to choose another master after receiving the timeout.

I'll only give slave information as they are what's relevant:

named.conf file:

options {
    directory "/var/powerdns/bind";
};

zone "example.com" IN {
    type slave;
    file "example.com";
    masters { 52.91.135.125; 54.175.233.140; 54.208.97.41; };
};

pdns.conf file:

config-dir=/etc/powerdns
daemon=yes
guardian=yes
disable-axfr=yes
disable-tcp=yes
local-port=53
log-dns-details=on
loglevel=3
master=no
slave=yes
setgid=pdns
setuid=pdns
socket-dir=/var/run
version-string=powerdns
launch=bind
bind-config=/var/powerdns/bind/named.conf
include-dir=/etc/powerdns/pdns.conf.d
bind-check-interval=600
allow-notify-from=0.0.0.0/0
distributor-threads=3
negquery-cache-ttl=10
query-cache-ttl=20
receiver-threads=1
retrieval-threads=2
slave-cycle-interval=60

Thanks in advance

pieterlexis:
Updated post with code tags.

Can you please share your OS and PowerDNS version number?

oridistor:
os - ubuntu 14.04 - trusty.

PowerDNS version - 3.4.7

oridistor:
A file called "allow-notify-from" with an 'allow-notify-from=' tag with a lot of IPs, including the relevant ones that are up.

pieterlexis:
We should try the master the NOTIFY was received from, so if it goes down jsut after sending the NOTIFY, this could be an issue. But I assume this isn't the case here. Can you increase your `loglevel` to 6 and provide the logs of when this happens?

oridistor:
will do right away.

Do notice that the NOTIFY goes to the same server, even if that server is down. As I do dns changes on masters side after the first master is down, I can say for certain that in my case, the request is sent to first master and not the one it recieved the notify from.

[...]

PR #3631, opened 2016-03-28
Fixed issue #3602 by adding a tracking to who the notifyier is

oridistor:
Changed the Add Slave Notify command to move the needed master to the head of the list. It also keeps the DomainInfo the same until the sucker command finished. This has been tested.

### DIFF
--- a/debian-pdns/pdns.postinst
+++ b/debian-pdns/pdns.postinst
@@ -10,7 +10,7 @@ update_init() {
 
 
 update_pdns() {
-	if [ -n "$2" ] ; then
+	if [ -z "$2" ] ; then
 		invoke-rc.d pdns start
 	else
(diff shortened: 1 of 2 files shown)


Habbie:
I'm sorry, did you test this at all? From the commit history I get the impression you have pushed untested and even uncompiled code a few times.

oridistor:
Hey, I have tested it, and it failed. I was just a bit excited about pushing that I did it irresponsibally.

After some development, we have developed something that solved that issue for us, but we only know it works on bind (the algorithm leaves only the notifyier as master for the current suck, doesn't touch the backend, to leave refreshing working on normal basis).

I wonder if I should PR our fix, as it solves our problem completely, but I'm not sure if anyone but us complained about it.

Will make a new PR anyways.

PR #3642, opened 2016-03-31
Protobuf cleanups

pieterlexis:


m4/pdns_with_protobuf.m4
```diff
@@ -0,0 +1,28 @@
...
+  AC_MSG_RESULT([$with_protobuf])
+
+  AS_IF([test "x$with_protobuf" != "xno"], [
+    AS_IF([test "x$with_protobuf" = "xyes" -o "x$with_protobuf" = "xauto"], [
+      PKG_CHECK_MODULES([PROTOBUF], [protobuf], [HAVE_PROTOBUF=1], [HAVE_PROTOBUF=0])
+      AC_CHECK_PROG([PROTOC], [protoc], [protoc])
+    ])
+  ])
+  AS_IF([test "x$with_protobuf" = "xyes"], [
+    AS_IF([test "$HAVE_PROTOBUF" -eq 0], [
```
rubenk: Why don't you just check if $PROTOBUF_LIBS is not empty?
Then you can drop the whole HAVE_PROTOBUF variable.

Issue #3647, opened 2016-04-01
disable implicit notifications


klaus3000:
Hi!

In issue #1340 I asked about explicit notify and it was closed due to the "allow-notify" option. But the "only-notify" option is not suitable for my use cases, it would only be a bad workaround.

So what is the problem with only-notify? only-notify works on IP addresses/masks. Thus, before evaluating if the target is allowed, the target has to be resolved:
1. fetching NS records from the backend (probably fast)
2. resolving the host names (from fast to terrible slow).

With 2., PowerDNS suddenly becomes dependent on external systems, eg. resolver. If PowerDNS hosts customer zones and customers may put arbitrary hostnames in their NS records (eg a host name whose authoritative nameserver is not responding), PowerDNS waits several seconds (in my case 30 seconds) until the (stub?) resolver gives up.

In our setup we hosts 100.000s of customer zones and we do not have control over their NS records. Thus, if we allow to slave zones from our PowerDNS (we do not need NOTIFYs to the hosts in the NS, but only ALSO-NOTIFY.) by enabling slave-renotify=yes, someone could hurt our PowerDNS and PowerDNS may become very slow waiting for unneeded host resolving. Thus, disabling these NOTIFYs with only-notify blocking this NOTIFYs is not suitble for scenarios like ours.

Thus, I again propose a new option:
notify=yes # default: current behavior, implicit (triggered by NS records) and explicit (also-notify)
notify=no # never send NOTIFYs
notify=explicit # only send NOTIFYs to host specified with ALSO-NOTIFY metadata and also-notify=

I will send a pull request too.

pieterlexis:
`notify=no` would overlap with `master=no` (i.e. the combination `master=no` and `notify=yes` makes no sense). A `notify-type={default,explicit}` option might make more sense

klaus3000:
That would be fine too for me. We just should agree on the naming and I will send another pull request.

zeha:
As a temporary thing, you could set a short timeout in /etc/resolv.conf and possibly use resolvers with a short timeout.

zeha:
If I understand you right, you don't even want `master=yes`, is that correct?

klaus3000:
In my setup I only have slave zones which I "reslave" (PowerDNS is a bump in the wire). But the option is useful for masters too. Eg. if you distribute the zone to the slaves with other means (eg. DB replication) but need to have some also-notifys.

klaus3000:
[...]

PR #3648, opened 2016-04-01
patch for #3647: notify=yes|no|explicit option

klaus3000:
I applied the patch in the online editor of github, I hope I didn't broke the patch.


Habbie:
Please also update pdns.conf-dist. I did not check if there were other reasons that Travis is unhappy.

klaus3000:
Thanks for reviewing.

First we should decide on which patch to use, there is "notify=yes|no|explicit" or "notify-type=default|explicit" in #3649. Please see discussion in #3647.

Second, the pdns.conf-dist begins with "# Autogenerated configuration file template". Thus I expected that it will be autogenerated. Isn't it? Then we should remove this header too.

mind04:
Why do we need a new config option for this? Add some logic to skip the NS lookup if only-notify is empty and you are done.

And if we need one, i like a more self describing one, like 'notify-nsset=yes/no"

klaus3000:
I realized that there is no option needed. just fixed the only-notify option, see #3890

PR #3649, opened 2016-04-01
Add notify-type option to disable implicit notifications

klaus3000:
For outgoing slaved zones PDNS automatically notifies all the
host in the zones NS records. This is often not needed and imposes
external dependencies as PDNS will try to resolve every hostname
of the NS records.

Using notify-type=explicit disables this implicit notifications
and only send notifications to targets specified with also-notify
or ALSO-NOTIFY metadata.

Further the log warning was canged to info as it may be common to not have any explicit notifications and having implicit slaves disabled.


klaus3000:
AFAIS the travis check only failed because there were options added to pdns.conf

klaus3000:
I realized that there is no option needed. just fixed the only-notify option, see #3890

Issue #1907, opened 2014-12-06
pdns_server 3.4.1 Ubuntu abort using chroot: libgcc_s.so.1 must be installed


lifeforms:
When trying PowerDNS 3.4.1 on Ubuntu 12.04, we get a signal 6 (abort) if the `chroot` feature is used.

Using the static deb file from http://downloads.powerdns.com/releases/deb/pdns-static_3.4.1-1_amd64.deb.

Interesting part in the stack trace is `#7  0x00007f97b1cb211e in __GI___libc_fatal (message=0x7f97b250ede8 "libgcc_s.so.1 must be installed for pthread_cancel to work\n")`

The problem goes away when `libgcc_s.so.1` is copied into the chroot, e.g. `/var/spool/powerdns/lib/x86_64-linux-gnu/libgcc_s.so.1`. Then pdns appears to be completely stable.

Downgrading back to version 3.3 also resolves the problem.

Not much is logged to syslog.

pdns.conf is pretty trivial (also attempted with numeric uid/gids; loglevel 9 does not give more interesting info):

```
# This file is managed by Puppet.
# Any manual edits will be overwritten. Please contact support for changes.

daemon=yes
guardian=yes
launch=gmysql
gmysql-host=127.0.0.1
gmysql-dbname=powerdns
gmysql-user=powerdns
gmysql-password=XXX
local-ipv6=::  # for IPv6
setuid=pdns
setgid=pdns
chroot=/var/spool/powerdns
disable-axfr=yes
slave=yes
loglevel=3
log-dns-details=no
version-string=anonymous
webserver=no
```

The chroot directory is empty.

Syslog output:

```
pdns[2999]: Guardian is launching an instance
pdns[2999]: Reading random entropy from '/dev/urandom'
pdns[2999]: This is a guarded instance of pdns
pdns[2999]: UDP server bound to 0.0.0.0:53
pdns[2999]: UDPv6 server bound to [::]:53
pdns[2999]: TCP server bound to 0.0.0.0:53
pdns[2999]: TCPv6 server bound to [::]:53
pdns[2999]: PowerDNS Authoritative Server 3.4.1 (jenkins@autotest.powerdns.com) (C) 2001-2014 PowerDNS.COM BV
pdns[2999]: Using 64-bits mode. Built on 20141030110058 by root@autotest.powerdns.com, gcc 4.7.2.
[...]

PR #3669, opened 2016-04-05
Do triggerLoadOfLibraries() only in glibc environment

Lance0312:
The workaround of preloading libgcc_s.so.1 to memory by creating a dummy
thread from #1907 makes pdns fail to run in musl libc environment with
`setuid` and `chroot` enabled at the same time. musl libc [implements
set\*id calls AS-safe](http://git.musl-libc.org/cgit/musl/commit/?id=78a8ef47c4d92b7680c52a85f80a81e29da86bb9)
by reading `/proc/self/task` to get information of all threads. And
since most users `chroot` to an empty directory for security reason,
`setuid` cannot find `/proc/self/task` in the new root.

I've started [a discussion thread on musl's mailing list](http://www.openwall.com/lists/musl/2016/04/01/15),
however, until the Linux kernel provides truly atomic set\*id calls,
which is a really tough task according to Rich Felker, there seems no
easy way to keep set\*id AS-safe without using `/proc/self/task`.

Since #1907 is an issue of glibc, I think it is rather safe to do the
fix only in glibc environment, and as a consequence mitigate the
combinational use of `pthread`, `chroot`, and `setuid` in musl libc
environment.

### DIFF
--- a/pdns/common_startup.cc
+++ b/pdns/common_startup.cc
@@ -475,7 +475,9 @@ void mainthread()
    secPollParseResolveConf();
 
    if(!::arg()["chroot"].empty()) {
+#ifdef __GLIBC__
      triggerLoadOfLibraries();
+#endif
      if(::arg().mustDo("master") || ::arg().mustDo("slave"))
         gethostbyname("a.root-servers.net"); // this forces all lookup libraries to be loaded
      Utility::dropGroupPrivs(newuid, newgid);


Habbie:
Can you clarify how `triggerLoadOfLibraries()` is causing failure on musl systems?

Lance0312:
I've created a POC here https://github.com/Lance0312/musl-pthread-chroot-setxid

Lance0312:
The main idea is, `setuid()` under threading environment requires `/proc/self/task` to exist. And `triggerLoadOfLibraries()` makes the environment a threading environment.

Habbie:
As far as I understand, we are triggering an issue specific to musl. Can you turn the ifdef around?

Lance0312:
That would be unfeasible, since there seems to be nothing like `__GLIBC__` in musl, and the musl team refuses to add something like that, here's their explanation http://openwall.com/lists/musl/2013/03/29/13. On the other hand, #1907 is related to glibc directly, so I think this change is quite reasonable.

Habbie:
Then please find another way to detect Musl. We are not going to add `#ifdef __GLIBC__` to work around a Musl-specific issue.

Lance0312:
There seems no elegant way to detect musl libc for now. I would like to leave the problem as-is until there's a proper way to detect musl or there's other proper method to deal with the problem.

Habbie:
Thank you - we agree

PR #3677, opened 2016-04-07
dnsdist: Add support for systemd-notify

pieterlexis:
We now also install our dnsdist.service file from make install.

Needs review

### DIFF
--- a/pdns/dnsdist.cc
+++ b/pdns/dnsdist.cc
@@ -42,6 +42,10 @@
 #include <getopt.h>
 #include "dnsdist-cache.hh"
 
+#ifdef HAVE_SYSTEMD
+#include <systemd/sd-daemon.h>
+#endif
+
 /* Known sins:
 
    Receiver is currently single threaded
@@ -1475,6 +1479,7 @@ try
       break;
     }
   }
+
   argc-=optind;
   argv+=optind;
   for(auto p = argv; *p; ++p) {
@@ -1724,6 +1729,9 @@ try
   thread healththread(healthChecksThread);
 
   if(g_cmdLine.beDaemon || g_cmdLine.beSupervised) {
+#ifdef HAVE_SYSTEMD
+    sd_notify(0, "READY=1");
+#endif
     healththread.join();
   }
   else {
--- a/pdns/dnsdistdist/m4/systemd.m4
+++ b/pdns/dnsdistdist/m4/systemd.m4
@@ -1,0 +1,123 @@
+# systemd.m4 - Macros to check for and enable systemd          -*- Autoconf -*-
+#
+# Copyright (C) 2014 Luis R. Rodriguez <mcgrof@suse.com>
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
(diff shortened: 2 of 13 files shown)


rgacogne:
Nice! Perhaps we could also fallback to 'simple' if systemd support is not available, like weakforced does since [1]?

[1]: https://github.com/PowerDNS/weakforced/pull/47/commits/d0bae52e114a5bf400796593cec97feed1bc813f

pieterlexis:
After pondering this, I feel there is no reason to generate a service file at all if there are no systemd(-dev) libraries installed, as the program is not intended to be run inside of systemd. We can stick a Type=simple service file in the contrib dir and have some documentation about this.

pdns/dnsdist.cc
```diff
@@ -1475,6 +1479,14 @@ try
       break;
     }
   }
+
+#ifdef HAVE_SYSTEMD
+  char *sock;
+  sock = getenv("NOTIFY_SOCKET");
+  if(sock != NULL) // Are we indeed running inside of systemd?
+    g_cmdLine.beSupervised=true;
+#endif
```
Habbie: This can be overridden from the commandline still, I presume?

pdns/dnsdistdist/m4/systemd.m4
```diff
@@ -0,0 +1,148 @@
...
+])
+
+AC_DEFUN([AX_ALLOW_SYSTEMD_OPTS], [
+	AX_ARG_DEFAULT_DISABLE([systemd], [Enable systemd support])
+	AX_SYSTEMD_OPTIONS()
+])
+
+AC_DEFUN([AX_CHECK_SYSTEMD_LIBS], [
+	AC_CHECK_HEADER([systemd/sd-daemon.h], [
+	    AC_CHECK_LIB([systemd-daemon], [sd_listen_fds], [libsystemddaemon="y"])
```
rgacogne: This seems to fail on Arch because libsystemd-daemon.so has been merged into libsystemd.so a while ago (looks like it was in version 209).

pdns/dnsdistdist/m4/systemd.m4
```diff
@@ -0,0 +1,148 @@
+# systemd.m4 - Macros to check for and enable systemd          -*- Autoconf -*-
+#
+# Copyright (C) 2014 Luis R. Rodriguez <mcgrof@suse.com>
+#
```
Habbie: Can we note the origin (including revision) of this file (or whatever we replace it with to fix @rgacogne's report)?

pdns/dnsdist.cc
pieterlexis: not if your actually running inside systemd. There is no other way (as daemonizing will break systemd's detection of a startup and not being supervised will spawn a console with a closed stdin, leading to program termination) to do this. We could just `exit(1)` here and print a message asking the user to start with `--supervised`

pdns/dnsdistdist/m4/systemd.m4
pieterlexis: Even centos 7's pkg-config mentions it as 'deprecated'. Debian jessie has systemd 215, so I'll just change the `systemd-daemon` to `systemd`

pieterlexis:
I fixed up this PR so our packages use this and tested the resulting packages

PR #3685, opened 2016-04-09
Fixed return to master problem

oridistor:
We found out that when notifying from a master, the pdns chooses his return address either randomly or by the first master defined (depending on the bind), this is problematic in case you can't promise to update the masters list every time a master fails (like in our case).

For that, we have fixed the slave communicator code to always return to the master who notified him in that case.

This was compiled and tested.

### DIFF
--- a/debian-pdns/pdns.postinst
+++ b/debian-pdns/pdns.postinst
@@ -10,7 +10,7 @@ update_init() {
 
 
 update_pdns() {
-	if [ -n "$2" ] ; then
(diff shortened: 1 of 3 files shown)


Habbie:
Hi,

there appear to be a lot of unrelated commits in this pull. Can you clean it up please?

oridistor:
Hi, can you please reopen this issue? I'm sending in a pull that has only the needed fix.

PR #3700, opened 2016-04-13
dnsdist: create user from the RPM package to drop privs

pieterlexis:
And add a paragraph about this in the readme.

### DIFF
--- a/build-scripts/build-dnsdist-rpm
+++ b/build-scripts/build-dnsdist-rpm
@@ -40,8 +40,7 @@ SODIUM_CONFIGURE=''
 
 # Some RPM platforms use systemd, others sysv, we default to systemd here
 INIT_BUILDREQUIRES='BuildRequires: systemd-devel'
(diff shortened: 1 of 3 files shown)


build-scripts/build-dnsdist-rpm
```diff
@@ -115,6 +114,41 @@ install -d %{buildroot}/%{_sysconfdir}/dnsdist
...
+%pre
+getent group dnsdist >/dev/null || groupadd -r dnsdist
+getent passwd dnsdist >/dev/null || \
+	useradd -r -g dnsdist -d / -s /sbin/nologin \
+	-c "dnsdist user" dnsdist
+exit 0
+
+
+%post
+%if 0%{?el6}
```
rubenk: No biggie, I guess this will work too, but usually we use  ```is %if 0%{?rhel} == 6```

rubenk:
Please add Requires(pre): shadow-utils since you're using useradd

build-scripts/build-dnsdist-rpm
pieterlexis: I took this from the existing EPEL specfile, but can change this.

rubenk: Nah, it's fine.

PR #3720, opened 2016-04-16
Fixes for dnswasher

edmonds:
This fixes a number of issues I found while trying to use `dnswasher` on a `DLT_RAW` pcap containing IPv6 traffic.

### DIFF
--- a/pdns/dnspcap.cc
+++ b/pdns/dnspcap.cc
@@ -88,7 +88,7 @@ try
       if(d_ip->ip_v==4)
 	contentCode = 0x0800;
       else
-	contentCode = 0x0806; 
+	contentCode = 0x86dd;
     }
     else if(d_pfh.linktype==113)
       contentCode=ntohs(d_lcc->lcc_protocol);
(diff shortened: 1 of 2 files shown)


ahupowerdns:
Robert, as a followup, I think the offset on IPv6 meant we were attempting to zero out the *last* 64 bits of the IP address. Thoughts?

edmonds:
Yeah, that didn't make much sense to me because the first 64 bits of the IPv6 address is still quite a lot of information. E.g., the first 64 bits of my IPv6 address is enough to uniquely identify me :-)

Also, due to the bug fixed in a8b9d93, since it would appear the IPv4 header checksum field physically overlaps the first 64 bits of the IPv6 source address field, I think you were also punching a 16 bit hole in the first 64 bits of the IPv6 source address, at least for queries.

In my local branch I wanted to obfuscate the full 128 bits so I ended up calling the obfuscator twice for each 64 bit half of the IPv6 address.

ahupowerdns:
7856684d63476a5d2dc0f4216be956c6f3d98019 has the proper fix, will be merged shortly.

PR #3722, opened 2016-04-16
dnsdist: use LT_INIT and disable static objects

rubenk:
Do the same thing pdns does.


rgacogne:
:+1: but it needs to be rebased

rubenk:
Done.

PR #3732, opened 2016-04-18
improve OSX portability

Habbie:
~~DO NOT MERGE THIS. If you merge this you will make every active dnsdist user very sad.~~


pdns/dnsdist-protobuf.cc
```diff
@@ -33,7 +34,7 @@ static void protobufFillMessage(PBDNSMessage& message, const DNSQuestion& dq)
   message.set_inbytes(dq.len);
 
   struct timespec ts;
-  clock_gettime(CLOCK_REALTIME, &ts);
+  gettime(&ts);
```
rgacogne: That's probably the only place were using CLOCK_MONOTONIC or CLOCK_MONOTONIC_RAW would be an issue. Perhaps we could add a `needRealTime` parameter to `gettime()`?

Habbie: Yes, one of the reasons this is not ready for merge, I'm flattening the different timescales. A bool sounds good; I was thinking about an enum (mirroring the original `clock_gettime` definition) but a bool is probably fine.

Habbie:
Updated. Systems with `clock_gettime` should see no change in behaviour, while systems without will get 'real' time even when asking for monotonic. Other than implementing more OS-specific APIs, this is the best we can do.

Review welcome of the time stuff (commits are messy but the full diff is quite short, will squash later). Remaining issue is the stuff in `cripple dnsdist lua to make it build on osx`, this is next on my list.

Habbie:
This is ready for review and merge.

pdns/dnsdistdist/configure.ac
```diff
@@ -56,6 +54,8 @@ AS_IF([test "x$enable_hardening" != "xno"], [
   AC_LD_RELRO
 ])
 
+PDNS_CHECK_OS
+
 PDNS_ENABLE_SANITIZERS
 
 PDNS_CHECK_PANDOC
```
rgacogne: No PDNS_CHECK_CLOCK_GETTIME? And it looks like pdns_check_clock_gettime.m4 is not symlinked to in dnsdist's m4 folder.

Habbie: `PDNS_CHECK_CLOCK_GETTIME` was in there already (line 17) - because on some systems it needs `-lrt` I think, and the symlink is already there because of it.

rgacogne: I don't know why I was convinced this file had just been added, my bad.

rgacogne:
LGTM!

Habbie:
Alright, let's wait for travis and hit it.

Issue #3555, opened 2016-03-11
Dynamic signing of synthesized ALIAS records


nickmarden:
(Copied over from https://github.com/PowerDNS/pdns/issues/2394, which is closed.)

@pieterlexis, I tried out the 4.0.0-alpha2 branch and ALIAS works mostly as I expected. However, it seems that A or AAAA records synthesized by the ALIAS mechanism are not signed by the available keys in a DNSSEC-enabled zone.

Your comment at the beginning of the https://github.com/PowerDNS/pdns/issues/2394 thread:

> We're considering releasing a version 3.5 that has this feature (with the warning that it doesn't do DNSSEC correctly on zones with ALIAS). But it could be we push the release of ALIAS into version 4.0.

seems somewhat ambiguous whether (a) the lack of DNSSEC support for ALIAS was expected to be a problem only if ALIAS made it quickly into 3.5, or (b) it was never expected that ALIAS + DNSSEC would work together, even in 4.x.

Unfortunately the absence of dynamic signing on ALIAS records is a deal-breaker for me on using PowerDNS for a cool, high-profile project. So that's why I'm asking.

I can imagine a lot of arguments for and against dynamic signing of the synthesized records in various cases, but since the ALIAS record _appears_ to be an A (or AAAA) record from the point of view of the DNS client - meaning that I am choosing to present the synthesized record to other users as a Very True Fact - shouldn't I (as the authoritative nameserver operator) at least have the option to sign it before I serve it?

Or am I misunderstanding something fundamental here?

Habbie:
Hi Nick,

thank you for your extensive question. The honest answer is that simply no work has been done on making it work with DNSSEC yet. We want this to just work (but we haven't looked at how hard it would be). We do not worry about the 'authority' argument :)

That said, we'd love to know about your high-profile project. Want to drop me a line at peter.van.dijk@powerdns.com to discuss? We'd love to help out, as explained in http://blog.powerdns.com/2014/05/05/how-to-talk-to-an-open-source-software-project-as-a-large-scale-or-otherwise-interesting-user/

Peter

Habbie:
Our 'work in progress' docs on ALIAS also mention we need to sign: https://github.com/PowerDNS/pdns/blob/master/docs/WIP/alias.md#dnssec-processing

nickmarden:
@Habbie - I'll drop you a private email to discuss our case.

[...]

PR #3733, opened 2016-04-18
improve ALIAS handling

Habbie:
This branch expands ALIAS to the backend A/AAAA on AXFR. It also adds some testing (but the DNSSEC tests are broken because of it).

* [x] make ALIAS expansion in AXFR optional (yes/no/reject, default to reject I think) (implemented without reject)
* [x] make AXFR abort when the lookups yield anything other than a -very- clear result (either 'data' or 'very clear NXDOMAIN/NODATA')
* [ ] investigate report of wildcard ALIAS leading to three identical A records in expansion
* [ ] finish #3808
* [ ] #3831 was caused by us punting a query to nproxy even if the client was TCP. Need to fix this.
* [ ] move ALIAS testing records from example.com to a smaller zone that we -do- check in `verify-dnssec-zone`


nickmarden:
Thanks @Habbie! This is going to be very useful for a lot of people :-)

Can you explain the semantics of `make ALIAS expansion in AXFR optional`? I think you mean:
- `yes`: Expand ALIAS records into the underlying A/AAAA values, failing the AXFR whenever the underlying A/AAAA records cannot be resolved
- `no`: Send ALIAS records as ALIAS records
- `reject`: Refuse to AXFR zones that contain ALIAS records

Is that correct?

Habbie:
Hi Nick, yes, that is my thinking for the three options! Note that for 'reject' the refusal may come halfway and thus in the behavioural sense that will be like 'failing the AXFR' on the `yes` side of things. Client are expected to deal with this by not replacing their local copy of the zone at all.

Habbie:
@nickmarden I have added the flag but it's a bool for now, missing the `refuse` option. Does that seem fine to you?

nickmarden:
Sounds great to me. I'd definitely be using the `yes` value, not the `refuse` value :-)

nickmarden:
@Habbie, I see that [a related issue](https://github.com/PowerDNS/pdns/issues/3555) is tagged auth-4.0.0, but I was wondering if there is an ETA for this PR to be on PowerDNS master?

Habbie:
@nickmarden as I may have mentioned to you (or not), the `getaddrinfo`-based infrastructure in this PR is unreliable, it does not tell me the difference between 'everything is terrible' and 'name does not exist'. To fix this, I have done some refactoring in #3802 so the ALIAS code can also use a reliably resolver. This was merged a few days ago, and I need to update this PR to use that code. After that (should be sometime next week) I feel we can merge this as it will make everything just a lot better. There are some nits on #3808, and for ALIAS purposes the 'this seems optimistic' bullet (currently the 4th one) in #3808 might be relevant, but that's incremental work.

TLDR I hope to finish and merge this PR sometime next week :)

nickmarden:
OK thanks @Habbie - I now understand the dependencies and timeline better. Godspeed!

Habbie:
Expecting travis to go green any second. Reviews welcome. Note that not all bullets at the top have been fixed - this is fine, they will go into a new ticket. As far as I can tell this PR is a strong improvement over the existing situation and more polish will be applied after user testing.

nickmarden:
@Habbie I'm not sufficiently familiar with the PowerDNS internals to offer useful commentary on most of the code changes, but the new option `outgoing-axfr-expand-alias` is exactly as we've discussed and so I give that a hearty 👍

Issue #3682, opened 2016-04-08
DNSSEC validation with +AD in the query


pieterlexis:
[RFC 6840 § 5.7](https://tools.ietf.org/html/rfc6840#section-5.7) specifies that a client sending a query with the AD bit set "understands and is interested in the value of the AD bit in the response". In discussion with @Habbie, this means to us 'please do DNSSEC validation'. As such, with `dnssec` set to anything but `off`, we should validate the records for these clients and send the appropriate response. 

This image explains the expected responses of the recursor based on flags, `dnssec` setting and DNSSEC status.
![This image](http://i.imgur.com/GuIwX2S.png)

In the same discussion, it was concluded that the DO bit on a query means 'give me DNSSEC records'. This means that responses will only be validated based on the `dnssec` setting and the CD bit.

HLFH:
I discovered this issue with alpha1 [at the end of December 2015](https://blog.powerdns.com/2015/12/24/technical-preview-releases-of-authoritative-server-recursor-and-dnsdist/) but I took time to understand it, and I forgot to submit this one. For this moment, I'm using PowerDNS 4.0 Alpha 2 as an authoritative DNS server and, on another server, Unbound 1.5.8 as a DNS resolver because Unbound does not miss the AD bit during DNSSEC validation.

PR #3741, opened 2016-04-20
recursor: Add DNSSEC testing infrastructure and tests

pieterlexis:
So we can actually test DNSSEC in the recursor.

This PR contains a few simple tests and the whole test-suite to test #3682 (some of the breaking tests are skipped).

Comments are welcome.

PR should be squashed when merging


regression-tests.recursor-dnssec/test_Flags.py
```diff
@@ -0,0 +1,679 @@
...
+        msg.flags = dns.flags.from_text(flags)
+        msg.flags += dns.flags.from_text('RD')
+        msg.use_edns(edns=0, ednsflags=dns.flags.edns_from_text(ednsflags))
+        return msg
+
+    def getQueryForSecure(self, flags='', ednsflags=''):
+        return self.createQuery('ns1.example.', 'A', flags, ednsflags)
+
+    ##
+    #   -AD -CD -DO
```
Habbie: I wonder if this file, starting here, couldn't look more like a table :)

regression-tests.recursor-dnssec/test_Flags.py
```diff
@@ -0,0 +1,679 @@
...
+
+    def testValidate_Insecure_None(self):
+        msg = self.getQueryForInsecure()
+        res = self.sendUDPQuery(msg, 'validate')
+        self.assertRcodeEqual(res, dns.rcode.NOERROR)
+        self.assertMessageHasFlags(res, ['QR', 'RA', 'RD'])
+        self.assertNoRRSIGsInAnswer(res)
+
+    ##
+    # +AD -CD -DO
```
Habbie: we have this comment header three times in the file, why not group them?

Habbie:
That's a bunch of nice work! Nits posted. Did not review every test yet, will wait for your replies.

regression-tests.recursor-dnssec/test_Flags.py
pieterlexis: The tests are grouped by domain state (secure, bogus, insecure) and then we test every combination of flags against 3 `dnssec=` settings. So there are multiple groupings :smile:

regression-tests.recursor-dnssec/test_Flags.py
pieterlexis: This could be awesome, would require a multidimensional list (dnssec setting, domain status, flags (possibly split by flag on/off))

PR #3752, opened 2016-04-22
recursor: DNSSEC related query flag processing

### LINKED ISSUE #3682: DNSSEC validation with +AD in the query

pieterlexis:
Please merge after #3741.

This PR ensures that we only do validation when asked (by means of the AD bit in the query) or in full-blown validation mode. The recursor now only returns DNSSEC records (RRSIG, NSEC, NSEC3) when the DO bit is set in the query (and won't validate on +DO unless in validate mode).

Furthermore, it solves a bug where a +CD in a query to a recursor with `dnssec=off` would lead to a response with the DO bit set.

Closes #3682

### DIFF
--- a/pdns/pdns_recursor.cc
+++ b/pdns/pdns_recursor.cc
@@ -736,12 +736,17 @@ void startDoResolve(void *p)
       sr.d_requestor=dc->d_remote;
     }
 
-    if(g_dnssecmode != DNSSECMode::Off)
+    if(g_dnssecmode != DNSSECMode::Off) {
       sr.d_doDNSSEC=true;
-    
-    if(pw.getHeader()->cd || (edo.d_Z & EDNSOpts::DNSSECOK)) {
-      DNSSECOK=true;
-      g_stats.dnssecQueries++;
+
+      // Does the requestor want DNSSEC records?
+      if(edo.d_Z & EDNSOpts::DNSSECOK) {
+        DNSSECOK=true;
+        g_stats.dnssecQueries++;
+      }
+    } else {
+      // Ignore the client-set CD flag
+      pw.getHeader()->cd=0;
     }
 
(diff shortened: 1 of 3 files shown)


pdns/dnsrulactions.hh
```diff
@@ -126,7 +126,7 @@ public:
   }
   bool matches(const DNSQuestion* dq) const override
   {
-    return dq->dh->cd || (getEDNSZ((const char*)dq->dh, dq->len) & EDNS_HEADER_FLAG_DO);    // turns out dig sets ad by default..
+    return dq->dh->ad || (getEDNSZ((const char*)dq->dh, dq->len) & EDNS_HEADER_FLAG_DO);
```
rgacogne: Hmm, this looks wrong, or at least would not be consistent with our documented behavior.

pieterlexis: ohw, this file is not part of the recursor. Based on the documentation, we should make this line
```
return dq->dh->cd || dq->dh->ad ||(getEDNSZ((const char*)dq->dh, dq->len) & EDNS_HEADER_FLAG_DO);
```

As any of these flags could mean "do something with DNSSEC"

PR #3762, opened 2016-04-25
Pls discuss: Teeaction: send copy of query to second nameserver, sponge responses

ahupowerdns:
This adds a getStats() too all Actions which is currently empty, but could be exposed via the API so people could query it remotely. I think it makes sense like this but @rgacogne is closed in tune with the architecture of dnsdist post 1.0.0 I think. So please ponder before merging!

### DIFF
--- a/pdns/dnsdistdist/dnsrulactions.cc
+++ b/pdns/dnsdistdist/dnsrulactions.cc
@@ -66,7 +66,7 @@ void TeeAction::worker()
     if(res==0)
       continue;
     res=recv(d_fd, packet, sizeof(packet), 0);
-    if(res <= 0) 
+    if(res <= (int)sizeof(struct dnsheader)) 
       d_recverrors++;
     else if(res > 0)
       d_responses++;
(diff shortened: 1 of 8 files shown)


pdns/dnsdistdist/dnsrulactions.cc
```diff
@@ -0,0 +1,87 @@
...
+  close(d_fd);
+  d_worker.join();
+}
+
+DNSAction::Action TeeAction::operator()(DNSQuestion* dq, string* ruleresult) const 
+{
+  if(dq->tcp) 
+    d_tcpdrops++;
+  else {
+    d_queries++;
```
rgacogne: Given that this may be called from different threads, perhaps we might want to make `d_tcpdrops` and `d_queries` atomic

pdns/dnsdistdist/dnsrulactions.cc
```diff
@@ -0,0 +1,87 @@
...
+    if(d_pleaseQuit)
+      break;
+    if(res < 0) {
+      usleep(250000);
+      continue;
+    }
+    if(res==0)
+      continue;
+    res=recv(d_fd, packet, sizeof(packet), 0);
+    if(res <= 0)
```
rgacogne: < sizeof(struct dnsheader), otherwise we might be reading garbage below

PR #3764, opened 2016-04-25
Add systemd notify support to auth and recursor

pieterlexis:


pieterlexis:
@rubenk, @rgacogne comments very welcome.

pdns/pdns_recursor.cc
```diff
@@ -2651,6 +2655,9 @@ int serviceMain(int argc, char*argv[])
   if(g_numThreads == 1) {
     L<<Logger::Warning<<"Operating unthreaded"<<endl;
     recursorThread(0);
+#ifdef HAVE_SYSTEMD
+    sd_notify(0, "READY=1");
```
rgacogne: Shouldn't the call be before calling `recursorThread()`?

rgacogne:
except for the previous comment, LGTM

pieterlexis:
The problem is we still can exit with a config error (webserver) in recursorThread, although.. But when you're running unthreaded you're (hopefully) not doing production.

rgacogne:
Not sure if relevant, but it might be a good idea to prevent the services files from being installed on trusty, as done in #3837 ?

pieterlexis:
I was planning to fix that after #3837 was merged

PR #3802, opened 2016-05-02
refactor doResolve out of secpoll

Habbie:
This is in preparation for switching #3733 from `FindNS` to `(stub)DoResolve`

### DIFF
--- a/pdns/stubresolver.cc
+++ b/pdns/stubresolver.cc
@@ -1,0 +1,121 @@
+#ifdef HAVE_CONFIG_H
+#include "config.h"
+#endif
+
+#include "logger.hh"
+#include "arguments.hh"
+#include "version.hh"
+#include "misc.hh"
+
+#include "sstuff.hh"
+#include "dnswriter.hh"
+#include "dns_random.hh"
+#include "namespaces.hh"
+#include "statbag.hh"
+#include "stubresolver.hh"
+
+// s_secpollresolvers contains the ComboAddresses that are used to resolve the
+// secpoll status of PowerDNS
+static vector<ComboAddress> s_secpollresolvers;
+
+/** Parse /etc/resolv.conf and add the nameservers to the vector
+ * s_secpollresolvers.
+ */
+void secPollParseResolveConf()
+{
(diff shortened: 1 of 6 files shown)


zeha:
Once this is used in #3733, some form of reloading for `s_stubresolvers` would be useful.

Habbie:
@zeha good point, will write that down somewhere after merge.

Ready for review/merge as far as I'm concerned.

pdns/stubresolver.cc
```diff
@@ -0,0 +1,120 @@
...
+        struct dnsheader d;
+        memcpy(&d, reply.c_str(), sizeof(d));
+        if(d.id != pw.getHeader()->id)
+          goto retry;
+      }
+    }
+    catch(...) {
+      continue;
+    }
+    MOADNSParser mdp(reply);
```
rgacogne: We could perhaps catch parsing exceptions here, to be able to retry with another resolver if the response contains garbage?

Habbie: Yes, makes sense, right now a parsing error would just throw immediately. I have plans to improve this code as I start using it for ALIAS by the way - I'm happy to merge this one if we feel I haven't at least broken anything for secpoll compared to before the PR :)

pdns/stubresolver.cc
```diff
@@ -0,0 +1,120 @@
...
+
+    waitForData(sock.getHandle(), 2, 0);
+    try {
+    retry:
+      sock.recvFrom(reply, dest);
+      if(reply.size() > sizeof(struct dnsheader)) {
+        struct dnsheader d;
+        memcpy(&d, reply.c_str(), sizeof(d));
+        if(d.id != pw.getHeader()->id)
+          goto retry;
```
rgacogne: We might want to limit the number of retries, otherwise it looks like we might loop as long as we keep receiving non-matching responses?

Habbie: Makes sense.

rgacogne:
AFAICT this is at least as good as it was before the PR, LGTM :)

Habbie:
Thanks! 3 nits noted in #3808. Merging now.

PR #3804, opened 2016-05-02
rec: Fix a memory leak in DNSSEC validation

rgacogne:
`DNSCryptoKeyEngine::makeFromPublicKeyString()` returns a naked
pointer to a new object.


ahupowerdns:
(small tip, not too relevant here, but scoped_ptr or unique_ptr would save a few cycles potentially)

PR #3806, opened 2016-05-03
auth: Move key validity check out of `fromISCMap()`

### LINKED ISSUE #3409: Checking keys in OpenSSL is expensive
As discused with @Habbie @rgacogne and @cmouse

The current OpenSSL signer does increase the CPU usage 5-10 times. After some investigation it turns out RSA_check_key() is the one to blame. Without this check CPU is back at normal (compared to mbed) levels.

Problems:
- RSA_check_key() and most likely EC_KEY_check_key() are expensive
- we fromiscmap a LOT (every query?)

Questions:
- Is the check necessary (in fromiscmap)
- Can we reduce the number of  fromiscmap invocations (for example, only if there is no signature cached)  Does not solve the CPU increase at startup or Thursdays
- Move the check to check-zone or 'check-keys' ?

rgacogne:
As reported by @mind04, it doesn't make a lot of sense to check the key validity at every
call of `fromISCMap()`, and it hurts performance a lot when keys
are not cached.

* Add separate `DNSSECKeeper::checkKeys()` and
`DNSCryptoKeyEngine::checkKeys()` methods
* Key validity is checked on import-zone-key, check-zone and
test-algorithm(s)

Closes #3409.

### DIFF
--- a/pdns/opensslsigners.cc
+++ b/pdns/opensslsigners.cc
@@ -99,6 +99,7 @@ public:
   std::string getPublicKeyString() const;
   void fromISCMap(DNSKEYRecordContent& drc, std::map<std::string, std::string>& stormap);
   void fromPublicKeyString(const std::string& content);
+  bool checkKeys() const override;
 
   static DNSCryptoKeyEngine* maker(unsigned int algorithm)
   {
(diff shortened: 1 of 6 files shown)


zeha:
Logic checks out. I'd think the operations in `pdnsutil hsm` that call `makeFromISCString` would also benefit from calling `checkKey`.

pdns/opensslsigners.cc
```diff
@@ -353,18 +354,17 @@ void OpenSSLRSADNSCryptoKeyEngine::fromISCMap(DNSKEYRecordContent& drc, std::map
...
-    throw runtime_error(getName()+" invalid public key");
-  }
-
   if (d_key)
     RSA_free(d_key);
 
   d_key = key;
 }
 
+bool OpenSSLRSADNSCryptoKeyEngine::checkKeys() const
```
Habbie: shouldn't this be `checkKey`?

rgacogne: Could be, it checks a key-pair but we mostly refer to it as key, so you are probably right, I'll change that along with the hsm stuff!

Habbie:
LGTM, if you are indeed done, go ahead and merge :)

PR #3807, opened 2016-05-03
auth: Add TTL settings for DNSSECKeeper's caches (key, medatada)

rgacogne:


pdns/pdnsutil.cc
```diff
@@ -120,6 +120,8 @@ void loadMainConfig(const std::string& configdir)
   ::arg().set("soa-expire-default","Default SOA expire")="604800";
   ::arg().set("soa-minimum-ttl","Default SOA minimum ttl")="3600";    
   ::arg().set("chroot","Switch to this chroot jail")="";
+  ::arg().set("dns-key-cache-ttl","Seconds to store domain DNS keys in cache")="30";
```
zeha: that could use a better description... without reading the code change I was unable to understand what this affects.

Habbie:
LGTM other than zeha's nit. How about "Seconds to cache DNSSEC keys from the database" ?

mind04:
And maybe ponder a bit what to do with 0 second ttls. Completely  disable the cache like we do for the packet and query cache? Right now a ttl of 0 result in sub second caching for the remaining part of the current second.

Habbie:
@mind04 agreed

rgacogne:
Pushed a new commit:
* Fix the description of the new settings
* Setting a 0-TTL disables caching
* Only get the value once, as it's done for `max-nsec3-iterations`

PR #3810, opened 2016-05-04
Remove pdns.conf-dist (generate it after compiling)

### LINKED ISSUE #1905: Remove pdns.conf-dist
Use the output of pdns_server --config instead.

It would be nice to override settings in the output, like pdns_server --config --setuid=powerdns for example.

pieterlexis:
Closes #1905. Custom options are not allowed/possible. Even if one would `--config --some-setting=foo` while generating the config, the setting would still be commented. Passing `true` to confgstring() will make pdns_server uncomment *all* settings, which is less desirable than using awk/sed during the packaging.


Habbie:
LGTM. Failing travis test restarted (was unrelated).

Habbie:
But, branch has conflicts. I bet they're in pdns.conf-dist ;)

PR #3817, opened 2016-05-04
Prevent launching similarly named backend

### LINKED ISSUE #3615: Its possible to load a psql server ( the same ) two times, white ugly side effects
When accidently doing

pdns.conf
`launch=gpgsql`

and then in pdns.d/mybacken.conf
`launch+=gpgsql`

things go ham. You will have every zone twice when using `pdns_control list-zones` but also the API gives back dupes and also, writing into the database using the API results in "dupe key" errors ( 2 write requests i guess).

It is clear that i made a configuration error.. the first definition was not intentionally and was not seen by me due to this huge "example configurations" with all the comments.

I guess, just throw an error or warning, that this is not designed to be, or the second gpgsql connection needs a specific name ( not default )

pieterlexis:
This catches configuration bugs (mostly when includes are used) where a
line `launch=backend` and a line `launch+=backend` occur in the sum of
the configuration.

Closes #3615

### DIFF
--- a/pdns/dnsbackend.cc
+++ b/pdns/dnsbackend.cc
@@ -134,6 +134,10 @@ void BackendMakerClass::launch(const string &instr)
   vector<string> parts;
   stringtok(parts,instr,", ");
 
+  for (const auto part : parts)
+    if (count(parts.begin(), parts.end(), part) > 1)
+      throw ArgException("Refusing to launch multiple backends with the same name '" + part + "', verify all 'launch' statements in your configuration");
+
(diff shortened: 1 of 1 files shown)


Habbie:
1. why are you sorting? doesn't that affect launch order?
2. did you mean 'prevent launching identical backends'?

pieterlexis:
1. oh, a leftover from an earlier implementation, will fix
2. well, yes... in our case this is the same, I'll edit the message

Habbie:
LGTM

PR #3827, opened 2016-05-06
set NSEC3PARAM algorithm to 1

### LINKED ISSUE #3796: wrong digest type in NSEC3PARAM breaks things
Inspired by web.de's digest type '8': it turns out that while we turn this into a 1 on AXFR (including matching signatures) we do send out the 8 on direct queries. In both cases we appear to be generating correct signatures, but validators (at least unbound) choke on the 8, and they are right. We should refuse anything but 1 in set-nsec3. In pdns_server, we should either robustly map to 1, or consistently SERVFAIL. Leaning towards just forcing '1'.

pieterlexis:
closes #3796

### DIFF
--- a/pdns/pdnsutil.cc
+++ b/pdns/pdnsutil.cc
@@ -2424,6 +2424,10 @@ loadMainConfig(g_vm["config-dir"].as<string>());
       cerr<<"Cannot enable NSEC3 for " << zone.toString() << " as it is too long (" << zone.wirelength() << " bytes, maximum is 222 bytes)"<<endl;
       return 1;
     }
+    if(ns3pr.d_algorithm != 1) {
+      cerr<<"NSEC3PARAM algorithm set to '"<<std::to_string(ns3pr.d_algorithm)<<"', but '1' is the only valid value"<<endl;
+      return EXIT_FAILURE;
+    }
     if (! dk.setNSEC3PARAM(zone, ns3pr, narrow)) {
       cerr<<"Cannot set NSEC3 param for " << zone.toString() << endl;
       return 1;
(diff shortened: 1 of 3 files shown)


Habbie:
Unsure about the logging in packethandler, might become a bit verbose. Otherwise LGTM. If you are fine with the verbosity, merge away.

pieterlexis:
This logging should never happen if the database contents are OK, I'm fine with the verbosity when a user messes up

mind04:
Why is the check not located in getNSEC3PARAM() / setNSEC3PARAm (like the max-nsec3-iterations check).
Right now it is so easy to miss things... Like rfc2136 updates and AXFR in.

pdns/pdnsutil.cc
```diff
@@ -408,9 +408,15 @@ int checkZone(DNSSECKeeper &dk, UeberBackend &B, const DNSName& zone, const vect
...
+  if (haveNSEC3){
+    if(isSecure && zone.wirelength() > 222) {
+      numerrors++;
+      cout<<"[Error] zone '" << zone.toStringNoDot() << "' has NSEC3 semantics but is too long to have the hash prepended. Zone name is " << zone.wirelength() << " bytes long, whereas the maximum is 222 bytes." << endl;
+    }
+    if(ns3pr.d_algorithm != 1) {
+      numerrors++;
+      cout<<"[Error] NSEC3PARAM algorithm for zone '"<<zone.toStringRootDot() << "' is '"<< std::to_string(ns3pr.d_algorithm) <<"', which is not a valid value."<<endl;
+    }
   }
```
mind04: this check is never triggered if you force  d_algorithm to 1 in getNSEC3PARAM()

mind04:
There is also a getNSEC3PARAM() in bind backend https://github.com/powerdns/pdns/blob/master/modules/bindbackend/binddnssec.cc#L153

PR #3833, opened 2016-05-09
Minor fixes to make coverity happier

rgacogne:
* Return false if `setsockopt()` failed in `setTCPNoDelay()`
* `writen2()` returns a size_t or throws an exception
* dnsdist: In verbose mode, warn if `sendto()`/`sendfromto()` failed

### DIFF
--- a/pdns/dnsdist.cc
+++ b/pdns/dnsdist.cc
@@ -144,10 +144,17 @@ struct DelayedPacket
   ComboAddress origDest;
   void operator()()
   {
-    if(origDest.sin4.sin_family == 0)
-      sendto(fd, packet.c_str(), packet.size(), 0, (struct sockaddr*)&destination, destination.getSocklen());
-    else
-      sendfromto(fd, packet.c_str(), packet.size(), 0, origDest, destination);
+    ssize_t res;
+    if(origDest.sin4.sin_family == 0) {
+      res = sendto(fd, packet.c_str(), packet.size(), 0, (struct sockaddr*)&destination, destination.getSocklen());
+    }
+    else {
+      res = sendfromto(fd, packet.c_str(), packet.size(), 0, origDest, destination);
+    }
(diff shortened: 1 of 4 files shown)


pdns/dnsdist.cc
```diff
@@ -144,10 +144,17 @@ struct DelayedPacket
   ComboAddress origDest;
   void operator()()
   {
-    if(origDest.sin4.sin_family == 0)
-      sendto(fd, packet.c_str(), packet.size(), 0, (struct sockaddr*)&destination, destination.getSocklen());
-    else
-      sendfromto(fd, packet.c_str(), packet.size(), 0, origDest, destination);
+    ssize_t res;
+    if(origDest.sin4.sin_family == 0) {
```
cmouse: this only adds { } around here which are not even needed?

pdns/dynmessenger.cc
```diff
@@ -119,10 +119,7 @@ DynMessenger::~DynMessenger()
 int DynMessenger::send(const string &msg) const
 {
   try {
-    if(writen2(d_s, msg+"\n") < 0) { // sue me
-      perror("sendto");
-      return -1;
-    }
+    writen2(d_s, msg+"\n");
```
cmouse: now you are ignoring return value?

pieterlexis: from the description: " writen2() returns a size_t or throws an exception"

Habbie: the return value is irrelevant

Habbie:
LGTM

pdns/dnsdist.cc
rgacogne: Well, the main point is to check the return value to be able to issue an informative message a few lines below. The { } are only there because the "if (res == -1)" line looked confusing to me otherwise.

PR #3837, opened 2016-05-10
Remove dnsdist service file on trusty

pieterlexis:
Having it there will lead to pain


Habbie:
Also need to do this for auth and rec.

Habbie:
LGTM

PR #3851, opened 2016-05-11
disable algorithm 13 and 14 in the opensslsigner if OpenSSL does not …

mind04:
…support ecdsa or the required curves

### DIFF
--- a/m4/ax_check_openssl.m4
+++ b/m4/ax_check_openssl.m4
@@ -109,6 +109,18 @@ AC_DEFUN([AX_CHECK_OPENSSL], [
         [AC_LANG_PROGRAM([#include <openssl/crypto.h>], [CRYPTO_free(NULL)])],
         [
             AC_MSG_RESULT([yes])
+            openssl_ecdsa=yes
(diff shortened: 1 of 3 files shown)


m4/ax_check_openssl.m4
```diff
@@ -109,6 +109,18 @@ AC_DEFUN([AX_CHECK_OPENSSL], [
...
+            AC_CHECK_FUNC(ECDSA_do_sign,
+            [
+                AC_CHECK_DECLS([NID_X9_62_prime256v1, NID_secp384r1], [ : ], [ openssl_ecdsa=no ], [AC_INCLUDES_DEFAULT
+#include <openssl/evp.h>
+                ])
+            ], [
+                openssl_ecdsa=no
+            ])
+            AS_IF([test "x$openssl_ecdsa" = "xyes"], [
+                AC_DEFINE([OPENSSL_USE_ECDSA], [1], [define to 1 if OpenSSL ecdsa support is avalable.])
```
pieterlexis: I would prefer `HAVE_OPENSSL_ECDSA` as name of the defined variable.

PR #3858, opened 2016-05-13
dnsdist: Only log YaHTTP's parser exceptions in verbose mode

rgacogne:
Log the remote client address in any case.


Habbie:
LGTM

PR #3859, opened 2016-05-13
build scripts: Stop patching config-dir in pdns.conf

zeha:
By the time this is read, it is too late anyway.


Habbie:
+1 - it's useless on both sides and the RPM one even gets it wrong

PR #3865, opened 2016-05-17
Allow fallback to no EDNS in DNSSEC mode

pieterlexis:
For @ahupowerdns's review pleasure.


ahupowerdns:
Um well - this does make us downgrade permanently on a bunch of timeouts and break all DNSSEC domains on that server. This is the issue BIND had. So this needs more thinking.

PR #3869, opened 2016-05-18
rec: Log outgoing queries / incoming responses via protobuf

rgacogne:
Also add ProtobufLogger.py, a script to display protobuf payloads

### DIFF
--- a/contrib/ProtobufLogger.py
+++ b/contrib/ProtobufLogger.py
@@ -30,11 +30,10 @@ class PDNSPBConnHandler(object):
                 self.printQueryMessage(msg)
             elif msg.type == dnsmessage_pb2.PBDNSMessage.DNSResponseType:
                 self.printResponseMessage(msg)
-            # PR #3869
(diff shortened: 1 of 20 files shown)


Habbie:
Postponed; maybe you can split ProtobufLogger.py into a separate PR?

rgacogne:
Do not merge. This code uses boost::uuid, which requires boost 1.42+, regardless of whether protobuf support is enabled. It doesn't play well with the boost version available in CentOS 6 (1.41).

rgacogne:
Updated, this only pulls boost::uuid in when protobuf is enabled.

Habbie:
This has conflicts. Was it ready for review besides that, @rgacogne?

rgacogne:
Rebased, ready for review!

pieterlexis:
more a question than a review: what is the purpose (eventually) for the ResolveContext?

rgacogne:
Right now its only use is to carry around the `initialRequestId`, ie the UUID of the initial incoming query resulting in the outgoing queries to authoritative servers. I'll welcome any suggestion for passing that in a cleaner way, I might have missed an obvious one!

PR #3872, opened 2016-05-18
Auth: add support for multiple carbon servers

### LINKED ISSUE #2093: multiple carbon-servers
Can we have the ability to have powerdns send data to multiple carbon-servers?

### LINKED ISSUE #3782: Authoritative server misses support for multiple carbon servers
#3485 adds this support to the recursor. Similar support should be added to the authoritative.

pieterlexis:
Could use a review

Closes #3782
Closes #2093

### DIFF
--- a/pdns/auth-carbon.cc
+++ b/pdns/auth-carbon.cc
@@ -15,52 +15,49 @@ try
 {
   extern StatBag S;
 
-  for(int numloops=0;;++numloops) {
-    if(arg()["carbon-server"].empty()) {
+  string hostname=arg()["carbon-ourname"];
+  if(hostname.empty()) {
+    char tmp[80];
+    memset(tmp, 0, sizeof(tmp));
+    gethostname(tmp, sizeof(tmp));
+    char *p = strchr(tmp, '.');
+    if(p) *p=0;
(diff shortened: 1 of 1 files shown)


pdns/auth-carbon.cc
```diff
@@ -15,52 +15,54 @@ try
...
-	char *p = strchr(tmp, '.');
-	if(p) *p=0;
-	hostname=tmp;
-	boost::replace_all(hostname, ".", "_");
-      }
-      for(const string& entry :  entries) {
-	str<<"pdns."<<hostname<<".auth."<<entry<<' '<<S.read(entry)<<' '<<now<<"\r\n";
+
+      if (msg.empty()) {
+        ostringstream str;
```
rgacogne: This could be done before entering the `for (const auto& carbonServer : carbonServers)` loop, it might perhaps be easier to read.

Habbie:
LGTM, but: when we do this we may want to make 'sleep' smarter, to prevent one slow/down target from increasing the interval for the others. Also, this PR drops the theoretical ability to change carbon servers at run time.

Still, LGTM, assuming this was tested.

PR #3874, opened 2016-05-18
Remote backend doc example change (types)

romuald:
Modify the examples for `getDomainMetadata`, since the `"YES"` value does not currently seems to be parsed correctly by the remote backend.
`"1"` is correctly parsed as a boolean true

### DIFF
--- a/docs/markdown/authoritative/backend-remote.md
+++ b/docs/markdown/authoritative/backend-remote.md
@@ -230,7 +230,7 @@ Query:
 
 Response:
 ```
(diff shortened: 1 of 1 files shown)


Habbie:
LGTM. @cmouse?

cmouse:
Ah right. I had forgotten that we actually use 0 and 1 instead of NO and YES. Yes :+1:

Habbie:
ITYM "1" :+1: ;)

Issue #3846, opened 2016-05-11
Feature request: dnsdist AddResponseAction() per server or per pool 


lark:
AddResponseAction() per server or per pool makes configuration simple, when used with AddPoolRule().

Only one (per pool) or some (per server) AddResponseAction() is needed, in case of multiple AddPoolRule("domain", "poolname").

lark:
It seems that this is relative simple: extending DNSQuestion to contain server state, then processResponse() can use server state to get server information or poolname.

rgacogne:
Out of curiosity, what are you trying to achieve? `AddResponseAction()` is very limited for now, and I'm wondering what users would like to be able to do in there.

lark:
A good candidate to replace dnsmasq with ipset functionality. That is, forwarding certain domain queries to certain servers, and then setting ipset set per responses, to interact with other network functionality, such as NAT, routing.

dnsmasq is great but has limitations. There is no failover, no server health check; logging is not flexible, etc.

rgacogne:
Closing as this can be implemented using tags in the current version.

PR #3879, opened 2016-05-20
Add ResponsePoolRule which matches pool name

lark:
This is a POC to add response rule based on which pool this reponse is returned. See #3846 . This pull request is RFC, not for merging.

UDP is tested and works, while TCP is not tested.


rgacogne:
Hi!

Firstly, thank you for this PR. My only worry is that if a server (say server1) belongs to more than one pool (pool1 and pool2), a query is routed to pool1 and then sent to server1, ResponsePoolRule("pool2") would match, which is probably not expected.
How do you feel about storing the PoolServer in the IDState object (for UDP) and then adding it to the DNSQuestion object when handling the response, instead of the DownstreamState?

lark:
A "server" is actually a request channel, and to the same server, multiple request channels can be specified.

One can split those requests to a channel which is in single pool. If this is explicitly and carefully documented, I think it's ok to match pool name. This is simple in implementation, and also simple for configuration if multiple "servers" in one pool.

I didn't read the IDState code thoroughly. I will look into it later. As it's UDP only, how to deal with TCP case?

My PR is for response only, but the concept is useful for request too. If multiple actions should be specified for some request rules, then pool is a good grouping/filtering mechanism which helps to simplify configuration drastically. (Think about it like a tag)

BTW, is there race condition to processResponse() and change server's pool? I am not sure.

rgacogne:
Sure, we could document that `ResponsePoolRule()` should only be used when a server is assigned to one pool at most, but storing the PoolServer would lift that limitation. There is no `IDState` for UDP because TCP handling is currently synchronous, so there wouldn't be any trouble keeping the PoolServer around over TCP.
I like your idea of adding some sort of tags to make configuration easier. However I don't think using the pool for that purpose is a good idea, if only because assigning the query to a pool stops the rules processing. But tags might definitely be a nice improvement.

lark:
@rgacogne 

I am not sure if adding tags is good/easy. How to carry tags across the request/response boundry and back? For now, I think Pool or Server is easier.

Habbie:
@rgacogne open question for you here as I read it

rgacogne:
I completely missed that question, sorry. We could keep the tags inside the `IDState` object, but I agree it would be more complicated that what you are suggesting. I still think it's misleading to match if the backend used is in a given pool, even if the query wasn't routed to that exact pool, but I guess it might be ok if that's clearly documented. Another option would be to store the pool the query was routed to in the `IDState` object.
That's unrelated but response rules now get a `DNSResponse` object instead of the `DNSQuestion` one, so you will need to update your PR. It also lacks at least one line documenting the new rule in the README.

lark:
@rgacogne I tried to rebase this PR to master, but no DNSResponseRule for DNSResponse yet.

rgacogne:
Oh right, I have been mistaken, sorry about that. `DNSResponseAction` get a `DNSResponse`, but there is no distinction between query and response rules. You should be fine just rebasing to master.

lark:
@rgacogne Any suggestion or feedback?

pieterlexis:
@rgacogne: ping?

rgacogne:
I forgot about this PR, sorry about that @lark. I would really like to prevent a possible issue if a server belongs to more than one pool. How about storing a `std::shared_ptr<ServerPool>` in `IDState`, instead of a `std::shared_ptr<DNSDistPacketCache>`? We can get the packet cache from the `ServerPool` in the UDP responder thread, and that way we can know exactly to which pool the query has been routed.

pieterlexis:
ping!

Habbie:
This branch has conflicts that must be resolved.

lark:
The new patch uses IDState for UDP as @rgacogne suggested. Now it's ok to put a server in multiple different pools.

We all admit a more general tag system which can mark a query/response in its lifecycle is the right way to go. But implementing it is beyond my affordable time.

And using pool is not that bad, IMHO. In my use case, I need to let a specific server resolve the query, and then take certain actions on response. If using tag system
```
addAction({"some.domain"}, tagAction("tagname"))
...
addAction(tagRule("tagname"), poolAction("poolname"))
addResponseAction(tagRule("tagname"), someAction())
```
While using pool
```
addAction({"some.domain"}, poolAction("poolname"))
...
addResponseAction(ResponsePoolRule("poolname"), someAction())
```

zeha:
I gave using the ServerPool ptr a try, see https://gist.github.com/zeha/d687634077f977a9af70c4594cb19477

`g_pools.getCopy()` looks like it'd be horribly slow? But maybe then one only pays the price if actually using a `ResponsePoolRule`.

Any other red flags?

rgacogne:
Now that we have a mechanism to add tags to a query (`dq::setTag()`..), I feel like that might be more suited to the task. We need to pass the tags from the query to the response (simply copying the shared pointer from the IDState structure and then to the DNSResponse should do) and add rules to be able to match on tags. It should not take long to write such a patch, I could work on that if there is an interest for it.

lark:
It's great that a tag system has been implemented. I do wish you can take some time to write the patch for DNSResponse.

rgacogne:
I opened #6037 a few days ago, proposing an implementation of tag-based routing. Any comments or suggestions are welcome!

zeha:
Is this still relevant with #6037 already being merged?

PR #3886, opened 2016-05-23
Some Recursor fixes

### LINKED ISSUE #3855: disable-syslog destroys recursor logmessages statistic
Under `disable-syslog` we assume that our stdout is being stored, by systemd for example. In that situation, `S.ringAccount("logmessages",msg);` is no longer executed, which means the stat stays at zero.

pieterlexis:
* Add summary to the recursor's configure output
* Remove the statbag from the recursor
* Fix #3855 as well


Habbie:
LGTM

PR #3889, opened 2016-05-23
backport two bug fixes to 3.4.x

klaus3000:


pdns/dynhandler.cc
```diff
@@ -269,8 +269,8 @@ string DLNotifyHandler(const vector<string>&parts, Utility::pid_t ppid)
   UeberBackend B;
   if(parts.size()!=2)
     return "syntax: notify domain";
-  if(!::arg().mustDo("master"))
-      return "PowerDNS not configured as master";
+  if(!::arg().mustDo("master") && !::arg().mustDo("slave-renotify"))
+      return "PowerDNS not configured as master or slave with re-notifications";
```
mind04: this change is useless without changing the code below

Habbie: without checking what you mean, that sounds like #3663 was broken too

klaus3000: I do not think this is useless. The code below is fine. Just the check was wrong: NOTIFYs can be sent either if PDNS has master mode active OR slave-renotify is enabled.
In my tests everthing worked fine.

mind04: We only notify 'all' if kind = 0 (master) so slave domains never get notified, and yes behaviour of #3663 is similar.
And second, if slave=no should we send out slave renotifies when notifying all?

klaus3000: > We only notify 'all' if kind = 0 (master) so slave domains never get notified
I do not understand what you mean. Currently PDNS has following outgoing NOTIFY logic:
1. zone is MASTER, or
2. zone is SLAVE and slave-renotify is enabled

This patch does not change this behaviour. But currently "pdns_control notify" supports only scenario 1, not scenario 2. This patch allows "pdns_control notify" in the second scenario.

> And second, if slave=no should we send out slave renotifies when notifying all?
You are right, I could add a check for pdns_control for this

klaus3000: I updated the pullrequest

mind04:
this line https://github.com/PowerDNS/pdns/pull/3889/commits/e64b1e21a4090316c06ef9b7112092db6bb4feff#diff-41c762c38ab4590eaf7b5a99356606f8L283 is preventing slave domains from being notified if you execute ```pdns_control notify '*'``` with slave-renotify enabled.
The counters also need some work if slave domains are added to the notify queue.

klaus3000:
Indeed. I never used pdns_control notify '*' up to now, I only notified zones explicitely. I will update the pull request. Once we have agreed, I will provide a pull request for master too.

klaus3000:
Hi! I checked the travis log but could not find out why the build failed. Can you please help me reading the build log?

klaus3000:
This pull request backports 2 things, 1 simple and 1 under discussion. Thus I close this pull request and add another one for the simple thing. The other thing has to be resolved in master first.

PR #3890, opened 2016-05-23
optimize only-notify logic

klaus3000:
do not execute the NS-records NOTIFY logic if the only-notify "whitelist"
is empty, as a target will never match an empty whitelist.

### DIFF
--- a/pdns/mastercommunicator.cc
+++ b/pdns/mastercommunicator.cc
@@ -49,30 +49,32 @@ void CommunicatorClass::queueNotifyDomain(const DNSName &domain, UeberBackend *B
   DNSResourceRecord rr;
   FindNS fns;
 
-  B->lookup(QType(QType::NS),domain);
-  while(B->get(rr))
-    nsset.insert(rr.content);
+  if (d_onlyNotify.size()) {
+    B->lookup(QType(QType::NS),domain);
+    while(B->get(rr))
+      nsset.insert(rr.content);
 
-  for(set<string>::const_iterator j=nsset.begin();j!=nsset.end();++j) {
-    vector<string> nsips=fns.lookup(DNSName(*j), B);
-    if(nsips.empty())
-      L<<Logger::Warning<<"Unable to queue notification of domain '"<<domain<<"': nameservers do not resolve!"<<endl;
-    else
-      for(vector<string>::const_iterator k=nsips.begin();k!=nsips.end();++k) {
-        const ComboAddress caIp(*k, 53);
-        if(!d_preventSelfNotification || !AddressIsUs(caIp)) {
-          if(!d_onlyNotify.match(&caIp))
-            L<<Logger::Info<<"Skipped notification of domain '"<<domain<<"' to "<<*j<<" because it does not match only-notify."<<endl;
-          else
-            ips.insert(caIp.toStringWithPort());
+    for(set<string>::const_iterator j=nsset.begin();j!=nsset.end();++j) {
+      vector<string> nsips=fns.lookup(DNSName(*j), B);
+      if(nsips.empty())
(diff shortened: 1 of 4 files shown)


pdns/mastercommunicator.cc
```diff
@@ -93,7 +95,7 @@ void CommunicatorClass::queueNotifyDomain(const DNSName &domain, UeberBackend *B
     }
   }
 
-  if (!hasQueuedItem)
+  if (!hasQueuedItem && d_onlyNotify.size())
```
mind04: Please don't hide the warning if only-notiy is empty, this also applies to also-notify.
If this warning pops up with an empty only-notiy list change the domain type from SLAVE to NATIVE

klaus3000: Your comment is wrong. I have PDNS with slave zones only and this slave zones should not send NOTIFYs to the NS-RRs. Some if my slave zones have further slave zones, thus I have slave-renotify enabled.
So this warning pops up every time a zone gets updated although everthing is correct.
It does not make sense to print out this warning if NS-RR NOTIFYs where explicitely disabled.

In my opinion this Warning was wrong from the beginning. There should not be a warning if a zone does not have any also-notifys (that's fine). IMO the warning makes only sense if the only-notify feature blocked all NS-RR NOTIFYs.

Habbie: How about a startup warning that empty `only-notify` prevents all notifications?

Habbie:
Can you reword 'fix' into 'optimize'? As far as I can see what we are fixing here is not actually broken.

pdns/mastercommunicator.cc
mind04: You are "fixing" a very specific re-notify case here. But you are hiding  a warning which also applies to master domains. If only-notify is empty they rely on also-notify to make things work. The warning is here to detect misconfigured master and slave-renotify domains. 
Yes, it will also warn for slave domains with an empty notification list, if global slave-renotify is enabled. But this is something you can't avoid without hiding real problems in other usecases.
In my opinion the only real fix, for your usecase, is add a SLAVE-RENOTIFY metadata option, so it is possible to enable renotification on a per domain basis. 

@Habbie a startup warning is useless because a part of the notify logic is stored in the domain metadata. Empty only-notify does not say a thing at startup.

klaus3000: OK, NOTIFYs to the NS hosts may be a common use case for Master zones, but not for slave zones. Thus, what about using the zone kind as decision if the warning should be logged? See last commit.

klaus3000:
Ping?

klaus3000:
I am a bit worried why this pull request is not accepted. In your blog you talk a lot about performance improvements for PDNS. This patch would disable DNS lookups if not needed anyways, so this would be a huge performance improvement in processing zone updates.

mind04:
Please revert the logging behaviour changes.
The last commit is very specific to your use-case. Silently failing slave re-notifications are a bad thing for most people.
And please read my previous comment about the startup warning. An empty only-notify list is only one part of the equation. The warning is at least confusing and is in some cases a complete lie.

Ps. Don't overestimate the performance improvement, most lookups are answered from the query cache.

klaus3000:
I still disagree about the logging. But let's split the issues. I made a new pull request to improve the logic:
https://github.com/PowerDNS/pdns/pull/4861

Once https://github.com/PowerDNS/pdns/pull/4861 is accepted, this here will be closed and I will create another issue regarding the logging.

ad PS: ...if the hostname is actually in the backend. Reviewing FindNS.lookup() seems strange too. Hostnames are resolved with getaddrinfo AND the backend.

pieterlexis:
closing in favor of #4861

PR #3891, opened 2016-05-23
fix only-notify logic

klaus3000:
do not execute the NS-records NOTIFY logic if the only-notify "whitelist"
is empty, as a target will never match an empty whitelist.


Habbie:
For future reference, please do not open backport PRs until something has been accepted on master :)

klaus3000:
I close this pull request. Once the issue is resolved in master I will open a new pull request with the backport.

PR #3905, opened 2016-05-26
Recursor: Add dnssec=process-no-validate option

pieterlexis:
This option turns the recursor into a "Security-Aware Recursive Name Server" (RFC 4033 §2), meaning it will pass on RRSIGs and NSEC(3)s but will not validate, even when the client requests it. This is similar to e.g. unbound without the validator enabled.

### DIFF
--- a/pdns/pdns_recursor.cc
+++ b/pdns/pdns_recursor.cc
@@ -960,7 +960,7 @@ void startDoResolve(void *p)
           }
 
           // Does the query or validation mode sending out a SERVFAIL on validation errors?
-          if(!pw.getHeader()->cd && (g_dnssecmode == DNSSECMode::ValidateAll || (dc->d_mdp.d_header.ad && g_dnssecmode != DNSSECMode::Off))) {
+          if(!pw.getHeader()->cd && (g_dnssecmode == DNSSECMode::ValidateAll || dc->d_mdp.d_header.ad)) {
             if(sr.doLog()) {
               L<<Logger::Warning<<"Sending out SERVFAIL for "<<dc->d_mdp.d_qname<<" because recursor or query demands it for Bogus results"<<endl;
             }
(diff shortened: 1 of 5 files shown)


pdns/pdns_recursor.cc
```diff
@@ -959,7 +959,7 @@ void startDoResolve(void *p)
           }
 
           // Does the query or validation mode sending out a SERVFAIL on validation errors?
-          if(!pw.getHeader()->cd && (g_dnssecmode == DNSSECMode::ValidateAll || (dc->d_mdp.d_header.ad && g_dnssecmode != DNSSECMode::Off))) {
+          if(!pw.getHeader()->cd && (g_dnssecmode == DNSSECMode::ValidateAll || (dc->d_mdp.d_header.ad && (g_dnssecmode != DNSSECMode::Off || g_dnssecmode != DNSSECMode::ProcessNoValidate)))) {
```
rgacogne: I think:
`(g_dnssecmode != DNSSECMode::Off || g_dnssecmode != DNSSECMode::ProcessNoValidate)`
should be:
`(g_dnssecmode != DNSSECMode::Off && g_dnssecmode != DNSSECMode::ProcessNoValidate)`

I guess it doesn't break because the condition on line 934 prevents us from getting here if `g_dnssecmode` is Off or ProcessNoValidate.

rgacogne:
Except for the previous comment, LGTM!

pdns/pdns_recursor.cc
pieterlexis: I think we can drop the whole `&& (g_dnssecmode != DNSSECMode::Off || g_dnssecmode != DNSSECMode::ProcessNoValidate)` as indeed the earlier condition prevents this

rgacogne: Indeed!

PR #3906, opened 2016-05-26
dnsdist: Keep the servers ordered inside pools

### LINKED ISSUE #3903: dnsdist - firstAvailable policy ignores order
There is a bug in the firstAvailable policy. It does not check the "order" in a server but selects the server in the order it is listed in the configuration.

Example:
newServer({address="192.168.0.1", order=2, qps=1000 })
newServer({address="192.168.0.2", order=1, qps=1000 })
newServer({address="192.168.0.3", order=3, qps=1000 })
setServerPolicy(firstAvailable)

In the above configuration 192.168.0.1 is selected instead of 192.168.0.2.

rgacogne:
Several policies expect the servers to be ordered on their 'order'.
In addition to that, we keep the servers in a `NumberedVector` to
be able to pass them as a Lua table to Lua custom policies, and
that means we need to get the numbers right there too, especially
when we remove a server from a pool.
Close #3903.

### DIFF
--- a/regression-tests.dnsdist/test_Routing.py
+++ b/regression-tests.dnsdist/test_Routing.py
@@ -296,3 +296,68 @@ class TestRoutingRoundRobinLBOneDown(DNSDistTest):
             total += value
 
         self.assertEquals(total, numberOfQueries * 2)
+
+class TestRoutingOrder(DNSDistTest):
+
+    _testServer2Port = 5351
+    _config_params = ['_testServerPort', '_testServer2Port']
+    _config_template = """
+    setServerPolicy(firstAvailable)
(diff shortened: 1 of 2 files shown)


regression-tests.dnsdist/test_Routing.py
```diff
@@ -213,7 +213,7 @@ def testRR(self):
         """
         Routing: Round Robin
 
-        Send 100 A queries to "rr.routing.tests.powerdns.com.",
+        Send 10 A queries to "rr.routing.tests.powerdns.com.",
```
Habbie: Was this change supposed to be in this commit?

Habbie:
One question/nit, otherwise LGTM.

PR #3910, opened 2016-05-27
Recursor: implement Negative Trust Anchor management and runtime Trust Anchor management

### LINKED ISSUE #3902: Recursor needs negative trust anchor management
[RFC 7646](https://tools.ietf.org/html/rfc7646).

pieterlexis:
This implementation *might* be a bit blunt, but it works. Please review.

Closes #3902

### DIFF
--- a/docs/manpages/rec_control.1.md
+++ b/docs/manpages/rec_control.1.md
@@ -43,8 +43,15 @@ To dump the cache to disk, execute:
      respond. Set to 0 for infinite.
 
 # COMMANDS
+add-nta *DOMAIN* [*REASON*]
+:    Add a Negative Trust Anchor for *DOMAIN*, suffixed optionally with *REASON*.
+
 current-queries
 :    Shows the currently active queries.
+
+clear-nta *DOMAIN*...
+:    Remove Negative Trust Anchor for one or more *DOMAIN*s. Set domain to `'*'`
+     to remove all NTA's.
 
 dump-cache *FILENAME*
 :    Dumps the entire cache to *FILENAME*. This file should
@@ -67,6 +74,9 @@ get *STATISTIC* [*STATISTIC*]...
 
 get-all
 :    Retrieve all known statistics.
+
+get-ntas
+:    Get a list of the currently configured Negative Trust Anchors.
 
 get-parameter *KEY* [*KEY*]...
 :    Retrieves the specified configuration parameter(s).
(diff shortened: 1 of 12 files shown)


pdns/rec_channel_rec.cc
```diff
@@ -341,6 +342,62 @@ string doSetCarbonServer(T begin, T end)
   return ret;
 }
 
+template<typename T>
+string doAddNTA(T begin, T end)
+{
+  if(begin == end)
+    return "No NTA specified, doing nothing\n";
+
+  auto luaconf = g_luaconfs.getLocal();
```
rgacogne: It looks like `luaconf` is unused.

pieterlexis: a leftover of an earlier incantation, removed

pdns/validate.cc
```diff
@@ -165,6 +165,19 @@ cspmap_t harvestCSPFromRecs(const vector<DNSRecord>& recs)
...
+   * tree
+   */
+  auto negAnchors = luaLocal->negAnchors;
+  if (!negAnchors.empty())
+    for (auto negAnchor : negAnchors)
+      if (zone.isPartOf(negAnchor.first)) {
+        LOG("Found a Negative Trust Anchor for "<<negAnchor.first<<", which was added with the message '"<<negAnchor.second<<"'. Returning Insecure."<<endl);
+        return NTA;
+      }
+
```
rgacogne: Shouldn't we check whether there is a (positive) trust anchor below? If I understand correctly section 1.1 of rfc7646, we would be expected to resume validation starting with this trust anchor, right?

pieterlexis:
Fixed @rgacogne's comments and added a bugfix that ensures the use of the correct, configured trust anchor (we only used the root trust anchor before)

pieterlexis:
an addition on the bugfix regarding positive trust anchors: this makes the Recursor behave as specified in [RFC 6840 §C.1](https://tools.ietf.org/html/rfc6840#appendix-C.1). i.e. using the closest encounter when validating responses.

pieterlexis:
Updated to also add runtime trust anchor management

docs/manpages/rec_control.1.md
```diff
@@ -43,9 +43,23 @@ To dump the cache to disk, execute:
...
+
+add-ta *DOMAIN* *DSRECORD*
+:    Add a Trust Anchor for *DOMAIN* with DS record data *DSRECORD*.
+
 current-queries
 :    Shows the currently active queries.
 
+clear-nta [*DOMAIN*]...
+:    Remove Negative Trust Anchor for one or more *DOMAIN*s. If *DOMAIN* is
+     empty, remove all configured Negative Trust Anchors.
```
Habbie: 'empty' or 'no domain specified'? I think we might want to be explicit here, use '*' or something. If somebody automates setting/clearing of NTAs it's pretty easy to accidentally pass an empty list

Habbie:
I approve of this PR!

Habbie:
(one nit)

PR #3912, opened 2016-05-28
Don't eat exceptions in GSQLBackend::get, include these as errors in …

hlindqvist:
Don't eat exceptions in GSQLBackend::get, include these as errors in pdnsutil check-zone.

An example of a problem that was otherwise silently ignored was a record with owner name `toolongxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx.example.com` (too long label).

### DIFF
--- a/pdns/backends/gsql/gsqlbackend.cc
+++ b/pdns/backends/gsql/gsqlbackend.cc
@@ -1055,7 +1055,6 @@ bool GSQLBackend::get(DNSResourceRecord &r)
   // L << "GSQLBackend get() was called for "<<qtype.getName() << " record: ";
   SSqlStatement::row_t row;
 
-skiprow:
   if(d_query_stmt->hasNextRow()) {
     try {
       d_query_stmt->nextRow(row);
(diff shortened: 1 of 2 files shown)


hlindqvist:
Alternatively, should it just abort immediately in pdnsutil? Is there a chance that it will loop endlessly printing errors from backend in other scenarios?

pdns/backends/gsql/gsqlbackend.cc
```diff
@@ -1065,8 +1064,8 @@ bool GSQLBackend::get(DNSResourceRecord &r)
     }
     try {
       extractRecord(row, r);
-    } catch (...) {
-      goto skiprow;
+    } catch (std::exception &e) {
+      throw PDNSException("Exception '"+string(e.what())+"' occured for row: "+boost::algorithm::join(row, " | "));
```
Habbie: occurred

Habbie:
This PR is for AFTER 4.0.0, and I suggest not putting it in 4.0.1 etc. either. It changes behaviour too much to do this now that we have already released a beta. In other words, merge to master when we have a 4.0.x release branch.

mind04:
Please never merge it like this. Blow up a working backend for one row of bad data, cant be right.
Passing state to pdnsutil check-zone is the way to go.

Habbie:
Agree - SERVFAIL is good, blowing up a whole backend is not.

pieterlexis:
moved to an issue ( #4941)

PR #3913, opened 2016-05-28
Validate names with address records to be valid hostnames in pdnsutil…

hlindqvist:
Validate names with address records to be valid hostnames in pdnsutil check-zone.

### DIFF
--- a/pdns/pdnsutil.cc
+++ b/pdns/pdnsutil.cc
@@ -564,6 +564,17 @@ int checkZone(DNSSECKeeper &dk, UeberBackend &B, const DNSName& zone, const vect
       }
     }
 
+    if(rr.qtype.getCode() == QType::A || rr.qtype.getCode() == QType::AAAA)
+    {
+      Regex hostnameRegex=Regex("^(([A-Za-z0-9]([A-Za-z0-9-]*[A-Za-z0-9])?)\\.)+$");
+      if (!hostnameRegex.match(rr.qname.toString()))
+      {
+        cout<<"[Error] A or AAAA record found at '"<<rr.qname.toString()<<"'. This name is not a valid hostname."<<endl;
+        numerrors++;
+        continue;
+      }
+    }
+
     if (rr.qtype.getCode() == QType::CNAME) {
       if (!cnames.count(rr.qname))
         cnames.insert(rr.qname);


pdns/pdnsutil.cc
```diff
@@ -564,6 +564,17 @@ int checkZone(DNSSECKeeper &dk, UeberBackend &B, const DNSName& zone, const vect
       }
     }
 
+    if(rr.qtype.getCode() == QType::A || rr.qtype.getCode() == QType::AAAA)
+    {
+      Regex hostnameRegex=Regex("^(([A-Za-z0-9]([A-Za-z0-9-]*[A-Za-z0-9])?)\\.)+$");
+      if (!hostnameRegex.match(rr.qname.toString()))
+      {
+        cout<<"[Error] A or AAAA record found at '"<<rr.qname.toString()<<"'. This name is not a valid hostname."<<endl;
```
Habbie: This could use an RFC reference. I'm pondering whether we should make it a warning instead of an error, but I'm unsure.

hlindqvist: @Habbie is it only https://tools.ietf.org/html/rfc952 + https://tools.ietf.org/html/rfc1123#section-2 that you had in mind (defines what is a valid hostname) or also something to support the inference that a name with address records should be a hostname?  Also, is this for a code comment or did you want the RFC references in the output?

Habbie: 1123 2 is fine. Comment would be good, message would become too long with it I think.

mind04:
1123 says "Host domain names MUST be translated to IP addresses". It does not limit the dnsname to ip translation to valid hostnames.

Habbie:
Ack, let's go for Warning then.

mind04:
I prefer [info], output warnings for valid dns is confusing.

Habbie:
Oh, we have Info? Works for me.

hlindqvist:
If the stance is that the existence of `A` records has no implication as to the type of name the owner name is, then I don't think it makes sense to have a message at all.
(That was not my interpretation but if that's what you're going with I think it makes more sense to do nothing.) 

However, I'm not sure why 1123 specifically was referenced in the context of what `A` may actually mean, that seems more like a 1034/1035 deal,  @mind04 ?

PR #3918, opened 2016-05-31
auth: Disable dnsbulktest and dnstcpbench when boost is too old

rgacogne:
An other option would have been to remove the corresponding targets from the Makefile if boost is too old, but comparing version in the Makefile seemed wrong.

### DIFF
--- a/pdns/dnsbulktest.cc
+++ b/pdns/dnsbulktest.cc
@@ -1,10 +1,15 @@
 #ifdef HAVE_CONFIG_H
 #include "config.h"
 #endif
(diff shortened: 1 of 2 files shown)


Habbie:
@mind04 I think you're the target audience for this, any comments?

Habbie:
I'm not a big fan of shipping binaries that can do nothing but print "I was compiled to not work"

mind04:
+1 on the not shipping dummy binaries.
Just don't build them if boot is too old

PR #3919, opened 2016-05-31
dnsdist: Add `sourceBindAny` to allow a nonlocal address as source

rgacogne:
It allows specifying a source address to reach this backend that might not exist yet
on the host but will later, for example in HA setups.

### DIFF
--- a/pdns/README-dnsdist.md
+++ b/pdns/README-dnsdist.md
@@ -202,6 +202,12 @@ The supported values for `source` are:
 
 Specifying the interface name is only supported on system having IP_PKTINFO.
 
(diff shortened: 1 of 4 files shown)


Habbie:
LGTM. Did not know you could do this on 'client' sockets, nice :)

rgacogne:
On Linux this only works for IPv6 sockets, not IPv4 (`bind()` succeed but then `connect()` doesn't). Therefore I'd be inclined to just give up on that idea. Any thought?

rgacogne:
Closing as it would be misleading to have this working only for IPv6.

PR #3926, opened 2016-06-01
Recursor: set +CD on queries forwarded to a recursor

pieterlexis:
to comply with [RFC 6840 § 5.9](https://tools.ietf.org/html/rfc6840#section-5.9)

### DIFF
--- a/pdns/lwres.cc
+++ b/pdns/lwres.cc
@@ -65,7 +65,18 @@ int asyncresolve(const ComboAddress& ip, const DNSName& domain, int type, bool d
 
   pw.getHeader()->rd=sendRDQuery;
   pw.getHeader()->id=dns_random(0xffff);
-  
+  /* RFC 6840 section 5.9:
+   *  This document further specifies that validating resolvers SHOULD set
+   *  the CD bit on every upstream query.  This is regardless of whether
+   *  the CD bit was set on the incoming query [...]
+   *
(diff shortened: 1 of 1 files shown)


Habbie:
As I mentioned on IRC, BIND does not do this, with some pretty strongly worded arguments. Of course, if you start with -CD, you need to be able to fall back to +CD when -CD keeps giving you SERVFAIL. So this patch is fine with me but we should keep this discussion in mind for later :)

pdns/lwres.cc
```diff
@@ -65,7 +65,18 @@ int asyncresolve(const ComboAddress& ip, const DNSName& domain, int type, bool d
...
   pw.getHeader()->id=dns_random(0xffff);
-  
+  /* RFC 6840 section 5.9:
+   *  This document further specifies that validating resolvers SHOULD set
+   *  the CD bit on every upstream query.  This is regardless of whether
+   *  the CD bit was set on the incoming query [...]
+   *
+   * sendRDQuery is only true if the qname is part of a forward-zone-recurse (or
+   * set in the forward-zone-file), so we use this as an indicator for it being
+   * an "upstream query". To stay true to "dnssec=off means 3.X behaviour, we
```
Habbie: missing "

Habbie:
one nit, then LGTM.

PR #3931, opened 2016-06-03
Rename notify to pdns_notify to avoid conflicts

zeha:
Note that this does not change regression-tests.recursor for now,
because that needs a newer pdns-tools deb first (for travis).

This is set out to fix [Debian bug #825804](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=825804).


Habbie:
travis failure is unrelated. LGTM. @pieterlexis?

Issue #2590, opened 2015-06-12
Please support emitting CollectD metrics


seitzbg:
Please support emitting CollectD metrics emitting.

carbon: 
lieter | phibs: [graphite] 2015/06/12 17:12:19 unable to parse data: received "pdns.influx-grafana-test.auth.udp6-queries" which doesn't conform to format of key.value.key.value.name or name 

lieter | I did some work on the collectd plugin for logstash some time ago
lieter | phibs: maybe we should hack in collectd metrics :D
lieter | phibs: can you make a ticket?

rubenk:
collectd has quite a nice pdns plugin for what it's worth ;)

pieterlexis:
yes, this is true. But emitting for e.g. [influxdb](http://influxdb.com) and [prometheus](http://prometheus.io/) (both HTTP) would be cool as well.

seitzbg:
Also not needing collectd if I don't already use it for other metrics, priceless ;)

janeczku:
@pieterlexis In the meantime: https://github.com/janeczku/powerdns_exporter

PR #3935, opened 2016-06-06
Prometheus statistics

giganteous:
This set of commits tries to mirror the graphite stats as a /prometheus endpoint on the API webserver. Prometheus likes to call this [directly instrumentated software](https://prometheus.io/docs/instrumenting/exporters/#directly-instrumentated-software).

Things I'd like to fix before a merge is
- [ ] naming / documenting all metrics on g_dstates
- [ ] naming / documenting all metrics on g_frontends
- [ ] naming / documenting all metrics on g_pools
- [ ] agreeing on the endpoint. Prometheus has a default of /metrics

But to get a general idea about if I'm solving this the correct way I'm creating the PR now.

To be clear: prometheus stats don't necessarily need a TYPE or HELP string but aids end users. Giving them the same name and structure should keep you from having to maintain 3 different branches where they're used, iiuc.

### DIFF
--- a/pdns/dnsdist-web.cc
+++ b/pdns/dnsdist-web.cc
@@ -296,12 +296,12 @@ static void connectionThread(int sock, ComboAddress remote, string password, str
         str<<"# a first stab at reporting prometheus metrics from inside powerdns\n";
         string mainPool = "main";
         for(const auto& e : g_stats.entries) {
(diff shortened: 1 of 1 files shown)


giganteous:
Prometheus was mentioned in issue #2590, but does not fix the graphite/carbon naming.

PR #3936, opened 2016-06-06
dnsdist: Filter on opcode, records count/type, trailing data

rgacogne:
* Add `OpcodeRule()` to filter on opcode + DNSOpcode.* Lua values
* Add `TrailingDataRule()` to filter queries with trailing data
* Add `RecordsCountRule(section, minCount, maxCount)` to match on
the number of records in a given section
* Add `RecordsTypeCountRule(section, type, minCount, maxCount)` to
match on the number of records of type `type` in a given section
* Add DNSSection.* Lua values
* Add DNSClass.* Lua values

### DIFF
--- a/pdns/dnsparser.cc
+++ b/pdns/dnsparser.cc
@@ -652,14 +652,16 @@ void ageDNSPacket(char* packet, size_t length, uint32_t seconds)
     uint64_t n;
     for(n=0; n < ntohs(dh.qdcount) ; ++n) {
       dpm.skipLabel();
-      dpm.skipBytes(4); // qtype, qclass
+      /* type and class */
+      dpm.skipBytes(4);
     }
    // cerr<<"Skipped "<<n<<" questions, now parsing "<<numrecords<<" records"<<endl;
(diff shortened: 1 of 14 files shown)


pdns/dnsparser.cc
```diff
@@ -713,3 +717,112 @@ uint32_t getDNSPacketMinTTL(const char* packet, size_t length)
...
+    const uint16_t qdcount = ntohs(dh->qdcount);
+    for(size_t n = 0; n < qdcount; ++n) {
+      dpm.skipLabel();
+      dpm.skipBytes(4); // qtype, qclass
+    }
+    const size_t numrecords = ntohs(dh->ancount) + ntohs(dh->nscount) + ntohs(dh->arcount);
+    for(size_t n = 0; n < numrecords; ++n) {
+      dpm.skipLabel();
+
+      /* const uint16_t dnstype */ dpm.get16BitInt();
```
zeha: i'd suggest not having multi-line comment syntax and actual code on the same line

PR #3941, opened 2016-06-06
make sure we can resolve 'x.com' if we have auth NS records loaded for x.com

ahupowerdns:


Habbie:
This would be easy to add a test for!

PR #3949, opened 2016-06-07
Implement IXFR-based slaving for Authoritative, fix duplicate AXFRs

ahupowerdns:
With this PR, we gain the conditional ability to slave a domain over IXFR, with some documented limitations mostly involving DNSSEC. 

At this late stage in 4.0.0 development, this change sadly also impacted regular AXFR but this could not be avoided. For starters, the old AXFR logic would happily launch two AXFR-attempts simultaneously, which messed with our ability to actually test IXFR. 

Secondly, the AXFR code performs a LOT of analysis on zones to filter records and determine zone status, an this logic had to be abstracted to also be available for IXFR. 

This PR passes all the regression tests, but I estimate (without proof) that it will inconvenience at least some users. I feel this is offset however by cleaning up the embarrassing double-AXFR situation, which must have been hurting other people!

### DIFF
--- a/docs/markdown/authoritative/modes-of-operation.md
+++ b/docs/markdown/authoritative/modes-of-operation.md
@@ -85,7 +85,26 @@ PowerDNS supports multiple masters. For the BIND backend, the native BIND
 configuration language suffices to specify multiple masters, for SQL based backends,
 list all master servers separated by commas in the 'master' field of the domains table.
 
-Since version 4.0.0, PowerDNS requires that masters sign their notifications. During transition and interoperation with other nameservers, you can use options **allow-unsigned-notify** to permit unsigned notifications. For 4.0.0 this is turned off by default, but it might be turned on permanently in future releases.
+Since version 4.0.0, PowerDNS requires that masters sign their
+notifications.  During transition and interoperation with other nameservers,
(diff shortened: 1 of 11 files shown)


docs/markdown/authoritative/modes-of-operation.md
```diff
@@ -85,7 +85,26 @@ PowerDNS supports multiple masters. For the BIND backend, the native BIND
 configuration language suffices to specify multiple masters, for SQL based backends,
 list all master servers separated by commas in the 'master' field of the domains table.
 
-Since version 4.0.0, PowerDNS requires that masters sign their notifications. During transition and interoperation with other nameservers, you can use options **allow-unsigned-notify** to permit unsigned notifications. For 4.0.0 this is turned off by default, but it might be turned on permanently in future releases.
+Since version 4.0.0, PowerDNS requires that masters sign their
+notifications.  During transition and interoperation with other nameservers,
+you can use options **allow-unsigned-notify** to permit unsigned
+notifications.  For 4.0.0 this is turned off by default, but it might be
```
pieterlexis: This comment is the wrong way around (still). the setting is set to 'yes' now, but might be 'no' later

PR #3957, opened 2016-06-07
recursor: Require = in forward-zones

### LINKED ISSUE #3956: pdns-recursor forward-zones eats whitespace-separated addresses with no warning
I recently discovered that my forward-zones syntax has been wrong (for years). I only discovered this when running pdns-recursor in foreground mode for another reason. I had configuration similar to:

forward-zones=\
        foo=1.2.3.4 1.2.3.5,\
        bar=1.2.3.4 1.2.3.5

which was silently being parsed as "foo=1.2.3.4,bar=1.2.3.4", thus never using the second address. The correct syntax is with ';' instead of ' '. Perhaps a warning should be emitted for this mistake?

cmouse:
Fixes #3956


sim-:
This passes my testing.

PR #3962, opened 2016-06-08
fix openbsd support in auth and dnsdist

### LINKED ISSUE #3009: b71b60e breaks sendto on OpenBSD(&others?)
```
Dec 07 11:19:09 Error sending reply with sendmsg (socket=4, dest=127.0.0.1:31360): Invalid argument
```

commit link b71b60ee73ef3c86f80a2179981eda2e61c4363f

Habbie:
fixes #3009

### DIFF
--- a/pdns/misc.cc
+++ b/pdns/misc.cc
@@ -845,7 +845,6 @@ void addCMsgSrcAddr(struct msghdr* msgh, void* cmsgbuf, const ComboAddress* sour
     memset(pkt, 0, sizeof(*pkt));
     pkt->ipi6_addr = source->sin6.sin6_addr;
     pkt->ipi6_ifindex = itfIndex;
-    msgh->msg_controllen = cmsg->cmsg_len; // makes valgrind happy and is slightly better style
   }
   else {
 #ifdef IP_PKTINFO
@@ -863,7 +862,6 @@ void addCMsgSrcAddr(struct msghdr* msgh, void* cmsgbuf, const ComboAddress* sour
     memset(pkt, 0, sizeof(*pkt));
     pkt->ipi_spec_dst = source->sin4.sin_addr;
     pkt->ipi_ifindex = itfIndex;
-    msgh->msg_controllen = cmsg->cmsg_len;
 #endif
 #ifdef IP_SENDSRCADDR
     struct in_addr *in;
@@ -878,7 +876,6 @@ void addCMsgSrcAddr(struct msghdr* msgh, void* cmsgbuf, const ComboAddress* sour
 
     in = (struct in_addr *) CMSG_DATA(cmsg);
     *in = source->sin4.sin_addr;
-    msgh->msg_controllen = cmsg->cmsg_len;
 #endif
   }
 }
--- a/pdns/misc.cc
+++ b/pdns/misc.cc
@@ -826,6 +826,9 @@ Regex::Regex(const string &expr)
     throw PDNSException("Regular expression did not compile");
 }
 
+// if you end up here because valgrind told you were are doing something wrong
+// with msgh->msg_controllen, please refer to https://github.com/PowerDNS/pdns/pull/3962
+// first.
 void addCMsgSrcAddr(struct msghdr* msgh, void* cmsgbuf, const ComboAddress* source, int itfIndex)
 {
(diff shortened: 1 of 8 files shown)


Habbie:
3rd commit basically reverts https://github.com/PowerDNS/pdns/commit/9af87ad80c04f5e52f26335dd0d94eb454bbec90 - have not given this deeper thought yet.

Habbie:
@fobser this PR may be relevant to your interest

Habbie:
Review welcome at this point. I have some minor nits of my own left that I will fix before merge.

pieterlexis:
nit: maybe guard the includes as they are not required on non-OpenBSD?

Habbie:
Not going to guard `unistd.h`; but yes, `history.h` is a candidate for guarding.

Habbie:
OpenBSD CURRENT (snapshot 6 June 2016) does not need the sendmsg ('anti-valgrind') change, by the way.

pdns/misc.cc
```diff
@@ -845,7 +845,6 @@ void addCMsgSrcAddr(struct msghdr* msgh, void* cmsgbuf, const ComboAddress* sour
     memset(pkt, 0, sizeof(*pkt));
     pkt->ipi6_addr = source->sin6.sin6_addr;
     pkt->ipi6_ifindex = itfIndex;
-    msgh->msg_controllen = cmsg->cmsg_len; // makes valgrind happy and is slightly better style
```
rgacogne: The initial commit introducing this line, 9af87ad80c04f5e52f26335dd0d94eb454bbec90, points to an interesting discussion on the matter: https://codereview.chromium.org/3026044 
rfc3542 states, in section 20.2 "The cmsghdr Structure": "While sending an application may or may not include padding at the end of last ancillary data in msg_controllen and implementations must accept both as valid.", and then uses CMSG_SPACE() in 21.1.

Habbie: Without having read it all, are you saying that OpenBSD was wrong to complain?

Habbie:
Updated my OpenBSD current comment because I was wrong. 5.9 and current are completely identical in this behaviour.

pdns/misc.cc
rgacogne: Well, yes and I wonder why, but valgrind is wrong too. After thinking about it a little more, this patch fixes a known issue and complies with the man page, so it's the right thing to do.

Habbie: Sounds good. I'll stick a comment in for when somebody finds it with valgrind again.

Habbie:
Ok, I feel this is done. Ready for merge and review.

fobser:
Thanks for doing this. I tried to compile 4.0.0-beta1 the other day on OpenBSD and it was failing. Didn't have the time to look into it then and report on it. With these patches the port compiles. Haven't tried running it yet though.

Habbie:
@fobser I did a few of our regression suites with it and they fully pass except the few tests that try to pass -p to dig ;)

pdns/nameserver.cc
```diff
@@ -291,12 +291,10 @@ void UDPNameserver::send(DNSPacket *p)
...
   fillMSGHdr(&msgh, &iov, cbuf, 0, (char*)buffer.c_str(), buffer.length(), &p->d_remote);
 
+  msgh.msg_control=NULL;
   if(p->d_anyLocal) {
     addCMsgSrcAddr(&msgh, cbuf, p->d_anyLocal.get_ptr(), 0);
   }
-  else {
-    msgh.msg_control=NULL;
-  }
   DLOG(L<<Logger::Notice<<"Sending a packet to "<< p->getRemote() <<" ("<< buffer.length()<<" octets)"<<endl);
```
rgacogne: I know this PR is about auth and dnsdist, but it looks like we have the exact same construct in two places in pdns_recursor.cc, perhaps we could apply the same fix there?

Habbie: Oh, that's a good idea. I did not look into recursor when I realised it would never run for lack of both ucontext and boost context.

Habbie: added!

pieterlexis:
Is this ready to be merged?

Habbie:
Yes

PR #3972, opened 2016-06-10
API: change PATCH/PUT on zones to return 204 No Content instead of full zone

zeha:
Breaks backwards compat, obviously.

Requested by a user with large zones; in general this makes sense to not waste bandwidth and processing power. Most users probably throw away the result without looking at it.


zeha:
I imagine pdnscontrol being broken, though.

ton31337:
:+1:

PR #3974, opened 2016-06-10
Add option to disable the stdout log timestamp

pieterlexis:
And add it to the unit files. This fixes the "double timestamps" issue.

### DIFF
--- a/pdns/logger.cc
+++ b/pdns/logger.cc
@@ -47,14 +47,17 @@ void Logger::log(const string &msg, Urgency u)
 #ifndef RECURSOR
   bool mustAccount(false);
 #endif
-  struct tm tm;
-  time_t t;
-  time(&t);
-  tm=*localtime(&t);
-
   if(u<=consoleUrgency) {
     char buffer[50];
-    strftime(buffer,sizeof(buffer),"%b %d %H:%M:%S ", &tm);
+    buffer[0] = '\0';
+    if(!d_disableTimeStamp) {
+      struct tm tm;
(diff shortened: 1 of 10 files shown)


Habbie:
The travis failures appear to be real. rebase perhaps?

pdns/logger.cc
```diff
@@ -47,14 +47,17 @@ void Logger::log(const string &msg, Urgency u)
...
     char buffer[50];
-    strftime(buffer,sizeof(buffer),"%b %d %H:%M:%S ", &tm);
+    buffer[0] = '\0';
+    if(!d_disableTimeStamps) {
+      struct tm tm;
+      time_t t;
+      time(&t);
+      tm=*localtime(&t);
+
+      strftime(buffer,sizeof(buffer),"%b %d %H:%M:%S ", &tm);
```
Habbie: Unrelated to this patch but we should, theoretically, be checking the return value of `strftime`.

pdns/logger.hh
```diff
@@ -60,6 +60,10 @@ public:
     d_disableSyslog = d;
   }
 
+  void disableTimeStamps(bool d) {
+    d_disableTimeStamps = d;
```
Habbie: the setting is singular but this bool and method are plural, looks a bit weird. I suggest making the setting plural.

Habbie:
Can you rebase and look at the nits? Thanks!

pieterlexis:
closing for now, this fails to build on travis but works locally

Issue #3118, opened 2015-12-25
4.0 tarball fileset


zeha:
Auth:

- ext/mbedtls/gpl-2.0.txt missing
- pdns/dnsdistdist is uselessly included (and has .js files with addtl. licenses...)
- missing source files for docs/markdown/theme/js/bootstrap-3.0.3.min.js docs/markdown/theme/js/prettify-1.0.min.js pdns/dnsdistdist/html/js/d3.v3.js pdns/dnsdistdist/html/js/jquery-1.8.3.min.js pdns/dnsdistdist/html/js/moment.min.js pdns/dnsdistdist/html/js/rickshaw.min.js pdns/dnsdistdist/html/js/underscore-min.js

Recursor:

- includes unusable build-scripts file(!)

pieterlexis:
> missing source files for 
* docs/markdown/theme/js/bootstrap-3.0.3.min.js
* docs/markdown/theme/js/prettify-1.0.min.js

These were (are) grabbed as-is from the [mkdocs](http://mkdocs.org) themes. I could add a small readme/license file.

zeha:
@pieterlexis please do, also debian needs the unminified sources, too.

pieterlexis:
mkdocs doesn't ship the un-minified sources :disappointed:

zeha:
> > missing source files for 
> * docs/markdown/theme/js/bootstrap-3.0.3.min.js
> * docs/markdown/theme/js/prettify-1.0.min.js

> These were (are) grabbed as-is from the [mkdocs](http://mkdocs.org) themes. I could add a small readme/license file.

Just looked at docs/markdown/theme/css/bootstrap-custom.min.css, and I'm thinking shipping source for that is either hopeless or a very enterprisey task.
Would you be ok with rearranging the prebuilt sources into a vendor/ folder?
That way I can just exclude them when importing your tarballs and symlink in the files from the libjs-bootstrap package (OTOH this means following major version upgrades in a timely manner, I think).

pieterlexis:
Updated the list, only the javascript issues remain.

rubenk:
The checklist says this is done for the recursor, but pdns-recursor-4.0.0-alpha2.tar.bz2 still contains minified javascript files.

rubenk:
IANAL, but can we mention something somewhere about the license of those files?

rubenk:
The jquery in recursordistdist is 1.8.3, which is unmaintained, and has some vulnerabilities.

[...]

PR #3981, opened 2016-06-11
Import Javascript sources for libs shipped with Recursor

zeha:
* d3 v3.3.6
* jquery v1.8.3
* moment v2.9.0
* rickshaw v1.4.6
* underscore v1.8.3

Fixes last missing files in Recursor tarball for #3118.

### DIFF
--- a/pdns/recursordist/html/js/d3.js
+++ b/pdns/recursordist/html/js/d3.js
@@ -1,0 +1,8981 @@
+d3 = function() {
+  var d3 = {
+    version: "3.3.6"
+  };
+  if (!Date.now) Date.now = function() {
+    return +new Date();
+  };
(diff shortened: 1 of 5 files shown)


zeha:
dnsdist appears to use completely different versions of these libs.

pieterlexis:
@rgacogne is it possible to sync these libraries between the recursor and dnsdist?

rgacogne:
dnsdist has newer versions for every common lib. At a quick glance, the only upgrade that might cause some trouble is jquery 1.8.3 -> 1.12.X.

Habbie:
This is the last open pull for recursor 4.0.0. Can we merge this to get 4.0.0 out the door and deal with dnsdist after that?

zeha:
Not seeing why not. Syncing the lib versions after 4.0.0 is certainly possible :)

PR #4005, opened 2016-06-16
Implement subcommand printing all KSK DS records in pdnsutil

horazont:
The subcommands prints all KSK DS records of the given zone to stdout. Diagnostics are exclusively printed to stderr, and if the zone is not secured this is fatal.

Merge this -or- #4007

### DIFF
--- a/pdns/pdnsutil.cc
+++ b/pdns/pdnsutil.cc
@@ -1673,6 +1673,86 @@ bool showZone(DNSSECKeeper& dk, const DNSName& zone)
   return true;
 }
 
+bool exportZoneDS(DNSSECKeeper& dk, const DNSName& zone)
+{
+  UeberBackend B("default");
+  DomainInfo di;
+
+  static const std::vector<std::pair<unsigned int, std::string> > DS_HASH_NAMES = {
+    std::make_pair(1, "SHA1"),
+    std::make_pair(2, "SHA256"),
+    std::make_pair(3, "GOST R 34.11-94"),
+    std::make_pair(4, "SHA384"),
+  };
+
+  if (!B.getDomainInfo(zone, di)) {
+    cerr << "No such zone in the database" << endl;
+    return false;
+  }
+
+  if (!dk.isSecuredZone(zone)) {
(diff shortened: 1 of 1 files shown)


pdns/pdnsutil.cc
```diff
@@ -1673,6 +1673,86 @@ bool showZone(DNSSECKeeper& dk, const DNSName& zone)
...
+
+  if(keys.empty()) {
+    cerr << "No keys for zone '" << zone.toString() << "'."<<endl;
+    return true;
+  }
+
+  sort(keys.begin(),keys.end());
+  reverse(keys.begin(),keys.end());
+  for(const auto& key : keys) {
+    if (key.d_flags != 257) {
```
Habbie: On zones with ONLY flags=256 keys, you should print those. I'm pretty sure you can reuse this logic from somewhere else.

Habbie:
This code is longer than I expected, giving weight to your initial idea of making it a flag to showZone.

horazont:
The decision not to make it an option was initially mainly due to the lack of an argument parsing framework for the subcommands.

However, after I then had a look at showZone while re-implementing the neccessary functionality in exportZoneDS, I figured that the simplifications I can make in exportZoneDS and the readability cost which would have incurred from adding a lot of branching to showZone is still worth the duplication.

As a middle way, it might make sense to factor out the actual printing of a single DS record into a separate function and use it from both showZone and exportZoneDS. I would be happy to do that and add it to this pull request.

showZone could in general use some clean up (as far as I can tell, it could be restructured in the same way I structured exportZoneDS: move the loop to print the key information out of the branches). I can also add that to this or a separate pull request.

Re zones with only ZSKs: sure, I can add that.

Habbie:
Ah yes, option goes both ways. I see your argument parsing issue, yes. But what you could also do is add a `DSexport` bool to `showZone` and call `showZone(.., .., true)` instead of your function. This feels like a better refactoring than your suggestion (but I could be wrong).

If a separate refactoring of showZone is in order, a separate PR would be appreciated. Thanks!

pieterlexis:
merged in #4007

PR #4007, opened 2016-06-16
Implement subcommand printing all KSK DS records in pdnsutil (by adapting showZone)

horazont:
The subcommands prints all KSK DS records of the given zone to stdout. Diagnostics are exclusively printed to stderr, and if the zone is not secured this is fatal.

This is a different approach to the same thing as #4005. Only one of the two should be merged.

### DIFF
--- a/pdns/pdnsutil.cc
+++ b/pdns/pdnsutil.cc
@@ -1517,87 +1517,102 @@ int setZoneKind(const DNSName& zone, const DomainInfo::DomainKind kind)
   return EXIT_SUCCESS;
 }
 
-bool showZone(DNSSECKeeper& dk, const DNSName& zone)
+bool showZone(DNSSECKeeper& dk, const DNSName& zone, bool exportDS = false)
 {
   UeberBackend B("default");
   DomainInfo di;
-  std::vector<std::string> meta;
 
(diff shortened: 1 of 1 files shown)


pdns/pdnsutil.cc
```diff
@@ -1494,87 +1494,102 @@ int setZoneKind(const DNSName& zone, const DomainInfo::DomainKind kind)
...
+      else
+        cout<<"No SOA serial found in database"<<endl;
     }
-    else
-      cout<<"No SOA serial found in database"<<endl;
   }
 
-
   if(!dk.isSecuredZone(zone)) {
-    cout<<"Zone is not actively secured"<<endl;
```
Habbie: For `show-zone`, this should stay `cout`. Perhaps move the `cerr` into `exportDS` and add an `else` for the `cout`.

Habbie:
One nit; rest looks good at a -quick- glance. Should get some more review before merge.

horazont:
Fixed the nit. The travis build seems to fail for other reasons, as far as I can tell.

pieterlexis:
Has conflicts, could you rebase? Thanks!

horazont:
@pieterlexis I rebased, but I cannot test as I do not have a 4.x installation at hand.

PR #4016, opened 2016-06-20
rec: Log protobuf messages for cache hits. Add policy tags in gettag()

rgacogne:
`gettag()` can now return an optional policy tags table in addition to
the existing `tag` integer.
Question and response protobuf messages are now sent even on cache hits.
When protobuf logging is enabled, the protobuf response message is
added to the cache and retrieved together with the response.

### DIFF
--- a/pdns/lua-recursor4.cc
+++ b/pdns/lua-recursor4.cc
@@ -44,7 +44,7 @@ bool RecursorLua4::ipfilter(const ComboAddress& remote, const ComboAddress& loca
   return false;
 }
 
-int RecursorLua4::gettag(const ComboAddress& remote, const Netmask& ednssubnet, const ComboAddress& local, const DNSName& qname, uint16_t qtype)
+int RecursorLua4::gettag(const ComboAddress& remote, const Netmask& ednssubnet, const ComboAddress& local, const DNSName& qname, uint16_t qtype, std::vector<std::string>* policyTags)
 {
   return 0;
 }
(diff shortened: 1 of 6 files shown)


pdns/lua-recursor4.cc
```diff
@@ -343,6 +343,12 @@ RecursorLua4::RecursorLua4(const std::string& fname)
   d_lw->registerFunction("getRecords", &DNSQuestion::getRecords);
   d_lw->registerFunction("setRecords", &DNSQuestion::setRecords);
 
+  d_lw->writeFunction("addPolicyTag", [](vector<string>* tags, const string& tag) {
+      if (tags) {
+        tags->push_back(tag);
+      }
+    });
+
```
Habbie: This weird idiom (If I see it correctly, inside `gettag` I am supposed to call `addPolicyTag(policytags, 'foo')`) is because we don't have any state object to speak of, like dq?

rgacogne: Yes. I couldn't find a way to make policytags directly writable from Lua, but I'm open to all suggestion!

Habbie: Options I have in mind:
1. make table.insert work on it, would be just about as weird though
2. change from `return 5` to `return 5, {"foo", "ipsum"}`

I don't think 1 is important. 2 might be nice but what you have now is fine with me!

Habbie:
conflicts.. probably because of the other merge of your stuff I just did

rgacogne:
Rebased to fix a conflict.

pdns/lua-recursor4.cc
rgacogne: I always forget that LuaWrapper handles returning multiple values. I'll try to make option 2 work, I like it a lot better than the current one.

rgacogne:
New version, this time `gettag()` can return an optional table of policy tags in addition to the existing tag.

Habbie:
Change looks good. Feel free to squash. I did not review the 'log protobuf for cache hits' part.

pdns/lua-recursor4.cc
```diff
@@ -480,10 +480,21 @@ bool RecursorLua4::ipfilter(const ComboAddress& remote, const ComboAddress& loca
...
-int RecursorLua4::gettag(const ComboAddress& remote, const Netmask& ednssubnet, const ComboAddress& local, const DNSName& qname, uint16_t qtype)
+int RecursorLua4::gettag(const ComboAddress& remote, const Netmask& ednssubnet, const ComboAddress& local, const DNSName& qname, uint16_t qtype, std::vector<std::string>* policyTags)
 {
-  if(d_gettag)
-    return d_gettag(remote, ednssubnet, local, qname, qtype);
+  if(d_gettag) {
+    auto ret = d_gettag(remote, ednssubnet, local, qname, qtype);
+
+    if (policyTags) {
+      boost::optional<std::unordered_map<int, std::string>> tags = std::get<1>(ret);
```
Habbie: I'm surprised this can't be `auto` - or can it?

Habbie:
I see no obvious defects in the protobuf side of things, but I did not read every line. If you have tested it, please feel free to merge.

pdns/lua-recursor4.cc
rgacogne: It can, I'm not sure which option is easier to read though.

Habbie: Alright, do what makes sense to you :)

rgacogne:
Minor update using `auto` in `RecursorLua4::gettag()`, will merge once travis is happy.

PR #4020, opened 2016-06-20
auth: Clarify comments and logs in `getAuth()`, rename `shorter` to `choppedOff`

rgacogne:
The possibility for a backend to report the best match it has
when asked for a SOA was already handled by `UeberBackend::getAuth()`
but didn't work as expected.


rgacogne:
I'm not sure we actually want to merge this as-is, opening this PR to track the discussion.

rgacogne:
The actual fix of `UeberBackend::getAuth()` is done in #4061 and the one to `DNSBackend::getSOA()` has been removed because `getSOA()` is not the right place, we should probably override `getAuth()` in backends willing to return their 'best/higher' match.

PR #4032, opened 2016-06-23
rec: Set the existing policy tags in `dq` for `{pre,post}resolve`

rgacogne:


rgacogne:
I just noticed that there was no point anymore in using a temporary vector for policy tags in `dq`, moved to a pointer instead.

PR #4044, opened 2016-06-26
Dnspacket comboaddr

cmouse:


cmouse:
Please review.

pdns/lua-auth.cc
```diff
@@ -193,13 +193,13 @@ static int ldp_addRecords(lua_State *L) {
...
   DNSPacket *p=ldp_checkDNSPacket(L);
-  lua_pushstring(L, p->getRemote().c_str());
+  lua_pushstring(L, p->getRemote().toString().c_str());
   return 1;
 }
 
 static int ldp_getRemoteRaw(lua_State *L) {
   DNSPacket *p=ldp_checkDNSPacket(L);
-  const ComboAddress& ca=p->d_remote;
+  const ComboAddress& ca=p->getRemote();
```
Habbie: why change this?

pdns/packethandler.cc
```diff
@@ -763,8 +763,8 @@ int PacketHandler::trySuperMasterSynchronous(DNSPacket *p, const DNSName& tsigke
   try {
     Resolver resolver;
     uint32_t theirserial;
-    resolver.getSoaSerial(p->getRemote(),p->qdomain, &theirserial);    
-    resolver.resolve(p->getRemote(), p->qdomain, QType::NS, &nsset);
+    resolver.getSoaSerial(p->getRemote().toString(),p->qdomain, &theirserial);    
+    resolver.resolve(p->getRemote().toString(), p->qdomain, QType::NS, &nsset);
```
Habbie: Wonder if these should eventually be taking CA as well :)

Habbie:
minor nits. LGTM.

pdns/lua-auth.cc
cmouse: it effectively returns same thing.

pdns/dnspacket.cc
```diff
@@ -81,9 +81,9 @@ const string& DNSPacket::getString()
   return d_rawpacket;
 }
 
-string DNSPacket::getRemote() const
+ComboAddress DNSPacket::getRemote() const
```
rgacogne: const ComboAddress& DNSPacket::getRemote() const, perhaps?

Habbie: Won't & give us risk of leaking if we're not careful? CA is pretty small in any case.

pdns/packethandler.cc
cmouse: Probably, but not in this change.

cmouse:
Changed lua backend too.

Habbie:
This looks good to me but there is no reason to merge it before 4.0.0, so I suggest doing it right after.

PR #4054, opened 2016-06-27
rec: Don't duplicate the policy name, use a shared_ptr instead

rgacogne:
The policy name we use to pass an `applied policy` value to our protobuf messages was duplicated for every policy object, and I just realized there could be quite a lot of them.
This commit moves to a shared_ptr instead.


rgacogne:
Obsoleted by #4324.

PR #4062, opened 2016-06-28
DNSName logging fixes

### LINKED ISSUE #3589: After #3569 may want to review DNSName logging
#3569 introduces `DNSName::toLogString`, we probably want to use this in lots of places that right now either choke on `empty` or check explicitly and verbosely.

### LINKED ISSUE #4002: replace all toString() and <<DNSName<< in logging to toLogString()
to avoid throwing on empty DNSNames

pieterlexis:
This PR removes the inconsistent use of `DNSName.toString()`, `DNSName.toStringNoDot()`, `DNSName.toStringRootDot()` and `DNSName.toLogString()` with regards to logging.

It makes the `<<` operator for DNSNames call `toLogString()`, which in turn uses `toStringRootDot()` on a non-empty DNSName. The output provided should therefor be consistent with the pre-DNSName logging, and print root always as '.'.

This PR also fixes dot-correctness to the `zone2json` tool and "root zone is a dot" semantics to the pipebackend (as is already the case with the gsql backends).

Closes #4002 
Closes #3589

### DIFF
--- a/docs/markdown/authoritative/backend-pipe.md
+++ b/docs/markdown/authoritative/backend-pipe.md
@@ -68,7 +68,8 @@ If this time is ever exceeded, the backend is declared dead and a new process is
 
 If set, only questions matching this regular expression are even sent to the backend.
 This makes sure that most of PowerDNS does not slow down if you you deploy a slow backend.
-A query for 'www.powerdns.com' would be presented to the regex as 'www.powerdns.com', a matching regex would be '^www.powerdns.com$'.
+A query for 'www.powerdns.com' would be presented to the regex as 'www.powerdns.com', a matching regex would be `^www\.powerdns\.com$`.
+**Note**: to match the root domain, use a dot, e.g. `^\.$`
 
 # PipeBackend protocol
 Questions come in over a file descriptor, by default standard input.
--- a/modules/bindbackend/bindbackend2.cc
+++ b/modules/bindbackend/bindbackend2.cc
@@ -211,7 +211,7 @@ bool Bind2Backend::startTransaction(const DNSName &qname, int id)
     }
     
     *d_of<<"; Written by PowerDNS, don't edit!"<<endl;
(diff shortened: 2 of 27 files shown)


pdns/dnsbackend.cc
```diff
@@ -357,7 +357,7 @@ string serializeSOAData(const SOAData &d)
 {
   ostringstream o;
   //  nameservername hostmaster serial-number [refresh [retry [expire [ minimum] ] ] ]
-  o<<d.nameserver.toString()<<" "<< d.hostmaster.toString() <<" "<< d.serial <<" "<< d.refresh << " "<< d.retry << " "<< d.expire << " "<< d.default_ttl;
+  o<<d.nameserver<<" "<< d.hostmaster <<" "<< d.serial <<" "<< d.refresh << " "<< d.retry << " "<< d.expire << " "<< d.default_ttl;
```
Habbie: given that ostreams now use `toLogString`, this code suddenly no longer throws while I think it should

docs/markdown/authoritative/backend-pipe.md
```diff
@@ -69,6 +69,7 @@ If this time is ever exceeded, the backend is declared dead and a new process is
 If set, only questions matching this regular expression are even sent to the backend.
 This makes sure that most of PowerDNS does not slow down if you you deploy a slow backend.
 A query for 'www.powerdns.com' would be presented to the regex as 'www.powerdns.com', a matching regex would be '^www.powerdns.com$'.
+**Note**: to match the root domain, use a dot, e.g. '^.$'
```
Habbie: I think this dot needs escaping

pdns/dnsbackend.cc
pieterlexis: This function is never wrapped directly inside a try/catch. But now that you mention it.. broken SOA data should indeed be not allowed.

docs/markdown/authoritative/backend-pipe.md
pieterlexis: yes, also the other dots above it (we do indeed use Regex here)

Habbie:
Review done. Few nits.

pieterlexis:
nits fixed

PR #4065, opened 2016-06-29
dnsdist: Include editline/readline.h instead of readline.h/history.h

rgacogne:
All libedit functions are in `<editline/readline.h`>, at least on Arch, OpenSUSE Leap and Fedora 22, and it looks like including `<readline.h`> and `<history.h`> would get non-libedit headers.

### DIFF
--- a/pdns/README-dnsdist.md
+++ b/pdns/README-dnsdist.md
@@ -30,7 +30,7 @@ make
 
 To build on OS X, `./configure LIBEDIT_LIBS='-L/usr/lib -ledit' LIBEDIT_CFLAGS=-I/usr/include/editline`
 
-To build on OpenBSD, `./configure CXX=eg++ CPP=ecpp LIBEDIT_LIBS='-ledit -lcurses' LIBEDIT_CFLAGS=-I/usr/include/readline`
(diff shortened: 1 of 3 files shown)


rgacogne:
This change was introduced to fix the compilation on OpenBSD, so we should probably test on that platform and on OSX before merging.

Habbie:
It turns out bffca8b9ff61471f97a5a25be8a85e87cb8c6535 also breaks OSX.

Habbie:
And this commit fixes OSX, but to be complete we should remove `LIBEDIT_CFLAGS` from the OSX configure line in README.

Habbie:
Meanwhile 65613131cd97b2fc738229ddad4fe8a3b484c073 (dnsdist-lua.cc) broke OpenBSD!

Habbie:
I'll continue with OpenBSD tomorrow.

Habbie:
I wonder why dnsdist.cc doesn't need this change

Habbie:
I have sent a PR to your branch that fixes dnsdist.cc in general and both files for OpenBSD. Also works on OSX with that change.

PR #4067, opened 2016-06-29
dnsdist: Disable eBPF support when BPF_FUNC_tail_call is not found

rgacogne:


rgacogne:
@pieterlexis review welcome, there might be a prettier way to check this I don't know about!

pieterlexis:
I think [`AC_CHECK_DECL`](https://www.gnu.org/software/autoconf/manual/autoconf-2.60/html_node/Generic-Declarations.html) is meant to check for functions

rgacogne:
Rewritten with `AC_CHECK_DECL`, as `AC_CHECK_FUNC` does not work for `BPF_FUNC_tail_call` since it's not a function. Thank you!

PR #4071, opened 2016-06-29
abort on backend failures at startup and retry while running

mind04:
TODO:

- [x] check bind, gmsql
- [x] double check @lieter gpgsql, pipe


zeha:
LGTM, but what happens when backends get instantiated on other threads/for other causes?

From the top of my head API should be fine (will just cause internal server error to be returned), but no idea about other callers.

pieterlexis:
this indeed works on several scenarios for the pgsql backend

pieterlexis:
Is merged onto master, but GH fails for some reason

PR #4080, opened 2016-06-30
Added metadata handlers for HTTP API

commx:
Added HTTP API handlers for zone metadata retrieval and modification. Tests were added as well.

Note: The changes were made for the 3.4.x version of the authoritative server, but can be ported to the 4.0.x version as well.

### DIFF
--- a/pdns/ws-auth.cc
+++ b/pdns/ws-auth.cc
@@ -509,6 +509,177 @@ static void updateDomainSettingsFromDocument(const DomainInfo& di, const string&
   }
 }
 
+static bool isValidMetadataKind(const string& kind, bool readonly) {
+  static string validOptions[] = {
+    "ALLOW-AXFR-FROM",
+    "AXFR-SOURCE",
+    "ALLOW-DNSUPDATE-FROM",
+    "TSIG-ALLOW-DNSUPDATE",
+    "FORWARD-DNSUPDATE",
+    "SOA-EDIT-DNSUPDATE",
+    "ALSO-NOTIFY",
+    "AXFR-MASTER-TSIG",
+    "GSS-ALLOW-AXFR-PRINCIPAL",
+    "GSS-ACCEPTOR-PRINCIPAL",
+    "IXFR",
+    "LUA-AXFR-SCRIPT",
+    "NSEC3NARROW",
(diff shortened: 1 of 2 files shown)


Habbie:
Thank you! I haven't checked out the work yet but adding this feature is awesome. However, we do not add such features to the 3.4.x branch. Can you port them to 4.x? If you want a review before you put in that effort, we can do that of course. Please let us know!

commx:
Sure, please review and let me know if I can port it.

Habbie:
@zeha your attention is welcomed here :)

zeha:
My feeling about domainmetadata was that the API should not expose that unmapped. I see the proposed code special-cases `NSEC3NARROW` etc., but my feeling is that none of the raw metadata is useful as is.

I assume you had need for something that isn't exposed today -- I'd guess designing and writing the API for just that part would also be easy but a lot more meaningful?

cmouse:
Is is possible to handle custom metadata from scripts or new features/backends without modifying HTTP API code? It's bit bothersome if you want to use some custom metadata in your script backend but can't modify it over HTTP. E.g. autorev.pl has custom metadata.

zeha:
That wouldn't work with the current PR either; and I still think it's a bad idea *in general*. I'd be "okay" with `X-<whatever>` metadata that can be read/written to in a generic way for custom scripts or such.

commx:
The implementation was made based on the description in the API specs. From another point of view the code currently prevents writes for unsupported metadata entries, but it would return the unsupported metadata entries anyway when retrieving metadata for a zone. At any time it should prevent writes for the three special-cases for reasons outlined in the API specs.

zeha:
Oh right, I even specced that!

So - in general this looks good. Note that in 4.0 the JSON library has changed, so code for 4.0 looks a bit different (and usually shorter).

I like the whitelist approach, and I think we'll need to trim the list a bit (given some of them now take filenames, and that looks like a security issue waiting to happen).

zeha:
Also please consider splitting the `test_domainmetadata` test, it's a tiny bit nicer if bugs in GET/POST/... cause different tests to fail.
Successful PUT isn't tested?

commx:
Replaced by #4093.

PR #4093, opened 2016-07-03
Added metadata handlers for HTTP API

commx:
This pull request replaces #4080.

I've also added `LUA-AXFR-SCRIPT` to the list of metadata entries which cannot be modified via API due to security concerns. As discussed previously, the implementation will reject creation and modification of custom metadata entries. We may choose from three possible options to relax the behavior:

1. Leave the implementation as-is, rejecting custom metadata entries at all
2. Accept custom metadata if the configuration option `api-allow-custom-metadata` is set to `yes`, reject otherwise
3. Accept custom metadata at all

### DIFF
--- a/pdns/ws-auth.cc
+++ b/pdns/ws-auth.cc
@@ -526,6 +526,181 @@ static void updateDomainSettingsFromDocument(const DomainInfo& di, const DNSName
   }
 }
 
+static bool isValidMetadataKind(const string& kind, bool readonly) {
+  static vector<string> builtinOptions {
+    "ALLOW-AXFR-FROM",
+    "AXFR-SOURCE",
+    "ALLOW-DNSUPDATE-FROM",
+    "TSIG-ALLOW-DNSUPDATE",
+    "FORWARD-DNSUPDATE",
+    "SOA-EDIT-DNSUPDATE",
+    "ALSO-NOTIFY",
+    "AXFR-MASTER-TSIG",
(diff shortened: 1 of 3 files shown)


pdns/ws-auth.cc
```diff
@@ -526,6 +526,181 @@ static void updateDomainSettingsFromDocument(const DomainInfo& di, const DNSName
...
+      throw ApiException("Could not update metadata entries for domain '" + zonename.toString() + "'");
+
+    Json::object key {
+      { "type", "Metadata" },
+      { "kind", document["kind"] },
+      { "metadata", metadata }
+    };
+
+    resp->status = 201;
+    resp->setBody(key);
```
zeha: If you return 201 `No Content`, there should be no body really 😄

zeha:
Small nit, can't find anything to complain about otherwise. </austrian>

Looks good!

zeha:
Oh, updating the docs would be heart-warming.

pdns/ws-auth.cc
commx: From the API spec about this:
> POST: Create. Success reply: 201 Created, with new object as body.

zeha: You are, once again, correct. (And I fail at remembering the spec.)

zeha:
IMO good to merge; if I was wrong with rgd. to response codes, this can be fixed afterwards.

Issue #4084, opened 2016-07-01
DNSPacketWriter::commit() called without startRecord ever having been called, but a record was added


mzealey:
Regression in 4.0 rc1 compared to 3-series:

`dig +noedns @10.10.7.77 nos-avg.cz. RRSIG` causes the following error in the logs:

```
DNS parser error (nos-avg.cz./RRSIG from 10.10.7.1): nos-avg.cz., DNSPacketWriter::commit() called without startRecord ever having been called, but a record was added
```

mzealey:
Doing this against a 3 series recursor returns a truncated response which causes dig to go into tcp mode for the correct query.

pieterlexis:
trace:

```
Jul 01 12:44:02 2 [1/1] question for 'nos-avg.cz.|RRSIG' from 127.0.0.1
Jul 01 12:44:02 [1] nos-avg.cz.: Wants DNSSEC processing in query for RRSIG
Jul 01 12:44:02 [1] nos-avg.cz.: Looking for CNAME cache hit of 'nos-avg.cz.|CNAME'
Jul 01 12:44:02 [1] nos-avg.cz.: No CNAME cache hit of 'nos-avg.cz.|CNAME' found
Jul 01 12:44:02 [1] nos-avg.cz.: No cache hit for 'nos-avg.cz.|RRSIG', trying to find an appropriate NS record
Jul 01 12:44:02 [1] nos-avg.cz.: Checking if we have NS in cache for 'nos-avg.cz.'
Jul 01 12:44:02 [1] nos-avg.cz.: no valid/useful NS in cache for 'nos-avg.cz.'
Jul 01 12:44:02 [1] nos-avg.cz.: Checking if we have NS in cache for 'cz.'
Jul 01 12:44:02 [1] nos-avg.cz.: no valid/useful NS in cache for 'cz.'
Jul 01 12:44:02 [1] nos-avg.cz.: Checking if we have NS in cache for '.'
Jul 01 12:44:02 [1] nos-avg.cz.: NS (with ip, or non-glue) in cache for '.' -> 'l.root-servers.net.'
Jul 01 12:44:02 [1] nos-avg.cz.: within bailiwick: 1,  in cache, ttl=86397
Jul 01 12:44:02 [1] nos-avg.cz.: NS (with ip, or non-glue) in cache for '.' -> 'e.root-servers.net.'
[...]

PR #4096, opened 2016-07-04
rec: Don't call `commit()` if we skipped all the records

rgacogne:
This should fix the "DNSPacketWriter::commit() called without startRecord ever having been called, but a record was added" part from #4084.


Habbie:
Closing this in favour of #4119 which includes it.

Habbie:
After talking to @mind04, merging this and awaiting more work from him for after 4.0.0.

PR #4097, opened 2016-07-04
Recursor: Allow logging DNSSEC bogus in any mode

pieterlexis:
Also allow setting this at runtime.

### DIFF
--- a/pdns/pdns_recursor.cc
+++ b/pdns/pdns_recursor.cc
@@ -975,7 +975,7 @@ void startDoResolve(void *p)
             pw.getHeader()->ad=0;
           }
           else if(state == Bogus) {
-            if(sr.doLog() || g_dnssecmode == DNSSECMode::ValidateForLog) {
+            if(g_dnssecLogBogus || sr.doLog() || g_dnssecmode == DNSSECMode::ValidateForLog) {
               L<<Logger::Warning<<"Answer to "<<dc->d_mdp.d_qname<<" for "<<dc->d_remote.toStringWithPort()<<" validates as Bogus"<<endl;
             }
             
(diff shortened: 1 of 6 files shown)


pdns/pdns_recursor.cc
```diff
@@ -975,7 +975,7 @@ void startDoResolve(void *p)
             pw.getHeader()->ad=0;
           }
           else if(state == Bogus) {
-            if(sr.doLog() || g_dnssecmode == DNSSECMode::ValidateForLog) {
+            if(::arg().mustDo("dnssec-log-bogus") || sr.doLog() || g_dnssecmode == DNSSECMode::ValidateForLog) {
```
Habbie: `::arg` is slow (that's why the the dnssec mode is in a `g_`)

pdns/rec_channel_rec.cc
```diff
@@ -343,6 +345,33 @@ string doSetCarbonServer(T begin, T end)
 }
 
 template<typename T>
+string doSetDnssecLogBogus(T begin, T end)
+{
+  if (begin == end)
+    return "No DNSSEC Bogus logging setting specified\n";
+
+  if (pdns_iequals(*begin, "on") || pdns_iequals(*begin, "yes")) {
+    if (!arg().mustDo("dnssec-log-bogus")) {
```
Habbie: Because `doSetDnssecLogBogus` does not update `arg`, this code might do the wrong thing when it is called more than once

pdns/rec_channel_rec.cc
```diff
@@ -343,6 +345,33 @@ string doSetCarbonServer(T begin, T end)
...
+      g_dnssecLogBogus = false;
+      return "DNSSEC Bogus logging disabled\n";
+    }
+    return "DNSSEC Bogus logging was already disabled\n";
+  }
+
+  return "Unknown DNSSEC Bogus setting: '" + *begin +"'\n";
+}
+
+template<typename T>
```
Habbie: Various parts of this feel like they might be duplicating other work.

pieterlexis: possibly, not an issue at this point imho.

PR #4106, opened 2016-07-05
Implementation of the cryptokeys REST-API (PUT, POST, DELETE)

### LINKED ISSUE #706: pdnssec add-zone-key and import-zone-key does not report key-id of added key
pdnssec add-zone-key and import-zone-key does not print key-id of recently added key, so it is quite difficult to do any further automatic processing concerning this key (for example deactivate key added with import-zone-key)

printing key-id would made simple scripting much more simple :-)

benj-zen:
We are a group of students at the Freie Universität Berlin. Our team consists of @zervnet, @MrM0nkey, @taudor and @benj-zen.

We implemented the Rest-Api methods for the cryptokeys module. While doing that we realized we need to fix #706, so we can return the id of the added key e.g. using POST. As proposed in #706 pdnsutil prints the added key-id to stdout.

As suggested in the discussion in the IRC channel, we only implemented the (return-id-)functionality for the major backend-modules (gmysql, gpgsql, gsqlite3, godbc, bind).

Documentation and tests are included.

### DIFF
--- a/docs/markdown/httpapi/api_spec.md
+++ b/docs/markdown/httpapi/api_spec.md
@@ -674,7 +674,7 @@ CryptoKeys
 cryptokey\_resource
 -------------------
 
-    {[
+    {
       "type": "CryptoKey",
       "id": <int>,
       "active": <bool>,
@@ -684,7 +684,7 @@ cryptokey\_resource
       "ds": [ <ds>,
               <ds>,
               .... ]
-    ]}
+    }
 
 
 ##### Parameters:
@@ -711,20 +711,42 @@ Returns all public data about cryptokeys, but not `privatekey`.
 
 #### POST
 
-Creates a new, single cryptokey.
-
-**TODO**: Not yet implemented.
+This method adds a key to a zone by generate it or content parameter.
 
 ##### Parameters:
 
-`content`: if `null`, the server generates a new key. In this case, the
+* `content` : "key The format used is compatible with BIND and NSD/LDNS" `<string>`
+
+If `content` == `null`, the server generates a new key. In this case, the
 following additional fields MAY be supplied:
 
-* `bits`: `<int>`
+* `bits`: number of bits `<int>`
 * `algo`: `<algo>`
+* `keytype` : "ksk|zsk" `<string>`
+* `active`: "true|false" `<value>`
 
 Where `<algo>` is one of the supported key algorithms in lowercase OR the
 numeric id, see [`the list`](../authoritative/dnssec.md#supported-algorithms).
+
+##### Response:
+* `422 Unprocessable Entity`:
+    * keytype isn't ksk|zsk:
+        * `json` {"error" : "Invalid keytype 'keytype'"}
+    * The "algo" isn't supported:
+        * `json` {"error" : "Unknown algorithm: 'algo'"}
+    * Algo <= 10 and no bits were passed:
+        * `json` {"error" : "Creating an algorithm algo key requires the size (in bits) to be passed"}
+    * The wrong keysize was passed:
+        * `json` {"error" : "Wrong bit size!"}
+    * If the server cant guess the keysize:
+        * `json` {"error" : "Can't guess key size for algorithm"}
+    * The key-creation failed:
+        * `json` {"error" : "Adding key failed, perhaps DNSSEC not enabled in configuration?"}
+    * The key in content has the wrong format:
+        * `json` {"error" : "Wrong key format!"}
+* `201 OK`:
+    * Everything was fine:
+        * `json` all public data about the new cryptokey. Look at cryptokey\_resource.
 
 URL: /api/v1/servers/:server\_id/zones/:zone\_name/cryptokeys/:cryptokey\_id
 ----------------------------------------------------------------------------
@@ -737,11 +759,37 @@ Returns all public data about cryptokeys, including `privatekey`.
 
 #### PUT
 
-**TODO**: Not yet implemented.
+This method handles PUT (execute) de/activates a key from `zone_name` specified by `cryptokey_id`.
+
+##### Parameters:
+
+* `:zone_name`: name of the zone for which the key with `cryptokey_id` should be de/activated
+* `cryptokey_id`: id of the key which wanted to be de/activated
+* `json`: {"active": true|false}
+
+##### Responses:
+* `200 OK`: The key with `cryptokey_id` is de/activated.
+* `400 Bad Request`: The `zone_name` is not found.
+* `422 Unprocessable Entity`:
+    &nbsp;&nbsp;The backend returns false on de/activation. An error occoured.
+    &nbsp;&nbsp;json {"error": "Could not de/activate Key: :cryptokey_id in Zone: :zone_name"}
 
 #### DELETE
 
-**TODO**: Not yet implemented.
+This Method deletes a key from a zone.
+
+##### Parameters:
+
+* `:zone_name`: name of the zone which is signed with a key with `cryptokey_id`
+* `cryptokey_id`: id of the key which wanted to be gone
+
+##### Responses:
+
+* `200 No Content`: The Key is gone.
+* `400 Bad Request`: The `zone_name` is not found.
+* `422 Unknown Status`:
+    &nbsp;&nbsp;The backend failed to remove the key.
+    &nbsp;&nbsp;json {"error": Could not DELETE :cryptokey_id"}
 
 Data searching
 ==============
--- a/docs/markdown/httpapi/api_spec.md
+++ b/docs/markdown/httpapi/api_spec.md
@@ -668,14 +668,14 @@ Modifies the metadata entries of a given kind for the zone.
 This returns `200 OK` on success.
 
 
-CryptoKeys
+Cryptokeys
 ==========
 
 cryptokey\_resource
 -------------------
 
     {
-      "type": "CryptoKey",
+      "type": "Cryptokey",
       "id": <int>,
       "active": <bool>,
       "keytype": <keytype>,
@@ -711,42 +711,44 @@ Returns all public data about cryptokeys, but not `privatekey`.
 
 #### POST
 
-This method adds a key to a zone by generate it or content parameter.
+This method adds a new key to a zone. The key can either be generated or imported by supplying the content parameter.
 
 ##### Parameters:
 
-* `content` : "key The format used is compatible with BIND and NSD/LDNS" `<string>`
+* `content` : "\<key\>" `<string>` (The format used is compatible with BIND and NSD/LDNS)
+* `keytype` : "ksk|zsk" `<string>`
+* `active`: "true|false" `<value>` (If not set the key will not be active by default)
 
 If `content` == `null`, the server generates a new key. In this case, the
 following additional fields MAY be supplied:
 
 * `bits`: number of bits `<int>`
-* `algo`: `<algo>`
-* `keytype` : "ksk|zsk" `<string>`
-* `active`: "true|false" `<value>`
+* `algo`: `<algo>` (Default: 13/ECDSA)
 
 Where `<algo>` is one of the supported key algorithms in lowercase OR the
 numeric id, see [`the list`](../authoritative/dnssec.md#supported-algorithms).
 
 ##### Response:
 * `422 Unprocessable Entity`:
-    * keytype isn't ksk|zsk:
-        * `json` {"error" : "Invalid keytype 'keytype'"}
-    * The "algo" isn't supported:
-        * `json` {"error" : "Unknown algorithm: 'algo'"}
-    * Algo <= 10 and no bits were passed:
-        * `json` {"error" : "Creating an algorithm algo key requires the size (in bits) to be passed"}
-    * The wrong keysize was passed:
-        * `json` {"error" : "Wrong bit size!"}
-    * If the server cant guess the keysize:
-        * `json` {"error" : "Can't guess key size for algorithm"}
+    * keytype is not ksk|zsk:
+        * `{"error" : "Invalid keytype 'keytype'"}`
+    * The "algo" is not supported:
+        * `{"error" : "Unknown algorithm: 'algo'"}`
+    * Algo <= 10 and the `bits` parameter is not set:
+        * `{"error" : "Creating an algorithm 'algo' key requires the size (in bits) to be passed."}`
+    * The provided bit size is not supported by the selected algorithm:
+        * `{"error" : "The algorithm does not support the given bit size."}`
+    * The `bits` parameter is not a positive integer value:
+        * `{"error" : "'bits' must be a positive integer value"}`
+    * If the server can not guess the key size:
+        * `{"error" : "Can not guess key size for algorithm"}`
     * The key-creation failed:
-        * `json` {"error" : "Adding key failed, perhaps DNSSEC not enabled in configuration?"}
-    * The key in content has the wrong format:
-        * `json` {"error" : "Wrong key format!"}
-* `201 OK`:
+        * `{"error" : "Adding key failed, perhaps DNSSEC not enabled in configuration?"}`
+    * The key in `content` has the wrong format:
+        * `{"error" : "Key could not be parsed. Make sure your key format is correct."}`
+* `201 Created`:
     * Everything was fine:
-        * `json` all public data about the new cryptokey. Look at cryptokey\_resource.
+        * Returns all public data about the new cryptokey. Look at cryptokey\_resource.
 
 URL: /api/v1/servers/:server\_id/zones/:zone\_name/cryptokeys/:cryptokey\_id
 ----------------------------------------------------------------------------
@@ -759,37 +761,28 @@ Returns all public data about cryptokeys, including `privatekey`.
 
 #### PUT
 
-This method handles PUT (execute) de/activates a key from `zone_name` specified by `cryptokey_id`.
+This method de/activates a key from `zone_name` specified by `cryptokey_id`.
 
 ##### Parameters:
 
-* `:zone_name`: name of the zone for which the key with `cryptokey_id` should be de/activated
-* `cryptokey_id`: id of the key which wanted to be de/activated
-* `json`: {"active": true|false}
+* `active`: "true|false" `<value>`
 
 ##### Responses:
-* `200 OK`: The key with `cryptokey_id` is de/activated.
-* `400 Bad Request`: The `zone_name` is not found.
+* `204 No Content`: The key with `cryptokey_id` is de/activated.
 * `422 Unprocessable Entity`:
-    &nbsp;&nbsp;The backend returns false on de/activation. An error occoured.
-    &nbsp;&nbsp;json {"error": "Could not de/activate Key: :cryptokey_id in Zone: :zone_name"}
+    &nbsp;&nbsp;The backend returns false on de/activation. An error occurred.
+    &nbsp;&nbsp;`{"error": "Could not de/activate Key: :cryptokey_id in Zone: :zone_name"}`
 
 #### DELETE
 
-This Method deletes a key from a zone.
-
-##### Parameters:
-
-* `:zone_name`: name of the zone which is signed with a key with `cryptokey_id`
-* `cryptokey_id`: id of the key which wanted to be gone
+This method deletes a key from `zone_name` specified by `cryptokey_id`.
 
 ##### Responses:
 
-* `200 No Content`: The Key is gone.
-* `400 Bad Request`: The `zone_name` is not found.
-* `422 Unknown Status`:
+* `200 OK`: The Key is gone.
+* `422 Unprocessable Entity`:
     &nbsp;&nbsp;The backend failed to remove the key.
-    &nbsp;&nbsp;json {"error": Could not DELETE :cryptokey_id"}
+    &nbsp;&nbsp;`{"error": Could not DELETE :cryptokey_id"}`
 
 Data searching
 ==============
--- a/modules/bindbackend/binddnssec.cc
+++ b/modules/bindbackend/binddnssec.cc
@@ -114,6 +114,7 @@ void Bind2Backend::setupStatements()
   d_getDomainKeysQuery_stmt = d_dnssecdb->prepare("select id,flags, active, content from cryptokeys where domain=:domain",1);
   d_deleteDomainKeyQuery_stmt = d_dnssecdb->prepare("delete from cryptokeys where domain=:domain and id=:key_id",2);
   d_insertDomainKeyQuery_stmt = d_dnssecdb->prepare("insert into cryptokeys (domain, flags, active, content) values (:domain, :flags, :active, :content)", 4);
+  d_GetLastInsertedKeyIdQuery_stmt = d_dnssecdb->prepare("select last_insert_rowid()", 0);
   d_activateDomainKeyQuery_stmt = d_dnssecdb->prepare("update cryptokeys set active=1 where domain=:domain and id=:key_id", 2);
   d_deactivateDomainKeyQuery_stmt = d_dnssecdb->prepare("update cryptokeys set active=0 where domain=:domain and id=:key_id", 2);
   d_getTSIGKeyQuery_stmt = d_dnssecdb->prepare("select algorithm, secret from tsigkeys where name=:key_name", 1);
@@ -136,6 +137,7 @@ void Bind2Backend::freeStatements()
   release(&d_getDomainKeysQuery_stmt);
   release(&d_deleteDomainKeyQuery_stmt);
   release(&d_insertDomainKeyQuery_stmt);
+  release(&d_GetLastInsertedKeyIdQuery_stmt);
   release(&d_activateDomainKeyQuery_stmt);
   release(&d_deactivateDomainKeyQuery_stmt);
   release(&d_getTSIGKeyQuery_stmt);
@@ -323,7 +325,22 @@ int Bind2Backend::addDomainKey(const DNSName& name, const KeyData& key)
   catch(SSqlException& se) {
     throw PDNSException("Error accessing DNSSEC database in BIND backend, addDomainKey(): "+se.txtReason());
   }
-  return true;
+
+  try {
+    d_GetLastInsertedKeyIdQuery_stmt->execute();
+    if (!d_GetLastInsertedKeyIdQuery_stmt->hasNextRow())
+      throw PDNSException("GSQLBackend unable to get id");
+    SSqlStatement::row_t row;
+    d_GetLastInsertedKeyIdQuery_stmt->nextRow(row);
+    int id = std::stoi(row[0]);
+    d_GetLastInsertedKeyIdQuery_stmt->reset();
+    return id;
+  }
+  catch (SSqlException &e) {
+    throw PDNSException("DNSSEC database in BIND backend unable to get id: "+e.txtReason());
+  }
+
+  return -1;
 }
 
 bool Bind2Backend::activateDomainKey(const DNSName& name, unsigned int id)
--- a/modules/bindbackend/binddnssec.cc
+++ b/modules/bindbackend/binddnssec.cc
@@ -89,6 +89,8 @@ void Bind2Backend::freeStatements()
 #include "pdns/logger.hh"
 #include "pdns/ssqlite3.hh"
 
+#define ASSERT_ROW_COLUMNS(query, row, num) { if (row.size() != num) { throw PDNSException(std::string(query) + " returned wrong number of columns, expected "  #num  ", got " + std::to_string(row.size())); } }
+
 void Bind2Backend::setupDNSSEC()
 {
   if(getArg("dnssec-db").empty() || d_hybrid)
@@ -334,6 +336,7 @@ bool Bind2Backend::addDomainKey(const DNSName& name, const KeyData& key, int64_t
     }
     SSqlStatement::row_t row;
     d_GetLastInsertedKeyIdQuery_stmt->nextRow(row);
+    ASSERT_ROW_COLUMNS("get-last-inserted-key-id-query", row, 1);
     int id = std::stoi(row[0]);
     d_GetLastInsertedKeyIdQuery_stmt->reset();
     return true;
--- a/modules/bindbackend/binddnssec.cc
+++ b/modules/bindbackend/binddnssec.cc
@@ -56,7 +56,7 @@ bool Bind2Backend::getDomainKeys(const DNSName& name, unsigned int kind, std::ve
 bool Bind2Backend::removeDomainKey(const DNSName& name, unsigned int id)
 { return false; }
 
-int Bind2Backend::addDomainKey(const DNSName& name, const KeyData& key)
+bool Bind2Backend::addDomainKey(const DNSName& name, const KeyData& key, int64_t& id)
 { return -1; }
 
 bool Bind2Backend::activateDomainKey(const DNSName& name, unsigned int id)
@@ -308,10 +308,10 @@ bool Bind2Backend::removeDomainKey(const DNSName& name, unsigned int id)
   return true;
 }
 
-int Bind2Backend::addDomainKey(const DNSName& name, const KeyData& key)
+bool Bind2Backend::addDomainKey(const DNSName& name, const KeyData& key, int64_t& id)
 {
   if(!d_dnssecdb || d_hybrid)
-    return -1;
+    return false;
 
   try {
     d_insertDomainKeyQuery_stmt->
@@ -328,19 +328,22 @@ int Bind2Backend::addDomainKey(const DNSName& name, const KeyData& key)
 
   try {
     d_GetLastInsertedKeyIdQuery_stmt->execute();
-    if (!d_GetLastInsertedKeyIdQuery_stmt->hasNextRow())
-      throw PDNSException("GSQLBackend unable to get id");
+    if (!d_GetLastInsertedKeyIdQuery_stmt->hasNextRow()) {
+      id = -2;
+      return true;
+    }
     SSqlStatement::row_t row;
     d_GetLastInsertedKeyIdQuery_stmt->nextRow(row);
     int id = std::stoi(row[0]);
     d_GetLastInsertedKeyIdQuery_stmt->reset();
-    return id;
+    return true;
   }
   catch (SSqlException &e) {
-    throw PDNSException("DNSSEC database in BIND backend unable to get id: "+e.txtReason());
+    id = -2;
+    return true;
   }
 
-  return -1;
+  return false;
 }
 
 bool Bind2Backend::activateDomainKey(const DNSName& name, unsigned int id)
--- a/modules/gpgsqlbackend/gpgsqlbackend.cc
+++ b/modules/gpgsqlbackend/gpgsqlbackend.cc
@@ -118,6 +118,7 @@ public:
     declare(suffix,"delete-names-query","","delete from records where domain_id=$1 and name=$2");
 
     declare(suffix,"add-domain-key-query","", "insert into cryptokeys (domain_id, flags, active, content) select id, $1, $2, $3 from domains where name=$4");
+    declare(suffix,"get-last-inserted-key-id-query","", "select lastval()");
     declare(suffix,"list-domain-keys-query","", "select cryptokeys.id, flags, case when active then 1 else 0 end as active, content from domains, cryptokeys where cryptokeys.domain_id=domains.id and name=$1");
     declare(suffix,"get-all-domain-metadata-query","", "select kind,content from domains, domainmetadata where domainmetadata.domain_id=domains.id and name=$1");
     declare(suffix,"get-domain-metadata-query","", "select content from domains, domainmetadata where domainmetadata.domain_id=domains.id and name=$1 and domainmetadata.kind=$2");
--- a/pdns/backends/gsql/gsqlbackend.cc
+++ b/pdns/backends/gsql/gsqlbackend.cc
@@ -671,6 +671,7 @@ bool GSQLBackend::addDomainKey(const DNSName& name, const KeyData& key, int64_t&
     }
     SSqlStatement::row_t row;
     d_GetLastInsertedKeyIdQuery_stmt->nextRow(row);
+    ASSERT_ROW_COLUMNS("get-last-inserted-key-id-query", row, 1);
     id = std::stoi(row[0]);
     d_GetLastInsertedKeyIdQuery_stmt->reset();
     return true;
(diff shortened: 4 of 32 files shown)


docs/markdown/httpapi/api_spec.md
```diff
@@ -707,11 +729,37 @@ Returns all public data about cryptokeys, including `privatekey`.
...
+This Method deletes a key from a zone.
+
+##### Parameters:
+
+* `:zone_name`: name of the zone which is signed with a key with `cryptokey_id`
+* `cryptokey_id`: id of the key which wanted to be gone
+
+##### Responses:
+
+* `200 No Content`: The Key is gone.
```
pieterlexis: small nit: `200 OK` (according to the code as well)

zeha: That should probably be 204 No Content then.

pieterlexis:
small doc nit: code is shown in normal font:

img

docs/markdown/httpapi/api_spec.md
zeha: `200 OK` would be consistent with the other API parts.

benj-zen:
@pieterlexis Should we remove the 'json' and instead format the actual json-object as code?

pdns/dnsseckeeper.hh
```diff
@@ -162,8 +162,8 @@ public:
   keyset_t getEntryPoints(const DNSName& zname);
   keyset_t getKeys(const DNSName& zone, bool useCache = true);
   DNSSECPrivateKey getKeyById(const DNSName& zone, unsigned int id);
-  bool addKey(const DNSName& zname, bool setSEPBit, int algorithm, int bits=0, bool active=true);
-  bool addKey(const DNSName& zname, const DNSSECPrivateKey& dpk, bool active=true);
+  int addKey(const DNSName& zname, bool setSEPBit, int algorithm, int bits=0, bool active=true);
+  int addKey(const DNSName& zname, const DNSSECPrivateKey& dpk, bool active=true);
```
rgacogne: `addKey()` returns a signed int, because it returns -1 in case of error, but the other `*Key()` functions take an unsigned int, and `LAST_INSERT_ID()` for example returns a `BIGINT UNSIGNED (64-bit)`. Perhaps addKey should at least return an int64_t, or a boolean to indicate success with an additional unsigned int& parameter to return the key?

docs/markdown/httpapi/api_spec.md
benj-zen: Yeah, that's a typo. `200 OK` is right.

pieterlexis:
@benj-zen yes :+1:

pdns/dnsseckeeper.hh
zeha: Returning the key id in `unsigned int&` would look better to me as well

pdns/pdnsutil.cc
```diff
@@ -2229,14 +2229,16 @@ loadMainConfig(g_vm["config-dir"].as<string>());
...
+    int id;
+    if((id = dk.addKey(zone, keyOrZone, algorithm, bits, active)) < 0) {
       cerr<<"Adding key failed, perhaps DNSSEC not enabled in configuration?"<<endl;
       exit(1);
     }
     else {
       cerr<<"Added a " << (keyOrZone ? "KSK" : "ZSK")<<" with algorithm = "<<algorithm<<", active="<<active<<endl;
       if(bits)
         cerr<<"Requested specific key size of "<<bits<<" bits"<<endl;
+      cout<<std::to_string(id)<<endl;
```
zeha: if you print the key id, please also print what you're printing here...

zeha:
Suggestion: use `INSERT INTO ... RETURNING ` for DBs that support it, and use `INSERT INTO; SELECT last_insert_id()` for those that don't. (If submitting multiple statements works?)

modules/gpgsqlbackend/gpgsqlbackend.cc
```diff
@@ -97,6 +97,7 @@ class gPgSQLFactory : public BackendFactory
     declare(suffix,"delete-names-query","","delete from records where domain_id=$1 and name=$2");
 
     declare(suffix,"add-domain-key-query","", "insert into cryptokeys (domain_id, flags, active, content) select id, $1, $2, $3 from domains where name=$4");
+    declare(suffix,"get-last-inserted-key-id-query","", "select lastval()");
```
zeha: this could take the name of the sequence, similar to oracle

pdns/dbdnsseckeeper.cc
```diff
@@ -91,7 +91,11 @@ int DNSSECKeeper::addKey(const DNSName& name, bool setSEPBit, int algorithm, int
   }
   DNSSECPrivateKey dspk;
   shared_ptr<DNSCryptoKeyEngine> dpk(DNSCryptoKeyEngine::make(algorithm));
-  dpk->create(bits);
+  try{
+    dpk->create(bits);
+  } catch (std::runtime_error error){
+    throw runtime_error("Wrong bit size!");
```
zeha: please use a more specific exception, and i think the error message could be a lot clearer (`Unsupported key size`?)

pdns/ws-auth.cc
```diff
@@ -561,20 +550,20 @@ static void apiZoneCryptokeys(HttpRequest* req, HttpResponse* resp) {
     }
 
     Json::object key {
-      { "type", "Cryptokey" },
-      { "id", (int)value.second.id },
-      { "active", value.second.active },
-      { "keytype", keyType },
-      { "flags", (uint16_t)value.first.d_flags },
-      { "dnskey", value.first.getDNSKEY().getZoneRepresentation() }
+        { "type", "Cryptokey" },
```
zeha: whitespace noise

pdns/ws-auth.cc
```diff
@@ -561,20 +550,20 @@ static void apiZoneCryptokeys(HttpRequest* req, HttpResponse* resp) {
...
+        { "dnskey", value.first.getDNSKEY().getZoneRepresentation() }
     };
 
     if (value.second.keyType == DNSSECKeeper::KSK || value.second.keyType == DNSSECKeeper::CSK) {
       Json::array dses;
       for(const int keyid : { 1, 2, 3, 4 })
-      try {
-        dses.push_back(makeDSFromDNSKey(zonename, value.first.getDNSKEY(), keyid).getZoneRepresentation());
-      } catch (...) {}
+        try {
```
zeha: wrong indent

pdns/ws-auth.cc
```diff
@@ -591,6 +580,208 @@ static void apiZoneCryptokeys(HttpRequest* req, HttpResponse* resp) {
...
   resp->setBody(doc);
+
+}
+
+/*
+ * This method handles DELETE requests for URL /api/v1/servers/:server_id/zones/:zone_name/cryptokeys/:cryptokey_id .
+ * It deletes a key from :zone_name specified by :cryptokey_id.
+ * Server Answers:
+ * Case 1: zone_name not found
+ *      The server returns 400 Bad Request
```
zeha: GET returns HttpNotFoundException in this case, and I think everything should

pdns/ws-auth.cc
```diff
@@ -591,6 +580,208 @@ static void apiZoneCryptokeys(HttpRequest* req, HttpResponse* resp) {
...
+ *      The sever returns 422 Unprocessable Entity with message "Could not DELETE :cryptokey_id"
+ * */
+static void apiZoneCryptokeysDELETE(DNSName zonename, int inquireKeyId, HttpRequest *req, HttpResponse *resp) {
+  if (!req->parameters.count("key_id"))
+    throw HttpBadRequestException();
+  UeberBackend B;
+  DNSSECKeeper dk(&B);
+  DomainInfo di;
+  if (!B.getDomainInfo(zonename, di))
+    throw HttpBadRequestException();
```
zeha: a lot of this code could be shared with GET/PUT/...

pdns/ws-auth.cc
```diff
@@ -591,6 +580,208 @@ static void apiZoneCryptokeys(HttpRequest* req, HttpResponse* resp) {
...
+ * Server Answers:
+ * Case 1: zone_name not found
+ *      The server returns 400 Bad Request
+ * Case 2: the backend returns true on removal. This means the key is gone.
+ *      The server returns 200 No Content
+ * Case 3: the backend returns false on removal. An error occoured.
+ *      The sever returns 422 Unprocessable Entity with message "Could not DELETE :cryptokey_id"
+ * */
+static void apiZoneCryptokeysDELETE(DNSName zonename, int inquireKeyId, HttpRequest *req, HttpResponse *resp) {
+  if (!req->parameters.count("key_id"))
```
zeha: why do you have `inquireKeyId` in the param list and then check for `key_id` in req->parameters too?

pdns/ws-auth.cc
```diff
@@ -591,6 +580,208 @@ static void apiZoneCryptokeys(HttpRequest* req, HttpResponse* resp) {
...
+    throw HttpBadRequestException();
+  UeberBackend B;
+  DNSSECKeeper dk(&B);
+  DomainInfo di;
+  if (!B.getDomainInfo(zonename, di))
+    throw HttpBadRequestException();
+  if (dk.removeKey(zonename, inquireKeyId)) {
+    resp->setSuccessResult("OK", 200);
+  } else {
+    resp->setErrorResult("Could not DELETE " + req->parameters["key_id"], 422);
```
zeha: feels like this should be an internal server error, OR have an actually useful error message?

pdns/ws-auth.cc
```diff
@@ -591,6 +580,208 @@ static void apiZoneCryptokeys(HttpRequest* req, HttpResponse* resp) {
...
+      resp->setErrorResult("Could not activate Key: " + req->parameters["key_id"] + " in Zone: " + zonename.toString(), 422);
+      return;
+    }
+  } else {
+    if (!dk.deactivateKey(zonename, inquireKeyId)) {
+      resp->setErrorResult("Could not deactivate Key: " + req->parameters["key_id"] + " in Zone: " + zonename.toString(), 422);
+      return;
+    }
+  }
+  resp->setSuccessResult("OK", 200);
```
zeha: this should be a 204 with no body

pdns/ws-auth.cc
```diff
@@ -591,6 +580,208 @@ static void apiZoneCryptokeys(HttpRequest* req, HttpResponse* resp) {
...
+static void apiZoneCryptokeysDELETE(DNSName zonename, int inquireKeyId, HttpRequest *req, HttpResponse *resp) {
+  if (!req->parameters.count("key_id"))
+    throw HttpBadRequestException();
+  UeberBackend B;
+  DNSSECKeeper dk(&B);
+  DomainInfo di;
+  if (!B.getDomainInfo(zonename, di))
+    throw HttpBadRequestException();
+  if (dk.removeKey(zonename, inquireKeyId)) {
+    resp->setSuccessResult("OK", 200);
```
zeha: The OK here ends up in the body, that's pointless. please change to 204 and no body (compare DELETE for zones)

pdns/ws-auth.cc
```diff
@@ -591,6 +580,208 @@ static void apiZoneCryptokeys(HttpRequest* req, HttpResponse* resp) {
...
+ *  Case 1: keytype isn't ksk|zsk
+ *    The server returns 422 Unprocessable Entity {"error" : "Invalid keytype 'keytype'"}
+ *  Case 2: The "algo" isn't supported
+ *    The server returns 422 Unprocessable Entity {"error" : "Unknown algorithm: 'algo'"}
+ *  Case 3: Algorithm <= 10 and no bits were passed
+ *    The server returns 422 Unprocessable Entity {"error" : "Creating an algorithm algo key requires the size (in bits) to be passed"}
+ *  Case 4: The wrong keysize was passed
+ *    The server returns 422 Unprocessable Entity {"error" : "Wrong bit size!"}
+ *  Case 5: If the server cant guess the keysize
+ *    The server returns 422 Unprocessable Entity {"error" : "Can't guess key size for algorithm"}
```
zeha: shortened forms ("can't", ...) should not be used in error messages

pdns/ws-auth.cc
```diff
@@ -591,6 +580,208 @@ static void apiZoneCryptokeys(HttpRequest* req, HttpResponse* resp) {
...
+ *  Case 6: The key-creation failed
+ *    The server returns 422 Unprocessable Entity {"error" : "Adding key failed, perhaps DNSSEC not enabled in configuration?"}
+ *  Case 7: The key in content has the wrong format
+ *    The server returns 422 Unprocessable Entity {"error" : "Wrong key format!"}
```
zeha: no exclamation mark please

pdns/ws-auth.cc
```diff
@@ -591,6 +580,208 @@ static void apiZoneCryptokeys(HttpRequest* req, HttpResponse* resp) {
...
+ *    The server returns 201 OK and all public data about the new cryptokey
+ *  Case 8: With specified content
+ *    The server returns 201 OK and all public data about the added cryptokey
+ */
+
+static void apiZoneCryptokeysPOST(DNSName zonename, HttpRequest *req, HttpResponse *resp) {
+  auto document = req->json();
+  auto content = document["content"];
+
+  bool active = boolFromJson(document, "active", false);
```
zeha: Q: do you really want a default here?

pdns/ws-auth.cc
```diff
@@ -591,6 +580,208 @@ static void apiZoneCryptokeys(HttpRequest* req, HttpResponse* resp) {
...
+
+  bool keyOrZone;
+  if (stringFromJson(document, "keytype") == "ksk") {
+    keyOrZone = true;
+  } else if (stringFromJson(document, "keytype") == "zsk") {
```
zeha: this lacks for combined signing keys I guess? @Habbie please chime in here

pdns/ws-auth.cc
```diff
@@ -591,6 +580,208 @@ static void apiZoneCryptokeys(HttpRequest* req, HttpResponse* resp) {
...
+    try{
+      shared_ptr<DNSCryptoKeyEngine> dke(DNSCryptoKeyEngine::makeFromISCString(dkrc, keyData));
+      dpk.d_algorithm = dkrc.d_algorithm;
+      if(dpk.d_algorithm == 7)
+        dpk.d_algorithm = 5;
+
+      if (keyOrZone)
+        dpk.d_flags = 257;
+      else
+        dpk.d_flags = 256;
```
zeha: IIRC there's nicer code for this in current pdnsutil, could you check?

pdns/ws-auth.cc
```diff
@@ -591,6 +580,208 @@ static void apiZoneCryptokeys(HttpRequest* req, HttpResponse* resp) {
...
+  UeberBackend B;
+  DNSSECKeeper dk(&B);
+
+  int insertedId;
+  if (content.is_null()) {
+    int bits = intFromJson(document, "bits", 0);
+    int algorithm = 13; // ecdsa256
+    auto providedAlgo = document["algo"];
+    if (providedAlgo.is_string()) {
+      int tmpAlgo = DNSSECKeeper::shorthand2algorithm(providedAlgo.string_value());
```
zeha: just assign into `algorithm` here?

pdns/ws-auth.cc
```diff
@@ -591,6 +580,208 @@ static void apiZoneCryptokeys(HttpRequest* req, HttpResponse* resp) {
...
+      if (tmpAlgo == -1)
+        throw ApiException("Unknown algorithm: " + providedAlgo.string_value());
+      algorithm = tmpAlgo;
+    } else if (providedAlgo.is_number()) {
+      algorithm = providedAlgo.int_value();
+    }
+
+    try{
```
zeha: missing space before brace

pdns/ws-auth.cc
```diff
@@ -591,6 +580,208 @@ static void apiZoneCryptokeys(HttpRequest* req, HttpResponse* resp) {
...
+      insertedId = dk.addKey(zonename, keyOrZone, algorithm, bits, active);
+    }catch (std::runtime_error& error){
+      throw ApiException(error.what());
+    }
+
+    if (insertedId < 0)
+      throw ApiException("Adding key failed, perhaps DNSSEC not enabled in configuration?");
```
zeha: can we find out if this is the case?

zeha: even better: can addKey just throw in this case?

pdns/ws-auth.cc
```diff
@@ -591,6 +580,208 @@ static void apiZoneCryptokeys(HttpRequest* req, HttpResponse* resp) {
...
+      if(dpk.d_algorithm == 7)
+        dpk.d_algorithm = 5;
+
+      if (keyOrZone)
+        dpk.d_flags = 257;
+      else
+        dpk.d_flags = 256;
+
+      dpk.setKey(dke);
+    } catch (std::runtime_error error){
```
zeha: missing &?

pdns/ws-auth.cc
```diff
@@ -591,6 +580,208 @@ static void apiZoneCryptokeys(HttpRequest* req, HttpResponse* resp) {
...
+      throw ApiException("Wrong key format!");
```
zeha: do we have a better error message?

pdns/ws-auth.cc
```diff
@@ -591,6 +580,208 @@ static void apiZoneCryptokeys(HttpRequest* req, HttpResponse* resp) {
...
+    int algorithm = 13; // ecdsa256
+    auto providedAlgo = document["algo"];
+    if (providedAlgo.is_string()) {
+      int tmpAlgo = DNSSECKeeper::shorthand2algorithm(providedAlgo.string_value());
+      if (tmpAlgo == -1)
+        throw ApiException("Unknown algorithm: " + providedAlgo.string_value());
+      algorithm = tmpAlgo;
+    } else if (providedAlgo.is_number()) {
+      algorithm = providedAlgo.int_value();
+    }
```
zeha: ... else throw

pdns/ws-auth.cc
```diff
@@ -591,6 +580,208 @@ static void apiZoneCryptokeys(HttpRequest* req, HttpResponse* resp) {
...
+ *      The server returns 400 Bad Request
+ * Case 3: the backend returns true on de/activation. This means the key is de/active.
+ *      The server returns 200 OK
+ * Case 4: the backend returns false on de/activation. An error occoured.
+ *      The sever returns 422 Unprocessable Entity with message "Could not de/activate Key: :cryptokey_id in Zone: :zone_name"
+ * */
+static void apiZoneCryptokeysPUT(DNSName zonename, int inquireKeyId, HttpRequest *req, HttpResponse *resp){
+
+  //check if there is a key_id specified
+  if (!req->parameters.count("key_id"))
```
zeha: can this be different from `inquireKeyId`?

pdns/ws-auth.cc
```diff
@@ -591,6 +580,208 @@ static void apiZoneCryptokeys(HttpRequest* req, HttpResponse* resp) {
...
+  auto document = req->json();
+  //throws an exception if the key does not exist or is not a bool
+  bool active = boolFromJson(document, "active");
+
+  UeberBackend B;
+  DNSSECKeeper dk(&B);
+  DomainInfo di;
+
+  if (!B.getDomainInfo(zonename, di))
+    throw HttpBadRequestException();
```
zeha: 404 IMO

docs/markdown/httpapi/api_spec.md
```diff
@@ -707,11 +729,37 @@ Returns all public data about cryptokeys, including `privatekey`.
 
 #### PUT
 
-**TODO**: Not yet implemented.
+This method handles PUT (execute) de/activates a key from `zone_name` specified by `cryptokey_id`.
```
zeha: grammar issue here?

pdns/ws-auth.cc
```diff
@@ -591,6 +580,208 @@ static void apiZoneCryptokeys(HttpRequest* req, HttpResponse* resp) {
...
+    try{
+      insertedId = dk.addKey(zonename, keyOrZone, algorithm, bits, active);
+    }catch (std::runtime_error& error){
+      throw ApiException(error.what());
+    }
+
+    if (insertedId < 0)
+      throw ApiException("Adding key failed, perhaps DNSSEC not enabled in configuration?");
+  } else {
+    auto keyData = stringFromJson(document, "content");
```
zeha: in this case, maybe check that the other fields are not given

docs/markdown/httpapi/api_spec.md
```diff
@@ -681,21 +681,43 @@ Returns all public data about cryptokeys, but not `privatekey`.
...
+        * `json` {"error" : "Creating an algorithm algo key requires the size (in bits) to be passed"}
+    * The wrong keysize was passed:
+        * `json` {"error" : "Wrong bit size!"}
+    * If the server cant guess the keysize:
+        * `json` {"error" : "Can't guess key size for algorithm"}
+    * The key-creation failed:
+        * `json` {"error" : "Adding key failed, perhaps DNSSEC not enabled in configuration?"}
+    * The key in content has the wrong format:
+        * `json` {"error" : "Wrong key format!"}
+* `201 OK`:
```
zeha: 201 is not `OK` (should be `Created`)

docs/markdown/httpapi/api_spec.md
```diff
@@ -707,11 +729,37 @@ Returns all public data about cryptokeys, including `privatekey`.
...
 #### PUT
 
-**TODO**: Not yet implemented.
+This method handles PUT (execute) de/activates a key from `zone_name` specified by `cryptokey_id`.
+
+##### Parameters:
+
+* `:zone_name`: name of the zone for which the key with `cryptokey_id` should be de/activated
+* `cryptokey_id`: id of the key which wanted to be de/activated
+* `json`: {"active": true|false}
```
zeha: please mirror how the rest of the spec looks here

pdns/ws-auth.cc
```diff
@@ -591,6 +580,208 @@ static void apiZoneCryptokeys(HttpRequest* req, HttpResponse* resp) {
...
+    inquireKeyId = std::stoi(req->parameters["key_id"]);
+  }
+
+  if (req->method == "GET") {
+    apiZoneCryptokeysGET(zonename,inquireKeyId, resp);
+  } else if (req->method == "DELETE") {
+    apiZoneCryptokeysDELETE(zonename, inquireKeyId, req, resp);
+  } else if (req->method == "POST") {
+    apiZoneCryptokeysPOST(zonename, req, resp);
+  }else if (req->method == "PUT"){
```
zeha: missing spaces

regression-tests.api/test_cryptokeys.py
```diff
@@ -0,0 +1,269 @@
...
+    def tearDown(self):
+        super(Cryptokeys,self).tearDown()
+        try:
+            subprocess.check_output(["../pdns/pdnsutil", "--config-dir=.", "delete-zone", Cryptokeys.zone])
+        except subprocess.CalledProcessError as e:
+            self.fail("Couldn't delete zone: "+e.output)
+
+
+    # This methode test the DELETE api call. Sometimes 200 or 422 are valid return codes,
+    # because it "could" happen that the backend crashes during the test or the remove function returns false.
```
zeha: well that... is a bug?

regression-tests.api/test_cryptokeys.py
```diff
@@ -0,0 +1,269 @@
...
+
+from test_helper import ApiTestCase, is_auth
+
+@unittest.skipIf(not is_auth(), "Not applicable")
+class Cryptokeys(ApiTestCase):
+    zone = "cryptokeyzone"
+    def setUp(self):
+        super(Cryptokeys, self).setUp()
+        try:
+            subprocess.check_output(["../pdns/pdnsutil", "--config-dir=.", "create-zone", Cryptokeys.zone])
```
zeha: do that using the API instead?

regression-tests.api/test_cryptokeys.py
```diff
@@ -0,0 +1,269 @@
+import subprocess
+import json
+import unittest
+
+from test_helper import ApiTestCase, is_auth
+
+@unittest.skipIf(not is_auth(), "Not applicable")
+class Cryptokeys(ApiTestCase):
+    zone = "cryptokeyzone"
```
zeha: create a new zone name for each run

regression-tests.api/test_cryptokeys.py
```diff
@@ -0,0 +1,269 @@
...
+        try:
+            keyid = subprocess.check_output(["../pdns/pdnsutil", "--config-dir=.", "add-zone-key", Cryptokeys.zone, "ksk"])
+        except subprocess.CalledProcessError as e:
+            self.fail("pdnsutil add-zone-key failed: "+e.output)
+
+        #checks the status code. I don't know how to test explicit that the backend fail removing a key.
+        r = self.session.delete(self.url("/api/v1/servers/localhost/zones/"+Cryptokeys.zone+"/cryptokeys/"+keyid))
+        self.assertTrue(r.status_code == 200 or r.status_code == 422)
+
+        # Check that the key is actually deleted
```
zeha: split the test function per each comment that you wrote. if you write comment lines describing what the test code below does, it's a new test case.

regression-tests.api/test_cryptokeys.py
```diff
@@ -0,0 +1,269 @@
...
+            self.fail("Couldn't delete zone: "+e.output)
+
+
+    # This methode test the DELETE api call. Sometimes 200 or 422 are valid return codes,
+    # because it "could" happen that the backend crashes during the test or the remove function returns false.
+    # In this case 422 is correct. I don't know how to force false backend behavior. So i assume that the backend
+    # works correct.
+    def test_delete(self):
+        try:
+            keyid = subprocess.check_output(["../pdns/pdnsutil", "--config-dir=.", "add-zone-key", Cryptokeys.zone, "ksk"])
```
zeha: do not refer to the class name, instead use the instance

regression-tests.api/test_cryptokeys.py
```diff
@@ -0,0 +1,269 @@
...
+            self.fail("pdnsutil add-zone-key failed: "+e.output)
+
+        #checks the status code. I don't know how to test explicit that the backend fail removing a key.
+        r = self.session.delete(self.url("/api/v1/servers/localhost/zones/"+Cryptokeys.zone+"/cryptokeys/"+keyid))
+        self.assertTrue(r.status_code == 200 or r.status_code == 422)
+
+        # Check that the key is actually deleted
+        try:
+            out = subprocess.check_output(["../pdns/pdnsutil", "--config-dir=.", "list-keys", Cryptokeys.zone])
+            self.assertFalse(Cryptokeys.zone in out)
```
zeha: self.assertNotIn(zonename, out)

regression-tests.api/test_cryptokeys.py
```diff
@@ -0,0 +1,269 @@
...
+        except subprocess.CalledProcessError as e:
+            self.fail("pdnsutil list-keys failed: " + e.output)
+
+        #checks for not covered zonename
+        r = self.session.delete(self.url("/api/v1/servers/localhost/zones/"+Cryptokeys.zone+"fail/cryptokeys/"+keyid))
+        self.assertEquals(r.status_code, 400)
+
+        #checks for key is gone. Its ok even if no key had to be deleted. Or something went wrong with the backend.
+        r = self.session.delete(self.url("/api/v1/servers/localhost/zones/"+Cryptokeys.zone+"/cryptokeys/"+keyid))
+        self.assertTrue(r.status_code == 200 or r.status_code == 422)
```
zeha: `assertIn`, if you must. but this just looks wrong.

regression-tests.api/test_cryptokeys.py
```diff
@@ -0,0 +1,269 @@
...
+        return r
+
+    # Removes a key by id using the pdnsutil command
+    def remove_key(self, key_id):
+        try:
+            subprocess.check_output(["../pdns/pdnsutil", "--config-dir=.", "remove-zone-key", Cryptokeys.zone, str(key_id)])
+        except subprocess.CalledProcessError as e:
+            self.fail("pdnsutil remove-zone-key failed: "+e.output)
+
+    # Tests POST for a positiv result and deletes the added key
```
zeha: positive*

regression-tests.api/test_cryptokeys.py
```diff
@@ -0,0 +1,269 @@
...
+        r = self.add_key(content=content, algo=algo, bits=bits)
+        self.assert_success_json(r)
+        self.assertEquals(r.status_code, 201)
+        response = r.json()
+        # Only a ksk added, so expected type is csk
+        self.assertEquals(response['keytype'], 'csk')
+        key_id = response['id']
+        # Check if the key is actually added
+        try:
+            out = subprocess.check_output(["../pdns/pdnsutil", "--config-dir=.", "list-keys", Cryptokeys.zone])
```
zeha: the API should be able to tell you ...

regression-tests.api/test_cryptokeys.py
```diff
@@ -0,0 +1,269 @@
...
+
+        # check if key is activated
+        try:
+            out = subprocess.check_output(["../pdns/pdnsutil", "--config-dir=.", "show-zone", Cryptokeys.zone])
+            self.assertTrue("Active" in out)
+        except subprocess.CalledProcessError as e:
+            self.fail("pdnsutil show-zone failed: " + e.output)
+
+        #activate again
+        z = self.session.put(
```
zeha: what's `z` for?

pdns/pdnsutil.cc
```diff
@@ -1745,7 +1745,7 @@ bool secureZone(DNSSECKeeper& dk, const DNSName& zone)
 
     int algo = DNSSECKeeper::shorthand2algorithm(z_algo);
 
-    if(!dk.addKey(zone, false, algo, z_size, true)) {
+    if(!dk.addKey(zone, false, algo, z_size, true) < 0) {
```
mind04: this is always true

modules/bindbackend/binddnssec.cc
```diff
@@ -323,7 +325,22 @@ int Bind2Backend::addDomainKey(const DNSName& name, const KeyData& key)
   catch(SSqlException& se) {
     throw PDNSException("Error accessing DNSSEC database in BIND backend, addDomainKey(): "+se.txtReason());
   }
-  return true;
+
+  try {
+    d_GetLastInsertedKeyIdQuery_stmt->execute();
+    if (!d_GetLastInsertedKeyIdQuery_stmt->hasNextRow())
+      throw PDNSException("GSQLBackend unable to get id");
```
mind04: s/GSQLBackend/bind and please add more context

pdns/backends/gsql/gsqlbackend.cc
```diff
@@ -661,7 +663,22 @@ int GSQLBackend::addDomainKey(const DNSName& name, const KeyData& key)
...
   }
-  return 1; // XXX FIXME, no idea how to get the id
+
+  try {
+    d_GetLastInsertedKeyIdQuery_stmt->execute();
+    if (!d_GetLastInsertedKeyIdQuery_stmt->hasNextRow())
+      throw PDNSException("GSQLBackend unable to get id");
+    SSqlStatement::row_t row;
+    d_GetLastInsertedKeyIdQuery_stmt->nextRow(row);
+    int id = std::stoi(row[0]);
```
mind04: missing ASSERT_ROW_COLUMNS() to prevent reading out of bound (custom queries !!!)

pdns/pdnsutil.cc
benj-zen: As described in #706 the idea was to make scripting more easy. We felt that printing only the id would make further processing more convenient.

modules/gpgsqlbackend/gpgsqlbackend.cc
benj-zen: `lastval()` doesn't take any arguments. Do you mean replacing it with `currval(sequence)`?

pdns/ws-auth.cc
benj-zen: The Api spec does say [here](https://github.com/PowerDNS/pdns/blob/master/docs/markdown/httpapi/api_spec.md#common-error-causes) if an invalid `zone_id` is provided, `400 Bad Request ` should be returned, but we definitely need to return the same response for all methods.

pdns/ws-auth.cc
benj-zen: We don't know of any reasons why `removeKey` returns `false`. Any input on this would be highly appreciated to provide a more specific error message and a fitting return code.

pdns/ws-auth.cc
benj-zen: Agree with the empty body. But the spec for DELETE says to return `200` ([Link to spec](https://github.com/PowerDNS/pdns/blob/master/docs/markdown/httpapi/api_spec.md#rest)).

pdns/ws-auth.cc
benj-zen: A: We mirrored the behaviour of pdnsutil here. A new key is not active by default.

pdns/ws-auth.cc
benj-zen: We mirrored the behaviour of pdnsutil. See [here](https://github.com/PowerDNS/pdns/blob/57c447da840d6a896c98940dfab9d3d07fda83db/pdns/pdnsutil.cc#L2230-L2233).

pdns/ws-auth.cc
benj-zen: Actually this exactly [how it's done](https://github.com/PowerDNS/pdns/blob/57c447da840d6a896c98940dfab9d3d07fda83db/pdns/pdnsutil.cc#L2639-L2653) in pdnsutil.

pdns/ws-auth.cc
benj-zen: We mirrored the behaviour of pdnsutil ([See here](https://github.com/PowerDNS/pdns/blob/57c447da840d6a896c98940dfab9d3d07fda83db/pdns/pdnsutil.cc#L2247-L2250)). Maybe a new issue should be opened regarding the error handling. We are considering this out of scope of this pull request.

pdns/ws-auth.cc
benj-zen: The Api spec does say [here](https://github.com/PowerDNS/pdns/blob/master/docs/markdown/httpapi/api_spec.md#common-error-causes) if an invalid `zone_id` is provided, `400 Bad Request ` should be returned.

regression-tests.api/test_cryptokeys.py
benj-zen: We don't see the benefits of using the api, could you please explain further?

regression-tests.api/test_cryptokeys.py
benj-zen: Could you state the reasons for this? We delete the created zone during "tear down".

pdns/pdnsutil.cc
benj-zen: Thanks, the exclamation mark is a mistake.

benj-zen:
> Suggestion: use INSERT INTO ... RETURNING for DBs that support it, and use INSERT INTO; SELECT last_insert_id() for those that don't. (If submitting multiple statements works?)

@zeha This would not be that easy. The GSQL-Backend provides a generic solution for all SQL-Backends. We would have to add special cases in the abstract class for the DBs that support `INSERT INTO ... RETURNING`. We consider the current solution as cleaner.

zeha:
> INSERT ... RETURNING. We would have to add special cases in the abstract class for the ... 
@benj-zen my suggestion specifically was to do that with one statement (but obviously different statements for different backends).

benj-zen:
> @benj-zen my suggestion specifically was to do that with one statement (but obviously different statements for different backends).

@zeha We would have to take a look at it. The current solution certainly wouldn't allow that.

pdns/ws-auth.cc
pieterlexis: I think a bool that tells the server if the SEP-bit should be set would make more sense? The second element of the addKey call is `bool setSEPBit`. And pdnsutil only uses zsk/ksk to determine the value of that bool.

And yes, the terminology in pdnsutil is somewhat flawed with the advent of our use of a CSK

benj-zen: So, should we add an optional parameter for the SEP bit, which if not provided is determined by checking zsk/ksk?

pieterlexis: imho, only the "SEPbit" parameter. but @zeha probably has a more informed opinion.

benj-zen:
We updated our pull request as suggested.

pdns/pdnsutil.cc
```diff
@@ -2649,12 +2661,18 @@ loadMainConfig(g_vm["config-dir"].as<string>());
...
       cerr<<"Adding key failed, perhaps DNSSEC not enabled in configuration?"<<endl;
       exit(1);
     }
-    cout<<std::to_string(id)<<endl;
+    if (id == -1) {
+      cerr<<std::to_string(id)<<"Key was added, but backend does not support returning of key id"<<endl;
+    } else if (id < -1) {
+      cerr<<std::to_string(id)<<"Key was added, but there was a failure while returning the key id"<<endl;
+    } else {
+      cout<<std::to_string(id)<<endl;
```
zeha: this code block is now duplicated three times

pdns/ws-auth.cc
```diff
@@ -656,36 +646,42 @@ static void apiZoneCryptokeysPOST(DNSName zonename, HttpRequest *req, HttpRespon
...
-    int bits = intFromJson(document, "bits", 0);
+    int bits = 0;
+    auto docbits = document["bits"];
+    if (!docbits.is_null()) {
+      if (!docbits.is_number() || (fmod(docbits.number_value(), 1.0) != 0) || docbits.int_value() < 0) {
+        throw ApiException("'bits' must be a positive integer value");
+      } else {
+        bits = docbits.int_value();
+      }
+    }
```
zeha: why was `intFromJson` bad here?

pdns/ws-auth.cc
```diff
@@ -697,65 +693,53 @@ static void apiZoneCryptokeysPOST(DNSName zonename, HttpRequest *req, HttpRespon
...
+    } try {
+      dk->addKey(zonename, dpk,insertedId, active);
+    } catch (std::runtime_error& error) {
       throw ApiException(error.what());
     }
     if (insertedId < 0)
       throw ApiException("Adding key failed, perhaps DNSSEC not enabled in configuration?");
-
+  } else {
+    throw ApiException("Either you submit just the 'content' field or you leave 'content' empty and submit the other fields.");
```
zeha: text could be nicer

pdns/ws-auth.cc
benj-zen: Because a wrong user input would have been changed to the value 0 without the user knowing.

zeha:
Code duplication + minor nits; I think the duplication should be fixed, but all of that could be done in a followup pr.

pieterlexis:
As master is now 'unstable' for 4.1, can you guys rebase this PR so we can (finally :smile:) merge it and create an issue ticket for the remaining few issues.

benj-zen:
Allright :+1:  The rebasing is done.

PR #4114, opened 2016-07-05
dnsdist: Add `QNameLabelsCountRule()` and `QNameWireLengthRule()`

rgacogne:
* `QNameLabelsCountRule(min, max)` matches if the qname has less than
`min` or more than `max` labels.
* `QNameWireLengthRule(min, max)` matches if the qname's length on the
wire is less than `min` or more than `max` bytes.
* Also add Lua bindings for DNSName's `countLabels()` and `wirelength()`

### DIFF
--- a/pdns/dnsdist-console.cc
+++ b/pdns/dnsdist-console.cc
@@ -248,7 +248,7 @@ char* my_generator(const char* text, int state)
       "PoolAction(", "printDNSCryptProviderFingerprint(",
       "RegexRule(", "RemoteLogAction(", "RemoteLogResponseAction(", "rmResponseRule(",
       "rmRule(", "rmServer(", "roundrobin",
-      "QTypeRule(",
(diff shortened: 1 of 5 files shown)


pdns/dnsdist-console.cc
```diff
@@ -248,7 +248,7 @@ char* my_generator(const char* text, int state)
       "PoolAction(", "printDNSCryptProviderFingerprint(",
       "RegexRule(", "RemoteLogAction(", "RemoteLogResponseAction(", "rmResponseRule(",
       "rmRule(", "rmServer(", "roundrobin",
-      "QTypeRule(",
+      "QNameLabelsCountRule(", "QNameWireLengthRule(", "QTypeRule(",
```
Habbie: wonder if this list should be one item per line for better diffs. Just a thought, not suggesting it for this PR.

Habbie:
LGTM.

PR #4115, opened 2016-07-05
Improve the bind parser to support more cases

ColMelvin:
At present, the bind parser logic cannot handle `zone`s that are nested inside `view`s nor can it support a configuration that does `logging` to a `file`, both of which are common.

Add more rules to the parser to handle these cases (and do so without introducing shift/reduce or reduce/reduce conflicts).

### DIFF
--- a/pdns/Makefile.am
+++ b/pdns/Makefile.am
@@ -1196,7 +1196,6 @@ if UNIT_TESTS
 if HAVE_BOOST_GE_148
 TESTS_ENVIRONMENT = env BOOST_TEST_LOG_LEVEL=message SRCDIR='$(srcdir)'
 TESTS=testrunner
-XFAIL_TESTS=testrunner
 else
 check-local:
 	@echo "Unit tests disabled, boost is too old"
--- a/pdns/bindlexer.l
+++ b/pdns/bindlexer.l
@@ -103,6 +103,7 @@ options                 return OPTIONSTOK;
 also-notify		return ALSONOTIFYTOK;
 acl                     return ACLTOK;
 logging                 return LOGGINGTOK;
+channel                 return CHANNELTOK;
 directory               return DIRECTORYTOK;
 view                    return VIEWTOK;
 masters                 return MASTERTOK;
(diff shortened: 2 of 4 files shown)


Habbie:
'Silently' parsing these things might confuse people when we then ignore them. We need to decide what's right here.

ColMelvin:
The parser silently ignores most statements with its generic `AWORD` token.  Indeed, I count only 5 statements that are actually used: `directory`, `file`, `type`, `master`, & `also-notify` (which may exist in the `zone` or `options` clauses).  Even the ACLs, which have their own parse tree (instead of using the generic `command` tree), are completely ignored.

It's probably a good idea to document/log which statements are ignored (or alternatively, which were actually read); however, that is out of scope for this change and it may be more appropriate for auth-4.1.

ColMelvin:
@Habbie Any traction on this?

Habbie:
Sorry, not currently; feel free to remind periodically. As I said I don't like the silent 'failure' modes, so maybe it makes sense to do work on documentation and logging already. I'm bumping milestone to 4.1.x in any case as this is a change that won't go into the 4.0.x branch now that it's already released.

ColMelvin:
I have a rewrite of bindparser that improves code readability & maintainability (and includes documentation to boot).  I was waiting to submit it, partially because it builds upon this PR, but mostly because the scope of the changes seem more appropriate for 4.1.  It now seems fitting to submit it.

The change is submitted in a separate pull request (https://github.com/PowerDNS/pdns/pull/4551), as it is an independent change that we don't need for our builds.  It should provide a better base for future modifications.

ColMelvin:
@Habbie Any updates on this?

Habbie:
@ColMelvin sorry, we are occupied with other things. Rest assured that as long as it sits in the Pull queue, we will get to it eventually.

Habbie:
I agree with ignoring 'logging', this should not surprise users. I do not want to silently import zones from views into the single global view that PowerDNS has. If you can drop the first two commits, this is ready to merge.

pdns/Makefile.am
```diff
@@ -1196,7 +1196,6 @@ if UNIT_TESTS
 if HAVE_BOOST_GE_148
 TESTS_ENVIRONMENT = env BOOST_TEST_LOG_LEVEL=message SRCDIR='$(srcdir)'
 TESTS=testrunner
-XFAIL_TESTS=testrunner
```
ahupowerdns: I'm unsure what XFAIL_TESTS is, why does it need to go?

Habbie: It was added in an earlier commit inside this same PR. The PR as a whole does not touch it.

Habbie:
Closing due to inactivity.

Issue #4104, opened 2016-07-05
dnsdist rule for matching destination address


skoef:
For my specific setup I would like to have a rule similar to `NetmaskGroupRule` except for that it matches on the address the request was sent to, instead of where the request was sent from. For now I'll manage with building something custom with `dq.localaddr` but it would really make things easier if such rule existed.

rgacogne:
Fixed by #4116.

PR #4116, opened 2016-07-05
added src boolean to NetmaskGroupRule to match destination address

skoef:
In order to solve #4104, I've created this possible solution to the problem. Also split `NetmaskGroupRule` into `NMGRule` and its subclass `NetmaskGroupRule` as @cmouse suggested.

### DIFF
--- a/pdns/dnsdist-lua.cc
+++ b/pdns/dnsdist-lua.cc
@@ -70,7 +70,7 @@ std::shared_ptr<DNSRule> makeRule(const luadnsrule_t& var)
   if(nmg.empty())
     return std::make_shared<SuffixMatchNodeRule>(smn);
   else
-    return std::make_shared<NetmaskGroupRule>(nmg);
+    return std::make_shared<NetmaskGroupRule>(nmg, true);
 }
 
 std::unordered_map<int, vector<boost::variant<string,double>>> getGenResponses(unsigned int top, boost::optional<int> labels, std::function<bool(const Rings::Response&)> pred) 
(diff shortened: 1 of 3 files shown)


pdns/dnsdist-lua.cc
```diff
@@ -779,8 +779,8 @@ vector<std::function<void(void)>> setupLua(bool client, const std::string& confi
       return std::shared_ptr<DNSRule>(new SuffixMatchNodeRule(smn, quiet ? *quiet : false));
     });
 
-  g_lua.writeFunction("NetmaskGroupRule", [](const NetmaskGroup& nmg) {
-      return std::shared_ptr<DNSRule>(new NetmaskGroupRule(nmg));
+  g_lua.writeFunction("NetmaskGroupRule", [](const NetmaskGroup& nmg, bool src = true) {
```
rgacogne: With the Lua wrapper we use, you need to make `src` a `boost::optional<bool>` if you want it to be optional, which is a good idea to keep compatibility. You'll then have to check whether it is set before using it. There is an example just above in `SuffixMatchNodeRule`.

rgacogne:
I'm not sure what the use of the `NMGRule` class with the current design. Perhaps two separate classes `NetmaskGroupRule` and `DestinationNetmaskGroupRule` both inheriting from `NMGRule` would make more sense?

skoef:
I believe that was my initial suggestion on IRC, but @cmouse made this suggestion which I went for. What to do now?

rgacogne:
Alright, it does not matter much, let's merge that. Thank you!

PR #4119, opened 2016-07-06
rec: improve dnssec record skipping for non dnssec queries

### LINKED ISSUE #4084: DNSPacketWriter::commit() called without startRecord ever having been called, but a record was added

mind04:
successor of #4096

closes #4096
fixes #4084

### DIFF
--- a/pdns/pdns_recursor.cc
+++ b/pdns/pdns_recursor.cc
@@ -952,8 +952,20 @@ void startDoResolve(void *p)
 
       bool needCommit = false;
       for(auto i=ret.cbegin(); i!=ret.cend(); ++i) {
-        if(!DNSSECOK && (i->d_type == QType::RRSIG || i->d_type==QType::NSEC || i->d_type==QType::NSEC3))
+        if( ! DNSSECOK &&
+            ( i->d_type == QType::NSEC3 ||
+              (
+                ( i->d_type == QType::RRSIG || i->d_type==QType::NSEC ) &&
+                (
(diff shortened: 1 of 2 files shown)


pdns/pdns_recursor.cc
```diff
@@ -1006,7 +1006,8 @@ void startDoResolve(void *p)
 
       bool needCommit = false;
       for(auto i=ret.cbegin(); i!=ret.cend(); ++i) {
-        if(!DNSSECOK && (i->d_type == QType::RRSIG || i->d_type==QType::NSEC || i->d_type==QType::NSEC3))
+        if(!DNSSECOK && ((((i->d_type == QType::RRSIG || i->d_type==QType::NSEC) && i->d_place != DNSResourceRecord::ANSWER) || i->d_type==QType::NSEC3) ||
+                        (i->d_type == QType::RRSIG && dc->d_mdp.d_qtype != i->d_type &&  dc->d_mdp.d_qtype != QType::ANY)))
```
Habbie: Can we reformat this to more lines so the () levels are clearer? And can you replace the last `i->d_type` with `QType::RRSIG`? I think this would make the code easier to understand.

Otherwise LGTM.

Habbie:
Note: first half already merged as #4096

PR #4123, opened 2016-07-06
check-zone: warn on mismatch between algo and NSEC

### LINKED ISSUE #3267: Drop automatic algo 5->algo 7 upgrade
Right now, `DNSSECKeeper::getKeys()` fakes algo 7 when it finds a zone with key algo 5 and NSEC3 semantics. This means that the database content is actually inconsistent with the output of the nameserver. We should

4.0.0:
- [x] Have `pdnsutil check-zone` warn about this

4.1.0:
- [x] Log a warning when such a key is read

4.2.0:
- [ ] Remove the code that fakes this
- [ ] Have `pdnsutil check-zone` error about this
- [ ] ~~Have a `pdnsutil fix-algo ZONE` command to fix this in the database~~

pieterlexis:
Closes #3267

### DIFF
--- a/pdns/dbdnsseckeeper.cc
+++ b/pdns/dbdnsseckeeper.cc
@@ -448,7 +448,7 @@ DNSSECKeeper::keyset_t DNSSECKeeper::getKeys(const DNSName& zone, bool useCache)
 
     dpk.d_flags = kd.flags;
     dpk.d_algorithm = dkrc.d_algorithm;
-    if(dpk.d_algorithm == 5 && getNSEC3PARAM(zone)) // XXX Needs to go, see #3267
+    if(dpk.d_algorithm == 5 && getNSEC3PARAM(zone))
       dpk.d_algorithm+=2;
 
     KeyMetaData kmd;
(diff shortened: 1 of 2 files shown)


mind04:
no warning for algorithm  7 in NSEC zones?

Habbie:
@mind04 why should we?

@pieterlexis if a comment says 'this should go' it does not mean itself ;) reopened #3267

pdns/dbdnsseckeeper.cc
```diff
@@ -448,7 +448,7 @@ DNSSECKeeper::keyset_t DNSSECKeeper::getKeys(const DNSName& zone, bool useCache)
 
     dpk.d_flags = kd.flags;
     dpk.d_algorithm = dkrc.d_algorithm;
-    if(dpk.d_algorithm == 5 && getNSEC3PARAM(zone)) // XXX Needs to go, see #3267
```
Habbie: Please reinstate this comment.

Habbie:
One nit, then LGTM.

PR #4124, opened 2016-07-06
Debian packaging: stop recommending/suggesting some packages

### LINKED ISSUE #4118: Ubuntu xenial repository forces mysql 5.7 - not working with mariadb
I have added pdns repository in my ubuntu xenial (16.04) to update from alpha2 to rc2.

`deb [arch=amd64] http://repo.powerdns.com/ubuntu xenial-auth-40 main`

Unfortunately it's forcing mysql 5.7 and doesn't support mariadb-server-10.0 (it was replaced on my end).

PowerDNS Authoritative Server 4.0.0-alpha2 installs and works correctly with mariadb.

Is this a bug or simply mariadb support was dropped in release candidates?

zeha:
Drop Suggests: pdns-recursor is not that common on the same machine.
Drop Recommends: mysql-client, as with default apt settings, that
would remove an installed mariadb server. (Drop Recommends:
postgresql-client for consistency.)

Fixes #4118

### DIFF
--- a/build-scripts/debian-authoritative/control.in
+++ b/build-scripts/debian-authoritative/control.in
@@ -11,7 +11,7 @@ Package: pdns-server
 Architecture: any
 Depends: ${shlibs:Depends}, ${misc:Depends}, ucf (>= 0.28), adduser, lsb-base (>= 3.2-14)
 Replaces: pdns
-Suggests: pdns-backend, pdns-recursor
+Suggests: pdns-backend
 Description: extremely powerful and versatile nameserver
  PowerDNS is a versatile nameserver which supports a large number
  of different backends ranging from simple zonefiles to relational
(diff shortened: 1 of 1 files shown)


Habbie:
LGTM, @pieterlexis?

mrlesmithjr:
Is this fixed in the latest 4.x repo? I am experiencing the same issue. Been trying to figure out why mariadb keeps dropping the cluster when installing PDNS 4.x and come to find out this is exactly why. Also referencing closed issue #4118

mrlesmithjr:
Also doing the same thing with `3.4.10-1` ...So why when this use to work just fine in previous versions with a MariaDB Galera Cluster.

mrlesmithjr:
Please disregard my previous two comments. This was actually my own fault...I still had the previous pre-req of `mysql-client` being installed using my Ansible playbook.

PR #4126, opened 2016-07-06
auth: Wait for the connection to the carbon server to be established

### LINKED ISSUE #4120: Carbon issues after upgrading to auth 4.0.0 RC2
After upgrading to auth 4.0.0 RC2 the following messages show up in my syslog:

Jul  6 13:00:21 lan pdns[72642]: Unable to write data to carbon server at 127.0.0.1:2003: failed in write2WithTimeout: Socket is not connected
Jul  6 13:00:51 lan pdns[72642]: Unable to write data to carbon server at 127.0.0.1:2003: failed in write2WithTimeout: Socket is not connected
Jul  6 13:01:21 lan pdns[72642]: Unable to write data to carbon server at 127.0.0.1:2003: failed in write2WithTimeout: Socket is not connected

Version 3.4.9 doesn't have this issue and neither does recursor 4.0.0 RC1.

I'm running FreeBSD 10.3-RELEASE-p5 amd64

The metronome version I'm running is github hash 1a6dd70 

You can check the metronome data via https://metronome.cainites.net

rgacogne:
Doing a non-blocking `connect()` immediately followed by a `write()`
cause the `write()` to fail with `ENOTCONN` on FreeBSD.
This commit instead wait for the `connect()` operation to finish
using `poll()` with a short timeout if it returned `EINPROGRESS`,
so that either we have a connected socket to write to, or we fail.

Fixes #4120.

### DIFF
--- a/pdns/auth-carbon.cc
+++ b/pdns/auth-carbon.cc
@@ -46,11 +46,12 @@ try
 
     for (const auto& carbonServer : carbonServers) {
       ComboAddress remote(carbonServer, 2003);
-      Socket s(remote.sin4.sin_family, SOCK_STREAM);
-      s.setNonBlocking();
-      s.connect(remote);  // we do the connect so the attempt happens while we gather stats
 
       try {
+        Socket s(remote.sin4.sin_family, SOCK_STREAM);
+        s.setNonBlocking();
+        s.connect(remote, 2);
+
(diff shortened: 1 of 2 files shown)


Habbie:
@RvdE can you test this?

RvdE:
I've applied the patch and everything works 100% again. Thanks

PR #4137, opened 2016-07-08
gsqlite3: Check whether foreign keys should be turned on

### LINKED ISSUE #3316: be explicit about sqlite3 foreign keys
Today we turn on foreign keys in SQLite if the admin requests so, but otherwise we trust the default. Based on https://sqlite.org/pragma.html#pragma_foreign_keys I suggest we always explicitly tell SQLite the intentions of the admin.

cmouse:


Habbie:
This will be merged for auth 4.0.0. Note that this does not actually close #3316.

PR #4140, opened 2016-07-09
resolver.cc: fix warnings with gcc on musl-libc

James-TR:
According to `man 2 poll` the correct syntax for the include is:

`#include <poll.h>`

There is possibly more cases of includes that can be rectified to be
more correct, however this will do for now.

### DIFF
--- a/pdns/resolver.cc
+++ b/pdns/resolver.cc
@@ -47,7 +47,7 @@
 
 
 #include "dns_random.hh"
-#include <sys/poll.h>
+#include <poll.h>
 #include "gss_context.hh"
 #include "namespaces.hh"
 


Habbie:
1. I presume you meant `man poll`, not `man poll.h`
2. Just to be clear, is this something you ran into in practice?

James-TR:
Yes, I meant `man poll`

I ran into this warning when compiling under musl-libc. I found all other instances where `poll.h` was included in powerdns it used the `#include <poll.h>` format as well. Musl-libc's include wrappers always emit warnings when something doesn't use the `POSIX` specified paths such as the one in `<sys/poll.h>` below:

```
#warning redirecting incorrect #include <sys/poll.h> to <poll.h>
#include <poll.h>
```

PR #4142, opened 2016-07-10
add used filedescriptor statistics to auth

mind04:


regression-tests.nobackend/counters/expected_result
```diff
@@ -6,6 +6,7 @@ dnsupdate-answers=0
 dnsupdate-changes=0
 dnsupdate-queries=0
 dnsupdate-refused=0
+fd-usage=18
```
ahupowerdns: this test will surely be brittle in the future.. can you grep this metric out?

PR #4162, opened 2016-07-12
Recursor 4.0.0 DNSSEC fixes

pieterlexis:
Needs scrutinizing


pdns/pdns_recursor.cc
```diff
@@ -787,10 +788,10 @@ void startDoResolve(void *p)
       break;
     }
 
-
     if(!t_pdl->get() || !(*t_pdl)->preresolve(dc->d_remote, dc->d_local, dc->d_mdp.d_qname, QType(dc->d_mdp.d_qtype), dc->d_tcp, ret, dc->d_ednsOpts.empty() ? 0 : &dc->d_ednsOpts, dc->d_tag, &appliedPolicy, &dc->d_policyTags, res, &variableAnswer)) {
```
Habbie: we most likely also shouldn't validate if preresolve indeed took the query

Habbie:
LGTM.

pdns/pdns_recursor.cc
pieterlexis: we discussed this today, the code already ensures no DNSSEC validation will take place if the query is handled by Lua

PR #4174, opened 2016-07-14
pdnsutil: If editor is not found in path and $EDITOR is not set then default to vi

hnsk:


zeha:
I assume that breaks setups that have `emacsclient -c` in EDITOR.

PR #4175, opened 2016-07-14
dnsdist: implemented query counting

skoef:
I like the `carbonServer` functionality, but want to know more about the queries (or in my specific case: domains) that are served by dnsdist. So I implemented query counting, a per-record counter of the amount of requests for a specific record. This then is included in the data sent to carbon and can be adjusted/filtered by setting a Lua function with `setQueryCountFilter()`


pdns/dnsdist-lua.cc
```diff
@@ -1213,10 +1213,37 @@ vector<std::function<void(void)>> setupLua(bool client, const std::string& confi
       return ret;
 
     });
-  
+
   g_lua.executeCode(R"(function topQueries(top, labels) top = top or 10; for k,v in ipairs(getTopQueries(top,labels)) do show(string.format("%4d  %-40s %4d %4.1f%%",k,v[1],v[2], v[3])) end end)");
+  
+  g_lua.writeFunction("clearQueryLog", []() { 
+      unsigned int size = g_qlog.records.size();
+      g_qlog.records.clear();
```
rgacogne: It looks like this code need to acquire a write lock.

pdns/dnsdist-lua.cc
```diff
@@ -1213,10 +1213,37 @@ vector<std::function<void(void)>> setupLua(bool client, const std::string& confi
...
+      unsigned int size = g_qlog.records.size();
+      g_qlog.records.clear(); 
+      boost::format fmt("%d records cleared from query log\n");
+      g_outputBuffer = (fmt % size).str();
+    });
 
+  g_lua.writeFunction("getQueryLog", [](boost::optional<unsigned int> optMax) {
+      setLuaNoSideEffect();
+      boost::format fmt("query logging is currently: %s (%d records in buffer)\n");
+      g_outputBuffer = (fmt % (g_qlog.enabled ? "enabled" : "disabled") % g_qlog.records.size()).str();
```
rgacogne: And probably a read lock there.

pdns/dnsdist-carbon.cc
```diff
@@ -104,6 +104,19 @@ try
             str<<base<<"cache-ttl-too-shorts" << " " << cache->getTTLTooShorts() << " " << now << "\r\n";
           }
         }
+
+        {
+          WriteLock wl(&g_qcount.queryLock);
+          QueryCountRecords::iterator it;
+          std::string qname;
+          for(it=g_qcount.records.begin(); it!=g_qcount.records.end(); ++it) {
```
cmouse: maybe use `for(auto qname: g_qcount.records)` here?

pdns/dnsdist-lua.cc
```diff
@@ -1235,10 +1235,44 @@ vector<std::function<void(void)>> setupLua(bool client, const std::string& confi
...
+      ReadLock rl(&g_qcount.queryLock);
+      g_outputBuffer = "query counting is currently: ";
+      g_outputBuffer+= g_qcount.enabled ? "enabled" : "disabled";
+      g_outputBuffer+= (boost::format(" (%d records in buffer)\n") % g_qcount.records.size()).str();
+
+      boost::format fmt("%-3d %s: %d request(s)\n");
+      QueryCountRecords::iterator it;
+      unsigned int max = optMax ? *optMax : 10;
+      unsigned int index{1};
+      for(it = g_qcount.records.begin(); it != g_qcount.records.end(); ++it, ++index) {
```
cmouse: `for(it = g_qcount.records.begin(); it != g_qcount.records.end() && index <= max; ++it, ++index);`

cmouse: or just `it = g_qcount.records.begin() + std::min(g_qcount.records.size(), max);`

pdns/README-dnsdist.md
```diff
@@ -971,6 +971,47 @@ latest version of [PowerDNS
 Metronome](https://github.com/ahupowerdns/metronome) comes with attractive
 graphs for `dnsdist` by default.
 
+Query counters
+-------------
+When using `carbonServer`, it is also possible to send per-records statistics of
+the amount of queries by using `setQueryCount(true)`. With query counting enabled,
+`dnsdist` will increase a counter for every unique record or the behaviour you define
+in a custom Lua function by setting `setQueryCountFilter(func)`. This filter can decide
+whether to keep count on a query at all or rewrite the queried record.
```
rgacogne: Perhaps clarify that only the counter key is rewritten, not the query. It should make sense of course, but I am worried about someone skimming over the README looking how to rewrite queries.

pdns/README-dnsdist.md
```diff
@@ -971,6 +971,47 @@ latest version of [PowerDNS
...
+```
+function filter(dq)
+  qname = dq.qname:toString()
+
+  -- don't count PTRs at all
+  if(qname:match('in%-addr.arpa$')) then
+    return false, ""
+  end
+
+  -- rewrite these queries
```
rgacogne: Same thing here, it might be useful to clarify that the counting key is rewritten, not the query.

pdns/README-dnsdist.md
```diff
@@ -971,6 +971,47 @@ latest version of [PowerDNS
...
+  return true, qname
+end
+
+setQueryCountFilter(filter)
+```
+
+Valid return values for `QueryCountFilter` functions are:
+
+* `true`: count the specified query
+* `false`: discard the query
```
rgacogne: Perhaps replace `discard the query` by `don't count the query`?

pdns/dnsdist.cc
```diff
@@ -748,6 +749,22 @@ bool processQuery(LocalStateHolder<NetmaskTree<DynBlock> >& localDynNMGBlock,
     g_rings.queryRing.push_back({now,*dq.remote,*dq.qname,dq.len,dq.qtype,*dq.dh});
   }
 
+  if(g_qcount.enabled) {
+    WriteLock wl(&g_qcount.queryLock);
```
rgacogne: I don't think you need to acquire a WriteLock until after `if (countQuery) {` evaluates to true below

pdns/dnsdist.cc
```diff
@@ -748,6 +749,22 @@ bool processQuery(LocalStateHolder<NetmaskTree<DynBlock> >& localDynNMGBlock,
...
+    string qname = (*dq.qname).toString(".", false);
+    bool countQuery{true};
+    if(g_qcount.filter) {
+      std::tie (countQuery, qname) = g_qcount.filter(dq);
```
rgacogne: I think you should acquire the Lua mutex here with something like `std::lock_guard<std::mutex> lock(g_luamutex);`

pdns/dnsdist.cc
```diff
@@ -748,6 +749,22 @@ bool processQuery(LocalStateHolder<NetmaskTree<DynBlock> >& localDynNMGBlock,
     g_rings.queryRing.push_back({now,*dq.remote,*dq.qname,dq.len,dq.qtype,*dq.dh});
   }
 
+  if(g_qcount.enabled) {
+    WriteLock wl(&g_qcount.queryLock);
+    string qname = (*dq.qname).toString(".", false);
```
rgacogne: Just a nit, but for a better readability we have `toStringNoDot()` that does `toString(".", false)`

rgacogne:
LGTM once travis is happy. Thank you!

PR #4182, opened 2016-07-14
Remove zone count from pdnsutil list-all-zones

### LINKED ISSUE #3750: Remove zone count from pdnsutil list-all-zones?
The output would be more useful for scripting if it's just a list of zone names. Of course, there's `head -n -1`…but then, there's also `wc -l`.

nivex:
Closes #3750


Habbie:
how about `cerr` instead?

nivex:
That could work, but it wasn't mentioned in the original issue, and I agree with the original reporter that "wc -l" is just as effective if that information is desired.

Habbie:
@maikzumstrull thoughts on `cerr`?

Habbie:
@cyclops1982 any idea why you added zone count? was it specifically requested?

mind04:
or only show zone count if pdnsutil is running in verbose mode (g_verbose)

maikzumstrull:
Splitting closely related messages between stdout and stderr doesn't seem like a good idea to me. "Only show in verbose mode" sounds okay, but so does just removing it.

Habbie:
+1 on verbose. Let's skip the `isatty` then.

cyclops1982:
@Habbie Probably just because that could be useful for some scripting and/or stats. Back then i also did some munin stats and you would be able to parse this output and make a graph based on that.

Then again, i didn't have a specific feature request to implement it, so it was more "i'll add it while i'm at it"

Habbie:
Can you update this patch to not break compilation, and to respect verbosity?

nivex:
I'll try this again when I figure out what I'm doing :)

PR #4189, opened 2016-07-15
Update EOL doc to include 4.x release information

setharnold:


cmouse:
You know you could leave mention about 3.x. Now it's like it never even existed.

PR #4195, opened 2016-07-15
API: prevent duplicate records in single RRset

### LINKED ISSUE #4132: API zone create  - nameservers array / NS RrSet
Hello,

I've noticed 'strange' behavior while creating new zone containing NS records via API.
pdns version / dist package: 4.0.0~alpha2-3build1 @ ubuntu 16.04 lts

assume having two NS fqdns (ns1.example.org. and ns2.example.org.)

scenario A)
- zone's nameservers array filled with valid NS fqdn's
- NS rrset present (containing two records, each record.content equals to certain NS fqdn)
- result is newly created zone, but NS RrSet contains 4 records (each NS fqdn duplicated)

scenario B)
- zones nameservers array empty
- NS rrset present (containing two records, each record.content equals to certain NS fqdn)
- result is newly created zone with 2 Records inside of the NS RrSet - which is correct

Is this by design or may it be considered as a bug ?
Afaik there needs to be checked whether the incoming zone object contains RrSet of type NS or not.
If so, there shouldn't take palce any 'appending' of RR's

Thank you,
Regards,
Adrian

zeha:
- [ ] documented the code
- [X] added regression tests
- [ ] added unit tests

### DIFF
--- a/pdns/ws-auth.cc
+++ b/pdns/ws-auth.cc
@@ -991,6 +991,26 @@ static void gatherRecordsFromZone(const std::string& zonestring, vector<DNSResou
   }
 }
 
+/** Throws ApiException if records with duplicate name/type/content are present.
+ *  NOTE: sorts records in-place.
+ */
+static void checkDuplicateRecords(vector<DNSResourceRecord>& records) {
+  sort(records.begin(), records.end(),
+    [](const DNSResourceRecord& rec_a, const DNSResourceRecord& rec_b) -> bool {
+      return rec_a.qname.toString() > rec_b.qname.toString() || \
+        rec_a.qtype.getCode() > rec_b.qtype.getCode() || \
+        rec_a.content < rec_b.content;
+    }
+  );
+  DNSResourceRecord previous;
+  for(const auto& rec : records) {
+    if (previous.qtype == rec.qtype && previous.qname == rec.qname && previous.content == rec.content) {
+      throw ApiException("Duplicate record in RRset " + rec.qname.toString() + " IN " + rec.qtype.getName() + " with content \"" + rec.content + "\"");
+    }
+    previous = rec;
+  }
+}
+
 static void apiServerZones(HttpRequest* req, HttpResponse* resp) {
   UeberBackend B;
   DNSSECKeeper dk(&B);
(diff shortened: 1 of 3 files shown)


zeha:
Somebody please review this.

pieterlexis:
looks good to me, please rebase :+1:

tuxis-ie:
May I please object? Although this might be technically correct, it is not very userfriendly. When importing a zone from another setup, I would just like to AXFR the zone, copy paste it into nsedit and be done with it. If I choose not to send an array with nameservers, just use the nameservers that are already in the zone, or b0rk. If I however do send an array with nameservers, override the nameservers in the zone.

If you b0rk on nameservers in the array and in the zone, you are forcing users to first review the whole zone before they are able to import.

So please reconsider this.

zeha:
@tuxis-ie This would imply the outcome of sending conflicting data is a silent loss of either data. If you want a fallback, client code can handle this. (Either by looking at the data before sending, or by sending 'replace NS' afterwards.)

zeha:
@pieterlexis rebased.

zeha:
Rebased once more.

pdns/ws-auth.cc
```diff
@@ -990,6 +990,24 @@ static void gatherRecordsFromZone(const std::string& zonestring, vector<DNSResou
   }
 }
 
+/** Throws ApiException if records with duplicate name/type/content are present.
+ *  NOTE: sorts records in-place.
+ */
+static void checkDuplicateRecords(vector<DNSResourceRecord>& records) {
+  sort(records.begin(), records.end(),
+    [](const DNSResourceRecord& rec_a, const DNSResourceRecord& rec_b) -> bool {
+      return rec_a.qname.toString() > rec_b.qname.toString() || rec_a.qtype.getCode() > rec_b.qtype.getCode();
```
mind04: Content must be part of the sorting, without it there is no guarantee duplicates are sequential if there are 3 or more records with the same "name type" Could use a test.
Is qname always lower case here?

zeha:
Worked in review feedback + rebased.

pdns/ws-auth.cc
```diff
@@ -991,6 +991,26 @@ static void gatherRecordsFromZone(const std::string& zonestring, vector<DNSResou
...
 
+/** Throws ApiException if records with duplicate name/type/content are present.
+ *  NOTE: sorts records in-place.
+ */
+static void checkDuplicateRecords(vector<DNSResourceRecord>& records) {
+  sort(records.begin(), records.end(),
+    [](const DNSResourceRecord& rec_a, const DNSResourceRecord& rec_b) -> bool {
+      return rec_a.qname.toString() > rec_b.qname.toString() || \
+        rec_a.qtype.getCode() > rec_b.qtype.getCode() || \
+        rec_a.content < rec_b.content;
```
Habbie: Has content been canonicalized at this point?

zeha: Yes!

PR #4198, opened 2016-07-18
newServer setting maxCheckFailures makes no sense

stutiredboy:
when dnsdist add `newServer` with `maxCheckFailures` parameter, such as `maxCheckFailures=3`, if the downstream server seems to be down, then the `dss->currentCheckFailures` will increase 1.

but, if `dss->currentCheckFailures` not reach the `maxCheckFailures` value, at this time, the downstream server seems to be up, the `dss->currentCheckFailures` was not reset to 0.

so when `dss->currentCheckFailures` reach `maxCheckFailures`, dnsdist will show downstream down, the change to up in the next second.

```
dnsdist[24734]: Marking downstream 10.63.49.12:53 as 'down'
dnsdist[24734]: Marking downstream 10.63.49.12:53 as 'up'
```

In my opinion, discontinuous failure retry makes no sense, so I post this pull request.

Thanks.

### DIFF
--- a/pdns/dnsdist.cc
+++ b/pdns/dnsdist.cc
@@ -1254,7 +1254,12 @@ void* healthChecksThread()
     for(auto& dss : g_dstates.getCopy()) { // this points to the actual shared_ptrs!
       if(dss->availability==DownstreamState::Availability::Auto) {
 	bool newState=upCheck(*dss);
-	if (!newState && dss->upStatus) {
+	if (newState) {
+		if (dss->currentCheckFailures != 0) {
+			dss->currentCheckFailures = 0;
+		}
+	}
+	else if (!newState && dss->upStatus) {
 	  dss->currentCheckFailures++;
 	  if (dss->currentCheckFailures < dss->maxCheckFailures) {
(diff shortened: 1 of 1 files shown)


rgacogne:
Hi,

This change makes sense, thank you! Could you just fix the indentation? I know it's not easy as we still have a mix of tabs and spaces in some places.

pdns/dnsdist.cc
```diff
@@ -1254,7 +1254,12 @@ void* healthChecksThread()
     for(auto& dss : g_dstates.getCopy()) { // this points to the actual shared_ptrs!
       if(dss->availability==DownstreamState::Availability::Auto) {
 	bool newState=upCheck(*dss);
-	if (!newState && dss->upStatus) {
+	if (newState) {
+		if (dss->currentCheckFailures != 0) {
```
rgacogne: You could perhaps skip that test, but that doesn't matter much.

stutiredboy:
rager that. I use space as the indent.

rgacogne:
LGTM

PR #4206, opened 2016-07-19
auth: Don't try to deallocate empty PG statements

### LINKED ISSUE #4201: DEALLOCATE: syntax error at end of input at character 12
This is Debian bug#831741, about pdns-backend-pgsql:

```
The pg log gets spammed with:

2016-07-18 20:45:26 GMT ERROR:  syntax error at end of input at character 12
2016-07-18 20:45:26 GMT STATEMENT:  DEALLOCATE

constantly, hundreds of times per second.

I had to set log_min_messages and log_min_error_statement to fatal to avoid it.

Something is wrong with the sql pdns sends to pg.

This is a recent regression.  I did not notice with with the first 4.0 pres in sid.
```

(And indeed, the line adding the `DEALLOCATE` statement was added in 4.0.0-rc1.)

rgacogne:
When a SPgSQLStatement is released without having been prepared,
we execute an invalid 'DEALLOCATE ' SQL command. This might happen
if the statement has not been used before being destroyed, for example.
Fixes #4201.

### DIFF
--- a/modules/gpgsqlbackend/spgsql.cc
+++ b/modules/gpgsqlbackend/spgsql.cc
@@ -184,9 +184,12 @@ private:
   void releaseStatement() {
     d_prepared = false;
     reset();
-    string cmd = string("DEALLOCATE " + d_stmt);
-    PGresult *res = PQexec(d_db(), cmd.c_str());
-    PQclear(res);
+    if (!d_stmt.empty()) {
+      string cmd = string("DEALLOCATE " + d_stmt);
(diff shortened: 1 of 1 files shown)


zeha:
Do we know /why/ this happens in the first place?

rgacogne:
No, I upgraded my personal PowerDNS w/ PostgreSQL setup recently and I don't have the issue. I'd like to see if there is anything in the corresponding PowerDNS logs.

PR #4217, opened 2016-07-21
turn on root-nx-trust by default and log-common-errors=off, and document that

ahupowerdns:


pieterlexis:
an explicit "yes" might be better (as this ends up in the generated config file)

PR #4230, opened 2016-07-24
Allow a pipe backend to send a higher SOA

pieterlexis:
This prevents superfluous queries to the pipe-command and will speed up
lookups.


mind04:
Does not work. Implement getAuth() in pipe if you want this.
There is no way for a backed to distinct a SOA query for authority from one for a negative answer or a direct lookup.

Issue #4226, opened 2016-07-23
Lua hook prior to RPZ filtering (rec 4.0)


mzealey:
It would be great if we could get a lua hook function prior to RPZ filtering. At the moment if something is listed in an RPZ the response gets decided without `preresolve` being called. We want the RPZ lookup, but it would be nice to be able to turn it off for selected hosts/domains too.

I would suggest this would be hooked into the code at pdns/pdns_recursor.cc:811 just after the `getQueryPolicy` call, but before it has been processed. This would call into lua perhaps a function named `postquerypolicy` with the same arguments as `preresolve`, but also having the `dfepol` data structure passed in read/write allowing it to be manipulated prior to the switch statement (return value ignored). This would then allow for any logic in the RPZ to be overridden by lua if required.

Example use cases: You want the RPZ to usually be enforced, but disable it for certain source IP/domain combinations, You want to change the RPZ response for a certain set of source IPs.

Issue #4088, opened 2016-07-01
RPZ lookup from within lua script


mzealey:
It would make the lua interface a lot more flexible if there was a way to:

1) Disable RPZ lookups for certain ips/domains eg in preresolve fn: `dq.disableRPZLookup = 1`

2) Do an inline RPZ lookup from within the lua script (or callback, but I think RPZ lookups are inline as they have to be on the local server). eg `local rpz_result = RPZLookup(dq.qname)` or perhaps the `dq` as you need to know if it was a tcp connection etc. Although it would be nice to be able to specify a custom domain or whatever rather than only being able to operate on the current dq object.

omoerbeek:
I think this is mostly covered by the current RPZ code and policy events.

Issue #4086, opened 2016-07-01
rpz blocks subdomains by default


mzealey:
pdns 4.0-rc1. Flat RPZ file like:

```
$TTL 2h;
$ORIGIN domain.example.com.
@               SOA powerdns.example.net. hostmaster.example.com ( 1 12h 15m 3w 2h)
                NS powerdns.example.net.
; begin RPZ RR definitions

nxdomain.org        CNAME .
```

```
dig +short @10.10.7.77 foo.nxdomain.org
1.2.3.4
```

Whereas spec says it should only do that if I specif '*.nxdomain.org'.

I actually prefer this way of blocking the whole domain, but it does seem to be against the spec and @Habbie agrees.

Habbie:
Going to bump this to rec-4.0.x as this bug is unlikely to hurt most RPZ users - who expects legal content on a subdomain of a malware domain?

mzealey:
@Habbie concern is more that if someone devs against it like this and forgets to put the wildcards in then a later release will allow the subdomains through. But agreed that it is pointless to expect whitelisting under a domain that is blocked.

Habbie:
Want to PR a line on it for the changelog? :)

Habbie:
rec-4.0.0 milestone just for mentioning this in the changelog as a known issue. Issue itself is 4.0.x.

PR #4236, opened 2016-07-27
Allow Lua access to the result of the Policy Engine decision

pieterlexis:
This PR fixes some small things with the RPZ code, adds basic RPZ tests and allows access to the policy decision inside of Lua. It also adds metrics for the taken decision.

Fixes:
 * Whitespace and comment fixes
 * Correct subzone blocking (#4086)

I think this PR addresses the issues raised in #4088 and #4226 by looking up the policy and allow it to be changed inside existing hooks. This should be flexible enough.

Comments and code reviews welcome

### DIFF
--- a/pdns/lua-recursor4.cc
+++ b/pdns/lua-recursor4.cc
@@ -6,6 +6,7 @@
 #include "namespaces.hh"
 #include "rec_channel.hh" 
 #include "ednssubnet.hh"
+#include "filterpo.hh"
 #include <unordered_set>
 
 #if !defined(HAVE_LUA)
@@ -24,12 +25,12 @@ bool RecursorLua4::nodata(const ComboAddress& remote,const ComboAddress& local, 
   return false;
 }
 
-bool RecursorLua4::postresolve(const ComboAddress& remote,const ComboAddress& local, const DNSName& query, const QType& qtype, bool isTcp, vector<DNSRecord>& ret, std::string* appliedPolicy, std::vector<std::string>* policyTags, int& res, bool* variable)
+bool RecursorLua4::postresolve(const ComboAddress& remote,const ComboAddress& local, const DNSName& query, const QType& qtype, bool isTcp, vector<DNSRecord>& ret, DNSFilterEngine::Policy* appliedPolicy, std::vector<std::string>* policyTags, int& res, bool* variable)
 {
   return false;
 }
 
-bool RecursorLua4::preresolve(const ComboAddress& remote, const ComboAddress& local, const DNSName& query, const QType& qtype, bool isTcp, vector<DNSRecord>& ret, const vector<pair<uint16_t,string> >* ednsOpts, unsigned int tag, std::string* appliedPolicy, std::vector<std::string>* policyTags, int& res, bool* variable)
+bool RecursorLua4::preresolve(const ComboAddress& remote, const ComboAddress& local, const DNSName& query, const QType& qtype, bool isTcp, vector<DNSRecord>& ret, const vector<pair<uint16_t,string> >* ednsOpts, unsigned int tag, DNSFilterEngine::Policy* appliedPolicy, std::vector<std::string>* policyTags, int& res, bool* variable)
 {
   return false;
 }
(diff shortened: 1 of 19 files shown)


pdns/lua-recursor4.cc
```diff
@@ -516,8 +553,10 @@ bool RecursorLua4::genhook(luacall_t& func, const ComboAddress& remote,const Com
   dq->ednsOptions = ednsOpts;
   dq->isTcp = isTcp;
   dq->policyTags = policyTags;
+  dq->appliedPolicy = appliedPolicy;
   bool handled=func(dq);
   if(variable) *variable |= dq->variable; // could still be set to indicate this *name* is variable, even if not 'handled'
+  appliedPolicy = dq->appliedPolicy; // The policy might have been changed, but the query can still be not handled by Lua
```
rgacogne: If I'm not mistaken some hooks are calling `genhook()` with `nullptr` in `appliedPolicy`.

pieterlexis: you are correct (test confirm it)

pieterlexis:
Added a way for Lua to disable RPZ processing and implemented NSDNAME and NSIP RPZ triggers.

pdns/lua-recursor4.cc
```diff
@@ -516,8 +554,11 @@ bool RecursorLua4::genhook(luacall_t& func, const ComboAddress& remote,const Com
   dq->ednsOptions = ednsOpts;
   dq->isTcp = isTcp;
   dq->policyTags = policyTags;
+  dq->appliedPolicy = appliedPolicy;
+  dq->wantsRPZ = wantsRPZ;
   bool handled=func(dq);
   if(variable) *variable |= dq->variable; // could still be set to indicate this *name* is variable, even if not 'handled'
+  *wantsRPZ = dq->wantsRPZ; // Even if we did not handle the query, RPZ could be disabled
```
rgacogne: `wantsRPZ` might be `NULL`, for example in the `nxdomain`, `nodata` or `preoutquery` hooks.

pdns/lua-recursor4.cc
```diff
@@ -516,8 +554,11 @@ bool RecursorLua4::genhook(luacall_t& func, const ComboAddress& remote,const Com
   dq->ednsOptions = ednsOpts;
   dq->isTcp = isTcp;
   dq->policyTags = policyTags;
+  dq->appliedPolicy = appliedPolicy;
+  dq->wantsRPZ = wantsRPZ;
```
rgacogne: Shouldn't that be `dq->wantsRPZ = *wantsRPZ;` if `wantsRPZ` is not `NULL`?

pdns/pdns_recursor.cc
```diff
@@ -611,15 +611,15 @@ catch(...)
 }
 
 #ifdef HAVE_PROTOBUF
-static void protobufLogQuery(const std::shared_ptr<RemoteLogger>& logger, uint8_t maskV4, uint8_t maskV6, const boost::uuids::uuid& uniqueId, const ComboAddress& remote, const ComboAddress& local, const Netmask& ednssubnet, bool tcp, uint16_t id, size_t len, const DNSName& qname, uint16_t qtype, uint16_t qclass, const std::string appliedPolicy, const std::vector<std::string>& policyTags)
+static void protobufLogQuery(const std::shared_ptr<RemoteLogger>& logger, uint8_t maskV4, uint8_t maskV6, const boost::uuids::uuid& uniqueId, const ComboAddress& remote, const ComboAddress& local, const Netmask& ednssubnet, bool tcp, uint16_t id, size_t len, const DNSName& qname, uint16_t qtype, uint16_t qclass, const DNSFilterEngine::Policy appliedPolicy, const std::vector<std::string>& policyTags)
```
rgacogne: Perhaps we should keep passing the name instead of the entire policy, it prevents us from having to create a temporary `DNSFilterEngine::Policy()` when we don't have one.

pdns/lua-recursor4.cc
pieterlexis: That shouldn't be an issue, at that point, all RPZ related code is run. but NULL might be bad though

pdns/pdns_recursor.cc
```diff
@@ -998,7 +1047,7 @@ void startDoResolve(void *p)
     if (luaconfsLocal->protobufServer) {
       pbMessage.setBytes(packet.size());
       pbMessage.setResponseCode(pw.getHeader()->rcode);
-      pbMessage.setAppliedPolicy(appliedPolicy);
+      pbMessage.setAppliedPolicy(appliedPolicy.d_name);
```
rgacogne: Since your `preresolve` and `postresolve` examples only override `dq.appliedPolicy.policyKind` with `pdns.policykinds.NoAction`, it looks like we might be logging the name of a policy that wasn't applied. Perhaps we should check whether `appliedPolicy.d_kind` is different from `DNSFilterEngine::PolicyKind::NoAction` and `wantsRPZ` is true?

pdns/pdns_recursor.cc
pieterlexis: maybe we should just dereference it? (`const DNSFilterEngine::Policy &appliedPolicy`)

pdns/pdns_recursor.cc
pieterlexis: I think testing if `appliedPolicy` differs from the (const) `dfepol`

pieterlexis:
incorporated in and superseded by #4324

PR #4248, opened 2016-07-28
 direct nsec nxdomain

mind04:


Habbie:
with `gsqlite3` test suite:
```
 dig @127.0.0.1 -p 5300 dsdelegation.example.com nsec

; <<>> DiG 9.11.0a2 <<>> @127.0.0.1 -p 5300 dsdelegation.example.com nsec
; (1 server found)
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 32281
;; flags: qr aa rd; QUERY: 1, ANSWER: 1, AUTHORITY: 0, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 1680
;; QUESTION SECTION:
;dsdelegation.example.com.	IN	NSEC

;; ANSWER SECTION:
dsdelegation.example.com. 86400	IN	NSEC	escapedtext.example.com. NS DS RRSIG NSEC

;; Query time: 2 msec
;; WHEN: Thu Jul 28 18:09:01 CEST 2016
;; MSG SIZE  rcvd: 98

```

this should hand out the (signed) delegation, like a query for `A` does.

(In related news, a non-dnssec `A` query hands out NS+DS while other servers only hand out NS there, wonder if we should change that.. probably not)

wido:
This should fix the issue I reported here: https://mailman.powerdns.com/pipermail/pdns-dev/2016-July/001608.html

Habbie:
@wido it does, I checked :)

PR #4252, opened 2016-07-29
auth: Don't include bind files if length <= 2 or > sizeof(filename)

rgacogne:
This prevents crashing on an invalid `include` directive in the bind backend configuration file.


Habbie:
code LGTM

PR #4269, opened 2016-08-02
contrib: simple bash completion for pdnsutil

j0ju:
Feel free to add this simple helper to PowerDNS distribution.


Habbie:
Works for me on OSX. Merging this will not actually cause distribution, another PR (or just ticket) might be handy for that!

Issue #4270, opened 2016-08-02
Boost 1.61 hides fcontext


Habbie:
Which means Recursor cannot use Boost Context on 1.61 and up.

phonedph1:
We've worked around this in our build by doing the following hack:

07:41 < ph1> Hey HAbbie, this is all we do to make boost 1.61 work for us: cp ${RPM_BUILD_DIR}/%{name}-%{version}/boost_1_61_0/boost/context/detail/fcontext.hpp 
             ${RPM_BUILD_DIR}/%{name}-%{version}/boost_1_61_0/boost/context/fcontext.hpp
07:41 < ph1> ./configure --with-boost=${RPM_BUILD_DIR}/%{name}-%{version}/boost_1_61_0/ --prefix=/usr --sysconfdir=/etc/powerdns

Habbie:
ph1's hack does not work for me on osx with brew boost 1.61.0:
```
In file included from mtasker_context.cc:6:
./mtasker_fcontext.cc:8:7: error: no member named 'make_fcontext' in namespace 'boost::context'; did you mean 'boost::context::detail::make_fcontext'?
using boost::context::make_fcontext;
      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      boost::context::detail::make_fcontext
/usr/local/include/boost/context/fcontext.hpp:33:12: note: 'boost::context::detail::make_fcontext' declared here
fcontext_t make_fcontext( void * sp, std::size_t size, void (* fn)( transfer_t) );
           ^
In file included from mtasker_context.cc:6:
./mtasker_fcontext.cc:39:7: error: no member named 'fcontext_t' in namespace 'boost::context'; did you mean 'boost::context::detail::fcontext_t'?
using boost::context::fcontext_t;
      ^~~~~~~~~~~~~~~~~~~~~~~~~~
      boost::context::detail::fcontext_t
/usr/local/include/boost/context/fcontext.hpp:23:15: note: 'boost::context::detail::fcontext_t' declared here
typedef void* fcontext_t;
              ^
In file included from mtasker_context.cc:6:
./mtasker_fcontext.cc:40:7: error: no member named 'jump_fcontext' in namespace 'boost::context'; did you mean 'boost::context::detail::jump_fcontext'?
using boost::context::jump_fcontext;
      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      boost::context::detail::jump_fcontext
[...]

PR #4271, opened 2016-08-02
fix type

Habbie:
I don't see how this code ever compiled. On OSX (10.10) and Debian 8 I can find `struct ucontext` and `ucontext_t` but never `ucontext`. Yet, git log tells me this code works on EL6 i386. Needs testing there.

For OSX, this works around #4270.


Habbie:
Got report from @mind04 that this passes on el6/i386.

Habbie:
Could use a review here.

zeha:
From my limited understanding of *context, LGTM.

Habbie:
As to how this code ever compiled, @zeha checked that on glibc indeed `ucontext` maps to `struct ucontext`. However, OSX does not ship `struct ucontext` by default which explains why the code failed there until this PR.

Habbie:
@zeha thanks for insights and review!

PR #4285, opened 2016-08-04
dnsdist: Handle unreachable servers at startup, reconnect stale sockets

### LINKED ISSUE #4131: dnsdist (on raspbian) does not start after reboot if network takes time to start
```
$ apt-get install dnsdist
$ netstat -anp|grep dnsdist
tcp        0      0 127.0.0.1:53            0.0.0.0:*               LISTEN      7584/dnsdist
udp        0      0 127.0.0.1:53            0.0.0.0:*                           7584/dnsdist
unix  3      [ ]         STREAM     CONNECTED     18741    7584/dnsdist

$ dnsdist --client
Unable to read configuration from '/etc/dnsdist/dnsdist.conf'
Fatal error: connecting socket to 127.0.0.1:5199: Connection refused
```

### LINKED ISSUE #4155: dnsdist downstream sockets can go stale but it never recovers
I'm using dnsdist (1.0) with some downstream servers over VPN and sometimes the connected UDP socket can go stale. It does not recover from this state and I have to restart dnsdist to fix it.

Although the downstream server can still be marked UP or DOWN by the `healthChecksThread()` as it does not use connected sockets there.

The UDP `send()` in `udpClientSendRequestToBackend` will actually return an error code of -1 but it's only checked to increase the error counters.

What I expected was that it would reconnect the socket to recover from this error.

I implemented the following `reconnect()` function in the `Downstream` class which I call after increasing the error counters:

```
void reconnect()
{
   shutdown(fd, SHUT_RDWR);
   close(fd);
   if (!IsAnyAddress(remote)) {
      fd = SSocket(remote.sin4.sin_family, SOCK_DGRAM, 0);
      if (!IsAnyAddress(sourceAddr)) {
        SSetsockopt(fd, SOL_SOCKET, SO_REUSEADDR, 1);
        SBind(fd, sourceAddr);
      }
      SConnect(fd, remote);
   }
}
```

The `shutdown()` was needed to signal the blocking `recv()` call in the `responderThread`.

I'm not sure if this is the correct way, but it works well for me.

rgacogne:
- If we can't `connect()` when the server is added, keep the server around and `connect()` the socket only when the health check succeeds. Fixes #4131.
- We now try to detect stale backend socket and to recover from it. Fixes #4155.

### DIFF
--- a/pdns/dnsdist-lua.cc
+++ b/pdns/dnsdist-lua.cc
@@ -226,13 +226,15 @@ vector<std::function<void(void)>> setupLua(bool client, const std::string& confi
 			  addServerToPool(localPools, "", ret);
 			  g_pools.setState(localPools);
 
-			  if(g_launchWork) {
-			    g_launchWork->push_back([ret]() {
-				ret->tid = move(thread(responderThread, ret));
+			  if (ret->connected) {
+			    if(g_launchWork) {
+			      g_launchWork->push_back([ret]() {
+			        ret->tid = move(thread(responderThread, ret));
 			      });
-			  }
-			  else {
-			    ret->tid = move(thread(responderThread, ret));
+			    }
+			    else {
+			      ret->tid = move(thread(responderThread, ret));
+			    }
 			  }
 
 			  return ret;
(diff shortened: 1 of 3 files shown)


rgacogne:
ready for review!

PR #4287, opened 2016-08-05
auth: better handling of invalid serial

### LINKED ISSUE #4281: Bad error report on invalid SOA
If your SOA record is invalid like:
`ns1.srv1.dk dns.srv1.dk 20160080400 14400 3600 604800 600`
instead of the correct:
`ns1.srv1.dk dns.srv1.dk 2016080400 14400 3600 604800 600`

PowerDNS refuses to start and you just get this error:
```
Exiting because communicator thread died with STL error: stop
pdns.service: Main process exited, code=exited, status=1/FAILURE
Failed to start PowerDNS Authoritative Server.
```

If you run check-all-zones..
```
root@ns1 ~# pdnsutil check-all-zones
Error: stou
```

Ubuntu 16.04.1 LTS
PowerDNS 4.0.1 from the repo

rgacogne:
Fixes #4281.

### DIFF
--- a/pdns/dnsbackend.cc
+++ b/pdns/dnsbackend.cc
@@ -337,20 +337,25 @@ void fillSOAData(const string &content, SOAData &data)
   if(pleft>1) 
     data.hostmaster=DNSName(attodot(parts[1])); // ahu@ds9a.nl -> ahu.ds9a.nl, piet.puk@ds9a.nl -> piet\.puk.ds9a.nl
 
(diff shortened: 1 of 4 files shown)


Habbie:
On the second commit: the pain is that we parse the wire format into text and then split it, which is terrible. In theory indeed we should not get out of range there but checking is good. One day we should stop making it text.

PR #4288, opened 2016-08-05
rec: Respect the timeout when connecting to a protobuf server

### LINKED ISSUE #4105: Recursor 4.0.0-rc1 hangs at startup if protobufServer( "..." ) is set.
I tried to play with protobufServer() in lua-config-file. But with this setting, the Recursor doesn't start anymore. It hangs at "Reading random entropy from":

    # /usr/sbin/pdns_recursor --daemon=no --write-pid=no
    Jul 05 13:20:09 PowerDNS Recursor 4.0.0-rc1 (C) 2001-2016 PowerDNS.COM BV
    Jul 05 13:20:09 Using 64-bits mode. Built using gcc 4.8.5 on Jun 17 2016 12:00:42 by abuild@lamb09.
    Jul 05 13:20:09 PowerDNS comes with ABSOLUTELY NO WARRANTY. This is free software, and you are welcome to redistribute it according to the terms of the GPL version 2. 
    Jul 05 13:20:09 Reading random entropy from '/dev/urandom'

rgacogne:
Fixes #4105.

### DIFF
--- a/docs/markdown/recursor/settings.md
+++ b/docs/markdown/recursor/settings.md
@@ -511,7 +511,7 @@ to detect and act on infected hosts.
 Protobuf export to a server is enabled using the `protobufServer()` directive:
 
 ```
-protobufServer("192.0.2.1:4242" [[[[[, timeout], maxQueuedEntries], reconnectWaitTime], maskV4], maskV6])
+protobufServer("192.0.2.1:4242" [[[[[[, timeout], maxQueuedEntries], reconnectWaitTime], maskV4], maskV6], asynConnect])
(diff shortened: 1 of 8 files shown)


Habbie:
A quick read suggests the recursor might still hang (and not answer queries) for the set timeout, is that right?

rgacogne:
During startup, yes, to be able to display a message indicating that the server couldn't be reached. I guess I could add an option to not block at all.

rgacogne:
Added a `asyncConnect` option to prevent hanging during startup.

PR #4291, opened 2016-08-05
rec: Add an option to only send protobuf messages with a policy or tag set

### LINKED ISSUE #4109: filter on protobuf logging so only dns messages with a tag or preason get logged
for privacy purposes

rgacogne:
Closes #4109.

### DIFF
--- a/docs/markdown/recursor/settings.md
+++ b/docs/markdown/recursor/settings.md
@@ -522,7 +522,7 @@ to detect and act on infected hosts.
 Protobuf export to a server is enabled using the `protobufServer()` directive:
 
 ```
-protobufServer("192.0.2.1:4242" [[[[[[, timeout], maxQueuedEntries], reconnectWaitTime], maskV4], maskV6], asynConnect])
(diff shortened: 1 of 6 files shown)


docs/markdown/recursor/settings.md
```diff
@@ -511,7 +511,7 @@ to detect and act on infected hosts.
 Protobuf export to a server is enabled using the `protobufServer()` directive:
 
 ```
-protobufServer("192.0.2.1:4242" [[[[[, timeout], maxQueuedEntries], reconnectWaitTime], maskV4], maskV6])
+protobufServer("192.0.2.1:4242" [[[[[[, timeout], maxQueuedEntries], reconnectWaitTime], maskV4], maskV6], taggedOnly])
```
Habbie: I wonder if there is a more readable way to write 'these ordered arguments are all optional'. Alternatively we could make them named of course.

Habbie:
One nit/question that does not need to be tackled now. LGTM.

docs/markdown/recursor/settings.md
rgacogne: I need to find a way, because it's getting out of control in dnsdist!

PR #4293, opened 2016-08-06
Remove SELinux policy

rubenk:
I wrote new policy from scratch which has been upstreamed a while ago.


Habbie:
Thanks!

mikalsande:
Where was this policy upstreamed? I have not found it in either
 * RHEL 7.4 (searched policy on a RHEL 7.4 machine)
 * Fedora's reference policy https://github.com/fedora-selinux/selinux-policy/
 * Tresys' reference policy https://github.com/TresysTechnology/refpolicy-contrib/

If there is an available SELinux module for PowerDNS Autoritative and Recursor I would like to use them. If there is no readily available SELinux module for them I would like to contribute to making them.

rubenk:
https://github.com/fedora-selinux/selinux-policy-contrib/blob/rawhide/pdns.te

mikalsande:
Thank you @rubenk for pointing me in the right direction, I had only searched fedora-selinux/selinux-policy and not fedora-selinux/selinux-policy-contrib.

rubenk:
You're welcome @mikalsande. Note that this is just policy for the authoritative server. I never got around to creating policy for the recursor. So if you'd like to that that would be nice.

mikalsande:
I'll give it a try :) I need a policy for the recursor before I deploy it in production at $work, might as well learn how to make proper selinux modules that others can use instead of just simple inhouse modules.

rubenk:
Ok, best of luck and let me know if you need any help. The way to do it is to open a PR against that selinux-policy-contrib repo.

Eventually Fedora is supposed to upstream their policy to the refpolicy, but for some reason that hasn't happened in quite a while.

PR #4298, opened 2016-08-08
auth: Reset d_tsigprevious between UDP queries

### LINKED ISSUE #4259: Standalone TSIG works only for the first time request
**Input**
Given PDNS 4.0.1(master) configured with remote backend to accept DNS updates signed with TSIG, no AXFR configured. Equivalent bind configuration
```
key somekey. {
  algorithm hmac-md5;
  secret "MD5-BASE64-KEY";
};

zone "example.com." {
    type master;
    file "example.com.zone";
    allow-update { key somekey.; };
};
```

**Action**
Update domain request using `nsupdate`

```
nsupdate <<!
server 127.0.0.1 5300
zone example.com.
key somekey MD5-BASE64-KEY
update add  test.example.com 3600 A 127.0.0.1
send
!
```
**Outcome**
First request after server start goes through fine, getting all `startTransaction->feedRecord->replaceRRSet->commitTransaction`  chain.

Second run of same command(nsupdate started again and have new pid) gives following line in the log file
`Packet for domain 'example.com' denied: TSIG signature mismatch using 'somekey' and algorithm 'hmac-md5.sig-alg.reg.int'`. Different command does not give any affect, all caching is disabled.

**Problem**

DNS receiver uses same instance of `DNSPacket` for each request - [ref](https://github.com/PowerDNS/pdns/blob/master/pdns/common_startup.cc#L373)
`PacketHandler` calls `DNSPacket::setTSIGDetails` that saves current mac as `d_tsigprevious` - [ref](https://github.com/PowerDNS/pdns/blob/master/pdns/packethandler.cc#L1176). (This might be not actual place where it got moved in between requests)

Verification is started in [DNSPacket::checkForCorrectTSIG](https://github.com/PowerDNS/pdns/blob/master/pdns/dnspacket.cc#L655) which getting packet message from [DNSPacket:: getTSIGDetails] (https://github.com/PowerDNS/pdns/blob/master/pdns/dnspacket.cc#L475) that uses [makeTSIGMessageFromTSIGPacket](https://github.com/PowerDNS/pdns/blob/master/pdns/dnspacket.cc#L499) to compose actual message.

One the first request `d_tsigprevious` is not defined and signature verification is passed successfully. On the second request it's set to the previous request `d_mac` that's was previously saved on the same `DNSPacket` instance that was used for the first request. That [adds previous mac](https://github.com/PowerDNS/pdns/blob/master/pdns/dnssecinfra.cc#L607]) to the message and breaks signature verification since this message has nothing to do with original one any more. 

I'm not entirely aware about that previous key got added nor I understand why `DNSPacket` is not cleaned up properly so i can't provide a solution.

There is [related](https://mailman.powerdns.com/pipermail/pdns-users/2016-April/024140.html) report but I can't find corresponding issue for it.

rgacogne:
Otherwise we try to compute the TSIG signature using an unrelated MAC, as well described in #4259. I looked briefly at the other members of `DNSPacket` and didn't see any other obvious reuse issue. The TCP path is not affected as it doesn't reuse `DNSPacket`.
Also remove the unused `DNSPacket::d_qlen` member.
Fixes #4259.

### DIFF
--- a/pdns/dnspacket.cc
+++ b/pdns/dnspacket.cc
@@ -68,7 +68,6 @@ DNSPacket::DNSPacket()
   d_havetsig = false;
   d_socket = -1;
   d_maxreplylen = 0;
-  d_qlen = 0;
   d_tsigtimersonly = false;
   d_haveednssection = false;
 }
@@ -96,7 +95,6 @@ DNSPacket::DNSPacket(const DNSPacket &orig)
   DLOG(L<<"DNSPacket copy constructor called!"<<endl);
   d_socket=orig.d_socket;
   d_remote=orig.d_remote;
-  d_qlen=orig.d_qlen;
   d_dt=orig.d_dt;
   d_compress=orig.d_compress;
   d_tcp=orig.d_tcp;
--- a/pdns/dnspacket.cc
+++ b/pdns/dnspacket.cc
@@ -470,6 +470,11 @@ void DNSPacket::setTSIGDetails(const TSIGRecordContent& tr, const DNSName& keyna
   d_tsigtimersonly=timersonly;
 }
 
+void DNSPacket::resetTSIGPrevious()
+{
+  d_tsigprevious.clear();
+}
+
 bool DNSPacket::getTSIGDetails(TSIGRecordContent* trc, DNSName* keyname, string* message) const
 {
   MOADNSParser mdp(d_rawpacket);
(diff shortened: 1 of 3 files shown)


mind04:
Different solution https://github.com/mind04/pdns/commit/5ae00bbc5b5df489f55c698dc539b8e6d59ac9c4
The problem is we use d_tsigprevious for questions.

rgacogne:
I was reluctant to modify the existing behavior of `DNSPacket::getTSIGDetails()`, but it looks like it's never used in a case where it needs the previous sig value, so I think your solution is better. Did you open a PR with it?

PR #4300, opened 2016-08-08
allow newDN to take a DNSName in; document missing methods

Habbie:
I expect this could also be done with `boost::variant`; in fact I'm not 100% sure I'm not overwriting the first definition here.


zeha:
> in fact I'm not 100% sure I'm not overwriting the first definition here.

sounds like you want a ("unit") test.

Habbie:
Maybe this needs docs. Maybe it's obvious this should work - it was to me!

rgacogne:
I'm pretty sure the first definition is overwritten. Based on [1], it looks like the only option is using `boost::variant`.

[1]: https://github.com/PowerDNS/luawrapper#polymorphic-functions

Habbie:
Changed to use boost::variant.

Habbie:
Ready for review.

rgacogne:
LGTM

PR #4306, opened 2016-08-11
limit size of mysql cell to 128 kilobytes

Habbie:
This avoids a crash when somebody uses LONGTEXT in their db schema (which may in the past have been given as advice by us, in the days when content was a `varchar(255)`!)


Habbie:
Reported by @imtek. Thanks!

PR #4309, opened 2016-08-12
Add showTCPStats function

RobinGeuze:
Add a handy function, allowing you to see the currently set values for the TCP connection settings and the amount of currently queued requests and connections. This is handy when determining the ideal settings for your specific use case.


rgacogne:
Thank you! Could you just add the new function to the README as well? I really need to find a way to get rid of that duplication somehow..

RobinGeuze:
I added it in the list of commands in a likely place in the README.md file, dunno if we want more documentation?

rgacogne:
That's perfect, thank you!

PR #4319, opened 2016-08-16
pipe: SERVFAIL when needed

### LINKED ISSUE #4308: Change in behaviour of pipe backend protocol
When updating a server from v3 to v4 of authoritative PDNS, I've noticed the the behaviour of FAIL in the protocol has changed.  In v3, a FAIL would stop the server from progressing to the next backend, and would return SRVFAIL to a client.  Now, it seems that the backend in simply killed, and the server continues on to the next backend.  

I'm not sure if this was an accidental feature that we happened to use, but the ability for the pipe backend to distinguish between "I don't know, ask someone else" and "The answer definitely doesn't exist, stop looking" was important for us.

Is this a bug in the new version, or the intended behaviour?

pieterlexis:
* SERVFAIL, but don't restart the coprocess if we get a FAIL.
 * SERVFAIL if the coprocess sends something we don't understand.
   * In this case, restart the coprocess.

Closes #4308

### DIFF
--- a/modules/pipebackend/pipebackend.cc
+++ b/modules/pipebackend/pipebackend.cc
@@ -324,10 +324,14 @@ bool PipeBackend::get(DNSResourceRecord &r)
         throw PDNSException("Coprocess backend sent incorrect response '"+line+"'");
     }
   }
+  catch (DBException &dbe) {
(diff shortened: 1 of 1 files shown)


pieterlexis:
ready for review

PR #4322, opened 2016-08-16
For those who build on ubuntu minimal server

kocoten1992:


pieterlexis:
Thanks for this patch. However, `dh-autoreconf` is a debian-specific debhelper package. On line 42 of the README, the following is stated:

>  When building from git, the following packages are also required: autoconf, automake,
 ragel, bison and flex, [...]

Is this perhaps in the wrong place and easy to miss?

kocoten1992:
I really miss that, guess I'm just reading too fast to get it done :smiley: , but that package was special than the rest, when other are missing, there is error hint to user which pack should be install (I don't even google the correct name of the pack, first try), that one wasn't, it come from google :smile:

pieterlexis:
These are the 'joys' of autotools :).

pieterlexis:
ah yes, before I forget, we provide packages with the latest version for ubuntu on https://repo.powerdns.com

kocoten1992:
@pieterlexis , people don't read doc before install, that one should definitely need on README, and thank you :+1:

PR #4324, opened 2016-08-17
Allow Lua access to the result of the Policy Engine decision, skip RPZ

rgacogne:
This PR includes and supersedes #4236 with:
* a new `prerpz` hook allowing to completely disable the processing filtering policies by setting `wantsRPZ` to false, or to disable only selected policies by using `dq:discardPolicy(policyname)`
* fix the recursor regression tests
* actually fails if the recursor regression tests fail

### DIFF
--- a/pdns/filterpo.cc
+++ b/pdns/filterpo.cc
@@ -39,24 +39,21 @@ bool findNamedPolicy(const map<DNSName, DNSFilterEngine::Policy>& polmap, const 
                     *.
    */
 
-  bool first=true;
   map<DNSName, DNSFilterEngine::Policy>::const_iterator iter;
-  do {
-    if(first) {
-      iter = polmap.find(s);
-      if(iter != polmap.end()) {
-        pol=iter->second;
-        return true;
-      }
-    }
-    first=false;
+  iter = polmap.find(s);
 
+  if(iter != polmap.end()) {
+    pol=iter->second;
+    return true;
+  }
+
+  while(s.chopOff()){
     iter = polmap.find(DNSName("*")+s);
     if(iter != polmap.end()) {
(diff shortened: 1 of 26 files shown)


pdns/pdns_recursor.cc
```diff
@@ -730,11 +730,8 @@ void startDoResolve(void *p)
     if(!g_quiet || tracedQuery) {
       L<<Logger::Warning<<t_id<<" ["<<MT->getTid()<<"/"<<MT->numProcesses()<<"] " << (dc->d_tcp ? "TCP " : "") << "question for '"<<dc->d_mdp.d_qname<<"|"
        <<DNSRecordContent::NumberToType(dc->d_mdp.d_qtype)<<"' from "<<dc->getRemote();
-#ifdef HAVE_PROTOBUF
```
pieterlexis: @rgacogne Judging by our IRC convo on ECS a few days ago, this will break when compiling without on non-protobuf. Could you revert this commit?

rgacogne:
Good catch, reverted.

pdns/filterpo.cc
```diff
@@ -11,30 +11,30 @@ bool findNamedPolicy(const map<DNSName, DNSFilterEngine::Policy>& polmap, const
...
-    if(!first) {
-      iter = polmap.find(DNSName("*")+s);
+    if(first) {
+      iter = polmap.find(s);
       if(iter != polmap.end()) {
-	pol=iter->second;
-	return true;
+        pol=iter->second;
+        return true;
       }
```
pieterlexis: needs an `s.chopOff()` in the `if(first)` block.. otherwise we will test *.full.domain.name for full.domain.name as well

PR #4329, opened 2016-08-17
Add options to LogAction() so it can append (instead of truncate) and…

wessels:
… write without buffering.

Was so happy to discover LogAction() but disappointed that it truncates
the file each time the process starts, and apparently doesn't flush the
file upon exiting.  With two new optional args, the file can be appended
rather than truncated and buffering disabled.  The defaults remain as
before, which is to say to truncate and to buffer.

### DIFF
--- a/pdns/README-dnsdist.md
+++ b/pdns/README-dnsdist.md
@@ -1257,7 +1257,7 @@ instantiate a server with additional parameters
     * `DelayAction(milliseconds)`: delay the response by the specified amount of milliseconds (UDP-only)
     * `DisableValidationAction()`: set the CD bit in the question, let it go through
     * `DropAction()`: drop these packets
-    * `LogAction([filename], [binary])`: Log a line for each query, to the specified file if any, to the console (require verbose) otherwise. When logging to a file, the `binary` optional parameter specifies whether we log in binary form (default) or in textual form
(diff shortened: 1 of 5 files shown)


pdns/README-dnsdist.md
```diff
@@ -1257,7 +1257,7 @@ instantiate a server with additional parameters
     * `DelayAction(milliseconds)`: delay the response by the specified amount of milliseconds (UDP-only)
     * `DisableValidationAction()`: set the CD bit in the question, let it go through
     * `DropAction()`: drop these packets
-    * `LogAction([filename], [binary])`: Log a line for each query, to the specified file if any, to the console (require verbose) otherwise. When logging to a file, the `binary` optional parameter specifies whether we log in binary form (default) or in textual form
+    * `LogAction([filename], [binary])`: Log a line for each query, to the specified file if any, to the console (require verbose) otherwise. When logging to a file, the `binary` optional parameter specifies whether we log in binary form (default) or in textual form, the `append` optional parameter specifies whether we open the file for appending or truncate each time (default), and the `buffered` optional parameter specifies whether writes to the file are buffered (default) or not.
```
rgacogne: The parameters have not been updated in the prototype here.

pdns/dnsrulactions.hh
```diff
@@ -913,6 +918,8 @@ private:
   string d_fname;
   FILE* d_fp{0};
   bool d_binary{true};
+  bool d_append{false};
+  bool d_buffered{true};
```
rgacogne: It looks like we don't need to keep `d_append` and `d_buffered` around, since they seem to be only used in the constructor.

pdns/README-dnsdist.md
wessels: Thanks, I missed that.  Its added now.

pdns/dnsrulactions.hh
wessels: Yes, thanks for the suggestion.

rgacogne:
Thank you!

PR #4330, opened 2016-08-18
auth: Add a `query-non-local-bind` option

### LINKED ISSUE #4299: Start of pdns_server fails with non-local-bind *and* query-local-address
We would like to start pdns_serever (4.0.1) with `non-local-bind=yes`.  But it fails if we also have `query-local-address=...` set.

`Exiting because communicator thread died with error: Resolver binding to local UDP socket on ...: Cannot assign requested address`

We define `query-local-address` and `query-local-address6` only for AXFRs. We have not set the `recursor` directive.

rgacogne:
This allows using a non-local address for `query-local-address`
or `query-local-address6`. This only makes sense if no query
is going to be sent before the address comes up, otherwise it will
fail.
Fixes #4299.

### DIFF
--- a/docs/markdown/authoritative/settings.md
+++ b/docs/markdown/authoritative/settings.md
@@ -632,6 +632,16 @@ system uses by default for outgoing packets.
 
 Source IP address for sending IPv6 queries.
 
+## `query-non-local-bind`
+* Boolean
+* Default: no
+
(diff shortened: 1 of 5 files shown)


rgacogne:
Replaced by #4332.

PR #4347, opened 2016-08-22
change default for any-to-tcp to yes

mind04:


regression-tests/backends/gmysql-master
```diff
@@ -23,9 +23,11 @@ gmysql-dbname=$GMYSQLDB
 gmysql-user=$GMYSQLUSER
 gmysql-host=$GMYSQLHOST
 gmysql-password=$GMYSQLPASSWD
+
+any-to-tcp=no
```
Habbie: Why doesn't this make the new any-to-tcp-query test fail?

regression-tests/backends/gmysql-master
```diff
@@ -23,9 +23,11 @@ gmysql-dbname=$GMYSQLDB
...
 __EOF__
 
-		gsql-master gmysql
+		gsql-master gmysql dyndns
```
Habbie: Oh, because of this. Bit of a hack

mind04: There are many ANY queries in the  dyndns tests.
And testing dyndns only via tcp feels wrong because you can't detect truncation issues with tcp

Habbie: Ah, like that. I always get confused about skipreasons. So we test mysql(+dyndns) with any-to-tcp off, and skip the actual any-to-tcp test; then for all other backends, we don't test dyndns (we never did) and we do test any-to-tcp. Sounds good.

PR #4351, opened 2016-08-23
RPZ: Use query-local-address(6) by default

### LINKED ISSUE #4343: RPZ make transfer source address configurable
[Apparently](https://github.com/PowerDNS/pdns/blob/ff6ca337e2f0d150a95a68a4fd9c6e20a87c6b1c/pdns/rpzloader.cc#L119) it is currently not possible to specify the IXFR transfer source address for RPZ . It would be nice if this could either be configured in `rpzMaster()` or alternatively if it honored `query-local-address` / `query-local-address6`

pieterlexis:
Add localAddress to rpzMaster options to allow override.

Fixes #4343


rgacogne:
Nice, a quick look at `RPZIXFRTracker()` suggests we need the same thing for the IXFR part.

pieterlexis:
Added the same support to `RPZIXFRTracker()`

PR #4359, opened 2016-08-23
Several documentation fixes

pieterlexis:


docs/markdown/recursor/settings.md
```diff
@@ -232,6 +232,13 @@ This setting can be used to expand or reduce the limitations.
 This is the value set for the EDNS0 buffer size in outgoing packets.
 Lower this if you experience timeouts.
 
+## `edns-subnet-whitelist`
+* Comma separated list of domain names and netmasks
+* Default: (none)
+
+List of netmasks and domains that [EDNS Client Subnet](https://tools.ietf.org/html/rfc7871) should be enabled for.
+By default, this option is empty, meaning no EDNS Client Subnet information is sent.
+
```
rgacogne: It's not obvious to me how exactly this option works with this description, how about:

> List of netmasks and domains that [EDNS Client Subnet](https://tools.ietf.org/html/rfc7871) should be enabled for in outgoing queries.
> For example, an EDNS Client Subnet option containing the address of the initial requestor will be added to an outgoing query sent to server 192.0.2.1 for domain X if 192.0.2.1 matches one of the supplied netmasks, or if X matches one the supplied domains. The initial requestor address will be truncated to 24 bits for IPv4 and to 56 bits for IPv6, as recommended in the privacy section of [rfc7871](https://tools.ietf.org/html/rfc7871#section-11.1).
> By default, this option is empty, meaning no EDNS Client Subnet information is sent.

docs/markdown/recursor/settings.md
```diff
@@ -232,6 +232,15 @@ This setting can be used to expand or reduce the limitations.
 This is the value set for the EDNS0 buffer size in outgoing packets.
 Lower this if you experience timeouts.
 
+## `edns-subnet-whitelist`
+* Comma separated list of domain names and netmasks
+* Default: (none)
+
+List of netmasks and domains that [EDNS Client Subnet](https://tools.ietf.org/html/rfc7871) should be enabled for in outgoing queries.
+For example, an EDNS Client Subnet option containing the address of the initial requestor will be added to an outgoing query sent to server 192.0.2.1 for domain X if 192.0.2.1 matches one of the supplied netmasks, or if X matches one the supplied domains.
```
rgacogne: >  or if X matches one **of** the supplied domains.

A word was missing in the text I suggested, sorry about that..

PR #4363, opened 2016-08-25
pdnsutil: Added change-slave-zone-master to change the master for a slave zone

hnsk:
A new pdnsutil command to change master server for a slave zone.

### DIFF
--- a/contrib/pdnsutil.bash_completion.d
+++ b/contrib/pdnsutil.bash_completion.d
@@ -15,7 +15,7 @@ have pdnsutil && {
   _pdnsutil_helper_local_() {
     local cur prev cmd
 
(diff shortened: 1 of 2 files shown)


pieterlexis:
In the help, you state that the user can add more than one master. However in `changeSlaveZoneMaster()`, I don't see an iteration through `cmd`, adding these addresses to a string and then passing it to `setMaster()`

hnsk:
I actually just pretty much copied the  create-slave-zone functionality which also states that you can define multiple master-ip's but it doesn't seem to work for that either. Maybe both should be changed in the help just to:

change-slave-zone-master ZONE master-ip
create-slave-zone ZONE master-ip

Or can a slave zone actually have multiple masters?

pieterlexis:
There is support for this, the addresses have to be comma-separated: https://doc.powerdns.com/md/authoritative/backend-generic-sql/#slave-operation

PR #4373, opened 2016-08-27
Revamp label compression code which (+ some cleanups) speeds up large packet creation by ~40%

ahupowerdns:
This is a big change that comes with a number of new tests. 

After:
'write big referral' 0.10 seconds: 67707.6 runs/s, 14.77 usec/run
Before:
'write big referral' 0.10 seconds: 42556.4 runs/s, 23.50 usec/run

A 37% improvement. 

Impact on root-server performance:
rootplot

Is in line with these numbers - we now have good service in 4 cores up to around 80kqps, trouble used to start a bit beyond 40kqps. This graph has 60% packet cacheability.

### DIFF
--- a/pdns/dnswriter.cc
+++ b/pdns/dnswriter.cc
@@ -29,6 +29,15 @@
 
 #include <limits.h>
 
+/* d_content:                                      <---- d_stuff ---->
+                                      v d_truncatemarker  
+   dnsheader | qname | qtype | qclass | {recordname| dnsrecordheader | record }
+                                        ^ d_rollbackmarker           ^ d_sor 
+    
+
+*/
+
+
 DNSPacketWriter::DNSPacketWriter(vector<uint8_t>& content, const DNSName& qname, uint16_t  qtype, uint16_t qclass, uint8_t opcode)
   : d_content(content), d_qname(qname), d_canonic(false), d_lowerCase(false)
 {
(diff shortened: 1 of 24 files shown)


pdns/dnswriter.cc
```diff
@@ -183,145 +162,198 @@ void DNSPacketWriter::xfr8BitInt(uint8_t val)
...
+        }
+        if(!c)
+          break;
+        pvect.push_back(iter - d_content.cbegin());
+        iter+=*iter+1;
+      }
+    }
+    catch(std::bad_alloc& ba) {
+      if(l_verbose)
+        cout<<"Domain "<<name<<" too large to compress"<<endl;
```
rgacogne: If I'm not mistaken, `name` is not the culprit here, the message might be misleading.

pdns/dnswriter.cc
```diff
@@ -183,145 +162,198 @@ void DNSPacketWriter::xfr8BitInt(uint8_t val)
...
+  for(auto p : d_namepositions) {
+    if(l_verbose) {
+      cout<<"Pos: "<<p<<", "<<d_content.size()<<endl;
+      DNSName pname((const char*)&d_content[0], d_content.size(), p, true); // only for debugging
+      cout<<"Looking at '"<<pname<<"' in packet at position "<<p<<"/"<<d_content.size()<<", option "<<counter<<"/"<<d_namepositions.size()<<endl;
+      ++counter;
+    }
+    // memcmp here makes things _slower_
+    pvect.clear();
+    try {
```
rgacogne: I wonder if we could compute `pvect` once when inserting a new entry into `d_namepositions` and store it instead.

rgacogne:
Looks fine!

PR #4389, opened 2016-09-01
Switch DNSSEC to using zone-cuts to find key material

pieterlexis:
This PR also generalizes the getDenial function so it can easier be used in the future to verify denials for non-DS records.

Needs a thorough review

### DIFF
--- a/pdns/validate.cc
+++ b/pdns/validate.cc
@@ -237,20 +237,22 @@ vState getKeysFor(DNSRecordOracle& dro, const DNSName& zone, keyset_t &keyset)
     }
   }
 
-  vector<string> labels = zone.getRawLabels();
-
-  dsmap_t dsmap;
   keyset_t validkeys;
+  dsmap_t dsmap;
 
-  DNSName qname = lowestTA;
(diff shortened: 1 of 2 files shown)


pdns/validate.cc
```diff
@@ -331,163 +334,152 @@ vState getKeysFor(DNSRecordOracle& dro, const DNSName& zone, keyset_t &keyset)
...
-            dsmap.insert(*dsrc);
-            // dotEdge(keyqname,
-            //         "DNSKEY", keyqname, ,
-            //         "DS", qname, std::to_string(dsrc.d_tag));
-            // cout<<"    "<<dotEscape("DNSKEY "+keyqname)<<" -> "<<dotEscape("DS "+qname)<<";"<<endl;
+        else if(v.first.second==QType::NSEC3) {
+          for(const auto& r : v.second.records) {
+            LOG("\t"<<r->getZoneRepresentation()<<endl);
+
+            auto nsec3 = std::dynamic_pointer_cast<NSEC3RecordContent>(r);
```
rgacogne: We should probably check that `nsec3` is not `NULL`, we do the check for `nsec` above.

PR #4391, opened 2016-09-01
DNSSEC: Actually follow RFC 7646 §2.1

pieterlexis:
We were off by one when counting labels, so when an NTA was added for a
name where a TA was configured, we would still attempt validation.

Reported by @jpmens


pieterlexis:
now with a test

PR #4398, opened 2016-09-01
Warn the user when DNSSEC related things are attempted when DNSSEC is disabled

### LINKED ISSUE #4390: Recursor accepts addDS() with dnssec=off
Recursor 4.0.2 accepts `addDS()` from the Lua configuration while `dnssec=off`. 
It would be nice if we could have a warning for this.

pieterlexis:
Fixes #4390

### DIFF
--- a/pdns/rec-lua-conf.cc
+++ b/pdns/rec-lua-conf.cc
@@ -15,6 +15,7 @@
 #include "base64.hh"
 #include "remote_logger.hh"
 #include "validate.hh"
+#include "validate-recursor.hh"
 #include "root-dnssec.hh"
 
 GlobalStateHolder<LuaConfigItems> g_luaconfs; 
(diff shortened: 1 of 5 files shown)


rgacogne:
Perhaps you could also warn when `addNTA()` is called with DNSSEC disabled?

pieterlexis:
added!

pdns/rec-lua-conf.cc
```diff
@@ -233,26 +234,34 @@ void loadRecursorLuaConfig(const std::string& fname)
 		    });
 
   Lua.writeFunction("addDS", [&lci](const std::string& who, const std::string& what) {
+      if(g_dnssecmode == DNSSECMode::Off)
+        theL()<<Logger::Warning<<"Warning: adding Trust Anchor for DNSSEC (addDS), but dnssec is set to 'off'!"<<endl;
```
Habbie: I see this code duplicated 3 more times with minor edits, maybe make a function?

pdns/rec_channel_rec.cc
```diff
@@ -348,6 +348,9 @@ string doSetCarbonServer(T begin, T end)
 template<typename T>
 string doSetDnssecLogBogus(T begin, T end)
 {
+  if(g_dnssecmode == DNSSECMode::Off)
+    return "DNSSEC is disabled in the configuration, not changing the Bogus logging setting\n";
```
Habbie: more code duplication

PR #3150, opened 2016-01-02
pdnsutil: complain about missing quotes around TXT record data

zeha:
Need this to match API behaviour.

Found by @stbuehler.

Issue #3503, opened 2016-03-03
pdns auth TXT parsing issues


hlindqvist:
There appear to be some issues with how `TXT` record data is treated (tested with pdns auth 3.4.8).

https://doc.powerdns.com/md/types/ expands a bit on the `TXT` handling over https://doc.powerdns.com/3/types/, but for what isn't covered there my fallback was _master file format_ expectations.

My general impression is that the `content` for records in the pdns database is in the format for the RRData part of the record from the _master file format_ specification as per the relevant RFCs?

If that is not the case, I still think that there are still issues, but maybe some of the observed behavior is expected and should just be clarified in documentation?

(This is possibly related to eg https://github.com/PowerDNS/pdns/pull/3061, where the discussion and tentative code inspired me to poke at how pdns deals with `TXT` RRs in general.)

Relevant data from database:

```
sqlite> SELECT * FROM records WHERE domain_id=2;
1005|2|example.com|SOA|ns.example.com. hostmaster.example.com. 7 3600 1800 3600000 7200|7200|||0||1
1006|2|test.example.com|TXT|"foo bar" baz|7200|||0||1
1007|2|test2.example.com|TXT|foo "bar baz"|7200|||0||1
1252|2|test3.example.com|TXT|"foo" "bar baz"|7200|||0||1
1253|2|test4.example.com|TXT|foo bar baz|7200|||0||1
1254|2|test5.example.com|TXT|"foo" bar baz|7200|||0||1
sqlite>
```

Below are my findings (all record data formatted like dig prints it, ie each string value is enclosed in quotation marks even in cases where quotation marks would not be necessary):

test.example.com works as expected:

```
test.example.com.       7200    IN      TXT     "foo bar" "baz"
```

test2.example.com is **answered with SERVFAIL**

expected for test2.example.com:

```
test2.example.com.      7200    IN      TXT     "foo" "bar baz"
```

test3.example.com works as expected:

```
test3.example.com.      7200    IN      TXT     "foo" "bar baz"
```

test4.example.com returns the **wrong result**:

```
test4.example.com.      7200    IN      TXT     "foo bar baz"
```

expected for test4.example.com:

```
test4.example.com.      7200    IN      TXT     "foo" "bar" "baz"
```

test5.example.com is **answered with SERVFAIL**

expected for test5.example.com:

```
test5.example.com.      7200    IN      TXT     "foo" "bar" "baz"
```

(The above tests are surely not comprehensive but it was a few variations that led to interesting results.)

PR #4406, opened 2016-09-02
Warn about TXT content without ""

azlev:
TXT records not enclosed by double quotes break the PowerDNS API and pdnsutil edit-zone. If they are not mandatory, at least emit a warning about that.


zeha:
If this makes it, please close #3150

azlev:
Sorry zeha. I didn't notice your PR. I wrote the same patch.

Habbie:
Closing; tracking issue in #3503

PR #4408, opened 2016-09-02
Authoritative: do not rely on webserver=yes if api=yes

zeha:
Alternative solution for #4290


zeha:
Not for 4.0 obviously.

zeha:
Rebased on top of current master to resolve conflicts.

PR #4429, opened 2016-09-06
Recursor 4.0.3 changelog

pieterlexis:


rgacogne:
Shouldn't #4388 and #4391 be in there too?

pieterlexis:
Both were indeed backported, thanks!

PR #4432, opened 2016-09-06
dnsdist: Add regression tests for protobuf export and DNSResponse bindings

rgacogne:
Because one regression is one too many.


rgacogne:
Added regression tests for protobuf alteration (anonymization) via Lua, along with missing `DNSResponse` bindings since LuaWrapper doesn't support inheritance.

PR #4446, opened 2016-09-08
fix version check for boost 1.61 fcontext include

anthraxx:
This fixes an issue where the wrong (pre boost 1.61 header) was
included as the check only evaluates to true after 1.61 instead if
reflecting what the autoconf changes are checking: greater-equals.

Additionally make sure to include the boost version header before the
version check as it may not be defined yet.

This issue was introduced via a30361f

### DIFF
--- a/pdns/mtasker_fcontext.cc
+++ b/pdns/mtasker_fcontext.cc
@@ -23,12 +23,12 @@
 #include <exception>
 #include <cassert>
 #include <type_traits>
-#if BOOST_VERSION > 106100
+#include <boost/version.hpp>
+#if BOOST_VERSION >= 106100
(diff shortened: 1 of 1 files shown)


pieterlexis:
Hi, with this patch, PowerDNS doesn't compile against boost 1.61, I've made this before. The API for `boost::context::detail::jump_fcontext()` has also changed (unfortunately). This means we need to redo the context switching for boost 1.61.

anthraxx:
@pieterlexis sure it was just to correct the include, it doesn't compile either way because of the new API. just don't forget to take the useful part out of this and use ```BOOST_VERSION >= 106100``` :smirk_cat:

PR #4457, opened 2016-09-11
fix incomplete commit for pdns_control (re)notify

klaus3000:
In #3889 it was revealed that #3663 was incomplete. This pull request handles 'notify *' correctly and verifies that 'slave' is enabled.

### DIFF
--- a/pdns/dynhandler.cc
+++ b/pdns/dynhandler.cc
@@ -297,19 +297,41 @@ string DLNotifyHandler(const vector<string>&parts, Utility::pid_t ppid)
     vector<DomainInfo> domains;
     B.getAllDomains(&domains);
 
-    int total = 0;
-    int notified = 0;
-    for (vector<DomainInfo>::const_iterator di=domains.begin(); di != domains.end(); di++) {
-      if (di->kind == 0) { // MASTER
-        total++;
-        if(Communicator.notifyDomain(di->zone))
-          notified++;
+    int master = 0, master_notified = 0;
+    int slave = 0, slave_notified = 0;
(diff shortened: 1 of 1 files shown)


pieterlexis:
looks good to me. @mind04 ?

mind04:
The logic is fine. 
But I'm no fan of the insane verbose response, and the truckload of extra code to generate it.
```
10 out of 11 MASTER zones added to queue - see log. 1 out of 2 SLAVE zones added to queue - see log.
```
Seems nice, but only tells you "11 out of 13 domains notified, check your log for details".

Bottom line is, if you care you have to read the logs.

klaus3000:
Indeed, there is some code to generate that kind of log. But that kind of log was there before - I only extended it to be "backwards compatible". I did not wanted to change anything, but fix it.

klaus3000:
Ping - almost one year for a simple fix. as mentioned. The unesthetic log message was there before.

klaus3000:
Any plans to commit this? I am cleaning up my pull requests. So please commit or close forever.

Habbie:
The first commit looks good; does the second commit fix anything or does it just add logging?

klaus3000:
The second commit is somehow "academic". When "notify *" is used, it checks if PowerDNS is configured as master or slave (or both) and then notifys only the master, slave or both domains. It also gives log output of how many master and how many slave domains were queued for notifications.

Habbie:
@klaus3000 thank you for this. I'm closing this PR in favour of #6691 which contains one of the commits.

PR #4459, opened 2016-09-11
check carefully: plugs memory leak in postgresql backend

ahupowerdns:
eleksir noted that we leak a ton of memory in postgresql. I'm no postgres expert, but this plugs my leak and still appears to function. In other news, do we need a transaction for every query?

### DIFF
--- a/modules/gpgsqlbackend/spgsql.cc
+++ b/modules/gpgsqlbackend/spgsql.cc
@@ -77,7 +77,8 @@ public:
       L<<Logger::Warning<<"Query: "<<d_query<<endl;
     }
     if (!d_parent->in_trx()) {
-      PQexec(d_db(),"BEGIN");
+      auto res=PQexec(d_db(),"BEGIN");
+      PQclear(res);
       d_do_commit = true;
     } else d_do_commit = false;
     d_res_set = PQexecPrepared(d_db(), d_stmt.c_str(), d_nparams, paramValues, paramLengths, NULL, 0);


ahupowerdns:
'You can keep a PGresult object around for as long as you need it; it does not go away when you issue a new command, nor even if you close the connection. To get rid of it, you must call PQclear. Failure to do this will result in memory leaks in your application.' - postgresql docs appear to confirm this is needed.

cmouse:
LGTM.

ahupowerdns:
I have checked the code, this is the only place where we make this mistake. I do note that the code looks 'leak prone'.

mschout:
+1 This bit us hard as well.  We applied the patch and things seem much better now.  As someone who has worked with libpq for 20 years, I can tell you that calling PQexec() without calling PQclear() on the result will definitely leak memory.  Only difference in the patch we applied locally to what was merged was to use "PGresult *res" instead of "auto res" just to be consistent with the style used elsewhere in that file.

Would love to see this fixed in 4.0.2 as this was a big problem for us since moving to 4.0

mschout:
Just adding some info here as evidence this patch fixes a leak.  Without the applied patch, dtrace on PQexec(), PQexecPrepared(), and PQclear() entries:

```
  PQexecPrepared                                                  702
  PQexec                                                         1404
  PQclear                                                        1408
```

The total PQclear()'s should be equal to the number of PQexec*()'s.  

After applying the patch:

```
  PQexecPrepared                                                  396
  PQexec                                                          792
  PQclear                                                        1188
```

Which is exactly as it should be.

PR #4461, opened 2016-09-12
Fall back to SystemV ucontexts on boost >= 1.61

pieterlexis:


Habbie:
With this patch, both Boost 1.60 (fcontext) and 1.61 (ucontext) work for me with the recursor on OSX.

PR #4478, opened 2016-09-15
docs enhancements; clarify error message when set-presigned fails with DNSSEC disabled

peterthomassen:
from IRC:

(12:04:06 PM) Habbie: wonder if that error from pdnsutil could be more helpful
(12:04:46 PM) Habbie: ah, indeed that error means -exactly- that gmysql-dnssec is not set
(12:04:57 PM) Habbie: every other error condition will give you an actual error from the backend i believe


peterthomassen:
regarding 8ad575f, from IRC:

(01:09:05 PM) pt01: During AXFR, pdns does rectify. Is this when acting as a (signing) slave (after receiving the zone), or when acting as a master (before signing)?
(01:09:38 PM) pt01: I am asking because docs:backend-generic-sql/#handling-dnssec-signed-zones says that the AXFR _client_ code does that, while the disable-axfr-rectify documentation says it's for _outgoing_ AXFR.
(01:10:19 PM) Habbie: the incoming code always does it
(01:10:40 PM) Habbie: and the outgoing code does it unless you set that disable, which you shouldn't :)
(01:15:28 PM) pt01: aha! so, on a master, I never need to rectify myself unless there are non-AXFR queries?
(01:15:50 PM) Habbie: correct

docs/markdown/authoritative/backend-generic-sql.md
```diff
@@ -100,16 +100,17 @@ this implies that a slaved domain will not stay disabled.
...
+`change_date` field in the "records" table).
 
 
 # Handling DNSSEC signed zones
 To enable DNSSEC processing, the `backend-dnssec` option must be set to 'yes'.
 
 ## Rules for filling out DNSSEC fields
 Two additional fields in the 'records' table are important: 'auth' and 'ordername'.
-These fields are set correctly on an incoming zone transfer, and also by running
+These fields are set correctly on a zone transfer, and also by running
```
mind04: the fields are only set during an incoming axfr, so the original text is more accurate

docs/markdown/authoritative/backend-generic-sql.md
```diff
@@ -136,7 +137,7 @@ encoded representation of the salted & iterated hash of the full record name.
 
 In addition, PowerDNS fully supports empty non-terminals. If you have a zone
 example.com, and a host a.b.c.example.com in it, rectify-zone (and the AXFR
-client code) will insert b.c.example.com and c.example.com in the records table
+code) will insert b.c.example.com and c.example.com in the records table
```
mind04: no record insertion during outgoing axfr

docs/markdown/authoritative/dnssec.md
```diff
@@ -218,7 +218,7 @@ In this way, if keying material is available for an unsigned zone that is
 retrieved from a master server, this keying material will be used when serving
 data from this zone.
 
-As part of the zone retrieval, the equivalent of `pdnsutil rectify-zone` is run
+As part of the zone transfer, the equivalent of `pdnsutil rectify-zone` is run
```
mind04: transfer out rectify does not update the db like pdnsutil rectify-zone

peterthomassen:
The information that I tried to add to the docs is that even an outgoing AXFR does some sort of rectify, as the outgoing AXFR contains signed NSEC3 records for empty non-terminals. This information is nowhere else to be found. The documentation therefore currently implies that the rectify is fully done by the receiving slave (including signing NSEC3 records for ENT), which requires access to keying material on the slave. However, this is not the case.

It was certainly misguided to modify backend-generic-sql.md, so I changed things so that this information is now in dnssec.md only.

PR #4494, opened 2016-09-22
Fix struct/class mismatch warning reported by apple clang 800.0.38

zeha:


rgacogne:
LGTM

pieterlexis:
no errors/warnings for misc.hh/cc on Linux with GCC 6.2

PR #4495, opened 2016-09-22
Extend zone2ldap to add the domain info attributes

goestreicher:
This adds two new flags to the command: --pdns-info to add the
attributes from the pdns-domaininfo schema and --domainid to
give the domain id of the first domain found in the file.


pieterlexis:
Awesome! Could you update the manpage (doc/manpages/zone2ldap.md) as well?

goestreicher:
I'm withdrawing this PR, because it'll be included, with many more goodies, in an upcoming PR to add DNSSEC in the LDAP backend

PR #4499, opened 2016-09-23
rec: Add support for boost::context >= 1.61

rgacogne:
Most of the pain comes from the fact that `jump_fcontext()` doesn't take a pointer to store the context we are switching from anymore, but instead make it available as a return value to the context we are switching to. Therefore we need to pass a pointer to the context we are switching to so it can store the context we are switching from into it.

### DIFF
--- a/pdns/mtasker_fcontext.cc
+++ b/pdns/mtasker_fcontext.cc
@@ -23,10 +23,15 @@
 #include <exception>
 #include <cassert>
 #include <type_traits>
-#include <boost/context/fcontext.hpp>
 #include <boost/version.hpp>
-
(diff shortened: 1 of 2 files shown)


Habbie:
Tested on OSX with 1.60 and 1.61, works perfectly!

zeha:
WFM on 10.12.

zeha:
style nit: having only one 'direction' for BOOST_VERSION >/<= makes reviewing, reading and later cleanup easier.

rgacogne:
@zeha Fixed, thank you!

olk:
why not using boost::context::execution_context<> - powerDNS seams to require C++11, so it will fit

PR #4504, opened 2016-09-24
update packetcache size in insert(), remove unnecessary locking, change cache cleaning strategy

### LINKED ISSUE #4503: stat: packetcache-size reaches impossible values
`packetcache-size=18446744073709551613` always happens for me as soon as a second record of a zone (and even in different zone) is changed. I'm using version 4.0.1 on ArchLinux.

My steps were:

1.  start pdns server (master=yes)
2.  request ttl change of A record (via API using nsedit)
3.  request ttl change of another A or AAAA record the same ways as above
4.  immediatly after second record is changed, packetcache-size jumps through all roofs

watch via:
```watch -n 1 "pdns_control show \* | sed -e 's/,/\n/g' | grep packetcache-size"```

### PR DESCRIPTION
**Pull request by mind04**
fixes #4503 and make metronome updates realtime

### CODE REVIEW

File: pdns/packetcache.cc [hunk 1]
```diff
@@ -386,76 +394,79 @@ map<char,int> PacketCache::getCounts()
...
+
+  DLOG(L<<"Done with cache clean, cacheSize: "<<*d_statnumentries<<", totErased"<<totErased<<endl);
+}
+
+void PacketCache::cleanupIfNeeded()
+{
+  if (d_ops++ == d_nextclean) {
+    int timediff = max((int)(time(0) - d_lastclean), 1);
+
+    L<<"cleaninterval: "<<d_cleaninterval<<", timediff: "<<timediff<<endl;
```
ahupowerdns: no DLOG?

pdns/packetcache.cc
```diff
@@ -386,76 +394,79 @@ map<char,int> PacketCache::getCounts()
...
       if(toTrim && erased > toTrim / d_maps.size())
-	break;
-      
+        break;
+
       if(lookedAt > lookAt / d_maps.size())
-	break;
+        break;
+
+      i++;
```
ahupowerdns: This looks to be wrong - it refers to i after erasing it. This is why it used to be blah.erase(i++) which moves to the next iterator before erasing it. There is also a new syntax I think i=blah.erase(i) but haven't checked.

pdns/packetcache.cc
```diff
@@ -386,76 +394,79 @@ map<char,int> PacketCache::getCounts()
...
+}
+
+void PacketCache::cleanupIfNeeded()
+{
+  if (d_ops++ == d_nextclean) {
+    int timediff = max((int)(time(0) - d_lastclean), 1);
+
+    L<<"cleaninterval: "<<d_cleaninterval<<", timediff: "<<timediff<<endl;
+
+    if (d_cleaninterval == 300000 && timediff < 30) {
```
ahupowerdns: what does this do? what is the magic 300000?

pdns/packetcache.cc
```diff
@@ -386,76 +394,79 @@ map<char,int> PacketCache::getCounts()
...
+      d_cleanskipped = true;
+      d_nextclean += d_cleaninterval;
+
+      L<<"cleaning skipped, timediff: "<<timediff<<endl;
+
+      return;
+    }
+    if(d_cleanskipped) {
```
ahupowerdns: what is the logic behind "cleanskipped"?

pdns/packetcache.cc
mind04: 300000 was the old cache clean interval, and since this was a good default for almost everyone I reused this number. At this interval I just clean time based (without further adoption). This way I prevent the clean interval from raising sky high on a small traffic spike.

pdns/packetcache.cc
mind04: make sure we stay at "max cleaninterval" if the cleaning was skipped, one or more times at "max cleaninterval". As timediff make no sense after the first skip.

pdns/packetcache.cc
mind04: oops, debugging leftover

PR #4510, opened 2016-09-26
Rewrite examples from SQL inserts to pdnsutil commands

bergzand:
This PR rewrites some of the docs to use pdnsutil example commands instead of raw SQL inserts. The TSIG documentation is split into a "generate secret" section and an "enable secret" section.


mind04:
Please document both options. The raw sql inserts and the corresponding pdnsutil commands.

Habbie:
ping

Habbie:
Ping!

PR #4516, opened 2016-09-30
On RPZ customPolicy, follow the resulting CNAME

### LINKED ISSUE #4188: Recursor RPZ does not follow CNAMEs
Hi,

I have pdns recursor v4.0.0 installed and I'm testing the RPZ functionality.

 With both rpzFile() and rpzMaster() the zone file is loaded correctly and is working, however defined policy actions are not applying. E.g.

rpzFile("/etc/pdns-recursor/example.rpz", {defpol=Policy.Drop})

This returns the IP address from the example.rpz file, not drop the query. Likewise Policy.Custom (e.g. {defpol=Policy.Custom, defcontent="test.example.com"}) does not return the custom response, it returns the IP from the file (or RPZ zone from master with rpzMaster()).

I'm using Centos 7, tried with packages from the v4 repo and the master repo (although at this time it looks like they're the same packages as v4 was just released).

Testing with SpamHaus DBL as well as the example.rpz from the RPZ announcement post:

$TTL 2h;
$ORIGIN domain.example.com.
@          SOA powerdns.example.net. hostmaster.example.com ( 1 12h 15m 3w 2h)
           NS powerdns.example.net.
; begin RPZ RR definitions

example.net A 192.168.2.5
*.example.net A 192.168.2.5

Example response with Policy.Drop:

Non-authoritative answer:
Name:    example.net
Addresses:  192.168.2.5
          192.168.2.5

Thanks.

pieterlexis:
Closes #4188

### DIFF
--- a/pdns/pdns_recursor.cc
+++ b/pdns/pdns_recursor.cc
@@ -637,6 +637,32 @@ static void protobufLogResponse(const std::shared_ptr<RemoteLogger>& logger, con
 }
 #endif
 
+/**
+ * Chases the CNAME provided by the PolicyCustom RPZ policy.
+ *
+ * @param spoofed: The DNSRecord that was created by the policy, should already be added to ret
+ * @param qtype: The QType of the original query
+ * @param sr: A SyncRes
+ * @param res: An integer that will contain the RCODE of the lookup we do
+ * @param ret: A vector of DNSRecords where the result of the CNAME chase should be appended to
+ */
+void handleRPZCustom(const DNSRecord& spoofed, const QType& qtype, SyncRes& sr, int& res, vector<DNSRecord>& ret)
+{
+  if (spoofed.d_type == QType::CNAME) {
+    bool oldWantsRPZ = sr.d_wantsRPZ;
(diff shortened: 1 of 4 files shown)


pdns/pdns_recursor.cc
```diff
@@ -637,6 +637,20 @@ static void protobufLogResponse(const std::shared_ptr<RemoteLogger>& logger, con
 }
 #endif
 
+void handleRPZCustom(const DNSRecord& spoofed, const QType& qtype, SyncRes& sr, int& res, vector<DNSRecord>& ret)
+{
+  if (spoofed.d_type == QType::CNAME) {
+    sr.d_wantsRPZ = false;
```
rgacogne: I wonder if we should save the existing value of `d_wantsRPZ` to restore it afterwards? I doesn't seem to matter right now, but it might prevent surprises in the future.

pieterlexis: I considered this, but realized when we hit RPZ we always send out the answer (and never use the SyncRes again). But to prevent future surprises, this is a good idea.

PR #4526, opened 2016-10-03
API: url in zone info should be absolute

### LINKED ISSUE #4524: API zone info returns an url without a starting /
When querying the API for zones with: `curl -H 'X-API-Key: secret' http://127.0.0.1:8081/api/v1/servers/localhost/zones` the returned json contains an url to the api zone info. This url does not start with a `/` whereas the docs specify that it should. (Other url fields from the api do start with a `/`).
The 3.4.x api does not seem to have this issue

Powerdns version 4.0.1
Distribution: Gentoo

Reproduce:
1. enable api
2. curl the api along the lines of `curl -H 'X-API-Key: secret' http://127.0.0.1:8081/api/v1/servers/localhost/zones`

expected behaviour:
URL field starts with a `/`, for example: `"url": "/api/v1/servers/localhost/zones/test.example.org."`

Observed behavior
URL does not start with a slash, for example: "url": "api/v1/servers/localhost/zones/test.example.org."

Proposed fix:
Add a `/` at the relevant place in getZoneInfo in ws-auth.cc

zeha:
Fixes #4524.

### DIFF
--- a/pdns/ws-auth.cc
+++ b/pdns/ws-auth.cc
@@ -308,7 +308,7 @@ static Json::object getZoneInfo(const DomainInfo& di) {
   return Json::object {
     // id is the canonical lookup key, which doesn't actually match the name (in some cases)
     { "id", zoneId },
-    { "url", "api/v1/servers/localhost/zones/" + zoneId },
+    { "url", "/api/v1/servers/localhost/zones/" + zoneId },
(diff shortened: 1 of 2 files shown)


zeha:
Did not see that the issue came with a PR. This PR has a test.

PR #4536, opened 2016-10-05
dnsdist: Add an option to 'mute' UDP responses per bind

### LINKED ISSUE #4527: dnsdist should have 'mute' option to do everything, except respond
This should be set per listening address. The usecase is 1) benchmarking with "real" IP addresses 2) setups where dnsdist is receiving packets from a port mirror for analysis, but should not be answering.

For some failover cases, we might want to set this at runtime, but let's only type that in if it isn't too hard. It would bring showListeners() and getListener(0):mute() and getListener(0):unmute() tec.

The goal is to only stop sending the answer at the sendmsg() call and really do all the other processing correctly.

rgacogne:
Fixes #4527.


Habbie:
auth is unhappy on travis.. rebase?

PR #4539, opened 2016-10-05
Remove 'kind' param in DNSBackend::getDomainKeys

### LINKED ISSUE #4493: Unused parameter in DNSBackend::getDomainKeys()
Hi,

The "kind" parameter to DNSBackend::getDomainKeys() seems not used at all and is always set to 0 in the calls. Maybe it could be removed?

Cheers,
Grégory

pieterlexis:
It was unused anyway. Closes #4493

### DIFF
--- a/modules/luabackend/dnssec.cc
+++ b/modules/luabackend/dnssec.cc
@@ -363,23 +363,19 @@ bool LUABackend::addDomainKey(const DNSName& name, const KeyData& key, int64_t& 
     return ok >= 0;
 }
 
-bool LUABackend::getDomainKeys(const DNSName& name, unsigned int kind, std::vector<KeyData>& keys) {
-    //what is kind used for?
-
+bool LUABackend::getDomainKeys(const DNSName& name, std::vector<KeyData>& keys) {
     if(f_lua_getdomainkeys == 0) 
 	return false;
 
     if(logging)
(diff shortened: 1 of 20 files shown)


modules/luabackend/dnssec.cc
```diff
@@ -363,19 +363,16 @@ bool LUABackend::addDomainKey(const DNSName& name, const KeyData& key, int64_t&
...
     if(logging)
-	L << Logger::Info << backend_name << "(getDomainKeys) BEGIN name: '" << name << "' kind: '" << kind << endl;
+	L << Logger::Info << backend_name << "(getDomainKeys) BEGIN name: '" << name << endl;
 
     lua_rawgeti(lua, LUA_REGISTRYINDEX, f_lua_getdomainkeys);
 
     lua_pushstring(lua, name.toString().c_str());
-    lua_pushinteger(lua, kind);
 
     if(lua_pcall(lua, 2, 1, f_lua_exec_error) != 0) {
```
mind04: Also remove 1 argument here and update the function in https://github.com/PowerDNS/pdns/blob/master/modules/luabackend/test2/pdns-luabackend.lua

zeha: @pieterlexis lua_pcall takes the number of arguments, so 2->1 here

PR #4545, opened 2016-10-06
Clean up getAuth/getSOA intertwining

zeha:


Habbie:
@mind04 @pieterlexis please help me clarify this: is it correct that `getAuth` should no longer live in backends, as `getSOA` is now allowed to return a 'higher' SOA?

Habbie:
ok, question answered as 'yes' from comment added in 9105d7b3110b30184288162a01f6723b2c54addb

zeha:
Rebased to fix ueberbackend.hh conflict.

mind04:
too bad the comment in 9105d7b is wrong.

Should be something like this:
```
// A backend can respond to our authority request with the 'best' match it
// has. For example, when asked for a.b.c.example.com. it might respond with
// com. We then store that and keep querying the other backends in case one
// of them has a more specific zone but don't bother asking this specific
// backend again for b.c.example.com. or c.example.com.
// If a backend has no match it may respond with an empty qname.
```

GetAuth() has implemented a default fallback to getSOA() but this does not mean getSOA() is always the most optimal solution for all backends. 

Problems with shorter SOA results:
- direct SOA queries
- caching

zeha:
Hmm, OTOH we do not have any backends implementing getAuth in the tree...

pieterlexis:
> Problems with shorter SOA results:
>   * direct SOA queries

Are SOA queries special-cased to use getSOA()?

>    * caching

This is a point, adding new subdomains to other backends could introduce weird support questions. But I don't think this feature is used very often.

pieterlexis:
@zeha, I think this PR makes a lot of sense. But I think some words in the backend writer's guide (and maybe the pipe backend docs?) about higher SOA's and the getSOA fallback for getAuth default might be in order?

mind04:
This pull does not add the option for backends to return higher SOA's. It is only removing getAuth()

zeha:
Yeah right, not sure what the higher SOA thing is about?

@mind04 you have hinted that you might need getAuth() to be available to backends? Do you have a specific implementation? What exactly does that do / how does it work / what are the limitations?

Habbie:
I'd like to get this one resolved so we can ship 4.1.0 soon, what's the story?

aj-gh:
Parts of the story, at least the ones dealing with returning "higher" SOAs, might be related to the issues we had with returning IPv6 reverse DNS responses from a pipe backend; also related to #4020.
We have a pipe backend that does IPv6 reverse DNS and knows that it's authoritative for let's say 1.0.0.2.ip6.arpa - but it can't return that info until specifically asked for the SOA of "1.0.0.2.ip6.arpa" which causes lots of unnecessary pipe backend lookups until it finally reaches that level.

We ended up running the following patch, in pipebackend.cc.
Comment from back then was: "so this needs a patch in the pipe backend .. this could be something we might merge"

> bool PipeBackend::getSOA(const DNSName &domain, SOAData &sd, DNSPacket *p)
> {
>   if (DNSBackend::getSOA(domain, sd, p)) {
>     /* check if we secretly have a higher SOA and set that in the returned record*/
>     this->lookup(QType(QType::SOA),domain,p);
>     DNSResourceRecord rr;
>     while(this->get(rr)) {
>       if (rr.qtype != QType::SOA) throw PDNSException("Got non-SOA record when asking for SOA");
>       sd.qname = rr.qname;
>     }
>     return true;
>   }
>   return false;
> }

Probably that's not state of the art anymore but I hope this helps and we don't need to patch anything anymore in 4.1.

Habbie:
@zeha can you PR "Stop the illusion that packet-specific SOAs were acceptable" separately while we ponder the rest of this PR some more?

zeha:
Will do!

zeha:
I think the important bits are merged; if getAuth should go away from DNSBackend or not can be tracked in a new ticket if still needed.

PR #4558, opened 2016-10-11
Auth 4.0.x backports

mind04:


Habbie:
Great! Want to add the postgres leak plug as well?

mind04:
There is already an open pull request for the postgres fix #4459

PR #4559, opened 2016-10-11
Fix a few warnings on rawhide

mind04:


Habbie:
I see boost.m4 was upstreamed already, nice work!

PR #4564, opened 2016-10-13
rec: Pass a `DNSQuestion` object to Lua hooks

rgacogne:
The main motivation is to get rid of the huge number of parameters
passed to our Lua hooks, and to make it easy to access new values
from every hook. It also prevents copying the `DNSName` and
`ComboAddress` for every Lua hook.
I'm still wondering whether we actually need to make `dq` a shared
pointer, or if we couldn't just allocate it on the stack.

Note that the content of the `DNSQuestion::data` map is now persistent between hooks 
(for the same query).

### DIFF
--- a/pdns/lua-recursor4.cc
+++ b/pdns/lua-recursor4.cc
@@ -36,32 +36,32 @@ RecursorLua4::RecursorLua4(const std::string &fname)
   throw std::runtime_error("Attempt to load a Lua script in a PowerDNS binary without Lua support");
 }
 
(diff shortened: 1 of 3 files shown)


Habbie:
> Note that the content of the DNSQuestion::data map is now persistent between hooks
(for the same query).

I've been telling people it already was! 😆

Habbie:
`dq` could definitely go on the stack, it will stay in the same call tree on the mthread. Merging this in any case!

Issue #1872, opened 2014-11-11
Notifies not sent out on Thursday with INCEPTION-INCREMENT


sfrost:
When the serial changes on Thursday due to INCEPTION-INCREMENT, notifies to slaves should be sent out but currently aren't.  According to Peter on IRC, this is due to a change which was missed in commit 535f4e31bcbf2a342069de729005aefa99454085; specifically that `CommunicatorClass::notifyDomain` also needs updating to send out notifies when the serial changes.

PR #4565, opened 2016-10-13
Spread DNSSEC signature inception over the week

jimmybergman:
A client uses PowerDNS as a master for a set of external slaves at a third party anycast provider that gets the zones with AXFR.

The client has a large number of zones (hundreds of thousands).

The anycast provider has for a long time been annoyed that they have to refetch all zones close to Thursday 00:00 UTC when the signatures wraps and thus the SOA serial is incremented.

This pull requests makes the signature inception be spread over the entire week, hashed over the zone name, instead of all beeing at a Thursday 00:00 UTC. The inception is still always at most a week in the past, i.e. the main logic of signature validity is the same - the difference is that we make the actual time of week of the wrap different for different zones.

It is done by taking the first 7 bits of the hash of the zone name and using this to offset when the wrap happens by dividing the week into 127 chunks and picking one of them.

Since this also affects the signature caching that uses the "blunt by effective" method of clearing the cache on the first cache miss after the week wrapped, the commit changes this to have 128 separate caches (each containing a maximum of max-signature-cache-entries/128 records to not start using much more memory with identical config).

This has the extra benefit of not purging the signature cache for all zones at the same time, only 1/128 of the zones every 4725 seconds - which in turn should make the performance over the week more consistent.


Habbie:
Just reading the description - hashing the name is exactly how I had this in mind, it's a very sane approach!

pieterlexis:
@mind04 you're comments will be more than welcome!

jimmybergman:
@Habbie @pieterlexis any toughts for this or more specifically, anything missing for the gods of the open source to accept this sacrifice and reward it with a merge? :-)

Habbie:
@jimmybergman we're a bit swamped with other work. Expect attention in December!

Habbie:
Ticket #1872 is related. If we fix #1872, then we basically also need this. But as long as we don't have #1872, this 'spreading over the week' may confuse matters for users because they can no longer predict when they need to send a manual NOTIFY. So my thinking is that we should fix these things hand-in-hand.

jimmybergman:
@Habbie Sorry for beeing a bit slow with my response, other tasks got in the way, the last commit fixes ticket #1872, there is a description of how I did it in the commit message.

I would have preferred to not have to change the module interface, but since it was defined as "getUpdatedMasters" I figured that out of the three alternatives:
a) To try to keep the database serial updated whenever it wrap. This would be hard with the current design, since the EDIT-SOA was applied whenever a query happened after the rollover, and we want to send NOTIFY to slaves when the serial wraps regardless of if anyone has actually queried for the domain yet.

b) To replicate all the EDIT-SOA logic in the backend modules. This would either mean lots of code duplication or depending on dnsseckeeper in the modules which would both be very ugly.

c) To change the interface to return all master zones and instead have the EDIT-SOA logic centrally when comparing which of them are updated. Since all but one of the backends anyway fetched all rows from the data store I figured that this was the cleanest way.

Howver, I have not had a chance to test any modules other than the gmysql one.

jimmybergman:
@Habbie Any comments? :-)

Habbie:
Can you rebase so we can test & review? Apologies for the delay.

jimmybergman:
@Habbie Sure, give me a few days, I'll ping you once rebased and let you know when you can test+review.

omoerbeek:
I would like to spend some time on this to make sure this can be merged,

A rebase would be very welcome, I hope you are still interested. In the mean time, some general comments:

* The code uses integer constants and constant expressions in a few places.  I prefer a single set of (commented) const definitions with clear names.
* I see some time_t -> uint32_t conversions that look suspicious, or at least need a proper check. It would be nice if all new code is 64-bit time_t safe. And additional thing is that time_t is a signed type. So a thorough check of all these (implicit) conversions must be done.
* I'm a bit worried by he backend interaction changes. Getting all domains and filtering on the C++ side might be expensive for a large number of zones. We have to study the performance impact and maybe come up with an alternative solution (that would probably require db schema changes, so that is also not very attractive either).

jimmybergman:
@omoerbeek 

I did start with a rebase, but didn't get it to working state and currently don't have the free time to drive this forward (also I don't have the pleasure to run or host PowerDNS anymore due to job changes).

The backend interaction changes are only needed for the additional changes related to ticket #1872, the original intention of the PR was simpler.

I wasn't super happy about having to change the module interface (also the cause of the messy conflicts since the PR got too large to lay around this long) but at the time I couldn't figure out any better way, see the comment at https://github.com/PowerDNS/pdns/pull/4565#issuecomment-280588605

At least at the time of the change all of the common backends already fetched all zones when figuring out which were updated, i.e. https://github.com/PowerDNS/pdns/pull/4565/files#diff-f7fa79052caf3807dc966882ddbed30a (the new code just keeps them for a bit longer).

omoerbeek:
OK, understood. Currently thinking (and talking to others) to see if it's worth the time and trouble.

pieterlexis:
Closing, needs a lot of work in the NOTIFY department

PR #4566, opened 2016-10-13
DNS Header and EDNS Flag access for Lua

pieterlexis:
This PR depends on #4564!

This PR adds a `dq.dh` member that has all the same functions as the `dh` in the ipfilter function. It also adds two functions to the `dq` object to get the EDNS Flags (only DO is supported at the moment)

### DIFF
--- a/pdns/lua-recursor4.cc
+++ b/pdns/lua-recursor4.cc
@@ -166,6 +166,25 @@ boost::optional<dnsheader> RecursorLua4::DNSQuestion::getDH() const
   return boost::optional<dnsheader>();
 }
 
+vector<string> RecursorLua4::DNSQuestion::getEDNSFlags() const
+{
+  vector<string> ret;
+  if (ednsFlags) {
+    if (*ednsFlags & EDNSOpts::DNSSECOK)
+      ret.push_back("DO");
+  }
+  return ret;
+}
+
+bool RecursorLua4::DNSQuestion::getEDNSFlag(string flag) const
+{
+  if (ednsFlags) {
+    if (flag == "DO" && (*ednsFlags & EDNSOpts::DNSSECOK))
+      return true;
+  }
+  return false;
+}
+
 vector<pair<uint16_t, string> > RecursorLua4::DNSQuestion::getEDNSOptions() const
 {
   if(ednsOptions)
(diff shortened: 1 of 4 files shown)


pdns/lua-recursor4.hh
```diff
@@ -57,6 +57,7 @@ public:
     const struct dnsheader* dh{nullptr};
     const bool isTcp;
     const std::vector<pair<uint16_t, string>>* ednsOptions{nullptr};
+    const uint16_t* ednsFlags;
```
rgacogne: Could you initialize that to `nullptr`?

pdns/lua-recursor4.cc
```diff
@@ -159,6 +159,21 @@ static int getFakePTRRecords(const DNSName& qname, const std::string& prefix, ve
 
 }
 
+vector<string> RecursorLua4::DNSQuestion::getEDNSFlags() const
+{
+  vector<string> ret;
+  if (*ednsFlags & EDNSOpts::DNSSECOK)
```
rgacogne: It would be safer to check whether `ednsFlags` is set before de-referencing it. It's not set in the `preoutquery` hook for example.

pdns/lua-recursor4.cc
```diff
@@ -159,6 +159,21 @@ static int getFakePTRRecords(const DNSName& qname, const std::string& prefix, ve
...
+    ret.push_back("DO");
+  return ret;
+}
+
+bool RecursorLua4::DNSQuestion::getEDNSFlag(string flag) const
+{
+  if (flag == "DO" && (*ednsFlags & EDNSOpts::DNSSECOK))
```
rgacogne: Same here, `ednsFlags` might be `NULL`.

pdns/lua-recursor4.cc
```diff
@@ -365,6 +365,12 @@ RecursorLua4::RecursorLua4(const std::string& fname)
   d_lw->registerMember<bool (DNSQuestion::*)>("variable", [](const DNSQuestion& dq) -> bool { return dq.variable; }, [](DNSQuestion& dq, bool newVariable) { dq.variable = newVariable; });
   d_lw->registerMember<bool (DNSQuestion::*)>("wantsRPZ", [](const DNSQuestion& dq) -> bool { return dq.wantsRPZ; }, [](DNSQuestion& dq, bool newWantsRPZ) { dq.wantsRPZ = newWantsRPZ; });
 
+  d_lw->registerMember<const dnsheader (DNSQuestion::*)>("dh",
+    [](const DNSQuestion& dq) -> const dnsheader {
+      return *dq.dh;
```
rgacogne: I know I wrote that code, but we should probably check that `dq.dh` is not `NULL` before de-referencing it. Sorry!

Habbie:
Needs rebase

Issue #4475, opened 2016-09-15
Negative notified_serial crashes pdns_server and pdnsutil check-all-zones


opeter:
Somehow a negative notified_serial sneaked into our domains table.  It causes pdns_server to crash/loop, should be easy to reproduce and easy to fix:

In our case:

```
mysql> select notified_serial from domains where notified_serial is not null order by notified_serial asc limit 10;  
+-----------------+
| notified_serial |
+-----------------+
|       -93865895 |
|               4 |
|      2006013100 |
|      2006050100 |
|      2006061900 |
|      2006061900 |
|      2006090500 |
|      2006090500 |
|      2006090500 |
|      2006091800 |
+-----------------+
10 rows in set (1.04 sec)

mysql> select \* from domains where notified_serial = -93865895;
+--------+------------+---------------+------------+--------+-----------------+---------+
| id     | name       | master        | last_check | type   | notified_serial | account |
+--------+------------+---------------+------------+--------+-----------------+---------+
| 845349 | xxxxxxx.de | xxx.xxx.xxx.x |       NULL | MASTER |       -93865895 | NULL    |
+--------+------------+---------------+------------+--------+-----------------+---------+
1 row in set (1.71 sec)

mysql> update domains set notified_serial = NULL where id = 845349 limit 1;  
Query OK, 1 row affected (0.00 sec)
Rows matched: 1  Changed: 1  Warnings: 0
```

The Error messages are:

```
    Sep 15 06:45:07 a pdns[14120]: PowerDNS Authoritative Server 4.0.1 (C) 2001-2016 PowerDNS.COM BV                                                                                                                              
    Sep 15 06:45:07 a pdns[14120]: Using 64-bits mode. Built using clang 3.4.1 (tags/RELEASE_34/dot1-final 208032) on Sep 15 2016 08:35:02 by root@localhost                                                                      
    Sep 15 06:45:07 a pdns[14120]: PowerDNS comes with ABSOLUTELY NO WARRANTY. This is free software, and you are welcome to redistribute it according to the terms of the GPL version 2.                                         
    Sep 15 06:45:07 a pdns[14120]: Creating backend connection for TCP                                                                                                                                                            
    Sep 15 06:45:07 a pdns[14120]: Master/slave communicator launching                                                                                                                                                            
    Sep 15 06:45:07 a pdns[14120]: About to create 8 backend threads for UDP                                                                                                                                                      
[...]

PR #4576, opened 2016-10-17
Issue #4475: Return absolute values from MySQL to work around serial overflow

opeter:
What about using the database (in my case MySQL) to do the sanatizing by an abs() envelope?

Any non-positive number or string will be returned as 0 and overwritten with the current timestamp by the daemon.

### DIFF
--- a/modules/gmysqlbackend/gmysqlbackend.cc
+++ b/modules/gmysqlbackend/gmysqlbackend.cc
@@ -92,9 +92,9 @@ public:
 
     declare(suffix,"master-zone-query","Data", "select master from domains where name=? and type='SLAVE'");
 
-    declare(suffix,"info-zone-query","","select id,name,master,last_check,notified_serial,type,account from domains where name=?");
+    declare(suffix,"info-zone-query","","select id,name,master,abs(last_check) as last_check,abs(notified_serial) as notified_serial,type,account from domains where name=?");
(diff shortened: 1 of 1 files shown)


rgacogne:
I'm not sure we want to silently ignore invalid serial values, especially at the database level, since it would prevent any chance of reporting that something is wrong.

opeter:
It would be ignored, yes, but since the absolute value of an invalid/negative/whatever entry would be 0, it will be overwritten during the next run with a fine value (certainly current unix timestamp) and everything will be alright - no crash, no additional code needed.

zeha:
Can we please not do this? Is there an issue that explains the actual problem so another fix can be found?

opeter:
@zeha Somehow we ended up with negative values for last_check and notified_serial in our database.  I do not have an explanation for this but the server process ended up with a crash / serial overflow.
To be honest, the fix of the overflow within the code definitely exceeds my knowledge of C++ so gave this PR a try.
The content in our database is quiet old (started ~2005) and it is not guaranteed that something besides PowerDNS inserted bad values into the tables, so this might not be an issue either.

We were able to fix the problem itself by updating negative values to 0.

The issue along with he PR can both be omitted.

Habbie:
This just hides problems. Not doing this.

PR #4580, opened 2016-10-17
[Documentation] Update some auth remove backend examples to add trailing '.'s

clokep:
This updates some of the examples in the remote backend documentation to include the trailing dot that PowerDNS 4 sends, e.g. the paragraph here:

> There is a breaking change on v4.0 and later. Before version 4.0, the DNS names passed in queries were without trailing dot, after version 4.0 the DNS names are sent with trailing dot. F.ex. example.org is now sent as example.org.

I've only fixed instances that I was actually able to reproduce, not ones that I haven't seen.


clokep:
Thanks! 👍

PR #4587, opened 2016-10-19
auth: Implement `Bind2Backend::getSOA()`, use the `zoneId` in `lookup()`

rgacogne:


rgacogne:
Closing this while we decide what to do with #4545. If we keep `DNSBackend::getAuth()` and `Bind2Backend::getAuth() should be implemented instead of `Bind2Backend::getSOA()`, if we don't the signature of `DNSBackend::getSOA()` will probably change anyway. I'm opening a separate PR for the `zoneId` fix.

PR #4596, opened 2016-10-20
dnsdist: Change the default max number of queued TCP conns to 1000

rgacogne:
With the existing default value of 0, we could consume an unlimited
number of file descriptors if the queued connections kept piling up.

### DIFF
--- a/pdns/README-dnsdist.md
+++ b/pdns/README-dnsdist.md
@@ -915,9 +915,9 @@ The maximum number of threads in the TCP pool is controlled by the
 increased to handle a large number of simultaneous TCP connections.
 If all the TCP threads are busy, new TCP connections are queued while
 they wait to be picked up. The maximum number of queued connections
-can be configured with `setMaxTCPQueuedConnections()`, and any value other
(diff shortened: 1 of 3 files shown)


paddg:
If I set `setMaxTCPQueuedConnections(1000)`, which is the default after this change, dnsdist complains at startup:
`Warning, this configuration can use more than 43768 file descriptors, web server and console connections not included, and the current limit is 4096.`

I have to set `ulimit -n`   at least to `58358`, to get rid of this message, which is strange because 58358 is much more than 43768. However,  `LimitNOFILE=` in the systemd unit file should also adjusted I think.

rgacogne:
I will check more carefully but it looks like there is a mistake in the way the maximum number of file descriptors is computed:
`requiredFDsCount += (tcpBindsCount * g_maxTCPQueuedConnections);`
should probably be:
`requiredFDsCount += g_maxTCPQueuedConnections;`

And dnsdist warns when the current FD limit is less than 4/3 of the maximum FD usage, since some FD can't be accurately predicted, like web and console connections. That formula doesn't work too well for large FD counts, I'm afraid.

PR #4547, opened 2016-10-06
Remove serializeSOAData, refactor calculate/edit/increaseSOA

zeha:
This also fixes #4515 and #2377.

PR #4598, opened 2016-10-20
auth: Add support for "NONE" SOA-EDIT kind

zilopbg:
Setting the "SOA-EDIT" value for a zone to "NONE" causes an error:
'SOA-EDIT type 'NONE' for zone ZONENAME is unknown.'

The calculateEditSOA function could return the SOA as-is in such cases.

### DIFF
--- a/pdns/serialtweaker.cc
+++ b/pdns/serialtweaker.cc
@@ -72,6 +72,9 @@ uint32_t calculateEditSOA(uint32_t old_serial, const string& kind, const DNSName
   else if(!kind.empty()) {
     g_log<<Logger::Warning<<"SOA-EDIT type '"<<kind<<"' for zone "<<zonename<<" is unknown."<<endl;
   }
+  // Seen strictly, this is a broken config: we can only come here if
+  // both SOA-EDIT and default-soa-edit are set to "", but the latter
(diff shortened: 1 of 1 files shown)


Habbie:
@zeha?

zeha:
Please merge this after #4547. Code itself LGTM

nmamn:
Hi, any update regarding this ?

thanks :)

zeha:
Not sure this is still needed. If some docs refer to "NONE", they should probably refer to "" instead...

nmamn:
unless I'm mistaken, it's still specified in the official documentation.

https://doc.powerdns.com/md/authoritative/dnssec/#soa-edit-ensure-signature-freshness-on-slaves

> (...)
> NONE
>
> Ignore default-soa-edit and/or default-soa-edit-signed settings.
>(...)

Habbie:
> Not sure this is still needed. If some docs refer to "NONE", they should probably refer to "" instead...

@zeha can you help us decide what to do with this PR?

zeha:
Okay, indeed we do need support for NONE (otherwise you can't override `default-soa-edit`). I've rebased this and added a comment, also killed a doubly checked case.

Issue #4601, opened 2016-10-21
new g.root-servers.net. AAAA 2001:500:12::d0d


Habbie:

Habbie:
Closed in #4603

PR #4603, opened 2016-10-21
g.root-servers.net added IPv6

nivex:
http://www.internic.net/domain/db.cache
last update:    October 20, 2016


rgacogne:
Closes #4601.

PR #4606, opened 2016-10-24
Use dnsname for nsec generation

mind04:
and a massive cleanup in internal before and aftername handling. Before and afternames are now relative DNSNames and the backends are responsible for converting these values to the right format if needed (labelreverse, replace dots by spaces and things like this) 

PS. replacing dots by spaces in dns name string is a very bad idea for a backend as it will break NSEC sorting rules for characters < 32

Closes #4345 (included in this pull)

### DIFF
--- a/modules/bindbackend/bindbackend2.cc
+++ b/modules/bindbackend/bindbackend2.cc
@@ -924,7 +924,7 @@ void Bind2Backend::queueReloadAndStore(unsigned int id)
   }
 }
 
-bool Bind2Backend::findBeforeAndAfterUnhashed(BB2DomainInfo& bbd, const DNSName& qname, DNSName& unhashed, string& before, string& after)
+bool Bind2Backend::findBeforeAndAfterUnhashed(BB2DomainInfo& bbd, const DNSName& qname, DNSName& unhashed, DNSName& before, DNSName& after)
 {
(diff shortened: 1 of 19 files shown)


mind04:
I prefer defect over enhancement. 
Btw. the fix for the canonical ordering also affect dnssec in the recursor and should be backported to recursor stable.

pieterlexis:
@rgacogne could you review at your leisure?

modules/bindbackend/bindbackend2.cc
```diff
@@ -924,97 +924,76 @@ void Bind2Backend::queueReloadAndStore(unsigned int id)
...
+  //   cerr<<record.qname<<"\t"<<makeHexDump(record.qname.toDNSString())<<endl;
 
-    while(iter == records->end() || (qname.canonCompare(iter->qname)) || (!(iter->auth) && (!(iter->qtype == QType::NS))) || (!(iter->qtype)))
-      iter--;
+  recordstorage_t::const_iterator iterBefore, iterAfter;
 
-    if(iter->qname.empty())
-      before.clear();
-    else { 
+  iterBefore = iterAfter = records->upper_bound(qname.makeLowerCase());
```
rgacogne: I don't think we need to lowercase here because it looks like the caller always does that, but that's a nit.

PR #4625, opened 2016-10-25
auth: Unify usage of randomness source by using `dns_random()`

rgacogne:
### Short description

`Utility::random()` is not impossible to predict, so even if we are not using it for anything sensitive it's better to just use `dns_random()` instead.
Reported by mongo (thanks!).
### Checklist

I have:
- [ ] documented the code
- [ ] added regression tests
- [ ] added unit tests

### DIFF
--- a/pdns/common_startup.cc
+++ b/pdns/common_startup.cc
@@ -475,7 +475,7 @@ static void triggerLoadOfLibraries()
 
 void mainthread()
 {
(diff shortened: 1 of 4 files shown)


zeha:
Maybe the remaining users of random() should also go away, so random() itself can go away? (dnsdist, geoipbackend, randombackend from a quick grep)

PR #4628, opened 2016-10-25
API: do not return dnssec info in domain list

zeha:
### Short description

Stop backend creation in a tight loop when listing all domains over the API. Should improve performance considerably. Also notes in the docs that "list" is incomplete.
### Checklist

I have:
- [ ] documented the code
- [ ] added regression tests
- [ ] added unit tests

### DIFF
--- a/regression-tests.api/test_Zones.py
+++ b/regression-tests.api/test_Zones.py
@@ -49,8 +49,11 @@ def eq_zone_rrsets(rrsets, expected):
 
 class Zones(ApiTestCase):
 
-    def test_list_zones(self):
-        r = self.session.get(self.url("/api/v1/servers/localhost/zones"))
+    def _test_list_zones(self, dnssec=True):
+        path = "/api/v1/servers/localhost/zones"
+        if not dnssec:
+            path = path + "?dnssec=false"
+        r = self.session.get(self.url(path))
         self.assert_success_json(r)
         domains = r.json()
         example_com = [domain for domain in domains if domain['name'] in ('example.com', 'example.com.')]
@@ -59,13 +62,22 @@ class Zones(ApiTestCase):
         print(example_com)
         required_fields = ['id', 'url', 'name', 'kind']
         if is_auth():
-            required_fields = required_fields + ['masters', 'last_check', 'notified_serial', 'edited_serial', 'serial', 'account']
+            required_fields = required_fields + ['masters', 'last_check', 'notified_serial', 'serial', 'account']
+            if dnssec:
+                required_fields = required_fields = ['dnssec', 'edited_serial']
             self.assertNotEquals(example_com['serial'], 0)
         elif is_recursor():
             required_fields = required_fields + ['recursion_desired', 'servers']
         for field in required_fields:
             self.assertIn(field, example_com)
+        if not dnssec:
+            self.assertNotIn('dnssec', example_com)
 
+    def test_list_zones_with_dnssec(self):
+        self._test_list_zones(True)
+
+    def test_list_zones_without_dnssec(self):
+        self._test_list_zones(False)
 
 class AuthZonesHelperMixin(object):
     def create_zone(self, name=None, **kwargs):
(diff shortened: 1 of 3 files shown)


Habbie:
How about moving the backend creation outside of the loop? Would at least allow db connection reuse.

zeha:
Yes, but we'll still send 2+ SQL queries in a loop then...

Habbie:
True. Maybe Wiebel can measure for us on his real load?

. o O ( ideally we would do all dnssec status in a single query )

vandalon:
Just did. The old pdns (version 4.0.1+225) takes about 8.5 seconds. the same version with this patch less then 0.3 seconds.

zeha:
Jeeves reports that nsedit uses this.

tuxis-ie:
Perhaps it is a good idea to change https://github.com/PowerDNS/pdns/blob/master/pdns/dbdnsseckeeper.cc#L59 to:

```
  keyset_t keys = getKeys(zone, true); // does the cache
```

and actually use the mentioned cache?

zeha:
> and actually use the mentioned cache?

this is actually the default with no param supplied.

zengxibai:
We have a few thousands zones (in the table ‘domains') hosted in PowerDNS database (MySQL). it looks very slow (10-12 minutes) to get a list of all domains with API call below:

curl -H 'X-API-Key: mykey' 'http://127.0.0.1:8080/api/v1/servers/localhost/zones/'

Is this a noticed performance problem? Any insight where the slowness comes from? 

I do see the DNSSEC info in the returned answer. below is one of the returned zones info. e.g. 

  {
    "account": "",
    "dnssec": false,
    "id": "188.28.10.in-addr.arpa.",
    "kind": "Native",
    "last_check": 0,
    "masters": [],
    "name": "188.28.10.in-addr.arpa.",
    "notified_serial": 0,
    "serial": 2017081704,
    "url": "api/v1/servers/localhost/zones/188.28.10.in-addr.arpa."
  },

zeha:
Rebased for 4.1.

Habbie:
Can we make this a choice by the API client? `&dnssec=true` or the like?

pieterlexis:
> We could, but if we flip the default to false, it's still a breaking change.

Please make the default `true` so we can merge this :)

zeha:
Well, this got a bit harder because since 1258fec `edited_serial` is also returned in the zone list, which needs a DNSSECKeeper and does at least two metadata lookups and calls isSecuredZone. 

So either dnssec=false also turns off returning `edited_serial` or ... I dunno.

wojas:
Turning off `edited_serial` is not ideal, but I would still be for merging this as is and documenting the behavior. The default stays the same, so this does not affect current users, only those that opt-in to the more efficient call.

Habbie:
Agreed - let's add some docs and merge this.

Habbie:
Testing and documenting this now.

Habbie:
Documentation pushed, renders like this:

image

Habbie:
> looks good, could use a test

Oh yes, I somehow imagined it had one already. Will do.

Habbie:
Test pushed.

regression-tests.api/test_Zones.py
```diff
@@ -59,13 +62,22 @@ def test_list_zones(self):
...
-            required_fields = required_fields + ['masters', 'last_check', 'notified_serial', 'edited_serial', 'serial', 'account']
+            required_fields = required_fields + ['masters', 'last_check', 'notified_serial', 'serial', 'account']
+            if dnssec:
+                required_fields = required_fields = ['dnssec', 'edited_serial']
             self.assertNotEquals(example_com['serial'], 0)
         elif is_recursor():
             required_fields = required_fields + ['recursion_desired', 'servers']
         for field in required_fields:
             self.assertIn(field, example_com)
+        if not dnssec:
```
pieterlexis: I think this should be inside the `if is_auth()` block

Habbie: Fixed, and now we also don't pointlessly test `dnssec=false` on the Recursor.

Habbie:
Rebased.

Issue #4312, opened 2016-08-12
dnsdist can be crashed by feeding bad maintenance function in on a console connection


phonedph1:
As discussed on IRC, feeding in a maintenance function with bad code inside does not get validated until it's actually run on the daemon process which can then cause the daemon to crash. There is no indication of this on the console connection until you try the next operation.

It seems like we should do better than allowing it to crash - either by denying adding anything that can't be fully validated or clearing the maintenance function on errors.

```
Read configuration from '/etc/dnsdist/dnsdist.conf'

Connecting to 0.0.0.0:5199

> function maintenance() addDynBlocks(exceedNXDOMAINS(1, 10), "Exceeding NXDOMAINs", 60) end



> 

> showServers()

Fatal error: failed in writen2: Broken pipe





> Got control connection from 127.0.0.1:52991

terminate called after throwing an instance of 'LuaContext::ExecutionErrorException'

  what():  [string "function maintenance() addDynBlocks(exceedN..."]:1: attempt to call global 'exceedNXDOMAINS' (a nil value)

Aborted

```

phonedph1:
```
13:04 < ahu> ok - well, it might be worth a ticket but I can't imagine how we'd improve this specific usecase
13:04 < ahu> I hadn't even realized you could do this
13:04 < sarnold> :)
13:04 < ph1> seems like something people would be doing to add blocks, no?
13:05 < ph1> perhaps I'm just doing something stupid.
13:05 < ahu> yes, but typing in a whole maintenance function on the console isn't what I imagined people would be doing
13:06 < ahu> an issue here is that the console does not get that error message
13:07 < ahu> since it arrives asynchronously
13:08 < ph1> Less concerned about an error on the console vs actually breaking the daemon process by feeding it unvalidated code.
13:08 < ahu> well, I could make it impossible to do this
13:08 < ahu> which is one way to fix it
13:08 < ahu> since we can't validate it usefully
13:09 < ahu> we could stop ourselves from crashing on an error in maintenance() and log it
13:09 < ahu> but that might still make people unhappy
13:09 < ahu> since it fails silently (from the console perspective)
13:10 -!- stirni [~Adium@2001:620:0:82::104] has joined #powerdns
13:10 < ahu> ph1: don't get me wrong, I'd love to be helpful
13:10 < ph1> FWIW even with something incorrect in the config on startup it starts running and then appears to break at the first maintenance call
13:11 < ahu> yes
[...]

PR #4640, opened 2016-10-26
Don't exit dnsdist on an exception in maintenance

pieterlexis:
### Short description

dnsdist would exit if an exception was thrown in the maintenance function.

This change logs every minute if the maintenance function throws an exception.
### Checklist

I have:

### DIFF
--- a/pdns/dnsdist.cc
+++ b/pdns/dnsdist.cc
@@ -1247,15 +1247,27 @@ void* maintThread()
 {
   int interval = 1;
   size_t counter = 0;
+  int32_t secondsToWaitLog = 0;
 
   for(;;) {
     sleep(interval);
 
     {
       std::lock_guard<std::mutex> lock(g_luamutex);
-      auto f =g_lua.readVariable<boost::optional<std::function<void()> > >("maintenance");
-      if(f)
-        (*f)();
+      auto f = g_lua.readVariable<boost::optional<std::function<void()> > >("maintenance");
+      if(f) {
+        try {
+          (*f)();
+          secondsToWaitLog = 0;
+        }
+        catch(std::exception &e) {
+          if (secondsToWaitLog <= 0) {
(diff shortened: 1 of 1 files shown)


phonedph1:
Nice! Fixes the biggest complaint in #4312

pdns/dnsdist.cc
```diff
@@ -1247,15 +1247,27 @@ void* maintThread()
...
       std::lock_guard<std::mutex> lock(g_luamutex);
       auto f =g_lua.readVariable<boost::optional<std::function<void()> > >("maintenance");
-      if(f)
-        (*f)();
+      if(f) {
+        try {
+          (*f)();
+          secondsToWaitLog = 0;
+        }
+        catch(LuaContext::ExecutionErrorException &e) {
```
rgacogne: Perhaps we could catch `std::runtime_error`, or even `std::exception` instead of `LuaContext::ExecutionErrorException`? Because otherwise we don't catch `LuaContext::SyntaxErrorException` nor `LuaContext::WrongTypeException` for example, although I'm not sure they can be raised in that context.

pdns/dnsdist.cc
```diff
@@ -1247,15 +1247,27 @@ void* maintThread()
...
+          if (secondsToWaitLog == 0) {
+            infolog("Error during execution of maintenance function: %s", e.what());
+            secondsToWaitLog = 61;
+          }
+          --secondsToWaitLog;
```
pieterlexis: we might amend the assignment of `secondsToWaitLog` to that configuration setting. as the decrement then still happens every second?

rgacogne: If `interval` is not guaranteed to be equal to 1 anymore, the decrement won't happen every second except if we alter the code to do so. This would probably be needed to keep the cache cleaning delay consistent anyhow.

PR #4648, opened 2016-10-28
Document dq.data (from @Habbie)

elad:
### Short description
Document `dq.data` - a table that is persistent across the lifetime of the `dq` object and can be used to store custom context.

### Checklist
I have:
- [ ] documented the code
- [ ] added regression tests
- [ ] added unit tests
- [ ] checked that this code was merged to master

### DIFF
--- a/docs/markdown/recursor/scripting.md
+++ b/docs/markdown/recursor/scripting.md
@@ -99,7 +99,7 @@ The DNSQuestion object contains at least the following fields:
     * policyCustom: The CNAME content for the `pdns.policyactions.Custom` response, a string
     * policyTTL: The TTL in seconds for the `pdns.policyactions.Custom` response
 * wantsRPZ - A boolean that indicates the use of the Policy Engine, can be set to `false` in `preresolve` to disable RPZ for this query
-* data - a table that is persistent throughout the lifetime of the `dq` object and can be used to store custom context.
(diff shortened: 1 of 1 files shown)


docs/markdown/recursor/scripting.md
```diff
@@ -99,6 +99,7 @@ The DNSQuestion object contains at least the following fields:
     * policyCustom: The CNAME content for the `pdns.policyactions.Custom` response, a string
     * policyTTL: The TTL in seconds for the `pdns.policyactions.Custom` response
 * wantsRPZ - A boolean that indicates the use of the Policy Engine, can be set to `false` in `preresolve` to disable RPZ for this query
+* data - a table that is persistent throughout the lifetime of the `dq` object and can be used to store custom context.
```
Habbie: custom context or custom content?

elad: I'm using to calling it "context" but I guess it may conflict with the "Lua context" - so I'll change to "content" (or maybe just "custom data"?)

Habbie: Yes, I understand how the data, in your context (haha), is context. I think 'data' is fine. Maybe also note that it is strictly string/string, you can't put tables inside this table like you can with normal Lua tables.

PR #4660, opened 2016-11-03
dnsdist: Add `statNodeRespRingSince(seconds, visitor)`

rgacogne:
### Short description
While `statNodeRespRing()` applies the `visitor` function to every entry in the response rings, `statNodeRespRingSince()` only applies it to entries added in the last `seconds` seconds.

I'd rather add an optional parameter to `statNodeRespRing()`, but I have been unable to successfully call a Lua function from `statNodeRespRing()` if the callback function is not the last parameter.

### Checklist

I have:
- [x] added regression tests


rgacogne:
Closing this since #4775 should allow adding an optional parameter to `statNodeRespRing()` instead.

PR #4672, opened 2016-11-10
On (re-)priming, fetch the root NS records

pieterlexis:
### Short description
Before, when repriming the root because it was evicted or expired from the cache, we did not go out to the internet to immediately update the list we inserted into the cache from the built-in hints or hintfile. While not in issue in the DNS world, in the DNSSEC world this means that we suddenly stop serving RRSIGs for `.|NS` at that point, confusing downstream validators.

This patch also ensures that we don't add the hint data as authoritative to the cache *and* adds DNSSEC validation for the results of this query if the settings support it.

Found by @giganteous 

### Checklist

I have:
- [x] documented the code

### DIFF
--- a/pdns/pdns_recursor.cc
+++ b/pdns/pdns_recursor.cc
@@ -1950,35 +1950,9 @@ static void houseKeeping(void *)
     }
 
     if(now.tv_sec - last_rootupdate > 7200) {
-      SyncRes sr(now);
-      sr.setDoEDNS0(true);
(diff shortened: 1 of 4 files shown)


bootc:
After being pointed at this from IRC, I've applied it to 4.0.3 (with a couple of tweaks). So far so good: BIND is no longer complaining at me all the time that `validating ./NS: got insecure response; parent indicates it should be secure`.

PR #4675, opened 2016-11-11
Backport #4672: On (re-)priming, fetch the root NS records

pieterlexis:
### Short description
see #4672 

### Checklist

I have:


bootc:
The patch in this PR matches the version I independently backported to 4.0.3 as I mentioned in #4672, so LGTM in case that helps.

PR #4692, opened 2016-11-16
SSql: Use unique_ptr for statements

cmouse:
### Short description
Convert gSQLBackend and BindBackend to use unique_ptr for storing SSqlStatement pointers. This will prevent any accidentical copying or misuse of said pointers.

### Checklist
I have:
- [ ] documented the code
- [ ] added regression tests
- [ ] added unit tests


nlyan:
Why not make the prepare() function return a unique_ptr by value?

cmouse:
I'll have to see what it causes.

cmouse:
Seems to be doable, fixed.

cmouse:
@mind04 has tested goracle that it compiles.

nlyan:
Use of std::move() in setupStatements() is redundant, since the return value from prepare() is already an rvalue and will be moved-from by default

cmouse:
fixed @nlyan's nit.

nlyan:
LGTM

Issue #2955, opened 2015-11-29
Recursor immediately crashes on OS X


zeha:
OS X 10.11.1
clang: Apple LLVM version 7.0.0 (clang-700.1.76)
boost: stable 1.59.0 (bottled), HEAD
git head 1ed6e213dc86533a15935c9ad0888f9a9c515017

```
ldb -- ../pdns/pdns_recursor --daemon=no --socket-dir=. --config-dir=. --local-port=5555 
(lldb) target create "../pdns/pdns_recursor"
Current executable set to '../pdns/pdns_recursor' (x86_64).
(lldb) settings set -- target.run-args  "--daemon=no" "--socket-dir=." "--config-dir=." "--local-port=5555"
(lldb) run
Process 60644 launched: '../pdns/pdns_recursor' (x86_64)
Nov 29 02:18:22 PowerDNS Recursor 0.0.apidnsnamedot.g43111af.dirty (C) 2001-2015 PowerDNS.COM BV
Nov 29 02:18:22 Using 64-bits mode. Built using clang 7.0.0 (clang-700.1.76) on Nov 28 2015 17:27:47 by ch@sx.
Nov 29 02:18:22 PowerDNS comes with ABSOLUTELY NO WARRANTY. This is free software, and you are welcome to redistribute it according to the terms of the GPL version 2.
Nov 29 02:18:22 Reading random entropy from '/dev/urandom'
Nov 29 02:18:22 Only allowing queries from: 127.0.0.0/8, 10.0.0.0/8, 100.64.0.0/10, 169.254.0.0/16, 192.168.0.0/16, 172.16.0.0/12, ::1/128, fc00::/7, fe80::/10
Nov 29 02:18:22 Will not send queries to: 127.0.0.0/8, 10.0.0.0/8, 100.64.0.0/10, 169.254.0.0/16, 192.168.0.0/16, 172.16.0.0/12, ::1/128, fc00::/7, fe80::/10, 0.0.0.0/8, 192.0.0.0/24, 192.0.2.0/24, 198.51.100.0/24, 203.0.113.0/24, 240.0.0.0/4, ::/96, ::ffff:0:0/96, 100::/64, 2001:db8::/32, 0.0.0.0, ::
Nov 29 02:18:22 PowerDNS Recursor itself will distribute queries over threads
Nov 29 02:18:22 NOT using IPv6 for outgoing queries - set 'query-local-address6=::' to enable
[...]

PR #4710, opened 2016-11-25
Fix crash on FreeBSD (and possibly other OS's)

RvdE:
### Short description

Fixes a crash on FreeBSD when built against boost 1.62

```
Assertion failed: (this != &x), function operator=, file /usr/local/include/boost/container/string.hpp, line 830.
```

This causes a SIGABRT (signal 6)

### Checklist

I have:
- [ ] documented the code
- [ ] added regression tests
- [ ] added unit tests

### DIFF
--- a/pdns/dnsname.hh
+++ b/pdns/dnsname.hh
@@ -30,7 +30,7 @@
 #include <boost/version.hpp>
 
 // it crashes on OSX and doesn't compile on OpenBSD
-#if BOOST_VERSION >= 105300 && ! defined( __APPLE__ ) && ! defined(__OpenBSD__)
+#if BOOST_VERSION >= 105300 && ! defined(_LIBCPP_VERSION) && ! defined(__OpenBSD__)
 #include <boost/container/string.hpp>
 #endif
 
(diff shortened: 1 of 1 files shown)


Habbie:
We need to understand why `_LIBCPP_VERSION` is the right define here, before we can merge.

Habbie:
Re 2: Indeed, OpenBSD still needs to be specified explicitly.

Habbie:
ping?

RvdE:
_LIBCPP_VERSION is set on FreeBSD in two files:
c++/v1/__config:#define _LIBCPP_VERSION 3800
c++/v1/tr1/__config:#define _LIBCPP_VERSION 3800

FreeBSD 11.0-RELEASE (which I'm running) comes with CLang v3.8.0 in base, so that's probably where it's coming from. On MacOSX I'm assuming replacing __APPLE_ with _LIBCPP_VERSION also has the desired effect since that also comes with CLang in base, but I can't say if it'll effect any other OS's.

Maybe somebody else has a more conclusive answer.

RvdE:
Just tried Recursor 4.0.6 (without this patch) against Boost 1.64, but the problem still exists.

zeha:
Ref #2955

Habbie:
As far we can see, this is no longer needed.

RvdE:
I've justed tested the recursor 4.1.3 port on FreeBSD 11.2-BETA3 without the patch and it crashes just like it did before.

Habbie:
I'd reopen the PR but the source repo is gone

PR #4711, opened 2016-11-26
Enable setting custom pgsql connection parameters

### LINKED ISSUE #2138: Allow overriding gpgsql connect string
Right now we support a subset of valid postgres connect string keys, notably excluding anything related to SSL/TLS. We should allow the user to specify the whole string.

The only hard part in this is hiding the password from logging (which is easy right now because we know where it is).

thusoy:
### Short description
Enable setting custom pgsql connection parameters, like TLS parameters.

Should this be added to documentation or examples anywhere? To validate TLS connections against a local trust root you can set `gpgsql-extra-connection-args=sslmode=verify-ca sslrootcert=<path-to-certs>`.

### Checklist

I have:
- [ ] documented the code
- [ ] added regression tests
- [ ] added unit tests

Closes #2138.

### DIFF
--- a/docs/markdown/authoritative/backend-generic-postgresql.md
+++ b/docs/markdown/authoritative/backend-generic-postgresql.md
@@ -48,6 +48,11 @@ The password to for [`gpgsql-user`](#gpgsql-user).
 ## `gpgsql-dnssec`
 Enable DNSSEC processing for this backend. Default=no.
 
+## `gpsql-extra-connection-parameters`
+Extra connection parameters to forward to postgres. If you want to pin a specific certificate for
+the connection you should set this to `sslmode=verify-full sslrootcert=<path-to-CA-cert>`. Accepted
+parameters are documented [in the PostgreSQL documentation](https://www.postgresql.org/docs/current/static/libpq-connect.html#LIBPQ-PARAMKEYWORDS).
+
 # Default schema
 ```
 !!include=../modules/gpgsqlbackend/schema.pgsql.sql
--- a/modules/gpgsqlbackend/gpgsqlbackend.cc
+++ b/modules/gpgsqlbackend/gpgsqlbackend.cc
@@ -44,7 +44,8 @@ gPgSQLBackend::gPgSQLBackend(const string &mode, const string &suffix)  : GSQLBa
         	  getArg("host"),
         	  getArg("port"),
         	  getArg("user"),
-        	  getArg("password")));
+        	  getArg("password"),
+        	  getArg("extra-connection-parameters")));
   }
 
   catch(SSqlException &e) {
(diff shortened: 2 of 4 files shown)


thusoy:
I tested that this actually stops a MitM by using this script: https://raw.githubusercontent.com/thusoy/postgres-mitm/master/postgres_mitm.py. As long as postgres is set to accept connections over tls (`hostssl` option in `pg_hha.conf` and `ssl on;` in `postgresql.conf`, run this script with `python postgres_mitm.py -p 5433 127.0.0.1` and set pdns to connect to localhost on 5433, you'll see that the mitm can intercept the authentication. If this patch is applied and `gpgsql-extra-connection-args=sslmode=verify-ca sslrootcert=<path-to-certs>` is set, the connection will be refused and the mitm won't learn the credentials.

thusoy:
@rgacogne Anything more I can do to get this merged? Should I add documentation somewhere?

zeha:
Please add docs, indeed. Apart from that, I think the pdns team is just somewhat busy right now.

rgacogne:
@thusoy Thank you for this PR! Adding a few words in `docs/markdown/authoritative/backend-generic-postgresql.md` would be great, otherwise it looks fine to me.

thusoy:
@rgacogne Cool, wrote some words and included an example to show how it can be used.. I also renamed the variable to `_parameters` instead of `_args` since the postgres docs are pretty consistent about calling it parameters.

docs/markdown/authoritative/backend-generic-postgresql.md
```diff
@@ -48,6 +48,11 @@ The password to for [`gpgsql-user`](#gpgsql-user).
 ## `gpgsql-dnssec`
 Enable DNSSEC processing for this backend. Default=no.
 
+## `gpsql-extra-connection-parameters`
+Extra connection parameters to forward to postgres. If you want to pin a specific certificate for
+the connection you should set this to `sslmode=verify-full sslrootcert=<path-to-CA-cert>`. Accepted
+parameters are documented [here.](https://www.postgresql.org/docs/current/static/libpq-connect.html#LIBPQ-PARAMKEYWORDS).
```
Habbie: please remove spurious period from `[here.]` i.e. replace it with `[here]`. Good to merge after that.

zeha: I'd suggest rewording this to `[in the PostgreSQL documentation]`... ('here' sounds somewhat meaningless in a web document)

thusoy:
@Habbie @zeha Good catch, both of you. Fixed.

Habbie:
Ready for merge when Travis passes.

thusoy:
@Habbie Just a heads up, Travis is happy!

Habbie:
@thusoy thanks for the ping, never hurts!

porbas:
It was merged after latest (auth-4.0.3) release. When should I expect next release including this PR?

pieterlexis:
@porbas this PR is not backported to 4.0.x, it exists only in master. If you need this in 4.0.x, can you open an issue about this?

PR #4726, opened 2016-12-02
dnsdist: Add an option to return ServFail when no server is available

rgacogne:
### Short description

By default dnsdist drops the query if there is no server available to forward it to. This PR adds the `setServFailWhenNoServer(bool)` directive to reply with a ServFail instead.

### Checklist

I have:
- [x] documented the code
- [x] added regression tests

### DIFF
--- a/pdns/README-dnsdist.md
+++ b/pdns/README-dnsdist.md
@@ -1385,6 +1385,7 @@ instantiate a server with additional parameters
     * `setServerPolicyLua(name, function)`: set server selection policy to one named 'name' and provided by 'function'
     * `showServerPolicy()`: show name of currently operational server selection policy
     * `newServerPolicy(name, function)`: create a policy object from a Lua function
+    * `setServFailWhenNoServer(bool)`: if set, return a ServFail when no servers are available, instead of the default behaviour of dropping the query
  * Available policies:
     * `firstAvailable`: Pick first server that has not exceeded its QPS limit, ordered by the server 'order' parameter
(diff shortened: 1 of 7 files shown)


pdns/README-dnsdist.md
```diff
@@ -1385,6 +1385,7 @@ instantiate a server with additional parameters
     * `setServerPolicyLua(name, function)`: set server selection policy to one named 'name' and provided by 'function'
     * `showServerPolicy()`: show name of currently operational server selection policy
     * `newServerPolicy(name, function)`: create a policy object from a Lua function
+    * `setServFailWhenNoServer(bool)`: if set, returns a ServFail when no server are available instead of dropping the query (the default)
```
Habbie: 'if set, return a ServFail when no servers are available, instead of the default behaviour of dropping the query'

pdns/dnsdist-console.cc
```diff
@@ -339,6 +339,7 @@ const std::vector<ConsoleKeyword> g_consoleKeywords{
   { "setRules", true, "list of rules", "replace the current rules with the supplied list of pairs of DNS Rules and DNS Actions (see `newRuleAction()`)" },
   { "setServerPolicy", true, "policy", "set server selection policy to that policy" },
   { "setServerPolicyLua", true, "name, function", "set server selection policy to one named 'name' and provided by 'function'" },
+  { "setServFailWhenNoServer", true, "bool", "if set, returns a ServFail when no server are available instead of dropping the query (the default)" },
```
Habbie: and update same text here. We still need to figure something out against this duplication :)

PR #4775, opened 2016-12-16
LuaWrapper: Use the correct index when storing a function

rgacogne:
### Short description
The LuaWrapper used to assume that the function was at the top of the stack, making it effectively impossible to have a callback function parameter anywhere else than as the last parameter.

### Checklist

I have:

### DIFF
--- a/ext/luawrapper/include/LuaContext.hpp
+++ b/ext/luawrapper/include/LuaContext.hpp
@@ -1642,10 +1642,10 @@ private:
     // structure that will ensure that a certain is stored somewhere in the registry
     struct ValueInRegistry {
         // this constructor will clone and hold the value at the top of the stack in the registry
-        ValueInRegistry(lua_State* lua) : lua{lua}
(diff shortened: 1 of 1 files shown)


ext/luawrapper/include/LuaContext.hpp
```diff
@@ -1642,10 +1642,10 @@ class LuaContext {
     // structure that will ensure that a certain is stored somewhere in the registry
     struct ValueInRegistry {
         // this constructor will clone and hold the value at the top of the stack in the registry
```
Habbie: This comment is now wrong?

ext/luawrapper/include/LuaContext.hpp
```diff
@@ -1642,10 +1642,10 @@ class LuaContext {
     // structure that will ensure that a certain is stored somewhere in the registry
```
Habbie: This comment already was wrong ;)

PR #4779, opened 2016-12-16
dnsdist: Save history to home-dir

### LINKED ISSUE #4562: .dnsdist_history deposited in cwd
Hi there. 

Just to mention that a .dnsdist_history file is created in the cwd when using the -c (client) connection. Doesn't really matter where you are. Seems to be a default. 

```
cd /var/tmp # or anywhere
ls -a | grep .dnsdist_history
echo "dumpStats()" | /usr/bin/dnsdist -c
ls -a | grep .dnsdist_history
.dnsdist_history
```

Likely(?) around this point needs to be modified:
https://github.com/PowerDNS/pdns/blob/master/pdns/dnsdist-console.cc#L88

(dnsdist 1.0.0)
dnsdist-1.0.0-1.el7.x86_64
CentOS Linux release 7.2.1511 (Core)

Keep up the good work 👍

pieterlexis:
### Short description
Only use CWD as a last resort. This prevents cluttering directories with `.dnsdist_history` files

Closes #4562

### Checklist

I have:
- [x] documented the code
- [ ] added regression tests
- [ ] added unit tests

### DIFF
--- a/pdns/dnsdist-console.cc
+++ b/pdns/dnsdist-console.cc
@@ -21,6 +21,7 @@
  */
 #include "dnsdist.hh"
 #include "sodcrypto.hh"
+#include "pwd.h"
 
 #if defined (__OpenBSD__)
 #include <readline/readline.h>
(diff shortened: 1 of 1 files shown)


pdns/dnsdist-console.cc
```diff
@@ -45,6 +46,25 @@ void feedConfigDelta(const std::string& line)
...
 }
 
+string historyFile(const bool &ignoreHOME = false)
+{
+  string ret;
+
+  struct passwd pwd;
+  struct passwd *result;
+  char buf[16384];
+  getpwuid_r(getuid(), &pwd, buf, sizeof(buf), &result);
```
rgacogne: I'm wondering if we should use `geteuid()` instead, in the unlikely event `dnsdist` is started with `-u` while in foreground mode.

pdns/dnsdist-console.cc
```diff
@@ -45,6 +46,25 @@ void feedConfigDelta(const std::string& line)
...
+
+  const char *homedir = getenv("HOME");
+  ret = "."; // CWD if nothing works..
+  if (result)
+    ret = string(pwd.pw_dir);
+  if (homedir && !ignoreHOME) // $HOME overrides what the OS tells us
+    ret = string(homedir);
+  ret.append("/.dnsdist_history");
```
rgacogne: How about not setting `ret` to `"."` beforehand and testing here if it's empty? If it is (somehow `$HOME` or `pwd.pw_dir` were empty or nothing worked), we could then fallback to `"."` before appending `"/.dnsdist_history"`. This would prevent trying to write in `"/.dnsdist_history"`.

PR #4780, opened 2016-12-17
dnsdist: Add an optional `seconds` parameter to `statNodeRespRing()`

rgacogne:
### Short description
By default `statNodeRespRing()` applies the visitor function to every entry in the responses ring. When passed a non-zero `seconds` parameter, it will only apply it to entries added in the last `seconds` seconds.
Based on top of #4775, replaces #4660.

### Checklist

I have:
- [x] added regression tests


pieterlexis:
I think commit 4ec1e17 was already merged in #4775?

rgacogne:
Indeed, rebased!

PR #4793, opened 2016-12-19
Don't call `hostname -f` on openbsd

### LINKED ISSUE #2579: `hostname -f` errors on OpenBSD [was: Illumos & Solaris systems]
At the very least the recursor build system uses hostname to bake user@host type strings into config files and binaries.  Understandably, it wants the FQDN.  With GNU hostname, specifying `-f` or `--fqdn` does that.  Unfortunately, the hostname command on Illumos-derived systems doesn't understand any options:

```
SYNOPSIS
       /usr/bin/hostname [name-of-host]
```

If run as a privileged user, any argument will be used as the new hostname.  So, `hostname -f` will set the hostname to `-f`.  This is definitely not what the user wanted.

Since Illumos started off as a fork of OpenSolaris, the same warning applies to all versions of OpenSolaris as well as Solaris 10 and older.  I have no idea if `hostname` changed in Solaris 11 as Solaris 11 and Illumos diverged quite a bit.

pieterlexis:
### Short description
OpenBSD would shout at you when `-f` is passed to `hostname`
Closes #2579

### Checklist

I have:
- [ ] documented the code
- [ ] added regression tests
- [ ] added unit tests

### DIFF
--- a/m4/pdns_enable_reproducible.m4
+++ b/m4/pdns_enable_reproducible.m4
@@ -15,7 +15,7 @@ AC_DEFUN([PDNS_ENABLE_REPRODUCIBLE], [
     build_user=$(id -u -n)
 
     case "$host_os" in
-    solaris2.1* | SunOS)
+    solaris2.1* | SunOS | openbsd*)
       build_host_host=$(hostname)
       build_host_domain=$(domainname)
       build_host="$build_host_host.$build_host_domain"


Habbie:
OpenBSD 6.0: PR removes the error message and yields
```
$ grep -i hostname config.log
hostname = postel.my.domain
```

:+1:

PR #4798, opened 2016-12-20
rec: put 15 seconds lower bound on max-cache-ttl

### LINKED ISSUE #4634: PowerDNS Recursor 4.0.3-1pdns.xenial always returns SERVFAIL when setting --max-cache-ttl=0
Setting `--max-cache-ttl=0` makes `pdns-recursor=4.0.3-1pdns.xenial` answer `SERVFAIL` to all kind of queries sent to the recursor. If the recursor is unable to deal with this option set to `0`, we should make it fail when trying to start instead of just silently replying with an error to queries.

pieterlexis:
### Short description
Having no cache makes no sense and anything below a few seconds will
lead to SERVFAILs on recursors that have a low query load.

Closes #4634

### Checklist

I have:
- [ ] documented the code
- [ ] added regression tests
- [ ] added unit tests

### DIFF
--- a/docs/markdown/recursor/settings.md
+++ b/docs/markdown/recursor/settings.md
@@ -584,6 +584,8 @@ for most installations.
 
 Maximum number of seconds to cache an item in the DNS cache, no matter what the
 original TTL specified.
+Since PowerDNS Recursor 4.1.0, the minimum value of this setting is 15.
+i.e. setting this to lower than 15 will make this value 15.
 
 ## `max-mthreads`
 * Integer
--- a/pdns/pdns_recursor.cc
+++ b/pdns/pdns_recursor.cc
@@ -2750,7 +2750,7 @@ int serviceMain(int argc, char*argv[])
   SyncRes::s_nopacketcache = ::arg().mustDo("disable-packetcache");
 
   SyncRes::s_maxnegttl=::arg().asNum("max-negative-ttl");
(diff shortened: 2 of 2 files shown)


pdns/pdns_recursor.cc
```diff
@@ -2750,7 +2750,7 @@ int serviceMain(int argc, char*argv[])
   SyncRes::s_nopacketcache = ::arg().mustDo("disable-packetcache");
 
   SyncRes::s_maxnegttl=::arg().asNum("max-negative-ttl");
-  SyncRes::s_maxcachettl=::arg().asNum("max-cache-ttl");
+  SyncRes::s_maxcachettl=max(::arg().asNum("max-cache-ttl"), 15);
```
rgacogne: `max()`?

pieterlexis: the title is weird, we cap the minimum for this setting to 15 :)

docs/markdown/recursor/settings.md
```diff
@@ -583,7 +583,7 @@ for most installations.
 * Available since: 3.2
 
 Maximum number of seconds to cache an item in the DNS cache, no matter what the
-original TTL specified.
+original TTL specified. The minimal value is capped to 15 since PowerDNS Recursor 4.1.0.
```
Habbie: suggest rewording this, perhaps like I reworded this PR title, to avoid confusing use of min vs. max.

PR #4809, opened 2016-12-22
Propose to remove redundant indexes from gmysqlbackend default schema

aj-gh:
### Short description
The gmysqlbackend default schema seems to contain redundant indexes which I'm proposing to remove unless they provide any benefit.

### Slighty longer description
As MySQL will happily use the left-most prefix of a multiple-column index "domain_id" on "records" is covered by "recordorder". Same applies to "comments_domain_id_idx" on "comments" (by "comments_order_idx")
It could be that these indices are left-overs or they might provide some benefit in certain cases, for example increased performance since they are smaller than their multi-column friends. But - is that really the case or does the additional overhead in maintaining both in fact degrade performance?
During my tests the MySQL query optimizer preferred the multi-column index also in cases where a lookup in the single-column one would have been sufficient. Based on that I think we could get rid of those indexes.

In any case this PR could be used to review these indices, and remove them if they are not needed (and otherwise document why they exist).

### Checklist

I have:
- [ ] documented the code
- [ ] added regression tests
- [ ] added unit tests

### DIFF
--- a/modules/gmysqlbackend/schema.mysql.sql
+++ b/modules/gmysqlbackend/schema.mysql.sql
@@ -28,7 +28,6 @@ CREATE TABLE records (
 ) Engine=InnoDB;
 
 CREATE INDEX nametype_index ON records(name,type);
-CREATE INDEX domain_id ON records(domain_id);
 CREATE INDEX recordorder ON records (domain_id, ordername);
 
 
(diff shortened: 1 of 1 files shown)


Habbie:
@mind04 has suggested that instead of removing the `domain_id` index, we should replace the `domain_id,ordername` index with an index on just `ordername`. Intuitively this seems wrong to me, but he says he saw serious performance improvements.

For the benefit of older MySQL optimisers that may not be as smart, I'm thinking:
1. `domain_id, ordername`
2. `ordername`

(plus keep `name,type`)

PR #4810, opened 2016-12-22
gmysqlbackend: Define IDs as unsigned in default schema

aj-gh:
### Short description
The gmysqlbackend default schema uses signed integers for both domain and record ids. Since these are auto_increment values and used as uint32_t in the code I suppose it doesn't make much sense to have them defined as signed integers, except for artificially reducing the maximum amount of possible records.
(And yes, in some broken setups admins were already hit by that)
If applied this should be mentioned in changelogs to allow for proper migration.

### Checklist

I have:
- [ ] documented the code
- [ ] added regression tests
- [ ] added unit tests

### DIFF
--- a/modules/gmysqlbackend/schema.mysql.sql
+++ b/modules/gmysqlbackend/schema.mysql.sql
@@ -1,5 +1,5 @@
 CREATE TABLE domains (
-  id                    INT AUTO_INCREMENT,
+  id                    INT UNSIGNED AUTO_INCREMENT,
   name                  VARCHAR(255) NOT NULL,
   master                VARCHAR(128) DEFAULT NULL,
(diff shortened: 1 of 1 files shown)


zeha:
Note that domain ids > UINT_MAX/2 are unsupported because some code treats the domain_id as a signed int32_t.

aj-gh:
Ok, I could remove domain_id. Any issues with the others being unsigned?

zeha:
`records.id` has been made into a `BIGINT` in another PR. Do you care about the other ids? If not, I'd suggest closing this PR.

aj-gh:
Well, I don't really get why we need signed (big-)integers instead of unsigned ones, but okay, there should be enough headroom now for "most" setups ;)
Closing.

PR #4815, opened 2016-12-23
dnsdist: Merge the client and server nonces to prevent replay attacks

rgacogne:
### Short description

Instead of using the local nonce to send messages (and so the remote one for received ones), split and merge the local and remote nonces to create two new nonces, one for client to server and one for server
to client.
This makes us more resistant against reply attacks, but note that it does break compatibility with previous versions.

### Checklist

I have:

### DIFF
--- a/pdns/dnsdist-console.cc
+++ b/pdns/dnsdist-console.cc
@@ -56,14 +56,16 @@ void doClient(ComboAddress server, const std::string& command)
   }
   SConnect(fd, server);
   setTCPNoDelay(fd);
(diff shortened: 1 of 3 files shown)


pieterlexis:
this is milestone'd for dnsdist 1.1. Due to the incompatible change, I suggest we delay and merge this first thing for dnsdist 1.2.

rgacogne:
> Shouldn't merge be an empty implementation here like init() and increment()?

It absolutely should, nice catch!

PR #4817, opened 2016-12-23
dnsdist: Add `setTCPUseSinglePipe()` to use a single TCP waiting queue

rgacogne:
### Short description
By default, every TCP worker thread has its own queue, and incoming TCP connections are dispatched to TCP workers on a round-robin basis.
This might cause issues if some connections are taking a very long time, since incoming ones will be waiting until the TCP worker they have been assigned to has finished handling its current query, while other TCP workers might be available.
This experimental `setTCPUseSinglePipe(true)` directive can be used so that all the incoming TCP connections are put into a single queue and handled by thefirst TCP worker available.

### Checklist

I have:
- [x] documented the code

### DIFF
--- a/pdns/dnsdist.hh
+++ b/pdns/dnsdist.hh
@@ -387,9 +387,9 @@ class TCPClientCollection {
   bool d_useSinglePipe;
 public:
 
-  TCPClientCollection(size_t maxThreads, bool useSinglePipe=false): d_maxthreads(maxThreads), d_useSinglePipe(useSinglePipe)
+  TCPClientCollection(size_t maxThreads, bool useSinglePipe=false): d_maxthreads(maxThreads), d_singlePipe{-1,-1}, d_useSinglePipe(useSinglePipe)
+
   {
-    d_maxthreads = maxThreads;
     d_tcpclientthreads.reserve(maxThreads);
 
     if (d_useSinglePipe) {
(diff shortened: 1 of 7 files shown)


pdns/dnsdist.hh
```diff
@@ -368,12 +368,30 @@ class TCPClientCollection {
...
   uint64_t d_maxthreads{0};
   std::mutex d_mutex;
+  int d_singlePipe[2];
+  bool d_useSinglePipe;
 public:
 
-  TCPClientCollection(size_t maxThreads)
+  TCPClientCollection(size_t maxThreads, bool useSinglePipe=false): d_maxthreads(maxThreads), d_useSinglePipe(useSinglePipe)
   {
     d_maxthreads = maxThreads;
```
RobinGeuze: Bit pointless to leave this here with the constructor initialisation list in place.

pdns/dnsdist.hh
```diff
@@ -368,12 +368,30 @@ class TCPClientCollection {
...
+      }
+      if (!setNonBlocking(d_singlePipe[1])) {
+        int err = errno;
+        close(d_singlePipe[0]);
+        close(d_singlePipe[1]);
+        throw std::runtime_error("Error setting the TCP single communication pipe non-blocking: " + string(strerror(err)));
+      }
+    }
+    else {
+      d_singlePipe[0] = -1;
```
RobinGeuze: Might be more consistent to preinitialize d_singlePipe to -1 for both, in that case you can do away with this else.

PR #4839, opened 2017-01-02
nsec ent-asterisk test is no longer bogus with unbound 1.6.0

mind04:
### Checklist

I have:


Habbie:
I can confirm that unbound 1.6.0 has fixed this bug. Maybe we can note that in the test description to avoid confusion for users?

PR #4854, opened 2017-01-04
In edns-subnet-processing use the RealRemote IP for AXFR ACLs

opeter:
### Short description
The real source IP gets lost if PowerDNS is behind dnsdist.
This patch fixes the problem when edns-subnet-processing is active.

### Checklist

I have:
- [x] documented the code
- [ ] added regression tests
- [ ] added unit tests

### DIFF
--- a/pdns/tcpreceiver.cc
+++ b/pdns/tcpreceiver.cc
@@ -564,6 +564,12 @@ bool TCPNameserver::canDoAXFR(shared_ptr<DNSPacket> q)
           // cerr<<"hit!"<<endl;
           return true;
         }
+        // EDNS Client Subnet matching
+        if(nm.match( q->getRealRemote().toStringNoMask() ) )
+        {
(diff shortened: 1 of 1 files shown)


rgacogne:
Hi, thank you for this pull request!
I'm a bit concerned by the fact that it blindly trusts the ECS information it received, regardless of the source. It means that anybody could send an `AXFR` request with an authorized IP in ECS to bypass the AXFR ACLs. We solved this for the `NOTIFY` case by only using the ECS information sent from whitelisted clients, based on the `trusted-notification-proxy` directive, as can be seen in https://github.com/PowerDNS/pdns/blob/master/pdns/packethandler.cc#L763
Perhaps we could add a similar directive to be able to whitelist the sources from which we trust the ECS information?

opeter:
In our specific case there the requests will be sent by only one IP anyway, in other words the IP we would put into such a trusted-proxy variable would be the same (from dnsdist). This is why I created the patch in the first place, the PowerDNS AuthServer will only see queries from the dnsdist's source IP but also needs to be able to see the IP from the original request to validate the ACLs.

Habbie:
I've unticked the "checked that this code was merged to master" box for you.

Habbie:
Furthermore, @rgacogne's concern is valid, we can't merge this as is.

PR #4856, opened 2017-01-05
dnsdist: Clarify docs: `maintenance()`, cache memory usage and `exceed*`

rgacogne:
### Short description
* Update the website for the 1.1.0 release
* Document how `maintenance()` is called
* Document the packet cache memory usage
* Document the return value of `exceed*` functions

### Checklist

I have:
- [ ] documented the code
- [ ] added regression tests
- [ ] added unit tests

### DIFF
--- a/pdns/README-dnsdist.md
+++ b/pdns/README-dnsdist.md
@@ -762,6 +762,8 @@ Dynamic Rule Generation
 -----------------------
 To set dynamic rules, based on recent traffic, define a function called `maintenance()` in Lua. It will
 get called every second, and from this function you can set rules to block traffic based on statistics.
+More exactly, the thread handling the `maintenance()` function will sleep for one second between each
+invocation, so if the function takes several seconds to complete it will not be invoked exactly every second.
 
 As an example:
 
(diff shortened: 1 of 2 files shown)


pdns/dnsdistdist/website/markdown/index.md
```diff
@@ -17,9 +17,9 @@ Finally, the DNS-OARC presentation is [here](https://indico.dns-oarc.net/event/2
 ([VIDEO!](https://www.youtube.com/watch?feature=player_embedded&v=PX3YYmBER7E#t=6403)).
 
 # Getting dnsdist
-Grab the master, 1.0 or pre-release packages from [our
-repositories](https://repo.powerdns.com).  The announcement of 1.0.0 is
-[here](https://blog.powerdns.com/2016/04/21/dnsdist-1-0-0-released/).
+Grab the master, 1.1, 1.0 or pre-release packages from [our
```
zeha: Maybe this should be "Grab the 1.1 release or development packages from..." instead?

pdns/README-dnsdist.md
```diff
@@ -857,6 +862,12 @@ The second one is the maximum lifetime of an entry in the cache, the third one i
 the minimum TTL an entry should have to be considered for insertion in the cache,
 the fourth one is the TTL used for a Server Failure or a Refused response. The last
 one is the TTL that will be used when a stale cache entry is returned.
+For performance reasons the cache will pre-allocate buckets based on the maximum number
+of entries, so be careful to set the first parameter to a reasonable value. Something
```
zeha: an exact number for "reasonable value" as a starting point might be nice...

rgacogne: Sure, but since it depends on the amount of memory you are willing to allocate to the cache, I'm not sure what to indicate here?

pieterlexis: Maybe a `X MB per Y buckets`?

zeha: or `on modern hardware with 8GB of RAM set this to NNNN as a start`...

rgacogne: How about adding:
> Assuming an average response size of 512 bytes, a cache size of 10000000 entries on a 64-bit host with 8GB of dedicated RAM would be a safe choice.

zeha: 👍

Issue #3097, opened 2015-12-23
validns and jdnssec-* do not support ECDSA


Habbie:
So we have temporarily stopped using them. When they have been fixed, revert d44560a7baeaa8c4d2f24abf70cfc9296b6453c1 (in part or in whole)

* [x] jdnssec
* [ ] validns

Habbie:
PR for validns ECDSA support: https://github.com/tobez/validns/pull/53

Habbie:
validns PR merged just now!

PR #4864, opened 2017-01-09
reinstate jdnssec usage now that it supports ECDSA

Habbie:
When we switched to ECDSA, we had to drop some of the validators. With jdnssec 0.13 supporting ECDSA, we are reinstating its usage.


pieterlexis:
Fixes a part of #3097

PR #4865, opened 2017-01-09
servfail-ttl: reduce from 60 to 5, and cap at 60

cnst:
According to http://tools.ietf.org/html/rfc2308#section-7.1,
server failure MUST NOT be cached for longer than 5 minutes.

In reality, caching server failure for an extended period of time
is be detrimental for client experience, where a short period of
downtime would then be amplified to a prolonged DNS outage.

This has been pointed out many years ago by DJB:

http://cr.yp.to/djbdns/third-party.html

>  The second tactic is to claim that widespread DNS clients will do something Particularly Evil when they are unable to reach all DNS servers. The problem with this argument is that the claim is false. Any such client is clearly buggy, and will be unable to survive in the marketplace: consider what happens if the *client's* routers briefly go down, or if the client's network is temporarily flooded.

Let's try not to do anything *Particularly Evil*. :-)

Additionally, when BIND first introduced this feature, it was set at 10 with a cap of 300, but was subsequently found to be detrimental, supposedly turned off for a release or two, and then came back with a default of 1 second (in place of the original 10), and a cap of 30 (in place of the original 300).

* https://source.isc.org/cgi-bin/gitweb.cgi?p=bind9.git;a=commit;h=a8783019814daa36dd57afe3f527462822834c3b

* https://kb.isc.org/article/AA-01178/

* https://source.isc.org/cgi-bin/gitweb.cgi?p=bind9.git;a=commit;h=90174e64f49bb7cba6a83fb665ebcb597aad7b57

Personally, I think any double-digit number should definitely not be the default.  On the other hand, `1s` is so small as to make it difficult to ascertain whether or not any cache is active in the first place.  I think `5s` may be the sweet spot of caching a failure, yet only enough for a couple of refresh clicks per user max (or maybe we should even go to `2s` or even `1s`, because software is meant for production, not for testing whether or not the caching works).

As for the cap, I think it's likewise important to ensure that admins cannot shoot their own users into the foot; I think 60s would be the sweet spot between pragmatism (cap is 30s in BIND) and flexibility (after all, 60s was the default in a few prior versions of PowerDNS).  I don't think it would be reasonable for anyone to cache failures for over a minute.

### DIFF
--- a/docs/markdown/recursor/settings.md
+++ b/docs/markdown/recursor/settings.md
@@ -668,13 +668,14 @@ the original TTL specified.
 
 ## `packetcache-servfail-ttl`
 * Integer
-* Default: 60
+* Default: 5 (since 4.1.0), 60 (before 4.1.0)
 * Available since: 3.2
 
 Maximum number of seconds to cache a 'server failure' answer in the packet cache.
 From 4.0.0 onward, this settings maximum is capped to [`packetcache-ttl`](#packetcache-ttl).
-i.e. setting `packetcache-ttl=15` and keeping `packetcache-servfail-ttl` at the
-default will lower `packetcache-servfail-ttl` to `15`.
+i.e. setting `packetcache-ttl=0` and keeping `packetcache-servfail-ttl` at the
+default will lower `packetcache-servfail-ttl` to `0`.
+From 4.1.0 onward, this setting is additionally capped at 60.
 
 ## `pdns-distributes-queries`
 * Boolean
--- a/pdns/pdns_recursor.cc
+++ b/pdns/pdns_recursor.cc
@@ -2754,6 +2754,8 @@ int serviceMain(int argc, char*argv[])
   SyncRes::s_packetcachettl=::arg().asNum("packetcache-ttl");
   // Cap the packetcache-servfail-ttl to the packetcache-ttl
   uint32_t packetCacheServFailTTL = ::arg().asNum("packetcache-servfail-ttl");
+  if (packetCacheServFailTTL > 60)
+    packetCacheServFailTTL = 60;
   SyncRes::s_packetcacheservfailttl=(packetCacheServFailTTL > SyncRes::s_packetcachettl) ? SyncRes::s_packetcachettl : packetCacheServFailTTL;
   SyncRes::s_serverdownmaxfails=::arg().asNum("server-down-max-fails");
   SyncRes::s_serverdownthrottletime=::arg().asNum("server-down-throttle-time");
@@ -3083,7 +3085,7 @@ int main(int argc, char **argv)
     ::arg().set("max-cache-ttl", "maximum number of seconds to keep a cached entry in memory")="86400";
     ::arg().set("packetcache-ttl", "maximum number of seconds to keep a cached entry in packetcache")="3600";
     ::arg().set("max-packetcache-entries", "maximum number of entries to keep in the packetcache")="500000";
-    ::arg().set("packetcache-servfail-ttl", "maximum number of seconds to keep a cached servfail entry in packetcache")="60";
+    ::arg().set("packetcache-servfail-ttl", "maximum number of seconds to keep a cached servfail entry in packetcache")="5";
     ::arg().set("server-id", "Returned when queried for 'server.id' TXT or NSID, defaults to hostname")="";
     ::arg().set("stats-ringbuffer-entries", "maximum number of packets to store statistics for")="10000";
     ::arg().set("version-string", "string reported on version.pdns or version.bind")=fullVersionString();


phonedph1:
Does your site produce a lot of ServFail's for clients and how did changing this impact your resolver performance? 

We produce a huge amount of ServFail's per second, because of some seriously broken setups out there. I am not sure reducing this so drastically would be a helpful default to most operators.

cnst:
@phonedph1:

Could you be more specific on the the source of so many SERVFAIL in you installation?

How long do you suggest SERVFAIL should be cached for?

I think you highly overestimate "reducing this so drastically" -- the change from 0 to 60s was made only in 3.2 (if docs are to be believed); change from 60s to 5s is hardly drastic, in that regard.

I also would like to point out my own bias in reducing it to only 5s.  I originally wanted to reduce it to just 5s to make sure you can still test that caching works, completely disregarding actual production use where each downtime (whether of server or client) would then be amplified by at least an extra 5 seconds of DNS being down, just for the point of having the feature more verifiable/testable from a user's perspective.  Actually writing down the rationale made me realise just how stupid it is, and that 5s already might as well be more than is reasonable as a default.

I would imagine the same bias might have struck the original implementer -- if they'd not cache for a minute, "noone would even notice the cool new feature".

But if we step back, we'll notice that noticing the downtime being amplified is hardly a feature that ought to be sought.

I'm open to discussion whether it should even be as high as 5s.  Perhaps 2s should be the default instead?

phonedph1:
There are a number of sites that use hostnames which exist on name servers only accessible on a local network but are still delegated on the internet's view. Some of these have extremely large NS sets which makes timing out fairly excessive on them. A good example of these are wpad lookups for internal names that literally occur almost constantly for the same name/clients. There are also hard-coded NTP names out there that are utterly terrible as well.

Before 4.x some of these would produce 50(!!) lookups by default before sending (and thankfully caching) SERVFAIL.  4.x is quite better in this regard but still can send a pretty large number of outgoing queries before failing (default timeouts are about 7, I think).

Changing from 60s to 5s will only amplify this amount of traffic.

We produce, across everything, thousands of servfail responses a second.

Habbie:
> I would imagine the same bias might have struck the original implementer -- if they'd not cache for a minute, "noone would even notice the cool new feature".

This is useless (and likely wrong) speculation. Also, the feature was done for an operator of a huge resolver farm, in fact - like @phonedph1. It serves a real need.

Habbie:
While we might consider lowering the default, we are not going to cap to 60 seconds. Even capping to the RFC mandate of 5 minutes feels wrong to me - operators should have freedom. I'm inclined to close this PR but I'll leave it open for a bit to allow operators to chime in.

cnst:
Can you kindly explain why do you feel that it's important to let people shoot themselves and their users in the foot?  There is no valid reason to allow caching for more than 5 minutes.

The cons of caching above a few seconds are clearly documented and easily understood.

What are the pros of caching for more than a minute?  Did anyone did any sort of a study to document why it may be so useful?  (How unique is each name that never resolves and must be cached?  How often do clients attempt to resolve it?  How many clients/name etc?  How many such names are served by a given NS set?  How many orgs delegate internal domains externally, and have the servers timeout?  Wouldn't it be much more effective to instead cache the availability of NS?)

If some setups are broken, breaking more things and allowing downtime amplification attacks seems like a wrong approach to take, especially if there is not even any sort of a study about the prevalence and details of these broken setups in the first place.

Habbie:
> There is no valid reason to allow caching for more than 5 minutes.

This, again, is idle speculation. Operators have real world needs to sometimes stretch parameters beyond what seems reasonable on paper. As an example, [the Recursor runtime minimum TTL feature](https://doc.powerdns.com/md/recursor/running/#set-minimum-ttl-value) obviously violates standards, by allowing the operator to override short TTLs and making them longer. However, this feature has in attack or overload situations made the difference between 'no service' and 'full service but some records may be a bit older than the auth admin intended'.

> especially if there is not even any sort of a study about the prevalence and details of these broken setups in the first place.

This could also be construed as an argument against changing anything :)

cnst:
The comparison with having a minimum TTL is not a valid one -- in the case of *minimum TTL*, you're *preventing* the downtime by having the ability to ignore absurdly low TTL values.

In the case *uncapped servfail-ttl*, you're *causing* the amplification of one.

> This could also be construed as an argument against changing anything :)

Was that meant to be a joke?  If not, then please construe it as an argument for backing out the original patch in full for causing downtime amplification. :)

phonedph1:
The vast majority of our servfails are because of incorrectly configured zones. It seems very unlikely that many of these would be fixed in the next 55 seconds.

In the event of a network connectivity issue, 60 seconds may seem excessive to you, which is why you can configure this as required. I would imagine on most networks if your resolvers can't talk to upstream auth servers, your clients can't talk to anything either.

cnst:
> The vast majority of our servfails are because of incorrectly configured zones. It seems very unlikely that many of these would be fixed in the next 55 seconds.

This is still a very vague description that ignores [my prior questions on the domain](https://github.com/PowerDNS/pdns/pull/4865#issuecomment-275184635); the solution screams "premature optimisation".

> In the event of a network connectivity issue, 60 seconds may seem excessive to you, which is why you can configure this as required. I would imagine on most networks if your resolvers can't talk to upstream auth servers, your clients can't talk to anything either.

No, the clients can still talk to the DNS server.  Also, as a result of default behaviour, they actually wouldn't be able to talk to anything else for as long as 60 seconds by default after the line comes back up, or even longer, if someone mistakenly thought to "optimise" the servfail-ttl to "fix" servfail "issues" further.

Seriously, I can't believe we're even having this argument.  If you have so much servfail going on, you better look at the underlying issues, and address those at the root (e.g., redelegate misbehaving zones, apply firewall rules, talk to upstream etc).

* Ignoring the fact that you just broke the internet for 60 extra seconds after a short downtime is an unacceptable way to look at this issue.

* Letting people break it for more than 60 seconds is nothing short of insane.

Habbie:
Closing this as it is not aligned with real world needs. Thank you for your submission!

Issue #4579, opened 2016-10-17
pdns_server 4.0.1 core dumps on Illumos/SmartOS when starting with allow-axfr-ips set


jussisallinen:
PowerDNS-4.0.1 core dumps on Illumos/SmartOS if allow-axfr-ips=127.0.0.0/8,::1
With allow-axfr-ips= empty pdns_server starts as it should.

Please ask for more specific output as needed.
[truss-pdns-4.0.1.txt](https://github.com/PowerDNS/pdns/files/533906/truss-pdns-4.0.1.txt)

[root@pkgtest ~]# pstack core.pdns_server.89125 | c++filt
core 'core.pdns_server.89125' of 89125: /opt/local/sbin/pdns_server --daemon=yes
 00000000005a07b4 NetmaskTree<bool>::insert(Netmask const&) () + 224
 000000000064d076 TCPNameserver::TCPNameserver() () + 2e6
 00000000005fbcc1 main () + 4921
 00000000004bf5cc _start () + 6c

[root@pkgtest ~]# ldd /opt/local/sbin/pdns_server
    libyaml-cpp.so.0.5 =>    /opt/local/lib//libyaml-cpp.so.0.5
    libGeoIP.so.1 =>     /opt/local/lib//libGeoIP.so.1
    libsocket.so.1 =>    /lib/64/libsocket.so.1
    libnsl.so.1 =>   /lib/64/libnsl.so.1
    libcrypto.so.1.0.0 =>    /opt/local/lib//libcrypto.so.1.0.0
    libsqlite3.so.0 =>   /opt/local/lib//libsqlite3.so.0
    libdl.so.1 =>    /lib/64/libdl.so.1
    librt.so.1 =>    /lib/64/librt.so.1
    libpthread.so.1 =>   /lib/64/libpthread.so.1
    libumem.so.1 =>  /lib/64/libumem.so.1
    libstdc++.so.6 =>    /opt/local/gcc49/x86_64-sun-solaris2.11/lib/amd64/libstdc++.so.6
    libm.so.2 =>     /lib/64/libm.so.2
    libssp.so.0 =>   /opt/local/gcc49/x86_64-sun-solaris2.11/lib/amd64/libssp.so.0
    libgcc_s.so.1 =>     /opt/local/gcc49/x86_64-sun-solaris2.11/lib/amd64/libgcc_s.so.1
    libc.so.1 =>     /lib/64/libc.so.1
[...]

PR #4877, opened 2017-01-11
issue #4579

skolot:
### Short description

in sunos byte order macroses LE_IN/BE_IN expected pointer to integer as argument

issue #4579 

### Checklist

I have:
- [ ] documented the code
- [ ] added regression tests
- [ ] added unit tests

### DIFF
--- a/pdns/iputils.hh
+++ b/pdns/iputils.hh
@@ -59,23 +59,22 @@
 #define le64toh(x) OSSwapLittleToHostInt64(x)
 #endif
 
-// for illumos
-#ifdef BE_64
+#ifdef __sun
 
 #define htobe16(x) BE_16(x)
 #define htole16(x) LE_16(x)
-#define be16toh(x) BE_IN16(x)
-#define le16toh(x) LE_IN16(x)
+#define be16toh(x) BE_IN16(&(x))
+#define le16toh(x) LE_IN16(&(x))
 
 #define htobe32(x) BE_32(x)
 #define htole32(x) LE_32(x)
-#define be32toh(x) BE_IN32(x)
-#define le32toh(x) LE_IN32(x)
+#define be32toh(x) BE_IN32(&(x))
+#define le32toh(x) LE_IN32(&(x))
 
 #define htobe64(x) BE_64(x)
 #define htole64(x) LE_64(x)
-#define be64toh(x) BE_IN64(x)
-#define le64toh(x) LE_IN64(x)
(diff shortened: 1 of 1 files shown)


pieterlexis:
@rgacogne could you review?

pdns/iputils.hh
```diff
@@ -59,23 +59,22 @@
 #define le64toh(x) OSSwapLittleToHostInt64(x)
 #endif
 
-// for illumos
-#ifdef BE_64
+#ifdef __sun
```
jeffpc: This will also be true on illumos, but in general isn't it considered bad form to check for specific systems instead of checking for features?

pdns/iputils.hh
```diff
@@ -59,23 +59,22 @@
...
 
 #define htobe16(x) BE_16(x)
 #define htole16(x) LE_16(x)
-#define be16toh(x) BE_IN16(x)
-#define le16toh(x) LE_IN16(x)
+#define be16toh(x) BE_IN16(&(x))
```
jeffpc: I don't understand why these use `{BE,LE}_INXX` instead of `{BE,LE}_XX`...

Habbie:
@jeffpc is any followup work required from your post-merge review of this?

rgacogne:
For the record, during the last discussion it was pointed out that the `{BE,LE}_INXX` function-like macros are documented as accepting "unaligned values" while the `{BE,LE}_XX` ones are not.

jeffpc:
IIRC, the change will work; it's just not necessarily the cleanest.  I don't know threshold in pdns for code cleanups/improvements.

I don't understand the asymmetry between the `hto{be,le}XX` and `{be,le}XXtoh`. Why is one using the IN versions while the other isn't?  Also, I just pointed out the use of a `__sun` guard.  Both are cleanliness issues IMO and nothing more.

rgacogne:
> I don't know threshold in pdns for code cleanups/improvements.

I'll be happy to do as most cleanup as possible once I'm sure it's not breaking anything :) 
 
> I don't understand the asymmetry between the hto{be,le}XX and {be,le}XXtoh. Why is one using the IN versions while the other isn't? 

I'm not sure that's the reason, but the host to network are far less likely to be used on non-aligned data than the network to host ones.

> Also, I just pointed out the use of a __sun guard.

Agreed, we should fix that no matter what we decide on the other issue.

Habbie:
@jeffpc want do a PR with your suggested cleanups? In the comments of this closed one we will all forget.

PR #4905, opened 2017-01-13
Revert "auth: In `Bind2Backend::lookup()`, use the `zoneId` when we have it"

rgacogne:
### Short description

This reverts commit 937a66255ff05f2e754ef113833e54cc4cf2004b.
It doesn't work with multiple backends since the `zoneId` is passed to
every available backend on `lookup()`.

### Checklist

I have:
- [ ] documented the code
- [ ] added regression tests
- [ ] added unit tests


zeha:
IMO this would be okay on master if the check *also* did a suffix match on qname. Then it's as good/bad as gsql.

PR #4908, opened 2017-01-14
Tidy up UeberBackend a bit

zeha:
### Short description
Code cleanup only. May improve compilation times.

### Checklist
I have:
- [X] compiled
- [ ] and tested this code
- [ ] documented the code
- [ ] added regression tests
- [ ] added unit tests

### DIFF
--- a/modules/remotebackend/unixconnector.cc
+++ b/modules/remotebackend/unixconnector.cc
@@ -22,6 +22,9 @@
 #ifdef HAVE_CONFIG_H
 #include "config.h"
 #endif
+#include <sys/types.h>
+#include <sys/socket.h>
+#include <sys/un.h>
 #include "remotebackend.hh"
 #ifndef UNIX_PATH_MAX 
 #define UNIX_PATH_MAX 108
--- a/pdns/ueberbackend.cc
+++ b/pdns/ueberbackend.cc
@@ -29,6 +29,7 @@
 #include "utility.hh"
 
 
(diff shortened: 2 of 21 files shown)


modules/remotebackend/unixconnector.cc
```diff
@@ -22,6 +22,9 @@
 #ifdef HAVE_CONFIG_H
 #include "config.h"
 #endif
+#include <sys/types.h>
+#include <sys/socket.h>
+#include <sys/un.h>
```
Habbie: commit message does not mention this - was this added intentionally?

pdns/ueberbackend.cc
```diff
@@ -29,6 +29,7 @@
 #include "utility.hh"
 
 
+#include <dlfcn.h>
```
Habbie: was this added intentionally?

zeha: this was just moved from ueberbackend.hh

modules/remotebackend/unixconnector.cc
zeha: they used to be in ueberbackend.hh, but nothing else appears to need them.

Habbie: Ah yes, works for me.

pdns/ueberbackend.cc
Habbie: :+1:

PR #4912, opened 2017-01-16
Refuse to start with chroot set in a systemd env

### LINKED ISSUE #4848: chroot option + systemd do not coexist well
There are various issues when running pdns with chroot= set under systemd.

IIRC:

  * notification to systemd fails because the systemd socket is unreachable
  * pdns_control has (with some specific settings) trouble talking to the daemon

Personally, I'd think refusing to start under systemd with chroot set would be good.

pieterlexis:
### Short description
When `chroot` is used while ruinning in systemd, PowerDNS can't use [`sd_notify()`](https://www.freedesktop.org/software/systemd/man/sd_notify.html)'s `NOTIFY_SOCKET` to send a notification to systemd that it started successfully, resulting in systemd believing the start up timed out.

This PR exits the server before starting up with a proper error message so the operator can take action.

Closes #4848

### Checklist

I have:
- [ ] documented the code
- [ ] added regression tests
- [ ] added unit tests

### DIFF
--- a/docs/markdown/recursor/settings.md
+++ b/docs/markdown/recursor/settings.md
@@ -129,6 +129,9 @@ in the configuration are relative to the new root.
 When using `chroot` and the API ([`webserver`](#webserver)), [`api-readonly`](#api-readonly)
 must be set and [`api-config-dir`](#api-config-dir) unset.
 
+When running on a system where systemd manages services, `chroot` does not work out of the box, as PowerDNS cannot use the `NOTIFY_SOCKET`.
+Either do not `chroot` on these systems or set the 'Type' of this service to 'simple' instead of 'notify' (refer to the systemd documentation on how to modify unit-files)
+
 ## `client-tcp-timeout`
 * Integer
 * Default: 2
--- a/pdns/common_startup.cc
+++ b/pdns/common_startup.cc
@@ -498,6 +498,14 @@ void mainthread()
    stubParseResolveConf();
 
    if(!::arg()["chroot"].empty()) {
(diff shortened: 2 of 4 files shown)


pdns/common_startup.cc
```diff
@@ -498,6 +498,14 @@ void mainthread()
    stubParseResolveConf();
 
    if(!::arg()["chroot"].empty()) {
+#ifdef HAVE_SYSTEMD
+     char *ns;
+     ns = getenv("NOTIFY_SOCKET");
+     if (ns != nullptr) {
+       L<<Logger::Error<<"Unable to chroot when running from systemd. Please disable chrooting or set the 'Type' for this service to 'simple'"<<endl;
```
zeha: `... Please disable chroot= or set ...`?

docs/markdown/recursor/settings.md
```diff
@@ -129,6 +129,9 @@ in the configuration are relative to the new root.
 When using `chroot` and the API ([`webserver`](#webserver)), [`api-readonly`](#api-readonly)
 must be set and [`api-config-dir`](#api-config-dir) unset.
 
+When running on a system where systemd manages services, `chroot` does not work out of the box, as PowerDNS cannot use the `NOTIFY_SOCKET`.
+Either don't `chroot` on these systems or set the 'Type' of the this service to 'simple' instead of 'notify' (refer to the systemd documentation on how to modify unit-files)
```
zeha: - `of the this`
- i think habbie usually wants `do not` not `don't`

PR #4915, opened 2017-01-16
Feature request: convert ws-auth to produce numerical stats.

rektide:
### Short description
Currently stats are produced as strings, even though they are numbers. This makes them very hard to consume from metrics systems (we're using New Relic Insights at work).

### Checklist

I have:
- [X] compiled the code
- [(could not figure out how to do this. regression-test instructions confusing, not direct!!)] ran tests on the code
- [ ] documented the code
- [ ] added regression tests
- [ ] added unit tests

### DIFF
--- a/pdns/rec_channel.hh
+++ b/pdns/rec_channel.hh
@@ -63,7 +63,7 @@ private:
   static bool s_init;
 };
 
-std::map<std::string, int> getAllStatsMap();
+std::map<std::string, uint64_t> getAllStatsMap();
 extern pthread_mutex_t g_carbon_config_lock;
 void sortPublicSuffixList();
 std::vector<std::pair<DNSName, uint16_t> >* pleaseGetQueryRing();
--- a/pdns/rec_channel.hh
+++ b/pdns/rec_channel.hh
@@ -63,7 +63,7 @@ private:
   static bool s_init;
 };
 
-std::map<std::string, std::string> getAllStatsMap();
+std::map<std::string, int> getAllStatsMap();
 extern pthread_mutex_t g_carbon_config_lock;
 void sortPublicSuffixList();
 std::vector<std::pair<DNSName, uint16_t> >* pleaseGetQueryRing();
--- a/pdns/rec_channel_rec.cc
+++ b/pdns/rec_channel_rec.cc
@@ -129,7 +129,7 @@ string getAllStats()
   varmap_t varmap = getAllStatsMap();
   string ret;
   for(varmap_t::value_type& tup :  varmap) {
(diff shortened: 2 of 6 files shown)


zeha:
Run tests for ws/api:

    make ...
    cd regression-tests.api 
    ./runtests authoritative

Ensure you have python + virtualenv installed first.

zeha:
- Recursor also needs conversion
- Some stats are 64bit items, so type for them can't be `int`

rektide:
Still working through this. First go at recursor, but it's a little more intensive & I have less knowledge of the project right now which will slow down some of my manual testing some.

I'm not sure whether "unsigned long" or uint64_t is preferred? The latter is just a typedef, no? I see some of each in different points in the codebase.

pdns/rec_channel.hh
```diff
@@ -63,7 +63,7 @@ private:
   static bool s_init;
 };
 
-std::map<std::string, std::string> getAllStatsMap();
+std::map<std::string, int> getAllStatsMap();
```
zeha: mismatched type between decl here and impl below.

zeha:
> I'm not sure whether "unsigned long" or uint64_t is preferred? The latter is just a typedef, no?

`unsigned long` is only guaranteed to fit 32bits.

Habbie:
Many JSON implementations limit integers to 53 bits; is this a concern?

zeha:
Right, I think we're limited to `double` actually?

Habbie:
@zeha exactly, which does lossless integers up to something around 53 bits (signed, I think)

rektide:
JSON the spec has no limit. Unfortunately as per https://github.com/dropbox/json11/issues/16 the JSON library we use insists that round-tripping a number through JSON in JS has higher precedence for their implementation than the spec, and refuses to offer any means to use large integers (tagged above issue wontfix).

So yes, metrics are- just before conversion to json- [cast to double](https://github.com/PowerDNS/pdns/pull/4915/files#diff-de98819b0ddb5a14027bafd29bd5a1d3R207).

pieterlexis:
@aerique I believe you were working on similar things with dnsdist, can you chime in?

aerique:
Yes, we 'fixed' it for dnsdist by casting to double as well. 53 bits should be enough for everyone.

See: https://github.com/PowerDNS/pdns/pull/5234

zeha:
@rektide mind rebasing this?

Habbie:
ping?

pieterlexis:
Closing for lack of response. We'll do a Prometheus endpoint in 4.3 or 4.4

PR #4921, opened 2017-01-18
dnsdist: Add `addWhitelistedDynBlocks()`

rgacogne:
### Short description

`addWhitelistedDynBlocks()` works similarly to `addDynBlocks` but takes a `NetmaskGroup` `whitelist` parameter containing a list of ranges to not block, even if they match.

### Checklist

I have:
- [x] documented the code
- [ ] added regression tests
- [ ] added unit tests

### DIFF
--- a/pdns/README-dnsdist.md
+++ b/pdns/README-dnsdist.md
@@ -783,6 +783,19 @@ representing the client's source address, and whose value is an integer represen
 the number of queries matching the corresponding condition (for example the
 `qtype` for `exceedQTypeRate()`, `rcode` for `exceedServFails()`).
 
+In order to prevent some ranges from being blocked, `addWhitelistedDynBlocks()` takes a
+`NetmaskGroup` as a fourth parameter named `whitelist`. The addresses it contains will not
+be blacklisted, even if they match.
+
+```
+whitelisted = newNMG()
+whitelisted:addMask("192.0.2.0/24")
+
+function maintenance()
+	addWhitelistedDynBlocks(exceedQRate(20, 10), "Exceeded query rate", 60, whitelisted)
+end
+```
+
 Dynamic blocks drop matched queries by default, but this behavior can be changed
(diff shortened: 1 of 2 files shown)


pdns/README-dnsdist.md
```diff
@@ -1425,6 +1438,7 @@ instantiate a server with additional parameters
     * `clearDynBlocks()`: clear all dynamic blocks
     * `showDynBlocks()`: show dynamic blocks in force
     * `addDynBlocks(addresses, message[, seconds])`: block the set of addresses with message `msg`, for `seconds` seconds (10 by default)
+    * `addWhitelistedDynBlocks(addresses, message[, seconds[, whitelist]])`: block the set of addresses with message `msg`, for `seconds` seconds. Addresses present in the`whitelist` `NetmaskGroup` will not be blocked even if they match
```
Habbie: Any reason you didn't just add another argument to `addDynBlocks`? I'm not sure about the naming of the new function either, maybe `addDynBlocksWithWhitelist` but I feel I'm channeling too much ObjC here ;)

rgacogne: > Any reason you didn't add another argument to `addDynBlocks`?

I tried but couldn't make it work, because mixing `boost::optional` with a reference to an object having a `boost::noncopyable` member wasn't going well IIRC. Passing a `boost::optional<const NetmaskGroup*>` was compiling fine but the passing a `NetmaskGroup` from Lua resulted then in:

`Error during execution of maintenance function: Unable to convert parameter from no value to N5boost8optionalIPK12NetmaskGroupEE`

I'll check again because I agree it would be so much nicer :) As for the name, `addDynBlocksWithWhitelist` was actually my first idea but it seemed to verbose, perhaps I should reconsider!

pieterlexis:
Needs a rebase

rgacogne:
Closing this PR, IMHO it should be implemented in the new `dynBlockRulesGroup` object.

Issue #3165, opened 2016-01-04
reverse lookups failed on pdns 3.4.7 and LDAP backend


tbrandysky:
Hello,

we've been using PowerDNS with LDAP backend for many years in company.
Recently we've updated some of our PowerDNS servers to version 3.4.7 and we've found out there is a problem with reverse DNS lookups while the configuration remains the very same as it was with 3.3.1 version which we'd use till then.

This is what happens:

PowerDNS 3.4.7 behaviour

```
dig -x 10.0.252.106 @<pdns_server_v3.4.7>

;; QUESTION SECTION:
;106.252.0.10.in-addr.arpa.     IN      PTR

in PowerDNS server log I can see:
Dec 11 14:53:27 server100 pdns[5630]: Backend reported permanent error which prevented lookup (Got non-SOA record when asking for SOA), aborting
Dec 11 14:53:27 server100 pdns[5630]: Backend error: Got non-SOA record when asking for SOA
```

PowerDNS 3.3.1 behaviour

```
dig -x 10.0.252.106 @<pdns_server_v3.3.1>

;; ANSWER SECTION:
106.252.0.10.in-addr.arpa. 86400 IN     PTR     server106.vm1.company.c.
```

this is our configuration (same on both PowerDNS servers except for local-address parameter):

```
module-dir=/usr/lib/pdns
socket-dir=/var/run/pdns-server
setuid=powerdns
setgid=powerdns

local-address=127.0.0.1,10.0.252.100                                                                                                                                                                                                        
local-port=53                                                                                                                                                                                                                               
recursor=127.0.0.1:5353                                                                                                                                                                                                                     

launch=ldap                                                                                                                                                                                                                                 
ldap-starttls=yes                                                                                                                                                                                                                           
ldap-host=ldap://10.0.252.116 ldap://10.0.252.117                                                                                                                                                                                           
ldap-basedn=ou=dns,ou=services,dc=company,dc=com                                                                                                                                                                                               
ldap-binddn=cn=dns,ou=systemusers,ou=users,dc=company,dc=com                                                                                                                                                                                
ldap-secret=XXXX                                                                                                                                                                                                                     
ldap-method=strict                                                                                                                                                                                                                          

distributor-threads=1                                                                                                                                                                                                                       
master=no                                                                                                                                                                                                                                   
slave=no                                                                                                                                                                                                                                    

negquery-cache-ttl=0                                                                                                                                                                                                                        
[...]

PR #4922, opened 2017-01-18
fix ldap-strict autoptr feature, including a test

Habbie:


Habbie:
Fixes #3165

zeha:
> do we need to init m_qtype in ctor

No. Calling ::get before ::lookup is wrong and broken, so having that uninit' is fine. (Actually, gsql would deref a nullptr in that path...)

Habbie:
Ready for review.

zeha:
lgtm

Issue #4849, opened 2017-01-03
Please ship all tarballs with a similar layout as git


zeha:
Right now the pdns tarball looks different than the dnsdist and pdns-recursor tarballs, and they do not mirror the layout in git.

This is painful for at least these:

  * applying/editing patches from git to tarballs
  * maintaining copyright notices (e.g. in debian/copyright)

omoerbeek:
This is not happening soon. Maybe when we change to a different build system...

Issue #2407, opened 2015-04-02
Directory layout should be modernized


pieterlexis:
Will be a big-bang commit (break ALL the things)!

suggestions:
* Move `/pdns/ext` to `/ext`
* Rename `/pdns` to `/src`

@rubenk what's your opinion on moving `/modules` to `/src/modules`?

rubenk:
@pieterlexis sounds good!

cmouse:
If you are going to muck the directories about, how about a separate build directory as well to keep objects out of source code?

cmouse:
http://www.gnu.org/software/automake/manual/html_node/VPATH-Builds.html

Habbie:
I believe @rubenk has done a lot of work to make this work already

cmouse:
Just wondering if we should do this by default, so that the code gets always built in build directory?

rubenk:
make distcheck handles this for you. The plan was to eventually let the builders always run this (#1604).

PR #4924, opened 2017-01-18
Make EDNS Client Subnet masks tunable, improve logging, improve subnet oddity

ahupowerdns:
### Short description
EDNS Client Subnet can reveal x bits of the client subnet to the authoritative server. This makes x tunable. Plus it pretties the logging, and fixes an oddity where we'd return netmasks like '1.2.3.4/8'.
### Checklist

I have:
- [X] documented the code
- [ ] added regression tests
- [ ] added unit tests

### DIFF
--- a/pdns/iputils.hh
+++ b/pdns/iputils.hh
@@ -276,7 +276,7 @@ union ComboAddress {
       return "["+toString() + "]:" + std::to_string(ntohs(sin4.sin_port));
   }
 
-  void truncate(unsigned int bits);
+  void truncate(unsigned int bits) noexcept;
 };
 
 /** This exception is thrown by the Netmask class and by extension by the NetmaskGroup class */
--- a/pdns/recursordist/ecs.cc
+++ b/pdns/recursordist/ecs.cc
@@ -1,0 +1,38 @@
+#include "syncres.hh"
+#include "arguments.hh"
+
+NetmaskGroup g_ednssubnets;
+SuffixMatchNode g_ednsdomains;
+
+boost::optional<Netmask> getEDNSSubnetMask(const ComboAddress& local, const DNSName&dn, const ComboAddress& rem)
+{
+  static int l_ipv4limit, l_ipv6limit;
+  if(!l_ipv4limit) {
+    l_ipv4limit = ::arg().asNum("ecs-ipv4-bits");
+    l_ipv6limit = ::arg().asNum("ecs-ipv6-bits");
+  }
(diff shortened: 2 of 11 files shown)


pdns/syncres.cc
```diff
@@ -1118,8 +1118,16 @@ int SyncRes::doResolveAt(NsSet &nameservers, DNSName auth, bool flawedNSSet, con
...
+              if(ednsmask) {
+                LOG(prefix<<qname<<": Adding EDNS Client Subnet Mask "<<ednsmask->toString()<<" to query"<<endl);
+              }
 	      resolveret=asyncresolveWrapper(*remoteIP, d_doDNSSEC, qname,  qtype.getCode(),
 					     doTCP, sendRDQuery, &d_now, ednsmask, &lwr);    // <- we go out on the wire!
+              if(ednsmask) {
+                LOG(prefix<<qname<<": Received EDNS Client Subnet Mask "<<ednsmask->toString()<<" on response"<<endl);
+              }
+
+
```
Habbie: trim some whitespace here?

pdns/iputils.hh
```diff
@@ -276,7 +276,7 @@ union ComboAddress {
       return "["+toString() + "]:" + std::to_string(ntohs(sin4.sin_port));
   }
 
-  void truncate(unsigned int bits);
+  void truncate(unsigned int bits) noexcept;
```
Habbie: my reading of http://en.cppreference.com/w/cpp/language/noexcept_spec does not tell me that `noexcept` -enforces- anything. There is no downside to having the keyword there, of course.

pdns/recursordist/ecs.cc
```diff
@@ -0,0 +1,38 @@
+#include "syncres.hh"
```
Habbie: So far recursordist has not seen actual .cc files, just symlinks one dir up. However, I feel that we should be moving in this direction anyway (i.e. unshared files -should- live in recursordist/), so unless @pieterlexis disagrees, I suggest keeping it like you did.

pieterlexis: I agree, but then there's also #4849 and #2407 :)

Habbie: Right. Neither are reasons to stop this PR I think?

pieterlexis: no, from a functionality perspective, this is fine :). It is just a little inconsistent. Merging is fine by me.

PR #4925, opened 2017-01-19
rec: Let ASAN know that we are switching stacks, which one is in use

rgacogne:
### Short description
Enabled via the --enable-asan configure switch if the `__sanitizer_start_switch_fiber` function is present in `sanitizer/common_interface_defs.h` (currently only supported by `LLVM/clang` AFAIK).

### Checklist

I have:
- [ ] documented the code
- [ ] added regression tests
- [ ] added unit tests


Habbie:
I am not qualified to review this but I did anyway, and it looks good to me. If @pieterlexis can review the autotools bit, this should be fine.

PR #4944, opened 2017-01-26
Fix API performance: one DNSSECKeeper per request

wojas:
### Short description

getZoneInfo() was instantiating a new DNSSECKeeper for every call. Since
DNSSECKeeper opens a new database connection, this resulted in severe
performance issues in /api/v1/servers/localhost/zones with many zones.

This is fixed by passing a DNSSECKeeper instance to getZoneInfo().
With 10,000 zones and the sqlite3 backend, this reduces the API call
execution time from ~2s to ~200ms (from ~7s to 700ms for first call).

### Checklist

I have:
- [ ] documented the code
- [ ] added regression tests
- [ ] added unit tests

### DIFF
--- a/pdns/ws-auth.cc
+++ b/pdns/ws-auth.cc
@@ -325,7 +325,7 @@ static void fillZone(const DNSName& zonename, HttpResponse* resp) {
   if(!B.getDomainInfo(zonename, di))
     throw ApiException("Could not find domain '"+zonename.toString()+"'");
 
-  DNSSECKeeper dk;
+  DNSSECKeeper dk(&B);
   Json::object doc = getZoneInfo(di, &dk);
   // extra stuff getZoneInfo doesn't do for us (more expensive)
   string soa_edit_api;
--- a/pdns/ws-auth.cc
+++ b/pdns/ws-auth.cc
@@ -302,8 +302,7 @@ static inline string makeBackendRecordContent(const QType& qtype, const string& 
   return makeRecordContent(qtype, content, true);
 }
 
-static Json::object getZoneInfo(const DomainInfo& di) {
-  DNSSECKeeper dk;
+static Json::object getZoneInfo(const DomainInfo& di, DNSSECKeeper *dk) {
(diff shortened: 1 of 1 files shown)


zeha:
As said elsewhere, I think the `dnssec` flag can just go away, and with it goes the need for DNSSECKeeper in listZones.

wojas:
@zeha The flag is currently being used. Removing it and requiring applications that need it for whatever reason to do individual calls for every single zone would not improve overal performance of the API.

How about adding an `?extra=...` GET parameter to this API call and only returning the `dnssec` field if you pass `?extra=dnssec`? This mechanism can also be easily extended to add other fields that the calling application might need (`?extra=dnssec,foo,bar`). The call takes slightly longer when the extra info is needed, but can save many individual requests that would otherwise be required. Users that don't need the extra fields simply do not provide it.

Such a change would be breaking for applications that use this field, but they can easily be fixed by adding the `extra` param. Adding this param to an application would also be backward compatible, as auth 4.0 will simply ignore it and always return the field.

wojas:
One way to keep it backward compatible is to default to`?extra=dnssec` if the extra param is not set. To to a light query you could pass an empty `?extra=`.

Based on my personal quick testing removing the field improves performance by 10-15% after this patch. This is significant enough to provide a way to not incur the overhead, but not to completely remove the functionality.

pdns/ws-auth.cc
```diff
@@ -326,7 +325,8 @@ static void fillZone(const DNSName& zonename, HttpResponse* resp) {
   if(!B.getDomainInfo(zonename, di))
     throw ApiException("Could not find domain '"+zonename.toString()+"'");
 
-  Json::object doc = getZoneInfo(di);
+  DNSSECKeeper dk;
```
zeha: Please make this `DNSSECKeeper dk(&B)` if that works.

wojas:
@zeha I made the suggested change to pass UeberBackend to DNSSECKeeper, thanks!

PR #4956, opened 2017-01-30
Improve slave-check: Check every configured master and AXFR from mast…

klaus3000:
…er with highest serial

Currently PDNS as slave has a poor logic when there are multiple masters:

1. On incoming NOTIFY, PDNS queues the zone for refresh. Then, the SOA query uses a
random master to get the master's serial. On timeout, PDNS does not try another master.

2. If the SOA query was successful and the serial increased, PDNS queues the domain for
AXFR, but this AXFR uses only the first master.

Thus, if the first master is offline, even if the second master sends
NOTIFYs and has a newer zone, PDNS will never transfer it.

This patch does 2 things:

a) If a slave check for a domain is requested and the domain has multiple masters, the SOA
check will be performed against every master. 
b) Further, the name server which answered with
the highest serial will be used for the AXFR.

Known issue: If there is a serial rollover, then choosing the highest serial is not the best
choice. But as soon as the rollover happend on every master the logic is correct again.

This pull request needs some more polishing (eg. replace some log messages with DLOG), but I wanted to ask first if you would accept this.

The same patch for 3.4 is running in production since ~1 year with nearly 1mio slave domains without problems.

### DIFF
--- a/docs/Makefile.am
+++ b/docs/Makefile.am
@@ -58,10 +58,10 @@ endif # if !HAVE_MANPAGES
 	.venv/bin/pip install -U pip setuptools setuptools-git
 	.venv/bin/pip install -r requirements.txt
 
-html-docs: common/** manpages/** .venv
+html-docs: common/** manpages/** .venv *.rst
 	.venv/bin/python -msphinx -b html . html-docs
 
-latex/PowerDNS-Authoritative.pdf: common/** manpages/** .venv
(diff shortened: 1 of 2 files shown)


klaus3000:
The patch did not worked with current master any more. I wanted to update the patch but during this the PR was automatically closed. Hence I opened a new one: https://github.com/PowerDNS/pdns/pull/5595

PR #4965, opened 2017-02-01
Feature request: PowerDNS Forward-Notify Patch

DrRemorse:
### Short description

This patch will allow you to redirect inbound notifications to a proxy server.  It's intended use is in anycast environments where it might be necessary for a proxy server to preform the AXFR.

The configuration option "forward-notify" has been added to the pdns.conf parser. The option accepts multiple IPv4 and IPv6 address values.

Obsoletes #4964 

### Checklist

I have:
- [ ] documented the code
- [ ] added regression tests
- [ ] added unit tests

### DIFF
--- a/docs/markdown/authoritative/settings.md
+++ b/docs/markdown/authoritative/settings.md
@@ -364,10 +364,10 @@ Forward DNS updates sent to a slave to the master.
 ## `forward-notify`
 * IP addresses, separated by commas
 
-IP addresses to send received notifications to regardless of master or slave settings.
+IP addresses to forward received notifications to regardless of master or slave settings.
 
 Note: The intended use is in anycast environments where it might be necessary for a
-proxy server to preform the AXFR.  The usual checks are preformed before any received
+proxy server to perform the AXFR.  The usual checks are performed before any received
 notification is forwarded.
 
 ## `guardian`
--- a/docs/markdown/authoritative/settings.md
+++ b/docs/markdown/authoritative/settings.md
@@ -361,6 +361,15 @@ If this is disabled (the default), ALIAS records will not expanded and the serve
 
 Forward DNS updates sent to a slave to the master.
 
+## `forward-notify`
+* IP addresses, separated by commas
+
+IP addresses to send received notifications to regardless of master or slave settings.
+
+Note: The intended use is in anycast environments where it might be necessary for a
+proxy server to preform the AXFR.  The usual checks are preformed before any received
+notification is forwarded.
+
 ## `guardian`
 * Boolean
 * Default: no
--- a/pdns/communicator.cc
+++ b/pdns/communicator.cc
@@ -92,6 +92,19 @@ void CommunicatorClass::go()
       exit(1);
     }
   }
+
+  vector<string> forwards;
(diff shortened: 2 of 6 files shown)


pdns/communicator.cc
```diff
@@ -92,6 +92,19 @@ void CommunicatorClass::go()
       exit(1);
     }
   }
+
```
Habbie: This is basically a copy of the for-loop right above, perhaps we can factor this out?

pdns/packethandler.cc
```diff
@@ -53,6 +53,8 @@
  
 AtomicCounter PacketHandler::s_count;
 NetmaskGroup PacketHandler::s_allowNotifyFrom;
+set<string> PacketHandler::s_forwardNotify;
```
Habbie: Why is this `string` and not `ComboAddress`?

Habbie: Ah, because d_alsoNotify works like that. Not your fault then.

Habbie:
Reviewed. I took the liberty of unchecking two checkboxes that were erroneously checked.

pdns/communicator.cc
DrRemorse: Thanks for reviewing this PR!  I'll take a look.

pdns/communicator.cc
```diff
@@ -37,6 +37,21 @@
 #include "arguments.hh"
 #include "packetcache.hh"
 
+#define CommunicatorLoadArgsIntoSet(listname,listset) do {                                 \
```
zeha: this looks like it could be a plain function?

pdns/communicator.cc
```diff
@@ -37,6 +37,21 @@
 #include "arguments.hh"
 #include "packetcache.hh"
 
+#define CommunicatorLoadArgsIntoSet(listname, listset) do {                                \
```
Habbie: Please change this into a function.

docs/markdown/authoritative/settings.md
```diff
@@ -361,6 +361,15 @@ If this is disabled (the default), ALIAS records will not expanded and the serve
 
 Forward DNS updates sent to a slave to the master.
 
+## `forward-notify`
+* IP addresses, separated by commas
+
+IP addresses to send received notifications to regardless of master or slave settings.
```
Habbie: s/send/forward/ ?

docs/markdown/authoritative/settings.md
```diff
@@ -361,6 +361,15 @@ If this is disabled (the default), ALIAS records will not expanded and the serve
...
+
+Note: The intended use is in anycast environments where it might be necessary for a
+proxy server to preform the AXFR.  The usual checks are preformed before any received
```
Habbie: s/preform/preform/g

pdns/communicator.hh
```diff
@@ -187,6 +187,7 @@ public:
   }
   bool notifyDomain(const DNSName &domain);
 private:
+  void LoadArgsIntoSet(const char *listname, set<string> &listset);
```
Habbie: Lowercase first `l` please.

PR #5006, opened 2017-02-13
docs: improve documentation of MySQL foreign keys

peterthomassen:
So far, the docs mentioned the foreign key for the records table only.
However, if foreign keys are used, they also make sense for the
domainmetadata, cryptokeys, and comments tables.

Corresponding SQL is already in an .sql file in the repository. This
commit replaces the incomplete information in the docs with a reference
to that file.


pieterlexis:
Hi, thanks for this, one nit, could you use 
 ```
 !!include=../modules/gmysqlbackend/enable-foreign-keys.mysql.sql
```
To include the sql in the docs (see lower down in the file you edited)

peterthomassen:
Done. I was not aware of that syntax.

PR #5062, opened 2017-02-21
Check for valid hostnames in SRV, NS and MX records

### LINKED ISSUE #512: check-zone does not check NS RDATA NSDNAME
If the content of an NS record is 'http://example.com/', check-zone does not complain. It should.

pieterlexis:
### Short description
Validate if records that contain a hostname in the RDATA actually have a valid hostname in the RDATA. 

This does not include CNAMEs, as they could be an alias for non-hostname names (i.e. `_443._tcp.www.example.com CNAME _cert.example.com` to only have one TLSA in the zone)

Fixes #512

### Checklist

I have:
- [x] documented the code

### DIFF
--- a/pdns/pdnsutil.cc
+++ b/pdns/pdnsutil.cc
@@ -579,13 +579,27 @@ int checkZone(DNSSECKeeper &dk, UeberBackend &B, const DNSName& zone, const vect
       }
     }
 
-    if(rr.qtype.getCode() == QType::A || rr.qtype.getCode() == QType::AAAA)
-    {
-      Regex hostnameRegex=Regex("^(([A-Za-z0-9]([A-Za-z0-9-]*[A-Za-z0-9])?)\\.)+$");
-      if (!hostnameRegex.match(rr.qname.toString()))
-      {
-        cout<<"[Info] A or AAAA record found at '"<<rr.qname.toString()<<"'. This name is not a valid hostname."<<endl;
-        continue;
+    if((rr.qtype.getCode() == QType::A || rr.qtype.getCode() == QType::AAAA) && !rr.qname.isWildcard() && !rr.qname.isHostname())
+      cout<<"[Info] "<<rr.qname.toString()<<" record for '"<<rr.qtype.getName()<<"' is not a valid hostname."<<endl;
+
+    // Check if the DNSNames that should be hostnames, are hostnames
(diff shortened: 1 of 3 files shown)


pdns/pdnsutil.cc
```diff
@@ -139,6 +139,12 @@ void loadMainConfig(const std::string& configdir)
   UeberBackend::go();
 }
 
+bool hostnameRegexCheck(const string& name)
+{
+  Regex hostnameRegex=Regex("^(([A-Za-z0-9]([A-Za-z0-9-]*[A-Za-z0-9])?)\\.)+$");
+  return hostnameRegex.match(name);
+}
+
 // irritatingly enough, rectifyZone needs its own ueberbackend and can't therefore benefit from transactions outside its scope
```
mind04: Move this to DNSName isHostname() ?

pdns/pdnsutil.cc
```diff
@@ -579,13 +585,30 @@ int checkZone(DNSSECKeeper &dk, UeberBackend &B, const DNSName& zone, const vect
       }
     }
 
-    if(rr.qtype.getCode() == QType::A || rr.qtype.getCode() == QType::AAAA)
+    if((rr.qtype.getCode() == QType::A || rr.qtype.getCode() == QType::AAAA) && !hostnameRegexCheck(rr.qname.toString()))
     {
```
mind04: Wildcards are valid for A and AAAA but do not match the hostnameRegexCheck

pdns/pdnsutil.cc
```diff
@@ -579,13 +585,30 @@ int checkZone(DNSSECKeeper &dk, UeberBackend &B, const DNSName& zone, const vect
...
-      Regex hostnameRegex=Regex("^(([A-Za-z0-9]([A-Za-z0-9-]*[A-Za-z0-9])?)\\.)+$");
-      if (!hostnameRegex.match(rr.qname.toString()))
-      {
-        cout<<"[Info] A or AAAA record found at '"<<rr.qname.toString()<<"'. This name is not a valid hostname."<<endl;
-        continue;
+      cout<<"[Info] A or AAAA record found at '"<<rr.qname.toString()<<"'. This name is not a valid hostname."<<endl;
+      continue;
+    }
```
mind04: Don't skip the checks below if an invalid hostname was found.

PR #5067, opened 2017-02-22
FreeBSD build notes and compile fixes.

cynthia:
### Short description
This adds a note for building on FreeBSD and fixes a compile error in json11.cpp.

It might make sense to look into setting c++11 for clang instead of gnuc++11 under clang, but the regression scope of that is a bit unclear due to the amount of versions of XCode and FreeBSD releases (each running different versions of clang!) out in the wild.

For the time being, I felt the safer route was to document a flag override only for FreeBSD, as that makes the regression surface minimal to non-existent.

### Checklist
I have:
- [x] documented the code

The one liner include change should have no side effects, so I have omitted extra tests to cover this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,6 +89,10 @@ FREEBSD NOTES
 -------------
 You need to compile using gmake - regular make only appears to work, but doesn't in fact. Use gmake, not make.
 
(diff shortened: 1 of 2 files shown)


rgacogne:
Thank you for this PR! Just one thing, it would be nice if the header addition to `json.cpp` could be fixed upstream at https://github.com/dropbox/json11, would you mind opening a PR there?

PR #5068, opened 2017-02-22
Change from time_t to uint32_t for serial in calculateSOASerial

### LINKED ISSUE #1010: calculateSOASerial should use uint32_t for serial instead time_t
```
dnsbackend.hh:  virtual bool calculateSOASerial(const string& domain, const SOAData& sd, time_t& serial);
```

This should be `uint32_t`, because `time_t` is not always a 32 bit unsigned integer.

pieterlexis:
### Short description
time_t can be bigger, which could lead to overflow/underflow issues.

Closes #1010

### Checklist

I have:
- [ ] documented the code

### DIFF
--- a/pdns/dnsbackend.cc
+++ b/pdns/dnsbackend.cc
@@ -250,7 +250,7 @@ bool DNSBackend::getSOA(const DNSName &domain, SOAData &sd)
   if(!sd.serial) { // magic time!
     DLOG(L<<Logger::Warning<<"Doing SOA serial number autocalculation for "<<rr.qname<<endl);
 
-    time_t serial;
+    uint32_t serial = 0;
     if (calculateSOASerial(domain, sd, serial)) {
       sd.serial = serial;
       //DLOG(L<<"autocalculated soa serialnumber for "<<rr.qname<<" is "<<newest<<endl);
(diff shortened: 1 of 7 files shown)


pdns/dnsbackend.cc
```diff
@@ -330,7 +330,7 @@ bool DNSBackend::getBeforeAndAfterNames(uint32_t id, const DNSName& zonename, co
  * \param sd Information about the SOA record already available
  * \param serial Output parameter. Only inspected when we return true
  */
-bool DNSBackend::calculateSOASerial(const DNSName& domain, const SOAData& sd, time_t& serial)
+bool DNSBackend::calculateSOASerial(const DNSName& domain, const SOAData& sd, uint32_t& serial)
 {
     // we do this by listing the domain and taking the maximum last modified timestamp
```
rgacogne: If we do this, we should probably make `newest` a `uint32_t` too then, and perhaps `DNSResourceRecord::last_modified` as well?

pieterlexis: Good one, although `last_modified` could be bigger when it comes from the backend (after 2038) so we might need some magic here to truncate properly?

Habbie: Won't truncation be properly by default, just by merit of stuffing a bigger number into a smaller one?

Habbie:
This PR has conflicts.

Habbie:
ping

PR #5070, opened 2017-02-22
dnsdist addAction: Also DNSName(s)

pieterlexis:
### Short description
This adds the possibility to feed a DNSName, or table of DNSNames to addAction.

### Checklist

I have:
- [ ] documented the code

### DIFF
--- a/pdns/dnsdist-lua.cc
+++ b/pdns/dnsdist-lua.cc
@@ -91,16 +91,13 @@ private:
   func_t d_func;
 };
 
-typedef boost::variant<string,vector<pair<int, string>>, std::shared_ptr<DNSRule> > luadnsrule_t;
-
 std::shared_ptr<DNSRule> makeRule(const luadnsrule_t& var)
 {
-  if(auto src = boost::get<std::shared_ptr<DNSRule>>(&var))
-    return *src;
-  
+  if (var.type() == typeid(std::shared_ptr<DNSRule>))
+    return *boost::get<std::shared_ptr<DNSRule>>(&var);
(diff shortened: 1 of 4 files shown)


pdns/dnsdist-lua.cc
```diff
@@ -91,13 +91,13 @@ class LuaResponseAction : public DNSResponseAction
   func_t d_func;
 };
 
-typedef boost::variant<string,vector<pair<int, string>>, std::shared_ptr<DNSRule> > luadnsrule_t;
+typedef boost::variant<string, vector<pair<int, string>>, std::shared_ptr<DNSRule>, DNSName, vector<pair<int, DNSName>> > luadnsrule_t;
```
rgacogne: Not your doing but this line should go, it's redundant with the one in `dnsdist-lua.hh`

pdns/dnsdist-lua.cc
```diff
@@ -110,11 +110,15 @@ std::shared_ptr<DNSRule> makeRule(const luadnsrule_t& var)
   };
   if(auto src = boost::get<string>(&var))
     add(*src);
-  else {
-    for(auto& a : boost::get<vector<pair<int, string>>>(var)) {
+  else if (auto src = boost::get<vector<pair<int, string>>>(&var))
```
rgacogne: g++ with `-Wshadow` is going to complain we declare several `src` in the same scope, perhaps we could avoid that?

pieterlexis:
Updated implementation to use `typeid` and added tests!

PR #5078, opened 2017-02-23
Throw an error when lua-conf-file can't be loaded

### LINKED ISSUE #4939: Bogus lua-config-file is ignored
- Program: Recursor
 - Issue type: Bug report

### Short description
If I point `lua-config-file` to a non-existing path, this appears to be silently ignored and the Recursor just starts up.

### Environment

 - Operating system: Debian 8.0
 - Software version: 4.0.3
 - Software source: repo

pieterlexis:
### Short description
This ensures we cannot start up if the file is unreadable.
Closes #4939

### Checklist

I have:
- [ ] documented the code


pieterlexis:
Interestingly, we already catch errors with a `rec_control reload-lua-config`

Issue #5086, opened 2017-02-27
Docs: Ldap-backend strict mode requires corresponding in-addr.arpa zone with SOA 


shantikulkarni:
- Program: Authoritative (ldap backend)
 - Issue type: Docs

### Short description
Please update ldap backend documentation to add that an reverse lookup zone with an SOA attribute is now required for strict mode reverse lookups. This was not the case for version 3.3.3 or prior.

### Description
I upgraded from 3.3.3 to git master (cbf3062), and found that my ldap strict mode reverse lookups were broken (either forwarded, or returned refused).  This was confusing because I've been using this tree since ~2.9.x. Eventually I decided to try creating an entry for the xxx.xxx.in-addr.arpa zone it was seeking, with an SOA attribute, and it started working again.  Since this is a change in behavior from earlier versions, I suggest the docs for ldap-backend should include that strict mode reverse lookups will fail without a unless there is a reverse lookup zone corresponding to the address, with a valid SOA attribute.

Habbie:
Agreed! If you want to speed this along, you can go to https://github.com/PowerDNS/pdns/blob/master/docs/markdown/authoritative/backend-ldap.md, edit it yourself, and file a pull request. Otherwise, this ticket you already wrote will make sure somebody will get to it eventually. Thanks!

Habbie:
Fixed in #5098. Thanks!

PR #5098, opened 2017-03-02
SOA object is required for all reverse zones

shantikulkarni:
### Short description
Updates documentation to reflect that reverse lookups require an object with an SOA entry for the reverse lookup zone, including in LDAP strict mode.  This requirement is a change between 3.3.3, which did not require a reverse zone SOA in strict mode, and git master. (Issue #5086)

### DIFF
--- a/docs/markdown/authoritative/backend-ldap.md
+++ b/docs/markdown/authoritative/backend-ldap.md
@@ -101,7 +101,7 @@ DN of a tree style example record (e.g. myhost.test.example.com):
 `dn:dc=myhost,dc=test,dc=example,dc=com,ou=hosts,...`
 
 ## Basic objects
-Each domain (or zone for BIND users) must include one object containing a SOA (Start Of Authority) record.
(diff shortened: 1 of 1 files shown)


docs/markdown/authoritative/backend-ldap.md
```diff
@@ -212,6 +211,9 @@ nsrecord:ns.example.com
 associateddomain:1.10.in-addr.arpa
 ```
 
+In `strict` mode, no other objects are required -- reverse queries that correspond to an arecord or aaaarecord of an existing object will be automagically serviced using the associateddomain entry of that object.
+
+In `simple` mode, you must then create objects for each reverse mapping:
```
Habbie: Also in `tree`, right?

shantikulkarni: The simple mode example is immediately followed by the tree mode example from the current version of the docs ("Tree mode requires each component to be a dc element of its own...").

Habbie: Ah, yes!

PR #5102, opened 2017-03-02
rec: Packet cache speedup / cleanup

rgacogne:
### Short description

* Don't parse the response's qname for every call to `getResponsePacket()`, this leads to a ~15% speed up on pure retrieval
* Only hash once, keep the hash result around, leading to a ~40% speed up on insertion
* Be consistent about the signedness of the packet cache's tag
* Consistently use a `time_t` for the packet cache's TTD, instead of mixing `time_t` and `uint32_t`
* Don't parse the qname again in the packet cache if we already have it, i.e. when `gettag()` exists

### Checklist

I have:
- [x] documented the code

### DIFF
--- a/pdns/recpacketcache.cc
+++ b/pdns/recpacketcache.cc
@@ -31,7 +31,7 @@ int RecursorPacketCache::doWipePacketCache(const DNSName& name, uint16_t qtype, 
       if(iter->d_name != name)
 	break;
     }
-    
+
     if(qtype==0xffff || iter->d_type == qtype) {
       iter=idx.erase(iter);
       count++;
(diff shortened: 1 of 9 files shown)


pdns/recpacketcache.cc
```diff
@@ -236,7 +261,7 @@ uint64_t RecursorPacketCache::doDump(int fd)
   for(auto i=sidx.cbegin(); i != sidx.cend(); ++i) {
     count++;
     try {
-      fprintf(fp, "%s %d %s  ; tag %d\n", i->d_name.toString().c_str(), (int32_t)(i->d_ttd - now), DNSRecordContent::NumberToType(i->d_type).c_str(), i->d_tag);
+      fprintf(fp, "%s %ld %s  ; tag %d\n", i->d_name.toString().c_str(), (i->d_ttd - now), DNSRecordContent::NumberToType(i->d_type).c_str(), i->d_tag);
```
ahupowerdns: boom broken on 32 bit time_t platforms with 64 bit time_t?

rgacogne: It's probably going to be more of an issue on 64bit platforms with a 32bit `time_t`, but indeed I should fix that!

PR #5103, opened 2017-03-02
rec: Use one listening socket per thread when reuseport is enabled

rgacogne:
### Short description
Except if `pdns-distributes-queries` is true, of course.
We used to share the same listening socket between all threads when `pdns-distributes-queries` is set to false, even with `reuseport` set to true and `SO_REUSEPORT` support available. After this commit:
* if `pdns-distributes-queries` is true, the distributor thread is still the only one listening to incoming queries
* if `pdns-distributes-queries` is false and either `reuseport` is false or `SO_REUSEPORT` support is not available, all threads share the same listening socket as it was before
* if `pdns-distributes-queries` is false, `SO_REUSEPORT` support is available and `reuseport` is true, we open a separate listening socket per thread to let the kernel distribute the incoming queries for us, avoiding any thundering herd issue as well as the distributor thread being a bottleneck.

### Checklist
I have:
- [x] documented the code


pieterlexis:
The commit message is clearer than the documentation included in the PR. Could you either open a ticket that we need to document this interaction better, or update `docs/markdown/recursor/performance.md` to include this info so we can point to it?

rgacogne:
Sure, I'll update `docs/markdown/recursor/performance.md`!

rgacogne:
@pieterlexis Updated!

PR #5112, opened 2017-03-06
Stubresolver: Use only `resolver` setting if given

### LINKED ISSUE #4655: Stub resolver does not use DNSproxy or ::arg()["recursor"] for ALIAS records if /etc/resolv.conf has "nameserver" entries
- Program: Authoritative
 - Issue type: Bug report

### Short description

I discovered an issue, where the correct A and AAAA records for an ALIAS record can not be resolved.

What I want to achieve is updating the A and AAAA records for an ALIAS record during the AXFR of the zone including the ALIAS record. The special thing is, that the domain to which the ALIAS points is also included in the same zone. Here's the part of the zone which is important for that problem:

    origin.example.com. A     127.0.0.1
    origin.example.com. AAAA  ::1
    alias.example.com.  ALIAS origin.example.com.

The problem is, that the PowerDNS server serving this zone is a hidden master. Thus, it is not reachable from the outside world. Its purpose is generating AXFR answers for an OpenDNSSEC service running on the same machine, which is itself creating AXFRs for the world reachable NSD which is presenting the signed zones to the world.

If an AXFR request is send from the OpenDNSSEC service to the PowerDNS server (hidden master), the ALIAS record is resolved to origin.example.com from the outside world (the public reachable NSD server). That is not what I want to achieve. I want to achive, that the ALIAS is resolved to origin.example.com on the PowerDNS server (hidden master). This is necessary, when origin.example.com is updated in the database of the hidden master server and a new AXFR is send to OpenDNSSEC for signing.

To solve this problem, I set up a local resolver (127.0.0.1:5678) with forwarding for the `example.com` domain to the hidden master (PowerDNS) and inserted the `resolver` and `outgoing-axfr-expand-alias` option to my `pdns.conf`:

    local-address=127.0.0.1
    local-port=1234
    resolver=127.0.0.1:5678
    outgoing-axfr-expand-alias=on

Here's a littel scheme of my setup.

               req        AXFR              AXFR
    Internet ------> NSD <----- OpenDNSSEC <----- PowerDNS <----- SQlite
                                                  |     ^
                                                 \|/    |
                                  Internet <----- Unbound
                                             req    

### Environment

 - Operating system: Ubuntu 14.04 LTS
 - Software version: 4.0.1

### Steps to reproduce

1. Setup as described above
2. Insert the above records into SQLite database
3. Query: `dig @127.0.0.1 -p 1234 alias.example.com.` and watch the log of the Unbound resolver. No requests are comming in from PowerDNS.

### Expected behaviour
I expect that the ALIAS record is replaced with the A and AAAA records of `origin.example.com.` in an AXRF, if the PowerDNS server is a hidden master.

### Actual behaviour
The ALIAS record is replaced with the A and AAAA records from the publicly deployed, signed version of the zone.

### Other information

I searched the source code and found a few locations where the problem could have it's roots.

In `pdns/tcpreceiver.cc`:

    750: int ret1 = stubDoResolve(getRR<ALIASRecordContent>(zrr.dr)->d_content, QType::A, ips);
    751: int ret2 = stubDoResolve(getRR<ALIASRecordContent>(zrr.dr)->d_content, QType::AAAA, ips);

the stub resolver is used to query for A and AAAA records of the target domain. Looking into `pdns/stubresolver.cc`:

    54: s_stubresolvers.push_back(ComboAddress(::arg()["recursor"], 53));

the recursor option is pushed to the end of the list of IP addresses which are used for querying. If there are `nameserver` entries in `/etc/resolv.conf` and these nameservers are working, the actual recursor is never queried. And maybe the custom port (5678) is rewritten to `53`. Which could also cause the problem.

Also there is no evidence of a DNSproxy which is also set up when the `recursor` option is set in `pdns.conf`, in `pdns/stubresolver.cc`.

pieterlexis:
Use resolv.conf otherwise. Also, do not use 127.0.0.1:53 as fallback,
as this could be ourselves.

Closes #4655

### Checklist

I have:
- [x] documented the code

### DIFF
--- a/docs/markdown/authoritative/howtos.md
+++ b/docs/markdown/authoritative/howtos.md
@@ -185,6 +185,8 @@ resolver=[::1]:5300
 expand-alias=yes
 ```
 
+**note**: If `resolver` is unset, ALIAS expension is disabled!
+
 and add the ALIAS record to your zone apex. e.g.:
 
 ```
--- a/docs/markdown/authoritative/settings.md
+++ b/docs/markdown/authoritative/settings.md
@@ -718,10 +718,11 @@ If set, recursive queries will be handed to the recursor specified here. See
 ["Recursion"](recursion.md).
 
 ## `resolver`
-* IP Address
+* IP Addresses with optional port, separated by commas
 * Added in: 4.1.0
 
-Use this resolver for ALIAS and the internal stub resolver.
+Use these resolver addresses for ALIAS and the internal stub resolver.
+If this is not set, `/etc/resolv.conf` is parsed for upstream resolvers.
 
 ## `retrieval-threads`
 * Integer
--- a/pdns/stubresolver.cc
+++ b/pdns/stubresolver.cc
@@ -14,71 +14,88 @@
 #include "statbag.hh"
 #include "stubresolver.hh"
 
-// s_stubresolvers contains the ComboAddresses that are used by
+// s_resolversForStub contains the ComboAddresses that are used by
 // stubDoResolve
-static vector<ComboAddress> s_stubresolvers;
+static vector<ComboAddress> s_resolversForStub;
 
-/** Parse /etc/resolv.conf and add the nameservers to the vector
- * s_stubresolvers.
+/*
+ * Returns false if no resolvers are configured, while emitting a warning about this
+ */
+bool resolversDefined()
+{
+  if (s_resolversForStub.empty()) {
+    L<<Logger::Warning<<"No upstream resolvers configured, stub resolving (including secpoll and ALIAS) impossible."<<endl;
+    return false;
+  }
+  return true;
+}
+
+/*
+ * Fill the s_resolversForStub vector with addresses for the upstream resolvers.
+ * First, parse the `resolver` configuration option for IP addresses to use.
+ * If that doesn't work, parse /etc/resolv.conf and add those nameservers to
+ * s_resolversForStub.
  */
 void stubParseResolveConf()
 {
-  ifstream ifs("/etc/resolv.conf");
-  if(!ifs)
-    return;
+  if(::arg().mustDo("resolver")) {
+    vector<string> parts;
+    stringtok(parts, ::arg()["resolver"], " ,\t");
+    for (const auto& addr : parts)
+      s_resolversForStub.push_back(ComboAddress(addr, 53));
+  }
 
-  string line;
-  while(std::getline(ifs, line)) {
-    boost::trim_right_if(line, is_any_of(" \r\n\x1a"));
-    boost::trim_left(line); // leading spaces, let's be nice
+  if (s_resolversForStub.empty()) {
+    ifstream ifs("/etc/resolv.conf");
+    if(!ifs)
+      return;
 
-    string::size_type tpos = line.find_first_of(";#");
-    if(tpos != string::npos)
-      line.resize(tpos);
+    string line;
+    while(std::getline(ifs, line)) {
+      boost::trim_right_if(line, is_any_of(" \r\n\x1a"));
+      boost::trim_left(line); // leading spaces, let's be nice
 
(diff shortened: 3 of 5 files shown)


docs/markdown/authoritative/settings.md
```diff
@@ -698,10 +698,11 @@ If set, recursive queries will be handed to the recursor specified here. See
 ["Recursion"](recursion.md).
 
 ## `resolver`
-* IP Address
+* IP Addresses, seperated by commas
```
Habbie: separated. Also they are not just IPs, they take optional ports.

pdns/stubresolver.cc
```diff
@@ -18,59 +18,76 @@
...
 static vector<ComboAddress> s_stubresolvers;
 
-/** Parse /etc/resolv.conf and add the nameservers to the vector
+/*
+ * Returns true if no stubresolvers are configured, while emitting a warning about this
+ */
+bool noStubsDefined()
+{
+  if (s_stubresolvers.empty()) {
+    L<<Logger::Warning<<"No recursors set, stub resolving (including secpoll and ALIAS) impossible."<<endl;
```
Habbie: s/recursors set/resolvers configured/

pdns/stubresolver.cc
```diff
@@ -18,59 +18,76 @@
 // stubDoResolve
 static vector<ComboAddress> s_stubresolvers;
 
-/** Parse /etc/resolv.conf and add the nameservers to the vector
+/*
+ * Returns true if no stubresolvers are configured, while emitting a warning about this
+ */
+bool noStubsDefined()
```
Habbie: how about `bool resolversDefined`? The things that are defined are not stubs, they are resolvers, and non-reverse logic is less confusing.

pdns/stubresolver.cc
```diff
@@ -18,59 +18,76 @@
...
+    return true;
+  }
+  return false;
+}
+
+/*
+ * Fill the s_stubresolvers vector with addresses for the upstream resolvers.
+ * First, parse the `resolver` configuration option for IP addresses to use.
+ * If that doesn't work, parse /etc/resolv.conf and add those nameservers 
  * s_stubresolvers.
```
Habbie: +to

Habbie: also we should not be calling them *stub* resolvers, again.

pdns/stubresolver.cc
```diff
@@ -18,59 +18,76 @@
...
       }
     }
   }
-
-  if(::arg().mustDo("resolver"))
-    s_stubresolvers.push_back(ComboAddress(::arg()["resolver"], 53));
-
-  // Last resort, add 127.0.0.1
-  if(s_stubresolvers.empty()) {
-    s_stubresolvers.push_back(ComboAddress("127.0.0.1", 53));
```
Habbie: agree on removing this

docs/markdown/authoritative/howtos.md
```diff
@@ -185,6 +185,8 @@ setting to an existing resolver:
 resolver=[::1]:5300
 ```
 
+**note**: If unset, `/etc/resolv.conf` is parsed to find upstream resolvers nameservers.
```
Habbie: -nameservers

pdns/stubresolver.cc
```diff
@@ -14,71 +14,88 @@
...
 
-/** Parse /etc/resolv.conf and add the nameservers to the vector
- * s_stubresolvers.
+/*
+ * Returns false if no resolvers are configured, while emitting a warning about this
+ */
+bool resolversDefined()
+{
+  if (s_resolversForStub.empty()) {
+    L<<Logger::Warning<<"No recursors configured, stub resolving (including secpoll and ALIAS) impossible."<<endl;
```
Habbie: s/recursors/upstream resolvers/

PR #5113, opened 2017-03-06
dnsdist: Add support for setting the server selection policy on a per pool basis

RobinGeuze:
### Short description
Add support to dnsdist for setting the server selection policy on a per pool basis, allowing for using different strategies for different pools.

### Checklist

I have:
- [x] documented the code


pdns/dnsdist-lua2.cc
```diff
@@ -1291,4 +1295,27 @@ void moreLua(bool client)
         }
 #endif /* HAVE_NET_SNMP */
       });
+
+    g_lua.writeFunction("setPoolServerPolicy", [](ServerPolicy policy, string pool) {
+        setLuaSideEffect();
+        auto localPools = g_pools.getCopy();
+        setPoolPolicy(localPools, pool, std::make_shared<ServerPolicy>(policy));
+      });
```
rgacogne: The issue here is that you are updating a local copy of the pools but not updating the global state. It won't work if the pool doesn't already exists and `setPoolPolicy()` creates it, because it will only exist in the local copy. You need to do `g_pools.setState(localPools)` afterwards. Same issue in `setPoolServerPolicyLua()`.

pdns/dnsdist.cc
```diff
@@ -692,11 +692,23 @@ std::shared_ptr<ServerPool> createPoolIfNotExists(pools_t& pools, const string&
     if (!poolName.empty())
       vinfolog("Creating pool %s", poolName);
     pool = std::make_shared<ServerPool>();
+    pool->policy = NULL;
```
rgacogne: Just make sure it's default initialized to `nullptr` by replacing:
`std::shared_ptr<ServerPolicy> policy;`
by
`std::shared_ptr<ServerPolicy> policy{nullptr};`

PR #5122, opened 2017-03-07
[WIP] Make Lua threaded for sync-like programming against kvquery

### LINKED ISSUE #5099: non-callback interface to kvresp
- Program: Recursor
 - Issue type: Feature request

### Short description
The current kvresp interface in the recursor is callback-based, which makes it very messy to use. On top of that, it is not supported in `gettag`. We should replace this with an interface that has a more 'synchronous' feeling. A first stab at this is in https://github.com/Habbie/pdns/tree/luathreads

Habbie:
### Short description
Fixes #5099 

### Checklist

I have:
- [ ] documented the code

### Open problems/questions
 - [ ] we leak ThreadIDs, I'm pretty sure
 - [ ] reloads are painful, most likely
 - [ ] should we make a single Lua thread per mthread and reuse that until the mthread dies? Would be slightly more efficient and might allow some cheap state keeping over the runtime of a query
 - [ ] kvquery is UDP right now, which can be flaky. Should we support TCP? This means keepalives, pooling, etc. and/or a tagged protocol.
 - [x] apparently questions are not NUL-safe
 - [x] check g_maxThreads (i.e. refactor that check into a separate function that we call early in case `start-mthread-early` is enabled)
 - [ ] response buffer size is limited to 512 bytes
 - [ ] we destroyed the 'table lookup' optimisation at https://github.com/PowerDNS/pdns/pull/5122/commits/348adb89b330d287b57604de9b80a6900f5669be#diff-7fb8a7d9628127b764d5aad70850602aL613 (if that link goes dead, it's where `genhook` now takes `const string &funcname` instead of `luacall_t &func`). It should be possible to reuse those refs in individual Lua threads but I haven't found out how to do that with LuaWrapper yet.
 - [ ] also do this change for TCP queries

### DIFF
--- a/pdns/pdns_recursor.cc
+++ b/pdns/pdns_recursor.cc
@@ -1526,7 +1526,7 @@ struct doProcessUDPQuestionArguments {
         int fd;
 };
 
-bool overCapacity(const ComboAddress& fromaddr)
+static bool overCapacity(const ComboAddress& fromaddr)
 {
   if(MT->numProcesses() > g_maxMThreads) {
     if(!g_quiet)
@@ -1710,7 +1710,7 @@ static void realDoProcessUDPQuestion(const std::string& question, const ComboAdd
   return;
 }
 
-void indirectDoProcessUDPQuestion(void *p)
+static void indirectDoProcessUDPQuestion(void *p)
 {
   doProcessUDPQuestionArguments* args=(doProcessUDPQuestionArguments *)p;
 
@@ -1725,7 +1725,7 @@ void indirectDoProcessUDPQuestion(void *p)
   realDoProcessUDPQuestion(question, fromaddr, destaddr, tv, fd, true);
 }
 
-string *doProcessUDPQuestion(const std::string& question, const ComboAddress& fromaddr, const ComboAddress& destaddr, struct timeval tv, int fd)
+static string *doProcessUDPQuestion(const std::string& question, const ComboAddress& fromaddr, const ComboAddress& destaddr, struct timeval tv, int fd)
 {
 
   if (::arg().mustDo("start-mthread-early")) {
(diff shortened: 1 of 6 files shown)


wojas:
While testing, I found that the kvquery interface is not binary safe. If I send the IP `1.0.1.1`, the outgoing UDP message has a length of 1 byte instead of 4 bytes. I have not checked yet if this is also the case with the callback interface. I was not able to pinpoint any unsafe conversions in the code.

While reading the code, I also found that the response buffer is limited to 512 bytes in `handleGenUDPQueryResponse`, while you can have packets of 16 kB by default on localhost. Can we increase this?

pdns/pdns_recursor.cc
```diff
@@ -1720,6 +1712,14 @@ void indirectDoProcessUDPQuestion(void *p)
 
 string *doProcessUDPQuestion(const std::string& question, const ComboAddress& fromaddr, const ComboAddress& destaddr, struct timeval tv, int fd)
 {
+  if(MT->numProcesses() > g_maxMThreads) {
+    if(!g_quiet)
+      L<<Logger::Notice<<t_id<<" ["<<MT->getTid()<<"/"<<MT->numProcesses()<<"] DROPPED question from "<<fromaddr.toStringWithPort()<<", over capacity"<<endl;
+
+    g_stats.overCapacityDrops++;
+    return 0;
+  }
```
rgacogne: This does however prevent serving a response from the packet cache even if we already have too many mthreads. Perhaps you can only check it here if `start-mthread-early` is set, and skip the check in `realDoProcessUDPQuestion()` if `inmthread` is true?

Habbie: ah, yes. No packetcache on TCP, that's why that code was simpler. I worried about this, thanks. Will indeed have to 'refactor' as I initally wrote down.

Habbie: Commit below should fix this.

PR #5132, opened 2017-03-10
auth: Hash the entire query in the packet cache, split caches

### LINKED ISSUE #4204: Packet cache messes with error reporting to clients
After receiving and parsing a packet in the Authoritative Server, the packetcache is consulted before the EDNS options and versions are checked. This leads e.g. to NOERROR responses to queries with EDNS0 version 1 set (which must be BADVERS).

rgacogne:
### Short description
* The packet cache now behaves a lot like the ones in dnsdist and the recursor, hashing the entire query (except the query id) to prevent matching queries with, for example, different EDNS version.
* Since the packet and query caches have now much less in common, split them into two different classes since it doesn't make sense to share the same storage and maximum number of entries and doing so clarifies the code a lot. This should also reduce contention.
* Add a new `max-packet-cache-entries` setting to control the number of entries in the packet cache.
* Add a new metric, `query-cache-size`, the number of entries in the query cache since `packetcache-size` is only about the packet cache.

Note that contrary to the recursor one's, the new packet cache doesn't skip the content of an EDNS Client Subnet option if present, in case it's used by a backend to produce the response. We could easily change that, or even make it configurable if we care.

Closes #4204.

### Checklist
I have:
- [x] documented the code


JcChimjj:
＃5232

JcChimjj:
＃5132

PR #5141, opened 2017-03-13
stop (de)serializing dq.data

Habbie:
### Short description
Right now we deserialize dq.data into an unordered map, which is inefficient and limiting. This PR makes `dq.data` an actual Lua object of any support type (table, string, number, ...), including returning any such type as the third value from `gettag`.

Open issues:
 - [x] ~I'm pretty sure this currently breaks non-Lua compilation, which I understand we still support.~ fixed when we merge #5146 
 - [x] documentation updates (removing a sentence or two no doubt)

Reviews very welcome.

### Checklist

I have:
- [x] documented the code

### DIFF
--- a/ext/luawrapper/include/LuaContext.hpp
+++ b/ext/luawrapper/include/LuaContext.hpp
@@ -189,6 +189,17 @@ public:
     class LuaFunctionCaller;
 
     /**
+     * Opaque type that identifies a Lua object
+     */
+    struct LuaObject {
+        LuaObject() = default;
+        LuaObject(lua_State* state, int index=-1) {
+            this->objectInRegistry = std::make_shared<LuaContext::ValueInRegistry>(state, index);
+        }
+        std::shared_ptr<LuaContext::ValueInRegistry> objectInRegistry;
+    };
+
+    /**
      * Opaque type that identifies a Lua thread
      */
     struct ThreadID {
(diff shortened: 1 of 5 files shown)


ext/luawrapper/include/LuaContext.hpp
```diff
@@ -2395,6 +2416,19 @@ struct LuaContext::Pusher<std::tuple<TTypes...>> {
...
 
+// opaque Lua references
+template<>
+struct LuaContext::Reader<LuaContext::LuaObject>
+{
+    static auto read(lua_State* state, int index)
+        -> boost::optional<LuaContext::LuaObject>
+    {
+        LuaContext::LuaObject obj;
+        obj.objectInRegistry = std::shared_ptr<LuaContext::ValueInRegistry>(new LuaContext::ValueInRegistry(state, index));
```
rgacogne: Just a nit, `std::make_shared<LuaContext::ValueInRegistry>(state, index)` would be a bit better (one allocation instead of two).

Habbie: pushed, thanks!

ext/luawrapper/include/LuaContext.hpp
```diff
@@ -189,6 +189,15 @@ class LuaContext {
     class LuaFunctionCaller;
 
     /**
+     * Opaque type that identifies a Lua object
+     */
+    struct LuaObject {
+    public:
+        friend LuaObject;
```
Habbie: Removed now :)

Habbie:
Ready for review!

PR #5159, opened 2017-03-15
dnsdist: Add expungeBySuffix function

### LINKED ISSUE #5109: [dnsdist] Feature Request:  Wildcard Support for DNSDistPacketCache::expungeByName
- Program: dnsdist 
 - Issue type: Feature request

### Short description
Similar to 'pdns_control purge <ZONE>' there should be away to purge the cache for all labels under a given DNSName

### Usecase
The expungeByName() function purges the cache only for the given DNSName:
`getPool(""):getCache():expungeByName(newDNSName("example.com"))`

### Description
It would be nice to have a function or parameter to purge all labels under the given DNSName as well, something like:
`getPool(""):getCache():expungeByName(newDNSName("%.example.com"))`

RobinGeuze:
### Short description
Adds an expungeBySuffix function that allows removing all names matching a suffix from cache. Closes #5109

### Checklist

I have:
- [x] documented the code

### DIFF
--- a/pdns/dnsdist-cache.cc
+++ b/pdns/dnsdist-cache.cc
@@ -258,18 +258,15 @@ void DNSDistPacketCache::expunge(size_t upTo)
   d_map.erase(beginIt, endIt);
 }
 
-void DNSDistPacketCache::expungeByName(const DNSName& name, uint16_t qtype)
+void DNSDistPacketCache::expungeByName(const DNSName& name, uint16_t qtype, bool suffixMatch)
 {
   WriteLock w(&d_lock);
 
   for(auto it = d_map.begin(); it != d_map.end(); ) {
     const CacheValue& value = it->second;
-    uint16_t cqtype = 0;
-    uint16_t cqclass = 0;
-    DNSName cqname(value.value.c_str(), value.len, sizeof(dnsheader), false, &cqtype, &cqclass, nullptr);
 
-    if (cqname == name && (qtype == QType::ANY || qtype == cqtype)) {
-        it = d_map.erase(it);
+    if ((value.qname == name || (suffixMatch && value.qname.isPartOf(name))) && (qtype == QType::ANY || qtype == value.qtype)) {
+      it = d_map.erase(it);
     } else {
       ++it;
     }
--- a/regression-tests.dnsdist/test_Caching.py
+++ b/regression-tests.dnsdist/test_Caching.py
@@ -819,6 +819,168 @@ class TestCacheManagement(DNSDistTest):
             total += self._responsesCounter[key]
         self.assertEquals(total, misses)
 
+    def testCacheExpungeByNameAndSuffix(self):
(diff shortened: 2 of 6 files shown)


pdns/dnsdist-cache.cc
```diff
@@ -276,6 +276,24 @@ void DNSDistPacketCache::expungeByName(const DNSName& name, uint16_t qtype)
...
 
+void DNSDistPacketCache::expungeBySuffix(const DNSName& name, uint16_t qtype)
+{
+    WriteLock w(&d_lock);
+
+    for (auto it = d_map.begin(); it != d_map.end(); ) {
+        const CacheValue& value = it->second;
+        uint16_t cqtype = 0;
+        uint16_t cqclass = 0;
+        DNSName cqname(value.value.c_str(), value.len, sizeof(dnsheader), false, &cqtype, &cqclass, nullptr);
```
rgacogne: The `qname`, `qtype` and `qclass` are stored in the cache entry, we don't need to parse them again. I just noticed `DNSDistPacketCache::expungeByName()` does the parsing too, which is awful, we should fix that!
Thinking about it a bit more, we should probably just add a boolean parameter to `DNSDistPacketCache::expungeByName()` and do an equal comparison or an `isPartOf()` one depending on the boolean value.

regression-tests.dnsdist/test_Caching.py
```diff
@@ -819,6 +819,168 @@ def testCacheExpungeByNameAndType(self):
...
         self.assertEquals(total, misses)
 
+    def testCacheExpungeBySuffix(self):
+        """
+        Cache: Expunge by name
+
+        """
+        misses = 0
+        ttl = 600
+        name = 'expungebyname.cache.tests.powerdns.com.'
```
rgacogne: You are reusing the exact same qname than a previous test in the same suite (`expungebyname.cache.tests.powerdns.com`) so depending on the execution order it might already be cached and the expected miss below will fail. I'd advise using a different qname for each test to prevent this kind of issues.

regression-tests.dnsdist/test_Caching.py
```diff
@@ -819,6 +819,168 @@ def testCacheExpungeByNameAndType(self):
...
+        self.assertEquals(total, misses)
+
+    def testCacheExpungeBySuffixAndType(self):
+        """
+        Cache: Expunge by name and type
+
+        """
+        misses = 0
+        ttl = 600
+        name = 'expungebynameandtype.cache.tests.powerdns.com.'
```
rgacogne: Same issue here, this name is used by another test of the same suite.

RobinGeuze:
I think the last commit should fix all your points.

rgacogne:
Looks good, thank you! Would you mind squashing a bit before I merge?

RobinGeuze:
All squashed :)

PR #5160, opened 2017-03-15
auth: dlso backend

baloo:
### Short description
Powerdns will dlopen a lib to act as a backend.

Missing:
 - [x] Api versionning. I intend to add an api version at the beginning of the `struct lib_so_api`.
 - [x] Packing. A bunch of the structs needs packing.
 - [x] Malloc responsability. `dlso_register_t` should be the one doing the malloc of `struct lib_so_api`, not `dlso` its struct.
 - [ ] Debugability: adds some logging to debug a lib. 

### Checklist

I have:
- [ ] documented the code

### DIFF
--- a/modules/dlsobackend/dlsobackend.cc
+++ b/modules/dlsobackend/dlsobackend.cc
@@ -1,0 +1,640 @@
+/*
+ * This file is part of PowerDNS or dnsdist.
+ * Copyright -- PowerDNS.COM B.V. and its contributors
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of version 2 of the GNU General Public License as
+ * published by the Free Software Foundation.
+ *
+ * In addition, for the avoidance of any doubt, permission is granted to
+ * link this program with OpenSSL and to (re)distribute the binaries
+ * produced as the result of such linking.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+#ifdef HAVE_CONFIG_H
+#include "config.h"
+#endif
+#include "dlsobackend.hh"
+#include <dlfcn.h>
+
+static const char *kBackendId = "[DlsoBackend]";
+
+/**
+ * Standard ctor and dtor
+ */
+DlsoBackend::DlsoBackend(const std::string &suffix)
+{
+  setArgPrefix("dlso"+suffix);
+
+  std::string libpath = getArg("path");
+  std::string args = getArg("args");
+
+  this->d_dnssec = mustDo("dnssec");
+  this->in_query = false;
+
+  this->dlhandle = dlopen(libpath.c_str(), RTLD_LAZY);
+
+  if (!this->dlhandle) {
+      throw PDNSException("Unable to load library: " + libpath + ":\n" + dlerror());
+  }
+
+  dlso_register_t register_api = (dlso_register_t) dlsym(this->dlhandle, "pdns_dlso_register");
+  if (register_api == NULL) {
+    dlclose(this->dlhandle);
+    throw PDNSException("Failed to initialize dlso, no pdns_dlso_register symbol exposed");
+  }
+
+  bool success = register_api(PDNS_DLSO_ABI_VERSION, &this->api, this->d_dnssec, args.c_str());
+  if (!success) {
+    dlclose(this->dlhandle);
+    throw PDNSException("Failed to initialize dlso, pdns_dlso_register returned false");
+  }
+
+  // Check pointer safety
+  if (this->api == NULL) {
+    dlclose(this->dlhandle);
+    throw PDNSException("Failed to initialize dlso, library did not reply with its pointer");
+  }
+
+  // Check version
+  if (this->api->abi_version != PDNS_DLSO_ABI_VERSION) {
+    if (this->api->release != NULL)
+      this->api->release(this->api);
+
+    dlclose(this->dlhandle);
+    throw PDNSException("Failed to initialize dlso, library does not use the same abi version than pdns");
+  }
+
+  // Sanity checks
+  if (this->api->get == NULL) {
+    if (this->api->release != NULL)
+      this->api->release(this->api);
+
+    dlclose(this->dlhandle);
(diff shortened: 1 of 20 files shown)


modules/dlsobackend/dlsobackend.cc
```diff
@@ -0,0 +1,297 @@
...
+  this->api.get_metas = (bool (*)(void * handle, uint8_t qlen, const char * qname, fill_metas_cb_t cb, void * metas)) dlsym(this->dlhandle, "pdns_dlso_domain_all_metadatas");
+  this->api.get_meta = (bool (*)(void * handle, uint8_t qlen, const char * qname, uint8_t meta_len, const char * kind, fill_meta_cb_t cb, void * meta)) dlsym(this->dlhandle, "pdns_dlso_domain_metadata");
+
+  this->api.get_before_after = (bool (*)(void * handle, uint32_t domain_id, uint8_t qlen, const char * qname, fill_before_after_t cb, void * beforeAfter)) dlsym(this->dlhandle, "pdns_dlso_get_before_after");
+  this->api.get_tsig_key = (bool (*)(void * handle, uint8_t qlen, const char * qname, uint8_t alg_len, const char * alg, fill_tsig_key_cb_t cb, void * content)) dlsym(this->dlhandle, "pdns_dlso_get_tsig_key");
+
+  this->handle = this->api.new_(this->d_dnssec, args.c_str());
+
+  if (this->handle == NULL)
+    throw PDNSException("Failed to initialize dlso");
```
rgacogne: It looks like a `dlclose()` is missing here.

modules/dlsobackend/dlsobackend.cc
```diff
@@ -0,0 +1,297 @@
...
+  if (!success) {
+    in_query = false;
+    return false;
+  }
+
+  return true;
+}
+
+void fill_meta_cb(std::vector<std::string>* meta, uint8_t value_len, const struct dns_meta_value * values) {
+  for (int i=0; i<value_len; i++) {
```
rgacogne: The loop index should be unsigned, otherwise the compiler will complain about comparing signed and unsigned values.

modules/dlsobackend/dlsobackend.cc
```diff
@@ -0,0 +1,297 @@
...
+    const struct dns_meta_value *value = &values[i];
+    string value_s = string(value->value, value->value_len);
+
+    (*meta).push_back(value_s);
+  }
+}
+
+void fill_metas_cb(std::map<std::string, std::vector<std::string>>* metas, uint8_t meta_len, const struct dns_meta * c_metas) {
+  for (int i=0; i<meta_len; i++) {
```
rgacogne: Same remark about the signedness.

zeha:
Possibly stupid question: what is the usecase for this backend?

zeha:
Thought: this would need some form of versioning, so API changes would not lead to difficult to debug crashes.

baloo:
The purpose is to have a backend that supports more databases, than the one provided by gsql (mysql, pg, sqlite, ...). The remote backend is interesting but involves a lot of parsing and queuing which hurts performance.
A `dlopen` would allow pdns to include custom code in any language that can export dynamic symbols, and open any backend database.
That allows to include custom code in various language (c, cpp, go, rust, ... you name it) using various databases (tokyo cabinet, rocksdb, leveldb, bdb, ... you name it).

This backend involves **a lot** of string copy, which I'm not fund of. I could probably do better, any advise taken.

Yeah, api should be versioned. The easiest way to go is probably to make the new method return a struct with version and pointers to various methods (`lookup`, `get`, `list`, ...).

modules/dlsobackend/dlsobackend.cc
```diff
@@ -0,0 +1,304 @@
...
+  this->in_query = false;
+
+  this->dlhandle = dlopen(libpath.c_str(), RTLD_LAZY);
+
+  if (!this->dlhandle) {
+      throw PDNSException("Unable to load library: " + libpath);
+  }
+
+  this->api.new_ = (void * (*)(bool, const char *)) dlsym(this->dlhandle, "pdns_dlso_new");
+  this->api.free_ = (void (*)(void *)) dlsym(this->dlhandle, "pdns_dlso_free");
```
bapt: No check against NULL is done, which means if one makes a plugin that does not define pdns_dlso_free will result in a segmentation fault

modules/dlsobackend/dlsobackend.cc
```diff
@@ -0,0 +1,304 @@
...
+  this->api.list = (bool (*)(void *, uint8_t, const char*)) dlsym(this->dlhandle, "pdns_dlso_list");
+  this->api.lookup = (bool (*)(void *, uint16_t, uint8_t, const char*, const sockaddr*)) dlsym(this->dlhandle, "pdns_dlso_lookup");
+  this->api.get = (bool (*)(void *, fill_cb_t, void *)) dlsym(this->dlhandle, "pdns_dlso_get");
```
bapt: same

zeha:
All those individual dlsym() calls look like a bad idea to me. dlsym'ing a `dlso_register` function that returns a function pointer table would be somewhat nicer... could also do the versioning thing in there.

baloo:
I agree with you, I will rewrite this.

baloo:
There is now a `pdns_dlso_register` method, this is expected to fill in a bunch of methods in the struct. Only `get` and `lookup` are required by dlso_backend. Others are optional.

Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+  flex \
+  ragel \
+  sqlite3 \
+  bc \
+  dnsutils
+
+RUN cd /tmp; \
+    wget http://ppa.launchpad.net/kalon33/gamesgiroll/ubuntu/pool/main/libs/libsodium/libsodium-dev_1.0.3-1~ppa14.04+1_amd64.deb \
+         http://ppa.launchpad.net/kalon33/gamesgiroll/ubuntu/pool/main/libs/libsodium/libsodium13_1.0.3-1~ppa14.04+1_amd64.deb; \
+    dpkg -i libsodium-dev_1.0.3-1~ppa14.04+1_amd64.deb libsodium13_1.0.3-1~ppa14.04+1_amd64.deb
```
baloo: This is an error, docker shouldn't appear here. This was an attempt to reproduce travis-ci environment.

baloo:
Needs a debug log in call

Habbie:
Can you rebase this and make travis pass?

baloo:
I won't have time for this until June 11. Please keep it open until then. Thank you

baloo:
I tried to give rebase a shot, but I'm off for holidays and I most likely won't have time to touch this PR before June 11th. So I most likely be inactive until then.

aerique:
@baloo Are you still off for holidays? :-)

baloo:
Nah, just fell off my schedule. Thanks for the reminder

pieterlexis:
Closing this for now. Feel free to re-open when there is movement :).

PR #5163, opened 2017-03-16
dnsdist: Cleanup closed TCP downstream connections

rgacogne:
### Short description
Regularly walk the TCP downstream connections to properly close whose that have been shutdown by the other end. It occurs only after a TCP client connection has been closed and at most every `setTCPDownstreamCleanupInterval()` seconds, defaulting to 60s.
Until now we only detected that the other end closed the connection when we tried to reuse it. While this is not an issue with a small number of backends because the connection are reused pretty quickly,
with a large number of backends dnsdist might end up with thousands of idle TCP connections to downstream servers in `CLOSE_WAIT` state, wasting open file descriptors.

### Checklist

I have:
- [ ] documented the code


Habbie:
Silly question: can't we put all these FDs in an (e)poll set we keep up to date, so the check does not involve iterating over all the sockets?

rgacogne:
Sure, the only downside is that managing an up-to-date set is going to be a pain. I'm not sure it's worth it, but I haven't looked at our existing FD multiplexer implementations so maybe we already have the code to manage that easily.

PR #5182, opened 2017-03-20
ALIAS: disable expansion by default

### LINKED ISSUE #5119: Add an option to disable ALIAS expanding
Currently it is not possible to disable ALIAS expanding. This may cause security issues in setups with untrusted customer zones. See thread https://mailman.powerdns.com/pipermail/pdns-users/2017-February/024670.html:

> On Thu, 9 Feb 2017 22:12:07 +0100
> Klaus Darilion <klaus.mailinglists at pernau.at> wrote:
> 
> > On 08.02.2017 18:53, Pieter Lexis wrote:
> > >> - If ALIAS is not enabled, will PDNS just ignore these records?  
> > > ALIAS is always "enabled". When we encounter an ALIAS record for the name queried, it is expanded.  
> > 
> > So, there is no means to disable ALIAS? Then this is IMO a bug. We use
> > PowerDNS to slave zones from our customers. When now one of these
> > customers put in an ALIAS, the customer can inject DNS queries in our
> > resolvers. E.g. if there is a day zero in a common resolver software -
> > the untrusted customer could trigger that the resolver resolves a
> > malicious domain and exploit the day zero.
> > 
> > This sounds very dangerous to me. Suddenly my resolvers, which were only
> > accessible from within my network, can be used by everybody (at least by
> > all my customers). This is a massive impact should be noted in more
> > details in the changelog, Because up to now I only had to deal with
> > authoritative name server security - but this feature forces me to setup
> > a dedicated resolver for this untrusted resolving-request.
> > 
> > Please add a feature to "disable-alias-expanding" and make it default
> > YES if you care about security.
> 
> Your comment makes sense, not only from the resolver perspective, but also from the "I don't want my slaves doing any dynamic things"-perspective.
> Could you file an issue[1] for this so we can track this?
> 
> Thank you for clarifying,
> Pieter
>

pieterlexis:
### Short description
Closes #5119

### Checklist

I have:
- [ ] documented the code

### DIFF
--- a/docs/markdown/authoritative/howtos.md
+++ b/docs/markdown/authoritative/howtos.md
@@ -178,11 +178,11 @@ If you have multiple IP addresses on the internet on one machine, UNIX often sen
 # Using ALIAS records
 The ALIAS record provides a way to have CNAME-like behaviour on the zone apex.
 
-In order to correctly serve ALIAS records, set the [`resolver`](settings.md#resolver)
-setting to an existing resolver: 
+In order to correctly serve ALIAS records, set the [`resolver`](settings.md#resolver) setting to an existing resolver and enable [`expand-alias`](settings.md#expand-alias):
 
 ```
 resolver=[::1]:5300
+expand-alias=yes
 ```
 
 and add the ALIAS record to your zone apex. e.g.:
@@ -212,6 +212,9 @@ ALIAS-record and send the results in the AXFR.
 Set `outgoing-axfr-expand-alias` to 'yes' if your slaves don't understand ALIAS
 or should not look up the addresses themselves. Note that slaves will not
 automatically follow changes in those A/AAAA records unless you AXFR regularly.
+
+**note:** The `expand-alias` setting does not exist in PowerDNS Authoritative Server 4.0.x.
+Hence, ALIAS records are always expanded on a direct A or AAAA query.
 
 ## ALIAS and DNSSEC
 Starting with the PowerDNS Authoritative Server 4.0.0, DNSSEC 'washing' of ALIAS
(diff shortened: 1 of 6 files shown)


docs/markdown/authoritative/settings.md
```diff
@@ -342,6 +342,17 @@ Enables EDNS subnet processing, for backends that support it.
...
+## `expand-alias`
+* Boolean
+* Default: no
+* Since: 4.1.0
+
+If this is enabled, ALIAS records are expanded (synthesised to their A/AAAA).
+
+If this is disabled (the default), ALIAS records will not expanded and the server will will return NODATA for A/AAAA queries for such names.
+
+**note**: In PowerDNS Authoritative Server 4.0.0, this setting did not exist and ALIAS was always expanded.
```
Habbie: 4.0.x

docs/markdown/authoritative/howtos.md
```diff
@@ -213,6 +213,9 @@ Set `outgoing-axfr-expand-alias` to 'yes' if your slaves don't understand ALIAS
 or should not look up the addresses themselves. Note that slaves will not
 automatically follow changes in those A/AAAA records unless you AXFR regularly.
 
+**note:** The `expand-alias` setting does not exist in PowerDNS Authoritative Server 4.0.X.
```
Habbie: lowercase x please

pdns/packethandler.hh
```diff
@@ -108,6 +108,7 @@ private:
   bool d_logDNSDetails;
   bool d_doIPv6AdditionalProcessing;
   bool d_doDNAME;
+  bool d_doExpandALIAS;
```
Habbie: I wonder if these flags, in general, shouldn't be static so we just parse them once on startup. Not a job for this PR though.

PR #5198, opened 2017-03-23
rec: Allow access to EDNS options from the `gettag()` hook

### LINKED ISSUE #5195: Recursor gettag(): allow access to EDNS options
- Program: Recursor 
 - Issue type: Feature request 

### Short description

The `gettag()` Lua callback currently provides no way to access EDNS options other than the EDNS client subnet.

### Usecase

I need access to EDNS option 65001 within the `gettag()` callback. This callback does not receive a `dq` object, so I have no access to `dq:getDH():getEDNSOption(65001)`, which is available in other callbacks.

### Description

The current `gettag()` interface is:

```
function gettag(remote, ednssubnet, local, qname, qtype)
```

I propose extending it with an extra parameter that is either a `dh` DNS headers object, or with the result of `getEDNSOptions()` if `dh` is too expensive.

Option 1:

```
function gettag(remote, ednssubnet, local, qname, qtype, dh)
    local opt65001 = dh:getEDNSOption(65001)
    ...
end
```

Option 2:

```
function gettag(remote, ednssubnet, local, qname, qtype, edns)
    local opt65001 = edns[65001]
    ...
end
```

Ideally, some kind of `gettag2()` that receives a light version of the `dq` object could be implemented to avoid an ever growing list of parameters to `gettag()`. This would also allow to lazily fetch the actually required data instead of always having to precalculate it and pass it in.

rgacogne:
### Short description

If `gettag-needs-edns-options` is set, the EDNS options are extracted and passed to the `gettag()` hook as a table whose keys are the EDNS option code and the values are `EDNSOptionView` object.
`EDNSOptionView` has two members, `content` and `size`, with `content` holding the raw, undecoded option value.

Closes #5195.

### Checklist

I have:
- [x] documented the code

### DIFF
--- a/pdns/lua-recursor4.cc
+++ b/pdns/lua-recursor4.cc
@@ -384,7 +384,10 @@ RecursorLua4::RecursorLua4(const std::string& fname)
   d_lw->registerMember("type", &DNSRecord::d_type);
   d_lw->registerMember("ttl", &DNSRecord::d_ttl);
   d_lw->registerMember("place", &DNSRecord::d_place);
-  
+
+  d_lw->registerMember("size", &EDNSOptionView::size);
+  d_lw->registerFunction<std::string(EDNSOptionView::*)()>("getContent", [](const EDNSOptionView& option) { return std::string(option.content, option.size); });
+
   d_lw->registerFunction<string(DNSRecord::*)()>("getContent", [](const DNSRecord& dr) { return dr.d_content->getZoneRepresentation(); });
   d_lw->registerFunction<boost::optional<ComboAddress>(DNSRecord::*)()>("getCA", [](const DNSRecord& dr) { 
       boost::optional<ComboAddress> ret;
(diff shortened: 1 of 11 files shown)


wojas:
I tested this change and found that it is not NUL safe.

If I send "\x01\x02\x00\xa4\xa5\xa6", only "\x01\x02" can be accessed within Lua. I verified with Wireshark that I am sending the full 6 bytes.

pdns/lua-recursor4.cc
```diff
@@ -384,7 +384,10 @@ RecursorLua4::RecursorLua4(const std::string& fname)
   d_lw->registerMember("type", &DNSRecord::d_type);
   d_lw->registerMember("ttl", &DNSRecord::d_ttl);
   d_lw->registerMember("place", &DNSRecord::d_place);
-  
+
+  d_lw->registerMember("content", &EDNSOptionView::content);
```
wojas: This is the reason it is not NUL safe. 

If this were a `std::string`, Luawrapper would handle it in a NUL safe way. Here Luawrapper is not aware of the size field and interprets it as a NUL terminated string.

This could also cause issues with content without NULs in it, as it will continue to read until it encounters a NUL.

wojas: One easy way to resolve this would be to not register content with Luawrapper, but instead add a `get()` method to `EDNSOptionView` and register that one.

Then the Lua code could do this:

```lua
function gettag(remote, edns, loc, qname, qtype, ednsoptions)
    if ednsoptions ~= nil then 
        local opt = ednsoptions[65001] 
        if opt ~= nil and opt.length == 6 then
            print("Found option 65001", opt:get())
        end
    end 
    ...
end
```

wojas: The `get()` method would return a `std::string`.

wojas:
I tested it and it works as expected. Thanks!

PR #5199, opened 2017-03-23
setKey() was actually "appendKey()". In addition, '-k' would not override for server configuration

ahupowerdns:
Unified -k behaviour for client and server mode now.

### Short description
Using setKey() at runtime actually did not work if a key had been set already. setKey() was in fact appendKey(). Furthermore, -k did not operate in server mode. In client mode, there was an ugly override to make -k work. Unified this now to too server and client modes.

### Checklist

I have:
- [X] documented the code

### DIFF
--- a/pdns/dnsdist-lua.cc
+++ b/pdns/dnsdist-lua.cc
@@ -1462,11 +1462,16 @@ vector<std::function<void(void)>> setupLua(bool client, const std::string& confi
     });
   
   g_lua.writeFunction("setKey", [](const std::string& key) {
+      if(!g_configurationDone && ! g_key.empty()) { // this makes sure the commandline -k key prevails over dnsdist.conf
+        return;                                     // but later setKeys() trump the -k value again
+      }
+
       setLuaSideEffect();
+      g_key.clear();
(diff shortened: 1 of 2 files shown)


pdns/dnsdist-lua.cc
```diff
@@ -1462,11 +1462,16 @@ vector<std::function<void(void)>> setupLua(bool client, const std::string& confi
...
+
       setLuaSideEffect();
+      g_key.clear();
       if(B64Decode(key, g_key) < 0) {
-	  g_outputBuffer=string("Unable to decode ")+key+" as Base64";
-	  errlog("%s", g_outputBuffer);
-	}
+        g_outputBuffer=string("Unable to decode ")+key+" as Base64";
+        errlog("%s", g_outputBuffer);
+      }
```
rgacogne: Shouldn't we try to decode the new key into a temporary string before clearing the existing one?

ahupowerdns: yes, will make it so. thanks!

Habbie: Ping

PR #5201, opened 2017-03-24
WIP: Add dnstap-compatible protobuf support to dnsdist.

jvalentini:
### Short description
This adds support for producing dnstap protobuf log messages to dnsdist.

### Checklist

I have:
- [ ] documented the code


pdns/dnsdistdist/configure.ac
```diff
@@ -74,7 +74,7 @@ PDNS_ENABLE_SANITIZERS
 
 PDNS_CHECK_PANDOC
 
-LDFLAGS="$RELRO_LDFLAGS $LDFLAGS"
+LDFLAGS="$RELRO_LDFLAGS $LDFLAGS -lfstrm"
```
rgacogne: The configure needs to be detect whether libfstrm is available, and dnsdist should be able to compile correctly if it isn't.

pdns/dnsdistdist/dnsrulactions.cc
```diff
@@ -1,132 +0,0 @@
-/*
```
rgacogne: You are moving dnsrulactions.cc from the dnsdistdist dir to the pdns one, while we try to keep files that are only used by dnsdist in the dnsdistdist directory. I know I haven't been very consistent about that in the past, sorry.

pdns/fstrm_logger.cc
```diff
@@ -0,0 +1,110 @@
...
+#include "logger.hh"
+#define WE_ARE_RECURSOR
+#else
+#include "dolog.hh"
+#endif
+#include <sys/un.h>
+
+#define DNSTAP_CONTENT_TYPE		"protobuf:dnstap.Dnstap"
+
+FrameStreamLogger::FrameStreamLogger(const std::string socket_path): socket_path(socket_path)
```
rgacogne: Just a nit, but socket_path could be a reference.

pdns/fstrm_logger.cc
```diff
@@ -0,0 +1,110 @@
...
+{
+  struct sockaddr_un local;
+  local.sun_family = AF_UNIX;
+  strcpy(local.sun_path, socket_path.c_str());
```
rgacogne: Please check that you do not overflow local.sun_path

pdns/fstrm_logger.cc
```diff
@@ -0,0 +1,110 @@
...
+
+  fwopt = fstrm_writer_options_init();
```
rgacogne: It looks like it can fail and return `nullptr`, please add a check for that case

pdns/fstrm_logger.cc
```diff
@@ -0,0 +1,110 @@
...
+  fstrm_res res = fstrm_writer_options_add_content_type(fwopt, DNSTAP_CONTENT_TYPE, sizeof(DNSTAP_CONTENT_TYPE) - 1);
+  if (res != fstrm_res_success) {
+#ifdef WE_ARE_RECURSOR
+    L<<Logger::Warning<<"Error: fstrm_writer_options_add_content_type failed: "<<res<<std::endl;
+#else
+    warnlog("Error: fstrm_writer_options_add_content_type failed: %s\n", res);
+#endif
+  }
+
+  uwopt = fstrm_unix_writer_options_init();
```
rgacogne: This can probably fail too

pdns/fstrm_logger.cc
```diff
@@ -0,0 +1,110 @@
...
+  writer = fstrm_unix_writer_init(uwopt, fwopt);
+  if (!writer) {
+#ifdef WE_ARE_RECURSOR
+    L<<Logger::Warning<<"Error: fstrm_unix_writer_init() failed."<<std::endl;
+#else
+    warnlog("Error: fstrm_unix_writer_init(%s) failed.\n", socket_path);
+#endif
+  }
+
+  iothr = fstrm_iothr_init(NULL, &writer);
```
rgacogne: Just a nit but using `nullptr` instead of `NULL` might help prevent mistakes

pdns/fstrm_logger.cc
```diff
@@ -0,0 +1,110 @@
...
+  }
+
+  ioqueue = fstrm_iothr_get_input_queue(iothr);
+  if (!ioqueue) {
+#ifdef WE_ARE_RECURSOR
+    L<<Logger::Warning<<"Error: fstrm_iothr_get_input_queue() failed."<<std::endl;
+#else
+    warnlog("Error: fstrm_iothr_get_input_queue() failed.\n");
+#endif
+  }
```
rgacogne: How about throwing an exception if anything fails, instead of just logging an error? because it looks like the `FrameStreamLogger` will be useless anyway if any step fails, right?
Of course you'll need to cleanup any object already created before throwing.

pdns/fstrm_logger.cc
```diff
@@ -0,0 +1,110 @@
...
+  iothr = fstrm_iothr_init(NULL, &writer);
+  if (!iothr) {
+#ifdef WE_ARE_RECURSOR
+    L<<Logger::Warning<<"Error: fstrm_iothr_init() failed."<<std::endl;
+#else
+    warnlog("Error: fstrm_iothr_init() failed.\n");
+#endif
+  }
+
+  ioqueue = fstrm_iothr_get_input_queue(iothr);
```
rgacogne: Since `queueData()` can be called from several threads, the queue model should be set to `FSTRM_IOTHR_QUEUE_MODEL_MPSC` with `fstrm_iothr_options_set_queue_model()`

pdns/fstrm_logger.cc
```diff
@@ -0,0 +1,110 @@
...
+    fstrm_res res;
+    res = fstrm_iothr_submit(iothr, ioqueue, frame,
+                             data.length(),
+                             fstrm_free_wrapper, NULL);
+
+    if (res == fstrm_res_success) {
+      // Frame successfully queued.
+      break;
+    } else if (res == fstrm_res_again) {
+      // Queue is full.
```
rgacogne: `queueData()` is busy-looping for as long as the queue is full, effectively blocking queries/responses processing which is not good. I think we should just give up on `fstrm_res_again` (don't forget to free the allocated memory then)

pdns/fstrm_logger.hh
```diff
@@ -0,0 +1,53 @@
...
+#pragma once
+#include "config.h"
+
+#include <atomic>
+#include <condition_variable>
+#include <queue>
+#include <thread>
+
+#include "iputils.hh"
+
```
rgacogne: The includes from `<atomic>` to `"iputils.hh"` seem unnecessary

pdns/protobuf-dnstap.cc
```diff
@@ -0,0 +1,248 @@
...
+#include "dnsparser.hh"
+#include "config.h"
+#ifdef PDNS_CONFIG_ARGS
+#include "logger.hh"
+#define WE_ARE_RECURSOR
+#else
+#include "dolog.hh"
+#endif
+
+#define MAXHOSTNAMELEN 256
```
rgacogne: it would be better not to redefine an existing standard macro

pdns/protobuf-dnstap.cc
```diff
@@ -0,0 +1,248 @@
...
+
+  dnstap::Message* message = proto_message.mutable_message();
+  message->set_response_message(packet, len);
+#endif /* HAVE_PROTOBUF */
+}
+
+void DnstapProtoBufMessage::serialize(std::string& data) const
+{
+#ifdef HAVE_PROTOBUF
+  proto_message.SerializeToString(&data);
```
rgacogne: You are duplicating of lot of code between `DnstapProtoBufMessage` and `DefaultProtoBufMessage`, how about creating a base class instead of an empty interface?

pdns/protobuf-dnstap.cc
```diff
@@ -0,0 +1,248 @@
...
+
+void DnstapProtoBufMessage::update(const boost::uuids::uuid& uuid, const ComboAddress* requestor, const ComboAddress* responder, bool isTCP, uint16_t id)
+{
+  struct timespec ts;
+  gettime(&ts, true);
+  setTime(ts.tv_sec, ts.tv_nsec / 1000);
+  setUUID(uuid);
+
+  char identity[MAXHOSTNAMELEN+1];
+  if (gethostname(identity, MAXHOSTNAMELEN) == 0) {
```
rgacogne: Please don't call `gethostname()` for every update() call. Either pass the host name to the constructor or compute it only once there and store it. You might even want to allow setting the identity as a parameter to `newFrameStreamLogger()`.

pdns/protobuf-dnstap.cc
```diff
@@ -0,0 +1,248 @@
...
+  dnstap::Message* message = proto_message.mutable_message();
+  message->set_query_message(packet, len);
+#endif /* HAVE_PROTOBUF */
+}
+
+void DnstapProtoBufMessage::setQuestion(const DNSName& qname, uint16_t qtype, uint16_t qclass)
+{
+#ifdef HAVE_PROTOBUF
+  std::string msg = "qname: " + qname.toString() + "; qtype: " +  std::to_string(qtype) + "; qclass: " + std::to_string(qclass);
+  proto_message.set_extra(msg);
```
gibson042: It seems highly opinionated to put data in `extra`; shouldn't this be a no-op?

pieterlexis:
@jvalentini ping?

jvalentini:
Sorry for the delay. I should be able to start looking at this now. Thanks for the feedback so far!

johnhtodd:
My organization has added tagging to protobuf messages so that downstream telemetry systems can get some ideas about what happened during the internal Lua processing cycle. (see: https://github.com/PowerDNS/pdns/pull/5396)  I'm happy to see dnstap abilities, and we'd probably have no problem moving to them but we would want to see the same type of tagging visible in the dnstap protobuf streams as well. Is this worth doing on the first pass or should we just wait to have this (and our patch) added, and then extend tagging afterwards to include dnstap messages?

pieterlexis:
*ping*

rgacogne:
Hello @jvalentini! We are seeing quite some interest for this feature, so I was wondering if you expected to be able to work on it, or if we should perhaps consider picking up from there? Thanks!

jvalentini:
@rgacogne If there's significant interest it's probably best to have someone take this over as it's been hard to find time. It's also somewhat complicated by the fact that this is my first c++ project and a lot of the feedback are things I'm not familiar with, which makes any progress slow going. Thanks!

rgacogne:
This PR has been superseded by #6170, which has been merged. A huge thanks again for this awesome work!

PR #5209, opened 2017-03-28
make our Lock classes uncopyable, make ReadLock and WriteLock moveable.

ahupowerdns:
### Short description
Right now it is possible to copy our Lock classes, after which bad things will happen. Both destructors will attempt to unlock the mutex. 

With this PR 1) copying becomes impossible 2) the ReadLock and WriteLock classes gain move support, which enables storing them in containers with expected semantics. This is good for having "thousands of locks" that exhibit RAII behaviour.
### Checklist

I have:
- [X] documented the code

### DIFF
--- a/pdns/Makefile.am
+++ b/pdns/Makefile.am
@@ -1156,6 +1156,7 @@ testrunner_SOURCES = \
 	test-dnsparser_hh.cc \
 	test-dnsrecords_cc.cc \
 	test-iputils_hh.cc \
+	test-lock_hh.cc \
 	test-md5_hh.cc \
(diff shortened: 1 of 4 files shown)


rgacogne:
Alternatively, perhaps `ReadLock::upgrade()` should just go since it doesn't appear to be used and is quite misleading, because it does release the existing read lock before acquiring a write one.

ahupowerdns:
I have added some tests and removed the upgrade() feature which was stupid. gotIt() will give you a nonsense answer on moved locks, but I'm told you should not be doing that.

PR #5212, opened 2017-03-29
pdns: remove duplicated dns2_tolower() function

tfarina:
There is already a version of dns2_tolower() in misc.hh, called
simply dns_tolower(), but due to some inclusion issues it was necessary
to extract it into another header file (named ascii.hh).


Habbie:
Travis is unhappy with your patch. Can you check the logs and find out why?

tfarina:
I know why it is unhappy, but I don't know autoconf well enough to change the Makefiles to add ascii.hh to the include path so the compiler could find it.

Habbie:
You just need symlinks in the recursordist and dnsdistdist dirs.

tfarina:
Done. Symlinks added.

Habbie:
Travis still unhappy (I did not check why).

Habbie:
Also, if you do requested fixes, please push additional commits, this makes review easier.

tfarina:
dnsname.hh:37:20: fatal error: ascii.hh: No such file or directory

rgacogne:
So, you need to add `ascii.hh` to the `dnsdist_SOURCES` variable in `pdns/dnsdistdist/Makefile.am` and the `pdns_recursor_SOURCES` one in `pdns/recursordist/Makefile.am`, otherwise they are not included in the dnsdist and recursor tarballs.

tfarina:
Done. It passed the trybots. Thanks Remi.

pdns/Makefile.am
```diff
@@ -133,6 +133,7 @@ EXTRA_PROGRAMS = \
 
 pdns_server_SOURCES = \
 	arguments.cc arguments.hh \
+        ascii.hh \
```
pieterlexis: Please replace with a tab to match style

tfarina:
Done.

pieterlexis:
hi @tfarina,

Pleas ealso do this to the other `Makefile.am`'s from this PR. They have the same issue 😉

tfarina:
Done.

Issue #5054, opened 2017-02-20
dnsdist query counters suffer from integer overflows


mikedamm:
![image](https://cloud.githubusercontent.com/assets/24718/23110178/1b8e2c94-f6d4-11e6-8fa3-3c479689b779.png)

dnsdist built and running on a 64-bit OS. Query counter for pools and match counter for rules overflow.

rgacogne:
Hi Mike,

Could you please tell us which version you are running? We have fixed similar issues in the past so if you are not running 1.1.0 already, upgrading might help. Would you mind also posting the result of running `showRules()` and `showServers()` via the console, so we know whether the issue is limited to the web interface? Thanks!

Habbie:
Ping

tmuncks:
Same problem in the web-interface here. The counters from showRules() and showServers() are fine though (for my current satte, I'm only seeing the rollover in rules)

> showRules()
#     Matches Rule                                               Action
1   2631607333 (Dst: x.x.x.x/32, y.y.y.y/32) && (Src: x.x.x.x/19, y.y.y.y/19, z.z.z.z/22) to pool secure
2     5727075 All                                                to pool public
![screenshot from 2017-03-22 08-08-05](https://cloud.githubusercontent.com/assets/9822186/24195502/beca6776-0ed8-11e7-847d-6efec2d08a36.png)

rgacogne:
Fixed by #5234.

PR #5222, opened 2017-04-04
Fix matches overflow in dnsdist API.

aerique:
The way it is currently fixed means that matches can still overflow
in the web interface because JavaScript has a 2^53 number limit.

However, this fix is in line with how earlier cases have been fixed.

This fixes issue #5054.

### Checklist

I have:
- [ ] documented the code

### DIFF
--- a/pdns/dnsdist-web.cc
+++ b/pdns/dnsdist-web.cc
@@ -380,7 +380,7 @@ static void connectionThread(int sock, ComboAddress remote, string password, str
       for(const auto& a : localRules) {
 	Json::object rule{
 	  {"id", num++},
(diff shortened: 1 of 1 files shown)


rgacogne:
Hi @aerique,
Thank you for this PR! The initial report also mentioned the backend queries counter, and if I'm not mistaken this only addresses the issue for the query and response rules counters?

aerique:
Yes, sorry, that is correct. I looked through the code with Pieter and we (well: he) determined that the matches and one of the query counters needed to be fixed.

Now that I'm almost done with my new machine I'll go and wrap this issue up.  Will probably be Monday though.

PR #5226, opened 2017-04-07
rec: Refactor the negative cache into a class

pieterlexis:
### Short description
This PR refactors the negative cache of the recursor into a class and uses that class in the SyncRes.

Needs a sturdy review.

It also has some cosmetic improvements:

* Adds `pragma once` to cachecleaner.hh
* Guards the `config.h` include in remote_logger.hh
* git-ignores files generated by the testrunner

### Checklist

I have:
- [x] documented the code

### DIFF
--- a/pdns/recursordist/negcache.cc
+++ b/pdns/recursordist/negcache.cc
@@ -1,0 +1,155 @@
+/*
+ * This file is part of PowerDNS or dnsdist.
+ * Copyright -- PowerDNS.COM B.V. and its contributors
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of version 2 of the GNU General Public License as
+ * published by the Free Software Foundation.
+ *
+ * In addition, for the avoidance of any doubt, permission is granted to
+ * link this program with OpenSSL and to (re)distribute the binaries
+ * produced as the result of such linking.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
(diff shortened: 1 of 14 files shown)


pdns/syncres.cc
```diff
@@ -795,11 +795,12 @@ bool SyncRes::doCNAMECacheCheck(const DNSName &qname, const QType &qtype, vector
   return false;
 }
 
-static const DNSName getLastLabel(const DNSName& qname)
-{
-  DNSName ret(qname);
-  ret.trimToLabels(1);
-  return ret;
+static inline void addTTLModifiedRecords(const vector<DNSRecord>& records, const uint32_t ttl, vector<DNSRecord>& ret) {
```
rgacogne: I would advise to drop the `inline` keyword. The compiler will try to inline when it makes sense anyway, but with the keyword it will generate a warning if the inlining doesn't happen.

pdns/recursordist/negcache.cc
```diff
@@ -0,0 +1,155 @@
...
+ *
+ * \param qname    The name to look up (only the last label is used)
+ * \param now      A timeval with the current time, to check if an entry is expired
+ * \param ne       A NegCacheEntry that is filled when there is a cache entry
+ * \return         true if ne was filled out, false otherwise
+ */
+bool NegCache::getRootNXTrust(const DNSName& qname, const struct timeval& now, NegCacheEntry& ne) {
+  // An 'ENT' QType entry, used as "whole name" in the neg-cache context.
+  static const QType qtnull(0);
+  pair<negcache_t::const_iterator, negcache_t::const_iterator> range;
```
pieterlexis: hrm, come to think of it. We should actually iterate over it and check `d_auth.isRoot() && d_type == QType(0)` (whole name, not just some record). This in case the root denies com.|A, we don't want to deny .com 👍

pieterlexis: hrm, come to think of it. We should actually iterate over it and check `d_auth.isRoot() && d_type == QType(0)` (whole name, not just some record). This in case the root denies com.|A, we don't want to deny .com 👍

pdns/syncres.cc
pieterlexis: ok!

PR #5236, opened 2017-04-11
rec: Move to thread_local, refactor `SyncRes::doResolveAt()` and auth zones

rgacogne:
### Short description
- Refactor and split `SyncRes::doResolveAt()`, making it easier to understand
- Get rid of `SyncRes::d_nocache`, makes sure we can't get into a root refresh loop
- Clean up the Auth and Forward zones a bit
- Limit the use of global variables in `SyncRes`, to make it easier to understand the interaction between components
- Use C++11's `thread_local` instead of `__thread`, as the former supports non-trivial types, allowing us to get rid of many existing naked pointers in the rec. This requires gcc 4.8+ or clang 3.6+. 

### Checklist

I have:
- [ ] documented the code


rgacogne:
Rebased to fix conflicts.

PR #5241, opened 2017-04-14
Feature except send servfail

magicbear:
### Short description
Improve the resolve service availability on cluster. When one of the server has exception, if it sent ServFail, the resolver dns server will be directly return to the client. When I don't replay anything, the resolver dns server will be sent the request against to another server in the cluster.

### Checklist

I have:
- [x] documented the code

### DIFF
--- a/pdns/common_startup.cc
+++ b/pdns/common_startup.cc
@@ -67,6 +67,7 @@ void declareArguments()
   ::arg().setSwitch("forward-dnsupdate","A global setting to allow DNS update packages that are for a Slave domain, to be forwarded to the master.")="yes";
   ::arg().setSwitch("log-dns-details","If PDNS should log DNS non-erroneous details")="no";
   ::arg().setSwitch("log-dns-queries","If PDNS should log all incoming DNS queries")="no";
+  ::arg().setSwitch("except-send-servfail","If PDNS should send ServFail on backends exception")="yes";
   ::arg().set("local-address","Local IP addresses to which we bind")="0.0.0.0";
   ::arg().setSwitch("local-address-nonexist-fail","Fail to start if one or more of the local-address's do not exist on this server")="yes";
   ::arg().setSwitch("non-local-bind", "Enable binding to non-local addresses by using FREEBIND / BINDANY socket options")="no";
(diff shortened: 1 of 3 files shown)


rgacogne:
I'm inclined to think that this is something that should be handled at the load-balancer level, trying on another server on `ServFail`.

magicbear:
so I make this is a option to do, when the server hasn't load balance server and directly in the internet. It may make useful.

phonedph1:
I'm confused. You have a recursive resolver that will not retry another server in the NS set if it gets a ServFail from any one of them? Is this what you're trying to fix?

magicbear:
phonedph1: This patch is for authoritative server, not for recursive.

rgacogne:
I mislabeled this earlier, sorry. If I understand correctly, the idea is to add an option so that the auth doesn't send a Server Failure answer when it encounters an error but just drop the query, because it would then make some LB try another server of the farm.

pdns/common_startup.cc
```diff
@@ -67,6 +67,7 @@ void declareArguments()
   ::arg().setSwitch("forward-dnsupdate","A global setting to allow DNS update packages that are for a Slave domain, to be forwarded to the master.")="yes";
   ::arg().setSwitch("log-dns-details","If PDNS should log DNS non-erroneous details")="no";
   ::arg().setSwitch("log-dns-queries","If PDNS should log all incoming DNS queries")="no";
+  ::arg().setSwitch("except-send-servfail","If PDNS should send ServFail on backends exception")="yes";
```
pieterlexis: maybe change this to "If a SERVFAIL should be sent on a backend exception, drop the query otherwise"

pdns/packethandler.cc
```diff
@@ -1513,12 +1514,18 @@ DNSPacket *PacketHandler::doQuestion(DNSPacket *p)
...
-    S.ringAccount("servfail-queries",p->qdomain.toLogString());
+    if (d_exceptSendServFail) {
+      L<<Logger::Error<<"Exception building answer packet for "<<p->qdomain<<"/"<<p->qtype.getName()<<" ("<<e.what()<<") sending out servfail"<<endl;
+      delete r;
+      r=p->replyPacket(); // generate an empty reply packet
+      r->setRcode(RCode::ServFail);
+      S.inc("servfail-packets");
+      S.ringAccount("servfail-queries",p->qdomain.toLogString());
+    }else {
+      L<<Logger::Error<<"Exception building answer packet ("<<e.what()<<") ignore for send out servfail"<<endl;
```
pieterlexis: Perhaps change the second part of the logline to "not sending a response"

pdns/common_startup.cc
Habbie: maybe rename to `exception-sends-servfail`?

pieterlexis:
ping! Still want to finish this feature?

Habbie:
Any configuration flag that causes query blackholing is a bad idea. Closing this.

PR #5244, opened 2017-04-15
dnsdist: Stop sending queries as soon as we are dyn-blocked in tests

rgacogne:
### Short description

Otherwise we might take too much time to finish sending our queries (2s timeout per query), ending up with the dynamic block rule gone by the time we finish.
Hopefully this will reduce or even get rid of the random travis failures this test is causing.

### Checklist

I have:
- [ ] documented the code

### DIFF
--- a/regression-tests.dnsdist/test_DynBlocks.py
+++ b/regression-tests.dnsdist/test_DynBlocks.py
@@ -394,7 +394,7 @@ class TestDynBlockResponseBytes(DNSDistTest):
                 # let's clear the response queue
                 self.clearToResponderQueue()
                 # and stop right there, otherwise we might
-                # wait for so long than the dynblock is gone
+                # wait for so long that the dynblock is gone
                 # by the time we finished
                 break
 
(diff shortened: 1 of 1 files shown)


regression-tests.dnsdist/test_DynBlocks.py
```diff
@@ -393,6 +393,10 @@ def testDynBlocksResponseByteRate(self):
                 # the query has not reached the responder,
                 # let's clear the response queue
                 self.clearToResponderQueue()
+                # and stop right there, otherwise we might
+                # wait for so long than the dynblock is gone
```
Habbie: s/than/that/

regression-tests.dnsdist/test_DynBlocks.py
```diff
@@ -450,6 +454,10 @@ def testDynBlocksResponseByteRate(self):
...
```
Habbie: s/than/that/

rgacogne:
Typos fixed, thanks!

PR #5250, opened 2017-04-17
Implement axfrfilter with LuaContext

cmouse:
### Short description
Implements axfrfilter using the new luacontext system. It also changes slightly how the AXFR filter works, by allowing the hook to also prepend records instead of just replacing them.

### Checklist
I have:
- [ ] documented the code

### DIFF
--- a/pdns/lua-auth4.cc
+++ b/pdns/lua-auth4.cc
@@ -11,6 +11,7 @@
 
 AuthLua4::AuthLua4(const std::string& fname) { }
 bool AuthLua4::updatePolicy(const DNSName &qname, QType qtype, const DNSName &zonename, DNSPacket *packet) { return false; }
+bool AuthLua4::axfrfilter(const ComboAddress& remote, const DNSName& zone, const DNSResourceRecord& in, vector<DNSResourceRecord>& out) { return false; }
 AuthLua4::~AuthLua4() { }
 
 #else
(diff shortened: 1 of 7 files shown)


pdns/lua-auth4.cc
```diff
@@ -222,8 +224,45 @@ AuthLua4::AuthLua4(const std::string& fname) {
...
+bool AuthLua4::axfrfilter(const ComboAddress& remote, const DNSName& zone, const DNSResourceRecord& in, vector<DNSResourceRecord>& out) {
+  luacall_axfr_filter_t::result_type ret;
+
+  if (d_axfr_filter == NULL) return false;
+
+  ret = d_axfr_filter(remote, zone, in);
+
+  if (std::get<0>(ret) < 0)
+    return false;
+  else if (std::get<0>(ret) > 0)
```
Habbie: maybe do `== 1` here so we can later give 2 meaning as well. Also it might be good to make these constants that the code can use, just like `pdns.PASS`

pdns/lua-auth4.cc
```diff
@@ -222,8 +224,45 @@ AuthLua4::AuthLua4(const std::string& fname) {
...
+
+  for(const auto& row: rows) {
+    DNSResourceRecord rec;
+    for(const auto& col: row.second) {
+      if (col.first == "qtype")
+        rec.qtype = QType(pdns_stou(col.second));
+      else if (col.first == "qname")
+        rec.qname = DNSName(col.second);
+      else if (col.first == "ttl")
+        rec.ttl = pdns_stou(col.second);
```
Habbie: Why isn't `second` a number here?

pdns/lua-auth4.hh
```diff
@@ -5,6 +5,7 @@
 #include "dnsrecords.hh"
 #include "dnspacket.hh"
 #include <unordered_map>
+#include <boost/variant/variant.hpp>
```
Habbie: This header is not used (but see ttl question above).

Habbie:
Approved with minor nits. As discussed on IRC, the return value could use some docs.

cmouse:
i forgot to retest after latest change.

cmouse:
seems to work.

PR #5252, opened 2017-04-18
rec: Only apply root-nx-trust if the received SOA is "."

rgacogne:
### Short description

If `root-nx-trust` is enabled and we got a NX answer from the root, check that the received SOA is for the root before negatively caching the entire TLD. This might happen if "." is forwarded, for example.

### Checklist

I have:
- [ ] documented the code


pieterlexis:
Maybe we should ensure both are true (both the SOA NAME and the `auth` value to be '.')? I think this fix can also remove the "check if forwarded" comment.

rgacogne:
We checked earlier that the SOA name is part of `auth` (`rec.d_name.isPartOf(auth)`), so we should be fine, but I'll add an explicit check to be safe in case the earlier check goes away in the future.

PR #5260, opened 2017-04-21
LDAP backend DNSSEC support

goestreicher:
### Short description
This PR adds DNSSEC support in the LDAP backend for NSEC and all NSEC3 modes of operation.

All travis tests pass, except for ent-asterisk because of the old unbound version, but this tests works just fine with 1.6.0.

### Checklist
I have:
- [x] documented the code


modules/ldapbackend/dnssec.cc
```diff
@@ -0,0 +1,1416 @@
...
+{
+  std::string dn;
+  std::string filter = strbind( ":domain:", m_pldap->escape( name.toStringRootDot() ), getArg( "metadata-searchfilter" ) );
+  const char* attributes[] = { "objectClass", NULL };
+  PowerLDAP::sentry_t result;
+
+  try {
+    PowerLDAP::SearchResult* search = m_pldap->search( m_metadata_searchdn, LDAP_SCOPE_SUBTREE, filter, attributes );
+
+    if ( search->getNext( result, true ) ) {
```
rgacogne: It looks like `PowerLDAP::SearchResult::getNext()` can throw, which would result in `search` being lost. How about making `PowerLDAP::search()` return a `unique_ptr` instead?

docs/manpages/zone2ldap.1.md
```diff
@@ -20,6 +22,15 @@ inserted to an LDAP server.
 --basedn=*DN*
 :    Base DN to store objects below
 
+--create-ent
```
zeha: maybe this should be `--dnssec` if it's only used for dnssec-enabled zones? (but is this part actually true?)

docs/markdown/authoritative/backend-ldap.md
```diff
@@ -82,12 +92,18 @@ In case the used LDAP client library doesn't support LDAP URIs as connection par
 (default: "") : Full path to the Kerberos credential cache file to use. Actually only files are supported, and the "FILE:" prefix must not be set. The PowerDNS process must be able to write to this file and it *must* be the only one able to read it.
 
 ## `ldap-basedn`
-(default "") : The PowerDNS LDAP DNS backend searches below this path for objects containing the specified DNS information. The retrieval of attributes is limited to this subtree. This option must be set to the path according to the layout of your LDAP tree, e.g. ou=hosts,o=linuxnetworks,c=de is the DN to my objects containing the DNS information.
+(default: "") : The PowerDNS LDAP DNS backend searches below this path for objects containing the specified DNS information. The retrieval of attributes is limited to this subtree. This option must be set to the path according to the layout of your LDAP tree, e.g. ou=hosts,o=linuxnetworks,c=de is the DN to my objects containing the DNS information.
```
zeha: i'd think `default: empty` would be nicer, but I don't know the style used in other pdns docs for this.

docs/markdown/authoritative/backend-ldap.md
```diff
@@ -97,16 +113,24 @@ In case the used LDAP client library doesn't support LDAP URIs as connection par
...
 
 Schema update
 -------------
 
-First off adding master support to the LDAP backend needs
-a schema update. This is required as some metadata must
-be stored by PowerDNS, such as the last successful transfer
-to slaves. The new schema is available in
-schema/pdns-domaininfo.schema.
+First off adding master support to the LDAP backend requires
```
zeha: "," after "off"?

modules/ldapbackend/dnssec.cc
```diff
@@ -0,0 +1,1385 @@
...
+        dn = result["dn"][0];
+    }
+
+    L<<Logger::Debug<< m_myname << " getDomainMetadataDN will return " << dn << endl;
+    return dn;
+  }
+  catch( LDAPTimeout &lt )
+  {
+    L << Logger::Warning << m_myname << " Unable to search LDAP directory: " << lt.what() << endl;
+    throw( DBException( "LDAP server timeout" ) );
```
zeha: avoid braces after throw

modules/ldapbackend/dnssec.cc
```diff
@@ -0,0 +1,1385 @@
...
+  const char* attributes[] = { "PdnsMetadataValue", NULL };
+  PowerLDAP::sentry_t result;
+
+  try {
+    PowerLDAP::SearchResult::Ptr search = m_pldap->search( basedn, LDAP_SCOPE_SUBTREE, filter, attributes );
+
+    // We're supposed to get just one entry
+    if ( search->getNext( result, m_getdn ) ) {
+      if ( result.count( "PdnsMetadataValue" ) && !result["PdnsMetadataValue"].empty() ) {
+        for ( auto value : result["PdnsMetadataValue"] )
```
zeha: 1) i imagine this could be const auto&
2) curly braces for the for body would be appreciated for readability (too many levels of indent...)

modules/ldapbackend/ldapauthenticator.cc
```diff
@@ -83,8 +82,32 @@ static int ldapGssapiAuthenticatorSaslInteractCallback( LDAP *conn, unsigned fla
 }
 
 LdapGssapiAuthenticator::LdapGssapiAuthenticator( const std::string& kt, const std::string &ccache, int tmout )
-  : logPrefix( "[LDAP GSSAPI] " ), keytabFile( kt ), cCacheFile( ccache ), timeout( tmout )
+  : m_logPrefix( "[LDAP GSSAPI] " ), m_keytabFile( kt ), m_cCacheFile( ccache ), m_timeout( tmout )
```
zeha: most other pdns parts use `d_` instead of `m_`...

docs/manpages/zone2ldap.1.md
goestreicher: Well, it's not necessarily only used for DNSSEC zone, but it's mandatory for them. My main concern would be that renaming this option for '--dnssec' may lead users to believe that only this option is necessary to ready a zone for DNSSEC usage, while this is not true (--domainid must also be set, and I don't want to imply --domainid when --create-ent is in use).

docs/markdown/authoritative/backend-ldap.md
goestreicher: It seems that there's not standard, but I agree that your idea is much better. It'll be part of the next push, thanks :)

docs/manpages/zone2ldap.1.md
Habbie: This should not even be a flag. Always create ENTs.

docs/markdown/authoritative/backend-ldap.md
goestreicher: I have to trust you on this one, English is not my first language. Then it seems better this way, it'll be part of the next push, thanks!

docs/manpages/zone2ldap.1.md
goestreicher: If you are using the "tree" layout, then ENTs will be created without this flag. However with other layouts it's not necessarily wanted by the LDAP admin, which is why I think it's best to leave this as a separate option.

Then maybe adding a blanket "--dnssec" flag, as @zeha, propsed, that would turn on the necessary flags is something that can be added.

Habbie: ENTs are mandatory in DNS, not just in DNSSEC. We should attempt to do the right thing in all modes that we can.

goestreicher: Just to sum up what we discussed on IRC: I'll remove this option and let zone2ldap do The Right Thing©® from the start.

modules/ldapbackend/dnssec.cc
goestreicher: Thanks, will be part of the next push!

modules/ldapbackend/dnssec.cc
goestreicher: Thanks! There are other places where const auto& can be used, this will be part of next push.

modules/ldapbackend/ldapauthenticator.cc
goestreicher: OK, renamed all member variables :)

modules/ldapbackend/dnssec.cc
```diff
@@ -0,0 +1,1386 @@
+/*
+ * This file is part of PowerDNS or dnsdist.
+ * Copyright -- PowerDNS.COM B.V. and its contributors
+ * originally authored by Norbert Sendetzky
```
Habbie: feel free to put your name here as well

modules/ldapbackend/dnssec.cc
```diff
@@ -0,0 +1,1386 @@
...
+  // Same as in setDomainMetadata, we don't create the required entries
+  std::string basedn = this->getDomainMetadataDN( name );
+  if ( basedn.empty() )
+    return false;
+
+  try {
+    // Prepare all the elements first. As LDAP doesn't have the concept of transaction
+    // we risk ending up with two keys with the same ID if there are insertions between
+    // the moment we get the list of keys and the moment we insert. So better not dawdle
+    // between those.
```
Habbie: This seems hacky, just shortening the race window. Is there no way we can do better than this?

modules/ldapbackend/dnssec.cc
```diff
@@ -0,0 +1,1386 @@
...
+  catch( LDAPTimeout &lt )
+  {
+    L << Logger::Warning << d_myname << " Unable to search LDAP directory: " << lt.what() << endl;
+    throw DBException( "LDAP server timeout" );
+  }
+  catch( LDAPNoConnection &lnc )
+  {
+    L << Logger::Warning << d_myname << " Connection to LDAP lost, trying to reconnect" << endl;
+    if ( reconnect() )
+      return this->addDomainKey( name, key, id );
```
Habbie: I sense a potential stack overflow here.

modules/ldapbackend/ldapbackend.cc
```diff
@@ -39,35 +39,38 @@ LdapBackend::LdapBackend( const string &suffix )
 
   try
   {
-    m_msgid = 0;
```
Habbie: Don't see a + `d_msgid`, is that okay?

Habbie:
Can you please rebase and reorder so that
* there are no 'merges to master' in the commit list
* each commit is suitable for individual review, especially the refactors (like moving things into native.cc)
* the `m_`/`d_` change is isolated

Thanks!

goestreicher:
As discussed with Habbie I'll close this request and split it in multiple parts based on functionnalities.

Issue #2423, opened 2015-04-10
Feature request: don't use /dev/urandom as entropy source


paultheweird:
While starting PowerDNS 3.4.3 on OpenBSD (preview port, not in ports tree yet), I see the following in the logs:

```
[weerd@pom] $ pdns_server
Apr 10 18:15:36 Reading random entropy from '/dev/urandom'
```

Opening files in `/dev/` can be problematic in chroot environments or in fd-exhausted scenarios. Better to use the `arc4random` family of functions to get random data (see http://www.openbsd.org/cgi-bin/man.cgi?query=arc4random for details). This could be implemented either via the portable code present in LibreSSL (available under a very liberal license) or simply on systems that support it via autoconf.

Thanks!

rubenk:
Atlhough I'd love to use this, it seems it's not that portable yet.
LibreSSL currently turns it of for freebsd, netbsd and darwin: https://github.com/libressl-portable/portable/blob/master/configure.ac#L256

omoerbeek:
With the new dns_random (upcoming), this is mostly solved. There are a few places were random() is used, mostly in utilities. Those are not harmful and require a bit more work to eliminate, because using dns_random() there introduces dependencies on libsodium and or libcrypto.

PR #5274, opened 2017-04-26
dns_random: Implement new dns_random

cmouse:
### Short description
dns_random: Use RAND_pseudo_random_bytes instead of the old code

### Checklist
I have:
- [x] documented the code


cmouse:
It could use RAND_poll() too for seeding.

cmouse:
It is not inherently clear if dns_random API is supposed to guarantee cryptographically strong bytes or not. Also, the seed parameter could imply that the number generator is supposed to be resettable to known state, but I am not sure. With this change, the number generator cannot be reset to known state anymore. The API can be changed to require strong bytes, or it can be made optional requirement.

cmouse:
It seems that resettability is not a requirement. After testing with same byte array several times using the current code, results do not repeat themselves.

rgacogne:
So, hum, what is this change trying to solve/enhance?

rgacogne:
By the way, OpenSSL 1.1 states:

> RAND_pseudo_bytes() has been deprecated. Users should use RAND_bytes() instead.

Habbie:
ping - we don't understand the purpose of this PR, and the code uses a deprecated API

cmouse:
Well, I have not updated it to use non-deprecated API since there is still no clarity on how dns_random should even behave.

rgacogne:
From previous discussions, my understanding is that we don't _need_:
- a cryptographically secure pseudorandom number generator since we don't use it for crypto operations ;
- resettability.

What we do need:
- not baking our own crypto ;
- fast enough so that we can continue calling it for every outgoing query ;
- secure enough that our query IDs can't be predicted ;
- long period, or a way to re-seed long before it's needed.

What would be nice to have:
- a way to re-seed (add entropy, not reset) periodically if we want to ;
- cryptographically secure, so we don't mess up in the future by using it for something really sensitive ;
- fork-safety, or at least a way to re-seed after `fork()` ;
- `getrandom()` / `getentropy()` support when available so we don't have to worry about not having enough file descriptors, or about being in some kind of `chroot()` ;
- several options, switchable at run-time (`getrandom()`, `/dev/urandom`, `OpenSSL random API`, ...) so users can switch if an issue is uncovered in one option.

neilcook:
> cryptographically secure, so we don't mess up in the future by using it for something really sensitive 

While I agree that this is not *required*, I think we should just do it anyway. I worry about:
1) Someone inadvertently using it for something sensitive
2) Code-audits (or indeed other H1 reports) reporting it as cryptographically insecure, mainly because it's not clear what the intention is.

We should also document all the assumptions that we have about what it's supposed to be used for (and what not).

Habbie:
related: #2423

cmouse:
So it needs configure checks whether we can support particular entropy source (such as arc4random, openssl_random_bytes, or some other crypto library's, /dev/urandom etc). Should we just go for the full hog and allow choosing from whatever is available?

Habbie:
> So it needs configure checks whether we can support particular entropy source (such as arc4random, openssl_random_bytes, or some other crypto library's, /dev/urandom etc). Should we just go for the full hog and allow choosing from whatever is available?

I recall @rgacogne also voiced this preference.

cmouse:
Now implements a choosable rng including a simple rng for test purposes.

cmouse:
Ready for review

pdns/dns_random.cc
```diff
@@ -22,73 +22,251 @@
...
+#if defined(HAVE_RAND_BYTES)
+#include <openssl/rand.h>
+#endif
+#if defined(HAVE_GETRANDOM)
+#include <linux/random.h>
+#endif
 
-using namespace std;
+static enum DNS_RNG {
+  RNG_AUTO = 0,
```
rgacogne: I find it a bit misleading to call this state `auto`, perhaps something like `uninitialized` might be clearer?

pdns/common_startup.cc
```diff
@@ -203,6 +203,8 @@ void declareArguments()
   ::arg().set("xfr-max-received-mbytes", "Maximum number of megabytes received from an incoming XFR")="100";
 
   ::arg().set("tcp-fast-open", "Enable TCP Fast Open support on the listening sockets, using the supplied numerical value as the queue size")="0";
+
+  ::arg().set("rng", "Specify random number generator to use. Valid values are auto,sodium,openssl,getrandom,arc4random,urandom,kiss.")="auto";
```
pieterlexis: Maybe not mention `kiss` here to protect people who don't read the docs but only --help (or the annotated pdns.conf)?

pdns/pdns_recursor.cc
```diff
@@ -3147,6 +3146,8 @@ static int serviceMain(int argc, char*argv[])
 
   openssl_thread_setup();
   openssl_seed();
+  /* setup rng before chroot */
+  (void)dns_random(1);
```
rgacogne: This seems like an ugly way to initialize the PRNG, perhaps we could make a version of `dns_random_init()` taking no parameter instead?

pdns/recursordist/configure.ac
```diff
@@ -123,7 +123,7 @@ PDNS_WITH_NET_SNMP
 PDNS_CHECK_RAGEL([pdns/dnslabeltext.cc], [www.powerdns.com])
 PDNS_CHECK_CURL
 
-AC_CHECK_FUNCS([strcasestr])
+AC_CHECK_FUNCS_ONCE([strcasestr localtime_r recvmmsg getrandom arc4random])
```
rgacogne: We don't need `localtime_r` and `recvmmsg` here.

cmouse:
nits fixed.

cmouse:
also prioritized getrandom and arc4random over openssl and libsodium.

pdns/common_startup.cc
```diff
@@ -203,6 +203,8 @@ void declareArguments()
   ::arg().set("xfr-max-received-mbytes", "Maximum number of megabytes received from an incoming XFR")="100";
 
   ::arg().set("tcp-fast-open", "Enable TCP Fast Open support on the listening sockets, using the supplied numerical value as the queue size")="0";
+
+  ::arg().set("rng", "Specify random number generator to use. Valid values are auto,sodium,openssl,getrandom,arc4random,urandom.")="auto";
```
rgacogne: s/Specify random number generator/Specify the random number generator/ ?

pdns/dns_random.cc
```diff
@@ -22,73 +22,263 @@
...
+    // some systems define getrandom but it does not really work, e.g. because it's
+    // not present in kernel.
+    if (getrandom(buf, sizeof(buf), 0) == -1) {
+       L<<Logger::Warning<<"getrandom() failed: "<<strerror(errno)<<", falling back to " + rdev<<std::endl;
+       chosen_rng = RNG_URANDOM;
+    }
+  }
+# endif
+
+#if defined(HAVE_RAND_BYTES)
```
rgacogne: Could use a space between # and if for consistency.

pdns/dns_random.cc
```diff
@@ -22,73 +22,263 @@
...
+         (data[8] + (data[9]<<8) + (data[10]<<16) + (data[11]<<24)) ^
+         (data[12] + (data[13]<<8) + (data[14]<<16) + (data[15]<<24));
+  kiss_init(seed);
 }
 
-#if 0
-int main()
-{
-  dns_random_init("0123456789abcdef");
+/* Parts of this code come from arc4random_uniform */
```
rgacogne: You might want to include the comments about "modulo bias" too, since that explain why you bother computing a lower threshold and looping until the number is ok.

pdns/dns_random.cc
```diff
@@ -22,73 +22,263 @@
...
+} chosen_rng = RNG_UNINITIALIZED;
 
-static AES_KEY aes_key;
-static unsigned int g_offset;
-static unsigned char g_counter[16], g_stream[16];
-static uint32_t g_in;
+static int urandom_fd = -1;
+
+/* KISS is intended for development use only */
+static unsigned int kiss_seed;
```
zeha: i'd argue it should be under a configure flag `--dangerous-unsafe-rng`

rgacogne: Agreed, or even an `#ifdef 0` so that you have to edit the code to be able to compile it in.

pdns/dns_random.cc
cmouse: added comment.

cmouse:
rebased to fix conflict. fixed build issue with calidns.

cmouse:
rebased, let's hope we can merge this soon =)

PR #5288, opened 2017-05-03
Make sure TTL is a positive signed 32 bit number as per RFC

mekmitasdigoat:
### Short description
Make sure the TTL does not go out of spec by going negative in the packet cache.

### DIFF
--- a/pdns/dnsparser.cc
+++ b/pdns/dnsparser.cc
@@ -640,10 +640,14 @@ public:
     const char *p = d_packet + d_offset;
     moveOffset(4);
 
-    uint32_t tmp;
+    // This is used for TTL. See RFC 1035 section 2.3.4
+    int32_t tmp;
     memcpy(&tmp, (void*) p, sizeof(tmp));
-    tmp = ntohl(tmp);
+    tmp = (int32_t)ntohl(tmp);
     tmp-=decrease;
+    if (tmp < 0) {
+      tmp = 0;
+    }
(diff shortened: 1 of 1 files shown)


Habbie:
This is wrong, see http://www.rfc-editor.org/errata_search.php?rfc=1035&rec_status=15&presentation=records on TTLs being unsigned 32 bits.

On 1 June 2017 16:25:09 CEST, Remi Gacogne <notifications@github.com> wrote:
>rgacogne commented on this pull request.
>
>The change as it is looks wrong to me, as nothing guarantees that the
>32-bit value about to be decreased is a TTL. This change might cause
>issues with unsigned 32-bit larger than `INT32_MAX`.
>What issue are you trying to solve exactly? Right now this code is only
>called by `ageDNSPacket()`, used by the packet caches of dnsdist and
>the recursor. Did you see a case where either one of them sent an
>invalid TTL on the wire? If so, I think we should fix the root cause of
>the issue instead of crippling `decreaseAndSkip32BitInt()`.
>
>
>
>-- 
>You are receiving this because you are subscribed to this thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/PowerDNS/pdns/pull/5288#pullrequestreview-41525166

-- 
Sent from my Android phone with K-9 Mail. Please excuse my brevity.

Habbie:
Closing this as it is wrong and there was no feedback.

PR #5293, opened 2017-05-04
rec: Allow retrieving stats from Lua via the `getStat("name")` call

rgacogne:
### Short description

Until now, only custom metrics could be set and retrieved from Lua hooks. This PR adds the possibility to retrieve the internal statistics via a new `getStat("name") call.

### Checklist

I have:
- [ ] documented the code

### DIFF
--- a/docs/markdown/recursor/scripting.md
+++ b/docs/markdown/recursor/scripting.md
@@ -377,7 +377,7 @@ output and weird results.
 
 ### Statistics
 
-You can retrieve statistics from Lua using the `getStat("name")` call. For example,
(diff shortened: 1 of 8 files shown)


docs/markdown/recursor/scripting.md
```diff
@@ -375,6 +375,18 @@ Note that metrics live in the same namespace as 'system' metrics. So if you
 generate one that overlaps with a PowerDNS stock metric, you will get double
 output and weird results.
 
+### Statistics
```
pieterlexis: Maybe add "since 4.1.0"

rgacogne: Nice catch, done. Thanks!

PR #5319, opened 2017-05-12
rec: Only check the netmask for subnet specific cache entries

rgacogne:
### Short description

We used to check the netmask for all entries for a qname if at least one of them was a subnet specific one. Since an empty `Netmask` doesn't match anything, we would effectively ignore every
non subnet specific entries if we had at least one subnet specific one.
This was part of a very hard to reproduce issue with for example f.root-servers.net that includes an EDNS Client Subnet option in its answer for `NS .` if the query has an EDNS Client Subnet option.
This caused the recursor to cache a subnet specific entry for `NS .`. When that entry expired, we retrieved and cached a non subnet specific one, but that new one was ignored as long as the subnet specific was not expunged from the cache.

### Checklist

I have:
- [ ] documented the code

### DIFF
--- a/pdns/recursor_cache.cc
+++ b/pdns/recursor_cache.cc
@@ -50,18 +50,11 @@ int32_t MemRecursorCache::get(time_t now, const DNSName &qname, const QType& qt,
   if(res)
     res->clear();
 
-  bool haveSubnetSpecific=false;
   if(d_cachecache.first!=d_cachecache.second) {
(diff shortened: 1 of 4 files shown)


zeha:
just reading this... might want a test?

rgacogne:
I just pushed a new commit adding unit tests for the `MemRecursorCache` class, one of them covering this exact issue.

PR #5320, opened 2017-05-13
 PowerDNS sdig does not truncate trailing bits of EDNS Client Subnet mask

ahupowerdns:
So if you'd truncate something as
a /9, we'd have to use 2 bytes anyhow, but we would not zero the last 7 bits.

We do now. Thanks Mukund & ISC!

### Short description
Use ComboAddress::truncate logic.

### Checklist

I have:
- [ ] documented the code


rgacogne:
Every single unit test falls on a byte boundary.. I'll have to add some to cover this case.

PR #5336, opened 2017-05-23
Implement a runtime changeable rule that matches IP address for a certain time

ahupowerdns:
This effectively allows (for example) pool selection from Lua, but then cached.

Sample code:

```
newServer({address="192.168.1.20", pool=""})
newServer({address="8.8.8.8", pool="elgoog"})

tisrElGoog=TimedIPSetRule()
tisrRest=TimedIPSetRule()
addAction(tisrElGoog:slice(), PoolAction("elgoog"))
addAction(tisrRest:slice(), PoolAction(""))

elgoogPeople=newNMG()
elgoogPeople:addMask("192.168.1.0/28")

function pickPool(dq)
	if(elgoogPeople:match(dq.remoteaddr)) -- in real life, this would be external
	then
		print("Lua caught query for a googlePerson")
		tisrElGoog:add(dq.remoteaddr, 10)
		return DNSAction.Pool, "elgoog"
	else
		print("Lua caught query for restPerson")
		tisrRest:add(dq.remoteaddr, 60)
		return DNSAction.None, ""
	end
end

addLuaAction(AllRule(), pickPool)
```

### Short description
A new DNSRule that can be manipulated from Lua

### Checklist

I have:
- [ ] documented the code

### DIFF
--- a/pdns/README-dnsdist.md
+++ b/pdns/README-dnsdist.md
@@ -439,6 +439,7 @@ A DNS rule can be:
  * a RecordsTypeCountRule
  * a SuffixMatchNodeRule
  * a TCPRule
+ * a TimedIPSetRule
  * a TrailingDataRule
 
 Some specific actions do not stop the processing when they match, contrary to all other actions:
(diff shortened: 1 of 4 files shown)


rgacogne:
Could use a bit of documentation, console completion support and tests :-)

rgacogne:
Looks very good but needs to be rebased!

ahupowerdns:
How does this PR compare to your dynamic rule based work @rgacogne ? Complete overlap? Or not? Is this PR still worth it? I think so, but would like to be sure we are not merging this PR because ahu wrote it.

rgacogne:
The recent dynamic block changes allow different actions than just dropping, but it's clearly a lot more limited than what can be achieved with this PR!

PR #5338, opened 2017-05-23
See #3602, slave with multiple masters always reach alive master

oridistor:
### Short description

It's a fix as part of my issue #3602. What this does is the following:

1. Make sure that if an AXFR request was made by a master, this master will be the only possible master to contact (as we know that it lives).
2. Make sure that we pass the randomized master list to the code that asks the master for its zone file, as that way we won't just ask the first master, that might be dead.

### Checklist

I have:
- [X] documented the code

### DIFF
--- a/pdns/slavecommunicator.cc
+++ b/pdns/slavecommunicator.cc
@@ -509,6 +509,19 @@ void CommunicatorClass::addSlaveCheckRequest(const DomainInfo& di, const ComboAd
   Lock l(&d_lock);
   DomainInfo ours = di;
   ours.backend = 0;
+  string remote_address = remote.toString();
+
+  // When adding a check, if the remote addr from which notification was
+  // received is a master, clear all other masters so we can be sure the
+  // query goes to that one.
+  for (const auto& master : ours.masters) {
+    if (master == remote_address) {
+      ours.masters.clear();
+      ours.masters.push_back(remote_address);
+      break;
+    }
+  }
+  d_tocheck.erase(di);
   d_tocheck.insert(ours);
   d_any_sem.post(); // kick the loop!
 }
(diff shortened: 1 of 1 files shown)


pdns/slavecommunicator.cc
```diff
@@ -509,6 +509,19 @@ void CommunicatorClass::addSlaveCheckRequest(const DomainInfo& di, const ComboAd
   Lock l(&d_lock);
   DomainInfo ours = di;
   ours.backend = 0;
+  string remote_address = remote.toString();
+
+  // When adding a check, if the remote addr from which notification was
+  // received is a master, clear all other masters so we can be sure the
+  // query goes to that one.
+  for (vector<string>::iterator it = ours.masters.begin(); it != ours.masters.end(); ++it) {
```
pieterlexis: Please use C++ 2011 for loops `for(const auto& master : ours.masters) { if (master == remote_address) { ... } }`

pdns/slavecommunicator.cc
```diff
@@ -509,6 +509,19 @@ void CommunicatorClass::addSlaveCheckRequest(const DomainInfo& di, const ComboAd
...
+    if (*it == remote_address) {
+	ours.masters.clear();
+        ours.masters.push_back(remote_address);
```
pieterlexis: Can you please fix the indenting?

pieterlexis:
Also, can you first PR this against master if this is relevant?

oridistor:
@pieterlexis so should I open a merge request for master only or for both master and auth-3.4.x?

Habbie:
@oridistor in general, PR against master, and when accepted, then do the backport. Further, note that it is unlikely we will accept this in 3.4.x. It might be nice for 4.0.x though.

oridistor:
@pieterlexis created the new PR with what you requested and a small fix we've seen in our production environment (but never on our QA environment, which is worriesome...) here:
https://github.com/PowerDNS/pdns/pull/5361

@Habbie so there won't be any new version for 3.4.x? I'm asking because we're using 3.4.x at our environment and I wonder if it means we should test upgrading up to the newer versions.

PR #5361, opened 2017-06-01
See #3602, made requests always return to sender, for usage in multi …

oridistor:
…master slave zones. Also - made sure that the master that is questioned for updates will be selected randomly, to prevent repeatidally asking a dead master for updates

### Short description
It's a fix as part of my issue #3602. What this does is the following:

1. Make sure that if an AXFR request was made by a master, this master will be the only possible master to contact (as we know that it lives).
2. Make sure that we pass the randomized master list to the code that asks the master for its zone file, as that way we won't just ask the first master, that might be dead.
3. Make sure that if there is no backend for the DomainInfo (because it was received from packethandler) we make sure it receives it from the Temporary copy (as to not SEGFAULT later on. - This occured on our production, only rarely, but still).

### Checklist

I have:
- [X] documented the code

### DIFF
--- a/pdns/slavecommunicator.cc
+++ b/pdns/slavecommunicator.cc
@@ -710,6 +710,19 @@ void CommunicatorClass::addSlaveCheckRequest(const DomainInfo& di, const ComboAd
   Lock l(&d_lock);
   DomainInfo ours = di;
   ours.backend = 0;
+  string remote_address = remote.toString();
+
+  // When adding a check, if the remote addr from which notification was
+  // received is a master, clear all other masters so we can be sure the
+  // query goes to that one.
+  for (const auto& master : ours.masters) {
+    if (master == remote_address) {
+      ours.masters.clear();
+      ours.masters.push_back(remote_address);
+      break;
+    }
+  }
+  d_tocheck.erase(di);
   d_tocheck.insert(ours);
   d_any_sem.post(); // kick the loop!
 }
@@ -856,10 +869,17 @@ void CommunicatorClass::slaveRefresh(PacketHandler *P)
   time_t now = time(0);
   for(val_t& val :  sdomains) {
     DomainInfo& di(val.di);
+    DomainInfo tempdi;
     // might've come from the packethandler
-    if(!di.backend && !B->getDomainInfo(di.zone, di)) {
+    // Please do not overwrite received DI just to make sure it exists in backend.
+    if(!di.backend) {
+      if (!B->getDomainInfo(di.zone, tempdi)) {
         L<<Logger::Warning<<"Ignore domain "<< di.zone<<" since it has been removed from our backend"<<endl;
         continue;
+      }
+      // Backend for di still doesn't exist and this might cause us to
+      // SEGFAULT on the setFresh command later on
+      di.backend = tempdi.backend;
     }
 
     if(!ssr.d_freshness.count(di.id)) { // If we don't have an answer for the domain


oridistor:
@pieterlexis @Habbie if there is anything more needed to be done so we can merge this in, as the tests have passed on this commit finally, I would be glad to know and take care of them ASAP

pdns/slavecommunicator.cc
```diff
@@ -710,6 +710,19 @@ void CommunicatorClass::addSlaveCheckRequest(const DomainInfo& di, const ComboAd
   Lock l(&d_lock);
   DomainInfo ours = di;
   ours.backend = 0;
+  string remote_address = remote.toString();
+
+  // When adding a check, if the remote addr from which notification was
+  // received is a master, clear all other masters so we can be sure the
+  // query goes to that one.
+  for (const auto& master : ours.masters) {
```
rgacogne: I'm not sure we should remove all other masters. If for some reason like a firewall issue we can receive a `NOTIFY` from a master but are unable to open a `TCP` connection to it for the actual `AXFR` we are stuck. Perhaps we should just move this master to the front of the vector so we try it first, but can still fall back to another master if that doesn't work out?

pdns/slavecommunicator.cc
```diff
@@ -856,10 +869,17 @@ void CommunicatorClass::slaveRefresh(PacketHandler *P)
...
-    if(!di.backend && !B->getDomainInfo(di.zone, di)) {
+    // Please do not overwrite received DI just to make sure it exists in backend.
+    if(!di.backend) {
+      if (!B->getDomainInfo(di.zone, tempdi)) {
         L<<Logger::Warning<<"Ignore domain "<< di.zone<<" since it has been removed from our backend"<<endl;
         continue;
+      }
+      // Backend for di still doesn't exist and this might cause us to
+      // SEGFAULT on the setFresh command later on
+      di.backend = tempdi.backend;
```
rgacogne: Perhaps we should explicitly check that `di.backend` is not `nullptr` after that and bail out otherwise? I don't think `getDomainInfo()` should succeed without setting `DomainInfo::backend` but the check is cheap and would prevent a `nullptr` dereference leading to a crash. And at a quick glance it looks like the `LDAP` backend doesn't set `DomainInfo::backend` in `getDomainInfo()`, but I'll open a separate issue for that.

pdns/slavecommunicator.cc
oridistor: @rgacogne the reason why I do it like this is that later on in the code we randomize the list of ours.masters again and ask a random master if there are any fixes, but in the end it'll still go to the first original master.
By removing all the masters temporarily I make sure that it'll only go back to the notifier, which is safer than the current code that always sends it to the same master.
I do agree that if for some weird reason there is only a one way connection, this thing will cause problems, but I also think that the original code will do a worse job in the case you have stated and we should assume that if a target in our masters list does not answer us it's because it's dead and not blocked.
It should be discouraged to have a master in the list that blocks us by firewall.

pdns/slavecommunicator.cc
oridistor: @rgacogne as I recall, the issue is that we have under the nullptr (at 0) a backend class up and running for the default use case. That's why I didn't check the nullptr because it didn't fix the problem when I tried it on a real use case.

rgacogne: Right, I'm fine with skipping that check since that should not happen in the first place.

pdns/slavecommunicator.cc
rgacogne: Oh, you are right, I missed the shuffling in `SlaveSenderReceiver::send()`. Perhaps it would make sense to remove the shuffling from `SlaveSenderReceiver::send()`, make sure the notifier is the first masters on the list and shuffle the remaining ones? This way we would try the notifier first, which is your original intent and makes sense to me, but we could still fallback to another one if needed?

oridistor: @rgacogne I can try adding that change, but I think that with how "send" is implemented it's redundant, as "send" shuffles the master list in order to check a random master, as it only works on "masters.begin()", meaning shuffling all masters but the first will not change who we address.

Also, I think we can get to the "send" (from slaveRefresh) through a flow that does not go through a notify (and addSlaveCheckRequest), in which case we do want to shuffle the entire list, and not leave the first master as the first master. We are unware in "send" if we got there through "notify" or not.

rgacogne: Oh, you are right, I missed the fact that the `SlaveSenderReceiver::send()` only tries one master even though you did mention it earlier, sorry. In that case your current solution makes sense, although I'm still a bit worried about an asymmetric network issue, but I guess we would need to refactor the code to try more than one master to address it.

PR #5362, opened 2017-06-01
rec: Add support for the new ASAN fiber switch API

rgacogne:
### Short description

`clang` >= 4.0.0 and `gcc` > 7.1, at least, use the new ASAN fiber switch interface taking three pointers instead of just one.

### Checklist

I have:
- [ ] documented the code

### DIFF
--- a/m4/pdns_enable_sanitizers.m4
+++ b/m4/pdns_enable_sanitizers.m4
@@ -44,8 +44,37 @@ AC_DEFUN([PDNS_ENABLE_ASAN], [
         AC_CHECK_HEADERS([sanitizer/common_interface_defs.h], asan_headers=yes, asan_headers=no)
         AS_IF([test x"$asan_headers" = "xyes" ],
           [AC_CHECK_DECL(__sanitizer_start_switch_fiber,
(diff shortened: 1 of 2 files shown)


m4/pdns_enable_sanitizers.m4
```diff
@@ -44,8 +44,45 @@ AC_DEFUN([PDNS_ENABLE_ASAN], [
...
+                    [
+                      AC_MSG_RESULT([unknown])
+                      AC_MSG_NOTICE([ASAN fiber switching is not available due to an unknown API version])
+                    ]
+                  )
+                ]
+              )
+            ],
+            [
+              AC_MSG_NOTICE([ASAN fiber switching is not available])
```
pieterlexis: Maybe use `AC_MSG_ERROR`? as we can't do ASAN but it was requested?

rgacogne: It would be nice for the recursor but we still want to be able to build the auth and dnsdist with ASAN enabled even if the fiber switching API is not available.

rgacogne:
Fixed the indentation, thanks!

PR #5381, opened 2017-06-06
Fix query in howtos.md doc

kevinquinnyo:
I'm not sure if this is a bug in the docs, a bug in pdns, or an old bug
that was fixed. I will open an issue to reference this.

### Short description
Query for SOA record as-is in documentation results in 500 errors when trying to view or edit the zone, like this:

```
Jun  6 13:52:05 debian pdns_server[1108]: Jun 06 13:52:05 HTTP ISE for "/api/v1/servers/localhost/zones/example.com": STL Exception: Parsing record content (try 'pdnsutil check-zone'): missing field at the end of record content 'localhost admin@example.com 1'
```

### Checklist
I have:
- [ ] documented the code


Habbie:
This is a nasty discrepancy between the API and the rest of PowerDNS.

PR #5383, opened 2017-06-06
dnsdist: Change dnsdist stats functions to always return lowercase names

### LINKED ISSUE #5287: Normalize 0x20 queries in dnsdist statistics
- Program: dnsdist
 - Issue type: Bug report

### Short description

dnsdist copies the QNAME as sent by the client into the buffer used for generating statistics. As a result a client sending 0x20 queries generates odd output in topQueries().

### Environment
 - Software version: 1.1.0
 - Software source: pdns repo

### Other information
Example from production:
```
> topQueries(1000)
 100  0.conTRol4.Pool.NTp.org.                   13  0.1%
 135  MX.pool.ntp.org.                            6  0.1%
 198  1.lEDe.pOoL.ntP.oRg.                        2  0.0%
 246  _ntp._UdP.SG.POOL.NtP.Org.                  1  0.0%
 272  0.XX.pool.ntp.org.                          1  0.0%
 275  1.fReEbsd.pOol.nTP.Org.                     1  0.0%
 276  0.Se.pool.ntp.Org.                          1  0.0%

```

RobinGeuze:
### Short description
Changes getTopQueries and getGenResponses to return lowercase versions for all names. Closes #5287 

### Checklist
I have:
- [ ] documented the code

### DIFF
--- a/pdns/dnsdist-lua.cc
+++ b/pdns/dnsdist-lua.cc
@@ -158,7 +158,7 @@ std::unordered_map<int, vector<boost::variant<string,double>>> getGenResponses(u
   vector<pair<int, DNSName>> rcounts;
   rcounts.reserve(counts.size());
   for(const auto& c : counts) 
-    rcounts.push_back(make_pair(c.second, c.first));
+    rcounts.push_back(make_pair(c.second, c.first.makeLowerCase()));
   
   sort(rcounts.begin(), rcounts.end(), [](const decltype(rcounts)::value_type& a, 
                                           const decltype(rcounts)::value_type& b) {
@@ -1351,7 +1351,7 @@ vector<std::function<void(void)>> setupLua(bool client, const std::string& confi
       vector<pair<int, DNSName>> rcounts;
       rcounts.reserve(counts.size());
       for(const auto& c : counts) 
-	rcounts.push_back(make_pair(c.second, c.first));
+	rcounts.push_back(make_pair(c.second, c.first.makeLowerCase()));
 
       sort(rcounts.begin(), rcounts.end(), [](const decltype(rcounts)::value_type& a, 
 					      const decltype(rcounts)::value_type& b) {


pdns/dnsdist-lua.cc
```diff
@@ -158,7 +158,7 @@ std::unordered_map<int, vector<boost::variant<string,double>>> getGenResponses(u
   vector<pair<int, DNSName>> rcounts;
   rcounts.reserve(counts.size());
   for(const auto& c : counts) 
-    rcounts.push_back(make_pair(c.second, c.first));
+    rcounts.push_back(make_pair(c.second, c.first.makeLowerCase()));
```
rgacogne: I think you need to lowercase earlier otherwise the counts are going to be off.

pdns/dnsdist-lua.cc
```diff
@@ -1351,7 +1351,7 @@ vector<std::function<void(void)>> setupLua(bool client, const std::string& confi
       vector<pair<int, DNSName>> rcounts;
       rcounts.reserve(counts.size());
       for(const auto& c : counts) 
-	rcounts.push_back(make_pair(c.second, c.first));
+	rcounts.push_back(make_pair(c.second, c.first.makeLowerCase()));
```
rgacogne: Same thing here.

pdns/dnsdist-lua.cc
RobinGeuze: Nope since it uses maps, which uses the default compare and less operators, which in turn lowercase everything themselves.

pdns/dnsdist-lua.cc
RobinGeuze: As above.

pdns/dnsdist-lua.cc
rgacogne: Oh, you are right, my bad!

rgacogne: I wonder if it wouldn't be better to lowercase a bit later then, replacing `rc.second.toString()` by something like `rc.second.makeLowerCase().toString()`, so we only do the lowercasing for the names we are going to print, not all of them.

PR #5386, opened 2017-06-08
dnsdist: Make a `truncate` action available to DynBlock and Lua

rgacogne:
### Short description

This allows forcing a client to retry over TCP instead of dropping or refusing the query.

### Checklist

I have:
- [ ] documented the code


rgacogne:
Yes, this is indeed in addition to the previous PR allowing other actions in a dynamic context than just dropping the query. For now only selected actions are available in that context though, but should the need arises we could add support for almost all existing actions.

PR #5390, opened 2017-06-11
Remove old Lua code

cmouse:
### Short description
Removes all old lua code from PowerDNS auth server, and provides codesharing between auth and recursor.

### Checklist
I have:
- [x] documented the code


cmouse:
Currently stuck at _tostring

The LuaContext used keeps the functions available stored in global table. I have not yet been able to decipher how the mapping between object<=>function name<=>function is done, but it seems to rely on pointer to the proper member function. This makes things difficult. I am **assuming** that there is, somewhere, some way to find out what `object:function` is, and that you can actually figure out this on run-time.

It is possible to get a live pointer to the C++ object in question, but I have not yet found a way to execute ::toString only if there is such member function, **or** figure out how to see if there is a _tostring function mapped for the object or not.

It is probably easier to find out the latter than the first, since this information is kept at global function table stored somewhere.

cmouse:
Please merge !5468 first

cmouse:
Also pending on https://github.com/ahupowerdns/luawrapper/pull/35

cmouse:
Not sure why the test fails, I cannot make it fail locally.

cmouse:
The reason the tests fail is that recursor test uses pdns auth server 4.0.4, which is obviously not going to include the lua changes done here. I am not sure how to resolve this problem, please advise.

pdns/lua-auth4.cc
```diff
@@ -136,6 +136,9 @@ bool AuthLua4::axfrfilter(const ComboAddress& remote, const DNSName& zone, const
 
 
 bool AuthLua4::updatePolicy(const DNSName &qname, QType qtype, const DNSName &zonename, DNSPacket *packet) {
+  // default decision is all goes
```
Habbie: Does this need a documentation update?

cmouse: no, it's just explicitly documented here.

Habbie:
In the "ext/luawrapper: Synchronize with upstream" commit message, can you clarify synchronize? If it is a full sync, please mention the commit you synced with. If it's an import of one patch, please mention that.

PR #5391, opened 2017-06-12
altered postgres schema to match mysql schema

stasic:
### Short description
mysql uses BIGINT for records id. 
postgres is also going to use BIGINT (BIGSERIAL) too.

### Checklist

I have:
- [ ] documented the code


zeha:
change lgtm, but it'd be good to have a schema migration SQL, too...

noteby: gpgsqlbackend never reads `records.id`.

stasic:
made a new pull request with your suggestions:  #5426

PR #5397, opened 2017-06-13
also set d_requestor without Lua: the ECS logic needs it

ahupowerdns:
### Short description
EDNS Client Subnet would only work if a Lua script was loaded in git master

### Checklist
I have:
- [X] documented the code


pieterlexis:
Some ECS code was backported to Recursor 4.0.5, would that version be affected too (a.k.a. do we need 4.0.6 soon?)

rgacogne:
@pieterlexis Yes, it's affected too but there is an easy workaround by just loading an empty Lua file. If we work on 4.0.6, we need #5372 and #5399 there too.

PR #5399, opened 2017-06-14
with this, EDNS Client Subnet becomes compatible with the packet cache

ahupowerdns:
with this, EDNS Client Subnet becomes compatible with the packet cache while using the existing variable answer facility.

### Short description
Before this PR, the packetcache applied to ECS-variable answers, which meant the packetcache had to be disabled for ECS service.

We now mark ECS answers as variable, using the existing meant-for-lua setVariable answers.
### Checklist

I have:
- [ ] documented the code

### DIFF
--- a/pdns/recursor_cache.cc
+++ b/pdns/recursor_cache.cc
@@ -34,7 +34,7 @@ unsigned int MemRecursorCache::bytes()
 }
 
 // returns -1 for no hits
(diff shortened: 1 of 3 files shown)


pdns/recursor_cache.cc
```diff
@@ -56,7 +56,9 @@ int32_t MemRecursorCache::get(time_t now, const DNSName &qname, const QType& qt,
 			    (qt.getCode()==QType::ADDR && (i->d_qtype == QType::A || i->d_qtype == QType::AAAA) )) 
 			    && (i->d_netmask.empty() || i->d_netmask.match(who)))
          ) {
-
+        if(variable && !i->d_netmask.empty()) {
+          *variable=true;
+        }
```
rgacogne: Indentation looks wrong here

PR #5403, opened 2017-06-14
rec: Use the incoming ECS for cache lookup if `use-incoming-edns-subnet` is set

rgacogne:
### Short description

Otherwise we insert into the cache based on the incoming ECS but later do the lookup based on the query's source IP.

### Checklist

I have:
- [ ] documented the code


rgacogne:
Rebased to fix a conflict.

PR #5406, opened 2017-06-14
rec: Make more specific Netmasks < to less specific ones

rgacogne:
### Short description

Having the most specific ones first, then the less specific ones then the empty one makes it easier to match the most specific first, in that case in the cache.

Please don't merge before we ponder the impact of this change a bit more!

### Checklist

I have:
- [ ] documented the code


Habbie:
PR is conflicted

PR #5410, opened 2017-06-15
rec: Do exact qname/qtype matching in the cache when possible

rgacogne:
### Short description

This prevents us from having to iterate over all types, and especially all subnet-specific entries for all types, if we are looking for a specific type (not ANY, not ADDR, think CNAME for example).
One drawback, it might make the `d_cachecache` hack much less efficient.

### Checklist

I have:
- [ ] documented the code


rgacogne:
Superseded by #5461.

PR #5416, opened 2017-06-15
rec: Move expired cache entries to the front so they are expunged

rgacogne:
### Short description

It turns out that the current code does not move them. We used to have per-record TTD so checking if the vector containing the results was empty was a good way to check that the entry only contained expired records, but the current code doesn't even enter the condition if the TTD is expired.

<strike>Don't merge this until I have taken the time to add unit tests for this somehow (or at the very least open an issue so I remember to).</strike>

### Checklist

I have:
- [ ] documented the code

### DIFF
--- a/pdns/cachecleaner.hh
+++ b/pdns/cachecleaner.hh
@@ -61,7 +61,7 @@ template <typename T> void pruneCollection(T& collection, unsigned int maxCached
     else
       ++iter;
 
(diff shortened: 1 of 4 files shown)


rgacogne:
Partially overlaps with #5461, I'll move the tests there.

rgacogne:
Tests have been added to #5461.

PR #5417, opened 2017-06-15
howtos: avoid unrestricted recursive resolution in 4.0.x ALIAS example

lifeforms:
### Short description
Previous 4.0.x example for ALIAS opened up an unrestricted recursive server due to `allow-recursion` default. Make the example secure.

### Checklist
I have:
- [ ] documented the code

### DIFF
--- a/docs/markdown/authoritative/howtos.md
+++ b/docs/markdown/authoritative/howtos.md
@@ -187,10 +187,11 @@ expand-alias=yes
 
 **note**: If `resolver` is unset, ALIAS expension is disabled!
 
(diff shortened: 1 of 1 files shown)


pieterlexis:
it might be worthwhile to fix the default to no networks.

Habbie:
> it might be worthwhile to fix the default to no networks.

If 4.1 has `recursor=`, yes. But it doesn't :)

pieterlexis:
> If 4.1 has `recursor=`, yes. But it doesn't :)

In auth 4.0.5 :)

Habbie:
I'm fine with just the doc note, changing the default might surprise people.

PR #5461, opened 2017-06-26
rec: Add an ECS index to the cache

rgacogne:
### Short description

The existing cache implementation has been not designed for a very large number of ECS-specific entries for a given (qname,qtype) entry. This is fine for most cases, but we can do better if we have several thousands of entries for the same (qname, qtype) and are looking for the longest-prefix match.
This PR adds a NetmaskTree based ECS index to the cache, keeping track of the netmask-specific entries to be able to get the best match quickly. The index is disabled by default and is only enabled once the cache holds netmask-specific entries, so there should not be any penalty for deployment not enabling ECS.

Remaining known issues:
 - ANY queries do not use the index ;
 - removal the cache is a bit slower since we need to update the ECS index, but it should be good enough ;
 - `NetmaskTree::erase()` does not compact the tree, so there is room for improvement here.

### Checklist

I have:
- [x] documented the code

### DIFF
--- a/pdns/cachecleaner.hh
+++ b/pdns/cachecleaner.hh
@@ -23,10 +23,10 @@
 
 #include "lock.hh"
 
-// this function can clean any cache that has a getTTD() method on its entries, and a 'sequence' index as its second index
+// this function can clean any cache that has a getTTD() method on its entries, a preRemoval() method and a 'sequence' index as its second index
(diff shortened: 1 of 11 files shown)


rgacogne:
Painfully rebased to fix a conflict, might use a quick review.

PR #5465, opened 2017-06-26
Use constant-time string comparison for the webserver's passwords

rgacogne:
### Short description

It prevents the (very theoretical) possibility for an attacker with very low-jitter access to the webserver and in one of the `webserver-allow-from` ranges from guessing the password with a timing attack.

### Checklist

I have:
- [ ] documented the code

### DIFF
--- a/pdns/misc.cc
+++ b/pdns/misc.cc
@@ -57,6 +57,7 @@
 #include <sys/types.h>
 #include <pwd.h>
 #include <grp.h>
+
 #ifdef __FreeBSD__
 #  include <pthread_np.h>
 #endif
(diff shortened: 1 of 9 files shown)


pdns/misc.cc
```diff
@@ -1311,3 +1310,28 @@ unsigned int pdns_stou(const std::string& str, size_t * idx, int base)
...
 
+#if HAVE_OPENSSL_LIBCRYPTO
+#include <openssl/opensslv.h>
+#include <openssl/crypto.h>
+#endif
+
+bool constantTimeStringEquals(const std::string& a, const std::string& b)
+{
+  if (a.size() != b.size()) {
+    return false;
```
rgacogne: You are right! That didn't matter until now because this function was only used to compare strings of the same length, and I haven't taken the time to check it before reusing it, my bad. It's a bit different in `C++11` because `std::string::size()` is mandated to be constant-time, but we leak information about the length nonetheless. I guess I'll have to rewrite that function if I ever find the time to do so.

rgacogne:
I'm pondering merging this PR as it is, since it would still improve the current situation. Sure, it would still be possible to guess the password's length but not the actual password. This doesn't matter much if the password is strong enough, and doesn't make it really worse if it's not.
On the other hand the actual implementation can in theory be used to guess the password.

gibson042:
Why not update the PR to use an algorithm that _doesn't_ leak length information?

rgacogne:
Sure, do you have one in mind? I'm pretty sure the one proposed in your earlier link still leaks the length because the compiler will optimize this branch away:

> if (s2[j] == '\0')
            k++;

I have looked at various implementations, including `CRYPTO_memcmp()` in OpenSSL, and they only provide constant-time comparison if the strings are of equal length.

gibson042:
Even if that branch does get optimized away, it's still an improvement over an explicit early exit like the one here.

rgacogne:
Yes, except that we then get different 'guarantees' depending on whether we are built against a recent version of OpenSSL or not, which I don't like. This function also needs to be binary-safe for the existing DNSSEC code. I'll ponder this a little longer.

zeha:
@rgacogne is your last commit in this branch post considerations or is there anything left to do here?

zeha:
Turned into issue #6153.

PR #5466, opened 2017-06-26
API: Clean up auth/recursor code mismatches

### LINKED ISSUE #5398: ws-recursor.cc is missing exception handling
- Program: Recursor
 - Issue type: Bug report

### Short description
The Recursor Webserver API (https://github.com/PowerDNS/pdns/blob/50b9a27c37aa240f13f8c6d3454246e82ff11d9f/pdns/ws-recursor.cc#L584) does not handle all possible HTTP parsing exceptions, resulting in missed log entries and (if you're debugging) a hung debugger.

### Other information
I've been able to identify three types of exception thrown. YaHTTP::ParseError which is handled; and std::out_of_range and char const* c which are not.

The exception handling elsewhere (e.g. https://github.com/PowerDNS/pdns/blob/353c76cb9d6fcbfb9eaca7bdfe9ef28768aecb40/pdns/webserver.cc#L326 ) looks like the correct approach.

zeha:
### Short description
serveConnection in auth caught a lot more exceptions than the recursor code, which then would just hang on an uncaught exception.

Also cleans up a duplicated data.clear().

Fixes #5398.

### Checklist
I have:
- [ ] documented the code

### DIFF
--- a/pdns/ws-recursor.cc
+++ b/pdns/ws-recursor.cc
@@ -571,7 +571,6 @@ void AsyncWebServer::serveConnection(Socket *client)
   string data;
   try {
     while(!req.complete) {
-      data.clear();
       int bytes = arecvtcp(data, 16384, client, true);
       if (bytes > 0) {
         req.complete = yarl.feed(data);
--- a/pdns/ws-recursor.cc
+++ b/pdns/ws-recursor.cc
@@ -560,9 +560,9 @@ void AsyncServer::newConnection()
   getMT()->makeThread(&AsyncServerNewConnectionMT, this);
 }
 
-
+// This is an entry point from FDM, so it needs to catch everything.
 void AsyncWebServer::serveConnection(Socket *client)
-{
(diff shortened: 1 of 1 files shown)


pdns/webserver.hh
```diff
@@ -126,7 +126,12 @@ public:
   ComboAddress d_local;
 
   Socket *accept() {
-    return d_server_socket.accept();
+    Socket* socket = nullptr;
+    // handle EAGAIN
+    while (socket == nullptr) {
+      socket = d_server_socket.accept();
+    }
+    return socket;
```
rgacogne: I'm sure I'm missing something but if the socket is in blocking mode we should never get `EAGAIN`, and if it's in non-blocking mode this looks like we would then enter busy-looping?

zeha: Right. On OS X EAGAIN isn't even in the accept man-page... Removed that commit.

PR #5485, opened 2017-07-03
rec: Don't always override `loglevel` to 6

rgacogne:
### Short description

It looks like we always overrode the `loglevel` value to 6, so change and document the default to 6, but don't override it otherwise.

### Checklist

I have:
- [ ] documented the code


Habbie:
Perhaps add this to the upgrade notes. Otherwise good!

rgacogne:
Rebased, added a few words to the upgrade notes and updated the new docs as well.

PR #5498, opened 2017-07-05
Add support for Botan 2.x

rgacogne:
### Short description

Initial testing (rec on https://rootcanary.org/test.html, `pdnsutil test-algorithms`) indicates that both 2.0.1 and 2.1.0 work fine, but signature is 10 times slower with 2.1.0, apparently due to blinding (callgrind reports a lot of CPU spent in the `RNG`).

### Checklist

I have:
- [ ] documented the code


Habbie:
rootcanary only tests verification, not signing; `pdnsutil test-algorithms` only verifies consistency within the same Botan version. Not saying this is an argument against merging but test vectors would be nice to have in the future.

The only real question I have is whether you tested 1.10 with rootcanary and `pdnsutil` after these changes as well.

rgacogne:
You are right! Don't merge it yet, I'll look into adding some test vectors.

Habbie:
You can steal one half from rootcanary.org :)

rgacogne:
I did test 1.10 after these changes, but I would feel better with test vectors anyway :-)

rgacogne:
BTW, the 2.1.0 performance regression has been confirmed: [1]

[1]: https://github.com/randombit/botan/issues/1019#issuecomment-296414253

rgacogne:
This PR should be rebased after https://github.com/PowerDNS/pdns/pull/5733 has been merged. Note that since Gost signatures are not deterministic, the test vectors are of no real help for testing the signing part.

rgacogne:
Removing the "WIP" part of the title since the tests have been added in #5733.

Habbie:
Given #5797, do you want to rework this PR to drop 1.10, or shall we do that separately?

rgacogne:
This PR now also drops Botan 1.10, thanks to @mind04!

PR #5512, opened 2017-07-11
Auth: drop broken support for packet-specific SOA replies from backends

zeha:
### Short description
Remove DNSPacket pointer from getSOA. This was never okay to use (at least with caching enabled).

Split from #4545 per @Habbie's wish.

### Checklist
I have:
- [ ] documented the code

### DIFF
--- a/modules/luabackend/minimal.cc
+++ b/modules/luabackend/minimal.cc
@@ -176,11 +176,9 @@ bool LUABackend::get(DNSResourceRecord &rr) {
     return !rr.content.empty();
 }
 
-bool LUABackend::getSOA(const DNSName &name, SOAData &soadata, DNSPacket *p) {
+bool LUABackend::getSOA(const DNSName &name, SOAData &soadata) {
     if (logging)
 	L << Logger::Info << backend_name << "(getsoa) BEGIN" << endl;
-
-    dnspacket = p;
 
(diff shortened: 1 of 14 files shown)


modules/luabackend/minimal.cc
```diff
@@ -176,12 +176,12 @@ bool LUABackend::get(DNSResourceRecord &rr) {
     return !rr.content.empty();
 }
 
-bool LUABackend::getSOA(const DNSName &name, SOAData &soadata, DNSPacket *p) {
+bool LUABackend::getSOA(const string &name, SOAData &soadata) {
```
Habbie: `string` ?!

modules/luabackend/minimal.cc
```diff
@@ -176,12 +176,12 @@ bool LUABackend::get(DNSResourceRecord &rr) {
...
+    dnspacket = NULL;
```
Habbie: `LUABackend::dnspacket` can be removed completely now.

zeha:
After this force-push battle... my commit does compile now and also sets luapacket=NULL in LUABackend::getSOA(), which is probably important?

Habbie:
`dnspacket=NULL` in getSOA can go away

mind04:
LGTM

PR #5515, opened 2017-07-12
rec: Fix cache handling of ECS queries with a source length of 0

rgacogne:
### Short description

The current behavior is to send a query without ECS information and to cache it as a non ECS-specific entry. Unfortunately doing so meant that we wouldn't send any further ECS-specific queries for
this `(qname,qtype)` since we could fallback to the non ECS-specific cached answer.

This commit adds a new parameter, `ecs-scope-zero-addr`, allowing to specify the address sent to ECS-whitelisted authoritative servers in that case. If unset, the default is to take the first usable (non-any) address in `query-local-address` then in `query-local-address6`, and finally to use `127.0.0.1` as a last resort.
### Checklist

I have:
- [x] documented the code

### DIFF
--- a/pdns/pdns_recursor.cc
+++ b/pdns/pdns_recursor.cc
@@ -2880,6 +2880,33 @@ static int serviceMain(int argc, char*argv[])
   SyncRes::s_ecsipv4limit = ::arg().asNum("ecs-ipv4-bits");
   SyncRes::s_ecsipv6limit = ::arg().asNum("ecs-ipv6-bits");
 
+  if (!::arg().isEmpty("ecs-scope-zero-address")) {
+    ComboAddress scopeZero(::arg()["ecs-scope-zero-address"]);
+    SyncRes::setECSScopeZeroAddress(Netmask(scopeZero, scopeZero.isIPv4() ? 32 : 128));
+  }
+  else {
+    bool found = false;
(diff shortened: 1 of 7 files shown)


pdns/pdns_recursor.cc
```diff
@@ -2812,6 +2812,33 @@ static int serviceMain(int argc, char*argv[])
...
+    for (const auto& addr : g_localQueryAddresses4) {
+      if (!IsAnyAddress(addr)) {
+        SyncRes::setECSScopeZeroAddress(Netmask(addr, 32));
+        found = true;
+        break;
+      }
+    }
+    if (!found) {
+      for (const auto& addr : g_localQueryAddresses6) {
+        if (!IsAnyAddress(addr)) {
```
zeha: what about link local addresses and other snowflakes?

rgacogne:
Updated to the new docs format.

pdns/pdns_recursor.cc
rgacogne: I agree this might be an issue, however this check is only intended to catch obvious configuration error since there is no way we can detect every one of them. My feeling is that the administrator setting `query-local-address` and/or `query-local-address6` to something like a link-local address and adding entries to the ECS whitelist should also set `ecs-scope-zero-addr`. I might be wrong :)

PR #5519, opened 2017-07-13
lookups one level (or more) below apex did confuse getAuth() for qytpe DS

mind04:
### Short description
lookups one level (or more) below apex did confuse getAuth() for qytpe DS

### Checklist
I have:
- [ ] documented the code

### DIFF
--- a/pdns/ueberbackend.cc
+++ b/pdns/ueberbackend.cc
@@ -337,7 +337,6 @@ bool UeberBackend::getAuth(const DNSName &target, const QType& qtype, SOAData* s
         DNSZoneRecord rr;
         rr.dr.d_name = sd->qname;
         rr.dr.d_type = QType::SOA;
-        
         rr.dr.d_content = makeSOAContent(*sd);
         rr.dr.d_ttl = sd->ttl;
         rr.domain_id = sd->domain_id;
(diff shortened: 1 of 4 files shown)


klaus3000:
I still think this fix is incomplete (only 90% sure as this logic is not easy to reverse engineer and the documentation of the code is sparse).

Consider the scenario where there is only 1 backend and this backend immediately responds with the best match. E.g. target="a.b.c.example.com" and the backend responses with "com".

In this case, shorter=target=a.b.c.example.com. If the query is for DS, PowerDNS would still "chasing next" as the if condition "target != shorter" is wrong. The proper comparison would be target against sd->qname (compare with the zone returned from the backend, not with the zone the backend was asked).

Further, in case of DS queries, I think getAuth() should return false if there is no parent zone found, instead of using the zone which is identical to the queried label:

Hence, I think the proper fix would be:

```diff
found:
+    // DS records must be answered from the parent zone (the zone
+    // which has the delegation). Hence, if we found Authority, we
+    // start again and chase the next zone.
+    // There is one exception: if the first found Authority is not
+    // identical to the queried target, we already have the parent
+    // zone found.
-    if(found == (qtype == QType::DS) || target != shorter) {
+    if(found == (qtype == QType::DS) || target != sd->qname) {
      DLOG(L<<Logger::Error<<"found: "<<sd->qname<<endl);
      return true;
    } else {
      DLOG(L<<Logger::Error<<"chasing next: "<<sd->qname<<endl);
      found = true;
    }

  } while(shorter.chopOff());
+  // DS queries must be answered from the parent zone. If we do not
+  // find a parent zone, return false
+  if (found && (qtype == QType::DS) && (target == sd->qname)) {
+    return false;
+  }
  return found;
}
```

Habbie:
Hello,

comments on closed tickets tend to get forgotten. If you believe more work needs to be done, can you please open a ticket?

klaus3000:
Done: #5697

PR #5543, opened 2017-07-19
Wrap the webserver's and Resolver::tryGetSOASerial objects into smart pointers

rgacogne:
### Short description

There was no issue AFAICT, but it makes it less likely to mess up in the future and leak memory, or trigger a use-after-free.

### Checklist

I have:
- [ ] documented the code


zeha:
👍  It looks like std::thread can take more than one argument, so the entire connectionThreadData struct could go away.

rgacogne:
Oh, I didn't think about that! Thanks for the suggestion, I'll amend this PR!

rgacogne:
Removed `connectionThreadData`, thanks again!

PR #5551, opened 2017-07-25
Turn on -fstack-check when supported.

cmouse:
### Short description
Protects against stack collision attacks.

### Checklist
I have:
- [ ] documented the code

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -62,6 +62,7 @@ AC_MSG_RESULT([$enable_hardening])
 AS_IF([test "x$enable_hardening" != "xno"], [
   AC_CC_PIE
   AC_CC_STACK_PROTECTOR
+  AC_CC_F_STACK_CHECK
   AC_CC_PARAM_SSP_BUFFER_SIZE([4])
   AC_CC_D_FORTIFY_SOURCE
   AC_LD_RELRO
--- a/m4/pdns_stack_check.m4
+++ b/m4/pdns_stack_check.m4
@@ -1,0 +1,31 @@
+dnl
+dnl Check for support -fstack-check
+dnl
+dnl This file is part of PowerDNS or dnsdist.
+dnl Copyright -- PowerDNS.COM B.V. and its contributors
+dnl
+dnl This program is free software; you can redistribute it and/or modify
+dnl it under the terms of version 2 of the GNU General Public License as
+dnl published by the Free Software Foundation.
+dnl
+dnl In addition, for the avoidance of any doubt, permission is granted to
+dnl link this program with OpenSSL and to (re)distribute the binaries
+dnl produced as the result of such linking.
+dnl
+dnl This program is distributed in the hope that it will be useful,
+dnl but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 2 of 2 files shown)


cmouse:
See https://www.qualys.com/2017/06/19/stack-clash/stack-clash.txt

setharnold:
I don't believe the current `-fstack-check` is suitable for this use. To quote from some sadly still-private emails:

```
We are still looking at this, but the -fstack-check flag is intended for
use with Ada, SJLJ-style exception handling, and a signal handler
(without sigaltstack) with calls sigsetjmp to recover from stack
overflow.  The -fstack-check flag also changes how the stack frame is
laid out, and we'd want to avoid that for performance reasons.

The need for running a SIGSEGV handler on the same stack explains why
the initial probe offset is so far away from the current stack pointer:
It's not based on the size of the guard page (it's larger than that on
x86-64), but on the stack space requirement of the SIGSEGV handler at
the time the patch was written.  (Register files get bigger all the
time, so the amount of data pushed by the kernel increases over time and
is now around 3 KiB with AVX-512F, and more with aarch64 SVE.)

[...]

We also do not use SJLJ-style exceptions, but DWARF-based ones, and the
code generated by -fstack-check appears not to be fully compatible with
asynchronous stack unwinding (there appear to be some missing CFI
annotations).  This could introduce sporadic crashes or hangs with
thread cancellation because stack unwinding does not work.
```

A later mail:

```
Also note that the existing -fstack-check code was written for Ada's
need.  It probes way below the current stack pointer, and on x86-64,
it's even more than a page.  The expectation is that all code is
compiled this way, and the Ada run time installs a signal handler which
raises an exception from the SIGSEGV handler once the guard page is hit,
and the application can catch that after unwinding.  This isn't what we
want because it can skip the guard page (unless everything is compiled
with -fstack-check) and has a substantial performance impact because
most functions need probes (not only those with large frames or
allocas).  The current implementation also forces the use of a frame
pointer.
```

The patches I saw looked like they were going to add new GCC command line options:

```
+@item -fstack-check-probe=@var{type}
+The @option{-fstack-check} option emits probes of stack locations to help
+detect stack overflows.  This option controls the strategy for when to
+emit those probes.
+       
+@option{-fstack-check-probe=ahead} emits probes ahead of need.  This is
+designed to allow a signal to be delivered and handled by ensuring there
+is at least one free stack page still available when the stack guard is
+encountered.  This option assumes all callers in the call stack are
+compiled with stack checking.
+
+@option{-fstack-check-probe=as-needed} emits probes as stack space is
+needed and allocated.  It does not ensure there is enough stack space
+to deliver and handle a signal, but it also does not assume anything about
+other calls in the call stack.
```

I suggest holding off until the GCC community has a command line option they're happy with.

Thanks

rgacogne:
> I suggest holding off until the GCC community has a command line option they're happy with.

FWIW, Arch is in the process of changing its default compilation flags, and reached the same conclusion about `-fstack-check` so far [1].

[1]: https://lists.archlinux.org/pipermail/arch-general/2017-July/043924.html

Habbie:
Closing this, we can revisit when gcc fixes this.

PR #5571, opened 2017-08-01
rec: Accept non-auth NS records as a zone cut hint

rgacogne:
### Short description

Some very popular auth servers are buggy and return an empty or REFUSED answer for the NS of zones they are authoritative for. Yeah.
Work around a validation issue that arises when we hit some "legacy" servers at fastly that are still in use, for example.

### Checklist

I have:
- [ ] documented the code


rgacogne:
Closing for now as it only fixes the REFUSED case.

PR #5579, opened 2017-08-03
dnsdist: Don't use square brackets for IPv6 in Carbon metrics

### LINKED ISSUE #5538: IPv6 addresses improperly encoded in carbon metrics
- Program: dnsdist
 - Issue type: Bug report

### Short description
When logging to carbon, IPv6 addresses are enclosed in square brackets ([]) within metric names. Square brackets have special meaning to many storage engines and frontends.

### Environment
 - Software version: dnsdist 1.1.0
 - Software source: repo

### Steps to reproduce
1. Configure dnsdist with an IPv6 interface and logging to a carbon server
2. Observe the metric names within your carbon instance
3. IPv4 addresses are encoded as `10_0_0_1:53_tcp`, IPv6 addresses are encoded as `[2001:db8::1]:53_tcp`

### Expected behaviour
A recommended encoding would be `2001:db8::1_53_tcp`. It is really a toss-up on the colons becoming ambiguous or breaking with the v4 encoding.

rgacogne:
### Short description

Use `2001:db8::1:53_tcp` instead of `[2001:db8::1]:53_tcp`, as square brackets have special meaning for some Carbon storage engines and backends. I went with the `:port` instead of `_port` as suggested in the initial report to be consistent with the current IPv4 encoding. Since the port is always present it shouldn't be ambiguous, but I'm of course open to switch to another encoding. 

Fixes #5538.

### Checklist

I have:
- [ ] documented the code

### DIFF
--- a/pdns/dnsdist-carbon.cc
+++ b/pdns/dnsdist-carbon.cc
@@ -84,7 +84,7 @@ try
         }
         const auto states = g_dstates.getCopy();
         for(const auto& state : states) {
-          string serverName = state->getName();
+          string serverName = state->name.empty() ? (state->remote.toString() + ":" + std::to_string(state->remote.getPort())) : state->getName();
           boost::replace_all(serverName, ".", "_");
           const string base = "dnsdist." + hostname + ".main.servers." + serverName + ".";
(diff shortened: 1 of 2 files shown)


pieterlexis:
lgtm!

PR #5580, opened 2017-08-04
dnsdist: Copy the DNS header before encrypting it in place

### LINKED ISSUE #5566: dnsdist: grepq returns garbage for DNSCrypt-encrypted responses
- Program: dnsdist
 - Issue type: Bug report

### Short description
As reported by @bjoe2k4 on IRC (thanks), `grepq()` returns strange results for DNSCrypt-encrypted responses. It looks like the pointer we use to insert the DNS header into our response ring buffer points the DNSCrypt response, and not to the actual response from the backend anymore, meaning we would be reading the resolver magic and part of the nonce instead of the actual DNS header.

rgacogne:
### Short description

When DNSCrypt is enabled, we encrypt the answer in place, but we need to keep a copy of the DNS header of the original answer to be able to populate the responses ring buffer.

Fixes #5566.

### Checklist

I have:
- [ ] documented the code

### DIFF
--- a/pdns/dnsdist-tcp.cc
+++ b/pdns/dnsdist-tcp.cc
@@ -233,6 +233,11 @@ void* tcpClientThread(int pipefd)
 #ifdef HAVE_PROTOBUF
   boost::uuids::random_generator uuidGenerator;
 #endif
+#ifdef HAVE_DNSCRYPT
+  /* when the answer is encrypted in place, we need to get a copy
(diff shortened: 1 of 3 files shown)


bjoe2k4:
Works for me. No Err#14 anymore. Thanks!

rgacogne:
@bjoe2k4 Thank you for reporting, testing and confirming!

Issue #4552, opened 2016-10-08
Lua DNSName equality should support ==


Habbie:
Because remembering `:equal` is awkward.

Habbie:
related: https://github.com/tomaka/luawrapper/issues/31

tcely:
Is this resolved now with the recent changes?

zeha:
Not so sure; a quick modification of the recursor tests to use `dq.qname == 'www5.example.net'` instead of `dq.qname:equal('www5.example.net')` doesn't seem to work...

PR #5582, opened 2017-08-06
Set DNSName metatable __eq function

tcely:
I believe this or something like it resolves issue #4552.

### Checklist

I have:
- [ ] documented the code

### DIFF
--- a/pdns/lua-recursor4.cc
+++ b/pdns/lua-recursor4.cc
@@ -249,7 +249,7 @@ RecursorLua4::RecursorLua4(const std::string& fname)
   );
   //d_lw->registerFunction("__eq", &DNSName::operator==);
   d_lw->writeVariable("DNSName", DNSName{});
-  d_lw->writeFunction<bool(DNSName::*)(DNSName::*)>(
+  d_lw->writeFunction(
     "DNSName", LuaContext::Metatable, "__eq",
     [](const DNSName& lhs, const DNSName& rhs) {
       return lhs==rhs;
--- a/pdns/lua-recursor4.cc
+++ b/pdns/lua-recursor4.cc
@@ -274,8 +274,7 @@ RecursorLua4::RecursorLua4(const std::string& fname)
     }
   );
   d_lw->executeCode("DNSName = { ud = DNSName, }");
-  d_lw->executeCode("DNSName.__call = function(self, a) local mt = getmetatable(self) local n = mt.ud(a or '') local nmt = getmetatable(n) local udmt = getmetatable(mt.u
-) nmt.__eq = udmt.__eq nmt.__tostring = udmt.__tostring return n end");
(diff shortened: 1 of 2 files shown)


Habbie:
@cmouse

cmouse:
Can I integrate this merge into #5390 or should it be merged separately? Also I think we should leave "__eq" function as well as some ppl might be using it?

Habbie:
Well, do we even need this one if we merge #5390? It's unclear to me.

cmouse:
Yes. It's still needed, #5390 only moves code around and adds stuff. I'll try and see if this works as advertized as it seems much cleaner way to implement `__eq` than the one I made for LuaWrapper.

cmouse:
This does not appear to work as expected, comparing two instances of DNSName do not actually invoke the lambda function.

tcely:
The way luawrapper constructs new metatables for each userdata type makes this slightly more painful than I had hoped, but wrapping this in a lua class that adds the missing functions to the newly created userdata instances allowed this to be mostly hidden from the lua script.

cmouse:
Unfortunately this still does not work as it should. When comparing against a C++ provided DNSName it invokes the __call method, but does not result in correct comparison. 

Tested using
```.cc
BOOST_AUTO_TEST_CASE(test_prequery) {
  const std::string script =
"function prequery(q)\n"
"  print(\"prequery\")\n"
"  if q.qdomain == DNSName(\"mod.unit.test.\")\n"
"  then\n"
"    print(\"true branch\")\n"
"    return true\n"
"  end\n"
"  print(\"false branch\")\n"
"  return false\n"
"end";
  AuthLua4 lua;
  DNSPacket *p = new DNSPacket(true);
  p->qdomain = DNSName("mod.unit.test.");
  lua.loadString(script);
  DNSPacket *r;
  try {
    r = lua.prequery(p);
    if (r != NULL)
      BOOST_CHECK_EQUAL(r->qdomain.toString(), "mod.unit.test.");
  } catch (const LuaContext::ExecutionErrorException& e) {
    try {
     std::rethrow_if_nested(e);
    } catch(const std::exception& exp) {
     theL()<<"Extra info: "<<exp.what();
    }
  }
  delete r;
  delete p;
}
```

tcely:
You will need to wrap the raw userdata like this:

`if DNSName(q.qdomain) == DNSName('mod.unit.test.')`

It's not perfect integration between lua and C++ but the syntax is friendlier than before.

cmouse:
Personally I think that https://github.com/ahupowerdns/luawrapper/pull/35 makes more sense now.

Habbie:
@cmouse @tcely is this still useful since we merged similar work from @cmouse ?

tcely:
At a first glance, I'm going with no.

A few concepts might be pulled out of this for the current code. It looks like LuaContext is often overwriting the metatable which makes persisting the functions more difficult. If the current syntax is working and good enough I don't see a reason to maintain the wrapped objects.

Habbie:
Thanks, then I'm closing this.

Issue #2835, opened 2015-10-27
listen on 127.0.0.1 during auth tests


Habbie:
Right now our auth regression tests (via `start-test-stop`) listen on 0.0.0.0 (as is default for pdns). This causes firewall popups on OSX. If we switch to 127.0.0.1 the popup goes away, and presumably the tests will still work.

aerique:
For IPv6 we still listen on ::, see also the comments for #5583 on this.

PR #5583, opened 2017-08-06
listen on 127.0.0.1 during regression tests

tcely:
### Short description
Added an address variable to regression-tests set to 127.0.0.1

Issue #2835 

### Checklist

I have:
- [ ] documented the code

### DIFF
--- a/regression-tests/backends/bind-master
+++ b/regression-tests/backends/bind-master
@@ -7,7 +7,7 @@ bind-config=./named.conf
 bind-ignore-broken-records=yes
 __EOF__
 
-		$RUNWRAPPER $PDNS --daemon=no --local-port=$port --config-dir=. \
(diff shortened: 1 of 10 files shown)


Habbie:
Do you think you can add an easy way to override this? An environment variable from outside perhaps?

Habbie:
Hmm, this does not actually address #2835 as for v6 we still listen on :: with this.

tcely:
I didn't see IPv6 mentioned in that issue, but we could always include `--local-ipv6=::1` if that would be desirable.

At this point though, I'd suggest a common flags file that these backend files all include instead of environment variables.

Habbie:
No, I only noticed v6 being an issue when I tested your patch. I agree on all this duplication being bad. I'm happy merging this as is and updating the ticket so somebody can go deeper when they have time.

PR #5594, opened 2017-08-11
[WIP] Add experimental XPF support

rgacogne:
### Short description

Don't merge this while it's still using a private code point!

This PR adds experimental support for `XPF`[1] to the recursor and dnsdist. It adds a new option to dnsdist to send the source and destination IPs and ports to the recursor via a `XPF` record.
It also adds a new option to the recursor to whitelist the source IPs allowed to send `XPF` records. Those records are then decoded before the `gettag` hook and the source and destination IPs are replaced by the one provided.
The PR is based on and contains @Habbie's work to add `XPF` support to `sdig` and the authoritative server's `bind` backend.

[1]: https://tools.ietf.org/html/draft-bellis-dnsop-xpf-01

### Checklist

I have:
- [x] documented the code


pdns/xpf.hh
```diff
@@ -0,0 +1,30 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+#ifndef PDNS_XPF_HH
```
pieterlexis: `#pragma once`?

rgacogne:
Rebased on master, and made the XPF code point configurable in dnsdist and the rec.

pdns/dnsparser.cc
```diff
@@ -295,20 +296,29 @@ void MOADNSParser::init(bool query, const char *packet, unsigned int len)
 
       d_answers.push_back(make_pair(dr, pr.d_pos));
 
+      /* XXX: XPF records should be allowed after TSIG as soon as the actual XPF option code has been assigned:
```
Habbie: Why would assignment change this rule?

pdns/xpf.cc
```diff
@@ -0,0 +1,120 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+
+#include "xpf.hh"
+
+std::string generateXPFPayload(bool tcp, const ComboAddress& remote, const ComboAddress& local)
```
Habbie: Maybe rename remote/local to source/destination? That mirrors the XPF spec text and seems less confusing to me.

pdns/qtype.hh
```diff
@@ -125,7 +125,8 @@ public:
     CAA=257,
     DLV=32769,
     ADDR=65400,
-    ALIAS=65401
+    ALIAS=65401,
+    XPF=65422
```
Habbie: Remove this?

pdns/qtype.hh
```diff
@@ -217,6 +218,7 @@ private:
       qtype_insert("DLV", 32769);
       qtype_insert("ADDR", 65400);
       qtype_insert("ALIAS", 65401);
+      qtype_insert("XPF", 65422);
```
Habbie: Remove this?

pdns/qtype.hh
Habbie: Hmm, I guess that would break `mastermake`. Now I see why fixing `sdig` is harder.

pdns/dnsparser.cc
rgacogne: Because this function is used in the auth, rec and dnsdist so using a variable is painful, while of course not impossible. This becomes very easy once we have a fixed value to allow.

pdns/recursordist/docs/settings.rst
```diff
@@ -1144,3 +1144,27 @@ TCP port where the webserver should listen on.
...
+.. versionadded:: 4.1.0
+
+-  IP ranges, separated by commas
+-  Default: empty
+
+This is an experimental implementation of `draft-bellis-dnsop-xpf`.
+The server will trust XPF records found in queries sent from those netmasks (both IPv4 and IPv6),
+and will adjust queries' source and destination accordingly. This is especially useful when the recursor
+is placed behind a proxy like dnsdist.
+Note that the `allow-from`_ setting is still applied to the original source address, and thus access restriction
```
Habbie: 'original source address' is a bit confusing, but I'm struggling to find better words.

pdns/xpf.cc
rgacogne: Looks better indeed, I'll do it after your changes have landed on this branch.

pdns/recursordist/docs/settings.rst
hlindqvist: "IP header source address", possibly? (If that is the correct interpretation)

The note removes what seems like probably the single most obvious application for this. Maybe name some example(s) where it can be used?

PR #5595, opened 2017-08-11
Improve slave-check: Check every configured master and AXFR from mast…

klaus3000:
…er with highest serial

Currently PDNS as slave has a poor logic when there are multiple masters:

1. On incoming NOTIFY, PDNS queues the zone for refresh. Then, the SOA query uses a
random master to get the master's serial. On timeout, PDNS does not try another master.

2. If the SOA query was successful and the serial increased, PDNS queues the domain for
AXFR, but this AXFR uses only the first master.

Thus, if the first master is offline, even if the second master sends
NOTIFYs and has a newer zone, PDNS will never transfer it.

This patch does 2 things:

a) If a slave check for a domain is requested and the domain has multiple masters, the SOA
check will be performed against every master. Further, the name server which answered with
the highest serial will be used for the AXFR.

Known issue: If there is a serial rollover, then choosing the highest serial is not the best
choice. But as soon as the rollover happend on every master the logic is correct again.

### DIFF
--- a/pdns/slavecommunicator.cc
+++ b/pdns/slavecommunicator.cc
@@ -244,7 +244,7 @@ static vector<DNSResourceRecord> doAxfr(const ComboAddress& raddr, const DNSName
   bool soa_received {false};
   while(retriever.getChunk(recs)) {
     if(first) {
-      L<<Logger::Error<<"AXFR started for '"<<domain<<"'"<<endl;
+      L<<Logger::Error<<"AXFR started for '"<<domain<<"' from remote "<<raddr.toStringWithPort()<<endl;
       first=false;
     }
 
@@ -650,6 +650,7 @@ struct SlaveSenderReceiver
     uint32_t theirSerial;
     uint32_t theirInception;
     uint32_t theirExpire;
+    string master;
   };
 
   map<uint32_t, Answer> d_freshness;
@@ -664,19 +665,18 @@ struct SlaveSenderReceiver
 
   Identifier send(DomainNotificationInfo& dni)
   {
-    random_shuffle(dni.di.masters.begin(), dni.di.masters.end());
     try {
-      ComboAddress remote(*dni.di.masters.begin());
+      L<<Logger::Info<<"Sending SOA request for '"<<dni.di.zone<<"' to "<<dni.di.masters.back()<<endl;
       if (dni.localaddr.sin4.sin_family == 0) {
         return make_pair(dni.di.zone,
-          d_resolver.sendResolve(ComboAddress(*dni.di.masters.begin(), 53),
+          d_resolver.sendResolve(ComboAddress(dni.di.masters.back(), 53),
             dni.di.zone,
             QType::SOA,
             dni.dnssecOk, dni.tsigkeyname, dni.tsigalgname, dni.tsigsecret)
         );
       } else {
         return make_pair(dni.di.zone,
-          d_resolver.sendResolve(ComboAddress(*dni.di.masters.begin(), 53), dni.localaddr,
+          d_resolver.sendResolve(ComboAddress(dni.di.masters.back(), 53), dni.localaddr,
             dni.di.zone,
             QType::SOA,
             dni.dnssecOk, dni.tsigkeyname, dni.tsigalgname, dni.tsigsecret)
@@ -691,14 +691,30 @@ struct SlaveSenderReceiver
   bool receive(Identifier& id, Answer& a)
   {
     if(d_resolver.tryGetSOASerial(&id.first, &a.theirSerial, &a.theirInception, &a.theirExpire, &id.second)) {
+      L<<Logger::Debug<<"Got response for zone '"<<id.first<<"' with serial "<<a.theirSerial<<endl;
       return 1;
     }
     return 0;
   }
 
-  void deliverAnswer(DomainNotificationInfo& dni, const Answer& a, unsigned int usec)
+  void deliverAnswer(DomainNotificationInfo& dni, Answer& a, unsigned int usec)
   {
-    d_freshness[dni.di.id]=a;
+    L<<Logger::Info<<"Received serial "<<a.theirSerial<<" for zone '"<<dni.di.zone<<"' in "<<usec<<" usec from master "<<dni.di.masters.back()<<endl;
+    a.master = dni.di.masters.back();
+    // Check if we already got an answer for this domain
+    if (d_freshness.count(dni.di.id)) {
+      // Note: This check may be incorrect if one of the masters had a serial overflow,
+      // but it will recover once all masters have wrapped their serial.
+      if (a.theirSerial > d_freshness[dni.di.id].theirSerial) {
+        L<<Logger::Debug<<"Received serial "<<a.theirSerial<<" for zone '"<<dni.di.zone<<"' is bigger than previous serial "<<d_freshness[dni.di.id].theirSerial<<endl;
+        d_freshness[dni.di.id]=a;
+      } else {
+        L<<Logger::Debug<<"Received serial "<<a.theirSerial<<" for zone '"<<dni.di.zone<<"' is equal or lower than previous serial "<<d_freshness[dni.di.id].theirSerial<<endl;
+      }
+    } else {
+      L<<Logger::Debug<<"Received serial "<<a.theirSerial<<" for zone '"<<dni.di.zone<<"' is first answer"<<endl;
+      d_freshness[dni.di.id]=a;
+    }
   }
 
   Resolver d_resolver;
@@ -730,6 +746,7 @@ void CommunicatorClass::slaveRefresh(PacketHandler *P)
   vector<DomainInfo> rdomains;
   vector<DomainNotificationInfo> sdomains; 
   vector<DNSPacket> trysuperdomains;
+  uint64_t numcheckdomains=0;
 
   {
     Lock l(&d_lock);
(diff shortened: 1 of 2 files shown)


pdns/slavecommunicator.cc
```diff
@@ -650,6 +650,7 @@ struct SlaveSenderReceiver
     uint32_t theirSerial;
     uint32_t theirInception;
     uint32_t theirExpire;
+    string master;
```
zeha: This should probably be a ComboAddress, if that integrates somewhat well with the rest (I know SlaveCommunicator likes casting between string and CA a lot...)

klaus3000: The "master" variable is used to hold data from DomainInfo.masters which is:

```
struct DomainInfo
{
...
  vector<string> masters;
```

So I just used the same data type.

zeha:
@klaus3000 This branch has conflicts that must be resolved :(

aerique:
@zeha @Habbie @rgacogne Is this PR still applicable?

rgacogne:
I have no idea.

zeha:
Its a new feature that some people want. Don’t think anything equivalent has been merged.

klaus3000:
The problem still exists but this patch is outdated. I have ported the patch to trunk but have not uploaded it as I want to test it a bit more. Maybe next week. If you want to add it to 4.2 I can put in some extra work hours ;)

Habbie:
I have not reviewed the code, but I agree we could do better in this area. I'd be open to reviewing an updated PR for 4.3 or 4.4.

klaus3000:
Good to know. I try to find some time to rebase/improve the patch. I think I will split the patch into 2 parts:

1. When there are multiple masters, do XFR from the one which was used for the serial check. (I consider this as an bugfix)

Then, there is space for improvement - not bugfixing. For example, Bind as Slave tries one masters, and if the serial check fails, it will retry with the other master. IMO this is complex to implement and actually not necessary, as during the next slave check PDNS again makes a random selection and soe with time also tries the other masters as well.

2. For scenaries where the masters are not synchron (ie. our customers somtimes fail to syncronize their masters and open plenty of support tickets) I would implement the option "check-every-master"  (default: no) were always all the masters are checked and XFR is done from the one with the highest serial.

Feedback?

Habbie:
> When there are multiple masters, do XFR from the one which was used for the serial check. (I consider this as an bugfix)

The serial check, or the one that sent the NOTIFY?

> For scenaries where the masters are not synchron (ie. our customers somtimes fail to syncronize their masters and open plenty of support tickets) I would implement the option "check-every-master" (default: no) were always all the masters are checked and XFR is done from the one with the highest serial.

Yes, i agree this would be nice, and I also agree it would be good to have separate from (1)

klaus3000:
> > When there are multiple masters, do XFR from the one which was used for the serial check. (I consider this as an bugfix)
> 
> The serial check, or the one that sent the NOTIFY?

Currently there is no connection from NOTIFY to SOA-Check. Hence, currently the SOA check always randomly selects a master, regardless if triggered by a NOTIFY or by expired refresh.

It would be smart, in case of NOTIFY to check against the NOTIFY source, but this is much more work, especially as the NOTIFY source may be an IP address which is not configured as master - then the master must be selected randomly anyway.

So to keep it simple, I would not touch how SOA checks are triggered, but only give the checked master IP to the XFR worker.

> > For scenaries where the masters are not synchron (ie. our customers somtimes fail to syncronize their masters and open plenty of support tickets) I would implement the option "check-every-master" (default: no) were always all the masters are checked and XFR is done from the one with the highest serial.
> 
> Yes, i agree this would be nice, and I also agree it would be good to have separate from (1)

Perfect.

pdns/slavecommunicator.cc
```diff
@@ -756,6 +773,7 @@ void CommunicatorClass::slaveRefresh(PacketHandler *P)
   }
   if(rdomains.empty()) { // if we have priority domains, check them first
     B->getUnfreshSlaveInfos(&rdomains);
+    L<<Logger::Info<<"No domains explicitely queued for freshness checks. Checking "<<rdomains.size()<<" unfresh slave zones reported by backend"<<endl;
```
jsoref, line 776: https://github.com/jsoref/pdns/commit/2b023c9666b3298dfc6caaa7503c63fb45218a4c#commitcomment-39230791
```suggestion
    L<<Logger::Info<<"No domains explicitly queued for freshness checks. Checking "<<rdomains.size()<<" unfresh slave zones reported by backend"<<endl;
```

pdns/slavecommunicator.cc
```diff
@@ -730,6 +746,7 @@ void CommunicatorClass::slaveRefresh(PacketHandler *P)
   vector<DomainInfo> rdomains;
   vector<DomainNotificationInfo> sdomains; 
   vector<DNSPacket> trysuperdomains;
+  uint64_t numcheckdomains=0;
```
jsoref, line 749:
```suggestion
  uint64_t numCheckDomains=0;
```

pdns/slavecommunicator.cc
```diff
@@ -806,7 +825,14 @@ void CommunicatorClass::slaveRefresh(PacketHandler *P)
         dni.localaddr.sin4.sin_family = 0;
       }
 
-      sdomains.push_back(dni);
+      numcheckdomains++;
```
jsoref, line 828:
```suggestion
      numCheckDomains++;
```

pdns/slavecommunicator.cc
```diff
@@ -820,9 +846,9 @@ void CommunicatorClass::slaveRefresh(PacketHandler *P)
   }
   else {
     Lock l(&d_lock);
-    L<<Logger::Warning<<sdomains.size()<<" slave domain"<<(sdomains.size()>1 ? "s" : "")<<" need"<<
-      (sdomains.size()>1 ? "" : "s")<<
-      " checking, "<<d_suckdomains.size()<<" queued for AXFR"<<endl;
+    L<<Logger::Warning<<numcheckdomains<<" slave domain"<<(numcheckdomains>1 ? "s" : "")<<" need"<<
+      (numcheckdomains>1 ? "" : "s")<<
```
jsoref, line 850:
```suggestion
    L<<Logger::Warning<<numCheckDomains<<" slave domain"<<(numCheckDomains>1 ? "s" : "")<<" need"<<
      (numCheckDomains>1 ? "" : "s")<<
```

pdns/slavecommunicator.cc
```diff
@@ -831,19 +857,25 @@ void CommunicatorClass::slaveRefresh(PacketHandler *P)
 
   ifl.d_maxInFlight = 200;
 
+  uint64_t stdexceptions=0, pdnsexceptions=0;
```
jsoref, line 860:
```suggestion
  uint64_t stdExceptions=0, pdnsExceptions=0;
```

pdns/slavecommunicator.cc
```diff
@@ -831,19 +857,25 @@ void CommunicatorClass::slaveRefresh(PacketHandler *P)
...
   for(;;) {
     try {
       ifl.run();
       break;
     }
     catch(std::exception& e) {
       L<<Logger::Error<<"While checking domain freshness: " << e.what()<<endl;
+      stdexceptions++;
```
jsoref, line 868:
```suggestion
      stdExceptions++;
```

pdns/slavecommunicator.cc
```diff
@@ -831,19 +857,25 @@ void CommunicatorClass::slaveRefresh(PacketHandler *P)
...
     }
     catch(PDNSException &re) {
       L<<Logger::Error<<"While checking domain freshness: " << re.reason<<endl;
+      pdnsexceptions++;
```
jsoref, line 872:
```suggestion
      pdnsExceptions++;
```

pdns/slavecommunicator.cc
```diff
@@ -831,19 +857,25 @@ void CommunicatorClass::slaveRefresh(PacketHandler *P)
...
     }
   }
-  L<<Logger::Warning<<"Received serial number updates for "<<ssr.d_freshness.size()<<" zone"<<addS(ssr.d_freshness.size())<<", had "<<ifl.getTimeouts()<<" timeout"<<addS(ifl.getTimeouts())<<endl;
+  L<<Logger::Warning<<"Received serial number updates for "<<ssr.d_freshness.size()<<" zone"<<addS(ssr.d_freshness.size())
+    <<", had "<<ifl.getTimeouts()<<" timeout"<<addS(ifl.getTimeouts())
+    <<", "<<stdexceptions<<" std-exception"<<addS(stdexceptions)<<" and "
+    <<pdnsexceptions<<" pdns-exception"<<addS(pdnsexceptions)<<endl;
```
jsoref, line 878:
```suggestion
    <<", "<<stdExceptions<<" std-exception"<<addS(stdExceptions)<<" and "
    <<pdnsExceptions<<" pdns-exception"<<addS(pdnsExceptions)<<endl;
```

zeha:
> Currently there is no connection from NOTIFY to SOA-Check. Hence, currently the SOA check always randomly selects a master, regardless if triggered by a NOTIFY or by expired refresh.
> 
> It would be smart, in case of NOTIFY to check against the NOTIFY source, but this is much more work, especially as the NOTIFY source may be an IP address which is not configured as master - then the master must be selected randomly anyway.
> 
> So to keep it simple, I would not touch how SOA checks are triggered, but only give the checked master IP to the XFR worker.

This was actually done in #5361 - pdns now uses the same master for SOA check and AXFR, and if the check source was a NOTIFY, it uses the master that sent the NOTIFY.

I've added the master address to the SOA check result logs in #9265.

Habbie:
> This was actually done in #5361 - - pdns now uses the same master for SOA check and AXFR, and if the check source was a NOTIFY, it uses the master that sent the NOTIFY.

I believe that solves most of the problem, so I'm closing this PR.

PR #5609, opened 2017-08-14
DNSPacket naming was confusing + small cleanups

ahupowerdns:
rename DNSPacket::parse() to questionparse() & add comment it does not parse answers. Remove needless parsing from dnsproxy.cc, make DNSPacket bool constructor explicit. ALIAS still works after these changes.

### Checklist

I have:
- [ ] documented the code


rgacogne:
I'm a bit worried that `DNSPacket::questionparse()` might also be confusing, because it does call `MOADNSParser` that parses the answers if `d_isQuery` is false, unless I'm mistaken?

ahupowerdns:
On second thought let's not do this. It does not clarify as much as you'd think. DNSPacket is a strange beast.

Issue #5357, opened 2017-05-31
SortList doesn't respect CNAME and A records ordering


killerwhile:
- Program: Recursor
 - Issue type: Bug report

### Short description

When using the Lua sortList feature, responses with CNAME and A records do not respect  RFC1034 section 4.3.1 (A record last RR). https://github.com/PowerDNS/pdns/issues/4339 and https://github.com/PowerDNS/pdns/pull/4340 reported similar behaviour but it seems to still happen when sortList is used.

### Environment

 - Operating system: Centos7
 - Software version: 4.0.4-1
 - Software source: PowerDNS repository

### Steps to reproduce

1. Create a file `/etc/pdns-recursor/sortList-test.lua` with the following content:
```
-- www.noisette.ch: return 87.98.177.202/32 as the first ip in any case.
addSortList("0.0.0.0/0", {"87.98.177.202/32"})
```

2. Add the following line to your config
```
lua-config-file=/etc/pdns-recursor/sortList-test.lua
```

3. ensure your `/etc/resolv.conf` query the right server

4. Run `getent hosts www.noisette.ch`. This doesn't return anything, but prints the following in syslog: `getent: gethostby*.getanswer: asked for "www.noisette.ch", got "noisette.ch"`

### Expected behaviour
I would expect to have the ip address of the domain returned, just like `nslookup` or `host` 
(which seems more permissive than gethostby) commands do:
```
$ host www.noisette.ch
noisette.ch has address 87.98.177.202
www.noisette.ch is an alias for noisette.ch.
```

### Actual behaviour
tcpdump'ing the traffic shows that the A rr comes before the CNAME record, which is not right.

Habbie:
Sidenote: your getent (so your libc?) is broken and you should report that to your distribution or to upstream.

killerwhile:
@Habbie libc seems to be more picky on the RFC than other tools, but pdns still sort too many RR when used with SortList feature.

killerwhile:
The PR listed above works fine for me. Thanks!

PR #5615, opened 2017-08-14
Fix sortlist in the presence of CNAME

ahupowerdns:
In #5357 @killerwhile discovered we were missorting CNAME records when using sortlist.
With this commit, we should get it right by moving to stable_sort and being more careful about type equivalence.

### Short description
Fix sortlist in the presence of CNAME

### Checklist

I have:
- [ ] documented the code

### DIFF
--- a/pdns/pdns_recursor.cc
+++ b/pdns/pdns_recursor.cc
@@ -1089,7 +1089,7 @@ static void startDoResolve(void *p)
       if(ret.size()) {
         orderAndShuffle(ret);
 	if(auto sl = luaconfsLocal->sortlist.getOrderCmp(dc->d_remote)) {
-	  sort(ret.begin(), ret.end(), *sl);
+	  stable_sort(ret.begin(), ret.end(), *sl);
 	  variableAnswer=true;
 	}
       }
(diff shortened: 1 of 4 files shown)


Habbie:
Unrelated to this PR: https://doc.powerdns.com/recursor/lua-config/sortlist.html#addsortlist does not mention this disables the packet cache.

Habbie:
```
;; ANSWER SECTION:
order.example.com.	120	IN	MX	25 mx.example.com.
ordercname.example.com.	120	IN	CNAME	order.example.com.
```

This fix works for A/AAAA but not for other types.

Habbie:
> This fix works for A/AAAA but not for other types.

Corrected in 'two non-address records sort equal'

Habbie:
Simplification proposal: -only- move address records around, leave everything else where it is:
```diff
-  // anything address related is always 'larger', rest is equal
-
-  if(aAddr && !bAddr)
-    return false;
-  else if(!aAddr && bAddr)
-    return true;
-  else if(!aAddr && !bAddr)
-    return false;
-
+  if(!aAddr || !bAddr) return false;
```

Habbie:
Now with test but it sometimes fails.

killerwhile:
I successfully tested in on my side (backported to 4.0.6). Thanks!

Habbie:
Thank you for reporting back!

PR #5621, opened 2017-08-17
rec: Log the policy type (QName, Client IP, NS IP...) over protobuf

rgacogne:
### Short description

Add a new field to our protocol buffer format to be able to know which type of policy matched. This is especially useful when using one or more `RPZ` zone containing different types of rules.

### Checklist

I have:
- [ ] documented the code

### DIFF
--- a/contrib/ProtobufLogger.py
+++ b/contrib/ProtobufLogger.py
@@ -66,6 +66,21 @@ class PDNSPBConnHandler(object):
                                               message.question.qType,
                                               message.question.qName))
 
+    @staticmethod
+    def getAppliedPolicyTypeAsString(polType):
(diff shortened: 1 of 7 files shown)


contrib/ProtobufLogger.py
```diff
@@ -66,6 +66,21 @@ def printQuery(self, message):
                                               message.question.qType,
                                               message.question.qName))
 
+    @staticmethod
+    def getAppliedPolicyTypeAsString(polType):
+        if polType == dnsmessage_pb2.PBDNSMessage.UNKNOWN:
```
Habbie: Would a dict lookup make more sense here?

PR #5637, opened 2017-08-21
remove unused SortList compare operator for ComboAddress

Habbie:
### Checklist

I have:
- [ ] documented the code


rgacogne:
It looks like there is also an unused `extern SortList g_sortlist;` in `pdns_recursor.cc` that should go, but that can be done in a separate PR.

Habbie:
> It looks like there is also an unused extern SortList g_sortlist; in pdns_recursor.cc that should go, but that can be done in a separate PR.

pushed

PR #5643, opened 2017-08-22
dnsdist: Add RewriteMapAction and RewriteMapResponseAction

rgacogne:
### Short description

Add `RewriteMapAction` to rewrite the qname of queries based on the content of the map. For example, if the map contains a key `example.com.` with the associated value `notexample.org.`, a query for `test.example.com.` will be sent to the backend with a QName of `test.notexample.org.`.
To function properly, a matching `RewriteMapResponseAction` action should be defined to rewrite the names and contents in responses. Only the following record types see their content rewritten, others are left untouched:
* CNAME
* MX
* NS
* SRV

Remaining questions:
* do we need to handle a query for zone `a` getting a record pointing to a name in zone `b`, where both `a` and `b` are in the map?
* do we want to handle the case where a query for a zone not present in the map receives a record pointing to a zone that is in the map?

### Checklist

I have:
- [ ] documented the code


pdns/dnsdist-lua-actions.cc
```diff
@@ -923,6 +924,189 @@ class TagResponseAction : public DNSResponseAction
...
+  DNSName rebased(*dq->qname);
+  rebased.rebase(pair->first, pair->second);
+  char* buffer = reinterpret_cast<char*>(dq->dh);
+  const size_t originalQNameSize = dq->qname->wirelength();
+  const std::string newQNameWire =  rebased.toDNSString();
+  const size_t newQNameSize = newQNameWire.size();
+  size_t newSize = dq->len - originalQNameSize + newQNameSize;
+
+  if (dq->len < originalQNameSize || newSize > dq->size) {
+    return Action::None;
```
zeha: some form of warnlog here?

pdns/dnsdist-lua-actions.cc
```diff
@@ -923,6 +924,189 @@ class TagResponseAction : public DNSResponseAction
...
+  catch(const std::exception& e) {
+    vinfolog("Got exception while rewriting answer: %s", e.what());
+  }
+
+  return false;
+}
+
+DNSResponseAction::Action RewriteMapResponseAction::operator()(DNSResponse* dr, string* ruleresult) const
+{
+  const auto pair = d_smt.lookup(*dr->qname);
```
zeha: could have a nicer name than `pair` here. maybe `fromTo`?

rgacogne:
Moving this to 1.4.0 since we still have open questions on that feature.

rgacogne:
Closing for lack of interest!

Issue #2613, opened 2015-06-23
rec: Queries answered from packetcache are not included in top-*-queries ringbuffers


qvr:
It would be nice to have them included in the stats seeing as how 90-95% of queries are answered from packetcache, but at minimum the docs should probably be fixed to reflect this. (Of course at the moment the commands aren't documented at all.)

PR #5644, opened 2017-08-22
Always add the QNAME|QTYPE to the query ringbuffer

pieterlexis:
### Short description
Before, queries answered from the PacketCache would not be added to the
query ring buffer (but would be in the answer related ring buffers).

Also document the behaviour _and_ the change.

Closes: #2613

Some questions (maybe for 4.2):

 * Would this be the correct behaviour?
 * Do we need a separate ringbuffer for non-packetcached queries?
 * Should DNSSEC Bogus SERVFAILs ens up in the SERVFAIL ringbuffer (right now they do not)
 * Should DNSSEC results be in a separate ringbuffer?

### Checklist

I have:
- [ ] documented the code

### DIFF
--- a/pdns/pdns_recursor.cc
+++ b/pdns/pdns_recursor.cc
@@ -953,9 +953,6 @@ static void startDoResolve(void *p)
 {
   DNSComboWriter* dc=(DNSComboWriter *)p;
   try {
-    if (t_queryring)
-      t_queryring->push_back(make_pair(dc->d_mdp.d_qname, dc->d_mdp.d_qtype));
-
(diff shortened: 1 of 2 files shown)


pieterlexis:
rebased, still needs discussion

phonedph1:
DNSSEC should be covered by #6713 now.

I'm somewhat partial to having something separate. A bunch of queries for google/facebook/whatever in stats is not overly helpful (which easily fill a smaller buffer). The current behaviour of showing recent/work-causing queries is much more helpful IMO.

rgacogne:
I also wonder what the performance impact of this PR is, since we try to keep packet cache hits as fast as possible.

PR #5647, opened 2017-08-24
Make dnsdist dynamic truncate do right thing on TCP/IP

ahupowerdns:
Winfried noted that our new dynamic truncation rule worked fine on UDP, but on TCP/IP a truncate would be converted into a drop, which was not the intended effect.
This commit makes dynamic truncate a NOOP on TCP.

### Checklist

I have:
- [ ] documented the code

### DIFF
--- a/pdns/dnsdist.cc
+++ b/pdns/dnsdist.cc
@@ -878,23 +878,33 @@ bool processQuery(LocalStateHolder<NetmaskTree<DynBlock> >& localDynNMGBlock,
 
   if(auto got=localDynNMGBlock->lookup(*dq.remote)) {
     if(now < got->second.until) {
-      g_stats.dynBlocked++;
-      got->second.blocks++;
       DNSAction::Action action = got->second.action;
       if (action == DNSAction::Action::None) {
         action = g_dynBlockAction;
       }
       if (action == DNSAction::Action::Refused) {
         vinfolog("Query from %s refused because of dynamic block", dq.remote->toStringWithPort());
+        g_stats.dynBlocked++;
+        got->second.blocks++;
+      
         dq.dh->rcode = RCode::Refused;
         dq.dh->qr=true;
         return true;
       }
-      else if (action == DNSAction::Action::Truncate && !dq.tcp) {
-        vinfolog("Query from %s truncated because of dynamic block", dq.remote->toStringWithPort());
-        dq.dh->tc = true;
-        dq.dh->qr = true;
-        return true;
+      else if (action == DNSAction::Action::Truncate) {
+        if(!dq.tcp) {
+          g_stats.dynBlocked++;
+          got->second.blocks++;
+
+          vinfolog("Query from %s truncated because of dynamic block", dq.remote->toStringWithPort());
+          dq.dh->tc = true;
+          dq.dh->qr = true;
+          return true;
+        }
+        else {
+          vinfolog("Query from %s for %s over TCP *not* truncated because of dynamic block", dq.remote->toStringWithPort(), dq.qname->toString());
+        }
+
       }
       else {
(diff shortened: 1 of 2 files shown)


ahupowerdns:
Note: there is already a test that should have caught this bug. However, that test did not function it appears. Still investigating.

ahupowerdns:
test_DynBlocks.py:TestDynBlockQPSActionTruncated.testDynBlocksQRate

pdns/dnsdist.cc
```diff
@@ -878,23 +878,33 @@ bool processQuery(LocalStateHolder<NetmaskTree<DynBlock> >& localDynNMGBlock,
...
+        if(!dq.tcp) {
+          g_stats.dynBlocked++;
+          got->second.blocks++;
+
+          vinfolog("Query from %s truncated because of dynamic block", dq.remote->toStringWithPort());
+          dq.dh->tc = true;
+          dq.dh->qr = true;
+          return true;
+        }
+        else {
```
pieterlexis: nit: Personally, I am not a fan superfluous `else` statements. As there is already a return in the body of the `if`.

pdns/dnsdist.cc
```diff
@@ -905,23 +915,32 @@ bool processQuery(LocalStateHolder<NetmaskTree<DynBlock> >& localDynNMGBlock,
...
+          vinfolog("Query from %s for %s truncated because of dynamic block", dq.remote->toStringWithPort(), dq.qname->toString());
+          dq.dh->tc = true;
+          dq.dh->qr = true;
+          return true;
+        }
+        else {
+          vinfolog("Query from %s for %s over TCP *not* truncated because of dynamic block", dq.remote->toStringWithPort(), dq.qname->toString());
+        }
       }
       else {
```
pieterlexis: Shouldn't this block also do 
```
g_stats.dynBlocked++;
got->blocks++;
```

pdns/dnsdist.cc
```diff
@@ -878,23 +878,33 @@ bool processQuery(LocalStateHolder<NetmaskTree<DynBlock> >& localDynNMGBlock,
...
+          dq.dh->tc = true;
+          dq.dh->qr = true;
+          return true;
+        }
+        else {
+          vinfolog("Query from %s for %s over TCP *not* truncated because of dynamic block", dq.remote->toStringWithPort(), dq.qname->toString());
+        }
+
       }
       else {
```
pieterlexis: Shouldn't this block also do 
```
g_stats.dynBlocked++;
got->blocks++;
```

pdns/dnsdist.cc
```diff
@@ -905,23 +915,32 @@ bool processQuery(LocalStateHolder<NetmaskTree<DynBlock> >& localDynNMGBlock,
...
+        if(!dq.tcp) {
+          g_stats.dynBlocked++;
+          got->blocks++;
+      
+          vinfolog("Query from %s for %s truncated because of dynamic block", dq.remote->toStringWithPort(), dq.qname->toString());
+          dq.dh->tc = true;
+          dq.dh->qr = true;
+          return true;
+        }
+        else {
```
pieterlexis: nit: Personally, I am not a fan superfluous `else` statements. As there is already a return in the body of the `if`.

pdns/dnsdist.cc
ahupowerdns: No, it precisely should not do that - on TCP "truncate" does nothing.

pdns/dnsdist.cc
ahupowerdns: See above.

pdns/dnsdist.cc
pieterlexis: This is in the 'dropped' block

ahupowerdns: good catch, redoing.

PR #5684, opened 2017-09-11
be more specific about trailing dot handling, fixes #5673

### LINKED ISSUE #5673: Segfault in pdnsutil check-zone
Sample database using postgresql backend:
```
insert into domains (name, type) values ('example.com', 'NATIVE'); 
insert into records (domain_id, type, name, content) values (currval('domains_id_seq'), 'SOA', 'example.com', 'ns1.example.com admin.example.com 2017010100 3600 600 1209600 3600');
insert into records (domain_id, name, type, ttl, prio, content) values (currval('domains_id_seq'), 'ns.example.com', 'A', 3600, default, '127.0.0.1');
insert into records (domain_id, name, type, ttl, prio, content) values (currval('domains_id_seq'), 'example.com', 'NS', 3600, default, 'ns.example.com');
insert into records (domain_id, type, name, content, ttl, prio) values (currval('domains_id_seq'), 'MX', 'example.com', 'alt4.aspmx.l.google.com.', 3600, 10);
```
```
[root@localhost ~]# pdnsutil check-zone example.com
[Warning] Parsed and original record content are not equal: example.com IN MX '10 alt4.aspmx.l.google.com.' (Content parsed as '10 alt4.aspmx.l.google.com')
Segmentation fault
```
If I delete "prio" from MX records, I got just the error:

```
powerdns=# update records set prio = null where name = 'example.com' and type = 'MX';

[root@localhost ~]# pdnsutil check-zone example.com
[Error] Following record had a problem: example.com IN MX  alt4.aspmx.l.google.com.
[Error] Error was: Parsing record content (try 'pdnsutil check-zone'): expected digits at position 1 in ' alt4.aspmx.l.google.com.'
Checked 4 records of 'example.com', 1 errors, 0 warnings.
```

rpm -q pdns
pdns-4.1.0-0.1.rc1.1pdns.el7.x86_64

Habbie:
### Checklist

I have:
- [ ] documented the code

### DIFF
--- a/pdns/dnsrecords.cc
+++ b/pdns/dnsrecords.cc
@@ -52,6 +52,8 @@ string DNSResourceRecord::getZoneRepresentation(bool noDot) const {
     case QType::SRV:
     case QType::MX:
       stringtok(parts, content);
+      if (!parts.size())
+        return "";
       last = *parts.rbegin();
       ret << content;
       if (last == ".")
--- a/pdns/dnsrecords.cc
+++ b/pdns/dnsrecords.cc
@@ -64,10 +64,8 @@ string DNSResourceRecord::getZoneRepresentation(bool noDot) const {
     case QType::NS:
     case QType::PTR:
       ret<<content;
-      if (content != ".") {
-        if(!noDot)
-          ret<<".";
-      }
+      if (*(content.rbegin()) != '.' && !noDot)
+        ret<<".";
       break;
     default:
       ret<<content;
--- a/pdns/dnsrecords.cc
+++ b/pdns/dnsrecords.cc
@@ -45,15 +45,26 @@ void DNSResourceRecord::setContent(const string &cont) {
 
 string DNSResourceRecord::getZoneRepresentation(bool noDot) const {
   ostringstream ret;
+  vector<string> parts;
+  string last;
+
   switch(qtype.getCode()) {
     case QType::SRV:
     case QType::MX:
+      stringtok(parts, content);
+      last = *parts.rbegin();
+      ret << content;
+      if (last == ".")
+        break;
+      if (*(last.rbegin()) != '.' && !noDot)
+        ret << ".";
+      break;
     case QType::CNAME:
     case QType::DNAME:
     case QType::NS:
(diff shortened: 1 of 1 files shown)


Habbie:
This is ready for review but I'm planning to add a test. The test itself is not an auth 4.1 blocker so merge away after review.

pdns/dnsrecords.cc
```diff
@@ -45,15 +45,26 @@ void DNSResourceRecord::setContent(const string &cont) {
...
+        ret << ".";
+      break;
     case QType::CNAME:
     case QType::DNAME:
     case QType::NS:
     case QType::PTR:
-      if (*(content.rbegin()) != '.') {
-        ret<<content;
+      ret<<content;
+      if (content != ".") {
```
rgacogne: I'm not sure I agree with this check, as it seems to me that it will add a dot if the existing content is not a single dot, even if it does already end with one? Perhaps the existing check makes more sense, as long as:
> res<<content;

is done before the check.

pdns/dnsrecords.cc
```diff
@@ -45,15 +45,26 @@ void DNSResourceRecord::setContent(const string &cont) {
...
+  vector<string> parts;
+  string last;
+
   switch(qtype.getCode()) {
     case QType::SRV:
     case QType::MX:
+      stringtok(parts, content);
+      last = *parts.rbegin();
+      ret << content;
+      if (last == ".")
```
rgacogne: Just a nit but this test looks redundant with the next one.

pdns/dnsrecords.cc
Habbie: If I feed in `15 example.com.` with noDot false, out comes `15 example.com.` so I think I'm missing the problem

Habbie: Sorry, you are right. Pushed another commit.

pdns/dnsrecords.cc
Habbie: As discussed on IRC, leaving this explicitness in.

PR #5689, opened 2017-09-14
verify parts.size(). Fixes #5688

### LINKED ISSUE #5688: Segfault#2 in pdnsutil check-zone
Sample data from postgresql:

```
insert into domains (name, type) values ('example.com', 'NATIVE'); 
insert into records (domain_id, type, name, content) values (currval('domains_id_seq'), 'SOA', 'example.com', 'ns1.example.com admin.example.com 2017010100 3600 600 1209600 3600');
insert into records (domain_id, name, type, ttl, prio, content) values (currval('domains_id_seq'), 'ns.example.com', 'A', 3600, default, '127.0.0.1');
insert into records (domain_id, name, type, ttl, prio, content) values (currval('domains_id_seq'), 'example.com', 'NS', 3600, default, 'ns.example.com');
insert into records (domain_id, type, name, content, ttl, prio) values (currval('domains_id_seq'), 'SRV', 'example.com', '0 443example.com', 3600, 10);
```
```
[root@localhost ~]# pdnsutil check-zone example.com
[Warning] Parsed and original record content are not equal: example.com IN SRV '0 0 443example.com' (Content parsed as '0 0 443 example.com')
Segmentation fault
```

Again, prio plays a role
```
update records set prio = null where domain_id = currval('domains_id_seq') and type = 'SRV';
```
```
[root@localhost ~]# pdnsutil check-zone example.com
[Error] Following record had a problem: example.com IN SRV  0 443example.com
[Error] Error was: Parsing record content (try 'pdnsutil check-zone'): expected digits at position 6 in ' 0 443example.com'
Checked 5 records of 'example.com', 2 errors, 0 warnings.
```

Habbie:
### Short description
We would read out of bounds if SRV or MX content did not have enough parts.

### Checklist

I have:
- [ ] documented the code

### DIFF
--- a/pdns/pdnsutil.cc
+++ b/pdns/pdnsutil.cc
@@ -584,16 +584,16 @@ int checkZone(DNSSECKeeper &dk, UeberBackend &B, const DNSName& zone, const vect
       if (rr.qtype.getCode() == QType::SRV) {
         vector<string> parts;
         stringtok(parts, rr.getZoneRepresentation());
-        toCheck = DNSName(parts[3]);
+        if (parts.size() == 4) toCheck = DNSName(parts[3]);
       } else if (rr.qtype.getCode() == QType::MX) {
         vector<string> parts;
         stringtok(parts, rr.getZoneRepresentation());
-        toCheck = DNSName(parts[1]);
+        if (parts.size() == 2) toCheck = DNSName(parts[1]);
       } else {
         toCheck = DNSName(rr.content);
       }
 
-      if (!toCheck.isHostname()) {
+      if (!toCheck.empty() && !toCheck.isHostname()) {
         cout<<"[Warning] "<<rr.qtype.getName()<<" record in zone '"<<zone<<"' has non-hostname content '"<<toCheck<<"'."<<endl;
         numwarnings++;
       }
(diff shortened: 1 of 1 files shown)


pdns/pdnsutil.cc
```diff
@@ -584,16 +584,16 @@ int checkZone(DNSSECKeeper &dk, UeberBackend &B, const DNSName& zone, const vect
...
         vector<string> parts;
         stringtok(parts, rr.getZoneRepresentation());
-        toCheck = DNSName(parts[1]);
+        if (parts.size() == 2) toCheck = DNSName(parts[1]);
       } else {
         toCheck = DNSName(rr.content);
       }
 
-      if (!toCheck.isHostname()) {
+      if (!toCheck.empty() && !toCheck.isHostname()) {
```
rgacogne: Perhaps we should display a warning if the hostname is empty? Something like:

` if (toCheck.empty() || !toCheck.isHostname()) {`

Habbie: I pondered that, but I couldn't imagine a code path that ends with an empty `toCheck` without having given some other warning before. But it doesn't hurt to check, so I'll fix it.

Habbie: pushed!

PR #5695, opened 2017-09-15
calidns: macOS stubs for send/recvmmsg

wojas:
### Short description

Make calidns compile on macOS, by adding stubs for send/recvmmsg.

### Checklist

I have:
- [ ] documented the code

### DIFF
--- a/pdns/calidns.cc
+++ b/pdns/calidns.cc
@@ -38,6 +38,38 @@
 using std::thread;
 using std::unique_ptr;
 
+#ifdef __APPLE__
+#define MSG_WAITFORONE 0
+// macOS does not support sendmmsg/recvmmsg, so we provide a (non-perfect) stub
+struct mmsghdr {
+    struct msghdr msg_hdr;
+    unsigned int msg_len;
+};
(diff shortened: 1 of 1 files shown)


rgacogne:
Just an idea, perhaps we could detect `sendmmsg()/recvmmsg()` during the configure, like is done in [1]  (from #5576, not yet merged), so the stubs can be used on other OS than macOS?

[1]: https://github.com/PowerDNS/pdns/pull/5576/commits/0beaa5c825059986d3ce108a4bd2578a08d6d1d0

zeha:
@wojas ping?

wojas:
A more general solution with would indeed be preferred. I just filed this PR as I had this hack sitting around locally and thought it might be useful.

Do I understand correctly that `#ifdef __APPLE__` needs to be replaced with `#ifndef HAVE_RECVMMSG` and friends? No other changes needed?

rgacogne:
> Do I understand correctly that #ifdef __APPLE__ needs to be replaced with #ifndef HAVE_RECVMMSG and friends? No other changes needed?

It looks like this would be it, yes. But please feel free to let me know if you don't have the time to look into  it and I will!

omoerbeek:
Was fixed by #7871. Never noticed this PR before...

PR #5699, opened 2017-09-15
implement dynamic cache sizeing for recursor

ahupowerdns:
### Short description
With this commit, the number of (packet)cache entries can be changed at runtime, although the effect may not be immediate in case of shrinking the cache.

### Checklist

I have:
- [ ] documented the code

### DIFF
--- a/pdns/pdns_recursor.cc
+++ b/pdns/pdns_recursor.cc
@@ -156,6 +156,7 @@ static bool g_useOneSocketPerThread;
 static bool g_gettagNeedsEDNSOptions{false};
 static time_t g_statisticsInterval;
 static bool g_useIncomingECS;
+std::atomic<uint32_t> g_maxCacheEntries, g_maxPacketCacheEntries;
 
 RecursorControlChannel s_rcc; // only active in thread 0
 RecursorStats g_stats;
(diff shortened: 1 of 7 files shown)


pdns/pdns_recursor.cc
```diff
@@ -2080,10 +2081,10 @@ static void houseKeeping(void *)
     if(now.tv_sec - last_prune > (time_t)(5 + t_id)) {
       DTime dt;
       dt.setTimeval(now);
-      t_RC->doPrune(); // this function is local to a thread, so fine anyhow
-      t_packetCache->doPruneTo(::arg().asNum("max-packetcache-entries") / g_numWorkerThreads);
+      t_RC->doPrune(g_maxCacheEntries / g_numThreads); // this function is local to a thread, so fine anyhow
+      t_packetCache->doPruneTo(g_maxPacketCacheEntries / g_numWorkerThreads);
 
-      SyncRes::pruneNegCache(::arg().asNum("max-cache-entries") / (g_numWorkerThreads * 10));
+      SyncRes::pruneNegCache(g_maxCacheEntries / (g_numWorkerThreads * 10));
```
rgacogne: Not related to this PR but I wonder where does the 10 comes from in `g_numWorkerThreads * 10`. And why we divide by `g_numWorkerThreads` (the number of _worker_ threads) for the packet and neg caches, but by `g_numThreads` (number of worker threads + 1 if `pdns-distribute-queries` is set) for the query cache.

ahupowerdns:
added metrics.

PR #5711, opened 2017-09-19
add ProbaRule to dnsdist: match with given probability

ahupowerdns:
This adds a ProbaRule, ProbaRule(1.0) means 'match always', 0.1 '10%'. Useful for TeeAction.

### Short description
Sometimes you don't want an action to happen 100% of the time. With ProbaRule you can for example 'Tee' 1% of your traffic to hot spares which then always have a hot cache:

addAction(ProbaRule(0.1), TeeAction("127.0.0.1:5301"))

### Checklist

I have:
- [ ] documented the code

### DIFF
--- a/pdns/dnsrulactions.hh
+++ b/pdns/dnsrulactions.hh
@@ -760,6 +760,18 @@ public:
 };
 
 
+class ProbaRule : public DNSRule
+{
+public:
+  ProbaRule(double proba) : d_proba(proba)
+  {
+  }
+  bool matches(const DNSQuestion* dq) const override;
+  string toString() const override;
+  double d_proba;
+};
+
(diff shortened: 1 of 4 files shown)


pdns/dnsrulactions.hh
```diff
@@ -760,6 +760,18 @@ public:
...
 
+class ProbaRule : public DNSRule
+{
+public:
+  ProbaRule(double proba) : d_proba(proba)
+  {
+  }
+  bool matches(const DNSQuestion* dq) const override;
+  string toString() const override;
+  double d_proba;
```
rgacogne: just a nit, but `d_proba` could be private.

PR #5716, opened 2017-09-20
rec: Detect zone cuts by asking for DS instead of NS

### LINKED ISSUE #5681: rec: Handle broken NS answers (like the ones from fastly) when looking for zone cuts
- Program: Recursor
 - Issue type: Bug report

### Short description

Some of fastly.net's authoritative servers answer with NODATA when asked fastly.net|NS, which messes up our zone cut detection code, leading to DNSSEC validation failure. They are aware of the issue and the offending servers will be removed, but we need to handle that case more gracefully, especially since it now looks like incapdns.net has the same issue.

rgacogne:
### Short description

Since there is so many broken DNS servers handing NS queries incorrectly, move to DS queries for figuring out zone cuts since we only care as long as the zone is Secure (including with a TA).
Please review this carefully before merging.
Fixes #5681.

### Checklist

I have:
- [ ] documented the code

### DIFF
--- a/pdns/validate.cc
+++ b/pdns/validate.cc
@@ -42,6 +42,59 @@ static bool isCoveredByNSEC(const DNSName& name, const DNSName& begin, const DNS
           (begin == next && name != begin));                        // "we have only 1 NSEC record, LOL!"
 }
 
+static std::string getHashFromNSEC3(const DNSName& qname, const std::shared_ptr<NSEC3RecordContent> nsec3)
+{
+  std::string result;
+
+  if (g_maxNSEC3Iterations && nsec3->d_iterations > g_maxNSEC3Iterations) {
+    return result;
+  }
+
+  return hashQNameWithSalt(nsec3->d_salt, nsec3->d_iterations, qname);
+}
+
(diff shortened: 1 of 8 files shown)


pdns/validate.cc
```diff
@@ -161,22 +173,38 @@ dState getDenial(const cspmap_t &validrrsets, const DNSName& qname, const uint16
...
         string beginHash=fromBase32Hex(v.first.first.getRawLabels()[0]);
 
+        /* RFC 6840 section 4.1 "Clarifications on Nonexistence Proofs":
+           Ancestor delegation NSEC or NSEC3 RRs MUST NOT be used to assume
+           nonexistence of any RRs below that zone cut, which include all RRs at
+           that (original) owner name other than DS RRs, and all RRs below that
+           owner name regardless of type.
+        */
+        if (nsec3->d_set.count(QType::NS) && !nsec3->d_set.count(QType::SOA) &&
+            getSigner(v.second.signatures).countLabels() < v.first.first.countLabels()) {
```
pieterlexis: should we check for `v.first.first.isPartOf(getSigner(v.second.signatures))` here?

PR #5726, opened 2017-09-25
Rec 4.0.x: be more resilient with broken auths

pieterlexis:
### Short description
This PR solves an issue where we would SERVFAIL while trying to validate zones for which the auths do either not understand EDNS0 or give NXDOMAIN/SERVFAIL answers for non-A/AAAA queries.

The first commit ensures we stop validating a name once we hit an Insecure zone cut, all the while cleaning up getZoneCuts (no more pointer magic).

The second commit makes EDNS0 not mandatory anymore for DNSSEC (to work around broken auths).

### Checklist

I have:
- [x] documented the code


pdns/syncres.cc
```diff
@@ -369,9 +369,12 @@ int SyncRes::asyncresolveWrapper(const ComboAddress& ip, bool ednsMANDATORY, con
     
     if(mode==EDNSStatus::NOEDNS) {
       g_stats.noEdnsOutQueries++;
+      if (ednsMANDATORY) {
```
rgacogne: We should probably rename `ednsMANDATORY` to something like `wantsDNSSEC`, since it's not mandatory anymore.

pdns/validate.cc
```diff
@@ -414,35 +405,41 @@ vState getKeysFor(DNSRecordOracle& dro, const DNSName& zone, keyset_t &keyset)
...
+    if(current_cut == zone) {
       LOG("requested keyset found! returning Secure for the keyset"<<endl);
       keyset.insert(validkeys.cbegin(), validkeys.cend());
       return Secure;
     }
 
     // We now have the DNSKEYs, use them to validate the DS records at the next zonecut
-    LOG("next name ["<<*(zoneCutIter+1)<<"], trying to get DS"<<endl);
+    DNSName next_cut = getNextCut(current_cut, zone, dro);
+    if (next_cut == current_cut) {
```
rgacogne: I'm not sure in what case this might happen? It seems to me that `getNextCut()` will never return `current_cut`, except if `current_cut` is equal to `zone` which is handled a few lines above. Not an issue, of course, but I'm wondering if I'm missing something here.

pieterlexis: This is a 'just in case' so we always break out of the loop. `getNextCut` returns `current_cut` if it cannot find one.

pdns/syncres.cc
pieterlexis: agreed! maybe `wantsEDNS` as it can be used for other things than DNSSEC (EDNSCookies etc.)

pdns/validate.cc
rgacogne: It looks like it would return `zone`, not `current_cut` (ie `target` and not `current` inside `getNextCut()`)?

pieterlexis: looks like you're right, fixing

Issue #5719, opened 2017-09-21
pdnsutil export-zone-dnskey erroneously exports DS records


jpmens:
Auth version `4.1.0-rc1`

The command `pdnsutil export-zone-ds` exports DS records, but the command `pdnsutil export-zone -dnskey` exports a DNSKEY *and* DS records. This is inconsistent behavior.

Apparently it's been that way since 2010; I propose to remove the printing of DS records from that particular subcommand.

aerique:
Notes to self:

16:15: habbie, rgacogne Is jpmens' proposed solution a good one?  I can imagine I can assign myself and do it Monday. (https://github.com/PowerDNS/pdns/issues/5719) (edited)
16:15 habbie: yes
16:15 habbie: i checked with a few operators and none of them are extracting DSes from export-zone-dnskey
16:16 habbie: if you just remove the DS bits from export-zone-dnskey you fix two(!) bugs
16:16 erikw: does the other bug have an issue?
16:16 habbie: no
16:16 habbie: the DS printing is wrong in export-zone-dnskey
16:17 habbie: so removing it will fix that as well

PR #5729, opened 2017-09-25
Remove printing of DS records from export-zone-dnskey.

aerique:
### Short description

Remove printing of DS records from `pdnsutil export-zone-dnskey …`.  This was not only inconsistent behaviour[1] but also done incorrectly[2].

Fixes https://github.com/PowerDNS/pdns/issues/5719.

[1] as noted by @jpmens
[2] as noted by @Habbie

### Checklist

I have:
- [ ] documented the code


jpmens:
As far as I can tell, removal of the DS printing looks ok. ;)

Issue #4703, opened 2016-11-23
pdnsutil edit-zone adds ascii whitespace \032 to some records


nomnomtom:
- Program: pdnsutil
 - Issue type: Bug report

### Short description

When adding certain records (AFSDB) via `pdnsutil edit-zone zonename` it pads the argument with an encoded space character (`\032`)

### Environment

 - Operating system: Debian Stretch
 - Software version: 4.0.1
 - Software source: stretch package
 - Backend: Postgresql on the same server

### Steps to reproduce

1. run `pdnsutil edit-zone example.com`
2. add the following: `example.com	3600	IN	AFSDB	1 afs.example.com.`
3. save.

### Expected behaviour

I'd expect pdnsutil to parse the line as it is or at least not add any spaces every time I save.

### Actual behaviour

pdnsutil prints the following warning and prompts to edit the zone or apply the changes:

`[Warning] Parsed and original record content are not equal: example.com IN AFSDB '1 \032afs.example.com.' (Content parsed as '1 \032afs.example.com')`

If I save, I can verify that the whitespace is in the record:

```
;; QUESTION SECTION:
;example.com.			IN	AFSDB

;; ANSWER SECTION:
example.com.		3600	IN	AFSDB	1 \032afs.example.com.
```

If I edit the zone and save it again, another whitespace is added.

Until now I could only observe this behaviour with AFSDB records.

### Other information

This does not seem to be connected to the editor used. Adding a dot at the end of the domain when editing the zone does not change the behaviour.

PR #5730, opened 2017-09-25
Unit test for AFSDB fix (PR #5667)

aerique:
### Short description

This adds the requested unit test to PR #5667.

**Question for reviewers**: in the unit test only check for QType and content or keeps all checks?

Closes #5667.
Closes https://github.com/PowerDNS/pdns/issues/4703.

### Checklist

I have:
  - both before and after the fix (so before the new unit test fails and after it passes)
- [ ] documented the code

### DIFF
--- a/pdns/zoneparser-tng.cc
+++ b/pdns/zoneparser-tng.cc
@@ -468,23 +468,22 @@ bool ZoneParserTNG::get(DNSResourceRecord& rr, std::string* comment)
     }
     break;
   case QType::AFSDB:
-    try {
-      stringtok(recparts, rr.content);
-      if(recparts.size() == 2)
-      {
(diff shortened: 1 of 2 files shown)


modules/remotebackend/test-remotebackend.cc
```diff
@@ -339,4 +339,15 @@ BOOST_AUTO_TEST_CASE(test_method_directBackendCmd) {
    BOOST_CHECK_EQUAL(be->directBackendCmd("PING 1234"), "PING 1234");
 }
 
+BOOST_AUTO_TEST_CASE(test_method_lookup) {
```
Habbie: This test is specific to remotebackend. Please add a test or two to pdns/test-dnsrecords_cc.cc instead.

PR #5738, opened 2017-09-27
rec: Do not allow direct queries for RRSIG or NSEC3

### LINKED ISSUE #5735: powerdns.com RRSIG query breaks subsequent queries
- Program: Recursor
 - Issue type: Bug report

### Short description
With #5715, a (rightfully failing) `dig rrsig powerdns.com` query will break a subsequent query for NSEC. (Without #5715, the NSEC query fails even without a preceding RRSIG query).

### Environment

 - Operating system: macOS
 - Software version: #5715 (7030a59b0)
 - Software source: git

### Steps to reproduce
1. run `./pdns_recursor  --socket-dir=. --local-port=5300 --dnssec=validate --trace`
2. run `dig rrsig powerdns.com @127.0.0.1 -p 5300 +dnssec`
3. run `dig nsec powerdns.com @127.0.0.1 -p 5300 +dnssec`

### Expected behaviour
On the second dig, NSEC + RRSIG, with `ad` bit.

### Actual behaviour
`SERVFAIL` (because Bogus). If I add `+cd`, I get the NSEC without the RRSIG.

### Additional information
After writing all of the above I realised this is not even specific to NSEC. MX also breaks, with or without #5715.

rgacogne:
### Short description

As pointed out in https://github.com/PowerDNS/pdns/issues/5735, the recursor has an interesting issue when handling a direct query for the `RRSIG` type. It would send a query to the corresponding authoritative server and assume that a `NOTIMPL` answer, as sent for example by PowerDNS 4.0.x, is an indication that this server does not support `EDNS`.
Since it does not make a lot of sense to directly ask for `RRSIG` or `NSEC3` records anyway, let's just refuse those queries as we already do for `AXFR` and `IXFR`.

One thing to ponder is that the current code sends a `ServFail` answer for `AXFR` and `IXFR`, and that feels wrong. Perhaps `REFUSED` would be more suited?

Fixes #5735.

### Checklist

I have:
- [ ] documented the code

### DIFF
--- a/pdns/recursordist/test-syncres_cc.cc
+++ b/pdns/recursordist/test-syncres_cc.cc
@@ -2151,13 +2151,13 @@ BOOST_AUTO_TEST_CASE(test_qclass_none) {
   BOOST_CHECK_EQUAL(queriesCount, 0);
 }
 
-BOOST_AUTO_TEST_CASE(test_xfr) {
+BOOST_AUTO_TEST_CASE(test_special_types) {
   std::unique_ptr<SyncRes> sr;
   initSR(sr);
 
   primeHints();
 
-  /* {A,I}XFR should be rejected right away */
+  /* {A,I}XFR, RRSIG and NSEC3 should be rejected right away */
   size_t queriesCount = 0;
 
   sr->setAsyncCallback([&queriesCount](const ComboAddress& ip, const DNSName& domain, int type, bool doTCP, bool sendRDQuery, int EDNS0Level, struct timeval* now, boost::optional<Netmask>& srcmask, boost::optional<const ResolveContext&> context, std::shared_ptr<RemoteLogger> outgoingLogger, LWResult* res) {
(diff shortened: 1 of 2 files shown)


pieterlexis:
Perhaps drop direct NSEC queries as well?

Habbie:
> One thing to ponder is that the current code sends a ServFail answer for AXFR and IXFR, and that feels wrong. Perhaps REFUSED would be more suited?

BIND sends NOTAUTH (because it could be, software-wise). 8.8.8.8 sends ServFail. Unbound sends REFUSED.

rgacogne:
> BIND sends NOTAUTH (because it could be, software-wise). 8.8.8.8 sends ServFail. Unbound sends REFUSED. Knot gives NOTIMP.

`NOTAUTH` seems a bit misleading, one could assume we looked at the qname and found we are not authoritative for it. `NOTIMP` has caused issues with `EDNS`-awareness detection in the past. `REFUSED` looks marginally better to me than `ServFail` but given that it means changing the current behavior, perhaps we should stick with `ServFail`.

Habbie:
Yes - NOTIMP is wrong. NOTAUTH is 'correct' but indeed a bit misleading. I am fine with sticking with ServFail.

PR #5740, opened 2017-09-28
rec: Lowercase all outgoing qnames when lowercase-outgoing is set

pieterlexis:
### Short description
Before, we would only lowercase the original qname before handing it to SyncRes. Now the syncresolveWrapper lowercases if it has to before handing the qname to asyncresolve.

### Checklist
I have:
- [ ] documented the code

### DIFF
--- a/pdns/recursordist/test-syncres_cc.cc
+++ b/pdns/recursordist/test-syncres_cc.cc
@@ -7434,6 +7434,68 @@ BOOST_AUTO_TEST_CASE(test_dnssec_rrsig_cache_validity) {
   BOOST_CHECK_EQUAL(queriesCount, 4);
 }
 
+BOOST_AUTO_TEST_CASE(test_lowercase_outgoing) {
+  g_lowercaseOutgoing = true;
+  std::unique_ptr<SyncRes> sr;
+  initSR(sr);
+
+  primeHints();
+
+  vector<DNSName> sentOutQnames;
+
+  const DNSName target("WWW.POWERDNS.COM");
+  const DNSName cname("WWW.PowerDNS.org");
+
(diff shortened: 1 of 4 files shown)


pdns/recursordist/test-syncres_cc.cc
```diff
@@ -6983,6 +6984,68 @@ BOOST_AUTO_TEST_CASE(test_nsec_ancestor_nxqtype_denial) {
...
+  primeHints();
+
+  vector<DNSName> sentOutQnames;
+
+  const DNSName target("WWW.POWERDNS.COM");
+  const DNSName cname("WWW.PowerDNS.org");
+
+  sr->setAsyncCallback([target, cname, &sentOutQnames](const ComboAddress& ip, const DNSName& domain, int type, bool doTCP, bool sendRDQuery, int EDNS0Level, struct timeval* now, boost::optional<Netmask>& srcmask, boost::optional<const ResolveContext&> context, std::shared_ptr<RemoteLogger> outgoingLogger, LWResult* res) {
+
+      cout<<domain<<endl;
```
rgacogne: Looks like an unwanted leftover :)

pdns/recursordist/test-syncres_cc.cc
```diff
@@ -6983,6 +6984,68 @@ BOOST_AUTO_TEST_CASE(test_nsec_ancestor_nxqtype_denial) {
...
+
+      return 0;
+    });
+
+  vector<DNSRecord> ret;
+  int res = sr->beginResolve(target, QType(QType::A), QClass::IN, ret);
+  BOOST_CHECK_EQUAL(res, RCode::NoError);
+  BOOST_CHECK_EQUAL(ret.size(), 2);
+
+  BOOST_CHECK_EQUAL(sentOutQnames.size(), 4);
```
rgacogne: This should probably be a `BOOST_CHECK_REQUIRE` and/or accesses to the vector should be done using `vector::at()` so we get a nice exception if the vector doesn't have enough elements instead of a crash.

PR #5741, opened 2017-09-28
Add DNSSEC status and 'total' time spent to non-quiet logging

ahupowerdns:
### Short description
With this change, the log emitted with 'quiet=false' includes:

 * DNSSEC status for queries that underwent DNSSEC processing
 * What we already logged: time spent waiting for network for this query
 * What we already measured but not logged: total question/answer latency time

This will be helpful to debug what queries cause us to use a lot of CPU in addition to waiting for the network.

### Checklist

I have:
- [ ] documented the code

### DIFF
--- a/build-scripts/test-recursor-bulk
+++ b/build-scripts/test-recursor-bulk
@@ -12,7 +12,7 @@ set -x
 
 cd regression-tests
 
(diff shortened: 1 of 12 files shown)


ahupowerdns:
I need to investigate. I made a clean branch for this with only 1 commit. Now to find where it is!

PR #5755, opened 2017-09-29
Improve dnsbulktest experience in travis for more robustness

ahupowerdns:
This commit changes our dnsbulktest source from Alexa to Cisco Umbrella, but this turned out not to be as important as we thought.
In addition, it turns out we had been installing pdns-tools incorrectly because of wrong apt-settings. We now install pdns-tools from the master repo at repo.powerdns.com
This commit also tunes pdns_recursor to use less simultaneous outbound connections during testing, which appears to make Travis NAT happier, leading to less errors.
Finally, we use new features of dnsbulktest to extract more statistics for how well we are doing. Success is now dependent on errors and timeouts, and less on NXDOMAIN.

### Checklist

I have:
- [ ] documented the code

### DIFF
--- a/build-scripts/test-recursor-bulk
+++ b/build-scripts/test-recursor-bulk
@@ -12,7 +12,7 @@ set -x
 
 cd regression-tests
 
-wget -c -N https://s3.amazonaws.com/alexa-static/top-1m.csv.zip
+wget -c -N http://s3-us-west-1.amazonaws.com/umbrella-static/top-1m.csv.zip
 
 unzip top-1m.csv.zip
 
--- a/build-scripts/travis.sh
+++ b/build-scripts/travis.sh
@@ -331,11 +331,11 @@ install_recursor() {
     moreutils \
     snmpd"
   run "cd .."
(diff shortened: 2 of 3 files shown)


build-scripts/test-recursor-bulk
```diff
@@ -12,7 +12,7 @@ set -x
 
 cd regression-tests
 
-wget -c -N https://s3.amazonaws.com/alexa-static/top-1m.csv.zip
+wget -c -N http://s3-us-west-1.amazonaws.com/umbrella-static/top-1m.csv.zip
```
pieterlexis: This file is available over https.

build-scripts/travis.sh
```diff
@@ -331,11 +331,11 @@ install_recursor() {
     moreutils \
     snmpd"
   run "cd .."
-  run "wget https://s3.amazonaws.com/alexa-static/top-1m.csv.zip"
+  run "wget http://s3-us-west-1.amazonaws.com/umbrella-static/top-1m.csv.zip"
```
pieterlexis: This file is available over https.

PR #5764, opened 2017-10-03
autoconf: set --enable-libsodium to 'auto'

pieterlexis:
### Short description
This was 'no' before. As we (want to) use libsodium for dnsdist's
console *and* for ed25519 for the auth and the recursor, we might as wel
use it when we detect it.

This would a good change to have in auth 4.1, rec 4.1 and dndist 1.3,
currently on master.

NOTE: will need an entry in the upgrade guides

### Checklist

I have:
- [ ] documented the code

### DIFF
--- a/m4/pdns_check_libsodium.m4
+++ b/m4/pdns_check_libsodium.m4
@@ -1,26 +1,30 @@
 AC_DEFUN([PDNS_CHECK_LIBSODIUM], [
   AC_MSG_CHECKING([whether we will be linking in libsodium])
   AC_ARG_ENABLE([libsodium],
-    AS_HELP_STRING([--enable-libsodium],[use libsodium @<:@default=no@:>@]),
+    AS_HELP_STRING([--enable-libsodium],[use libsodium @<:@default=auto@:>@]),
     [enable_libsodium=$enableval],
-    [enable_libsodium=no],
+    [enable_libsodium=auto],
   )
   AC_MSG_RESULT([$enable_libsodium])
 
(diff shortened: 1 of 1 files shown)


m4/pdns_check_libsodium.m4
```diff
@@ -1,26 +1,30 @@
...
+        AC_CHECK_FUNCS([crypto_box_easy_afternm])
+        CFLAGS=$save_CFLAGS
+        LIBS=$save_LIBS
+      ])
+    ])
+  ])
+  AM_CONDITIONAL([LIBSODIUM], [test "x$LIBSODIUM_LIBS" != "x"])
+  AS_IF([test "x$enable_libsodium" = "xyes"], [
+    AS_IF([test x"$LIBSODIUM_LIBS" = "x"], [
+      AC_MSG_ERROR([Protobuf requested but libraries were not found])
```
rgacogne: s/Protobuf/libsodium/ ?

pdns/dnsdistdist/m4/dnsdist_enable_dnscrypt.m4
```diff
@@ -9,7 +9,7 @@ AC_DEFUN([DNSDIST_ENABLE_DNSCRYPT], [
   AM_CONDITIONAL([DNSCRYPT], [test "x$enable_dnscrypt" != "xno"])
 
   AM_COND_IF([DNSCRYPT], [
-    AM_COND_IF([LIBSODIUM], [
+    AS_IF([test x"$LIBSODIUM_LIBS" != "x"], [
```
rgacogne: I'm not sure we actually need this change since it looks like the `LIBSODIUM` conditional is still defined?

PR #5766, opened 2017-10-05
auth: Correctly purge entries from the caches after a transfer

rgacogne:
### Short description

Since the QC/PC split up, we only removed entries for the AXFR'd domain from the packet cache, not the query cache.
We also did not remove entries in case of IXFR. We might want to backport that second part to 4.0.

### Checklist

I have:
- [ ] documented the code


Habbie:
This fixes my inconsistency issues.

PR #5767, opened 2017-10-05
PoC: stop doing individual RRSIG queries during outbound AXFR

Habbie:
### Short description
When we serve out a PRESIGNED zone over AXFR, we get a full zone listing from the database, including the RRSIGs. We then throw those RRSIGs away only to ask for all of them *individually*, via the query cache. Before #5766 this caused inconsistency issues. After #5766 the inconsistency issues are gone but all of this is very inefficient.

This PR stops the throwing away of RRSIGs that we got from `list()` (in tcpreceiver.cc); in dnssecsigner.cc we stop hitting the database.

Further thoughts: every outbound AXFR passes through the Chunked Signing Pipe, and this does not appear to make sense for unsigned or presigned zones.

This PR is not for merging yet - it breaks direct queries.

### Checklist

I have:
- [ ] documented the code


pieterlexis:
closing in favor of #5838

PR #5773, opened 2017-10-05
rec: Don't negcache entries for longer than their RRSIG validity

rgacogne:
### Checklist

I have:
- [ ] documented the code

### DIFF
--- a/pdns/recursordist/test-syncres_cc.cc
+++ b/pdns/recursordist/test-syncres_cc.cc
@@ -278,10 +278,10 @@ static bool addRRSIG(const testkeysset_t& keys, std::vector<DNSRecord>& records,
   }
 
   DNSRecord rec;
+  rec.d_type = QType::RRSIG;
   rec.d_place = records[recordsCount-1].d_place;
   rec.d_name = records[recordsCount-1].d_name;
-  rec.d_type = QType::RRSIG;
-  rec.d_ttl = sigValidity;
+  rec.d_ttl = records[recordsCount-1].d_ttl;
 
   rec.d_content = std::make_shared<RRSIGRecordContent>(rrc);
   records.push_back(rec);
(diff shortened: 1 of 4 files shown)


pdns/recursordist/test-syncres_cc.cc
```diff
@@ -278,10 +278,10 @@ static bool addRRSIG(const testkeysset_t& keys, std::vector<DNSRecord>& records,
   }
 
   DNSRecord rec;
+  rec.d_type = QType::RRSIG;
   rec.d_place = records[recordsCount-1].d_place;
   rec.d_name = records[recordsCount-1].d_name;
-  rec.d_type = QType::RRSIG;
-  rec.d_ttl = sigValidity;
+  rec.d_ttl = records[recordsCount-1].d_ttl;
```
Habbie: shouldn't we cap this to the max of `d_ttl` and `sigValidity`?

rgacogne: Section 3 of rfc4034 [1] states:

> The TTL value of an RRSIG RR MUST match the TTL value of the RRset it covers.

so this looks right to me?

[1]: https://tools.ietf.org/html/rfc4034#section-3

Habbie: I may be confused, but say we receive an RRSIG with an 86400 TTL but expiry is in a few hours. We should only cache it for those few hours, and I get the impression this change would cache it for a day?

rgacogne: Well this code is only there to generate mockup `RRSIG`s records for our unit tests, so it is intentionally designed to allow the creation of weird records :-)

Habbie: durrrr

PR #5207, opened 2017-03-27
Forward-zone NS speed tracking

phonedph1:
### Short description
Fix for #5187 by slightly abusing our current tracking mechanism by stuffing our ComboAddress in as a DNSName and reusing much of shuffleInSpeedOrder.

This also prints a + in the trace log if the query is recursive - much like previous versions did.

### Checklist
I have:
- [x] read the [CONTRIBUTING.md](https://github.com/PowerDNS/pdns/blob/master/CONTRIBUTING.md) document
- [x] compiled and tested this code
- [ ] included documentation (including possible behaviour changes)
- [ ] documented the code
- [ ] added or modified regression test(s)
- [ ] added or modified unit test(s)

Issue #5187, opened 2017-03-21
forward-zone behavior change 


phonedph1:
Somewhere between 3.6.x and 4.0.4 the behavior of forward-zones appears to have changed. In the 3.6.x branch it used to keep track of response times of forwarded zones, so if one of your downstreams when down it would actually depreference it and use any alternative. This can be seen in the trace logs on 3.6.x.

In 4.0.4 the behavior seems to always use the first server in your forward-zone list, before trying the next after the 1500ms network timeout (if the system is completely unresponsive).

Is this a regression or a "how we want it to work now"?

We could proxy our forward zones with dnsdist + health checking locally to get around this if it's a design change.

Thanks,

phonedph1:
I should add it does appear to throttle after server-down-max-fails properly. Most people might be running this at the default setting of 0 though.

pieterlexis:
Hi,

I did not know the behaviour was like that before, during the spring-cleaning, forward-zones were fully broken at some point. This was addressed in #3244, but this did not change how we forward. So it must have been broken before or after that. 

I think we could mark this as a regression, but I can't promise a fix soon.

pieterlexis:
I hoped fa1b87ff63cfe5da19be7b3a21e7b355ef86744f would be the culprit, but it isn't.

note:
We track response times based on the nameserver's name (e.g. ns1.example.com) not it's IP address(es). For forwarded servers, we only have one name.

phonedph1:
Totally crazy, but https://github.com/phonedph1/pdns/commit/c6ecf5ce8fd36fdfbd1369ae4e0416c769d4884f would mimic the previous behaviour.

Sadly I don't know enough about the internals (or C++) to make this something decent myself.

pieterlexis:
Nice solution, It will even work for v6 addresses (39 chars max). Maybe storing the nsspeeds as IP addresses makes more sense in general... This needs some pondering.

phonedph1:
Bit less sketchy: https://github.com/PowerDNS/pdns/commit/f42440341a9ea0c276e777562b2e4b0b718517ed

phonedph1:
Fixed by #6168

PR #5774, opened 2017-10-06
Implement experimental metric tracking time spent within PowerDNS per query

ahupowerdns:
### Short description

With this commit, PowerDNS provides metrics on the difference between the time spent waiting for authoritative servers, and the amount of time elapsed between arrival of query
and sending out the response. This metric should be seen as experimental until operational experience proves its relevance.

### Checklist

I have:
- [ ] documented the code

### DIFF
--- a/pdns/pdns_recursor.cc
+++ b/pdns/pdns_recursor.cc
@@ -1237,7 +1237,7 @@ static void startDoResolve(void *p)
     }
 
     sr.d_outqueries ? t_RC->cacheMisses++ : t_RC->cacheHits++;
-
+            
     if(spent < 0.001)
       g_stats.answers0_1++;
     else if(spent < 0.010)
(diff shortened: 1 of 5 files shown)


pdns/recursordist/docs/metrics.rst
```diff
@@ -439,3 +439,45 @@ number of seconds process has been running (since 3.1.5)
 user-msec
 ^^^^^^^^^
 number of CPU milliseconds spent in 'user' mode
+
+x-our-latency
+^^^^^^^^^^^^^
+New metric, which is not yet proven to be reliable. PowerDNS measures per query how much time has been spent waiting on
+authoritative servers. In addition, the Recursor measures the total amount of time needed to answer a question. The
+difference between these two durations is a measure of how much time was spent within PowerDNS. This metric is the average
+of that difference, in microseconds. Since 4.1.
```
pieterlexis: please the following syntax:
```rst
x-our-latency
^^^^^^^^^^^^^
.. versionadded:: 4.1

TEXT
```

pdns/pdns_recursor.cc
```diff
@@ -1253,6 +1253,28 @@ static void startDoResolve(void *p)
     newLat = min(newLat,(uint64_t)(((uint64_t) g_networkTimeoutMsec)*1000)); // outliers of several minutes exist..
     g_stats.avgLatencyUsec=(1-1.0/g_latencyStatSize)*g_stats.avgLatencyUsec + (float)newLat/g_latencyStatSize;
     // no worries, we do this for packet cache hits elsewhere
+
+    auto ourtime = 1000.0*spent-sr.d_totUsec/1000.0; // in msec
```
rgacogne: It looks like we are not updating `SyncRes::d_totUsec` when we forward a cache-only (`RD`=0) query to a hard-coded server in `SyncRes::doResolve()`. This would look like we spent a lot of time in the recursor when we didn't.

pdns/pdns_recursor.cc
```diff
@@ -1237,7 +1237,7 @@ static void startDoResolve(void *p)
     }
 
     sr.d_outqueries ? t_RC->cacheMisses++ : t_RC->cacheHits++;
-
+
```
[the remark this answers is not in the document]
ahupowerdns: removed

pdns/pdns_recursor.cc
ahupowerdns: nice! solved.

pdns/syncres.cc
```diff
@@ -536,6 +536,10 @@ int SyncRes::doResolve(const DNSName &qname, const QType &qtype, vector<DNSRecor
 
           boost::optional<Netmask> nm;
           res=asyncresolveWrapper(remoteIP, d_doDNSSEC, qname, qtype.getCode(), false, false, &d_now, nm, &lwr);
+
+          d_totUsec += lwr.d_usec;
+          accountAuthLatency(lwr.d_usec, remoteIP.sin4.sin_family);
```
pieterlexis: I think this incidentally solves  #5207 and #5187?

rgacogne: I don't think it does?

pieterlexis: ah, this has indeed nothing to do with the speed tracking per-ns

Issue #5158, opened 2017-03-15
Cleanup or remove preprocessor directives for specific GCC versions


rgacogne:
- Program: Authoritative, Recursor, dnsdist
 - Issue type: Feature request

### Short description
We still have several occurrences of preprocessor directives dealing with old gcc versions -some of them as old as gcc 2.x- while our usage of C++11 now requires at least 4.7, for example for delegating constructors. We should clean that up.

zeha:
I cannot find any of those, maybe #5775 got rid of all of them?

pieterlexis:
yes! 😄

PR #5775, opened 2017-10-06
Remove preprocessor directives for older GCC versions

MatusKysel:
…5158

### Short description
Removing preprocessor directives for GCC version lower then 4.7 issue #5158 

### Checklist

I have:
- [ ] documented the code

### DIFF
--- a/pdns/test-dns_random_hh.cc
+++ b/pdns/test-dns_random_hh.cc
@@ -2,7 +2,7 @@
 #define BOOST_TEST_NO_MAIN
 
 // Disable this code for gcc 4.8 and lower
-#if (__GNUC__ == 4 && __GNUC_MINOR__ > 8) || !__GNUC__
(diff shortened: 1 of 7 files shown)


pdns/test-dns_random_hh.cc
```diff
@@ -2,7 +2,7 @@
 #define BOOST_TEST_NO_MAIN
 
 // Disable this code for gcc 4.8 and lower
-#if (__GNUC__ > 4) || (__GNUC__ == 4 && __GNUC_MINOR__ > 8) || !__GNUC__
+#if (__GNUC__ == 4 && __GNUC_MINOR__ > 8) || !__GNUC__
```
rgacogne: This looks wrong, we want to disable this code for versions of GCC lower than 4.8, and this would only enable it for the 4 branch after 4.8.

pieterlexis: I am not even sure we can build on GCC < 4.8 tbh

MatusKysel: You are right, will fix it

PR #5779, opened 2017-10-06
Rectify zones via the API 

### LINKED ISSUE #5712: API zone endpoint should support an 'rrset' parameter
- Program: Authoritative
 - Issue type: Feature request

### Short description

Right now, requesting zones/:zone_id will also return all RRSets in the zone, which can be slow on large zones. If I only want the 'meta' information, I don't need these.

My suggestion is to add an `rrset` URL parameter that can be set to `false` to not include the RRsets in the response.

A query would look like: `GET /api/v1/servers/localhost/zones/example.com.?rrsets=false HTTP/1.1`

pieterlexis:
### Short description
This PR adds zone rectification to the API. Work done:

 - Move the pdnsutil rectification code to the DNSSECKeeper
 - Generate DNSSEC keys for a zone when "dnssec" is true in an API POST/PATCH for zones
 - Rectify DNSSEC zones after POST/PATCH when API-RECTIFY metadata is 1
 - Allow setting this metadata via the "api-rectify" param in a Zone object
 - Show "nsec3param" and "nsec3narrow" in Zone API responses
 - Add an "rrsets" request parameter for a zone to skip sending RRSets in the response (Closes #5712)
 - Add checkNSEC3PARAM function
 - Add rectify endpoint in the API

This PR takes a lot of ideas and code from #3417 and subsequent development.

Closes #3417

Many thanks to Nils Wisiol (@nils-wisiol) for the initial implementation.

### Checklist

I have:
- [ ] documented the code


nils-wisiol:
@pieterlexis, thank you for completing this! :+1:

peterthomassen:
@pieterlexis Does this also allow setting NSEC3 settings?

pieterlexis:
> @pieterlexis Does this also allow setting NSEC3 settings?

[yes](https://github.com/PowerDNS/pdns/pull/5779/files#diff-b358b714dc43e45c0ca862ec9507a8b8R636)

pdns/dbdnsseckeeper.cc
```diff
@@ -558,6 +559,181 @@ bool DNSSECKeeper::getTSIGForAccess(const DNSName& zone, const string& master, D
   return false;
 }
 
+bool DNSSECKeeper::rectifyZone(const DNSName& zone, string& error) {
+  if (isPresigned(zone)) {
+    error =  "Rectify presigned zone '"+zone.toLogString()+"' is not allowed/necessary.";
+    return false;
+  }
+
+  UeberBackend B("default");
```
zeha: DNSSECKeeper is supposed to have an UeberBackend instance already, so can you reuse that?

docs/http-api/zone.rst
```diff
@@ -37,24 +38,19 @@ Comments are per-RRset.
 
   Switching ``dnssec`` to ``true`` (from ``false``) sets up DNSSEC signing
   based on the other flags, this includes running the equivalent of
-  ``secure-zone`` and ``rectify-zone``. This also applies to newly created
-  zones. If ``presigned`` is ``true``, no DNSSEC changes will be made to
-  the zone or cryptokeys.
-
-  ``dnssec``, ``nsec3narrow``, ``nsec3param``, ``presigned`` are not yet implemented.
+  ``secure-zone`` and ``rectify-zone`` (if ``api_rectify`` is set to "1").
+  This also applies to newly created zones. If ``presigned`` is ``true``,
```
zeha: the mismatch of 1 and true is quite ugly here. I think API consumers should not carry the burden of dealing with different truthiness values

pdns/ws-auth.cc
```diff
@@ -497,27 +520,142 @@ static void gatherComments(const Json container, const DNSName& qname, const QTy
...
+
+        int k_algo = DNSSECKeeper::shorthand2algorithm(::arg()["default-ksk-algorithm"]);
+        int z_algo = DNSSECKeeper::shorthand2algorithm(::arg()["default-zsk-algorithm"]);
+        int k_size = arg().asNum("default-ksk-size");
+        int z_size = arg().asNum("default-zsk-size");
+
+        if (k_algo != -1) {
+          int64_t id;
+          if (!dk.addKey(zonename, true, k_algo, id, k_size)) {
+            throw ApiException("No backend was able to secure '" + zonename.toString() + "', most likely because no DNSSEC"
```
zeha: This message is duplicated at least three times.

Also I can't say I like these super-long messages that suggest backend-specific config settings...

pdns/ws-auth.cc
```diff
@@ -497,27 +520,142 @@ static void gatherComments(const Json container, const DNSName& qname, const QTy
...
+    string error_msg = "";
+    if (!isDNSSECZone) {
+      throw ApiException("NSEC3PARAMs provided for zone '"+zonename.toString()+"', but zone is not DNSSEC secured.");
+    }
+    if (!dk.checkNSEC3PARAM(ns3pr, error_msg)) {
+      throw ApiException("NSEC3PARAMs provided for zone '"+zonename.toString()+"' are invalid. " + error_msg);
+    }
+    if (!dk.setNSEC3PARAM(zonename, ns3pr, boolFromJson(document, "nsec3narrow", false))) {
+      throw ApiException("NSEC3PARAMs provided for zone '" + zonename.toString() +
+          "' passed our basic sanity checks, but cannot be used with the current backend.");
```
zeha: s/current/hosting/?

pdns/ws-auth.cc
```diff
@@ -1510,6 +1687,16 @@ static void patchZone(HttpRequest* req, HttpResponse* resp) {
...
+  DNSSECKeeper dk;
+  string api_rectify;
+  di.backend->getDomainMetadataOne(zonename, "API-RECTIFY", api_rectify);
+  if (dk.isSecuredZone(zonename) && !dk.isPresigned(zonename) && api_rectify == "1") {
+    string error_msg = "";
+    if (!dk.rectifyZone(zonename, error_msg))
+      throw ApiException("Failed to rectify '" + zonename.toString() + "' " + error_msg);
+  }
+
   di.backend->commitTransaction();
```
zeha: dk.rectifyZone calls commitTransaction too, right? Maybe dk.rectifyZone shouldn't deal with backend lookups and transactions at all.

pdns/ws-auth.cc
nils-wisiol: I believe current is the better word here; "hosting" over-specifies the backend and suggests there is more than one backend at a time.

docs/http-api/zone.rst
pieterlexis: I think I do make a bool out of, but it is wrongly documented

zeha: the code below indeed confused me. please check whats actually there and i think boolifying it is a good idea (:

pdns/ws-auth.cc
pieterlexis: That would mean that all callers are responsible for this? I would prefer dk.rectifyZone() to do some sanity checking and perform adding the ENTs in a safe way

zeha: Probably. But having the transaction being committed underneath you is also super surprising!

zeha: Probably. But it'd be great if the API could modify *and* rectify the zone in one TX, so no unrectified stuff is ever served by the DNS parts...

pieterlexis: that is a sexy idea......

docs/http-api/zone.rst
pieterlexis: The POST/PUT and GET for `api_rectify` were not consistent. It is fully boolified on the API level in the upcoming push.

pdns/dbdnsseckeeper.cc
pieterlexis: Will do!

pdns/ws-auth.cc
pieterlexis: Will be implemented

pieterlexis:
@zeha I think the new commits will satisfy your comments?

pdns/dbdnsseckeeper.cc
pieterlexis: Interestingly, re-using the UeberBackend in the DNSSECKeeper makes the bind-dnssec-both test fail because `getSOAUncached()` returns false for secure-delegated.dnssec-parent.com. Investigating

pieterlexis: confirmed, reverting 1d7630f631b1e81373d158021a9eb1daa9442762 makes all the tests pass.

zeha: That sounds reaaaally bad. Like something keeping a list() operation open or sth.

pieterlexis: ah, I think I got it. In pdnsutil.cc:1983 we create a DNSSECKeeper object _without_ handing it an UeberBackend. So when rectifyZone (in pdnsutil) is called we indeed can not get a SOA from that dk.

pieterlexis: so this is a bug(ish) in pdnsutil, but it does mean that calling `DNSSECKeeper::rectifyZone` means we need the Ueber backend. Maybe it indeed makes more sense to move rectifyZone to the UeberBackend (but that has other pain points)

pieterlexis: Alright, I have made a solution that uses the existing UeberBackend if it works and allocates a new one when needed. C++ wizards are welcome to make this better......

PR #5788, opened 2017-10-07
docs: note default for auth server-id

jpmens:
note default for auth server-id

### DIFF
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -1007,7 +1007,7 @@ this to an empty string disables secpoll.
 -  String
 -  Default: The hostname of the server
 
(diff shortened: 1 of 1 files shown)


docs/settings.rst
```diff
@@ -1007,7 +1007,7 @@ this to an empty string disables secpoll.
 -  String
 -  Default: The hostname of the server
 
-This is the server ID that will be returned on an EDNS NSID query.
+This is the server ID that will be returned on an EDNS NSID query. It defaults to the server's hostname.
```
zeha: this is already mentioned two lines above?

jpmens: If only I could read ...

PR #5799, opened 2017-10-09
Fix shadowed variables

rgacogne:
### Short description

Let's try not to shadow variable by declaring a new variable with the same name in the scope of the previous one. Fix a bug in the way `isTCPSocketUsable()` handles `EINTR` errors.
Reported by GCC's `-Wshadow`.

### Checklist

I have:
- [ ] documented the code


rgacogne:
> I don't see any obvious issues with un-shadowing these vars. We should consider running in travis with -Wshadow -Werror to prevent stuff like this

I think we should:
* enable `-Wshadow` by default, even outside travis
* since `-Werror` is super annoying, especially on old compilers, see if the versions of gcc and clang we have on travis supports `-Werror=shadow` which turns only warnings about shadowed variables into error, not all warnings.

PR #5803, opened 2017-10-10
Fix throwing exceptions from MThreads, plus add unit tests

ahupowerdns:
### Short description
In MTasker, MThreads can throw exceptions, which then get caught by the 'kernel' thread. This mechanism was faulty in our 'old boost' support, leading to crashes in std::rethrow_exception. 
This PR fixes the 'old boost' support, plus adds a testcase for throwing & catching such exceptions. In addition, this simple testcase found a bug in MTasker with timeoutless waiters. This has also been solved.
### Checklist

I have:
- [ ] documented the code

### DIFF
--- a/pdns/mtasker_fcontext.cc
+++ b/pdns/mtasker_fcontext.cc
@@ -152,7 +152,7 @@ threadWrapper (transfer_t const t) {
 #if BOOST_VERSION < 106100
     jump_fcontext (reinterpret_cast<fcontext_t*>(&ctx->uc_mcontext),
                    static_cast<fcontext_t>(next_ctx),
-                   static_cast<bool>(ctx->exception));
+                   reinterpret_cast<intptr_t>(ctx));
 #else
     jump_fcontext (static_cast<fcontext_t>(next_ctx), 0);
 #endif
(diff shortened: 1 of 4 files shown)


pdns/recursordist/Makefile.am
```diff
@@ -206,6 +206,7 @@ testrunner_SOURCES = \
 	ixfr.cc ixfr.hh \
 	logger.cc logger.hh \
 	misc.cc misc.hh \
+	mtasker_fcontext.cc \
```
rgacogne: It should be mtasker_context.cc instead

pdns/mtasker_fcontext.cc
```diff
@@ -189,10 +189,12 @@ pdns_swapcontext
...
-    if (jump_fcontext (reinterpret_cast<fcontext_t*>(&octx.uc_mcontext),
-                       static_cast<fcontext_t>(ctx.uc_mcontext), 0)) {
-        std::rethrow_exception (ctx.exception);
-    }
+    intptr_t ptr = jump_fcontext(reinterpret_cast<fcontext_t*>(&octx.uc_mcontext),
+                                 static_cast<fcontext_t>(ctx.uc_mcontext), 0);
+
+    auto origctx = reinterpret_cast<pdns_ucontext_t*>(ptr);
+    if(origctx && origctx->exception)
+        std::rethrow_exception (origctx->exception);
```
rgacogne: So, I understand why your fix works but I don't get why this didn't work in the first place. Shouldn't `origctx` be equal to `ctx` here, after `jump_fcontext()` returns, or am I getting this backward once again?

ahupowerdns: so I tried that first, and printed the addresses of 'ctx', and they are different from the address of ctx in the MThread. I tried ctx and octx both. I admit this is somewhat of a "montecarlo programming" fix. There is a little bit of a dance around what 'ctx' means, and even the new boost code has to do some work to extract the ctx we returned from. Perhaps you don't know that when you do the jump_fcontext?

rgacogne: Ah yes, I remember something like that, we can have different `pdns_ucontext_t` and of course the `fcontext_t` keeps changing.

PR #5808, opened 2017-10-11
rec: Check that the NSEC covers an ENT when looking for NODATA

rgacogne:
### Short description

Otherwise we might consider that a NSEC record covers a name when it does not.

### Checklist

I have:
- [ ] documented the code

### DIFF
--- a/pdns/validate.cc
+++ b/pdns/validate.cc
@@ -42,6 +42,15 @@ static bool isCoveredByNSEC(const DNSName& name, const DNSName& begin, const DNS
           (begin == next && name != begin));                        // "we have only 1 NSEC record, LOL!"
 }
 
+static bool nsecProvesENT(const DNSName& name, const DNSName& begin, const DNSName& next)
+{
+  /* if name is an ENT:
+     - begin < name
+     - next is a child of name
+  */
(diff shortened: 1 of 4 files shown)


pdns/validate.cc
```diff
@@ -95,10 +104,7 @@ bool denialProvesNoDelegation(const DNSName& zone, const std::vector<DNSRecord>&
   return false;
 }
 
-// FIXME: needs a zone argument, to avoid things like 6840 4.1
-// FIXME: Add ENT support
-// FIXME: Make usable for non-DS records and hook up to validateRecords (or another place)
-dState getDenial(const cspmap_t &validrrsets, const DNSName& qname, const uint16_t qtype, bool referralToUnsigned)
+dState getDenial(const cspmap_t &validrrsets, const DNSName& qname, const uint16_t qtype, bool referralToUnsigned, bool wantsNoDataProof)
```
pieterlexis: I am not a fan of telling getDenial what we are looking for. I think this function should always return the correct dState. Although this might be a project for later.

rgacogne: I completely agree, but we will need to refactor our handling of DNSSEC denials for that.

Issue #5807, opened 2017-10-11
Auth: b2b-migrate from BIND to gsqlite does not migrate 'master' servers


pieterlexis:
- Program: Authoritative
 - Issue type: Bug report

### Short description
While migrating from the BIND backend to the gsqlite3 backend, the IP addresses of the masters for slave zones are not added to the database's "masters" column in the "domains" table.

### Environment
 - Operating system: Debian Stretch
 - Software version: Master (2b55d75)
 - Software source: repo.powerdns.com

### Steps to reproduce
1. Have a slave zone in the bind backend:
```
zone "powerdns.space" in {
  type slave;
  file "/var/lib/powerdns/zones/powerdns.space.zone";
  masters {
    188.166.104.87;
    2a03:b0c0:2:d0::4af:b001;
  };
};
```
2. Run `pdnsutil b2b-migrate bind gsqlite3` (after configging both backends in pdns.conf)
3. Observe no masters in the domains table:
```
$ sqlite3 -column -header /var/lib/powerdns/pdns-public.sqlite3 "select * from domains where name='powerdns.space'"
id          name            master      last_check  type        notified_serial  account   
----------  --------------  ----------  ----------  ----------  ---------------  ----------
7           powerdns.space                          SLAVE                                  
```

### Expected behaviour
That the master IP addresses show up in the table.

### Actual behaviour
That the addresses did not show up.

PR #5810, opened 2017-10-11
BIND backend: Several improvements for b2b-migrate

pieterlexis:
### Short description
While attempting to do a migration from the BIND backend to gsqlite3, several issues arose:

* No masters were set in the target db (#5807)
* Only the last master in the list of masters would be added to the target database
* The BIND backend was not fully aware of native zones

The last issue stemmed from an incomplete implementation in #5115.

This PR aims to fix all these issues.

Note: the b2b-migrate patch for setting the masters correctly and the patch to add the masters to the DomainInfo might need to be backported to 4.0.x.

### Checklist

I have:
- [ ] documented the code

### DIFF
--- a/modules/bindbackend/bindbackend2.cc
+++ b/modules/bindbackend/bindbackend2.cc
@@ -302,7 +302,7 @@ void Bind2Backend::getUpdatedMasters(vector<DomainInfo> *changedDomains)
     ReadLock rl(&s_state_lock);
 
     for(state_t::const_iterator i = s_state.begin(); i != s_state.end() ; ++i) {
-      if(!i->d_masters.empty() && this->alsoNotify.empty() && i->d_also_notify.empty())
+      if(i->d_kind != DomainInfo::Master)
         continue;
 
       DomainInfo di;
@@ -352,7 +352,7 @@ void Bind2Backend::getAllDomains(vector<DomainInfo> *domains, bool include_disab
       di.id=i->d_id;
       di.zone=i->d_name;
       di.last_check=i->d_lastcheck;
-      di.kind=i->d_masters.empty() ? DomainInfo::Master : DomainInfo::Slave; //TODO: what about Native?
+      di.kind=i->d_kind;
       di.masters=i->d_masters;
       di.backend=this;
       domains->push_back(di);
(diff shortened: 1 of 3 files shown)


modules/bindbackend/bindbackend2.cc
```diff
@@ -353,6 +353,7 @@ void Bind2Backend::getAllDomains(vector<DomainInfo> *domains, bool include_disab
       di.zone=i->d_name;
       di.last_check=i->d_lastcheck;
       di.kind=i->d_masters.empty() ? DomainInfo::Master : DomainInfo::Slave; //TODO: what about Native?
```
cmouse: `i->d_kind` instead?

modules/bindbackend/bindbackend2.cc
```diff
@@ -302,7 +302,8 @@ void Bind2Backend::getUpdatedMasters(vector<DomainInfo> *changedDomains)
     ReadLock rl(&s_state_lock);
 
     for(state_t::const_iterator i = s_state.begin(); i != s_state.end() ; ++i) {
-      if(!i->d_masters.empty() && this->alsoNotify.empty() && i->d_also_notify.empty())
+      if(i->d_kind != DomainInfo::Slave ||
+          (!i->d_masters.empty() && this->alsoNotify.empty() && i->d_also_notify.empty()))
```
cmouse: should these be removed or are Master domains with masters considered slaves?

pieterlexis:
paging @mind04 for a review

modules/bindbackend/bindbackend2.cc
pieterlexis: Good question, I'd have to check what the expected behaviour is.

pieterlexis: In the gsqlite3 backend, we `SELECT ... where type='MASTER'`. This also means that alsoNotify and d_masters don't even matter. We should not use NATIVE or SLAVE zones here to make the BIND backend behave consistent with the gsql backends. I'll make that so.

mind04:
Two questions:

1. Was the interaction between master and slave checked? (notifies and freshness (soa) checks still working)
2. Is the behaviour of bind backed still compatible with real bind after this change for master/slave zones and also-notify?

pieterlexis:
> Was the interaction between master and slave checked? (notifies and freshness (soa) checks still working)

Yes, this was tested and it works

> Is the behaviour of bind backed still compatible with real bind after this change for master/slave zones and also-notify?

The behaviour indeed differs from before (now that I re-read this code), fixing.

PR #5842, opened 2017-10-25
4.1 polish: Add log-timestamp option

pieterlexis:
### Short description
This option can be used to disable printing timestamps to stdout, this is useful when using systemd-journald or another supervisor that timestamps stdout by itself. As the logs will not have 2 timestamps.

### Checklist

I have:
- [ ] documented the code

### DIFF
--- a/pdns/logger.cc
+++ b/pdns/logger.cc
@@ -47,14 +47,16 @@ void Logger::log(const string &msg, Urgency u)
 #ifndef RECURSOR
   bool mustAccount(false);
 #endif
-  struct tm tm;
-  time_t t;
-  time(&t);
-  tm=*localtime(&t);
-
   if(u<=consoleUrgency) {
-    char buffer[50];
-    strftime(buffer,sizeof(buffer),"%b %d %H:%M:%S ", &tm);
+    char buffer[50] = "";
+    if (d_timestamps) {
+      struct tm tm;
+      time_t t;
+      time(&t);
+      tm=*localtime(&t);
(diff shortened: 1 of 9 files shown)


pdns/logger.cc
```diff
@@ -47,17 +47,20 @@ void Logger::log(const string &msg, Urgency u)
...
-    char buffer[50];
-    strftime(buffer,sizeof(buffer),"%b %d %H:%M:%S ", &tm);
+    if (d_timestamps) {
+      char buffer[50] = "";
+      struct tm tm;
+      time_t t;
+      time(&t);
+      tm=*localtime(&t);
+      strftime(buffer,sizeof(buffer),"%b %d %H:%M:%S ", &tm);
+      clog << string(buffer);
```
rgacogne: This write is unlocked, perhaps it would be easier to declare buffer out of the `if (d_timestamps) {` block, initialize it to an empty string like it's done in your code, and keep the existing `clog << string(buffer) + msg <<endl;`?

pieterlexis: nice catch, I was hoping to avoid initializing the buffer for every log line if unneeded, maybe already lock before `if (d_timestamps)`?

rgacogne: That would mean calling `localtime()` and `strftime()` while holding the lock, which we should try very hard to avoid. A 50 bytes buffer on the stack is cheap and initializing it to "" is just setting the first byte to `\0`, so IMHO that's a better choice.

pieterlexis: ack

PR #5876, opened 2017-10-30
--trace logging with more details about trust anchors, plus ttl of auth data

ahupowerdns:
Print more details of trust anchors. In addition, the --trace output that mentions if data from authoritative servers gets accepted now also prints the TTL and clarifies the 'place' number previously printed.

### Checklist

I have:
- [ ] documented the code

### DIFF
--- a/pdns/syncres.cc
+++ b/pdns/syncres.cc
@@ -1440,19 +1440,23 @@ vState SyncRes::getTA(const DNSName& zone, dsmap_t& ds)
   auto luaLocal = g_luaconfs.getLocal();
 
   if (luaLocal->dsAnchors.empty()) {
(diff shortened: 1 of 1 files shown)


pieterlexis:
Maybe we should have a static vector for the names of the places (like `vState` and `vStates`) so we can show an actual name?

```cpp
.... <<DNSPacketSections[rec.d_place]<<....
```

ahupowerdns:
I'm actually hoping for a slightly more generic solution. I know there are classes out there that create fast and simple mapping between enums and names. We have this issue for QType, RCode, validation states, record places, opcodes etc. https://stackoverflow.com/questions/28828957/enum-to-string-in-modern-c11-c14-and-future-c17-c20 has some discussion.

pieterlexis:
alright, let's make that an item for the 4.2 or 4.3 timeframe

PR #5877, opened 2017-10-30
rec: Sort NS addresses by speed, remove old ones

### LINKED ISSUE #1066: Odd NS selection after updates to an NS's IP address(es)
There is an issue in pdns's NS selection when the IP address(es) for a given NS change. pdns will send all recursive queries to the updated NS, regardless of whether or not a better NS is available. The issue seems to persist until pdns is restarted.

We ran across this issue while examining traffic captures for the d-root server during the changeover process which occurred earlier this year (Jan 3rd 2013). We then recreated it in an emulated environment (https://www.emulab.net/), emulating 13 root servers, a replacement d-root server, and a test server running pdns_recursor with a script to generate queries with invalid TLDs (guaranteed to be sent to a root server).

Though the experiment was specifically interested in testing root servers, I believe the same issue appears for non-root servers.

The timeline of events:
1. pdns working normally.
2. d-root IP address changes; all root servers begin returning a new IP address for d.root-servers.net
3. pdns still operates as in step 1, until the next priming query (every 2 hours).
4. Shortly after pdns re-primes the root server cache (and gets the new IP address), it begins sending all root server traffic to the new d-root IP address.

In the emulated network, the recursor had controlled latencies to each of the root servers. Of note, h-root (which should be prefered as the lowest latency server) had a latency of ~11ms, while both the new and the old d-root had around ~100ms. The issue causes pdns to prefer the 100ms server, leading to an overall decrease in performance for those queries.

Here is a graph of the recorded network traffic:

reqs

Here is another graph zoomed in on the moment when pdns starts acting incorrectly:

reqs_zoomed

And here is a large (89M compressed, 1.2G uncompressed) debug log from the experimental run: http://www.cs.umd.edu/~gjp/pdns_ns_selection_log.tar.gz

I looked into the pdns source and I believe this is the sequence of events leading to this issue (these same events happen independently for each thread):
1. houseKeeping thread issues a priming query every 2 hours.
2. The priming query response returns the new d-root IP address, which replaces the cached IP address for the d.root-servers.net NS. The DEWMA collection for d.root-servers.net still contains the old IP address and its decayed latency.
3. pdns selects which NS to use for a query by sorting the NSs according to the lowest latency associated with the NS (in the NS's DEWMA collection).
4. The latency for the old d-root IP address continues to decay until it is the lowest latency, at which point pdns chooses to use d.root-servers.net for the current query.
5. pdns lookups the IP address for d.root-servers.net in the resolver cache, which returns the new d-root IP address (as updated in step 2).
6. Once the query completes, pdns submits the latency to the DEWMA collection for d.root-servers.net, using the new d-root IP address. The old d-root IP address and latency are still in the DEWMA collection, but the latency is not updated (and continues to decay).
7. Because the latency for the old d-root IP address continues to decay, pdns continues to select d.root-servers.net for root zone queries.

rgacogne:
### Short description

We used to not sort the different addresses we had for a given NS by speed, only taking care of placing the first one in front. However we also didn't remove existing entries for a given NS, meaning that if a given IP stopped being advertised it would stay in our NS speeds map and keep being used to determine the fastest NS, even if we would only send queries to the new IPs after the selection. Since we didn't send any query to the old IP anymore, its latency would only keep decaying meaning the computed latency of the corresponding NS would only keep decreasing, completely uncorrelated from its real latency.

This commit removes old entries from the NS speeds map if they are no longer present when we refresh the addresses of a given NS. In addition, it orders all NS IPs by decaying latency, meaning new ones will have a fair chance of being picked up.

It might fix #1066, although it did not test that exact case.

Don't merge without a careful review!

### Checklist

I have:
- [ ] documented the code

### DIFF
--- a/pdns/rec_channel_rec.cc
+++ b/pdns/rec_channel_rec.cc
@@ -215,7 +215,16 @@ string doDumpNSSpeeds(T begin, T end)
   try {
     total = broadcastAccFunction<uint64_t>(boost::bind(pleaseDumpNSSpeeds, fd));
   }
-  catch(...){}
+  catch(std::exception& e)
+  {
+    close(fd);
+    return "error dumping NS speeds: "+string(e.what())+"\n";
+  }
+  catch(PDNSException& e)
+  {
+    close(fd);
+    return "error dumping NS speeds: "+e.reason+"\n";
+  }
 
   close(fd);
   return "dumped "+std::to_string(total)+" records\n";
--- a/pdns/syncres.cc
+++ b/pdns/syncres.cc
@@ -368,7 +368,9 @@ uint64_t SyncRes::doDumpNSSpeeds(int fd)
   for(const auto& i : t_sstorage.nsSpeeds)
   {
     count++;
-    fprintf(fp, "%s -> ", i.first.toString().c_str());
+
+    // an <empty> can appear hear in case of authoritative (hosted) zones
+    fprintf(fp, "%s -> ", i.first.toLogString().c_str());
     for(const auto& j : i.second.d_collection)
     {
       // typedef vector<pair<ComboAddress, DecayingEwma> > collection_t;
@@ -685,7 +687,6 @@ vector<ComboAddress> SyncRes::getAddrs(const DNSName &qname, unsigned int depth,
 
   if(ret.size() > 1) {
     map<ComboAddress, double> speeds;
-
     auto& collection = t_sstorage.nsSpeeds[qname].d_collection;
     for(const auto& val: ret) {
       double speed;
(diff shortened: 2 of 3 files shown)


giganteous:
Observing build a88e082 's behaviour in a simulated environment still gives preference to the new IP for a root nameserver.

```shell
rec_control dump-nsspeeds /tmp/out && grep '^[\;dh]' /tmp/out
dumped 39 records
; nsspeed dump from thread follows
;
d.root-servers.net -> 10.0.1.4/398.325378 
h.root-servers.net -> 10.0.1.8/701.068420 
; nsspeed dump from thread follows
;
d.root-servers.net -> 10.0.1.4/61.136986 10.0.2.4/201549.312500 
h.root-servers.net -> 10.0.1.8/61.137657 
; nsspeed dump from thread follows
;
d.root-servers.net -> 10.0.1.4/60.774120 10.0.2.4/200368.500000 
h.root-servers.net -> 10.0.1.8/61.066288 
```

It really is storing the wrong value for 10.0.1.4 somehow, as I'm measuring >200 msec responses with dig in this environment.

PR #5880, opened 2017-10-30
Add configuration option to disable IP_BIND_ADDRESS_NO_PORT.

dmccombs:
### Short description
This adds the configuration option `ipBindAddrNoPort` for downstream servers to optionally disable `IP_BIND_ADDRESS_NO_PORT` if `dnsdist` was built with support for it. It defaults to `true` to maintain current behavior by default.

This addresses the problem when `dnsdist` was built on a kernel newer than 4.2, but is running (in my case, in a container) on a system with an older kernel version which does not support `IP_BIND_ADDRESS_NO_PORT`, causing TCP connections to immediately fail.

### Checklist
I have:
- [X] documented the code

### DIFF
--- a/pdns/dnsdistdist/docs/reference/config.rst
+++ b/pdns/dnsdistdist/docs/reference/config.rst
@@ -224,7 +224,7 @@ Servers
       tcpSendTimeout=NUM,    -- The timeout (in seconds) of a TCP write attempt
       tcpRecvTimeout=NUM,    -- The timeout (in seconds) of a TCP read attempt
       tcpFastOpen=BOOL,      -- Whether to enable TCP Fast Open
-      ipBindAddrNoPort=BOOL, -- Whether to enable IP Bind Address No Port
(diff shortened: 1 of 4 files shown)


pdns/dnsdistdist/docs/reference/config.rst
```diff
@@ -224,6 +224,7 @@ Servers
       tcpSendTimeout=NUM,    -- The timeout (in seconds) of a TCP write attempt
       tcpRecvTimeout=NUM,    -- The timeout (in seconds) of a TCP read attempt
       tcpFastOpen=BOOL,      -- Whether to enable TCP Fast Open
+      ipBindAddrNoPort=BOOL, -- Whether to enable IP Bind Address No Port
```
rgacogne: "Whether to enable IP_BIND_ADDRESS_NO_PORT if available, default: true"?

dmccombs: Thanks @rgacogne. Updated with that documentation change.

Issue #4436, opened 2016-09-07
recursor: add option to validate dnssec and log bogus but do not send servfail to client


CaptainQwark:
It would be nice if it was possible to let the recursor validate DNSSEC and log bogus validations, but never actually send a SERVFAIL to any client (even those who have the AD bit set).

use case:
For an ISP it's crucial to know what the impact will be of turning on validation **before** actually turning it on and have you customers impacted by failing dns queries.
By just validating and logging without servfails we can see what domains that our customers request are bogus and:
- validate that the recursor is correct about the bogus validation
- address the issue with the domain admins of the failing domains to get things fixed before we start validating (I know there will always be crappy configured domains, but anything helps)
- determine if it's feasible at any point in time to turn on validation for our customers

pieterlexis:
So I am pondering the following 3 options to replace `dnssec` and `dnssec-log-bogus`:

- `dnssec-validation`
    - `yes` -- Always validate
    - `client` -- Validate when the client requests it
    - `no` -- Disable DNSSEC validation
- `dnssec-servfail`
    - `yes` -- Always SERVFAIL on bogus
    - `client` -- Only when the client requests it
    - `no` -- Never send a SERVFAIL on bogus
- `dnssec-log`
  - `yes` or `all` -- Log all validation results
  - `bogus` -- Log all bogus results
  - `no` -- Don't log validation results

This will allow for greater operator flexibility without having too many options to be confusing.

pieterlexis:
Alternatively, the following might also work, but might also be pretty confusing:

- `dnssec`
    - `yes` -- Always validate and SERVFAIL
    - `yes-client` -- Always validate and SERVFAIL if the client asks for DNSSEC
    - `yes-nofail` -- Always validate but don't SERVFAIL
    - `client` -- Only validate when the client asks for it
    - `client-nofail` -- Validate when the client asks for it, but don't SERVFAIL
    - `no` -- Disable DNSSEC validation
- `dnssec-log`
  - `yes` or `all` -- Log all validation results
  - `bogus` -- Log all bogus results
  - `no` -- Don't log validation results

CRtEurope:
The first proposal from pieterlexis is better and not confusing at all. The second one, hmmm, is confusing.

CaptainQwark:
+1 for the first proposal. The only downside may be that it breaks current dnssec configurations.

paddg:
Yes, first proposal sounds good! Maybe useless combinations will be possible:
[...]

PR #5891, opened 2017-11-01
[WIP] Split the DNSSEC mode, logging and SERVFAIL config

pieterlexis:
### Short description
This should implement the suggestions in #4436 and allow operators to
control their log volumes when e.g. only validation results are
required.

Not that running with `--trace` will also print the DNSSEC traces.

I open this PR for feedback now, as we have to decide if it goes in 4.1.0 or 4.2.0. What is really missing here, is the documentation.

### Checklist

I have:
- [ ] documented the code

### DIFF
--- a/pdns/pdns_recursor.cc
+++ b/pdns/pdns_recursor.cc
@@ -770,7 +770,7 @@ static void startDoResolve(void *p)
       sr.setLuaEngine(t_pdl);
     }
     sr.d_requestor=dc->d_remote; // ECS needs this too
-    if(g_dnssecmode != DNSSECMode::Off) {
+    if(g_dnssecValidationMode != DNSSECValidationMode::Off) {
       sr.setDoDNSSEC(true);
 
       // Does the requestor want DNSSEC records?
@@ -782,7 +782,8 @@ static void startDoResolve(void *p)
       // Ignore the client-set CD flag
       pw.getHeader()->cd=0;
     }
-    sr.setDNSSECValidationRequested(g_dnssecmode == DNSSECMode::ValidateAll || g_dnssecmode==DNSSECMode::ValidateForLog || ((dc->d_mdp.d_header.ad || DNSSECOK) && g_dnssecmode==DNSSECMode::Process));
+    sr.setDNSSECValidationRequested(g_dnssecValidationMode == DNSSECValidationMode::Validate ||
+        ((dc->d_mdp.d_header.ad || DNSSECOK) && g_dnssecValidationMode == DNSSECValidationMode::ClientOnly));
 
 #ifdef HAVE_PROTOBUF
     sr.setInitialRequestId(dc->d_uuid);
@@ -1012,7 +1013,7 @@ static void startDoResolve(void *p)
       dc=0;
       return;
     }
-    if(tracedQuery || res == -1 || res == RCode::ServFail || pw.getHeader()->rcode == RCode::ServFail)
+    if(tracedQuery || g_dnssecLogMode == DNSSECLogMode::Trace || res == -1 || res == RCode::ServFail || pw.getHeader()->rcode == RCode::ServFail)
     { 
       string trace(sr.getTrace());
       if(!trace.empty()) {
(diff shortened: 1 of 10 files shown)


pdns/pdns_recursor.cc
```diff
@@ -1042,30 +1043,34 @@ static void startDoResolve(void *p)
...
               L<<Logger::Warning<<"Answer to "<<dc->d_mdp.d_qname<<"|"<<QType(dc->d_mdp.d_qtype).getName()<<" for "<<dc->d_remote.toStringWithPort()<<" validates as Bogus"<<endl;
             }
             
             // Does the query or validation mode sending out a SERVFAIL on validation errors?
-            if(!pw.getHeader()->cd && (g_dnssecmode == DNSSECMode::ValidateAll || dc->d_mdp.d_header.ad || DNSSECOK)) {
-              if(sr.doLog()) {
+            if(!pw.getHeader()->cd &&
+                ((g_dnssecBogusServfailMode == DNSSECBogusServfailMode::ClientOnly && (dc->d_mdp.d_header.ad || DNSSECOK)) ||
+                g_dnssecBogusServfailMode == DNSSECBogusServfailMode::On)) {
+              if(sr.doLog() || shouldLog || g_dnssecLogMode == DNSSECLogMode::BogusOnly) {
```
rgacogne: `sr.doLog()` seems redundant here.

pdns/pdns_recursor.cc
```diff
@@ -3268,8 +3294,9 @@ int main(int argc, char **argv)
     ::arg().set("local-address","IP addresses to listen on, separated by spaces or commas. Also accepts ports.")="127.0.0.1";
     ::arg().setSwitch("non-local-bind", "Enable binding to non-local addresses by using FREEBIND / BINDANY socket options")="no";
     ::arg().set("trace","if we should output heaps of logging. set to 'fail' to only log failing domains")="off";
-    ::arg().set("dnssec", "DNSSEC mode: off/process-no-validate (default)/process/log-fail/validate")="process-no-validate";
-    ::arg().set("dnssec-log-bogus", "Log DNSSEC bogus validations")="no";
+    ::arg().set("dnssec-validation", "DNSSEC validation mode: yes/no/client")="no";
```
rgacogne: Shouldn't the default be at least `process`?

pdns/rec_channel_rec.cc
```diff
@@ -370,30 +370,39 @@ string doSetCarbonServer(T begin, T end)
...
-    return "No DNSSEC Bogus logging setting specified\n";
+    return "No DNSSEC validation logging setting specified\n";
 
   if (pdns_iequals(*begin, "on") || pdns_iequals(*begin, "yes")) {
-    if (!g_dnssecLogBogus) {
+    if (g_dnssecLogMode != DNSSECLogMode::On) {
       L<<Logger::Warning<<"Enabling DNSSEC Bogus logging, requested via control channel"<<endl;
-      g_dnssecLogBogus = true;
+      g_dnssecLogMode = DNSSECLogMode::On;
       return "DNSSEC Bogus logging enabled\n";
```
rgacogne: The two messages above mention "Bogus" logging but it will enable the logging of all validation results.

pdns/syncres.cc
```diff
@@ -1099,29 +1100,31 @@ bool SyncRes::doCacheCheck(const DNSName &qname, const QType &qtype, vector<DNSR
...
 
         if (cachedState != Indeterminate) {
-          LOG(prefix<<qname<<": got Indeterminate state from the cache, validation result is "<<vStates[cachedState]<<endl);
+          DNSSECLOG(prefix<<qname<<": got Indeterminate state from the cache, validation result is "<<vStates[cachedState]<<endl);
           t_RC->updateValidationStatus(d_now.tv_sec, sqname, sqt, d_incomingECSFound ? d_incomingECSNetwork : d_requestor, d_requireAuthData, cachedState);
         }
       }
     }
 
+    DNSSECLOG(prefix<<qname<<" DNSSEC validated the following record content:"<<endl);
```
rgacogne: This message seems misleading, the content might not have been validated.

pieterlexis: Can you explain this some more. I might have misinterpreted the code here.

rgacogne: If I'm not mistaken we will log the content of every record retrieved from the cache, be it validated or not. It might not have been validated if validation was not requested, or the cached data was not authoritative, for example.

pieterlexis: ok, then this should not be DNSSECLOG but LOG (with a better message)

PR #5893, opened 2017-11-01
auth: Warn about algo 5 -> 7 upgrade in log

pieterlexis:
### Short description
One step toward fixing #3267 in 4.2.0

### Checklist

I have:
- [ ] documented the code


mind04:
Warn in pdnsutil show-zone / check-zone, not in the server logs
Imagine what will happen with 100000 zones like this and a 30 seconds (default) key-cache-ttl.

pieterlexis:
We already warn about this in pdnsutil (#4123), but not all users take our advice and run `check-all-zones` on new releases. Warning them in the log will make them aware they need to fix something.

I also think now many users have enabled NSEC3 for algo 5 zones.

Issue #4616, opened 2016-10-24
remove `recursor=` setting from auth?


Habbie:
This is a discussion ticket.

At least once a week, we talk to a user that is confused about `recursor` because it is not doing what he/she expects. Meanwhile we/I feel that there are very, very few legitimate use cases for `recursor` that cannot be covered by reversing the setup (putting the Recursor in front) or deploying dnsdist. So, some of us would like to entirely remove the `recursor` feature.

If you disagree, please describe, carefully and in detail, your use case, so we can figure out how we can support that.

apfohl:
Hi, as described in issue #4655, it would be nice to have an option for the stub resolver to set an upstream nameserver. Maybe `recursor` is the wrong name for that, but it should definitely be possible.

Habbie:
@apfohl yes, I'd like to keep that around, but would indeed rename the flag for it if we remove `recursor`

sndrsmnk:
Please note that auth fails to do security checks / ALIAS expansion and logs about this with no `recursor=` set:

```
No recursors set, stub resolving (including secpoll and ALIAS) impossible.
Could not retrieve security status update for '4.0.3' on 'auth-4.0.3.security-status.secpoll.powerdns.com.', RCODE = Server Failure
```

pieterlexis:
@sndrsmnk by default, we do parse /etc/resolv.conf. if recursor is set to (empty), we indeed cannot resolve.

sndrsmnk:
> @sndrsmnk by default, we do parse /etc/resolv.conf. if recursor is set to (empty), we indeed cannot resolve.

But having no 'nameserver' in resolv.conf should default to localhost, right?

ahupowerdns:
@tomsommer - noted, but this is what we have chosen since the feature was confusing and made our life very difficult. What I hear is that you would like to be able to provision zones to the PowerDNS Recursor, is that correct?

tomsommer:
That's exactly how we use Auth in our internal DNS structure, for split horizon on thousands of hostnames, with the recursor as fallback for any domains unknown to the auth.

ahupowerdns:
So 1) you can actually provision authoritative zones to the PowerDNS Recursor with the API 2) have you studied https://doc.powerdns.com/authoritative/guides/recursion.html for workarounds?

tomsommer:
[...]

PR #5900, opened 2017-11-02
Mark recursor as removed

tomsommer:
According to #4616 the recursor setting was removed entirely in 4.1.0

### DIFF
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -1179,8 +1179,7 @@ Seconds to store recursive packets in the :ref:`packet-cache`.
 ------------
 
 -  IP Address
-
-.. deprecated:: 4.1.0
+-  Removed in: 4.1.0
 
 If set, recursive queries will be handed to the recursor specified here.
 


docs/settings.rst
```diff
@@ -1179,8 +1179,7 @@ Seconds to store recursive packets in the :ref:`packet-cache`.
 ------------
 
 -  IP Address
-
-.. deprecated:: 4.1.0
+-  Removed in: 4.1.0
```
pieterlexis: We prefer to use the [Sphinx markup](http://www.sphinx-doc.org/en/stable/markup/para.html) for this. I suggest changing this to 

```rst
.. deprecated:: 4.1.0
  recursion is removed
```

With perhaps a link to the migration docs.

Habbie:
ping!

aerique:
@pieterlexis Please review again.

docs/settings.rst
aerique: There's a lot of other entries using "Removed in", why should this one as the only one use `.. deprecated::…` ?

PR #5910, opened 2017-11-06
Have the REST-API use the `disable-dnssec` function from pdnsutil and…

### LINKED ISSUE #5909: API: Ability to un-secure zone
Currently (4.1.0) it is not possible to un-secure a zone using the REST-API:

https://github.com/PowerDNS/pdns/blob/29fab30a73e01fe8e32ca6093350113782ddb362/pdns/ws-auth.cc#L629

tomsommer:
Have the REST-API use the `disable-dnssec` function from pdnsutil and copy behaviour

Fixes #5909

### Checklist

I have:
- [x] documented the code

### DIFF
--- a/pdns/ws-auth.cc
+++ b/pdns/ws-auth.cc
@@ -627,8 +627,13 @@ static void updateDomainSettingsFromDocument(UeberBackend& B, const DomainInfo& 
       // "dnssec": false in json
       if (isDNSSECZone) {
         disableDNSSECOnZone(dk, zonename);
-        if (!dk.isSecuredZone(zonename)) {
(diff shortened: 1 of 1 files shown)


tomsommer:
Notes from IRC:
- This copies behaviour from `pdnsutil`, if this behaviour is wrong/too agressive, it should be changed in both instances. Personally I don't feel this is the case.
- It makes sense to move the disableDNSSECOnZone() function to the DNSSECKeeper class, just like rectifyZone()
- The same functionality can be achieved by using the cryptokeys API-endpoints and unsetting presigned and NSEC3PARAM. Very complicated.

@pieterlexis

PR #5917, opened 2017-11-07
use _exit() when we really really want to exit, for example after a fatal error

ahupowerdns:
### Short description
This stops us dying while we die. A call to exit() will trigger destructors, which may paradoxically stop the process from exiting, taking down only one thread, but harming the rest of the process.

### Checklist

I have:
- [ ] documented the code


rgacogne:
I see some nasty looking `exit()` calls in `distributor.hh` too, did you skip these on purpose?

PR #5930, opened 2017-11-08
in the recursor secpoll code, we ASSumed the TXT record would be the first record

ahupowerdns:
### Short description
in the recursor secpoll code, we ASSumed the TXT record would be the first record first record we received. Sometimes it was the RRSIG, leading to a silent error, and no secpoll check. Fixed the assumption, added an error.

### Checklist

I have:
- [ ] documented the code

### DIFF
--- a/pdns/pdns_recursor.cc
+++ b/pdns/pdns_recursor.cc
@@ -2162,6 +2162,15 @@ static void houseKeeping(void *)
         {
           L<<Logger::Error<<"Exception while performing security poll: "<<e.what()<<endl;
         }
+        catch(PDNSException& e)
+        {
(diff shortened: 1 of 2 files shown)


pdns/pdns_recursor.cc
```diff
@@ -2158,7 +2158,10 @@ static void houseKeeping(void *)
 	try {
 	  doSecPoll(&last_secpoll);
 	}
-	catch(...) {}
+	catch(std::exception& e)
+        {
+          L<<Logger::Error<<"Exception while performing security poll: "<<e.what()<<endl;
+        }
```
rgacogne: We should probably keep catching other types of exceptions here, because it looks like a `PDNSException` for example would bubble up and terminate the thread.

PR #5932, opened 2017-11-09
Improve logging in the webserver

### LINKED ISSUE #5923: Log raw requests received by the webserver
 - Issue type: Feature request 

### Short description

Webserver should print the HTTP request (requesting IP, headers, parameters, request body etc) in the logs at the high(est) levels of logging.

### Usecase

The webserver API provider programmatic access to PDNS records. Sometimes the HTTP API requests fail and are hard to debug simply based on the response. When debugging client/server issues it would be nice to be able to dump the HTTP request received by the server in the logs (at higher levels of logging).

### Description

The webserver should log information about HTTP Requests. Any standard format (taking inspiration from httpd's log format for example), will do.

pieterlexis:
### Short description
This makes the webserver log all requests in a format similar to the Common Log Format.
The following fields are not logged: the two user fields and datetime field.

On loglevel 6 and higher, the GET and POST parameters, alongside the request and response headers.

On loglevel 7 (debug) the full request body and response bodies are logged as well.

Fixes #5923

Points of discussion:
 * What parts of the existing logging should be replaced?
 * Should the API(s) log at the different levels (in another PR)

### Checklist

I have:
- [ ] documented the code


zeha:
For an API, most of the value is in logging the incoming request body ...

pdns/webserver.cc
```diff
@@ -308,17 +308,54 @@ try {
   resp.write(ss);
   string reply = ss.str();
 
+  ComboAddress remote;
+  client->getRemote(remote);
+  L<<Logger::Warning<<d_logprefix<<remote<<" \""<<req.method<<" "<<req.url.path<<" HTTP/"<<req.versionStr(req.version)<<"\" "<<resp.status<<" "<<reply.size()<<endl;
```
rgacogne: Logging this at warning might be a bit too verbose? Perhaps we should hide that behind a specific flag?

pieterlexis: Perhaps we should log this line at the NOTICE level?

pieterlexis:
rebased!

pdns/pdns_recursor.cc
```diff
@@ -4185,6 +4185,7 @@ int main(int argc, char **argv)
     ::arg().set("webserver-port", "Port of webserver to listen on") = "8082";
     ::arg().set("webserver-password", "Password required for accessing the webserver") = "";
     ::arg().set("webserver-allow-from","Webserver access is only allowed from these subnets")="127.0.0.1,::1";
+    ::arg().set("webserver-log-level","Amount of logging in the webserver (none, normal, detailed)")="normal";
```
zeha: can i get a space after the ,? :)

pieterlexis: there is..

Habbie:
```suggestion
    ::arg().set("webserver-log-level", "Amount of logging in the webserver (none, normal, detailed)")="normal";
```

Habbie: I think @zeha means this ^

pieterlexis:
I think this is ready for a thorough review.

pdns/ws-recursor.cc
```diff
@@ -620,50 +622,120 @@ void AsyncServer::newConnection()
...
-void AsyncWebServer::serveConnection(std::shared_ptr<Socket> client) const
-try {
-  HttpRequest req;
-  YaHTTP::AsyncRequestLoader yarl;
-  yarl.initialize(&req);
-  client->setNonBlocking();
-
-  string data;
+void AsyncWebServer::serveConnection(std::shared_ptr<Socket> client) const {
+  const string logprefix = d_logprefix + "<" + to_string(getUniqueID()) + "> ";
```
zeha: i'd favor removing the <> symbols, as I think they add little value, make parsing the message harder (would need stripping in awk etc) and sometimes can cause operational boo-boos (think inadvertent pasting of the log message into a shell instead of a textfile)

pdns/webserver.cc
```diff
@@ -276,49 +278,119 @@ void WebServer::handleRequest(HttpRequest& req, HttpResponse& resp) const
...
+    } catch (YaHTTP::ParseError &e) {
+      // request stays incomplete
+    }
+
+    if (d_loglevel >= WebServer::LogLevel::None) {
+      client->getRemote(remote);
+    }
+
+    if (d_loglevel >= WebServer::LogLevel::Detailed) {
+      g_log<<Logger::Notice<<logprefix<<"Request Details:"<<endl;
```
zeha: wouldnt it make sense to split out the logging stuff into a separate (member) function that can be shared by auth and rec? (remember that rec's AsyncWebServer is a WebServer child). might also make the diff easier to read :)

func name suggestions could be "logRequest" and "logResponse"

pdns/pdns_recursor.cc
```diff
@@ -4196,6 +4196,7 @@ int main(int argc, char **argv)
     ::arg().set("webserver-port", "Port of webserver to listen on") = "8082";
     ::arg().set("webserver-password", "Password required for accessing the webserver") = "";
     ::arg().set("webserver-allow-from","Webserver access is only allowed from these subnets")="127.0.0.1,::1";
+    ::arg().set("webserver-log-level", "Amount of logging in the webserver (none, normal, detailed)") = "normal";
```
zeha: the other programs use only one dash?

Habbie:
```
[peter:~/projects/powerdns/pdns] rec-docs-move-mib+* ± git grep log-level | grep -w set
[peter:~/projects/powerdns/pdns] rec-docs-move-mib+* 1 ± git grep loglevel | grep -w set
pdns/common_startup.cc:  ::arg().set("loglevel","Amount of logging. Higher is more. Do not set below 3")="4";
pdns/pdns_recursor.cc:    ::arg().set("loglevel","Amount of logging. Higher is more. Do not set below 3")="6";
pdns/pdnsutil.cc:  ::arg().set("loglevel","Amount of logging. Higher is more.")="3";
[peter:~/projects/powerdns/pdns] rec-docs-move-mib+* ±
```

pieterlexis: fun fact, it was correct in the docs!

Habbie:
While debugging an API problem (on master, not on this PR) I noticed we basically throw away error information from yahttp.

Here's a small improvement that helped me debug:
```diff
diff --git a/pdns/webserver.cc b/pdns/webserver.cc
index 979cc7085..8e3bf3d24 100644
--- a/pdns/webserver.cc
+++ b/pdns/webserver.cc
@@ -289,17 +289,20 @@ try {
       int bytes;
       char buf[1024];
       bytes = client->readWithTimeout(buf, sizeof(buf), timeout);
       if (bytes > 0) {
         string data = string(buf, bytes);
         req.complete = yarl.feed(data);
       } else {
         // read error OR EOF
         break;
       }
     }
     yarl.finalize();
   } catch (YaHTTP::ParseError &e) {
-    // request stays incomplete
+    cerr<<"YaHTTP::ParseError: "<<e.reason<<endl;
   }

   HttpResponse resp;
```

It does not look like the PR improves this situation, so here's a suggestion. It looks like we throw away this info in one or two other places too.

Bonus suggestion: report errors back to clients instead of just telling them 'Bad Request'.

pieterlexis:
> Bonus suggestion: report errors back to clients instead of just telling them 'Bad Request'.

That is something for later, see:

* #7097
* #7090
* #7100

zeha:
land this!

PR #5933, opened 2017-11-09
Check return value for all getTSIGKey calls

### LINKED ISSUE #5931: pdns-auth terminates when a TSIG key is not found
It seems PowerDNS-Auth 4.x added support for outgoing NOTIFYs with TSIG. Therefore, in mastercommunicator.cc CommunicatorClass::sendNotification() the TSIG key is fetched from DB. The problem happens when the domainmetadata TSIG-ALLOW-AXFR references a key, but this key is not in the tsigkeys' table. Because B.getTSIGKey() is used without checking the return value (like everywhere else in PowerDNS code). Hence, the unset DNSName tsigalgorithm will be used which throws a fatal exception.

With perfect provisioning this should not happen, but IMO PowerDNS should not terminate if there is a missing key in the DB, better log an error.

So please check the return value of B.getTSIGKey(), any maybe also of B64Decode(), as it is done in slavecommunicator.cc CommunicatorClass::suck().

pieterlexis:
### Short description
This would lead to crashes if the TSIG key was referenced in TSIG-ALLOW-FROM but the key was not in the tsigkeys table.

Closes #5931

I do need to add some regression tests, but need to have a good look at how.

Note: some unchecked calls to `B64Decode` remain in the code, these will be fixed in another PR.

### Checklist

I have:
- [ ] documented the code

### DIFF
--- a/pdns/mastercommunicator.cc
+++ b/pdns/mastercommunicator.cc
@@ -238,7 +238,10 @@ void CommunicatorClass::sendNotification(int sock, const DNSName& domain, const 
   pw.getHeader()->aa = true; 
 
   if (tsigkeyname.empty() == false) {
-    B.getTSIGKey(tsigkeyname, &tsigalgorithm, &tsigsecret64);
+    if (!B.getTSIGKey(tsigkeyname, &tsigalgorithm, &tsigsecret64)) {
+      L<<Logger::Error<<"TSIG key '"<<tsigkeyname<<"' for domain '"<<domain<<"' not found"<<endl;
+      return;
+    }
     TSIGRecordContent trc;
     if (tsigalgorithm.toStringNoDot() == "hmac-md5")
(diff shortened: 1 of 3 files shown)


pdns/mastercommunicator.cc
```diff
@@ -248,7 +251,10 @@ void CommunicatorClass::sendNotification(int sock, const DNSName& domain, const
     trc.d_fudge = 300;
     trc.d_origID=ntohs(id);
     trc.d_eRcode=0;
-    B64Decode(tsigsecret64, tsigsecret);
+    if (!B64Decode(tsigsecret64, tsigsecret)) {
```
rgacogne: Isn't that the other way around? It looks like `B64Decode` returns `0` when everything is fine, although our unit tests are not checking that at all..

pieterlexis: you are right, looking at the code for B64Decode I wonder why we don't just return a bool.....

PR #5936, opened 2017-11-09
API: Allow disabling DNSSEC

### LINKED ISSUE #5909: API: Ability to un-secure zone

pieterlexis:
### Short description

This PR makes sure we can disable DNSSEC via the API, this is equivalent to doing `pdnsutil disable-dnssec`.

Closes #5909
Closes #5910 

### Checklist

I have:
- [x] documented the code

### DIFF
--- a/pdns/dbdnsseckeeper.cc
+++ b/pdns/dbdnsseckeeper.cc
@@ -589,6 +589,26 @@ bool DNSSECKeeper::getTSIGForAccess(const DNSName& zone, const string& master, D
   return false;
 }
 
+bool DNSSECKeeper::unSecureZone(const DNSName& zone, string& error, string& info) {
+  // Not calling isSecuredZone(), as it will return false for zones with zero
+  // active keys.
+  DNSSECKeeper::keyset_t keyset=getKeys(zone);
+
+  if(keyset.empty())  {
+    error = "No keys for zone '" + zone.toLogString() + "'.";
+    return false;
+  }
+
+  for(auto& key : keyset) {
+    deactivateKey(zone, key.second.id);
+    removeKey(zone, key.second.id);
+  }
+
+  unsetNSEC3PARAM(zone);
+  unsetPresigned(zone);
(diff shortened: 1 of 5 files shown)


pdns/dbdnsseckeeper.cc
```diff
@@ -589,6 +589,26 @@ bool DNSSECKeeper::getTSIGForAccess(const DNSName& zone, const string& master, D
   return false;
 }
 
+bool DNSSECKeeper::unSecureZone(const DNSName& zone, string& error, string& info) {
+  // Not calling isSecuredZone(), as it will return false for zones with zero
+  // active keys.
+  DNSSECKeeper::keyset_t keyset=getKeys(zone);
```
rgacogne: Just a nit, but perhaps `isSecureZone()` could get an optional boolean to return `true` as long as there is one key, be it inactive?

pdns/dbdnsseckeeper.cc
```diff
@@ -589,6 +589,26 @@ bool DNSSECKeeper::getTSIGForAccess(const DNSName& zone, const string& master, D
   return false;
 }
 
+bool DNSSECKeeper::unSecureZone(const DNSName& zone, string& error, string& info) {
```
rgacogne: `info` doesn't seem to be used?

pieterlexis: Correct, should we ever add more functionality to this function, then it might be. I kept is similar to `rectifyZone()`.

pdns/dbdnsseckeeper.cc
pieterlexis: That might make sense, perhaps turn that into a ticket?

PR #5960, opened 2017-11-14
Check more thoroughly the source of UDP answers

rgacogne:
### Short description

Since several implementations suffer from a race allowing an attacker to insert `UDP` packets between the `bind()` and `connect()` calls from a different source that the expected one (see [1]), this PR adds additional checks: 
* Check the source address with `recvfrom()` in the recursor's `handleGenUDPQueryResponse()` to prevent `bind()`/`connect()` race
* Store the remote address in the auth's `DNSProxy` and use `recvfrom()`, so we can prevent a race condition if a packet arrives between `bind()` and `connect()`
* Check the remote peer address in the auth's `Resolver` when we receive a response since the socket is not connected

[1]: http://seclists.org/oss-sec/2017/q4/224

### Checklist

I have:
- [ ] documented the code

### DIFF
--- a/pdns/dnsproxy.cc
+++ b/pdns/dnsproxy.cc
@@ -44,18 +44,21 @@ DNSProxy::DNSProxy(const string &remote)
 
   vector<string> addresses;
   stringtok(addresses, remote, " ,\t");
-  ComboAddress remaddr(addresses[0], 53);
-  
-  if((d_sock=socket(remaddr.sin4.sin_family, SOCK_DGRAM,0))<0)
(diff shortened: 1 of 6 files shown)


rgacogne:
Just to be clear, this does not touch the recursor's handling of `UDP` answers from authoritative servers, since we already do all the possible checks there.

Habbie:
Looks like this conflicts with #5979 which I just merged.

rgacogne:
Rebased!

PR #5964, opened 2017-11-16
API: Make the /cryptokeys endpoint consistently use CryptoKey objects

pieterlexis:
### Short description
Add algorithm and bits to the Cryptokey objects and let the POST command accept this CryptoKey object.

### Checklist

I have:
- [x] documented the code


pieterlexis:
We should probably land this in 4.1.0 final so there is never a release with inconsistent API cryptokeys

zeha:
++

PR #5971, opened 2017-11-20
rec: Fix the use of a deleted iterator in SyncRes::getDSRecords()

rgacogne:
### Short description

The iterator is invalidated after being erased, so we have to handle that case explicitly. 

### Checklist

I have:
- [ ] documented the code


stbuehler:
* https://stackoverflow.com/questions/34624223/how-to-erase-elements-from-a-vector-based-on-a-specific-condition-in-c11#34624284
* http://en.cppreference.com/w/cpp/algorithm/remove

PR #5982, opened 2017-11-21
improve error message of pdnsutil if it can't find configuration

ahupowerdns:
### Short description
pdnsutil only says it can't load your config it it can't, doesn't tell you why. With this PR it does.

### Checklist

I have:
- [ ] documented the code

### DIFF
--- a/pdns/pdnsutil.cc
+++ b/pdns/pdnsutil.cc
@@ -121,7 +121,7 @@ void loadMainConfig(const std::string& configdir)
 
   // Keep this line below all ::arg().set() statements
   if (! ::arg().laxFile(configname.c_str()))
-    cerr<<"Warning: unable to read configuration file '"<<configname<<"'."<<endl;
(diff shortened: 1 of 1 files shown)


pdns/pdnsutil.cc
```diff
@@ -121,7 +121,7 @@ void loadMainConfig(const std::string& configdir)
 
   // Keep this line below all ::arg().set() statements
   if (! ::arg().laxFile(configname.c_str()))
-    cerr<<"Warning: unable to read configuration file '"<<configname<<"'."<<endl;
+    cerr<<"Warning: unable to read configuration file '"<<configname<<"': "<<strerror(errno)<<endl;
```
pieterlexis: Is there a reason `laxFile` can return false and `errno` is not set? If that would be the case, the error message would be lying (as it uses an old `errno`).

zeha: maybe laxFile should just return (-)errno...?

PR #5862, opened 2017-10-26
Add Draft of swagger spec for Authoritative Server HTTP API

ffledgling:
### Short description
- Defines most object types
- Documents GET requests for /servers and /zones

Issue #5854 

### Checklist

I have:
- [x] read the [CONTRIBUTING.md](https://github.com/PowerDNS/pdns/blob/master/CONTRIBUTING.md) document
- [x] included documentation

PR #5988, opened 2017-11-23
API: Add TSIG key manipulation endpoints

pieterlexis:
### Short description
This PR implements a new endpoint at `/api/v1/servers/{server_id}/tsigkeys` where TSIG keys can be manipulated.

This also adds 2 fields to the `Zone` object that list the TSIG master and slave keys.

This PR also adds a more flexible way to disclose HTTP error codes to the client. e.g. an HTTP 404 can now also contain a separate message instead of the default "Not Found" (and json as well).

At the moment, documentation is missing. I would like to add this endpoint to the swagger spec after #5862 is merged (and the docs are generated form this spec). However, I already open this PR for feedback and comments.

### Checklist
I have:
- [ ] documented the code


pdns/ws-auth.cc
```diff
@@ -663,6 +681,29 @@ static void updateDomainSettingsFromDocument(UeberBackend& B, const DomainInfo&
...
+
+  if (!document["master_tsig_key_ids"].is_null()) {
+    vector<string> metadata;
+    for(auto value : document["master_tsig_key_ids"].array_items()) {
+      auto keyname(value.string_value());
+      // XXX test if the key actually exists?
+      metadata.push_back(apiZoneIdToName(keyname).toString());
+    }
+    if (!di.backend->setDomainMetadata(zonename, "TSIG-ALLOW-AXFR", metadata)) {
+      throw ApiException("Unable to set new TSIG master keys for zone '" + zonename.toLogString() + "'");
```
pieterlexis: Should become internal server error.

pdns/ws-auth.cc
```diff
@@ -663,6 +681,29 @@ static void updateDomainSettingsFromDocument(UeberBackend& B, const DomainInfo&
...
+  }
+  if (!document["slave_tsig_key_ids"].is_null()) {
+    vector<string> metadata;
+    for(auto value : document["slave_tsig_key_ids"].array_items()) {
+      auto keyname(value.string_value());
+      // XXX test if the key actually exists?
+      metadata.push_back(apiZoneIdToName(keyname).toString());
+    }
+    if (!di.backend->setDomainMetadata(zonename, "AXFR-MASTER-TSIG", metadata)) {
+      throw ApiException("Unable to set new TSIG slave keys for zone '" + zonename.toLogString() + "'");
```
pieterlexis: Should become internal server error.

pdns/ws-auth.cc
```diff
@@ -1179,6 +1222,154 @@ static void checkDuplicateRecords(vector<DNSResourceRecord>& records) {
   }
 }
 
+static void checkTSIGKey(UeberBackend& B, const DNSName& keyname, const DNSName& algo, const string& content) {
+  DNSName algoFromDB;
+  string contentFromDB;
+  B.getTSIGKey(keyname, &algoFromDB, &contentFromDB);
+  if (!contentFromDB.empty() || !algoFromDB.empty()) {
+    throw ApiException("A TSIG key with the name '"+keyname.toLogString()+"' already exists");
```
zeha: theres probably a dedicated http status code for that. ("Conflict"?)

pdns/ws-auth.cc
```diff
@@ -1179,6 +1222,154 @@ static void checkDuplicateRecords(vector<DNSResourceRecord>& records) {
...
+    { "type", "TSIGKey" }
+  };
+  return tsigkey;
+}
+
+static Json::object makeJSONTSIGKey(const struct TSIGKey& key) {
+  return makeJSONTSIGKey(key.name, key.algorithm, key.key);
+}
+
+static void apiServerTsigKeys(HttpRequest* req, HttpResponse* resp) {
```
zeha: `apiServerTSIGKeys` for consistency?

pdns/ws-auth.cc
```diff
@@ -1179,6 +1222,154 @@ static void checkDuplicateRecords(vector<DNSResourceRecord>& records) {
...
+        klen = 32;
+      }
+      char tmpkey[64];
+      for(size_t i = 0; i < klen; i+=4) {
+        *(unsigned int*)(tmpkey+i) = dns_random(0xffffffff);
+      }
+      content = Base64Encode(std::string(tmpkey, klen));
+    }
+
+    // Will throw and ApiException on error
```
zeha: and=an

pdns/ws-auth.cc
```diff
@@ -1179,6 +1222,154 @@ static void checkDuplicateRecords(vector<DNSResourceRecord>& records) {
...
+    if (content.empty()) {
+      size_t klen = 64;
+      if (algo.toStringNoDot() == "hmac-md5"
+          || algo.toStringNoDot() == "hmac-sha1"
+          || algo.toStringNoDot() == "hmac-sha224") {
+        klen = 32;
+      }
+      char tmpkey[64];
+      for(size_t i = 0; i < klen; i+=4) {
+        *(unsigned int*)(tmpkey+i) = dns_random(0xffffffff);
```
zeha: that looks like it wants to go through memcpy() to avoid breaking aliasing rules.

pdns/ws-auth.cc
```diff
@@ -1179,6 +1222,154 @@ static void checkDuplicateRecords(vector<DNSResourceRecord>& records) {
...
+    DNSName algo(stringFromJson(document, "algorithm"));
+    string content = document["key"].string_value();
+
+    if (content.empty()) {
+      size_t klen = 64;
+      if (algo.toStringNoDot() == "hmac-md5"
+          || algo.toStringNoDot() == "hmac-sha1"
+          || algo.toStringNoDot() == "hmac-sha224") {
+        klen = 32;
+      }
```
zeha: stupid q: utility function in tsigsomething.hh?

pdns/ws-auth.cc
```diff
@@ -1179,6 +1222,154 @@ static void checkDuplicateRecords(vector<DNSResourceRecord>& records) {
...
+
+  if (!B.getTSIGKey(keyname, &algo, &content)) {
+    throw HttpNotFoundException("TSIG key with name '"+keyname.toLogString()+"' not found");
+  }
+
+  json11::Json document;
+
+  if (!req->body.empty()) {
+    document = req->json();
+  }
```
zeha: maybe move this into the if(...PUT) block

pdns/ws-auth.cc
```diff
@@ -1179,6 +1222,154 @@ static void checkDuplicateRecords(vector<DNSResourceRecord>& records) {
...
+      string decoded;
+      if (B64Decode(new_content, decoded) == -1) {
+        throw ApiException("Can not base64 decode key content '" + new_content + "'");
+      }
+      tsk.key = new_content;
+    }
+    if (!B.setTSIGKey(tsk.name, tsk.algorithm, tsk.key)) {
+      throw HttpInternalServerErrorException("Unable to save TSIG Key");
+    }
+    if (tsk.name != keyname) {
```
zeha: I'd favour disallowing this?

pdns/ws-auth.cc
```diff
@@ -355,6 +356,23 @@ static void fillZone(const DNSName& zonename, HttpResponse* resp, bool doRRSets)
...
   doc["api_rectify"] = (api_rectify == "1");
 
+  // TSIG
+  vector<string> tsig_master, tsig_slave;
+  di.backend->getDomainMetadata(zonename, "TSIG-ALLOW-AXFR", tsig_master);
+  di.backend->getDomainMetadata(zonename, "AXFR-MASTER-TSIG", tsig_slave);
+
+  Json::array tsig_master_keys;
+  for (const auto& keyname : tsig_master) {
+    tsig_master_keys.push_back(apiZoneNameToId(DNSName(keyname)));
```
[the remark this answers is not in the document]
pieterlexis: they are, see [here](https://github.com/pieterlexis/pdns/blob/f910b71cb47c7ac75ca4ebe9435ae4a553ab7bf9/pdns/dnsbackend.hh#L91-L95) and [here](https://github.com/pieterlexis/pdns/blob/f910b71cb47c7ac75ca4ebe9435ae4a553ab7bf9/pdns/dbdnsseckeeper.cc#L578-L590)

pdns/webserver.cc
```diff
@@ -258,7 +258,9 @@ void WebServer::handleRequest(HttpRequest& req, HttpResponse& resp) const
       resp.body = "<!html><title>" + what + "</title><h1>" + what + "</h1>";
     } else if (req.accept_json) {
       resp.headers["Content-Type"] = "application/json";
-      resp.setErrorResult(what, resp.status);
+      if (resp.body.empty()) {
```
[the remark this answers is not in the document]
pieterlexis: and append `"error"`to the body item?

pdns/ws-auth.cc
pieterlexis: it works and is tested. The zone endpoint docs alluded to this being possible for zones.

pdns/ws-auth.cc
pieterlexis: makes sense

pdns/ws-auth.cc
pieterlexis: I copied this from pdnsutil, but I'll see what I can do

pieterlexis:
Putting this on hold for a moment, as #5274 needs deciding on wrt to the use of `dns_random` for key material.

pieterlexis:
rebased on master as #5274 is merged :tada:

zeha:
from travis:

```
tsigutils.cc: In function ‘std::string makeTSIGKey(const DNSName&)’:
tsigutils.cc:48:19: error: variable length array ‘tmpkey’ is used [-Werror=vla]
   char tmpkey[klen];
                   ^
cc1plus: some warnings being treated as errors
make[3]: *** [tsigutils.o] Error 1
```

docs/http-api/tsigkey.rst
```diff
@@ -0,0 +1,59 @@
+TSIGKeys
+========
+.. versionadded:: 4.2.0
+
+TSIGKeys and be manipulated via the API.
```
rgacogne: This sentence doesn't seem right.

pdns/tsigutils.cc
```diff
@@ -0,0 +1,54 @@
...
+  }
+
+  size_t klen = 64;
+  if (tsigHashEnum == TSIG_MD5
+      || tsigHashEnum == TSIG_SHA1
+      || tsigHashEnum == TSIG_SHA224) {
+    klen = 32;
+  }
+
+  char tmpkey[64];
```
rgacogne: Since we are going to allocate a string anyway, how about allocating a string right way, resizing it to the correct size and then accessing it with `std::string::at()`, so we have boundary checks?

pdns/tsigutils.cc
```diff
@@ -0,0 +1,54 @@
...
+  for (size_t i = 0; i < klen; i += 4) {
+    unsigned int t = dns_random(0xffffffff);
```
rgacogne: We shouldn't assume that `sizeof(unsigned int) == 4`, it will not be true on some architectures. Perhaps just use `sizeof(unsigned int)` instead of `4` above, and of course in the `memcpy()` call below?

pdns/dns_random.cc
```diff
@@ -210,7 +210,7 @@ void dns_random_init(const string& data __attribute__((unused)), bool force) {
 }
 
 /* Parts of this code come from arc4random_uniform */
-unsigned int dns_random(unsigned int upper_bound) {
+uint32_t dns_random(uint32_t upper_bound) {
```
rgacogne: It would be nice to fix the return type for the `getrandom()` and `openssl` cases as well.

rgacogne: Also `ULONG_MAX` should probably be changed to `UINT32_MAX`.

pdns/tsigutils.cc
```diff
@@ -0,0 +1,63 @@
...
+      || tsigHashEnum == TSIG_SHA1
+      || tsigHashEnum == TSIG_SHA224) {
+    klen = 32;
+  }
+
+  string tmpkey;
+  tmpkey.resize(klen);
+
+  for (size_t i = 0; i < klen; i += sizeof(unsigned int)) {
+    unsigned int t = dns_random(std::numeric_limits<unsigned int>::max());
```
rgacogne: So you changed `dns_random()` return type to `uint32_t` but you still use `unsigned int` here?

pdns/tsigutils.cc
```diff
@@ -0,0 +1,63 @@
...
+    memcpy(&tmpkey.at(i), &t, sizeof(unsigned int));
+    if (i + sizeof(unsigned int) > klen) {
+      size_t needed_bytes = klen - i;
+      for (size_t j = 0; j < needed_bytes; j++) {
```
rgacogne: I'm pretty sure we can remove that code, perhaps throwing an exception if `klen` is not a multiple of 32 which is not going to happen.

pieterlexis:
I think this PR is done?

pieterlexis:
@rgacogne can you re-review the `dns_random` changes?

pdns/dns_random.cc
```diff
@@ -224,7 +224,7 @@ unsigned int dns_random(unsigned int upper_bound) {
      On applicable rngs, we loop until the rng spews out
      value larger than min, and then take modulo out of that.
   */
-#if (ULONG_MAX > 0xffffffffUL)
+#if (UINT32_MAX > 0xffffffffUL)
```
rgacogne: Nope, it should remain `ULONG_MAX` here.

Habbie:
> Is `ULONG_MAX` the only remaining issue with this PR?

Rest of the code looks good to me, in any case.

PR #5993, opened 2017-11-26
Recursor: add ecs-add-for option

zeha:
### Short description

Adds an option named `ecs-add-for` to the Recursor. Clients matching this NetmaskGroup will get their real IP address in the ECS option in outgoing packets. This should be especially useful for setups running PowerDNS Recursor behind a DNS loadbalancer like dnsdist.

The existing `edns-subnet-whitelist` stays in effect, and not matching clients will get the scope zero fallback. Also, if the client already supplied an ECS option, it will be passed on according to the existing rules. Previously, all clients would get this behaviour, except if they already sent an ECS option.

This commit also restructures the ECS code in SyncRes to be easier to understand.

### Checklist

I have:
- [ ] documented the code

### DIFF
--- a/pdns/pdns_recursor.cc
+++ b/pdns/pdns_recursor.cc
@@ -167,6 +167,7 @@ uint16_t g_outgoingEDNSBufsize;
 bool g_logRPZChanges{false};
 
 #define LOCAL_NETS "127.0.0.0/8, 10.0.0.0/8, 100.64.0.0/10, 169.254.0.0/16, 192.168.0.0/16, 172.16.0.0/12, ::1/128, fc00::/7, fe80::/10"
+#define LOCAL_NETS_INVERSE "!127.0.0.0/8, !10.0.0.0/8, !100.64.0.0/10, !169.254.0.0/16, !192.168.0.0/16, !172.16.0.0/12, !::1/128, !fc00::/7, !fe80::/10"
 // Bad Nets taken from both:
 // http://www.iana.org/assignments/iana-ipv4-special-registry/iana-ipv4-special-registry.xhtml
 // and
@@ -789,7 +790,6 @@ static void startDoResolve(void *p)
     if(t_pdl) {
       sr.setLuaEngine(t_pdl);
     }
-    sr.d_requestor=dc->d_remote; // ECS needs this too
     if(g_dnssecmode != DNSSECMode::Off) {
       sr.setDoDNSSEC(true);
 
@@ -808,12 +808,7 @@ static void startDoResolve(void *p)
     sr.setInitialRequestId(dc->d_uuid);
 #endif
 
-    if (g_useIncomingECS) {
-      sr.setIncomingECSFound(dc->d_ecsFound);
-      if (dc->d_ecsFound) {
-        sr.setIncomingECS(dc->d_ednssubnet);
-      }
-    }
+    sr.setQuerySource(dc->d_remote, g_useIncomingECS && !dc->d_ednssubnet.source.empty() ? boost::optional<const EDNSSubnetOpts&>(dc->d_ednssubnet) : boost::none);
 
     bool tracedQuery=false; // we could consider letting Lua know about this too
     bool variableAnswer = false;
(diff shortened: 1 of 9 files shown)


pdns/pdns_recursor.cc
```diff
@@ -2986,6 +2981,10 @@ static int serviceMain(int argc, char*argv[])
     }
   }
 
+  SyncRes::parseEDNSSubnetWhitelist(::arg()["edns-subnet-whitelist"]);
+  SyncRes::parseEDNSSubnetAddFor(::arg()["ecs-add-for"]);
+  g_useIncomingECS = ::arg().mustDo("use-incoming-edns-subnet");
```
rgacogne: The processing of `edns-subnet-whitelist` and `use-incoming-edns-subnet` is duplicated line 3033.

pdns/syncres.cc
```diff
@@ -2699,11 +2700,25 @@ int SyncRes::doResolveAt(NsSet &nameservers, DNSName auth, bool flawedNSSet, con
...
+
+  if (incomingECS && incomingECS->source.getBits() > 0) {
+    d_cacheRemote = incomingECS->source.getMaskedNetwork();
+    uint8_t bits = std::min(incomingECS->source.getBits(), (incomingECS->source.isIpv4() ? s_ecsipv4limit : s_ecsipv6limit));
+    ComboAddress trunc = incomingECS->source.getNetwork();
+    trunc.truncate(bits);
+    d_outgoingECSNetwork = boost::optional<Netmask>(Netmask(trunc, bits));
+  } else {
+    d_cacheRemote = d_requestor;
+    if(s_ednslocalsubnets.match(d_requestor)) {
```
rgacogne: It looks like we don't honor an incoming `ECS` with a 0-mask. If `use-incoming-edns-subnet` is set and we receive a 0-mask, we should use `s_ecsScopeZero` even if the source address is in `s_ednslocalsubnets`.

zeha: Right. I see (now) the RFC mandates this. OTOH I've got a production report that passing on 0.0.0.0 confuses (some) upstreams.

rgacogne: Yes, that's why we try hard not to use that in `s_ecsScopeZero`. Not hard enough, perhaps?

pdns/pdns_recursor.cc
zeha: Merge conflict resolution leftover that I've now removed. Thanks for noticing!

zeha:
@rgacogne all your points should be addressed now!

zeha:
Updated doc change to say 4.2.0.

zeha:
Rebased to fix conflicts.

pdns/recursordist/docs/settings.rst
```diff
@@ -322,6 +322,20 @@ This setting can be used to expand or reduce the limitations.
...
+--------------------------
+.. versionadded:: 4.2.0
+
+-  Comma separated list of netmasks
+-  Default: 0.0.0.0/0, ::, !127.0.0.0/8, !10.0.0.0/8, !100.64.0.0/10, !169.254.0.0/16, !192.168.0.0/16, !172.16.0.0/12, !::1/128, !fc00::/7, !fe80::/10
+
+List of requestor netmasks for which the requestor IP Address should be used as the :rfc:`EDNS Client Subnet <7871>` for outgoing queries. Instead, `ecs-scope-zero-address`_ would be used.
+Valid incoming ECS values from `use-incoming-edns-subnet`_ are not replaced.
+
+This defaults to not using the requestor address inside RFC1918 and similar "private" IP address spaces.
```
rgacogne: Re-reading this explanation without reminding the context, I think it could use more words. How about:

> List of requestor netmasks for which the requestor IP Address should be used as the :rfc:`EDNS Client Subnet <7871>` for outgoing queries. Outgoing queries for requestors that do not match this list will use the `ecs-scope-zero-address`_ instead.
Valid incoming ECS values from `use-incoming-edns-subnet`_ are not replaced.
Regardless of the value of this setting, ECS values are only sent for outgoing queries matching the conditions in the `edns-subnet-whitelist`_ setting. This setting only controls the actual value being sent.

zeha:
Added your suggested doc text.

rgacogne:
Code and documentation are good. I'm afraid the name of the setting might be misleading but I don't have a better proposal yet.

PR #5947, opened 2017-11-13
dnsdist: Split the Lua files, remove deprecated functions

rgacogne:
### Short description

Split the existing Lua files (dnsdist-lua.cc, dnsdist-lua2.cc) into:
- dnsdist-lua.cc: main one, holding configurations directives
- dnsdist-lua-actions.cc: actions-related code
- dnsdist-lua-bindings.cc: bindings for most structures like ComboAddress, Netmask, etc
- dnsdist-lua-bindings-dnsquestion.cc: bindings for DNSQuestion, DNSResponse
- dnsdist-lua-inspection.cc: inspection-related code (topXX, grepq(), ..)
- dnsdist-lua-rules.cc rules-related code
- dnsdist-lua-vars.cc: constants like DNSClass, DNSTypes, rcodes..

The main goal was to improve readability, but this might also make merges less painful and reduce the memory needed to compile dnsdist.
Note that it does remove functions deprecated in 1.2.0 because they prevented splitting actions from rules.
Comments welcome!

### Checklist

I have:
- [x] read the [CONTRIBUTING.md](https://github.com/PowerDNS/pdns/blob/master/CONTRIBUTING.md) document
- [x] compiled and tested this code
- [ ] included documentation (including possible behaviour changes)
- [ ] documented the code
- [ ] added or modified regression test(s)
- [ ] added or modified unit test(s)

PR #6003, opened 2017-11-27
Add TempFailureCacheTTLAction

zeha:
### Short description
Adds a request Action to set the packetcache TTL used for temporary failures, (= ServFail + Refused).

RFC. Will add docs, test etc when not everyone is shouting at me for this.

Usage example:
```
addAction("deduktiva.com.", TempFailureCacheTTLAction(1))
addAction("244.77.in-addr.arpa.", TempFailureCacheTTLAction(86400))
```

### Checklist

I have:
- [ ] documented the code

### DIFF
--- a/pdns/dnsdist-cache.cc
+++ b/pdns/dnsdist-cache.cc
@@ -89,7 +89,7 @@ void DNSDistPacketCache::insertLocked(CacheShard& shard, uint32_t key, const DNS
   value = newValue;
 }
 
-void DNSDistPacketCache::insert(uint32_t key, const DNSName& qname, uint16_t qtype, uint16_t qclass, const char* response, uint16_t responseLen, bool tcp, uint8_t rcode)
+void DNSDistPacketCache::insert(uint32_t key, const DNSName& qname, uint16_t qtype, uint16_t qclass, const char* response, uint16_t responseLen, bool tcp, uint8_t rcode, boost::optional<uint32_t> tempFailureTTL)
 {
   if (responseLen < sizeof(dnsheader))
     return;
@@ -97,7 +97,7 @@ void DNSDistPacketCache::insert(uint32_t key, const DNSName& qname, uint16_t qty
   uint32_t minTTL;
 
   if (rcode == RCode::ServFail || rcode == RCode::Refused) {
-    minTTL = d_tempFailureTTL;
+    minTTL = tempFailureTTL == boost::none ? d_tempFailureTTL : *tempFailureTTL;
     if (minTTL == 0) {
       return;
     }
--- a/pdns/dnsdist-tcp.cc
+++ b/pdns/dnsdist-tcp.cc
@@ -553,7 +553,7 @@ void* tcpClientThread(int pipefd)
         }
 
 	if (packetCache && !dq.skipCache) {
-	  packetCache->insert(cacheKey, qname, qtype, qclass, response, responseLen, true, dh->rcode);
+	  packetCache->insert(cacheKey, qname, qtype, qclass, response, responseLen, true, dh->rcode, dq.tempFailureTTL);
 	}
 
 #ifdef HAVE_DNSCRYPT
--- a/pdns/dnsdist.cc
+++ b/pdns/dnsdist.cc
@@ -467,7 +467,7 @@ try {
       }
 
       if (ids->packetCache && !ids->skipCache) {
-        ids->packetCache->insert(ids->cacheKey, ids->qname, ids->qtype, ids->qclass, response, responseLen, false, dh->rcode);
+        ids->packetCache->insert(ids->cacheKey, ids->qname, ids->qtype, ids->qclass, response, responseLen, false, dh->rcode, ids->tempFailureTTL);
       }
 
       if (ids->cs && !ids->cs->muted) {
@@ -1371,6 +1371,7 @@ static void processUDPQuery(ClientState& cs, LocalHolders& holders, const struct
     ids->qtype = dq.qtype;
     ids->qclass = dq.qclass;
     ids->delayMsec = delayMsec;
+    ids->tempFailureTTL = dq.tempFailureTTL;
     ids->origFlags = origFlags;
     ids->cacheKey = cacheKey;
     ids->skipCache = dq.skipCache;
--- a/pdns/dnsdist.hh
+++ b/pdns/dnsdist.hh
@@ -129,7 +129,7 @@ extern thread_local boost::uuids::random_generator t_uuidGenerator;
 
 struct DNSQuestion
 {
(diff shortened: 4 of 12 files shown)


pdns/dnsdist-cache.cc
```diff
@@ -89,15 +89,19 @@ void DNSDistPacketCache::insertLocked(CacheShard& shard, uint32_t key, const DNS
...
+void DNSDistPacketCache::insert(uint32_t key, const DNSName& qname, uint16_t qtype, uint16_t qclass, const char* response, uint16_t responseLen, bool tcp, uint8_t rcode, uint32_t tempFailureTTL)
 {
   if (responseLen < sizeof(dnsheader))
     return;
 
   uint32_t minTTL;
 
   if (rcode == RCode::ServFail || rcode == RCode::Refused) {
-    minTTL = d_tempFailureTTL;
+    if (tempFailureTTL != std::numeric_limits<uint32_t>::max()) {
```
rgacogne: It might be cleaner to make `tempFailureTTL` a `boost::optional<uint32_t>` instead of relying on a magic value? I know we haven't been consistent in the existing code..

pdns/dnsdist-lua2.cc
```diff
@@ -832,6 +832,10 @@ void moreLua(bool client)
 #endif
       });
 
+    g_lua.writeFunction("TempFailureTTLAction", [](int maxTTL) {
```
rgacogne: It would be more consistent to make `maxTTL` a `uint32_t` here.

pdns/dnsdist-tcp.cc
```diff
@@ -337,13 +337,14 @@ void* tcpClientThread(int pipefd)
 
 	string poolname;
 	int delayMsec=0;
+	uint32_t tempFailureTTL = std::numeric_limits<uint32_t>::max();
```
rgacogne: I have the feeling it would be better to make `tempFailureTTL` a member of `DNSQuestion`, it would make it easier to expose it to Lua later, if we want to. We should probably do the same thing for `delayMsec` and `poolname`, but that's another story.

pdns/dnsdist.cc
```diff
@@ -1009,6 +1009,9 @@ bool processQuery(LocalHolders& holders, DNSQuestion& dq, string& poolname, int*
       case DNSAction::Action::Delay:
         *delayMsec = static_cast<int>(pdns_stou(ruleresult)); // sorry
         break;
+      case DNSAction::Action::SetTempFailureTTL:
```
rgacogne: If we make `tempFailureTTL` a member of `DNSQuestion` we don't need to add a new `DNSAction::Action` value, `TempFailureTTLAction` can just set the value the way `SkipCacheAction` does.

pdns/dnsdist.hh
```diff
@@ -418,6 +419,7 @@ struct IDState
   uint16_t origID;                                            // 2
   uint16_t origFlags;                                         // 2
   int delayMsec;
+  uint32_t tempFailureTTL;
```
rgacogne: It would be slightly better to move this field with the other `uint32_t` one, for alignment purpose.

pdns/dnsrulactions.hh
```diff
@@ -1301,6 +1301,25 @@ private:
...
 };
 
+class TempFailureTTLAction : public DNSAction
+{
+public:
+  TempFailureTTLAction(int maxTTL) : d_maxTTL(maxTTL)
+  {}
+  TempFailureTTLAction::Action operator()(DNSQuestion* dq, string* ruleresult) const override
+  {
+    cerr << "going through TempFailureTTLAction " << endl;
```
rgacogne: The log message will have to go, of course :)

pdns/dnsrulactions.hh
```diff
@@ -1301,6 +1301,25 @@ private:
   std::string d_reason;
 };
 
+class TempFailureTTLAction : public DNSAction
```
rgacogne: I would like the name of the action to reflect that it only applies to the cache, but something like `TempFailureCacheTTLAction` is quite a mouthful..

zeha:
Addressed @rgacogne notes.

I'd think dq.tempFailureTTL should also be exposed to Lua, but now that it is a boost::optional<uint32_t> I'm not so sure how...

rgacogne:
> I'd think dq.tempFailureTTL should also be exposed to Lua, but now that it is a boost::optional<uint32_t> I'm not so sure how...

You can easily write custom member handlers using `registerMember`, making it easy to set and get the actual value. The not-so-easy part is how to convey the fact that the value is currently not set to Lua. Perhaps you could write a custom `isTempFailureTTLSet()` method?

zeha:
> Perhaps you could write a custom isTempFailureTTLSet() method?

Yeah... for symmetry maybe that also needs an unsetTempFailure() method?

This is probably not the last boost::optional that will be exposed... can we expose an Optional interface to Lua? (Happy to chat about this on IRC, haven't fully thought out this yet.)

rgacogne:
This PR will need to be rebased following the merge of #5947, sorry about that!

zeha:
Tests, Docs, Lua binding added. Turns out LuaWrapper happily exposes boost::optional!

pdns/test-dnsdistpacketcache_cc.cc
```diff
@@ -109,6 +109,52 @@ BOOST_AUTO_TEST_CASE(test_PacketCacheSimple) {
...
+    DNSQuestion dq(&a, QType::A, QClass::IN, &remote, &remote, (struct dnsheader*) query.data(), query.size(), query.size(), false);
+    bool found = PC.get(dq, a.wirelength(), 0, responseBuf, &responseBufSize, &key);
+    BOOST_CHECK_EQUAL(found, false);
+
+    // Insert with failure-TTL of 0 (-> should not enter cache).
+    PC.insert(key, a, QType::A, QClass::IN, (const char*) response.data(), responseLen, false, RCode::ServFail, boost::optional<uint32_t>(0));
+    found = PC.get(dq, a.wirelength(), pwR.getHeader()->id, responseBuf, &responseBufSize, &key, 0, true);
+    BOOST_CHECK_EQUAL(found, false);
+
+    // Insert with failure-TTL non-zero (-> should not enter cache).
```
rgacogne: s/should not enter cache/should enter cache/ ?

regression-tests.dnsdist/test_Caching.py
```diff
@@ -354,6 +354,52 @@ def testCacheDifferentCase(self):
...
+
+        (receivedQuery, receivedResponse) = self.sendUDPQuery(query, response)
+        self.assertTrue(receivedQuery)
+        self.assertTrue(receivedResponse)
+        receivedQuery.id = query.id
+        self.assertEquals(query, receivedQuery)
+        self.assertEquals(receivedResponse, response)
+
+        # next query should hit the cache
+        (receivedQuery, receivedResponse) = self.sendUDPQuery(query, response)
```
rgacogne: If you expect the query to hit the cache, I don't think you should provide a response. Otherwise it will be added to the queue and this might prove an issue later because you won't get the expected answer for the next test. It's not an issue there because you pass the exact same answer to the next test, but you might want to use:
```
self.sendUDPQuery(query, response=None, useQueue=False)
```

zeha:
Nits addressed, thanks!

Issue #6010, opened 2017-11-28
TXTRecordContent needs a "native" interface, and native storage


ahupowerdns:
Within the MOADNS infrastructure, we have the TXTRecordContent class. TXT records are odd things, and internally consist of a sequence of <character-string>s. 

On the wire, "test" is encoded on the wire as: 4test
In a zone this might appear as test or as "test".

If there are spaces, the quotes become mandatory, so "test one" always has to look like that in a zone, and is encoded as: 8test one

It is also possible to have multiple character-strings, like: "test" "one", which goes out as 4test3one

Within PowerDNS, we frequently have the need to send a message over TXT and not care about quotes. Right now, we do that by adding quotes, and passing the quoted string to the constructor. This is error prone, and we sometimes accidentally rely on the unquoted fallback.

What the TXTRecordContent needs is setContent(const vector<string>&), so we could natively set a TXT record without bothering about quotes, and still be able to do setContent({"test", "one"}).

This is complicated however by the fact that the TXTRecordContent class right now stores its contents.. in zonefile format. Yes. This means it contains escaped values, as one string. 

A proper transition would be to store TXT records in a native format, without escaping, and then setContent becomes a mostly empty function. The reading from zones file then however becomes slightly more complicated, as does getZoneRepresentation(). This is well worth it in terms of correctness however.

cmouse:
is it permissible to have `"test "one" content" "two"`?

Habbie:
> is it permissible to have "test "one" content" "two"?

no

cmouse:
and as discussed elsewhere:

`"test "one" content" "two"` should not work, but `"test \"one\" content" "two"` should. Also `"one" two "three"` should work fine.

pieterlexis:
refs:
#3503
#3150

PR #6011, opened 2017-11-28
quote server-id to hosts with - or . in their name get reported again

ahupowerdns:
### Short description
If our server name contained a dash or a dot, 'dig chaos txt server.id @yourserver' would break in auth.
This applies a minimal fix for the 4.1.early timeframe. I'll leave it up to everyone if this is 4.1.0 material or not. If anyone asks, my vote is 'yes'.

This PR should eventually be succeeded by the result of #6010.
 
### Checklist

I have:
- [X] documented the code

### DIFF
--- a/pdns/packethandler.cc
+++ b/pdns/packethandler.cc
@@ -274,7 +274,10 @@ int PacketHandler::doChaosRequest(DNSPacket *p, DNSPacket *r, DNSName &target)
         r->setRcode(RCode::Refused);
         return 0;
       }
-      rr.dr.d_content=DNSRecordContent::mastermake(QType::TXT, 1, id);
+      string tid=id;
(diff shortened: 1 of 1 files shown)


pdns/packethandler.cc
```diff
@@ -274,7 +274,10 @@ int PacketHandler::doChaosRequest(DNSPacket *p, DNSPacket *r, DNSName &target)
         r->setRcode(RCode::Refused);
         return 0;
       }
-      rr.dr.d_content=DNSRecordContent::mastermake(QType::TXT, 1, id);
+      string tid=id;
+      if(!tid.empty() && tid[0]!='"') // see #6010 however
+	tid = "\"" + tid + "\"";
```
pieterlexis: style gripe: please use spaces and braces

PR #6012, opened 2017-11-28
dnsdist: Keep the TCP connection open on cache hit, generated answers

rgacogne:
### Short description

We used to close the TCP connection right away on cases where that did not make sense:
- on a cache hit
- on a self-generated answer
- on a servfail answer caused by the lack of usable downstream servers

We still close the TCP connections on drops, dynamic blocks, lack of usable downstream servers without `setServFailWhenNoServer()` set, invalid queries, network errors..
### Checklist

I have:
- [ ] documented the code


zeha:
LGTM

PR #6021, opened 2017-11-30
query silliness

mind04:
### Short description
PowerDNS was  wasting a lot of queries while processing notifies
Use comboaddress for IPs (was strings)
Add 'supermaster' setting

### Checklist

I have:
- [ ] documented the code

### DIFF
--- a/modules/gmysqlbackend/gmysqlbackend.cc
+++ b/modules/gmysqlbackend/gmysqlbackend.cc
@@ -94,8 +94,6 @@ public:
     declare(suffix, "remove-empty-non-terminals-from-zone-query", "remove all empty non-terminals from zone", "delete from records where domain_id=? and type is null");
     declare(suffix, "delete-empty-non-terminal-query", "delete empty non-terminal from zone", "delete from records where domain_id=? and name=? and type is null");
 
-    declare(suffix,"master-zone-query","Data", "select master from domains where name=? and type='SLAVE'");
-
     declare(suffix,"info-zone-query","","select id,name,master,last_check,notified_serial,type,account from domains where name=?");
 
     declare(suffix,"info-all-slaves-query","","select id,name,master,last_check from domains where type='SLAVE'");
--- a/pdns/dnsbackend.hh
+++ b/pdns/dnsbackend.hh
@@ -90,8 +90,7 @@ struct DomainInfo
   const bool isMaster(const ComboAddress& ip)
   {
     for( const auto& master: masters) {
-      const ComboAddress caMaster(master);
-      if(ComboAddress::addressOnlyEqual()(ip, caMaster))
+      if(ComboAddress::addressOnlyEqual()(ip, master))
         return true;
     }
     return false;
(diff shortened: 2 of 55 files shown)


pdns/dnsbackend.hh
```diff
@@ -86,6 +87,16 @@ struct DomainInfo
       return DomainInfo::Native;
   }
 
+  const bool isMaster(const ComboAddress& ip)
+  {
+    for( const auto& master: masters) {
+      const ComboAddress caMaster(master);
```
ahupowerdns: this needs to compare only IP addresses, and not also the port, I'm almost sure :-)

pdns/slavecommunicator.cc
```diff
@@ -745,15 +745,7 @@ void CommunicatorClass::slaveRefresh(PacketHandler *P)
...
-    int res;
-    res=P->trySuperMasterSynchronous(&dp, tsigkeyname);
-    if(res>=0) {
-      DNSPacket *r=dp.replyPacket();
-      r->setRcode(res);
-      r->setOpcode(Opcode::Notify);
-      N->send(r);
-      delete r;
-    }
+    P->trySuperMasterSynchronous(&dp, tsigkeyname); // FIXME could use some error loging
```
ahupowerdns: was there a specific reason for removing the error packet generation? or can it no longer fail?

modules/gmysqlbackend/gmysqlbackend.cc
```diff
@@ -94,8 +94,6 @@ class gMySQLFactory : public BackendFactory
     declare(suffix, "remove-empty-non-terminals-from-zone-query", "remove all empty non-terminals from zone", "delete from records where domain_id=? and type is null");
     declare(suffix, "delete-empty-non-terminal-query", "delete empty non-terminal from zone", "delete from records where domain_id=? and name=? and type is null");
 
-    declare(suffix,"master-zone-query","Data", "select master from domains where name=? and type='SLAVE'");
```
ahupowerdns: <3 cleanup!

pdns/dnsbackend.hh
```diff
@@ -51,7 +52,7 @@ struct DomainInfo
   DNSName zone;
   time_t last_check;
   string account;
-  vector<string> masters;
+  vector<string> masters; // FIXME use ComboAddress
```
ahupowerdns: I can contribute that cleanup if you want, and commit it to this PR?

mind04: Would be awesome.

pdns/dnsbackend.hh
mind04: I almost sure you are right  ;) Fixing...

pdns/slavecommunicator.cc
mind04: this was responsible for the spurious answer (log is in the commit message). After removal of this piece of code the error is gone. There is no way to let the master know at this point it is failing.
A nice error in the log is all we can do here.

pieterlexis:
Judging by the title update, this is ready for another review?

mind04:
Yes, ready for review.

PR #6042, opened 2017-12-04
rec: Don't cache merged answers from different sections in a single packet

rgacogne:
### Short description

We incorrectly merged answers for the same qname and qtype coming from a single packet but from different sections when storing them in the cache. It resulted in duplicates for the IP addresses of some NS, for example.

### Checklist

I have:
- [ ] documented the code


phonedph1:
Seems to work for me against all the examples I had.

PR #6048, opened 2017-12-04
LDAP: fix getDomainInfo() to set this as di.backend

goestreicher:
### Checklist

I have:
- [ ] documented the code

### DIFF
--- a/modules/ldapbackend/ldapbackend.cc
+++ b/modules/ldapbackend/ldapbackend.cc
@@ -678,7 +678,6 @@ bool LdapBackend::getDomainInfo( const DNSName& domain, DomainInfo& di )
 {
   string filter;
   SOAData sd;
-  int msgid=0;
   PowerLDAP::sentry_t result;
   const char* attronly[] = {
     "sOARecord",
--- a/modules/ldapbackend/ldapbackend.cc
+++ b/modules/ldapbackend/ldapbackend.cc
@@ -585,8 +585,6 @@ void LdapBackend::getUpdatedMasters( vector<DomainInfo>* domains )
     if ( !getDomainInfo( DNSName( result["associatedDomain"][0] ), di ) )
       continue;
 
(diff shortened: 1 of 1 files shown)


zeha:
Note that getDomainInfo may be called for domains that are not stored in the backend. In this case, the DomainInfo object must NOT be changed.

goestreicher:
@zeha Yes, and although the diff doesn't show it, this case is already considered (the DomainInfo object will only be called if there's a SOARecord)

zeha:
@goestreicher Ah, the diff was just too confusing.

goestreicher:
Is there something preventing this from being merged? Just ping me if that's the case :)

Habbie:
We where in strong release mode the last few weeks. Now that we have released, and created maintenance branches, git master is back in "just do it" mode. I'll see what I can do during the coming holidays!

modules/ldapbackend/ldapbackend.cc
```diff
@@ -697,7 +695,7 @@ bool LdapBackend::getDomainInfo( const DNSName& domain, DomainInfo& di )
     // search for SOARecord of domain
     filter = "(&(associatedDomain=" + toLower( m_pldap->escape( domain.toStringRootDot() ) ) + ")(SOARecord=*))";
     m_msgid = m_pldap->search( getArg( "basedn" ), LDAP_SCOPE_SUBTREE, filter, attronly );
-    m_pldap->getSearchEntry( msgid, result );
```
rgacogne: Just a nit but it looks like `msgid` is now unused so we should just remove it.

aerique: @goestreicher Could you please address the nit mentioned above? Then I'll merge it.

aerique:
Gah, meant to press cancel to edit the merge commit but accidentally pressed the green button!

PR #6057, opened 2017-12-05
Handle bracketed IPv6 addresses without ports

zeha:
dnsdist's newServer source parameter is documented to take values
of those forms (plus some more):
  - v4 address ("192.0.2.1")
  - v6 address ("2001:DB8::1")

For consistency, bracketed addresses ("[2001:DB8::1]") should work too, and all of them
should have a test.

Example that would previously "fail":
```
newServer({name="pdns", address="[::1]:5300", source="[::1]"}):setUp()
```

Gives:
```
Dismissing source [::1] because '[::1]' is not a valid interface name
Added downstream server [::1]:5300
```

### Checklist
I have:
- [ ] documented the code

### DIFF
--- a/pdns/misc.cc
+++ b/pdns/misc.cc
@@ -719,9 +719,7 @@ int makeIPv6sockaddr(const std::string& addr, struct sockaddr_in6* ret)
     if(pos == string::npos)
       return -1;
     ourAddr.assign(addr.c_str() + 1, pos-1);
-    if (pos + 1 == addr.size()) {
-      port = 0;
-    } else {
+    if (pos + 1 != addr.size()) { // complete after ], no port specified
       if (pos + 2 > addr.size() || addr[pos+1]!=':')
         return -1;
       try {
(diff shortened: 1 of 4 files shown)


pdns/misc.cc
```diff
@@ -716,15 +716,21 @@ int makeIPv6sockaddr(const std::string& addr, struct sockaddr_in6* ret)
...
       return -1;
     ourAddr.assign(addr.c_str() + 1, pos-1);
-    try {
-      port = pdns_stou(addr.substr(pos+2));
-      portSet = true;
-    }
-    catch(std::out_of_range) {
-      return -1;
+    if (pos + 1 == addr.size()) {
+      port = 0;
```
rgacogne: It looks like setting `port` to 0 without setting `portSet` to ` true` is useless?

zeha: Hmmm yes. Will change that if()...

zeha: @rgacogne fixed this!

PR #6063, opened 2017-12-06
Add colour to diff output of pdnsutil

### LINKED ISSUE #3760: colored diffs from edit-zone
Just now I managed to miss the minus in a line that read `-_acme-challenge....`, which is all me, but many other tools today color these things. Might be a nice feature.

cyclops1982:
### Short description
Fixes #3760.
Simply adds the colours to the console output. Red for deletes. Green for additions.

### Checklist
- [ ] documented the code

### DIFF
--- a/pdns/pdnsutil.cc
+++ b/pdns/pdnsutil.cc
@@ -918,7 +918,7 @@ int editZone(DNSSECKeeper& dk, const DNSName &zone) {
   set_difference(pre.cbegin(), pre.cend(), post.cbegin(), post.cend(), back_inserter(diff), DNSRecord::prettyCompare);
   for(const auto& d : diff) {
     ostringstream str;
-    str<<'-'<< d.d_name <<" "<<d.d_ttl<<" IN "<<DNSRecordContent::NumberToType(d.d_type)<<" "<<d.d_content->getZoneRepresentation(true)<<endl;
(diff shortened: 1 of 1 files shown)


rgacogne:
> I think we'll need to check if we are on a tty so we don't write escape codes to a file or a pipe

I do agree on the principle, but note that we are in edit-zone so file or pipe is unlikely. The existing code also already uses color for at least one message.

pieterlexis:
> I do agree on the principle, but note that we are in edit-zone so file or pipe is unlikely. The existing code also already uses color for at least one message.

Right, I'll approve this for now. But perhaps we need to correct this in the future.

beamerblvd:
@cyclops1982, @pieterlexis: Was `if (isatty(fileno(stdin))) {` ever added to this after the fact? I think I've come across a problem with this change. I'm calling this utility using `docker exec` through the Docker API (HTTP), and the data I'm getting back has garbled characters at the beginning of each line. I suspect it's trying to colorize the output that's coming back to me? I'm calling the Docker API with `"tty": false`.

beamerblvd:
You can ignore that comment above. Apparently I actually needed to have `"tty": true`, and it's a function of the way Docker works, not related to this change here. (I wasn't even calling the command to which this change was made; I was calling `check-zone`.) We still should probably consider this possibility and add `isatty` in there somewhere.

Habbie:
Hi @beamerblvd,

you've commented on a closed PR. While I see your comment now, nobody will ever notice it again. Can you please open a separate ticket? Thank you!

PR #6068, opened 2017-12-07
dnsdist: Add DNSQuestion:getEDNSOptions() to access incoming EDNS options

rgacogne:
### Short description

Enabling access to the raw binary content of incoming `EDNS` options from `Lua` rules via `DNSQuestion::getEDNSOptions()`. This call returns a table with option code as keys and `EDNSOptionView`s as values. The size of the raw binary content can be retrieved via `EDNSOptionView::size` and the content with `EDNSOptionView::getContent()`, the later returning the binary content as a string.

### Checklist

I have:
- [x] documented the code

### DIFF
--- a/regression-tests.dnsdist/dnsdisttests.py
+++ b/regression-tests.dnsdist/dnsdisttests.py
@@ -491,18 +491,25 @@ class DNSDistTest(unittest.TestCase):
             for option in received.options:
                 self.assertEquals(option.otype, 10)
 
-    def checkMessageEDNSWithECS(self, expected, received):
+    def checkMessageEDNSWithECS(self, expected, received, additionalOptions=0):
         self.assertEquals(expected, received)
         self.assertEquals(received.edns, 0)
-        self.assertEquals(len(received.options), 1)
-        self.assertEquals(received.options[0].otype, clientsubnetoption.ASSIGNED_OPTION_CODE)
+        self.assertEquals(len(received.options), 1 + additionalOptions)
+        hasECS = False
+        for option in received.options:
+            if option.otype == clientsubnetoption.ASSIGNED_OPTION_CODE:
+                hasECS = True
+            else:
+                self.assertNotEquals(additionalOptions, 0)
+
         self.compareOptions(expected.options, received.options)
+        self.assertTrue(hasECS)
 
-    def checkQueryEDNSWithECS(self, expected, received):
-        self.checkMessageEDNSWithECS(expected, received)
+    def checkQueryEDNSWithECS(self, expected, received, additionalOptions=0):
+        self.checkMessageEDNSWithECS(expected, received, additionalOptions)
 
-    def checkResponseEDNSWithECS(self, expected, received):
-        self.checkMessageEDNSWithECS(expected, received)
+    def checkResponseEDNSWithECS(self, expected, received, additionalOptions=0):
+        self.checkMessageEDNSWithECS(expected, received, additionalOptions)
 
     def checkQueryEDNSWithoutECS(self, expected, received):
         self.checkMessageEDNSWithoutECS(expected, received)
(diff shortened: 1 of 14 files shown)


Habbie:
Works as advertised. Still needs a code review.

regression-tests.dnsdist/dnsdisttests.py
```diff
@@ -451,18 +451,24 @@ def checkMessageEDNSWithoutECS(self, expected, received, withCookies=0):
...
                 self.assertEquals(option.otype, 10)
 
-    def checkMessageEDNSWithECS(self, expected, received):
+    def checkMessageEDNSWithECS(self, expected, received, additionalOptions=0):
         self.assertEquals(expected, received)
         self.assertEquals(received.edns, 0)
-        self.assertEquals(len(received.options), 1)
-        self.assertEquals(received.options[0].otype, clientsubnetoption.ASSIGNED_OPTION_CODE)
+        self.assertEquals(len(received.options), 1 + additionalOptions)
+        hasECS = False
```
pieterlexis: `hasECS` seems unused

zeha:
This is going to conflict a bit with #6170 - both change the DNSQuestion constructor params.

zeha:
Already has merge conflicts:
```
dnsdist-tcp.cc:365:152: error: no matching function for call to ‘DNSResponse::DNSResponse(const DNSName*&, const uint16_t&, const uint16_t&, const ComboAddress*&, const ComboAddress*&, dnsheader*, size_t&, uint16_t&, bool, timespec*)’
           DNSResponse dr(dq.qname, dq.qtype, dq.qclass, dq.local, dq.remote, reinterpret_cast<dnsheader*>(query), dq.size, dq.len, true, &queryRealTime);
...
```

MangeshBhamre:
Habbie, 
I wish to try this code. Could you please fix the code merge conflict? 

Thanks 
Mangesh

Habbie:
@MangeshBhamre I may have time for that this week, but you could also check out rgacogne's branch and build that, if you just want to try out this feature. You'll get a slightly older dnsdist that way, of course.

rgacogne:
Rebased!

rgacogne:
Note than for some EDNS option codes, it might make sense to have several options with the same code in one OPT RR, which this API does not allow, since it mirrors the one currently existing in the recursor. We might want to change that before merging that PR.

Habbie:
```
   self.assertFalse(self._protobufQueue.empty())
AssertionError: True is not false
```

IIRC a rebase should correct this.

zeha:
(New) conflicts :(

rgacogne:
I need to think the API through so we can handle more than one value for the same option code, postponing to 1.3.x

rgacogne:
Updated to support more than one value per `EDNS` option, rebased on top of #6744 to use the same code base and API (so this PR will need a rebase once #6744 has been merged).

rgacogne:
Rebased after the merge of #6744.

rgacogne:
Rebased.

eterry1388:
I'm super excited to try this out. When will it be published here? https://repo.powerdns.com/debian/dists/jessie-dnsdist-master

MangeshBhamre:
How can I access the content of ecsOption? Here is my code snippet. content is nil always

function getCpeID(dq)
   local cpeID
   if dq:getEDNSOptions() ~= nil then
        local ednsOptions = dq:getEDNSOptions()

        local ecsOption = ednsOptions[65073]
           if ecsOption then
                -- Assuming 'content' is the correct way to access the data you're interested in
                errlog("EDNS Option Code: " .. type(ecsOption)  .. tostring(ecsOption.content) )
                cpeID = tostring(ecsOption.content)
            end
    end
    return cpeID
end

Habbie:
Hello,

Support questions are handled in our other online communities: [GitHub Discussions](https://github.com/PowerDNS/pdns/discussions), IRC, and our mailing lists. Please see https://www.powerdns.com/opensource.html for information about those.

MangeshBhamre:
Done. Thanks @Habbie https://github.com/PowerDNS/pdns/discussions/13891

PR #6072, opened 2017-12-08
dnsdist: add inClientStartup() function

Habbie:
### Short description
This allows the user to, for example, print certain things during daemon startup but not while connecting a client.

### Checklist

I have:
- [ ] documented the code

### DIFF
--- a/pdns/dnsdistdist/docs/reference/config.rst
+++ b/pdns/dnsdistdist/docs/reference/config.rst
@@ -118,6 +118,10 @@ Control Socket, Console and Webserver
 
   :param str address: An IP address with optional port. By default, the port is 5199.
 
+.. function:: inClientStartup()
+
+  Returns true while the console client is parsing the configuration.
+
 .. function:: makeKey()
 
   Generate and print an encryption key.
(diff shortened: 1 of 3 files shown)


pdns/dnsdistdist/docs/reference/config.rst
```diff
@@ -118,6 +118,10 @@ Control Socket, Console and Webserver
 
   :param str address: An IP address with optional port. By default, the port is 5199.
 
+.. function:: inClientStartup()
+
+  Returns true while the console client is parsing the configuration.
```
rgacogne: Doesn't it return `true` also after the client is done parsing the configuration? I think we have something like `g_configurationDone` to indicate whether we are done parsing the configuration.

Habbie: It returns false in the client console. I couldn't imagine a situation in which somebody could call this in the client state after the `>` prompt was shown so I did not think about it.

rgacogne: You're right, after the client is done parsing the configuration all commands are passed to the server over the console, so my comment made no sense!

Habbie: pushed something for this

Habbie:
@zeha did you want any change in the logic here?

zeha:
@Habbie I couldn't find a variant that was strictly better, so no.

PR #6078, opened 2017-12-10
Make everything compile and run on NetBSD, including on multi-homed hosts

tih:
### Short description

These are the minimal changes to make everything (auth, recursor, and
dnsdist) compile and run properly on NetBSD.  The changes fall into
three groups:

1) A number of existing "#if defined" lines already in place for other
   BSD variants needed to be expanded with "|| defined(__NetBSD__)" to
   get the same effect here.

2) The pthread CPU affinity code in pdns/misc.cc needed changes,
   because the underlying cpuset_t is opaque in NetBSD, and one must
   use function calls operating on a pointer to a CPU set returned by
   cpuset_create(), instead of declaring a local struct and operating
   on that.

3) The use of socket options IP_PKTINFO et al is slightly different in
   NetBSD.  It can be used to make recvmsg() supply information about
   the source address of an incoming packet, or to make sendmsg() use
   a specific source address for an outgoing packet.  Thus, it is a
   direct parallel to the IPV6_PKTINFO option, and uses similarly
   named struct members.  To get information about the destination
   address of an incoming packet (which multi-homed servers will need
   to be able to respond from the right address), one would use the
   IP_RECVPKTINFO option instead -- while Linux conflates these two
   features into one, by adding a struct member "ipi_spec_dst" for the
   "specific destination" of an incoming packet.  (It then overloads
   that field by using it for the "specific source" of an outgoing
   packet, which looks somewhat inelegant to me.)

   So the IP_PKTINFO option can be used to set the source for an
   outgoing packet for sendmsg(), but we have to use the "ipi_addr"
   struct member instead of "ipi_spec_dst", so I #ifdef that for
   NetBSD.  For getting the destination address of an incoming packet,
   PowerDNS already has code to use the IP_RECVDSTADDR option instead
   of IP_PKTINFO, so I just added an "&& !defined(__NetBSD__) to the
   test for IP_PKTINFO being available.  (Pretending that IP_PKTINFO
   isn't available is fine here, but wouldn't help for setting the
   source address of outgoing packets, as IP_SENDSRCADDR (which there
   was already code in place to use) is not available on NetBSD.)

   Another tiny discrepancy: the existing code uses the CMSG_DATA()
   macro to fetch optional information for packets, of course.  The
   implementation of this on NetBSD clashes with the declaration
   "const struct cmsghdr* cmsg;", and I had to use CCMSG_DATA()
   instead.

The result works as it should, and is -- I've verified that the
servers now respond correctly to queries addressed to them on IPv4
addresses other than the one on the interface the query arrives on.

One quirk I haven't done anything about is the difficulty configure
experiences finding libedit on NetBSD, where it is in the base
installation.  The workaround is to configure dnsdist by saying
"./configure LIBEDIT_CFLAGS=-I/usr/include LIBEDIT_LIBS=-L/usr/lib -ledit"
but I'm sure something could be done to configure to help it.  I don't
know the tool well enough to try, though.

### Checklist

I have:
- [ ] documented the code

### DIFF
--- a/pdns/dns.hh
+++ b/pdns/dns.hh
@@ -130,7 +130,7 @@ struct EDNS0Record
 
 static_assert(sizeof(EDNS0Record) == 4, "EDNS0Record size must be 4");
 
-#if defined(__FreeBSD__) || defined(__APPLE__) || defined(__OpenBSD__) || defined(__DragonFly__) || defined(__FreeBSD_kernel__)
+#if defined(__FreeBSD__) || defined(__APPLE__) || defined(__OpenBSD__) || defined(__DragonFly__) || defined(__FreeBSD_kernel__) || defined(__NetBSD__)
 #include <machine/endian.h>
 #elif __linux__ || __GNU__
 # include <endian.h>
--- a/pdns/dnsdist-console.cc
+++ b/pdns/dnsdist-console.cc
@@ -23,7 +23,7 @@
 #include "sodcrypto.hh"
 #include "pwd.h"
 
-#if defined (__OpenBSD__)
+#if defined (__OpenBSD__) || defined(__NetBSD__)
 #include <readline/readline.h>
 #include <readline/history.h>
 #else
--- a/pdns/dnsdist.cc
+++ b/pdns/dnsdist.cc
@@ -27,7 +27,7 @@
 #include <limits>
 #include "dolog.hh"
 
-#if defined (__OpenBSD__)
+#if defined (__OpenBSD__) || defined(__NetBSD__)
 #include <readline/readline.h>
 #else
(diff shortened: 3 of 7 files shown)


tih:
Hmpf. There are improvements that can still be made, here. It turns out that the documentation doesn't match the implementation in NetBSD, so some of what I thought were differences between Linux and NetBSD here turns out to be wrong. This means I can get it to work with fewer tweaks.

Don't merge this just yet, in other words.

tih:
I ended up modifying NetBSD, to make these socket options compatible with Linux (and FreeBSD, and Solaris). I'd like to create a new pull request instead of making this one even more complicated.

Closing this pull request now, in other words.

PR #6092, opened 2017-12-13
rec: Fix the computation of the closest encloser for positive answers

rgacogne:
### Short description

When the positive answer is expanded from a wildcard with `NSEC3`, the closest encloser is not always parent of the `qname`, depending on the number of labels in the initial wildcard.

### Checklist

I have:
- [ ] documented the code

### DIFF
--- a/pdns/syncres.cc
+++ b/pdns/syncres.cc
@@ -1898,7 +1898,7 @@ vState SyncRes::validateRecordsWithSigs(unsigned int depth, const DNSName& qname
   return Bogus;
 }
 
-RCode::rcodes_ SyncRes::updateCacheFromRecords(unsigned int depth, LWResult& lwr, const DNSName& qname, const QType& qtype, const DNSName& auth, bool wasForwarded, const boost::optional<Netmask> ednsmask, vState& state, bool& needWildcardProof)
+RCode::rcodes_ SyncRes::updateCacheFromRecords(unsigned int depth, LWResult& lwr, const DNSName& qname, const QType& qtype, const DNSName& auth, bool wasForwarded, const boost::optional<Netmask> ednsmask, vState& state, bool& needWildcardProof, unsigned int& wildcardLabelsCount)
 {
   tcache_t tcache;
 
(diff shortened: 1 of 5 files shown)


pdns/validate.cc
```diff
@@ -306,9 +306,12 @@ static bool provesNSEC3NoWildCard(DNSName wildcard, uint16_t const qtype, const
   useful when we have a positive answer synthetized from a wildcard and we only need to prove that the exact
   name does not exist.
 */
-dState getDenial(const cspmap_t &validrrsets, const DNSName& qname, const uint16_t qtype, bool referralToUnsigned, bool wantsNoDataProof, bool needWildcardProof)
+dState getDenial(const cspmap_t &validrrsets, const DNSName& qname, const uint16_t qtype, bool referralToUnsigned, bool wantsNoDataProof, bool needWildcardProof, unsigned int wildcardLabelsCount)
 {
   bool nsec3Seen = false;
+  if (!needWildcardProof && wildcardLabelsCount == 0) {
+    throw std::runtime_error("Invalid wildcard labels count for the validation of a positive answer synthetized from a wildcard");
```
pieterlexis: ImmidiateServFailException or PDNSException is maybe better?

pdns/syncres.cc
```diff
@@ -1898,7 +1898,7 @@ vState SyncRes::validateRecordsWithSigs(unsigned int depth, const DNSName& qname
   return Bogus;
 }
 
-RCode::rcodes_ SyncRes::updateCacheFromRecords(unsigned int depth, LWResult& lwr, const DNSName& qname, const QType& qtype, const DNSName& auth, bool wasForwarded, const boost::optional<Netmask> ednsmask, vState& state, bool& needWildcardProof)
+RCode::rcodes_ SyncRes::updateCacheFromRecords(unsigned int depth, LWResult& lwr, const DNSName& qname, const QType& qtype, const DNSName& auth, bool wasForwarded, const boost::optional<Netmask> ednsmask, vState& state, bool& needWildcardProof, unsigned int& wildcardLabelsCount)
```
pieterlexis: perhaps we can use `uint8_t` for the labelcount. The RRSIG labels field is never bigger than this.

rgacogne: That would work, the current type is derived from the return value of `DNSName::countLabels()`.

pdns/validate.cc
rgacogne: `ImmediateServFailException` is in `syncres.hh` so it would be an issue for the authoritative server, which uses `validate.cc`. I'll update to throw a `PDNSException`.

PR #6095, opened 2017-12-14
 rec: Pass the correct buffer size to arecvfrom()

rgacogne:
### Short description

The size we used to pass to `arecvfrom()` along with our buffer was off by one byte, resulting in the last byte of our buffer to be uninitialized for answers exactly matching our outgoing buffer size. Since we passed the correct size to `MOADNSParser`, we were reading one bye of uninitialized memory for such answers.
This caused issue with some authoritative servers sending an answer of our exact buffer size, causing a parsing error. We would then retry without `EDNS`, causing `DNSSEC` validation failures for some domains on such authoritative servers.

### Checklist

I have:
- [ ] documented the code

### DIFF
--- a/pdns/lwres.cc
+++ b/pdns/lwres.cc
@@ -161,7 +161,7 @@ int asyncresolve(const ComboAddress& ip, const DNSName& domain, int type, bool d
   
     // sleep until we see an answer to this, interface to mtasker
     
-    ret=arecvfrom(reinterpret_cast<char *>(buf.get()), bufsize-1,0, ip, &len, qid,
+    ret=arecvfrom(reinterpret_cast<char *>(buf.get()), bufsize, 0, ip, &len, qid,
                   domain, type, queryfd, now);
   }
   else {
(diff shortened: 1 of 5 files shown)


zeha:
While this looks right, I must say there is a lot of copying of the packet going on.

`handleUDPServerResponse` (set up as the MT handler by `asendto`) always makes a new char buffer of `g_outgoingEDNSBufsize` size, and sticks that into a std::string. `arecvfrom` then takes that string and copies it into the `scoped_array<unsigned char>` buffer (passed as a `char*`) supplied by `asyncresolve`. Each of these functions also has its own idea about buffer sizes. That doesn't make it any clearer on what is going on ...

rgacogne:
> While this looks right, I must say there is a lot of copying of the packet going on.

I completely agree but my feeling is that making that better belongs to a different PR :)

zeha:
Agreed, probably better for backporting too.

PR #6102, opened 2017-12-18
add actual EDNS buffer size logging, not just our interpretation

ahupowerdns:
### Short description
DNSPacket exposes the aptly named getMaxReplyLen() function which we mistakenly thought meant this was the supplied EDNS buffer size. In some places we want to know the actual raw EDNS buffer size. This PR adds that, and hooks it up to our logging.

### Checklist

I have:
- [ ] documented the code

### DIFF
--- a/pdns/dnspacket.cc
+++ b/pdns/dnspacket.cc
@@ -121,7 +121,7 @@ DNSPacket::DNSPacket(const DNSPacket &orig)
   d_tsigtimersonly = orig.d_tsigtimersonly;
   d_trc = orig.d_trc;
   d_tsigsecret = orig.d_tsigsecret;
-  
+  d_ednsRawPacketSizeLimit = orig.d_ednsRawPacketSizeLimit;
   d_havetsig = orig.d_havetsig;
   d_wrapped=orig.d_wrapped;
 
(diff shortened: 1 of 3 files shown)


pdns/dnspacket.cc
```diff
@@ -580,9 +580,10 @@ try
     }
     d_ednsversion = edo.d_version;
     d_ednsrcode = edo.d_extRCode;
-  }
+ }
```
rgacogne: Looks like the indentation is now wrong?

pdns/dnspacket.hh
```diff
@@ -162,7 +162,8 @@ public:
   bool checkForCorrectTSIG(UeberBackend* B, DNSName* keyname, string* secret, TSIGRecordContent* trc) const;
 
   static bool s_doEDNSSubnetProcessing;
-  static uint16_t s_udpTruncationThreshold; //2
+  static uint16_t s_udpTruncationThreshold; 
+  int d_ednsRawPacketSizeLimit; // only used for Lua record
```
rgacogne: This comment might not be true anymore :-)

pdns/dnspacket.cc
```diff
@@ -547,13 +547,13 @@ try
...
   d_haveednssubnet = false;
   d_haveednssection = false;
-  
 
   if(getEDNSOpts(mdp, &edo)) {
     d_haveednssection=true;
     /* rfc6891 6.2.3:
        "Values lower than 512 MUST be treated as equal to 512."
     */
+    d_ednsRawPacketSizeLimit=edo.d_packetsize;
```
zeha: there's a comment above saying:
```
// ANY OPTION WHICH *MIGHT* BE SET DOWN BELOW SHOULD BE CLEARED FIRST!
```

does this apply?

Issue #6115, opened 2017-12-22
Caching is providing incorrect responses with gmysql backend and a stored procedure


robsheldon:
- Program: Authoritative
 - Issue type: Bug report

### Short description
I can't get PowerDNS to provide consistent answers when using a stored procedure with the gmysql backend. It keeps providing different/incorrect answers when the backend data hasn't changed.

Relevant pdns.local.conf:
```
# Something in pdns caching is screwy
max-cache-entries=1
max-packet-cache-entries=1
cache-ttl=1
query-cache-ttl=1

# Launch gmysql backend
launch=gmysql:genericdns,gmysql:ridenips,gmysql:ridenmagic

# Used for generic DNS handling
gmysql-genericdns-socket=/var/run/mysqld/mysqld.sock
gmysql-genericdns-dbname=dns
gmysql-genericdns-user=...
gmysql-genericdns-password=...

# Used specifically for riden-jailed IP addresses
gmysql-ridenips-socket=/var/run/mysqld/mysqld.sock
gmysql-ridenips-dbname=riden
gmysql-ridenips-user=...
gmysql-ridenips-password=...

# Used for netrange scoring
gmysql-ridenmagic-socket=/var/run/mysqld/mysqld.sock
gmysql-ridenmagic-dbname=riden
gmysql-ridenmagic-user=pdns
gmysql-ridenmagic-password=...
gmysql-ridenmagic-any-id-query=CALL network_query(?, ?)
```

The network_query stored procedure looks like:
```
CREATE PROCEDURE network_query (name VARCHAR(255), whoknows VARCHAR(1))
network_query:begin
    declare ip_long int(10) unsigned default 0;
    declare return_value varchar(15) default '127.0.0.0';
    declare netrange_score float default 0;
    declare search int(10) unsigned default 0;
    declare offset int unsigned default 0;
    set offset = instr(name, '.rtidefense.net');
    if offset = 0 then
        select null as content, 600 as ttl, null as prio, 'A' as type, null as domain_id, null as disabled, name, null as auth;
        LEAVE network_query;
    end if;
    set search = inet_aton(substring(name, 1, offset - 1));
    if search is null then
        select null as content, 600 as ttl, null as prio, 'A' as type, 1 as domain_id, null as disabled, name, 1 as auth;
        LEAVE network_query;
    end if;
[...]

PR #6121, opened 2017-12-27
Add releaseStatement() to the end of hasNextRow()

robsheldon:
Fixes stale data returned by MySQL when calling stored procedures.

### Short description
Closes issue #6115 by adding releaseStatement() to the end of the "#if MYSQL_VERSION_ID >= 50500" block. Without this, mysql_stmt_close() is never called for stored procedures, which appears to cause MySQL to return stale results from past calls to msql_stmt_fetch().

### Checklist
I have:
- [ ] documented the code

### DIFF
--- a/modules/gmysqlbackend/smysql.cc
+++ b/modules/gmysqlbackend/smysql.cc
@@ -268,6 +268,7 @@ public:
         }
         mysql_stmt_free_result(d_stmt);
       }
+      releaseStatement();
     }
 #endif
     return this; 


rgacogne:
> My reading of the code here was that releaseStatement() would only get called after mysql_stmt_next_result() returned !0 (i.e., an error occurred or no error and no more results to read).

It will also get executed when the `break` statement is reached, with no error and something to read, unless I'm mistaken :)

robsheldon:
Whoop, you're right. I should've seen that. Since your commit fixes this behavior anyway, I'll revert my change to my codebase and apply yours and test. If there are any issues I'll just reply to your PR.

Closing this one.

PR #6124, opened 2017-12-28
Improve tests and two bugfixes

### LINKED ISSUE #6101: RecordTextReader::xfrIP accepts invalid addresses
- Program: all that parse text representations of DNS records
 - Issue type: Bug report

### Short description

`RecordTextReader::xfrIP`:
- will read an incomplete IPv4 address and pad it with zeroes from the left.  E.g. it will read `10.0` as `0.0.10.0`, which is certainly not what users would expect. Short forms would be padded from the right (i.e. `10.0.0.0`), but RFC 1035 only allows "four decimal numbers separated by dots without any imbedded spaces" anyway.
- will read a 4th dot as "field separator" (like white space); if the next field would allow `.` as first character parsing would actually succeed.  This is probably not intentional.  Currently only base64 encoded fields might follow an IPv4 address, which fails reading a `.`, so isn't actually a problem right now.

### LINKED ISSUE #6120: RRSIG doesn't accept timestamps in seconds since epoch
- Program: Authoritative
 - Issue type: Bug report

### Short description

For `RRSIG` the following text is not accepted:

    SOA 8 3 300 6666 7777 54216 bar1.example.net. AAAA

Only when the timestamps are in `YYYYMMDDHHmmSS` format it is accepted, like :

    SOA 8 3 300 19900101000000 19900101000000 54216 bar1.example.net. AAAA

[RFC 4034 section 3.2](https://tools.ietf.org/html/rfc4034#section-3.2) states:

> The Signature Expiration Time and Inception Time field values MUST be represented either as an unsigned decimal integer indicating seconds since 1 January 1970 00:00:00 UTC, or in the form YYYYMMDDHHmmSS in UTC, where:
> [...]
> Note that it is always possible to distinguish between these two formats because the YYYYMMDDHHmmSS format will always be exactly 14 digits, while the decimal representation of a 32-bit unsigned integer can never be longer than 10 digits.

### Environment
 - Operating system: debian
 - Software version: `pdns-server 4.0.3-1+deb9u2`
 - Software source: debian (stable)

stbuehler:
### Short description

- unit tests: fix uninitialized pointer
- test-dnsrecords: simplify CASE_L to CASE_S when inval == zoneval
- test-dnsrecords: handle TSIG like everything else
- test-dnsrecords: remove duplicate tests
- test-dnsrecords: test wire instead of zone for truncated AAAA wire value
- test-dnsrecords: use enum class for case_type_enum_t, fix compile error testing wrong tuple entry
- test-dnsrecords: warn that wire tests are not actually supported
- test-dnsrecords: fix reversed logic for broken bad-values check
- test-dnsrecords: add tests related to #6101
- fix xfrIP to reject invalid ips (fixes #6101)
- accept seconds since epoch in RRSIG timestamps too (fixes #6120)

### Checklist

I have:
- [x] compiled and tested this code (ran: `testrunner -t test_dnsrecords_cc`)
- [ ] documented the code


stbuehler:
I rebased locally to current master (no conflicts), `testrunner -t test_dnsrecords_cc` still good; can push if desired.

I also have a second queue of patches that's waiting for this to be merged: https://github.com/stbuehler/pdns/compare/improve-tests-and-two-bugfixes...stbuehler:improve-tests-part-two

PR #2603, opened 2015-06-18
Expose ResponseStats via REST API

zeha:
With the intention to fix #2602

PR #6136, opened 2018-01-03
regression-tests.api: clean up runtests.py wrapper

zeha:
### Short description
Cleans up `runtests.py` a bit. Extracted and improved from #2603.

Now shows which commands and tests are executed, and the common server arguments are prepared in a single place.

### Checklist

I have:
- [ ] documented the code


zeha:
Right, everything gets quoted, compared to bash which (under set -x) apparently only quotes the "relevant" arguments:

     + echo 'foo bar' foo

PR #6139, opened 2018-01-03
Forbid creating algo 5/8/10 keys with out-of-spec sizes

zeha:
### Short description
Previously, creating keys with tiny sizes would fail because OpenSSL would not allow it. Newer versions apparently accept sizes lower than the RFCs allow.

Might need to go into the upgrade notes?

### Checklist

I have:
- [ ] documented the code

### DIFF
--- a/pdns/opensslsigners.cc
+++ b/pdns/opensslsigners.cc
@@ -278,14 +278,14 @@ DNSCryptoKeyEngine::storvector_t OpenSSLRSADNSCryptoKeyEngine::convertToISCVecto
 
   string algorithm=std::to_string(d_algorithm);
   switch(d_algorithm) {
-    case 5:
-    case 7:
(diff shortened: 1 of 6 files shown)


pdns/opensslsigners.cc
```diff
@@ -211,6 +211,19 @@ class OpenSSLRSADNSCryptoKeyEngine : public DNSCryptoKeyEngine
 
 void OpenSSLRSADNSCryptoKeyEngine::create(unsigned int bits)
 {
+  if (d_algorithm == 5 && (bits < 512 || bits > 4096)) {
```
Habbie: Don't we have named constants for the algorithm?

zeha: Indeed we do. Not sure why half of the code is using the numbers directly.

PR #6146, opened 2018-01-05
Changes to compile and run on NetBSD

tih:
### Short description

These are the minimal changes to make the authoritative server, the
recursor, and dnsdist compile and run properly on NetBSD.

The changes fall into three groups:

1) A number of existing "#if defined" lines already in place for other
   BSD variants needed to be expanded with "|| defined(__NetBSD__)" to
   get the same effect here.

2) The pthread CPU affinity code in pdns/misc.cc needed changes,
   because the underlying cpuset_t is opaque in NetBSD, and one must
   use function calls operating on a pointer to a CPU set returned by
   cpuset_create(), instead of declaring a local struct and operating
   on that.

3) The socket control message interface uses the CMSG_DATA() macro to
   fetch optional information for packets, but expects that to assume
   a "const struct cmsghdr" declaration.  NetBSD has a CCMSG_DATA()
   that does this.

4) The implementation of socket options IP_PKTINFO et al in NetBSD was
   not compatible with anything, and this has required source code
   workarounds in software that uses them.  As part of this project,
   I submitted a set of fixes to NetBSD that makes it source code
   compatible with Linux, Solaris, and FreeBSD -- the two former by
   making the IP_PKTINFO support detect which of their behaviours the
   programmer expects, the latter by implementing IP_SENDSRCADDR.  My
   modifications are expected to be in NetBSD 8.0 when it is released.

   My change here is simply to "#undef IP_PKTINFO" unless the version
   of NetBSD we're compiling on is new enough to have my changes.  For
   older versions, server sockets bound to INADDR_ANY won't work right.
   This has been the way NetBSD "pkgsrc" has handled this so far, too,
   so it's not likely to surprise anyone.

The result works as it should, and I've verified that the servers now
respond correctly to queries addressed to them on IPv4 addresses other
than the one that's the primary address of the interface the query
arrives on.

One quirk I haven't done anything about is the difficulty configure
experiences finding libedit on NetBSD, where it is in the base
installation.  The workaround is to configure dnsdist by saying
./configure LIBEDIT_CFLAGS="-I/usr/include" "LIBEDIT_LIBS=-L/usr/lib -ledit"
but I'm sure something could be done to configure to help it.  I don't
know the tool well enough to try, though.

### Checklist

I have:
- [ ] documented the code

### DIFF
--- a/pdns/dns.hh
+++ b/pdns/dns.hh
@@ -130,7 +130,7 @@ struct EDNS0Record
 
 static_assert(sizeof(EDNS0Record) == 4, "EDNS0Record size must be 4");
 
-#if defined(__FreeBSD__) || defined(__APPLE__) || defined(__OpenBSD__) || defined(__DragonFly__) || defined(__FreeBSD_kernel__)
+#if defined(__FreeBSD__) || defined(__APPLE__) || defined(__OpenBSD__) || defined(__DragonFly__) || defined(__FreeBSD_kernel__) || defined(__NetBSD__)
 #include <machine/endian.h>
 #elif __linux__ || __GNU__
 # include <endian.h>
--- a/pdns/dnsdist-console.cc
+++ b/pdns/dnsdist-console.cc
@@ -23,7 +23,7 @@
 #include "sodcrypto.hh"
 #include "pwd.h"
 
-#if defined (__OpenBSD__)
+#if defined (__OpenBSD__) || defined(__NetBSD__)
 #include <readline/readline.h>
 #include <readline/history.h>
 #else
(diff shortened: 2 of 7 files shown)


tih:
Re-reading the patch, there's a clarification I guess I should have made: in pdns/misc.cc, there was an #ifdef, #endif, #ifdef, #endif sequence, that I replace with #if, #elif, #endif. This sequence would have created unnecessary extra code on NetBSD, it being the only OS where both of the conditions would be true. My modifications, commited to NetBSD-current, and expected to be released with NetBSD 8, make both the IP_PKTINFO and the IP_SENDSRCADDR control messages available for use with sendmsg().

PR #6147, opened 2018-01-05
Add ERCodeRule

zeha:
### Short description
Add ERCodeRule to match on EDNS Extended RCodes.

Separate rule to avoid performance cost for non-EDNS packets. Unclear if this is a good idea for correctness.

### Checklist
I have:
- [ ] documented the code

### DIFF
--- a/pdns/dnsdist-lua-rules.cc
+++ b/pdns/dnsdist-lua-rules.cc
@@ -724,7 +724,7 @@ private:
 class RCodeRule : public DNSRule
 {
 public:
-  RCodeRule(int rcode) : d_rcode(rcode)
+  RCodeRule(uint8_t rcode) : d_rcode(rcode)
   {
   }
   bool matches(const DNSQuestion* dq) const override
@@ -736,13 +736,13 @@ public:
     return "rcode=="+RCode::to_s(d_rcode);
   }
 private:
-  int d_rcode;
+  uint8_t d_rcode;
 };
 
 class ERCodeRule : public DNSRule
 {
 public:
-  ERCodeRule(int rcode) : d_rcode(rcode & 0xF), d_extrcode(rcode >> 4)
+  ERCodeRule(uint8_t rcode) : d_rcode(rcode & 0xF), d_extrcode(rcode >> 4)
   {
   }
   bool matches(const DNSQuestion* dq) const override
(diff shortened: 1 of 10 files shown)


pdns/dnsdist-lua-rules.cc
```diff
@@ -738,6 +739,52 @@ class RCodeRule : public DNSRule
...
+class ERCodeRule : public DNSRule
+{
+public:
+  ERCodeRule(int rcode) : d_rcode(rcode & 0xF), d_extrcode(rcode >> 4)
+  {
+  }
+  bool matches(const DNSQuestion* dq) const override
+  {
+    // avoid parsing EDNS OPT RR when not needed.
+    if (d_rcode != dq->dh->rcode) {
```
rgacogne: Given that `dq->dh->rcode` is `unsigned`, perhaps `d_rcode` should be too? Same for `d_extrcode` since `edns0.extRCode` is an `uint8_t`.

pdns/dnsdist-lua-rules.cc
```diff
@@ -738,6 +739,52 @@ class RCodeRule : public DNSRule
...
+      return false;
+    }
+
+    char * optStart = NULL;
+    size_t optLen = 0;
+    bool last = false;
+    int res = locateEDNSOptRR((char*)dq->dh, dq->len, &optStart, &optLen, &last);
```
rgacogne: `const_cast<char*>(reinterpret_cast<const char*>(dq->dh))` ? I don't think we'll manage to get rid of the existing C-style casts, but it would be nice to add new ones as few as possible :)

pdns/dnsdist-lua-rules.cc
```diff
@@ -738,6 +739,52 @@ class RCodeRule : public DNSRule
...
+      return false;
+    }
+
+    if (*optStart != 0) {
+      // OPT RR Name != '.'
+      return false;
+    }
+    EDNS0Record edns0;
+    static_assert(sizeof(EDNS0Record) == sizeof(uint32_t), "sizeof(EDNS0Record) must match sizeof(uint32_t) AKA RR TTL size");
+    memcpy(&edns0, optStart + 5, sizeof edns0);
```
rgacogne: Adding a comment explaining where does the `5` comes from might be nice for later review.

zeha:
Nits addressed

PR #6155, opened 2018-01-07
Update EDNS Option code list

zeha:
### Short description
Update list in enum, correct DAU code.

From https://www.iana.org/assignments/dns-parameters/dns-parameters.xhtml#dns-parameters-11

### Checklist
I have:
- [ ] documented the code

### DIFF
--- a/pdns/ednsoptions.hh
+++ b/pdns/ednsoptions.hh
@@ -26,7 +26,7 @@
 
 struct EDNSOptionCode
 {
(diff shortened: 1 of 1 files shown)


Habbie:
@rgacogne why backport to stable? these two constants are not reference anywhere in the code

rgacogne:
They are not today, no, but I'm worried we might someday backport some piece of code that uses them, and not check if they are correct on this branch.

PR #6157, opened 2018-01-08
Lua2 backend

cmouse:
### Short description
This is a rewrite of the lua backend. It uses AuthLua4 as basis and more strongly typed access using LuaContext.

### Checklist
I have:
- [x] documented the code


docs/backends/lua2.rst
```diff
@@ -0,0 +1,173 @@
...
+* Comments: No
+* Module name: lua2
+* Launch name: ``lua2``
+
+This is a rewrite of existing Lua backend.
+This backend is stub between your Lua script and PowerDNS authoritative server.
+The backend uses AuthLua4 base class, and you can use same functions and types as in any other Lua script.
+
+.. warning::
+  Some of the function calls and configuration settings have been changed, please review this document carefully.
```
Habbie: changed compared to what?

docs/backends/lua2.rst
```diff
@@ -0,0 +1,173 @@
...
+
+.. warning::
+  All places which use DNS names now use DNSName class which cannot be compared directly to a string.
+  To compare them against a string use either ``tostring(dnsname)`` or ``newDN(string)``.
+
+API description (v1)
```
Habbie: maybe call this v2?

docs/backends/lua2.rst
```diff
@@ -0,0 +1,173 @@
...
+``bool dns_dnssec``
+~~~~~~~~~~~~~~~~~~~
+If your script supports DNSSEC, set this to true.
+
+``dns_lookup(qtype, qname, domain_id)``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Perform lookup of given resource record name and type.
+
+INPUT:
+ - string qtype - Type of queried resource record
```
Habbie: Why not an integer or a qtype type?

docs/backends/lua2.rst
```diff
@@ -0,0 +1,173 @@
...
+ - DNSName qname - Name of queried resource record
+ - int domain_id - ID of associated domain
+
+OUTPUT:
+ Expects a table which has rows of following tables:
+ - DNSName name - resource record name (can also be string)
+ - string type - type of resource record
+ - string content - resource record content
+ - int ttl - time to live for this resource record (default: configured value)
```
Habbie: How would one trigger the defaults?

docs/backends/lua2.rst
```diff
@@ -0,0 +1,173 @@
...
+ - int domain_id - ID of associated domain (default: -1)
+ - bool auth - Whether data is authoritative or not (default: true)
+ - int last_modified - UNIX timestamp of last modification
```
Habbie: What's this for, if not for autoserial?

docs/backends/lua2.rst
```diff
@@ -0,0 +1,173 @@
...
+ - int scope_mask - How many bytes of source IP netmask was used for this result
```
Habbie: What use is this if `lookup` does not get passed the source IP?

docs/backends/lua2.rst
```diff
@@ -0,0 +1,173 @@
...
+ 
+``dns_get_domaininfo(domain)``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Get domain information for given domain.
+
+INPUT:
+ - DNSName domain - Domain to get info for
+
+OUTPUT:
+ Return false if not supported or found, otherwise expects a table of:
```
Habbie: This comment also applies to other functions: should I return a table with these things in this order, or do I return a table with keys `account`, `kind`, etc.?

pdns/pdnsutil.cc
```diff
@@ -80,6 +80,7 @@ void loadMainConfig(const std::string& configdir)
   string configname=::arg()["config-dir"]+"/"+s_programname+".conf";
   cleanSlashes(configname);
 
+  ::arg().set("resolver","Use this resolver for ALIAS and the internal stub resolver")="no";
```
Habbie: why add this?

docs/backends/lua2.rst
```diff
@@ -0,0 +1,173 @@
...
+``dns_get_before_and_after_names_absolute(id, qname)``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Calculate NSEC before/after value for the given qname for domain with id.
+
+INPUT:
+ - int id - Associated domain id
+ - DNSName qname - DNS name to calculate
+
+OUTPUT:
+ Table of three DNSNames, in order of Unhashed, before and after.
```
Habbie: Regarding some previous comments, I see from the testing script that some functions indeed expect keyed tables. This one does not - suggest saying 'Array of three DNSNames' for clarity.

pdns/pdnsutil.cc
cmouse: this one is required by lua-auth4.cc

docs/backends/lua2.rst
cmouse: Compared to lua backend. Need to clarify this.

docs/backends/lua2.rst
cmouse: Good question, not sure. I'm happy with any way you decide. =)

docs/backends/lua2.rst
cmouse: I guess I could pass the DNSPacket here... or maybe a special table with values of interest?

docs/backends/lua2.rst
Habbie: Let's go with v2 to avoid any confusion.

docs/backends/lua2.rst
Habbie: Table makes more sense to me.

docs/backends/lua2.rst
cmouse: i included all fields on the structure.

docs/backends/lua2.rst
cmouse: Updated.

docs/backends/lua2.rst
cmouse: Changed. Also added check that if version 1 is used, it instructs to use luabackend instead.

docs/backends/lua2.rst
cmouse: Uses QType now.

docs/backends/lua2.rst
cmouse: clarified documentation

docs/backends/lua2.rst
cmouse: special table added.

pdns/pdnsutil.cc
cmouse: this is ok..?

Habbie: yes

PR #6171, opened 2018-01-11
LUA Records (yes we know it is Lua)

ahupowerdns:
## Tasks

Not all of these tasks are requirements for merging this PR, but all of them need to happen at some point. Your help is welcome!

- [x] go through all failed builds spawned from https://builder.powerdns.com/#/builders/8/builds/2167, and identify why; then comment below. PRs extra welcome of course!
- [x] figure out how to create our own small, fake geoip database so we can write tests that don't rely on possibly changing data from Maxmind
- [x] add a license to minicurl.[cc|hh]
- [x] add `m4/pdns_check_libcurl` and replace `-lcurl` in Makefile.am with the output of it; drop curl support if CURLOPT_RESOLVE (which is not a #define!) is not available. It's likely that https://github.com/curl/curl/blob/master/docs/libcurl/libcurl.m4 can do this all.
- [x] add tests
- [x] make Lua non-optional for building the auth; see the recursor configure.ac for inspiration. Ideally this also involves removing all the related #if(def)s.
- [ ] expand the docs with more examples
- [ ] document geoiplookup (see comment below for an example)
- [ ] expose `enum GeoIPQueryAttribute` from geoipinterface.hh to Lua (this is the second argument to geoiplookup)
- [x] commit 'REVERT ME: allow way longer execution 218a055' was done because just importing Redis broke that limit so the number might be too low. It should probably be configurable.
- [x] `throw std::runtime_error("Script took too long");` does not make it to the logs

### Short description
With this PR, the Authoritative Server gains Lua powered LUA records that deliver high availability, load balancing, geographical load balancing and various other cool things.
### Checklist

I have:
- [ ] documented the code


pdns/lua-record.cc
```diff
@@ -386,6 +386,25 @@ static std::vector<DNSZoneRecord> lookup(const DNSName& name, uint16_t qtype, in
...
+  }
+
+  if(selector=="random")
+    return pickrandom(candidates);
+  else if(selector=="closest")
+    return closest(bestwho, candidates);
+  else if(selector=="hashed")
+    return hashed(bestwho, candidates);
+
+  L<<Logger::Warning<<"LUA Record ifportup called with unknown selector '"<<selector<<"'"<<endl;
```
mnordhoff: The warning message still assumes it's being called by ifportup.

pdns/minicurl.hh
```diff
@@ -0,0 +1,20 @@
+#pragma once
```
pieterlexis: minicurl.{cc,hh} has no license.

pdns/Makefile.am
```diff
@@ -232,7 +234,7 @@ pdns_server_LDADD = \
 	$(YAHTTP_LIBS) \
 	$(JSON11_LIBS) \
 	$(LIBCRYPTO_LIBS) \
-	$(SYSTEMD_LIBS)
+	$(SYSTEMD_LIBS) -lcurl
```
pieterlexis: If possible, please use `$(CURL_LIBS)`.

pieterlexis: In order to do so, add an `m4/pdns_check_libcurl`.

pfak:
Hello,

The documentation is not clear what behaviour is expected if the LUA record is not "hot" and a query is done against the record. I also don't expect the behaviour to be as below.

In the example section (https://powerdns.org/auth-4.1.1-docs/lua-records.html#examples), it says that a record will be randomly returned if a query has not been previously run (or TTL expired.) However, the behaviour I'm seeing is all records are returned.

Record:

{code}
dns.pfak.org. IN LUA A  "ifurlup('https://dns.pfak.org/pdnsmanager/', {{'198.50.219.2',  '51.254.142.100', '198.50.219.55'}, {'198.50.219.2'}},  {stringmatch='PDNS Manager', selector='hashed'})"
{code}

Results:

{code}
[peter@crash ~]$ host dns.pfak.org ns3
Using domain server:
Name: ns3
Address: 198.50.219.55#53
Aliases:

dns.pfak.org has address 51.254.142.100
dns.pfak.org has address 198.50.219.55
dns.pfak.org has address 198.50.219.2

[peter@crash ~]$ host dns.pfak.org ns3
Using domain server:
Name: ns3
Address: 198.50.219.55#53
Aliases:

dns.pfak.org has address 198.50.219.2
{code}

Expected behaviour:

Return one record if lookup has not been done at all.

Nice-to-have behaviour:

Allow a 'default' record to be specified and returned if the record is not hot in the cache
OR
Have all resource checks be run on startup and periodically even if the record isn't hot so that no default or random record is returned

Thank you!

Habbie:
> However, the behaviour I'm seeing is all records are returned.

I am also seeing this on the first query.

Habbie:
The documentation says " If all IP addresses are down, all candidates are returned." but as @pfak points out, 'down' and 'not checked yet' is not the same. It should at least be -possible- to have a single fallback IP.

Habbie:
Log message: `Lua record reported: Unable to convert parameter from nil to NSt3__112basic_stringIcNS_11char_traitsIcEENS_9allocatorIcEEEE`. dig result: NOERROR+SOA. Is this how we want to handle failures?

Habbie:
This PR does not build without Lua.

zeha:
Can we make Lua a build requirement? (possibly for both auth + rec)

Habbie:
> Can we make Lua a build requirement? (possibly for both auth + rec)

It already is for rec and dnsdist. I am fine with extending to auth for the sake of simplicity.

Habbie:
> Can we make Lua a build requirement? (possibly for both auth + rec)

Yes, we are going to do that.

Habbie:
geoiplookup example:

```
geo.example.net 3600 IN LUA TXT ";local t={} for i,v in ipairs{'asn', 'city', 'continent', 'country', 'country2', 'name', 'region', 'latlon'} do table.insert(t, string.format('%s=%s', v, geoiplookup(bestwho:toString(), i-1))) end return table.concat(t, ' ')"
```

```
$ dig txt geo.example.net @127.0.0.1 +subnet=194.109.0.0 +short
"asn=3265 city=unknown continent=eu country=nl country2=nl name=xs4all internet bv region=unknown latlon=52.382400 4.899500"
```

mnordhoff:
Don't know if this was expected, but `ifportup()` stopped working between the Ubuntu Xenial packages 0.0.luarec.2167g201b6d6-1pdns.xenial and 0.0.luarec.2168g04983ff-1pdns.xenial.

`pdns_server[20610]: Lua record reported: [string "return ifportup(443, {'45.79.1.70', '159.203...."]:1: attempt to call global 'ifportup' (a nil value)`

I guess the build environment didn't have `CURLOPT_RESOLVE` defined?

Would it be practical for `ifportup()` to work whether or not the curl stuff is available?

Habbie:
> Don't know if this was expected, but ifportup() stopped working between the Ubuntu Xenial packages 0.0.luarec.2167g201b6d6-1pdns.xenial and 0.0.luarec.2168g04983ff-1pdns.xenial.

This is not intended but not a surprise, I did not test that commit outside el6. I'll fix it soon.

> Would it be practical for ifportup() to work whether or not the curl stuff is available?

The curl stuff should be available everywhere except EL6 (as soon as I unbreak the ifdefs), so I think we're fine.

Habbie:
Reverted that commit - now we're broken on el6 again but we work everywhere else.

mnordhoff:
Maybe feature request:

`ifurlup` lets you specify multiple groups of IPs to be used at different priorities. `ifportup` only supports one group of IPs.

If `ifportup` supported groups too, I'm 66% sure I'd use it. So, if no one else is interested in it, there's a 34% chance adding it would be a waste of time. :sweat:

pdns/Makefile.am
```diff
@@ -236,6 +236,11 @@ pdns_server_LDADD = \
 	$(LIBCRYPTO_LIBS) \
 	$(SYSTEMD_LIBS)
 
+if HAVE_LUA_RECORDS
+pdns_server_SOURCES += lua-record.cc minicurl.cc minicurl.hh
```
pieterlexis: Please add these files to `EXTRA_DIST` as well

chbruyand: Done !

docs/lua-records.rst
```diff
@@ -0,0 +1,366 @@
...
+
+Functions available
+-------------------
+
+Record creation functions
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. function:: ifportup(portnum, addresses[, options])
+
+  Simplistic test to see if an IP address listens on a certain port. Note that
```
maikzumstrull: Should probably explicitly mention that this will test TCP. Could mention that it will attempt a connection and consider it "up" if the connection establishes - no data will be sent or read from the connection.

PR #6172, opened 2018-01-12
IXFR: correct behavior of dealing with DNS Name with multiple records; speed up IXFR transaction

cassvin:
##### commit 397ed71
Makes IXFR client actively close connection to IXFR server when IXFR is considered to be done. Right now the client would keep reading data from socket even server has sent out all the IXFR responses, until server close the TCP connection. The whole transaction can last for 30s or longer, depends on IXFR server.

##### commit 668624c
Fix bug.

Say a.test.com has two A records.
```
a.test.com.  IN  A  10.0.0.1
a.test.com.  IN  A  10.0.0.2
```
IXFR client receives an add change that a.test.com has another A record 10.0.0.3. The update will remove all A records of a.test.com and only keep the newly added one. The final result will be:
```
a.test.com.  IN  A  10.0.0.3
```

### Checklist
I have:
- [ ] documented the code

### DIFF
--- a/pdns/ixfr.cc
+++ b/pdns/ixfr.cc
@@ -171,7 +171,8 @@ vector<pair<vector<DNSRecord>, vector<DNSRecord> > > getIXFRDeltas(const ComboAd
   int8_t ixfrInProgress = -2;
 
   for(;;) {
-    if (!ixfrInProgress)
+    // IXFR end
+    if (ixfrInProgress >= 0)
       break;
 
     if(s.read((char*)&len, sizeof(len)) != sizeof(len))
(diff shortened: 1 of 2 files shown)


pdns/ixfr.cc
```diff
@@ -213,6 +217,12 @@ vector<pair<vector<DNSRecord>, vector<DNSRecord> > > getIXFRDeltas(const ComboAd
           return ret;
         }
         masterSOA = sr;
+      } else {
+        // we hit the last SOA record
+        // ixfr is considered to be done if we hit the last SOA record twice
+        if (r.first.d_type == QType::SOA && masterSOA->d_st.serial == getRR<SOARecordContent>(r.first)->d_st.serial) {
```
rgacogne: Please check that `getRR<SOARecordContent>(r.first)` is not `nullptr` before dereferencing it. It might happen is the class is not `IN` for example.

pdns/ixfr.cc
```diff
@@ -168,8 +168,12 @@ vector<pair<vector<DNSRecord>, vector<DNSRecord> > > getIXFRDeltas(const ComboAd
   std::shared_ptr<SOARecordContent> masterSOA = nullptr;
   vector<DNSRecord> records;
   size_t receivedBytes = 0;
+  int8_t ixfrInProgress = -2;
 
   for(;;) {
+    if (!ixfrInProgress)
```
rgacogne: This test would fail if we somehow received a transaction containing the last SOA record _thrice_. I agree this shouldn't happen, but how about testing `if (ixfrInProgress >= 0)` instead?

pdns/ixfr.cc
cassvin: Thanks rgacogne for reviewing. Good sense of that. I've updated the code accordingly.

pdns/ixfr.cc
cassvin: Yeah agree with that. It's more protective.

rgacogne:
For the record, this PR fixes an actual issue in the auth (removing all existing records for the name) and improves IXFR in the auth and the rec.

PR #6177, opened 2018-01-12
dnsdist: Fix building without libssl

rgacogne:
### Short description

* Only fail if DNS over TLS is enabled and we don't have either `GnuTLS` or `OpenSSL` libssl
* Switch `GnuTLS` to `auto`, we will detect it but should only link it if DNS over TLS is enabled

### Checklist

I have:
- [ ] documented the code


rgacogne:
> Tiny (un)related nit, configure help mentions s2n

Nice catch, fixed!

PR #6170, opened 2018-01-11
dnsdist: Add simple dnstap logging

zeha:
### Short description
Adds a simple version of http://dnstap.info logging to dnsdist. Can use FrameStreamLogger or RemoteLogger with DnstapLogAction/DnstapLogResponseAction. Can use libfstrm TCP code if present.

Based off @jvalentini's PR.

Moves queryTime from DNSResponse to DNSQuestion.

### Checklist

I have:
- [X] read the [CONTRIBUTING.md](https://github.com/PowerDNS/pdns/blob/master/CONTRIBUTING.md) document
- [X] compiled and tested this code
- [ ] included documentation (including possible behaviour changes)
- [ ] documented the code
- [X] added or modified regression test(s)
- [ ] added or modified unit test(s)

PR #6185, opened 2018-01-15
dnsdist: run self-answered responses through a ruleset

### LINKED ISSUE #6182: dnsdist: apply rules to locally generated responses
Responses generated by (query) rules are immediately sent out after the first terminating action.

For purposes of logging these generated responses, it would be great to reuse the existing Response Rules/Actions. These responses may actually be more interesting to log for failure analysis than "normal" responses.

zeha:
### Short description
Add a new ruleset for locally generated responses ("self-answered queries"). Previously those were impossible to run ResponseActions on, possibly for logging.
ServFail-on-no-policy responses are handled, too.

PR #6170 is a bit useless without this.

Docs are probably not the nicest.

Reviewers: I wasn't too sure which sizes to pass to processQuery at some points.

Fixes #6182.

### Checklist
I have:
- [ ] documented the code


rgacogne:
Please note that this PR will conflict heavily with #6190. It absolutely does not mean it shouldn't be merged, but we should think a bit about the order in which we merge them.

zeha:
I'm happy to rebase this, but sooner is better than later :)

zeha:
Rebased, also took the refactor to the next level.

PR #6190, opened 2018-01-16
dnsdist: Replace the Lua mutex with a rw lock to limit contention

rgacogne:
### Short description

Most policies can run concurrently, and only need to hold a lock because they access the pools. The Lua ones however need a write lock. If you really know what you are doing, you can request a read lock for custom Lua policies too.

### Checklist

I have:
- [x] read the [CONTRIBUTING.md](https://github.com/PowerDNS/pdns/blob/master/CONTRIBUTING.md) document
- [x] compiled and tested this code
- [x] included documentation (including possible behaviour changes)
- [ ] documented the code
- [ ] added or modified regression test(s)
- [ ] added or modified unit test(s)

PR #6191, opened 2018-01-16
dnsdist: Ring buffers sharding

rgacogne:
### Short description

I did test this code, but I had to rebase it to be able to open a pull request against the current master and the rebase was a bit painful, so please don't merge this before I can find the time to test it more.
Combining #6190 and this PR allowed me to reach a bit more than 1M qps with a single `dnsdist` process at almost 100% cache hit rate and in a very controlled lab environment, so YMMV.

### Checklist

I have:
- [x] documented the code


rgacogne:
Cleaned up the code, could still use a review!

rgacogne:
Rebased to fix a conflict. Added unit tests and some optional locking statistics.

rgacogne:
Rebased to fix a conflict with https://github.com/PowerDNS/pdns/pull/6327.

rgacogne:
Rebased so the rec tests pass, and be a little more permissive in the new unit tests because Travis is slow.

rgacogne:
Rebased to fix conflicts.

rgacogne:
Rebased, moved to C++11-style loops for walking the ring buffers as suggested by @chbruyand (thanks!).

rgacogne:
Rebased to fix conflicts.

rgacogne:
Tests failure is unrelated (CDB issue in the auth's tests).

pdns/dnsdist-rings.cc
```diff
@@ -19,35 +19,39 @@
...
+
+#include "dnsdist-rings.hh"
 
 size_t Rings::numDistinctRequestors()
 {
   std::set<ComboAddress, ComboAddress::addressOnlyLessThan> s;
-  ReadLock rl(&queryLock);
-  for(const auto& q : queryRing)
-    s.insert(q.requestor);
+  for (auto& shard : d_shards) {
```
chbruyand: Maybe add `const` to the `shard` reference in all of your loops against `d_shards` (not in `dnsdist-rings.hh`) ?

pdns/dnsdist-rings.hh
```diff
@@ -0,0 +1,209 @@
...
+    }
+  }
+  std::unordered_map<int, vector<boost::variant<string,double> > > getTopBandwidth(unsigned int numentries);
+  size_t numDistinctRequestors();
+  void setCapacity(size_t newCapacity, size_t numberOfShards)
+  {
+    if (numberOfShards < d_numberOfShards) {
+      throw std::runtime_error("Decreasing the number of shards in the query and response rings is not supported");
+    }
+
```
chbruyand: Maybe make sure the `newCapacity / numberOfShards` ratio is coherent ?

chbruyand: This may reduce the size of the indivual rings and  have some strange behaviour regarding `d_nbQueryEntries`

pdns/dnsdist-lua-inspection.cc
```diff
@@ -31,26 +32,27 @@ static std::unordered_map<unsigned int, vector<boost::variant<string,double>>> g
...
+        for(const auto& a : shard->respRing) {
+          if(!pred(a))
+            continue;
+          counts[a.name]++;
+          total++;
+        }
       }
+      else {
+        unsigned int lab = *labels;
+        for(auto a : shard->respRing) {
```
chbruyand: maybe `auto& a` ?

pdns/dnsdist-rings.cc
rgacogne: I would except I can't because I'm locking the mutex. I could make it mutable but that doesn't seem right here.

pdns/dnsdist-rings.hh
rgacogne: Indeed, that's why it's only allowed at configuration time and not at run-time. I could write a few words about that here as well though.

pdns/dnsdist-lua-inspection.cc
rgacogne: We can't because we modify the `name` member by calling `DNSName::trimToLabels()` on it below. We could consider only copying the `DNSName` instead of the whole entry, but I'm not sure it would make a noticeable difference. It might worth benchmarking it, or simply doing it.

rgacogne:
I pushed a few commits addressing @chbruyand's comments (thanks!).

PR #6209, opened 2018-01-19
Minimal fix to make primeHints threadsafe

zeha:
### Short description
wim on IRC discovered that there is a chance on startup that the root-server IPs will be incorrect.

### Checklist
I have:
- [ ] documented the code


zeha:
Using hint-file=<real-hint-file> avoids the broken code.

ahupowerdns:
note: this issue should not have survived the actual refreshing of the . records. So there is at least one more bug.

PR #6218, opened 2018-01-22
rec: Correctly handle ancestor delegation NSEC{,3} for children

aerique:
### Short description

Fixes the DNSSEC validation issue found in Knot Resolver, where a NSEC{3} ancestor delegation is wrongly use to prove the non-existence of a RR below the delegation.
We already had the correct check for the exact owner name, but not for RRs below the delegation.

### Checklist

I have:
- [ ] documented the code


aerique:
Merging even though the auth build failed after a discussion with @habbie on our internal communication.

PR #6230, opened 2018-01-25
remove all traces of selectmplexer, fix up pollmplexer

ahupowerdns:
select is dangerous and should never be used (as it trips over at a certain number of fds used). Our pollmplexer was severely bitrotted and may never have worked. It is revived in this PR.
No one "should" have been using the selectmplexer, but this makes sure.
### Checklist

I have:
- [ ] documented the code


zeha:
nproxy uses it, apparently.

PR #6242, opened 2018-02-02
Add support for encrypting IP addresses #gdpr

ahupowerdns:
### Short description
With this change, PowerDNS core gains ability to encrypt & decrypt IP addresses as described in https://medium.com/@bert.hubert/on-ip-address-encryption-security-analysis-with-respect-for-privacy-dabe1201b476
For IPv4 this uses ipcrypt, for IPv6 it uses a 128-bit AES ECB operation.
This PR also hooks up ipencrypt() and ipdecrypt() methods for dnsdist use, specifically to pseudonomyse logging.

### Checklist

I have:
- [x] documented the code


pdns/dnsdistdist/Makefile.am
```diff
@@ -152,14 +152,13 @@ if HAVE_RE2
 dnsdist_LDADD += $(RE2_LIBS)
 endif
 
-if HAVE_DNS_OVER_TLS
```
rgacogne: This would cause `dnsdist` to be linked against `libgnutls` and `$(LIBSSL_LIBS)` when they are available, regardless of whether DNS over TLS support is enabled. If I'm not mistaken you only need to link against `$(LIBCRYPTO_LIBS)` for this to work?

pdns/dnsdistdist/configure.ac
```diff
@@ -52,10 +52,11 @@ PDNS_CHECK_LUA_HPP
 DNSDIST_ENABLE_DNS_OVER_TLS
 DNSDIST_CHECK_GNUTLS
 DNSDIST_CHECK_LIBSSL
+PDNS_CHECK_LIBCRYPTO
 AS_IF([test "x$enable_dns_over_tls" != "xno"], [
   AS_IF([test "$HAVE_LIBSSL" = "1"], [
     # we need libcrypto if libssl is enabled
```
rgacogne: This comment is now useless

pdns/dnsdistdist/Makefile.am
ahupowerdns: yes, that is correct, but I did not find a way to make that happen correctly.

rgacogne:
Superseded by #7481.

PR #6251, opened 2018-02-09
Allow running `clang-analyze` from `make`

pieterlexis:
### Short description
This adds the `--enable-static-analysis` flag to `configure`. And when the compiler supports `--analyze` (which Clang supports), `make analyze` can be run. This will dump all Errors and Warnings per-source file to stdout and write some .plist files with information as well.

This allows us to analyze (new) code before coverity is run.

### Checklist

I have:
- [ ] documented the code


pdns/Makefile.am
```diff
@@ -1366,5 +1369,20 @@ ixfrdist.service: ixfrdist.service.in
...
 endif # TOOLS
-
 endif # HAVE_SYSTEMD
+
+analyze_srcs = $(filter %.cc, $(pdns_server_SOURCES))
+analyze_plists = $(analyze_srcs:%.cc=%.plist)
+
+if HAVE_STATIC_ANALYSIS
+AM_LDFLAGS += \
+	-export-symbols
```
rgacogne: I don't get the point of adding `-export-symbols` to `LD_FLAGS` here? It makes the linking fail on my host because libtool expects a path to a file to follow this parameter.

pieterlexis: On my system building the plists fails without this flag

cmouse: Do you by chance mean `-export-dynamic`?

pieterlexis:
Needs much more thinking

PR #6264, opened 2018-02-14
ixfrdist: Prevent a nullptr exception and guard reads/writes

pieterlexis:
### Short description
It was found that ixfrdist would crash with a segfault under certain circumstances. It seems that we would accidentally write a nullptr into `g_soas` on an update exception. This PR ensures we only write a SOA to g_soas when an update is fully done.

Also, protect every read access to `g_soas` with the mutex.

### Checklist

I have:
- [ ] documented the code

### DIFF
--- a/pdns/ixfrdist.cc
+++ b/pdns/ixfrdist.cc
@@ -195,9 +195,11 @@ void updateThread() {
     time_t now = time(nullptr);
     for (const auto &domain : g_domains) {
       shared_ptr<SOARecordContent> current_soa;
-      if (g_soas.find(domain) != g_soas.end()) {
+      {
         std::lock_guard<std::mutex> guard(g_soas_mutex);
-        current_soa = g_soas[domain];
+        if (g_soas.find(domain) != g_soas.end()) {
+          current_soa = g_soas[domain];
+        }
       }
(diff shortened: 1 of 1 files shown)


pdns/ixfrdist.cc
```diff
@@ -194,8 +194,13 @@ void updateThread() {
     }
     time_t now = time(nullptr);
     for (const auto &domain : g_domains) {
-      if ((g_soas.find(domain) != g_soas.end() && now - lastCheck[domain] < g_soas[domain]->d_st.refresh) || // Only check if we have waited `refresh` seconds
-          (g_soas.find(domain) == g_soas.end() && now - lastCheck[domain] < 30))  {                          // Or if we could not get an update at all still, every 30 seconds
+      shared_ptr<SOARecordContent> current_soa;
+      if (g_soas.find(domain) != g_soas.end()) {
+        std::lock_guard<std::mutex> guard(g_soas_mutex);
```
rgacogne: `g_soas.find()` is still accessing the `map` without holding the lock. In that case it might be fine because we are in the `updateThread` thread which is the only one updating the `map`, but it might be an issue in the future.

PR #6268, opened 2018-02-14
ixfrdist: Allow setting the inbound AXFR timeout

pieterlexis:
### Short description
When transferring huge zones, we might not make in the hardcoded 10 seconds. This PR makes `AXFRRetriever::getChunk()` and `timeoutReadn()` get an optional timeout parameter and hooks it up to ixfrdist.

We'll nee dto hook this up to the recursor and the auth in the future as well.

### Checklist

I have:
- [ ] documented the code

### DIFF
--- a/pdns/ixfrdist.cc
+++ b/pdns/ixfrdist.cc
@@ -240,7 +240,11 @@ void updateThread() {
         Resolver::res_t nop;
         vector<DNSRecord> chunk;
         records_t records;
-        while(axfr.getChunk(nop, &chunk, g_axfrTimeout)) {
+        time_t t_start = time(nullptr);
+        while(axfr.getChunk(nop, &chunk)) {
+          if (time(nullptr) - t_start > g_axfrTimeout) {
+            throw PDNSException("AXFR timeout exceeded");
+          }
           for(auto& dr : chunk) {
             if(dr.d_type == QType::TSIG)
               continue;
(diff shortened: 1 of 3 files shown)


pdns/ixfrdist.cc
```diff
@@ -707,6 +710,7 @@ int main(int argc, char** argv) {
       ("server-address", po::value<string>()->default_value("127.0.0.1:5300"), "server address")
       ("work-dir", po::value<string>()->default_value("."), "Directory for storing AXFR and IXFR data")
       ("keep", po::value<uint16_t>()->default_value(KEEP_DEFAULT), "Number of old zone versions to retain")
+      ("axfr-timeout", po::value<uint16_t>()->default_value(AXFRTIMEOUT_DEFAULT), "Timeout in seconds for an AXFR to complete")
```
rgacogne: It looks like it configures the timeout for getting a chunk of the `AXFR`, not for the whole `AXFR`, right?

pieterlexis: Correct, the wording is a bit off...

rgacogne: One option would be to keep track of the total elapsed time in `AXFRRetriever`, as we do for the number of bytes, but I'm not sure if that's you have in mind?

pieterlexis: That could also work, but that should be initialized on the first `getChunk` and be checked every consecutive call to `getChunk`. It also does not feel right.

PR #6278, opened 2018-02-18
pkcs11signers: Fix yubikey NEO to work

cmouse:
### Short description

Some devices have slots that do not start from 0, so we ask for the slot IDs from the device and iterate those instead of just guessing the slot ID.

### Checklist
I have:
- [x] documented the code

### DIFF
--- a/pdns/pkcs11signers.cc
+++ b/pdns/pkcs11signers.cc
@@ -279,6 +279,8 @@ class Pkcs11Token {
     std::string d_ecdsa_params;
 
     std::string d_label;
+    std::string d_pub_label;
+
     bool d_loaded;
     CK_RV d_err;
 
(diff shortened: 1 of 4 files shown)


cmouse:
Tested with a physical Yubikey NEO key.

pdns/pkcs11signers.cc
```diff
@@ -289,8 +295,39 @@ class Pkcs11Token {
       }
     }
 
+    unsigned int ecparam2bits(const std::string& obj) {
```
rgacogne: Perhaps that method could be `const`?

pdns/pkcs11signers.cc
```diff
@@ -289,8 +295,39 @@ class Pkcs11Token {
...
+      // if we can use some library to parse the EC parameters, better use it.
+      // otherwise fall back to using hardcoded primev256 and secp384r1
+#ifdef HAVE_LIBCRYPTO_ECDSA
+      EC_KEY *key = NULL;
+      BIGNUM *order;
+      unsigned int bits = 0;
+      unsigned char data[obj.size()];
+      const unsigned char *in = data;
+      obj.copy((char*)in, obj.size());
```
rgacogne: I'm very likely missing something but I don't see why we need to copy the content of `obj`?

cmouse:
Fixed nits on `ecparams2bits`

PR #6281, opened 2018-02-19
Recursor: properly truncate with EDNS, do what `server-id` promises and add NSID support

### LINKED ISSUE #6159: rec: actually support NSID
- Program: Recursor
 - Issue type: Bug report

### Short description
While our config description claims we do NSID, we do not.

### Environment
 - Operating system: any
 - Software version: 4.x, master
 - Software source: any

### Steps to reproduce
1. Notice `# server-id	Returned when queried for 'id.server' TXT or NSID, defaults to hostname` in config
2. start recursor
3. do +nsid query

### Expected behaviour
Response includes NSID.

### Actual behaviour
Response does not include NSID.

### Other information
NSID does work in auth.

pieterlexis:
### Short description
This PR does several things:

* Adds NSID support to the recursor
* Actually defaults to using the hostname for the `server-id` setting (as promised by the documentation)
* Allow setting `server-id` to `disabled` to prevent sending the server id to clients.

I have not yet figured out a nice way to test this though....

Closes #6159

### Checklist

I have:
- [ ] documented the code


zeha:
add NSID support to sdig, similar to ecs?

pieterlexis:
> add NSID support to sdig, similar to ecs?

Sure, but I'd need to run a bunch of recursors with different settings, that is the part I need to think about.

zeha:
> I'd need to run a bunch of recursors with different settings...

I see python test runner in your future.

pdns/pdns_recursor.cc
```diff
@@ -1188,9 +1188,16 @@ static void startDoResolve(void *p)
          OPT record.  This MUST also occur when a truncated response (using
          the DNS header's TC bit) is returned."
       */
-      if (addRecordToPacket(pw, makeOpt(edo.d_packetsize, 0, edo.d_Z), minTTL, maxanswersize)) {
```
rgacogne: `addRecordToPacket()` takes care of not exceeding the maximum answer size, which the replacement code does not.

pdns/pdns_recursor.cc
```diff
@@ -1188,9 +1188,16 @@ static void startDoResolve(void *p)
...
-        pw.commit();
+      DNSPacketWriter::optvect_t opts;
+      if(wantsNSID) {
+        const static string mode_server_id = ::arg()["server-id"];
+        if(mode_server_id != "disabled" && !mode_server_id.empty()) {
+          opts.push_back(make_pair(3, mode_server_id));
```
rgacogne: Using `EDNSOptionCode::NSID` instead of `3` would make it easier to read.

pdns/pdns_recursor.cc
```diff
@@ -1188,9 +1188,16 @@ static void startDoResolve(void *p)
...
+          variableAnswer = true; // Can't packetcache an answer with NSID
+        }
       }
+      pw.addOpt(maxanswersize, 0, DNSSECOK ? EDNSOpts::DNSSECOK : 0, opts);
```
rgacogne: I don't think we want `maxanswersize` here, since it's capped by the value advertised by the requestor. However the current code seems to just echo back the value advertised by the requestor (`edo.d_packetsize`) which is just wrong. I guess we should use `g_udpTruncationThreshold` instead?

pdns/pdns_recursor.cc
```diff
@@ -3148,6 +3155,12 @@ static int serviceMain(int argc, char*argv[])
   openssl_thread_setup();
   openssl_seed();
 
+  if(::arg()["server-id"].empty()) {
+    char tmp[128];
+    gethostname(tmp, sizeof(tmp)-1);
+    ::arg().set("server-id") = tmp;
```
rgacogne: if `gethostname()` fails the content of `tmp` is just plain garbage.

pdns/pdns_recursor.cc
```diff
@@ -3451,7 +3464,7 @@ int main(int argc, char **argv)
     ::arg().set("packetcache-ttl", "maximum number of seconds to keep a cached entry in packetcache")="3600";
     ::arg().set("max-packetcache-entries", "maximum number of entries to keep in the packetcache")="500000";
     ::arg().set("packetcache-servfail-ttl", "maximum number of seconds to keep a cached servfail entry in packetcache")="60";
-    ::arg().set("server-id", "Returned when queried for 'id.server' TXT or NSID, defaults to hostname")="";
+    ::arg().set("server-id", "Returned when queried for 'id.server' TXT or NSID, defaults to hostname, set custom or 'disabled'")="";
```
rgacogne: Perhaps we should default to "disabled" to prevent leaking the hostname? Especially since the behavior will change on an upgrade. Probably not a big deal though.

pdns/pdns_recursor.cc
pieterlexis: I was afraid of this indeed

pdns/pdns_recursor.cc
pieterlexis: This behaviour mimics the Auth. Users that change the `version-string` setting will also change this one.

Habbie:
The travis failures (at least in auth) appear to be real:

```
--- /home/travis/build/PowerDNS/pdns/regression-tests.nobackend/edns1/expected_result   2018-02-20 23:46:57.941441345 +0000
+++ /home/travis/build/PowerDNS/pdns/regression-tests.nobackend/edns1/real_result       2018-02-21 00:26:26.958597478 +0000
@@ -1 +1 @@
-EDNS response OK
+Invalid EDNS response, expected extended RCODE=BADVERS, no SOA, and OPT version 0
+exit 1
```

pieterlexis:
> The travis failures (at least in auth) appear to be real

Yes, I have problems running the regression tests locally as well.

pieterlexis:
I think the comments are fixed now. I also added tests. Please review.

regression-tests.recursor-dnssec/test_LargeAnswer.py
```diff
@@ -81,12 +99,20 @@ def datagramReceived(self, datagram, address):
 
         if request.question[0].name == dns.name.from_text('AAAA.large-answer.example.'):
             value = 'A'
-        else:
+            final_count = 5
```
rgacogne: This makes the large answer test useless, as it can no longer catch an of-by-one if the answer from an auth matches exactly `udp-truncation-threshold`. I'm guessing you altered that test because the recursor now always keep space to add the `EDNS OPT` record when the client supports `EDNS0`, so the existing answer was truncated. Please consider simply disabling `EDNS` in the queries sent in `testLargeAnswer()` instead, if I'm not mistaken it should be enough to restore the existing behavior.

pieterlexis: Disabling EDNS from the client would make the nameserver send a TC response, as the 1680 bytes won't fit into the reply packet. I'll make a noEDNS test that will max out the packet size.

rgacogne: But it did fit before without EDNS, what am I missing?

pieterlexis: the tests always had EDNS enabled, the client would send 4096 as the payload size, the recursor would send 1680 to the "auth", get a response of exactly that size and stuff in the packet to the client and forget about the EDNS OPT record in the reply packet as it would not fit.

pieterlexis:
Rebased, tested. Could use another review

rgacogne:
Exact same comments than my last review.

regression-tests.recursor-dnssec/test_LargeAnswer.py
pieterlexis: We discussed this on IRC, because we now reserve space for the OPT, truncation works a little different. Added a test without EDNS to also test 512 byte truncation.

rgacogne: But does the modified version test the issue the initial version was testing for, an off-by-one in the buffer processing response from authoritative servers whose size is exactly the same than our outgoing buffer size? The fact that `_udpTruncationThreshold` did not change while the size of the answer generated from `UDPLargeResponder()` did makes me think it doesn't.

pieterlexis: aha, *now* I understand what we were testing. This had nothing to do with the response to the client but with the response from the auth and the outgoing buffer size

pieterlexis:
Ready for another review round!

regression-tests.recursor-dnssec/test_EDNSBufferSize.py
```diff
@@ -0,0 +1,267 @@
...
+      client bufsize (stub => recursor)
+     bufsize to client (recursor => stub)
+
+    The subclasses will test the following scenarios:
+
+    test | udp-trunc | edns-outgoing | client bufsize | response size   | result to client | bufsize to client |
+    -----+-----------+---------------+----------------+-----------------+------------------+-------------------+
+     01  | 1680      | 1680          | 4096           | 1680 (inc EDNS) | 1680 (inc EDNS)  | 1680              |
+     02  | 1680      | 1680          | 1679           | 1680 (inc EDNS) | TC (+EDNS)       | 1680              |
+     03  | 1680      | 1681          | 4096           | 1680 (inc EDNS) | TC (+EDNS)       | 1680              |
```
rgacogne: Shouldn't the response size be 1681 for the third test?

regression-tests.recursor-dnssec/test_EDNSBufferSize.py
```diff
@@ -0,0 +1,267 @@
...
+     04  | 1680      | 1679          | 4096           | 1679 (inc EDNS) | 1679 (inc EDNS)  | 1680              |
+     05  | 1680      | 1680          | 1680           | 1680 (inc EDNS) | 1680 (inc EDNS)  | 1680              |
+     06  | 1680      | 1680          | 512 (No EDNS)  | 512 (+EDNS)     | 512 (no EDNS)    | (no EDNS)         |
+     07  | 1680      | 1680          | 512 (No EDNS)  | 513 (+EDNS)     | TC (no EDNS)     | (no EDNS)         |
+
```
rgacogne: Would you mind adding a test with a client advertising a buffer below 512? I would like to check we correctly use 512 in that case.

pdns/pdns_recursor.cc
```diff
@@ -785,22 +785,35 @@ static void startDoResolve(void *p)
...
     uint16_t maxanswersize = dc->d_tcp ? 65535 : min(static_cast<uint16_t>(512), g_udpTruncationThreshold);
     EDNSOpts edo;
+    bool variableAnswer = false;
     bool haveEDNS=false;
+    DNSPacketWriter::optvect_t returnedEdnsOptions; // Here we stuff all the options for the return packet
+    uint8_t ednsExtRCode = 0;
     if(getEDNSOpts(dc->d_mdp, &edo)) {
+      haveEDNS=true;
+      if (edo.d_version != 0) {
+        ednsExtRCode = ERCode::BADVERS;
```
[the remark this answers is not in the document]
pieterlexis: or immediately, I don't see a reason to parse the packet further, initialize the protobuf message or the dq object

pdns/pdns_recursor.cc
```diff
@@ -785,22 +785,35 @@ static void startDoResolve(void *p)
...
+      for (const auto& o : edo.d_options) {
+        if (o.first == EDNSOptionCode::ECS && g_useIncomingECS && !dc->d_ecsParsed) {
+          dc->d_ecsFound = getEDNSSubnetOptsFromString(o.second, &dc->d_ednssubnet);
+        } else if (o.first == EDNSOptionCode::NSID) {
+          const static string mode_server_id = ::arg()["server-id"];
+          if(mode_server_id != "disabled" && !mode_server_id.empty()) {
+            returnedEdnsOptions.push_back(make_pair(EDNSOptionCode::NSID, mode_server_id));
+            variableAnswer = true; // Can't packetcache an answer with NSID
+            // Option Code and Option Length are both 2
+            maxanswersize -= 2 + 2 + mode_server_id.size();
```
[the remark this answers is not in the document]
pieterlexis: hmm, RFC 5001 says:
>    Including the NSID option in a response is always optional, so this
   specification never requires name servers to truncate response
   messages.

We can check the size before we add it to returnedEDNSOptions (I prefer not to have this logic _after_ resolution)

regression-tests.recursor-dnssec/test_EDNSBufferSize.py
pieterlexis: it is in my hand-written table and in the test as well.. fixing the table

pdns/pdns_recursor.cc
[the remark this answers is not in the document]
rgacogne: We might still want to send a protobuf message for this answer, though?

pdns/pdns_recursor.cc
```diff
@@ -1284,9 +1301,8 @@ static void startDoResolve(void *p)
          OPT record.  This MUST also occur when a truncated response (using
          the DNS header's TC bit) is returned."
       */
-      if (addRecordToPacket(pw, makeOpt(edo.d_packetsize, 0, edo.d_Z), minTTL, dc->d_ttlCap, maxanswersize)) {
-        pw.commit();
-      }
+      pw.addOpt(g_udpTruncationThreshold, ednsExtRCode, DNSSECOK ? EDNSOpts::DNSSECOK : 0, returnedEdnsOptions);
```
Habbie: `handleNewUDPQuestion` accepts queries of up to 256 bytes, not `udp-truncation-threshold`

Habbie: I lied, it's 1500 (see #6627). 256 would be a fine limit, though!

pdns/pdns_recursor.cc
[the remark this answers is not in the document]
pieterlexis: Fixed with

```cpp
    if(ednsExtRCode != 0) {
      goto sendit;
    }

    if(dc->d_mdp.d_qtype==QType::ANY && !dc->d_tcp && g_anyToTcp) {
      pw.getHeader()->tc = 1;
      res = 0;
....
```

pdns/pdns_recursor.cc
pieterlexis: 256 is forbidden by RFC 6891 section 6.2.5:

>    Values of less than 512 bytes MUST be treated as equal to 512 bytes.

(ok, not forbidden technically, but ignored...). So I've set this to 512.

pieterlexis:
Rebased, I would appreciate a merge soon as rebasing is quite the thing every time

PR #6283, opened 2018-02-19
dnsdist: Report Lua(Response)Action failures

zeha:
### Short description
Previously, an error from inside Lua code in an LuaAction or LuaResponseAction would cause the query to not be answered at all, with no diagnostics.

- Creates a new statistic item: rule-servfail
- Logs Lua errors with warnlog
- Failed Lua code causes an immediate ServFail now

Example log:

```
> LuaAction failed inside lua, returning ServFail: [string "chunk"]:21: attempt to call a nil value (global 'rl')
```

### Checklist

I have:
- [ ] documented the code

### DIFF
--- a/pdns/dnsdist.cc
+++ b/pdns/dnsdist.cc
@@ -992,6 +992,12 @@ bool processQuery(LocalHolders& holders, DNSQuestion& dq, string& poolname, int*
         g_stats.ruleRefused++;
         return true;
         break;
+      case DNSAction::Action::ServFail:
+        dq.dh->rcode = RCode::ServFail;
+        dq.dh->qr=true;
+        g_stats.ruleServFail++;
+        return true;
+        break;
       case DNSAction::Action::Spoof:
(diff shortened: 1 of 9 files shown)


pdns/dnsdist.cc
```diff
@@ -1039,6 +1045,10 @@ bool processResponse(LocalStateHolder<vector<DNSDistResponseRuleAction> >& local
       case DNSResponseAction::Action::HeaderModify:
         return true;
         break;
+      case DNSResponseAction::Action::ServFail:
+        dr.dh->rcode = RCode::ServFail;
+        return true;
+        break;
```
rgacogne: I wonder if we should increment `g_stats.ruleServFail` here too?

zeha: We don't increment any of the ruleXXX counters for ResponseActions right now, and I think that would be confusing because the sums will not match up.

rgacogne: Alright, that makes sense :)

PR #6289, opened 2018-02-20
dnsdist: unify global stats accounting

zeha:
### Short description
Clean up some statistics accounting which was handled inconsistently before:

- Before, UDP selfAnswered was increased even if no response was sent, while this isn't done for TCP or UDP/TCP cacheHits. Now it is only increased if a response goes out.
- ~~Global latency stats were not updated for TCP at all~~ - to avoid changing current behaviour, this is dropped
- Treat selfAnswered, cacheHits, and noPolicy the same (=latency 0) for global latency stats, both UDP and TCP

Additional stats issues that this PR does not fix:
- UDP and TCP update aclDrops, nonCompliantQueries, queries at different times
- nonCompliantQueries is increased either before or after queries is increased

### Checklist
I have:
- [ ] documented the code

### DIFF
--- a/pdns/dnsdist-tcp.cc
+++ b/pdns/dnsdist-tcp.cc
@@ -378,9 +378,10 @@ void* tcpClientThread(int pipefd)
           }
 #endif
           handler.writeSizeAndMsg(query, dq.len, g_tcpSendTimeout);
-	  g_stats.selfAnswered++;
-	  continue;
-	}
+          g_stats.selfAnswered++;
+          doLatencyStats(0);  // we're not going to measure this
+          continue;
+        }
 
         std::shared_ptr<ServerPool> serverPool = getPool(*holders.pools, poolname);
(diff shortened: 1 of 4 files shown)


rgacogne:
Thanks! Code looks good, although I'm not sure we want to take the TCP response time into account in our latency calculations. It is my understanding that by design we only care about the UDP response time, especially because we use it for our policies, but I might be mistaken.

zeha:
I think latency is only used by the leastOutstanding policy, which looks at DownstreamState latencyUsec. This PR does not touch the DSS latencyUsec for TCP. (This could have been clearer...)

zeha:
OTOH: I could add extra timekeeping for TCP, just before writing the response out. Then we could use that and update DSS latencyUsec too. Y/N/Scary?

zeha:
I've dropped the TCP latency updating for now, can discuss this later.

PR #6290, opened 2018-02-21
rec: Implement settable AXFR timeout for RPZ

pieterlexis:
### Short description
This allows the operator to set the timeout of the AXFR transaction. This can be handy when the zone is huge or the link is slow.

### Checklist

I have:
- [ ] documented the code


pdns/rpzloader.cc
```diff
@@ -189,8 +189,10 @@ shared_ptr<SOARecordContent> loadRPZFromServer(const ComboAddress& master, const
   Resolver::res_t nop;
   vector<DNSRecord> chunk;
   time_t last=0;
+  time_t axfrStart = time(0);
```
rgacogne: s/`time(0)/`time(nullptr)` would be more consistent.

pdns/rpzloader.cc
```diff
@@ -205,6 +207,10 @@ shared_ptr<SOARecordContent> loadRPZFromServer(const ComboAddress& master, const
       RPZRecordToPolicy(dr, zone, true, defpol, maxTTL);
       nrecords++;
     } 
+    axfrNow = time(nullptr);
+    if (axfrNow - axfrStart > axfrTimeout) {
```
rgacogne: Perhaps we should check that `axfrNow` >= `axfrStart` in case the time goes backward because of NTP adjustment for example

PR #6291, opened 2018-02-21
ixfrdist: Read only the amount of bytes we need

pieterlexis:
### Short description
spotted by @rgacogne

We would read the size of the buffer on a possibly blocking call.

### Checklist

I have:
- [ ] documented the code

### DIFF
--- a/pdns/ixfrdist.cc
+++ b/pdns/ixfrdist.cc
@@ -23,6 +23,7 @@
 #include "config.h"
 #endif
 #include <boost/program_options.hpp>
+#include <arpa/inet.h>
 #include <sys/types.h>
 #include <grp.h>
 #include <pwd.h>
@@ -619,8 +620,7 @@ void handleTCPRequest(int fd, boost::any&) {
     return;
   }
 
-  char buf[4096];
-  // Discard the first 2 bytes (qlen)
+  char buf[65535];
   int res;
   res = recv(cfd, &buf, 2, 0);
   if (res != 2) {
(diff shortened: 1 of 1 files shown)


pdns/ixfrdist.cc
```diff
@@ -619,8 +620,7 @@ void handleTCPRequest(int fd, boost::any&) {
     return;
   }
 
-  char buf[4096];
-  // Discard the first 2 bytes (qlen)
+  char buf[65535];
```
rgacogne: That's a huge buffer on the stack, perhaps 4k was enough for an `XFR` query? BTW you are shadowing this variable line 710:
```
char buf[2];
```

pdns/ixfrdist.cc
```diff
@@ -619,8 +620,7 @@ void handleTCPRequest(int fd, boost::any&) {
...
   int res;
```
rgacogne: `recv()` returns a `ssize_t`, even if we should be fine given that the buffer is currently smaller than `2^16`

pdns/ixfrdist.cc
```diff
@@ -636,7 +636,8 @@ void handleTCPRequest(int fd, boost::any&) {
     }
   }
 
-  res = recv(cfd, &buf, sizeof(buf), 0);
+  size_t toRead = std::min(static_cast<size_t>(ntohs((buf[0]<<8) + buf[1])), sizeof(buf));
```
rgacogne: Just a nit but it might be easier to read the size into a `uint16_t` than doing bits shifting

PR #6296, opened 2018-02-22
Add a default include-dir for recursor as per auth

kaosdrachen:
### Short description

Modifies the debian build scripts so that the recursor.conf by default features include-dir=/etc/powerdns/recursor.d (and that said directory is created during install)

I added this because it's a QOL feature that as far as I can tell does not interfere with unmodified existing installs.

Documentation seems superfluous since this PR shortdesc is already longer than the entirety of the patch.

### Checklist
I have:
- [ ] documented the code


zeha:
https://salsa.debian.org/dns-team/pdns-recursor/commit/1d7428121a8593672e5abfc8778a6c3f501a8c6e

Issue #6241, opened 2018-01-31
Entries in ecsIndex cache are not updated after they expire once


liordot:
- Program: Recursor 
 - Issue type: Bug report 

### Short description

When using ecs, entries are only served from cache until they expire one time, after that they are not cached anymore.

### Environment

 - Operating system: Ubuntu 16.04
 - Software version: 4.1.0
 - Software source: Ubuntu Repository

### Steps to reproduce

1. Enable ecs
2. Make a request for a host
3. Wait for the response to expire

### Expected behaviour

Entry should be served from cache
### Actual behaviour

Entry is not cached.
### Other information

It looks like the code here:
https://github.com/PowerDNS/pdns/blob/master/pdns/recursor_cache.cc#L271
Only update the ecsIndex if the entry is not in d_cache.
Here the code only erase the ecsIndex but d_cache is still populated.
https://github.com/PowerDNS/pdns/blob/master/pdns/recursor_cache.cc#L121
The result is that the entry is not in the ecsIndex and the response is not cached.

The code below fixes the issue:

```
if (stored == d_cache.end()) {
    stored = d_cache.insert(CacheEntry(key, CacheEntry::records_t(), auth)).first;
    isNew = true;

  }
    if (ednsmask) {
      auto ecsIndexKey = boost::make_tuple(qname, qt.getCode());
      auto ecsIndex = d_ecsIndex.find(ecsIndexKey);
      if (ecsIndex == d_ecsIndex.end()) {
        ecsIndex = d_ecsIndex.insert(ECSIndexEntry(qname, qt.getCode())).first;
      }
      ecsIndex->addMask(*ednsmask);
    }

```

ahupowerdns:
thanks liordot! We're looking into it.

ahupowerdns:
@liordot - we confirm the issue & that your fix appears to be correct. How can we thank you in the release notes?

liordot:
You're welcome :)

Can you share an ETA on when the next version with this fix will be released?

rgacogne:
The fix proposed in e0dfa1c44bc8a356e9f8c0e35f382c2ca2fff239 looks good to me, thank you again for reporting this!

ahupowerdns:
I think this one can be close, @rgacogne ?

PR #6300, opened 2018-02-22
Fix ECS-based cache entry refresh code

ahupowerdns:
This would lead to us not refreshing ECS-varying answers in the cache, and therefore an even more depressed cache rate.

Thanks to @liordot who found the issue & provided a fix in #6241. This code has been tested by @paddg where it slightly reduced CPU load.

### Short description
The fix seems trivial, but a second look is required.

### Checklist

I have:
- [ ] documented the code

### DIFF
--- a/pdns/recursor_cache.cc
+++ b/pdns/recursor_cache.cc
@@ -93,7 +93,6 @@ MemRecursorCache::cache_t::const_iterator MemRecursorCache::getEntryUsingECSInde
         /* we have nothing more specific for you */
         break;
       }
-
       auto key = boost::make_tuple(qname, qtype, best);
       auto entry = d_cache.find(key);
       if (entry == d_cache.end()) {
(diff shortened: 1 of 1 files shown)


pdns/recursor_cache.cc
```diff
@@ -264,23 +263,23 @@ bool MemRecursorCache::attemptToRefreshNSTTL(const QType& qt, const vector<DNSRe
...
-    /* don't bother building an ecsIndex if we don't have any netmask-specific entries */
-    if (ednsmask && !ednsmask->empty()) {
-      auto ecsIndexKey = boost::make_tuple(qname, qt.getCode());
-      auto ecsIndex = d_ecsIndex.find(ecsIndexKey);
-      if (ecsIndex == d_ecsIndex.end()) {
-        ecsIndex = d_ecsIndex.insert(ECSIndexEntry(qname, qt.getCode())).first;
-      }
-      ecsIndex->addMask(*ednsmask);
+  /* don't bother building an ecsIndex if we don't have any netmask-specific entries */
+  if (ednsmask && !ednsmask->empty()) {
```
rgacogne: Since the only case where we might have an existing entry in `d_cache` for this exact (`qname`, `qtype`, `ednsmask`) tuple, but not one in `d_ecsIndex`, is that the entry has expired and was removed from `d_ecsIndex` but it's still waiting for garbage collection in `d_cache`, we could skip the lookup into `d_ecsIndex` and the call to `ecsIndex->addMask()` unless `isNew` or `stored->d_ttd <= now`. It would require a comment explaining why, though.

PR #6313, opened 2018-02-28
rec: Count a lookup into an internal auth zone as a cache miss

rgacogne:
### Short description

We only considered a query a cache miss if we had to do at least one outgoing query, but having to look up into one of the internally hosted zone (auth-zones) should count as a cache miss too.

### Checklist

I have:
- [ ] documented the code

### DIFF
--- a/pdns/syncres.hh
+++ b/pdns/syncres.hh
@@ -697,6 +697,7 @@ public:
 
   std::unordered_map<std::string,bool> d_discardedPolicies;
   DNSFilterEngine::Policy d_appliedPolicy;
+  unsigned int d_authzonequeries;
   unsigned int d_outqueries;
   unsigned int d_tcpoutqueries;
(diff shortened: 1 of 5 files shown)


pdns/syncres.hh
```diff
@@ -697,6 +697,7 @@ public:
 
   std::unordered_map<std::string,bool> d_discardedPolicies;
   DNSFilterEngine::Policy d_appliedPolicy;
+  unsigned int d_authzonequeries;
```
pieterlexis: This variable should be initialized [here](https://github.com/rgacogne/pdns/blob/f7b8cffae1ff94c2e78ac6f3b3ee494735a0c23c/pdns/syncres.cc#L111)

rgacogne:
Fixed the initialized member issue (thanks!) and added the corresponding metric.

PR #6317, opened 2018-03-01
dnsdist: Add an option to use several source ports toward a backend

rgacogne:
### Short description

This is very useful if the backend is distributing queries based only on (source IP, source port, destination IP, destination port), which is for example the case of PowerDNS Recursor with several
threads, `reuseport` set and `pdns-distribute-queries` not set.

### Checklist

I have:
- [ ] documented the code

### DIFF
--- a/pdns/dnsdist-lua.cc
+++ b/pdns/dnsdist-lua.cc
@@ -217,9 +217,13 @@ void setupLuaConfig(bool client)
 			  }
 			}
 
-			if(vars.count("sockets")) {
-			  numberOfSockets=std::stoi(boost::get<string>(vars["sockets"]));
-			}
+                        if (vars.count("sockets")) {
+                          numberOfSockets = std::stoul(boost::get<string>(vars["sockets"]));
+                          if (numberOfSockets == 0) {
+                            warnlog("Dismissing invalid number of sockets '%s', using 1 instead", boost::get<string>(vars["sockets"]));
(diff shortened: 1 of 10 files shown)


pdns/dnsdist-lua.cc
```diff
@@ -216,6 +217,10 @@ void setupLuaConfig(bool client)
 			  }
 			}
 
+			if(vars.count("sockets")) {
+			  numberOfSockets=std::stoi(boost::get<string>(vars["sockets"]));
```
zeha: maybe add a check > 0

rgacogne:
Added a check that the number of sockets is greater than 0, handle more gracefully a failed (re-)connection, and updated the file descriptors check to account for more than one socket per backend.

rgacogne:
Note that the current implementation will not perform well with a large number of sockets per backend since it currently uses `poll()`. It would be better to use our existing multiplexer classes, but it requires:
- removing / adding the socket descriptors to the multiplexer on reconnection ;
- refactoring the existing code to support the multiplexer callback interface, or adding a new function to the multiplexer interface that would return a list of 'ready' file descriptors.

rgacogne:
Pushed an additional commit using our existing multiplexer code. I went with the second option described above, adding a new function to the multiplexer interface that returns a list of 'ready' file descriptors. This code needs a serious review before merging!

PR #6323, opened 2018-03-05
dnsdist: create RemoteLoggers in client mode, but avoid connecting

zeha:
### Short description
Fixes a nullptr deref under --check-config, found by Quad9.

To make this work, RemoteLogger with asyncConnect now waits for the first message before  connecting. FrameStreamLogger created in client mode will not create its IO Thread, making logging impossible.

All tests will now run --check-config on their generated configuration files before running, so errors like this one should be caught by the tests.

### Checklist

I have:
- [ ] documented the code


zeha:
Not creating FrameStreamLogger delays checking the remote address (ip:port or unix socket name), so I thought that was worth it.

Also just realised my doc change is bullshit, will fix in a bit!

rgacogne:
You are right, this PR fixes a real issue and I don't have a better alternative to suggest at the moment so I think I will just merge it once it's green. We can improve that later if a better way comes to mind.

zeha:
Fixed the doc change: asyncConnect is only exposed in recursor, and in dnsdist it's always false (except now in client mode).

PR #6324, opened 2018-03-05
Fix unused variable compiler warning

pieterlexis:
### Short description
Turns out, the compiler was right...

### Checklist

I have:
- [ ] documented the code

### DIFF
--- a/pdns/misc.cc
+++ b/pdns/misc.cc
@@ -380,12 +380,10 @@ int waitForMultiData(const set<int>& fds, const int seconds, const int useconds,
     return ret;
 
   set<int> pollinFDs;
-  ctr = 0;
-  for (const auto& fd : realFDs) {
-    if (pfds[ctr].revents & POLLIN) {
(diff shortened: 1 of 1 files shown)


pdns/misc.cc
```diff
@@ -380,12 +380,10 @@ int waitForMultiData(const set<int>& fds, const int seconds, const int useconds,
     return ret;
 
   set<int> pollinFDs;
-  ctr = 0;
-  for (const auto& fd : realFDs) {
-    if (pfds[ctr].revents & POLLIN) {
-      pollinFDs.insert(pfds[ctr].fd);
+  for (unsigned int i = 0; i < realFDs.size(); i++) {
```
rgacogne: You could probably also do something like (untested, might rely on a gcc extension):
```
for (const auto& pfd : pfds) {
  if (pfd.revents & POLLIN) {
    pollinFDs.insert(pfd.fd);
  }
}
```

pieterlexis: that might work as well, let me check with clang

PR #6327, opened 2018-03-08
Add Metric "server-up" Count to Carbon Reporting

lowellmower:
### Short description

This PR adds a metric (`servers-up`) to be pushed to carbon for servers in a pool which are in a state of `UP`. There already existed a metric for `servers` but that did not offer insight into the health of a `ServerPool`, rather was just a count for the total list of downstreams irrespective of their `State`. There did not appear to be an open issue/request for this.

### Checklist

I have:
- [x] documented the code

Proposed Metric:
```
dnsdist.carbonname1.main.pools._default_.servers-up 2 1520457187
```

Regression Test Output:
```
$ DNSDISTBIN=../pdns/dnsdistdist/dnsdist ./runtests test_Carbon.py
Generating a 2048 bit RSA private key
.......................................................................+++
...............................................+++
writing new private key to 'ca.key'
-----
Generating a 2048 bit RSA private key
.........................................................................+++
.......................................................+++
writing new private key to 'server.key'
-----
Signature ok
subject=/CN=tls.tests.dnsdist.org/OU=PowerDNS.com BV/C=NL
Getting CA Private Key
..
Ran 2 tests in 12.345s

OK
```

### DIFF
--- a/pdns/dnsdist.hh
+++ b/pdns/dnsdist.hh
@@ -648,12 +648,12 @@ struct ServerPool
   std::shared_ptr<ServerPolicy> policy{nullptr};
 
   int countServersUp() {
-    int upFound = 0;
-    for(unsigned idx=0; idx<servers.size(); idx++) {
-      if ( std::get<1>(servers[idx])->isUp() ) {
+    size_t upFound = 0;
+    for (const auto& server : servers) {
+      if (std::get<1>(server)->isUp() ) {
         upFound++;
-      }
-    }
+      };
+    };
     return upFound;
   };
 };
--- a/pdns/dnsdist.hh
+++ b/pdns/dnsdist.hh
@@ -90,12 +90,12 @@ struct DNSResponse : DNSQuestion
     DNSQuestion(name, type, class_, lc, rem, header, bufferSize, responseLen, isTcp, queryTime_) { }
 };
 
-/* so what could you do: 
-   drop, 
-   fake up nxdomain, 
-   provide actual answer, 
-   allow & and stop processing, 
-   continue processing, 
+/* so what could you do:
+   drop,
+   fake up nxdomain,
+   provide actual answer,
+   allow & and stop processing,
+   continue processing,
    modify header:    (servfail|refused|notimp), set TC=1,
    send to pool */
 
@@ -172,7 +172,7 @@ struct DNSDistStats
   stat_t cacheHits{0};
   stat_t cacheMisses{0};
   stat_t latency0_1{0}, latency1_10{0}, latency10_50{0}, latency50_100{0}, latency100_1000{0}, latencySlow{0};
-  
+
   double latencyAvg100{0}, latencyAvg1000{0}, latencyAvg10000{0}, latencyAvg1000000{0};
   typedef std::function<uint64_t(const std::string&)> statfunction_t;
   typedef boost::variant<stat_t*, double*, statfunction_t> entry_t;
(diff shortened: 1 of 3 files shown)


pdns/dnsdist.hh
```diff
@@ -646,6 +646,16 @@ struct ServerPool
   NumberedVector<shared_ptr<DownstreamState>> servers;
   std::shared_ptr<DNSDistPacketCache> packetCache{nullptr};
   std::shared_ptr<ServerPolicy> policy{nullptr};
+
+  int countServersUp() {
+    int upFound = 0;
```
rgacogne: The type of `upFound` and the function return value should be unsigned, so `size_t` would be a good choice.

pdns/dnsdist.hh
```diff
@@ -646,6 +646,16 @@ struct ServerPool
...
+    for(unsigned idx=0; idx<servers.size(); idx++) {
```
rgacogne: Please use a c++11 loop whenever possible, as it's much more readable:
```
for (const auto& server : servers) {
  if (std::get<1>(server)->isUp() ) {
    upFound++;
  }
}

File: regression-tests.dnsdist/test_Carbon.py [hunk 3]
```diff
@@ -64,6 +77,30 @@ def startResponders(cls):
         cls._CarbonResponder2.setDaemon(True)
         cls._CarbonResponder2.start()
 
+    @classmethod
```
rgacogne: This is a quite complicated way to have some servers up, perhaps you could just force them up in the configuration, so you don't have to actually start responder threads? The more responder threads we start the more issue we potentially have later with another test reusing the same port too quickly. Something like:
```
server = newServer(...)
server:setDown()
server = newServer(...)
server:setUp()
server = newServer(...)
server:setUp
```

lowellmower:
@rgacogne - Sorry about the whitespace commit, happy to reset and commit that change separately if you like. Good call on the configuration for the regression test, it was feeling like a little much and turns out it was.

Thanks for the quick feedback, let me know if there is something else you'd like me to address.

regression-tests.dnsdist/test_Carbon.py
```diff
@@ -7,6 +7,10 @@
 
 class TestCarbon(DNSDistTest):
 
+    _serverCount = 3
+    _serverUpCount = 2
+    _toResponderQueue1 = Queue()
```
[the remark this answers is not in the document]
lowellmower: Good catch - also removed  `_serverCount` and `_serverUpCount` and just assert against `2` and `3` respectively.

lowellmower:
@rgacogne - Thanks again for such quick feedback. Let me know if you see anything else.

pdns/dnsdist.hh
```diff
@@ -646,6 +646,16 @@ struct ServerPool
   NumberedVector<shared_ptr<DownstreamState>> servers;
   std::shared_ptr<DNSDistPacketCache> packetCache{nullptr};
   std::shared_ptr<ServerPolicy> policy{nullptr};
+
+  const size_t countServersUp() {
```
rgacogne: This should be `size_t countServersUp() const {` instead, the current version is marking the return value `const`, which does not matter since it's neither a pointer nor a reference.

lowellmower: Ah, I misunderstood.

lowellmower:
@rgacogne - thank you for the approval, is there anything else you need from me in order for this to be merged with `master`?

rgacogne:
Ah sorry, this PR was not on my radar because the tests were failing but it's not your doing. Merging now, thanks again!

PR #6331, opened 2018-03-09
add TCP support for ALIAS

Habbie:
### Short description
The dnsproxy does not support TCP. For questions that come in over TCP that rely on ALIAS expansion, this PR short circuits the request via the stub resolver, instead of the proxy.

### Checklist

I have:
- [ ] documented the code

### DIFF
--- a/pdns/dnsproxy.cc
+++ b/pdns/dnsproxy.cc
@@ -32,6 +32,9 @@
 #include "logger.hh"
 #include "statbag.hh"
 #include "dns_random.hh"
+#include "stubresolver.hh"
+#include "tcpreceiver.hh"
+#include "arguments.hh"
 
 extern StatBag S;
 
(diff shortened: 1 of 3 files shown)


pdns/dnsproxy.cc
```diff
@@ -85,41 +88,39 @@ void DNSProxy::go()
...
-  return true;
-
-}
+extern TCPNameserver *TN;
 
 //! look up qname aname with r->qtype, plonk it in the answer section of 'r' with name target
 bool DNSProxy::completePacket(DNSPacket *r, const DNSName& target,const DNSName& aname)
 {
+  if(r->d_tcp) {
+    cerr<<"qtype="<<r->qtype.getCode()<<endl;
```
rgacogne: This should probably be removed before merging :)

pdns/dnsproxy.cc
```diff
@@ -85,41 +88,38 @@ void DNSProxy::go()
...
-  const string& buffer = p->getString();
-  
-  if(send(d_sock,buffer.c_str(), buffer.length() , 0)<0) { // zoom
-    L<<Logger::Error<<"Unable to send a packet to our recursing backend: "<<stringerror()<<endl;
-  }
-  (*d_resquestions)++;
-  return true;
-
-}
+extern TCPNameserver *TN;
```
rgacogne: Not blocking but why do we need to add that definition?

zeha: `tcpreceiver.hh` include also looks unused?

Habbie: Unsure why those were in there, possibly from some intermediate iteration. Removed now.

PR #6342, opened 2018-03-13
Auth: Reload /etc/resolv.conf when modified. Fix #6263

### LINKED ISSUE #6263: Stub resolver should reload /etc/resolv.conf
- Program: Authoritative 
 - Issue type: Bug report 

### Short description

When "`resolver`" isn't set, auth's stub resolver reads `/etc/resolv.conf` once at startup to determine what nameservers to use.

It uses them forever, even if `/etc/resolv.conf` is later changed.

### Environment

 - Operating system: Ubuntu 16.04 (x86-64)
 - Software version: 4.1.0-1pdns.xenial

### Steps to reproduce

1. Start PowerDNS.
2. Change `/etc/resolv.conf` to use different nameservers.
3. Wait for PowerDNS to do a secpoll query.

### Expected behaviour

PowerDNS querying one of the nameservers currently in `/etc/resolv.conf`.

### Actual behaviour

PowerDNS querying the nameserver that was in `/etc/resolv.conf` when it started.

### Other information

I use resolvconf. Last time I rebooted, it changed `/run/resolvconf/resolv.conf` from a nameserver I don't like to a nameserver I do like 1 second after PowerDNS was started. Sad face.

It's not a problem for me, it just makes me unhappy.

I don't often change `/etc/resolv.conf`.

chbruyand:
### Short description
Reload /etc/resolv.conf when file has been modified
Issue #6263

### Checklist

I have:
- [ ] documented the code

### DIFF
--- a/pdns/stubresolver.cc
+++ b/pdns/stubresolver.cc
@@ -114,7 +114,7 @@ void stubParseResolveConf()
 // s_resolversForStub contains the ComboAddresses that are used to resolve the
 int stubDoResolve(const DNSName& qname, uint16_t qtype, vector<DNSZoneRecord>& ret)
 {
-  // only check if resolvers come from he local resolv.conf in the first place
+  // only check if resolvers come from local resolv.conf in the first place
   if (s_localResolvConfMtime != 0) {
         parseLocalResolvConf();
   }
--- a/pdns/stubresolver.cc
+++ b/pdns/stubresolver.cc
@@ -1,3 +1,5 @@
+#include <sys/stat.h>
+
 #ifdef HAVE_CONFIG_H
 #include "config.h"
 #endif
@@ -14,9 +16,15 @@
 #include "statbag.hh"
 #include "stubresolver.hh"
 
+#define LOCAL_RESOLV_CONF_PATH "/etc/resolv.conf"
+
 // s_resolversForStub contains the ComboAddresses that are used by
 // stubDoResolve
 static vector<ComboAddress> s_resolversForStub;
+static pthread_mutex_t s_resolversForStubLock = PTHREAD_MUTEX_INITIALIZER;
+
+// /etc/resolv.conf last modification time
+static auto s_localResolvConfMtime = 0;
 
 /*
  * Returns false if no resolvers are configured, while emitting a warning about this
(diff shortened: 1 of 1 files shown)


pdns/stubresolver.cc
```diff
@@ -14,9 +16,15 @@
...
 
+#define LOCAL_RESOLV_CONF_PATH "/etc/resolv.conf"
+
 // s_resolversForStub contains the ComboAddresses that are used by
 // stubDoResolve
 static vector<ComboAddress> s_resolversForStub;
+static pthread_mutex_t s_resolversForStubLock = PTHREAD_MUTEX_INITIALIZER;
+
+// /etc/resolv.conf last modification time
+static auto s_localResolvConfMtime = 0;
```
ahupowerdns: In this case, you leave the compiler guessing with = 0. I suggest static time_t s_localResolvConfTime

pdns/stubresolver.cc
```diff
@@ -30,6 +38,47 @@ bool resolversDefined()
...
+  struct stat st;
+  string line;
+
+  if(!ifs)
+    return;
+  if (stat(LOCAL_RESOLV_CONF_PATH, &st) == -1)
+    return ;
+  s_localResolvConfMtime = st.st_mtime;
+
+  pthread_mutex_lock(&s_resolversForStubLock);
```
ahupowerdns: the lock.hh include file has 'resource acquisition is initialization' (RAII) wrappers for setting and releasing locks. These are much recommended as you can't forget to unlock - the destructor does that for you.

pdns/stubresolver.cc
```diff
@@ -80,6 +104,16 @@ void stubParseResolveConf()
 // s_resolversForStub contains the ComboAddresses that are used to resolve the
 int stubDoResolve(const DNSName& qname, uint16_t qtype, vector<DNSZoneRecord>& ret)
 {
+  // only check if resolvers come from he local resolv.conf in the first place
+  if (s_localResolvConfMtime != 0) {
+    struct stat st;
+
+    if (stat(LOCAL_RESOLV_CONF_PATH, &st) != -1) {
+      if (st.st_mtime > s_localResolvConfMtime) {
```
ahupowerdns: I'd say != instead of > - any change is a good reason to reload

mnordhoff: What about using `!=` instead of `>`? It's a bit weird, but someone could revert to an older backup copy of the file or something.

pdns/stubresolver.cc
```diff
@@ -80,6 +114,10 @@ void stubParseResolveConf()
 // s_resolversForStub contains the ComboAddresses that are used to resolve the
 int stubDoResolve(const DNSName& qname, uint16_t qtype, vector<DNSZoneRecord>& ret)
 {
+  // only check if resolvers come from he local resolv.conf in the first place
```
zeha: s/he/the/

Issue #6321, opened 2018-03-02
Avoid using UDP source port 11211


job:
- Program: Recursor
- Issue type: Feature request

### Short description

The recursor uses `1025 + dns_random(64510)` as ephemeral port - but UDP/11211 is now tainted due to the open memcached problem and in tons of people's rate-limiters and ACLs. Perhaps that source port should simply be avoided.

majek:
In the mean time
```
sysctl -w net.ipv4.ip_local_reserved_ports=11211
```

ahupowerdns:
So I think we should do this since this question keeps coming up. In a more generic sense, we should be able to set the range of ports we use, and then have a list of ports we don't use. It is overkill to make this too generic so we could set 1024-2048,32768- etc. So I suggest something like min-udp-source-port, max-udp-source-port and avoid-udp-source-port=80,11211

PR #6355, opened 2018-03-15
recursor: Add options to tune udp source ports

chbruyand:
Add min-udp-source-port, max-udp-source-port and avoid-udp-source-port variables to tune the range of ports we use

This should address #6321 

### Checklist

I have:
- [ ] documented the code

### DIFF
--- a/pdns/pdns_recursor.cc
+++ b/pdns/pdns_recursor.cc
@@ -27,6 +27,7 @@
 #include <sys/stat.h>
 #include <unistd.h>
 
+#include <boost/container/flat_set.hpp>
 #include "ws-recursor.hh"
 #include <pthread.h>
 #include "recpacketcache.hh"
@@ -60,7 +61,6 @@
 #include <boost/shared_array.hpp>
 #include <boost/function.hpp>
 #include <boost/algorithm/string.hpp>
-#include <boost/container/flat_set.hpp>
 #ifdef MALLOC_TRACE
 #include "malloctrace.hh"
 #endif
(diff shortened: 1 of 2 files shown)


pdns/pdns_recursor.cc
```diff
@@ -3516,6 +3548,10 @@ int main(int argc, char **argv)
     ::arg().set("xpf-allow-from","XPF information is only processed from these subnets")="";
     ::arg().set("xpf-rr-code","XPF option code to use")="0";
 
+    ::arg().set("min-udp-source-port", "Minimum UDP port to bind on")="1025";
+    ::arg().set("max-udp-source-port", "Maximum UDP port to bind on")="65535";
+    ::arg().set("avoid-udp-source-port", "List of comma separated UDP port number to avoid")="11211";
```
pieterlexis: Small nit:

Perhaps name the settings `udp-source-port-$SETTING` so they are sorted nicely in the generated config file.

rgacogne:
Probably needs some documentation :)

pdns/pdns_recursor.cc
zeha: this is kinda why i thought having a single udp-source-port setting would be a lot nicer and maybe easier to understand

krionbsd:
Fails on fbsd with:

In file included from pdns_recursor.cc:63:
In file included from /usr/local/include/boost/container/flat_set.hpp:29:
In file included from /usr/local/include/boost/container/detail/flat_tree.hpp:30:
In file included from /usr/local/include/boost/container/vector.hpp:58:
In file included from /usr/local/include/boost/move/algo/adaptive_merge.hpp:16:
/usr/local/include/boost/move/algo/detail/adaptive_sort_merge.hpp:813:14: error: illegal initializer
      (only variables can be initialized)
   size_type L = size_type(last - first);
             ^
./recursor_cache.hh:45:11: note: expanded from macro 'L'
#define L theL()
          ^

ahupowerdns:
the solution for @krionbsd's problem is to move the boost include all the way up so it happens before logger.hh gets included. It is a historical mistake that L...

chbruyand:
I moved the include on top and renamed the settings.

It should build. @krionbsd, is that easy for you to test ?

krionbsd:
I confirm, now it builds fine on fbsd head with clang 6.0.0, thank you, Charles-Henri!

PR #6377, opened 2018-03-23
auth: add support for MB and MG RR types

### LINKED ISSUE #5687: RR TYPE7 MB not supported
During regression testing I stumbled accross a problem. I have a test zone with pretty much different RR types. This zone is on a Bind Master and slaved by PowerDNS.

RFC3597 specifies that label compression can be used for well-known types = the types defined in RFC1035. RFC1035 defines the MB type. Hence, Bind compresses the RDATA section of the MB record.

PowerDNS does not actively support MB, it just handles the RR as "TYPE7". Hence, PowerDNS is not aware of the label compression and stores the RDATA as provided (compressed). When answering queries, the provided RDATA is the previously compressed data which is not valid in this response message (the compression data was only valid in the incoming AXFR). That may not be a big problem, as MB is rarely used/queried. But if PowerDNS provides outgoing AXFR to another Slave which understands MB, the transfer will fail as the MB RDATA (old compression data) is invalid.

I see 2 solutions: implement MB (should be easy, only a hostname as RDATA) or convince all other vendors to disable label compression for MB.

chbruyand:
### Short description
This PR adds support for MB and MR RR types as defined in RFC 1035.

This should fix #5687 

### Checklist

I have:
- [ ] documented the code


regression-tests.nobackend/tinydns-data-check/expected_result
```diff
@@ -1,4 +1,4 @@
-17a1219ca79020c1c67ad173a291d3c8  ../regression-tests/zones/example.com
+9f88091d3d1b793bf15a2cf5015bf92d  ../regression-tests/zones/example.com
```
Habbie: data.cdb should be regenerated and the checksum below updated. Once that is done, it is possible you won't need `skip.tinydns` at all. `description` in this dir has docs on the regenerating.

chbruyand: I'll have a look, but it looks like auth tests now succeed.
In `generate-data.sh`, what is the `tcpclient` tool ?

Habbie: it comes from the ucspi-tcp package from djb (http://cr.yp.to/). The tests indeed pass here when they sometimes should not.

PR #6378, opened 2018-03-23
dnsdist: Reduce the amount of times we copy our shared objects

rgacogne:
### Short description

Almost all of those copies should be quite fast and not noticeable, but let's not make a habit of doing a full copy.

### Checklist

I have:
- [ ] documented the code


rgacogne:
Rebased to fix a conflict.

PR #6381, opened 2018-03-23
dnsdist: Use a separate lock for accessing the pool's servers

rgacogne:
### Short description

We used to hold the `Lua` lock while applying the load-balancing policy to select a backend, which is only needed by `Lua` policies, not core ones. We do need a lock to make sure that the vector of servers is not altered under our feet, but a per-pool read-write lock is enough and reduces contention a lot, especially when the `maintenance` thread is doing some heavy-lifting.

### Checklist

I have:
- [ ] documented the code


rgacogne:
This PR supersedes #6190.

PR #6384, opened 2018-03-26
 dnsdist: Log when we add an eBPF dynamic block, like regular ones

### LINKED ISSUE #5845: bpf dynamic ip block filter to generate feedback or log lines
- Issue type: Feature request
### Short description
When dynamic IP Blocking in dnsdist using bpf, when IPs are blocked/unblocked by bpf please generate a feedback lines in dnsdist.log (syslog) like "Inserting dynamic block for 172.18.0.1 for 300 seconds: Exceeded-query-rate" in /var/log/dnsdist/dnsdist.log  (without bpf using dynamic ip blocking and any block of IPs is generating feedbacks

### Usecase
when source 1.1.1.1 is sending more volume of dns queries and dynamic IP Blocking bpf (exceededquery rate) if voilated, to generate a feedback lines as "Inserting dynamic block for 1.1.1.1 for 300 seconds: Exceeded-query-rate"

Feedbacks are important for us to get notified on the blocking.

rgacogne:
### Short description

Fixes #5845.

### Checklist

I have:
- [ ] documented the code

### DIFF
--- a/pdns/dnsdist-dynbpf.cc
+++ b/pdns/dnsdist-dynbpf.cc
@@ -23,8 +23,9 @@
 
 #ifdef HAVE_EBPF
 
-void DynBPFFilter::block(const ComboAddress& addr, const struct timespec& until)
+bool DynBPFFilter::block(const ComboAddress& addr, const struct timespec& until)
 {
+  bool inserted = false;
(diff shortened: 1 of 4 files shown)


zeha:
For the unbreak part: No clue how easy/hard this would be, but it sounds like something test-worthy?

PR #6394, opened 2018-03-26
dnsdist: Remove the --daemon option

### LINKED ISSUE #6329: dnsdist loses remote logger thread on daemonizing (-d)
- Program:  dnsdist
 - Issue type: Bug report

### Short description

When starting _dnsdist_ with `-d` (`--daemonize`), the program doesn't log to a remote logger. It's as though this action were never specified, though the connect is seen:

```
addAction(AllRule(), RemoteLogAction(newRemoteLogger("10.0.0.1:4242", 2, 2)))
```
In a private conversation with @rgacogne it turns out that:

>  we call daemonize() after creating the remote logger thread
> daemonize() calls fork() and the parent process exits
> so there is no remote logger worker thread anymore

For the record, Remi also says

> same pain for TeeAction and, obviously, dnstap

### Environment

 - Operating system:  SLES12SP3
 - Software version: 1.2.1
 - Software source: OBS package

### Expected behaviour

Logging should occur as without _daemonize_

rgacogne:
### Short description

We keep finding bugs with this option because we don't use it on a regular basis and most users run dnsdist under some kind of supervision anyway.

Closes #6329.

### Checklist

I have:
- [ ] documented the code

### DIFF
--- a/pdns/dnsdist.cc
+++ b/pdns/dnsdist.cc
@@ -723,41 +723,6 @@ shared_ptr<DownstreamState> roundrobin(const NumberedServerVector& servers, cons
   return (*res)[(counter++) % res->size()].second;
 }
 
-static void writepid(string pidfile) {
-  if (!pidfile.empty()) {
-    // Clean up possible stale file
-    unlink(pidfile.c_str());
-
-    // Write the pidfile
-    ofstream of(pidfile.c_str());
-    if (of) {
-      of << getpid();
-    } else {
-      errlog("Unable to write PID-file to '%s'.", pidfile);
-    }
-    of.close();
-  }
-}
-
-static void daemonize(void)
-{
-  if(fork())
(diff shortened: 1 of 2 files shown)


pdns/dnsdist.cc
```diff
@@ -2564,24 +2517,17 @@ try
...
-    vector<string> vec;
-    std::string acls;
-    g_ACL.getLocal()->toStringVector(&vec);
-    for(const auto& s : vec) {
-      if (!acls.empty())
-        acls += ", ";
-      acls += s;
-    }
-    infolog("ACL allowing queries from: %s", acls.c_str());
+  vinfolog("Running in the foreground");
```
pieterlexis: nit: not sure this message is still relevant

rgacogne: Oh right

PR #6395, opened 2018-03-26
dnsdist: Require the right gnutls version

pieterlexis:
### Short description
We use `gnutls_memset`, introduced in 3.4.0.

### Checklist

I have:
- [ ] documented the code


zeha:
Using memset_s instead might be a good idea. TBD

rgacogne:
I don't like the idea of raising the minimum `GnuTLS` to 3.4.0, since we only require `gnutls_memset()` when `libsodium` is not available. Jessie for example doesn't have 3.4.0 but has a libsodium package, so raising the minimum version would prevent the use of the `GnuTLS` TLS backend on Jessie for no good reason.
We could fallback to `memset_s()` indeed, although it might be a bit tricky because it's an extension defined for `C11` but not for `C++`, and if I'm not mistaken might not even be implemented by some `C11` implementations.

pieterlexis:
Superseded by the superior #6412

PR #6399, opened 2018-03-27
dnsdist: Restrict remote connection to the console via an ACL

### LINKED ISSUE #4654: dnsdist: Restrict console access with an ACL
- Program: dnsdist 
 - Issue type: Feature request 

### Short description

It would be nice to be able to pass an ACL (list of addresses as a string or a `NMG`) to `controlSocket()` to restrict access to the console by IP addresses.

### Usecase
Limiting access to the console socket to some hosts on the internal network.

rgacogne:
### Short description

This PR adds an ACL to restrict remote connection to the console, only allowing local connections by default. Please note that this is a breaking change from previous versions, where using any non-loopback address allowed remote connections, although the secret key was still required provided that encryption had been enabled.
Closes #4654.

### Checklist

I have:
- [ ] documented the code


pdns/dnsdist.cc
```diff
@@ -2264,6 +2239,10 @@ try
     g_ACL.setState(acl);
   }
 
+  auto consoleACL = g_consoleACL.getCopy();
+  consoleACL.addMask("127.0.0.1/8");
```
pieterlexis: For completeness-sake, maybe we should add `::1/128` here as well?

pdns/dnsdistdist/docs/reference/config.rst
```diff
@@ -205,6 +229,10 @@ Access Control Lists
 
   :param {str} netmasks: A table of CIDR netmask, e.g. ``{"192.0.2.0/24", "2001:DB8:14::/56"}``. Without a subnetmask, only the specific address is allowed.
 
+.. function:: showACL()
+
+  Print a list of all allowed netmasks.
```
zeha: this should probably say what this ACL is used for...

pdns/dnsdist.cc
zeha: yes please, various things default to ::1 now...

rgacogne: Makes sense!

pdns/dnsdistdist/docs/reference/config.rst
rgacogne: Right, I'll write a few words!

rgacogne:
Added `::1/128` and a few words about `addACL(), `setACL()` and `showACL()`.

pieterlexis:
A test would be nice. e.g. setting the acl to 127.0.0.1 and attempting to connect from 127.0.0.2?

rgacogne:
Regression tests added (the second version is even consistent :))!

PR #6407, opened 2018-03-27
dnsdist: Fix warnings reported by clang's analyzer and cppcheck

rgacogne:
### Short description

Mostly non-noticeable performance hints:
- pass shared pointers by reference whenever possible
- use pre-increment instead of post-increment for non-primitive types
- use initialization list instead of assigning values in the constructor's body

`LuaWrapper` moving the function variable before using it again might be a real issue though, and the locks' move constructor might have become one at some point.

### Checklist

I have:
- [ ] documented the code

### DIFF
--- a/pdns/dnsdist-lua.hh
+++ b/pdns/dnsdist-lua.hh
@@ -25,7 +25,7 @@ class LuaAction : public DNSAction
 {
 public:
   typedef std::function<std::tuple<int, boost::optional<string> >(DNSQuestion* dq)> func_t;
-  LuaAction(LuaAction::func_t& func) : d_func(func)
+  LuaAction(const LuaAction::func_t& func) : d_func(func)
   {}
   Action operator()(DNSQuestion* dq, string* ruleresult) const override;
   string toString() const override
@@ -40,7 +40,7 @@ class LuaResponseAction : public DNSResponseAction
 {
 public:
   typedef std::function<std::tuple<int, boost::optional<string> >(DNSResponse* dr)> func_t;
-  LuaResponseAction(LuaResponseAction::func_t& func) : d_func(func)
+  LuaResponseAction(const LuaResponseAction::func_t& func) : d_func(func)
   {}
   Action operator()(DNSResponse* dr, string* ruleresult) const override;
   string toString() const override
--- a/pdns/dnsdist-lua.hh
+++ b/pdns/dnsdist-lua.hh
@@ -25,7 +25,7 @@ class LuaAction : public DNSAction
 {
 public:
   typedef std::function<std::tuple<int, boost::optional<string> >(DNSQuestion* dq)> func_t;
-  LuaAction(LuaAction::func_t func) : d_func(func)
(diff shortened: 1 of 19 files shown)


pdns/dnsdist-lua.hh
```diff
@@ -25,7 +25,7 @@ class LuaAction : public DNSAction
 {
 public:
   typedef std::function<std::tuple<int, boost::optional<string> >(DNSQuestion* dq)> func_t;
-  LuaAction(LuaAction::func_t func) : d_func(func)
+  LuaAction(LuaAction::func_t& func) : d_func(func)
```
chbruyand: Maybe `const LuaAction::func_t& func` ?

pdns/dnsdist-lua.hh
```diff
@@ -40,7 +40,7 @@ class LuaResponseAction : public DNSResponseAction
 {
 public:
   typedef std::function<std::tuple<int, boost::optional<string> >(DNSResponse* dr)> func_t;
-  LuaResponseAction(LuaResponseAction::func_t func) : d_func(func)
+  LuaResponseAction(LuaResponseAction::func_t& func) : d_func(func)
```
chbruyand: Maybe `const LuaResponseAction::func_t& func` ?

pdns/dnsdist.hh
```diff
@@ -324,12 +337,10 @@ struct ClientState;
 struct IDState
 {
   IDState() : origFD(-1), sentTime(true), delayMsec(0), tempFailureTTL(boost::none) { origDest.sin4.sin_family = 0;}
-  IDState(const IDState& orig)
+  IDState(const IDState& orig): origRemote(orig.origRemote), origDest(orig.origDest)
```
chbruyand: Why not also moving other member variables in initialization list ? (`origFD`, `origID`, `delayMsec` and `tempFailureTTL`)

pdns/dnsparser.cc
```diff
@@ -199,9 +199,8 @@ DNSRecordContent::zmakermap_t& DNSRecordContent::getZmakermap()
   return zmakermap;
 }
 
-DNSRecord::DNSRecord(const DNSResourceRecord& rr)
+DNSRecord::DNSRecord(const DNSResourceRecord& rr): d_name(rr.qname)
 {
-  d_name = rr.qname;
   d_type = rr.qtype.getCode();
   d_ttl = rr.ttl;
```
chbruyand: Same remark as with `IDState()`. But I may be missing something. Sorry if that's the case.

rgacogne:
Pushed a fix to constify the function references passed to `LuaAction` and `LuaResponseAction`.

pdns/dnsdist.hh
rgacogne: I had to keep the initialization of some members in the constructor body (copy constructor is missing for `std::atomic` for example), so I only changed the one that mattered, ie non-Plain Old Data types.

pdns/dnsparser.cc
rgacogne: Same reason here.

PR #6412, opened 2018-03-28
dnsdist: Use safe_memset if we have neither sodium nor gnutls_memset

rgacogne:
### Short description

Use our own `safe_memset()` (thanks Dovecot!) if we have neither `libsodium`'s `sodium_m(un)lock()` nor `gnutls_memset()` available.

Supersedes #6395.

### Checklist

I have:
- [ ] documented the code


zeha:
This safe_memset is what gnutls_memset does, too. See http://lists.gnu.org/archive/html/emacs-devel/2017-07/msg00687.html why this is not necessarily good thing...

rgacogne:
I just pushed support to use `explicit_bzero()` or `explicit_memset()` before `gnutls_memset()` when available. We could also add support for `memset_s()` but the whole `__STDC_WANT_LIB_EXT1__` dance scares me and I don't have access to a platform supporting it to test. It's also starting to look like a joke, at some point if we don't trust what `GnuTLS` is providing to wipe the cryptographic materials from memory perhaps we should reconsider whether we trust `GnuTLS` in the first place.. I'm _not_ adding inline asm to that code.

zeha:
> We could also add support for memset_s() but the whole __STDC_WANT_LIB_EXT1__ dance scares me and I don't have access to a platform supporting it to test.

macOS has memset_s but not explicit_bzero, so I could test that. But not hugely important...

pdns/dnsdistdist/tcpiohandler.cc
```diff
@@ -595,7 +621,7 @@ class GnuTLSTicketsKey
 #ifdef HAVE_LIBSODIUM
       sodium_munlock(d_key.data, d_key.size);
```
pieterlexis: maybe `safe_memzero` should always be called and in that function `HAVE_LIBSODIUM` is checked?

pdns/dnsdistdist/m4/dnsdist_check_gnutls.m4
```diff
@@ -14,6 +14,14 @@ AC_DEFUN([DNSDIST_CHECK_GNUTLS], [
       PKG_CHECK_MODULES([GNUTLS], [gnutls >= 3.1.11], [
         [HAVE_GNUTLS=1]
         AC_DEFINE([HAVE_GNUTLS], [1], [Define to 1 if you have GnuTLS])
+        save_CFLAGS=$CFLAGS
+        save_LIBS=$LIBS
+        CFLAGS="$GNUTLS_CFLAGS $CFLAGS"
+        LIBS="$GNUTLS_LIBS $LIBS"
+        AC_CHECK_FUNCS([gnutls_memset explicit_bzero explicit_memset])
```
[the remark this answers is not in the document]
rgacogne: That's true, but we currently only use them for the GnuTLS code so I didn't bother.

pdns/dnsdistdist/tcpiohandler.cc
rgacogne: That doesn't feel right because `sodium_munlock()` does more than just zeroing the buffer.

rgacogne:
Renamed as `safe_memory_lock()` and `safe_memory_release()`, moved the `explicit_bzero()` and `explicit_memset()` checks to a separate m4 file.

PR #6429, opened 2018-03-31
dnsdist: reorder headers to fix OpenBSD build

Habbie:
### Short description
OpenBSD requires `sys/socket.h` to come before `net/if.h`.

### Checklist

I have:
- [ ] documented the code


rgacogne:
We should add some comment but otherwise it's bound to happen again.

Habbie:
> We should add some comment but otherwise it's bound to happen again.

Amended.

rgacogne:
thanks!

PR #6433, opened 2018-04-02
dnsdist: remove -d flag after #6394

Habbie:
### Short description
#6394 forgot to remove this, leading to a silent non-daemonized startup when passing `-d`.

Question: should the `switch` block have a `default:` that complains loudly?

Question: is `-s` missing intentionally even if we use it as the key to match `--supervised`?

### Checklist

I have:
- [ ] documented the code

### DIFF
--- a/pdns/dnsdist.cc
+++ b/pdns/dnsdist.cc
@@ -2082,9 +2082,9 @@ try
   string optstring;
   for(;;) {
 #ifdef HAVE_LIBSODIUM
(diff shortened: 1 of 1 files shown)


rgacogne:
Oh, right, I indeed forgot to remove it! Note that #6394 also removed the `pid` option so `p` should go as well. I don't remember if `-s` was left out intentionally but as it's not a very likely option to use manually on the command line, perhaps we just didn't bother adding the short version?

zeha:
Would agree on not having `-s`.

Habbie:
> so p should go as well

pushed

Habbie:
Replaced by #6435

PR #6434, opened 2018-04-02
Fix build with LibreSSL 2.7

Sp1l:
### Short description
LibreSSL 2.7 implements OpenSSL 1.1 API leading to conflicts in method names

### Checklist

I have:
- [ ] documented the code

Signed-off-by: Bernard Spil <brnrd@FreeBSD.org>

### DIFF
--- a/pdns/opensslsigners.cc
+++ b/pdns/opensslsigners.cc
@@ -35,7 +35,7 @@
 #include "dnsseckeeper.hh"
 
 #if (OPENSSL_VERSION_NUMBER < 0x1010000fL || \
-     (defined(LIBRESSL_VERSION_NUMBER) && LIBRESSL_VERSION_NUMBER < 0x2070000fL)
+     (defined(LIBRESSL_VERSION_NUMBER) && LIBRESSL_VERSION_NUMBER < 0x2070000fL) )
 /* OpenSSL < 1.1.0 needs support for threading/locking in the calling application. */
 static pthread_mutex_t *openssllocks;
 
--- a/pdns/opensslsigners.cc
+++ b/pdns/opensslsigners.cc
@@ -34,7 +34,8 @@
 #include "dnssecinfra.hh"
 #include "dnsseckeeper.hh"
 
-#if (OPENSSL_VERSION_NUMBER < 0x1010000fL || defined LIBRESSL_VERSION_NUMBER)
+#if (OPENSSL_VERSION_NUMBER < 0x1010000fL || \
+     (defined(LIBRESSL_VERSION_NUMBER) && LIBRESSL_VERSION_NUMBER < 0x2070000fL)
 /* OpenSSL < 1.1.0 needs support for threading/locking in the calling application. */
 static pthread_mutex_t *openssllocks;
 


Habbie:
> compiled and tested this code

Did you?

Sp1l:
Hi @Habbie 
Clearly not well enough... And not against the master branch. Meanwhile a version of the patch landed in FreeBSD ports. These were tested by me in FreeBSD ports.
https://bugs.freebsd.org/227183 https://bugs.freebsd.org/227184
Looks like a typo 😭 I redid the work in my private pdns fork. From the CI logs:
```
opensslsigners.cc:37:5: error: missing ')' in expression
 #if (OPENSSL_VERSION_NUMBER < 0x1010000fL || \
```
I'll cycle back to this and retest. Expect a follow up later in the week.

Sp1l:
> Note that I wouldn't be surprised to see the same fix needed in pdns/dnsdistdist/tcpiohandler.cc too if we want to enable DNS over TLS on FreeBSD.

Correct. Just checked and LibreSSL 2.7 has both `CRYPTO_set_id_callback` and `CRYPTO_set_locking_callback`.

The change is not for FreeBSD, it is currently only/mostly OpenBSD that is affected as they have LibreSSL 2.7 in the base OS. For FreeBSD users they must be using `security/libressl-devel` from ports.
Once I've landed sufficient patches for ports, I will switch `security/libressl` to 2.7 branch in FreeBSD.

Habbie:
dnsdist 1.3.0 compiles fine with DNS-over-TLS support on an OpenBSD 6.3 snapshot from this week, with LibreSSL 2.7.2, without anything like this patch.

Habbie:
@Sp1l ping

Sp1l:
Thanks for the poke @Habbie . Meanwhile I've updated the security/libressl port on FreeBSD to 2.7.2.
I'll check up on dnsdist and pdns builds.

RvdE:
When @Sp1l created this PR, the security/libressl port was still on the 2.6.x branch, which worked fine (only security/libressl-devel was on the 2.7.x branch). Since april 28th the ports has been updated to version 2.7.2 (and is currently on 2.7.3), which does require the patch.

Hopefully that clarifies things enough to get this into master, so I can remove the patch from my port on FreeBSD

tcely:
A similar patch is being used for AlpineLinux too. https://github.com/alpinelinux/aports/commit/c841cf3b79ba701cc99a6f3b3fed37278e00606c

aerique:
I had similar build issues on Void Linux which uses LibreSSL as well.

Cloned this branch and merged master into it. It compiled fine and I'm currently using it to work on another issue. I have not run into any problems so far.

rgacogne:
The fix looks good to me, compatibility with OpenSSL 1.1.0's API has been implemented in LibreSSL 2.7 so our own compatibility layer is colliding with it. I believe we should also backport this fix to 4.1.x and in dnsdist's `tcpiohandler.cc`.

rgacogne:
It turns out that dnsdist compiles correctly with 2.7.3 since it does not redefine the various accessors that collide in the auth and rec. I'm very concerned however about the locking code. OpenSSL >= 1.1.0 does not need it anymore, but looking at the code it's absolutely not obvious that `LibreSSL` >= 2.7.0 doesn't, and this PR removes it. If `LibreSSL` >= 2.7.0 does indeed still need it, merging this PR would cause race conditions, so in the end I don't think we should merge this as it is.
I would be OK to merge a PR removing the `RSA_` and `ECDSA_` helpers with `LibreSSL` >= 2.7.0, but unless someone can prove to me that the locking code is not needed I don't think we should remove it.

zeha:
`CRYPTO_w_lock` and similar appear to be quite in use in current libressl.

Sp1l:
Thanks for picking this up!

github-k8n:
Any news on this?
Is the locking code still needed with LibreSSL >= 2.8.0 (2.8.1)?
Or any other way to make libressl work with pdns?

Habbie:
I want to release auth 4.2.0-alpha soon; I'm going to need guidance on this from somebody who cares.

rgacogne:
To sum up a bit, the main remaining question is whether LibreSSL >= 2.7.0 still need the application to provide and setup locking callbacks the way OpenSSL before 1.1.0 did.

If it does, and so far I have no reason to believe it does not, this PR should be amended so that the `RSA_*` and `ECDSA_*` helper functions are not defined when building with LibreSSL >= 2.7.0, because these do conflict with the versions present in LibreSSL, but should keep the locking callbacks in place to be thread-safe.

Habbie:
Ping!

omoerbeek:
Since LibreSSL 2.9 the CRYPTO_LOCK functions are deprecated and stubbed.
For LibreSSL 2.7 and 2.8 there are still needed. But for the callback functions the story is different:
`CRYPTO_set_id_callback` and `CRYPTO_set_locking_callback` are still implemented and are required for current LibreSSL. AFAIKS master seems to do the right thing: not implement the helpers, but do setup the lock callbacks.

Using OpenBSD current, on master rec, auth and dnsdist run and build fine.

Habbie:
I agree with Otto, so I'm closing this.

PR #6440, opened 2018-04-03
WIP dnsdist: Allow removing incoming ECS data

zeha:
### Short description
If a client included EDNS Client Subnet data in a query, there was no way in dnsdist to get rid of that data. This turns ECSOverrideAction into a multi-purpose tool.

WIP: tests are probably broken, need to add tests for new format, and need tests for `remove`

### Checklist
I have:
- [ ] documented the code


zeha:
I'm now thinking:

- have one enum with keepIncoming, remove, replace (=override), addClientAddr (=old default)
- merge dq.useECS, dq.ecsOverride
- turn newServer useClientSubnet into the enum
- get rid of the global g_ECSOverride

thats not backward compatible though.

pieterlexis:
Let's turn this into a ticket so we can think about implementing this for the next version.

PR #6443, opened 2018-04-04
rec: Use a separate protobuf exporter thread per worker thread

rgacogne:
### Short description

Use a dedicated protobuf queue for each worker thread, instead of sharing a single one for all worker threads, leading to some contention under heavy load.
Also properly stops the existing protobuf exporter threads on a `Lua` configuration reload.

### Checklist

I have:
- [ ] documented the code


rgacogne:
Rebased to fix a conflict.

rgacogne:
Rebased to fix a (new) conflict.

PR #6447, opened 2018-04-04
[WIP] add memory limits to recursor caches

Habbie:
### Short description
This adds -bytes limits to the three recursor caches, in addition to the existing -entries limits. Cache pruning has been extended to make sure caches stay below both limits. 'bytes' reporting has been reinstated in `rec-control get_all` as that has become cheap.

TODO:
- [ ] make sure all xfr* methods, and all specific implementations of type conversion (like I did for NSEC/NSEC3) account for their usage
- [ ] measure against actual memory usage, then adjust the fudge factors (easily recognised by their use of `sizeof`); for this, it would be nice if we actually had a 'real memory usage' metric
- [ ] document
- [ ] make tests pass again
- [ ] expose `max-negcache-[bytes|entries]` metrics
- [ ] account for ECS
- [ ] add new metrics to SNMP?

### Checklist

I have:
- [ ] documented the code

### DIFF
--- a/pdns/nsecrecords.cc
+++ b/pdns/nsecrecords.cc
@@ -211,6 +211,8 @@ void NSEC3RecordContent::toPacket(DNSPacketWriter& pw)
 NSEC3RecordContent::DNSRecordContent* NSEC3RecordContent::make(const DNSRecord &dr, PacketReader& pr) 
 {
   NSEC3RecordContent* ret=new NSEC3RecordContent();
+
+  pr.d_bytesout = 0;
   pr.xfr8BitInt(ret->d_algorithm);
   pr.xfr8BitInt(ret->d_flags);
   pr.xfr16BitInt(ret->d_iterations);
@@ -220,7 +222,8 @@ NSEC3RecordContent::DNSRecordContent* NSEC3RecordContent::make(const DNSRecord &
 
   pr.xfr8BitInt(len);
   pr.xfrBlob(ret->d_nexthash, len);
-  
+  ret->d_size_in_bytes = pr.d_bytesout;
+
   string bitmap;
   pr.xfrBlob(bitmap);
   
@@ -246,6 +249,7 @@ NSEC3RecordContent::DNSRecordContent* NSEC3RecordContent::make(const DNSRecord &
       for(int bit = 0; bit < 8 ; ++bit , val>>=1)
         if(val & 1) {
           ret->d_set.insert((7-bit) + 8*(k) + 256*window);
+          ret->d_size_in_bytes += 2 + 2*sizeof(void*);
         }
       }
   }
--- a/pdns/nsecrecords.cc
+++ b/pdns/nsecrecords.cc
@@ -85,9 +85,7 @@ NSECRecordContent::DNSRecordContent* NSECRecordContent::make(const DNSRecord &dr
 {
   NSECRecordContent* ret=new NSECRecordContent();
 
-  pr.d_bytesout = 0;
   pr.xfrName(ret->d_next);
-  ret->d_size_in_bytes = pr.d_bytesout;
 
   string bitmap;
   pr.xfrBlob(bitmap);
@@ -113,7 +111,6 @@ NSECRecordContent::DNSRecordContent* NSECRecordContent::make(const DNSRecord &dr
       for(int bit = 0; bit < 8 ; ++bit , val>>=1)
         if(val & 1) {
           ret->d_set.insert((7-bit) + 8*(k) + 256*window);
-          ret->d_size_in_bytes += 2 + 2*sizeof(void*);
         }
       }
   }
@@ -134,8 +131,24 @@ string NSECRecordContent::getZoneRepresentation(bool noDot) const
   return ret;
 }
 
+size_t NSECRecordContent::bytes() const
+{
+  return sizeof(*this)
+    + d_next.getStorage().capacity() + 1
+    + (sizeof(uint16_t) + 2 * sizeof(void* )) * d_set.size();
+}
+
 ////// begin of NSEC3
 
+size_t NSEC3RecordContent::bytes() const
+{
+  return sizeof(*this)
+    + d_salt.capacity() + 1
+    + d_nexthash.capacity() + 1
+    // TODO hasn't std::set an overhead of 3 pointers per entry ?
+    // also see NSECRecordContent::bytes
+    + (sizeof(uint16_t) + 2 * sizeof(void* )) * d_set.size();
+}
 void NSEC3RecordContent::report(void)
 {
   regist(1, 50, &make, &make, "NSEC3");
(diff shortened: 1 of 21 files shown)


pdns/recursor_cache.cc
```diff
@@ -371,6 +372,7 @@ int MemRecursorCache::doWipeCache(const DNSName& name, bool sub, uint16_t qtype)
 	break;
       if(iter->d_qtype == qtype || qtype == 0xffff) {
 	count++;
+        d_bytes -= iter->d_bytes;
```
rgacogne: The indentation seems wrong here

pdns/recpacketcache.cc
```diff
@@ -190,20 +198,17 @@ uint64_t RecursorPacketCache::size()
 
 uint64_t RecursorPacketCache::bytes()
```
rgacogne: Could be marked `const`.

pdns/recpacketcache.cc
```diff
@@ -161,6 +165,7 @@ void RecursorPacketCache::insertResponsePacket(unsigned int tag, uint32_t qhash,
     }
```
rgacogne: We need to account for the real size of the protobuf message somehow :-/

pdns/recursor_cache.cc
```diff
@@ -310,6 +302,7 @@ void MemRecursorCache::replace(time_t now, const DNSName &qname, const QType& qt
     }
   }
   ce.d_records.clear();
+  ce.d_bytes=sizeof(ce)+qname.wirelength();
```
rgacogne: In addition to `sizeof(ce)` we need to consider the cost of the indexes. Last time I checked (not recently I'm afraid), the cost for one additional entry was:
* ordered: 3 pointers
* unordered: 2 pointers
* sequenced: 2 pointers

pdns/recpacketcache.cc
rgacogne: And for the size of the indexes, see below.

pdns/nsecrecords.cc
```diff
@@ -109,6 +113,7 @@ NSECRecordContent::DNSRecordContent* NSECRecordContent::make(const DNSRecord &dr
       for(int bit = 0; bit < 8 ; ++bit , val>>=1)
         if(val & 1) {
           ret->d_set.insert((7-bit) + 8*(k) + 256*window);
+          ret->d_size_in_bytes += 2 + 2*sizeof(void*);
```
rgacogne: Shouldn't we have the same thing for `NSEC3` records too?

pdns/recursordist/negcache.cc
```diff
@@ -97,8 +97,21 @@ bool NegCache::get(const DNSName& qname, const QType& qtype, const struct timeva
...
-  replacing_insert(d_negcache, ne);
+void NegCache::add(NegCacheEntry& ne) {
+  ne.d_bytes = sizeof(ne);
+  for (const auto& rec : ne.DNSSECRecords.records) {
+    ne.d_bytes+=rec.d_content->d_size_in_bytes;
+  }
+  for (const auto& sig : ne.DNSSECRecords.signatures) {
+    ne.d_bytes+=sig.d_content->d_size_in_bytes;
+  }
+
```
rgacogne: We also need to account for the records and signatures in `authoritySOA`.

pdns/recursordist/negcache.cc
```diff
@@ -97,8 +97,21 @@ bool NegCache::get(const DNSName& qname, const QType& qtype, const struct timeva
  *
  * \param ne The NegCacheEntry to add to the cache
  */
-void NegCache::add(const NegCacheEntry& ne) {
-  replacing_insert(d_negcache, ne);
+void NegCache::add(NegCacheEntry& ne) {
+  ne.d_bytes = sizeof(ne);
```
rgacogne: We need to account for the size of the indexes, see above.

pdns/recpacketcache.cc
Habbie: > We need to account for the real size of the protobuf message somehow :-/

The python protobuf implementation keeps a running count of the (wire!) size. I haven't spotted such a thing in the C++ lib, but it would be too low anyway. I wonder if just assuming it doubles the size of the whole cache entry is close enough.

pdns/nsecrecords.cc
Habbie: > Shouldn't we have the same thing for NSEC3 records too?

Typed it in but forgot to commit&push; now pushed. The code duplication bothers me there btw.

pdns/recursordist/negcache.cc
Habbie: ah, I missed those because I stole the code from `NegCache::dumpToFile` which also ignores i. I'll fix both.

Habbie: pushed

pdns/recursor_cache.cc
Habbie: mixed tabs and spaces that github gets confused about. pushed.

pdns/recursor_cache.cc
```diff
@@ -333,11 +327,17 @@ void MemRecursorCache::replace(time_t now, const DNSName &qname, const QType& qt
     //    cerr<<"To store: "<<i.d_content->getZoneRepresentation()<<" with ttl/ttd "<<i.d_ttl<<", capped at: "<<maxTTD<<endl;
     ce.d_records.push_back(i.d_content);
     // there was code here that did things with TTL and auth. Unsure if it was good. XXX
+    ce.d_bytes += i.d_content->d_size_in_bytes;
   }
 
+  for(const auto i : signatures) {
+    ce.d_bytes += i->d_size_in_bytes;
+  }
```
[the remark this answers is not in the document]
Habbie: pushed but untested. I note this is missing from dump as well.

pdns/recursor_cache.cc
```diff
@@ -321,6 +314,7 @@ void MemRecursorCache::replace(time_t now, const DNSName &qname, const QType& qt
   if(auth && (qname.isRoot() || !attemptToRefreshNSTTL(qt, content, ce) ) ) {
     // cerr<<"\tGot auth data, and it was not refresh attempt of an unchanged NS set, nuking storage"<<endl;
     ce.d_records.clear(); // clear non-auth data
+    ce.d_bytes=qname.wirelength();
```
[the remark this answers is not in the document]
Habbie: Missed that when improving it after having duplicated it. Pushed.

pdns/recpacketcache.cc
Habbie: pushed; also added const to a few others that did not have it before.

pdns/recursor_cache.cc
Habbie: It might even be implementation dependent, but this is exactly the kind of stuff I put in the NSEC(3) byte count based on some quick googling. For ce/nce I'm pondering leaving it to `*-entries` to map to this overhead but it may not be what people expect.

chbruyand:
I changed the way size of DNSRecordContent objects is calculated.  Could you give me your opinion on that approach ?

Habbie:
I have given up on doing it this way, and i strongly suspect there should be a simpler and better way. Feature request moved to #8327

PR #6448, opened 2018-04-05
Restrict value range for weight parameter, avoid sum overflows dropping queries

dmccombs:
### Short description
This restricts weight values on downstreams and caps the calculated sum of all weights to avoid the following issues:
* Sum of all downstream weights combined exceeding the max value of `int`, causing queries to be dropped with `no-policy`
* Negative values on some downstreams causing a sum of all downstream weights to be 0 or less, causing queries to be dropped with `no-policy`
* Negative value weight preventing any queries from going to a downstream

### Testing
* Run regression tests as described in `regression-tests.dnsdist/README`, particularly `test_Routing.py`
* Manually try adding downstreams with weights that are negative or greater than `2147483647` and observe the logged error and that they are not added, similar behavior to master when adding a bad address
* Add multiple downstreams (with different address/port combinations) that have weights whose sum is greater than `2147483647`, and note that no queries are dropped as `no-policy` shown in `dumpStats()`, as they are when testing this on master

### Checklist
I have:
- [x] documented the code

### Notes

When very high value weights are specified for downstreams to where the sum is greater than the value of `int`, more queries will be sent to downstreams earlier in the list than later ones with the same weight, since the earlier downstream's weight will be a high percentage of the total sum. However, this seems preferable to having queries be dropped with `no-policy` in this edge case.

### DIFF
--- a/pdns/dnsdistdist/docs/reference/config.rst
+++ b/pdns/dnsdistdist/docs/reference/config.rst
@@ -294,7 +294,8 @@ Servers
       address="IP:PORT",     -- IP and PORT of the backend server (mandatory)
       qps=NUM,               -- Limit the number of queries per second to NUM, when using the `firstAvailable` policy
       order=NUM,             -- The order of this server, used by the `leastOustanding` and `firstAvailable` policies
-      weight=NUM,            -- The weight of this server, used by the `wrandom` and `whashed` policies
+      weight=NUM,            -- The weight of this server, used by the `wrandom` and `whashed` policies, default: 1
+                             -- Supported values are a minimum of 1, and a maximum of 2147483647.
       pool=STRING|{STRING},  -- The pools this server belongs to (unset or empty string means default pool) as a string or table of strings
       retries=NUM,           -- The number of TCP connection attempts to the backend, for a given query
       tcpConnectTimeout=NUM, -- The timeout (in seconds) of a TCP connection attempt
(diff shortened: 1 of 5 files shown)


rgacogne:
Unrelated authoritative tests failure in travis (ALIAS), restarted.

pdns/dnsdistdist/docs/reference/config.rst
```diff
@@ -294,7 +294,8 @@ Servers
       address="IP:PORT",     -- IP and PORT of the backend server (mandatory)
       qps=NUM,               -- Limit the number of queries per second to NUM, when using the `firstAvailable` policy
       order=NUM,             -- The order of this server, used by the `leastOustanding` and `firstAvailable` policies
-      weight=NUM,            -- The weight of this server, used by the `wrandom` and `whashed` policies
+      weight=NUM,            -- The weight of this server, used by the `wrandom` and `whashed` policies, default: 1
+                             -- Supported values are a minimum of 1, and a maximum of 2147483647.
```
rgacogne: Technically a `signed int` maximum value could be as small as `32767`, but let's assume we are not going to run on 16-bit.

rgacogne:
Thank you for this PR!

pdns/dnsdistdist/docs/reference/config.rst
dmccombs: Yeah, the actual check is pulling the max dynamically, but I went with the common value for documentation here. If someone happens to run on 16-bit and tries to add something larger than it, at least the error log will say the correct max.

PR #6459, opened 2018-04-09
fix documentation for out-of-zone-additional-processing

gryphius:
### Short description
The documentation for `out-of-zone-additional-processing` (default: `yes` ) currently states 'Do not **enable** this setting if you run a public DNS service with untrusted users.'
However, as the default is `yes` this should probably read "do not **disable** this setting"

### Checklist

I have:
- [ ] documented the code

### DIFF
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -1064,7 +1064,7 @@ always receive AXFR NOTIFYs.
 
 Do out of zone additional processing. This means that if a malicious
 user adds a '.com' zone to your server, it is not used for other domains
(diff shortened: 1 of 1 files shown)


Habbie:
Perhaps we should just remove this feature instead.

gryphius:
I'll downgrade this to a github then, if that's ok. Removing features from PowerDNS is not something I feel comfortable PR'ing ;-)

gryphius:
closing in favour of #6461

PR #6465, opened 2018-04-10
rec: Don't account chained queries more than once

### LINKED ISSUE #6462: Recursor: server-down-max-fails is amplified by chained queries.
- Program: Recursor
 - Issue type: Bug report

### Short description
We recently ran into a problem with ```qname=sip.voip.blackberry.com, qtype=A``` failing to respond from both their auth servers. This situation was compounded by CPE requesting the name at extremely bursty rates (eg: requesting this name hundreds of times within the same second). This lead to a situation where both auth servers ended up in a throttled state.

### Environment

 - Operating system: Linux
 - Software version: 4.1.2
 - Software source: Compiled

### Steps to reproduce
1. Launch 200 queries for a name that upstream will not respond to.
2. Query for a name that would normally succeed

### Expected behaviour
Since the 200 queries for the same name only trigger 1 outgoing query (per NS) due to query chaining (located within ```asendto```) it was expected that it would only count for 1 increment in ```server-down-max-fails``` and the following query would succeed.

### Actual behaviour
1. Initial 200 queries very quickly increment the remoteIP failure rate to exceed ```server-down-max-fails```
2. The re-try for these queries against the second NS quickly does the same
2. This query fails because both servers have been throttled

### Other information
I passed a ```bool& chained``` down from ```doResolveAtThisIP``` into ```asendto``` and changed the value if the query was going to be chained. I then excluded such queries from the timeout code:

```if (s_serverdownmaxfails > 0 && (auth != g_rootdnsname) && !chained && t_sstorage.fails.incr(remoteIP) >= s_serverdownmaxfails) {``` 

Although I did not do extensive testing on this yet, I believe it would limit counting only on-wire (non-chained) timeouts.

### Usecase
I don't believe this situation is limited to just broken auth (although in this case it was). Consider where people only have 2 auth servers and one might be down. In that situation popular names may trigger this based on unlucky packet loss. We often see 'micro bursts' of lookups like this due to local CPE caching expiring. 

### Description
Looking for guidance, but I think failures should only be accounted for if they were actually sent on the wire. This may have implications for lots of timeout/outgoing metric counting as well.

Thanks for reading.

rgacogne:
### Short description

Hopefully fixes #6462, I haven't tested this exact case. Thanks to @phonedph1 for the report and the suggested fix!

### Checklist

I have:
- [ ] documented the code

### DIFF
--- a/pdns/lwres.cc
+++ b/pdns/lwres.cc
@@ -90,7 +90,7 @@ static void logIncomingResponse(std::shared_ptr<RemoteLogger> outgoingLogger, bo
 /** lwr is only filled out in case 1 was returned, and even when returning 1 for 'success', lwr might contain DNS errors
     Never throws! 
  */
-int asyncresolve(const ComboAddress& ip, const DNSName& domain, int type, bool doTCP, bool sendRDQuery, int EDNS0Level, struct timeval* now, boost::optional<Netmask>& srcmask, boost::optional<const ResolveContext&> context, std::shared_ptr<RemoteLogger> outgoingLogger, LWResult *lwr)
+int asyncresolve(const ComboAddress& ip, const DNSName& domain, int type, bool doTCP, bool sendRDQuery, int EDNS0Level, struct timeval* now, boost::optional<Netmask>& srcmask, boost::optional<const ResolveContext&> context, std::shared_ptr<RemoteLogger> outgoingLogger, LWResult *lwr, bool* chained)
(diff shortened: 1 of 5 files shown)


phonedph1:
I tested this patch against 4.1.2 and found it indeed fixes the original behavior described, and very close to what I had come up with the other night.

Thanks @rgacogne !

PR #6469, opened 2018-04-11
rec: Respect the AXFR timeout while connecting to the RPZ server

rgacogne:
### Short description

The initial `AXFR` query to a `RPZ` server was using a hard-coded ten seconds timeout regardless of the `axfrTimeout` setting.
We also messed up error reporting in `waitForRWData()` when `poll` returns `-1`.

### Checklist

I have:
- [ ] documented the code


Habbie:
Does the `waitForRWData` thing affect auth?

rgacogne:
The only thing backporting this PR would change for the auth is make error reporting better when the `poll()` call in `waitForRWData()` fails, which at a quick glance might happen in the `TCP` handling code, in `DNSProxy`, `Carbon`, and while doing an outgoing `AXFR`.

PR #6479, opened 2018-04-12
make lock test pass on macOS

Habbie:
### Short description
With this change, auth `make check` passes on macOS.

### Checklist

I have:
- [ ] documented the code

### DIFF
--- a/pdns/test-lock_hh.cc
+++ b/pdns/test-lock_hh.cc
@@ -41,8 +41,17 @@ BOOST_AUTO_TEST_CASE(test_pdns_lock)
   for(auto& pp : g_locks)
     wlocks.emplace_back(&*pp);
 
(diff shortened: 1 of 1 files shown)


Habbie:
I am unsure this fix is in the right place. One alternative solution would be to treat EDEADLK as non-failure in lock.hh, as that would also fix it for actual production use.

rgacogne:
> One alternative solution would be to treat EDEADLK as non-failure in lock.hh, as that would also fix it for actual production use.

I don't think this would be a good idea. `EDEADLK` means that the current thread already has the lock, while `EBUSY` means another thread has it. While the second case is expected, the first is clearly a programming mistake.

Habbie:
After discussion with @rgacogne, it appears this PR is correct after all.

PR #6484, opened 2018-04-12
recheck serial when AXFR is done

Habbie:
### Short description
When a slave receives a NOTIFY, it (if the serial check says it should) re-transfers the zone. If additional NOTIFYs come in during the transfer, those are acknowledged, even if they mean that the zone we are transferring is not the latest version. With this patch, we will recheck the serial when the transfer is done and, if necessary, do another transfer to make sure are truly up to date with the master.

### Checklist

I have:
- [ ] documented the code


mind04:
LGTM

PR #6486, opened 2018-04-13
Update dns64.rst

phonedph1:
Provide the same link below as in the middle of the document. Currently renders as "To enhance DNS64, see the Lua scripting <lua-scripting/index> documentation." with no actual link on https://doc.powerdns.com/recursor/dns64.html


phonedph1:
.. or maybe it was destined to be https://doc.powerdns.com/recursor/lua-scripting/hooks.html#dns64 instead?

pieterlexis:
> .. or maybe it was destined to be https://doc.powerdns.com/recursor/lua-scripting/hooks.html#dns64 instead?

I think so, yes :)

PR #6489, opened 2018-04-13
ixfrdist: use worker threads to handle TCP connections

### LINKED ISSUE #6468: ixfrdist: starting a new thread to handle TCP connections would be nice
- Program: ixfrdist 
 - Issue type: Feature request

### Short description

Currently `ixfrdist` only uses one thread to handle `UDP` and `TCP` queries, which means that we can only serve one `TCP` client at once. It would be nice to be able to serve more than one.

### Usecase
When serving large zones, the initial `AXFR` might be huge and take a few seconds, during which I would like to be able to handle new `TCP` connections.

pieterlexis:
### Short description
This makes `ixfrdist` no longer block when sending out an AXFR.
To improve the logging in `ixfrdist`, it now uses `Logger`. This PR also patches the `Logger` class to optionally prepend a severity prefix to emitted log lines.

Closes #6468 

### Checklist

I have:
- [x] documented the code

### DIFF
--- a/pdns/ixfrdist.cc
+++ b/pdns/ixfrdist.cc
@@ -531,13 +531,13 @@ void handleUDPRequest(int fd, boost::any&) {
     return;
   }
 
-  if (!allowedByACL(saddr)) {
-    g_log<<Logger::Warning<<"UDP query from "<<saddr.toString()<<" is not allowed, dropping"<<endl;
+  if (saddr == ComboAddress("0.0.0.0", 0)) {
+    g_log<<Logger::Warning<<"Could not determine source of message"<<endl;
     return;
   }
 
-  if (saddr == ComboAddress("0.0.0.0", 0)) {
-    g_log<<Logger::Warning<<"Could not determine source of message"<<endl;
+  if (!allowedByACL(saddr)) {
+    g_log<<Logger::Warning<<"UDP query from "<<saddr.toString()<<" is not allowed, dropping"<<endl;
     return;
   }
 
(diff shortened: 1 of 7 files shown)


pdns/ixfrdist.cc
```diff
@@ -610,109 +574,149 @@ void handleTCPRequest(int fd, boost::any&) {
     cfd = SAccept(fd, saddr);
     setBlocking(cfd);
```
rgacogne: I don't think you need to set it blocking, it will be by default.

pdns/ixfrdist.cc
```diff
@@ -610,109 +574,149 @@ void handleTCPRequest(int fd, boost::any&) {
...
-    cerr<<"[WARNING] TCP query from "<<saddr.toString()<<" is not allowed, dropping"<<endl;
-    close(cfd);
+    g_log<<Logger::Error<<e.what()<<endl;
     return;
   }
 
   if (saddr == ComboAddress("0.0.0.0", 0)) {
-    cerr<<"[WARNING] Could not determine source of message"<<endl;
+    g_log<<Logger::Warning<<"Could not determine source of message"<<endl;
     return;
```
rgacogne: We are leaking `cfd` here.

pdns/ixfrdist.cc
```diff
@@ -610,109 +574,149 @@ void handleTCPRequest(int fd, boost::any&) {
...
-      packets.push_back(packet);
-    }
 
-    if (mdp.d_qtype == QType::AXFR) {
-      if (!makeAXFRPackets(mdp, packets)) {
-        close(cfd);
-        return;
+      vector<vector<uint8_t>> packets;
+      if (mdp.d_qtype == QType::SOA) {
+      vector<uint8_t> packet;
```
rgacogne: Indentation seems off.

pdns/ixfrdist.cc
```diff
@@ -727,7 +731,8 @@ int main(int argc, char** argv) {
       ("server-address", po::value<string>()->default_value("127.0.0.1:5300"), "server address")
       ("work-dir", po::value<string>()->default_value("."), "Directory for storing AXFR and IXFR data")
       ("keep", po::value<uint16_t>()->default_value(KEEP_DEFAULT), "Number of old zone versions to retain")
-      ("axfr-timeout", po::value<uint16_t>()->default_value(AXFRTIMEOUT_DEFAULT), "Timeout in seconds for an AXFR to complete")
+      ("axfr-timeout", po::value<uint16_t>()->default_value(AXFRTIMEOUT_DEFAULT), "Timeout in seconds for an inbound AXFR to complete")
+      ("tcp-out-threads", po::value<uint16_t>()->default_value(10), "Number of maximum simultaneous outbound TCP connections")
```
rgacogne: Isn't that the maximum number of _inbound_ `TCP` connections?

pdns/ixfrdist.cc
pieterlexis: I don't think it hurts right?

Issue #3814, opened 2016-05-04
pdnsutil edit-zone should offer to increase the serial


Habbie:
Right now `edit-zone` does not automatically touch the SOA. Suggestion: if edits are made but SOA is untouched, maybe we can do one of these things:
1. warn about it
2. offer to fix it
3. respect soa-edit settings

zeha:
Related: API has lots of lines to do the SOA-EDIT stuff, and it feels like they are way too many. There's no specific area to work on, except cutting down the number of lines.

Habbie:
2136 also uses SOA-EDIT, and of course SOA queries and the mastercommunicator (for NOTIFY), etc. Some refactoring/centralisation might be in order.

tinuzz:
In the context of this issue, I was trying to write a shell script that:

1. invokes pdns-util edit zone 
2. checks the output for a line that says 'No changes to apply.
1. invoke pdns-util increase-serial if no such line is found

However, step 2 is hard to do, because the message is printed to stdout, rather than stderr, and I haven't found a way to capture stdout without causing at least _some_ issues with the editor.

My request would be that this particular message be sent to stderr instead of stdout.

https://github.com/PowerDNS/pdns/blob/master/pdns/pdnsutil.cc#L1068

`sed '1068 s/cout/cerr/' pdnsutil.cc`

Best regards,
Martijn.

bjoe2k4:
:+1: would prefer option 3

PR #6496, opened 2018-04-14
#3814 - Add question for to increase-soa after non SOA edit

cyclops1982:
### Short description
First stab at #3814.
Simple implementation that does simple SOA change detection. Should work for most people
who just want to let increase-serial do the work.

Logic to increase serial is not centralized and it does not detect if the SERIAL has changed in the SOA. It just looks for a change of the SOA record in general.

I'm happy to also look at the rework of the increase-serial in general, but will need some guidance on that...

### Checklist
I have:
- [X] documented the code


Habbie:
I think it would be good to print all changes first, and then ask the two questions.

cyclops1982:
Ok, that can be easily done, but it will not print the changed SOA records than. Is that ok?

cyclops1982:
Code got updated, please have a quick look.

PR #6495, opened 2018-04-13
only parse resolv.conf once - this avoids race conditions

Habbie:
### Short description
@goestreicher ran into crashes because AuthLua4 again tells us to parse resolv.conf (and add the results to a global vector). This PR fixes that.

### Checklist

I have:
- [x] read the [CONTRIBUTING.md](https://github.com/PowerDNS/pdns/blob/master/CONTRIBUTING.md) document
- [x] compiled and tested this code
- [ ] included documentation (including possible behaviour changes)
- [ ] documented the code
- [ ] added or modified regression test(s)
- [ ] added or modified unit test(s)
- [ ]  checked that this code was merged to master

PR #6499, opened 2018-04-15
stubresolver: Improve locking

cmouse:
### Short description
Improves locking in stub resolver to avoid race conditions.

### Checklist

I have:
- [ ] documented the code

### DIFF
--- a/pdns/stubresolver.cc
+++ b/pdns/stubresolver.cc
@@ -23,7 +23,8 @@
 // s_resolversForStub contains the ComboAddresses that are used by
 // stubDoResolve
 static vector<ComboAddress> s_resolversForStub;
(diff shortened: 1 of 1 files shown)


goestreicher:
So I couldn't reproduce the crash with this patch and *without* #6495, but could still reproduce it without any of them. So I guess it works for me :)

goestreicher:
Additionally I've also tested this patch alone on Travis, where the job failed every time, and it now works.

rgacogne:
@cmouse Ping! Do you disagree with my previous comment? I might be wrong, in which case I'd be happy to merge this :-)

cmouse:
I'll revisit the locking code, it will take few days.

cmouse:
Hoping it'll finally compile nicely.

Issue #6173, opened 2018-01-12
rec: add a default-soa-edit-api setting


atosatto:
- Program: Authoritative
 - Issue type: Feature request

### Short description

Add a `default-soa-edit-api` setting to automatically set the default `soa-edit-api` METADATA for all the created zone.

### Usecase & Description

Editing PowerDNS Authoritative DNS zones RRsets via API it would be convenient to a have a default configuration setting for the `soa-edit-api` METADATA instead of having to set it per-zone.

jangrewe:
Oh yes, this would be very appreciated, especially considering that a lot of people who use Let's Encrypt will switch to the DNS-based authentication challenge soon, as that's required for wildcard certificates, and you really, really, really want to automate this - preferably via the API! ;-)
(And you then want slave notifications as quickly as possible, which currently doesn't happen unless you set the metadata.)

zeha:
Note that, if you create the zone using the api, SOA-EDIT-API defaults to DEFAULT, if `soa_edit_api` is not present in the create request.

Napsty:
What about if pdns would lookup the pdns.conf file and check for the option `default-soa-edit`?

```
# grep "^default-soa-edit" /etc/powerdns/pdns.conf
default-soa-edit=INCEPTION-INCREMENT
```
 
If a zone is created through the API and the request not contain the `soa_edit_api` key, use the default from the config file. Any reason, why this is not read from the config file?
Or, if that somehow breaks something, add a second option into the config file, e.g. `default-soa-edit-api`.

apezio:
I could definitely use this setting.  I am using master/slave and if the a zone is created on master, without  setting this metadata to INCREASE (which can only be done manually) the zone never axfrs after being updated on the master via the API resulting in stale records on the slave.

zeha:
If you use the API to create zones, it should just work. Otherwise pass SOA-EDIT-API in the zone create API call.

bessone:
To celebrate this issue's 8th anniversary 🎂, is there any chance that this setting will be integrated?

Thanks!

PR #6501, opened 2018-04-15
Implement #6173

cyclops1982:
### Short description
This adds very simple default-soa-edit-api support, which is requested in #6173 
I'm not sure if:
- we maybe should re-use default-soa-edit
- We should check for valid input (maybe soa-edit-api should be an enum or some sort to be able to parse it and validate that it's correct)
- The docs are good. I haven't gotten them working, but wanted to submit this pull request to get a review.

### Checklist
I have:
- [ ] documented the code

### DIFF
--- a/pdns/ws-auth.cc
+++ b/pdns/ws-auth.cc
@@ -568,6 +568,10 @@ static void updateDomainSettingsFromDocument(UeberBackend& B, const DomainInfo& 
   if (document["kind"].is_string()) {
     di.backend->setKind(zonename, DomainInfo::stringToKind(stringFromJson(document, "kind")));
   }
+  
+  if (arg()["default-soa-edit-api"] != "" && !document["soa_edit_api"].is_string()) {
+    di.backend->setDomainMetadataOne(zonename, "SOA-EDIT-API", arg()["default-soa-edit-api"]);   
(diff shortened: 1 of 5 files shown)


pdns/ws-auth.cc
```diff
@@ -568,6 +568,10 @@ static void updateDomainSettingsFromDocument(UeberBackend& B, const DomainInfo&
   if (document["kind"].is_string()) {
     di.backend->setKind(zonename, DomainInfo::stringToKind(stringFromJson(document, "kind")));
   }
+  
+  if (arg()["default-soa-edit-api"] != "" && !document["soa_edit_api"].is_string()) {
```
pieterlexis: There is a side-effect here. `updateDomainSettingsFromDocument` is not only called when a domain is created, but also when a zone is modified with HTTP PUT. This means that existing zones with an empty (or non-existing) `SOA-EDIT-API` metadata will suddenly have this metadata data attached to them when editing e.g. records.

I think this default metadata should only be set on creation around [here](https://github.com/cyclops1982/pdns/blob/e6b32a057708bfd70f017ad6a264a03466b8ce47/pdns/ws-auth.cc#L1224-L1226)

PR #6504, opened 2018-04-16
rec: Add an optional stack cache to MTasker

rgacogne:
### Short description

The stack cache is disabled by default and can be enabled by setting `mtasker-stacks-cache-size` to a non-zero value. When the cache is enabled the internal vector is not used, and a new one is
dynamically allocated instead, to be able to store it into a per-MTasker free list when the context is released and then reuse it for a new context later.
Otherwise there should be no noticeable overhead.

Ideally the allocator should have a per-thread cache and save us the trouble. While it seems to be the case on recent (>= 2.26) glibc, it looks like older versions don't have a per-thread free-list, or at least not for so large objects. 

### Checklist

I have:
- [ ] documented the code

### DIFF
--- a/pdns/recursordist/docs/settings.rst
+++ b/pdns/recursordist/docs/settings.rst
@@ -779,6 +779,19 @@ This setting artificially raises all TTLs to be at least this long.
 While this is a gross hack, and violates RFCs, under conditions of DoS, it may enable you to continue serving your customers.
 Can be set at runtime using ``rec_control set-minimum-ttl 3600``.
 
+.. _setting-mtasker-stacks-cache-size:
+
+``mtasker-stacks-cache-size``
+------------------------
+.. versionadded:: 4.2.0
+
+-  Integer
+-  Default: 0 (disabled)
+
+This setting enables a cache for the stacks used by the recursor. When enabled, stacks are put into a free-list when
+a context is released until the number of available stacks reaches the configured value, and new contexts will reuse
+available stacks before allocating new ones.
+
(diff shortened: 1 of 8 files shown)


pdns/recursordist/docs/settings.rst
```diff
@@ -779,6 +779,19 @@ This setting artificially raises all TTLs to be at least this long.
...
+``mtasker-stacks-cache-size``
+------------------------
+.. versionadded:: 4.2.0
+
+-  Integer
+-  Default: 0 (disabled)
+
+This setting enables a cache for the stacks used by the recursor. When enabled, stacks are put into a free-list when
+a context is released until the number of available stacks reaches the configured value, and new contexts will reuse
+available stacks before allocating new ones.
```
pieterlexis: Perhaps a few words about this should also go into the tuning guide?

rgacogne:
Added a few words about the `MTasker` settings in the performance guide!

ahupowerdns:
have we measured if this makes a difference in performance?

rgacogne:
> have we measured if this makes a difference in performance?

I have not done the measurement on an older distribution yet, so no. On a recent distribution (Arch current) it does not seem to improve nor degrade performance, I'm guessing because the per-thread cache is doing well enough on its own.
This PR might very well prove useless in the end, but we have suspected for a while that these large allocations might be hurting us on older `libc`, so I thought it was worth writing the code so we could build packages and see if our suspicions were accurate (or not, in which case we can concentrate on the other leads we have).

PR #6505, opened 2018-04-16
rec: Add hashed indexes to the caches, for faster retrieval

rgacogne:
### Short description

Most of the lookups in the query (positive) and negative caches are done on the exact name only, and doing so using a hash is much more faster than using canonical ordering.
In both cases we still need canonical ordering to be able to wipe a domain and its subtree from the cache efficiently, but regular retrieval (cachecache, get()) doesn't.

Pros:
- Faster retrieval: ~8x faster

Cons:
- Increased memory usage: ~10% overhead worst case (entries with only one small record)
- Slower insertion: 1.08x slower

### Checklist

I have:
- [ ] documented the code

### DIFF
--- a/pdns/recursor_cache.cc
+++ b/pdns/recursor_cache.cc
@@ -33,7 +33,7 @@ unsigned int MemRecursorCache::bytes() const
     ret+=sizeof(struct CacheEntry);
     ret+=(unsigned int)i.d_qname.toString().length();
     for(const auto& record : i.d_records)
-      ret+= sizeof(record); // XXX WRONG we don't know the stored size! j->size();
+      ret+= sizeof(record); // XXX WRONG we don't know the stored size!
   }
   return ret;
 }
--- a/pdns/recursor_cache.cc
+++ b/pdns/recursor_cache.cc
@@ -29,16 +29,16 @@ unsigned int MemRecursorCache::bytes() const
 {
   unsigned int ret=0;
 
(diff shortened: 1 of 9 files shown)


pdns/recursor_cache.cc
```diff
@@ -29,16 +29,16 @@ unsigned int MemRecursorCache::bytes() const
...
 
-  for(cache_t::const_iterator i=d_cache.begin(); i!=d_cache.end(); ++i) {
+  for(const auto& i : d_cache) {
     ret+=sizeof(struct CacheEntry);
-    ret+=(unsigned int)i->d_qname.toString().length();
-    for(auto j=i->d_records.begin(); j!= i->d_records.end(); ++j)
-      ret+= sizeof(*j); // XXX WRONG we don't know the stored size! j->size();
+    ret+=(unsigned int)i.d_qname.toString().length();
+    for(const auto& record : i.d_records)
+      ret+= sizeof(record); // XXX WRONG we don't know the stored size! j->size();
```
pieterlexis: 1. Is this comment still relevant?
2. `j->size()` has no meaning here

Habbie: given #6447, maybe just remove the comment here

rgacogne: The comment is still relevant (agreed, `j` doesn't make sense anymore) until #6447 is merged so I'd rather keep it. I'll remove the  `j->size();` part.

PR #6512, opened 2018-04-17
[WIP] luawrapper: work with luajit on ARM64

Habbie:
### Short description
Lua `lightuserdata` on ARM64 with LuaJIT is limited to addresses that can fit in 47 bits. This PR removes all usage of `lua_pushlightuserdata` from our copy of Luawrapper. The `ValueInRegistry` and `plain function pointer` changes are mostly harmless; the type index changes might have some performance impact.

This makes the dnsdist test suite pass on Linux/ARM64.

PRing this now for reviews and comments. If it turns out the type index changes have real performance impact, I will consider making that a compile time choice.

~When force applied to upstream Luawrapper, this breaks compilation of the `custom_types`, `movable` and `metatables` test suite components.~ Fixed in `Revert "no longer need this" `

### TODO
- [x] deal with the last remaining `lua_pushlightuserdata` call (in `template<> struct LuaContext::Pusher<const std::type_info*>`) - either make sure we never call it, or fix it. Just removing that template causes the upstream LuaWrapper failure mentioned above. I checked this box after adding an assert to that call, proving we do not rely on it.
- [x] update comments
- [ ] check performance impact of these changes

### Checklist

I have:
- [ ] documented the code

### DIFF
--- a/ext/luawrapper/include/LuaContext.hpp
+++ b/ext/luawrapper/include/LuaContext.hpp
@@ -32,12 +32,14 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include <array>
 #include <cassert>
 #include <cmath>
+#include <cstdlib>
 #include <cstring>
 #include <functional>
 #include <limits>
 #include <list>
 #include <map>
 #include <memory>
+#include <iostream>
 #include <random>
 #include <set>
 #include <stdexcept>
@@ -97,9 +99,10 @@ public:
         // setting the panic function
         lua_atpanic(mState, [](lua_State* state) -> int {
             const std::string str = lua_tostring(state, -1);
+            std::cerr<<"lua panic: "<<str<<std::endl;
             lua_pop(state, 1);
             assert(false && "lua_atpanic triggered");
-            exit(0);
+            exit(EXIT_FAILURE);
         });
 
         // opening default library if required to do so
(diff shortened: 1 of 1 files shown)


Habbie:
Travis and Buildbot pass this as of 5a01fb3

Habbie:
This is ready for review, and, as far as I'm concerned, merge to master after some squashing.

Habbie:
Some benchmark results: https://gist.github.com/Habbie/4da8f77e66248ad0778d4481c9ef78f0

Habbie:
> Code looks good. Am I understanding correctly that our current FFI API will not work on ARM64 or anything else that uses more than 47 bits of addresses?

No, it's only lightuserdata that is affected by the 47 bit limit, as far as I can tell.

rgacogne:
> No, it's only lightuserdata that is affected by the 47 bit limit, as far as I can tell.

But isn't that what we use for our `pdns_ffi_param` type, to maximize performance and because I couldn't get `FFI` to work when the type was wrapped by `LuaWrapper`?

Habbie:
> But isn't that what we use for our pdns_ffi_param type, to maximize performance and because I couldn't get FFI to work when the type was wrapped by LuaWrapper?

Oh, yes. We'll have to look into that.

rgacogne:
Benchmarks look good, and as far as I understand it, it won't break the existing recursor `FFI` `API` on `x86_64` so I'd be happy to merge this once you are done with it. I think you mentioned wanting to do some squashing first?

Habbie:
Closing this until we figure out the performance impact. Still open, in parts, at:
* https://github.com/ahupowerdns/luawrapper/pull/43
* https://github.com/ahupowerdns/luawrapper/pull/44
* https://github.com/ahupowerdns/luawrapper/pull/45
* https://github.com/ahupowerdns/luawrapper/pull/46

Issue #6510, opened 2018-04-17
Add Swagger codegen test for docs


ffledgling:
- Program: Authoritative, 
 - Issue type: Feature request

### Short description

We've had swagger Auth HTTP for a little while now, however, generating client code from this seems to be susceptible to breakage as the docs evolve over time. The idea is to add a travis task (using the build matrix's include directive in `.travis.yml`) that tries to generate the bindings from the swagger spec and complains loudly if it can't.

### Environment

 - Operating system: All
 - Software version: Master

### Steps to reproduce

```
#!/bin/bash

set -euxo pipefail

wget http://central.maven.org/maven2/io/swagger/swagger-codegen-cli/2.3.0/swagger-codegen-cli-2.3.0.jar -O swagger-codegen-cli.jar

java -jar swagger-codegen-cli.jar generate \
    -c swagger-config.json \
    -i pdns/docs/http-api/swagger/authoritative-api-swagger.yaml \
    -l go \
    -o out/go

tree out/go
```
### Expected behaviour
A client library for golang should be generated.

### Actual behaviour
Returns a bunch of errors and warnings, like:
```
[main] WARN io.swagger.codegen.DefaultCodegen - skipping invalid array property
...
[main] ERROR io.swagger.codegen.DefaultCodegen - unexpected missing property for name Response
...
[main] ERROR io.swagger.codegen.DefaultCodegen - unexpected missing property for name response
```
### Other information
Most of these are syntax/semantic issues in the swagger spec definition. If there are tests for the codegen, it'll provide feedback at the time of reviewing the PR and allow people to fix issues instead of letting them slide unnoticed.

PR #6515, opened 2018-04-19
Add tests for swagger-codegen in travis

ffledgling:
We now attempt to generate valid client bindings from the in-repo
swagger spec. If this fails, there's something wrong with the spec.

### Short description
Fixes issue #6510 , by adding a test for the swagger codegen client.

### Checklist

I have:
- [ ] documented the code

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,7 @@ matrix:
       addons:
         apt:
           packages: ['clang-3.6']
+    - env: PDNS_BUILD_PRODUCT=nobuild_swagger
 
 before_script:
   - git describe --always --dirty=+
--- a/build-scripts/travis.sh
+++ b/build-scripts/travis.sh
@@ -369,6 +369,20 @@ install_dnsdist() {
   run "sudo chmod 0755 /var/agentx"
 }
 
+install_nobuild_swagger() {
+  # Install swagger codegen
+
+  TMP_DIR="$(mktemp -d)"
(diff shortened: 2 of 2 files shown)


ffledgling:
To add some context to this seemingly random PR:

The test that this PR adds is *supposed* to fail, because the swagger spec in the master branch is broken at the moment. The reason for the addition of this test is precisely to prevent commits from breaking the spec in the future.

However, fixing the spec is an entirely different ball game, one which should probably not be tied into this PR.

So the decision for the moment (quoting @Habbie ) is:
```
06:33 < Habbie> so, i understand that adding this test was 'easy'
06:33 < Habbie> but making it pass is 'hard'
06:33 < Habbie> i am okay with noting that clearly in the PR
06:34 < Habbie> so we have this test setting ready for the day we are, well, ready for it
```

jsoref:
Hi, for the time being (and hopefully going forward), pdns master is running the [check-spelling](https://github.com/marketplace/actions/check-spelling) action. In order to avoid a ❌ when merged https://github.com/jsoref/pdns/commit/27fbadcc761a9e20a204d6c5de3d3633581fa28f#commitcomment-39230833 -- you'll need to merge/rebase on `master` to get the check (and the files) -- you can accept the advice (which should add the items to `expect.txt`)

Habbie:
Do you want to rebase this for a fresh look at the status of our yaml?

pieterlexis:
we no longer use travis. Closing this

PR #6516, opened 2018-04-19
auth: have a global g_log (backport)

pieterlexis:
### Short description
Not really a backport, but a re-do because of too many conflicts.

### Checklist

I have:
- [ ] documented the code
- [x] checked that this code was merged to master

### DIFF
--- a/modules/bindbackend/bindbackend2.cc
+++ b/modules/bindbackend/bindbackend2.cc
@@ -488,7 +488,7 @@ void Bind2Backend::insertRecord(BB2DomainInfo& bb2, const DNSName &qname, const 
   else {
     string msg = "Trying to insert non-zone data, name='"+bdr.qname.toLogString()+"', qtype="+qtype.getName()+", zone='"+bb2.d_name.toLogString()+"'";
     if(s_ignore_broken_records) {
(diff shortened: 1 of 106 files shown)


zeha:
is backporting this really a good idea...?

Habbie:
> is backporting this really a good idea...?

We don't compile against Boost 1.67.0 without it (or, perhaps, something like it).

https://github.com/freebsd/freebsd-ports/blob/6fa3dca03cf2/dns/dnsdist/files/patch-boost-1.67

https://github.com/Homebrew/homebrew-core/pull/26812

zeha:
> We don't compile against Boost 1.67.0 without it (or, perhaps, something like it).

I see. Maybe 4.1 should carry the "ugly" fix to keep the diff between 4.1.x -> 4.1.y small, so patches from 4.1.y can be applied to earlier versions more easily.

pieterlexis:
> I see. Maybe 4.1 should carry the "ugly" fix to keep the diff between 4.1.x -> 4.1.y small, so patches from 4.1.y can be applied to earlier versions more easily.

patches _to_ 4.1.y would come from master and would be harder to apply though

PR #6518, opened 2018-04-19
rec: Add a new max-udp-queries-per-round setting

rgacogne:
### Short description

This new setting limits the number of `UDP` queries we attempt to handle after being woken up by the multiplexer and before returning back to process other events.
Before this, we could end up trying to process queries after queries and almost never return from `handleNewUDPQuestion()` to process new events, meaning we could eventually end up never scheduling new `mthreads` or handle responses from authoritative servers for a long time, only sending responses for packetcache hits and creating new `mthreads`.

This would also make `rec_control` seem unresponsive and return with:
```
Fatal: Timeout waiting for answer from control channel
```

### Checklist

I have:
- [ ] documented the code


pieterlexis:
Could you add some words on what a "round" is in this context?

PR #6526, opened 2018-04-23
 calidns: Add an option to read ECS values from the query file, skip comments

rgacogne:
### Short description

If the `--ecs-from-file` option is used, `calidns` read an IP address or netmask from a third space-separated field in the query file, and adds it to the query as an `EDNS Client Subnet` option.
This PR also makes `calidns` ignore lines starting with a `#` in the query file, to be able to put comments in it.

### Checklist

I have:
- [ ] documented the code

### DIFF
--- a/pdns/calidns.cc
+++ b/pdns/calidns.cc
@@ -318,6 +318,9 @@ try
     vector<uint8_t> packet;
     DNSPacketWriter::optvect_t ednsOptions;
     boost::trim(line);
+    if (line.empty() || line.at(0) == '#') {
+      continue;
+    }
 
     auto fields = splitField(line, ' ');
     std::string qname = fields.first;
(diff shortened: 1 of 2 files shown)


pdns/calidns.cc
```diff
@@ -316,14 +318,28 @@ try
     vector<uint8_t> packet;
     DNSPacketWriter::optvect_t ednsOptions;
     boost::trim(line);
-    const auto fields = splitField(line, ' ');
-    DNSPacketWriter pw(packet, DNSName(fields.first), DNSRecordContent::TypeToNumber(fields.second));
+    if (line.empty() || line.at(0) == '#') {
+      continue;
+    }
+
+    auto fields = splitField(line, ' ');
```
pieterlexis: Would using `vector<string> parts; stringtok(parts, line, " ")` not be easier? Using the resulting vector's length to determine if subnet info was required (and ignoring it if the switch is not used).

rgacogne: Sure, `stringtok()` is a bit more complex but that might makes things easier to read, so why not.

PR #6529, opened 2018-04-24
configure: ensure lua is setup when lua records are enabled

chbruyand:
### Short description
This will trigger an error if LUA hasn't been found and LUA records were not disabled.

### Checklist

I have:
- [ ] documented the code


Habbie:
Works!

Could you improve this one as well?

```
checking whether libcurl is usable... no
configure: error: libcurl minimum version requirement not met. This is required for LUA records. You can disable it with the --disable-lua-records switch or use --with-libcurl to select another curl installation.
```

my curl is not too old - I simply have no curl.

PR #6531, opened 2018-04-24
avoid interleaved access to B (via d_dk)

Habbie:
### Short description
Before this patch, the meta lookup would interfere with the already-started `B.lookup`. This caused failures with odbc/MSSQL.

### Checklist

I have:
- [ ] documented the code


Habbie:
> style nit: i think any two calls to .lookup() and .get() should really be on directly following lines, so errors like this are not (re-)introduced by accident.

:+1:

PR #6538, opened 2018-04-30
Add custom config for lgtm.com

zeha:
### Short description

Setup auth, recursor and dnsdist for lgtm.com.

Unclear what to do about Lua - which is required for dnsdist and previously we thought making this non-optional for recursor, too.

### Checklist

I have:
- [ ] documented the code

### DIFF
--- a/lgtm.yml
+++ b/lgtm.yml
@@ -1,0 +1,34 @@
+path_classifiers:
+  test: "regression-tests*"
+  library:
+  - "ext"
+  - exclude: ext/luawrapper
+  - exclude: ext/yahttp
+  - "docs"
+  - "pdns/recursordist/html/js"
+  - "pdns/dnsdistdist/src_js"
+  - "pdns/dnsdistdist/html/js"
+extraction:
+  cpp:
+    configure:
+      command:
+      - "pushd pdns/dnsdistdist"
+      - "sed -i '/^dist_man_MANS/d' Makefile.am"
+      - "autoreconf -vi && ./configure"
+      - "popd"
+      - "pushd pdns/recursordist/"
+      - "sed -i '/^dist_man_MANS/d' Makefile.am"
+      - "autoreconf -vi && ./configure"
+      - "popd"
+      - "sed -i 's/codedocs docs//' Makefile.am"
+      - "autoreconf -vi"
+      - "./configure --with-modules='bind gsqlite3 gmysql gpgsql ldap'"
+    index:
+      build_command:
+      - "pushd pdns/dnsdistdist"
+      - "make"
+      - "popd"
+      - "pushd pdns/recursordist"
+      - "make"
+      - "popd"
+      - "make"


sjvs:
I think you may need to `pushd` and `popd` in the `index` section as well

Habbie:
This pull request **introduces 6 alerts** and **fixes 36** when merging 0e333f5caa9e279da5dfb7aec1e4d970e1a681db into 2ed977eae67bda7a8d3b1fad247e43185d3c632b - [view on lgtm.com](https://lgtm.com/projects/g/PowerDNS/pdns/rev/pr-213dff6cece5030853ce712ad8291f4654d3b82b)

**new alerts:**

* 2 for Constant return type
* 2 for Constant return type on member
* 1 for Catching by value
* 1 for Comparison result is always the same

**fixed alerts:**

* 22 for Short global name
* 5 for Catching by value
* 2 for Missing header guard
* 2 for Resource not released in destructor
* 2 for Comparison result is always the same
* 1 for Inconsistent definition of copy constructor and assignment \(&#39;Rule of Two&#39;\)
* 1 for Missing return statement
* 1 for Function declared in block

---

*Comment posted by [lgtm.com](https://lgtm.com)*

sjvs:
@zeha: your config clearly worked! The new alerts are a result of analysing more code, so that's great (in some way :slightly_smiling_face:)

I'm experimenting with adding the `gmysql` and `gpgsql` modules for analysis. Will post here if/when I have some results.

Habbie:
@sjvs any reason lgtm does not have a github 'bot' account to post the comments from?

sjvs:
@Habbie: that's something we're working on. In the meantime, you can of course create your own bot account for this purpose. That's what the AMPHTML guys have done: https://github.com/amphtml-lgtm-bot

The way to do that:
- Disable LGTM's PR integration from your @Habbie account (https://lgtm.com/projects/g/PowerDNS/pdns/ci/)
- Create a bot account, give it admin permissions on the `pdns` repo (only required temporarily)
- Log into GitHub and LGTM as the bot account
- Enable LGTM's PR integration: https://lgtm.com/projects/g/PowerDNS/pdns/ci/
- Optional: revoke the bot's admin permissions on the `pdns` repo

lgtm.yml
```diff
@@ -0,0 +1,30 @@
...
+      - "sed -i '/^dist_man_MANS/d' Makefile.am"
+      - "autoreconf -vi && ./configure"
+      - "popd"
+      - "pushd pdns/recursordist/"
+      - "sed -i '/^dist_man_MANS/d' Makefile.am"
+      - "autoreconf -vi && ./configure"
+      - "popd"
+      - "sed -i 's/codedocs docs//' Makefile.am"
+      - "autoreconf -vi"
+      - "./configure --with-modules=\"bind\""
```
sjvs: I'm fairly sure that `--with-modules='bind gmysql gpgsql'` would just work: the dependencies will be detected and installed automagically by LGTM. I think you may have to use single quotes though.

zeha:
@sjvs it appears to have trouble with us running make three times though.
https://lgtm.com/projects/g/PowerDNS/pdns/rev/pr-213dff6cece5030853ce712ad8291f4654d3b82b shows many files created/deleted.

zeha:
Maybe this is my own doing, hopefully fixed in c0aa906.

sjvs:
@zeha: LGTM compares the 'before' and 'after' situation. This PR changes the build process significantly, so the 'after' is very different from the 'before'. Although those files are not 'new', it is the first time they're visible to the LGTM build/analysis process. With that in mind, do you think the results are wrong?

sjvs:
@zeha: I just noticed that your new build script only partially works. It only builds `dnsdistdist` but fails to build `recursordist` and the main code. That's why so many source code files appear to be missing/deleted: because the new build command never considers them :slightly_smiling_face:.

Your recent fix should resolve this (by using `pushd` and `popd` in the `index` step), so LGTM will probably comment soon with the results.

Issue #6535, opened 2018-04-26
DNSDist: Error message is useless userdata


kaosdrachen:
- Program:  dnsdist 
 - Issue type: Bug report

### Short description

While scripting an advanced config I encountered a bug in the way dnsdist returns errors 
back to lua that renders the error message as returned by pcall() into a useless userdata object

### Environment
 - Operating system: Debian
 - Software version: 1.3.0
 - Software source: PDNS repository for DNSdist 1.3

### Steps to reproduce
1. Install dnsdist
2. Add the following into a config file:

```
x = newNMG()
print(pcall(function() x:addMask('banana') end))
```
3. Run the code; 

### Expected behaviour
A readable error message; In this case it should return 'Unable to convert 'banana' to a netmask'

### Actual behaviour
it will print 'false userdata' followed by a pointer.

### Other information
I've discussed this in IRC with Habbie; they're the one who recommended filing a bug report.

Habbie:
> it will print 'false userdata' followed by a pointer.

To be clear, the return value from pcall is `false` and the useless `userdata`. This needs a sprinkling of `rethrow_if_nested` and/or some LuaWrapper bindings for exception objects.

chbruyand:
Closing the issue as it has been fixed by #6541

PR #6541, opened 2018-04-30
dnsdist: allow known exception types to be converted to string

chbruyand:
### Short description
This tries to improve Lua error reporting, see #6535. Exceptions thrown in a Lua subroutine can be printed:
```
x = newNMG()
res, err = pcall(function() x:addMask('banana') end)
if not res then
   print("An error has been thrown: ", err)
end
```

### Checklist

I have:
- [ ] documented the code

### DIFF
--- a/pdns/dnsdist-lua-bindings.cc
+++ b/pdns/dnsdist-lua-bindings.cc
@@ -44,6 +44,21 @@ void setupLuaBindings(bool client)
       g_outputBuffer+="\n";
     });
 
+  /* Exceptions */
+  g_lua.registerFunction<string(std::exception_ptr::*)()>("__tostring", [](const std::exception_ptr& eptr) {
+      try {
+        if (eptr) {
+          std::rethrow_exception(eptr);
+        }
+      } catch(const std::exception& e) {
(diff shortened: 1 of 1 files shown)


chbruyand:
For the record, we could also catch some specific exception and convert them to `lua_error` directly in the Lua wrapper : https://gist.github.com/chbruyand/586820716b128cf680d12599fd46adc7

This allows to get more details about the error
```
res, err = pcall(function() throwSomeException() end)
if not res then
   print (err)
end
```
will output
```
[string "chunk"]:1: Caught exception: Something bad happened
```
But it would also require to catch `PDNSExceptions`s as they  diverge from `std::exception`s, which would not be so clean.

zeha:
> But it would also require to catch PDNSExceptionss as they diverge from std::exceptions, which would not be so clean.

Maybe its time to ditch PDNSException or make it a std subclass.

ahupowerdns:
will this work with Lua's assert() idiom? https://www.lua.org/pil/8.3.html

ahupowerdns:
looks useful!

PR #6562, opened 2018-05-02
rec api: add subtree option to the cache flush endpoint

### LINKED ISSUE #6550: rec: Flushing a subtree, not just the exact name, from the caches via the API
- Program: Recursor
 - Issue type: Feature request

### Short description
Our API has a nice endpoint to flush a given name from the cache, as documented in [1]. However it only supports the removal of an exact name, not a whole subtree, as is possible via `rec_control wipe-cache '.$'`. Most of the needed code is already there, we would probably just need an additional parameter to be able to pass `true` instead of `false` to the existing cache cleaning functions, plus the documentation and some tests.

[1]: https://doc.powerdns.com/recursor/http-api/endpoint-cache.html

chbruyand:
### Short description
This adds support for the `subtree` parameter to the cache flush API endpoint.
```
/api/v1/servers/localhost/cache/flush?domain=example.com.&subtree=true
```

Fix #6550 

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/recursordist/docs/http-api/endpoint-cache.rst
+++ b/pdns/recursordist/docs/http-api/endpoint-cache.rst
@@ -7,7 +7,10 @@ Cache manipulation endpoint
 
   :query server_id: The name of the server
   :query domain: The domainname to flush for
-  :query subtree: If set to `true`, also flush the whole subtree (default=`false`)
+
+  .. versionadded:: 4.1.3
+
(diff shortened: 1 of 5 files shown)


zeha:
Sorry, the review request got buried in other gh mail :(

PR #6566, opened 2018-05-03
rec: Use a separate, non-blocking pipe to distribute queries

rgacogne:
### Short description

This allows us to drop queries when a pipe goes full, thus still distributing queries to other threads instead of blocking. It also adds a new metric to keep track of queries dropped because the pipe was full.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/pdns_recursor.cc
+++ b/pdns/pdns_recursor.cc
@@ -2415,7 +2415,10 @@ static void makeThreadPipes()
       unixDie("Creating pipe for inter-thread communications");
     tps.readQueriesToThread = fd[0];
     tps.writeQueriesToThread = fd[1];
-    setNonBlocking(tps.writeQueriesToThread);
+
+    if (!setNonBlocking(tps.writeQueriesToThread)) {
+      unixDie("Making pipe for inter-thread communications non-blocking");
+    }
 
     g_pipes.push_back(tps);
   }
--- a/pdns/pdns_recursor.cc
+++ b/pdns/pdns_recursor.cc
@@ -128,6 +128,8 @@ struct ThreadPipeSet
   int readToThread;
   int writeFromThread;
   int readFromThread;
+  int writeQueriesToThread; // this one is non-blocking
+  int readQueriesToThread;
 };
 
 typedef vector<int> tcpListenSockets_t;
@@ -2409,6 +2411,12 @@ static void makeThreadPipes()
     tps.readFromThread = fd[0];
     tps.writeFromThread = fd[1];
 
+    if(pipe(fd) < 0)
+      unixDie("Creating pipe for inter-thread communications");
+    tps.readQueriesToThread = fd[0];
+    tps.writeQueriesToThread = fd[1];
+    setNonBlocking(tps.writeQueriesToThread);
+
     g_pipes.push_back(tps);
   }
 }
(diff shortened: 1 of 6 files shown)


pdns/pdns_recursor.cc
```diff
@@ -2463,17 +2471,29 @@ void distributeAsyncFunction(const string& packet, const pipefunc_t& func)
   tmsg->func = func;
   tmsg->wantAnswer = false;
 
-  if(write(tps.writeToThread, &tmsg, sizeof(tmsg)) != sizeof(tmsg)) {
+  ssize_t written = write(tps.writeQueriesToThread, &tmsg, sizeof(tmsg));
```
chbruyand: Is there any specific reason you did not keep the test like it was before ? I find it more readable like that
```
  if (write(tps.writeQueriesToThread, &tmsg, sizeof(tmsg)) != sizeof(tmsg)) {                                                                                                 
    delete tmsg;                                                                                                                                                              
    int error = errno;                                                                                                                                                        
    if (error == EAGAIN || error == EWOULDBLOCK) {                                                                                                                            
      g_stats.queryPipeFullDrops++;                                                                                                                                           
    } else {                                                                                                                                                                  
      unixDie("write to thread pipe returned wrong size or error:" + error);                                                                                                  
    }                                                                                                                                                                         
  } 
```

pdns/recursordist/RECURSOR-MIB.txt
```diff
@@ -758,6 +758,14 @@ policyResultCustom OBJECT-TYPE
         "Number of policy-mandated custom results"
     ::= { stats 91 }
 
+queryPipeFullDrops OBJECT-TYPE
+    SYNTAX Counter64
+    MAX-ACCESS read-only
+    STATUS current
+    DESCRIPTION
+        "Number of responses dropped because the query distribution pipe was full"
```
chbruyand: Number of queries maybe ?

pdns/pdns_recursor.cc
```diff
@@ -2409,6 +2411,12 @@ static void makeThreadPipes()
     tps.readFromThread = fd[0];
     tps.writeFromThread = fd[1];
 
+    if(pipe(fd) < 0)
+      unixDie("Creating pipe for inter-thread communications");
+    tps.readQueriesToThread = fd[0];
+    tps.writeQueriesToThread = fd[1];
+    setNonBlocking(tps.writeQueriesToThread);
```
chbruyand: Would it be a problem if `setNonBlocking()` would fail ?

pdns/pdns_recursor.cc
rgacogne: The value or `errno` only makes sense if `write()` returned `-1`, so we need to make sure it's the case, because otherwise we might be reading a garbage value of `errno` because of a short write. Agreed, it should not be happening because writing a 64-bits pointer on a pipe is supposed to be atomic since 64 < `PIPE_BUF`, but I'd rather be on the safe side.

pdns/recursordist/RECURSOR-MIB.txt
rgacogne: Indeed, good catch!

pdns/pdns_recursor.cc
rgacogne: It would not really be an issue since we would just be in the same situation than we were previously but I'll add a check nevertheless!

PR #6567, opened 2018-05-03
rec: Move carbon/webserver/control/stats handling to a separate thread

rgacogne:
### Short description

This makes sure that no worker or distributor thread will get blocked while waiting for a response from another thread, for example while gathering stats or executing a command coming from the control
channel.

This PR needs a serious review, don't merge blindly!

It might also need some documentation update.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/pdns_recursor.cc
+++ b/pdns/pdns_recursor.cc
@@ -2462,13 +2462,14 @@ void broadcastFunction(const pipefunc_t& func)
       unixDie("write to thread pipe returned wrong size or error");
     }
 
-    string* resp;
+    string* resp = nullptr;
     if(read(tps.readFromThread, &resp, sizeof(resp)) != sizeof(resp))
       unixDie("read from thread pipe returned wrong size or error");
 
     if(resp) {
 //      cerr <<"got response: " << *resp << endl;
       delete resp;
+      resp = nullptr;
     }
   }
 }
@@ -2585,7 +2586,7 @@ template<class T> T broadcastAccFunction(const boost::function<T*()>& func)
       unixDie("write to thread pipe returned wrong size or error");
     }
 
-    T* resp;
+    T* resp = nullptr;
     if(read(tps.readFromThread, &resp, sizeof(resp)) != sizeof(resp))
       unixDie("read from thread pipe returned wrong size or error");
 
@@ -2593,6 +2594,7 @@ template<class T> T broadcastAccFunction(const boost::function<T*()>& func)
       //~ cerr <<"got response: " << *resp << endl;
       ret += *resp;
       delete resp;
+      resp = nullptr;
     }
   }
   return ret;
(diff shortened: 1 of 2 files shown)


pdns/pdns_recursor.cc
```diff
@@ -2451,13 +2459,19 @@ void broadcastFunction(const pipefunc_t& func, bool skipSelf)
 
 void distributeAsyncFunction(const string& packet, const pipefunc_t& func)
 {
+  if (t_id != 0) {
```
zeha: possibly stupid q: why do all of these funcs have different t_id checks?

pdns/pdns_recursor.cc
```diff
@@ -3347,36 +3354,39 @@ static int serviceMain(int argc, char*argv[])
     s_avoidUdpSourcePorts.insert(port);
   }
 
+  /* This thread handles the web server, carbon, statistics and the control channel */
+  std::thread handlerThread(recursorThread, -1, false);
```
zeha: if -1 is a magic t_id, maybe stick this into a #define or const?

pdns/pdns_recursor.cc
rgacogne: Not stupid at all, there is some level of pain hidden here. `distributeAsyncFunction()` can only be called by the distributor thread (`t_id` 0 at the moment), `broadcastFunction()` only by the new `handler` thread (`t_id` -1), except once at startup where it's called by the first thread (`t_id` 0), and `broadcastAccFunction()` only by the new `handler` thread.

pdns/pdns_recursor.cc
rgacogne: Agreed, I'll do that. I'm also pondering adding the possibility to have more than one distributor thread so this whole logic will have to be refactored anyway (but in a different PR).

pdns/pdns_recursor.cc
```diff
@@ -2438,26 +2458,33 @@ void broadcastFunction(const pipefunc_t& func, bool skipSelf)
...
+    exit(1);
+  }
+
   unsigned int hash = hashQuestion(packet.c_str(), packet.length(), g_disthashseed);
   unsigned int target = 1 + (hash % (g_pipes.size()-1));
 
-  if(target == t_id) {
-    func();
-    return;
+  if(target == 0) {
```
zeha: s/0/s_distributorThreadID/?

rgacogne:
I just noticed a small issue with this PR. The `doStats()` function is called from the `houseKeeping()` (except if the user requested a stats dump via `USR1`), and `houseKeeping()` is invoked every 500 iterations of the main recursor thread loop. But in the new `handler` thread that has no `DNS` query to process, 500 iterations is a very long time, that in theory could be as long as 500 times the 500ms timeout of the multiplexer.
I'll move the invocation of `doStats()` out of `houseKeeping()` so it's called every `g_statisticsInterval` seconds.

rgacogne:
Rebased to fix a conflict with #6566.

PR #6571, opened 2018-05-04
auth-api: increase serial after dnssec related updates

mind04:
### Short description
The soa serial was not increased after dnssec related updates via the api

### Checklist
I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


zeha:
Well okay, but this func is also used when creating domains and then `X-PDNS-Old-Serial` makes no sense.

Personally I also totally dislike the use of headers for this; sticking the values into the output json would be a lot nicer from my pov.

mind04:
The headers are gone.
 
The "new-serial" in the json output is something for an other pull, since this is not limited to the dnssec updates.

PR #6585, opened 2018-05-08
Geoip

### LINKED ISSUE #6584: GeoIP backend - Hit on service's default network matching poisons the cache
- Program: Authoritative
 - Issue type: Bug report

### Short description

When using the GeoIP backend, as described on https://doc.powerdns.com/md/authoritative/backend-geoip/, we can configure services with subnet matching rules and a default matching rule.

I can send queries from all the configured subnets and I always get the correct answer.
If we send a query that matches no subnet, it returns the default value configured for the service, which is normal.

But then, after receiving that "default matched" query, it always answer with that value for everyone, for 20 seconds (query-cache-ttl delay I guess), even for clients that should be matched against one of the configured subnet.

### Environment
 - Operating system: Docker
 - Software version: PowerDNS Authoritative Server 4.1.1

### Configuration

Command line : --config-dir=/powerdns/ --resolver=1.2.3.4:53

/powerdns/pdns.conf :
```
expand-alias=yes
launch=geoip
geoip-zones-file=/powerdns/zones.yml

```
/powerdns/zones.yml
```
domains:
- domain: example.com
  ttl: 30
  records:
    example.com:
      - soa: ns-aaa.example.com dns.example.com 2018050100 7200 3600 1209600 300
      - ns: ns-aaa.example.com
      - ns: ns-bbb.example.com
    ns-aaa.example.com:
      - a: 10.1.0.1
    ns-bbb.example.com:
      - a: 10.2.0.1
    foo-aaa.example.com:
      - alias: aaa.another.domain.com
    foo-bbb.example.com:
      - alias: bbb.another.domain.com
  services:
    foo.example.com:
      default: 'foo-aaa.example.com'
      10.1.0.0/16: 'foo-aaa.example.com'
      10.2.0.0/16: 'foo-bbb.example.com'
```

### Expected behaviour

The service's subnet mapping should always be respected.
The query cache should be involved only after resolving the service based on the subnet.

### Actual behaviour

Query Cache is filled only when hitting the default service mapping, and the cached value is then wrongly used for the duration of query-cache-ttl.

### Usecase

It's the whole point of the GeoIP backend to vary the answered value based on the client's IP address.
One query from an unknown IP should not poison the cache.

###

I will try to look into the implementation code if I have some time.
Thanks already.

cmouse:
### Short description
Fixes #6584, and does few small code quality improvements.

Includes an optimization to make lookups faster.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

The change here is mainly that we try and avoid returning netmask of 0 for any question to avoid poisoning cache.

### DIFF
--- a/modules/geoipbackend/geoipbackend.cc
+++ b/modules/geoipbackend/geoipbackend.cc
@@ -406,7 +406,7 @@ bool GeoIPBackend::get(DNSResourceRecord &r) {
   return true;
 }
 
-string queryGeoIP(const string &ip, bool v6, GeoIPInterface::GeoIPQueryAttribute attribute, GeoIPNetmask& gl) {
+static string queryGeoIP(const string &ip, bool v6, GeoIPInterface::GeoIPQueryAttribute attribute, GeoIPNetmask& gl) {
   string ret = "unknown";
 
   for(auto const& gi: s_geoip_files) {
--- a/modules/geoipbackend/geoipbackend.cc
+++ b/modules/geoipbackend/geoipbackend.cc
@@ -35,13 +35,11 @@
 pthread_rwlock_t GeoIPBackend::s_state_lock=PTHREAD_RWLOCK_INITIALIZER;
 
 struct GeoIPDNSResourceRecord: DNSResourceRecord {
-public:
   int weight;
   bool has_weight;
 };
 
-class GeoIPDomain {
-public:
+struct GeoIPDomain {
   int id;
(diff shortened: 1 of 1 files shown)


pieterlexis:
The last two commits work for me.

Testing 16K netmasks as services, before this PR, my CPUs would be overloaded at 400 QPS, after this commit, 40000 QPS raise CPU usage to ~20%.

PR #6589, opened 2018-05-09
rec: Add lua maintenance callback

### LINKED ISSUE #5991: rec: add a periodic maintenance callback to Lua
- Program: Recursor
 - Issue type: Feature request 

### Short description

It would be helpful if the Recursor could call a maintenance function in Lua once per second, if that one is defined.

### Usecase

Currently doing periodic maintenance -- like flushing metrics, or checking a database for changes --
 in a Lua script requires checking on every request if the current time has changed.

This is cumbersome and does not guarantee that it will be actually run every second. For example, if you have a period with no requests, the maintenance function is never run.

### Description

I suggest adding a `maintenance()` callback function to the Lua API that would be called every second. 

Example usage:

```lua
local foo_metric = getMetric("foo")
local foo_count = 0

function maintenance()
    -- This would be called every second
    foo_metric:incBy(foo_count)
    foo_count = 0
    -- Other maintenance ...
end
```

Implementation notes:

* This does not need to offer guarantees that it will be called exactly every second.
* It should be called even if no requests are being handled.
* The maintenance function expects no arguments and returns no values.
* This would need to be called for every Lua state (Recursor OS thread).
* If a script requires less frequent maintenance, the Lua function can check every second if it is time to do maintenance. No need to make this more complicated by making the interval configurable.
* Maybe this needs to be explicitly enabled with an option?

chbruyand:
### Short description
This adds a lua `maintenance()` callback that will be called periodically.

Fix #5991 

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/pdns_recursor.cc
+++ b/pdns/pdns_recursor.cc
@@ -3556,7 +3556,7 @@ try
         last_carbon = g_now.tv_sec;
       }
     }
-    if(!t_id && (g_now.tv_sec - last_lua_maintenance >= luaMaintenanceInterval)) {
+    if(g_now.tv_sec - last_lua_maintenance >= luaMaintenanceInterval) {
       t_pdl->maintenance();
       last_lua_maintenance = g_now.tv_sec;
     }
--- a/pdns/pdns_recursor.cc
+++ b/pdns/pdns_recursor.cc
@@ -3556,7 +3556,7 @@ try
         last_carbon = g_now.tv_sec;
       }
     }
-    if (t_pdl != nullptr) {
+    if (worker && t_pdl != nullptr) {
       // lua-dns-script directive is present, call the maintenance callback if needed
       if(g_now.tv_sec - last_lua_maintenance >= luaMaintenanceInterval) {
         t_pdl->maintenance();
--- a/pdns/pdns_recursor.cc
+++ b/pdns/pdns_recursor.cc
@@ -3556,11 +3556,14 @@ try
         last_carbon = g_now.tv_sec;
       }
     }
-    if (worker && t_pdl != nullptr) {
+    if (t_pdl != nullptr) {
       // lua-dns-script directive is present, call the maintenance callback if needed
-      if(g_now.tv_sec - last_lua_maintenance >= luaMaintenanceInterval) {
-        t_pdl->maintenance();
-        last_lua_maintenance = g_now.tv_sec;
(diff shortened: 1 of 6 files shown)


pdns/pdns_recursor.cc
```diff
@@ -3505,6 +3506,10 @@ try
       MT->makeThread(doCarbonDump, 0);
       last_carbon = g_now.tv_sec;
     }
+    if(!t_id && (g_now.tv_sec - last_lua_maintenance >= luaMaintenanceInterval)) {
```
rgacogne: So the `maintenance()` hook will only be called from thread `0`? I think we need it to be called for all threads.

rgacogne: Since we have one Lua context per thread, I think we really need to be called for all of them. If one needs to do something in a specific thread only, it can use `getRecursorThreadId()` to know from which thread it has been invoked.

chbruyand:
Rebased to fix conflicts.

rgacogne:
Crap, I wanted to merge this PR before #6567 and I got carried away. The consequence is that the `maintenance()` hook is now called in the `handler` thread in addition to the `distributor` and regular `workers`, which probably does not make sense. Any opinion?

chbruyand:
Your right ! `maintenance()` callback should now be restricted to worker threads.

chbruyand:
Anything left to do on that PR ?

pdns/pdns_recursor.cc
```diff
@@ -3555,6 +3556,13 @@ try
         last_carbon = g_now.tv_sec;
       }
     }
+    if (worker && t_pdl != nullptr) {
```
rgacogne: `worker` is true for the distributor thread, which means we will also call the `maintenance` function from this thread which is probably useless. It doesn't matter much though.

rgacogne:
Looks good to me, perhaps @wojas might want to chime in?

phonedph1:
Version bump, now that 4.1.3 is out already?

wojas:
LGTM

It looks like maintenance will not be called if no requests arrive, but I think this is good enough.

It would be nice to be able to detect the availability of this and other new features. Right now the only way to detect it is to add a `maintenance` function, set a global flag when it's actually called and check for this flag in `gettag` (or any other callback) to decide if it should perform maintenance directly. But we can keep this out of scope for this PR.

PR #6594, opened 2018-05-10
pdnsutil: also load modules through the load-modules directive

chbruyand:
### Short description
Also load load modules defined in `pdns.conf` and that may be used by the `launch` directive.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/pdnsutil.cc
+++ b/pdns/pdnsutil.cc
@@ -58,32 +58,6 @@ string humanTime(time_t t)
   return ret;
 }
 
-static void loadModules()
-{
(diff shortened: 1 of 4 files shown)


pdns/pdnsutil.cc
```diff
@@ -58,6 +58,32 @@ string humanTime(time_t t)
   return ret;
 }
 
+static void loadModules()
```
pieterlexis: Is it possible to move this copied code to a header+cc file?

PR #6609, opened 2018-05-15
our recursor lua policy override example did not work

ahupowerdns:
### Short description
Fix an example Lua script from the documentation.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/recursordist/docs/lua-scripting/hooks.rst
+++ b/pdns/recursordist/docs/lua-scripting/hooks.rst
@@ -272,13 +272,14 @@ Example script
     myDomain = newDN("example.com")
 
     myNetblock = newNMG()
-    myNetblock:addMasks("192.0.2.0/24")
+    myNetblock:addMasks({"192.0.2.0/24"})
 
     function preresolve(dq)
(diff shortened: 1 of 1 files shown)


pdns/recursordist/docs/lua-scripting/hooks.rst
```diff
@@ -272,13 +272,14 @@ Example script
     myDomain = newDN("example.com")
 
     myNetblock = newNMG()
-    myNetblock:addMasks("192.0.2.0/24")
+    myNetblock:addMasks({"192.0.2.0/24"})
```
rgacogne: Just a nit but `myNetblock:addMask("192.0.2.0/24")` should also work.

ahupowerdns: ok - but it is somewhat of a wash if replacing it will improve things now, do you agree?

rgacogne: Sure!

PR #6614, opened 2018-05-16
auth-api: restrict creation of OPT and TSIG rrsets

### LINKED ISSUE #6441: API allows OPT RRsets to be created
- Program: Authoritative
 - Issue type: Bug report

### Short description
PowerDNS allows the creation of `OPT` RRsets via the API (and supposedly via AXFR), which then end up in the `records` table, although `OPT` is not a valid RRset type.

As a consequence, other functionality breaks down. I have tested this only with a DNSSEC-enabled domain whose AXFR used to have 173 lines; the AXFR length is now wildly varying. I got the following sequence of line numbers using `dig | wc -l`: 122, 98, 86, 168, 127, 134, 87, 98

### Environment

 - Operating system: Linux
 - Software version: 4.1.1

### Steps to reproduce
1. Create an `OPT` RRset via the API (or, presumably, by hand in the database, or by incoming AXFR). It may be required to have DNSSEC enabled.
2. Compare AXFR before and after

### Expected behaviour
When an `OPT` RRset comes in via API, it should not be created (invalid type error).
The same should happen for `OPT` in an incoming AXFR.

### Actual behaviour
`OPT` RRset is created, resulting in issues e.g. with AXFR

### Other information
Whether record creation succeeds depends on the record content. `9999` is a working example.

Error examples:
- `BIO_read failed to read all data from memory buffer` is caused by content `9`.
- `All data was not consumed` seems to appear when the content begins with `"`.

chbruyand:
### Short description
OPT and TSIG rrsets are already ignored in incoming AXFR. This restricts creation of such records through the authoritative API and fix #6441 

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/ws-auth.cc
+++ b/pdns/ws-auth.cc
@@ -306,7 +306,7 @@ static Json::object getZoneInfo(const DomainInfo& di, DNSSECKeeper *dk) {
   vector<string> masters;
   for(const auto& m : di.masters)
     masters.push_back(m.toStringWithPortExcept(53));
-  
+
   return Json::object {
     // id is the canonical lookup key, which doesn't actually match the name (in some cases)
     { "id", zoneId },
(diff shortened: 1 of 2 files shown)


zeha:
`gatherRecordsFromZone` would need the same check though, except if `ZoneParserTNG` already does it. So maybe moving this check into a new func that is called from `gatherRecords`  and `gatherRecordsFromZone` would make more sense.

zeha:
My suggestion would be to use `qtype.getName()` instead of `stringFromJson(rrset, "type")` in the error message.

chbruyand:
Oh yeah, thanks! did change that while refactoring. This was a dumb copy/paste indeed.

zeha:
I can see some code branches that could also be moved around, but lets do this in another PR.

Issue #6624, opened 2018-05-18
pdns_server crashes during startup when the first non-dynloaded module initializes


tih:
- Program: Authoritative 
 - Issue type: Bug report  

### Short description

After the big change from "theL" to "g_log", commit e6a9dde524b57bb57f1d063ef195bb1e2667c5fc, pdns_server crashes during initialization of any module that is not dynamically loaded, when the module tries to log its initialization. If all modules are dynamically loaded, there is no crash. Removing that particular log action from the module, so that it is no longer the first thing that attempts to log something, will also avoid the crash: the module may then log things later, with no ill effects.

### Environment

 - Operating system: NetBSD/amd64 8.0RC1 and NetBSD/amd64-current
 - Software version: master branch HEAD
 - Software source: Locally compiled from git clone 

### Steps to reproduce

1.  ./configure --with-modules="gpgsql" 
2.  gmake

### Expected behaviour

That pdns_server and its supporting files are properly built.

### Actual behaviour

The freshly built pdns_server dumps core when it is run by gmake to generate its sample config file.

The backtrace looks like this:

```
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x000077a0174b7cf5 in ?? () from /usr/lib/libc.so.12
(gdb) bt
#0  0x000077a0174b7cf5 in ?? () from /usr/lib/libc.so.12
#1  0x000077a0174b93bb in free () from /usr/lib/libc.so.12
#2  0x000077a017c9b667 in std::string::reserve(unsigned long) () from /usr/lib/libstdc++.so.8
#3  0x000077a017c9b6dd in std::string::append(std::string const&) () from /usr/lib/libstdc++.so.8
#4  0x0000000156938451 in Logger::operator<< (s=..., this=0x156e00fe0 <g_log>) at logger.cc:177
#5  Logger::operator<< (this=0x156e00fe0 <g_log>, 
    s=s@entry=0x156b8ad48 "[gpgsqlbackend] This is the gpgsql backend version 0.0.g23de6c095") at logger.cc:183
#6  0x0000000156b3707f in gPgSQLLoader::gPgSQLLoader (this=<optimized out>) at gpgsqlbackend.cc:184
#7  0x000000015689704f in ?? ()
#8  0x00007f7fffa83fe0 in ?? ()
#9  0x0000000156895f69 in _init ()
#10 0x0000000000000001 in ?? ()
#11 0x0000000156896f57 in ___start ()
[...]

PR #6625, opened 2018-05-18
logger: Use thread_local instead of pthread_getspecific()

rgacogne:
### Short description

It might help with #6624, but feels cleaner to me anyhow.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/logger.cc
+++ b/pdns/logger.cc
@@ -31,10 +31,8 @@ extern StatBag S;
 #include "lock.hh"
 #include "namespaces.hh"
 
-pthread_once_t Logger::s_once;
-pthread_key_t Logger::g_loggerKey;
-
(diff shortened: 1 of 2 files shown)


tih:
So I tried this patch, and it did, indeed, fix the crash I described in #6624. It also points out another quirk: once the pdns_server runs without crashing, it's interesting how the very first thing it tries to log is the initialization of each module, even before it announces its general presence at all. This is unfortunate, as the modules try to log their initialization before the desired log level has been set, and so, the log announcements of their starting is lost.

Habbie:
While losing the log messages is a pity, they mostly exist so we can spot mismatched pdns_server vs. dynamic backend module versions in bug report logs, so it's not a big problem.

PR #6627, opened 2018-05-18
rec: Drop queries truncated because they were larger than our buffer

rgacogne:
### Short description

Incoming queries larger than our buffer (1500 bytes) are truncated by the kernel, so let's not even try to parse them.

### Checklist

I have:
- [x] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/pdns_recursor.cc
+++ b/pdns/pdns_recursor.cc
@@ -2019,107 +2019,121 @@ static void handleNewUDPQuestion(int fd, FDMultiplexer::funcparam_t& var)
   char cbuf[256];
   bool firstQuery = true;
 
-  data.resize(maxIncomingQuerySize);
-  fromaddr.sin6.sin6_family=AF_INET6; // this makes sure fromaddr is big enough
-  fillMSGHdr(&msgh, &iov, cbuf, sizeof(cbuf), &data[0], data.size(), &fromaddr);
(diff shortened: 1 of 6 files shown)


pieterlexis:
in #6281, I lowered the advertised maximum to 512 (minimum allowed by EDNS), perhaps the `data.resize()` should mirror this

pieterlexis:
@ahupowerdns will know. My feeling is "something something ethernet"

rgacogne:
Alright, thanks! So does anyone see a reason why we shouldn't go with `512` from now on?

Habbie:
> Alright, thanks! So does anyone see a reason why we shouldn't go with 512 from now on?

For rec, 512 has my full support. 

(Auth, and by extension, dnsdist may need to deal with big UPDATE messages, I suppose.)

pieterlexis:
> So does anyone see a reason why we shouldn't go with 512 from now on?

For the recursor? No.

rgacogne:
- Updated to switch to & 512-byte buffer for incoming queries
- Fixed an issue where the buffer could have been invalidated between queries because of the `resize()` call, since nothing guarantees a reallocation won't happen when the size is shrank
- Rebased because it conflicted with one commit from #6518

rgacogne:
Rebased to fix a conflict.

PR #6630, opened 2018-05-20
-rdynamic is for the linker

Habbie:
### Short description
This moves `-rdynamic` from compiler to linker flags for the recursor. Ref #6511 

I DID NOT test if the FFI symbols are correctly exported with this patch.

### Checklist

I have:
- [x] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/recursordist/configure.ac
+++ b/pdns/recursordist/configure.ac
@@ -103,8 +103,7 @@ PDNS_WITH_LUA([mandatory])
 AS_IF([test "x$LUAPC" = "xluajit"], [
   # export all symbols to be able to use the Lua FFI interface
   AC_MSG_NOTICE([Adding -rdynamic to export all symbols for the Lua FFI interface])
-  CFLAGS="$CFLAGS -rdynamic"
-  CXXFLAGS="$CXXFLAGS -rdynamic"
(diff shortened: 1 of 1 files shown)


cmouse:
FWIW, I'd recommend replacing all instances of `-rdynamic` with `-export-dynamic`. This is a libtool **specific** flag (not to be confused with `-Wl,-export-dynamic`), that tells libtool that you want symbols to be dlopenable. It is more portable way of achieving the same result.

rgacogne:
> I DID NOT test if the FFI symbols are correctly exported with this patch.

I did, they are.

So I'm all for using `-export-dynamic`if it's more portable but I have no idea how to use it?

Habbie:
> So I'm all for using -export-dynamicif it's more portable but I have no idea how to use it?

I'll give this a shot.

cmouse:
It works by setting
`LDFLAGS=-export-dynamic`

Habbie:
Investigation shows that `-rdynamic` is portable to any platform on which GCC or Clang has been correctly built and configured, and the compiler will pass the right flags (if needed) to ld.

(Also, passing `-export-dynamic` in the right place would mean some ugly refactoring of configure.ac and Makefile.am).

So, let's just merge this.

PR #6633, opened 2018-05-21
reject duplicate RRsets in patchZone

Habbie:
### Short description
We frequently get users who pass two identical RRsets, each holding one record, with different content. The user might then be confused to only see the second one ending up in the zone. Erroring out early should avoid confusion.

Comments welcome. ~~Still needs a test.~~

Suggested by @hlindqvist 

### Checklist

I have:
- [x] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/ws-auth.cc
+++ b/pdns/ws-auth.cc
@@ -1642,14 +1642,15 @@ static void patchZone(HttpRequest* req, HttpResponse* resp) {
       apiCheckQNameAllowedCharacters(qname.toString());
       QType qtype;
       qtype = stringFromJson(rrset, "type");
-      if(seen.count({qname, qtype}))
-      {
-        throw ApiException("Duplicate RRset "+qname.toString()+" IN "+stringFromJson(rrset, "type"));
-      }
-      seen.insert({qname, qtype});
       if (qtype.getCode() == 0) {
         throw ApiException("RRset "+qname.toString()+" IN "+stringFromJson(rrset, "type")+": unknown type given");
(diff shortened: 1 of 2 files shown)


Habbie:
This does not pass `ERROR: test_zone_rr_update_multiple_rrsets (test_Zones.AuthZones)`

Habbie:
> This does not pass ERROR: test_zone_rr_update_multiple_rrsets (test_Zones.AuthZones)

fixed (amended), I was ignoring the qtype.

pdns/ws-auth.cc
```diff
@@ -1634,12 +1634,19 @@ static void patchZone(HttpRequest* req, HttpResponse* resp) {
...
+
     for (const auto& rrset : rrsets.array_items()) {
       string changetype = toUpper(stringFromJson(rrset, "changetype"));
       DNSName qname = apiNameToDNSName(stringFromJson(rrset, "name"));
       apiCheckQNameAllowedCharacters(qname.toString());
       QType qtype;
       qtype = stringFromJson(rrset, "type");
+      if(seen.count({qname, qtype}))
+      {
+        throw ApiException("Duplicate RRset "+qname.toString()+" IN "+stringFromJson(rrset, "type"));
```
zeha: i'd push this check down below the qtype.getCode() check and then use qtype.toString in the error message.

Habbie: pushed, thanks!

PR #6639, opened 2018-05-22
logger: Use thread_local, move config into a function-level static variable

rgacogne:
### Short description

This PR includes the commit from #6625 and additionally move the `Logger` configuration to a different struct kept inside a function-level static variable, making sure it's always properly initialized before being used.
The diff is not very large and should not be harder to maintain and use than the current version. It avoids relying on an undefined order of initialization that might prove a more serious issue in the future.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/logger.cc
+++ b/pdns/logger.cc
@@ -36,6 +36,15 @@ thread_local Logger::PerThread Logger::t_perThread;
 
 static Logger::Config& getLoggerConfig()
 {
+  /* Since the Logger can be called very early, we need to make sure
+     that the relevant parts are initialized no matter what, which is tricky
+     because we can't easily control the initialization order, especially with
+     built-in backends.
+     t_perThread is thread_local, so it will be initialized when first accessed,
+     but we need to make sure that the rest of the config is too, and making
+     it a function-level static variable achieves that, because it will be
+     initialized the first time we enter this function at the very last.
+  */
   static Logger::Config config;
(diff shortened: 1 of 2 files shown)


tih:
Tested on NetBSD, and verified to fix the crash I reported in #6624.

pdns/logger.cc
```diff
@@ -31,19 +31,25 @@ extern StatBag S;
...
 
-pthread_once_t Logger::s_once;
-pthread_key_t Logger::g_loggerKey;
-
 Logger g_log("", LOG_DAEMON);
+thread_local Logger::PerThread Logger::t_perThread;
+
+static Logger::Config& getLoggerConfig()
+{
+  static Logger::Config config;
```
zeha: "magic" code should have a comment, if you ask me (:

tih:
As previously mentioned, this fixes my NetBSD/amd64 crash, and both the auth server and the recursor work (and log) fine with these patches installed. On NetBSD/arm, however (the 32 bit arm platform, to be specific), the auth server is fine, but the recursor crashes. The backtrace looks exactly like what I saw on NetBSD/amd64 without your patch, and it happens at the first logging attempt made by recursorThread(); the "Done priming cache with root hints" one.

Might there be something the recursor isn't doing during startup, that it ought to?

PR #6643, opened 2018-05-22
auth: Use MySQL's transaction_isolation instead of tx_isolation

### LINKED ISSUE #6632: SET SESSION tx_isolation='READ-COMMITTED' deprecated in MySQL 8.0
- Program: Authoritative
- Issue type: Bug report

### Short description
Using MySQL 8.0 as my backend, the Authoritative Server always crashes upon launch when having gmysql-innodb-read-committed=yes, despite the following in my my.cnf file:

[mysqld]
log_bin
binlog_format=ROW
transaction-isolation=READ-COMMITTED

### Environment
- Operating system: CentOS 7
- Software version: Authoritative Server 4.1.2 / MySQL Server 8.0.11
- Software source: Official PowerDNS Repository for Authoritative Server / Official MySQL Repository for MySQL Server

### Steps to reproduce
1. Install PowwerDNS Authoritative Server and MySQL 8.0 from their respective repositories
2. Set binlog_format=ROW and transaction-isolation=READ-COMMITTED in /etc/my.cnf
3. Set up MySQL as the PowerDNS Backend, and set gmysql-innodb-read-committed=yes
4. Launch PowerDNS

### Expected behaviour
1. I would expect PowerDNS to just launch

### Actual behaviour
1. Receive error: Distributor caught fatal exception: Unable to launch gmysql connection: Please add '(gmysql-)innodb-read-committed=no' to your PowerDNS configuration, and reconsider yourstorage engine if it does not support transactions.:

### Other information
Looks like in modules/gmysqlbackend/smysql.cc, PowerDNS tries to set a tx_isolation variable:
```
if (d_setIsolation && (retry == 1))
    mysql_options(&d_db, MYSQL_INIT_COMMAND,"SET SESSION tx_isolation='READ-COMMITTED'");
```

However, the tx_isolation variable is deprecated since MySQL 8.0.3: https://bugs.mysql.com/bug.php?id=88227

Think if this piece of code can be set to SET SESSION transaction_isolation='READ-COMMITTED', we should be good to go.

rgacogne:
### Short description

`tx_isolation` has been deprecated since 5.7.20 and removed in 8.0.3, see https://bugs.mysql.com/bug.php?id=88227

Fixes #6632.

### Checklist

I have:
- [x] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/modules/gmysqlbackend/smysql.cc
+++ b/modules/gmysqlbackend/smysql.cc
@@ -426,8 +426,13 @@ void SMySQL::connect()
     mysql_options(&d_db, MYSQL_SET_CHARSET_NAME, MYSQL_AUTODETECT_CHARSET_NAME);
 #endif
 
-    if (d_setIsolation && (retry == 1))
+    if (d_setIsolation && (retry == 1)) {
+#if MYSQL_VERSION_ID >= 50720
+      mysql_options(&d_db, MYSQL_INIT_COMMAND,"SET SESSION transaction_isolation='READ-COMMITTED'");
+#else
       mysql_options(&d_db, MYSQL_INIT_COMMAND,"SET SESSION tx_isolation='READ-COMMITTED'");
+#endif /* MYSQL_VERSION_ID >= 50720 */
+    }
 
     mysql_options(&d_db, MYSQL_READ_DEFAULT_GROUP, d_group.c_str());
 


rgacogne:
@zeha suggested we should look into using `SET TRANSACTION ISOLATION LEVEL`  [1] instead , and a compile-time check makes no sense here anyway, we should at least check the version the server is running, not the one of the client library.

[1]: https://dev.mysql.com/doc/refman/5.5/en/set-transaction.html

Issue #5736, opened 2017-09-27
Option for the API to check if a domain exists


nlmark:
- Program: Authoritative
 - Issue type: Feature request

### Short description
If you request a domain that does not exist it returns a http status code 422. It would be nice to have a separate check/test for this I believe.

### Usecase
It makes automation easier to see if you have to remove a domain if it exists or add a domain if it doesn't exists.

### Description
A 200 http status code with a short message informing about the status would be nice.

pieterlexis:
Doesn't the /api/v1/server/localhost/zones/example.com. return a 404 if it doesn't exist?

rvdh:
@pieterlexis No, it returns 422.

zeha:
Directly going to that URL is also bad - that relies on knowing the pdns-internal applied transformation of the domain name...

pieterlexis:
so how about `/zones?exists=example.com` that returns a json:
```json
{ 
  "name": "example.com",
  "exists": false
}
```
```json
{ 
  "name": "example.com",
  "exists": true,
  "url": "/zones/example.com."
}
```

zeha:
I think we should "just" have a filter on /zones that limits the result to the one zone. Doesn't need any special "exists" thingy...

Habbie:
> Directly going to that URL is also bad - that relies on knowing the pdns-internal applied transformation of the domain name...

Me pointing this out to @nlmark and others is what led to this ticket, indeed!

zeha:
#6668 did this.

PR #6668, opened 2018-05-24
Auth API: add zone lookup by /zones?zone=example.org.

zeha:
### Short description
Implements #5736.

Someone else please type in docs.

### Checklist
I have:
- [X] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/ws-auth.cc
+++ b/pdns/ws-auth.cc
@@ -1390,7 +1390,20 @@ static void apiServerZones(HttpRequest* req, HttpResponse* resp) {
     throw HttpMethodNotAllowedException();
 
   vector<DomainInfo> domains;
-  B.getAllDomains(&domains, true); // incl. disabled
+
+  auto zone_it = req->getvars.find("zone");
+  if (zone_it != req->getvars.end()) {
+    string zone = zone_it->second;
+    apiCheckNameAllowedCharacters(zone);
+    DNSName zonename = apiNameToDNSName(zone);
+    zonename.makeUsLowerCase();
+    DomainInfo di;
+    if (B.getDomainInfo(zonename, di)) {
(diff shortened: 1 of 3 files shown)


pieterlexis:
TODO:

- [ ] Swagger update

pdns/ws-auth.cc
```diff
@@ -1390,7 +1390,20 @@ static void apiServerZones(HttpRequest* req, HttpResponse* resp) {
     throw HttpMethodNotAllowedException();
 
   vector<DomainInfo> domains;
-  B.getAllDomains(&domains, true); // incl. disabled
+
+  auto zone_it = req->getvars.find("zone");
+  if (zone_it != req->getvars.end()) {
```
chbruyand: I think you can check `req->getvars.count("zone") != 0` and then use `req->getvars["zone"]`.

zeha:
> calling /zones/example.org would make more sense IMO

Well yes, but then we're back in silly URL encoding land, where it's not clear how a zone called `foo/bar.example.org` would be encoded correctly by all possible clients.

pdns/ws-auth.cc
zeha: Yeah, but that sounded like doing the work twice?

chbruyand: That's just that something like the following looks better IMHO :
```
if (req->getvars.count("zone") != 0) {
  auto zone = req->getvars["zone"];
}
```

zeha: Pushed the style change.

zeha:
@pieterlexis did you find some time to type in the swagger changes? <3

zeha:
@pieterlexis thanks!

PR #6672, opened 2018-05-25
 dnsdist: Fix reconnection handling

rgacogne:
### Short description

The first issue was that we closed the socket toward the backend if the first connection attempt failed, but tried to reconnect without reopening it, as reported by Aleš Rygl on the mailing-list [1].
The second issue is that we would close and reconnect the backend socket on some error codes even if the error was occurring on the health check socket, which is a separate one.
This PR also makes sure we don't try to reconnect from the healthcheck and responder threads at the same time, and that we don't accidentally start the same responder thread twice.

[1]: https://mailman.powerdns.com/pipermail/dnsdist/2018-May/000425.html
### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdist-lua.cc
+++ b/pdns/dnsdist-lua.cc
@@ -345,6 +345,8 @@ void setupLuaConfig(bool client)
                         g_pools.setState(localPools);
 
 			if (ret->connected) {
+			  ret->threadStarted.test_and_set();
+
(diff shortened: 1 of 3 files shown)


rgacogne:
OP confirms that the packages built from this branch are fixing the issue for him.

PR #6675, opened 2018-05-25
Fix maybe-uninitialized warning

tcely:
### Short description

test-mtasker.cc: In function 'void mtasker_cc::doSomething(void*)':
test-mtasker.cc:20:12: warning: 'o' may be used uninitialized in this function [-Wmaybe-uninitialized]
   g_result = o;

### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/recursordist/test-mtasker.cc
+++ b/pdns/recursordist/test-mtasker.cc
@@ -15,7 +15,7 @@ static int g_result;
 static void doSomething(void* p)
 {
   MTasker<>* mt = reinterpret_cast<MTasker<>*>(p);
-  int i=12, o;
+  int i=12, o=0;
   mt->waitEvent(i, &o);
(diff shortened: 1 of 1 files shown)


rgacogne:
I'm not sure we should do this because it would prevent smarter tools like valgrind from reporting an error if it turns out that we really use it before it has been initialized.

omoerbeek:
I could not reproduce this warning with gcc 6, but looking at the test, shouldn't it be modified to only use o if an actual event was returned?
something like:

  if (mt->waitEvent(i, &o) == 1)
    g_result = o;

That should cause the test to fail if an error or timeout was returned by waitEvent which is even better than relying on valgrind imo.

rgacogne:
Agreed!

PR #6689, opened 2018-05-29
logger: Use a function-level static var for the logger object

rgacogne:
### Short description

This PR includes the commit from #6625, replaces #6639. I don't like the macro trick at all but I don't see any other way to make sure that the `Logger` object is correctly initialized before any use. 

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/pdns/logger.cc
+++ b/pdns/logger.cc
@@ -31,8 +31,22 @@ extern StatBag S;
 #include "lock.hh"
 #include "namespaces.hh"
 
-Logger g_log("", LOG_DAEMON);
 thread_local Logger::PerThread Logger::t_perThread;
+
+Logger& getLogger()
+{
+  /* Since the Logger can be called very early, we need to make sure
+     that the relevant parts are initialized no matter what, which is tricky
+     because we can't easily control the initialization order, especially with
+     built-in backends.
+     t_perThread is thread_local, so it will be initialized when first accessed,
+     but we need to make sure that the object itself is initialized, and making
+     it a function-level static variable achieves that, because it will be
+     initialized the first time we enter this function at the very last.
+  */
+  static Logger log("", LOG_DAEMON);
+  return log;
+}
 
 void Logger::log(const string &msg, Urgency u)
 {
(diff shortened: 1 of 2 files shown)


tih:
I tested this on NetBSD/amd64, and verified that the authoritative server and the recursor both work correctly with this modification. Thus, it solves the issue I reported in #6624.

I will also test on NetBSD/arm, and report further once that has been done.

tih:
The recursor still crashes in exactly the same way on NetBSD/arm, I'm afraid.

rgacogne:
I'm now wondering whether the switch to `thread_local` is causing this issue on `NetBSD/arm`..

tih:
Can I do something to test whether thread_local is working properly? Might be a platform specific bug in GCC 6.4, I suppose.

tih:
Just to see what would happen, I built the authoritative server on the Pi, from the same code base, including the gpgsql module. It's running (and logging) just as it should, happily letting me create and modify zones, which it then serves correctly. Securing zones on it works, too.

So the crash in the recursor, the first time recursorThread() tries to log something, *may* of course be a GCC bug in the thread_local handling, but it doesn't seem all that likely to me.

For reference, the backtrace looks like this:

Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x7a5598c4 in std::string::append(std::string const&) ()
   from /usr/lib/libstdc++.so.8
[Current thread is 1 (process 2)]
(gdb) bt
#0  0x7a5598c4 in std::string::append(std::string const&) ()
   from /usr/lib/libstdc++.so.8
#1  0x088d8b74 in Logger::operator<< (s=..., this=0x8da7f6c <getLogger()::log>)
    at logger.cc:166
#2  Logger::operator<< (this=0x8da7f6c <getLogger()::log>, s=<optimized out>)
    at logger.cc:172
#3  0x08971ddc in recursorThread (n=<optimized out>, worker=<optimized out>)
    at pdns_recursor.cc:3436

PR #6696, opened 2018-05-30
(WIP) Newly observed domain support

neilcook:
### Short description
This PR implements newly observed domain support.  It uses the magic of hyperloglog to achieve this with minimal memory, cpu and disk overhead.

Upon starting the recursor, the NOD DB is initialised - this reads in all the previously saved hyperloglog cache files and merges then into a "master" HLL which is used for all checks. There is also a "current" HLL that isn't used for checks (although it is merged into master on startup if a snapshot exists on disk), rather it contains all the lookups for the current period (24 hours), and is snapshotted to disk every 10 minutes. Once per day, the current HLL is archived and a new one is started. Once the number of cache files exceeds a configurable number (default 365 i.e. one year), the oldest are pruned.

Whenever the recursor gets a cache miss, it checks for a newly observed domain in the NOD DB. If it finds one, it logs and/or sends a DNS query to powerdns. (TODO - the log should be a separate log, and the DNS query isn't implemented yet).

If the recursor is shutdown using rec_control, the current HLL is snapshotted to disk before exit.

The upshot of all of the above is that after few hours, and with increasing fidelity, the newly observed domain log should be pretty accurate. It also survives restarts (the most that can be lost is 10 minutes if the recursor is killed or crashes).

It remains to be seen if the Hyperloglog settings I've used will provide a good balance between speed and effectiveness.  Currently we're using 10 register bits with 1024 registers, which gives an error rate of around 3%. It's not the end of the world if we get a new domain false-positive.

### Checklist

I have:
- [ x] read the [CONTRIBUTING.md](https://github.com/PowerDNS/pdns/blob/master/CONTRIBUTING.md) document
- [ x] compiled this code
- [ x] tested this code
- [ x] documented the code
- [ x] added or modified unit test(s)


pieterlexis:
> If the recursor is shutdown using rec_control, the current HLL is snapshotted to disk before exit.

Keep in mind that we do not shut it down like this via the systemd unit files

neilcook:
I will almost certainly rewrite the hyperloglog function at some point to use a bigger than 32 bit hash.

neilcook:
This needs a rethink - HLLs are not appropriate for this task. I'll look at other algorithms such as bloom filters, and scale down the ambitions for how long we track domains for.

PR #6702, opened 2018-06-01
add "enable-notification-warning" option to make logging of warnings …

klaus3000:
…configureable

### Short description
By default PowerDNS will log a warning if all NOTIFYs were suppressed due to
`only-notify` and there were no explict ALSO-NOTIFYs. This warning can be suppressed by
setting this option to `no`

### Checklist

I have:
- [x ] compiled this code
- [ ] tested this code
- [x ] included documentation (including possible behaviour changes)
- [ ] documented the code

### DIFF
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -554,16 +554,16 @@ Enables EDNS subnet processing, for backends that support it.
 
 Enable globally the LUA records feature
 
-.. _setting-enable-notification-warning:
+.. _setting-enable-no-notification-targets-warning:
 
-``enable-notification-warning``
+``enable-no-notification-targets-warning``
 --------------------------
 
 -  Boolean
 -  Default: yes
 
-By default PowerDNS will log a warning if all NOTIFYs were suppressed due to
-`only-notify` and there were no explict ALSO-NOTIFYs. This warning can be suppressed by
+By default PowerDNS will log a warning if all outgoing NOTIFYs were suppressed due to
+`only-notify` and there were no explicit ALSO-NOTIFYs. This warning can be suppressed by
 setting this option to `no`.
 
 .. _setting-entropy-source:
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -554,6 +554,18 @@ Enables EDNS subnet processing, for backends that support it.
 
 Enable globally the LUA records feature
 
+.. _setting-enable-notification-warning:
+
+``enable-notification-warning``
+--------------------------
+
+-  Boolean
+-  Default: yes
+
+By default PowerDNS will log a warning if all NOTIFYs were suppressed due to
+`only-notify` and there were no explict ALSO-NOTIFYs. This warning can be suppressed by
+setting this option to `no`.
+
 .. _setting-entropy-source:
 
 ``entropy-source``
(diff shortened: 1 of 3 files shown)


docs/settings.rst
```diff
@@ -554,6 +554,18 @@ Enables EDNS subnet processing, for backends that support it.
 
 Enable globally the LUA records feature
 
+.. _setting-enable-notification-warning:
+
+``enable-notification-warning``
```
Habbie: This name is a bit broad; how about 'enable-no-notification-targets-warning'?

docs/settings.rst
```diff
@@ -554,6 +554,18 @@ Enables EDNS subnet processing, for backends that support it.
...
+--------------------------
+
+-  Boolean
+-  Default: yes
+
+By default PowerDNS will log a warning if all NOTIFYs were suppressed due to
+`only-notify` and there were no explict ALSO-NOTIFYs. This warning can be suppressed by
```
Habbie: s/explict/explicit/

klaus3000:
> somebody could have a setup without also-notify, without only-notify, and a broken zone with no NS records. This setting (when set to no) would also hide the warning in that case, and I'm wondering if that is what we want.

To detect broken zones this is the wrong place. There could be much wrong with a zone and from experience this happens quite often (PDNS luckily transfers a zone with NS incoming and outgoing). So, if you want to detect broken zones, do it somewhere else. Eg outofband with recitify tools. Further the log warning would not show the real problem (missing NS record). Actually I am against this warning in the first place - e.g. Bind also does not log warning if you configure "notify no" or "notify explicit".

docs/settings.rst
klaus3000: changed

docs/settings.rst
klaus3000: fixed

Habbie:
So what's the use case? Why is such a domain set to MASTER?

klaus3000:
> So what's the use case? Why is such a domain set to MASTER?

Our domains are not set to master. We are a secondary service (hence all zones have type=SLAVE) and customers can refetch their zones, ie after DNSSEC signing on our side (hence we have slave-renotify=yes).

For those zones were the customers want to refetch their zones we set ALSO-NOTIFY domain metadata. Further, to avoid false NOTIFYs we have set 'only-notify=' (to imitate Bind's notify=explicit).

Hence, for every zone which does not have ALSO-NOTIFYs set, we receive this warning,

mind04:
So the real problem is a missing SLAVE-RENOTIFY, per domain, metadata option. 
I prefer to add the metadata option instead of hiding the warning with a config option.

klaus3000:
I do not like this patch - maybe you have a better idea how to avoid these warnings.

klaus3000:
As mentioned before, IMO the warning is argueable. If someone configures PDNS to not send NOTIFYs then PDNS logs a warning. IMO it is not PDNS which should take care of all provisioning errors.

Regarding metadata. Where to start and where to stop? So also add ONLY-NOTIFY domain metadata?
I think then metadata needs an overhaul. Currently, if I understand it correctly, every metadata lookup is a dedicated uncached DB-lookup. So I am a bit afraid that adding plenty of metadata adds plenty of DB lookups and puts additional load on the DB. It should be possible to always load all metadata of a zone and cache it for the current tasks, ie for request handling, for a slave_check run, ...

So, to prevent additional DB lookups due to additional metadata lookups I would close the PR and, as before, just delete the log statement before building PDNS.

Habbie:
metadata lookups are cached. I'm closing this PR as we can't really agree on how to solve this niche situation.

PR #6703, opened 2018-06-02
dnsdist: Support DNSAction.Allow in DynBlocks for testing purposes

lochiiconnectivity:
### Short description
Supports DNSAction.Allow in DynBlocks for testing purposes, based on discussion in the IRC channel on 31/05. Also makes documentation more verbose.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [X] documented the code
- [] added or modified unit test(s)

### DIFF
--- a/pdns/dnsdist-lua.cc
+++ b/pdns/dnsdist-lua.cc
@@ -936,12 +936,23 @@ void setupLuaConfig(bool client)
 
   g_lua.writeFunction("setDynBlocksAction", [](DNSAction::Action action) {
       if (!g_configurationDone) {
-        if (action == DNSAction::Action::Drop || action == DNSAction::Action::Refused || action == DNSAction::Action::Truncate) {
-          g_dynBlockAction = action;
-        }
-        else {
-          errlog("Dynamic blocks action can only be Drop, Refused or Truncate!");
-          g_outputBuffer="Dynamic blocks action can only be Drop, Refused or Truncate!\n";
+        switch(action) {
+          case DNSAction::Action::Allow: 
+            g_dynBlockAction = action;
+            break;
+          case DNSAction::Action::Drop:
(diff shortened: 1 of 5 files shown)


pdns/dnsdist-lua.cc
```diff
@@ -936,12 +936,23 @@ void setupLuaConfig(bool client)
...
+          case DNSAction::Action::Allow: 
+            g_dynBlockAction = action;
+            break;
+          case DNSAction::Action::Drop:
+            g_dynBlockAction = action;
+            break;
+          case DNSAction::Action::Refused:
+            g_dynBlockAction = action;
+            break;
+          case DNSAction::Action::Truncate:
```
rgacogne: I'm not sure the switch is needed there, but perhaps we could at least collapse it into two single blocks since all blocks are the same except for the default, so something like:
```
case DNSAction::Action::Allow:
case DNSAction::Action::Drop:
case DNSAction::Action::Refused:
case DNSAction::Action::Truncate:
  g_dynBlockAction = action;
  break;
default:
...
```

lochiiconnectivity: Ack, will collapse the switch statement. 

With regards to the rule processing, how about permitting the use of DNSAction::Action::None here?

rgacogne: Right, it looks like `DNSAction::Action::None` would work for `setDynBlocksAction()`, although it wouldn't for the optional action parameter of `addDynBlocks()` since we already use `None` there to indicate that the default action should be used. I don't think that's an issue, but we would need to make that clear in the documentation.

rgacogne:
I'm closing this pull request because it looks like it has been superseded by #6776, please correct me if I'm wrong and thank you again!

PR #6704, opened 2018-06-04
logger: Use a function-level static var for the logger object

rgacogne:
### Short description

This is the same commit than in #6689, ie making the `Logger` object a function-level static variable to make sure it's properly initialized. It fixes the crash reported in #6624 without adding a new one like the use of `thread_local` proposed in #6689 and #6625.
This should restore the behavior we had before the move to `g_log`, but is not perfect: @tih reports that the recursor still crashes a bit later on `NetBSD/arm32`, apparently because of an issue regarding the initialization of non-`POD` `unique_ptr`.
I still think using `thread_local` instead of `pthread_get_specific()` makes sense, so I'd be more interested in merging #6689 than this PR but I'm opening it so we can discuss that.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/logger.cc
+++ b/pdns/logger.cc
@@ -34,7 +34,20 @@ extern StatBag S;
 pthread_once_t Logger::s_once;
 pthread_key_t Logger::g_loggerKey;
 
-Logger g_log("", LOG_DAEMON);
+Logger& getLogger()
+{
+  /* Since the Logger can be called very early, we need to make sure
+     that the relevant parts are initialized no matter what, which is tricky
+     because we can't easily control the initialization order, especially with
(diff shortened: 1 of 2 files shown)


tih:
For now, I'll be giving up on PowerDNS on NetBSD/arm32 - this looks like a tool chain bug, and anyway it's too painful to test on my Raspberry Pi 1.

For the record, the problem was observed on NetBSD/evbarm-current 8.99.17, with gcc 6.4.0 and binutils 2.27.

rgacogne:
Closing since #6689 has been merged.

PR #6698, opened 2018-05-31
rec: Refactor Protobuf options, add query/response selection

rgacogne:
### Short description

- Turn the protobuf options into a table because there are already too many of them
- Split the masks applied to the initiator to a separate `setProtobufMasks` directive
- Add the possibility to log only queries, or only responses
- Add the possibility to select queries (FFI only) and responses for export from the Lua hooks
- Add regression tests for the protobuf features

Closes #6361.

### Checklist

I have:
- [x] read the [CONTRIBUTING.md](https://github.com/PowerDNS/pdns/blob/master/CONTRIBUTING.md) document
- [x] compiled this code
- [x] tested this code
- [x] included documentation (including possible behaviour changes)
- [x] documented the code
- [x] added or modified regression test(s)
- [ ] added or modified unit test(s)

PR #6708, opened 2018-06-05
rec: Support exporting more record types via protobuf

rgacogne:
### Short description

Several users have reported the need to export the content of of more record types found in responses. This pull request adds support for optionally exporting the content of `MX`, `NS`, `PTR`,  `SPF`, `SRV` and `TXT` record types, in addition to the existing `A`, `AAAA` and `CNAME`. The implementation is limited to those types because it special cases them.
An other option would be to implement all `xfr*()` methods to be able to get the "wire version" of all types.  If we decided to go that road, I think it might make sense to extract the `xfr*()`methods from `DNSPacketWriter` into a `DNSRecordWriter` implementing all `xfr*()` methods, that would be called by `DNSPacketWriter` but could also be used from the protobuf logger.

This PR is based on #6698 and will need a rebase after it has been merged.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/pdns/dnsrecords.hh
+++ b/pdns/dnsrecords.hh
@@ -211,6 +211,10 @@ class SPFRecordContent : public DNSRecordContent
 {
 public:
   includeboilerplate(SPF)
+  const std::string& getText() const
+  {
+    return d_text;
+  }
 
 private:
   string d_text;
@@ -240,6 +244,7 @@ class PTRRecordContent : public DNSRecordContent
 public:
   includeboilerplate(PTR)
   explicit PTRRecordContent(const DNSName& content) : d_content(content){}
+  const DNSName& getContent() const { return d_content; }
 private:
   DNSName d_content;
 };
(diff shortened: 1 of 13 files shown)


chbruyand:
I'd vote in favor of the `DNSRecordWriter`.

As far as I understand, your current implementation is not really currently limited to exporting the types listed in your comment as you check know types against `QType::names`.

rgacogne:
> As far as I understand, your current implementation is not really currently limited to exporting the types listed in your comment as you check known types against QType::names.

It is, because `RecProtoBufMessage::addRR()` only knows how to deal with those types.

rgacogne:
> I'd vote in favor of the `DNSRecordWriter`.

It would be nicer, yes, but would also be quite more work. My current feeling is to merge this pull request and see if the demand for exporting other types really exists before considering the `DNSRecordWriter` option.

rgacogne:
Rebased since #6698 has been merged in the meantime.

PR #6711, opened 2018-06-05
dnsdist: default set "Connection: close" header for web requests

### LINKED ISSUE #6532: dnsdist: webserver should set Connection: close in standard headers
- Program: dnsdist
 - Issue type: Bug report

### Short description
The dnsdist webserver seems to close connections after each request without signaling to the client that it's going to do so with a `Connection: close` header. This introduces a race condition where an HTTP/1.1 client that implements connection reuse may try to send a second request over the connection if it doesn't receive the TCP FIN in time, and this second request will get hung up on.

Specifying `{["Connection"]="close"}` in `webserver()`'s custom headers seems to work around the problem.

### Environment
 - Operating system: Debian buster (aka "testing" as of right now)
 - Software version: 1.3.0
 - Software source: Debian repo

chbruyand:
### Short description
Default add the `Connection: close` header to dnsdist http responses to indicate that the connection will be closed after completion of the response.

Fix #6532 
### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdist-web.cc
+++ b/pdns/dnsdist-web.cc
@@ -194,7 +194,6 @@ static void addSecurityHeaders(YaHTTP::Response& resp, const boost::optional<std
     { "X-Permitted-Cross-Domain-Policies", "none" },
     { "X-XSS-Protection", "1; mode=block" },
     { "Content-Security-Policy", "default-src 'self'; style-src 'self' 'unsafe-inline'" },
-    { "Connection", "close" },
   };
 
   for (const auto& h : headers) {
@@ -275,6 +274,8 @@ static void connectionThread(int sock, ComboAddress remote, string password, str
 
     addCustomHeaders(resp, customHeaders);
     addSecurityHeaders(resp, customHeaders);
+    /* indicate that the connection will be closed after completion of the response */
+    resp.headers["Connection"] = "close";
 
     /* no need to send back the API key if any */
     resp.headers.erase("X-API-Key");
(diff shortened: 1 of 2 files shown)


pdns/dnsdist-web.cc
```diff
@@ -194,6 +194,7 @@ static void addSecurityHeaders(YaHTTP::Response& resp, const boost::optional<std
     { "X-Permitted-Cross-Domain-Policies", "none" },
     { "X-XSS-Protection", "1; mode=block" },
     { "Content-Security-Policy", "default-src 'self'; style-src 'self' 'unsafe-inline'" },
+    { "Connection", "close" },
```
rgacogne: The name of the function (`addSecurityHeaders`) might not be relevant anymore after that change.

pdns/dnsdistdist/docs/guides/webserver.rst
```diff
@@ -21,6 +21,7 @@ By default, our web server sends some security-related headers::
    X-Permitted-Cross-Domain-Policies: none
    X-XSS-Protection: 1; mode=block
    Content-Security-Policy: default-src 'self'; style-src 'self' 'unsafe-inline'
+   Connection: close
```
rgacogne: I don't think `Connection: close` belongs here, it's not a security-related header at all.

chbruyand:
Sorry, didn't pay attention to the name of the function. I did move setting the header where it cannot be overriden as we will always close the connection.

PR #6713, opened 2018-06-05
rec: add bogus ring to make it more easy to detect high profile domains with broken dnssec

mind04:
### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


rgacogne:
The packet cache unit tests do not compile anymore :'(

pdns/recpacketcache.cc
```diff
@@ -66,6 +67,12 @@ bool RecursorPacketCache::checkResponseMatches(std::pair<packetCache_t::index<Ha
         i = i + labellen;
       }
 
+      if(iter->d_vstate == Bogus) {
+        if(t_bogusremotes)
+          t_bogusremotes->push_back(source);
+        if(t_bogusqueryring)
+          t_bogusqueryring->push_back(make_pair(qname, qtype));
+      }
```
rgacogne: It feels a bit wrong to have this code here, IMHO the packet cache should not have any knowledge of the rings or the requestor's source address. Perhaps this logic could be moved in `doProcessUDPQuestion()`?

pdns/recpacketcache.cc
```diff
@@ -174,6 +178,7 @@ void RecursorPacketCache::insertResponsePacket(unsigned int tag, uint32_t qhash,
     e.d_ttd = now+ttl;
     e.d_creation = now;
     e.d_tag = tag;
+    e.d_vstate = valState;
```
rgacogne: We should likely also update the validation state when we replace an existing entry, just above.

Issue #6709, opened 2018-06-05
dnsdist 1.3.x local console no longer seems to work without using setKey()


aj-gh:
- Program: dnsdist
 - Issue type: Bug report

### Short description
Local console commands on dnsdist 1.2.x worked without having to use setKey() - in 1.3.x it seems that setKey() is required, otherwise the console client does not work at all and closes the command connection. dnsdist itself logs "Could not decrypt message".
Improvement of console output on mismatched keys is requested in #6683.

### Environment
 - Operating system: Ubuntu 16.04
 - Software version: dnsdist 1.3.0-1pdns.xenial
 - Software source: PowerDNS repository

### Steps to reproduce
1. Create dnsdist.conf with controlSocket("127.0.0.1") but without setKey()
2. Start dnsdist and attempt to connect to the console with -c
3. Compare to behavior of 1.2.x with the same config, then 1.3.x with setKey() configured

### Expected behaviour
Local console access working even without using setKey()

### Actual behaviour
Local console access fails in 1.3.x without setKey() but works in 1.2.x

rgacogne:
Hi, thanks for reporting this! Using the console without `setKey()` doesn't work when `dnsdist` has been compiled with `libsodium` support enabled, and that hasn't changed between 1.2.x and 1.3.0. We did however change `libsodium`'s default detection from `no` to `auto`, meaning that it's not automatically enabled if the library is present.
I'm not 100% sure on how to fix this, we could make the console work without any defined key regardless of whether `libsodium` support is enabled, but this would have security implications, even on localhost where it would allow any user to connect to the console.

hlindqvist:
Isn't it a good thing if it always requires a key to allow console connections? (With appropriate error messages, of course)

aj-gh:
@rgacogne, well, it seems it did work here ;)
```
# dnsdist --version
dnsdist 1.2.0 (Lua 5.1.4)
Enabled features: dnscrypt libsodium protobuf re2 systemd
# grep -i setkey /etc/dnsdist/dnsdist.conf
# dnsdist -c
> showVersion();
dnsdist 1.2.0
>
```

@hlindqvist, I'm also fine with it requiring a key locally but then I'd propose a few documentation changes because at the moment the docs only strongly suggest to enable encryption (setKey) when using remote connections, not that everything including local would fail when not having used setKey().
I agree that enforcing a key makes sense when having untrusted local users.

rgacogne:
Fixed by #6715.

PR #6715, opened 2018-06-05
dnsdist: Refuse console connection without a proper key set

### LINKED ISSUE #6683: Please have dnsdist -c verify it can talk to the server
- Program: dnsdist 
 - Issue type: Feature request 

### Short description

dnsdist config - in non-default place - with:
```
setKey("Z8JaN17Ba8zkNBx8fa11VWJYm5RwHPtf+Esp8G2KwtE=")
addLocal('127.0.0.1:5400')
```

Running `dnsdist -c` gives:
```
Unable to read configuration from '/usr/local/etc/dnsdist.conf'
> newRemoteLogger()
Connection closed by the server.
```

On the server:
```
Accepting control connections on 127.0.0.1:5199
> Got control connection from 127.0.0.1:49564
Got an exception in client connection from 127.0.0.1:49564: Could not decrypt message
```

### Usecase

Mostly me being stupid and dnsdist could be nicer to me. Sending a dummy message to the server should probably suffice to check for the mismatched key.

rgacogne:
### Short description

Fixes #6683 and explicitly prevents the situation from #6709, which was working in some cases because of a very brittle side-effect.
This PR should add a few words to the documentation to clarify that the use of encryption is also strongly recommended for local connections.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/pdns/dnsdist-console.cc
+++ b/pdns/dnsdist-console.cc
@@ -86,7 +86,6 @@ static bool sendMessageToServer(int fd, const std::string& line, SodiumNonce& re
   putMsgLen32(fd, static_cast<uint32_t>(msgLen));
 
   if (!msg.empty()) {
-    cerr << "sending message of size " << msgLen << endl;
     writen2(fd, msg);
   }
 
--- a/pdns/dnsdist-console.cc
+++ b/pdns/dnsdist-console.cc
@@ -79,7 +79,7 @@ static bool sendMessageToServer(int fd, const std::string& line, SodiumNonce& re
   string msg = sodEncryptSym(line, g_consoleKey, writingNonce);
   const auto msgLen = msg.length();
   if (msgLen > std::numeric_limits<uint32_t>::max()) {
-    cout << "Encrypted essage is too long to be sent to the server, "<< std::to_string(msgLen) << " > " << std::numeric_limits<uint32_t>::max() << endl;
+    cout << "Encrypted message is too long to be sent to the server, "<< std::to_string(msgLen) << " > " << std::numeric_limits<uint32_t>::max() << endl;
     return true;
   }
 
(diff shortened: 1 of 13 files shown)


pdns/dnsdist-console.cc
```diff
@@ -73,10 +74,55 @@ static string historyFile(const bool &ignoreHOME = false)
   return ret;
 }
 
+static bool sendMessageToServer(int fd, const std::string& line, SodiumNonce& readingNonce, SodiumNonce& writingNonce, const bool outputEmptyLine)
+{
+  string msg = sodEncryptSym(line, g_consoleKey, writingNonce);
+  const auto msgLen = msg.length();
+  if (msgLen > std::numeric_limits<uint32_t>::max()) {
+    cout << "Encrypted essage is too long to be sent to the server, "<< std::to_string(msgLen) << " > " << std::numeric_limits<uint32_t>::max() << endl;
```
chbruyand: *message ?

pdns/dnsdistdist/docs/guides/console.rst
```diff
@@ -11,7 +11,7 @@ The console can be enabled with :func:`controlSocket`:
 
   controlSocket('192.0.2.53:5199')
 
-Exposing the console to the network without encryption enabled is not recommended. To enable encryption, first generate a key with :func:`makeKey`::
+Enabling the console without encryption enabled is not recommended. To enable encryption, first generate a key with :func:`makeKey`::
```
chbruyand: Maybe adding a word about the need of libsodium for encryption to actually happen would be nice ?

rgacogne: We do so a few lines below but you are right in that it would be better to do it here, I'll move it up. Thanks!

Issue #4556, opened 2016-10-11
No SOA on ALIAS record types - pdn recursor


gitshaw:
Hopefully someone can help here - not sure if this is expected or not.

Im using pdns auth and rec. 

My pdns server is authoritative for the domain foo.mydomain.net and i am using the ALIAS record type to point to an external hostname on akamai. The problem im having is when i request an IPv6 record no SOA value is returned. 

```
dig foo.mydomain.net @localhost AAAA

; <<>> DiG 9.9.5-3ubuntu0.9-Ubuntu <<>> foo.mydomain.net @localhost AAAA
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 5163
;; flags: qr aa rd ra; QUERY: 1, ANSWER: 0, AUTHORITY: 0, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 1680
;; QUESTION SECTION:
;foo.mydomain.net.      IN      AAAA

;; Query time: 31 msec
;; SERVER: 127.0.0.1#53(127.0.0.1)
;; WHEN: Tue Oct 11 06:47:06 UTC 2016
;; MSG SIZE  rcvd: 53`
```

however you can see in the logs that the akamai SOA value is returned its just not passed through to the client requesting.

```
Oct 11 06:41:02 09d8cbc01d94 pdns_recursor[1981]: 2 [1/1] question for 'XXX.com.akadns.net|AAAA' from 172.17.0.3
Oct 11 06:41:02 09d8cbc01d94 pdns_recursor[1981]: [1] XXX.com.akadns.net: Wants NO DNSSEC processing in query for AAAA
Oct 11 06:41:02 09d8cbc01d94 pdns_recursor[1981]: [1] XXX.com.akadns.net: Looking for CNAME cache hit of 'XXX.com.akadns.net|CNAME'
Oct 11 06:41:02 09d8cbc01d94 pdns_recursor[1981]: [1] XXX.com.akadns.net: No CNAME cache hit of 'XXX.com.akadns.net|CNAME' found
Oct 11 06:41:02 09d8cbc01d94 pdns_recursor[1981]: [1] XXX.com.akadns.net: No cache hit for 'XXX.com.akadns.net|AAAA', trying to find an appropriate NS record
[...]

PR #6727, opened 2018-06-11
ALIAS: Respond SERVFAIL on non-NOERRORs from resolver

### LINKED ISSUE #6666: ALIAS: behaviour when resolver returns anything but NOERROR
- Program: Authoritative
 - Issue type: Bug report

### Short description
At the moment, in 4.1.2 and master, the Authoritative Server will happily forward whatever information it gets in response to an ALIAS lookup. For instance, when the ALIAS target does not exist, the response to the client will be NXDOMAIN. This could effectively "blank out" a name or domain, especially if the ALIAS is at the apex of a zone.

A comment in the code for ALIAS expansion over TCP leads me to believe (and I agree with this) that we should answer with SERVFAIL in this case. So the client can retry another authoritative server.

I have a branch [here](https://github.com/pieterlexis/pdns/commits/alias-servfail-on-nxd), but the tests are failing because the SERVFAIL also (rightfully) kills the AXFR with an expanded ALIAS. I think I'll make a separate zone for this test to test this behaviour. This ticket is created so we don't forget.

### Steps to reproduce
1. create ALIAS to non-existent domain
2. Query for A record of the ALIAS name
3. Get NXDOMAIN for that name

### Expected behaviour
SERVFAIL

### Actual behaviour
NXDOMAIN

pieterlexis:
### Short description
In #6659 is was discovered that the Authoritative Server would send NXDOMAIN if the ALIAS target would yield NXDOMAIN (actually, it would pass the RCODE from the resolver). Leading to potential issues of blanking out a domain if the ALIAS was on the apex.

This PR fixes these issues by setting the rcode to SERVFAIL if the recursor's answer is not NOERROR. 

On timeouts from the resolver, we will send SERVFAIL as well over TCP. But due to the design of the dnsproxy there is no response over UDP, it will take some work to be able to detect this, we'll have to read our conntrack table and return SERVFAIL if some time has expired. All in a loop separate from the recvfrom() loop we already have in `DNSProxy::mainloop`.

Closes #6666

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code


pieterlexis:
A yes, I know remember why this was not PR'd before. Because we SERVFAIL the AXFR when the ALIAS can not the expanded, the tests fail.

gibson042:
> setting the rcode to SERVFAIL if the recursor's answer is not NOERROR

Isn't that overly blunt? NXDOMAIN in particular seems like it should be translated to NODATA (i.e., RCODE=NOERROR + ANCOUNT=0).

And there's probably already an ALIAS bug with respect to NODATA—negative responses must include the SOA of the containing zone.

pieterlexis:
> Isn't that overly blunt? NXDOMAIN in particular seems like it should be translated to NODATA (i.e., RCODE=NOERROR + ANCOUNT=0).

This is something that could be argued. NXDOMAIN of the ALIAS target in my eyes means that the zone was misconfigured as the ALIAS target must exist. But I am open to be convinced differently.

> And there's probably already an ALIAS bug with respect to NODATA—negative responses must include the SOA of the containing zone.

From a quick test, this indeed seems to be the case. We'll fix that. Thanks!

pieterlexis:
@Habbie the tests pass, could you review?

pdns/dnsproxy.cc
```diff
@@ -100,7 +100,7 @@ bool DNSProxy::completePacket(DNSPacket *r, const DNSName& target,const DNSName&
       ret2 = stubDoResolve(target, QType::AAAA, ips);
 
     if(ret1 != RCode::NoError || ret2 != RCode::NoError) {
-      g_log<<Logger::Error<<"Error resolving for "<<aname<<" ALIAS "<<target;
+      g_log<<Logger::Error<<"Error resolving for "<<aname<<" ALIAS "<<target<<" over TCP";
```
Habbie: I didn't think we actually did a TCP query ourselves, here, so probably needs better wording. Given our entirely distinct handling for whether the client query came via UDP or TCP, logging the difference does seem sensible.

pdns/dnsproxy.cc
```diff
@@ -245,26 +245,31 @@ void DNSProxy::mainloop(void)
         /* Set up iov and msgh structures. */
         memset(&msgh, 0, sizeof(struct msghdr));
         string reply; // needs to be alive at time of sendmsg!
-        if(i->second.complete) {
-
+        if(i->second.complete) { // TODO in the world of non-recursive backends, isn't this always true?
```
Habbie: Indeed! Nothing in our codebase calls `DNSProxy::recurseFor` anymore.

regression-tests.auth-py/test_ALIAS.py
```diff
@@ -0,0 +1,201 @@
...
+    def testNxDomain(self):
+        query = dns.message.make_query('nxd.example.org', 'A')
+        res = self.sendUDPQuery(query)
+        self.assertRcodeEqual(res, dns.rcode.SERVFAIL)
+
+        query = dns.message.make_query('nxd.example.org', 'AAAA')
+        res = self.sendUDPQuery(query)
+        self.assertRcodeEqual(res, dns.rcode.SERVFAIL)
+
+        # TODO this should actually return SOA + NS?
```
Habbie: No - ANY includes A/AAAA and thus ALIAS and thus I think SERVFAIL is fine.

Habbie:
> And there's probably already an ALIAS bug with respect to NODATA—negative responses must include the SOA of the containing zone.

The PR does not appear to fix this, but this is tracked in #4556

regression-tests.auth-py/test_ALIAS.py
pieterlexis: ANY is not ALL though....

Habbie: > ANY is not ALL though....

Yes it is. Cloudflare can go ** a ** :)

PR #6728, opened 2018-06-11
Newly Observed Domain Support

neilcook:
### Short description
This PR provides support for detecting newly observed domains. It's based on a probabilistic data structure called a Stable Bloom Filter (basically a bloom filter that expires random entries every time you add a new entry).

It still needs work:
- Documentation
- I'd really like to have it run on some sample of "real" traffic to see how it performs (functionally, speed, accuracy etc.)
- Currently the "log" option logs to normal pdns log - this should be to a different log
- Packaging - the recursor packaging scripts need to create the cache directory (PREFIX/var/lib/pdns-recursor/nod)

This PR does take into account the comments I received from lieter. Although it does still contain boost::filesystem.

I've added unit tests for the underlying functionality, but no regression tests for the recursor as yet.

By default this functionality is DISABLED. It needs a configure option to enable.

### Checklist
I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code


pdns/nod.cc
```diff
@@ -0,0 +1,211 @@
...
+  }
+  return ret;
+}
+
+void NODDB::addDomain(const DNSName& dname)
+{
+  // The only time this should block is when snapshotting from the
+  // housekeeping thread
+  std::lock_guard<std::mutex> lock(NODDB::d_sbf_mutex);
+  std::string native_domain = dname.toDNSStringLC();
```
rgacogne: the call to `DNSName::toDNSStringLC()` could be done outside of the critical section.

pdns/nod.cc
```diff
@@ -0,0 +1,211 @@
...
+  return isNewDomain(dname);
+}
+
+bool NODDB::isNewDomain(const DNSName& dname)
+{
+  // The only time this should block is when snapshotting from the
+  // housekeeping thread
+  std::lock_guard<std::mutex> lock(NODDB::d_sbf_mutex);
+  // the result is always the inverse of what is returned by the SBF
+  return !d_sbf.testAndAdd(dname.toDNSStringLC());
```
rgacogne: the call to `DNSName::toDNSStringLC()` could be done outside of the critical section.

pdns/pdns_recursor.cc
```diff
@@ -861,6 +873,42 @@ static bool checkOutgoingProtobufExport(LocalStateHolder<LuaConfigItems>& luacon
...
+    if (t_nodDBp && t_nodDBp->isNewDomain(dname)) {
+      if (g_nodLog) {
+        // This should probably log to a dedicated log file
+        g_log<<Logger::Notice<<"Newly observed domain nod="<<dname.toLogString()<<endl;
+      }
+      if (!(g_nodLookupDomain.compare("") == 0)) {
+        // Send a DNS A query to <domain>.g_nodLookupDomain
+        DNSName qname = dname;
+        vector<DNSRecord> dummy;
+        qname.appendRawLabel(g_nodLookupDomain);
```
rgacogne: This will only add one label, so I don't think it that's what you want.

pdns/recursordist/ext/Makefile.am
```diff
@@ -1,12 +1,15 @@
...
-	json11
+	json11 \
+	probds
 
 EXTRA_DIST = \
 	luawrapper/include/LuaContext.hpp \
 	yahttp/LICENSE \
-	json11/LICENSE.txt incbin/incbin.h 
+	json11/LICENSE.txt incbin/incbin.h \
+	probds/hyperloglog.hpp
```
rgacogne: hyperloglog.hpp doesn't seem to be used?

pdns/recursordist/stable-bloom.hh
```diff
@@ -0,0 +1,140 @@
...
+          return false;
+      }
+      return true;
+    }
+    bool testAndAdd(const std::string& data) {
+      auto hashes = hash(data);
+      bool retval = true;
+      for (auto& i : hashes) {
+        if (d_cells.test(i % d_num_cells) == false)
+          retval = false;
```
rgacogne: Couldn't we break here?

pdns/nod.cc
neilcook: yes good point

pdns/recursordist/ext/Makefile.am
neilcook: No it's not. Well it was, but not anymore. We might use it in future for something, but in the meantime I'll remove it.

pdns/recursordist/stable-bloom.hh
neilcook: yes I expect so

pdns/pdns_recursor.cc
neilcook: suggestions for what you think I might want would be welcome :)

pdns/nod.cc
neilcook: yep

pdns/pdns_recursor.cc
rgacogne: Of course, sorry :-) I think the easiest way would be to make `g_nodLookupDomain` a `DNSName`, then `qname += g_nodLookupDomain` should do the needful. You would need to replace `g_nodLookupDomain.compare("") == 0` with `g_nodLookupDomain.empty()`, and I think you're set.

neilcook: Oh I see, so even if g_nodLookupDomain contains "foo.bar.com", it will only append "foo"?

rgacogne: Actually no, it would append a single label of length 11, ie `\11foo.bar.com`.

neilcook: Because that's not what I see - I see the whole domain being appended

rgacogne: So, if the original name is `www`, do you get `\3www\3foo\3bar\3com\0` or `\3www\11foo.bar.com\0`?

rgacogne: I haven't tested to be clear :)

neilcook: Oh I see, I just looked cursorily at the lookup - you're right, it's treating "foo.bar.com" as a single label

neilcook: Ok modified to append the raw labels one at a time

pdns/pdns_nod.cc
```diff
@@ -0,0 +1,24 @@
+/*
```
rgacogne: It looks like this file and the corresponding .hh (and symlinks) are not in use?

neilcook:
Excellent! Except someone else will have to merge, cos I don't have permission.

rgacogne:
> Excellent! Except someone else will have to merge, cos I don't have permission.

I'd be willing to merge it, but I'm cautious about PR with "WIP" in their title :) I was also wondering if you wanted to squash a bit before the merge?

neilcook:
Ok removed WIP, because it's not anymore. Shall i just squash into one commit?

rgacogne:
One commit would be good yes, especially since it looks like a few of the existing ones didn't compile, which is always painful while bisecting!

neilcook:
Ok done

Issue #6693, opened 2018-05-29
execution failure in wildcard LUA records leads to NOERROR


Habbie:
Because https://github.com/PowerDNS/pdns/blob/master/pdns/packethandler.cc#L399 does nothing to report errors upstream.

Habbie:
- [ ] document all available variables/functions

Contents of `_G` with git master at ce633f7804e765481a7d7c14a1d6b0eecdc73d5d compiled against LuaJIT:
```
_G	table: 0x0e100960
_VERSION	Lua 5.1
asnum	function: 0x0e11c6b8
assert	function: builtin#2
bestwho	127.0.0.1
bit	table: 0x0e105ca8
closestMagic	function: 0x0e11b778
collectgarbage	function: builtin#27
continent	function: 0x0e11c638
coroutine	table: 0x0e102708
country	function: 0x0e11c730
createForward	function: 0x0e100ac8
createForward6	function: 0x0e100b50
createReverse	function: 0x0e100a40
createReverse6	function: 0x0e100be0
debug	table: 0x0e105690
dh	userdata 0x0e11a5b8
dnssecOK	false
dofile	function: builtin#25
dupDNSPacket	function: 0x0e114040
e5ddced079fc405aa4937b386ca387d2	function: 0x0e106c18
ednsPKTSize	512
error	function: builtin#19
gcinfo	function: builtin#26
geoiplookup	function: 0x0e11c480
getfenv	function: builtin#10
getmetatable	function: builtin#8
ifportup	function: 0x0e100c60
ifurlup	function: 0x0e100ce0
include	userdata: 0x0e11c8b8
io	table: 0x0e1039e0
ipairs	function: builtin#7
jit	table: 0x0e106268
latlon	function: 0x0e11b698
latlonloc	function: 0x0e11b6f0
latlonMagic	function: 0x0e1009b0
load	function: builtin#23
loadfile	function: builtin#22
loadstring	function: builtin#24
math	table: 0x0e104ab0
module	function: 0x0e102e70
netmask	function: 0x0e11c7c0
newCA	function: 0x0e10d070
newCAS	function: 0x0e10d8f0
newDN	function: 0x0e108870
newDNSPacket	function: 0x0e113fc0
newDR	function: 0x0e111958
newDRR	function: 0x0e10a6d0
newNetmask	function: 0x0e10e788
newNMG	function: 0x0e110de8
newproxy	function: builtin#28
newQType	function: 0x0e10e028
[...]

PR #6731, opened 2018-06-12
Address Auth LUA Records issues

chbruyand:
### Short description
This PR addresses some known LUA Records issues :
 - better check input lists
 - report lua wildcards errors
 - exposes `DNSName::getRawLabels` in lua env
 - better document LUA functions and objects

This is an effort to tick some boxes in #6693 

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


chbruyand:
Added some documentation which still has TODOs.

pdns/lua-record.cc
```diff
@@ -356,11 +368,15 @@ static bool getLatLon(const std::string& ip, string& loc)
...
 {
+  if (wips.empty()) {
+    throw std::invalid_argument("The IP list cannot be empty");
+  }
   map<double,vector<ComboAddress> > ranked;
   double wlat=0, wlon=0;
   getLatLon(bestwho.toString(), wlat, wlon);
   //        cout<<"bestwho "<<wlat<<", "<<wlon<<endl;
   vector<string> ret;
+  g_log<<Logger::Debug<< __FILE__ << ":" << __LINE__<< " " << __func__ << " wips.size()="<< wips.size() <<endl;
```
Habbie: Please wrap these in `DLOG(..)` (unless they will be removed before merge, of course!)

docs/dnsupdate.rst
```diff
@@ -1,519 +0,0 @@
-Dynamic DNS Update (RFC2136)
```
Habbie: I presume removing this file was a mistake :)

docs/lua-records/reference/comboaddress.rst
```diff
@@ -0,0 +1,112 @@
+.. _ComboAddress:
```
Habbie: Wonder if we could symlink this to the dnsdist comboaddress docs?

pdns/lua-record.cc
Habbie: (I see now they do get removed later)

docs/dnsupdate.rst
chbruyand: Indeed! I restored the file.

docs/lua-records/reference/comboaddress.rst
chbruyand: The content is not exactly the same as I also documented the `ComboAddressSet` functionalities.

chbruyand:
Rebased onto master.

This PR doesn't fix all issues reported in #6693 but is already a great documentation improvement. We should merge it IMHO to avoid further PR-rebasing-related efforts. Otherwise, could we state what's *really* missing here ?

Could we have an another review ?

PR #6734, opened 2018-06-13
dnsdist: Add SetECSAction to set an arbitrary outgoing ECS value

### LINKED ISSUE #6404: dnsdist needs Action to send a specific EDNS Client Subnet option to downstream
- Program: dnsdist
 - Issue type: feature request

### Short description
Some operators support EDNS Client Subnet operations on part of their listening IP addresses. Depending on where a query came in on, they would like to tell the backend the IP address of the client, or not. Instead of the client IP, they may want to send an IPv4 or IPv6 netmask that represents at least the resolver location in a useful way. 

Currently, dnsdist can supply the IP address of the client via EDNS Client Subnet to backend servers, and this can be turned off with an action or from Lua.

It would be great to have a `SetECSSubnetAction(netmask1[, netmask2])`. If only 1 netmask is provided, that one is always used. If a second one is provided and it is IPv6, that one is used for IPv6 users. 

### Usecase
Provide ECS for some people, but give CDNs something useful for the rest.

### Description
Get an action that allows to override ECSSubnets.

rgacogne:
### Short description

Closes #6404.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/pdns/dnsdist-lua-actions.cc
+++ b/pdns/dnsdist-lua-actions.cc
@@ -651,11 +651,11 @@ public:
 class SetECSAction : public DNSAction
 {
 public:
-  SetECSAction(Netmask v4): d_v4(v4), d_hasV6(false)
+  SetECSAction(const Netmask& v4): d_v4(v4), d_hasV6(false)
   {
   }
 
-  SetECSAction(Netmask v4, Netmask v6): d_v4(v4), d_v6(v6), d_hasV6(true)
+  SetECSAction(const Netmask& v4, const Netmask& v6): d_v4(v4), d_v6(v6), d_hasV6(true)
   {
   }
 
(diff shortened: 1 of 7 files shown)


pdns/dnsdist-lua-actions.cc
```diff
@@ -648,6 +648,47 @@ class DisableECSAction : public DNSAction
...
 };
 
+class SetECSAction : public DNSAction
+{
+public:
+  SetECSAction(Netmask v4): d_v4(v4), d_hasV6(false)
+  {
+  }
+
+  SetECSAction(Netmask v4, Netmask v6): d_v4(v4), d_v6(v6), d_hasV6(true)
```
chbruyand: Couldn't we `const &` constructor arguments (here and with v4 only)?

rgacogne: Done!

PR #6737, opened 2018-06-14
dnsdist: add consistent hash builtin policy

chbruyand:
### Short description
This adds a new load balancing policy that is based on consistent hashing. Unlike the ``whashed`` policy, this distribution will keep consistent over time. Adding or removing servers will only remap a small part of the queries.

For the record, this also adds a`uuid` field to the `DownstreamState` class.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdist.cc
+++ b/pdns/dnsdist.cc
@@ -649,6 +649,10 @@ void DownstreamState::setId(const boost::uuids::uuid& newId)
 
 void DownstreamState::setWeight(int newWeight)
 {
+  if (newWeight < 1) {
+    errlog("Error setting server's weight: downstream weight value must be greater than 0.");
+    return ;
+  }
   weight = newWeight;
   if (!hashes.empty()) {
     hash();
--- a/pdns/dnsdist.cc
+++ b/pdns/dnsdist.cc
@@ -627,6 +627,7 @@ bool DownstreamState::reconnect()
 
 DownstreamState::DownstreamState(const ComboAddress& remote_, const ComboAddress& sourceAddr_, unsigned int sourceItf_, size_t numberOfSockets): remote(remote_), sourceAddr(sourceAddr_), sourceItf(sourceItf_)
 {
+  id = t_uuidGenerator();
   threadStarted.clear();
 
   mplexer = std::unique_ptr<FDMultiplexer>(FDMultiplexer::getMultiplexerSilent());
@@ -642,6 +643,7 @@ DownstreamState::DownstreamState(const ComboAddress& remote_, const ComboAddress
     sw.start();
     infolog("Added downstream server %s", remote.toStringWithPort());
   }
+
 }
 
 std::mutex g_luamutex;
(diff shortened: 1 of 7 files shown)


pdns/dnsdist.cc
```diff
@@ -721,6 +723,48 @@ shared_ptr<DownstreamState> whashed(const NumberedServerVector& servers, const D
...
+ */
+
+shared_ptr<DownstreamState> chashed(const NumberedServerVector& servers, const DNSQuestion* dq)
+{
+  std::map<unsigned int, shared_ptr<DownstreamState>> circle = {};
+  unsigned int qhash = dq->qname->hash(g_hashperturb);
+
+  for (const auto& d: servers) {
+    if (d.second->isUp()) {
+      if (d.second->hashes.empty()) {
```
rgacogne: We probably need a per-server `mutex` when accessing `hashes`.

pdns/dnsdist.cc
```diff
@@ -721,6 +723,48 @@ shared_ptr<DownstreamState> whashed(const NumberedServerVector& servers, const D
...
+        // computes server's points
+        // @todo check if server's weight can change over time
```
rgacogne: I don't think so, but even if it does I think it might be cleaner to compute the points at the server's creation and when the weight changes, instead of doing it here?

pdns/dnsdist.cc
```diff
@@ -721,6 +723,48 @@ shared_ptr<DownstreamState> whashed(const NumberedServerVector& servers, const D
...
+        while (w > 0) {
+          std::string uuid = boost::str(boost::format("%s-%d") % d.second->id % w);
+          unsigned int wshash = burtleCI((const unsigned char*)uuid.c_str(), uuid.size(), g_hashperturb);
+          d.second->hashes.insert(wshash);
+          --w;
+        }
+      }
+      for (const auto& h: d.second->hashes) {
+        // put server's hashes on the circle
+        circle.insert(std::make_pair(h, d.second));
```
rgacogne: Do you think it would make sense to compute the circle when a server is added / removed or changes state?

pdns/dnsdist.cc
chbruyand: Definitely !

rgacogne:
Given that this PR has currently very little impact to the existing code I'm scheduling it for 1.3.x, we might need to rethink that if the impact grows.

pdns/dnsdist.cc
chbruyand: Yes, but I did it there because it prevents unnecessary computation when the policy isn't set to `chashed` and I got the feeling the policy handler could be called even if the different pools were not set to it.

pdns/dnsdist.cc
chbruyand: Hum, yes we could, but as the circle is pool dependent, it would require to change the signature of the policy handlers and pass the pool the query is assigned to. Didn't get the feeling it was an easy move as the policy system is quiet generic.

chbruyand: I mean, to cache it, it would require to be aware of the context we are called in. Is that something we want ?

pdns/dnsdist.cc
rgacogne: Right, I feel pre-computing the points might be worth it since it should happen very rarely but let's not rush that decision.

pdns/dnsdist.cc
rgacogne: Ah yes, it would lead to quite substantial changes so let's not go there until we have properly benchmarked that new policy.

chbruyand:
Moves hashes computation to object initialization and added two setters for `weight` and `id` members.

chbruyand:
It looks like the performances of this is comparable to other policies :
metronome

(from left to right, `chashed`, `whashed`, `leastOutstanding` and `wrandom`)

rgacogne:
It looks like the regression test "Routing: WRandom (overflow)" is failing with this PR, we need to check whether the test is testing something it shouldn't or if the PR actually is breaking something.

pdns/dnsdist.cc
```diff
@@ -721,6 +754,47 @@ shared_ptr<DownstreamState> whashed(const NumberedServerVector& servers, const D
...
+    if (d.second->isUp()) {
+      // make sure hashes have been computed
+      if (d.second->hashes.empty()) {
+        d.second->hash();
+      }
+      {
+        ReadLock rl(&(d.second->d_lock));
+        const auto& server = d.second;
+        // we want to keep track of the last hash
+        if (max < *(server->hashes.rbegin())) {
```
rgacogne: Don't we have an issue here if the weight is set to zero, or even a negative since it's a signed integer,?

chbruyand: Didn't consider that as the weight value is bounded when the server is added (see https://github.com/PowerDNS/pdns/blob/04437f1a023bbe90f13e3f8372f1b49f3554c9bb/pdns/dnsdist-lua.cc#L269)

rgacogne: Right, although it looks like we have a `Lua` binding for the `weight` member that does not check anything :'(

chbruyand: Just added some checks around the `weight` member binding

pdns/dnsdist.cc
```diff
@@ -2258,6 +2336,15 @@ try
 
   auto todo=setupLua(false, g_cmdLine.config);
 
+  if (g_policy.getLocal()->name == "chashed") {
+    vinfolog("Pre-computing hashes for consistent hash load-balancing policy");
```
rgacogne: I haven't tested but it looks like this only work for the global load-balancing policy, perhaps we should also test pool-specific load-balancing policies (set with `setPoolServerPolicy()`)?

chbruyand: https://github.com/PowerDNS/pdns/pull/6737/files#diff-194369649d922d5a26ab47ac9c895d33R770
Indeed, this only tests against global policy in order to decide whether or not to pre-compute hashes. However, they will be computed on demand if needed.

rgacogne: Right, I agree this is not an issue, but I was just wondering if we might want to iterate over the pools there and check their policy as well? But that's not blocking either way.

chbruyand: I'll update the check :)

PR #6740, opened 2018-06-14
 dnsdist: Add a negative TTL option to the packet cache

### LINKED ISSUE #6579: dnsdist: max-negative-ttl setting for PacketCache
- Program: dnsdist
 - Issue type: Feature request

### Short description
dnsdist uses the minimum TTL of the records contained in the response
when adding an entry to the cache, capped by the maximum TTL setting.
For a negative answers e.g. for the root zone, this TTL is 86400 s.

It has been noticed that Google Chrome (at least on mobile devices) will send on each app start three randomly generated one label query names, likely nonexistent, to detect if the ISP resolver is hijacking NXDOMAIN responses. On a network with many mobile users this can take up a lot of space in the PacketCache and result in a bad cache hit rate.

See also discussion on dnsdist mailinglist
https://mailman.powerdns.com/pipermail/dnsdist/2018-May/000421.html

It would be useful to have a max negative TTL setting in the dnsdist's packet cache.

### Usecase
I would like to prevent the dnsdist's packet cache from being filled up with negative answers. If the packet cache is configured with a limited size it can reduce cache hit rate.

### Description
For negative answers (NXDOMAIN and NODATA) the new introduced max-negative-ttl setting (default 3600 s or lower) should cap the effective negative TTL from the response.  I guess, max-negative-ttl could be an optional parameter for the newPacketCache statement.

rgacogne:
### Short description

So it's possible to cap the `TTL` of `NODATA` and `NXDOMAIN` answers.
Also fixes a bug in the `editTTLs()` feature, where the wrong section was passed to the callback.

Closes #6579.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/pdns/dnsdistdist/docs/reference/config.rst
+++ b/pdns/dnsdistdist/docs/reference/config.rst
@@ -506,10 +506,13 @@ PacketCache
 A Pool can have a packet cache to answer queries directly in stead of going to the backend.
 See :doc:`../guides/cache` for a how to.
 
-.. function:: newPacketCache(maxEntries[, maxTTL=86400[, minTTL=0[, temporaryFailureTTL=60[, staleTTL=60[, dontAge=false[, numberOfShards=1[, deferrableInsertLock=true]]]]]]]) -> PacketCache
+.. function:: newPacketCache(maxEntries[, maxTTL=86400[, minTTL=0[, temporaryFailureTTL=60[, staleTTL=60[, dontAge=false[, numberOfShards=1[, deferrableInsertLock=true[, maxNegativeTTL=3600]]]]]]]) -> PacketCache
 
   .. versionchanged:: 1.3.0
     ``numberOfShards`` and ``deferrableInsertLock`` parameters added.
+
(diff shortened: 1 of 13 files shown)


pdns/dnsdistdist/docs/reference/config.rst
```diff
@@ -521,6 +524,7 @@ See :doc:`../guides/cache` for a how to.
   :param bool dontAge: Don't reduce TTLs when serving from the cache. Use this when :program:`dnsdist` fronts a cluster of authoritative servers
   :param int numberOfShards: Number of shards to divide the cache into, to reduce lock contention
   :param bool deferrableInsertLock: Whether the cache should give up insertion if the lock is held by another thread, or simply wait to get the lock
+  :param bool maxNegativeTTL: Cache a NXDomain or NoData answer from the backend for at most this amount of seconds, even if the TTL of the SOA record is higher
```
pr0vieh: i think this param must be an int not a bool just doc ;)

PR #6744, opened 2018-06-19
rec: Support multiple values for the same EDNS option in gettag

### LINKED ISSUE #6307: Additional EDNS options with same `OPTION-CODE` overwrite previous entry
- Program: Recursor
 - Issue type: Bug report

### Short description

When sending a DNS message to the recursor (and possible also to the auth server? (untested)) with multiple EDNS options with the same `OPTION-CODE` (see [RFC 6891](https://tools.ietf.org/html/rfc6891)) earlier EDNS options with `OPTION-CODE` will be overwritten by later ones with the same code and will not be available in a Lua script `gettag` function.

### Environment

 - Operating system: Void Linux rolling release (4.14.12_3, x86_64)
 - Software version: HEAD (as of 2018-02-26) of master branch, commit 32a290946d2ec0f740a437cf58afc8fe1612281c
 - Software source: PowerDNS/pdns:master

### Steps to reproduce

1. Acquire software that can send a DNS message with EDNS options
2. Set two EDNS options with the same `OPTION-CODE` but with different `OPTION-DATA` (as per https://tools.ietf.org/html/rfc6891#section-6.1.2)
3. Create the file `tmp.lua` with the following content:

```
function gettag(remote_ca, edns, loc, qname, qtype, ednsoptions)
    io.stderr:write("--- BEGIN ednsoptions\n")
    for k, v in pairs(ednsoptions) do
       io.stderr:write("- " .. k .. ": " .. v:getContent() .. "\n")
    end
    io.stderr:write("--- END   ednsoptions\n")
    -- NOTE: ednsoptions is only passed if `gettag-needs-edns-options=yes`

    local tag, policytags, dqdata, username = nil, nil, nil, nil
    return tag, policytags, dqdata, username, deviceid_binary
end
```

4. Start recursor as follows: `sudo ./pdns_recursor --loglevel=1 --quiet=no --gettag-needs-edns-options=yes --lua-dns-script=tmp.lua`

### Expected behaviour

I would expect the following to be printed:

```
    --- BEGIN ednsoptions
    - OPTION-CODE: OPTION-DATA-1
    - OPTION-CODE: OPTION-DATA-2
    --- END   ednsoptions
```

### Actual behaviour

The following is printed:

```
    --- BEGIN ednsoptions
    - OPTION-CODE: OPTION-DATA-2
    --- END   ednsoptions
```

### Other information

This line seems to be the problem https://github.com/PowerDNS/pdns/blob/master/pdns/ednsoptions.cc#L98 in assuming there will only every be 1 option with `optionCode`.  However this restriction is not mentioned in [RFC 6891](https://tools.ietf.org/html/rfc6891) *and* Internet Draft [Client ID in Forwarded DNS Queries](https://tools.ietf.org/html/draft-tale-dnsop-edns0-clientid-01) explicitly mentions:

> Multiple ECID options MAY appear in the OPT record.  However, the same IDENTIFIER-TYPE SHOULD not appear more than once, and each ECID option MUST only carry one IDENTIFIER-TYPE and CLIENT-IDENTIFIER pair.

rgacogne:
### Short description

This PR adds support for retrieving multiple values for the same `EDNS` option in the recursor, not breaking the existing API. The regular interface does not support more than one value, so it will keep returning only the first value for a given option, while two new methods make it possible to retrieve all the values:
- `EDNSOptionView:count()`
- `EDNSOptionView:getValues()`

The `FFI` interface already supported returning more than one value so support for multiple values has been transparently added to it.

This PR also adds several regression tests for the Lua interface.
 
Closes #6307.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/pdns/ednsoptions.hh
+++ b/pdns/ednsoptions.hh
@@ -32,14 +32,21 @@ struct EDNSOptionCode
 /* extract a specific EDNS0 option from a pointer on the beginning rdLen of the OPT RR */
 int getEDNSOption(char* optRR, size_t len, uint16_t wantedOption, char ** optionValue, size_t * optionValueSize);
 
-struct EDNSOptionView
+struct EDNSOptionViewValue
 {
   const char* content{nullptr};
   uint16_t size{0};
 };
 
+struct EDNSOptionView
+{
+  std::vector<EDNSOptionViewValue> values;
+};
+
+typedef std::map<uint16_t, EDNSOptionView> EDNSOptionViewMap;
+
 /* extract all EDNS0 options from a pointer on the beginning rdLen of the OPT RR */
-int getEDNSOptions(const char* optRR, size_t len, std::map<uint16_t, EDNSOptionView>& options);
+int getEDNSOptions(const char* optRR, size_t len, EDNSOptionViewMap& options);
 
 void generateEDNSOption(uint16_t optionCode, const std::string& payload, std::string& res);
 
--- a/pdns/dnsrecords.cc
+++ b/pdns/dnsrecords.cc
@@ -548,12 +548,12 @@ bool getEDNSOpts(const MOADNSParser& mdp, EDNSOpts* eo)
     for(const MOADNSParser::answers_t::value_type& val :  mdp.d_answers) {
       if(val.first.d_place == DNSResourceRecord::ADDITIONAL && val.first.d_type == QType::OPT) {
         eo->d_packetsize=val.first.d_class;
-       
+
         EDNS0Record stuff;
         uint32_t ttl=ntohl(val.first.d_ttl);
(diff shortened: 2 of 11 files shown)


pdns/ednsoptions.hh
```diff
@@ -32,14 +32,21 @@ struct EDNSOptionCode
...
 int getEDNSOption(char* optRR, size_t len, uint16_t wantedOption, char ** optionValue, size_t * optionValueSize);
 
-struct EDNSOptionView
+struct EDNSOptionViewValue
 {
   const char* content{nullptr};
   uint16_t size{0};
 };
 
+struct EDNSOptionView
```
chbruyand: Did you keep a struct instead of a type alias for style/readability ?

rgacogne: Mostly because otherwise it was mapped as a Lua table by LuaWrapper and I couldn't get my bindings to work.

PR #6749, opened 2018-06-22
dnsdist: Add the ability to dump a summary of the cache content

rgacogne:
### Short description

The new feature writes a summary of the packet cache content to a file, dumping for each entry:
- qname
- remaining TTL
- qtype
-  computed hash
- response's length
- TCP or UDP
- the timestamp of when the entry was added to the cache. 

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/pdns/dnsdist-cache.cc
+++ b/pdns/dnsdist-cache.cc
@@ -160,6 +160,7 @@ bool DNSDistPacketCache::get(const DNSQuestion& dq, uint16_t consumed, uint16_t 
 {
   std::string dnsQName(dq.qname->toDNSString());
   uint32_t key = getKey(dnsQName, consumed, (const unsigned char*)dq.dh, dq.len, dq.tcp);
+
   if (keyOut)
     *keyOut = key;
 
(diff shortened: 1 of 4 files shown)


phonedph1:
This is much better than what I had come up with!

pieterlexis:
Looks good. Perhaps the TTL and QType should be swapped?

Also, I would like a 'header' on the first line like this:

```
; qname TTL type
```

rgacogne:
>  Perhaps the TTL and QType should be swapped?

I had the same feeling but decided to keep the same order than the one when dumping the recursor's packet cache. But the more I think about it, the less relevant it is so I'm OK with swapping it.

> Also, I would like a 'header' on the first line like this:

I'm not sure it's really needed, because that's almost what we already have:
```
; dnsdist's packet cache dump follows
;
powerdns.com. 3572 A ; key 2000355501, length 57, tcp 0, added 1529912626
www.powerdns.com. 3590 A ; key 4247463626, length 61, tcp 0, added 1529912644
nx.powerdns.com. 3595 A ; key 1069543761, length 109, tcp 0, added 1529912649
```

PR #6754, opened 2018-06-25
 dnsdist: Detect ECS collisions in the packet cache

rgacogne:
### Short description

We did not detect a collision involving two queries for the same qname, qtype, qclass and flags but with different `ECS` values hashing to the same key.
This commit adds an option to the packet cache to parse and keep the `ECS` value in order to detect this kind of collisions.
It's not enabled by default because parsing the `ECS` value has a cost and most deployment probably don't care.

This PR is based on the branch from #6747, so will need to be rebased after this one has been merged.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/pdns/dnsdist-cache.cc
+++ b/pdns/dnsdist-cache.cc
@@ -23,8 +23,11 @@
 #include "dolog.hh"
 #include "dnsparser.hh"
 #include "dnsdist-cache.hh"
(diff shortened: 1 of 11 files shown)


rgacogne:
Rebased following the merge of #6747.

PR #6774, opened 2018-07-06
calidns: accurate qps targets

wojas:
### Short description

Instead of a naive fixed sleep between batch sends, check the actual time
passed to determine how long to sleep. This allows calidns to actually
reach qps targets at higher rates.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code


wojas:
Thanks, added a comment!

PR #6779, opened 2018-07-09
rec: use native QType in lua scripting

mind04:
### Short description
qtype.getName() and qtype.getCode() is now possible without type casting.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/recursordist/docs/lua-scripting/dq.rst
+++ b/pdns/recursordist/docs/lua-scripting/dq.rst
@@ -17,7 +17,11 @@ The DNSQuestion object contains at least the following fields:
 
   .. attribute:: DNSQuestion.qtype
 
-      Type this query is for as an integer, can be compared against ``pdns.A``, ``pdns.AAAA``.
+     .. versionchanged:: 4.2.0
+
+       The qtype parameter is now a QType instead of an int
+
+      Type this query is for as an QType, can be compared against ``pdns.A``, ``pdns.AAAA``.
 
   .. attribute:: DNSQuestion.rcode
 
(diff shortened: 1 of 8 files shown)


pdns/recursordist/docs/lua-scripting/dq.rst
```diff
@@ -17,7 +17,11 @@ The DNSQuestion object contains at least the following fields:
 
   .. attribute:: DNSQuestion.qtype
 
-      Type this query is for as an integer, can be compared against ``pdns.A``, ``pdns.AAAA``.
+     .. versionchanged:: 4.2.0
+
+       The qtype parameter is now a QType instead of an int
+
+      Type this query is for as an QType, can be compared against ``pdns.A``, ``pdns.AAAA``.
```
Habbie: This sentence is unreadable (it already was), but could you perhaps fix it for us anyway? :)

rgacogne:
I'm not sure what the impact of this PR is, performance-wise, since we now have to go back to `C++` from `Lua` to do anything with the `qtype`. It might also be a bit annoying that existing scripts will break after an upgrade, but I guess we can deal with that.

Habbie:
> I'm not sure what the impact of this PR is, performance-wise, since we now have to go back to C++ from Lua to do anything with the qtype. It might also be a bit annoying that existing scripts will break after an upgrade, but I guess we can deal with that.

Under LuaJIT, the performance hit of going back to C++ is huge.

pieterlexis:
> It might also be a bit annoying that existing scripts will break after an upgrade, but I guess we can deal with that.

This needs upgrade notes

PR #6783, opened 2018-07-10
Add Basic Auth support to changelog-from-pr.py

rgacogne:
### Short description

GitHub's API rate limiting is quite aggressive, but is more permissive for authenticated users.
This PR allows authenticating via username and password, but you can also submit a personal access token instead of a password if, for example, you have 2-factors authentication enabled on your account.

### Checklist

I have:
- [ ] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/build-scripts/changelog-from-pr.py
+++ b/build-scripts/changelog-from-pr.py
@@ -4,10 +4,13 @@ import requests
 import sys
 import argparse
 import re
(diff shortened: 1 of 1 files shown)


pieterlexis:
haha cool!

I have a `.netrc` to do this ;). But this is better

Issue #5469, opened 2017-06-27
PowerDNS Auth does not forward EDNS Client Subnet for ALIAS records


DarkDeviL:
- Program: Authoritative 
 - Issue type: Bug report 

### Short description

Authoritative PowerDNS server does not forward EDNS Client Subnet to the server specified as "resolver=" when processing ALIAS records, which quite much destroys the performance when using it together with CDN services that rely on the EDNS Client Subnet to route users to the nearest server.

Not forwarding isn't really consistent with the documentation about ALIAS processing either:
-> https://github.com/PowerDNS/pdns/blob/master/docs/WIP/alias.md
```
## EDNS Subnet
Authoritative servers sometimes give out different answers based on the IP
address of the resolver asking.  As a further refinement, some resolvers can
pass along (part) of the actual stub-resolver asking the question, and base
its answer on that 'real' address. 

For ALIAS processing, implementors are encouraged to pass along or use all
knowledge of the remote client IP address when retrieving A or AAAA records.
```

### Environment

 - Operating system: Debian 8
 - Software version: master branch (0.0.1728g70bd485-1pdns.jessie)
 - Software source: PowerDNS repository

### Steps to reproduce

1. Install (or compile) PowerDNS
2. In the configuration, set:
```
edns-subnet-processing=yes
expand-alias=yes
resolver=8.8.8.8
```
3. Create zone "example.com" where you ALIAS "example.com" and/or "www.example.com" to "www.google.com"
4. Look up either "example.com" or "www.example.com" and compare "www.google.com"
```
dig @ns1.example.com example.com
dig @ns1.example.com www.example.com
dig @8.8.8.8 www.google.com
```
5. With your server "ns1.example.com" being in the United States, and you being in e.g. United Kingdom where Google has a local cache server within your ISP's network, the first two dig will route you towards Google's nearest instances (e.g. Netherlands), rather than the local ones within your ISP in United Kingdom.
With CDN services, the same happens where both "example.com" and "www.example.com" are reaching locations closer to the United States rather than locations close to (or in) United Kingdom.

### Expected behaviour

When 192.0.2.123 queries ns1.example.com for "example.com" or "www.example.com" with a EDNS Client Subnet of "198.51.100.0/24", then "198.51.100.0/24" should be forwarded to the resolver 8.8.8.8, as set in "resolver=".

### Actual behaviour

[...]

PR #6811, opened 2018-07-26
[do not merge] apply ALIAS scopemask after chasing

Habbie:
### Short description
Before this patch, the scopebits on an ALIAS result from a backend would be thrown away. With this patch, we pass the scopebits down into the dnsproxy so it can apply it to the resulting A/AAAA records.

Related: #5469 

Should this be configurable?

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsproxy.cc
+++ b/pdns/dnsproxy.cc
@@ -88,7 +88,7 @@ void DNSProxy::go()
 }
 
 //! look up qname target with r->qtype, plonk it in the answer section of 'r' with name aname
-bool DNSProxy::completePacket(DNSPacket *r, const DNSName& target,const DNSName& aname)
+bool DNSProxy::completePacket(DNSPacket *r, const DNSName& target,const DNSName& aname, const uint8_t scopeMask)
 {
   if(r->d_tcp) {
     vector<DNSZoneRecord> ips;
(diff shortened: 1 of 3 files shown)


zeha:
I think this should be transparent, so not configurable. If one needs to tamper with the IPs I’d suggest doing it in the recursor instead.

DarkDeviL:
For privacy reasons/concerns, it could be made configurable in regards to what scope you're forwarding, e.g.

```
alias-minimum-scope-v4=22
alias-minimum-scope-v6=48
```

If client IP is 203.0.113.123/32 , with ECS 198.51.100.0/24, then 198.51.100.0/22 (converted out of ECS+ alias-minimum-scope-v4) would be forwarded.

If client IP is 203.0.113.123/32, without no ECS at all, then 203.0.112.0/22 (converted out of Client IP + alias-minimum-scope-v4, due to missing ECS) would be forwarded.

Other than that, I would also say as transparent as possible as mentioned above.

Habbie:
> For privacy reasons/concerns, it could be made configurable in regards to what scope you're forwarding, e.g.

But this query is coming from a resolver, which should already have applied some sensible scope mask - or am I missing something?

DarkDeviL:
> But this query is coming from a resolver, which should already have applied some sensible scope mask - or am I missing something?

I was more thinking if you're providing authoritative DNS, and not providing the resolver yourself.

That way, you're relying on a third party, and forwarding "ALIAS" queries through them. You're obviously the lord of your queries yourself, and shouldn't be pointing the resolver= to something or someone you don't trust. Both OpenDNS and Google DNS does not seem to rely on the subnet you're sending, but only the subnet based on the true client IP they see.

So yeah, in the end, it might just be a "waste", to implement something like that for the ALIAS part.

PR #6812, opened 2018-07-26
rec: load Lua scripts only in worker threads

Habbie:
### Short description
The extra thread introduced in #6567 would load the Lua script but then do nothing with it. This PR removes the Lua script from the extra thread.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/pdns_recursor.cc
+++ b/pdns/pdns_recursor.cc
@@ -3603,7 +3603,7 @@ try
   setupNODThread();
 #endif /* NOD_ENABLED */
   
-  if(worker) {
+  if(worker && (!g_weDistributeQueries || t_id != s_distributorThreadID)) {
     try {
       if(!::arg()["lua-dns-script"].empty()) {
         t_pdl = std::make_shared<RecursorLua4>();


pdns/pdns_recursor.cc
```diff
@@ -3603,16 +3603,18 @@ try
   setupNODThread();
 #endif /* NOD_ENABLED */
   
-  try {
-    if(!::arg()["lua-dns-script"].empty()) {
-      t_pdl = std::make_shared<RecursorLua4>();
-      t_pdl->loadFile(::arg()["lua-dns-script"]);
-      g_log<<Logger::Warning<<"Loaded 'lua' script from '"<<::arg()["lua-dns-script"]<<"'"<<endl;
+  if(worker) {
```
chbruyand: I was wondering if you want to load this script on the distributor thread when query distribution is enabled. If not, you may want to update the condition to something like `if (worker && (!g_weDistributeQueries || t_id != s_distributorThreadID)) {`.

Habbie: yes! pushed!

PR #6813, opened 2018-07-26
dnsdist: DSTPortRule 

phonedph1:
### Short description
Allows matching based on the destination port of the question. This allows one to apply more powerful rules to ports bound for DoT for instance.

### Checklist
I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdistdist/dnsdist-rules.hh
+++ b/pdns/dnsdistdist/dnsdist-rules.hh
@@ -603,11 +603,10 @@ class DSTPortRule : public DNSRule
 public:
   DSTPortRule(uint16_t port) : d_port(port)
   {
-    d_port_htons = htons(d_port);
   }
   bool matches(const DNSQuestion* dq) const override
   {
-    return d_port_htons == dq->local->sin4.sin_port;
+    return htons(d_port) == dq->local->sin4.sin_port;
   }
   string toString() const override
   {
--- a/pdns/dnsdistdist/dnsdist-rules.hh
+++ b/pdns/dnsdistdist/dnsdist-rules.hh
@@ -598,6 +598,26 @@ private:
   uint8_t d_opcode;
 };
 
+class DSTPortRule : public DNSRule
+{
+public:
+  DSTPortRule(uint16_t port) : d_port(port)
+  {
+    d_port_htons = htons(d_port);
+  }
+  bool matches(const DNSQuestion* dq) const override
+  {
+    return d_port_htons == dq->local->sin4.sin_port;
+  }
+  string toString() const override
+  {
+    return "dst port=="+std::to_string(d_port);
+  }
+private:
+  uint16_t d_port;
+  uint16_t d_port_htons;
+};
+
 class TCPRule : public DNSRule
 {
 public:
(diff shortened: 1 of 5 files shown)


pdns/dnsdistdist/dnsdist-rules.hh
```diff
@@ -598,6 +598,26 @@ private:
...
+  {
+    return d_port_htons == dq->local->sin4.sin_port;
+  }
+  string toString() const override
+  {
+    return "dst port=="+std::to_string(d_port);
+  }
+private:
+  uint16_t d_port;
+  uint16_t d_port_htons;
```
rgacogne: Given how cheap `htons()` is, keeping the two values here feels a bit wrong.

zeha:
I felt like the rule should match the entire local address, not just the port... and/or match the object returned from addLocal() (if it does return something)

rgacogne:
Unless I'm mistaken, you can already use `NetmaskGroupRule` if you want to match the destination address.

zeha:
Yeah, but somehow it's weird that you cannot do both things with one matcher. I'd kinda expect something like this: LocalRule("*:443")
Exact semantics not thought through.

phonedph1:
@zeha while I don't disagree with one potentially being nice - the already existing combo of using a NetmaskGroupRule for that purpose seems just as easy to combine into whatever you want and might provide more re-use for some people.

FWIW we use NMGs to shove a number of destination addresses into a single group, so re-using that group for any port matching we do would still be nice for us. The semantics of something like ```AndRule({nmg group, LocalAddr("*:whatever")})``` seem less intuitive. 

```
range = newNMG()
range:addMask("8.0.0.0/24")
addAction(AndRule({NetmaskGroupRule(range, false), DSTPortRule(911)}), PoolAction("woot"))

localR = newNMG()
localR:addMask("10.10.10.10")
addAction(AndRule({NetmaskGroupRule(localR, false), DSTPortRule(911)}), PoolAction("woot"))

> showRules()
#     Matches Rule                                                     Action
0           0 (Dst: 8.0.0.0/24) && (dst port==911)                     to pool woot
1           0 (Dst: 10.10.10.10/32) && (dst port==911)                 to pool woot
>
```

pdns/dnsdistdist/dnsdist-rules.hh
```diff
@@ -598,6 +598,25 @@ private:
...
+  {
+    return htons(d_port) == dq->local->sin4.sin_port;
+  }
+  string toString() const override
+  {
+    return "dst port=="+std::to_string(d_port);
+  }
+private:
+  uint16_t d_port;
+  uint16_t d_port_htons;
```
rgacogne: You removed the code using this member, but somehow forgot to remove the member itself ;-)

rgacogne:
Unless someone is really motivated to work on the more advanced `LocalRule` suggested by @zeha, which looks nice, I'd be in favor of merging this one. The code is simple enough, it has documentation and tests, and doesn't prevent adding a more advanced rule later if we want to. Any thought?

zeha:
Agreed

PR #6816, opened 2018-07-31
 auth: Make sure that we use strict weak records ordering in the API

rgacogne:
### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


zeha:
Please (manually) check that this doesn't break returning records+comments when there is an RRset with no records but comments left...

Apart from that LGTM

rgacogne:
> Please (manually) check that this doesn't break returning records+comments when there is an RRset with no records but comments left...

Nice catch, I reverted the ordering of records and comments in `fillZone()` to the existing behavior, and I added a test detecting a mismatch between records and comments.

rgacogne:
For the record, this solves an issue reported to the mailing-list: https://mailman.powerdns.com/pipermail/pdns-users/2018-July/025420.html

PR #6822, opened 2018-07-31
add incremental slave-check backoff also for failed AXFR 

klaus3000:
### Short description
If SOA-checks against a master with failing AXFR still succeed, we would
constantly try to AXFR the zone. To avoid this, we add the zone to the list
of failed slave-checks to suspend slave-checks (and subsequent AXFR) for
this zone for some time.

### Checklist
I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/pdns/slavecommunicator.cc
+++ b/pdns/slavecommunicator.cc
@@ -615,17 +615,20 @@ void CommunicatorClass::suck(const DNSName &domain, const ComboAddress& remote)
     }
   }
   catch(ResolverException &re) {
-    // The AXFR probably failed due to a problem on the master server. If SOA-checks against this master
-    // still succeed, we would constantly try to AXFR the zone. To avoid this, we add the zone to the list of
-    // failed slave-checks. This will suspend slave-checks (and subsequent AXFR) for this zone for some time.
-    uint64_t newCount = 1;
-    time_t now = time(0);
-    const auto failedEntry = d_failedSlaveRefresh.find(domain);
-    if (failedEntry != d_failedSlaveRefresh.end())
-      newCount = d_failedSlaveRefresh[domain].first + 1;
-    time_t nextCheck = now + std::min(newCount * d_tickinterval, (uint64_t)::arg().asNum("soa-retry-default"));
-    d_failedSlaveRefresh[domain] = {newCount, nextCheck};
-    g_log<<Logger::Error<<"Unable to AXFR zone '"<<domain<<"' from remote '"<<remote<<"' (resolver): "<<re.reason<<" (This was the "<<(newCount == 1 ? "first" : std::to_string(newCount) + "th")<<" time. Excluding zone from slave-checks until "<<nextCheck<<")"<<endl;
+    {
+      Lock l(&d_lock);
+      // The AXFR probably failed due to a problem on the master server. If SOA-checks against this master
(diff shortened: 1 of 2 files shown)


pdns/slavecommunicator.cc
```diff
@@ -615,7 +615,17 @@ void CommunicatorClass::suck(const DNSName &domain, const ComboAddress& remote)
     }
   }
   catch(ResolverException &re) {
-    g_log<<Logger::Error<<"Unable to AXFR zone '"<<domain<<"' from remote '"<<remote<<"' (resolver): "<<re.reason<<endl;
+    // The AXFR probably failed due to a problem on the master server. If SOA-checks against this master
+    // still succeed, we would constantly try to AXFR the zone. To avoid this, we add the zone to the list of
+    // failed slave-checks. This will suspend slave-checks (and subsequent AXFR) for this zone for some time.
+    uint64_t newCount = 1;
+    time_t now = time(0);
+    const auto failedEntry = d_failedSlaveRefresh.find(domain);
```
rgacogne: We need to acquire `d_lock` before accessing `d_failedSlaveRefresh`.

pdns/slavecommunicator.cc
```diff
@@ -615,7 +615,17 @@ void CommunicatorClass::suck(const DNSName &domain, const ComboAddress& remote)
...
+    if (failedEntry != d_failedSlaveRefresh.end())
+      newCount = d_failedSlaveRefresh[domain].first + 1;
+    time_t nextCheck = now + std::min(newCount * d_tickinterval, (uint64_t)::arg().asNum("soa-retry-default"));
```
rgacogne: It might be a bit misleading to use `soa-retry-default` here, at the very least we should document it if we do.

klaus3000: Where shall we document it? AFAIK this whole incremental backoff is not documentet somewhere.

pdns/slavecommunicator.cc
klaus3000: Can you point me to an example how to use d_lock?

rgacogne: Sure, sorry! You need to acquire the lock by doing something like:
```
Lock l(&d_lock);
```
It will create a `l` object that will hold the lock as long as it exists, and will release it as soon as it is destroyed. In this case the easiest way would be to create it right after entering the `catch` block you are in, and then just let it be destroyed when you exit this block.

pdns/slavecommunicator.cc
rgacogne: I think https://github.com/PowerDNS/pdns/blob/master/docs/modes-of-operation.rst#slave-operation might be the right place to write a few words about this change.

klaus3000: I think it would be wise to add an option to turn the backoff (failed SOA and failed AXFR) on/off. Then we can document the exact behavior next to the option

klaus3000:
Is this the second commit the correct usage of d_lock?

rgacogne:
> Is this the second commit the correct usage of `d_lock`?

Yes, thanks!

klaus3000:
Is there something missing from my side?

klaus3000:
I added some docs

PR #6823, opened 2018-07-31
On incoming NOTIFY load our serial from backend to have it available …

### LINKED ISSUE #6821: regression: serial not loaded during getDomainInfo on incoming NOTIFY
- Program: Authoritative
 - Issue type: Bug report

### Short description
In PDNS 4.1, on incoming NOTIFYs, DomainInfo was loaded with:
`  if(!B.getDomainInfo(p->qdomain, di) || !(db=di.backend)) {`

In current Master this has changed to:
`  if(!B.getDomainInfo(p->qdomain, di, false) || !di.backend) {`
Hence,  getSerial is set to false.

If the NOTIFY is valid, the domain is queued for slave-checks. But the serial is needed during the slave-check to determine if the master's serial differs. This is not working anymore:

pdns[6955]: Domain 'nxdomain.at' is stale, master serial 2018031201, our serial 0

So, an incoming NOTIFYs currently always triggers an XFR as the comparison is always done against 0 as "ourserial" is not available.

I think the NOTIFY-processing was optimized a bit too much. I guess replacing false with true is the key.

### Environment
 - Operating system: Debian
 - Software version: master

klaus3000:
### Short description
On incoming NOTIFY load our serial from backend to have it available during slave-check.

Also log ourserial to ease debugging.

Fixes #6821 
### Checklist
I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/packethandler.cc
+++ b/pdns/packethandler.cc
@@ -887,7 +887,7 @@ int PacketHandler::processNotify(DNSPacket *p)
   // Domain verification
   //
   DomainInfo di;
-  if(!B.getDomainInfo(p->qdomain, di, false) || !di.backend) {
+  if(!B.getDomainInfo(p->qdomain, di, true) || !di.backend) {
     if(::arg().mustDo("supermaster")) {
       g_log<<Logger::Warning<<"Received NOTIFY for "<<p->qdomain<<" from "<<p->getRemote()<<" for which we are not authoritative, trying supermaster"<<endl;
       return trySuperMaster(p, p->getTSIGKeyname());
@@ -921,7 +921,7 @@ int PacketHandler::processNotify(DNSPacket *p)
   }
 
   if(::arg().mustDo("slave")) {
-    g_log<<Logger::Debug<<"Queueing slave check for "<<p->qdomain<<endl;
+    g_log<<Logger::Debug<<"Queueing slave check for "<<p->qdomain<<", our serial is "<<di.serial<<endl;
     Communicator.addSlaveCheckRequest(di, p->d_remote);
   }
   return 0;


mind04:
This will always result in an extra db query for every incoming notify and will partial revert #6021
Since the result of the query is only used for debug logging I don't think this should be merged like this

It would be better to only get the serial when the log level is debug. And even then I'm not sure it is worth the performance hit.

klaus3000:
When a NOTIFY is received and processed successfully, the one will be queued for a slave check. The slave check compares the master's serial with our serial. Hence, during slave check, PDNS needs to know our serial.
So, the serial MUST be loaded somewhere (see #6821). This was always done during NOTIFY processing. When you remove it from NOTIFY processing, you have to add it somewhere else. E.g during slave-check. But until now slave-check could rely that ourserial is available, which is not the case anymore after your improvements.

klaus3000:
I guess you are referring to https://github.com/PowerDNS/pdns/pull/6021/commits/47bddbb7cfaa02b4afbbdf1029d9188cb1b5b87a

I think your assumption that the serial was never used war wrong - it was used later during the slave-check.

klaus3000:
Any chance to get SLAVE mode working again on master (with triggering XFR on every incoming NOTIFY)?

rgacogne:
We do indeed always use the serial in `CommunicatorClass::slaveRefresh()`. We could skip the extra DB query in there and do it later. I guess this might help if we get a lot of duplicated NOTIFY messages, but I'm not sure it's worth the trouble?

klaus3000:
Ping?

PS: if you are concerned about needless getSOA queries then take a look at this thread:
http://powerdns.13854.n7.nabble.com/SLAVE-refresh-performance-td7389.html

The logic to find out if a domain needs a slave-check is horrible bad: the "next check" is calculated every slave-check (for all slave domains) instead of calculating it during the last zone-activity (SOA-check or XFR).

Habbie:
Without this PR, we always XFR, even if it is not necessary. So I am merging this PR now for 4.2.0-alpha1 and we can figure out any duplicate DB work before 4.2.0.

PR #6824, opened 2018-07-31
Add the serials when logging the final result of a slave-check

klaus3000:
### Short description
When the final result of the slave-check get logged, always add the respective serials to the log line. This is very useful when debugging slave-check issues
 
### Checklist
I have:
- [x] compiled this code
- [partly] tested this code (only the cases without DNSSEC workarounds, but the code compiles)
- [ ] documented the code

### DIFF
--- a/pdns/slavecommunicator.cc
+++ b/pdns/slavecommunicator.cc
@@ -743,7 +743,7 @@ void CommunicatorClass::slaveRefresh(PacketHandler *P)
         requeue.insert(di);
       }
       else {
-        g_log<<Logger::Debug<<"Got NOTIFY for "<<di.zone<<", going to check SOA serial, ourserial is "<<di.serial<<endl;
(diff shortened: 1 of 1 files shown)


pdns/slavecommunicator.cc
```diff
@@ -743,7 +743,7 @@ void CommunicatorClass::slaveRefresh(PacketHandler *P)
         requeue.insert(di);
       }
       else {
-        g_log<<Logger::Debug<<"Got NOTIFY for "<<di.zone<<", going to check SOA serial"<<endl;
+        g_log<<Logger::Debug<<"Got NOTIFY for "<<di.zone<<", going to check SOA serial, ourserial is "<<di.serial<<endl;
```
rgacogne: "ourserial is" -> "our serial is"

PR #6826, opened 2018-08-01
Make getOutstanding be usable from both lua and console

phonedph1:
### Short description
So that we can do sketchy things with it from maintenance and luarule actions.

### Checklist
I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdist-lua-bindings.cc
+++ b/pdns/dnsdist-lua-bindings.cc
@@ -105,7 +105,7 @@ void setupLuaBindings(bool client)
       g_pools.setState(localPools);
       s->pools.erase(pool);
     });
-  g_lua.registerFunction<void(DownstreamState::*)()>("getOutstanding", [](const DownstreamState& s) { g_outputBuffer=std::to_string(s.outstanding.load()); });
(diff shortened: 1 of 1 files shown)


pdns/dnsdist-lua-bindings.cc
```diff
@@ -105,7 +105,7 @@ void setupLuaBindings(bool client)
       g_pools.setState(localPools);
       s->pools.erase(pool);
     });
-  g_lua.registerFunction<void(DownstreamState::*)()>("getOutstanding", [](const DownstreamState& s) { g_outputBuffer=std::to_string(s.outstanding.load()); });
+  g_lua.registerFunction<int(DownstreamState::*)()>("getOutstanding", [](const DownstreamState& s) { return s.outstanding.load(); });
```
rgacogne: If you don't mind, please make the return value the same type than the `outstanding` member (`uint64_t`) otherwise we'll likely have conversion issues down the road.

phonedph1:
Right - thanks! Updated.

PR #6831, opened 2018-08-07
Get rid of some allocs/copies in DNS parsing

rgacogne:
### Short description

This PR:
- Prevents an allocation and copy of the `DNS` query or response passed to `MOADNSParser`, and get rid of the weird `-12` dance (auth, rec) ;
- Makes `PacketReader` accepts a `string` instead of a `vector<uint18_t>` so we don't need to allocate a vector and copy our data to it (auth, rec, dnsdist) ;
- Removes a useless copy of the response in `arecvfrom()` (rec) ;
- Removes some string allocations and copies during `EDNS` processing (dnsdist) ;
- replaces a fixed 65k buffer allocated for every `TCP` connection by a dynamically allocated one (rec).

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdist.cc
+++ b/pdns/dnsdist.cc
@@ -273,11 +273,12 @@ bool fixUpResponse(char** response, uint16_t* responseLen, size_t* responseSize,
   }
 
   if (ednsAdded || ecsAdded) {
-    char * optStart = NULL;
+    uint16_t optStart = NULL;
     size_t optLen = 0;
     bool last = false;
 
-    int res = locateEDNSOptRR(*response, *responseLen, &optStart, &optLen, &last);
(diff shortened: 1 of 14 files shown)


pdns/dnsdist.cc
```diff
@@ -273,11 +273,12 @@ bool fixUpResponse(char** response, uint16_t* responseLen, size_t* responseSize,
   }
 
   if (ednsAdded || ecsAdded) {
-    char * optStart = NULL;
+    uint16_t optStart = NULL;
```
chbruyand: Initializing this to `NULL` is confusing

rgacogne: Indeed!

PR #6843, opened 2018-08-10
check-all-zones: find duplicate zones and SOAs

### LINKED ISSUE #6829: 'pdns_control list-zones' zones prints zones out twice
- Program: Authoritative 
 - Issue type: Bug report 

### Short description

I noticed 'pdns_control list-zones' had couple of zones printed out twice. And good people on irc #powerdns (esp. ahu) suggested to see if these zones have more than one SOA record. And indeed they had. After deleting all but one, everything seems to be back to normal.

### Environment

 - Operating system: Ubuntu 16.04 amd64
 - Software version: 4.0.3-1pdns.xenial
 - Software source: repo.powerdns.com 

### Steps to reproduce

1. open master zone in edit-zone
2. add next SOA which is different for example by serial number
3. save
4. look at 'pdns_control list-zones'

### Expected behaviour

Actually i would expect that zone could contain exactly one SOA record (or at least 'pdns_control list-zones' shows the zone not twice).

### Actual behaviour

edit-zone permits to enter more than one SOA record and check-zone sees no error. 'pdns_control list-zones' shows zones twice.

### Other information

Since i use postgresql backend I looked also into postgresql database and saw that in general everything is ok but there are too several SOA entries found (but i didnt notice it until ahu on irc suggested it might have to do with it).

Also, since i use nsedit i have in zones this metadata

Metadata items: 
	SOA-EDIT	INCEPTION-INCREMENT
	SOA-EDIT-API	INCEPTION-INCREMENT

zeha:
### Short description
If you have duplicate zones, zone IDs or SOAs, check-all-zones will now produce errors like these:

duplicate SOA:
```
[Error] Another SOA for zone 'zeha1.example.org' (serial 2) has already been seen (serial 1).
[Error] Domain ID 99 of 'zeha1.example.org' in backend gsqlite3 has already been used by zone 'zeha1.example.org' in backend gsqlite3.
```

duplicate zone ID in different backends:
```
[Error] Another SOA for zone 'example.com' (serial 2847484148) has already been seen (serial 2847484148).
[Error] Domain ID 3 of 'example.com' in backend bind has already been used by zone 'example.com' in backend gsqlite3.
```

Fixes #6829 by means of adding checking in check-all-zones.
### Checklist
I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/pdnsutil.cc
+++ b/pdns/pdnsutil.cc
@@ -604,15 +604,41 @@ int checkAllZones(DNSSECKeeper &dk, bool exitOnError)
 {
   UeberBackend B("default");
   vector<DomainInfo> domainInfo;
+  struct name{};
+  struct id{};
+  multi_index_container<
+    DomainInfo,
+    indexed_by<
+      ordered_non_unique< member<DomainInfo,DNSName,&DomainInfo::zone>, CanonDNSNameCompare >,
+      ordered_non_unique< member<DomainInfo,uint32_t,&DomainInfo::id> >
+    >
+  > seenInfos;
+  auto& seenNames = seenInfos.get<0>();
+  auto& seenIds = seenInfos.get<1>();
 
   B.getAllDomains(&domainInfo, true);
   int errors=0;
   for(auto di : domainInfo) {
     if (checkZone(dk, B, di.zone) > 0) {
       errors++;
-      if(exitOnError)
-        return EXIT_FAILURE;
     }
+
+    auto seenName = seenNames.find(di.zone);
+    if (seenName != seenNames.end()) {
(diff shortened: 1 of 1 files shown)


pdns/pdnsutil.cc
```diff
@@ -604,15 +604,41 @@ int checkAllZones(DNSSECKeeper &dk, bool exitOnError)
 {
   UeberBackend B("default");
   vector<DomainInfo> domainInfo;
+  struct name{};
+  struct id{};
```
rgacogne: I'm probably missing something but I don't see why you declare these?

zeha: You're not missing anything, they were leftover tag structs, but then I got rid of their usage!

PR #6845, opened 2018-08-10
Drop api-readonly configuration setting

### LINKED ISSUE #6786: api-readonly design questions
 - Issue type: Design discussion

### Short description
#6781 points out that the `rectify-zone` API endpoint works even when `api-readonly` is set.  This is not the first time we 'missed a spot'.

Two proposals have been made informally:
1. when the request is not GET (i.e. it is POST/PUT/PATCH), and `api-readonly` is set, never even dispatch the call to the defined handlers
2. drop `api-readonly`

In the interest of reducing complexity, I would currently prefer option 2.

Thoughts?

zeha:
### Short description
Fixes #6786.

### Checklist
I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/recursordist/docs/settings.rst
+++ b/pdns/recursordist/docs/settings.rst
@@ -81,7 +81,8 @@ Static pre-shared authentication key for access to the REST API.
 ``api-readonly``
 ----------------
 .. versionadded:: 4.0.0
-.. versionremoved:: 4.2.0
+.. versionchanged:: 4.2.0
+  This setting has been removed.
 
 -  Boolean
 -  Default: no


pdns/recursordist/docs/settings.rst
```diff
@@ -81,6 +81,7 @@ Static pre-shared authentication key for access to the REST API.
 ``api-readonly``
 ----------------
 .. versionadded:: 4.0.0
+.. versionremoved:: 4.2.0
```
pieterlexis: `.. deprecated::`, there is not `versionremoved` in sphinx. Alternatively:

```rst
.. versionchanged:: 4.2.0
  This setting has been removed
```

franklouwers:
Docs are only updated for the recursor it seems, not for auth.

PR #6850, opened 2018-08-13
rec: Disable only our own TCP listening socket when reuseport is enabled

### LINKED ISSUE #6849: pdns_recursor: Tried to remove unlisted fd X from multiplexer
- Program: Recursor
 - Issue type: Bug report

### Short description

After 50 min of operation, pdns_recursor crashes with `STL Exception: Tried to remove unlisted fd 5 from multiplexer`

### Environment

 - Software version:  4.1.3 (edited from 4.1)
 - Software source:  PowerDNS repository

### Steps to reproduce
Config:
```
allow-from=127.0.0.0/8, 10.0.0.0/8, 100.64.0.0/10, 169.254.0.0/16, 192.168.0.0/16, 172.16.0.0/12, ::1/128, fc00::/7, fe80::/10, REMOVED
config-dir=/etc/powerdns
cpu-map=0=20 1=21 2=22 3=23 4=24 5=25 6=26 7=27 8=28 9=29 10=30 11=31 12=32 13=33 14=34 15=35 16=36 17=37 18=38 19=39 20=40 21=41 22=42 23=43 24=44 25=45
include-dir=/etc/powerdns/recursor.d/
local-address=REMOVED
local-port=53
pdns-distributes-queries=no
quiet=yes
reuseport=yes
security-poll-suffix=
setgid=pdns
setuid=pdns
threads=25
```
1. start
2. send queries (our test setup: 5000 queries/s)

### Expected behaviour
No crash

### Actual behaviour
crashes

### Other information
I suspect it kind of relates to the reuseport=yes, I see the error message for each thread (25 times)

I operate the pdns_recursor in a separate network namespace with its own network interface.

I now activated coredumps for the process, so when it crashes again I will attach the coredump.

rgacogne:
### Short description

Fixes #6849.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/pdns_recursor.cc
+++ b/pdns/pdns_recursor.cc
@@ -143,6 +143,11 @@ struct RecThreadInfo
     int readQueriesToThread{-1};
   };
 
+  /* FD corresponding to TCP sockets this thread is listening
+     on if reuseport is set. Otherwise g_tcpSockets is used instead.
+     These FDs are also in deferredAdds when we have one
+     socket per listener, and in g_deferredAdds instead. */
+  std::set<int> tcpSockets;
   /* FD corresponding to listening sockets if we have one socket per
      listener (with reuseport), otherwise all listeners share the
      same FD and g_deferredAdds is then used instead */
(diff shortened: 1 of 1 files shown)


pdns/pdns_recursor.cc
```diff
@@ -3490,24 +3496,26 @@ static int serviceMain(int argc, char*argv[])
...
+        makeTCPServerSockets(deferredAdds, tcpSockets);
       }
     }
   }
   else {
     /* we don't have reuseport so we can only open one socket per
        listening addr:port and everyone will listen on it */
     makeUDPServerSockets(g_deferredAdds);
-    makeTCPServerSockets(g_deferredAdds);
+    makeTCPServerSockets(g_deferredAdds, g_tcpSockets);
```
zeha: feels weird to me that in the "only one thread listens" case another var is used. but maybe I just didn't understand what is going on here...

rgacogne: It's tricky because if we have more than one thread handling incoming queries (distributor or worker when `pdns-distributes-queries` is off) but we don't have `reuseport`, we actually have several threads listening to the same socket, so it's not obvious which one owns it. Your remark actually made me spot what looks like a bug, though, so thank you :)

PR #6881, opened 2018-08-24
Keep track of what destinations are causing the most timeouts

phonedph1:
### Short description
Keep track of what downstreams are causing the most timeouts. Useful for troubleshooting network connectivity issues rather than trolling servfails and manual testing.

- [ ] not really remotes so should split that out in ws/api
- [ ] the naming timeouts could probably be improved, suggestions?

### Checklist
I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/pdns_recursor.cc
+++ b/pdns/pdns_recursor.cc
@@ -116,7 +116,7 @@ thread_local std::unique_ptr<MT_t> MT; // the big MTasker
 thread_local std::unique_ptr<MemRecursorCache> t_RC;
 thread_local std::unique_ptr<RecursorPacketCache> t_packetCache;
 thread_local FDMultiplexer* t_fdm{nullptr};
-thread_local std::unique_ptr<addrringbuf_t> t_remotes, t_servfailremotes, t_largeanswerremotes, t_bogusremotes, t_timeouts;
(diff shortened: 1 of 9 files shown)


phonedph1:
apparently accessing a TLS extern doesn't compile properly on old gcc or clang?

zeha:
> apparently accessing a TLS extern doesn't compile properly on old gcc or clang?

Note that linking the `testrunner` fails. I'd guess you need to add the actual `t_timeouts` variable into `test-syncrec_cc.cc`. Or move it outside of `pdns_recursor.cc`.

rgacogne:
> Note that linking the testrunner fails. I'd guess you need to add the actual t_timeouts variable into test-syncrec_cc.cc. Or move it outside of pdns_recursor.cc.

You are right. Since the new `t_timeouts` variable is used from `syncres.cc`, we need it to be defined in one of the files linked into the `testrunner` binary, and `pdns_recursor.cc` is not (and couldn't be since it defines `main()`, among other things). This is not an issue for the existing variables since they are only ever used from `pdns_recursor.cc` and `rec_channel_rec.cc`, which are not linked into `testrunner`.
IMHO the easiest way would be to move the definition of `t_timeouts` from `pdns_recursor.cc` into `syncres.cc`.

PR #6882, opened 2018-08-24
Dump the throttle map

phonedph1:
### Short description
Dumps the throttle map.

Probably not the cleanest method, open to suggestions and education on how to accomplish this better.

### Checklist
I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/syncres.cc
+++ b/pdns/syncres.cc
@@ -402,7 +402,7 @@ uint64_t SyncRes::doDumpThrottleMap(int fd)
   {
     count++;
     // remote IP, dns name, qtype, count, ttd
-    fprintf(fp, "%s\t%s\t%d\t%d\t%s", i.first.get<0>().toString().c_str(), i.first.get<1>().toString().c_str(), i.first.get<2>(), i.second.count, ctime(&i.second.ttd));
+    fprintf(fp, "%s\t%s\t%d\t%u\t%s", i.first.get<0>().toString().c_str(), i.first.get<1>().toString().c_str(), i.first.get<2>(), i.second.count, ctime(&i.second.ttd));
   }
   fclose(fp);
   return count;
--- a/pdns/syncres.cc
+++ b/pdns/syncres.cc
@@ -388,6 +388,26 @@ uint64_t SyncRes::doDumpNSSpeeds(int fd)
   return count;
 }
 
+uint64_t SyncRes::doDumpThrottleMap(int fd)
+{
+  FILE* fp=fdopen(dup(fd), "w");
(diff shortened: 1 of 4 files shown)


pdns/syncres.cc
```diff
@@ -388,6 +388,26 @@ uint64_t SyncRes::doDumpNSSpeeds(int fd)
...
+    return 0;
+  fprintf(fp, "; throttle map dump follows\n");
+  fprintf(fp, "; remote IP, DNSName, QType\n");
+  uint64_t count=0;
+
+  for(const auto& i : t_sstorage.throttle.getThrottleTuples())
+  {
+    count++;
+    // remote IP, dns name, qtype
+    fprintf(fp, "%s %s %d", i.get<0>().toLogString().c_str(), i.get<1>().toString().c_str(), i.get<2>());
```
rgacogne: It would be nice to have the content of the `entry` as well, the `TTD` and the `count` fields.

pdns/syncres.hh
```diff
@@ -120,10 +120,23 @@ public:
     return (unsigned int)d_cont.size();
   }
 
+  typedef std::vector<boost::tuple<ComboAddress,DNSName,uint16_t> > throttles_t;
+  throttles_t getThrottleTuples() {
```
rgacogne: I'm not sure we need to do a full copy there, we could probably get away with returning a const reference to the actual map, and it would give us the content of the map in addition to the keys.

phonedph1:
Thanks @rgacogne - I *think* this does what you suggest? Hopefully doesn't abuse ``auto`` too badly.

pdns/syncres.cc
```diff
@@ -388,6 +388,26 @@ uint64_t SyncRes::doDumpNSSpeeds(int fd)
...
+  fprintf(fp, "; throttle map dump follows\n");
+  fprintf(fp, "; remote IP\tqname\tqtype\tcount\tttd\n");
+  uint64_t count=0;
+
+  const auto& throttleMap = t_sstorage.throttle.getThrottleMap();
+  for(const auto& i : throttleMap)
+  {
+    count++;
+    // remote IP, dns name, qtype, count, ttd
+    fprintf(fp, "%s\t%s\t%d\t%d\t%s", i.first.get<0>().toString().c_str(), i.first.get<1>().toString().c_str(), i.first.get<2>(), i.second.count, ctime(&i.second.ttd));
```
rgacogne: The `count` field is an `unsigned int` so the second `%d` should be a `%u` here.

PR #6890, opened 2018-08-28
ixfrdist: Reduce memory usage and lock contention

rgacogne:
### Short description

* Avoid some copies during zone retrieval, reducing the memory usage
* Reserve memory in advance when possible to avoid reallocation
* Reduce lock contention and copy of zone information (records, diffs) when serving an actual query, by using shared pointers so we only need to hold the lock while retrieving or replacing a pointer, instead of having to hold it while copying the whole zone.   

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/ixfrdist.cc
+++ b/pdns/ixfrdist.cc
@@ -226,8 +226,9 @@ void updateThread(const string& workdir, const uint16_t& keep, const uint16_t& a
           loadZoneFromDisk(records, fname, domain);
         }
         std::lock_guard<std::mutex> guard(g_soas_mutex);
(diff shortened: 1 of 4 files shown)


rgacogne:
Note that the memory usage is still quite consequent for very large zones, for example a `RPZ` zone with 8M entries will take around 1.6 GB of memory, and twice that during an update since we need to have both the previous and new versions in memory at the same time. Since such a zone is mostly composed of `CNAME`s, I guess something like roughly 200 bytes per `CNAME` might make sense.

Issue #6892, opened 2018-08-28
Lua Records - ifurlup + selector improvements


blop:
- Program: Authoritative
 - Issue type: Feature request

### Short description

I just found out about the Lua records feature coming in the 4.2 version.
I build a new docker image from the master and it's working great ! Thank you!

I'm testing the ifurlup() function but I do have a few questions (they could apply to ifportup() too) : 

#1 - How do we define the time the UP/down status are cached ? What are the default values ?
#2 - Which HTTP status code are considered UP ? 2xx and 3xx ? It's not mentioned in the documentation.
#3 - Can we add an optional selector to define the behaviour when all units are down ? currently there is a random among all nodes, i'd like to have a 'none' selector that just returns nothing (NXDOMAIN)
#4 - Can we also add an 'all' selector that we could use to return all units that are available/up among the first group matched ? (multiple A values returned)

Thanks again.

blop:
I'm now able to perform my health checks and return all the available IP addresses from the first set with at least one available, or nothing if everything is down.

Example:
`
ifurlup('https://hostname/path, {{'1.1.1.1','1.1.1.2','1.1.1.3'},{'2.2.2.1','2.2.2.2','2.2.2.3'}}, {selector='all', defaultSelector='none'})
`

chbruyand:
An URL is currently considered UP if the HTTP response code is equal to 200 ([see minicurl.cc](https://github.com/PowerDNS/pdns/blob/7d625762bc3d250c2f96ebc86197f85657829764/pdns/minicurl.cc#L108))

PR #6894, opened 2018-08-29
Improved Lua records - Added all selector, and backupSelector fallbacks

blop:
### Short description

Fixes -3 and -4 of #6892

This PR adds a `backupSelector` parameter to the `ifportup` and `ifurlup` methods. This is used to be able to customize the selector used when no candidate is declared as UP (instead of the default `random` one). This also adds the `all` selector that can be used to return all candidates.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [X] documented the code

What do you think ? ;-)

### DIFF
--- a/docs/lua-records/functions.rst
+++ b/docs/lua-records/functions.rst
@@ -46,7 +46,7 @@ Record creation functions
 
   Various options can be set in the ``options`` parameter:
 
-  - ``selector``: used to pick the IP address from list of viable candidates. Choices include 'pickclosest', 'random', 'hashed'.
+  - ``selector``: used to pick the IP address from list of viable candidates. Choices include 'pickclosest', 'random', 'hashed', 'all', 'none'.
   - ``source``: Source IP address to check from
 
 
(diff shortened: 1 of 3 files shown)


blop:
Previous behaviour is preserved by default.

pieterlexis:
This PR is conflicted, would you mind rebasing on master?

blop:
Done ;-)

chbruyand:
I also have a last nit: what would you think of renaming `defaultSelector` into `backupSelector` ?

chbruyand:
It still requires a few regression tests that I'll push tomorrow.

chbruyand:
So, I added a quick test for the backupSelector.
I also removed the `none` selector as it can lead to weird behavior if used with DNSSEC. We could be returning `NOERROR` answers if none of the candidates is `UP` and still put A in the NSEC bitmap which might confuse validators.
We are trying to release the `auth-4.2.0-alpha` ASAP so we vdecided to postpone the addition of the `none` validator and discuss with you of your need for that.

blop:
Then I won't be able to use the new version :/

I need the none selector so that if no service replicas is up, the DNS server returns an empty response.

The whole idea is to return only values for replicas that pass the health checks.

Habbie:
> I need the none selector so that if no service replicas is up, the DNS server returns an empty response.

Sorry about dropping that part - it will interact badly with DNSSEC so we want to take a little more time figuring out how to do it right, but right now we want to get 4.2.0-alpha1 out the door. Please PR the none selector separately so we can discuss it. Thank you for your contribution!

blop:
I'll create a new PR when I have more time.

PR #6343, opened 2018-03-14
Dnsdist prometheus stats

giganteous:
### Short description
Rebase of the original PR#3935, which is disconnected, hence a new PR.
See discussion #4947, and more recently #6002 and #6088.

I think I'm terrible at C++, but here we go.

### Checklist

I have:
- [x] read the [CONTRIBUTING.md](https://github.com/PowerDNS/pdns/blob/master/CONTRIBUTING.md) document
- [x] compiled and tested this code
- [ ] included documentation (including possible behaviour changes)
- [x] documented the code
- [ ] added or modified regression test(s)
- [ ] added or modified unit test(s)

PR #6901, opened 2018-08-30
Added Prometheus stats support for dnsdist

pavel-odintsov:
### Short description

My patch based on https://github.com/PowerDNS/pdns/pull/6343 prepared by @giganteous 

I made number of changes for it:
- Ported code to current master branch
- Introduced class which stores meta information about metrics (metric type, description). I do not like idea about mixing atomic counters and meta fields with big text blobs and I divided them
- Slightly changed code formatting
- Changed endpoint to /metrics as suggested by @dannyk81 
- Tested code and it works quite well for my case

### Checklist
I have:
- [*] read the [CONTRIBUTING.md](https://github.com/PowerDNS/pdns/blob/master/CONTRIBUTING.md) document
- [*] compiled this code
- [*] tested this code
- [*] included documentation (including possible behaviour changes)
- [*] documented the code
- [] added or modified regression test(s)
- [] added or modified unit test(s)

Thank you!

### DIFF
--- a/pdns/dnsdist-web.cc
+++ b/pdns/dnsdist-web.cc
@@ -471,7 +471,7 @@ static void connectionThread(int sock, ComboAddress remote, string password, str
           }
           const string label = "{pool=\"" + poolName + "\"}";
           const std::shared_ptr<ServerPool> pool = entry.second;
-          output << "dnsdist_pools_servers" << label << " " << pool->countServers(false) << "\n";
+          output << "dnsdist_pool_servers" << label << " " << pool->countServers(false) << "\n";
 
           if (pool->packetCache != nullptr) {
             const auto& cache = pool->packetCache;
--- a/pdns/dnsdist-web.cc
+++ b/pdns/dnsdist-web.cc
@@ -128,7 +128,7 @@ static bool isAnAPIRequestAllowedWithWebAuth(const YaHTTP::Request& req)
 
 static bool isAStatsRequest(const YaHTTP::Request& req)
 {
-  return req.url.path == "/jsonstat" || req.url.path == "/prometheus";
+  return req.url.path == "/jsonstat" || req.url.path == "/metrics";
 }
 
 static bool compareAuthorization(const YaHTTP::Request& req, const string &expected_password, const string& expectedApiKey)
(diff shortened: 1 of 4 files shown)


pdns/dnsdist-web.cc
```diff
@@ -386,6 +386,120 @@ static void connectionThread(int sock, ComboAddress remote, string password, str
...
+        for (const auto& e : g_stats.entries) {
+          std::string metricName = std::get<0>(e);
+
+          // Prometheus suggest using '_' instead of '-'
+          std::string prometheusMetricName = "dnsdist_main_" + boost::replace_all_copy(metricName, "-", "_");
+
+          MetricDefinition metricDetails; 
+
+          if (!g_metricDefinitions.getMetricDetails(metricName, metricDetails)) {
+              warnlog("Do not have metric details for %s", metricName);
```
rgacogne: Perhaps this would be better as a `vinfolog()`? Granted, that would make it easier to miss it, but it would also prevent flooding the logs if we ever forget to add a metric.

pdns/dnsdist-web.cc
```diff
@@ -386,6 +386,120 @@ static void connectionThread(int sock, ComboAddress remote, string password, str
...
+            output << (*val)->load();
+          else if (const auto& dval = boost::get<double*>(&std::get<1>(e)))
+            output << **dval;
+          else
+            output << (*boost::get<DNSDistStats::statfunction_t>(&std::get<1>(e)))(std::get<0>(e));
+          
+          output << "\n";
+        }
+
+        const auto states = g_dstates.getCopy();
```
rgacogne: How about using `g_dstates.getLocal()` to get a shared pointer on the current version, instead of making a full copy?

pdns/dnsdist-web.cc
```diff
@@ -386,6 +386,120 @@ static void connectionThread(int sock, ComboAddress remote, string password, str
...
+              serverName = state->getName();
+
+          boost::replace_all(serverName, ".", "_");
+
+          const string label = "{server=\"" + serverName + "\"}";
+          output << statesbase << "queries"     << label << " " << state->queries.load()     << "\n";
+          output << statesbase << "drops"       << label << " " << state->reuseds.load()     << "\n";
+          output << statesbase << "latency"     << label << " " << state->latencyUsec/1000.0 << "\n";
+          output << statesbase << "senderrors"  << label << " " << state->sendErrors.load()  << "\n";
+          output << statesbase << "outstanding" << label << " " << state->outstanding.load() << "\n";
```
rgacogne: Perhaps the weight and order values might make sense as well?

pdns/dnsdist-web.cc
```diff
@@ -386,6 +386,120 @@ static void connectionThread(int sock, ComboAddress remote, string password, str
...
+
+          string frontName = front->local.toString() + ":" + std::to_string(front->local.getPort());
+          boost::replace_all(frontName, ".", "_");
+          string proto = (front->udpFD >= 0 ? "udp" : "tcp");
+
+          output << "dnsdist_main_frontend_queries{frontend=\"" << frontName << "\",proto=\"" << proto
+              << "\"} " << front->queries.load() << "\n";
+        }
+
+        const auto localPools = g_pools.getCopy();
```
rgacogne: Same, `g_pools.getLocal()` might make more sense.

pdns/dnsdist-web.cc
```diff
@@ -386,6 +386,120 @@ static void connectionThread(int sock, ComboAddress remote, string password, str
...
+          const string qnamebase = "dnsdist_querycount_queries";
+
+          for(auto &record: g_qcount.records) {
+            qname = record.first;
+            boost::replace_all(qname, ".", "_");
+	    
+            const std::string label = "{qname=\"" + qname + "\"}";
+            output << qnamebase << label << " " << record.second << "\n";
+          }
+          g_qcount.records.clear();
```
rgacogne: Are you sure you want to do that? Because right now the querycount stats are cleared after a carbon export, but with this change they would be cleared after a carbon export _and_ after an `API` query, which seems wrong. If you don't clear them then you only need a `ReadLock`, btw.

pdns/dnsdist.hh
```diff
@@ -262,7 +262,75 @@ struct DNSDistStats
   };
 };
 
+// Keeps additional information about metrics
+struct MetricDefinition {
+  MetricDefinition(std::string description, std::string prometheusType) {
```
rgacogne: It doesn't matter much but it could take const references to string instead of copying the strings. The fields could also be initialized in the member initializer list.

pdns/dnsdist.hh
```diff
@@ -262,7 +262,75 @@ struct DNSDistStats
...
+
+  // Metric description
+  std::string description;
+  // Metric type for Prometheus
+  std::string prometheusType;
+};
+
+struct MetricDefinitionStorage {
+  // Return metric definition by name
+  bool getMetricDetails(std::string metricName, MetricDefinition& metric) {
```
rgacogne: `metricName` should really be a `const` reference here. Also the body of the method seems to have some indentation issues.

pdns/dnsdist.hh
```diff
@@ -262,7 +262,75 @@ struct DNSDistStats
...
+    this->description = description;
+    this->prometheusType = prometheusType;
+  }
+ 
+  MetricDefinition() = default;
+
+  // Metric description
+  std::string description;
+  // Metric type for Prometheus
+  std::string prometheusType;
```
rgacogne: Given that we have a limited range of possible values (counter or gauge if I'm not mistaken), perhaps a enum class would be better suited?

pdns/dnsdist-web.cc
giganteous: Why is it reset after a carbon export? That will make these counters behave really weird, and graphs will show silly drops.

pdns/dnsdist-web.cc
pavel-odintsov: Good point! I changed code to vinfolog. Thank you!

pdns/dnsdist-web.cc
pavel-odintsov: Done! Thank you!

pdns/dnsdist-web.cc
pavel-odintsov: Done!

pdns/dnsdist.hh
pavel-odintsov: Done, good point. Copying std::string isn't a good idea.

pdns/dnsdist.hh
pavel-odintsov: I reworked it to scoped enum and added function to convert scoped enum value to string representation. Also, I fixed bug about misplaced type and help messages for Prometheus.

pdns/dnsdist-web.cc
```diff
@@ -386,6 +386,127 @@ static void connectionThread(int sock, ComboAddress remote, string password, str
...
+    else if (req.url.path == "/metrics") {
+        handleCORS(req, resp);
+        resp.status = 200;
+
+        std::ostringstream output;
+        for (const auto& e : g_stats.entries) {
+          std::string metricName = std::get<0>(e);
+
+          // Prometheus suggest using '_' instead of '-'
+          std::string prometheusMetricName = "dnsdist_main_" + boost::replace_all_copy(metricName, "-", "_");
```
pieterlexis: I think we only added "main" in the carbon output because metronome would not ingest it otherwise. If so (cc @ahupowerdns) I suggest we don't add the "main" to the prometheus output.

ahupowerdns: Let is not continue that hack indeed. Sorry :-)

pdns/dnsdist-web.cc
pavel-odintsov: Added weight and order

pdns/dnsdist-web.cc
pavel-odintsov: I removed clear() here and replaced WriteLock by ReadLock. Also, I tested dnsdist with setQueryCount(true) and it works as expected.

pdns/dnsdist-web.cc
pavel-odintsov: I removed "main" from all metric names and added up to date raw output for /metrics.

pavel-odintsov:
I think I fixed all issues discovered during review.

pdns/dnsdist-web.cc
```diff
@@ -386,6 +386,128 @@ static void connectionThread(int sock, ComboAddress remote, string password, str
...
+          }
+        }
+
+        {
+          ReadLock rl(&g_qcount.queryLock);
+          std::string qname;
+          const string qnamebase = "dnsdist_querycount_queries";
+
+          for(auto &record: g_qcount.records) {
+            qname = record.first;
```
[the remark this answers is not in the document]
pavel-odintsov: Yes, I noticed it too. I removed it completely. I used DNSTAP for such precise logging and can suggest it as workaround for per qname logging.

pdns/dnsdist-web.cc
```diff
@@ -386,6 +386,128 @@ static void connectionThread(int sock, ComboAddress remote, string password, str
...
+          output << statesbase << "order"       << label << " " << state->order              << "\n";
+          output << statesbase << "weight"      << label << " " << state->weight             << "\n";
+        }
+
+        for (const auto& front : g_frontends) {
+          if (front->udpFD == -1 && front->tcpFD == -1)
+            continue;
+
+          string frontName = front->local.toString() + ":" + std::to_string(front->local.getPort());
+          boost::replace_all(frontName, ".", "_");
```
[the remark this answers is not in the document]
pavel-odintsov: Removed replace_all.

pdns/dnsdist-web.cc
```diff
@@ -386,6 +386,128 @@ static void connectionThread(int sock, ComboAddress remote, string password, str
...
+          else if (const auto& dval = boost::get<double*>(&std::get<1>(e)))
+            output << **dval;
+          else
+            output << (*boost::get<DNSDistStats::statfunction_t>(&std::get<1>(e)))(std::get<0>(e));
+          
+          output << "\n";
+        }
+
+        auto states = g_dstates.getLocal();
+        const string statesbase = "dnsdist_servers_";
```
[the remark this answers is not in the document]
pavel-odintsov: Done!

pdns/dnsdist-web.cc
```diff
@@ -386,6 +386,128 @@ static void connectionThread(int sock, ComboAddress remote, string password, str
...
+        
+        for (const auto& state : *states) {
+          string serverName;
+           
+          if (state->name.empty())
+              serverName = state->remote.toString() + ":" + std::to_string(state->remote.getPort());
```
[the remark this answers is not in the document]
pavel-odintsov: Done, thank you for suggestion!

pdns/dnsdist-web.cc
```diff
@@ -386,6 +386,128 @@ static void connectionThread(int sock, ComboAddress remote, string password, str
...
+        for (const auto& entry : *localPools) {
+          string poolName = entry.first;
+          boost::replace_all(poolName, ".", "_");
+          
+          if (poolName.empty()) {
+            poolName = "_default_";
+          }
+          const string label = "{pool=\"" + poolName + "\"}";
+          const std::shared_ptr<ServerPool> pool = entry.second;
+          output << "dnsdist_pools_servers" << label << " " << pool->countServers(false) << "\n";
```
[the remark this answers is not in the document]
pavel-odintsov: Done, renamed.

pdns/dnsdist-web.cc
```diff
@@ -386,6 +386,128 @@ static void connectionThread(int sock, ComboAddress remote, string password, str
...
+          output << "dnsdist_frontend_queries{frontend=\"" << frontName << "\",proto=\"" << proto
+              << "\"} " << front->queries.load() << "\n";
+        }
+
+        auto localPools = g_pools.getLocal();
+        const string cachebase = "dnsdist_pool_";
+        
+        for (const auto& entry : *localPools) {
+          string poolName = entry.first;
+          boost::replace_all(poolName, ".", "_");
```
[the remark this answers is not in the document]
pavel-odintsov: Done!

pdns/dnsdist.hh
pavel-odintsov: Reworked in previous patch set.

pdns/dnsdist-web.cc
```diff
@@ -386,6 +386,128 @@ static void connectionThread(int sock, ComboAddress remote, string password, str
...
+          string serverName;
+           
+          if (state->name.empty())
+              serverName = state->remote.toString() + ":" + std::to_string(state->remote.getPort());
+          else
+              serverName = state->getName();
+
+          boost::replace_all(serverName, ".", "_");
+
+          const string label = "{server=\"" + serverName + "\"}";
```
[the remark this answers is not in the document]
pavel-odintsov: Good idea, implemented it. Also, I migrated old string processing code for boost::format.

pavel-odintsov:
I fixed all issues again. Also, I updated example endpoint output to match new code.

pavel-odintsov:
Thank your for reviewing it! :)

PR #6910, opened 2018-08-31
Add support for OpenSSL 1.1.1's ed25519 and ed448 for signing and verifying

pieterlexis:
### Short description
This implements algorithm 15 and 16 support in the OpenSSL signer. It will only be used if the (as of yet unreleased) OpenSSL 1.1.1 is used.

Please review carefully.

CC: @mind04 

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/m4/pdns_check_libcrypto_eddsa.m4
+++ b/m4/pdns_check_libcrypto_eddsa.m4
@@ -25,7 +25,7 @@ AC_DEFUN([PDNS_CHECK_LIBCRYPTO_EDDSA], [
   [AC_INCLUDES_DEFAULT
   #include <$ssldir/include/openssl/evp.h>])
 
-  AS_IF([test "$libcrypto_ed448" = "yes" -o "$libcrypto_ed448" = "yes"], [
+  AS_IF([test "$libcrypto_ed25519" = "yes" -o "$libcrypto_ed448" = "yes"], [
     AC_DEFINE([HAVE_LIBCRYPTO_EDDSA], [1], [define to 1 if OpenSSL EDDSA support is available.])
   ], [ : ])
 
--- a/m4/pdns_check_libcrypto_eddsa.m4
+++ b/m4/pdns_check_libcrypto_eddsa.m4
@@ -1,0 +1,36 @@
+AC_DEFUN([PDNS_CHECK_LIBCRYPTO_EDDSA], [
+  AC_REQUIRE([PDNS_CHECK_LIBCRYPTO])
+
+  # Set the environment correctly for a possibly non-default OpenSSL path that was found by/supplied to PDNS_CHECK_LIBCRYPTO
+  save_CPPFLAGS="$CPPFLAGS"
+  save_LDFLAGS="$LDFLAGS"
+  save_LIBS="$LIBS"
+
+  CPPFLAGS="$LIBCRYPTO_INCLUDES $CPPFLAGS"
+  LDFLAGS="$LIBCRYPTO_LDFLAGS $LDFLAGS"
+  LIBS="$LIBCRYPTO_LIBS $LIBS"
+
+  libcrypto_ed25519=no
+  libcrypto_ed448=no
+  AC_CHECK_DECLS([NID_ED25519], [
+    libcrypto_ed25519=yes
+    AC_DEFINE([HAVE_LIBCRYPTO_ED25519], [1], [define to 1 if OpenSSL ed25519 support is available.])
+  ], [ : ],
+  [AC_INCLUDES_DEFAULT
+  #include <$ssldir/include/openssl/evp.h>])
+  AC_CHECK_DECLS([NID_ED448], [
+    libcrypto_ed448=yes
+    AC_DEFINE([HAVE_LIBCRYPTO_ED448], [1], [define to 1 if OpenSSL ed448 support is available.])
+  ], [ : ],
+  [AC_INCLUDES_DEFAULT
+  #include <$ssldir/include/openssl/evp.h>])
+
+  AS_IF([test "$libcrypto_ed448" = "yes" -o "$libcrypto_ed448" = "yes"], [
+    AC_DEFINE([HAVE_LIBCRYPTO_EDDSA], [1], [define to 1 if OpenSSL EDDSA support is available.])
+  ], [ : ])
+
+  # Restore variables
+  CPPFLAGS="$save_CPPFLAGS"
+  LDFLAGS="$save_LDFLAGS"
+  LIBS="$save_LIBS"
+])
--- a/pdns/opensslsigners.cc
+++ b/pdns/opensslsigners.cc
@@ -26,6 +26,9 @@
 #ifdef HAVE_LIBCRYPTO_ECDSA
 #include <openssl/ecdsa.h>
 #endif
+#if defined(HAVE_LIBCRYPTO_ED25519) || defined(HAVE_LIBCRYPTO_ED448)
+#include <openssl/evp.h>
+#endif
 #include <openssl/sha.h>
 #include <openssl/rand.h>
 #include <openssl/rsa.h>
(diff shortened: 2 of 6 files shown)


pdns/opensslsigners.cc
```diff
@@ -910,6 +913,216 @@ void OpenSSLECDSADNSCryptoKeyEngine::fromPublicKeyString(const std::string& inpu
...
+  static std::shared_ptr<DNSCryptoKeyEngine> maker(unsigned int algorithm)
+  {
+    return std::make_shared<OpenSSLEDDSADNSCryptoKeyEngine>(algorithm);
+  }
+
+private:
+  size_t d_len;
+  int d_id;
+
+  EVP_PKEY *d_edkey = NULL;
```
rgacogne: I would rather see `nullptr` than `NULL` here, and in many places below.

pdns/opensslsigners.cc
```diff
@@ -910,6 +913,216 @@ void OpenSSLECDSADNSCryptoKeyEngine::fromPublicKeyString(const std::string& inpu
...
+  buf.resize(len);
+  EVP_PKEY_get_raw_private_key(d_edkey, reinterpret_cast<unsigned char*>(&buf.at(0)), &len);
+  storvect.push_back(make_pair("PrivateKey", buf));
+  return storvect;
+}
+
+std::string OpenSSLEDDSADNSCryptoKeyEngine::sign(const std::string& msg) const
+{
+  auto ctx = EVP_MD_CTX_new();
+  auto pctx = d_edctx;
```
rgacogne: I'm not sure what the point of `pctx` is?

pdns/opensslsigners.cc
```diff
@@ -910,6 +913,216 @@ void OpenSSLECDSADNSCryptoKeyEngine::fromPublicKeyString(const std::string& inpu
...
+  }
+  return signature;
+}
+
+bool OpenSSLEDDSADNSCryptoKeyEngine::verify(const std::string& msg, const std::string& signature) const
+{
+  auto ctx = EVP_MD_CTX_new();
+  auto pctx = d_edctx;
+  if(EVP_DigestVerifyInit(ctx, &pctx, NULL, NULL, d_edkey) < 1) {
+    throw runtime_error(getName()+" unable to initialize verifyer");
```
rgacogne: We are leaking `ctx`.

pdns/opensslsigners.cc
```diff
@@ -910,6 +913,216 @@ void OpenSSLECDSADNSCryptoKeyEngine::fromPublicKeyString(const std::string& inpu
...
+  }
+
+  string checkSignature = signature;
+  string checkMsg = msg;
+
+  auto r = EVP_DigestVerify(ctx,
+      reinterpret_cast<unsigned char*>(&checkSignature.at(0)), checkSignature.length(),
+      reinterpret_cast<unsigned char*>(&checkMsg.at(0)), checkMsg.length());
+  if (r < 0) {
+    throw runtime_error(getName()+" verification failure");
```
rgacogne: Still leaking `ctx`.

pdns/opensslsigners.cc
```diff
@@ -910,6 +913,216 @@ void OpenSSLECDSADNSCryptoKeyEngine::fromPublicKeyString(const std::string& inpu
...
+  }
+
```
rgacogne: Also leaking `ctx`.

pdns/opensslsigners.cc
```diff
@@ -910,6 +913,216 @@ void OpenSSLECDSADNSCryptoKeyEngine::fromPublicKeyString(const std::string& inpu
...
+        reinterpret_cast<unsigned char*>(&signature.at(0)), &siglen,
+        reinterpret_cast<unsigned char*>(&msgToSign.at(0)), msgToSign.length()) < 1) {
+    throw runtime_error(getName()+" signing error");
+  }
+  return signature;
+}
+
+bool OpenSSLEDDSADNSCryptoKeyEngine::verify(const std::string& msg, const std::string& signature) const
+{
+  auto ctx = EVP_MD_CTX_new();
```
rgacogne: We should check the return value of `EVP_MD_CTX_new()` against `nullptr`.

pdns/opensslsigners.cc
```diff
@@ -910,6 +913,216 @@ void OpenSSLECDSADNSCryptoKeyEngine::fromPublicKeyString(const std::string& inpu
...
+  auto pctx = d_edctx;
```
rgacogne: I don't think we need `pctx`, it seems that it's only written to by `EVP_DigestVerifyInit()` and we don't use it afterwards.

pdns/opensslsigners.cc
```diff
@@ -910,6 +913,216 @@ void OpenSSLECDSADNSCryptoKeyEngine::fromPublicKeyString(const std::string& inpu
...
+    throw std::runtime_error(getName() + " could not create key structure from private key");
+  }
+  // Replace the context for a new one with the key
+  EVP_PKEY_CTX_free(d_edctx);
+  d_edctx = EVP_PKEY_CTX_new(d_edkey, NULL);
+}
+
+void OpenSSLEDDSADNSCryptoKeyEngine::fromPublicKeyString(const std::string& content)
+{
+  const unsigned char* raw = (const unsigned char*)content.c_str();
```
rgacogne: `reinterpret_cast<const unsigned char*>(content.c_str());` ?

pdns/opensslsigners.cc
```diff
@@ -910,6 +913,216 @@ void OpenSSLECDSADNSCryptoKeyEngine::fromPublicKeyString(const std::string& inpu
...
+  const size_t inputLen = content.length();
+
+  if (inputLen < 1) {
+    throw runtime_error(getName()+" invalid input size for the public key");
```
rgacogne: Unsure what the point of this check is since we check it again below?

pdns/opensslsigners.cc
```diff
@@ -910,6 +913,216 @@ void OpenSSLECDSADNSCryptoKeyEngine::fromPublicKeyString(const std::string& inpu
...
+  }
+
+  int type{0};
+
+  if (inputLen == 32) {
+    type = NID_ED25519;
```
rgacogne: Same remark regarding the `HAVE_` than above.

pdns/opensslsigners.cc
```diff
@@ -910,6 +913,216 @@ void OpenSSLECDSADNSCryptoKeyEngine::fromPublicKeyString(const std::string& inpu
...
+      d_len = 32;
+      d_id = NID_ED25519;
+    } else if (d_algorithm == 16) {
+      d_len = 57;
+      d_id = NID_ED448;
+    } else {
+      throw runtime_error(getName()+" unknown algorithm "+std::to_string(d_algorithm));
+    }
+
+    d_edctx = EVP_PKEY_CTX_new_id(d_id, NULL);
```
rgacogne: From what I see, it doesn't look like we need a `EVP_PKEY_CTX` object except for generating a new key so perhaps we could just deal with it in `::create()` and be done with it afterwards?

m4/pdns_check_libcrypto_eddsa.m4
```diff
@@ -0,0 +1,36 @@
...
+  [AC_INCLUDES_DEFAULT
+  #include <$ssldir/include/openssl/evp.h>])
+  AC_CHECK_DECLS([NID_ED448], [
+    libcrypto_ed448=yes
+    AC_DEFINE([HAVE_LIBCRYPTO_ED448], [1], [define to 1 if OpenSSL ed448 support is available.])
+  ], [ : ],
+  [AC_INCLUDES_DEFAULT
+  #include <$ssldir/include/openssl/evp.h>])
+
+  AS_IF([test "$libcrypto_ed448" = "yes" -o "$libcrypto_ed448" = "yes"], [
```
mnordhoff: That checks `ed448` twice. Should one of them be `ed25519`?

pdns/opensslsigners.cc
```diff
@@ -910,6 +913,216 @@ void OpenSSLECDSADNSCryptoKeyEngine::fromPublicKeyString(const std::string& inpu
...
+    }
+
+    d_edkey = EVP_PKEY_new();
+    if (d_edkey == NULL) {
+      throw runtime_error(getName()+" allocation of key structure failed");
+    }
+
+    if(d_algorithm == 15) {
+      d_len = 32;
+      d_id = NID_ED25519;
```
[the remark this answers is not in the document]
pieterlexis: I think `HAVE_LIBCRYPTO_ED25519` and `HAVE_LIBCRYPTO_ED448` can go, ed25519 and ed448 support were added in the same commit in OpenSSL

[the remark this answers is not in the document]
rgacogne: Is it possible to compile with only one of them available?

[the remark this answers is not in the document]
rgacogne: (openssl, I mean)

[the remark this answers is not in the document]
pieterlexis: It does not looks like it from the `perl configdata.pm --dump`

pdns/opensslsigners.cc
```diff
@@ -910,6 +913,239 @@ void OpenSSLECDSADNSCryptoKeyEngine::fromPublicKeyString(const std::string& inpu
...
+};
+
+bool OpenSSLEDDSADNSCryptoKeyEngine::checkKey() const
+{
+  return (d_edkey != nullptr);
+}
+
+void OpenSSLEDDSADNSCryptoKeyEngine::create(unsigned int bits)
+{
+  auto pctx = EVP_PKEY_CTX_new(d_edkey, nullptr);
```
rgacogne: I think we should generate the context from the algorithm id:
```
auto pctx = EVP_PKEY_CTX_new_id(d_id, nullptr);
```

Otherwise I don't see how the correct type of key can be generated.

pdns/opensslsigners.cc
```diff
@@ -910,6 +913,239 @@ void OpenSSLECDSADNSCryptoKeyEngine::fromPublicKeyString(const std::string& inpu
...
+  if (pctx == nullptr) {
+    throw runtime_error(getName()+" context initialization failed");
+  }
+  if (EVP_PKEY_keygen_init(pctx) < 1) {
+    throw runtime_error(getName()+" keygen initialization failed");
```
rgacogne: I think we are leaking `pctx` here.

pdns/opensslsigners.cc
```diff
@@ -910,6 +913,239 @@ void OpenSSLECDSADNSCryptoKeyEngine::fromPublicKeyString(const std::string& inpu
...
+  }
+  if (EVP_PKEY_keygen(pctx, &d_edkey) < 1) {
+    throw runtime_error(getName()+" key generation failed");
```
rgacogne: Also here.

pdns/opensslsigners.cc
```diff
@@ -910,6 +913,239 @@ void OpenSSLECDSADNSCryptoKeyEngine::fromPublicKeyString(const std::string& inpu
...
+  }
```
rgacogne: We need to release `pctx` here.

pdns/opensslsigners.cc
```diff
@@ -910,6 +913,239 @@ void OpenSSLECDSADNSCryptoKeyEngine::fromPublicKeyString(const std::string& inpu
...
+  if (EVP_PKEY_get_raw_private_key(d_edkey, reinterpret_cast<unsigned char*>(&buf.at(0)), &len) < 1) {
+    throw runtime_error(getName() + " Could not get private key from d_edkey");
+  }
+  storvect.push_back(make_pair("PrivateKey", buf));
+  return storvect;
+}
+
+std::string OpenSSLEDDSADNSCryptoKeyEngine::sign(const std::string& msg) const
+{
+  auto pctx = EVP_PKEY_CTX_new(d_edkey, nullptr);
```
rgacogne: I don't think we need `pctx` at all here, `EVP_DigestSignInit()` will accept a `nullptr` instead.

pdns/opensslsigners.cc
```diff
@@ -910,6 +913,239 @@ void OpenSSLECDSADNSCryptoKeyEngine::fromPublicKeyString(const std::string& inpu
...
+    throw runtime_error(getName()+" signing error");
+  }
+  EVP_PKEY_CTX_free(pctx);
+  EVP_MD_CTX_free(mdctx);
+  return signature;
+}
+
+bool OpenSSLEDDSADNSCryptoKeyEngine::verify(const std::string& msg, const std::string& signature) const
+{
+  auto pctx = EVP_PKEY_CTX_new(d_edkey, nullptr);
```
rgacogne: Same here, I don't think `pctx` is needed.

pieterlexis:
Could use a new review

PR #6911, opened 2018-08-31
Dnsdist doh [DO NOT MERGE]

ahupowerdns:
### Short description
This PR adds DNS over HTTPS support to dnsdist, based on libh2o. 
### Checklist
I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code


johndickinson:
I had to upgrade to bionic (from trusty) to get libh2o-evloop and also had to specify LIBS=-lwslay to fix link errors like /usr/lib/gcc/x86_64-linux-gnu/7/../../../x86_64-linux-gnu/libh2o-evloop.so: undefined reference to `wslay_event_want_read’. The full configure line I used was ./configure --enable-dns-over-tls --enable-dns-over-https --with-libcrypto=/opt/openssl-1.1.1-pre9 --enable-re2 CXXFLAGS=-I/opt/openssl-1.1.1-pre9/include CPPFLAGS=-I/opt/openssl-1.1.1-pre9/include LDFLAGS=-L/opt/openssl-1.1.1-pre9/lib LIBS=-lwslay

pieterlexis:
@johndickinson,

What does `config.log` say about libh2o? And can you share the `libh2o-evloop.pc`? On arch linux, the `-lwslay` is not needed (and not in the .pc file). I _think_ this is because the arch PKGBUILD installs the library as a `.a` and Debian/ubuntu as a `.so`.

johndickinson:
config.log:
...
configure:20525: checking for LIBH2OEVLOOP
configure:20532: $PKG_CONFIG --exists --print-errors "libh2o-evloop"
configure:20535: $? = 0
configure:20549: $PKG_CONFIG --exists --print-errors "libh2o-evloop"
configure:20552: $? = 0
configure:20590: result: yes
...
ac_cv_env_LIBH2OEVLOOP_CFLAGS_set=
ac_cv_env_LIBH2OEVLOOP_CFLAGS_value=
ac_cv_env_LIBH2OEVLOOP_LIBS_set=
ac_cv_env_LIBH2OEVLOOP_LIBS_value=
...
pkg_cv_LIBH2OEVLOOP_CFLAGS=
pkg_cv_LIBH2OEVLOOP_LIBS=-lh2o-evloop
...
HAVE_LIBH2OEVLOOP_FALSE='#'
HAVE_LIBH2OEVLOOP_TRUE=''
...
LIBH2OEVLOOP_CFLAGS=''
LIBH2OEVLOOP_LIBS='-lh2o-evloop'
...
#define HAVE_LIBH2OEVLOOP 1


$ cat /usr/lib/x86_64-linux-gnu/pkgconfig/libh2o-evloop.pc
prefix=/usr
exec_prefix=${prefix}
includedir=${prefix}/include
libdir=/usr/lib/x86_64-linux-gnu

Name: libh2o-evloop
Description: An optimized HTTP/1.x & HTTP/2 library
URL: https://h2o.examp1e.net/
Version: 0.13.4
Libs: -L${libdir} -lh2o-evloop
Cflags: -I${includedir}

zeha:
Rgd -lwslay: https://sources.debian.org/src/h2o/2.2.5+dfsg1-6/debian/patches/link-against-wslay.patch/

pdns/dnsdistdist/doh.cc
```diff
@@ -0,0 +1,718 @@
...
+    }
+
+    string poolname;
+    int delayMsec = 0;
+    const uint16_t * flags = getFlagsFromDNSHeader(dh);
+    const uint16_t origFlags = *flags;
+    uint16_t qtype, qclass;
+    unsigned int consumed = 0;
+    DNSName qname(query, len, sizeof(dnsheader), false, &qtype, &qclass, &consumed);
+    DNSQuestion dq(&qname, qtype, qclass, consumed, &du->dest, &du->remote, dh, 1500, len, false, &queryRealTime);
```
rgacogne: The `1500` here is wrong, it pretends that there is at least 1500 bytes available from the address pointed to by `dh`, which is not true. Rules and `handleEDNSClientSubnet()` will very likely trigger an heap-based out-of-bounds write.

pdns/dnsdistdist/doh.cc
```diff
@@ -0,0 +1,718 @@
...
+      }
+    }
+
+    uint32_t cacheKey = 0;
+    if (packetCache && !dq.skipCache) {
+
+      uint32_t allowExpired = ss ? 0 : g_staleCacheEntriesTTL;
+      boost::optional<Netmask> subnet;
+      char cquery[1500];
+      memcpy(cquery, query, du->query.length());
```
rgacogne: I don't _think_ you need to copy the query here, unless I missing something, since either it will be overwritten by the cached response on a hit, or we won't use `cquery` again before leaving this block.

pdns/dnsdistdist/doh.cc
```diff
@@ -0,0 +1,718 @@
...
+    h2o_add_header(&du->req->pool, &du->req->res.headers, H2O_TOKEN_CONTENT_TYPE, NULL, H2O_STRLIT("application/dns-message"));
+    
+    //    struct dnsheader* dh = (struct dnsheader*)du->query.c_str();
+    //    cout<<"Attempt to send out "<<du->query.size()<<" bytes over https, TC="<<dh->tc<<", RCODE="<<dh->rcode<<", qtype="<<du->qtype<<", req="<<(void*)du->req<<endl;
+
+    if(du->ednsAdded) { // take it out again
+      vector<uint8_t> stripped;      
+      rewriteResponseWithoutEDNS(du->query, stripped);
+      du->query.assign((const char*)&stripped[0], stripped.size());
+    }
```
rgacogne: The case of an incoming query with `EDNS` but without `ECS` is not handled here.

rgacogne:
Unless I'm mistaken, we always send the request received over DoH to the backend over UDP, advertising an EDNS Payload Size of 4096 (unless there was an existing EDNS OPT RR). Doesn't that mean that we will never be able to handle responses larger than 4096 bytes over DoH?

MaxWichern:
Starting dnsdist one way results in an error, while another way it works fine.
"nohup dnsdist --supervised &" works without any issues.
"systemctl start dnsdist" results in:
"_dnsdist: what():  DOH thread failed to launch: DOH server failed to listen on 10.142.0.3:443: Permission denied
systemd: dnsdist.service: main process exited, code=killed, status=6/ABRT_"

The error message itself should be more specific.
And the error should not end with an abort.

rgacogne:
@MaxWichern It should have been fixed by 9bad969, is that not the case?

MaxWichern:
> @MaxWichern It should have been fixed by [9bad969](https://github.com/PowerDNS/pdns/commit/9bad969691fb930ffedeb62ac7e77d59b859ae1b), is that not the case?

That was the initial thought, but it didn't.

Habbie:
When #7676 gets merged, we need to extend that code to reload h2o certs as well.

rgacogne:
Superseded by #7726.

PR #6928, opened 2018-09-05
Add operator<<(long long) to be able to log time_t.

fobser:
On some systems time_t is defined as long long so that it is at least 64 bit on 32 bit and 64 bit architectures.

This prevents

slavecommunicator.cc:902:151: error: use of overloaded operator '<<' is
      ambiguous (with operand types 'Logger' and 'time_t' (aka 'long long'))

### DIFF
--- a/pdns/logger.cc
+++ b/pdns/logger.cc
@@ -222,6 +222,16 @@ Logger& Logger::operator<<(unsigned long i)
   return *this;
 }
 
+Logger& Logger::operator<<(long long i)
+{
+  ostringstream tmp;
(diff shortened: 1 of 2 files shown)


ahupowerdns:
before merging this, please know it has been attempted before and broke many other platforms. This is because there is overlap between long long, long, int etc. Perhaps it works these days.

omoerbeek:
I propose a better way to fix this: https://github.com/PowerDNS/pdns/pull/7316

pieterlexis:
fixed by #7316

PR #6935, opened 2018-09-06
Backup server type for dnsdist

pavel-odintsov:
### Short description

Implemented backup server type for dnsdist. dnsdist uses backup server only when all other servers are dead. It's not used during normal operations. Only single backup server allowed.

It supported and tested for wrandom and whashed policies only.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/pdns/dnsdist-lua.cc
+++ b/pdns/dnsdist-lua.cc
@@ -341,6 +341,10 @@ void setupLuaConfig(bool client)
 			  ret->mustResolve=boost::get<bool>(vars["mustResolve"]);
 			}
 
+			if(vars.count("backup")) {
+			  ret->backup=boost::get<bool>(vars["backup"]);
+			}
+
 			if(vars.count("useClientSubnet")) {
 			  ret->useECS=boost::get<bool>(vars["useClientSubnet"]);
 			}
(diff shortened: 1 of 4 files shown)


pavel-odintsov:
Added backup server support for "roundrobin" algorithm.

pavel-odintsov:
Added support for leastOutstanding and tested it.

pavel-odintsov:
Tested support for firstAvailable and added it to documentation (it was implemented in commit about leastOutstanding).

pavel-odintsov:
I added backup server support for all load balancing algorithms. Ready for review again.

pavel-odintsov:
I'm going to rework it to Action. I also do not like idea about hacking each load balancing algo.

pavel-odintsov:
I finished very early implementation using global action: https://github.com/Line-Rate-Development/pdns/commit/39164e25c71216e943254c06633f8183cd68dcf8

rgacogne:
What I would really love is to be able to pass a list of pool names instead of only one to `PoolAction()`, so that we can then try them in order and stop as soon as one of them returns a valid server.

pavel-odintsov:
Nice idea!

What about serve stale from cache feature? Should I try it for each pool before falling back to next one?

rgacogne:
> What about serve stale from cache feature? Should I try it for each pool before falling back to next one?

That's a very good question and a nice example of features interacting in weird ways.. I'd think so, yes.

pavel-odintsov:
I prepared improved version of this feature and will create new PR.

PR #6937, opened 2018-09-06
dnsdist: allow web API calls to trigger Lua code

Habbie:
### Short description
Given this config:

```
webserver('127.0.0.1:8083', 'secret', 'apisecret')
setAPIWritable(true)

function webcall(s) 
	print("got webcall", s)
	return string.format('you said: %s', s)
end
```

This works:

```
$ curl -X POST --data 'hi' -H 'X-API-Key: apisecret' http://127.0.0.1:8083/api/v1/servers/localhost/luahandler
you said: hi
```

Obviously this needs error handling, and perhaps some serialisation (how hard would it be to get JSON in and out without importing a JSON parser into Lua?).

For now I am posting this for discussion. The power and flexibility offered by this extension should be obvious; most API feature requests could be diverted to this feature until we flesh out a real API for the important functions. @rgacogne has called the idea 'a can of worms', and given the foot shooting ability we are offering here, it is obvious he is right.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdist-web.cc
+++ b/pdns/dnsdist-web.cc
@@ -161,6 +161,11 @@ static bool isMethodAllowed(const YaHTTP::Request& req)
       return true;
     }
   }
+  if (req.method == "POST" && g_apiReadWrite) {
+    if (req.url.path == "/api/v1/servers/localhost/luahandler") {
+      return true;
+    }
+  }
   return false;
 }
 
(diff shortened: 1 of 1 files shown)


pdns/dnsdist-web.cc
```diff
@@ -161,6 +161,11 @@ static bool isMethodAllowed(const YaHTTP::Request& req)
       return true;
     }
   }
+  if (req.method == "POST" && g_apiReadWrite) {
```
pavel-odintsov: Maybe you can add two version of lua hook? One for R/O API mode and one more for R/W API mode flag?

pieterlexis:
Closing for now, the usecase it small and a feature like this is easily abused

PR #6962, opened 2018-09-12
Store NetmaskTree nodes in a set for faster removal

rgacogne:
### Short description

The insertion is a bit slower as a result (~ +25%) but removal is much, much more faster for large sets as it was O(n) previously.
Walking all the entries might be a bit slower as well, but this change has no impact on the lookup speed, which is the critical point for the NMT.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/iputils.hh
+++ b/pdns/iputils.hh
@@ -632,11 +632,11 @@ public:
     return *this;
   }
 
-  const typename std::vector<node_type*>::const_iterator begin() const { return _nodes.begin(); }
-  const typename std::vector<node_type*>::const_iterator end() const { return _nodes.end(); }
+  const typename std::set<node_type*>::const_iterator begin() const { return _nodes.begin(); }
+  const typename std::set<node_type*>::const_iterator end() const { return _nodes.end(); }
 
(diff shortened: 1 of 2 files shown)


rgacogne:
This PR makes the `test_config_allow_from_replace` test in the `API` regression suite fail, because it looks like the `ACL` is not correctly updated and contains both `127.0.0.1/32` and `::1/128` when it should only contain the former. Since this PR makes the order not deterministic anymore, it fails half the time. We should investigate why the `ACL` is not updated and update the test to make sure that only one value is returned, instead of failing only if the first value is not the one we expect.

rgacogne:
It turns out that this test is broken because the `ACL` is read from the `acl.list` file, since `--allow-from-file` is passed on the command line, overriding `allow-from`.

```
Done parsing 2 allow-from ranges from file 'acl.list' - overriding 'allow-from' setting
```

rgacogne:
Rebased now that the `test_config_allow_from_replace` test has been fixed.

PR #6897, opened 2018-08-29
dnsdist: Refactor trailing data tests

gibson042:
### Short description
Improve dnsdist regression testing:
* DRY out TCP vs. UDP transport.
* Verify that trailing data makes it to the backend.
* Verify that `TrailingDataRule()` doesn't match queries without trailing data.

### Checklist
I have:
- [x] read the [CONTRIBUTING.md](https://github.com/PowerDNS/pdns/blob/master/CONTRIBUTING.md) document
- [x] compiled this code
- [x] tested this code
- [ ] included documentation (including possible behaviour changes)
- [x] documented the code
- [x] added or modified regression test(s)
- [ ] added or modified unit test(s)

PR #6967, opened 2018-09-13
dnsdist: Expose trailing data

### LINKED ISSUE #6846: dnsdist: ability to append data to packet
- Program: dnsdist
 - Issue type: Feature request

### Short description
This is a feature request that I would intend to implement myself, but want to get a general idea of whether this functionality would be accepted in a pull request before I go forward, so I can weigh other implementation paths I could take before starting.

### Usecase
The ability to pass additional environment information from `dnsdist` to downstreams that would otherwise not have that information, outside of standard DNS data to avoid invalidating TSIG hashes.

### Description
I would like to add the following functionality:

- Ability to append data to the end of the DNS packet before it's sent to a downstream from a rule and/or Lua

The use case here is that we would have a `dnsdist` instance with access to additional client information handling queries signed with TSIG, which needs to pass the additional client information on to downstreams we also control to be used in determining responses, without invalidating the TSIG hash. By appending the data to the end of the packet after the length of DNS data, it would normally be ignored by resolvers.

I recognize this is a pretty specific use case, so I'm trying to get an idea of whether we can add a generic enough and useful feature to `dnsdist` that would be accepted, but would meet our needs, before we go forward with deciding on how we proceed ourselves. Feedback and thoughts welcome.

gibson042:
### Short description
Expose data following a DNS message to Lua for both reading and writing.
Fixes #6846.

Incorporates #6897 ([diff from that base](https://github.com/gibson042/pdns/compare/2018-08-trailing-data-tests...2018-08-trailing-data)).

### Checklist
I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/pdns/dnsdist-lua-bindings-dnsquestion.cc
+++ b/pdns/dnsdist-lua-bindings-dnsquestion.cc
@@ -69,6 +69,21 @@ void setupLuaBindingsDNSQuestion()
       vector<uint8_t> tail(message + length, message + dq.len);
       return tail;
     });
+  g_lua.registerFunction<bool(DNSQuestion::*)(vector<pair<int, uint8_t>>)>("setTrailingData", [](DNSQuestion& dq, const vector<pair<int, uint8_t>>&data) {
+      uint8_t* message = reinterpret_cast<uint8_t*>(dq.dh);
+      const uint16_t length = getDNSPacketLength(reinterpret_cast<const char*>(message), dq.len);
+      if(length + data.size() > dq.size) {
+        return false;
+      }
+
+      /* Copy data from the Lua array, whose first index is 1 instead of 0. */
+      dq.len = length + data.size();
+      uint8_t* tail = message + length - 1;
+      for(const auto& pair : data) {
+        *(tail + pair.first) = pair.second;
+      }
+      return true;
+    });
 
   g_lua.registerFunction<void(DNSQuestion::*)(std::string)>("sendTrap", [](const DNSQuestion& dq, boost::optional<std::string> reason) {
 #ifdef HAVE_NET_SNMP
--- a/pdns/dnsdist-lua-bindings-dnsquestion.cc
+++ b/pdns/dnsdist-lua-bindings-dnsquestion.cc
@@ -63,24 +63,24 @@ void setupLuaBindingsDNSQuestion()
 
       return *dq.ednsOptions;
     });
-  g_lua.registerFunction<vector<uint8_t>(DNSQuestion::*)(void)>("getTrailingData", [](const DNSQuestion& dq) {
-      const uint8_t* message = reinterpret_cast<const uint8_t*>(dq.dh);
-      const uint16_t length = getDNSPacketLength(reinterpret_cast<const char*>(message), dq.len);
-      vector<uint8_t> tail(message + length, message + dq.len);
+  g_lua.registerFunction<std::string(DNSQuestion::*)(void)>("getTrailingData", [](const DNSQuestion& dq) {
+      const char* message = reinterpret_cast<const char*>(dq.dh);
+      const uint16_t messageLen = getDNSPacketLength(message, dq.len);
+      const std::string tail = std::string(message + messageLen, dq.len - messageLen);
       return tail;
     });
-  g_lua.registerFunction<bool(DNSQuestion::*)(vector<pair<int, uint8_t>>)>("setTrailingData", [](DNSQuestion& dq, const vector<pair<int, uint8_t>>&data) {
-      uint8_t* message = reinterpret_cast<uint8_t*>(dq.dh);
-      const uint16_t length = getDNSPacketLength(reinterpret_cast<const char*>(message), dq.len);
-      if(length + data.size() > dq.size) {
+  g_lua.registerFunction<bool(DNSQuestion::*)(std::string)>("setTrailingData", [](DNSQuestion& dq, const std::string& tail) {
+      char* message = reinterpret_cast<char*>(dq.dh);
+      const uint16_t messageLen = getDNSPacketLength(message, dq.len);
+      const uint16_t tailLen = tail.size();
+      if(messageLen + tailLen > dq.size) {
         return false;
       }
 
-      /* Copy data from the Lua array, whose first index is 1 instead of 0. */
-      dq.len = length + data.size();
-      uint8_t* tail = message + length - 1;
-      for(const auto& pair : data) {
-        *(tail + pair.first) = pair.second;
+      /* Update length and copy data from the Lua string. */
+      dq.len = messageLen + tailLen;
+      if(tailLen > 0) {
+        tail.copy(message + messageLen, tailLen);
       }
       return true;
     });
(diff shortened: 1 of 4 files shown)


regression-tests.dnsdist/test_Trailing.py
```diff
@@ -40,35 +87,257 @@ def testTrailingAllowed(self):
...
+    addLuaAction("echoed.trailing.tests.powerdns.com.", reportTrailingData)
+
+    function replaceTrailingData(dq)
+        local success = dq:setTrailingData({65, 66, 67}) -- "ABC"
+        if not success then
+            return DNSAction.ServFail, ""
+        end
+        return DNSAction.None, ""
+    end
+    addLuaAction("replaced.trailing.tests.powerdns.com.", replaceTrailingData)
```
rgacogne: This line looks like a c/p mistake.

pdns/dnsdist-lua-bindings-dnsquestion.cc
```diff
@@ -54,6 +54,27 @@ void setupLuaBindingsDNSQuestion()
...
+      const uint16_t length = getDNSPacketLength(reinterpret_cast<const char*>(message), dq.len);
+      if(length + data.size() > dq.size) {
+        return false;
+      }
+
+      /* Copy data from the Lua array, whose first index is 1 instead of 0. */
+      dq.len = length + data.size();
+      uint8_t* tail = message + length - 1;
+      for(const auto& pair : data) {
+        *(tail + pair.first) = pair.second;
```
rgacogne: I'm probably paranoid but I'd prefer the use of a counter instead of using the indexes passed by `Lua`.

regression-tests.dnsdist/test_Trailing.py
gibson042: No, the idea is that queries for replaced.trailing.tests.powerdns.com will be subject first to `replaceTrailingData` (which sets the data to ASCII "ABC") and then to `reportTrailingData` (which uses trailing data to spoof a response), allowing us to verify that the first succeeded.

pdns/dnsdist-lua-bindings-dnsquestion.cc
gibson042: I was also wondering about that... but is the order of data from Lua tables guaranteed? And if not, should I just guard the write like e.g.
```cpp
int size = data.size();
for(const auto& pair : data) {
  if(pair.first < 1 || pair.first > size) {
    /* reject sparse input */
    return false;
  }
  *(tail + pair.first) = pair.second;
}
```

regression-tests.dnsdist/test_Trailing.py
rgacogne: ah, right, I missed that! Thanks for the explanation!

pdns/dnsdist-lua-bindings-dnsquestion.cc
rgacogne: To be fair if the order is not right I'd consider it a bug in the wrapper. Still I'd like some sort of check, yes, but we need to be careful because we already increased `dq.len` at that point.

gibson042: So what kind of protection would you like to see?

gibson042: @rgacogne would it make more sense to follow what #6068 is doing with EDNS option data and expose the trailing data as a string rather than a uint8 list? It would work as long as 0x00 is allowed and strings are sufficiently manipulable in Lua, and would resolve the ordering concern.

rgacogne: Sorry it took me so long to answer this comment :-/
Yes, we have been using binary-safe strings in dnsdist for this kind of things so far, so IMHO it would be a better choice. Again, thanks a lot for contributing.

gibson042:
@rgacogne I updated this PR to expose the data as a string rather than an array of bytes, and added tests for non-ASCII octets in the trailing data (both inbound and outbound). This should be ready for final review.

pdns/dnsdist-lua-bindings-dnsquestion.cc
```diff
@@ -63,6 +63,27 @@ void setupLuaBindingsDNSQuestion()
...
+      const char* message = reinterpret_cast<const char*>(dq.dh);
+      const uint16_t messageLen = getDNSPacketLength(message, dq.len);
+      const std::string tail = std::string(message + messageLen, dq.len - messageLen);
+      return tail;
+    });
+  g_lua.registerFunction<bool(DNSQuestion::*)(std::string)>("setTrailingData", [](DNSQuestion& dq, const std::string& tail) {
+      char* message = reinterpret_cast<char*>(dq.dh);
+      const uint16_t messageLen = getDNSPacketLength(message, dq.len);
+      const uint16_t tailLen = tail.size();
+      if(messageLen + tailLen > dq.size) {
```
rgacogne: Not a big deal but it looks like `messageLen` + `tailLen` could overflow before being promoted. Perhaps something like:
```suggestion
      if(tailLen > (dq.size - messageLen)) {
```

gibson042: Nice catch! Updated.

regression-tests.dnsdist/dnsdisttests.py
```diff
@@ -142,42 +142,54 @@ def _ResponderIncrementCounter(cls):
...
-            request = dns.message.from_wire(data, ignore_trailing=ignoreTrailing)
-            response = cls._getResponse(request, fromQueue, toQueue)
-
+            forceRcode = None
+            try:
+                request = dns.message.from_wire(data, ignore_trailing=ignoreTrailing)
+            except dns.message.TrailingJunk as e:
+                if trailingDataResponse is False:
+                    raise
+                print("UDP query with trailing data, synthesizing response")
```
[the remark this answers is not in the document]
gibson042: The existing structure was already a bit confusing, and I was trying to preserve it as much as possible (which unfortunately didn't help). But the idea is to extend the existing `ignoreTrailing` parameter into `trailingDataResponse` as follows:
* `False`: reject queries with trailing data (_preexisting behavior_)
* `True`: ignore trailing data after queries (_preexisting behavior_)
* anything else: interpret as RCODE to be sent in response to queries with trailing data (_new behavior, used by `TestTrailingDataToBackend` in regression-tests.dnsdist/test_Trailing.py_)
We will never pass `synthesize=True` to `_getResponse`, because `forceRcode` is only set to `trailingDataResponse` when dns.message.from_wire raises dns.message.TrailingJunk, which cannot happen when it is ignoring trailing data.

I have added comments and safety-enforcing code.

gibson042:
@rgacogne Why was dnsdist 1.3.3 released including [DNSQuestion:getEDNSOptions()](https://github.com/PowerDNS/pdns/commit/ef8640ddc0fcff7c84b955f2ec2110939140ca2e) but not these additions?

rgacogne:
> Why was dnsdist 1.3.3 released including DNSQuestion:getEDNSOptions() but not these additions?

The main reason is that we are having a hard time deciding whether we want to make it possible to use trailing data within dnsdist. I personally don't like the idea too much, but if people are going to do it anyway I'd prefer the feature to be merged and be done with it.

gibson042:
> if people are going to do it anyway I'd prefer the feature to be merged and be done with it.

I really wish you would do that. Remember that this feature also provides a means of [protecting downstream servers _from_ trailing data](https://github.com/PowerDNS/pdns/pull/6967/files#diff-f25095fe509924812e11dd26ab8ea6b6R179):

```lua
function removeTrailingData(dq)
    local success = dq:setTrailingData("")
    if not success then
        return DNSAction.ServFail, ""
    end
    return DNSAction.None, ""
end
```

rgacogne:
Sorry it took so long, thanks a lot for this pull request!

PR #6969, opened 2018-09-14
a clean re-do of an earlier PR to add the glorious log-log histograms

ahupowerdns:
log-log histograms from https://blog.powerdns.com/2017/11/02/dns-performance-metrics-the-logarithmic-percentile-histogram/

As a side effect, 'dnsscope' no longer uses MOADNSParser which makes it both faster and more resilient. There are other speedups as well.

It also removes the confusing word 'bogus' from the statistics, which we used in a non-DNSSEC context.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code


pdns/histog.hh
```diff
@@ -0,0 +1,105 @@
+#pragma once
+#ifdef L
```
pieterlexis: We now have `g_log` `L` is dead and gone

pdns/dnsscope.cc
```diff
@@ -257,85 +278,91 @@ try
...
 	    qd.d_answercount++;
 
 	    if(qd.d_qcount) {
 	      uint32_t usecs= (pr.d_pheader.ts.tv_sec - qd.d_firstquestiontime.tv_sec) * 1000000 +  
 		(pr.d_pheader.ts.tv_usec - qd.d_firstquestiontime.tv_usec) ;
-	      //            cout<<"Took: "<<usecs<<"usec\n";
-	      if(usecs<2049000)
-		cumul[usecs]++;
-	      else
-		reallylate++;
```
phonedph1: This still prints ```0 responses (0.0000%) older than 2 seconds``` now since we never seem to increment it.

Should that part be removed all together (or add the tracking back in)?

PR #6979, opened 2018-09-19
rec: Authority records in AA=1 CNAME answer are authoritative

rgacogne:
### Short description

The records other than the `CNAME` for the initial target in `ANSWER` are not, nor are the `ADDITIONAL` ones, but `AUTHORITY` records are.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/syncres.cc
+++ b/pdns/syncres.cc
@@ -2117,7 +2117,7 @@ RCode::rcodes_ SyncRes::updateCacheFromRecords(unsigned int depth, LWResult& lwr
        set the TC bit solely because these RRSIG RRs didn't fit."
     */
     bool isAA = lwr.d_aabit && i->first.place != DNSResourceRecord::ADDITIONAL;
-    if (isAA && isCNAMEAnswer && (i->first.place != DNSResourceRecord::ANSWER || i->first.type != QType::CNAME || i->first.name != qname)) {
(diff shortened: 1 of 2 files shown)


rgacogne:
@pieterlexis Ping? :-)

pdns/syncres.cc
```diff
@@ -2117,7 +2117,7 @@ RCode::rcodes_ SyncRes::updateCacheFromRecords(unsigned int depth, LWResult& lwr
        set the TC bit solely because these RRSIG RRs didn't fit."
     */
     bool isAA = lwr.d_aabit && i->first.place != DNSResourceRecord::ADDITIONAL;
-    if (isAA && isCNAMEAnswer && (i->first.place != DNSResourceRecord::ANSWER || i->first.type != QType::CNAME || i->first.name != qname)) {
+    if (isAA && isCNAMEAnswer && i->first.place == DNSResourceRecord::ANSWER && (i->first.type != QType::CNAME || i->first.name != qname)) {
```
pieterlexis: Code change looks good (as it seems to do what you describe). Perhaps a comment here and multi-lining the statement makes sense?

rgacogne: Multi-lining might be a good idea indeed!
I briefly pondered adding a comment, but it would basically be "AUTHORITY records in AA=1 answers are authoritative", which is well, kind of obvious when you put it that way :) Also this test is already in the middle of 20 lines of comments, so... :)

PR #6984, opened 2018-09-21
rec: Delay the creation of RPZ threads until we have dropped privileges

Habbie:
This is a backport of #6792.

On Linux/glibc, calling `set*id()` from a thread results in the other
threads being sent the `SIGRT_1` signal so they are aware that they
should switch credentials too, because `POSIX` requires that all threads
use the same credentials but Linux actually handles it per thread.
The reception of the signal interrupts the current `syscall` with
`EINTR`, causing the loading of the `RPZ` zone to fail.

(cherry picked from commit e6ec15bfe4c391a51eab7c51c38307c7e009768f)

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


Habbie:
Builds should appear at https://builder.powerdns.com/#/builders/9/builds/2314

PR #6986, opened 2018-09-25
dnsdist: Add warning rates to dynBlockRulesGroup rules

### LINKED ISSUE #6907: dnsdist: support several dynBlockRules of the same type with different actions
- Program: dnsdist 
 - Issue type: Feature request 

### Short description

It would be very useful to be able to insert more than one dynBlockRule of the same type (Query rate, rcode rate, response byte rate...) but with different thresholds and actions.

### Usecase

That would allow generating an alert without blocking any query over a certain threshold, while blocking over a different one.

### Description

```
dbr:setQueryRate(50, 10, "Exceeded query rate", 60, DNSAction.NoOp)
dbr:setQueryRate(100, 10, "Exceeded query rate", 60, DNSAction.Drop)
```

rgacogne:
### Short description

After thinking a bit more about the needs described in #6907, I could not find a meaningful scenario where we want different actions for different rates except for the warning case. So for now I took the easy route of just adding a warning threshold, and we can improve on it later if we find that we need more than that.
Closes #6907.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/pdns/dnsdist-dynblocks.hh
+++ b/pdns/dnsdist-dynblocks.hh
@@ -44,9 +44,6 @@ private:
 
     DynBlockRule(const std::string& blockReason, unsigned int blockDuration, unsigned int rate, unsigned int warningRate, unsigned int seconds, DNSAction::Action action): d_blockReason(blockReason), d_blockDuration(blockDuration), d_rate(rate), d_warningRate(warningRate), d_seconds(seconds), d_action(action), d_enabled(true)
     {
-      if (d_warningRate > 0) {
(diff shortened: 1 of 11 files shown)


rgacogne:
> It may be nice to add a test where only the warning threshold is enabled.

This is actually not handled, since you can do that by using the regular threshold and a `NoOp` action, instead of using the warning threshold. Do you think it would make sense to handle that? We could make it work by passing a `0` value to the regular `rate`, but I'm not sure it would be very intuitive..

PR #7023, opened 2018-10-04
Clarify meaning of dnsdist statistics

ahupowerdns:
The various statistics kept by dnsdist do add up to a coherent picture, this PR documents how.

The wording leaves a little bit of wiggle room that this is how it 'should' be. If we find that it isn't like this, the documentation should be leading.

### DIFF
--- a/pdns/dnsdistdist/docs/statistics.rst
+++ b/pdns/dnsdistdist/docs/statistics.rst
@@ -10,9 +10,21 @@ dnsdist keeps statistics on the queries is receives and send out. They can be ac
 - via Carbon / Graphite / Metronome export (see :doc:`../guides/carbon`)
 - via SNMP (see :doc:`../advanced/snmp`)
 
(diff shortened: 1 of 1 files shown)


pdns/dnsdistdist/docs/statistics.rst
```diff
@@ -10,9 +10,21 @@ dnsdist keeps statistics on the queries is receives and send out. They can be ac
 - via Carbon / Graphite / Metronome export (see :doc:`../guides/carbon`)
 - via SNMP (see :doc:`../advanced/snmp`)
 
+To make sense of the statistics, the following relation should hold:
+
+	queries - noncompliant-queries
```
rgacogne: I think this should be `queries - noncompliant-queries - empty-queries`

rgacogne: I was wrong :-)

PR #7030, opened 2018-10-05
dnsdist did not set TCP_NODELAY, causing needless latency

ahupowerdns:
Winfried and others noted that our TLS responses were super slow. This turned out to be caused by the Nagle
algorithm which we failed to disable. https://en.wikipedia.org/wiki/Nagle%27s_algorithm

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [X] documented the code

### DIFF
--- a/pdns/dnsdist-tcp.cc
+++ b/pdns/dnsdist-tcp.cc
@@ -716,7 +716,7 @@ void* tcpAcceptorThread(void* p)
         continue;
       }
 #endif
-      SSetsockopt(ci->fd, SOL_TCP, TCP_NODELAY, 1); // disable NAGLE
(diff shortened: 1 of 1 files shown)


pdns/dnsdist-tcp.cc
```diff
@@ -715,7 +716,7 @@ void* tcpAcceptorThread(void* p)
         continue;
       }
 #endif
-
+      SSetsockopt(ci->fd, SOL_TCP, TCP_NODELAY, 1); // disable NAGLE
```
rgacogne: Just a nit but we have a nice `setTCPNoDelay()` function in `misc.hh` :-) It also uses `IPPROTO_TCP` instead of `SOL_TCP`, which is more portable.

PR #7039, opened 2018-10-08
Speedup DNSName canonical compare with two-sided algorithm

ahupowerdns:
DNSName canonical compare currently performs a<b and b<a tests to find equality.
With this change, it does both in one pass. This 'should' be a speedup.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsname.hh
+++ b/pdns/dnsname.hh
@@ -154,6 +154,23 @@ private:
 size_t hash_value(DNSName const& d);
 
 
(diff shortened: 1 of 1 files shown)


ahupowerdns:
As far as I can tell this makes no CPU impact whatsoever. This is interesting though. Apparently we spend all the time not on this part.

PR #7056, opened 2018-10-11
rec: remove the old dnssec root key, goodbye 19036 ....

mind04:
### Checklist
I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/root-dnssec.hh
+++ b/pdns/root-dnssec.hh
@@ -22,5 +22,4 @@
 
 #pragma once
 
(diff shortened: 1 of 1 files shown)


paulehoffman:
This PR should either be rejected or put on hold until the old trust anchor has been revoked.

mind04:
Please enlighten me why is this bad for new installations. The old key will never be in use again and caching is clearly not an issue.

paulehoffman:
The new key can be rolled back up until the old key is revoked. There is no expectation that this will happen, but it could.

PR #7059, opened 2018-10-12
Recursor: Allow reading root DNSSEC keys from file

### LINKED ISSUE #3530: Please support easy way of using distro-supplied root data
It'd be great if root hints and key/anchor could easily be used from a distro-supplied file (makes updating those easier without rebuilding all the dns software a distro ships).

Ideally I'd like to just symlink two files in /usr/share/pdns-recursor to some centrally provided files.

(Debian bug #760470)

pieterlexis:
### Short description
This PR adds the `readTrustAnchorsFromFile` function to the recursor's Lua config. This will allow us to read the Debian-provided `/usr/share/dns/root.keys` file.

It also renames `addDS` to `addTA` and `clearDS` to `clearTA` and documents some missing config functions.

Fixes #3530

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


pdns/rec-lua-conf.cc
```diff
@@ -389,6 +411,39 @@ void loadRecursorLuaConfig(const std::string& fname, luaConfigDelayedThreads& de
         lci.negAnchors.clear();
     });
 
+  Lua.writeFunction("readTrustAnchorsFromFile", [&lci](const std::string& fname) {
+      warnIfDNSSECDisabled("Warning: reading Trust Anchors from file (readTrustAnchorsFromFile), but dnssec is set to 'off'!");
+      auto zp = ZoneParserTNG(fname);
```
rgacogne: Perhaps this should be moved inside the `try..` block so the message states that the error, if any, is about a `TA` file?

pieterlexis:
Added an additional feature that makes the housekeeping thread reload the Trust Anchors evey day (by default). Needs new approval.

pdns/pdns_recursor.cc
```diff
@@ -2516,9 +2516,16 @@ static void doStats(void)
...
 static void houseKeeping(void *)
 {
-  static thread_local time_t last_rootupdate, last_prune, last_secpoll;
+  static thread_local time_t last_rootupdate, last_prune, last_secpoll, last_trustAnchorUpdate;
   static thread_local int cleanCounter=0;
   static thread_local bool s_running;  // houseKeeping can get suspended in secpoll, and be restarted, which makes us do duplicate work
+
+  auto luaconfsLocal = g_luaconfs.getLocal();
+  if (!luaconfsLocal->trustAnchorFileInfo.second.empty() && luaconfsLocal->trustAnchorFileInfo.first != 0) {
+    // Loading the config updated this, so we fake the update time here
```
rgacogne: I feel like I'm clearly missing something, but isn't `last_trustAnchorUpdate` reset every time we enter `houseKeeping`?

pdns/pdns_recursor.cc
```diff
@@ -2516,9 +2516,16 @@ static void doStats(void)
...
+    last_trustAnchorUpdate = time_t(nullptr) + luaconfsLocal->trustAnchorFileInfo.first * 3600;
```
rgacogne: Perhaps we could use `now.tv_sec` instead of `time_t(nullptr)`.

pdns/pdns_recursor.cc
```diff
@@ -2571,8 +2578,18 @@ static void houseKeeping(void *)
           g_log<<Logger::Error<<"Exception while performing security poll"<<endl;
         }
       }
+
+      if (!luaconfsLocal->trustAnchorFileInfo.second.empty() && luaconfsLocal->trustAnchorFileInfo.first != 0 &&
+          now.tv_sec - last_trustAnchorUpdate >= (luaconfsLocal->trustAnchorFileInfo.first * 3600)) {
+        g_log<<Logger::Debug<<"Refreshing Trust Anchors from file"<<endl;
+        try {
+          updateTrustAnchorsFromFile(luaconfsLocal->trustAnchorFileInfo.second);
```
rgacogne: Shouldn't `last_trustAnchorUpdate` be updated?

pdns/rec-lua-conf.hh
```diff
@@ -47,6 +47,7 @@ public:
   LuaConfigItems();
   SortList sortlist;
   DNSFilterEngine dfe;
+  std::pair<uint32_t, std::string> trustAnchorFileInfo; // Used to update the Trust Anchors from file periodically
```
rgacogne: I would prefer the use of a struct, or perhaps two separate members if that's easier than making a new struct, because otherwise I can't read the code without having to loop up what `first` and `second` are.

pdns/pdns_recursor.cc
```diff
@@ -2571,8 +2578,24 @@ static void houseKeeping(void *)
           g_log<<Logger::Error<<"Exception while performing security poll"<<endl;
         }
       }
+
+      if (!luaconfsLocal->trustAnchorFileInfo.fname.empty() && luaconfsLocal->trustAnchorFileInfo.interval != 0 &&
+          g_now.tv_sec - last_trustAnchorUpdate >= (luaconfsLocal->trustAnchorFileInfo.interval * 3600)) {
+        g_log<<Logger::Debug<<"Refreshing Trust Anchors from file"<<endl;
+        try {
+          map<DNSName, dsmap_t> dsAnchors;
```
rgacogne: We might want to document that, when this feature is enabled, any `TA` added manually via `rec_control add-ta` will get overridden after a while.

pieterlexis: done!

PR #7060, opened 2018-10-12
dnsdist: Release memory on DNS over TLS handshake failure

rgacogne:
### Short description

Otherwise we leak the memory allocated to handle the `TLS` connection if the handshake doesn't succeed correctly.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdistdist/tcpiohandler.cc
+++ b/pdns/dnsdistdist/tcpiohandler.cc
@@ -246,6 +246,8 @@ public:
     }
 
     if (!SSL_set_fd(d_conn, d_socket)) {
+      SSL_free(d_conn);
+      d_conn = nullptr;
       throw std::runtime_error("Error assigning socket");
     }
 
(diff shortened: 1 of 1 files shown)


rgacogne:
We should really use smart pointers there, although `OpenSSL` structures being opaque makes it a bit harder.

pdns/dnsdistdist/tcpiohandler.cc
```diff
@@ -246,19 +246,30 @@ class OpenSSLTLSConnection: public TLSConnection
...
 
     int res = 0;
     do {
       res = SSL_accept(d_conn);
       if (res < 0) {
-        handleIORequest(res, timeout);
+        try {
+          handleIORequest(res, timeout);
+        }
+        catch(const std::exception&) {
```
zeha: maybe this should catch `(...)` instead, to also get `PDNSExceptions`?

rgacogne: Done, thanks!

PR #7064, opened 2018-10-12
 dnsdist: Wrap GnuTLS and OpenSSL pointers in smart pointers

rgacogne:
### Short description

This PR is based on #7060 but wraps the naked pointers returned by `GnuTLS` and `OpenSSL` in smart pointers instead of carefully tracking them.
It needs a more serious review, though.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


rgacogne:
> I don't know if this is a significant delay or if this should be considered for dnsdist.

This might be, but I'm not sure how we could do better by managing that ourselves? Calling `SSL_CTX_flush_sessions()` will block anyone from accessing the session store anyway, even if we called it from a different thread, so the only thing we could do is to call it less often.
IMHO we can ignore it for now, until it happens to be an issue.

zeha:
> This might be, but I'm not sure how we could do better by managing that ourselves? Calling `SSL_CTX_flush_sessions()` will block anyone from accessing the session store anyway, even if we called it from a different thread, so the only thing we could do is to call it less often.

Right. I was thinking calling it from the maintenance thread would at least give consistent behaviour. But I agree it's probably a good idea to delay this for now!

PR #7070, opened 2018-10-16
pdns-recursor: avoid a memory leak in catch-all exception handler

r-bk:
This commit prevents a leak of DNSComboWriter in the catch-all exception
handler.


rgacogne:
Out of curiosity, did you experience this leak? At a quick glance it looks like we catch all exceptions making sense explicitly, but of course you never know.

r-bk:
> Out of curiosity, did you experience this leak? At a quick glance it looks like we catch all exceptions making sense explicitly, but of course you never know.

@rgacogne Yes, I did experience it. 
Preparing another PR to handle these exceptions not in this catch-all handler.

PR #7081, opened 2018-10-18
rec: extend the validity period of signatures by a number of seconds

mind04:
### Short description
There are some weird, maybe even broken, dnssec implementations out there. This pull is a workaround for the sign with inception is NOW problem. This is done by a loadbalancer vendor with a letter and a number in it's name. This is bad because only a very small time difference between signer and validator is enough for a domain to go bogus from time to time. 
This pull is adding an option to extent the signature validity by a number of seconds at inception and expire. This will increase the resilience of the validation process and avoid a number of hard to explain bogus results for high profile domains. The default is to add 60 seconds to the signature lifetime. If you think you need to increase this value check your clock. Ntp is your friend...

### Checklist
I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/recursordist/docs/settings.rst
+++ b/pdns/recursordist/docs/settings.rst
@@ -1096,6 +1096,15 @@ Query example (where 192.0.2.14 is your server):
 PowerDNS can change its user and group id after binding to its socket.
 Can be used for better :doc:`security <security>`.
 
+.. _setting-signature-inception-skew:
+
+``signature-inception-skew``
+----------------------------------
+-  Integer
+-  Default: 60
+
+Allow the signture inception to be off by this number of seconds. Negative values are not allowed.
+
 .. _setting-single-socket:
 
 ``single-socket``
--- a/pdns/recursordist/docs/settings.rst
+++ b/pdns/recursordist/docs/settings.rst
@@ -1100,6 +1100,8 @@ Can be used for better :doc:`security <security>`.
 
 ``signature-inception-skew``
 ----------------------------------
+.. versionadded:: 4.1.5
+
 -  Integer
 -  Default: 60
 
--- a/pdns/recursordist/docs/settings.rst
+++ b/pdns/recursordist/docs/settings.rst
@@ -1105,7 +1105,11 @@ Can be used for better :doc:`security <security>`.
 -  Integer
 -  Default: 60
 
-Allow the signture inception to be off by this number of seconds. Negative values are not allowed.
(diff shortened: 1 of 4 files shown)


pdns/recursordist/docs/settings.rst
```diff
@@ -460,6 +460,15 @@ If set, all hostnames in the `export-etc-hosts`_ file are loaded in canonical fo
 So an entry called 'pc' with ``export-etc-hosts-search-suffix='home.com'`` will lead to the generation of 'pc.home.com' within the recursor.
 An entry called 'server1.home' will be stored as 'server1.home', regardless of this setting.
 
+.. _setting-extend-signature-validity:
+
+``extend-signature-validity``
+----------------------------------
+-  Integer
```
rgacogne: A `versionadded:: 4.2.0` (for example) would be nice

mnordhoff:
If you do backport this to 4.1, I'd suggest having it off by default. (And turning it on by default in 4.2.)

This is weird and paranoid, but people could have compliance reasons for being unable to accept expired RRSIGs, and people are less careful about reading the changelogs of point releases.

(A CA using a different resolver got in trouble because this feature was on by default. But they just made a mistake -- it was always on by default, they didn't miss a change.)

mind04:
We are talking about 60 seconds to compensate for clock skew, not hours or days (like in other software).

When your validation process is this critical and compliance is that important, you should always read change-logs. If you fail to do so, don't blame software for a more sane mode of operation. For normal users this small number of seconds will prevent some insanely hard to debug validation failures. And by doing so it will save many hours of debugging and speed up the dnssec adoption as a whole. Something CAs should love...

I'm against a switched off default in 4.1.x, because this will mean people will choose a shitty value for this option when they hit this problem. And this value stay in there config for years.

In my opinion this is one of those pulls where “The Needs of the Many Outweigh the Needs of the Few”.

Or maybe we should just release 4.2.0 ;)

Habbie:
> (A CA using a different resolver got in trouble because this feature was on by default. But they just made a mistake -- it was always on by default, they didn't miss a change.)

Do you have a reference for this? It sounds juicy.

phonedph1:
Looks to be this  https://bugzilla.mozilla.org/show_bug.cgi?id=1398427 for anyone else interested.

Habbie:
The 4.1.x version of this (with default '0') was merged. Changed milestone back to 4.2.0.

PR #7087, opened 2018-10-19
dnsdist: Allow NoRecurse for use in dynamic blocks or lua rules

phonedph1:
### Short description
We can do NoRecurseAction() but not similar actions within a Lua rule nor as a dynamic block action.

This is potentially silly, please just close if it doesn't make sense to be doing - notably it's somewhat confusing because the query is put into the query ring before actions so it makes things like grepq look a bit weird.

### Checklist
I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


rgacogne:
PR is now conflicted, though :-/

phonedph1:
Yeah I am not sure it makes much sense, and was considering just closing this. I think a separate PR to set rd=0 as part of the lua actions will still make sense though.

PR #7088, opened 2018-10-22
rec: Add support for more than one protobuf server

rgacogne:
### Short description

So protobuf messages can be exported to several servers. At the moment we support only using the same exact options (timeouts, log queries, messages or both, ...) for all servers.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/pdns/lwres.cc
+++ b/pdns/lwres.cc
@@ -50,9 +50,9 @@
 
 #ifdef HAVE_PROTOBUF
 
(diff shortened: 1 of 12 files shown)


rgacogne:
> As the ownership of the list of protobuf servers is already shared via shared_ptr, is there any reason to also share the ownership of the protobuf server instances themselves ?

`RemoteLogger` objects are not copyable or movable since they hold a condition variable, a mutex and a thread, so we need to use pointers to be able to store them in a container like a `vector`, which might get resized. I don't want to deal with naked pointers, but you are right, we don't seem to need to be able to share the ownership so I just pushed a commit using unique pointers instead. Thanks!

rgacogne:
Rebased to fix a conflict with #7094.

PR #7094, opened 2018-10-23
Unique DNS Responses

neilcook:
### Short description
This PR extends the work on Newly Observed Domains by adding support for Unique DNS Responses; these are never-seen-before triplets of (qname, qtype, RR(rrname, rrtype, rrdata)). These are used by folks to detect potential evilness, such as fast-flux DNS.

Completely new to this PR is the ability to pass the NOD and UDR status via Protobuf. These are only set in the Response Protobuf packet. Protobuf responses from the packet cache have these flags cleared, as we only want to indicate this once. 

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/contrib/ProtobufLogger.py
+++ b/contrib/ProtobufLogger.py
@@ -114,6 +114,7 @@ class PDNSPBConnHandler(object):
                 if rr.HasField('class'):
                     rrclass = getattr(rr, 'class')
                 rrtype = rr.type
+                rrudr = rr.udr
                 if (rrclass == 1 or rrclass == 255) and rr.HasField('rdata'):
                     if rrtype == 1:
                         rdatastr = socket.inet_ntop(socket.AF_INET, rr.rdata)
@@ -122,11 +123,12 @@ class PDNSPBConnHandler(object):
                     elif rrtype == 28:
                         rdatastr = socket.inet_ntop(socket.AF_INET6, rr.rdata)
 
-                print("\t - %d, %d, %s, %d, %s" % (rrclass,
+                print("\t - %d, %d, %s, %d, %s, %d" % (rrclass,
                                                    rrtype,
                                                    rr.name,
                                                    rr.ttl,
-                                                   rdatastr))
+                                                   rdatastr,
+                                                   rrudr))
 
     def printSummary(self, msg, typestr):
         datestr = datetime.datetime.fromtimestamp(msg.timeSec).strftime('%Y-%m-%d %H:%M:%S')
@@ -168,9 +170,10 @@ class PDNSPBConnHandler(object):
 
         deviceId = binascii.hexlify(bytearray(msg.deviceId))
         requestorId = msg.requestorId
+        nod = msg.newlyObservedDomain
 
         print('[%s] %s of size %d: %s%s -> %s (%s), id: %d, uuid: %s%s '
-                  'requestorid: %s deviceid: %s serverid: %s' % (datestr,
+                  'requestorid: %s deviceid: %s serverid: %s nod: %d' % (datestr,
                                                     typestr,
                                                     msg.inBytes,
                                                     ipfromstr,
@@ -182,7 +185,8 @@ class PDNSPBConnHandler(object):
                                                     initialrequestidstr,
                                                     requestorId,
                                                     deviceId,
-                                                    serveridstr))
+                                                    serveridstr,
+                                                    nod))
 
     def getRequestorSubnet(self, msg):
         requestorstr = None
--- a/contrib/ProtobufLogger.py
+++ b/contrib/ProtobufLogger.py
@@ -111,10 +111,12 @@ class PDNSPBConnHandler(object):
             for rr in response.rrs:
                 rrclass = 1
                 rdatastr = ''
+                rrudr = 0
                 if rr.HasField('class'):
                     rrclass = getattr(rr, 'class')
                 rrtype = rr.type
-                rrudr = rr.udr
+                if rr.HasField('udr'):
+                    rrudr = rr.udr
                 if (rrclass == 1 or rrclass == 255) and rr.HasField('rdata'):
                     if rrtype == 1:
                         rdatastr = socket.inet_ntop(socket.AF_INET, rr.rdata)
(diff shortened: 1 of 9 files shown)


contrib/ProtobufLogger.py
```diff
@@ -114,6 +114,7 @@ def printResponse(self, message):
                 if rr.HasField('class'):
                     rrclass = getattr(rr, 'class')
                 rrtype = rr.type
+                rrudr = rr.udr
```
rgacogne: Doesn't this generate an error if the RR does not have a `udr` field set, which might happen since it's optional?

contrib/ProtobufLogger.py
```diff
@@ -168,9 +170,10 @@ def printSummary(self, msg, typestr):
 
         deviceId = binascii.hexlify(bytearray(msg.deviceId))
         requestorId = msg.requestorId
+        nod = msg.newlyObservedDomain
```
rgacogne: Same question here.

pdns/pdns_recursor.cc
```diff
@@ -949,6 +955,25 @@ static void nodAddDomain(const DNSName& dname)
...
+  bool ret = false;
+  if (record.d_place == DNSResourceRecord::ANSWER ||
+      record.d_place == DNSResourceRecord::ADDITIONAL) {
+    // Create a string that represent a triplet of (qname, qtype and RR[type, name, content])
+    std::stringstream ss;
+    ss << dname.toDNSStringLC() << ":" << qtype <<  ":" << qtype << ":" << record.d_type << ":" << record.d_name.toDNSStringLC() << ":" << record.d_content->getZoneRepresentation();
+    if (t_udrDBp && t_udrDBp->isUniqueResponse(ss.str())) {
+      if (g_udrLog) {
+	// This should also probably log to a dedicated file. 
+	g_log<<Logger::Notice<<"Unique response observed: qname="<<dname.toLogString()<<" qtype="<<QType(qtype).getName()<< " rrtype=" << QType(record.d_type).getName() << " rrname=" << record.d_name.toLogString() << " rrcontent=" << record.d_content->getZoneRepresentation() << endl;
```
rgacogne: Indentation seems off

pdns/pdns_recursor.cc
```diff
@@ -1426,7 +1462,17 @@ static void startDoResolve(void *p)
       pbMessage->setQueryTime(dc->d_now.tv_sec, dc->d_now.tv_usec);
       pbMessage->setRequestorId(dq.requestorId);
       pbMessage->setDeviceId(dq.deviceId);
+#ifdef NOD_ENABLED
+      if (g_nodEnabled) {
+        if (nodCheckNewDomain(dc->d_mdp.d_qname))
```
rgacogne: It looks like you won't have `NOD` without `protobuf`, I don't think you want that!

pdns/pdns_recursor.cc
```diff
@@ -4080,6 +4139,11 @@ int main(int argc, char **argv)
     ::arg().set("new-domain-lookup", "Perform a DNS lookup newly observed domains as a subdomain of the configured domain")="";
     ::arg().set("new-domain-history-dir", "Persist new domain tracking data here to persist between restarts")=string(NODCACHEDIR)+"/nod";
     ::arg().set("new-domain-whitelist", "List of domains (and implicitly all subdomains) which will never be considered a new domain")="";
+    ::arg().set("new-domain-db-size", "Size of the DB used to track new domains in terms of number of cells. Defaults to 67108864")="67108864";
+    ::arg().set("unique-response-tracking", "Track unique responses (tuple of query name, type and RR).")="no";
+    ::arg().set("unique-response-log", "Log unique responses")="yes";
+    ::arg().set("unique-response-history-dir", "Persist unique response tracking data here to persist between restarts")=string(NODCACHEDIR)+"/udr";
+    ::arg().set("unique-response-db-size", "Size of the DB used to track unique responses in terms of number of cells. Defaults to 67108864")="67108864";
```
rgacogne: We should get a few lines of documentation about those new settings.

rgacogne:
Rebasing on top of master after https://github.com/PowerDNS/pdns/pull/7095 has been merged should be enough to make Travis happy.

contrib/ProtobufLogger.py
neilcook: Good point

contrib/ProtobufLogger.py
neilcook: yep

pdns/pdns_recursor.cc
neilcook: No, indeed!

pdns/pdns_recursor.cc
neilcook: Yep

neilcook:
Still need to add the tags for NOD and UDR

pdns/nod.cc
```diff
@@ -121,6 +112,59 @@ void NODDB::setCacheDir(const std::string& cachedir)
   }
 }
 
+// Dump the SBF to a file
+// To spend the least amount of time inside the mutex, we dump to an
+// intermediate stringstream, otherwise the lock would be waiting for
+// file IO to complete
+bool PersistentSBF::snapshotCurrent(std::thread::id tid)
```
chbruyand: Maybe use a const reference for the arg?

pdns/nod.cc
```diff
@@ -121,6 +112,59 @@ void NODDB::setCacheDir(const std::string& cachedir)
...
+    else {
+      g_log<<Logger::Warning<<"NODDB snapshot: Cannot write file: " << f.string() << endl;
+    }
+  }
+  return false;
+}
+
+// NODDB Implementation
+
+void NODDB::housekeepingThread(std::thread::id tid)
```
chbruyand: Same here

pdns/nod.cc
```diff
@@ -173,42 +213,24 @@ void NODDB::addDomain(const std::string& domain)
...
-      }
-    }
-    else {
-      g_log<<Logger::Warning<<"NODDB snapshot: Cannot write file: " << f.string() << endl;
+void UniqueResponseDB::addResponse(const std::string& response)
+{
+  d_psbf.add(response);
+}
+
+void UniqueResponseDB::housekeepingThread(std::thread::id tid)
```
chbruyand: And there

rgacogne:
To be fair, according to C++11 `thread::id is a lightweight, trivially copyable class` so it might not matter much.

neilcook:
BTW I can't merge as I don't have permission

PR #7111, opened 2018-10-28
Rename skipLabel function to skipDomainName

gibson042:
### Short description
Rename the `skipLabel` function to `skipDomainName`, since that's what it actually does. I found the parser code very confusing until I figured this out.

### Checklist
I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/pdns/dnsparser.cc
+++ b/pdns/dnsparser.cc
@@ -582,7 +582,7 @@ public:
     : d_packet(packet), d_length(length), d_notyouroffset(12), d_offset(d_notyouroffset)
   {}
   
-  void skipLabel()
+  void skipDomainName()
   {
     uint8_t len; 
     while((len=get8BitInt())) { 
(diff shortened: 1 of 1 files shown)


pdns/dnsparser.cc
```diff
@@ -582,7 +582,7 @@ class DNSPacketMangler
     : d_packet(packet), d_length(length), d_notyouroffset(12), d_offset(d_notyouroffset)
   {}
   
-  void skipLabel()
+  void skipDomainName()
```
pieterlexis: It would make sense to add a docblock above the function as well, so the next person who stumbles unto this does not have to figure out what this function _exactly_ does, regardless of the name

pieterlexis: [This](https://github.com/PowerDNS/pdns/blob/50294c448e545f7ba33756e2db1003be075b608f/pdns/syncres.cc#L154-L166) and [this](https://github.com/PowerDNS/pdns/blob/50294c448e545f7ba33756e2db1003be075b608f/pdns/syncres.cc#L1042-L1048) should give some guidance

gibson042: Thanks; done.

Issue #6936, opened 2018-09-06
roundrobin policy does not respect server status


pavel-odintsov:
- Program: dnsdist
 - Issue type: Bug report/

### Short description

roundrobin policy does not respect server status

### Environment

 - Operating system:  Debian
 - Software version: Jessie
 - Software source: compiled yourself

### Steps to reproduce

1. I created configuration like this:
```
setServerPolicy(roundrobin)

-- Unbound Upper instances
newServer({address="1.1.1.1:53", name="1_1_1_1", retries=1, pool="" })
newServer({address="1.0.0.1:53", name="1_0_0_1", retries=1, pool="" })
newServer({address="8.8.8.8:53", name="backup", backup=true})
```

2. By default, all these servers marked as Up:
 ```
./dnsdist --client  --config /etc/dnsdist/dnsdist.conf
> showServers()
#   Name                 Address                       State     Qps    Qlim Ord Wt    Queries   Drops Drate   Lat Outstanding Pools
0   1_1_1_1              1.1.1.1:53                       up     0.0       0   1  1          0       0   0.0   0.0           0 
1   1_0_0_1              1.0.0.1:53                       up     0.0       0   1  1          0       0   0.0   0.0           0 
2   backup               8.8.8.8:53                       up     0.0       0   1  1          0       0   0.0   0.0           0 
All                                                              0.0                         0       0       
```
3. I made few queries to them
4. I explicitly marked all servers as down:
```
getServer(0):setDown()
getServer(1):setDown()
getServer(2):setDown()
```
5. I run few queris to dnsdist:
```
dig @127.0.0.1 -p1153 cloudflare.com; sleep 1; dig @127.0.0.1 -p1153 yandex.ru; sleep 2; dig @127.0.0.1 -p1153 habr.com
```
6. dnsdist queried all my downstreams ignoring their status:
```
showServers()
#   Name                 Address                       State     Qps    Qlim Ord Wt    Queries   Drops Drate   Lat Outstanding Pools
0   1_1_1_1              1.1.1.1:53                     DOWN     0.0       0   1  1          1       0   0.0   0.0           0 
1   1_0_0_1              1.0.0.1:53                     DOWN     0.0       0   1  1          1       0   0.0   0.0           0 
2   backup               8.8.8.8:53                     DOWN     1.0       0   1  1          1       0   0.0   0.1           0 
All                                                              0.0                         3       0                         
> 
```
 
### Expected behaviour

I expect SERVFAIL. 

[...]

PR #7137, opened 2018-11-02
Added option to specify multiple backup pools for any pool for dnsdist

pavel-odintsov:
### Short description

It's second attempt to implement backup pool. Previous one was: https://github.com/PowerDNS/pdns/pull/6935 

I Implemented option to specify any number of backup pools for specific pool.

It can be enabled using following Lua commands
```
addPoolBackupPool("", "backup1")
addPoolBackupPool("", "backup2")
showPoolBackupPool("")
```

dnsdist uses backup server only when all other servers in specified pool are dead. It's not used during normal operations.

All load balancing policies supported. But be careful with round robin: https://github.com/PowerDNS/pdns/issues/6936 

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/pdns/dnsdist-lua.cc
+++ b/pdns/dnsdist-lua.cc
@@ -1477,6 +1477,20 @@ void setupLuaConfig(bool client)
       g_pools.setState(localPools);
     });
 
+  g_lua.writeFunction("addPoolBackupPool", [](string pool, string backupPool) {
+      setLuaSideEffect();
+      auto localPools = g_pools.getCopy();
+      addBackupPool(localPools, pool, backupPool);
+      g_pools.setState(localPools);
+    });
+
+  g_lua.writeFunction("showPoolBackupPool", [](string pool) {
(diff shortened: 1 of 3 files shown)


pdns/dnsdist-lua.cc
```diff
@@ -1477,6 +1477,20 @@ void setupLuaConfig(bool client)
       g_pools.setState(localPools);
     });
 
+  g_lua.writeFunction("setPoolBackupPolicy", [](string backupPool, string pool) {
```
ahupowerdns: I think this Lua function is actually 'addPoolBackupPool'? and not 'set'? It also does not add a policy but a pool I think?

pavel-odintsov: Done! Renamed :)

pavel-odintsov:
I think https://github.com/PowerDNS/pdns/pull/7140 implements same approach but much better way.

PR #7138, opened 2018-11-02
dnsdist/rec: Drop remaining capabilities after startup

rgacogne:
### Short description

Drop any remaining capabilities after startup, for example if we have been started as root without `--uid` or `--gid` (please don't do that) or as an unprivileged user with ambient capabilities like `CAP_NET_BIND_SERVICE`.
This does change the existing behavior for setups not using `--uid` or `--gid`, because `dnsdist` used to keep `CAP_NET_BIND_SERVICE` around in that case, and won't after this PR is merged. The only case I can think of right now where it might matter is someone willing to start the webserver on a privileged port after startup, using the console.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code


rgacogne:
Added the same mechanism for the recursor.

pieterlexis:
> This does change the existing behavior for setups not using --uid or --gid, because dnsdist used to keep CAP_NET_BIND_SERVICE around in that case, and won't after this PR is merged.

This should be documented somewhere indeed.

Another thing is that this capability should be added to the spec files, dockerfiles and debian/rules etc. so it is actually used in our packages (I can PR that if you want)

rgacogne:
> Another thing is that this capability should be added to the spec files, dockerfiles and debian/rules etc. so it is actually used in our packages (I can PR that if you want)

Do you know how dnsdist is currently started in our packages? With `--uid dnsdist` ? Do we create the `dnsdist` user using sysusers where available?

pieterlexis:
> Do you know how dnsdist is currently started in our packages? With --uid dnsdist ? Do we create the dnsdist user using sysusers where available?

1. yes ([debian](https://github.com/PowerDNS/pdns/blob/c2ca8dc0dbf73969759a5812deff0ddb4990d363/builder-support/debian/dnsdist/debian-stretch/rules#L57), [centos](https://github.com/PowerDNS/pdns/blob/c2ca8dc0dbf73969759a5812deff0ddb4990d363/builder-support/specs/dnsdist.spec#L125-L126)
2. no sysusers atm, they are created in the postinstall

> Same question for the rec :) I can check if you don't know, of course :)

Similar answers as above :)

pdns/recursordist/Makefile.am
```diff
@@ -369,6 +371,9 @@ rec_control_SOURCES = \
 	rec_control.cc \
 	unix_utility.cc
 
+rec_control_LDADD = \
```
pieterlexis: I don't see why we'd need libcap support in `rec_control` (apart from "this is `#include`d because reasons")

rgacogne: Because it links `misc.o` in, which now refers to symbols from `libcap`.

rgacogne: Fixed!

PR #7148, opened 2018-11-06
sdig: handle non-IN class records better

Habbie:
### Short description
This avoids garbage output for non-IN class records in responses.

### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code


pdns/sdig.cc
```diff
@@ -27,6 +27,19 @@ void usage() {
   cerr<<"Syntax: sdig IP-ADDRESS PORT QUESTION QUESTION-TYPE [dnssec] [recurse] [showflags] [hidesoadetails] [hidettl] [tcp] [ednssubnet SUBNET/MASK] [xpf XPFDATA]"<<endl;
 }
 
+const string nameForClass(uint16_t qclass, uint16_t qtype)
```
pieterlexis: perhaps we need a `QClass::toString()` so we don't need this function

Habbie: Yes, makes sense, and then we can also fix this in the other tools.

PR #7152, opened 2018-11-06
PowerDNS Recursor 4.0.9

aerique:


rgacogne:
LGTM failure is because it's trying to analyse code for the auth in the rec-4.0.x branch, where the auth is not in great shape.

PR #7168, opened 2018-11-08
dnsdist: Fix warnings, mostly unused parameters, reported by -Wextra

rgacogne:
### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


rgacogne:
Note that since `std::thread::thread()` accepts any Callable, we should just change the return type of our thread functions to `void` instead of `void *`.

PR #7178, opened 2018-11-09
rec: Skip NS for the exact zone in CNAME answers

rgacogne:
### Short description

Redo of https://github.com/PowerDNS/pdns/pull/6979.

These NS can't be authoritative since we have a CNAME answer for which only the record describing that alias is necessarily authoritative.
But if we allow the current auth, which might be serving the child zone, to raise the TTL of non-authoritative NS in the cache, they might be able to keep a "ghost" zone alive forever, even after the
delegation is gone from the parent.
So let's just do nothing with them, we can fetch them directly ifwe need them.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/pdns/recursordist/test-syncres_cc.cc
+++ b/pdns/recursordist/test-syncres_cc.cc
@@ -10119,6 +10119,77 @@ BOOST_AUTO_TEST_CASE(test_getDSRecords_multialgo_two_highest) {
   }
 }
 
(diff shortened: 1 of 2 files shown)


rgacogne:
> I don't see any obvious issues, but we should test this against some real domains as well.

I just tested this against the top 10k domains and saw no significant difference from the same test against master so IMHO this can be merged.

PR #7181, opened 2018-11-12
dnsdist: fix wrong client mode when running check-config

chbruyand:
### Short description
`setupLua()` is getting called as if it where acting in client mode (`--client`) when asked to check config. This PR fixes that and marks client mode as `false`.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


zeha:
Will this not conflict with a running dnsdist?

chbruyand:
Indeed! This requires more work actually, see https://github.com/PowerDNS/pdns/issues/7185

PR #7182, opened 2018-11-12
Add metrics to ixfrdist

pieterlexis:
### Short description
This PR adds prometheus statistics to ixfrdist. It also refactors the `WebServer` class somewhat to make it more reusable (like for ixfrdist :smile:).

I've opened this PR early for comments on the code and to get some more ideas for statistics.

ToDo:

- [x] Add tests for the statistics

Future work:

- Make the webserver PW/API-key runtime-configable
- Add an API to ixfrdist

CC: @atosatto

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/ixfrdist-stats.hh
+++ b/pdns/ixfrdist-stats.hh
@@ -25,6 +25,7 @@
 #include <string>
 
 #include "dnsname.hh"
+#include "pdnsexception.hh"
 
 class ixfrdistStats {
   public:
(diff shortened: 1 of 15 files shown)


pdns/ixfrdist-web.hh
```diff
@@ -0,0 +1,37 @@
...
+#include "iputils.hh"
+
+class IXFRDistWebServer
+{
+  public:
+    explicit IXFRDistWebServer(const ComboAddress &listenAddress, const NetmaskGroup &acl);
+    void go();
+
+  private:
+    WebServer *d_ws;
```
rgacogne: unique_ptr?

pdns/ixfrdist-web.cc
```diff
@@ -0,0 +1,50 @@
...
+#include "ixfrdist-web.hh"
+#include <thread>
+#include "threadname.hh"
+#include "iputils.hh"
+#include "ixfrdist-stats.hh"
+
+string doGetStats();
+
+IXFRDistWebServer::IXFRDistWebServer(const ComboAddress &listenAddress, const NetmaskGroup &acl) {
+  // TODO wrap in smart pointer
```
rgacogne: Yes, please :) :+1:

pdns/ixfrdist-stats.hh
```diff
@@ -0,0 +1,88 @@
...
+
+        std::atomic<uint64_t> numAXFRFailures;
+        std::atomic<uint64_t> numIXFRFailures;
+    };
+    class programStats {
+      public:
+        time_t startTime;
+    };
+
+    std::map<DNSName, perDomainStat> domainStats;
```
rgacogne: It looks to me that with the current implementation, you can have several threads inserting into this map, because the way `setSOASerial()` and the `increment*()` methods access it, they would create a new entry if it doesn't already exist.
Now, it seems that this currently can't happen because `registerDomain()` will have been called for every zone first, creating the entry in the map. But at the very least we should document that, and perhaps alter the `setSOASerial()` and `increment*()` methods so that they can't insert into the table by mistake.

pdns/ixfrdist.cc
```diff
@@ -1046,6 +1093,26 @@ static bool parseAndCheckConfig(const string& configpath, YAML::Node& config) {
...
+      retval = false;
+    }
+  }
+
+  if (config["webserver-acl"]) {
+    try {
+      config["webserver-acl"].as<vector<Netmask>>();
+    }
+    catch (const runtime_error &e) {
+      g_log<<Logger::Error<<"Unable to read 'webserver-address' value: "<<e.what()<<endl;
```
rgacogne: s/webserver-address/webserver-acl/

pieterlexis:
Added tests and fixed review items.

pdns/ixfrdist-stats.hh
```diff
@@ -0,0 +1,102 @@
...
+    ixfrdistStats() {
+      progStats.startTime = time(nullptr);
+    }
+
+    std::string getStats();
+
+    void setSOASerial(const DNSName& d, const uint32_t serial) {
+      wasRegistered(d);
+      domainStats[d].currentSOA = serial;
+      domainStats[d].haveZone = true;
```
rgacogne: It's a bit sad to do the lookup in the map twice every time you update the statistics, how about having `wasRegistered()` return the iterator?

pieterlexis: fixed!

Issue #5366, opened 2017-06-02
Compiler warning in sillyrecords.cc


pieterlexis:
- Program: Authoritative, Recursor,
 - Issue type: Bug report

### Short description

GCC 7.1 (`gcc (GCC) 7.1.1 20170516`) issues the following warning when compiling the recursor and the authoritative server:

```
sillyrecords.cc: In member function ‘virtual std::__cxx11::string LOCRecordContent::getZoneRepresentation(bool) const’:
sillyrecords.cc:296:8: warning: ‘%.2f’ directive output may be truncated writing between 4 and 313 bytes into a region of size between 0 and 37 [-Wformat-truncation=]
 string LOCRecordContent::getZoneRepresentation(bool noDot) const
        ^~~~~~~~~~~~~~~~
In file included from /usr/include/stdio.h:939:0,
                 from /usr/include/c++/7.1.1/cstdio:42,
                 from /usr/include/c++/7.1.1/ext/string_conversions.h:43,
                 from /usr/include/c++/7.1.1/bits/basic_string.h:6159,
                 from /usr/include/c++/7.1.1/string:52,
                 from utility.hh:45,
                 from sillyrecords.cc:4:
/usr/include/bits/stdio2.h:65:44: note: ‘__builtin___snprintf_chk’ output between 48 and 1938 bytes into a destination of size 79
        __bos (__s), __fmt, __va_arg_pack ());
                                            ^
```

### Environment
 - Operating system: Arch Linux
 - Software version: master
 - Software source: git repo

### Steps to reproduce
1. `./bootstrap && ./configure --without-modules --without-dynmodules && make -C pdns sillyrecords.o`

### Expected behaviour
No warning 😄 

### Actual behaviour
The above warning is shown

Habbie:
While poking at this I noticed we 32-bit truncate SIZE-encoded things. This helps:
```diff
diff --git a/pdns/sillyrecords.cc b/pdns/sillyrecords.cc
index 289cdb18c..970b2aa6f 100644
--- a/pdns/sillyrecords.cc
+++ b/pdns/sillyrecords.cc
@@ -16,7 +16,7 @@ static unsigned int poweroften[10] = {1, 10, 100, 1000, 10000, 100000,
 /* converts ascii size/precision X * 10**Y(cm) to 0xXY. moves pointer.*/
 static uint8_t precsize_aton(const char **strptr)
 {
-  unsigned int mval = 0, cmval = 0;
[...]

PR #7211, opened 2018-11-19
Check values before formatting a LOC record into a string

rgacogne:
### Short description

This makes the warning from #5366 go away (GCC 8.2.1). I'm not sure silently "fixing" the values is the right way to go, perhaps we might want to raise an exception instead. 

### Checklist

I have:
- [x] compiled this code
- [ ] tested this code
- [ ] documented the code


pieterlexis:
This is conflicted due to 0e12635d6975731441ae27e5b98f070ecfb0cf8a, which also fixes this issue.

PR #7217, opened 2018-11-21
auth: geoip, properly delete libGeoIP return values

### LINKED ISSUE #7184: There is a memory leak at the powerdns geoip backend
- Program: Authoritative
- Version: 4.1.3
- Issue type: Bug Report

### Description
I'm using powerdns authoritative 4.1.3 geoip backend. I configured a number of domains, the most of which had over 1000 records. Geoip database has approximately 30,000 + subnet records.

When I use it, I find that the server's memory has been growing, and the reset has not decreased. I had to restart the service once a day.
image

I also found that the memory growth rate is proportional to the number of accesses, I suspect that the geoip back end caches requests and responses, but does not free memory properly. Here is the average number of requests on both servers.
image

What information do I need to provide to fix this problem?

chbruyand:
### Short description
This PR wraps libGeoIP return values in `unique_ptr` to make sure they are properly deleted.

This should fix #7184 

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/modules/geoipbackend/geoipinterface-dat.cc
+++ b/modules/geoipbackend/geoipinterface-dat.cc
@@ -26,11 +26,30 @@
 #include "geoipinterface.hh"
 #ifdef HAVE_GEOIP
 #include "GeoIPCity.h"
+#include "GeoIP.h"
 
 struct geoip_deleter {
   void operator()(GeoIP* ptr) {
-    if (ptr) GeoIP_delete(ptr);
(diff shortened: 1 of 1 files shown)


rgacogne:
Perhaps @cmouse will want to take a look before we consider merging this?

cmouse:
LGTM too.

PR #7219, opened 2018-11-24
auth: geoip - Fix static lookup when using weighted records on multiple record types

### LINKED ISSUE #7051: GeoIP sometimes does not answer on when using weighted records and different records types
- Program: Authoritative
 - Issue type: Bug report

### Short description

Hello, when using the GeoIP backend with weighted records and different types (A and AAAA) the server respond nothing half of the time (the sum of the weight of the A records is the same as the AAAA records)

### Environment

 - Operating system: CenOS 7.4.1708
 - Software version: 4.1.4

### Steps to reproduce

Configuration:

```
setgid=pdns
setuid=pdns
launch=geoip,bind

bind-config=/etc/pdns/bind.conf
do-ipv6-additional-processing=no
edns-subnet-processing=yes
geoip-database-cache=memory
geoip-database-files=/usr/share/GeoIP/GeoIPCity.dat
geoip-zones-file=/etc/pdns/geoip.yaml
local-port=53
```

geoip.yaml

```---
domains:
- domain: example.com
  ttl: 300
  records:
    example.com:
    - a:
        content: 192.168.0.1
        weight: 50
    - a:
        content: 192.168.0.2
        weight: 50
    - aaaa:
        content: 2042::1
        weight: 50
    - aaaa:
        content: 2042::2
        weight: 50
```

```
# dig AAAA example.com @localhost  +short
# dig AAAA example.com @localhost  +short
# dig AAAA example.com @localhost  +short
2042::2
# dig AAAA example.com @localhost  +short
2042::2
# dig AAAA example.com @localhost  +short
2042::2
# dig AAAA example.com @localhost  +short
2042::2
# dig AAAA example.com @localhost  +short
2042::1
# dig AAAA example.com @localhost  +short
2042::2
# dig AAAA example.com @localhost  +short
2042::1
# dig AAAA example.com @localhost  +short
# dig AAAA example.com @localhost  +short
2042::1
# dig AAAA example.com @localhost  +short
# dig AAAA example.com @localhost  +short
# dig AAAA example.com @localhost  +short
# dig AAAA example.com @localhost  +short
# dig AAAA example.com @localhost  +short
# dig AAAA example.com @localhost  +short
2042::2
```
### Expected behaviour

I expect to get an answer each time, 2042::1 or 2042::2.

### Other information

Reading the geoip backend here https://github.com/PowerDNS/pdns/blob/auth-4.1.4/modules/geoipbackend/geoipbackend.cc#L327, the weight is computed for each records without distinction of record types.
Perhaps it could be relevant to compute weight for each type.

Annih:
### Short description
Computes the record weight/probability per QType and adapt the code to avoid empty answer in a mixed-weighted-Qtype situation.

Closes #7051.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/modules/geoipbackend/geoipbackend.cc
+++ b/modules/geoipbackend/geoipbackend.cc
@@ -318,6 +318,8 @@ bool GeoIPBackend::lookup_static(const GeoIPDomain &dom, const DNSName &search, 
 
   if (i != dom.records.end()) { // return static value
     for(const auto& rr : i->second) {
+      if (qtype != QType::ANY && rr.qtype != qtype) continue;
+
       if (rr.has_weight) {
         gl.netmask = (v6?128:32);
         int comp = cumul_probabilities[rr.qtype.getCode()];
@@ -325,13 +327,11 @@ bool GeoIPBackend::lookup_static(const GeoIPDomain &dom, const DNSName &search, 
         if (rr.weight == 0 || probability_rnd < comp || probability_rnd > (comp + rr.weight))
           continue;
       }
-      if (qtype == QType::ANY || rr.qtype == qtype) {
-        const string& content = format2str(rr.content, ip, v6, gl);
-        if (rr.qtype != QType::ENT && rr.qtype != QType::TXT && content.empty()) continue;
-        d_result.push_back(rr);
-        d_result.back().content = content;
-        d_result.back().qname = qdomain;
-      }
+      const string& content = format2str(rr.content, ip, v6, gl);
+      if (rr.qtype != QType::ENT && rr.qtype != QType::TXT && content.empty()) continue;
+      d_result.push_back(rr);
+      d_result.back().content = content;
+      d_result.back().qname = qdomain;
     }
     // ensure we get most strict netmask
     for(DNSResourceRecord& rr: d_result) {
--- a/modules/geoipbackend/geoipbackend.cc
+++ b/modules/geoipbackend/geoipbackend.cc
@@ -261,24 +261,30 @@ void GeoIPBackend::initialize() {
 
     // finally fix weights
     for(auto &item: dom.records) {
-      float weight=0;
-      float sum=0;
+      map<uint16_t, float> weights;
+      map<uint16_t, float> sums;
+      map<uint16_t, GeoIPDNSResourceRecord> lasts;
       bool has_weight=false;
       // first we look for used weight
       for(const auto &rr: item.second) {
-        weight+=rr.weight;
+        weights[rr.qtype.getCode()] += rr.weight;
         if (rr.has_weight) has_weight = true;
       }
       if (has_weight) {
         // put them back as probabilities and values..
         for(auto &rr: item.second) {
-          rr.weight=static_cast<int>((static_cast<float>(rr.weight) / weight)*1000.0);
-          sum += rr.weight;
+          uint16_t rr_type = rr.qtype.getCode();
+          rr.weight=static_cast<int>((static_cast<float>(rr.weight) / weights[rr_type])*1000.0);
+          sums[rr_type] += rr.weight;
           rr.has_weight = has_weight;
+          lasts[rr_type] = rr;
         }
         // remove rounding gap
-        if (sum < 1000)
-          item.second.back().weight += (1000-sum);
+        for(auto &x: lasts) {
+          float sum = sums[x.first];
+          if (sum < 1000)
+            x.second.weight += (1000-sum);
+        }
       }
     }
 
@@ -307,15 +313,15 @@ GeoIPBackend::~GeoIPBackend() {
 
 bool GeoIPBackend::lookup_static(const GeoIPDomain &dom, const DNSName &search, const QType &qtype, const DNSName& qdomain, const std::string &ip, GeoIPNetmask &gl, bool v6) {
   const auto& i = dom.records.find(search);
-  int cumul_probability = 0;
+  map<uint16_t,int> cumul_probabilities;
   int probability_rnd = 1+(dns_random(1000)); // setting probability=0 means it never is used
 
   if (i != dom.records.end()) { // return static value
     for(const auto& rr : i->second) {
       if (rr.has_weight) {
         gl.netmask = (v6?128:32);
-        int comp = cumul_probability;
-        cumul_probability += rr.weight;
+        int comp = cumul_probabilities[rr.qtype.getCode()];
+        cumul_probabilities[rr.qtype.getCode()] += rr.weight;
         if (rr.weight == 0 || probability_rnd < comp || probability_rnd > (comp + rr.weight))
           continue;
       }
--- a/regression-tests/backends/geoip-master
+++ b/regression-tests/backends/geoip-master
@@ -26,6 +26,13 @@ domains:
       - ns: ns1.example.com
       - ns: ns2.example.com
       - mx: 10 mx.example.com
+    mixed_weight.geo.example.com:
+      - a:
+          content: "%ip4"
+          weight: 10
+      - txt:
+          content: text
+          weight: 10
     ip.geo.example.com:
       - a: "%ip4"
       - aaaa: "%ip6"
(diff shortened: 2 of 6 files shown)


modules/geoipbackend/geoipbackend.cc
```diff
@@ -261,24 +261,30 @@ void GeoIPBackend::initialize() {
...
+        weights[rr.qtype.getCode()] += rr.weight;
         if (rr.has_weight) has_weight = true;
       }
       if (has_weight) {
         // put them back as probabilities and values..
         for(auto &rr: item.second) {
-          rr.weight=static_cast<int>((static_cast<float>(rr.weight) / weight)*1000.0);
-          sum += rr.weight;
+          uint16_t rr_type = rr.qtype.getCode();
+          rr.weight=static_cast<int>((static_cast<float>(rr.weight) / weights[rr_type])*1000.0);
```
rgacogne: This was already the case with the existing code, but it looks like we would happily divide by `0.0` if the weight of all records for a given type is set to `0`.

Annih: What would be the desired behavior in that case?
Maybe the yaml parser should just ignore records with a weight of 0?


Anyway it seems out of the scope of this review, what do you think @rgacogne?

rgacogne: I think the parser should error out on a weight of `0`.

I agree it's a bit out of the scope so if you don't feel like fixing that in your PR just let me know and I'll either open a new PR or an issue so we don't forget to fix it.

Annih: OK I'm doing it in another review :)

Annih:
@rgacogne I'm new in the pdns opensource world, in order to backport this change to the *4.1.x* branch, what is the process? 

I read I need first to wait that this is one is merged, but I still have some question:
- Should I do the backport or a maintainer will do it?
- Should all the fixes be backported, or only the main fix?

Habbie:
> * Should I do the backport or a maintainer will do it?

If you want it, it's best if you do it, instead of having to wait for us.



> * Should all the fixes be backported, or only the main fix?

We prefer minimal backports whenever possible.

rgacogne:
Perhaps @cmouse might want to review this before we merge it?

modules/geoipbackend/geoipbackend.cc
```diff
@@ -307,25 +313,25 @@ GeoIPBackend::~GeoIPBackend() {
...
 
   if (i != dom.records.end()) { // return static value
     for(const auto& rr : i->second) {
+      if (qtype != QType::ANY && rr.qtype != qtype) continue;
+
       if (rr.has_weight) {
         gl.netmask = (v6?128:32);
-        int comp = cumul_probability;
-        cumul_probability += rr.weight;
+        int comp = cumul_probabilities[rr.qtype.getCode()];
```
cmouse: I am bit worried that QType::ANY will no longer work here.

regression-tests/backends/geoip-master
```diff
@@ -26,6 +26,13 @@ domains:
       - ns: ns1.example.com
       - ns: ns2.example.com
       - mx: 10 mx.example.com
+    mixed_weight.geo.example.com:
+      - a:
+          content: "%ip4"
+          weight: 10
+      - txt:
+          content: text
+          weight: 10
```
cmouse: this test does not really test the functionality

Annih: Could you be more specific on the "functionality" which should be tested and which is not?

Is there something unclear in the commit message where I introduce the test? If yes, how should I reword it to make it clearer?

My goal is to demonstrate we do not regress/retrigger #7051, and ensure that the ANY question answer both records, the A question return the A record and the the TXT return the TXT.
Do you think my test is covering this goal?
Do you think my test is enough?

modules/geoipbackend/geoipbackend.cc
Annih: Do you mean in the case where someone put a ANY record in the yaml? (I would say this is not valid)

Otherwise I would actually say that this patch fixes the behaviour with the QType::ANY. Before this patch a ANY request with mixed QType records would only return 1 record (or zero due to the bug in #7051)
i.e. as I tried to demonstrate with the new regression test, if you specify *A* and *TXT* values in a record, I would expect the ANY question to return 1 record for the *A* type and another for the *TXT* record.

regression-tests/backends/geoip-master
cmouse: although not sure if it's even possible to test it well.

modules/geoipbackend/geoipbackend.cc
cmouse: I am wondering if this still makes sense, because now it can return "different" TXT-A pair if there are more than one specified.

Consider following example:
 - A 1.2.3.4, weight = 10
 - A 4.3.2.1, weight = 10
 - TXT "1-2-3-4", weight = 10
 - TXT "4-3-2-1", weight = 10

If I am reading this code correctly, it can end up returning four different kinds of results.

So, is the requirement to have corresponding records returned, or is it ok for them to behave independently of each other.

Seems that ANY query is handled correctly after all, I read it wrong the first round. But only question now remains is that is it OK that records are chosen randomly by type? If it's OK, then this works correctly.

Annih: @Habbie @rgacogne or @cmouse who should take this decision?

Could we move forward? I'll send the backport review once this is merged.

Annih:
@rgacogne & @cmouse any thought about this?
I know you are working hard on 4.2.0, it would be great if this change could be included in the release.

modules/geoipbackend/geoipbackend.cc
cmouse: I think the result-set needs to contain one of TXT and one of A in this configuration to be correct.

Annih: That's what the new implementation produces; a single random record per QTYPE.

cmouse: so can you paste what `dig any name.geo.example.com` gives?

Annih: Yes certainly I can :), this is exactly the behavior I tried to demonstrate in the added regression test with the ANY query, but due to the randomness of the result I had to simplify the data.

So with the a setup similar to what you described:
```
    name.geo.example.com:
      - a:
          content: 1.2.3.4
          weight: 10
      - a:
          content: 4.2.3.1
          weight: 10
      - txt:
          content: 1-2-3-4
          weight: 10
      - txt:
          content: 4-2-3-1
          weight: 10
```

Here are the results:
```
#1:
name.geo.example.com.   30      IN      TXT     "4-2-3-1"
name.geo.example.com.   30      IN      A       4.2.3.1
#2:
name.geo.example.com.   30      IN      TXT     "4-2-3-1"
name.geo.example.com.   30      IN      A       4.2.3.1
#3:
name.geo.example.com.   30      IN      TXT     "1-2-3-4"
name.geo.example.com.   30      IN      A       1.2.3.4
#4:
name.geo.example.com.   30      IN      TXT     "4-2-3-1"
name.geo.example.com.   30      IN      A       4.2.3.1
#5:
name.geo.example.com.   30      IN      TXT     "1-2-3-4"
name.geo.example.com.   30      IN      A       1.2.3.4
```

Interestingly the pairs are preserved, but I'm not sure this is a feature that should be advertised ^^

aerique:
@rgacogne do you still approve and @cmouse do you approve?

Annih:
Thanks @aerique for the reminder, it has been a while that this PR didn't move :)
@cmouse could you please tell us if you approve or not? If you don't, could you drive me on the right path to fix this patch and move forward?

cmouse:
i already wrote in earlier code that i approve, but lets approve here too

Annih:
Thanks @cmouse :)
So what is required to merge this, who is allowed to do it?

aerique:
> So what is required to merge this, who is allowed to do it?

I or one of the other PowerDNS people will merge it in the coming days.

PR #7233, opened 2018-11-28
Make api changes do a rectify by default, add an option to disable

### LINKED ISSUE #7231: api_rectify as default
- Program: Authoritative
 - Issue type: Feature request

### Short description
We would like the possibility to have api_rectify set as default.

### Usecase
After a modification of any record where the domain has DNSSec, we would like to have a still working DNSSec domain.

### Description
We use the PowerDNS API to modify our domains. We also use DNSSec on all our domains. After modifications we always want to run rectify, if needed. Currently it is possible to modify the domain object (or set on creation) API-RECTIFY to 1. Since we would need to set it always, I hope it is possible to either switch api_rectify to be on by default (since I figure most of the time this is what you want) and/or to provide an option to set the default for this.

tjikkun:
### Short description
Previously, you had to set API-RECTIFY metadata on every zone you wanted to
have automatically rectified on changes through the API. With this change, the
default behaviour is to do a rectify, but this can globally be cahged with the
setting default-api-rectify, and overriden per zone.
It fixes #7231

### Checklist
I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/docs/domainmetadata.rst
+++ b/docs/domainmetadata.rst
@@ -56,7 +56,9 @@ This metadata item controls whether or not a zone is fully rectified on changes
 to the contents of a zone made through the :doc:`API <http-api/index>`.
 
 When the ``API-RECTIFY`` value is "1", the zone will be rectified on changes.
-Any other other value means that it will not be rectified.
+Any other other value means that it will not be rectified. If this is not set
+at all, rectifying of the zone depends on the config variable
+:ref:`setting-default-api-rectify`.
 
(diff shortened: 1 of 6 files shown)


chbruyand:
I'm not sure we want this to be the default. Changes through the API should not have to be rectified after (except implicitly for some changes like securing a zone). Maybe this new setting could be set to "no" by default and enabled by admins that _want_ that ?

Issue #7234, opened 2018-11-28
No DNSSEC validation in case of "--forward-zones-recurse"


paddg:
- Program: Recursor
 - Issue type: Bug report

### Short description
If I forward queries to a full DNSSEC validating Resolver, answers are always NOERROR even if I ask for dnssec-failed.org.

### Environment
 - Operating system: SUSE Linux Enterprise 12 SP3
 - Software version: PowerDNS Recursor 4.1.8
 - Software source: self compiled

### Steps to reproduce
Start the Recursor:
`pdns_recursor --config-dir=. --socket-dir=. --daemon=no --forward-zones-recurse=.=8.8.8.8 --dnssec=validate`

Send a query against it:
```
dig dnssec-failed.org @127.0.0.1

; <<>> DiG 9.10.4-P5 <<>> dnssec-failed.org @127.0.0.1
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 17625
;; flags: qr rd ra ad; QUERY: 1, ANSWER: 1, AUTHORITY: 0, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 4096
;; QUESTION SECTION:
;dnssec-failed.org.             IN      A

;; ANSWER SECTION:
dnssec-failed.org.      7199    IN      A       69.252.80.75

;; Query time: 133 msec
;; SERVER: 127.0.0.1#53(127.0.0.1)
;; WHEN: Wed Nov 28 10:27:47 CET 2018
;; MSG SIZE  rcvd: 62
```
### Expected behaviour
If I run a Recursor with `--dnssec=validate` and `--forward-zones-recurse=.=` to a full validating Resolver I'd expect a SERVFAIL in case of a bogus DNSSEC domain.

### Actual behaviour
The answer status is NOERROR in each case. Even the AD bit is set in the answer if I ask for `dnssec-failed.org`

M4t7e:
I have seen the same behavior when forwarding some zones to an OpenNIC DNS resolver. Thanks for reporting @paddg!

rgacogne:
We did a lot of cleanup in that area, would you happen to know if this issue still remains in 4.2.0?

paddg:
Yes I can confirm, it is fixed. Tested wit 4.2.0 and 4.1.12. Thanks @rgacogne!

PR #4377, opened 2016-08-29
Rec: make the record cache forward-aware

pieterlexis:
This fixes #4369.

We now consistently SERVFAIL on undelegated, forwarded zones inside secure zone (because validation rightfully fails)

For undelegated, forwarded zones inside insecure zones, we answer the query (as validation notices the parent zone is already insecure).

(unrelated) This PR also moves the DNSSEC regression tests to the local loopback address range.

Issue #4369, opened 2016-08-26
Recursor: with dnssec inconsistently answers with NXDOMAIN sometimes (instead of SERVFAIL or answer)


gertvdijk:
Recursor version 4.0.2.
Relevant entries in my `recursor.conf`:

```
local-address=0.0.0.0
forward-zones=test.gert.lan=192.168.123.45:5300
dnssec=validate
```

Documentation says:

> `validate`
>
> Full blown DNSSEC validation. Send SERVFAIL to clients on bogus responses.

Here's what I see on a resolver after a recursor restart _sometimes_:

```
$ host host1.test.gert.lan; sleep 1; host host2.test.gert.lan
Host host1.test.gert.lan not found: 2(SERVFAIL)
Host host2.test.gert.lan not found: 3(NXDOMAIN)
```

PCAP analysis shows this:

```
1  workstation    -> recursor402    DNS 78 Standard query 0x0192 A host1.test.gert.lan
2  recursor402    -> 192.168.123.45 DNS 89 Standard query 0xc832 A host1.test.gert.lan OPT
3  192.168.123.45 -> recursor402    DNS 105 Standard query response 0xc832 A host1.test.gert.lan A 192.168.45.67 OPT
4  recursor402    -> 192.228.79.201 DNS 70 Standard query 0x5238 DNSKEY <Root> OPT
5  192.228.79.201 -> recursor402    DNS 778 Standard query response 0x5238 DNSKEY <Root> DNSKEY DNSKEY RRSIG OPT
6  recursor402    -> 199.7.83.42    DNS 74 Standard query 0x6f5c DS lan OPT
7  199.7.83.42    -> recursor402    DNS 685 Standard query response 0x6f5c No such name DS lan NSEC lancaster RRSIG NSEC aaa RRSIG SOA a.root-servers.net RRSIG OPT
8  recursor402    -> workstation    DNS 78 Standard query response 0x0192 Server failure A host1.test.gert.lan
9  workstation    -> recursor402    DNS 80 Standard query 0xc27f A host2.test.gert.lan
10 recursor402    -> workstation    DNS 155 Standard query response 0xc27f No such name A host2.test.gert.lan SOA a.root-servers.net
```

Corresponding Recursor logs with `quiet=no` and `loglevel=999`:

```
2 [4/1] question for 'host1.test.gert.lan|A' from <ip-workstation>
[...]

PR #7238, opened 2018-11-28
rec: Fix DNSSEC validation with forward-zone-recurse

rgacogne:
### Short description

Related to #7234.

Only did a few quick tests, needs a lot more testing and regression tests.
We also need to investigate the:

```
!(wasForwardedOrAuthZone && ne->d_auth != authname)) { // Only the authname nameserver can neg cache entries
```

removal because as of now this code makes no sense to me.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/pdns/syncres.cc
+++ b/pdns/syncres.cc
@@ -1137,9 +1137,7 @@ bool SyncRes::doCacheCheck(const DNSName &qname, const DNSName& authname, bool w
     giveNegative = true;
     cachedState = ne->d_validationState;
   }
-  else if (t_sstorage.negcache.get(qname, qtype, d_now, &ne) &&
-           !(wasForwardedOrAuthZone && ne->d_auth != authname)) { // Only the authname nameserver can neg cache entries
-
+  else if (t_sstorage.negcache.get(qname, qtype, d_now, &ne)) {
     /* If we are looking for a DS, discard NXD if auth == qname
        and ask for a specific denial instead */
     if (qtype != QType::DS || ne->d_qtype.getCode() || ne->d_auth != qname ||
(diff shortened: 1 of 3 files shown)


pdns/syncres.cc
```diff
@@ -2143,7 +2142,7 @@ RCode::rcodes_ SyncRes::updateCacheFromRecords(unsigned int depth, LWResult& lwr
     if (shouldValidate() && recordState == Secure) {
       vState initialState = recordState;
 
-      if (isAA) {
+      if (isAA || wasForwardRecurse) {
```
rgacogne: This most likely breaks for `CNAME` answers containing more than the exact name sought.

rgacogne:
Rebased on master because the DNSPython / DNSCrypt issue has been fixed there.

pieterlexis:
>  We also need to investigate the:
> `!(wasForwardedOrAuthZone && ne->d_auth != authname)) { // Only the authname nameserver can neg cache entries`
> removal because as of now this code makes no sense to me.

Commit 64a8b6a1c1e1f165c8eb716f399e4da6d71156e7 has the reason:

> Make the negcache forwarded zones aware
>     
> Because of DNSSEC (without NTAs configured), we 'leak' the forwarded
> name to the cache when looking for the DNSKEY. This resulted in NXDOMAIN
> answers for actual NXDOMAINs, but with the root SOA record. Leading to
> inconsistent answers to the client.

We'll need to check what happens with this commit (it looks like no regression tests exist)

pieterlexis:
I lied, #4377 adds tests for this issue (#4369)! So if travis says "OK", it hopefully is. Will do the code review now

rgacogne:
> Commit 64a8b6a has the reason:

Yes, but that only makes sense to me for the first part of the commit, along `// when forwarding, the root may only neg-cache if it was forwarded to`. I don't get the second part, the one I removed in this PR.

pieterlexis:
The second part existed for non-delegated subdomains not from the root afaik (but this was some time ago though, so I don't remember exactly)

PR #7258, opened 2018-12-04
rec: Skip NS records in authority/additional sections of NXD / NODATA

rgacogne:
### Short description

These records are quite susceptible to fragmentation poisoning.
Based on a suggestion by Tsunehiko Suzuki (thanks!).

Note that Unbound introduced the same change in 1.8.2:
- NLnetLabs/unbound@f7e9913
- NLnetLabs/unbound@7458729

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/recursordist/test-syncres_cc.cc
+++ b/pdns/recursordist/test-syncres_cc.cc
@@ -10146,6 +10146,116 @@ BOOST_AUTO_TEST_CASE(test_getDSRecords_multialgo_two_highest) {
   }
 }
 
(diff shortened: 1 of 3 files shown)


rgacogne:
Note that unbound doesn't accept additional `A` and `AAAA` records that are unrelated to an accepted record in the previous sections, so they would for example drop an `AAAA` for the skipped `NS`, which we could currently insert into our cache.
So that PR probably needs more work.

rgacogne:
I'm working on a stricter version of this PR, don't merge yet!

PR #7267, opened 2018-12-06
dnsdist: Check that SO_ATTACH_BPF is defined before enabling eBPF

rgacogne:
### Short description

It turns out that RH decided to backport only the tracing subsystem part of eBPF as a "Technology Preview" in the RHEL / Centos 7.6 kernel, so most of the eBPF stuff is present but not what is needed
to use the networking parts.

With this change, eBPF support is disabled when `SO_ATTACH_BPF` is missing since we can't attach our filter to a socket without that anyway. Tested on CentOS 7.6 and Arch.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


Habbie:
This builds on el7, and `dnsdist --version` is unchanged from before (no ebpf on either side).

Habbie:
Debian stretch: no `--version` changes with/without this PR.

PR #7284, opened 2018-12-13
auth: Fallback to sha1 for the signatures cache if md5 is not available

rgacogne:
### Short description

This requires switching from the deprecated `MD5()` API to the `EVP` one, otherwise OpenSSL aborts before we can check whether MD5 is available. It makes computing a hash a bit slower (~10% on my desktop) because we need to allocate the context, which is opaque in 1.1.1 but now might be a good time to bit the bullet.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


pdns/digests.hh
```diff
@@ -0,0 +1,81 @@
...
+  if (written != result.size()) {
+    throw std::runtime_error(std::string(EVP_MD_name(md)) + " EVP final wrote " + std::to_string(written) + ", expected " + std::to_string(result.size()));
+  }
+
+  return result;
+}
+
+inline std::string pdns_md5sum(const std::string& input)
+{
+  const auto& md = EVP_md5();
```
chbruyand: Maybe remove the reference (here and on line 75) as it is a pointer ?

pdns/digests.hh
```diff
@@ -0,0 +1,81 @@
...
+#pragma once
+
+#include <stdexcept>
+#include <string>
+
+#include <openssl/evp.h>
+
+inline std::string pdns_hash(const EVP_MD * md, const std::string& input)
+{
+  std::string result;
```
chbruyand: Maybe it could be declared and resized just before it is used ?

rgacogne:
Fixed the two comments from the review and squashed the commits into one.

PR #7293, opened 2018-12-18
pdns_control notify * also notify slaves zones

alebeta90:
### Short description

pdns_control notify * also notify slaves zones when slave-renotify option is enabled

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code


Habbie:
Thanks! Want to take this opportunity to replace the numbers with the names from here?
`  enum DomainKind : uint8_t { Master, Slave, Native } kind;`

alebeta90:
@Habbie perfect, it is such a good idea :)

omoerbeek:
Please use enum names for di->kind, as @Habbie asked earlier.

If slave-renotify is set and there are slave domains in the db but server is configured as master only this code still will send notifies for the slave zones. This is a bit strange but not necessarily wrong.

Issue #7169, opened 2018-11-08
pdns-recursor dnssec issues with default Debian 9 install


firthmj:
- Program: Recursor
 - Issue type: Bug report

### Short description

On installing pdns-recursor with minimal configuration (only changing "local-address" from the default of localhost only) on Debian 9, many queries appear to fail with SERVFAIL

Adding "dnssec=off" to the configuration file allows the queries to resolve, but seems like a bad thing to have to set.

### Environment

 - Operating system: Debian 9 (stretch) with latest system patches
 - Software version: 4.0.4 (from Debian repositories) or 4.1.6-1pdns (from PowerDNS repository)
 - Software source: Tried both from default Debian repository, and latest version from PowerDNS repository 

### Steps to reproduce

1. apt-get install pdns-recursor
2. edit /etc/powerdns/recursor.conf to change the "local-address" parameter to also include system's Ethernet IP address
3. (for example) dig isc.org soa @hostname (replace hostname with actual hostname)

Same result for local queries sent to @localhost

### Expected behaviour

The "dig" command would return the SOA record for the isc.org domain

### Actual behaviour

The "dig" command just indicates that the server returned "SERVFAIL":

; <<>> DiG 9.10.3-P4-Ubuntu <<>> www.isc.org @XXXXX
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: SERVFAIL, id: 64900
;; flags: qr rd ra; QUERY: 1, ANSWER: 0, AUTHORITY: 0, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 4096
;; QUESTION SECTION:
;www.isc.org.			IN	A

;; Query time: 1784 msec
;; SERVER: 10.69.80.56#53(10.69.80.56)
;; WHEN: Thu Nov 08 17:00:49 GMT 2018
;; MSG SIZE  rcvd: 40

This example was using "dig" on a remote Ubuntu system, which is why the output has "ubuntu" in it - the server being queried was Debian 9.

### Other information

This failure is only for certain domains - at least "isc.org" and "gnu.org" are failing, but others are working (e.g. "bbc.com")

As mentioned above, adding "dnssec=off" to the configuration file makes the queries to the failing domains start working.

I do have a PCAP of the DNS packets to/from the machine running pdns-recursor for a few queries, but want to make sure I'm not doing something stupid first.

[...]

PR #7307, opened 2019-01-02
rec: Lower default EDNS0 buffersizes to 1232 bytes

pieterlexis:
### Short description
This should prevent issues with networks that drop fragments, have
broken PMTU detection, etc.

This commit lowers both the udp-truncation-threshold (stub<->res) and
edns-outgoing-bufsize (rec<->auths).

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/recursordist/docs/settings.rst
+++ b/pdns/recursordist/docs/settings.rst
@@ -419,8 +419,16 @@ found, the recursor fallbacks to sending 127.0.0.1.
 
 ``edns-outgoing-bufsize``
 -------------------------
+.. versionchanged:: 4.2.0
+  Before 4.2.0, the default was 1680
+
 -  Integer
--  Default: 1680
+-  Default: 1232
+
+.. note:: Why 1232?
+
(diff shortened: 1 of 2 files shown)


pdns/recursordist/docs/settings.rst
```diff
@@ -419,8 +419,11 @@ found, the recursor fallbacks to sending 127.0.0.1.
...
 -------------------------
+.. versionchanged:: 4.2.0
+  Before 4.2.0, the default was 1680
+
 -  Integer
--  Default: 1680
+-  Default: 1232
 
 This is the value set for the EDNS0 buffer size in outgoing packets.
 Lower this if you experience timeouts.
```
rgacogne: It would be nice to document where the `1232` value is coming from, so we don't have to guess 10 years from now ;)

aerique:
Maybe also mention #7169 ?

pieterlexis:
> Maybe also mention #7169 ?

That would be something for the blogpost, not the docs

PR #7310, opened 2019-01-03
dnsdist: Add a 'keepStaleData' option to the packet cache

### LINKED ISSUE #7239: dnsdist: stop expunging entries from the cache if all servers are down
- Program: dnsdist
 - Issue type: Feature request

### Short description
It would be nice to be able to prevent expired entries from being expunged from the cache when all the backends this cache is associated to are unavailable.

### Usecase
dnsdist already has the ability to serve stale data from the cache when no backend are available, but it currently keeps cleaning the cache, removing those entries.
This new option would make sure we can continue serving stale data until the backend are available again.

### Description
Add a new optional `keepStaleData` parameter to the `newPacketCache()` command.

rgacogne:
### Short description

This new option prevents expired entries from being expunged from the packet cache associated to a pool if all the backends in this pool are down, so we can keep serving stale data until at least one backend gets available again. 

Closes #7239.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/pdns/dnsdist-lua-bindings.cc
+++ b/pdns/dnsdist-lua-bindings.cc
@@ -220,8 +220,63 @@ void setupLuaBindings(bool client)
 #endif /* HAVE_EBPF */
 
   /* PacketCache */
-  g_lua.writeFunction("newPacketCache", [](size_t maxEntries, boost::optional<uint32_t> maxTTL, boost::optional<uint32_t> minTTL, boost::optional<uint32_t> tempFailTTL, boost::optional<uint32_t> staleTTL, boost::optional<bool> dontAge, boost::optional<size_t> numberOfShards, boost::optional<bool> deferrableInsertLock, boost::optional<uint32_t> maxNegativeTTL, boost::optional<bool> ecsParsing) {
-      return std::make_shared<DNSDistPacketCache>(maxEntries, maxTTL ? *maxTTL : 86400, minTTL ? *minTTL : 0, tempFailTTL ? *tempFailTTL : 60, maxNegativeTTL ? *maxNegativeTTL : 3600, staleTTL ? *staleTTL : 60, dontAge ? *dontAge : false, numberOfShards ? *numberOfShards : 1, deferrableInsertLock ? *deferrableInsertLock : true, ecsParsing ? *ecsParsing : false);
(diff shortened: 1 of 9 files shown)


pdns/dnsdist-lua-bindings.cc
```diff
@@ -220,8 +220,63 @@ void setupLuaBindings(bool client)
...
-      return std::make_shared<DNSDistPacketCache>(maxEntries, maxTTL ? *maxTTL : 86400, minTTL ? *minTTL : 0, tempFailTTL ? *tempFailTTL : 60, maxNegativeTTL ? *maxNegativeTTL : 3600, staleTTL ? *staleTTL : 60, dontAge ? *dontAge : false, numberOfShards ? *numberOfShards : 1, deferrableInsertLock ? *deferrableInsertLock : true, ecsParsing ? *ecsParsing : false);
+  g_lua.writeFunction("newPacketCache", [](size_t maxEntries, boost::optional<uint32_t> maxTTL, boost::optional<uint32_t> minTTL, boost::optional<uint32_t> tempFailTTL, boost::optional<uint32_t> staleTTL, boost::optional<bool> dontAge, boost::optional<size_t> numberOfShards, boost::optional<bool> deferrableInsertLock, boost::optional<uint32_t> maxNegativeTTL, boost::optional<bool> ecsParsing, boost::optional<std::unordered_map<std::string, boost::variant<bool, size_t>>> vars) {
```
pieterlexis: Is there a reason to keep this backwards compatible? I believe 1.4.0 will be the next version so we could just _only_ accept an `options` table.

pdns/dnsdistdist/docs/reference/config.rst
```diff
@@ -583,13 +602,19 @@ See :doc:`../guides/cache` for a how to.
     :param int qtype: The type to expunge
     :param bool suffixMatch: When set to true, remove al entries under ``name``
 
+  .. method:: PacketCache:printStats()
```
pieterlexis: `s/printStats/getStats/`

rgacogne: Fixed, thanks!

pdns/dnsdist-lua-bindings.cc
rgacogne: I'm actually not sure yet, so until a breaking change comes up I'm going to assume we will release a 1.3.4 first. Otherwise I'll have to change all the 1.3.4 to 1.4.0, and it will indeed be the time to break this as well :)

PR #7311, opened 2019-01-03
Rec: Set socket buf size for control socket.

### LINKED ISSUE #5745: doc update: add "net.local.dgram.recvspace=65536" to FreeBSD section.
- Program: Recursor 
 - Issue type: Feature request 

### Short description

Update docs and mention that for FreeBSD or for systems with too low datagram receive space setting, it's needed to increase it in order for rec_control to work correctly.

### Environment

 - Operating system: FreeBSD 11.1-STABLE
 - Software version: PowerDNS-recursor 4.0.6
 - Software source: FreeBSD ports collection 

### Expected behaviour

rec_control should work

### Actual behaviour

rec_control produces:
**Fatal: Timeout waiting for answer from control channel**

### Description

Add _net.local.dgram.recvspace=65536_ to FreeBSD-specific section in documentation.

omoerbeek:
Fixes #5745 without changing settings for all processes.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [X] documented the code

### DIFF
--- a/pdns/rec_channel.cc
+++ b/pdns/rec_channel.cc
@@ -42,9 +42,9 @@ static void setSocketBuffer(int fd, int optname, uint32_t size)
 
   if (psize > size)
     return;
-
-  if (setsockopt(fd, SOL_SOCKET, optname, (const void*)&size, sizeof(size)) < 0 )
-    throw PDNSException("Unable to raise socket buffer size: "+stringerror());
+  
+  // failure to raise is not fatal
+  setsockopt(fd, SOL_SOCKET, optname, (const void*)&size, sizeof(size));
 }
 
 
--- a/pdns/rec_channel.cc
+++ b/pdns/rec_channel.cc
@@ -54,6 +54,12 @@ int RecursorControlChannel::listen(const string& fname)
   if(bind(d_fd, (sockaddr*)&d_local,sizeof(d_local))<0) 
     throw PDNSException("Unable to bind to controlsocket '"+fname+"': "+stringerror());
 
+  // receive buf should be size of max datagram plus address size
+  int bufsz = 60*1024;
+  setsockopt(d_fd, SOL_SOCKET, SO_SNDBUF, &bufsz, sizeof(bufsz));
+  bufsz = 64*1024;
+  setsockopt(d_fd, SOL_SOCKET, SO_RCVBUF, &bufsz, sizeof(bufsz));
+  
   return d_fd;
 }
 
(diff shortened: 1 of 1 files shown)


pdns/rec_channel.cc
```diff
@@ -54,6 +54,12 @@ int RecursorControlChannel::listen(const string& fname)
   if(bind(d_fd, (sockaddr*)&d_local,sizeof(d_local))<0) 
     throw PDNSException("Unable to bind to controlsocket '"+fname+"': "+stringerror());
 
+  // receive buf should be size of max datagram plus address size
+  int bufsz = 60*1024;
```
rgacogne: Out of curiosity, why 60k for the send buffer and 64k for the receive one?

pdns/rec_channel.cc
```diff
@@ -54,6 +54,12 @@ int RecursorControlChannel::listen(const string& fname)
...
+  setsockopt(d_fd, SOL_SOCKET, SO_SNDBUF, &bufsz, sizeof(bufsz));
+  bufsz = 64*1024;
+  setsockopt(d_fd, SOL_SOCKET, SO_RCVBUF, &bufsz, sizeof(bufsz));
```
rgacogne: I'm a bit worried about the fact that this might actually _lower_ the buffer size, that's the case on several Linux boxes I tested, for example. Perhaps we could reuse the `setSocketReceiveBuffer()` and `setSocketSendBuffer()` functions from `pdns_recursor.cc` so that we only increase the size? We should probably lower the severity of the messages from `Error` to `Warning` if we do that, by the way.

pdns/rec_channel.cc
omoerbeek: At least on OpenBSD, the receive buffer must be bigger than the send buffer, since it is also used to receive the peer address internally. This might not be true on all systems, though.

pdns/rec_channel.cc
omoerbeek: Agreed, it is better to first check the current size. I'll make a 2nd revision.

pdns/rec_channel.cc
```diff
@@ -32,6 +32,32 @@ RecursorControlChannel::~RecursorControlChannel()
...
+  socklen_t len=sizeof(psize);
+
+  if (getsockopt(fd, SOL_SOCKET, optname, (void*)&psize, &len))
+    throw PDNSException("Unable to getsocket buffer size: "+stringerror());
+
+  if (psize > size)
+    return;
+
+  if (setsockopt(fd, SOL_SOCKET, optname, (const void*)&size, sizeof(size)) < 0 )
+    throw PDNSException("Unable to raise socket buffer size: "+stringerror());
```
rgacogne: Thanks for making this change! I'm not sure we should throw an exception if we can't raise the buffer size, though, because it can occur if the system has a low maximum buffer size, and a lower buffer size might not prevent us from successfully complete the operation while raising this exception will.

omoerbeek: This is a bit of a hard decision; e.g. on OpenBSD, the buffer sizes are pretty small by default, making e.g. rec_control help fail and possible get-all as well (once the numbers are grown a bit or another stat is added). On the other hand rec itself will probably run fine.  So I'll remove the exception.

PR #7326, opened 2019-01-03
Add type filter to search-data api. Closes #5430

### LINKED ISSUE #5430: search-data api call with type argument returning rrsets or comments or both or none
- Program: Authoritative 
 - Issue type: Feature request 

### Short description

Using search-data for a large number of zones takes forever due to recursive calls per id for records and comments. There should be an option for limiting either record or comment array or both or none as an option in the argument when calling search-data. 
eg: type=z will return zone info only, type=r returns only records and no comments, type=c returns comments only,  and type=a will return all 
'search-data?q=*&max=10000&type=z'  

i have 'working patch' I came up with quickly and by no means it's production ready code as I am not a C++ programmer but it solves my problem querying all the zones to compare when necessary without timing out.
https://jpst.it/104Oo

would be awesome if upstream can modify and implement it properly. 

### Environment

 - Operating system:  centos 
 - Software version:  4.0.1
 - Software source:  PowerDNS repository, compiled yourself and patched for above 

### Steps to reproduce

1. 
2. 
3. 

### Expected behaviour

api/v1/servers/localhost/'search-data?q=*&max=1000000'  | jq . returns all the data without timing out 

### Actual behaviour

times out 
### Usecase

only return top zone data and selectively choose the records 
### Description

only return top zone data

jonathaneen:
### Short description
This pull request adds a 'type' argument to the data-search api. Using this argument you can filter the api results on type, which can be 'all', 'zone', 'record' or 'comment'.
This PR fixes issue #5430 and is a rework of code provided by @mimianddaniel 

### Checklist
I have:
- [X] compiled this code
- [X] tested this code
- [X] documented the code

### DIFF
--- a/docs/http-api/swagger/authoritative-api-swagger.yaml
+++ b/docs/http-api/swagger/authoritative-api-swagger.yaml
@@ -436,6 +436,11 @@ paths:
           required: true
           description: 'Maximum number of entries to return'
           type: integer
+        - name: type
+          in: query
+          required: false
+          description: 'Type of data to search for, one of “all”, “zone”, “record”, “comment”'
+          type: string
       responses:
         '200':
           description: Returns a JSON array with results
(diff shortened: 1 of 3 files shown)


Habbie:
Ping!

zeha:
I've renamed type to object_type.

PR #7334, opened 2019-01-06
Travis integrated spelling checker - based on a whitelist

jsoref:
### Short description
Rough implementation of spell checking. Based on my hook for checkstyle/checkstyle

### Checklist
I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

I'm not really sure which part of the Travis system is the right place to put this hook, offhand, I'm pretty sure where I placed it is wrong, but it can be easily changed.

### Travis changes:
* `.travis.yml`: A non-matrix build called "none" is added.
* `travis.sh`: For none, a bunch of stuff is skipped (although this isn't re-indented, it could be), including the install and build phases.

### The `$spellchecker` directory has:
* `exclude.pl` -- use this to define files/paths that should be excluded -- note: the list can't be empty (arguably a bug, but if you don't want to exclude anything, you can just not call the script...)
* `rpm2cpio.sh` -- this script is used to unpack a standardized English word list -- this could be skipped if one likes the system `/usr/share/dict/words`
* `spelling-unknown-word-splitter.pl` this is the engine that does the word splitting / checks against the English word list (note: there's nothing that requires the word list to be "English")
* `whitelist.words` -- this is the list of "words" that are considered "ok"
* `unknown.words` -- when the script runs, it generates this file (which is compared against the white list)

### The `$temp` directory gets:
* `english.words` -- the retrieved English word list
* `words.rpm` -- an rpm package with the word list

### Notes
* I'm not tied to *any* of the paths or filenames (they're based on the ones that the checkstyle project picked).
* Currently this code retrieves `w` from my repository which means that in the rare instance where I improve it, those updates are included.
  * I can imagine some projects disliking this and preferring to maintain a static copy of the script instead.
  * I changed `w` specifically to address one edge involving the pdns repository: it has lots of dangling symlinks.

* I'm not providing tests for this code
  * I did some testing of this work here: https://travis-ci.org/jsoref/pdns/builds
* There are some comments in the code
  * The code is admittedly pretty dense
* There is only minimal documentation


jsoref:
[passing builds](https://travis-ci.org/jsoref/pdns/jobs/475940183) look roughly like this: 
```
No new words and misspellings found.
The command "./build-scripts/travis.sh" exited with 0.
```
[failing builds](https://travis-ci.org/jsoref/pdns/jobs/475939650) will look roughly like this: 
```
New misspellings found, please review:
geoipdatabase
If you are ok with the output of this run, you will need to
patch build-scripts/jsoref-spellchecker/whitelist.words <<EOF
@@ -2814,0 +2815 @@
+geoipdatabase
EOF
```

pieterlexis:
I like this a lot :), do let us know if it is finished!

jsoref:
Oh, it's ready, minus whatever changes you request

PR #7335, opened 2019-01-07
Tell people to use --with-modules=... when they use --without-mysql

### LINKED ISSUE #5140: Auth `./configure --without-mysql` should disable the gmysql module
- Program: Authoritative 
 - Issue type: Bug report 

### Short description

The gmysql module is built by default. I expected that `--without-mysql` would skip everything that requires mysql, but instead the configure script still checks for mysql and exits with an error.

### Environment

 - Software version: master 1efb94eda4f8500c3568c149b9ca4ca5fa74e1fc
 - Software source: github

### Steps to reproduce

1. `./configure --disable-mysql`

### Expected behaviour

Do not build any mysql modules.

### Actual behaviour

The configure scripts do no properly check for the value 'no', which is set by `--without-mysql`:

```
checking for no/bin/mysql_config... configure: error: not found
```

jsoref:
closes #5140

### Short description
Detect `--without-mysql` and suggest `--with-modules="bind random" --with-dyn-modules="pipe"`

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

fwiw, this isn't my preference, I'd rather we just remove `gmysql` from `modules` when someone uses `--without-mysql`, but...

```sh
$ ./configure --without-mysql --with-libcrypto=/usr/local/opt/openssl 2>&1|tail -5
checking for arc4random... yes
checking whether to enable verbose logging... no
checking whether to enable PKCS11 support... no
checking whether to enable experimental GSS-TSIG support... no
configure: error: instead of --without-mysql try --with-modules="bind random" --with-dyn-modules="pipe"
```

### DIFF
--- a/m4/pdns_with_mysql.m4
+++ b/m4/pdns_with_mysql.m4
@@ -2,6 +2,12 @@ AC_DEFUN([PDNS_WITH_MYSQL],[
   AC_ARG_WITH([mysql],
     [AS_HELP_STRING([--with-mysql=<path>], [root directory path of MySQL installation])],
     [
+      if test "$withval" = "no"; then
+        modules_without_gmysql=$(echo $modules|sed -e 's/gmysql//;s/  */ /g;')
+        dynmodules_without_gmysql=$(echo $dynmodules|sed -e 's/gmysql//;s/  */ /g;')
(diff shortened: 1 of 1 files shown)


pieterlexis:
As discussed on IRC, please make `--without-mysql` an error condition as we only attempt to find it when we really need it.

PR #7340, opened 2019-01-07
auth: error on DNSSEC default misconfiguration

pieterlexis:
### Short description
This commit ensures `pdns_server` does not start when there is a
misconfiguration in default-[kz]sk-algorithm:

 * Either is set to an unknown algorithm
 * When using RSA, key size is unset
 * zsk is set but ksk is not
 * zsk is not set to the same as the ksk

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/common_startup.cc
+++ b/pdns/common_startup.cc
@@ -557,33 +557,41 @@ void mainthread()
   }
   catch(...) {}
 
-  // Some sanity checking on default key settings
-  bool hadKeyError = false;
-  for (const string& algotype : {"ksk", "zsk"}) {
-    int algo, size;
-    if (::arg()["default-"+algotype+"-algorithm"].empty())
-      continue;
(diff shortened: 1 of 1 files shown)


pdns/common_startup.cc
```diff
@@ -558,16 +558,32 @@ void mainthread()
...
+    }
+  }
+  if (hadKeyError) {
+    exit(1);
+  }
+  if (::arg()["default-ksk-algorithm"].empty() && !::arg()["default-zsk-algorithm"].empty()) {
+    g_log<<Logger::Error<<"Error: default-zsk-algorithm is set, but default-ksk-algorithm is not set."<<endl;
+    exit(1);
+  }
+  if (!::arg()["default-zsk-algorithm"].empty() && ::arg()["default-zsk-algorithm"] != ::arg()["default-ksk-algorithm"]) {
```
Habbie: This `!=` relies on string representations, why not use `int algo` from above?

Issue #7357, opened 2019-01-09
Response statistics on auth api


tduboys:
- Program: Authoritative
 - Issue type: Feature request

### Short description
Since 4.2 (tested on the alpha1 version), there is more statistics on the API statistics entrypoint.
I can now see queries-by-qtype that wasnt present on 4.1.5 version.

But I cant see any statistics to know what kind of response my auth serves to clients.
It should be cool to add more statistics like response by rcode, type, or something like that.

### Usecase
The idea is to graph more metrics, and detect if users do some wrong things.
I already can graph all type of queries by qtype, but I dont have information about responses

### Description
On the statistics api entrypoint, add more statistics like, for example : 
```
{
    "name": "responses-by-rcode",
    "type": "MapStatisticItem",
    "value": [
      {
        "name": "NOERROR",
        "value": "27"
      },
      {
        "name": "NXDOMAIN",
        "value": "4"
      },
      {
        "name": "REFUSED",
        "value": "12"
      },
    ]
  }
```

PR #7359, opened 2019-01-10
Add Rcode response statistics on API

tduboys:
- Add new statistic variable about rcode responses
- Add metrics about RCode on statistics API entrypoint
- Rename queries-by-qtype as response-by-qtype on API stats because it's calculated on response side

### Short description
Following Issue #7357, new statistic counter about rcode responses, published on the API on statistics entrypoint

Point to discuss : Name of variables.
On /statistics entrypoint, I've renamed queries-by-qtype as response-by-qtype, and named my new variable as response-by-rcode, because all of the work is made on responsestats.cc before sending the response. It's confusing to call the metrics as « queries » because it supposes that the metrics is measured when the querie cames and not before sending the response.

### Checklist
I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/responsestats.cc
+++ b/pdns/responsestats.cc
@@ -8,7 +8,7 @@
 
 #include "dnsparser.hh"
 
-ResponseStats::ResponseStats() :   d_qtypecounters(new std::atomic<unsigned long>[65536])
+ResponseStats::ResponseStats() :   d_qtypecounters(new std::atomic<unsigned long>[65536]), d_rcodecounters(new std::atomic<unsigned long>[256])
 {
   d_sizecounters.push_back(make_pair(20,0));
   d_sizecounters.push_back(make_pair(40,0));
@@ -21,6 +21,8 @@ ResponseStats::ResponseStats() :   d_qtypecounters(new std::atomic<unsigned long
   d_sizecounters.push_back(make_pair(std::numeric_limits<uint16_t>::max(),0));
   for(unsigned int n =0 ; n < 65535; ++n)
     d_qtypecounters[n] = 0;
+  for(unsigned int n =0 ; n < 256; ++n)
+    d_rcodecounters[n] = 0;
 }
 
 ResponseStats g_rs;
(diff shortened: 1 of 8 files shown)


pdns/responsestats.cc
```diff
@@ -8,7 +8,7 @@
 
 #include "dnsparser.hh"
 
-ResponseStats::ResponseStats() :   d_qtypecounters(new std::atomic<unsigned long>[65536])
+ResponseStats::ResponseStats() :   d_qtypecounters(new std::atomic<unsigned long>[65536]), d_rcodecounters(new std::atomic<unsigned long>[65536])
```
rgacogne: We don't need 65536 entries for the response code.
```suggestion
ResponseStats::ResponseStats() :   d_qtypecounters(new std::atomic<unsigned long>[65536]), d_rcodecounters(new std::atomic<unsigned long>[256])
```

pdns/ws-api.cc
```diff
@@ -188,7 +189,7 @@ void apiServerStatistics(HttpRequest* req, HttpResponse* resp) {
 
     doc.push_back(Json::object {
       { "type", "MapStatisticItem" },
-      { "name", "queries-by-qtype" },
+      { "name", "response-by-qtype" },
       { "value", values },
```
rgacogne: Perhaps we could add a new `response-by-qtype` field while deprecating the existing `queries-by-qtype` one in 4.2.0, so it doesn't break existing setups right away? It probably doesn't matter much, though.

tduboys: Is it needed to deprecate the existing `queries-by-qtype` ? This value wasn't on the last stable 4.1.5 release. Maybe just informing the alpha1 users that this value changes ?
But if we need to deprecate it, what's the best way to do ?

pdns/responsestats.cc
tduboys: Thanks, I didnt noticed that 65536 was because it's the uint16_t max value. rcode is uint8_t :)
I just updated this file

pdns/ws-api.cc
rgacogne: Sorry, I had in mind that this metric already existed in 4.1.x but I was wrong. Please forget that comment :-)

tduboys:
Looks that there is a new spellcheck on the master, and I need to rebase my branch and add a word on the whitelist. (but the master is already broken on this test)

pieterlexis:
> not sure its worth breaking existing users

I would be fine with a breaking change in the statistics endpoint in 4.2.0, if it is documented. @Habbie?

zeha:
Oh if that wasn't in 4.1 then sure!

tduboys:
I just rebased my branch to fix conflict about unit test and spellcheck.

tduboys:
> Oh if that wasn't in 4.1 then sure!

I confirm. The renamed endpoint was added on auth-4.2-alpha1 and not present on the last auth-4.1.x version.

pieterlexis:
Please add some documentation on the name-change and these new statistics and then we'll merge it :+1:

tduboys:
> Please add some documentation on the name-change and these new statistics and then we'll merge it +1

Hi @pieterlexis 
I will add more docs on the statistics entrypoint https://doc.powerdns.com/authoritative/http-api/statistics.html
But I didnt see any docs about the old value I've renamed, I dont know where to put some infos about this change.
Any idea ?

tduboys:
Hi,
I've pushed things in changelog and /docs/http-api/{statistics.rst, authoritative-api-swagger.yaml} . 
It adds informations about the statistics endpoint (the current page is quite empty : https://doc.powerdns.com/authoritative/http-api/statistics.html )

If anyone have idea about what I can do more, please ask :)

PR #7404, opened 2019-01-21
rec: Sanitize records received before doing anything else

rgacogne:
### Short description

Stricter version of https://github.com/PowerDNS/pdns/pull/7258

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/pdns/syncres.cc
+++ b/pdns/syncres.cc
@@ -2078,6 +2078,12 @@ void SyncRes::sanitizeRecords(const std::string& prefix, LWResult& lwr, const DN
         continue;
       }
 
+      if (!(lwr.d_aabit || wasForwardRecurse)) {
+        LOG(prefix<<"Removing irrelevant record '"<<rec->d_name<<"|"<<DNSRecordContent::NumberToType(rec->d_type)<<"|"<<rec->d_content->getZoneRepresentation()<<"' in the "<<(int)rec->d_place<<" section received from "<<auth<<endl);
+        rec = lwr.d_records.erase(rec);
+        continue;
+      }
+
       if (!haveAnswers) {
         if (lwr.d_rcode == RCode::NXDomain) {
           isNXDomain = true;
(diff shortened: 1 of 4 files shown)


pdns/syncres.cc
```diff
@@ -1969,6 +1969,162 @@ vState SyncRes::validateRecordsWithSigs(unsigned int depth, const DNSName& qname
...
+    }
+
+    /* dealing with the records in additional */
+    if (rec->d_place == DNSResourceRecord::ADDITIONAL && rec->d_type != QType::A && rec->d_type != QType::AAAA && rec->d_type != QType::RRSIG) {
+      LOG(prefix<<"Removing irrelevant record '"<<rec->d_name<<"|"<<DNSRecordContent::NumberToType(rec->d_type)<<"|"<<rec->d_content->getZoneRepresentation()<<"' in the "<<(int)rec->d_place<<" section received from "<<auth<<endl);
+      rec = lwr.d_records.erase(rec);
+      continue;
+    }
+
+    if (rec->d_place == DNSResourceRecord::ADDITIONAL && allowedAdditionals.count(rec->d_name) == 0) {
```
pieterlexis: I _think_ we might need to check if the allowedAdditional isPartOf `auth`?

rgacogne: You mean if `rec->d_name` ispartOf `auth`? If so we already checked on line 2030, unless I'm missing your point entirely.

rgacogne: But how would that `A` record not be discarded by the check on line 2030? I mean sure, we would accept the `SRV` record and so add `server.jabber.example` to the `allowedAdditionals` whitelist, but the additional `A` record should be rejected for not being part of `auth` well before being checked against that whitelist, right?

pieterlexis: ah, yes. You're right!

PR #7406, opened 2019-01-22
Add EDNS unknown version handling

alenichev:
### Short description
Respond with RCODE=BADVERS on EDNS version >0 (as per rfc6891).
Turned out to be a rule to match on the EDNS version and a action to return ERCODE.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


Habbie:
Thank you for this work. Can you please update your message based on our pull request template at https://github.com/PowerDNS/pdns/blob/master/.github/PULL_REQUEST_TEMPLATE.md ?

Also, it appears your work breaks a bunch of our tests, as reported by Travis.

pdns/dnsdist-ecs.cc
```diff
@@ -741,3 +741,45 @@ bool queryHasEDNS(const DNSQuestion& dq)
...
+  uint16_t qtype = (reinterpret_cast<const unsigned char*>(packet)[pos])*256 + reinterpret_cast<const unsigned char*>(packet)[pos+1];
+  pos += DNS_TYPE_SIZE;
+  pos += DNS_CLASS_SIZE;
+
+  if (qtype != QType::OPT || (pos + EDNS_EXTENDED_RCODE_SIZE + 1) >= dq.len) {
+    return 0;
+  }
+
+  const uint8_t* z = reinterpret_cast<const uint8_t*>(packet) + pos + EDNS_EXTENDED_RCODE_SIZE;
+  return 0x100 * (*z) + *(z+1);
```
rgacogne: This doesn't look right, the `EDNS` version field is only one byte, so this is very likely a bad c/p.

pdns/dnsdist.cc
```diff
@@ -1403,6 +1403,12 @@ static void processUDPQuery(ClientState& cs, LocalHolders& holders, const struct
       return;
     }
 
+    if (queryHasEDNS(dq) && getEDNSVersion(dq) > 0) {
+        dq.dh->qr = true;
+        dq.dh->rcode = (16 & 0xF);
+        dq.ednsRCode = ((16 & 0xFFF0)>>4); // set rcode to BADVERS
+    }
```
rgacogne: Note that this only handles queries received over `UDP`, not `TCP`.

pdns/dnsdist.cc
```diff
@@ -1403,6 +1403,12 @@ static void processUDPQuery(ClientState& cs, LocalHolders& holders, const struct
       return;
     }
 
+    if (queryHasEDNS(dq) && getEDNSVersion(dq) > 0) {
+        dq.dh->qr = true;
+        dq.dh->rcode = (16 & 0xF);
+        dq.ednsRCode = ((16 & 0xFFF0)>>4); // set rcode to BADVERS
```
rgacogne: I'd really prefer the use of `ERCode::BADVERS` over `16`. I'm also not sure why you set the "regular" `rcode` here.

pdns/dnsdistdist/dnsdist-rules.hh
```diff
@@ -882,6 +882,49 @@ private:
...
+    }
+
+    if (optStart < dq->len && packetStr.at(optStart) != 0) {
+      // OPT RR Name != '.'
+      return false;
+    }
+    EDNS0Record edns0;
+    static_assert(sizeof(EDNS0Record) == sizeof(uint32_t), "sizeof(EDNS0Record) must match sizeof(uint32_t) AKA RR TTL size");
+    // copy out 4-byte "ttl" (really the EDNS0 record), after root label (1) + type (2) + class (2).
+    memcpy(&edns0, packet + optStart + 5, sizeof edns0);
```
rgacogne: There is a lot of code duplication with `ERCodeRule`, I think we need to turn that into a function that would be better off in `dnsdist-ecs.cc`. But we can do that refactoring after this PR has been merged if needed.

Issue #7407, opened 2019-01-22
cryptokeys missing when DNSSEC-enabled zone is deleted/recreated quickly


peterthomassen:
- Program: Authoritative
 - Issue type: Bug report

### Short description
All this is using the API.

- When deleting a DNSSEC-enabled zone and recreating it again within 30 seconds, cryptokeys will be missing (both via `GET` to the `cryptokeys` endpoint, and also in the `cryptokeys` table of the database).
- Trying again (by deleting and recreating once more) reliably produces a zone that has keys.
- Trying even more often leads to an alternating pattern.
- Whenever there were more than 30 seconds between the attempts, keys are created reliably.

### Environment
 - Operating system: Debian stretch
 - Software version: 4.1.5-1pdns.stretch
 - Software source: PowerDNS repository

### Steps to reproduce
1. Create a shell script with the following content (adjust hostname if it is not `pdns`):
    ```
    curl -s -X POST http://pdns:8081/api/v1/servers/localhost/zones -H "Content-Type: application/json" -H "X-API-Key: $TOKEN" -d '{"name": "test.de.", "kind": "MASTER", "dnssec": true, "nsec3param": "1 0 127 abd9404aa407e63d", "nameservers": ["ns1.test.com.", "ns2.test.com."]}' | wc
    curl -s -X GET http://pdns:8081/api/v1/servers/localhost/zones/test.de./cryptokeys -H "Content-Type: application/json" -H "X-API-Key: $TOKEN" | wc
    curl -s -X DELETE http://pdns:8081/api/v1/servers/localhost/zones/test.de. -H "Content-Type: application/json" -H "X-API-Key: $TOKEN" | wc
    ```

2. `export TOKEN=...`, then run the script with various delays
3. Observe how the second line varies (it has the response length in the last column)

Remove `| wc` to see the full (confusing) output.

### Expected behaviour
Keys should always be present, i.e. second line should never have last column equal to 2.

### Actual behaviour
Last column of second line often is 2, i.e. the `cryptokeys` endpoint returned `[]` (keys missing).

Habbie:
```
# dnssec-key-cache-ttl	Seconds to cache DNSSEC keys from the database
# dnssec-key-cache-ttl=30
```

this default (that you did not override) looks like a smoking gun

peterthomassen:
Setting `dnssec-key-cache-ttl=0` indeed is a workaround for this problem. Thanks for the very very quick response!

[...]

PR #7408, opened 2019-01-22
[WIP] auth API, DNSSEC (key) caching

Habbie:
### Short description
#7407 demonstrates our DNSSEC key cache persisting over domain deletion, and thereby preventing the correct setup of DNSSEC when immediately adding that domain again.

This PR mitigates the specific symptom. It's obvious that the key cache shouldn't have held on to those keys when the domain was deleted, so there is more work to be done here.

Please feel free to push (if you have access) or PR (if you don't) to this branch.

### Checklist

I have:
- [x] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dbdnsseckeeper.cc
+++ b/pdns/dbdnsseckeeper.cc
@@ -57,12 +57,12 @@ bool DNSSECKeeper::doesDNSSEC()
   return d_keymetadb->doesDNSSEC();
 }
 
-bool DNSSECKeeper::isSecuredZone(const DNSName& zone) 
+bool DNSSECKeeper::isSecuredZone(const DNSName& zone, bool useCache)
 {
   if(isPresigned(zone))
     return true;
 
-  keyset_t keys = getKeys(zone); // does the cache
+  keyset_t keys = getKeys(zone, useCache);
 
   for(keyset_t::value_type& val :  keys) {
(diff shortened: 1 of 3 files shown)


Habbie:
This PR adds a caching bool to isSecuredZone, like a few other calls already have. Perhaps caching should be a property of the object, chosen when instantiating the Keeper? That way the API could operate fully uncached.

rgacogne:
Perhaps we should just clear the caches on zone deletion, something like https://github.com/PowerDNS/pdns/commit/5a4fa722749dbd5daa80b3a3f8feba6e4662aec0 (untested)?

pieterlexis:
> Perhaps we should just clear the caches on zone deletion, something like [5a4fa72](https://github.com/PowerDNS/pdns/commit/5a4fa722749dbd5daa80b3a3f8feba6e4662aec0) (untested)?

Perhaps check first if the domain was signed so we don't wipe the cache needlessly :)

rgacogne:
> Perhaps check first if the domain was signed so we don't wipe the cache needlessly :)

I pondered that, but I'm actually not sure that check is less expensive that just trying to clear the cache no matter what :)

Habbie:
Fixed in #7529 instead.

PR #7410, opened 2019-01-23
Ignore Path MTU Discovery on UDP server socket

rgacogne:
### Short description

It might help prevent Path MTU poisoning attacks.
Untested, don't merge it yet!

### Checklist

I have:
- [x] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/iputils.cc
+++ b/pdns/iputils.cc
@@ -136,6 +136,25 @@ int SSetsockopt(int sockfd, int level, int opname, int value)
   return ret;
 }
 
+void setSocketIgnorePMTU(int sockfd)
+{
+#ifdef IP_PMTUDISC_OMIT
+  /* Linux 3.15+ has IP_PMTUDISC_OMIT, which discards PMTU information to prevent
+     poisoning, but still allows fragmentation if the packet size exceeds the
+     outgoing interface MTU, which is good.
+  */
+  try {
+    SSetsockopt(sockfd, IPPROTO_IP, IP_MTU_DISCOVER, IP_PMTUDISC_OMIT);
+    return;
+  }
+  catch(const std::exception& e) {
+    /* failed, let's try IP_PMTUDISC_DONT instead */
+  }
+#endif /* IP_PMTUDISC_OMIT */
+
+  /* IP_PMTUDISC_DONT disables Path MTU discovery */
+  SSetsockopt(sockfd, IPPROTO_IP, IP_MTU_DISCOVER, IP_PMTUDISC_DONT);
+}
 
 bool HarvestTimestamp(struct msghdr* msgh, struct timeval* tv) 
 {
--- a/pdns/iputils.cc
+++ b/pdns/iputils.cc
@@ -138,6 +138,7 @@ int SSetsockopt(int sockfd, int level, int opname, int value)
 
 void setSocketIgnorePMTU(int sockfd)
 {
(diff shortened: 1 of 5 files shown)


pdns/iputils.cc
```diff
@@ -136,6 +136,25 @@ int SSetsockopt(int sockfd, int level, int opname, int value)
...
+    SSetsockopt(sockfd, IPPROTO_IP, IP_MTU_DISCOVER, IP_PMTUDISC_OMIT);
+    return;
+  }
+  catch(const std::exception& e) {
+    /* failed, let's try IP_PMTUDISC_DONT instead */
+  }
+#endif /* IP_PMTUDISC_OMIT */
+
+  /* IP_PMTUDISC_DONT disables Path MTU discovery */
+  SSetsockopt(sockfd, IPPROTO_IP, IP_MTU_DISCOVER, IP_PMTUDISC_DONT);
```
rgacogne: We should check that `IP_MTUDISC_DONT` is defined. Note that unbound set the don't fragment flag on systems that don't have `IP_MTUDISC_DONT`: https://github.com/NLnetLabs/unbound/blob/master/services/listen_dnsport.c#L571

rgacogne:
As a datapoint, Knot decided against doing this: https://gitlab.labs.nic.cz/knot/knot-dns/issues/467

pdns/iputils.cc
rgacogne: Bind is doing the same: https://gitlab.isc.org/isc-projects/bind9/blob/master/lib/isc/unix/socket.c#L2396

mnordhoff:
Knot decided against it in 2016 when fragmentation attacks weren't in vogue, though.

pdns/iputils.cc
rgacogne: Added the missing check for `IP_MTUDISC_DONT`. I'm still undecided about setting DF=1.

rgacogne:
Rebased to fix a conflict.

rgacogne:
As pointed out by @hdais `NSD` did bite the bullet as well recently: https://www.nlnetlabs.nl/bugs-script/show_bug.cgi?id=4235

rgacogne:
Move this PR into the rec-4.2.0-beta1 milestone. Now is the time to speak if you are against this change ;-)

jedisct1:
Hi Remi,

That change seems to be what rmade DNSCrypt servers running `dnsdist` unreliable after they upgraded to version 1.4.0.

As you know since you implemented the protocol, DNSCrypt requires questions sent over UDP to be as large as responses, using padding.

Blocking fragmented questions prevents large responses from being received.

Ignoring fragments in responses from authoritative servers fine, but even with unencrypted queries, I'm not sure that there is any value in dropping fragments on the server socket. A fragment attack would just allow the question to be modified, but a stub resolver would ignore a response for a different query.

rgacogne:
Hi Frank,

Thanks a lot for reporting this! I have to admit I had completely forgotten that propriety of `DNSCrypt` when I submitted this change, sorry about that!
I'll re-evaluate whether it makes sense to keep setting `setSocketIgnorePMTU()` on our incoming sockets. At the very least I'll disable it for `DNSCrypt` binds.

jedisct1:
Thank you so much Remi!

/cc @welwood08

PR #7417, opened 2019-01-24
While we all enjoy grate spelling, this was just to much hassle

ahupowerdns:
The spellchecker was yet an additional hurdle in getting people involved. I appreciate the good intent but this was one bridge too far.


jsoref:
Sorry to hear. Thanks for trying.

Notes:
1. You could scope it to only check docs files
1. You could make the reporting non fatal and let someone else clean up as a chore.

PR #7421, opened 2019-01-25
Hack in support for GitHub access tokens.

aerique:
It's not very pretty but I'm not going to spend more time on it today.

`git blame ./build-scripts/changelog-from-pr.py` > reviewers

Possible improvements:
- merge `if access_token …` occurences back into one assignment
- `var.get(attribute, None)` → `var.get(attribute)` ?

### DIFF
--- a/build-scripts/changelog-from-pr.py
+++ b/build-scripts/changelog-from-pr.py
@@ -11,6 +11,8 @@ argp.add_argument('--oneline', action='store_true',
                   help='Make one-lined changelog entries (for 4.0 and older)')
 argp.add_argument('--username',
                   help='Use the specified username for Basic Authentication to the GitHub API, allowing an higher rate limit')
+argp.add_argument('--access_token',
+                  help='Use API access token instead of username & password combination')
 argp.add_argument('pullrequest', metavar='PULL_REQUEST', nargs='+',
                   help='Make changelogs for these Pull Request #\'s')
 arguments = argp.parse_args()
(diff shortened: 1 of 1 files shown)


build-scripts/changelog-from-pr.py
```diff
@@ -11,6 +11,8 @@
                   help='Make one-lined changelog entries (for 4.0 and older)')
 argp.add_argument('--username',
                   help='Use the specified username for Basic Authentication to the GitHub API, allowing an higher rate limit')
```
Habbie: s/an higher/a higher/ while you're here?

Habbie:
```suggestion
                  help='Use the specified username for Basic Authentication to the GitHub API, allowing a higher rate limit')
```

aerique:
> See comment posted earlier. Besides that, if this has been tested with and without a token, just merge it.

I don't think other credentials besides the token are possible anymore for GitHub.  I'd have to double-check that.

Without credentials still works but you'll run into rate limiting really quickly.

Habbie:
> I don't think other credentials besides the token are possible anymore for GitHub. I'd have to double-check that.

"nothing" and "token" are the two ways I mean.


> Without credentials still works but you'll run into rate limiting really quickly.

Understood!

aerique:
Ah right!  I'll give it a go later today since I need it for dstore and then I'll merge it.

PR #7422, opened 2019-01-25
LUA: report nested exception

Habbie:
### Short description
Before:
```
Jan 25 12:53:43 Lua record reported: Exception thrown by a callback function called by Lua
```
After:
```
Jan 25 19:42:24 Lua record reported: Exception thrown by a callback function called by Lua: Unable to convert presentation address '10.0.0.0/8'
```

Ideally this would also return a backtrace, but this is a start.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


Habbie:
The code I added here exists in ~five more places, we might want to centralise the unnest handling (and, for kicks, while we're doing that, deal with more than one level of nesting).

PR #7428, opened 2019-01-29
Reduce systemcall usage in protobuf logging

ahupowerdns:
Since Spectre/Meltdown, system calls have become more expensive.  In
addition, relevant versions of glibc turn out to implement pthread_cond_wait
and pthread_cond_signal in such a way that they use multiple system calls always.
There is an optimization in glibc to improve this but it is disabled.

This new setup changes our protobuf logging so it amortizes system calls so we perform
far less than one call per message.

Note that our previous RemoteLogger was configured in terms of how many
*messages* it would buffer. Our new code is configured in terms of how many
*bytes*. I have multiplied the configured numbers by 100 elsewhere (recursor
config, dnsdist config) to sort of maintain parity.

In addition, the old RemoteLogger would buffer messages while there was no
connection available. We no longer do this.

Finally new, every 'reconnectTimeout' seconds we will flush our buffers
opportunistically to not keep people waiting.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/namespaces.hh
+++ b/pdns/namespaces.hh
@@ -22,7 +22,6 @@
 #ifndef PDNS_NAMESPACES_HH
 #define PDNS_NAMESPACES_HH
 #include <boost/tuple/tuple.hpp>
-
 #include <boost/shared_array.hpp>
 #include <boost/scoped_array.hpp>
 #include <boost/optional.hpp>
@@ -78,4 +77,5 @@ using boost::trim_right_copy_if;
 using boost::equals;
 using boost::ends_with;
 using boost::iends_with;
+
 #endif
--- a/pdns/remote_logger.cc
+++ b/pdns/remote_logger.cc
@@ -1,6 +1,7 @@
 #include <unistd.h>
 #include "threadname.hh"
 #include "remote_logger.hh"
+#include <sys/uio.h>
 #ifdef HAVE_CONFIG_H
 #include "config.h"
 #endif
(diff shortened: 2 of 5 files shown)


pdns/namespaces.hh
```diff
@@ -78,4 +78,18 @@ using boost::trim_right_copy_if;
...
 using boost::ends_with;
 using boost::iends_with;
+
+// apple compiler somehow has string_view even in c++11!
+#if __cplusplus < 201703L && !defined(__APPLE__)
+#if BOOST_VERSION > 105400
+#include <boost/utility/string_view.hpp>
+using boost::string_view;
+#else
+#include <boost/utility/string_ref.hpp>
```
rgacogne: Looks like this requires Boost 1.53, we previously required 1.42+ with protobuf. Given that we already have 
a reference to an existing string in every place where we currently use a `string_view`, perhaps we can skip that for now?

pdns/remote_logger.cc
```diff
@@ -31,73 +87,70 @@ bool RemoteLogger::reconnect()
...
+      }
     }
-    catch(const std::runtime_error& e) {
-#ifdef WE_ARE_RECURSOR
-      g_log<<Logger::Info<<"Error sending data to remote logger "<<d_remote.toStringWithPort()<<": "<< e.what()<<endl;
-#else
-      vinfolog("Error sending data to remote logger (%s): %s", d_remote.toStringWithPort(), e.what());
-#endif
-      busyReconnectLoop();
+    else if(reconnect()) { // if it was zero, it will remain zero, we are the only ones setting it!
```
rgacogne: Perhaps we could try to reconnect right away if we just closed the connection? It feels especially bad to wait one second since we are no buffering messages when we are not connected.

pdns/remote_logger.hh
```diff
@@ -30,6 +30,31 @@
...
+
+   The filedescriptor can be in non-blocking mode.
+   
+   This class is not threadsafe.
+*/
+
+class CircularWriteBuffer
+{
+public:
+  explicit CircularWriteBuffer(int fd, int size) : d_fd(fd), d_buffer(size)
```
rgacogne:
```suggestion
  explicit CircularWriteBuffer(int fd, size_t size) : d_fd(fd), d_buffer(size)
```

PR #7431, opened 2019-01-29
pass empty response

alenichev:
### Short description
pass empty responses without errors (BADVERS without soa for example).

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


rgacogne:
Allowing a `qdcount` of 0 is not necessary for a response without `SOA`, what this does allow is a `NoError` response without the query. I'm not sure we should allow that, do you have an example of a server returning that kind of response?

alenichev:
Sure - it is a nsd, latest version. Here is result of dig against it, which give us no errors in isc's "EDNS Compliance Tester":

    $ dig -p 534 +edns=200 soa vk.com @localhost
    ../../../../lib/isc/unix/net.c:581: sendmsg() failed: Operation not permitted
    
    ; <<>> DiG 9.10.3-P4-Debian <<>> -p 534 +edns=200 soa vk.com @localhost
    ;; global options: +cmd
    ;; Got answer:
    ;; ->>HEADER<<- opcode: QUERY, status: BADVERS, id: 44817
    ;; flags: qr rd; QUERY: 0, ANSWER: 0, AUTHORITY: 0, ADDITIONAL: 1
    
    ;; OPT PSEUDOSECTION:
    ; EDNS: version: 0, flags:; udp: 4096
    ;; Query time: 0 msec
    ;; SERVER: 127.0.0.1#534(127.0.0.1)
    ;; WHEN: Tue Jan 29 19:41:49 MSK 2019
    ;; MSG SIZE  rcvd: 23

dnsdist before patch:

    $  dig +edns=200 soa vk.com @localhost
    ../../../../lib/isc/unix/net.c:581: sendmsg() failed: Operation not permitted
    
    ; <<>> DiG 9.10.3-P4-Debian <<>> +edns=200 soa vk.com @localhost
    ;; global options: +cmd
    ;; connection timed out; no servers could be reached

and after:

    $  dig +edns=200 soa vk.com @localhost
    ../../../../lib/isc/unix/net.c:581: sendmsg() failed: Operation not permitted
    
    ; <<>> DiG 9.10.3-P4-Debian <<>> +edns=200 soa vk.com @localhost
    ;; global options: +cmd
    ;; Got answer:
    ;; ->>HEADER<<- opcode: QUERY, status: BADVERS, id: 13180
    ;; flags: qr rd; QUERY: 0, ANSWER: 0, AUTHORITY: 0, ADDITIONAL: 1
    
    ;; OPT PSEUDOSECTION:
    ; EDNS: version: 0, flags:; udp: 4096
    ;; Query time: 0 msec
    ;; SERVER: 127.0.0.1#53(127.0.0.1)
    ;; WHEN: Tue Jan 29 19:47:11 MSK 2019
    ;; MSG SIZE  rcvd: 23

rgacogne:
Alright, I'm not too happy about this but I can guess we need to allow such responses if `NSD` is sending them. This does increase the risk of collision a lot, though, since we can't check that the qname, qtype and qclass match the ones in the initial query, so perhaps we should turn that into an option and not accept it by default.
We also need to remove a regression test that checks explicitly that we don't let these kind of empty responses go through:
```
FAIL: Basics: Header-only NoError response should be dropped
  File "/home/travis/build/PowerDNS/pdns/regression-tests.dnsdist/test_Basics.py", line 403, in testHeaderOnlyNoErrorResponse
    self.assertEquals(receivedResponse, None)
AssertionError: <DNS message, ID 65339> != None
```

rgacogne:
Would you mind making this optional? Then we can worry about fixing the failing test, or I can do it if you prefer.

alenichev:
new patch added; not sure about option name - setAllowEmptyResponse()

pdns/dnsdist.cc
```diff
@@ -222,7 +223,7 @@ bool responseContentMatches(const char* response, const uint16_t responseLen, co
   }
 
   if (dh->qdcount == 0) {
-    if (dh->rcode != RCode::NoError && dh->rcode != RCode::NXDomain) {
+    if (dh->rcode != RCode::NXDomain && g_allowEmptyResponse) {
```
rgacogne:
```suggestion
    if ((dh->rcode != RCode::NoError && dh->rcode != RCode::NXDomain) || g_allowEmptyResponse) {
```

PR #7438, opened 2019-01-31
travis auth: install dnsmasq, configure it on 127.0.0.53

Habbie:
### Short description
This introduces some local caching, and, I hope, a bit of retrying on ALIAS lookups, because we don't do that internally.

I considered hardcoding the google-public-dns-a.google.com./8.8.8.8 record we use for testing, but I'd like to see if just adding dnsmasq as a layer improves the situation without hardcoding anything. If it turns out we do need to add that static record, it will be just one extra item around line 336 of `build-scripts/travis.sh` at that time.

### Checklist

I have:
- [ ] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/build-scripts/travis.sh
+++ b/build-scripts/travis.sh
@@ -328,6 +328,19 @@ install_auth() {
     faketime"
   run "sudo touch /etc/authbind/byport/53"
   run "sudo chmod 755 /etc/authbind/byport/53"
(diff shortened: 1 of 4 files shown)


ahupowerdns:
I think this is worth a try.

PR #7451, opened 2019-02-05
Make sure that the systemd binaries are available during the build

rgacogne:
### Short description

Otherwise the detection of systemd's sandboxing features does not work properly, since it uses `systemctl` to determine the version of systemd.
A better option might be to use `pkg-config --modversion systemd` to retrieve the version but that's a fix for a later time.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


zeha:
debian bits lgtm

PR #7460, opened 2019-02-07
auth: speedup getUpdatedMasters() for the gsql backends

mind04:
### Short description
The query is optimized for the existing schemas. If you join on records.domain_id = domains.id it is possible to speed-up things a little bit more, but this requires an index change. I'm not sure this is worth the trouble and the possible slowdown when this index is missing.

Test with 750K zones

Before:
```
Feb 07 09:27:03 getUpdatedMasters(): 87087980 usec to execute
Feb 07 09:27:03 750000 domains for which we are master need notifications
```

After:
```
Feb 07 09:30:43 getUpdatedMasters(): 7002059 usec to execute.
Feb 07 09:30:43 750000 domains for which we are master need notifications
```

### Checklist
I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/modules/gmysqlbackend/gmysqlbackend.cc
+++ b/modules/gmysqlbackend/gmysqlbackend.cc
@@ -120,7 +120,7 @@ public:
     declare(suffix,"update-account-query","", "update domains set account=? where name=?");
     declare(suffix,"update-serial-query","", "update domains set notified_serial=? where id=?");
     declare(suffix,"update-lastcheck-query","", "update domains set last_check=? where id=?");
-    declare(suffix,"info-all-master-query","", "select id,name,master,last_check,notified_serial,type from domains where type='MASTER'");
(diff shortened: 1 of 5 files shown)


zeha:
The schema files in git appear to have different ideas if there should be an index on `records.domain_id` or not (`schema.mysql.sql` has it, the upgrade schema files appear to drop it...). 
So it's somewhat hard to tell what users are actually running.

Intuition-wise I'd favour the index, but my feelings are often wrong for MySQL.

Habbie:
> the upgrade schema files appear to drop it..

Hmm, can you point me to that?

zeha:
> > the upgrade schema files appear to drop it..
> 
> Hmm, can you point me to that?

I can't seem to find what I was referencing back then. Ignore my comment? :)

Habbie:
Did you test the performance of the other backends as well?

mind04:
No, but i expect similar results for the other sql backends.

Habbie:
sqlite3, 500000 domains:
```
May 27 15:27:37 getUpdatedMasters: 10452069 usec to execute
```
->
```
May 27 15:37:12 getUpdatedMasters: 1935608 usec to execute
```

Issue #7452, opened 2019-02-05
recursor Unable to load RPZ zone 'rpz.mypdns.cloud' from '127.0.0.1:53'


spirillen:
- Program: Recursor, (maybe also some Authoritative)
 - Issue type: Bug report

### Short description

journal -xefu pdns-recursor | grep rpz
```
Loading RPZ zone 'whitelist.rpz.mypdns.cloud' from [2a01:4f8:1c1c:abe4::53]
Feb 05 17:01:51 pdns_recursor[3457]: Loaded & indexed 0 policy records so far for RPZ zone 'whitelist.rpz.mypdns.cloud'
Feb 05 17:01:51 pdns_recursor[3457]: Unable to load RPZ zone 'whitelist.rpz.mypdns.cloud' from '[2a01:4f8:1c1c:abe4::53]': 'Adding a QName-based filter policy of kind rpz-passthru. but a policy of kind rpz-passthru. already exists for the following QName: .'. (Will try again in 60 seconds...)
Feb 05 17:01:51 pdns_recursor[3457]: Loading RPZ zone 'rpz.mypdns.cloud' from [2a01:4f8:1c1c:abe4::53]
Feb 05 17:01:51 pdns_recursor[3457]: Loaded & indexed 0 policy records so far for RPZ zone 'rpz.mypdns.cloud'
Feb 05 17:01:51 pdns_recursor[3457]: Unable to load RPZ zone 'rpz.mypdns.cloud' from '[2a01:4f8:1c1c:abe4::53]': 'Adding a QName-based filter policy of kind . but a policy of kind . already exists for the following QName: .'. (Will try again in 60 seconds...)
```

### Environment

 - Operating system: Ubuntu 18.04
 - Software version: 0.0.15970.0.g70093b3d00
 - Software source: PowerDNS repository

### Steps to reproduce

1. add in recursor.lua
```
rpzMaster({"[2a01:4f9:c010:2166::53]:53","[2a01:4f8:1c1c:abe4::53]:53"}, "whitelist.rpz.mypdns.cloud", {defpol=Policy.NoAction, refresh="60"})
rpzMaster({"[2a01:4f9:c010:2166::53]:53","[2a01:4f8:1c1c:abe4::53]:53"}, "rpz.mypdns.cloud", {defpol=Policy.NXDOMAIN, refresh="60", defttl="604800", axfrTimeout="600"})

```
2. restart pdns-recursor
3. sedund test also fails, in recursor.lua add:
```
[...]

PR #7476, opened 2019-02-11
rec: Add an option to not override custom RPZ types with the default policy

rgacogne:
### Short description

Also be a bit more permissive when we do override them, because it might turn several `A` or `AAAA` records into several `CNAME`s which would trigger an error.
Fixes https://github.com/PowerDNS/pdns/issues/7452.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/pdns/recursordist/docs/lua-config/rpz.rst
+++ b/pdns/recursordist/docs/lua-config/rpz.rst
@@ -41,7 +41,7 @@ To instead drop all queries matching a rule, while slaving from a master:
 
     rpzMaster("192.0.2.4", "policy.rpz", {defpol=Policy.Drop})
 
-Note that since 4.2.0, it is possible for the override policy specified via 'defpol' to no longer applied to local data entries present in the zone by setting the 'defpolOverrideLocalData' parameter to false.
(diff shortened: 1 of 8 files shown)


pdns/recursordist/docs/lua-config/rpz.rst
```diff
@@ -18,16 +18,31 @@ An RPZ can be loaded from file or slaved from a master. To load from file, use f
...
+
+    rpzFile("dblfilename", {defpol=Policy.Custom, defcontent="badserver.example.com"})
+
+To instead drop all queries matching a rule, while slaving from a master:
+
+.. code-block:: Lua
+
+    rpzMaster("192.0.2.4", "policy.rpz", {defpol=Policy.Drop})
+
+Note that since 4.2.0, it is possible for the override policy specified via 'defpol' to no longer applied to local data entries present in the zone by setting the 'defpolOverrideLocalData' parameter to false.
```
chbruyand:
```suggestion
Note that since 4.2.0, it is possible for the override policy specified via 'defpol' to no longer be applied to local data entries present in the zone by setting the 'defpolOverrideLocalData' parameter to false.
```

PR #7478, opened 2019-02-11
rec: Add a new 'max-cache-bogus-ttl' option

### LINKED ISSUE #7445: rec: Add an option to cap the TTL of a Bogus answer
- Program: Recursor
 - Issue type: Feature request

### Short description

Since 4.1.0, the DNSSEC validation state of records is stored along with them so that we only need to validate once. It would be nice to cap the TTL of a record that has been validated as `Bogus` in the query cache, so we don't keep it around for days if the initial TTL is high. The initial value of this new setting (`max-cache-bogus-tll`?) should not be too low either so we are not too penalized by a broken domain. `3600` might be a good choice.
For the packet cache, I think the issue is already covered by `packetcache-servfail-ttl` [1].

[1]: https://docs.powerdns.com/recursor/settings.html#packetcache-servfail-ttl

rgacogne:
### Short description

Since 4.1.0, the DNSSEC validation state of records is stored along with them so that we only need to validate once. This PR adds a new 'max-cache-bogus-ttl' option to cap the TTL of a record that has been validated as `Bogus` in the query cache, so we don't keep it around for days if the initial TTL is high.
Closes #7445.

This PR still needs a bit of testing, as well as regression tests.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/recursordist/test-syncres_cc.cc
+++ b/pdns/recursordist/test-syncres_cc.cc
@@ -9382,7 +9382,7 @@ BOOST_AUTO_TEST_CASE(test_dnssec_rrsig_negcache_bogus_validity) {
   BOOST_REQUIRE_EQUAL(ret.size(), 3);
   BOOST_CHECK_EQUAL(queriesCount, 4);
 
-  /* check that the entry has not been negatively cached but not longer than s_maxbogusttl */
+  /* check that the entry has been negatively cached but not longer than s_maxbogusttl */
   const NegCache::NegCacheEntry* ne = nullptr;
   BOOST_CHECK_EQUAL(SyncRes::t_sstorage.negcache.size(), 1);
   BOOST_REQUIRE_EQUAL(SyncRes::t_sstorage.negcache.get(target, QType(QType::A), sr->getNow(), &ne), true);
--- a/pdns/recursordist/test-syncres_cc.cc
+++ b/pdns/recursordist/test-syncres_cc.cc
@@ -121,6 +121,7 @@ static void init(bool debug=false)
   SyncRes::s_maxtotusec = 1000*7000;
   SyncRes::s_maxdepth = 40;
   SyncRes::s_maxnegttl = 3600;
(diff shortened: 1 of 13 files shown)


rgacogne:
I just added a very simple regression test, and did various tests by hand.

pdns/syncres.cc
```diff
@@ -2366,6 +2397,13 @@ RCode::rcodes_ SyncRes::updateCacheFromRecords(unsigned int depth, LWResult& lwr
       }
     }
 
+    if (recordState == Bogus) {
+      /* this is a TTD by now, be careful */
+      for(auto& record : i->second.records) {
+        record.d_ttl = std::min(record.d_ttl, static_cast<uint32_t>(s_maxbogusttl + d_now.tv_sec));
```
chbruyand: isn't `record.d_ttl` TTL and not a TTD ?

rgacogne: Unless I'm mistaken it's a TTD at this point, since `dr.d_ttl += d_now.tv_sec;` line 2283.

pdns/recursordist/test-syncres_cc.cc
```diff
@@ -9321,6 +9332,76 @@ BOOST_AUTO_TEST_CASE(test_dnssec_rrsig_negcache_validity) {
...
+  SyncRes::s_maxbogusttl = 360;
+
+  vector<DNSRecord> ret;
+  int res = sr->beginResolve(target, QType(QType::A), QClass::IN, ret);
+  BOOST_CHECK_EQUAL(res, RCode::NoError);
+  BOOST_CHECK_EQUAL(sr->getValidationState(), Bogus);
+  BOOST_REQUIRE_EQUAL(ret.size(), 3);
+  BOOST_CHECK_EQUAL(queriesCount, 4);
+
+  /* check that the entry has not been negatively cached but not longer than s_maxbogusttl */
```
pieterlexis: `s/has not been/has been/`?

PR #7480, opened 2019-02-12
rec: implement a way to disallow throtteling of auths

pieterlexis:
### Short description

TODO:

- [ ] Add new tests for this functionality
- [x] Allow updating the lists via `rec_control`

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


rgacogne:
I did not review the code but some SyncRes unit tests related to the throttle code are failing.

rgacogne:
This PR makes it possible to disable throttling for all forward-recurse servers, but it's all or nothing, and limited to these kind of servers. How about having a map instead, so you could disable throttling for selected servers, based on the IP address (but suffix matching would be nice too), which would make it usable for all servers, even regular auths?

pdns/pdns_recursor.cc
```diff
@@ -4219,6 +4240,9 @@ int main(int argc, char **argv)
     ::arg().set("max-tcp-clients","Maximum number of simultaneous TCP clients")="128";
     ::arg().set("server-down-max-fails","Maximum number of consecutive timeouts (and unreachables) to mark a server as down ( 0 => disabled )")="64";
     ::arg().set("server-down-throttle-time","Number of seconds to throttle all queries to a server after being marked as down")="60";
+    ::arg().set("throttle-forward-recurse", "Throttle servers that are forward to with the RD bit set")="yes";
```
pieterlexis: needs to be removed

pdns/syncres.cc
```diff
@@ -2618,7 +2618,7 @@ bool SyncRes::processRecords(const std::string& prefix, const DNSName& qname, co
   return done;
 }
 
-bool SyncRes::doResolveAtThisIP(const std::string& prefix, const DNSName& qname, const QType& qtype, LWResult& lwr, boost::optional<Netmask>& ednsmask, const DNSName& auth, bool const sendRDQuery, const DNSName& nsName, const ComboAddress& remoteIP, bool doTCP, bool* truncated)
+bool SyncRes::doResolveAtThisIP(const std::string& prefix, const DNSName& qname, const QType& qtype, LWResult& lwr, boost::optional<Netmask>& ednsmask, const DNSName& auth, bool const sendRDQuery, const DNSName& nsName, const ComboAddress& remoteIP, bool doTCP, bool* truncated, const bool wasForwarded)
```
rgacogne: I don't think we need `wasForwarded` anymore.

rgacogne:
I don't really understand why but the interface shows a lot of commits that are not related to this PR?!

pieterlexis:
> I don't really understand why but the interface shows a lot of commits that are not related to this PR?!

rebased on master, looks ok now

pdns/dnsname.hh
```diff
@@ -340,14 +385,33 @@ struct SuffixMatchNode
...
+    std::string ret;
+    bool first = true;
+    for (const auto &n: d_tree.getNodes()) {
+      if (!first) {
+        ret += ", ";
+      }
+      first = false;
+      ret += n.toString();
+    }
+    return ret;
```
rgacogne: I'm a bit worried about this operation going from `O(1)` to `O(n)`, especially in dnsdist where someone could trigger this by retrieving the list of rules from the `API` for example. Perhaps we could cache the result of this operation to a mutable member variable and invalidate it whenever `::add()` or `::remove()` is called?

pdns/dnsname.hh
```diff
@@ -286,6 +286,42 @@ struct SuffixMatchTree
...
+      // No subnode found, we're done
+      return;
+    }
+
+    // We have found a child
+    labels.pop_back();
+    if (labels.empty()) {
+      // The child is no longer an endnode
+      child->endNode = false;
+      // TODO clear d_value for this node as d_value = T() would break for types
```
rgacogne: I'm not sure that's up to this class to handle that. If we really need to, perhaps we could do something with an optional `boost::optional<T> resetValue=boost::none` parameter and set `d_value = resetValue` when clearing, but that seems brittle.

pdns/test-dnsname_cc.cc
```diff
@@ -558,6 +563,11 @@ BOOST_AUTO_TEST_CASE(test_suffixmatch_tree) {
   BOOST_CHECK_EQUAL(*smt.lookup(examplenet), examplenet);
   BOOST_REQUIRE(smt.lookup(net));
   BOOST_CHECK_EQUAL(*smt.lookup(net), net);
+
+  // Remove .net, and check that example.net remains
+  smt.remove(net);
+  BOOST_CHECK(smt.lookup(net) == nullptr);
+  BOOST_CHECK_EQUAL(*smt.lookup(examplenet), examplenet);
 }
```
rgacogne: I have a few more tests in a different PR, I'll PR them against this branch.

pdns/dnsname.hh
pieterlexis: we could store the dnsnames in a set and refresh `d_human` on `::add()` and `::remove()`

pieterlexis: Although this means that `::add()` and `::remove()` will be a bit more complex.. I wonder what would be called more often, `toString()` or `::add()`/`::remove()`

pdns/dnsname.hh
```diff
@@ -371,49 +371,75 @@ struct SuffixMatchTree
...
+  public:
+    SuffixMatchNode()
+    {}
+    SuffixMatchTree<bool> d_tree;
+
+    void add(const DNSName& dnsname)
+    {
+      d_tree.add(dnsname, true);
+      d_nodes.insert(dnsname);
+      updateHuman();
```
rgacogne: I'm a bit worried that we are now doing an `O(n)` operation on insertion, although that seems fine with the way we currently use `SuffixMatchNode`, except perhaps for the `SuffixMatchNodeRule` in dnsdist where it depends how it's used from `Lua`.

pdns/dnsname.hh
pieterlexis: :+1:, I'll remove the comment and keep it like this.

rgacogne:
Additional tests PR'ed at https://github.com/pieterlexis/pdns/pull/13

pdns/rec_channel_rec.cc
```diff
@@ -1353,6 +1353,190 @@ static string* nopFunction()
   return new string("pong\n");
 }
 
+string getDontThrottleNames() {
```
rgacogne: Just a nit but should be `static`.

pdns/rec_channel_rec.cc
```diff
@@ -1353,6 +1353,190 @@ static string* nopFunction()
...
+  auto dtn = g_dontThrottleNames.getLocal();
+  return dtn->toString() + "\n";
+}
+
+string getDontThrottleNetmasks() {
```
rgacogne: same

pdns/rec_channel_rec.cc
```diff
@@ -1353,6 +1353,190 @@ static string* nopFunction()
...
+  auto dtn = g_dontThrottleNetmasks.getLocal();
+  return dtn->toString() + "\n";
+}
+
+template<typename T>
+string addDontThrottleNames(T begin, T end) {
```
rgacogne: as well

pdns/rec_channel_rec.cc
```diff
@@ -1353,6 +1353,190 @@ static string* nopFunction()
...
+
+  g_dontThrottleNames.setState(dnt);
+
+  ret += " to the list of nameservers that may not be throttled";
+  g_log<<Logger::Info<<ret<<", requested via control channel"<<endl;
+  return ret + "\n";
+}
+
+template<typename T>
+string addDontThrottleNetmasks(T begin, T end) {
```
rgacogne: this one too

pdns/rec_channel_rec.cc
```diff
@@ -1353,6 +1353,190 @@ static string* nopFunction()
...
+
+  g_dontThrottleNetmasks.setState(dnt);
+
+  ret += " to the list of nameserver netmasks that may not be throttled";
+  g_log<<Logger::Info<<ret<<", requested via control channel"<<endl;
+  return ret + "\n";
+}
+
+template<typename T>
+string clearDontThrottleNames(T begin, T end) {
```
rgacogne: almost there ;)

pdns/rec_channel_rec.cc
```diff
@@ -1353,6 +1353,190 @@ static string* nopFunction()
...
+
+  g_dontThrottleNames.setState(dnt);
+
+  ret += " from the list of nameservers that may not be throttled";
+  g_log<<Logger::Info<<ret<<", requested via control channel"<<endl;
+  return ret + "\n";
+}
+
+template<typename T>
+string clearDontThrottleNetmasks(T begin, T end) {
```
rgacogne: Last one!

pdns/syncres.cc
```diff
@@ -2780,7 +2780,12 @@ bool SyncRes::doResolveAtThisIP(const std::string& prefix, const DNSName& qname,
...
     }
 
-    if(resolveret != -2 && !chained) { // don't account for resource limits, they are our own fault
+    auto dontThrottleNames = g_dontThrottleNames.getLocal();
+    auto dontThrottleNetmasks = g_dontThrottleNetmasks.getLocal();
+
+    if(resolveret != -2 && !chained && !(dontThrottleNames->check(nsName) || dontThrottleNetmasks->match(remoteIP))) {
+      // don't account for resource limits, they are our own fault
+      // And don't throttle when the IP address is on the dontThrottleNetmasks list or the name is part of dontThrottleNames
       t_sstorage.nsSpeeds[nsName.empty()? DNSName(remoteIP.toStringWithPort()) : nsName].submit(remoteIP, 1000000, &d_now); // 1 sec
```
rgacogne: It looks like the server can still be throttle on ServFail, Refused, TC=1 over TCP or a lame answer. Is that expected? If so perhaps we should document that.

pieterlexis: it is expected, this only disable throttling on timeouts

pieterlexis: otoh, it might defeat the purpose, I'll make it never throttle

pdns/recursordist/docs/settings.rst
```diff
@@ -257,6 +257,40 @@ Operate in the background.
...
+.. _setting-dont-throttle-names:
+
+``dont-throttle-names``
+----------------------------
+.. versionadded:: 4.2.0
+
+-  Comma separated list of domain-names
+-  Default: (empty)
+
+When an authotitative server does not answer a query or sends a reply the recursor does lot like, it is throttled.
```
```suggestion
When an authoritative server does not answer a query or sends a reply the recursor does not like, it is throttled.
```

pdns/recursordist/docs/settings.rst
```diff
@@ -257,6 +257,40 @@ Operate in the background.
...
+.. _setting-dont-throttle-netmasks:
+
+``dont-throttle-netmasks``
+----------------------------
+.. versionadded:: 4.2.0
+
+-  Comma separated list of netmasks
+-  Default: (empty)
+
+When an authotitative server does not answer a query or sends a reply the recursor does lot like, it is throttled.
```
```suggestion
When an authoritative server does not answer a query or sends a reply the recursor does not like, it is throttled.
```

pdns/recursordist/docs/settings.rst
```diff
@@ -257,6 +257,40 @@ Operate in the background.
...
+.. versionadded:: 4.2.0
+
+-  Comma separated list of domain-names
+-  Default: (empty)
+
+When an authotitative server does not answer a query or sends a reply the recursor does lot like, it is throttled.
+Any servers' name suffix-matching the supplied names will never be throttled.
+
+.. warning::
+  Most servers on the internet do not respond for a good reason (overloaded or unreachable), ``dont-throttle-names`` could make this load on the upstream server even higher, resulting in further service degredation.
```
```suggestion
  Most servers on the internet do not respond for a good reason (overloaded or unreachable), ``dont-throttle-names`` could make this load on the upstream server even higher, resulting in further service degradation.
```

pdns/recursordist/docs/settings.rst
```diff
@@ -257,6 +257,40 @@ Operate in the background.
...
+
+When an authotitative server does not answer a query or sends a reply the recursor does lot like, it is throttled.
+Any servers matching the supplied netmasks will never be throttled.
+
+This can come in handy on lossy networks when forwarding, where the same server is configured multiple times (e.g. with ``forward-zones-recurse=example.com=192.0.2.1;192.0.2.1``).
+By default, the PowerDNS Recursor would throttle the "first" server on a timeout and hence not retry the "second" one.
+In this case, ``dont-throttle-netmasks`` could be set to ``192.0.2.1``.
+
+.. warning::
+  Most servers on the internet do not respond for a good reason (overloaded or unreachable), ``dont-throttle-netmasks`` could make this load on the upstream server even higher, resulting in further service degredation.
```
```suggestion
  Most servers on the internet do not respond for a good reason (overloaded or unreachable), ``dont-throttle-netmasks`` could make this load on the upstream server even higher, resulting in further service degradation.
```

PR #7484, opened 2019-02-14
Utility::random() and srandom() is not used anymore.

omoerbeek:
### Short description

Eliminate rand() and srand() usage. These are really bad random generators and should not be used. Leaving usage of random() alone.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/common_startup.cc
+++ b/pdns/common_startup.cc
@@ -496,6 +496,8 @@ static void triggerLoadOfLibraries()
 
 void mainthread()
 {
+   Utility::srandom(time(0) ^ getpid());
+
    int newgid=0;
    if(!::arg()["setgid"].empty()) 
      newgid=Utility::makeGidNumeric(::arg()["setgid"]);      
--- a/pdns/common_startup.cc
+++ b/pdns/common_startup.cc
@@ -496,7 +496,7 @@ static void triggerLoadOfLibraries()
 
 void mainthread()
 {
-   Utility::srandom(time(0) ^ getpid());
+   Utility::srandom();
 
    int newgid=0;
    if(!::arg()["setgid"].empty()) 
--- a/pdns/common_startup.cc
+++ b/pdns/common_startup.cc
@@ -496,9 +496,7 @@ static void triggerLoadOfLibraries()
 
 void mainthread()
 {
-   Utility::srandom(time(0) ^ getpid());
-
-   int newgid=0;      
+   int newgid=0;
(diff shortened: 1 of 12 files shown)


pdns/common_startup.cc
```diff
@@ -496,9 +496,7 @@ static void triggerLoadOfLibraries()
 
 void mainthread()
 {
-   Utility::srandom(time(0) ^ getpid());
```
rgacogne: It looks like `random()` is still used in `waitForMultiData()`, `waitFor2Data()` (this one seems unused, though) and `Lua` records, so perhaps we still need to seed it properly?

omoerbeek: Yes, looking into this problem as we speak. e.g. pdns_notify uses random() for it's DNS id and always sends the same one.

omoerbeek:
The calls of random() in misc are a bit hard to avoid by replacing them with dns_random(), since the would introduce dependencies on libcrypto and possibly libsodium in a lot of utils. Don't know if that is worth the trouble.
AFAIKS, the non-seeded use of random() are not a big issue in misc.cc.

pieterlexis:
> The calls of random() in misc are a bit hard to avoid by replacing them with dns_random(), since the would introduce dependencies on libcrypto and possibly libsodium in a lot of utils. Don't know if that is worth the trouble.

libcrypto is already a dependency for the auth and the recursor, I don't think it is a problem to have several of our tools depend on it as well.

omoerbeek:
If dependencies on libcrypto (and libsodium) are not considered a problem, I might prepare a bigger hammer to move everything to dns_random and eliminate random() everywhere. A kind of preventive measure to avoid having a case where predictable random is actually problem while we thought it was not.

rgacogne:
libcrypto is already required in the auth and rec, but not in dnsdist, so please be careful about that.

omoerbeek:
Looks like dnsdist already has a (conditional) dependency on libcrypto if HAVE_DNS_OVER_TLS && HAVE_LIBSSL

rgacogne:
Indeed, but I really would like that dependency to remain optional.

omoerbeek:
I'm parking this for the moment. It introduces a strange error in the rec testrunner on travis I cannot reproduce on any system available to me. I need to reproduce the travis environment exactly, it seems.

omoerbeek:
Failure on travis tracked and fixed. Should be good to merge.

PR #7494, opened 2019-02-18
Correctly interpret an empty AXFR response to an IXFR query

rgacogne:
### Short description

If the remote server answered with an empty AXFR response to a RPZ IXFR query, the recursor incorrectly parsed this answer as an empty IXFR, instead of removing all records except the SOA from the zone.
It looks like the authoritative server is affected as well when slaving a, IXFR-enabled zone.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code


rgacogne:
Rebased to fix a conflict following the merge of #7476.

Issue #7373, opened 2019-01-13
Authoritative ringbuffers are ASCII & slow


ahupowerdns:
In the Authoritative Server, we store queries and responses in several ringbuffers. These ringbuffers are templatized and could contain anything we want. As it stands, they contain strings. This means that every query goes through several DNSName::toLogString() conversions. 

Under high load, this represents around 5%-10% of what we are doing. Ringbuffers  with DNSNames are updated from several places:
```
common_startup.cc:    S.ringAccount("queries", P->qdomain.toLogString()+"/"+P->qtype.getName());
packethandler.cc:  S.ringAccount("noerror-queries",p->qdomain.toLogString()+"/"+p->qtype.getName());
packethandler.cc:    S.ringAccount("servfail-queries",p->qdomain.toLogString());
packethandler.cc:    S.ringAccount("servfail-queries",p->qdomain.toLogString());
responsestats-auth.cc:      S.ringAccount("nxdomain-queries",p.qdomain.toLogString()+"/"+p.qtype.getName());
responsestats-auth.cc:    S.ringAccount("unauth-queries",p.qdomain.toLogString()+"/"+p.qtype.getName());
```
To fix, templataize these ringbuffers to they contain pair<DNSName,QType> & adjust API and webserver so *they* do the conversion if needed.

zeha:
Quoting @Habbie on a very insightful comment:
```
<Habbie> so writing to them is slow, but that means this does not suddenly hurt operations when you start looking at them
```

rgacogne:
Aren't we much more likely to write to them than to read from them, though?

zeha:
Summary of some FD:

- only the top10 are printed
- obv. everything is always written

Result: deferring the string work to printing will likely have very little operational impact at print time, and would save a ton of work when writing.

Related discoveries: the current code always does a by-string lookup to find the same three ringbuffers. That should also go away.

rgacogne:
Mostly fixed by #7503, except perhaps that last discovery:

> the current code always does a by-string lookup to find the same three ringbuffers. That should also go away.

PR #7503, opened 2019-02-20
Auth ringbuffer simmaries were case sensitive & accounted delegations incorrectly

ahupowerdns:
### Short description
When we previously summarised a ringbuffer we would do so case sensitively. In addition, packetcached delegations were put in the rings for queries for domains we are not auth for.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code


pieterlexis:
This should fix the issue mentioned in #7373

ToDo:

- [ ] Templatize

Issue #7498, opened 2019-02-18
recursor needs more ecs metrics to expose subnet response sizes


ahupowerdns:
### Short description
The PowerDNS recursor support EDNS Client Subnet for outgoing queries. Sometimes ECS auths send back answers for large subnets (/16, for example), but sometimes this also goes down to /25 for highly fragmented networks.

Short of doing a cache dump or a packet capture we currently do not expose what kind of masks we get back from authoritatives.

Although finer grained data would be nice, counters that expose 'ecs-v4-response-bits-32', through to 'ecs-v4-response-bits-1' are a good start. And of course the 128 counters for IPv6....

### Usecase
The usecase is to help debug sudden changes in ECS traffic that happen for no reason.

ahupowerdns:
we might make this an API-only metric btw, I'm not looking forward to sending everyone 160 new metrics every 30 seconds that will mostly be 0.. We already have the concept of 'metrics you have to ask for specifically'. We may have to make this a carbon setting perhaps. Unsure.

rgacogne:
Implemented in #7504.

PR #7504, opened 2019-02-20
Add rec statistics about ECS response sizes, API endpoint to get a specific stat

rgacogne:
### Short description

This PR:
- adds statistics about ECS response sizes in the recursor, closing #7498 ;
- adds an API endpoint to request a specific statistic only, in the recursor and the auth, even if it is otherwise excluded from the list of all statistics ;
- disable the polling of expensive statistics (cache-bytes, packetcache-bytes) via SNMP by default ;
- make the list of metrics that are excluded from either the list of all statistics (API, rec-control) or completely (Carbon, SNMP) configurable per retrieval method.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/pdns/recursordist/docs/metrics.rst
+++ b/pdns/recursordist/docs/metrics.rst
@@ -236,6 +236,14 @@ ecs-responses
 ^^^^^^^^^^^^^
 number of responses received from authoritative servers with an EDNS Client Subnet option we used (since 4.1)
 
+ecs-v4-response-bits-*
+^^^^^^^^^^^^^^^^^^^^^^
+number of responses received from authoritative servers with an IPv4 EDNS Client Subnet option we used, of this subnet size (1 to 32). Added in 4.2.0.
+
+ecs-v6-response-bits-*
+^^^^^^^^^^^^^^^^^^^^^^
+number of responses received from authoritative servers with an IPv6 EDNS Client Subnet option we used, of this subnet size (1 to 128). Added in 4.2.0.
+
 edns-ping-matches
 ^^^^^^^^^^^^^^^^^
 number of servers that sent a valid EDNS PING   response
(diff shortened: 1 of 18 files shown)


pdns/ws-api.cc
```diff
@@ -159,15 +159,29 @@ void apiServerStatistics(HttpRequest* req, HttpResponse* resp) {
   if(req->method != "GET")
     throw HttpMethodNotAllowedException();
 
+  Json::array doc;
+  string name = req->getvars["statistic"];
+  if (!name.empty()) {
+    auto stat = productServerStatisticsFetch(name);
```
pieterlexis: would a `serverStatisticsFetch` function that uses `#ifdef RECURSOR` with an `#else` be more readable than this indirection?

pdns/recursordist/docs/metrics.rst
```diff
@@ -236,6 +236,14 @@ ecs-responses
 ^^^^^^^^^^^^^
 number of responses received from authoritative servers with an EDNS Client Subnet option we used (since 4.1)
 
+ecs-v4-response-bits-*
+^^^^^^^^^^^^^^^^^^^^^^
+number of responses received from authoritative servers with an IPv4 EDNS Client Subnet option we used, of this subnet size (1 to 32). Added in 4.2.0.
```
pieterlexis:
```rst
.. versionadded:: 4.2.0
```

rgacogne:
> I would prefer the naming of the settings to `stats-$TYPE-blacklist` so in the future new `stats-$TYPE-*` settings are together when sorted alphabetically.

Makes sense to me! Done!

pdns/recursordist/docs/metrics.rst
rgacogne: Done!

pdns/ws-api.cc
rgacogne: I actually find the current way cleaner than having conditional includes and code blocks.

PR #7507, opened 2019-02-20
rec: Use a bounded load-balancing algo to distribute queries

rgacogne:
### Short description

If `pdns-distributes-queries` is set and this setting is set to another value than 0, the distributor thread will use a bounded load-balancing algorithm while distributing queries to worker threads, making sure that no thread is assigned more queries than distribution-load-factor times the average number of queries currently processed by all the workers.
For example, with a value of 1.25, no server should get more than 125 % of the average load. This helps making sure that all the workers have roughly the same share of queries, even if the incoming traffic is very skewed, with a majority of requests asking for the same qname.

### Checklist

I have:
- [x] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/pdns_recursor.cc
+++ b/pdns/pdns_recursor.cc
@@ -2888,7 +2888,7 @@ static unsigned int selectWorker(unsigned int hash)
   // cerr<<"total load is "<<currentLoad<<", number of workers is "<<g_numWorkerThreads<<", target load is "<<targetLoad<<endl;
 
   unsigned int worker = hash % g_numWorkerThreads;
-  /* at least one server has to be below the average load */
+  /* at least one server has to be at or below the average load */
   if (load[worker] > targetLoad) {
     ++g_stats.rebalancedQueries;
     do {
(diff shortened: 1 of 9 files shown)


ahupowerdns:
HOT

pdns/pdns_recursor.cc
```diff
@@ -2843,9 +2855,52 @@ static bool trySendingQueryToWorker(unsigned int target, ThreadMSG* tmsg)
...
+    load[idx] = getWorkerLoad(idx);
+    currentLoad += load[idx];
+    // cerr<<"load for worker "<<idx<<" is "<<load[idx]<<endl;
+  }
+
+  double targetLoad = (currentLoad / g_numWorkerThreads) * s_balancingFactor;
+  // cerr<<"total load is "<<currentLoad<<", number of workers is "<<g_numWorkerThreads<<", target load is "<<targetLoad<<endl;
+
+  unsigned int worker = hash % g_numWorkerThreads;
+  /* at least one server has to be below the average load */
```
chbruyand: That may not be the case if `s_balancingFactor` is set to a low value (`< 1/(g_numWorkerThreads + 1)` or something like that). Maybe we should ensure `distribution-load-factor` is greater than `1.0` ?

rgacogne: Nice catch!

PR #7508, opened 2019-02-21
rec: Add 'protobuf-use-kernel-timestamp' for sharper latencies

rgacogne:
### Short description

Using `SO_TIMESTAMP` when available, so we can really compute the latency from the moment the datagram was actually received by the kernel instead of the moment we started processing it.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/pdns_recursor.cc
+++ b/pdns/pdns_recursor.cc
@@ -208,6 +208,7 @@ static bool g_reusePort{false};
 static bool g_gettagNeedsEDNSOptions{false};
 static time_t g_statisticsInterval;
 static bool g_useIncomingECS;
+static bool g_useKernelTimestamp;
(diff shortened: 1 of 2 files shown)


ahupowerdns:
Is there a reason to make this optional? Would you ever not want this perhaps?

rgacogne:
> Is there a reason to make this optional? Would you ever not want this perhaps?

My main worry was that the reported latency would increase artificially between 4.1.x and 4.2.0, leading to misguided users complaining about some regression. Especially if we later decide not to limit that timestamp to protobuf but to also use it for our other latency measurements.

rgacogne:
Rebased to fix a conflict.

PR #7509, opened 2019-02-21
plug mysql_thread_init memory leak, fixes #6231

### LINKED ISSUE #6231: Authoritative: Memory leak with bulk API request
- Program: Authoritative
 - Issue type: Bug report

### Short description

PDNS authoritative server consumes much memory with API access.

### Environment

 - Operating system: Debian jessie
 - Software version: 4.1.0

```
$  dpkg -l | grep pdns
ii  pdns-backend-mysql                   4.1.0-1pdns.jessie           amd64        generic MySQL backend for PowerDNS
ii  pdns-server                          4.1.0-1pdns.jessie           amd64        extremely powerful and versatile nameserver
ii  pdns-server-dbg                      4.1.0-1pdns.jessie           amd64        debugging symbols for PowerDNS
```

### Steps to reproduce

1. Set up pdns server with mysql backend like https://gist.github.com/nonylene/c7d85d68a1b71411397eafd3e995c03e 
2. Create `local.` domain by `crate_local_domain.py` at https://gist.github.com/nonylene/eef81b1a3b36974fdccce118e31ca196 
3. Create `example.local.` record by `create_example_record.py` at  https://gist.github.com/nonylene/eef81b1a3b36974fdccce118e31ca196
4. Send bulk GET API request to pdns

```sh
$ while true; do curl -H 'X-API-Key: testkey' http://127.0.0.1:8081/api/v1/servers/localhost/zones 2&>1 > /dev/null ; done
```

### Expected behaviour

pdns does not consume all memory.

### Actual behaviour

screen shot 2018-01-26 at 17 51 52

Memory usage increased gradually. Finally, pdns used all memory (including swap) and instance was  killed.

### Other information

Here is valgrind summary.

```
$ sudo valgrind --leak-check=full --log-file=pdns-valgrind.out -v --trace-children=yes --track-origins=yes -- /usr/sbin/pdns_server --guardian=no --daemon=no --disable-syslog --write-pid=no
```

```
...
==3186== 49,152 bytes in 3 blocks are possibly lost in loss record 2,746 of 2,767
==3186==    at 0x4C28C20: malloc (vg_replace_malloc.c:296)
==3186==    by 0x81E0C54: my_malloc (in /usr/lib/x86_64-linux-gnu/libmysqlclient.so.18.1.0)
==3186==    by 0x81D81BE: mysql_socket_vio_new (in /usr/lib/x86_64-linux-gnu/libmysqlclient.so.18.1.0)
==3186==    by 0x81D82C0: vio_new (in /usr/lib/x86_64-linux-gnu/libmysqlclient.so.18.1.0)
==3186==    by 0x81B4725: mysql_real_connect (in /usr/lib/x86_64-linux-gnu/libmysqlclient.so.18.1.0)
==3186==    by 0x7F65205: SMySQL::connect() (in /usr/lib/x86_64-linux-gnu/pdns/libgmysqlbackend.so)
==3186==    by 0x7F65680: SMySQL::SMySQL(std::string const&, std::string const&, unsigned short, std::string const&, std::string const&, std::string const&, std::string const&, bool, unsigned int) (in /usr/lib/x86_64-linux-gnu/pdns/libgmysqlbackend.so)
==3186==    by 0x7F576D9: gMySQLBackend::reconnect() (in /usr/lib/x86_64-linux-gnu/pdns/libgmysqlbackend.so)
==3186==    by 0x7F57C04: gMySQLBackend::gMySQLBackend(std::string const&, std::string const&) (in /usr/lib/x86_64-linux-gnu/pdns/libgmysqlbackend.so)
==3186==    by 0x7F64185: gMySQLFactory::make(std::string const&) (in /usr/lib/x86_64-linux-gnu/pdns/libgmysqlbackend.so)
==3186==    by 0x2265B0: BackendMakerClass::all(bool) (in /usr/sbin/pdns_server)
==3186==    by 0x3B2F88: UeberBackend::UeberBackend(std::string const&) (in /usr/sbin/pdns_server)
==3186==
==3186== 3,924,600 bytes in 19,623 blocks are definitely lost in loss record 2,767 of 2,767
==3186==    at 0x4C2AD10: calloc (vg_replace_malloc.c:623)
==3186==    by 0x81E2436: my_thread_init (in /usr/lib/x86_64-linux-gnu/libmysqlclient.so.18.1.0)
==3186==    by 0x81AB324: mysql_server_init (in /usr/lib/x86_64-linux-gnu/libmysqlclient.so.18.1.0)
==3186==    by 0x81B1A36: mysql_init (in /usr/lib/x86_64-linux-gnu/libmysqlclient.so.18.1.0)
==3186==    by 0x7F65103: SMySQL::connect() (in /usr/lib/x86_64-linux-gnu/pdns/libgmysqlbackend.so)
==3186==    by 0x7F65680: SMySQL::SMySQL(std::string const&, std::string const&, unsigned short, std::string const&, std::string const&, std::string const&, std::string const&, bool, unsigned int) (in /usr/lib/x86_64-linux-gnu/pdns/libgmysqlbackend.so)
==3186==    by 0x7F576D9: gMySQLBackend::reconnect() (in /usr/lib/x86_64-linux-gnu/pdns/libgmysqlbackend.so)
==3186==    by 0x7F57C04: gMySQLBackend::gMySQLBackend(std::string const&, std::string const&) (in /usr/lib/x86_64-linux-gnu/pdns/libgmysqlbackend.so)
==3186==    by 0x7F64185: gMySQLFactory::make(std::string const&) (in /usr/lib/x86_64-linux-gnu/pdns/libgmysqlbackend.so)
==3186==    by 0x2265B0: BackendMakerClass::all(bool) (in /usr/sbin/pdns_server)
==3186==    by 0x3B2F88: UeberBackend::UeberBackend(std::string const&) (in /usr/sbin/pdns_server)
==3186==    by 0x3E21F0: apiServerZoneDetail(HttpRequest*, HttpResponse*) (in /usr/sbin/pdns_server)
==3186==
==3186== LEAK SUMMARY:
==3186==    definitely lost: 3,924,616 bytes in 19,624 blocks
==3186==    indirectly lost: 0 bytes in 0 blocks
==3186==      possibly lost: 204,733 bytes in 1,721 blocks
==3186==    still reachable: 5,245,137 bytes in 6,105 blocks
==3186==         suppressed: 0 bytes in 0 blocks
==3186== Reachable blocks (those to which a pointer was found) are not shown.
==3186== To see them, rerun with: --leak-check=full --show-leak-kinds=all
==3186==
==3186== ERROR SUMMARY: 1307 errors from 1307 contexts (suppressed: 0 from 0)
==3186== ERROR SUMMARY: 1307 errors from 1307 contexts (suppressed: 0 from 0)
```

- Normal DNS queries didn't cause any leak.

Habbie:
### Short description
Whenever a thread uses MySQL, make sure to call `mysql_thread_end()` when the thread terminates.

~NOTE: not PRed to master yet.~ #7517 

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/modules/gmysqlbackend/smysql.cc
+++ b/modules/gmysqlbackend/smysql.cc
@@ -38,6 +38,32 @@
 typedef bool my_bool;
 #endif
 
+/*
+ * Older versions of the MySQL and MariaDB client leak memory
+ * because they expect the application to call mysql_thread_end()
+ * when a thread ends. This thread_local static object provides
+ * that closure, but only when the user has asked for it
+ * by setting gmysql-thread-cleanup.
+ * For more discussion, see https://github.com/PowerDNS/pdns/issues/6231
+ */
+class MySQLThreadCloser
+{
+public:
+  ~MySQLThreadCloser() {
+    if(d_enabled) {
+      mysql_thread_end();
+    }
+  }
+  void enable() {
+   d_enabled = true;
+  }
+
+private:
+  bool d_enabled = false;
+};
+
+static thread_local MySQLThreadCloser threadcloser;
+
 bool SMySQL::s_dolog;
 pthread_mutex_t SMySQL::s_myinitlock = PTHREAD_MUTEX_INITIALIZER;
 
@@ -411,6 +437,10 @@ void SMySQL::connect()
   int retry=1;
 
   Lock l(&s_myinitlock);
+  if (d_threadCleanup) {
+    threadcloser.enable();
+  }
+
   if (!mysql_init(&d_db))
     throw sPerrorException("Unable to initialize mysql driver");
 
@@ -459,8 +489,8 @@ void SMySQL::connect()
 }
 
 SMySQL::SMySQL(const string &database, const string &host, uint16_t port, const string &msocket, const string &user,
-               const string &password, const string &group, bool setIsolation, unsigned int timeout):
-  d_database(database), d_host(host), d_msocket(msocket), d_user(user), d_password(password), d_group(group), d_timeout(timeout), d_port(port), d_setIsolation(setIsolation)
+               const string &password, const string &group, bool setIsolation, unsigned int timeout, bool threadCleanup):
+  d_database(database), d_host(host), d_msocket(msocket), d_user(user), d_password(password), d_group(group), d_timeout(timeout), d_port(port), d_setIsolation(setIsolation), d_threadCleanup(threadCleanup)
 {
   connect();
 }
--- a/modules/gmysqlbackend/gmysqlbackend.cc
+++ b/modules/gmysqlbackend/gmysqlbackend.cc
@@ -59,7 +59,8 @@ void gMySQLBackend::reconnect()
                    getArg("password"),
                    getArg("group"),
                    mustDo("innodb-read-committed"),
-                   getArgAsNum("timeout")));
+                   getArgAsNum("timeout"),
+                   mustDo("thread-cleanup")));
 }
 
 class gMySQLFactory : public BackendFactory
@@ -78,6 +79,7 @@ public:
     declare(suffix,"group", "Database backend MySQL 'group' to connect as", "client");
     declare(suffix,"innodb-read-committed","Use InnoDB READ-COMMITTED transaction isolation level","yes");
     declare(suffix,"timeout", "The timeout in seconds for each attempt to read/write to the server", "10");
+    declare(suffix,"thread-cleanup","Explicitly call mysql_thread_end() when threads end","no");
 
     declare(suffix,"dnssec","Enable DNSSEC processing","no");
 
--- a/modules/gmysqlbackend/smysql.hh
+++ b/modules/gmysqlbackend/smysql.hh
@@ -32,7 +32,8 @@ public:
   SMySQL(const string &database, const string &host="", uint16_t port=0,
          const string &msocket="",const string &user="",
          const string &password="", const string &group="",
-         bool setIsolation=false, unsigned int timeout=10);
+         bool setIsolation=false, unsigned int timeout=10,
+         bool threadCleanup=false);
 
   ~SMySQL();
 
@@ -61,6 +62,7 @@ private:
   unsigned int d_timeout;
   uint16_t d_port;
   bool d_setIsolation;
+  bool d_threadCleanup;
 };
 
 #endif /* SSMYSQL_HH */


modules/gmysqlbackend/smysql.cc
```diff
@@ -38,6 +38,22 @@
 typedef bool my_bool;
 #endif
 
+class MySQLThreadCloser
```
zeha: this black magic code should have a comment on the "why" :)

Habbie:
As pointed out by @rgacogne in conversation before I filed this PR:

> You could do per-thread refcounting of SMySQL objects, but you can’t call mysql_thread_end() if the refcount goes to 0 because we could still create a new one (note that it might work and just cause a new allocation, but the documentation is not clear on that point so it’s a bit iffy) and if the refcount isn’t at 0 when your destructor is called you won’t get called again. I guess you could then set a flag so that the last SMySQL destructor knows we are exiting the thread, but what if your flag is destructed before that?

We need to check if this is a problem in practice. The best answer I can hope for is 'hide this cleanup trick behind a pdns.conf flag and we will get away with it'.

Habbie:
With https://gist.github.com/Habbie/92f0a5b389339c09d7d0931addf94329:

API call:
```
298088192 SMySQL
298088192 MySQLThreadCloser enabled
298088192 ~SMySQL
298088192 ~MySQLThreadCloser closing ...  mysql_thread_end()
```

AXFR:
```
298088192 SMySQL
298088192 MySQLThreadCloser enabled
323266304 SMySQL
323266304 MySQLThreadCloser enabled
306480896 SMySQL
314873600 SMySQL
306480896 MySQLThreadCloser enabled
314873600 MySQLThreadCloser enabled
314873600 ~SMySQL
314873600 ~SMySQL
323266304 ~SMySQL
314873600 ~MySQLThreadCloser closing ...  mysql_thread_end()
314873600 ~MySQLThreadCloser closing ...  mysql_thread_end()
323266304 ~MySQLThreadCloser closing ...  mysql_thread_end()
306480896 ~SMySQL
306480896 ~MySQLThreadCloser closing ...  mysql_thread_end()
298088192 ~SMySQL
298088192 ~MySQLThreadCloser closing ...  mysql_thread_end()
```

modules/gmysqlbackend/smysql.cc
Habbie: Yes, agreed!

Habbie:
> With https://gist.github.com/Habbie/92f0a5b389339c09d7d0931addf94329, and this PR cherry-picked on top of master (available at https://github.com/Habbie/pdns/tree/mysqlthreadcloser)

This debugging output is not portable (works on el7, doesn't work on OSX):
```
  CXX      smysql.lo
smysql.cc:44:10: error: cannot initialize return object of type 'int' with an rvalue of type 'pthread_t _Nonnull' (aka '_opaque_pthread_t *')
  return pthread_self();
         ^~~~~~~~~~~~~~
1 error generated.
make: *** [smysql.lo] Error 1
```

rgacogne:
I think this might work (with some caveats below) because:
- destructors of `thread_local` variables are called before pthread thread-specific data is destroyed [1] ;
- objects with an automatic storage duration will already have been destroyed when `thread_local` objects are destroyed.

But that means that we need to make sure that no `UeberBackend` object outlives the duration of the thread that created it. This means no global (static), thread-local (because then the order of destruction is very hard to predict), or static local objects. Also we need to be careful about dynamically allocated objects outliving their creator thread.
I already see one possible issue in [2], a static local `UeberBackend` that might be created in one thread and be used by another one after the initial thread has been destroyed, but there might be other. :-/

[1]: https://sourceware.org/glibc/wiki/Destructor%20support%20for%20thread_local%20variables
[2]: https://github.com/PowerDNS/pdns/blob/master/pdns/lua-record.cc#L387

omoerbeek:
After reading the mysql 5.5 docs on `mysql_thread_end()` and related pages (e.g. https://dev.mysql.com/doc/refman/5.5/en/c-api-threaded-clients.html):

* `mysql_thread_init()` should be called (directly or indirectly via `mysql_connect()` for every thread that uses mysql functions.
* `mysql_thread_end()` should be called just before `pthread_exit()`
* The case where a `mysql_connect()` is done by one thread but the resulting handle is used by another is covered as well. A thread using mysql functions should call `mysql_thread_init()` even if it din not create the mysql connection. And per consequence it should also call `mysql_thread_end()` on thread exit.

I suspect that sharing sql connections between threads could cause trouble if the thread that did not create the connection uses mysql functions and did not call `mysql_thread_init()`. But likely we should have seen this trouble already if that happens in the current code. If the thread that uses but not created the mysql connection exits at some point, it should call  `mysql_thread_end()`. If it lives forever, no active leak occurs. In general I say that sharing mysql connection objects between threads is something that makes a hard problem even harder and should be avoided.

As for short lived threads, it looks like the mechanism provided is sufficient, provided that we keep the sql connection to the thread and no mysql functions are used *by the thread exiting* between the call to `mysql_thread_end()` and `pthread_exit()`.  As we are already in thread teardown state when the  C++ thread local destructor calls are done this abusive use of mysql functions can only be achieved by (other) thread local destructor calls, which can be avoided easily.   

So I agree with Remi's analysis, we are likely safe (with some reservations). 

We should probably make this thread local object mechanism something conditional. If we know exactly which client libs expose the problem, we could use https://dev.mysql.com/doc/refman/5.5/en/mysql-get-client-version.html to decide to enable the mechanism. Otherwise use a config flag.

omoerbeek:
> 314873600 ~MySQLThreadCloser closing ...  mysql_thread_end()
> 314873600 ~MySQLThreadCloser closing ...  mysql_thread_end()

Hmm, a double call, that should not happen, right?

Habbie:
> Only one create, but two dt calls.. Is this log complete?

Well spotted, this is weird. Will try to rerun that test.

omoerbeek:
To avoid the double call to `mysql_thread_end()` the bool should probably be a counter: increment on enable, decrement in dt and call `mysql_thread_end()` if zero.

Habbie:
> To avoid the double call to `mysql_thread_end()` the bool should probably be a counter: increment on enable, decrement in dt and call `mysql_thread_end()` if zero.

That sounds like the kind of thing that would -cause- double calls...

omoerbeek:
Yes, that is one step further. First need to find out why there is a single construct with two dt calls in the log.

Habbie:
> Yes, that is one step further. First need to find out why there is a single construct with two dt calls in the log.

That version does not log the construct, just the enable, but it indeed logs just one enable..

Habbie:
> Hmm, a double call, that should not happen, right?

It usually doesn't. I've added `this` to the logging and retried until it happened again:
```
-1300248832 SMySQL
-1300248832 0x7fefb27fc6f9 MySQLThreadCloser construct
-1300248832 0x7fefb27fc6f9 MySQLThreadCloser enabled
-1325426944 SMySQL
-1325426944 0x7fefb0ff96f9 MySQLThreadCloser construct
-1325426944 0x7fefb0ff96f9 MySQLThreadCloser enabled
-1317034240 SMySQL
-1308641536 SMySQL
-1317034240 0x7fefb17fa6f9 MySQLThreadCloser construct
-1317034240 0x7fefb17fa6f9 MySQLThreadCloser enabled
-1308641536 0x7fefb1ffb6f9 MySQLThreadCloser construct
-1308641536 0x7fefb1ffb6f9 MySQLThreadCloser enabled
-1308641536 ~SMySQL
-1308641536 ~SMySQL
-1317034240 ~SMySQL
-1308641536 0x7fefb1ffb6f9 ~MySQLThreadCloser closing ...  mysql_thread_end()
-1308641536 0x7fefb1ffb6f9 ~MySQLThreadCloser closing ...  mysql_thread_end()
Feb 25 12:35:08 AXFR of domain 'example.com' to 127.0.0.1:46709 finished
-1317034240 0x7fefb17fa6f9 ~MySQLThreadCloser closing ...  mysql_thread_end()
-1325426944 ~SMySQL
-1325426944 0x7fefb0ff96f9 ~MySQLThreadCloser closing ...  mysql_thread_end()
-1300248832 ~SMySQL
-1300248832 0x7fefb27fc6f9 ~MySQLThreadCloser closing ...  mysql_thread_end()
```

4 constructs, 4 enables, 5 destruct+close

Habbie:
It turns out I'm using `cerr` in a thread unsafe way, which is duplicating output lines. With `static std::atomic<int> s_unique;` and `(s_unique++)` once in every cerr line, here's the smoking gun:

```
1518323456 SMySQL
1518323456 0x7f6a5a7fc6ec MySQLThreadCloser constructed, d_constructed=1, s_unique=32
1518323456 0x7f6a5a7fc6ec MySQLThreadCloser enabled, s_unique=33
1509930752 SMySQL
1509930752 0x7f6a59ffb6ec MySQLThreadCloser constructed, d_constructed=1, s_unique=34
1509930752 0x7f6a59ffb6ec MySQLThreadCloser enabled, s_unique=35
1493145344 SMySQL
1501538048 SMySQL
1493145344 0x7f6a58ff96ec MySQLThreadCloser constructed, d_constructed=1, s_unique=36
1493145344 0x7f6a58ff96ec MySQLThreadCloser enabled, s_unique=37
1501538048 0x7f6a597fa6ec MySQLThreadCloser constructed, d_constructed=1, s_unique=38
1501538048 0x7f6a597fa6ec MySQLThreadCloser enabled, s_unique=39
1493145344 ~SMySQL
1509930752 ~SMySQL
1493145344 0x7f6a58ff96ec ~MySQLThreadCloser closing ... , d_constructed=1, d_destructed=1, s_unique=40 mysql_thread_end()
1493145344 0x7f6a58ff96ec ~MySQLThreadCloser closing ... , d_constructed=1, d_destructed=1, s_unique=40 mysql_thread_end()
1501538048 ~SMySQL
1493145344 0x7f6a58ff96ec ~MySQLThreadCloser closing ... , d_constructed=1, d_destructed=1, s_unique=40 mysql_thread_end()
1501538048 ~SMySQL
1509930752 0x7f6a59ffb6ec ~MySQLThreadCloser closing ... , d_constructed=1, d_destructed=1, s_unique=41 mysql_thread_end()
1501538048 0x7f6a597fa6ec ~MySQLThreadCloser closing ... , d_constructed=1, d_destructed=1, s_unique=42 mysql_thread_end()
1518323456 ~SMySQL
1518323456 0x7f6a5a7fc6ec ~MySQLThreadCloser closing ... , d_constructed=1, d_destructed=1, s_unique=43 mysql_thread_end()
```

Habbie:
#7517 has a configurable version of this for master, please review

Habbie:
Now identical to #7517, minus the documentation changes. Ready for review, merge.

PR #7529, opened 2019-02-28
auth: Clear caches (meta-data, keys) on domain deletion

rgacogne:
### Short description

Alternative fix for #7407.

### Checklist

I have:
- [x] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/ws-auth.cc
+++ b/pdns/ws-auth.cc
@@ -1676,6 +1676,7 @@ static void apiServerZoneDetail(HttpRequest* req, HttpResponse* resp) {
     // clear caches
     DNSSECKeeper dk(&B);
     dk.clearCaches(zonename);
+    purgeAuthCaches(zonename.toString() + "$");
 
     // empty body on success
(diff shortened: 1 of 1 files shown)


mind04:
What about clearing the packet cache and maybe even more important the query cache?

rgacogne:
> What about clearing the packet cache and maybe even more important the query cache?

It would make a lot of sense, yes, I'll add a `purgeAuthCachesExact(zonename);`.

rgacogne:
> It would make a lot of sense, yes, I'll add a `purgeAuthCachesExact(zonename);`.

Or not 'Exact', actually.

zeha:
👍

rgacogne:
Done.

rgacogne:
> Would it make sense to also clear caches in pdnsutil `deleteZone()` ?

I believe it would, but I don't think we can do that from `pdnsutil`, can we?

PR #7537, opened 2019-03-01
DNSNameSet and QNameSetRule

andreydomas:
### Short description
Set of DNSName based on std::set.
Query rule for test if set contains qname.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [X] documented the code

### DIFF
--- a/pdns/dnsdist-lua-bindings.cc
+++ b/pdns/dnsdist-lua-bindings.cc
@@ -180,7 +180,7 @@ void setupLuaBindings(bool client)
   /* DNSNameSet */
   g_lua.registerFunction<string(DNSNameSet::*)()>("toString", [](const DNSNameSet&dns ) { return dns.toString(); });
   g_lua.registerFunction<void(DNSNameSet::*)(DNSName&)>("add", [](DNSNameSet& dns, DNSName& dn) { dns.insert(dn); });
-  g_lua.registerFunction<bool(DNSNameSet::*)(DNSName&)>("contains", [](DNSNameSet& dns, DNSName& dn) { return dns.find(dn) != dns.end(); });
+  g_lua.registerFunction<bool(DNSNameSet::*)(DNSName&)>("check", [](DNSNameSet& dns, DNSName& dn) { return dns.find(dn) != dns.end(); });
   g_lua.registerFunction("delete",(size_t (DNSNameSet::*)(const DNSName&)) &DNSNameSet::erase);
   g_lua.registerFunction("size",(size_t (DNSNameSet::*)() const) &DNSNameSet::size);
   g_lua.registerFunction("clear",(void (DNSNameSet::*)()) &DNSNameSet::clear);
--- a/pdns/dnsdist-lua-bindings.cc
+++ b/pdns/dnsdist-lua-bindings.cc
@@ -175,6 +175,16 @@ void setupLuaBindings(bool client)
   g_lua.registerFunction<string(DNSName::*)()>("toString", [](const DNSName&dn ) { return dn.toString(); });
   g_lua.writeFunction("newDNSName", [](const std::string& name) { return DNSName(name); });
   g_lua.writeFunction("newSuffixMatchNode", []() { return SuffixMatchNode(); });
+  g_lua.writeFunction("newDNSNameSet", []() { return DNSNameSet(); });
+
+  /* DNSNameSet */
+  g_lua.registerFunction<string(DNSNameSet::*)()>("toString", [](const DNSNameSet&dns ) { return dns.toString(); });
+  g_lua.registerFunction<void(DNSNameSet::*)(DNSName&)>("add", [](DNSNameSet& dns, DNSName& dn) { dns.insert(dn); });
+  g_lua.registerFunction<bool(DNSNameSet::*)(DNSName&)>("contains", [](DNSNameSet& dns, DNSName& dn) { return dns.find(dn) != dns.end(); });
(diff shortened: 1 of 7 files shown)


pdns/dnsdist-lua-bindings.cc
```diff
@@ -175,6 +175,16 @@ void setupLuaBindings(bool client)
   g_lua.registerFunction<string(DNSName::*)()>("toString", [](const DNSName&dn ) { return dn.toString(); });
   g_lua.writeFunction("newDNSName", [](const std::string& name) { return DNSName(name); });
   g_lua.writeFunction("newSuffixMatchNode", []() { return SuffixMatchNode(); });
+  g_lua.writeFunction("newDNSNameSet", []() { return DNSNameSet(); });
+
+  /* DNSNameSet */
+  g_lua.registerFunction<string(DNSNameSet::*)()>("toString", [](const DNSNameSet&dns ) { return dns.toString(); });
+  g_lua.registerFunction<void(DNSNameSet::*)(DNSName&)>("add", [](DNSNameSet& dns, DNSName& dn) { dns.insert(dn); });
+  g_lua.registerFunction<bool(DNSNameSet::*)(DNSName&)>("contains", [](DNSNameSet& dns, DNSName& dn) { return dns.find(dn) != dns.end(); });
```
rgacogne: Just a nit but we have `SuffixMatchNode::check()`, perhaps it would make sense to be consistent (although I agree that `::contains()` is actually more explicit)?
```suggestion
  g_lua.registerFunction<bool(DNSNameSet::*)(DNSName&)>("check", [](DNSNameSet& dns, DNSName& dn) { return dns.find(dn) != dns.end(); });
```

pdns/dnsdistdist/dnsdist-rules.hh
```diff
@@ -542,6 +543,22 @@ private:
   DNSName d_qname;
 };
 
+class QNameSetRule : public DNSRule {
+public:
+    QNameSetRule(const DNSNameSet names) : qname_idx(names) {}
```
rgacogne: It doesn't really matter but I'm pretty sure we will get a warning from `cppcheck` otherwise:
```suggestion
    QNameSetRule(const DNSNameSet& names) : qname_idx(names) {}
```

pdns/dnsdistdist/docs/rules-actions.rst
```diff
@@ -610,6 +610,11 @@ These ``DNSRule``\ s be one of the following items:
 
    :param string qname: Qname to match
 
+.. function:: QNameSetRule(set)
+   Matches if the set contains qname.
```
rgacogne: It might be nice to be explicit about the difference between `QNameSetRule` and `SuffixMatchNodeRule`, ie exact vs suffix match.

rgacogne: Mentioning `QnameSetRule` in the documentation for `SuffixMatchNodeRule` might make sense as well.

ahupowerdns: So I think we must make this difference very explicit, because it confused me too.

pdns/dnsname.hh
```diff
@@ -376,3 +378,11 @@ bool DNSName::operator==(const DNSName& rhs) const
 }
 
 extern const DNSName g_rootdnsname, g_wildcarddnsname;
+
+struct DNSNameSet: public std::set<DNSName> {
```
ahupowerdns: Why not an std::unordered_set? This is almost always way faster.

pdns/dnsdist-lua-bindings.cc
andreydomas: Ok, changed it to `::check()`

pdns/dnsdistdist/dnsdist-rules.hh
andreydomas: Fixed.

pdns/dnsdistdist/docs/rules-actions.rst
andreydomas: Wrote more about the difference and add mentionings.

pdns/dnsname.hh
andreydomas: Right you are, I was in a hurry. Fixed.

pdns/dnsdistdist/docs/reference/dnsnameset.rst
```diff
@@ -0,0 +1,60 @@
+.. _DNSNameSet:
+
+DNSNameSet objects
+==================
+
+A :class:`DNSNameSet` object is a set of :class:`DNSName` objects. 
+Based on std::set (usually implemented as red-black trees).
```
rgacogne: This line is not true anymore; sorry about that.

pdns/dnsdistdist/docs/rules-actions.rst
```diff
@@ -610,6 +610,13 @@ These ``DNSRule``\ s be one of the following items:
 
    :param string qname: Qname to match
 
+.. function:: QNameSetRule(set)
+  Matches if the set contains excact qname.
```
rgacogne:
```suggestion
  Matches if the set contains exact qname.
```

pdns/dnsdistdist/docs/reference/dnsnameset.rst
andreydomas: Fixed.

PR #7538, opened 2019-03-05
Rec dnstap

omoerbeek:
### Short description

Whenever the recursor sends a query to an auth server (on a cache miss) generate a dnstap formatted message with the query and/or response and send that message using a frame stream connection to one or more framestream servers. 

Options are implemented to disable logging of queries and responses respectively and to set the various parameters of the framestream lib.

Note that typically no framestream lib will be avaialble. Work is being done to change that (although we might need to distribute the lib ourselves).

This code is only active if the configure option -with-dnstap is active.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [X] documented the code


pdns/dnstap.cc
```diff
@@ -2,7 +2,7 @@
 #include "gettime.hh"
 #include "dnstap.hh"
 
-DnstapMessage::DnstapMessage(const std::string& identity, const ComboAddress* requestor, const ComboAddress* responder, bool isTCP, const char* packet, const size_t len, const struct timespec* queryTime, const struct timespec* responseTime)
+DnstapMessage::DnstapMessage(const std::string& identity, const ComboAddress* requestor, const ComboAddress* responder, bool isTCP, const char* packet, const size_t len, const struct timespec* queryTime, const struct timespec* responseTime, bool recursor)
```
rgacogne: Perhaps a RecDnstapMessage would be better than a `recursor` boolean, as we do for protobuf?

pdns/fstrm_logger.cc
```diff
@@ -9,7 +9,8 @@
 
 #ifdef HAVE_FSTRM
 
-FrameStreamLogger::FrameStreamLogger(const int family, const std::string& address, bool connect): d_family(family), d_address(address)
+FrameStreamLogger::FrameStreamLogger(const int family, const std::string& address, bool connect,
+    const std::unordered_map<string,unsigned>& options): d_family(family), d_address(address)
```
rgacogne: I'm not very fond of passing a map of options to a constructor, I'd rather see setters if we have too many of them. Please feel free to disagree, though :)

pdns/pdns_recursor.cc
```diff
@@ -121,6 +121,11 @@ static thread_local std::shared_ptr<std::vector<std::unique_ptr<RemoteLogger>>>
 static thread_local uint64_t t_outgoingProtobufServersGeneration;
 #endif /* HAVE_PROTOBUF */
 
+#ifdef HAVE_FSTRM
+static thread_local std::shared_ptr<std::vector<std::unique_ptr<RemoteLoggerInterface>>> t_frameStreamServers{nullptr};
```
rgacogne: I'm not sure why you are not storing these as `FrameStreamLogger` instead of `RemoteLoggerInterface`?

pdns/recursordist/Makefile.am
```diff
@@ -368,6 +388,14 @@ pdns_recursor_SOURCES += \
 
 endif
 
+if FSTRM
+pdns_recursor_SOURCES += \
+	dnstap.cc dnstap.hh fstrm_logger.cc fstrm_logger.hh dolog.hh
```
rgacogne: I don't think we should use `dolog.hh` in the rec, perhaps we can do the same kind of ugly hack we currently have in `remote_logger.cc` with `#ifdef PDNS_CONFIG_ARGS` ?

pdns/dnstap.cc
omoerbeek: Yes. Will do.

pdns/fstrm_logger.cc
omoerbeek: The problem is that the `FrameStreamLogger` create the framestream connection on construct. Setters would be too late. Options:
* rework the FrameStreamLogger with a three step setup (construct; set setters; connect)
* use a settings object: `frameStreamSettings.setOption(x)....; FrameStreamLogger(... frameStreamSettings);`

pdns/pdns_recursor.cc
omoerbeek: The `FrameStreamLogger` class is not available if !HAVE_FRAMESTREAM. But I could make it a dummy object in that case if you think that is more clear.

pdns/fstrm_logger.cc
rgacogne: Ah, right, I missed that indeed! Let's keep it that way for now, then.

pdns/pdns_recursor.cc
rgacogne: Do you mean if `!HAVE_FSTRM`? But in that case this variable doesn't exist at all? Or am missing something?

pdns/recursordist/Makefile.am
pieterlexis: > `#ifdef PDNS_CONFIG_ARGS`

We have a `RECURSOR` variable as well

pdns/pdns_recursor.cc
omoerbeek: I mean !HAVE_FSTRM indeed. Sorry for the confusion. The actual problem is in `lwrec.c asyncresolve()`, there the type of the `fstrmLoggers` vector needs to be known.

rgacogne: Understood, thanks! I usually prefer a dummy object in that case rather than having to cast, yes, mostly because I find it easier to read afterwards, but that's only my personal preference :-)

pdns/recursordist/Makefile.am
omoerbeek: Will fix in upcoming commit (using RECURSOR)

pdns/pdns_recursor.cc
omoerbeek: Well, actually I might share that preference, I'll rework this.

pdns/recursordist/Makefile.am
```diff
@@ -390,7 +390,7 @@ endif
 
 if FSTRM
 pdns_recursor_SOURCES += \
-	dnstap.cc dnstap.hh fstrm_logger.cc fstrm_logger.hh
+	dnstap.cc dnstap.hh fstrm_logger.cc fstrm_logger.hh rec-dnstap.hh
```
pieterlexis: pretty sure this needs to go into `EXTRA_DIST` as well

regression-tests.recursor-dnssec/test_RecDnstap.py
```diff
@@ -0,0 +1,301 @@
...
+def checkDnstapQuery(testinstance, dnstap, protocol, query, initiator='127.0.0.1'):
+    testinstance.assertEquals(dnstap.message.type, dnstap_pb2.Message.RESOLVER_QUERY)
+    checkDnstapBase(testinstance, dnstap, protocol, initiator)
+
+    testinstance.assertTrue(dnstap.message.HasField('query_time_sec'))
+    testinstance.assertTrue(dnstap.message.HasField('query_time_nsec'))
+
+    testinstance.assertTrue(dnstap.message.HasField('query_message'))
+    wire_message = dns.message.from_wire(dnstap.message.query_message)
+    #testinstance.assertEqual(wire_message, query)
```
rgacogne: It wouldn't mind a comment explaining that we can't compare the outgoing query with the incoming one.

PR #7548, opened 2019-03-08
Eliminate the loop in SyncRes::getAddrs()

omoerbeek:
Makes the code much more readable.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [X] documented the code

### DIFF
--- a/pdns/syncres.cc
+++ b/pdns/syncres.cc
@@ -694,9 +694,9 @@ vector<ComboAddress> SyncRes::getAddrs(const DNSName &qname, unsigned int depth,
   res_t resv4;
   // If IPv4 ever becomes second class, we should revisit this
   if (doResolve(qname, QType::A, resv4, depth+1, beenthere, newState) == 0) {  // this consults cache, OR goes out
-    for (auto i = resv4.cbegin(); i != resv4.end(); ++i) {
-      if (i->d_type == QType::A) {
-        if (auto rec = getRR<ARecordContent>(*i)) {
+    for (auto const &i : resv4) {
+      if (i.d_type == QType::A) {
+        if (auto rec = getRR<ARecordContent>(i)) {
           ret.push_back(rec->getCA(53));
         }
       }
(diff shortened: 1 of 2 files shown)


pdns/syncres.cc
```diff
@@ -680,61 +680,54 @@ struct speedOrderCA
...
-	      }
-	    }
-	  }
-	}
+  }
+  if (s_doIPv6) {
+    if (ret.empty()) {
+      // We did not find IPv4 addresses, try to get IPv6 ones
+      vState newState = Indeterminate;
+      res_t res;
```
rgacogne: The two declarations above shadow existing variables, we should avoid that whenever possible.

pdns/syncres.cc
```diff
@@ -680,61 +680,54 @@ struct speedOrderCA
...
-	  if(auto rec = getRR<ARecordContent>(*i))
-	    ret.push_back(rec->getCA(53));
-	  else if(auto aaaarec = getRR<AAAARecordContent>(*i))
-	    ret.push_back(aaaarec->getCA(53));
-          done=true;
+  vState newState = Indeterminate;
+  res_t res;
+  // If IPv4 ever becomes second class, we should revisit this
+  if (doResolve(qname, QType::A, res, depth+1, beenthere, newState) == 0) {  // this consults cache, OR goes out
+    for (res_t::const_iterator i = res.begin(); i != res.end(); ++i) {
```
rgacogne: Just a nit but it looks like a C++11-style loop would be easier to read.

pdns/syncres.cc
omoerbeek: Agreed, will fix.

pdns/syncres.cc
omoerbeek: Right, will do.

pdns/syncres.cc
```diff
@@ -680,61 +680,54 @@ struct speedOrderCA
...
-	  if(auto rec = getRR<ARecordContent>(*i))
-	    ret.push_back(rec->getCA(53));
-	  else if(auto aaaarec = getRR<AAAARecordContent>(*i))
-	    ret.push_back(aaaarec->getCA(53));
-          done=true;
+  vState newState = Indeterminate;
+  res_t resv4;
+  // If IPv4 ever becomes second class, we should revisit this
+  if (doResolve(qname, QType::A, resv4, depth+1, beenthere, newState) == 0) {  // this consults cache, OR goes out
+    for (auto i = resv4.cbegin(); i != resv4.end(); ++i) {
```
```suggestion
    for (const auto& i : resv4) {
```

Note that `i` becomes a const reference to the actual type, and not an iterator anymore so it implies editing a bit more code than just that line.

PR #7550, opened 2019-03-08
auth web: make max request/response body size configurable

Habbie:
### Short description
Because some people have big zones.

Bumping the buffer size from 1024 to 16000 changes the read time for a 4 megabyte request from 2 minutes to around 10 seconds.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/recursordist/docs/settings.rst
+++ b/pdns/recursordist/docs/settings.rst
@@ -1725,17 +1725,6 @@ The value between the hooks is a UUID that is generated for each request. This c
 .. note::
   The webserver logs these line on the NOTICE level. The :ref:`settings-loglevel` seting must be 5 or higher for these lines to end up in the log.
 
-.. _setting-webserver-max-bodysize:
-
-``webserver-max-bodysize``
---------------------------
-.. versionadded:: 4.2.0
-
--  Integer
--  Default: 2
-
-Maximum request/response body size in megabytes.
-
 .. _setting-webserver-password:
 
 ``webserver-password``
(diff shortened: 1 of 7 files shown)


pdns/webserver.cc
```diff
@@ -281,13 +281,14 @@ try {
...
   yarl.initialize(&req);
+  req.max_request_size=d_maxbodysize;
   int timeout = 5;
   client->setNonBlocking();
 
   try {
     while(!req.complete) {
       int bytes;
-      char buf[1024];
+      char buf[16000];
```
rgacogne: I wonder if we should make this configurable (and in that case perhaps move it to the heap?). This should be fine on most OS, though.

pdns/recursordist/docs/settings.rst
```diff
@@ -1523,6 +1523,15 @@ IP address for the webserver to listen on.
 
 These subnets are allowed to access the webserver.
 
+.. _setting-webserver-max-bodysize:
+
+``webserver-max-bodysize``
+--------------------------
+-  Integer
```
```suggestion
.. versionadded:: 4.2.0

-  Integer
```

Habbie: Oh yes, sorry. Will hold off on this a bit in case we also backport.

zeha:
Maybe default to 20MB?

Habbie:
Rebased without changes.

Habbie:
> Maybe default to 20MB?

As long as auth happens after receiving the whole request, I'm more comfortable with a smaller limit.

pdns/webserver.cc
Habbie: At 16000 my impression is that actual processing of the request will be a way bigger part of execution time anyway, so I'm not sure we need yet another config option.

pdns/recursordist/docs/settings.rst
Habbie: committed

PR #7526, opened 2019-02-27
 dnsdist: Refactoring to merge the UDP and TCP paths

rgacogne:
### Short description

This PR refactors the handling of queries and responses a bit so that the existing code for the UDP and TCP paths can be merged into one. This reduces code duplication and will make it easier to add another processing path in the future, like DoH or a more scalable TCP/DoT stack.
It also removes a bit of code duplication in the rules, and create empty objects for the DNSCrypt-related classes when support is disabled, so that we can mostly ignore whether it's enabled in most parts of the code.

### Checklist

I have:
- [x] read the [CONTRIBUTING.md](https://github.com/PowerDNS/pdns/blob/master/CONTRIBUTING.md) document
- [x] compiled this code
- [x] tested this code
- [ ] included documentation (including possible behaviour changes)
- [x] documented the code
- [x] added or modified regression test(s)
- [x] added or modified unit test(s)

PR #7559, opened 2019-03-11
dnsdist: Refactoring of the TCP stack

### LINKED ISSUE #4814: dnsdist TCP stack needs improving
- Program: dnsdist 
 - Issue type: Bug report

### Description
The current dnsdist TCP stack has some weaknesses particularly when it is used in front of recursors. Due to the way dnsdist handles distributing queries over its tcp threads it can cause a "jam" of (very) slow queries on a certain thread. When it then assigns a "normal" query to this thread this query might timeout. Another case is that the queue for a certain thread can fill up the entire global TCP queue while the other thread still have plenty of time for processing queries causing queries to get dropped since the queue is "full".

These effects can be mitigated by spawning *alot* of tcp threads or setting the tcp recv timeouts on the server very low, but neither solution is very desirable.

### Environment
 - Operating system: Any 
 - Software version: 1.0.0, 1.1.0-beta2, git (probably)

### Steps to reproduce
The easiest way to reproduce is to fire a bunch of known slow queries at dnsdist. dnsdist will then stop responding properly to normal queries as well or at least encountering failures at random (although I managed to temporarily break it completely with about 1 query every 2 seconds in my case).

rgacogne:
### Short description

This PR completely changes the way dnsdist handles TCP connections, moving to an event-based mode that allows a single thread to handle a large number of connections, instead of only one. It makes DNS over TCP and DNS over TLS much more scalable.
This needs a very serious review, lots of testing, including weird corner-cases like TCP Fast Open, etc.. 
The documentation needs to be updated, and probably the way we handle the number of TCP threads to start.

Based on #7526. Hopefully fixes #4814.

IMHO merging this PR calls for the move to the 1.4.x branch at least, perhaps 2.0.x.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdist-tcp.cc
+++ b/pdns/dnsdist-tcp.cc
@@ -302,7 +302,7 @@ IOState tryRead(int fd, std::vector<uint8_t>& buffer, size_t& pos, size_t toRead
   return IOState::Done;
 }
 
-std::shared_ptr<TCPClientCollection> g_tcpclientthreads;
+std::unique_ptr<TCPClientCollection> g_tcpclientthreads;
 
 class TCPClientThreadData
 {
--- a/pdns/dnsdist-tcp.cc
+++ b/pdns/dnsdist-tcp.cc
@@ -880,12 +880,15 @@ static void handleDownstreamIOCallback(int fd, FDMultiplexer::funcparam_t& param
        Let's just drop the connection
     */
     vinfolog("Got an exception while handling (%s backend) TCP query from %s: %s", (state->d_lastIOState == IOState::NeedRead ? "reading from" : "writing to"), state->d_ci.remote.toStringWithPort(), e.what());
-    /* remove this FD from the IO multiplexer */
-    ++state->d_downstreamFailures;
+    /* don't increase this counter when reusing connections */
+    if (state->d_freshDownstreamConnection) {
+      ++state->d_downstreamFailures;
+    }
     if (state->d_outstanding && state->d_ds != nullptr) {
       --state->d_ds->outstanding;
       state->d_outstanding = false;
     }
+    /* remove this FD from the IO multiplexer */
     iostate = IOState::Done;
     connectionDied = true;
   }
--- a/pdns/dnsdist-tcp.cc
+++ b/pdns/dnsdist-tcp.cc
@@ -332,7 +332,7 @@ static void handleDownstreamIOCallback(int fd, FDMultiplexer::funcparam_t& param
 class IncomingTCPConnectionState
 {
 public:
-  IncomingTCPConnectionState(ConnectionInfo&& ci, TCPClientThreadData& threadData, time_t now): d_buffer(4096), d_responseBuffer(4096), d_threadData(threadData), d_ci(std::move(ci)), d_handler(d_ci.fd, g_tcpRecvTimeout, d_ci.cs->tlsFrontend ? d_ci.cs->tlsFrontend->getContext() : nullptr, now), d_connectionStartTime(now)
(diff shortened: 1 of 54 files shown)


rgacogne:
Rebased to fix a conflict.

pdns/dnsdist-tcp.cc
```diff
@@ -201,544 +260,863 @@ void TCPClientCollection::addTCPClientThread()
...
+      return;
+    }
+    throw std::runtime_error("Error while reading from the TCP acceptor pipe (" + std::to_string(pipefd) + ") in " + std::string(isNonBlocking(pipefd) ? "non-blocking" : "blocking") + " mde:" + strerror(errno));
+  }
+  else if (got != sizeof(citmp)) {
+    throw std::runtime_error("Partial read while reading from the TCP acceptor pipe (" + std::to_string(pipefd) + ") in " + std::string(isNonBlocking(pipefd) ? "non-blocking" : "blocking") + " mode");
+  }
+
+  g_tcpclientthreads->decrementQueuedCount();
+  auto ci = std::move(*citmp);
```
chbruyand: I was wondering if we could get rid of `ci` and just construct `state` directly by moving `citmp`'s content

pdns/dnsdist-tcp.cc
```diff
@@ -201,544 +260,863 @@ void TCPClientCollection::addTCPClientThread()
...
+
+  setThreadName("dnsdist/tcpClie");
+
+  TCPClientThreadData data;
+
+  data.mplexer->addReadFD(pipefd, handleIncomingTCPQuery, &data);
+  time_t lastTCPCleanup = time(nullptr);
+  time_t lastTimeoutScan = time(nullptr);
+  struct timeval now;
+  gettimeofday(&now, 0);
```
chbruyand: Maybe we can just init `lastTCPCleanup` and `lastTimeoutScan` with `now.tv_sec`

pdns/tcpiohandler.hh
```diff
@@ -176,6 +193,77 @@ public:
...
     }
   }
+
+  /* Tries to read exactly toRead bytes into the buffer, starting at position pos.
+     Updates pos everytime a successful read occurs,
+     throws an std::runtime_error in case of IO error,
+     return Done when toRead bytes have been read, needRead or needWrite if the IO operation
+     would block.
+  */
+  IOState tryRead(std::vector<uint8_t>& buffer, size_t& pos, size_t toRead)
```
chbruyand: Wow, looks like the same code as in `dnsdist-tcp.cc`

pdns/dnsdist-tcp.cc
```diff
@@ -201,544 +260,863 @@ void TCPClientCollection::addTCPClientThread()
...
+}
+
+static void handleNewIOState(std::shared_ptr<IncomingTCPConnectionState>& state, IOState iostate, const int fd, FDMultiplexer::callbackfunc_t callback, boost::optional<struct timeval> ttd)
+{
+  //cerr<<"in "<<__func__<<" for fd "<<fd<<", last state was "<<(int)state->d_lastIOState<<", new state is "<<(int)iostate<<endl;
+
+  if (state->d_lastIOState == IOState::NeedRead && iostate != IOState::NeedRead) {
+    state->d_threadData.mplexer->removeReadFD(fd);
+    //cerr<<__func__<<": remove read FD "<<fd<<endl;
+    state->d_lastIOState = IOState::Done;
```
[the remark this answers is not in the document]
rgacogne: But if the next state (`iostate`) is `NeedWrite` we need to ask the multiplexer to let us know when the FD is writable, right?

pdns/tcpiohandler.hh
rgacogne: Yes :-) That's why the one in `dnsdist-tcp.cc` has this comment:

> // XXX could probably be implemented as a TCPIOHandler

but I haven't looked into it yet. It would be nice but it can wait a bit.

pdns/dnsdist-tcp.cc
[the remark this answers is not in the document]
chbruyand: Indeed, got it.

rgacogne:
Just pushed a few commits, the only important one being 705cf3caf8544fe681567fc5501114a3ace7385d that fixes a nice bug with DoT, uncovered while testing with https://github.com/DNS-OARC/flamethrower

rgacogne:
Rebased to fix conflicts.

PR #7569, opened 2019-03-12
Improve error reporting with garbage in the 'master' field of the database

pieterlexis:
### Short description
This improves the error reporting when running in `slave` mode, as well as via `pdnsutil` and the API.

Before, we should silently ignore slave zones with broken IPs in 'master'. Now we report on this:

```
Mar 12 16:25:55 Could not parse master address (foo) for zone 'example.com': Unable to convert presentation address 'foo'
Mar 12 16:25:55 No masters for slave zone 'example.com' found in the database
Mar 12 16:25:55 No new unfresh slave domains, 0 queued for AXFR already, 0 in progress
```

```
$ pdnsutil check-zone example.com
[Error] non-IP address for masters: Unable to convert presentation address 'foo'
[Error] No SOA record present, or active, in zone 'example.com'
Checked 0 records of 'example.com', 2 errors, 0 warnings.
```

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/backends/gsql/gsqlbackend.cc
+++ b/pdns/backends/gsql/gsqlbackend.cc
@@ -339,12 +339,16 @@ void GSQLBackend::getUnfreshSlaveInfos(vector<DomainInfo> *unfreshDomains)
 
   vector<DomainInfo> allSlaves;
 
+  bool loggedAssertRowColumns = false;
   for(const auto& row : d_result) { // id,name,master,last_check
     DomainInfo sd;
     try {
       ASSERT_ROW_COLUMNS("info-all-slaves-query", row, 4);
     } catch(const PDNSException &e) {
-      g_log<<Logger::Warning<<e.reason<<endl;
+      if (!loggedAssertRowColumns) {
+        g_log<<Logger::Warning<<e.reason<<endl;
+      }
+      loggedAssertRowColumns = true;
       continue;
     }
 
--- a/pdns/backends/gsql/gsqlbackend.cc
+++ b/pdns/backends/gsql/gsqlbackend.cc
@@ -1304,8 +1304,13 @@ void GSQLBackend::getAllDomains(vector<DomainInfo> *domains, bool include_disabl
       if (!row[4].empty()) {
         vector<string> masters;
         stringtok(masters, row[4], " ,\t");
-        for(const auto& m : masters)
-          di.masters.emplace_back(m, 53);
+        for(const auto& m : masters) {
+          try {
+            di.masters.emplace_back(m, 53);
+          } catch(const PDNSException &e) {
+            throw PDNSException("Could not parse master address (" + m + ") for zone '" + di.zone.toLogString() + "': " + e.reason);
+          }
+        }
       }
 
       SOAData sd;
--- a/pdns/backends/gsql/gsqlbackend.cc
+++ b/pdns/backends/gsql/gsqlbackend.cc
@@ -1312,7 +1312,7 @@ void GSQLBackend::getAllDomains(vector<DomainInfo> *domains, bool include_disabl
           try {
             di.masters.emplace_back(m, 53);
           } catch(const PDNSException &e) {
-            throw PDNSException("Could not parse master address (" + m + ") for zone '" + di.zone.toLogString() + "': " + e.reason);
+            g_log<<Logger::Warning<<"Could not parse master address ("<<m<<") for zone '"<<di.zone<<"': "<<e.reason;
           }
         }
       }
--- a/pdns/backends/gsql/gsqlbackend.cc
+++ b/pdns/backends/gsql/gsqlbackend.cc
@@ -341,23 +341,51 @@ void GSQLBackend::getUnfreshSlaveInfos(vector<DomainInfo> *unfreshDomains)
 
   for(const auto& row : d_result) { // id,name,master,last_check
     DomainInfo sd;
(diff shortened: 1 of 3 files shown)


pdns/backends/gsql/gsqlbackend.cc
```diff
@@ -341,23 +341,50 @@ void GSQLBackend::getUnfreshSlaveInfos(vector<DomainInfo> *unfreshDomains)
 
   for(const auto& row : d_result) { // id,name,master,last_check
     DomainInfo sd;
-    ASSERT_ROW_COLUMNS("info-all-slaves-query", row, 4);
+    try {
+      ASSERT_ROW_COLUMNS("info-all-slaves-query", row, 4);
```
mind04: Why is this assert changed in a non fatal, very verbose one?

pdns/backends/gsql/gsqlbackend.cc
```diff
@@ -341,23 +341,50 @@ void GSQLBackend::getUnfreshSlaveInfos(vector<DomainInfo> *unfreshDomains)
...
+    } catch(const PDNSException &e) {
+      g_log<<Logger::Warning<<e.reason<<endl;
+      continue;
+    }
+
+    try {
+      sd.zone = DNSName(row[1]);
+    } catch(const PDNSException &e) {
```
mind04: DNSName is not throwing PDNSException

pdns/backends/gsql/gsqlbackend.cc
```diff
@@ -1265,12 +1292,24 @@ void GSQLBackend::getAllDomains(vector<DomainInfo> *domains, bool include_disabl
       } catch (...) {
         continue;
       }
+
+      if (pdns_iequals(row[3], "MASTER"))
+        di.kind = DomainInfo::Master;
+      else if (pdns_iequals(row[3], "SLAVE"))
+        di.kind = DomainInfo::Slave;
+      else
+        di.kind = DomainInfo::Native;
```
mind04: Please add some logging if the type is not equal to 'NATIVE'

pdns/backends/gsql/gsqlbackend.cc
```diff
@@ -1265,12 +1292,24 @@ void GSQLBackend::getAllDomains(vector<DomainInfo> *domains, bool include_disabl
...
       if (!row[4].empty()) {
         vector<string> masters;
         stringtok(masters, row[4], " ,\t");
-        for(const auto& m : masters)
-          di.masters.emplace_back(m, 53);
+        for(const auto& m : masters) {
+          try {
+            di.masters.emplace_back(m, 53);
+          } catch(const PDNSException &e) {
+            throw PDNSException("Could not parse master address (" + m + ") for zone '" + di.zone.toLogString() + "': " + e.reason);
```
mind04: I'm prety sure this throw will lead to new issues since a single broken master ip will terminate getAllDomans(). 
For example "pdnsutil rectify-all-zones" in a cron, will stop working if there is one broken master ip in your database.

pdns/backends/gsql/gsqlbackend.cc
pieterlexis: before, this would lead to the whole `getUnfreshSlaveInfos` failing. _in theory_ other could be good still.

pdns/backends/gsql/gsqlbackend.cc
pieterlexis: `rectify-all-zones` works, but `list-all-zones` and `check-all-zones` break indeed. Not throwing will make `pdnsutil` swallow the exception (but `check-all-zones` will report it). Fixed

pdns/backends/gsql/gsqlbackend.cc
mind04: In my opinion this is not the right place to fix this. Catch the exception and add some logging in getUnfreshSlaveInfos() is a better approach. When this assert is triggered it is likely for all rows in the result.
With a million domains this will result in a million lines of logging for every slave freshness check.

pieterlexis: I'll make sure it logs only once but keep the current semantics of "keep trying all other domains regardless"

pdns/backends/gsql/gsqlbackend.cc
```diff
@@ -339,25 +339,56 @@ void GSQLBackend::getUnfreshSlaveInfos(vector<DomainInfo> *unfreshDomains)
...
+    } catch(const std::runtime_error &e) {
+      g_log<<Logger::Warning<<"Domain name '"<<row[1]<<"' is not a valid DNS name: "<<e.what()<<endl;
+      continue;
+    }
+
     try {
       sd.id=pdns_stou(row[0]);
-      sd.zone= DNSName(row[1]);
+    } catch (const std::exception &e) {
+      g_log<<Logger::Warning<<"Could not convert id ("<<row[0]<<") for domain '"<<sd.zone<<"' into an integer: "<<e.what()<<endl;
```
zeha: missing continue; here?

pdns/backends/gsql/gsqlbackend.cc
zeha: now it will always log this even if the db already says "NATIVE"?

pdns/backends/gsql/gsqlbackend.cc
pieterlexis: yes....

PR #7571, opened 2019-03-13
rec: Add a distribution-pipe-buffer-size setting

rgacogne:
### Short description

Allows setting the size in bytes of the internal buffer of the pipe used by the distributor to pass incoming queries to a worker thread.
Requires support for `F_SETPIPE_SZ` which is present in Linux since 2.6.35. The actual size might be rounded up to a multiple of a page size. 0 means that the OS default size is used.
A large buffer might allow the recursor to deal with very short-lived load spikes during which a worker thread gets overloaded, but it will be at the cost of an increased latency.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/misc.cc
+++ b/pdns/misc.cc
@@ -1485,3 +1485,36 @@ std::vector<ComboAddress> getResolvers(const std::string& resolvConfPath)
 
   return results;
 }
+
+size_t getPipeBufferSize(int fd)
+{
+#ifdef F_GETPIPE_SZ
+  int res = fcntl(fd, F_GETPIPE_SZ);
+  if (res == -1) {
+    return 0;
+  }
+  return res;
+#else
+  errno = ENOSYS;
+  return 0;
+#endif /* F_GETPIPE_SZ */
+}
+
+bool setPipeBufferSize(int fd, size_t size)
+{
+#ifdef F_SETPIPE_SZ
+  if (size > std::numeric_limits<int>::max()) {
+    errno = EINVAL;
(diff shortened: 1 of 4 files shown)


rgacogne:
Rebased to fix a conflict.

rgacogne:
Rebased to fix a conflict.

pdns/misc.cc
```diff
@@ -1485,3 +1485,36 @@ std::vector<ComboAddress> getResolvers(const std::string& resolvConfPath)
...
+    return 0;
+  }
+  return res;
+#else
+  errno = ENOSYS;
+  return 0;
+#endif /* F_GETPIPE_SZ */
+}
+
+bool setPipeBufferSize(int fd, size_t size)
```
omoerbeek: Why a `size_t` ? 
`argsvMap.asNum()` returns an `int`, and the `fcntl` takes an `int` as well.

rgacogne: I used a `size_t` because a negative value makes no sense here. `asNum()` and `fcntl()` are generic interfaces, but IMHO that does not mean we can't use a more correct type.

omoerbeek: I'm happy with the unsigned part of size_t, but negative numbers will be converted to very large values by the int to size_t conversion. We probably need an asSize() or similar that rejects negative values.

rgacogne: That would be awesome, yes!

PR #7574, opened 2019-03-13
rec: Add a new ecs-minimum-ttl-override setting

rgacogne:
### Short description

To be able to artificially raise the TTLs of records in the ANSWER section of ECS-specific answers to be at least this long.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/pdns/rec_channel_rec.cc
+++ b/pdns/rec_channel_rec.cc
@@ -696,8 +696,13 @@ static string setMinimumTTL(T begin, T end)
 {
   if(end-begin != 1)
     return "Need to supply new minimum TTL number\n";
-  SyncRes::s_minimumTTL = pdns_stou(*begin);
-  return "New minimum TTL: " + std::to_string(SyncRes::s_minimumTTL) + "\n";
+  try {
+    SyncRes::s_minimumTTL = pdns_stou(*begin);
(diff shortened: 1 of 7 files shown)


pdns/rec_channel_rec.cc
```diff
@@ -668,12 +668,21 @@ static string getTAs()
...
   SyncRes::s_minimumTTL = pdns_stou(*begin);
   return "New minimum TTL: " + std::to_string(SyncRes::s_minimumTTL) + "\n";
 }
 
+template<typename T>
+string setMinimumECSTTL(T begin, T end)
+{
+  if(end-begin != 1)
+    return "Need to supply new ECS minimum TTL number\n";
+  SyncRes::s_minimumECSTTL = pdns_stou(*begin);
```
pieterlexis: `pdns_stou` can throw, which would abort the process I think.

rgacogne:
Rebased to fix a conflict, plus two commits to deal with the issue reported in @pieterlexis' review.

pdns/rec_channel_rec.cc
rgacogne: Nice catch!

PR #7584, opened 2019-03-15
GSQL: Log more data in error messages

pieterlexis:
### Short description
This could help zooming into the source of the error.

### Checklist

I have:
- [x] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/backends/gsql/gsqlbackend.cc
+++ b/pdns/backends/gsql/gsqlbackend.cc
@@ -1140,7 +1140,7 @@ bool GSQLBackend::superMasterBackend(const string &ip, const DNSName &domain, co
         reset();
     }
     catch (SSqlException &e) {
-      throw PDNSException("GSQLBackend unable to search for a domain: "+e.txtReason());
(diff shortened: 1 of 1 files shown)


pdns/backends/gsql/gsqlbackend.cc
```diff
@@ -797,7 +797,7 @@ bool GSQLBackend::getTSIGKey(const DNSName& name, DNSName* algorithm, string* co
     d_getTSIGKeyQuery_stmt->reset();
   }
   catch (SSqlException &e) {
-    throw PDNSException("GSQLBackend unable to retrieve named TSIG key: "+e.txtReason());
+    throw PDNSException("GSQLBackend unable to retrieve TSIG key with name '" + name.toString() + "': "+e.txtReason());
```
mind04: Why is toString() used here (and in some other places) instead of toLogString() ?

pieterlexis: I believed we logged TSIG key names with a dot in other palces, but a look at the tcpreceiver tells me we're not

PR #7593, opened 2019-03-18
Update changelog, secpoll and advisories for auth-4.0.7 and auth-4.1.7

aerique:


docs/security-advisories/powerdns-advisory-2019-03.rst
```diff
@@ -0,0 +1,33 @@
...
+-  Date: March 18th 2019
+-  Affects: PowerDNS Authoritative up to and including 4.1.6
+-  Not affected: 4.1.7, 4.0.7
+-  Severity: High
+-  Impact: Denial of Service, Information Disclosure, Content spoofing
+-  Exploit: This problem can be triggered via crafted queries
+-  Risk of system compromise: No
+-  Solution: Upgrade to a non-affected version
+
+An issue has been found in PowerDNS Authoritative Server allowing a
```
cmouse: should this mention that it only happens if you use the RESTful urls?

rgacogne: Is it possible to use the HTTP connector with something else than RESTful URLs?

cmouse: Yes. You can configure it to send everything as json to the endpoint directly and not alter the URL at all. 
> You can also use post to change behaviour so that it will send POST request to url/method + url_suffix with parameters=json-formatted-parameters. If you use post and post_json, it will POST url with text/javascript containing JSON formatted RPC request, just like for pipe and unix.

cmouse: To clarify that a bit, the post=1 mode means http://foobar.com/dns/lookup.json, the post=1, post_json=1 means http://foobar.com/dns

rgacogne: So maybe something like:

> An issue has been found in PowerDNS Authoritative Server when the
HTTP remote backend is used in RESTful mode (without post=1 set),
allowing a remote user to cause the HTTP backend to connect to
an attacker-specified host instead of the configured one, via a
crafted DNS query.

PR #7615, opened 2019-03-21
RKEY is missing algorithm field

DNS-Leo:
This RR type is missing a data field.
draft-reid-dnsext-rkey-00 defines the data as flags (0) / protocol (1) / algorithm / key, and for these last two is refering to RFC4034

https://tools.ietf.org/html/draft-reid-dnsext-rkey-00
https://tools.ietf.org/html/rfc4034#section-2.1

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsrecords.cc
+++ b/pdns/dnsrecords.cc
@@ -397,6 +397,7 @@ CDNSKEYRecordContent::CDNSKEYRecordContent() {}
 boilerplate_conv(RKEY, 57, 
                  conv.xfr16BitInt(d_flags); 
                  conv.xfr8BitInt(d_protocol); 
+                 conv.xfr8BitInt(d_algorithm); 
                  conv.xfrBlob(d_key);
                  )
 RKEYRecordContent::RKEYRecordContent() {}


Habbie:
Alternative suggestion: drop support for the type.

DNS-Leo:
If you don't support it correctly then you can't use it for what it was meant.
If you drop support then it can never be used anymore.
Clearly nobody else noticed the bug - at least didn't reported it, so I estimate the demand among powerdns users is close to nihil.
Anyway, it's JFYI + simple solution; up to you to decide, but AFAIK the record is not obsolete.

DNS-Leo:
I'm not such an avid Github user - am I now suppose to close this request?

Habbie:
> AFAIK the record is not obsolete.

There's nothing to obsolete - it was never published as RFC!

But the big question here is, why/how did you find out? Did you try to use RKEY for something?

DNS-Leo:
Before making a suggestion to draft-muks-dnsop-dns-catalog-zones I verified how well (or bad) NINFO records (TYPE56) are supported among common DNS software, but found out that it was missing in pdns. My adventually planned intended use was exactly like it was designated for.
As it's -like SPF- is just another TXT clone, I had the illusion it wasn't too hard to add it, did some fruitless attempts and while doing so I noticed the error at the next RR type (57) of which I always -mistakenly- thought to be a TXT-clone also.

PR #7622, opened 2019-03-23
Change all occurrences of string to std::string

shane-kerr:
I noticed that there were both `string` and `std::string` in the file that I am working on. This converts them all to `std::string`.

### Short description
No issues, just some code hygiene improvement by making the style consistent. 

### Checklist
I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdist-lua-actions.cc
+++ b/pdns/dnsdist-lua-actions.cc
@@ -35,11 +35,11 @@
 class DropAction : public DNSAction
 {
 public:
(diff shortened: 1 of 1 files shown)


rgacogne:
Thank you for this PR! I'm bit torn about it because while I like having a consistent style, merging this now will cause a lot of conflicts with existing pull requests, so I think we should at least wait a bit before merging it.

PR #7572, opened 2019-03-13
rec: implement a configurable ECS cache limit

rgacogne:
### Short description

This PR implements a configurable ECS cache limit so responses with an ECS scope more specific than a certain threshold are not inserted into the records cache at all.

### Checklist

I have:
- [x] read the [CONTRIBUTING.md](https://github.com/PowerDNS/pdns/blob/master/CONTRIBUTING.md) document
- [x] compiled this code
- [x] tested this code
- [x] included documentation (including possible behaviour changes)
- [ ] documented the code
- [x] added or modified regression test(s)
- [x] added or modified unit test(s)

PR #7631, opened 2019-03-27
Rec ecs cache limit with ttl

omoerbeek:
### Short description

This is further work based on https://github.com/PowerDNS/pdns/pull/7572
It introduces a new parameter to control caching of ECS specific answers: ecs-cache-limit-ttl
Ineffective by default. If set, an ECS specific answer must have a TTL >= the given limit to be cached.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [X] documented the code

### DIFF
--- a/pdns/recursor_cache.cc
+++ b/pdns/recursor_cache.cc
@@ -236,7 +236,7 @@ int32_t MemRecursorCache::get(time_t now, const DNSName &qname, const QType& qt,
   return -1;
 }
 
-void MemRecursorCache::replace(time_t now, const DNSName &qname, const QType& qt, const vector<DNSRecord>& content, const vector<shared_ptr<RRSIGRecordContent>>& signatures, const std::vector<std::shared_ptr<DNSRecord>>& authorityRecs, bool auth, boost::optional<Netmask> ednsmask, time_t minTTD, vState state)
+void MemRecursorCache::replace(time_t now, const DNSName &qname, const QType& qt, const vector<DNSRecord>& content, const vector<shared_ptr<RRSIGRecordContent>>& signatures, const std::vector<std::shared_ptr<DNSRecord>>& authorityRecs, bool auth, boost::optional<Netmask> ednsmask, vState state)
 {
   d_cachecachevalid = false;
   //  cerr<<"Replacing "<<qname<<" for "<< (ednsmask ? ednsmask->toString() : "everyone") << endl;
@@ -305,11 +305,9 @@ void MemRecursorCache::replace(time_t now, const DNSName &qname, const QType& qt
   for(const auto i : content) {
     /* Yes, we have altered the d_ttl value by adding time(nullptr) to it
        prior to calling this function, so the TTL actually holds a TTD. */
-    if (minTTD == 0 || static_cast<time_t>(i.d_ttl) >= minTTD) {
-      ce.d_ttd = min(maxTTD, static_cast<time_t>(i.d_ttl));   // XXX this does weird things if TTLs differ in the set
-      //    cerr<<"To store: "<<i.d_content->getZoneRepresentation()<<" with ttl/ttd "<<i.d_ttl<<", capped at: "<<maxTTD<<endl;
-      ce.d_records.push_back(i.d_content);
-    }
+    ce.d_ttd=min(maxTTD, static_cast<time_t>(i.d_ttl));   // XXX this does weird things if TTLs differ in the set
+    //    cerr<<"To store: "<<i.d_content->getZoneRepresentation()<<" with ttl/ttd "<<i.d_ttl<<", capped at: "<<maxTTD<<endl;
+    ce.d_records.push_back(i.d_content);
   }
 
   if (!isNew) {
--- a/pdns/recursor_cache.cc
+++ b/pdns/recursor_cache.cc
@@ -236,7 +236,7 @@ int32_t MemRecursorCache::get(time_t now, const DNSName &qname, const QType& qt,
   return -1;
 }
 
-void MemRecursorCache::replace(time_t now, const DNSName &qname, const QType& qt, const vector<DNSRecord>& content, const vector<shared_ptr<RRSIGRecordContent>>& signatures, const std::vector<std::shared_ptr<DNSRecord>>& authorityRecs, bool auth, boost::optional<Netmask> ednsmask, vState state)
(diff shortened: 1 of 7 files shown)


pdns/recursor_cache.cc
```diff
@@ -305,9 +305,11 @@ void MemRecursorCache::replace(time_t now, const DNSName &qname, const QType& qt
   for(const auto i : content) {
     /* Yes, we have altered the d_ttl value by adding time(nullptr) to it
        prior to calling this function, so the TTL actually holds a TTD. */
-    ce.d_ttd=min(maxTTD, static_cast<time_t>(i.d_ttl));   // XXX this does weird things if TTLs differ in the set
-    //    cerr<<"To store: "<<i.d_content->getZoneRepresentation()<<" with ttl/ttd "<<i.d_ttl<<", capped at: "<<maxTTD<<endl;
-    ce.d_records.push_back(i.d_content);
+    if (minTTD == 0 || static_cast<time_t>(i.d_ttl) >= minTTD) {
```
rgacogne: I don't think this is the right place to do this check. Unless I'm mistaken, we could end up adding a new entry to the cache without any records, which is wrong.

pdns/syncres.hh
```diff
@@ -720,6 +720,7 @@ public:
   static uint8_t s_ecsipv6limit;
   static uint8_t s_ecsipv4cachelimit;
   static uint8_t s_ecsipv6cachelimit;
+  static unsigned int s_ecscachelimitttl;
```
rgacogne: Just a nice but it would be nice to move the new field with the other `unsigned int` ones.

omoerbeek: > It looks like this setting prevents the insertion of an ECS-specific entry into the cache regardless of the actual scope. I think what we want is to prevent the entry from entering the cache if the scope is greater or equal to `ecs-ipv{4,6}-cache-bits` **AND** the TTL is lower than `ecs-cache-limit-ttl`, and unless I'm mistaken this PR does prevent it if the scope is greater or equal to `ecs-ipv{4,6}-cache-bits` **OR** the TTL is lower than `ecs-cache-limit-ttl`.

Yes, alreay determined myself that the condition is not correct, will revisit.

pdns/recursor_cache.cc
omoerbeek: Hmm, thing get complicated because it is not a given all entries have the same TTL.
A scan of the min or max TTL of all records could probaly work... have to think about this after coffee.

omoerbeek: New code in update to thi PR.

pdns/syncres.cc
```diff
@@ -2416,7 +2420,37 @@ RCode::rcodes_ SyncRes::updateCacheFromRecords(unsigned int depth, LWResult& lwr
...
+              // Case: many bits and ttlIsSmall
+              doCache = false;
+            }
+          } else {
+              // Case: few mask bits
+              doCache = true;
+          }
+        } else {
+          // no applicable TTL limit, scope determines cacheability
+          doCache = !manyMaskBits;
```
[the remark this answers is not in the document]
omoerbeek: Initially I wanted to keep "existing" behavior (the original PR) the same. But I agree this probably should be true.

pdns/recursordist/docs/settings.rst
```diff
@@ -399,6 +421,17 @@ This setting artificially raises the TTLs of records in the ANSWER section of EC
...
+.. _setting-ecs-cache-limit-ttl:
+
+``ecs-cache-limit-ttl``
+-----------------------
+.. versionadded:: 4.1.12
+
+-  Integer
+-  Default: 0 (disabled)
+
+The minumum TTL for an ECS-specific answer to be inserted into the query cache.
```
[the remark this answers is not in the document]
omoerbeek: will do

pdns/recursordist/docs/settings.rst
```diff
@@ -399,6 +423,18 @@ This setting artificially raises the TTLs of records in the ANSWER section of EC
...
+.. _setting-ecs-cache-limit-ttl:
+
+``ecs-cache-limit-ttl``
+-----------------------
+.. versionadded:: 4.1.12
+
+-  Integer
+-  Default: 0 (disabled)
+
+The minumum TTL for an ECS-specific answer to be inserted into the query cache. This condition applies in conjunction with ``ecs-ipv4-cache-bits`` or ``ecs-ipv6-cache-bits``.
```
```suggestion
The minimum TTL for an ECS-specific answer to be inserted into the query cache. This condition applies in conjunction with ``ecs-ipv4-cache-bits`` or ``ecs-ipv6-cache-bits``.
```

pdns/recursordist/test-syncres_cc.cc
```diff
@@ -2099,6 +2104,204 @@ BOOST_AUTO_TEST_CASE(test_skip_negcache_for_variable_response) {
...
+      return 1;
+    });
+
+  const time_t now = sr->getNow().tv_sec;
+  vector<DNSRecord> ret;
+  int res = sr->beginResolve(target, QType(QType::A), QClass::IN, ret);
+  BOOST_CHECK_EQUAL(res, RCode::NoError);
+  BOOST_CHECK_EQUAL(ret.size(), 1);
+
+  /* should have been cached because /24 is more specific than /16 but TTL limit is nof efective */
```
rgacogne:
```suggestion
  /* should have been cached because /24 is more specific than /16 but TTL limit is not effective */
```

pdns/syncres.cc
```diff
@@ -2416,7 +2420,32 @@ RCode::rcodes_ SyncRes::updateCacheFromRecords(unsigned int depth, LWResult& lwr
        - NS, A and AAAA (used for infra queries)
     */
     if (i->first.type != QType::NSEC3 && (i->first.type == QType::DS || i->first.type == QType::NS || i->first.type == QType::A || i->first.type == QType::AAAA || isAA || wasForwardRecurse)) {
-      t_RC->replace(d_now.tv_sec, i->first.name, QType(i->first.type), i->second.records, i->second.signatures, authorityRecs, i->first.type == QType::DS ? true : isAA, i->first.place == DNSResourceRecord::ANSWER ? ednsmask : boost::none, recordState);
+
+      bool doCache = i->first.place != DNSResourceRecord::ANSWER || !ednsmask;
+      // if ednsmask is relevant, we do not want to cache if the scope > ecslimit and TTL < limitttl
+      if (!doCache && ednsmask) {
```
rgacogne: I was a bit confused by the fact that we might still cache even though the initial value of `doCache` is `false`. How about something like that instead?
```
      bool doCache = true;
      if (i->first.place ==  DNSResourceRecord::ANSWER && ednsmask) {
```

PR #7632, opened 2019-03-27
fix of problems in oraclebackend.hh

DNS-Leo:
With this changes installation went well - on condition you do `./configure --with-modules="goracle oracle"`

This pull contains the modification to allow an (errorless) install of a (working) Oracle backend.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/modules/oraclebackend/oraclebackend.hh
+++ b/modules/oraclebackend/oraclebackend.hh
@@ -75,7 +75,7 @@ public:
                                       DNSName& before,
                                       DNSName& after) override;
   bool get(DNSResourceRecord &rr) override;
-  vector<string> getDomainMasters(const DNSName& domain, int zoneId) override;
+  vector<string> getDomainMasters(const DNSName& domain, int zoneId);
   bool getDomainInfo(const DNSName& domain, DomainInfo &di) override;
   void alsoNotifies(const DNSName& domain, set<string> *addrs) override;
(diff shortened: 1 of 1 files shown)


modules/oraclebackend/oraclebackend.hh
```diff
@@ -84,7 +84,7 @@ public:
   void setNotified(uint32_t zoneId, uint32_t serial) override;
   bool list(const DNSName& domain, int zoneId, bool include_disabled=false) override;
   bool startTransaction(const DNSName& domain, int zoneId) override;
-  bool feedRecord(const DNSResourceRecord &rr, const DNSName ordername) override;
+  bool feedRecord(const DNSResourceRecord &rr, const DNSName ordername);
```
rgacogne: This doesn't seem right, contrary to the two other ones `feedRecord()` is really inherited and should match the declaration of the base class, unless I'm missing something.

DNS-Leo:
AFAIK it didn't compile (or run) without this.
After the weekend I will double check and provide feedback as soon as I know more.

rgacogne:
> AFAIK it didn't compile (or run) without this.

I agree the current code is broken because the definition does not match the one in the base class, but removing the `override` only hides the error in that case. It does not fix it.

PR #7634, opened 2019-03-27
rec-4.1.x: Use a bounded load-balancing algo to distribute queries

rgacogne:
### Short description

Backport of https://github.com/PowerDNS/pdns/pull/7507 to rel/rec-4.1.x without exporting the new metric over SNMP, otherwise the OID would change between 4.1.x and master.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/pdns_recursor.cc
+++ b/pdns/pdns_recursor.cc
@@ -3200,6 +3200,10 @@ static int serviceMain(int argc, char*argv[])
   g_statisticsInterval = ::arg().asNum("statistics-interval");
 
   s_balancingFactor = ::arg().asDouble("distribution-load-factor");
+  if (s_balancingFactor != 0.0 && s_balancingFactor < 1.0) {
+    s_balancingFactor = 0.0;
+    L<<Logger::Warning<<"Asked to run with a distribution-load-factor below 1.0, disabling it instead"<<endl;
+  }
 
 #ifdef SO_REUSEPORT
   g_reusePort = ::arg().mustDo("reuseport");
--- a/pdns/pdns_recursor.cc
+++ b/pdns/pdns_recursor.cc
@@ -2398,7 +2398,7 @@ static unsigned int selectWorker(unsigned int hash)
   // cerr<<"total load is "<<currentLoad<<", number of workers is "<<g_numWorkerThreads<<", target load is "<<targetLoad<<endl;
 
   unsigned int worker = hash % g_numWorkerThreads;
-  /* at least one server has to be below the average load */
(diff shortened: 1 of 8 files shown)


pdns/pdns_recursor.cc
```diff
@@ -2348,10 +2365,53 @@ static bool trySendingQueryToWorker(unsigned int target, ThreadMSG* tmsg)
...
+
   return true;
 }
 
-// This function is only called by the distributor thread, when pdns-distributes-queries is set
+static unsigned int getWorkerLoad(size_t workerIdx)
+{
+  const auto mt = s_threadInfos[/* skip distributor */ 1 + workerIdx].mt;
+  if (mt != nullptr) {
+    return mt->numProcesses();
```
ahupowerdns: So is this safe? Doesn't this need a lock to protect d_threads in MTasker? Also, I think this would suck, perhaps there is a better way to do this. Also this assumes the load is proportional to the number of processes and not to the number of queries per second - although I am also not sure how we'd measure that.

ahupowerdns: We might have to do a numThreads variable in MTasker and keep it updated.

rgacogne: It would be safer to keep a separate variable, updated, yes. I'm pretty sure that's how it's implemented in practice but it's not guaranteed by the standard. I'll update this PR.

rgacogne: I agree estimating the load this ways is far from being perfect, but it was the cheapest estimation I could find. We could keep track of the number of queries distributed to each thread, or even the ones actually in the distribution pipe, but it would come at a cost that it's not easy to estimate.

pdns/pdns_recursor.cc
```diff
@@ -2348,10 +2365,53 @@ static bool trySendingQueryToWorker(unsigned int target, ThreadMSG* tmsg)
...
+    load[idx] = getWorkerLoad(idx);
+    currentLoad += load[idx];
+    // cerr<<"load for worker "<<idx<<" is "<<load[idx]<<endl;
+  }
+
+  double targetLoad = (currentLoad / g_numWorkerThreads) * s_balancingFactor;
+  // cerr<<"total load is "<<currentLoad<<", number of workers is "<<g_numWorkerThreads<<", target load is "<<targetLoad<<endl;
+
+  unsigned int worker = hash % g_numWorkerThreads;
+  /* at least one server has to be below the average load */
```
omoerbeek: "on or below"

PR #7636, opened 2019-03-28
Action for extended error codes

shane-kerr:
### Short description
This PR implements the Extended DNS Errors draft:

https://tools.ietf.org/html/draft-ietf-dnsop-extended-error-05

It was started at the IETF 104 hackathon and finished up in the week of the IETF.

I looked through the documented error codes in the draft and think that only three really apply to proxies:

REFUSED Extended DNS Error Code 2 - Prohibited
NXDOMAIN Extended DNS Error Code 1 - Blocked
NXDOMAIN Extended DNS Error Code 2 - Censored

Since adding an ACL to `dnsdist` causes it to drop packets on the floor, none of these make sense to happen "automagically" somehow.

Instead, this PR takes the approach of enhancing the existing `ERCodeAction` to provide optional additional parameters which can be used to deliver extended error code information, for example:
```
setLocal("[::1]:5321")
newServer("9.9.9.9")
addAction(RegexRule("."), ERCodeAction(3, 1, "DENIED ✋"))
```
This will return `NXDomain` with the extended code 1 (blocked), while providing a helpful error message.

I used a few C-isms while poking around the packet bytes. There are probably more verbose C++ ways to do this which - while not actually being any safer or faster - are at least really long and hard to read, therefore preferred by C++ developers. 😉

I also note that there did not seem to be straightforward ways to manipulate packet contents (for example, adding OPT RR and so on). I tried to mimic the approach taken in the ECS stuff, although that has a lot more work involved than the extended error codes. Probably a refactoring to handle OPT record processing in a more elegant way would be worth it, but outside of the scope of the hackathon. 

The  `dnsdist-lua-actions.cc` file has a lot of changes because of the `string` to `std::string` conversion. I can clean these up if desired, but my guess is that this PR will probably need other changes anyway so I've left them in. 

While the existing tests work, I did not bother to add any new ones. I guess there would be something like 10 additional cases to cover all the combinations of ways to do things correctly and broken with such an option.

Finally, it might be also be useful to provide a way to read the extended error code provided by the backends, and allow a rule (or Lua) to act on them. If the extended error code draft is still a work in progress by the next IETF hackathon, maybe I'll give that a shot then.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dns.cc
+++ b/pdns/dns.cc
@@ -69,6 +69,69 @@ std::string ERCode::to_s(uint8_t rcode) {
   return RCode::rcodes_s[rcode];
 }
 
+
+static std::vector<std::string> noerror_info_codes = boost::assign::list_of
+  ("")
+  ("Unsupported DNSKEY Algorithm")
+  ("Unsupported DS Algorithm")
+  ("Stale Answer")
+  ("Forged answer")
+  ("DNSSEC Indeterminate")
+;
+
+static std::vector<std::string> formerr_info_codes = boost::assign::list_of
+  ("")
+;
+
+static std::vector<std::string> servfail_info_codes = boost::assign::list_of
+  ("")
+  ("DNSSEC Bogus")
+  ("Signature Expired")
+  ("Signature Not Yet Valid")
+  ("DNSKEY missing")
+  ("RRSIGs missing")
+  ("No Zone Key Bit Set")
+  ("No Reachable Authority")
+  ("NSEC Missing")
+  ("Cached Error")
+  ("Not Ready")
+;
+
+static std::vector<std::string> nxdomain_info_codes = boost::assign::list_of
+  ("")
+  ("Blocked")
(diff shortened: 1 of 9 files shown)


pieterlexis:
@rgacogne is this still a relevant PR? If not, please close.

rgacogne:
The final version is quite different so we are unlikely to merge this PR, but I would like to keep it open as long as we don't have that feature.

rgacogne:
It doesn't really make sense to keep this pull request based on an early draft around so I opened a new feature request instead: https://github.com/PowerDNS/pdns/issues/12572
Thanks a lot for the pull request, and sorry we did not react sooner!

shane-kerr:
No worries! This was hackathon work, so meant to vet the draft not necessarily for something to go into production code. Happy to see the feature request issue. :+1:

Issue #7625, opened 2019-03-26
serving root-zone from mysql backend incomplete


miesi:
Program: Authoritative
Issue type: Bug report

### Short description

queries for DS co.uk  are answered differently from what bind does with the same zone file.

### Environment
 - Operating system:  CentOS 7 x86_64
 - Software version: 4.1.8, MariaDB 10.3
 - Software source: rpm compiled by mind04

### Steps to reproduce

Setup a powerdns with mysql backend and slave support enabled. Add the following entry to domains table and wait for the root-zone to transfer.

name | master | type
--------|-----------|-------
. |192.0.47.132 | SLAVE

### Expected behaviour
```
 dig DS co.uk @f.root-servers.net

; <<>> DiG 9.11.5-P4-RedHat-9.11.5-4.P4.fc29 <<>> DS co.uk @f.root-servers.net
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 50704
;; flags: qr rd; QUERY: 1, ANSWER: 0, AUTHORITY: 8, ADDITIONAL: 14
;; WARNING: recursion requested but not available

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 1472
;; QUESTION SECTION:
;co.uk.				IN	DS

;; AUTHORITY SECTION:
uk.			172800	IN	NS	nsa.nic.uk.
uk.			172800	IN	NS	nsb.nic.uk.
uk.			172800	IN	NS	nsc.nic.uk.
uk.			172800	IN	NS	nsd.nic.uk.
uk.			172800	IN	NS	dns1.nic.uk.
uk.			172800	IN	NS	dns2.nic.uk.
uk.			172800	IN	NS	dns3.nic.uk.
uk.			172800	IN	NS	dns4.nic.uk.

;; ADDITIONAL SECTION:
nsa.nic.uk.		172800	IN	A	156.154.100.3
nsa.nic.uk.		172800	IN	AAAA	2001:502:ad09::3
nsb.nic.uk.		172800	IN	A	156.154.101.3
nsc.nic.uk.		172800	IN	A	156.154.102.3
nsd.nic.uk.		172800	IN	A	156.154.103.3
dns1.nic.uk.		172800	IN	A	213.248.216.1
dns1.nic.uk.		172800	IN	AAAA	2a01:618:400::1
dns2.nic.uk.		172800	IN	A	103.49.80.1
[...]

PR #7637, opened 2019-03-28
auth: fix referral response for DS queries

mind04:
### Short description
Glue and ents derived from glue did confuse the DS referral logic.

Fixes: #7625

### Checklist
I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/packethandler.cc
+++ b/pdns/packethandler.cc
@@ -1455,6 +1455,9 @@ DNSPacket *PacketHandler::doQuestion(DNSPacket *p)
       bool doReferral = true;
       if(d_dk.doesDNSSEC()) {
         for(auto& loopRR: rrset) {
+          // In a dnssec capable backend auth=true means, there is no delagation at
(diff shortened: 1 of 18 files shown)


pdns/packethandler.cc
```diff
@@ -1448,19 +1448,38 @@ DNSPacket *PacketHandler::doQuestion(DNSPacket *p)
...
-      DLOG(g_log<<"checking if qtype is DS"<<endl);
-      if(p->qtype.getCode() == QType::DS)
-      {
+
+    // referral for DS query
+    if(p->qtype.getCode() == QType::DS) {
+      DLOG(g_log<<"Qtype is DS"<<endl);
+      bool doReferral = true;
+      if(d_dk.doesDNSSEC()) {
+        for(auto& loopRR: rrset) {
```
rgacogne: It might be nice to have a comment explaining why we look for `auth=true` records in the DNSSEC case and for `DS` records only otherwise, as it's not obvious to me right away.

PR #7644, opened 2019-03-31
remotebackend: fix SOA in unittests

zeha:
Otherwise check-zone and NS-queries fail.

Noticed while verifying a remotebackend patch in Debian against 4.1.5. Not tested on master.

### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code


rgacogne:
It looks like the pipe connector tests do not like this change :-/

PR #7641, opened 2019-03-30
DNSSEC docs: Improve wording in "BIND-mode operation"

gertvdijk:
### Short description

Documentation only - slightly improved the text as the title says.

Also:
* Mention that not just keys are part of this database.
* Link to the pdns.conf setting, that includes a warning with #7640.

### Checklist

I have:
- [x] read the [CONTRIBUTING.md](https://github.com/PowerDNS/pdns/blob/master/CONTRIBUTING.md) document
- [x] compiled this code
  (yes built the docs, did not introduce new Sphinx warnings)
- [x] tested this code
  (the link works)
- [x] included documentation (including possible behaviour changes)
- [x] documented the code
  n/a
- [x] added or modified regression test(s)
  n/a
- [x] added or modified unit test(s)
  n/a

PR #7640, opened 2019-03-30
bindbackend docs: Add warning on empty bind-dnssec-db for slave operation

gertvdijk:
### Short description

I fell right into the pitfall of configuring a slave with the Bind backend, serving presigned records, assuming it will serve the RRSIGs just fine, but no, my domain went bogus. This was documented, but
not as clearly as I hoped for, this PR improves the documentation regarding that.

### Checklist

I have:
- [x] read the [CONTRIBUTING.md](https://github.com/PowerDNS/pdns/blob/master/CONTRIBUTING.md) document
  (sorry for the slightly longer first line in commit message)
- [x] compiled this code
  (yes built the docs, did not introduce new Sphinx warnings)
- [x] tested this code
  (the admonition looks fine to me)
- [x] included documentation (including possible behaviour changes)
- [x] documented the code
  n/a
- [x] added or modified regression test(s)
  n/a
- [x] added or modified unit test(s)
  n/a

PR #7629, opened 2019-03-26
dnssec clarifications

jsoref:
### Short description
minor fixes...

### Checklist

I have:
- [x] read the [CONTRIBUTING.md](https://github.com/PowerDNS/pdns/blob/master/CONTRIBUTING.md) document
- [ ] compiled this code
- [ ] tested this code
- [x] included documentation (including possible behaviour changes)
- [ ] documented the code
- [ ] added or modified regression test(s)
- [ ] added or modified unit test(s)

PR #7645, opened 2019-03-31
Various fixes/updates/rewording to documentation

gertvdijk:
### Short description
Documentation-only fixes, mostly OCD/formatting fixes but some additions too while at it. Please see individual commit messages for details - it's too much to list all here...

Offered as one huge PR as [per request](https://twitter.com/PowerDNS_Bert/status/1112123798915035141).

Feel free to cherry-pick individual commits if you don't agree with the why - I tried my best to keep them as independent as possible with the reasoning in the message.

(Includes #7640, #7641, closing those.)

### Checklist

I have:
- [x] compiled this code
  (verified that this does not introduce new Sphinx warnings, tested with `sphinx-build -E -a -n`)
- [x] tested this code
- [x] documented the code


docs/appendices/backend-writers-guide.rst
```diff
@@ -729,22 +729,22 @@ other update/remove functionality at a later stage.
 
 .. cpp:function:: virtual bool DNSBackend::startTransaction(const string &qname, int id)
 
-  See :cpp:func:`above <DNSBackend::beginTransaction>`. Please
+  See :cpp:func:`above <DomainInfo::startTransaction>`. Please
```
rgacogne:
```suggestion
  See :cpp:func:`above <DNSBackend::startTransaction>`. Please
```

docs/appendices/backend-writers-guide.rst
```diff
@@ -729,22 +729,22 @@ other update/remove functionality at a later stage.
...
   note that this function now receives a negative number (-1), which
   indicates that the current zone data should NOT be deleted.
 
 .. cpp:function:: virtual bool DNSBackend::commitTransaction()
 
-  See :cpp:func:`above <DNSBackend::commitTransaction>`.
+  See :cpp:func:`above <DomainInfo::commitTransaction>`.
```
rgacogne: `::commitTransaction()`, `::abortTransaction()` and `::feedRecord()` belong to the `DNSBackend` class.

docs/appendices/backend-writers-guide.rst
gertvdijk: ah okay, but why does is the of the link mention "above" (we're *in* the DNSBackend class) here? pdns-authoritative-ns2

If I apply your suggestion, please note that it links to itself... :confused: 

So the whole link can be removed, imo?

docs/appendices/backend-writers-guide.rst
gertvdijk: (same as other comment)

gertvdijk:
@rgacogne Thanks for the review; but the comment is a bit confusing to me.

docs/appendices/backend-writers-guide.rst
rgacogne: My understanding is that the "see above" link is because we are in the "DNS update support" section, and should point to the description of `::startTransaction()` in the "Read/write slave-capable backends" section. However it seems wrong there too, unless I'm mistaken, ` bool DomainInfo::isMaster()` should be `bool DNSBackend::isMaster()` and so on up to `bool DomainInfo::setFresh()` that should be `bool DNSBackend::setFresh()`.

gertvdijk: Yeah, I think I understand it now - quite a lot of references seem indeed wrong. I think the initial confusion by the author may be caused by the forward declaration of the DNSBackend class before the DomainInfo struct?

https://github.com/PowerDNS/pdns/blob/b4b9cf2d683845bba29933e1cb40cf917fcbadcb/pdns/dnsbackend.hh#L47-L48

But then, `isMaster()` is indeed part of the `DomainInfo` struct, not `DNSBackend`!

https://github.com/PowerDNS/pdns/blob/b4b9cf2d683845bba29933e1cb40cf917fcbadcb/pdns/dnsbackend.hh#L90-L97

So, this gets a bit too hairy to get right in this big line of commits. I suggest to take that commit out and we should open a separate PR for that document. WDYT?

rgacogne: `isMaster()` is indeed part of `DomainInfo` in master since a2dfc9ea78614a9eef514a612117e507d70d4664, but not in 4.1.x, so it's going to be fun to document it properly..

rgacogne: > I suggest to take that commit out and we should open a separate PR for that document. WDYT?

Makes sense to me!

gertvdijk:
@rgacogne 
* Rebased onto latest master (new doc changes merged earlier today)
* Taken out the "Backend writers’ guide" commit - created #7650.

docs/appendices/backend-writers-guide.rst
gertvdijk: (not very fluent with GitHub workflow - I should have marked this as resolved?)
created #7650

gertvdijk:
Rebased on master - fixing conflicts due to merge of #7629.

PR #7651, opened 2019-04-01
Backport Rec ecs cache limit with tt (7631)

omoerbeek:
### Short description

Backport of https://github.com/PowerDNS/pdns/pull/7631

**Watchout**! **Untested**! Syncres unit test does not work since is needs some extra functionality that I did not backport yet.

Additionally, code is not merged yet to master.

### Checklist

I have:
- [X] compiled this code
- [ ] tested this code
- [X] documented the code

### DIFF
--- a/pdns/syncres.cc
+++ b/pdns/syncres.cc
@@ -73,6 +73,8 @@ uint8_t SyncRes::s_ecsipv4limit;
 uint8_t SyncRes::s_ecsipv6limit;
 uint8_t SyncRes::s_ecsipv4cachelimit;
 uint8_t SyncRes::s_ecsipv6cachelimit;
+unsigned int SyncRes::s_ecscachelimitttl;
+
 bool SyncRes::s_doIPv6;
 bool SyncRes::s_nopacketcache;
 bool SyncRes::s_rootNXTrust;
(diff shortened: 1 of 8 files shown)


pdns/syncres.cc
```diff
@@ -2136,8 +2140,32 @@ RCode::rcodes_ SyncRes::updateCacheFromRecords(unsigned int depth, LWResult& lwr
        - denial of existence proofs in wildcard expanded positive responses are stored in authorityRecs
        - denial of existence proofs for negative responses are stored in the negative cache
     */
-    if (i->first.type != QType::NSEC3) {
-      t_RC->replace(d_now.tv_sec, i->first.name, QType(i->first.type), i->second.records, i->second.signatures, authorityRecs, i->first.type == QType::DS ? true : isAA, i->first.place == DNSResourceRecord::ANSWER ? ednsmask : boost::none, recordState);
+    if (i->first.type != QType::NSEC3 && (i->first.type == QType::DS || i->first.type == QType::NS || i->first.type == QType::A || i->first.type == QType::AAAA || isAA || wasForwardRecurse)) {
```
```suggestion
    if (i->first.type != QType::NSEC3) {
```

ahupowerdns:
Looks good to me, thorough work on on a complicated feature.

pdns/syncres.hh
```diff
@@ -639,6 +639,12 @@ public:
     d_skipCNAMECheck = skip;
   }
 
+  void setQuerySource(const ComboAddress& requestor, boost::optional<const EDNSSubnetOpts&> incomingECS) {
```
rgacogne: I'm not sure this is a good idea to create a function in 4.1.x that does something quite different from the one in master. It's clearly very useful when backporting the `SyncRes` unit tests, but I'm worried it might hide the fact that the handling of incoming ECS values is very different in 4.1.x during a backport. I would not worry so much about that normally, but we have been backporting a lot of things to 4.1.x recently..
Perhaps we could have a wrapper in the `SyncRes` unit tests but not make it part of the `SyncRes` class itself to prevent any confusion?

omoerbeek: Yes, better indeed. Will do.

PR #7654, opened 2019-04-02
Update changelog and secpoll for rec-4.1.12.

aerique:


rgacogne:
Note sure about the `Internals` tag, though, but it's just a nit.

Issue #7646, opened 2019-03-31
Authoritative server exits on parse error while looking for NS addresses to notify


gertvdijk:
- Program: Authoritative 
 - Issue type: Bug report 

### Short description

Very similar to #7493 (exit/crash with invalid zone data), but for regular non-SOA records on the BIND backend.

I do not expect a PowerDNS server to completely go down / exit upon a parser error.

### Environment

 - Operating system: Debian Stretch
 - Software version: 4.1.8-1pdns.stretch & 4.2.0~rc1-1pdns.stretch
 - Software source: PowerDNS repository 

### Steps to reproduce

1. Set up a master with the BIND backend.
2. Add a zone with zone file containing an invalid A record (example with quotes as invalid character, probably other cases to trigger this too?), e.g.:

       @       300     A       "1.2.3.4"
3. Increment the serial, reload the zone or await automatic reload.
4. Observe exit of server with log line:

       Zone 'mydomain.tld' (bind-zones/mydomain.tld.db) needs reloading
       Zone 'mydomain.tld' (bind-zones/mydomain.tld.db) reloaded
       1 domain for which we are master needs notifications
       Exiting because communicator thread died with STL error: Parsing record content (try 'pdnsutil check-zone'): while parsing IP address, expected digits at position 0 in '"1.2.3.4"'

### Expected behaviour

Log error/warning about zone, but keep running for other valid zones loaded.

### Actual behaviour

Complete server exiting.

### Other information

I think it's very similar to  #7493, but the bug appears only fixed for SOA record parser errors?

If I have another example error, in an MX record (data, again with quotes that are wrong: `"10 myhost"`), instead of an A record I get another issue on zone transfer (but not a full crash):

    TCP Connection Thread died because of STL error: Parsing record content (try 'pdnsutil check-zone'): expected digits at position 0 in '"10 myhost".mydomain.tld'

<sup>I noticed this when I accidentally added quotes to records' data, not just the ones of type TXT.</sup>

omoerbeek:
Unable to reproduce using 4.1.8 (and master branch) on my development system.
Using 4.1.8 the broken record do produce a log entry when i retrieve them:
```
Apr 03 10:40:36 Exception building answer packet for example1.com/A (Parsing record content (try 'pdnsutil check-zone'): while parsing IP address, expected digits at position 0 in '"1.2.3.4"') sending out servfail
```

I do see an error message on notify:
```
[...]

PR #7662, opened 2019-04-03
Do not exit on exception resolving addresses to notify

omoerbeek:
Fix https://github.com/PowerDNS/pdns/issues/7646. 

It might not be the right place, though, but it prevents fatal exception on
unparseable A (or AAAA) addresss for nameserver addresses needed
to send notifies.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/communicator.hh
+++ b/pdns/communicator.hh
@@ -251,7 +251,7 @@ private:
 class FindNS
 {
 public:
-  vector<string> lookup(const DNSName &name, UeberBackend *b)
+  vector<string> lookup(const DNSName &name, UeberBackend *b, const DNSName& zone)
   {
     vector<string> addresses;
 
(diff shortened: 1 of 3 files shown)


pdns/communicator.hh
```diff
@@ -259,10 +259,25 @@ public:
...
+          }
+          catch (std::exception &e) {
+            g_log << Logger::Error << "Skipping record: " << e.what() << endl;
+            continue;
+          }
+          if (ok) {
+            if (rr.dr.d_type == QType::A || rr.dr.d_type == QType::AAAA)
+              addresses.push_back(rr.dr.d_content->getZoneRepresentation());   // SOL if you have a CNAME for an NS
+          }
+        } while (ok);
```
rgacogne: Isn't `ok` used uninitialized here if the first call to `::get()` threw an exception?

omoerbeek: yes right, i'm not terribly happy with this construct anyway...

PR #7703, opened 2019-04-10
Use a smaller test size to not fill the pipe buffer on systems with small buffer

omoerbeek:
Also, explictly set arguments to avoid order dependencies and make it possible
to run single tests.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/test-distributor_hh.cc
+++ b/pdns/test-distributor_hh.cc
@@ -68,11 +68,17 @@ struct BackendSlow
 
 
 BOOST_AUTO_TEST_CASE(test_distributor_queue) {
+  ::arg().set("overload-queue-length","Maximum queuelength moving to packetcache only")="0";
+  ::arg().set("max-queue-length","Maximum queuelength before considering situation lost")="1000";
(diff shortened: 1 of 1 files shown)


pdns/test-distributor_hh.cc
```diff
@@ -68,11 +68,17 @@ struct BackendSlow
...
+  ::arg().set("queue-limit","Maximum number of milliseconds to queue a query")="1500";
+  S.declare("servfail-packets","Number of times a server-failed packet was sent out");
+  S.declare("timedout-packets", "timedout-packets");
+
   auto d=Distributor<DNSPacket, Question, BackendSlow>::Create(2);
 
   BOOST_CHECK_EXCEPTION( {
     int n;
-    for(n=0; n < 6000; ++n)  {
+    for(n=0; n < 2000; ++n)  {
```
rgacogne: Perhaps a comment stating that the upper bound should be greater than `max-queue-length`?

Issue #7705, opened 2019-04-10
gsql getAllDomainMetadata() does not get all domain metadata


ahupowerdns:
For reasons which probably made sense at the time, the gsql getAllDomainMetadata() call filters out DNSSEC related metadata (PRESIGNED, NSEC3PARAM, NSEC3NARROW). This in turn makes that these items are not shown in the pdnsutil show-zone meta display. It shows them in other ways though.

It also means that b2b-migrate does not migrate these vital settings, leading to broken servers.

A quick scan of the code appears to show no harm would be done by simply removing this filtering from `getAllDomainMetadata()`.

Habbie:
Fixed in #7706

PR #7706, opened 2019-04-10
No longer filter DNSSEC metadata when DNSSEC is enabled in gsql

ahupowerdns:
As noted in #7705, gsql getAllDomainMetadata() did not return all domain
metadata.  This function would always filter DNSSEC related metadata, even
if a backend was configured to support DNSSEC.

### Checklist

I have:
- [X] compiled this code
- [x] tested this code
- [ ] documented the code


ahupowerdns:
@mind04 - what do you think?

PR #7714, opened 2019-04-11
 rec: Fix DNSSEC validation of non-expanded wildcards

rgacogne:
### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code


Habbie:
`unbound-host` chokes on a related problem:
```
*.foo.h4ha.com has address 192.155.81.104 (BOGUS (security failure))
validation failure <*.foo.h4ha.com. A IN>: signature crypto failed from 45.79.4.83 and 52.55.168.70
```

Maybe we can add a test for that?

rgacogne:
> The code works correctly under anything I can throw at it. I have trouble with calling them 'non-expanded wildcards', as they -are- expanded, just onto themselves, but I see how it would be hard to change the code flow to accommodate my feelings on this.

You are right, it's much more correct to call them expanded onto themselves that non-expanded. I'll rewrite that fix because I don't want to introduce more confusion, this topic is already complicated enough.

rgacogne:
> I do note that we do not provide a wildcard proof NSEC to our client for a direct *.example.com query, but I'm not sure this is a problem - after all, just the RRSIG with the minus-1 labelcount proves the existence of the expanded-onto-itself wildcard.

That's an interesting case, because we technically don't have to prove that the exact name does not exist because it does, but I can see how some DNSSEC validators might not like that.

rgacogne:
> Maybe we can add a test for that?

Is it just an answer expanded from the `*.h4ha.com.` wildcard? If so I think `test_dnssec_validation_nsec_wildcard` already test pretty much the same case.

Habbie:
> Is it just an answer expanded from the `*.h4ha.com.` wildcard? If so I think `test_dnssec_validation_nsec_wildcard` already test pretty much the same case.

It is - but like this PR has `isWildcard` to treat a special case, it looks like unbound also has similar code except it gets it wrong, and I imagine this is something we could also mess up in the future.

Habbie:
> That's an interesting case, because we technically don't have to prove that the exact name does not exist because it does, but I can see how some DNSSEC validators might not like that.

Yes - one can write a validator that does not need it, but I imagine many have been written such that they do need it.

rgacogne:
> It is - but like this PR has isWildcard to treat a special case, it looks like unbound also has similar code except it gets it wrong, and I imagine this is something we could also mess up in the future.

I just realized after writing my answer what the special case is, will write a test!

rgacogne:
> You are right, it's much more correct to call them expanded onto themselves that non-expanded. I'll rewrite that fix because I don't want to introduce more confusion, this topic is already complicated enough.

Done.

> That's an interesting case, because we technically don't have to prove that the exact name does not exist because it does, but I can see how some DNSSEC validators might not like that.

Fixed.

> I just realized after writing my answer what the special case is, will write a test!

Done.

Habbie:
> > That's an interesting case, because we technically don't have to prove that the exact name does not exist because it does, but I can see how some DNSSEC validators might not like that.
> 
> Fixed.

I'm confused:
```
 dig a \*.h4ha.com @127.0.0.1 -p 5300 +dnssec

; <<>> DiG 9.14.0 <<>> a *.h4ha.com @127.0.0.1 -p 5300 +dnssec
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 10645
;; flags: qr rd ra; QUERY: 1, ANSWER: 2, AUTHORITY: 0, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags: do; udp: 512
;; QUESTION SECTION:
;*.h4ha.com.			IN	A

;; ANSWER SECTION:
*.h4ha.com.		260	IN	A	192.155.81.104
*.h4ha.com.		260	IN	RRSIG	A 13 2 300 20190509000000 20190418000000 29592 h4ha.com. i+0hOzjiMYiRGrYc7x/3hr3psHDIUP/1Mom7YsK5v9yDOU/4d+wDsnOb 64jP7vxy4AFYAPLElOIj5Oxwoo0Z4A==

;; Query time: 0 msec
;; WHEN: Fri Apr 26 14:50:18 CEST 2019
;; MSG SIZE  rcvd: 159

```

Habbie:
Besides that, the PR still behaves correctly, like the previous version of it.

Habbie:
> I'm confused:

I'm no longer confused:
```
$ dig a \*.h4ha.com  +dnssec @ns3.epik.com

; <<>> DiG 9.14.0 <<>> a *.h4ha.com +dnssec @ns3.epik.com
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 38918
;; flags: qr aa rd; QUERY: 1, ANSWER: 2, AUTHORITY: 0, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags: do; udp: 1232
;; QUESTION SECTION:
;*.h4ha.com.			IN	A

;; ANSWER SECTION:
*.h4ha.com.		300	IN	A	192.155.81.104
*.h4ha.com.		300	IN	RRSIG	A 13 2 300 20190509000000 20190418000000 29592 h4ha.com. wP4O4n874h/HuuIZoLcC5vpq1fJSJPslZ/XganMRFnFfkied4EEAuwQR zvlRZRr7Is/lmf9/qQpGMYb5ti3ZSg==

;; Query time: 116 msec
;; SERVER: 52.55.168.70#53(52.55.168.70)
;; WHEN: Fri Apr 26 14:55:44 CEST 2019
;; MSG SIZE  rcvd: 159
```

PR #7723, opened 2019-04-15
rename supermaster to superslave, and mention in upgrade notes

### LINKED ISSUE #7720: Auth 4.2 supermaster setting clarification
Two minor things regarding the 4.2 Auth nameserver.

In 4.2, you specifically require to enable 'supermaster' in the configuration file if you want to run as a superslave. This might be confusing (i.e. you enable support for notifies from supermasters, but it reads as if the machine should be supermaster itself, which isn't something you configure). Perhaps change wording to 'superslave' instead to clarify?

Also, very minor issue, the repository page shows 4.2 as if it would be an actual release - Given its still a release candidate, perhaps add this to the repository name as well, so people grab 4.1 for production instead. :)

Habbie:
### Short description
The old name was very confusing. Fixes #7720.

### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/docs/modes-of-operation.rst
+++ b/docs/modes-of-operation.rst
@@ -201,10 +201,10 @@ itself as a slave for that zone.
 Before a supermaster notification succeeds, the following conditions
 must be met:
 
- - :ref:`setting-supermaster` support must be enabled
+ - :ref:`setting-superslave` support must be enabled
  - The supermaster must carry a SOA record for the notified domain
- - The supermaster IP must be present in the 'supermaster' table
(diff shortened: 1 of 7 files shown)


Habbie:
This SHOULD fail on Travis.

Habbie:
> This SHOULD fail on Travis.

Right after pushing, I noticed I forgot to fix the supermaster tests in regression-tests.nobackend. I decided to let Travis run to see if they would in fact fail, and they did. I have now amended the commit to update those too, and now Travis should pass.

Habbie:
This has passed on Travis.

PR #7724, opened 2019-04-15
Fix unittest order deps

omoerbeek:
### Short description

To make it possible to run individual unit test they should not depend on each other.
This makes sure all tests setup globals properly. Additionally, the testrunners of auth and rec call `reportAllTypes()` in init code since a lot of tests need that and a large set did not call it but depended silently on other tests to do it for them.

`./testrunner --random` now works all the time for all subprojects as well.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/recursordist/testrunner.cc
+++ b/pdns/recursordist/testrunner.cc
@@ -30,7 +30,6 @@
 #include <dnsrecords.hh>
 
 bool init_unit_test() {
-  std::cerr << "Initing..." << std::endl;
   reportAllTypes();
   return true;
 }
--- a/pdns/recursordist/testrunner.cc
+++ b/pdns/recursordist/testrunner.cc
@@ -20,12 +20,23 @@
  * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
  */
 #define BOOST_TEST_DYN_LINK
-#define BOOST_TEST_MAIN
-#define BOOST_TEST_MODULE unit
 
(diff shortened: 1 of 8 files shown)


pdns/test-packetcache_cc.cc
```diff
@@ -206,6 +207,7 @@ BOOST_AUTO_TEST_CASE(test_PacketCacheThreaded) {
     PC.setTTL(3600);
 
     g_PC=&PC;
+    g_QCmissing = 0;
```
rgacogne:
```suggestion
    g_PCmissing = 0;
```

pdns/recursordist/testrunner.cc
```diff
@@ -20,12 +20,23 @@
...
 #ifdef HAVE_CONFIG_H
 #include "config.h"
 #endif
 #include <boost/test/unit_test.hpp>
 
+#include <iostream>
+#include <dnsrecords.hh>
 
+bool init_unit_test() {
+  std::cerr << "Initing..." << std::endl;
```
rgacogne: Looks like a left-over?

pdns/testrunner.cc
```diff
@@ -19,3 +17,16 @@ ArgvMap &arg()
   static ArgvMap theArg;
   return theArg;
 }
+
+
+bool init_unit_test() {
+  cerr << "Initing..." << endl;
```
rgacogne: Same, unless you really wanted it :-)

PR #7726, opened 2019-04-15
dnsdist: Add DNS over HTTPS support based on libh2o

rgacogne:
### Short description

This PR is a rebase of #6911 on top of the current master, with a few fixes.
It also has less code duplication because of the cleanup done in #7526.
The only remaining issue I'm aware of is that we always send the request received over DoH to the backend over UDP, advertising an EDNS Payload Size of 4096 (unless there was an existing EDNS OPT RR). It means that the current design can't handle responses larger than 4096 bytes over DoH, and leads to a weird situation if the backend answers with `TC=1`.

~~Still needs documentation, and quite a lot of regression tests.~~ I'm not sure how to proceed for that last point since the version of Ubuntu used by our Travis tests is much too old to have the required `libh2o` library, and Python libraries with support for HTTP/2.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


pdns/dnsdistdist/doh.cc
```diff
@@ -0,0 +1,603 @@
...
+
+  auto sock = h2o_evloop_socket_create(dsc->h2o_ctx.loop, dsc->dohresponsepair[1], H2O_SOCKET_FLAG_DONT_READ);
+  sock->data = dsc;
+
+  // this listens to responses from dnsdist to turn into http responses
+  h2o_socket_read_start(sock, on_dnsdist);
+
+  // we should probably make that hash, algorithm etc line configurable too
+  if(setup_ssl(dsc, df->d_certFile.c_str(), df->d_keyFile.c_str(),
+                "DEFAULT:!MD5:!DSS:!DES:!RC4:!RC2:!SEED:!IDEA:!NULL:!ADH:!EXP:!SRP:!PSK") != 0)
```
rgacogne: Note to myself, this needs to be configurable!

rgacogne:
Now with:
- custom ciphersuites, including for TLS 1.3 ;
- certificate reloading ;
- TCP Fast Open, reuseport, interface and CPU pinning ;
- more than one certificate (ECDSA and RSA, for example) ;
- documentation.

rgacogne:
I think this PR is done, it's still missing tests but I'm afraid we won't be able to run them on Travis anyway, so I'd be happy to merge this PR as soon as it has been reviewed.
In the meantime I'll write regression tests and disable them in Travis.

pieterlexis:
> Travis tests is much too old to have the required libh2o library

We can build it ourselves though (as we do for packages on OSs that have no libh2o-evloop)

> Python libraries with support for HTTP/2.

Perhaps we can take [hyper](https://hyper.readthedocs.io/en/latest/) for a spin

Habbie:
> > Travis tests is much too old to have the required libh2o library

How about testing on a fresher docker image in circleci?

Habbie:
> How about testing on a fresher docker image in circleci?

I've done (ugly, unfinished) circleci work based on this branch, in https://github.com/Habbie/pdns/tree/dnsdist-redoh-circleci (py2, with hacks, but all tests pass) - probably best to integrate that into the bigger matrix @pieterlexis is working on, when that has landed.

pdns/dnsdist-lua.cc
```diff
@@ -1650,6 +1648,113 @@ void setupLuaConfig(bool client)
...
+          if (index < g_dohlocals.size()) {
+            result = g_dohlocals.at(index);
+          }
+          else {
+            errlog("Error: trying to get DOH frontend with index %zu but we only have %zu\n", index, g_dohlocals.size());
+            g_outputBuffer="Error: trying to get DOH frontend with index " + std::to_string(index) + " but we only have " + std::to_string(g_dohlocals.size()) + "\n";
+          }
+        }
+        catch(const std::exception& e) {
+          g_outputBuffer="Error: "+string(e.what())+"\n";
```
chbruyand: A word on context between "Error" and the exception may help debugging if an issue is caught there.

pdns/dnsdistdist/doh.cc
```diff
@@ -0,0 +1,707 @@
...
+        else
+          ++dsc->df->d_http1queries;
+
+        doh_dispatch_query(dsc, self, req, std::move(decoded), remote);
+      }
+    }
+    else
+    {
+      vinfolog("HTTP request without DNS parameter: %s", req->path.base);
+      h2o_send_error_400(req, "Bad Request", "dnsdist " VERSION " could not find DNS parameter", 0);
```
chbruyand: I see that errors - mostly - send the dnsdist's version in their explanation message. Maybe this would be nicer to add a HTTP header (such as `x-dnsdist-version`) with dnsdist version and and option to disable that ?

rgacogne: I really don't think we should advertise the product and version at all. Unless someone strongly disagrees, I'll remove "dnsdist' and the version from these messages.

Habbie: If we need to make the choice, it's better not to advertise. `version.bind` and friends are configurable in auth and rec, I suppose that would be the obvious thing to do -eventually-.

pdns/dnsdistdist/docs/reference/config.rst
```diff
@@ -100,6 +100,29 @@ Listen Sockets
...
+
+  .. versionadded:: 1.4.0
+
+  Listen on the specified address and TCP port for incoming DNS over HTTPS connections, presenting the specified X.509 certificate.
+
+  :param str address: The IP Address with an optional port to listen on.
+                      The default port is 443.
+  :param str certFile(s): The path to a X.509 certificate file in PEM format, or a list of paths to such files.
+  :param str keyFile(s): The path to the private key file corresponding to the certificate, or a list of paths to such files, whose order should match the certFile(s) ones.
+  :param list url: A list of URLs to accept queries on. The default is /.
```
chbruyand: This is a "base" path specifier and not an "exact match". This means that any path under one of the given path will be treated (if one of the URL is "/foo", any DoH request against "/foo/bar?dns=..." will be treated). Maybe we could make that clear here ?

Also accepting a string OR a list of strings would be so nice. I can PR that if needed...

pdns/dnsdistdist/doh.cc
chbruyand: On the same topic, h2o advertises it's version "Server: h2o/2.2.5". Maybe you could also remove it :)

pdns/dnsdistdist/doh.cc
```diff
@@ -0,0 +1,718 @@
...
+  auto accept_ctx = d_dsc->accept_ctx->get();
+  accept_ctx->ssl_ctx = tlsCtx.release();
+  d_dsc->accept_ctx->release();
+}
+
+// this is the entrypoint from dnsdist.cc
+void dohThread(ClientState* cs)
+try
+{
+  std::shared_ptr<DOHFrontend>& df = cs->dohFrontend;
```
chbruyand: Would be nice to add a call to `setThreadName("dnsdist/dohThre");`

chbruyand:
\o/

PR #7730, opened 2019-04-17
rec: Move replaced negcache entries to the back of the expunge queue

rgacogne:
### Short description

Otherwise they might get expunged very quickly while they just have been inserted.
Similar to https://github.com/PowerDNS/pdns/pull/3451, which fixed this issue for the positive caches.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


rgacogne:
I refactored this commit a bit since the existing `replacing_insert()` function was used by at least one `multi_index` container with only one index (in the bind backend) so I could not alter it.

PR #7731, opened 2019-04-18
rec: Fix the cache cleaning code being only run once for workers

rgacogne:
### Short description

The `s_running` guard was only set to `false` for the handler thread and remained `true` after the first run for all the others.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


rgacogne:
The regression was introduced in #7059 so only master and 4.2.0-alpha1 are impacted.

PR #7732, opened 2019-04-18
block stdin when spawning dnsdist for testing

Habbie:
### Short description
Otherwise, if stdin is a not a terminal, dnsdist might immediately exit because it reads EOF.

### Checklist

I have:
- [ ] compiled this code
- [x] tested this code
- [ ] documented the code


pieterlexis:
this is why there is the `--supervised` switch to dnsdist

Habbie:
> this is why there is the `--supervised` switch to dnsdist

Ah yes!

I have amended the PR to do that instead of making stdin a pipe.

PR #7749, opened 2019-04-24
refactor tls/https lib handling to cover all option combinations

Habbie:
### Short description
This fixes linking against a custom libcrypto, and building with no TLS and no HTTPS.

### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdistdist/Makefile.am
+++ b/pdns/dnsdistdist/Makefile.am
@@ -191,18 +191,18 @@ if HAVE_RE2
 dnsdist_LDADD += $(RE2_LIBS)
 endif
 
+if HAVE_LIBSSL
+dnsdist_LDADD += $(LIBSSL_LIBS)
+endif
+
 if HAVE_LIBCRYPTO
-dnsdist_LDADD += $(LIBCRYPTO_LIBS)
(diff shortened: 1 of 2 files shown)


Habbie:
Travis failure is unrelated (auth, apt-get).

pdns/dnsdistdist/Makefile.am
```diff
@@ -191,19 +191,19 @@ if HAVE_RE2
 dnsdist_LDADD += $(RE2_LIBS)
 endif
 
+if HAVE_LIBSSL
+dnsdist_LDADD += $(LIBSSL_LIBS)
```
rgacogne: The addition of `$(LIBSSL_CFLAGS)` to `AM_CPPFLAGS` is still only done when either `HAVE_DNS_OVER_TLS` or `HAVE_DNS_OVER_HTTPS` is set, perhaps we should change that? Otherwise we might include the wrong files.

pieterlexis: HAVE_LIBSSL cannot be true if neither is set (enforced in the configure.ac)

rgacogne: You are right so it doesn't matter!

PR #7770, opened 2019-05-02
Revert "auth: fix referral response for DS queries"

pieterlexis:
Reverts PowerDNS/pdns#7637

this _might_ have broken master


pieterlexis:
Fixed by #7772

Issue #7785, opened 2019-05-06
https://github.com/PowerDNS/pdns/pull/7772 breaks recursor regression test


omoerbeek:
Above commit breaks a recursor regression test: `test_basicNSEC.py:basicNSEC.testSecureDNAMEToInsecureCNAMEAnswer`

Resursor test log is attached. It is not  clear yet if this is an auth or recursor bug, but the first suspect would be auth.

The recursor tests uses the auth process. Bisecting the auth code lead to the above commit.
[recursor.log](https://github.com/PowerDNS/pdns/files/3148528/recursor.log)

@mind04 any clue?

Habbie:
auth+rec at 5510ebdcd breaks; auth+rec at the previous merge commit, 5d270e867, works. Auth at the parent (5d270e867) and rec at the breaking commit (5510ebdcd) works. Just confirming that the cause is an auth change (which does not mean it's an auth bug).

Habbie:
Here's an auth test - https://gist.github.com/Habbie/e462e2e58c00ae027dab78646b7b66b6

With `./start-test-stop 5300 gsqlite3`, at 5510ebd and up, you get:
```
+./tests/secure-cname-to-insecure/drillchase.out:|---Bogus DNSSEC signature:
+./tests/secure-cname-to-insecure/drillchase.out:No trusted keys found in tree: first error was: Bogus DNSSEC signature
+./tests/secure-cname-to-insecure/unbound-host.out:cname-to-insecure.example.com is an alias for www.insecure.dnssec-parent.com. (BOGUS (security failure))
+./tests/secure-cname-to-insecure/unbound-host.out:www.insecure.dnssec-parent.com has address 192.0.2.88 (BOGUS (security failure))
```

The reason is that the CNAME RRSIG, that signs the CNAME, after 5510ebd also signs `0	www.insecure.dnssec-parent.com.	IN	A	120	192.0.2.88`

Habbie:
The comment at https://github.com/PowerDNS/pdns/blob/master/pdns/dnssecsigner.cc#L134 is a lie.

Habbie:
Here's a patch that (taken out of context so maybe I'm wrong!) makes the test above pass - https://gist.github.com/Habbie/1761b84b8b3671173db3d52787571f45

However, this patch is wrong. It will, for example, break if the CNAME comes from a parent zone.

Habbie:
[...]

PR #7786, opened 2019-05-07
Part one of Circle-CI integration

pieterlexis:
### Short description
This is the first part of the integration of circle-ci.

This PR adds the Recursor to circle-ci, including regression tests and the bulk test (with the default 95% successrate).

Also added are a bunch of Authoritative backends:

* BIND
* gsql (gmysql, gsqlite3, odbc-sqlite3 (tests), gpgsql)
* LDAP
* LMDB
* MyDNS
* tinydns

Finally, several speed improvements have been made:

* Parallel testing is implemented
* ccache is used when building
* The repository is cloned 'shallow'

### Checklist

I have:
- [ ] compiled this code
- [x] tested this code
- [ ] documented the code


build-scripts/circleci.sh
```diff
@@ -74,4 +74,4 @@ __EOF__
         echo unknown command "$1"
         exit 1
         ;;
-esac
```
Habbie: This file can go, right?

.circleci/config.yml
```diff
@@ -1,68 +1,370 @@
...
 commands:
+  checkout-shallow:
+    description: "Do a shallow checkout of the repository"
+    steps:
+      - run:
+          name: Install git and openssh
+          command: apt-get update && apt-get -qq -y install openssh-client git
+      - run:
+          name: Store the keyid for the git repo
+          command: mkdir -p $HOME/.ssh && ssh-keyscan $(echo ${CIRCLE_REPOSITORY_URL} | sed 's,.*@\([^:]*\):.*,\1,') > $HOME/.ssh/known_hosts
```
Habbie: Any reason we are doing this instead of telling git/ssh to just don't care about the key? Or, alternatively, why not do an https checkout?

pieterlexis: This is mostly done to keep the git checkout very close to what circle is doing by default (SSH checkout)

build-scripts/circleci.sh
pieterlexis: yes, will fix

Habbie:
This PR breaks LDAP testing on Travis. I propose removing it there.

Habbie:
Travis recursor failure is unrelated (#7785)

Issue #7579, opened 2019-03-14
High CPU usage after receiving mass notify. 


george2asenov:
- Program: Authoritative
 - Issue type: Bug report

### Short description
After receiving many NOTIFY queries at once (like when restarting BIND master with more than 1000 domains in this case) PDNS (working as a hidden slave - BIND -> PDNS superslave/supermaster -> slave ) and MYSQL start to consume ~100%  constantly till next restart. This continues even after all AXFR-s are completed and log stop logging anything ( at loglevel=6) and no DNS queries (just in test environment). Just doing nothing but keep CPU consumption high.

issue: https://github.com/PowerDNS/pdns/issues/622 is the same but appears to be resolved back then. Apparently it is not. 

### Environment
 - Operating system:  CentOS 73.10.0-862.9.1.el7.x86_64
 - Software version: 4.1.6
 - Software source:  PowerDNS repository  - http://repo.powerdns.com/centos/$basearch/$releasever/auth-41

### Steps to reproduce
(Lua) scripts that are loaded. -->  none
1. configure PDNS with gmysql or gsqlite3  
2. configure PDNS as superslave of BIND with more than for example 500 domains 
3. restart the MASER BIND server so it send notifications for all domains.
4. wait all AXFR-s  complete 
5. watch  CPU start using ~1 core constanty

### Expected behaviour
After all AXFR complete pdns to become idle and consume ~0%CPU

### Actual behaviour
PDNS stuck on using too much cpu for an idling service. 
strace:
```
[pid 21655] write(29, "!\0\0\0\27\2\0\0\0\0\1\0\0\0\0\1\376\0\376\0\3SOA\obfusc-domain3.com", 37) = 37
[...]

PR #7790, opened 2019-05-07
Do not busy loop if we get lots of notifies. 

omoerbeek:
Also rewrite main slave loop to use a timed wait for the semaphore.

See https://github.com/PowerDNS/pdns/issues/7579

If the semaphore get incremented a lot (i.e. a tonne of notifies), the loop degenerates to a busy loop (rc=0 in that case since the semaphore is incremented for each post).
For each iteration all slave checks and notifies are done, so it does not make sense to iterate with the semaphore as counter.

Rewrite the loop to only iterate once per set of posts, and use a timedWait() since that makes the loop easier to grok.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [X] documented the code


pdns/communicator.cc
```diff
@@ -113,37 +113,19 @@ void CommunicatorClass::mainloop(void)
...
-              extraSlaveRefresh = true;
-          }
-          if (extraSlaveRefresh)
-            slaveRefresh(&P);
-        }
-        else { 
-          break; // something happened
-        }
-        // this gets executed at least once every second
-        doNotifications();
```
rgacogne: so `doNotifications()` used to be called at least every seconds, apparently on purpose, and it looks like it might not be called for as long as `slave-cycle-interval` (60s) after this PR. Is that intended?

pdns/communicator.cc
```diff
@@ -113,37 +113,19 @@ void CommunicatorClass::mainloop(void)
...
+      while (d_any_sem.tryWait() == 0) {
+       // eat up remaining posts, will do next iteration shortly
```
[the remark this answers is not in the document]
omoerbeek: yeah, already came to that conclusion myself in the train home. Will do.

pdns/communicator.cc
omoerbeek: Indeed. The solution probably ties in with our suggestion to use two threads. Have to think about this a bit.

omoerbeek: I decide to move the splitting of the master and slave to the future and concentrate on the solution of the busy loop using a condition var.

omoerbeek:
https://github.com/omoerbeek/pdns/commit/187d3ea64919544f1fc144768287aacaecf7d616

has a supposedly better version of the condition variable approach. 
I would suggest to merge the simple fix and leave the condition var one for after 4.2 is branched.

Issue #7458, opened 2019-02-06
max-cache-entries and max-packet-cache-entries have no effect


aldem:
- Program: Authoritative
 - Issue type: Bug report

### Short description
Looks like configuration parameters `max-cache-entries` and `max-packet-cache-entries` are either ignored or misbehaving. Even when set to low value (like 1000 entries), pdns actually ignores them, at least, according to statistics (`packetcache-size` and `query-cache-size`) and RAM usage (process consumed more than 1G or RAM despite cache size limited to 10000 entries).

### Environment

 - Operating system:  CentOS 7
 - Software version: pdns-4.1.6-1pdns.el7.x86_64
 - Software source: PowerDNS repository

### Steps to reproduce
1. Create relatively big set records
2. Set `max-cache-entries` and `max-packet-cache-entries`  to value significantly less than number of records
3. Query all records and observe that all of them are cached regardless of limit

### Expected behaviour
Configuration parameters `max-cache-entries` and `max-packet-cache-entries`  must actually define upper limit for number of cached entries.

### Actual behaviour
Configuration parameters `max-cache-entries` and `max-packet-cache-entries` are ignored. Not sure if this matters but backend is sqlite.

rgacogne:
Thank you for this report!
It might not be clear from our documentation (sorry about that!) but these numbers are not hard limits, meaning that both caches are allowed to cache more entries than the limit, but the cache cleaning algorithm will then remove enough entries to make sure that we are back under the limit. 
Note however that we might stay over the limit for a while depending on your traffic, since the cache cleaning algorithm is run roughly every 30s, but only if we receive queries.
Would you mind checking if the number of entries in the caches go back to normal after a while?

aldem:
Just checked - cache is not cleared at all (it is more than one hour, and at least 20 minutes since last query), number of cached entries has not changed (server is idling).

This "soft" limit is actually not really good idea - in constrained environment (limited RAM) it may kill process and there is no way to prevent it, and restarts are expensive.

It would be much better to be able to specify really hard limits.

rgacogne:
Adding the 'backport to stable?' label since the milestone has been updated to auth-4.2.0-beta1.

Habbie:
For now, I suggest setting `max-cache-entries` and `max-packet-cache-entries` way lower, so you have some headroom.

Habbie:
Hello! Are you able to see if #8713 (for master) or #8736 (for 4.2.x) fixes this for you?

[...]

PR #7793, opened 2019-05-08
Prevent unlimited memory usage by cache

NameWeb:
In my test-environment the cache would cause out of memory while max-cache-entries seemed to be ignored.
I first tried actually respecting max-cache-entries by not writing to the cache once it was more than full, but this seemed to cause a big penalty (in one test going from 17k QPS to about 11k QPS). 
The problem however seems to be reasonably under control with this change: if we don't skip cleanup if the cache is already too full. 

This change does still make things slower (now at around 14k QPS on my test-system when I try to trigger the out-of-memory), but the alternative was everything completely failing because of possible out of memory.
If s_maxcleaninterval is correctly chosen, then I expect the extra code to only be executed if we might be in the situation where the cache is being filled quicker than expected (and an out of memory is therefore a possibility).

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/auth-querycache.cc
+++ b/pdns/auth-querycache.cc
@@ -233,7 +233,7 @@ void AuthQueryCache::cleanupIfNeeded()
       uint64_t maxCached = (int)(2*d_maxEntries);
       uint64_t cacheSize = *d_statnumentries;
       
-      if (maxCached > cacheSize) {
+      if (cacheSize <= maxCached) {
         d_cleanskipped = true;
         d_nextclean += d_cleaninterval;
 
--- a/pdns/auth-querycache.cc
+++ b/pdns/auth-querycache.cc
@@ -226,12 +226,21 @@ void AuthQueryCache::cleanupIfNeeded()
     DLOG(g_log<<"cleaninterval: "<<d_cleaninterval<<", timediff: "<<timediff<<endl);
 
     if (d_cleaninterval == s_maxcleaninterval && timediff < 30) {
-      d_cleanskipped = true;
-      d_nextclean += d_cleaninterval;
+      /* 
+      There might be very good reasons for the cache size to be limited, 
+      lets not skip cleanup if we notice the cache is already oversized
+      */
+      uint64_t maxCached = (int)(2*d_maxEntries);
+      uint64_t cacheSize = *d_statnumentries;
+      
+      if (maxCached > cacheSize) {
+        d_cleanskipped = true;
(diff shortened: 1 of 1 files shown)


NameWeb:
Should make out of memory reported in #7458 much less likely.

pdns/auth-querycache.cc
```diff
@@ -226,12 +226,21 @@ void AuthQueryCache::cleanupIfNeeded()
...
-      d_nextclean += d_cleaninterval;
-
-      DLOG(g_log<<"cleaning skipped, timediff: "<<timediff<<endl);
-
-      return;
+      /* 
+      There might be very good reasons for the cache size to be limited, 
+      lets not skip cleanup if we notice the cache is already oversized
+      */
+      uint64_t maxCached = (int)(2*d_maxEntries);
```
rgacogne:
```suggestion
      uint64_t maxCached = 2*d_maxEntries;
```

pdns/auth-querycache.cc
```diff
@@ -226,12 +226,21 @@ void AuthQueryCache::cleanupIfNeeded()
...
+      uint64_t cacheSize = *d_statnumentries;
+      
+      if (maxCached > cacheSize) {
```
rgacogne: Completely identical, but it seems easier to read to me:
```suggestion
      if (cacheSize <= maxCached) {
```

NameWeb:
The same would indeed apply to the packet cache. 

But I have been pondering a bit myself and in retrospect I think my second attempt to solve this problem (#7796) would be a better solution. 

Pro's of #7796:
- Less likely to be triggered/cause overhead if not needed
- Also works for packet cache
- It makes the actual cache size more predictable, which would be good thing because they can be set in the config.

Con's of #7796:
- The cache can still exceed the max cache settings between two clean-ups and with only #7796 applied and not this patch, it can take 30 seconds before there is an other cleanup. So if the user is able to fill the cache up to out-of-memory within 30 seconds time, we would need this patch on top of #7796. I was however unable to trigger out-of-memory on a 2CPU 1G RAM machine with LMDB backend. For this to occur you would need a machine that has many CPU's, very quick back-end while only having very limited RAM. 

I'm not sure if we need to cater for that unlikely use case.

Therefore, my current advice is to not accept this patch, but accept #7796

Habbie:
We have merged a different fix for this in #8713. Thank you for your work!

PR #7796, opened 2019-05-08
Respect max-packet-cache-entries and max-cache-entries

NameWeb:
### Short description

Fixed proposed for #7458
This might replace or be combined with #7793 
This solution however seems to carry a smaller speed penalty and is less likely to be triggered when it's not actually needed.

If the initial attempt to clean the cache by only throwing away the old stuff wasn't sufficient to get below maxCached, then we will start throwing away from the top until we are at or close to the maximum. This way users are able to prevent out of memory by settings max-packet-cache-entries and max-cache-entries to appropriate values.

Code will execute only when max-cache has actually been exceeded. Overhead should be limited to when it's actually needed. 

I used dnsperf to test the speed and to stress-test on a 1G machine with LMDB backend and using fully random subdomains. Without this patch (and without #7793), PDNS would run out of memory after about 60 seconds. With the patch and with max-packet-cache-entries and max-cache-entries set to the default of 1000000 it would remain well within the memory limits.

My stress-tests triggered the code execution and even then speed penalty seems limited:
Two 60 second runs prior to the patch (runs cancelled just before it would otherwise go out of memory): 16981QPS and 17133QPS reported by dnsperf
Two 60 second runs with the patch: 16340QPS and 16988QPS

### Checklist

I have:
- [ X ] read the [CONTRIBUTING.md](https://github.com/PowerDNS/pdns/blob/master/CONTRIBUTING.md) document
- [ X ] compiled this code
- [ X ] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/cachecleaner.hh
+++ b/pdns/cachecleaner.hh
@@ -147,7 +147,27 @@ template <typename T> uint64_t pruneLockedCollectionsVector(vector<T>& maps, uin
     }
     totErased += erased;
   }
+  
+  toTrim = toTrim - totErased;
 
+  if (toTrim > 0) {
+    //Just trowing away the old stuff wasn't sufficient, start from the top
+    
+    for(auto& mc : maps) {
+      WriteLock wl(&mc.d_mut);
+      auto& sidx = boost::multi_index::get<2>(mc.d_map);
+      uint64_t erased = 0, lookedAt = 0;
+      for(auto i = sidx.begin(); i != sidx.end(); lookedAt++) {
+        i = sidx.erase(i);
+        erased++;
+
+        if(erased > toTrim / maps.size())
+          break;
+      }
+      totErased += erased;
+    }
+  }
+  
   return totErased;
 }
 
--- a/pdns/cachecleaner.hh
+++ b/pdns/cachecleaner.hh
@@ -159,12 +159,13 @@ template <typename T> uint64_t pruneLockedCollectionsVector(vector<T>& maps, uin
       
       //Is it possible for toThrowAway to be larger than sidx.size()? Better safe than sorry
       uint64_t toThrowAway = std::min(toTrim / maps.size(), sidx.size());
+      if (toThrowAway > 0) {
+        auto eraseIter = sidx.begin();
+        std::advance(eraseIter, toThrowAway);
+        sidx.erase(sidx.begin(),eraseIter);
 
-      auto eraseIter = sidx.begin();
-      std::advance(eraseIter, toThrowAway);
-      sidx.erase(sidx.begin(),eraseIter);
-
-      totErased += toThrowAway;
+        totErased += toThrowAway;
+      }
     }
   }
   
--- a/pdns/cachecleaner.hh
+++ b/pdns/cachecleaner.hh
@@ -151,7 +151,7 @@ template <typename T> uint64_t pruneLockedCollectionsVector(vector<T>& maps, uin
   toTrim = toTrim - totErased;
 
   if (toTrim > 0) {
-    //Just trowing away the old stuff wasn't sufficient, start from the top
+    //Just throwing away the old stuff wasn't sufficient, start from the top
     
     for(auto& mc : maps) {
       WriteLock wl(&mc.d_mut);
--- a/pdns/cachecleaner.hh
+++ b/pdns/cachecleaner.hh
@@ -156,8 +156,8 @@ template <typename T> uint64_t pruneLockedCollectionsVector(vector<T>& maps, uin
     for(auto& mc : maps) {
       WriteLock wl(&mc.d_mut);
       auto& sidx = boost::multi_index::get<2>(mc.d_map);
-      uint64_t erased = 0, lookedAt = 0;
-      for(auto i = sidx.begin(); i != sidx.end(); lookedAt++) {
+      uint64_t erased = 0;
+      for(auto i = sidx.begin(); i != sidx.end();) {
         i = sidx.erase(i);
         erased++;
 
--- a/pdns/cachecleaner.hh
+++ b/pdns/cachecleaner.hh
@@ -156,15 +156,15 @@ template <typename T> uint64_t pruneLockedCollectionsVector(vector<T>& maps, uin
     for(auto& mc : maps) {
       WriteLock wl(&mc.d_mut);
       auto& sidx = boost::multi_index::get<2>(mc.d_map);
-      uint64_t erased = 0;
-      for(auto i = sidx.begin(); i != sidx.end();) {
-        i = sidx.erase(i);
-        erased++;
+      
+      //Is it possible for toThrowAway to be larger than sidx.size()? Better safe than sorry
+      uint64_t toThrowAway = std::min(toTrim / maps.size(), sidx.size());
 
-        if(erased > toTrim / maps.size())
-          break;
-      }
-      totErased += erased;
+      auto eraseIter = sidx.begin();
+      std::advance(eraseIter, toThrowAway);
+      sidx.erase(sidx.begin(),eraseIter);
+
+      totErased += toThrowAway;
     }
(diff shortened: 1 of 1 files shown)


pdns/cachecleaner.hh
```diff
@@ -147,7 +147,27 @@ template <typename T> uint64_t pruneLockedCollectionsVector(vector<T>& maps, uin
...
+
+  if (toTrim > 0) {
+    //Just trowing away the old stuff wasn't sufficient, start from the top
+    
+    for(auto& mc : maps) {
+      WriteLock wl(&mc.d_mut);
+      auto& sidx = boost::multi_index::get<2>(mc.d_map);
+      uint64_t erased = 0, lookedAt = 0;
+      for(auto i = sidx.begin(); i != sidx.end(); lookedAt++) {
+        i = sidx.erase(i);
```
rgacogne: There is a version of `erase()` using iterators that is likely much faster then you are removing more than one entry at a time.

pdns/cachecleaner.hh
```diff
@@ -147,7 +147,27 @@ template <typename T> uint64_t pruneLockedCollectionsVector(vector<T>& maps, uin
...
+        erased++;
 
+        if(erased > toTrim / maps.size())
+          break;
```
rgacogne: One potential issue here is that this code will always nuke the content of the shards in the same order until it has nuked enough entries, which means the last shards will often not be reached. Perhaps we should start with a random shard every time?
It's not that bad with the existing code because we are expunging expired entries, but that would be an improvement there as well.

NameWeb: I think it would even be better to simply have the new code start from the bottom and keep removing the expired entries from the top. This will cause no extra overhead at all, and the two combined we have a good chance having checked the complete list and a small chance of going over entries twice.

rgacogne: Do you mean iterating over the shards in reverse order to remove all entries, during the second phase? That might help but I'm afraid it would still be very skewed.
And if you mean removing the entries inside the shards in reverse order it would remove the most recently used ones, which would not be a good idea.

NameWeb: Sorry, I wasn't taking into account the fact that most recently used would be at the bottom.

I'm very new to this code, so sorry if I'm mistaking. But taking that order into account, I would think it would be beneficial to always start at the top. Randomly starting lower to the bottom would mean we're deleting more recently used entries?

rgacogne: You are mixing two things, but that's because I wasn't clear enough about that, sorry. `maps` is a vector of shards, and I think it would make sense to start with a random shard for the final cleanup. `d_map` is a multi-indexed object whose second index should be a LRU. It's a LRU for the query and packet cache of the recursor, but surprisingly it looks like it was never a LRU in the auth.. It should be.

NameWeb: Sorry if this is becoming a back-and-forth, but aren't we always going over all elements in maps, thus always going over all the shards?

pdns/cachecleaner.hh
NameWeb: I changed it. The new version actually seem to have a 5% speed improvement.

pdns/cachecleaner.hh
rgacogne: Oh, right, you remove the same number of entries from each shard so the order doesn't matter. Something in the initial version made me think that you could have removed a lot of entries from the first shards and none at all from the last ones but I was clearly mistaken, sorry about that.

pdns/cachecleaner.hh
```diff
@@ -147,7 +147,27 @@ template <typename T> uint64_t pruneLockedCollectionsVector(vector<T>& maps, uin
...
+
+  if (toTrim > 0) {
+    //Just throwing away the old stuff wasn't sufficient, start from the top
+    
+    for(auto& mc : maps) {
+      WriteLock wl(&mc.d_mut);
+      auto& sidx = boost::multi_index::get<2>(mc.d_map);
+      
+      //Is it possible for toThrowAway to be larger than sidx.size()? Better safe than sorry
+      uint64_t toThrowAway = std::min(toTrim / maps.size(), sidx.size());
```
rgacogne: Perhaps check that `toThrowAway` is greater than 0 and `continue` otherwise?

NameWeb: We just checked if toTrim > 0 
and 
toThrowAway = toTrim / maps.size()

So for toThrowAway to be < 1, toTrim needed to be < maps.size(). I'm not sure how large maps.size() can be, but
statistical chances for this seems to be very slim and nothing will break of we don't check. Are we sure we want an extra check?

NameWeb:
> Code looks good, we still need to decide whether we want to apply this change.

Regarding the need for this patch: I think that at least something would need to be done against the possibility of not respecting memory usage boundaries for the cache. Currently a memory-bound installation of PDNS with LMDB backend will go out of memory after about 1 minute when under DDoS. With this patch it will keep responding. Yes, when under DDoS it will respond slightly slower than without the patch (before that goes out of memory). But this slow-down should only be triggered when there is indeed a problem with the cache being filled quicker than normal.
The only event I can think if when this runs unwanted would be when plenty of memory is available but the max-cache settings are set at a value too low for that system. If this would be a draw-back, then I would suggest we look into having an "auto" option for max-cache that doesn't have a hard-coded number, but calculates a value based on available system memory when PDNS is started.

pdns/cachecleaner.hh
rgacogne: Unless `sidx.size()` is 0, meaning this shard is empty.

NameWeb: Okay, I added a check to make sure toThrowAway > 0

NameWeb:
Quick note that I tried some other solutions for this problem, including deferring the cache insert if the maximum is reached, but from all my tests, this solutions gave the best QPS.

mind04:
In auth updated items are not moved back in the cache. After this change pdns will delete possible brand new cache items. This will result in additional backed lookups, especially under heavy load.

rgacogne:
> In auth updated items are not moved back in the cache.

Yes, we should fix that but that will probably be in 4.3.x.

> After this change pdns will delete possible brand new cache items. This will result in additional backed lookups, especially under heavy load.

When the cache is full and expunging expired entries was not enough, yes.

NameWeb:
True, but the alternative is out of memory and crashing. 
This clean-up can be prevented by setting the max-cache-entries to an appropriately large value.

I tried to find a way to remove the oldest elements first, but from my insight on how the cache is structured, there doesn't seem to be an easy/fast way of doing this.

mind04:
This pull does not solve the running out of memory. If you are able to fill the cache in one clean cycle you are still in trouble. Lower cleaninterval and maxcleaninterval (make them configurable) is likely a better approach to deal with low memory servers and high qps bursts.

NameWeb:
> This pull does not solve the running out of memory. If you are able to fill the cache in one clean cycle you are still in trouble. 

max-cache-entries is indeed not respected completely. The actual maximum would now be max-cache-entries + anything you can add in 30 seconds.
I did some attempts to solve this in an other way, but this patch seemed to have the smallest impact on the QPS that could be handled under high load.

One obvious attempt to solve this was: don't put anything new in the cache when it's full. The penalty in QPS was higher than with this fix. Probably because that way every single fresh cache entry which would have been handy, was never cached.

> Lower cleaninterval and maxcleaninterval (make them configurable) is likely a better approach to deal with low memory servers and high qps bursts.

#7793 attempts something similar: it will lower the interval if the cache is more than full. The conversation on that commit contains more details comparing it with this patch.

We are indeed not yet fully respecting max-cache-entries. But we're getting closer. And respecting it even more, comes with a speed penalty.

Habbie:
We have merged a different fix for this in #8713. Thank you for your work!

PR #7811, opened 2019-05-12
dnsdist: test on circleci; support py3 in tests

Habbie:
### Short description
This adds circleci testing for dnsdist, running on debian buster, because it's easier than stretch-backports (for libh2o-evloop) and buster will be out soon anyway.

This fixes running dnsdist tests on python3 on my local systems (debian buster); it also does not appear to break testing on Travis, somehow. Specifically this removes the 2.x/3.x split in the python protobuf dependency.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1131,6 +1131,8 @@ jobs:
   build-dnsdist:
     docker:
       - image: debian:buster
+        environment:
+          UBSAN_OPTIONS: 'print_stacktrace=1:halt_on_error=1'
     steps:
       - restore-cache-ccache:
           product: dnsdist
@@ -1207,6 +1209,8 @@ jobs:
   test-dnsdist-regression:
     docker:
       - image: debian:buster
+        environment:
+          UBSAN_OPTIONS: 'print_stacktrace=1:halt_on_error=1'
     steps:
       - install-dnsdist-deps
       - checkout-shallow
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -170,6 +170,27 @@ commands:
             default-libmysqlclient-dev \
             unixodbc
 
+  install-dnsdist-deps:
+    description: "Install all libraries needed for testing dnsdist"
+    steps:
+      - run: apt-get update
+      - run:
+          command: |
+            apt-get install -qq -y \
+              libluajit-5.1 \
+              libboost-all-dev \
+              libcap2 \
+              libcurl4-openssl-dev \
+              libfstrm0 \
+              libh2o-evloop0.13 \
+              libre2-5 \
+              libssl-dev \
+              libsystemd0 \
+              libsodium23 \
+              libprotobuf17 \
+              protobuf-compiler \
+              virtualenv
+
   install-auth-dev-deps:
     description: Install all packages needed to build the auth
     steps:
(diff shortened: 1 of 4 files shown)


Habbie:
We should add SNMP testing to recursor after this now that we have it set up anyway.

.circleci/config.yml
```diff
@@ -711,6 +732,106 @@ jobs:
...
+            --enable-dnstap \
+            --enable-dnscrypt \
+            --enable-dns-over-tls \
+            --enable-dns-over-https \
+            --prefix=/opt/dnsdist \
+            --with-libsodium \
+            --with-lua=luajit \
+            --with-libcap \
+            --with-protobuf=yes \
+            --with-re2
```
rgacogne: No ASAN or UBSAN?

Habbie: ah yes! those are missing on rec and auth too it appears. I'll have a go at it.

Habbie: We also don't test clang at all on circleci, but travis can keep doing that.

Habbie: gcc on buster appears to allow lsan+asan+ubsan, shall I do that?

Habbie: lsan+asan+ubsan pushed for dnsdist

Habbie:
@omoerbeek can you check that my py3 changes did not break dnsdist testing on OpenBSD?

.circleci/config.yml
```diff
@@ -711,6 +732,109 @@ jobs:
...
+      - run:
+          name: Install dependencies
+          command: |
+            apt-get update && apt-get -qq --no-install-recommends install \
+            autoconf \
+            automake \
+            bison \
+            ca-certificates \
+            curl \
+            flex \
```
zeha: dnsdist needs flex?

Habbie: no. pushed :)

Habbie:
The py3 changes did not break testing on macOS.

Adding the sanitizers broke the circleci build ('10 minutes without output'), now trying `-j1`.

Habbie:
squashed

.circleci/config.yml
Habbie: (removed lsan)

omoerbeek:
> @omoerbeek can you check that my py3 changes did not break dnsdist testing on OpenBSD?

doing that now

omoerbeek:
Only little problem I see is a call to `virtualenv` that on OpenBSD maps to the python2 version.
But that is easy to fix with a symbolic link in my ~/bin. Tests themselves seem to run.

Habbie:
> Only little problem I see is a call to `virtualenv` that on OpenBSD maps to the python2 version. The pythin 3 version is called `virtualenv-3`.

Yes, portably making a py3 venv is even harder than portably calling py3.

Thanks :)

pieterlexis:
> Only little problem I see is a call to virtualenv that on OpenBSD maps to the python2 version. The pythin 3 version is called virtualenv-3.

perhaps we should switch to running `python3 -m venv --symlinks $DIRECTORY`

Habbie:
> perhaps we should switch to running `python3 -m venv --symlinks $DIRECTORY`

I like it - let's do a separate PR for that though.

Habbie:
rebased

Habbie:
rebased

Habbie:
> PR looks good! Could we just export `UBSAN_OPTIONS=print_stacktrace=1:halt_on_error=1`, otherwise UBSAN is almost useless since it only prints some errors that nobody will ever see?

pushed, without the suppression we use in travis, just to see what happens.

PR #7813, opened 2019-05-13
Ensure a valid range to string() in PacketReader::getUnquotedText()

rgacogne:
### Short description

This PR is quite similar to #7272.

In some cases we might have called:
```
string::string(InputIt first, InputIt last)
```
with last < first, which is invalid.

`libstdc++` handles that gracefully by throwing an out-of-range exception but `libc++` tries to allocate a negative value of bytes, which in turns triggers a request for a very large memory allocation, which fails.

This issue was triggered on `OSS-fuzz`.
### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsparser.cc
+++ b/pdns/dnsparser.cc
@@ -484,6 +484,11 @@ string PacketReader::getUnquotedText(bool lenField)
   else
     stop_at = d_recordlen;
 
+  /* think unsigned overflow */
+  if (stop_at < d_pos) {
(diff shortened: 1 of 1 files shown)


Habbie:
Want to rebase so circleci can test it?

rgacogne:
> Want to rebase so circleci can test it?

Done!

PR #7825, opened 2019-05-15
 dnsdist: Implement SNIRule for DoT and DoH

### LINKED ISSUE #7210: Add dnsdist rules for DoT and DoH connections to detect the DNS server name
- Program: dnsdist 
 - Issue type: Feature request 

### Short description

For DoT and DoH it would be very useful to differenciate between different services based on the DNS servers domain name. With TLS based connections a DNS client can add the Server Name Indication (SNI) in the Client Hello message. The idea is to store the SNI for the whole session context and make it available via a dnsdist rule. For DoH the server name can be extracted from the request URI.

### Other information

**For DoH it can be extracted from the request URI:**
https://**service1.example.com**/dns-query

**For DoT and DoH it can be extracted from the SNI in the Client Hello:**
Extension: server_name (len=25)
    Type: server_name (0)
    Length: 25
    Server Name Indication extension
        Server Name list length: 23
        Server Name Type: host_name (0)
        Server Name length: 20
        Server Name: **service1.example.com**

### Usecase
Example rules in dnsdist:
- **TLSServerNameRule(servername)**
    Matches queries with SNI value specified in servername.

- **HTTPServerNameRule(servername)**
    Matches queries with DNS server domain name specified in servername.

If there would be a possibility to detect the DNS server name with dnsdist rules, there are countless possibilities. Here two examples:

Enable or disable options based on the DNS server name:
```
addAction(TLSServerNameRule("no-ecs.example.com"), DisableECSAction())
addAction(HTTPServerNameRule("no-ecs.example.com"), DisableECSAction())
addAction(AllRule(), PoolAction("DEFAULT"))
```

Send query to special downstream servers that offer different services (e.g. adblock, family-shield, safesearch for popular search engines, DNSSEC, etc.)
```
addAction(TLSServerNameRule("adblock.example.com"), PoolAction("BLOCK-ADS"))
addAction(HTTPServerNameRule("adblock.example.com"), PoolAction("BLOCK-ADS"))
addAction(AllRule(), PoolAction("DEFAULT"))
```

### Description
It seems that DoT and DoH are becoming more and more popular and the number of (native) client implemenations is increasing. For those two DNS transports it is normally possible to set a domain name or a URL (just DoH) in the DNS resolver setting. Also https://dnscrypt.info/stamps-specifications is saying the hostname will "_be used as a SNI name_" for DoT and DoH.

Here some interesting DoT/DoH implementations:
- Android has a native DoT implementation called "Private DNS" starting with "Android 9 Pie": https://android-developers.googleblog.com/2018/04/dns-over-tls-support-in-android-p.html
- A DoH implementation for Android is available with the Intra App (Btw. already tested, it is additionally sending the SNI): https://getintra.org
- kdig >=2.7.0 has a possibility to send a specific SNI ("+[no]tls-sni=STR"): https://www.knot-dns.cz/docs/2.7/singlehtml/

In the past the only possibility to differenciate between different services was to have multiple different IPs on the same DNS Resolver or different ports (might be very hard to configure for most DNS clients). Another options is to make use of the source IP address. But this is sometimes very hard because you need to know the current IP address in realtime and especially for mobile devices the IP changes all the time because of different networks (RAN, WiFi, etc.). The DNS server domain name will not change and is independent of IP addresses, NAT scenarios and transport networks. Making use it to offer different services for DoT and DoH connections can be very useful for a lot of applications.

rgacogne:
### Short description

Support for DoT should work with any version of OpenSSL or GnuTLS but for DoH it requires the addition of `h2o_socket_get_ssl_server_name()` in `libh2o` (PR in preparation). It's already possible to do HTTP Host header matching for DoH, though.

Closes #7210.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/pdns/dnsdist-console.cc
+++ b/pdns/dnsdist-console.cc
@@ -546,6 +546,7 @@ const std::vector<ConsoleKeyword> g_consoleKeywords{
   { "showVersion", true, "", "show the current version" },
   { "shutdown", true, "", "shut down `dnsdist`" },
   { "SkipCacheAction", true, "", "Don’t lookup the cache for this query, don’t store the answer" },
+  { "SNIRule", true, "name", "Create a rule which matches on the incoming TLS SNI value, if any (DoT or DoH)" },
   { "snmpAgent", true, "enableTraps [, masterSocket]", "enable `SNMP` support. `enableTraps` is a boolean indicating whether traps should be sent and `masterSocket` an optional string specifying how to connect to the master agent"},
   { "SNMPTrapAction", true, "[reason]", "send an SNMP trap, adding the optional `reason` string as the query description"},
   { "SNMPTrapResponseAction", true, "[reason]", "send an SNMP trap, adding the optional `reason` string as the response description"},
--- a/pdns/dnsdist-lua-bindings-dnsquestion.cc
+++ b/pdns/dnsdist-lua-bindings-dnsquestion.cc
@@ -85,6 +85,10 @@ void setupLuaBindingsDNSQuestion()
       return true;
     });
 
+  g_lua.registerFunction<std::string(DNSQuestion::*)()>("getServerNameIndication", [](const DNSQuestion& dq) {
+      return dq.sni;
+    });
+
   g_lua.registerFunction<void(DNSQuestion::*)(std::string)>("sendTrap", [](const DNSQuestion& dq, boost::optional<std::string> reason) {
 #ifdef HAVE_NET_SNMP
       if (g_snmpAgent && g_snmpTrapsEnabled) {
--- a/pdns/dnsdist-lua-rules.cc
+++ b/pdns/dnsdist-lua-rules.cc
@@ -295,6 +295,10 @@ void setupLuaRules()
     });
 #endif
 
+  g_lua.writeFunction("SNIRule", [](const std::string& name) {
+      return std::shared_ptr<DNSRule>(new SNIRule(name));
+  });
+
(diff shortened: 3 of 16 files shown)


rgacogne:
The corresponding PR to h2o is https://github.com/h2o/h2o/pull/2054. I don't think we need to wait until it is either merged or refused, though, because even if it does not make it we properly detect that the support is not present.

pieterlexis:
It might make sense to document that as of h2o version 2.3.0-beta, it is not possible to extract the SNI value from DOH connections and that HTTP Host should be used.

Habbie:
> It might make sense to document that as of h2o version 2.3.0-beta, it is not possible to extract the SNI value from DOH connections and that HTTP Host should be used.

Linking to HTTPHeaderRule in that bit of text would also be nice.

M4t7e:
I have tested the SNIRule for DoT and it works as expected so far. Thank you very much! It would be great to see this feature in the upcoming release :smiley:

rgacogne:
The `h2o` PR (https://github.com/h2o/h2o/pull/2054) has been merged so `h2o_socket_get_ssl_server_name()` should be in the upcoming 2.3.0 release.

rgacogne:
Thanks! I added a few words stating that `SNIRule` will unfortunately not work with h2o <= 2.3.0-beta, with a link to `HTTPHeaderRule`.

rgacogne:
Rebased to fix a conflict following the merge of #7830.

PR #7832, opened 2019-05-18
add DoH support to sdig

Habbie:
### Short description
This will need some autotooling help - I'm pretty sure building sdig without LUA records is now illegal, which is not exactly what we want.

This ignores the port number argument - should that argument be removed altogether in DoH mode?

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/minicurl.hh
+++ b/pdns/minicurl.hh
@@ -31,16 +31,21 @@
 class MiniCurl
 {
 public:
+  using MiniCurlHeaders = std::map<std::string, std::string>;
+
   static void init();
 
   MiniCurl(const string& useragent="MiniCurl/0.0");
   ~MiniCurl();
   MiniCurl& operator=(const MiniCurl&) = delete;
   std::string getURL(const std::string& str, const ComboAddress* rem=0, const ComboAddress* src=0);
-  std::string postURL(const std::string& str, const std::string& postdata);
+  std::string postURL(const std::string& str, const std::string& postdata, MiniCurlHeaders& headers);
 private:
   CURL *d_curl;
   static size_t write_callback(char *ptr, size_t size, size_t nmemb, void *userdata);
   std::string d_data;
+  struct curl_slist* d_header_list = nullptr;
   void setupURL(const std::string& str, const ComboAddress* rem=0, const ComboAddress* src=0);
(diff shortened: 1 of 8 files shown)


Habbie:
> I'm pretty sure building sdig without LUA records is now illegal

CircleCI has delivered proof of this:
```
In file included from minicurl.cc:25:0:
minicurl.hh:27:23: fatal error: curl/curl.h: No such file or directory
```

pdns/sdig.cc
```diff
@@ -249,7 +263,9 @@ try
           cerr<<"EDNS Subnet response: "<<reso.source.toString()<<", scope: "<<reso.scope.toString()<<", family = "<<reso.scope.getNetwork().sin4.sin_family<<endl;
 	}
       }
-
+      else if(iter->first == EDNSOptionCode::PADDING) {
+        cerr<<"EDNS Padding size: "<<(iter->second.size())<<endl;
```
Habbie: I think this `cerr` (and the ones before and after it) should be `cout`.

Habbie:
> This will need some autotooling help - I'm pretty sure building sdig without LUA records is now illegal, which is not exactly what we want.

This should be fixed now.

pdns/minicurl.hh
```diff
@@ -28,17 +28,22 @@
 #include "iputils.hh"
 // turns out 'CURL' is currently typedef for void which means we can't easily forward declare it
 
+using MiniCurlHeaders = std::map<std::string, std::string>;
```
rgacogne: Perhaps that could be moved inside the `MiniCurl` class to avoid cluttering the global namespace?

pdns/sdig.cc
Habbie: pushed

pdns/minicurl.hh
Habbie: pushed

Habbie:
rebased after coverity fixes

pdns/sdig.cc
Habbie: unpushed and saved at https://gist.github.com/Habbie/3719f3c68f7b0ced64c6228f8b60ddac because it involves updating a bunch of tests

PR #7838, opened 2019-05-21
build docs in circleci

pieterlexis:
### Short description
This will make circle ci build the documentation. When building master from the PowerDNS/pdns repository, we'll also upload the documents to the webserver.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


.circleci/config.yml
```diff
@@ -265,6 +265,128 @@ commands:
...
+            latexmk \
+            libboost-all-dev \
+            libedit-dev \
+            libluajit-5.1-dev \
+            libssl-dev \
+            make \
+            pkg-config \
+            ragel \
+            virtualenv
+            if [ "${CIRCLE_PROJECT_USERNAME}" = "PowerDNS" -a "${CIRCLE_PROJECT_REPONAME}" = "pdns" -a "${CIRCLE_BRANCH}" = "master" ]; then
```
Habbie: This long line is repeated a few times - if we ever need to change it, it would be easy to make mistakes. Perhaps we can make it an alias or a script?

pieterlexis: Looks like [conditional steps](https://circleci.com/docs/2.0/reusing-config/#defining-conditional-steps) don't do what we want. I'll have a look at other solutions

PR #7841, opened 2019-05-22
recursor: log UDP TC bits during trace

Habbie:
### Short description
Before:
```
May 22 15:48:06 [1] tipsport.cz: Trying IP 90.183.101.40:53, asking 'tipsport.cz|NS'
May 22 15:48:06 [1] tipsport.cz: using TCP with 90.183.101.40:53
```

leading to the question 'why is it using TCP?'

After:
```
May 22 15:48:06 [1] tipsport.cz: Trying IP 90.183.101.40:53, asking 'tipsport.cz|NS'
May 22 15:48:06 [1] tipsport.cz: truncated bit set, over UDP
May 22 15:48:06 [1] tipsport.cz: using TCP with 90.183.101.40:53
```

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/syncres.cc
+++ b/pdns/syncres.cc
@@ -2985,6 +2985,7 @@ bool SyncRes::doResolveAtThisIP(const std::string& prefix, const DNSName& qname,
       }
       return false;
     }
+    LOG(prefix<<qname<<": truncated bit set, over UDP"<<endl);
 
     return true;
   }
--- a/pdns/syncres.cc
+++ b/pdns/syncres.cc
@@ -2977,10 +2977,12 @@ bool SyncRes::doResolveAtThisIP(const std::string& prefix, const DNSName& qname,
   if(lwr.d_tcbit) {
     *truncated = true;
 
-    if (doTCP && !dontThrottle) {
+    if (doTCP) {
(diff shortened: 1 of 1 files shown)


pdns/syncres.cc
```diff
@@ -2983,6 +2983,7 @@ bool SyncRes::doResolveAtThisIP(const std::string& prefix, const DNSName& qname,
       t_sstorage.throttle.throttle(d_now.tv_sec, boost::make_tuple(remoteIP, qname, qtype.getCode()), 60, 3);
       return false;
     }
+    LOG(prefix<<qname<<": truncated bit set, over UDP"<<endl);
```
rgacogne: At a quick glance, it looks like this line could be logged if the query was sent over TCP provided that `dontThrottle` is set?
Actually it's quite weird that we skip the ` truncated bit set, over TCP?` log and the return when `dontThrottle` is set, shouldn't we only skip the throttling?

Habbie: Looks like 5c8c72bf1e4cec59c31445cdc6d97e7669c74669 failed to make that distinction. I'll refactor.

Habbie: refactored, force-pushed

PR #7843, opened 2019-05-23
rec: Small speedups in the recursion 'slow' path

rgacogne:
### Short description

We are mostly focused on the performance of the 'fast' path, when the query can be served from the packet cache, and we had some low-hanging inefficiencies in the 'regular' path:
- `SyncRes::shuffleInSpeedOrder()` was very inefficient, doing several copies of `DNSName` objects including one in a map. The new version does only one copy, computes the needed speeds directly there and returns it ;
- We used `dns_random()` when shuffling `NS` records, and their corresponding `A` and `AAAA` entries. These operations really doesn't require a CSPRNG and `dns_random()` is much slower than the default randomness source used in C++11 ;
- The `UDPClientSocks` class was keeping every file descriptor it allocated in a set, checking that returned FDs belonged to that set before removing them from the FD multiplexer and closing them in any case. I couldn't find a case where we could return a socket that was not allocated by that class, and even if we did the FD multiplexer already does an internal accounting and would throw an exception, so I don't think it makes sense to do the work twice ;
- This one is just a nit but we used to set `errno` to 0 in `asyncresolve()`, and I couldn't find a reason why. `errno` is a special variable and setting it has a non-negligible cost so let's skip that. 

Without the first two items, `SyncRes::shuffleInSpeedOrder()` sometimes accounted for 10% of the total CPU cost in my flame graphs and is at a more reasonable level with this PR applied.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/lwres.cc
+++ b/pdns/lwres.cc
@@ -169,7 +169,6 @@ int asyncresolve(const ComboAddress& ip, const DNSName& domain, int type, bool d
 
   srcmask = boost::none; // this is also our return value, even if EDNS0Level == 0
 
-  errno=0;
   if(!doTCP) {
     int queryfd;
     if(ip.sin4.sin_family==AF_INET6)
--- a/pdns/pdns_recursor.cc
+++ b/pdns/pdns_recursor.cc
@@ -544,7 +544,6 @@ public:
 
     if(connect(*fd, (struct sockaddr*)(&toaddr), toaddr.getSocklen()) < 0) {
       int err = errno;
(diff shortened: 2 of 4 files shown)


rgacogne:
Pushed a small additional commit making sure we don't leak a socket descriptor in the unlikely case where we fail to bind to a local address, or where `setsockopt()` fails. We might want to backport this last commit to the 4.2.0 branch, btw.

PR #7852, opened 2019-05-27
sqlite3: make journal mode configurable; default to WAL

### LINKED ISSUE #7251: Add support for JOURNAL_MODE when using SQLite3 backend
- Program: Authoritative
 - Issue type:Feature request 

### Short description

For using powerdns with an SQLite database, it would be helpful if "PRAGMA JOURNAL_MODE" could be supported.

### Environment

 - Operating system: Linux Arch x64
 - Software version: 4.1.1
 - Software source: docker hub psitrax/powerdns

### Usecase

I run powerdns in a docker container and bind mount the SQLite database file into the container. However, for each update, a *-journal file needs be created. Therefore, the parent directory (the one containing the database file) has also to be mounted into the container.

With JOURNAL_MODE=MEMORY, this would be not necessary at all. With JOURNAL_MODE=TRUNCATE, it would be sufficient to mount both the database and the journal file.

### Description

Suggestion: Add a gsqlite3-pragma-journal-mode configuration option to allow defining the journal mode to be used. (Has to be set on each connection.)

Possible values are: DELETE | TRUNCATE | PERSIST | MEMORY | WAL | OFF

For compatibility, gslite3-pragma-journal-mode should default to DELETE.

Whichever value is set, it should be forwarded to the SQLite3 engine immediately after opening any new database connection, i.e. "PRAGMA JOURNAL_MODE=xxx;" should be sent.

Habbie:
### Short description
WAL enables tremendous concurrency and avoids a lot of locking problems. It is a good default.

Fixes #7251 

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/modules/gsqlite3backend/gsqlite3backend.cc
+++ b/modules/gsqlite3backend/gsqlite3backend.cc
@@ -43,7 +43,7 @@ gSQLite3Backend::gSQLite3Backend( const std::string & mode, const std::string & 
 {
   try
   {
-    SSQLite3 *ptr = new SSQLite3( getArg( "database" ));
+    SSQLite3 *ptr = new SSQLite3( getArg( "database" ), getArg( "pragma-journal-mode") );
     setDB(ptr);
     if(!getArg("pragma-synchronous").empty()) {
       ptr->execute("PRAGMA synchronous="+getArg("pragma-synchronous"));
@@ -77,6 +77,7 @@ public:
     declare(suffix, "database", "Filename of the SQLite3 database", "powerdns.sqlite");
     declare(suffix, "pragma-synchronous", "Set this to 0 for blazing speed", "");
     declare(suffix, "pragma-foreign-keys", "Enable foreign key constraints", "no" );
+    declare(suffix, "pragma-journal-mode", "SQLite3 journal mode", "WAL");
 
     declare(suffix, "dnssec", "Enable DNSSEC processing","no");
 
--- a/pdns/pdnsutil.cc
+++ b/pdns/pdnsutil.cc
@@ -2081,7 +2081,7 @@ try
       return 0;
     }
     try {
-      SSQLite3 db(cmds[1], true); // create=ok
+      SSQLite3 db(cmds[1], "", true); // create=ok
       vector<string> statements;
       stringtok(statements, sqlCreate, ";");
       for(const string& statement :  statements) {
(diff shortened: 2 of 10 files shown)


modules/gsqlite3backend/gsqlite3backend.cc
```diff
@@ -77,6 +77,7 @@ class gSQLite3Factory : public BackendFactory
     declare(suffix, "database", "Filename of the SQLite3 database", "powerdns.sqlite");
     declare(suffix, "pragma-synchronous", "Set this to 0 for blazing speed", "");
     declare(suffix, "pragma-foreign-keys", "Enable foreign key constraints", "no" );
+    declare(suffix, "journal-mode", "SQLite3 journal mode", "WAL");
```
mind04: Why journal-mode instead of pragma-journal-mode, like the settings above this one?

pdns/pdnsutil.cc
```diff
@@ -2081,7 +2081,7 @@ try
       return 0;
     }
     try {
-      SSQLite3 db(cmds[1], true); // create=ok
+      SSQLite3 db(cmds[1], "WAL", true); // create=ok
```
mind04: Why is this WAL hard-coded and not imported from pdns.conf or the command line?
Ps I'm not even sure you need a journal-mode when you create a database.

modules/gsqlite3backend/gsqlite3backend.cc
Habbie: Good point!

pdns/pdnsutil.cc
Habbie: Because any operations after this one will then also use WAL, even if not done through pdnsutil. 
If we take it from pdns.conf, most users will also get WAL.

It's not super important, if you think it's confusing or surprising, we can take it out.

modules/gsqlite3backend/gsqlite3backend.cc
Habbie: pushed

pdns/pdnsutil.cc
mind04: Yes, I think it is confusing. The docs say about journal-mode "Set to empty to leave the journal mode alone." Which is true for existing databases. But for new ones you silently default to WAL regardless of journal-mode being empty or not.

If you don't set the journal-mode to WAL at create time it will happen after the first connection to the db from powerdns or pdnsutil. This time the value depends on the settings in dns.conf or command line.

Habbie: Ok! Pushed!

Habbie:
squashed/rebased

PR #7853, opened 2019-05-27
dnsdist: Support Prometheus latency histograms

### LINKED ISSUE #6088: Expose sum of all observations for histogram statistics
 - Issue type: Feature request

### Usecase
To be able to present pdns statistics histograms (i.e. `answers`, `latency`, `auth4-answers`, `auth6-answers`) as Prometheus histograms the sum of all observations also needs to be presented as a metric, alongside the buckets.

### Description
While trying to implement [Prometheus histograms](https://prometheus.io/docs/practices/histograms/) we've ran into an issue, namely that Prometheus histograms require the sum of all observations to be presented as well. e.g. in the case of auth answers histogram (`answers-*` metrics) the "total time spent answering queries" should also be exposed.

thanks!

Marlinc:
### Short description
This PR adds support for Prometheus latency histograms as described in: https://prometheus.io/docs/concepts/metric_types/#histogram

It adds 3 metrics, `latency_sum`, `latency_count` and `latency_bucket`. The `latency_bucket` metrics have a `le` label which has the `lower and equal to` latency of the request in ms.

This resolves #6088 for dnsdist.

Example graph:
image

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/pdns/dnsdist-web.cc
+++ b/pdns/dnsdist-web.cc
@@ -443,6 +443,22 @@ static void connectionThread(int sock, ComboAddress remote)
           output << "\n";
         }
 
+        // Latency histogram buckets
+        output << "# HELP dnsdist_latency Histogram of responses by latency\n";
+        output << "# TYPE dnsdist_latency histogram\n";
+        uint64_t latency_amounts = g_stats.latency0_1;
+        output << "dnsdist_latency_bucket{le=\"1\"} " << latency_amounts << "\n";
+        latency_amounts += g_stats.latency1_10;
+        output << "dnsdist_latency_bucket{le=\"10\"} " << latency_amounts << "\n";
+        latency_amounts += g_stats.latency10_50;
+        output << "dnsdist_latency_bucket{le=\"50\"} " << latency_amounts << "\n";
+        latency_amounts += g_stats.latency50_100;
+        output << "dnsdist_latency_bucket{le=\"100\"} " << latency_amounts << "\n";
+        latency_amounts += g_stats.latency100_1000;
+        output << "dnsdist_latency_bucket{le=\"1000\"} " << latency_amounts << "\n";
+        latency_amounts += g_stats.latencySlow; // Should be the same as latency_count
+        output << "dnsdist_latency_bucket{le=\"+Inf\"} " << latency_amounts << "\n";
+
         auto states = g_dstates.getLocal();
         const string statesbase = "dnsdist_server_";
 
--- a/pdns/dnsdist.hh
+++ b/pdns/dnsdist.hh
@@ -225,6 +225,8 @@ extern GlobalStateHolder<NetmaskTree<DynBlock>> g_dynblockNMG;
 
 extern vector<pair<struct timeval, std::string> > g_confDelta;
 
+extern uint64_t getLatencyCount(const std::string&);
+
 struct DNSDistStats
 {
   using stat_t=std::atomic<uint64_t>; // aww yiss ;-)
@@ -251,7 +253,7 @@ struct DNSDistStats
   stat_t noPolicy{0};
   stat_t cacheHits{0};
   stat_t cacheMisses{0};
-  stat_t latency0_1{0}, latency1_10{0}, latency10_50{0}, latency50_100{0}, latency100_1000{0}, latencySlow{0};
+  stat_t latency0_1{0}, latency1_10{0}, latency10_50{0}, latency50_100{0}, latency100_1000{0}, latencySlow{0}, latencySum{0};
   stat_t securityStatus{0};
 
   double latencyAvg100{0}, latencyAvg1000{0}, latencyAvg10000{0}, latencyAvg1000000{0};
@@ -298,7 +300,10 @@ struct DNSDistStats
     {"fd-usage", getOpenFileDescriptors},
     {"dyn-blocked", &dynBlocked},
     {"dyn-block-nmg-size", [](const std::string&) { return g_dynblockNMG.getLocal()->size(); }},
-    {"security-status", &securityStatus}
+    {"security-status", &securityStatus},
+    // Latency histogram
+    {"latency-sum", &latencySum},
+    {"latency-count", getLatencyCount},
   };
 };
 
@@ -390,6 +395,9 @@ struct MetricDefinitionStorage {
     { "dyn-blocked",            MetricDefinition(PrometheusMetricType::counter, "Number of queries dropped because of a dynamic block")},
     { "dyn-block-nmg-size",     MetricDefinition(PrometheusMetricType::gauge,   "Number of dynamic blocks entries") },
     { "security-status",        MetricDefinition(PrometheusMetricType::gauge,   "Security status of this software. 0=unknown, 1=OK, 2=upgrade recommended, 3=upgrade mandatory") },
+    // Latency histogram
+    { "latency-sum",            MetricDefinition(PrometheusMetricType::counter, "Total response time in milliseconds")},
+    { "latency-count",          MetricDefinition(PrometheusMetricType::counter, "Number of queries contributing to response time histogram")},
   };
 };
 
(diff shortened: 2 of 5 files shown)


pdns/dnsdist.hh
```diff
@@ -298,7 +298,10 @@ struct DNSDistStats
     {"fd-usage", getOpenFileDescriptors},
     {"dyn-blocked", &dynBlocked},
     {"dyn-block-nmg-size", [](const std::string&) { return g_dynblockNMG.getLocal()->size(); }},
-    {"security-status", &securityStatus}
+    {"security-status", &securityStatus},
+    // Latency histogram
+    {"latency-sum", &latencySum},
+    {"latency-count", &responses}, // This is the same metric as responses but uses the metric name as expected by Prometheus
```
rgacogne: I'm afraid this is not accurate, we do update the latency metrics for cache-hits and self-generated answers (see `prepareOutgoingResponse()` and the `responses` counter is not incremented in these cases.

rgacogne: `g_stats.responses` + `g_stats.selfAnswered` + `g_stats.cacheHits` would be closer to the value you are looking for.

rgacogne: Adding these values here would cause them to be sent over carbon and the HTTP API as well, where they have no meaning. Perhaps it's time to replace the pair in this vector with a proper structure so we can specify that these are for prometheus use only? This is not really a blocker since carbon and API consumers could simply discard them, but it feels a bit wrong.

Marlinc: Oh okay didn't know that, will try and change that

Marlinc: Now added a stats function which returns the sum of those 3 values

Marlinc: Well if Prometheus can use them to generate histograms other consumers might be able to use those stats as well

pdns/dnsdistdist/docs/statistics.rst
```diff
@@ -99,6 +99,18 @@ latency-slow
 ------------
 Number of queries answered in more than 1 second.
 
+latency-sum
+------------
+Total response time in milliseconds.
+
+latency-count
+------------
```
rgacogne:
```suggestion
-------------
```

pdns/dnsdistdist/docs/statistics.rst
```diff
@@ -99,6 +99,18 @@ latency-slow
...
+Number of queries contributing to response time histogram
+
+latency-bucket
+------------
```
rgacogne:
```suggestion
--------------
```

pdns/dnsdistdist/docs/statistics.rst
```diff
@@ -99,6 +99,18 @@ latency-slow
 ------------
 Number of queries answered in more than 1 second.
 
+latency-sum
+------------
```
rgacogne:
```suggestion
-----------
```

pdns/dnsdist.hh
rgacogne: Fair enough. I would bet they are never going to get used outside of prometheus but that's fine, it's probably not worth the additional code needed to hide them.

Marlinc:
Fixes all the things!

Marlinc:
image

wojas:
Am I correct that the metrics are not prefixed with `dnsdist_server_` unlike the other ones?

Marlinc:
That is a good one and is by mistake, I will fix that

wojas:
@Marlinc Great! Correction, the prefix should be `pdns_dnsdist_`.

Marlinc:
Actually, what is the difference between the `dnsdist_server_` metrics and `dnsdist_`? None of the current metrics start with `pdns_dnsdist_`.

wojas:
I just realize that this will conflict with the existing metrics...

How about something like `pdns_dnsdist_response_latency_(bucket|sum|count)`? 

The `server` metrics are the ones derived from the per-server stats.

wojas:
@Marlinc Sorry, you're right, I was looking at the output of an old dnsdist exporter instead of the new native dnsdist metrics... `dnsdist_latency_bucket` etc will be perfectly fine!

Marlinc:
Added the `dnsdist_` prefix and added `HELP` and `TYPE` definitions to `dnsdist_latency_bucket`

pdns/dnsdist-web.cc
```diff
@@ -443,6 +443,22 @@ static void connectionThread(int sock, ComboAddress remote)
           output << "\n";
         }
 
+        // Latency histogram buckets
+        output << "# HELP dnsdist_latency_bucket Number of queries contributing to response time histogram per latency bucket\n";
+        output << "# TYPE dnsdist_latency_bucket counter\n";
```
wojas: A Prometheus histogram definition should look like this:

```
# HELP django_http_requests_body_total_bytes Histogram of requests by body size.
# TYPE django_http_requests_body_total_bytes histogram
django_http_requests_body_total_bytes_bucket{le="0.0"} 31794.0
django_http_requests_body_total_bytes_bucket{le="1.0"} 31794.0
(...)
django_http_requests_body_total_bytes_bucket{le="+Inf"} 31897.0
django_http_requests_body_total_bytes_count 31897.0
django_http_requests_body_total_bytes_sum 11935.0
```

Note that the metric name does not include `_bucket`. The type is `histogram`, not `counter` (the sum and count are also part of this metric type).

Marlinc: I now included a histogram type as well. As some of the metrics already get their help and type from pre-existing code I've simply added a histogram type now. Hope that works as expected.

pdns/dnsdistdist/docs/statistics.rst
```diff
@@ -99,6 +99,18 @@ latency-slow
 ------------
 Number of queries answered in more than 1 second.
 
+latency-sum
+-----------
+Total response time in milliseconds.
```
pieterlexis: It might make sense to mention the difference of this statistic in the time axis might say something about the change in latency. As latency usually is a gauge-type statistic over the last X seconds?

Marlinc: I changed the documentation of this statistic a bit. I'm not sure how to really explain this metric as I'm not entirely sure what Prometheus will be using it for (other than it having to do with histograms).

Marlinc:
I have now hopefully fixed the tests

pdns/dnsdistdist/docs/statistics.rst
wojas: You can use it to calculate the average latency over a period of time: https://prometheus.io/docs/practices/histograms/#count-and-sum-of-observations

pdns/dnsdist-web.cc
```diff
@@ -443,6 +443,24 @@ static void connectionThread(int sock, ComboAddress remote)
           output << "\n";
         }
 
+        // Latency histogram buckets
+        output << "# HELP dnsdist_latency Histogram of responses by latency\n";
+        output << "# TYPE dnsdist_latency histogram\n";
+        output << "# HELP dnsdist_latency_bucket Number of queries contributing to response time histogram per latency bucket\n";
+        output << "# TYPE dnsdist_latency_bucket counter\n";
```
rgacogne: I don't think we need "#HELP" and "#TYPE" lines for `dnsdist_latency_bucket`, based on https://github.com/prometheus/docs/blob/master/content/docs/instrumenting/exposition_formats.md#text-format-example

pdns/dnsdistdist/docs/statistics.rst
Marlinc: Updated the documentation

pdns/dnsdist-web.cc
Marlinc: Removed the type and help lines for `dnsdist_latency_bucket`

Marlinc:
Applied the last feedback

PR #7856, opened 2019-05-28
Don't initialize and deinitialize with OpenSSL >= 1.1

neheb:
Will not compile without deprecated APIs enabled.

Tested with released version of PowerDNS

### DIFF
--- a/pdns/dnsdistdist/libssl.cc
+++ b/pdns/dnsdistdist/libssl.cc
@@ -54,29 +54,23 @@ static void openssl_thread_cleanup()
   OPENSSL_free(openssllocks);
 }
 
-#else
-static void openssl_thread_setup()
-{
-}
-
-static void openssl_thread_cleanup()
-{
-}
-#endif /* (OPENSSL_VERSION_NUMBER < 0x1010000fL || defined LIBRESSL_VERSION_NUMBER) */
-
 static std::atomic<uint64_t> s_users;
+#endif /* (OPENSSL_VERSION_NUMBER < 0x1010000fL || defined LIBRESSL_VERSION_NUMBER) */
 
 void registerOpenSSLUser()
 {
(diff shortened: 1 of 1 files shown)


pdns/dnsdistdist/libssl.cc
```diff
@@ -53,31 +53,25 @@ static void openssl_thread_cleanup()
...
-static void openssl_thread_setup()
-{
-}
-
-static void openssl_thread_cleanup()
-{
-}
 #endif /* (OPENSSL_VERSION_NUMBER < 0x1010000fL || defined LIBRESSL_VERSION_NUMBER) */
 
 static std::atomic<uint64_t> s_users;
```
rgacogne: Would you mind moving the `#endif` after the `static std::atomic<uint64_t> s_users;` line, since we don't need `s_users` at all when OpenSSL >= 1.1.0 is in use?

pdns/dnsdistdist/libssl.cc
```diff
@@ -53,31 +53,25 @@ static void openssl_thread_cleanup()
...
 
 void registerOpenSSLUser()
 {
   if (s_users.fetch_add(1) == 0) {
+#if (OPENSSL_VERSION_NUMBER < 0x1010000fL || defined LIBRESSL_VERSION_NUMBER)
```
rgacogne: How about including the `if (s_users.fetch_add(1) == 0) {` line in that `#if`?

pdns/dnsdistdist/libssl.cc
```diff
@@ -53,31 +53,25 @@ static void openssl_thread_cleanup()
...
     SSL_load_error_strings();
     OpenSSL_add_ssl_algorithms();
     openssl_thread_setup();
+#endif
   }
 }
 
 void unregisterOpenSSLUser()
 {
   if (s_users.fetch_sub(1) == 1) {
```
rgacogne: Same here.

neheb:
Fixed.

PR #7859, opened 2019-05-28
auth: Remove disable-tcp option

pieterlexis:
### Short description
Also turns the TCPNameserver into a shared_ptr instead of a naked pointer.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/common_startup.hh
+++ b/pdns/common_startup.hh
@@ -46,7 +46,7 @@ extern CommunicatorClass Communicator;
 extern std::shared_ptr<UDPNameserver> N;
 extern vector<std::shared_ptr<UDPNameserver> > g_udpReceivers;
 extern int avg_latency;
-extern TCPNameserver *TN;
+extern std::unique_ptr<TCPNameserver> TN;
 extern ArgvMap & arg( void );
 extern void declareArguments();
 extern void declareStats();
--- a/pdns/receiver.cc
+++ b/pdns/receiver.cc
@@ -590,8 +590,7 @@ int main(int argc, char **argv)
       }
     }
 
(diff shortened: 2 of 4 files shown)


pdns/common_startup.hh
```diff
@@ -46,7 +46,7 @@ extern CommunicatorClass Communicator;
 extern std::shared_ptr<UDPNameserver> N;
 extern vector<std::shared_ptr<UDPNameserver> > g_udpReceivers;
 extern int avg_latency;
-extern TCPNameserver *TN;
+extern std::shared_ptr<TCPNameserver> TN;
```
rgacogne:
```suggestion
extern std::unique_ptr<TCPNameserver> TN;
```

pieterlexis: I made it a `shared_ptr` as the `UDPNameserver` is also a `shared_ptr`, but it might indeed be unique

rgacogne: I haven't looked into why but `UDPNameserver` objects are apparently also stored in a vector so they can't be unique, but `TCPNameserver` ones should be fine.

pdns/receiver.cc
```diff
@@ -590,8 +590,7 @@ int main(int argc, char **argv)
       }
     }
 
-    if(!::arg().mustDo("disable-tcp"))
-      TN=new TCPNameserver; 
+    TN = std::make_unique<TCPNameserver>();
```
rgacogne: C++11 doesn't provide `make_unique` in the `std` namespace, we do have our own `make_unique` in `misc.hh`, though.

PR #7861, opened 2019-05-28
Check if -latomic is needed instead of hardcoding

neheb:
ARM XScale, ARM 926EJ-S, and ARM FA526

Logs: https://downloads.openwrt.org/snapshots/faillogs/arm_arm926ej-s/packages/dnsdist/compile.txt
https://downloads.openwrt.org/snapshots/faillogs/arm_fa526/packages/dnsdist/compile.txt
https://downloads.openwrt.org/snapshots/faillogs/arm_xscale/packages/dnsdist/compile.txt


James-TR:
There's been additional discussion in the OpenWRT repo which might be useful as an alternate way to achieve this https://github.com/openwrt/packages/pull/9090#issuecomment-496649653

neheb:
Changed to alternate solution. Please review.

rgacogne:
Unrelated failure on Travis (perhaps https://github.com/PowerDNS/pdns/pull/7865), restarted.

PR #7864, opened 2019-05-31
Fix warnings reported by Coverity

rgacogne:
### Short description

Nothing too serious, and some changes are even just there to make the code more readable. We might want to backport most of these changes to 4.2.x (auth and rec).

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/nsecrecords.cc
+++ b/pdns/nsecrecords.cc
@@ -29,7 +29,9 @@ class NSECBitmapGenerator
 public:
   NSECBitmapGenerator(DNSPacketWriter& pw_): pw(pw_)
   {
+    memset(res, 0, 34);
   }
+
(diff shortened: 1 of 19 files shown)


pdns/nsecrecords.cc
```diff
@@ -29,7 +29,9 @@ class NSECBitmapGenerator
 public:
   NSECBitmapGenerator(DNSPacketWriter& pw_): pw(pw_)
   {
+    memset(res, 0, 34);
```
omoerbeek: I prefer `sizeof(res)`
There's another occurrence of this memset.

rgacogne: Done.

PR #7869, opened 2019-06-03
auth LUA: optionally reuse Lua state

Habbie:
### Short description
Nonscientific measurements:
* 60kqps from a plain A records
* 5kqps from `LUA A "'192.0.2.1'"`
* 35kqps from that same LUA with this patch and `enable-lua-records=shared`

Needs docs, maybe a test?

We could (presumably) win even more if we avoided doing 22 `writeFunction` calls for each LUA query still, but because of the captured variables, fixing that is not trivial. ~(After I just talked to @ahupowerdns  I'd like to try anyway)~ we'll try that for 4.3

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


Habbie:
User reports positive results!

PR #7870, opened 2019-06-04
stubquery: Fix handling of optional type arg.

omoerbeek:
### Short description

Currently: 
```
[otto@x1:3]$ stubquery www.google.com
Segmentation fault (core dumped) 
```

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/stubquery.cc
+++ b/pdns/stubquery.cc
@@ -38,10 +38,12 @@ try
     }
   }
 
-  if(argc < 2) {
+  if(argc <= 1) {
     usage();
     exit(EXIT_FAILURE);
   }
+
+  string type(argc == 2 ? "A" : argv[2]);
 
   ::arg().set("resolver","Use this resolver for ALIAS and the internal stub resolver")="no"; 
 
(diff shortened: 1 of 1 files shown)


pdns/stubquery.cc
```diff
@@ -38,19 +38,21 @@ try
     }
   }
 
-  if(argc < 2) {
+  if(argc <= 1) {
     usage();
     exit(EXIT_FAILURE);
   }
 
+  string type(argc == 2 ? "A" : argv[2]);
```
Habbie:
```suggestion
  string type(argc >= 2 ? "A" : argv[2]);
```

omoerbeek: I must be missing something. We want to take arg[2] if it is given.

Habbie: Oh, my logic is reversed! I was thinking about what if we later add more arguments (like sdig flags, although we should really redo those in a better way).

My suggestion should have been `<= 2`, I suppose, but let's drop it.

PR #7871, opened 2019-06-04
Adapt calidns for openbsd and other systems without rcvmmsg(2)

omoerbeek:
### Short description

Implement receiving using `recvmsg(2)` and guard `sched_setscheduler(2)`.
Also, a non-null control msg of length 0 is illegal on OpenBSD.

I tried tried to avoid reformatting too much, formatting is a broken in a lot of places.

The error handling in the `poll` loop was broken afaiks, so I fixed that as well.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -130,9 +130,11 @@ PDNS_CHECK_VIRTUALENV
 PDNS_FROM_GIT
 
 dnl Checks for library functions.
-AC_CHECK_FUNCS_ONCE([strcasestr localtime_r recvmmsg getrandom arc4random])
+AC_CHECK_FUNCS_ONCE([strcasestr localtime_r recvmmsg sched_setscheduler getrandom arc4random])
 
 AM_CONDITIONAL([HAVE_RECVMMSG], [test "x$ac_cv_func_recvmmsg" = "xyes"])
+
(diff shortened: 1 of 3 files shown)


Habbie:
This makes calidns work on macOS 10.12

configure.ac
```diff
@@ -130,10 +130,12 @@ PDNS_CHECK_VIRTUALENV
 PDNS_FROM_GIT
 
 dnl Checks for library functions.
-AC_CHECK_FUNCS_ONCE([strcasestr localtime_r recvmmsg getrandom arc4random])
+AC_CHECK_FUNCS_ONCE([strcasestr localtime_r recvmmsg sched_setscheduler getrandom arc4random])
 
 AM_CONDITIONAL([HAVE_RECVMMSG], [test "x$ac_cv_func_recvmmsg" = "xyes"])
 
+AM_CONDITIONAL([HAVE_SCHED_SETSCHEDULER], [test "x$ac_cv_func_sched_setscheduler" = "xyes"])
```
pieterlexis: I think you can remove this line. The automake conditional is not used (and not needed). `AC_CHECK_FUNCS_ONCE` will define `HAVE_SCHED_SETSCHEDULER` in config.h

PR #7872, opened 2019-06-04
LUA record documentation (version number, shared model)

Habbie:
### Short description
I enthousiastically merged #7869 without docs. This PR rectifies that situation.

### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/docs/lua-records/index.rst
+++ b/docs/lua-records/index.rst
@@ -197,6 +197,20 @@ Note that to protect operators, support for LUA records must be enabled
 explicitly, either globally (``enable-lua-records``) or per zone
 (``ENABLE-LUA-RECORDS`` = 1).
 
+.. _lua-records-shared-state:
+
+Shared Lua state model
+----------------------
+
+The default mode of operation for LUA records is to create a fresh Lua state for every query that hits a LUA record.
+This way, different LUA records cannot accidentally interfere with each other, by leaving around global objects, or perhaps even deleting relevant functions.
+However, creating a Lua state (and registering all our functions for it, see Reference below) takes measurable time.
+For users that are confident they can write Lua scripts that will not interfere with eachother, a mode is supported where Lua states are created on the first query, and then reused forever.
+In non-scientific testing this has yielded up to 10x QPS increases.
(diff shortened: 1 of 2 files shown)


docs/settings.rst
```diff
@@ -596,7 +596,11 @@ Enables EDNS subnet processing, for backends that support it.
 -  Boolean
```
pieterlexis: not a boolean :)

I'd say 'string', 'ne of: no, yes, shared'

Habbie: amended

docs/lua-records/index.rst
```diff
@@ -197,6 +197,20 @@ Note that to protect operators, support for LUA records must be enabled
 explicitly, either globally (``enable-lua-records``) or per zone
 (``ENABLE-LUA-RECORDS`` = 1).
 
+.. _lua-records-shared-state:
+
+Shared Lua state model
+----------------------
+
+The default mode of operation for LUA records is to create a fresh Lua state for every query that hits a LUA records.
```
franklouwers: a LUA record~~s~~

docs/lua-records/index.rst
```diff
@@ -197,6 +197,20 @@ Note that to protect operators, support for LUA records must be enabled
...
+This way, different LUA records cannot accidentally interfere with eachother, by leaving around global objects, or perhaps even deleting relevant functions.
```
franklouwers: ~~eachother~~ each other

Habbie:
suggestions from @franklouwers commited and squashed

mnordhoff:
It creates (up to) one state per thread and you still wind up with a handful of states, right? It might be worth mentioning that explicitly, so someone doesn't think there's precisely one state **total** and try to make a global counter or something.

Habbie:
> It creates (up to) one state per thread and you still wind up with a handful of states, right? It might be worth mentioning that explicitly, so someone doesn't think there's precisely one state **total** and try to make a global counter or something.

pushed

PR #7882, opened 2019-06-07
rec fix racy negcache test

omoerbeek:
This test is racey, if a sig is only valid for the current sec that might have passed once we get to the cache check. Also fix comment.

I can make the test always fail by using 0 as an arg to `addRRSIG`.

### Checklist

I have:
- [X read the [CONTRIBUTING.md](https://github.com/PowerDNS/pdns/blob/master/CONTRIBUTING.md) document
- [X] compiled this code
- [X] tested this code
- [x] documented the code


omoerbeek:
I made a mess of this.

PR #7883, opened 2019-06-07
rec fix racy negcache test

omoerbeek:
This test is racey, if a sig is only valid for the current sec a tick might have happened once we get to the cache check.
Also fix comment.

I could make this test fail always if I put in 0 as arg to `addRRSIG`

### Checklist

I have:
- [X] compiled this code
- [X ] tested this code
- [X] documented the code


rgacogne:
I know what's going on, but this fix would only hide the real issue which is that we mix `time()` and `gettimeofday()` (see #6160). I'll open a PR with what I think is the right fix.

omoerbeek:
Obsolete

PR #7884, opened 2019-06-07
rec: Don't mix time() and gettimeofday() in our unit tests (again)

rgacogne:
### Short description

We used to set time returned by `sr->getNow()` via `gettimeofday()` and later the one used to compute the RRSIG validity via `time(nullptr)`. It turns out that the second call can return a value lesser than the first one, without the time going backward, making our tests fail.
After this commit we base our calculations on the same time value, so it should not fail randomly anymore.

Yes another occurrence of #6160 and #7235, but hopefully the last remaining one.
Supersedes #7883.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


omoerbeek:
This look like a better fix indeed and in line wit the other cases.

PR #7886, opened 2019-06-07
 SuffixMatchTree: Fix root removal, partial match of non-leaf nodes

rgacogne:
### Short description

If we insert fr and www.domain.fr, domain.fr should not match www.domain.fr, the leaf node, and it should not match the intermediary domain.fr node, but it should match the fr end-node.
It should also be possible to remove the root from the tree.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsname.hh
+++ b/pdns/dnsname.hh
@@ -336,11 +336,7 @@ struct SuffixMatchTree
       return nullptr;
     }
 
-    auto result = lookup(name.getRawLabels());
-    if (result) {
-      return result;
-    }
-    return endNode ? &d_value : nullptr;
+    return lookup(name.getRawLabels());
   }
 
   T* lookup(std::vector<std::string> labels) const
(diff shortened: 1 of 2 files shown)


pdns/dnsname.hh
```diff
@@ -328,28 +333,37 @@ struct SuffixMatchTree
     if(children.empty()) { // speed up empty set
       if(endNode)
         return &d_value;
-      return 0;
+      return nullptr;
     }
-    return lookup(name.getRawLabels());
+
```
omoerbeek: Wondering if this is neccesary. I would think that this case is handled by `lookup(std::vector<string>)`

pdns/dnsname.hh
```diff
@@ -299,6 +299,11 @@ struct SuffixMatchTree
    */
   void remove(std::vector<std::string> labels) const
   {
```
omoerbeek: Not related to this dif, but would it be better to make the argument type `const std::vector<std::string> &`?

pdns/dnsname.hh
rgacogne: Good catch, I don't think this is necessary! I'll update this PR.

PR #7891, opened 2019-06-07
auth API, pdnsutil: improve backend transaction correctness 

Habbie:
### Short description
We do deletes+inserts outside of transactions. This makes it possible to briefly observe NXDOMAIN while replacing RRsets. With this PR, most operations are correctly wrapped in transactions.

### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/backends/gsql/gsqlbackend.cc
+++ b/pdns/backends/gsql/gsqlbackend.cc
@@ -1349,6 +1349,10 @@ bool GSQLBackend::replaceRRSet(uint32_t domain_id, const DNSName& qname, const Q
   try {
     reconnectIfNeeded();
 
+    if (!d_inTransaction) {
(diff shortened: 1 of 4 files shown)


Habbie:
I asked on IRC 'is there a reason that ws-auth.cc apiZoneMetadataKind uses B instead of di.backend?'. The answer arrived just after I PRed this: metadata can live in a different backend than the zone itself. This will need some work.

Habbie:
> I asked on IRC 'is there a reason that ws-auth.cc apiZoneMetadataKind uses B instead of di.backend?'. The answer arrived just after I PRed this: metadata can live in a different backend than the zone itself. This will need some work.

Pushed.

PR #7892, opened 2019-06-07
auth: auto-generate lua.hpp

### LINKED ISSUE #4959: autogenerate lua.hpp for auth as well
- Program: Authoritative
 - Issue type: Enhancement

### Short description
in dnsdist and recursor, if lua.hpp is missing on the local system, we generate our own, by including lua_hpp.mk in the Makefile. For auth, we do not do this yet. We should.

James-TR:
### Short description
On systems without an available lua.hpp, the builds for the lua backends and pdns-server
will fail. To fix this, we can auto-generate the file at build time. This fixes #4959

Only issue I can see with this is including the same code in multiple places. I will also test including this in the root Makefile and simply employing symlinks instead and see how that works.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code

### DIFF
--- a/modules/lua2backend/Makefile.am
+++ b/modules/lua2backend/Makefile.am
@@ -9,5 +9,15 @@ liblua2backend_la_SOURCES = \
 	lua2backend.cc lua2backend.hh \
 	lua2api2.hh lua2api2.cc
 
+if !HAVE_LUA_HPP
+BUILT_SOURCES = lua.hpp
+nodist_liblua2backend_la_SOURCES = lua.hpp
+CLEANFILES = lua.hpp
+endif
+
 liblua2backend_la_LDFLAGS = -module -avoid-version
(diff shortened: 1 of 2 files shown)


jsoref:
Hi, for the time being (and hopefully going forward), pdns master is running the [check-spelling](https://github.com/marketplace/actions/check-spelling) action. In order to avoid a ❌ when merged https://github.com/jsoref/pdns/commit/d0d81fca29aa79ec3b6398d7787d6c2747bba5d1#commitcomment-39230815 -- you'll need to merge/rebase on `master` to get the check (and the files) -- you can accept the advice (which should add `libluabackend` to `expect.txt`)

pieterlexis:
This PR is no longer needed, all OS's we care about ship this hpp. The issue has been updated to reflect this.

PR #7901, opened 2019-06-11
rec: Reuse the outgoing query protobuf for the incoming response

rgacogne:
### Short description

Allocating memory for the ProtoBuf objects an their fields is costly, as is converting the qname and potential UUID to strings, for example.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/lwres.cc
+++ b/pdns/lwres.cc
@@ -118,48 +118,46 @@ static void logFstreamResponse(const std::shared_ptr<std::vector<std::unique_ptr
 
 #endif // HAVE_FSTRM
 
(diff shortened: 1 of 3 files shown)


rgacogne:
CircleCI failure seems unrelated, probably https://github.com/PowerDNS/pdns/pull/7899 ?

Habbie:
> CircleCI failure seems unrelated, probably #7899 ?

yes

rgacogne:
> Only logging answers an no replies is not possible.

I noticed that as well and was a bit surprised about it because we do support it for the logging of incoming queries and outgoing responses, but I considered it was out of scope for this PR.

PR #7902, opened 2019-06-11
dnsdist TLS: remove unused doHandshake methods

Habbie:
### Short description
This code appears to be unused.

### Checklist

I have:
- [x] compiled this code
- [ ] tested this code
- [ ] documented the code


rgacogne:
I'm a bit torn about this PR. I'm usually all for removing dead code but in that case I'm wondering whether we will need this if we ever intend on adding DoT in the authoritative server, for example.
Not that the `::read()` and `::write()` methods are also currently unused but I'd like to keep them around because I plan to eventually add DoT support to `sdig`.

Habbie:
Right! closing :)

PR #7904, opened 2019-06-11
 rec: Optimize for large number of filtering policies, empty sections

rgacogne:
### Short description

The existing code was optimized for very larger filtering policies (RPZ) but was duplicating a lot of work when a large number of policies was in use. This PR makes sure we don't copy `DNSName` objects and iterate over their labels one time for each policy.
It also check whether a policy really has rules for the type of lookup (qname, source IP, NS name, etc) before doing any processing, so we can return right away if there is no matching rules.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/filterpo.cc
+++ b/pdns/filterpo.cc
@@ -31,19 +31,9 @@ DNSFilterEngine::DNSFilterEngine()
 {
 }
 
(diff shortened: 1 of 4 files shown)


rgacogne:
CircleCI failure is unrelated.

PR #7802, opened 2019-05-09
Allow disabling packet cache with max-packet-cache-entries=0

NameWeb:
### Short description

In some configurations (dnsdist in front of pdns?) a packet cache might hurt performance. I didn't find an option to actually disable this. Setting max-packet-cache-entries=0 will now disable writing to it or reading from it.

Performance gain of well over 10% in usage case where packet cache wasn't being hit.
### Checklist

I have:
- [X] read the [CONTRIBUTING.md](https://github.com/PowerDNS/pdns/blob/master/CONTRIBUTING.md) document
- [X] compiled this code
- [X] tested this code
- [ ] included documentation (including possible behaviour changes)
- [ ] documented the code
- [ ] added or modified regression test(s)
- [ ] added or modified unit test(s)

PR #7910, opened 2019-06-12
auth: completely disable the packet when cache-ttl=0

mind04:
This was inspired by #7802 but is more in line with the current query cache behaviour

closes #7802

### Checklist
I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -180,8 +180,7 @@ Also AXFR a zone from a master with a lower serial.
 -  Integer
 -  Default: 20
 
-Seconds to store packets in the :ref:`packet-cache`. A value of 0 will disable
-the cache.
+Seconds to store packets in the :ref:`packet-cache`. A value of 0 will disable the cache.
 
 .. _setting-carbon-instance:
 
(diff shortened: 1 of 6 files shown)


docs/settings.rst
```diff
@@ -180,7 +180,8 @@ Also AXFR a zone from a master with a lower serial.
 -  Integer
 -  Default: 20
 
-Seconds to store packets in the :ref:`packet-cache`.
+Seconds to store packets in the :ref:`packet-cache`. A value of 0 will disable
```
Habbie: Please use one line per sentence. We don't do that everywhere yet, but we should :)

Habbie:
ci/circleci: test-auth-regress-odbc-mssql failure can be ignored.

pdns/packethandler.cc
```diff
@@ -1546,8 +1546,7 @@ DNSPacket *PacketHandler::doQuestion(DNSPacket *p)
     if(doSigs)
       addRRSigs(d_dk, B, authSet, r->getRRS());
       
-    r->wrapup(); // needed for inserting in cache
```
Habbie: was this comment wrong?

mind04: There is no need to do it here getString() will warpup when necessary https://github.com/powerdns/pdns/blob/master/pdns/dnspacket.cc#L63

PR #7926, opened 2019-06-13
Update secpoll for rec-4.1.14 release.

aerique:


aerique:
> Current SOA is already 2019061301, please update to 2019061302

How did that happen?  HEAD from `master` still shows 2019060601: https://github.com/PowerDNS/pdns/blob/master/docs/secpoll.zone#L1

(Will update regardless.)

PR #7927, opened 2019-06-13
dnsdist: Proper HTTP response for timeouts over DoH

rgacogne:
### Short description

Timeouts, actively discovered or not, were not properly handled for DoH queries.
This PR is based on top of #7917 so we can send a proper HTTP status code, it will need to be rebased.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdist.cc
+++ b/pdns/dnsdist.cc
@@ -570,6 +570,7 @@ try {
           continue;
         }
 
+        bool isDoH = du != nullptr;
         /* atomically mark the state as available */
         int oldFD = ids->origFD.exchange(-1);
         if (oldFD == origFD) {
@@ -630,7 +631,7 @@ try {
 
         double udiff = ids->sentTime.udiff();
         vinfolog("Got answer from %s, relayed to %s%s, took %f usec", dss->remote.toStringWithPort(), ids->origRemote.toStringWithPort(),
-                 ids->du ? " (https)": "", udiff);
+                 isDoH ? " (https)": "", udiff);
 
         struct timespec ts;
         gettime(&ts);
--- a/pdns/dnsdist.cc
+++ b/pdns/dnsdist.cc
@@ -546,9 +546,17 @@ try {
         IDState* ids = &dss->idStates[queryId];
         int origFD = ids->origFD;
 
-        if(origFD < 0) // duplicate
+        if(origFD < 0) {
+          /* the corresponding state is marked as not in use, meaning that:
+             - it was already reused by another thread and the state is gone ;
+             - we already got a response for this query and this one is a duplicate.
+             Either way, we don't touch it.
+          */
           continue;
+        }
 
+        /* read the potential DOHUnit state as soon as possible, but don't use it
+           until we have confirmed that we own this state by updating origFD */
         auto du = ids->du;
         /* setting age to 0 to prevent the maintainer thread from
            cleaning this IDS while we process the response.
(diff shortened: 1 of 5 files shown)


rgacogne:
Fixed the use-after-free race reported by ASAN on Circle CI.

rgacogne:
Rebased now that https://github.com/PowerDNS/pdns/pull/7917 has been merged.

pdns/dnsdist.cc
```diff
@@ -588,15 +592,15 @@ try {
...
-          if (ids->du) {
+          if (du) {
 #ifdef HAVE_DNS_OVER_HTTPS
             // DoH query
-            ids->du->query = std::string(response, responseLen);
-            if (send(ids->du->rsock, &ids->du, sizeof(ids->du), 0) != sizeof(ids->du)) {
-              delete ids->du;
+            du->response = std::string(response, responseLen);
+            if (send(du->rsock, &du, sizeof(du), 0) != sizeof(du)) {
+              delete du;
```
omoerbeek: Hmm, is `ids->du` stll the same pointer? That would now be dangling...

pdns/dnsdist.cc
```diff
@@ -1564,16 +1568,25 @@ static void processUDPQuery(ClientState& cs, LocalHolders& holders, const struct
...
+      du = ids->du;
+    }
 
     int oldFD = ids->origFD.exchange(cs.udpFD);
     if(oldFD < 0) {
-      // if we are reusing, no change in outstanding
       ++ss->outstanding;
+      /* either it was already -1 so no DOH unit to handle,
+         or someone handled it before us */
+      du = nullptr;
```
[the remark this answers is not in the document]
rgacogne: It is a dead assignment today, but it prevents accessing it by mistake later.

pdns/dnsdistdist/doh.cc
```diff
@@ -122,6 +126,22 @@ struct DOHServerConfig
...
+    return;
+  }
+
+/* we are about to erase an existing DU */
+  oldDU->error = true;
+  oldDU->status_code = 502;
+
+  if (send(oldDU->rsock, &oldDU, sizeof(oldDU), 0) != sizeof(oldDU)) {
+    delete oldDU;
+    oldDU = nullptr;
```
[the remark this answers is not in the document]
rgacogne: same

pdns/dnsdist.cc
rgacogne: Unless I'm mistaken, the only case where `du` is not a `NULL` pointer is when `ids->origFD` was still equal to `oldFD` after we atomically updated it, meaning that the `oldFD == origFD` condition on line 566 was true and therefore `ids->du` was set to `nullptr`.

pdns/dnsdist.cc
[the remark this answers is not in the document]
omoerbeek: true, agree

pdns/dnsdist.cc
omoerbeek: A comment and/or an assert on that `ids->du == nullptr` condition would be nice

omoerbeek:
I must say I don't grasp all of dnsdist yet, but this PR look ok (with the one comment on the nullptr case).

pdns/dnsdist.cc
rgacogne: I pushed a bunch of comments about our `IDState`s object and in particular its `origFD` member since it's probably the hardest part of dnsdist to grasp. We can't add an assertion because it would be racy, the state might have already been reused by another thread at that point.

omoerbeek: 👍

rgacogne:
Giving up on the LGTM check, it is simply taking too long.

PR #7928, opened 2019-06-14
rec: Better detection of Bogus zone cuts for DNSSEC validation

rgacogne:
### Short description

`SyncRes::getDSRecords()` used to not set the `foundCut` flag as soon as the processing returned a `Bogus` state. In some situations, for example if the time was completely off and the `RRSIG` from the root zone were thus invalid, this wrong zone cut determination lead to validation issues.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/syncres.cc
+++ b/pdns/syncres.cc
@@ -1784,6 +1784,10 @@ vState SyncRes::getDSRecords(const DNSName& zone, dsmap_t& ds, bool taOnly, unsi
 
     if (rcode == RCode::NoError) {
       if (ds.empty()) {
+        /* we have no DS, it's either:
+           - a delegation to a non-DNSSEC signed zone
+           - no delegation, we stay in the same zone
+        */
         if (gotCNAME || denialProvesNoDelegation(zone, dsrecords)) {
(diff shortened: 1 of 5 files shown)


rgacogne:
CircleCI failure is unrelated.

pdns/syncres.cc
```diff
@@ -1747,58 +1747,63 @@ vState SyncRes::getDSRecords(const DNSName& zone, dsmap_t& ds, bool taOnly, unsi
...
 
+        /* delegation with no DS, might be Secure -> Insecure */
+        if (foundCut) {
           *foundCut = true;
         }
 
-        return Insecure;
-      } else if (foundCut && rcode == RCode::NoError && !ds.empty()) {
-        *foundCut = true;
+        return state == Secure ? Insecure : state;
```
omoerbeek: This line can do with a comment

rgacogne: I added some comments, let me know if that clarifies it for you!

omoerbeek: Yes, nice.

PR #7931, opened 2019-06-14
rec backport to 4.2.x: Don't mix time() and gettimeofday() in our unit tests (again)

omoerbeek:
We used to set time returned by `sr->getNow()` via `gettimeofday()`
and later the one used to compute the RRSIG validity via
`time(nullptr)`. It turns out that the second call can return a
value lesser than the first one, without the time going backward,
making our tests fail.
After this commit we base our calculations on the same time value,
so it should not fail randomly anymore.

(cherry picked from commit d0add506df4075aa8a7e2eb7c66f58ea850ff6d4)

### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code


omoerbeek:
including a test of a modified cherry-pick-pr.py

Issue #7358, opened 2019-01-09
dnsdist Carbon multiple frontend counters


nb71:
- Program: dnsdist 
 - Issue type: Bug report 

### Short description

Frontend queries metrics are send to carbon with the same name when using multiple listeners
(addLocal). 
Carbon overwrites the same counter multiple times, so its imposible to see individual front end behavior o summarize.
Carbon capture (with 3 frontends):
dnsdist.dnsdist1.main.frontends.1_8_8_8:53_udp.queries 314529726 1547057845.
dnsdist.dnsdist1.main.frontends.1_8_8_8:53_udp.queries 312262776 1547057845.
dnsdist.dnsdist1.main.frontends.1_8_8_8:53_udp.queries 313196027 1547057845.
dnsdist.dnsdist1.main.frontends.1_8_8_8:53_tcp.queries 407090 1547057845.
dnsdist.dnsdist1.main.frontends.1_8_8_8:53_tcp.queries 407732 1547057845.
dnsdist.dnsdist1.main.frontends.1_8_8_8:53_tcp.queries 405958 1547057845.

### Environment

 - Operating system:  Any
 - Software version:  dnsdist 1.3.2 (Lua 5.1.4 [LuaJIT 2.0.4])
 - Software source: Power DNS repo 

### Steps to reproduce

1. Configure multiple listeners (setLocal(), addLocal(), addLocal()
2. Configure Carbon: dnsdist.conf:carbonServer("1.2.3.4:2023", "dnsdist1", 10)
3. Capture metrics send 

### Expected behaviour

The frontend counters should have a suffix to differentiate each one of them.
The counters are OK, but being the same name they are overwritten by carbon instead of 
saving a different counter for each frontend.

### Actual behaviour

dnsdist.dnsdist1.main.frontends.1_8_8_8:53_udp.queries 314529726 1547057845.
dnsdist.dnsdist1.main.frontends.1_8_8_8:53_udp.queries 312262776 1547057845.
dnsdist.dnsdist1.main.frontends.1_8_8_8:53_udp.queries 313196027 1547057845.
dnsdist.dnsdist1.main.frontends.1_8_8_8:53_tcp.queries 407090 1547057845.
dnsdist.dnsdist1.main.frontends.1_8_8_8:53_tcp.queries 407732 1547057845.
dnsdist.dnsdist1.main.frontends.1_8_8_8:53_tcp.queries 405958 1547057845.

### Other information

[...]

PR #7934, opened 2019-06-14
dnsdist: Deduplicate frontends entries with carbon and prometheus

rgacogne:
### Short description

This is a work-around, and not a very pretty one, to avoid messing up our carbon/prometheus metrics when we add the same frontends several times with `addLocal()`, `addTLSLocal()` or `addDOHLocal()` and `reuseport`.
The proper fix would be to be able to provide the number of threads directly as a parameter to the `add*Local()` and provide aggregated metrics (but also per-thread metrics, at least for the number of queries, so we can check that the distribution done by the kernel is working correctly), but this would require an important refactoring and we are too late in the 1.4.0 release process for that IMHO.

We will to implement the deduplication for the DoH metrics in prometheus as well, either in this PR if https://github.com/PowerDNS/pdns/pull/7933 is merged first, or in #7933 otherwise.

Fixes https://github.com/PowerDNS/pdns/issues/7358.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdist-carbon.cc
+++ b/pdns/dnsdist-carbon.cc
@@ -104,12 +104,20 @@ try
           str<<base<<"tcpavgqueriesperconnection" << ' '<< state->tcpAvgQueriesPerConnection.load() << " " << now << "\r\n";
           str<<base<<"tcpavgconnectionduration" << ' '<< state->tcpAvgConnectionDuration.load() << " " << now << "\r\n";
         }
+
+        std::map<std::string,uint64_t> frontendDuplicates;
         for(const auto& front : g_frontends) {
           if (front->udpFD == -1 && front->tcpFD == -1)
             continue;
 
           string frontName = front->local.toString() + ":" + std::to_string(front->local.getPort()) +  (front->udpFD >= 0 ? "_udp" : "_tcp");
           boost::replace_all(frontName, ".", "_");
+          auto dupPair = frontendDuplicates.insert({frontName, 1});
+          if (dupPair.second == false) {
+            frontName = frontName + "_" + std::to_string(dupPair.first->second);
+            ++dupPair.first->second;
+          }
+
           const string base = namespace_name + "." + hostname + "." + instance_name + ".frontends." + frontName + ".";
           str<<base<<"queries" << ' ' << front->queries.load() << " " << now << "\r\n";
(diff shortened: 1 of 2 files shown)


pdns/dnsdist-carbon.cc
```diff
@@ -104,12 +104,20 @@ try
...
+
+        std::map<std::string,uint64_t> frontendDuplicates;
         for(const auto& front : g_frontends) {
           if (front->udpFD == -1 && front->tcpFD == -1)
             continue;
 
           string frontName = front->local.toString() + ":" + std::to_string(front->local.getPort()) +  (front->udpFD >= 0 ? "_udp" : "_tcp");
           boost::replace_all(frontName, ".", "_");
+          auto dupPair = frontendDuplicates.insert({frontName, 1});
+          if (dupPair.second == false) {
```
pieterlexis:
```suggestion
          if (!dupPair.second) {
```

pdns/dnsdist-carbon.cc
```diff
@@ -104,12 +104,20 @@ try
...
+            frontName = frontName + "_" + std::to_string(dupPair.first->second);
+            ++dupPair.first->second;
```
pieterlexis: This could use some braces for clarity :)

pdns/dnsdist-carbon.cc
```diff
@@ -104,12 +104,20 @@ try
           str<<base<<"tcpavgqueriesperconnection" << ' '<< state->tcpAvgQueriesPerConnection.load() << " " << now << "\r\n";
           str<<base<<"tcpavgconnectionduration" << ' '<< state->tcpAvgConnectionDuration.load() << " " << now << "\r\n";
         }
+
+        std::map<std::string,uint64_t> frontendDuplicates;
```
pieterlexis: Perhaps `typedef` this?

pdns/dnsdist-carbon.cc
rgacogne: Done!

pdns/dnsdist-carbon.cc
rgacogne: Done!

pdns/dnsdist-carbon.cc
rgacogne: I'm not sure that would make the code more readable, since it's a quite simple type, and you would then have to lookup the typedef definition if you want to known what the actual type is.

Issue #7894, opened 2019-06-07
DoH serverTokens: duplicate server header and ignored setting


appliedprivacy:
- Program: dnsdist
 - Issue type: Bug report

### Short description

There are actually 3 issues but since they are closely related we'll document them in one go:

**a) ignored serverTokens value**

Setting the serverTokens option via 
`addDOHLocal( ..., "/query", { serverTokens='abc'})`

has no effect on the HTTP `server` header field for HTTP requests not going to `/query`.

**b) duplicate HTTP `server` header**

Setting the serverTokens option via 
`addDOHLocal( ..., "/query", { serverTokens='abc'})`

causes the server to send multiple HTTP `server` headers:

```
HTTP/2.0 200 OK
server: h2o/2.2.5
date: Fri, 07 Jun 2019 18:17:52 GMT
server: abc
content-type: application/dns-message
content-length: 80
X-Firefox-Spdy: h2
```

when the client issued a valid DoH request.

**c) documentation issue** 

The documentation says the default value for serverTokens is `h2o/dnsdist`
but it is actually `h2o/<h2oversion>` currently that is `h2o/2.2.5`

### Environment
 - Operating system: Debian Buster
 - Software version: 1.4.0~beta1-1pdns.buster
 - Software source: PowerDNS repository

### Steps to reproduce

**a)**
`curl -si https://doh.appliedprivacy.net`

**b)**
`curl -si "https://doh.appliedprivacy.net/query?dns=l1sBAAABAAAAAAAAA3d3dw1rbm90LXJlc29sdmVyAmN6AAAcAAE"`

**c)**
https://dnsdist.org/reference/config.html#addDOHLocal

### Expected behaviour

* dnsdist should set the HTTP `server` header to `h2o/dnsdist` by default (as per documentation)
* dnsdist should set the HTTP `server` header to the value provided in the `serverTokens` option (no matter which URL the client requested)
* dnsdist should not send multiple HTTP `server` header

### Actual behaviour

* dnsdist ignores the `serverTokens` option complete when answering HTTP requests that are not valid DoH requests
* dnsdist sends multiple HTTP `server` header when the request is a valid DoH request and the configuration contains the `serverTokens` option

rgacogne:
Thank you for reporting this!
The a/ issue is clearly an oversight on my part, I guess we could make sure that we route all queries by registering "/".
[...]

PR #7938, opened 2019-06-16
Dnsdist/server token test

melissavoegeli:
### Short description
I am learning this codebase and decided before I make a real change, that I will add a test to an area I am eventually going to make modifications.  I did this to A) convince myself that I know how these tests work and B) see how close I get to the community code expectations.

I am not familiar with good python practices, so please let me know if anything is fucked up. In particular I didn't want to add another return value to sendDOHQuery because that would be more code changes.  But I think my implemented solution could be confusing.

### Checklist

I have:
- [x] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/regression-tests.dnsdist/test_DOH.py
+++ b/regression-tests.dnsdist/test_DOH.py
@@ -53,7 +53,6 @@ class DNSDistDOHTest(DNSDistTest):
         message = None
         cls._response_headers = ''
         data = conn.perform_rb()
-
         rcode = conn.getinfo(pycurl.RESPONSE_CODE)
         if rcode == 200:
             message = dns.message.from_wire(data)
--- a/regression-tests.dnsdist/test_DOH.py
+++ b/regression-tests.dnsdist/test_DOH.py
@@ -5,7 +5,6 @@ import clientsubnetoption
 from dnsdisttests import DNSDistTest
 
 import pycurl
-import re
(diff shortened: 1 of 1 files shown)


melissavoegeli:
> Thanks for this PR! It looks good to me, the only comment I can think of is that the regex matching would not detect a duplicated 'Server' header in the response, for example, but I think it's good enough that way.

Good point.  I also thought about how to test custom server token headers not being present for non explicitly determined paths.  (As mentioned in #7894)  But it required more changes and I felt bad.  I like your idea because it helps document the behavior a little bit more plainly in the test expectations.  I'll revise it.

rgacogne:
Note that the duplicated 'Server' header is a bug and will disappear once https://github.com/PowerDNS/pdns/pull/7911 is merged!

melissavoegeli:
Does that mean you prefer I change the test expectations or abandon this PR?

rgacogne:
> Does that mean you prefer I change the test expectations or abandon this PR?

Please don't abandon it! :-) Ideally I'd like you to rebase it on the current master, and check that you have only one 'Server:' header. But I'm also fine with just changing this after rebasing:

```
self.assertTrue((self._h2oServerToken) in self._response_headers)
```

to:

```
self.assertFalse((self._h2oServerToken) in self._response_headers)
```

PR #7939, opened 2019-06-17
rec backport to 4.2.x: : Compare the CacheKey type and place first then the name

omoerbeek:
We don't care about the actual ordering, so let's do the cheapest
comparisons first then the quite expensive DNSName comparison only
if we really need to.

(cherry picked from commit 186e99c98d195329af05129cd4d6512657edf4bc)

### Checklist

I have:
- [X] compiled this code
- [ ] tested this code
- [ ] documented the code


omoerbeek:
Branch created with modified cherry-pick-pr.py

PR #7941, opened 2019-06-17
rec: Add bindings to craft answers from the FFI version of gettag

rgacogne:
### Short description

This will not skip the creation of a MThread since we will need it to follow CNAME records, if any, and we don't want to worry about a full parsing of the query and records creation in gettag, but it prevents the need of calling 'preresolve' which has no FFI equivalent, as well as the need to carry data around from one hook to another.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


rgacogne:
Rebased to fix a conflict.

PR #7942, opened 2019-06-17
Add debug informations to the MPlexer unit tests, tracking a failure

rgacogne:
### Short description

I can't reproduce the failure on Arch nor stretch, but I have now seen it occur two times on our CI tests, the most recent one in CircleCI, so I would like to understand what is happening.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/test-mplexer.cc
+++ b/pdns/test-mplexer.cc
@@ -90,9 +90,9 @@ BOOST_AUTO_TEST_CASE(test_MPlexer) {
                         *calledPtr = true;
                 };
   mplexer->addReadFD(pipes[0],
-                      readCB,
-                      &readCBCalled,
-                      &ttd);
+                     readCB,
+                     &readCBCalled,
+                     &ttd);
 
   /* not ready for reading yet */
   readyFDs.clear();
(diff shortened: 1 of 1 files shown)


pdns/test-mplexer.cc
```diff
@@ -124,6 +124,25 @@ BOOST_AUTO_TEST_CASE(test_MPlexer) {
...
+    cerr<<"pipes[1] is "<<pipes[1]<<endl;
+    cerr<<"The file descripttor returned as ready is "<<readyFDs.at(0)<<endl;
+    char buffer[2];
+    ssize_t res = read(pipes[0], &buffer[0], sizeof(buffer));
+    cerr<<"Reading from pipes[0] returns "<<res<<endl;
+    if (res == -1) {
+      cerr<<"errno is "<<errno<<endl;
+    }
+    res = write(pipes[1], "0", 1);
+    cerr<<"Writing to pipes[1] returns "<<res<<endl;
```
Habbie: this might clobber `errno`? cc @omoerbeek

rgacogne: I'm not sure I understand how that would be a problem?

omoerbeek: A problem is that writing to `cerr` might clobber `errno`. Better save `errno` into a local var immediately after the syscall and use that value for printing.

rgacogne: I see. I pushed a commit saving `errno` before printing so that we still get the initial value in case printing somehow fails and clobbers it.

rgacogne: I see. I pushed a commit saving `errno` before printing so that we still get the initial value in case printing somehow fails and clobbers it.

PR #7943, opened 2019-06-17
Improve error when notification comes in for non-slave zone

ahupowerdns:
### Short description
If you neglected tos etup a zone as a slave, but received a notification, you'd get the following error:
```
Can't determine backend for domain 'example.com'
```
It personally took me a while to debug a problem because of this error (was something wrong with my backend?).

The new error mentions a zone not being a slave zone in this case.

### Checklist

I have:
- [X] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/slavecommunicator.cc
+++ b/pdns/slavecommunicator.cc
@@ -92,6 +92,7 @@ void CommunicatorClass::ixfrSuck(const DNSName &domain, const TSIGTriplet& tt, c
     DNSSECKeeper dk (&B); // reuse our UeberBackend copy for DNSSECKeeper
 
     bool wrongDomainKind = false;
+    // this checks three error conditions, and sets wrongDomainKind if we hit the third & had an error
     if(!B.getDomainInfo(domain, di) || !di.backend || (wrongDomainKind = true, di.kind != DomainInfo::Slave)) { // di.backend and B are mostly identical
       if(wrongDomainKind)
         g_log<<Logger::Error<<"Can't determine backend for domain '"<<domain<<"', not configured as slave"<<endl;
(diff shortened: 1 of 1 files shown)


pdns/slavecommunicator.cc
```diff
@@ -91,8 +91,12 @@ void CommunicatorClass::ixfrSuck(const DNSName &domain, const TSIGTriplet& tt, c
   try {
     DNSSECKeeper dk (&B); // reuse our UeberBackend copy for DNSSECKeeper
 
-    if(!B.getDomainInfo(domain, di) || !di.backend || di.kind != DomainInfo::Slave) { // di.backend and B are mostly identical
-      g_log<<Logger::Error<<"Can't determine backend for domain '"<<domain<<"'"<<endl;
+    bool wrongDomainKind = false;
+    if(!B.getDomainInfo(domain, di) || !di.backend || (wrongDomainKind = true, di.kind != DomainInfo::Slave)) { // di.backend and B are mostly identical
```
Habbie: This looks super clever and I cannot read it. Can we change it? Maybe some `()` ?

ahupowerdns: if this is problematic, a few more ()'s are not going to help I'm afraid - unless I misunderstand where your confusion is coming from. I can also rewrite this into 7 lines of not much more obvious code. The way to read the above is that it sets a flag if it hit the third || component. Let me know!

Habbie: The () would help if `(wrongDomainKind = true, di.kind != DomainInfo::Slave)` meant `(wrongDomainKind = (true, di.kind != DomainInfo::Slave))` but it does not, so that's okay.

So, if we hit the third component, `wrongDomainKind` is set to true, even if the third component then evaluates to false. But this is fine because we then only look at `wrongDomainKind` if the third component was true. I get it now, thanks!

ahupowerdns: I'll add a comment.

Issue #7900, opened 2019-06-10
DoH: allow configuration of HTTP header


appliedprivacy:
- Program: dnsdist
 - Issue type: Feature request

### Usecase

RFC8484 states:
> also allowing web applications
> to access DNS information via existing browser APIs in a safe way
> consistent with Cross Origin Resource Sharing (CORS)

This usecase requires CORS headers otherwise browsers will not allow it.

dnsdist does not yet set the CORS header:
```
HTTP/2 200 
server: h2o/2.2.5
date: Mon, 10 Jun 2019 11:12:24 GMT
content-type: application/dns-message
content-length: 80
```

This feature request is about allowing operators to add HTTP headers to dnsdist's HTTP responses to allow the described usecase and potentially others. One obvious usecase is CORS
- `Access-Control-Allow-Origin: *`

but this feature should allow the operator to configure arbitrary HTTP header (i.e. also HSTS).
Some exceptions may apply, for example the server header is already set via `serverTokens` #7894.

HTTP header should be configurable per frontend and per SNI.

ref: https://gitlab.labs.nic.cz/knot/knot-resolver/merge_requests/823

Habbie:
With reference to https://dnsdist.org/reference/config.html#addDOHLocal, perhaps adding 'headers' to the options, taking a table, might make sense. I'm assuming that we can easily deal with tables in that place.

rgacogne:
Agreed, we already do something like that for the internal webserver: https://dnsdist.org/reference/config.html?highlight=headers#webserver

melissavoegeli:
For transparency, I'm going to work on this for this week.  I'll keep you posted.

PR #7957, opened 2019-06-19
7900/melissavoegeli/dnsdist custom http response headers

melissavoegeli:
### Short description
Implements a solution for #7900 to allow custom response headers to be set with the addDOHLocal configuration.  H2o provides a [token table](https://github.com/h2o/h2o/blob/master/lib/common/token_table.h) of acceptable headers.  This PR utilizes that list as the allowable response headers.  I still need to figure out how to do more appropriate warn logging and write a few more tests.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


pdns/dnsdist-lua.cc
```diff
@@ -1691,6 +1691,12 @@ void setupLuaConfig(bool client)
       if (vars->count("serverTokens")) {
         frontend->d_serverTokens = boost::get<const string>((*vars)["serverTokens"]);
       }
+      if (vars->count("customResponseHeaders")) {
+        for (auto const& headerMap : boost::get<std::map<std::string,std::string>>((*vars)["customResponseHeaders"])) {
+          std::pair<std::string,std::string> headerResponse = std::make_pair(headerMap.first, headerMap.second);
```
melissavoegeli: Is this a silly conversion?  It feels silly, but I see pair getting used a lot.

rgacogne: It's not, we want to be able to iterate quickly over the headers so a vector of pairs makes sense to me.

pdns/dnsdistdist/doh.cc
```diff
@@ -333,6 +333,16 @@ try
   h2o_socket_getsockname(sock, reinterpret_cast<struct sockaddr*>(&local));
   DOHServerConfig* dsc = reinterpret_cast<DOHServerConfig*>(req->conn->ctx->storage.entries[0].data);
 
+  // I wanted to put this near the other h2o_set_header method call, but should I move it?
+  for (auto const& headerPair : dsc->df->d_customResponseHeaders) {
+    const h2o_token_t *token = h2o_lookup_token(headerPair.first.c_str(), headerPair.first.size());
+    // TODO: Should I provide so WARN logging when we can't find a matching token?
+    if (token != NULL) {
```
rgacogne: Just a nit but we try to use `nullptr`, which is not silently converted to a numerical value, for example.
```suggestion
    if (token != nullptr) {
```

pdns/dnsdistdist/doh.cc
```diff
@@ -333,6 +333,16 @@ try
...
+      // h2o_set_header(pool, headers, token, value, value_len, overwrite_if_exists)
+      h2o_set_header(&req->pool, &req->res.headers, token, headerPair.second.c_str(), headerPair.second.size(), 1);
```
rgacogne: How about calling `h2o_set_header_by_str()` instead? It will do the lookup for you (provided that you set `maybe_token` to 1) so it should be as fast, but will also add an unknown (one for which there is no token) header.

melissavoegeli: Do we want to allow unknown headers?  I thought this was a surety concern as well as a performance consideration.

rgacogne: I think we should allow it, if it really has a negative impact the admin can just stick to known ones :)

pdns/dnsdist-lua.cc
rgacogne: I missed your question on IRC yesterday but we can iterate more quickly over a vector of pairs than a set or a map because the memory is contiguous. I'm not worried about duplicated entries, since you use a map in the configuration step, and even otherwise it would be up to the admin to be careful.

rgacogne:
Regression tests are currently failing with this error on Travis:
```
pycurl.error: (23, 'Failed writing header')
```
and the log shows:
```
TypeError: string argument expected, got 'bytes'
```

rgacogne:
> Regression tests are currently failing with this error on Travis:

On CircleCI, actually, sorry!

rgacogne:
Hi! I'd love to merge this PR once the tests are fine, let us know if you need some help, or if you are just too busy at the moment :-)

rgacogne:
Superseded by #8148. Thanks a lot!

PR #7971, opened 2019-06-22
dnsdist: after partial write to TCP client, only send remaining data

Habbie:
### Short description
If dnsdist could not send all buffered data to a TCP client, it would later try to send the full size of the send buffer, from the correct non-zero position, leading it to also send whatever was -behind- the buffer.

I have a somewhat flakey setup for testing this. I haven't figured out how we could write a reliable test for this.

### Checklist

I have:
- [x] read the [CONTRIBUTING.md](https://github.com/PowerDNS/pdns/blob/master/CONTRIBUTING.md) document
- [x] compiled this code
- [x] tested this code
- [ ] included documentation (including possible behaviour changes)
- [ ] documented the code
- [ ] added or modified regression test(s)
- [ ] added or modified unit test(s)
- [ ]  checked that this code was merged to master

PR #7974, opened 2019-06-24
dnsdist: Fix short IOs over TCP

rgacogne:
### Short description

`TCPIOHandler::tryRead()` and `TCPIOHandler::tryWrite()` are supposed to update the position and read/write the remaining number of bytes based on the initial total number minus the position, so that the caller can just pass the exact same parameters on short reads/writes until it succeeds or an exception is raised, without having to care about updating the position and number of bytes to read/write.
<strike>Untested yet.</strike>

Supersedes #7971.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/tcpiohandler.hh
+++ b/pdns/tcpiohandler.hh
@@ -245,7 +245,7 @@ public:
     }
 
     do {
-      ssize_t res = ::write(d_socket, reinterpret_cast<char*>(&buffer.at(pos)), toWrite - pos);
+      ssize_t res = ::write(d_socket, reinterpret_cast<const char*>(&buffer.at(pos)), toWrite - pos);
       if (res == 0) {
         throw runtime_error("EOF while sending message");
       }
--- a/pdns/tcpiohandler.hh
+++ b/pdns/tcpiohandler.hh
@@ -192,7 +192,7 @@ public:
     }
   }
 
-  /* Tries to read exactly toRead bytes into the buffer, starting at position pos.
+  /* Tries to read exactly toRead - pos bytes into the buffer, starting at position pos.
      Updates pos everytime a successful read occurs,
      throws an std::runtime_error in case of IO error,
      return Done when toRead bytes have been read, needRead or needWrite if the IO operation
@@ -208,9 +208,8 @@ public:
       return d_conn->tryRead(buffer, pos, toRead);
     }
 
-    size_t got = 0;
     do {
-      ssize_t res = ::read(d_socket, reinterpret_cast<char*>(&buffer.at(pos)), toRead - got);
+      ssize_t res = ::read(d_socket, reinterpret_cast<char*>(&buffer.at(pos)), toRead - pos);
       if (res == 0) {
         throw runtime_error("EOF while reading message");
       }
@@ -224,14 +223,13 @@ public:
       }
 
       pos += static_cast<size_t>(res);
-      got += static_cast<size_t>(res);
     }
-    while (got < toRead);
+    while (pos < toRead);
 
     return IOState::Done;
   }
 
-  /* Tries to write exactly toWrite bytes from the buffer, starting at position pos.
+  /* Tries to write exactly toWrite - pos bytes from the buffer, starting at position pos.
      Updates pos everytime a successful write occurs,
      throws an std::runtime_error in case of IO error,
      return Done when toWrite bytes have been written, needRead or needWrite if the IO operation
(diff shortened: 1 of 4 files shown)


pdns/tcpiohandler.hh
```diff
@@ -208,9 +208,8 @@ public:
       return d_conn->tryRead(buffer, pos, toRead);
     }
```
omoerbeek: The cases `toRead == pos` (will cause `::read` to rerturn `0`) and `pos > toRead` (sign wrap) should be handled.

pdns/tcpiohandler.hh
```diff
@@ -243,9 +241,8 @@ public:
       return d_conn->tryWrite(buffer, pos, toWrite);
     }
```
omoerbeek: same here

rgacogne:
Added some checks suggested by @omoerbeek and fixed the existing one, which compared the size of the buffer to the requested size, not the requested size minus the existing position.

Habbie:
>  Untested yet.

It fixes the problem in my test environment!

rgacogne:
I forgot to update my comment but I actually tested it in the meantime! I reproduced the issues with short reads and short writes with master, and confirmed that this PR fixed them.
I'll try to write a test for these cases, since it's already the second time I manage to introduce a bug in that exact spot.

omoerbeek:
I approved this, but now it looks to me like not all cases of tryRead were covered.

pdns/tcpiohandler.hh
```diff
@@ -224,28 +223,29 @@ public:
...
+      throw std::out_of_range("Calling tryWrite() with a too small buffer (" + std::to_string(buffer.size()) + ") for a write of " + std::to_string(toWrite - pos) + " bytes starting at " + std::to_string(pos));
+    }
     if (d_conn) {
       return d_conn->tryWrite(buffer, pos, toWrite);
     }
 
-    size_t sent = 0;
     do {
-      ssize_t res = ::write(d_socket, reinterpret_cast<char*>(&buffer.at(pos)), toWrite - sent);
+      ssize_t res = ::write(d_socket, reinterpret_cast<char*>(&buffer.at(pos)), toWrite - pos);
```
cmouse: are you accidentically removing constness here?

cmouse: also, what if res < 0?

pdns/tcpiohandler.hh
```diff
@@ -192,25 +192,24 @@ public:
...
     }
 
     if (d_conn) {
       return d_conn->tryRead(buffer, pos, toRead);
     }
 
-    size_t got = 0;
     do {
-      ssize_t res = ::read(d_socket, reinterpret_cast<char*>(&buffer.at(pos)), toRead - got);
+      ssize_t res = ::read(d_socket, reinterpret_cast<char*>(&buffer.at(pos)), toRead - pos);
```
cmouse: what if res < 0?

pdns/tcpiohandler.hh
omoerbeek: res < 0 case handled below

cmouse: uhm, how?

omoerbeek: line 252

cmouse: *sigh*, i got fooled by the review view.

rgacogne:
> Seems a few calls to `tryRead` were missed when converting the 3rd arg. There are 4 calls here,
> line 1045 got converted, 1067 looks right, but 936 and 946 need fixing.

You are mistaking `TCPIOHandler::tryRead()` and a static function named `tryRead()`, which exists because we don't speak DoT toward backends atm and so have no need for the filters abstraction. I have a WIP branch unifying those two.

omoerbeek:
> > Seems a few calls to `tryRead` were missed when converting the 3rd arg. There are 4 calls here,
> > line 1045 got converted, 1067 looks right, but 936 and 946 need fixing.
> 
> You are mistaking `TCPIOHandler::tryRead()` and a static function named `tryRead()`, which exists because we don't speak DoT toward backends atm and so have no need for the filters abstraction. I have a WIP branch unifying those two.

Ah, right. Did not spot that and jumped to conlcusions. Having two conventions for the third arg *is* confusing...

rgacogne:
Now with regression tests.

Issue #1956, opened 2014-12-17
Build fails looking for `execinfo.h` when building against musl


James-TR:
`execinfo.h` is a GNU specific header, and doesn't exist under musl.

A suggestion would be to make it so this feature is auto-detected to exist, and even add a configure flag.

Relevant log entry:

```
x86_64-gentoo-linux-musl-g++ -DHAVE_CONFIG_H -I. -I..   -pthread     -I.. -I.. -DSYSCONFDIR=\"/etc/powerdns\" -DPKGLIBDIR=\"/usr/lib/powerdns/pdns\" -DLOCALSTATEDIR=\"/var/run\" -pthread    -I../pdns/ext/rapidjson/include -I../pdns/ext/yahttp -D_GNU_SOURCE -march=native -O2 -fstack-protector-all -fomit-frame-pointer -fstack-protector --param ssp-buffer-size=4 -fPIE -DPIE -c -o receiver.o receiver.cc
receiver.cc:377:22: fatal error: execinfo.h: No such file or directory
 #include <execinfo.h>
                      ^
```

PR #7980, opened 2019-06-25
auth: make sure we really are using glibc

James-TR:
### Short description
Make sure we're using glibc and not uclibc pretending to be glibc

### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/receiver.cc
+++ b/pdns/receiver.cc
@@ -354,7 +354,7 @@ static int guardian(int argc, char **argv)
   }
 }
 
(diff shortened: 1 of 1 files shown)


James-TR:
This is related to work I did in #1958 to fix #1956, which obviously wasn't fully completed.

Habbie:
I have verified that this does not break tracebacks on my Debian Buster.

PR #7983, opened 2019-06-25
Add script for generating repo files.

aerique:
This is mainly used for quickly testing new releases by starting up a Docker container for each of our repositories, this saves a lot of handwork.

Although this script is very blunt and already starting to get out of hand I still would like this in our repo to get more eyes on and as a reminder that we need something like this.

In time, we should replace it with better tools or port it to a real programming languages, but this has not happened in the last months as it will not happen in the foreseeable future.  For that time, this will come in handy.


Habbie:
Merging this as I couldn't find it last time I needed it!

PR #7984, opened 2019-06-25
Add SLAVE-RENOTIFY zone metadata support

qvr:
### Short description
Adds support for new per-zone metadata SLAVE-RENOTIFY, which does the same thing as the global slave-renotify setting, but per-zone.

Needed for example when pdns is acting as a slave for a hidden master, but still needs to notify upstream slaves for specific zones.

### Checklist
I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/docs/domainmetadata.rst
+++ b/docs/domainmetadata.rst
@@ -172,6 +172,16 @@ This metadata can also be set using the
 and ``set-publish-cds``. For an example for an :rfc:`7344` key rollover,
 see the :doc:`guides/kskrollcdnskey`.
 
+.. _metadata-slave-renotify:
+
+SLAVE-RENOTIFY
+--------------
+.. versionadded:: 4.3.0
+
+If set to 1, will make PowerDNS renotify the slaves after an AXFR is received from a master.
+Any other value means that no renotifies are done. If not set at all, action will depend on
+the :ref:`setting-slave-renotify` setting.
+
 .. _metadata-soa-edit:
 
 SOA-EDIT
--- a/pdns/slavecommunicator.cc
+++ b/pdns/slavecommunicator.cc
@@ -599,7 +599,19 @@ void CommunicatorClass::suck(const DNSName &domain, const ComboAddress& remote)
 
 
     g_log<<Logger::Error<<"AXFR done for '"<<domain<<"', zone committed with serial number "<<zs.soa_serial<<endl;
+
+    bool renotify = false;
     if(::arg().mustDo("slave-renotify"))
+      renotify = true;
+    vector<string> meta;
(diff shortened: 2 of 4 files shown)


pdns/slavecommunicator.cc
```diff
@@ -599,7 +599,8 @@ void CommunicatorClass::suck(const DNSName &domain, const ComboAddress& remote)
 
 
     g_log<<Logger::Error<<"AXFR done for '"<<domain<<"', zone committed with serial number "<<zs.soa_serial<<endl;
-    if(::arg().mustDo("slave-renotify"))
+    vector<string> meta;
+    if(::arg().mustDo("slave-renotify") || (B.getDomainMetadata(domain, "SLAVE-RENOTIFY", meta) && meta.size() > 0 && meta[0]=="1"))
```
mind04: Please use the same logic as API-RECTIFY 

> When the API-RECTIFY value is “1”, the zone will be rectified on changes. Any other other value means that it will not be rectified. If this is not set at all, rectifying of the zone depends on the config variable default-api-rectify.`

SLAVE_RENOTIFY == 1 -> always renotify
SLAVE_RENOTIFY != 1 -> never renotify (this part is missing right now)

qvr: Changed to similar logic

docs/domainmetadata.rst
```diff
@@ -172,6 +172,14 @@ This metadata can also be set using the
 and ``set-publish-cds``. For an example for an :rfc:`7344` key rollover,
 see the :doc:`guides/kskrollcdnskey`.
 
+.. _metadata-slave-renotify:
+
+SLAVE-RENOTIFY
+--------------
```
pieterlexis: Please add:

```
.. versionadded:: 4.3.0
```

pdns/slavecommunicator.cc
```diff
@@ -599,7 +599,18 @@ void CommunicatorClass::suck(const DNSName &domain, const ComboAddress& remote)
...
     g_log<<Logger::Error<<"AXFR done for '"<<domain<<"', zone committed with serial number "<<zs.soa_serial<<endl;
+
+    bool renotify = false;
     if(::arg().mustDo("slave-renotify"))
+      renotify = true;
+    vector<string> meta;
+    if (B.getDomainMetadata(domain, "SLAVE-RENOTIFY", meta) && meta.size() > 0) {
+      if (meta[0] == "1") {
+        renotify = true;
+      } else
```
pieterlexis: Please use braces here as well

docs/domainmetadata.rst
qvr: done, and fixed doc to mention the new logic

pdns/slavecommunicator.cc
qvr: done

Issue #7390, opened 2019-01-18
dnsdist: make SuffixMatchNode::add() accept a string


rgacogne:
- Program: dnsdist 
 - Issue type: Feature request 

### Short description
Right now `SuffixMatchNode::add()` accepts only a `DNSName`, it would be nice to accept a `string` as well.

PR #7985, opened 2019-06-25
SuffixMatchNode:add: Accept more types

pieterlexis:
### Short description
The SuffixMatchNode now also accepts a table of DNSNames, a table of
string, and singular strings.

Closes https://github.com/PowerDNS/pdns/issues/7390.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdist-lua-bindings.cc
+++ b/pdns/dnsdist-lua-bindings.cc
@@ -205,7 +205,32 @@ void setupLuaBindings(bool client)
   g_lua.registerFunction("empty",(bool (DNSNameSet::*)()) &DNSNameSet::empty);
 
   /* SuffixMatchNode */
-  g_lua.registerFunction("add",(void (SuffixMatchNode::*)(const DNSName&)) &SuffixMatchNode::add);
+  g_lua.registerFunction<void (SuffixMatchNode::*)(const boost::variant<DNSName, string, vector<pair<int, DNSName>>, vector<pair<int, string>>> &name)>("add", [](SuffixMatchNode &smn, const boost::variant<DNSName, string, vector<pair<int, DNSName>>, vector<pair<int, string>>> &name) {
+      if (name.type() == typeid(DNSName)) {
+          auto n = boost::get<DNSName>(name);
+          smn.add(n);
+          return;
+      }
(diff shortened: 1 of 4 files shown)


pdns/dnsdist-lua-bindings.cc
```diff
@@ -205,7 +205,32 @@ void setupLuaBindings(bool client)
...
-  g_lua.registerFunction("add",(void (SuffixMatchNode::*)(const DNSName&)) &SuffixMatchNode::add);
+  g_lua.registerFunction<void (SuffixMatchNode::*)(const boost::variant<DNSName, string, vector<pair<int, DNSName>>, vector<pair<int, string>>> &name)>("add", [](SuffixMatchNode &smn, const boost::variant<DNSName, string, vector<pair<int, DNSName>>, vector<pair<int, string>>> &name) {
+      if (name.type() == typeid(DNSName)) {
+          auto n = boost::get<DNSName>(name);
+          smn.add(n);
+          return;
+      }
+      if (name.type() == typeid(string)) {
+          auto n = boost::get<string>(name);
+          smn.add(n);
```
rgacogne: We might want to catch the exception that might be raised when the string is implicitly converted to a `DNSName` to provide a nice error message.

pdns/dnsdist-lua-bindings.cc
```diff
@@ -205,7 +205,32 @@ void setupLuaBindings(bool client)
...
+          auto names = boost::get<vector<pair<int, DNSName>>>(name);
+          for (auto const n : names) {
+            smn.add(n.second);
+          }
+          return;
+      }
+      if (name.type() == typeid(vector<pair<int, string>>)) {
+          auto names = boost::get<vector<pair<int, string>>>(name);
+          for (auto const n : names) {
+            smn.add(n.second);
```
rgacogne: Same here.

pdns/dnsdist-lua-bindings.cc
rgacogne: It's actually explicit since you added `SuffixMatchNode::add(const std::string& name)` :)

Issue #7981, opened 2019-06-25
Investigate valgrind report wrt uninitalized mem in sendmsg 


omoerbeek:
- Program: dnsdist, but potentially more 
 - Issue type: Bug report 

### Short description

As reported by valgrind
```==30898== Thread 3 dnsdist/respond:
==30898== Syscall param sendmsg(msg.msg_control) points to uninitialised byte(s)
==30898==    at 0x5DECEB0: sendmsg (sendmsg.c:28)
==30898==    by 0xC2C010: sendfromto(int, char const*, unsigned long, int, ComboAddress const&, ComboAddress const&) (iputils.cc:241)
==30898==    by 0x74CBC0: sendUDPResponse(int, char const*, unsigned short, int, ComboAddress const&, ComboAddress const&) (dnsdist.cc:479)
```
But do consider:
```
// if you end up here because valgrind told you were are doing something wrong
// with msgh->msg_controllen, please refer to https://github.com/PowerDNS/pdns/pull/3962
// first.
void addCMsgSrcAddr(struct msghdr* msgh, void* cmsgbuf, const ComboAddress* source, int itfIndex)
```

But note that this valgrind report is about `msg_control` and not `msg_controllen`

omoerbeek:
Apart form the alignment thing below the code looks ok.
So the reason why valgrind complains could be `msg_controllen` is set to `CMSG_SPACE(size)`, but only `size` is filled in `addCMsgSrcAddr`. So alignment bytes are left uninitialized. It's debatble if valgrind should complain about that.

That said, I do noticed that the buffer is not aligned explicitly. At least OpenBSD recommends that. A PR will be upcoming.

PR #7990, opened 2019-06-26
Explicitly align the buffer used for cmsgs

omoerbeek:
### Short description

OpenBSDs man page http://man.openbsd.org/CMSG_DATA recommend explicit alignment of the buffer to hold cmsg data. 
This probably does not matter on non-strict alignment architectures, but is good practise.
(Disclaimer: I was one of the devs involved in this change to the man page, see http://cvsweb.openbsd.org/src/share/man/man3/CMSG_DATA.3).

Noticed while studying #7981

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [X] documented the code

### DIFF
--- a/pdns/calidns.cc
+++ b/pdns/calidns.cc
@@ -71,11 +71,13 @@ static void* recvThread(const vector<Socket*>* sockets)
 #if HAVE_RECVMMSG
   vector<struct mmsghdr> buf(100);
   for(auto& m : buf) {
-    fillMSGHdr(&m.msg_hdr, new struct iovec, new char[512], 512, new char[1500], 1500, new ComboAddress("127.0.0.1"));
+    cmsgbuf_aligned *cbuf = new cmsgbuf_aligned;
+    fillMSGHdr(&m.msg_hdr, new struct iovec, cbuf, sizeof(*cbuf), new char[1500], 1500, new ComboAddress("127.0.0.1"));
   }
 #else
   struct msghdr buf;
-  fillMSGHdr(&buf, new struct iovec, new char[512], 512, new char[1500], 1500, new ComboAddress("127.0.0.1"));
+  cmsgbuf_aligned *cbuf = new cmsgbuf_aligned;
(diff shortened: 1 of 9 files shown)


pdns/pdns_recursor.cc
```diff
@@ -1654,12 +1654,12 @@ static void startDoResolve(void *p)
...
       struct iovec iov;
-      char cbuf[256];
-      fillMSGHdr(&msgh, &iov, cbuf, 0, (char*)&*packet.begin(), packet.size(), &dc->d_remote);
+      cmsgbuf_aligned cbuf;
+      fillMSGHdr(&msgh, &iov, &cbuf, 0, (char*)&*packet.begin(), packet.size(), &dc->d_remote);
       msgh.msg_control=NULL;
 
       if(g_fromtosockets.count(dc->d_socket)) {
-	addCMsgSrcAddr(&msgh, cbuf, &dc->d_local, 0);
+	      addCMsgSrcAddr(&msgh, &cbuf, &dc->d_local, 0);
```
rgacogne: Just a nit but the indentation seems off.

pdns/calidns.cc
```diff
@@ -71,11 +71,13 @@ static void* recvThread(const vector<Socket*>* sockets)
 #if HAVE_RECVMMSG
   vector<struct mmsghdr> buf(100);
   for(auto& m : buf) {
-    fillMSGHdr(&m.msg_hdr, new struct iovec, new char[512], 512, new char[1500], 1500, new ComboAddress("127.0.0.1"));
```
rgacogne: I wonder if we used to allocate 512 bytes instead of the usual 256 ones for a reason?

pdns/pdns_recursor.cc
omoerbeek: will fix

pdns/calidns.cc
omoerbeek: No idea. 256 is already generous.

PR #7991, opened 2019-06-26
auth: add new metadata ALLOW-AXFR-ONLY-FROM

qvr:
### Short description
Same as ALLOW-AXFR-FROM, but allows only the specified subnets.
Global allow-axfr-ips= setting and any possible ALLOW-AXFR-FROM entries are ignored. Empty value "" will deny transfers from all addresses.

GSS/TSIG authenticated AXFRs will still work regardless of this setting.

### Checklist
I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/docs/domainmetadata.rst
+++ b/docs/domainmetadata.rst
@@ -46,6 +46,18 @@ Each ACL has its own row in the database:
 To disallow all IP's, except those explicitly allowed by domainmetadata
 records, add ``allow-axfr-ips=`` to ``pdns.conf``.
 
+.. _metadata-allow-axfr-only-from:
+
+ALLOW-AXFR-ONLY-FROM
+--------------------
+.. versionadded:: 4.3.0
+
+Same as ``ALLOW-AXFR-FROM``, but allows only the specified subnets.
+Global ``allow-axfr-ips=`` setting and any possible ``ALLOW-AXFR-FROM`` entries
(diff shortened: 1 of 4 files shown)


mind04:
What is your use case for this new metadata option? Is it something you cannot accomplish with the existing (metadata)options?

qvr:
Yes, my usecase is lots of zones that have many addresses to allow axfr from, and then a small number of zones for which the list has to be different. Without this new exception option, the only other option with pdns would be to have an empty allow-axfr-ips and keep hundreds of thousands allow-axfr-from entries in sync.

mind04:
I do not see a problem in the per domain solution. I assume you have automation in place to manage these zones, so adding the metadata is just one additional query/api call.
Allowing things by default and disable them per zone has security risks.
Sorry, I'm no fan of this new metadata option since this will add an other layer to the already complicated "allow axfr" configuration.

qvr:
Keeping the metadata in sync is not just one additional call when you have to loop all the zones to get their current metadata entries, and then as many update/delete's as it takes to end up synced.

As for security risks, it would only have a negative effect if you *wanted* the functionality but then failed to add it to a zone, which would be an operator error. Checking if it's set is easy and (atleast for my usecase but I'd guess in general as well) far less calls than checking all zones.

Habbie:
Closed after private conversation with @qvr.

PR #7992, opened 2019-06-26
dnsdist: Add `quiet` parameter to NetmaskGroupRule

pieterlexis:
### Short description
This allows the user to hide the list of matched IPs in `showRules()`

### Checklist

I have:
- [x] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdistdist/docs/rules-actions.rst
+++ b/pdns/dnsdistdist/docs/rules-actions.rst
@@ -625,7 +625,7 @@ These ``DNSRule``\ s be one of the following items:
 
   :param NetMaskGroup nmg: The NetMaskGroup to match on
   :param bool src: Whether to match source or destination address of the packet. Defaults to true (matches source)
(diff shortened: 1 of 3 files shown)


pdns/dnsdistdist/docs/rules-actions.rst
```diff
@@ -622,6 +625,7 @@ These ``DNSRule``\ s be one of the following items:
 
   :param NetMaskGroup nmg: The NetMaskGroup to match on
   :param bool src: Whether to match source or destination address of the packet. Defaults to true (matches source)
+  :param bool quiet: Do not return the list of matched netmasks. Default is false.
```
rgacogne: Perhaps we could replace 'return' with 'display', although I see we already use 'return' for the SMN version:
```suggestion
  :param bool quiet: Do not display the list of matched netmasks. Default is false.
```

pieterlexis: I'll change both!

PR #7993, opened 2019-06-26
dnsdist: improve Lua error reporting during config load

Habbie:
### Short description
This replaces the useless `[string "chunk"]` with the file name. It preserves the showing of the actual error:
```
$ ./dnsdist -C ./dnsdist.conf
Fatal Lua error: ./dnsdist.conf:3: unfinished string near ''dnsdist2.lua)()'
$ ./dnsdist -C ./dnsdist2.conf
Fatal Lua error: cannot open ./dnsdist2.conf: No such file or directory
$ ./dnsdist -C /dev/mem
Fatal Lua error: cannot open /dev/mem: Permission denied
```

TODO: also fix this in `includeDirectory`.

### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdist-lua.cc
+++ b/pdns/dnsdist-lua.cc
@@ -1967,13 +1967,7 @@ vector<std::function<void(void)>> setupLua(bool client, const std::string& confi
   setupLuaRules();
   setupLuaVars();
 
-  std::ifstream ifs(config);
-  if(!ifs)
-    warnlog("Unable to read configuration from '%s'", config);
-  else
-    vinfolog("Read configuration from '%s'", config);
-
-  g_lua.executeCode(ifs);
+  g_lua.executeCode("dofile('"+config+"')");
 
   auto ret = *g_launchWork;
   delete g_launchWork;


pdns/dnsdist-lua.cc
```diff
@@ -1967,13 +1967,7 @@ vector<std::function<void(void)>> setupLua(bool client, const std::string& confi
   setupLuaRules();
   setupLuaVars();
 
-  std::ifstream ifs(config);
-  if(!ifs)
-    warnlog("Unable to read configuration from '%s'", config);
```
rgacogne: Not being able to open the configuration file used to be only a warning (which may or may not have been a good idea) but will now prevent dnsdist from starting, correct?

pdns/dnsdist-lua.cc
```diff
@@ -1967,13 +1967,7 @@ vector<std::function<void(void)>> setupLua(bool client, const std::string& confi
...
-  else
-    vinfolog("Read configuration from '%s'", config);
```
rgacogne: Perhaps we could keep this line, after the `g_lua.executeCode()` one since we don't need to have the name of the file printed before anymore.

pdns/dnsdist-lua.cc
Habbie: Yes, it's fatal now - it probably should be

pdns/dnsdist-lua.cc
Habbie: yes, makes sense

pdns/dnsdist-lua.cc
pieterlexis: I use `-l 127.0.0.1:5300 127.0.0.1:5301` to quickly spawn `dnsdist` to test things, this would make life a bit harder (having to create a config file)

Habbie: right

pieterlexis: unless `-C /dev/null` also works :)

Habbie: > unless `-C /dev/null` also works :)

Haha, I came here to post this! Yes, that works.

Habbie: @rgacogne @pieterlexis with that addition, is the change acceptable?

rgacogne: I really would like dnsdist to keep starting even if it couldn't read its configuration file, at least for 1.4.0. Perhaps we could still try to open it and only try to parse it that succeeded?
I'm OK with changing that behavior for 1.5.0.

pieterlexis:

PR #7997, opened 2019-06-28
rec: Fix the export of only outgoing queries or incoming responses

rgacogne:
### Short description

The documentation states that you can log only outgoing queries or incoming responses, but this was not actually supported.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


omoerbeek:
According to doc "New in version 4.2.0" so no need to backport to 4.1.x I suppose.

PR #8001, opened 2019-06-28
recursor: add deviceName field to protobuf messages

chbruyand:
### Short description
This PR adds a new `deviceName` field to the protobuf messages. Like other fields such as `deviceId` or `requestorId`, this can be set dynamically in the `gettag()` LUA function. This also fixes the FFI corresponding API to actually set this field (and not the `deviceId`).

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/recursordist/docs/lua-scripting/dq.rst
+++ b/pdns/recursordist/docs/lua-scripting/dq.rst
@@ -108,6 +108,12 @@ The DNSQuestion object contains at least the following fields:
 
       A string that will be used to set the ``deviceId`` field in :doc:`protobuf <../lua-config/protobuf>` messages.
 
+  .. attribute:: DNSQuestion.deviceName
+
+      .. versionadded:: 4.1.15
+
+      A string that will be used to set the ``deviceName`` field in :doc:`protobuf <../lua-config/protobuf>` messages.
+
   .. attribute:: DNSQuestion.udpAnswer
 
       Answer to the :attr:`udpQuery <DNSQuestion.udpQuery>` when when using the ``udpQueryResponse`` :attr:`followupFunction <DNSQuestion.followupFunction>`.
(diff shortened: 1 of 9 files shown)


pdns/recursordist/docs/lua-scripting/dq.rst
```diff
@@ -108,6 +108,12 @@ The DNSQuestion object contains at least the following fields:
 
       A string that will be used to set the ``deviceId`` field in :doc:`protobuf <../lua-config/protobuf>` messages.
 
+  .. attribute:: DNSQuestion.deviceName
```
rgacogne: It looks like we don't actually export this field (or `deviceId` for that matter, look at `requestorId` if you want a working example) of the `DNSQuestion` object in the non-FFI Lua interface.

chbruyand: Thanks for your review. Should be fixed now.

pdns/recursordist/docs/lua-scripting/dq.rst
```diff
@@ -108,6 +108,12 @@ The DNSQuestion object contains at least the following fields:
...
+
+      .. versionadded:: 4.1.15
```
omoerbeek: Version is not right I suppose.

pdns/recursordist/docs/lua-scripting/hooks.rst
```diff
@@ -69,6 +69,9 @@ Interception Functions
     .. versionadded:: 4.1.0
 
         It can also return a table whose keys and values are strings to fill the :attr:`DNSQuestion.data` table, as well as a ``requestorId`` value to fill the :attr:`DNSQuestion.requestorId` field and a ``deviceId`` value to fill the :attr:`DNSQuestion.deviceId` field.
+    .. versionadded:: 4.1.15
```
omoerbeek: Version is not right I suppose.

pdns/recursordist/docs/lua-scripting/dq.rst
chbruyand: Fixed, thanks!

PR #8008, opened 2019-07-01
rec: Don't go Bogus if the auth zone delegation test takes too long

rgacogne:
### Short description

On some very slow platforms where the test was taking too long, the signature's inception would have ended up _before_ the current SyncRes time used to do the validation, causing a validation issue.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/recursordist/test-syncres_cc.cc
+++ b/pdns/recursordist/test-syncres_cc.cc
@@ -414,7 +414,7 @@ void generateKeyMaterial(const DNSName& name, unsigned int algo, uint8_t digest,
   dsAnchors[name].insert(keys[name].second);
 }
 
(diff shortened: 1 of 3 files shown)


zeha:
I've applied this in Debian as part of 4.2.0-5, as on mips the `test_auth_zone_delegation` failed like this:

```
test-syncres_cc4.cc(93): Entering test case "test_auth_zone_delegation"
test-syncres_cc4.cc(162): error: in "syncres_cc4/test_auth_zone_delegation": check queriesCount == 4 has failed [6 != 4]
test-syncres_cc4.cc(93): Leaving test case "test_auth_zone_delegation"; testing time: 18952ms
```

PR #8013, opened 2019-07-02
Don't create temporary strings to escape DNSName labels

rgacogne:
### Short description

Shaves a nice 20%+ off `DNSName::toString()` for regular names, and cut in half the CPU time when we encounter a 'special' character (not `.` or `\` and lesser or equal to `0x20` or greater or equal to `0x7f`).

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsname.cc
+++ b/pdns/dnsname.cc
@@ -177,7 +177,8 @@ std::string DNSName::toString(const std::string& separator, const bool trailing)
     const char* end = p + d_storage.size();
 
     while (p < end && *p) {
-      ret += escapeLabel(p + 1, static_cast<size_t>(*p)) + separator;
+      appendEscapedLabel(ret, p + 1, static_cast<size_t>(*p));
+      ret += separator;
       p += *p + 1;
     }
   }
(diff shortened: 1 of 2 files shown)


pdns/dnsname.cc
```diff
@@ -430,29 +431,27 @@ size_t hash_value(DNSName const& d)
...
 {
-  return escapeLabel(label.c_str(), label.size());
-}
-
-string DNSName::escapeLabel(const char* orig, size_t len)
-{
-  std::string ret;
   size_t pos = 0;
 
-  ret.reserve(len);
```
chbruyand: Maybe this would be worth to keep `reserve()`ing enough space to append the escaped label ?

omoerbeek: I think that is covered by line 172. In the regular case that should be enough.

rgacogne: In theory we should have reserved enough space already at the beginning of `DNSName::toString()`, but that might not be true due to escaping, I'll try to measure whether a `ret.reserve(ret.size() + len)` helps.

chbruyand: > I think that is covered by line 172. In the regular case that should be enough.

Yes indeed, my bad.

PR #8024, opened 2019-07-04
 dnsdist: Display TCP/DoT queries and responses in verbose mode, opcode in grepq

rgacogne:
### Short description

- We display the queries and responses received over UDP and DoH in verbose mode, but we were not displaying the ones received over TCP or DoT ;
- Display any non-zero opcode in `grepq()`.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dns.cc
+++ b/pdns/dns.cc
@@ -70,7 +70,7 @@ std::string ERCode::to_s(uint8_t rcode) {
 }
 
 std::string Opcode::to_s(uint8_t opcode) {
-  static std::vector<std::string> s_opcodes = { "Query", "IQuery", "Status", "3", "Notify", "Update" };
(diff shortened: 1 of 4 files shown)


pdns/dns.cc
```diff
@@ -69,6 +69,16 @@ std::string ERCode::to_s(uint8_t rcode) {
   return RCode::rcodes_s[rcode];
 }
 
+std::string Opcode::to_s(uint8_t opcode) {
+  static std::vector<std::string> s_opcodes = { "Query", "IQuery", "Status", "3", "Notify", "Update" };
```
omoerbeek: nit: cold be const

omoerbeek: Could be `static const`

rgacogne: Done!

PR #8028, opened 2019-07-05
limit compression pointers to 14 bits

Habbie:
### Short description
In a very specific edge case, we would get compression pointers above the 0-0x3fff range, and truncate those to a very low number, causing clients to throw parse errors.

I have added a fix from @pieterlexis where DNSPacketWriter did not honour `compress=false` for the content of any type (like NS) that is permitted to compress by RFC.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnswriter.cc
+++ b/pdns/dnswriter.cc
@@ -201,6 +201,7 @@ void DNSPacketWriter::xfrUnquotedText(const string& text, bool lenField)
 
 
 static constexpr bool l_verbose=false;
+static constexpr uint16_t maxCompressionOffset=16384;
 uint16_t DNSPacketWriter::lookupName(const DNSName& name, uint16_t* matchLen)
 {
   // iterate over the written labels, see if we find a match
@@ -268,7 +269,7 @@ uint16_t DNSPacketWriter::lookupName(const DNSName& name, uint16_t* matchLen)
         if(!c)
           break;
         auto offset = iter - d_content.cbegin();
-        if (offset >= 16384) break; // compression pointers cannot point here
+        if (offset >= maxCompressionOffset) break; // compression pointers cannot point here
         pvect.push_back(offset);
         iter+=*iter+1;
       }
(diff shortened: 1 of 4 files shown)


pdns/dnswriter.cc
```diff
@@ -325,7 +325,7 @@ void DNSPacketWriter::xfrName(const DNSName& name, bool compress, bool)
 
   uint16_t li=0;
   uint16_t matchlen=0;
-  if(compress && (li=lookupName(name, &matchlen))) {
+  if(compress && (li=lookupName(name, &matchlen)) && li < 16384) {
```
rgacogne: I wonder if that check should live in `DNSPacketWriter::lookupName()` instead so we can try to find an existing name with less matching labels?

Habbie: ah yes

Habbie: pushed

Habbie:
This PR contains two commits. i have verified that either commit, independently, fixes the bug. The second commit is an optimisation suggested by @rgacogne.

Habbie:
After discussion with @omoerbeek and @rgacogne, I have pushed a third commit that also, independently, fixes the issue. I think we want to ship the first and third commits.

pdns/dnswriter.cc
```diff
@@ -266,6 +266,8 @@ uint16_t DNSPacketWriter::lookupName(const DNSName& name, uint16_t* matchLen)
         }
         if(!c)
           break;
+        auto offset = iter - d_content.cbegin();
+        if (offset >= 16384) break; // compression pointers cannot point here
         pvect.push_back(iter - d_content.cbegin());
```
rgacogne:
```suggestion
        pvect.push_back(offset);
```

Habbie: duh

Habbie: pushed

omoerbeek:
Would it be posisble to transform you test code in a unit test?

From close source reading the suggested fix looks right (commits 1+3+4). Although 1 is not strictly necessary. Existing unittests also succeed for all three products here.

Habbie:
> Would it be possible to transform you test code into a unit test?

Probably, and I think it's a good idea. It might be the kind of test that 'breaks' whenever we change something in this area, though.

> Although 1 is not strictly necessary.

Correct, but it has no downsides and thus is a nice failsafe.

Habbie:
I pushed one extremely specific test.

pdns/dnswriter.cc
```diff
@@ -266,7 +267,9 @@ uint16_t DNSPacketWriter::lookupName(const DNSName& name, uint16_t* matchLen)
         }
         if(!c)
           break;
-        pvect.push_back(iter - d_content.cbegin());
+        auto offset = iter - d_content.cbegin();
+        if (offset >= 16384) break; // compression pointers cannot point here
```
pieterlexis: It might make sense to
```cpp
#define MAX_COMPRESSION_POINTER 16384
```

and use it everywhere

Habbie:
Squashed, ready for final review I hope!

PR #8029, opened 2019-07-05
bind-load-threads config improves (re)load

mzealey:
### Short description

Changed the bindbackend config loader to be multi-threaded when loading individual zones. Added a new bind-load-threads to configure how many this should be (default 1). I see linear improvement in load speed from 22sec at 140k zones to 6 sec with 4 threads on my laptop, I believe this probably scales linearly to roughly 16 cores before the master thread controlling access to the main data structure gets overworked.

I have not fully tested this code in all exception cases yet although I don't think there should be many surprises. I am requesting a code review to ensure you are happy with the idea behind this and the code done so far before I continue.

### Checklist
I have:
- [x] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/docs/backends/bind.rst
+++ b/docs/backends/bind.rst
@@ -64,6 +64,15 @@ Interval in seconds to check for zone file changes. Default is 0 (disabled).
 
 See :ref:`bind-operation` section for more information.
 
+.. _setting-bind-load-threads:
+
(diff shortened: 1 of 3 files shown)


pieterlexis:
@mind04 could you share your thoughts on this code?

PR #7879, opened 2019-06-06
Recursor: don't start as root in systemd

pieterlexis:
### Short description
This PR also moves the socket-dir by default to /var/run/pdns-recursor. Due to `Runtimedir` in systemf, this directory is created with the correct owenership and permissions. The init-script is amended to create this directory as well and as the recursor is started as root, all is fine :).

### Checklist

I have:
- [x] read the [CONTRIBUTING.md](https://github.com/PowerDNS/pdns/blob/master/CONTRIBUTING.md) document
- [x] compiled this code
- [x] tested this code
- [ ] included documentation (including possible behaviour changes)
- [ ] documented the code
- [ ] added or modified regression test(s)
- [ ] added or modified unit test(s)
- [ ]  checked that this code was merged to master

PR #8034, opened 2019-07-08
Ensure Debian SysV users get set{g,u}id

pieterlexis:
### Short description
As we'd now run as root in those instances. We also add a warning in the config-file about this. See https://github.com/PowerDNS/pdns/pull/7879#issuecomment-509033651 for the discussion

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


zeha:
Thanks. I meant to do this myself but didn't get around to it.

Issue #7875, opened 2019-06-05
bindbackend parser issue on unquoted TXT fields


mzealey:
- Program: Authoritative
 - Issue type: Bug report

### Short description

We have a bind record like the below which works with named:

```
default._domainkey IN TXT "v=DKIM1; k=rsa; p=MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEAsZvaAEWSj+ntYKqVPw99HQE5k0etIrot8/QF8kTqAe4J8O+qWuq9qNapiSEeV3OAcQlsmcnELrQo3zItT+oGN43bl3V2H/yFQkBHF0mPfUgPWB4fzRTvwNBACaOGTI390X7X9jX1wWXxjP/h/pa5UdJphoiiF+lRrjAhB2ck2jAJSMNW8DGnlkbf1EPTCS3av" 5mkBzlQDNLQWhfy/2amX5iNYAJh9wdOteY+KTIFKXTKF0RqWrzEXedHqBtN1vHX3WdMkWD0HL3ZLmWLowUdnYu3bvJ4ob3XVse6Wz54CJe7BHkRh4c/7zmUOstQftWBBJnfYjnLpCHt4glVkaJAdwIDAQAB\;
```

When queried on powerdns it fails and logs the following error:

```
Exception building answer packet for default._domainkey.XXX/TXT (Parsing record content (try 'pdnsutil check-zone'): Data field in DNS should start with quote (") at position 258
```

Placing quotes around the second string fixes the issue.

Unfortunately we don't easily control these files, and my understanding is that bindbackend should be a drop-in replacement for named so filing a bug report to ask you to fix this issue and not require quotes on TXT blocks. Thanks!

### Environment
 - Operating system: linux
 - Software version: 4.2.0-rc1
 - Software source: PowerDNS repository RPM

PR #8035, opened 2019-07-08
Improve BIND zone file parsing for TXT

mzealey:
### Short description

BIND allows multiple TXT sections to be specified and they do not have
to be quoted. This commit adds functionality to handle this correctly
rather than SERVFAIL as at present.

This is a fix for #7875

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code


pdns/zoneparser-tng.cc
```diff
@@ -272,6 +272,38 @@ pair<string,int> ZoneParserTNG::getLineNumAndFile()
...
+        ++i;
+        in_quote = false;
+      }
+      else if( !needs_end_quote && txt[i] == '"' ) {
+        in_quote = false;
+      }
+      else if( txt[i] == '\\' ) {     // escape - skip the next character
+        ++i;
+      }
+    } else if( txt[i] != ' ' ) {      // start of a new string
```
gibson042: Tab characters are also separators rather than content, cf. [RFC 1035](https://tools.ietf.org/html/rfc1035#section-5.1).

pdns/zoneparser-tng.cc
```diff
@@ -272,6 +272,38 @@ pair<string,int> ZoneParserTNG::getLineNumAndFile()
...
+      in_quote = true;
+      if( txt[i] != '"' ) {
+        txt.insert(i, "\"");
+        ++i;
```
gibson042: It looks like this will skip over the first character of an unquoted \<character-string>, which could be problematic if it is a backslash (e.g., `TXT \  2` should be recognized as identical to `TXT " " "2"`).

pdns/zoneparser-tng.cc
```diff
@@ -272,6 +272,38 @@ pair<string,int> ZoneParserTNG::getLineNumAndFile()
     return {d_filestates.top().d_filename, d_filestates.top().d_lineno};
 }
 
+void fixTxtRecord(string &txt) {
+  bool in_quote = false;
+  bool needs_end_quote = false;
+
+  // Bind automatically puts quotes around text blocks that don't have them
+  for( string::size_type i = 0; i < txt.size(); ++i ) {
+    if( in_quote ) {
```
gibson042: I think this will break on unseparated \<character-string>s like `TXT foo"bar"`, though to be honest I'm not sure what other software does with those either. But does seem to accept e.g. `TXT "foo"bar` and `TXT "foo""bar"`, so there's an asymmetry if nothing else.

pdns/zoneparser-tng.cc
```diff
@@ -272,6 +272,38 @@ pair<string,int> ZoneParserTNG::getLineNumAndFile()
...
+      if( txt[i] != '"' ) {
+        txt.insert(i, "\"");
+        ++i;
+        needs_end_quote = true;
+      } else
+        needs_end_quote = false;
+    }
+  }
+  if( in_quote && needs_end_quote )
+    txt.append("\"");
```
gibson042: Wouldn't it be better to reject bad input than to try correcting it and likely screw things up?

mzealey: I agree it would be best to reject bad input, however not quoting your TXT is not bad input as it works fine in named.

gibson042: named is not the definition of master file syntax; [RFC 1035](https://tools.ietf.org/html/rfc1035#section-5.1) is. If the former allows unterminated \<character-string>s like e.g. `TXT "foo`, then it's going off-script.

mzealey: It's not unterminated as much as implicitly quoting unquoted sections eg `TXT foo`

mzealey: Which, looking at the RFC is how it should be so pdns is currently in regression to this...

```
<character-string> is expressed in one or two ways: as a contiguous set
of characters without interior spaces, or as a string beginning with a "
and ending with a ".  Inside a " delimited string any character can
occur, except for a " itself, which must be quoted using \ (back slash)
```

mzealey:
Have produced a simpler patch which corrects the handling in rcpgenerator.cc now. Assuming the code is correct, I cannot see an issue with regressions on production as it simply allows a slightly larger subset of unquoted TXT records compared to the present situation. It makes handling closer to RFC1035.

pdns/rcpgenerator.cc
```diff
@@ -377,30 +377,36 @@ void RecordTextReader::xfrText(string& val, bool multi, bool lenField)
...
-      throw RecordTextException("Data field in DNS should start with quote (\") at position "+std::to_string(d_pos)+" of '"+d_string+"'");
-    }
-    val.append(1, '"');
-    while(++d_pos < d_end && d_string[d_pos]!='"') {
-      if(d_string[d_pos]=='\\' && d_pos+1!=d_end) {
+
+      if(is_plenus && dns_isspace(d_string[d_pos]))
+        break;
+
+      if(d_string[d_pos]=='\\' && d_pos+1!=d_end)
```
gibson042: Are you allowing `\` to escape line terminators (e.g., line feed `\n` and carriage return `\r`), and if so is that intentional? This is also ambiguous in the RFC, and BIND allows it only inside _quoted_ \<character-string>s.

pdns/rcpgenerator.cc
```diff
@@ -377,30 +377,36 @@ void RecordTextReader::xfrText(string& val, bool multi, bool lenField)
...
         val.append(1, d_string[d_pos++]);
```
gibson042: This code fails to reject the following invalid cases (but I don't know if they get caught elsewhere):
* too-short \\DDD escapes (e.g., `foo\03bar`)
* out-of-range \\DDD escapes (e.g., `foo\256bar`)
* unterminated escape sequences (e.g., `foo\`)

EvanCarroll: Is this what's holding up the patch @gibson042

gibson042: They should certainly be fixed in my opinion, but I'm not a maintainer of this repository.

miodvallat:
Superseded by #16087, closing.

PR #8044, opened 2019-07-08
add key make time to pdnsutil test-all-algorithms, cleanup return type

Habbie:
### Short description
`Testing algorithm 8: 'OpenSSL RSA' ->'OpenSSL RSA' -> 'OpenSSL RSA' Signature & verify ok, create 173usec, signature 75usec, verify 14usec` - the `create` part is new.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnssecinfra.cc
+++ b/pdns/dnssecinfra.cc
@@ -196,7 +196,7 @@ bool DNSCryptoKeyEngine::testAll()
         
         for(maker_t* verifier :  value.second) {
           try {
-            /* pair<unsigned int, unsigned int> res=*/ testMakers(value.first, creator, signer, verifier);
+            testMakers(value.first, creator, signer, verifier);
           }
           catch(std::exception& e)
           {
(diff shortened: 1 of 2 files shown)


pdns/dnssecinfra.cc
```diff
@@ -253,7 +253,9 @@ pair<unsigned int, unsigned int> DNSCryptoKeyEngine::testMakers(unsigned int alg
   else
     throw runtime_error("Can't guess key size for algorithm "+std::to_string(algo));
 
+  DTime dt; dt.set();
   dckeCreate->create(bits);
+  unsigned int udiffCreate = dt.udiff() / 100;
```
rgacogne: We only create once so unless I'm mistaken the `/ 100` part should go.

Habbie: blind copy pasting - maybe we should create 100 times too?

rgacogne: That might make sense, yes. Verify as well, I guess?

Habbie: amended

PR #8047, opened 2019-07-09
Another time sensistive test fixed with a fixedNow construct.

omoerbeek:
Fix borrowed from #8008.
But this test also does `addRRSIG()` calls, these need adaptation as well.

### Short description

See https://builder.powerdns.com/#/builders/114/builds/301

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code
- [] added or modified regression test(s)

### DIFF
--- a/pdns/recursordist/test-syncres_cc5.cc
+++ b/pdns/recursordist/test-syncres_cc5.cc
@@ -26,7 +26,13 @@ BOOST_AUTO_TEST_CASE(test_dnssec_secure_various_algos) {
 
   size_t queriesCount = 0;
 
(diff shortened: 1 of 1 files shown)


omoerbeek:
Indeed, but making the test less time sensitive is a good thing anyway imo. I would love to be able to run the tests in a slow (debug/trace/sanitize) environment without spurious failures.

rgacogne:
> Indeed, but making the test less time sensitive is a good thing anyway imo. I would love to be able to run the tests in a slow (debug/trace/sanitize) environment without spurious failures.

I totally agree, it's just that we need to find a better way than editing the tests one by one every time a new failure appears :)

PR #8031, opened 2019-07-05
Remove useless chopOff loop in LMDB

mzealey:
### Short description

UeberBackend already calls lookup() multiple times with the ending
chopped off, so doing it again here doesn't make any sense.

### Checklist

I have:
- [x] read the [CONTRIBUTING.md](https://github.com/PowerDNS/pdns/blob/master/CONTRIBUTING.md) document
- [x] compiled this code
- [x] tested this code
- [ ] included documentation (including possible behaviour changes)
- [ ] documented the code
- [ ] added or modified regression test(s)
- [ ] added or modified unit test(s)

Issue #4907, opened 2017-01-13
bind-backend regression 3.x -> 4.x with multiple backends slips in data from wrong zone


ZaphodB:
- Program: Authoritative 
 - Issue type: Bug report 

### Short description

3 backends in order:
 - bind for service checks loading zone "localhost"
 - gmysql DB pdns_slave for slave zones
 - gmysql DB pdns for NATIVE zones

After upgrading from 3.4.1-4+deb8u6 to 4.0.2-1pdns.jessie PowerDNS Server returned zone "localhost" data along with the SOA for the first slave zone in pdns_slave DB although that slave zone does not have anything like localhost in it.

with
`launch=bind:servicecheck,gmysql:slave,gmysql:master`

```
➜  ~ dig any collins.kg @wv-adns2-ha2.wavecloud.de
;; Truncated, retrying in TCP mode.

; <<>> DiG 9.8.3-P1 <<>> any collins.kg @wv-adns2-ha2.wavecloud.de
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 34362
;; flags: qr aa rd; QUERY: 1, ANSWER: 3, AUTHORITY: 0, ADDITIONAL: 0
;; WARNING: recursion requested but not available

;; QUESTION SECTION:
;collins.kg.			IN	ANY

;; ANSWER SECTION:
localhost.		86400	IN	NS	localhost.
localhost.		86400	IN	A	127.0.0.1
collins.kg.		3600	IN	SOA	ns1.wavedns.de. it-guys.project-collins.com. 2017011201 600 3600 604800 600

;; Query time: 36 msec
;; SERVER: 185.22.222.15#53(185.22.222.15)
;; WHEN: Fri Jan 13 19:47:57 2017
;; MSG SIZE  rcvd: 144
```

with
`launch=gmysql:slave,gmysql:master,bind:servicecheck`

```
➜  ~ dig any collins.kg @wv-adns2-ha2.wavecloud.de
;; Truncated, retrying in TCP mode.

; <<>> DiG 9.8.3-P1 <<>> any collins.kg @wv-adns2-ha2.wavecloud.de
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 41358
;; flags: qr aa rd; QUERY: 1, ANSWER: 10, AUTHORITY: 0, ADDITIONAL: 2
[...]

PR #8050, opened 2019-07-09
Improved version of #8031 and some other fixes and optimizations.

mind04:
### Short description
The first commit is an improved version of #8031, since this optimization is only possible for SOA queries. The second commit is a cleanup to make it more easy to spot incorrect use of lookup() without a zone_id. The other commits are 'collateral damage' found while testing the first two commits.

Closes #8031

### Checklist
I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/modules/bindbackend/bindbackend2.cc
+++ b/modules/bindbackend/bindbackend2.cc
@@ -1053,7 +1053,7 @@ bool Bind2Backend::getBeforeAndAfterNamesAbsolute(uint32_t id, const DNSName& qn
   }
 }
 
-void Bind2Backend::lookup(const QType &qtype, const DNSName &qname, DNSPacket *pkt_p, int zoneId )
+void Bind2Backend::lookup(const QType &qtype, const DNSName &qname, int zoneId, DNSPacket *pkt_p )
 {
   d_handle.reset();
   DNSName domain(qname);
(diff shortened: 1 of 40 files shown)


rgacogne:
Unless I'm missing something, the latest commit (34a9e979be42aa4c14f6441311c7acb39d55edc1) looks a lot like https://github.com/PowerDNS/pdns/pull/4638 which we had to revert in https://github.com/PowerDNS/pdns/pull/4905. Perhaps the additional check suggested in https://github.com/PowerDNS/pdns/pull/4905#issuecomment-272714469 would be good enough?

mind04:
I think this issue #4907 was responsible for the revert. The problem here is a duplicate id in both backends. This is not allowed in pdns and pdnsutil check-all-zones is throwing errors if this does occur.

```
Checked 35 records of 'example.com', 0 errors, 0 warnings.
[Error] Another SOA for zone 'example.com' (serial 0) has already been seen (serial 2019052902).
[Error] Domain ID 1 of 'example.com' in backend gmysql has already been used by zone '0.0.127.in-addr.arpa' in backend bind.
```

Duplicates id's in will cause a lot of different issues (wrong keys, wrong metadata, ect.), I'm not sure we should slow down backends to be less broken when there are clearly problems in a database/configuration. 

With the current bind backend powerdns is happily serving records from the wrong zone when there is a root zone or when there is a child zone in backend one and the parent zone in backend two (sharing the same id).

We can however do a little bit better here without slowing us down as already suggested in #4638. With the 'fix' the level of brokenness with duplicate id's in bind backend will be equal to the other backends.

PR #8059, opened 2019-07-11
UeberBackend optimization

mzealey:
When all backends have returned a result then stop the loop. More modern
backends may be able to return a shorter- or empty DNSName so even with
the optimization to prevent re-querying on that specific backend, it
doesn't make sense to keep looping and checking caches until the shorter
query name reaches the specific qname that was returned.

This is just an idea at the moment I have NOT fully tested, want to see what the automated checks pick up.

### Checklist

I have:
- [x] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/ueberbackend.cc
+++ b/pdns/ueberbackend.cc
@@ -287,7 +287,8 @@ bool UeberBackend::getAuth(const DNSName &target, const QType& qtype, SOAData* s
   bool found = false;
   int cstat;
   DNSName shorter(target);
-  vector<pair<size_t, SOAData> > bestmatch (backends.size(), make_pair(target.wirelength()+1, SOAData()));
+  const size_t initial_length = target.wirelength()+1;
+  vector<pair<size_t, SOAData> > bestmatch (backends.size(), make_pair(initial_length, SOAData()));
   do {
 
     // Check cache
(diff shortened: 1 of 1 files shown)


Habbie:
> This is just an idea at the moment I have NOT fully tested, want to see what the automated checks pick up.

Note that you can enable Travis and Circle on your own fork too - then you don't need to PR!

pdns/ueberbackend.cc
```diff
@@ -344,15 +345,31 @@ bool UeberBackend::getAuth(const DNSName &target, const QType& qtype, SOAData* s
...
+            if(d_negcache_ttl) {
+              DLOG(g_log<<Logger::Error<<"add neg cache entry:"<<shorter<<endl);
+              d_question.qname=shorter;
+              addNegCache(d_question);
+            }
+            continue;
+          }
+      }
+
+      // Definately got a result from one of the backends, cache it
```
jsoref, line 371:
```suggestion
      // Definitely got a result from one of the backends, cache it
```
per https://github.com/jsoref/pdns/commit/3d71a43f243eeb6b8f4361e81ae0bfe851c935e0#commitcomment-39230327

pdns/ueberbackend.cc
```diff
@@ -344,15 +345,31 @@ bool UeberBackend::getAuth(const DNSName &target, const QType& qtype, SOAData* s
...
-      // Add to cache
-      if(i == backends.end()) {
-        if(d_negcache_ttl) {
-          DLOG(g_log<<Logger::Error<<"add neg cache entry:"<<shorter<<endl);
-          d_question.qname=shorter;
-          addNegCache(d_question);
-        }
-        continue;
-      } else if(d_cache_ttl) {
+      // If all backends have return a result (which may be empty or shorter
```
Godwottery, line 348: "have return a result" should be "have returned a result"

PR #8067, opened 2019-07-11
Make sure we always compile with BOOST_CB_ENABLE_DEBUG set to 0

rgacogne:
### Short description

Even though the documentation states that:

> simultaneous read accesses to a shared circular_buffer are safe

`boost::circular_buffer` < 1.62 was built with a debugging facility *by default* that made it unsafe to access the same circular buffer from more than one thread at once, in addition to making iterator operations slower. 

- https://svn.boost.org/trac10/ticket/6277
- https://github.com/boostorg/circular_buffer/pull/9

It's only a real issue in dnsdist since our DNS over TLS code is the only place where we read from a circular buffer from several threads at once, but this change might provide a small performance boost in the auth and rec as well on platforms where boost is older than 1.62.

Note that I have compiled and tested this code only on Arch, we need to test it on other distributions before merging this PR, especially ones with a boost version lesser than 1.62. 

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/pdns/Makefile.am
+++ b/pdns/Makefile.am
@@ -160,6 +160,7 @@ pdns_server_SOURCES = \
 	bindlexer.l \
 	bindparser.cc \
 	cachecleaner.hh \
+	circular_buffer.hh \
 	comment.hh \
 	common_startup.cc common_startup.hh \
 	communicator.cc communicator.hh \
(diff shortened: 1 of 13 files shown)


omoerbeek:
Works fine on OpenBSD (boost 1.66, so that is no surprise). Checked that the only direct ref to boost/circular_buffer.hpp is in the new header file. I don not have a system with older boost at hand.

rgacogne:
Just tested it on CentOS 7, it doesn't work as expected because `BOOST_CB_ENABLE_DEBUG` is blindly overridden, we need to set `BOOST_CB_DISABLE_DEBUG` to 1 instead. Fix coming.

PR #8081, opened 2019-07-15
m4/systemd.m4: Fail when systemctl is not available

pieterlexis:
### Short description
We use systemctl to determine the version as to find out what features
to enable in the unit file.

Thanks to @zeha for mentioning this
### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -231,12 +231,14 @@ commands:
               libsodium-dev \
               libsqlite3-dev \
               libssl-dev \
+              libsystemd-dev \
               libtool \
               libyaml-cpp-dev \
               make \
               pkg-config \
               ragel \
               sqlite3 \
+              systemd \
               unixodbc-dev \
               virtualenv \
               wget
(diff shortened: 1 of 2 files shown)


zeha:
failure from circleci:

```
configure: error: conditional "HAVE_SYSTEMD_DYNAMIC_USER" was never defined.
```

Habbie:
CircleCI: `configure: error: systemctl not found`

pieterlexis:
> CircleCI: configure: error: systemctl not found

Hmm, are we building with systemd support there?

Habbie:
> Hmm, are we building with systemd support there?

We install libsystemd but not systemd; some builds have `--disable-systemd` but most don't.

rgacogne:
> We install libsystemd but not systemd; some builds have --disable-systemd but most don't.

Shall we enable systemd there, since I guess most of our users will run under systemd nowadays?

pieterlexis:
> Shall we enable systemd there, since I guess most of our users will run under systemd nowadays?

I just did the exact opposite, mostly because we don't test the systemd functionality at all.

rgacogne:
But that means that we won't build the code paths specific to systemd on our CI, right? While if we enabled it, we could at least see compilation errors even if we don't test anything else?

pieterlexis:
makes sense, fixing

PR #8082, opened 2019-07-15
Move sbin manpages to section 8

zeha:
### Short description
This should avoid the warning "command-in-sbin-has-manpage-in-incorrect-section" from Debian's lintian.

### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code


pieterlexis:
I'm unsure about this... perhaps the binaries should not live in `sbin` at all?

zeha:
The bin/sbin distinction appears to be somewhat arbitrary nowadays. I guess for daemon binaries some argument could be found for shipping them in neither.

Habbie:
> I guess for daemon binaries some argument could be found for shipping them in neither.

what the BSDs call libexec always made sense to me, yes

jsoref:
Hi, for the time being (and hopefully going forward), pdns master is running the [check-spelling](https://github.com/marketplace/actions/check-spelling) action. In order to avoid a ❌ when merged https://github.com/jsoref/pdns/commit/7d2b4c641d5cf27a7d79adcd196158a254bf519a#commitcomment-39230593 -- you'll need to merge/rebase on `master` to get the check (and the files) -- you can accept the advice (which should add the items to `expect.txt`)

zeha:
On Linux, "libexec" appears to be `(/usr)/lib/$YOURPROGRAM/`. I'm not sure we want to go there, yet?

pieterlexis:
I think the proper solution is to move the programs to `/bin`, as PowerDNS is not really a 'system administration command'

omoerbeek:
> I think the proper solution is to move the programs to `/bin`, as PowerDNS is not really a 'system administration command'. I'll make a PR for this and close this.

Section 8 historically also contains daemons and their related tools, so section 8 still feels right to me. To quote intro*0* form OpenBSD:
```
    Section 8 pages also describe network services and daemons.  The rest of
     this page discusses some of the main daemons available on the OpenBSD
     system, and how to enable/disable them.
```

pieterlexis:
`man man` on linux only mentions:

```
       8   System administration commands (usually only for root)
```

I also don't want to bikeshed over this.

Habbie:
intro(8) on Linux, though:

> Section 8 of the manual describes commands which either can be or are used only by the superuser, like system-administration commands, daemons, and hardware-related commands.

PR #8087, opened 2019-07-16
Don't fail the test right away if we don't have the right number of FDs ready

rgacogne:
### Short description

This way we can at least check that the callbacks are called later.
Also, since it looks like the read pipe is always reported as ready, let's try to write first to see if it blocks.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


rgacogne:
LGTM failure is unrelated (analysis took too long).

PR #8089, opened 2019-07-17
Fix the rfc1982LessThan template, it only works properly if the cast …

omoerbeek:
…is done to a same-sized type.

If we use it for uint8_t or uint64_t it breaks currenty.  Add unit tests while there.

### Checklist

I have:
- [X ] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/misc.hh
+++ b/pdns/misc.hh
@@ -112,6 +112,7 @@ stringtok (Container &container, string const &in,
 
 template<typename T> bool rfc1982LessThan(T a, T b)
 {
+  static_assert(std::is_unsigned<T>::value, "rfc1982LessThan only works for unsigned types");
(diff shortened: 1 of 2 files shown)


pdns/misc.hh
```diff
@@ -112,7 +112,8 @@ stringtok (Container &container, string const &in,
 
 template<typename T> bool rfc1982LessThan(T a, T b)
 {
-  return ((signed)(a - b)) < 0;
+  typedef typename std::make_signed<T>::type signed_t;
```
rgacogne: It's not clear to me what `std::make_signed` does if we ever pass a signed type by mistake, perhaps we could add a `static_assert()` using `std::is_unsigned()` ?

PR #8094, opened 2019-07-17
auth: add full option to 'pdns_control show-config'

mind04:
### Short description
Without the full option only modified options are included in the output.
In full mode the output has the same format as 'pdns_server --config'.

### Checklist
I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


pdns/arguments.hh
```diff
@@ -106,9 +106,11 @@ public:
   void setCmd(const string &, const string &); //!< Add a command flag
   string &setSwitch(const string &, const string &); //!< Add a command flag
   string helpstring(string prefix=""); //!< generates the --help
-  string configstring(bool current=false); //!< generates the --mkconfig
+  string configstring(bool current=false, bool full=true); //!< generates the --mkconfig
```
zeha: the pre-existing comment here is... outdated?

zeha:
orthogonal: for the debian init scripted i would have needed a way of asking the pdns_server binary (not a running instance) what the value of a config setting (socket-dir) would be (after applying defaults and any config files)...

Habbie:
> orthogonal: for the debian init scripted i would have needed a way of asking the pdns_server binary (not a running instance) what the value of a config setting (socket-dir) would be (after applying defaults and any config files)...

I suggest making a separate ticket for that, as we're about to merge this PR and then nobody will ever see your comment again.

pdns/arguments.hh
Habbie: looks like this PR fixes that now

Habbie:
> orthogonal: for the debian init scripted i would have needed a way of asking the pdns_server binary (not a running instance) what the value of a config setting (socket-dir) would be (after applying defaults and any config files)...

With this PR:
```
$ pdns/pdns_server --config=diff --config-dir=. | grep ^socket-dir
socket-dir=.
```

pdns/arguments.hh
```diff
@@ -106,9 +106,11 @@ public:
   void setCmd(const string &, const string &); //!< Add a command flag
   string &setSwitch(const string &, const string &); //!< Add a switch flag
   string helpstring(string prefix=""); //!< generates the --help
-  string configstring(bool current=false); //!< generates the --mkconfig
+  string configstring(bool current=false, bool full=true); //!< generates the --config
```
zeha: methods with multiple default boolean args are weird; should one of them maybe not have a default arg or this be split into two (perhaps three) methods?

Habbie: I also wondered about that, perhaps it would be best to not give them defaults at all?

pdns/receiver.cc
```diff
@@ -569,7 +587,7 @@ int main(int argc, char **argv)
     DynListener::registerFunc("REMOTES", &DLRemotesHandler, "get top remotes");
     DynListener::registerFunc("SET",&DLSettingsHandler, "set config variables", "<var> <value>");
     DynListener::registerFunc("RETRIEVE",&DLNotifyRetrieveHandler, "retrieve slave domain", "<domain>");
-    DynListener::registerFunc("CURRENT-CONFIG",&DLCurrentConfigHandler, "retrieve the current configuration");
+    DynListener::registerFunc("CURRENT-CONFIG",&DLCurrentConfigHandler, "retrieve the current configuration", "[diff|default]");
```
zeha: i wonder if `default` is a useful thing to ask a running instance?

Habbie:
Have the review comments been addressed? If so, can you mark them resolved? :)

pdns/arguments.hh
mind04: the defaults are gone

PR #8097, opened 2019-07-17
auth: use BIGINT for notified_serial in pg schema

### LINKED ISSUE #6815: domains table notified_serial needs bigint instead of integer
- Program: Authoritative
 - Issue type: Bug report

### Short description
It seems there was a change from PDNS 4.0 to PDNS 4.1 (or even earlier) how serials in the range from 2^31 - 2^32-1 were handled.

Old version: internally the notified_serial was treated as unsigned, but when written into the DB, it was stored as INTEGER (signed). Hence, serials > 2^31 where stored as negative values. When read from DB the notified_serial was transparently converted again into the correct unsigned value.

PDNS 4.1: now the database code wants to write the unsigned type as unsigned type into the DB, and this fails (testet with Postgresql backend): 
> pdns[32458]: AXFR done for 'xxx.wien', zone committed with serial number 2824191541
> pdns[32458]: Queued also-notification of domain 'xxx.wien' to 176.9.98.135:53
> pdns[32458]: Unable to AXFR zone 'xxx.wien' from remote 'xx.59.115.xx' (PDNSException): GSQLBackend unable to refresh domain_id 533900: Fatal error during query: update soap_domains set notified_serial=$1 where id=$2: ERROR:  value "2824191541" is out of range for type integer

Further, reading the old-style negative values will fail to:
> Unable to parse record during incoming AXFR of 'xxx.ch' (std::exception): stoul returned result out of unsigned int range; data was "-1470775755"

Fix new installations:
- In the DB schema change the notified_serial type from INTEGER to BIGINT

Fix existing installations:
> ALTER TABLE domains ALTER notified_serial TYPE BIGINT;
BEGIN;
> NOTE: at this point PowerDNS will receive Postgresql Errors (prepared statements) and should be restartet
> UPDATE domains SET notified_serial = NULL WHERE notified_serial < 0;
COMMIT;

### Environment

 - Operating system: Debian
 - Software version: PDNS 4.1 + Postgresql Backend

Habbie:
### Short description
Fixes #6815.

Also adds the `change_date` and autoserial removal to the auth 4.2 upgrade docs.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/modules/gpgsqlbackend/4.1.0_to_4.2.0_schema.pgsql.sql
+++ b/modules/gpgsqlbackend/4.1.0_to_4.2.0_schema.pgsql.sql
@@ -1,1 +1,2 @@
 ALTER TABLE records DROP COLUMN change_date;
+ALTER TABLE domains ALTER notified_serial TYPE bigint USING CASE WHEN notified_serial >= 0 THEN notified_serial::bigint END;
--- a/pdns/pdnsutil.cc
+++ b/pdns/pdnsutil.cc
@@ -1868,6 +1868,20 @@ void testSchema(DNSSECKeeper& dk, const DNSName& zone)
     return;
   }
   cout<<"[+] ordername sorting is correct for names starting with _"<<endl;
+  cout<<"Setting low notified serial"<<endl;
+  db->setNotified(di.id, 500);
+  db->getDomainInfo(zone, di);
+  if(di.notified_serial != 500) {
+    cout<<"[-] Set serial 500, got back "<<di.notified_serial<<endl;
+  }
+  cout<<"Setting serial that needs 32 bits"<<endl;
+  db->setNotified(di.id, 2147484148);
+  db->getDomainInfo(zone, di);
+  if(di.notified_serial != 2147484148) {
+    cout<<"[-] Set serial 2147484148, got back "<<di.notified_serial<<endl;
+  } else {
+    cout<<"[+] Big serials work correctly"<<endl;
+  }
   cout<<endl;
   cout<<"End of tests, please remove "<<zone<<" from domains+records"<<endl;
 }
--- a/pdns/pdnsutil.cc
+++ b/pdns/pdnsutil.cc
@@ -1776,6 +1776,7 @@ void testSchema(DNSSECKeeper& dk, const DNSName& zone)
   cout<<"Note: test-schema will try to create the zone, but it will not remove it."<<endl;
   cout<<"Please clean up after this."<<endl;
   cout<<endl;
+  cout<<"If this test reports an error and aborts, please check your database schema."<<endl;
   cout<<"Constructing UeberBackend"<<endl;
   UeberBackend B("default");
   cout<<"Picking first backend - if this is not what you want, edit launch line!"<<endl;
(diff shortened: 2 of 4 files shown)


pdns/pdnsutil.cc
```diff
@@ -1868,6 +1868,20 @@ void testSchema(DNSSECKeeper& dk, const DNSName& zone)
...
+  db->setNotified(di.id, 500);
+  db->getDomainInfo(zone, di);
+  if(di.notified_serial != 500) {
+    cout<<"[-] Set serial 500, got back "<<di.notified_serial<<endl;
+  }
+  cout<<"Setting serial that needs 32 bits"<<endl;
+  db->setNotified(di.id, 2147484148);
+  db->getDomainInfo(zone, di);
+  if(di.notified_serial != 2147484148) {
+    cout<<"[-] Set serial 2147484148, got back "<<di.notified_serial<<endl;
```
pieterlexis: "Check the database schema"?

pdns/pdnsutil.cc
```diff
@@ -1868,6 +1868,20 @@ void testSchema(DNSSECKeeper& dk, const DNSName& zone)
...
   }
   cout<<"[+] ordername sorting is correct for names starting with _"<<endl;
+  cout<<"Setting low notified serial"<<endl;
+  db->setNotified(di.id, 500);
+  db->getDomainInfo(zone, di);
+  if(di.notified_serial != 500) {
+    cout<<"[-] Set serial 500, got back "<<di.notified_serial<<endl;
+  }
+  cout<<"Setting serial that needs 32 bits"<<endl;
+  db->setNotified(di.id, 2147484148);
```
pieterlexis: Won't this crash/error on non-bigint?

pdns/pdnsutil.cc
Habbie: That seems obvious to me when test-schema is unhappy...

pdns/pdnsutil.cc
Habbie: it will crash with an exception, yes

pdns/pdnsutil.cc
pieterlexis: > That seems obvious to **me**

Not to everyone though

pdns/pdnsutil.cc
pieterlexis: perhaps `try` it and log the failure?

pdns/pdnsutil.cc
Habbie: I put a note at the top so that I don't have to edit each error message.

pdns/pdnsutil.cc
Habbie: pushed!

modules/gpgsqlbackend/4.1.0_to_4.2.0_schema.pgsql.sql
```diff
@@ -1 +1,5 @@
 ALTER TABLE records DROP COLUMN change_date;
+ALTER TABLE domains ALTER notified_serial TYPE BIGINT;
```
Habbie: This entire file should be replaced with
```
ALTER TABLE domains ALTER notified_serial TYPE bigint USING CASE WHEN notified_serial >= 0 THEN notified_serial::bigint END;
```

Issue #3688, opened 2016-04-10
dnsdist: allow dynamic EDNS option stuffing


Habbie:
We have `addAction(AllRule(), MacAddrAction(65001))` but if I want to stuff something other than a MAC address, currently I need to edit C++ code and recompile. I would like to be able to generate the option data from Lua, dynamically, when a query comes in, before passing it to the backend.

andatche:
Another vote. This would be very useful!

PR #8115, opened 2019-07-22
dnsdist: Fix ECS addition when the OPT record is not the last one

### LINKED ISSUE #8098: AXFR + TSIG + ECS with dnsdist and pdns_auth backend fails with broken error
- Program: Recursor (maybe dnsdist)
 - Issue type: Bug report

### Short description
My setup consists of a dnsdist and a single pdns-authoritative backend. I have configured tsig for AXFRs, and everything runs fine when I perform a tsig-based dig axfr.

When I enable ECS in dnsdist, the dig +axfr returns a nasty error.

### Environment
 - Operating system: Centos 7
 - Software version: pdns-auth 4.1.latest + dnsdist 1.3.latest

### Steps to reproduce
1. Install the components, dnsdist on port 53, pdns-auth on 5300, dnsdist configured with:
`newServer({ name="a1", address="127.0.0.1:5300"})`
2. Enable axfr, create a zone, create a tsig key, assign to zone
3. Ensure you can use dig axfr with the tsig key
4. Change dnsdist config to:
`newServer({ name="a1", address="127.0.0.1:5300", useClientSubnet=true})`
5. dig axfr still works without tsig
6. when I try axfr WITH the tsig key, I get this:
```
hello@i1-r1 ~]$ dig axfr example.tld @35.228.242.29 -y 'hmac-sha1:testkey:xkS1J57nDi18rku+t8VlSNPwo9wmhjgZW6kP3N/tUL0='
;; communications error to 35.228.242.29#53: end of file
```
### Expected behaviour
axfr digs should work, just as before setting `useClientSubnet=true`

### Actual behaviour
EOF communications error in dig

### Other information
Will update the ticket later with a pcap

### Usecase
Enable ECS without breaking tsig axfr. Ultimate goal: log the ECS client ip when doing the tsig-based axfr instead of just the dnsdist ip.

### Description
Work :)

rgacogne:
### Short description

When the `OPT` record in the query is not the last one, for example if it is followed by a `TSIG`, dnsdist would not detect it correctly and would wrongly add a new `OPT` record if configured to add `ECS` information to the incoming query.

This issue went unnoticed for quite some time and the `TSIG` signature will be broken anyway so I don't think we should rush this fix for 1.4.0, therefore I'm scheduling it for 1.5.0 instead.

Fixes #8098.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


rgacogne:
You are assuming no compression, which I don't think will hold.

gibson042:
No, I explicitly mention fixing up compression pointers but have just not provided sample code for doing so ("_TODO: update compression pointers in trailing records that reference offsets >= oldSuffix_"). But even without that, adopting them would still be an improvement—the currently proposed changes of this PR corrupt the DNS message whenever _anything_ follows the insertion point; my suggestions unnecessarily corrupt the DNS message only when the inserted/altered OPT is not the last record **and** a compression pointer in a later record references an offset that falls on or after the insertion point. And as I also state in the TODO, "_it's rare for any record to follow OPT and rarer still to reference a name appearing so deep, but we shouldn't leave potentially exploitable holes_".

rgacogne:
I do not understand what you mean, the goal of this PR is to deal with cases where the OPT record is not the last one. Why are you stating that "the currently proposed changes of this PR corrupt the DNS message whenever anything follows the insertion point for new data"?

Most of the code in this PR is here because we don't want to send garbage when compression pointers are in use after the OPT record, so yes, I'm a bit worried about a proposal to remove that code while stating that it does not handle that case.

gibson042:
You misunderstand... I wasn't saying to skip the pointer rewriting, I was stubbing that work out in my short suggestions because the code is substantial. However, creating a copy of the packet with `slowRewriteQueryWithExistingEDNS` is still overkill when all that's needed is to shift and preserve the existing content as-is **except** for pointers that reference offsets at or after the insertion point (which necessarily follow the insertion point)—and such pointers should either get arithmetically adjusted by the amount that the content after the insertion point shifted OR expanded because that shift pushed the compression reference out of bounds. Consider:
```c
size_t nextNameOffset = sizeof(dnsheader) + consumed + 4;
unsigned int nameLen = 0;
uint16_t remaining, rtype, rclass, rdlen;
try {
  /* skip over the question/answer/authority sections, then try to find an OPT RR */
  for(remaining = ntohs(dh->qdcount) - 1; remaining-- > 0; ) {
    DNSName q(packet, packetSize, nextNameOffset, true, &rtype, &rclass, &nameLen);
    nextNameOffset += (size_t)nameLen + 4
  }
  for(remaining = ntohs(dh->ancount) + ntohs(dh->nscount); remaining-- > 0; ) {
    DNSName rr(packet, packetSize, nextNameOffset, true, &rtype, &rclass, &nameLen);
    nextNameOffset += (size_t)nameLen + 8;
    rdlen = (packet[nextNameOffset] * 256) + packet[nextNameOffset + 1];
    nextNameOffset += 2 + (size_t)rdlen;
  }
  for(remaining = ntohs(dh->arcount); remaining-- > 0; ) {
    DNSName rr(packet, packetSize, nextNameOffset, true, &rtype, &rclass, &nameLen);
    nextNameOffset += (size_t)nameLen + 8;
    rdlen = (packet[nextNameOffset] * 256) + packet[nextNameOffset + 1];
    nextNameOffset += 2 + (size_t)rdlen;
    if (rtype == QType::OPT && rclass == QClass::IN) {
      size_t optRdlenOffset = nextNameOffset - (size_t)rdlen - 2;
      /* found an OPT; replace or insert ECS and update disrupted compression pointers */
      return replaceOrInsertEDNSOption(packet, packetSize, len,
          optRdlenOffset, remaining, nextNameOffset,
          EDNSOptionCode::ECS, newECSOption, overrideExisting, ecsAdded);
    }
  }
}
catch(const std::range_error& e) {
  return false;
}

/* no OPT; add one as the last RR and don't worry about compression pointers */
return addEDNSWithOption(packet, packetSize, len, newECSOption, ednsAdded, ecsAdded);
```
```c
bool replaceOrInsertEDNSOption(char* const packet, const size_t packetSize, uint16_t* const len,
    size_t optRdlenOffset, uint16_t followingRRs, size_t nextRROffset,
    uint16_t optionCode, const string& newOption,
    bool overrideExisting, bool* const optionAdded)
{
  /* look for a preexisting option */
  unsigned char* optRdlenPos = reinterpret_cast<unsigned char*>(packet) + optRdlenOffset;
  uint16_t optRdlen = (packet[optRdlenPos] * 256) + packet[optRdlenPos + 1];
  char * optionStart = nullptr;
  size_t oldOptionSize = 0;
  int res = getEDNSOption(reinterpret_cast<char*>(optRdlenPos), len - optRdlenOffset,
      optionCode, &optionStart, &oldOptionSize);
  if (res != 0) {
    /* no preexisting option; start after other options and keep oldOptionSize 0 */
    optionStart = optRdlenPos + 2 + optRdlen;
  } else if (!overrideExisting) {
    /* not allowed to override */
    return true;
  }

  /* overwrite in place, if possible */
  int shift = newOption.size() - oldOptionSize;
  if (shift == 0) {
    memcpy(optionStart, newOption.c_str(), newOption.size());
    return true;
  }

  /* shift the old suffix and insert the new option */
  int remainingCapacity = packetSize - *len;
  remainingCapacity -= shift;
  if (remainingCapacity < 0) {
    return false;
  }
  unsigned char* oldSuffix = optionStart + oldOptionSize;
  assert(oldSuffix > packet);
  memmove(optionStart + newOption.size(), oldSuffix, *len - (oldSuffix - packet));
  memcpy(optionStart, newOption.c_str(), newOption.size());
  if (oldOptionSize != 0) {
    *ecsAdded = true;
    optRdlen++;
    packet[optRdlenPos] = optRdlen / 256;
    packet[optRdlenPos + 1] = optRdlen % 256;
  }

  /* update compression pointers as necessary */
  if (followingRRs > 0) {
    res = adjustCompressionPointersFrom(packet, packetSize, len,
        oldSuffix - packet, shift, followingRRs, nextRROffset + shift);
    if (res != 0) {
      return false;
    }
  }
}
```
```c
bool adjustCompressionPointersFrom(char* const packet, const size_t packetSize, uint16_t* const len,
    size_t movedOffset, int adjustment, uint16_t followingRRs, size_t nextRROffset)
{
  try {
    unsigned int nameLen = 0;
    uint16_t rtype, rclass, rdlen;
    const unsigned char* pos;
    unsigned char labelLen;
    while(followingRRs-- > 0) {
      /* read the owner name to verify that it's good, then rescan it for compression pointers */
      DNSName rr(packet, packetSize, nextRROffset, true, &rtype, &rclass, &nameLen);
      pos = packet + nextRROffset;
      nextRROffset += (size_t)nameLen + 8;
      rdlen = (packet[nextRROffset] * 256) + packet[nextRROffset + 1];
      nextRROffset += 2 + (size_t)rdlen;
      for(labelLen = *pos; labelLen != 0; labelLen = *pos) {
        if (labelLen < 0xc0) {
          /* normal label; advance to the next */
          pos += labelLen;
          continue;
        }
        labelLen &= (~0xc0);
        int target = (labelLen << 8) + *(const unsigned char*)(pos + 1);
        if (target < movedOffset) {
          /* unaffected compression pointer; stop scanning labels */
          break;
        }
        target += shift;
        if ((target & 0xc000) != 0) {
          /* newly out-of-bounds compression pointer */
          /* TODO if possible, expand, move following contents, and adjust nextRROffset accordingly */
          return false;
        }
        /* in-place adjustable compression pointer */
        target |= 0xc000;
        *pos = (unsigned char)(target >> 8);
        *(pos + 1) = (unsigned char)(target & 0xff);
        break;
      }
      /* TODO also adjust compression pointers inside well-known RDATA per RFC 3597 */
    }
  }
  catch(const std::range_error& e) {
    return false;
  }
}
```

P.S. There are problems in both `getEDNSOptionsStart` (it returns ENOENT unless the only record in the entire message is an additional-section OPT) and `DNSPacketWriter::xfrBlob` (it ignores compression pointers in RDATA, contrary to https://tools.ietf.org/html/rfc3597#section-4 ). So that would seem to diminish your hard-line stance of fixing compression pointers in post-OPT RRs rather than just shifting them to make room for new contents.  Regardless, should I open issues for those bugs?

gibson042:
> > There are problems in both getEDNSOptionsStart (it returns ENOENT unless the only record in the entire message is an additional-section OPT) and DNSPacketWriter::xfrBlob (it ignores compression pointers in RDATA, contrary to https://tools.ietf.org/html/rfc3597#section-4 ).
> 
> Unless I'm mistaken, `getEDNSOptionsStart()` returning `ENOENT` unless the only record in the entire message is an additional-section `OPT` should be fine in this PR since other cases will be handled by `slowRewriteQueryWithExistingEDNS()` or do you see another possibility that I missed?

A DNS query can have RRs in the answer and/or authority sections (e.g., for [DNS UPDATE](https://tools.ietf.org/html/rfc2136), or as attempted exploits, or even just from bugs). When processing such a message whose additional section is a single OPT (i.e., ARCOUNT 1), the `handleEDNSClientSubnet` of this PR will skip the slow path, get a nonzero ENOENT from `getEDNSOptionsStart`, and invoke `addEDNSWithECS`, which will add a _second_ OPT. `parseEDNSOptions` behavior is even worse—it proceeds as if there were no OPT. And `addEDNSToQueryTurnedResponse` leaves it untouched rather than removing/replacing it as designed.

As for the rest, I understand. We're actually not that far apart, and the intentional `DNSPacketWriter::xfrBlob` limitation does make sense (and further, you really could ignore compression pointers in post-OPT owner names for the same reason—if they were actually pointing at or after the OPT, then the input was likely already bad anyway). I appreciate how you've worked with me and hope you'll continue to do so.

If you take a fresh look at https://github.com/PowerDNS/pdns/pull/8115#issuecomment-534449535 , perhaps ignoring the gratuitous `adjustCompressionPointersFrom`, I think you'll find that the resulting code covers relevant post-OPT RR cases that are the prime focus of this PR plus trailing data cases like #8321 that it currently misses, _and_ gets rid of some ECS-specific clutter in favor of a much more straightforward and #3688-generalization-ready `replaceOrInsertEDNSOption`, _and_ does so by shifting the tail end of a packet rather than incurring the performance penalty of copying the whole thing. `handleEDNSClientSubnet` is currently confusing and reaches out to **four** distinct ECS-aware functions, but it could be as simple as
```c
int res = scanPacketForOpt(packet, packetSize, len,
    &optRDLenPosition, &unscannedRRCount, &nextRRPosition);
if (res != 0) {
  /* invalid input detected */
  return false;
}
if (optRDLenPosition != 0) {
  /* valid OPT found */
  return replaceOrInsertEDNSOption(packet, packetSize, len,
          optRDLenPosition, unscannedRRCount,
          EDNSOptionCode::ECS, newECSOption, overrideExisting, &ecsAdded);
}
return addEDNSWithOption(packet, packetSize, len,
    nextRRPosition, newECSOption, &ednsAdded, &ecsAdded);
```

Anyway, I think I've done enough hijacking here. Thanks again, and please do consider the above. I truly do appreciate the software, and come here with a desire to help it succeed.

PR #8117, opened 2019-07-23
dnsdist: Implement ContinueAction()

rgacogne:
### Short description

As discussed on the mailing-list [1], this new action makes it possible to use an otherwise terminal action without stopping the rule processing. For example:
```
addAction(AllRule(), PoolAction("mypool"))
addAction(AllRule(), DisableValidationAction())
```
would route all incoming queries to the `mypool` pool, and since the `PoolAction()` terminates the processing the `DisableValidationAction()` one would never be applied.
With the new action, it is possible to do:
```
addAction(AllRule(), ContinueAction(PoolAction("mypool")))
addAction(AllRule(), DisableValidationAction())
```
This would result in all queries being routed to the `mypool` pool *with* the CD `flag` being set.

[1]: https://mailman.powerdns.com/pipermail/dnsdist/2019-July/000590.html

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/pdns/dnsdist.cc
+++ b/pdns/dnsdist.cc
@@ -1036,7 +1036,71 @@ static void spoofResponseFromString(DNSQuestion& dq, const string& spoofContent)
   }
 }
 
-static bool applyRulesToQuery(LocalHolders& holders, DNSQuestion& dq, string& poolname, const struct timespec& now)
+bool processRulesResult(const DNSAction::Action& action, DNSQuestion& dq, std::string& ruleresult, bool& drop)
+{
+  switch(action) {
+  case DNSAction::Action::Allow:
+    return true;
+    break;
+  case DNSAction::Action::Drop:
+    ++g_stats.ruleDrop;
+    drop = true;
+    return true;
+    break;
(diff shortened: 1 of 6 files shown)


pdns/dnsdist.cc
```diff
@@ -1036,7 +1036,71 @@ static void spoofResponseFromString(DNSQuestion& dq, const string& spoofContent)
...
 
-static bool applyRulesToQuery(LocalHolders& holders, DNSQuestion& dq, string& poolname, const struct timespec& now)
+bool processRulesResult(const DNSAction::Action& action, DNSQuestion& dq, std::string& ruleresult, bool& drop)
+{
+  switch(action) {
+  case DNSAction::Action::Allow:
+    return true;
+    break;
+  case DNSAction::Action::Drop:
+    ++g_stats.ruleDrop;
```
zeha: will this cause incorrect stats?

rgacogne: If you do `ContinueAction(DropAction())` then yes, but IMHO you then deserve it :)

PR #8122, opened 2019-07-24
Bail out when no Context library is available

pieterlexis:
### Short description
On MIPS, there is no Boost Context. On OpenWRT, this is combined with a lack of `getcontext` in musl. This means we can never compile there. This PR makes configure do this, instead of waiting for the linker to error after compilation.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


omoerbeek:
This caused a build failure on the build bot so it was reverted.
https://builder.powerdns.com/#/builders/97/builds/857

PR #8128, opened 2019-07-26
auth 4.2: use BIGINT for notified_serial in pg schema

Habbie:
### Short description
Backport of #8097, minus the doc commit.

### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code


Habbie:
> I'd add a note to the merge that this also changes the exit codes of pdnsutil, but otherwise great. This note is mostly defensive in case someone ends up here with a bisect and wonders how this schema change affected pdnsutil return codes.

Right - they'd end up on the commit saying it changed the exit codes, so I think we're good here.

PR #8131, opened 2019-07-27
dnsdist-console: flush cout after printing g_outputBuffer

### LINKED ISSUE #8130: makeKey can not be redirected using shell
- Program: dnsdist
 - Issue type: Bug report

### Short description
I was attempting to use `echo "makeKey()" | dnsdist` to get a usable console key, it works, but when I then did `echo "makeKey()" | dnsdist | grep setKey` it stopped working.

### Environment
 - Operating system: Fedora 30
 - Software version: 1.3.3 (Lua 5.1.4 [LuaJIT 2.1.0-beta3]) with dns-over-tls(openssl) dnscrypt ebpf libsodium protobuf recvmmsg/sendmmsg systemd

### Steps to reproduce
1. `echo "makeKey()" | dnsdist` works
2. `echo "makeKey()" | dnsdist -C /dev/null` works (config is not relevant)
3. `echo "makeKey()" | dnsdist | cat` does not work, because setKey is not printed when piped

### Expected behaviour
The setKey line gets printed.

### Actual behaviour
The setKey does not get printed when stdout/stderr is redirected.

### Other information
`makeKey` is implemented [here](https://github.com/PowerDNS/pdns/blob/f2298404045814f3708e5991aca45452219c6208/pdns/dnsdist-lua.cc#L850) but I can not find where `g_outputBuffer` is printed.

### Usecase
My idea was to make ansible run `echo "makeKey()" | dnsdist  | grep setKey > /etc/dnsdist/key/key.conf` if it didn't exist and then to do a `includeDirectory /etc/dnsdist/key` in the main config file.

### Description
It would be nice if there was a `dnsdist --print-new-key` command that _just_ printed a new key, no reading configs, or starting the server or listening on sockets or anything.

dwfreed:
### Short description

cout may not be a tty, in which case it is block buffered by stdlib, so flush it after printing g_outputBuffer.

Fixes #8130

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdist-console.cc
+++ b/pdns/dnsdist-console.cc
@@ -299,7 +299,7 @@ void doConsole()
           }
         }
         else 
-          cout << g_outputBuffer;
+          cout << g_outputBuffer << std::flush;
         if(!getLuaNoSideEffect())
           feedConfigDelta(line);
       }


Habbie:
I was confused about why this patch makes a difference. The answer is that dnsdist uses `_exit(...)` a lot, which avoids the automatic flush on process exit.

PR #8139, opened 2019-07-29
dnsdist: Add a KeyValueStoreLookup action based on CDB or LMDB

rgacogne:
### Short description

A lookup into a key value store can be done via the `KeyValueStoreLookupAction` action,
using the usual selectors to match the incoming queries for which the lookup should be done.

The first step is to get a `KeyValueStore` object via one of the following functions:

 * `newCDBKVStore` for a CDB database ;
 * `newLMDBKVStore` for a LMDB one.

Then the key used for the lookup can be selected via one of the following functions:

 * the exact qname with `KeyValueLookupKeyQName` ;
 * a suffix match via `KeyValueLookupKeySuffix`, meaning that several lookups will be done, removing one label from the qname at a time, until a match has been found or there is no label left ;
 * the source IP with `KeyValueLookupKeySourceIP` ;
 * the value of an existing tag with `KeyValueLookupKeyTag`.

For example, to do a suffix-based lookup into a LMDB KVS database, the following rule can be used:

  > kvs = newLMDBKVStore('/path/to/lmdb/database', 'database name')
  > addAction(AllRule(), KeyValueStoreLookupAction(kvs, KeyValueLookupKeySuffix(), 'kvs-suffix-result'))

For a query whose qname is "sub.domain.powerdns.com.", and for which only the "\8powerdns\3com\0" key exists in the database, this would result in the following lookups:

 * \3sub\6domain\8powerdns\3com\0
 * \6domain\8powerdns\3com\0
 * \8powerdns\3com\0

Then a match is found for the last key, and the corresponding value is stored into the 'kvs-suffix-result' tag. This tag can now be used in subsequent rules to take an action based on the result of the lookup.

 > addAction(TagRule('kvs-suffix-result', 'this is the value obtained from the lookup'), SpoofAction('2001:db8::1'))

If the value found in the LMDB database for the key '\8powerdns\3com\0' was 'this is the value obtained from the lookup', then the query is immediately answered with a AAAA record.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code


modules/tinydnsbackend/cdb.cc
```diff
@@ -22,33 +22,30 @@
...
 
   memset(&d_cdbf,0,sizeof(struct cdb_find));
   int cdbinit = cdb_init(&d_cdb, d_fd);
   if (cdbinit < 0)
   {
-    g_log<<Logger::Error<<"Failed to initialize cdb structure. ErrorNr: '"<<cdbinit<<endl;
-    throw new PDNSException("Failed to initialize cdb structure.");
+    close(d_fd);
+    d_fd = -1;
+    throw std::runtime_error("Failed to initialize cdb structure. ErrorNt: '" + std::to_string(cdbinit) + "'");
```
zeha: ErrorN_t_?

modules/lmdbbackend/Makefile.am
```diff
@@ -4,6 +4,7 @@ pkglib_LTLIBRARIES = liblmdbbackend.la
 
 EXTRA_DIST = OBJECTFILES OBJECTLIBS
 
-liblmdbbackend_la_SOURCES = lmdbbackend.cc lmdbbackend.hh lmdb-safe.hh lmdb-safe.cc lmdb-typed.hh lmdb-typed.cc
+liblmdbbackend_la_SOURCES = lmdbbackend.cc lmdbbackend.hh lmdb-typed.hh lmdb-typed.cc \
+	../../pdns/lmdb-safe.hh ../../pdns/lmdb-safe.cc
```
zeha: so if lmdb-safe.* come from ahu's repo, and this should continue, should the files live in ext/lmdb-safe/ instead?

modules/tinydnsbackend/cdb.cc
rgacogne: This was the text in the existing code, yes, but unless I'm mistaken I replaced it in a later commit?

modules/lmdbbackend/Makefile.am
rgacogne: That would make sense, yes. I guess we that applies to lmdb-typed.{cc,hh} as well.

rgacogne:
Rebased to fix a conflict in `dnsdist-lua-actions.cc`.

pdns/dnsdistdist/dnsdist-kvs.cc
```diff
@@ -0,0 +1,172 @@
...
+    if (stat(d_fname.c_str(), &st) == 0) {
+      if (st.st_mtime > d_mtime) {
+        reload(st);
+      }
+    }
+    else {
+      warnlog("Error while retrieving the last modification time of CDB database '%s': %s", d_fname, stringerror());
+    }
+    d_nextCheck = now + d_refreshDelay;
+  }
```
chbruyand: I don't think `d_refreshing` is cleared upon successful refresh

pdns/dnsdist-lua-bindings.cc
```diff
@@ -713,4 +714,97 @@ void setupLuaBindings(bool client)
...
+#ifdef HAVE_CDB
+  g_lua.writeFunction("newCDBKVStore", [client](const std::string& fname, time_t refreshDelay) {
+    if (client) {
+      return std::shared_ptr<KeyValueStore>(nullptr);
+    }
+    return std::shared_ptr<KeyValueStore>(new CDBKVStore(fname, refreshDelay));
+  });
+#endif /* HAVE_CDB */
+
+  g_lua.registerFunction<std::string(std::shared_ptr<KeyValueStore>::*)(const std::string&)>("lookup", [](std::shared_ptr<KeyValueStore>& kvs, const std::string& keyStr) {
```
chbruyand: Couldn't we expose multiple `lookup(ComboAddress)`, `lookup(DNSName)` and `lookup(string)` ?

rgacogne: LuaWrapper doesn't support polymorphism so we would need to use a `boost::variant<ComboAddress, DNSName, std::string>`, but sure. That would require creating the correct object beforehand:
```
> kvs:lookup(newDNSName("name.to.lookup."))
> kvs:lookup(newCA("192.0.2.1"))
```
instead of:
```
> kvs:lookup("name.to.lookup.")
> kvs:lookup("192.0.2.1")
```

I agree it would prevent surprises, though, so it's likely a good idea.

gregcockroft:
2 suggestions

It would be nice to limit the depth of the KeyValueLookupKeySuffix match.
Right now a lookup for "cnn.com" looks up "cnn.com" and "com".
This would save a low level lookup in almost all cases.

Matching on wire format names makes debugging harder. It would be nice if this was optional. A flag so that KeyValueLookupKeySuffix::getKeys could use toString() instead of toDNSString().

rgacogne:
Thanks for the feedback!

> It would be nice to limit the depth of the KeyValueLookupKeySuffix match.
> Right now a lookup for "cnn.com" looks up "cnn.com" and "com".
> This would save a low level lookup in almost all cases.

That makes complete sense, I'll likely add that soon.

> Matching on wire format names makes debugging harder. It would be nice if this was optional. A flag so that KeyValueLookupKeySuffix::getKeys could use toString() instead of toDNSString().

Agreed, the reason I used wire format was to remove the ambiguity between something like `\3www\3sub\4main\3tld\0` and `\3www\8sub.main\3tld\0`, but I guess we might not care, or not care by default. Adding a flag would makes sense as well.

rgacogne:
I just pushed a commit adding suffix matching limits and plain text lookups.

rgacogne:
I moved the lmdb-safe files to ext/lmdb-safe/.

modules/lmdbbackend/Makefile.am
rgacogne: Done!

PR #8141, opened 2019-07-29
dnsdist: Add OCSP stapling (from files) for DoT and DoH

### LINKED ISSUE #7812: add support for OCSP stapling
- Program: dnsdist 
 - Issue type: Feature request

### Short description

Adding support for OCSP stapling would:

* reduce latency (no need to contact OCSP server)
* improve privacy (no need to disclose the visit to the 3rd party OCSP server)

### Expected behaviour

DoH and DoT should support OCSP stapling.
And maybe enable it by default?

### Actual behaviour

DoH listener does not appear to support OCSP stapling.
(DoT status not tested)

### Other information

h2o appears to support it:

https://h2o.examp1e.net/index.html 

Key Features: automatic OCSP stapling

rgacogne:
### Short description

The OCSP responses are loaded from files on startup, and reloaded when the `reloadAllCertificates()` command is issued.

Fixes #7812.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/pdns/dnsdistdist/tcpiohandler.cc
+++ b/pdns/dnsdistdist/tcpiohandler.cc
@@ -422,6 +422,7 @@ public:
       SSL_CTX_sess_set_cache_size(d_tlsCtx.get(), fe.d_maxStoredSessions);
     }
 
+    std::vector<int> keyTypes;
     for (const auto& pair : fe.d_certKeyPairs) {
       if (SSL_CTX_use_certificate_chain_file(d_tlsCtx.get(), pair.first.c_str()) != 1) {
         ERR_print_errors_fp(stderr);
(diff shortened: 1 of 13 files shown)


pdns/dnsdistdist/tcpiohandler.cc
```diff
@@ -431,6 +432,25 @@ class OpenSSLTLSIOCtx: public TLSCtx
         ERR_print_errors_fp(stderr);
         throw std::runtime_error("Error loading key from " + pair.second + " for the TLS context on " + fe.d_addr.toStringWithPort());
       }
+      if (SSL_CTX_check_private_key(d_tlsCtx.get()) != 1) {
+        ERR_print_errors_fp(stderr);
+        throw std::runtime_error("Key from '" + pair.second + "' does not match the certificate from '" + pair.first + "' for the TLS context on " + fe.d_addr.toStringWithPort());
+      }
+
+      /* store the type of the new key, we might need it later to select the right OCSP stapling response */
+      keyTypes.push_back(libssl_get_last_key_type(d_tlsCtx));
```
chbruyand: Maybe we should check `libssl_get_last_key_type()` returned something positive ?

rgacogne:
Thanks for the review!

> I'm just wondering how hard it would be to add a basic unit test ?

The issue I encountered was that I could not find any simple way to generate an `OCSP` response for a certificate using the OpenSSL tools. I did see a (complicated) way to set up an `OCSP` server responding to queries so I guess we could do that and use the corresponding client to get a response, but that sounded horribly complicated and brittle.
I did all my testing using OCSP responses from Let's Encrypt for valid certificates.

rgacogne:
Now with regression tests for DoH and DoT OCSP Stapling!
I have to write a bit of code involving undocumented OpenSSL APIs but I hope it will be worth it :-)

PR #8149, opened 2019-07-31
auth: stop using select() in places where FDs can be >1023

Habbie:
### Short description
pipebackend and remotebackend use select() to wait on a single FD. In big setups, these FDs can be >1023, which crashes select().

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/modules/pipebackend/coprocess.cc
+++ b/modules/pipebackend/coprocess.cc
@@ -78,7 +78,7 @@ void CoProcess::launch(const char **argv, int timeout, int infd, int outfd)
     if(!(d_fp=fdopen(d_fd2[0],"r")))
       throw PDNSException("Unable to associate a file pointer with pipe: "+stringerror());
     if( d_timeout)
-      setbuf(d_fp,0); // no buffering please, confuses select
+      setbuf(d_fp,0); // no buffering please, confuses poll
   }
   else if(!d_pid) { // child
     signal(SIGCHLD, SIG_DFL); // silence a warning from perl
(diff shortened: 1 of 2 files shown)


modules/pipebackend/coprocess.cc
```diff
@@ -164,20 +164,11 @@ void CoProcess::receive(string &receive)
...
-    fd_set rds;
-    FD_ZERO(&rds);
-    FD_SET(fileno(d_fp),&rds);
-    int ret=select(fileno(d_fp)+1,&rds,0,0,&tv);
-    if(ret<0)
-      throw PDNSException("Error waiting on data from coprocess: "+stringerror());
-    if(!ret)
-      throw PDNSException("Timeout waiting for data from coprocess");
-  }
+  int ret = waitForRWData(fileno(d_fp), true, 0, d_timeout*1000);
```
rgacogne: Just a nit but I find the use of `waitForData()` easier to read (instead of having to look up what the second parameter to `waitForRWData()` does):

```suggestion
  int ret = waitForData(fileno(d_fp), 0, d_timeout*1000);
```

Habbie: Oh that is better, yes. Changed, tested, amended and force-pushed.

Habbie:
I have now also pushed the fix for the remotebackend pipeconnector. I did not test that commit.

modules/pipebackend/coprocess.cc
```diff
@@ -164,20 +164,11 @@ void CoProcess::receive(string &receive)
...
-    fd_set rds;
-    FD_ZERO(&rds);
-    FD_SET(fileno(d_fp),&rds);
-    int ret=select(fileno(d_fp)+1,&rds,0,0,&tv);
-    if(ret<0)
-      throw PDNSException("Error waiting on data from coprocess: "+stringerror());
-    if(!ret)
-      throw PDNSException("Timeout waiting for data from coprocess");
-  }
+  int ret = waitForData(fileno(d_fp), 0, d_timeout*1000);
```
rgacogne: if `d_timeout` is `0` I believe this would return immediately with `0`?

Habbie: yes

Habbie:
I've redone the whole thing, smaller diffs now, and timeout=0 again waits forever.

paddg:
It looks very good. The patch (4.1.13) is been running for a few days now. No crash so far. Thank you!

rgacogne:
Thanks for the feedback, that's much appreciated!

PR #8153, opened 2019-08-02
dnsdist: Add Lua bindings, rules and action for DoH

### LINKED ISSUE #8133: dnsdist: Support for redirects in DOH
- Program: dnsdist 
 - Issue type: Feature request 

### Short description
Provide support for 3xx-style redirects in DOH query responses

### Usecase
Being able to redirect clients to alternate servers would be useful.  This could be used for load sharing/shedding, policy implementation, geographic re-location, or any of the typical features that are used in HTTP 3xx responses. 

### Description
When a specific pathname regexp is requested, there needs to be a method (Lua, probably) that allows for examination of the path, examination of data about the query (IP address, HTTP headers) and then an ability to respond to the query with one of the 3xx responses that are defined by the HTTP specifications, including a custom URL constructed by logic within the dnsdist configuration model.

One specific use case would be JSON-style DNS responses, which could be handled by an external application/server if requested.

Another use case would be to redirect users to a server containing content about the DNS service policies, general contact information, etc.  As it stands today, a browser query to port 443 on a DOH-enabled dnsdist server results in a non-user-friendly result.

This also could possibly serve as an ugly stand-in for 4xx style responses, so that error messages would be offloaded to a different server which would provide more useful insight into the reason the HTTPS query failed, while leaving dnsdist unburdened from those object delivery tasks.

This was discussed in chat with Habbie on July 5.

rgacogne:
### Short description

This PR adds:
- `HTTPPathRegexRule` to be able to match the path of a DoH query with a regular expression ;
- `HTTPStatusAction` to be able to return a specific HTTP response to a DoH query ;
- Lua bindings (`DNQuestion:getHTTPHeaders()`, `DNSQuestion:getHTTPHost()`, `DNSQuestion:getHTTPPath()`, `DNSQuestion:getHTTPQueryString() and `DNSQuestion:getHTTPScheme()` to be able to inspect a DoH query from Lua ;
- `DNSQuestion:setHTTPResponse()` to be able to return a specific HTTP response from Lua as well.

Fixes #8133.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/pdns/dnsdist-carbon.cc
+++ b/pdns/dnsdist-carbon.cc
@@ -197,6 +197,7 @@ try
               {"post-queries", doh->d_postqueries},
               {"bad-requests", doh->d_badrequests},
               {"error-responses", doh->d_errorresponses},
+              {"redirect-responses", doh->d_redirectresponses},
               {"valid-responses", doh->d_validresponses}
             };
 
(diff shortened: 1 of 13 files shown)


rgacogne:
Travis failures are unrelated (network issues, apparently).

rgacogne:
Rebased to fix a conflict in `dnsdist-lua-actions.cc`.

PR #8154, opened 2019-08-02
dnsdist: Use a counter to mark IDState usage instead of the FD

rgacogne:
### Short description

This change fixes a possible ABA issue and, coupled with stronger checks when updating the state, should reduce the likelihood of any issue when a state is reused from a different thread.

Needs a serious review.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/pdns/dnsdist.cc
+++ b/pdns/dnsdist.cc
@@ -544,11 +544,11 @@ try {
         }
 
         IDState* ids = &dss->idStates[queryId];
-        int origFD = ids->origFD;
+        int64_t usageIndicator = ids->usageIndicator;
 
-        if(origFD < 0) {
+        if(usageIndicator < 0) {
           /* the corresponding state is marked as not in use, meaning that:
-             - it was already reused by another thread and the state is gone ;
+             - it was already cleaned up by another thread and the state is gone ;
              - we already got a response for this query and this one is a duplicate.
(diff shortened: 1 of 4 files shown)


rgacogne:
Travis failures are unrelated (network issues, apparently).

pdns/dnsdist.cc
```diff
@@ -2091,19 +2090,15 @@ static void healthChecksThread()
       dss->prev.reuseds.store(dss->reuseds.load());
       
       for(IDState& ids  : dss->idStates) { // timeouts
-        int origFD = ids.origFD;
-        if(origFD >=0 && ids.age++ > g_udpTimeout) {
-          /* We set origFD to -1 as soon as possible
+        int64_t usageIndicator = ids.usageIndicator;
```
chbruyand: Maybe use `.load()` ?

rgacogne: Isn't that equivalent? Or do you think the use of `.load()` more readable, perhaps?

chbruyand: It is. That was just to match https://github.com/PowerDNS/pdns/pull/8154/files#diff-41f3f8b6ff377ad8fb04d1a7a307a50eR547

PR #8196, opened 2019-08-14
Gracefully handle the mplexer test pipe having only room for one

rgacogne:
### Short description

Many thanks to @stbuehler for the investigation, explanation and reproducer!

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/pdns/test-mplexer.cc
+++ b/pdns/test-mplexer.cc
@@ -122,35 +122,41 @@ BOOST_AUTO_TEST_CASE(test_MPlexer) {
                       &ttd);
 
   /* both should be available */
-  readyFDs.clear();
-  mplexer->getAvailableFDs(readyFDs, 0);
-  if (readyFDs.size() == 1) {
-    /* something is wrong, we need some debug infos */
-    cerr<<"FDMultiPlexer implementation is "<<mplexer->getName()<<endl;
-    cerr<<"Watching "<<mplexer->getWatchedFDCount(false)<<" FDs for read and "<<mplexer->getWatchedFDCount(true)<<" for write"<<endl;
-    cerr<<"pipes[0] is "<<pipes[0]<<endl;
-    cerr<<"pipes[1] is "<<pipes[1]<<endl;
-    cerr<<"The file descriptor returned as ready is "<<readyFDs.at(0)<<endl;
-    char buffer[2];
-    ssize_t res = write(pipes[1], "0", 1);
-    int saved = errno;
-    cerr<<"Writing to pipes[1] returns "<<res<<endl;
-    if (res == -1) {
-      cerr<<"errno is "<<saved<<endl;
-    }
-    res = read(pipes[0], &buffer[0], sizeof(buffer));
-    saved = errno;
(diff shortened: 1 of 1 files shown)


stbuehler:
Looks good to me.

I guess this is a good place to leave a more detailed description :)

A pipe is backed by a ring buffer of pages. When "polling" a pipe (`select`, `poll`, `epoll`, ...) the pipe only signals `EPOLLOUT` if there is at least one page free (see [`pipe_poll` in `fs/pipe.cs`](https://github.com/torvalds/linux/search?q=pipe_poll+path:fs+filename:pipe.c)), although when actually trying to write the pipe it will try to fill the last page too.

Usually a pipe is backed by 16 pages (see definition of [`PIPE_DEF_BUFFERS`](https://github.com/torvalds/linux/search?q=PIPE_DEF_BUFFERS+path:include)), but there is a global kernel limit (across all users, namespaces, containers, ...) for the amount of pages that can be used for pipes.  The default soft limit is `16*1024` pages, i.e. corresponds to `1024` default-sized pipes (a single pipe can be resized (`F_SETPIPE_SZ`) up to `sysctl fs.pipe-max-size`, by default `1MiB`, corresponding to 256 `4KiB` pages or 16 default-sized pipes on systems with `4KiB` page size).  If the soft limit (`sysctl fs.pipe-user-pages-soft`) is hit new pipes will only be backed by a single page.  The hard limit (`sysctl fs.pipe-user-pages-hard`), disabled (`0`) by default, would prevent creation of further pipes.

Anyway, if the soft limit is hit and a pipe is backed only by a single page, writing a single byte to the pipe will make it getting marked "unwritable" in any poll API, although technically you can still write more data.

The following program tries to create many pipes in order to hit the soft limit:
```C
#include <errno.h>
#include <poll.h>
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <sys/time.h>
#include <sys/resource.h>
#include <unistd.h>

#define ERROR(msg) do { \
        fprintf(stderr, "round %d: %s: %s\n", i, msg, strerror(errno)); \
        exit(1); \
    } while (0)

int main() {
	int pipefds[2];
	struct pollfd pfd;

	struct rlimit nofiles;
	if (-1 == getrlimit(RLIMIT_NOFILE, &nofiles)) perror("getrlimit(NOFILE)");
	nofiles.rlim_cur = nofiles.rlim_max;
	if (-1 == setrlimit(RLIMIT_NOFILE, &nofiles)) perror("setrlimit(NOFILE)");

	for (int i = 0; i < 1025; ++i) {
		if (-1 == pipe(pipefds)) ERROR("pipe");

		if (1 != write(pipefds[1], "a", 1)) ERROR("write");

		pfd.fd = pipefds[1];
		pfd.events = POLLOUT;
		poll(&pfd, 1, 0);
		if ((pfd.revents & POLLOUT)) continue;

		printf("Round %i missing POLLOUT revents: %i\n", i, pfd.revents);
		char c;
		read(0, &c, 1);
		break;
	}
}
```

Issue #8106, opened 2019-07-20
recursor fails to use NS to resolve itself under forward-zones


hawken93:
Hi,
I've gotten help from @Habbie on IRC and would like to report a bug with the way recursor handles a certain edgecase.
I don't really know how to describe the issue briefly, but I hope the appendices will help you understand.

 - Program: Recursor
 - Issue type: Bug report

### Short description

My suspicion is that recursor doesn't put the records from the additional section in the cache, but I don't know.

I would like to refer to the steps in the "actual behavior" section below :)

### Environment

 - Operating system:  debian 10 (buster)
 - Software version:  4.1.11 (but tested 4.2 and master also)
 - Software source: operating system repo for 4.1.11, powerdns repo for 4.2 and master

### Steps to reproduce

* A recursor forwards the zone "vpn" to an authorative server
* the vpn zone contains the following records:
  * router.blu.vpn A 10.159.1.1
  * blu.vpn NS router.blu.vpn
* router.blu.vpn is an authorative server for the zone blu.vpn and also contains
  * router.blu.vpn A 10.159.1.1

### Expected behaviour

1. recursor sends the query to the hardcoded nameserver (authorative for the vpn zone)
2. The authorative server sends back authorative: blu.vpn NS router.blu.vpn and additional: router.blu.vpn A 10.159.1.1
3. recursor puts the A record in the cache
4a. recursor sees the record in the cache and returns it to the client.
4b. recursor decides to ask the nameserver about itself, using the A record in the cache to contact it
5b. recursor contacts router.blu.vpn, gets the A record and returns this to the client

### Actual behaviour

1. recursor sends the query to the hardcoded nameserver that contains the vpn zone
2. the authorative server sends blu.vpn NS router.blu.vpn in the authority section and router.blu.vpn A 10.159.1.1 in the additional section
3. recursor sees this and figures out that it is being asked to ask the nameserver about itself, figures out that this should be in the cache already
4. Somehow it is not in the cache so it asks the vpn. server again and gets the same answer
5. This fails, recursor gives up and gives back SERVFAIL to the client

### Other information

* 4.1.0-rc2 was tested by @Habbie (version before #2758) and notes that this version fails even earlier
* I'd like to attach IRC logs from the help that Habbie gave me, it provides extra insight.
https://gist.github.com/hawken93/14e529e50c5e10f2002891a10dad3032 + give an extra thanks, this was very helpful and the attitude positive :)

my configuration:
[...]

PR #8197, opened 2019-08-14
rec: cachonly vs forwarding

omoerbeek:
The cacheonly vs forwarding logic is twisted. If we get an non-aa answer from the forward, continue consulting the cache in a regular way.

#8106 

But not confident yet this is the real fix. Hence the draft status.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/recursordist/test-syncres_cc3.cc
+++ b/pdns/recursordist/test-syncres_cc3.cc
@@ -677,12 +677,10 @@ BOOST_AUTO_TEST_CASE(test_forward_zone_nord) {
         addRecordToLW(res, domain, QType::A, "192.0.2.42");
         return 1;
       }
-
+      // Not reached
+      BOOST_CHECK(false);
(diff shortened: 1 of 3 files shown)


hawken93:
I have built and installed this branch, I'm going to keep running with this for a while and I'll let you know if anything goes horribly wrong :)

rgacogne:
> The cacheonly vs forwarding logic is twisted.

Yes, perhaps we should clean up this part first. I'm not even sure we should consult the servers we forward to in the `cacheonly` case..

omoerbeek:
@hawken93 can you please test the updated code? Thanks!

pieterlexis:
@omoerbeek is this PR still needed? I believe several improvements in this area have been made since August 2019

omoerbeek:
Yes, this is outdated.

PR #8206, opened 2019-08-16
dnsdist: Add support for early DoH HTTP responses

rgacogne:
### Short description

The `HTTPPathRule` and `HTTPPathRegexRule`, and the `Lua` bindings to access the HTTP informations are only invoked for valid DNS over HTTP queries. This PR adds support for responding to HTTP queries _before_ the DNS payload has been parsed, thus allowing to respond to all HTTP queries.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/pdns/dnsdistdist/doh.cc
+++ b/pdns/dnsdistdist/doh.cc
@@ -149,7 +149,111 @@ static void on_socketclose(void *data)
   ctx->release();
 }
 
-/* this duplicates way too much from the UDP handler. Sorry.
+
+static const std::string& getReasonFromStatusCode(uint16_t statusCode)
+{
+  /* no need to care too much about this, HTTP/2 has no 'reason' anyway */
+  static const std::unordered_map<uint16_t, std::string> reasons = {
+    { 200, "OK" },
+    { 301, "Moved Permanently" },
+    { 302, "Found" },
+    { 303, "See Other" },
+    { 304, "Not Modified" },
+    { 305, "Use Proxy" },
+    { 306, "Switch Proxy" },
+    { 307, "Temporary Redirect" },
+    { 308, "Permanent Redirect" },
+    { 400, "Bad Request" },
+    { 401, "Unauthorized" },
+    { 402, "Payment Required" },
+    { 403, "Forbidden" },
+    { 404, "Not Found" },
+    { 405, "Method Not Allowed" },
+    { 406, "Not Acceptable" },
+    { 407, "Proxy Authentication Required" },
+    { 408, "Request Timeout" },
+    { 409, "Conflict" },
+    { 410, "Gone" },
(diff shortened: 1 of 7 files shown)


pdns/dnsdistdist/doh.cc
```diff
@@ -470,6 +574,13 @@ try
 
   string path(req->path.base, req->path.len);
 
+  for (const auto& entry : dsc->df->d_responsesMap) {
+    if (entry->matches(path)) {
+      handleResponse(*dsc->df, req, entry->getStatusCode(), entry->getContent(), std::string());
```
chbruyand: Content-type will be set to `application/dns-message` which may not be what we want

rgacogne:
> I'm just wondering if matching shouldn't be also done on HTTP method ?

It might be a request at some point, yes, but for now I'm trying to keep it simple.

rgacogne:
> I'm also wondering if we may want to send the possibly set custom headers in the response ?

I'd really like dnsdist not to become a full-fledged HTTP server :-)

chbruyand:
> > I'm also wondering if we may want to send the possibly set custom headers in the response ?
> 
> I'd really like dnsdist not to become a full-fledged HTTP server :-)

I know, this is just that you may expect to get the custom headers you set via `customResponseHeaders` in the response.

pdns/dnsdistdist/doh.cc
chbruyand: `Content-type` header will be set by default depending on the desired status code.
This would be `application/dns-message` in case of a 200  response which may not be what we want.

rgacogne:
> I know, this is just that you may expect to get the custom headers you set via `customResponseHeaders` in the response.

Oh, you are right! I'm not 100% sure it's a good idea to apply these to "early" responses, but I guess it might be something that users will expect..

chbruyand:
> > I know, this is just that you may expect to get the custom headers you set via `customResponseHeaders` in the response.
> 
> Oh, you are right! I'm not 100% sure it's a good idea to apply these to "early" responses, but I guess it might be something that users will expect..

Yes, really depends on your custom headers usage. But I think you would expect it in use cases like https://github.com/PowerDNS/pdns/issues/7900

rgacogne:
I'm worried about the case where one does not want the `customResponseHeaders` applied here, so I guess we indeed need to accept an optional list of custom headers in `newDOHResponseMapEntry()`, where no list means to apply `customResponseHeaders` and any other value overrides it (so an empty list means no headers).

rgacogne:
Fixed the default `application/dns-message` `content-type` header added to early 200 responses. Added support for custom HTTP headers in early responses as well.

rgacogne:
Rebased to fix a conflict.

PR #8207, opened 2019-08-16
dnsdist: Add minTLSVersion for DoH and DoH

### LINKED ISSUE #8202: Ability to set not only ciphers but also TLS version in dnsdist
- Program: dnsdist
 - Issue type /Feature request

### Short description
In dnsdist, we can specify the TLS Ciphers to use when negotiating DoT/DoH connections. We should have a parameter to specify the TLS version as well

### Usecase
Some environments have strict security requirements, and the accepted TLS version is often one of those. There might be other reasons as well (library bugs that only manifest themselves in a certain version, a protocol fault that can be mitigated using a more recent TLS version ...

### Description
We can already specify the ciphers to use, as an option. It would be good to be able to specify the TLS version as well. Most other TLS-capable software (nginx, apache, mysql, ...) allow you to specify the TLS version and most (all?) current TLS libraries support setting the version.

rgacogne:
### Short description

Fixes #8202.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


pdns/dnsdistdist/docs/reference/config.rst
```diff
@@ -128,6 +128,7 @@ Listen Sockets
   * ``serverTokens``: str - The content of the Server: HTTP header returned by dnsdist. The default is "h2o/dnsdist".
   * ``customResponseHeaders={}``: table - Set custom HTTP header(s) returned by dnsdist.
   * ``ocspResponses``: list - List of files containing OCSP responses, in the same order than the certificates and keys, that will be used to provide OCSP stapling responses.
+  * ``minTLSVersion``: str - Minimum version of the TLS protocol to support. Possible values are 'tls-1.0', 'tls-1.1', 'tls-1.2' and 'tls-1.3'.
```
franklouwers: In the Docs for addTLSLocal you mention that this parameter doesn't do anything when using gnuTLS. Does that apply here as well? If so:

```suggestion
  * ``minTLSVersion``: str - Minimum version of the TLS protocol to support. Possible values are 'tls-1.0', 'tls-1.1', 'tls-1.2' and 'tls-1.3'. Note that this value is ignored when the GnuTLS provider is in use, and the ``ciphers`` option should be set accordingly instead. For example, 'NORMAL:!VERS-TLS1.0:!VERS-TLS1.1' will disable TLS 1.0 and 1.1.
```

rgacogne: It does not, DoH is OpenSSL-only.

rgacogne:
> Does the remark about gnuTLS apply to addDOHLocal() as well? If so, best to document it.
It does not, DoH only uses OpensSL.

> I can't directly where the default is being set. I assume there's no minimum (well, I hope we're not trying to do sslv2/sslv3)?

The default is TLS 1.0+, I'll document that.

Issue #8205, opened 2019-08-16
rec: Lua feature flags


wojas:
- Program: Recursor (perhaps also dnsdist)
 - Issue type: Feature request

### Short description

Currently it is not always easy to detect from Lua if we are running on a recursor that has certain features enabled or has a certain bugfix.

I suggest adding a global `pdns_features` table to the global Lua namespace with keys that can be checked in Lua code. Every time a certain feature would be added or a bug fixed that would be useful to test for, as key is added to this table.

### Usecase

PR #8001 fixed an issue where setting the `deviceName` would overwrite the `deviceId` in the protobuf instead.

There is however no way to test for the presence of this bugfix and the bug is not fixed in the current 4.2.0. Lua code cannot safely use the API to set the deviceName if it's running on a recursor that runs on an unfixed version, while it would be able to on a newer. The API itself did not change, so we cannot check for the presence of the function call to decide if we can call it.

Exposing the version is not very useful, as this would not give any meaningful results when running development versions from git.

### Description

Add an easy way to define feature flags in the Lua namespace. For example, for PR #8001 and the backport in #8101, we would effectively set the following global:

```lua
pdns_features = {
    PR8001_devicename = true,
}
```

Adding these flags needs to be convenient and should be done every time we add a significant Lua feature we might want to test for.

omoerbeek:
Done.

PR #8210, opened 2019-08-16
rec: Lua pdns_features table

omoerbeek:
### Short description

Introduce a feature table to be able to check Lua features. 
#8205 

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [X] documented the code

### DIFF
--- a/pdns/lua-base4.cc
+++ b/pdns/lua-base4.cc
@@ -29,6 +29,9 @@ void BaseLua4::loadString(const std::string &script) {
   loadStream(iss);
 };
 
+//  By default no features
+void BaseLua4::getFeatures(Features &) { }
+
 #if !defined(HAVE_LUA)
 
 void BaseLua4::prepareContext() { return; }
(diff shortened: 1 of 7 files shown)


pdns/lua-base4.cc
```diff
@@ -42,6 +42,15 @@ BaseLua4::~BaseLua4() { }
...
 
+  // lua features available
+  d_lw->writeVariable("pdns_features",
+                      vector<pair<string, boost::variant<string,bool,int,double> > > {
+  // Add key-values pairs below. Make sure you add string values explicity converted to string.
+  // e.g. { "somekey", string("stringvalue") }
+  // Both int and double end up as a lua number type.
+      { "PR8001_devicename", true },
+    });
+
```
rgacogne: This file is shared with the Auth, so that might not be the right place to define this variable?

Habbie: Or at least, not its contents!

Habbie: From IRC conversation: if we want to share this feature between the different products, maybe the features should be reported into this table at runtime. For example, the one auth 'feature' I can think of right now is 'geoip is available and configured', which we don't know at compile time. Even if the recursor does know everything at compile time, filling the table at startup would make `pdns_features` "portable" between the products.

Alternatively, go with with @rgacogne said and move it into a Recursor-specific file, we can always think about features reporting for Auth (and dnsdist?) later.

omoerbeek: I adapted the setup. Base class now has the common stuff. It calls a method that can be overriden to define the product specifics values.

pdns/lua-recursor4.cc
```diff
@@ -460,6 +460,14 @@ void RecursorLua4::postLoad() {
   d_gettag_ffi = d_lw->readVariable<boost::optional<gettag_ffi_t>>("gettag_ffi").get_value_or(0);
 }
 
+void RecursorLua4::getFeatures(Features & features) {
+  // Add key-values pairs below.
+  // Make sure you add string values explicity converted to string.
+  // e.g. features.push_back(make_pair("somekey", string("stringvalue"));
+  // Both int and double end up as a lua number type.
+   features.push_back(make_pair("PR8001_devicename", true));
```
rgacogne: Just a nit but I would have expected this function to return a `Features` object, perhaps as a const reference, rather than take it as an out parameter. That doesn't matter, though.

omoerbeek: I have done it this way to allow for a more easy way to add features in several places. Not needed now, but maybe in the future. Each method call will then add its features to an existing vector.

Issue #7439, opened 2019-02-01
tinydns backend timestamp generates huge TTLs


smellyspice:
### Description
The timestamp feature of tinydns allows for the expiry of records at a certain time. If you set this time way into the future (say 2030-01-01), stock tinydns will limit the TTL to 3600 secs. However, with the `pdns tinydns backend`, the actual `timestamp - now` value is inserted into the `Answer` section, resulting in an enormous TTL being sent back.  
### Environment
PowerDNS version: `PowerDNS Authoritative Server 4.1.1`
Where you got the software from: `Ubuntu repo`
Operating System and version: `Ubuntu 18.04`

### Steps to reproduce
Given a Data file with this data: 
`+www.lanechange.net:52.60.72.197:0:4000000070dbd880`

**Expected behavior**: A query for this data from a vanilla `tinydns` server returns: 
`lanechange.net.         3600    IN      A       52.60.72.197`

**Observed behavior**: Same query against `pdns tinydns backend` results in: 
`lanechange.net.         344475198 IN    A       52.60.72.197`

### Recommended Solution - Implement stock djbdns logic
Stock `tdlookup.c` from djbdns has 2 checks for extremely small or large values to limit the advertised TTLs when `timestamps` are being used. You can see that on lines 79 & 80 here: 
https://github.com/abh/djbdns/blob/0715ca5de14ad5eae2c779c536beee2ef56d07a2/tdlookup.c#L75-L82

TTLs < 2 seconds are fixed at 2 seconds and values > 3600 seconds are capped at 3600 seconds. 

I believe this same logic needs to be inserted in and around here: 
https://github.com/PowerDNS/pdns/blob/08a373a10af4f7dbf44576f4af1677f6cb07a038/modules/tinydnsbackend/tinydnsbackend.cc#L287

Habbie:
Fixed in #8225

PR #8225, opened 2019-08-21
timestamp TTL limiting to fix Issue #7439 - Take 2

smellyspice:
### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code


Habbie:
@smellyspice your persistence in navigating git, github, your editor, etc. in this is appreciated! I planned to, if necessary, get your contribution into our tree by hand, because I think the fix looks good and tooling issues should not prevent a fix from going in. I intend to get this in next week, no matter what.

smellyspice:
@Habbie - Thanks for the good words, it's nice to know that my efforts and screw-ups are taken in stride! I really didn't know how else to proceed other than to start fresh after my editor ended up reformatting that file. It is a simple fix, which I've been running in production for over 6 months with no issues, so it would be nice to get it into master. Thanks!

Habbie:
> I always feel that boundary conditions checks should be < and > not <= or >=

I agree, but in this specific case, it mirrors code from the original tinydns:

```
tdlookup.c-     if (newttl <= 2.0) newttl = 2.0;
tdlookup.c:     if (newttl >= 3600.0) newttl = 3600.0;
```

which also has readability benefits.

Habbie:
I have a testcase for this sitting on a branch, will add it soon.

pieterlexis:
@Habbie **poke**

Habbie:
The 'large ttl' test i wrote turns out to not exercise this code. I'll push it to this branch anyway, to show that this code does not affect non-timestamp large TTLs.

PR #8240, opened 2019-08-28
Fix some conditionally added source files missing from extra dist

rgacogne:
### Short description

They are conditionally added to `dnsdist_SOURCES` or `pdns_recursor_SOURCES` so we need to add them to `EXTRA_DIST` in order for them to be included in the dist even if the relevant dependencies are not available at `make dist` time.

### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code


rgacogne:
It looks like we don't actually need this, as found by @pieterlexis :

> When using conditionals like this you don’t need to use the EXTRA_ variable, because Automake will examine the contents of each variable to construct the complete list of source files.

from https://www.gnu.org/software/automake/manual/html_node/Conditional-Sources.html

PR #8245, opened 2019-08-28
dns_random: verify upper bound; fix boundary condition for QID generation

Habbie:
### Short description
Some of our `dns_random` usage looked like it expected the parameter to be an inclusive upper bound. This PR adds a test that ensures the upper bound is exclusive. The common `dns_random(0xffff)` pattern, which was off by one, is replaced by a single function with a clear name.

Reported by Matt Nordhoff, thanks!

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dns_random.hh
+++ b/pdns/dns_random.hh
@@ -25,5 +25,6 @@
 
 void dns_random_init(const std::string& data = "", bool force_reinit = false);
 uint32_t dns_random(uint32_t n);
+uint16_t dns_random_uint16();
 
 #endif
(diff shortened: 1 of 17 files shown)


pdns/dns_random.hh
```diff
@@ -25,5 +25,6 @@
 
 void dns_random_init(const std::string& data = "", bool force_reinit = false);
 uint32_t dns_random(uint32_t n);
+uint16_t dns_random_uint16();
```
pieterlexis: should this perhaps be `inline`?

mnordhoff:
There's also at least one in calidns.cc:

https://github.com/PowerDNS/pdns/blob/4fac301bd7fcd9fb2d8c3d056f8cb0a00958d185/pdns/calidns.cc#L397

Habbie:
> There's also at least one in calidns.cc:

Amended, thanks!

PR #8246, opened 2019-08-29
fix a typo in comment in syncres.cc: 'FRC 7129' to 'RFC 7129'

Enna1:
There is a typo in comment in "/pdns/syncres.cc:2883" :
> as described in section 5.3.4 of RFC 4035 and 5.3 of FRC 7129
The corrected comment should be:
"as described in section 5.3.4 of RFC 4035 and 5.3 of RFC 7129"


rgacogne:
I don't really understand what happened with the `ci/circleci: test-dnsdist-regression` test, it looks like the `build-dnsdist` step was executed before the merge of #8139 and the `test-dnsdist-regression` one after.. Not related to this PR anyway.

PR #8256, opened 2019-08-30
rec: fix two coverity issues

omoerbeek:
1. Uninitialized field in DNSComboWriter.
2. TOCTOU wrt stat() and then opendir(). opendir() already tells us if the path
is not a dir or not accesible, so no need to do a TOCTOU sensitive pre-check.
Use uniform error messages in log and exceptions while here.

### Short description

** CID 1404735:  Uninitialized members  (UNINIT_CTOR)
** CID 1401970:  Security best practices violations  (TOCTOU)

With some whitespace fixes, since the indentation was completely wrong

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/arguments.cc
+++ b/pdns/arguments.cc
@@ -460,43 +460,35 @@ bool ArgvMap::file(const char *fname, bool lax, bool included)
 
 void ArgvMap::gatherIncludes(std::vector<std::string> &extraConfigs) {
   extraConfigs.clear();
-  if (params["include-dir"].empty()) return; // nothing to do
-    struct stat st;
-    DIR *dir;
-    struct dirent *ent;
+  if (params["include-dir"].empty())
+    return; // nothing to do
 
-    // stat
-    if (stat(params["include-dir"].c_str(), &st)) {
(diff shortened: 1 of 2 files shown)


pdns/arguments.cc
```diff
@@ -460,43 +460,35 @@ bool ArgvMap::file(const char *fname, bool lax, bool included)
...
-    }
+  if (params["include-dir"].empty())
+    return; // nothing to do
+
+  DIR *dir;
+  if (!(dir = opendir(params["include-dir"].c_str()))) {
+    int err = errno;
+    string msg = params["include-dir"] + " is not accessible: " + strerror(err);
+    g_log << Logger::Error << msg << std::endl;
+    throw ArgException(msg);
```
rgacogne: Not a big deal since I believe we will exit right away, but this would leak the `DIR*`.

omoerbeek: I don't think so since the opendir() call failed...

rgacogne: Oh, right, sorry about that!

Issue #8263, opened 2019-09-02
Feature request: DoH frontend but plain HTTP


Jamesits:
- Program: dnsdist
 - Issue type: Feature request

### Short description
Currently the DoH implementation in 1.4.0-alpha2 requires TLS to use DoH. I'm suggesting the TLS part to be optional (i.e. DoH over plain HTTP protocol). 

### Usecase

We'd like to use dnsdist as a DoH server, but behind an SSL-offloading reverse proxy (e.g. Nginx). This kind of setup is very common in modern infrastructure (for example, in a Kubernetes cluster), and works better with automatic cert signing service (Let's Encrypt, etc.). Surely we can create a self-signed cert for dnsdist, but in this case, doing another TLS handshake from reverse proxy to dnsdist does not benefit much and wastes a lot of energy. 

### Description

* make arguments `certFile` and `keyFile` optional for `addDOHLocal` function definition
* If the 2 arguments are not set, use plain HTTP mode to communicate with the server
* And set the default port to 80 maybe?

PR #8267, opened 2019-09-02
dnsdist: Allow accepting DoH queries over HTTP instead of HTTPS

rgacogne:
### Short description

It allows using dnsdist behind a reverse-proxy or any other device offloading TLS.
Closes https://github.com/PowerDNS/pdns/issues/8263.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdistdist/docs/reference/config.rst
+++ b/pdns/dnsdistdist/docs/reference/config.rst
@@ -103,11 +103,12 @@ Listen Sockets
                                   higher than 0 to enable TCP Fast Open when available.
                                   Default is 0.
 
-.. function:: addDOHLocal(address, certFile(s), keyFile(s) [, urls [, options]])
+.. function:: addDOHLocal(address, [certFile(s) [, keyFile(s) [, urls [, options]]]])
 
   .. versionadded:: 1.4.0
 
   Listen on the specified address and TCP port for incoming DNS over HTTPS connections, presenting the specified X.509 certificate.
+  If no certificate (or key) files are specified, listen for incoming DNS over HTTP connections instead.
 
   :param str address: The IP Address with an optional port to listen on.
                       The default port is 443.
--- a/pdns/dnsdist-lua.cc
+++ b/pdns/dnsdist-lua.cc
@@ -1678,7 +1678,7 @@ void setupLuaConfig(bool client)
     setSyslogFacility(facility);
   });
 
-  g_lua.writeFunction("addDOHLocal", [client](const std::string& addr, boost::variant<std::string, std::vector<std::pair<int,std::string>>> certFiles, boost::variant<std::string, std::vector<std::pair<int,std::string>>> keyFiles, boost::optional<boost::variant<std::string, vector<pair<int, std::string> > > > urls, boost::optional<localbind_t> vars) {
+  g_lua.writeFunction("addDOHLocal", [client](const std::string& addr, boost::optional<boost::variant<std::string, std::vector<std::pair<int,std::string>>>> certFiles, boost::optional<boost::variant<std::string, std::vector<std::pair<int,std::string>>>> keyFiles, boost::optional<boost::variant<std::string, vector<pair<int, std::string> > > > urls, boost::optional<localbind_t> vars) {
 #ifdef HAVE_DNS_OVER_HTTPS
     if (client) {
       return;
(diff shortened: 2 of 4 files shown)


bop1991:
Like

Habbie:
> Like

Please use the emoji button instead of comment like, thank you!

pdns/dnsdistdist/docs/reference/config.rst
```diff
@@ -103,11 +103,12 @@ Listen Sockets
                                   higher than 0 to enable TCP Fast Open when available.
                                   Default is 0.
 
-.. function:: addDOHLocal(address, certFile(s), keyFile(s) [, urls [, options]])
+.. function:: addDOHLocal(address, [certFile(s) [, keyFile(s) [, urls [, options]]]])
```
Jamesits: I think it should be 

```
function:: addDOHLocal(address, [certFile(s) , keyFile(s) [, urls [, options]]])
```

Since if you have `certFile` then you should have `keyFile` too.

cmouse:
Should this come with some check(s) that the remote actually offloaded SSL? Or at least support for such thing?

krombel:
I would not require such header nor would I add such restriction.
The person which is configuring this should know what it is doing.
I think the log output that encryption is disabled (as done) is sufficient.

pdns/dnsdistdist/docs/reference/config.rst
rgacogne: In practice that's pretty much what the code does, it requires the same number of keys.

rgacogne:
> Should this come with some check(s) that the remote actually offloaded SSL? Or at least support for such thing?
> 
> HAPROXY supports sending TLS status with in version 2 header. Apache/nginx can send header(s) that inform about TLS usage. Also these connections should not be accepted from untrusted IP sources.

I think we should let the administrator configure the ACL properly, and accept plaintext connections when told to do so.

christianbur:
Could someone please tell me how I can use DNS over HTTP.

DNS over HTTP with haproxy runs like this 
```
addDOHLocal("0.0.0.0:443", {"/dnsdist_certs/rsa/fullchain.cer", "/dnsdist_certs/ecdsa/fullchain.cer"}, {"/dnsdist_certs/rsa/example.de.key", "/dnsdist_certs/ecdsa/example.de.key"}, "/doh_654321", {serverTokens="h2o/dnsdist", tcpFastOpenSize=0, reusePort=true})
addDOHLocal("[::]:443",    {"/dnsdist_certs/rsa/fullchain.cer", "/dnsdist_certs/ecdsa/fullchain.cer"}, {"/dnsdist_certs/rsa/example.de.key", "/dnsdist_certs/ecdsa/example.de.key"}, "/doh_654321", {serverTokens="h2o/dnsdist", tcpFastOpenSize=0, reusePort=true})
```

DNS over HTTP (HTTPS with haproxy) -- but the following two lines do not work. 
```
addDOHLocal("0.0.0.0:80",,, "/doh_654321", {serverTokens="h2o/dnsdist", tcpFastOpenSize=0, reusePort=true})
addDOHLocal("[::]:80",   ,, "/doh_654321", {serverTokens="h2o/dnsdist", tcpFastOpenSize=0, reusePort=true})
```
The [documentary](https://dnsdist.org/reference/config.html) says "If no certificate (or key) files are specified, listen for incoming DNS over HTTP connections instead."

But how do I define "no certificate"?

- addDOHLocal("0.0.0.0:80",,, "/doh_654321", ..  
    - the start of dnsdist fails
    - Fatal error: [string "chunk"]:25: unexpected symbol near ','

- addDOHLocal("0.0.0.0:80", "", "", "/doh_654321", ..  -
    - the start of dnsdist fails
    - 139740307332928:error:02001002:system library:fopen:No such file or directory:../crypto/bio/bss_file.c:288:fopen('','r')
 
- addDOHLocal("0.0.0.0:80", "/doh_654321", ..  
    -  dnsdist starts, but I don't get a doh answer.
    - Fatal Lua error: [string "chunk"]:25: Unable to convert parameter from table to N5boost8optionalINS_7variantINSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEEJSt6vectorISt4pairIiS7_ESaISA_EEEEEEE

- addDOHLocal("0.0.0.0:80",  {}, {}, "/doh_654321", ..  
    - dnsdist starts, but I don't get a doh answer.

Habbie:
> Could someone please tell me how I can use DNS over HTTP.

Please find us on IRC or one of our mailinglists, via https://www.powerdns.com/opensource.html, so that we can help you. We do not provide support via GitHub.

PR #8270, opened 2019-09-03
Parameterize dnstap configure default (rec=no, dndist=auto)

omoerbeek:
### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code


pieterlexis:
You might need to set `--disable-dnstap` in `configure` for the `rules` and `*.spec` files

omoerbeek:
Changing the default is good enough. This way, we only have to change one line if we want to enable it.

PR #8271, opened 2019-09-04
auth: add configurable timeout for inbound AXFR

qvr:
### Short description
Adds a configurable timeout for inbound AXFR. Default is 10 seconds which it was previously hardcoded to.

Needed when transferring large zones from slow masters which take a while before sending anything. Same timeout setting is also applied for subsequent `getChunk` calls.

Fun fact: bind has a default timeout of 2 hours for the whole AXFR, and 1 hour for stalls during transfer.

### Checklist
I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/pdns/slavecommunicator.cc
+++ b/pdns/slavecommunicator.cc
@@ -241,13 +241,14 @@ static bool processRecordForZS(const DNSName& domain, bool& firstNSEC3, DNSResou
 
 static vector<DNSResourceRecord> doAxfr(const ComboAddress& raddr, const DNSName& domain, const TSIGTriplet& tt, const ComboAddress& laddr,  scoped_ptr<AuthLua4>& pdl, ZoneStatus& zs)
 {
+  uint16_t axfr_timeout=::arg().asNum("axfr-fetch-timeout");
   vector<DNSResourceRecord> rrs;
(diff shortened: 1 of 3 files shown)


pdns/slavecommunicator.cc
```diff
@@ -242,12 +242,12 @@ static bool processRecordForZS(const DNSName& domain, bool& firstNSEC3, DNSResou
...
 {
   vector<DNSResourceRecord> rrs;
-  AXFRRetriever retriever(raddr, domain, tt, (laddr.sin4.sin_family == 0) ? NULL : &laddr, ((size_t) ::arg().asNum("xfr-max-received-mbytes")) * 1024 * 1024);
+  AXFRRetriever retriever(raddr, domain, tt, (laddr.sin4.sin_family == 0) ? NULL : &laddr, ((size_t) ::arg().asNum("xfr-max-received-mbytes")) * 1024 * 1024, ::arg().asNum("axfr-fetch-timeout"));
   Resolver::res_t recs;
   bool first=true;
   bool firstNSEC3{true};
   bool soa_received {false};
-  while(retriever.getChunk(recs)) {
+  while(retriever.getChunk(recs, 0, ::arg().asNum("axfr-fetch-timeout"))) {
```
Habbie: The `::arg().asNum(..)` call is not free - perhaps you can stick it in a `uint16_t` at the beginning of the function?

chbruyand: Also we prefer using `nullptr` over `0` when dealing with pointers

PR #8289, opened 2019-09-10
 pdnsutil increase-serial: under SOA-EDIT=INCEPTION-EPOCH, bump as if it is EPOCH

### LINKED ISSUE #8218: auth: pdnsutil increase-serial misbehaving under INCEPTION-EPOCH
- Program: Authoritative
 - Issue type: Surprising behaviour

### Short description
With `SOA-EDIT=INCEPTION-EPOCH`, `pdnsutil increase-serial` will bump the serial to the latest inception if necessary, and otherwise increase the existing number by one. Users might expect, instead, that the number is always bumped to the current UNIX time.

Habbie:
### Short description
Fixes #8218.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/docs/dnsupdate.rst
+++ b/docs/dnsupdate.rst
@@ -236,6 +236,8 @@ These are the settings available for **SOA-EDIT-DNSUPDATE**.
 -  SOA-EDIT-INCREASE: Change the serial to whatever SOA-EDIT would
    provide. If what SOA-EDIT provides is lower than the current serial,
    increase the current serial by 1.
+   Exception: with SOA-EDIT=INCEPTION-EPOCH, the serial is bumped to at
+   least the current EPOCH time.
 
 DNS update How-to: Setup dyndns/rfc2136 with dhcpd
 --------------------------------------------------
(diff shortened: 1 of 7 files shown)


zeha:
Well, maybe this should be done more generally. We have documented this:

```
SOA-EDIT-INCREASE: Change the serial to whatever SOA-EDIT would provide. If what SOA-EDIT provides is lower than the current serial, increase the current serial by 1.
```

At the very least I think this needs a doc update; I'm not sure right now what is the least confusing behaviour.

Habbie:
> Well, maybe this should be done more generally.

I agree, but I wasn't so confident that any other types were doing a surprising thing, so I figured I'd do a narrow fix here and give it some more thought for 4.3.

> `SOA-EDIT-INCREASE: Change the serial to whatever SOA-EDIT would provide. If what SOA-EDIT provides is lower than the current serial, increase the current serial by 1.`

Right, and my PR deviates from that. Then again, we only documented that for rfc2136..

zeha:
> Right, and my PR deviates from that. Then again, we only documented that for rfc2136..

True, but the docs do refer to the rfc2136 rules when explaining related options. I couldn't find a separate explanation for SOA-EDIT-INCREASE in any case.

Habbie:
> True, but the docs do refer to the rfc2136 rules when explaining related options. I couldn't find a separate explanation for SOA-EDIT-INCREASE in any case.

Fair enough. I'll think about how I can make the docs clear on what I did here.

Habbie:
Doc update pushed.

PR #8296, opened 2019-09-13
webserver: handle exceptions instead of SIGABRTing the world

Habbie:
### Short description
Without this, any exception that manages to bubble out of `serveConnection` will take down the whole pdns_server (or ixfrdist or recursor or ...) process with SIGABRT. Except on 'Linux', the user would also be left none the wiser as to the reason of the crash.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/webserver.cc
+++ b/pdns/webserver.cc
@@ -202,7 +202,18 @@ void WebServer::registerWebHandler(const string& url, HandlerFunction handler) {
 
 static void *WebServerConnectionThreadStart(const WebServer* webServer, std::shared_ptr<Socket> client) {
   setThreadName("pdns-r/webhndlr");
-  webServer->serveConnection(client);
+  try {
+    webServer->serveConnection(client);
+  }
+  catch(PDNSException &e) {
+    g_log<<Logger::Error<<"PDNSException while serving a connection in main webserver thread: "<<e.reason<<endl;
+  }
+  catch(std::exception &e) {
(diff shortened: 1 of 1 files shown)


Habbie:
I used this patch to test the PR:
```diff
diff --git a/pdns/uuid-utils.cc b/pdns/uuid-utils.cc
index 12340aefc..fd8c54bc5 100644
--- a/pdns/uuid-utils.cc
+++ b/pdns/uuid-utils.cc
@@ -36,7 +36,9 @@ thread_local boost::uuids::random_generator t_uuidGenerator;
 
 boost::uuids::uuid getUniqueID()
 {
-  return t_uuidGenerator();
+                   BOOST_THROW_EXCEPTION(boost::uuids::entropy_error(5, "getentropy"));
+
+ return t_uuidGenerator();
 }
 
 boost::uuids::uuid getUniqueID(const std::string& str)
```

as the SIGABRT was discovered by a user that managed to throw `entropy_error` by not having `/dev/urandom` on FreeBSD.

zeha:
AFAICT all these exceptions - except as caused by the UUID generator - are already being caught inside handleRequest. the code dup is a bit of a pity :(

rgacogne:
The code handling these exceptions in `serveConnection()` uses the `UUID` when logging so it was likely not straightforward. Handling them there also reduces the likelihood of messing up later when introducing new code in `serveConnection()`.

zeha:
Yes, as discussed with @Habbie on IRC, the extra "exception boundary" is a good thing in any case also from my PoV.

PR #8325, opened 2019-09-20
auth API: make disabled optional for Record

pieterlexis:
# Short description
`disabled` defaults to false.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/ws-auth.cc
+++ b/pdns/ws-auth.cc
@@ -537,7 +537,10 @@ static void gatherRecords(const string& logprefix, const Json container, const D
   const auto& items = container["records"].array_items();
   for(const auto& record : items) {
     string content = stringFromJson(record, "content");
-    rr.disabled = boolFromJson(record, "disabled");
+    rr.disabled = false;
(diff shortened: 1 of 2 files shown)


pdns/ws-auth.cc
```diff
@@ -537,7 +537,10 @@ static void gatherRecords(const string& logprefix, const Json container, const D
   const auto& items = container["records"].array_items();
   for(const auto& record : items) {
     string content = stringFromJson(record, "content");
-    rr.disabled = boolFromJson(record, "disabled");
+    rr.disabled = false;
+    if(!record["disabled"].is_null()) {
+      rr.disabled = boolFromJson(record, "disabled");
```
zeha: `boolFromJson(record, "disabled", false);`?

pieterlexis: That will send an APIError when `disabled` is not a bool

zeha: Right. That's probably a bug in all of the <type>FromJson functions.

PR #8344, opened 2019-09-25
Add signal handling for SIGTERM and SIGINT in pdns_recursor, if we are PID1

franklouwers:
### Short description
The Linux kernel handles signals for PID 1 processes differently. It
doesn't implement a default handler for some signals such as
SIGTERM/SIGINT.

When running pdns_recursor as a container, this causes a few annoyances.
You can work around those by running your containers with --init or by
installing `tini` inside the container. Or you can handle the signals in
the application itself.

This commit adds signal() handlers for SIGTERM and SIGINT for
pdns_recursor, if we are running as PID 1.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/pdns_recursor.cc
+++ b/pdns/pdns_recursor.cc
@@ -2786,6 +2786,11 @@ static void daemonize(void)
   }
 }
 
+static void termIntHandler(int)
+{
+  doExitNicely();
+}
+
 static void usr1Handler(int)
 {
   statsWanted=true;
@@ -4075,6 +4080,8 @@ static int serviceMain(int argc, char*argv[])
     g_log.toConsole(Logger::Critical);
     daemonize();
   }
+  signal(SIGTERM,termIntHandler);
+  signal(SIGINT,termIntHandler);
   signal(SIGUSR1,usr1Handler);
   signal(SIGUSR2,usr2Handler);
   signal(SIGPIPE,SIG_IGN);
--- a/pdns/pdns_recursor.cc
+++ b/pdns/pdns_recursor.cc
@@ -4081,7 +4081,19 @@ static int serviceMain(int argc, char*argv[])
     daemonize();
   }
   if(Utility::getpid() == 1) {
-    /* We are running as pid 1, register sigterm and sigint handler */
+    /* We are running as pid 1, register sigterm and sigint handler
+     
+      The Linux kernel will handle SIGTERM and SIGINT for all processes, except PID 1.
+      It assumes that the processes running as pid 1 is an "init" like system.
+      For years, this was a safe assumption, but containers change that: in
+      most (all?) container implementations, the application itself is running
+      as pid 1. This means that sending signals to those applications, will not
+      be handled by default. Results might be "you container not responsing
+      when asking it to stop", or "ctrl-c" not working even when the app is
+      running in the foreground inside a container.
+
+      So TL;DR: If we're running pid 1 (container), we should handle SIGTERM and SIGINT ourselves */
+
     signal(SIGTERM,termIntHandler);
     signal(SIGINT,termIntHandler);
   } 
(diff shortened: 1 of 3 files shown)


franklouwers:
Please verify carefully. It has been over 20 years since I last wrote c++ code :)

pdns/pdns_recursor.cc
```diff
@@ -2786,6 +2786,11 @@ static void daemonize(void)
   }
 }
 
+static void termIntHandler(int)
+{
+  doExitNicely();
```
rgacogne: I think it would be better to call `doExit()` instead, I'm worried about race conditions between threads with the use of `exit()`, where one thread tries to access a global structure already destroyed by another, for example.

omoerbeek: Agree. Calling exit(3) from a signal handler is not safe. 

Even doing stream I/O (which might lead to heap allocations or deallocations) is already dangerous: if the signal is caught while the heap is in an inconsistent state (e.g. inside a new) more heap operations will lead to big trouble.
`doExit()` is already walking a thin rope...

franklouwers: Thanks for the comments! I'll do some research and see what I can up with. However, as the main use-case for this fix is Docker, where the container (including all memory/heap allocated to it) will be stopped when the PID 1 (pdns_recursor) stops, this might be not so much of an issue. 

Maybe it would make sense to determine the PID, and only register the handlers if we are PID 1?

Habbie: > Maybe it would make sense to determine the PID, and only register the handlers if we are PID 1?

Please let us not dig us even deeper into a rabbit hole for what is an avoidable Docker bug.

franklouwers: oh sigh. Just close it then. Again: it is not a docker bug. It's Linux kernel behaviour: if pid1: you don't get (all) signals. Unless you want to convince Linus to change that behaviour...

rgacogne: Ideally we should just set a `sig_atomic_t` flag from the signal handler and call `doExit()` in the main event loop later, but I'm fine with calling `doExit()`. Worst case scenario it causes an abort or a crash, but we were exiting anyway.

omoerbeek: Agreed. Call doExit() for now. I'll file an issue to do proper shutdown (as rgacogne suggests).

franklouwers: @omoerbeek @rgacogne Ok thanks. I'll change the code. Do you prefer I put some "if we are pid 1" logic around the handler registration?

pdns/pdns_recursor.cc
```diff
@@ -4075,6 +4080,12 @@ static int serviceMain(int argc, char*argv[])
     g_log.toConsole(Logger::Critical);
     daemonize();
   }
+  if(Utility::getpid() == 1) {
+    /* We are running as pid 1, register sigterm and sigint handler */
```
omoerbeek: A mention of why pid 1 is special would be nice. On "normal" systems pid 1 mean we're the init process, so testing on 1 might confuse the reader.

franklouwers: See additional code comments

pdns/pdns_recursor.cc
```diff
@@ -4075,6 +4080,24 @@ static int serviceMain(int argc, char*argv[])
...
   }
+  if(Utility::getpid() == 1) {
+    /* We are running as pid 1, register sigterm and sigint handler
+     
+      The Linux kernel will handle SIGTERM and SIGINT for all processes, except PID 1.
+      It assumes that the processes running as pid 1 is an "init" like system.
+      For years, this was a safe assumption, but containers change that: in
+      most (all?) container implementations, the application itself is running
+      as pid 1. This means that sending signals to those applications, will not
+      be handled by default. Results might be "you container not responsing
```
phonedph1: Minor typo here, s/you/your/

PR #8349, opened 2019-09-26
dnsdist: Implement TLS Session Ticket Keys management for DoH

rgacogne:
### Short description

<strike>Still needs regression tests (in progress).</strike>

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/pdns/doh.hh
+++ b/pdns/doh.hh
@@ -48,11 +48,22 @@ struct DOHFrontend
   std::string d_ciphers13;
   std::string d_serverTokens{"h2o/dnsdist"};
   LibsslTLSVersion d_minTLSVersion{LibsslTLSVersion::TLS10};
+#ifdef HAVE_DNS_OVER_HTTPS
+  std::unique_ptr<OpenSSLTLSTicketKeysRing> d_ticketKeys{nullptr};
+#endif
   std::vector<std::pair<std::string, std::string>> d_customResponseHeaders;
   ComboAddress d_local;
 
   uint32_t d_idleTimeout{30};             // HTTP idle timeout in seconds
   std::vector<std::string> d_urls;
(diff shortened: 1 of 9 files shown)


rgacogne:
We should make sure that dnsdist still compiles without libssl after these changes, I realize I didn't test that!

pdns/doh.hh
```diff
@@ -48,11 +48,22 @@ struct DOHFrontend
...
   LibsslTLSVersion d_minTLSVersion{LibsslTLSVersion::TLS10};
+#ifdef HAVE_DNS_OVER_HTTPS
+  std::unique_ptr<OpenSSLTLSTicketKeysRing> d_ticketKeys{nullptr};
+#endif
   std::vector<std::pair<std::string, std::string>> d_customResponseHeaders;
   ComboAddress d_local;
 
   uint32_t d_idleTimeout{30};             // HTTP idle timeout in seconds
   std::vector<std::string> d_urls;
+  std::string d_ticketKeyFile;
```
chbruyand: Maybe also wrap these new members in a `#ifdef HAVE_DNS_OVER_HTTPS #endif` block ?

pdns/doh.hh
```diff
@@ -82,6 +93,7 @@ struct DOHFrontend
   HTTPVersionStats d_http1Stats;
   HTTPVersionStats d_http2Stats;
 
+
```
chbruyand: To remove if any other modification in this file.

pdns/doh.hh
rgacogne: I don't like having a lot of `#ifdef` block around, so as long as it doesn't prevent compiling when DoH support is disabled I'd rather leave them alone.

Issue #8161, opened 2019-08-05
Authoritative server 4.1.10 memory leak with bind backend


sthaug:
- Program: Authoritative server version 4.1.10 
 - Issue type: Bug report 

### Short description

- Pdns configured as slave nameserver for the "no" zone, with a BIND server as master. The "no" zone is reasonably large - currently 3512726 lines and 354990510 bytes as stored by Pdns. The "no" zone contains DNSSEC information.
- Pdns configured with bind backend.
- When the master name server sends a notify, the "no" zone is transferred and stored. However, I can also see from "ps" that the memory size of the pdns process increases with more than a Gigabyte for each transfer of the zone from the master. Eventually the pdns process is killed because the system runs out of swap.

### Environment

 - Operating system: FreeBSD 11.3-STABLE and FreeBSD 12.0-STABLE (same problem for both).
 - Software version: Pdns authoritative 4.1.10.
 - Software source: Operating system repository (installed by the FreeBSD pkg system)

### Steps to reproduce

1. Install Pdns. "pkg install powerdns-4.1.10_3"
2. Create Sqlite3 database (because DNSSEC): pdnsutil create-bind-db /usr/local/etc/pdns/data/bind-dnssec-db.sqlite3
3. Configure Pdns to run as slave for the "no" domain using bind backend:
% diff pdns.conf-dist pdns.conf
10a11
> allow-axfr-ips=127.0.0.0/8, 193.69.2.2
20a22
> allow-notify-from=193.75.4.22
250a253,255
> launch=bind
> bind-config=/usr/local/etc/pdns/named.conf
> bind-dnssec-db=/usr/local/etc/pdns/data/bind-dnssec-db.sqlite3
260a266
> local-address=127.0.0.1,193.75.4.59
270a277
> local-ipv6=::1
495a503
> slave=yes
625a634
> xfr-max-received-mbytes=0

4. Add BIND config file: 
% cat /usr/local/etc/pdns/named.conf
options {
        directory "/usr/local/etc/pdns";
};

zone "no" { type slave; masters { 193.75.4.22; }; file "slave/no"; };

5. Start Pdns.

6. Observe memory usage regularly with the "ps" command.
### Expected behaviour

Large increase in memory usage the first time the "no" zone is transferred. Only very small increases thereafter, as new "no" subdomains are added.
### Actual behaviour

Dramatic increase in memory usage, e.g. "VSZ RSS" numbers:
Aug 5 10:10:15 7996788 7690468
[...]

PR #8351, opened 2019-09-27
auth statbag: move to std::mutex, avoid copies

Habbie:
### Short description
During the debugging of #8161, we discovered more moving around of mutexes. This fixes the other place we did that.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/statbag.cc
+++ b/pdns/statbag.cc
@@ -167,6 +167,16 @@ StatRing<T,Comp>::StatRing(unsigned int size)
 }
 
 template<typename T, typename Comp>
+StatRing<T,Comp>::StatRing(const StatRing<T,Comp> &arg)
+{
+  std::lock_guard<std::mutex> thislock(d_lock);
+  std::lock_guard<std::mutex> arglock(arg.d_lock);
+  
+  d_items = arg.d_items;
+  d_help = arg.d_help;
+}
+
+template<typename T, typename Comp>
 void StatRing<T,Comp>::account(const T& t)
 {
   std::lock_guard<std::mutex> l(d_lock);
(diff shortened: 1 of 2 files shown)


rgacogne:
Doesn't compile in Travis, unfortunately :-/

omoerbeek:
Travis uses gcc 4.8 and clang 5.0.0 both fail here:
```
In file included from statbag.cc:25:
In file included from ./utility.hh:45:
In file included from /usr/lib/gcc/x86_64-linux-gnu/4.8/../../../../include/c++/4.8/string:40:
In file included from /usr/lib/gcc/x86_64-linux-gnu/4.8/../../../../include/c++/4.8/bits/char_traits.h:39:
In file included from /usr/lib/gcc/x86_64-linux-gnu/4.8/../../../../include/c++/4.8/bits/stl_algobase.h:64:
/usr/lib/gcc/x86_64-linux-gnu/4.8/../../../../include/c++/4.8/bits/stl_pair.h:113:21: error: 
      call to implicitly-deleted copy constructor of 'StatRing<std::string,
      CIStringCompare>'
      : first(__a), second(__b) { }
```

I guess we still need to support 4.8. Then  we need to find a solution by either not copying the mutex or maybe move to a static one? The latter might cause too much contention though.

I'll try to find a way.

rgacogne:
I'm surprised that constructing the new element in place invokes the copy constructor? We could declare a copy constructor for `StatRing` that locks the mutexes, copy `d_items` and `d_help`, if needed.

omoerbeek:
Above addition makes Travis happy and isn't too ugly imo.

PR #8365, opened 2019-10-01
dnsdist: Lowercase custom DoH header names

### LINKED ISSUE #8353: [dnsdist] Custom response headers must be in lower case
- Program: dnsdist 
 - Issue type: Feature request

### Short description
If you put custom headers in upper-case or mixed case, HTTP/2 clients fail. This does not seem documented.

### Usecase
Trying to use customResponseHeaders in the server but the problem is probably the same with the extra headers of the internal webserver.

### Description
addDOHLocal("[::]:443", "/etc/dnsdist/server-doh.pem", "/etc/dnsdist/server-doh.key", "/", {customResponseHeaders={["Link"]="<https://www.bortzmeyer.org/doh-bortzmeyer-fr-policy.html> rel=\"\service-meta\"; type=\"text/html\""}})

This will be accepted by dnsdist, no warning sent, but it will break HTTP/2 clients (curl reports "'HTTP/2 stream 0 was not closed cleanly: PROTOCOL_ERROR (err 1)'")

Headers must be in lower case.

Two solutions: document this limitation or, better, lowercase headers automatically

### Related

This bug in the ACME client dehydrated is another case of programmers forgetting that headers must be in lower case. https://github.com/lukas2511/dehydrated/issues/559

rgacogne:
### Short description

Fixes #8353.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdist-lua-bindings.cc
+++ b/pdns/dnsdist-lua-bindings.cc
@@ -374,7 +374,7 @@ void setupLuaBindings(bool client)
     if (customHeaders) {
       headers = std::vector<std::pair<std::string, std::string>>();
       for (const auto& header : *customHeaders) {
-        headers->push_back({ header.first, header.second });
+        headers->push_back({ boost::to_lower_copy(header.first), header.second });
       }
     }
     return std::make_shared<DOHResponseMapEntry>(regex, status, content, headers);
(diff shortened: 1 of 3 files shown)


rgacogne:
This should also be done for `newDOHResponseMapEntry()`.

rgacogne:
> This should also be done for newDOHResponseMapEntry().

Done.

PR #8367, opened 2019-10-01
Implement RFC 8020 "NXDOMAIN: There Really Is Nothing Underneath"

pieterlexis:
### Short description
This commit implements the "NXDOMAIN: There Really Is Nothing Underneath".
When enabled (the default), the SyncRes will check the negative cache if
there exists a higher denied name and uses that data to send an NXDOMAIN
to the client. In essence, it is a more aggressive version of
root-nx-trust (which could be removed in the future).

There are several advantages:

 * We potentially send fewer queries to the internet
 * The record cache is not "polluted" with useless NXDOMAINs

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/pdns/pdns_recursor.cc
+++ b/pdns/pdns_recursor.cc
@@ -3918,6 +3918,7 @@ static int serviceMain(int argc, char*argv[])
   SyncRes::s_ecscachelimitttl = ::arg().asNum("ecs-cache-limit-ttl");
 
   SyncRes::s_qnameminimization = ::arg().mustDo("qname-minimization");
+  SyncRes::s_hardenNXD = ::arg().mustDo("nothing-below-nxdomain");
 
   if (!::arg().isEmpty("ecs-scope-zero-address")) {
     ComboAddress scopeZero(::arg()["ecs-scope-zero-address"]);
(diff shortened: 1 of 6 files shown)


pieterlexis:
I will also add a metric for this

pdns/recursordist/test-syncres_cc5.cc
```diff
@@ -1176,15 +1176,15 @@ BOOST_AUTO_TEST_CASE(test_dnssec_validation_nsec3_wildcard) {
   BOOST_CHECK_EQUAL(res, RCode::NoError);
   BOOST_CHECK_EQUAL(sr->getValidationState(), Secure);
   BOOST_REQUIRE_EQUAL(ret.size(), 6);
-  BOOST_CHECK_EQUAL(queriesCount, 10);
+  BOOST_CHECK_EQUAL(queriesCount, 9);
```
rgacogne: Did you check why the number of queries change? I'm pretty sure this is not normal, and at first glance it is caused by a `NODATA` for `sub.powerdns.com|DS` somehow being leading to `www.sub.powerdns.com: Entire name 'sub.powerdns.com' and below,is negatively cached via 'powerdns.com' for another 300 seconds`.

PR #8370, opened 2019-10-02
auth [WIP] An attempt at zero-copying of packet data in auth

omoerbeek:
### Short description

This moves the 
```
allocate p
loop {
  receive p
  new q
  copy p into q
  process q
  delete q
}
```
into:
```
loop {
  new p
  receive p
  process p
  delete p
}
```
So far this is only an experiment, although unit tests and regression tests succeed.

After feedback, the loop is now:

``` 
p = nullptr
loop {
  new p if it is null or moved to distributor
  receive p
  process p: move to distributor only if needed
}
```
### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/common_startup.cc
+++ b/pdns/common_startup.cc
@@ -415,8 +415,12 @@ try
     NS = N;
   }
 
+  unique_ptr<DNSPacket> question;
   for(;;) {
-    unique_ptr<DNSPacket> question = make_unique<DNSPacket>(true);
+    // On first iteration and whenever the object was moved to the distributor, reallocate
+    if (!question) {
+      question = make_unique<DNSPacket>(true);
+    }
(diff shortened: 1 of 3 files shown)


rgacogne:
It does improve the cache-miss path, but please note that it adds an allocation to the cache-hit fast path.

omoerbeek:
> It does improve the cache-miss path, but please note that it adds an allocation to the cache-hit fast path.

Indeed. I have to think on how to avoid that.

pieterlexis:
@omoerbeek is this still relevant after #8714? If not, please close

omoerbeek:
Yes, I still think this could be relevant.

omoerbeek:
I'm closing this. The idea is might still be relevant, but it needs to be redone as everything got moved around.

PR #8372, opened 2019-10-02
dnsdist: Use SO_BINDTODEVICE when available for newServer's source itf

rgacogne:
### Checklist

I have:
- [x] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdist-lua.cc
+++ b/pdns/dnsdist-lua.cc
@@ -195,6 +195,7 @@ void setupLuaConfig(bool client)
       }
 
       ComboAddress sourceAddr;
+      std::string sourceItfName;
       unsigned int sourceItf = 0;
       size_t numberOfSockets = 1;
       std::set<int> cpus;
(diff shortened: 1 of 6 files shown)


pdns/dnsdist-lua.cc
```diff
@@ -237,10 +238,14 @@ void setupLuaConfig(bool client)
 			        sourceAddr = ComboAddress(source.substr(0, pos));
 			        sourceItf = itfIdx;
 			      }
+#ifdef SO_BINDTODEVICE
+                              /* we need to retain CAP_NET_RAW to be able to set SO_BINDTODEVICE in the health checks */
+                              g_capabilitiesToRetain.insert("CAP_NET_RAW");
```
omoerbeek: Indenting wrong or a github render error?

pdns/dnsdist-tcp.cc
```diff
@@ -79,6 +79,12 @@ static std::unique_ptr<Socket> setupTCPDownstream(shared_ptr<DownstreamState>& d
         if (ds->ipBindAddrNoPort) {
           SSetsockopt(result->getHandle(), SOL_IP, IP_BIND_ADDRESS_NO_PORT, 1);
         }
+#endif
+#ifdef SO_BINDTODEVICE
+        int res = setsockopt(result->getHandle(), SOL_SOCKET, SO_BINDTODEVICE, ds->sourceItfName.c_str(), ds->sourceItfName.length());
```
omoerbeek: Small nit: I had to check what happens if the name is empty. You might consider doing an explicit `if (!ds->sourceItfName.empty())`...

rgacogne: Oh yes, indeed!

pdns/dnsdist-lua.cc
rgacogne: I'll check but this function has a mix of indenting styles so it's always a pain. The upcoming reformat will be very welcome here!

PR #8377, opened 2019-10-03
dnsdist: Compile with -Wall -Wextra -Wshadow by default

rgacogne:
### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1097,8 +1097,8 @@ jobs:
       - run:
           name: configure
           command: |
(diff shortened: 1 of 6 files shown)


rgacogne:
> A bit curious about the (already existing) -O3 though. In OpenBSD, we patch it out in all ports, since we do not trust it to do the right thing. But maybe we are too conservative...

I remember we did some testing before the 1.0.0 release and -O3 was giving us the best results. I think it should be safe to use as long as the code doesn't do anything crazy but I have been bitten in the past.

PR #8385, opened 2019-10-07
More strict flags

omoerbeek:
### Short description

Specifically `-Wextra -Wshadow`. using clang, this created a tonne of signed vs unisgned warnings (mostly in test-* files.).
Some cases of shadowing are also fixed. While reviewing, please pay extra attention to these, it's easy to make a mistake there.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/modules/godbcbackend/sodbc.cc
+++ b/modules/godbcbackend/sodbc.cc
@@ -344,13 +344,13 @@ SSqlStatement* SODBCStatement::nextRow(row_t& row)
     d_result = SQLFetch(d_statement);
     // cerr<<"subsequent SQLFetch done, d_result="<<d_result<<endl;
     if(d_result == SQL_NO_DATA) {
-      SQLRETURN result = SQLMoreResults(d_statement);
-      // cerr<<"SQLMoreResults done, result="<<d_result<<endl;
-      if (result == SQL_NO_DATA) {
-        d_result = result;
+      SQLRETURN result2 = SQLMoreResults(d_statement);
+      // cerr<<"SQLMoreResults done, result="<<d_result2<<endl;
+      if (result2 == SQL_NO_DATA) {
+        d_result = result2;
       }
       else {
-        testResult( result, SQL_HANDLE_STMT, d_statement, "Could not fetch next result set for ("+d_query+").");
+        testResult( result2, SQL_HANDLE_STMT, d_statement, "Could not fetch next result set for ("+d_query+").");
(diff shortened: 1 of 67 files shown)


modules/godbcbackend/sodbc.cc
```diff
@@ -344,13 +344,13 @@ SSqlStatement* SODBCStatement::nextRow(row_t& row)
     d_result = SQLFetch(d_statement);
     // cerr<<"subsequent SQLFetch done, d_result="<<d_result<<endl;
     if(d_result == SQL_NO_DATA) {
-      SQLRETURN result = SQLMoreResults(d_statement);
-      // cerr<<"SQLMoreResults done, result="<<d_result<<endl;
-      if (result == SQL_NO_DATA) {
-        d_result = result;
+      SQLRETURN result2 = SQLMoreResults(d_statement);
+      // cerr<<"SQLMoreResults done, result="<<d_result2<<endl;
```
rgacogne: Should either be `d_result` or `result2` (likely this one). Or perhaps we should remove this line :)

modules/luabackend/lua_functions.cc
```diff
@@ -230,148 +230,148 @@ void register_lua_functions(lua_State *lua) {
     lua_setglobal(lua, "QTypes");
 }
 
-bool LUABackend::getValueFromTable(lua_State *lua, const std::string& key, string& value) {
-  lua_pushstring(lua, key.c_str());
-  lua_gettable(lua, -2);
+bool LUABackend::getValueFromTable(lua_State *lua_state, const std::string& key, string& value) {
```
rgacogne: I'm missing something here, why aren't you also updating the name of the `lua_State` in the functions above?

omoerbeek: clang did not complain about them.... I guess the difference is function vs method. But it wold be better to do them all indeed.

omoerbeek:
I'll try to find the reason why I do get complaints in my environment (clang 8.0.1/boost 1.66 on OpenBSD, which does have a few local changes to clang).

modules/luabackend/lua_functions.cc
omoerbeek: To make it more clear: `lua` is a field in `LUABackend` so for global functions there is no clash.

omoerbeek:
Above addition makes Travis happy and isn't too ugly imo.

PR #8391, opened 2019-10-08
rec: Allow multiple simultaneous incoming TCP queries over a connection

### LINKED ISSUE #8358: Out-of-order processing
Recursor,( dnsdist )
 - Issue type: Feature request

### Short description
Send replies in a different order than the order of the incoming tcp requests.

omoerbeek:
Fixes #8358 

### Short description

Answers are sent out the moment the become available, so not necessarily in the
same order as received. There's a limit on how many queries per TCP
induced connection we may have in flight.

Only very lightly tested!

Includes an sdig hack to read multiple queries from a file.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/pdns_recursor.cc
+++ b/pdns/pdns_recursor.cc
@@ -1780,11 +1780,19 @@ static void startDoResolve(void *p)
           Utility::gettimeofday(&g_now, 0); // needs to be updated
           struct timeval ttd = g_now;
           if (dc->d_tcpConnection->d_requestsInFlight == TCPConnection::s_maxInFlight - 1) {
+            // A read error might have happened. If we add the fd back, it will most likely error again.
+            // This is not a big issue, the next handleTCPClientReadable() will see another read error
+            // and take action.
             //cerr << "Reenabling " << dc->d_socket << ' ' << dc->d_tcpConnection->d_requestsInFlight << endl;
             ttd.tv_sec += g_tcpTimeout;
             t_fdm->addReadFD(dc->d_socket, handleRunningTCPQuestion, dc->d_tcpConnection, &ttd);
           } else {
-            t_fdm->setReadTTD(dc->d_socket, ttd, g_tcpTimeout);
+            // fd might have been removed by read error code, so expect an exception
+            try {
+              t_fdm->setReadTTD(dc->d_socket, ttd, g_tcpTimeout);
+            }
+            catch (FDMultiplexerException &) {
+            }
           }
         }
       }
--- a/pdns/pdns_recursor.cc
+++ b/pdns/pdns_recursor.cc
@@ -746,7 +746,7 @@ static void writePid(void)
   }
 }
 
-TCPConnection::TCPConnection(int fd, const ComboAddress& addr) : data(2, 0), d_remote(addr), d_requestsInFlight(0), d_fd(fd)
+TCPConnection::TCPConnection(int fd, const ComboAddress& addr) : data(2, 0), d_remote(addr), d_fd(fd)
 {
   ++s_currentConnections;
   (*t_tcpClientCounts)[d_remote]++;
@@ -767,7 +767,7 @@ TCPConnection::~TCPConnection()
   --s_currentConnections;
 }
 
-int TCPConnection::s_maxInFlight = 10;
+uint16_t TCPConnection::s_maxInFlight = 10;
 
 AtomicCounter TCPConnection::s_currentConnections;
 
@@ -1750,27 +1750,41 @@ static void startDoResolve(void *p)
       else
         hadError=false;
 
-      // update tcp connection status, either by closing or moving to 'BYTE0'
+      // update tcp connection status, closing if needed and doing the fd multiplexer accounting
 
+      dc->d_tcpConnection->d_requestsInFlight--;
+
+      // In the code below, we try to remove the fd from the set, but
+      // we don't know if another mthread already did the remove, so we can get a
+      // "Tried to remove unlisted fd" exception.  Not that an inflight < limit test
+      // will not work since we do not know if the other mthread got an error or not.
       if(hadError) {
-        // no need to remove us from FDM, we weren't there
+        try {
+          t_fdm->removeReadFD(dc->d_socket);
+        }
+        catch (FDMultiplexerException &) {
+        }
         dc->d_socket = -1;
       }
       else {
         dc->d_tcpConnection->queriesCount++;
         if (g_tcpMaxQueriesPerConn && dc->d_tcpConnection->queriesCount >= g_tcpMaxQueriesPerConn) {
+          try {
+            t_fdm->removeReadFD(dc->d_socket);
+          }
+          catch (FDMultiplexerException &) {
+          }
           dc->d_socket = -1;
         }
         else {
           Utility::gettimeofday(&g_now, 0); // needs to be updated
           struct timeval ttd = g_now;
-          dc->d_tcpConnection->d_requestsInFlight--;
           if (dc->d_tcpConnection->d_requestsInFlight == TCPConnection::s_maxInFlight - 1) {
-            //cerr << "Reenabling... " << dc->d_tcpConnection->d_requestsInFlight << ' ' << dc->d_socket << endl;
+            //cerr << "Reenabling " << dc->d_socket << ' ' << dc->d_tcpConnection->d_requestsInFlight << endl;
             ttd.tv_sec += g_tcpTimeout;
             t_fdm->addReadFD(dc->d_socket, handleRunningTCPQuestion, dc->d_tcpConnection, &ttd);
           } else {
-            t_fdm->setReadTTD(dc->d_socket, ttd,  g_tcpTimeout);
+            t_fdm->setReadTTD(dc->d_socket, ttd, g_tcpTimeout);
           }
         }
       }
@@ -2010,8 +2024,12 @@ static void handleRunningTCPQuestion(int fd, FDMultiplexer::funcparam_t& var)
     if(conn->bytesread==conn->qlen) {
       conn->d_requestsInFlight++;
       if (conn->d_requestsInFlight >= TCPConnection::s_maxInFlight) {
-        //cerr << "Disabling... " << conn->d_requestsInFlight << ' ' << fd << endl;
+        //cerr << "Disabling " << fd << ' ' << conn->d_requestsInFlight << endl;
         t_fdm->removeReadFD(fd); // should no longer awake ourselves when there is data to read
+      } else {
+        Utility::gettimeofday(&g_now, 0); // needed?
+        struct timeval ttd = g_now;
+        t_fdm->setReadTTD(fd, ttd, g_tcpTimeout);
       }
 
       std::unique_ptr<DNSComboWriter> dc;
--- a/pdns/pdns_recursor.cc
+++ b/pdns/pdns_recursor.cc
@@ -746,7 +746,7 @@ static void writePid(void)
   }
 }
 
-TCPConnection::TCPConnection(int fd, const ComboAddress& addr) : data(2, 0), d_remote(addr), d_fd(fd)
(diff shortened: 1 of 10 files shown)


pdns/syncres.hh
```diff
@@ -1012,8 +1012,10 @@ public:
   enum stateenum {BYTE0, BYTE1, GETQUESTION, DONE} state{BYTE0};
   uint16_t qlen{0};
   uint16_t bytesread{0};
-
+  std::atomic<int> d_requestsInFlight;
```
rgacogne: Does it really need to be atomic? I don't think we have any concurrency issue?

rgacogne: Also just a nit but I'd prefer an unsigned type of known size :-)

pdns/pdns_recursor.cc
```diff
@@ -2002,7 +2008,11 @@ static void handleRunningTCPQuestion(int fd, FDMultiplexer::funcparam_t& var)
     }
     conn->bytesread+=(uint16_t)bytes;
     if(conn->bytesread==conn->qlen) {
-      t_fdm->removeReadFD(fd); // should no longer awake ourselves when there is data to read
+      conn->d_requestsInFlight++;
+      if (conn->d_requestsInFlight >= TCPConnection::s_maxInFlight) {
+        //cerr << "Disabling... " << conn->d_requestsInFlight << ' ' << fd << endl;
+        t_fdm->removeReadFD(fd); // should no longer awake ourselves when there is data to read
+      }
```
rgacogne: Shouldn't we update the read TTD when we don't disable the polling?

pdns/pdns_recursor.cc
```diff
@@ -1760,12 +1762,16 @@ static void startDoResolve(void *p)
...
-
-          t_fdm->addReadFD(dc->d_socket, handleRunningTCPQuestion, dc->d_tcpConnection, &ttd);
+          dc->d_tcpConnection->d_requestsInFlight--;
+          if (dc->d_tcpConnection->d_requestsInFlight == TCPConnection::s_maxInFlight - 1) {
+            //cerr << "Reenabling... " << dc->d_tcpConnection->d_requestsInFlight << ' ' << dc->d_socket << endl;
+            ttd.tv_sec += g_tcpTimeout;
+            t_fdm->addReadFD(dc->d_socket, handleRunningTCPQuestion, dc->d_tcpConnection, &ttd);
+          } else {
+            t_fdm->setReadTTD(dc->d_socket, ttd,  g_tcpTimeout);
+          }
```
rgacogne: It looks like we disable the polling on a timeout even if we have queries in-flight (good), but we could resume it later if `dc->d_tcpConnection->d_requestsInFlight == TCPConnection::s_maxInFlight - 1` which might not be what we want?

pdns/pdns_recursor.cc
```diff
@@ -1760,12 +1762,16 @@ static void startDoResolve(void *p)
...
-          ttd.tv_sec += g_tcpTimeout;
-
-          t_fdm->addReadFD(dc->d_socket, handleRunningTCPQuestion, dc->d_tcpConnection, &ttd);
+          dc->d_tcpConnection->d_requestsInFlight--;
+          if (dc->d_tcpConnection->d_requestsInFlight == TCPConnection::s_maxInFlight - 1) {
+            //cerr << "Reenabling... " << dc->d_tcpConnection->d_requestsInFlight << ' ' << dc->d_socket << endl;
+            ttd.tv_sec += g_tcpTimeout;
+            t_fdm->addReadFD(dc->d_socket, handleRunningTCPQuestion, dc->d_tcpConnection, &ttd);
+          } else {
+            t_fdm->setReadTTD(dc->d_socket, ttd,  g_tcpTimeout);
```
rgacogne: Would that fail if the polling was disabled because of a timeout but we still had at least one query in-flight?

pdns/syncres.hh
omoerbeek: I wasn't sure if tcpconnection was read from multiple threads. Will double check.

pdns/pdns_recursor.cc
omoerbeek: Will look into these timeout related issues tomorrow.

pdns/syncres.hh
omoerbeek: Indeed, TCPConnections are only shared by mthreads

pdns/pdns_recursor.cc
omoerbeek: yes, I think so. The original code didn't do that, but it should have.

pdns/pdns_recursor.cc
omoerbeek: Right in flight count should be decremented always and the fd should be remove if it is in the set

omoerbeek:
I think I addressed the FD accounting. It's a bit more complex than you'd think, because you don't know if another mthread before you already had a write error and removed the fd from the multiplexer.

pdns/pdns_recursor.cc
```diff
@@ -1748,24 +1750,50 @@ static void startDoResolve(void *p)
...
-          dc->d_tcpConnection->state=TCPConnection::BYTE0;
           Utility::gettimeofday(&g_now, 0); // needs to be updated
           struct timeval ttd = g_now;
-          ttd.tv_sec += g_tcpTimeout;
-
-          t_fdm->addReadFD(dc->d_socket, handleRunningTCPQuestion, dc->d_tcpConnection, &ttd);
+          if (dc->d_tcpConnection->d_requestsInFlight == TCPConnection::s_maxInFlight - 1) {
+            // A read error might have happened. If we add the fd back, it will most likely error again.
+            // This is not a big issue, the next handleTCPClientReadable() will see another read error
+            // and take action.
```
rgacogne: One possible issue here is that we might have decided to stop reading because the idle timeout was reached, and in that case we will start reading again. I wonder if that might open us to some sort of slow loris attack where a client can keep the TCP connection alive for a very long time without reaching the maximum number of queries on a single TCP connection.

omoerbeek:
The code in the `GETQUESTION` block case needs some fixes, since there are multiple exit paths. In those case the state isn't updated and the `d_requestsInFlight` counter will be wrong.

Additionally, a regression test would be nice.

As for the slow loris attack, I'll try to see if this is possible and/or how to defend against it.

omoerbeek:
As for the slow loris attack: the FD is only added back after we processed a query. If we want to come back in the same state again, we can only do that by processing another query, AFAIKS. So the query count does get increased.

omoerbeek:
Some remaining questions to be answered:
- do we want the sdig part (it includes some refactoring that might be nice)
- what should be the default for `max-concurrent-requests-per-tcp-connection`? Currently 10, but we might be conservative and make it 1.

regression-tests.recursor-dnssec/test_SimpleTCP.py
```diff
@@ -0,0 +1,123 @@
...
+        expectedCNAME = dns.rrset.from_text('cname-to-islandofsecurity.secure.example.', 0, 'IN', 'CNAME', 'node1.islandofsecurity.example.')
+        expectedA = dns.rrset.from_text('node1.islandofsecurity.example.', 0, 'IN', 'A', '192.0.2.20')
+
+        res = self.sendTCPQuery(query)
+
+        self.assertRcodeEqual(res, dns.rcode.NOERROR)
+        self.assertRRsetInAnswer(res, expectedA)
+
+
+    def testVeryBasicOOO(self):
```
rgacogne: Would that test fail without OOO support? It looks like it would work as long as we support pipelining.

pdns/syncres.hh
```diff
@@ -1012,7 +1012,9 @@ public:
   enum stateenum {BYTE0, BYTE1, GETQUESTION, DONE} state{BYTE0};
   uint16_t qlen{0};
   uint16_t bytesread{0};
-
+  uint16_t d_requestsInFlight{0}; // number of mthreads spawned for this connection
+  // The max number of concurrent TCP requests we're willing to process
+  uint16_t d_maxInFlight;
```
rgacogne: I'm wondering if that could not be a static member, only updated once instead of reading `max-concurrent-requests-per-tcp-connection` for each incoming TCP connection?

omoerbeek:
Actually, looking back, what I implemented is pipelining with the additionall bonus of ooo processing.

For unbound it was implemented in Feb 2019, so pretty recent.  For knot and bind resolver IO could not find details, except that they do support pipelining/ooo according to https://dnsprivacy.org/wiki/display/DP/DNS+Privacy+Implementation+Status 

I'll look into making the test a bit more extensive and the static thing.

pdns/pdns_recursor.cc
```diff
@@ -1748,24 +1749,49 @@ static void startDoResolve(void *p)
       else
         hadError=false;
 
-      // update tcp connection status, either by closing or moving to 'BYTE0'
+      // update tcp connection status, closing if needed and doing the fd multiplexer accounting
 
+      dc->d_tcpConnection->d_requestsInFlight--;
```
cmouse: could be a good idea to make sure d_requestsInFlight>0 here...

omoerbeek:
As for the default value, I think I'll leave it at 10 to see if alpha tester catch anything.

pdns/pdns_recursor.cc
```diff
@@ -1778,7 +1780,7 @@ static void startDoResolve(void *p)
         else {
           Utility::gettimeofday(&g_now, 0); // needs to be updated
           struct timeval ttd = g_now;
-          if (dc->d_tcpConnection->d_requestsInFlight == dc->d_tcpConnection->d_maxInFlight - 1) {
+          if (dc->d_tcpConnection->d_requestsInFlight == TCPConnection::s_maxInFlight - 1) {
```
cmouse: this would be better formed as `dc->d_tcpConnection->d_requestsInFlight + 1 == TCPConnection::s_maxInFlight` given it's unsigned int.

regression-tests.recursor-dnssec/test_SimpleTCP.py
omoerbeek: I added a note in the test file I'm not sure yet how to test the OOO code actually.

pdns/pdns_recursor.cc
omoerbeek: `TCPConnection::s_maxInFlight` should be `>= 1` (though we don't check for that atm).

omoerbeek:
BTW tests after latest commit will probably fail because it needs the lua sockets lib.

Habbie:
> BTW tests after latest commit will probably fail because it needs the lua sockets lib.

Fix for that being tested in https://github.com/Habbie/pdns/tree/rec-out-of-order-luasocket - if that goes green on circle and travis, you can steal that commit.

Habbie:
> Fix for that being tested in https://github.com/Habbie/pdns/tree/rec-out-of-order-luasocket - if that goes green on circle and travis, you can steal that commit.

This went green (except for Travis auth which is unrelated) so I pushed the commit here.

pdns/pdns_recursor.cc
```diff
@@ -4014,6 +4014,13 @@ static int serviceMain(int argc, char*argv[])
   g_numThreads = g_numDistributorThreads + g_numWorkerThreads;
   g_maxMThreads = ::arg().asNum("max-mthreads");
 
+  
+  if (TCPConnection::s_maxInFlight < 1 || TCPConnection::s_maxInFlight > USHRT_MAX || TCPConnection::s_maxInFlight >= g_maxMThreads) {
+    g_log<<Logger::Warning<<"Asked to run with illegal max-concurrent-requests-per-tcp-connection, setting to default"<<endl;
```
pieterlexis: You might want to log the default as well :)

omoerbeek: Oh, how I wish the default could be accessed by code ....

omoerbeek: Actually, the logic is wrong. I need to validate before assigning to a ushort.

PR #8396, opened 2019-10-09
Do not use variable printf format strings

omoerbeek:
### Short description

Using a variable format string opens up all kinds of cans of worms.
So use a fixed string and enumerate the remaining (after using `*` for  parametrised field width) 4 possibilities.
While there, add a unit test for the $TEMPLATE case.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code


pdns/zoneparser-tng.cc
```diff
@@ -193,11 +193,22 @@ bool ZoneParserTNG::getTemplateLine()
...
+          break;
+        case 'x':
+          snprintf(tmp, sizeof(tmp), "%0*x", width, d_templatecounter + offset);
+          break;
+        case 'X':
+          snprintf(tmp, sizeof(tmp), "%0*X", width, d_templatecounter + offset);
+          break;
+        case 'd':
+        default:
+          snprintf(tmp, sizeof(tmp), "%0*d", width, d_templatecounter + offset);
```
rgacogne: I see a 'n' and 'N' types in the documentation for "$GENERATE" ( http://www.zytrax.com/books/dns/ch8/generate.html ),  do we support that?

omoerbeek: Should we? If so, I can do it.

omoerbeek: See #8405

PR #8408, opened 2019-10-11
dnsdist: Fix the caching of large entries

rgacogne:
### Short description

It turns out that the packet cache suffered a regression since 1.3.0: the buffer allocated to incoming queries, and used to store the cached response if any, was not big enough to allow large (>1500 bytes for UDP, > 512 bytes for TCP/DoT/DoH) answers to be retrieved. We can now always retrieve entries up to 4096 bytes.
We also cached entries larger than what we could actually retrieve, we now no longer cache entries larger than 4096 bytes. 

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/pdns/dnsdistdist/doh.cc
+++ b/pdns/dnsdistdist/doh.cc
@@ -304,9 +304,9 @@ static int processDOHQuery(DOHUnit* du)
     struct timespec queryRealTime;
     gettime(&queryRealTime, true);
     uint16_t len = du->query.length();
-    /* allocate a bit more memory to be able to spoof the content,
-       or to add ECS without allocating a new buffer */
-    du->query.resize(du->query.size() + 512);
+    /* We reserve at least 512 additional bytes to be able to add EDNS, but we also want
+       at least s_maxPacketCacheEntrySize bytes to be able to spoof the content or fill the answer from the packet cache */
+    du->query.resize((du->query.size() + 512 < s_maxPacketCacheEntrySize) ? s_maxPacketCacheEntrySize : (du->query.size() + 512));
     size_t bufferSize = du->query.size();
     auto query = const_cast<char*>(du->query.c_str());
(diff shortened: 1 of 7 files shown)


pdns/dnsdistdist/doh.cc
```diff
@@ -304,9 +304,9 @@ static int processDOHQuery(DOHUnit* du)
     struct timespec queryRealTime;
     gettime(&queryRealTime, true);
     uint16_t len = du->query.length();
-    /* allocate a bit more memory to be able to spoof the content,
-       or to add ECS without allocating a new buffer */
-    du->query.resize(du->query.size() + 512);
+    /* We reserve at least 512 additional bytes to be able to add EDNS, but we also want
+       at least s_maxPacketCacheEntrySize bytes to be able to spoof the content or fill the answer from the packet cache */
+    du->query.resize((du->query.size() + 512 < s_maxPacketCacheEntrySize) ? s_maxPacketCacheEntrySize : (du->query.size() + 512));
```
omoerbeek: Slight preference for std::min() here

pdns/dnsdistdist/doh.cc
```diff
@@ -304,9 +304,9 @@ static int processDOHQuery(DOHUnit* du)
     struct timespec queryRealTime;
     gettime(&queryRealTime, true);
     uint16_t len = du->query.length();
-    /* allocate a bit more memory to be able to spoof the content,
-       or to add ECS without allocating a new buffer */
-    du->query.resize(du->query.size() + 512);
+    /* We reserve at least 512 additional bytes to be able to add EDNS, but we also want
+       at least s_maxPacketCacheEntrySize bytes to be able to spoof the content or fill the answer from the packet cache */
```
omoerbeek: Slight preference for std::max() here

rgacogne:
Rebased to fix a conflict.

PR #8410, opened 2019-10-11
Fix typo in setQueryRate docs

franklouwers:
### Short description
Fix small typo in the dnsdist DynBlocks guide

### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [x] documented the code

### DIFF
--- a/pdns/dnsdistdist/docs/guides/dynblocks.rst
+++ b/pdns/dnsdistdist/docs/guides/dynblocks.rst
@@ -81,6 +81,6 @@ action is applied.
 .. code-block:: lua
 
   local dbr = dynBlockRulesGroup()
-  -- generate a warning above 100 qps for 10s, and start dropping incoming queries above 300 qps for 10s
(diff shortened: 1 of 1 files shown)


pdns/dnsdistdist/docs/guides/dynblocks.rst
```diff
@@ -81,6 +81,6 @@ action is applied.
 .. code-block:: lua
 
   local dbr = dynBlockRulesGroup()
-  -- generate a warning above 100 qps for 10s, and start dropping incoming queries above 300 qps for 10s
+  -- generate a warning above 100 qps for 10s, and start dropping incoming queries above 300 qps for 60s
```
rgacogne: The text is misleading, we indeed:
- generate a warning if we receive more than 100 qps during 10s ;
- drop queries for the next 60s if we receive more than 300 qps during 10s.

Perhaps it's only me but I'm afraid that the proposed version is as misleading as the existing one?

franklouwers: please check new version

PR #8418, opened 2019-10-15
Deb: Load DNSSEC Keys from disk by default

pieterlexis:
### Short description
This also automatically reloads them each 24 hours by default. A similar method was implemented by @zeha in Debian downstream for recursor 4.1.x.

@zeha you might want to port this to downstream (in 4.2.0+) as well.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/builder-support/debian/recursor/debian-buster/rules
+++ b/builder-support/debian/recursor/debian-buster/rules
@@ -39,10 +39,15 @@ override_dh_auto_configure:
 
 override_dh_auto_install:
 	dh_auto_install
+	install -d debian/pdns-recursor/usr/share/pdns-recursor/lua-config
+	install -m 644 -t debian/pdns-recursor/usr/share/pdns-recursor/lua-config debian/lua-config/rootkeys.lua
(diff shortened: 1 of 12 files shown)


zeha:
I liked the comment in your initial commit, so I have imported that into Debian. See https://salsa.debian.org/dns-team/pdns-recursor/commit/c437c59fa87d18d7e8cfe1a9f4892372e477a40c

builder-support/debian/recursor/debian-buster/rules
```diff
@@ -60,5 +65,5 @@ override_dh_gencontrol:
 
 override_dh_fixperms:
 	dh_fixperms
-        # these files often contain passwords. 640 as it is chowned to root:pdns
+# these files often contain passwords. 640 as it is chowned to root:pdns
```
zeha: It looks like this bit came from the auth packaging, but recursor.conf can never have passwords, right?

omoerbeek: There's the webserver's credentials

PR #8430, opened 2019-10-17
builder: demand user chooses a module except when building docs

Habbie:
### Short description
Without this, running something like `builder/build.sh centos-7` (without any `-m`) will spend a lot of time, and then fail confusingly.

Only lightly tested. PRed because @pieterlexis and I half-solved this problem 3 times before but it never landed.

### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code


Habbie:
After at least a small review, I suggest passing this through buildbot to see if anything broke.

PR #8439, opened 2019-10-19
geoipbackend: Use Netmask instead of string

### LINKED ISSUE #8438: Lua GeoIP legacy bug for IPv6 clients
- Program: Authoritative
 - Issue type: Bug report

### Short description

For IPv6 client addrs the Lua backend is unable to successfully perform legacy GeoIP lookups.
Lookups via the MaxMind GeoIP protocol against MaxMind GeoIP databases works fine.

### Environment

 - Software version: 4.2.0

### Steps to reproduce

1. enable the GeoIP backend
2. enable LUA records
3. configure the geoip-database-files according to your environment
4. Create a LUA record that does a geoip lookup, e.g.:
```
continent.example.com     60      IN      LUA     TXT "; if(continent('AS')) then return 'AS' elseif(continent('EU')) then return 'EU' elseif(continent('NA')) then return 'NA' elseif(continent('OC')) then return 'OC' elseif(continent('SA')) then return 'SA' else return 'unknown' end"
```
5. Query this record from an IPv6 client address

Now pdns will return "unknown", no matter if `geoiplookup6` is able to successfully locate the address or not.

### Expected behaviour

For this example record above pdns should return the correct continent.

### Actual behaviour

Pdns just returns "unknown" due to the inability to have a thruthy lookup against any of the given continents.

### Other information

This has been already briefly debugged and discussed on IRC, where @cmouse has already posted a first shot at a fix, which is still subject to discussion and change:

```
diff --git a/modules/geoipbackend/geoipbackend.cc b/modules/geoipbackend/geoipbackend.cc
index 682501e..02fa2a9 100644
--- a/modules/geoipbackend/geoipbackend.cc
+++ b/modules/geoipbackend/geoipbackend.cc
@@ -504,8 +504,9 @@ string getGeoForLua(const std::string& ip, int qaint)
 {
   GeoIPInterface::GeoIPQueryAttribute qa((GeoIPInterface::GeoIPQueryAttribute)qaint);
   try {
+    const ComboAddress addr{ip};
     GeoIPNetmask gl;
-    string res=queryGeoIP(ip, false, qa, gl);
+    string res=queryGeoIP(ip, addr.isIPv6(), qa, gl);
     //    cout<<"Result for "<<ip<<" lookup: "<<res<<endl;
     if(qa==GeoIPInterface::ASn && boost::starts_with(res, "as"))
       return res.substr(2);
```

cmouse:
### Short description
Consistently use Netmask instead of string and boolean in API. Fixes #8438 

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/modules/geoipbackend/geoipbackend.cc
+++ b/modules/geoipbackend/geoipbackend.cc
@@ -89,7 +89,7 @@ GeoIPBackend::GeoIPBackend(const string& suffix) {
 static vector<std::unique_ptr<GeoIPInterface> > s_geoip_files;
 
 string getGeoForLua(const std::string& ip, int qaint);
-static string queryGeoIP(const string &ip, bool v6, GeoIPInterface::GeoIPQueryAttribute attribute, GeoIPNetmask& gl);
+static string queryGeoIP(const Netmask& addr, GeoIPInterface::GeoIPQueryAttribute attribute, GeoIPNetmask& gl);
 
 void GeoIPBackend::initialize() {
   YAML::Node config;
@@ -311,7 +311,7 @@ GeoIPBackend::~GeoIPBackend() {
   }
 }
 
-bool GeoIPBackend::lookup_static(const GeoIPDomain &dom, const DNSName &search, const QType &qtype, const DNSName& qdomain, const std::string &ip, GeoIPNetmask &gl, bool v6) {
+bool GeoIPBackend::lookup_static(const GeoIPDomain &dom, const DNSName &search, const QType &qtype, const DNSName& qdomain, const Netmask& addr, GeoIPNetmask &gl) {
   const auto& i = dom.records.find(search);
   map<uint16_t,int> cumul_probabilities;
   int probability_rnd = 1+(dns_random(1000)); // setting probability=0 means it never is used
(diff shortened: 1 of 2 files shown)


foxxx0:
just applied this patch against the `4.2.0` release tarball and it resolves #8438 for me, great job!

Habbie:
```diff
diff --git a/modules/geoipbackend/geoipbackend.cc b/modules/geoipbackend/geoipbackend.cc
index 023f760c7..93455e879 100644
--- a/modules/geoipbackend/geoipbackend.cc
+++ b/modules/geoipbackend/geoipbackend.cc
@@ -527,7 +527,7 @@ bool queryGeoLocation(const Netmask& addr, GeoIPNetmask& gl, double& lat, double
     if (addr.isIpv6()) {
       if (gi->queryLocationV6(gl, addr.toString(), lat, lon, alt, prec))
         return true;
-     } else if (gi->queryLocation(gl, addr.toString(), lat, lon, alt, prec))
+     } else if (gi->queryLocation(gl, addr.toStringNoMask(), lat, lon, alt, prec))
         return true;
   }
   return false;
```

this unbreaks the `loc-resolution` test for me. I did not check if the same mistake exists in more places.

Habbie:
>  The same mistake exists at least once more, two lines up.

I've taken the liberty of pushing a commit that fixes these two.

cmouse:
We get Netmask from the DNSPacket class, so to use ComboAddress would require converting that Netmask into ComboAddress, which is extra work.

rgacogne:
OK

rgacogne:
The failure in the recursor's regression tests is unrelated.

PR #8440, opened 2019-10-19
Fix -WShadow warnings

cmouse:
### Short description
Fix lots of shadowing variables

### Checklist

I have:
- [x] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/nsecrecords.cc
+++ b/pdns/nsecrecords.cc
@@ -76,9 +76,9 @@ void NSECBitmap::toPacket(DNSPacketWriter& pw)
 {
   NSECBitmapGenerator nbg(pw);
   if (d_bitset) {
-    size_t count = d_bitset->count();
     size_t found = 0;
-    for(size_t idx = 0; idx < nbTypes && found < count; ++idx){
+    size_t l_count = d_bitset->count();
+    for(size_t idx = 0; idx < nbTypes && found < l_count; ++idx){
       if (!d_bitset->test(idx)) {
         continue;
       }
@@ -138,9 +138,9 @@ string NSECBitmap::getZoneRepresentation() const
   string ret;
 
   if (d_bitset) {
-    size_t count = d_bitset->count();
     size_t found = 0;
-    for(size_t idx = 0; idx < nbTypes && found < count; ++idx) {
+    size_t l_count = d_bitset->count();
+    for(size_t idx = 0; idx < nbTypes && found < l_count; ++idx) {
       if (!d_bitset->test(idx)) {
         continue;
       }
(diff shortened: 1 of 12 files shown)


pdns/opensslsigners.cc
```diff
@@ -346,13 +346,13 @@ int OpenSSLRSADNSCryptoKeyEngine::hashSizeToKind(const size_t hashSize)
...
-  string hash = this->hash(msg);
-  int hashKind = hashSizeToKind(hash.size());
+  string l_hash = this->hash(msg);
+  int hashKind = hashSizeToKind(l_hash.size());
   std::string signature;
   signature.resize(RSA_size(d_key.get()));
   unsigned int signatureLen = 0;
 
-  int res = RSA_sign(hashKind, reinterpret_cast<unsigned char*>(&hash.at(0)), hash.length(), reinterpret_cast<unsigned char*>(&signature.at(0)), &signatureLen, d_key.get());
+  int res = RSA_sign(hashKind, (unsigned char*)l_hash.c_str(), l_hash.length(), (unsigned char*)signature.c_str(), &signatureLen, d_key.get());
```
rgacogne: I'm not a big fan of the C++-style cast to C-style cast change, here..

rgacogne: I'm also not sure why this commit also moves from `&.data(0)` to `c_str()`?

cmouse: because it seemed to be in minority. i think it would be sensible to do either one instead of doing both ways. i'm happy to rectify the others to use data(0) instead.

cmouse: also RSA_sign is C style API.

cmouse: ping?

rgacogne: OK, I don't care much about `data()` vs `c_str()`,although I find the latter a bit misleading for binary strings, but I was surprised to see this change in a PR/commit about shadowed variables.
Regarding the casts, my humble point of view is that using a C-style cast is a regression in code readability, and I'm also not sure I see the relation with `RSA_sign()`?

cmouse: Ok, so unrelated change. I'll revert it.

pdns/nsecrecords.cc
```diff
@@ -76,9 +76,8 @@ void NSECBitmap::toPacket(DNSPacketWriter& pw)
 {
   NSECBitmapGenerator nbg(pw);
   if (d_bitset) {
-    size_t count = d_bitset->count();
     size_t found = 0;
-    for(size_t idx = 0; idx < nbTypes && found < count; ++idx){
+    for(size_t idx = 0; idx < nbTypes && found < d_bitset->count(); ++idx){
```
rgacogne: `std::bitset::count()` is likely O(n) and I'm not 100% confident that the compiler will be able to cache it, since the body of the loop calls functions that, as far as the compiler know, could possibly alter the bitset.

pdns/nsecrecords.cc
```diff
@@ -138,9 +137,8 @@ string NSECBitmap::getZoneRepresentation() const
   string ret;
 
   if (d_bitset) {
-    size_t count = d_bitset->count();
     size_t found = 0;
-    for(size_t idx = 0; idx < nbTypes && found < count; ++idx) {
+    for(size_t idx = 0; idx < nbTypes && found < d_bitset->count(); ++idx) {
```
rgacogne: Same.

PR #8447, opened 2019-10-22
dnsdist: Add metrics about TLS handshake failures for DoH and DoT

rgacogne:
### Short description

So we can detect an abnormal rate of TLS connections failing because the cipher suite or the protocols supported are too restrictive, for example.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code


rgacogne:
Doesn't compile on Travis because `SSL_R_VERSION_TOO_LOW` is not defined there, I'll investigate later.

rgacogne:
> Doesn't compile on Travis because `SSL_R_VERSION_TOO_LOW` is not defined there, I'll investigate later.

Fixed.

Issue #7185, opened 2018-11-13
dnsdist: incorrect check-config behaviour


chbruyand:
- Program: dnsdist 
 - Issue type: Bug report 

### Short description

`check-config` mode actually relies on the same behaviour than `client` mode to run. However, all objects won't be properly created while launched under `check-config` as it will be inhibited by the `client` mode.
We should separate the two modes and ensure proper behaviour.

rgacogne:
Postponing to 1.5.0 since it might break some setups.

PR #8456, opened 2019-10-23
dnsdist: Separate the check-config and client modes

rgacogne:
### Short description

In client (and single command) mode we do not care too much about parsing the syntax of commands that do not relate to the console itself, and we do not want to create sockets and touch files on the local filesystem while parsing the configuration.
In check-config mode, however, we want to parse as much configuration directives as possible, and we to generate some files that might be required to validate the rest of the configuration (DNSCrypt certificates and keys, OCSP responses..).

This PR separates the two modes and tries to skip only the following parts when in check-config mode, compared to 'normal' configuration parsing:
- we still parse and create objects when processing 'add*Local()'   and 'set*Local()' directives, since the actual binding is delayed anyway ;
- we do not connect the outgoing sockets toward our backends ;
- we do not create Protocol Buffer or DNSTap outgoing connections ;
- we do not start the console, SNMP or webserver threads.

Should fix https://github.com/PowerDNS/pdns/issues/7185

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/pdns/dnsdist-lua.cc
+++ b/pdns/dnsdist-lua.cc
@@ -206,15 +206,19 @@ static void parseTLSConfig(TLSConfig& config, const std::string& context, boost:
 
 #endif // defined(HAVE_DNS_OVER_TLS) || defined(HAVE_DNS_OVER_HTTPS)
 
-void setupLuaConfig(bool client)
+void setupLuaConfig(bool client, bool configCheck)
 {
   typedef std::unordered_map<std::string, boost::variant<bool, std::string, vector<pair<int, std::string> >, DownstreamState::checkfunc_t > > newserver_t;
   g_lua.writeFunction("inClientStartup", [client]() {
         return client && !g_configurationDone;
   });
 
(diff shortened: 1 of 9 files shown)


rgacogne:
One regression test fails with:
```
  File "/home/travis/build/rgacogne/pdns/regression-tests.dnsdist/dnsdisttests.py", line 96, in startDNSDist
    raise AssertionError('dnsdist --check-config failed: %s' % output)
AssertionError: dnsdist --check-config failed: Error creating new server: downstream weight value must be greater than 0.
Error creating new server: downstream weight value must be between 1 and 2147483647
Configuration 'configs/dnsdist_TestRoutingBadWeightWRandom.conf' OK!
```
which is actually an improvement but the test must be fixed.

rgacogne:
> but the test must be fixed.

Done.

rgacogne:
The rec failure is unrelated (a rebase would fix it).

rgacogne:
Rebased.

PR #8462, opened 2019-10-24
Fixes #8422 abort transaction and return refused to client when update policy returns false

### LINKED ISSUE #8422: dns lua update policy sends wrong response to client when update is denied
when lua script is used for update policy client gets wrong response when update is denied.
server sends NOERROR instead of REFUSED.

as described in rfc even when multiple updates are requested and one of them is refused. rollback has to be perfomed and REFUSED sent to client.

querplis:
### Short description
Fixes #8422 abort transaction and return refused to client when update policy returns false
### Checklist
I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/rfc2136handler.cc
+++ b/pdns/rfc2136handler.cc
@@ -938,7 +938,8 @@ int PacketHandler::processUpdate(DNSPacket& p) {
         if (this->d_update_policy_lua != NULL) {
           if (this->d_update_policy_lua->updatePolicy(rr->d_name, QType(rr->d_type), di.zone, p) == false) {
             g_log<<Logger::Warning<<msgPrefix<<"Refusing update for " << rr->d_name << "/" << QType(rr->d_type).getName() << ": Not permitted by policy"<<endl;
-            continue;
+            di.backend->abortTransaction();
+            return RCode::Refused;
           } else {
             g_log<<Logger::Debug<<msgPrefix<<"Accepting update for " << rr->d_name << "/" << QType(rr->d_type).getName() << ": Permitted by policy"<<endl;
           }


rgacogne:
This would change the documented behavior of the `updatepolicy` callback which, according to the documentation "is called for each record at a time and you can approve or reject any or all".
I'm not sure it makes sense to accept only a part of a DNS update, to be clear, but I think we need to be certain that we actually want to change the semantics of an existing functionality.

cmouse:
It might be that you don't want to accept IPv6 (or IPv4) updates from hosts. Or you want to avoid accepting anything but A/AAAA updates. I think it might be best to keep track if *any* update succeeded.

querplis:
@rgacogne it is all or nothing at the moment with this patch , and if i did read rfc correctly it is how it shoud be done

Habbie:
This patch breaks the documented behaviour. I'd be happy to take a different patch that adds a third return value that causes a transaction abort.

Habbie (changes requested):
This patch breaks the documented behaviour. I'd be happy to take a different patch that adds a third return value that causes a transaction abort.

Habbie:
Closing this. As I mentioned above, we'd be happy to take a more flexible patch. Thanks!

jdreichmann:
As the current behaviour makes it impossible to know from the Answer if an update actually went through when using a lua script for validation, i would be interested in what exactly is meant with the
> "third return value that causes a transaction abort."

Is the lua scripts return value `true`, `false`, `<to_be_added>` meant or something else?

Habbie:
> Is the lua scripts return value `true`, `false`, `<to_be_added>` meant or something else?

to be added, yes

jdreichmann:
As that return value changes the result of the entire transaction, it seems a bit counterintuitive to have that as a return value of the evaluation of a single record affecting the whole set - wouldn't a "strict" mode, where a single record failing the policy check cancels the transaction, be a bit more intuitive?

i.e. how important is it to stick to a third possible return value in the lua script which then in turn cancels the entire transaction?

Habbie:
Note that we are discussing on a closed ticket, and likely these comments will eventually be forgotten. If you want to discuss a new proposal, please open a new ticket.

jdreichmann:
I'm most interested in gettting the problem described in #8462 fixed and am trying to find out which kind of exact behaviour would be wanted from a patch, as this one sadly got stalled after a review.

PR #8470, opened 2019-10-25
rec: prime NS records of root-servers.net parent (.net)

omoerbeek:
### Short description

Do not only put the root hints into the cache, but also make sure the NS records of the top level domains of the *names* of the root servers are in the cache. We need these to correctly determine the security status of that specific domain (normally `root-servers.net`). This is caused by the accident that the root servers are authoritative for `root-servers.net`, and some implementations reply not with a delegation on a `root-servers.net` `DS` query, but with a `NODATA` response (the domain is unsigned).

Discussed with @Habbie and @pieterlexis 

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [X] documented the code


pdns/reczones.cc
```diff
@@ -88,12 +92,38 @@ void primeHints(void)
         rr.content=toLower(rr.content);
         nsset.push_back(DNSRecord(rr));
       }
+      t_rootNSZones.insert(rr.qname.getLastLabel());
     }
   }
   t_RC->doWipeCache(g_rootdnsname, false, QType::NS);
```
pieterlexis: Should we wipe the cache for the entries in `t_rootNSZones` as well?

PR #8474, opened 2019-10-28
auth: Fix auth logging if no packet cache; from Habbie

omoerbeek:
### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code


mind04:
A single log message mentioning the disabled PC at startup will give you all the information you need. There is no need to repeat it for every query.

PR #8477, opened 2019-10-28
rec: enable qname minimization

omoerbeek:
### Short description

Enablq qname minimization by default.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/regression-tests.recursor-dnssec/test_Protobuf.py
+++ b/regression-tests.recursor-dnssec/test_Protobuf.py
@@ -129,7 +129,7 @@ class TestRecursorProtobuf(RecursorTest):
         self.assertEquals(msg.id, query.id)
         self.assertTrue(msg.HasField('inBytes'))
         if normalQueryResponse:
-            # compare inBytes with length off query/response
+            # compare inBytes with length of query/response
             # Note that for responses, the size we received might differ
             # because dnspython might compress labels differently from
             # the recursor
@@ -158,7 +158,7 @@ class TestRecursorProtobuf(RecursorTest):
         if length is not None:
           self.assertEquals(msg.inBytes, length)
         else:
-          # compare inBytes with length off query/response
+          # compare inBytes with length of query/response
           self.assertEquals(msg.inBytes, len(query.to_wire()))
 
     def checkProtobufQuery(self, msg, protocol, query, qclass, qtype, qname, initiator='127.0.0.1'):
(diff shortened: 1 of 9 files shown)


omoerbeek:
Not yet ready for primetime. e.g. reggresion test failure that need to be investigated.

omoerbeek:
All tests succeed now.

There is one rather fundamental change in this PR,  and that is the handling of forwarding in `syncres.cc`. To be discussed....

regression-tests.recursor-dnssec/test_Protobuf.py
```diff
@@ -129,7 +129,7 @@ def checkProtobufBase(self, msg, protocol, query, initiator, normalQueryResponse
         self.assertEquals(msg.id, query.id)
         self.assertTrue(msg.HasField('inBytes'))
         if normalQueryResponse:
-            # compare inBytes with length of query/response
+            # compare inBytes with length off query/response
```
rgacogne: Wasn't `of` actually the right version?

regression-tests.recursor-dnssec/test_Protobuf.py
```diff
@@ -351,7 +351,10 @@ class OutgoingProtobufDefaultTest(TestRecursorProtobuf):
 
     _confdir = 'OutgoingProtobufDefault'
     _config_template = """
-auth-zones=example=configs/%s/example.zone""" % _confdir
+    # Switch of QName Minimization, it generates much more protobuf messages
```
rgacogne:
```suggestion
    # Switch off QName Minimization, it generates much more protobuf messages
```

regression-tests.recursor-dnssec/test_Protobuf.py
```diff
@@ -380,7 +383,10 @@ class OutgoingProtobufNoQueriesTest(TestRecursorProtobuf):
 
     _confdir = 'OutgoingProtobufNoQueries'
     _config_template = """
-auth-zones=example=configs/%s/example.zone""" % _confdir
+    # Switch of QName Minimization, it generates much more protobuf messages
```
rgacogne:
```suggestion
    # Switch off QName Minimization, it generates much more protobuf messages
```

regression-tests.recursor-dnssec/test_Protobuf.py
Habbie: Indeed

omoerbeek: Yeah, only the other was correct, thanks

PR #8491, opened 2019-10-30
dnsdist: Implement parallel health checks

rgacogne:
### Short description

We used to do health checking sequentially, which could take a long time with a huge list of backends, or simply if several of them were timing out.
This PR implements parallel health checks, we first loop through all the backends we have to check and send them a query, then use multiplexing on the sockets to get the responses, or handle the timeouts if any. This way a single round of checks should only take as long as the greatest timeout of the backends, instead of the total of all timeouts today.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


rgacogne:
Rebased to fix a conflict.

PR #8492, opened 2019-10-30
Add a parameter to limit the number of '$GENERATE' steps

rgacogne:
### Short description

A very large number of steps in one bind-style zone could trigger a lot of CPU and memory usage. This is not really a problem in most of the deployment scenarios but OSS-Fuzz has recently figured out how to trigger it.  

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -927,6 +927,21 @@ will generally suffice for most installations.
 Maximum number of empty non-terminals to add to a zone. This is a
 protection measure to avoid database explosion due to long names.
 
+.. _setting-max-generate-steps:
+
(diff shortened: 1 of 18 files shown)


Habbie:
`comfun` is a tool for doing statistics over large zones, like .com. It should never hit $GENERATE - so either ignore it, or put a very low limit in it so we never get the exploding kind of surprise.

rgacogne:
We should now be enforcing the limit, or completely disabling '$GENERATE' support, everywhere. I had to rebase to fix some conflicts, sorry about that.

PR #8521, opened 2019-11-11
rec: Make threads run until asked to stop.

omoerbeek:
This is safer since the atexit handler is not ran while threads are
still active. Also, when using valgrind we get more clean leak reports.

Retry of the accidentally merged #8518 that was reverted.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/rec_channel.cc
+++ b/pdns/rec_channel.cc
@@ -17,6 +17,8 @@
 
 #include "namespaces.hh"
 
+volatile sig_atomic_t RecursorControlChannel::stop;
+
 RecursorControlChannel::RecursorControlChannel()
 {
   d_fd=-1;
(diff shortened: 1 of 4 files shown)


pdns/rec_channel.cc
```diff
@@ -17,6 +17,8 @@
 
 #include "namespaces.hh"
 
+volatile sig_atomic_t RecursorControlChannel::stop;
```
rgacogne: Is the initial value set somewhere?

omoerbeek: https://en.cppreference.com/w/cpp/language/initialization defines that statics are zero initialized. But it might be better to do it explicitly indeed .

Issue #7771, opened 2019-05-02
Recursor 4.1.12 leaks memory


cmouse:
- Program: Recursor
 - Issue type: Bug report

### Short description
It appears that 4.1.12 might be leaking memory on long-term basis. Node with average of 70 kpq/s and 2k validations/s is leaking memory about 4G per month. 

### Environment
 - Operating system: 14.04 ubuntu LTS
 - Software version: 4.1.12
 - Software source: PowerDNS repository

### Steps to reproduce
1. Run PowerDNS recursor for a month

### Expected behaviour
Memory usage should not be increasing constantly.

### Actual behaviour
Memory usage increases.

![memory usage](https://user-images.githubusercontent.com/1204643/57092962-56d43780-6d15-11e9-835a-59aee9c7d964.png)

As a curiosity, it seems memory leak has been happening over several software versions over the years.

![long term memory usage](https://user-images.githubusercontent.com/1204643/57093033-85eaa900-6d15-11e9-886f-ed2f6c0c972e.png)

### Other information
Cache can be ruled out due to cache usage being fairly constant.

![cache entries](https://user-images.githubusercontent.com/1204643/57093062-94d15b80-6d15-11e9-84d0-4a648a739d3c.png)

smaps breakdown
-------------------
```
===============================================================================
Private    Private     Shared     Shared
 Clean   +  Dirty   +  Clean   +  Dirty   =  Total   : library
===============================================================================
    0 kB + 3451432 kB +     0 kB +     0 kB = 3451432 kB : [anonymous]
 5204 kB +    96 kB +     0 kB +     0 kB =  5300 kB : pdns_recursor
    0 kB +  1996 kB +     0 kB +     0 kB =  1996 kB : [heap]
    0 kB +    24 kB +  1512 kB +     0 kB =  1536 kB : libc-2.19.so
    0 kB +   152 kB +  1300 kB +     0 kB =  1452 kB : libcrypto.so.1.0.0
  848 kB +    40 kB +     0 kB +     0 kB =   888 kB : libstdc++.so.6.0.19
  856 kB +    20 kB +     0 kB +     0 kB =   876 kB : libprotobuf.so.8.0.0
   96 kB +     8 kB +   460 kB +     0 kB =   564 kB : libm-2.19.so
  408 kB +    12 kB +     0 kB +     0 kB =   420 kB : libluajit-5.1.so.2.0.2
[...]

PR #8525, opened 2019-11-12
Rec: Purge map of failed auths periodically by keeping a last changed timestamp.

omoerbeek:
### Short description

SyncRes thread local storage includes a map of failed auths which was
only cleaned if a specific IP was contacted again and that contact
succeeded. Persistent failing auths or auths that are never tried
again remained in the map.

While here add code to dump the failed servers map. Might (partially?)
solve #7771.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/pdns_recursor.cc
+++ b/pdns/pdns_recursor.cc
@@ -2844,7 +2844,8 @@ static void doStats(void)
 
     g_log<<Logger::Notice<<"stats: throttle map: "
       << broadcastAccFunction<uint64_t>(pleaseGetThrottleSize) <<", ns speeds: "
-      << broadcastAccFunction<uint64_t>(pleaseGetNsSpeedsSize)<<endl;
+      << broadcastAccFunction<uint64_t>(pleaseGetNsSpeedsSize)<<", failed ns: "
+      << broadcastAccFunction<uint64_t>(pleaseGetFailedServersSize)<<endl;
     g_log<<Logger::Notice<<"stats: outpacket/query ratio "<<(int)(SyncRes::s_outqueries*100.0/SyncRes::s_queries)<<"%";
     g_log<<Logger::Notice<<", "<<(int)(SyncRes::s_throttledqueries*100.0/(SyncRes::s_outqueries+SyncRes::s_throttledqueries))<<"% throttled, "
      <<SyncRes::s_nodelegated<<" no-delegation drops"<<endl;
@@ -2908,6 +2909,8 @@ static void houseKeeping(void *)
       if(!((cleanCounter++)%40)) {  // this is a full scan!
 	time_t limit=now.tv_sec-300;
         SyncRes::pruneNSSpeeds(limit);
+        limit = now.tv_sec - SyncRes::s_serverdownthrottletime * 10;
+        SyncRes::pruneFailedServers(limit);
       }
       last_prune=time(0);
     }
--- a/pdns/pdns_recursor.cc
+++ b/pdns/pdns_recursor.cc
@@ -2845,7 +2845,8 @@ static void doStats(void)
     g_log<<Logger::Notice<<"stats: throttle map: "
       << broadcastAccFunction<uint64_t>(pleaseGetThrottleSize) <<", ns speeds: "
       << broadcastAccFunction<uint64_t>(pleaseGetNsSpeedsSize)<<", failed ns: "
-      << broadcastAccFunction<uint64_t>(pleaseGetFailedServersSize)<<endl;
+      << broadcastAccFunction<uint64_t>(pleaseGetFailedServersSize)<<", ednsmap: "
+      <<broadcastAccFunction<uint64_t>(pleaseGetEDNSStatusesSize)<<endl;
     g_log<<Logger::Notice<<"stats: outpacket/query ratio "<<(int)(SyncRes::s_outqueries*100.0/SyncRes::s_queries)<<"%";
     g_log<<Logger::Notice<<", "<<(int)(SyncRes::s_throttledqueries*100.0/(SyncRes::s_outqueries+SyncRes::s_throttledqueries))<<"% throttled, "
      <<SyncRes::s_nodelegated<<" no-delegation drops"<<endl;
@@ -2911,6 +2912,8 @@ static void houseKeeping(void *)
         SyncRes::pruneNSSpeeds(limit);
         limit = now.tv_sec - SyncRes::s_serverdownthrottletime * 10;
         SyncRes::pruneFailedServers(limit);
+        limit = now.tv_sec - 2*3600;
+        SyncRes::pruneEDNSStatuses(limit);
       }
       last_prune=time(0);
     }
--- a/pdns/syncres.cc
+++ b/pdns/syncres.cc
@@ -472,6 +472,27 @@ uint64_t SyncRes::doDumpThrottleMap(int fd)
   return count;
 }
 
+uint64_t SyncRes::doDumpFailedServers(int fd)
+{
+  auto fp = std::unique_ptr<FILE, int(*)(FILE*)>(fdopen(dup(fd), "w"), fclose);
+  if(!fp)
+    return 0;
+  fprintf(fp.get(), "; failed servers dump follows\n");
+  fprintf(fp.get(), "; remote IP\tcount\ttimestamp\n");
+  uint64_t count=0;
+
+  for(const auto& i : t_sstorage.fails.getMap())
+  {
+    count++;
+    char tmp[26];
+    ctime_r(&i.second.last, tmp);
+    fprintf(fp.get(), "%s\t%lld\t%s", i.first.toString().c_str(),
+            static_cast<long long>(i.second.value), tmp);
+  }
+
+  return count;
+}
+
 /* so here is the story. First we complete the full resolution process for a domain name. And only THEN do we decide
    to also do DNSSEC validation, which leads to new queries. To make this simple, we *always* ask for DNSSEC records
    so that if there are RRSIGs for a name, we'll have them.
@@ -3150,7 +3171,7 @@ bool SyncRes::doResolveAtThisIP(const std::string& prefix, const DNSName& qname,
       t_sstorage.nsSpeeds[nsName.empty()? DNSName(remoteIP.toStringWithPort()) : nsName].submit(remoteIP, 1000000, &d_now); // 1 sec
 
       // code below makes sure we don't filter COM or the root
-      if (s_serverdownmaxfails > 0 && (auth != g_rootdnsname) && t_sstorage.fails.incr(remoteIP) >= s_serverdownmaxfails) {
+      if (s_serverdownmaxfails > 0 && (auth != g_rootdnsname) && t_sstorage.fails.incr(remoteIP, d_now) >= s_serverdownmaxfails) {
         LOG(prefix<<qname<<": Max fails reached resolving on "<< remoteIP.toString() <<". Going full throttle for "<< s_serverdownthrottletime <<" seconds" <<endl);
         // mark server as down
         t_sstorage.throttle.throttle(d_now.tv_sec, boost::make_tuple(remoteIP, "", 0), s_serverdownthrottletime, 10000);
--- a/pdns/syncres.cc
+++ b/pdns/syncres.cc
@@ -536,16 +536,15 @@ int SyncRes::asyncresolveWrapper(const ComboAddress& ip, bool ednsMANDATORY, con
      If '3', send bare queries
   */
 
-  SyncRes::EDNSStatus* ednsstatus;
-  ednsstatus = &t_sstorage.ednsstatus[ip]; // does this include port? YES
+  SyncRes::EDNSStatus* ednsstatus = &t_sstorage.ednsstatus[ip]; // does this include port? YES
 
-  if(ednsstatus->modeSetAt && ednsstatus->modeSetAt + 3600 < d_now.tv_sec) {
-    *ednsstatus=SyncRes::EDNSStatus();
+  if (ednsstatus->modeSetAt && ednsstatus->modeSetAt + 3600 < d_now.tv_sec) {
+    *ednsstatus = SyncRes::EDNSStatus();
     //    cerr<<"Resetting EDNS Status for "<<ip.toString()<<endl);
   }
 
-  SyncRes::EDNSStatus::EDNSMode& mode=ednsstatus->mode;
-  SyncRes::EDNSStatus::EDNSMode oldmode = mode;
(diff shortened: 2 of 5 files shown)


omoerbeek:
I'll take a look at the EDNS status map as well.

omoerbeek:
The mode is set to `UNKNOWN` if the entry is older than an hour in `asyncResolveWrapper()`, so the prune age can be an hour if I am not mistaken.

rgacogne:
> The mode is set to `UNKNOWN` if the entry is older than an hour in `asyncResolveWrapper()`, so the prune age can be an hour if I am not mistaken.

I expected a longer validity but OK.

pdns/pdns_recursor.cc
```diff
@@ -2908,6 +2910,10 @@ static void houseKeeping(void *)
       if(!((cleanCounter++)%40)) {  // this is a full scan!
 	time_t limit=now.tv_sec-300;
         SyncRes::pruneNSSpeeds(limit);
+        limit = now.tv_sec - SyncRes::s_serverdownthrottletime * 10;
+        SyncRes::pruneFailedServers(limit);
+        limit = now.tv_sec - 2*3600;
+        SyncRes::pruneEDNSStatuses(limit);
```
rgacogne: I'm wondering if we can have an issue due to the fact that `SyncRes::asyncresolveWrapper()` keeps a pointer to an entry of the map during the call to `asyncresolve()`, which will yield the control back to the mthread kernel. I'm pretty sure we have a race condition if the map is cleaned up at that time, and the pointer is no longer valid when `asyncresolve()` returns.
We should likely do a lookup again when we return instead.

omoerbeek: Indeed, the same thought occurred to me this morning. Will redraft.

pdns/syncres.hh
```diff
@@ -214,60 +214,74 @@ private:
...
+  counter_t incr(const Thing& t, const struct timeval & now)
   {
-    typename cont_t::iterator i=d_cont.find(t);
+    typename cont_t::iterator i = d_cont.find(t);
 
-    if(i==d_cont.end()) {
-      d_cont[t]=1;
+    if (i == d_cont.end()) {
+      d_cont[t].value = 1;
+      d_cont[t].last = now.tv_sec;
```
rgacogne: Might be worth taking a reference via `std::map::insert()` to avoid doing the lookup twice?

pdns/syncres.hh
```diff
@@ -214,60 +214,74 @@ private:
...
     d_cont.clear();
   }
+
   size_t size() const
   {
     return d_cont.size();
   }
+
+  void prune(time_t cutoff) {
+    for (auto it = d_cont.begin(); it != d_cont.end(); ) {
```
rgacogne: Any idea of how many entries we usually have in there? If it might be a lot perhaps we need to use a multi_index keyed on the `last` field to avoid having to scan the whole map.

pdns/syncres.hh
```diff
@@ -498,6 +512,16 @@ public:
   {
     t_sstorage.ednsstatus.clear();
   }
+  static void pruneEDNSStatuses(time_t cutoff)
+  {
+    for (auto it = t_sstorage.ednsstatus.begin(); it != t_sstorage.ednsstatus.end(); ) {
+      if (it->second.modeSetAt <= cutoff) {
+        it = t_sstorage.ednsstatus.erase(it);
```
rgacogne: Same question about the number of entries?

phonedph1: On one of our small recursors (sub 1kqps) that has only been up for a week we have 200,000 edns status entries - if that data point helps.

phonedph1: With cleanup, I suppose it will follow the same trend as the nsspeed size, which in that instance is an order of magnitude smaller.

omoerbeek:
In my test runs (with the purging) continously resolving  `alexa-top-1m` in a loop with calidns, I see the ednsstatus count grow to about 600k.  Failed servers hovers around 30k and nsspeeds is the same order (bit a bit bigger) than ednstable.

rgacogne:
> In my test runs (with the purging) continously resolving alexa-top-1m in a loop with calidns, I see the ednsstatus count grow to about 600k. Failed servers hovers around 30k and nsspeeds is the same order (bit a bit bigger) than ednstable.

Are these numbers before without this PR? Or does it stay that high even when regularly purged? 
Although, we won't start purging the EDNS statuses until 2 hours after startup, so I'm guessing it will grow almost up to 600k quickly. So perhaps it would male sense to use a multi_index, with the existing index plus one on the timestamp, so that we only scan expired entries?

omoerbeek:
The numbers I reported are with this PR included, so with purging. I'll look into the multi-index thing soon. I currently do not want to stop this test run so see how it is doing in the long run (multi-days). After than I can add some code to see how fast the purging is and compare it with a multi-index solution.

rgacogne:
> The numbers I reported are with this PR included, so with purging. I'll look into the multi-index thing soon.

OK! Let me know if you are not familiar with how multi-index works and want some pointers (no pun intended)!

omoerbeek:
I'm not familiar with multi-index, but I'd like to give it a shot myself as a start.

BTW, the numbers above are the accumulated ones for a rec with 16 threads. So to get per-thread estimates you need to divide by 16.

omoerbeek:
@rgacogne Can you take a look to see if I'm doing this right? If so we might need to generalize this, since the other tables are purged in a similar manner.

pdns/syncres.hh
```diff
@@ -339,7 +346,30 @@ public:
...
+  struct EDNSStatus {
+    EDNSStatus(const ComboAddress &arg) : address(arg) {}
+    ComboAddress address;
+    time_t modeSetAt{0};
+    enum EDNSMode { UNKNOWN=0, EDNSOK=1, EDNSIGNORANT=2, NOEDNS=3 } mode{UNKNOWN};
+  };
+  
+  struct ednsstatus_t : public multi_index_container<EDNSStatus,
+                                indexed_by<
+                                  ordered_unique<member<EDNSStatus, ComboAddress, &EDNSStatus::address>>,
```
rgacogne: I'd advise the use of named indexes, using tags, see in `recpacketcache.hh` for an example. It is much easier to know which index you are using later without having to look at the definition, and prevent renumbering if you ever want to add an index.

pdns/syncres.hh
```diff
@@ -339,7 +346,30 @@ public:
...
+                                  ordered_non_unique<member<EDNSStatus, time_t, &EDNSStatus::modeSetAt>>
+                                  >> {
+    void reset(nth_index<0>::type &index, iterator it) {
+      index.modify(it, [](EDNSStatus &s) { s.mode = EDNSStatus::EDNSMode::UNKNOWN; s.modeSetAt = 0; }); 
+    }
+    void setMode(nth_index<0>::type &index, iterator it, EDNSStatus::EDNSMode mode) {
+      index.modify(it, [mode](EDNSStatus &s) { s.mode = mode; });
```
rgacogne: You could make the `mode` member `mutable`, since it is not involved in any index, making it possible to update it directly.

pdns/syncres.cc
```diff
@@ -515,16 +536,18 @@ int SyncRes::asyncresolveWrapper(const ComboAddress& ip, bool ednsMANDATORY, con
      If '3', send bare queries
   */
 
-  SyncRes::EDNSStatus* ednsstatus;
-  ednsstatus = &t_sstorage.ednsstatus[ip]; // does this include port? YES
+  SyncRes::ednsstatus_t::iterator ednsstatus = t_sstorage.ednsstatus.find(ip); // does this include port? YES
+  ednsstatus = t_sstorage.ednsstatus.insert(ip).first;
```
rgacogne: I'm not sure why this calls `find()` then always `insert()`?

omoerbeek: Ah, I started with find and then thought: insert does what I want, but some old code remained.

pdns/syncres.hh
omoerbeek: Will do.

omoerbeek:
Here's the graph of resident memory usage after more than 5 days running 4.1.x wih a rough backport of this PR.

Schermafbeelding 2019-11-18 om 08 05 44

omoerbeek:
Friday I became smart enough to have a periodic ps running each hour.
Tow lines from it:
```
2019-11-15T14:00:01+00:00       nobody    1158  286 15.2 5415688 2507572 pts/3 Sl+  Nov12 12048:21 ./pdns_recursor --config-dir=.
```
```
2019-11-18T08:00:01+00:00       nobody    1158  286 16.6 5415688 2740668 pts/3 Sl+  Nov12 23414:51 ./pdns_recursor --config-dir=.
```
While the RSS creep is viible here, it must be noted that the VSZ *does not grow*.
Since there is no pressure on the allocated physical pages on this machine:
```
MiB Mem :  16044.8 total,   5760.7 free,   3137.4 used,   7146.7 buff/cache
MiB Swap:  16384.0 total,  16384.0 free,      0.0 used.  12431.9 avail Mem 
```

omoerbeek:
Another observation: I applied memory pressure to the point where the machine started needing swap by running a program allocating and touching lots of memory. I saw almost the whole rec process being pages out to swap. After I stopped the memory eating process, about 1.1G of rec was paged in again after 10min or so. The whole time VSZ remained constant. Now the process is at:
```
nobody    1158  286  7.0 5415688 1151420 pts/3 Sl+  Nov12 23729:33 ./pdns_recursor --config-dir=.
```
So the short-term working set of this process is about 1.1G.

omoerbeek:
Measurement shows that a linear scan of the threads ednsstatus map takes about 0.024s for  123515 entries.  This was on a 2 thread test run, where (with the given test setup) it is expected to grow to about 300k entries per thread. For the 16 threads case the table size would be about 600k/16 = 37500. Testing that now...

omoerbeek:
I'll be squashing the multi_index code out, save a copy for future reference.

```c++
class fails_t : public boost::noncopyable
{
public:
  typedef unsigned long counter_t;
  struct value_t {
    value_t(const ComboAddress &a) : address(a) {}
    ComboAddress address;
    mutable counter_t value{0};
    time_t last{0};
  };

  typedef multi_index_container<value_t,
                                indexed_by<
                                  ordered_unique<tag<ComboAddress>, member<value_t, ComboAddress, &value_t::address>>,
                                  ordered_non_unique<tag<time_t>, member<value_t, time_t, &value_t::last>>
                                  >> cont_t;

  cont_t getMap() const {
    return d_cont;
  }
  counter_t value(const ComboAddress& t) const
  {
    auto i = d_cont.find(t);

    if (i == d_cont.end()) {
      return 0;
    }
    return i->value;
  }

  counter_t incr(const ComboAddress& t, const struct timeval & now)
  {
    auto i = d_cont.insert(t).first;

    if (i->value < std::numeric_limits<counter_t>::max()) {
      i->value++;
    }
    auto &ind = d_cont.get<ComboAddress>();
    ind.modify(i, [t = now.tv_sec](value_t &val) { val.last = t;});
    return i->value;
  }

  void clear(const ComboAddress& a)
  {
    d_cont.erase(a);
  }

  void clear()
  {
    d_cont.clear();
  }

  size_t size() const
  {
    return d_cont.size();
  }

  void prune(time_t cutoff) {
    auto &ind = d_cont.get<time_t>();
    ind.erase(ind.begin(), ind.upper_bound(cutoff));
  }

private:
  cont_t d_cont;
};

struct EDNSStatus {
    EDNSStatus(const ComboAddress &arg) : address(arg) {}
    ComboAddress address;
    time_t modeSetAt{0};
    mutable enum EDNSMode { UNKNOWN=0, EDNSOK=1, EDNSIGNORANT=2, NOEDNS=3 } mode{UNKNOWN};
  };
  
  struct ednsstatus_t : public multi_index_container<EDNSStatus,
                                indexed_by<
                                  ordered_unique<tag<ComboAddress>, member<EDNSStatus, ComboAddress, &EDNSStatus::address>>,
                                  ordered_non_unique<tag<time_t>, member<EDNSStatus, time_t, &EDNSStatus::modeSetAt>>
                                  >> {
    void reset(index<ComboAddress>::type &ind, iterator it) {
      ind.modify(it, [](EDNSStatus &s) { s.mode = EDNSStatus::EDNSMode::UNKNOWN; s.modeSetAt = 0; }); 
    }
    void setMode(index<ComboAddress>::type &ind, iterator it, EDNSStatus::EDNSMode mode) {
      it->mode = mode;
    }
    void setTS(index<ComboAddress>::type &ind, iterator it, time_t ts) {
      ind.modify(it, [ts](EDNSStatus &s) { s.modeSetAt = ts; });
    }

  };
```

rgacogne:
> For the 16 threads case the table size would be smaller.

This assumes that the number of authoritative servers contacted by a thread decreases with the number of threads, which might be true in the long-term. It's clearly not true when `pdns-distributes-queries` is off.

>  I think a good plan is:
> 1. merge the linear versions of the cleanup for failed servers and ednsstatus into master
> 2. backport it
> 3. later rework all tables that should be purged on a time based decision to use a multi_index map. A unification to avoid re-implementing the same logic in multiple places is needed. It might be even possible to use fewer tables that combine the data that is now stored in several separate tables.

OK, I can see how that makes sense. I'm still a bit worried about the impact of linear scanning in real-world setups, though. We have learned in the past that things that were negligible in small setups could have dramatic consequences in large scale ones, like metrics reporting, pipes being blocking and so on. So I think we really should get the multi_index version in master quickly, so that we have a solution ready if the linear scan proves to be a problem.

omoerbeek:
Agreed. But note that the 600k edns size was an observed *aggregate* number of a 16 threads test run.

omoerbeek:
@rgacogne can you take a last look? esp. the `SyncRes::asyncresolveWrapper()` part. tx.

pdns/syncres.cc
```diff
@@ -556,32 +576,35 @@ int SyncRes::asyncresolveWrapper(const ComboAddress& ip, bool ednsMANDATORY, con
     else {
       ret=asyncresolve(ip, sendQname, type, doTCP, sendRDQuery, EDNSLevel, now, srcmask, ctx, d_outgoingProtobufServers, d_frameStreamServers, luaconfsLocal->outgoingProtobufExportConfig.exportTypes, res, chained);
     }
+    // ednsstatus might be cleared, so do a new lookup
+    ednsstatus = &t_sstorage.ednsstatus[ip]; // does this include port? YES
+    mode = &ednsstatus->mode;
```
rgacogne: I wonder if we should update `oldmode` as well here? The only difference it would make is that if the status has already been updated in the meantime, we will set `modeSetAt` again. It's a matter of seconds, though, so it probably doesn't matter.

PR #8528, opened 2019-11-13
rec: Backport 8470 to rec 4.2.x:  prime NS records of root-servers.net parent (.net) 

omoerbeek:
### Checklist

I have:
- [X] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/reczones.cc
+++ b/pdns/reczones.cc
@@ -34,6 +34,14 @@ extern char** g_argv;
 
 static thread_local set<DNSName> t_rootNSZones;
 
+static void insertIntoRootNSZones(const DNSName &name) {
+  // do not insert dot, wiping dot's NS records from the cache in primeRootNSZones()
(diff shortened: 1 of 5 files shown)


pdns/reczones.cc
```diff
@@ -88,12 +100,39 @@ void primeHints(void)
...
+{
+  struct timeval now;
+  gettimeofday(&now, 0);
+  SyncRes sr(now);
+
+  if (dnssecmode) {
+    sr.setDoDNSSEC(true);
+    sr.setDNSSECValidationRequested(true);
+  }
+  for (const auto & qname: t_rootNSZones) {
```
rgacogne: Aren't you introducing the same issue than the one fixed in #8510 ?

omoerbeek: Grumble... I thought I included that fix... glad you are paying attention

PR #8532, opened 2019-11-14
rec: qname-minimization metrics 

phonedph1:
### Short description
Keep a counter of the number of outqueries that resort to 'use the full qname'

I have only tested rec_control gets for this so far.

Thoughts?

### Checklist
I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


phonedph1:
This might make more sense to call it something like `fullqname-success` and only count noerror results.

omoerbeek:
Thanks for working on this.

I'd suggest a name like `qname-min-fallback-success` and indeed only count successful ones.

phonedph1:
Okay, update and tested rec_control, /metrics & SNMP (which was a huge pain).

PR #8537, opened 2019-11-15
dnsdist: Rule stats for Prometheus

phonedph1:
### Short description
We have a desire to capture rule matches and aggregate them in Prometheus.  This does create some nasty long attributes in some cases, but putting this out there to see how offensive this really is. 

Produces output like:

```
dnsdist_rule_action{id="0",creationOrder="0",uuid="4bc426c5-b41e-4661-b559-ede4ee56472f",rule="All",action="delay by 40 msec"} 0
dnsdist_rule_response{id="0",creationOrder="1",uuid="86d7b044-677d-4b3e-94fc-10e9f9f43056",rule="All",action="delay by 50 msec"} 0
dnsdist_rule_cacheHit{id="0",creationOrder="2",uuid="94470e79-0c4b-48dc-be4d-d327bee56783",rule="All",action="delay by 60 msec"} 0
dnsdist_rule_selfAnsweredResponse{id="0",creationOrder="3",uuid="194b8181-3b34-4a16-9ac9-a3e8c2d3efa7",rule="All",action="delay by 70 msec"} 0
```

TODO: HELP/TYPE stuff.

### Checklist
I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdist-web.cc
+++ b/pdns/dnsdist-web.cc
@@ -245,6 +245,22 @@ static json11::Json::array someResponseRulesToJson(GlobalStateHolder<vector<T>>*
   return responseRules;
 }
 
+template<typename T>
+static string someResponseRulesToPrometheus(const string basename, GlobalStateHolder<vector<T>>* someResponseRules)
+{
+  int num = 0;
+  auto localResponseRules = someResponseRules->getLocal();
+  ostringstream output;
+  for(const auto& a : *localResponseRules) {
+    const std::string rule_info = boost::str(boost::format("{id=\"%1%\",creationOrder=\"%2%\",uuid=\"%3%\",rule=\"%4%\",action=\"%5%\"}")
+      % num % (double)a.d_creationOrder % boost::uuids::to_string(a.d_id) % a.d_rule->toString() % a.d_action->toString());
+    output << basename << rule_info << " " << a.d_rule->d_matches << "\n";
+    num++;
+  }
+  return output.str();
+}
+
+
 static void connectionThread(int sock, ComboAddress remote)
 {
   setThreadName("dnsdist/webConn");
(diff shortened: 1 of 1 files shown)


hhoffstaette:
I've added this to my 1.4.0 package and noticed that the uuid of every rules changes between restarts/reboots, essentially invalidating a rule series and leading to label bloat. Is the uuid really required or would it suffice to use the rule itself to match series? Same for id & creationOrder - the series of a rule shouldn't change/be invalidated just because an unrelated rule was inserted/deleted.
Also I only see _action, no _response/_cacheHit/_selfAnsweredResponse labels.

phonedph1:
Good question. I'm guessing most people probably don't need them, but I think the JSON endpoint exposed them so I just did the same. Our use case is typically going to be aggregating based on something in the rule or action text. 

Do you actually have non-action rules setup for it to show?

Some others on IRC have also commented that they have so many rules this part should perhaps be optional all togerher.

hhoffstaette:
> Do you actually have non-action rules setup for it to show?

So far I only use Spoof or NXDomain actions. It seems empty tables result in empty strings and therefore don't result in output; IMHO in this case the metrics should still appear and have a default 0 value.

rgacogne:
> I've added this to my 1.4.0 package and noticed that the uuid of every rules changes between restarts/reboots, essentially invalidating a rule series and leading to label bloat. Is the uuid really required or would it suffice to use the rule itself to match series?

Note that you can fix the UUID in the configuration so that it stays the same across restarts.

hhoffstaette:
> Note that you can fix the UUID in the configuration so that it stays the same across restarts.

True - thanks for the reminder. I suspect the number of people who do so is very low, and it doesn't change the wrong behaviour by default (which is surprising, to say the least). It would probably be much easier to switch the rule ID to a stable value - something like action code xor rule hash. I don't know whether the actual format of the id matters, but without stability the value is essentially useless anyway. If people want to autogenerate rules with explicit externally-generated UUIDs they could still do so.

pdns/dnsdist-web.cc
```diff
@@ -531,6 +546,30 @@ static void connectionThread(int sock, ComboAddress remote)
...
+        auto responseRules = someResponseRulesToPrometheus(rule_response, &g_resprulactions);
+        output << responseRules;
+
+        const string rule_cacheHit = "dnsdist_rule_cachehit";
+        output << "# HELP " << rule_cacheHit << " Number of queries matching the specified cache hit rule" << "\n";
+        output << "# TYPE " << rule_cacheHit << " counter" << "\n";
+        auto cacheHitResponseRules = someResponseRulesToPrometheus(rule_cacheHit, &g_cachehitresprulactions);
+        output << cacheHitResponseRules;
+
+        const string rule_selfAnsweredResponse = "dnsdist_rule_selfansweredresponse";
```
jsoref, line 567: Hi, for the time being (and hopefully going forward), pdns master is running the [check-spelling](https://github.com/marketplace/actions/check-spelling) action. In order to avoid a ❌ when merged https://github.com/jsoref/pdns/commit/6352b9816ec2624646bfc68380f29a90422e36c2#commitcomment-39229536 -- you'll need to merge/rebase on `master` to get the check (and the files) -- you can just accept the advice (which should add `selfansweredresponse` to `expect.txt`) as `selfansweredresponse` appears to be the proper thing here.

pieterlexis:
@rgacogne is something like this already implemented? If so, please close this pr

phonedph1:
We've diverged a bit from the initial PR here. We are only using `rule` and `action` now and will likely add `name` for 1.6.0. There was also some feedback on IRC from a large operator saying having these rules display by default would be too much for their prometheus collection. Not sure how much agreement here will be on what should be included or not.

phonedph1:
Extremely similar functionality is implemented in #9746 and as such I am closing this.

PR #8538, opened 2019-11-18
Use move semantics when updating the content of the StateHolder

rgacogne:
### Short description

This PR prevents unnecessary copies when adding/updating rules in dnsdist, and possibly for some configuration updates of the recursor as well. I only measured the dnsdist case, to be clear.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdist-lua-actions.cc
+++ b/pdns/dnsdist-lua-actions.cc
@@ -1173,7 +1173,7 @@ private:
 };
 
 template<typename T, typename ActionT>
-static void addAction(GlobalStateHolder<vector<T> > *someRulActions, luadnsrule_t& var, std::shared_ptr<ActionT>& action, boost::optional<luaruleparams_t>& params) {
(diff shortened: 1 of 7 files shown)


pdns/dnsdist-lua-actions.cc
```diff
@@ -1173,16 +1173,16 @@ class KeyValueStoreLookupAction : public DNSAction
 };
 
 template<typename T, typename ActionT>
-static void addAction(GlobalStateHolder<vector<T> > *someRulActions, luadnsrule_t var, std::shared_ptr<ActionT> action, boost::optional<luaruleparams_t> params) {
+static void addAction(GlobalStateHolder<vector<T> > *someRulActions, luadnsrule_t& var, std::shared_ptr<ActionT>& action, boost::optional<luaruleparams_t>& params) {
```
omoerbeek: I don't know how strict we want to be with respect to const-correctness, but you can probably add some `const` for these parameters.

rgacogne: Thanks, fixed!

PR #8562, opened 2019-11-20
YAHTTP: Accept headers without spaces

pieterlexis:
### Short description
cc @cmouse 

YaHTTP did not accept a header like

```
Accept:application/json
```

Which should be allowed according to RFC 7230:

>    Each header field consists of a case-insensitive field name followed by a colon (":"), optional leading whitespace, the field value, and optional trailing whitespace.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/ext/yahttp/yahttp/reqresp.cpp
+++ b/ext/yahttp/yahttp/reqresp.cpp
@@ -101,10 +101,11 @@ namespace YaHTTP {
           break;
         }
         // split headers
-        if ((pos1 = line.find(": ")) == std::string::npos)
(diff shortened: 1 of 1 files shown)


Habbie:
This does not appear to accept *more* than one space after the `:`

omoerbeek:
This was only backported to auth 4.2.x, is this relevant to rec-4.2.x?

Habbie:
> This was only backported to auth 4.2.x, is this relevant to rec-4.2.x?

Yes, it is!

PR #8564, opened 2019-11-20
geoipbackend: propagate weighted rounding gap fix

jeremy-clerc:
### Short description

618824cb73db384d52cf379c7d49744dbb5b8f6d (PR #7219) fixed computation
of weighted round robin computation per Qtype.

That commit also introduced a bug on the removal of the rounding gap.

Because it works on a copy of a GeoIPDNSResourceRecord, the updated
weight of the last record is not propagated to the domain object.

The bug is as follow:
  1. The total weight is < 1000
  2. The random probabability gives 1000
  3. There is no resource to be given

This commit restores the original behavior of ensuring the gap is added
to the last resource so we always have 1000.

### Checklist
I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/modules/geoipbackend/geoipbackend.cc
+++ b/modules/geoipbackend/geoipbackend.cc
@@ -263,7 +263,7 @@ void GeoIPBackend::initialize() {
     for(auto &item: dom.records) {
       map<uint16_t, float> weights;
       map<uint16_t, float> sums;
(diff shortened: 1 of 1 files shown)


jeremy-clerc:
Do you think it would be possible to have a bugfix release once merged ? Or there is some sort of nightly build on 4.2.x branch (when cherry picked) ?
Thanks

PR #8565, opened 2019-11-20
geoipbackend: stop looking after first weighted match

jeremy-clerc:
### Short description

In the case of weighted round robin, we will give back only one
GeoIPDNSResourceRecord, so once we found it, there is no point to
keep looking except inducing a performance penalty.

This commit ensures we stop looking after we found our first weighted
match.

### Checklist
I have:
- [X] compiled this code
- [X] tested this code
- [X] documented the code


Habbie:
@cmouse

PR #8567, opened 2019-11-21
dnsdist: Add bounded loads to the consistent hashing policy

### LINKED ISSUE #7387: dnsdist needs a 'bounded hashing' server selection policy
See #7386, but then for dnsdist. Could use number of outstanding queries as metric, or QPS. Or perhaps configurable.

In short, this means we hash a query and pick the same server for the same query UNLESS this server is receiving too large a share of traffic or is too busy.

rgacogne:
### Short description

This PR adds support for a bounded-load version of the consistent hashing policy, preventing one server from receiving much more queries than the others, even if the distribution of queries is not perfect.
This "consistent hashing with bounded loads" algorithm is enabled by setting `setConsistentHashingBalancingFactor()` to a value other than 0, which is the default. This value is the maximum number of outstanding queries that a given server can have at a given time, as a ratio of the average number of outstanding queries for all the active servers in the pool.
For example, setting `setConsistentHashingBalancingFactor(1.5)` means that no server will be allowed to have more outstanding queries than 1.5 times the average of all outstanding queries in the pool. The algorithm will try to select a server based on the hash of the qname, as is done when no bounded-load is set, but will disqualify all servers that have more outstanding queries than the average times the factor, until a suitable server is found. The higher the factor, the more imbalance between the servers is allowed.

Implements the `chashed` part of #7387, not the `whashed` one.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdist.cc
+++ b/pdns/dnsdist.cc
@@ -896,7 +896,7 @@ shared_ptr<DownstreamState> chashed(const NumberedServerVector& servers, const D
   shared_ptr<DownstreamState> ret = nullptr, first = nullptr;
 
   double targetLoad = std::numeric_limits<double>::max();
-  if (g_consistentHashBalancingFactor != 0) {
+  if (g_consistentHashBalancingFactor > 0) {
     /* we start with one, representing the query we are currently handling */
     double currentLoad = 1;
     for (const auto& pair : servers) {
--- a/pdns/dnsdist.cc
+++ b/pdns/dnsdist.cc
@@ -882,6 +882,7 @@ shared_ptr<DownstreamState> wrandom(const NumberedServerVector& servers, const D
 }
 
 uint32_t g_hashperturb;
+double g_consistentHashBalancingFactor = 0;
 shared_ptr<DownstreamState> whashed(const NumberedServerVector& servers, const DNSQuestion* dq)
 {
   return valrandom(dq->qname->hash(g_hashperturb), servers, dq);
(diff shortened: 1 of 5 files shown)


pdns/dnsdist.cc
```diff
@@ -894,8 +895,18 @@ shared_ptr<DownstreamState> chashed(const NumberedServerVector& servers, const D
   unsigned int min = std::numeric_limits<unsigned int>::max();
   shared_ptr<DownstreamState> ret = nullptr, first = nullptr;
 
+  double targetLoad = std::numeric_limits<double>::max();
+  if (g_consistentHashBalancingFactor != 0) {
```
chbruyand: Maybe verify it is `> 0`, or validate `setConsistentHashingBalancingFactor()` input parameter ?

pdns/dnsdistdist/docs/guides/serverselection.rst
```diff
@@ -46,10 +46,19 @@ The current hash algorithm is based on the qname of the query.
 ``chashed``
 ~~~~~~~~~~~
 
+.. versionadded: 1.3.3
+
 ``chashed`` is a consistent hashing distribution policy. Identical questions with identical hashes will be distributed to the same servers. But unlike the ``whashed`` policy, this distribution will keep consistent over time. Adding or removing servers will only remap a small part of the queries.
 
+Increasing the weight of servers to a value larger than the default is required to get a good distribution of queries. Small values like 100 or 1000 should be enough to get a correct distribution.
```
chbruyand: I was wondering if it might help to cap weight to a min value (100?) when using consistent hashing-based distribution ?

rgacogne:
> This could be useful to also implement that for `whashed` policy, maybe remove the closes #7387 comment ?

True, I edited it.

PR #8571, opened 2019-11-22
Rec: better time based data structures

omoerbeek:
### Short description

- Use `multi_index` for all time based data structures except `nsSpeeds`. That will come later.
- `nsSpeeds` *is* cleaned up up to be more memory-compact and to use `float` instead of `double` consistently.
- some cleanup of the time based cleanup condition in in housekeeping. Adding the thread id make no sense to me am at all. Not testing on a `time_t` only while spread out the cleanup in various threads to be not simultaneously.
- One condition in the throttle class in `syncres.hh` is marked ????, I copied it from the original code, but I have a hard time seeing this is the right condition.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [X] documented the code

### DIFF
--- a/pdns/syncres.hh
+++ b/pdns/syncres.hh
@@ -113,7 +113,9 @@ public:
     if (i == d_cont.end()) {
       entry_t e = { t, ttd, count };
       d_cont.insert(e);
-    } else if (i->ttd > ttd || i->count < count) { // ????
+    } else if (ttd > i->ttd || count > i->count) {
+      ttd = std::max(i->ttd, ttd);
+      count = std::max(i->count, count);
       auto &ind = d_cont.template get<Thing>();
       ind.modify(i, [ttd,count](entry_t &e) { e.ttd = ttd; e.count = count; });
     }
--- a/pdns/syncres.hh
+++ b/pdns/syncres.hh
@@ -152,39 +152,27 @@ private:
 class DecayingEwma
 {
 public:
-  DecayingEwma() :  d_val(0.0)
-  {
-    d_needinit=true;
-    d_last.tv_sec = d_last.tv_usec = 0;
-    d_lastget=d_last;
-  }
-
+  DecayingEwma() {}
   DecayingEwma(const DecayingEwma& orig) = delete;
   DecayingEwma & operator=(const DecayingEwma& orig) = delete;
   
   void submit(int val, const struct timeval& now)
   {
-    if(d_needinit) {
-      d_last=now;
-      d_lastget=now;
-      d_needinit=false;
+    if (d_last.tv_sec == 0 && d_last.tv_usec == 0) {
+      d_last = now;
       d_val = val;
     }
     else {
-      float diff= makeFloat(d_last - now);
-
-      d_last=now;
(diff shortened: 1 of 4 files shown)


pdns/syncres.hh
```diff
@@ -141,10 +134,13 @@ public:
     d_cont.clear();
   }
 
+  void prune() {
+    time_t now = time(NULL);
```
rgacogne:
```suggestion
    time_t now = time(nullptr);
```

pdns/syncres.hh
```diff
@@ -156,110 +152,84 @@ private:
...
+      float diff = makeFloat(d_last - now);
+      d_last = now;
+      float factor = expf(diff)/2.0f; // might be '0.5', or 0.0001
+      d_val = (1-factor)*val + factor*d_val;
     }
   }
 
-  float get(const struct timeval* tv)
+  float get(float factor)
   {
```
rgacogne: I understand that not computing the factor for every collection in a `DecayingEwmaCollection` is a nice optimization, but it feels like we are crippling the `DecayingEwma` class here, since it can no longer work outside of a `DecayingEwmaCollection`.

omoerbeek: True, but in this case I think it is worth it. 

It's not only avoiding the computation which produces the same value for every item in the collection, but also not having to store a `struct timeval` per item.

`DecayEwma` is only used for one purpose atm. If we encounter a case where it does make sense to store a timestamp per item it's easy to extend the class, adding a `struct timeval and a `get()` taking a now value and computing the factor.

But maybe the biggest benefit is that we have the `d_lastget` member available in the `DecayingEwmaCollection` class, making it more easy to define an index on that field in the future.

pdns/syncres.hh
```diff
@@ -78,60 +78,53 @@ typedef std::unordered_map<
...
+    time_t ttd = now + ttl;
+    if (i == d_cont.end()) {
+      entry_t e = { t, ttd, count };
+      d_cont.insert(e);
+    } else if (i->ttd > ttd || i->count < count) { // ????
+      auto &ind = d_cont.template get<Thing>();
+      ind.modify(i, [ttd,count](entry_t &e) { e.ttd = ttd; e.count = count; });
     }
-    else if(i->second.ttd > e.ttd || (i->second.count) < e.count)
-      d_cont[t]=e;
```
omoerbeek: @rgacogne can you explain the conditions used to decide to modify or not?

rgacogne: IMHO we should always replace by the maximum of the TTDs and the maximum of the number of attempts. It looks like the existing code replaces if the new TTD is smaller or equal to the existing one which doesn't make sense to me.

omoerbeek: Good, I also could not make any sense of the existing condition.

pdns/syncres.hh
rgacogne: Understood. It seems to me that `getFactor()` factor might then be better suited as a static method of `DecayingEwma` taking a `struct timeval` indicating the latest retrieval time and another one indicating the current time, but that's not important here.

PR #8585, opened 2019-11-27
Auth API: be a bit more lenient

pieterlexis:
### Short description
While working with openapi-codegen generated code, it seems that the auth API might be a bit too strict when it comes to the createZone call.

This PR addresses 2 of these issues.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/ws-auth.cc
+++ b/pdns/ws-auth.cc
@@ -1535,8 +1535,8 @@ static void apiServerZones(HttpRequest* req, HttpResponse* resp) {
       throw ApiException("You cannot give rrsets AND zone data as text");
 
     auto nameservers = document["nameservers"];
-    if (!nameservers.is_array() && zonekind != DomainInfo::Slave)
-      throw ApiException("Nameservers list must be given (but can be empty if NS records are supplied)");
+    if (!nameservers.is_null() && !nameservers.is_array() && zonekind != DomainInfo::Slave)
+      throw ApiException("Nameservers is not a list");
 
     string soa_edit_api_kind;
     if (document["soa_edit_api"].is_string()) {


Habbie:
Swagger/OpenAPI 2.0 (which we use) does not actually define what the types mean (or at least, I cannot find out). However, 3.0 has this to say:

> Boolean
> type: boolean represents two values: true and false. Note that truthy and falsy values such as "true", "", 0 or null are not considered boolean values.

pieterlexis:
> Swagger/OpenAPI 2.0 (which we use) does not actually define what the types mean (or at least, I cannot find out). However, 3.0 has this to say: [...]

Although I agree with you. Right now this solves an immediate problem I have with the generated code (and others might have similar issues).

Habbie:
Suggestion:
* don't change the .yml
* add comments to the .cc explaining that one wrapper misbehaves like this

zeha:
https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md has to say this:

> Primitive data types in the Swagger Specification are based on the types supported by the JSON-Schema Draft 4. Models are described using the Schema Object which is a subset of JSON Schema Draft 4.

JSON Schema Draft 4 then goes on to define:

>    boolean  A JSON boolean.

In my understanding a JSON boolean is only the literals true or false.

Maybe its useful to wrap the comparisons in a function, and also add the comment there.

pieterlexis:
I'll update this PR to remove the 1/0 commit, but I still believe that we should accept `null` for `nameservers`. I've fixed the client issue: https://github.com/Mermade/openapi-codegen/pull/71

PR #8594, opened 2019-11-28
auth: add default-publish-{cds|cdnskey} options

### LINKED ISSUE #6466: Feature suggestion: configuration option to publish CDS/CDNSKEY by default
- Program: Authoritative
 - Issue type: Feature request

### Short description

I would like to suggest two new configuration settings `default-publish-cds` and `default-publish-cdnskey` which would instruct PowerDNS to automatically publish `CDS` / `CDNSKEY` records when a zone is being secured with DNSSEC. 

### Usecase

Right now, .cz is the only TLD which supports RFC8078/RFC7344 to automatically manage DS record in the parent, but other TLDs are starting to implement it (.ch and .li ) and hopefully more will follow. About 4% of DNSSEC signed domains in .CH currently are islands of security (no DS in parent), one of the reasons being poor or missing DNSSEC support at the registrar.
We're hoping to boost DNSSEC adoption with CDS.

For hosting companies which serve most domains in these ccTLDs this new setting would allow a atomic operation (`secure-zone`) to simultaneously sign a zone and automatically
update the chain of trust. Right now this requires two operations ( `secure-zone` + `set-publish-cds` ), combining them would make automation simpler and of course be a nice bragging opportunity ("protecting your domain with DNSSEC requires only one command!")

### Description
The setting would work similar to current `default-*` settings. As soon as a zone is secured, for example through `pdnsutil secure-zone` , PowerDNS would check this setting and automatically 
activate the `PUBLISH-CDNSKEY` or `PUBLISH-CDS` domain metadata if not empty.

The format could be the same as the allowed values for the domainmetadata:

 - `default-publish-cdnskey`: `boolean`, default `no`
 - `default-publish-cds`: comma separated list of digest algorithm numbers , default empty

Thanks for considering!

Habbie:
### Short description
Fixes #6466. ~Needs docs.~ has docs

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/common_startup.cc
+++ b/pdns/common_startup.cc
@@ -207,6 +207,8 @@ void declareArguments()
   ::arg().set("default-zsk-algorithm","Default ZSK algorithm")="";
   ::arg().set("default-zsk-size","Default ZSK size (0 means default)")="0";
   ::arg().set("max-nsec3-iterations","Limit the number of NSEC3 hash iterations")="500"; // RFC5155 10.3
+  ::arg().set("default-publish-cdnskey","Default value for PUBLISH-CDNSKEY")="";
+  ::arg().set("default-publish-cds","Default value for PUBLISH-CDS")="";
 
   ::arg().set("include-dir","Include *.conf files from this directory");
   ::arg().set("security-poll-suffix","Domain name from which to query security update notifications")="secpoll.powerdns.com.";
--- a/pdns/dbdnsseckeeper.cc
+++ b/pdns/dbdnsseckeeper.cc
@@ -388,6 +388,11 @@ bool DNSSECKeeper::setPublishCDS(const DNSName& zname, const string& digestAlgos
   return d_keymetadb->setDomainMetadata(zname, "PUBLISH-CDS", meta);
 }
 
+void DNSSECKeeper::getPublishCDS(const DNSName& zname, std::string& value)
+{
+  getFromMetaOrDefault(zname, "PUBLISH-CDS", value, ::arg()["default-publish-cds"]);
+}
+
 /**
  * Remove domainmetadata to stop publishing CDS records for zone zname
  *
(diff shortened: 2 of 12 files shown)


pdns/common_startup.cc
```diff
@@ -207,6 +207,8 @@ void declareArguments()
   ::arg().set("default-zsk-algorithm","Default ZSK algorithm")="";
   ::arg().set("default-zsk-size","Default ZSK size (0 means default)")="0";
   ::arg().set("max-nsec3-iterations","Limit the number of NSEC3 hash iterations")="500"; // RFC5155 10.3
+  ::arg().set("default-publish-cdnskey","Default value for PUBLISH-CDNSKEY")="";
```
pieterlexis: Perhaps this should be a switch? It is only an integer in the metadata because other metadata uses 0/1 for bools.

```suggestion
  ::arg().setSwitch("default-publish-cdnskey","Set PUBLISH-CDNSKEY to 1 for all zones without this metadata item");
```

pdns/dbdnsseckeeper.cc
```diff
@@ -191,10 +191,18 @@ bool DNSSECKeeper::activateKey(const DNSName& zname, unsigned int id)
...
+    return;
+  else
+    value = defaultvalue;
+}
 
-void DNSSECKeeper::getFromMeta(const DNSName& zname, const std::string& key, std::string& value)
+bool DNSSECKeeper::getFromMeta(const DNSName& zname, const std::string& key, std::string& value)
 {
   static int ttl = ::arg().asNum("domain-metadata-cache-ttl");
+  bool isset = false;
```
pieterlexis: I like!

pdns/common_startup.cc
Habbie: It's not a switch, it has three potential values (0/1/empty).

Habbie:
> Could use some tests

Test pushed

PR #8608, opened 2019-12-04
geoipbackend: accept custom lookup mapping

jeremy-clerc:
If for example you want a per country granularity, but still want to
group by custom regions: uk, fr, be -> eu-central, pt, es -> eu-south;
you need to to use %cc and create identical country records. It means
you zones file can become huge and powerdns does not like it too much.

For a user to have a custom mapping without the need to rewrite the
GeoIP database, with this commit, he/she can specify a custom mapping,
so if we get from the GeoIP database fr, we will look in the custom
mapping and find eu-central. So we only need to create the eu-central
record and it will be used by for uk, fr and be as per our first
example.

Definition of mapping_lookup_formats or custom_mapping at the domain
level has priority, but if not defined the global config will be used
as default.

The custom lookup formats and mapping are specified in the zones file:

```
  ---
  mapping_lookup_formats: ['%cc-%re', '%cc']
  custom_mapping:
    'fr': 'eu-central'
    'be': 'eu-central'
    'es': 'eu-south'
    'pt': 'eu-south'
    'us-ca': 'us-west'
    'us-tx': 'us-south'
  domains:
  - domain: example.com
    services:
      www.example.com: [ '%mp.www.example.com' ]
    records:
      eu-central.www.example.com:
        - A: 1.1.1.1
      eu-south.www.example.com:
        - A: 1.1.1.2
      us-west.www.example.com:
        - A: 1.1.1.3
      us-south.www.example.com:
        - A: 1.1.1.4
  - domain: example2.com
    mapping_lookup_formats: ['%cc']
```

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [X] documented the code

### DIFF
--- a/docs/backends/geoip.rst
+++ b/docs/backends/geoip.rst
@@ -145,6 +145,14 @@ that the ``‐`` before certain keys is part of the syntax.
         service.geo.example.com:
           default: [ '%co.%cn.service.geo.example.com', '%cn.service.geo.example.com' ]
           10.0.0.0/8: 'internal.service.geo.example.com'
+    mapping_lookup_formats: ['%cc-%re', '%cc']
+    custom_mapping:
+      fr: eu-central
+      be: eu-central
+      es: eu-south
+      pt: eu-south
+      us-tx: us-south
+      us-ca: us-south
 
 Keys explained
 ~~~~~~~~~~~~~~
@@ -165,6 +173,16 @@ Keys explained
 
   :services: Defines one or more services for querying.
              Each service name may have one or more placeholders.
+  :mapping_lookup_formats: Defines which format to interpolate when using the ``%mp`` placeholder. Each entry
+                           is looked up in the given order and stops at first match.
+                           This allows using a fine granularity, (e.g. per country), while limiting the number
+                           of records to create.
+                           You can use any placeholder, except ``%mp`` to avoid recursion, within the given
+                           format (e.g. %cc).
+  :custom_mapping: Defines the mapping between the lookup format and a custom value to replace ``%mp`` placeholder.
+
+:mapping_lookup_formats: Same as per domain, but used as default value if not defined at the domain level.
+:custom_mapping: Same as per domain, but used as default value if not defined at the domain level.
 
 .. note::
 
@@ -203,6 +221,16 @@ These placeholders disable caching for the record completely:
 :%ip: Client IP address
 :%ip4: Client IPv4 address
 :%ip6: Client IPv6 address
+
+Following placeholder allows custom mapping:
+
+:%mp: Use formats in ``mapping_lookup_formats`` and use user defined ``custom_mapping``
+
+.. versionadded:: 4.4.0
+
+  These placeholders have been added in version 4.4.0:
+
+  - %mp to expand user defined custom formats.
 
 .. versionadded:: 4.2.0
 
--- a/modules/geoipbackend/geoipbackend.cc
+++ b/modules/geoipbackend/geoipbackend.cc
@@ -51,6 +51,8 @@ struct GeoIPDomain {
   int ttl;
   map<DNSName, GeoIPService> services;
   map<DNSName, vector<GeoIPDNSResourceRecord> > records;
+  vector<string> mapping_lookup_formats;
+  map<std::string, std::string> custom_mapping;
 };
 
 static vector<GeoIPDomain> s_domains;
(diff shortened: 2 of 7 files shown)


Habbie:
Wow! Nice work! docs, tests, everything!

modules/geoipbackend/geoipbackend.cc
```diff
@@ -547,7 +598,16 @@ string GeoIPBackend::format2str(string sformat, const Netmask& addr, GeoIPNetmas
     string rep;
     int nrep=3;
     tmp_gl.netmask = 0;
-    if (!sformat.compare(cur,3,"%cn")) {
+    if (!sformat.compare(cur,3,"%mp")) {
+      rep = "unknown";
+      for (auto lformat : dom.mapping_lookup_formats) {
```
cmouse: const auto& lformat

cmouse:
other than that one nit, lgtm.

modules/geoipbackend/geoipbackend.cc
jeremy-clerc: Done. Waiting for the tests.

cmouse:
The requested changes are mainly that please use references and const references when possible. Otherwise it can easily lead into expensive copy operations for the YAML nodes.

jeremy-clerc:
So I tried to put the reference for mapping and formats, but I get:
 ```
                                                               ^
geoipbackend.cc:258:56: error: invalid initialization of non-const reference of type ‘YAML::Node&’ from an rvalue of type ‘YAML::Node’
       if (YAML::Node &mapping = domain["custom_mapping"]) {
```
 (for the 2 mapping and 2 formats occurrences)

I am far from proficient in c++, so searching for this I found: https://stackoverflow.com/a/28156535 which states:

> YAML::Node is already a reference semantic type. There is no need to bind instances of YAML::Node to references. They are designed to be passed around by value, but behave like references.

I'll remove the `&` let me know if there is another way or something I am missing.

File: modules/geoipbackend/geoipbackend.cc [hunk 2]
```diff
@@ -91,6 +93,25 @@ static vector<std::unique_ptr<GeoIPInterface> > s_geoip_files;
 string getGeoForLua(const std::string& ip, int qaint);
 static string queryGeoIP(const Netmask& addr, GeoIPInterface::GeoIPQueryAttribute attribute, GeoIPNetmask& gl);
 
+// validateMappingLookupFormats validates any custom format provided by the
+// user does not use the custom mapping placeholder again, else it would do an
+// infinite recursion.
+bool validateMappingLookupFormats(const vector<string>& formats) {
+  string::size_type cur,last;
+  for (const auto& lformat : formats) {
```
jsoref, line 101: The [check-spelling](https://github.com/marketplace/actions/check-spelling) action doesn't like `lformat`, you could use `lookupFormat` or just `format`:
https://github.com/jsoref/pdns/commit/2170772e6e21da5af191065cc5e5f4dbe228c357#commitcomment-39229499

modules/geoipbackend/geoipbackend.cc
```diff
@@ -210,6 +244,23 @@ void GeoIPBackend::initialize() {
         nmt.insert(Netmask("::/0")).second.swap(value);
       }
 
+      // Allow per domain override of mapping_lookup_formats and custom_mapping.
+      // If not defined, the global values will be used.
+      if (YAML::Node formats = domain["mapping_lookup_formats"]) {
+        vector<string> mlf = formats.as<vector<string>>();
```
jsoref, line 250: Similarly, you could use `mapping` or `mapping_lookup_formats`

modules/geoipbackend/regression-tests/custom-mapping-txt-resolution/expected_result
```diff
@@ -0,0 +1,8 @@
+0	map.geo.example.com.	IN	TXT	30	"custom mapping"
+2	.	IN	OPT	0	AAgACAABIBgBAQEB
+Rcode: 0 (No Error), RD: 0, QR: 1, TC: 0, AA: 1, opcode: 0
+Reply to question for qname='map.geo.example.com.', qtype=TXT
+0	map.geo2.example.com.	IN	TXT	30	"overriden moon mapping"
```
jsoref, line 5: The check-spelling action found a typo:
```suggestion
0	map.geo2.example.com.	IN	TXT	30	"overridden moon mapping"
```

jeremy-clerc:
Hello, sorry for the delay, I was out for few days, I suppose the spelling mistakes are blocking the PR ? I'll have a look this week to send a patch. Thanks.

Habbie:
Hi Jeremy, we removed the spell checker from CI again. Of course we prefer well-spelled PRs so please do fix typos :)

modules/geoipbackend/geoipbackend.cc
jeremy-clerc: done

modules/geoipbackend/geoipbackend.cc
jeremy-clerc: It was for `mlf` right ? Done

modules/geoipbackend/regression-tests/custom-mapping-txt-resolution/expected_result
jeremy-clerc: Done

Habbie:
The build failure is around LUA support, which seems weird. Can you rebase this?

jeremy-clerc:
Done, tests look good now. Thanks for the hint.

docs/backends/geoip.rst
```diff
@@ -204,6 +222,16 @@ These placeholders disable caching for the record completely:
 :%ip4: Client IPv4 address
 :%ip6: Client IPv6 address
 
+Following placeholder allows custom mapping:
+
+:%mp: Use formats in ``mapping_lookup_formats`` and use user defined ``custom_mapping``
+
+.. versionadded:: 4.3.0
```
Habbie, line 229:
```suggestion
.. versionadded:: 4.4.0
```

docs/backends/geoip.rst
```diff
@@ -204,6 +222,16 @@ These placeholders disable caching for the record completely:
...
+
+  These placeholders have been added in version 4.3.0:
```
Habbie, line 231:
```suggestion
  These placeholders have been added in version 4.4.0:
```

Habbie (approved):
Please update the version number so that we can merge this, thanks!

jeremy-clerc:
Updated, and rebased (else unrelated tests would not pass), thank you.

PR #8616, opened 2019-12-10
Rec: more bulk test variations

omoerbeek:
### Short description

This runs more variations of the bulk tests on circleci with different thread/mthread settings.
Also, the test itself is repeated: once to fill the record cache and then twice to use the record cache. I do two runs with the hot cache to see timing variations.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code


omoerbeek:
Obsoleted.

PR #8622, opened 2019-12-12
Use utf8mb4 instead of deprecated utf8

Rawk:
### Short description
utf8 is an alias for utf8mb3 that is deprecated since MySQL 8.0, released 2018.
utf8mb4 is supported since MySQL 5.5.3, released 2010.

Deprecation info:
https://dev.mysql.com/doc/refman/8.0/en/charset-unicode-utf8.html

### Checklist
I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/modules/gmysqlbackend/schema.mysql.sql
+++ b/modules/gmysqlbackend/schema.mysql.sql
@@ -5,7 +5,7 @@ CREATE TABLE domains (
   last_check            INT DEFAULT NULL,
   type                  VARCHAR(6) NOT NULL,
   notified_serial       INT UNSIGNED DEFAULT NULL,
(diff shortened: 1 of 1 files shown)


mind04:
utf8 is not deprecated, it is an alias for utf8mb3. At some point this alias will become a reference for utf8mb4.

Rawk:
Well, you are right, it is utf8mb3 that is deprecated. But MySQL writes: "To avoid ambiguity about the meaning of utf8, consider specifying utf8mb4 explicitly for character set references instead of utf8." It will take many years until MySQL versions with fixed utf8 alias is in widespread use.

That said, i am totaly ok with rejecting this pull request.

Habbie:
I don't think this warrants a schema change for our users, so I'm closing this. Thanks!

Issue #8610, opened 2019-12-06
Combining 'delete' and 'replace' for the same RRSet in one request leads to duplicate RRSet error


similarbg:
I noticed unexpected behaviour in the PowerDNS v4.2.0 API, when pushing new contents for a zone.

Combining 'delete' and 'replace' for the same RRSet in one request leads to a duplicate RRSet error.
This issue was discussed on IRC with 'Habbie'.

A simple snippet to reproduce this error is:
https://gist.github.com/Habbie/edfafea62231b322ae5f8de96436a50b

Req.3 works, req.2 fails with the error: "ApiError (422): Duplicate RRset aaa1._domainkey.example.com. IN CNAME"

The reason I want to combine the 'delete' and the 'replace' rrsets in one request is to replace the complete contents of zone in an atomic way, without risking some rrsets being leftover from the old state.

james-stevens:
Creating duplicate records, especially CNAME records, is clearly an error, but I would have though the whole point of REPLACE is to give you an atomic operation for changing RRs, obviating the need for the DELETE, surley?

zeha:
Yeah, sending delete + replace for the same RRset is generally pointless.

similarbg:
My use case is a situation in which a newly pushed zone might be out of sync with the zone in PowerDNS. Specifically, when a CNAME exists in PowerDNS and an A record is pushed for the same name.
In this case, we need to request the current RRs from PowerDNS, do a smart local diff to determine what RRs should be replaced and what should be deleted (CNAME, because of conflicts with the A RRset) and then call the API with the results. This is not atomic and could lead to race conditions, in case the zone is changed in between these requests.

Ideally we could just push the complete new zone to the PowerDNS API and let PowerDNS atomically replace the old zone with the new zone data. 

This feature does not seem available (yet), so for now I try to simplify the procedure by making it possible to push deletes for all current RRs in PowerDNS and replaces for the new RRs for the zone, so the local diff can be reduced to a basic comparison.

zeha:
You could still just send REPLACE name/CNAME + REPLACE name/A?
I don't see why you need to send a DELETE for name/A and a REPLACE for name/A?

Maybe you can send which data you actually send and why?

james-stevens:
[...]

PR #8623, opened 2019-12-12
Make a combination of delete and replace for rrset possible

jonathaneen:
### Short description
This is an attempt to fix issue 8610, allowing for 'delete' and 'replace' rrsets to be combined in the same request, for the same record type. https://github.com/PowerDNS/pdns/issues/8610

### Checklist
I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/ws-auth.cc
+++ b/pdns/ws-auth.cc
@@ -1965,7 +1965,7 @@ static void patchZone(UeberBackend& B, HttpRequest* req, HttpResponse* resp) {
     di.backend->getDomainMetadataOne(zonename, "SOA-EDIT", soa_edit_kind);
     bool soa_edit_done = false;
 
-    set<pair<DNSName, QType>> seen;
+    set<tuple<DNSName, QType, string>> seen;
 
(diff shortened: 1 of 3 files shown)


similarbg:
I'm afraid I cannot link the build error on Travis to my changes in this PR. Please let me know if I'm expected to fix the build in some way.

omoerbeek:
Likely if you merge https://github.com/PowerDNS/pdns/pull/8632 into your branch you'll be fine.

zeha:
As mentioned in #8610 this is probably “technically correct”, but has an unclear use case: REPLACE already deletes the old records, so sending DELETE for the same RRset shouldn’t change the outcome at all.

zeha:
> As mentioned in #8610 this is probably “technically correct”, but has an unclear use case: REPLACE already deletes the old records, so sending DELETE for the same RRset shouldn’t change the outcome at all.

However: DELETE doesn't validate; so if you have bad zone data today, there is a difference. *shrug*

PR #8627, opened 2019-12-12
gpgsqlbackend: Avoid actually prepared statements

zeha:
### Short description

Remove real prepared statements from the generic PostgreSQL backend. Instead use the parameterised libpq API, which was (as far as I remember) the original intent.

This should help users with transaction level pooling servers (like pgbouncer) where relying on cross-transaction session state is not possible (and prepared statement are such state).

### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/modules/gpgsqlbackend/spgsql.cc
+++ b/modules/gpgsqlbackend/spgsql.cc
@@ -35,7 +35,7 @@
 class SPgSQLStatement: public SSqlStatement
 {
 public:
-  SPgSQLStatement(const string& query, bool dolog, int nparams, SPgSQL* db, unsigned int nstatement) {
+  SPgSQLStatement(const string& query, bool dolog, int nparams, SPgSQL* db) {
     d_query = query;
     d_dolog = dolog;
     d_parent = db;
@@ -45,7 +45,6 @@ public:
     d_res_set = NULL;
     paramValues = NULL;
     paramLengths = NULL;
-    d_nstatement = nstatement;
     d_paridx = 0;
     d_residx = 0;
     d_resnum = 0;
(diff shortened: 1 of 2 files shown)


zeha:
This should also remove d_nstatement.

sfrost:
This patch allows using pgbouncer with PowerDNS, which is very handy for dealing with spikes in zone transfers, for example, because PowerDNS currently creates a new connection to the database for every zone transfer.  This does, however, mean that each query is re-planned every time it's called instead of being prepared once and then just executed and therefore there's an increase in planning time (looks to be something like 0.1-0.2ms, but this will depend on the query and the system involved, of course).

I'd recommend adding a flag to make this configurable, have it default to the pre-patch behavior, and users who want to use pgbouncer with PowerDNS can change the flag to have the behavior introduced by this patch.

Habbie:
I like this patch. I'd love to not have a flag for it, but I don't see a clear choice between 'true' and 'false' for that flag that is the obvious right thing for everybody, so I think the only remaining problem is naming and describing the flag.

zeha:
Based on what other projects do with rgd. to PG and really my feelings I think we should avoid the flag and just do what I proposed here.
The 0.2ms planning cost is "just one number" and IMO doesn't tell us very much about the real world effect of this. TTBOMK the change to prepared statements wasn't (at least for PG) done based off hard numbers either.

If someone does a proper benchmark I'm happy to reconsider my position; otherwise I'm in strong favour of just calling PQexecParams.

Habbie:
> TTBOMK the change to prepared statements wasn't (at least for PG) done based off hard numbers either.

This is correct. We always wanted those numbers but nobody ever delivered them. All along there has, in fact, been this nagging feeling that the prepared statements might slow things down - we just don't know.

So.. is this ready for merge? :)

zeha:
> So.. is this ready for merge? :)

I've removed `d_nstatement`, and @sfrost has tested it (the previous version though). I'm all for it.

PR #8628, opened 2019-12-12
auth: make sure get() is always returning the default value for d_place

### LINKED ISSUE #8626: Response ends up on in ADDITIONAL SECTION instead of ANSWER SECTION (only with LMDB)
- Program Authoritative
- Issue type Bug report

### Short description
This issue causes answers to end up in the ADDITIONAL SECTION instead of ANSWER SECTION. Which is interpreted by at least some resolvers as an empty response (no record of that record type). The address is therefore seen as non-existent and no other nameservers will be tried. 

### Environment
I'm using PDNS 4.2.1 installed from the yum repo, with LMDB backend. I have query-cache enabled.

### Steps to reproduce

This is an example zone file with which I can reproduce this:
```
nameweb.it.		901	IN	SOA	ns.nscluster.eu. hostmaster.nscluster.eu. 2019121203 43200 3600 1209600 901
nameweb.it.		901	IN	A	127.0.0.1
nameweb.it.		901	IN	NS	ns.nscluster.eu.
nameweb.it.		901	IN	NS	ns.nscluster.fr.
nameweb.it.		901	IN	NS	ns.nscluster.se.
nameweb.it.		901	IN	NS	ns.nscluster.uk.
nameweb.it.		901	IN	NS	ns.nscluster.us.
nameweb.it.		901	IN	MX	10 mail.nameweb.it.
nameweb.it.		901	IN	MX	10 mail2.nameweb.it.
localhost.nameweb.it.	901	IN	A	127.0.0.1
mail.nameweb.it.	901	IN	A	127.0.0.1
mail2.nameweb.it.	901	IN	A	127.0.0.2
```

I restart PDNS to make sure the cache is empty and then run
```
dig @127.0.0.1 -t MX nameweb.it
sleep 1
dig @127.0.0.1 -t A mail.nameweb.it
dig @127.0.0.1 -t A mail2.nameweb.it
```

The answer to the last dig, for mail2.nameweb.it, ends up in the "ADDITIONAL SECTION" while it should have been in the ANSWER SECTION. I assume it was put in an incorrect way in the cache when it was returned in the additional section while requesting the MX-record.
```
dig @127.0.0.1 -t A mail2.nameweb.it

; <<>> DiG 9.9.4-RedHat-9.9.4-73.el7_6 <<>> @127.0.0.1 -t A mail2.nameweb.it
; (1 server found)
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 49950
;; flags: qr aa rd; QUERY: 1, ANSWER: 0, AUTHORITY: 0, ADDITIONAL: 2

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 1232
;; QUESTION SECTION:
;mail2.nameweb.it.		IN	A

;; ADDITIONAL SECTION:
mail2.nameweb.it.	901	IN	A	127.0.0.2

;; Query time: 0 msec
;; SERVER: 127.0.0.1#53(127.0.0.1)
;; WHEN: Thu Dec 12 14:44:39 CET 2019
;; MSG SIZE  rcvd: 61
```

Any following requests for mail2.nameweb.it within the query-cache-ttl will also return the result in the wrong section. 

I tried this also on PDNS with MySQL back-end, but couldn't reproduce it there. It seems to only be an issue when using LMDB.

I'm also unable to reproduce it if the additional section for the MX-record has only one entry.

### Other information
A temporary workaround is to set query-cache-ttl=0

If anybody wants a hand with reproducing this, do let me know and I'd be happy to help.

mind04:
### Short description
Make sure get() is always returning the default value for d_place and not some, random, previously set value.

Fixes #8626

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/ueberbackend.cc
+++ b/pdns/ueberbackend.cc
@@ -633,6 +633,9 @@ bool UeberBackend::get(DNSZoneRecord &rr)
     d_answers.clear();
     return false;
   }
+
+  rr.dr.d_place=DNSResourceRecord::ANSWER;
+
   d_ancount++;
   d_answers.push_back(rr);
   return true;


rgacogne:
Travis failure is unrelated (see https://github.com/PowerDNS/pdns/pull/8632).

PR #8647, opened 2019-12-20
rec: Bump max-qperq default to 100 

### LINKED ISSUE #8646: Default max-qperq too low for QNAME minimization?
- Program: Recursor
 - Issue type: Bug report

### Short description
The default `max-qperq` setting is 50. (And `max-recursion-depth` is 40, but I don't know what that does.) I'm concerned that may not be enough for QNAME minimization, DNSSEC validation, cold caches, and fun things like IPv6 reverse DNS.

An `ip6.arpa` name has, what, 34 labels. From a cold cache, to resolve a secure name that exists, that's 34 `A` queries, `34` DS queries, a half dozen `DNSKEY` queries... you're already over 70 queries without resolving any nameservers or sending a single actual `PTR` query.

A hot cache should run fine, but resolvers may `SERVFAIL` some queries when they're warming up.

Should it be increased to something like 100? Or more?

### Environment
 - Operating system: Ubuntu 16.04 x86-64
 - Software version: 4.3.0~beta1+master.19.gfa178d8fd-1pdns.xenial

### Steps to reproduce
1. `sudo systemctl restart pdns-recursor`
2. `dig -x 2620:57:4011:1452::62`

### Expected behaviour
`NXDOMAIN`.

### Actual behaviour
`SERVFAIL`.

Retrying it a minute later after the `SERVFAIL` cache clears works.

### Other information
My example IP is a random IP that doesn't even have a `PTR` record. To pick something more traditional, the (signed) root server IPs 2001:500:1::53, 2001:7fd::1 and 2001:500:9f::42 appear to be suitable.

omoerbeek:
This is enough for rev v6 queries with cold cache with validations and qname minimization enabled.  Fixes #8646.

These cases (with cold cache) might still take more than 5s though, so timeouts can still happen.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/pdns_recursor.cc
+++ b/pdns/pdns_recursor.cc
@@ -3970,7 +3970,7 @@ static int serviceMain(int argc, char*argv[])
   if (SyncRes::s_qnameminimization) {
     // With an empty cache, a rev ipv6 query with dnssec enabled takes
     // almost 100 queries. Default maxqperq is 60.
-    SyncRes::s_maxqperq = SyncRes::s_maxqperq * 5 / 3;
+    SyncRes::s_maxqperq = std::max(SyncRes::s_maxqperq, 100);
   }
 
   SyncRes::s_hardenNXD = SyncRes::HardenNXD::DNSSEC;
--- a/pdns/pdns_recursor.cc
+++ b/pdns/pdns_recursor.cc
@@ -3970,7 +3970,7 @@ static int serviceMain(int argc, char*argv[])
   if (SyncRes::s_qnameminimization) {
     // With an empty cache, a rev ipv6 query with dnssec enabled takes
     // almost 100 queries. Default maxqperq is 60.
-    SyncRes::s_maxqperq = std::max(SyncRes::s_maxqperq, 100);
+    SyncRes::s_maxqperq = std::max(SyncRes::s_maxqperq, static_cast<unsigned int>(100));
   }
 
   SyncRes::s_hardenNXD = SyncRes::HardenNXD::DNSSEC;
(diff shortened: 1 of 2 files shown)


rgacogne:
I wonder if we should only raise this value when qname-minimization is enabled? The existing value has been a nice safeguard for several years.

omoerbeek:
Thought about that as well, but did not go with that:

QName minimization is switched on by default. With QM switched off and a cold cache, for the v6 rev ip question 50 is already not enough; it should be at least 55.

Also, the argument processing does not allow for an easy way to say: if qname minimization is switched off and max-qperq is not set take this default and otherwise that.

So I went for fort a new default that's covers the rev v6 case with qm + validation.

Maybe set a new default to 60 and if qm is switched on add a bonus of 40?

pdns/pdns_recursor.cc
```diff
@@ -3967,6 +3967,12 @@ static int serviceMain(int argc, char*argv[])
 
   SyncRes::s_qnameminimization = ::arg().mustDo("qname-minimization");
 
+  if (SyncRes::s_qnameminimization) {
+    // With an empty cache, a rev ipv6 query with dnssec enabled takes
+    // almost 100 queries. Default maxqperq is 60.
+    SyncRes::s_maxqperq = SyncRes::s_maxqperq * 5 / 3;
```
rgacogne: I find this arithmetic a bit confusing, especially if the value of `max-qperq` has been updated from the default. Perhaps we could set it to `std::max(SyncRes::s_maxqperq, 100)` when qname minimization has been enabled instead?

pdns/pdns_recursor.cc
```diff
@@ -3967,6 +3967,12 @@ static int serviceMain(int argc, char*argv[])
 
   SyncRes::s_qnameminimization = ::arg().mustDo("qname-minimization");
 
+  if (SyncRes::s_qnameminimization) {
+    // With an empty cache, a rev ipv6 query with dnssec enabled takes
+    // almost 100 queries. Default maxqperq is 60.
+    SyncRes::s_maxqperq = std::max(SyncRes::s_maxqperq, 100);
```
rgacogne: I see this failed on CircleCI, we need to cast the constant to the same type than `s_maxqperq`:
```suggestion
    SyncRes::s_maxqperq = std::max(SyncRes::s_maxqperq, static_cast<unsigned int>(100));
```

PR #8665, opened 2020-01-02
rec: An Opt-Out NSEC3 RR only proves that there is no secure delegation

### LINKED ISSUE #8664: Recursor will set AD bit in NSEC3 opt-out negative responses
- Program: Recursor
 - Issue type: Bug report

### Short description
The Recursor will set the `AD` bit on `NXDOMAIN` responses, and `DS` `NODATA` responses for insecure delegations, in NSEC3 opt-out zones. (Excluding, of course, circumstances where it never sets the `AD` bit, like when DNSSEC is off.)

[RFC 5155 section 9.2](https://tools.ietf.org/html/rfc5155#section-9.2) says this MUST NOT be done:

>   The AD bit, as defined by [RFC4035], MUST NOT be set when returning a
>   response containing a closest (provable) encloser proof in which the
>   NSEC3 RR that covers the "next closer" name has the Opt-Out bit set.

### Environment
 - Operating system: Ubuntu 16.04 amd64
 - Software version: 4.3.0~beta1+master.25.g732b1dbd0-1pdns.xenial

### Steps to reproduce
1. Enable DNSSEC.
2. `dig +dnssec nsec3-oh-no.com`
3. `dig +dnssec google.com ds`

### Expected behaviour
No `AD` bit.

### Actual behaviour
`AD` bit.

### Other information
To be honest, I'm unsure about all of this. I would not be surprised if I'm wrong. On the other hand, Recursor's behavior is different from the other implementations I've checked.

Everybody sets `AD` on `NODATA` responses where the name is proven to exist, e.g. `dig +dnssec com` or `dig +dnssec cctld.us` (an ENT).

rgacogne:
### Short description

So we should not set the `AD` bit for a `DS` `NODATA` or a `NXDOMAIN` answer proved by an `Opt-Out` `NSEC3` `RR`.
 
Hopefully fixes #8664.

Could use regression tests!

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/pdns/syncres.cc
+++ b/pdns/syncres.cc
@@ -1417,7 +1417,7 @@ void SyncRes::computeNegCacheValidationStatus(const NegCache::NegCacheEntry* ne,
     vState neValidationState = ne->d_validationState;
     dState expectedState = res == RCode::NXDomain ? NXDOMAIN : NXQTYPE;
     dState denialState = getDenialValidationState(*ne, state, expectedState, false);
-    updateDenialValidationState(neValidationState, ne->d_name, state, denialState, expectedState, qtype == QType::DS);
+    updateDenialValidationState(neValidationState, ne->d_name, state, denialState, expectedState, qtype == QType::DS || expectedState == NXDOMAIN);
   }
   if (state != Indeterminate) {
     /* validation succeeded, let's update the cache entry so we don't have to validate again */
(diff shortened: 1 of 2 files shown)


omoerbeek:
The code does seem to do what the RFC says and the testcases are fixed. 
But I have to admit that I am not able to see if these changed might have unwanted side-effects.
I also would  love a unittest for the test cases.

PR #8666, opened 2020-01-02
Fix build on systems without Python 2.

jsha:
On systems with only Python 3 (e.g. Ubuntu Eoan), virtualenv still
defaults to building venvs that expect python2. This causes a build
failure:

```
virtualenv .venv
The path python2 (from --python=python2) does not exist
```

Here's the documentation from virtualenv --help:

> The Python interpreter to use, e.g.,
> --python=python2.5 will use the python2.5 interpreter
> to create the new environment.  The default is the
> python2 interpreter on your path (e.g.
> /usr/bin/python2)

This change adds an explicity `--python=python3` flag to virtualenv.
Since most modern systems have python3, believe this should work even
on those systems that also have python2.

As an FYI, Python2.7 is now unmaintained: https://pythonclock.org/

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -254,6 +254,7 @@ commands:
               libyaml-cpp-dev \
               make \
               pkg-config \
+              python3-venv \
               ragel \
               sqlite3 \
               systemd \
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -359,6 +359,7 @@ commands:
             libssl-dev \
             make \
             pkg-config \
+            python3-venv \
             ragel \
             rsync \
             virtualenv
@@ -867,6 +868,7 @@ jobs:
             libtool \
             make \
             pkg-config \
+            python3-venv \
             protobuf-compiler \
             ragel \
             systemd \
@@ -1123,6 +1125,7 @@ jobs:
             make \
             pkg-config \
             protobuf-compiler \
+            python3-venv \
             ragel \
             virtualenv
       - install-coverity-tools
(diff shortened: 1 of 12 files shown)


zeha:
There was some discussion on #powerdns about this.
Some relevant points:

* this virtualenv call is only relevant at `make dist` time, not (tarball) build time
* configure could avoid checking for virtualenv if building from a tarball (can't build docs from it anyway)
* each product has a copy of the virtualenv call, so they all need fixing
* maybe its smarter to use pyvenv (bundled with the interpreter) instead of virtualenv nowadays

jsha:
Thanks! That feedback makes sense. Updated to `python3 -m venv` and increased the number of files affected.

> configure could avoid checking for virtualenv if building from a tarball (can't build docs from it anyway)

If it's alright I'd like to leave this part for a later PR. I'm not particularly expert at autotools and I'm not sure exactly what I would need to tweak.

jsha:
I'm a little stuck on trying to get the Travis tests to pass. I've added this to build-scripts/travis.sh:

```
  python3 \
  python3.4-venv \
```

However, when this runs in Travis I get the error:

```
python3 -m venv .venv
pyenv: python3: command not found
The `python3' command exists in these Python versions:
  3.6
  3.6.3
```

(I've also tried a variant where I install `python3.6-venv`. IIRC that errors out because Trusty doesn't have that package).

python3 definitely should exist after the apt package is installed. I'm thinking maybe the `pyenv global 2.7` in the before_script of .travis.yml could be interfering? However, I wasn't able to reproduce locally. Any thoughts on how to fix?

Habbie:
>  I'm thinking maybe the `pyenv global 2.7` in the before_script of .travis.yml could be interfering?

Without having looked at anything, I wonder if we can't get rid of that? Py3 everywhere!

Habbie:
We are now finally py3 all the way on the master branch, so I'm closing this. Thanks!

PR #8668, opened 2020-01-06
ws-auth: Allow DNAME in apex with SOA and NS records

### LINKED ISSUE #8641: DNAME at apex via API leads to "Conflicts with other RRSet"
- Program: Authoritative
 - Issue type: Bug report

### Short description
Attempts to create a zone with a single DNAME via the API at the apex of the domain fail with the error: "Conflicts with another RRset"

### Environment
 - Operating system: Ubuntu (Bionic)
 - Software version: 4.2.1
 - Software source: deb [arch=amd64] http://repo.powerdns.com/ubuntu bionic-auth-42 main

### Steps to reproduce
1. Attempt to post `{"kind": "master", "name": "example.net.", "rrsets": [{"ttl": 3600, "name": "example.net.", "records": [{"content": "example.co.uk.", "disabled": false}], "type": "DNAME"}], "masters": [], "nameservers": []}` to the Zones API at `http://localhost:8081/api/v1/servers/localhost/zones`
2. Observe a reply `{"error": "RRset example.net. IN DNAME: Conflicts with another RRset"}`

### Expected behaviour
I'd expect the zone to be created, with a single DNAME at the apex of the domain.

### Actual behaviour
422 error from the API citing that it conflicts with another RRset.  No other RRset exists.

### Other information
I discussed this briefly on IRC with dwfreed - who assured me that DNAMEs could exist with other records.  I think, however, there's something odd going on with the apex of a domain here.

While I vaguely understand DNAMEs, I'm unsure if DNAME at the apex is permissible - seemingly not in PDNS as it stands, but I think it is in BIND (which one of my customers is moving away from).

cmouse:
### Short description
Permits DNAME with SOA and NS in zone apex. Closes #8641.

I have:
- [x] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/ws-auth.cc
+++ b/pdns/ws-auth.cc
@@ -59,7 +59,7 @@ static void makePtr(const DNSResourceRecord& rr, DNSResourceRecord* ptr);
 // QTypes that MUST NOT have multiple records of the same type in a given RRset.
 static const std::set<uint16_t> onlyOneEntryTypes = { QType::CNAME, QType::DNAME, QType::SOA };
 // QTypes that MUST NOT be used with any other QType on the same name.
-static const std::set<uint16_t> exclusiveEntryTypes = { QType::CNAME, QType::DNAME };
+static const std::set<uint16_t> exclusiveEntryTypes = { QType::CNAME };
 
 AuthWebServer::AuthWebServer() :
   d_tid(0),
@@ -2029,6 +2029,8 @@ static void patchZone(UeberBackend& B, HttpRequest* req, HttpResponse* resp) {
 
         if (replace_records) {
           bool ent_present = false;
+          bool dname_seen = false, ns_seen = false;
+
           di.backend->lookup(QType(QType::ANY), qname, di.id);
           DNSResourceRecord rr;
           while (di.backend->get(rr)) {
@@ -2037,6 +2039,10 @@ static void patchZone(UeberBackend& B, HttpRequest* req, HttpResponse* resp) {
               /* that's fine, we will override it */
               continue;
             }
+            if (qtype == QType::DNAME || rr.qtype == QType::DNAME)
+              dname_seen = true;
+            if (qtype == QType::NS || rr.qtype == QType::NS)
+              ns_seen = true;
             if (qtype.getCode() != rr.qtype.getCode()
               && (exclusiveEntryTypes.count(qtype.getCode()) != 0
                 || exclusiveEntryTypes.count(rr.qtype.getCode()) != 0)) {
@@ -2049,6 +2055,9 @@ static void patchZone(UeberBackend& B, HttpRequest* req, HttpResponse* resp) {
             }
           }
 
+          if (dname_seen && ns_seen && qname != zonename) {
+            throw ApiException("RRset "+qname.toString()+" IN "+qtype.getName()+": Cannot have both NS and DNAME except in zone apex");
+          }
           if (!new_records.empty() && ent_present) {
             QType qt_ent{0};
             if (!di.backend->replaceRRSet(di.id, qname, qt_ent, new_records)) {
--- a/regression-tests.api/test_Zones.py
+++ b/regression-tests.api/test_Zones.py
@@ -1298,7 +1298,6 @@ $ORIGIN %NAME%
 
     @parameterized.expand([
         ('CNAME', ),
-        ('DNAME', ),
     ])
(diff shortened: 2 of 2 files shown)


pdns/ws-auth.cc
```diff
@@ -1356,11 +1357,31 @@ static void checkNewRecords(vector<DNSResourceRecord>& records) {
...
       }
     }
 
+    // check apex records
+    if (apexRecords.count(QType::CNAME) > 0)
+       throw ApiException("Apex record has CNAME");
+    if (apexRecords.count(QType::DNAME) > 0 && apexRecords[QType::DNAME] > 1)
+       throw ApiException("Apex record has duplicate DNAME");
+    if (apexRecords.count(QType::SOA) > 0 && apexRecords[QType::SOA] > 1)
+       throw ApiException("Apex record has duplicate DNAME");
```
Habbie:
```suggestion
       throw ApiException("Apex record has duplicate SOA");
```

pdns/ws-auth.cc
```diff
@@ -1356,11 +1357,31 @@ static void checkNewRecords(vector<DNSResourceRecord>& records) {
...
+    if (apexRecords.count(QType::DNAME) > 0) {
+       // allow SOA and NS
+       for(auto it = apexRecords.cbegin(); it != apexRecords.cend(); it++)
+         if (it->first != QType::DNAME && it->first != QType::SOA && it->first != QType::NS)
+            throw ApiException("Apex with DNAME can only hae SOA and NS records");
```
Habbie: This is wrong. A, MX, TXT, etc. can all coexist with DNAME.

Habbie: And this is not an apex rule, this is a general rule.

regression-tests.api/test_Zones.py
```diff
@@ -1298,31 +1298,6 @@ def test_rrset_unknown_type(self):
 
     @parameterized.expand([
         ('CNAME', ),
-        ('DNAME', ),
-    ])
-    def test_rrset_exclusive_and_other(self, qtype):
```
zeha: this test is still valid for CNAMEs, no?

cjsoftuk:
What can I reasonably do to help push this forward?  Do you need me to test it in the specific case that triggered the original report?

Habbie:
> Do you need me to test it in the specific case that triggered the original report?

That never hurts!

cjsoftuk:
This seems to have fixed one of the issues that was preventing us from moving to PDNS - so thank you.  Given that it seems to be OK - is there anything else productive I can do to help this down the road?

Habbie:
You've done what you could! This should be reviewed and merged soon.

pdns/ws-auth.cc
```diff
@@ -2029,23 +2036,37 @@ static void patchZone(UeberBackend& B, HttpRequest* req, HttpResponse* resp) {
...
+          getAllRecords(di.backend, records);
+
+          for(auto& existing_rrset: records) {
+            bool dname_seen = false, ns_seen = false;
+            sort(existing_rrset.second.begin(), existing_rrset.second.end(),
+              [](const DNSResourceRecord& rec_a, const DNSResourceRecord& rec_b) -> bool {
+              /* we need _strict_ weak ordering */
+                return std::tie(rec_a.qname, rec_a.qtype, rec_a.content) < std::tie(rec_b.qname, rec_b.qtype, rec_b.content);
+              }
+            );
```
mind04: Please explain why you need the sorting. Lookup will only return records from one qname, and the code below is  iterating over all rr's.

pdns/ws-auth.cc
```diff
@@ -2029,23 +2036,37 @@ static void patchZone(UeberBackend& B, HttpRequest* req, HttpResponse* resp) {
...
+            for(const auto& rr: existing_rrset.second) {
```
mind04: Without the sorting a normal get() will do in this loop and will avoid some copying of rrs.

cmouse: We agreed to add the getAllRecords here so it might eventually maybe replace get().

pdns/ws-auth.cc
cmouse: I *suppose* we don't need to sort them for this.

cjsoftuk:
Hi guys - any news on this?  Sorry to keep chasing!

PR #8682, opened 2020-01-08
auth: Add an extended status report in the bind backend

rgacogne:
### Short description

I find very useful to be able to get much more information about the current status of the domain.

Sample output:
```
$ pdns_control BIND-DOMAIN-EXTENDED-STATUS
example.zone: 
         Status: parsed into memory at 2020-01-08 17:38:12 +0100
         Internal ID: 1
         On-disk file: /etc/powerdns/example.zone (1578495472)
         Kind: Slave
         Masters: 
                 - 192.0.2.1:53
         Also Notify: 
         Number of records: 42
         Loaded: 1
         Check now: 0
         Check interval: 0
         Last check: 1578501492
         Last notified: 0
```

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/docs/backends/bind.rst
+++ b/docs/backends/bind.rst
@@ -144,7 +144,16 @@ will be loaded at first request.
 .. note::
   This does not add the zone to the :ref:`setting-bind-config` file.
 
-``bind-domain-status <domain> [domain]``
+``bind-domain-extended-status [domain ...]``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. versionadded:: 4.3.0
+
+Output an extended status of a domain or domains, containing much more information than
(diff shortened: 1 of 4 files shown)


docs/backends/bind.rst
```diff
@@ -144,6 +144,15 @@ will be loaded at first request.
 .. note::
   This does not add the zone to the :ref:`setting-bind-config` file.
 
+``bind-domain-extended-status <domain> [domain]``
```
Habbie: Looks like you wrote this one before you made the arguments entirely optional.

rgacogne: Fixed, thanks!

rgacogne:
Rebased to fix a conflict, fixed the doc mistake at the same time.

PR #8702, opened 2020-01-14
Add the source and destination ports to the protobuf msg

rgacogne:
### Short description

- for all protobuf messages in dnsdist ;
- for incoming queries / outgoing responses in the recursor.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code


rgacogne:
> Looks good. If I see things correctly, protobuf receivers using the old definition will not be hindered by this, right?

They will just ignore the new fields, yes.

> If so, I like to include this in rec-4.3.0-beta2

Fine by me! :-)

PR #8705, opened 2020-01-15
rec: Fix precedence order for RPZ policies rules

rgacogne:
### Short description

- Since 272e9a0 we scanned all policies for an exact match before looking for wildcard matches. It brokes the promise that filtering policies are evaluated in the order they are defined ;
- After a "pass-thru" match, we should not process any other RPZ rule ;
- Export the filtering policy type to `Lua`.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/pdns/recursordist/test-filterpo_cc.cc
+++ b/pdns/recursordist/test-filterpo_cc.cc
@@ -431,10 +431,10 @@ BOOST_AUTO_TEST_CASE(test_multiple_filter_policies)
   const DNSName badWildcard("*.bad-wildcard.example.com.");
   const DNSName badUnderWildcard("sub.bad-wildcard.example.com.");
 
-  zone1->addQNameTrigger(bad, DNSFilterEngine::Policy(DNSFilterEngine::PolicyKind::Custom, DNSFilterEngine::PolicyType::QName, 0, nullptr, {DNSRecordContent::mastermake(QType::CNAME, QClass::IN, "garden1.example.net.")}));
-  zone2->addQNameTrigger(bad, DNSFilterEngine::Policy(DNSFilterEngine::PolicyKind::Custom, DNSFilterEngine::PolicyType::QName, 0, nullptr, {DNSRecordContent::mastermake(QType::CNAME, QClass::IN, "garden2.example.net.")}));
-  zone1->addQNameTrigger(badWildcard, DNSFilterEngine::Policy(DNSFilterEngine::PolicyKind::Custom, DNSFilterEngine::PolicyType::QName, 0, nullptr, {DNSRecordContent::mastermake(QType::CNAME, QClass::IN, "garden1.example.net.")}));
-  zone2->addQNameTrigger(badUnderWildcard, DNSFilterEngine::Policy(DNSFilterEngine::PolicyKind::Custom, DNSFilterEngine::PolicyType::QName, 0, nullptr, {DNSRecordContent::mastermake(QType::CNAME, QClass::IN, "garden2.example.net.")}));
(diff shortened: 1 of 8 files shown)


pdns/syncres.cc
```diff
@@ -3508,8 +3527,10 @@ int SyncRes::doResolveAt(NsSet &nameservers, DNSName auth, bool flawedNSSet, con
             }
           }
           LOG(endl);
-          if (hitPolicy) //implies d_wantsRPZ
-            return -2;
+          if (hitPolicy) { //implies d_wantsRPZ
+            /* RPZ hit */
+            throw PolicyHitException();
+          }
         }
```
omoerbeek: There are still a few places in the code (and comments) were return -2 is used for a RPZ hit. Since `doResolve()` now  throws an exception for all these cases, these lines of code and commens should be adapted. Specifically: 767, 889, 3428 (hope I got them all)

pdns/syncres.hh
```diff
@@ -1054,6 +1054,10 @@ public:
   string reason; //! Print this to tell the user what went wrong
 };
 
+class PolicyHitException
```
omoerbeek: Maybe extend from `PDNSException`?

pdns/recursordist/test-filterpo_cc.cc
```diff
@@ -428,9 +428,13 @@ BOOST_AUTO_TEST_CASE(test_multiple_filter_policies)
   zone2->setName("Unit test policy 1");
 
   const DNSName bad("bad.example.com.");
+  const DNSName badWildcard("*.bad-wildcard.example.com.");
+  const DNSName badUnderWildcard("sub.bad-wildcard.example.com.");
 
   zone1->addQNameTrigger(bad, DNSFilterEngine::Policy(DNSFilterEngine::PolicyKind::Custom, DNSFilterEngine::PolicyType::QName, 0, nullptr, {DNSRecordContent::mastermake(QType::CNAME, QClass::IN, "garden1.example.net.")}));
   zone2->addQNameTrigger(bad, DNSFilterEngine::Policy(DNSFilterEngine::PolicyKind::Custom, DNSFilterEngine::PolicyType::QName, 0, nullptr, {DNSRecordContent::mastermake(QType::CNAME, QClass::IN, "garden2.example.net.")}));
+  zone1->addQNameTrigger(badWildcard, DNSFilterEngine::Policy(DNSFilterEngine::PolicyKind::Custom, DNSFilterEngine::PolicyType::QName, 0, nullptr, {DNSRecordContent::mastermake(QType::CNAME, QClass::IN, "garden1.example.net.")}));
+  zone2->addQNameTrigger(badUnderWildcard, DNSFilterEngine::Policy(DNSFilterEngine::PolicyKind::Custom, DNSFilterEngine::PolicyType::QName, 0, nullptr, {DNSRecordContent::mastermake(QType::CNAME, QClass::IN, "garden2.example.net.")}));
```
omoerbeek: Suggestio to use `"garden1a.example.net."` etc to distinguish the 4 cases and test against the specific names below, So not only test if the correct policy is hit, but also the correct rule.
[x.txt](https://github.com/PowerDNS/pdns/files/4076362/x.txt)

pdns/syncres.cc
rgacogne: Done, thanks!

pdns/recursordist/test-filterpo_cc.cc
rgacogne: Nice idea, applied!

pdns/syncres.hh
rgacogne: I'm not sure it's a good idea, the new exception might get caught in unexpected places, which should not happen with a new, unrelated exception.

pdns/syncres.cc
rgacogne: Done, thanks!

pdns/syncres.hh
omoerbeek: Right

PR #8707, opened 2020-01-15
lmdb: fill di.serial (code stolen straight from gsqlbackend)

Habbie:
### Short description
This changes the serial from `0` to the correct value in a domain-specific API call such as `/api/v1/servers/localhost/zones/example.com.`.

`/api/v1/servers/localhost/zones` yields the right serial without this patch, and I see that `void LMDBBackend::getAllDomains(vector<DomainInfo> *domains, bool include_disabled)` uses more 'direct' code to get the serial. PRing this as draft until I figure out which code is right.

### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/modules/lmdbbackend/lmdbbackend.cc
+++ b/modules/lmdbbackend/lmdbbackend.cc
@@ -811,6 +811,25 @@ bool LMDBBackend::getDomainInfo(const DNSName &domain, DomainInfo &di, bool getS
   if(!(di.id=txn.get<0>(domain, di)))
     return false;
   di.backend = this;
+
+  di.serial = 0;
+
+  if(getSerial) {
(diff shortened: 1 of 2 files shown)


Habbie:
@mind04 points out that the version in `getAllDomains` is a lot faster. Perhaps we should move that one into a private method so we can reuse it.

Habbie:
Also, I broke a test.

peterthomassen:
@Habbie Do you think this will make it into an upcoming 4.2.x release, or at least auth 4.3.0? There's not much more my heart could wish for :-p

Habbie:
Hopefully all good now, with some hints from @mind04 :)

mind04 (approved):

PR #8713, opened 2020-01-17
auth: Enforce a strict maximum size for the packet and records caches

rgacogne:
### Short description

Before this change, both the query and packet caches in the authoritative server can exceed their maximum size by a lot, until the next cleaning cycle.
This is particularly nasty since the current cleaning algorithm will never remove entries from the cache until they expire, as opposed to what we do in the recursor, for example, where we nuke the least-recently used entries, even if they are still valid, when the cache is full.
This commit changes that by removing the least recently inserted or updated entry from the cache after inserting a new one when the cache is full, thus enforcing the maximum size more strictly.

Note that this is really the least recently inserted/updated and not the least recently used one, as is done in the recursor. Having a proper `LRU` in the auth would require acquiring a write lock for a
simple lookup, instead of a potentially concurrent read-lock at the moment. We might want to consider changing that at some point, as a `LRU` might be fairer and the lock contention might be very small since the caches are sharded.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/pdns/auth-packetcache.cc
+++ b/pdns/auth-packetcache.cc
@@ -154,6 +154,7 @@ void AuthPacketCache::insert(DNSPacket& q, DNSPacket& r, unsigned int maxTTL)
         continue;
       }
 
+      moveCacheItemToBack<SequencedTag>(mc.d_map, iter);
       iter->value = entry.value;
       iter->ttd = now + ourttl;
       iter->created = now;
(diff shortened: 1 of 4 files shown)


pdns/auth-packetcache.cc
```diff
@@ -162,7 +163,15 @@ void AuthPacketCache::insert(DNSPacket& q, DNSPacket& r, unsigned int maxTTL)
 
     /* no existing entry found to refresh */
     mc.d_map.insert(entry);
-    (*d_statnumentries)++;
+
+    if (*d_statnumentries >= d_maxEntries) {
+      /* remove the least recently inserted or replaced entry */
+      auto& sidx = mc.d_map.get<SequencedTag>();
```
Habbie: Question unrelated to this change: why do we actively manage `d_statnumentries`? Does this map not expose a cheap `.size()` ?

rgacogne: We have several shards, so we would need to iterate over all of them, taking a lock in the process, to know the total number of entries.

Habbie: Ah of course, thanks.

PR #8714, opened 2020-01-17
auth: Reduce the number of temporary memory allocations

rgacogne:
### Short description

This PR <strike>is a work-in-progress to reduce</strike> reduces the number of temporary memory allocations and copies in the authoritative server. Comments welcome!

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/packethandler.cc
+++ b/pdns/packethandler.cc
@@ -125,7 +125,7 @@ bool PacketHandler::addCDNSKEY(DNSPacket& p, std::unique_ptr<DNSPacket>& r, cons
     rr.dr.d_name=p.qdomain;
     rr.dr.d_content=std::make_shared<DNSKEYRecordContent>(value.first.getDNSKEY());
     rr.auth=true;
-    r->addRecord(rr);
+    r->addRecord(std::move(rr));
     haveOne=true;
   }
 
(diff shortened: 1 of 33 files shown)


omoerbeek:
#8370 is somewhat related. I just rebased it.

rgacogne:
Rebased to fix conflicts.

Habbie:
> This PR is a work-in-progress to reduce the number of temporary memory allocations and copies in the authoritative server. Comments welcome!

It was marked 'ready for review', I trust it is no longer work-in-progress?

rgacogne:
> It was marked 'ready for review', I trust it is no longer work-in-progress?

Yes, I should have added a comment! I have a few remaining ideas but I can always open a separate PR later :)

Habbie:
> Yes, I should have added a comment! I have a few remaining ideas but I can always open a separate PR later :)

I just edit the original comment in such cases. Thanks!

pdns/packethandler.cc
```diff
@@ -944,18 +946,21 @@ int PacketHandler::processNotify(const DNSPacket& p)
 static bool validDNSName(const DNSName &name)
 {
```
Habbie: with how much this function now assumes about how `DNSName` works, perhaps we should move it into the class at some point? Doesn't have to be in this PR, just a thought :)

rgacogne: Yes, that would make a lot of sense!

rgacogne:
Rebased. I'll move the `validDNSName` fonction into the `DNSName` class, but I don't mind if that happens outside of this merge request.

rgacogne:
>  I'll move the validDNSName fonction into the DNSName class, but I don't mind if that happens outside of this merge request.

Done!

rgacogne:
Rebased to fix the `PG sql` tests issue (see #8833).

PR #8733, opened 2020-01-21
dnsdist: Load an openssl configuration file, if any, during startup

rgacogne:
### Short description

This way dnsdist will load the default OpenSSL configuration, or a custom one specified via the `OPENSSL_CONF` environment variable.
It allows loading an engine or configuration various options supported by OpenSSL.
This requires OpenSSL >= 1.1.0.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/pdns/dnsdistdist/libssl.cc
+++ b/pdns/dnsdistdist/libssl.cc
@@ -19,7 +19,7 @@
 #include <sodium.h>
 #endif /* HAVE_LIBSODIUM */
 
-#if (OPENSSL_VERSION_NUMBER < 0x1010000fL || defined LIBRESSL_VERSION_NUMBER)
+#if (OPENSSL_VERSION_NUMBER < 0x1010000fL || (defined LIBRESSL_VERSION_NUMBER) && LIBRESSL_VERSION_NUMBER < 0x2090100fL)
 /* OpenSSL < 1.1.0 needs support for threading/locking in the calling application. */
 static pthread_mutex_t *openssllocks{nullptr};
 
@@ -62,7 +62,7 @@ static void openssl_thread_cleanup()
   OPENSSL_free(openssllocks);
 }
 
-#endif /* (OPENSSL_VERSION_NUMBER < 0x1010000fL || defined LIBRESSL_VERSION_NUMBER) */
+#endif /* (OPENSSL_VERSION_NUMBER < 0x1010000fL || (defined LIBRESSL_VERSION_NUMBER) && LIBRESSL_VERSION_NUMBER < 0x2090100fL) */
 
 static std::atomic<uint64_t> s_users;
 static int s_ticketsKeyIndex{-1};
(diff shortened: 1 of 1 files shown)


pdns/dnsdistdist/libssl.cc
```diff
@@ -72,7 +72,11 @@ static int s_keyLogIndex{-1};
 void registerOpenSSLUser()
 {
   if (s_users.fetch_add(1) == 0) {
-#if (OPENSSL_VERSION_NUMBER < 0x1010000fL || defined LIBRESSL_VERSION_NUMBER)
+#if (OPENSSL_VERSION_NUMBER >= 0x1010000fL && !defined LIBRESSL_VERSION_NUMBER)
```
omoerbeek: `OPENSSL_init_crypto` has been available in libressl for a few years already...

omoerbeek: specifically the test can be:
`if (OPENSSL_VERSION_NUMBER >= 0x1010000fL && (!defined LIBRESSL_VERSION_NUMBER) || LIBRESSL_VERSION_NUMBER >= 0x2070000fL)`

rgacogne: Does this version of LibreSSL really support automatic initialization and, more importantly, automatic setup of thread-specific callbacks? Otherwise this will result in random memory corruption..
I really wish LibreSSL had not fiddled with `OPENSSL_VERSION_NUMBER`...

omoerbeek: `CRYPTO_set_id_callback` and `CRYPTO_set_locking_callback` are not needed in libressl anymore since `LIBRESSL_VERSION_NUMBER >= 0x2090100fL`. So to be conservative, we could check for this value.

rgacogne: Thanks! I pushed a commit updating these version number checks, I'd be great if you could check I did not mess anything up :)

pieterlexis:
Should the `#if` perhaps be centralized and do a `#define PDNS_OPENSSL_RECENT_ENOUGH` but a better name?

rgacogne:
> Should the #if perhaps be centralized and do a #define PDNS_OPENSSL_RECENT_ENOUGH but a better name?

I don't really see what you mean by that, we have several checks referring to different versions of OpenSSL and LibreSSL, with only one of thel being used twice?

omoerbeek:
Maybe, but not all version nummbers are the same, though I also could live with "libressl mus be new enough" test.

Additionally, `SSL_CTX_set_min_proto_version` is available as well with libressl >= 2.7, but `SSL_CTX_get0_privatekey`is not at all.

omoerbeek:
Current version of PR compiles and links fine on OpenBSD.

rgacogne:
> Additionally, `SSL_CTX_set_min_proto_version` is available as well with libressl >= 2.7, but `SSL_CTX_get0_privatekey` is not at all.

I guess we could try to properly detect if these functions are present during `configure`, but it's a bit out-of-scope for this PR that I would like to be able to merge quickly.

PR #8706, opened 2020-01-15
Include LuaContext.hpp for LuaContext::ExecutionErrorException

swegener:
```
test-lua_auth4_cc.cc: In member function ‘void lua_auth4_cc::test_prequery::test_method()’:
test-lua_auth4_cc.cc:37:53: error: expected unqualified-id before ‘&’ token
   37 |   } catch (const LuaContext::ExecutionErrorException& e) {
      |                                                     ^
```

### Checklist

I have:
- [ ] read the [CONTRIBUTING.md](https://github.com/PowerDNS/pdns/blob/master/CONTRIBUTING.md) document
- [ ] compiled this code
- [ ] tested this code
- [ ] included documentation (including possible behaviour changes)
- [ ] documented the code
- [ ] added or modified regression test(s)
- [ ] added or modified unit test(s)
- [ ]  checked that this code was merged to master

PR #8744, opened 2020-01-25
Make Lua mandatory for Auth

zeha:
### Short description
Recursor and dnsdist already treat Lua as mandatory; with the growing Lua-based feature set in the Auth, it would make sense to also treat Lua as mandatory. At times in the past we've discovered that nobody really tests building without Lua anymore.

Also see #8706.

### Checklist

I have:
- [x] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -178,9 +178,11 @@ commands:
           apt-get install -qq -y --no-install-recommends \
             libboost-all-dev \
             libcdb1 \
+            libcurl4 \
             libkrb5-3 \
             libldap-2.4-2 \
             liblmdb0 \
+            libluajit-5.1-2 \
             libpq5 \
             libssl1.1 \
             libsodium23 \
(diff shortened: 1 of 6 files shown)


Habbie:
Turns out our CircleCI build is without Lua.

Habbie:
> Turns out our CircleCI build is without Lua.

I'll work on this.

Habbie:
> I'll work on this.

Commit pushed to your branch!

zeha:
Thanks!

james-stevens:
LUA is not RFC, so should not be mandatory :(

zeha:
You can still pass `--disable-lua-records` to get rid of LUA, but not of Lua. Building without Lua just makes everybody's life harder.

james-stevens:
We strongly avoid installing development tools on production machines.
PDNS requires PDNS server + MySql Server, nothing else ... until now, 
now we also have to install Lua programming language runtime, even though we don't want it or use it :(

Couldn't you just provide a stub-Lua that has blank calls corresponding to the ones you use?

Or will PDNS *run* without Lua, just not *compile* without it?
i.e. does `--disable-lua-record` block PDNS trying to do **any** Lua, or will it still try some kind of Lua initialization?

Habbie:
Hello James,

you are replying on a closed ticket; often, such comments go entirely unnoticed. Normally I'd ask for the request to be posted in a new ticket, but in this case that does not make sense.

If you do not configure pdns to do anything with Lua, there should be no way that any Lua code gets executed.

I'm closing conversation on this ticket.

PR #8750, opened 2020-01-28
docs: clarify Zone object description, closes #8748

### LINKED ISSUE #8748: auth API (doc issue): Retrieve all zone names and serials (edited_serial) without zone contents
- Program: Authoritative
 - Issue type: Feature request

### Short description
We would like to retrieve a list of all zone names and corresponding serials (edited_serial) from the pdns API, without retrieving all the record contents.

### Usecase
To extract the state of all zones, one needs all zone names and serials. This helps, for example, with the implementation of catalog zones (enriched with serials).

### Description
One way of doing this would be a query parameter to the `zones/` endpoint, like `rrsets=false` or similar. When given, the objects in the returned JSON array would not contain the `rrsets` field (and ideally, pdns would also save all the associated database queries).

A more generic way would be passing `fields` query parameter. In my usecase, I would then issue `GET .../zones/?fields=name,edited_serial`.

peterthomassen:
Point out that the `rrsets` is only included at the zone detail endpoint

### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/docs/http-api/swagger/authoritative-api-swagger.yaml
+++ b/docs/http-api/swagger/authoritative-api-swagger.yaml
@@ -918,7 +918,7 @@ definitions:
         type: array
         items:
           $ref: '#/definitions/RRSet'
-        description: 'RRSets in this zone'
+        description: 'RRSets in this zone (for zones/{zone_id} endpoint only; omitted during GET on the .../zones list endpoint)'
       serial:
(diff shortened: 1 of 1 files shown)


Habbie:
Thanks!

PR #8756, opened 2020-01-28
rec: Document the difference between rec_control quit and quit-nicely.

### LINKED ISSUE #2267: Please clarify difference between "quit" and "quit-nicely"
I think it is very uncommon to have two ways to end a daemon (especially when you have a command named `quit-nicely` which sounds better than just "quit" but which isn't used in the runscripts in source). So please clarify the difference between `_quit_` and `_quit-nicely_` which can be called using `rec_control`.

From my understanding `quit` calls `_exit()` and `quit-nicely` calls `exit()` ... but when should we use which command?

If we want to keep two ways to end pdns/pdns_recursor we should add documentation and give recommendation to package maintainers which one they should use for normal shutdown in their init system.

omoerbeek:
Fixes #2267

### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/recursordist/docs/manpages/rec_control.1.rst
+++ b/pdns/recursordist/docs/manpages/rec_control.1.rst
@@ -197,7 +197,7 @@ ping
 
 quit
     Request shutdown of the recursor, exiting the process while
-    letting the OS clean up recources.
+    letting the OS clean up resources.
 
 quit-nicely
     Request nice shutdown of the recursor. This method allows all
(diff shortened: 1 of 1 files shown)


pdns/recursordist/docs/manpages/rec_control.1.rst
```diff
@@ -196,10 +196,13 @@ ping
     Check if server is alive.
 
 quit
-    Request shutdown of the recursor.
+    Request shutdown of the recursor, exiting the process while
+    letting the OS clean up recources.
```
mnordhoff:
```suggestion
    letting the OS clean up resources.
```

PR #8766, opened 2020-01-30
Update my name in old changelogs

zeha:
Updates my name in old changelogs, documentation, etc.

I guess not putting this MR into the changelogs would make sense ;)


Habbie:
> I guess not putting this MR into the changelogs would make sense ;)

We can use Squash&merge to avoid it even showing up on the radar for that :)

PR #8783, opened 2020-02-05
Add 'IO wait' and 'steal' metrics on Linux

rgacogne:
### Short description

The 'steal' one is useful when investigating latency issues in virtualized environment. The 'IO wait' one is a bit more of a long shot although it might make sense when investigating certain issues with the auth so why not.
I did not actually test the 'steal' metric yet.

### Checklist

I have:
- [x] compiled this code
- [ ] tested this code
- [ ] documented the code


pdns/common_startup.cc
```diff
@@ -346,6 +346,12 @@ void declareStats(void)
 
   S.declare("sys-msec", "Number of msec spent in system time", getSysUserTimeMsec);
   S.declare("user-msec", "Number of msec spent in user time", getSysUserTimeMsec);
+
+#ifdef __linux__
+  S.declare("cpu-iowait", "Time spent waiting for I/O to complete by the whole system", getCPUIOWait);
+  S.declare("cpu-steal", "Stolen time, which is the time spent by the whole system in other operating systems when running in a virtualized environment", getCPUSteal);
+#endif
+
```
omoerbeek: Would be nice to know the unit. From the man page: 
"The amount of time, measured in units of USER_HZ
(1/100ths of a second on most architectures, use
sysconf(_SC_CLK_TCK) to obtain the right value)"

rgacogne: Done!

PR #8792, opened 2020-02-07
DNSName: Don't call strlen() when the length is already known

rgacogne:
### Short description

Saving a call to `strlen()` when we already know the size, i.e. when the `DNSName` is constructed from a string. This is often the case when we retrieved the name from a `SQL` backend, for example.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


rgacogne:
Rebased to fix a conflict.

PR #8811, opened 2020-02-11
adjust NSEC TTLs to negative TTL

Habbie:
### Short description
So that RFC8198 aggressive NSEC use does not extend negative TTLs when this is not wanted.

~Needs update notes, some doc changes,~ perhaps a blog post?

~Need to check if this makes us do the wrong thing for NSEC(3) TTLs on presigned/slaved zones from other signers (including pdns 4.2).~

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


Habbie:
> Need to check if this makes us do the wrong thing for NSEC(3) TTLs on presigned/slaved zones from other signers (including pdns 4.2).

Using the lower TTL on those is fine. It will help caches do the right thing, and it will not affect signatures anyway.

Issue #2214, opened 2015-02-10
Auth: MySQL database schema allows duplicate entries in domainmetadata table


jpmens:
Release 3.4.2

The database schema for MySQL's `domainmetadata` table allows rows with duplicate content, the effects of which are unknown to me. I think the index definition

```
CREATE INDEX domainmetadata_idx ON domainmetadata (domain_id, kind);
```

should encompass `content` and be unique.

```
mysql: CREATE DATABASE j1;
mysql: SOURCE /usr/share/doc/pdns/schema.mysql.sql;
mysql: INSERT INTO domainmetadata (domain_id, kind, content) VALUES (13, 'SOA-EDIT-API', 'INCEPTION-INCREMENT');
mysql: INSERT INTO domainmetadata (domain_id, kind, content) VALUES (13, 'SOA-EDIT-API', 'INCEPTION-INCREMENT');
mysql: SELECT * FROM domainmetadata;
+----+-----------+--------------+---------------------+
| id | domain_id | kind         | content             |
+----+-----------+--------------+---------------------+
|  1 |        13 | SOA-EDIT-API | INCEPTION-INCREMENT |
|  2 |        13 | SOA-EDIT-API | INCEPTION-INCREMENT |
+----+-----------+--------------+---------------------+
2 rows in set (0.00 sec)
```

Habbie:
Several metadata keys (TSIG-ALLOW-AXFR, FORWARD-DNSUPDATE, TSIG-ALLOW-DNSUPDATE, ALLOW-AXFR-FROM, possibly others) are currently allowed to exist multiple times. However, I can't think of a reason for any of them to exists multiple times with the same content, so I don't have an immediate objection to your suggestion.

As for what your example means - most code that expects a single metadata entry picks the first or last one (terms that are meaningless in an SQL context, of course - so it's best to say they pick an arbitrary one).

Your suggestion will not fix the situation where somebody manages to put SOA-EDIT=X, SOA-EDIT=Y in the database. I do not have a better suggestion in any case!

zeha:
check-zone could possibly check this.

RobinGeuze:
I've done some research.

SQLite and MySQL both support a REPLACE like statement. Additionally MySQL supports "ON DUPLICATE KEY UPDATE'

postgresql > 9.5 has "upsert" support.

With oracle it can be achieved using a MERGE INTO statememt.

[...]

PR #8813, opened 2020-02-11
Make check-zone also check whether there are duplicate key value pair metadatas for the zone

RobinGeuze:
### Short description
This PR is the initial work for fixing #2214, initially returning errors in check-zone for the offending values.

### Checklist

I have:
- [x] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/pdnsutil.cc
+++ b/pdns/pdnsutil.cc
@@ -713,6 +713,27 @@ static int checkZone(DNSSECKeeper &dk, UeberBackend &B, const DNSName& zone, con
     }
   }
 
+  std::map<std::string, std::vector<std::string>> metadatas;
+  if (!B.getAllDomainMetadata(zone, metadatas)) {
+    cout << "[Error] Unable to retrieve metadata for zone " << zone << endl;
+    numerrors++;
+  }
+
+  for (const auto &metaData : metadatas) {
+    set<string> seen;
+    set<string> messaged;
+
+    for (const auto &value : metaData.second) {
+      if (seen.count(value) <= 0) {
+        seen.insert(value);
+      } else if (messaged.count(value) <= 0) {
+        cout << "[Error] Found duplicate metadata key value pair for zone " << zone << " with key " << metaData.first << " and value " << value << endl;
+        numerrors++;
+        messaged.insert(value);
+      }
+    }
+  }
+
   cout<<"Checked "<<records.size()<<" records of '"<<zone<<"', "<<numerrors<<" errors, "<<numwarnings<<" warnings."<<endl;
   if(!numerrors)
     return EXIT_SUCCESS;
--- a/pdns/pdnsutil.cc
+++ b/pdns/pdnsutil.cc
@@ -727,7 +727,7 @@ static int checkZone(DNSSECKeeper &dk, UeberBackend &B, const DNSName& zone, con
       if (seen.count(value) <= 0) {
         seen.insert(value);
       } else if (messaged.count(value) <= 0) {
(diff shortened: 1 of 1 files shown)


pdns/pdnsutil.cc
```diff
@@ -598,6 +598,27 @@ int checkZone(DNSSECKeeper &dk, UeberBackend &B, const DNSName& zone, const vect
     }
   }
 
+  std::map<std::string, std::vector<std::string>> metadatas;
+  if (!B.getAllDomainMetadata(zone, metadatas)) {
```
Habbie: For terrible reasons, this skips `PRESIGNED`, `NSEC3PARAM`, and `NSEC3NARROW`for gsqlbackend (but not for other backends) - but I don't think we can blame this PR for that!

pdns/pdnsutil.cc
```diff
@@ -598,6 +598,27 @@ int checkZone(DNSSECKeeper &dk, UeberBackend &B, const DNSName& zone, const vect
...
+
+  for (const auto &metaData : metadatas) {
+    set<string> seen;
+    set<string> messaged;
+
+    for (const auto &value : metaData.second) {
+      if (seen.count(value) <= 0) {
+        seen.insert(value);
+      } else if (messaged.count(value) <= 0) {
+        cout << "[Error] Found duplicate metadata key value pair for zone " << zone << " with key " << metaData.first << " and value " << value << endl;
```
Habbie: `[Error] Found duplicate metadata key value pair for zone 1.example.com with key PUBLISH-CDS and value `


```suggestion
        cout << "[Error] Found duplicate metadata key value pair for zone " << zone << " with key '" << metaData.first << "' and value '" << value << "'" << endl;
```

pdns/pdnsutil.cc
```diff
@@ -598,6 +598,27 @@ int checkZone(DNSSECKeeper &dk, UeberBackend &B, const DNSName& zone, const vect
...
+    cout << "[Error] Unable to retrieve metadata for zone " << zone << endl;
+    numerrors++;
+  }
+
+  for (const auto &metaData : metadatas) {
+    set<string> seen;
+    set<string> messaged;
+
+    for (const auto &value : metaData.second) {
+      if (seen.count(value) <= 0) {
```
Habbie: while not wrong, the `<` looks weird to me :)

Habbie:
ping!

pdns/pdnsutil.cc
RobinGeuze: How best to address the issue?

pdns/pdnsutil.cc
RobinGeuze: Right, looking at the docs for count it can only be 0 or 1, so I will change it.

pdns/pdnsutil.cc
RobinGeuze: So based on the code it seems that the only case where they are not included is if dnssec is disabled on the backend. Do we really care about the rare case where dnssec is disabled and there are DNSSEC params?

RobinGeuze:
@Habbie could you take another look? There is one review comment open, however its covers a corner case we can probably ignore in my opinion, do you agree?

pdns/pdnsutil.cc
Habbie: I've decided we don't :)

Habbie:
I agree. Can you rebase?

RobinGeuze:
Done.

Habbie (approved):

PR #8815, opened 2020-02-12
YaHTTP: Support bracketed IPv6 addresses

pieterlexis:
### Short description
This ensures that urls like `http://[2001::db8::fff:22]/` and `http://[2001::db8::fff:22]:2500/`. It also now sends a `Host` header with bracketed IPv6 addresses when needed.

cc: @cmouse 

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/ext/yahttp/yahttp/reqresp.hpp
+++ b/ext/yahttp/yahttp/reqresp.hpp
@@ -249,7 +249,7 @@ public:
     }
     void setup(const std::string& method_, const std::string& url_) {
       this->url.parse(url_);
-      this->headers["host"] = this->url.host;
+      this->headers["host"] = this->url.host.find(":") == std::string::npos ? this->url.host : "[" + this->url.host + "]";
       this->method = method_;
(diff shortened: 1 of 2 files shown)


ext/yahttp/yahttp/reqresp.hpp
```diff
@@ -249,7 +249,7 @@ namespace YaHTTP {
     }
     void setup(const std::string& method_, const std::string& url_) {
       this->url.parse(url_);
-      this->headers["host"] = this->url.host;
+      this->headers["host"] = this->url.host.find(":") == std::string::npos ? this->url.host : "[" + this->url.host + "]";
```
Habbie: This would break if `:9100` for a port number ever made it into the host header (which is legal).

pieterlexis: At the moment, YaHTTP doesn't add the port anyway,  that is stored in `url.port`.

PR #8852, opened 2020-02-21
Add SpoofRecordAction

cmouse:
### Short description
Constructs spoofed response for qtype by parsing content. Somewhat easier to use than SpoofRawAction.

Example:
```
addAction("test.unit.", SpoofRecordAction(15, "10 mx.text.unit"))
```

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


rgacogne:
Thank you for this PR! Unfortunately, we decided a while ago that we really did not want to have DNS record parsing in dnsdist, this is simply not where we want to go with this tool. It adds a lot of code and complexity and we believe that should not be done in a DNS load-balancer.

PR #8854, opened 2020-02-24
auth LUA records: make sure weights are positive non-zero numbers

Habbie:
### Short description
Coverity reported that we let zero/negative numbers through. This PR prevents that.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/lua-record.cc
+++ b/pdns/lua-record.cc
@@ -40,6 +40,15 @@ using wiplist_t = std::unordered_map<int, string>;
 using ipunitlist_t = vector<pair<int, iplist_t> >;
 using opts_t = std::unordered_map<string,string>;
 
+unsigned int pdns_stou_nonzero(const std::string& str, size_t * idx = 0, int base = 10)
+{
+  unsigned long result = pdns_stou(str, idx, base);
+  if (result == 0) {
+    throw std::out_of_range("number cannot be zero");
+  }
+  return result;
+}
+
 class IsUpOracle
 {
 private:
@@ -508,7 +517,7 @@ static vector<pair<int, ComboAddress> > convWIplist(std::unordered_map<int, wipl
   vector<pair<int,ComboAddress> > ret;
 
   for(const auto& i : src) {
-    ret.emplace_back(atoi(i.second.at(1).c_str()), ComboAddress(i.second.at(2)));
+    ret.emplace_back(pdns_stou_nonzero(i.second.at(1).c_str()), ComboAddress(i.second.at(2)));
   }
 
   return ret;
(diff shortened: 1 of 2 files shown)


pdns/lua-record.cc
```diff
@@ -508,7 +517,7 @@ static vector<pair<int, ComboAddress> > convWIplist(std::unordered_map<int, wipl
   vector<pair<int,ComboAddress> > ret;
 
   for(const auto& i : src) {
-    ret.emplace_back(atoi(i.second.at(1).c_str()), ComboAddress(i.second.at(2)));
+    ret.emplace_back(pdns_stou_nonzero(i.second.at(1).c_str()), ComboAddress(i.second.at(2)));
```
cmouse: shouldn't the vector pair be unsigned int too?

Habbie: I ignored that at first because people shouldn't be using weights that care about that, but it would be safer, I agree. Will fix.

pdns/lua-record.cc
```diff
@@ -40,6 +40,15 @@ using wiplist_t = std::unordered_map<int, string>;
 using ipunitlist_t = vector<pair<int, iplist_t> >;
 using opts_t = std::unordered_map<string,string>;
 
+unsigned int pdns_stou_nonzero(const std::string& str, size_t * idx = 0, int base = 10)
```
chbruyand: Maybe this could be useful to leave it in `misc.cc` ?

Habbie: I was unsure. If you also feel that way, let's do it :)

regression-tests.auth-py/test_LuaRecords.py
```diff
@@ -55,6 +55,7 @@ class TestLuaRecords(AuthTest):
 all.noneup.ifportup          3600 IN LUA  A     "ifportup(8080, {{'192.168.42.21', '192.168.21.42'}}, {{ backupSelector='all' }})"
 
 whashed.example.org.         3600 IN LUA  A     "pickwhashed({{ {{15, '1.2.3.4'}}, {{42, '4.3.2.1'}} }})"
+whashedzero.example.org.     3600 IN LUA  A     "pickwhashed({{ {{15, '1.2.3.4'}}, {{0, '4.3.2.1'}} }})"
```
jsoref, line 58: You'll need to add `whashedzero` to `expect.txt` (alternatively you could exclude `^regression-tests\.auth-py/test_LuaRecords\.py$` to `exclude.txt`) 
https://github.com/jsoref/pdns/commit/f08ce8ad14a3070ddf5893ae267289fae0912779#commitcomment-39229497

Habbie:
Rebased, will wait for CI and then do the discussed fixes.

pdns/lua-record.cc
cmouse: is there a reason we cannot use boost::lexical_cast instead? do we really need non-base10?

Habbie: Ah, makes sense. Of course it's weird we're parsing at all - it was a number before we pulled it out of Lua as a string!

miodvallat:
Sorry for the PR grave digging.

Is there still an interest in this feature? Do you need help to bring the PR up-to-date?

PR #8448, opened 2019-10-22
PROXYv2 work

Habbie:
### Short description
This is just a quick start. I'm pretty sure `sdig` does the right thing on UDP and TCP here.

References:
* https://github.com/dovecot/core/blob/master/src/lib-master/master-service-haproxy.c (LGPL v2.1)
* https://github.com/haproxy/haproxy/blob/master/include/types/connection.h (LGPL v2.1)
* https://www.haproxy.org/download/1.9/doc/proxy-protocol.txt

TODO:
* [ ] write tests for the two functions in proxy-protocol.cc
* [ ] rewrite those functions so they are less horrible
* [ ] implement PROXY receive support in recursor
* [ ] implement PROXY receive support in Auth
* [ ] implement PROXY receive support in dnsdist
* [ ] implement PROXY send support in dnsdist

### Checklist

I have:
- [ ] read the [CONTRIBUTING.md](https://github.com/PowerDNS/pdns/blob/master/CONTRIBUTING.md) document
- [ ] compiled this code
- [ ] tested this code
- [ ] included documentation (including possible behaviour changes)
- [ ] documented the code
- [ ] added or modified regression test(s)
- [ ] added or modified unit test(s)
- [ ]  checked that this code was merged to master

PR #8874, opened 2020-02-27
Add support for Proxy Protocol between dnsdist and the recursor

rgacogne:
### Short description

This PR implements:
- outgoing Proxy Protocol (UDP, TCP, healthchecks) in dnsdist ;
- incoming Proxy Protocol (UDP, TCP) in the recursor.

It allows dnsdist to pass the initial source and destination addresses and ports to the recursor, along with custom values.
Compared to the (ab)use of EDNS Client Subnet, this doesn't prevent preserving an existing ECS value (zeo scope, notably) and provides the ports, along with custom values.
Compared to ECS and XPF, this prevents having to parse the existing payload and inject either an EDNS option or a new record, preserving TSIG-signatures.

Based on a rebased version of https://github.com/PowerDNS/pdns/pull/8448.

Could use a good review :)

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code


pdns/pdns_recursor.cc
```diff
@@ -1993,11 +1998,92 @@ static void getQNameAndSubnet(const std::string& question, DNSName* dnsname, uin
...
+    }
+    else if (remaining < 0) {
+      conn->proxyProtocolNeed = -remaining;
+      conn->data.resize(conn->proxyProtocolGot + conn->proxyProtocolNeed);
+      return;
+    }
+    else {
+      /* proxy header received */
+      /* we ignore the TCP field for now, but we could properly set whether
+         the connection was received over UDP or TCP if neede */
```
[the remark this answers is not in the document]
rgacogne: Fixed, thanks!

rgacogne:
I realized that the current logic of trying to reuse TCP backend connections is completely flawed when the proxy protocol is enabled. I'm working on a fix.

rgacogne:
It's in fact more complicated than that because of the custom values. Currently they can be set per-query which works for UDP but doesn't match the fact that they can only be sent once per backend connection over TCP. We could disable connection reuse when the proxy protocol is enabled, but that would be quite bad for performances. On the other hand, if we keep it, we at least need to ensure that the same backend connection can only be used for the same incoming client connection which is a bit better for performances but cumbersome.

Habbie:
> I realized that the current logic of trying to reuse TCP backend connections is completely flawed when the proxy protocol is enabled. I'm working on a fix.

I've been wondering if the protocol should come with some framing, but that's another rabbit hole.

rgacogne:
Fixed by disabling TCP connection reuse in dnsdist when a backend has the proxy protocol feature enabled. We could try to still reuse the connection for the same client, as long as the TLV values are empty, but it's error-prone since in that case we should not add the proxy protocol payload for subsequent queries, except if the connection breaks at some point and we need to open a new one.

rgacogne:
> I'm very excited about this, but surprised to see the asymmetry of dnsdist being able to send PROXY messages but not receive them. Do you think it could also get settings like the recursor's proxy-protocol-from and proxy-protocol-maximum-size in this PR?

Yes, we are planning on adding support for incoming proxy protocol payload! Unfortunately our time is finite so we had to make some choices. This PR will not extend beyond outgoing proxy protocol, and I'm not sure yet whether we will be able to implement incoming support before 1.5.0, but it will definitely be there in 1.6.0 at the latest.

rgacogne:
Rebased on master to fix a conflict with #8851.

rgacogne:
@omoerbeek I requested a review from you mostly for the recursor-related bits, although your review on the rest of the code would be nice to have as well :-)

pdns/proxy-protocol.cc
```diff
@@ -0,0 +1,250 @@
...
+
+  size_t valuesSize = 0;
+  for (const auto& value : values) {
+    if (value.content.size() > std::numeric_limits<uint16_t>::max()) {
+      throw std::runtime_error("The size of proxy protocol values is limited to " + std::to_string(std::numeric_limits<uint16_t>::max()) + ", trying to add a value of size " + std::to_string(value.content.size()));
+    }
+    valuesSize += sizeof(uint8_t) + sizeof(uint8_t) * 2 + value.content.size();
+  }
+
+  const uint16_t contentlen = htons((addrSize * 2) + sizeof(sourcePort) + sizeof(destinationPort) + valuesSize);
```
omoerbeek:
```suggestion
  size_t total = (addrSize * 2) + sizeof(sourcePort) + sizeof(destinationPort) + valuesSize;
  if (total > std::numeric_limits<uint16_t>::max()) {
    throw(...);
  }
  const uint16_t contentlen = total;
```

pdns/recursordist/docs/lua-scripting/dq.rst
```diff
@@ -193,6 +193,16 @@ The DNSQuestion object contains at least the following fields:
 
       Returns the :class:`DNSHeader` of the query or nil.
 
+  .. method:: DNSQuestion:getPolicyTags() -> {str}
```
omoerbeek: Now we have two entries describind `getPolicyTags`.... The order is indeed confusing. I think I re-ordered them in my routing tag PR.

pdns/dnsdistdist/docs/reference/dq.rst
```diff
@@ -204,6 +204,14 @@ This state can be modified from the various hooks.
     :param int expire: The value of the expire field in the SOA record
     :param int minimum: The value of the minimum field in the SOA record
 
+  .. method:: DNSQuestion:setProxyProtocolValues(values)
+
+    .. versionadded:: 1.5.0
+
+    Set the Proxy-Protocol Type-Length values to send to the backend along with this query.
+
+    :param table values: A table of types and values to send, for example: ``{ ["0"] = foo", ["42"] = "bar" }``
```
omoerbeek: Do the indexes need to be strings? The C++ code expects `uint8_t`s.

pdns/dnsdistdist/docs/rules-actions.rst
```diff
@@ -1251,6 +1251,14 @@ The following actions exist.
   * ``ad``: bool - Set the AD bit to this value (true means the bit is set, false means it's cleared). Default is to clear it.
   * ``ra``: bool - Set the RA bit to this value (true means the bit is set, false means it's cleared). Default is to copy the value of the RD bit from the incoming query.
 
+.. function:: SetProxyProtocolValuesAction(values)
+
+  .. versionadded:: 1.5.0
+
+  Set the Proxy-Protocol Type-Length values to be sent to the server along with this query to ``values``.
+
+  :param table values: A table of types and values to send, for example: ``{ ["0"] = foo", ["42"] = "bar" }``
```
omoerbeek: Same here

pdns/recursordist/docs/lua-scripting/hooks.rst
```diff
@@ -53,13 +53,18 @@ Interception Functions
...
-.. function:: gettag(remote, ednssubnet, localip, qname, qtype, ednsoptions, tcp) -> int
+.. function:: gettag(remote, ednssubnet, localip, qname, qtype, ednsoptions, tcp, proxyprotocolvalues) -> int
+              gettag(remote, ednssubnet, localip, qname, qtype, ednsoptions, tcp) -> int
               gettag(remote, ednssubnet, localip, qname, qtype, ednsoptions) -> int
 
     .. versionchanged:: 4.1.0
 
       The ``tcp`` parameter was added.
 
+    .. versionchanged:: 4.3.0
```
omoerbeek:
```suggestion
    .. versionchanged:: 4.4.0
```

pdns/dnsdistdist/docs/reference/dq.rst
rgacogne: If you can point me to a better working Lua syntax, I'm all for it!

pdns/pdns_recursor.cc
```diff
@@ -2021,17 +2108,28 @@ static void handleRunningTCPQuestion(int fd, FDMultiplexer::funcparam_t& var)
       conn->bytesread=0;
     }
     if(!bytes || bytes < 0) {
```
omoerbeek:
```suggestion
    if(bytes <= 0) {
```

pdns/pdns_recursor.cc
```diff
@@ -2008,11 +2094,12 @@ static void handleRunningTCPQuestion(int fd, FDMultiplexer::funcparam_t& var)
       conn->state=TCPConnection::GETQUESTION;
     }
     if(!bytes || bytes < 0) {
```
omoerbeek:
```suggestion
    if (bytes <= 0) {
```

pdns/pdns_recursor.cc
```diff
@@ -1993,11 +1998,92 @@ static void getQNameAndSubnet(const std::string& question, DNSName* dnsname, uin
...
+      else if (static_cast<size_t>(used) > g_proxyProtocolMaximumSize) {
+        if (g_logCommonErrors) {
+          g_log<<Logger::Error<<"Proxy protocol header in packet from TCP client "<< conn->d_remote.toStringWithPort() << " is larger than proxy-protocol-maximum-size (" << used << "), dropping"<< endl;
+        }
+        ++g_stats.proxyProtocolInvalidCount;
+        t_fdm->removeReadFD(fd);
+        return;
+      }
+
+      /* check the real source */
```
omoerbeek: "real source" is confusing. With my test here I see that in the case of proxied queries the proxy source is used to check against `allow-from`. This should be documented more explicitly as well.

pdns/sdig.cc
```diff
@@ -32,7 +33,8 @@ void usage()
   cerr << "sdig" << endl;
   cerr << "Syntax: sdig IP-ADDRESS-OR-DOH-URL PORT QNAME QTYPE "
           "[dnssec] [ednssubnet SUBNET/MASK] [hidesoadetails] [hidettl] "
-          "[recurse] [showflags] [tcp] [xpf XPFDATA] [class CLASSNUM]"
+          "[recurse] [showflags] [tcp] [xpf XPFDATA] [class CLASSNUM] "
+          "[proxy PROXYDATA]"
```
omoerbeek: make it explicit what PROXYDATA is

pdns/recursordist/docs/settings.rst
```diff
@@ -1226,6 +1226,28 @@ Improves performance on Linux.
...
+.. _setting-proxy-protocol-from:
+
+``proxy-protocol-from``
+-----------------------
+.. versionadded:: 4.4.0
+
+-  IP ranges, separated by commas
+-  Default: empty
+
+Ranges that are allowed to send a Proxy Protocol header in front of UDP and TCP queries, to pass the original source and destination addresses and ports to the recursor, as well as custom values.
```
omoerbeek: "Allowed to" or "must"? From my testing it looks that regular queries are not accepted if the query comes from an address in `proxy-protocol-from`. This matches what I see in the code.

pdns/dnsdistdist/docs/reference/dq.rst
omoerbeek: Doesn't lua accept ``{ [0] = foo", [42] = "bar" }``?

pdns/recursordist/docs/lua-scripting/dq.rst
rgacogne: I'm pretty sure I messed up while rebasing, sorry!

pdns/dnsdistdist/docs/reference/dq.rst
rgacogne: It does!

rgacogne:
I have fixed most of the reported shortcomings.

> It should be documented how the proxy protocol and allow-from interact. AFAKS, if a request is proxied, the from address in the proxy header is used to check against allow-from.

Done!

> I don't know if the recursor packer parsing is subject to being fuzzed, but it would be nice, especially with these additions.

I'm planning on adding a fuzzing target for the proxy protocol parser itself, but if you are thinking of the TCP handling code it's a bit complicated because it's event-based, with a state machine. 

> A check for interoperability with another implementation would really be nice.

Agreed, I'll try to do that.

pdns/recursordist/docs/settings.rst
```diff
@@ -1235,7 +1237,10 @@ Whether to compute the latency of responses in protobuf messages using the times
 -  IP ranges, separated by commas
 -  Default: empty
 
-Ranges that are allowed to send a Proxy Protocol header in front of UDP and TCP queries, to pass the original source and destination addresses and ports to the recursor, as well as custom values.
+Ranges that are required to send a Proxy Protocol header in front of UDP and TCP queries, to pass the original source and destination addresses and ports to the recursor, as well as custom values.
+Queries that are not prefix with such a header will not be accepted from clients in these ranges. Queries prefixed by headers from clients that are not listed in these ranges will be dropped.
```
omoerbeek:
```suggestion
Queries that are not prefixed with such a header will not be accepted from clients in these ranges. Queries prefixed by headers from clients that are not listed in these ranges will be dropped.
```

pdns/recursordist/docs/settings.rst
```diff
@@ -1235,7 +1237,10 @@ Whether to compute the latency of responses in protobuf messages using the times
...
+
+Note that once a Proxy Protocol header has been received, the original source address will be checked against the `allow-from`_ ACL, instead of the address of the proxy.
```
omoerbeek:
```suggestion
Note that once a Proxy Protocol header has been received, the source address from the proxy header instead of the address of the proxy will be checked against the `allow-from`_ ACL, 
```

omoerbeek:
The test-recursor-bulk tets failure is due to a CirlceCI issue, making the CircleCI config and the checked out tree out-of-sync. Rebasing should fix it.

rgacogne:
I successfully tested the recursor's support for incoming proxy protocol over TCP against HAProxy 2.1.3, as well as dnsdist's outgoing support over TCP against nginx 1.16.1.

Unfortunately HAProxy doesn't do UDP. nginx doesn't support incoming proxy protocol for UDP, and only supports the first version of the proxy protocol for outgoing connections/datagrams.

rgacogne:
I couldn't find any implementation to test incoming or outgoing support over UDP, Cloudflare's Spectrum is supposed to support it but it requires an enterprise plan.
So unless someone disagrees, I'll rebase this PR to fix the regression test issue and then merge it as it is.

Habbie:
Do it!

omoerbeek:
> I couldn't find any implementation to test incoming or outgoing support over UDP, Cloudflare's Spectrum is supposed to support it but it requires an enterprise plan.
> So unless someone disagrees, I'll rebase this PR to fix the regression test issue and then merge it as it is.

Did you manage to do a TCP test? That is at least something...

rgacogne:
> Did you manage to do a TCP test? That is at least something...

I did, see https://github.com/PowerDNS/pdns/pull/8874#issuecomment-600004028

omoerbeek:
> > Did you manage to do a TCP test? That is at least something...
> 
> I did, see [#8874 (comment)](https://github.com/PowerDNS/pdns/pull/8874#issuecomment-600004028)

Great! I missed that comment earlier. LGTM!

PR #8882, opened 2020-03-02
dnsdist: add sessionTimeout setting for TLS session lifetime

qvr:
### Short description
Adds a new `sessionTimeout` setting for TLS session lifetime. It is used both for sessions stored server-side and also to indicate to clients for how long their given TLS ticket is valid for. If unset or set to 0, will use OpenSSL's default (which for TLS protocols is 2 hours).

It makes sense to not tell the client to throw the ticket away if the ticket key hasn't yet been rotated.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdist-lua.cc
+++ b/pdns/dnsdist-lua.cc
@@ -183,6 +183,10 @@ static void parseTLSConfig(TLSConfig& config, const std::string& context, boost:
     config.d_preferServerCiphers = boost::get<bool>((*vars)["preferServerCiphers"]);
   }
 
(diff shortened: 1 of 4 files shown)


qvr:
Ok, the default is now left up to OpenSSL, and I changed the setting name as well because the timeout is also used for sessions stored server-side (and that's a good reason to decouple it from ticket key rotation too).

And the default value for all TLS protocols is 2 hours:
https://github.com/openssl/openssl/blob/master/ssl/t1_lib.c#L106

PR #7489, opened 2019-02-15
Recursor: allow basic auth for some statistics APIs

zeha:
### Short description
Removes need for API key on builtin webserver pages.

Fixes #5942.

### Checklist
I have:
- [x] read the [CONTRIBUTING.md](https://github.com/PowerDNS/pdns/blob/master/CONTRIBUTING.md) document
- [x] compiled this code
- [x] tested this code
- [ ] included documentation (including possible behaviour changes)
- [ ] documented the code
- [ ] added or modified regression test(s)
- [x] added or modified unit test(s)

PR #8890, opened 2020-03-03
Recursor: allow access to some API endpoints with webserver-password even if api-key is empty

swoga:
### Short description
Since PR #7489 access to some API endpoints is allowed with the webserver-password (or no auth if the webserver-password is empty) instead of the api-key, but an api-key must still be configured or the requests will be denied.

This PR lifts this requirment and allows access to these endpoints even when the api-key is empty.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/webserver.cc
+++ b/pdns/webserver.cc
@@ -130,12 +130,11 @@ void WebServer::apiWrapper(WebServer::HandlerFunction handler, HttpRequest* req,
 
   resp->headers["access-control-allow-origin"] = "*";
 
-  if (d_apikey.empty()) {
-    g_log<<Logger::Error<<req->logprefix<<"HTTP API Request \"" << req->url.path << "\": Authentication failed, API Key missing in config" << endl;
-    throw HttpUnauthorizedException("X-API-Key");
-  }
+  bool auth_ok = false;
 
-  bool auth_ok = req->compareHeader("x-api-key", d_apikey) || req->getvars["api-key"] == d_apikey;
+  if (!d_apikey.empty()) {
+    auth_ok = req->compareHeader("x-api-key", d_apikey) || req->getvars["api-key"] == d_apikey;
+  }
 
   if (!auth_ok && allowPassword) {
     if (!d_webserverPassword.empty()) {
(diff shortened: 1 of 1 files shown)


swoga:
@rgacogne would you mind reviewing and merging this? :)

pdns/webserver.cc
```diff
@@ -130,12 +130,11 @@ void WebServer::apiWrapper(WebServer::HandlerFunction handler, HttpRequest* req,
...
-  if (d_apikey.empty()) {
-    g_log<<Logger::Error<<req->logprefix<<"HTTP API Request \"" << req->url.path << "\": Authentication failed, API Key missing in config" << endl;
-    throw HttpUnauthorizedException("X-API-Key");
-  }
+  bool auth_ok = false;
 
-  bool auth_ok = req->compareHeader("x-api-key", d_apikey) || req->getvars["api-key"] == d_apikey;
+  if (!d_apikey.empty()) {
+    auth_ok = req->compareHeader("x-api-key", d_apikey) || req->getvars["api-key"] == d_apikey;
+  }
```
omoerbeek, line 138: The logic is now: if the api-key is passed in but does not match, password auth is still tried if allowed. I think that can be confusing.

pdns/webserver.cc
```diff
@@ -145,6 +144,11 @@ void WebServer::apiWrapper(WebServer::HandlerFunction handler, HttpRequest* req,
     }
   }
 
+  if (!auth_ok && d_apikey.empty()) {
+    g_log<<Logger::Error<<req->logprefix<<"HTTP API Request \"" << req->url.path << "\": Authentication failed, API Key missing in config" << endl;
+    throw HttpUnauthorizedException("X-API-Key");
+  }
+
   if (!auth_ok) {
     g_log<<Logger::Error<<req->logprefix<<"HTTP Request \"" << req->url.path << "\": Authentication by API Key failed" << endl;
```
omoerbeek, line 153: A message talking about API Key failure is also given on password failure. Return code is also not relevant.

omoerbeek (changes requested):
I think this is only acceptable if the code flow of the two types of authentication are more separated and now intertwined as they are now, this PR makes that a bit worse even.

PR #8894, opened 2020-03-04
Rec: shared cache and routing tag

omoerbeek:
### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [X] documented the code

This introduces  shared record cache for all recursor threads.

Additionally, a possibility to add a "routingTag" (final name to be decided) to a query, settable form Lua code. This tag is associated with RC entries generated from the client query. If a routing tag is present, the EDNS subnet mask is ignored.
If the RC is consulter with tag and no records are found, the code queries the RC without tag.

Updates to the RC are stored with a tag, if the original query has a tag. That is likely a problem, since e.g. NS records that are retrieved as a result are tagged as well. This is different when using subnet, since the reply indicates if the auth actually did use a subnet to determine the answer. We have to discuss possible solutions to this. This is why I submit this as a draft PR.

IMO the shared cache part (without the tagging,https://github.com/PowerDNS/pdns/commit/a264db6bd2fbf06596a7c182ac9670b0b4490960 ) is solid and probably is ready to be merged. Maybe a separate PR fo that part?


rgacogne:
> Updates to the RC are stored with a tag, if the original query has a tag. That is likely a problem, since e.g. NS records that are retrieved as a result are tagged as well. This is different when using subnet, since the reply indicates if the auth actually did use a subnet to determine the answer. We have to discuss possible solutions to this. This is why I submit this as a draft PR.

I'm not sure I understand this correctly, aren't we still sending an ECS value to authoritative servers? If so we should still be able to detect whether they used it in the returned scope, and then store them with a routing tag if the ECS scope was != 0 and without a tag otherwise?

omoerbeek:
Or course, i must have developed a blind sport for that. To be continued. Need a new PR for this anyway once the shared cache part is merged.

omoerbeek:
See #8910

PR #8896, opened 2020-03-04
Add a new endpoint to the api to find existing subzones for a zone

jonathaneen:
### Short description
This PR adds a new endpoint /zones/<id>/sub-zones, which provides a fast search for the names and zone ids for all existing subdomain zones for a selected zone.
This should resolve issue PowerDNS#8841

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/modules/remotebackend/remotebackend.cc
+++ b/modules/remotebackend/remotebackend.cc
@@ -933,12 +933,12 @@ bool RemoteBackend::searchComments(const string &pattern, int maxResults, vector
   return false;
 }
 
-bool RemoteBackend::getSubZones(const string &pattern, vector<std::tuple<string, string>>& result)
+bool RemoteBackend::getSubZones(const string &zoneName, vector<std::tuple<string, string>>& result)
 {
     Json query = Json::object{
     { "method", "getSubZones" },
     { "parameters", Json::object{
-      { "pattern", pattern }
+      { "zoneName", zoneName }
     }}
   };
 
--- a/modules/remotebackend/remotebackend.cc
+++ b/modules/remotebackend/remotebackend.cc
@@ -933,12 +933,12 @@ bool RemoteBackend::searchComments(const string &pattern, int maxResults, vector
   return false;
 }
 
-bool RemoteBackend::getSubZones(const string &zoneName, vector<std::tuple<string, string>>& result)
+bool RemoteBackend::listSubDomains(const DNSName &parent_zone, vector<std::tuple<uint32_t, string>>& result)
 {
     Json query = Json::object{
     { "method", "getSubZones" },
     { "parameters", Json::object{
-      { "zoneName", zoneName }
+      { "parent_zone", parent_zone.toStringNoDot() }
     }}
   };
 
@@ -950,7 +950,7 @@ bool RemoteBackend::getSubZones(const string &zoneName, vector<std::tuple<string
     return false;
   
   for(const auto& row: answer["result"].array_items()) {
-    result.push_back(std::tuple<string, string>{stringFromJson(row, "zone_id"), stringFromJson(row, "name")});
+    result.push_back(std::tuple<uint32_t, string>{intFromJson(row, "zone_id"), stringFromJson(row, "name")});
   }
   return true;
 }
--- a/modules/remotebackend/remotebackend.cc
+++ b/modules/remotebackend/remotebackend.cc
@@ -933,6 +933,28 @@ bool RemoteBackend::searchComments(const string &pattern, int maxResults, vector
   return false;
 }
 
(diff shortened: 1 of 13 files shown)


modules/remotebackend/remotebackend.cc
```diff
@@ -922,6 +922,28 @@ bool RemoteBackend::searchComments(const string &pattern, int maxResults, vector
   return false;
 }
 
+bool RemoteBackend::getSubZones(const string &pattern, vector<std::tuple<string, string>>& result)
+{
+    Json query = Json::object{
+    { "method", "getSubZones" },
+    { "parameters", Json::object{
+      { "pattern", pattern }
```
zeha: `pattern` is a parent zone name here, right?
if so, a different name would be a lot better.

jonathaneen:
@swoga I'm no expert in PowerDNS API structure, so I'm open for suggestions.
In the issue related to this PR @zeha suggested adding the endpoint `../zones/<id>/parent-zones` in case I needed parent zones, that is the basis for the current `../zones/<id>/sub-zones` format.

pdns/backends/gsql/gsqlbackend.cc
```diff
@@ -1836,6 +1870,11 @@ void GSQLBackend::extractComment(const SSqlStatement::row_t& row, Comment& comme
   comment.content = row[5];
 }
 
+void GSQLBackend::extractZoneReference(const SSqlStatement::row_t& row, std::tuple<string, string>& reference)
```
zeha: i guess this is here because `extractRecord` and `extractComment` exist? they are called from multiple places, but the one line of code here could live inline in `getSubZones`.

pdns/backends/gsql/gsqlbackend.cc
```diff
@@ -1796,6 +1798,38 @@ bool GSQLBackend::searchComments(const string &pattern, int maxResults, vector<C
   return false;
 }
 
+bool GSQLBackend::getSubZones(const string &pattern, vector<std::tuple<string, string>>& result)
```
zeha: the internal api uses "Domain" instead of "Zone", and (while there are some outliers), "list" instead of "get" if multiple results are expected.

if pattern is a domain/zone name, i think it should be called "parent_zone" and be of type DNSName, as should the returned domain names. the first part of the vector appears to be the domain_id, which should be a uint32_t not a string?

pdns/ws-auth.cc
```diff
@@ -2250,6 +2250,28 @@ static void apiServerSearchData(HttpRequest* req, HttpResponse* resp) {
...
+  DNSName zonename = apiZoneIdToName(req->parameters["id"]);
+  Json::array doc;
+  vector<std::tuple<string, string>> subZoneResult;
+  UeberBackend B;
+  B.getSubZones(zonename.toStringNoDot(), subZoneResult);
+  for(const std::tuple<string, string> &z: subZoneResult)
+  {
+    auto object = Json::object {
+        { "object_type", "subzone" },
+        { "zone_id", std::get<0>(z) },
```
zeha: is the integer id useful here -- maybe you wanted `apiZoneNameToId` on the zone name instead?

if this endpoint returned the same type as `apiServerZones`, one wouldn't need to introduce a new object_type (and the documentation for it).

zeha:
Some more thoughts:

- it could be reasoned that the result should only include "direct" subzones (so, excluding zones that are subzones of a subzone). The current implementation would provide all direct and indirect subzones, even those without a delegation in their respective parent, or where an intermediate zone is hosted at another nameserver. I'd say this should be spelled out in the documentation, so there are no surprises.
- some users might want the results ordered in some particular way (flattened tree with closest child first for example). Right now the result is unordered -- this should probably also be in the doc.
- remotebackend needs a doc update to say what is expected of implementors (unordered is fine, all results, etc)
- bindbackend might want an implementation too

Habbie:
This branch is conflicted, can you rebase it? Also, can you indicate which of the review comments you believe you have addressed? Thanks!

jonathaneen:
Hi Peter, this PR has escaped my attention for while.
I'll make sure to rebase it this week and try to process the review comments.

jonathaneen:
I'm currently working on the review comments. I've marked the comments that I think I have processed with a thumbsup

jonathaneen:
To be honest, I have a hard time determining which datastructure is used in apiServerZones (as per the suggestion to use the same format instead of introducing a new tuple) 

Currently my work schedule doesn't leave much room for work on this PR. 
I am happy to process the rest of the review comments later, but feel free to apply any code improvements in the meantime.

pieterlexis:

PR #8900, opened 2020-03-05
Detect {Libre,Open}SSL functions availability during configure

### LINKED ISSUE #8739: Better detection of OpenSSL/LibreSSL features availability
 - Issue type: Feature request

### Short description
We currently detect the availability of some OpenSSL / LibreSSL features based on the version number, which is tricky because LibreSSL does weird things with `OPENSSL_VERSION_NUMBER`. It was suggested in #8733 that we could do proper detection of some functions during configure instead, and that we don't need to set up thread-specific callbacks with recent enough (`>= 0x2090100fL`) versions of LibreSSL (which we do in auth, dnsdist and rec at the moment).

rgacogne:
### Short description

Instead of using version numbers, which is brittle, notably because LibreSSL faked the OpenSSL version numbers at some point without implementing all the newly introduced functions.

Fixes #8739.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/m4/pdns_check_libcrypto.m4
+++ b/m4/pdns_check_libcrypto.m4
@@ -111,7 +111,7 @@ AC_DEFUN([PDNS_CHECK_LIBCRYPTO], [
         [AC_LANG_PROGRAM([#include <openssl/crypto.h>], [ERR_load_CRYPTO_strings()])],
         [
             AC_MSG_RESULT([yes])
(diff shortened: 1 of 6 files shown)


rgacogne:
@omoerbeek This is low priority but, if you don't mind, I would really appreciate it if you could check that this doesn't break anything on OpenBSD :-)

omoerbeek:
auth, dnsdist and rec configure and build show no surprises.

PR #8909, opened 2020-03-06
dnsdist: Implement bounded loads for the whashed and wrandom policies

rgacogne:
### Short description

Also make sure that the `chashed` implementation respects the weights, and properly handle backends that are not available during the bounds computation.

Closes https://github.com/PowerDNS/pdns/issues/7387.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdist-lua.cc
+++ b/pdns/dnsdist-lua.cc
@@ -1652,6 +1652,18 @@ static void setupLuaConfig(bool client, bool configCheck)
       }
     });
 
+  g_lua.writeFunction("setWeightedBalancingFactor", [](double factor) {
+      setLuaSideEffect();
+      if (factor >= 0) {
+        g_weightedBalancingFactor = factor;
+      }
+      else {
+        errlog("Invalid value passed to setWeightedBalancingFactor()!");
+        g_outputBuffer="Invalid value passed to setWeightedBalancingFactor()!\n";
+        return;
+      }
+    });
+
   g_lua.writeFunction("setRingBuffersSize", [](size_t capacity, boost::optional<size_t> numberOfShards) {
       setLuaSideEffect();
       if (g_configurationDone) {
--- a/pdns/dnsdist-lua.cc
+++ b/pdns/dnsdist-lua.cc
@@ -1642,7 +1642,7 @@ static void setupLuaConfig(bool client, bool configCheck)
 
   g_lua.writeFunction("setConsistentHashingBalancingFactor", [](double factor) {
       setLuaSideEffect();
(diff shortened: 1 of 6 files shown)


pdns/dnsdistdist/docs/guides/serverselection.rst
```diff
@@ -55,10 +67,11 @@ This is a side-effect of the internal implementation of the consistent hashing a
...
-The higher the factor, the more imbalance between the servers is allowed.
+The algorithm will try to select a server based on the hash of the qname, as is done when no bounded-load is set, but will disqualify all servers that have more outstanding queries than intended times the factor, until a suitable server is found. The higher the factor, the more imbalance between the servers is allowed.
+
+For example, if we have two servers, with respective weights of 1 and 4, we expect the first server to get a fifth of the queries, and the second one 4/5. If the qname of the queries are not perfectly distributed, some server might get more queries than expected. Setting :func:`setWeightedBalancingFactor` to 1.1 limits the imbalance between the ratio of outstanding queries actually handled by a server and the expected number, so in this example the first server would not be allowed to handle more than 1.1/5 of all the outstanding queries at a given time.
```
chbruyand:
```suggestion
For example, if we have two servers, with respective weights of 1 and 4, we expect the first server to get a fifth of the queries, and the second one 4/5. If the qname of the queries are not perfectly distributed, some server might get more queries than expected. Setting :func:`setConsistentHashingBalancingFactor` to 1.1 limits the imbalance between the ratio of outstanding queries actually handled by a server and the expected number, so in this example the first server would not be allowed to handle more than 1.1/5 of all the outstanding queries at a given time.
```

rgacogne: Nice catch, thanks!

pdns/dnsdist-lua.cc
```diff
@@ -1647,6 +1647,18 @@ static void setupLuaConfig(bool client, bool configCheck)
       }
     });
 
+  g_lua.writeFunction("setWeightedBalancingFactor", [](double factor) {
+      setLuaSideEffect();
+      if (factor >= 0) {
```
chbruyand: we could get strange behavior in case of a factor that is set between 0 and 1 no ?

rgacogne: Done!

pdns/dnsdistdist/docs/guides/serverselection.rst
rgacogne: Fixed!

PR #8911, opened 2020-03-06
Add '/api/docs' endpoint to Auth server

kpfleming:
### Short description
Adds an '/api/docs' endpoint to allow clients to obtain the OpenAPI (Swagger) specification document directly from the server.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code


kpfleming:
There are some open issues here:

* Some clients will prefer to receive the OpenAPI document in JSON instead of YAML. There is no mechanism available in the Auth server code today to convert the YAML document into JSON, but we could configure Swagger to produce JSON instead of YAML and then supply the resulting document to clients who ask for either JSON or YAML (since JSON documents are also YAML documents).

* The new endpoint is part of the `/api` route and thus requires the api-key for access; it may be better to make the endpoint available without authentication.

* Requiring the administrator to set a configuration setting to point to the document is a bit of additional work for them, and will also require that the document be packaged and installed with the server. It may be possible to embed the document into the server binary itself so that neither of these are necessary, but I didn't see any existing examples in the code of that being done. I'm happy to make that change with some guidance for the preferred method.

Habbie:
Thanks!


>     * we could configure Swagger to produce JSON instead of YAML and then supply the resulting document to clients who ask for either JSON or YAML (since JSON documents are also YAML documents)

That sounds like something we could do during `make dist` or package build.



>     * require that the document be packaged and installed with the server

We can ship the file in our packages.



>     * It may be possible to embed the document into the server binary itself so that neither of these are necessary, but I didn't see any existing examples in the code of that being done

dnsdist-web.cc uses `incbin` to include some static HTML, but I prefer the packaging route.

kpfleming:
I'll also need to update the index.rst in the http-api docs to indicate that this new endpoint is available.

pdns/ws-api.cc
```diff
@@ -114,22 +114,50 @@ void apiDiscovery(HttpRequest* req, HttpResponse* resp) {
...
+
+  string fname = ::arg()["api-document-file"];
+  if (fname.empty())
+    throw HttpNotFoundException();
+
+#ifdef HAVE_CXX11
+  std::ifstream ifs(fname, std::ifstream::binary);
+#else
+  std::ifstream ifs(fname.c_str(), std::ifstream::binary);
+#endif
```
Habbie:
```suggestion
   std::ifstream ifs(fname, std::ifstream::binary);
```

pdns already requires C++11

omoerbeek:
Please rebase, there's a change in the test setup that CircleCI picks up only partially. That explains the test-recursor-bulk test falure. Rebasing should fix that.

kpfleming:
As far as I can tell the existing OAS (Open API Specification) document wasn't generated from any sources in this repository; is that correct? Is it being maintained by hand?

Habbie:
> As far as I can tell the existing OAS (Open API Specification) document wasn't generated from any sources in this repository; is that correct? Is it being maintained by hand?

Yes, it is maintained by hand.

kpfleming:
OK, so would you be willing to maintain it in JSON format instead of YAML? That would only require a one-time conversion, and no additional tools would be necessary during the packaging process.

Habbie:
> OK, so would you be willing to maintain it in JSON format instead of YAML? That would only require a one-time conversion, and no additional tools would be necessary during the packaging process.

@pieterlexis ?

pieterlexis:
I dislike typing JSON a lot more than YAML

> > * we could configure Swagger to produce JSON instead of YAML and then supply the resulting document to clients who ask for either JSON or YAML (since JSON documents are also YAML documents)

> That sounds like something we could do during make dist or package build.

> >    * require that the document be packaged and installed with the server

> We can ship the file in our packages.

> >   * It may be possible to embed the document into the server binary itself so that neither of these are necessary, but I didn't see any existing examples in the code of that being done

> dnsdist-web.cc uses incbin to include some static HTML, but I prefer the packaging route.

I really prefer us creating the JSON at build-time and including it in the binary. If we don't, the schema served could be changed by an operator on-disk, leading to support questions.

kpfleming:
Indeed, that is why I proposed including it into the binary, to avoid the risk of mismatches.

There are a number of mechanisms which could be used to do the conversion; the simplest is to use Python and the PyYAML package (easily installable on any system with Python). We could also bring a C++ YAML library into the source tree to avoid external dependencies.

Habbie:
> Indeed, that is why I proposed including it into the binary, to avoid the risk of mismatches.

Ok, agreed!



> We could also bring a C++ YAML library into the source tree to avoid external dependencies.

We use plenty of Python in our distribution and build already, so the PyYAML route sounds good to me!

kpfleming:
Excellent, I'll update this PR with the necessary changes. Stay tuned.

kpfleming:
New version has been posted here now. It embeds both the YAML and JSON forms of the document into the server binary, and I've tested retrieving with no Accept header, and with headers which specify application/json and application/x-yaml.

The LGTM analysis fails because it is unable to create a Python virtualenv which is required for the YAML->JSON conversion step; none of other virtualenv processes in the tree would have been triggered by LGTM previously because they are all for building documentation. We might need some help from the LGTM team figuring out how to solve this problem.

Habbie:
We should be able to fix it in `lgtm.yml`.

kpfleming:
It's also using Python 2.7 even though lgtm.yml explicitly asks for Python 3. I suspect the proper fix is to switch to using 'python3 -m venv .venv' instead of the virtualenv tool throughout the tree, but that's too large a change to bundle into this PR.

Habbie:
Hack: before calling make, make an empty or sufficiently empty `apidocfiles.h` so that the compile succeeds. We never run the binary that comes out of the LGTM compile anyway.

kpfleming:
Good call, a one-line hack in lgtm.yml is now in place.

kpfleming:
Success... all checks are green now :-)

zeha:
Would be great if the Recursor would support the same endpoint :)

Habbie:
> Would be great if the Recursor would support the same endpoint :)

Worth a request ticket I suppose, I don't want to hold this PR for every product :)

kpfleming:
Does the recursor support the same API?

Habbie:
> Does the recursor support the same API?

No, there are a lot of differences, and we don't have a Swagger definition for it anyway.

kpfleming:
Then I concur, an issue for that in the future seems like a good choice :-)

bootc:
Thanks for doing this! However I don't think this is enough by itself: the server will need to substitute some values into the spec file for the `host` and `schemes`:

https://swagger.io/docs/specification/2-0/api-host-and-base-path/

When loading the spec from disk for use with Bravado I currently substitute in the values myself; it would be nice to avoid having to do this:

```
def powerdns():
    url = urlparse(cfg.powerdns.url)

    http_client = RequestsClient()
    http_client.set_api_key(
        url.netloc, cfg.powerdns.key,
        param_name='X-API-Key', param_in='header'
    )

    spec_dict = load_file('authoritative-api-swagger.yaml')
    spec_dict['host'] = url.netloc
    spec_dict['schemes'] = [url.scheme]

    return SwaggerClient.from_spec(spec_dict, http_client=http_client)
```

kpfleming:
I do that substitution as well, for the same reason. From my understanding of the Bravado docs if `from_url` is invoked Bravado ignores the `host`, `scheme`, and `port` attributes in the spec and substitutes the ones which were used to obtain the spec. I don't know if this behavior is specific to Bravado or is common among Swagger clients.

In any case removing those attributes from the spec file would be a simple fix.

zeha:
This PR needs a rebase to fix conflicts in ws-recursor.cc.

Should it also change the swagger spec?

jsoref:
To help the [check-spelling](https://github.com/marketplace/actions/check-spelling) action, please rebase to `master` and add `apidocfiles` to `.github/actions/spell-check/expect.txt` as suggested:
https://github.com/jsoref/pdns/commit/e9f8a6344832bacb992c1ed5186d9cd6b95ce925#commitcomment-39229522

pdns/ws-recursor.cc
```diff
@@ -382,10 +382,10 @@ static void apiServerCacheFlush(HttpRequest* req, HttpResponse* resp) {
   DNSName canon = apiNameToDNSName(req->getvars["domain"]);
   bool subtree = (req->getvars.count("subtree") > 0 && req->getvars["subtree"].compare("true") == 0);
 
-  int count = broadcastAccFunction<uint64_t>(std::bind(pleaseWipeCache, canon, subtree, 0xffff));
-  count += broadcastAccFunction<uint64_t>(std::bind(pleaseWipePacketCache, canon, subtree, 0xffff));
-  count += broadcastAccFunction<uint64_t>(std::bind(pleaseWipeAndCountNegCache, canon, subtree));
-  resp->setBody(Json::object {
+  int count = broadcastAccFunction<uint64_t>(boost::bind(pleaseWipeCache, canon, subtree, 0xffff));
```
Habbie, line 385: why the change from `std` to `boost`?

kpfleming: I honestly have no idea where this change came from! I don't remember changing it nor can I see any reason why it would have been changed as it's not related to the topic of this PR at all.

Habbie: Probably happened during a rebase then. Can you take it out? :)

kpfleming: Done.

Habbie:
This breaks when building from a distribution tarball: `builder/build.sh -m authoritative debian-buster` ends with `make[3]: *** No rule to make target '../docs/http-api/swagger/authoritative-api-swagger.yaml', needed by 'api-swagger.yaml'.  Stop.`

I can probably look at this at some later time if you don't have time or don't feel like fighting our Makefiles.

kpfleming:
That would indicate that the file is not being included in the tarball... that's unfortunate. I hadn't looked into those portions of the Makefiles yet, so if you're able to investigate this one that would probably get to a solution more quickly.

kpfleming:
Is this still a candidate for inclusion in 4.4.0 (since alpha releases are already being produced)? If so, I'll get it rebased and fix up the distribution tarball issue, since I've spent some time working in the builder tooling recently :-)

Habbie:
> Is this still a candidate for inclusion in 4.4.0

Yes, within a few weeks (at beta) this is fine! Let me know if you need any help.

kpfleming:
I'm actively working on this again, should have a new version pushed in the next 24 hours.

kpfleming:
Branch updated; it has been rebased on current `master`, and builds from source tarballs (produced by `make dist`) now complete as they should.

pieterlexis:
The github action config might need an update:

```
You need virtualenv to generate the JSON API document
exit 1
Makefile:3590: recipe for target 'api-swagger.json' failed
```

kpfleming:
Ugh. It's getting harder and harder to use Python 2.7 and the `virtualenv` tool... as noted in another PR, we're long past the point where upgrading the build system to use Python 3.x is necessary.

New commit pushed which adds `python-virtualenv` package to the CodeQL workflow. Fingers crossed :-)

kpfleming:
Another thought: since `virtualenv` is no longer optional for building the Authoritative Server, the configure script for it should bail out if it was not found, instead of proceeding and then producing an error during the build.

Habbie:
> Another thought: since `virtualenv` is no longer optional for building the Authoritative Server, the configure script for it should bail out if it was not found, instead of proceeding and then producing an error during the build.

Is this also true when building from a tarball?

kpfleming:
At the moment it is, because the source tarball only contains the YAML version of the Swagger document. It may be possible to have the 'dist' targets do the YAML->JSON conversion and then include both formats in the tarball, I can experiment with that.

Habbie:
Yes, that would make sense, then it's similar to how we handle manpages (those also need Python to generate them, but then we stick them in the tarball)

kpfleming:
OK, I'll make those changes too.

pieterlexis:
If you need help with the autotools/autoconf/m4, let me known

kpfleming:
OK, that's fixed. `api-swagger.yaml` and `api-swagger.json` are now included in source tarballs, so `virtualenv` is not required to build from them. If `virtualenv` is available it will still be used and changes can be made to those files as normal for inclusion into the server binary.

`virtualenv` is still required to build from a plain repository clone, though.

zeha:
lgtm!

kpfleming:
I dare say this might finally be ready for merging :-)

Habbie:
Looks good at a quick skim! I've kicked off a test build at https://builder.powerdns.com/#/buildrequests/129582?redirect_to_build=true - normally we just merge and see, but as this touches the build system directly, I'd like to wait for that.

Habbie:
@pieterlexis can you check the autotools bits in here?

pieterlexis (approved):
I like it

Habbie:
@kpfleming many thanks!

kpfleming:
You're all very welcome... now I need to update my Ansible modules which use the API to pull the Swagger spec directly from the server :-)

kpfleming:
With the release of 4.4.0-beta1, my Ansible modules which use the API to manage Zones and TSIG Keys are now passing automated tests using direct extraction of the Swagger spec from the server, instead from a local file. Success!

https://github.com/kpfleming/ansible-pdns-auth-api

PR #8918, opened 2020-03-10
rec: Implement EDNS0 padding (rfc7830) for outgoing responses

rgacogne:
### Short description

The initial idea is to be able to add EDNS padding when answering to queries received over DoT or DoH. This PR allows adding this option based on the source (so you could use a different source when forwarding a query from dnsdist to the recursor, based on whether it was received over DoH or DoT), and optionally on whether the initial query had padding.

One of the challenge is that this duplicates the content of the packet cache to avoid mixing padded and non-padded responses.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/pdns/ednspadding.cc
+++ b/pdns/ednspadding.cc
@@ -1,0 +1,31 @@
+/*
+ * This file is part of PowerDNS or dnsdist.
+ * Copyright -- PowerDNS.COM B.V. and its contributors
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of version 2 of the GNU General Public License as
+ * published by the Free Software Foundation.
+ *
+ * In addition, for the avoidance of any doubt, permission is granted to
+ * link this program with OpenSSL and to (re)distribute the binaries
+ * produced as the result of such linking.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+
+#include "ednspadding.hh"
+
+std::string makeEDNSPaddingOptString(size_t bytes)
+{
+  std::string ret;
+  ret.resize(bytes, '0');
+  return ret;
+}
+
--- a/pdns/ednspadding.cc
+++ b/pdns/ednspadding.cc
@@ -25,7 +25,7 @@
 std::string makeEDNSPaddingOptString(size_t bytes)
 {
   std::string ret;
-  ret.resize(bytes, '0');
+  ret.resize(bytes, 0);
   return ret;
 }
 
--- a/pdns/pdns_recursor.cc
+++ b/pdns/pdns_recursor.cc
@@ -5602,7 +5602,7 @@ int main(int argc, char **argv)
 
     ::arg().setSwitch("aggressive-nsec-cache-size", "The number of records to cache in the aggressive cache. If set to a value greater than 0, and DNSSEC validation is enabled, the recursor will cache NSEC and NSEC3 records to generate negative answers, as defined in rfc8198")="100000";
 
-    ::arg().set("edns-padding-from", "List of netmasks (proxy IP in case of XPF or proxy-protocol presence, client IP otherwise) for which EDNS padding will be enabled in responses, provided that 'edns-padding-mode' is set")="";
+    ::arg().set("edns-padding-from", "List of netmasks (proxy IP in case of XPF or proxy-protocol presence, client IP otherwise) for which EDNS padding will be enabled in responses, provided that 'edns-padding-mode' applies")="";
     ::arg().set("edns-padding-mode", "Whether to add EDNS padding to all responses ('always') or only to responses for queries containing the EDNS padding option ('padded-queries-only', the default). In both modes, padding will only be added to responses for queries coming from `edns-padding-from`_ sources")="padded-queries-only";
     ::arg().set("edns-padding-tag", "Packetcache tag associated to responses sent with EDNS padding, to prevent sending these to non-whitelisted clients.")="7830";
 
@@ -5719,4 +5719,3 @@ int main(int argc, char **argv)
 
   return ret;
 }
-
--- a/pdns/pdns_recursor.cc
+++ b/pdns/pdns_recursor.cc
@@ -359,6 +359,7 @@ struct DNSComboWriter {
   bool d_followCNAMERecords{false};
   bool d_logResponse{false};
   bool d_tcp{false};
+  bool d_responsePaddingDisabled{false};
 };
 
 MT_t* getMT()
@@ -1490,7 +1491,7 @@ static void startDoResolve(void *p)
     bool variableAnswer = dc->d_variable;
     bool haveEDNS=false;
     bool paddingAllowed = false;
-    bool padResponse = false;
+    bool addPaddingToResponse = false;
 #ifdef NOD_ENABLED
     bool hasUDR = false;
 #endif /* NOD_ENABLED */
@@ -1502,7 +1503,7 @@ static void startDoResolve(void *p)
         ednsExtRCode = ERCode::BADVERS;
       }
 
-      if(!dc->d_tcp) {
+      if (!dc->d_tcp) {
         /* rfc6891 6.2.3:
            "Values lower than 512 MUST be treated as equal to 512."
         */
@@ -1511,10 +1512,10 @@ static void startDoResolve(void *p)
       ednsOpts = edo.d_options;
       maxanswersize -= 11; // EDNS header size
 
-      if (g_paddingFrom.match(dc->d_remote)) {
+      if (!dc->d_responsePaddingDisabled && g_paddingFrom.match(dc->d_remote)) {
         paddingAllowed = true;
         if (g_paddingMode == PaddingMode::Always) {
-          padResponse = true;
+          addPaddingToResponse = true;
         }
       }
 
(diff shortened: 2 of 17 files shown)


Habbie:
> One of the challenge is that this duplicates the content of the packet cache to avoid mixing padded and non-padded responses.

Without having read the code - the one thing our packet cache responses currently special-case is the QID, and the exact caPiTAlISation of the QNAME. Perhaps we can do the padding the same way?

pdns/pdns_recursor.cc
```diff
@@ -1172,9 +1179,24 @@ static void startDoResolve(void *p)
             // Option Code and Option Length are both 2
             maxanswersize -= 2 + 2 + mode_server_id.size();
           }
+        } else if (paddingAllowed && padResponse == false && g_paddingMode == PaddingMode::PaddedQueries && o.first == EDNSOptionCode::PADDING) {
+          /* we should only pad on 'secure' cases to limit amplification:
+             - over TCP ;
+             - from 'known-good' sources (dnsdist for example).
+             We should also support a strict mode (only pad if the query has a padding option)
```
Habbie:
```suggestion
             TODO: We should also support a strict mode (only pad if the query has a padding option)
```

pdns/pdns_recursor.cc
```diff
@@ -1172,9 +1179,24 @@ static void startDoResolve(void *p)
...
+             and a relaxed mode (pad if the client supports EDNS).
```
Habbie: This reminds me of https://community.cloudflare.com/t/doh-service-adds-edns-padding-even-if-client-does-not-indicate-edns-support/84199 :)

pdns/pdns_recursor.cc
rgacogne: It's actually already implemented, I'll update the comment :)

rgacogne:
> (The next question then would be if what we put in the packet cache is padded or not; for 'padded', we could send a 'truncated' version when we're not padding; this might be cheaper than a resize when doing a 'padded' from a cached non-padded response. Or perhaps none of this matters because the QID/QNAME fixing already involves a copy, and we can fix that copy to already allocate the right padded/nonpadded size in one go.)

Fixing the padding is not that easy, because we also need to update the size of the OPT rdata length.

Habbie:
> Fixing the padding is not that easy, because we also need to update the size of the OPT rdata length.

Ah, yes :(

omoerbeek:
@rgacogne this PR is still in "Draft" status. It is already worth the time doing a review? Or is it better to wait?

rgacogne:
> @rgacogne this PR is still in "Draft" status. It is already worth the time doing a review? Or is it better to wait?

IMHO we need to decide whether we are interested in implementing this, and if we are what to do for the packet cache. It wouldn't be _that_ hard to add the padding on a packet cache hit, if we decide it's worth it.

rgacogne:
I'm wondering whether the packet duplication is a real issue. Until #8993 is merged we already store  different responses depending on the incoming padding anyway, and once it's merged we will only store two.

pieterlexis:
I've closed #10018 in favor of this one. @omoerbeek can you milestone this PR?

omoerbeek:
@rgacogne As there's been some interest in getting this, would you be able to get this into shape for 4.5? If not, we'll move the milestone, but if it is possible that would be nice.

rgacogne:
> As there's been some interest in getting this, would you be able to get this into shape for 4.5? If not, we'll move the milestone, but if it is possible that would be nice.

Sure, but the reason is stayed a draft PR for so long is that I would like some feedback on the general logic, and especially on what we should with regard to the packet cache.

pdns/ednspadding.cc
```diff
@@ -0,0 +1,31 @@
...
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+
+#include "ednspadding.hh"
+
+std::string makeEDNSPaddingOptString(size_t bytes)
+{
+  std::string ret;
+  ret.resize(bytes, '0');
```
omoerbeek, line 28: De we want zero or NUL?

pdns/recursordist/docs/settings.rst
```diff
@@ -548,6 +548,40 @@ found, the recursor fallbacks to sending 127.0.0.1.
...
+``edns-padding-mode``
+---------------------
+.. versionadded:: 4.4.0
+
+-  One of ``always``, ``padded-queries-only``, String
+-  Default: ``padded-queries-only``
+
+Whether to add EDNS padding to all responses (``always``) or only to the ones for padded queries (``padded-queries-only``, the default) coming from `edns-padding-from`_ sources.
+
+.. _setting-edns-padding-mode:
```
omoerbeek, line 573:
```suggestion
.. _setting-edns-padding-tag:
```

pdns/recursordist/docs/settings.rst
```diff
@@ -548,6 +548,40 @@ found, the recursor fallbacks to sending 127.0.0.1.
...
+
+``edns-padding-mode``
+---------------------
+.. versionadded:: 4.4.0
+
+-  One of ``always``, ``padded-queries-only``, String
+-  Default: ``padded-queries-only``
+
+Whether to add EDNS padding to all responses (``always``) or only to the ones for padded queries (``padded-queries-only``, the default) coming from `edns-padding-from`_ sources.
```
omoerbeek, line 571: From this description it is not immediate clear if queries outside `edns-padding-from` are padded or not.

pdns/pdns_recursor.cc
```diff
@@ -1640,6 +1661,30 @@ static void startDoResolve(void *p)
...
+
+      if (currentSize < (maxSize - 4)) {
+        size_t remaining = maxSize - (currentSize + 4);
+        /* from rfc8647, "4.1.  Recommended Strategy: Block-Length Padding":
+           If a server receives a query that includes the EDNS(0) "Padding"
+           option, it MUST pad the corresponding response (see Section 4 of
+           RFC 7830) and SHOULD pad the corresponding response to a
+           multiple of 468 octets (see below).
+        */
+        static const size_t blockSize = 468;
```
omoerbeek, line 1678: Little use making a `const int` static here

omoerbeek (commented):
- I think the extra pressure on the packetcache is acceptable, especially since we now have #8993 indeed.
- If it does becomes a problem we can implement padding after packet cache retrieval. Wonder if it's worth the complexity though.
- Please make sure in the description of the PR that we're talking client side only

pdns/ednspadding.cc
rgacogne: A `CharT` whose value value is `0`.

pdns/recursordist/docs/settings.rst
rgacogne: This PR implements padding for outgoing responses, not for outgoing queries, perhaps this description should be clearer about that?

omoerbeek: My comment should have read:
From this description it is not immediate clear if responses to queries outside edns-padding-from are padded or not.

pdns/recursordist/docs/settings.rst
```diff
@@ -548,6 +548,40 @@ found, the recursor fallbacks to sending 127.0.0.1.
...
+-  Comma separated list of netmasks
+-  Default: (none)
+
+List of netmasks (proxy IP in case of XPF or proxy-protocol presence, client IP otherwise) for which EDNS padding will be enabled in responses. See also `edns-padding-mode`_.
+
+.. _setting-edns-padding-mode:
+
+``edns-padding-mode``
+---------------------
+.. versionadded:: 4.4.0
```
pieterlexis, line 566: 4.5.0

pdns/recursordist/docs/settings.rst
```diff
@@ -548,6 +548,40 @@ found, the recursor fallbacks to sending 127.0.0.1.
...
+
+``edns-padding-tag``
+--------------------
+.. versionadded:: 4.4.0
+
+-  Integer
+-  Default: 7830
+
+The packetcache tag to use for padded responses, to prevent a client not allowed by the `edns-padding-from`_ list to be served a cached answer generated for an allowed one. This
+effectively divides the packet cache in two when `edns-padding-from`_ is used. Note that this will not override a tag set from one of the ``Lua`` hooks.
```
pieterlexis, line 583: it might be good to expose whether or not a response would be padded in the `dq` object?

pieterlexis: (if at all possible without making a trail of blood through the recursor)

rgacogne:
Rebased, fixing a few reported issues in the process. I'll see how hard it would be to make the padding status available to Lua hooks.

pdns/recursordist/docs/settings.rst
rgacogne: I have updated the description, please let me know if it's still unclear. It's sometimes hard to take the user point-of-view when you wrote you spent some time writing the actual code! :)

pdns/ednspadding.cc
omoerbeek: Form the RFC: `The PADDING octets SHOULD be set to 0x00.`.
So I think should be `0` (or `'\0'`and not `'0'`. The latter is ASCII zero with value `0x30`.

pdns/recursordist/docs/settings.rst
omoerbeek: Yes, much more clear.

pdns/ednspadding.cc
rgacogne: Ouch, sorry, will fix!

rgacogne:
Fixed the documentation as it turned out my memory was not right and the options provided by this PR are:
- Don't send padded responses (the default) ;
- Send padded responses to all queries from allow-listed (`edns-padding-from`) clients (`edns-padding-mode` set to `always`) ;
- Send padded responses to queries from allow-listed (`edns-padding-from`) clients provided that the query contains an EDNS Padding option (`edns-padding-mode` set to `padded-queries-only`).

Otto, does that seem OK to you?

I also added the ability to disable EDNS Padding from `gettag_ffi` and `DNSQuestion` based Lua hooks.

pdns/pdns_recursor.cc
```diff
@@ -5530,6 +5602,10 @@ int main(int argc, char **argv)
 
     ::arg().setSwitch("aggressive-nsec-cache-size", "The number of records to cache in the aggressive cache. If set to a value greater than 0, and DNSSEC validation is enabled, the recursor will cache NSEC and NSEC3 records to generate negative answers, as defined in rfc8198")="100000";
 
+    ::arg().set("edns-padding-from", "List of netmasks (proxy IP in case of XPF or proxy-protocol presence, client IP otherwise) for which EDNS padding will be enabled in responses, provided that 'edns-padding-mode' is set")="";
```
omoerbeek, line 5605:
```suggestion
    ::arg().set("edns-padding-from", "List of netmasks (proxy IP in case of XPF or proxy-protocol presence, client IP otherwise) for which EDNS padding will be enabled in responses, provided that 'edns-padding-mode' applies")="";
```

pdns/recursordist/docs/settings.rst
```diff
@@ -610,6 +610,41 @@ found, the recursor fallbacks to sending 127.0.0.1.
...
+.. _setting-edns-padding-from:
+
+``edns-padding-from``
+---------------------
+.. versionadded:: 4.5.0
+
+-  Comma separated list of netmasks
+-  Default: (none)
+
+List of netmasks (proxy IP in case of XPF or proxy-protocol presence, client IP otherwise) for which EDNS padding will be enabled in responses, provided that `edns-padding-mode`_ is set.
```
omoerbeek, line 622:
```suggestion
List of netmasks (proxy IP in case of XPF or proxy-protocol presence, client IP otherwise) for which EDNS padding will be enabled in responses, provided that `edns-padding-mode`_ applies.
```

omoerbeek (approved):
This time I compared the docs against the code and observed behaviour. It looks good with two small nits.

Issue #8922, opened 2020-03-12
Service is not reloaded on systemd during %post on updates


atoomic:
### Short description

the builder-support/specs/dnsdist.spec is using `%systemd_post` in the `%post` section, but unfortunately on several distro that macro does not enforce a reload of systemd.

We should consider adding `systemctl daemon-reload` on update otherwise any change to the service file will be ignored by systemd.

```
%post
...
%if 0%{?rhel} >= 7
%systemd_post %{name}.service
%endif
```

Example on CentOS7 for the macro `%systemd_post`
```
> rpm --eval "%systemd_post foo"

if [ $1 -eq 1 ] ; then
        # Initial installation
        systemctl preset foo >/dev/null 2>&1 || :
fi
```

### Expected behaviour
call `systemctl daemon-reload` on reload

rgacogne:
Fixed in #8923.

PR #8923, opened 2020-03-12
Reload systemctl service on updates

atoomic:
Fix GH #8922

### DIFF
--- a/builder-support/specs/dnsdist.spec
+++ b/builder-support/specs/dnsdist.spec
@@ -152,6 +152,7 @@ fi
 %service_add_post %{name}.service
 %endif
 %if 0%{?rhel} >= 7
+systemctl daemon-reload ||:
 %systemd_post %{name}.service
 %endif
 
--- a/builder-support/specs/pdns-recursor.spec
+++ b/builder-support/specs/pdns-recursor.spec
@@ -116,6 +116,7 @@ exit 0
 %if 0%{?rhel} == 6
 chkconfig --add %{name}
 %else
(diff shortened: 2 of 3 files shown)


builder-support/specs/dnsdist.spec
```diff
@@ -153,6 +153,10 @@ fi
 %endif
 %if 0%{?rhel} >= 7
 %systemd_post %{name}.service
+if [ $1 -eq 2 ] ; then
```
pieterlexis: I think this should come before `%systemd_post %{name}.service`

atoomic: note that order does not really matters as long as we are using a check for update as `%systemd_post` translate to `if [ $1 -eq 1 ] ; then ... fi`
I was assuming `preset` was implying a daemon-reload but after reading some docs https://freedesktop.org/wiki/Software/systemd/Preset/ it s mainly going to trigger enable 
so we would have to move it above and remove the condition

atoomic:
probably better to just do `systemctl daemon-reload ||:` otherwise it will be impossible to know if it fails and adjust it if needed...

phonedph1:
So in EL7 this happens in the postun triggers, and now in EL8 they are removed. Does it make sense to duplicate the same effort in the postun sections instead, to match what EL7 does?

atoomic:
@phonedph1 I think it s a pretty bad idea to use the `postun` macro during post, as you have no guarantee that this will not change overtime/distro... and trigger some unexpected behavior

PR #8924, opened 2020-03-12
auth: make sure we look at 10% of all cached items during cleanup

mind04:
### Short description
Purging was very inefficient for large shards when the cached items are not equally distributed over all shards. We now look at 10% of the cached items in every shard. This was 10% of all items divided over all shards.

In some setups this will result in a dramatic decrease of cached items (up to 100x) with an identical cache hit rate.

### Checklist
I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/auth-packetcache.cc
+++ b/pdns/auth-packetcache.cc
@@ -248,11 +248,7 @@ uint64_t AuthPacketCache::purge(const string &match)
 			   
 void AuthPacketCache::cleanup()
 {
-  uint64_t maxCached = d_maxEntries;
-  uint64_t cacheSize = *d_statnumentries;
-  uint64_t totErased = 0;
-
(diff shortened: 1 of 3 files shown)


mind04:
Scanning 100k entries is exactly what we did before this pull as well. (assuming every shard is for at least 10% filled in a large cache) What has changed is that we now scan the 100k entries with the biggest change of encountering expired entries. The behavior change is only for small caches. In the large ones it wiill still look at 10% of every shard.
The above max cleaning was remove since it is useless now. The cache will never grow beyond the maximum size.

rgacogne:
You are completely right, I somehow managed to forget that we could not go above the maximum number of entries anymore.
I'm still a bit worried about the latency impact but it's not caused by this PR, but is a collateral effect of not exceeding the maximum anymore.

PR #8941, opened 2020-03-17
lua: better errors when a key/value table is expected

Habbie:
### Short description

```
> pc = newPacketCache(5000000, {86400, 0, 60, 60, false, 40} )
```

Before: `Error parsing parameters, did you forget parameter name?` (lua 5.3) or `Unknown Lua error` (luajit), without a line number.

Now: `[string "pc = newPacketCache(5000000, {86400, 0, 60, 6..."]:1: Caught exception: Expected key/value table, got array`

The enforcement on unordered map to not have a length is pretty brutal. I wonder if CI will pick anything up, I only tested dnsdist.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/ext/luawrapper/include/LuaContext.hpp
+++ b/ext/luawrapper/include/LuaContext.hpp
@@ -2802,6 +2802,17 @@ struct LuaContext::Reader<std::unordered_map<TKey,TValue,THash,TKeyEqual>>
         if (!lua_istable(state, index))
             return boost::none;
 
+        size_t len;
+
+#if     LUA_VERSION_NUM >= 502
+        len = lua_rawlen(state, index);
+#else
+        len = lua_objlen(state, index);
+#endif
+
+        if (len != 0)
+            throw std::logic_error("Expected key/value table, got array");
+
         std::unordered_map<TKey,TValue,THash,TKeyEqual> result;
 
         // we traverse the table at the top of the stack
--- a/ext/luawrapper/include/LuaContext.hpp
+++ b/ext/luawrapper/include/LuaContext.hpp
@@ -2649,11 +2649,21 @@ struct LuaContext::Reader<std::string>
     static auto read(lua_State* state, int index)
         -> boost::optional<std::string>
     {
+        std::string result;
+
+        // lua_tolstring might convert the variable that would confuse lua_next, so we
+        //   make a copy of the variable.
+        lua_pushvalue(state, index);
+
         size_t len;
-        const auto val = lua_tolstring(state, index, &len);
-        if (val == 0)
-            return boost::none;
-        return std::string(val, len);
+        const auto val = lua_tolstring(state, -1, &len);
+
+        if (val != 0)
+          result.assign(val, len);
+
+        lua_pop(state, 1);
+
+        return val != 0 ? boost::optional<std::string>{ std::move(result) } : boost::none;
     }
 };
 
--- a/ext/luawrapper/include/LuaContext.hpp
+++ b/ext/luawrapper/include/LuaContext.hpp
@@ -2764,6 +2764,17 @@ struct LuaContext::Reader<std::map<TKey,TValue>>
         if (!lua_istable(state, index))
             return boost::none;
 
+        size_t len;
+
+#if     LUA_VERSION_NUM >= 502
+        len = lua_rawlen(state, index);
+#else
+        len = lua_objlen(state, index);
+#endif
+
+        if (len != 0)
+            throw std::logic_error("Expected key/value table, got array");
+
         std::map<TKey,TValue> result;
 
         // we traverse the table at the top of the stack
--- a/pdns/dnsdist-console.cc
+++ b/pdns/dnsdist-console.cc
@@ -316,10 +316,7 @@ void doConsole()
       // tried to return something we don't understand
     }
     catch(const LuaContext::ExecutionErrorException& e) {
-      if(!strcmp(e.what(),"invalid key to 'next'"))
-        std::cerr<<"Error parsing parameters, did you forget parameter name?";
-      else
-        std::cerr << e.what(); 
+      std::cerr << e.what();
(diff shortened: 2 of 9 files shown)


Habbie:
> ``
> 
> The enforcement on unordered map to not have a length is pretty brutal.

Should probably extend this to `std::map` etc. as well.

Habbie:
I have local fixes for the CI failures, will push soon.

Habbie:
Perhaps it would make sense to add a typedef or two (`LuaArray` -> `vector<pair<int, ..>>`, `LuaMap` -> `map` or `unordered_map`) to make these things clearer to someone reading the code, and also less awkward to type?

Habbie:
> The PR looks good but I'm worried we might need to update a few more places:

I'm fixing up the last(?) failing tests, which covers about half of your list. I'll grep some more as well :)

rgacogne:
Oh and also:
```
pdns/dnsdist-lua-inspection.cc:29:static std::unordered_map<unsigned int, vector<boost::variant<string,double>>> getGenResponses(unsigned int top, boost::optional<int> labels, std::function<bool(const Rings::Response&)> pred)
pdns/dnsdist-lua-inspection.cc:70:  std::unordered_map<unsigned int, vector<boost::variant<string,double>>> ret;
pdns/dnsdist-lua-inspection.cc:297:      std::unordered_map<unsigned int, vector<boost::variant<string,double>>> ret;
```

rgacogne:
> Perhaps it would make sense to add a typedef or two (LuaArray -> vector<pair<int, ..>>, LuaMap -> map or unordered_map) to make these things clearer to someone reading the code, and also less awkward to type?

That sounds like a very nice idea :+1:

Habbie:
> pdns/lua-auth4.cc:27:      std::unordered_map<int, DNSResourceRecord> luaResult;

This one is returned from C++ to Lua, so it does not hit the assertion.

rgacogne:
> This one is returned from C++ to Lua, so it does not hit the assertion (but I changed it anyway for clarity).

Ah, yes, several of the ones I mentioned earlier do. I did not realize at first that they would not be impacted by the change but I agree it would be better to change them for clarity anyway :)

Habbie:
> Ah, yes, several of the ones I mentioned earlier do. I did not realize at first that they would not be impacted by the change but I agree it would be better to change them for clarity anyway :)

The DNSCrypt `getCertificatePairs` is the same, BUT:
# before
```
> t=getDNSCryptBind(0):getCertificatePairs()
> t
Command returned an object we can't print: Trying to cast a lua variable from "table" to "N5boost8optionalINS_7variantINSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEEJSt10shared_ptrI15DownstreamStateEP11ClientStateSt13unordered_mapIS7_dSt4hashIS7_ESt8equal_toIS7_ESaISt4pairIKS7_dEEEEEEEE"
```

current PR state:
```
> t
Expected key/value table, got array
```

(This does not change if I fix the type, because the problem appears to be the variant set in g_lua.executeCode in dnsdist-console.cc. We really want to do better here but I'm tempted to not bloat this PR to that point.)

Habbie:
> pdns/dnsdist-lua-inspection.cc:297:      std::unordered_map<unsigned int, vector<boost::variant<string,double>>> ret;

This is in `getTopQueries` which is undocumented. It is also unaffected. I'm going to ignore the rings for now.

Habbie:
> `getTopQueries` which is undocumented

Right, used internally by `topQueries` which still works.

Habbie:
Going to leave this for today (and wait for CI), then do another grep. Then I'll refactor with the suggested types or something similar to them.

omoerbeek:
For rec:
`dnstapFrameStreamServer("127.0.0.1:9999", {1}) ` produces:
`Mar 25 09:53:19 Unable to load Lua script from 'lua-config-file': STL Exception: [string "chunk"]:1: Caught exception: Expected key/value table, got array`

which is better than before. At some point it would even be nicer to show the offending code, like dnsdist is doing in your example. Wondering how difficult that would be to do in the recursor.

This type feels really strange: 
`std::vector<std::pair<int, std::string>>` which is filled with `pair(++i, ...)`, the first element of the seems 100% redundant, since it equals the index of the vector (apart from the lua off-by-one)... But I suppose this is needed by the lua interface?

Habbie:
> This type feels really strange:
> `std::vector<std::pair<int, std::string>>` which is filled with `pair(++i, ...)`, the first element of the seems 100% redundant, since it equals the index of the vector (apart from the lua off-by-one)... But I suppose this is needed by the lua interface?

Yes - this is why I suggested defining a convenient 'LuaArray' type or similar to see what we can simplify here.

Habbie:
> At some point it would even be nicer to show the offending code, like dnsdist is doing in your example. Wondering how difficult that would be to do in the recursor.

We should check if #8942 fixes/improves that. If not, I'll have a separate look at it.

omoerbeek:
> > At some point it would even be nicer to show the offending code, like dnsdist is doing in your example. Wondering how difficult that would be to do in the recursor.
> 
> We should check if #8942 fixes/improves that. If not, I'll have a separate look at it.

Nope:
`Mar 25 10:54:28 Unable to load Lua script from 'lua-config-file': STL Exception: Unknown Lua error`
This is with only #8942

Habbie:
> Nope:

I can look at this later, or you can see if stealing the `rethrow_if_nested` bit from dnsdist-console.cc helps :)

omoerbeek:
There's alreay code calling `std::rethrow_if_nested(e)` in `rec-lua-conf.cc`
Debugging this seems to show that there's no more detailed info in the exception...

Habbie:
Ah yes. I'll have go at it later.

pieterlexis:

Issue #8866, opened 2020-02-26
status NXDOMAIN on remote http backend when ANY request returns HTTP 500


administratorius:
- Program: Authoritative
 - Issue type: Bug report

### Short description

We're implementing HTTP DNS storage using "remote" backend, keeping SOA record statically defined in our code. If our DB is failing and /ANY GET request returns non 20x HTTP status, but /SOA is 200 - query status becomes NXDOMAIN

### Environment

 - Operating system: CentOS 7.7.1908
 - Software version: 4.2.1
 - Software source: PowerDNS repository powerdns-auth-42

### Steps to reproduce

/etc/pdns/pdns.conf:

```
config-dir=/etc/pdns
setuid=pdns
setgid=pdns
log-dns-details=yes
log-dns-queries=yes
loglevel=6
webserver=yes
webserver-address=0.0.0.0
launch=remote
remote-connection-string=http:url=http://localhost:8080
```

Forcing ANY return status is 302 on the backend, anything not 20x results the same,  only SOA GET status is 200

HTTP backend log snippet:
```
<...>
GET /lookup/cn%2eexample%2ecom%2e/SOA 200 159 - 0.242 ms
<...>
GET /lookup/%2a%2ecn%2eexample%2ecom%2e/ANY 302 5 - 0.229 ms
<...>
```

dig returns:
`status: NXDOMAIN`

### Expected behaviour
dig returning:
`status: REFUSED` or `status: SERVFAIL`

### Actual behaviour
Backend returning HTTP status 302 with `Error` string. pdns log:
```
Feb 26 12:46:56 pdns pdns_server: Cannot parse JSON reply: expected value, got 'E' (69)
Feb 26 12:46:56 pdns pdns_server: Cannot parse JSON reply: expected value, got 'E' (69)
Feb 26 12:46:57 pdns pdns_server: Cannot parse JSON reply: expected value, got 'E' (69)
Feb 26 12:46:57 pdns pdns_server: Cannot parse JSON reply: expected value, got 'E' (69)
Feb 26 12:46:57 pdns pdns_server: Cannot parse JSON reply: expected value, got 'E' (69)
Feb 26 12:46:57 pdns pdns_server: Cannot parse JSON reply: expected value, got 'E' (69)
Feb 26 12:46:57 pdns pdns_server: Cannot parse JSON reply: expected value, got 'E' (69)
[...]

PR #8943, opened 2020-03-17
Remote Backend: Throw DBException in functions that allow it

pieterlexis:
### Short description
These functions are called inside a try block in e.g. the
slavecommunicator and them `throw`ing actually makes the reporting
better.

### Checklist

I have:
- [x] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/modules/remotebackend/httpconnector.cc
+++ b/modules/remotebackend/httpconnector.cc
@@ -418,8 +418,9 @@ int HTTPConnector::recv_message(Json& output) {
 
     arl.finalize();
 
-    if (resp.status < 200 || resp.status >= 400) {
+    if ((resp.status < 200 || resp.status >= 400) && resp.status != 404) {
       // bad. 
+      g_log<<Logger::Warning<<"Received unacceptable HTTP status code "<<std::to_string(resp.status)<<" from HTTP endpoint"<<endl;
       return -1;
     }
 
--- a/modules/remotebackend/httpconnector.cc
+++ b/modules/remotebackend/httpconnector.cc
@@ -384,7 +384,6 @@ int HTTPConnector::recv_message(Json& output) {
     if (d_socket == nullptr ) return -1; // cannot receive :(
     char buffer[4096];
     int rd = -1;
-    bool fail = false;
     time_t t0;
 
     arl.initialize(&resp);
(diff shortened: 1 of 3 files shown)


rgacogne:
Out if curiosity, did you consider making `RemoteBackend::send()` and `RemoteBackend::recv()` throw instead? It might solve https://github.com/PowerDNS/pdns/issues/8866 as well.

pieterlexis:
> Out if curiosity, did you consider making RemoteBackend::send() and RemoteBackend::recv() throw instead? It might solve #8866 as well.

I was afraid with blowing up the auth process on a failure.. but it looks like the database backend do the same as well. I will change this PR to indeed throw on send and recv failures.

pieterlexis:
hhrm, by the looks of the CI, `throw`ing on `{"result": false}` might be a bit too much :)

rgacogne:
> hhrm, by the looks of the CI, throwing on {"result": false} might be a bit too much :)

That's a valid result for some operations, I believe?

pieterlexis:
> That's a valid result for some operations, I believe?

it is

modules/remotebackend/remotebackend.cc
```diff
@@ -42,18 +42,27 @@ bool Connector::send(Json& value) {
...
-    return false;
+    for(const auto& message: value["log"].array_items()) {
+      g_log<<Logger::Info<<"[remotebackend]: "<< message.string_value() <<std::endl;
+    }
+    return retval;
+  }
+  throw PDNSException("Unknown error while receiving data");
+}
+
+void makeErrorAndThrow(Json &value) {
```
cmouse: is this intentionally not a class member?

pieterlexis: no, accidentally :)

modules/remotebackend/httpconnector.cc
```diff
@@ -418,8 +418,9 @@ int HTTPConnector::recv_message(Json& output) {
 
     arl.finalize();
 
-    if (resp.status < 200 || resp.status >= 400) {
+    if ((resp.status < 200 || resp.status >= 400) && resp.status != 404) {
       // bad. 
+      g_log<<Logger::Warning<<"Received unacceptable HTTP status code "<<std::to_string(resp.status)<<" from HTTP endpoint"<<endl;
```
cmouse, line 423: should this be DBException instead?

pieterlexis: hmm, I typed that in, but I _think_ blowing up the connection on a 404 is a bit much.....

cmouse: uhm, the if condition precludes 404, so it's not blowing up the connection.. I am bit puzzled how 404's are treated now.

cmouse: should there be unit/integration test for that?

pieterlexis: > uhm, the if condition precludes 404, so it's not blowing up the connection.. I am bit puzzled how 404's are treated now.

ah yes :)

> should there be unit/integration test for that?

It is implicitly tested, as I got exceptions on 404s before :)

Habbie:
Title says DBException, code says PDNSException, I trust the title is outdated?

modules/remotebackend/httpconnector.cc
Habbie: This looks resolved to me, except for the exception type, right?

pieterlexis:
> Title says DBException, code says PDNSException, I trust the title is outdated?

We use PDNSExceptions to signal problems, when they aren't recoverable by the backend, we throw a DBException toward the Ueberbackend.

Habbie:
> We use PDNSExceptions to signal problems, when they aren't recoverable by the backend, we throw a DBException toward the Ueberbackend.

Oh, I see now that the code does both. Ok!

PR #8955, opened 2020-03-20
On OpenBSD string_view is both in boost and std

omoerbeek:
### Checklist

I have:
- [X] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/ext/lmdb-safe/lmdb-safe.hh
+++ b/ext/lmdb-safe/lmdb-safe.hh
@@ -13,7 +13,7 @@
 #include <algorithm>
 
 // apple compiler somehow has string_view even in c++11!
-#if __cplusplus < 201703L && !defined(__APPLE__)
+#if __cplusplus < 201703L && !defined(__APPLE__) && !defined(__OpenBSD__) && !defined(__FreeBSD__)
 #include <boost/version.hpp>
 #if BOOST_VERSION >= 106100
 #include <boost/utility/string_view.hpp>
--- a/pdns/dnsdistdist/views.hh
+++ b/pdns/dnsdistdist/views.hh
@@ -23,7 +23,7 @@
 #pragma once
 
 // apple compiler somehow has string_view even in c++11!
-#if __cplusplus < 201703L && !defined(__APPLE__) && !defined(__OpenBSD__)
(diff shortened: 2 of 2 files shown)


pdns/dnsdistdist/doh.cc
```diff
@@ -655,15 +655,15 @@ static void doh_dispatch_query(DOHServerConfig* dsc, h2o_handler_t* self, h2o_re
   }
 }
 
-static bool getHTTPHeaderValue(const h2o_req_t* req, const std::string& headerName, string_view& value)
+static bool getHTTPHeaderValue(const h2o_req_t* req, const std::string& headerName, boost::string_view& value)
 {
   bool found = false;
   /* early versions of boost::string_ref didn't have the ability to compare to string */
-  string_view headerNameView(headerName);
+  boost::string_view headerNameView(headerName);
```
rgacogne: Well, we really would like to use `std::string_view` when available.. How does lmdb-safe.hh compiles on OpenBSD? Because the `string_view` detection is a copy-paste from that code..

omoerbeek: I normally don't build lmdb support on OpenBSD. But indeed, it gives similar errors.

RalfvdEnden:
FreeBSD suffers from the same issue. After adding ' && !defined(__FreeBSD)' to the #if line in the patch it builds fine.

Habbie:
NetBSD 9 does not need this.

omoerbeek:
`lmdb-safe.cc` build on OpenBSD now, but warns:

```
  CXX      ext/lmdb-safe/lmdb-safe.o
ext/lmdb-safe/lmdb-safe.cc:338:10: warning: moving a temporary object prevents copy elision [-Wpessimizing-move]
  return std::move(getRWTransaction());
         ^
ext/lmdb-safe/lmdb-safe.cc:338:10: note: remove std::move call here
  return std::move(getRWTransaction());
         ^~~~~~~~~~                  ~
1 warning generated.
```

rgacogne:
> `lmdb-safe.cc` build on OpenBSD now, but warns:
[...]

clang++ does warn on Arch as well. AFAICT the `std::move()` should go as it prevents RVO.

omoerbeek:
OK, IMO we need a separate PR for that.

PR #8960, opened 2020-03-21
Correct typo in Swagger specification

kpfleming:
Found when loading the spec into a validating parser :-)


Habbie:
> Found when loading the spec into a validating parser :-)

Why would you do such a thing! :laughing:

Habbie:
thanks Kevin!

PR #8967, opened 2020-03-23
rec: Implement native DNS64 support, without Lua

rgacogne:
### Short description

Native support is much less flexible than `Lua` hooks but should satisfy most of the DNS64 setups. It is also much faster since it does not involve calling a `Lua` hook for all queries.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/pdns/pdns_recursor.cc
+++ b/pdns/pdns_recursor.cc
@@ -1164,10 +1164,7 @@ int getFakeAAAARecords(const DNSName& qname, ComboAddress prefix, vector<DNSReco
     if (rr.d_type == QType::A && rr.d_place == DNSResourceRecord::ANSWER) {
       if (auto rec = getRR<ARecordContent>(rr)) {
         ComboAddress ipv4(rec->getCA());
-        uint32_t tmp;
-        memcpy(&tmp, &ipv4.sin4.sin_addr.s_addr, 4);
-        // tmp=htonl(tmp);
-        memcpy(((char*)&prefix.sin6.sin6_addr.s6_addr)+12, &tmp, 4);
+        memcpy(&prefix.sin6.sin6_addr.s6_addr[12], &ipv4.sin4.sin_addr.s_addr, sizeof(ipv4.sin4.sin_addr.s_addr));
         rr.d_content = std::make_shared<AAAARecordContent>(prefix);
         rr.d_type = QType::AAAA;
       }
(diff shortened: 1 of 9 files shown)


pdns/pdns_recursor.cc
```diff
@@ -1132,6 +1134,91 @@ int followCNAMERecords(vector<DNSRecord>& ret, const QType& qtype)
   return rcode;
 }
 
+int getFakeAAAARecords(const DNSName& qname, ComboAddress prefix, vector<DNSRecord>& ret)
```
omoerbeek, line 1137: Suggest to use a 'const &' for prefix.

rgacogne: Note that it won't save the copy since we modify it later in the function (I did not change that code, I guess we could).

omoerbeek: Oh, missed that. That code sure looks ugly.

pdns/pdns_recursor.cc
```diff
@@ -1132,6 +1134,91 @@ int followCNAMERecords(vector<DNSRecord>& ret, const QType& qtype)
...
+
+  bool seenA = false;
+  for (DNSRecord& rr : ret) {
+    if (rr.d_type == QType::A && rr.d_place == DNSResourceRecord::ANSWER) {
+      if (auto rec = getRR<ARecordContent>(rr)) {
+        ComboAddress ipv4(rec->getCA());
+        uint32_t tmp;
+        memcpy(&tmp, &ipv4.sin4.sin_addr.s_addr, 4);
+        // tmp=htonl(tmp);
+        memcpy(((char*)&prefix.sin6.sin6_addr.s6_addr)+12, &tmp, 4);
```
omoerbeek, line 1170:
```suggestion
        memcpy(&prefix.sin6.sin6_addr.s6_addr[12], &ipv4.sin4.sin_addr.s_addr, sizeof(ipv4.sin4.sin_addr.s_addr));
```

omoerbeek, line 1170:
```suggestion
        memcpy(&prefix.sin6.sin6_addr.s6_addr[12], &ipv4.sin4.sin_addr.s_addr, sizeof(ipv4.sin4.sin_addr.s_addr));
```

pdns/pdns_recursor.cc
```diff
@@ -4205,6 +4310,20 @@ static int serviceMain(int argc, char*argv[])
   g_proxyProtocolACL.toMasks(::arg()["proxy-protocol-from"]);
   g_proxyProtocolMaximumSize = ::arg().asNum("proxy-protocol-maximum-size");
 
+  if (!::arg()["dns64-prefix"].empty()) {
+    auto dns64Prefix = Netmask(::arg()["dns64-prefix"]);
```
pieterlexis, line 4314: the Netmask ctor can throw a NetmaskException in `makeComboAddress`, please try/catch this to prevent abort()

pdns/pdns_recursor.cc
rgacogne: Done!

PR #8974, opened 2020-03-26
dnsdist: Keep accepting fragmented UDP datagrams on DNSCrypt binds

rgacogne:
### Short description

`DNSCrypt` pads its queries for privacy purposes, and thus requires larger queries than plain DNS ones. Discarding fragmented datagrams doesn't make sense in that case, and actually leads to a very degraded service, as reported in https://github.com/PowerDNS/pdns/pull/7410#issuecomment-604318988.
At the moment I don't see any reason to accept fragmented datagrams for plain DNS queries, so this MR only applies to `DNSCrypt` binds for now.
 
### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/pdns/dnsdist.cc
+++ b/pdns/dnsdist.cc
@@ -1844,14 +1844,15 @@ static void setUpLocalBind(std::unique_ptr<ClientState>& cs)
 #endif
   }
 
-  if (!cs->tcp) {
-    if (cs->local.isIPv4()) {
-      try {
-        setSocketIgnorePMTU(cs->udpFD);
-      }
-      catch(const std::exception& e) {
-        warnlog("Failed to set IP_MTU_DISCOVER on UDP server socket for local address '%s': %s", cs->local.toStringWithPort(), e.what());
-      }
+  /* Only set this on IPv4 UDP sockets.
+     Don't set it for DNSCrypt binds. DNSCrypt pads queries for privacy
+     purposes, so we do receive large, sometimes fragmented datagrams. */
+  if (!cs->tcp && cs->local.isIPv4() && !cs->dnscryptCtx) {
+    try {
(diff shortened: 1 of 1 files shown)


jedisct1:
Hi @rgacogne ,

Packets larger than 1500 bytes are still rejected. Here is the root cause:

https://github.com/PowerDNS/pdns/blob/d93a215ceb0d232746242662da1580b2e1c8d264/pdns/dnsdist.hh#L1147

Packets are rejected in `isUDPQueryAcceptable()`.

rgacogne:
Hi Frank,

That time we are no longer talking about fragmented and reassembled packets, but about packets that are truncated because they are larger than the size of our incoming buffer, or actually than the size we are willing to accept since our actual buffer is a bit larger (to allow addition of EDNS options, handling of responses from the packet cache, ...). We could advertise a larger buffer when DNSCrypt is enabled but it might be a bit more complicated. I'll look into it.

jedisct1:
Hi Remi,

I totally understand that can't just change that constant as it would have implications for non-DNSCrypt users :(

On the bright side, I tested `dnsdist` with an increased incoming buffer size, and this, in combination with your previous change, seems to have fixed all known issues. No more dropped queries.

rgacogne:
I have not tested it yet, but I believe something like this might be enough, it uses the full available size for DNSCrypt binds: https://github.com/rgacogne/pdns/commit/0e214db287e6952b61b78d3e1fd4328403686fc6

jedisct1:
Just tested it, and it totally works!

rgacogne:
Thanks a lot! I just opened a new merge request with this fix, I'll merge it once our CI is green :)

PR #8975, opened 2020-03-26
auth: improve sql schema updates

mind04:
### Upgrade notes

If you are here because our upgrade notes told you to come here, please extract the appropriate `4.3.0-beta2_to_4.3.0_schema.X.sql` file from the FIRST COMMIT and apply it to your database. After that, your schema is identical to the 4.3.0 schema.

### Short description
The colunn order from a new 4.3.0 schema was different from an updated 4.2.0 one. This is no problem for PowerDNS itself, but it will complicate future schema updates. After this pull a new schema and an updated one are identical.

There are also update files to correct the column order for a beta2 schema.

Please do NOT include the beta2 update files in the tar. Picking the wrong update schema will break dnssec signing and most people will only need the release version.

### Checklist
I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


Habbie:
> Please do NOT include the beta2 update files in the tar. Picking the wrong update schema will break dnssec signing and most people will only need the release version.

This means: that this PR does not touch any `Makefile.am` is on purpose. In the upgrade notes, we need to put a notice aimed at ONLY beta2 users, to tell them to grab such a file from our git.

Habbie:
Missing: column reordering for the godbc mssql schema. I don't think we need the whole upgrade-from-beta2 path for those, because of the limited number of users, but we do need to fix the schema and 4.2-4.3 upgrade to put the column in the right spot.

Habbie:
via @mind04, untested fix for the godbc mssql schema:
```
BEGIN;
  ALTER TABLE cryptokeys ADD published BIT DEFAULT TRUE;

  ALTER TABLE cryptokeys ADD COLUMN content_new VARCHAR(MAX)
  UPDATE cryptokeys SET content_new = content;
  ALTER TABLE cryptokeys DROP COLUMN content;
  EXEC sp_RENAME 'cryptokeys.content_new', 'content', 'COLUMN'
COMMIT;
```

tinuzz:
Maybe you could add something in the text that explains which columns are the problem, so in addition to "The colunn order from a new 4.3.0 schema was different from an updated 4.2.0 one.", explain that the order should be 'active' -> 'published' -> 'content, rather than 'active' -> 'content' -> 'published'. It could help people verify whether they're actually doing the correct thing.

Habbie:
This ticket is the text. You have just added that to the text. Thank you :)

tinuzz:
Oh, ok ;-)

PR #8987, opened 2020-03-31
rec: fix referral unit test

omoerbeek:
test_referral_depth is actually hitting the limit in doCNAMECacheCheck() and not the generic depth limit. Use one less of referral to fix that.  Additionally, test and fix the non-limited case.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [X] documented the code

### DIFF
--- a/pdns/recursordist/test-syncres_cc2.cc
+++ b/pdns/recursordist/test-syncres_cc2.cc
@@ -5,7 +5,7 @@
 
 BOOST_AUTO_TEST_SUITE(syncres_cc2)
 
-BOOST_AUTO_TEST_CASE(test_referral_depth)
+void do_test_referral_depth(bool limited)
 {
   std::unique_ptr<SyncRes> sr;
   initSR(sr);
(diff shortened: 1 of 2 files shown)


pdns/recursordist/test-syncres_cc2.cc
```diff
@@ -62,6 +63,22 @@ BOOST_AUTO_TEST_CASE(test_referral_depth)
...
+    BOOST_CHECK(e.reason.find("max-recursion-depth") != string::npos);
+  }
+
+  // Then check if the setup with high limit is OK.
+  SyncRes::s_maxdepth = 50;
+  try {
+    vector<DNSRecord> ret;
+    int rcode = sr->beginResolve(target, QType(QType::A), QClass::IN, ret);
+    BOOST_CHECK_EQUAL(rcode, RCode::NoError);
+    BOOST_CHECK_EQUAL(ret.size(), 1U);
```
rgacogne, line 75: This should be a `require` instead of a `check` so we don't read out-of-bounds just after if the test fails:
```suggestion
    BOOST_REQUIRE_EQUAL(ret.size(), 1U);
```

pdns/recursordist/test-syncres_cc2.cc
```diff
@@ -62,6 +63,22 @@ BOOST_AUTO_TEST_CASE(test_referral_depth)
     BOOST_CHECK(false);
   }
   catch (const ImmediateServFailException& e) {
+    BOOST_CHECK(e.reason.find("max-recursion-depth") != string::npos);
+  }
+
+  // Then check if the setup with high limit is OK.
```
rgacogne, line 69: It doesn't seem right to do a new test without clearing all the caches (records, EDNS statuses, etc..) which would be painful, perhaps this should be a separate test case instead, something like `BOOST_AUTO_TEST_CASE(test_referral_depth_ok)`?

rgacogne:
Perhaps we could also apply this change in `SyncRes::doResolveNoQNameMinimization`?:
```diff
-  if(s_maxdepth && depth > s_maxdepth)
+  if (s_maxdepth && depth > s_maxdepth) {
+    LOG(prefix<<qname<<": More than "<<std::to_string(s_maxdepth)<<" levels of recursion (max-recursion-depth) needed while resolving ("<<std::to_string(depth)<<")"<<endl);
     throw ImmediateServFailException("More than "+std::to_string(s_maxdepth)+" (max-recursion-depth) levels of recursion needed while resolving "+qname.toLogString());
+  }
```

pdns/recursordist/test-syncres_cc2.cc
omoerbeek: I hit it in the (fixed) positive test case.

rgacogne:
CI failure is unrelated (ALIAS).

PR #8993, opened 2020-04-02
Skip EDNS Cookies in the packet cache

### LINKED ISSUE #5131: dnsdist, auth, rec: EDNS COOKIE option busts packet cache?
Hi,

I noticed that with a packet cache enabled:

```lua
pc = newPacketCache(10000, 86400, 0, 60, 60)
getPool(""):setCache(pc)
```

When I queried this dnsdist server with dig 9.11, the result was always a cache miss, unless I specified `+nocookie` to dig. It looks like dnsdist doesn't support the EDNS COOKIE option, so a query with a COOKIE will always bust the packet cache.

This is with the dnsdist Debian package version 1.1.0-2.

rgacogne:
### Short description

This PR changes how our packet caches operate with regard to EDNS Cookies:
- the authoritative server's one will skip the content of EDNS cookies, if any, potentially serving an answer from the packet cache for a query with a different cookie than the one that caused the answer to be generated. This is fine since the authoritative server doesn't include EDNS cookies in its responses. EDNS Client Subnet values are still hashed given that they might be used to provide the answer ;
- the dnsdist's one will skip the content of EDNS cookies as well by default, but this behaviour can be changed by passing `cookieHashing=true` to the `newPacketCache()` function. Doing so is needed if at least one of the backend return responses containing EDNS cookies. EDNS Client Subnet values are still hashed ;
- the recursor's one will skip the content of EDNS cookies and EDNS Client Subnet.

This PR could use a serious review (or several :)) and should not be merged lightly!

Fixes #5131.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code


pdns/ednsoptions.cc
```diff
@@ -23,6 +23,23 @@
...
+    return false;
+  }
+
+  size_t pos = 0;
+
+  optionCode = (static_cast<unsigned char>(data[pos]) * 256) + static_cast<unsigned char>(data[pos + 1]);
+  pos += EDNS_OPTION_CODE_SIZE;
+
+  optionLen = (static_cast<unsigned char>(data[pos]) * 256) + static_cast<unsigned char>(data[pos + 1]);
+  pos += EDNS_OPTION_LENGTH_SIZE;
```
omoerbeek, line 38: I prefer to have the conversion to uint16_t explicit:

```suggestion
  const uint8_t* p = reinterpret_cast<const uint8_t*>(data);
  optionCode = (static_cast<uint16_t>(p[pos]) * 256) + p[pos + 1];
  pos += EDNS_OPTION_CODE_SIZE;

  optionLen = (static_cast<uint16_t>(p[pos]) * 256) + p[pos + 1];
  pos += EDNS_OPTION_LENGTH_SIZE;

 ```

File: pdns/packetcache.hh [hunk 2]
```diff
@@ -23,80 +23,137 @@
...
-      int res = getEDNSOption(const_cast<char*>(reinterpret_cast<const char*>(p)) + 14, end - (p + 14), EDNSOptionCode::ECS, &optionBegin, &optionLen);
-      if (res == 0) {
-        skipBegin = optionBegin;
-        skipEnd = optionBegin + optionLen;
-        if (ecsBegin != nullptr && ecsEnd != nullptr) {
-          *ecsBegin = optionBegin - packet.c_str();
-          *ecsEnd = *ecsBegin + optionLen;
-        }
+    const struct dnsheader* dh = reinterpret_cast<const struct dnsheader*>(packet.data());
+    if (ntohs(dh->qdcount) != 1 || ntohs(dh->ancount) != 0 || ntohs(dh->nscount) != 0 || ntohs(dh->arcount) != 1 || (pos + 15) >= packetSize) {
```
omoerbeek, line 49: You could choose to `htons()` the right hand side of the compares here, since they all test for inequality.

pdns/packetcache.hh
```diff
@@ -23,80 +23,137 @@
...
-    if (skipBegin > p) {
-      ret = burtle(reinterpret_cast<const unsigned char*>(p), skipBegin-p, ret);
+
+    currentHash = burtle(reinterpret_cast<const unsigned char*>(&packet.at(pos)), 15, currentHash);
+    /* skip the qtype (2), qclass (2) */
+    /* root label (1), type (2), class (2) and ttl (4) */
+    /* already hashed above */
+    pos += 13;
+
+    const uint16_t rdLen = ((static_cast<unsigned char>(packet.at(pos)) * 256) + static_cast<unsigned char>(packet.at(pos + 1)));
```
omoerbeek, line 62: Some suggestion as wrt `ednsoptions.cc`: make the cast to `uint16_t` explicit.

pdns/packetcache.hh
```diff
@@ -23,80 +23,137 @@
...
+    return currentHash;
   }
 
-  static uint32_t canHashPacket(const std::string& packet)
+  static uint32_t hashHeaderAndQName(const std::string& packet, size_t& pos)
   {
-    uint32_t ret = 0;
-    ret = burtle(reinterpret_cast<const unsigned char*>(packet.c_str()) + 2, sizeof(dnsheader) - 2, ret); // rest of dnsheader, skip id
+    uint32_t currentHash = 0;
     size_t packetSize = packet.size();
```
omoerbeek, line 119:
```suggestion
    const size_t packetSize = packet.size();
```

pdns/packetcache.hh
```diff
@@ -23,80 +23,137 @@
...
+      }
+      else if (optionCode == EDNSOptionCode::ECS) {
+        if (skipECS) {
+          skip = true;
+        }
+      }
+
+      if (!skip) {
+        /* hash the option code, length and content */
+        currentHash = burtle(reinterpret_cast<const unsigned char*>(&packet.at(pos)), 4 + optionLen, currentHash);
```
omoerbeek, line 98: An option having the same opcode but different length does not hash the same.
If options are variable length, this would miss some cache hits. But I guess that's ok, since the equality test also does considers them equal. Actually, do variable length options exist?

pdns/packetcache.hh
rgacogne: I'd really prefer to keep it that way as I believe it's more easy to read than the alternative you suggest. Even if the compiler were somehow not smart enough to optimize it, I'm sure we can live with it.

pdns/packetcache.hh
rgacogne: Yes, it was intended for options with the same opcode but different lengths to compare differently, as the content simply can't be the same. That might change, however, if/when we decide to skip EDNS padding in queries. I guess we would need to special case the EDNS padding option to skip the size of the content.

rgacogne: I wonder if it would make sense to skip the length of all options right now. It would make sense for EDNS padding and would also mean that we would fully ignore the content of EDNS Client Subnet options, including the size of the source, which seems like a win when we really do want to ignore ECS.

rgacogne: That might actually be an issue for ECS source zero [1], which is a way for the client to ask that no EDNS value be sent to authoritative servers. So we don't want to ignore that one, I guess..

[1]: https://tools.ietf.org/html/rfc7871#section-7.1.2

rgacogne:
Pushed a few changes:
- Making conversions to uint16_t explicit, as suggested by Otto ;
- Skipping EDNS padding in addition to EDNS cookies in the auth and rec ;
- Skipping EDNS padding, still optionally skipping EDNS cookies in dnsdist ;
- Skipping the size of the rdata length for the OPT record, as well as the size of some padding options (padding, cookies). The size of EDNS Client Subnet options is not ignored as we don't want to confuse an EDNS source zero with a different one.

jsoref:
Hi, for the time being (and hopefully going forward), pdns master is running the [check-spelling](https://github.com/marketplace/actions/check-spelling) action. In order to avoid a ❌ when merged https://github.com/jsoref/pdns/commit/03669c04901f518b72cdb7b4ace867451ab57fae#commitcomment-39229219, please see https://github.com/jsoref/pdns/commit/03669c04901f518b72cdb7b4ace867451ab57fae#commitcomment-39229308 -- you'll need to merge `master` to get the check (and the files).

rgacogne:
Rebased to fix conflicts.

rgacogne:
Rebased to fix a conflict.

rgacogne:
I'll try to do more performance testing of this PR then the numbers look good I'd like to get it merged, since it might have a very nice impact on our packet cache hit ratio.

rgacogne:
I've done a fair number of tests, and for all products there does not seem to be any measurable slowdown when there is no EDNS options present (or no EDNS at all). As expected there is a small slowdown when EDNS options are present (it doesn't seem noticeable in the rec when only one option is present, since we were already looking for ECS), I measured around 7% when three EDNS options are present. On dnsdist this is completely eclipsed by the gains from #9420. 
In my opinion this slowdown is worth it based on the much better cache hit ratio that skipping Cookies (and perhaps padding later?) would get us for these queries :-)

pdns/packetcache.hh
```diff
@@ -23,80 +23,137 @@
...
     }
-    if (skipEnd < end) {
-      ret = burtle(reinterpret_cast<const unsigned char*>(skipEnd), end-skipEnd, ret);
+
+    return currentHash;
+  }
+
+  static uint32_t hashHeaderAndQName(const std::string& packet, size_t& pos)
+  {
+    uint32_t currentHash = 0;
```
omoerbeek, line 118: redundant init, suggest to move the declaration down

omoerbeek (approved):
I did a review back in April and did a quick read again today. After a rebase it should be good.

rgacogne:
Thanks Otto! I guess I'll merge this PR once an "auth-4.4.x" branch has been created, then :-)

Habbie:
> the authoritative server's one will skip the content of EDNS cookies, if any, potentially serving an answer from the packet cache for a query with a different cookie than the one that caused the answer to be generated. This is fine since the authoritative server doesn't include EDNS cookies in its responses. EDNS Client Subnet values are still hashed given that they might be used to provide the answer ;

I am fine with this for auth 4.4.0. I did not review the code, though.

PR #8999, opened 2020-04-05
Add error response objects to all API operations in spec file

kpfleming:
This patch documents the possible error response objects for all API
operations. When a spec-validating client is used, returning a
response code (and object) which are not included in the spec will
disallow access to the response object, and may cause an exception to
be thrown.

In order to do this without repeating the possible responses for each
API operation, the YAML anchor/alias mechanism is used so that the
list of response codes and their object schemas can be written in one
place. To make that possible, the patch adds a fictitious '/error' API
operation, which of course will actually return an error since it does
not exist.

The spec version has been bumped from 0.0.13 to 0.0.15 (0.0.14 should
have been used for the recent bug fixes to the spec file).

Note: If the spec file was upgraded to OpenAPI 3.0.0, instead of
Swagger 2.0, the response codes could be specified as '4XX' and '5XX',
which would cover all possible response codes in those ranges without
having to audit the code for error responses.

Signed-off-by: Kevin P. Fleming <kevin@km6g.us>

### DIFF
--- a/.github/actions/spell-check/excludes.txt
+++ b/.github/actions/spell-check/excludes.txt
@@ -3,6 +3,7 @@
 (?:^|/)go\.sum$
 (?:^|/)m4/
 (?:^|/)package-lock\.json$
+(?:^|/)requirements\.txt$
 /expected_result
 /test-dnsrecords_cc\.cc$
 SUMS$
--- a/docs/http-api/swagger/authoritative-api-swagger.yaml
+++ b/docs/http-api/swagger/authoritative-api-swagger.yaml
@@ -1,6 +1,6 @@
 swagger: '2.0'
 info:
-  version: "0.0.13"
+  version: "0.0.15"
   title: PowerDNS Authoritative HTTP API
   license:
     name: MIT
@@ -23,6 +23,28 @@ security:
 #       We need to look at the code and figure out the default HTTP response
 #       codes and adjust docs accordingly.
 paths:
+  '/error':
+    get:
+      summary: Will always generate an error
+      operationId: error
+      responses: &commonErrors
+        '400':
+          description: The supplied request was not valid
+          schema:
+            $ref: '#/definitions/Error'
+        '404':
+          description: Requested item was not found
+          schema:
+            $ref: '#/definitions/Error'
+        '422':
+          description: The input to the operation was not valid
+          schema:
+            $ref: '#/definitions/Error'
+        '500':
+          description: Internal server error
+          schema:
+            $ref: '#/definitions/Error'
+
   '/servers':
     get:
       summary: List all servers
@@ -36,6 +58,7 @@ paths:
             type: array
             items:
               $ref: '#/definitions/Server'
+        <<: *commonErrors
 
   '/servers/{server_id}':
     get:
@@ -54,6 +77,7 @@ paths:
           description: An server
           schema:
             $ref: '#/definitions/Server'
+        <<: *commonErrors
 
   '/servers/{server_id}/cache/flush':
     put:
@@ -77,6 +101,7 @@ paths:
           description: Flush successful
           schema:
             $ref: '#/definitions/CacheFlushResult'
+        <<: *commonErrors
 
   '/servers/{server_id}/zones':
     get:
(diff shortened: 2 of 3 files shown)


kpfleming:
I have confirmed that the YAML->JSON conversion using the Python standard library modules supports the anchor/alias mechanism; I've used the JSONified version of this new spec file in tests with my Ansible modules and it works fine.

kpfleming:
The current version of sphinxcontrib-openapi (and all previous versions) do *not* work with YAML merge keys; while that will be fixed in a release in the near future, the upstream version doesn't support the 'definitions' tag used in the RST files here.

I've included one option for fixing this. Another option is to upgrade to Python 3 for building the docs, and removing the OrderedLoader from the version of sphinxcontrib-openapi (an old pinned fork) since it will no longer be necessary. A third option would be to maintain the docs in a new file and have the 'canonical' YAML file produced from it when required.

Habbie:
> Another option is to upgrade to Python 3 for building the docs

This!

kpfleming:
I'll pick this up again and require Python 3 for all of the build process steps which use Python.

kpfleming:
I've got a branch in the works for the Python 3 upgrade, which I'll send  a PR for once #8911 has been merged (since that branch also modifies the usage of Python in the build system). I've already gotten it working properly for local builds (both from repo clones and from source tarballs), but haven't tackled the various CI configurations or the builder-support Dockerfiles yet. More details on that once I open that PR.

kpfleming:
With the merge of the use-python3 branch I've updated this branch, but it's still an option to simplify the sphinxcontrib-openapi fork and remove the YAML-to-YAML conversion in this branch.

Habbie:
Fixing the fork makes sense to me - @pieterlexis?

pieterlexis:
yes, let's do that!

kpfleming:
Alright, I'll fork it and prepare a PR. Should be pretty easy to do.

kpfleming:
PR opened: https://github.com/pieterlexis/sphinxcontrib-openapi/pull/1

With those changes, no workarounds are necessary for Sphinx to handle the merge keys in the YAML spec file. I've built the PDF and HTML docs with that branch and the changes here and the results are as expected (the error responses show up each for each API operation as they should).

kpfleming:
One final thought: since the API URL refers to this API as 'v1', the version number in the spec file should probably use '1' as its major version, instead of '0'.

kpfleming:
If someone can poke the Travis-CI and Circle-CI jobs, they should pass now.

Habbie:
https://github.com/pieterlexis/sphinxcontrib-openapi/pull/1#issuecomment-730087209

> Merging this will 'break' existing pdns build trees, since the order will not be preserved. Creating a new branch name might be an easy way to fix that (use-jsondomain-py3 would be logical).

only master I hope? if not, I'd really like the two-branch solution proposed in that PR.

kpfleming:
I was wrong, there; the use-jsondomain branch now requires Python 3.6, so if *any* pdns branch is used to build docs, it will require Python 3.6 or higher, and the results will be correct. Those branches use 'virtualenv' without specifying a Python version, though, so this won't be automatic.

I think the two-branch solution is probably safer.

Habbie:
@pieterlexis can you make a new branch with Kevin's PR on it and roll the current branch back?

pieterlexis:
done, the use-jsondomain branch is rolled back. I have created a use-jsondomain-kpfleming branch with the changes from @kpfleming

kpfleming:
Thanks. I'll update this PR with that branch name; once it is ready for merging you can choose a less personalized name for the branch :-)

kpfleming:
All tests passed except the spell check, and I assume we're not going to add 'kpfleming' to the dictionary :-)

Habbie:
We have plenty of names in the spellcheck exceptions, but in this case the exception should be all of requirements.txt :D

Habbie:
If we can rename the branch on https://github.com/pieterlexis/sphinxcontrib-openapi.git to something else than '-kpfleming', we can merge this.

pieterlexis:
Renamed the branch to use-jsondomain-pdns

Habbie:
@kpfleming can you update the branch name in this PR? Then we can merge it!

kpfleming:
Changed, rebased on current 'master', squashed too.

PR #9007, opened 2020-04-06
auth: Do an ANY lookup for all types then filter

rgacogne:
### Short description

Most of our backends have a very high latency, meaning that it takes a long time to send a query and get the answer, regardless of whether we are asking for one type or several. Our code base often asks for a specific type, and our current code stores separately the answers for ANY queries and the ones for a specific type. This seems wasteful since the answer to an ANY query already contains the records for a more specific one, and our in-memory records cache is must faster than going to the backend. We could save a round-trip by looking for ANY answers when we don't find a specific one in the record cache, but our first query is often for the specific NS type because we are looking for a referral.
This PR converts all lookups to an ANY lookup instead, making sure that we fill the cache as fast as possible to save round-trips to the backend later.

My tests showed roughly one third less queries to the backend in simple cases, and probably more in DNSSEC cases, while achieving higher QPS boundaries (~ +30%). CPU usage is also significantly reduced while replying a real-world PCAP.
The number of entries in the records cache is also significantly lower since we don't need to store a record twice, for ANY and for the exact type itself.

We could easily enable that change for specific backends only if we believe it might have a negative effect on some of them, although testing with the bind backend showed a slight improvement there as well, even though lookups in the bind backend are already quite fast.
I have not tested LMDB.

We could reduce the number of round-trips to the backend even more by getting rid of the 'SOA' special case, since I'm not aware of any backend currently implementing it in a special way.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/pdns/ueberbackend.cc
+++ b/pdns/ueberbackend.cc
@@ -54,6 +54,7 @@ std::mutex UeberBackend::instances_lock;
 
 // initially we are blocked
 bool UeberBackend::d_go = false;
+bool UeberBackend::s_doANYLookupsOnly = false;
 std::mutex UeberBackend::d_mut;
 std::condition_variable UeberBackend::d_cond;
 AtomicCounter* UeberBackend::s_backendQueries = nullptr;
@@ -97,6 +98,11 @@ void UeberBackend::go(void)
 {
   S.declare("backend-queries", "Number of queries sent to the backend(s)");
   s_backendQueries = S.getPointer("backend-queries");
+
+  if (::arg().mustDo("any-lookups-only")) {
+    s_doANYLookupsOnly = true;
+  }
+
   {
     std::unique_lock<std::mutex> l(d_mut);
     d_go = true;
@@ -518,31 +524,34 @@ int UeberBackend::cacheHas(const Question &q, vector<DNSZoneRecord> &rrs)
   rrs.clear();
   //  g_log<<Logger::Warning<<"looking up: '"<<q.qname<<"'|N|"<<q.qtype.getName()<<"|"<<q.zoneId<<endl;
 
-  std::vector<DNSZoneRecord> anyRecs;
-  bool ret = QC.getEntry(q.qname, QType::ANY, anyRecs, q.zoneId);
+  bool ret = false;
+  if (s_doANYLookupsOnly) {
+    std::vector<DNSZoneRecord> anyRecs;
+    ret = QC.getEntry(q.qname, QType::ANY, anyRecs, q.zoneId);
 
-  if (ret) {
-    if (anyRecs.empty()) {// negatively cached
-      return 0;
-    }
-
-    for (auto& rec : anyRecs) {
-      if (q.qtype.getCode() == QType::ANY || rec.dr.d_type == q.qtype.getCode()) {
-        rrs.push_back(std::move(rec));
+    if (ret) {
+      if (anyRecs.empty()) {// negatively cached
+        return 0;
       }
-    }
 
-    if (rrs.empty()) {
-      // we know there is no record of this type */
-      return 0;
-    }
+      for (auto& rec : anyRecs) {
+        if (q.qtype.getCode() == QType::ANY || rec.dr.d_type == q.qtype.getCode()) {
+          rrs.push_back(std::move(rec));
+        }
+      }
 
-    return 1;
+      if (rrs.empty()) {
+        // we know there is no record of this type */
+        return 0;
+      }
+
+      return 1;
+    }
   }
 
   // miss for ANY, see if by any chance we have a cached entry for the exact type (mostly SOA)
   // or a negative cache entry (all types)
-  if (q.qtype != QType::ANY) {
+  if (!s_doANYLookupsOnly || q.qtype != QType::ANY) {
     ret = QC.getEntry(q.qname, q.qtype, rrs, q.zoneId);
     if (ret) {
       if (rrs.empty()) {
@@ -619,7 +628,7 @@ void UeberBackend::lookup(const QType &qtype,const DNSName &qname, int zoneId, D
   d_domain_id = zoneId;
 
   d_handle.i = 0;
-  d_handle.qtype = QType::ANY;
+  d_handle.qtype = s_doANYLookupsOnly ? QType::ANY : qtype;
   d_handle.qname = qname;
   d_handle.pkt_p = pkt_p;
   d_ancount = 0;
@@ -640,7 +649,7 @@ void UeberBackend::lookup(const QType &qtype,const DNSName &qname, int zoneId, D
       d_negcached = d_cached = false;
       d_answers.clear();
       d_handle.d_hinterBackend = backends.at(d_handle.i++);
-      d_handle.d_hinterBackend->lookup(QType::ANY, qname, zoneId, pkt_p);
+      d_handle.d_hinterBackend->lookup(d_handle.qtype, qname, zoneId, pkt_p);
       ++(*s_backendQueries);
     } 
     else if (cstat == 0) {
@@ -698,12 +707,13 @@ bool UeberBackend::get(DNSZoneRecord &rr)
     }
 
     ++d_anyCount;
+
     d_answers.push_back(std::move(anyRecord));
   }
 
   if (!gotRecord) {
     // cout<<"end of ueberbackend get, seeing if we should cache"<<endl;
-    if (d_anyCount == 0 && d_handle.qname.countLabels()) {
+    if (s_doANYLookupsOnly && d_anyCount == 0 && d_handle.qname.countLabels()) {
       /* we can negcache the whole name */
       // cerr<<"we can negcache the whole name"<<endl;
       addNegCache(d_question, QType::ANY);
@@ -714,8 +724,13 @@ bool UeberBackend::get(DNSZoneRecord &rr)
       addNegCache(d_question, d_question.qtype);
     }
 
-    if (d_anyCount) {
-      addCache(d_question, QType::ANY, std::move(d_answers));
+    if (s_doANYLookupsOnly) {
+      if (d_anyCount) {
+        addCache(d_question, QType::ANY, std::move(d_answers));
+      }
+    }
+    else if (d_ancount > 0) {
+      addCache(d_question, d_question.qtype, std::move(d_answers));
     }
 
     d_answers.clear();
--- a/pdns/ueberbackend.cc
+++ b/pdns/ueberbackend.cc
@@ -53,9 +53,10 @@ vector<UeberBackend *>UeberBackend::instances;
 std::mutex UeberBackend::instances_lock;
 
 // initially we are blocked
-bool UeberBackend::d_go=false;
+bool UeberBackend::d_go = false;
 std::mutex UeberBackend::d_mut;
 std::condition_variable UeberBackend::d_cond;
+AtomicCounter* UeberBackend::s_backendQueries = nullptr;
 
 //! Loads a module and reports it to all UeberBackend threads
 bool UeberBackend::loadmodule(const string &name)
@@ -94,6 +95,8 @@ bool UeberBackend::loadModules(const vector<string>& modules, const string& path
 
 void UeberBackend::go(void)
 {
+  S.declare("backend-queries", "Number of queries sent to the backend(s)");
+  s_backendQueries = S.getPointer("backend-queries");
   {
     std::unique_lock<std::mutex> l(d_mut);
     d_go = true;
@@ -638,6 +641,7 @@ void UeberBackend::lookup(const QType &qtype,const DNSName &qname, int zoneId, D
       d_answers.clear();
       d_handle.d_hinterBackend = backends.at(d_handle.i++);
       d_handle.d_hinterBackend->lookup(QType::ANY, qname, zoneId, pkt_p);
+      ++(*s_backendQueries);
     } 
     else if (cstat == 0) {
       //      cout<<"UeberBackend::lookup("<<qname<<"|"<<DNSRecordContent::NumberToType(qtype.getCode())<<"): NEGcached"<<endl;
@@ -761,6 +765,7 @@ bool UeberBackend::handle::get(DNSZoneRecord &r)
       
       d_hinterBackend = parent->backends.at(i++);
       d_hinterBackend->lookup(qtype, qname, parent->d_domain_id, pkt_p);
+      ++(*s_backendQueries);
     }
     else {
       break;
--- a/pdns/ueberbackend.cc
+++ b/pdns/ueberbackend.cc
@@ -368,8 +368,8 @@ bool UeberBackend::getAuth(const DNSName &target, const QType& qtype, SOAData* s
       if(i == backends.end()) {
         if(d_negcache_ttl) {
           DLOG(g_log<<Logger::Error<<"add neg cache entry:"<<shorter<<endl);
-          d_question.qname=shorter;
-          addNegCache(d_question);
+          d_question.qname = shorter;
+          addNegCache(d_question, d_question.qtype);
         }
         continue;
       } else if(d_cache_ttl) {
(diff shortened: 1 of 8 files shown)


Habbie:
Our metronome templates currently use `query-cache-miss / udp-answers` to calculate 'query to DB amplification', is your `backend-queries` metric very different?

rgacogne:
> Our metronome templates currently use query-cache-miss / udp-answers to calculate 'query to DB amplification', is your backend-queries metric very different?

Yes, if only because this PR does currently up to two lookups into the records cache (ANY then specific type) because of the SOA case, but only one will go to the backend. The 'uncached' SOA query is not accounted for in `query-cache-miss`. It also differs if you have more than one backend. There might be other cases as well.

rgacogne:
Failure of ci/circleci: test-auth-regress-gpgsql is unrelated (ALIAS).

rgacogne:
Failure of test-auth-regress-gsqlite3 this time:
```
Apr 06 14:11:39 Error resolving for ALIAS google-public-dns-a.google.com., aborting AXFR
Apr 06 14:11:39 Unable to AXFR zone 'example.com' from remote '127.0.0.1:5200' (resolver): AXFR chunk error: Server Failure (This was the first time. Excluding zone from slave-checks until 1586182599)
Apr 06 14:11:39 Signing thread died because of std::exception: Reading from socket in Signing Pipe loop: Connection reset by peer
```

Habbie:
Ugh, ALIAS was quite stable in CircleCI for a while. I'll see if I can make the ALIAS target something local.

rgacogne:
> We could reduce the number of round-trips to the backend even more by getting rid of the 'SOA' special case, since I'm not aware of any backend currently implementing it in a special way.

I looked into that idea a bit, but it's complicated by the fact that some backends (bind and `LMDB`, as of today) return their best `SOA` when the `qtype` passed to the `::lookup()` operation is `SOA`. I'm not sure how useful that optimization is today (I'm pretty sure it's useless for the bind backend since the round-trip to the `Ueberbackend` is a function call, for `LMDB` it depends on the cost of opening a RO transaction) but we might not want to lose it.

Habbie:
That optimisation has been requested for the pipebackend, I did not check if we do that today.

rgacogne:
In fact it looks like this optimization is gone in both the Bind and LMDB backends since https://github.com/PowerDNS/pdns/commit/6678e5ac11362caedebfdb0e9049c94f4a248a1a (#8050).

rgacogne:
The more I look into this "optimization" the less I'm actually convinced it ever worked that way in the bind backend.

rgacogne:
Unless I'm missing something huge, none of the backends I know of (including the not merged DLSO and Cassandra backends) overrides `DNSBackend::getAuth()`. `DNSBackend::getAuth()` directly calls `DNSBackend::getSOA()`, which is not overridden either. `DNSBackend::getSOA()` does not support sending a higher SOA than the one asked for, and it actually forces the qname to the requested one so sending a higher SOA would not be ignored and cause issues.

pdns/ueberbackend.cc
```diff
@@ -502,48 +502,81 @@ int UeberBackend::cacheHas(const Question &q, vector<DNSZoneRecord> &rrs)
...
+  bool ret = QC.getEntry(q.qname, QType::ANY, anyRecs, q.zoneId);
+
+  if (ret) {
+    if (anyRecs.empty()) {// negatively cached
+      return 0;
+    }
+
+    for (auto& rec : anyRecs) {
+      if (q.qtype.getCode() == QType::ANY || rec.dr.d_type == q.qtype.getCode()) {
+        rrs.push_back(std::move(rec));
```
cmouse, line 522: emplace_back?

rgacogne: Won't the move-constructor be called in the exact same way in both cases?

cmouse: possibly. i just wonder if it could be rrs.emplace_back(rec)?

rgacogne: That would allocate and copy two `DNSName`s and one `shared_ptr<DNSRecordContent>` instead of moving them, so I don't think that would be better.

rgacogne:
After discussing it ith @mind04 on the IRC channel, this might cause issues with some setups involving multiple backends. Unless we are willing to take the risk of breaking some setups, we should either limit this new behaviour to single-backend setups (easiest) or make it configurable (enable this optimization for most multiple-backends setups while not breaking the other ones).

zeha:
If I read the diff right, only setups that have the same name in multiple backends could be affected?
If this is true, and if delegations from one backend to another work - great, and those mystical people that do other weird shit need to step out of the shadows and explain what they are doing.

rgacogne:
Yes, my worry is that it might be more common that we think for custom backends, in order not to implement DNSSEC there and delegate that to a second backend instead. That's quite ugly in my humble opinion since it requires duplicating the names and types in the second backend, but it clearly exists..

zeha:
My Problem with this is: there are only rumors of this existing, and nobody can say if this would be a supported thing to do or not. The mere rumor of it existing prevents making stuff better for everyone else.
If this is a usecase, maybe something actually well-supported should be implemented instead.
However my fear is, until we actually break these setups, no code will be written to make anything better - nothing will ever change.

mind04:
Limiting this behaviour to singe backend setups is a nice middle ground. And maybe add a configuration option to enable this for multi backed setups as well. 
This will make it better for most people, without breaking more complex multi backend setups.

klaus3000:
I just wonder if this approach can be extend to fetch all RRs of a zone from the backend and filtering the label afterwards. Then all RRs of a zone can be put in the query cache. And if PDNS knows know that the query cache either contains all or no RRs of a zone, it could defeat random subdomain attacks as the NXDOMAIN is a result of not finding the requested label in the query cache, with having to ask the backend for every query.

rgacogne:
> I just wonder if this approach can be extend to fetch all RRs of a zone from the backend and filtering the label afterwards. Then all RRs of a zone can be put in the query cache. And if PDNS knows know that the query cache either contains all or no RRs of a zone, it could defeat random subdomain attacks as the NXDOMAIN is a result of not finding the requested label in the query cache, with having to ask the backend for every query.

There are several issues with that idea, mostly because the underlying database might be updated, and records might have different TTLs which might mean that we would need to invalidate the whole records cache after the shortest TTL expires.
I think the closest that we can get to that would be:
- an in-memory database like the bind or LMDB backends ;
- a (partial?) implementation of rfc8198 in the auth.

klaus3000:
> > I just wonder if this approach can be extend to fetch all RRs of a zone from the backend and filtering the label afterwards. Then all RRs of a zone can be put in the query cache. And if PDNS knows know that the query cache either contains all or no RRs of a zone, it could defeat random subdomain attacks as the NXDOMAIN is a result of not finding the requested label in the query cache, with having to ask the backend for every query.
> 
> There are several issues with that idea, mostly because the underlying database might be updated, and records might have different TTLs which might mean that we would need to invalidate the whole records cache after the shortest TTL expires.

Since when does PDNS Auth respect TTLs for packet/query cache? Actually I think that TTLs should be be considered for these caches. TLLs are designed for recursive DNS. Cache policies of the Auth are a decision of the DNS operator - regardless what the TTL says.

> I think the closest that we can get to that would be:
> 
> * an in-memory database like the bind or LMDB backends ;
Lacks replication. I do not know LMDB details, but SOA checks and AXFR does not scale to millions of zones and hundred secondaries.

> * a (partial?) implementation of rfc8198 in the auth.
Sounds good. But - when you have a random subdomain attack, within a second the aggressiveNSEC(3) cache is filled with the complete zone - so not much difference then loading the whole zone.

Anyways agressiveNSEC(3) caching sounds useful also to reduce the backend queries also without random subdomain attacks. Of course the technique should also be used for queries without the +DO bit set, to improve also non-DNSSEC queries.

rgacogne:
> Since when does PDNS Auth respect TTLs for packet/query cache? 

We have capped the TTD of the packet and the query caches with the lowest TTL for as long as I can remember.

> Actually I think that TTLs should NOT be considered for these caches. TTLs are designed for recursive DNS. Cache policies of the Auth are a decision of the DNS operator - regardless what the TTL says.

I would agree with you if all backends could properly detect a change in the zone, which is not the case for database backends. That means we need to keep the TTD duration in the caches quite small, which reduces the effect of loading the whole zone in memory a lot.
Note that it might not be possible to load everything in for quite some setups anyway, because they have a huge number of very large zones that would not fit in memory at the same time. 

> Sounds good. But - when you have a random subdomain attack, within a second the aggressiveNSEC(3) cache is filled with the complete zone - so not much difference then loading the whole zone.

The SOA, NSEC and corresponding RRSIG records are loaded, not all the records. This might make a significant difference.

> Anyways agressiveNSEC(3) caching sounds useful also to reduce the backend queries also without random subdomain attacks. Of course the technique should also be used for queries without the +DO bit set, to improve also non-DNSSEC queries.

Agreed.

klaus3000:
> > Since when does PDNS Auth respect TTLs for packet/query cache?
> 
> We have capped the TTD of the packet and the query caches with the lowest TTL for as long as I can remember.

I was not aware of this. It should be mentioned in the docs.

> > Actually I think that TTLs should NOT be considered for these caches. TTLs are designed for recursive DNS. Cache policies of the Auth are a decision of the DNS operator - regardless what the TTL says.
> 
> I would agree with you if all backends could properly detect a change in the zone, which is not the case for database backends. That means we need to keep the TTD duration in the caches quite small, which reduces the effect of loading the whole zone in memory a lot.

I use a database backend. Actually, when I as DNS operator sets a query cache of X, but my customer sets a TTL of Y<X, the PDNS ignores the Admin and follows the user. IMO not fine. So, a TTL should be that TTL regardless of record values - IMO.

> Note that it might not be possible to load everything in for quite some setups anyway, because they have a huge number of very large zones that would not fit in memory at the same time.

OF course it should be a config option. I guess for fast backends which do not suffer during random subdomain attacks there is no need to cache the whole zone as the whole zone is already in memory. And if someone with hughe zones uses a DB backend than it probably will not use this feature.

> > Sounds good. But - when you have a random subdomain attack, within a second the aggressiveNSEC(3) cache is filled with the complete zone - so not much difference then loading the whole zone.
> 
> The SOA, NSEC and corresponding RRSIG records are loaded, not all the records. This might make a significant difference.

In my experience the size of "normal" RRs can be neglected compared to NSEC(3) and RRSIG RRs.

> > Anyways agressiveNSEC(3) caching sounds useful also to reduce the backend queries also without random subdomain attacks. Of course the technique should also be used for queries without the +DO bit set, to improve also non-DNSSEC queries.
> 
> Agreed.

Of course there always exists a certain usecase were things may get worse than better. Hence, such features should always be a config option for power users.

rgacogne:
> Limiting this behaviour to singe backend setups is a nice middle ground. And maybe add a configuration option to enable this for multi backed setups as well.
This will make it better for most people, without breaking more complex multi backend setups.

I pushed a commit making that behaviour configurable. It's enabled by default, even in multi-backend setups, because I believe it will be fine for most users and will bring a noticeable performance improvement.

zeha:
While I think the config option is a good idea, maybe we want to name it differently, so further performance improvements can be done under the same option. IIRC `consistent-backends` was suggested at some point.

rgacogne:
> While I think the config option is a good idea, maybe we want to name it differently, so further performance improvements can be done under the same option. IIRC consistent-backends was suggested at some point.

Agreed, I'll change the name and the description so it means that can all records for a given name should be unique to a backend. Or perhaps we should have a setting to declare that zones are not spread across multiple backends, which is a bit more strict but would allow more optimizations later?

zeha:
> Agreed, I'll change the name and the description so it means that can all records for a given name should be unique to a backend. Or perhaps we should have a setting to declare that zones are not spread across multiple backends, which is a bit more strict but would allow more optimizations later?

@Habbie I hoped you'd chime in here.

I think "no overlays" would make sense as a general idea. Supporting real, delegated sub-zones should still work IMO?

Habbie:
Hello! I like `consistent-backends`. Maybe we can even switch it to `yes` in a release or two - especially if we can invent a `warn` setting that logs when a zone split over two backends is found? (but this warn feature should not be a blocker for this PR).

> Supporting real, delegated sub-zones should still work IMO?

I'd assume those would work by virtue of having distinct domain IDs.

rgacogne:
> Maybe we can even switch it to yes in a release or two - especially if we can invent a warn setting that logs when a zone split over two backends is found? (but this warn feature should not be a blocker for this PR).

We might even be able to set it to `yes` by default in the next major release in my opinion, as long as it's clearly documented?

> I'd assume those would work by virtue of having distinct domain IDs.

Yes, separate zones are fine.

Habbie:
I believe this is good to merge once the option is renamed and defaulted to off!

Habbie:
> I believe this is good to merge once the option is renamed and defaulted to off!

I pushed a commit for this.

zeha:
circle-ci build-auth failed with:

```
unknown location(0): fatal error: in "test_ueberbackend_cc/test_multi_backends_overlay_name": unknown type
test-ueberbackend_cc.cc(779): last checkpoint: "test_multi_backends_overlay_name" test entry
Undefined but needed argument: 'consistent-backends'
```

Habbie:
Thanks, that confused me for a bit, but I have now seen the light and am rebasing :)

Habbie:
Rebased, fixed, pushed. This passes locally. If I enable consistent backends, `make check` no longer passes, will investigate this.

pdns/ueberbackend.cc
```diff
@@ -632,36 +678,65 @@ void UeberBackend::getAllDomains(vector<DomainInfo> *domains, bool include_disab
...
+  if (d_negcached) {
     return false; 
   }
 
-  if(d_cached) {
-    if(d_cachehandleiter != d_answers.end()) {
-      rr=*d_cachehandleiter++;;
+  if (d_cached) {
+    if (d_cachehandleiter != d_answers.end()) {
+      rr = *d_cachehandleiter++;
```
mind04, line 688: I think we need a check here to make sure the zone_id in rr matches the id passed to lookup()

mind04 (changes requested):
All lookups with id = -1 (SOA and lookups in findNS()) may result in answers from multiple zones. Please add logic to prevent/detect answers from multiple zones in the cache.

pdns/ueberbackend.cc
```diff
@@ -368,8 +377,8 @@ bool UeberBackend::getAuth(const DNSName &target, const QType& qtype, SOAData* s
       if(i == backends.end()) {
         if(d_negcache_ttl) {
           DLOG(g_log<<Logger::Error<<"add neg cache entry:"<<shorter<<endl);
-          d_question.qname=shorter;
-          addNegCache(d_question);
+          d_question.qname = shorter;
+          addNegCache(d_question, d_question.qtype);
```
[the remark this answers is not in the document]
rgacogne: I'm confused, how would we know the requested qtype then, so that we can filter the records in `UeberBackend::get()`, if we don't store it in `d_question.qtype`? `d_handle.qtype` is already set to `ANY` in that case so we do the correct lookup when we iterate over backends in `UeberBackend::handle::get()`.

zeha:
> > All lookups with id = -1 (SOA and lookups in findNS()) may result in answers from multiple zones. Please add logic to prevent/detect answers from multiple zones in the cache.
> 
> I'm wondering how that works today? I'm not sure this PR changes that behaviour much, the SOA should not appear in more than one zones, the NSs will if we have a parent and a child zone but wasn't that already the case?

I gave this a stern look and would agree. IMO there are two remaining B->getSOA cases that could go away; and then we're down to a) FindNS and b) pdnsutil benchmark for callers passing `-1`. None of them should have a problem here though.

pdns/ueberbackend.cc
zeha: QC.getEntry already uses the zoneId for the cache lookup, no?

zeha:
I'm running this for today (in combination with #9464) and so far it's looking good.

zeha:
For some reason this needs a trivial rebase on master when merging locally, but I don't see why.

Habbie:
I merged #9483 (inspired by, and partially taken from, this PR) instead, as it appears to give the same benefits, while changing a lot less code. Thanks!

PR #9008, opened 2020-04-06
rec: test what's I believe was intended for test_RecursorCacheSimple

### LINKED ISSUE #9006: rec: test_RecursorCacheSimple has confusing/wrong parts
`test_RecursorCacheSimple` could be improved. This issue is mostly a reminder to look at it.

 - Program: Recursor 
 - Issue type: Bug report

### Short description

The initial part of `test_RecursorCacheSimple` fills the cache with entries that end up with ttd 0 in the cache, so they never are retrieved. This has to do with he insert being done with an empty vector of records. In that case the cache code never sets a proper ttd.

Initially `records` is empty. The ` BOOST_CHECK_EQUAL(retrieved.size(), records.size())` tests succeeds because no records were retrieved because of the `ttd == 0` issue, even though there are the expected number of matches because `get()` returns -1.

### Environment

 - Operating system: OpenBSD
 - Software version: master
 - Software source:  compiled yourself

### Steps to reproduce

1. Run unit tests, check source code.

### Expected behaviour

A test that actually tests expected behaviour.

### Actual behaviour

A test that tests some edge cases, and not what you would expect.

### Other information

For the routing tag test code I based my test on `test_RecursorCacheSimple` which causes me a lot of head scratching.

omoerbeek:
Also, zap unneeded clear() calls, get() already does that.
Fixes #9006 

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/recursordist/test-recursorcache_cc.cc
+++ b/pdns/recursordist/test-recursorcache_cc.cc
@@ -20,6 +20,19 @@ BOOST_AUTO_TEST_CASE(test_RecursorCacheSimple)
   std::vector<std::shared_ptr<RRSIGRecordContent>> signatures;
   time_t now = time(nullptr);
 
+  time_t ttd = now + 30;
+  DNSName power("powerdns.com.");
+  DNSRecord dr0;
+  ComboAddress dr0Content("2001:DB8::");
+  dr0.d_name = power;
+  dr0.d_type = QType::AAAA;
+  dr0.d_class = QClass::IN;
+  dr0.d_content = std::make_shared<AAAARecordContent>(dr0Content);
+  dr0.d_ttl = static_cast<uint32_t>(ttd);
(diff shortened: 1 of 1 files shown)


rgacogne:
> Also, zap unneeded clear() calls, get() already does that.

I'm not sure how I feel about that change, would we notice if `get()` stopped doing that?

omoerbeek:
> > Also, zap unneeded clear() calls, get() already does that.
> 
> I'm not sure how I feel about that change, would we notice if `get()` stopped doing that?

I think so, test would start failing; while currently get() stopping doing that might go unnoticed. Will double check tomorrow.

omoerbeek:
If I neuter the clear in `get()` tests start failing, both on master and on this PR branch. So we'd notice if `get()` would change in that respect.

PR #9016, opened 2020-04-08
random engine

omoerbeek:
### Short description

Continuation of #9004, which does not allow me to push anymore.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code


pdns/dnsname.hh
```diff
@@ -63,6 +63,15 @@ class DNSName
...
   DNSName()  {}          //!< Constructs an *empty* DNSName, NOT the root!
+  // Work around assertion in some boost versions that do not like self-assignment of boost::container::string
+  DNSName& operator=(const DNSName& a)
+  {
+    if (this != &a) {
+      d_storage = a.d_storage;
+    }
+    return *this;
+  }
+  DNSName(const DNSName& a) = default;
```
rgacogne, line 74: I know this is a work-in-progress but be careful that declaring a copy assignment operator or a copy constructor will prevent the move operations from being generated.

omoerbeek: Yes. But so far this is the only "solution" I could think of.

pdns/dnsname.hh
```diff
@@ -63,6 +63,16 @@ class DNSName
...
+  DNSName(DNSName&& a) = default;
```
rgacogne, line 75:
```suggestion
  DNSName(DNSName&& a) = default;
  DNSName& operator=(DNSName&& rhs) 
  {
    if (this != &rhs) {
      d_storage = std::move(rhs.d_storage);
    }
    return *this;
  }
```

pdns/shuffle.cc
```diff
@@ -0,0 +1,109 @@
...
+#endif
+
+#include <string>
+
+#include "shuffle.hh"
+#include "dns_random.hh"
+#include "dnsparser.hh"
+
+// shuffle, maintaining some semblance of order
+void pdns::shuffle(std::vector<DNSZoneRecord>& rrs)
```
rgacogne, line 33: So if I understand correctly, this function assumes that we have first CNAMEs in the ANSWER section, then other ANSWER records? In which case it will then shuffle the second ones only. And Then then the same for additional records? I know it's has not been modified by this PR but I think some comment might help for the next time, especially since the move to a new file means the history will be harder to find.

pdns/shuffle.cc
```diff
@@ -0,0 +1,109 @@
...
+
+  if(second-first>1)
+    shuffle(first, second, r);
+
+  // we don't shuffle the rest
+}
+
+
+// shuffle, maintaining some semblance of order
+void pdns::shuffle(std::vector<DNSRecord>& rrs)
```
rgacogne, line 63: And this one is mostly the same, except it also assumes that RRSIGs will come at the end of each section. It looks like it's never called directly, always from `orderAndShuffle()`, perhaps we could make it static?

pdns/dnsdist-lua-actions.cc
```diff
@@ -553,8 +554,9 @@ DNSAction::Action SpoofAction::operator()(DNSQuestion* dq, std::string* ruleresu
     }
   }
 
+  static std::default_random_engine r;
```
rgacogne, line 557: Unless `std::default_random_engine::operator()` is thread-safe, I don't think we should make `r` static.

omoerbeek: Indeed. I do not think we want to re-create the engine object all the time, since it might use a static seed (and it does afaks in at least the clang c++ lib). Maybe a thread local?

rgacogne: How about placing it in the `SpoofAction` class? I can write that change if you would like me to :)

omoerbeek: Right. If that takes care of the multi-threaded access that is preferred. Please do so. There might be caveats I do not know with the `DNSAction` classes :-)

rgacogne: On it!

rgacogne: For a reason unknown to me Github doesn't allow me to open a pull request against your branch, so please pull https://github.com/PowerDNS/pdns/commit/387221d594adb3a915ff93791e9225a4ecdd8ea7 from https://github.com/rgacogne/pdns/tree/random-engine :)

omoerbeek:
I think it is good to merge now. I'll squash it first though.

Issue #8663, opened 2019-12-31
API: gmysql Deadlock


Bodenhaltung:
- Program: Authoritative
 - Issue type: Bug report

### Short description
API Updates result in a MySQL Deadlock 

### Environment
 - Operating system: ArchLinux
 - Software version: pdns-4.2.1-1
 - Software source: Operating system repository

### Steps to reproduce
1. Send a few IP changes to the api.

### Expected behaviour
No Errors.

### Actual behaviour
MySQL Error:
```
[webserver] 50746175-0898-4b7c-812a-b8244a6c0a81 HTTP ISE for "/api/v1/servers/localhost/zones/domain.tld": Exception: GSQLBackend unable to nullify ordername and update auth for domain.tld for domain_id 1: Could not execute mysql statement: update records set ordername=NULL,auth=? where domain_id=? and name=? and disabled=0: Deadlock found when trying to get lock; try restarting transaction
```

Version 4.2.0-7 works still fine.

MySQL 8.0.16-7 Percona.

Record Table:

```
| records | CREATE TABLE `records` (
  `id` int(11) NOT NULL AUTO_INCREMENT,
  `domain_id` int(11) DEFAULT NULL,
  `name` varchar(255) CHARACTER SET latin1 COLLATE latin1_swedish_ci DEFAULT NULL,
  `type` varchar(10) CHARACTER SET latin1 COLLATE latin1_swedish_ci DEFAULT NULL,
  `content` varchar(64000) CHARACTER SET latin1 COLLATE latin1_swedish_ci DEFAULT NULL,
  `ttl` int(11) DEFAULT NULL,
  `prio` int(11) DEFAULT NULL,
  `disabled` tinyint(1) DEFAULT '0',
  `change_date` int(11) DEFAULT NULL,
  `ordername` varchar(255) CHARACTER SET latin1 COLLATE latin1_swedish_ci DEFAULT NULL,
  `auth` tinyint(1) DEFAULT NULL,
  PRIMARY KEY (`id`),
  KEY `rec_name_index` (`name`),
  KEY `nametype_index` (`name`,`type`),
  KEY `domain_id` (`domain_id`),
  KEY `orderindex` (`ordername`)
) ENGINE=InnoDB AUTO_INCREMENT=9601267 DEFAULT CHARSET=utf8 |
```

pieterlexis:
Can you please provide logs when running with these settings:

```
webserver-loglevel=detailed
loglevel=9
query-logging=yes
```

[...]

PR #9064, opened 2020-04-26
rectifyZone: do not update ordernames/auth when there is no update

nils-wisiol:
### Short description
Rectify-zone, triggered via API and CLI, computes and updates NSEC3 ordernames for _all_ rr sets in a given zone, but in many cases the time spend will result on no-op updates triggered in the backend. While both wastes CPU time, this PR removes no-op updates of NSEC3 ordernames. Removing the hash iterations that result in no-ops requires more elaborate changes and contribute less to the computational effort and is hence not part of this PR.

To be able to judge from within `DNSSECKeeper::rectifyZone` whether nor not an update of the ordername is required, the ordername is attached to `DNSResourceRecord` upon loading of the zone. Doing so should come at little computational cost.

In a test, I used a zone with ~100k rr sets and ~100k records. To benchmark `rectifyZone`, I inserted a new rr set via the API. The change in this commit increases performance by 600%, from before 30s and ~100k SQL queries updating ordernames down to 5s and 3 SQL queries that update ordernames. Similar numbers for CLI rectify-zone.

The current code has one backend-specific computation, `ordername.labelReverse().toString(" ", false)` which needs to be removed before merging. Any advise on how to proceed with this is appreciated.

I tested this against master and 4.2.x.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

IMHO, these items do not apply for this PR:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -54,3 +54,4 @@ built_pkgs
 *-shm
 __pycache__
 .circleci/config.yml-local
+.idea
--- a/docs/backends/generic-sql.rst
+++ b/docs/backends/generic-sql.rst
@@ -377,11 +377,11 @@ basic-query for the Generic MySQL backend would appear as:
 
 ::
 
-    gmysql-basic-query=SELECT content,ttl,prio,type,domain_id,disabled,name,auth FROM records WHERE disabled=0 and type=? and name=?
+    gmysql-basic-query=SELECT content,ttl,prio,type,domain_id,disabled,name,auth,ordername FROM records WHERE disabled=0 and type=? and name=?
 
 Queries can span multiple lines, like this:
 
 ::
 
-    gmysql-basic-query=SELECT content,ttl,prio,type,domain_id,disabled,name,auth \
+    gmysql-basic-query=SELECT content,ttl,prio,type,domain_id,disabled,name,auth,ordername \
     FROM records WHERE disabled=0 and type=? and name=?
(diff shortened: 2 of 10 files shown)


Habbie:
> The current code has one backend-specific computation, `ordername.labelReverse().toString(" ", false)` which needs to be removed before merging. Any advise on how to proceed with this is appreciated.

I've set the PR to Draft for this.

Habbie:
Several CI tests failed, please click Details on them to see why. At least it looks like you'll have to adjust `basic-query` in the other SQL backends as well.

nils-wisiol:
I believe the tests fail because upon loading of RR sets from the database, when the ordername is restored as `r.ordername = DNSName(boost::replace_all_copy(row[8]," ",".")).labelReverse()`, database ordername `NULL` and ordername `""` evaluate to the same thing, and things go south from there. Any advise how to fix this?

nils-wisiol:
It appears that mssql backend handles NULL differently than the other backends. Does anyone here have insight into that?

Other than that and the git history, this is ready for review.

nils-wisiol:
Should this be marked as "fixing" #8663? It does reduce chances to deadlock, but it (fundamentally) cannot stop deadlocks from occurring when different RR sets are inserted in the same NSEC gap.

Habbie:
> Should this be marked as "fixing" #8663

no :)

Issue #6993, opened 2018-09-27
Authoritative server uses `pthread_create`, should use `std::thread`


pieterlexis:
- Program: Authoritative
 - Issue type: Feature request

### Short description
In dnsdist and the recursor, `std::thread`s are used for threading, in the auth this is `pthread_create`. To ease development (when switching between programs), it would be nice to use `std::thread` everywhere.

V1shvesh:
Can I work upon this issue?

pieterlexis:
> Can I work upon this issue?

Hell yeah!

V1shvesh:
@pieterlexis Awesome! 

Can you point me to the files where these changes are to be made?

I see these lines,
```C++
pthread_rwlock_init(&mc.d_mut, 0);
```
over at `pdns/auth-querycache.cc` and `pdns/auth-packagecache.cc`.

Do I have to modify these files?

pieterlexis:
> Can you point me to the files where these changes are to be made?

I guess all files that are used in the authoritative server using `pthread_create`.

> Do I have to modify these files?

I guess these can be changed to a `std::mutex`.
@rgacogne am I correct here?

rgacogne:
> I guess these can be changed to a `std::mutex`.

Please don't do that, this is completely unrelated to the `pthread_create` -> `std::thread` move, and we use read-write locks instead of mutexes to keep the contention to a minimal level.

V1shvesh:
@rgacogne Got it!

I see that in file `pdns/ws-auth.hh`, the `pthread.h` lib is included. So I've changed it to the standard thread library, and I'll work my way from there. Sounds cool?

miodvallat:
Fixed by #9067. There are still minor dependencies on `pthread` functions rather than stl or boost routines, mostly for thread names and stack size, as well as the `Semaphore` class.

PR #9067, opened 2020-04-28
Wrap pthread objects

rgacogne:
### Short description

This PR wraps the remaining `pthread_*` objects in our code base:
- `pthread_mutex_t` -> `std::mutex` (always correctly initialized, can't be copied, properly destroyed via RAII) ;
- `pthread_rwlock_t` -> `ReadWriteLock` (light wrapper, always correctly initialized, can't be copied, properly destroyed via RAII) ;
- `pthread_create` -> `std::thread` (no need of an awkward wrapper to pass the parameters, easier to join or detach).

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/modules/bindbackend/bindbackend2.cc
+++ b/modules/bindbackend/bindbackend2.cc
@@ -77,9 +77,9 @@ Bind2Backend::state_t Bind2Backend::s_state;
 int Bind2Backend::s_first=1;
 bool Bind2Backend::s_ignore_broken_records=false;
 
-pthread_rwlock_t Bind2Backend::s_state_lock=PTHREAD_RWLOCK_INITIALIZER;
-pthread_mutex_t Bind2Backend::s_supermaster_config_lock=PTHREAD_MUTEX_INITIALIZER; // protects writes to config file
-pthread_mutex_t Bind2Backend::s_startup_lock=PTHREAD_MUTEX_INITIALIZER;
+ReadWriteLock Bind2Backend::s_state_lock;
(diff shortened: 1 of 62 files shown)


pieterlexis:
Related: #6993

rgacogne:
I missed a few spots, fixing that now.

rgacogne:
Rebased to fix a conflict with #9069.

rgacogne:
> I missed a few spots, fixing that now.

Done, ready for review!

omoerbeek:
wrt lock.hh: Is `g_singleThreaded` still useful? AFAIK it is never set.
Is the class `Lock` still useful?

rgacogne:
> wrt lock.hh: Is g_singleThreaded still useful? AFAIK it is never set.

Yeah, I was looking at the same thing. I think it should go, the idea was nice but we never used it and IMHO this is an optimization best left to the implementation.

> Is the class Lock still useful?

It's gone :)

rgacogne:
I just removed `g_singleThreaded`.

omoerbeek:
I looked at making the constructors taking a `pthread_rwlock_t *` private to avoid re-introducing lock objects using pthread rw locks. That does compie, so if you agree I can push the diff to this branch.

rgacogne:
> I looked at making the constructors taking a pthread_rwlock_t * private to avoid re-introducing lock objects using pthread rw locks. That does compie, so if you agree I can push the diff to this branch.

Nice idea, I like it!

PR #9070, opened 2020-04-28
Boost 1.73 moved boost::bind placeholders to the placeholders namespace

rgacogne:
### Short description

Reported by @dwfreed on IRC (thanks!), based on a patch from Gentoo [1].

[1]: https://bugs.gentoo.org/719926

### Checklist

I have:
- [x] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/ixfrdist-web.cc
+++ b/pdns/ixfrdist-web.cc
@@ -30,7 +30,7 @@ IXFRDistWebServer::IXFRDistWebServer(const ComboAddress &listenAddress, const Ne
 {
   d_ws->setACL(acl);
   d_ws->setLogLevel(loglevel);
-  d_ws->registerWebHandler("/metrics", boost::bind(&IXFRDistWebServer::getMetrics, this, _1, _2));
+  d_ws->registerWebHandler("/metrics", std::bind(&IXFRDistWebServer::getMetrics, this, std::placeholders::_1, std::placeholders::_2));
   d_ws->bind();
 }
 
--- a/pdns/webserver.cc
+++ b/pdns/webserver.cc
@@ -107,7 +107,7 @@ static void bareHandlerWrapper(WebServer::HandlerFunction handler, YaHTTP::Reque
 
 void WebServer::registerBareHandler(const string& url, HandlerFunction handler)
 {
-  YaHTTP::THandlerFunction f = boost::bind(&bareHandlerWrapper, handler, _1, _2);
+  YaHTTP::THandlerFunction f = std::bind(&bareHandlerWrapper, handler, std::placeholders::_1, std::placeholders::_2);
   YaHTTP::Router::Any(url, f);
 }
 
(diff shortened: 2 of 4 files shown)


rgacogne:
Fails on Travis:
```
webserver.cc:39:24: error: ‘placeholders’ is not a namespace-name
```

rgacogne:
The boost::placeholders namespace did not exist before 1.60.

rgacogne:
Pushed a new version that fixes the issue by moving to `std::bind` and `std::placeholders` instead.

neheb:
LLVM recommends avoiding bind: https://clang.llvm.org/extra/clang-tidy/checks/modernize-avoid-bind.html

No idea if lambdas are used here.

omoerbeek:
The existence of a `clang-tidy` test that marks a construct as undesirable is not enough reason to go about changing it everywhere IMO. 

Furthermore the mentioned test is a C++14 test, while we target C++11.

neheb:
Ehhh, the only thing C++14 specific thing is that check is using auto as a parameter.

rgacogne:
The reasons to use lambdas instead of `std::bind` are not very strong ones, IMHO:
> std::bind can be hard to read and can result in larger object files and binaries due to type information that will not be produced by equivalent lambdas.

Still I don't mind moving to lambdas if that makes the core more readable, but I don't think it prevents this PR from being merged. First we fix the real issue that our code doesn't compile anymore with boost 1.73+, then we can move to lambdas if we want to.

omoerbeek:
Backporting: tt seems this is a change that touches not to much stuff, but somehow I ended up needing to change a lot more wrt (optional) args to `pleaseWipeCache` and friends.

So I keep the backport on hold for the moment.

Habbie:
I'm removing the backport labels in the assumption that any affected downstreams will have managed by now. If not, I'm sure they'll let us know.

Habbie:
I put the 4.3 labels back because of a report on pdns-users

zeha:
> I'm removing the backport labels in the assumption that any affected downstreams will have managed by now. If not, I'm sure they'll let us know.

Debian still has 1.71, so it didn't fail -yet- :)

AdamMajer:
Looks like most of the changes here are actually missing in the 4.3.5 release although in the changelog. But yes, we manage ;)

omoerbeek:
Ugh, looks like I missed the webserver.cc ones somehow

Any other?

AdamMajer:
> Looks like mail gateway is foobared so I'll post what I sent in email earlier.

The patch that is still carried is,

 webserver.cc   |    6 +++---
 ws-recursor.cc |    8 ++++----
 2 files changed, 7 insertions(+), 7 deletions(-)

https://build.opensuse.org/package/view_file/server:dns/pdns-recursor/9070.patch?expand=1

Part of the ws-recursor.cc patch was applied in this release.

omoerbeek:
OK, but some of those lines do not use placeholders, so I think we can skip changing those.

AdamMajer:
Indeed. Works with only the webserver.cc placeholder changes.

PR #9073, opened 2020-04-29
Ensure runtime dirs for virtual services differ

pieterlexis:
### Short description
Systemd ensures the RuntimeDirectory is emptied when the service is started. This works fine until you start a second service (pdns@foo.service), which then removes the existing /run/pdns. This PR ensures the virtual services have their own runtime directories.

Note that we might need a code-change for virtual hosting to set the socket-dir automatically if pdns is virtually hosted and the socket-dir matches the socket-dir from `configure`.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


pieterlexis:
> Looks good. I guess that means that `rec_control` and `pdns_control` will need a `--socket-dir` argument if `socket-dir` is left empty in the configuration file, right?

Yes, they would.

PR #9085, opened 2020-05-01
builder: improve shipped config files

zeha:
### Short description

- #8094 introduced --config=default, but the official packages did not get updated to use that.
- Also the pdns-recursor.conf on Debian changes with each update as it contains the version string. Even if there are no other changes, this causes a conffile prompt for each update, which is annoying when using the master repositories + automatic upgrades.

### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/builder-support/debian/authoritative/debian-buster/rules
+++ b/builder-support/debian/authoritative/debian-buster/rules
@@ -55,7 +55,7 @@ override_dh_installinit:
 
 override_dh_install:
 	dh_install
-	./pdns/pdns_server --no-config --config | sed \
+	./pdns/pdns_server --no-config --config=default | sed \
 	  -e 's!# module-dir=.*!!' \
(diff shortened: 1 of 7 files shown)


mnordhoff:
> Also the pdns-recursor.conf on Debian changes with each update as it contains the version string. Even if there are no other changes, this causes a conffile prompt for each update, which is annoying when using the master repositories + automatic upgrades.

Auth does as well, because the comment at the top contains the date and time.

zeha:
> Edit: Ah, using `default` fixes that issue too, doesn't it?

Yes.

pieterlexis:
I was thinking, perhaps we should port the version-string setting from the auth to the recursor

zeha:
> I was thinking, perhaps we should port the version-string setting from the auth to the recursor

Had the same thought, maybe in a followup?

PR #9093, opened 2020-05-09
auth/rec/dnsdist: dockerise

Habbie:
### Short description
Todo:
* [ ] all TODO comments
* [ ] all tasks in DOCKER.md
* [ ] remove DOCKER.md
* [ ] write docs
* [ ] write tests for the dnsdist-resolver
* [x] this doesn't actually call `resolver.maintenance()` from `maintenance` in dnsdist  ~~, and when it does, it appears to re-add the servers every second~~ fixed
* [ ] test and demo modes (or perhaps those are the same thing?) that add a domain to the auth, make dnsdist point at it, etc. and test a few DNS queries

Reviews welcome though!

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/Dockerfile-auth
+++ b/Dockerfile-auth
@@ -1,0 +1,91 @@
+# our chosen base image
+FROM debian:10 AS builder
+
+# TODO: make sure /source looks roughly the same from git or tar
+
+# Reusable layer for base update
+RUN apt-get update && apt-get -y dist-upgrade && apt-get clean
+
+# devscripts gives us mk-build-deps (and a lot of other stuff)
+RUN apt-get update && apt-get -y dist-upgrade && apt-get install -y  --no-install-recommends devscripts equivs && apt-get clean
+
+# import everything - this could be pdns.git OR an auth tarball!
+COPY . /source
+
+# TODO: control file is not in tarballs at all right now
+RUN mk-build-deps -i -t 'apt-get -y -o Debug::pkgProblemResolver=yes --no-install-recommends' /source/builder-support/debian/authoritative/debian-buster/control && \
+    apt-get clean
+
+# build and install (TODO: before we hit this line, rearrange /source structure if we are coming from a tarball)
+WORKDIR /source/
+
+ARG MAKEFLAGS=
+ENV MAKEFLAGS ${MAKEFLAGS:--j2}
+
+RUN autoreconf -vfi
+
+# simplify repeated -C calls with SUBDIRS?
+RUN mkdir /build && \
+    ./configure \
+      --with-lua=luajit \
+      --sysconfdir=/etc/powerdns \
+      --enable-option-checking=fatal \
+      --with-dynmodules='bind geoip gmysql godbc gpgsql gsqlite3 ldap lmdb lua2 pipe random remote tinydns' \
+      --enable-tools \
+      --enable-ixfrdist && \
+    make clean && \
+    make $MAKEFLAGS -C ext && make $MAKEFLAGS -C modules && make $MAKEFLAGS -C pdns && \
+    make -C pdns install DESTDIR=/build && make -C modules install DESTDIR=/build && make clean && \
+    strip /build/usr/local/bin/* /build/usr/local/sbin/*
+RUN cd /tmp && mkdir /build/tmp/ && mkdir debian && \
+    echo 'Source: pdns' > debian/control && \
+    dpkg-shlibdeps /build/usr/local/bin/* /build/usr/local/sbin/* /build/usr/local/lib/pdns/*.so && \
+    sed 's/^shlibs:Depends=/Depends: /' debian/substvars >> debian/control && \
+    equivs-build debian/control && \
+    dpkg-deb -I equivs-dummy_1.0_all.deb && cp equivs-dummy_1.0_all.deb /build/tmp/
+
+# Runtime
+FROM debian:10
+
+# Reusable layer for base update - Should be cached from builder
+RUN apt-get update && apt-get -y dist-upgrade && apt-get clean
+
+# Ensure python3 is present (for startup script), and sqlite3 (for db schema), and tini (for signal management)
+RUN apt-get install -y python3 sqlite3 tini && apt-get clean
+
+# Output from builder
+COPY --from=builder /build /
+RUN chmod 1777 /tmp # FIXME: better not use /build/tmp for equivs at all
+RUN setcap 'cap_net_bind_service=+eip' /usr/local/sbin/pdns_server
+
+# Ensure dependencies are present
+RUN apt install -y /tmp/equivs-dummy_1.0_all.deb && apt clean
+
+# Start script
+COPY dockerdata/startup.py /usr/local/sbin/pdns_server-startup
+
+# Config file(s) from builder
+# Should not grab this from builder - since it isn't being built
+COPY --from=builder /source/dockerdata/pdns.conf /etc/powerdns/
+RUN mkdir -p /etc/powerdns/pdns.d /var/run/pdns
+RUN touch /etc/powerdns-api.conf && chown 953 /etc/powerdns-api.conf
+RUN ln -s /etc/powerdns-api.conf /etc/powerdns/pdns.d/api.conf
+
+# Make database dir before we drop root
+RUN mkdir -p /var/lib/powerdns && chown 953 /var/lib/powerdns
+
+# Work with pdns user - not root
+RUN adduser --system --disabled-password --disabled-login --no-create-home --group pdns --uid 953
+RUN chown pdns:pdns /var/run/pdns
+USER pdns
+
+# Set up database - this needs to be smarter
+RUN sqlite3 /var/lib/powerdns/pdns.sqlite3 < /usr/local/share/doc/pdns/schema.sqlite3.sql
+
+# DNS ports
+EXPOSE 53/udp
+EXPOSE 53/tcp
+# webserver port
+EXPOSE 8081/tcp
+
+ENTRYPOINT ["/usr/bin/tini", "--", "/usr/local/sbin/pdns_server-startup"]
--- a/Dockerfile-recursor
+++ b/Dockerfile-recursor
@@ -1,0 +1,97 @@
+# USAGE
+
+#   docker build --build-arg MAKEFLAGS=-j8 -t recursor -f docker/Dockerfile-recursor .
+#   docker run -p 1053:53 -p 1053:53/udp -ti --rm recursor
+#   dig a www.example.com @0 -p 1053
+
+# Builder
+FROM debian:10 AS builder
+
+# Reusable layer for base update
+RUN apt-get update && apt-get -y dist-upgrade && apt-get clean
+
+# devscripts gives us mk-build-deps (and a lot of other stuff)
+RUN apt-get install -y --no-install-recommends devscripts equivs git curl && apt-get clean
+
+# import everything - this could be pdns.git OR a recursor tarball!
+COPY . /source
+
+# TODO: make sure /source looks roughly the same from git or tar
+
+# TODO: control file is not in tarballs at all right now
+RUN mk-build-deps -i -t 'apt-get -y -o Debug::pkgProblemResolver=yes --no-install-recommends' /source/builder-support/debian/recursor/debian-buster/control && \
+    apt-get clean
+# RUN apt-get -y install protobuf-compiler && apt-get clean
+
+# build and install (TODO: before we hit this line, rearrange /source structure if we are coming from a tarball)
+WORKDIR /source/pdns/recursordist
+
+ARG MAKEFLAGS=
+ENV MAKEFLAGS ${MAKEFLAGS:--j2}
+
+# Manpage deps
+# RUN apt-get install -y virtualenv && apt-get clean
+
+# Manpage prevent
+RUN touch pdns_recursor.1 rec_control.1 # avoid installing pandoc
+
+RUN autoreconf -vfi
+
+RUN mkdir /build && \
+    ./configure \
+      --with-lua=luajit \
+      LDFLAGS=-rdynamic \
+      --sysconfdir=/etc/powerdns \
+      --enable-option-checking=fatal && \
+      make clean && \
+      make $MAKEFLAGS install DESTDIR=/build && make clean && \
+      strip /build/usr/local/bin/* /build/usr/local/sbin/*
+RUN cd /tmp && mkdir /build/tmp/ && mkdir debian && \
+    echo 'Source: pdns' > debian/control && \
+    dpkg-shlibdeps /build/usr/local/bin/rec_control /build/usr/local/sbin/pdns_recursor && \
+    sed 's/^shlibs:Depends=/Depends: /' debian/substvars >> debian/control && \
+    equivs-build debian/control && \
+    dpkg-deb -I equivs-dummy_1.0_all.deb && cp equivs-dummy_1.0_all.deb /build/tmp/
+
+# Runtime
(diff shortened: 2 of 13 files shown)


zeha:
dont you want a two step docker build: use one image to compile + build the runnable image?

Dockerfile-auth
```diff
@@ -0,0 +1,39 @@
...
+ENV MAKEFLAGS ${MAKEFLAGS:--j2}
+
+RUN autoreconf -vfi
+
+RUN ./configure --with-lua=luajit --sysconfdir=/etc/powerdns --enable-option-checking=fatal --with-modules='bind geoip gsqlite3' --enable-tools --enable-ixfrdist && make clean && make $MAKEFLAGS  -C ext && make $MAKEFLAGS -C modules && make $MAKEFLAGS -C pdns && make -C pdns install && make clean
+
+RUN mkdir -p /etc/powerdns/pdns.d && cp /source/dockerdata/pdns.conf /etc/powerdns
+RUN cp /source/dockerdata/startup.py /usr/local/sbin/pdns_server-startup
+
+RUN mkdir -p /var/lib/powerdns && sqlite3 /var/lib/powerdns/pdns.sqlite3 < /source/modules/gsqlite3backend/schema.sqlite3.sql
```
zeha, line 31: /var/lib/powerdns/ wants to be a volume and if its empty, `pdns.sqlite3` wants to be created anew?

Dockerfile-auth
```diff
@@ -0,0 +1,39 @@
...
+
+# DNS ports
+EXPOSE 53/udp
+EXPOSE 53/tcp
+# webserver port
+EXPOSE 8081/tcp
+
+ENTRYPOINT ["/usr/bin/tini", "--", "/usr/local/sbin/pdns_server-startup"]
```
jsoref, line 39: Newline at eof ;-) -- a couple of files

DOCKER.md
```diff
@@ -0,0 +1,25 @@
+
+# Use local Ubuntu mirrors
+RUN sed -i 's%http://archive.ubuntu.com/ubuntu/%mirror://mirrors.ubuntu.com/mirrors.txt%' /etc/apt/sources.list
+
+Ik weet niet of mirror: daar werkt
```
jsoref, line 5: :-)

dockerdata/dnsdist-resolver.lua
```diff
@@ -0,0 +1,110 @@
...
+    if not ret
+    then
+        error(resout)
+    end
+
+    -- on purpose no pcall, an error here is a bug
+    resout = resout()
+
+    -- check for servers removed by controller
+    for name,v in pairs(ourservers)
```
jsoref, line 76:
```suggestion
    for name, v in pairs(ourservers)
```

Habbie:
> dont you want a two step docker build: use one image to compile + build the runnable image?

Probably - this is currently a note in DOCKER.md :)

Dockerfile-auth
Habbie: makes sense!

dockerdata/dnsdist.conf
```diff
@@ -0,0 +1,9 @@
+setLocal('0.0.0.0')
+
+resolver = require 'dnsdist-resolver'
+resolver.servers.auth = {pool='auths'}
+resolver.servers.recursor = {pool='recursors'}
```
wojas, line 5: This hardcodes the docker-compose setup. We probably need a more generic way of doing this, like passing the pools in env vars. At the very least add comments, because it's not obvious what the hardcoded assumptions are here.

For that static setup using [depends_on](https://docs.docker.com/compose/compose-file/#depends_on) in `docker-compose.yml` and `os.getenv` here would probably suffice, but this approach could be useful in other setups.

dockerdata/dnsdist-resolver.lua
```diff
@@ -0,0 +1,113 @@
...
+
+    -- it does not exist, let's add it
+    local settings = tablecopy(_M.servers[name])
+    settings.name = name
+    -- TODO: we only take the first IP
+    settings.address = ip
+    ourservers[name] = {ip, newServer(settings)}
+end
+
+function _M.maintenance()
```
wojas, line 58: Does all of this block dnsdist?

dockerdata/startup.py
```diff
@@ -0,0 +1,49 @@
...
+setConsoleACL('0.0.0.0/0')
+"""
+
+apikey = os.getenv(apienvvar)
+print("apikey=", apikey)
+if apikey is not None:
+	with open(apiconffile, 'w') as conf:
+		conf.write(apiconftemplate.format(apikey=apikey))
+
+os.execv(program, [program]+args+sys.argv[1:])
```
wojas, line 49: Perhaps first print what you are about to exec, like you would with `set -x` in a shell script.

Dockerfile-auth
```diff
@@ -0,0 +1,39 @@
+# our chosen base image
+FROM debian:10
+
+# import everything - this could be pdns.git OR an auth tarball!
+COPY . /source
```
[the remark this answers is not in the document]
wojas: Same applies to the other Dockerfiles.

dockerdata/dnsdist-resolver.lua
```diff
@@ -0,0 +1,113 @@
...
+    -- TODO: check return values
+    for k,v in pairs(_M.servers)
+    do
+        resolverpipe:write(k..' ')
+    end
+    resolverpipe:write('\n')
+    resolverpipe:flush()
+
+    -- TODO: maybe this failure should be quiet for the first X seconds?
+    local ret, resout = pcall(loadfile, './dnsdist-resolver.out')
```
wojas, line 69: This looked like a race condition in that you cannot be sure it has already been written. It does not matter, because it will be read during a future maintenance. Perhaps add a comment explaining this intended behavior?

Dockerfile-auth
```diff
@@ -0,0 +1,39 @@
...
+
+# build and install (TODO: before we hit this line, rearrange /source structure if we are coming from a tarball)
+WORKDIR /source/
+
+ARG MAKEFLAGS=
+ENV MAKEFLAGS ${MAKEFLAGS:--j2}
+
+RUN autoreconf -vfi
+
+RUN ./configure --with-lua=luajit --sysconfdir=/etc/powerdns --enable-option-checking=fatal --with-modules='bind geoip gsqlite3' --enable-tools --enable-ixfrdist && make clean && make $MAKEFLAGS  -C ext && make $MAKEFLAGS -C modules && make $MAKEFLAGS -C pdns && make -C pdns install && make clean
```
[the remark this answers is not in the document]
wojas: Perhaps also add a `PDNS_AUTH_CONFIGURE_EXTRA` buildarg.

And you can split lines in a Dockerfile with a backslash:

```docker
RUN ./configure --with-lua=luajit --sysconfdir=/etc/powerdns --enable-option-checking=fatal \
        --with-modules='bind geoip gsqlite3' --enable-tools --enable-ixfrdist && \
    make clean && \
    make $MAKEFLAGS -C ext && \
    make $MAKEFLAGS -C modules && \
    make $MAKEFLAGS -C pdns && \
    make -C pdns install && \
    make clean
```

Dockerfile-recursor
```diff
@@ -0,0 +1,45 @@
...
+RUN mk-build-deps -i -t 'apt-get -y -o Debug::pkgProblemResolver=yes --no-install-recommends' /source/builder-support/debian/recursor/debian-buster/control && apt-get clean
+RUN apt-get -y install protobuf-compiler && apt-get clean
+
+# build and install (TODO: before we hit this line, rearrange /source structure if we are coming from a tarball)
+WORKDIR /source/pdns/recursordist
+
+ARG MAKEFLAGS=
+ENV MAKEFLAGS ${MAKEFLAGS:--j2}
+
+#RUN touch pdns_recursor.1 rec_control.1 # avoid installing pandoc
```
wojas, line 30: Do we still avoid installing 1 GB of pandoc in some other way?

Dockerfile-auth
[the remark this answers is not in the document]
Habbie: > Do this as late as possible, definitely _after_ the installation of all the build dependencies.

I currently steal the list of build dependencies, at build time, from our Debian sources. To move the COPY to later, I'd have to copy that list and then we'd be maintaining it in two places.. or perhaps I could copy just the data for the build deps here!

I should definitely have a `.dockerignore`.

As for figuring out a version number, I don't think a Dockerfile can decide what version number the image gets anyway, so that would have to happen externally I believe?

Dockerfile-auth
```diff
@@ -0,0 +1,39 @@
+# our chosen base image
+FROM debian:10
+
+# import everything - this could be pdns.git OR an auth tarball!
+COPY . /source
+
+# TODO: make sure /source looks roughly the same from git or tar
+
+# TODO: ccache
```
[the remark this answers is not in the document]
Habbie: I am also unsure, but it would be good to have if is was possible.

dockerdata/dnsdist-resolver.lua
Habbie: Yes, that's why the DNS resolving is an external process.

Dockerfile-recursor
Habbie: The comment should say 'avoid having to install pandoc'. pandoc does not currently end up in the images (but for various other reasons, many of which I'm sure I will avoid by doing multistage builds, the images are around 4GB right now).

Habbie: It is possible I get away with this because `COPY ... /source` is taking my unclean host tree that already has built manpages, will check.

Habbie: > It is possible I get away with this because `COPY ... /source` is taking my unclean host tree that already has built manpages, will check.

Confirmed!

Dockerfile-auth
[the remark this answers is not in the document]
wojas: Yes, you could just copy the the files you need at each stage.

For example, for Python projects I usually first add `requirements.txt` and `pip install` those, and only then add the rest of the sources. This way changing the sources does not trigger a reinstall of all third party packages.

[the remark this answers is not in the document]
wojas: You could pass the version you want through an optional build arg, I'm just not sure if you can easily pass that to the rest of the build system.

Dockerfile-auth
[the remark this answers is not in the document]
wojas: Unfortunately there is currently no support for something like a build cache volume.

In pdns-builder we add it in a convoluted way that requires tight support from the outer build.sh script.

Dockerfile-recursor
wojas: Hehe, definitely add a `.dockerignore` file to prevent shipping all your local build output to Docker.

spirillen:
Just an applause and codo to @Habbie & Co. for this idea. I can see the potential in this :+1: 

I'm looking forward to see this implementation as it could save a lot of ns to ip lookup code for 1 to many special rules

M4t7e:
Really great to see that you are planning to dockerize pdns servers. It makes life easier, especially for testing and development purposes.

For dnsdist I sometimes start Docker containers with interactive mode (no supervised flag) for testing stuff. For this reason, I created an entrypoint script to distinguish automatically between the two modes:
```bash
if [ -t 0 ]; then
    supervised=""
else
    supervised="--supervised"
fi

exec dnsdist $supervised --disable-syslog --uid pdns --gid pdns "${@}"
```
... and start the container with `docker run -p 127.0.0.1:53:53/udp -it pdns-dnsdist -l 0.0.0.0`. Would be good to have a similar possibility in your images and maybe the default listening address could be 0.0.0.0 by default?

I have seen that there is a resolver function in dnsdist. That sounds very good to me. In the past, it was also somewhat difficult with an auth or recursor container together with dnsdist, since only raw IP addresses could be configured in dnsdist. I wrote a lua ffi function to make the gethostbyname function available. I use this to resolve host names to IP addresses and dynamically add new servers. It is also synchronized whether a new server was added later or deleted using the maintenance function. This supports scaling upstream servers, but my approach doesn't feel that good...

When scaling dnsdist containers in host network mode, handling the control sockets is also difficult. My dirty trick is to randomly generate the IP address to avoid conflicts. Here is an excerpt from my entrypoint script:
```bash
control_address="127.$((RANDOM % 256)).$((RANDOM % 256)).$((1 + RANDOM % 254))"
dnsdist --disable-syslog --uid pdns --gid pdns --local "${control_address}:5199" <<EOF | grep setKey > /etc/dnsdist/dnsdist.conf
makeKey()
EOF
echo "controlSocket('${control_address}:5199')" >> /etc/dnsdist/dnsdist.conf
```
It would be nice to have the option to use a Unix socket instead. And in addition, a cleaner way to dynamically create the socket key. My approach is hacky here too I think.

Configuring the containers can sometimes not be that easy. For Recursor and Auth, I use `envtpl` and a jinja script in my entry point to generate the configuration based on the environment variable starting with `RECURSOR_` and `AUTH_` prefix. 
```jinja
{% for key, value in environment('RECURSOR_') %}{{ (key|replace('_', '-')).lower() }}={{ value }}
{% endfor %}
```
Since dnsdist is so dynamically also in configuration, I currently use a combination of jinja scripts with predefined environment variables and a special variable called `DNSDIST_CONFIG` where a can dynamically put multiline config e.g. with a docker-compose.yml file that is appended to the dnsdist config file. Maybe a (additional) different approach like a special directory for custom config (e.g. `/etc/dnsdist/conf.d/`) could be good.

franklouwers:
> > dont you want a two step docker build: use one image to compile + build the runnable image?
> 
> Probably - this is currently a note in DOCKER.md :)

You might want to take step back: do we need to compile? Or can we (re-)use the existing packages? In my own docker images for pdns for instance, I usually base them on `FROM debian:slim` and just add the repo, apt-get the images etc.

Habbie:
> do we need to compile? Or can we (re-)use the existing packages?

We decided to compile, so that local changes to the code are also reflected in the image.

Habbie:
rebased

chrisss404:
I did something similar. I'll leave this here in case it's useful to anyone: https://github.com/chrisss404/powerdns

Habbie:
> dont you want a two step docker build: use one image to compile + build the runnable image?

Done!

Dockerfile-recursor
```diff
@@ -78,4 +79,4 @@ EXPOSE 53/tcp
 # webserver port
 EXPOSE 8082/tcp
 
-ENTRYPOINT ["/usr/local/sbin/pdns_recursor-startup"]
\ No newline at end of file
+ENTRYPOINT ["/usr/bin/tini", "--", "/usr/local/sbin/pdns_recursor-startup"]
```
jsoref, line 82: Please add the missing newline at eof :-)

Habbie: fixed I think. I might unfix it by accident at some later point though..

Dockerfile-recursor
Habbie: Just adding the Dockerfiles to .dockerignore has done wonders for build times at least!

Dockerfile-auth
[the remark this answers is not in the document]
Habbie: > While the default should not require a real database, do build with postgres, mysqland any other potentially useful modules for people that want to use this image

did this

Dockerfile-recursor
Habbie: We now avoid pandoc by being explicit about our make targets.

pieterlexis (approved):
lgtm for a first implementation

PR #9099, opened 2020-05-13
rec: don't suggest to add a TA for powerdns.com

omoerbeek:
### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code


phonedph1:
There is a line just above which is related;

`To e.g. add a trust anchor for the root and powerdns.com, use the following config in the Lua file:`

omoerbeek:
> There is a line just above which is related;
> 
> `To e.g. add a trust anchor for the root and powerdns.com, use the following config in the Lua file:`

Fixed; thanks!

PR #9106, opened 2020-05-15
Tell about the release cycles.

omoerbeek:
### Short description

If this table format is OK, I'll add the auth one.

### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/docs/appendices/EOL.rst
+++ b/docs/appendices/EOL.rst
@@ -1,6 +1,12 @@
 End of life statements
 ======================
 
+We aim to have a release every six months.
+The latest and previous release receive correctness, stability and security updates.
+The release before that gets critical security updates only.
+Older releases are marked end of life and receive no updates at all.
+Pre-releases do not receive immediate security updates.
+
 The currently supported release train of PowerDNS Authoritative Server is 4.3.
 
 PowerDNS Authoritative Server 4.2 will only receive correctness, stability and security updates and will be receiving security updates only after PowerDNS Authoritative Server 4.4 is released.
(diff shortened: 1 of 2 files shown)


pdns/recursordist/docs/appendices/EOL.rst
```diff
@@ -1,5 +1,10 @@
 End of life statements
-======================
+=====================
```
zeha, line 2: did this remove one actually required =?

omoerbeek: yes, tx

docs/appendices/EOL.rst
```diff
@@ -1,6 +1,11 @@
 End of life statements
 ======================
 
+We aim to have a release every six months.
+The latest and previous release receive correctness, stability and security updates.
+The release before that gets critical security updates only.
```
zeha, line 6:
```suggestion
The second-to-last release before that gets critical security updates only.
```

docs/appendices/EOL.rst
```diff
@@ -1,6 +1,11 @@
...
+Older releases are marked end of life.
```
zeha, line 7:
```suggestion
Older releases are marked end of life and receive no updates at all.
```

docs/appendices/EOL.rst
```diff
@@ -1,6 +1,11 @@
...
+
```
zeha, line 8:
```suggestion
Pre-releases do not receive immediate security updates.

```

if this is true?

docs/appendices/EOL.rst
omoerbeek: I find second-to-last always confusing. To me it means the one before last here.

PR #9125, opened 2020-05-19
Recursor 4.3.1: Fix compilation on FreeBSD

RvdE:
Fix compilation of Recursor 4.3.1 on FreeBSD. Hopefully this works on other platforms as well.

### Checklist

I have:
- [ X ] read the [CONTRIBUTING.md](https://github.com/PowerDNS/pdns/blob/master/CONTRIBUTING.md) document
- [ X ] compiled this code
- [ X ] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/rec-carbon.cc
+++ b/pdns/rec-carbon.cc
@@ -40,16 +40,16 @@ try
     namespace_name="pdns";
   }
   if(hostname.empty()) {
-    char tmp[HOST_NAME_MAX+1];
-    memset(tmp, 0, sizeof(tmp));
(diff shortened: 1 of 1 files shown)


omoerbeek:
```
rec-carbon.cc:40:23: error: variable length array 'tmp' is used [-Werror=vla]
 
     char tmp[hostmax+1];
 ```
We do not allow vla's...

omoerbeek:
Thanks for the PR, but #9127  presents a different and more C++ way if solving this.

Issue #2726, opened 2015-09-03
Default carbon-ourname won't replace dots with underscores


friek:
As documented on [this documentation page](https://doc.powerdns.com/md/common/logging/#sending-to-carbongraphitemetronome), pdns recursor is supposed to replace the dots in a hostname with underscores. Instead, it strips everything after the first dot, leaving only the first part of the hostname as carbon-ourname.

It may be related to [this patch](https://github.com/PowerDNS/pdns/pull/2381).

ahupowerdns:
so after some thinking, I think it is worth changing this even though it will break some setups. We might get too many collisions now between recursor.delft.corp  and recursor.denhaag.corp.

omoerbeek:
Fixed

PR #2381, opened 2015-03-26
Carbon-ourname fix

cyclops1982:
Replace '.' with '_' in carbon-ourname

PR #9127, opened 2020-05-19
Fix compilation on systems that do not define HOST_NAME_MAX

rgacogne:
### Short description

On FreeBSD at least, HOST_NAME_MAX is not defined and we need to use sysconf() to get the value at runtime instead.
Based on a work done by @RvdE to make the recursor compile on FreeBSD (many thanks!).

This PR generalizes the work done in #9125 to cover the authoritative server and dnsdist as well.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/misc.cc
+++ b/pdns/misc.cc
@@ -57,6 +57,7 @@
 #include <sys/types.h>
 #include <pwd.h>
 #include <grp.h>
+#include <limits.h>
 #ifdef __FreeBSD__
 #  include <pthread_np.h>
 #endif
@@ -1559,4 +1560,35 @@ DNSName reverseNameFromIP(const ComboAddress& ip)
   }
 
   throw std::runtime_error("Calling reverseNameFromIP() for an address which is neither an IPv4 nor an IPv6");
+}
+
+static size_t getMaxHostNameSize()
+{
+#if defined(HOST_NAME_MAX)
+  return HOST_NAME_MAX;
+#endif
+
+#if defined(_SC_HOST_NAME_MAX)
+  auto tmp = sysconf(_SC_HOST_NAME_MAX);
+  if (tmp != -1) {
+    return tmp;
+  }
+#endif
+
+  /* _POSIX_HOST_NAME_MAX */
+  return 255;
+}
+
+std::string getCarbonHostName()
+{
+  std::string hostname;
+  hostname.resize(getMaxHostNameSize() + 1, 0);
+
+  if (gethostname(const_cast<char*>(hostname.c_str()), hostname.size()) != 0) {
+    throw std::runtime_error(stringerror());
+  }
+
+  boost::replace_all(hostname, ".", "_");
+
+  return hostname;
 }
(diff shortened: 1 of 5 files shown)


pdns/misc.cc
```diff
@@ -1560,3 +1561,39 @@ DNSName reverseNameFromIP(const ComboAddress& ip)
...
+  if (gethostname(const_cast<char*>(hostname.c_str()), hostname.size()) != 0) {
+    throw std::runtime_error(stringerror());
+  }
+
+  auto pos = hostname.find(".");
+  if (pos != std::string::npos) {
+    hostname.resize(pos);
+  }
+
+  boost::replace_all(hostname, ".", "_");
```
omoerbeek, line 1596: I might be missing something, but if you just truncated the string until the first dot there are no dots left in the string. So to me this `replace_all` reads as a no-op.

rgacogne: It was present in the original code so I left it there since I don't want to break anything else but it doesn't make sense to me either. I looked at the history but as far as I can tell it was never needed. I would suggest removing it on master but keeping it around on the already released branches.

Habbie: #2726 #2381

omoerbeek: So should we truncate or replace? We can't have both. Or was the intend to strip a trailing dot?

rgacogne: I gather from Habbie's links that the original intent was to keep only the local part of the name, then later we tried to use the full host name by replacing dots with underscores but failed to do so. So I propose to stop truncating on master, do you agree?

omoerbeek: yes

Habbie: #2726 says 'we should replace dots with underscores in `carbon-ourname`'. #2381 says "no don't do that".

Ignoring the potential for breakage, I'd say the right way is to do underscores-for-dots for the machine hostname, and do nothing for `carbon-ourname`.

Not ignoring the potential for breakage, I'm not sure what we should do.

@pieterlexis ?

rgacogne: Oh, yeah, I meant to propose replacing dots with hyphens instead of underscores, I don't think underscores would ever work. I agree `carbon-ourname` should stay unaltered.

Habbie: I don't see why hyphens would be better than underscores, but @pieterlexis would know better

zeha:
This is also a problem on macOS :)

pdns/misc.cc
pieterlexis: hyphens or underscores don't matter in carbon, only dots do. This function should indeed replace the dots with _something_ that doesn't confuse graphite. When a user sets dot in `carbon-hostname`, they are on their own. We try to provide a safe default for those who don't set `carbon-hostname`.

rgacogne: Alright, so I'll change this code to replace dots with underscores, and only when `carbon-ourname` is not set.

PR #9142, opened 2020-05-25
rec: Defer the NOD lookup until after the response has been sent

### LINKED ISSUE #8836: Newly Observed Domain Tracking slows down resolving in certain circumstances
- Program: Recursor
 - Issue type: Bug report

### Short description
If I use `new-domain-tracking` in conjunction with `new-domain-lookup` and `forward-zones`, resolving of "newly observed domains" is dog slow. But only if I send the `new-domain-lookup` to an IP address, which is not responding (neither DNS nor ICMP).

### Environment
 - Operating system: SLES15-SP1
 - Software version: PowerDNS Recursor 4.2.1

### Steps to reproduce
1. cd into an empty directory
`# mkdir -p nod-test && cd nod-test`
2. Delete possibly existing new-domain-history files and start the Recursor from command line
`# rm -f nod_*; /usr/sbin/pdns_recursor --daemon=no --config-dir=. --local-address=127.0.0.53 --new-domain-tracking=yes --new-domain-history-dir=. --new-domain-lookup=new-domain-lookup --forward-zones=new-domain-lookup=100::`
3. dig it
```
$ dig +tries=1 +time=10 @127.0.0.53 . SOA
...
;; Query time: 1506 msec
...
```
### Expected behaviour
"Query time:" below 100 msec

### Actual behaviour
"Query time:" above 1000 msec

rgacogne:
### Short description

If the NOD lookup is slow, for example because the destination authoritative server is down, doing the NOD lookup before the response has been sent increases the latency a lot.
This commit moves the actual NOD lookup after the response has been sent, so we can still use the existing mthread (we might actually need to do a proper DNS resolution to find the target authoritative server) without keeping the client waiting.

Fixes #8836.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/pdns_recursor.cc
+++ b/pdns/pdns_recursor.cc
@@ -1038,8 +1038,6 @@ static bool checkFrameStreamExport(LocalStateHolder<LuaConfigItems>& luaconfsLoc
 #ifdef NOD_ENABLED
 static bool nodCheckNewDomain(const DNSName& dname)
 {
-  static const QType qt(QType::A);
-  static const uint16_t qc(QClass::IN);
   bool ret = false;
   // First check the (sub)domain isn't whitelisted for NOD purposes
   if (!g_nodDomainWL.check(dname)) {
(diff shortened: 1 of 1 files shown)


rgacogne:
Rebased to fix travis failures.

PR #9152, opened 2020-05-25
Make sure we can install unsigned packages.

aerique:
Sometimes we need to install unsigned packages from our own ad-hoc repo,
installing `apt-transport-https` makes sure we can do this (at least on
Debian Stretch).


pieterlexis:
does this actually work? AFAIK, apt still refuses unsigned packages, even downloaded over https. See [apt-secure(8)](http://manpages.ubuntu.com/manpages/bionic/man8/apt-secure.8.html)

aerique:
> does this actually work? AFAIK, apt still refuses unsigned packages, even downloaded over https. See [apt-secure(8)](http://manpages.ubuntu.com/manpages/bionic/man8/apt-secure.8.html)

I needed to add this for Debian Stretch to get it to install packages from the Pulp repo.

pieterlexis:
> I needed to add this for Debian Stretch to get it to install packages from the Pulp repo.

Hrm, we do a 301 redir to HTTPS, but those are indeed unsigned. perhaps apt does allow that.

PR #9169, opened 2020-05-28
auth: fetch all metadata at once

mind04:
### Short description
Fetch all metadata at once when a query is being answered. This will reduce the number of backend/database queries by roughly 20%.

### Checklist
I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


Habbie:
> Looks great, my only worry is that a backend could implement `getDomainMetadata()` without implementing `getAllDomainMetadata()`, but given that this is required to work properly with several `pdnsutil` commands it's probably unlikely.

Perhaps we could (not necessarily now) remove `getDomainMetadata` from all backends, and implement it in `DNSBackend` on top of `getAllDomainMetadata` (which all backends then need to start supporting). Then your surprise can never happen.

zeha:
I agree that we should remove getDomainMetadata soon, then.

mind04:
I'm against backporting to 4.3.x The change in behaviour is too big for a minor release.

zeha:
Right.

PR #9171, opened 2020-05-28
RFC: addRRSigs/getPreRRSIGs: use internal UeberBackend

zeha:
To start a discussion!

### Short description

From my analysis of the code calling these functions, everybody passes the same UeberBackend that is used to construct the DBDNSSECKeeper on which the functions are called. So this is no practical change, and it makes the code simpler to follow.

I also see some optimisations that can be done if this is acceptable.

It certainly takes away the theoretical possibility of passing in another UeberBackend, but this does not appear to be useful today.

NB: yes, some callers of addRRSigs pass in a keymeta-only UeberBackend, but this is fine, as they don't hit the codepath leading to any list/get calls. I'd propose splitting up addRRSigs in a later step, possibly as part of the optimisation mentioned above. 

### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dbdnsseckeeper.cc
+++ b/pdns/dbdnsseckeeper.cc
@@ -567,19 +567,20 @@ bool DNSSECKeeper::checkKeys(const DNSName& zone, vector<string>* errorMessages)
   return retval;
 }
 
-bool DNSSECKeeper::getPreRRSIGs(UeberBackend& db, const DNSName& signer, const DNSName& qname,
+// A fully functional UeberBackend is required.
+bool DNSSECKeeper::getPreRRSIGs(const DNSName& signer, const DNSName& qname,
         const DNSName& wildcardname, const QType& qtype,
         DNSResourceRecord::Place signPlace, vector<DNSZoneRecord>& rrsigs, uint32_t signTTL)
 {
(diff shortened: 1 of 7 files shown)


mind04:
d_keymetadb is not always capable of doing record lookups (bind backend)

rgacogne:
> d_keymetadb (DNSSECKeeper db) is not always capable of doing record lookups (bind backend)

That might be fine as long as we don't call those functions with such a DNSSECKeeper object, I guess?

zeha:
> d_keymetadb (DNSSECKeeper db) is not always capable of doing record lookups (bind backend)

Yes, this is why I said:

> NB: yes, some callers of addRRSigs pass in a keymeta-only UeberBackend, but this is fine, as they don't hit the codepath leading to any list/get calls. I'd propose splitting up addRRSigs in a later step, possibly as part of the optimisation mentioned above.

zeha:
Quite the conflict on `DNSSECKeeper::getPreRRSIGs`, need to look at that closely.

Habbie:
I have no idea whether this still makes sense. It's also highly conflicted. If you feel this work is still relevant, please feel free to reopen after unconflicting it!

PR #9174, opened 2020-05-29
tinydnsbackend: Ignore duplicate SOA in getAllDomains()

### LINKED ISSUE #9173: tinydns: handle duplicate SOAs
When I migrate data from tinydns to gpgsql, I get this error:
 ```
# pdnsutil b2b-migrate tinydns gpgsql
May 29 02:53:03 [bindbackend] Done parsing domains, 0 rejected, 0 new, 0 removed
Moving zone(s) from tinydns to gpgsql
Processing 'klonopin-no-prescription.us'
Moved 10 record(s), 0 comment(s), 0 metadata(s) and 0 cryptokey(s)
Processing 'klonopin-no-prescription.us'
Error: Database error trying to insert new domain 'klonopin-no-prescription.us.': Fatal error during query: insert into domains (type,name,master,account,last_check, notified_serial) values($1,$2,$3,$4,null,null): ERROR:  duplicate key value violates unique constraint "name_index"
DETAIL:  Key (name)=(klonopin-no-prescription.us) already exists.
```
Operating system: Ubuntu 18.04.4 LTS
Software version: 4.1.13
Software source: PowerDNS repository

Does someone know how to deal with it, please?

### PR DESCRIPTION
**Pull request by cmouse**
### Short description
This patch prevents tinydns from providing same domain twice when listing all domains. This can happen if there is same SOA record more than once.

Fixes #9173 

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/modules/tinydnsbackend/tinydnsbackend.cc
+++ b/modules/tinydnsbackend/tinydnsbackend.cc
@@ -163,9 +163,10 @@ void TinyDNSBackend::getAllDomains(vector<DomainInfo> *domains, bool include_dis
 
   d_cdbReader->searchAll();
   DNSResourceRecord rr;
+  std::unordered_set<DNSName> dupcheck;
 
   while (get(rr)) {
-    if (rr.qtype.getCode() == QType::SOA) {
+    if (rr.qtype.getCode() == QType::SOA && dupcheck.insert(rr.qname).second) {
       SOAData sd;
       fillSOAData(rr.content, sd);
 

### CODE REVIEW

File: modules/tinydnsbackend/tinydnsbackend.cc [hunk 1]
```diff
@@ -166,6 +166,8 @@ void TinyDNSBackend::getAllDomains(vector<DomainInfo> *domains, bool include_dis
 
   while (get(rr)) {
     if (rr.qtype.getCode() == QType::SOA) {
+      if (std::find_if(domains->begin(), domains->end(), [rr](const DomainInfo& d){ return rr.qname == d.zone; }) != domains->end())
```
rgacogne, line 169: It looks like the complexity is terrible (O(n^2)), which could be an issue with very large datasets. Perhaps we could use an `unordered_set` instead?

cmouse: True.. it's possible, but would though mean that someone is keeping several thousand domains in tinydns. But sure, I'll change it to unordered_set instead.

modules/tinydnsbackend/tinydnsbackend.cc
```diff
@@ -164,6 +164,8 @@ void TinyDNSBackend::getAllDomains(vector<DomainInfo> *domains, bool include_dis
   d_cdbReader->searchAll();
   DNSResourceRecord rr;
 
+  std::unordered_set<DomainInfo> result;
+
   while (get(rr)) {
     if (rr.qtype.getCode() == QType::SOA) {
```
mind04, line 170: You can replace the entire pull by this
 ```
std::unordered_set<DNSName> dupcheck;

 while (get(rr)) {
    if (rr.qtype.getCode() == QType::SOA && dupcheck.insert(rr.qname).second) {
```

Comment (cmouse) [reply modules/tinydnsbackend/tinydnsbackend.cc:170]: i didn't want to use dupcheck, because I feel we should be using unordered_set for the getAllDomains instead of Vector.

Comment (cmouse) [reply modules/tinydnsbackend/tinydnsbackend.cc:170]: but i didn't feel like doing it for getAllDomains now.

File: modules/tinydnsbackend/tinydnsbackend.cc [hunk 3]
```diff
@@ -177,9 +179,14 @@ void TinyDNSBackend::getAllDomains(vector<DomainInfo> *domains, bool include_dis
       di.notified_serial = sd.serial;
       di.kind = DomainInfo::Master;
       di.last_check = time(0);
-      domains->push_back(di);
+      if (result.find(di) != result.end())
+        continue;
+      result.insert(di);
```
rgacogne, line 184: You could always insert instead of looking up first, by the way ;-)

cmouse: what happens if i do that though

modules/tinydnsbackend/tinydnsbackend.cc
```diff
@@ -177,9 +179,14 @@ void TinyDNSBackend::getAllDomains(vector<DomainInfo> *domains, bool include_dis
...
     }
   }
+
```
rgacogne, line 187: Just a nit but we could call `domains->reserve(result.size())` as a small optimization :-)

modules/tinydnsbackend/tinydnsbackend.cc
rgacogne: The insertion will be prevented and the existing value will remain in the set.

cmouse: huh. I would've expected exception at least.

modules/tinydnsbackend/tinydnsbackend.cc
```diff
@@ -177,9 +179,13 @@ void TinyDNSBackend::getAllDomains(vector<DomainInfo> *domains, bool include_dis
       di.notified_serial = sd.serial;
       di.kind = DomainInfo::Master;
       di.last_check = time(0);
-      domains->push_back(di);
+      result.insert(di);
     }
   }
+
+  domains->reserve(result.size);
```
rgacogne, line 186:
```suggestion
  domains->reserve(result.size());
```

rgacogne (approved):
The PR looks good to me. I wonder however if the hash and equality operator for `DomainInfo` should not take the `id` field account? It doesn't matter for this PR but I'm worried it might come back to bite us later..

zeha:
> I wonder however if the hash and equality operator for `DomainInfo` should not take the `id` field account?

I guess the operators are there only for `unordered_set`? If so, at least equality should be defined specifically for _this_ `unordered_set` instance. I don't think we -generally- want to treat two DomainInfos with differing SOA values equal, etc.

> It doesn't matter for this PR but I'm worried it might come back to bite us later..

Yeah.

mind04:
I wonder if this pull should be split in two separate pull requests. This one with the actual fix for this bug (which is 1,5 lines of code) and an other one with the di changes. 
This will make it possible to quickly fix this issue and it will allow us to give the di changes the attention they need.

Habbie:
> I wonder if this pull should be split in two separate pull requests. This one with the actual fix for this bug (which is 1,5 lines of code) and an other one with the di changes.

Yes please! @cmouse do you want to move the DI stuff to a separate PR?

cmouse:
Dropped the equality commit, I'll move that to new PR.

PR #9175, opened 2020-05-29
dnsdist: Add per-thread Lua FFI load-balancing policies

rgacogne:
### Short description

This allows Lua FFI load-balancing policies that don't need access to the global, shared Lua state to be lock-less, avoiding lock contention.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


wojas:
Do I understand correctly that this injects all dnsdist globals into every per-thread Lua context? Is it safe and useful to call those? If not, it's perhaps best to keep those Lua states as clean as possible to avoid confusion and mistakes.

I was originally thinking about requiring those per-thread functions to be pure Lua functions without access to the dnsdist Lua API, but perhaps that was naive for real world load balancing functions. Do you have an example of a policy that would require more than this?

rgacogne:
> Do I understand correctly that this injects all dnsdist globals into every per-thread Lua context? Is it safe and useful to call those? If not, it's perhaps best to keep those Lua states as clean as possible to avoid confusion and mistakes.

Not all of them, no, it injects bindings and constants, mostly. Most of these should be safe to use, a few of them might not be and should be of very little use for a load-balancing policy, but reviewing every single bindings and constants and separating the safe ones from the unsafe would be a huge task. 

> I was originally thinking about requiring those per-thread functions to be pure Lua functions without access to the dnsdist Lua API, but perhaps that was naive for real world load balancing functions. Do you have an example of a policy that would require more than this?

I don't understand, how do you access the content of the `cdata` objects passed to the load-balancing function without access to the dnsdist bindings?

rgacogne:
Rebased on master.

rgacogne:
Rebased to fix a conflict.

Issue #7795, opened 2019-05-08
Triple database connection on incoming IXFR requests


klaus3000:
- Program: Authoritative
 - Issue type: Bug report

### Short description
I currently debug a problem where I hit the connection limit on my database. Watchting the logs, it seems that for incoming IXFR, every XFR thread opens 3 DB connections - although IXFR is just a fallback to AXFR.

### Environment
 - Operating system: Ubuntu 14.04
 - Software version: 4.1.5
 - Software source: self compiled .deb package

### Expected behaviour
Only open a single DB connection

### Actual behaviour
For IXFR PDNS opens 3 DB connections

### Logs for AXFR and IXFR
Incoming AXFR: dig @1.2.3.29 nxdomain.at axfr
```
: AXFR of domain 'nxdomain.at' initiated by 1.2.3.27
: AXFR of domain 'nxdomain.at' allowed: client IP 1.2.3.27 is in allow-axfr-ips
: gpgsql Connection successful. Connected to database 'pdns' on ''.
: AXFR of domain 'nxdomain.at' to 1.2.3.27 finished
```

Incoming IXFR with different serial: dig @1.2.3.29 nxdomain.at ixfr=0
```
: IXFR of domain 'nxdomain.at' initiated by 1.2.3.27 with serial 0
: AXFR of domain 'nxdomain.at' allowed: client IP 1.2.3.27 is in allow-axfr-ips
: gpgsql Connection successful. Connected to database 'pdns' on ''.
: gpgsql Connection successful. Connected to database 'pdns' on ''.
: IXFR fallback to AXFR for domain 'nxdomain.at' our serial 2019031507
: AXFR of domain 'nxdomain.at' initiated by 1.2.3.27
: AXFR of domain 'nxdomain.at' allowed: client IP 1.2.3.27 is in allow-axfr-ips
: gpgsql Connection successful. Connected to database 'pdns' on ''.
: AXFR of domain 'nxdomain.at' to 1.2.3.27 finished
```

Incoming IXFR with identical serial: dig @1.2.3.29 nxdomain.at ixfr=2019031507
```
: IXFR of domain 'nxdomain.at' initiated by 1.2.3.27 with serial 2019031507
: AXFR of domain 'nxdomain.at' allowed: client IP 1.2.3.27 is in allow-axfr-ips
: gpgsql Connection successful. Connected to database 'pdns' on ''.
: gpgsql Connection successful. Connected to database 'pdns' on ''.
[...]

PR #9176, opened 2020-05-29
auth-4.3.x: Optimize IXFR-to-AXFR fallback path

zeha:
### Short description

Backport of #9163 to 4.3.x.

Avoid making new backends when we are going to either deny the XFR, or
fall back to AXFR anyway.

This cuts down the number of new backends from four (three for IXFR
pre-checks plus one for AXFR) to one (just the AXFR one).
When replying in IXFR mode, we keep making _one_ new backend, which is
also better than before.

While we now hold the s_plock for a while longer, we only take it once
in doIXFR; before we took it twice -- for TSIG retrieval, which now
re-uses the IXFR backend.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/tcpreceiver.cc
+++ b/pdns/tcpreceiver.cc
@@ -1094,8 +1094,10 @@ int TCPNameserver::doIXFR(std::unique_ptr<DNSPacket>& q, int outsock)
 
   g_log<<Logger::Error<<"IXFR of domain '"<<q->qdomain<<"' initiated by "<<q->getRemote()<<" with serial "<<serial<<endl;
 
-  // determine if zone exists and AXFR is allowed using existing backend before spawning a new backend.
+  // determine if zone exists, XFR is allowed, and if IXFR can proceed using existing backend before spawning a new backend.
   SOAData sd;
+  bool securedZone;
+  bool serialPermitsIXFR;
   {
     Lock l(&s_plock);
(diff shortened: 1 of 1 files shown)


zeha:
Addresses #7795 for 4.3.x. Note: #9163 is not merged yet.

pdns/tcpreceiver.cc
```diff
@@ -1111,48 +1113,42 @@ int TCPNameserver::doIXFR(std::unique_ptr<DNSPacket>& q, int outsock)
...
-  if (!rfc1982LessThan(serial, calculateEditSOA(sd.serial, dk, sd.qname))) {
+  if (serialPermitsIXFR) {
+    DNSName target = q->qdomain;
     TSIGRecordContent trc;
     DNSName tsigkeyname;
     string tsigsecret;
 
+    UeberBackend db;
+    DNSSECKeeper dk(&db);
+    DNSSECKeeper::clearCaches(target);
```
mind04, line 1142: I don't think it is necessary to clean the caches here. There is an other DNSSECKeeper::clearCaches();  a few lines up.

zeha: right. fixed this in #9163, will cherry-pick here once 9163 is merged

Habbie: 9163 is merged

zeha: cherry-picked here

PR #9177, opened 2020-05-29
auth: when a backend ships us multiple SOAs, explicitly take the first one

Habbie:
### Short description
See https://github.com/PowerDNS/pdns/issues/9173#issuecomment-635923565

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


pdns/dnsbackend.cc
```diff
@@ -230,6 +230,8 @@ bool DNSBackend::getSOA(const DNSName &domain, SOAData &sd)
   int hits=0;
 
   while(this->get(rr)) {
+    if(hits>0)
+      continue;
```
zeha, line 234: should this be after the qtype check?

mind04: If this is necessary for a specific backed i think you should fix this in the backend.
And put a if(hits>0) throw PDNSException("say something ugly"); here

Habbie: yes, makes sense. I'll change that after #9174 lands

pieterlexis (changes requested):
Could use a test where the second SOA is put into the database via SQL.

Habbie:
Test pushed, but need to wait for CI to see if I did the backend exclusions correctly (for simplicity, this test is only done for SQLite).

klaus-nicat:
Ist it really necessary to make PDNS more strict then necessary? Since years PowerDNS worked fine with multiple SOAs (or other duplicate RRs). I remember the PDNS bug where overlapping incoming zone transfers generated duplicate RRs. In our setup we still have race conditions (hardly but we have them) were overlapping provisioning requests or failing DB-triggers lead to duplicate RRs or duplicate SOAs [1].

We catch them via cron jobs and fix the provisioing issues manually - but we have those issue. But having PowerDNS throwing an exception is IMO to hard, as this immediately makes a domain unusable in case of provisionig errors in the DB. I guess plenty of PowerDNS+DB-backend users will have severe problems when they update to the next PowerDNS release.

If you really want to throw an expception please make a config option to allow duplicate SOAs.

thanks
Klaus


[1] Our records/domains tables are partitioned. Postgres does not support foreign keys over partitions. Hence, we have to delete from records via triggers when a domain is deleted from domains table. And where we have sometimes race conditions that cleanup does not work as expected.

Habbie:
Proposal: we log a warning, and make sure `check-zone` notices this.

mind04:
Proposal2: Output an error in check-zone and check a zone before a b2b-migrate.

PR #9182, opened 2020-06-01
Implemented prometheus metrics-endpoint for auth

### LINKED ISSUE #4947: auth: Prometheus statistics endpoint
 - Issue type: Feature request 

### Short description
[Prometheus](http://prometheus.io/) is a statistics collector that pulls metrics over HTTP(S). The [format](https://prometheus.io/docs/instrumenting/writing_clientlibs/) to be returned is quite simple and adding new metrics is trivial.

The question is: Do we want to implement a prometheus endpoint (path) in the built-in webserver, or will we only support carbon output and all other outputs by means of 3rd party programs?

See also #3935.

supervacuus:
### Short description
This PR introduces a simple metrics-endpoint to `auth` in a similar fashion as was done in #6901 for `dnsdist` (and should fix #4947). The metrics reflected on the end-point are all of the `stats` and `funcstats` in the global `StatBag` instance (which should be the same as the "variables" listed in the webservers index page).

Due to a different metrics infrastructure in the `auth` compared to `dnsdist`, I took the liberty to extend the `StatBag` with a `StatType` map to configure the metric-type at the point where it is declared. This might not be what you want, please let me know if you'd prefer this configuration in a separate structure.

### Checklist
I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/docs/http-api/index.rst
+++ b/docs/http-api/index.rst
@@ -79,7 +79,7 @@ A simple ``GET`` request will return a response similar to the following:
   # TYPE pdns_auth_packetcache_miss counter
   pdns_auth_packetcache_miss 0
   # HELP pdns_auth_packetcache_size Number of entries in the packet cache
-  # TYPE pdns_auth_packetcache_size counter
+  # TYPE pdns_auth_packetcache_size gauge
   pdns_auth_packetcache_size 0
   # HELP pdns_auth_query_cache_hit Number of hits on the query cache
   # TYPE pdns_auth_query_cache_hit counter
@@ -88,7 +88,7 @@ A simple ``GET`` request will return a response similar to the following:
   # TYPE pdns_auth_query_cache_miss counter
   pdns_auth_query_cache_miss 0
   # HELP pdns_auth_query_cache_size Number of entries in the query cache
-  # TYPE pdns_auth_query_cache_size counter
+  # TYPE pdns_auth_query_cache_size gauge
   pdns_auth_query_cache_size 0
   # HELP pdns_auth_rd_queries Number of recursion desired questions
   # TYPE pdns_auth_rd_queries counter
@@ -103,7 +103,7 @@ A simple ``GET`` request will return a response similar to the following:
   # TYPE pdns_auth_recursion_unanswered counter
   pdns_auth_recursion_unanswered 0
   # HELP pdns_auth_security_status Security status based on regular polling
-  # TYPE pdns_auth_security_status counter
+  # TYPE pdns_auth_security_status gauge
   pdns_auth_security_status 0
   # HELP pdns_auth_servfail_packets Number of times a server-failed packet was sent out
   # TYPE pdns_auth_servfail_packets counter
--- a/docs/http-api/index.rst
+++ b/docs/http-api/index.rst
@@ -21,6 +21,276 @@ The following webserver related configuration items are available:
 * :ref:`setting-webserver-allow-from`: Netmasks that are allowed to connect to the webserver
 * :ref:`setting-webserver-max-bodysize`: Maximum request/response body size in megabytes
 
+
+Metrics Endpoint
+----------------
+
+The webserver exposes a metrics-endpoint that follows the `prometheus exposition-format <https://github.com/prometheus/docs/blob/master/content/docs/instrumenting/exposition_formats.md>`_ on path ``/metrics``.
+
+The metrics listed are equivalent to the variables section on the index-page of the webserver (prefixed with ``pdns_auth_`` and replacing dashes with underscores).
+
+A simple ``GET`` request will return a response similar to the following:
+
+.. http:get:: /metrics
+
+  HTTP/1.1 200 OK
+  Connection: close
+  Content-Length: 12044
+  Content-Type: text/plain
+  Server: PowerDNS/0.0.19015.0.master.ge719aae4e8
+
+  # HELP pdns_auth_corrupt_packets Number of corrupt packets received
+  # TYPE pdns_auth_corrupt_packets counter
+  pdns_auth_corrupt_packets 0
+  # HELP pdns_auth_deferred_cache_inserts Amount of cache inserts that were deferred because of maintenance
+  # TYPE pdns_auth_deferred_cache_inserts counter
+  pdns_auth_deferred_cache_inserts 0
+  # HELP pdns_auth_deferred_cache_lookup Amount of cache lookups that were deferred because of maintenance
+  # TYPE pdns_auth_deferred_cache_lookup counter
+  pdns_auth_deferred_cache_lookup 0
+  # HELP pdns_auth_deferred_packetcache_inserts Amount of packet cache inserts that were deferred because of maintenance
+  # TYPE pdns_auth_deferred_packetcache_inserts counter
+  pdns_auth_deferred_packetcache_inserts 0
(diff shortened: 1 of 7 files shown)


docs/http-api/index.rst
```diff
@@ -21,6 +21,276 @@ The following webserver related configuration items are available:
...
+  # TYPE pdns_auth_overload_drops counter
+  pdns_auth_overload_drops 0
+  # HELP pdns_auth_packetcache_hit Number of hits on the packet cache
+  # TYPE pdns_auth_packetcache_hit counter
+  pdns_auth_packetcache_hit 0
+  # HELP pdns_auth_packetcache_miss Number of misses on the packet cache
+  # TYPE pdns_auth_packetcache_miss counter
+  pdns_auth_packetcache_miss 0
+  # HELP pdns_auth_packetcache_size Number of entries in the packet cache
+  # TYPE pdns_auth_packetcache_size counter
```
rgacogne, line 82:
```suggestion
  # TYPE pdns_auth_packetcache_size gauge
```

docs/http-api/index.rst
```diff
@@ -21,6 +21,276 @@ The following webserver related configuration items are available:
...
+  pdns_auth_packetcache_size 0
+  # HELP pdns_auth_query_cache_hit Number of hits on the query cache
+  # TYPE pdns_auth_query_cache_hit counter
+  pdns_auth_query_cache_hit 0
+  # HELP pdns_auth_query_cache_miss Number of misses on the query cache
+  # TYPE pdns_auth_query_cache_miss counter
+  pdns_auth_query_cache_miss 0
+  # HELP pdns_auth_query_cache_size Number of entries in the query cache
+  # TYPE pdns_auth_query_cache_size counter
```
rgacogne, line 91:
```suggestion
  # TYPE pdns_auth_query_cache_size gauge
```

docs/http-api/index.rst
```diff
@@ -21,6 +21,276 @@ The following webserver related configuration items are available:
...
+  # TYPE pdns_auth_recursing_answers counter
+  pdns_auth_recursing_answers 0
+  # HELP pdns_auth_recursing_questions Number of questions sent to recursor
+  # TYPE pdns_auth_recursing_questions counter
+  pdns_auth_recursing_questions 0
+  # HELP pdns_auth_recursion_unanswered Number of packets unanswered by configured recursor
+  # TYPE pdns_auth_recursion_unanswered counter
+  pdns_auth_recursion_unanswered 0
+  # HELP pdns_auth_security_status Security status based on regular polling
+  # TYPE pdns_auth_security_status counter
```
rgacogne, line 106:
```suggestion
  # TYPE pdns_auth_security_status gauge
```

docs/http-api/index.rst
```diff
@@ -21,6 +21,276 @@ The following webserver related configuration items are available:
...
+  # TYPE pdns_auth_qsize_q gauge
+  pdns_auth_qsize_q 0
+  # HELP pdns_auth_real_memory_usage Actual unique use of memory in bytes (approx)
+  # TYPE pdns_auth_real_memory_usage gauge
+  pdns_auth_real_memory_usage 133189632
+  # HELP pdns_auth_ring_logmessages_capacity Maximum number of entries in the logmessages ring
+  # TYPE pdns_auth_ring_logmessages_capacity counter
+  pdns_auth_ring_logmessages_capacity 10000
+  # HELP pdns_auth_ring_logmessages_size Number of entries in the logmessages ring
+  # TYPE pdns_auth_ring_logmessages_size counter
```
rgacogne, line 205: I think these entries can be cleared from the web server, so likely a gauge?

docs/http-api/index.rst
supervacuus: You are absolutely right. I wrongly assumed that all call-sites accessing `d_stats' use the `StatBag::inc` interface and forgot about `StatBag::getPointer` and `StatBag::set`. I checked against those interfaces again and found the same entries as you did. Thanks for hint. Will fix in the next commit.

docs/http-api/index.rst
supervacuus: If the ring `size` metrics can be `reset()` to zero entries, but otherwise does not result in a negative difference between data-points, then according to [prometheus-docs](https://prometheus.io/docs/concepts/metric_types/#counter) they would still be considered a counter.

But if someone changes the ring `capacity` to a lower value via `resize()`, then the `capacity` metric would be considered a gauge.

Should I change this accordingly?

docs/http-api/index.rst
supervacuus: Agreed, will fix in the next commit.

docs/http-api/index.rst
supervacuus: Agreed, will fix in the next commit.

docs/http-api/index.rst
rgacogne: I think you are right, since it can only be set to zero but not decremented, a counter type might work.

supervacuus: Should I leave the ring `capacity` metrics as counters or change them to gauges? I think they would make more sense as gauges.

rgacogne: `Capacity` metrics should clearly be gauges :+1:

supervacuus: Fixed in latest commit.

rgacogne:
Looks good to me. Perhaps @Habbie and/or @pieterlexis want to take a look as well?

Habbie:
I will!

pdns/statbag.cc
```diff
@@ -237,8 +244,8 @@ vector<pair<T, unsigned int> >StatRing<T,Comp>::get() const
 
 void StatBag::registerRingStats(const string& name)
 {
-  declare("ring-" + name + "-size", "Number of entries in the " + name + " ring", [this,name](const std::string&) { return static_cast<uint64_t>(getRingEntriesCount(name)); });
-  declare("ring-" + name + "-capacity", "Maximum number of entries in the " + name + " ring", [this,name](const std::string&) { return static_cast<uint64_t>(getRingSize(name)); });
+  declare("ring-" + name + "-size", "Number of entries in the " + name + " ring", [this,name](const std::string&) { return static_cast<uint64_t>(getRingEntriesCount(name)); }, StatType::counter);
```
Habbie, line 247: gauge, I think?

supervacuus: @rgacogne raised the same issue here: https://github.com/PowerDNS/pdns/pull/9182#discussion_r433730340

I am happy either way, but following the [prometheus docs](https://prometheus.io/docs/concepts/metric_types/#counter) the ring-buffer `size` metric can be considered a counter. It either increments and saturates or the buffer is `reset()` and `size` is 0.

Since the ring-buffer can be resized, we considered `capacity` as a gauge. But `resize()` will also affect `size` if the resized buffer is smaller than the number of items it currently holds. Which would make `size` a gauge.

supervacuus: When considering the case of `resize()` to a `capacity` smaller than the current `size` it is clear, that `size` must be a gauge too. Fixed in the latest commit.

Habbie: ah, sorry, I did see that comment (but did not read it carefully). I see now how the definitions of 'gauge' and 'metric' in Prometheus do not exactly match my intuition of the terms.

resize to smaller but non-zero seems like a sensible argument, thanks!

PR #9183, opened 2020-06-02
API: Allow rectifying Slave zones

### LINKED ISSUE #9066: Rectify via API not allowed for slave zones
- Program: Authoritative
 - Issue type: Bug report

### Short description

PowerDNS supports signing of slave zones. For this, slave zones need to be recitified. This works fine on incoming AXFR. If an existing slave zone is switched from unsigned to signed (signed on the PDNS slave), the zone needs to be rectified locally. This work fine with pdnsutil, but not via the REST API.

### Environment
 - Operating system: Ubuntu
 - Software version: 4.1 and trunk

### Steps to reproduce
1. add a slave zone
2. add keys for the slave zone (we add them manually directly to cryptokeys table)
3. rectify via API

### Expected behaviour
Zone should be rectified

### Actual behaviour
Recitify is rejected: 
https://github.com/PowerDNS/pdns/blob/0ddde5f/pdns/ws-auth.cc#L1866

### Other information
A workaround would be to retransfer the zone, but this is unreliable as the master may not be available when DNSSEC is enabled for this zone.

zeha:
## Short description

Fixes #9066.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/ws-auth.cc
+++ b/pdns/ws-auth.cc
@@ -1863,9 +1863,6 @@ static void apiServerZoneRectify(HttpRequest* req, HttpResponse* resp) {
   if (!dk.isSecuredZone(zonename))
     throw ApiException("Zone '" + zonename.toString() + "' is not DNSSEC signed, not rectifying.");
 
-  if (di.kind == DomainInfo::Slave)
-    throw ApiException("Zone '" + zonename.toString() + "' is a slave zone, not rectifying.");
-
   string error_msg = "";
   string info;
   if (!dk.rectifyZone(zonename, error_msg, info, true))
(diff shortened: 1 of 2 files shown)


pdns/ws-auth.cc
```diff
@@ -1863,9 +1863,6 @@ static void apiServerZoneRectify(HttpRequest* req, HttpResponse* resp) {
   if (!dk.isSecuredZone(zonename))
```
mind04, line 1863: Unrelated to the pull itself, but not allowing rectify for non dnssec zones is also unwanted.
I think we should remove this limitation as well.

zeha: Maybe this should actually be `if (!dk.isPresigned...`?

mind04: Yes, rectifying presigned zones will not end well.

zeha: pushed followup commit for this

Issue #5292, opened 2017-05-04
add query parameters to backend query logging


klaus3000:
Since 4.0 the query logging of the backend (query-logging=yes) the logs do not show parameters anymore, only the prepared statements, e.g.:
pdns <4:
select ...  and name=E'dev-nsec.rc0-testing.dnssec-signiert.at' and domainmetadata.kind=E'PRESIGNED'

pdns>=4:
select ... and name=$1 and domainmetadata.kind=$2

The old version was much more helpful while debugging. Please add some parameter logging too, if possible

klaus3000:
IIRC in PDNS 4.3 the prepared statements were removed. So maybe this issue is obsolete, if now the logging contains the actual query parameters again. (I have not deployed 4.3 yet)

Habbie:
We still use parameter binding, so I doubt this is fixed.

zeha:
This is fixed for gpgsqlbackend, but not for gsqlite/gmysql/godbc.

klaus3000:
For me, the problem is fixed by #9189. Hence I will close this issue.
If you think that should be addressed also for other backends, please reopen the issue or create a new one.

Habbie:
Thanks for reporting back, Klaus. I'll reopen for other backends, indeed.

PR #9189, opened 2020-06-02
gpgsqlbackend: add parameters to query logging

zeha:
### Short description

Addresses #5292 (for postgres only).

Example logs:
```
Jun 02 19:23:19 Query 140202644522992: Statement: SELECT content,ttl,prio,type,domain_id,disabled::int,name,auth::int FROM records WHERE disabled=false and type=$1 and name=$2
Jun 02 19:23:19 Query 140202644522992: Parameters: $1 = 'SOA', $2 = 'example.com'
```

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/modules/gpgsqlbackend/spgsql.cc
+++ b/modules/gpgsqlbackend/spgsql.cc
@@ -77,7 +77,19 @@ public:
   SSqlStatement* execute() {
     prepareStatement();
     if (d_dolog) {
-      g_log<<Logger::Warning<< "Query "<<((long)(void*)this)<<": " << d_query << endl;
+      g_log<<Logger::Warning<< "Query "<<((long)(void*)this)<<": Statement: " << d_query << endl;
+      if (d_paridx) {
+        // Log message is similar, bot not exactly the same as the postgres server log.
+        std::stringstream log_message;
+        log_message<< "Query "<<((long)(void*)this)<<": Parameters: ";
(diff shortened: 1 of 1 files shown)


modules/gpgsqlbackend/spgsql.cc
```diff
@@ -77,7 +77,19 @@ class SPgSQLStatement: public SSqlStatement
   SSqlStatement* execute() {
     prepareStatement();
     if (d_dolog) {
-      g_log<<Logger::Warning<< "Query "<<((long)(void*)this)<<": " << d_query << endl;
+      g_log<<Logger::Warning<< "Query "<<((long)(void*)this)<<": Statement: " << d_query << endl;
```
Habbie, line 80: hmm, how about no `endl` here? Not a strong opinion but wanted to ask.

zeha: I wanted this to look somewhat similar to the postgres server logs, which also puts statement and params on separate lines. Might also be nicer (or worse?) for grepping.

Habbie: Yes, goes both ways for grepping. I see that we use the same prefix (numberised `this`) for more than just these two lines, so splitting here is fine too.

PR #9190, opened 2020-06-02
gpgsql: Reintroduce prepared statements

### LINKED ISSUE #9029: Make prepared statements against the gpgsql backend optional
- Program: PowerDNS Authoritative 4.3
 - Issue type: Feature request

### Short description
PowerDNS 4.3 abandoned the use of prepared statements for the gpgsql backend. This has a bigger impact on performance in some situations.

### Usecase
If you are not using a connection pooler such as pgbouncer it would be great to get the better performance by being able to use prepared statements.

### Description
I recently upgraded from PowerDNS auth 4.2.1 to PowerDNS auth 4.3, and was a bit surprised to see latency reported by PowerDNS go up from ~2.5 ms to around 4.2 ms.  I know we are talking milliseconds, but they matter when it comes to DNS.

Here's a graph of the latency of one of our DNS servers before and after the upgrade to PowerDNS 4.3:
powerdns-latency

We don't use pgbouncer, so it would be great to be able to enable prepared statements again.

zeha:
### Short description

Fixes #9029.

Defaults to using prepared statements, but they can be turned off for pgBouncer (et al) users.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/modules/gpgsqlbackend/spgsql.cc
+++ b/modules/gpgsqlbackend/spgsql.cc
@@ -39,18 +39,8 @@ public:
     d_query = query;
     d_dolog = dolog;
     d_parent = db;
-    d_prepared = false;
     d_nparams = nparams;
-    d_res = NULL;
-    d_res_set = NULL;
-    paramValues = NULL;
-    paramLengths = NULL;
     d_nstatement = nstatement;
-    d_paridx = 0;
-    d_residx = 0;
-    d_resnum = 0;
-    d_fnum = 0;
-    d_cur_set = 0;
   }
 
   SSqlStatement* bind(const string& name, bool value) { return bind(name, string(value ? "t" : "f")); }
@@ -82,9 +72,9 @@ public:
       d_dtime.set();
     }
     if (!d_stmt.empty()) {
-      d_res_set = PQexecPrepared(d_db(), d_stmt.c_str(), d_nparams, paramValues, paramLengths, NULL, 0);
+      d_res_set = PQexecPrepared(d_db(), d_stmt.c_str(), d_nparams, paramValues, paramLengths, nullptr, 0);
     } else {
-      d_res_set = PQexecParams(d_db(), d_query.c_str(), d_nparams, NULL, paramValues, paramLengths, NULL, 0);
+      d_res_set = PQexecParams(d_db(), d_query.c_str(), d_nparams, nullptr, paramValues, paramLengths, nullptr, 0);
     }
     ExecStatusType status = PQresultStatus(d_res_set);
     if (status != PGRES_COMMAND_OK && status != PGRES_TUPLES_OK && status != PGRES_NONFATAL_ERROR) {
@@ -103,10 +93,10 @@ public:
   }
 
   void nextResult() {
-    if (d_res_set == NULL) return; // no refcursor
+    if (d_res_set == nullptr) return; // no refcursor
     if (d_cur_set >= PQntuples(d_res_set)) {
       PQclear(d_res_set);
-      d_res_set = NULL;
+      d_res_set = nullptr;
       return;
     }
     // this code handles refcursors if they are returned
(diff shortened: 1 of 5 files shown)


modules/gpgsqlbackend/spgsql.cc
```diff
@@ -80,7 +81,11 @@ class SPgSQLStatement: public SSqlStatement
       g_log<<Logger::Warning<< "Query "<<((long)(void*)this)<<": " << d_query << endl;
       d_dtime.set();
     }
-    d_res_set = PQexecParams(d_db(), d_query.c_str(), d_nparams, NULL, paramValues, paramLengths, NULL, 0);
+    if (!d_stmt.empty()) {
+      d_res_set = PQexecPrepared(d_db(), d_stmt.c_str(), d_nparams, paramValues, paramLengths, NULL, 0);
+    } else {
+      d_res_set = PQexecParams(d_db(), d_query.c_str(), d_nparams, NULL, paramValues, paramLengths, NULL, 0);
```
rgacogne, line 87:
```suggestion
      d_res_set = PQexecParams(d_db(), d_query.c_str(), d_nparams, nullptr, paramValues, paramLengths, NULL, 0);
```

modules/gpgsqlbackend/spgsql.cc
```diff
@@ -207,10 +212,28 @@ class SPgSQLStatement: public SSqlStatement
...
+      d_stmt.clear();
+    }
   }
 
   void prepareStatement() {
     if (d_prepared) return;
+    if (d_parent->usePrepared()) {
+      // prepare a statement; name must be unique per session (using d_nstatement to ensure this).
+      this->d_stmt = string("stmt") + std::to_string(d_nstatement);
+      PGresult* res = PQprepare(d_db(), d_stmt.c_str(), d_query.c_str(), d_nparams, NULL);
```
rgacogne, line 228:
```suggestion
      PGresult* res = PQprepare(d_db(), d_stmt.c_str(), d_query.c_str(), d_nparams, nullptr);
```

zeha:
Addressed the nits!

Habbie:
@sfrost could I convince you to review this? :)

sfrost:
Not sure when I'll have a chance to really review it, but adding in the prepared statements as an option seems like a good thing to me- pretty sure I had asked for that initially and then folks thought we could get away with just not bothering to do it at all.

I'd think that, in an ideal world, statements would be prepared with distinct/known names and then used.  If there's an error returned because the prepared statement doesn't already exist on the connection then it could be prepared and then used automatically.

Habbie:
>  pretty sure I had asked for that initially and then folks thought we could get away with just not bothering to do it at all.

Yes. We were so very wrong!

Habbie:
> I'd think that, in an ideal world, statements would be prepared with distinct/known names and then used. If there's an error returned because the prepared statement doesn't already exist on the connection then it could be prepared and then used automatically.

That appears to be impossible: https://github.com/PowerDNS/pdns/issues/9029#issuecomment-614141270

sfrost:
Uh, why not just accept that you got an error and then try to prepare it and run it again?  If it fails again, or the prepare fails, then fail out.

Why do you need to bother parsing the error or doing anything further?

Habbie:
That sounds like a recipe for lots of extra roundtrips in some setups.

(we are discussing on a closed ticket, replies may not be seen)

rgacogne:
I agree it would make sense to try to prepare and run it again if could be sure that the connection is going to be long-lived, which is true for some of them but unfortunately not all of them, far from it. Doing that for short-lived connections would likely increase latency quite significantly. 
In a ideal situation we would know the expected duration of a connection in advance, perhaps not bothering to prepare statements for short-lived connections, but being ready to invest a bit more time preparing and re-preparing statements for other ones, but at the moment we don't have that information.

PR #9196, opened 2020-06-03
rec: Disable outgoing v4 when QLA has no v4 addresses

pieterlexis:
### Short description
The documentation actually mentions that not configuring a certain address family in `query-local-address` means no queries are sent out from that address family. This was only true for IPv4 though. This PR ensures this is true for v6 only installations as well.

### Checklist
I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/pdns/pdns_recursor.cc
+++ b/pdns/pdns_recursor.cc
@@ -4729,6 +4729,7 @@ try
   if (threadInfo.isHandler) {
     if (!primeHints()) {
       threadInfo.exitCode = EXIT_FAILURE;
+      RecursorControlChannel::stop = 1;
       g_log<<Logger::Critical<<"Priming cache failed, stopping"<<endl;
       return nullptr;
     }
--- a/pdns/pdns_recursor.cc
+++ b/pdns/pdns_recursor.cc
@@ -4166,7 +4166,7 @@ static int serviceMain(int argc, char*argv[])
     g_log<<Logger::Warning<<"Enabling IPv4 transport for outgoing queries"<<endl;
   }
   else {
-    g_log<<Logger::Warning<<"NOT using IPv6 for outgoing queries - add an IPv4 address (like '0.0.0.0') to query-local-address to enable"<<endl;
+    g_log<<Logger::Warning<<"NOT using IPv4 for outgoing queries - add an IPv4 address (like '0.0.0.0') to query-local-address to enable"<<endl;
   }
 
 
@@ -4176,6 +4176,11 @@ static int serviceMain(int argc, char*argv[])
   }
   else {
     g_log<<Logger::Warning<<"NOT using IPv6 for outgoing queries - add an IPv6 address (like '::') to query-local-address to enable"<<endl;
+  }
+
+  if (!(SyncRes::s_doIPv6 && SyncRes::s_doIPv4)) {
+    g_log<<Logger::Error<<"No outgoing addresses specified! Can not continue"<<endl;
+    exit(99);
   }
 
   // keep this ABOVE loadRecursorLuaConfig!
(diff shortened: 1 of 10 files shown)


pdns/pdns_recursor.cc
```diff
@@ -4102,6 +4102,15 @@ static int serviceMain(int argc, char*argv[])
     exit(99);
   }
 
+  if(pdns::isQueryLocalAddressFamilyEnabled(AF_INET)) {
+    SyncRes::s_doIPv4=true;
+    g_log<<Logger::Warning<<"Enabling IPv4 transport for outgoing queries"<<endl;
+  }
+  else {
+    g_log<<Logger::Warning<<"NOT using IPv6 for outgoing queries - add an IPv4 address (like '0.0.0.0') to query-local-address to enable"<<endl;
```
phonedph1, line 4110: s/IPv6/IPv4/ ?

pdns/recursordist/test-syncres_cc10.cc
```diff
@@ -0,0 +1,138 @@
...
+  sr->setAsyncCallback([target, &v4Hit, &v6Hit, &queries](const ComboAddress& ip, const DNSName& domain, int type, bool doTCP, bool sendRDQuery, int EDNS0Level, struct timeval* now, boost::optional<Netmask>& srcmask, boost::optional<const ResolveContext&> context, LWResult* res, bool* chained) {
+    queries++;
+    if (isRootServer(ip)) {
+      setLWResult(res, 0, false, false, true);
+
+      if (domain == DNSName("powerdns.com.")) {
+        addRecordToLW(res, domain, QType::NS, "ns1.powerdns.com.", DNSResourceRecord::AUTHORITY, 172800);
+        addRecordToLW(res, "ns1.powerdns.com.", QType::AAAA, "2001:DB8:1::53", DNSResourceRecord::ADDITIONAL, 3600);
+        addRecordToLW(res, "ns1.powerdns.com.", QType::A, "192.0.2.1", DNSResourceRecord::ADDITIONAL, 3600);
+      }
```
rgacogne, line 27: Don't we have a 50% chance of passing this test even if the "no v4" switch is broken? Perhaps we could check that the root server address used is v4 as well?

pdns/recursordist/test-syncres_cc10.cc
```diff
@@ -0,0 +1,138 @@
...
+  std::unique_ptr<SyncRes> sr;
+  initSR(sr);
+  SyncRes::s_doIPv4 = false;
+  SyncRes::s_doIPv6 = true;
+  primeHints();
+  int queries = 0;
+
+  const DNSName target("powerdns.com.");
+  sr->setAsyncCallback([target, &queries](const ComboAddress& ip, const DNSName& domain, int type, bool doTCP, bool sendRDQuery, int EDNS0Level, struct timeval* now, boost::optional<Netmask>& srcmask, boost::optional<const ResolveContext&> context, LWResult* res, bool* chained) {
+    cout<<ip.toString()<<endl;
```
rgacogne, line 109: I'm pretty sure the `cout` lines are not wanted here :-)

pieterlexis:
The failing test is because that test uses `query-local-address=::1` and the auths don't listen on ipv6.

omoerbeek:
Rebased and fixed a little bug that prevented error exit in unthreaded mode after failure to load hints.

omoerbeek:
@rgacogne could you check the new exit code from serviceMain? There are more cases to be converted later, but we first need to make sure the approach is right.

pdns/pdns_recursor.cc
```diff
@@ -179,6 +179,7 @@ struct RecThreadInfo
   bool isListener{false};
   /* process queries */
   bool isWorker{false};
+  int exitCode{0};
```
rgacogne, line 182: It would be nicer to move that above the booleans and below the `uint64_t` :-)

PR #9203, opened 2020-06-05
Rec: rpz policy should override gettag_ffi answer by default

omoerbeek:
### Short description

If an RPZ is active, an answer set by `gettag_ffi` should not be used unless there is no RPZ match. There are some subtle points;

1. If no policy is matched or it is `NoAction`, use the result from `gettag_ffi` if available.
2. You still might want to prefer the exising behaviour: set `overridesGettag=false` for that case when loading the RPZ.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [X] documented the code

### DIFF
--- a/pdns/recursordist/docs/lua-config/rpz.rst
+++ b/pdns/recursordist/docs/lua-config/rpz.rst
@@ -117,7 +117,7 @@ overridesGettag
 .. versionadded:: 4.4.0
 
 `gettag_ffi` can set an answer to a query.
-By default an RPZ hit overrides this answer, unless the policy is `rpz-passthru` or this option is set to `false`.
+By default an RPZ hit overrides this answer, unless this option is set to `false`.
 The default is `true`.
 
 zoneSizeHint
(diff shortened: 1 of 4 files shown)


pdns/recursordist/docs/lua-config/rpz.rst
```diff
@@ -112,6 +112,14 @@ tags
 
 List of tags as string, that will be added to the policy tags exported over protobuf when a policy of this zone matches.
 
+overridesGettag
+^^^^^^^^^^^^^^^
+.. versionadded:: 4.4.0
+
+`gettag_ffi` can set an answer to a query.
+By default an RPZ hit overrides this answer, unless the policy is `rpz-passthru` or this option is set to `false`.
```
rgacogne, line 120: I wonder if we really want to ignore `rpz-passthru` matches, which are a "whitelisting" mechanism? Would it make sense to only apply the answer set from `gettag_ffi` if there was no match or if this option is set to `false`?

omoerbeek: Yes, this is open for discussion. I'll sleep a night on it :-)

PR #9207, opened 2020-06-06
use std::string_view when available

neheb:
There's a standard C++ macro to check for its existence.

libstdc++ from GCC makes it available under C++17 and up. libcxx from
LLVM makes it available everywhere.

I have:
- [x ] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/ext/lmdb-safe/lmdb-safe.hh
+++ b/ext/lmdb-safe/lmdb-safe.hh
@@ -12,8 +12,9 @@
 #include <vector>
 #include <algorithm>
 
-// apple compiler somehow has string_view even in c++11!
-#if __cplusplus < 201703L && !defined(__APPLE__) && !defined(__OpenBSD__) && !defined(__FreeBSD__)
+#ifdef __cpp_lib_string_view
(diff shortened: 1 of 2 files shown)


omoerbeek:
Do you happen to have a reference where this define is standardized? I could not spot it on https://en.cppreference.com

neheb:
https://en.cppreference.com/w/cpp/feature_test

omoerbeek:
> https://en.cppreference.com/w/cpp/feature_test

Right. I read that as the feature macros are only available for C++20 and later.  but I checked the define is included for `g++ -std=c++17` so this works allright and is much nicerr than the existing condition.

neheb:
This is available in both GCC libstdcpp++ and LLVM libc++. The C++20 thing might be that the page is updated for that version. I’ll check the C++17 standard to see what it says later.

neheb:
I’ll also mention that the define is 2018 for GNU libstdc++ and 2016 for LLVM libcxx. I don’t think that makes a practical difference.

Issue #9206, opened 2020-06-05
dnsdist stops handling DoH after a while


krombel:
- Program: dnsdist
 - Issue type: Bug report

### Short description
after a while dnsdist stops responding to DoH-Queries (while all other protocols continue to work)

### Environment
 - Operating system: debian buster
 - Software version: 1.5.0-rc2
 - Software source: PowerDNS repository

### Steps to reproduce
I am not sure. It happens when we run dnsdist for a while.

### Other information
Some stats of dnsdist directly:
https://stats.ffmuc.net/d/tlvoghcZk/doh-dot?orgId=1&from=1591373515179&to=1591376176006
The probably the most relevant graph: DNS queries/minute
![image](https://user-images.githubusercontent.com/11167142/83910764-679e7a80-a76b-11ea-8f11-cbd10d25ae8e.png)
Green: All requests; yellow DoHv4; blue DNSCrypt; ...
Note: I have two dedicated listeners for IPv4 and IPv6 and only the IPv4-Listener broke. The DoHv6 listener kept answering (but gets hit that low that you cannot really see it in the graph)

nginx (which is reverse proxying DoH) reports following stats:
![image](https://user-images.githubusercontent.com/11167142/83910147-543edf80-a76a-11ea-8014-729d56c4c5be.png)

`strace -p <PID_of_dnsdist> -o dnsdist_faulty.strace.txt -f`
[dnsdist_faulty.strace.txt](https://github.com/PowerDNS/pdns/files/4737823/dnsdist_faulty.strace.txt)
`gdb -p <PID_of_dnsdist> -x gdb-script --batch > dnsdist_faulty.gdb.txt` where gdb-script contains
`thread apply all bt full`:
[dnsdist_faulty.gdb.txt](https://github.com/PowerDNS/pdns/files/4737824/dnsdist_faulty.gdb.txt)

rgacogne:
Leads that I investigated that seem to be dead ends:
- one thread stopping to work because of an exception not being handled: all threads are accounted for in GDB, nothing weird pops to mind ;
- the internal response socket pair getting full: the GDB trace shows a lot of threads in the process of writing to the response socket pair, which could have indicated that they are blocked because the buffer is full. The strace log does not confirm that, although it shows that some of the "new" DoH connections processed by h2o are in fact quite old and already closed on the OS side (`getpeername()` returns `ENOTCONN`), which seems to indicate that thread was indeed blocked for a short while, or at the very least completely overwhelmed. 

[...]

PR #9211, opened 2020-06-08
dnsdist: Use non-blocking pipes to pass DoH queries/responses around

rgacogne:
### Short description

This commit makes the internal sockets non-blocking so we don't freeze if they ever fill up, and log errors/increment metrics instead.

It also replaces the socket pairs by pipes, since the default buffer size for sockets seems to allow only ~278 pending queries which might be reached given how libh2o batches events. On Linux, a pipe gives us 8192 pending queries by default due to the lower overhead, and it can easily be incremented to 131072 pending queries by setting the pipe size to 1048576. This commits adds a new setting to do just that.

Might help with the issue reported in #9206.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdist-lua.cc
+++ b/pdns/dnsdist-lua.cc
@@ -1936,6 +1936,10 @@ static void setupLuaConfig(bool client, bool configCheck)
         frontend->d_trustForwardedForHeader = boost::get<bool>((*vars)["trustForwardedForHeader"]);
       }
 
+      if (vars->count("internalPipeBufferSize")) {
+        frontend->d_internalPipeBufferSize = boost::get<int>((*vars)["internalPipeBufferSize"]);
+      }
+
       parseTLSConfig(frontend->d_tlsConfig, "addDOHLocal", vars);
     }
     g_dohlocals.push_back(frontend);
--- a/pdns/dnsdist.cc
+++ b/pdns/dnsdist.cc
@@ -638,7 +638,12 @@ try {
 #ifdef HAVE_DNS_OVER_HTTPS
             // DoH query
             du->response = std::string(response, responseLen);
-            if (send(du->rsock, &du, sizeof(du), 0) != sizeof(du)) {
+            ssize_t sent = write(du->rsock, &du, sizeof(du));
+            if (sent != sizeof(du)) {
+              if (errno == EAGAIN || errno == EWOULDBLOCK) {
+                ++g_stats.dohResponsePipeFull;
+              }
+
               /* at this point we have the only remaining pointer on this
                  DOHUnit object since we did set ids->du to nullptr earlier,
                  except if we got the response before the pointer could be
--- a/pdns/dnsdistdist/doh.cc
+++ b/pdns/dnsdistdist/doh.cc
@@ -43,11 +43,11 @@ using namespace std;
    dnsdist worker thread which we also launched.
 
    This dnsdist worker thread injects the query into the normal dnsdist flow
-   (as a datagram over a socketpair). The response also goes back over a
-   (different) socketpair, where we pick it up and deliver it back to h2o.
+   (over a pipe). The response also goes back over a (different) pipe,
+   where we pick it up and deliver it back to h2o.
 
    For coordination, we use the h2o socket multiplexer, which is sensitive to our
-   socketpair too.
+   pipe too.
 */
 
 /* h2o notes.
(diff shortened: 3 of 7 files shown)


omoerbeek:
This looks good, although on OpenBSD en FreeBSD there is not way to set the pipe buffer size, while a socketpair has standardized socket options to set both receive and send buffer size. So I'm not 100% convinced moving to a pipe is the best way.

In the meantime I'll do some tests on OpenBSD.

rgacogne:
> This looks good, although on OpenBSD en FreeBSD there is not way to set the pipe buffer size, while a socketpair has standardized socket options to set both receive and send buffer size. So I'm not 100% convinced moving to a pipe is the best way.

The thing is that, on Linux at least, the default value for a socket pair allows up to ~277 queries. Increasing it to the maximum size without changing system-wide settings via `sysctl` yields a maximum of ~554 queries in my tests, ~87382 after a lot of system tuning.
The default value for a pipe is `8192` queries on Linux, `131072` by setting the pipe buffer size to `1048576` which requires not system tuning. I would also not be surprised if there was more overhead than just the amount of memory when sockets are used instead of pipes. And for what it's worth we have been using a pipe for that task in the recursor for quite some time now :-)

omoerbeek:
Could it be that the limits to buffer sizes are because the socketpair created specifiies `SOCK_DGRAM`?  I think a `SOCK_STREAM` socketpair would not have those strict limits. 

Come to think of it, it even *migh*t explain the root-cause of the issue: datagram socketpairs are not reliable, I suspect even for local sockets.

rgacogne:
> Could it be that the limits to buffer sizes are because the socketpair created specifies `SOCK_DGRAM`? I think a `SOCK_STREAM` socketpair would not have those strict limits.

Writing and reading to a `SOCK_STREAM` socket, even a local one, is not atomic. It is for a `SOCK_DGRAM` one.

rgacogne:
> Come to think of it, it even might explain the root-cause of the issue: datagram socketpairs are not reliable, I suspect even for local sockets.

UNIX domain sockets are usually reliable even in datagram mode on _most_ UNIX implementations, although I don't think this is actually required by POSIX.

omoerbeek:
There is also `SOCK_SEQPACKET`, but I agree, it is better to move to a pipe. Noting that atomicity is only guaranteed for writes up to `PIPE_BUF`, but `sizeof(uw)` is likely much smaller. A `static_assert` might be in place though.

omoerbeek:
As for reliability of AF_UNIX datagram socket pairs, Posix does not require that indeed.

pdns/dnsdist-lua.cc
```diff
@@ -1936,6 +1936,10 @@ static void setupLuaConfig(bool client, bool configCheck)
         frontend->d_trustForwardedForHeader = boost::get<bool>((*vars)["trustForwardedForHeader"]);
       }
 
+      if (vars->count("internalPipeBufferSize")) {
+        frontend->d_internalPipeBufferSize = boost::get<int>((*vars)["internalPipeBufferSize"]);
```
omoerbeek, line 1940:
```suggestion
        frontend->d_internalPipeBufferSize = boost::get<uint32_t>((*vars)["internalPipeBufferSize"]);
```

pdns/dnsdist.cc
```diff
@@ -638,7 +638,12 @@ try {
 #ifdef HAVE_DNS_OVER_HTTPS
             // DoH query
             du->response = std::string(response, responseLen);
-            if (send(du->rsock, &du, sizeof(du), 0) != sizeof(du)) {
+            ssize_t sent = write(du->rsock, &du, sizeof(du));
+            if (sent != sizeof(du)) {
+              if (errno == EAGAIN || errno == EWOULDBLOCK) {
+                ++g_stats.dohResponsePipeFull;
+              }
+
```
omoerbeek, line 646: maybe internal erorr on other errors?

pdns/dnsdistdist/doh.cc
```diff
@@ -222,9 +240,16 @@ void handleDOHTimeout(DOHUnit* oldDU)
 
   /* increase the ref counter before sending the pointer */
   oldDU->get();
-  if (send(oldDU->rsock, &oldDU, sizeof(oldDU), 0) != sizeof(oldDU)) {
+
+  ssize_t sent = write(oldDU->rsock, &oldDU, sizeof(oldDU));
+  if (sent != sizeof(oldDU)) {
+    if (errno == EAGAIN || errno == EWOULDBLOCK) {
+      ++g_stats.dohResponsePipeFull;
+    }
```
omoerbeek, line 248: maybe internal error on other errors?

pdns/dnsdistdist/doh.cc
```diff
@@ -432,7 +457,13 @@ static int processDOHQuery(DOHUnit* du)
       }
       /* increase the ref counter before sending the pointer */
       du->get();
-      if (send(du->rsock, &du, sizeof(du), 0) != sizeof(du)) {
+
+      ssize_t sent = write(du->rsock, &du, sizeof(du));
+      if (sent != sizeof(du)) {
+        if (errno == EAGAIN || errno == EWOULDBLOCK) {
+          ++g_stats.dohResponsePipeFull;
+        }
```
omoerbeek, line 465: here too...

pdns/dnsdist-lua.cc
rgacogne: This will not compile because the `uint32_t` type is not present in that variant. I could add it but I would prefer keeping the number of types limited because it quickly becomes confusing :)

pdns/dnsdist.cc
rgacogne: I don't really know what to do in that case besides releasing the memory.. We can't trigger a response from libh2o because we are not in the right thread for that, and we can't talk to that thread without writing to this pipe. So apart from releasing the memory.. Perhaps I could log an error in verbose mode?

rgacogne:
I added a few static assertions to make sure we don't write more than `PIPE_BUF` bytes at a time in our pipes.

PR #9215, opened 2020-06-09
auth: PKCS11 improvements

Habbie:
### Short description
* ECDSA support
* tests upgraded to softhsm2
* tests enabled in CircleCI

TODO:
* [x] code could use a bit of cleanup now (centralise the two Curve parameter strings; make sure we don't crash on unmapped algorithms; ...)
* [x] documentation is wildly outdated and contains a confusing deprecation message about 4.0.0
* [ ] ~add per algo tests?~ maybe later

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/pdnsutil.cc
+++ b/pdns/pdnsutil.cc
@@ -1682,7 +1682,7 @@ static bool showZone(DNSSECKeeper& dk, const DNSName& zone, bool exportDS = fals
         cout<<"ID = "<<value.second.id<<" ("<<DNSSECKeeper::keyTypeToString(value.second.keyType)<<")";
       }
       if (value.first.getKey()->getBits() < 1) {
-        cerr<<" <key missing or defunct>" <<endl;
+        cout<<" <key missing or defunct>" <<endl;
         continue;
       }
       if (!exportDS) {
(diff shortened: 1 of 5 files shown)


pdns/pdnsutil.cc
```diff
@@ -1682,7 +1682,7 @@ static bool showZone(DNSSECKeeper& dk, const DNSName& zone, bool exportDS = fals
         cout<<"ID = "<<value.second.id<<" ("<<DNSSECKeeper::keyTypeToString(value.second.keyType)<<")";
       }
       if (value.first.getKey()->getBits() < 1) {
-        cerr<<" <key missing or defunct>" <<endl;
+        cout<<" <key missing or defunct>" <<endl;
```
rgacogne, line 1685: I wonder if the move from `stderr` to `stdout` really makes sense?

Habbie: During regression tests, the output of pdnsutil is passed through `grep` and friends to get key IDs, and this message prints to the terminal repeatedly with no context. Changing to `cout` avoids that. In general I'm not sure it ever makes sense to mix stdout and stderr in the same output line.

Does that help?

rgacogne: Yes and no! I'm now wondering if there are cases where this message is actually helpful? :)

Habbie: Perhaps if we add `run pdnsutil hsm create-key?` to it, I'll do that :)

Habbie: pushed that

PR #9225, opened 2020-06-15
Wrap more FILE objects in smart pointers

rgacogne:
### Checklist

I have:
- [x] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/modules/pipebackend/coprocess.cc
+++ b/modules/pipebackend/coprocess.cc
@@ -232,12 +232,7 @@ UnixRemote::UnixRemote(const string& path, int timeout)
   if(connect(d_fd, (struct sockaddr*)&remote, sizeof(remote)) < 0)
     unixDie("Unable to connect to remote '"+path+"' using UNIX domain socket");
 
-  d_fp = fdopen(d_fd, "r");
-}
-
-UnixRemote::~UnixRemote()
-{
-  fclose(d_fp);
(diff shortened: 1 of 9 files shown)


modules/pipebackend/coprocess.cc
```diff
@@ -232,12 +232,7 @@ UnixRemote::UnixRemote(const string& path, int timeout)
   if(connect(d_fd, (struct sockaddr*)&remote, sizeof(remote)) < 0)
     unixDie("Unable to connect to remote '"+path+"' using UNIX domain socket");
 
-  d_fp = fdopen(d_fd, "r");
-}
-
-UnixRemote::~UnixRemote()
-{
-  fclose(d_fp);
+  d_fp = std::unique_ptr<FILE, int(*)(FILE*)>(fdopen(d_fd, "r"), fclose);
```
zeha, line 235: small suggestion: stick the type + "fclose" in some helper header?

rgacogne: We could declare an alias to save a few declarations:
```c++
using UniqueFilePointer = std::unique_ptr<FILE, int (*)(FILE*)>;
```

It won't save writing the `fclose` part, though, and I'm not sure I like hiding the real type. It means you need to find the declaration to know what to do with that object. We could save the `fclose` part by writing a real wrapper class, but then we need to implement several methods of `std::unique_ptr` for it to be useful, and it seems not worth the trouble.

PR #9226, opened 2020-06-15
rec: Fix three shared cache issues

omoerbeek:
- Only prime share cache once on startup

- Cache pruning could go into an infinite loop if not enough expired
entries could be pruned.

- Handler thread isn't run very often, but now the record cache
pruning is done by it, so increase frequency of the housekeeping
call for the handler thread.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [X] documented the code

### DIFF
--- a/pdns/pdns_recursor.cc
+++ b/pdns/pdns_recursor.cc
@@ -4662,11 +4662,13 @@ try
   t_allowFrom = g_initialAllowFrom;
   t_udpclientsocks = std::unique_ptr<UDPClientSocks>(new UDPClientSocks());
   t_tcpClientCounts = std::unique_ptr<tcpClientCounts_t>(new tcpClientCounts_t());
-  primeHints();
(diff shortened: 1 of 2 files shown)


pdns/pdns_recursor.cc
```diff
@@ -4784,7 +4786,9 @@ try
   while (!RecursorControlChannel::stop) {
     while(MT->schedule(&g_now)); // MTasker letting the mthreads do their thing
 
-    if(!(counter%500)) {
+    // Use primes, it avoid not being echeduled in casdes where the counter has a reguykar pattern.
```
rgacogne, line 4789:
```suggestion
    // Use primes, it avoid not being scheduled in cases where the counter has a regular pattern.
```

omoerbeek: Thanks, that's what I get by editing & committing quickly before a call...

PR #9229, opened 2020-06-15
dnsdist: Implement an ACL in the internal web server

rgacogne:
### Short description

Implement the same kind of ACL for the internal web server that we have in the auth and rec, and restricts to 127.0.0.1 and ::1 by default.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdistdist/dnsdist-web.hh
+++ b/pdns/dnsdistdist/dnsdist-web.hh
@@ -1,0 +1,22 @@
+#pragma once
+
+struct WebserverConfig
+{
+  WebserverConfig()
+  {
+    acl.toMasks("127.0.0.1, ::1");
+  }
+
+  NetmaskGroup acl;
+  std::string password;
+  std::string apiKey;
+  boost::optional<std::map<std::string, std::string> > customHeaders;
+  std::mutex lock;
+};
+
+void setWebserverAPIKey(const boost::optional<std::string> apiKey);
+void setWebserverPassword(const std::string& password);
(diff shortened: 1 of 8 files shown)


pdns/dnsdistdist/dnsdist-web.hh
```diff
@@ -0,0 +1,22 @@
+#pragma once
```
chbruyand, line 1: Maybe add the license header ?

pdns/dnsdistdist/docs/guides/webserver.rst
```diff
@@ -9,6 +9,13 @@ To visually interact with dnsdist, try add :func:`webserver` to the configuratio
 
 Now point your browser at http://127.0.0.1:8083 and log in with any username, and that password. Enjoy!
 
+Since 1.5.0, only connections from 127.0.0.1 and ::1 are allowed by default. To allow connections from 192.0.2.1, instead:
```
chbruyand, line 12: This is actually a list of netmasks. So this would probably be better to document here and after that it accepts various forms (eg. `192.0.2.1`, `192.0.0.0/16`, and even `!192.0.0.0/16`)

pdns/dnsdistdist/dnsdist-web.hh
rgacogne: Didn't we stop adding it? I'm a bit lost on that topic to be fair :-)

PR #9231, opened 2020-06-15
Use explicit flag for the specific version of c++ wer'e targeting.

omoerbeek:
### Short description

Been running this on my test OpenBSD/arm64 box for a while.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -19,7 +19,7 @@ AM_SILENT_RULES([yes])
 AC_CANONICAL_HOST
 # Add some default CFLAGS and CXXFLAGS, can be appended to using the environment variables
 CFLAGS="-g -O2 -Wall -Wextra -Wshadow -Wno-unused-parameter -Wmissing-declarations -Wredundant-decls $CFLAGS"
-CXXFLAGS="-g -O2 -Wall -Wextra -Wshadow -Wno-unused-parameter -Wmissing-declarations -Wredundant-decls $CXXFLAGS"
+CXXFLAGS="-std=c++11 -g -O2 -Wall -Wextra -Wshadow -Wno-unused-parameter -Wmissing-declarations -Wredundant-decls $CXXFLAGS"
 
 AC_PROG_CC
 AM_PROG_CC_C_O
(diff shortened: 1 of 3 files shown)


configure.ac
```diff
@@ -19,7 +19,7 @@ AM_SILENT_RULES([yes])
 AC_CANONICAL_HOST
 # Add some default CFLAGS and CXXFLAGS, can be appended to using the environment variables
 CFLAGS="-g -O2 -Wall -Wextra -Wshadow -Wno-unused-parameter -Wmissing-declarations -Wredundant-decls $CFLAGS"
-CXXFLAGS="-g -O2 -Wall -Wextra -Wshadow -Wno-unused-parameter -Wmissing-declarations -Wredundant-decls $CXXFLAGS"
+CXXFLAGS="-std=c++11 -g -O2 -Wall -Wextra -Wshadow -Wno-unused-parameter -Wmissing-declarations -Wredundant-decls $CXXFLAGS"
```
zeha, line 22: isn't this what `AX_CXX_COMPILE_STDCXX_11` is supposed to do/avoid? having both sounds bad to me, but i'm not an autoconf expert.

omoerbeek: Hmmm, lemme doublecheck the flags actually added.

pieterlexis: afaik, `AX_CXX_COMPILE_STDCXX_11` checks if the flag is needed or not to support C++11 features. However, if your compiler default is to use C++14, it will happily run the compiler without `-std=c++11`, allowing the (accidental) use of newer C++ features

omoerbeek: Using the `AX_CXX_COMPILE_STDCXX_11` we know the compiler supports c++11, but there are two issues:

- an extended mode might be used. This can be fixed by using `[std]` instead of `[ext]`.
- Depending on the compiler, it might support a newer version of c++. One of the goals of this commit is to detect (accidental) use of features > c++11. Using  `AX_CXX_COMPILE_STDCXX_11` does not do that.

zeha: Right. Maybe `AX_CXX_COMPILE_STDCXX_11` can then just go away. (I hope compilers abort if they do not understand `-std=c++11`.

PR #9237, opened 2020-06-16
rec: Don't validate a NXD with a NSEC proving that the name is an ENT

rgacogne:
### Short description

A NSEC from a.powerdns.com. to a.b.powerdns.com. proves that b.powerdns.com. exists and is an empty non terminal, so it cannot be used to prove that b.powerdns.com. does not exist.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


omoerbeek:
This makes sense to me though I'd like @Habbie to double check as well.

PR #9238, opened 2020-06-16
dnsdist: Clean up dnsdistconf.lua as a default configuration file

### LINKED ISSUE #8038: dnsdist + auth-ns: insecure default settings
- Program: pdns-auth + dnsdist
 - Issue type: request for comments

### Short description
When running an authoritative name server, thery are often configured to allow XFR for 127.0.0.1. I.e. PDNS-Auth has as default:
```
disable-axfr=no
allow-axfr-ips=127.0.0.0/8,::1

```

When putting dnsdist in front of the auth name server, running on the same host, the communication between dnsdist and the auth name server uses 127.0.0.1. As dnsdist also forwards XFR requests, suddenly the safe default xfr settings become insecure.

I would propose a default setting for dnsdist to refuse XFR or have a default entry like this in the default config file:

```
--
-- Refuse incoming AXFR,IXFR,NOTIFY,UPDATE from unknown sources.
-- Add trusted sources (slaves, masters) explicitely in front of this rule
--
addAction(OrRule({OpcodeRule(DNSOpcode.Notify), OpcodeRule(DNSOpcode.Update), QTypeRule(DNSQType.AXFR), QTypeRule(DNSQType.IXFR)}), RCodeAction(DNSRCode.REFUSED))

```

rgacogne:
### Short description

The idea is to be able to provide a default configuration file, including the basic blocks to build a basic configuration of dnsdist. We should then be able to install it as a default dnsdist.conf.

Hopefully closes #8038.

### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdistconf.lua
+++ b/pdns/dnsdistconf.lua
@@ -63,7 +63,7 @@
 -- addAction({"abuse.example.org.", "xxx."}, PoolAction("abuse"))
 
 -- drop queries for this exact qname
--- addAction(QNameRule("drop-me.powerdns.com."), DropAction())
+-- addAction(QNameRule("drop-me.example.org."), DropAction())
 
 -- send the queries from a selected subnet to the
 -- abuse pool
--- a/pdns/dnsdistconf.lua
+++ b/pdns/dnsdistconf.lua
@@ -60,7 +60,7 @@
 
 -- send the queries for selected domain suffixes to the servers
 -- in the 'abuse' pool
--- addAction({"powerdns.org.", "xxx."}, PoolAction("abuse"))
+-- addAction({"abuse.example.org.", "xxx."}, PoolAction("abuse"))
 
 -- drop queries for this exact qname
 -- addAction(QNameRule("drop-me.powerdns.com."), DropAction())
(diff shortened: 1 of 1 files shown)


pdns/dnsdistconf.lua
```diff
@@ -1,208 +1,113 @@
...
+-- setMaxTCPClientThreads(20)
+
+-- == Sample Actions ==
+
+-- https://dnsdist.org/rules-actions.html
+
+-- send the queries for selected domain suffixes to the servers
 -- in the 'abuse' pool
-addAction({"ezdns.it.", "xxx."}, PoolAction("abuse"))
+-- addAction({"powerdns.org.", "xxx."}, PoolAction("abuse"))
```
zeha, line 63:
```suggestion
-- addAction({"abuse.example.org.", "xxx."}, PoolAction("abuse"))
```

pdns/dnsdistconf.lua
```diff
@@ -1,208 +1,113 @@
...
+
+-- drop queries for this exact qname
+-- addAction(QNameRule("drop-me.powerdns.com."), DropAction())
```
zeha, line 66:
```suggestion
-- addAction(QNameRule("drop-me.example.org."), DropAction())
```

Issue #5375, opened 2017-06-04
Postgres backend doesn't handle large ids in the domain table properly


nickjacques:
- Program: Authoritative
 - Issue type: Bug report

### Short description
When using the `gpgsql` backend, PowerDNS exhibits issues when `domain_id` values exceed the `int` boundaries of Postgres. While this seems trivial on its face - the backend expects the column type specified in the schema - consider a scenario where this may not be the case. 

For instance, CockroachDB does not use incrementing SERIALs, [but rather, generates a large integer that is likely to be unique](https://www.cockroachlabs.com/docs/serial.html). After initial testing, it appears that PowerDNS and CockroachDB look to be compatible (besides some small quibbles, like the [INET type in the schema](https://github.com/PowerDNS/pdns/blob/master/modules/gpgsqlbackend/schema.pgsql.sql#L40), which is [actively being worked on](https://github.com/cockroachdb/cockroach/issues/6981).

I would suggest making use of `bigint` or `bigserial` types, which should allow for compatibility with CockroachDB, as well as providing additional headroom for (*very*) large deployments of PowerDNS.

### Environment

 - Operating system: CentOS Linux release 7.3.1611 (Core)
 - Software version: PowerDNS Authoritative Server 4.0.3 Built using gcc 4.8.5 20150623 (Red Hat 4.8.5-4) on Jan 17 2017 08:55:46 by buildbot@c17f0608c0cd. Postgres 9.6.
 - Software source: PowerDNS CentOS Repo

### Steps to reproduce

1. Start with a functioning base install of PowerDNS Authoritative 4.0.3 connected to a fresh/empty Postgres 9.6 backend.
2. Create a zone using `pdnsutil create-zone example.com` and observe that a record is created in the `domains` table with `id` of 1.
3. Execute `pdnsutil list-all-zones` and observe that `example.com` is displayed.
4. Update the `id` column to the type `bigint`.
5. Repeat step 3 and observe proper functionality.
6. Insert a record in the `domains` table with id `12345678901` (larger than the `int` type).
7. Execute `pdnsutil list-all-zones` (as in step 3), but observe the following error: `Error: stou`

### Expected behaviour
The zone should be listed and should successfully resolve when queried.

### Actual behaviour
The following is returned: `Error: stou` and queries are not successful.

### Other information
I am happy to provide additional debugging or log information if needed. Thank you!

wrouesnel:
[...]

PR #9240, opened 2020-06-16
Add instructions for running PowerDNS on CockroachDB using generic pgsql

LordGaav:
### Short description
Adds documentation on how to run PowerDNS authoritative on CockroachDB. Based on testing in #5375 . Includes pointers on what schema and configuration changes to make, and a reference to the original content on Github for further details.

### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/docs/backends/generic-postgresql.rst
+++ b/docs/backends/generic-postgresql.rst
@@ -152,7 +152,11 @@ Given the normal pgsql schema, change the following:
   );
 
 
-3. Do not add the recordorder INDEX to the records table, the text_pattern_ops operator class is not supported.
+3. Do **not** add the following index to the records table, the text_pattern_ops operator class is not supported:
+
+.. code-block:: SQL
+
+  CREATE INDEX recordorder ON records (domain_id, ordername text_pattern_ops);
 
 
 Configuration changes
--- a/docs/backends/generic-postgresql.rst
+++ b/docs/backends/generic-postgresql.rst
@@ -141,14 +141,14 @@ Given the normal pgsql schema, change the following:
 
   -- Before
   CREATE TABLE domains (
-    id                    SERIAL PRIMARY KEY,
-    ...
+    id SERIAL PRIMARY KEY,
+    --
   }
 
   -- After
   CREATE TABLE domains (
-    id                    INT PRIMARY KEY DEFAULT nextval('domain_id') PRIMARY KEY,
-    ...
+    id INT DEFAULT nextval('domain_id') PRIMARY KEY,
+    --
   );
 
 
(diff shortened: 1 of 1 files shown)


docs/backends/generic-postgresql.rst
```diff
@@ -109,3 +109,66 @@ This is the 4.3 schema. Please find `the 4.2 schema <https://github.com/PowerDNS
...
+
+  -- Before
+  CREATE TABLE domains (
+    id                    SERIAL PRIMARY KEY,
+    ...
+  }
+
+  -- After
+  CREATE TABLE domains (
+    id                    INT PRIMARY KEY DEFAULT nextval('domain_id') PRIMARY KEY,
```
RobinGeuze, line 150: PRIMARY KEY twice?

docs/backends/generic-postgresql.rst
```diff
@@ -109,3 +109,66 @@ This is the 4.3 schema. Please find `the 4.2 schema <https://github.com/PowerDNS
...
+    ...
+  );
+
+
+3. Do not add the recordorder INDEX to the records table, the text_pattern_ops operator class is not supported.
```
RobinGeuze, line 155: Maybe show the specific index you mean?

I also wonder if a simple script that modifies the schema is in order to make things convenient.

docs/backends/generic-postgresql.rst
LordGaav: Now that you mention it, it does execute successfully though.

docs/backends/generic-postgresql.rst
LordGaav: I'd rather include the entire SQL dump in that case, which is way more descriptive.

docs/backends/generic-postgresql.rst
RobinGeuze: I would call that weird :P. But eh, maybe test it with PRIMARY KEY once to make sure it works properly.

docs/backends/generic-postgresql.rst
RobinGeuze: The downside of that is that it will need to be maintained alongside the normal postgresql schema which might be sub-optimal. Maybe @pieterlexis or @Habbie have an opinion on that?

Habbie: https://github.com/PowerDNS/pdns/issues/5375#issuecomment-644822573

docs/backends/generic-postgresql.rst
LordGaav: Tested with the single statement, and it does work.

docs/backends/generic-postgresql.rst
LordGaav: I clarified which index to skip in the text.

PR #9265, opened 2020-06-20
auth: add used master address to slave check logs

zeha:
### Short description

Adds the used master address to all slave check "result" logs.
Should help with debugging issues when multiple masters are configured, and also clarify once and for all that the checked master is being used for the following AXFR.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


zeha:
Wanted to make the DomainInfo const in the sdomains loop, but unfortunately di.backend can be reset inside. Not sure if this is actually relevant anymore today.

Issue #9257, opened 2020-06-18
down 1 distributor thread when use distributor-threads mode v4.2.1


liwei0526vip:
- Program: Recursor
 - Issue type: Bug report
 - Version: 4.2.1

### Short description

We are trying to use the distributor-thread method of pdns-recursor, configured with 3 distributor-threads, N worker threads, and ran for about 3 days, and found that distributor-thread died one of the threads, and some of the logs are very simple The message is below

```bash
any other exception in main:
```

### Environment
 - Operating system: CentOS7.3 kernel 3.10
 - Software version: 4.2.1

### Steps to reproduce

1. There is no stable way to reproduce, but there have been two problems so far

omoerbeek:
Please share the config file.

liwei0526vip:
#### my configuration information

```bash
allow-from=0.0.0.0/0,::/0
daemon=yes
disable-packetcache=yes
disable-syslog=no
dnssec=off
dont-query=127.0.0.0/8, 10.0.0.0/8, 100.64.0.0/10, 169.254.0.0/16, 192.168.0.0/16, 172.16.0.0/12, ::1/128, fc00::/7, fe80::/10, 0.0.0.0/8, 192.0.0.0/24, 192.0.2.0/24, 198.51.100.0/24, 203.0.113.0/24, 240.0.0.0/4, ::/96, ::ffff:0:0/96, 100::/64, 2001:db8::/32
edns-subnet-whitelist=xxxx.com.hk
log-timestamp=yes
loglevel=7
packetcache-servfail-ttl=1
pdns-distributes-queries=yes
processes=1
query-local-address6=::
quiet=off
reuseport=yes
statistics-interval=0
trace=off
use-incoming-edns-subnet=yes
webserver=no
non-local-bind=yes
api-key=changeme
threads=19
local-address=10.10.10.10,10.10.10.11,172.16.42.31,172.16.42.32,10.41.0.45,10.41.0.46,172.16.234.38,172.16.234.41,172.16.153.24,172.16.153.87
socket-dir=/usr/local/pdns-recursor.debug/sbin/
cpu-map=0=0 1=1 2=2 3=3 4=4 5=5 6=6 7=7 8=8 9=9 10=10 11=11 12=12 13=13 14=14 15=15 16=16 17=17 18=18 19=19
forward-zones-recurse=xxx.com.hk=x.x.x.x;x.x.x.x
logging-facility=0
distributor-threads=3
[...]

PR #9268, opened 2020-06-22
rec: Better exception handling in houseKeeping / handlePolicyHit

rgacogne:
### Short description

- An uncaught exception being raised in `houseKeeping` could prevent it from being run again because `s_running` would stay true. That could for example happen when an `ImmediateServFailException` was raised during a call to `primeRootNSZones()` ;
- An `ImmediateServFailException` being raised during the call to `SyncRes::beginResolve()` from `handleRPZCustom()` could lead to the resolution process to stop without any response being sent:
```
Any other exception in a resolver context (www.example.net/A from 127.0.0.1:4242)
```
This commit turns into a ServFail answer instead.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


rgacogne:
I suspect this would fix https://github.com/PowerDNS/pdns/issues/9257 once backported to 4.2.x, since the exception is passed up to the function scheduling the `mthread` on boost versions older than 1.61 (like in CentOS 7), while it seems that it doesn't in 1.61+, likely because of the changes made in `boost::context` 1.61.

PR #9270, opened 2020-06-22
Remove archaic spellings

jsoref:
### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code


.github/actions/spell-check/reject.txt
```diff
@@ -0,0 +1,18 @@
+benefitting
+spae
+spaebook
+spaecraft
```
Habbie, line 4: what

jsoref: I'm still working through the right way to handle this.

Roughly, the base dictionary currently in use has all these words.

[spae](https://en.wiktionary.org/wiki/spae) is a word that means (meant): To divine; foretell
[spae-craft](https://en.wiktionary.org/wiki/spae-craft) is a word that means (meant): The magical art of foretelling the future

Mostly, the problem with this family of words is that they're easy enough typos for `space` / `spaces` / `spaced` / `spacecraft` / `spaceman`.

One approach would be to commit a custom dictionary and delete these words from it (that's what microsoft/terminal did when they found these words -- because someone made one of these typos).

The cost there is holding the entire dictionary in the repo instead of the exceptions.

In the vaguely near future, I'm going to release a version that makes it easier for clients to specify a version of a dictionary and I'll ship a version where a bunch of these archaic words are dropped and other newer words are folded in.

pieterlexis:

PR #9275, opened 2020-06-26
dnsdist: Force a reconnection when a downstream transitions to the UP state

### LINKED ISSUE #5220: dnsdist should reopen sockets after interface or routing change
- Program: dnsdist
 - Issue type: Bug report

### Short description
dnsdist opens sockets for doing queries, but if interfaces or routing changes, the sockets are not reopened, meaning that dnsdist queries go out using wrong interface and possibly wrong source IP. This means they get unanswered, but probe queries are still answered as they reopen new sockets. So we end up with non-responsive UP servers.

### Environment
 - Operating system: Ubuntu
 - Software version: 16.04.02 LTS
 - Software source: PowerDNS

### Steps to reproduce
1. Configure DNSDist with two servers, so that the second server requires a, say, VPN connection.
2. Start dnsdist
3. Check that the second server is down
4. Enable the interface (or connect to VPN)
5. Observe the second server is up

### Expected behaviour
Observe the second server won't reply to real queries.

### Actual behaviour
The real queries would be answered.

Nuitari:
Force a reconnection when a downstream transitions to the UP state

### Short description
Force the reconnection of a downstream when a server comes back up. 

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdistdist/dnsdist-healthchecks.cc
+++ b/pdns/dnsdistdist/dnsdist-healthchecks.cc
@@ -60,7 +60,7 @@ void updateHealthCheckResult(const std::shared_ptr<DownstreamState>& dss, bool n
   if(newState != dss->upStatus) {
     warnlog("Marking downstream %s as '%s'", dss->getNameWithAddr(), newState ? "up" : "down");
 
-    if (newState) {
+    if (newState && dss->reconnectOnUp) {
       newState = dss->reconnect();
 
       if (dss->connected && !dss->threadStarted.test_and_set()) {
--- a/pdns/dnsdistdist/dnsdist-healthchecks.cc
+++ b/pdns/dnsdistdist/dnsdist-healthchecks.cc
@@ -60,7 +60,7 @@ void updateHealthCheckResult(const std::shared_ptr<DownstreamState>& dss, bool n
   if(newState != dss->upStatus) {
     warnlog("Marking downstream %s as '%s'", dss->getNameWithAddr(), newState ? "up" : "down");
 
-    if (newState && !dss->connected) {
+    if (newState) {
       newState = dss->reconnect();
 
       if (dss->connected && !dss->threadStarted.test_and_set()) {
--- a/pdns/dnsdistdist/dnsdist-healthchecks.cc
+++ b/pdns/dnsdistdist/dnsdist-healthchecks.cc
@@ -60,7 +60,7 @@ void updateHealthCheckResult(const std::shared_ptr<DownstreamState>& dss, bool n
   if(newState != dss->upStatus) {
     warnlog("Marking downstream %s as '%s'", dss->getNameWithAddr(), newState ? "up" : "down");
 
-    if (newState && dss->reconnectOnUp) {
+    if (newState && (!dss->connected || dss->reconnectOnUp)) {
       newState = dss->reconnect();
 
       if (dss->connected && !dss->threadStarted.test_and_set()) {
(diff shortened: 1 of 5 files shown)


Nuitari:
> Thank you for submitting this pull request! I'm a bit cautious about reconnecting the socket(s) every time we transition from down to up, however, because it might have a negative impact (responder thread failing to send at least one response, queued TCP connections being lost). If I read the IRC backlog correctly, it looks like in your case the backend was never considered up? It would probably make sense to reconnect in that case indeed.

In that case the backend was never up, and would come up when a vpn connected. tcpdump would show the packet coming out of the correct interface but with the wrong source IP. 
We also saw similar issues where the interface for a server changed (think VM migrations).

I have no way to test those cases with TCP, but I would assume that a down backend would not have a response waiting to be sent upstream. I don't know if dnsdist would redispatch things to another server in a pool if a healthcheck fails. 

Does dnsdist open multiple tcp connections to a backend outside of healthchecks?


> Out of curiosity, do you know what `sendmsg()` returned (in `udpClientSendRequestToBackend()`) before reconnecting the socket? We try to detect such a network configuration change there already, and perhaps we could enhance our logic.

I did not check, however I can infer that it was not -1. 
The reason is that the error message in the logs matched the one at line 1698 in dnsdist.cc which is at the info log level. 
If the return was -1, there should be an info log entry with "Error sending request to backend", which we never saw.

rgacogne:
> In that case the backend was never up, and would come up when a vpn connected. tcpdump would show the packet coming out of the correct interface but with the wrong source IP.

OK, I completely agree it would make sense to reconnect when the backend was never up.
Out of curiosity, did you try specifying the source IP via the `source` parameter? It might not work if the IP doesn't exist yet in the system, though.

> We also saw similar issues where the interface for a server changed (think VM migrations).

Right, I can see how this is an issue.

> I have no way to test those cases with TCP, but I would assume that a down backend would not have a response waiting to be sent upstream. I don't know if dnsdist would redispatch things to another server in a pool if a healthcheck fails.
> Does dnsdist open multiple tcp connections to a backend outside of healthchecks?

I did not think this through last time but we don't prepare sockets for TCP in that case, of course, and already established connections will not be impacted by that change. Only UDP sockets are set up in advance for performance reasons.

So I think there are two different cases:
- if the backend was never up before, it should not hurt to reconnect the sockets right away when it does get up ;
- in the general case, I don't think we should reconnect the sockets on a backend going down and up, since we have seen setups with flaky backends going up and down all the time, and recreating the sockets is not cheap (we need to remove them from the IO multiplexer, close them, create new ones and register them to the IO multiplexer, then wait for the receiving thread to realize the sockets have been closed and switch to the new ones). On the other hand I understand this might be useful for some setups where interfaces come and go (VPN, VM migrations, ...) so in my opinion it would make sense to make that automatic re-connection optional, by adding an option (`reconnectOnUp`?) to `newServer` and a boolean to `DownstreamState`.

Nuitari:
> Out of curiosity, did you try specifying the source IP via the `source` parameter? It might not work if the IP doesn't exist yet in the system, though.

The IP is not always known either in some of these cases.

>     * in the general case, I don't think we should reconnect the sockets on a backend going down and up, since we have seen setups with flaky backends going up and down all the time, and recreating the sockets is not cheap (we need to remove them from the IO multiplexer, close them, create new ones and register them to the IO multiplexer, then wait for the receiving thread to realize the sockets have been closed and switch to the new ones). On the other hand I understand this might be useful for some setups where interfaces come and go (VPN, VM migrations, ...) so in my opinion it would make sense to make that automatic re-connection optional, by adding an option (`reconnectOnUp`?) to `newServer` and a boolean to `DownstreamState`.

I've added the option to this PR.

pdns/dnsdistdist/dnsdist-healthchecks.cc
```diff
@@ -60,7 +60,7 @@ void updateHealthCheckResult(const std::shared_ptr<DownstreamState>& dss, bool n
   if(newState != dss->upStatus) {
     warnlog("Marking downstream %s as '%s'", dss->getNameWithAddr(), newState ? "up" : "down");
 
-    if (newState && !dss->connected) {
+    if (newState && dss->reconnectOnUp) {
```
rgacogne, line 63:
```suggestion
    if (newState && (!dss->connected || dss->reconnectOnUp)) {
```

PR #9278, opened 2020-06-29
dnsdist: Prevent race between the DoH handling threads

rgacogne:
### Short description

This PR fixes a race between the different threads involved in DoH handling, leading to use-after-free and a crash in some cases. This is done by making sure that the h2o req object is only accessed from the main DoH thread, where it can't be altered by h2o in the meantime, and by preventing any access to it once the request has been canceled by h2o. It also makes sure that the pointer to our internal DOHUnit object is cleared from the h2o request pool before releasing that DOHUnit object.
This change involves a penalty cost because we now need to copy the HTTP headers, if any, because we need to access them from a different thread than the main DoH one and the internal memory handled by h2o could be gone then, or even be altered under our feet.

This issue was discovered by Tomas Krizek from CZ.NIC while using the DNS Shotgun tool, and reported to us. Many thanks for that!

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/pdns/dnsdistdist/doh.cc
+++ b/pdns/dnsdistdist/doh.cc
@@ -652,9 +652,9 @@ static h2o_pathconf_t *register_handler(h2o_hostconf_t *hostconf, const char *pa
    We use this to signal to the 'du' that this req is no longer alive */
 static void on_generator_dispose(void *_self)
 {
-  DOHUnit** du = (DOHUnit**)_self;
-  if(*du) { // if 0, on_dnsdist cleaned up du already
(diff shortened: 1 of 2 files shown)


tomaskrizek:
I've tested this version and dnsdist now seems to be stable even under heavy DoH load.

Thanks for the prompt fix!

rgacogne:
Thanks a lot for the initial report, your wonderful help to reproduce the issue and testing the fix, that's very much appreciated!

Issue #9220, opened 2020-06-09
Configure DNSKEY TTL


klaus3000:
- Program: Authoritative
 - Issue type: Feature request

### Short description
Currently, PDNS sets the DNSKEY TTL of synthesized reponses to the negative-TTL of the zone. Short negative TTLs are not uncommon, hence this may result in short DNSKEY TTL.

I just learned in the DNS-OARC online meeting that "Recursive resolvers trim dependent signed RRsets to no more than their parent". Which means resolvers stop caching when the zones DNSKEY TTL expires.

Is there a specific reason why the DNSKEY TTL uses the zone's negative-TLL?

I think it may be useful to have a pdns.conf option to override the default value, ie: dnskey-ttl=86400.

mnordhoff:
Note:

> If you want a ‘normal’ TTL (3600, 86400, etc.) for your DNSKEY but a low TTL on negative answers, set your SOA minimum TTL to the high number, and set the TTL on the SOA record itself to the low TTL you want for negative answers. Note that [before 4.3.0] the NSEC/NSEC3 records proving those negatives will get the high TTL in that case, and this may affect subsequent resolution in resolvers that do aggressive NSEC caching (RFC 8198).

https://doc.powerdns.com/authoritative/dnssec/operational.html#some-notes-on-ttl-usage

klaus3000:
I was not aware that a low SOA TTL has impact on the negative caching of resolvers. Anyways, we are a secondary DNS service and we sign our customer zones. Hence, when we slave a zone we do not have control over the SOA.

Habbie:
> I just learned in the DNS-OARC online meeting that "Recursive resolvers trim dependent signed RRsets to no more than their parent".

I also saw that claim. I asked him what resolvers he had actually seen doing that, but sadly he did not respond.

PowerDNS Recursor does not do this; I don't know if any resolvers do.

Habbie:
Given the notes that @mnordhoff linked to, I'm leaning towards closing this. Any objections?

aldem:
>set your SOA minimum TTL to the high number

As far as I know, "SOA minimim" is deprecated in favor of "SOA negative" in [RFC 2308](https://tools.ietf.org/html/rfc2308) ("4 - SOA Mimimum Field"):
>  Despite being the original defined meaning, the first of these, the  minimum TTL value of all RRs in a zone, has never in practice been used and is hereby deprecated.

and a bit later:
[...]

PR #9281, opened 2020-06-29
auth: Add dnskey-minimum-ttl-override

zeha:
### Short description

Secondary operators regularly do not have control over the SOA minimum field, which is used as the DNSKEY TTL in replies. This allows setting a minimum value for these replies.

Implements #9220.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/docs/dnssec/operational.rst
+++ b/docs/dnssec/operational.rst
@@ -248,3 +248,8 @@ Note that the NSEC/NSEC3 records proving those negatives will get the high TTL i
 
   NSEC(3) records now get the negative TTL (which is the lowest of the SOA TTL and the SOA minimum), which means their TTL matches that of an error such as NXDOMAIN.
   The warning about RFC8198 no longer applies.
+
+.. note::
+
+  Since 4.4.0, the DNSKEY TTL can be overridden using :ref:`setting-dnskey-minimum-ttl-override` if the SOA minimum TTL is too low.
+  This is a per server setting, mostly useful for secondary operators.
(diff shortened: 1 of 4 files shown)


mind04:
I do not like the idea of enforcing a certain minimum value. Overriding the soa minimum ttl with a fixed value (when set) does make a lot more sense. As an operator you want consistent dnskey ttls. This will make the timing of dnssec related updates more easy.

Adding a 'dnskey-ttl' and optionally a 'nsec3param-ttl' which override the soa minimum ttl is in my opinion the way to go.

Ps. do not  forget the ttls in an afxr

zeha:
I don't think dnskey-ttl/nsec3param-ttl would be very useful for a secondary operator.
In an earlier discussion with @Habbie this came up, and the outcome for those was more or less "could make sense as a per-zone metadata item" - but per pdns instance, they would probably hurt more than be useful.

zeha:
(The "secondary" angle is also why I wasn't going to touch AXFR-out.)

klaus-nicat:
Hello! I would like to start the discussion again. After some more thinking and some DNSSEC quirks on our system I think Kees approach is useful. As a secondary provider, we sign customer zones but do not have control over the SOA. So, some customer zones DNSKEY may have a small TTL, others may have a huge TTL. Both are not good. For example with hugh TTLs it takes very long to cleanup in resolver caches after DNSKEY errors.

So I would propose a global dnskey-ttl setting. I think 'per zone metadata' is overkill, as I want all my zones to behave similar. Maybe related TTLs (CDS, ...) may be set to the same value.

Habbie:
Hello Klaus,

you commented on a closed ticket (PR, in this case) which means we will immediately forget about what you said. It's best to open a new issue to discuss this.

klaus-nicat:
I hoped the ticket will open up again ;-)

PR #9290, opened 2020-07-01
rec: Refuse QType 0 right away, based on rfc6895 section 3.1

rgacogne:
### Short description

There is no good reason to query that type, and we used to accept the query, even sending a query with TYPE0 to authoritative servers.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


zeha:
is this also true for auth, dnsdist?

rgacogne:
> is this also true for auth, dnsdist?

dnsdist doesn't care and will pass the query along, which seems OK to me. I did not look at the auth but it would make sense to me to reject it there as well.

rgacogne:
I'll prepare a separate PR for the auth.

PR #9294, opened 2020-07-05
Update dynlistener.cc

ihsinme:
according to definition   
https://en.cppreference.com/w/c/io/fgets  

```
Reads at most count - 1 characters from the given file stream and stores them in the character array pointed to by str. Parsing stops if a newline character is found, in which case str will contain that newline character, or if end-of-file occurs. If bytes are read and no errors occur, writes a null character at the position immediately after the last character written to str.
```

and according to definition    
https://en.cppreference.com/w/cpp/container/vector/resize

```
1) additional default-inserted elements are appended
```

I propose to exclude dead code as required by "CWE-561: Dead Code"

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dynlistener.cc
+++ b/pdns/dynlistener.cc
@@ -259,11 +259,7 @@ string DynListener::getLine()
         continue;
       }
       
-      if(strlen(&mesg[0]) == mesg.size()) {
-        g_log<<Logger::Error<<"Line on controlsocket ("<<d_client<<") was too long"<<endl;
-        close(d_client);
-        continue;
-      }
+
       break;
     }
   }


pdns/dynlistener.cc
```diff
@@ -259,11 +259,7 @@ string DynListener::getLine()
         continue;
       }
       
-      if(strlen(&mesg[0]) == mesg.size()) {
```
rgacogne, line 262: I agree this check does not seem to be needed. I wonder if we should check that the string is terminated by a newline instead, since otherwise it might indicate a truncated line?

ihsinme: definitely fgets will leave / n at the end of the line

ihsinme: maybe I do not understand the question correctly. can you specify it?

zeha: I imagine the check should have compared to mesg.size() - 1, to preserve the original spirit. Checking for `\n` might be a more approachable variant :)

ihsinme: I can be wrong, but I don’t understand what’s wrong with reading a lot of bytes, quite a lot even up to size-1. I think this is normal, the main thing is to keep the termination zero, but it is saved here.
I want to apologize for not making another proposal at the same time - this line also needs to be deleted; it is not needed:
https://github.com/PowerDNS/pdns/blob/2014b205f4db0a13b0b73203eca1b47c1c5221b6/pdns/dynlistener.cc#L282

rgacogne: Unless I'm mistaken, the issue is that if the original line is too long we will get a truncated version of up to `mesg.size() - 1` bytes, that we will then try to parse as a full line. I think we should check that `mesg` does end up on a `\n` character, to be sure that we have the full line, and discard it instead.

ihsinme: but in this case (if we discard part of the line), repeated reading will put the end of the line in the buffer, and we will assume that this is the entire line.
therefore, if we want to process the entire package correctly, we must either dynamically increase the size of the buffer or limit the maximum size.

rgacogne: Agreed, we should not discard the line but close the connection, which is what we did before that path.

ihsinme:
I wanted to clarify if work is underway on this PR?

Habbie:
As @rgacogne mentioned, the disconnect on truncated line is gone, which is not good.

PR #9297, opened 2020-07-06
rec: Ignore cache-only for DNSKEYs retrieval

rgacogne:
### Short description

When the DNSKEYs are needed for validation, the initial RD flag should not prevent us from going to the network. This was fine in most cases because the DNSKEYs are usually retrieved during the computation of the zone cuts, where we already ignore the RD flag, but we might also need to retrieve the DNSKEYs during just-in-time validation later on.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/syncres.cc
+++ b/pdns/syncres.cc
@@ -2330,7 +2330,9 @@ vState SyncRes::getDNSKeys(const DNSName& signer, skeyset_t& keys, unsigned int 
   LOG(d_prefix<<"Retrieving DNSKeys for "<<signer<<endl);
 
   vState state = Indeterminate;
+  const bool oldCacheOnly = setCacheOnly(false);
   int rcode = doResolve(signer, QType(QType::DNSKEY), records, depth + 1, beenthere, state);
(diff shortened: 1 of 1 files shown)


omoerbeek:
I am wondering if this issue also is relevant to`getDSRecords`...

rgacogne:
That's likely, yes. In theory we should have seen all the required DSs during the zone cuts computation, since all other queries should be "infra" queries to get NS and their corresponding A/AAAAs, which we are not validating, but that might not be 100% true for just-in-time validation..

rgacogne:
I pushed an additional commit to also ignore the cache-only flag when retrieving DSs. I believe the current code was fine without that change given the callers of `getDSRecords()`, but better safe than sorry.

Note that this PR is mostly a work-around, I think a better fix would be to skip validation for cache-only queries:
- no zone cut computation ;
- retrieving the validation status for the cache, but not performing any validation if the retrieved state is "Indeterminate" ;
- and of course not updating the cache no matter what.
But let's do that for master only in a different pull request.

PR #9300, opened 2020-07-06
auth: Refuse queries for QType values of zero or OPT

rgacogne:
### Short description

Follow-up of https://github.com/PowerDNS/pdns/pull/9290 for the authoritative server.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/packethandler.cc
+++ b/pdns/packethandler.cc
@@ -1261,6 +1261,11 @@ std::unique_ptr<DNSPacket> PacketHandler::doQuestion(DNSPacket& p)
     if(p.qclass==QClass::ANY)
       r->setA(false);
 
+    /* rfc6895 section 3.1 */
+    if (p.qtype.getCode() == 0 || p.qtype.getCode() == QType::OPT) {
+      r->setRcode(RCode::Refused);
+      return r;
+    }
 
   retargeted:;
     if(retargetcount > 10) {    // XXX FIXME, retargetcount++?


mnordhoff:
FWIW, I'd vote against this, unless there's a specific need for it. NODATA and NXDOMAIN are cacheable. If resolvers ask silly questions, I'd rather give them something that they can cache, so they'll go away instead of retrying.

rgacogne:
That's a very good point, I'll wait for more feedback but I believe you might be right!

mnordhoff:
Also FWIW, Knot DNS returns NOTIMP for type 0 and treats OPT queries normally.

zeha:
> Also FWIW, Knot DNS returns NOTIMP for type 0 and treats OPT queries normally.

What does "normally" mean here?

zeha:
It might be worthwhile changing the qclass not in (IN, ANY) branch above to whatever will be done for this case, too.

zeha:
FORMERR appears to be a popular choice for OPT queries among other Auth servers.

Habbie:
> FWIW, I'd vote against this, unless there's a specific need for it. NODATA and NXDOMAIN are cacheable. If resolvers ask silly questions, I'd rather give them something that they can cache, so they'll go away instead of retrying.

This makes sense to me.

rgacogne:
Closing this PR, then! Thanks for the feedback :-)

PR #9308, opened 2020-07-08
rec: Check that DNSKEYs have the zone flag set

rgacogne:
### Short description

As required by rfc4034 Section 5.2. That should not make any difference in practice.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/recursordist/test-syncres_cc4.cc
+++ b/pdns/recursordist/test-syncres_cc4.cc
@@ -684,6 +684,159 @@ BOOST_AUTO_TEST_CASE(test_dnssec_bogus_no_dnskey)
   BOOST_CHECK_EQUAL(queriesCount, 2U);
 }
 
+BOOST_AUTO_TEST_CASE(test_dnssec_bogus_dnskey_without_zone_flag)
+{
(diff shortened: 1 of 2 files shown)


Habbie:
Perhaps also https://tools.ietf.org/html/rfc4034#section-2.1.2 then?

> The Protocol Field MUST have value 3, and the DNSKEY RR MUST be
> treated as invalid during signature verification if it is found to be
> some value other than 3.

(also looks like 2.1.1 might be a better reference for the flags than 5.2)

rgacogne:
> Perhaps also https://tools.ietf.org/html/rfc4034#section-2.1.2 then?

This one is already checked in `getByTag()`. In fact, perhaps I should move the "zone" bit check there as well.

> (also looks like 2.1.1 might be a better reference for the flags than 5.2)

Yes!

Habbie:
Oh! What about the REVOKED bit?

rgacogne:
Moved the "zone" check to `getByTag()` with the better reference, added a "not revoked" check and some unit tests.

rgacogne:
Rebased to fix a conflict.

PR #9313, opened 2020-07-09
rec: Add the ability to log validation failures for selected names

rgacogne:
### Short description

This is a very rough draft, the general idea is to add the ability to get a one-line summary in the logs when a DNSSEC failure happens for selected domains, or suffixes. This is not as useful as getting a full trace, but can stay enabled pretty much forever since the summary line is only logged when the validation is done, and not when the entry is retrieved from the cache.
Comments welcome! 

### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/pdns_recursor.cc
+++ b/pdns/pdns_recursor.cc
@@ -4181,6 +4181,22 @@ static int serviceMain(int argc, char*argv[])
     exit(1);
   }
 
+  if (!::arg()["dnssec-validation-failures-log-filter"].empty()) {
+    std::vector<string> names;
+    stringtok(names, ::arg()["dnssec-validation-failures-log-filter"], ", ");
+    for (auto& name : names) {
+      if (name.empty()) {
+        continue;
+      }
+      bool exact = false;
+      if (name.at(name.size()-1) == '$') {
+        name = name.substr(0, name.size()-1);
+        exact = true;
+      }
+      SyncRes::addToValidationFilter(DNSName(name), exact);
+    }
+  }
+
   g_signatureInceptionSkew = ::arg().asNum("signature-inception-skew");
   if (g_signatureInceptionSkew < 0) {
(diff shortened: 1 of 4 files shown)


johnhtodd:
I have an idea for a complimentary feature to specific inclusion as this PR is written. It would be useful if there was an ability to exclude certain domains from being logged as DNSSEC Bogus. We wish to have bogus domains logged, but a large percentage are "bogus bogus" domains, if that is a term I can invent. "rootcanary.net", and "dotnxdomain.net", "gexperiments3.com", "cloudflareresolve.com" all are domains that are purposely replying with invalid DNSSEC results, but yet produce (in aggregate) probably around half our Bogus log messages. This means that we see a continuing stream of these messages that we have to filter out further downstream during post-processing. This is not a significant problem in itself, but it also means that the DNSSEC failure counter metrics that PDNS stores are also incremented for these domains, which is more problematic. We cannot easily & reliably create a baseline of actual failed DNSSEC results with the noise floor of these intentional failures creating a continual increase in the counters.  While we may never know the domain suffixes for _all_ intentionally bogus DNSSEC lookups, it is possible to filter a large volume of them out based on some commonly-visible domain suffixes, which would go a long way towards making the statistics more meaningful without including intentionally broken results.  Maybe this should be a feature request instead of a note on this PR.

rgacogne:
Right, that makes sense to me to have a list of "known bad" domains, as far as DNSSEC validation is concerned, for which we would neither log nor increase the counters.

johnhtodd:
If there is any objection to not incrementing the failure counters on such excluded domains, perhaps just adding a new counter of "excluded bogus domains" or something like that would allow for quick math to be done to reveal the true volume of interesting DNSSEC failures ( {all failures} - {excluded bogus domains failures} = actionable data)

pieterlexis:
@rgacogne ping!

rgacogne:
Closing since https://github.com/PowerDNS/pdns/pull/9597 has made debugging such issues much, much easier.

PR #9318, opened 2020-07-09
gsqlbackend: allow backend-specific queries

zeha:
### Short description
Nothing uses this (yet?), but this untangles the constructor, setDB and allocateStatements order.

This way one can setup extra queries in one of the gsql backends -- without this, you'll end up unable to initialise either your query string or stmt. Also, without access to `d_query_stmt` you can't implement a custom list query.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/modules/gpgsqlbackend/gpgsqlbackend.cc
+++ b/modules/gpgsqlbackend/gpgsqlbackend.cc
@@ -52,8 +52,7 @@ gPgSQLBackend::gPgSQLBackend(const string &mode, const string &suffix)  : GSQLBa
     g_log<<Logger::Error<<mode<<" Connection failed: "<<e.txtReason()<<endl;
     throw PDNSException("Unable to launch "+mode+" connection: "+e.txtReason());
   }
-
-  reconnect();
+  allocateStatements();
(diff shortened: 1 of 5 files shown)


modules/gpgsqlbackend/gpgsqlbackend.cc
```diff
@@ -52,6 +52,8 @@ gPgSQLBackend::gPgSQLBackend(const string &mode, const string &suffix)  : GSQLBa
     g_log<<Logger::Error<<mode<<" Connection failed: "<<e.txtReason()<<endl;
     throw PDNSException("Unable to launch "+mode+" connection: "+e.txtReason());
   }
+
+  reconnect();
```
rgacogne, line 56: This seems a bit wasteful since it will close the connection and establish a new one, while we likely just did that? I'm guessing there is a reason you are not calling `allocateStatements()` here but it's not obvious to me.

zeha: You're quite right, calling `reconnect()` here is wrong. I've pushed a change to use `allocateStatements()` here but will verify this works on my test branch!

PR #9360, opened 2020-07-29
dnsdist: Prevent a copy of a pool's backends when selecting a server

rgacogne:
### Short description

It makes a noticeable difference as soon as the number of backends in a pool increases to more than 8 backends, and a huge difference when there is a large number of backends.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


omoerbeek:
The regress failures a bit worrying, though

rgacogne:
> The regress failures a bit worrying, though

Both failures seemed unrelated and the tests passed after a restart.

PR #9363, opened 2020-07-29
Rec: cname chasing and rpz

### LINKED ISSUE #8233: PowerDNS Recursor RPZ CNAME resolution.
. Program: PowerDNS Recursor 4.2.0 with RZP zone
. Issue type: Feature request

### Usecase
Recursor try A record resolution locally on the RPZ Zone before the external DNS resolution.

### Description

Hi,
We're using pdns recursor with RPZ resolution for our company internal name resolution. In some of our Zones are CNAMEs.
Getting the A records works fine. But when looking up a CNAME, the answer only contains the CNAME.
The recursor seems to try to resolve the CNAME result only with external DNS not with the internal RPZ zone. If there is a A record on external DNS they work, if not, they only return CNAME without telling again RPZ Zone. 
Is there a way for the recursor pdns to try A record resolution locally on the RPZ Zone before the external DNS resolution?

Is there a way to have a complete resolution with recursor?

Full Example (if we have this entries on the RPZ zone):

```
*.devs.gesc.galilee.fr	CNAME	devs.gesc.galilee.fr
devs.gesc.galilee.fr		A	192.168.171.188	
```
```
$ dig @pcc.rpz.dns.galilee.fr *.devs.gesc.galilee.fr
;; ANSWER SECTION:
*.devs.gesc.galilee.fr. 3600	IN	CNAME	devs.gesc.galilee.fr.

$ dig @pcc.rpz.dns.galilee.fr devs.gesc.galilee.fr
;; ANSWER SECTION:
devs.gesc.galilee.fr. 3600	IN	A	192.168.171.188
```
```
With a BIND RPZ zone for example i have this result:
$ dig @pcc.rpz.dns.galilee.fr *.devs.gesc.galilee.fr
;; ANSWER SECTION:
*.devs.gesc.galilee.fr. 3600	IN	CNAME	devs.gesc.galilee.fr.
devs.gesc.galilee.fr. 3600	IN	A	192.168.171.188
```

Is there a way for the recursor pdns to try A record resolution locally on the RPZ Zone before the external DNS resolution?

Thanks in advance.
Best regards.

omoerbeek:
### Short description

This is a start for matching the CNAMEs encountered during resolving with RPZ custom matches. In particular, further resolving is subject to RPZ processing as well  (even with the same priority to allow matches in the same RPZ zone).

To allow this, the names checks are split from the client IP checks and are now done in SyncRres, including CNAME answers. This introduces some complexity and potential performance issues. One regression test also fails with this, this is to be analyzed.

Before I continue on this path, I like some feedback if this is going in the right direction and if we want this at all.

Fixes #8233

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/filterpo.cc
+++ b/pdns/filterpo.cc
@@ -200,41 +200,13 @@ bool DNSFilterEngine::getProcessingPolicy(const ComboAddress& address, const std
 
 bool DNSFilterEngine::getClientPolicy(const ComboAddress& ca, const std::unordered_map<std::string,bool>& discardedPolicies, Policy& pol) const
 {
-  // cout<<"Got question for "<<qname<<" from "<<ca.toString()<<endl;
-  std::vector<bool> zoneEnabled(d_zones.size());
-  size_t count = 0;
-  bool allEmpty = true;
+  // cout<<"Got question from "<<ca.toString()<<endl;
   for (const auto& z : d_zones) {
-    bool enabled = true;
     if (z->getPriority() >= pol.getPriority()) {
-      enabled = false;
-    } else {
-      const auto& zoneName = z->getName();
-      if (discardedPolicies.find(zoneName) != discardedPolicies.end()) {
-        enabled = false;
-      }
-      else {
-        if (z->hasClientPolicies()) {
-          allEmpty = false;
-        }
-        else {
-          enabled = false;
-        }
-      }
+      break;
     }
-
-    zoneEnabled[count] = enabled;
-    ++count;
-  }
-
-  if (allEmpty) {
-    return false;
-  }
-
-  count = 0;
-  for (const auto& z : d_zones) {
-    if (!zoneEnabled[count]) {
-      ++count;
+    const auto& zoneName = z->getName();
+    if (discardedPolicies.find(zoneName) != discardedPolicies.end()) {
       continue;
     }
 
@@ -242,10 +214,7 @@ bool DNSFilterEngine::getClientPolicy(const ComboAddress& ca, const std::unorder
       // cerr<<"Had a hit on the IP address ("<<ca.toString()<<") of the client"<<endl;
       return true;
     }
-
-    ++count;
   }
-
   return false;
 }
 
--- a/pdns/filterpo.cc
+++ b/pdns/filterpo.cc
@@ -198,9 +198,8 @@ bool DNSFilterEngine::getProcessingPolicy(const ComboAddress& address, const std
   return false;
 }
 
-bool DNSFilterEngine::getQueryPolicy(const DNSName& qname, const ComboAddress& ca, const std::unordered_map<std::string,bool>& discardedPolicies, Policy& pol) const
+bool DNSFilterEngine::getClientPolicy(const ComboAddress& ca, const std::unordered_map<std::string,bool>& discardedPolicies, Policy& pol) const
 {
-  // cout<<"Got question for "<<qname<<" from "<<ca.toString()<<endl;
   std::vector<bool> zoneEnabled(d_zones.size());
   size_t count = 0;
   bool allEmpty = true;
@@ -214,7 +213,7 @@ bool DNSFilterEngine::getQueryPolicy(const DNSName& qname, const ComboAddress& c
         enabled = false;
       }
       else {
-        if (z->hasQNamePolicies() || z->hasClientPolicies()) {
+        if (z->hasClientPolicies()) {
           allEmpty = false;
         }
         else {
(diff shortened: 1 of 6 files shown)


pdns/filterpo.cc
```diff
@@ -214,7 +214,58 @@ bool DNSFilterEngine::getQueryPolicy(const DNSName& qname, const ComboAddress& c
         enabled = false;
```
rgacogne, line 214: We only need the two-pass loop when we are matching a `DNSName` (to avoid computing all labels + wildcards as many times as the number of policies), you can mimic `bool DNSFilterEngine::getProcessingPolicy(const ComboAddress& address, const std::unordered_map<std::string,bool>& discardedPolicies, Policy& pol) const` here.

pdns/pdns_recursor.cc
```diff
@@ -1231,9 +1231,8 @@ static PolicyResult handlePolicyHit(const DNSFilterEngine::Policy& appliedPolicy
     return PolicyResult::NoAction;
 
   case DNSFilterEngine::PolicyKind::Custom:
-    ret.clear();
     res = RCode::NoError;
-    {
+    if (post && ret.size() == 0) { // can happen with NS matches, those do not fill the result
```
rgacogne, line 1235: I'm not sure we should treat that case differently?

pdns/syncres.cc
```diff
@@ -637,10 +637,57 @@ int SyncRes::asyncresolveWrapper(const ComboAddress& ip, bool ednsMANDATORY, con
...
+  auto spoofed = d_appliedPolicy.getCustomRecords(target, qtype.getCode());
+  for (const auto& dr : spoofed) {
+    if (dr.d_place != DNSResourceRecord::ANSWER) {
+      continue;
+    }
+    ret.push_back(dr);
+    auto content = getRR<CNAMERecordContent>(dr);
+    if (content) {
+      target = content->getTarget();
+      return qnameRPZHit(dfe, target, qtype, ret);
```
rgacogne, line 672: Perhaps we should do a sanity check that the target is actually different from the qname here?

rgacogne, line 672: So this would chase CNAME targets but only if they exist in a policy, this would not actually chase all CNAMEs, right? I understand this is just a start, I want to make sure I understand whether you intend to chase all CNAMEs in the end :)

omoerbeek: Added  generalized case: depth check

omoerbeek: I nthin k it is ok to keep chasing withing the RPZ, the resulting new target will be further resolved by the caller. I just need to make sure we return `false` in that case. So that is an actual bug.

pdns/pdns_recursor.cc
omoerbeek: For the moment I decided to just copy what has been done until now for NS name or IP triggers. The current code always trows an exception if a NS policy is hit, and then does a resolve without RPZ processing starting with the custom record from the policy. I have trouble seeing how to modify the code in SyncRes  to handle the NS RPZ cases in SyncRes itself.

So I decided to keep the current behaviour (with an ugly hack, true). Need to revisit this, but I might need some help in that.

omoerbeek:
Current state/issues

1. To keep current behaviour wrt an answer from gettag_ffi, we de a check on qname and if there is a hit we do not return the gettag_ffi answer but continue with RPZ processingt. Hits doing resolve will not be considered.
2. While chasing CNAMEs we do consider policies with the same priority (so that CNAME chains defined a specific RPZ zone work, see #8233). This has the consequence that there's might be another hit in the same zone for other reasons...
3. point 2 above and the fact that we do try to chase CNAMEs with RPZ  with same priority enabled causes some changes in the regressions test, specifically `regression-tests.recursor/RPZ`
4. There is code to see if we have an answer to decide if we want to continue chasing CNAMEs or are done. This all feels a bit brittle.

All in all I'm not too happy about all this, but I am out of ideas to improve this further.

rgacogne:
> This is a start for matching the CNAMEs encountered during resolving with RPZ custom matches. In particular, further resolving is subject to RPZ processing as well (even with the same priority to allow matches in the same RPZ zone).

Re-reading the RPZ specs, I'm wondering if this is the right thing to do. Especially based on Section 6 [1]:
> RPZ rules do not apply to synthetic data generated by using RPZ
   rules.  For example, if RPZ supplies a CNAME pointing to a walled
   garden, RPZ policies will not be used while following that CNAME.  If
   RPZ supplies local data giving a particular A record, RPZ policies
   will not apply to that response IP address.

[1]: https://tools.ietf.org/html/draft-vixie-dnsop-dns-rpz-00#section-6

omoerbeek:
> > This is a start for matching the CNAMEs encountered during resolving with RPZ custom matches. In particular, further resolving is subject to RPZ processing as well (even with the same priority to allow matches in the same RPZ zone).
> 
> Re-reading the RPZ specs, I'm wondering if this is the right thing to do. Especially based on Section 6 [1](https://tools.ietf.org/html/draft-vixie-dnsop-dns-rpz-00#section-6):
> 
> > RPZ rules do not apply to synthetic data generated by using RPZ
> > rules.  For example, if RPZ supplies a CNAME pointing to a walled
> > garden, RPZ policies will not be used while following that CNAME.  If
> > RPZ supplies local data giving a particular A record, RPZ policies
> > will not apply to that response IP address.

Did not realize that. I was set on the wrong foot by #8233. I think we can close that one, referring to the quote above.

rgacogne:
> Did not realize that. I was set on the wrong foot by #8233. I think we can close that one, referring to the quote above.

We still need to apply filtering policies to CNAME chains, though, so I'll keep working on this.

omoerbeek:
Continued in #9414

PR #9369, opened 2020-07-31
SVCB and HTTPS support

pieterlexis:
### Short description
PR to add SVCB and HTTPS support.

ToDo:

- [x] Implement `SvcParam` class for internal representation of service parameters
- [x] Add SVB
  - [x] qtype.cc/hh
  - [x] dnsrecords.cc/hh + `reportOtherTypes`
- [x] Add HTTPS to QTypes and dnsrecords
  - [x] qtype.cc/hh
  - [x] dnsrecords.cc/hh + `reportOtherTypes`
- [x] Implement Reading/Writing SvcParams (`xfrSvcParamKeyVals`)
  - [x] `RecordTextReader`
  - [x] `RecordTextWriter`
  - [x] `DNSPacketWriter`
  - [x] `PacketReader`
- [x] Add `pdnsutil` tests for SVCB/HTTPS correctness
- [x] Implement auth additional processing
  - [x] Add tests for auth additional processing
- [ ] Implement rec additional processing
  - [ ] Add tests for rec additional processing

### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code


pdns/Makefile.am
```diff
@@ -1143,6 +1166,7 @@ dnsgram_SOURCES = \
 	nsecrecords.cc \
 	qtype.cc \
 	rcpgenerator.cc rcpgenerator.hh \
+	svc-records.cc \
```
zeha, line 1169: supernit: somewhat-but-not-quite alphabetical sorting ;)

pieterlexis:
After a discussion with @Habbie, some features will be moved to a next iteration so this initial code can land in auth 4.4.0-alpha1. These features will be:

* Recursor additional processing
* CNAME processing for additional records (requires a refactor of `packethandler::doQuestion`)

Habbie:
upgrading.rst notes that IPSECKEY is no longer unknown; it should note the same for SVCB and HTTPS.

pdns/dnsrecords.hh
```diff
@@ -496,6 +497,31 @@ private:
   string d_keyring;
 };
 
+class SVCBRecordContent : public DNSRecordContent
+{
+public:
+  includeboilerplate(SVCB)
+  const DNSName getTarget() const {return d_target;}
```
rgacogne, line 504:
```suggestion
  const DNSName& getTarget() const {return d_target;}
```
Otherwise it does not make a lot of sense to mark the return value as `const`.

pdns/dnsrecords.hh
```diff
@@ -496,6 +497,31 @@ private:
...
+  uint16_t d_priority;
+  DNSName d_target;
+  set<SvcParam> d_params;
+};
+
+class HTTPSRecordContent : public DNSRecordContent
+{
+public:
+  includeboilerplate(HTTPS)
+  const DNSName getTarget() const {return d_target;}
```
rgacogne, line 517: Same here
```suggestion
  const DNSName& getTarget() const {return d_target;}
```

pdns/packethandler.cc
```diff
@@ -438,13 +438,51 @@ bool PacketHandler::getBestWildcard(DNSPacket& p, const SOAData& sd, const DNSNa
...
+          auto rrc = getRR<HTTPSRecordContent>(rr.dr);
+          r->addRecord(std::move(rr));
+          ret = rrc->getTarget().isRoot() ? ret : rrc->getTarget();
+          if (rrc->getPriority() == 0) {
+            done = false;
+          }
+          break;
+        }
+        default:
+          abort();
```
rgacogne, line 471: That sounds a bit harsh?

pdns/pdnsutil.cc
```diff
@@ -401,6 +405,39 @@ static int checkZone(DNSSECKeeper &dk, UeberBackend &B, const DNSName& zone, con
       continue;
     }
 
+    if (rr.qtype.getCode() == QType::SVCB || rr.qtype.getCode() == QType::HTTPS) {
+      vector<string> parts;
+      stringtok(parts, rr.content);
+      if (std::atoi(parts.at(0).c_str()) == 0 && parts.size() > 2) {
```
rgacogne, line 411: Wouldn't `parts.at(0)` throw if `rr.content` is empty, or contains only white spaces?

pdns/svc-records.hh
```diff
@@ -0,0 +1,98 @@
...
+  static std::string keyToString(const SvcParamKey &k);
+
+  bool operator< (const SvcParam &other) const;
+
+  SvcParamKey getKey() const {
+    return d_key;
+  }
+
+  uint16_t getPort() const;
+  std::vector<ComboAddress> getIPHints() const;
```
rgacogne, line 81: How about returning const references instead, for all these methods returning heavy objects?

pdns/dnsparser.cc
```diff
@@ -553,6 +553,107 @@ void PacketReader::xfrBlob(string& blob, int length)
...
+        throw std::out_of_range("empty 'mandatory' values");
+      }
+      std::set<SvcParam::SvcParamKey> paramKeys;
+      size_t stop = d_pos + len;
+      while (d_pos < stop) {
+        uint16_t keyval;
+        xfr16BitInt(keyval);
+        paramKeys.insert(static_cast<SvcParam::SvcParamKey>(keyval));
+      }
+      kvs.insert(SvcParam(key, paramKeys));
```
rgacogne, line 587: It would be a nice improvement to make the constructors move the second parameter, instead of doing copies.

modules/remotebackend/Makefile.am
```diff
@@ -123,6 +123,7 @@ libtestremotebackend_la_SOURCES = \
 	../../pdns/nameserver.cc \
 	../../pdns/rcpgenerator.cc \
 	../../pdns/unix_utility.cc \
+	../../pdns/svc-records.cc \
```
rgacogne, line 126:
```suggestion
	../../pdns/svc-records.hh ../../pdns/svc-records.cc \
```

pdns/Makefile.am
```diff
@@ -199,6 +199,7 @@ pdns_server_SOURCES = \
 	qtype.cc qtype.hh \
 	query-local-address.hh query-local-address.cc \
 	rcpgenerator.cc \
+	svc-records.cc \
```
rgacogne, line 202:
```suggestion
	svc-records.cc svc-records.hh \
```

modules/ldapbackend/dnsdomain2.schema
```diff
@@ -212,6 +212,18 @@ attributetype ( 1.3.6.1.4.1.2428.20.1.61 NAME 'openPGPKeyRecord'
 	SUBSTR caseIgnoreIA5SubstringsMatch
 	SYNTAX 1.3.6.1.4.1.1466.115.121.1.26 )
 
+attributetype ( 1.3.6.1.4.1.2428.20.1.64 NAME 'sVCBRecord'
```
zeha, line 215: idk, but newer records apparently stopped with the 'first char downcase' rule already, so lets not continue it? (cf EUI64Record)

pdns/Makefile.am
```diff
@@ -315,6 +316,7 @@ pdnsutil_SOURCES = \
 	pdnsutil.cc \
 	qtype.cc \
 	rcpgenerator.cc rcpgenerator.hh \
+	svc-records.cc svc-records.hh \
```
zeha, line 319: you're the autoconf expert, but it looks weird sometimes having and sometimes _not_ having the `.hh` file in the `..._SOURCES` lists.

pdns/dnsrecords.cc
```diff
@@ -326,6 +326,22 @@ boilerplate_conv(OPENPGPKEY, 61,
                  conv.xfrBlob(d_keyring);
                  )
 
+boilerplate_conv(SVCB, 64,
+                 conv.xfr16BitInt(d_priority);
+                 conv.xfrName(d_target, false, true);
+                 if (d_priority != 0) {
+                  conv.xfrSvcParamKeyVals(d_params);
```
zeha, line 333: 1space indent seems weird

rgacogne (commented):
A few nits but the code looks solid. :heart: the tests :-)

pdns/packethandler.cc
zeha, line 471: throw PDNSException(...

zeha (changes requested):
some initial nits.

Also: the draft suggests that svcParamKeys need to be in increasing numeric order. Is this something the code should do?

zeha:
> Also: the draft suggests that svcParamKeys need to be in increasing numeric order. Is this something the code should do?

To answer my own question: probably not, because thats only per RR and not per RRset. Fine (:

zeha:
Might make sense to add an API test (didn't verify this works):

```patch
diff --git a/regression-tests.api/test_Zones.py b/regression-tests.api/test_Zones.py
index e2069d8ef..2a4c62717 100644
--- a/regression-tests.api/test_Zones.py
+++ b/regression-tests.api/test_Zones.py
@@ -1462,6 +1462,25 @@ $ORIGIN %NAME%
                                headers={'content-type': 'application/json'})
         self.assert_success(r)  # user should be able to create DNAME at APEX as per RFC 6672 section 2.3

+    def test_rr_svcb(self):
+        name, payload, zone = self.create_zone()
+        rrset = {
+            'changetype': 'replace',
+            'name': 'svcb.' + name,
+            'type': 'SOA',
+            'ttl': 3600,
+            'records': [
+                {
+                    "content": 'mandatory=alpn alpn=h2,h3 ipv4hint=192.0.2.1,192.0.2.2 echconfig="dG90YWxseSBib2d1cyBlY2hjb25maWcgdmFsdWU="',
+                    "disabled": False
+                },
+            ]
+        }
+        payload = {'rrsets': [rrset]}
+        r = self.session.patch(self.url("/api/v1/servers/localhost/zones/" + name), data=json.dumps(payload),
+                               headers={'content-type': 'application/json'})
+        self.assert_success(r)
+
     def test_rrset_ns_dname_exclude(self):
         name, payload, zone = self.create_zone()
         rrset = {
```

pdns/Makefile.am
pieterlexis: You only need it in one place so it is disted, but it might make sense to specify it everywhere indeed

pdns/packethandler.cc
pieterlexis: it is a programmer error if it ever triggered, but as @zeha suggests, I'll make it a PDNSException

pdns/pdnsutil.cc
pieterlexis: It shoudl have already error'd above (in the parsing step). But perhaps we need to eb more careful indeed.

pdns/pdnsutil.cc
```diff
@@ -541,6 +578,54 @@ static int checkZone(DNSSECKeeper &dk, UeberBackend &B, const DNSName& zone, con
...
+
+    if (name == target) {
+      cout<<"[Error] SVCB record "<<name<<" has itself as target."<<endl;
+      numerrors++;
+    }
+
+    if (prio == 0) {
+      if (target.isPartOf(zone)) {
+        if (svcbAliases.find(target) != svcbAliases.end()) {
+          cout << "[Warning] SVB record for " << name << " has an aliasform target (" << target << ") that is in aliasform itself." << endl;
```
zeha, line 594:
```suggestion
          cout << "[Warning] SVCB record for " << name << " has an aliasform target (" << target << ") that is in aliasform itself." << endl;
```

pdns/pdnsutil.cc
```diff
@@ -541,6 +578,54 @@ static int checkZone(DNSSECKeeper &dk, UeberBackend &B, const DNSName& zone, con
...
+          numwarnings++;
+        }
+        if (addresses.find(target) == addresses.end() && svcbRecords.find(target) == svcbRecords.end()) {
+          cout<<"[Error] SVBC record "<<name<<" has a target "<<target<<" that has neither address nor SVCB records."<<endl;
```
zeha, line 598:
```suggestion
          cout<<"[Error] SVCB record "<<name<<" has a target "<<target<<" that has neither address nor SVCB records."<<endl;
```

pdns/rcpgenerator.cc
```diff
@@ -607,6 +718,61 @@ void RecordTextWriter::xfrHexBlob(const string& val, bool)
   }
 }
 
+void RecordTextWriter::xfrSvcParamKeyVals(const set<SvcParam>& val) {
+  for (auto const &param : val) {
+    if (!d_string.empty())
```
zeha, line 723: is the intent "d_string is empty" or "is this not the first value"? if the latter, i'd suggest a `bool first` instead :)

pdns/svc-records.cc
```diff
@@ -0,0 +1,181 @@
...
+  }
+  d_value = value;
+}
+
+SvcParam::SvcParam(const SvcParamKey &key, const std::vector<std::string> & value) {
+  d_key = key;
+  if (d_key != SvcParamKey::alpn) {
+    throw std::invalid_argument("can not create SvcParam for " + keyToString(key) + " with a string-set value");
+  }
+  if (d_key == SvcParamKey::alpn) {
```
zeha, line 87: this if looks redundant :)

pdns/pdnsutil.cc
rgacogne: If we are not supposed to reach that part of the code with an invalid content, I'm fine with an exception being raised. :-)

pieterlexis: fwiw, on [line 399](https://github.com/PowerDNS/pdns/pull/9369/files#diff-e69997cee73802012aa0260286fbd5c7R399) we `continue` if marshalling fails.

pdns/svc-records.cc
pieterlexis: for now, yes. But there might be other string-set Service params in the future.

pdns/Makefile.am
pieterlexis: I'll make a ticket that we need to sort them all :)

modules/ldapbackend/dnsdomain2.schema
Habbie: Per Howard Chu: don't know why people do that, probably historical. Conformity in a schema is a good thing. Also, they are case-insensitive.

Please uppercase :)

Habbie (approved):

PR #9374, opened 2020-08-07
Be explicit about IPs vs netmasks.

omoerbeek:
### Checklist

I have:
- [X] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/recursordist/docs/settings.rst
+++ b/pdns/recursordist/docs/settings.rst
@@ -16,7 +16,7 @@ As an example:
 
 ``allow-from``
 --------------
--  IP ranges, separated by commas
+-  IP addresses or netmasks, separated by commas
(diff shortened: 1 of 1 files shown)


pdns/recursordist/docs/settings.rst
```diff
@@ -1838,14 +1838,16 @@ IP address for the webserver to listen on.
 
 ``webserver-allow-from``
 ------------------------
--  IP addresses, comma separated
+-  IP ranges, comma separated
```
cmouse, line 1841: range would be something like 127.0.0.1-127.0.0.5, this should probably use word like prefix or something else?

pieterlexis: "IP ranges as netmasks"?

omoerbeek: I am copied some words from other "allow-from" settings. But I think the term IP range is confusing. I'll change the others as well.

PR #9375, opened 2020-08-10
rec: Allow some more depth headroom for the no-qname-minimization fallback case

omoerbeek:
If we hit maxdepth, we still want to have some room to do the doResolveNoQNameMinimization call.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code


omoerbeek:
@rgacogne any comment? This could also be done by throwing an exception and doing the fallback resolve in the catch block, but that is scary as it involves way more restructuring.

rgacogne:
I don't have a better idea on how to handle this at the moment, so I think we should go ahead with that solution. Arbitrarily cutting the depth in half seems weird but no value would feel right either so I'm fine with that. If we document that change I'm OK with this pull request.

Issue #9058, opened 2020-04-24
Additional data for RPZ logging/protobuf


FrancisTurner:
In order to provide details for reporting and/or troubleshooting it would be helpful if there was an option to report the actual zone record that caused an RPZ hit.

In bind syslogs you have a "via blahblah.." stanza in the log line that does this e.g.
```
17-Mar-2020 09:34:49.887 rpz: info: client 192.168.123.10#53112 (casasur.cl): rpz QNAME NODATA rewrite casasur.cl via casasur.cl.phishy.di000001.rpz.threatstop.local
```

This allows a reporting tool to look at the via section and get more information on it. It also allows troubleshooting where you (for example) discover that you were blocking *.cl instead of *.cn

For RPZ hits that work on dnames, such as in the example above, the qname is (plus or minus a *.) enough of a hint that it is relatively simple (if hacky) to identify the line with a suitable regex on the text form of the zone but this does not work at all for hits based on IP, NSIP or NSDNAME . 
e.g. if I have a hit based on NSIP a bind server tells me this
```
19-Mar-2020 09:00:45.878 rpz: info: client 192.168.123.12#55929 (peccsr.com): rpz NSIP CNAME rewrite peccsr.com via 29.120.82.251.162.rpz-nsip.phishy.di000001.rpz.threatstop.local
```

I can see no way that power dns would be able to give me the equivalent information. I've looked at the lua scripting to see if there's a way to get access to this information but as far as I can see the appliedPolicy part of the DNSQuestion object that lua scripting has access to just reports the policyName but not the line within the policy that caused the rewrite.

Ideally there would be a rpz flag that would allow this data to be output in the protobuf but it would be OK if it were an option in the DQ object so I could write a lua handler to process it.

omoerbeek:
#9376 should fix this

PR #9376, opened 2020-08-10
Rec: rpz store trigger in appliedPolicy and protobuf message

omoerbeek:
### Short description

In RPZ speak, the trigger is the lefthand side of an RPZ rule. If a RPZ rule is hit, store the value of this trigger in the `appliedPolicy.policyTrigger` field so LUA can access it and also store it in the protobuf message.
Also add the value that caused the hit. 

While working on this, I noticed the docs mention `appliedPolicy.policyAction`, but the code does not set it afaiks. I'm assuming `appliedPolicy.policyType` was meant. Adapt the code and docs to agree with respect to this  and add `policytypes` listing the possible values for `policyType`.

Also added code to use the information in appliedPolicy to do logging in a consistent manner.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [X] documented the code

### DIFF
--- a/pdns/filterpo.cc
+++ b/pdns/filterpo.cc
@@ -27,6 +27,14 @@
 #include "namespaces.hh"
 #include "dnsrecords.hh"
 
+static const string rpzDropName("rpz-drop."),
+  rpzTruncateName("rpz-tcp-only."),
+  rpzNoActionName("rpz-passthru."),
+  rpzClientIPName("rpz-client-ip"),
+  rpzIPName("rpz-ip."),
+  rpzNSDnameName("rpz-nsdname."),
+  rpzNSIPName("rpz-nsip.");
+
 DNSFilterEngine::DNSFilterEngine()
 {
 }
@@ -174,7 +182,7 @@ bool DNSFilterEngine::getProcessingPolicy(const DNSName& qname, const std::unord
       if (z->findExactNSPolicy(wc, pol)) {
         // cerr<<"Had a hit on the nameserver ("<<qname<<") used to process the query"<<endl;
         pol.d_trigger = wc;
-        pol.d_trigger.appendRawLabel("rpz-nsdname");
+        pol.d_trigger.appendRawLabel(rpzNSDnameName);
         return true;
       }
     }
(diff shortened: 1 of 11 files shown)


omoerbeek:
The extra logging should address #9058

omoerbeek:
This PR will likely conflict with other ongoing RPZ work. After that is merged, I'll rebase.

pdns/filterpo.cc
```diff
@@ -27,6 +27,18 @@
 #include "namespaces.hh"
 #include "dnsrecords.hh"
 
+// Names below are RPZ Actions and end with a dot (execpt "Local Data")
```
rgacogne, line 30:
```suggestion
// Names below are RPZ Actions and end with a dot (except "Local Data")
```

pdns/filterpo.cc
```diff
@@ -94,6 +106,8 @@ bool DNSFilterEngine::Zone::findNamedPolicy(const std::unordered_map<DNSName, DN
     iter = polmap.find(g_wildcarddnsname+s);
     if(iter != polmap.end()) {
       pol=iter->second;
+      pol.d_trigger = g_wildcarddnsname+s;
```
rgacogne, line 109:
```suggestion
      pol.d_trigger = iter->first;
```

pdns/lua-recursor4.cc
```diff
@@ -188,6 +188,8 @@ void RecursorLua4::postPrepareContext()
   d_lw->registerMember("policyKind", &DNSFilterEngine::Policy::d_kind);
   d_lw->registerMember("policyType", &DNSFilterEngine::Policy::d_type);
   d_lw->registerMember("policyTTL", &DNSFilterEngine::Policy::d_ttl);
+  d_lw->registerMember("policyTrigger", &DNSFilterEngine::Policy::d_trigger);
+  d_lw->registerMember("policyHit", &DNSFilterEngine::Policy::d_hit);
```
rgacogne, line 192: Looks like these are not added to the documentation?

rgacogne (commented):
I love that pull request! I wonder if we should lower the log level, though, or perhaps hide that logging behind a setting?

omoerbeek:
> I love that pull request! I wonder if we should lower the log level, though, or perhaps hide that logging behind a setting?

rebased and processed your suggestions. Logging is now centralized in the `handlePolicyHit` functions. The log lines are only generated when tracing (in general or for a specific name). Do you want to change that?

rgacogne (approved):

rgacogne:
> Logging is now centralized in the handlePolicyHit functions. The log lines are only generated when tracing (in general or for a specific name). Do you want to change that?

No, that's great!

PR #9387, opened 2020-08-17
auth slave: log successful NOTIFY

zeha:
### Short description

The Auth already logs all NOTIFY cases except the successful one on `Notice` or higher, lets do the same on success.

### Checklist

I have:
- [x] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/packethandler.cc
+++ b/pdns/packethandler.cc
@@ -870,7 +870,7 @@ int PacketHandler::trySuperMasterSynchronous(const DNSPacket& p, const DNSName& 
 
 int PacketHandler::processNotify(const DNSPacket& p)
 {
-  /* now what? 
+  /* now what?
      was this notification from an approved address?
      was this notification approved by TSIG?
      We determine our internal SOA id (via UeberBackend)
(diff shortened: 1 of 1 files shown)


pdns/packethandler.cc
```diff
@@ -948,7 +948,7 @@ int PacketHandler::processNotify(const DNSPacket& p)
   }
 
   if(::arg().mustDo("slave")) {
-    g_log<<Logger::Debug<<"Queueing slave check for "<<p.qdomain<<endl;
+    g_log<<Logger::Notice<<"Received NOTIFY for "<<p.qdomain<<" from "<<p.getRemote()<<" - queuing check"<<endl;
```
Habbie, line 951:
```suggestion
    g_log<<Logger::Notice<<"Received NOTIFY for "<<p.qdomain<<" from "<<p.getRemote()<<" - queueing check"<<endl;
```

While a quick search suggests both spellings might be valid, I much prefer the extra e.

klaus3000:
I prefer "queueing slave check" to explicit what kind of check is queued

zeha:
> I prefer "queueing slave check" to be explicit what kind of check is queued

I was trying to avoid "slave" here, to avoid another change later.

pdns/packethandler.cc
zeha: fixed

Habbie (approved):

PR #9388, opened 2020-08-17
Fix docs for DNSName::equal()

### LINKED ISSUE #7510: Recursor Lua's DNSName:equal() does not work as documented
- Program: Recursor
 - Issue type: Bug report

### Short description
Within the Recursor's Lua scripting, `DNSName:equal()` doesn't accept other `DNSName` values as a parameter, but does seem to accept strings. The Recursor documentation describes [`DNSName:equal`](https://doc.powerdns.com/recursor/lua-scripting/dnsname.html#DNSName:equal) as accepting a `name` parameter of type `DNSName`.

### Environment
 - Operating system: CentOS Linux 7.5
 - Software version: PowerDNS Recursor 4.1.3, Lua 5.1
 - Software source: RPM built from [source tar from PowerDNS.com](https://downloads.powerdns.com/releases/pdns-recursor-4.1.3.tar.bz2).

### Steps to reproduce
Compare a `DNSName` to itself or any other `DNSName` in the Recursor scripting context.

For example, configure the recursor to load this Lua script:
```lua
function preresolve(dq)
    pdnslog("Is example.com. equal to itself?")
    local result = newDN("example.com."):equal(newDN("example.com."))
    pdnslog("result is " .. tostring(result))

    return false
end
```

### Expected behaviour
The script has no affect on DNS queries. Recursor logs should include:
```
Is example.com. equal to itself?
result is true
```

### Actual behaviour
Recursor logs include an STL error and the query fails:
```
Is example.com. equal to itself?
STL error (myexample.com/A from 192.168.32.2): Unable to convert parameter from userdata to Ss
```

### Other information
The Authority's reference documentation for [`DNSName`](https://doc.powerdns.com/authoritative/lua-records/reference/dnsname.html#DNSName) describes `:equal` taking the `name` parameter as a string instead. If a string value is given, the scripting doesn't produce an error, and it *seems* to compare to the `DNSName` correctly. I'm unsure if Authority and Recursor Lua scripting both use the same implementation of `DNSName`.

Using `DNSName:equal()` with a string instead of another `DNSName` does seem to be how the type is used in example code [in other issues](https://github.com/PowerDNS/pdns/issues/4552#issuecomment-356083098).

omoerbeek:
Since overloading Lua wrapper functions does not seem to be possible,
fix docs for DNSName:equal(). But note that == is defined for DNSNames.

Fixes #7510

### Short description

Also zap a duplicated line in `lua-base4.cc`.

### Checklist

I have:
- [X] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/recursordist/docs/lua-scripting/dnsname.rst
+++ b/pdns/recursordist/docs/lua-scripting/dnsname.rst
@@ -50,8 +50,10 @@ A small example of the functionality of a :class:`DNSName` is shown below:
   .. method:: DNSName:equal(name) -> bool
 
     Returns true when both names are equal for the DNS, i.e case insensitive.
+    
+    To compare two ``DNSName`` objects, use ``==``.
 
-    :param DNSName name: The name to compare against.
+    :param DNSName string: The name to compare against.
 
   .. method:: DNSName:isPartOf(name) -> bool
 


Habbie:
> Since overloading Lua wrapper functions does not seem to be possible,

Just for later reference, this is how overloading can be done. It is a bit tedious: https://github.com/PowerDNS/pdns/blob/master/pdns/lua-base4.cc#L130

pdns/recursordist/docs/lua-scripting/dnsname.rst
```diff
@@ -51,7 +51,7 @@ A small example of the functionality of a :class:`DNSName` is shown below:
 
     Returns true when both names are equal for the DNS, i.e case insensitive.
```
Habbie, line 52:
```suggestion
    Returns true when both names are equal for the DNS, i.e case insensitive.
    
    To compare two ``DNSName`` objects, use ``==``.
```

omoerbeek:
Hmm, I now see `wirelength` vbs `wireLength`. That part should be left out.

PR #9401, opened 2020-08-20
statbag: handle key redeclarations better

Habbie:
### Short description
This avoids auth testrunner failures on 32 bit architectures. Possibly it also avoids silent memory corruption on 64 bit architectures.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/receiver.cc
+++ b/pdns/receiver.cc
@@ -636,7 +636,13 @@ int main(int argc, char **argv)
     exit(1);
   }
   
-  declareStats();
+  try {
+    declareStats();
+  }
+  catch(PDNSException &PE) {
+    g_log<<Logger::Error<<"Exiting because: "<<PE.reason<<endl;
+    exit(1);
+  }
   S.blacklist("special-memory-usage");
 
   DLOG(g_log<<Logger::Warning<<"Verbose logging in effect"<<endl);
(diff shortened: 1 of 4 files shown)


rgacogne:
> Possibly it also avoids silent memory corruption on 64 bit architectures.

To be clear that should not happen in the regular binaries, right?

Habbie:
> To be clear that should not happen in the regular binaries, right?

Right. I've been pondering making the behaviour configurable - set to zero in 'test' mode, throw on redeclare in production mode. Thoughts?

Habbie:
This commit (on a different branch) throws on re-declare, it's running through CI now: https://github.com/Habbie/pdns/commit/2cdcd11271e20d93359c4676e20429d7dbab26b5

Habbie:
> This branch throws on re-declare, it's running through CI now: https://github.com/Habbie/pdns/commits/statbag-redeclare-throw

It passed. This should mean we don't do any redeclarations in our existing production code.

rgacogne:
> I've been pondering making the behaviour configurable - set to zero in 'test' mode, throw on redeclare in production mode.

That seems like a very neat idea, as long as it doesn't complicate the code much :-)

rgacogne (approved):

PR #9408, opened 2020-08-23
Fix building with LLVM11

RvdE:
There's a typo on line 529, where it says "true", but the double quotes cause a type cast error on LLVM11.

dnsdist-console.cc:529:32: error: type 'const char [5]' cannot be narrowed to 'bool' in initializer list [-Wc++11-narrowing]
  { "SetNegativeAndSOAAction", "true", "nxd, zone, ttl, mname, rname, serial, refresh, retry, expire, minimum [, options]", "Turn a query into a NXDomain or NoData answer and sets a SOA record in the additional section" },
                               ^~~~~~
dnsdist-console.cc:529:32: note: insert an explicit cast to silence this issue
  { "SetNegativeAndSOAAction", "true", "nxd, zone, ttl, mname, rname, serial, refresh, retry, expire, minimum [, options]", "Turn a query into a NXDomain or NoData answer and sets a SOA record in the additional section" },
                               ^~~~~~
                               static_cast<bool>( )
1 error generated.

### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code


rgacogne (approved):
Thanks!

PR #9411, opened 2020-08-24
rec: Fix warning: initialized lambda captures are a C++14 extension

omoerbeek:
### Short description

Please double check this twice if it is ok for the capture to be done like this. data is thread local, is it safe to use a reference here?

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/pdns_recursor.cc
+++ b/pdns/pdns_recursor.cc
@@ -2855,7 +2855,7 @@ static void handleNewUDPQuestion(int fd, FDMultiplexer::funcparam_t& var)
           }
 
           if(g_weDistributeQueries) {
-            const std::string& localdata = data;
+            std::string localdata = data;
             distributeAsyncFunction(data, [localdata, fromaddr, dest, source, destination, tv, fd, proxyProtocolValues]() mutable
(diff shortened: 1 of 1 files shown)


pdns/pdns_recursor.cc
```diff
@@ -2855,8 +2855,9 @@ static void handleNewUDPQuestion(int fd, FDMultiplexer::funcparam_t& var)
           }
 
           if(g_weDistributeQueries) {
-            distributeAsyncFunction(data, [data = data, fromaddr, dest, source, destination, tv, fd, proxyProtocolValues]() mutable
-              { return doProcessUDPQuestion(data, fromaddr, dest, source, destination, tv, fd, proxyProtocolValues); });
+            const std::string& localdata = data;
```
rgacogne, line 2858: How about we make it obvious here that the data is copied? It will be captured by copy in the lambda anyway (references are stripped) but it might currently not be immediately obvious to the reader.
```suggestion
            std::string localdata = data;
```

omoerbeek: I was bit worried about excessive copying.., but yes.

PR #9414, opened 2020-08-24
rec: Apply filtering policies (RPZ) on CNAME chains as well

rgacogne:
### Short description

This PR changes the way we handle filtering policies (RPZ) in the recursor:
- Name encountered in CNAME chains are now filtered as well ;
- The actual handling of QNAME and response records rules is moved to the SyncRes, so that we can better match the RPZ specifications by applying rules during the correct ""stage of resolution" (mainly CNAME sub-queries) ;
- A new `policyEventFilter` Lua hook is created, and is called every time a rule matches. This makes it easier to intercept and alter a policy hit, as it was before hard to understand what could be handled in `preresolve` or in `postresolve`.

This PR is based on the great work started by Otto in #9363 (thanks!).

Comments welcome, especially as this changes the existing behaviour with regard to intercepting RPZ hits via Lua hooks.
Maybe a few unit tests would be nice as well :-)

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code


rgacogne:
My last update broke our DNS64 code, investigating why.

rgacogne:
> My last update broke our DNS64 code, investigating why.

So what happens is that the "Step0" part of QName Minimization calls `doResolveNoQNameMinimization()` but with recursion disabled. So the `doCNAMECacheCheck()` calls done from there calls `doResolve` for the target of the CNAME with no recursion either, and of course that fails if the answer is not in the cache. It worked before that pull request because `fromCache` was not set when `doCNAMECacheCheck()`, causing the QM code to go to the next step. Now that `fromCache` is set, QM considers that the resolution is done and stops there, which is wrong.
One simple solution would be to only set `fromCache` when recursion is enabled, but that feels wrong so I'll try to find a better one..

omoerbeek:
The old comment hints at that. But it should have read something like:: even if the CNAMe is in the cache, that does not mean the final result (which involves chasing the CNAME) will be.

rgacogne:
Oh, yeah, trust me I'll update the comment when I'm done :-)

rgacogne:
I pushed a fix with a comment, let me know if you want to add more words to it!

pdns/syncres.cc
```diff
@@ -876,7 +876,13 @@ int SyncRes::doResolveNoQNameMinimization(const DNSName &qname, const QType &qty
...
     if (doCNAMECacheCheck(qname, qtype, ret, depth, res, state, wasAuthZone, wasForwardRecurse)) { // will reroute us if needed
       d_wasOutOfBand = wasAuthZone;
-      if (fromCache) {
+      // Here we have an issue. If we were prevented from going out to the network (cache-only was set, possibly because we
+      // are in QM Step0) we might have a CNAME but not the corresponding target.
+      // It means that we will sometimes go to the next steps when we are in fact done, but that's fine since
+      // we will get the records from the cache, resulting in a small overhead.
+      // This might be a real problem if we had a RPZ hit, though, because we do not want the processing to continue, since
+      // RPZ rules will not be evaluated anymore (we already matched).
+      if (fromCache && (!d_cacheonly || (d_appliedPolicy.d_type != DNSFilterEngine::PolicyType::None && d_appliedPolicy.d_kind != DNSFilterEngine::PolicyKind::NoAction))) {
```
omoerbeek, line 885: For readability, add sth like:
```
bool rpzHit = d_appliedPolicy.d_type != DNSFilterEngine::PolicyType::None && d_appliedPolicy.d_kind != DNSFilterEngine::PolicyKind::NoAction;
```
and then use rpzHIt/!rpzHit below in two places?

rgacogne: Will do!

omoerbeek (approved):

PR #9432, opened 2020-09-01
Rec: make parse ip:port a bit smarter

### LINKED ISSUE #7743: Misinterpretation of IPv6 in forward-zones/forward-zones-recurse configuration
- Program: Recursor 
 - Issue type: Bug report 

### Short description
The configuration recursor.conf is not properly processed for non-bracket IPv6 adresses, even when in most other places processing IPv6 adresses, including when mixing IPv4 and IPv6 without special indication for IPv6 works fine.

### Environment

 - Operating system: uname -a says: Linux 4.15.0-47-generic #50-Ubuntu SMP Wed Mar 13 10:44:52 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux
 - Software version: PowerDNS Recursor 4.1.1 (C) 2001-2017 PowerDNS.COM BV
 - Software source:  
```
curl https://repo.powerdns.com/CBC8B383-pub.asc | sudo apt-key add -
apt-get update
```
Later in the install i see
```
<13>Apr 23 14:35:26 dns-resolver-dev2: Get:4 http://archive.ubuntu.com/ubuntu bionic/universe amd64 pdns-recursor amd64 4.1.1-2 [1424 kB]
<13>Apr 23 14:35:27 dns-resolver-dev2: Fetched 1546 kB in 0s (43.8 MB/s)
[...]
<13>Apr 23 14:35:27 dns-resolver-dev2: Selecting previously unselected package pdns-recursor.
<13>Apr 23 14:35:27 dns-resolver-dev2: Preparing to unpack .../pdns-recursor_4.1.1-2_amd64.deb ...
<13>Apr 23 14:35:27 dns-resolver-dev2: Unpacking pdns-recursor (4.1.1-2) ...
[...]
<13>Apr 23 14:35:27 dns-resolver-dev2: Setting up pdns-recursor (4.1.1-2) ...
```

### Steps to reproduce

1.  Stop pdns_recursor
2. Insert configuration into /etc/powerdns/recursor.conf as follows
```
#################################
# forward-zones-recurse Zones for which we forward queries with recursion bit, comma separated domain=ip pairs
# forward-zones-recurse=
# this fails by getting misconfiguration live without throwing an error
forward-zones-recurse=testthis.net=2003:1234::f561;2003:1234::f562
```
2. Start pdns_recursor
3. Check which adresses were chosen for forwarding by looking into syslog
4. Stop pdns_recursor 
5. Modify configuration as follows
```
forward-zones-recurse=testthis.net=[2003:1234::f561];[2003:1234::f562]
```
6. Start pdns_recursor
7. Check which addresses were chosen for forwarding by looking into syslog

### Expected behavior
Expected is that pdns_recursor 
a) will bind to the same address in both choices of configuration
b) would throw an error if there is ambiguity in configuration

When no errror coccurs, syslog shows that configuration for forwarding chose a pair of proper IPv6 addresses:
```
Apr 23 13:41:20 bn831x5a-idns4-lab pdns_recursor[2683]: Redirecting queries for zone 'testthis.net' with recursion to: [2003:1234::f561]:53, [2003:1234::f562]:53
```
### Actual behavior
When choosing this line - which should be expected to work the same way as elsewhere in the configuration file, that is: Without having to embrace the IPv6 address with edged brackets.

```
forward-zones-recurse=testthis.net=2003:1234::f561;2003:1234::f562
```
but syslog shows that configuration for forwarding has seemingly tried to interpret part of config as IPv4 address, and omitted the portion it did not understand, without throwing an error:

```
Apr 23 13:40:04 bn831x5a-idns4-lab pdns_recursor[2627]: Redirecting queries for zone 'testthis.net' with recursion to: 0.0.7.211:1234, 0.0.7.211:1234
```

### Other information

The same inconsistent behavior is observed for forward-zones and forward-zones-recurse

### Usecase

I try to have the resolver instance resolve data from a three independent pairs of IPv6-only 
on premise internal DNS Authoritiative services 

### Description

Consistently accept IPv6 addresses without requiring user to put brackets around them in some places in recursor.conf whereas such brackets are not required in other places of same file

omoerbeek:
### Short description

Currently an IPv6 addres makes it use the explicit port case, interpreting the prefix as an IPv4 address.
Fix this by looking from the right for a colon. Also add unit tests. Fixes #7743
I'm moving this function to iputils as well (linking reczones with testrunner does no go smoothly).

There still remains an ambiguity I'm not really happy with: `string:port` can either be a full IPv6 address or a IPv6:port combo. If the string happens to parse as an IPv6 address the latter is returned. 
e.g.:
```
1::33:99 -> [1::33]:99
1:33::99 -> [1:33::99]
```
To avoid ambiguity always use square brackets for IPv6 addresss.
Since the original code allowed for naked IPv6 addresses we cannot just force that, alas. Or can we?

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/iputils.cc
+++ b/pdns/iputils.cc
@@ -436,3 +436,36 @@ bool isTCPSocketUsable(int sock)
 
   return false;
 }
+/* mission in life: parse three cases
+   1) 1.2.3.4
+   2) 1.2.3.4:5300
+   3) 2001::1
+   4) [2002::1]:53
+*/
+
+ComboAddress parseIPAndPort(const std::string& input, uint16_t port)
+{
+  if (input.find(':') == string::npos || input.empty()) { // common case
+    return ComboAddress(input, port);
+  }
+
+  pair<string,string> both;
+  try { // case 2
+    string::size_type cpos = input.rfind(':');
+    both.first = input.substr(0, cpos);
+    both.second = input.substr(cpos + 1);
+
+    uint16_t newport = static_cast<uint16_t>(pdns_stou(both.second));
+    return ComboAddress(both.first, newport);
+  }
+  catch(...) {
+  }
+
+  if (input[0] == '[') { // case 4
+    both = splitField(input.substr(1), ']');
+    return ComboAddress(both.first, both.second.empty() ? port : static_cast<uint16_t>(pdns_stou(both.second.substr(1))));
+  }
+
+  return ComboAddress(input, port); // case 3
+}
+
--- a/pdns/iputils.cc
+++ b/pdns/iputils.cc
@@ -437,15 +437,15 @@ bool isTCPSocketUsable(int sock)
   return false;
 }
 /* mission in life: parse four cases
-   1) 1.2.3.4
-   2) 1.2.3.4:5300
-   3) 2001::1
-   4) [2002::1]:53
+   1) [2002::1]:53
+   2) 1.2.3.4
+   3) 1.2.3.4:5300
+   4) 2001::1 no port allowed
 */
 
 ComboAddress parseIPAndPort(const std::string& input, uint16_t port)
 {
-  if (input[0] == '[') { // case 4
+  if (input[0] == '[') { // case 1
     auto both = splitField(input.substr(1), ']');
     return ComboAddress(both.first, both.second.empty() ? port : static_cast<uint16_t>(pdns_stou(both.second.substr(1))));
   }
@@ -460,9 +460,9 @@ ComboAddress parseIPAndPort(const std::string& input, uint16_t port)
     }
   }
   switch (count) {
-  case 0: // case 1
+  case 0: // case 2
     return ComboAddress(input, port);
-  case 1: { // case 2
+  case 1: { // case 3
     string::size_type cpos = input.rfind(':');
     pair<std::string,std::string> both;
     both.first = input.substr(0, cpos);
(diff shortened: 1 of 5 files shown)


pdns/iputils.cc
```diff
@@ -436,3 +436,36 @@ bool isTCPSocketUsable(int sock)
 
   return false;
 }
+/* mission in life: parse three cases
```
pieterlexis, line 439: yet I see 4 :)

omoerbeek: heh, that's a copy-paste form the original

omoerbeek:
Now tinking about an alternative approach: making the number of colons leading
starts with '[': [IPv6] with optional port
Else count colons:
0 -> naked IPv4
1 -> IPv4:port
2 or more: naked IPv6 without port

Habbie:
> To avoid ambiguity always use square brackets for IPv6 addresss.
> Since the original code allowed for naked IPv6 addresses we cannot just force that, alas. Or can we?

Bit late to the party, but: we can and we must. 'naked IPv6+port' is a recipe for disaster and confusion.

pdns/iputils.cc
```diff
@@ -436,3 +436,43 @@ bool isTCPSocketUsable(int sock)
...
+  for (char c : input) {
+    if (c == ':') {
+      count++;
+    }
+    if (count > 1) {
+      break;
+    }
+  }
+  switch (count) {
+  case 0: // case 1
```
Habbie, line 463: I would seriously consider renumbering the list in the comment at the top to make this less confusing :)

omoerbeek:
Rebased

Habbie (approved):

PR #9448, opened 2020-09-07
rec: We only want to do QName Minimization for the names in a forwarded domain.

### LINKED ISSUE #9438: Unnecessary requests to intermediate levels of the hard-coded domain.
- Program: Recursor
 - Issue type: Bug report

### Short description
Performing unnecessary operations in the process of processing a hard-coded domain leading to sending part of the requests to other servers.

### Environment
 - Operating system: Oracle Linux Server release 8.2
 - Software version: 4.4.0-beta1

### Steps to reproduce

1. Add the forwarding "baz.bar.com = ip_yuo_auth_internal_server" to your recursive server config in any way you want;
2. Enable request tracing: rec_control trace-regex '.*\.bar.com\.$'
3. Send a request to the recurser: "dig bar.baz.com @recursor_ip";
4. In the recursor log, we observe requests to all servers of intermediate levels of the bar.baz.com domain if they are not in the cache.

### Expected behaviour
The recurser must immediately send a request to the server, which is specified in the "forward-zones" directive or in the "forward-zones-file".

### Actual behaviour
The recurser queries the entire chain of authoritative servers of the parent zones (".", "Com", "baz") trying to find the current NS up to the subdomain "bar", but regardless of whether it succeeded or not, reaching the "bar" level it detects that the domain is hardcoded and further communication is performed with the server specified in the "forward-zones" or "forward-zones-file" directives.

### Other information
The behavior does not appear if you turn off qname minification.
You also need to take into account that in 4.4.0-beta1, the cache of ns records for subdomains that fall under the path of hard-coded domains finally started working. I would like this behavior to remain operational.

P.S.: Please edit the title of this issue if you find a more appropriate phrase.

omoerbeek:
E.g. if foo.bar.com is forwarded and the qname is x.foo.bar.com,
start the QM process with ancestor foo.bar.com, so the query is
directed to the forwarder.  But if the qname is baz.bar.com, we do
regular QM, starting with the regular ancestor.

Should fix #9438 without breaking having forward for dot.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/syncres.cc
+++ b/pdns/syncres.cc
@@ -725,17 +725,23 @@ int SyncRes::doResolve(const DNSName &qname, const QType &qtype, vector<DNSRecor
     for (int tries = 0; tries < 2 && bestns.empty(); ++tries) {
       bool flawedNSSet = false;
       set<GetBestNSAnswer> beenthereIgnored;
-      getBestNSFromCache(nsdomain, qtype, bestns, &flawedNSSet, depth, beenthereIgnored);
+      getBestNSFromCache(nsdomain, qtype, bestns, &flawedNSSet, depth, beenthereIgnored, forwarded ? fwdomain : g_rootdnsname);
+      if (forwarded) {
+        break;
+      }
     }
 
     if (bestns.size() == 0) {
-      // Something terrible is wrong
-      QLOG("Step1 No ancestor found return ServFail");
-      return RCode::ServFail;
+      if (!forwarded) {
+        // Something terrible is wrong
+        QLOG("Step1 No ancestor found return ServFail");
+        return RCode::ServFail;
+      }
+    } else {
+      QLOG("Step1 Ancestor from cache is " << bestns[0].d_name);
     }
-    QLOG("Step1 Ancestor from cache is " << bestns[0].d_name);
-
-    const DNSName& ancestor(!forwarded || bestns[0].d_name.isPartOf(fwdomain) ? bestns[0].d_name : fwdomain);
+    const DNSName& ancestor(!forwarded || (bestns.size() > 0 && bestns[0].d_name.isPartOf(fwdomain)) ?
+                            bestns[0].d_name : fwdomain);
     if (forwarded) {
       QLOG("Step1 Final Ancestor (using forwarding info) is " << ancestor);
     }
@@ -1110,7 +1116,7 @@ vector<ComboAddress> SyncRes::getAddrs(const DNSName &qname, unsigned int depth,
   return ret;
 }
 
-void SyncRes::getBestNSFromCache(const DNSName &qname, const QType& qtype, vector<DNSRecord>& bestns, bool* flawedNSSet, unsigned int depth, set<GetBestNSAnswer>& beenthere)
+void SyncRes::getBestNSFromCache(const DNSName &qname, const QType& qtype, vector<DNSRecord>& bestns, bool* flawedNSSet, unsigned int depth, set<GetBestNSAnswer>& beenthere, const DNSName& cutOffDomain)
 {
   string prefix;
   DNSName subdomain(qname);
@@ -1121,6 +1127,9 @@ void SyncRes::getBestNSFromCache(const DNSName &qname, const QType& qtype, vecto
   bestns.clear();
   bool brokeloop;
   do {
+    if (!subdomain.isPartOf(cutOffDomain)) {
+      break;
+    }
     brokeloop=false;
     LOG(prefix<<qname<<": Checking if we have NS in cache for '"<<subdomain<<"'"<<endl);
     vector<DNSRecord> ns;
--- a/pdns/syncres.cc
+++ b/pdns/syncres.cc
@@ -725,7 +725,7 @@ int SyncRes::doResolve(const DNSName &qname, const QType &qtype, vector<DNSRecor
     for (int tries = 0; tries < 2 && bestns.empty(); ++tries) {
       bool flawedNSSet = false;
       set<GetBestNSAnswer> beenthereIgnored;
-      getBestNSFromCache(nsdomain, qtype, bestns, &flawedNSSet, depth, beenthereIgnored, forwarded ? fwdomain : g_rootdnsname);
+      getBestNSFromCache(nsdomain, qtype, bestns, &flawedNSSet, depth, beenthereIgnored, boost::make_optional(forwarded, fwdomain));
       if (forwarded) {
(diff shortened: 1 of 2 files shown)


pdns/syncres.cc
```diff
@@ -714,22 +714,30 @@ int SyncRes::doResolve(const DNSName &qname, const QType &qtype, vector<DNSRecor
...
-    }
 
-    if (bestns.size() == 0) {
-      // Something terrible is wrong
-      QLOG("Step1 No ancestor found return ServFail");
-      return RCode::ServFail;
+    DNSName fwdomain(qname);
+    bool forwarded = getBestAuthZone(&fwdomain) != t_sstorage.domainmap->end();
+    if (forwarded) {
+      QLOG("Step1 qname is in a forwarded domain");
```
rgacogne, line 721: Couldn't that domain be one from a local auth zone as well?

rgacogne (commented):
I wonder if we should still check if we have a more specific ancestor in our cache even after looking up the forward/auth zones. If we have a query for `x.y.z.foo.bar.com.`, a forward for `foo.bar.com.` and a NS for `y.z.foo.bar.com.` in the cache, shouldn't we use the latest?

pdns/syncres.cc
omoerbeek: I do not think so, since for auth and recursive forwards QM is not used at all.

omoerbeek:
> I wonder if we should still check if we have a more specific ancestor in our cache even after looking up the forward/auth zones. If we have a query for `x.y.z.foo.bar.com.`, a forward for `foo.bar.com.` and a NS for `y.z.foo.bar.com.` in the cache, shouldn't we use the latest?

Yes, I think so. Will look into this.

pdns/syncres.cc
rgacogne: Right!

brlumen:
I think the latter. By the way, I wrote about this in #9438  :

> You also need to take into account that in 4.4.0-beta1, the cache of ns records for subdomains that fall under the path of hard-coded domains finally started working. I would like this behavior to remain operational.

omoerbeek:
> I think the latter. By the way, I wrote about this in #9438 :
> 
> > You also need to take into account that in 4.4.0-beta1, the cache of ns records for subdomains that fall under the path of hard-coded domains finally started working. I would like this behavior to remain operational.

Yes, I did test that, including checking to see if qname minimization was effective. I think my latest commit is just a more expilcit way of steering the algorithm which was otherwise done by the called functions to determine the nameserver to be used.

brlumen:
I tested it, everything works as intended. The only confusion is that we are checking for a valid NS in the cache below the hard-coded domain. This is not like a similar process when there is indeed a valid NS in the cache and no further verification of parent domains is performed.
  I think we are doing some extra work by checking if there is a valid NS in the cache for the parent zone. This also applies to query trace analysis. Stopping validation of the correct NS on the parents of the hard-coded zone would make more sense.

brlumen:
In the current version, this looks like:
> QM foo.bar.com.|A child=(empty): Step0 qname is in a forwarded domain bar.com
foo.bar.com: Checking if we have NS in cache for 'foo.bar.com'
foo.bar.com: no valid/useful NS in cache for 'foo.bar.com'
foo.bar.com: Checking if we have NS in cache for 'bar.com'
foo.bar.com: no valid/useful NS in cache for 'bar.com'
foo.bar.com: Checking if we have NS in cache for 'com'
foo.bar.com: no valid/useful NS in cache for 'com'
foo.bar.com: Checking if we have NS in cache for '.'
foo.bar.com: no valid/useful NS in cache for '.'
...

but more logical:
>QM foo.bar.com.|A child=(empty): Step0 qname is in a forwarded domain bar.com
foo.bar.com: Checking if we have NS in cache for 'foo.bar.com'
foo.bar.com: no valid/useful NS in cache for 'foo.bar.com' 
foo.bar.com authOrForwDomain: bar.com nsFromCacheDomain: . isPartof: 1
foo.bar.com: using forwarder as NS
foo.bar.com: initial validation status for foo.bar.com is Indeterminate
foo.bar.com: Cache consultations done, have 1 NS to contact
foo.bar.com: Domain has hardcoded nameserver
...

omoerbeek:
Thanks, that looks lile a valid observation/optimization. I'll look into it.

pdns/syncres.cc
```diff
@@ -1111,6 +1127,9 @@ void SyncRes::getBestNSFromCache(const DNSName &qname, const QType& qtype, vecto
   bestns.clear();
   bool brokeloop;
   do {
+    if (!subdomain.isPartOf(cutOffDomain)) {
```
rgacogne, line 1130: I did not test it but at first sight it looks like we will look the NS for `cutOffDomain` itself. We want that when it's `g_rootdnsname` but I'm not sure we do when it's `fwdomain`?

pdns/syncres.cc
```diff
@@ -719,17 +725,27 @@ int SyncRes::doResolve(const DNSName &qname, const QType &qtype, vector<DNSRecor
...
-      return RCode::ServFail;
+      if (!forwarded) {
+        // Something terrible is wrong
+        QLOG("Step1 No ancestor found return ServFail");
+        return RCode::ServFail;
+      }
+    } else {
+      QLOG("Step1 Ancestor from cache is " << bestns[0].d_name);
+    }
+    const DNSName& ancestor(!forwarded || (bestns.size() > 0 && bestns[0].d_name.isPartOf(fwdomain)) ?
```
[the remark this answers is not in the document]
omoerbeek: Yes, originally I kept the names to be the same as the the ines in the algorithm in the RFC, but this is indeed easier to read.

pdns/syncres.cc
omoerbeek: Correct, but I did not want to complicate matters further. It should not hurt (TM).

rgacogne: I wonder if we can somehow end up with `NS` records in the cache for a forwarded domain? We should refuse them from other authoritative servers than the ones we forward to, but what happens if these send us a `NS` record that does not match our forward?

omoerbeek: OK, I made the code more robust against that by using boost::optional and only doing cutoff when we are passed a cutoff point. This has the nice consequence that the original behaviour (with no cutoff) is not modified at all.

rgacogne (approved):

PR #9450, opened 2020-09-08
Support for backup servers

1848:
Not ready for merging, POC only for now.

This adds support for backup servers in newServer().
Works only for the roundrobin policy for now.

When a new server is registered using newServer() and backup is set to true, the server will not be used in the RR policy unless all other non-backup servers are down.

I will add support to all other policies if you think this PR is useful in general.

Use case:
Multiple equal downstream servers and some backup servers which should only be used if all primary ds servers are down.

Dont mind the debug code, only used for testing.

### Checklist
I have:
- [yes] read the [CONTRIBUTING.md](https://github.com/PowerDNS/pdns/blob/master/CONTRIBUTING.md) document
- [yes] compiled this code
- [yes] tested this code
- [not yet] included documentation (including possible behaviour changes)
- [not needed] documented the code
- [not yet] added or modified regression test(s)
- [not yet] added or modified unit test(s)

### DIFF
--- a/pdns/dnsdist-lua.cc
+++ b/pdns/dnsdist-lua.cc
@@ -378,6 +378,10 @@ static void setupLuaConfig(LuaContext& luaCtx, bool client, bool configCheck)
         ret->checkInterval=static_cast<unsigned int>(std::stoul(boost::get<string>(vars["checkInterval"])));
       }
 
+      if(vars.count("backup")) {
+        ret->is_backup=boost::get<bool>(vars["backup"]);
+      }
+
       if(vars.count("tcpConnectTimeout")) {
         ret->tcpConnectTimeout=std::stoi(boost::get<string>(vars["tcpConnectTimeout"]));
       }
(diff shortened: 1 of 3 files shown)


rgacogne:
Thank you for this pull request! I'm open to the idea of implementing backup servers, although it feels a bit weird to have to implement that in every load-balancing policy. I'm not sure that can be helped with our current design.
Note that it can roughly be implemented via the `PoolAvailableRule` rule, placing backup servers in a different pool and routing the query to that pool if the "main" one does not have any server available. It's a bit more cumbersome because you have to declare a different pool, though.

1848:
First of all sorry for my late reply (too busy the last days).
I agree that it would be not the best idea to implement this for every LB policy.


> I'm not sure that can be helped with our current design.

I see, will close this PR for now. PoolAvailableRule seems to be good enough to skip the trouble implementing this properly.

Thanks for your time and opinion.

PR #9464, opened 2020-09-14
auth: cache list of all domains

zeha:
### Short description

Add a cache of all domains, avoiding backend lookups for domains that do not exist, and for non-existing subdomains.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/modules/lmdbbackend/lmdbbackend.cc
+++ b/modules/lmdbbackend/lmdbbackend.cc
@@ -969,12 +969,12 @@ bool LMDBBackend::createDomain(const DNSName& domain, const DomainInfo::DomainKi
     di.account = account;
 
     txn.put(di);
-    txn.commit();
-  }
 
-  if (zoneId != nullptr) {
-    auto txn = d_tdomains->getROTransaction();
-    *zoneId = txn.get<0>(domain, di);
+    if (zoneId != nullptr) {
+      *zoneId = txn.get<0>(domain, di);
+    }
+
+    txn.commit();
   }
 
   return true;
--- a/modules/lmdbbackend/lmdbbackend.cc
+++ b/modules/lmdbbackend/lmdbbackend.cc
@@ -953,22 +953,29 @@ bool LMDBBackend::setMasters(const DNSName& domain, const vector<ComboAddress>& 
   });
 }
 
-bool LMDBBackend::createDomain(const DNSName& domain, const DomainInfo::DomainKind kind, const vector<ComboAddress>& masters, const string& account)
+bool LMDBBackend::createDomain(const DNSName& domain, const DomainInfo::DomainKind kind, const vector<ComboAddress>& masters, const string& account, int* zoneId)
 {
   DomainInfo di;
 
-  auto txn = d_tdomains->getRWTransaction();
-  if (txn.get<0>(domain, di)) {
-    throw DBException("Domain '" + domain.toLogString() + "' exists already");
+  {
+    auto txn = d_tdomains->getRWTransaction();
+    if (txn.get<0>(domain, di)) {
+      throw DBException("Domain '" + domain.toLogString() + "' exists already");
+    }
+
+    di.zone = domain;
+    di.kind = kind;
+    di.masters = masters;
+    di.account = account;
+
+    txn.put(di);
+    txn.commit();
   }
 
-  di.zone = domain;
-  di.kind = kind;
-  di.masters = masters;
-  di.account = account;
-
-  txn.put(di);
-  txn.commit();
+  if (zoneId != nullptr) {
+    auto txn = d_tdomains->getROTransaction();
+    *zoneId = txn.get<0>(domain, di);
+  }
 
   return true;
 }
--- a/pdns/auth-domaincache.cc
+++ b/pdns/auth-domaincache.cc
@@ -160,7 +160,9 @@ void AuthDomainCache::setReplacePending()
   if (!d_ttl)
     return;
 
-  WriteLock globalLock(d_mut);
-  d_replacePending = true;
-  d_pendingAdds.clear();
+  {
+    WriteLock globalLock(d_mut);
+    d_replacePending = true;
+    d_pendingAdds.clear();
+  }
 }
--- a/pdns/auth-domaincache.cc
+++ b/pdns/auth-domaincache.cc
@@ -30,15 +30,16 @@
 #include "cachecleaner.hh"
 extern StatBag S;
 
-AuthDomainCache::AuthDomainCache(size_t mapsCount): d_maps(mapsCount)
+AuthDomainCache::AuthDomainCache(size_t mapsCount) :
+  d_maps(mapsCount)
 {
   S.declare("domain-cache-hit", "Number of hits on the domain cache");
   S.declare("domain-cache-miss", "Number of misses on the domain cache");
   S.declare("domain-cache-size", "Number of entries in the domain cache", StatType::gauge);
 
-  d_statnumhit=S.getPointer("domain-cache-hit");
-  d_statnummiss=S.getPointer("domain-cache-miss");
-  d_statnumentries=S.getPointer("domain-cache-size");
+  d_statnumhit = S.getPointer("domain-cache-hit");
+  d_statnummiss = S.getPointer("domain-cache-miss");
+  d_statnumentries = S.getPointer("domain-cache-size");
 
   d_ttl = 0;
 }
@@ -47,16 +48,16 @@ AuthDomainCache::~AuthDomainCache()
 {
   try {
     vector<WriteLock> locks;
-    for(auto& mc : d_maps) {
+    for (auto& mc : d_maps) {
       locks.push_back(WriteLock(mc.d_mut));
     }
     locks.clear();
   }
-  catch(...) {
+  catch (...) {
   }
 }
 
-bool AuthDomainCache::getEntry(const DNSName &domain, int& zoneId)
+bool AuthDomainCache::getEntry(const DNSName& domain, int& zoneId)
 {
   auto& mc = getMap(domain);
   bool found = false;
@@ -71,7 +72,8 @@ bool AuthDomainCache::getEntry(const DNSName &domain, int& zoneId)
 
   if (found) {
     (*d_statnumhit)++;
-  } else {
+  }
+  else {
     (*d_statnummiss)++;
   }
   return found;
@@ -87,16 +89,16 @@ void AuthDomainCache::clear()
   purgeLockedCollectionsVector(d_maps);
 }
 
-void AuthDomainCache::replace(const vector<tuple<DNSName, int>> &domain_indices)
+void AuthDomainCache::replace(const vector<tuple<DNSName, int>>& domain_indices)
 {
-  if(!d_ttl)
+  if (!d_ttl)
     return;
 
   size_t count = domain_indices.size();
   vector<MapCombo> newMaps(d_maps.size());
 
   // build new maps
-  for(const tuple<DNSName, int>& tup: domain_indices) {
+  for (const tuple<DNSName, int>& tup : domain_indices) {
     const DNSName& domain = tup.get<0>();
     CacheValue val;
     val.zoneId = tup.get<1>();
@@ -104,8 +106,7 @@ void AuthDomainCache::replace(const vector<tuple<DNSName, int>> &domain_indices)
     mc.d_map.emplace(domain, val);
   }
 
-  for(size_t mapIndex = 0; mapIndex < d_maps.size(); mapIndex++)
-  {
+  for (size_t mapIndex = 0; mapIndex < d_maps.size(); mapIndex++) {
     auto& mc = d_maps[mapIndex];
     WriteLock l(mc.d_mut);
     mc.d_map = std::move(newMaps[mapIndex].d_map);
@@ -116,7 +117,7 @@ void AuthDomainCache::replace(const vector<tuple<DNSName, int>> &domain_indices)
 
 void AuthDomainCache::add(const DNSName& domain, const int zoneId)
 {
-  if(!d_ttl)
+  if (!d_ttl)
     return;
 
   CacheValue val;
--- a/pdns/auth-domaincache.cc
+++ b/pdns/auth-domaincache.cc
@@ -33,8 +33,8 @@ extern StatBag S;
 AuthDomainCache::AuthDomainCache(size_t mapsCount) :
   d_maps(mapsCount)
 {
-  S.declare("domain-cache-hit", "Number of hits on the domain cache");
-  S.declare("domain-cache-miss", "Number of misses on the domain cache");
+  S.declare("domain-cache-hit", "Number of domain cache hits");
+  S.declare("domain-cache-miss", "Number of domain cache misses");
   S.declare("domain-cache-size", "Number of entries in the domain cache", StatType::gauge);
 
   d_statnumhit = S.getPointer("domain-cache-hit");
@@ -109,7 +109,7 @@ void AuthDomainCache::replace(const vector<tuple<DNSName, int>>& domain_indices)
   {
     WriteLock globalLock(d_mut);
     if (d_replacePending) {
-      // add/replace all domains created while data collection for replace() was already running.
+      // add/replace all domains created while data collection for replace() was already in progress.
       for (const tuple<DNSName, int>& tup : d_pendingAdds) {
         const DNSName& domain = tup.get<0>();
         CacheValue val;
--- a/pdns/auth-domaincache.cc
+++ b/pdns/auth-domaincache.cc
@@ -106,10 +106,27 @@ void AuthDomainCache::replace(const vector<tuple<DNSName, int>>& domain_indices)
     mc.d_map.emplace(domain, val);
   }
 
-  for (size_t mapIndex = 0; mapIndex < d_maps.size(); mapIndex++) {
-    auto& mc = d_maps[mapIndex];
-    WriteLock l(mc.d_mut);
-    mc.d_map = std::move(newMaps[mapIndex].d_map);
+  {
+    WriteLock globalLock(d_mut);
+    if (d_replacePending) {
+      // add/replace all domains created while data collection for replace() was already running.
+      for (const tuple<DNSName, int>& tup : d_pendingAdds) {
+        const DNSName& domain = tup.get<0>();
+        CacheValue val;
+        val.zoneId = tup.get<1>();
+        auto& mc = newMaps[getMapIndex(domain)];
+        mc.d_map[domain] = val;
+      }
+    }
+
+    for (size_t mapIndex = 0; mapIndex < d_maps.size(); mapIndex++) {
+      auto& mc = d_maps[mapIndex];
+      WriteLock mcLock(mc.d_mut);
+      mc.d_map = std::move(newMaps[mapIndex].d_map);
+    }
+
+    d_pendingAdds.clear();
+    d_replacePending = false;
   }
 
   d_statnumentries->store(count);
@@ -120,13 +137,30 @@ void AuthDomainCache::add(const DNSName& domain, const int zoneId)
   if (!d_ttl)
     return;
 
+  {
+    WriteLock globalLock(d_mut);
+    if (d_replacePending) {
+      d_pendingAdds.push_back({domain, zoneId});
+    }
+  }
+
   CacheValue val;
   val.zoneId = zoneId;
 
   int mapIndex = getMapIndex(domain);
   {
     auto& mc = d_maps[mapIndex];
-    WriteLock l(mc.d_mut);
+    WriteLock mcLock(mc.d_mut);
     mc.d_map.emplace(domain, val);
   }
+}
+
+void AuthDomainCache::setReplacePending()
+{
+  if (!d_ttl)
+    return;
+
+  WriteLock globalLock(d_mut);
+  d_replacePending = true;
+  d_pendingAdds.clear();
 }
--- a/pdns/auth-domaincache.cc
+++ b/pdns/auth-domaincache.cc
@@ -1,0 +1,131 @@
+/*
+ * This file is part of PowerDNS or dnsdist.
+ * Copyright -- PowerDNS.COM B.V. and its contributors
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of version 2 of the GNU General Public License as
+ * published by the Free Software Foundation.
+ *
+ * In addition, for the avoidance of any doubt, permission is granted to
+ * link this program with OpenSSL and to (re)distribute the binaries
+ * produced as the result of such linking.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+#ifdef HAVE_CONFIG_H
+#include "config.h"
+#endif
+
+#include "auth-domaincache.hh"
+#include "logger.hh"
+#include "statbag.hh"
+#include "arguments.hh"
+#include "cachecleaner.hh"
+extern StatBag S;
+
+AuthDomainCache::AuthDomainCache(size_t mapsCount): d_maps(mapsCount)
+{
+  S.declare("domain-cache-hit", "Number of hits on the domain cache");
+  S.declare("domain-cache-miss", "Number of misses on the domain cache");
+  S.declare("domain-cache-size", "Number of entries in the domain cache", StatType::gauge);
+
+  d_statnumhit=S.getPointer("domain-cache-hit");
+  d_statnummiss=S.getPointer("domain-cache-miss");
+  d_statnumentries=S.getPointer("domain-cache-size");
+
+  d_ttl = 0;
+}
+
+AuthDomainCache::~AuthDomainCache()
+{
+  try {
+    vector<WriteLock> locks;
+    for(auto& mc : d_maps) {
+      locks.push_back(WriteLock(mc.d_mut));
+    }
+    locks.clear();
+  }
+  catch(...) {
+  }
+}
+
+bool AuthDomainCache::getEntry(const DNSName &domain, int& zoneId)
+{
+  auto& mc = getMap(domain);
+  bool found = false;
+  {
+    ReadLock rl(mc.d_mut);
+    auto iter = mc.d_map.find(domain);
+    if (iter != mc.d_map.end()) {
+      found = true;
+      zoneId = iter->second.zoneId;
+    }
+  }
+
+  if (found) {
+    (*d_statnumhit)++;
+  } else {
+    (*d_statnummiss)++;
+  }
+  return found;
+}
+
+bool AuthDomainCache::isEnabled() const
+{
+  return d_ttl > 0;
+}
+
+void AuthDomainCache::clear()
+{
+  purgeLockedCollectionsVector(d_maps);
+}
+
+void AuthDomainCache::replace(const vector<tuple<DNSName, int>> &domain_indices)
+{
+  if(!d_ttl)
+    return;
+
+  size_t count = domain_indices.size();
+  vector<MapCombo> newMaps(d_maps.size());
+
+  // build new maps
+  for(const tuple<DNSName, int>& tup: domain_indices) {
+    const DNSName& domain = tup.get<0>();
+    CacheValue val;
+    val.zoneId = tup.get<1>();
+    auto& mc = newMaps[getMapIndex(domain)];
+    mc.d_map.emplace(domain, val);
+  }
+
+  for(size_t mapIndex = 0; mapIndex < d_maps.size(); mapIndex++)
+  {
+    auto& mc = d_maps[mapIndex];
+    WriteLock l(mc.d_mut);
+    mc.d_map = newMaps[mapIndex].d_map;
+  }
+
+  d_statnumentries->store(count);
+}
+
+void AuthDomainCache::add(const DNSName& domain, const int zoneId)
+{
+  if(!d_ttl)
+    return;
+
+  CacheValue val;
+  val.zoneId = zoneId;
+
+  int mapIndex = getMapIndex(domain);
+  {
+    auto& mc = d_maps[mapIndex];
+    WriteLock l(mc.d_mut);
+    mc.d_map.emplace(domain, val);
+  }
+}
--- a/pdns/auth-domaincache.cc
+++ b/pdns/auth-domaincache.cc
@@ -108,7 +108,7 @@ void AuthDomainCache::replace(const vector<tuple<DNSName, int>> &domain_indices)
   {
     auto& mc = d_maps[mapIndex];
     WriteLock l(mc.d_mut);
-    mc.d_map = newMaps[mapIndex].d_map;
+    mc.d_map = std::move(newMaps[mapIndex].d_map);
   }
 
   d_statnumentries->store(count);
--- a/pdns/auth-domaincache.hh
+++ b/pdns/auth-domaincache.hh
@@ -21,13 +21,8 @@
  */
 #pragma once
 #include <string>
-#include <map>
-#include "dns.hh"
-
 #include <unordered_map>
-
-#include "dns.hh"
-#include "dnspacket.hh"
+#include "dnsname.hh"
 #include "lock.hh"
 
 class AuthDomainCache : public boost::noncopyable
@@ -91,3 +86,5 @@ private:
 
   time_t d_ttl;
 };
+
+extern AuthDomainCache g_domainCache;
--- a/pdns/auth-domaincache.hh
+++ b/pdns/auth-domaincache.hh
@@ -35,7 +35,7 @@ public:
 
   void replace(const vector<tuple<DNSName, int>>& domains);
   void add(const DNSName& domain, const int zoneId);
-  void setReplacePending(); //!< call this when data collection for subsequent replace() call starts.
+  void setReplacePending(); //!< call this when data collection for the subsequent replace() call starts.
 
   bool getEntry(const DNSName& domain, int& zoneId);
 
--- a/pdns/auth-domaincache.hh
+++ b/pdns/auth-domaincache.hh
@@ -79,7 +79,6 @@ private:
     return d_maps[getMapIndex(qname)];
   }
 
-  AtomicCounter d_ops{0};
   AtomicCounter *d_statnumhit;
   AtomicCounter *d_statnummiss;
   AtomicCounter *d_statnumentries;
--- a/pdns/auth-domaincache.hh
+++ b/pdns/auth-domaincache.hh
@@ -28,21 +28,23 @@
 class AuthDomainCache : public boost::noncopyable
 {
 public:
-  AuthDomainCache(size_t mapsCount=1024);
+  AuthDomainCache(size_t mapsCount = 1024);
   ~AuthDomainCache();
 
-  void replace(const vector<tuple<DNSName, int>> &domains);
+  void replace(const vector<tuple<DNSName, int>>& domains);
   void add(const DNSName& domain, const int zoneId);
 
-  bool getEntry(const DNSName &domain, int &zoneId);
+  bool getEntry(const DNSName& domain, int& zoneId);
 
   size_t size() { return *d_statnumentries; } //!< number of entries in the cache
 
-  uint32_t getTTL() const {
+  uint32_t getTTL() const
+  {
     return d_ttl;
   }
 
-  void setTTL(uint32_t ttl) {
+  void setTTL(uint32_t ttl)
+  {
     d_ttl = ttl;
   }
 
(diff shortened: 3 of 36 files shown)


pdns/auth-domaincache.hh
```diff
@@ -0,0 +1,111 @@
...
+
+  struct HashTag{};
+  struct NameTag{};
+  struct SequencedTag{};
+  typedef multi_index_container<
+    CacheEntry,
+    indexed_by <
+      hashed_unique<tag<HashTag>, composite_key<CacheEntry,
+                                                         member<CacheEntry,DNSName,&CacheEntry::domain> > > ,
+      ordered_non_unique<tag<NameTag>, member<CacheEntry,DNSName,&CacheEntry::domain>, CanonDNSNameCompare >,
```
rgacogne, line 70: I'm not sure we actually need this one? At a quick glance it looks like this cache could be an `unordered_map` rather than a `multi-index` :-)

zeha: Right. Added as a todo.

zeha:
On a dnsdist-fronted install, I can see a drop in "query to db amplification %" from 190 to 110, while keeping latency ~unchanged.

Memory usage is ~ 1MB/100k domains.

zeha:
This temporarily adds a getSOA call inside UeberBackend, but `getSOA` got replaced by a `lookup` call. I might inline this later into getAuth.

zeha:
I don't understand why the gmysql tests fail:

https://circleci.com/api/v1.1/project/github/PowerDNS/pdns/55104/output/115/0?file=true&allocation-id=5f9ab1b9d9375c3ff53e5c3c-0-build%2F7F358AB9

```
Oct 29 12:14:51 gmysql Connection failed: Unable to connect to database: ERROR 2002 (HY000): Can't connect to MySQL server on '127.0.0.1' (99)
Oct 29 12:14:51 Caught an exception instantiating a backend: Unable to launch gmysql connection: Unable to connect to database: ERROR 2002 (HY000): Can't connect to MySQL server on '127.0.0.1' (99)
Oct 29 12:14:51 Cleaning up
```

zeha:
getAuth needs to be made nicer, but everything else is I think good to go now.

pieterlexis:
@zeha what's the status of this?

zeha:
> @zeha what's the status of this?

It works and its in production at a customer. Needs a rebase though, apparently.

Habbie:
I milestoned this 4.5.0-alpha0 which means 'merge as soon as doing that merge will not put it into 4.4' so I'd also like this for 4.5. If you can rebase, I'll try to do a review soon.

zeha:
Rebased.

pdns/auth-domaincache.hh
```diff
@@ -0,0 +1,93 @@
...
+    return d_maps[getMapIndex(qname)];
+  }
+
+  AtomicCounter d_ops{0};
+  AtomicCounter *d_statnumhit;
+  AtomicCounter *d_statnummiss;
+  AtomicCounter *d_statnumentries;
+
+  time_t d_ttl;
+};
```
rgacogne, line 93: How about adding `extern AuthDomainCache g_domainCache;` in that file, so we don't need it anywhere else?

pdns/auth-domaincache.cc
```diff
@@ -0,0 +1,131 @@
...
+#include "config.h"
+#endif
+
+#include "auth-domaincache.hh"
+#include "logger.hh"
+#include "statbag.hh"
+#include "arguments.hh"
+#include "cachecleaner.hh"
+extern StatBag S;
+
```
rgacogne, line 32: `AuthDomainCache g_domainCache`; ?

pdns/auth-domaincache.cc
```diff
@@ -0,0 +1,131 @@
...
+    val.zoneId = tup.get<1>();
+    auto& mc = newMaps[getMapIndex(domain)];
+    mc.d_map.emplace(domain, val);
+  }
+
+  for(size_t mapIndex = 0; mapIndex < d_maps.size(); mapIndex++)
+  {
+    auto& mc = d_maps[mapIndex];
+    WriteLock l(mc.d_mut);
+    mc.d_map = newMaps[mapIndex].d_map;
```
rgacogne, line 111:
```suggestion
    mc.d_map = std::move(newMaps[mapIndex].d_map);
```

pdns/ueberbackend.cc
```diff
@@ -273,16 +281,35 @@ void UeberBackend::reload()
...
+  }
+
+  vector<tuple<DNSName, int>> domain_indices;
+
+  for (vector<DNSBackend*>::iterator i = backends.begin(); i != backends.end(); ++i )
+  {
+    vector<DomainInfo> domains;
+    (*i)->getAllDomains(&domains, false);
+    for(const auto& di: domains) {
+      domain_indices.push_back({di.zone, (int)di.id});  // this cast should not be necessary
```
rgacogne, line 296: If we make `di` non-const we can move `di.zone` here.

pdns/test-ueberbackend_cc.cc
```diff
@@ -362,8 +369,11 @@ static void testWithoutThenWithCache(std::function<void(UeberBackend& ub)> func)
     ::arg().set("query-cache-ttl")="20";
     ::arg().set("negquery-cache-ttl")="60";
     QC.cleanup();
+    ::arg().set("domain-cache-ttl")="60";
```
rgacogne, line 372: I would not mind adding a separate test here. We currently have a first test with no cache at all and a second one with the positive and negative query caches. This PR adds the domain cache to the second test, I'd rather keep the second test without the domain cache and have a third one with the positive, negative and domain caches.

pdns/auth-domaincache.hh
```diff
@@ -0,0 +1,93 @@
...
+  {
+    return d_maps[getMapIndex(qname)];
+  }
+
+  AtomicCounter d_ops{0};
+  AtomicCounter *d_statnumhit;
+  AtomicCounter *d_statnummiss;
+  AtomicCounter *d_statnumentries;
+
+  time_t d_ttl;
```
rgacogne, line 92:
```suggestion
  time_t d_ttl{0};
```

pdns/auth-domaincache.hh
```diff
@@ -0,0 +1,93 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+#pragma once
+#include <string>
+#include <map>
```
rgacogne, line 24: Is this still needed?

pdns/auth-domaincache.hh
```diff
@@ -0,0 +1,93 @@
...
+#include "dns.hh"
+
+#include <unordered_map>
+
+#include "dns.hh"
+#include "dnspacket.hh"
```
rgacogne, line 30: That doesn't seem needed?

rgacogne (approved):
A few comments but the concept and code look good!

I wonder if it would make sense to remove entries from the cache in `deleteDomain()`?

pdns/auth-domaincache.hh
zeha: done!

pdns/auth-domaincache.cc
zeha: That unfortunately crashes on startup inside StatBag, due to globals init order. :(

pdns/auth-domaincache.cc
zeha: done

pdns/ueberbackend.cc
zeha: done

pdns/auth-domaincache.cc
rgacogne: Right, I'm really not a big fan of the `StatBag` because by design it's creeping up everywhere.

zeha:
@zeha look at the now failed tests, please

pdns/auth-domaincache.cc
zeha: Yeah :-/ I think it could be time for moving a few of those globals around, but I don't know who should be driving this.

And it's painful for backports / security patches.

klaus-nicat:
FYI: Running stable since 2 weeks in production on all our 52 nodes (40kq/s). Before it was running stable for 3 months on 3 nodes.

rgacogne:
Crazy idea, did you ever consider the possibility to add an optional cache for the whole content of selected zones? I have not thought this through, but something like:
- take a list of zones to load in memory at startup (perhaps a smart list, with the ability to include sub-zones or not and to exclude some zones) ;
- load all records for these zones in a in-memory cache ;
- check now and then (every 60s or so?) if the SOA has been updated in the backend, and if so add the list to a queue of zones to reload ;
- when an incoming query comes in, check that cache first (likely with a read-write lock, giving up and falling back to normal processing if we can't get the lock right away).

Of course that would only work for backends that supports listing the content of a zone, and only if any edit is guaranteed to update the SOA, but as long as you have to allow-list the cached zones I guess that's fine. It would provide nice performance while keeping native DB replication, and could be enabled only when a zone is under attack, for example.

Habbie:
> Crazy idea, did you ever consider the possibility to add an optional cache for the whole content of selected zones?

Here's a variant I proposed to @zeha earlier, also not entirely thought through:
* have a pdns_server instance that replicates however you normally replicate
* have a second pdns_server instance, using LMDB, that serves your data
* do 'something clever that smells a lot like `pdnsutil b2b-migrate`' in an outside process

Habbie:
The ubuntu-bionic build failure looks real and related to this PR:
```
test-ueberbackend_cc.cc: In member function 'void test_ueberbackend_cc::test_simple::test_method()':
test-ueberbackend_cc.cc:537:5: error: 'testWithoutThenWithCache' was not declared in this scope
     testWithoutThenWithCache(testFunction);
     ^~~~~~~~~~~~~~~~~~~~~~~~
test-ueberbackend_cc.cc:537:5: note: suggested alternative: 'testWithoutThenWithAuthCache'
     testWithoutThenWithCache(testFunction);
     ^~~~~~~~~~~~~~~~~~~~~~~~
     testWithoutThenWithAuthCache
test-ueberbackend_cc.cc: In member function 'void test_ueberbackend_cc::test_multi_backends_separate_zones::test_method()':
test-ueberbackend_cc.cc:683:5: error: 'testWithoutThenWithCache' was not declared in this scope
     testWithoutThenWithCache(testFunction);
     ^~~~~~~~~~~~~~~~~~~~~~~~
test-ueberbackend_cc.cc:683:5: note: suggested alternative: 'testWithoutThenWithAuthCache'
     testWithoutThenWithCache(testFunction);
     ^~~~~~~~~~~~~~~~~~~~~~~~
     testWithoutThenWithAuthCache
test-ueberbackend_cc.cc: In member function 'void test_ueberbackend_cc::test_multi_backends_overlay::test_method()':
test-ueberbackend_cc.cc:808:5: error: 'testWithoutThenWithCache' was not declared in this scope
     testWithoutThenWithCache(testFunction);
     ^~~~~~~~~~~~~~~~~~~~~~~~
test-ueberbackend_cc.cc:808:5: note: suggested alternative: 'testWithoutThenWithAuthCache'
     testWithoutThenWithCache(testFunction);
     ^~~~~~~~~~~~~~~~~~~~~~~~
     testWithoutThenWithAuthCache
test-ueberbackend_cc.cc: In member function 'void test_ueberbackend_cc::test_multi_backends_overlay_name::test_method()':
test-ueberbackend_cc.cc:930:5: error: 'testWithoutThenWithCache' was not declared in this scope
     testWithoutThenWithCache(testFunction);
     ^~~~~~~~~~~~~~~~~~~~~~~~
test-ueberbackend_cc.cc:930:5: note: suggested alternative: 'testWithoutThenWithAuthCache'
     testWithoutThenWithCache(testFunction);
     ^~~~~~~~~~~~~~~~~~~~~~~~
     testWithoutThenWithAuthCache
test-ueberbackend_cc.cc: In member function 'void test_ueberbackend_cc::test_child_zone::test_method()':
test-ueberbackend_cc.cc:1007:5: error: 'testWithoutThenWithCache' was not declared in this scope
     testWithoutThenWithCache(testFunction);
     ^~~~~~~~~~~~~~~~~~~~~~~~
test-ueberbackend_cc.cc:1007:5: note: suggested alternative: 'testWithoutThenWithAuthCache'
     testWithoutThenWithCache(testFunction);
     ^~~~~~~~~~~~~~~~~~~~~~~~
     testWithoutThenWithAuthCache
test-ueberbackend_cc.cc: In member function 'void test_ueberbackend_cc::test_multi_backends_best_soa::test_method()':
test-ueberbackend_cc.cc:1059:5: error: 'testWithoutThenWithCache' was not declared in this scope
     testWithoutThenWithCache(testFunction);
     ^~~~~~~~~~~~~~~~~~~~~~~~
test-ueberbackend_cc.cc:1059:5: note: suggested alternative: 'testWithoutThenWithAuthCache'
     testWithoutThenWithCache(testFunction);
     ^~~~~~~~~~~~~~~~~~~~~~~~
     testWithoutThenWithAuthCache
test-ueberbackend_cc.cc: At global scope:
test-ueberbackend_cc.cc:383:13: warning: 'void testWithoutThenWithDomainCache(std::function<void(UeberBackend&)>)' defined but not used [-Wunused-function]
 static void testWithoutThenWithDomainCache(std::function<void(UeberBackend& ub)> func)
             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
test-ueberbackend_cc.cc:349:13: warning: 'void testWithoutThenWithAuthCache(std::function<void(UeberBackend&)>)' defined but not used [-Wunused-function]
 static void testWithoutThenWithAuthCache(std::function<void(UeberBackend& ub)> func)
             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
make[4]: *** [test-ueberbackend_cc.o] Error 1
```

Habbie:
Also, there are a few conflicts, presumably due to formatting of `modules/*`. Applying the formatting to your branch might be enough to fix those conflicts, but I did not try this.

klaus-nicat:
> Here's a variant I proposed to @zeha earlier, also not entirely thought through:
>     * have a pdns_server instance that replicates however you normally replicate
>     * have a second pdns_server instance, using LMDB, that serves your data
>     * do 'something clever that smells a lot like `pdnsutil b2b-migrate`' in an outside process

I often have thought about such hacks outside of PDNS. But I would use Knot+XDP and probably standard NOTIFY+AXFR. I do not know if this is reliable in sync with 3 mio zones and it still required dynamic additon/deletion of zones (at least until catalog zones + SERIAL are implemented). And for sure it would double RAM requirements - as the Powerdns-DB and the Knot-or-LMDB zones would need to be in RAM. So then we could just cache whole zones in PDNS+DB, I guess.

klaus-nicat:
> Crazy idea, did you ever consider the possibility to add an optional cache for the whole content of selected zones? I have not thought this through, but something like:
>     * take a list of zones to load in memory at startup (perhaps a smart list, with the ability to include sub-zones or not and to exclude some zones) ;
>     * load all records for these zones in a in-memory cache ;
>     * check now and then (every 60s or so?) if the SOA has been updated in the backend, and if so add the list to a queue of zones to reload ;
>     * when an incoming query comes in, check that cache first (likely with a read-write lock, giving up and falling back to normal processing if we can't get the lock right away).
> 
> Of course that would only work for backends that supports listing the content of a zone, and only if any edit is guaranteed to update the SOA, but as long as you have to allow-list the cached zones I guess that's fine. It would provide nice performance while keeping native DB replication, and could be enabled only when a zone is under attack, for example.

I would immediatly buy your solution if those "hot" zones are detected automatically and automatically added. Either for example using a threshold (ie 1000q/s per zone) or for example keeping the hottest 1000 zones in the "full-zone-cache".

But the "cache list of all domains" would be needed anyways. For example there is a random subdomain attack against the nic.at Zone using a 1.2.3.4.5.6.$random.nic.at pattern - then PDNS should know the serving zone without performing 8 getSOA lookups.

pdns/common_startup.cc
```diff
@@ -673,12 +675,36 @@ void mainthread()
...
-    try {
-      doSecPoll(false);
+    const uint32_t slept = g_domainCache.getTTL() == 0 ? secpollInterval : std::min(secpollInterval, g_domainCache.getTTL());
+    sleep(slept);  // if any signals arrive, we might run more often than expected.
+
+    domainCacheUpdateSince += slept;
+    if (domainCacheUpdateSince >= g_domainCache.getTTL()) {
+      try {
+        UeberBackend B;
+        B.updateDomainCache();
```
mind04, line 689: There is a race here. Any zones added or removed while this is running are not taken in account. As a result of this a new zone might show up briefly and then disappear for the duration of the cache ttl.
Running with large ttls will be problematic because of this.  For api only systems this will force you to more frequent domain list updates than necessary.

Habbie:
>  (at least until catalog zones + SERIAL are implemented)

Indeed, once we have that, you could reduce the load on your secondary to just pdns+LMDB.

paddg:
> Crazy idea ...
I have also often thought why not combine the advantages of BIND and DB backend. A backend that works like the BIND backend, only that the zones do not come from a file, but from the database. So serve everything only from memory like the BIND backend does.

rgacogne:
> I would immediatly buy your solution if those "hot" zones are detected automatically and automatically added. Either for example using a threshold (ie 1000q/s per zone) or for example keeping the hottest 1000 zones in the "full-zone-cache".

That's a nice idea :+1:

> But the "cache list of all domains" would be needed anyways. For example there is a random subdomain attack against the nic.at Zone using a 1.2.3.4.5.6.$random.nic.at pattern - then PDNS should know the serving zone without performing 8 getSOA lookups.

Indeed, I was not suggesting to replace this PR but merely to build on it once it is merged.

pdns/common_startup.cc
rgacogne: Nice catch! I guess one option would be to set a flag during the refresh so that additions happening during that time are also added to a queue, which would be cleared at the end of the update, or something like that.

zeha:
Rebased + reformatted, and the test failure is fixed. Still need to look at the add/replace race.

pdns/common_startup.cc
zeha: I've added the flag/queue code now.

zeha:
Rebased again, and added a queue for adding domains while data for replace() is being collected.

zeha:
@klaus-nicat do you want to give this version a try?

zeha:
I've got positive feedback about this MR. Review/Merge would be greatly appreciated!

pdns/auth-domaincache.cc
```diff
@@ -0,0 +1,166 @@
...
+#include "statbag.hh"
+#include "arguments.hh"
+#include "cachecleaner.hh"
+extern StatBag S;
+
+AuthDomainCache::AuthDomainCache(size_t mapsCount) :
+  d_maps(mapsCount)
+{
+  S.declare("domain-cache-hit", "Number of hits on the domain cache");
+  S.declare("domain-cache-miss", "Number of misses on the domain cache");
```
jsoref, line 37:
```suggestion
  S.declare("domain-cache-hit", "Number of hits of the domain cache");
  S.declare("domain-cache-miss", "Number of misses of the domain cache");
```
?

pdns/auth-domaincache.cc
```diff
@@ -0,0 +1,166 @@
...
+    CacheValue val;
+    val.zoneId = tup.get<1>();
+    auto& mc = newMaps[getMapIndex(domain)];
+    mc.d_map.emplace(domain, val);
+  }
+
+  {
+    WriteLock globalLock(d_mut);
+    if (d_replacePending) {
+      // add/replace all domains created while data collection for replace() was already running.
```
jsoref, line 112:
```suggestion
      // add/replace all domains created while data collection for replace() was in progress.
```

pdns/auth-domaincache.hh
```diff
@@ -0,0 +1,96 @@
...
+
+class AuthDomainCache : public boost::noncopyable
+{
+public:
+  AuthDomainCache(size_t mapsCount = 1024);
+  ~AuthDomainCache();
+
+  void replace(const vector<tuple<DNSName, int>>& domains);
+  void add(const DNSName& domain, const int zoneId);
+  void setReplacePending();  //!< call this when data collection for subsequent replace() call starts.
```
jsoref, line 36: I can't parse this at 1am, maybe it makes sense at some other hour

pdns/common_startup.cc
```diff
@@ -176,6 +177,7 @@ void declareArguments()
   ::arg().set("cache-ttl","Seconds to store packets in the PacketCache")="20";
   ::arg().set("negquery-cache-ttl","Seconds to store negative query results in the QueryCache")="60";
   ::arg().set("query-cache-ttl","Seconds to store query results in the QueryCache")="20";
+  ::arg().set("domain-cache-ttl","Seconds to cache list of known domains")="0";
   ::arg().set("server-id", "Returned when queried for 'id.server' TXT or NSID, defaults to hostname - disabled or custom")="";
   ::arg().set("default-soa-content","Default SOA content")="a.misconfigured.dns.server.invalid hostmaster.@ 0 10800 3600 604800 3600";
   ::arg().set("default-soa-edit","Default SOA-EDIT value")="";
```
jsoref, line 183: Some of these have a space after comma and some don't

pdns/ueberbackend.cc
```diff
@@ -120,12 +121,18 @@ bool UeberBackend::getDomainInfo(const DNSName &domain, DomainInfo &di, bool get
...
-      return true;
+    if(mydb->createDomain(domain, kind, masters, account, &zoneId)) {
+      success = true;
+      break;
     }
   }
-  return false;
+  if (success) {
+    g_domainCache.add(domain, zoneId);  // make new domain visible
+  }
```
jsoref, line 134: For the peanut gallery: Code scanning isn't smart enough to know that `zoneId` is guarded by `success` and success will only be true if zoneId is filled in.

To make things more complicated, `zoneId` is a pointer not a reference on the callee side, and callees obviously only write to it if the pointer is non null.

pdns/ueberbackend.cc
```diff
@@ -322,20 +349,39 @@ bool UeberBackend::getAuth(const DNSName &target, const QType& qtype, SOAData* s
...
-
   bool found = false;
   int cstat;
   DNSName shorter(target);
   vector<pair<size_t, SOAData> > bestmatch (backends.size(), make_pair(target.wirelength()+1, SOAData()));
   do {
+    int zoneId{-1};
+    if(cachedOk && g_domainCache.isEnabled()) {
+      if (g_domainCache.getEntry(shorter, zoneId)) {
+        // Zone exists in domain cache, directly lookup SOA.
```
jsoref, line 360:
```suggestion
        // Zone exists in domain cache, directly look up SOA.
```

pdns/auth-domaincache.cc
Habbie: `on` is better than `of`. `in` might be better than `on`.

Also, `on` is consistent with existing entries.

pdns/auth-domaincache.cc
Habbie: 'already in progress', perhaps

pdns/auth-domaincache.cc
jsoref:
```suggestion
  S.declare("domain-cache-hit", "Number of hits on the domain cache");
  S.declare("domain-cache-miss", "Number of domain cache misses");
```

jsoref: (and similar for the other line if this seems like a good style)

Habbie: This makes sense to me, but then we should also change the existing similar entries.

modules/lmdbbackend/lmdbbackend.cc
```diff
@@ -953,22 +953,29 @@ bool LMDBBackend::setMasters(const DNSName& domain, const vector<ComboAddress>&
...
-  di.account = account;
+    txn.put(di);
+    txn.commit();
+  }
 
-  txn.put(di);
-  txn.commit();
+  if (zoneId != nullptr) {
+    auto txn = d_tdomains->getROTransaction();
+    *zoneId = txn.get<0>(domain, di);
```
Habbie, line 977: Can't you do this in the RW transaction that made the domain?

pdns/auth-domaincache.cc
```diff
@@ -0,0 +1,166 @@
...
+    auto& mc = d_maps[mapIndex];
+    WriteLock mcLock(mc.d_mut);
+    mc.d_map.emplace(domain, val);
+  }
+}
+
+void AuthDomainCache::setReplacePending()
+{
+  if (!d_ttl)
+    return;
```
Habbie, line 161: does this manage to skip the lock?

pdns/backends/gsql/gsqlbackend.cc
```diff
@@ -1287,6 +1287,20 @@ bool GSQLBackend::createDomain(const DNSName &domain, const DomainInfo::DomainKi
       bind("account", account)->
       execute()->
       reset();
+
+    if (zoneId != nullptr) {
+      // XXX: needs its own stmt as godbc has a table name in it
```
Habbie, line 1292: :exclamation:

pdns/common_startup.cc
```diff
@@ -673,12 +675,36 @@ void mainthread()
...
 
+  const uint32_t secpollInterval = 1800;
+  uint32_t secpollSince = 0;
+  uint32_t domainCacheUpdateSince = 0;
   for(;;) {
-    sleep(1800);
-    try {
-      doSecPoll(false);
+    const uint32_t slept = g_domainCache.getTTL() == 0 ? secpollInterval : std::min(secpollInterval, g_domainCache.getTTL());
+    sleep(slept);  // if any signals arrive, we might run more often than expected.
```
Habbie, line 683: This is a very confusing line, because `sleep` -returns- the number of seconds slept, but we don't use that value (and the 'might run more often' comment also suggests you actually wanted the return value!).

Maybe rename your `slept` to `sleeptime`?

pdns/test-auth-domaincache_cc.cc
```diff
@@ -0,0 +1,95 @@
...
+  cache.setTTL(3600);
+
+  vector<tuple<DNSName, int>> domain_indices{
+    {DNSName("example.org."), 1},
+  };
+  cache.replace(domain_indices);
+
+  int zoneId = 0;
+  bool found = cache.getEntry(DNSName("example.org."), zoneId);
+  if (!found || zoneId == 0) {
```
Habbie, line 49:
```suggestion
  if (!found || zoneId != 1) {
```

Habbie (commented):
Awesome! I posted a few questions and nits. The godbc thing will have to be fixed (or at least documented) and then I think we're good to merge.

pdns/common_startup.cc
zeha: Indeed, and I do not know what the correct style is.

pdns/backends/gsql/gsqlbackend.cc
zeha: oh right, totally forgot about this :(

pdns/common_startup.cc
jsoref: Personally, I'd vote for with. But obviously defer to someone else

Habbie: If you're not making it worse, I'll accept it. You could run the format script over it to see what it does.

Habbie: (I'd also vote for the space)

pdns/backends/gsql/gsqlbackend.cc
Habbie: If at some point this is the last blocker, I'm okay with noting this one in a separate ticket, so we can get 4.5.0-alpha1 out the door.

pdns/test-auth-domaincache_cc.cc
zeha: I can't say I understand the logic here?

zeha:
Note to self: createDomain now fails

pdns/auth-domaincache.cc
zeha: i've added a new {} block to make sure of it

pdns/auth-domaincache.cc
zeha: I've done it here, but the other places need to be in a separate PR

pdns/common_startup.cc
zeha: spaced! the other occurrences should be in another PR imo

pdns/backends/gsql/gsqlbackend.cc
zeha: I've added a new statement.

pdns/auth-domaincache.cc
Habbie: agreed

pdns/auth-domaincache.hh
Habbie: I have no trouble with it, and I do not have better words

pdns/auth-domaincache.hh
```diff
@@ -0,0 +1,98 @@
...
+
+class AuthDomainCache : public boost::noncopyable
+{
+public:
+  AuthDomainCache(size_t mapsCount = 1024);
+  ~AuthDomainCache();
+
+  void replace(const vector<tuple<DNSName, int>>& domains);
+  void add(const DNSName& domain, const int zoneId);
+  void setReplacePending(); //!< call this when data collection for subsequent replace() call starts.
```
jsoref, line 38:
```suggestion
  void setReplacePending(); //!< call this when data collection for a subsequent replace() call starts.
```

jsoref, line 38:
```suggestion
  void setReplacePending(); //!< call this when data collection for the subsequent replace() call starts.
```

jsoref (commented):
It isn't 1am

pdns/auth-domaincache.hh
zeha: thanks, applied

zeha:
I'll probably want to apply the same trick as in #10392 for the new query here.

Habbie:
Yep! Was going to ping you about that :)

Habbie:
Related, while I was talking to @mind04 about this, he suggested that generally, the last-id query could be made more robust by asking for the domain name you just added.

mind04 (changes requested):

pdns/ueberbackend.cc
```diff
@@ -322,20 +349,39 @@ bool UeberBackend::getAuth(const DNSName &target, const QType& qtype, SOAData* s
...
+    int zoneId{-1};
+    if(cachedOk && g_domainCache.isEnabled()) {
+      if (g_domainCache.getEntry(shorter, zoneId)) {
+        // Zone exists in domain cache, directly look up SOA.
+        // XXX: this code path and the cache lookup below should be merged; but that needs the code path below to also use ANY.
+        // Or it should just also use lookup().
+        DNSZoneRecord zr;
+        lookup(QType(QType::SOA), shorter, zoneId, nullptr);
+        if (!get(zr)) {
+          throw PDNSException("Backend returned no SOA for existing domain '"+shorter.toLogString()+"'");
```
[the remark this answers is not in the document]
zeha: I can make this `return false;` instead. If the zone ended up in the cache, we should probably be using it. If its disabled, it should not be in the cache -- if it is newly disabled, we can make it "not found" until the cache TTL expires.

modules/tinydnsbackend/tinydnsbackend.cc
```diff
@@ -170,7 +170,15 @@ void TinyDNSBackend::getAllDomains(vector<DomainInfo>* domains, bool include_dis
...
 
-  while (get(rr)) {
+  while (true) {
+    try {
+      if (!get(rr))
+        break;
+    }
+    catch (...) {
+      // Error already logged by get().
+      continue;
```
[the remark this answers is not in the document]
zeha: This is true. The other real option is `return`, which will cause one broken RR to hide all domains -after- that RR.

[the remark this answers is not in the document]
Habbie: How did this come up, and became part of this PR?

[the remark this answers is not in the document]
zeha: The tinydns tests fail otherwise, IIRC.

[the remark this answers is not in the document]
zeha: I've removed commit 029338bb8affe1389985012363c5968ca2f26463

pdns/auth-domaincache.hh
Habbie: done, I think?

pdns/ueberbackend.cc
```diff
@@ -322,20 +349,43 @@ bool UeberBackend::getAuth(const DNSName &target, const QType& qtype, SOAData* s
...
+    if(cachedOk && g_domainCache.isEnabled()) {
+      if (g_domainCache.getEntry(shorter, zoneId)) {
+        // Zone exists in domain cache, directly look up SOA.
+        // XXX: this code path and the cache lookup below should be merged; but that needs the code path below to also use ANY.
+        // Or it should just also use lookup().
+        DNSZoneRecord zr;
+        lookup(QType(QType::SOA), shorter, zoneId, nullptr);
+        if (!get(zr)) {
+          g_log << Logger::Info << "Backend returned no SOA for domain '" << shorter.toLogString() << "', which it reported as existing " << endl;
+          return false;
```
mind04, line 367: continue; is the proper action here. Now you are hiding parent zones of disabled zone.
Please put the log message behind log-dns-details. Disabling a zone with a lot of qps will flood the logs and this might hurt performance for the other zones.

regression-tests/backends/gsql-common
```diff
@@ -50,7 +50,7 @@ gsql_master()
         --config-name=$backend --socket-dir=./ --no-shuffle \
         --dnsupdate=yes --resolver=$RESOLVERIP --outgoing-axfr-expand-alias=yes \
         --expand-alias=yes \
-        --cache-ttl=$cachettl --dname-processing \
+        --cache-ttl=$cachettl --domain-cache-ttl=$cachettl --dname-processing \
```
mind04, line 53: I don't think we have Ci with a set cachettl. A fixed value in one of the backends will make sure we test the domain cache in the regressiontests. gsqlite3 is a good candidate (also used for consistent-backends).

pdns/ueberbackend.cc
```diff
@@ -273,16 +280,36 @@ void UeberBackend::reload()
...
+    return;
+  }
+
+  vector<tuple<DNSName, int>> domain_indices;
+  g_domainCache.setReplacePending();
+
+  for (vector<DNSBackend*>::iterator i = backends.begin(); i != backends.end(); ++i )
+  {
+    vector<DomainInfo> domains;
+    (*i)->getAllDomains(&domains, false);
```
mind04, line 294: Please include disabled domains as well, otherwise re-enabled zones and new secondary zones will not show up until the domain cache has expired. This will prevent the use of large ttl's on secondaries and api only installs.

mind04 (changes requested):

pdns/ueberbackend.cc
zeha: My opinion: it's perfectly fine that disabled zones are not visible until the cache expires.

zeha:
Not sure I edited the correct file for adding cachettl to one CI run.

pdns/ueberbackend.cc
mind04: What is the advantage of not including disabled zones? 
Performance comes to mind, but the performance impact of including disabled zones is close to zero (the extra required SOA lookup is cached by the query cache).
For new secondary zones (zones without records) fast availability of new zone content is important. Some registries have a predelegation checks (.de). Not including disabled zones here will delay this check for no reason and will reduce the usefulness of this cache.

Habbie:
> Not sure I edited the correct file for adding cachettl to one CI run.

That file is not used anywhere, perhaps we should delete it.

Habbie:
```
regression-tests$ git grep consistent-backends
backends/gsqlite3-master:consistent-backends
backends/gsqlite3-slave:consistent-backends
```

is how consistent-backends is tested; you probably want to do something similar

pdns/ueberbackend.cc
zeha: The original usecase doesn't really care about either. I'll flip this to true, but I can't do much thinking what impact it will have.

zeha:
> > Not sure I edited the correct file for adding cachettl to one CI run.
> 
> That file is not used anywhere, perhaps we should delete it.

I tried to understand which files in ./build-scripts/ are used, but its really hard to tell for me. Someone else should do that (in another PR, anyway).

Habbie:
This looks awesome to me, except for two things:
* CI (I have the fixes locally but I'll hold them so I can update them after the next item)
* the use of `ttl` in both the setting and the code is confusing. It's not expiry on items; the items do not have a TTL. The domain cache has a refresh interval. Do you want to change that? I'm also okay with 'that is a good point, please go ahead and change it'.

Habbie:
@mind04 points out that, if we're renaming anyway, we should really replace `domain` with `zone`. I'm happy to do this.

zeha:
I'm happy to change domain-cache-ttl to something else, but I don't have good ideas on the name. zone-cache-refresh-interval?

Habbie:
> I'm happy to change domain-cache-ttl to something else, but I don't have good ideas on the name. zone-cache-refresh-interval?

perfect

zeha:
Renamed. Re CI: please apply the changes as needed.

mind04 (approved):

Issue #1885, opened 2014-11-22
Please dump cache over socket


zeha:
Given new security features in init systems (e.g. systemd's `PrivateTmp`, and others), having a daemon write to an admin-controlled path is likely to get harder / impossible.

Please add support for `rec_control dump-cache` to send the cache over the control socket, instead of directly writing the contents to a file.

(If `pdns_control` has a similar thing, same concern...)

omoerbeek:
Different solution in #9468, same result.

PR #9468, opened 2020-09-14
rec: Rec control pass fd

omoerbeek:
### Short description

This makes `rec_control` open the file used by `dump-cache` and other dumping command and pass the file descriptor over the socket. This makes it more easy to export data if the recursor is running with restricted file access (e.g. chroot).

Also allows to dump the data to stdout (`rec_control dump-cache -`). 

I also added code to separate the error code from the string returned. This principle could also be extended to any command that could return an error.

Note that this causes an incompatibility between `rec_control` and `pdns_recursor`. I don't think we ever promised the API between the two would be stable or compatible between versions, but you know, people will stumble over this anyway. A note in the upgrade guide will be needed.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [X] documented the code

### DIFF
--- a/pdns/pdns_recursor.cc
+++ b/pdns/pdns_recursor.cc
@@ -3842,13 +3842,6 @@ static vector<pair<DNSName, uint16_t> >& operator+=(vector<pair<DNSName, uint16_
   return a;
 }
 
-static pair<int, string>& operator+=(std::pair<int, string>& a, const std::pair<int, string>& b)
-{
-  a.first |= b.first;
-  a.second += b.second;
-  return a;
-}
-
 /*
   This function should only be called by the handler to gather metrics, wipe the cache,
   reload the Lua script (not the Lua config) or change the current trace regex,
@@ -3891,7 +3884,7 @@ template<class T> T broadcastAccFunction(const boost::function<T*()>& func)
 }
 
 template string broadcastAccFunction(const boost::function<string*()>& fun); // explicit instantiation
-template std::pair<int, string> broadcastAccFunction(const boost::function<pair<int,string>*()>& fun); // explicit instantiation
+template RecursorControlChannel::Answer broadcastAccFunction(const boost::function<RecursorControlChannel::Answer *()>& fun); // explicit instantiation
 template uint64_t broadcastAccFunction(const boost::function<uint64_t*()>& fun); // explicit instantiation
 template vector<ComboAddress> broadcastAccFunction(const boost::function<vector<ComboAddress> *()>& fun); // explicit instantiation
 template vector<pair<DNSName,uint16_t> > broadcastAccFunction(const boost::function<vector<pair<DNSName, uint16_t> > *()>& fun); // explicit instantiation
@@ -3901,7 +3894,7 @@ static void handleRCC(int fd, FDMultiplexer::funcparam_t& var)
 {
   try {
     string remote;
-    string msg = s_rcc.recv(&remote).second;
+    string msg = s_rcc.recv(&remote).d_str;
     RecursorControlParser rcp;
     RecursorControlParser::func_t* command;
 
@@ -4117,14 +4110,14 @@ static FDMultiplexer* getMultiplexer()
 }
 
 
-static std::pair<int, string>* doReloadLuaScript()
+static RecursorControlChannel::Answer* doReloadLuaScript()
 {
   string fname= ::arg()["lua-dns-script"];
   try {
     if(fname.empty()) {
       t_pdl.reset();
       g_log<<Logger::Info<<t_id<<" Unloaded current lua script"<<endl;
-      return new std::pair<int, string>(0, string("unloaded\n"));
+      return new RecursorControlChannel::Answer{0, string("unloaded\n")};
     }
     else {
       t_pdl = std::make_shared<RecursorLua4>();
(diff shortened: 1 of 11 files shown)


zeha:
This would close #1885

pieterlexis:
> Before spending more time on this I'd like to know if we want anything like this.

yes, although I would prefer an HTTP API for control eventually

pdns/pdns_recursor.cc
```diff
@@ -3777,6 +3777,13 @@ static vector<pair<DNSName, uint16_t> >& operator+=(vector<pair<DNSName, uint16_
   return a;
 }
 
+static pair<int, string>& operator+=(std::pair<int, string>& a, const std::pair<int, string>& b)
```
rgacogne, line 3780: I'm a bit worried about defining an operator for a generic type like that, perhaps defining a new struct would be better?

pdns/rec_channel.cc
```diff
@@ -162,30 +197,45 @@ void RecursorControlChannel::send(const std::string& msg, const std::string* rem
...
+  if (ret == 0) {
     throw PDNSException("Timeout waiting for answer from control channel");
-  
-  if( ret < 0 || (len=::recvfrom(d_fd, buffer, sizeof(buffer), 0, (struct sockaddr*)&remoteaddr, &addrlen)) < 0)
+  }
+  int err;
+  if (ret < 0 || ::recvfrom(d_fd, &err, sizeof(err), 0, (struct sockaddr*)&remoteaddr, &addrlen) < 0) {
+    throw PDNSException("Unable to receive return status over control channel: " + stringerror());
+  }
+  if (ret < 0 || (len = ::recvfrom(d_fd, buffer, sizeof(buffer), 0, (struct sockaddr*)&remoteaddr, &addrlen)) < 0) {
```
rgacogne, line 231: The second `ret < 0` check seems redundant here, or did you meant to call `waitForData` again, maybe?

pdns/rec_channel_rec.cc
```diff
@@ -291,24 +291,56 @@ static string doDumpNSSpeeds(T begin, T end)
...
+  io_vector[0].iov_base = &ch;
+  io_vector[0].iov_len = 1;
+
+  memset(&msg, 0, sizeof(msg));
+  msg.msg_control = &cmsgbuf.buf;
+  msg.msg_controllen = sizeof(cmsgbuf.buf);
+  msg.msg_iov = io_vector;
+  msg.msg_iovlen = 1;
+
+  if (recvmsg(s, &msg, 0) == -1)
```
rgacogne, line 316: Just a nit but I'd prefer to see braces around all conditional blocks.

pdns/rec_channel_rec.cc
```diff
@@ -291,24 +291,56 @@ static string doDumpNSSpeeds(T begin, T end)
...
+    throw PDNSException("recvmsg");
+  if ((msg.msg_flags & MSG_TRUNC) || (msg.msg_flags & MSG_CTRUNC))
+    throw PDNSException("control message truncated");
+  for (cmsg = CMSG_FIRSTHDR(&msg); cmsg != NULL;
+       cmsg = CMSG_NXTHDR(&msg, cmsg)) {
+    if (cmsg->cmsg_len == CMSG_LEN(sizeof(int)) &&
+        cmsg->cmsg_level == SOL_SOCKET &&
+        cmsg->cmsg_type == SCM_RIGHTS) {
+      fd = *(int *)CMSG_DATA(cmsg);
```
rgacogne, line 325: Perhaps we should break there?

pdns/pdns_recursor.cc
```diff
@@ -3777,6 +3777,13 @@ static vector<pair<DNSName, uint16_t> >& operator+=(vector<pair<DNSName, uint16_
...
+{
+  a.first |= b.first;
```
rgacogne, line 3782: I wonder if we should really do a OR here, because if we ever get two different non-zero values the result will not make any sense. Perhaps we should only override `a.first` if `a.first` is zero and `b.first` is non-zero?

rgacogne (commented):
I like the concept of passing a file descriptor to prevent chroot/namespaces issues a lot, for commands that are dumping output to a file.

Trying to sum up what this change means from a security perspective:
- the recursor can write outside of its own chroot, to a file that it might not have the rights to write to otherwise, but only if a process with enough privileges to write to that file does request it. That seems fine.
- the other way around, someone with enough privileges to write to the recursor's communication socket can request the recursor to write to an-attacker controller file descriptor, but the attacker needs enough privileges to open the file. That seems fine as well. In fact, that seems better than that we have today.

rgacogne:
Writing it down because I always need to think about it when I consider this change: it does indeed means that some `rec_control` operations do not make sense from a remote host since we can't pass a valid file descriptor, but the existing code was already dumping to a local file anyway so that does not really change anything.

pieterlexis:
@omoerbeek what is the status here?

omoerbeek:
> @omoerbeek what is the status here?

I hope to transform this from a POC into a real PR soon.

omoerbeek:
> > Before spending more time on this I'd like to know if we want anything like this.
> 
> yes, although I would prefer an HTTP API for control eventually

eventually yes, atm we cannot do all that's needed via the HTTP API, plus if you're talking cache dumps, it seems a bit overkill to change a local file write into a download of sometimes huge files.

omoerbeek:
Fixes #9634

pdns/rec_channel_rec.cc
```diff
@@ -219,6 +219,47 @@ string static doGetParameter(T begin, T end)
   return ret;
 }
 
+/* Read an (open) fd from the control channel */
+static int
+getfd(int s)
```
rgacogne, line 224: A small nit but I would really like to see that file descriptor wrapped in an object like our `Socket` object, to make sure it is properly closed when it goes out of scope. It looks like you are properly handling exceptions today, though, but it might not always be the case.

pdns/pdns_recursor.cc
```diff
@@ -4109,35 +4110,40 @@ static FDMultiplexer* getMultiplexer()
...
     if(fname.empty()) {
       t_pdl.reset();
       g_log<<Logger::Info<<t_id<<" Unloaded current lua script"<<endl;
-      return new string("unloaded\n");
+      return new RecursorControlChannel::Answer{0, string("unloaded\n")};
     }
     else {
       t_pdl = std::make_shared<RecursorLua4>();
-      t_pdl->loadFile(fname);
+      int err = t_pdl->loadFile(fname);
```
rgacogne, line 4124: Not related to this PR but adding the same kind of check on the initial loading of the file would fix https://github.com/PowerDNS/pdns/issues/10079

rgacogne (approved):
Looks good, a couple minor nits.

PR #9470, opened 2020-09-15
Remove UeberBackend::getSOA

zeha:
### Short description
Makes reasoning about #9007 easier.

### Checklist

I have:
- [x] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/slavecommunicator.cc
+++ b/pdns/slavecommunicator.cc
@@ -941,8 +941,16 @@ void CommunicatorClass::slaveRefresh(PacketHandler *P)
 
     bool hasSOA = false;
     SOAData sd;
-    try{
-      hasSOA = B->getSOA(di.zone, sd);
+    sd.serial = 0;
+    try {
+      // Use UeberBackend cache for SOA. Cache gets cleared after AXFR/IXFR.
+      B->lookup(QType(QType::SOA), di.zone, di.id, nullptr);
+      DNSZoneRecord zr;
+      hasSOA = B->get(zr);
+      if (hasSOA) {
+        fillSOAData(zr, sd);
+        while(B->get(zr));
+      }
     }
     catch(...) {}
 
--- a/pdns/slavecommunicator.cc
+++ b/pdns/slavecommunicator.cc
@@ -900,7 +900,9 @@ void CommunicatorClass::slaveRefresh(PacketHandler *P)
   time_t now = time(0);
   for(auto& val : sdomains) {
     DomainInfo& di(val.di);
-    // might've come from the packethandler
+    // If our di comes from packethandler (caused by incoming NOTIFY), di.backend will not be filled out,
+    // and di.serial will not either.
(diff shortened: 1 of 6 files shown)


pdns/slavecommunicator.cc
```diff
@@ -939,22 +942,15 @@ void CommunicatorClass::slaveRefresh(PacketHandler *P)
...
-    bool hasSOA = false;
-    SOAData sd;
-    try{
-      hasSOA = B->getSOA(di.zone, sd);
-    }
-    catch(...) {}
-
     uint32_t theirserial = ssr.d_freshness[di.id].theirSerial;
-    uint32_t ourserial = sd.serial;
+    uint32_t ourserial = di.serial;
```
rgacogne, line 946: I was a bit worried that this change was undoing https://github.com/PowerDNS/pdns/pull/7245, but it looks fine since di.serial is now properly updated above.

pdns/packethandler.cc
```diff
@@ -456,7 +456,7 @@ int PacketHandler::doAdditionalProcessingAndDropAA(DNSPacket& p, std::unique_ptr
          !(i->dr.d_type==QType::MX || i->dr.d_type==QType::NS || i->dr.d_type==QType::SRV))
         continue;
 
-      if(r->d.aa && i->dr.d_name.countLabels() && i->dr.d_type==QType::NS && !B.getSOA(i->dr.d_name,sd) && !retargeted) { // drop AA in case of non-SOA-level NS answer, except for root referral
+      if(r->d.aa && i->dr.d_name.countLabels() && i->dr.d_type==QType::NS && !i->auth && !retargeted) { // drop AA in case of non-SOA-level NS answer, except for root referral
```
rgacogne, line 459: Is `auth` always correctly set, even in non DNSSEC zones or non DNSSEC-capable backends?

rgacogne (commented):
The general idea looks to me since I don't know of any clear case where it makes sense for a backend to override `getSOA()`.

pdns/packethandler.cc
zeha: This is indeed a potential problem - one where I'm not sure how to test this / what would actually break.

IMO we could also call getAuth instead, which would at least be correct over getSOA. The perf gain from avoiding another DB lookup could be done in a separate PR.

zeha:
Generally my idea was:
- apply this MR
- then #9007
- then see about #9464, in which we could think about caching the zone SOA too, removing that lookup as well (assuming consistent-backends=yes)

pdns/slavecommunicator.cc
zeha: Yes, I hope I understood the intention of #7245 correctly.

pdns/packethandler.cc
zeha: In #9478 @mind04 fixed this bit.

PR #9471, opened 2020-09-15
rec: Log when going Bogus because of a missing SOA in authority

rgacogne:
### Short description

A missing SOA in the authority section of negative (NXDOMAIN, NODATA) answers in a DNSSEC-secure zone currently leads to a Bogus result, because the needed NSEC/NSEC3 could not be validated.

Perhaps we could try to detect such a case and check for a valid denial of proof if a NSEC/NSEC3 record shows up in authority?

In the meantime this PR makes it clearer why the DNSSEC status was updated to Bogus in the trace.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


rgacogne:
> Perhaps we could try to detect such a case and check for a valid denial of proof if a NSEC/NSEC3 record shows up in authority?

We need to be careful about the DS denial of existence in delegations if we ever do that, though.

omoerbeek (approved):

PR #9472, opened 2020-09-16
dnsbackend: log error on catching std::exception

pieterlexis:
### Short description
This prevents the following logging when an std::exception is thrown by a backend.

```
Sep 16 07:39:31 debian-9 pdns_server[3011]: Caught an exception instantiating a backend, cleaning up
Sep 16 07:39:31 debian-9 pdns_server[3011]: Uncaught exception of unknown type - sorry
```

### Checklist

I have:
- [x] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsbackend.cc
+++ b/pdns/dnsbackend.cc
@@ -211,6 +211,13 @@ vector<DNSBackend *> BackendMakerClass::all(bool metadataOnly)
       delete i;
     }
     throw;
+  } catch (const std::exception &e) {
+    g_log<<Logger::Error<<"Caught an exception instantiating a backend: "<<e.what()<<endl;
+    g_log<<Logger::Error<<"Cleaning up"<<endl;
+    for (auto i : ret) {
+      delete i;
+    }
+    throw;
   } catch(...) {
     // and cleanup
     g_log<<Logger::Error<<"Caught an exception instantiating a backend, cleaning up"<<endl;


Habbie (approved):

zeha:
The `Uncaught exception of unknown type - sorry` message will stay, right?

pieterlexis:
> The `Uncaught exception of unknown type - sorry` message will stay, right?

We re-throw after logging, so it depends on where it is caught :)

zeha:
> > The `Uncaught exception of unknown type - sorry` message will stay, right?
> 
> We re-throw after logging, so it depends on where it is caught :)

Right. I'm a bit worried about how this works, because you add:
```
  } catch (const std::exception &e) {
```

and receiver.cc already has:

```
  catch(std::exception &e) {
    ...
    g_log<<Logger::Error<<"Exiting because of STL error: "<<e.what()<<endl;
  }
  catch(...) {
    cerr<<"Uncaught exception of unknown type - sorry"<<endl;
  }
```

So i'm not sure how you can end up with "unknown type" for a `std::exception` in the first place ... 😕

pieterlexis:
hmmmmm... perhaps it is a DBException that leaks here

zeha:
> hmmmmm... perhaps it is a DBException that leaks here. That would mean that _that_ is not caught properly in the backend (mysql I think in this case),

Could also be `SSqlException` from a quick glance...

PR #9475, opened 2020-09-16
Rec: shared and sharded negcache

omoerbeek:
### Short description

While the main record cache is shared between threads since some time, the negcache was not. This makes is so.

While there, simplify the `lruReplacingInsert()` result to just signal if an entry was inserted or replaced (this makes counting more easy) and generalize the invalidating of the per-shard cache used in the record cache so the code can be more easily be reused in the negcache.

This code keeps track of mutex acquired/contended counts. Don't know if this is going to stay, it is not exported in any way atm.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/pdns_recursor.cc
+++ b/pdns/pdns_recursor.cc
@@ -131,6 +131,7 @@ static thread_local uint64_t t_frameStreamServersGeneration;
 
 thread_local std::unique_ptr<MT_t> MT; // the big MTasker
 std::unique_ptr<MemRecursorCache> s_RC;
+std::unique_ptr<NegCache> s_negcache;
 
 thread_local std::unique_ptr<RecursorPacketCache> t_packetCache;
 thread_local FDMultiplexer* t_fdm{nullptr};
@@ -3229,7 +3230,6 @@ static void houseKeeping(void *)
     past.tv_sec -= 5;
     if (last_prune < past) {
       t_packetCache->doPruneTo(g_maxPacketCacheEntries / g_numWorkerThreads);
-      SyncRes::pruneNegCache(g_maxCacheEntries / (g_numWorkerThreads * 10));
 
       time_t limit;
       if(!((cleanCounter++)%40)) {  // this is a full scan!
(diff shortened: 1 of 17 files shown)


pdns/pdns_recursor.cc
```diff
@@ -131,6 +131,7 @@ static thread_local uint64_t t_frameStreamServersGeneration;
 
 thread_local std::unique_ptr<MT_t> MT; // the big MTasker
 std::unique_ptr<MemRecursorCache> s_RC;
+std::unique_ptr<NegCache> s_negcache;
```
rgacogne, line 134: The `s_` prefix is usually for variables static to a file, or static to a class, so perhaps use `g_` instead?

pdns/recursordist/negcache.cc
```diff
@@ -70,24 +101,27 @@ bool NegCache::getRootNXTrust(const DNSName& qname, const struct timeval& now, N
  */
 bool NegCache::get(const DNSName& qname, const QType& qtype, const struct timeval& now, NegCacheEntry& ne, bool typeMustMatch)
 {
-  const auto& idx = d_negcache.get<2>();
+  auto& map = getMap(qname);
+  const lock l(map);
+
+  const auto& idx = map.d_map.get<2>();
```
rgacogne, line 107:
```suggestion
  const auto& idx = map.d_map.get<NegCacheEntry>();
```

pdns/recursordist/negcache.cc
```diff
@@ -70,24 +101,27 @@ bool NegCache::getRootNXTrust(const DNSName& qname, const struct timeval& now, N
...
   auto range = idx.equal_range(qname);
   auto ni = range.first;
 
   while (ni != range.second) {
     // We have an entry
     if ((!typeMustMatch && ni->d_qtype.getCode() == 0) || ni->d_qtype == qtype) {
       // We match the QType or the whole name is denied
-      auto firstIndexIterator = d_negcache.project<0>(ni);
+      auto firstIndexIterator = map.d_map.project<0>(ni);
```
rgacogne, line 115:
```suggestion
      auto firstIndexIterator = map.d_map.project<CompositeKey>(ni);
```

rgacogne (approved):

pdns/pdns_recursor.cc
omoerbeek: Right. I opt for` g_NegCache`  and might rename `s_RC`  then into `g_RecCache` as well.

PR #9488, opened 2020-09-21
Sync string_view definition in lmdb-safe

rgacogne:
### Short description

While we do not actually care about `string_view::at()` in lmdb-safe, we need to keep the two in sync so the type aliases do not collide.
Perhaps we should consider using `pdns_string_view` as an alias (outside of lmdb-safe) instead?

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/ext/lmdb-safe/lmdb-safe.hh
+++ b/ext/lmdb-safe/lmdb-safe.hh
@@ -16,12 +16,16 @@
 using std::string_view;
 #else
 #include <boost/version.hpp>
-#if BOOST_VERSION >= 106100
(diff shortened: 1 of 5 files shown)


Habbie:
>  Perhaps we should consider using `pdns_string_view` as an alias (outside of lmdb-safe) instead?

That makes sense to me.

rgacogne:
> Perhaps we should consider using pdns_string_view as an alias (outside of lmdb-safe) instead?

Done.

PR #9490, opened 2020-09-21
rec: Check return value of dup(2) as noted by coverity.

omoerbeek:
### Checklist

I have:
- [X] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/rec_channel_rec.cc
+++ b/pdns/rec_channel_rec.cc
@@ -201,14 +201,16 @@ string static doGetParameter(T begin, T end)
 
 static uint64_t dumpNegCache(int fd)
 {
-  auto fp = std::unique_ptr<FILE, int(*)(FILE*)>(fdopen(dup(fd), "w"), fclose);
-  if(!fp) { // dup probably failed
+  int newfd = dup(fd);
+  if (newfd == -1) {
     return 0;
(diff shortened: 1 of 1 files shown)


pdns/rec_channel_rec.cc
```diff
@@ -201,14 +201,16 @@ string static doGetParameter(T begin, T end)
...
 static uint64_t dumpNegCache(int fd)
 {
-  auto fp = std::unique_ptr<FILE, int(*)(FILE*)>(fdopen(dup(fd), "w"), fclose);
-  if(!fp) { // dup probably failed
+  int newfd = dup(fd);
+  if (newfd == -1) {
+    return 0;
+  }
+  auto fp = std::unique_ptr<FILE, int(*)(FILE*)>(fdopen(newfd, "w"), fclose);
+  if (fp == nullptr) {
```
cmouse, line 209: will it compare unique_ptr or within unique ptr though?

omoerbeek: https://en.cppreference.com/w/cpp/memory/unique_ptr/operator_cmp case 8.
But if you find it confusing maybe it's better to put the `!fp` check back.

PR #9495, opened 2020-09-21
rec: Watch the descriptor again after an out-of-order read timeout

rgacogne:
### Short description

It might be that there was no other incoming query on that connection and we timed out while the response had not been sent yet, but the client might want to re-use the connection after receiving the response.
We try to reset the TTD, but that might fail when the socket descriptor has already been removed.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/pdns_recursor.cc
+++ b/pdns/pdns_recursor.cc
@@ -1915,7 +1915,7 @@ static void startDoResolve(void *p)
               t_fdm->setReadTTD(dc->d_socket, ttd, g_tcpTimeout);
             }
             catch (const FDMultiplexerException &) {
-              // but the FD was removed because of a timeout while we were sending a response,
+              // but if the FD was removed because of a timeout while we were sending a response,
               // we need to re-arm it. If it was an error it will error again.
               ttd.tv_sec += g_tcpTimeout;
               t_fdm->addReadFD(dc->d_socket, handleRunningTCPQuestion, dc->d_tcpConnection, &ttd);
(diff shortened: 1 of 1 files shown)


rgacogne:
I forgot to mention that the issue was reported by Tomas Krizek from CZ.NIC, while doing some test with DNS Shotgun [1]. Thanks a lot!

[1]: https://github.com/CZ-NIC/shotgun

pdns/pdns_recursor.cc
```diff
@@ -1910,11 +1910,15 @@ static void startDoResolve(void *p)
...
             t_fdm->addReadFD(dc->d_socket, handleRunningTCPQuestion, dc->d_tcpConnection, &ttd);
           } else {
-            // fd might have been removed by read error code, so expect an exception
+            // fd might have been removed by read error code, or a read timeout, so expect an exception
             try {
               t_fdm->setReadTTD(dc->d_socket, ttd, g_tcpTimeout);
             }
-            catch (FDMultiplexerException &) {
+            catch (const FDMultiplexerException &) {
+              // but the FD was removed because of a timeout while we were sending a response,
```
omoerbeek, line 1918:
```suggestion
              // but if the FD was removed because of a timeout while we were sending a response,
```

omoerbeek (approved):
This looks good with small nit. Did you have a chance to provoke the case in testing?

rgacogne:
> This looks good with small nit. Did you have a chance to provoke the case in testing?

Yes, it's very easy to reproduce with dnsdist, or any other tool that wait for the answer to come back before sending a new query.

PR #9492, opened 2020-09-21
rec: Detach snmp thread to avoid trouble when trying to quit nicely.

omoerbeek:
This avoids a case where the thread object and the RecursorSNMPAgent
object get destroyed in the wrong order.

### Short description

```#0  thrkill () at /tmp/-:3
#1  0x000008fc7b1a55ee in _libc_abort () at /usr/src/lib/libc/stdlib/abort.c:51
#2  0x000008fc60181b3c in abort_message (format=<optimized out>) at /usr/src/lib/libcxxabi/src/abort_message.cpp:77
#3  0x000008fc60181f5f in demangling_terminate_handler () at /usr/src/lib/libcxxabi/src/cxa_default_handlers.cpp:76
#4  0x000008fc6016153f in std::__terminate (func=0x0) at /usr/src/lib/libcxxabi/src/cxa_handlers.cpp:60
#5  0x000008fc601614a9 in std::terminate () at /usr/src/lib/libcxxabi/src/cxa_handlers.cpp:91
#6  0x000008fc23cbc835 in std::__1::thread::~thread (this=0x0) at /usr/src/lib/libcxx/src/thread.cpp:44
#7  0x000008f9e1d8d0df in std::__1::__shared_count::__release_shared (this=0x8fc34474d00)
    at /usr/include/c++/v1/memory:3544
#8  std::__1::__shared_weak_count::__release_shared (this=0x8fc34474d00) at /usr/include/c++/v1/memory:3586
#9  std::__1::shared_ptr<RecursorSNMPAgent>::~shared_ptr (this=<optimized out>) at /usr/include/c++/v1/memory:4522
#10 0x000008fc7b207de3 in _libc___cxa_finalize (dso=<optimized out>) at /usr/src/lib/libc/stdlib/atexit.c:177
#11 0x000008fc7b1d82c1 in _libc_exit (status=0) at /usr/src/lib/libc/stdlib/exit.c:54
#12 0x000008f9e1ad9588 in ___start ()
#13 0x0000000000000000 in ?? ()
```

### Checklist

I have:
- [X] read the [CONTRIBUTING.md](https://github.com/PowerDNS/pdns/blob/master/CONTRIBUTING.md) document
- [X] compiled this code
- [X] tested this code
- [ ] included documentation (including possible behaviour changes)
- [ ] documented the code
- [ ] added or modified regression test(s)
- [ ] added or modified unit test(s)

PR #9496, opened 2020-09-21
auth: ignore cryptokeys in presigned zones

### LINKED ISSUE #9489: Confusion when DNSSEC key(s) exist for PRESIGNED zone
- Program: Authoritative
 - Issue type: Bug report

### Short description
If a PRESIGNED DNSSEC secured zone has key(s) attached to it, they are used to sign the zone. This is not indicated by `pdnsutil show-zone` output which shows only the AXFR'd keys from primary. To figure out you have such key(s) one has to do `pdnsutil list-keys`.

### Environment
 - Operating system: Debian 9.13
 - Software version:  PowerDNS Authoritative Server 4.2.2
 - Software source: PowerDNS

### Steps to reproduce
1. Add keys for a zone
2. Configure zone as slave
3. AXFR presigned zone

### Expected behaviour
Not sure what should happen, but at least `pdnsutil show-zone` should indicate what the keys are that are used.

### Actual behaviour
PowerDNS tools indicate that zone is using presigned data, and correctly show the presigned key from master. Yet the server uses existing DNSSEC keys to live sign the record data.

### Other information
Supporting live signing on slaves could be used to avoid an extra unsigned master, and could be used to support NSEC3 NARROW slaving?

mind04:
### Short description
Fixes #9489

### Checklist
I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


mnordhoff:
> Fixes #9492 

That should be fixes #9489, right?

mind04:
oops, updated

cmouse:
Can we add also feature that will warn about this with `pdnsutil show-zone`? Something like `cryptokeys found for presigned zone`?

pdns/tcpreceiver.cc
```diff
@@ -609,70 +609,73 @@ int TCPNameserver::doAXFR(const DNSName &target, std::unique_ptr<DNSPacket>& q,
...
+    for (auto const& value : entryPoints)
+      entryPointIds.insert(value.second.id);
+
+    for(const DNSSECKeeper::keyset_t::value_type& value :  keys) {
+      if (!value.second.published) {
+        continue;
+      }
+      zrr.dr.d_type = QType::DNSKEY;
+      zrr.dr.d_content = std::make_shared<DNSKEYRecordContent>(value.first.getDNSKEY());
+      DNSName keyname = NSEC3Zone ? DNSName(toBase32Hex(hashQNameWithSalt(ns3pr, zrr.dr.d_name))) : zrr.dr.d_name;
```
cmouse, line 637: `const DNSName&`

pdns/tcpreceiver.cc
```diff
@@ -609,70 +609,73 @@ int TCPNameserver::doAXFR(const DNSName &target, std::unique_ptr<DNSPacket>& q,
...
+      NSECXEntry& ne = nsecxrepo[keyname];
```
cmouse, line 638: `const NSECXEntry&`

cmouse (changes requested):

PR #9504, opened 2020-09-22
rec: Better const-ness when dealing with timestamps

rgacogne:
### Short description

This PR should have no functional impact, it just makes reviewing the code easier. It also saves a `gettimeofday()` call in `DTime::udiff()` which should not have any meaningful impact either.

### Checklist

I have:
- [x] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/misc.hh
+++ b/pdns/misc.hh
@@ -196,8 +196,12 @@ public:
   DTime & operator=(const DTime &dt) = default;
   time_t time();
   inline void set();  //!< Reset the timer
-  inline int udiff(); //!< Return the number of microseconds since the timer was last set.
-  inline int udiffNoReset(); //!< Return the number of microseconds since the timer was last set.
+  inline int udiff(bool reset = true); //!< Return the number of microseconds since the timer was last set.
+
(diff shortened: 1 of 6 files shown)


pdns/misc.hh
```diff
@@ -215,19 +219,17 @@ inline void DTime::set()
   gettimeofday(&d_set,0);
```
omoerbeek, line 219:
```suggestion
  gettimeofday(&d_set, nullptr);
```

pdns/misc.hh
```diff
@@ -196,8 +196,12 @@ public:
   DTime & operator=(const DTime &dt) = default;
   time_t time();
```
omoerbeek, line 197: can be const and inline instead of in `time.cc` file

omoerbeek (approved):
Further small nits:

`getTimeval()` can be const
d_set can be initialized using {0, 0} and the default ct can be `= default`

rgacogne:
I pushed a new commit with the changes you suggested, thanks a lot!

PR #9518, opened 2020-09-24
auth: make sure the metadata cache is not filled from within an api transaction

mind04:
### Short description
And use rectify for full zone rectifies in the rfc2136 code

### Checklist
I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/ws-auth.cc
+++ b/pdns/ws-auth.cc
@@ -331,7 +331,9 @@ static Json::object getZoneInfo(const DomainInfo& di, DNSSECKeeper* dk) {
   };
   if (dk) {
     obj["dnssec"] = dk->isSecuredZone(di.zone);
-    obj["edited_serial"] = (double)calculateEditSOA(di.serial, *dk, di.zone);
+    string soa_edit;
+    dk->getSoaEdit(di.zone, soa_edit, false);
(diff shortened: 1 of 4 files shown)


pdns/ws-auth.cc
```diff
@@ -737,7 +740,7 @@ static void updateDomainSettingsFromDocument(UeberBackend& B, const DomainInfo&
         if (!dk.unSecureZone(zonename, error, info)) {
           throw ApiException("Error while un-securing zone '"+ zonename.toString()+"': " + error);
         }
-        isDNSSECZone = dk.isSecuredZone(zonename);
+        isDNSSECZone = dk.isSecuredZone(zonename, false);
```
zeha, line 743: there's a similar call in the other branch (line 726), that could use it as well?

mind04: The short answer is no. The longer version is...
There was no metadata update inside the transaction at line 726, so the cache is valid and it is also fine to fill the cache when expired.
Here the unsetNSEC3PARAM(); and unsetPresigned(); in unSecureZone() are responsible for a write operation inside the transaction, so the isPresigned() in side isSecuredZone() is no longer allowed to read/update the cache.

PR #9534, opened 2020-09-25
enable code scanning

Habbie:
### Short description
This should enable GitHub Code Scanning on all branches, all PRs, plus periodically.

This needs #9583 merged first.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,0 +1,118 @@
+name: "CodeQL"
+
+on:
+  push:
+  pull_request:
+  schedule:
+    - cron: '0 22 * * 2'
+
+jobs:
+  analyze:
+    name: Analyze
+    runs-on: ubuntu-latest
+
+    strategy:
+      fail-fast: false
+      matrix:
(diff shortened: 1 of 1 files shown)


Habbie:
Right now this builds our code.. six times. Also for Python and JS scanning. Probably don't want the language matrix at all.

zeha:
Might be good to keep the larger check set - I'm hoping in a PR it would only show new alerts?

Habbie:
> Might be good to keep the larger check set

The options as I see them:

1. start small, clean out the large set, then go large
2. go large now and triage them all before we start scanning PRs
3. go large, accept that the list is big, clean up at some point

Having written it out like that, perhaps 3 is reasonable after all.

> I'm hoping in a PR it would only show new alerts?

I also assume so!

Habbie:
> Having written it out like that, perhaps 3 is reasonable after all.

I have decided I like 3. We'll find out how noisy it is on PRs.

Habbie:
#9583 was merged so I rebased this.

PR #9550, opened 2020-09-28
dnsrecords: fix APL test on macOS

zeha:
### Short description

macOS does not life fe80:: without an %interface specifier, so inet_pton
fails, and we end with an error like this:

```
check rec->getZoneRepresentation() == zoneval has failed [2:invalid Non-recoverable failure in name resolution/128 != 2:fe80:1234:5678:9910:8bc:3359:b2e8:720e/128]
```

Fix test by switching to a "normal" internet class prefix.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


stbuehler:
Given `inet_pton` can't properly return the scope id it really shouldn't require one...

Also this sounds rather bad; afaict this means `AAAA` won't work with link-local addresses on macOS either (`ComboAddress` -> `makeIPv6sockaddr` -> `inet_pton` with fallback `getaddrinfo`).

Found this with more (old) background: https://stackoverflow.com/a/13826646/1478356 (which claims back then unscoped `inet_pton` should have worked).

Perhaps a proper fix could workaround broken `inet_pton` implementations by replacing the leading `fe80:` with something else, and then reverthing that in the result?

I think it would be good to keep link-local address tests around - I think you really want to know if this breaks on other platforms as well.

stbuehler:
Actually: if the address begins with `fe80:` (case insensitive) one can simply skip the first 4 characters; the resulting address should be valid extactly when the original was valid, and simply contain a zero as first word instead of 0xfe80.

So this should be "easy" to wrap in C without allocating new strings.

zeha:
I had a quick look at inet_pton/inet_ntop on macos earlier today, and that doesn't seem to be the problem. Maybe ComboAddress calls some other "bad" function internally. Should probably investigate a bit more.

PR #9562, opened 2020-09-30
rec: Expose typed cache flush via Web API

omoerbeek:
### Checklist

I have:
- [X] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/recursordist/docs/http-api/endpoint-cache.rst
+++ b/pdns/recursordist/docs/http-api/endpoint-cache.rst
@@ -12,6 +12,10 @@ Cache manipulation endpoint
 
   :query subtree: If set to `true`, also flush the whole subtree (default = `false`)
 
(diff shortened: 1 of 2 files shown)


pdns/recursordist/docs/http-api/endpoint-cache.rst
```diff
@@ -12,6 +12,10 @@ Cache manipulation endpoint
 
   :query subtree: If set to `true`, also flush the whole subtree (default = `false`)
 
+  .. versionadded:: 4.4.0
+
+  :query type: If set only flush records of the specified type name.
```
zeha, line 17: Needs a hint that this works only in the recursor, not auth

omoerbeek: It's rec specific doc, but fair enough.

PR #9564, opened 2020-09-30
el6: set __STDC_FORMAT_MACROS during package builds

Habbie:
### Short description
Tested with build.sh. I did not check if the resulting binaries run and do the right thing in the affected bit of code.

### Checklist

I have:
- [x] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/builder-support/specs/pdns.spec
+++ b/builder-support/specs/pdns.spec
@@ -223,7 +223,8 @@ export CPPFLAGS="-DLDAP_DEPRECATED"
   --disable-lua-records \
   --without-protobuf \
   --with-boost=/usr/include/boost148/ LDFLAGS=-L/usr/lib64/boost148 \
-  CXXFLAGS=-std=gnu++11
+  CXXFLAGS=-std=gnu++11 \
+  CPPFLAGS="${CPPFLAGS} -D__STDC_FORMAT_MACROS"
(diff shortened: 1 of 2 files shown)


builder-support/specs/pdns.spec
```diff
@@ -223,7 +223,8 @@ export CPPFLAGS="-DLDAP_DEPRECATED"
   --disable-lua-records \
   --without-protobuf \
   --with-boost=/usr/include/boost148/ LDFLAGS=-L/usr/lib64/boost148 \
-  CXXFLAGS=-std=gnu++11
+  CXXFLAGS=-std=gnu++11 \
+  CPPFLAGS=-D__STDC_FORMAT_MACROS
```
rgacogne, line 227: Doesn't this override the `CPPFLAGS="-DLDAP_DEPRECATED"` value set just above? Perhaps it would be better to append to `CPPFLAGS` instead of setting it?

Habbie: thanks!

Habbie: Fixing that, although apparently the build does not need it to complete :)

Habbie: amended

PR #9571, opened 2020-10-01
rec: Throttle servers sending invalid data and rcodes

rgacogne:
### Short description

Instead of throttling explicitly on ServFail and Refused, do it for all rcodes except `NoError` and `NXDomain`, with the exception of `FormError` which is handled earlier.
Also throttle servers sending data that we fail to parse.

This PR is huge because it adds a `LWResult::Result` type to stop dealing with numerical values all over the place, sorry about that.
Need more tests if the general idea is deemed correct.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/lwres.cc
+++ b/pdns/lwres.cc
@@ -220,11 +220,10 @@ static void logIncomingResponse(const std::shared_ptr<std::vector<std::unique_pt
 }
 #endif /* HAVE_PROTOBUF */
 
-//! returns -2 for OS limits error, -1 for permanent error that has to do with remote **transport**, 0 for timeout, 1 for success
 /** lwr is only filled out in case 1 was returned, and even when returning 1 for 'success', lwr might contain DNS errors
     Never throws! 
  */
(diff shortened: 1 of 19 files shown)


omoerbeek:
I like the general idea & direction. Reviewing is kinda hard because the various conditions both change because of the new symbolic constants (that have different values compared to the old literal ones) *and* because the PR introduces behavioural changes.

rgacogne:
Note that I tried to separate the symbolic constants changes from the other ones in distinct commits, so you could try reviewing them independently?  I could also open two separate PRs if you think it would help.

omoerbeek:
Ah, I was looking at the combined result. Looking at the separate commits will make it easer, yes.

omoerbeek:
This looks good indeed. But as you already mention, tests would really be a good thing.

rgacogne:
Rebased to fix a conflict, and added unit tests! It seems ready for a review now :)

omoerbeek (approved):

PR #9572, opened 2020-10-01
rec: Terminate TCP connections instead of 'ignoring' errors

rgacogne:
### Short description

We used to ignore questions that we consider invalid (unexpected opcode, qdcount != 1, QR=1, parse error, ...) but also those
received from source addresses blocked by ipfilter, still waiting for a new question to come up on the socket.
That might be fine for clients that will keep sending queries, even though they will still end up wondering what happened to the ignored queries, but some clients like dnsdist will wait until a response is sent, or a time out occurs.
Closing the TCP connection instead allows dnsdist to keep going, possibly retrying over a new connection but finally giving up,
instead of keeping the connection alive.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/pdns_recursor.cc
+++ b/pdns/pdns_recursor.cc
@@ -752,6 +752,16 @@ TCPConnection::~TCPConnection()
 
 AtomicCounter TCPConnection::s_currentConnections;
 
+static void terminateTCPConnection(int fd)
+{
+  try {
+    t_fdm->removeReadFD(fd);
+  }
+  catch (const FDMultiplexerException& fde)
+  {
+  }
+}
+
 static void handleRunningTCPQuestion(int fd, FDMultiplexer::funcparam_t& var);
 
 // the idea is, only do things that depend on the *response* here. Incoming accounting is on incoming.
@@ -1882,11 +1892,7 @@ static void startDoResolve(void *p)
       // "Tried to remove unlisted fd" exception.  Not that an inflight < limit test
       // will not work since we do not know if the other mthread got an error or not.
       if(hadError) {
-        try {
-          t_fdm->removeReadFD(dc->d_socket);
-        }
-        catch (FDMultiplexerException &) {
-        }
+        terminateTCPConnection(dc->d_socket);
         dc->d_socket = -1;
       }
       else {
@@ -2122,12 +2128,12 @@ static bool handleTCPReadResult(int fd, ssize_t bytes)
 {
   if (bytes == 0) {
     /* EOF */
-    t_fdm->removeReadFD(fd);
+    terminateTCPConnection(fd);
     return false;
   }
   else if (bytes < 0) {
     if (errno != EAGAIN && errno != EWOULDBLOCK) {
-      t_fdm->removeReadFD(fd);
+      terminateTCPConnection(fd);
       return false;
     }
   }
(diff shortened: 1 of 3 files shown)


paddg:
Do you think this could improve DoT benchmarks? I've made the observation that benchmarks looks much better when the queries are chosen in such a way that they will be answered safely and no timeouts occurres

rgacogne:
> Do you think this could improve DoT benchmarks? I've made the observation that benchmarks looks much better when the queries are chosen in such a way that they will be answered safely and no timeouts occurres

That would very much depend on the benchmarks but I believe it would improve some of them, yes. As a matter of cat I was looking at reproducing the results from the CZ.NIC benchmarks when I noticed this :)

pspacek:
For the record we (CZ.NIC) are benchmarking with traffic captures from real-world traffic. In other words all the garbage in the data sample was in practice really sent to DNS resolvers.

pspacek:
Ah, I forgot about the main point: It might be better to send back FORMERR so the client/proxy, it would allow to reuse the connection open for other queries (= save connection setup/teardown overhead).

Of course there is a thin line between weird queries (qdcount != 1) which warrant FORMERR and total crap which ought to be completely ignored.

rgacogne:
> For the record we (CZ.NIC) are benchmarking with traffic captures from real-world traffic. In other words all the garbage in the data sample was in practice really sent to DNS resolvers.

Understood! And to be fair we have seen much worse than that, that's actually the reason we drop a lot of garbage as early as possible :-)

rgacogne:
> Ah, I forgot about the main point: It might be better to send back FORMERR to the client/proxy, it would allow to reuse the connection open for other queries (= save connection setup/teardown overhead).
> Of course there is a thin line between weird queries (qdcount != 1) which warrant FORMERR and total crap which ought to be completely ignored or treated as unrecoverable error.

Agreed, it would be nicer. I'm pondering doing that for some cases (opcode == UPDATE or opcode == NOTIFY, QDCOUNT == 0) but we can't do it for a number of cases where we our parser just bail out too early for us to send a response (qdcount > 1, opcode other than QUERY, UPDATE or NOTIFY. We also don't want to do it in some cases, like QR=1, because we have seen some weird case of endless loops between servers in that particular case.

omoerbeek (approved):

rgacogne:
>  I'm pondering doing that for some cases (opcode == UPDATE or opcode == NOTIFY, QDCOUNT == 0) but we can't do it for a number of cases where we our parser just bail out too early for us to send a response (qdcount > 1, opcode other than QUERY, UPDATE or NOTIFY.

I have a commit doing just that (sending NOTIMP for `UPDATE`, `NOTIFY` and `qdcount == 0`). Shall I push it to that branch or open a new pull request once that one has been merged? Or do we not want that at all?

omoerbeek:
This branch is OK imo, since it is very much related.

rgacogne:
> This branch is OK imo, since it is very much related.

Pushed!

pdns/pdns_recursor.cc
```diff
@@ -762,6 +762,67 @@ static void terminateTCPConnection(int fd)
...
+         "The minimal response MUST be the DNS header, question section, and an
+         OPT record.  This MUST also occur when a truncated response (using
+         the DNS header's TC bit) is returned."
+      */
+      pw.addOpt(512, 0, 0);
+      pw.commit();
+    }
+  }
+
+  dnsheader& header = reinterpret_cast<dnsheader&>(packet.at(0));;
```
omoerbeek, line 813:
```suggestion
  dnsheader& header = reinterpret_cast<dnsheader&>(packet.at(0));
```

omoerbeek (approved):

PR #9573, opened 2020-10-01
rec draft: Lua/C++ teamwork for MX additionals

Habbie:
### Short description
With the packetcache disabled (could probably do something smarter with `dq.variable`) this allows filling IPs (v4) in ADDITIONALS for MX records in ANSWER, strictly from the recursor cache.

Unfinished; posting this to see if we feel something that has this shape could still go into rec 4.4. Then for 4.5 (perhaps also with some more experience from this small thing) we could do it better in (mostly) C++.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/lua-recursor4.cc
+++ b/pdns/lua-recursor4.cc
@@ -24,6 +24,7 @@
 #include "logger.hh"
 #include "dnsparser.hh"
 #include "syncres.hh"
+#include "resolver.hh"
 #include "namespaces.hh"
 #include "rec_channel.hh"
 #include "ednsoptions.hh"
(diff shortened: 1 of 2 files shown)


omoerbeek:
I like the general idea, but way have to be careful about the API (we might want to expose more of the cache info available) and the exact functionality: ATM AFAKS the code is happy to add Bogus records. 

Given these, I do not think we should not push it to make 4.4.0.

Habbie:
> I like the general idea, but way have to be careful about the API (we might want to expose more of the cache info available) and the exact functionality: ATM AFAKS the code is happy to add Bogus records.

Yes, I later realised at least TTL was missing, and indeed, I was unsure how solid this data is.

> Given these, I do not think we should not push it to make 4.4.0.

:+1:

omoerbeek:
For the standard cases superseded by #11302 (pure C++ implementation).

Habbie:
Yes, closing this. Thanks!

PR #9574, opened 2020-10-01
Ensure socket-dir matches runtimedir on old systemd

### LINKED ISSUE #9485: rec: Different runtime dirs for virtual services (#9397) introduced slight incompatibility with older systemd releases
- Program: Recursor
 - Issue type: Bug report

### Short description
#9397 (based on #9073) introduced different runtime directories for virtual services. systemd then creates runtime directories in the format _/var/run/pdns-recursor-&lt;instance&gt;_ however this relies on the _RUNTIME_DIRECTORY_ environment variable being set by systemd to be properly passed to recursor. On systemd releases &lt;240 such as on Ubuntu 18.04 this is not set causing recursor to default to _/var/run/pdns-recursor_ instead of _pdns-recursor-&lt;instance&gt;_.
This causes recursor to exit with a socket-directory problem error and needs to be manually fixed, for example by specifying `socket-dir` in the instances configuration - which is unexpected.

### Environment

 - Operating system: Ubuntu 18.04.5 LTS
 - Software version: 4.3.4
 - Software source: PowerDNS repo, pdns-recursor 4.3.4-1pdns.bionic

### Steps to reproduce

1. Start virtual service instance on systemd &lt;240, e.g. Ubuntu 18.04, without specifying `socket-dir`
2. (Compare to 4.3.3 behavior - but yes, this had other issues)

### Expected behaviour
Recursor running just fine.

### Actual behaviour
Recursor quits:
`Problem with socket directory /var/run/pdns-recursor: it does not exist or cannot access; see https://docs.powerdns.com/recursor/upgrade.html#x-to-4-3-0-or-master`

### Other information
As discussed with the team it's not sure if this can be fixed or would just get documented as known issue with virtual hosting / virtual services under older systemd releases.

pieterlexis:
### Short description
This is mostly a hack for systemd version <240 so using virtual hosting
on Ubuntu Bionic and EL7 does not lead to surprises.

This Commit explicitly adds the `--socket-dir` to the invocation of the
unit. Any users who want to use their own runtimedirs should use a
drop-in unit overriding ExecStart. I believe this does not validate the
principle of least surprise for those using the PowerDNS provided
packages and virtual hosting.

Fixes #9485

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/Makefile.am
+++ b/pdns/Makefile.am
@@ -1607,6 +1607,9 @@ dnsdist:
 if HAVE_SYSTEMD
 pdns.service: pdns.service.in
 	$(AM_V_GEN)sed -e 's![@]sbindir[@]!$(sbindir)!' -e 's![@]service_user[@]!$(service_user)!' -e 's![@]service_group[@]!$(service_group)!' < $< > $@
+if !HAVE_SYSTEMD_WITH_RUNTIME_DIR_ENV
+	$(AM_V_GEN)sed -e 's!/pdns_server!& --socket-dir=%t/pdns!' -i $@
+endif
 if !HAVE_SYSTEMD_LOCK_PERSONALITY
 	$(AM_V_GEN)perl -ni -e 'print unless /^LockPersonality/' $@
 endif
@@ -1668,6 +1671,9 @@ pdns@.service: pdns.service
 	  -e 's!RuntimeDirectory=.*!&-%i!' \
 	  -e 's!SyslogIdentifier=.*!&-%i!' \
 	  < $< > $@
+if !HAVE_SYSTEMD_WITH_RUNTIME_DIR_ENV
+	$(AM_V_GEN)sed -e 's!--socket-dir=[^ ]\+!&-%i !' -i $@
+endif
 
 systemdsystemunitdir = $(SYSTEMD_DIR)
 
--- a/pdns/recursordist/Makefile.am
+++ b/pdns/recursordist/Makefile.am
@@ -517,6 +517,9 @@ endif
 if HAVE_SYSTEMD
 pdns-recursor.service: pdns-recursor.service.in
 	$(AM_V_GEN)sed -e 's![@]sbindir[@]!$(sbindir)!' -e 's![@]service_user[@]!$(service_user)!' -e 's![@]service_group[@]!$(service_group)!' < $< > $@
+if !HAVE_SYSTEMD_WITH_RUNTIME_DIR_ENV
+	$(AM_V_GEN)sed -e 's!/pdns_recursor!& --socket-dir=%t/pdns-recursor!' -i $@
+endif
 if !HAVE_SYSTEMD_LOCK_PERSONALITY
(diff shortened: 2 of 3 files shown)


omoerbeek:
I possible, I would prefer a runtime check instead of a compile time check.

pieterlexis:
> I possible, I would prefer a runtime check instead of a compile time check.

The only runtime check is 'is this env-var set', but if it is *not* set, it can mean 2 things: we're running outside of systemd (i.e. started on the shell) or we're running in an old systemd version

zeha:
Manually applying this to the service unit file makes 4.4.0-beta1 start for me on Ubuntu bionic. Would be great if this could make it into 4.4.0!

pdns/Makefile.am
```diff
@@ -1570,6 +1570,9 @@ dnsdist:
 if HAVE_SYSTEMD
 pdns.service: pdns.service.in
 	$(AM_V_GEN)sed -e 's![@]sbindir[@]!$(sbindir)!' -e 's![@]service_user[@]!$(service_user)!' -e 's![@]service_group[@]!$(service_group)!' < $< > $@
+if !HAVE_SYSTEMD_WITH_RUNTIME_DIR_ENV
+	$(AM_V_GEN)sed -e 's!/pdns_server!& --socket-dir=/var/run/pdns-%i!' -i $@
```
ekohl, line 1574: Also noting it here: there is `%t` as variable for the runtime directory.

pieterlexis: Implemented

pieterlexis:
I've updated this PR to have the non-@ service use `%t/pdns` and the @-service `%t/pdns-%i`, which now matched `RuntimeDirectory`

pdns/recursordist/Makefile.am
```diff
@@ -517,6 +517,9 @@ endif
 if HAVE_SYSTEMD
 pdns-recursor.service: pdns-recursor.service.in
 	$(AM_V_GEN)sed -e 's![@]sbindir[@]!$(sbindir)!' -e 's![@]service_user[@]!$(service_user)!' -e 's![@]service_group[@]!$(service_group)!' < $< > $@
+if !HAVE_SYSTEMD_WITH_RUNTIME_DIR_ENV
+	$(AM_V_GEN)sed -e 's!/pdns_recursor!& --socket-dir=/var/run/pdns-recursor-%i!' -i $@
```
ekohl, line 521: Should this get the same treatment as auth and use `%t` + drop `%-i` from the single instance.

pieterlexis: yes! I totally missed the rec

Habbie:
@zeha can you review/test this again please?

zeha (approved):
I have tried the Auth with these changes on Ubuntu bionic and can confirm it works!

zeha:
I would like to see a backport to auth-4.4, if possible.

Habbie:
> I would like to see a backport to auth-4.4, if possible.

Yes!

PR #9578, opened 2020-10-02
Fix error checking on one getaddrinfo call

zeha:
### Short description
Noticed this while looking at something related.

`getaddrinfo(3)` documents the return values as only zero meaning success. Indeed all other `getaddrinfo` call sites already check this correctly.

### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/notify.cc
+++ b/pdns/notify.cc
@@ -97,7 +97,7 @@ try
        parts.push_back("domain");
      else if (parts.size() != 2)
        throw runtime_error("Invalid hostname:port syntax");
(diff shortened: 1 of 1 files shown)


rgacogne (approved):
It looks like all error values are negative on Linux, but that's not guaranteed at all.

Habbie (approved):

omoerbeek:
The BSDs also use negative values, but the fix is correct anyway.

PR #9580, opened 2020-10-02
Support serving APL records pointing to fe80:: on macOS

zeha:
### Short description
Improves PR #9550. Thanks to @stbuehler for nagging "this cant really be the case".

While `ComboAddress` tries hard to avoid `getaddrinfo()`, in `toString()` it calls `getnameinfo()`, which shares the same problems. As `ComboAddress` callers might need the interface qualifiers, I've added a new `toStringNoInterface()`, which is used by `class Netmask` now. (APL uses `Netmask` internally.)

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


rgacogne:
I'm not sure I understand what caused the initial issue? #9550 stated that "macOS does not life fe80:: without an %interface specifier, so inet_pton fails" but this PR since to fix an issue when the interface specifier is there instead?
I wonder if we could still use `getnameinfo()` in that case by setting `sin6_scope_id` to 0 first, since POSIX hints that this is where the interface index should be. Although not all implementations might have taken the hint..

zeha:
In #9550 I was mistaken: inet_pton doesn't fail, we never call it (the code path is hard to follow).

But really, we try a roundtrip from a string address (without a %interface), using inet_ntop and getnameinfo. And getnameinfo fails. inet_pton actually succeeds (and I think it cannot do %interface). I think this is fine (actually correct?) for addresses we're treating as data (DNS records).

For addresses used as local configuration, we need to preserve the %interface specifier if given. Now if a local admin specifies fe80:: without %interface on macOS, it will probably fail. But thats a (good?) limitation which I don't want to work around.

I thought about resetting `sin6_scope_id`, but I don't think it would make a difference here (judging by Darwins inet_ntop source).

rgacogne (approved):

PR #9582, opened 2020-10-02
dnsdist: Implement TCP out-of-order

rgacogne:
### Short description

This PR implements TCP out-of-order support for dnsdist:
- for incoming queries (between the client and dnsdist) ;
- as well as for outgoing queries (between dnsdist and the backend).

We try to reuse the existing TCP connections as much as possible, but we might need to open many connections to the same backend, especially if that backend does not support out-of-order. We also don't mix queries from different incoming TCP connections on the same TCP connection to the backend, since that would require taking care of query ID collisions, which we currently detect. An existing TCP connection to the backend can be re-used for a different incoming connection later, though.

This PR break the TCP handling code in two parts, one handling incoming TCP connections and a second one handling outgoing ones.

It still needs a serious review and more testing.

DNS Shotgun seems to be quite happier with it, by the way:
master
oor

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code


paddg:
How many TCP connections did you simulate during your 6kqps test?

rgacogne:
Roughly 10k:

connection

paddg:
Very good! Did you test with real DNS queries or with synthetic? If I test with real DNS traffic, the response rate is below 60%

rgacogne:
> Did you test with real DNS queries or with synthetic? If I test with real DNS traffic, the response rate is below 60%

Yes, that test is done with the data set mentioned in #9572, which is real-word traffic. Note that the test is done over DNS over TLS, and I tuned a few things:
- `setTCPRecvTimeout(30)`: read time out set to 30s ;
- `setMaxTCPClientThreads(8)`: 8 TCP worker threads ;
- 2 * `addTLSLocal("[::1]:8443", 'server.pem', 'server.key', { reusePort=true}, numberOfStoredSessions=0, maxInFlight=5 })`: 2 threads doing `accept()` ;
- 8 * `newServer({ address = "[::1]", maxInFlight=10})`: 8 backend threads ;

On the recursor side, both #9495 and #9572 are applied, and I have:
- `client-tcp-timeout=30` ;
- `dnssec=validate` ;
- `max-mthreads=8192` ;
- `max-tcp-clients=500000` ;
- `pdns-distributes-queries=no` ;
- `qname-minimization=yes` ;
- `reuseport=yes` ;
- `threads=8`.

rgacogne:
> Is this meant as unlimited? Because "Setting this value to 0 disables stored session entirely"

No, it is intended to disable sessions stored server-side, since client-side tickets are much nicer to the server and more realistic in anycast setups anyway, since the session-ticket key can be shared between servers much more easily than the server-side sessions.

rgacogne:
Squashed to clean up the history a bit, and to make sure all commits build and pass the unit tests.

pdns/dnsdistdist/dnsdist-tcp-downstream.cc
```diff
@@ -0,0 +1,543 @@
...
+
+      try {
+        if (conn->reconnect()) {
+          conn->d_ioState = make_unique<IOStateHandler>(conn->d_clientConn->getIOMPlexer(), conn->d_socket->getHandle());
+
+          /* we need to resend the queries that were in flight, if any */
+          for (auto& pending : conn->d_pendingResponses) {
+            conn->d_pendingQueries.push_back(std::move(pending.second));
+            if (!conn->d_usedForXFR) {
+              --conn->d_ds->outstanding;
```
omoerbeek, line 224: Maybe a >0 check here to detect accounting errors wrt `outstanding`?

pdns/dnsdistdist/dnsdist-tcp-downstream.cc
```diff
@@ -0,0 +1,543 @@
...
+
+  notifyAllQueriesFailed(now, FailureReason::timeout);
+}
+
+void TCPConnectionToBackend::notifyAllQueriesFailed(const struct timeval& now, FailureReason reason)
+{
+  d_connectionDied = true;
+
+  if (!d_usedForXFR) {
+    d_ds->outstanding -= d_pendingResponses.size();
```
omoerbeek, line 405: Maybe a >= d_pendingResponses.size() check here to detect accounting errors wrt outstanding?

pdns/dnsdistdist/dnsdist-tcp-downstream.cc
```diff
@@ -0,0 +1,543 @@
...
+  d_downstreamFailures = 0;
+
+  auto& clientConn = d_clientConn;
+  if (!clientConn || !clientConn->active()) {
+    // a client timeout occured, or something like that */
+    d_connectionDied = true;
+    d_clientConn.reset();
+
+    if (!conn->d_usedForXFR) {
+      --conn->d_ds->outstanding;
```
omoerbeek, line 451: And here...

pdns/dnsdistdist/dnsdist-tcp-downstream.cc
```diff
@@ -0,0 +1,543 @@
...
+
+    auto it = d_pendingResponses.find(queryId);
+    if (it == d_pendingResponses.end()) {
+      DEBUGLOG("could not find any corresponding query for ID "<<queryId<<". This is likely a duplicated ID over the same TCP connection, giving up!");
+      notifyAllQueriesFailed(now, FailureReason::unexpectedQueryID);
+      return IOState::Done;
+    }
+
+    if (!conn->d_usedForXFR) {
+      --conn->d_ds->outstanding;
```
omoerbeek, line 488: here

pdns/dnsdistdist/dnsdist-tcp-downstream.cc
rgacogne: We have tried that in the past, and unfortunately it did not really help because it covered issues up for light loads, while the race between the test and the decrement was happening quite frequently at high loads.

pdns/dnsdistdist/dnsdist-tcp-upstream.hh
```diff
@@ -0,0 +1,229 @@
...
+  }
+
+  ~ConnectionInfo()
+  {
+    if (fd != -1) {
+      close(fd);
+      fd = -1;
+    }
+    if (cs) {
+      --cs->tcpCurrentConnections;
```
omoerbeek, line 48: There's some asymmetry here; the increment happens in `tcpAcceptorThread`, but that's because it's only done after the accept I suppose?

pdns/dnsdistdist/dnsdist-tcp-downstream.cc
omoerbeek: right. making it safe would be  more complex indeed.

pdns/dnsdistdist/dnsdist-tcp-upstream.hh
rgacogne: Yes, the object is allocated before the call to `accept()` so we can directly store the descriptor there, but we only want to increment after the connection has been accepted.

omoerbeek (approved):
It's quite a big PR, so I won't state I looked at every line of code. But what I checked looks good and basic testing show it works as advertised once I git the settings right.

related to that: please document  that for it to work you should set both incoming and outgoing `maxInFlight`.

rgacogne:
> It's quite a big PR, so I won't state I looked at every line of code. But what I checked looks good and basic testing show it works as advertised once I git the settings right.

I did not expect you to look at every single detail, don't worry! :-) Thanks a lot for the review and tests!

> related to that: please document that for it to work you should set both incoming and outgoing maxInFlight.

I just added more words on how that feature works, that was clearly lacking!

PR #9588, opened 2020-10-05
circleci: authenticate to Docker Hub

Habbie:
### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -494,6 +494,9 @@ jobs:
 
     docker:
       - image: debian:buster
+        auth:
+          username: powerdnsreadonly
+          password: $DOCKERHUB_PASSWORD
 
     steps:
       - checkout-shallow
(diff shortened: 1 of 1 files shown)


pieterlexis:
This might break on forks that use circleci?

Habbie:
> This might break on forks that use circleci?

I'm surprised it even works on the PR. There is no log output indicating whether it's logging in or not, succeeding to do so, or falling back to not logged in.

Very tempted to ignore this until the problem kicks in and we can actually tell the difference between 'login worked' and 'ratelimit hit because one million other CircleCI users did not log in'.

Habbie:
I've reverted the intentional breakage (that broke nothing -inside the PR-) and am marking this as draft so we can have it ready to go on November 1st.

PR #9590, opened 2020-10-06
auth: Rewrite OpenSSLRSADNSCryptoKeyEngine::fromISCMap to be less tricky

omoerbeek:
### Short description

Prompted by CodeQL

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [x] documented the code

### DIFF
--- a/pdns/opensslsigners.cc
+++ b/pdns/opensslsigners.cc
@@ -194,7 +194,7 @@ public:
   bool verify(const std::string& hash, const std::string& signature) const override;
   std::string getPubKeyHash() const override;
   std::string getPublicKeyString() const override;
-  std::unique_ptr<BIGNUM, void(*)(BIGNUM*)>parse(std::map<std::string, std::string>& stormap, const std::string& key);
+  std::unique_ptr<BIGNUM, void(*)(BIGNUM*)>parse(std::map<std::string, std::string>& stormap, const std::string& key) const;
   void fromISCMap(DNSKEYRecordContent& drc, std::map<std::string, std::string>& stormap) override;
   void fromPublicKeyString(const std::string& content) override;
   bool checkKey(vector<string> *errorMessages) const override;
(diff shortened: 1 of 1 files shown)


pdns/opensslsigners.cc
```diff
@@ -429,88 +430,44 @@ std::string OpenSSLRSADNSCryptoKeyEngine::getPublicKeyString() const
 }
 
 
+std::unique_ptr<BIGNUM, void(*)(BIGNUM*)>OpenSSLRSADNSCryptoKeyEngine::parse(std::map<std::string, std::string>& stormap, const std::string& key)
+{
+  const std::string& v = stormap[key];
+  auto n = std::unique_ptr<BIGNUM, void(*)(BIGNUM*)>(BN_bin2bn((const unsigned char*)v.c_str(), v.length(), nullptr), BN_clear_free);
```
rgacogne, line 436:
```suggestion
  auto n = std::unique_ptr<BIGNUM, void(*)(BIGNUM*)>(BN_bin2bn(reinterpret_cast<const unsigned char*>(v.data()), v.length(), nullptr), BN_clear_free);
```

rgacogne (approved):

pdns/opensslsigners.cc
```diff
@@ -429,88 +430,44 @@ std::string OpenSSLRSADNSCryptoKeyEngine::getPublicKeyString() const
 }
 
 
+std::unique_ptr<BIGNUM, void(*)(BIGNUM*)>OpenSSLRSADNSCryptoKeyEngine::parse(std::map<std::string, std::string>& stormap, const std::string& key)
+{
+  const std::string& v = stormap[key];
```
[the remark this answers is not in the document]
omoerbeek: If absent, the choice is exception on getting the value or on parsing an empty string

pdns/opensslsigners.cc
omoerbeek: Any particular reason on `data()` vs `c_str()`? In C++11, they are equivalent, AFAIK.

rgacogne: In my mind `c_str()` implies that we need a null-terminated string, which is not true here. But I don't really care either way :)

PR #9591, opened 2020-10-06
Long overdue table driven option setting for libfstrm

omoerbeek:
### Short description

It's long overdue since I was wanting to do it for more than a year. Nothing fancy,  but I was triggered by CodeQL saying: this function is too large and has too little comments.
### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/fstrm_logger.cc
+++ b/pdns/fstrm_logger.cc
@@ -84,43 +84,26 @@ FrameStreamLogger::FrameStreamLogger(const int family, const std::string& addres
       throw std::runtime_error("FrameStreamLogger: fstrm_iothr_options_set_queue_model failed: " + std::to_string(res));
     }
 
-    if (options.find("bufferHint") != options.end() && options.at("bufferHint")) {
-      res = fstrm_iothr_options_set_buffer_hint(d_iothropt, options.at("bufferHint"));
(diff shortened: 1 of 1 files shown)


pdns/fstrm_logger.cc
```diff
@@ -84,43 +84,26 @@ FrameStreamLogger::FrameStreamLogger(const int family, const std::string& addres
       throw std::runtime_error("FrameStreamLogger: fstrm_iothr_options_set_queue_model failed: " + std::to_string(res));
     }
 
-    if (options.find("bufferHint") != options.end() && options.at("bufferHint")) {
-      res = fstrm_iothr_options_set_buffer_hint(d_iothropt, options.at("bufferHint"));
-      if (res != fstrm_res_success) {
-        throw std::runtime_error("FrameStreamLogger: fstrm_iothr_options_set_buffer_hint failed: " + std::to_string(res));
+    const struct {
+      const char *name;
```
rgacogne, line 88: How about storing the names as strings? It does not really matter since this is only called at configuration time but it looks like that would save a lot of temporary string construction below.
```suggestion
      const std::string name;
```

rgacogne (approved):

Issue #9091, opened 2020-05-06
Serial 0 in SOA record


sadequenoor:
I have installed Authoritative Server 4.3.0-1pdns.e17 on 3 identical Centos 7 machines. 
I have 1 master and 2 slaves.

When I add a SOA record with serial 0 into the records table, the notified_serial in the domains table is changed to 0 on the master. On the slaves the notified_serial is set to NULL in the domains table. Everything else looks like it is the same, the entries all replicate with the exception of this one cell value.

Habbie:
If I understood correctly (from IRC discussion), the problem is that the 0 serial correctly replicates to the slave via AXFR, but the slave then keeps saying it does not have the zone, because we treat the zero serial wrong.

zeha:
Didn't @mind04 fix this recently?

mind04:
Yes, fixed in #9649

zeha:
> Yes, fixed in #9649

I was thinking about another PR, but can't remember which one :(

Habbie:
As far as I know, this should work in 4.4. Closing this. Please report back if you find otherwise!

PR #9593, opened 2020-10-06
Auth: remove deprecated features/settings

### LINKED ISSUE #7919: auth: remove SOA autofilling
- Program: Authoritative
 - Issue type: Feature removal

### Short description
Proposal: in 4.3.0, remove these settings:
```
# default-soa-mail	mail address to insert in the SOA record if none set in the backend
# default-soa-mail=
# default-soa-name	name to insert in the SOA record if none set in the backend
# default-soa-name=a.misconfigured.powerdns.server
# soa-expire-default	Default SOA expire
# soa-expire-default=604800
# soa-minimum-ttl	Default SOA minimum ttl
# soa-minimum-ttl=3600
# soa-refresh-default	Default SOA refresh
# soa-refresh-default=10800
# soa-retry-default	Default SOA retry
# soa-retry-default=3600
```

This would close #6248, #5382, #4321, #3518.

It is my suspicion that not many users rely on the feature, or at least, that most of those that do, could also pretty easily live without, now that we've been pushing the API for a while.

@mind04 mentions that PowerAdmin users might be relying on this feature; we should check that and maybe help them change that.

mind04:
### Short description
Closes #7919

### Checklist
I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


Habbie (approved):
This looks good to me. Needs some upgrade notes.

I wouldn't mind a second reviewer.

docs/settings.rst
```diff
@@ -434,7 +446,7 @@ Overrides :ref:`setting-default-soa-edit`
 -  String
 
 .. deprecated:: 4.2.0
-  This setting has been deprecated and will be removed in 4.4.0
+  This setting has been removed in 4.4.0
```
zeha, line 449: i was thinking it might be nicer if a new block would be added:
```
.. versionchanged:: 4.4.0
  This setting has been removed
```

so to keep the 4.2/4.3 docs "correct".

but maybe its clear enough.

pdns/pdnsutil.cc
```diff
@@ -1236,12 +1231,10 @@ static int createZone(const DNSName &zone, const DNSName& nsname) {
...
 
-  string soa = (boost::format("%s %s 1")
-                % (nsname.empty() ? ::arg()["default-soa-name"] : nsname.toString())
-                % (::arg().isEmpty("default-soa-mail") ? (DNSName("hostmaster.") + zone).toString() : ::arg()["default-soa-mail"])
-  ).str();
+  string soa = ::arg()["default-soa-content"];
+  boost::replace_all(soa, "@", zone.toStringNoDot());
   SOAData sd;
-  fillSOAData(soa, sd);  // fills out default values for us
+  fillSOAData(soa, sd);
```
zeha, line 1237: do we gain anything from keeping the fillSOAData / makeSOAContent roundtrip?

zeha:
If the default serial is 0, it should be ensured that SOA-EDIT is applied in all cases; otherwise some coe parts may treat the zone as being absent / not yet filled out (like a secondary zone that has not been AXFRed yet).

klaus3000:
0 is a valid serial. Hence, 0 should not be used as "no serial info available". For this, -1 may be more appropriate or use a dedicated flag.

pdns/pdnsutil.cc
mind04: Validation of the value in "default-soa-content"

docs/settings.rst
mind04: Both versions exist in the current documentation so I guess it is a matter of taste. I do not see how this this will make the 4.2/4.3 docs less "correct"

pdns/pdnsutil.cc
zeha: Right. Might be a bit confusing at this point, but very well.

zeha (approved):

mind04:
@Habbie done. Ci failure seems unrelated to this pull.

Habbie:
> @Habbie done. Ci failure seems unrelated to this pull.

Yes, fixed in #9622

Habbie:
> 0 is a valid serial. Hence, 0 should not be used as "no serial info available". For this, -1 may be more appropriate or use a dedicated flag.

Yes, we have this problem in a broader sense, like in #9091 :(

PR #9596, opened 2020-10-09
List products per distro release

aerique:
This is an attempt to list the required files to be written differently but after writing it, I'm not sure it is much of an improvement, still I'm curious about your opinions.

Pros:
- clear if-blocks per distribution release
- per if-block the supported products are listed

Con:
- the code blocks themselves are not simpler due to the need of `$PRODUCT`

It has been tested for each script argument using `sha256sum` that the exact same file are produced.

This also remove the `auth-40` and `rec-40` arguments since they're EOL and do not work anymore.

### DIFF
--- a/build-scripts/docker/generate-repo-files.sh
+++ b/build-scripts/docker/generate-repo-files.sh
@@ -24,8 +24,8 @@
 if [ "$1" = "" -o "$1" = "-?" -o "$1" = "-h" -o "$1" = "--help" ]; then
     echo "Usage: generate-repo-files.sh RELEASE"
     echo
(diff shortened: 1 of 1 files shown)


Habbie:
> This is an attempt to list the required files to be written differently but after writing it, I'm not sure it is much of an improvement.

This really does not want to be a shell script. It probably wants to be a Python script :-)

aerique:
> This really does not want to be a shell script. It probably wants to be a Python script :-)

Yes :-D

That will be the next step that's on my list and perhaps we should skip this step (but you know how it goes, there's more important things looming on the horizon and that might be another 2 years).

aerique:
Let's kill this PR.  After looking at this again it doesn't add much, if anything.

PR #9597, opened 2020-10-12
rec: Move to several distinct Bogus states, for easier debugging

rgacogne:
### Short description

This is especially useful after the fact, when the Bogus state is retrieved from a cache and we don't have any clue left as to how we ended up with that state.
Documentation is missing, and I wouldn't be surprised to see some regression tests fail.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/validate.cc
+++ b/pdns/validate.cc
@@ -745,9 +745,14 @@ static const vector<DNSName> getZoneCuts(const DNSName& begin, const DNSName& en
   return ret;
 }
 
-bool isRRSIGNotExpired(const time_t now, const shared_ptr<RRSIGRecordContent> sig)
+bool isRRSIGNotExpired(const time_t now, const shared_ptr<RRSIGRecordContent>& sig)
 {
-  return sig->d_siginception - g_signatureInceptionSkew <= now && sig->d_sigexpire >= now;
+  return sig->d_sigexpire >= now;
+}
+
+bool isRRSIGIncepted(const time_t now, const shared_ptr<RRSIGRecordContent>& sig)
+{
+  return sig->d_siginception - g_signatureInceptionSkew <= now;
 }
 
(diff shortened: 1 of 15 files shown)


omoerbeek:
Love this. Puzzled by the sudden Travis failure....

rgacogne:
My money is on a dependency error that has already been fixed for the travis failure:
```
ERROR: Package 'PyHamcrest' requires a different Python: 2.7.14 not in '>=3.5'
```
I restarted it.

omoerbeek (approved):

rgacogne:
I wonder if it would make sense to have a look at the cases covered in rfc8914 before merging this? Perhaps we could add some Bogus states that would make the implementation of rfc8914 easier.

rgacogne:
I added more states to better match the extended error codes defined in rfc8914.

rgacogne:
Otto, are you OK with me merging this?

pdns/validate.cc
```diff
@@ -952,6 +982,64 @@ void validateDNSKeysAgainstDS(time_t now, const DNSName& zone, const dsmap_t& ds
       //        if(validkeys.empty()) cerr<<"did not manage to validate DNSKEY set based on DS-validated KSK, only passing KSK on"<<endl;
     }
   }
+
+  if (validkeys.size() < tkeys.size()) {
+    /* so we failed to validate the whole set, let's try to find out why exactly */
```
pieterlexis, line 987: What if one key is bad but the rest are good? do we still bail with a Bogus?

rgacogne: You mean if we did not find at least one DNSKEY, corresponding to a DS, that validates the signature of the whole DNSKEY RRset? Yes

omoerbeek:
> Otto, are you OK with me merging this?

yes

PR #9599, opened 2020-10-12
fix the initialization order in the class.

ihsinme:
### Short description
```
The uninitialized class member 'd_trc' is used to initialize the 'd_tsigVerifier' member.
```

According to the language standard, class members are initialized in the constructor in the same order as they are declared inside the class.

Your code contains a constructor where the initialization of one class member depends on another.

```cpp
AXFRRetriever::AXFRRetriever(const ComboAddress& remote,
                             const DNSName& domain,
                             const TSIGTriplet& tt, 
                             const ComboAddress* laddr,
                             size_t maxReceivedBytes,
                             uint16_t timeout)
  : d_tsigVerifier(tt, remote, d_trc), d_receivedBytes(0), d_maxReceivedBytes(maxReceivedBytes)
```
At the same time, a variable used for initialization is not yet initialized itself.

To fix the bug, we need to put the declaration of the 'd_trc' class member before the declaration of the 'd_tsigVerifier' class member.
  
### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/axfr-retriever.hh
+++ b/pdns/axfr-retriever.hh
@@ -48,9 +48,9 @@ class AXFRRetriever : public boost::noncopyable
     int d_sock;
     int d_soacount;
     ComboAddress d_remote;
+    TSIGRecordContent d_trc;
     TSIGTCPVerifier d_tsigVerifier;
 
     size_t d_receivedBytes;
     size_t d_maxReceivedBytes;
-    TSIGRecordContent d_trc;
 };


rgacogne (approved):
Thank you for this pull request! At first glance it looks like the `TSIGTCPVerifier` constructor only takes a reference on the `TSIGRecordContent` object so I don't expect any effective bug, but I agree we need to apply this fix.

omoerbeek:
Yes, this is right. Always find this dependency on the order of declaration awkward. But it's the way it is.

omoerbeek (approved):

PR #9607, opened 2020-10-13
improved implementation of the operator=.

ihsinme:
### Short description
In the implementation of operator = of the NSECBitmap class, there is no check for assigning an object to itself.
However, the memory allocation is performed in the function itself.
This is a potentially dangerous place, as memory leaks can occur during use.

You can fix this area in various ways, even using assert.
But considering how a similar check was applied in another class.

```cpp
  DNSName& operator=(const DNSName& rhs)
  {
    if (this != &rhs) {
      d_storage = rhs.d_storage;
    }
    return *this;
  }
```
I propose this fix.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsrecords.hh
+++ b/pdns/dnsrecords.hh
@@ -583,6 +583,10 @@ public:
   }
   NSECBitmap& operator=(const NSECBitmap& rhs)
   {
+    if (this == &rhs) {
+      return *this;
+    }
+
     d_set = rhs.d_set;
 
     if (rhs.d_bitset) {


rgacogne (commented):
I think the situation is a bit different here than it was for the `DNSName`, where some versions of `boost::container::string` assert on self-assignment. In the `NSECBitmap` case I don't see any harm from a self-assignment, except of course the wasting some CPU cycles. What makes you think there is a possible leak? Have you seen actual cases where we do self-assign a `NSECBitmap`?

ihsinme:
from my point of view, the leakage problem can arise in the situation of assigning an object to itself, since at this moment the memory in the heap will be allocated for d_bitset.
And the previously allocated memory for d_bitset will be lost and can no longer be freed.

rgacogne:
> And the previously allocated memory for d_bitset will be lost and can no longer be freed.

Isn't `d_bitset` a unique pointer that will be automatically released?

rgacogne:
`d_bitset` is a unique pointer, which is a kind of smart pointer (see [1]). The goal of this kind of pointer is to prevent memory leaks by making sure that the memory is released if the unique pointer goes out of scope (using C++'s RAII, see [2]), and that the previously allocated memory is released if a new value is set. 
In that case I believe the memory allocated by new is immediately stored in a unique pointer, making sure it will be released if either that pointer goes out of scope before it can be stored in `d_bitset` (that seems unlikely here but can sometimes happen when an exception is raised), if the object is destroyed or later if a new value is stored into `d_bitset`.

[1]: https://en.wikipedia.org/wiki/Smart_pointer
[2]: https://en.wikipedia.org/wiki/Resource_acquisition_is_initialization

ihsinme:
Thanks a lot for the clarification.
it will take me a while to analyze this on a practical mock-up.

ihsinme:
i checked on mock.
you were absolutely right.
the heap is automatically cleaned up.

sorry to distract and waste your time.

PR #9616, opened 2020-10-15
dnsdist: Add support for incoming Proxy Protocol

rgacogne:
### Short description

This PR:
- unifies the way UDP, TCP and DoH deal with incoming data by using a `std::vector<uint8_t>` everywhere. This makes it possible to pass the size around without any risk of mismatch, allow expanding the buffer at any point, and prevents signedness issues that could arise with signed chars. It uses a new type of vector that does not initialize the memory on a resize, because calling zero-ing a bunch of bytes is expensive when you do it 100k times per second ;
- implements incoming proxy protocol, making it possible to chain several dnsdist, or simply to pass data to dnsdist from a "smart" client. Incoming data will be forwarded to a downstream server if it supports the proxy protocol as well, although TLV data can of course be inspected and edited from dnsdist.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


rgacogne:
This PR also gets rid of the "keep trailing data" feature, since it does not really make sense to maintain it once we have a way to pass data from a client to the backend.

rgacogne:
Rebased on the squashed version of https://github.com/PowerDNS/pdns/pull/9582, and squashed as well.

rgacogne:
Rebased to fix a conflict.

rgacogne:
Rebased to fix a conflict.

rgacogne:
Rebased to fix a conflict with the protozero PR.

rgacogne:
Rebased to fix a conflict., plus fixed a few nits reported by cppcheck (`try` blocks being function's bodies, an uninitialized `dnsheader` variable in one unit test).

rgacogne:
@omoerbeek @Habbie I'm quite happy with the dnsdist's parts of this PR, and there are very few changes to code used by the Authoritative server or the Recursor, mostly the templating of `DNSPacketWriter` to be able to use a vector with a specific allocator (`NoInitVector` in that code) that does not set the memory to zero when a vector is made bigger via `std::vector::resize()`. That sounds like a small optimization but does matter in dnsdist because for every `UDP` query we do resize the vector to the actual number of bytes we received, before resizing it again to a larger capacity for the next query. It takes real CPU time to do that hundred of thousands of times per second and we don't care about that memory being set to zero.
So I'll likely merge this PR once our CI turns green, but please let me know if you would like to skim over the changes first!

Habbie:
Skimmed, please go ahead!

omoerbeek:
> Skimmed, please go ahead!

Yes. I remember skimming over it earlier (without doing an actual review).

Issue #804, opened 2013-05-13
generic syntax fails for known types


Habbie:
bindbackend zonefile content:

```
weirdtxt.example.net.    3600 IN TYPE16 \# 4 03 78 0e 78 
```

dig output:

```
weirdtxt.example.net.   3600    IN  TXT "# 4 03 78 0e 78"
```

(TYPE16 is TXT)

silbe:
Has there been any progress on fixing that? This kind of incompatibility makes it really hard to use new RR types in production setups. One has to start using the generic syntax because the version of PowerDNS shipped by the distro will usually be too "old" to support the new RR type. But on the next distro update of one of the servers (synchronised via something other than AXFR) the whole setup will break down as the different versions need different content.

Habbie:
There has been no progress but this is high on my list, indeed other people have also been reporting pain at upgrade time.

klaus3000:
Any progress on this? This hits us quite hard at the moment. Customers begin to use CAA with presigned zones and those "fancy" records in <4.0 are not implemented correctly. Hence, pdns 3.4 does not deliver RRSIGs for CAA (MBOXFW). So we need to update to 4.0 but have no clean upgrade path.

ahupowerdns:
klaus, can you clarify what you would need to upgrade to 4.x?

klaus3000:
Actually I am not sure what would be the best way. At the moment we use pdns 3.4.11. The "control" server receives the zones from customers (AXFR) and stores it into the DB. The DB is replicated to ~30 nodes where pdns "slaves" reads from the DB and serves requests. CAA used to work fine (stored in DB as MBOXFW). Now we have customers with CAA and pre-signed zones. As the MBOXFW is implemented as "fancy type" it seems incomplete and RRSIGs are not served.

So we need a smooth way to upgrade the control server and the slaves, one after the other. The problem is, once we upgrade a slave to 4.0.3 it does not serve CAA anymore (as in the DB is a MBOXFW RR) until we update the control server - and the other way round. Even worse, if not using 4.0.3, but using newest rel/auth-4.0.x PDNS we respond with servfail to all queries (even for A queries) if there is an MBOXFW in DB: Exception building answer packet for example.com/CAA (All data was not consumed) sending out servfail

ahupowerdns:
[...]

PR #9623, opened 2020-10-19
gsql,bind: allow seamless serving of newly-supported TYPExxx records

zeha:
### Short description

When upgrading to a release supporting new record types, all existing database entries using the TYPExxx format stop working. This prevents testing of a "master" release while preserving compatibility with a production database.

Makes TYPExxx work by always sending them through UnknownRecordContent and deserialising that using the specific QType handler.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -1731,6 +1731,25 @@ of reflection attacks. Maximum value is 65535, but values above
   1232 is the largest number of payload bytes that can fit in the smallest IPv6 packet.
   IPv6 has a minimum MTU of 1280 bytes (:rfc:`RFC 8200, section 5 <8200#section-5>`), minus 40 bytes for the IPv6 header, minus 8 bytes for the UDP header gives 1232, the maximum payload size for the DNS response.
 
+.. _setting-upgrade-unknown-types:
+
+``upgrade-unknown-types``
+-------------------------
+
+-  Boolean
+-  Default: no
+
+.. versionadded:: 4.4.0
+
+Transparently upgrade records stored as `TYPE#xxx` and RFC 3597 (hex format)
+contents, if the type is natively supported.
+When this is disabled, records stored in this format cannot be served.
+
+Recommendation: keep disabled for better performance.
+Enable for testing PowerDNS upgrades, without changing stored records.
+
+This option is supported by the bind and Generic SQL backends. 
+
 .. _setting-version-string:
 
 ``version-string``
--- a/pdns/common_startup.cc
+++ b/pdns/common_startup.cc
@@ -238,6 +238,7 @@ void declareArguments()
   ::arg().set("tcp-fast-open", "Enable TCP Fast Open support on the listening sockets, using the supplied numerical value as the queue size")="0";
 
   ::arg().set("max-generate-steps", "Maximum number of $GENERATE steps when loading a zone from a file")="0";
+  ::arg().setSwitch("upgrade-unknown-types","Transparently upgrade known TYPExxx records. Recommended to keep off, except for PowerDNS upgrades until data sources are cleaned up")="no";
 
   ::arg().set("rng", "Specify the random number generator to use. Valid values are auto,sodium,openssl,getrandom,arc4random,urandom.")="auto";
   ::arg().setDefaults();
--- a/pdns/dnsparser.cc
+++ b/pdns/dnsparser.cc
@@ -26,74 +26,56 @@
 
 #include "namespaces.hh"
 
-class UnknownRecordContent : public DNSRecordContent
+UnknownRecordContent::UnknownRecordContent(const string& zone) 
 {
-public:
-  UnknownRecordContent(const DNSRecord& dr, PacketReader& pr) 
-    : d_dr(dr)
-  {
-    pr.copyRecord(d_record, dr.d_clen);
+  // parse the input
+  vector<string> parts;
+  stringtok(parts, zone);
+  // we need exactly 3 parts, except if the length field is set to 0 then we only need 2
+  if (parts.size() != 3 && !(parts.size() == 2 && equals(parts.at(1), "0"))) {
+    throw MOADNSException("Unknown record was stored incorrectly, need 3 fields, got " + std::to_string(parts.size()) + ": " + zone);
   }
 
-  UnknownRecordContent(const string& zone) 
-  {
-    // parse the input
-    vector<string> parts;
-    stringtok(parts, zone);
-    // we need exactly 3 parts, except if the length field is set to 0 then we only need 2
-    if (parts.size() != 3 && !(parts.size() == 2 && equals(parts.at(1), "0"))) {
-      throw MOADNSException("Unknown record was stored incorrectly, need 3 fields, got " + std::to_string(parts.size()) + ": " + zone);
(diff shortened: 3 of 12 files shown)


klaus3000:
This would solve issue #804. (at least for SQL backend?)

Habbie:
That's a neat trick! Please hide this behind an option so users still have an incentive to update contents when upgrading their PowerDNS :)

Longer term (so outside of this PR), perhaps we should have a pdnsutil command to 'upgrade' existing content.

zeha:
Setting name suggestions very welcome!

klaus3000:
> That's a neat trick! Please hide this behind an option so users still have an incentive to update contents when upgrading their PowerDNS :)
> 

IMO this is a bug fix, as currently PowerDNS is not compatible with RRs created by previous PowerDNS versions. A bug fix should not be hidden behind an option, but be enabled always (be liberal what you expect). If you are concerned about old TYPExx RRs, then PowerDNS could log a warning `old style RR TYPExxx detectected for label www.example.com - please use "pdnsutil upgrade example.com" to convert the RR into format FOOBAR`.

In this approach it seems that there is a performance penalty. Maybe the logic can be changed that unkown-RR-parsing happens only when the normal parsing triggerend an exception. If this would be too complex I agree on the config option.

> Longer term (so outside of this PR), perhaps we should have a pdnsutil command to 'upgrade' existing content.

Yes.

Habbie:
> Maybe the logic can be changed that unkown-RR-parsing happens only when the normal parsing triggerend an exception.

I suspect that that exception would trigger always, and thus just cost extra.

klaus3000:
> > Maybe the logic can be changed that unkown-RR-parsing happens only when the normal parsing triggerend an exception.
> 
> I suspect that that exception would trigger always, and thus just cost extra.

Currently, when PowerDNS parses a meanwhile well-known TYPExxx record it throws an exception. So, why not catch this exception and reparse with the generic TYPExxx parser? But I am happy with above solution too :-)

zeha:
> Currently, when PowerDNS parses a meanwhile well-known TYPExxx record it throws an exception. So, why not catch this exception and reparse with the generic TYPExxx parser? But I am happy with above solution too :-)

The exception on this happens way too late for most backends. - bind knows while parsing, but for other backends this happens when building the actual response packet (or AXFR response or whatever).

One could do an internal refactoring and make the backends return wire format, but this is a larger undertaking. Would be a nice thing to do, though.

docs/settings.rst
```diff
@@ -1731,6 +1731,25 @@ of reflection attacks. Maximum value is 65535, but values above
...
+.. versionadded:: 4.4.0
+
+Transparently upgrade records stored as `TYPE#xxx` and RFC 3597 (hex format)
+contents, if the type is natively supported.
+When this is disabled, records stored in this format cannot be served.
+
+Recommendation: keep disabled for better performance.
+Enable for testing PowerDNS upgrades, without changing stored records.
+
+Not all backends can use this option.
```
cmouse, line 1751: `can use` => `support`

zeha:
> That's a neat trick! Please hide this behind an option so users still have an incentive to update contents when upgrading their PowerDNS :)

I've manually tested gpgsql and bindbackend (successfully). Ready for review IMO.

pdns/dnsparser.cc
```diff
@@ -26,74 +26,56 @@
...
+  string out;
+  out.reserve(total + 1);
 
-    for (unsigned int n = 0; n < total; ++n) {
-      int c;
-      if (sscanf(&relevant.at(2*n), "%02x", &c) != 1) {
-        throw MOADNSException("unable to read data at position " + std::to_string(2 * n) + " from unknown record of size " + std::to_string(relevant.size()));
-      }
-      out.append(1, (char)c);
+  for (unsigned int n = 0; n < total; ++n) {
```
cmouse, line 52: you could use boost::algorithm::unhex here..

```.c
std::stringstream ss;
boost::algorithm::unhex(relevant, std::ostream_iterator<char>(ss));
out = ss.str();
```

pdns/dnsparser.cc
```diff
@@ -26,74 +26,56 @@
...
 
-  void toPacket(DNSPacketWriter& pw) override
-  {
-    pw.xfrBlob(string(d_record.begin(),d_record.end()));
+string UnknownRecordContent::getZoneRepresentation(bool noDot) const
+{
+  ostringstream str;
+  str<<"\\# "<<(unsigned int)d_record.size()<<" ";
+  char hex[4];
+  for (size_t n=0; n<d_record.size(); ++n) {
```
cmouse, line 68: you could use `boost::algorithm::hex` here.

cmouse (changes requested):
You should add unit tests

pdns/dnsparser.cc
zeha: none of this is new code, and i'm not going to change it.

pdns/dnsparser.cc
zeha: same as above.

docs/settings.rst
Habbie, line 1751:
```suggestion
Not all backends support this option.
```

Habbie (approved):
Please apply the doc nit, then I'll merge!

docs/settings.rst
mind04: If you do, please add a list of supported backends as well.

Habbie: ah, yes please

pdns/common_startup.cc
```diff
@@ -238,6 +238,7 @@ void declareArguments()
   ::arg().set("tcp-fast-open", "Enable TCP Fast Open support on the listening sockets, using the supplied numerical value as the queue size")="0";
 
   ::arg().set("max-generate-steps", "Maximum number of $GENERATE steps when loading a zone from a file")="0";
+  ::arg().setSwitch("upgrade-unknown-types","Transparently upgrade known TYPExxx records. Recommended to keep off, except for PowerDNS upgrades until data sources are cleaned up")="no";
```
mind04, line 241: Are we sure about the "upgrade-unknown-types" because the comment says "upgrade known TYPExxx records" and this pull is upgrading known types.

Maybe "upgrade-known-typex-records" is a better name or "serve-known-typex-records". But my "naming things" skills are terrible, so feel free to pick something else.

mind04 (approved):

pdns/common_startup.cc
zeha: I had pretty much the same question myself. But I too suck at naming things.

IDK if "typex-records" is something understandable. Maybe "serve-hex-encoded-records"?

Habbie: > Are we sure about the "upgrade-unknown-types" because the comment says "upgrade known TYPExxx records" and this pull is upgrading known types.

It's upgrading the known unknown types. I think it might be fine just like this :)

zeha: Okay, lets ship with this name for now.

docs/settings.rst
zeha: added!

PR #9630, opened 2020-10-21
Rec: start using protozero to create protobuf messages

omoerbeek:
### Short description

Google's protbuf library has many features, but allocates quite a lot of memory. Using momery takes time...

Start using protozero for query and response logging. Performance impact of logging is now way less. In a particular test using dnsperf with 100 names no protobuf logging would hit 28kqps. Old code (using `logQueries` only) made it drop to 21kqps. New code hovers comes in around 26kqps.

Previously the packet cache would hold a protobuf object, now it stores serialized data in two chunks.
These strings are used to construct a new protobuf message by appending the fields that differ per query.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [X] documented the code

### DIFF
--- a/pdns/pdns_recursor.cc
+++ b/pdns/pdns_recursor.cc
@@ -2706,9 +2706,9 @@ static string* doProcessUDPQuestion(const std::string& question, const ComboAddr
 
 #ifdef HAVE_PROTOBUF
       if(t_protobufServers && logResponse && !(luaconfsLocal->protobufExportConfig.taggedOnly && pbData && !pbData->d_tagged)) { // XXX
-        pdns::ProtoZero::Message pbMessage(pbData->d_message, pbData->d_response, 64, 10); // The extra bytes we are going to add
+        pdns::ProtoZero::Message pbMessage(pbData ? pbData->d_message : "", pbData ? pbData->d_response : "", 64, 10); // The extra bytes we are going to add
         if (pbData) {
-          // We take the unmutable string from the cache an are appending a few values
+          // We take the inmutable string from the cache and are appending a few values
         } else {
           pbMessage.setType(2); // Response
           pbMessage.setServerIdentity(SyncRes::s_serverID);
--- a/pdns/pdns_recursor.cc
+++ b/pdns/pdns_recursor.cc
@@ -885,8 +885,8 @@ static void protobufLogQuery(uint8_t maskV4, uint8_t maskV6, const boost::uuids:
   Netmask requestorNM(remote, remote.sin4.sin_family == AF_INET ? maskV4 : maskV6);
   ComboAddress requestor = requestorNM.getMaskedNetwork();
   requestor.setPort(remote.getPort());
-  
-  pdns::ProtoZero::Message m{128}; // guess at size
+
+  pdns::ProtoZero::Message m{128, std::string::size_type(policyTags.empty() ? 0 : 64)}; // XXX guesses at sizes
   m.request(uniqueId, requestor, local, qname, qtype, qclass, id, tcp, len);
   m.setServerIdentity(SyncRes::s_serverID);
   m.setEDNSSubnet(ednssubnet, ednssubnet.isIPv4() ? maskV4 : maskV6);
@@ -895,25 +895,24 @@ static void protobufLogQuery(uint8_t maskV4, uint8_t maskV6, const boost::uuids:
   m.setDeviceName(deviceName);
 
   if (!policyTags.empty()) {
-    m.startResponse();
     m.setPolicyTags(policyTags);
-    m.finishResponse();
   }
 
+  std::string msg(m.movebuf());
   for (auto& server : *t_protobufServers) {
-    server->queueData(m.getbuf());
+    server->queueData(msg);
   }
 }
 
-static void protobufLogResponse(const pdns::ProtoZero::Message& message)
+static void protobufLogResponse(pdns::ProtoZero::Message& message)
 {
   if (!t_protobufServers) {
     return;
   }
 
-  string data = message.getbuf();
+  std::string msg(message.movebuf());
   for (auto& server : *t_protobufServers) {
-    server->queueData(data);
+    server->queueData(msg);
   }
 }
 #endif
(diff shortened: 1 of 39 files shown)


pdns/pdns_recursor.cc
```diff
@@ -1320,15 +1317,15 @@ static void startDoResolve(void *p)
     auto luaconfsLocal = g_luaconfs.getLocal();
     // Used to tell syncres later on if we should apply NSDNAME and NSIP RPZ triggers for this query
     bool wantsRPZ(true);
-    boost::optional<RecProtoBufMessage> pbMessage(boost::none);
+    std::unique_ptr<pdns::ProtoZero::Message> pbMessage;
```
rgacogne, line 1320: Perhaps we could make that a `boost::optional<pdns::ProtoZero::Message>` instead, to prevent a dynamic memory allocation?

pdns/pdns_recursor.cc
```diff
@@ -797,38 +798,34 @@ static void protobufLogQuery(uint8_t maskV4, uint8_t maskV6, const boost::uuids:
...
   ComboAddress requestor = requestorNM.getMaskedNetwork();
   requestor.setPort(remote.getPort());
-  RecProtoBufMessage message(DNSProtoBufMessage::Query, uniqueId, &requestor, &local, qname, qtype, qclass, id, tcp, len);
-  message.setServerIdentity(SyncRes::s_serverID);
-  message.setEDNSSubnet(ednssubnet, ednssubnet.isIPv4() ? maskV4 : maskV6);
-  message.setRequestorId(requestorId);
-  message.setDeviceId(deviceId);
-  message.setDeviceName(deviceName);
+
+  pdns::ProtoZero::Message m{128, std::string::size_type(policyTags.empty() ? 0 : 64)}; // XXX guesses at sizes
```
rgacogne, line 802: Did you consider a `pdns::ProtoZero::Message` constructor taking a reference to existing string objects instead? That way I think we could pass already allocated strings, like `thread_local` ones, which would allow us to reuse the same allocations over and over.
That won't work in `startDoResolve()`, of course, because we can switch to a different `mthread` from that function.

pdns/protozero.hh
```diff
@@ -0,0 +1,254 @@
...
+      // Construct a Message with (partially) constructed content
+      Message(const std::string& buf1, const std::string& buf2, std::string::size_type sz1, std::string::size_type sz2) :
+        d_msgbuf{buf1}, d_rspbuf{buf2}, d_message{d_msgbuf}, d_response{d_rspbuf}
+      {
+        // We expect to grow the buffers, in the end the d_message will contains the (grown) d_response
+        // This is extra space in addition to what's already there
+        // Different from what string.reserve() does
+        d_message.reserve(sz1 + d_rspbuf.length() + sz2);
+        d_response.reserve(sz2);
+      }
```
rgacogne, line 55: Perhaps we could mark the copy constructor, copy assignment operator, move constructor and move assignment operator as `delete`?

rgacogne: That might not play too well with the `boost::optional` suggestion, though, we would need to check.

pdns/protozero.hh
```diff
@@ -0,0 +1,254 @@
...
+namespace pdns {
+  namespace ProtoZero {
+    class Message {
+    public:
+      // Start a new messagebug, constaining separate data for the response part
+      Message(std::string::size_type sz1, std::string::size_type sz2) : d_message{d_msgbuf}, d_response{d_rspbuf}
+      {
+        // This is extra space in addition to what's already there
+        // Different from what string.reserve() does
+        d_message.reserve(sz1);
```
rgacogne, line 42: Be careful when reserving a `std::string`, it can actually trigger a reallocation of a smaller buffer (before C++20). I would check the `capacity()` before calling `reserve()`.

pdns/protozero.hh
```diff
@@ -0,0 +1,254 @@
...
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+#pragma once
+
+#include <protozero/pbf_writer.hpp>
+#include <string>
+
+#include "config.h"
+#include "iputils.hh"
+#include "filterpo.hh"
```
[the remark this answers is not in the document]
omoerbeek: Yes, we can split  the general part and the recursor specific part.

pdns/protozero.hh
```diff
@@ -0,0 +1,254 @@
...
+      {
+        encodeComboAddress(6, ca);
+      }
+      void setTo(const ComboAddress& ca)
+      {
+        encodeComboAddress(7, ca);
+      }
+      void setInBytes(uint64_t len)
+      {
+        if (len) {
```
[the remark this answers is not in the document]
omoerbeek: I was thinking: len = zero is implied by absence...

pdns/pdns_recursor.cc
omoerbeek: My original code had that but it caused bugs. Will investigate a bit further now general part is working.

pdns/pdns_recursor.cc
omoerbeek: I considered it, but wanted to avoid the extra complexity at the moment. We can always optimize this later.

pdns/protozero.hh
omoerbeek: This is calling reserve on `protozero::pbf_writer` which has different semantics compared to `std::string`.

pdns/pdns_recursor.cc
omoerbeek: I think it had to do with the default constructor of `protozero::pbf_writer` being used, which creates a non-working writer (as documented). So I'd have to see if that can be avoided somehow even if  the object is embedded in a `boost::optional`

pdns/protozero.hh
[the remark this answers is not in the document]
rgacogne: I don't see how we could have a DNS message smaller than `sizeof(dnsheader)` anyway, so that should be fine.

pdns/protozero.hh
rgacogne: Right, but the issue is the same because that does call `std::string::reserve()` in the end, adding the new size to the existing size, not to the existing capacity. That doesn't matter now, but if we end up reusing an existing string to avoid allocations we sometimes will cause a reallocation by calling reserve() with a smaller size than the current capacity of the string.

pdns/protozero.hh
omoerbeek: I tried to use `boost::optional` again, but failed to do so. My guess is that because `pbf_writer` contains a pointer to the associated string, something goes wrong somewhere, even though the `pbf_writer` object itself *is* movable... But I did not figure out yet how to fix that.

rgacogne (approved):
A couple nits but that PR looks very good!

rgacogne:
Any plans for implementing outgoing queries / incoming responses using protozero as well? I would feel quite good about getting rid of Google's Protocol Buffer library :-)

omoerbeek:
> Any plans for implementing outgoing queries / incoming responses using protozero as well? I would feel quite good about getting rid of Google's Protocol Buffer library :-)

Certainly, even if the performance impact would be way smaller. But I have some other tasks to do first...

rgacogne (approved):

PR #9633, opened 2020-10-21
rec: two OpenBSD improvemenst wrt UDP sockets: port randomization and EGAIN errors

omoerbeek:
Use OpenBSD's reliable local UDP port randomization instead of a home-grown mechanism and retry on sendmsg EAGAIN. The latter can and does happen on non-blocking UDP sockets on OpenBSD.

This makes running multiple recursor threads on OpenBSD much more reliable on heavy load. Previously
you would see quite some EAGAINs (leading to reply loss) on e.g. a dnsperf load test with threads > 1.
The arbitrary max loop count is a bit unsatisfactory....

There might be more systems that return EAGAIN once in a while on sendmsg(2) on a non-blocking UDP socket.

Apology for the white-space changes.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [X] documented the code

### DIFF
--- a/pdns/iputils.cc
+++ b/pdns/iputils.cc
@@ -242,7 +242,8 @@ int sendOnNBSocket(int fd, const struct msghdr *msgh)
 {
   int sendErr = 0;
 #ifdef __OpenBSD__
-  for (int i = 0; i < 10; i++) {
+  // OpenBSD can and does return EAGIN on non-blocking datagram sockets
+  for (int i = 0; i < 10; i++) { // Arbitrary upper bound
     if (sendmsg(fd, msgh, 0) != -1) {
       sendErr = 0;
       break;
--- a/pdns/iputils.cc
+++ b/pdns/iputils.cc
@@ -242,7 +242,7 @@ int sendOnNBSocket(int fd, const struct msghdr *msgh)
 {
   int sendErr = 0;
 #ifdef __OpenBSD__
-  // OpenBSD can and does return EAGIN on non-blocking datagram sockets
+  // OpenBSD can and does return EAGAIN on non-blocking datagram sockets
   for (int i = 0; i < 10; i++) { // Arbitrary upper bound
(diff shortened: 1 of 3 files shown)


Habbie:
>  The arbitrary max loop count is a bit unsatisfactory....

I had to read the tries=2, `--tries`, `tries==1` trick a few too many times, so perhaps that needs more comments, or a different approach.

Habbie:
Otherwise, looks good!

pdns/iputils.cc
```diff
@@ -238,6 +238,28 @@ bool IsAnyAddress(const ComboAddress& addr)
   
   return false;
 }
+int sendOnNBSocket(int fd, const struct msghdr *msgh)
+{
+  int sendErr = 0;
+#ifdef __OpenBSD__
+  // OpenBSD can and does return EAGIN on non-blocking datagram sockets
```
Habbie, line 245:
```suggestion
  // OpenBSD can and does return EAGAIN on non-blocking datagram sockets
```

pdns/pdns_recursor.cc
```diff
@@ -578,50 +578,52 @@ class UDPClientSocks
...
-
-    //    setCloseOnExec(ret); // we're not going to exec
-
-    int tries=10;
+    // The loop below runs the body with [tries-1 tries-2 ... 1]. Last iteration with tries == 1 is special: it uses a kernel
+    // allocated UDP port.
+#if !defined( __OpenBSD__)
+    int tries = 10;
+#else
+    int tries = 2; // hit the reliable kernel random case for OpenBSD immediately, using sysctl net.inet.udp.baddynamic to exclude ports
```
Habbie, line 595:
```suggestion
    int tries = 2; // hit the reliable kernel random case for OpenBSD immediately (because it will match tries==1 below), using sysctl net.inet.udp.baddynamic to exclude ports
```

Habbie (approved):
approved, with two more comment improvement suggestions

PR #9639, opened 2020-10-23
rec: the clock can tick while the negcache test is running.

omoerbeek:
I'm assuming the test always finishes within 1 second...

Fix tested by waiting in a loop until the second is *almost* over and then run the test.

While there zap a redundant test: it's legal to call `free(nullptr)`.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [X] documented the code


rgacogne (commented):
Perhaps we could pass a `const struct timeval&` to `NegCache::dumpToFile` instead?

rgacogne (approved):

PR #9648, opened 2020-10-27
Auth: Add zoneContentAvailable to DomainInfo

zeha:
### Short description

Some parts of the code need to know if a zone exists, but there are no records yet. Previously this was more or less provided by `getSOA`, but #9470 removed that.

Add a new `zoneContentAvailable` bool on DomainInfo, explicitly indicating what `getSOA` did earlier.

For gsql, while there is now a JOIN in getDomainInfo and getUnfreshDomains, they no longer have to call getSOA internally.

TODO:
- [ ] understand the weird stuff lmdb does and set zoneContentAvailable properly
- [ ] more testing

### Checklist

I have:
- [x] compiled this code
- [ ] tested this code
- [ ] documented the code


Habbie (commented):
This makes sense to me.

Habbie:
Some tests appear to have failed.

modules/lmdbbackend/lmdbbackend.cc
```diff
@@ -1005,6 +1006,7 @@ void LMDBBackend::getUnfreshSlaveInfos(vector<DomainInfo>* domains)
     compoundOrdername co;
     MDBOutVal val;
     uint32_t serial = 0;
+    bool zoneContentAvailable = false;
     if (!txn2->txn->get(txn2->db->dbi, co(iter.getID(), g_rootdnsname, QType::SOA), val)) {
       LMDBResourceRecord lrr;
       serFromString(val.get<string_view>(), lrr);
```
zeha, line 1012: I think lmdbbackend neglects to check lrr.disabled here!?

PR #9651, opened 2020-10-28
rec: Allow to specify a name in getMetric() that is used for Prometheus export only.

omoerbeek:
This can be used to specify names that are structured using Prometheus
conventions. If no name Prometheus name is given, do a more thorough
conversion to a name Prometheus likes by replacing any non-alnum
char by an underscore.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/ws-recursor.cc
+++ b/pdns/ws-recursor.cc
@@ -50,8 +50,12 @@ using json11::Json;
 
 void productServerStatisticsFetch(map<string,string>& out)
 {
-  map<string,string> stats = getAllStatsMap(StatComponent::API);
-  out.swap(stats);
+  auto stats = getAllStatsMap(StatComponent::API);
+  map<string,string> ret;
+  for (const auto& entry: stats) {
+    ret.insert(make_pair(entry.first, entry.second.d_value));
+  }
+  out.swap(ret);
 }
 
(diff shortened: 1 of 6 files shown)


pdns/ws-recursor.cc
```diff
@@ -437,13 +437,14 @@ static void prometheusMetrics(HttpRequest *req, HttpResponse *resp) {
...
+    // Argument controls blacklisting of any stats. So
+    // stats-api-blacklist will be used to block returned stats.
+    // Second arg tells to use the prometheus names.
+    varmap_t varmap = getAllStatsMap(StatComponent::API, true);
     for (const auto &tup :  varmap) {
         std::string metricName = tup.first;
-
-        // Prometheus suggest using '_' instead of '-'
-        std::string prometheusMetricName = "pdns_recursor_" + boost::replace_all_copy(metricName, "-", "_");
+        std::string prometheusMetricName = "pdns_recursor_" + metricName;
```
wojas, line 447: If the Prometheus was explicitly set, it must not be prefixed with `pdns_recursor_`.

omoerbeek: Right. I'll move the prepending of the prefix to the proper place.

wojas:
Can this be backported to the 4.4.x?

rgacogne (approved):

PR #9655, opened 2020-10-29
auth: cleanup warnings in ueberbackend, dbdnsseckeeper

zeha:
### Short description
Should clean up:
```
>>>     CID 1435947:  Uninitialized members  (UNINIT_CTOR)
>>>     Non-static class member "d_qtype" is not initialized in this constructor nor in any functions that it calls.
>>>     CID 1030015:  Uninitialized members  (UNINIT_CTOR)
>>>     Non-static class member "zoneId" is not initialized in this constructor nor in any functions that it calls.
```

and:

```
dbdnsseckeeper.cc:621:13: warning: moving a temporary object prevents copy elision [-Wpessimizing-move]
rrsig = std::move(getRR<RRSIGRecordContent>(dzr.dr));
```

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dbdnsseckeeper.cc
+++ b/pdns/dbdnsseckeeper.cc
@@ -618,7 +618,7 @@ void DNSSECKeeper::getPreRRSIGs(UeberBackend& db, vector<DNSZoneRecord>& rrs, ui
 
   db.lookup(QType(QType::RRSIG), !rr.wildcardname.empty() ? rr.wildcardname : rr.dr.d_name, rr.domain_id);
   while(db.get(dzr)) {
-    rrsig = std::move(getRR<RRSIGRecordContent>(dzr.dr));
+    rrsig = getRR<RRSIGRecordContent>(dzr.dr);
     if(rrsig->d_type == rr.dr.d_type) {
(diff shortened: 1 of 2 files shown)


mind04:
Moving d_type to handle is not obvious to me. Handle is somewhat special and passed around to various backends. Since the backends do not use the lookup qtype, this is in my opinion is not the place to store it.
I prefer to not move d_qtype and just initialize the variables.

zeha:
To me, `d_handle` looked like the internal state for `get()`.

What's confusing me a bit more, is why we have `d_handle` **and** `d_question`.

Maybe we should get rid of `d_handle` instead. 🤷🏻

Habbie:
I'd also prefer the simpler version where we initialise the variables that were uninitialised. If the backend does not need something, it does not need to be moved into d_handle - things are confusing enough as they are :)

zeha:
ok, will change in a moment

zeha:
Done!

mind04 (approved):

Habbie (approved):

PR #9659, opened 2020-10-29
Ignore unknown settings

zeha:
### Short description

Add a new `ignore-unknown-settings` setting, similar to ssh's `IgnoreUnknown`. Anything listed in there does not abort config parsing, even if PowerDNS does not know the setting otherwise.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


zeha:
Rebased.

zeha:
Rebased.

pdns/arguments.cc
```diff
@@ -163,40 +197,28 @@ string ArgvMap::configstring(bool running, bool full)
   else
     help="# Autogenerated configuration file template\n\n";
 
+  // Effects parsing, should come first.
```
Habbie, line 200:
```suggestion
  // Affects parsing, should come first.
```

Habbie (approved):
One comment typo, otherwise awesome! Will need review from Otto too. Also, you have one conflict.

pdns/reczones.cc
```diff
@@ -362,17 +362,16 @@ std::shared_ptr<SyncRes::domainmap_t> parseAuthAndForwards()
```
omoerbeek, line 362: These changes do not seem related, better leave them out.

pdns/arguments.cc
```diff
@@ -365,23 +387,39 @@ void ArgvMap::parseOne(const string &arg, const string &parseOnly, bool lax)
...
+        d_params[var] = val;
         d_cleared.insert(var);
       }
     }
-    else if(!lax)
-      throw ArgException("Trying to set unknown parameter '"+var+"'");
+    else
+    {
+      // unknown setting encountered. see if its on the ignore list before throwing.
+      vector<string> parts;
```
omoerbeek, line 408: Why not a set? saves you a loop below

docs/settings.rst
```diff
@@ -678,6 +678,19 @@ or slave settings.
 
 Run within a guardian process. See :ref:`running-guardian`.
 
+.. _setting-ignore-unknown-settings:
```
omoerbeek, line 681: .. versionadded:: xyz missing

pdns/recursordist/docs/settings.rst
```diff
@@ -720,6 +720,19 @@ If set, EDNS options in incoming queries are extracted and passed to the :func:`
 
 If set, the root-hints are read from this file. If unset, default root hints are used.
 
+.. _setting-ignore-unknown-settings:
+
+``ignore-unknown-settings``
+---------------------------
+
```
omoerbeek, line 727: .. versionadded:: xyz missing

pdns/arguments.cc
zeha: set does not actually work with stringtok:

```
./misc.hh:102:17: error: no member named 'push_back' in 'std::__1::set<std::__1::basic_string<char>,
      std::__1::less<std::__1::basic_string<char> >, std::__1::allocator<std::__1::basic_string<char> > >'
      container.push_back (in.substr(i));
      ~~~~~~~~~ ^
arguments.cc:409:7: note: in instantiation of function template specialization 'stringtok<std::__1::set<std::__1::basic_string<char>,
      std::__1::less<std::__1::basic_string<char> >, std::__1::allocator<std::__1::basic_string<char> > > >' requested here
      stringtok(parts, d_params["ignore-unknown-settings"], " ,\t\n\r");
      ^
```

omoerbeek: Bah, I don't see an immediate solution to this, `vector` and `set` are a bit too different...
You still might want to use `std::find` for the searching though

docs/settings.rst
zeha: added

pdns/arguments.cc
```diff
@@ -365,23 +387,39 @@ void ArgvMap::parseOne(const string &arg, const string &parseOnly, bool lax)
...
+      parts.clear();
```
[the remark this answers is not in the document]
zeha: removed, thanks

pdns/arguments.cc
zeha: I've added a stringtok template variant for set.

pdns/reczones.cc
zeha: gone!

pdns/arguments.cc
omoerbeek: Since `vector` actually both has `push_back` and `insert`, will the first template be selected in that case?

Habbie: I don't think the `insert` variant would compile if it picked `vector`?

zeha: `vector`s `insert` shouldn't compile, and IIRC non-compiling templates are not picked if there are other variants?

I haven't seen any trait types that would be useful here.

omoerbeek: I was reading the template chapters of the C++ reference, but I couldn't find a clear description of the case at hand between all the thousands of words. The build does not seem to agree now though, but I did not look in detail.

zeha: Yeah, for some reason now `ld` also hates it locally. Which is weird, because that should never be a symbol problem!?
I'll remove it.

PR #9662, opened 2020-10-30
Coverity fixes

cmouse:
### Short description
Some coverity fixes.

### Checklist

I have:
- [x] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/comfun.cc
+++ b/pdns/comfun.cc
@@ -68,9 +68,13 @@ struct SendReceive
   SendReceive(map<ComboAddress, namecount, ComboAddress::addressOnlyLessThan>& res) : d_res(res)
   {
     d_socket = socket(AF_INET, SOCK_DGRAM, 0);
+    if (d_socket < 0)
+       throw runtime_error("socket(AF_INET, SOCK_DGRAM, 0) failed: " + stringerror());
     int val=1;
-    setsockopt(d_socket, SOL_SOCKET, SO_REUSEADDR, &val, sizeof(val));
     
+    if (setsockopt(d_socket, SOL_SOCKET, SO_REUSEADDR, &val, sizeof(val)) != 0)
+       cerr<<"Failed to turn on SO_REUSEADDR: " << stringerror() << endl;
+
     for(unsigned int id =0 ; id < std::numeric_limits<uint16_t>::max(); ++id) 
       d_idqueue.push_back(id);
   }
@@ -198,8 +202,11 @@ struct SendReceiveRes
   {
     d_socket = socket(AF_INET, SOCK_DGRAM, 0);
     int val=1;
-    setsockopt(d_socket, SOL_SOCKET, SO_REUSEADDR, &val, sizeof(val));
-    connect(d_socket, (struct sockaddr*)&remote, remote.getSocklen());
+
+    if (setsockopt(d_socket, SOL_SOCKET, SO_REUSEADDR, &val, sizeof(val)) != 0)
+       cerr<<"Failed to turn on SO_REUSEADDR: " << stringerror() << endl;
+    if (connect(d_socket, (struct sockaddr*)&remote, remote.getSocklen()) != 0)
+       throw DNSException("connect(" + remote.toStringWithPort() + ") failed: " + stringerror());
     for(unsigned int id =0 ; id < std::numeric_limits<uint16_t>::max(); ++id) 
       d_idqueue.push_back(id);
   }
@@ -299,7 +306,7 @@ struct SendReceiveRes
 };
 
 
-void printStats()
+static void printStats()
 {
   while(!g_stop) {
     sleep(1);
@@ -309,7 +316,7 @@ void printStats()
   cerr<<"\n";
 }
 
-int parseZone(const std::string& str, unsigned int limit)
+static int parseZone(const std::string& str, unsigned int limit)
 {
   ZoneParserTNG zpt(str);
   zpt.disableGenerate();
@@ -353,7 +360,7 @@ int parseZone(const std::string& str, unsigned int limit)
   return 0;
 }
 
-int resolveNS(const std::string& fname)
+static int resolveNS(const std::string& fname)
 {
   string line;
   ifstream needres(fname);
@@ -385,7 +392,7 @@ int resolveNS(const std::string& fname)
   return EXIT_SUCCESS;
 }
 
-void readRESNames(const std::string& fname, map<DNSName, vector<ComboAddress>>& addrs)
+static void readRESNames(const std::string& fname, map<DNSName, vector<ComboAddress>>& addrs)
 {
   ifstream ifs(fname);
   if(!ifs)
--- a/pdns/dbdnsseckeeper.cc
+++ b/pdns/dbdnsseckeeper.cc
@@ -127,7 +127,8 @@ void DNSSECKeeper::clearAllCaches() {
   s_metacache.clear();
 }
 
-
+/* This function never fails, the return value is to simplify call chains
+   elsewhere so we can do mutate<cache> && clear<cache> */
 bool DNSSECKeeper::clearKeyCache(const DNSName& name)
 {
   WriteLock l(&s_keycachelock);
@@ -145,8 +146,8 @@ bool DNSSECKeeper::clearMetaCache(const DNSName& name)
 
 void DNSSECKeeper::clearCaches(const DNSName& name)
 {
-  clearKeyCache(name);
-  clearMetaCache(name);
+  (void)clearKeyCache(name);
+  (void)clearMetaCache(name);
 }
 
 bool DNSSECKeeper::addKey(const DNSName& name, const DNSSECPrivateKey& dpk, int64_t& id, bool active, bool published)
--- a/pdns/distributor.hh
+++ b/pdns/distributor.hh
@@ -101,8 +101,8 @@ public:
     }
 
     Question Q;
-    callback_t callback;
-    int id;
+    callback_t callback{nullptr};
+    int id{0};
   };
 
   bool isOverloaded() override
--- a/pdns/distributor.hh
+++ b/pdns/distributor.hh
@@ -181,6 +181,7 @@ template<class Answer, class Question, class Backend>MultiThreadDistributor<Answ
 // start of a new thread
 template<class Answer, class Question, class Backend>void MultiThreadDistributor<Answer,Question,Backend>::distribute(int ournum)
 {
+  // this is the longest name we can use, not a typo
   setThreadName("pdns/distributo");
 
   try {
--- a/pdns/dnsdist.cc
+++ b/pdns/dnsdist.cc
@@ -2006,6 +2006,8 @@ static void checkFileDescriptorsLimits(size_t udpBindsCount, size_t tcpBindsCoun
   }
 }
 
+static bool g_warned_ipv6_recvpktinfo = false;
+
 static void setUpLocalBind(std::unique_ptr<ClientState>& cs)
 {
   /* skip some warnings if there is an identical UDP context */
@@ -2039,9 +2041,13 @@ static void setUpLocalBind(std::unique_ptr<ClientState>& cs)
 
   if(!cs->tcp && IsAnyAddress(cs->local)) {
     int one=1;
-    setsockopt(fd, IPPROTO_IP, GEN_IP_PKTINFO, &one, sizeof(one));     // linux supports this, so why not - might fail on other systems
+    (void)setsockopt(fd, IPPROTO_IP, GEN_IP_PKTINFO, &one, sizeof(one)); // linux supports this, so why not - might fail on other systems
 #ifdef IPV6_RECVPKTINFO
-    setsockopt(fd, IPPROTO_IPV6, IPV6_RECVPKTINFO, &one, sizeof(one));
+    if (setsockopt(fd, IPPROTO_IPV6, IPV6_RECVPKTINFO, &one, sizeof(one)) < 0 &&
+        !g_warned_ipv6_recvpktinfo) {
+        warnlog("Warning: IPV6_RECVPKTINFO setsockopt failed: %s", stringerror());
+        g_warned_ipv6_recvpktinfo = true;
+    }
 #endif
   }
 
--- a/pdns/signingpipe.cc
+++ b/pdns/signingpipe.cc
@@ -32,7 +32,8 @@ int readn(int fd, void* buffer, unsigned int len)
       if(errno == EAGAIN || errno == EINTR) {
         if(pos==0)
           return -1;
-        waitForData(fd, -1); 
+        // error handled later
+        (void)waitForData(fd, -1);
         continue;
       }
       unixDie("Reading from socket in Signing Pipe loop");
--- a/pdns/test-svc_records_cc.cc
+++ b/pdns/test-svc_records_cc.cc
@@ -73,7 +73,6 @@ BOOST_AUTO_TEST_CASE(test_SvcParam_ctor_no_value) {
 BOOST_AUTO_TEST_CASE(test_SvcParam_ctor_value) {
     string val = "foobar";
     auto base64val = Base64Encode(val);
-    SvcParam param;
 
     BOOST_CHECK_THROW(SvcParam(SvcParam::mandatory, val), std::invalid_argument);
     BOOST_CHECK_THROW(SvcParam(SvcParam::alpn, val), std::invalid_argument);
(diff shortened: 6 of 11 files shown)


pdns/dbdnsseckeeper.cc
```diff
@@ -144,8 +144,8 @@ bool DNSSECKeeper::clearMetaCache(const DNSName& name)
 
 void DNSSECKeeper::clearCaches(const DNSName& name)
 {
-  clearKeyCache(name);
-  clearMetaCache(name);
+  if (!clearKeyCache(name) or !clearMetaCache(name))
```
rgacogne, line 147:
```suggestion
  if (!clearKeyCache(name) || !clearMetaCache(name))
```

pdns/signingpipe.cc
```diff
@@ -32,7 +32,9 @@ int readn(int fd, void* buffer, unsigned int len)
       if(errno == EAGAIN || errno == EINTR) {
         if(pos==0)
           return -1;
-        waitForData(fd, -1); 
+        // error handled later
+        (void)waitForData(fd, -1);
+          return -1;
```
rgacogne, line 37: Why the additional `return -1` here?

rgacogne (changes requested):
Thanks for this PR! I'm afraid I'm really uneasy with asserting on the return value of `chopOff()`, I'm not 100% sure a return value of `false` is an error in all these occurrences. And even if it is, it would turn a silent error that we might be handling later into a process termination, which sounds much worse.

pdns/distributor.hh
```diff
@@ -181,7 +183,7 @@ template<class Answer, class Question, class Backend>MultiThreadDistributor<Answ
 // start of a new thread
 template<class Answer, class Question, class Backend>void MultiThreadDistributor<Answer,Question,Backend>::distribute(int ournum)
 {
-  setThreadName("pdns/distributo");
+  setThreadName("pdns/distributor");
```
rgacogne, line 186: Pieter would know but I'm pretty sure we truncated this because there wasn't any space left.

pieterlexis: correct!

cmouse:
Fixed nits, dropped the assert patch.

pdns/dbdnsseckeeper.cc
```diff
@@ -144,8 +144,8 @@ bool DNSSECKeeper::clearMetaCache(const DNSName& name)
 
 void DNSSECKeeper::clearCaches(const DNSName& name)
 {
-  clearKeyCache(name);
-  clearMetaCache(name);
+  if (!clearKeyCache(name) || !clearMetaCache(name))
+     g_log<<Logger::Warning<<"Failed to clear cache for "<<name<<std::endl;
```
mind04, line 148: This is dead code. clearKeyCache and clearMetaCache always return true.

cmouse: wonder why they bother returning value at all? is it some future-planning thing?

Habbie: as discussed on IRC: the `return addKey(name, dspk, id, active, published) && clearKeyCache(name);` construction that is used in 15 variations throughout dbdnsseckeeper.cc needs the `true` return value.

cmouse:
fixed nits.

pdns/dbdnsseckeeper.cc
```diff
@@ -126,7 +126,8 @@ void DNSSECKeeper::clearAllCaches() {
   s_metacache.clear();
 }
 
-
+/* This function never fails, the return value is to simplify call chains
+   elsewhere to we can do mutate<cache> && clear<cache> */
```
Habbie, line 130:
```suggestion
   elsewhere so we can do mutate<cache> && clear<cache> */
```

pdns/dnsdist.cc
```diff
@@ -1850,9 +1850,12 @@ static void setUpLocalBind(std::unique_ptr<ClientState>& cs)
 
   if(!cs->tcp && IsAnyAddress(cs->local)) {
     int one=1;
-    setsockopt(fd, IPPROTO_IP, GEN_IP_PKTINFO, &one, sizeof(one));     // linux supports this, so why not - might fail on other systems
+
+    if (setsockopt(fd, IPPROTO_IP, GEN_IP_PKTINFO, &one, sizeof(one)) < 0)    // linux supports this, so why not - might fail on other systems
```
Habbie, line 1854: the comment suggests that this does not warrant a warning

pdns/test-svc_records_cc.cc
```diff
@@ -73,15 +73,8 @@ BOOST_AUTO_TEST_CASE(test_SvcParam_ctor_no_value) {
 BOOST_AUTO_TEST_CASE(test_SvcParam_ctor_value) {
     string val = "foobar";
     auto base64val = Base64Encode(val);
-    SvcParam param;
-
-    BOOST_CHECK_THROW(SvcParam(SvcParam::mandatory, val), std::invalid_argument);
-    BOOST_CHECK_THROW(SvcParam(SvcParam::alpn, val), std::invalid_argument);
-    BOOST_CHECK_THROW(SvcParam(SvcParam::no_default_alpn, val), std::invalid_argument);
```
Habbie, line 80: why is this block gone?

pdns/logger.cc
```diff
@@ -120,6 +120,9 @@ void Logger::log(const string &msg, Urgency u) noexcept
       catch (const runtime_error& e) {
         cerr << e.what() << endl;
       }
+      catch (boost::exception_detail::clone_impl<boost::exception_detail::error_info_injector<std::length_error> >& e2) {
```
Habbie, line 123: is there no less specific form we can catch? what throws this?

pdns/receiver.cc
```diff
@@ -566,8 +566,13 @@ int main(int argc, char **argv)
       
       if(::arg().mustDo("control-console"))
         dl=make_unique<DynListener>();
-      else
-        dl=std::unique_ptr<DynListener>(new DynListener(s_programname));
+      else {
+        try {
+           dl=std::unique_ptr<DynListener>(new DynListener(s_programname));
+        } catch (boost::exception_detail::clone_impl<boost::exception_detail::error_info_injector<boost::io::too_few_args> >& e) {
```
Habbie, line 572: any less specific exception we can catch?

Habbie: where is this exception coming from, anyway? :)

cmouse: it's what coverity complained, so i copied it from there.

Habbie: let's remove it

pdns/logger.cc
cmouse: ` S.ringAccount("logmessages",msg);`

cmouse:
```
1. fun_call_w_exception: Called function throws an exception of type boost::exception_detail::clone_impl<boost::exception_detail::error_info_injector<std::length_error> >.
 d_rings[name].account(item);
```

pdns/receiver.cc
cmouse: lets remove it indeed, the source for this is impossible

```
throw runtime_error((boost::format("Error: Unable to parse group ID %s") % cstr).str() );
```

pdns/dnsdist.cc
rgacogne: Indeed, I don't think we want a warning for this.

pieterlexis:
@cmouse ping

pdns/test-svc_records_cc.cc
cmouse: too aggressive cleanup =)

pdns/logger.cc
cmouse: ping @Habbie

Habbie: If you can put a comment in there explaining why we catch this very specific thing, it's fine by me.

Habbie (changes requested):
One last change requested. After that I'd like a review from at least Otto or Rémi just to be sure.

pdns/comfun.cc
```diff
@@ -69,8 +69,9 @@ struct SendReceive
   {
     d_socket = socket(AF_INET, SOCK_DGRAM, 0);
```
omoerbeek, line 70: While here check return `socket()` return val

pdns/logger.cc
cmouse: I think it is possible to happen if d_rings[name] does not exist, but I wonder if boost would let me catch std::length_error instead.

pdns/comfun.cc
```diff
@@ -69,8 +69,9 @@ struct SendReceive
...
     int val=1;
-    setsockopt(d_socket, SOL_SOCKET, SO_REUSEADDR, &val, sizeof(val));
-    
+    if (setsockopt(d_socket, SOL_SOCKET, SO_REUSEADDR, &val, sizeof(val)) != 0)
+       g_log<<Logger::Warning<<"Failed to turn on SO_REUSEADDR: " << stringerror() << endl;
```
omoerbeek, line 73: Are yo sure this is not going to spam the logs if it fails. Is it really useful to know?

pdns/comfun.cc
```diff
@@ -198,8 +199,10 @@ struct SendReceiveRes
   {
     d_socket = socket(AF_INET, SOCK_DGRAM, 0);
     int val=1;
-    setsockopt(d_socket, SOL_SOCKET, SO_REUSEADDR, &val, sizeof(val));
-    connect(d_socket, (struct sockaddr*)&remote, remote.getSocklen());
+    if (setsockopt(d_socket, SOL_SOCKET, SO_REUSEADDR, &val, sizeof(val)) != 0)
+       g_log<<Logger::Warning<<"Failed to turn on SO_REUSEADDR: " << stringerror() << endl;
+    if (connect(d_socket, (struct sockaddr*)&remote, remote.getSocklen()) != 0)
```
omoerbeek, line 204: A connect failure feels fatal to me

pdns/comfun.cc
```diff
@@ -198,8 +199,10 @@ struct SendReceiveRes
   {
     d_socket = socket(AF_INET, SOCK_DGRAM, 0);
     int val=1;
-    setsockopt(d_socket, SOL_SOCKET, SO_REUSEADDR, &val, sizeof(val));
-    connect(d_socket, (struct sockaddr*)&remote, remote.getSocklen());
+    if (setsockopt(d_socket, SOL_SOCKET, SO_REUSEADDR, &val, sizeof(val)) != 0)
```
omoerbeek, line 202: some worry about filling the log here. Also I'm wondering if it is of use to set `SO_REUSEADDR` on a client socket.

pdns/distributor.hh
```diff
@@ -98,6 +98,8 @@ public:
   {
     QuestionData(const Question& query): Q(query)
     {
+       id = 0;
```
omoerbeek, line 101: I prefer {} initializers with the fields themselves: nicething is anybody adds a ct and forget the init, things will keep on working.

pdns/logger.cc
omoerbeek: IMO this is too specific: if boost changes an implementation detail this won't match.

cmouse: removed as impossible.

omoerbeek (approved):

pdns/dnsdist.cc
```diff
@@ -1885,9 +1885,10 @@ static void setUpLocalBind(std::unique_ptr<ClientState>& cs)
 
   if(!cs->tcp && IsAnyAddress(cs->local)) {
     int one=1;
-    setsockopt(fd, IPPROTO_IP, GEN_IP_PKTINFO, &one, sizeof(one));     // linux supports this, so why not - might fail on other systems
+    (void)setsockopt(fd, IPPROTO_IP, GEN_IP_PKTINFO, &one, sizeof(one)); // linux supports this, so why not - might fail on other systems
 #ifdef IPV6_RECVPKTINFO
-    setsockopt(fd, IPPROTO_IPV6, IPV6_RECVPKTINFO, &one, sizeof(one));
+    if (setsockopt(fd, IPPROTO_IPV6, IPV6_RECVPKTINFO, &one, sizeof(one)) < 0)
+        warnlog("Warning: IPV6_RECVPKTINFO setsockopt failed: %s", stringerror());
```
rgacogne, line 1891: I don't really see the point of warning about this. There is nothing the user can do about it anyway.

rgacogne (changes requested):

pdns/dnsdist.cc
cmouse:
```
18:52 < cmouse> rgacogne: what happens if setting IPV6_RECVPKTINFO fails?
18:53 < cmouse> rgacogne: will it break anything? does something stop working?
19:00 < rgacogne> cmouse: we might not be able to properly detect the destination address for datagrams received over IPv6 for sockets bound to ::
19:01 < cmouse> rgacogne: point being; if setting that fails, will it cause something that might concern an admin?
19:01 < Habbie> yes
19:01 < Habbie> if they have :: in local-address
19:01 < cmouse> then I would suggest that warning would be logged
19:01 < Habbie> but
19:02 < Habbie> we really only want that warning once
19:02 < rgacogne> Habbie: it's for dnsdist in that case
19:02 < Habbie> oh
19:02 < rgacogne> anyway there is nothing an admin can do
19:03 < Habbie> except stop listening on ::
19:03 < Habbie> like the pdns faq used to warn
19:03 < rgacogne> and we would log that every time we start
19:03 < rgacogne> true
19:03 < Habbie> but, when would this happen? on some obsecure solaris fork?
19:03 < cmouse> if kernel does not support that option, most likely.
```

so what do we do?

Habbie: Warn once? (maybe have a (`static`?) `bool` to make sure it really is once.)

rgacogne: Warning only once would be nice but please keep it simple, I don't care that much :-)

cmouse: Fixed.

pdns/dnsdist.cc
```diff
@@ -1852,6 +1852,8 @@ static void checkFileDescriptorsLimits(size_t udpBindsCount, size_t tcpBindsCoun
   }
 }
 
+static bool g_warned_ipv6_recvpktinfo = FALSE;
```
rgacogne, line 1855:
```suggestion
static bool g_warned_ipv6_recvpktinfo = false;
```

pdns/dnsdist.cc
```diff
@@ -1885,9 +1887,13 @@ static void setUpLocalBind(std::unique_ptr<ClientState>& cs)
...
   if(!cs->tcp && IsAnyAddress(cs->local)) {
     int one=1;
-    setsockopt(fd, IPPROTO_IP, GEN_IP_PKTINFO, &one, sizeof(one));     // linux supports this, so why not - might fail on other systems
+    (void)setsockopt(fd, IPPROTO_IP, GEN_IP_PKTINFO, &one, sizeof(one)); // linux supports this, so why not - might fail on other systems
 #ifdef IPV6_RECVPKTINFO
-    setsockopt(fd, IPPROTO_IPV6, IPV6_RECVPKTINFO, &one, sizeof(one));
+    if (unlikely(setsockopt(fd, IPPROTO_IPV6, IPV6_RECVPKTINFO, &one, sizeof(one))) < 0 &&
+        !g_warned_ipv6_recvpktinfo) {
+        warnlog("Warning: IPV6_RECVPKTINFO setsockopt failed: %s", stringerror());
+        g_warned_ipv6_recvpktinfo == TRUE;
```
rgacogne, line 1895:
```suggestion
        g_warned_ipv6_recvpktinfo = true;
```

pdns/dnsdist.cc
```diff
@@ -1885,9 +1887,13 @@ static void setUpLocalBind(std::unique_ptr<ClientState>& cs)
 
   if(!cs->tcp && IsAnyAddress(cs->local)) {
     int one=1;
-    setsockopt(fd, IPPROTO_IP, GEN_IP_PKTINFO, &one, sizeof(one));     // linux supports this, so why not - might fail on other systems
+    (void)setsockopt(fd, IPPROTO_IP, GEN_IP_PKTINFO, &one, sizeof(one)); // linux supports this, so why not - might fail on other systems
 #ifdef IPV6_RECVPKTINFO
-    setsockopt(fd, IPPROTO_IPV6, IPV6_RECVPKTINFO, &one, sizeof(one));
+    if (unlikely(setsockopt(fd, IPPROTO_IPV6, IPV6_RECVPKTINFO, &one, sizeof(one))) < 0 &&
```
rgacogne, line 1892: I don't think we have the `unlikely` macro defined?

rgacogne (changes requested):

pdns/dnsdist.cc
cmouse: ah, my bad.

pdns/comfun.cc
Habbie: This does not look resolved.

Habbie (changes requested):
Found one older comment that was not resolved. Otherwise, this looks fine to me. Please do a rebase on fresh master so this runs in our current CI settings.

pdns/comfun.cc
cmouse: so - what should I do here? raise exception?

Habbie (approved):

Issue #9322, opened 2020-07-11
Remove archaic or offensive spelling (master, slave)


pyloor:
- Program: Authoritative, Recursor, dnsdist
 - Documentation
 - Issue type: Feature request

### Short description
Remove the archaic or offensive spelling like super-master, master or slave. there are better words for naming these entities.

### Usecase
Use a language which is less offensive or archaic

### Description
Replace the words like super-master, master or slave with a new set of namings. a few examples:
- Primary / Secondary
- Parent / Child

Note: I searched for a similiar or same issue through the issue tracker but could find sth. I only found a PR which removes some spelling errors but not those words. If there was a discussion or issue earlier which i didn't see, please point me in the right direction.

Contribution:
I am here to help and want to contribute the changes for this issue.

jpmens:
I hope DNS server vendors can agree on terms, e.g. by consulting [draft-knodel-terminology-02](https://tools.ietf.org/html/draft-knodel-terminology-02). ISC implemented the terms `primary` and `secondary` for BIND9. In order to not break configurations until much later releases these are currently aliased to `master` and `slave` respectively.

E.g. from [Bv9ARM](https://bind9.readthedocs.io/en/v9_16_4/reference.html#configuration-file-grammar):

> A primary zone has a master copy of the data for the zone and is able to provide authoritative answers for it. Type master is a synonym for primary.

Habbie:
Yes, using primary and secondary makes sense. We do need to invent something for 'supermaster' then (https://en.wiktionary.org/wiki/superprimary won't fly).

We also have several instances of 'blacklist' and 'whitelist', those should be easy to replace by 'blocklist', 'allowlist' or perhaps in some cases 'ignorelist' or 'droplist'.

The main technical (and user-impacting) challenge is the database schemas (they contain a `supermasters` table and a `domains.masters` column) and database contents (`domains.type` is currently expected to hold `MASTER` or `SLAVE` [or `NATIVE`]). For reading database contents, we should make sure we parse the type into an Enum as quickly as possible to avoid having to distribute the list of aliases (primary=master etc.) all over the code base (#9280 improved on this already).

The other 'user-facing' challenge is the HTTP API. I expect we can introduce aliases there as well on the input side, but the output side is tricker. Ideas for a decent migration strategy are welcome. Perhaps we can duplicate the `masters` field for a while?

[...]

PR #9669, opened 2020-11-03
auth pdns.conf, pdnsutil, pdns_control: modern aliases for master/slave

Habbie:
### Short description
Part of #9322. Not part of 4.4.0.

Not done yet:
* ~doc updates~ settings changes documented, other uses of the words in next PR
* ~perhaps reformat some `if` conditions that have now become huge (or move `primary` and `secondary` into static global bools to centralise the pain a bit)~ fixed by mirroring the settings centrally
* ~rename `allow-unsigned-supermaster`~ done
* ~add an upgrade note~ done

Not in scope for this PR:
* HTTP API
* constants, enums, source file names
* db schema
* db contents (metadata)
* going through all word usages in doc text
* log messages

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


zeha:
i like the new approach a lot more!

zeha:
@Habbie asked for this:

I think it'd be nice if --config showed only the new names. After #9659 that should be possible to do.

Habbie:
> d_unknownParams

which is in #9659

pdns/common_startup.cc
```diff
@@ -183,7 +187,8 @@ void declareArguments()
   ::arg().set("domain-metadata-cache-ttl","Seconds to cache domain metadata from the database")="60";
 
   ::arg().set("trusted-notification-proxy", "IP address of incoming notification proxy")="";
-  ::arg().set("slave-renotify", "If we should send out notifications for slaved updates")="no";
+  ::arg().set("slave-renotify", "If we should send out notifications for secondaried updates")="no";
+  ::arg().set("secondary-renotify", "If we should send out notifications for secondaried updates")="no";
```
jsoref, line 191: Not a big fan of secondaried

Habbie: Hmm, do you have an alternative suggestion?

jsoref: Can you explain what it is?

Habbie: secondaried means 'The secondary has transferred, from the primary, a new version of the zone.'

jsoref: Does this fit your intent?
```suggestion
  ::arg().set("secondary-renotify", "If we should send out notifications when secondary receives updates")="no";
```

Habbie: it's close, something with 'after' would be even closer

jsoref: Hmm, can you explain a bit?

Note:
`when` to me implies that it's the secondary sending the update

```suggestion
  ::arg().set("secondary-renotify", "If we should send out notifications after secondary receives updates")="no";
```

I worry that `after` doesn't indicate who is doing the sending

(I'm not particularly confident on this.)

Habbie: 'Whether we should send out notifications after receiving updates from the primary'

Habbie: from IRC, proposals to also rename the option while we're at it, because the current name is not 100% accurate: `notify-for-secondary`, `notify-by-secondary`, `notify-as-secondary`

Habbie: current proposal: secondary-do-notify: Whether this secondary sends out notifications after receiving updates from the primary for a zone

Habbie: we should say 'zone transfers' or 'XFR(s)' instead of 'updates'

jsoref: And `a primary` instead of `the primary` :-)

Habbie: s/the primary/a primary/

Habbie:
`replication-cycle-interval` -> `xfr-cycle-interval`

Habbie:
Rebased.

Habbie:
This is ready for review. Once merged I'll make additional PRs with changes in other area (SQL, API, non-functional wording in documentation, log messages).

docs/modes-of-operation.rst
```diff
@@ -86,6 +87,7 @@ the :ref:`pdns_control <running-pdnscontrol>` tool:
    happen in contrived configurations.
 
 .. _slave-operation:
+.. _secondary-operation:
 
 Slave operation
 ---------------
```
pieterlexis, line 93:
```suggestion
Secondary operation
-------------------
```

docs/modes-of-operation.rst
```diff
@@ -29,6 +29,7 @@ master or slave in zone transfers, or has a frontend attached for
 managing records etc.
 
 .. _master-operation:
+.. _primary-operation:
 
 Master operation
 ----------------
```
pieterlexis, line 35:
```suggestion
Primary operation
-----------------
```

docs/settings.rst
```diff
@@ -1771,6 +1840,22 @@ If the webserver should print arguments.
...
+.. versionchanged:: 4.5.0
+  This was called :ref:`setting-slave-cycle-interval` before 4.5.0.
+
+-  Integer
+-  Default: 60
+
+On a master, this is the amount of seconds between the master checking
+the SOA serials in its database to determine to send out NOTIFYs to the
+slaves. On slaves, this is the number of seconds between the slave
+checking for updates to zones.
```
pieterlexis, line 1857:
```suggestion
On a primary, this is the amount of seconds between the primary checking
the SOA serials in its database to determine to send out NOTIFYs to the
secondaries. On secondaries, this is the number of seconds between the secondary
checking for updates to zones.
```

pdns/receiver.cc
```diff
@@ -427,6 +427,19 @@ int main(int argc, char **argv)
         g_log<<Logger::Error<<"Unknown logging facility "<<::arg().asNum("logging-facility") <<endl;
     }
 
+    if (::arg().mustDo("master")) ::arg().set("primary")="yes";
```
pieterlexis, line 430: :heart:

pieterlexis (changes requested):
few nits, good otherwise

Habbie:
Rebased

pieterlexis (approved):

PR #9676, opened 2020-11-05
dnsdist: Implement Lua custom web endpoints

### LINKED ISSUE #9120: dnsdist: add support for a Lua HTTP API handler
- Program: dnsdist
- Issue type: Feature request

### Short description

It would be nice if dnsdist supported adding an API HTTP handler in Lua.

### Usecase

This would allow HTTP clients to update the Lua configuration in arbitrary - but controlled - ways without having to implement custom API endpoints for every possible feature. Currently this can only be achieved by opening a socket to the dnsdist console and writing Lua function calls to it.

### Description

For example, dnsdist could define a `/api/v1/servers/localhost/custom[/....]` endpoint that is handled in Lua. When an HTTP request is received, dnsdist calls a user-defined `customhttphandler` Lua function to handle the request.

The Lua callback function could look as follows:

```lua
function customhttphandler(req)
    print(req.method, req.path, req.body)
    return 200, {["Content-Type"] = "text/plain"}, "hello world!"
end 
```

The request object would be a table for easy future extensibility without an explosion of positional arguments.  Suggested attributes to include in the request object:

- method: the HTTP method
- path: the full request path
- body: the raw request body as string
- params: parsed GET params
- form: parsed POST values if this is a POST request and the content type is `application/x-www-form-urlencoded`
- subpath: the subpath under `/api/v1/servers/localhost/custom`, allowing for custom routing in Lua.

The response args are:

- The HTTP status code
- An optional table of extra HTTP headers to set on the response (nil allowed)
- A string with the response contents

In order to limit the time spent in the Lua code (which will block dnsdist), dnsdist buffers the whole request with body before calling the Lua code. This processing is all done in a separate thread. For the same reason, the Lua handler returns the whole response at once, instead of streaming the response as is common in modern frameworks.

If the `customhttphandler` callback is not defined, dnsdist returns a 404 and an appropriate message.

The HTTP request is subject to API key validation.

rgacogne:
### Short description

Closes #9120.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdistdist/docs/reference/config.rst
+++ b/pdns/dnsdistdist/docs/reference/config.rst
@@ -362,6 +362,26 @@ Webserver configuration
   - ...
   But not queries for /foobar or /foo/bar.
 
+  A sample handler function could be:
+
+  .. code-block:: lua
+
+    function customHTTPHandler(req, resp)
+      local get = req.getvars
+      local headers = req.headers
+
+      if req.path ~= '/foo' or req.version ~= 11 or req.method ~= 'GET' or get['param'] ~= '42' or headers['customheader'] ~= 'foobar' then
+        resp.status = 500
+        return
+      end
+
+      resp.status = 200
+      resp.body = 'It works!'
+      resp.headers = { ['Foo']='Bar'}
+    end
+
+    registerWebHandler('/foo', customHTTPHandler)
+
   :param str path: Path to register the handler for.
   :param function handler: The Lua function to register.
 
--- a/pdns/dnsdistdist/docs/reference/config.rst
+++ b/pdns/dnsdistdist/docs/reference/config.rst
@@ -370,7 +370,7 @@ Webserver configuration
       local get = req.getvars
       local headers = req.headers
 
-      if req.path ~= '/foo' or req.version ~= 11 or req.method ~= 'GET' or get['param'] ~= '42' or headers['customheader'] ~= 'foobar' then
+      if req.path ~= '/foo' or req.version ~= 11 or req.method ~= 'GET' or get['param'] ~= '42' or headers['custom'] ~= 'foobar' then
         resp.status = 500
         return
       end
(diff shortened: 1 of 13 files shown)


pdns/dnsdistdist/docs/reference/web.rst
```diff
@@ -0,0 +1,52 @@
...
+
+  .. attribute:: WebRequest.version
+
+    The HTTP version of this query, as an integer.
+
+.. class:: WebResponse
+
+  Represent a HTTP response.
+
+  .. attribute:: WebRequest.body
```
Habbie, line 42:
```suggestion
  .. attribute:: WebResponse.body
```

? (and two more)

pdns/dnsdistdist/docs/reference/config.rst
```diff
@@ -349,6 +349,22 @@ Webserver configuration
   * ``custom_headers={[str]=str,...}``: map of string - Allows setting custom headers and removing the defaults.
   * ``acl=newACL``: string - List of IP addresses, as a string, that are allowed to open a connection to the web server. Defaults to "127.0.0.1, ::1".
 
+.. function:: registerWebHandler(path, handler)
+
+  .. versionadded: 1.6.0
+
+  Register a function named ``handler`` that will be called for every query sent to the exact ``path`` path. The function will receive a :class:`WebRequest` object
+  and a :class:`WebResponse` object, representing respectively the HTTP request received and the HTTP response to send.
```
Habbie, line 357: could use an example of handling a simple request and setting resp.body etc.

rgacogne: That makes total sense, will do!

rgacogne:
Weird test-auth-regress-bind error in bind-dnssec-nsec3-both, during verify-dnssec-zone:
```
Nov 06 10:40:12 AXFR of domain 'tsig.com' allowed: client IP 127.0.0.1:46172 is in allow-axfr-ips
AXFR failed.
Syntax error, could not parse the RR's TTL at 0
Nov 06 10:40:12 Received serial number updates for 1 zone, had 0 timeouts
```

wojas:
Nice!

PR #9680, opened 2020-11-06
rec: If a.b.c CNAME x.a.b.c is encoutered, switch off QM for x.a.b.c resolve

omoerbeek:
### Short description

The is an optimization, avoiding hitting maxdepth followeed by no-QM fallback.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [X] documented the code

### DIFF
--- a/pdns/syncres.cc
+++ b/pdns/syncres.cc
@@ -1498,6 +1498,13 @@ bool SyncRes::doCNAMECacheCheck(const DNSName &qname, const QType &qtype, vector
         throw ImmediateServFailException(msg);
       }
 
+      if (newTarget.isPartOf(qname)) {
+        // a.b.c. CNAME x.a.b.c will go to great depths with QM on
+        string msg = "got a CNAME referral (from cache) to child, switching off QM";
+        LOG(prefix<<qname<<": "<<msg<<endl);
+        setQNameMinimization(false);
+      }
+
       // Check to see if we already have seen the new target as a previous target
       if (scanForCNAMELoop(newTarget, ret)) {
         string msg = "got a CNAME referral (from cache) that causes a loop";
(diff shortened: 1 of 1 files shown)


pdns/syncres.cc
```diff
@@ -1498,10 +1498,20 @@ bool SyncRes::doCNAMECacheCheck(const DNSName &qname, const QType &qtype, vector
         throw ImmediateServFailException(msg);
       }
 
+      const bool doingQM = getQNameMinimization();
+
+      if (newTarget.isPartOf(qname)) {
```
rgacogne, line 1503:
```suggestion
      if (doingQM && newTarget.isPartOf(qname)) {
```

pdns/syncres.cc
```diff
@@ -3701,9 +3712,18 @@ void SyncRes::handleNewTarget(const std::string& prefix, const DNSName& qname, c
     return;
   }
 
+  const bool doingQM = getQNameMinimization();
+
+  if (newtarget.isPartOf(qname)) {
```
rgacogne, line 3717:
```suggestion
  if (doingQM && newtarget.isPartOf(qname)) {
```

rgacogne (approved):

omoerbeek:
We're seeing unexplained bulk test failures. This will need more investigation.

omoerbeek:
So what was happening: `getQNameMinimization()` reflect the top-level state of the current resolve action. Deeper nested calls might already have switched off QM. So the "optimized" code was switching QM on in some cases, leading to loops. The QM code needs to make a more clear distinction between teh top-level and deeper "doing QM" properties.
Settle for now for the more simple original approach, just switching off QM for the rest of this resolve.

Issue #9514, opened 2020-09-23
AXFR and IXFR requests are logged at Error level


tih:
- Program: Authoritative
 - Issue type: Bug report

### Short description
When normal (expected) AXFR and IXFR requests arrive, they are logged at Error level, which causes console noise. There are also log entries that are made at the Warning level, that I would suggest be at Info.  I suspect that the logging related to AXFR and IXFR was accidentally left in a state that was useful during development, and would suggest that someone go through the pdns/tcpreceiver.cc file, and check all the Logger:: calls there.

### Environment
 - Operating system: NetBSD-current
 - Software version: git HEAD from master branch as per 2020-09-23
 - Software source: locally built from a clone of the PowerDNS git repo

### Steps to reproduce
1. Observation.

### Expected behaviour
Normal operations should not trigger Warning or Error log entries.

### Actual behaviour
In the log entries shown here, the "127.0.0.1" is because I use dnsdist to filter the requests, and it then performs the actual xfer call into the authoritative server locally. The requests are really from the secondary server for the zone, at my provider.

Sep 23 11:07:30 barsoom pdns[1775]: IXFR of domain 'hamartun.priv.no' initiated by 127.0.0.1:63701 with serial 2020091701
Sep 23 11:07:30 barsoom pdns[1775]: AXFR of domain 'hamartun.priv.no' allowed: client IP 127.0.0.1:63701 is in allow-axfr-ips
Sep 23 11:07:30 barsoom pdns[1775]: IXFR of domain 'hamartun.priv.no' to 127.0.0.1:63701 finished

The first and last of these are at Error level, which causes them to be printed on the console with my syslogd configuration. The middle one is at Warning level, which doesn't, but I still think the level is wrong; allowing an IP that's explicitly allowed is normal operations, and deserves a simple Info level log entry.

### Other information
None.

klaus3000:
Comment from an administrator: Very often I have to look up the logs and inspect XFRs. When do they start, when do they finish. Both, on master and slave. Because often customers complain that zones were not fetched, or zones could not be fetched. Hence, if log levels are changes, the default PowerDNS log level should log XFR related tasks. Thanks

Habbie:
@tih I agree with Info. Do you want to write a PR? :)

@klaus3000 that suggests that either you need to run at a higher log level than default, or we need more fine grained logging controls. I would entertain a PR for better controls if somebody made one.

tih:
My suggested modifications are in PR #9658 .

[...]

PR #9688, opened 2020-11-09
Pdns AXFR, IXFR loglevels

mind04:
### Short description
Decrease the verbosity of AXFR and IXFR in the logs

Inspired by and successor of #9658 This pull will change the log levels for both, primary and secondary.

Closes #9658 #9514

Most XFR related events are now logged at notice level. And will not show up in the logs at the default log level.

This is the only (non error) message still logged at warning level on the primary:
```
Nov 09 13:33:31 AXFR of domain 'example.com' initiated by 127.0.0.2:36543
```

The secondary is only showing this at warning level:
```
Nov 09 13:33:31 AXFR done for 'example.com', zone committed with serial number 2005092501
```

You can argue about these two lines, but in my opinion it is important to not completely hide XFR at default log level.

### Checklist
I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/tcpreceiver.cc
+++ b/pdns/tcpreceiver.cc
@@ -346,7 +346,7 @@ void TCPNameserver::doConnection(int fd)
       {
         std::lock_guard<std::mutex> l(s_plock);
         if(!s_P) {
-          g_log<<Logger::Error<<"TCP server is without backend connections, launching"<<endl;
+          g_log<<Logger::Warning<<"TCP server is without backend connections, launching"<<endl;
           s_P=make_unique<PacketHandler>();
         }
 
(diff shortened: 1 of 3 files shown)


tih:
I like it! Much more complete than my quick hack in #9658, and I also never thought of using "Notice" actively.

pdns/tcpreceiver.cc
```diff
@@ -544,26 +544,20 @@ int TCPNameserver::doAXFR(const DNSName &target, std::unique_ptr<DNSPacket>& q,
...
   bool narrow;
   bool NSEC3Zone=false;
   if(securedZone && dk.getNSEC3PARAM(target, &ns3pr, &narrow)) {
     NSEC3Zone=true;
     if(narrow) {
-      g_log<<Logger::Error<<"Not doing AXFR of an NSEC3 narrow zone '"<<target<<"' for "<<q->getRemote()<<endl;
-      noAXFRBecauseOfNSEC3Narrow=true;
+      g_log<<Logger::Warning<<"Not doing AXFR of an NSEC3 narrow zone '"<<target<<"' for "<<q->getRemote()<<endl;
+      outpacket->setRcode(RCode::Refused);
+      // FIXME: should actually figure out if we are auth over a zone, and send out 9 if we aren't
```
Habbie, line 555: I thought 9 was reserved for RFC2136

Habbie (approved):
Approved, with one open question (about a comment that was already there). I've asked the other vendors if they use NOTAUTH like that.

pdns/tcpreceiver.cc
mind04: There is nothing to figure out here. The zone is narrow so it does exist. I will remove the comment.

Habbie: ah yes, in this context there is nothing to figure out. Thanks :)

Habbie (approved):

PR #9694, opened 2020-11-09
Several docker image improvements

pieterlexis:
### Short description
This PR contains 3 improvements:

1. It copies less into the build environment, improving caching somewhat
2. It makes dnsdist-resolve not print each second
3. It adds templating to the startup script.

The last item needs docs, but I did not find a readme for the docker images.. so where will we do this?

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/Dockerfile-auth
+++ b/Dockerfile-auth
@@ -10,7 +10,7 @@ RUN apt-get update && apt-get -y dist-upgrade && apt-get clean
 RUN apt-get update && apt-get -y dist-upgrade && apt-get install -y  --no-install-recommends devscripts equivs git && apt-get clean
 
 # import everything - this could be pdns.git OR an auth tarball!
-COPY . /source
+COPY builder-support /source/builder-support
 
 # TODO: control file is not in tarballs at all right now
 RUN mk-build-deps -i -t 'apt-get -y -o Debug::pkgProblemResolver=yes --no-install-recommends' /source/builder-support/debian/authoritative/debian-buster/control && \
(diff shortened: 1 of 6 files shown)


Makefile.docker
```diff
@@ -1,15 +1,22 @@
...
 	docker build --build-arg VERSION=$(VERSION) -t powerdns/pdns-auth-master:$(VERSION)-$(REVISION) --rm -f Dockerfile-auth .
+
+recursor:
 	docker build --build-arg VERSION=$(VERSION) -t powerdns/pdns-recursor-master:$(VERSION)-$(REVISION) --rm -f Dockerfile-recursor .
+
+dnsdist:
 	docker build --build-arg VERSION=$(VERSION) -t powerdns/dnsdist-master:$(VERSION)-$(REVISION) --rm -f Dockerfile-dnsdist .
 
+build: auth recursor dnsdist
+	@true
```
Habbie, line 18: I don't think you need this line

Dockerfile-auth
```diff
@@ -67,16 +67,11 @@ COPY dockerdata/startup.py /usr/local/sbin/pdns_server-startup
 # Config file(s) from builder
 # Should not grab this from builder - since it isn't being built
 COPY --from=builder /source/dockerdata/pdns.conf /etc/powerdns/
-RUN mkdir -p /etc/powerdns/pdns.d /var/run/pdns
-RUN touch /etc/powerdns-api.conf && chown 953 /etc/powerdns-api.conf
-RUN ln -s /etc/powerdns-api.conf /etc/powerdns/pdns.d/api.conf
```
Habbie, line 72: why remove this?

pieterlexis: startup.py now takes care of writing the API config if needed (I'll check)

Habbie: I believe that, but this construct was designed so that startup.py would not have to write into /etc/powerdns just to set an API key. I understand that for startup.py's templating, this is expected, but just for the API key, somebody mentioned that I should not be writing into somebody's bindmounted /etc/powerdns.

pieterlexis: The use should not set that env var then :). We won't write into the config dir then

Habbie: But then they cannot set the API key..

pieterlexis: if their bind-mounting.... they can

pieterlexis: I think we're fine actually!

If a user bind-mounts over `/etc/powerdns` or `/etc/powerdns/conf.d` (or the equivs for the rec and dnsdist) and does not set the TEMPLATE_FILES or `*_API_KEY` variables, the startup will never write to the config dirs

PR #9697, opened 2020-11-10
rec: do not send overly long NOD lookups

Habbie:
### Short description
This avoids a confusing 'too long' log message that looks like the user's request failed.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


Habbie:
Forgot to add it to the SNMP MIB, will do that soon.

pdns/recursordist/rec_metrics.hh
```diff
@@ -251,6 +251,9 @@ private:
     {"no-packet-error",
       MetricDefinition(PrometheusMetricType::counter,
         "Number of erroneous received packets")},
+    {"nod-lookups-dropped-oversize",
+      MetricDefinition(PrometheusMetricType::counter,
+        "Number of NOD lookups dropped because of their length")},
```
omoerbeek, line 256:
```suggestion
        "Number of NOD lookups dropped because they would exceed the maximum name length")},
```

pdns/pdns_recursor.cc
```diff
@@ -1218,7 +1218,14 @@ static void sendNODLookup(const DNSName& dname)
     // Send a DNS A query to <domain>.g_nodLookupDomain
     static const QType qt(QType::A);
     static const uint16_t qc(QClass::IN);
-    DNSName qname = dname + g_nodLookupDomain;
+    DNSName qname;
+    try {
+      qname = dname + g_nodLookupDomain;
+    }
+    catch(const std::range_error &e) {
```
omoerbeek, line 1225: Any other exception that could happen here?

Habbie: Unsure - perhaps we could do a separate catchall that does log, so we'll find out some day.

Habbie: DNSName can also throw `out_of_range` and `runtime_error`, but not in the code paths we are triggering here

Habbie:
> Forgot to add it to the SNMP MIB, will do that soon.

Done, and squashed.

omoerbeek (approved):

PR #9698, opened 2020-11-10
namespaces.hh cleanup

omoerbeek:
### Short description

This reduces the number of files included in `namespaces.hh` to only the most popular.

Also switch to using the `boost` namespace explicitly, with the exception of `tuple`, `make_tuple` and `tie`.
For these we should be able to uste the 'std::' ones, but that produces compile errors on Travis (which uses trusty).

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code


rgacogne (approved):
Great work!

PR #9699, opened 2020-11-10
Rec: refresh almost expired

### LINKED ISSUE #440: Feature request: DNS Prefetching
This is a feature request for DNS prefetching in PowerDNS Recursor. Attached is a proof of concept implementation. It seems to work. But it's hard to say whether it improves the average latency from the customer perspective. Prefetching simply ensures that popular domain names are always fresh in the Recursors cache.

The attached patch already includes the patch suggestion from [http://wiki.powerdns.com/trac/ticket/438] which is a requirement for the prefetch proof of concept patch.

omoerbeek:
### Short description

This add two related mechanisms to the record cache:

1. When run normally, determining that an RC entry is "almost expired" and adding it to a queue of "to be refetched" entries (filtering duplicates).

2. Running a query done by SyncRes in a mode that considers "almost expired" entries as expired, triggering re-fetch of the RC "almost expired" entries used to resolve the query (including e.g. NS records that are used and almost expired).

The queue of almost expired entries is processed by the handler thread atm (this should change). It will refetch the names in "refetch almost expired" mode.

The whole mechanism is enabled by setting `refresh-on-ttl-perc`, e.g. to 10

If you want to play with this, _disable the packetcache_, since the entries in the PC and the RC mostly "dance to the same beat" and the trigger mechanism is only implemented for the RC so far. If requests are served by PC hits, the RC refresh mechanism won't be triggered. *This is no longer needed, refetch triggered by PC is implemented now but tested only very lightly!*

This is WIP, but creating a draft PR so people can take a look. I suspect I won't be able to do much work on this the coming time.

Todo: 

- [x] use own thread instead of the handler thread,
- [x] trigger refetch from the PC
- [x] mark RC(PC) entries as already in the process of being re-fetched so the adding to the queue only happens once per record and not each time the RC/PC record is used and almost expired (done for PC where the impact is highest) 
- [x] performance measurements
- [ ] study interaction with client specific stuff like EDNS client subnet
- [x] check how this behaves with very low TTLs
- [x] logging (wtithout spamming) and catching exceptions

Should fix #440 in the end (after making it better and validating it is beneficial).

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [x] documented the code

### DIFF
--- a/pdns/pdns_recursor.cc
+++ b/pdns/pdns_recursor.cc
@@ -99,6 +99,7 @@
 #include "query-local-address.hh"
 
 #include "rec-snmp.hh"
+#include "rec-taskqueue.hh"
 
 #ifdef HAVE_SYSTEMD
 #include <systemd/sd-daemon.h>
@@ -2175,6 +2176,8 @@ static void startDoResolve(void *p)
     g_log<<Logger::Error<<"Any other exception in a resolver context "<< makeLoginfo(dc) <<endl;
   }
 
+  runTaskOnce(g_logCommonErrors);
+
   g_stats.maxMThreadStackUsage = max(MT->getMaxStackUsage(), g_stats.maxMThreadStackUsage);
 }
 
@@ -3356,6 +3359,9 @@ static void doStats(void)
   auto rc_stats = g_recCache->stats();
   double r = rc_stats.second == 0 ? 0.0 : (100.0 * rc_stats.first / rc_stats.second);
   uint64_t negCacheSize = g_negCache->size();
+  auto taskPushes = getTaskPushes();
+  auto taskExpired = getTaskExpired();
+  auto taskSize = getTaskSize();
 
   if(g_stats.qcounter && (cacheHits + cacheMisses) && SyncRes::s_queries && SyncRes::s_outqueries) {
     g_log<<Logger::Notice<<"stats: "<<g_stats.qcounter<<" questions, "<<
@@ -3388,6 +3394,7 @@ static void doStats(void)
       }
     }
 
+    g_log<<Logger::Notice<<"stats: tasks pushed/expired/queuesize: " << taskPushes << '/' << taskExpired << '/' << taskSize << endl; 
     time_t now = time(0);
     if(lastOutputTime && lastQueryCount && now != lastOutputTime) {
       g_log<<Logger::Notice<<"stats: "<< (SyncRes::s_queries - lastQueryCount) / (now - lastOutputTime) <<" qps (average over "<< (now - lastOutputTime) << " seconds)"<<endl;
@@ -3420,6 +3427,8 @@ static void houseKeeping(void *)
       return;
     }
     s_running=true;
+
+    runTaskOnce(g_logCommonErrors);
 
     struct timeval now, past;
     Utility::gettimeofday(&now, nullptr);
@@ -3776,7 +3785,6 @@ static vector<pair<DNSName, uint16_t> >& operator+=(vector<pair<DNSName, uint16_
   a.insert(a.end(), b.begin(), b.end());
   return a;
 }
-
 
 /*
   This function should only be called by the handler to gather metrics, wipe the cache,
@@ -4536,6 +4544,9 @@ static int serviceMain(int argc, char*argv[])
   SyncRes::s_maxtotusec=1000*::arg().asNum("max-total-msec");
   SyncRes::s_maxdepth=::arg().asNum("max-recursion-depth");
   SyncRes::s_rootNXTrust = ::arg().mustDo( "root-nx-trust");
+  SyncRes::s_refresh_ttlperc = ::arg().asNum("refresh-on-ttl-perc");
+  RecursorPacketCache::s_refresh_ttlperc = SyncRes::s_refresh_ttlperc;
+
   if(SyncRes::s_serverID.empty()) {
     SyncRes::s_serverID = myHostname;
   }
@@ -5403,6 +5414,7 @@ int main(int argc, char **argv)
     ::arg().setSwitch("nothing-below-nxdomain", "When an NXDOMAIN exists in cache for a name with fewer labels than the qname, send NXDOMAIN without doing a lookup (see RFC 8020)")="dnssec";
     ::arg().set("max-generate-steps", "Maximum number of $GENERATE steps when loading a zone from a file")="0";
     ::arg().set("record-cache-shards", "Number of shards in the record cache")="1024";
+    ::arg().set("refresh-on-ttl-perc", "If a record is requested from the cache and only this % of original TTL remains, refetch") = "0";
 
 #ifdef NOD_ENABLED
     ::arg().set("new-domain-tracking", "Track newly observed domains (i.e. never seen before).")="no";
@@ -5535,3 +5547,4 @@ int main(int argc, char **argv)
 
   return ret;
 }
+
--- a/pdns/recpacketcache.cc
+++ b/pdns/recpacketcache.cc
@@ -8,11 +8,14 @@
 #include "cachecleaner.hh"
 #include "dns.hh"
 #include "namespaces.hh"
+#include "rec-taskqueue.hh"
 
 RecursorPacketCache::RecursorPacketCache()
 {
   d_hits = d_misses = 0;
 }
+
+unsigned int RecursorPacketCache::s_refresh_ttlperc{0};
 
 int RecursorPacketCache::doWipePacketCache(const DNSName& name, uint16_t qtype, bool subtree)
 {
@@ -41,7 +44,7 @@ int RecursorPacketCache::doWipePacketCache(const DNSName& name, uint16_t qtype, 
 
 bool RecursorPacketCache::qrMatch(const packetCache_t::index<HashTag>::type::iterator& iter, const std::string& queryPacket, const DNSName& qname, uint16_t qtype, uint16_t qclass)
 {
-  // this ignores checking on the EDNS subnet flags! 
+  // this ignores checking on the EDNS subnet flags!
   if (qname != iter->d_name || iter->d_type != qtype || iter->d_class != qclass) {
     return false;
   }
@@ -60,10 +63,20 @@ bool RecursorPacketCache::checkResponseMatches(std::pair<packetCache_t::index<Ha
 
     if (now < iter->d_ttd) { // it is right, it is fresh!
       *age = static_cast<uint32_t>(now - iter->d_creation);
+      // we know ttl is > 0
+      uint32_t ttl = static_cast<uint32_t>(iter->d_ttd - now);
+      if (s_refresh_ttlperc > 0 && !iter->d_submitted) {
+        const uint32_t deadline = iter->getOrigTTL() * s_refresh_ttlperc / 100;
+        const bool almostExpired = ttl <= deadline;
+        if (almostExpired) {
+          iter->d_submitted = true;
+          pushTask(qname, qtype, iter->d_ttd);
+        }
+      }
       *responsePacket = iter->d_packet;
       responsePacket->replace(0, 2, queryPacket.c_str(), 2);
       *valState = iter->d_vstate;
-    
+
       const size_t wirelength = qname.wirelength();
       if (responsePacket->size() > (sizeof(dnsheader) + wirelength)) {
         responsePacket->replace(sizeof(dnsheader), wirelength, queryPacket, sizeof(dnsheader), wirelength);
@@ -83,7 +96,7 @@ bool RecursorPacketCache::checkResponseMatches(std::pair<packetCache_t::index<Ha
       return true;
     }
     else {
-      moveCacheItemToFront<SequencedTag>(d_packetCache, iter); 
+      moveCacheItemToFront<SequencedTag>(d_packetCache, iter);
       d_misses++;
       break;
     }
@@ -158,14 +171,14 @@ void RecursorPacketCache::insertResponsePacket(unsigned int tag, uint32_t qhash,
     iter->d_ttd = now + ttl;
     iter->d_creation = now;
     iter->d_vstate = valState;
-
+    iter->d_submitted = false;
     if (pbdata) {
       iter->d_pbdata = std::move(*pbdata);
     }
 
     break;
   }
-  
+
   if(iter == range.second) { // nothing to refresh
     struct Entry e(qname, std::move(responsePacket), std::move(query));
     e.d_qhash = qhash;
@@ -175,7 +188,7 @@ void RecursorPacketCache::insertResponsePacket(unsigned int tag, uint32_t qhash,
     e.d_creation = now;
     e.d_tag = tag;
     e.d_vstate = valState;
-
+    e.d_submitted = false;
     if (pbdata) {
       e.d_pbdata = std::move(*pbdata);
     }
--- a/pdns/recpacketcache.hh
+++ b/pdns/recpacketcache.hh
@@ -49,6 +49,8 @@ using namespace ::boost::multi_index;
 class RecursorPacketCache: public PacketCache
 {
 public:
+  static unsigned int s_refresh_ttlperc;
+
   struct PBData {
     std::string d_message;
     std::string d_response;
@@ -92,11 +94,16 @@ private:
     uint16_t d_type;
     uint16_t d_class;
     mutable vState d_vstate;
+    mutable bool d_submitted;   // whether this entry has been queued for refetch
     inline bool operator<(const struct Entry& rhs) const;
 
     time_t getTTD() const
     {
       return d_ttd;
+    }
+
+    uint32_t getOrigTTL() const {
+      return d_ttd - d_creation;
     }
   };
 
--- a/pdns/recursor_cache.cc
+++ b/pdns/recursor_cache.cc
@@ -13,6 +13,7 @@
 #include "recursor_cache.hh"
 #include "namespaces.hh"
 #include "cachecleaner.hh"
+#include "rec-taskqueue.hh"
 
 MemRecursorCache::MemRecursorCache(size_t mapsCount) : d_maps(mapsCount)
 {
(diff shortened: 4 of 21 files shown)


pdns/recursordist/taskqueue.cc
```diff
@@ -0,0 +1,70 @@
...
+  std::unique_lock<std::mutex> lock(d_mutex);
+  ResolveTask ret = d_queue.get<SequencedTag>().front();
+  d_queue.get<SequencedTag>().pop_front();
+  return ret;
+}
+
+void TaskQueue::runner()
+{
+  while (!d_queue.empty()) {
+    ResolveTask task = pop();
```
rgacogne, line 56: There will be a race here if we ever get more than one reader.

rgacogne (commented):
> Running a query done by SyncRes in a mode that considers "almost expired" entries as expired, triggering re-fetch of the RC "almost expired" entries used to resolve the query (including e.g. NS records that are used and almost expired).

I'm a bit worried about this concept, since we will consider any "almost expired" entries as expired while attempting to refresh one specific entry. Perhaps it would be wiser to ignore only the specific target of the resolution, as I think that any other "almost expired" entry will either be updated automatically if we get a fresher version during the resolution of the initial target, or added to the list of entries to refresh?

pdns/recursordist/taskqueue.cc
omoerbeek: Yes, agreed. I took a shortcut assuming there is only one reader.

omoerbeek:
> > Running a query done by SyncRes in a mode that considers "almost expired" entries as expired, triggering re-fetch of the RC "almost expired" entries used to resolve the query (including e.g. NS records that are used and almost expired).
> 
> I'm a bit worried about this concept, since we will consider any "almost expired" entries as expired while attempting to refresh one specific entry. Perhaps it would be wiser to ignore only the specific target of the resolution, as I think that any other "almost expired" entry will either be updated automatically if we get a fresher version during the resolution of the initial target, or added to the list of entries to refresh?

An issue isis that intermediate records used might be almost expired while the target is not. Do you want to force refresh of these?

rgacogne:
> An issue isis that intermediate records used might be almost expired while the target is not. Do you want to force refresh of these?

It looked to me like the current code would have noticed they were going to expire soon during the resolution process and add them to the task list, but perhaps I'm mistaken? I think I would prefer that, that they are added to the list but still used for the current resolution, if they are still usable.

omoerbeek:
> > An issue isis that intermediate records used might be almost expired while the target is not. Do you want to force refresh of these?
> 
> It looked to me like the current code would have noticed they were going to expire soon during the resolution process and add them to the task list, but perhaps I'm mistaken? I think I would prefer that, that they are added to the list but still used for the current resolution, if they are still usable.

During regular resolving all the entries in the RC that are consulted and almost expired are added to the queue but still used for resolving since they are not really expired.

In refresh mode (used by the queue runner) the entries in the RC that are almost expired are considered expired (-1 is returned by `get()`) so they trigger a refresh. 

So both RC entries needed for resolving a name and the name itself  end up in the queue (if considered almost expired). The queue is processed in order, so e.g. the NS records needed will be refreshed before the final name and won't be re-refreshed during the resolving of the final name: after they are refreshed that are not amost expired anymore.

omoerbeek:
Refetch from PacketCache implemented. Beware: only tested very lightly!

omoerbeek:
There is at least one bug: during the "almost expired period" multiple identical tasks can be queued. Those extra queued tasks actual cause work once the previous one is done: if there is an identical in the queue, the new one will be ignored, but if no more identical task is in the queue, the new one will be accepted in the queue. Since the PC entry isn't refreshed until expires, this is actually a likely scenario for popular entries.

omoerbeek:
Packet issue should be fixed with latest commit.

omoerbeek:
Performance measures show for 5 dnsperf runs with 1000 names of 300s a slight decrease in latency and the stddev of the latency. Note that the variation of the test runs is there, but also the trend afaiac.

[refetch.pdf](https://github.com/PowerDNS/pdns/files/5765028/refetch.pdf)

omoerbeek:
I believe this is now in a state ready for review.

pdns/recursordist/taskqueue.cc
```diff
@@ -0,0 +1,122 @@
...
+      g_log << Logger::Error << "Exception while running the background task queue: " << e.what() << endl;
+    }
+    catch (const PDNSException& e) {
+      g_log << Logger::Error << "Exception while running the background task queue: " << e.reason << endl;
+    }
+    catch (const ImmediateServFailException& e) {
+      g_log << Logger::Error << "Exception while running the background task queue: " << e.reason << endl;
+    }
+    catch (const PolicyHitException& e) {
+      g_log << Logger::Error << "Policy hit while running the background task queue" << endl;
```
jsoref, line 86: Is this a good thing or a bad thing?

pdns/recursor_cache.cc
```diff
@@ -288,16 +308,16 @@ int32_t MemRecursorCache::get(time_t now, const DNSName &qname, const QType& qt,
         *state = *cachedState;
       }
 
-      return ret > 0 ? static_cast<int32_t>(ret-now) : ret;
+      return ret > 0 ? fakeTTD(nullptr, qname, qtype, ret, now, origTTL, refresh) : ret;
```
rgacogne, line 311: Does that means that `ADDR` queries (for the A and AAAA of NS) will not be added to the refresh list? I'm OK with that if that's the case, but does it still makes sense to return a fake TTD? Does that allow other entries to be refreshed?

pdns/recursor_cache.cc
```diff
@@ -633,7 +649,7 @@ uint64_t MemRecursorCache::doDump(int fd)
       for (const auto &sig : i.d_signatures) {
         count++;
         try {
-          fprintf(fp.get(), "%s %" PRId64 " IN RRSIG %s ; %s\n", i.d_qname.toString().c_str(), static_cast<int64_t>(i.d_ttd - now), sig->getZoneRepresentation().c_str(), i.d_netmask.empty() ? "" : i.d_netmask.toString().c_str());
+          fprintf(fp.get(), "%s %" PRIu32 " %" PRId64 " IN RRSIG %s ; %s\n", i.d_qname.toString().c_str(), i.d_orig_ttl, static_cast<int64_t>(i.d_ttd - now), sig->getZoneRepresentation().c_str(), i.d_netmask.empty() ? "" : i.d_netmask.toString().c_str());
```
rgacogne, line 652: I'm not sure we care about the original TTL for RRSIGs since we already dump it for the entry itself.

pdns/recursordist/taskqueue.cc
```diff
@@ -0,0 +1,122 @@
...
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+
+#include "taskqueue.hh"
+
+#include "logger.hh"
+#include "syncres.hh"
+
+extern std::shared_ptr<SyncRes::domainmap_t> g_initialDomainMap;
```
rgacogne, line 28: Is this needed?

pdns/recursordist/taskqueue.cc
```diff
@@ -0,0 +1,122 @@
...
+size_t TaskQueue::size() const
+{
+  return d_queue.size();
+}
+
+void TaskQueue::push(const ResolveTask&& task)
+{
+  // Insertion fails if it's already there, no problem since we're already scheduled
+  // and the deadline would remain the same anyway.
+  auto result = d_queue.insert(std::move(task));
```
rgacogne, line 47: I don't think that's true, the deadline is a fraction of the original TTL that doesn't change plus the time of the insertion, which will.

pdns/recursordist/taskqueue.cc
rgacogne: Most these messages should be at Notice level, I guess?

pdns/recursordist/test-recursorcache_cc.cc
```diff
@@ -455,11 +455,11 @@ BOOST_AUTO_TEST_CASE(test_RecursorCache_ExpungingExpiredEntries)
 
   /* the remaining entry should be power2, but to get it
      we need to go back in the past a bit */
-  BOOST_CHECK_EQUAL(MRC.get(ttd - 1, power2, QType(dr2.d_type), false, &retrieved, who, boost::none, nullptr), 1);
+  BOOST_CHECK_EQUAL(MRC.get(ttd - 1, power2, QType(dr2.d_type), false, &retrieved, who, 0, boost::none, nullptr), 1);
```
rgacogne, line 458: Just a nit but I'd prefer `false` over `0` here and below.

pdns/pdns_recursor.cc
```diff
@@ -2170,6 +2173,8 @@ static void startDoResolve(void *p)
     }
 #endif /* NOD_ENABLED */
 
+    t_taskQueue.runOnce();
```
rgacogne, line 2176: Perhaps we should run that outside of this exception block? I know we are not supposed to bubble up that far but if we ever do we might be confused about seeing the context of the initial query in there.

pdns/pdns_recursor.cc
```diff
@@ -136,6 +137,8 @@ thread_local std::unique_ptr<MT_t> MT; // the big MTasker
 std::unique_ptr<MemRecursorCache> g_recCache;
 std::unique_ptr<NegCache> g_negCache;
 
+static thread_local pdns::TaskQueue t_taskQueue;
```
rgacogne, line 140: Out of curiosity, did you consider making the queue global? I know we should not push the same entry twice into different queues since we mark them as "submitted" (although, I guess we could trigger once from the PC and once from the RC via different threads?) Of course this is not easy to shard if we want to keep global ordering, but we only have relative ordering (per thread) now so I was wondering if you had another reason? To avoid contention, I guess?

rgacogne (approved):
Code looks good, most of my comments are about understanding the design and not issues. Nice work!

omoerbeek:
Rebased

pdns/recpacketcache.hh
```diff
@@ -92,11 +94,13 @@ private:
 #endif
     mutable time_t d_ttd;
     mutable time_t d_creation; // so we can 'age' our packets
+    mutable uint32_t d_orig_ttl;
```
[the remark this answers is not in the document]
omoerbeek: Right. The  record cache does not have it, but here I can use it.

pdns/recursor_cache.cc
omoerbeek: The `nullptr` is there to avoid having to keep track which cache entry was used. You are right that in this case using the fakeTTD isn;t of much use. This also means that the first arg of can be a ref since it will always be there

pdns/recursor_cache.cc
omoerbeek: But doesn't the same hold for the actual ttl that was already there?

pdns/recursordist/taskqueue.cc
omoerbeek: This is an error in the calls I think. The deadline was supposed to be the original ttd, with the idea that it does not make a lot of sense to execute the scheduled task after that. But I see the calls use the "almost expired" point in time.

pdns/recursor_cache.cc
rgacogne: It does indeed, and for the netmask as well. Perhaps it makes sense to stop dumping these? Or we can keep the code as it is, it's not a big issue. I guess the only drawback is making the dump bigger, and perhaps a bit longer to generate.

pdns/pdns_recursor.cc
omoerbeek: I started with a global taskQueue, but then decided to run a task at the end of a resolve always, to avoid being forced to have a thread dedicated to tasks. This meant that to avoid contention, a per-thread queue was preferred. There is indeed a scenario where the same task could end up in two queues. If one of the queue is executed first, the second task will be a no-op. If there are executed simultaneously, I guess the would be handles the same as two client-queries for the same name, i.e. chaining will take place.

omoerbeek:
Comments processed. Lets do one more round of testing to see if I did not botch something in the last minute.

pdns/pdns_recursor.cc
rgacogne: Right, that makes sense to me. Thanks!

pdns/pdns_recursor.cc
```diff
@@ -5403,6 +5414,7 @@ int main(int argc, char **argv)
     ::arg().setSwitch("nothing-below-nxdomain", "When an NXDOMAIN exists in cache for a name with fewer labels than the qname, send NXDOMAIN without doing a lookup (see RFC 8020)")="dnssec";
     ::arg().set("max-generate-steps", "Maximum number of $GENERATE steps when loading a zone from a file")="0";
     ::arg().set("record-cache-shards", "Number of shards in the record cache")="1024";
+    ::arg().set("refresh-on-ttl-perc", "If a record is requsted from the cache and only this % of original TTL remains, refetch") = "0";
```
rgacogne, line 5417:
```suggestion
    ::arg().set("refresh-on-ttl-perc", "If a record is requested from the cache and only this % of original TTL remains, refetch") = "0";
```

pdns/recpacketcache.cc
```diff
@@ -60,10 +63,20 @@ bool RecursorPacketCache::checkResponseMatches(std::pair<packetCache_t::index<Ha
 
     if (now < iter->d_ttd) { // it is right, it is fresh!
       *age = static_cast<uint32_t>(now - iter->d_creation);
+      // we know ttl is > 0
+      uint32_t ttl = static_cast<uint32_t>(iter->d_ttd - now);
+      if (s_refresh_ttlperc > 0) {
+        const uint32_t deadline = iter->getOrigTTL() * s_refresh_ttlperc / 100;
```
Habbie, line 69: could we win a few CPU cycles by checking `d_submitted` before doing the rest of the checks?

pdns/recpacketcache.hh
```diff
@@ -92,12 +94,17 @@ private:
     uint16_t d_type;
     uint16_t d_class;
     mutable vState d_vstate;
+    mutable bool d_submitted;
```
Habbie, line 97:
```suggestion
    mutable bool d_submitted;   // whether this entry has been queued for refetch
```

pdns/recursor_cache.hh
```diff
@@ -95,8 +95,10 @@ private:
     OptTag d_rtag;
     mutable vState d_state;
     mutable time_t d_ttd;
+    uint32_t d_orig_ttl;
     uint16_t d_qtype;
     bool d_auth;
+    mutable bool d_submitted;
```
Habbie, line 101:
```suggestion
    mutable bool d_submitted;     // whether this entry has been queued for refetch
```

pdns/recursordist/taskqueue.cc
```diff
@@ -0,0 +1,120 @@
...
+      sr.beginResolve(task.d_qname, QType(task.d_qtype), QClass::IN, ret);
+    }
+    catch (const std::exception& e) {
+      g_log << Logger::Error << "Exception while running the background task queue: " << e.what() << endl;
+    }
+    catch (const PDNSException& e) {
+      g_log << Logger::Notice << "Exception while running the background task queue: " << e.reason << endl;
+    }
+    catch (const ImmediateServFailException& e) {
+      g_log << Logger::Notice << "Exception while running the background task queue: " << e.reason << endl;
```
Habbie, line 81:
```suggestion
      if(g_logCommonErrors) {
        g_log << Logger::Notice << "Exception while running the background task queue: " << e.reason << endl;
      }
```

because we also gate it behind that for client-initiated resolves

pdns/recursordist/taskqueue.cc
```diff
@@ -0,0 +1,120 @@
...
+    }
+    catch (const PolicyHitException& e) {
+      g_log << Logger::Notice << "Policy hit while running the background task queue" << endl;
```
Habbie, line 84:
```suggestion
      if(g_logCommonErrors) {
        g_log << Logger::Notice << "Policy hit while running the background task queue" << endl;
      }
```

pdns/recursordist/taskqueue.hh
```diff
@@ -0,0 +1,87 @@
...
+
+namespace pdns
+{
+
+struct ResolveTask
+{
+  DNSName d_qname;
+  uint16_t d_qtype;
+  time_t d_deadline;
+  bool d_refreshMode;
```
Habbie, line 47: what does this bool mean?

pdns/syncres.hh
```diff
@@ -610,6 +610,13 @@ public:
     return old;
   }
 
+  uint32_t setRefreshAlmostExpired(bool doit)
```
Habbie, line 613: why is the return type `uint32_t` and not `bool`? (besides, the one caller in taskqueue.cc does not use the return value)

Habbie (approved):
Looks sane! Nice work :+1: 

Couple of nits/comments.

pdns/recpacketcache.cc
omoerbeek: yes

pdns/syncres.hh
omoerbeek: fixed

pdns/recursordist/taskqueue.hh
omoerbeek: Comment added

Habbie (dismissed):

paddg:
It came to mind, that refresh almost expired records unfortunately does not work when using a caching dnsdist in front of a Recursor. That is a pity. Unfortunately, I also don't have a good suggestion on how this could be improved.

omoerbeek:
> It came to mind, that refresh almost expired records unfortunately does not work when using a caching dnsdist in front of a Recursor. That is a pity. Unfortunately, I also don't have a good suggestion on how this could be improved.

Indeed that will not work. Marking almost expired entries for re-fetching on query requires that they *are* queried. But once we have more experience with this, we can research other methods of marking.

pdns/recursor_cache.cc
omoerbeek: I decided to keep the current code, it makes it a tad more easy to spot which lines belong to the same cache entry. We could make the dump more compact by having a entry header en then the individual records without repeating info and then a clear end of entry marker, but that's for some other time.

Habbie (approved):

johnhtodd:
> > It came to mind, that refresh almost expired records unfortunately does not work when using a caching dnsdist in front of a Recursor. That is a pity. Unfortunately, I also don't have a good suggestion on how this could be improved.
> 
> Indeed that will not work. Marking almost expired entries for re-fetching on query requires that they _are_ queried. But once we have more experience with this, we can research other methods of marking.

This seems to be a key "next feature" as I suspect most larger recursive resolver infrastructure will have dnsdist in front of the back-end resolvers, and probably multiple instances of dnsdist pointed to multiple instances of recursors. If there was a way to have dnsdist include "number of queries to this qname/qtype/[ECS] request" as result data included in an RD=0 EDNS query (with filters on IP origin or an EDNS-encoded hash password to prevent privacy leaks) then it might be possible for the backend to determine if the refresh is worth doing. That is a hack, but it might be a useful thing to have for other reasons as well.  How this would skew various counter metrics on dnsdist would also need to be taken into consideration to prevent over-counting.  This is a "non-trivial" signaling problem.

paddg:
> > > It came to mind, that refresh almost expired records unfortunately does not work when using a caching dnsdist in front of a Recursor. That is a pity. Unfortunately, I also don't have a good suggestion on how this could be improved.
> > 
> > 
> > Indeed that will not work. Marking almost expired entries for re-fetching on query requires that they _are_ queried. But once we have more experience with this, we can research other methods of marking.
> 
> This seems to be a key "next feature" as I suspect most larger recursive resolver infrastructure will have dnsdist in front of the back-end resolvers, and probably multiple instances of dnsdist pointed to multiple instances of recursors. If there was a way to have dnsdist include "number of queries to this qname/qtype/[ECS] request" as result data included in an RD=0 EDNS query (with filters on IP origin or an EDNS-encoded hash password to prevent privacy leaks) then it might be possible for the backend to determine if the refresh is worth doing. That is a hack, but it might be a useful thing to have for other reasons as well. How this would skew various counter metrics on dnsdist would also need to be taken into consideration to prevent over-counting. This is a "non-trivial" signaling problem.

The problem is, dnsdist does not know the original TTL or the re-fetch TTL. What about this approach: Limit the maximum TTL within the Recursor to 2^16. This way two TTLs can transmitted to dnsdist. TTL field length is 32 bit.

PR #9700, opened 2020-11-10
Allow setting a fake-ish version number when building release images

pieterlexis:
### Short description
This requires setting DOCKER_FAKE_RELEASE to YES as a build arg (on the docker hub). Combined with automatic builds based on tags (also on the docker hub) will lead to binaries with the correct version number

### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/Dockerfile-auth
+++ b/Dockerfile-auth
@@ -28,11 +28,18 @@ COPY m4 /source/m4
 COPY ext /source/ext
 COPY .git /source/.git
 ADD configure.ac Makefile.am /source/
+COPY builder/helpers/set-configure-ac-version.sh /usr/local/bin
 
 ARG MAKEFLAGS=
 ENV MAKEFLAGS ${MAKEFLAGS:--j2}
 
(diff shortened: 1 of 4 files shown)


Habbie:
Looks good, but what about alpha/beta/RC?

Habbie:
> Looks good, but what about alpha/beta/RC?

In fact, won't this strip '4.4.0-alpha2' to '4.4.0'?

pieterlexis:
won't work, but also wouldn't be built with the current regex for tags on the docker hub :)

Habbie:
Ok, so manual builds for those for now?

pieterlexis:
> Ok, so manual builds for those for now?

I guess, or we can revisit this code by that time

Habbie:
> I guess, or we can revisit this code by that time

4.4.0-beta1 is next week ;) but we'll do that by hand then

PR #9704, opened 2020-11-11
auth: nxdomain performance

mind04:
### Short description
This pull will increase uncached NXDOMAIN performance by roughly 10% in single backend setups. 

getSOAUncached() is used to figure out which backend is responsible for the before and after name. But in single backed setups this is obvious and does not require additional backend queries. 

Commit 1-3 are improvements, commit 4 is clean-up and is not changing functionality.

### Checklist
I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


zeha:
Tests say DS glue is broken now. Generally this PR looks like a good idea to me.

Issue #9208, opened 2020-06-07
Misleading timeout logs for recursor with high load and small number of threads


vin01:
- Program: Recursor
 - Issue type: Improvement

### Short description

Recently I used trace queries to debug some ServFail responses under medium load (300/400 queries per second) and saw following failure logs:

```
timeout resolving after 1515.71msec
Trying IP 127.0.0.1:5300, asking 'example.intranet.domain:AAAA'
...
```

Network timeout is set to default 1500 msec and the pdns auth server running locally on the same node. The log seemed to indicate some issue with the pdns auth server which did not seem to respond within expected 1500 msec. 

However, the issue turned out to be the recursor itself running with a small number of threads, raising thread count with `threads=400` to 400 got rid ot these timeouts and hence `ServFail`s. Before this, recursor was running with default thread count of 2 for my 2 core machine.

I also tried setting `distributor-threads` and `receiver-threads` to higher values which did not help much. Also auth server indicated no issues in logs so I believe this was purely a recursor issue. The [performance guide](https://doc.powerdns.com/recursor/performance.html#performance-guide) also indicates to set `threads` to the number of cores which did not really work well for me with high number of queries.

### Environment

 - Operating system: 

Ubuntu 18.04

 - Software version: 
 - Software source: 

```
pdns-backend-bind/unknown,now 4.3.0-1pdns.bionic amd64 [installed,automatic]
pdns-backend-pgsql/unknown,now 4.3.0-1pdns.bionic amd64 [installed]
pdns-recursor/unknown,now 4.3.1-1pdns.bionic amd64 [installed]
pdns-server/unknown,now 4.3.0-1pdns.bionic amd64 [installed]
```

### Steps to reproduce

High number of concurrent `AAAA` queries (400/s) in my case.

### Expected behaviour

A better log message perhaps indicating internal wait time exhaustion would be more helpful here rather than generic timeout log.

### Actual behaviour

A log indicating timeout which implies a timeout in getting a response from the auth server.
Alternatively, perhaps it can be mentioned in documentation explicitly to avoid confusion. 

### Other information

[...]

PR #9707, opened 2020-11-11
rec: Do not add request to a wait chain that's already processed or being processed

omoerbeek:
The following scenario can occur. Multiple concurrent clients doing the same query A
are needed to trigger it:

1. Incoming request A, which has a need for request X
2. Add request X to chain because we already have an identical outstanding request
3. We receive the reply for X
4. We process the chain
5. In the meantime a new request for X that's identical is added to the chain
6. The added id in step 5 is not being processed anymore -> timeout

This can happen if request X has TTL 0, otherwise the record cache would have a hit.

Should fix timeouts seen for auth serving records with TTL 0. I suspect #9208 is a case of that as increasing the number of threads avoids the issue. Another workaound it to set `override-minimum-ttl`.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [X] documented the code

### DIFF
--- a/pdns/pdns_recursor.cc
+++ b/pdns/pdns_recursor.cc
@@ -653,7 +653,7 @@ LWResult::Result asendto(const char *data, size_t len, int flags,
   pair<MT_t::waiters_t::iterator, MT_t::waiters_t::iterator> chain=MT->d_waiters.equal_range(pident, PacketIDBirthdayCompare());
 
   for(; chain.first != chain.second; chain.first++) {
-    if(chain.first->key.fd > -1) { // don't chain onto existing chained waiter!
+    if(chain.first->key.fd > -1 && !chain.first->key.closed) { // don't chain onto existing chained waiter or a chain already processed
       /*
       cerr<<"Orig: "<<pident.domain<<", "<<pident.remote.toString()<<", id="<<id<<endl;
       cerr<<"Had hit: "<< chain.first->key.domain<<", "<<chain.first->key.remote.toString()<<", id="<<chain.first->key.id
@@ -3814,6 +3814,9 @@ static void handleTCPClientWritable(int fd, FDMultiplexer::funcparam_t& var)
 // resend event to everybody chained onto it
 static void doResends(MT_t::waiters_t::iterator& iter, PacketID resend, const string& content)
 {
+  // We close the chain for new entries, since they won't be processed anyway
+  iter->key.closed = true;
+
   if(iter->key.chain.empty())
     return;
   //  cerr<<"doResends called!\n";
--- a/pdns/mtasker.hh
+++ b/pdns/mtasker.hh
@@ -81,7 +81,7 @@ public:
     EventKey key;
     std::shared_ptr<pdns_ucontext_t> context;
     struct timeval ttd;
-    int tid;    
+    int tid;
   };
 
   typedef multi_index_container<
--- a/pdns/syncres.hh
+++ b/pdns/syncres.hh
@@ -908,7 +908,7 @@ LWResult::Result arecvtcp(string& data, size_t len, Socket* sock, bool incomplet
 
 struct PacketID
 {
-  PacketID() : id(0), type(0), sock(0), inNeeded(0), inIncompleteOkay(false), outPos(0), nearMisses(0), fd(-1)
+  PacketID() : id(0), type(0), sock(0), inNeeded(0), inIncompleteOkay(false), outPos(0), nearMisses(0), fd(-1), closed(false)
   {
     remote.reset();
   }
@@ -930,6 +930,7 @@ struct PacketID
   mutable chain_t chain;
   mutable uint32_t nearMisses; // number of near misses - host correct, id wrong
   int fd;
+  mutable bool closed; // Processing already started, don't accept new chained ids
 
   bool operator<(const PacketID& b) const
   {


pdns/pdns_recursor.cc
```diff
@@ -653,7 +653,7 @@ LWResult::Result asendto(const char *data, size_t len, int flags,
   pair<MT_t::waiters_t::iterator, MT_t::waiters_t::iterator> chain=MT->d_waiters.equal_range(pident, PacketIDBirthdayCompare());
 
   for(; chain.first != chain.second; chain.first++) {
-    if(chain.first->key.fd > -1) { // don't chain onto existing chained waiter!
+    if(chain.first->key.fd > -1 && !chain.first->key.closed) { // don't chain onto existing chained waiter or a chain already processed
```
rgacogne, line 656: So if I understand that right, we yield in `doResends()` by calling `MTasker::sendEvent()`, and a different mthread then tries to add the same query again. This change makes it so that it does not try to reuse the old one, which is good, but I guess that mthread will then call `MTasker::waitEvent()` via `arecvfrom()`. How does that not fail (`d_waiters.count(key)` != 0) if the old entry is still there? And if it does not fail, is there any risk that our iterator gets invalidated?

omoerbeek: I'm not sure we yield. What I've seen is add to chain, process, add to chainin the same mthread. I suspect the specific record is needed twice in the same resolution process and bcause of the TTL 0 the cached one is not used.

rgacogne: I'm not sure I understand how that would happen with only one mthread. Wouldn't the entry be removed in `MTasker::sendEvent()`, called from `handleUDPServerResponse()`, before the correct mthread is woken up from the kernel one?

omoerbeek: If I run with the diff below on top of the PR (it disabled the !close condition and adds logging). I see 👍 
```
...
36 asendto() add to chain
36 doResends called!
35 asendto() add to chain
35 doResends called!
36 asendto() add to chain
36 doResends called!
38 asendto() add to chain
38 doResends called!
38 asendto() add to chain
...
```
So there does seem to be yielding going on...
This is with two clients running dig in a loop:
```
while true; do dig +time=5 +tries=1 -p 5301 euaz.tr.teams.microsoft.com ; sleep .05; done | grep "Query time"
```
```
diff --git a/pdns/pdns_recursor.cc b/pdns/pdns_recursor.cc
index 49cf2c463..d95be13d7 100644
--- a/pdns/pdns_recursor.cc
+++ b/pdns/pdns_recursor.cc
@@ -653,7 +653,8 @@ LWResult::Result asendto(const char *data, size_t len, int flags,
   pair<MT_t::waiters_t::iterator, MT_t::waiters_t::iterator> chain=MT->d_waiters.equal_range(pident, PacketIDBirthdayCompare());
 
   for(; chain.first != chain.second; chain.first++) {
-    if(chain.first->key.fd > -1 && !chain.first->key.closed) { // don't chain onto existing chained waiter or a chain already processed
+    if(chain.first->key.fd > -1) { // don't chain onto existing chained waiter or a chain already processed
+      cerr << MT->getTid() << " asendto() add to chain" << endl;
       /*
       cerr<<"Orig: "<<pident.domain<<", "<<pident.remote.toString()<<", id="<<id<<endl;
       cerr<<"Had hit: "<< chain.first->key.domain<<", "<<chain.first->key.remote.toString()<<", id="<<chain.first->key.id
@@ -3819,7 +3820,7 @@ static void doResends(MT_t::waiters_t::iterator& iter, PacketID resend, const st
 
   if(iter->key.chain.empty())
     return;
-  //  cerr<<"doResends called!\n";
+  cerr<< MT->getTid() << " doResends called!\n";
   for(PacketID::chain_t::iterator i=iter->key.chain.begin(); i != iter->key.chain.end() ; ++i) {
     resend.fd=-1;
     resend.id=*i;
```

omoerbeek:
A trace with a little bit more info printed:
```
31 asendto() add to chain b-tr-teasc-ukwe-07.ukwest.cloudapp.azure.com 40.90.4.5 34647

31 doResends called b-tr-teasc-ukwe-07.ukwest.cloudapp.azure.com 40.90.4.5
        Resending 89 bytes for fd=-1 and id=34647
33 asendto() add to chain euaz.tr.teams.trafficmanager.net 13.107.252.10 46311

33 doResends called euaz.tr.teams.trafficmanager.net 13.107.252.10
        Resending 124 bytes for fd=-1 and id=46311
32 asendto() add to chain b-tr-teasc-euno-04.northeurope.cloudapp.azure.com 13.107.160.201 39754

32 doResends called b-tr-teasc-euno-04.northeurope.cloudapp.azure.com 13.107.160.201
        Resending 94 bytes for fd=-1 and id=39754
33 asendto() add to chain euaz.tr.teams.trafficmanager.net 13.107.252.10 16682

33 doResends called euaz.tr.teams.trafficmanager.net 13.107.252.10
        Resending 123 bytes for fd=-1 and id=16682
```
Until here the pattern: asendto followed by doResends
```
35 asendto() add to chain euaz.tr.teams.trafficmanager.net 13.107.252.10 36039

35 doResends called euaz.tr.teams.trafficmanager.net 13.107.252.10
        Resending 123 bytes for fd=-1 and id=36039
35 asendto() add to chain euaz.tr.teams.trafficmanager.net 13.107.252.10 17934
```
This a case of an asendto *not* followed by doResends. It also does not come later. The timeing corresponds to a dig with 1.5s delay.
```
53 asendto() add to chain euaz.tr.teams.trafficmanager.net 13.107.252.10 17579

53 doResends called euaz.tr.teams.trafficmanager.net 13.107.252.10
        Resending 119 bytes for fd=-1 and id=17579
52 asendto() add to chain a-tr-teasc-ukwe-13.ukwest.cloudapp.azure.com 64.4.48.5 58860

52 doResends called a-tr-teasc-ukwe-13.ukwest.cloudapp.azure.com 64.4.48.5
        Resending 89 bytes for fd=-1 and id=58860
53 asendto() add to chain euaz.tr.teams.trafficmanager.net 13.107.252.10 15611

53 doResends called euaz.tr.teams.trafficmanager.net 13.107.252.10
        Resending 123 bytes for fd=-1 and id=15611
52 asendto() add to chain a-tr-teasc-euwe-13.westeurope.cloudapp.azure.com 40.90.4.201 59407

52 doResends called a-tr-teasc-euwe-13.westeurope.cloudapp.azure.com 40.90.4.201
        Resending 93 bytes for fd=-1 and id=59407
```
The rest is also regular, though intertwined.

omoerbeek:
All logs are created with`threads=1`

rgacogne:
Alright, I finally realized what I was missing, and I think your fix is right. Sorry it took me so long!
What I did not understand was that while we do look for similar queries matching on (remote, qtype, qname) only, the actual comparison operator also takes the ID into account. So we can have two concurrent events for the same (remote, qtype, qname), meaning that with your fix we can not only correctly skip the ones that are already being processed, but also insert a new one without overwriting the existing one.

rgacogne (approved):

omoerbeek:
> Alright, I finally realized what I was missing, and I think your fix is right. Sorry it took me so long!

Well, my explanation was missing some points as well :-)

PR #9709, opened 2020-11-12
Add a simple fuzzing target for YaHTTP

rgacogne:
### Short description

Since we use YaHTTP to parse incoming web requests (and responses in the remote HTTP backend), better fuzz it :)

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


Habbie (approved):
not an expert on the fuzzing infra, but I see nothing obviously wrong here

rgacogne:
The CIFuzz build fails because the YaHTTP lib has not been built at that point. I need to either find a way to make that a dependency, or to modify the OSSFuzz build process to build YaHTTP beforehand.

rgacogne:
In the end I modified the OSSFuzz build process to build YaHTTP beforehand.

rgacogne:
Reminder to self: please submit the change to the oss-fuzz repository so that the new corpus for the YaHTTP fuzzing target is actually used.

PR #9712, opened 2020-11-12
Use Python 3 in build system

kpfleming:
### Short description
The various tasks in the build system which use Python
will now use Python 3 (3.6 or later) and the 'venv' module,
instead of the 'virtualenv' tool which was primarily used
with Python 2.7.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code


.travis.yml
```diff
@@ -1,5 +1,5 @@
 sudo: required
-dist: trusty
+dist: bionic
```
pieterlexis, line 2: here be several angry dragons

Habbie: @rgacogne points out that a new travis would not add much compared to Debian Buster on CircleCI, and that as long as we have Travis, we might as well keep it on the old stuff. This makes sense to me.

Habbie: but, we should still switch that to py3

kpfleming: Alright, I'll work on reverting this to Trusty.

kpfleming: The latest version of  Python available for Trusty is 3.5, which is EoL, and package maintainers are already adopting 3.6-and-later features in their packages. If we can make this work with Python 3.5 it will have a limited lifetime, as at some point a package we need for building docs will no longer be compatible.

Habbie: I understand -Travis- offers py3.6 in the Trusty image. If that is easy to use, then please try that. If not, we'll spend some more time thinking about what we want with Travis.

Personally I'm tempted to solve the "let's make sure we periodically test on old systems" problem in Github Actions separately.

kpfleming: Travis build reverted to Ubuntu Trusty, but with Python 3.6.

kpfleming:
All of the build processes are working, but there are various regression tests which are failing when run with Python 3.6. I'm guessin there are either syntax changes required in the regression test code, or there are package dependencies which are not Python 3 compatible.

Habbie (commented):
just one comment i spotted while going over this

Habbie:
> All of the build processes are working, but there are various regression tests which are failing when run with Python 3.6. I'm guessin there are either syntax changes required in the regression test code, or there are package dependencies which are not Python 3 compatible.

I'll have a go at this.

pdns/dnsdistdist/m4/ax_python_module.m4
```diff
@@ -0,0 +1,56 @@
+# ===========================================================================
```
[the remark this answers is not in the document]
kpfleming: Ahh, I wasn't sure if symlinks were safe here. I can symlink `pdns_check_python_venv.m4` too, since it's identical in all three locations (the module-specific tests were moved into each module's configure script).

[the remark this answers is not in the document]
Habbie: yes! symlinking the other m4s was implied :)

Habbie:
> I'll have a go at this.

I've made some progress, I'll provide some work soon.

Habbie:
dnsdist in Travis is failing with `configure.ac:127: error: AM_COND_IF: no such condition "HAVE_VENV"`, I don't immediately see why.

kpfleming:
That's already been fixed, there were broken symlinks...

Habbie:
> That's already been fixed, there were broken symlinks...

ah good!

kpfleming:
I'm currently working on testing package builds; debian-buster works, running ubuntu-xenial right now, centos-7 appears to have a problem which I'll need to investigate.

Habbie:
https://github.com/Habbie/pdns/commit/92c71f0a6872543500d9fa3c0d977bed79f16937 turns recursor green on circleci by fixing a few tests and skipping a lot. Will need some more chipping away at.

Habbie:
travis auth, I'm unsure why this happens, it works for me locally:
```
=====================================================================

FAIL: test_post_broken_key_name (test_TSIG.AuthTSIG)

  File "/home/travis/build/PowerDNS/pdns/regression-tests.api/test_TSIG.py", line 179, in test_post_broken_key_name

    self.assertEqual(r.status_code, 422)

AssertionError: 201 != 422

-------------------- >> begin captured logging << --------------------

urllib3.connectionpool: DEBUG: Starting new HTTP connection (1): 127.0.0.1:5556

urllib3.connectionpool: DEBUG: http://127.0.0.1:5556 "POST /api/v1/servers/localhost/tsigkeys HTTP/1.1" 201 139

--------------------- >> end captured logging << ---------------------

FAIL: test_put_broken_key (test_TSIG.AuthTSIG)

  File "/home/travis/build/PowerDNS/pdns/regression-tests.api/test_TSIG.py", line 146, in test_put_broken_key

    self.assertEqual(r.status_code, 422)

AssertionError: AssertionError: 200 != 422

-------------------- >> begin captured stdout << ---------------------

sending {'name': 'test-13115543342849-key', 'algorithm': 'hmac-md5'}

reply {'algorithm': 'hmac-md5', 'id': 'test-13115543342849-key.', 'key': 'ukOl6UkWZxERF0NduMU1q8s460V7vFzm5fJc5N+B5Wo=', 'name': 'test-13115543342849-key', 'type': 'TSIGKey'}

--------------------- >> end captured stdout << ----------------------

-------------------- >> begin captured logging << --------------------

urllib3.connectionpool: DEBUG: Starting new HTTP connection (1): 127.0.0.1:5556

urllib3.connectionpool: DEBUG: http://127.0.0.1:5556 "POST /api/v1/servers/localhost/tsigkeys HTTP/1.1" 201 168

urllib3.connectionpool: DEBUG: Resetting dropped connection: 127.0.0.1

urllib3.connectionpool: DEBUG: http://127.0.0.1:5556 "PUT /api/v1/servers/localhost/tsigkeys/test-13115543342849-key. HTTP/1.1" 200 139

--------------------- >> end captured logging << ---------------------
```

kpfleming:
Found the problem with the centos-7 package builds (don't know why it doesn't happen on debian-buster or ubuntu-xenial): api-swagger.yaml contains Unicode left/right single/double quotation marks, which are not permitted in YAML, so the YAML->JSON conversion fails.

Habbie:
Huh? Why would those not be permitted?

Habbie:
https://yaml.org/spec/current.html#id2513364

kpfleming:
In spite of its URL, that appears to be a working draft; the final YAML 1.1 spec has a 'Character Set' section:

https://yaml.org/spec/1.1/#id868524

Habbie:
Ah! But that also appears to allow it.

kpfleming:
Indeed it does... could be a bug in PyYAML i suspect. I'll look into that, but the quickest solution for now is to just not use them :-)

kpfleming:
OK, it appears to be a bug in PyYAML >= 5.1. I'll restrict the installation to versions lower than that.

kpfleming:
Well that was fun. The actual cause was that the default encoding for Python stdin/stdout is platform-dependent, and on CentOS 7 (at least) it is ASCII, not UTF-8. The solution was to move the transformation code into a small Python script where I can specify the input and output encoding directly.

Habbie:
> The actual cause was that the default encoding for Python stdin/stdout is platform-dependent, and on CentOS 7 (at least) it is ASCII, not UTF-8.

Ah! When stdout is a terminal, Python looks at the locale I think, but that's not what we're doing here.

I almost have the Recursor tests ported to py3.

Habbie:
https://github.com/Habbie/pdns/tree/use-python3-next now has passing recursor tests (on my system) and I rebased it on your last push. Going to wait and see what all the CI thinks of it :)

kpfleming:
I have a complete batch of centos/debian/ubuntu package builds running now as well, since they are not dependent on the regression tests.

Habbie:
(by the way, rebasing is easier for me if you add commits instead of squashing+force pushing; we can squash when all of the whole work is done)

Habbie:
> https://github.com/Habbie/pdns/tree/use-python3-next now has passing recursor tests (on my system) and I rebased it on your last push. Going to wait and see what all the CI thinks of it :)

Appears to have fully passed. I wonder where the Travis auth failure went!

Habbie:
Can you, at your convenience, pick the 3 commits from my use-python3-next branch? I'd prefer those remain unsquashed vs. your work :)

```
commit 38e926fafc1a326b4fc30d6606e2759bf3db8e0a (HEAD -> use-python3-next, habbie/use-python3-next)
Author: Peter van Dijk <peter.van.dijk@powerdns.com>
Date:   Fri Nov 13 19:43:35 2020 +0100

    add venv to Dockerfile-auth for JSON API file

commit a0c538514108abeafc3962f1ea1fc18c75e0acfa
Author: Peter van Dijk <peter.van.dijk@powerdns.com>
Date:   Fri Nov 13 17:03:17 2020 +0100

    recursor DNSSEC regression tests: skip debug cachedump

commit 0c091ab8afb477e21b62eb68c415214df2007b35
Author: Peter van Dijk <peter.van.dijk@powerdns.com>
Date:   Fri Nov 13 12:40:23 2020 +0100

    recursor DNSSEC regression tests: migrate to Python 3

```

kpfleming:
Will do! I'm waiting for the local package builds to complete, so far everything is working but as you know it takes a long time to build and test all the packages for 8 platforms...

kpfleming:
OK, latest push has:

* Habbie's fixes for regression tests
* Addition of IUS Community Repository for Python 3.6 on CentOS 6

Has been package-build tested on el6, el7, el8, stretch, buster, xenial, bionic, and focal.

kpfleming:
Blargh... the TSIG auth test failures on Travis-CI are back.

Habbie:
> Blargh... the TSIG auth test failures on Travis-CI are back.

Ok, so they're intermittent. I'll have another look at that.

Habbie:
in travis.sh, augmenting

```
  run "DNSDISTBIN=$HOME/dnsdist/bin/dnsdist ./runtests -v --ignore-files='(?:^\.|^_,|^setup\.py$|^test_TLSSessionResumption\.py$)'"
```

to also skip `test_Dnstap.py` and perhaps `test_Protobuf.py` should make dnsdist pass on Travis. The very old protobuf compiler on Travis is not compatible with the Python 3 choice we are making here. Meanwhile those features are tested on CircleCI, so skipping them here is fine.

Habbie:
Ah, and in fact the same applies to Recursor - that's also failing because of the old protobuf compiler and the new Python. In build-scripts/test-recursor, `./runtests -e 'SNMP' || EXIT=1` can be changed to `'SNMP|DNSTap|Protobuf'` or similar.

(I might have time to do test pushes for these changes later, but for now, you have my notes).

Habbie:
Also, can you add:
```diff
--- a/regression-tests.recursor-dnssec/printlogs.py
+++ b/regression-tests.recursor-dnssec/printlogs.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python2
+#!/usr/bin/env python3
```

:)

kpfleming:
`test_TSIG.py` now passes; there are two strings with embedded Unicode characters, which Python 2 ignored and Python 3 parses. Adding a backslash prefix turns them back into normal strings which will be passed through unmodified.

Habbie:
auth-py also fails on my system. Fixing.

Habbie:
Please:
1. copy clientsubnetoption.py from regression-tests.recursor-dnssec to regression-tests.auth-py
2. apply:
```diff
diff --git a/regression-tests.auth-py/requirements.txt b/regression-tests.auth-py/requirements.txt
index 27efa3234..6f6ac0bab 100644
--- a/regression-tests.auth-py/requirements.txt
+++ b/regression-tests.auth-py/requirements.txt
@@ -2,4 +2,4 @@ dnspython>=1.11
 nose>=1.3.7
 Twisted>0.15.0
 requests>=2.18.4
-git+https://github.com/PowerDNS/xfrserver.git@0.2
+git+https://github.com/PowerDNS/xfrserver.git@0.3
diff --git a/regression-tests.auth-py/test_IXFR.py b/regression-tests.auth-py/test_IXFR.py
index 0970d9232..fa8c85082 100644
--- a/regression-tests.auth-py/test_IXFR.py
+++ b/regression-tests.auth-py/test_IXFR.py
@@ -239,6 +239,6 @@ newrecord.example.        8484    A       192.0.2.42
         ret = subprocess.check_output([os.environ['PDNSUTIL'],
                            '--config-dir=configs/auth',
                            'list-zone', 'example'], stderr=subprocess.STDOUT)
-        rets = ret.split('\n')
+        rets = ret.split(b'\n')
 
-        self.assertEqual(1, sum('SOA' in l for l in ret.split('\n')))
\ No newline at end of file
+        self.assertEqual(1, sum(b'SOA' in l for l in ret.split(b'\n')))
\ No newline at end of file
```

Habbie:
Or cherry-pick from https://github.com/Habbie/pdns/tree/use-python3-next - where I pushed this commit and it is now going through CI.

rgacogne:
> copy clientsubnetoption.py from regression-tests.recursor-dnssec to regression-tests.auth-py

It might be nice to use symbolic links for the EDNS option files, since we are using them in various regression test directories.

Habbie:
> It might be nice to use symbolic links for the EDNS option files, since we are using them in various regression test directories.

I was pondering the same, except we had different versions at some point. I believe we're back to 3 identical versions now though :)

kpfleming:
Travis-CI is now green :-)

Habbie:
Otto, Rémi, can you take a look at the changes made to your respective test suites please?

Habbie:
(I see now we did not actually touch dnsdist tests!)

.github/workflows/codeql-analysis.yml
```diff
@@ -64,9 +63,12 @@ jobs:
     #    and modify them (or add more) to build your code if your project
     #    uses a compiled language
 
+    - name: Update repository metadata
+      run: |
+        sudo apt-get update
     - name: Install dependencies
       run: |
-        sudo apt-get update && sudo apt-get --no-install-recommends install -qq -y \
```
Habbie, line 69: why is the update gone?

builder-support/dockerfiles/Dockerfile.rpmbuild
```diff
@@ -1,8 +1,6 @@
 FROM dist-base as package-builder
 RUN touch /var/lib/rpm/* && \
-    yum install -y rpm-build rpmdevtools /usr/bin/python3 && \
-    yum groupinstall -y "Development Tools" && \
-    rpmdev-setuptree
+    yum install -y rpm-build rpmdevtools python3 "@Development Tools"
```
Habbie, line 3: @pieterlexis any opinion on this removal of `rpmdev-setuptree` ?

builder-support/dockerfiles/Dockerfile.target.centos-6
```diff
@@ -10,6 +10,10 @@ RUN touch /var/lib/rpm/* && yum clean all
 RUN touch /var/lib/rpm/* && yum install -y --verbose epel-release centos-release-scl-rh && \
     yum install -y --nogpgcheck devtoolset-7-gcc-c++
 
+# Enable IUS Community Repository (for Python 3.6)
```
Habbie, line 13: the 3.4 in EPEL is too old?

Habbie (commented):
Excellent work, thank you very much! A few questions remain.

.github/workflows/codeql-analysis.yml
kpfleming: It wasn't removed, just moved to a separate step above this one, which is not visible in this context for some reason.

rgacogne (approved):

.github/workflows/codeql-analysis.yml
Habbie: ah, totally missed that, sorry!

builder-support/dockerfiles/Dockerfile.rpmbuild
kpfleming: It was called twice; once here, and once after setting WORKDIR.

builder-support/dockerfiles/Dockerfile.target.centos-6
kpfleming: To be candid I didn't even try to use it, as it's far beyond EOL and adding the IUS repository wasn't difficult. You've already announced that Auth 4.4.x won't be supported on CentOS 6 anyway, and I assume that will be true for Recursor and DNSdist in their next releases too, so this target file will probably be removed soon :-)

Habbie: Works for me, and this should not have made IUS a runtime dep. Ok!

builder-support/dockerfiles/Dockerfile.rpmbuild
Habbie: :+1:

Habbie:
Thank you so much, Kevin!

PR #9719, opened 2020-11-13
rec: Backport to 4.4.x: Do not add request to a wait chain that's already processed or being processed.

omoerbeek:
The following scenario can occur. Multiple concurrent clients doing the same query A
are needed to trigger it:

1. Incoming request A, which has a need for request X
2. Add request X to chain because we already have an identical outstanding request
3. We receive the reply for X
4. We process the chain
5. In the meantime a new request for X that's identical is added to the chain
6. The added id in step 5 is not being processed anymore -> timeout

This can happen if request X has TTL 0, otherwise the record cache would have a hit.

(cherry picked from commit c647a254a0f863aabeaea9d33f673afa26c60457)

Backport of #9707 

### Checklist

I have:
- [X] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/mtasker.hh
+++ b/pdns/mtasker.hh
@@ -81,7 +81,7 @@ public:
     EventKey key;
     std::shared_ptr<pdns_ucontext_t> context;
     struct timeval ttd;
-    int tid;    
+    int tid;
   };
 
   typedef multi_index_container<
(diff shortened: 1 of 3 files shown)


paddg:
I can confirm, it works, thanks!!

rgacogne (approved):

PR #9721, opened 2020-11-13
dnsdist: Improve const-correctness of Lua bindings

georgeto:
### Short description
Since LuaWrapper enforces const-correctness, a const (shared) pointer
to an object can only read members and call const methods.
Therefore we have to take care to use the const qualifier during
registration of constant lambda Lua function bindings.

An example where the absence of the const qualifier was problematic is
LuaRule, which invokes its user provided match function with a constant
DNSQuestion object. Because the functions of DNSQuestion lacked
the const qualifier (e.g. getTag()), it was not possible to call them
from within the LuaRule match function.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


rgacogne (approved):
That PR looks great, thanks a lot!

PR #9735, opened 2020-11-16
auth: pdnsutil check-zone: DNAME fixes

Habbie:
### Short description
* don't warn about occlusion of records with the same name as the DNAME record
* do warn about occlusion when the DNAME is at the apex

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/pdnsutil.cc
+++ b/pdns/pdnsutil.cc
@@ -468,6 +468,7 @@ static int checkZone(DNSSECKeeper &dk, UeberBackend &B, const DNSName& zone, con
     }
 
     if(rr.qname==zone) {
+      // apex checks
       if (rr.qtype.getCode() == QType::NS) {
         hasNsAtApex=true;
       } else if (rr.qtype.getCode() == QType::DS) {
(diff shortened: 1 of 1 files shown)


pdns/pdnsutil.cc
```diff
@@ -641,6 +644,11 @@ static int checkZone(DNSSECKeeper &dk, UeberBackend &B, const DNSName& zone, con
         continue;
       }
       if( rr.qname.isPartOf( qname.first ) ) {
+        if( qname.second == QType::DNAME && rr.qname == qname.first ) {
```
zeha, line 647: can't say these if/if/if/if blocks are very readable anymore. took me quite some time to understand why this is a new case over the first if. i don't understand however why this is inside the rr.qname.isPartOf bit.

Habbie: I see now that 3 lines up also mentions DNAME. I can probably do better.

Habbie: combining the ifs would not have made things more readable. Instead I added a bunch of comments.

zeha:
The new comments are much appreciated, thanks!

PR #9744, opened 2020-11-17
rec: Switch to TCP in case of spoofing (near-miss) attempts

rgacogne:
### Short description

Instead of treating this like an unrecoverable network error and trying the next server, let's switch to TCP instead. This might
prevent a DoS by making us try every single servers and failing, and will make the spoofing attempt a bit much harder.

Could use a regression test.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/pdns_recursor.cc
+++ b/pdns/pdns_recursor.cc
@@ -693,7 +693,7 @@ LWResult::Result asendto(const char *data, size_t len, int flags,
 LWResult::Result arecvfrom(std::string& packet, int flags, const ComboAddress& fromaddr, size_t *d_len,
                            uint16_t id, const DNSName& domain, uint16_t qtype, int fd, const struct timeval* now)
 {
-  static const boost::optional<unsigned int> nearMissLimit = ::arg().asNum("spoof-nearmiss-max");
+  static const unsigned int nearMissLimit = ::arg().asNum("spoof-nearmiss-max");
 
   PacketID pident;
   pident.fd=fd;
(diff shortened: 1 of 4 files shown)


Habbie (approved):
Tidy!

omoerbeek (changes requested):

pdns/pdns_recursor.cc
```diff
@@ -693,9 +693,7 @@ LWResult::Result asendto(const char *data, size_t len, int flags,
 LWResult::Result arecvfrom(std::string& packet, int flags, const ComboAddress& fromaddr, size_t *d_len,
                            uint16_t id, const DNSName& domain, uint16_t qtype, int fd, const struct timeval* now)
 {
-  static boost::optional<unsigned int> nearMissLimit;
-  if(!nearMissLimit)
-    nearMissLimit=::arg().asNum("spoof-nearmiss-max");
+  static const boost::optional<unsigned int> nearMissLimit = ::arg().asNum("spoof-nearmiss-max");
```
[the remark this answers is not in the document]
rgacogne: Nice catch, I don't know how I missed that!

pdns/pdns_recursor.cc
```diff
@@ -715,10 +713,10 @@ LWResult::Result arecvfrom(std::string& packet, int flags, const ComboAddress& f
 
     *d_len=packet.size();
 
-    if (*nearMissLimit && pident.nearMisses > *nearMissLimit) {
-      g_log<<Logger::Error<<"Too many ("<<pident.nearMisses<<" > "<<*nearMissLimit<<") bogus answers for '"<<domain<<"' from "<<fromaddr.toString()<<", assuming spoof attempt."<<endl;
+    if (nearMissLimit > 0 && pident.nearMisses > nearMissLimit) {
+      g_log<<Logger::Error<<"Too many ("<<pident.nearMisses<<" > "<<nearMissLimit<<") bogus answers for '"<<domain<<"' from "<<fromaddr.toString()<<", assuming spoof attempt."<<endl;
```
setharnold, line 717: Is this `bogus` in the sense of dnssec or is this in the sense of "not even close to right"? If the latter, it might be worth changing the wording while we're here. ("invalid", "improper", "bad <field name>", etc.)

omoerbeek: Yes the word bogus is misleading. It's more "close but no cigar" ;-)

rgacogne: I'll update it to "Too many ($attempts > $limit) answers with a wrong message ID for $domain from $source, assuming spoof attempt.".

setharnold: yay thanks!

omoerbeek (approved):

PR #9748, opened 2020-11-18
Avoid accidental removal of generated files

kpfleming:
### Short description
When generated files (man pages, api-swagger files, etc.) are in the tree because they were part of a source distribution
but were not locally generated, 'make clean' should not remove them. Doing so results in a broken tree, as the Makefiles
expect the files to be present, and running 'configure' again will introduce the additional dependencies required to
generate them.

### Checklist
I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/docs/Makefile.am
+++ b/docs/Makefile.am
@@ -87,7 +87,11 @@ mans/.complete: .venv
 	html-docs.tar.bz2
 
 clean-local:
+if !HAVE_MANPAGES
+# don't clean these files if they were present
+# at 'configure' time (e.g. from a source dist)
 	rm -f latex/PowerDNS-Authoritative.pdf $(MANPAGES_DIST) html-docs.tar.bz2
+endif
 	rm -rf mans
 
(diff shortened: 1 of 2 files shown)


kpfleming:
Well, that's unexpected. These changes should not have affected the fuzzing tests at all.

Habbie:
> Well, that's unexpected. These changes should not have affected the fuzzing tests at all.

Yes, looks unrelated.

kpfleming:
And the Travis-CI jobs failed because the Ubuntu package repositories were broken... this is not a good sign.

Habbie:
> And the Travis-CI jobs failed because the Ubuntu package repositories were broken... this is not a good sign.

this PR is cursed, sorry.

Habbie:
i poked some restart buttons

Habbie:
fuzz passed now; travis is still red from the previous run while the jobs are pending

pieterlexis (approved):
nice.

PR #9749, opened 2020-11-19
Fix the DNSName move assignment operator

rgacogne:
### Short description

A misplaced 'const' prevented it from being called, making every move of a DNSName into a full copy.

Introduced in d720eb8add5ebda11867e8b404125e0b68ed2911.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsname.hh
+++ b/pdns/dnsname.hh
@@ -71,7 +71,7 @@ public:
     }
     return *this;
   }
-  DNSName& operator=(const DNSName&& rhs)
+  DNSName& operator=(DNSName&& rhs)
   {
     if (this != &rhs) {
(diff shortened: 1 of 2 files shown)


omoerbeek (approved):
Ugh, my mistake :-( If only some static analyzer would warn about this....

rgacogne:
> If only some static analyzer would warn about this....

I was really angry at the compiler for not warning about this, although I do understand why..

Habbie (approved):

rgacogne:
I pushed a tentative fix for the auth failures, but although I'm quite sure the altered code was wrong it might not have been the only issue.

rgacogne:
Looks good now!

PR #9779, opened 2020-11-26
gh actions: build centos6 packages

Habbie:
### Short description
This includes running `make check`.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/.github/workflows/builder.yml
+++ b/.github/workflows/builder.yml
@@ -1,0 +1,24 @@
+name: 'Test package building for specific distributions'
+
+on:
+  push:
+  pull_request:
+  schedule:
+    - cron: '0 22 * * 2'
+
+jobs:
+  build:
+    # a centos-6 Docker container
+    name: centos-6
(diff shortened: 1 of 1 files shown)


.github/workflows/builder.yml
```diff
@@ -0,0 +1,21 @@
...
+on:
+  push:
+  pull_request:
+  schedule:
+    - cron: '0 22 * * 2'
+
+jobs:
+  build:
+    name: centos-6
+    runs-on: ubuntu-20.04
```
rgacogne, line 12: Just to be sure I understand, this will use pdns-builder to run a CentOS 6 docker container from inside an Ubuntu 20.04, right?

Habbie: Yes! and `make check` also runs inside the CentOS 6 container.

Habbie: I'll add some more comments before merging.

PR #9782, opened 2020-11-27
dnsdist: Use an eBPF filter for Dynamic blocks when available

### LINKED ISSUE #6763: dnsdist: Add transparent support for eBPF dynamic bloc in DynBlockRulesGroup
- Program: dnsdist
 - Issue type: Feature request 

### Short description

It would be nice to be able to use `eBPF` filtering when available with a `DynBlockRulesGroup`. 

### Usecase

Take full advantage of the new `DynBlockRulesGroup` interface in combination with `eBPF` filtering.

### Description

Since the number of entries in an `eBPF` table is limited the implementation would need to fallback to "regular" rules if the insertion of a new entry fails.
Making sure that `eBPF` support is fully functional from a program is not easy so I don't think this feature should be enabled by default.

rgacogne:
### Short description

This PR will use the default `BPF` filter, if any, to apply dynamic block generated via a `DynBlockRulesGroup`.  `eBPF` blocks are applied in kernel space and are much more efficient than user space ones.
We still also insert a regular dynamic block so that any failure to get the `eBPF` working will fall back to user space blocking automatically. The metrics will also transparently sum the hits from the `eBPF` block and the regular one.

<strike>This PR is based on top of #9756 so only the last commit matters here, and the PR will need to get rebased once #9756 has been merged.</strike>

Closes #6763.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/bpf-filter.cc
+++ b/pdns/bpf-filter.cc
@@ -214,16 +214,15 @@ void BPFFilter::removeSocket(int sock)
 
 void BPFFilter::block(const ComboAddress& addr)
 {
-  std::lock_guard<std::mutex> lock(d_mutex);
-
   uint64_t counter = 0;
   int res = 0;
-  if (addr.sin4.sin_family == AF_INET) {
+  if (addr.isIPv4()) {
     uint32_t key = htonl(addr.sin4.sin_addr.s_addr);
     if (d_v4Count >= d_maxV4) {
       throw std::runtime_error("Table full when trying to block " + addr.toString());
     }
 
+    std::lock_guard<std::mutex> lock(d_mutex);
     res = bpf_lookup_elem(d_v4map.fd, &key, &counter);
     if (res != -1) {
       throw std::runtime_error("Trying to block an already blocked address: " + addr.toString());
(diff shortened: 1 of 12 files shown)


omoerbeek (commented):
For the non EBPF case a few fields are unused, consider the diff below (plus adapting the ct).
```
+#ifdef HAVE_EBPF
   uint32_t d_maxV4;
   uint32_t d_maxV6;
   uint32_t d_maxQNames;
   uint32_t d_v4Count{0};
   uint32_t d_v6Count{0};
   uint32_t d_qNamesCount{0};
+#endif /* HAVE_EBPF */
```

pdns/bpf-filter.cc
```diff
@@ -386,6 +376,19 @@ std::vector<std::pair<ComboAddress, uint64_t> > BPFFilter::getAddrStats()
     v6Key[idx] = 0;
```
omoerbeek, line 376: memset?

pdns/bpf-filter.cc
```diff
@@ -386,6 +376,19 @@ std::vector<std::pair<ComboAddress, uint64_t> > BPFFilter::getAddrStats()
...
+      v4Addr.sin_addr.s_addr = ntohl(v4Key);
+      result.push_back(make_pair(ComboAddress(&v4Addr), value));
+    }
+
+    res = bpf_get_next_key(d_v4map.fd, &v4Key, &nextV4Key);
+  }
+
   res = bpf_get_next_key(d_v6map.fd, &v6Key, &nextV6Key);
 
   while (res == 0) {
```
omoerbeek, line 394: memcpy below?

omoerbeek (approved):
A few nits, I really like the "belt and suspenders" approach ;-)

Issue #9742, opened 2020-11-17
rec: qname minimization chases CNAMEs on intermediate names


Habbie:
- Program: Recursor
 - Issue type: Bug report

### Short description
While looking up `a.b.c.example.com`, recursor with QM also asks for `c` and `b.c`. If any of those have a CNAME, we uselessly chase that CNAME.

### Environment
 - Software version: any

### Steps to reproduce
1. tcpdump or enable trace or monitor queries some other way
2. ask for `terminal.cname.ent.experiments.powerdns.space`

### Expected behaviour
No query to the example.com NSes.

### Actual behaviour
Queries to the example.com NSes

Habbie:
Duplicate of #8699

omoerbeek:
Fixed by #9790

Issue #8699, opened 2020-01-13
rec: QNAME minimization resolves intermediate CNAMEs


mnordhoff:
- Program: Recursor
 - Issue type: Bug report

### Short description
For example, if you look up `foo.bar.example.com`, and there is a `CNAME` at `bar.example.com`, the Recursor resolves the `CNAME` target before continuing.

It's inefficient, and eats away at `max-qperq`, but it's pretty harmless. (Everything is cached, for better or worse.)

It could potentially cause problems in poorly managed zones, e.g. with very deep names, slow nameservers, and low TTLs.

If you want my opinion, it's not worth putting effort into unless it's very easy to change, or people run into real problems.

### Environment
 - Operating system: Ubuntu 16.04 amd64
 - Software version: 4.3.0~beta1+master.56.g3e56dd19a-1pdns.xenial
 - Software source: PowerDNS repository

### Steps to reproduce
1. Enable QNAME minimization, wipe cache.
2. `dig signaler-pa.clients6.google.com`

### Expected behaviour
No queries for `l.google.com` or `clients.l.google.com`.

### Actual behaviour
2 queries. 34 millisecond higher overall response time. ;-)

### Other information
```
2020-01-13 18:51:04.265042 IP 127.0.0.1.38740 > 127.0.0.1.53: 4829+ [1au] A? signaler-pa.clients6.google.com. (72)
2020-01-13 18:51:04.265907 IP 104.237.139.208.23352 > 216.239.38.10.53: 59105 [1au] A? clients6.google.com. (48)
2020-01-13 18:51:04.273093 IP 216.239.38.10.53 > 104.237.139.208.23352: 59105*- 2/0/1 CNAME clients.l.google.com., A 172.217.9.14 (88)
2020-01-13 18:51:04.273955 IP6 2600:3c00::f03c:91ff:fe96:6bcc.4091 > 2001:4860:4802:32::a.53: 42440 [1au] A? l.google.com. (41)
2020-01-13 18:51:04.289560 IP6 2001:4860:4802:32::a.53 > 2600:3c00::f03c:91ff:fe96:6bcc.4091: 42440*- 0/1/1 (91)
2020-01-13 18:51:04.290534 IP 104.237.139.208.59153 > 216.239.32.10.53: 604 [1au] A? clients.l.google.com. (49)
[...]

PR #9790, opened 2020-11-30
rec: Do not chase CNAME during qname minization step 4

rgacogne:
### Short description

Not chasing CNAMEs while we are only trying to determine where the next zone cut lies helps with some broken domains.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/syncres.cc
+++ b/pdns/syncres.cc
@@ -771,9 +771,12 @@ int SyncRes::doResolve(const DNSName &qname, const QType &qtype, vector<DNSRecor
 
       // Step 4
       QLOG("Step4 Resolve A for child");
+      bool oldFollowCNAME = d_followCNAME;
+      d_followCNAME = false;
       retq.resize(0);
       StopAtDelegation stopAtDelegation = Stop;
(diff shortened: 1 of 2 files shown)


omoerbeek:
Does seem to fix both #9742 and #8699

omoerbeek:
I am wondering if we want to move the check before the CNAME loop detection code. If we are not going to chase, do we still need the loop detection?

rgacogne:
> I am wondering if we want to move the check before the CNAME loop detection code. If we are not going to chase, do we still need the loop detection?

It's a good question. My immediate answer was that it still makes sense to detect a loop early, but then I realized not doing that might allow us to detect an issue during QM, and to retry with QM disabled. I'll do a few tests with the check moved before the CNAME loop detection.

rgacogne:
I did not see any issue by moving the check before the CNAME loop detection, so I just moved it. I amended the existing commit instead of adding a second one to make backports easier.

omoerbeek (approved):

PR #9801, opened 2020-12-02
dnsdist: Clean up the internal queues use for self-answered and trailing test responses

rgacogne:
### Short description

Hopefully it might help with the transient failures of the trailing data regression test we are recurrently experiencing in CI.

### Checklist

I have:
- [ ] compiled this code
- [x] tested this code
- [ ] documented the code


omoerbeek:
On my OpenBSD/arm64 machine (which picks up every commit on mater and builds & unit tests everything): 
```
unknown location(0): fatal error: in "dnsdistdynblocks_hh/test_DynBlockRulesMetricsCache_GetTopN": std::bad_alloc: std::bad_alloc
```

omoerbeek:
Turns out it was hitting a rather strict data ulimit. Raising it fixed the issue.

rgacogne:
Right, https://github.com/PowerDNS/pdns/pull/9756 increased the size of the ring buffers in that test suite, among other things.

PR #9807, opened 2020-12-02
rec: Untangle the validation/resolving qnames and qtypes

rgacogne:
### Short description

There was a confusion in `SyncRes::validateRecordsWithSigs` between the name and type we are currently resolving, and the name and type we are validating. It might have led to a spurious DNSSEC validation failure in some weird corner cases.

Could use some unit tests.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


omoerbeek (approved):
Reads correct. But please add a unit test indeed.

PR #9817, opened 2020-12-03
rec: Keep a cached, valid entry over a fresher Bogus one

rgacogne:
### Short description

It turns out to be quite difficult to make us accept a record that we already have in cache, thanks to sanitization, but let's make
sure that we will not replace a valid entry with a Bogus one if that happens.
It might happen for SOA records, and for DS records when the TTL of the corresponding NS records is shorter than the TTL of the DS.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/pdns/recursor_cache.cc
+++ b/pdns/recursor_cache.cc
@@ -405,7 +405,6 @@ void MemRecursorCache::replace(time_t now, const DNSName &qname, const QType& qt
   ce.d_qtype=qt.getCode();
   ce.d_signatures=signatures;
   ce.d_authorityRecs=authorityRecs;
-  ce.d_state=state;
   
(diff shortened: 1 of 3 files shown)


pdns/recursor_cache.cc
```diff
@@ -421,6 +420,16 @@ void MemRecursorCache::replace(time_t now, const DNSName &qname, const QType& qt
     }
   }
 
+  if (auth) {
+    /* only if the new entry is auth, we don't want to keep non-auth entry while we have a auth one */
```
omoerbeek, line 424: This comment is hard to read for me. Too much negatives. It is actually  not clear to me atm why we do not want to skip overwriting in the !auth case.

omoerbeek: Ah, it should be "we don't want to skip overwriting a non-auth entry while we have an auth one"

rgacogne:
Updated!

omoerbeek (approved):

PR #9819, opened 2020-12-04
Improve the way QType behaves by defining proper conversion operators and a hash function.

omoerbeek:
I converted syncres.* for starters to stop using uint16_t for qtype
and use QType everywhere.  Note that I also changed const QType&
in arg lists by QType, since it makes little sense to pass a 16 bit value
by const reference.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/qtype.cc
+++ b/pdns/qtype.cc
@@ -30,7 +30,7 @@
 
 static_assert(sizeof(QType) == 2, "QType is not 2 bytes in size, something is wrong!");
 
-const vector<QType::namenum> QType::names = {
+const map<const string, uint16_t> QType::names = {
   {"A", 1},
   {"NS", 2},
   {"CNAME", 5},
@@ -92,14 +92,21 @@ const vector<QType::namenum> QType::names = {
   {"LUA", 65402},
 };
 
+static map<uint16_t, const string> swapElements(const map<const string, uint16_t>& names) {
+  map<uint16_t, const string> ret;
+
+  for (const auto& n : names) {
+    ret.insert(make_pair(n.second, n.first));
+  }
+  return ret;
+}
+
+const map<uint16_t, const string> QType::numbers = swapElements(names);
+
+
 bool QType::isSupportedType() const
 {
-  for (const auto& pos : names) {
-    if (pos.second == code) {
-      return true;
-    }
-  }
-  return false;
+  return numbers.count(code) == 1;
 }
 
 bool QType::isMetadataType() const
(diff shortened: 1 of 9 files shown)


pdns/qtype.cc
```diff
@@ -30,23 +30,79 @@
 
 static_assert(sizeof(QType) == 2, "QType is not 2 bytes in size, something is wrong!");
 
-vector<QType::namenum> QType::names;
-// XXX FIXME we need to do something with initializer order here!
-QType::init QType::initializer; 
-
-QType::QType()
+const vector<QType::namenum> QType::names = {
```
pieterlexis, line 33: I wonder if we could wrap this into `boilerplate_conv` via DNSRecords. Or make `boilerplate_conv` take a QType, so we never have the problem of types not being in sync with supported types?

omoerbeek: I changed the boilerplate so that it uses QType::RNAME, enforcing it must exists. That's a start, I'd say.

pieterlexis:
> Note that I also changed const QType& in arg lists by QType, since it makes little sense to pass a 16 bit value by const reference.

It might still be good to keep them `const` just in case?

omoerbeek:
> > Note that I also changed const QType& in arg lists by QType, since it makes little sense to pass a 16 bit value by const reference.
> 
> It might still be good to keep them `const` just in case?

It is uncommon the specify `const` in a prototype non-reference or pointer args.

It's call by value anyway so the caller is safe form modifications by the callee.  As for the callee, `const` only prevents changing the value in its body.

pdns/qtype.hh
```diff
@@ -38,206 +36,123 @@
...
 
-
-
 class QType
 {
 public:
-  QType(); //!< Naked constructor
-  explicit QType(uint16_t); //!< convert from an integer to a QType
-  QType(const QType& orig) : code(orig.code)
+  QType(uint16_t qtype = 0) : code(qtype) {}
```
rgacogne, line 42: I'm not sure it's such a good idea to remove the `explicit` keyword on the constructor from `uint16_t`?

rgacogne:
I agree it does not make sense to mark the `QType` as `const` in the declaration, but it might still make sense to mark it as `const` in the definition, as it does prevent mistakenly changing the value in the function.

pdns/qtype.hh
omoerbeek: It's conventient to be able to pass e.g. `QType::A` to a function taking a `QType`.

rgacogne: That makes sense, so let's keep it that way.

pieterlexis:
@omoerbeek can you rebase so we can review?

omoerbeek:
Rebased

pieterlexis (approved):
:heart:

Habbie (approved):

PR #9821, opened 2020-12-07
rec-4.4.x: Move to several distinct Bogus states, for easier debugging

rgacogne:
### Short description

Backport of #9597 to rel/rec-4.4.x, with an additional commit to export the sum of all the new 'Bogus*' metrics as the existing 'Bogus' metric, to prevent messing up with existing dashboards.
At second thought, I think it would actually make sense to forward port that commit :-)

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/pdns/lua-recursor4.cc
+++ b/pdns/lua-recursor4.cc
@@ -355,6 +355,13 @@ void RecursorLua4::postPrepareContext()
         {"BogusNoRRSIG", static_cast<unsigned int>(vState::BogusNoRRSIG) },
         {"BogusNoValidRRSIG", static_cast<unsigned int>(vState::BogusNoValidRRSIG) },
         {"BogusMissingNegativeIndication", static_cast<unsigned int>(vState::BogusMissingNegativeIndication) },
+        {"BogusSignatureNotYetValid", static_cast<unsigned int>(vState::BogusSignatureNotYetValid)},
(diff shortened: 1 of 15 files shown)


rgacogne:
This might introduce a regression for Lua scripts detecting whether an existing answer is in the `Bogus` state. I'll try to find a neat way to prevent that.

rgacogne:
> This might introduce a regression for Lua scripts detecting whether an existing answer is in the Bogus state. I'll try to find a neat way to prevent that.

I pushed a commit keeping compatibility with Lua scripts using the existing 'Bogus' state.

omoerbeek:
Nice, that last one should be merged to master as well.

rgacogne:
> Nice, that last one should be merged to master as well.

I was actually wondering whether we should add that to master, or better document the change? I'll add the documentation anyway, of course, but I guess we might want the compatibility layer in 4.5, and perhaps remove it in the next major version?

omoerbeek:
> > Nice, that last one should be merged to master as well.
> 
> I was actually wondering whether we should add that to master, or better document the change? I'll add the documentation anyway, of course, but I guess we might want the compatibility layer in 4.5, and perhaps remove it in the next major version?

Yes, that sounds like a good plan. Though the tricky thing is actually doing the last part in m y experience.

omoerbeek (approved):

PR #9832, opened 2020-12-08
UUID: Use the non-cryptographic variant of the boost::uuid

rgacogne:
### Short description

Since Boost 1.67.0 the default UUID generator is cryptographically strong, which is neat but quite slower. Since we don't need that, just use the fastest version.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/pdns/Makefile.am
+++ b/pdns/Makefile.am
@@ -891,7 +891,8 @@ speedtest_SOURCES = \
 	speedtest.cc \
 	statbag.cc \
 	svc-records.cc svc-records.hh \
-	unix_utility.cc
+        unix_utility.cc \
+        uuid-utils.cc
 
 speedtest_LDFLAGS = $(AM_LDFLAGS) $(LIBCRYPTO_LDFLAGS)
 speedtest_LDADD = $(LIBCRYPTO_LIBS) \
--- a/pdns/speedtest.cc
+++ b/pdns/speedtest.cc
@@ -8,6 +8,7 @@
 #include "dnsrecords.hh"
 #include "iputils.hh"
 #include <fstream>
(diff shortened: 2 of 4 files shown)


rgacogne:
Whoops, looks like I cleaned my code a bit too much after testing since it doesn't compile anymore, fixing that!

omoerbeek:
We have to make sure a non-seeded `mt19937` does not always starts in the same state, and thus generate the same uuids for different runs. I'll check this soon.

rgacogne:
That's a very good point! From the documentation in the uuid class I did not expect we needed to seed it, but the documentation in the random class makes me think it can very well be. Now I wonder if we have been doing it wrong in the past..

zeha:
The old boost implementations (which used MT) seem to seed it at construct time.

rgacogne:
The `basic_random_generator` default constructor has a comment pretending it will be correctly seeded, but let's check ourselves:

```
    // default constructor creates the random number generator and                                                                                                                                                   
    // if the UniformRandomNumberGenerator is a PseudoRandomNumberGenerator                                                                                                                                          
    // then it gets seeded by a random_provider.   
```

omoerbeek:
Test on OpenBSD (boost 1.70 shows the MT is seeded. Original code: 
```
'UUIDGenTest' 0.21 seconds: 730515.1 runs/s, 1.37 usec/run
```
New code:
```
'UUIDGenTest' 0.10 seconds: 9190128.8 runs/s, 0.11 usec/run
```
I'll push the speedtest to this branch.

omoerbeek (approved):

PR #9836, opened 2020-12-09
circleci: extend coverage some more and disable travis

Habbie:
### Short description

* switch to clang-8
* test Recursor SNMP, DNSTAP, Protobuf
* enable asan+ubsan for all products
* fixes bugs that came out of the previous two items

Missing:
* ~ubsan for rec~
* ~asan+ubsan for auth (#9835)~
* ~`--enable-option-checking=fatal`~
* ~remotebackend: see if there is a more modern iteration form that does not break on asan~ moved to #9915 

Problem:
* ~recursor bulktest appears to be failing~
* ~`==15170==WARNING: invalid path to external symbolizer!` (works on my system, but not in CircleCI, need to figure out why)~

### Checklist

I have:
- [ ] read the [CONTRIBUTING.md](https://github.com/PowerDNS/pdns/blob/master/CONTRIBUTING.md) document
- [ ] compiled this code
- [ ] tested this code
- [ ] included documentation (including possible behaviour changes)
- [ ] documented the code
- [ ] added or modified regression test(s)
- [ ] added or modified unit test(s)
- [ ]  checked that this code was merged to master

PR #9843, opened 2020-12-09
Use protozero for Protocol Buffer operations in dnsdist, and dnstap/outgoing for the rec

### LINKED ISSUE #9780: dnsdist: Ponder switching to protozero
- Program: dnsdist 
 - Issue type: Feature request

### Short description
Basically https://github.com/PowerDNS/pdns/pull/9630 but for dnsdist. It's a bit more complicated because dnsdist allows alteration of the protobuf message by Lua hooks, but I believe we can work something out and the performance gains are too good to ignore.

### LINKED ISSUE #9781: rec: more protozero
Move to protozero for outgoing protobuf logging as well.

rgacogne:
### Short description

This PR:
- switches from libprotobuf to protozero to generate Protocol Buffer and dnstap messages in dnsdist ;
- same for outgoing Protocol Buffer and dnstap messages in the recursor ;
- same for dnscap2protobuf in the auth ;
- removes the dependencies on `libprotobuf` and `proto-c` ;
- removes the `--with-protobuf` option since protozero is always available, removing a ton of `#ifdef`s.

Using protozero in dnsdist results in no new allocation when a message in generated, against ~8 of them before.

Closes #9780.
Closes #9781.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


rgacogne:
Great, it looks like applying our formatting script locally gives a different result than the one expected in Circle CI..

pdns/lwres.cc
```diff
@@ -81,9 +80,8 @@ static void logFstreamQuery(const std::shared_ptr<std::vector<std::unique_ptr<Fr
 
   struct timespec ts;
   TIMEVAL_TO_TIMESPEC(&queryTime, &ts);
-  RecDnstapMessage message(SyncRes::s_serverID, nullptr, &ip, doTCP, auth, reinterpret_cast<const char*>(&*packet.begin()), packet.size(), &ts, nullptr);
   std::string str;
-  message.serialize(str);
+  DnstapMessage message(str, 4, SyncRes::s_serverID, nullptr, &ip, doTCP, reinterpret_cast<const char*>(&*packet.begin()), packet.size(), &ts, nullptr, auth);
```
omoerbeek, line 84:
```suggestion
  DnstapMessage message(str, 3, SyncRes::s_serverID, nullptr, &ip, doTCP, reinterpret_cast<const char*>(&*packet.begin()), packet.size(), &ts, nullptr, auth);
```

pdns/lwres.cc
```diff
@@ -112,9 +110,8 @@ static void logFstreamResponse(const std::shared_ptr<std::vector<std::unique_ptr
   struct timespec ts1, ts2;
   TIMEVAL_TO_TIMESPEC(&queryTime, &ts1);
   TIMEVAL_TO_TIMESPEC(&replyTime, &ts2);
-  RecDnstapMessage message(SyncRes::s_serverID, nullptr, &ip, doTCP, auth, static_cast<const char*>(&*packet.begin()), packet.size(), &ts1, &ts2);
   std::string str;
-  message.serialize(str);
+  DnstapMessage message(str, 3, SyncRes::s_serverID, nullptr, &ip, doTCP, static_cast<const char*>(&*packet.begin()), packet.size(), &ts1, &ts2, auth);
```
omoerbeek, line 114:
```suggestion
  DnstapMessage message(str, 4, SyncRes::s_serverID, nullptr, &ip, doTCP, static_cast<const char*>(&*packet.begin()), packet.size(), &ts1, &ts2, auth);
```

pdns/dnsparser.hh
```diff
@@ -37,6 +37,7 @@
 #include "pdnsexception.hh"
```
omoerbeek, line 37: At first sight this looks a bit like an unrelated change,

omoerbeek (approved):
This is a big one! I did not read everything, but in general it looks very good and I'm very happy with it.

I tested dnstap, regular protobuf and outgoing protobuf in rec (and noted one issue in the dnstap code as already mentioned). 

I also reviewed the recursor changes in detail, since they are a bit tricky with adding to a partially constructed message. That all looks fine and tests ok.

Two things remain:
- It really could use more extensive regression tests.
- Also, a few symbolic constants or enums would make the code easier to read and might even have prevented the 3/4 mixup.

rgacogne:
> It really could use more extensive regression tests.

It turns out that the current regression tests are quite good, and would have caught the issue if they were run in our CI.. Which https://github.com/PowerDNS/pdns/pull/9836 will solve. And apparently I forgot to run them locally, as I was too focused on the protobuf ones.

> Also, a few symbolic constants or enums would make the code easier to read and might even have prevented the 3/4 mixup.

I'll push a couple commits doing just that in a few moments.

Thanks for the review, Otto, much appreciated!

rgacogne:
Otto, do you want to take a look at the two latest commits before I merge this PR?

There is no rush, I won't merge it before auth-4.4.0 since I'm afraid the removal of libprotobuf might cause a few build issues here and there.

rgacogne:
Rebased to fix a conflict introduced by the `incbin` removal.

PR #9856, opened 2020-12-11
rec: Use a short-lived NSEC3 hashes cache for denial validation

rgacogne:
### Short description

It turns out that computing those SHA1 hashes is far from cheap, and in almost all cases the salt and iterations are identical
so no need to compute them several times.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/validate.cc
+++ b/pdns/validate.cc
@@ -90,14 +90,15 @@ static std::string getHashFromNSEC3(const DNSName& qname, const std::shared_ptr<
     return result;
   }
 
-  auto it = cache.find({qname, nsec3->d_salt, nsec3->d_iterations});
+  auto key = std::make_tuple(qname, nsec3->d_salt, nsec3->d_iterations);
+  auto it = cache.find(key);
   if (it != cache.end())
   {
     return it->second;
   }
 
   result = hashQNameWithSalt(nsec3->d_salt, nsec3->d_iterations, qname);
-  cache[{qname, nsec3->d_salt, nsec3->d_iterations}] = result;
+  cache[key] = result;
   return result;
 }
 
--- a/pdns/validate.cc
+++ b/pdns/validate.cc
@@ -80,7 +80,9 @@ static bool nsecProvesENT(const DNSName& name, const DNSName& begin, const DNSNa
   return begin.canonCompare(name) && next != name && next.isPartOf(name);
 }
 
-static std::string getHashFromNSEC3(const DNSName& qname, const std::shared_ptr<NSEC3RecordContent>& nsec3)
+using nsec3HashesCache = std::map<std::tuple<DNSName, std::string&, uint16_t>, std::string>;
+
+static std::string getHashFromNSEC3(const DNSName& qname, const std::shared_ptr<NSEC3RecordContent>& nsec3, nsec3HashesCache& cache)
 {
   std::string result;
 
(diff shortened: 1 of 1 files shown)


omoerbeek (commented):
Initially I was confused since the order of things is different between the map and the call to `hashQNameWithSalt`. It's also not immediately clear why `DNSName` is not a ref in the tuple, but salt is. Added comments to explain that would be nice.

rgacogne:
> It's also not immediately clear why DNSName is not a ref in the tuple, but salt is.

Because `salt` lives from the duration of the cache while the name does not. I'll drop that optimization, it might lead to issues later.

> Initially I was confused since the order of things is different between the map and the call to hashQNameWithSalt
 
I'm afraid I don't understand the confusion. Do you mean the order in the tuple? It does not really matter, but since the name is more likely to differs than the salt or the number of iterations, I want to compare the name first.

omoerbeek:
> > It's also not immediately clear why DNSName is not a ref in the tuple, but salt is.
> 
> Because `salt` lives from the duration of the cache while the name does not. I'll drop that optimization, it might lead to issues later.

Agreed.

> 
> > Initially I was confused since the order of things is different between the map and the call to hashQNameWithSalt
> 
> I'm afraid I don't understand the confusion. Do you mean the order in the tuple? It does not really matter, but since the name is more likely to differs than the salt or the number of iterations, I want to compare the name first.

Yes, I meant the order of the tuple. Initially my brain did (wrongly) assume the order of args in the function call and the order of the tuple would be the same. It was probably a end-of-week thing.

omoerbeek (approved):

rgacogne:
Compilation fails on Travis:
```
validate.cc:93:67: error: converting to ‘const key_type {aka const std::tuple<DNSName, std::basic_string<char, std::char_traits<char>, std::allocator<char> >&, short unsigned int>}’ from initializer list would use explicit constructor ‘constexpr std::tuple< <template-parameter-1-1> >::tuple(_UElements&& ...) [with _UElements = {const DNSName&, std::basic_string<char, std::char_traits<char>, std::allocator<char> >&, short unsigned int&}; <template-parameter-2-2> = void; _Elements = {DNSName, std::basic_string<char, std::char_traits<char>, std::allocator<char> >&, short unsigned int}]’
   auto it = cache.find({qname, nsec3->d_salt, nsec3->d_iterations});
                                                                   ^
validate.cc:100:52: error: converting to ‘std::map<std::tuple<DNSName, std::basic_string<char, std::char_traits<char>, std::allocator<char> >&, short unsigned int>, std::basic_string<char> >::key_type {aka std::tuple<DNSName, std::basic_string<char, std::char_traits<char>, std::allocator<char> >&, short unsigned int>}’ from initializer list would use explicit constructor ‘constexpr std::tuple< <template-parameter-1-1> >::tuple(_UElements&& ...) [with _UElements = {const DNSName&, std::basic_string<char, std::char_traits<char>, std::allocator<char> >&, short unsigned int&}; <template-parameter-2-2> = void; _Elements = {DNSName, std::basic_string<char, std::char_traits<char>, std::allocator<char> >&, short unsigned int}]’
   cache[{qname, nsec3->d_salt, nsec3->d_iterations}] = result;
```

Although it does compile on CentOS 6 so we could get away with it, but let's try to be as portable as possible.

PR #9859, opened 2020-12-11
Don't copy the same salt for all iterations in hashQNameWithSalt()

rgacogne:
### Short description

The salt does not change between iterations, and the hash size is constant, so we can just overwrite the hash instead.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/pdns/dnssecinfra.cc
+++ b/pdns/dnssecinfra.cc
@@ -491,6 +491,7 @@ string hashQNameWithSalt(const NSEC3PARAMRecordContent& ns3prc, const DNSName& q
 
 string hashQNameWithSalt(const std::string& salt, unsigned int iterations, const DNSName& qname)
 {
+  // rfc5155 section 5
   unsigned int times = iterations;
   unsigned char hash[SHA_DIGEST_LENGTH];
   string toHash(qname.toDNSStringLC() + salt);
--- a/pdns/dnssecinfra.cc
+++ b/pdns/dnssecinfra.cc
@@ -492,16 +492,34 @@ string hashQNameWithSalt(const NSEC3PARAMRecordContent& ns3prc, const DNSName& q
 string hashQNameWithSalt(const std::string& salt, unsigned int iterations, const DNSName& qname)
 {
   unsigned int times = iterations;
-  unsigned char hash[20];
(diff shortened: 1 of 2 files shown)


pdns/dnssecinfra.cc
```diff
@@ -492,16 +492,34 @@ string hashQNameWithSalt(const NSEC3PARAMRecordContent& ns3prc, const DNSName& q
...
-
-  for(;;) {
-    toHash.append(salt);
-    SHA1((unsigned char*)toHash.c_str(), toHash.length(), hash);
-    toHash.assign((char*)hash, sizeof(hash));
-    if(!times--)
+  unsigned char hash[SHA_DIGEST_LENGTH];
+  string toHash(qname.toDNSStringLC() + salt);
+  if (toHash.capacity() < (salt.size() + sizeof(hash))) {
+    toHash.reserve(salt.size() + sizeof(hash));
```
rgacogne, line 498: I'm wondering now if that call to `std::string::reserve()` is not a misoptimization:
- when `iterations` == 1, which seems to happen quite often, we might be just doing a needless reallocation if the string is already large enough to hold the hash ;
- with a larger number of iterations, doing the reallocation now or after the first iteration does not really matter.

I'll remove these lines.

omoerbeek (commented):
It looks good from code inspection. Are these changes are covered by some unit test?

rgacogne:
> Are these changes are covered by some unit test?

Weirdly enough, they are not. I'll add some tests!

rgacogne:
> I'll add some tests!

Done :)

rgacogne:
And even a few more unit tests while I was at it :)

omoerbeek:
Failing regression test seems unrelated...

omoerbeek (approved):

Habbie:
> Failing regression test seems unrelated...

SNMP, so a rebase might help I think.

rgacogne:
Rebased!

rgacogne:
Still failing, and this time it looks like it might be a crash. I'll investigate.

rgacogne:
I couldn't reproduce it locally and it went away after restarting the test. On the bright side it yielded #10036 :)

PR #9862, opened 2020-12-11
dnsdist: Add a Lua binding for the number of queries dropped by a server

### LINKED ISSUE #9861: getServer property for drops
- Program: dnsdist 
 - Issue type: Feature request

### Short description
It would be nice to have a directly accessible property for backend DNS server "Drops" 
These are listed under showServers() for all backends, but not accessible on an individual level.

Something like getServer(0).drops or getServer(0).dropCount

### Usecase
I am using https://github.com/sysadminblog/dnsdist-configs/blob/master/orderedLeastOutstanding.lua to provide fixed failover between servers (i.e. always use X until X is down, then use Y), although have modified the line:

`if (servers[i].upStatus == true) then`

to be

`if (servers[i].upStatus == true) then
      if (servers[i]:isUp() == true) then
`

...as isUp will respond differently based on if the server is marked down outside of a health check.

However, if a healthcheck is set to a high number (to avoid log pollution on the respective backend) then it takes whatever interval is present before failover happens.

I would like to use a Lua query (and local var) to track the number of drops, and once the drops increment beyond a given number (since the last set point), I can manually mark the backend as down, so my Lua if statement (above) will catch it,
dnsdist's own healthcheck will "catch up" on it's own.

A separate "last check" variable on the healthcheck can then be used to note dnsdists own healthcheck resuming (marked back up) and manually mark the server up, so "isUp" responds positively also. 

### Description
Provide a property for the server class that responds with that backend servers "drop" count

rgacogne:
### Short description

Closes #9861.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdistdist/docs/reference/config.rst
+++ b/pdns/dnsdistdist/docs/reference/config.rst
@@ -609,7 +609,7 @@ A server object returned by :func:`getServer` can be manipulated with these func
 
     Get the number of dropped queries for this server.
 
-    :returns: The number of outstanding queries
+    :returns: The number of dropped queries
 
   .. method:: Server:getOutstanding() -> int
 
--- a/pdns/dnsdistdist/docs/reference/config.rst
+++ b/pdns/dnsdistdist/docs/reference/config.rst
@@ -603,6 +603,14 @@ A server object returned by :func:`getServer` can be manipulated with these func
 
     :returns: A string containing the server name if any plus the server address and port
 
+  .. method:: Server:getDrops() -> int
+
+    .. versionadded:: 1.6.0
+
+    Get the number of dropped queries for this server.
+
+    :returns: The number of outstanding queries
+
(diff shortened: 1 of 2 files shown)


pdns/dnsdistdist/docs/reference/config.rst
```diff
@@ -603,6 +603,14 @@ A server object returned by :func:`getServer` can be manipulated with these func
 
     :returns: A string containing the server name if any plus the server address and port
 
+  .. method:: Server:getDrops() -> int
+
+    .. versionadded:: 1.6.0
+
+    Get the number of dropped queries for this server.
+
+    :returns: The number of outstanding queries
```
omoerbeek, line 612:
```suggestion
    :returns: The number of dropped queries
```

PR #9876, opened 2020-12-16
More toolsets in boost m4 code

### LINKED ISSUE #9871: boost.m4 does not contain modern toolchain tests
boost.m4 is too old, eg. clang 10 on OpenBSD is detected as 
```
checking for the toolset name used by Boost for c++... gcc42 -gcc               
```
At lest for rec I do not see immediate consequences of this misdetection, but better safe than sorry.
Also newest versions of gcc are missing in the checks.

omoerbeek:
### Short description

Fixes #9871 

Also upstreamed.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/m4/boost.m4
+++ b/m4/boost.m4
@@ -22,7 +22,7 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 m4_define([_BOOST_SERIAL], [m4_translit([
-# serial 32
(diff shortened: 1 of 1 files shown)


pieterlexis (approved):
:+1: merged upstream

PR #9881, opened 2020-12-16
el6 docker builds: prefer v4 resolution

Habbie:
### Short description
This avoids yum breaking on machines without v6.

### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/builder-support/dockerfiles/Dockerfile.target.centos-6
+++ b/builder-support/dockerfiles/Dockerfile.target.centos-6
@@ -5,6 +5,12 @@
 # Put only the bare minimum of common commands here, without dev tools
 FROM centos:6 as dist-base
 COPY /builder-support/dockerfiles/Centos-Base-Vault.repo /etc/yum.repos.d/CentOS-Base.repo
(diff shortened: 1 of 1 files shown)


rgacogne (approved):

treysis:
This is bad. It shouldn't be the default. Default is IPv6 enabled. Whoever disables IPv6 has to apply the workarounds manually. Again: defaulting to IPv4 is the wrong solution.

dwfreed:
This Dockerfile is for *building* PowerDNS, not for running it.  The primary consumer (and I expect only consumer) is PowerDNS's own build infrastructure.  They are merely included in this repo for (I imagine) build transparency and simplicity of implementation.  This change is what works best for their infrastructure, in particular the automated nature of it precludes manually applying these changes on containers that don't have v6 capability, since the containers are created for the purpose of running the build and then destroyed.  Should you wish to build PowerDNS yourself using this Dockerfile, you're welcome to modify it in your fork to remove these changes, but I find it highly unlikely that a PR to revert these changes would be accepted.

PR #9883, opened 2020-12-16
rec: Lookup DS entries before CNAME entries

### LINKED ISSUE #9621: rec: www.dpaste.com sometimes bogus; insecure domain with apex CNAME
- Program: Recursor
 - Issue type: Bug report

### Short description
`www.dpaste.com` sometimes validates as bogus. `dpaste.com` is always fine.

`dpaste.com` is an insecure domain with a zone apex `CNAME`, so none of it deserves to resolve, but the precise way in which it's failing is a bit odd.

I'm not sure what is required for it to be bogus, but it seems to involve learning the `dpaste.com` `SOA` record and/or the `CNAME` record.

### Environment
 - Operating system: Ubuntu 16.04 x86-64
 - Software version: 4.5.0~alpha0+master.190.g8446e56b8-1pdns.xenial and a couple slightly older builds

### Steps to reproduce
1. Restart Recursor
2. `dig dpaste.com aaaa`
3. `dig www.dpaste.com`

### Expected behaviour
Consistent success or consistent failure.

### Actual behaviour
Some things are bogus and some things aren't depending on what steps you take.

### Other information
Unless I'm making a mistake, this isn't bogus:

1. `dig dpaste.com`
2. `dig www.dpaste.com`

Or this:

1. `dig dpaste.com soa`
2. `dig www.dpaste.com`

But this is:

1. `dig dpaste.com`
2. `dig dpaste.com soa`
3. `dig www.dpaste.com`

`trace-regex 'dpaste\.com\.$'` for `dig dpaste.com aaaa` and `dig www.dpaste.com`: https://mn0.us/ojCK

`recursor.conf`:
```
allow-from=127.0.0.0/8, ::1/128
allow-trust-anchor-query
carbon-ourname=clover_mattnordhoff_net
carbon-server=2a02:2770:8::2635:0:1
config-dir=/etc/powerdns
dnssec=validate
dnssec-log-bogus
hint-file=/usr/share/dns/root.hints
include-dir=/etc/powerdns/recursor.d
local-address=127.0.0.1, ::1
log-common-errors
lua-config-file=/etc/powerdns/recursor.lua
max-cache-ttl=172800
max-negative-ttl=10800
query-local-address=0.0.0.0, ::
quiet=yes
setgid=pdns
setuid=pdns
threads=1
udp-truncation-threshold=4096
```
(I know, I'm a Flag Day traitor!)

rgacogne:
### Short description

When we are looking for a `DS`, we want to do the non-`CNAME` cache check first because we can actually have a `DS` (from the parent zone) AND a `CNAME` (from the child zone), and what we really want is the `DS`.

<strike>Note that this works because we do compute the zone cuts first, so we will have the `DS` (or denial of) in cache. We will need to be careful about that if we stop doing that, since the `DS` entry might expire while the `CNAME` one is still valid.</strike> I was mistaken, this PR makes the situation better when the `DS` entry (or denial of) is still there but doesn't work when it has expired and the `CNAME` hasn't.

<strike>Still need a unit test.</strike>

Fixes #9621.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/pdns/reczones.cc
+++ b/pdns/reczones.cc
@@ -46,6 +46,7 @@ bool primeHints(void)
 {
   // prime root cache
   const vState validationState = vState::Insecure;
+  static const ComboAddress from("255.255.255.255");
   vector<DNSRecord> nsset;
   t_rootNSZones.clear();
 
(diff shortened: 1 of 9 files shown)


rgacogne:
I added a unit test showing the solved and unsolved issues.

pdns/reczones.cc
```diff
@@ -46,6 +46,7 @@ bool primeHints(void)
 {
   // prime root cache
   const vState validationState = vState::Insecure;
+  static const ComboAddress from("255.255.255.255");
```
omoerbeek, line 49: No use in making this one `static`, there is no extra value to save this one between calls.

pdns/syncres.cc
```diff
@@ -945,6 +949,37 @@ int SyncRes::doResolveNoQNameMinimization(const DNSName &qname, const QType &qty
 
       return res;
     }
+
+    /* if we have not found a cached DS (or denial of), now is the time to look for a CNAME */
```
omoerbeek, line 953: Don't like the repeated block a lot but I do not see an easy way to get rid of it.

omoerbeek (approved):

pdns/syncres.cc
rgacogne: Agreed, I tried a few ideas but they all turned out worse than duplicating the block, so I left it that way.

PR #9885, opened 2020-12-16
Clear the LMDB set state when performing a new lookup or list to prevent corruption cases

RobinGeuze:
### Short description
Prevents errors of the following kind from occurring if LMDBBackend::get should throw an exception for some other reason:
```
Exception building answer packet for example.com/SOA (vector::_M_range_check: __n (which is 1410) >= this->size() (which is 1)) sending out servfail
```
Port of #9884 to master.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/modules/lmdbbackend/lmdbbackend.cc
+++ b/modules/lmdbbackend/lmdbbackend.cc
@@ -781,20 +781,24 @@ bool LMDBBackend::get(DNSZoneRecord& rr)
       key = d_currentKey.get<string_view>();
     }
 
-    const auto& drr = d_currentrrset.at(d_currentrrsetpos++);
+    try {
+      const auto& drr = d_currentrrset.at(d_currentrrsetpos++);
 
-    rr.dr.d_name = compoundOrdername::getQName(key) + d_lookupdomain;
-    rr.domain_id = compoundOrdername::getDomainID(key);
-    rr.dr.d_type = compoundOrdername::getQType(key).getCode();
-    rr.dr.d_ttl = drr.ttl;
-    rr.dr.d_content = deserializeContentZR(rr.dr.d_type, rr.dr.d_name, drr.content);
-    rr.auth = drr.auth;
+      rr.dr.d_name = compoundOrdername::getQName(key) + d_lookupdomain;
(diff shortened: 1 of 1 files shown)


rgacogne (approved):

Habbie:
Reproduction (we should make a test out of this):

Start here:
```
$ ppdnsutil --config-dir=. --config-name=lmdb list-zone example.com
$ORIGIN .
example.com	3005	IN	NS	ns.
example.com	3005	IN	SOA	a.misconfigured.dns.server.invalid hostmaster.example.com 12347 10800 3600 604800 3600
bar.example.com	3005	IN	A	1.2.3.4
foo.example.com	3005	IN	CNAME	foofoo.example.com.
```

Then:

```
$ mdb_dump -n -p -a pdns.lmdb-0  > dump
```

(I have `lmdb-shards=1`)

Edit the dump like this:

```diff
- \09\00\06foofoo\c0\15\bd\0b\00\00\01\00\01
+ \09\00\06foofoof\c0\15\bd\0b\00\00\01\00\01
```

Load it:
```
$ mdb_dump -n -p -a pdns.lmdb-0  > dump
```

(on my LMDB version, I got an error about `db_pagesize` so I removed that line from the dump and imported again)

Then, run pdns with `distributor-threads=1` and `receiver-threads=1`.

Then query: `dig a @127.0.0.1 -p 5300 {foo,bar}.example.com`

Without this PR, both queries fail. With this PR, the second query succeeds.

Habbie:
@mind04 suggests that, on top of this fix, the failure mode should perhaps throw `PDNSException` to avoid any other possibilities of backend state being confused.

Habbie:
Even though this PR in its current state has been backported, I'm adding a backport label for the PDNSException idea.

Habbie:
> @mind04 suggests that, on top of this fix, the failure mode should perhaps throw `PDNSException` to avoid any other possibilities of backend state being confused.

Pushed. I verified that -either commit-, -individually-, will make the second (`bar`) lookup work.

Habbie:
And rebased.

PR #9899, opened 2020-12-23
YAML Configuration parser 

pieterlexis:
### Short description
This WIP-PR includes a new, YAML-based configuration parser. This parser and assorted functions live in the `pdns::config` namespace.

Programs need to register their options, including several callbacks (ixfrdist uses lambda's, but those are optional) to check and apply new configuration, to return the current configuration and one to return the defaults.

This code supports the following things:

- Reading YAML nodes as DNSNames, ComboAddresses, Netmasks, and NetmaskGroups
- Reading vectors etc. of the above
- Potentially allow for runtime-updating of configuration (by re-reading the config-file and calling the check and apply functions).

It works for IXFRDist. This PR is opened to receive feedback on the design and to get ideas for improvements.

For PowerDNS (auth and rec), some questions remain:

 - How will this work in combination with command-line arguments (or will those disappear)?
 - How to migrate from the old .conf to the new YAML config

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code


zeha:
Design question:
How will this work for `launch+=...` across multiple included files?

omoerbeek:
> Design question:
> How will this work for `launch+=...` across multiple included files?

In am more general way: how are multiple config files (with e.g. a config dir) handled? I can think of extending, overwriting or erroring? It probably should be part of the meta data how to handle that.

omoerbeek:
Some general comments

- `registerOption` takes a shared pointer, I do not really see why, I think a ref could do.
- There's the issue of merging vs overwriting and having multiple config files. In particular

```
bla:
  value: 10
  anotherValue: 20
```
and
```
bla:
  value: 20
```
in a file processed later is hairy. I think the result should likely be merged to
```
bla:
  value: 20
  anotherValue: 20
```
But for configs like launch and forward defs that won't do: you typically want to extend the existing values:
```
  forwarder:
     - domain: aap.nl
       address: 1.2.3.4
```
and another file 
```
  forwarder:
     - domain: noot.nl
       address: 1.2.3.4
```
should result in two domains being forwarded.

We have to come up with clear semantics for the diffferent situations.

pieterlexis:
needs more thinking and work

PR #9902, opened 2020-12-27
Add Debian Bullseye support to package builder

kpfleming:
### Short description
In preparation for support of Debian Bullseye, add package-building mechanics so they can be tested.

Also address a couple of other small issues found during testing.

### Checklist

I have:
- [ ] compiled this code
- [x] tested this code
- [ ] documented the code


zeha:
The idea was to do a packaging resync with Debian after 4.4.0 was out.

kpfleming:
That sounds promising, although I suspect in the long term there will still be a need to publish packages separately... once 4.5.0 is released, it won't be included in Bullseye.

zeha:
Well yes, but we (or least I) don't want the packaging to diverge too much. I'll see about doing the resync sooner than later.
(4.4.0 is already in bullseye. Binary packages for lmdb and lua2 are also on their way.)

kpfleming:
Ha! I wish I had checked for that before I setup my own private repository and built packages of it :-)

This also solves the lack of Debian armhf and arm64 packages, since the upstream repos will have those by default.

kpfleming:
If you do decide that the Debian packaging bits of this PR aren't needed, I'll split out the first commit (the one that changes to fully-qualified registry/image names for container images), since that is still useful. I'm making builds using Podman instead of Docker, and in Bullseye it already requires explicit configuration to support unqualified image names.

zeha:
@kpfleming my debian bits are in #9964

kpfleming:
That's much more comprehensive than mine, so I'm fine with dropping this PR.

pieterlexis:
Closing in favor of #9964

PR #9909, opened 2021-01-05
Enable address sanitizer on MacOS

omoerbeek:
### Short description

Tests OK on Intel MacOS. Apple Silicon shows a few issues which I am investigating.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/recursordist/configure.ac
+++ b/pdns/recursordist/configure.ac
@@ -95,7 +95,7 @@ PDNS_CHECK_CURL
 
 dnl the *_r functions are in posix so we can use them unconditionally, but the ext/yahttp code is
 dnl using the defines.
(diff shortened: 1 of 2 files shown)


pdns/recursordist/configure.ac
```diff
@@ -95,7 +95,7 @@ PDNS_CHECK_CURL
 
 dnl the *_r functions are in posix so we can use them unconditionally, but the ext/yahttp code is
```
rgacogne, line 96: That comment is no longer true, though.

rgacogne (approved):

pdns/recursordist/configure.ac
rgacogne: It actually is, my bad!

PR #9913, opened 2021-01-05
Move to c++17

omoerbeek:
### Short description

Only compile tested with clang on OpenBSD so far.

Creating a PR so the CI and others can easily use this branch to test.

### Checklist

I have:
- [X] compiled this code
- [ ] tested this code
- [ ] documented the code


omoerbeek:
Old Boost is a hurlde.

rgacogne:
Do we know which version of Boost fixed C++17 compatibility? Would it make sense to try C++14? I see some errors (in the fuzzers build) related to the removal of `auto_ptr` in C++17, which is still there in C++14.

omoerbeek:
> Do we know which version of Boost fixed C++17 compatibility? Would it make sense to try C++14? I see some errors (in the fuzzers build) related to the removal of `auto_ptr` in C++17, which is still there in C++14.

I searched for some explicitl statement about C++17 compatibility, but found none.

I am trying CentOS7 now, which has boost 1.53 but seems to have patched `get_pointer.hpp`. Lets see how much other stufff I encounter. For Ubuntu 16, (as used by the fuzzer) which has boost 1.57 another workaround might be needed.  If it all does not work out, I'll try C++14

rgacogne:
> For Ubuntu 16, (as used by the fuzzer) which has boost 1.57 another workaround might be needed.

OSS-Fuzz intends to move to 20.04 eventually (https://github.com/google/oss-fuzz/issues/3756) but I wouldn't hold my breath.

omoerbeek:
So Unbuntu 16 as used by the fuzzer is a real problem. I was trying to set up a debug env like the fuzzer uses and found out: while the fuzzer build environment uses clang 12, the newest available clang from ubuntu xenial is 8. And if we use that we hit an issue with the c++ libs not being c++17 level. So getting a debug environment up and running is a hassle. Is there a more direct way to get the same env as the fuzzer uses?

Habbie:
> Is there a more direct way to get the same env as the fuzzer uses?

https://github.com/google/oss-fuzz/blob/master/infra/base-images/base-builder/Dockerfile ?

rgacogne:
https://google.github.io/oss-fuzz/getting-started/new-project-guide/#testing-locally

omoerbeek:
Ubunitu xenial is a swamp. It's boost 1.57 is using too much deprecated C++11 stuff that has been removed in C++17. Now going for C++14.

omoerbeek:
Back to c++17 now that the fuzzer issue should be fixed

omoerbeek:
Both CI and buildbot runs are fine. I needed to add one workaround in the bindbackend code.

builder-support/dockerfiles/Dockerfile.rpmbuild
```diff
@@ -21,20 +21,34 @@ ADD builder-support/specs/ /pdns/builder-support/specs
...
 
 @IF [ -n "$M_dnsdist$M_all" ]
 RUN touch /var/lib/rpm/* && mkdir /libh2o && cd /libh2o && \
       yum install -y curl openssl-devel cmake && \
       curl -L https://github.com/h2o/h2o/archive/v2.2.6.tar.gz | tar xz && \
       CFLAGS='-fPIC' cmake -DWITH_PICOTLS=off -DWITH_BUNDLED_SSL=off -DWITH_MRUBY=off -DCMAKE_INSTALL_PREFIX=/opt ./h2o-2.2.6 && \
-      make install
-RUN touch /var/lib/rpm/* && builder/helpers/build-specs.sh builder-support/specs/dnsdist.spec
+      make install && \
+      cd /pdns;
```
Habbie, line 45: this line does nothing

m4/ax_cxx_compile_stdcxx_14.m4
```diff
@@ -0,0 +1,34 @@
+# =============================================================================
```
Habbie, line 1: do we need this file?

Habbie (changes requested):
Two small things, looks good otherwise!

modules/bindbackend/bindbackend2.cc
```diff
@@ -1349,13 +1349,16 @@ bool Bind2Backend::superMasterBackend(const string &ip, const DNSName& domain, c
...
 
+
 BB2DomainInfo Bind2Backend::createDomainEntry(const DNSName& domain, const string &filename)
 {
   int newid=1;
   {   // Find a free zone id nr.  
     ReadLock rl(&s_state_lock);
     if (!s_state.empty()) {
-      newid = s_state.rbegin()->d_id+1;
+      // older (1.53) versions of boost have a exprression for s_state.rbegin()
```
rgacogne, line 1359:
```suggestion
      // older (1.53) versions of boost have an expression for s_state.rbegin()
```

rgacogne (approved):

omoerbeek:
Processed review comments and squashed it.

modules/bindbackend/bindbackend2.cc
```diff
@@ -1349,13 +1349,16 @@ bool Bind2Backend::superMasterBackend(const string &ip, const DNSName& domain, c
...
 
+
 BB2DomainInfo Bind2Backend::createDomainEntry(const DNSName& domain, const string &filename)
 {
   int newid=1;
   {   // Find a free zone id nr.  
     ReadLock rl(&s_state_lock);
     if (!s_state.empty()) {
-      newid = s_state.rbegin()->d_id+1;
+      // older (1.53) versions of boost have an exprression for s_state.rbegin()
```
Habbie, line 1359:
```suggestion
      // older (1.53) versions of boost have an expression for s_state.rbegin()
```

PR #9921, opened 2021-01-07
dnsdist: Fix SNI on resumed sessions by acknowledging the name sent by the client

rgacogne:
### Short description

Otherwise `SSL_get_servername()` only returns true when the session has been freshly established, and will return `nullptr` when it is resumed.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/pdns/dnsdistdist/libssl.cc
+++ b/pdns/dnsdistdist/libssl.cc
@@ -161,6 +161,18 @@ int libssl_ticket_key_callback(SSL *s, OpenSSLTLSTicketKeysRing& keyring, unsign
   return 1;
 }
 
+static long libssl_server_name_callback(SSL* ssl, int* al, void* arg)
+{
+  (void) al;
+  (void) arg;
+
+  if (SSL_get_servername(ssl, TLSEXT_NAMETYPE_host_name)) {
+    return SSL_TLSEXT_ERR_OK;
+  }
+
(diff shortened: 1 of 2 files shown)


pdns/dnsdistdist/libssl.cc
```diff
@@ -685,6 +697,11 @@ std::unique_ptr<SSL_CTX, void(*)(SSL_CTX*)> libssl_init_server_context(const TLS
     SSL_CTX_sess_set_cache_size(ctx.get(), config.d_maxStoredSessions);
   }
 
+  /* we need to set this callback to acknowledge the server name sent by the client,
+     otherwise it will not stored in the session and will not be accessible when the
+     session is resumed, causing SSL_get_servername to return nullptr */
+  SSL_CTX_set_tlsext_servername_callback(ctx.get(), &libssl_server_name_callback);
```
omoerbeek, line 703: extremely pedantic nit: redundant `&`

rgacogne: I personally prefer it that way to make it more clear that we pass the address of the function, and not the result of a call to the function.

omoerbeek: To my eyes it's clear it isn't a call since no arguments are passed. But fair enough.

rgacogne:
Regression tests added.

Issue #9903, opened 2020-12-28
auth/lmdb: on first run after 4.4 upgrade, new zones are considered "empty"


peterthomassen:
- Program: Authoritative
 - Issue type: Bug report

### Short description
I observed that on the first run after upgrading pdns auth with lmdb backend to version 4.4.0, new zones created after the upgrade were considered empty and continually triggered AXFRs. **But,** zones actually were not empty: `pdnsutil show-zone / list-zone` showed the expected contents, and queries were answered properly.

As far as I can tell, the problem went away by shutting down and restarting auth.

### Environment
 - Operating system: Ubuntu 18.04
 - Software version: 4.4.0
 - Software source: PowerDNS repository

### Steps to reproduce
I'm not sure whether the issue can be reproduced reliably. I will comment here if I encounter the problem on other instances as well.

The configuration is here: https://github.com/desec-io/desec-frontend-node/blob/ec1a713dca87f79591e03b072a76dfb5bf76a5f9/ns/conf/pdns.conf.var
Before the upgrade, there was an additional `soa-minimum-ttl=60`, which we removed (it is now unsupported).

### Expected behaviour
Zone emptiness status should be detected correctly.

### Actual behaviour
Zones that are not empty and otherwise function correctly are claimed empty, so that thousands or AXFRs per day are triggered per affected zone. This happened more frequently than once per minute, although `slave-cycle-interval` is not configured. This caused hundreds of thousands of AXFRs per day, leading to significant extra traffic.

### Other information
Opened ticket according to discussion on IRC.

@nils-wisiol

Habbie:
> I'm not sure whether the issue can be reproduced reliably.

I can reproduce it semi-reliably, I get an 'empty' every 2 or 3 startups.

Habbie:
This happens if the slavecommunicator thread is -not- the first one to get to the LMDB file to upgrade it.

Habbie:
When you add a domain, do you use pdnsutil, or the API in the running affected process?

Habbie:
Also, if it ever happens again, can you share `lsof` output for the affected process?

peterthomassen:
> When you add a domain, do you use pdnsutil, or the API in the running affected process?

API.

We are going to upgrade a few more nodes this month. Shall post `lsof` if the issue appears.

Habbie:
I filed #9933 which fixes the similar-but-different issue I've reproduced.

Habbie:
[...]

PR #9933, opened 2021-01-11
auth lmdb: serialise LMDBBackend construction

Habbie:
### Short description
Related to, but possibly not fixing, #9903.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/modules/lmdbbackend/lmdbbackend.cc
+++ b/modules/lmdbbackend/lmdbbackend.cc
@@ -32,6 +32,7 @@
 #include "pdns/logger.hh"
 #include "pdns/version.hh"
 #include "pdns/arguments.hh"
+#include "pdns/lock.hh"
 #include <boost/archive/binary_oarchive.hpp>
 #include <boost/archive/binary_iarchive.hpp>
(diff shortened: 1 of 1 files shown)


modules/lmdbbackend/lmdbbackend.cc
```diff
@@ -53,8 +54,14 @@
 // List the class version here. Default is 0
 BOOST_CLASS_VERSION(LMDBBackend::KeyDataDB, 1)
 
+static ReadWriteLock s_lmdbOpenUpgradeLock;
+
 LMDBBackend::LMDBBackend(const std::string& suffix)
 {
+  // we lock to avoid a race condition when we do a schema upgrade during startup
+  // non-upgrade startups should be very cheap so this lock should not hurt performance
+  WriteLock l(&s_lmdbOpenUpgradeLock);
```
rgacogne, line 63: Just a nit but if we don't have a lot of concurrent readers a simple `std::mutex` would be faster :)

Habbie: ah thanks!

PR #9950, opened 2021-01-12
dnsdist: Improve error with a hint on how to fix it

pieterlexis:
### Short description
The warning emitted did not give guidance on how to fix the issue.

### Checklist

I have:
- [x] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdist-tcp.cc
+++ b/pdns/dnsdist-tcp.cc
@@ -222,7 +222,7 @@ void TCPClientCollection::addTCPClientThread()
     std::lock_guard<std::mutex> lock(d_mutex);
 
     if (d_numthreads >= d_tcpclientthreads.size()) {
-      vinfolog("Adding a new TCP client thread would exceed the vector size (%d/%d), skipping", d_numthreads.load(), d_tcpclientthreads.size());
+      vinfolog("Adding a new TCP client thread would exceed the vector size (%d/%d), skipping. Consider increasing the maximum amount of TCP client threads with setMaxTCPClientThreads() in the configuration.", d_numthreads.load(), d_tcpclientthreads.size());
       if (!d_useSinglePipe) {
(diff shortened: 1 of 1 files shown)


Habbie (approved):

rgacogne:
I don't understand right away how we could reach that point since both `tcpAcceptorThread` and `healthChecksThread` should be checking that we still have some room left before calling `TCPClientCollection::addTCPClientThread`.
That reminds me that now that one worker can handle tens of thousands of connections, and thus the number of workers should be relatively low, we might want to create all of them at startup instead of dynamically adding them later.

pieterlexis:
> I don't understand right away how we could reach that point since both `tcpAcceptorThread` and `healthChecksThread` should be checking that we still have some room left before calling `TCPClientCollection::addTCPClientThread`.
> That reminds me that now that one worker can handle tens of thousands of connections, and thus the number of workers should be relatively low, we might want to create all of them at startup instead of dynamically adding them later.

Want me to turn this PR into a ticket?

rgacogne:
Got it, there is a nice race when we create the TCP acceptor threads in a tight loop at startup, which is why we call `addTCPClientThread()` in the first place. Then we have a lock preventing the race. So this PR actually makes sense :) :+1:

rgacogne (approved):

Habbie:
Please do not restart jobs, I'm going to ask CircleCI why two checks did not report.

rgacogne:
> Please do not restart jobs, I'm going to ask CircleCI why two checks did not report.

Is that still relevant? :)

Also this PR is now conflicted, sorry about that!

Habbie:
> Is that still relevant? :)

No, sorry for not reporting back, please carry on!

pieterlexis:
rebased

PR #9957, opened 2021-01-12
dnsdist: Start all TCP worker threads on startup

rgacogne:
### Short description

Instead of starting only as many TCP worker threads on startup as the number of TCP listeners, then starting more at runtime, start all TCP worker threads on startup.
Change the default number of worker threads to at least 10, up to the number of TCP listener threads.
Document that change and improve the tuning page a bit.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdist-tcp.cc
+++ b/pdns/dnsdist-tcp.cc
@@ -221,8 +221,8 @@ void TCPClientCollection::addTCPClientThread()
   {
     std::lock_guard<std::mutex> lock(d_mutex);
 
-    if (d_numthreads >= d_tcpclientthreads.capacity()) {
-      warnlog("Adding a new TCP client thread would exceed the vector capacity (%d/%d), skipping", d_numthreads.load(), d_tcpclientthreads.capacity());
+    if (d_numthreads >= d_tcpclientthreads.size()) {
+      vinfolog("Adding a new TCP client thread would exceed the vector capacity (%d/%d), skipping", d_numthreads.load(), d_tcpclientthreads.size());
       if (!d_useSinglePipe) {
         close(pipefds[0]);
         close(pipefds[1]);
@@ -244,7 +244,7 @@ void TCPClientCollection::addTCPClientThread()
       return;
     }
 
-    d_tcpclientthreads.push_back(pipefds[1]);
+    d_tcpclientthreads.at(d_numthreads) = pipefds[1];
     ++d_numthreads;
   }
 }
@@ -1023,10 +1023,6 @@ void tcpAcceptorThread(ClientState* cs)
   bool tcpClientCountIncremented = false;
   ComboAddress remote;
   remote.sin4.sin_family = cs->local.sin4.sin_family;
-
-  if(!g_tcpclientthreads->hasReachedMaxThreads()) {
-    g_tcpclientthreads->addTCPClientThread();
-  }
 
   auto acl = g_ACL.getLocal();
   for(;;) {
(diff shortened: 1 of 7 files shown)


pdns/dnsdistdist/docs/advanced/tuning.rst
```diff
@@ -58,3 +63,18 @@ The UDP threads handling the responses from the backends do not use a lot of CPU
   newServer({address="192.0.2.127:53", name="Backend2"})
   newServer({address="192.0.2.127:53", name="Backend3"})
   newServer({address="192.0.2.127:53", name="Backend4"})
+
+For DNS over HTTPS, every :func:`addDOHLocal` directive adds a new thread dealing with incoming connections, so it might be useful to add more than one directive, as indicated above.
+
+When dealing with a large traffic load, it might happen that the internal pipe used to pass queries between the threads handling the incoming connections and the one getting a response from the backend become full too quickly, degrading performance and causing timeouts. This can be prevented by increasing the size of the internal pipe buffer, via the `internalPipeBufferSize` option of :func:`addDOHLocal`. Setting a value of `1048576` is known to yield good results on Linux.
+
+When dispatching UDP queries to backend servers, dnsdist keeps track of at most **n** outstanding queries for each backend.
+This number **n** can be tuned by the :func:`setMaxUDPOutstanding` directive, defaulting to 10240 (65535 since 1.4.0), with a maximum value of 65535.
```
pieterlexis, line 72: I would say "defaulting to 65535 (10240 before 1.4.0)" or even

```rst
defaulting to 65535, which is the maximum value.

.. versionchanged:: 1.4.0
  The default was 10240 before 1.4.0
```

rgacogne: Applied, thanks!

pdns/dnsdist-tcp.cc
```diff
@@ -221,8 +221,8 @@ void TCPClientCollection::addTCPClientThread()
   {
     std::lock_guard<std::mutex> lock(d_mutex);
 
-    if (d_numthreads >= d_tcpclientthreads.capacity()) {
-      warnlog("Adding a new TCP client thread would exceed the vector capacity (%d/%d), skipping", d_numthreads.load(), d_tcpclientthreads.capacity());
+    if (d_numthreads >= d_tcpclientthreads.size()) {
+      vinfolog("Adding a new TCP client thread would exceed the vector capacity (%d/%d), skipping", d_numthreads.load(), d_tcpclientthreads.size());
```
omoerbeek, line 225:
```suggestion
      vinfolog("Adding a new TCP client thread would exceed the vector size (%d/%d), skipping", d_numthreads.load(), d_tcpclientthreads.size());
```

pdns/dnsdist.cc
```diff
@@ -1801,14 +1797,18 @@ static void checkFileDescriptorsLimits(size_t udpBindsCount, size_t tcpBindsCoun
...
   /* listening sockets */
   requiredFDsCount += udpBindsCount;
   requiredFDsCount += tcpBindsCount;
-  /* max TCP connections currently served */
-  requiredFDsCount += g_maxTCPClientThreads;
-  /* max pipes for communicating between TCP acceptors and client threads */
-  requiredFDsCount += (g_maxTCPClientThreads * 2);
+  /* number of TCP connections currently served, assuming 1 connection per worker thread which is of course not right */
+  if (g_maxTCPClientThreads) {
+    requiredFDsCount += *g_maxTCPClientThreads;
```
omoerbeek, line 1808: This looks weird... why adding three times `*g_maxTCPClientThreads` ?

rgacogne: It should always be set at this point, I just check it to prevent a crash if I'm wrong :)

PR #9968, opened 2021-01-13
rec: As found by coverity, origTTL can be left uninitialized if all the entries found are expired.

omoerbeek:
In that case the loop exits without setting origTTL via handleHit().

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/recursor_cache.cc
+++ b/pdns/recursor_cache.cc
@@ -318,7 +318,7 @@ int32_t MemRecursorCache::get(time_t now, const DNSName &qname, const QType& qt,
     else {
       auto entry = getEntryUsingECSIndex(map, now, qname, qtype, requireAuth, who);
       if (entry != map.d_map.end()) {
-          int32_t ret = handleHit(map, entry, qname, origTTL, res, signatures, authorityRecs, variable, cachedState, wasAuth, fromAuthZone);
+        int32_t ret = handleHit(map, entry, qname, origTTL, res, signatures, authorityRecs, variable, cachedState, wasAuth, fromAuthZone);
         if (state && cachedState) {
           *state = *cachedState;
         }
@@ -336,6 +336,7 @@ int32_t MemRecursorCache::get(time_t now, const DNSName &qname, const QType& qt,
       for (auto i=entries.first; i != entries.second; ++i) {
 
         firstIndexIterator = map.d_map.project<OrderedTag>(i);
+        origTTL = firstIndexIterator->d_orig_ttl;
         if (i->d_ttd <= now) {
           moveCacheItemToFront<SequencedTag>(map.d_map, firstIndexIterator);
           continue;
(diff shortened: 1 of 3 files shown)


pdns/recursor_cache.cc
```diff
@@ -336,6 +336,7 @@ int32_t MemRecursorCache::get(time_t now, const DNSName &qname, const QType& qt,
       for (auto i=entries.first; i != entries.second; ++i) {
 
         firstIndexIterator = map.d_map.project<OrderedTag>(i);
+        origTTL = firstIndexIterator->d_orig_ttl;
```
rgacogne, line 339: So we are setting `origTTL` even though that entry might not match, that sounds a bit weird. It might not matter since we don't set `ttd`, so it will remain at `0` and thus it looks like `origTTL` will not be used. When were we reading the uninitialized value of `origTTL` exactly?

Habbie: in line 325 and 386 i believe

omoerbeek: In `fakeTTD()` we use origTTL, it returns `ttd-now` normally as before, but it uses `origTTL` to do the refetch logic.
Alternatively we could return -1 if the loop did not find a match by using g a `found` bool. It would break two unit tests that expect `-now` and then would see -1. Also, expired entries would never be submitted for refetch but that might be a good thing since they will be refetched anyway by the regular resolve.

rgacogne: But do we really use `origTTL` if `ttd` is still `0`? Or am I missing a way for `ttd` to be set while `origTTL` is not?

omoerbeek: `ttd` follows the same logic, but is initialized to 0 (hence the `-now` value the unit test sees.

rgacogne: So what I don't understand is, we only use `origTTL` in `fakeTTD`, and even there only when `ttl > 0` which implies that `ret` (coming from `ttd` in the caller) is not zero (or `now` is negative but let's not go there). When `ttd` is zero, which is its default value, it doesn't look like we actually read the uninitialized value of `origTTL`? And if `ttd` is non-zero, I think `origTTL` has been set as well?

omoerbeek: That looks right. So it does look like a false positive after all. The `ttl > 0` condition saves the day
remains the question if we want to make the code a bit easier to follow by using a `found` bool and only call fakeTTD if that is true. Note that the same logic also applies to `*state` which might be set even if `cachedState` is not set in the loop.

omoerbeek: I pushed the alternative approach to make it explitit what I'm talking about.

pdns/recursor_cache.cc
```diff
@@ -344,27 +345,32 @@ int32_t MemRecursorCache::get(time_t now, const DNSName &qname, const QType& qt,
...
           break;
         }
       }
-      if (state && cachedState) {
-        *state = *cachedState;
+      if (found) {
+        if (state && cachedState) {
+          *state = *cachedState;
+        }
+        return fakeTTD(firstIndexIterator, qname, qtype, ttd, now, origTTL, refresh);
```
rgacogne, line 359: So that means we will do a second lookup without the tag, even though we did find an (expired, or possibly not auth, ...) entry for that (qname, qtype, routingTag)?

omoerbeek: Yes, that is indeed a change in behaviour... looking into that

rgacogne (approved):
Looks good to me, it makes the code a bit easier to read. I think we should not rely on knowing whether an entry is present but expired or simply not there from outside the cache anyway.

PR #9969, opened 2021-01-13
auth: Document resolver setting

zeha:
### Short description
`resolver=` was not documented, and references to it were shown as "setting-resolver" instead.

### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -1277,7 +1277,7 @@ Number of receiver (listening) threads to start. See :doc:`performance`.
 ``resolver``
 ------------
 
--  IP Address with port
+-  IP Address with optional port
 -  Default: unset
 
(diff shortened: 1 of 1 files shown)


docs/settings.rst
```diff
@@ -1272,6 +1272,21 @@ Maximum number of milliseconds to queue a query. See :doc:`performance`.
 
 Number of receiver (listening) threads to start. See :doc:`performance`.
 
+.. _setting-resolver:
+
+``resolver``
+------------
+
+-  IP Address with port
```
mnordhoff, line 1280:
```suggestion
-  IP Address with optional port
```
It says "with port" and then one of the two examples is without a port, which is a bit confusing.

zeha: thanks, applied

Habbie (approved):

PR #9970, opened 2021-01-13
rec: Account for the NSEC(3) denial TTL in expanded wildcard answers

### LINKED ISSUE #9967: signed wildcard expansions might get cached for too long
In DNSSEC, a wildcarded name (say, `*.example.com`) is served as two things together:
* the wildcard, expanded (`hello.example.com`), with the original signature over `*.example.com`
* a NSEC(3) proof showing that `hello.example.com` does not exist and thus the wildcard expansion is legit

All of these records have TTLs, and so do their RRSIGs. The proof, and thus the expansion, is only valid for the lowest of all these TTLs (the TTLs themselves also being capped by signature expiry).

It appears that our Recursor, however, caches wildcards (plus NSEC(3) proofs) for the TTL of the expanded wildcard. If somebody validates downstream from us, we might serve them an expanded wildcard with expired proofs.

Thus, we should limit the caching of expanded wildcards to the lowest TTL involved in the expansion and the proof.

Note from @rgacogne: 

> we should likely take authorityRecs into account here: https://github.com/PowerDNS/pdns/blob/master/pdns/syncres.cc#L3104

rgacogne:
### Short description

We need to take into account the TTL of NSEC(3) proofs in expanded wildcard answers, as well as the expiration time of their signatures.

Fixes #9967.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/pdns/recursordist/test-syncres_cc5.cc
+++ b/pdns/recursordist/test-syncres_cc5.cc
@@ -788,10 +788,10 @@ BOOST_AUTO_TEST_CASE(test_dnssec_validation_nsec_wildcard)
           }
         }
         else {
-          addRecordToLW(res, domain, QType::A, "192.0.2.42");
+          addRecordToLW(res, domain, QType::A, "192.0.2.42", DNSResourceRecord::ANSWER, 600);
           addRRSIG(keys, res->d_records, DNSName("powerdns.com"), 300, false, boost::none, DNSName("*.powerdns.com"));
           /* we need to add the proof that this name does not exist, so the wildcard may apply */
-          addNSECRecordToLW(DNSName("a.powerdns.com."), DNSName("wwz.powerdns.com."), {QType::A, QType::NSEC, QType::RRSIG}, 600, res->d_records);
+          addNSECRecordToLW(DNSName("a.powerdns.com."), DNSName("wwz.powerdns.com."), {QType::A, QType::NSEC, QType::RRSIG}, 60, res->d_records);
           addRRSIG(keys, res->d_records, DNSName("powerdns.com"), 300);
         }
         return LWResult::Result::Success;
@@ -814,6 +814,10 @@ BOOST_AUTO_TEST_CASE(test_dnssec_validation_nsec_wildcard)
   BOOST_CHECK_EQUAL(res, RCode::NoError);
   BOOST_CHECK_EQUAL(sr->getValidationState(), vState::Secure);
   BOOST_REQUIRE_EQUAL(ret.size(), 4U);
+  for (const auto& rec : ret) {
+    /* check that we applied the lowest TTL, here this is from the NSEC proving that the exact name did not exist */
+    BOOST_CHECK_LE(rec.d_ttl, 60U);
+  }
   BOOST_CHECK_EQUAL(queriesCount, 9U);
 }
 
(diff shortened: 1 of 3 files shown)


pdns/syncres.cc
```diff
@@ -2327,9 +2328,32 @@ uint32_t SyncRes::computeLowestTTD(const std::vector<DNSRecord>& records, const
 
     for(const auto& sig : signatures) {
       if (isRRSIGNotExpired(d_now.tv_sec, sig)) {
-        // we don't decerement d_sigexpire by 'now' because we actually want a TTD, not a TTL */
+        // we don't decrement d_sigexpire by 'now' because we actually want a TTD, not a TTL */
         lowestTTD = min(lowestTTD, static_cast<uint32_t>(sig->d_sigexpire));
       }
+      else {
+        /* if the signature IS expired, we should not cache this */
+        lowestTTD = d_now.tv_sec;
```
rgacogne, line 2336: This bit might be controversial, especially for backports. I think I should remove it, we already cap the TTL of Bogus records anyway.

pdns/recursordist/test-syncres_cc5.cc
```diff
@@ -814,6 +814,10 @@ BOOST_AUTO_TEST_CASE(test_dnssec_validation_nsec_wildcard)
   BOOST_CHECK_EQUAL(res, RCode::NoError);
   BOOST_CHECK_EQUAL(sr->getValidationState(), vState::Secure);
   BOOST_REQUIRE_EQUAL(ret.size(), 4U);
+  for (const auto& rec : ret) {
+    /* check that we applied the lowest TTL, here this is from the NSEC proving that the exact name did not exist */
+    BOOST_CHECK_LE(rec.d_ttl, 60);
```
omoerbeek, line 819:
```suggestion
    BOOST_CHECK_LE(rec.d_ttl, 60U);
```

pdns/recursordist/test-syncres_cc5.cc
```diff
@@ -932,6 +936,10 @@ BOOST_AUTO_TEST_CASE(test_dnssec_validation_nsec_wildcard_proof_before_rrsig)
...
```
omoerbeek, line 941:
```suggestion
    BOOST_CHECK_LE(rec.d_ttl, 60U);
```

pdns/recursordist/test-syncres_cc5.cc
```diff
@@ -1408,6 +1416,10 @@ BOOST_AUTO_TEST_CASE(test_dnssec_validation_nsec3_wildcard)
   BOOST_CHECK_EQUAL(res, RCode::NoError);
   BOOST_CHECK_EQUAL(sr->getValidationState(), vState::Secure);
   BOOST_REQUIRE_EQUAL(ret.size(), 6U);
+  for (const auto& rec : ret) {
+    /* check that we applied the lowest TTL, here this is from the NSEC3 proving that the exact name did not exist (next closer) */
+    BOOST_CHECK_LE(rec.d_ttl, 60);
```
omoerbeek, line 1421:
```suggestion
    BOOST_CHECK_LE(rec.d_ttl, 60U);
```

omoerbeek (approved):
Three nits

rgacogne:
> Three nits

Sorry, Otto, I always mess that up :-/

PR #9972, opened 2021-01-13
dnsdist: Deprecate parameters to webserver(), add 'statsRequireAuthentication' parameter

### LINKED ISSUE #8710: dnsdist: please allow retrieving /metrics without authorization
I'd like to request a way to serve Prometheus /metrics without authorization, which I haven't been able to accomplish (apparently due to the way all webServer access is hardcoded to require BasicAuth).
The reason is purely due to the way Prometheus jobs are configured. While this works fine for a single target (setting credentials in the Prometheus job config), it only sort-of works when jobs are dynamically discovered via any of the service discovery mechanisms (requiring all discovered targets to have the same credentials).
Most exporters do not require any auth since metrics are generally not considered confidential or security-sensitive.

### LINKED ISSUE #9311: [docs] dnsdist webserver parameters are global
- Program: dnsdist
 - Issue type: Bug report

### Short description
webserver parameters (password, apikey, custom headers, acl) are global to all webserver instances.
This behaviour is not expected when reading the documentation https://dnsdist.org/reference/config.html#webserver-configuration. It would be useful if the documentation could state this behaviour.

### Environment
 - Operating system: CentOS 7.6
 - Software version: dnsdist 1.5.0-rc4
 - Software source: compiled myself

### Steps to reproduce
The following configuration sets different ACLs per webserver instance. However, as the parameters are globally applied "::/0" is cleared.
```
webserver("[2001:620:0:1006:21a:4aff:fede:ee3d]:8853","supersecret", "", {}, "::/0")
webserver("[::1]:8853","supersecret", "", {}, "127.0.0.1, ::1")
```

### Expected behaviour
Either apply webserver parameters per webserver instance or state in the documentation that parameters (password, apikey, custom headers, acl) are global to all webserver instances. For example:

> All webserver parameters except `listen_address` are global to all webserver instances.

rgacogne:
### Short description

This PR deprecates the use of additional parameters with `webserver()`, as the syntax is confusing and could lead to believe that the parameters are per-instance while they actually are global.
Also implements an additional 'statsRequireAuthentication' parameter to allow scraping the statistics without any kind of authentication, which is useful to Prometheus setups with dynamic service discovery.

Closes #8710.
Closes #9311.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


pdns/dnsdist-lua.cc
```diff
@@ -855,15 +855,28 @@ static void setupLuaConfig(LuaContext& luaCtx, bool client, bool configCheck)
         return;
       }
 
+      if (password || apiKey || customHeaders || acl) {
+	warnlog("Passing additional parameters to 'webserver()' is deprecated, please use 'setWebserverConfig()' instead.");
```
pieterlexis, line 859: nit: needs an indent

rgacogne: As it turns out we still have some stranded tabs here and there, fixed.

PR #9990, opened 2021-01-19
Rec: packetcache for tcp queries

omoerbeek:
### Short description

Up until now the packetcache was not used for incoming queries over tcp.

Includes re-factoring of the cache-hit path to avoid code duplication.

Note that udp and tcp entries are separated, so we do not have to deal with deciding if a udp packet would fit a tcp query or vice versa.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [X] documented the code

### DIFF
--- a/pdns/pdns_recursor.cc
+++ b/pdns/pdns_recursor.cc
@@ -860,40 +860,40 @@ static void finishTCPReply(std::unique_ptr<DNSComboWriter>& dc, bool hadError, b
   if (hadError) {
     terminateTCPConnection(dc->d_socket);
     dc->d_socket = -1;
+    return;
   }
-  else {
-    dc->d_tcpConnection->queriesCount++;
-    if (g_tcpMaxQueriesPerConn && dc->d_tcpConnection->queriesCount >= g_tcpMaxQueriesPerConn) {
-      try {
-        t_fdm->removeReadFD(dc->d_socket);
-      }
-      catch (FDMultiplexerException &) {
-      }
-      dc->d_socket = -1;
+  dc->d_tcpConnection->queriesCount++;
+  if (g_tcpMaxQueriesPerConn && dc->d_tcpConnection->queriesCount >= g_tcpMaxQueriesPerConn) {
+    try {
+      t_fdm->removeReadFD(dc->d_socket);
     }
-    else {
-      Utility::gettimeofday(&g_now, 0); // needs to be updated
-      struct timeval ttd = g_now;
-      // If we cross from max to max-1 in flight requests, the fd was not listened to, add it back
-      if (updateInFlight && dc->d_tcpConnection->d_requestsInFlight == TCPConnection::s_maxInFlight - 1) {
-        // A read error might have happened. If we add the fd back, it will most likely error again.
-        // This is not a big issue, the next handleTCPClientReadable() will see another read error
-        // and take action.
-        ttd.tv_sec += g_tcpTimeout;
-        t_fdm->addReadFD(dc->d_socket, handleRunningTCPQuestion, dc->d_tcpConnection, &ttd);
-      } else {
-        // fd might have been removed by read error code, or a read timeout, so expect an exception
-        try {
-          t_fdm->setReadTTD(dc->d_socket, ttd, g_tcpTimeout);
-        }
-        catch (const FDMultiplexerException &) {
-          // but if the FD was removed because of a timeout while we were sending a response,
-          // we need to re-arm it. If it was an error it will error again.
-          ttd.tv_sec += g_tcpTimeout;
-          t_fdm->addReadFD(dc->d_socket, handleRunningTCPQuestion, dc->d_tcpConnection, &ttd);
-        }
-      }
+    catch (FDMultiplexerException &) {
     }
+    dc->d_socket = -1;
+    return;
+  }
+
+  Utility::gettimeofday(&g_now, nullptr); // needs to be updated
+  struct timeval ttd = g_now;
+
+  // If we cross from max to max-1 in flight requests, the fd was not listened to, add it back
+  if (updateInFlight && dc->d_tcpConnection->d_requestsInFlight == TCPConnection::s_maxInFlight - 1) {
+    // A read error might have happened. If we add the fd back, it will most likely error again.
+    // This is not a big issue, the next handleTCPClientReadable() will see another read error
+    // and take action.
+    ttd.tv_sec += g_tcpTimeout;
+    t_fdm->addReadFD(dc->d_socket, handleRunningTCPQuestion, dc->d_tcpConnection, &ttd);
+    return;
+  }
+  // fd might have been removed by read error code, or a read timeout, so expect an exception
+  try {
+    t_fdm->setReadTTD(dc->d_socket, ttd, g_tcpTimeout);
+  }
+  catch (const FDMultiplexerException &) {
+    // but if the FD was removed because of a timeout while we were sending a response,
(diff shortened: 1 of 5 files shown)


pdns/pdns_recursor.cc
```diff
@@ -918,6 +920,54 @@ static void protobufLogResponse(pdns::ProtoZero::RecMessage& message)
...
+
+  // In response part
+  if (g_useKernelTimestamp && tv.tv_sec) {
+    pbMessage.setQueryTime(tv.tv_sec, tv.tv_usec);
+  }
+  else {
+    pbMessage.setQueryTime(g_now.tv_sec, g_now.tv_usec);
+  }
+
+  auto luaconfsLocal = g_luaconfs.getLocal();
```
rgacogne, line 946: Perhaps we could pass the local state holder to the function? `GlobalStateHolder::getLocal` is cheap but the next dereference of `luaconfsLocal` will call `GlobalStateHolder::getState` which is not.

pdns/pdns_recursor.cc
```diff
@@ -2566,21 +2658,80 @@ static void handleRunningTCPQuestion(int fd, FDMultiplexer::funcparam_t& var)
...
+          if (t_protobufServers && dc->d_logResponse && !(luaconfsLocal->protobufExportConfig.taggedOnly && pbData && !pbData->d_tagged)) {
+            struct timeval tv{0, 0};
+            protobufLogResponse(dh, pbData, tv, true, dc->d_source, dc->d_destination, dc->d_ednssubnet, dc->d_uuid, dc->d_requestorId, dc->d_deviceId, dc->d_deviceName);
+          }
+
+          if (!g_quiet) {
+            g_log<<Logger::Notice<<t_id<< " TCP question answered from packet cache tag="<<dc->d_tag<<" from "<<dc->d_source.toStringWithPort()<<(dc->d_source != dc->d_remote ? " (via "+dc->d_remote.toStringWithPort()+")" : "")<<endl;
+          }
+
+          bool hadError = sendResponseOverTCP(dc, response);
```
rgacogne, line 2685: I wonder if we could try to factorize that code, which seems duplicated from `startDoResolve`?

pdns/pdns_recursor.cc
```diff
@@ -2566,21 +2658,80 @@ static void handleRunningTCPQuestion(int fd, FDMultiplexer::funcparam_t& var)
...
       else {
+        // We have read a proper query
         ++g_stats.qcounter;
         ++g_stats.tcpqcounter;
-        ++conn->d_requestsInFlight;
-        if (conn->d_requestsInFlight >= TCPConnection::s_maxInFlight) {
-          t_fdm->removeReadFD(fd); // should no longer awake ourselves when there is data to read
+
+        string response;
+        uint32_t qhash = 0;
```
rgacogne, line 2666: Perhaps use `dc->d_qhash` directly?

rgacogne (approved):
Looks good, a few nits but the PR seems sound to me! :+1:

pdns/pdns_recursor.cc
```diff
@@ -918,6 +920,54 @@ static void protobufLogResponse(pdns::ProtoZero::RecMessage& message)
...
+static void protobufLogResponse(const struct dnsheader* dh,
+                                const RecursorPacketCache::OptPBData& pbData, const struct timeval& tv,
+                                bool tcp, const ComboAddress& source, const ComboAddress& destination,
+                                const EDNSSubnetOpts& ednssubnet,
+                                const boost::uuids::uuid& uniqueId, const string& requestorId, const string& deviceId,
+                                const string& deviceName) 
+{
+  pdns::ProtoZero::RecMessage pbMessage(pbData ? pbData->d_message : "", pbData ? pbData->d_response : "", 64, 10); // The extra bytes we are going to add
+  if (pbData) {
+    // We take the inmutable string from the cache and are appending a few values
```
pieterlexis, line 932: empty?

Habbie, line 932:
```suggestion
    // We take the immutable string from the cache and are appending a few values
```

omoerbeek: This is a kind of "cannot happen situation". There is a hit in the PC, but no protobuf packet form the cache. In that case we start with an empty one (instead of once having various things filled in) and add the bare minimum in the `else` branch of the `if (pbData)` block.

pieterlexis: so like this:

```cpp
if (!pbData) {
  // content of the current else block
}
```

omoerbeek: A, now I see what you mean. I was looking at the limited github view. Will fix.

pdns/pdns_recursor.cc
omoerbeek: Yes, it's similar, though a bit more simpler, lacking the `requestsInFlight` accounting since it can answer immediately. I will take a look.

omoerbeek:
Processed the review comments, including the re-factoring of the fd accouting after a TCP reply has been sent. This part needs an additional review.

pdns/pdns_recursor.cc
```diff
@@ -846,6 +846,57 @@ static void sendErrorOverTCP(std::unique_ptr<DNSComboWriter>& dc, int rcode)
...
+        }
+        catch (const FDMultiplexerException &) {
+          // but if the FD was removed because of a timeout while we were sending a response,
+          // we need to re-arm it. If it was an error it will error again.
+          ttd.tv_sec += g_tcpTimeout;
+          t_fdm->addReadFD(dc->d_socket, handleRunningTCPQuestion, dc->d_tcpConnection, &ttd);
+        }
+      }
+    }
+  }
```
pieterlexis, line 897: It might be more readable to return early instead of having a bunch of `if`s in `else` blocks:

```suggestion
  if (hadError) {
    terminateTCPConnection(dc->d_socket);
    dc->d_socket = -1;
    return;
  }

  dc->d_tcpConnection->queriesCount++;
  if (g_tcpMaxQueriesPerConn && dc->d_tcpConnection->queriesCount >= g_tcpMaxQueriesPerConn) {
    try {
      t_fdm->removeReadFD(dc->d_socket);
    }
    catch (FDMultiplexerException &) {
    }
    dc->d_socket = -1;
    return;
  }

  Utility::gettimeofday(&g_now, 0); // needs to be updated
  struct timeval ttd = g_now;
  // If we cross from max to max-1 in flight requests, the fd was not listened to, add it back
  if (updateInFlight && dc->d_tcpConnection->d_requestsInFlight == TCPConnection::s_maxInFlight - 1) {
    // A read error might have happened. If we add the fd back, it will most likely error again.
    // This is not a big issue, the next handleTCPClientReadable() will see another read error
    // and take action.
    ttd.tv_sec += g_tcpTimeout;
    t_fdm->addReadFD(dc->d_socket, handleRunningTCPQuestion, dc->d_tcpConnection, &ttd);
    return;
  }
  // fd might have been removed by read error code, or a read timeout, so expect an exception
  try {
    t_fdm->setReadTTD(dc->d_socket, ttd, g_tcpTimeout);
  }
  catch (const FDMultiplexerException &) {
    // but if the FD was removed because of a timeout while we were sending a response,
    // we need to re-arm it. If it was an error it will error again.
    ttd.tv_sec += g_tcpTimeout;
    t_fdm->addReadFD(dc->d_socket, handleRunningTCPQuestion, dc->d_tcpConnection, &ttd);
  }
```

rgacogne (approved):

PR #9995, opened 2021-01-20
Rec and sdig: support tcp fastopen connect

### LINKED ISSUE #7982: rec: Add support for outgoing TCP Fast Open with TCP_FASTOPEN_CONNECT
- Program: Recursor
 - Issue type: Feature request

### Short description
When the `TCP_FASTOPEN_CONNECT` option is available (Linux >= 4.11) and `tcp-fast-open` is set, it would be nice to enable TCP fast open toward authoritative servers, as requested in https://github.com/PowerDNS/pdns/pull/5138.

### Usecase
Support outgoing TFO to speed up TCP connection toward authoritative servers, without having to worry about not connecting the socket and using `sendto()` with the `MSG_FASTOPEN` flag, and that only for the first write.

### Description

When `TCP_FASTOPEN_CONNECT` is available and `tcp-fast-open` is set, we should try to set  this flag to speed up the connection. This should require very few changes, apart from calling `setsockopt()` with the flag, making sure that we correctly handle `EINPROGRESS` for the `connect()` and `write()`/`sendmsg()` calls.
See [1] for more information.

[1]: https://www.mail-archive.com/netdev@vger.kernel.org/msg149369.html

omoerbeek:
### Short description

Start supporting `TCP_FASTOPEN` for outgoing TCP connections.
We originally did this for `sdig` and the recursor, but the `sdig` case is now handler by tcpiohandler and no longer part  of this PR.

Don't forget to do `sysctl -w net.ipv4.tcp_fastopen=3` if you want to use this (or `1` if you only want it for outgoing and `2` for incoming only.

While testing this, I had a hard time finding auths that actually support FAST_OPEN.
8.8.8.8 and the google auths appear to do so, but trying to use it, the cookie is rejected and a new cookie is sent, causing the client to retransmit the request payload.

Fixes #7982

TODO

- [ ] deciding if we want this since auth support seems to be almost non-existent or buggy.
- [x] separate setting for outgoing, e.g. `tcp-fast-open-connect`
- [x] add check of sysctl setting so people get a warning if `tcp-fast-open` (or `tcp-fast-open-connect`?) setting isn;t allowed by sysctl
- [X] docs incl mentioning sysctl and guidance for the value of `tcp-fast-open`

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [X] documented the code

### DIFF
--- a/pdns/lwres.cc
+++ b/pdns/lwres.cc
@@ -332,7 +332,7 @@ LWResult::Result asyncresolve(const ComboAddress& ip, const DNSName& domain, int
           s.setFastOpenConnect();
         }
         catch (const NetworkError& e) {
-          g_log << Logger::Error << "tcp-fast-connect enabled but returned error: " << e.what() << endl;
+          // Ignore error, we did a pre-check in pdns_recursor.cc:checkTFOconnect()
         }
       }
 
--- a/pdns/lwres.cc
+++ b/pdns/lwres.cc
@@ -327,8 +327,7 @@ LWResult::Result asyncresolve(const ComboAddress& ip, const DNSName& domain, int
       Socket s(ip.sin4.sin_family, SOCK_STREAM);
 
       s.setNonBlocking();
-      // v6 tcp does not seem to have fast open
-      if (ip.sin4.sin_family == AF_INET && SyncRes::s_tcp_fast_open_connect) {
+      if (SyncRes::s_tcp_fast_open_connect) {
         try {
           s.setFastOpenConnect();
         }
--- a/pdns/lwres.cc
+++ b/pdns/lwres.cc
@@ -328,11 +328,14 @@ LWResult::Result asyncresolve(const ComboAddress& ip, const DNSName& domain, int
 
       s.setNonBlocking();
       ComboAddress local = pdns::getQueryLocalAddress(ip.sin4.sin_family, 0);
+      if (SyncRes::s_tcp_fast_open > 0) {
+        s.setFastOpen(SyncRes::s_tcp_fast_open);
+      }
 
       s.bind(local);
-        
+
       s.connect(ip);
-      
+
       uint16_t tlen=htons(vpacket.size());
       char *lenP=(char*)&tlen;
       const char *msgP=(const char*)&*vpacket.begin();
@@ -341,7 +344,7 @@ LWResult::Result asyncresolve(const ComboAddress& ip, const DNSName& domain, int
       if (ret != LWResult::Result::Success) {
         return ret;
       }
-      
+
       packet.clear();
       ret = arecvtcp(packet, 2, &s, false);
       if (ret != LWResult::Result::Success) {
--- a/pdns/lwres.cc
+++ b/pdns/lwres.cc
@@ -327,14 +327,20 @@ LWResult::Result asyncresolve(const ComboAddress& ip, const DNSName& domain, int
       Socket s(ip.sin4.sin_family, SOCK_STREAM);
 
       s.setNonBlocking();
-      ComboAddress local = pdns::getQueryLocalAddress(ip.sin4.sin_family, 0);
-      if (SyncRes::s_tcp_fast_open > 0) {
-        s.setFastOpenConnect();
+      // v6 tcp does not seem to have fast open
+      if (ip.sin4.sin_family == AF_INET && SyncRes::s_tcp_fast_open_connect) {
+        try {
+          s.setFastOpenConnect();
+        }
+        catch (const NetworkError& e) {
+          g_log << Logger::Error << "tcp-fast-connect enabled but returned error: " << e.what() << endl;
+        }
       }
 
+      ComboAddress local = pdns::getQueryLocalAddress(ip.sin4.sin_family, 0);
       s.bind(local);
 
-      s.connect(ip, g_networkTimeoutMsec * 1000); // needed for fastopen EINPROGRESS and better anyway than the system wide connect timeout
+      s.connect(ip);
 
       uint16_t tlen=htons(vpacket.size());
       char *lenP=(char*)&tlen;
--- a/pdns/lwres.cc
+++ b/pdns/lwres.cc
@@ -329,7 +329,7 @@ LWResult::Result asyncresolve(const ComboAddress& ip, const DNSName& domain, int
       s.setNonBlocking();
       ComboAddress local = pdns::getQueryLocalAddress(ip.sin4.sin_family, 0);
       if (SyncRes::s_tcp_fast_open > 0) {
-        s.setFastOpen(SyncRes::s_tcp_fast_open);
+        s.setFastOpenConnect();
       }
 
       s.bind(local);
--- a/pdns/lwres.cc
+++ b/pdns/lwres.cc
@@ -334,7 +334,7 @@ LWResult::Result asyncresolve(const ComboAddress& ip, const DNSName& domain, int
 
       s.bind(local);
 
-      s.connect(ip);
+      s.connect(ip, g_networkTimeoutMsec * 1000); // needed for fastopen EINPROGRESS and better anyway than the system wide connect timeout
 
       uint16_t tlen=htons(vpacket.size());
       char *lenP=(char*)&tlen;
--- a/pdns/pdns_recursor.cc
+++ b/pdns/pdns_recursor.cc
@@ -3236,11 +3236,23 @@ static void checkFastOpenSysctl(bool active)
     }
   }
   else {
-    g_log << Logger::Error << "Cannot determine if kernel setting allow fast-open" << endl;
+    g_log << Logger::Notice << "Cannot determine if kernel settings allow fast-open" << endl;
  }
 #else
-  g_log << Logger::Error << "Cannot determine if kernel setting allow fast-open" << endl;
+  g_log << Logger::Notice << "Cannot determine if kernel settings allow fast-open" << endl;
 #endif
+}
+
+static void checkTFOconnect()
+{
+  try {
+    Socket s(AF_INET, SOCK_STREAM);
+    s.setNonBlocking();
+    s.setFastOpenConnect();
+  }
+  catch (const NetworkError& e) {
+    g_log << Logger::Error << "tcp-fast-open-connect enabled but returned error: " << e.what() << endl;
+  }
 }
 
 static void makeTCPServerSockets(deferredAdd_t& deferredAdds, std::set<int>& tcpSockets)
@@ -4698,6 +4710,7 @@ static int serviceMain(int argc, char*argv[])
 
   if (SyncRes::s_tcp_fast_open_connect) {
     checkFastOpenSysctl(true);
+    checkTFOconnect();
   }
 
   if(SyncRes::s_serverID.empty()) {
--- a/pdns/pdns_recursor.cc
+++ b/pdns/pdns_recursor.cc
@@ -3292,7 +3292,7 @@ static void makeTCPServerSockets(deferredAdd_t& deferredAdds, std::set<int>& tcp
 
     if (SyncRes::s_tcp_fast_open > 0) {
 #ifdef TCP_FASTOPEN
-      if (setsockopt(fd, IPPROTO_TCP, TCP_FASTOPEN, &SyncRes::tcp_fast_open, sizeof SyncRes::tcp_fast_open) < 0) {
+      if (setsockopt(fd, IPPROTO_TCP, TCP_FASTOPEN, &SyncRes::s_tcp_fast_open, sizeof SyncRes::s_tcp_fast_open) < 0) {
         int err = errno;
         g_log<<Logger::Error<<"Failed to enable TCP Fast Open for listening socket: "<<strerror(err)<<endl;
       }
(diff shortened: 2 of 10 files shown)


omoerbeek:
https://issuetracker.google.com/68040969
indeed, fixing the src port makes the issue go away

pdns/sdig.cc
```diff
@@ -346,7 +349,11 @@ try {
   } else if (tcp) {
     uint16_t counter = 0;
     Socket sock(dest.sin4.sin_family, SOCK_STREAM);
-    sock.connect(dest);
+    sock.setNonBlocking();
+    if (fastopen) {
+      sock.setFastOpenConnect();
+    }
+    sock.connect(dest, INT_MAX);
```
rgacogne, line 356: That's one huge timeout :) It would be nice to be able to configure it, but for now perhaps we can set it to 10s, as we do for UDP?

pdns/sstuff.hh
```diff
@@ -127,6 +128,15 @@ public:
     }
   }
 
+  void setFastOpenConnect()
+  {
+#ifdef TCP_FASTOPEN_CONNECT
+    int on = 1;
+    if (setsockopt(d_socket, IPPROTO_TCP, TCP_FASTOPEN_CONNECT, &on, sizeof(on)) < 0)
+      throw NetworkError("While setting TCP_FASTOPEN_CONNECT: " + stringerror());
```
rgacogne, line 136: Do we want to fall on our sword on that? I'm pretty sure we will have some cases where `TCP_FASTOPEN_CONNECT` was defined at compile time but it not available at run-time. So it might make sense to catch that exception in the callers?

pdns/sdig.cc
```diff
@@ -355,23 +362,26 @@ try {
...
       counter++;
 
-      uint16_t len = htons(packet.size());
-      if (sock.write((const char *)&len, 2) != 2)
-        throw PDNSException("tcp write failed");
-      string question(packet.begin(), packet.end());
+      // Prefer to do a single write, so that fastopen can send all the data on SYN
+      uint16_t len = packet.size();
+      string question;
+      question.reserve(2 + packet.size());
```
rgacogne, line 368:
```suggestion
      question.reserve(sizeof(len) + packet.size());
```

rgacogne (changes requested):

pdns/lwres.cc
```diff
@@ -328,11 +328,14 @@ LWResult::Result asyncresolve(const ComboAddress& ip, const DNSName& domain, int
...
+      if (SyncRes::s_tcp_fast_open > 0) {
+        s.setFastOpenConnect();
+      }
 
       s.bind(local);
-        
-      s.connect(ip);
-      
+
+      s.connect(ip, g_networkTimeoutMsec * 1000); // needed for fastopen EINPROGRESS and better anyway than the system wide connect timeout
```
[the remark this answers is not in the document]
omoerbeek: I'll check the existing behaviour in mode detail tomorrow and make sure not to alter behaviour. I might have confused myself seeing unhandled EINPROGRESS in straces.

pdns/sstuff.hh
omoerbeek: or log and continue, as we do for the listening case.

pdns/sdig.cc
omoerbeek: I did rewrite sdig to use non-blocking mode, so the code is more equal to the rec case. This isn't necessary, but handy for study and diagnosis. But if this code survives a more decent timeout is needed.

pdns/sstuff.hh
rgacogne: Agreed, provided that we catch and log in the caller.

omoerbeek:
Some more knowledge gained by experimenting and linux src inspection:
- for `setsockopt(TCP_FASTOPEN_CONNECT)` fails if the sysctl isn't allowing it (good since it allows us to give a proper warning
- but `setsockopt(TCP_FASTOPEN)` (for passive sockets) does not, we do need to chekc the sysctk value "by hand" to give a warning 
- queries google nameservers multiple times with TCP_FASTOPEN_CONNECT triggers the `TCPFastOpenBlackhole` conditions here. This has the consequence that *all* future fast open will be (silently) disabled, even onces that previously worked. See 
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/net/ipv4/tcp_fastopen.c  `tcp_fastopen_active_disable()` and further. I do not completely understand yet why that condition is triggered here, since I'm not seeing timeouts

pieterlexis:
>     * while testing, i really need to set  `net.ipv4.tcp_fastopen_blackhole_timeout_sec=0` otherwise the feature gets disabled all the time.

It would be good to mention this both in the docs for the setting(s) and the tuning guide

pdns/recursordist/docs/performance.rst
```diff
@@ -107,6 +107,34 @@ The settings can be made permanent by using the ``--permanent`` flag::
 
 Following the instructions above, you should be able to attain very high query rates.
 
+.. _tcp-fast-open-support:
+
+TCP Fast Open Support
+---------------------
+On Linux systems, the recursor can use TCP Fast Open for passive (incoming, since 4.1) and active (outgoing, since 4.5) TCP connections.
+TCP Fast Open allows the initial SYN packet to carry data, saving one network round-trip.
+For details, consult RFC 7413.
```
pieterlexis, line 116:
```suggestion
For details, consult `:rfc:7413`.
```

omoerbeek:
An example of a NS causing trouble: ns1.yahoo.com (68.180.131.16)
test script:
```
netstat -s | grep TCPFastOpenBlackhole
echo hi | curl --connect-timeout 10 --tcp-fastopen $1:53 
netstat -s | grep TCPFastOpenBlackhole
```
```
$  ./t 68.180.131.16
    TCPFastOpenBlackhole: 60
curl: (28) Connection timed out after 10000 milliseconds
    TCPFastOpenBlackhole: 61
```

pieterlexis:
@omoerbeek can you milestone this?

omoerbeek:
Undecided, we gain little from fastopen as things are now. As I'm also moving sdig (and later rec) to tcpiohandler so it may be is better to shelf this and solve it in tcpiohandler for all outgoing TCP, be it rec, dnsdist or sdig.

omoerbeek:
This should be reworked once #10108 is merged to use tcpiohandler, which then does FastOpen as well.

omoerbeek:
Rebased. Integrating the lwrec TCP I/O with tcpiohandler is a bit more tricky, since the recursor has the special `asendtcp` and `arecvtcp` functions that tie into the mthreads. Still this is needed, also for DoT to auth support.

pdns/sdig.cc
```diff
@@ -207,6 +207,7 @@ try {
   bool dnssec = false;
   bool recurse = false;
   bool tcp = false;
+  bool fastopen = false;
```
rgacogne, line 210: Duplicate below (line 215).

pdns/lwres.cc
```diff
@@ -327,12 +327,21 @@ LWResult::Result asyncresolve(const ComboAddress& ip, const DNSName& domain, int
       Socket s(ip.sin4.sin_family, SOCK_STREAM);
 
       s.setNonBlocking();
-      ComboAddress local = pdns::getQueryLocalAddress(ip.sin4.sin_family, 0);
+      // v6 tcp does not seem to have fast open
+      if (ip.sin4.sin_family == AF_INET && SyncRes::s_tcp_fast_open_connect) {
```
rgacogne, line 331: Based on https://www.spinics.net/lists/netdev/msg416206.html I believe `TCP_FASTOPEN_CONNECT` will work on IPv6 sockets as well.

rgacogne (changes requested):

pdns/lwres.cc
omoerbeek: I'm not really sure. On my buster machines I see this:
```
# sudo sysctl -a | grep fast
net.ipv4.tcp_fastopen = 3
net.ipv4.tcp_fastopen_blackhole_timeout_sec = 3600
```
No mention of TFO for v6.

rgacogne: I'm pretty sure these settings control both the IPv4 and IPv6 sides, and lives in `net.ipv4` for historical reasons, but I have not actually verified this.

omoerbeek: I will verify

omoerbeek:
I changed the milestone to 4.5 since this is low risk: disabled by default.

pdns/lwres.cc
```diff
@@ -327,12 +327,20 @@ LWResult::Result asyncresolve(const ComboAddress& ip, const DNSName& domain, int
       Socket s(ip.sin4.sin_family, SOCK_STREAM);
 
       s.setNonBlocking();
-      ComboAddress local = pdns::getQueryLocalAddress(ip.sin4.sin_family, 0);
+      if (SyncRes::s_tcp_fast_open_connect) {
+        try {
+          s.setFastOpenConnect();
+        }
+        catch (const NetworkError& e) {
+          g_log << Logger::Error << "tcp-fast-connect enabled but returned error: " << e.what() << endl;
```
rgacogne, line 335: I'm a bit afraid of flooding the logs with that.. Perhaps we could do that test at startup by creating a socket and trying to enable Fast Open Connect? If that works we close the socket right away, if it doesn't we log _once_.

pdns/pdns_recursor.cc
```diff
@@ -3162,6 +3162,27 @@ static void handleNewUDPQuestion(int fd, FDMultiplexer::funcparam_t& var)
...
+    }
+    if (!active && !(flag & 2)) {
+      g_log << Logger::Error << "tcp-fast-open enabled but net.ipv4.tcp_fastopen does not allow it" << endl;
+    }
+  }
+  else {
+    g_log << Logger::Error << "Cannot determine if kernel setting allow fast-open" << endl;
+ }
+#else
+  g_log << Logger::Error << "Cannot determine if kernel setting allow fast-open" << endl;
```
rgacogne, line 3182: How about `Notice` instead? The administrator can't fix this anyway.

pdns/recursordist/docs/performance.rst
```diff
@@ -107,6 +107,34 @@ The settings can be made permanent by using the ``--permanent`` flag::
...
 
+.. _tcp-fast-open-support:
+
+TCP Fast Open Support
+---------------------
+On Linux systems, the recursor can use TCP Fast Open for passive (incoming, since 4.1) and active (outgoing, since 4.5) TCP connections.
+TCP Fast Open allows the initial SYN packet to carry data, saving one network round-trip.
+For details, consult `:rfc:7413`.
+
+To enable TCP Fast Open, it might be need change the value of the ``net.ipv4.tcp_fastopen`` sysctl.
```
rgacogne, line 118:
```suggestion
To enable TCP Fast Open, it might be needed to change the value of the ``net.ipv4.tcp_fastopen`` sysctl.
```

pdns/recursordist/docs/performance.rst
```diff
@@ -107,6 +107,34 @@ The settings can be made permanent by using the ``--permanent`` flag::
...
+    netstat -s | grep TCPFastOpen
+
+Please note that if active TCP Fast Open attempts fail in particular ways, the Linux kernel stops using active TCP Fast Open for a while for all connections, even connection to servers that previously worked.
+This behaviour can be monitored by watching the ``TCPFastOpenBlackHole`` kernel metric and influenced by setting the ``net.ipv4.tcp_fastopen_blackhole_timeout_sec`` sysctl.
+While developing active TCP Fast Open, it was needed to set ``net.ipv4.tcp_fastopen_blackhole_timeout_sec`` to zero to circumvent the issue, since it was triggered regularly while forcing TCP connections to nameservers for popular domains.
+
+At the moment of writing, the Google operated nameservers (both recursive and authoritative) indicate Fast Open support in the TCP handshake, but do not accept the cookie they sent previously and send a new one for each connection.
+We can only hope Google will fix this issue soon.
+
+If you operate an anycast pool of machines, make them share the TCP Fast Open Key by setting the ``net.ipv4.tcp_fastopen_key`` sysctl, otherwise you wil create a similar issue the Google servers have.
```
rgacogne, line 132:
```suggestion
If you operate an anycast pool of machines, make them share the TCP Fast Open Key by setting the ``net.ipv4.tcp_fastopen_key`` sysctl, otherwise you will create a similar issue the Google servers have.
```

pdns/recursordist/docs/settings.rst
```diff
@@ -1800,7 +1800,18 @@ A list of comma-separated statistic names, that are prevented from being exporte
...
+.. _setting-tcp-fast-open-connect:
+
+``tcp-fast-open-connect``
+-------------------------
+.. versionadded:: 4.5.0
+
+-  Boolean
+-  Default: no (disabled)
+
+Enable TCP Fast Open Connect support, if available, on the outgoing connections to authoritatively servers. See :ref:`tcp-fast-open-support`.
```
rgacogne, line 1814:
```suggestion
Enable TCP Fast Open Connect support, if available, on the outgoing connections to authoritative servers. See :ref:`tcp-fast-open-support`.
```

rgacogne (approved):
:+1:

PR #10002, opened 2021-01-22
recursor: Add EDNS0 padding

pieterlexis:
### Short description
This adds a very crude from of EDNS0 padding to the recursor.

When the client sends a PADDING option in the request, the recursor will send a padded response, with at most 512 padding bytes, limited by the buffersize of the UDP packet.

Do note that this is quite the packetcache-buster and might require some special handling.

This PR also needs tests.

~~This commit also ensures the EDNS0 options in the response are always sorted.~~

To consider:
 * [x] `edns-padding` config option: `off` (never send padded responses, `on-request` (send padded responses to queries with padding), `always` (always send a padded response to queries with EDNS0). Implemented as a single boolean now.
 * [x]  `edns-padding-max-bytes` option: a way to specify the maximum amount of padding to add (as it is fixed to 512 now)

In the future, we could consider adding several of the policies described in [RFC 8467](https://tools.ietf.org/html/rfc8467).

### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/pdns_recursor.cc
+++ b/pdns/pdns_recursor.cc
@@ -244,6 +244,9 @@ static uint16_t s_maxUdpSourcePort;
 static double s_balancingFactor;
 static bool s_addExtendedResolutionDNSErrors;
 
+static bool s_EDNS0Padding;
+static long unsigned int s_EDNS0PaddingMaxBytes;
+
 RecursorControlChannel s_rcc; // only active in the handler thread
 RecursorStats g_stats;
 string s_programname="pdns_recursor";
@@ -1930,10 +1933,10 @@ static void startDoResolve(void *p)
         }
       }
 
-      if (shouldPad && pw.size() < maxanswersize && (maxanswersize - pw.size()) >= (EDNSOptionCodeSize + EDNSOptionLengthSize)) {
+      if (s_EDNS0Padding && shouldPad && pw.size() < maxanswersize && (maxanswersize - pw.size()) >= (EDNSOptionCodeSize + EDNSOptionLengthSize)) {
         string to_add;
         // Add 512 bytes, or fewer if the packet is full
-        to_add.resize(std::min(maxanswersize - pw.size() - EDNSOptionCodeSize - EDNSOptionLengthSize, static_cast<long unsigned int>(512)));
+        to_add.resize(std::min(maxanswersize - pw.size() - EDNSOptionCodeSize - EDNSOptionLengthSize, s_EDNS0PaddingMaxBytes));
         returnedEdnsOptions.push_back(std::make_pair(EDNSOptionCode::PADDING, to_add));
       }
       /* we try to add the EDNS OPT RR even for truncated answers,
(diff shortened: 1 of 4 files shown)


omoerbeek:
I think padding should be irrelevant for the hash. Looks like  the place to handle that is `PacketCache::hashAfterQname()`.

Habbie:
> This commit also ensures the EDNS0 options in the response are always sorted.

why?

pieterlexis:
> > This commit also ensures the EDNS0 options in the response are always sorted.
> 
> why?

I just read the RFC, there is no need indeed

pieterlexis:
> Do note that this is quite the packetcache-buster and might require some special handling.

It might be hard to make padding not bust the cache:

Ignoring the value of the PADDING option is not enough, as the OPTION LENGTH can change per-request. However, ignoring both the LENGTH and the value, still makes the packet uncachable as the the total EDNS0 RDLEN still counts those bytes (and is already hashed by the time we reach the code to hash the PADDING option).

pdns/pdns_recursor.cc
```diff
@@ -1927,6 +1933,12 @@ static void startDoResolve(void *p)
         }
       }
 
+      if (s_EDNS0Padding && shouldPad && pw.size() < maxanswersize && (maxanswersize - pw.size()) >= (EDNSOptionCodeSize + EDNSOptionLengthSize)) {
+        string to_add;
+        // Add 512 bytes, or fewer if the packet is full
+        to_add.resize(std::min(maxanswersize - pw.size() - EDNSOptionCodeSize - EDNSOptionLengthSize, s_EDNS0PaddingMaxBytes));
```
omoerbeek, line 1939: This allocates a string of NULs and then appends that string to the packet (potentially causing a allocation and copying data). It would be nice if you could  compute the desired length and just extend the packet data. That might require some code in the code handling `returnedEdnsOptions` though.

pieterlexis: I have added a commit with this change, but I'm not happy with it.

omoerbeek:
Wouldn't it be a better approach to

- Completely ignore incoming padding for hash calculation, but do include the presence of the option
- Store the unpadded response in the packet cache, and append padding only if needed for the outgoing data.

pieterlexis:
>     * Completely ignore incoming padding for hash calculation, but do include the presence of the option

It would, but we already hashed RDLEN of the OPT record when we find out there is a PADDING option, so the cache is already busted (could be fixed with more code and using a second position counter).

>     * Store the unpadded response in the packet cache, and append padding only if needed for the outgoing data.

That would be good, but that would mean we'd need to modify the OPT record (RDLEN and append PADDING) from the packet cache.

It might be good to see what stubs that support PADDING actually do. If they (e.g.) always pad the same amount, or always pad _to_ a certain amount, then it could be ok to only ignore the value of the padding bytes. As the packet-cache would contain 1 answer per qname/qtype per stub implementation. Although a malicious user can still bust the packet-cache.

omoerbeek:
Always padding with the same amount of bytes does not make sense to me. Indeed, it's even mentioned in rfc 8467 Appendix A.  Padding Policies That Are Not Sensible :-)

rgacogne:
My initial thoughts on the packet-cache busting issue, as well as an implementation of the `Block-Length Padding` strategy can be found in #8918.

Issue #10000, opened 2021-01-21
Time types and computations audit


omoerbeek:
The code bases uses `time_t`, `int32_t` and `uint32_t` (and maybe more) for time related values. These values can be absolute or relative. Some of the values directly represent data from DNS records, some are used for internal computations.
A review of all the types used and computations is needed to make sure things will keep working in 2038 or when we start seeing e.g. RRSIG expiration values in 2038 and later, which might happen earlier.

The audit should include the unit and regression tests to see if the coverage is adequate.

This audit is quite some work, but will be a little bit less complex if we can assume `time_t` is a 64-bit value.

pieterlexis:
fwiw, the regression-tests./recursor-dnssec already uses libfaketime for some of its tests. So adding tests that fake being in 2038 should be easy

jsoref:
Can you add an assert to the build system to force it to die if `time_t` isn't 64-bit?

omoerbeek:
> Can you add an assert to the build system to force it to die if `time_t` isn't 64-bit?

Sure, that will be part of the work.

pieterlexis:
adding [this](https://www.gnu.org/software/autoconf-archive/ax_compile_check_sizeof.html) macro to `m4` and adding this the configure.ac's will give you that:

```diff
+AX_COMPILE_CHECK_SIZEOF(time_t)
+AS_IF([test $ac_size -lt 8], [AC_MSG_ERROR([size of time_t is $ac_size, which is too small for the y2k38 bug])])
```

omoerbeek:
Another point to consider:

We have absolute time and relative time (i.e. duration/interval). The latter can be represented by a 32-bit quantity without issues. The format should be `time_t`. Now there are cases in the codebase where `uint32_t` is used for absolute time. While they last for quite some time after 2038, I think they should be avoided and made `time_t` unless there is a very good and documented reason to keep them `unint32_t`. The main issue is that expressions computing absolute time plus relative time should be safe without risk of overflow. If both operand are 32-bit there is no immediate guarantee so careful checking is needed. Plus reasoning about expressions using mixed unsigned and signed is harder in general.

omoerbeek:
I did an audit of the cache code in dnsdist, auth and rec.

- Rec needs fixes (#10004) and will break in 2106 since it contains code stuffing a timestamp into a `uint32_t` ttl field.
- The packetcaches in auth and rec are fine, and also the query-cache: they use `time_t` in relevant places.

[...]

PR #10004, opened 2021-01-25
rec: fix recursor caches to handle timestamps > 2038

omoerbeek:
Recursor cache one was failing, a few functions return `int32_t` instead of `time_t`.

This is a demonstration of the type of issues #10000 talks about.

There are still issues caused by using `DNSRecord`'s `d_ttl` field (a 32-bit quantity since it corresponds to the phycial representation) for absolute time.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [x] documented the code

### DIFF
--- a/pdns/recursor_cache.cc
+++ b/pdns/recursor_cache.cc
@@ -112,10 +112,10 @@ static void updateDNSSECValidationStateFromCache(boost::optional<vState>& state,
   }
 }
 
-int32_t MemRecursorCache::handleHit(MapCombo& map, MemRecursorCache::OrderedTagIterator_t& entry, const DNSName& qname, uint32_t& origTTL, vector<DNSRecord>* res, vector<std::shared_ptr<RRSIGRecordContent>>* signatures, std::vector<std::shared_ptr<DNSRecord>>* authorityRecs, bool* variable, boost::optional<vState>& state, bool* wasAuth, DNSName* fromAuthZone)
+time_t MemRecursorCache::handleHit(MapCombo& map, MemRecursorCache::OrderedTagIterator_t& entry, const DNSName& qname, uint32_t& origTTL, vector<DNSRecord>* res, vector<std::shared_ptr<RRSIGRecordContent>>* signatures, std::vector<std::shared_ptr<DNSRecord>>* authorityRecs, bool* variable, boost::optional<vState>& state, bool* wasAuth, DNSName* fromAuthZone)
 {
   // MUTEX SHOULD BE ACQUIRED
-  int32_t ttd = entry->d_ttd;
+  time_t ttd = entry->d_ttd;
   origTTL = entry->d_orig_ttl;
 
   if (variable && (!entry->d_netmask.empty() || entry->d_rtag)) {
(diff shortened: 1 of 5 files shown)


pdns/recursordist/test-recursorcache_cc.cc
```diff
@@ -19,7 +19,7 @@ BOOST_AUTO_TEST_CASE(test_RecursorCacheSimple)
   std::vector<std::shared_ptr<DNSRecord>> authRecords;
   std::vector<std::shared_ptr<RRSIGRecordContent>> signatures;
   const DNSName authZone(".");
-  time_t now = time(nullptr);
+  time_t now = INT_MAX - 15;
```
Habbie, line 22: need to remember to remove this, right?

omoerbeek: well, my plan was to do both a check for time(nullptr); and `INT_MAX - 15`;

pdns/recursor_cache.cc
```diff
@@ -112,10 +112,10 @@ static void updateDNSSECValidationStateFromCache(boost::optional<vState>& state,
   }
 }
 
-int32_t MemRecursorCache::handleHit(MapCombo& map, MemRecursorCache::OrderedTagIterator_t& entry, const DNSName& qname, uint32_t& origTTL, vector<DNSRecord>* res, vector<std::shared_ptr<RRSIGRecordContent>>* signatures, std::vector<std::shared_ptr<DNSRecord>>* authorityRecs, bool* variable, boost::optional<vState>& state, bool* wasAuth, DNSName* fromAuthZone)
+time_t MemRecursorCache::handleHit(MapCombo& map, MemRecursorCache::OrderedTagIterator_t& entry, const DNSName& qname, uint32_t& origTTL, vector<DNSRecord>* res, vector<std::shared_ptr<RRSIGRecordContent>>* signatures, std::vector<std::shared_ptr<DNSRecord>>* authorityRecs, bool* variable, boost::optional<vState>& state, bool* wasAuth, DNSName* fromAuthZone)
```
rgacogne, line 115: What bothers me with that change is that `time_t` is not guaranteed to be signed by POSIX..

omoerbeek: True, though I don't think you'll find such a system soon. We might do a `ax_check_sign` check in addition to the sizeof check in #10010 in case people start doing crazy things.

pdns/recursordist/test-negcache_cc.cc
```diff
@@ -69,6 +69,30 @@ BOOST_AUTO_TEST_CASE(test_get_entry)
...
+
+  timeval now{INT_MAX - 300, 0};
+
+  NegCache cache;
+  cache.add(genNegCacheEntry(qname, auth, now));
+
+  BOOST_CHECK_EQUAL(cache.size(), 1U);
+
+  NegCache::NegCacheEntry ne;
+  bool ret = cache.get(qname, QType(1), now, ne);
```
rgacogne, line 88:
```suggestion
  bool ret = cache.get(qname, QType(QType::A), now, ne);
```

rgacogne (approved):
Looks good, I wouldn't mind the `ax_check_sign` mentioned in the comments :)

PR #10010, opened 2021-01-26
Check sizeof(time_t) to be at least 8

omoerbeek:
### Short description

We can do this once we get rid of the raspbian builds. Code suggested by @pieterlexis.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/recursordist/configure.ac
+++ b/pdns/recursordist/configure.ac
@@ -25,11 +25,7 @@ AC_PROG_CC
 AC_PROG_CXX
 AC_LANG([C++])
 
-AX_COMPILE_CHECK_SIZEOF(time_t)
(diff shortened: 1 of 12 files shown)


pieterlexis:
This PR doesn't actually check the size of `time_t` :)

pdns/recursordist/configure.ac
```diff
@@ -25,6 +25,12 @@ AC_PROG_CC
 AC_PROG_CXX
 AC_LANG([C++])
 
+AX_COMPILE_CHECK_SIZEOF(time_t)
```
pieterlexis, line 28: It would be nice to put this in its own macro (`PDNS_CHECK_TIME_T`)

pieterlexis (approved):

PR #10013, opened 2021-01-26
auth: run lmdb init and upgrade code only once

mind04:
### Short description
Only lock until the first backend is ready.

### Checklist
I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/modules/lmdbbackend/lmdbbackend.cc
+++ b/modules/lmdbbackend/lmdbbackend.cc
@@ -58,6 +58,11 @@ static std::mutex s_lmdbStartupLock;
 
 LMDBBackend::LMDBBackend(const std::string& suffix)
 {
+  // overlapping domain ids in combination with relative names are a recipe for disater
+  if (!suffix.empty()) {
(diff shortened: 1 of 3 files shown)


modules/lmdbbackend/lmdbbackend.cc
```diff
@@ -38,30 +39,25 @@
...
 
 #define SCHEMAVERSION 3
-#define SCHEMAVERSION_TEXT "3"
+
 // List the class version here. Default is 0
 BOOST_CLASS_VERSION(LMDBBackend::KeyDataDB, 1)
 
-static std::mutex s_lmdbOpenUpgradeLock;
+static bool s_first = true;
+static int s_shards = 0;
```
Habbie, line 56: This breaks if multiple LMDB backends are launched.

Habbie (changes requested):
This looks like it would not work correctly if multiple LMDB backends are launched.

Habbie (approved):

PR #10014, opened 2021-01-26
Rec: validate signatures after 2038 and more

omoerbeek:
### Short description

This PR allows the test code to run unit tests for SyncRes at a specific point in time and uses that to test valdiating resolution at those points in time. Currently this fails in the year 2106, since the expiry and inceptions validation code in `validate.cc` does not take rfc 4034 into account. 2038 is OK.

Part of the work for #10000

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [X] documented the code

### DIFF
--- a/pdns/validate.cc
+++ b/pdns/validate.cc
@@ -763,11 +763,13 @@ static const vector<DNSName> getZoneCuts(const DNSName& begin, const DNSName& en
 
 bool isRRSIGNotExpired(const time_t now, const shared_ptr<RRSIGRecordContent>& sig)
 {
+  // Should use https://www.rfc-editor.org/rfc/rfc4034.txt section 3.1.5 
   return sig->d_sigexpire >= now;
 }
 
 bool isRRSIGIncepted(const time_t now, const shared_ptr<RRSIGRecordContent>& sig)
 {
+  // Should use https://www.rfc-editor.org/rfc/rfc4034.txt section 3.1.5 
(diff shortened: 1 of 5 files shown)


pdns/validate.cc
```diff
@@ -763,11 +763,13 @@ static const vector<DNSName> getZoneCuts(const DNSName& begin, const DNSName& en
 
 bool isRRSIGNotExpired(const time_t now, const shared_ptr<RRSIGRecordContent>& sig)
 {
+  // Should use https://www.rfc-editor.org/rfc/rfc4034.txt section 3.1.5 
   return sig->d_sigexpire >= now;
```
rgacogne, line 767: We have `rfc1982LessThan` in our code base, but I'm not sure it can be used with these types.

omoerbeek: I tried to implement this, but the 2106 still fails. From a quick look because the record cache cannot handle timestamps beyond 2106, causing trouble finding the root servers. I'd say we leave this for some other time.

rgacogne: Fine by me :) Perhaps we can open a new issue unless the validation case is already covered by the existing one?

rgacogne (approved):

pdns/validate.cc
omoerbeek: #10000 mentions it.

PR #10016, opened 2021-01-26
builder: add arch-specific targets

Habbie:
### Short description
This allows us to be specific about what architecture to build for. The 'base' target remains, for anybody that just wants to build for their default arch. It might make sense to disable the base targets at some point.

### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/builder-support/dockerfiles/Dockerfile.target.centos-7
+++ b/builder-support/dockerfiles/Dockerfile.target.centos-7
@@ -3,7 +3,13 @@
 
 # This defines the distribution base layer
 # Put only the bare minimum of common commands here, without dev tools
+@IF [ ${BUILDER_TARGET} = centos-7 ]
 FROM centos:7 as dist-base
+@ENDIF
+@IF [ ${BUILDER_TARGET} = centos-7-amd64 ]
+FROM amd64/centos:7 as dist-base
+@ENDIF
+
 ARG BUILDER_CACHE_BUSTER=
 RUN touch /var/lib/rpm/* && yum install -y epel-release centos-release-scl-rh
 RUN touch /var/lib/rpm/* && yum install -y --nogpgcheck devtoolset-8-gcc-c++
--- a/builder-support/dockerfiles/Dockerfile.target.ubuntu-focal
+++ b/builder-support/dockerfiles/Dockerfile.target.ubuntu-focal
@@ -1,7 +1,15 @@
 # First do the source builds
 @INCLUDE Dockerfile.target.sdist
 
+@IF [ ${BUILDER_TARGET} = ubuntu-focal ]
 FROM ubuntu:focal as dist-base
+@ENDIF
+@IF [ ${BUILDER_TARGET} = ubuntu-focal-amd64 ]
+FROM amd64/ubuntu:focal as dist-base
+@ENDIF
+@IF [ ${BUILDER_TARGET} = ubuntu-focal-arm64 ]
+FROM arm64v8/ubuntu:focal as dist-base
+@ENDIF
 ARG BUILDER_CACHE_BUSTER=
 ARG APT_URL
 RUN apt-get update && apt-get -y dist-upgrade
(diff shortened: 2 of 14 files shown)


Habbie:
I'll file a pdns-builder PR to list the targets vertically instead of on one line, and then I'll see if we can bump the submodule here.

builder-support/dockerfiles/Dockerfile.target.centos-7
```diff
@@ -3,7 +3,13 @@
 
 # This defines the distribution base layer
 # Put only the bare minimum of common commands here, without dev tools
+@IF [ ${BUILDER_TARGET} = centos-7 ]
 FROM centos:7 as dist-base
+@ENDIF
+@IF [ ${BUILDER_TARGET} = centos-7-amd64 ]
```
james-crowley, line 9: @Habbie Is there a reason why you are calling the arch specific image? Most of these images all support manifest and should pull down the correct image based on your architecture. It would make sense to have an if/branch if the specific images does not have support. 

For example `centos` images do not support `s390x`. So a lot of the times I have logic that checks what architecture I am building for, if its `s390x`, than I would pull from `clefos`. ClefOS being a `s390x` CentOS clone.

pieterlexis: Hi James, this PR has nothing to do with building images. We utilize docker to produce RPM/DEB packages for different distros (and soon architectures), but out scripts need to know exactly what they're building.

james-crowley: @pieterlexis Thanks for providing the needed context. When building the RPM/DEB packages is calling specific architecture image needed? Unless you are using an emulation layer like QEMU, Docker should detect the correct architecture and pull the correct image from the manifest. 

Are you building RPM/DEB packages using QEMU/emulation or on architecture currently?

Habbie: Currently (without this PR), we build amd64 and raspbian armhf images, on amd64 machines. For armhf, we use qemu.

After this PR (plus related work in our buildbot setup), the plan is to build amd64 on amd64, and arm64+armhf on Amazon Graviton2 ARM64 boxes (that, unlike some other ARM64 chips, can run arm32 natively too).

The original reason for the change you are asking about is that the chosen target also decides the tarball name for the built packages. As a bonus effect, it allows one to force building packages for different architectures than the host, possibly, indeed, involving qemu.

Note that the diff does not remove the unprefixed version, that would, as you say, pull the correct image.

And, I did not realise this before, indeed it would allow us to choose forks based on their architecture support. :)

But as Pieter pointed out, none of this is about our *Docker images*, which is what you filed a request for in a separate ticket.

Habbie:
I've bumped pdns-builder, and compared a bunch of the debian-buster and centos-8 packages. I found no differences.

Habbie:
This PR really wants https://github.com/PowerDNS/pdns-builder/pull/46

Habbie:
> This PR really wants [PowerDNS/pdns-builder#46](https://github.com/PowerDNS/pdns-builder/pull/46)

(But can be merged independently)

builder-support/dockerfiles/Dockerfile.target.ubuntu-focal
```diff
@@ -1,7 +1,15 @@
 # First do the source builds
 @INCLUDE Dockerfile.target.sdist
 
+@IF [ ${BUILDER_TARGET} = ubuntu-focal ]
```
pieterlexis, line 4: Is what would be built dependent on the host system?

Habbie: yes

pieterlexis: it was phrased as a question, but my feeling was "this might be painful". Although we have specific arch, one can use those.

Habbie: Yes, I left this one in on purpose, but perhaps we can take it out when we're done setting up multiarch in Buildbot.

pieterlexis (approved):

PR #10022, opened 2021-01-27
rec: Move prometheus table to web server and add missing entries.

### LINKED ISSUE #10021: rec: init and retrieval of stats map is racy
- Program:  Recursort 
 - Issue type: Bug report

### Short description

In particular, the webserver code might access the stats map while it isn't completely initialized yet. There's is code to avoid initing the stats map twice, but not against accessing it half-initialized.

omoerbeek:
Also fixes #10021

There is code to check the tables, but it is disabled.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code


omoerbeek:
Wondering if some form of memory barrier or atomic is needed to be safe accesing and modifying `s_inited` even on weak memory order machines.

pieterlexis:
@omoerbeek can you milestone this?

rgacogne (approved):

PR #10027, opened 2021-01-29
-Werror experiment

omoerbeek:
### Short description

See if `-Werror` is feasible. 

Its a bit complicated, since not all compiler agree on warnings. I had to use `-Wno-error=xyz` for a few things. Let's see how this fares with the various toolsets. 

It still might make the lives of maintainers harder since their environment might need more `Wno-error=...`. 

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code


omoerbeek:
This is causing  more subtle configure errors like failing to detect the proper boost context mode. Abandoning for now.

PR #10035, opened 2021-02-01
dnsdist: Fix a few warnings reported by clang's static analyzer and cppcheck

rgacogne:
### Short description

Mostly false positive that I'm tired of seeing, or performance tips in places we don't really care about performance.
But also a real bug in the code loading the ring buffers from a file (only used for debugging) and a few hardening measures (additional checks, initializations) that will not hurt and might prevent an issue later. 

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdist-xpf.cc
+++ b/pdns/dnsdist-xpf.cc
@@ -52,6 +52,7 @@ bool addXPF(DNSQuestion& dq, uint16_t optionCode)
   pos += sizeof(drh);
   memcpy(reinterpret_cast<char*>(&data.at(pos)), payload.data(), payload.size());
   pos += payload.size();
+  (void) pos;
 
   dq.getHeader()->arcount = htons(ntohs(dq.getHeader()->arcount) + 1);
 
--- a/ext/luawrapper/include/LuaContext.hpp
+++ b/ext/luawrapper/include/LuaContext.hpp
@@ -204,7 +204,7 @@ public:
     class WrongTypeException : public std::runtime_error
     {
     public:
-        WrongTypeException(std::string luaType_, const std::type_info& destination_) :
+        WrongTypeException(const std::string& luaType_, const std::type_info& destination_) :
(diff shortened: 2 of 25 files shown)


pdns/dnsdist-xpf.cc
```diff
@@ -52,6 +52,7 @@ bool addXPF(DNSQuestion& dq, uint16_t optionCode)
   pos += sizeof(drh);
   memcpy(reinterpret_cast<char*>(&data.at(pos)), payload.data(), payload.size());
   pos += payload.size();
+  (void) pos;
```
Habbie, line 55: This is to mark the result from the previous line used, and you don't want to delete the previous line in case we ever add more code?

Habbie (approved):
approved with one unimportant question

pdns/dnsdist-xpf.cc
rgacogne: Exactly. I can live with deleting the previous line if you think it's cleaner, though :)

Habbie: What you did makes sense :)

Habbie:
CodeQL reports 4 fixed alerts (all 'Potentially uninitialized local variable ')

rgacogne:
> CodeQL reports 4 fixed alerts (all 'Potentially uninitialized local variable ')

Where does this get reported? I don't see it in the check details?

Habbie:
> Where does this get reported? I don't see it in the check details?

It's a few clicks into the Checks tab, under Security scanning > CodeQL

rgacogne:
Fun fact, when I click on "All checks have passed" then "CodeQL / Analyze..." -> Details the "Code scanning results" menu is empty. But it's there when I go through the "Checks" tag. Thanks :)

Unfortunately these 4 alerts seem to randomly get fixed and found again, I don't think it's related to my PR..

PR #10037, opened 2021-02-01
auth: run deleteDomain() inside a transaction

mind04:
### Short description
deleteDomain() is running a few queries to delete domain data from different tables. If one of those queries is failing (for example due to locks) a partial deleted domain is left behind.

### Checklist
I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/pdnsutil.cc
+++ b/pdns/pdnsutil.cc
@@ -819,8 +819,18 @@ static int deleteZone(const DNSName &zone) {
     return EXIT_FAILURE;
   }
 
-  if(di.backend->deleteDomain(zone))
-    return EXIT_SUCCESS;
+  di.backend->startTransaction(zone, -1);
+  try {
+    if(di.backend->deleteDomain(zone)) {
+      di.backend->commitTransaction();
+      return EXIT_SUCCESS;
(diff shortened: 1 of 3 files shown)


Habbie (approved):

pdns/pdnsutil.cc
```diff
@@ -819,8 +819,18 @@ static int deleteZone(const DNSName &zone) {
...
+      return EXIT_SUCCESS;
+    }
+  } catch (...) {
+    di.backend->abortTransaction();
+    throw;
+  }
+
+  di.backend->abortTransaction();
 
   cerr<<"Failed to delete domain '"<<zone<<"'"<<endl;;
```
Habbie, line 835: this code has become unreachable (and the few lines before and after it too)

Habbie (changes requested):

pdns/pdnsutil.cc
mind04: The code is not unreachable. You hit this when a backend does not support deleteDomain()

Habbie (approved):

PR #10041, opened 2021-02-02
rec: Also account work done by handler thread, like priming

omoerbeek:
### Short description

`broadcastFunction` does the right thing, but `broadcastAccFunction` not.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/pdns_recursor.cc
+++ b/pdns/pdns_recursor.cc
@@ -3798,9 +3798,16 @@ template<class T> T broadcastAccFunction(const boost::function<T*()>& func)
   }
 
   unsigned int n = 0;
-  T ret=T();
+  T ret = T();
+
   for (const auto& threadInfo : s_threadInfos) {
+    T* resp = nullptr;
     if (n++ == t_id) {
+      resp = func();
+      if (resp != nullptr) {
+        ret += *resp;
+        delete resp;
+      }
       continue;
     }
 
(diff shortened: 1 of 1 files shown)


pdns/pdns_recursor.cc
```diff
@@ -3809,19 +3816,18 @@ template<class T> T broadcastAccFunction(const boost::function<T*()>& func)
...
-    if(write(tps.writeToThread, &tmsg, sizeof(tmsg)) != sizeof(tmsg)) {
+    if (write(tps.writeToThread, &tmsg, sizeof(tmsg)) != sizeof(tmsg)) {
       delete tmsg;
       unixDie("write to thread pipe returned wrong size or error");
     }
 
-    T* resp = nullptr;
-    if(read(tps.readFromThread, &resp, sizeof(resp)) != sizeof(resp))
+    if (read(tps.readFromThread, &resp, sizeof(resp)) != sizeof(resp)) {
+      delete tmsg;
```
rgacogne, line 3825: I'm not so sure it's a good idea to delete it now. It has successfully been written to the pipe, the fact that we are not getting an answer might not mean that the other end has not received it? Perhaps it did, got an exception but already deleted the object? A small leak might be better than a double-free?

omoerbeek: Right, the responsibility is at the other end.

pdns/pdns_recursor.cc
```diff
@@ -3798,9 +3798,16 @@ template<class T> T broadcastAccFunction(const boost::function<T*()>& func)
   }
 
   unsigned int n = 0;
-  T ret=T();
+  T ret = T();
+
   for (const auto& threadInfo : s_threadInfos) {
+    T* resp = nullptr;
     if (n++ == t_id) {
+      resp = func();
```
rgacogne, line 3806: I wonder if it's OK for all `broadcastAddFunction` calls. For example this change will cause the `lua-dns-script` to be loaded on the handler thread on a `reload-lua-script` command, while it's not normally loaded.

omoerbeek: That's a valid concern indeed. The handler thread is in some way a regular recursor thread, but in other ways not.  I have to think what's the best solution: don't be concerned the handler thread stats are not accounted for, or careful checking of all cases to see what would happen. Moving this to `Draft` for now.

omoerbeek:
Other work done by the handler threads is incoming tcp. Now that we have a packet cache that's also used for tcp, it would be nice if those numbers end up in the stats as well. 

BTW, the pruning of the packet cache is now based on the max size / number of workers, should maybe become number of workers + 1, for the above reason.

rgacogne:
Isn't incoming TCP handled by the listener thread(s), which are the distributors if there is any and the workers otherwise?

omoerbeek:
With 2 worker threads, I saw the PC dumps of three threads in a cache dump (first one being all tcp) and the concluded (maybe too quickly) it was the handler thread. Will investigate in more detail later.

omoerbeek:
We have indeed 1 PC cache per distributor thread plus 1 per worker thread. It still seems a bit wrong to do the sizing on the number of worker threads only.

rgacogne:
> It still seems a bit wrong to do the sizing on the number of worker threads only.

Agreed! So perhaps number of workers + number of distributors?

omoerbeek:
yes, will prep a PR for that soon

PR #10045, opened 2021-02-02
Rec: experimental NS endpoint type and TCP connection pooling 

omoerbeek:
### Short description

This is a experimental code, implementing two related features

- Ability to specify the type of connection to use for forwarding. e.g.  `forward-zones-recurse=.=TCP/8.8.8.8`
To be able to track what kind of connection to use for a specific NS, this records the type of endpoint for NS data. ATM, mostly `Unspecified` i.e. use regular UDP with fallback to TCP. In the future this can be extended to force DoT (or other?) connection methods to auths.
- A connection pooling  mechnism for outgoing TCP connections. This avoid having to set up a new TCP connection for every outgoing TCP query always. 

Mostly creating a PR to be able to run CI on it (now that forked repos are not supposed to do that) and making discussion the code more easy.

Note that this also contains a version of #10041

TODO:
- [ ] fix file descriptor limit code to avoid underflow
- [ ] imake sure we never grow the map beyond maxPerThread so the cleanup code can be much simpler
- [ ] change description (and maybe name of setting) of maxIdle: "after a connection has been idle for this time is it eligible for GC", there is no guarantee it will be GC'ed immediately
- [ ] tests!

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/lwres.cc
+++ b/pdns/lwres.cc
@@ -48,6 +48,7 @@
 #include "validate-recursor.hh"
 #include "ednssubnet.hh"
 #include "query-local-address.hh"
+#include "rec-tcp-out.hh"
 
 #include "rec-protozero.hh"
 #include "uuid-utils.hh"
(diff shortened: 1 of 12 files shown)


omoerbeek:
Should be reworked once #10108  and #9995 are merged.

omoerbeek:
Obsolete, I'm working on an updated version that uses the new `TCPIOHandler` code.

PR #9377, opened 2020-08-10
rec: It's ok to apply nothing-below-nxdomain for the auth/forwarding case.

omoerbeek:
This becomes more relevant when the root stub PR (#9351) is merged.
The root-nx-trust case looks already covered.

### Checklist

I have:
- [X] read the [CONTRIBUTING.md](https://github.com/PowerDNS/pdns/blob/master/CONTRIBUTING.md) document
- [X] compiled this code
- [X] tested this code
- [ ] included documentation (including possible behaviour changes)
- [ ] documented the code
- [ ] added or modified regression test(s)
- [ ] added or modified unit test(s)
- [ ]  checked that this code was merged to master

PR #10047, opened 2021-02-02
rec: Implement rfc 8198 - Aggressive Use of DNSSEC-Validated Cache

rgacogne:
### Short description

This is a work-in-progress PR, to see how it fares against our CI tests and get the feedback on the design choices.
Some of these are:
- using a separate cache instead of reusing/abusing the existing record cache: we don't store NSEC3 in the record cache, and the sharding makes it very hard to reuse it efficiently without writing some horrible hack, but more importantly I believe keeping the NSEC/NSEC3 ordered by zone reduce the amount of work doing the search by a nice margin.
- checking that new cache early means that we can be searching in the wrong zone, like in the root zone, even though we do have some information in our record cache about a more specific zone existing. This should not have any functional impact but likely has a performance impact. On the other hand doing the check after computing the zone cuts is not efficient at all. Perhaps we could do a quick cache-only lookup similar to what we do to get the best NSs? 

This PR needs regression tests before being moved from the draft status, at the very least.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/pdns/recursordist/aggressive_nsec.cc
+++ b/pdns/recursordist/aggressive_nsec.cc
@@ -172,6 +172,7 @@ void AggressiveNSECCache::insertNSEC(const DNSName& zone, const DNSName& owner, 
       entry->d_entries.insert({record.d_content, signatures, owner, std::move(next), record.d_ttl});
     }
   }
+  ++d_entriesCount;
 }
 
 bool AggressiveNSECCache::getNSECBefore(time_t now, std::shared_ptr<AggressiveNSECCache::ZoneEntry>& zoneEntry, const DNSName& name, ZoneEntry::CacheEntry& entry) {
@@ -219,6 +220,7 @@ bool AggressiveNSECCache::getNSECBefore(time_t now, std::shared_ptr<AggressiveNS
 
   if (it->d_ttd <= now) {
     idx.erase(it);
+    --d_entriesCount;
     return false;
   }
 
@@ -245,6 +247,7 @@ bool AggressiveNSECCache::getNSEC3(time_t now, std::shared_ptr<AggressiveNSECCac
 
     if (it->d_ttd <= now) {
       idx.erase(it);
+      --d_entriesCount;
       return false;
     }
 
@@ -332,6 +335,7 @@ bool AggressiveNSECCache::synthesizeFromNSEC3Wildcard(time_t now, const DNSName&
   /* and of course we won't deny the wildcard either */
 
   LOG("Synthesized valid answer from NSEC3s and wildcard!"<<endl);
+  ++d_nsec3WildcardHits;
   return true;
 }
 
@@ -351,6 +355,7 @@ bool AggressiveNSECCache::synthesizeFromNSECWildcard(time_t now, const DNSName& 
   addRecordToRRSet(now, nsec.d_owner, QType::NSEC3, nsec.d_ttd - now, nsec.d_record, nsec.d_signatures, doDNSSEC, ret);
 
   LOG("Synthesized valid answer from NSECs and wildcard!"<<endl);
+  ++d_nsecWildcardHits;
   return true;
 }
 
@@ -389,6 +394,7 @@ bool AggressiveNSECCache::getNSEC3Denial(time_t now, std::shared_ptr<AggressiveN
     }
 
     LOG(": done!"<<endl);
+    ++d_nsec3Hits;
     res = RCode::NoError;
     addToRRSet(now, soaSet, soaSignatures, zoneEntry->d_zone, doDNSSEC, ret);
     addRecordToRRSet(now, exactNSEC3.d_owner, QType::NSEC3, exactNSEC3.d_ttd - now, exactNSEC3.d_record, exactNSEC3.d_signatures, doDNSSEC, ret);
@@ -477,6 +483,7 @@ bool AggressiveNSECCache::getNSEC3Denial(time_t now, std::shared_ptr<AggressiveN
   addRecordToRRSet(now, wcEntry.d_owner, QType::NSEC3, wcEntry.d_ttd - now, wcEntry.d_record, wcEntry.d_signatures, doDNSSEC, ret);
 
   LOG("Found valid NSEC3s covering the requested name and type!"<<endl);
+  ++d_nsec3Hits;
   return true;
 }
 
@@ -589,5 +596,6 @@ bool AggressiveNSECCache::getDenial(time_t now, const DNSName& name, const QType
   }
 
   LOG("Found valid NSECs covering the requested name and type!"<<endl);
+  ++d_nsecHits;
   return true;
 }
--- a/pdns/recursordist/aggressive_nsec.cc
+++ b/pdns/recursordist/aggressive_nsec.cc
@@ -1,0 +1,578 @@
+/*
+ * This file is part of PowerDNS or dnsdist.
+ * Copyright -- PowerDNS.COM B.V. and its contributors
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of version 2 of the GNU General Public License as
+ * published by the Free Software Foundation.
+ *
+ * In addition, for the avoidance of any doubt, permission is granted to
+ * link this program with OpenSSL and to (re)distribute the binaries
+ * produced as the result of such linking.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+
+#include "aggressive_nsec.hh"
+#include "recursor_cache.hh"
+#include "validate.hh"
+
+std::unique_ptr<AggressiveNSECCache> g_aggressiveNSECCache{nullptr};
+
+/* this is defined in syncres.hh and we are not importing that here */
+extern std::unique_ptr<MemRecursorCache> g_recCache;
+
+std::shared_ptr<AggressiveNSECCache::ZoneEntry> AggressiveNSECCache::getBestZone(const DNSName& zone)
+{
+  std::shared_ptr<AggressiveNSECCache::ZoneEntry> entry{nullptr};
+  {
+    ReadLock rl(d_lock);
+    auto got = d_zones.lookup(zone);
+    if (got) {
+      return *got;
+    }
+  }
+  return entry;
+}
+
+std::shared_ptr<AggressiveNSECCache::ZoneEntry> AggressiveNSECCache::getZone(const DNSName& zone)
+{
+  std::shared_ptr<AggressiveNSECCache::ZoneEntry> entry{nullptr};
+  {
+    ReadLock rl(d_lock);
+    auto got = d_zones.lookup(zone);
+    if (got && *got && (*got)->d_zone == zone) {
+      return *got;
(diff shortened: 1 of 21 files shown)


rgacogne:
Forgot to mention that it does implement NSEC and NSEC3 aggressive caching, including wildcard expansion.

Habbie:
> Forgot to mention that it does implement NSEC and NSEC3 aggressive caching, including wildcard expansion.

I can confirm that this absolutely breaks wildcard `LUA createForward6`:

```
2001-0-0-0-0-0-0-2.example.com.	40 IN	AAAA	2001::1
```

:-)

Habbie:
Double domainname (example.com.example.com) here:

```
; aggressive NSEC cache dump follows
;
; Zone example.com.
*.example.com. 33 IN NSEC example.com.example.com. AAAA RRSIG NSEC
- RRSIG NSEC 13 2 60 20210211000000 20210121000000 39646 example.com. Gu9kzkf2suGcfownpxTdkC6XFxxpxhOzzUqAkf+BsWBonj5xAghZ8RagjTfJN15uyVAd9/XHJ4o8VxlbFl0uwA==
```

rgacogne:
> *.example.com. 33 IN NSEC example.com.example.com. AAAA RRSIG NSEC

That's very weird, because while we do reset the owner name to the original wildcard value, we should not touch the content. Do you happen to have the original NSEC lying around?

Habbie:
> Do you happen to have the original NSEC lying around?

in my zone:
`example.com.example.com	3005	IN	NS	ns1.example.com.`

please ignore my report :)

pdns/recursordist/aggressive_nsec.cc
```diff
@@ -0,0 +1,758 @@
...
+          }
+        }
+      }
+
+      if (node.d_value->d_entries.size() == 0) {
+        emptyEntries.push_back(node.d_value->d_zone);
+      }
+    });
+  }
+
```
omoerbeek, line 190: Here I would have expected a loop through `emptyEntries` to clear the empty zones.

pdns/syncres.cc
```diff
@@ -3276,13 +3295,38 @@ RCode::rcodes_ SyncRes::updateCacheFromRecords(unsigned int depth, LWResult& lwr
           }
         }
       }
+
       if (doCache) {
         g_recCache->replace(d_now.tv_sec, i->first.name, QType(i->first.type), i->second.records, i->second.signatures, authorityRecs, i->first.type == QType::DS ? true : isAA, auth, i->first.place == DNSResourceRecord::ANSWER ? ednsmask : boost::none, d_routingTag, recordState, remoteIP);
+
+        if (needWildcardProof && recordState == vState::Secure && i->first.place == DNSResourceRecord::ANSWER && g_aggressiveNSECCache && i->first.name == qname && !i->second.signatures.empty() && !d_routingTag && !ednsmask) {
```
omoerbeek, line 3302: Move this `g_aggressiveNSECCache` part to the front? That makes it very clear this block is only relevent if it is true

pdns/syncres.cc
```diff
@@ -3276,13 +3295,38 @@ RCode::rcodes_ SyncRes::updateCacheFromRecords(unsigned int depth, LWResult& lwr
...
+            }
+
+            g_recCache->replace(d_now.tv_sec, realOwner, QType(i->first.type), content, i->second.signatures, /* no additional records in that case */ {}, i->first.type == QType::DS ? true : isAA, auth, boost::none, boost::none, recordState, remoteIP);
+          }
+        }
       }
     }
 
-    if(i->first.place == DNSResourceRecord::ANSWER && ednsmask)
+    if ((i->first.type == QType::NSEC || i->first.type == QType::NSEC3) && recordState == vState::Secure && !seenAuth.empty() && g_aggressiveNSECCache) {
```
omoerbeek, line 3322: Move this `g_aggressiveNSECCache` part to the front? That makes it very clear this block is only relevant if it is true

omoerbeek:
>     * using a separate cache instead of reusing/abusing the existing record cache: we don't store NSEC3 in the record cache, and the sharding makes it very hard to reuse it efficiently without writing some horrible hack, but more importantly I believe keeping the NSEC/NSEC3 ordered by zone reduce the amount of work doing the search by a nice margin.

Yes, and it is a very big advantage that the aggressive cache does not interfere with the existing record cache. That gives a lot more confidence that this can be merged without regressions. The very limited changes in `SyncRes` also helps in tat respect. So I 100% agree with this choice.

> 
>     * checking that new cache early means that we can be searching in the wrong zone, like in the root zone, even though we do have some information in our record cache about a more specific zone existing. This should not have any functional impact but likely has a performance impact. On the other hand doing the check after computing the zone cuts is not efficient at all. Perhaps we could do a quick cache-only lookup similar to what we do to get the best NSs?

If it is: we can often decide much faster if a name does not exist, but in some cases we get performance that might be the same as it was before, I do not see a problem. Or can performance be negatively impacted compared to the current state? i.e. ca a lookup with no result in the aggressive cache slow down the resolving?

Is this influenced by the zone cute elimination PR (#10057)?

pieterlexis:
> Is this influenced by the zone cute elimination PR (#10057)?

How big are the conflict with either? I don't want to find @rgacogne crying for having to go through another painful rebase :)

omoerbeek:
> > Is this influenced by the zone cute elimination PR (#10057)?
> 
> How big are the conflict with either? I don't want to find @rgacogne crying for having to go through another painful rebase :)

Another advantage of the small amount of changes in `SyncRes` of this PR: Almost all changes of the other PR are in `SyncRes`, so no large and complex conflicts afaiks.

I was just wondering how big the performance impact (good or bad) would be if the zonecut elimination PR is taken into account.

rgacogne:
> If it is: we can often decide much faster if a name does not exist, but in some cases we get performance that might be the same as it was before, I do not see a problem. Or can performance be negatively impacted compared to the current state? i.e. ca a lookup with no result in the aggressive cache slow down the resolving?

It will slow down the resolving a bit, since the aggressive cache lookup is not free. I'll do more measurements but I expect to that lookup to be quite cheap, especially considering that we are going to have to go to the network anyway. 

> Is this influenced by the zone cute elimination PR (#10057)?

#10057 removes the "easy" way to make sure we don't do a lookup in the wrong zone, or for an insecure zone. So either we find a way to get some information from something we would be paying for anyway (like getting the best NS), or I think it's not worth it and the aggressive cache lookup is likely cheaper than anything else we could do to know the best zone and its status.

Habbie:
What happens if a zone is in the middle of an NSEC3 salt/iterations roll? Or if multiple auths for a zone use different parameters?

rgacogne:
> What happens if a zone is in the middle of an NSEC3 salt/iterations roll? Or if multiple auths for a zone use different parameters?

The short version is that we will use the latest NSEC3 parameters (salt/iterations) that we have seen for a zone. If we find something with that in our cache that's great, otherwise too bad. We could remember more than one but honestly I don't think it's worth it.

Habbie:
> The short version is that we will use the latest NSEC3 parameters (salt/iterations) that we have seen for a zone. If we find something with that in our cache that's great, otherwise too bad. We could remember more than one but honestly I don't think it's worth it.

That makes sense - looking at the code however I am not sure we check that what we found matches the params, did I miss it?

rgacogne:
I don't think we do check, and that's indeed bad because we could use a NSEC3 hashed with the wrong parameters. I'll add some checks, thanks!

Habbie:
> I don't think we do check, and that's indeed bad because we could use a NSEC3 hashed with the wrong parameters. I'll add some checks, thanks!

Right, that was the point. Thanks!

rgacogne:
> I don't think we do check, and that's indeed bad because we could use a NSEC3 hashed with the wrong parameters. I'll add some checks, thanks!

Done!

rgacogne:
Did a test against umbrella top 1m, I see a 0.4% user CPU increase and a 0.1% system CPU increase with that feature enabled. I also get:
```
aggressive-nsec-cache-nsec-hits 17894
aggressive-nsec-cache-nsec-wc-hits      54
aggressive-nsec-cache-nsec3-hits        21286
aggressive-nsec-cache-nsec3-wc-hits     86
```
which is much more than I expected to see, I'll investigate.

I also did a quick test simulating a PRSD attack against a single zone. First that feature enabled then disabled:
Screenshot_20210222_181506
Screenshot_20210222_181524
Screenshot_20210222_181540
Screenshot_20210222_181606

rgacogne:
Rebased on master to fix conflicts.

Habbie:
> I also did a quick test simulating a PRSD attack against a single zone. First that feature enabled then disabled:

That looks amazing, but why the SERVFAILs in the disabled case? Auth overloaded?

omoerbeek:
> > I also did a quick test simulating a PRSD attack against a single zone. First that feature enabled then disabled:
> 
> That looks amazing, but why the SERVFAILs in the disabled case? Auth overloaded?

Could also be recursor overload, it goes together with a large proportion of slow auth4 answers and throttling.

rgacogne:
Clearly auth overload, the dnsdists in front of my auths (same box) report a lot of timeouts at that time.

rgacogne:
> which is much more than I expected to see, I'll investigate.

After closer inspection that looks legit. We get some hits because of the fact that `dnsbulktest` adds a second query with `www` appended by default (and that also explains most of the wildcard hits), some because the denial for a DS also denies the queried type, etc.

omoerbeek (approved):
I'm happy for this to be merged if you are happy too. Please enable it by default for now. Based on feedback we'll decide to leave it enabled or not for 4.5.0

rgacogne:
I'm writing some regression tests and I think we will be good to go. To enable it by default I need to pick a value for the maximum number of entries in the NSEC cache, do you have any opinion? 100k seems to be a good value but it might be quite an increase in memory usage, so perhaps 10k for a start?

rgacogne:
I just added several regression tests.

rgacogne:
> 100k seems to be a good value but it might be quite an increase in memory usage, so perhaps 10k for a start?

On second thoughts, the default packet cache size is 500kn and 1M for the record cache, so 100k might be a good default. Still our default DNSSEC mode is `process-no-validate`, which makes the aggressive NSEC cache useless, do we have any plan to change the default DNSSEC mode?

omoerbeek:
> > 100k seems to be a good value but it might be quite an increase in memory usage, so perhaps 10k for a start?
> 
> On second thoughts, the default packet cache size is 500kn and 1M for the record cache, so 100k might be a good default. Still our default DNSSEC mode is `process-no-validate`, which makes the aggressive NSEC cache useless, do we have any plan to change the default DNSSEC mode?

100k feels right. As for default dnssec mode, no changes planned for this release. But it would be nice if people enabling dncsec get the aggressive cache as well. Quite some people enable dnssec.

rgacogne:
Done! Note that it does mean that the default configuration yields a warning at startup:
```
Aggressive NSEC/NSEC3 caching is enabled but DNSSEC validation is not set to 'validate', 'log-fail' or 'process', ignoring
```

pieterlexis:
we could consider enabling DNSSEC validation by default nowadays? knot, bind, and unbound do this afaik (and is easy to disable)

rgacogne:
if we don't want to go with `validate` as a default, I would not mind having `process` :-)

rgacogne:
Running the regression tests suite with that feature enabled revealed a nasty bug related to the handling of the DNAME bit in parent NSEC(3)s, which is now fixed.

rgacogne:
So we have a problem with the Interop test for #4369, which is the same than the issue Otto had in #9377. If we want to allow fowarding to a different auth in the middle of a Secure zone, with no zone cuts, we can't apply the aggressive NSEC cache any more than "nothing-below-nxdomain" to any forwarded zone. That makes me sad because all forwarded zones, even the ones with a proper zone cut, will lose the aggressive cache protection, but I'll prepare a patch in that direction.

PR #10054, opened 2021-02-04
Stop enforcing format

omoerbeek:
Most important reason: not all version of check-format do the same. I still encourage people to use the script `build-scrips/format-code` though.

### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -529,39 +529,6 @@ jobs:
           paths:
             - project
 
-  check-formatting:
-    docker:
-      - image: debian:buster
-        auth:
-          username: powerdnsreadonly
-          password: $DOCKERHUB_PASSWORD
-    steps:
-      - run:
-          name: Install dependencies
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -529,39 +529,6 @@ jobs:
...
-      - get-workspace
-      - run:
-          name: Check formatting
-          command: |
-            ./build-scripts/format-code pdns/recursordist/*.[ch][ch]
-            git --no-pager diff
-            exit $(git diff | wc -l)
-          working_directory: ~/project
-      - run:
-          name: Check Makefile.am SOURCES sort order
```
pieterlexis, line 554: This needs to stay

omoerbeek:
The feeling seems to be that the annoyance is quite limited, I'll try to find a way around it anyway so we can keep the enforcing.

Issue #9666, opened 2020-11-03
auth: add LMDB to API testing


Habbie:
With #9664 and #9665 applied, the LMDB backend no longer crashes during normal or API testing, but some API tests fail.

With just #9665 applied, we still crash, so it appears 9665 is hiding some other issue beyond what 9664 fixes.

Also, `pdnsutil secure-zone example.com` segfaults if the zone does not exist.

Habbie:
Part of the API failures are explained by `disabled` not working at all in LMDB.

Habbie:
This also makes sense (although it fixes nothing) - or a rewrite of that function:
```diff
diff --git a/modules/lmdbbackend/lmdbbackend.cc b/modules/lmdbbackend/lmdbbackend.cc
index daba3ae16..87f0f480e 100644
--- a/modules/lmdbbackend/lmdbbackend.cc
+++ b/modules/lmdbbackend/lmdbbackend.cc
@@ -241,7 +241,9 @@ std::string serToString(const vector<DNSResourceRecord>& rrs)
 static inline size_t serOneRRFromString(const string_view& str, DNSResourceRecord& rr)
 {
   uint16_t len;
+  if (str.size() < 2) throw std::out_of_range("serialized RR is too short for length prefix");
   memcpy(&len, &str[0], 2);
+  if (str.size() < (2 + len + 7)) throw std::out_of_range("serialized RR is too short for full content length");
   rr.content.assign(&str[2], len);    // len bytes
   memcpy(&rr.ttl, &str[2] + len, 4);
   rr.auth = str[2 + len + 4];
```

Habbie:
A useful testcase for 9664/9665:
```
$ cat tests/pdnsutil-delete-nonpresent-rrset/command 
#!/bin/sh

$PDNSUTIL --config-dir=. --config-name=lmdb list-zone minimal.com 2>&1
echo $?

$PDNSUTIL --config-dir=. --config-name=lmdb delete-rrset minimal.com test A

$PDNSUTIL --config-dir=. --config-name=lmdb list-zone minimal.com 2>&1
echo $?
```

zeha:
> Part of the API failures are explained by `disabled` not working at all in LMDB.

When I last looked at the LMDB code, `disabled` was all wrong on the LMDB side. Not just "not implemented".

pieterlexis:
@Habbie I think we can close this?

Habbie:
> @Habbie I think we can close this?

[...]

PR #10056, opened 2021-02-04
Auth: test the api with more backends

mind04:
### Short description
This pull will add gmysql, pgsql and lmd to the api tests. And fixes some api issues with gpgsql.
Lmdb testing is incomplete but useful for manual testing. Extending the lmdb backend and the tests is for an other pull.

Related #9666

### Checklist
I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


.circleci/config.yml
```diff
@@ -208,7 +208,7 @@ commands:
             libkrb5-3 \
             libldap-2.4-2 \
             liblmdb0 \
-            libluajit-5.1-2 \
+            libluajit-5.1 \
```
Habbie, line 211: why this change?

Habbie (commented):
> Extending the lmdb backend and the tests is for an other pull.

How about telling the test suite to skip some tests for LMDB? That way we can always test the things we do know work.

Habbie:
I'll do a full review soon. Two questions in the review above.

Habbie (approved):

PR #10057, opened 2021-02-04
rec: Get rid of early zone cut computation

rgacogne:
### Short description

This commits removes the early zone cut computation using DS queries.
It instead determines the status of a record when it is either received from an authoritative server or retrieved from a cache, detecting missed zone cuts when the expected signatures are not present.
This results in roughly 8% less outgoing queries in quick testing.

Needs a lot of testing, and there are several low-hanging optimizations that can be done to the new code.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/pdns/recursordist/test-aggressive_nsec_cc.cc
+++ b/pdns/recursordist/test-aggressive_nsec_cc.cc
@@ -290,7 +290,8 @@ BOOST_AUTO_TEST_CASE(test_aggressive_nsec_wildcard_synthesis)
   setDNSSECValidation(sr, DNSSECMode::ValidateAll);
 
   primeHints();
-  /* we first ask a.powerdns.com. | A, get an answer synthesized from the wildcard,
+  /* we first ask a.powerdns.com. | A, get an answer synthesized from the wildcard.
+     We can use it yet because we need the SOA, so let's request a non-existing type
      then check that the aggressive NSEC cache will use the wildcard to synthesize an answer
      for b.powerdns.com */
   const DNSName target("a.powerdns.com.");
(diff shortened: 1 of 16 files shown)


Habbie:
This gives www.icann.org the AD bit it somewhat deserves :)

rgacogne:
Rebased to fix conflicts + a few fixes.

rgacogne:
Did more tests, fixing a few issues in the process. Over the umbrella top1m this PR results in:
- ~17% less outgoing queries
- less DNSSEC validation failures (~19% less in my tests but that depends too much on external variables like which NS you hit during the run, for some domains)
- less CPU (~ %21 less user, ~ 14% less system)
- smaller latency (dnsbulktest reports 14% less on mean response time, 30% less on median, but I would not trust these numbers too much).

rgacogne:
I'll take a better look on Monday but there is something I don't get in the failing DNAME-related tests:
```
Mar 05 17:06:26 [16] dname-insecure.secure.example: Resolved 'secure.example' NS ns.secure.example to: 127.0.0.9
Mar 05 17:06:26 [16] dname-insecure.secure.example: Trying IP 127.0.0.9:53, asking 'dname-insecure.secure.example|A'
Mar 05 17:06:26 [16] dname-insecure.secure.example: Got 5 answers from ns.secure.example (127.0.0.9), rcode=0 (No Error), aa=1, in 0ms
Mar 05 17:06:26 [16] dname-insecure.secure.example: accept answer 'secure.example|SOA|ns1.example.net. hostmaster.example.net. 1 3600 1800 1209600 300' from 'secure.example' nameservers? ttl=15, place=2 YES!
Mar 05 17:06:26 [16] dname-insecure.secure.example: accept answer 'secure.example|RRSIG|SOA 13 2 3600 20210318000000 20210225000000 64723 secure.example. TmXuxYZ/uLo4NebW1+6+4mu9c6mRWpry8inby9gihR6v7hdLWlLTju9g4gYU1Eb+YOOqDiW5DF6ljb+Sv3YLPg==' from 'secure.example' nameservers? ttl=15, place=2 RRSIG - separate
Mar 05 17:06:26 [16] dname-insecure.secure.example: accept answer 'dname-insecure.secure.example|NSEC|dname-secure.secure.example. DNAME RRSIG NSEC' from 'secure.example' nameservers? ttl=15, place=2 YES!
Mar 05 17:06:26 [16] dname-insecure.secure.example: accept answer 'dname-insecure.secure.example|RRSIG|NSEC 13 3 300 20210318000000 20210225000000 64723 secure.example. 1TK2iL7SjfD3kdjn1dPtUimrRYJPjPhuObSlNK8gTFKlj39UeUAVWTXlqVHeDj1Zi114JeXMzDaKGFesrufx2Q==' from 'secure.example' nameservers? ttl=15, place=2 RRSIG - separate
Mar 05 17:06:26 [16] dname-insecure.secure.example: OPT answer '.' from 'secure.example' nameservers
```
Doesn't the NSEC prove that there is a `DNAME` at `dname-insecure.secure.example`? Shouldn't we get a `CNAME` then?

Habbie:
> Mar 05 17:06:26 [16] dname-insecure.secure.example: Trying IP 127.0.0.9:53, asking 'dname-insecure.secure.example|A'

`dname-insecure.secure.example` is not ...

> Doesn't the NSEC prove that there is a `DNAME` at `dname-insecure.secure.example`? Shouldn't we get a `CNAME` then?

.. below `dname-insecure.secure.example`. DNAMEs only generate CNAMEs for any names below them. DNAMEs do nothing at their own name.

rgacogne:
Ah, right! It's too late for me, I guess :)

rgacogne:
I forgot to mention that after this PR we see a bit more responses than before ending up in the 'Indeterminate' bucket. That happens when the resolution failed, for example because of a loop or because all servers are down. The resulting `ServFail` used to end up in the last known zone cut state for that query. Since we no longer determine the zone cuts first, the validation state remains as `Indeterminate` in most cases where we could not resolve.
The new behaviour seems actually more logic to me, though.

omoerbeek:
This withstood all test cases I've thrown at it and works ad advertized: DS queris are very rare with this.
The amount and complexity of the changes in SyncRes are a bit overwhelming, but I'll try to review soon.

omoerbeek:
> The new behaviour seems actually more logic to me, though.

Indeed. We have no data to decide, so `Indeterminate` is appropriate.

omoerbeek:
Continuing the review I did not find stuff that is obviously wrong or even suspect. OTOH, to be honest, I do no grasp the whole algorithm change either, alas, though the general picture is clear. We have to decide if this is worth the risk to incorporate into 4.5

rgacogne:
> OTOH, to be honest, I do no grasp the whole algorithm change either, alas, though the general picture is clear. We have to decide if this is worth the risk to incorporate into 4.5

Can you think of anything (code change, documentation, tests, ...) that would ease your mind? Unfortunately, due to just-in-time validation, validation can happen in several places (processing the response, fetching records from one of the caches, noticing that a response is a NXDOMAIN/NXQTYPE, ...). We could reduce that a bit by refactoring part of the code but that might actually make that PR scarier :)

omoerbeek (approved):

PR #10063, opened 2021-02-05
dnsdist: Add option to spoofRawAction to spoof multiple answers

tjikkun:
### Short description
This adds the option to specify multiple responses with spoofRawAction

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdist-console.cc
+++ b/pdns/dnsdist-console.cc
@@ -667,7 +667,7 @@ const std::vector<ConsoleKeyword> g_consoleKeywords{
   { "SNMPTrapResponseAction", true, "[reason]", "send an SNMP trap, adding the optional `reason` string as the response description"},
   { "SpoofAction", true, "ip|list of ips [, options]", "forge a response with the specified IPv4 (for an A query) or IPv6 (for an AAAA). If you specify multiple addresses, all that match the query type (A, AAAA or ANY) will get spoofed in" },
   { "SpoofCNAMEAction", true, "cname [, options]", "Forge a response with the specified CNAME value" },
-  { "SpoofRawAction", true, "raw [, options]", "Forge a response with the specified record data as raw bytes" },
+  { "SpoofRawAction", true, "raw|list of raws [, options]", "Forge a response with the specified record data as raw bytes. If you specify multiple raws (it is assumed they match the query type), all will get spoofed in" },
   { "SuffixMatchNodeRule", true, "smn[, quiet]", "Matches based on a group of domain suffixes for rapid testing of membership. Pass true as second parameter to prevent listing of all domains matched" },
   { "TagRule", true, "name [, value]", "matches if the tag named 'name' is present, with the given 'value' matching if any" },
   { "TCAction", true, "", "create answer to query with TC and RD bits set, to move to TCP" },
--- a/pdns/dnsdist-lua-actions.cc
+++ b/pdns/dnsdist-lua-actions.cc
@@ -540,21 +540,29 @@ DNSAction::Action SpoofAction::operator()(DNSQuestion* dq, std::string* ruleresu
   uint16_t qtype = dq->qtype;
   // do we even have a response?
   if (d_cname.empty() &&
-      d_rawResponse.empty() &&
+      d_rawResponses.empty() &&
       d_types.count(qtype) == 0) {
     return Action::None;
   }
 
   vector<ComboAddress> addrs;
+  vector<std::string> rawResponses;
   unsigned int totrdatalen = 0;
   uint16_t numberOfRecords = 0;
   if (!d_cname.empty()) {
     qtype = QType::CNAME;
     totrdatalen += d_cname.getStorage().size();
     numberOfRecords = 1;
-  } else if (!d_rawResponse.empty()) {
-    totrdatalen += d_rawResponse.size();
-    numberOfRecords = 1;
+  } else if (!d_rawResponses.empty()) {
+    rawResponses.reserve(d_rawResponses.size());
+    for(const auto& rawResponse : d_rawResponses){
+      totrdatalen += rawResponse.size();
+      rawResponses.push_back(rawResponse);
+      ++numberOfRecords;
+    }
+    if (rawResponses.size() > 1) {
+      shuffle(rawResponses.begin(), rawResponses.end(), t_randomEngine);
+    }
   }
   else {
     for(const auto& addr : d_addrs) {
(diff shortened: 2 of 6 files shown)


pdns/dnsdist-lua-actions.cc
```diff
@@ -540,21 +540,28 @@ DNSAction::Action SpoofAction::operator()(DNSQuestion* dq, std::string* ruleresu
...
   // do we even have a response?
   if (d_cname.empty() &&
-      d_rawResponse.empty() &&
+      d_rawResponses.empty() &&
       d_types.count(qtype) == 0) {
     return Action::None;
   }
 
   vector<ComboAddress> addrs;
+  vector<std::string> rawResponses;
```
rgacogne, line 549: We could use a vector of pointers to string to avoid copying the content. Not sure we care much.

pdns/dnsdist-lua-actions.cc
```diff
@@ -540,21 +540,28 @@ DNSAction::Action SpoofAction::operator()(DNSQuestion* dq, std::string* ruleresu
...
   uint16_t numberOfRecords = 0;
   if (!d_cname.empty()) {
     qtype = QType::CNAME;
     totrdatalen += d_cname.getStorage().size();
     numberOfRecords = 1;
-  } else if (!d_rawResponse.empty()) {
-    totrdatalen += d_rawResponse.size();
-    numberOfRecords = 1;
+  } else if (!d_rawResponses.empty()) {
+    for(const auto& rawResponse : d_rawResponses){
```
rgacogne, line 557: We could call `rawResponses.reserve(d_rawResponses.size());` to make sure we do only one allocation, no matter the number of items.

pdns/dnsdist-lua-actions.cc
```diff
@@ -1728,8 +1740,17 @@ void setupLuaActions(LuaContext& luaCtx)
       return ret;
     });
 
-  luaCtx.writeFunction("SpoofRawAction", [](const std::string& raw, boost::optional<responseParams_t> vars) {
-      auto ret = std::shared_ptr<DNSAction>(new SpoofAction(raw));
+  luaCtx.writeFunction("SpoofRawAction", [](boost::variant<std::string,vector<pair<int, std::string>>> inp, boost::optional<responseParams_t> vars) {
+      vector<string> raws;
+      if(auto s = boost::get<std::string>(&inp))
```
rgacogne, line 1745: Just a nit but I'd rather always have braces, it prevents mistakes if someone later wants to add a second instruction in that `if` case.

pdns/dnsdist-lua-actions.cc
```diff
@@ -1728,8 +1740,17 @@ void setupLuaActions(LuaContext& luaCtx)
...
+        raws.push_back(*s);
+      else {
+        const auto& v = boost::get<vector<pair<int,std::string>>>(inp);
+        for(const auto& raw: v)
```
rgacogne, line 1749: Same.

pdns/dnsdist.cc
```diff
@@ -735,7 +735,9 @@ static void spoofResponseFromString(DNSQuestion& dq, const string& spoofContent,
   string result;
 
   if (raw) {
-    SpoofAction sa(spoofContent);
+    std::vector<std::string> raws;
+    stringtok(raws, spoofContent, " ,");
```
rgacogne, line 739: Perhaps it would be better to only split on "," and not on " "? I'm not sure it really matters, though.

regression-tests.dnsdist/test_Spoofing.py
```diff
@@ -357,6 +359,50 @@ def testSpoofRawAction(self):
             self.assertEquals(expectedResponse, receivedResponse)
             self.assertEquals(receivedResponse.answer[0].ttl, 3600)
 
+    def testSpoofRawActionMulti(self):
+        """
+        Spoofing: Spoof a response from raw bytes
```
rgacogne, line 364:
```suggestion
        Spoofing: Spoof a response from several raw bytes
```
Just to prevent having two tests with the same description, so it's less confusing if that test fails :-)

rgacogne (approved):
Nice work! A few nits but nothing that would prevent a merge as it is.

pdns/dnsdist-console.cc
```diff
@@ -667,7 +667,7 @@ const std::vector<ConsoleKeyword> g_consoleKeywords{
   { "SNMPTrapResponseAction", true, "[reason]", "send an SNMP trap, adding the optional `reason` string as the response description"},
   { "SpoofAction", true, "ip|list of ips [, options]", "forge a response with the specified IPv4 (for an A query) or IPv6 (for an AAAA). If you specify multiple addresses, all that match the query type (A, AAAA or ANY) will get spoofed in" },
   { "SpoofCNAMEAction", true, "cname [, options]", "Forge a response with the specified CNAME value" },
-  { "SpoofRawAction", true, "raw [, options]", "Forge a response with the specified record data as raw bytes" },
+  { "SpoofRawAction", true, "raw|list of raws [, options]", "Forge a response with the specified record data as raw bytes. If you specify multiple raws, all will get spoofed in, it is assumed they match the query type" },
```
jsoref, line 670: I feel like this is closer to the comment structure for `SpoofAction`:
```suggestion
  { "SpoofRawAction", true, "raw|list of raws [, options]", "Forge a response with the specified record data as raw bytes. If you specify multiple raws (it is assumed they match the query type), all will get spoofed in" },
```

pdns/dnsdist-lua-actions.cc
tjikkun: I would if I knew how :)

pdns/dnsdist-lua-actions.cc
tjikkun: Added, and as a bonus for you I also updated the place where I copy/pasted it from ;)

pdns/dnsdist-lua-actions.cc
tjikkun: added

pdns/dnsdist.cc
tjikkun: Agree, changed

pdns/dnsdist-console.cc
tjikkun: changed

pdns/dnsdist-lua-actions.cc
rgacogne: Let's keep that version, it's safer without juggling with pointers :-)

rgacogne (approved):

PR #10064, opened 2021-02-05
rec: Return current rcode instead of 0 if there are no CNAME records to follow

### LINKED ISSUE #9547: recursor lua: setting rcode NXDOMAIN lost when followupFunction is followCNAMERecords
- Program: Recursor 4.3.0
 - Issue type: Bug report

### Short description

```
function preresolve(dq)
dq.rcode = pdns.NXDOMAIN
dq.followupFunction = "followCNAMERecords"
dq.variable = true
return true
end
```

returns

```
;; ->>HEADER<<- opcode: QUERY; status: NOERROR; id: 62032
;; Flags: qr rd ra; QUERY: 1; ANSWER: 0; AUTHORITY: 0; ADDITIONAL: 1
```

If I disable the followupFunction line it correctly returns `NXDOMAIN`.

I would have expected that if there were no CNAMEs in the response the followupFunction should not change the other parameters specified?

omoerbeek:
Note that this is a change in behavior. While it is for the good and fixes #9547, it might be existing code depends on the old 0 value...

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/lua-recursor4.cc
+++ b/pdns/lua-recursor4.cc
@@ -664,7 +664,7 @@ loop:;
     
     if(!dq.followupFunction.empty()) {
       if(dq.followupFunction=="followCNAMERecords") {
-        ret = followCNAMERecords(dq.records, QType(dq.qtype));
+        ret = followCNAMERecords(dq.records, QType(dq.qtype), ret);
       }
       else if(dq.followupFunction=="getFakeAAAARecords") {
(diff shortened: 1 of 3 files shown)


rgacogne (approved):
Code looks good. While I kind of agree that it is a bit weird to call `followCNAMERecords` when there is no `CNAME`, doing nothing (ie not changing the `rcode`) in that case seems better than the current behaviour.

PR #10066, opened 2021-02-05
Spell check 0.0.17a

jsoref:
### Short description

Upgrade the spell checking action

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### Plurals and paste tenses
In the past, plurals `foo`+`s` and past tenses `foo`+`ed` were automatically tolerated. This turned out to be a bad design choice on my part.

The basic example is that `potatos` would sometimes be treated as a mistake and sometimes not (depending on the presence of `potato`).

This is also why `transfered`/`crashs` were accepted as a word in the past, so `transfer` + `ed` / `crash` + `s` were acceptable).

The downside is that plurals of terms of art have to be added individually. Given that the bot helps out w/ this (and the next release should let you have the bot do it for you), this hasn't turned out to be a huge cost.

### Pull requests

When GitHub initially introduced GitHub Actions, the event for `pull_request` was created without enough permission for a tool like this to work properly. In 2020, they introduced a replacement event `pull_request_target` which has enough permission (and thus works better).

### Validation Steps Performed
I've created a temporary repo which is using this as its default branch.

There's a sample PR here:
https://github.com/check-spelling/pdns/pull/1

If people want, they're welcome to create PRs into that repo to test (you can also test in your own forks).


docs/dnsupdate.rst
```diff
@@ -382,7 +382,7 @@ via the domainmetadata table:
 
 This will:
 
-1. Add the 'dhcpdupdate' key to our PowerDNSinstallation
+1. Add the 'dhcpdupdate' key to our PowerDNS installation
```
jsoref, line 385: This fix is because @Habbie spotted the typo while reviewing the expect file changes in this PR

Habbie (approved):

Habbie (changes requested):
Just noticed this disables merge testing, which is not good.

Habbie (approved):
After more discussion, it turns out testing merge commits is hard to do without extra magic under the new `pull_request_target` event, which does make everything else easier, so we're accepting the head vs. merge race condition (which would still be present, at lower incidence, if this check in fact did test the merge commit).

Issue #10024, opened 2021-01-27
Support for Multi-Arch Docker Images


james-crowley:
- Program: Authoritative, Recursor, dnsdist
 - Issue type: Feature request

### Short description
I would like to see the support for multi-arch images as PowerDNS starts to build out their Docker image offerings. Specifically, support for `s390x` and `ppc64le` would be extremely helpful.  

### Usecase
I want to be able to deploy and utilize PowerDNS in any environment/system without worrying about the architecture I am running on. I do a lot of work on Linux on Z(s390x) and having a `s390x` image would be helpful.

### Description
I helped multiple open source communities with enabling multi-arch support for their Docker images using manifest. There are a couple build tools that make the heavy lifting easy but the Dockerfiles need to be design to be architecture aware.

I saw two issues/PRs open talking about multi-arch support:

#10016
#8655

@Habbie, you seem to be involved a lot in the Docker development and also enabling multiple architectures. Would supporting `s390x` and `ppc64le` be something you could be open too?

Additionally, if you need access to `s390x` and `ppc64le` resources there are multiple ways I can assist with that. I did see mentioned of Travis CI in, #8655 , but do not see a lot use of the CI/CD in this repo. How are builds currently be handled?

Habbie:
> I saw two issues/PRs open talking about multi-arch support:

As discussed on #10016, those are not actually about the Docker images, although the work sometimes interleaves a bit!

>  There are a couple build tools that make the heavy lifting easy but the Dockerfiles need to be design to be architecture aware.

Help with that would be much appreciated.

> @Habbie, you seem to be involved a lot in the Docker development and also enabling multiple architectures. 

Yes, both things somehow ended up mostly on my plate.

> Would supporting `s390x` and `ppc64le` be something you could be open too?

Yes, although it depends on the effort from our side, and the maintenance burden. Somewhat related to that, one of the goals of the work in #10016 (that helps us migrate armhf builds off of qemu) is reducing our build times from six hours (armhf in qemu) to 20-30 minutes (armhf on hardware arm64), and we'd prefer that adding platforms does not bring back the six hour builds.

> Additionally, if you need access to `s390x` and `ppc64le` resources there are multiple ways I can assist with that.

That would definitely solve the 'qemu' problem!

> I did see mentioned of Travis CI in, #8655 but do not see a lot use of the CI/CD in this repo. How are builds currently be handled?

Broadly: https://github.com/PowerDNS/pdns/wiki/Automatic-testing

[...]

PR #9964, opened 2021-01-12
authoritative: update debian packaging

zeha:
### Short description
Import from Debian bullseye, with the following changes:
- keep PowerDNS Maintainer:
- drop debhelper compat to 12 (causes no changes)
- ignore changelog, patches

Should also go to the 4.4 branch if possible.

### Checklist

I have:
- [x] read the [CONTRIBUTING.md](https://github.com/PowerDNS/pdns/blob/master/CONTRIBUTING.md) document
- [x] compiled this code
- [ ] tested this code
- [ ] included documentation (including possible behaviour changes)
- [ ] documented the code
- [ ] added or modified regression test(s)
- [ ] added or modified unit test(s)

PR #10067, opened 2021-02-05
Added Support for Architecture aware Dockerfiles

james-crowley:
### Short description
Modifies the existing Dockerfiles to allow for multiple architectures to be build without modification. Needed to make small changes to the Dockerfiles and slight modifications to the control files for the various packages. 

### Checklist
I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

Some of the modifications came from #10024. I do not think that issue should be closed just yet as we need to build out a way to build the manifest files. This PR just removes the issues with the existing Dockerfiles to allow for multiple architectures. 

My testing was building the containers on multiple architectures: `s390x` and `ppc64le`.

I want to also thank @Habbie and @pieterlexis for help debugging the issues I ran into.


james-crowley:
Please note, currently building Dockerfiles on master is broken but not do to my changes. Looks like something has gone wrong with the included sql schema files.

Which causes:

```
Step 38/42 : RUN sqlite3 /var/lib/powerdns/pdns.sqlite3 < /usr/local/share/doc/pdns/schema.sqlite3.sql
 ---> Running in f8cfd7bf3269
/bin/sh: 1: cannot open /usr/local/share/doc/pdns/schema.sqlite3.sql: No such file
The command '/bin/sh -c sqlite3 /var/lib/powerdns/pdns.sqlite3 < /usr/local/share/doc/pdns/schema.sqlite3.sql' returned a non-zero code: 2
```

Habbie:
> Please note, currently building Dockerfiles on master is broken but not do to my changes. Looks like something has gone wrong with the included sql schema files.

That's weird, especially as we added the Docker builds to GH Actions today and those are passing. What revision?

james-crowley:
@Habbie It looks like it was just from master, maybe the most recent commit? If you look at your CircleCI tests, https://github.com/PowerDNS/pdns/pull/10067/checks?check_run_id=1842380066, you can see that exact same issue.

Habbie:
Found it, it's from this PR:

```
/bin/sh: 1: [[: not found
```

and then later
```
configure: Built-in modules: bind gmysql random
configure: Dynamic modules: pipe
```

That's the default set of modules. It looks like your trick to use bash did not work, and sh truncated the configure line at the expression it did not understand.

Habbie:
> sh truncated the configure line at the expression it did not understand.

No, bash only executed the first word (`./configure`), perhaps you need to quote the full argument to `-c` to fix both issues :)

Habbie:
It might also be cleaner/easier to read to use a separate step (or part of the `&&` chain) to set a LUAVER var?

james-crowley:
@Habbie I had to switch to a fully `sh` implementation. Adding `/bin/bash -c` did not seem to work. As I mentioned before the shell "magic" is strong :)

The command now looks like: `$([ -z "${NO_LUA_JIT##*$(uname -m)*}" ] && echo 'lua5.3' || echo 'luajit')`. 

If you want me set to `LUAVER` var in the previous step, I can and then just use `--with-lua=${LUAVER}`.

Habbie:
> If you want me set to `LUAVER` var in the previous step, I can and then just use `--with-lua=${LUAVER}`.

Yes please :)

james-crowley:
@Habbie Looks green/passed on the Docker builds!

Habbie (approved):

Dockerfile-auth
```diff
@@ -43,14 +45,15 @@ RUN if [ "${DOCKER_FAKE_RELEASE}" = "YES" ]; then \
 
 # simplify repeated -C calls with SUBDIRS?
 RUN mkdir /build && \
+    LUAVER=$([ -z "${NO_LUA_JIT##*$(uname -m)*}" ] && echo 'lua5.3' || echo 'luajit') && \
```
zeha, line 48: `uname -m` is bad when cross compiling. Might not be relevant in docker, but before anyone copies this elsewhere.

Also, if you can ensure luajit is not installed (Build-Conflicts), you don't need LUAVER at all.

Habbie: > `uname -m` is bad when cross compiling. Might not be relevant in docker, but before anyone copies this elsewhere.

Ah,  we should reuse the `$(dpkg-architecture -q DEB_BUILD_GNU_TYPE)` (perhaps a different value) for that then.

james-crowley: @zeha Switch `uname -m` to something like `dpkg --print-architecture` is fine. I am not sure what `dpkg --print-architecture` produces on `arm64`. That might be worth checking out so I can update my PR. @Habbie Can you confirm what `dpkg --print-architecture` on `arm64` produces?

Just an FYI `dpkg --print-architecture` produces `ppc64el` not `ppc64le`. It's not an issue for this statement as `ppc64le` has a Lua JIT. 

Another thing is `dpkg` works fine on `debian`/`ubuntu` based systems, but installing `dpkg` on other distros like `centos` or `ubi` do not make sense. Is PDNS going to expand their docker image offerings? Or just keep it to `debian` for now? I did see some interest in `alpine`.

Habbie:
```
peter@buildbot-worker-arm64-1:~$ dpkg --print-architecture
arm64
```

Habbie: We have no plans to support anything other than Debian for our Docker images.

james-crowley: Just replaced `uname -m` with `dpkg --print-architecture`. With that I had to modify the `NO_LUA_JIT` variable.

builder-support/debian/recursor/debian-buster/control
```diff
@@ -8,8 +8,8 @@ Build-Depends: debhelper (>= 10),
                dh-systemd [linux-any],
                libboost-all-dev,
                libcap-dev,
-               libluajit-5.1-dev [!arm64],
-               liblua5.3-dev [arm64],
+               libluajit-5.1-dev [!arm64 !s390x],
```
Habbie, line 11: I prefer this form: https://github.com/PowerDNS/pdns/pull/9964/files#diff-3c0edaabae71305c965b62ac80ffa36de75c6919d543bd0d1dbee6db824731efR19-R20, can you copy that?

Habbie: @zeha tells me he's still experimenting with the right form.

Habbie: I'll merge this, @zeha has accepted that he'll get a small merge conflict.

james-crowley: Just an FYI, you would need to include `ppc64el` in the example you posted. PowerPC has a JIT but it looks like the other supported architectures, s390x and arm64, do not.

Habbie: I expect #9964 to land soon with the exactly right predicates :)

zeha: @james-crowley do you know if luajit _actually_works_ on ppc64el?

james-crowley: @zeha Yup! I built the `auth` Dockerfile and ran the container. Had no issues. I think the confusion might be because its not "official". But a lot of distros thought it was of good enough quality and has included into their packages/repos. 

From reading the `luajit` repo, it looks like the dev is willing to support other architectures but does have the time/money to do so. So there are unofficial patches/branches for various architectures like `ppc64el` and `s390x`. Not sure why `s390x` did not get included into the distro's packages/repos, maybe due to lack of testing?

PR #10073, opened 2021-02-09
dnsdist: Add 'spoof' and 'spoofRaw' Lua bindings

rgacogne:
This commits makes it possible to spoof multiple values from Lua:

```
function myfunc(dq)
  local ret = { newCA("192.0.2.1"), newCA("192.0.2.2") }
  dq:spoof(ret)
  return DNSAction.HeaderModify
end

addAction(AllRule(), LuaAction(myfunc))
```

<strike>Barely tested, need tests and documentation.</strike>
Now with documentation and tests courtesy of tjikkun, thanks a lot!

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


rgacogne:
It might be better to take a variant instead of having two functions, by the way.

PR #10078, opened 2021-02-11
add CSYNC type

### LINKED ISSUE #9802: auth: Add support for CSYNC record type
- Program: Authoritative
 - Issue type: Feature request

### Short description
It would be great if pdns auth supported CSYNC records (RFC 7477).

### Usecase
In combination with RFC 8901 (which requires no additional support in pdns), CSYNC records facilitate transferring authority over a zone to another DNS provider without incurring any insecure time period.

### Description
RFC 7477 introduces the CSYNC record type, which contains a serial field, some flags, and a record type bit map (as in ~RRSIG~ NSEC). It would be great if such records could be provisionen in pdns auth, including through the API.

@ulrichwisser @nils-wisiol

Habbie:
### Short description
Has one test which passes. Having to implement (well, steal from NSEC3) all those methods makes me wish we had native NSECBitmap (should we rename that?) support in the conv* functions.

Fixes #9802.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


Habbie:
> Having to implement (well, steal from NSEC3) all those methods makes me wish we had native NSECBitmap (should we rename that?) support in the conv* functions.

I added `xfrNSECBitmap` so CSYNC is a boilerplate conversion now, but NSEC/NSEC3 will need an additional `xfr` for blobs with length prefixes.

pieterlexis:
It would be good to mention in the (eventual) docs that the SOA in the database might not the be one in the zone due to SOA-EDIT and that operators should be aware of this

Habbie:
This is failing because dnsdist is lacking symlinks and Makeflie entries for the new files and some other dependencies. I'll see if I can reorganise so that that is not necessary.

Habbie:
> This is failing because dnsdist is lacking symlinks and Makeflie entries for the new files and some other dependencies. I'll see if I can reorganise so that that is not necessary.

That is one big yak stack. i've stored the refactor at https://github.com/Habbie/pdns/pull/new/csync-refactor and will undo it here.

Habbie:
rebased

pieterlexis (approved):

PR #10087, opened 2021-02-16
pdnsutil: Warn on CNAME targets for NS, MX and SRV

### LINKED ISSUE #10085: pdnsutil check-zone doesn’t complain about MX pointing to CNAME record
- Program: Authoritative
 - Issue type: Bug report

### Short description

`pdnsutil check-zone` doesn’t complain about MX record that points to a CNAME in the same zone file.

### Environment
cPanel & WHM 94 (PowerDNS 4.3.1) on CentOS 7

### Steps to reproduce
Create a zone file that has:
```
example.com. 14400 IN MX 0 mail.example.com.
mail 14400 IN CNAME example.com.
```
… and run `pdnsutil check-zone` on that domain.

### Expected behaviour
Since this is forbidden per RFC 2181 (10.3), PowerDNS probably ought to give _some_ notice about this.

### Actual behaviour
There’s no mention of it.

### Other information
Per discussion on IRC (timeless, kpfleming) this prohibition may actually be “more honour’d in the breach than the observance”. I personally would like to see it relaxed since adhering to it will significantly complicate my own work.

pieterlexis:
### Short description
Fixes #10085

This only warns, even though having this is an RFC violation. Moderns resolvers and clients should be able to handle this situation just fine.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/pdnsutil.cc
+++ b/pdns/pdnsutil.cc
@@ -324,6 +324,7 @@ static int checkZone(DNSSECKeeper &dk, UeberBackend &B, const DNSName& zone, con
 
   bool hasNsAtApex = false;
   set<DNSName> tlsas, cnames, noncnames, glue, checkglue, addresses, svcbAliases, httpsAliases, svcbRecords, httpsRecords;
+  vector<DNSResourceRecord> checkCNAME;
   set<pair<DNSName, QType> > checkOcclusion;
   set<string> recordcontents;
   map<string, unsigned int> ttl;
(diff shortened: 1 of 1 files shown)


pdns/pdnsutil.cc
```diff
@@ -323,7 +323,8 @@ static int checkZone(DNSSECKeeper &dk, UeberBackend &B, const DNSName& zone, con
 
 
   bool hasNsAtApex = false;
-  set<DNSName> tlsas, cnames, noncnames, glue, checkglue, addresses, svcbAliases, httpsAliases, svcbRecords, httpsRecords;
+  set<DNSName> tlsas, cnames, noncnames, glue, checkglue, addresses, svcbAliases, httpsAliases, svcbRecords, httpsRecords, mxrecords, nsrecords;
```
Habbie, line 326: These two new sets appear to be unused in the current version of the code :-)

Habbie (changes requested):

Habbie (approved):

PR #10088, opened 2021-02-16
dnsdist: Create TCP worker threads before acceptors ones

rgacogne:
### Short description

Otherwise we might crash when the first TCP query arrives because we are dividing by zero when trying to select a worker.
This PR also adds an explicit check to turn the crash into an exception if we ever manage to break that again.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/pdns/dnsdist.cc
+++ b/pdns/dnsdist.cc
@@ -2418,6 +2418,13 @@ int main(int argc, char** argv)
     }
     handleQueuedHealthChecks(mplexer, true);
 
+    /* we need to create the TCP worker threads before the
+       acceptor ones, otherwise we might crash when processing
+       the first TCP query */
+    while (!g_tcpclientthreads->hasReachedMaxThreads()) {
+      g_tcpclientthreads->addTCPClientThread();
+    }
+
     for(auto& cs : g_frontends) {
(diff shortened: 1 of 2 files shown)


Habbie (approved):

rgacogne:
Unrelated failure in remotebackend-pipe tests, with not much to investigate:
```
basic-a-resolution
--- ../modules/remotebackend/regression-tests/basic-a-resolution/expected_result	2021-02-16 10:46:16.000000000 +0000
+++ ../modules/remotebackend/regression-tests/basic-a-resolution/real_result	2021-02-16 11:10:45.582292569 +0000
@@ -1,3 +1,2 @@
-0	outpost.example.com.	IN	A	120	192.168.2.1
 Rcode: 0 (No Error), RD: 0, QR: 1, TC: 0, AA: 1, opcode: 0
 Reply to question for qname='outpost.example.com.', qtype=A
./start-test-stop: line 192: kill: (12520) - No such process

Exited with code exit status 1

CircleCI received exit code 1
```

Habbie:
> Unrelated failure in remotebackend-pipe tests, with not much to investigate:

I've seen similar, very seldom, failures in the remotebackend. I haven't found why yet.

Issue #10058, opened 2021-02-04
Counters and handling for intentionally bogus domain list (DNSSEC)


johnhtodd:
- Program: Recursor
 - Issue type: Feature request

### Short description
Add separate counter for "bogus bogus" DNSSEC failures, so appropriate math can be applied and real DNSSEC failure rates can be computed.

### Usecase
We would like to be able to see in our reporting tools the actual number of bogus domains, excluding intentionally bogus domains that are being used by third-party testing systems. Keeping separate counters for these "bogus bogus" domains would allow our alerting and display systems to compute real figures.

### Description
Many DNSSEC bogus domains sent to our resolvers for resolution are "intentional" bogus records, from zones like dotnxdomain.net, ripe-hackathon2.nlnetlabs.nl, and rootcanary.net.  While these test domain lookups are still bogus, they are corrupting our ability to see actual trends in real bogus domain counts as they become (percentage-wise) a significant portion if not the majority of our DNSSEC failures.

We would like to have a separate set of counters that allow matching against a static list of suffix domains (or more flexibly, a regular expression) that would allow us to track the volume of this sub-set of bogus domains.  We would then be able to do a subtraction of this sub-set value from the total counters for dnssec bogus failures.  This would not change the logic of the existing counters at all; it would simply add a new counter that allows post-processing to be applied to derive more sane numbers.

It may be useful also to tag those "bogus bogus" matches in logging output, so that downstream logging systems would not have to keep a separate list, which risks issues of de-synchronizing.

Items:
 - Adding counters for "bogus bogus" domains matching the list
 - The ability to list these "bogus bogus" domains via a command
 - Tagging "bogus bogus" queries on logging output for differentiation

johnhtodd:
Please, for the love of all things holy, find some other term than "bogus bogus" if this actually gets implemented.

PR #10089, opened 2021-02-16
rec: Count DNSSEC stats for some names in a different set of counters

omoerbeek:
### Short description

This implements the feature as requested in #10058.

Idea is that you can set `x-dnssec-names` to a suffixmatch list. DNSSEC counting actions will then be done for on `x-dnssec-result-bogus-...` counter instead of the regular one.

Missing: 
- [ ] Adding suffixes using `rec_control`.
- [x] Marking of log lines as requested in the feature request.
- [x] Docs (settings + metrics)

This PR is mainly to gauge if this approach is the way forward. I'm also open for a better name than `x-dnssec...`

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/pdns_recursor.cc
+++ b/pdns/pdns_recursor.cc
@@ -1847,10 +1847,6 @@ static void startDoResolve(void *p)
             }
           }
 
-          if(sr.doLog()) {
-            g_log<<Logger::Warning<<"Starting validation of answer to "<<dc->d_mdp.d_qname<<"|"<<QType(dc->d_mdp.d_qtype).getName()<<x_marker<<" for "<<dc->getRemote()<<endl;
-          }
-
           if(state == vState::Secure) {
             if(sr.doLog()) {
               g_log<<Logger::Warning<<"Answer to "<<dc->d_mdp.d_qname<<"|"<<QType(dc->d_mdp.d_qtype).getName()<<x_marker<<" for "<<dc->getRemote()<<" validates correctly"<<endl;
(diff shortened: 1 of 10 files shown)


omoerbeek:
Note that I count in one set or another, not both as suggested by the feature request.
Also, I might consider adding the `x-` metrics only if the corresponding setting is non-empty.

omoerbeek:
We decidd to leave the `rec_control` manipulation of `x-dnssec-names` out for now.

pdns/pdns_recursor.cc
```diff
@@ -1835,26 +1837,34 @@ static void startDoResolve(void *p)
...
+          if (sr.doLog() || vStateIsBogus(state)) {
+            auto xdnssec = g_xdnssec.getLocal();
+            if (xdnssec->check(dc->d_mdp.d_qname)) {
+              x_marker = " [in x-dnssec-names]";
+            }
           }
 
-          auto state = sr.getValidationState();
+          if(sr.doLog()) {
+            g_log<<Logger::Warning<<"Starting validation of answer to "<<dc->d_mdp.d_qname<<"|"<<QType(dc->d_mdp.d_qtype).getName()<<x_marker<<" for "<<dc->getRemote()<<endl;
```
rgacogne, line 1851: I think this line should go, actually. It used to make sense in 4.0 when we were doing validation here, after everything else, but I think it does not make sense anymore.

rgacogne (approved):
Looks good to me! I wonder if we will actually use the separate metrics, or if we could get away with not accounting the validation states for the names in that list at all. On the other hand we have often been very happy to have additional metrics when investigating an issue, and in that case it could for example make it clear that a mistake has been made by adding the wrong name to the "separate" list, for example.

PR #10103, opened 2021-02-18
dnsdist: Add a "TLS sessions management" guide

rgacogne:
### Short description

Add more words about how TLS sessions are handled in dnsdist, especially with regard to TLS Session Ticket Keys (STEKs).

### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code


pdns/dnsdistdist/docs/guides/tls-sessions-management.rst
```diff
@@ -0,0 +1,56 @@
+TLS Sessions Management
+=======================
+
+TLS sessions
+------------
+
+One of the most costly TLS operation if the negotiation of a new session, since both the client and the server need to generate and agree on cryptographic materials. In order to reduce that cost, TLS implements what is called session resumption, where a client opening a new connection to a server can reuse the cryptographic materials negotiated for a previous TLS session.
```
mnordhoff, line 7:
```suggestion
One of the most costly TLS operation is the negotiation of a new session, since both the client and the server need to generate and agree on cryptographic materials. In order to reduce that cost, TLS implements what is called session resumption, where a client opening a new connection to a server can reuse the cryptographic materials negotiated for a previous TLS session.
```
Typo, if -> is

pdns/dnsdistdist/docs/guides/tls-sessions-management.rst
```diff
@@ -0,0 +1,56 @@
...
+
+It is also possible to manually request a STEK rotation using the :func:`getDOHFrontend` (DoH) and :func:`getTLSContext` (DoT) functions to retrieve the bind object, and calling its ``rotateTicketsKey`` method (:meth:`DOHFrontend.rotateTicketsKey`, :meth:`TLSContext.rotateTicketsKey`).
+
+The default settings should be fine for most deployments, but generating a random key for every dnsdist instance will not allow resuming the session from a different instance in a cluster. In that case it is possible to generate the STEK outside of dnsdist, write it to a file on a non-persistent storage, distribute it to all instances using something like rsync over SSH, and load that file from dnsdist.
```
omoerbeek, line 26: "on a non-persistent storage" is confusing. Files are normally persistent and since you need the bytes when starting up, shouldn't it be on persistent storage?

omoerbeek: I do think it should be mentioned the STEK file is sensitive data and should be protected from access by unauthorized users.

rgacogne: Ideally, no, the STEK should only be written to a tmpfs or equivalent.

pdns/dnsdistdist/docs/guides/tls-sessions-management.rst
```diff
@@ -0,0 +1,56 @@
...
+
+Nowadays pretty much all clients support the second option, TLS tickets, where the need information is signed and encrypted by the server before being sent to the client, which is responsible for storing it and sending it back when it wants to establish a new session. That reduces the burden of the server while providing the same benefits.
+
+The server uses Session Ticket Encryption Key (STEK) to sign and encrypt the information sent to the client, making it possible to ensure that it is genuine and has not been tampered when the client provides it later. That STEK can be shared by all dnsdist instances in the same cluster, making it possible for any server to resume a session initially generated by a different server.
+
+Knowing the STEK is all the information needed to be able to decrypt a live TLS session, but also a recorded one, so it is very important to keep that key well-protected. It should never be exchanged in clear-text, and ideally should not be written to persistent storage but be kept in a tmpfs with no swap configured. It should also be regularly rotated to preserve TLS' forward secrecy properties.
+
+Keys management in dnsdist
+--------------------------
+
```
omoerbeek, line 21: dnsdist only support TLS t tickets, right? Please mention that if so.

rgacogne: It supports both server-side sessions and client-side tickets, although my advice is to only use the latter.

pdns/dnsdistdist/docs/guides/tls-sessions-management.rst
mnordhoff: Example bad idea: For an unrelated service, I have a hacky script that creates or rotates STEKs in `/run/foo`, executed by a systemd timer and in the service's 'ExecStartPre`.

omoerbeek: > Ideally, no, the STEK should only be written to a tmpfs or equivalent.

I agree in principle, but this does present a bit of a challenge, since you do need to generate, distribute a make it available to all dnsdist instances in a cluster. If you want to be able to reboot members of a cluster they either need to have it stored locally or retrieve it again from a central location so the central location should have it available, likely even after you reboot the central location. Or you need to accept that it is lost after the central location is rebooted...

rgacogne: Yes? I agree it's difficult and there is a trade-off between security (not losing confidentiality for too many sessions when the STEK is compromised) and availability (not being able to resume sessions if the current STEK is lost will increase the load on the server for a while) that everyone needs to consider for themselves.

omoerbeek: Yes, I'm rambling too much... , since this is very much dependent on the environment I'd say leave at "be careful with the STEK as it is sensitive data".

PR #10106, opened 2021-02-19
Rec: refactor latency and histogram calculations

omoerbeek:
### Short description

This intended as code cleanup without changing externally observable behavior. It will make it easier to make the HIstogram configurable and add cumulative ones in the future.

Current latency computations are a mix of different types (float, double, uint64_t) and units (secs, millisecs, microsecs. Harmonize to always work in an integral number of microseconds, and only use doubles when needed to compute running averages.

Also make sure we use atomics for all data written to from different threads, specifically  `avgLatencyUsec` and `avgLatencyOursUsec`.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [X] documented the code

### DIFF
--- a/pdns/histogram.hh
+++ b/pdns/histogram.hh
@@ -1,0 +1,120 @@
+#pragma once
+
+#include <algorithm>
+#include <atomic>
+#include <vector>
+#include <sstream>
+
+namespace pdns {
+
+// By convention, we are using microsecond units
+struct Bucket
+{
+  std::string d_name;
+  uint64_t d_boundary;
+  uint64_t d_count;
+};
+
+inline bool operator<(uint64_t b, const Bucket& bu)
+{
+  // we are using less-or-equal
+  return b <= bu.d_boundary;
+}
+
+struct AtomicBucket
+{
+  // We need the constrcutors in this case, since atomics have a disabled
+  // copy constructor.
+  AtomicBucket() {}
+  AtomicBucket(std::string name, uint64_t boundary, uint64_t val) : d_name(std::move(name)), d_boundary(boundary), d_count(val) {}
+  AtomicBucket(const AtomicBucket& rhs) : d_name(rhs.d_name), d_boundary(rhs.d_boundary), d_count(rhs.d_count.load()) {}
+
+  std::string d_name;
+  uint64_t d_boundary;
+  std::atomic<uint64_t> d_count;
+};
+
+inline bool operator<(uint64_t b, const AtomicBucket& bu)
+{
+  // we are using less-or-equal
+  return b <= bu.d_boundary;
+}
+
+template<class B>
+class BaseHistogram
+{
+public:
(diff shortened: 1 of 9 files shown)


pdns/histogram.hh
```diff
@@ -0,0 +1,137 @@
...
+  }
+
+  std::vector<B> getCumulativeBuckets() const
+  {
+    std::vector<B> ret;
+    ret.reserve(d_buckets.size());
+    uint64_t c{0};
+    for (const auto& b : d_buckets) {
+      c += b.d_count;
+      ret.push_back(B{b.d_name, b.d_boundary, c});
```
rgacogne, line 103: `emplace_back`?

pdns/histogram.hh
```diff
@@ -0,0 +1,137 @@
...
+{
+  std::string d_name;
+  uint64_t d_boundary{0};
+  uint64_t d_count{0};
+};
+
+inline bool operator<(uint64_t b, const Bucket& bu)
+{
+  // we are using less-or-equal
+  return b <= bu.d_boundary;
```
rgacogne, line 41: I find that a bit confusing. I wonder if it would not be clearer to define a different comparison operator (`lessOrEqual`?) and pass that to `std::upper_bound` and `std::is_sorted`?

pdns/pdns_recursor.cc
```diff
@@ -2389,6 +2367,11 @@ static bool checkForCacheHit(bool qnameParsed, unsigned int tag, const string& d
     }
     g_stats.avgLatencyUsec = (1.0 - 1.0 / g_latencyStatSize) * g_stats.avgLatencyUsec + 0.0; // we assume 0 usec
     g_stats.avgLatencyOursUsec = (1.0 - 1.0 / g_latencyStatSize) * g_stats.avgLatencyOursUsec + 0.0; // we assume 0 usec
+#if 0
+    // XXX changes behaviour compared to old code!
+    g_stats.answers(0);
+    g_stats.ourtime(0);
+#endif
```
rgacogne, line 2374: I'm afraid this change will cause a lot of confusion about a "latency regression" :-) Perhaps we could add a new metric but keep the existing one (but that means double work)? Otherwise I think we need to write some "impossible to miss" upgrade note :)

rgacogne (approved):

pdns/pdns_recursor.cc
omoerbeek: That's why I `#if 0`ed that part... in the future, the plan is to add configurable histograms. Those will be an addition, not a replacement iof the current metrics, as not to disturb.

pdns/histogram.hh
omoerbeek: Not needed for `std::is_sorted`, it's comparing `uints64_t`. Actually it should also check if each bound is unique.

PR #10108, opened 2021-02-19
dnsdist: Fix several bugs in the TCP code path, add unit tests

rgacogne:
### Short description

This PR fixes several issues in corner cases, introduced during the refactoring for out-of-order support. It also adds a lot of unit tests for that TCP code, which was planned for 1.7.0 but fast-forwarded because of issues reported with 1.6.0-alpha1.

It also adds a new configuration option, `setTCPInternalPipeBufferSize`, that can be used to set the size of the buffer of the internal pipe used to pass queries to TCP worker threads, thus allowing a smoother handling of incoming TCP connections peaks.

The diff seems huge but most of that (3200 out of 3900 additions) comes from the new unit tests, which actually amount to more lines of code than the TCP code itself and provides a very nice code coverage (which unfortunately does not prevent bugs from still hiding in a unexpected state.. ;-)).

One caveat is that this drops source selection support other than via `SO_BINDTODEVICE`, because having to pass the source interface name and index down to the actual `sendmsg` code is very annoying. Perhaps we should at least look into supporting `IP_SENDIF` in addition to `SO_BINDTODEVICE`?

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/pdns/dnsdistdist/test-dnsdisttcp_cc.cc
+++ b/pdns/dnsdistdist/test-dnsdisttcp_cc.cc
@@ -2107,14 +2107,15 @@ BOOST_AUTO_TEST_CASE(test_IncomingConnectionOOOR_BackendOOOR)
       { ExpectedStep::ExpectedRequest::readFromBackend, IOState::Done, responses.at(3).size() - 2 },
       /* sending it to the client (4) but short write */
       { ExpectedStep::ExpectedRequest::writeToClient, IOState::NeedWrite, responses.at(3).size() - 1, [&threadData](int desc, const ExpectedStep& step) {
-        /* set the client descriptor as ready */
-        dynamic_cast<MockupFDMultiplexer*>(threadData.mplexer.get())->setReady(desc);
+        /* set the client descriptor as NOT ready */
+        dynamic_cast<MockupFDMultiplexer*>(threadData.mplexer.get())->setNotReady(desc);
       } },
-#warning FIXME? We are trying to write again when coming back from handleQuery, this is a bit useless..
-      { ExpectedStep::ExpectedRequest::writeToClient, IOState::NeedWrite, 0 },
       /* reading a response from the backend (5) */
       { ExpectedStep::ExpectedRequest::readFromBackend, IOState::Done, 2 },
-      { ExpectedStep::ExpectedRequest::readFromBackend, IOState::Done, responses.at(4).size() - 2 },
+      { ExpectedStep::ExpectedRequest::readFromBackend, IOState::Done, responses.at(4).size() - 2, [&threadData](int desc, const ExpectedStep& step) {
+        /* set the client descriptor as ready to resume sending */
+        dynamic_cast<MockupFDMultiplexer*>(threadData.mplexer.get())->setReady(-1);
+      } },
       /* resume sending it to the client (4) */
       { ExpectedStep::ExpectedRequest::writeToClient, IOState::Done, 1 },
       /* sending it to the client (5) */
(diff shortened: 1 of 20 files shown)


pdns/dnsdistdist/test-dnsdisttcp_cc.cc
```diff
@@ -0,0 +1,3210 @@
...
+    BOOST_CHECK_EQUAL(s_writeBuffer.size(), 0U);
+    BOOST_CHECK_EQUAL(s_backendWriteBuffer.size(), query.size());
+    BOOST_CHECK(s_backendWriteBuffer == query);
+
+    /* we need to clear them now, otherwise we end up with dangling pointers to the steps via the TLS context, etc */
+    IncomingTCPConnectionState::clearAllDownstreamConnections();
+  }
+
+  {
+#if 0
```
omoerbeek, line 1551: Intended to be switched off? If so, please add a comment why

omoerbeek:
Nice to see more general TCP FastOpen support. I'm a bit worried about the source binding changes as you are using Linux specific functions. Both alternatives are not generally available.

omoerbeek:
> Nice to see more general TCP FastOpen support. I'm a bit worried about the source binding changes as you are relying on specific functions. Both alternatives are not generally available.

Thinking about this a bit more, the old-fashioned way of binding to a source *address* before connect should still work, right?

pdns/tcpiohandler.hh
```diff
@@ -197,35 +205,89 @@ public:
...
+
+#ifdef MSG_FASTOPEN
+    if (!d_conn && fastOpen) {
+      d_fastOpen = true;
+    }
+    else {
+      SConnectWithTimeout(d_socket, remote, /* no timeout, we will handle it ourselves */ 0);
+    }
+#else
+    SConnectWithTimeout(d_socket, d_ds->remote, /* no timeout, we will handle it ourselves */ 0);
```
omoerbeek, line 253:
```suggestion
    SConnectWithTimeout(d_socket, remote, /* no timeout, we will handle it ourselves */ 0);
```

pdns/tcpiohandler.hh
```diff
@@ -197,35 +205,89 @@ public:
...
+
+#ifdef MSG_FASTOPEN
+    if (!d_conn && fastOpen) {
+      d_fastOpen = true;
+    }
+    else {
+      SConnectWithTimeout(d_socket, remote, timeout);
+    }
+#else
+    SConnectWithTimeout(d_socket, d_ds->remote, timeout);
```
omoerbeek, line 285:
```suggestion
    SConnectWithTimeout(d_socket, remote, timeout);
```

pdns/tcpiohandler.hh
```diff
@@ -297,8 +359,24 @@ public:
       return d_conn->tryWrite(buffer, pos, toWrite);
     }
 
+    if (d_fastOpen) {
```
omoerbeek, line 362:
```suggestion
#ifdef MSG_FASTOPEN
    if (d_fastOpen) {
```

pdns/tcpiohandler.hh
```diff
@@ -297,8 +359,24 @@ public:
...
+        pos += sent;
+      }
+
+      if (pos < toWrite) {
+        return IOState::NeedWrite;
+      }
+
+      return IOState::Done;
+    }
+
```
omoerbeek, line 376:
```suggestion
#endif /* MSG_FASTOPEN */
```

pdns/tcpiohandler.hh
```diff
@@ -367,13 +445,14 @@ public:
...
+  bool getUnknownTicketKey() const
   {
     return d_conn && d_conn->getUnknownTicketKey();
   }
 
 private:
   std::unique_ptr<TLSConnection> d_conn{nullptr};
+  ComboAddress d_remote;
   int d_socket{-1};
   bool d_fastOpen{false};
```
omoerbeek, line 457:
```suggestion
#if defined(MSG_FASTOPEN) || defined(TCP_FASTOPEN_CONNECT)
  bool d_fastOpen{false};
#endif
```

rgacogne:
> Thinking about this a bit more, the old-fashioned way of binding to a source address before connect should still work, right?

Yes, that only concerns source **interface** selection, sorry!

pdns/tcpiohandler.hh
```diff
@@ -106,6 +106,14 @@ protected:
 class TLSFrontend
```
omoerbeek, line 106: Add `#include <netinet/tcp.h>` above, so we have the `TCP_FASTOPEN_CONNECT` connect if there.

omoerbeek:
Something to investigate: I did not have `TCP_FASTOPEN_CONNECT`, but it seems I did not get the fallback to `MSG_FASTOPEN` either.

pdns/tcpiohandler.hh
rgacogne: We only use it for `MSG_FASTOPEN`, even. On it, thanks!

rgacogne:
> Something to investigate: I did not have TCP_FASTOPEN_CONNECT, but it seems I did not get the fallback to MSG_FASTOPEN either.
> 
> The fallback to MSG_FASTOPEN if the include is not there is enabled, but does not seem tot work as intended for sdig using the write call.

Fixed the detection of `TCP_FASTOPEN_CONNECT` and the handling of `MSG_FASTOPEN` in blocking mode (only the non-blocking one was supported). I also added a `fastOpen` option to `sdig`.

pdns/dnsdistdist/test-dnsdisttcp_cc.cc
rgacogne: Oops, I only intended to disable them for a while when I was writing more tests. Fixing, thanks!

omoerbeek:
NIce, when this is merged, I'll make a PR to change the two write calls in sdig into as single one.

omoerbeek (approved):
FastOpen works now as expected in sdig. I do see two writes in captures, but that is expected. I browsed the rest of the changes, and found nothing surprising.

rgacogne:
Rebased to fix some commits not compiling.

PR #10111, opened 2021-02-22
rec: Handle policy (if needed) after postresolve and document the hooks better

### LINKED ISSUE #10080: pdns.DROP doesn't work any more
- Program: Recursor
 - Issue type: Bug report

### Short description
`pdns.DROP` sends malformed reply to the client instead of silently dropping the request.

### Environment
 - Software version: powerdns-recursor 4.4.2-1

The powerdns-recursor was upgraded from 4.3.5-4:

```
[2021-02-12T10:55:56+0200] [ALPM] upgraded powerdns-recursor (4.3.5-4 -> 4.4.2-1)
```

### Steps to reproduce
1. intercept queries like this:

```lua
blocklist_drop = newDS()
blocklist_drop:add(dofile("/etc/powerdns/blocklist-drop.lua"))

function preresolve(dq)
    -- drop explicitly restricted
    if blocklist_drop:check(dq.qname) then
        dq.rcode = pdns.DROP
        return true
    end
    return false
end
```

2. in tcpdump observe that the queries are not dropped, but are replied with weird code:

```
17:29:42.570307 IP 192.168.1.27.43457 > 192.168.1.1.53: 32798+ A? www.stalker.com. (33)
        0x0000:  4500 003d be4b 0000 3f11 39f8 c0a8 011b  E..=.K..?.9.....
        0x0010:  c0a8 0101 a9c1 0035 0029 13a8 801e 0100  .......5.)......
        0x0020:  0001 0000 0000 0000 0377 7777 0773 7461  .........www.sta
        0x0030:  6c6b 6572 0363 6f6d 0000 0100 01         lker.com.....
17:29:42.570557 IP 192.168.1.1.53 > 192.168.1.27.43457: 32798 Resp14 0/0/0 (33)
        0x0000:  4500 003d 74a8 0000 4011 829b c0a8 0101  E..=t...@.......
        0x0010:  c0a8 011b 0035 a9c1 0029 83a7 801e 818e  .....5...)......
        0x0020:  0001 0000 0000 0000 0377 7777 0773 7461  .........www.sta
        0x0030:  6c6b 6572 0363 6f6d 0000 0100 01         lker.com.....
```

### Expected behaviour

In powerdns-recursor 4.3 dropped queries were just dropped. This is what is also expected in 4.4.

### Actual behaviour

In powerdns-recursor 4.4 dropped queries are actually answered with some code.

### Other information

From the IRC chat:

```
[16:05] <post-factum> hi. in pdns-recursor 4.3 setting rcode to pdns.drop caused the query to just be dropped, while in 4.4 this caused something like "17749 Resp14 0/0/0 (33)" (as per tcpdump) to be sent back to client. is this a bug? how to restore old behaviour?
[16:06] <rgacogne> post-factum: how do you set the rcode, exactly?
[16:07] <post-factum> rgacogne: in `function preresolve(dq)` i do a check for domain using `if blocklist_drop:check(dq.qname) then`, and then set rcode with `dq.rcode = pdns.DROP` and then just `return true`
[16:14] <post-factum> rgacogne: if it helps, this is what it looks like in tcpdump: https://paste.debian.net/1185214/
[16:14] <rgacogne> hmm, I might have removed support for that during a cleanup without even being aware it was a thing
[16:15] <post-factum> whoopsie
[16:16] <rgacogne> so doing dq.rcode = pdns.DROP might now set the rcode to a negative value..
[16:17] <post-factum> nonono please bring it back ☺
[16:21] <Habbie> DROP=-2
[16:21] <Habbie> 14 is 16-2
[16:21] <Habbie> :)
[16:22] <ph1> bobby drop dns 
[16:23] <rgacogne> I think it might still be possible to make it work, just a sec
[16:25] <rgacogne> hmm, no, I don't see how, we only check for -2 after a RPZ hit now
[16:26] <post-factum> shall i create a bugreport?
[16:26] <rgacogne> yes, please
[16:26] <post-factum> ok
[16:26] <rgacogne> for now, I'm not sure it's possible to completely drop a query based on the qname without using RPZ
[16:28] <rgacogne> if we do make this work again, it really needs some doc
[16:28] <Habbie> and a test
```

omoerbeek:
Fixes #10080 (which contains a more detailed explanation)

Please review the docs changes extra carefully, I tried to make them more clear, but I don't know if I succeeded.

Maybe the removal of `PolicyDecision` and the corresponding Lua is a bit too radical and should be postponed.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/recursordist/docs/lua-scripting/hooks.rst
+++ b/pdns/recursordist/docs/lua-scripting/hooks.rst
@@ -25,7 +25,7 @@ Once a script is loaded, PowerDNS looks for the interception functions in the lo
 All of these functions are optional.
 
 If ``ipfilter`` returns ``true``, the query is dropped.
-If ``preresolve`` returns ``true``, it will indicate it handled a query, and the recursor will send to result as constructed in the functions to the client.
+If ``preresolve`` returns ``true``, it will indicate it handled a query, and the recursor will send the result as constructed in the function to the client.
 If it returns ``false``, the Recursor will continue processing.
 For the other functions, the return value will indicate that an alteration has been made. In that case DNSSEC validation will be automatically disabled since the content might not be genuine anymore.
 At specific points the Recursor will check if policy handling should take place.
--- a/pdns/recursordist/docs/lua-scripting/hooks.rst
+++ b/pdns/recursordist/docs/lua-scripting/hooks.rst
@@ -24,8 +24,13 @@ It also means that to print such names, the ``:toString`` method must be used (o
 Once a script is loaded, PowerDNS looks for the interception functions in the loaded script.
 All of these functions are optional.
 
-If a function returns true, it will indicate that it handled a query.
-If it returns false, the Recursor will continue processing unchanged (with one minor exception).
+If ``ipfilter`` returns ``true``, the query is dropped.
+If ``preresolve`` returns ``true``, it will indicate it handled a query, and the recursor will send to result as constructed in the functions to the client.
+If it returns ``false``, the Recursor will continue processing.
+For the other functions, the return value will indicate if DNSSEC validation should take place.
+For modified results, the functions should return ``true`` to avoid failing DNSSEC validations.
+At specific points the Recursor will check if policy handling should take place.
+These points are immediatly after ``preresolve``, after resolving and after ``postresolve``.
 
 Interception Functions
 ----------------------
@@ -115,7 +120,7 @@ Interception Functions
     It accepts a single, scalable parameter which can be accessed using FFI accessors.
     Like the non-FFI version, it has the ability to set a tag for the packetcache, policy tags, a routing tag, the :attr:`DNSQuestion.requestorId` and :attr:`DNSQuestion.deviceId`values and to fill the :attr:`DNSQuestion.data` table. It also offers ways to mark the answer as variable so it's not inserted into the packetcache, to set a cap on the TTL of the returned records, and to generate a response by adding records and setting the RCode. It can also instruct the recursor to do a proper resolution in order to follow any `CNAME` records added in this step.
 
-.. function:: prerpz(dq)
+.. function:: prerpz(dq) -> bool
 
   This hook is called before any filtering policy have been applied,  making it possible to completely disable filtering by setting  :attr:`dq.wantsRPZ <DNSQuestion.wantsRPZ>` to false.
   Using the :meth:`dq:discardPolicy() <DNSQuestion:discardPolicy>` function, it is also possible to selectively disable one or more filtering policy, for example RPZ zones, based on the content of the ``dq`` object.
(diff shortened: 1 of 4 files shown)


pdns/recursordist/docs/lua-scripting/hooks.rst
```diff
@@ -238,9 +243,24 @@ A minimal sample script:
 **Warning**: Please do NOT use the above sample script in production!
 Responsible NXDomain redirection requires more attention to detail.
 
-Useful 'rcodes' include 0 for "no error", ``pdns.NXDOMAIN`` for "NXDOMAIN", ``pdns.DROP`` to drop the question from further processing.
+Useful 'rcodes' include 0 for "no error" and ``pdns.NXDOMAIN`` for "NXDOMAIN", ``pdns.DROP`` to drop the question from further processing.
```
rgacogne, line 246:
```suggestion
Useful 'rcodes' include 0 for "no error" and ``pdns.NXDOMAIN`` for "NXDOMAIN". Before 4.4.0, ``pdns.DROP`` can also be used to drop the question without any further processing.
```

pdns/recursordist/docs/lua-scripting/hooks.rst
```diff
@@ -24,8 +24,13 @@ It also means that to print such names, the ``:toString`` method must be used (o
 Once a script is loaded, PowerDNS looks for the interception functions in the loaded script.
 All of these functions are optional.
 
-If a function returns true, it will indicate that it handled a query.
-If it returns false, the Recursor will continue processing unchanged (with one minor exception).
+If ``ipfilter`` returns ``true``, the query is dropped.
+If ``preresolve`` returns ``true``, it will indicate it handled a query, and the recursor will send to result as constructed in the functions to the client.
+If it returns ``false``, the Recursor will continue processing.
+For the other functions, the return value will indicate if DNSSEC validation should take place.
```
rgacogne, line 30: While it's technically the only thing that changes today when `true` is returned, I'm not sure we should document it that way. Perhaps:
```suggestion
For the other functions, the return value will indicate that an alteration has been made. In that case DNSSEC validation will be automatically disabled since the content might not be genuine anymore.
```

pdns/recursordist/docs/lua-scripting/hooks.rst
```diff
@@ -24,8 +24,13 @@ It also means that to print such names, the ``:toString`` method must be used (o
...
+For modified results, the functions should return ``true`` to avoid failing DNSSEC validations.
+At specific points the Recursor will check if policy handling should take place.
+These points are immediatly after ``preresolve``, after resolving and after ``postresolve``.
```
rgacogne, line 33:
```suggestion
These points are immediately after ``preresolve``, after resolving and after ``postresolve``.
```

pdns/recursordist/docs/lua-scripting/hooks.rst
```diff
@@ -24,8 +24,13 @@ It also means that to print such names, the ``:toString`` method must be used (o
 Once a script is loaded, PowerDNS looks for the interception functions in the loaded script.
 All of these functions are optional.
 
-If a function returns true, it will indicate that it handled a query.
-If it returns false, the Recursor will continue processing unchanged (with one minor exception).
+If ``ipfilter`` returns ``true``, the query is dropped.
+If ``preresolve`` returns ``true``, it will indicate it handled a query, and the recursor will send to result as constructed in the functions to the client.
+If it returns ``false``, the Recursor will continue processing.
+For the other functions, the return value will indicate if DNSSEC validation should take place.
+For modified results, the functions should return ``true`` to avoid failing DNSSEC validations.
```
rgacogne, line 31:
```suggestion
```

rgacogne (changes requested):

pdns/recursordist/docs/lua-scripting/hooks.rst
```diff
@@ -24,8 +24,12 @@ It also means that to print such names, the ``:toString`` method must be used (o
 Once a script is loaded, PowerDNS looks for the interception functions in the loaded script.
 All of these functions are optional.
 
-If a function returns true, it will indicate that it handled a query.
-If it returns false, the Recursor will continue processing unchanged (with one minor exception).
+If ``ipfilter`` returns ``true``, the query is dropped.
+If ``preresolve`` returns ``true``, it will indicate it handled a query, and the recursor will send to result as constructed in the functions to the client.
```
rgacogne, line 28:
```suggestion
If ``preresolve`` returns ``true``, it will indicate it handled a query, and the recursor will send the result as constructed in the function to the client.
```

rgacogne (approved):
One nit, looks good!

PR #10115, opened 2021-02-22
dnsdist: Warn on unsupported parameters

cmouse:
### Short description
Change behavior of many commands to actually check the passed in vars / parameters that they are consumed fully. If they are not, we assume that they were not supported, and indicate first failing var / parameter.

### Checklist
I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdist-lua-actions.cc
+++ b/pdns/dnsdist-lua-actions.cc
@@ -31,7 +31,6 @@
 #include "dnsdist-kvs.hh"
 #include "dnsdist-svc.hh"
 
-#include "dolog.hh"
 #include "dnstap.hh"
 #include "dnswriter.hh"
 #include "ednsoptions.hh"
@@ -2100,6 +2099,7 @@ static void addAction(GlobalStateHolder<vector<T> > *someRuleActions, const luad
   boost::uuids::uuid uuid;
   uint64_t creationOrder;
   parseRuleParams(params, uuid, name, creationOrder);
+  checkAllParametersConsumed("addAction", params);
 
   auto rule = makeRule(var);
   someRuleActions->modify([&rule, &action, &uuid, creationOrder, &name](vector<T>& ruleactions){
@@ -2109,22 +2109,12 @@ static void addAction(GlobalStateHolder<vector<T> > *someRuleActions, const luad
 
 typedef std::unordered_map<std::string, boost::variant<bool, uint32_t> > responseParams_t;
 
-static void parseResponseConfig(boost::optional<responseParams_t> vars, ResponseConfig& config)
+static void parseResponseConfig(boost::optional<responseParams_t>& vars, ResponseConfig& config)
 {
-  if (vars) {
-    if (vars->count("ttl")) {
-      config.ttl = boost::get<uint32_t>((*vars)["ttl"]);
-    }
-    if (vars->count("aa")) {
-      config.setAA = boost::get<bool>((*vars)["aa"]);
-    }
-    if (vars->count("ad")) {
-      config.setAD = boost::get<bool>((*vars)["ad"]);
-    }
-    if (vars->count("ra")) {
-      config.setRA = boost::get<bool>((*vars)["ra"]);
-    }
-  }
+  getOptionalValue<uint32_t>(vars, "ttl", config.ttl);
+  getOptionalValue<bool>(vars, "aa", config.setAA);
+  getOptionalValue<bool>(vars, "ad", config.setAD);
+  getOptionalValue<bool>(vars, "ra", config.setRA);
 }
 
 void setResponseHeadersFromConfig(dnsheader& dh, const ResponseConfig& config)
@@ -2153,6 +2143,7 @@ void setupLuaActions(LuaContext& luaCtx)
       uint64_t creationOrder;
       std::string name;
       parseRuleParams(params, uuid, name, creationOrder);
+      checkAllParametersConsumed("newRuleAction", params);
 
       auto rule = makeRule(dnsrule);
       DNSDistRuleAction ra({std::move(rule), action, std::move(name), uuid, creationOrder});
@@ -2277,6 +2268,7 @@ void setupLuaActions(LuaContext& luaCtx)
       auto ret = std::shared_ptr<DNSAction>(new SpoofAction(addrs));
       auto sa = std::dynamic_pointer_cast<SpoofAction>(ret);
       parseResponseConfig(vars, sa->d_responseConfig);
+      checkAllParametersConsumed("SpoofAction", vars);
       return ret;
     });
 
@@ -2291,6 +2283,7 @@ void setupLuaActions(LuaContext& luaCtx)
       auto ret = std::shared_ptr<DNSAction>(new SpoofAction(DNSName(a)));
       auto sa = std::dynamic_pointer_cast<SpoofAction>(ret);
       parseResponseConfig(vars, sa->d_responseConfig);
+      checkAllParametersConsumed("SpoofCNAMEAction", vars);
       return ret;
     });
 
@@ -2308,6 +2301,7 @@ void setupLuaActions(LuaContext& luaCtx)
       auto ret = std::shared_ptr<DNSAction>(new SpoofAction(raws));
       auto sa = std::dynamic_pointer_cast<SpoofAction>(ret);
       parseResponseConfig(vars, sa->d_responseConfig);
+      checkAllParametersConsumed("SpoofRawAction", vars);
       return ret;
     });
 
@@ -2402,6 +2396,7 @@ void setupLuaActions(LuaContext& luaCtx)
       auto ret = std::shared_ptr<DNSAction>(new RCodeAction(rcode));
       auto rca = std::dynamic_pointer_cast<RCodeAction>(ret);
       parseResponseConfig(vars, rca->d_responseConfig);
+      checkAllParametersConsumed("RCodeAction", vars);
       return ret;
     });
 
@@ -2409,6 +2404,7 @@ void setupLuaActions(LuaContext& luaCtx)
       auto ret = std::shared_ptr<DNSAction>(new ERCodeAction(rcode));
       auto erca = std::dynamic_pointer_cast<ERCodeAction>(ret);
       parseResponseConfig(vars, erca->d_responseConfig);
+      checkAllParametersConsumed("ERCodeAction", vars);
       return ret;
     });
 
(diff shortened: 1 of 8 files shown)


omoerbeek:
This idea would be nice for the recursor as well. Some general thoughts:
The three lines
```
 if (vars->count("idleTimeout")) {
        frontend->d_idleTimeout = boost::get<int>((*vars)["idleTimeout"]);
        vars->erase("idleTimeout");
 }
```
are repeated over and over again.
I'd love to see a function in the line of
```
getOptionalValue<int>(frontend->d_idleTimeout , vars, "idleTimeout");
```
to avoid the repeating.
Also, the exceptions thrown mean that previously working config (with typos) now cause startup failure. That is likely too much, a warning would be better imo.

cmouse:
Tested that it does indeed print warning

```
> pc = newPacketCache(100, {dontAget=true})
Unknown key 'dontAget' given - ignored
```

pdns/dnsdist-lua-actions.cc
```diff
@@ -1576,18 +1576,11 @@ typedef std::unordered_map<std::string, boost::variant<bool, uint32_t> > respons
 static void parseResponseConfig(boost::optional<responseParams_t> vars, ResponseConfig& config)
```
rgacogne, line 1576:
```suggestion
static void parseResponseConfig(boost::optional<responseParams_t>& vars, ResponseConfig& config)
```

pdns/dnsdist-lua.cc
```diff
@@ -102,26 +102,16 @@ typedef std::unordered_map<std::string, boost::variant<bool, int, std::string, s
 static void parseLocalBindVars(boost::optional<localbind_t> vars, bool& reusePort, int& tcpFastOpenQueueSize, std::string& interface, std::set<int>& cpus, int& tcpListenQueueSize, size_t& maxInFlightQueriesPerConnection)
```
rgacogne, line 102:
```suggestion
static void parseLocalBindVars(boost::optional<localbind_t>& vars, bool& reusePort, int& tcpFastOpenQueueSize, std::string& interface, std::set<int>& cpus, int& tcpListenQueueSize, size_t& maxInFlightQueriesPerConnection)
```

pdns/dnsdist-lua.hh
```diff
@@ -107,3 +107,37 @@ void setupLuaInspection(LuaContext& luaCtx);
...
+static inline int getOptionalValue(boost::optional<V>& vars, const std::string& key, T& value) {
+  /* nothing found, nothing to return */
+  if (!vars)
+    return 0;
+  if (vars->count(key)) {
+    try {
+      value = boost::get<G>((*vars)[key]);
+    } catch (boost::bad_get& e) {
+      /* key is there but isn't compatible */
+      return -1;
```
rgacogne, line 129: Would it make sense to log a warning there? I don't think we used to but it might be a nice improvement?

pdns/dnsdist-lua.hh
```diff
@@ -107,3 +107,37 @@ void setupLuaInspection(LuaContext& luaCtx);
...
+    }
+  }
+  return vars->erase(key);
+}
+
+template<class V>
+static inline void checkAllParametersConsumed(const boost::optional<V>& vars) {
```
rgacogne, line 136: Did you consider passing a `const std::string&` indicating where the parameters were not consumed? It would be a bit of extra work but I think it might be very helpful, so I'm willing to write that code if that makes sense to you :)

pdns/dnsdist-lua.cc
```diff
@@ -102,26 +102,16 @@ typedef std::unordered_map<std::string, boost::variant<bool, int, std::string, s
...
+    std::vector<std::pair<int,int>> setCpus;
+
+    getOptionalValue<bool>(vars, "reusePort", reusePort);
+    getOptionalValue<int>(vars, "tcpFastOpenQueueSize", tcpFastOpenQueueSize);
+    getOptionalValue<int>(vars, "tcpListenQueueSize", tcpListenQueueSize);    
+    getOptionalValue<int>(vars, "maxInFlight", maxInFlightQueriesPerConnection);
+    getOptionalValue<std::string>(vars, "interface", interface);
+    if (getOptionalValue<decltype(setCpus)>(vars, "cpus", setCpus) > 0)
+      for (const auto& cpu : setCpus)
         cpus.insert(cpu.second);
```
rgacogne, line 114: Just a nit but I would really prefer braces around the if and the for bodies.

rgacogne (changes requested):
Looks like we often pass a _copy_ of the variables, so we need to fix that at least. The general concept is awesome and the code looks good.

pdns/dnsdist-lua.hh
cmouse: What would the string do?

pdns/dnsdist-lua.hh
cmouse: This is there mostly to facilitate
``` 

      if (getOptionalValue<std::string>(vars, "pool", valueStr) > 0) {
        ret->pools.insert(valueStr);
      } else if (getOptionalValue<decltype(pools)>(vars, "pool", pools) > 0) {
        for(auto& p : pools) {
          ret->pools.insert(p.second);
        }
      }
 ```

Comment (rgacogne) [reply pdns/dnsdist-lua.hh:129]: Right, I thought we could catch an invalid value that way, but I guess it's not practical.

File: pdns/dnsdist-lua.hh (hunk 4 above)
Comment (rgacogne) [reply pdns/dnsdist-lua.hh:136]: Tell what directive was involved, like `newServer` or `addLocal`?

File: pdns/dnsdist-lua.cc [hunk 6]
```diff
@@ -254,18 +225,18 @@ static void setupLuaConfig(LuaContext& luaCtx, bool client, bool configCheck)
...
-      newserver_t vars;
+      boost::optional<newserver_t> vars;
 
       ComboAddress serverAddr;
       std::string serverAddressStr;
       if(auto addrStr = boost::get<string>(&pvars)) {
         serverAddressStr = *addrStr;
         if(qps) {
-          vars["qps"] = std::to_string(*qps);
+          (*vars)["qps"] = std::to_string(*qps);
```
rgacogne, line 235: Does that work? I would expect that de-referencing an unset `boost::optional` would be a problem.

pdns/dnsdist-lua.cc
```diff
@@ -379,134 +357,160 @@ static void setupLuaConfig(LuaContext& luaCtx, bool client, bool configCheck)
...
       }
 
-      if(vars.count("ipBindAddrNoPort")) {
-        ret->ipBindAddrNoPort=boost::get<bool>(vars["ipBindAddrNoPort"]);
-      }
+      getOptionalValue<bool>(vars, "setCD", ret->setCD);
+      getOptionalValue<bool>(vars, "mustResolve", ret->mustResolve);
+      getOptionalValue<bool>(vars, "useClientSubnet", ret->useECS);
+      getOptionalValue<bool>(vars, "useProxyProtocol", ret->useProxyProtocol);
+      getOptionalValue<bool>(vars, "disableZeroScoping", ret->disableZeroScope);
```
rgacogne, line 465:
```suggestion
      getOptionalValue<bool>(vars, "disableZeroScope", ret->disableZeroScope);
```

pdns/dnsdist-lua.cc
```diff
@@ -2358,13 +2345,10 @@ static void setupLuaConfig(LuaContext& luaCtx, bool client, bool configCheck)
 
         if (vars) {
           parseLocalBindVars(vars, reusePort, tcpFastOpenQueueSize, interface, cpus, tcpListenQueueSize, maxInFlightQueriesPerConn, tcpMaxConcurrentConns);
-
-          if (vars->count("provider")) {
-            frontend->d_provider = boost::get<const string>((*vars)["provider"]);
-            boost::algorithm::to_lower(frontend->d_provider);
-          }
-
+          getOptionalValue<std::string>(vars, "provider", frontend->d_provider);
```
rgacogne, line 2348: Isn't the `to_lower` part gone?

rgacogne (changes requested):
I love the general idea, and the implementation looks good! A few issues remain and I would not mind a rebase on the current master to ensure the checks are still valid, but other than that I'd like to merge it soon!

pdns/dnsdist-lua.cc
cmouse: hm. that would also change the parameter name no?

pdns/dnsdist-lua.cc
cmouse: I have to try.

pdns/dnsdist-lua.cc
cmouse: so it seems.

pdns/dnsdist-lua.cc
rgacogne: I believe the name is currently `disableZeroScope`?

rgacogne:
I'm in the process of (painfully, but that's on me) rebasing this pull request on master, just FYI.

rgacogne:
Rebased to fix conflicts.

rgacogne (approved):

rgacogne:
Rebased on master, and ready to be merged. A second review would be nice to have to make sure I did not mess up :)

pdns/dnsdist-lua.cc
```diff
@@ -181,65 +171,47 @@ static bool loadTLSCertificateAndKeys(const std::string& context, std::vector<TL
...
-  }
+  getOptionalValue<std::string>(vars, "ciphers", config.d_ciphers);
+  getOptionalValue<std::string>(vars, "ciphersTLS13", config.d_ciphers13);
 
 #ifdef HAVE_LIBSSL
-  if (vars->count("minTLSVersion")) {
-    config.d_minTLSVersion = libssl_tls_version_from_string(boost::get<const string>((*vars)["minTLSVersion"]));
-  }
+  std::string minVersion;
+  if (getOptionalValue<std::string>(vars, "minTLSVersion", minVersion) > 0)
```
chbruyand, line 181: I think we want the conditionnal content to be surrounded by `{}` (here and on various other places). Maybe have a `clang-tidy` pass ?

pdns/dnsdist-lua.cc
```diff
@@ -494,74 +466,60 @@ static void setupLuaConfig(LuaContext& luaCtx, bool client, bool configCheck)
...
 
-                         if (vars.count("checkFunction")) {
-                           config.checkFunction = boost::get<DownstreamState::checkfunc_t>(vars["checkFunction"]);
-                         }
+                         getOptionalValue<std::string>(vars, "checkType", config.checkType);
 
-                         if (vars.count("checkTimeout")) {
-                           config.checkTimeout = std::stoi(boost::get<string>(vars["checkTimeout"]));
+                         if (getOptionalValue<std::string>(vars, "checkClass", valueStr) > 0) {
+                           config.checkClass = std::stoi(valueStr);
```
chbruyand, line 476: As done a bit later, do we want to guard potential conversion exceptions for optional values ?

chbruyand: I'm also wondering if there is a good reason to not directly `getOptionalValue<int>(vars, "checkClass", config.checkClass)`

chbruyand (commented):
Maybe we could unify a bit some integer values processing. Otherwise this lgtm and feels like a really good feature.

pdns/dnsdist-lua.cc
rgacogne: > I'm also wondering if there is a good reason to not directly `getOptionalValue<int>(vars, "checkClass", config.checkClass)`

Wouldn't that fail because the Lua wrapper used the `string` type for that variable? We could extend the variant to allow integers, but then we need to change all the existing types and make sure the conversion works properly. One drawback that I did not notice before with the new code is that if we pass a string when we expect an integer, it is no longer considered a parsing error and we don't even warn, except at the end to say something like `newServer: Unknown key 'checkClass' given - ignored`, which is confusing: the key exists, but the type is wrong.

rgacogne: I pushed a new commit with better handling of numerical values, comments welcome!

chbruyand (approved):
lgtm!

PR #10118, opened 2021-02-24
rec: Change dnssec default to `process`

pieterlexis:
### Short description
This is a sane default nowadays. It will also greatly help if the aggressive NSEC cache (#10047) is enabled by default.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/pdns_recursor.cc
+++ b/pdns/pdns_recursor.cc
@@ -5320,7 +5320,7 @@ int main(int argc, char **argv)
     ::arg().set("local-address","IP addresses to listen on, separated by spaces or commas. Also accepts ports.")="127.0.0.1";
     ::arg().setSwitch("non-local-bind", "Enable binding to non-local addresses by using FREEBIND / BINDANY socket options")="no";
     ::arg().set("trace","if we should output heaps of logging. set to 'fail' to only log failing domains")="off";
-    ::arg().set("dnssec", "DNSSEC mode: off/process-no-validate (default)/process/log-fail/validate")="process-no-validate";
+    ::arg().set("dnssec", "DNSSEC mode: off/process-no-validate/process (default)/log-fail/validate")="process";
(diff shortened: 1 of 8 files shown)


omoerbeek (approved):

pieterlexis:
>  ci/circleci: test-recursor-regression — Your tests failed on CircleCI 

seems like a legit failure

rgacogne:
One drawback of that latest commit is that it adds 25s for each test suite because we start several auth processes, run secure-zone, etc. Perhaps we could start only the auth processes that we really need instead?

pieterlexis:
> One drawback of that latest commit is that it adds 25s for each test suite because we start several auth processes, run secure-zone, etc. Perhaps we could start only the auth processes that we really need instead?

we do, just the root. As having no reachable root server breaks the recursor

rgacogne:
> we do, just the root.

Ah, I was missing the fact that we only prepare the zones present in `_auth_zones`, and start the processes needed for these only. That's nice! You can indeed ignore my comment, then, sorry :)

omoerbeek:
Failing test:

```
FAIL: testA (test_Protobuf.OutgoingProtobufDefaultTest)
  File "/opt/project/regression-tests.recursor-dnssec/test_Protobuf.py", line 372, in testA
    self.checkProtobufOutgoingQuery(msg, dnsmessage_pb2.PBDNSMessage.UDP, query, dns.rdataclass.IN, dns.rdatatype.A, name)
  File "/opt/project/regression-tests.recursor-dnssec/test_Protobuf.py", line 226, in checkProtobufOutgoingQuery
    self.checkOutgoingProtobufBase(msg, protocol, query, initiator, length=length)
  File "/opt/project/regression-tests.recursor-dnssec/test_Protobuf.py", line 166, in checkOutgoingProtobufBase
    self.assertEquals(msg.inBytes, len(query.to_wire()))
AssertionError: 32 != 44
```

Issue #8284, opened 2019-09-09
Feature request: Add vrf support


krombel:
- Programs: Authoritative, Recursor, dnsdist
 - Issue type: Feature request

### Usecase
At the moment it is not possible to define per service on which VRF they should listen on. This applies to the Authoriative, Recursor and dnsdist.

The main use case for me is the webserver component:
I want it running on localhost (to e.g. allow metrics scraping) but not my VRF in which the service should listen to because that would require using a public ip address

When I use e.g. `ip vrf exec external /path/to/dnsdist [...]` I am enforced to have the webserver running on a public interface as well.

When I try to let it listen on 127.0.0.1 (started with `ip vrf exec ...`) it fails with

    Unable to bind to webserver socket on 127.0.0.1:8080: binding socket to 127.0.0.1:8080: Cannot assign requested address

When I do not use the vrf the service is not publicly accessible

Same applies for the Control Port for dnsdist: Even when using a public IP it is not possible to bind the port. My first guess is an UDP protocol but I guess you might know that better.

### Description
Have an option to bind to vrf per service. If there is such option already only the doc for this is missing

Habbie:
For dnsdist, can you tell us if the `interface` option at https://dnsdist.org/reference/config.html#addLocal covers your needs?

krombel:
You are right. Adding "interface=external" worked as expected.
Have missed that option.
Thanks for clearing things up :slightly_smiling_face:

krombel:
I just realize, that I missed that this config is only available for dnsdist.
How is this working for authoritative and recursor?

Habbie:
>  How is this working for authoritative and recursor?

I don't think they can do it right now.

awlx:
It also seems that the newServer call cannot handle vrfs.

If I add for example:

`newServer({address="2001:608:a01::40", name="gw01", source="vrf_external"})     -- downstream servers for recursion`

It still picks the outgoing ip of the master vrf. And if I define an IP with ip@vrf_external it cannot bind anymore :/.

`Oct 02 12:00:16 webfrontend01 dnsdist[27956]: Fatal Lua error: [string "chunk"]:22: Caught exception: binding socket to [2001:608:a01::3]:0: Cannot assign requested address`

rgacogne:
If you only specifies the interface we ask the kernel to use that interface, and let the choice of the IP to use to it.
[...]

PR #10130, opened 2021-03-02
dnsdist: Remove forgotten warning about source interface selection

rgacogne:
### Short description

This removes a warning introduced in #10108.

Before that PR we supported source interface selection over UDP and TCP when talking to our backends, via:
- the "generic" way, binding with the right source address, and passing the interface index down to the actual `sendmsg` call via `IP_PKTINFO` and `IPV6_PKTINFO` in case that was not enough ;
- `SO_BINDTODEVICE` on Linux, which was reported as the only working solution on Linux in https://github.com/PowerDNS/pdns/issues/8284.

After that PR we don't set the interface index anymore on the TCP path, and instead rely on:
- binding with the right source address being good enough on most OS ;
- setting `SO_BINDTODEVICE` on Linux.

As far as I can tell there is no other option, `IP_SENDIF` never made it to the BSD world, but from what I read binding the socket should be enough on all non-Linux systems, and `SO_BINDTODEVICE` should fix these.

If we find out that this is not enough, we will need to pass the interface down to the actual `sendmsg` call again, but note that this will not work once we start doing `DoT` or `DoQ` toward the backends because right now the TLS libraries handle these calls themselves. 

### Checklist

I have:
- [x] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdistdist/dnsdist-tcp-downstream.cc
+++ b/pdns/dnsdistdist/dnsdist-tcp-downstream.cc
@@ -52,7 +52,6 @@ IOState TCPConnectionToBackend::sendQuery(std::shared_ptr<TCPConnectionToBackend
 {
   DEBUGLOG("sending query to backend "<<conn->getDS()->getName()<<" over FD "<<conn->d_handler->getDescriptor());
 
-#warning FIXME: TODO: this drops 1/ source selection other than SO_BINDTODEVICE, perhaps we should look into IP_SENDIF?
   IOState state = conn->d_handler->tryWrite(conn->d_currentQuery.d_buffer, conn->d_currentPos, conn->d_currentQuery.d_buffer.size());
 
   if (state != IOState::Done) {


omoerbeek:
Yes, binding on address should be enough. 

With IPv6 you can specify the interface name or index for interface scoped addresses, like `fe80::b62e:99ff:fef2:6600%rge0`. 
`iputils.cc:makeIPv6sockaddr` does fall back to using `getaddrinfo` if `inet_pton` fails (as it does on scoped addresses). 

Part of a ktrace on OpenBSD with `query-local-address=fe80::b62e:99ff:fef2:6600%rge0`:
```
 72796 pdns_recursor CALL  bind(27,0x1442d3f4620,28)
 72796 pdns_recursor STRU  struct sockaddr { AF_INET6, [fe80::b62e:99ff:fef2:6600%1]:0 }
 72796 pdns_recursor RET   bind 0
```
Note that the name `rge0` is internally translated as index 1 (`%1`)
The config does not work since it is a non-routable source address, but at least this shows the binding is working.

Issue #10128, opened 2021-03-02
dnsdist: Setting setCacheCleaningPercentage() breaks packetcache in the long run


aj-gh:
- Program: dnsdist
 - Issue type: Bug report

### Short description
Setting setCacheCleaningPercentage() to a value != 100 seems to break the packetcache - not immediately but in the long run, depending on the amount of queries, packetcache size and sharding. During runtime the packetcache fill level seems to become almost constant and the cache hit rate worse per minute.
I could be wrong but from a quick look at the code it seems [DNSDistPacketCache::purgeExpired](https://github.com/PowerDNS/pdns/blob/2753956a56928f6c996dc457340fb353bb5ee755/pdns/dnsdist-cache.cc#L312-L332) iterates over packetcache shards until it has cleaned sufficiently many entries. However I'd assume it manages to do this in the first few shards and never reaches the last causing these to fill up.
On a full shard `DNSDistPacketCache::insertLocked` seems to just drop the entry. Note that at this stage there doesn't seem to be a metric for this while it might be an interesting figure (but I might have missed it).
Sharding was introduced in 2b3eefc37bc815538059105c3c49d0e6b6f383f4, `setCacheCleaningPercentage()` earlier in f65ea0c29202c50561b0ca98fdbd1cb630b703b6 - guess `setCacheCleaningPercentage()` might need a review :wink:
Also proposing a metric for "full shards insert drops" in case there isn't one (and it still makes sense after this fix).

### Environment

 - Operating system: Ubuntu 20.04.2 LTS
 - Software version: dnsdist 1.5.1-1pdns.focal
 - Software source: PowerDNS repo

### Steps to reproduce
1. Set `setCacheCleaningPercentage(30)`
2. Let dnsdist run and feed it with some real-live traffic
3. Watch packet cache hit rate degrade over time

### Expected behaviour
setCacheCleaningPercentage() not having much impact on cache hit rate.

### Actual behaviour
Last packetcache shards seem to become filled up with no further entries / updates possible.

rgacogne:
> I could be wrong but from a quick look at the code it seems DNSDistPacketCache::purgeExpired iterates over packetcache shards until it has cleaned sufficiently many entries. However I'd assume it manages to do this in the first few shards and never reaches the last causing these to fill up.

`d_expungeIndex` should ensure we always start with a shard that has not been scanned the last time, if any, though?

rgacogne:
Should be fixed by #10133.

PR #10133, opened 2021-03-02
dnsdist: Clean up expired entries from all the packet cache's shards

rgacogne:
### Short description

Otherwise we might remove enough entries from the first shards only and stop there, which means that the other shards might remain full.
This might be fine if we clean up often enough since the next cleaning run will start with the remaining shards, but that's sub-optimal when we are often nearly full because it will prevent new entries from being inserted in the shards that are full.

This should improve https://github.com/PowerDNS/pdns/issues/10128.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/pdns/dnsdist-cache.cc
+++ b/pdns/dnsdist-cache.cc
@@ -336,34 +336,39 @@ size_t DNSDistPacketCache::purgeExpired(size_t upTo, const time_t now)
 }
 
 /* Remove all entries, keeping only upTo
-   entries in the cache */
+   entries in the cache.
+   If the cache has more than one shard, we will try hard
+   to make sure that every shard has free space remaining.
+*/
 size_t DNSDistPacketCache::expunge(size_t upTo)
 {
+  const size_t maxPerShard = upTo / d_shardCount;
+
   size_t removed = 0;
-  const uint64_t size = getSize();
 
-  if (upTo >= size) {
-    return removed;
-  }
+  for (auto& shard : d_shards) {
+    WriteLock w(&shard.d_lock);
+    auto& map = shard.d_map;
 
-  size_t toRemove = size - upTo;
+    if (map.size() <= maxPerShard) {
+      continue;
+    }
 
-  for (uint32_t shardIndex = 0; shardIndex < d_shardCount; shardIndex++) {
-    WriteLock w(&d_shards.at(shardIndex).d_lock);
-    auto& map = d_shards[shardIndex].d_map;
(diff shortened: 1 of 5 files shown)


pdns/dnsdist-cache.cc
```diff
@@ -294,75 +294,80 @@ bool DNSDistPacketCache::get(DNSQuestion& dq, uint16_t queryId, uint32_t* keyOut
...
-
-  size_t toRemove = size - upTo;
+  const size_t maxPerShard = upTo / d_shardCount;
 
+  size_t removed = 0;
   size_t scannedMaps = 0;
 
-  const time_t now = time(nullptr);
   do {
     uint32_t shardIndex = (d_expungeIndex++ % d_shardCount);
```
omoerbeek, line 308: Since  `d_expungeIndex` is incremented `d_shardCount` times I believe the scanning will always start at shard 0. If you wamnt to start at a different shard for each `purge` call the increment needs to be above the loop. Or just removed `expungeIndex` ?

rgacogne: Right, since we now always scan all buckets ` d_expungeIndex` is not relevant anymore, I'll remove it.

pdns/dnsdist-cache.cc
```diff
@@ -294,75 +294,80 @@ bool DNSDistPacketCache::get(DNSQuestion& dq, uint16_t queryId, uint32_t* keyOut
 
 /* Remove expired entries, until the cache has at most
    upTo entries in it.
+   If the cache has more than one shard, we will try hard
```
omoerbeek, line 297: Is this true? I think if no expired entries are in a shard, nothing will be deleted...

rgacogne: We will only remove expired entries, yes, we are in `purgeExpired()`, but we will scan all entries if we need to.

pdns/dnsdist-cache.cc
```diff
@@ -302,14 +302,14 @@ size_t DNSDistPacketCache::purgeExpired(size_t upTo, const time_t now)
   const size_t maxPerShard = upTo / d_shardCount;
 
   size_t removed = 0;
-  size_t scannedMaps = 0;
+  uint32_t shardIndex = 0;
 
   do {
```
omoerbeek, line 307: Not wrong, but now that `d_expungeIndex` is gone you might as well turn this into a regular `for (range)` loop

omoerbeek (approved):

PR #10152, opened 2021-03-08
Make the non-resolving cache even less aggressive and add a unit test.

omoerbeek:
In particular, we clear the entry if we see a succeeding resolve.
This means that we only a couple of failing resolves in a row will
cause a ban. Also change to default non-resolving-ns-max-fails from
1 to 5.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [X] documented the code

### DIFF
--- a/pdns/syncres.cc
+++ b/pdns/syncres.cc
@@ -2288,11 +2288,14 @@ vector<ComboAddress> SyncRes::retrieveAddressesForNS(const std::string& prefix, 
 {
   vector<ComboAddress> result;
 
-
+  size_t nonresolvingfails = 0;
   if (!tns->first.empty()) {
-    if (s_nonresolvingnsmaxfails > 0 && t_sstorage.nonresolving.value(tns->first) >= s_nonresolvingnsmaxfails) {
-      LOG(prefix<<qname<<": NS "<<tns->first<< " in non-resolving map, skipping"<<endl);
-      return result;
+    if (s_nonresolvingnsmaxfails > 0) {
+      nonresolvingfails = t_sstorage.nonresolving.value(tns->first);
+      if (nonresolvingfails >= s_nonresolvingnsmaxfails) {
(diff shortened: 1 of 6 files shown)


pdns/syncres.cc
```diff
@@ -2310,10 +2313,16 @@ vector<ComboAddress> SyncRes::retrieveAddressesForNS(const std::string& prefix,
...
-      if (!dontThrottleNames->check(tns->first)) {
-        t_sstorage.nonresolving.incr(tns->first, d_now);
+    if (s_nonresolvingnsmaxfails > 0 && nretrieveAddressesForNS > oldnretrieveAddressesForNS) {
+      if (result.empty()) {
+        auto dontThrottleNames = g_dontThrottleNames.getLocal();
+        if (!dontThrottleNames->check(tns->first)) {
+          t_sstorage.nonresolving.incr(tns->first, d_now);
+        }
+      }
+      else if (nonresolvingfails > 0) {
```
pieterlexis, line 2323: it would be nice to have this bit tested as well?

pieterlexis (approved):
tests are good!

pdns/syncres.cc
omoerbeek: Indeed, I'll see if I can do that as well,

rgacogne (approved):
Nice, looks good to me! I have a test setup that reliably triggered an issue with the previous version, I'll test that one later today.

omoerbeek:
> Nice, looks good to me! I have a test setup that reliably triggered an issue with the previous version, I'll test that one later today.

That would be nice, after that, I'll set the alpha3 tag. But note that I chnaged the code slightly after your comment.

rgacogne:
I can confirm that 59ab9505d9cfe3ba2a3a4508d7c0c334f73fa375 fixed the issue for me, and it does not come back with the current version of this PR!

PR #10154, opened 2021-03-09
Add SQLITE_BUSY wait for schema migration.

hansborr:
A schema migration can fail if the dnssec db is used by pdns
when the script runs, as the database may be locked.

Add a 60 second SQLITE_BUSY timeout, so that the migration will
wait up to 60 total seconds if it initially fails to acquire the
lock.

### Short description
If the schema migration script bind-dnssec.4.2.0_to_4.3.0_schema.sqlite3.sql runs at a time when the dnssec db happens to be in use, it can fail to acquire a lock for the transaction. This will abort the schema migration script, and no change gets made.

The change in this PR ensures that the sqlite queries will wait for up to 60 seconds if it cannot acquire the lock.

I did not do any unit test, as I'm unsure of how to script this, but manual testing can be done as follows:
1. Start with an older dnssec db.
2. In one terminal, use "sqlite3 /path/to/dnssec.db", and then force a lock with "BEGIN EXCLUSIVE TRANSACTION;"
3. In another terminal, run the schema migration script "sqlite3 /path/to/dnssec.db < /path/to/bind-dnssec.4.2.0_to_4.3.0_schema.sqlite3.sql"

Prior to the PR, this step would fail immediately, leaving the schema in the old format. With this PR, it should wait for up to 60 seconds.

4. Quickly (before the 60 second time limit) switch to your other terminal and quit the sqlite3 session with ".q" -- this should allow the schema migration to proceed as it releases the lock.

### Checklist

I have:
- [ ] compiled this code (N/A -- the change is not in a compiled portion of code)
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/bind-dnssec.4.2.0_to_4.3.0_schema.sqlite3.sql
+++ b/pdns/bind-dnssec.4.2.0_to_4.3.0_schema.sqlite3.sql
@@ -1,3 +1,4 @@
+PRAGMA busy_timeout = 60000;
 BEGIN TRANSACTION;
   CREATE TABLE cryptokeys_temp (
     id                  INTEGER PRIMARY KEY,


cmouse:
Why not just shutdown pdns when doing this?

hansborr:
> Why not just shutdown pdns when doing this?

I do now, but it still seems like a good idea to have this in the migration script in case anyone else runs into the problem in the future.

mind04:
Once the database is locked for the upgrade powerdns does not answer to DNS queries.
This pull not a solution since it will only move the pain to a different place.
I agree with @cmouse. Shutting down pdns is the best practice here. The upgrade notes could use a few words about this.

hansborr:
> Once the database is locked for the upgrade powerdns does not answer to DNS queries.
> This pull not a solution since it will only move the pain to a different place.

The migration itself happens very quickly for most DB sizes on modern hardware -- it usually takes a small fraction of a second. It would only cause any appreciable downtime for very large DBs -- but even then, you end up with the same amount of downtime no matter what approach you use (stopping pdns before the migration, or using sqlite_busy timeout).

But, if someone runs into the locking problem without using sqlite_busy, then the downtime is indefinite -- pdns will remain in a broken statue until they notice dns is broken, check server logs, and manually correct the issue,

This PR is just intended to cover the situation where some person is ignorant of the locking problem does a scripted upgrade without thinking of the potential of locking failures. It makes the upgrade more likely to succeed for them, and does not negatively harm anyone else.

Habbie:
ping

hansborr:
```
I don't follow - why would that happen?
```

I'll try to give you the specific steps to reproduce that, but it will be a bit before I have the time to go through it.

Habbie:
I'm closing this for now. Please let us know when you have any more interesting details.

PR #10157, opened 2021-03-09
Move to hashed passwords for the web interface

### LINKED ISSUE #7937: Feature request: store authentication verifier in hashed format
### Description

It is a common security best practice to store authentication/password verifiers on the server side in a hashed format to reduce the impact / increase the effort for attackers should they be able to gain access to the configuration data. This could happend via an unsecured backup file for example.

There are multiple places where plaintext password/key verifiers are stored:

https://dnsdist.org/reference/config.html#setKey
https://dnsdist.org/reference/config.html#webserver

and maybe other locations?

This issue is about supporting a strong hashformat to protect password verifier / to avoid storing plaintext passwords/keys in the server configuration where possible.

Default filesystem permissions of configuration files should not allow any access to "other".

rgacogne:
### Short description

It is not very good practice to keep the password in plaintext in the configuration file as well as in memory at runtime.

Fixes #7937.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


pdns/dnsdist-lua.cc
```diff
@@ -939,6 +950,10 @@ static void setupLuaConfig(LuaContext& luaCtx, bool client, bool configCheck)
       }
     });
 
+  luaCtx.writeFunction("hashPassword", [](const std::string& password) {
+    return hashPassword(password);
```
Habbie, line 954: Just a note: the copy of this string on the Lua side will not be mlocked.

pdns/dnsdistdist/docs/reference/config.rst
```diff
@@ -317,13 +325,17 @@ Webserver configuration
   .. versionchanged:: 1.6.0
     ``statsRequireAuthentication``, ``maxConcurrentConnections`` optional parameters added.
 
+  .. versionchanged:: 1.7.0
+    The ``password``, optional parameter has been deprecated, ``hashedPassword`` added instead.
```
Habbie, line 329:
```suggestion
    The optional ``password``parameter has been deprecated and replaced with ``hashedPassword``.
```

Habbie (approved):
One note, one text suggestion, definitely approved!

rgacogne:
Otto, Peter, shall I try to extend that PR to also include the authoritative server and the recursor, at least when `sodium` support is available? I was thinking of adding password-hashing capabilities to `pdnsutil` and `rec_control` for that.

Habbie:
Yes please!

omoerbeek:
That would be wonderful!

Habbie:
> `rec_control`

Bit weird to have a 'client side' function in `rec_control`, but the recent FD passing support for `dump-*` at least offers some precedent for `rec_control` being more than a simple socket command client :)

rgacogne:
Rebased to fix a conflict plus:
- automatically detect whether the password is hashed instead of using a separate parameter, moving to a separate class
- use that class in the auth and rec
- add hashing support to `pdnsutil` and `rec_control`
- add unit and speed tests

The auth and rec parts need more testing, doing that now.

rgacogne:
We now also try to do credentials verification in constant time even without `libsodium`, by computing a random seed, hashing the credentials found in the configuration, then later hashing the credentials supplied by the user using the same seed, and comparing the password only if the hashes match.

rgacogne:
> The auth and rec parts need more testing, doing that now.

Looks good!

pieterlexis (commented):
It would be good to add libsodium to the auth and rec packages as well (adding a sodium `BuildRequires` for RPM and adding it in `Build-Depends` in the debbian control file should to the trick if `configure` checks for sodium).

rgacogne:
Don't we already build with libsodium for `ed25519` DNSSEC support?

Habbie:
> Don't we already build with libsodium for `ed25519` DNSSEC support?

Yep, all our package builds already have libsodium!

pieterlexis:
> Don't we already build with libsodium for `ed25519` DNSSEC support?

we do, excellent :)

rgacogne:
The ` test-auth-regress-gpgsql` failure seems real (occurred twice already) but I don't see any useful output in the log. I'll have a look later.

Habbie:
I also saw a few (different) gsql failures today that I could not immediately explain..

zeha:
Q (feel free to ignore): is there a good reason to build without libsodium any longer?

mind04:
el7 and el8 do not have libsodium out of the box

pieterlexis:
> el7 and el8 do not have libsodium out of the box

fwiw: our package builds pull this in via EPEL.

rgacogne:
Rebased to fix a conflict.

Habbie (approved):
Some comments, LGTM really!

rgacogne:
I just implemented your suggestions, thanks a lot! I added support for `sodium_memcmp()` but opted to use `CRYPTO_memcp()` when both options are available, since the assembly implementation used in OpenSSL seems less likely to be helpfully optimized away by the compiler.

pdns/pdnsutil.cc
```diff
@@ -2461,6 +2461,14 @@ try
 
     return 0;
   }
+  else if (cmds[0]=="hash-password") {
+    if (cmds.size() < 2) {
+      cerr<<"Syntax: pdnsutil hash-password PASSWORD"<<endl;
```
pieterlexis, line 2466: this would show up in the shell history, perhaps we should read the password from a prompt?

omoerbeek: `readpassphrase(3)` exists in the BSDs and is available in libbsd on linux.

cmouse:
I would like to raise few small concerns on this:

1. Is it possible to detect and use different algorithms in future if you, for whatever reason, have to add support for some new algorithm?
2. ARGON2 is pretty slow, so if you use it for API authentication, have you considered the impact on API performance? This might be a pointless concern, yet if you spend 1s per request to do authentication, it can have rather severe impact on large changes.

 How hard is ti to

pdns/pdnsutil.cc
rgacogne: Adding a new dependency for that feature seems a bit too much, but why not..

Habbie: We can copy the one .c file instead, Otto pointed out.

rgacogne:
> Is it possible to detect and use different algorithms in future if you, for whatever reason, have to add support for some new algorithm?

Yes, libsodium does handle that for us, you can see several sample of the storage format that does include versioning in the diff.

> ARGON2 is pretty slow, so if you use it for API authentication, have you considered the impact on API performance? This might be a pointless concern, yet if you spend 1s per request to do authentication, it can have rather severe impact on large changes.

The PR adds password hashing and verification to speedtest, so we have some numbers:
```
'Credentials verification test' 0.17 seconds: 17.3 runs/s, 57670.67 usec/run
```

That's with the recommended values for "interactive" usage.
So yes, it's a lot slower, but that's kind of the point. I guess we could make hashing optional if one is willing to accept the risk..

rgacogne:
After further discussions with Habbie, cmouse and Otto on IRC:
- the current behaviour of this PR is "if you stick a plaintext password in the config and hashing is available, we will hash the supplied password and keep only the hashed version in memory" which would lead to an unexpected increase of CPU and memory usage which is not good
- I'll change that to 1/ hashed passwords are hashed 2/ plaintext passwords stay in plaintext in memory unless you ask for it (and then you pay the price)

rgacogne:
I pushed a new commit to read the credentials from the terminal (or stdin), review welcome!

pdns/credentials.cc
```diff
@@ -156,3 +160,89 @@ bool CredentialsHolder::isHashingAvailable()
...
 }
+
+#include <signal.h>
+#include <termios.h>
+
+std::string CredentialsHolder::readFromTerminal()
+{
+  struct termios term;
+  struct termios oterm;
+  memset(&term, 0, sizeof(term));
```
omoerbeek, line 171: WHy init `term` and `oterm` if you are going to overwrite them?

pdns/credentials.cc
```diff
@@ -156,3 +160,89 @@ bool CredentialsHolder::isHashingAvailable()
...
+  term.c_lflag |= ECHO;
+  memset(&oterm, 0, sizeof(oterm));
+  oterm.c_lflag |= ECHO;
```
Habbie, line 174: Won't this be overwritten by `tcgetattr`?

pdns/credentials.cc
rgacogne: They are not overwritten when we fall back to stdin, and we check the value of `term` no matter what. It seemed cleaner to initialize them both.

pdns/credentials.cc
```diff
@@ -156,3 +160,89 @@ bool CredentialsHolder::isHashingAvailable()
...
+  bool restoreTermSettings = false;
+  int termAction = TCSAFLUSH;
+#ifdef TCSASOFT
+  termAction |= TCSASOFT
```
omoerbeek, line 178: missing ;

pdns/credentials.cc
Habbie: but if it doesn't (because somebody breaks this code in the future), this will cause restoreTermSettings, if it happens when it shouldn't, to at least restore ECHO.

pdns/rec_control.cc
```diff
@@ -142,10 +141,17 @@ int main(int argc, char** argv)
           throw PDNSException("Command needs a file argument");
         }
       }
+      else if (commands.at(i) == "hash-password") {
```
omoerbeek, line 144: hash-password missing in help

pdns/credentials.cc
omoerbeek: I kinda like the structrure in the original where term and oterm are only initialised if input != stdin or tcgetattr fails better

rgacogne: OK, I'll see what I can do, hopefully next week.

omoerbeek:
This tests well for rec, but the main page (doing quite some requests for very 1s refresh) does cause a noticeable load.

rgacogne:
I guess argon2 is too slow for us, in that case :-/

Habbie:
> I guess argon2 is too slow for us, in that case :-/

I suppose most other argon2 users have more of a 'session' thing than we do..

setharnold:
Can there be two tokens? One for API use that's fast (maybe stored via hmac, or uses public/private keys, etc) and one that's argon2 based for humans to type into a web interface? A fast-to-check token should be long and entropic, and the friendlier things that humans type can be slower to check.

Thanks

rgacogne:
It would be good to indeed have an endpoint taking argon2-based credentials and returning a token with a limited life-time that could be used for API operations. The problem is that the existing API clients don't handle that mechanism, but I guess we could introduce that while keeping the existing API key around for quite some time.

In the meantime I think our only option is to use a much weaker option to store passwords and API keys :-/ I tried the weakest option support by libsodium (`crypto_pwhash_OPSLIMIT_MIN`) and it's roughly twice as fast than the interactive one, which is still too slow. We could go with PBKDF2 hmac-sha256, for example. That means implementing the parsing of the versioning, salt and iteration count ourselves, though :-/

cmouse:
> 
> 
> It would be good to indeed have an endpoint taking argon2-based credentials and returning a token with a limited life-time that could be used for API operations. The problem is that the existing API clients don't handle that mechanism, but I guess we could introduce that while keeping the existing API key around for quite some time.
> 
> In the meantime I think our only option is to use a much weaker option to store passwords and API keys :-/ I tried the weakest option support by libsodium (`crypto_pwhash_OPSLIMIT_MIN`) and it's roughly twice as fast than the interactive one, which is still too slow. We could go with PBKDF2 hmac-sha256, for example. That means implementing the parsing of the versioning, salt and iteration count ourselves, though :-/

I think you could keep the current implementation. You could add support for bearer token authentication and in-memory token endpoint that returns something like maybe GUID under /token/generate. Then you can use that token as Auth bearer guid to login after this.

This would make it possible for API clients to transition to this, and would allow using encrypted credentials.

rgacogne:
As a quick note, I'd prefer some sort of authenticated token rather than a GUID so we don't have to keep a state (make things harder in anycast setups, restart, ...).
I still think we need a different option, otherwise we would have to keep credentials in plaintext for clients not using the new scheme.

cmouse:
Then you should use JWT tokens. That's the easiest way to do stateless authenticated tokens. And there are several libraries for handling those.

rgacogne:
Oh, have the security nightmares (non-fixed algorithms, having to parse untrusted JSON before checking signatures, ...) surrounding JWT been fixed?

cmouse:
Probably not. No one asks you to fully implement every niche of JWT. But at least it's a standard way to do things. =)

rgacogne:
> In the meantime I think our only option is to use a much weaker option to store passwords and API keys.

The more I think about it the more that seems to be the case. For example in several deployments most of the queries received by the dnsdist and recursor webservers are prometheus ones. As far as I can tell, most tools scraping prometheus support basic auth but would not be able to get an authentication token, so we need to be able to store the credentials in a non-plaintext way even in the direct, non-token mode. That means supporting an algorithm with a cheaper verification cost than argon2. OpenSSL 1.1.1 has support for scrypt KDF for which a low `N` can be selected [1], or we could go the regular `PBKDF2` way with `PKCS5_PBKDF2_HMAC`, which is supported since 1.0.0.

[1]: https://blog.filippo.io/the-scrypt-parameters/

dch:
cross-referencing https://github.com/PowerDNS/pdns/issues/10360 (allow unauthenticated requests for dashboard & lazy sysadmins)

rgacogne:
Some numbers using the `scrypt` KDF:
```
p=1, r8, N=32768
'Credentials hashing test' 0.16 seconds: 12.8 runs/s, 78345.99 usec/run
'Credentials verification test' 0.15 seconds: 13.2 runs/s, 75881.71 usec/run

p=1, r=8, N=16384
'Credentials hashing test' 0.13 seconds: 30.1 runs/s, 33222.99 usec/run
'Credentials verification test' 0.12 seconds: 34.2 runs/s, 29268.85 usec/run

p=1, r=8, N=8192
'Credentials hashing test' 0.12 seconds: 42.6 runs/s, 23479.36 usec/run
'Credentials verification test' 0.11 seconds: 46.1 runs/s, 21671.49 usec/run

p=1, r=8, N=4096
'Credentials hashing test' 0.12 seconds: 80.8 runs/s, 12382.01 usec/run
'Credentials verification test' 0.11 seconds: 85.1 runs/s, 11753.92 usec/run

p=1, r=8, N=1024
'Credentials hashing test' 0.11 seconds: 422.0 runs/s, 2369.60 usec/run
'Credentials verification test' 0.11 seconds: 433.5 runs/s, 2306.86 usec/run
```
Would it make sense to go for an implementation using scrypt with N=1024 for now? That's not great, but it's a lot better than what we have now in term of credentials security, and it seems like a nice middle ground between usability and security.
As a second step we can introduce a new endpoint delivering a token for heavy users of the API.

setharnold:
Yes, I think even `N=1024` makes a huge impact; hashcat sha1 bruteforcing does [billions of attempts every second](https://gist.github.com/epixoip/a83d38f412b4737e99bbef804a270c40). Even if someone had hardware that was a thousand times better at brute-forcing argon2 than you used here, it's still a huge improvement in brute force resistance.

omoerbeek:
Yeah, that shounds like a reasonable compromise.

rgacogne:
I just pushed a new version that uses `OpenSSL`'s `scrypt` support, when available, to support hashed and salted credentials. The default is not to hash anything, although the new code will try to do more constant-time comparison of credentials by first computing a burtle hash and comparing that, only comparing the full string if the hash matches (the comparison is constant time as long as the strings have the same lengths). The user can decide to hash the credentials (using the console, `pdnsutil` or `rec_control`) before putting them into the configuration, in which case they can decide the work factor they want (default to 1024 which is quite fast), being aware that there is a cost involved.

omoerbeek (approved):
Some small nits. Reads and tests good: existing plaintext password keep working as well.

On LibreSSL does not have the scrypt API, so this cannot be used there ATM. But do not let it hold you back!. I'll either move to OpenSSL or check if libsodium's scrypt API can be used as an alternative.

m4/pdns_check_libcrypto.m4
```diff
@@ -112,6 +112,7 @@ AC_DEFUN([PDNS_CHECK_LIBCRYPTO], [
         [
             AC_MSG_RESULT([yes])
             AC_CHECK_FUNCS([RAND_bytes RAND_pseudo_bytes CRYPTO_memcmp OPENSSL_init_crypto EVP_MD_CTX_new EVP_MD_CTX_free RSA_get0_key])
+            AC_CHECK_DECL(EVP_PKEY_CTX_set1_scrypt_salt, [AC_DEFINE([HAVE_EVP_PKEY_CTX_SET1_SCRYPT_SALT], [1], [Define to 1 if you have EVP_PKEY_CTX_set1_scrypt_saltEVP_PKEY_CTX_set1_scrypt_salt]) ], [], [#include <openssl/kdf.h>])
```
omoerbeek, line 115:
```suggestion
            AC_CHECK_DECL(EVP_PKEY_CTX_set1_scrypt_salt, [AC_DEFINE([HAVE_EVP_PKEY_CTX_SET1_SCRYPT_SALT], [1], [Define to 1 if you have EVP_PKEY_CTX_set1_scrypt_salt) ], [], [#include <openssl/kdf.h>])
```

pdns/rec_control.cc
```diff
@@ -106,9 +107,32 @@ int main(int argc, char** argv)
 
     sockname.append(".controlsocket");
 
-    rccS.connect(arg()["socket-dir"], sockname);
-
     const vector<string>&commands=arg().getCommands();
+
+    if (commands.size() >= 1 && commands.at(0) == "hash-password") {
```
omoerbeek, line 112: help missing in ` RecursorControlParser::getAnswer()`

rgacogne:
Oh, right, I did not notice that `scrypt` was implemented in `libsodium`. The nice API has a lower bound at 32768 for the work factor which would be an issue, but there is a low-level one that might work indeed (`crypto_pwhash_scryptsalsa208sha256_ll`).

pdns/misc.cc
```diff
@@ -1642,3 +1642,36 @@ size_t parseSVCBValueList(const std::string &in, vector<std::string> &val) {
...
+#ifdef HAVE_SODIUM_MEMCMP
+#include <sodium.h>
+#endif /* HAVE_SODIUM_MEMCMP */
+#endif /* HAVE_CRYPTO_MEMCMP */
+
+bool constantTimeStringEquals(const std::string& a, const std::string& b)
+{
+  if (a.size() != b.size()) {
+    return false;
+  }
```
cmouse, line 1658: this is not really constant =)

rgacogne: Did you look at how it is used, or even my comment above?

cmouse: Fair enough, I seem to have read it poorly.

omoerbeek:
Some words in the docs making it extra clear that you *can* use hashed passwords in the config files without setting the option would be nice. I suspect quite some readers will conclude that they need to set the option in order to be able to use hashed password in the config files.

rgacogne:
> Some words in the docs making it extra clear that you can use hashed passwords in the config files without setting the option would be nice. I suspect quite some readers will conclude that they need to set the option in order to be able to use hashed password in the config files.

Done!

rgacogne:
@omoerbeek @Habbie Would now a good time for me to merge this?

Habbie:
works for me!

rgacogne:
Rebased to fix a conflict, I'm not sure whether it's going to work out with the recent CI changes but we will find out :-)

docs/settings.rst
```diff
@@ -1771,6 +1774,18 @@ IP Address for webserver/API to listen on.
...
+.. _setting-webserver-hash-plaintext-credentials:
+
+``webserver-hash-plaintext-credentials``
+----------------------------------------
+..versionadded:: 4.6.0
+
+-  Boolean
+-  Default: no
+
+Whether passwords and API keys supplied in the configuration as plaintext should be hashed during startup, to prevent the plaintext versions from staying in memory. Doing so increases significantly the cost of verifying credentials and is thus disabled by default.
```
gibson042, line 1786: I'm concerned about this unsafe default; why is it even an option to keep plain-text passwords in memory? How "significant" is the cost in terms of absolute time and/or memory? Would it be more sensible to instead make the hashing algorithm and/or tunable parameters configurable for constrained hosts?

cmouse: This has been extensively discussed in earlier comments.

gibson042: I saw https://github.com/PowerDNS/pdns/pull/10157#issuecomment-816516664 and a couple of comments suggesting that a previous implementation imposed a slowdown of 110 to 170 milliseconds per API request, which hardly seem "extensive". Nor could I find any decision regarding my primary point here about the default behavior remaining insecure, although again there was discussion that seemed to peter out.

rgacogne: That's actually very expensive, given that simply loading the main page of the recursor does dozens of such queries, refreshes itself every second doing on average 2 requests per second. So that means we would spent half a second of the time of a CPU core to refresh that page, which in my book is a huge issue in a software whose main goal is to serve DNS queries. Does that sound like a good idea to you?
Now, the default is to use hashed webserver passwords and API keys if the administrator uses that in the configuration file. If they don't, keeping the plaintext in memory is not ideal, sure, but there isn't a lot of actual thread models where it does matter. An attacker would need to find some remote infoleak flaw and then, in most cases, they would be able to access read-only metrics and a random secret that is not used anywhere else. If you are worried about that, great, there is an option to hash these credentials in memory, just enable it.

gibson042: I think you misunderstood me. I said that a few comments' worth of conversation does not qualify as "extensive", not that a couple hundred milliseconds per request isn't _expensive_ (although making dozens of requests per page load seems like a problem in itself). But to rephrase my last question from above, why are the only options plain text vs. "huge issue" hashing? There's **lots** of software used for much more sensitive access control that gets by with far less computationally expensive authorization, and even in the DNS space BIND `rdnc` suffers from many other problems but avoids this one altogether.

rgacogne: Sure, how would you do it? I'd be happy to review your pull request anytime.

gibson042: Default to a suitably fast hashing algorithm and don't allow opting out, but support opting _in_ to use of intentionally expensive key-stretching alternatives such as scrypt. Since supporting SHA-256 is already a requirement for DNSSEC anyway, that seems like a good choice.

> I'd be happy to review your pull request anytime.

This pull request is almost there now; I'm just arguing for safer defaults. But is that approach even on the table? If not then there's no point in preparing it.

rgacogne: > Default to a suitably fast hashing algorithm and don't allow opting out, but support opting in to use of intentionally expensive key-stretching alternatives such as scrypt. Since supporting SHA-256 is already a requirement for DNSSEC anyway, that seems like a good choice.

That sounds good. One issue is that SHA-256 support is not a requirement for dnsdist so we need to deal with the case where it's not available. It might be nice to use `crypto_generichash` from libsodium when available, then SHA-256 from libcrypto, but we still need a fallback when none of this is available.

> This pull request is almost there now; I'm just arguing for safer defaults. But is that approach even on the table? If not then there's no point in preparing it.

Yes, I would be happy to have what you suggest. It would require writing more code than this PR currently have, so I might merge that PR first as it already improves things significantly, but I would definitely merge a suitable follow-up PR.

rgacogne:
I'm merging this PR now even though a nice improvement was suggested in the comments, but the improvement would take a non-negligible amount of work and this PR already improves the situation enough that I would like to have it in dnsdist 1.7.0.

PR #10160, opened 2021-03-10
rec: Structured logging

chbruyand:
### Short description

This pull request is a WIP on how we could move current logging to structured logging. This Logr interface is based on a minimal logging API originally written for Go, see https://github.com/go-logr/logr

I'm not satisfied with the usage of intermediate value objects, but this gives for now the ability to accept various types, to unify their string conversion and to prevent conversion to happen if verbosity is too high.

Comments welcome!

### Checklist

I have:
- [x] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/recursordist/logging.cc
+++ b/pdns/recursordist/logging.cc
@@ -79,7 +79,7 @@ namespace Logging
 
   std::shared_ptr<Logr::Logger> Logger::withValues(const std::string& key, const Logr::Loggable& value)
   {
-    auto res = std::make_shared<Logger>(getptr(), _name.get(), _level, _callback);
+    auto res = std::make_shared<Logger>(getptr(), _name, _level, _callback);
     res->_values.insert({key, value.to_string()});
     res->setVerbosity(getVerbosity());
     return res;
@@ -135,10 +135,10 @@ namespace Logging
   Logger::Logger(EntryLogger callback) : _callback(callback)
   {
   }
-  Logger::Logger(EntryLogger callback, boost::optional<const std::string> name) : _callback(callback), _name(name)
+  Logger::Logger(EntryLogger callback, boost::optional<std::string> name) : _callback(callback), _name(name)
   {
   }
-  Logger::Logger(std::shared_ptr<Logger> parent, boost::optional<const std::string> name, size_t lvl,  EntryLogger callback) : _parent(parent), _callback(callback), _name(name), _level(lvl)
+  Logger::Logger(std::shared_ptr<Logger> parent, boost::optional<std::string> name, size_t lvl,  EntryLogger callback) : _parent(parent), _callback(callback), _name(name), _level(lvl)
   {
   }
 
--- a/pdns/recursordist/logging.cc
+++ b/pdns/recursordist/logging.cc
@@ -32,23 +32,34 @@ namespace Logging
     return shared_from_this();
   }
 
-  bool Logger::enabled() const
+  bool Logger::enabled(Logr::Priority p) const
   {
-    return _level <= _verbosity;
+    return _level <= _verbosity || p != Logr::Absent;
   }
 
   void Logger::info(const std::string& msg) const
   {
-    logMessage(msg, boost::none);
+    logMessage(msg, Logr::Absent, boost::none);
+  }
+
+  void Logger::info(Logr::Priority p, const std::string& msg) const
+  {
+    logMessage(msg, p, boost::none);
   }
 
   void Logger::logMessage(const std::string& msg, boost::optional<const std::string> err) const
   {
-    if (!enabled()) {
+    return logMessage(msg, Logr::Absent, err);
+  }
+
+  void Logger::logMessage(const std::string& msg, Logr::Priority p, boost::optional<const std::string> err) const
+  {
+    if (!enabled(p)) {
       return ;
     }
     Entry entry;
     entry.level = _level;
+    entry.d_priority = p;
     Utility::gettimeofday(&entry.d_timestamp);
     entry.name = _name;
     entry.message = msg;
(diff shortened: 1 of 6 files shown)


rgacogne (commented):
I'm not familiar at all with structured logging so I don't think I'm the right person to comment on whether this makes sense, but I like the examples in this PR quite a lot, nice work!
I have a few remarks on the implementation, I believe we might be able to avoid some allocations and temporary objects, but let's keep that for later, when we have decided how we want to progress on that topic.

omoerbeek:
API
----
Is going in the right direction as far as I am concerned. Some small remarks:

- I would prefer the ability to add an errno as int argument to `error()`, and then convert it later into string. This would avoid all the stringerror() calls in the callers. Likely as last argument is preferable, so you can leave it out if no errno is available.
 
- I look forward to be able to pass objects to `Logger`, like `DNSName` or` QType`. Some more work needs to in this area, but the basics are there in the `Loggable` interface.
  
- I think we need to adopt the existing `Logger::Urgency` levels for the `v()` method. This would make it easier to do a gradual introduction of structured logging and also a syslog backend would need some well -defined values for that.
  
- An auto-generated high precision timestamps added to the entry would be nice, that is: a timestamp generated at the moment of logging, not at the moment of output.

Formatting and output
--

Formatting to a string representation and outputting to a backend should be separated. I think we should start with a backend for stdout or stderr and and one that logs to syslog. These can share the formatter. At startup we need to setup a Logger that writes to stdout or stderr that gets used until parsing of the config is done and then switch to the configured backend.

Tracing

I do not think the current approach is usable for tracing as-is. There you want to avoid that the all the arguments evaluated and pushed to the stack etc and then likely decide later that you want no
tracing for this query at all. Here we need some smarft use of inline function and/or macro. Also, I think in the future we want to be able to send  tracing information to a different backend than the one for general logging. I would love to be able to do rec_control trace ... and then get the trace on *my* tty. 

How to proceed?
--
Some thoughts how to move forward step by step:

1. Work on the proposed error api change(s)

2. Work with Remi to avoid copying etc

3. Implement code to use the existing rec Logger as backend (but only for output), formatting is done by the new mechanism). The idea being that the old and new mechanism exists together and that we start with using the new mechanism step-by-step. See also the section about formatting vs outputting above.

4. For tracing we could for the moment continue to use the existing mechanism, which basically logs the trace information to a stringbuffer and then (conditionally) logs that buffer.

omoerbeek:
@chbruyand @rgacogne any opinion? I might take a shot at doing some of the tasks listed above, if you agree they make sense.

omoerbeek:
So I did some of the changes suggested above (including a rebase, so a forced push)
- changed to eror function to alos accept an int errno
- Pubbed the existing logger as a backend.
- Always log as key-value pairs, including msg and err if applicable and quote the value part.
- Fixed a `const` error wrt `boost::optional<string>` vs `boost::optional<const string>` that caused some trouble passing the name to the `Logger` constructor.

This is all  tentative but shows they way some of the point above could be solved.
level <-> syslog urgency mapping (if wanted) has to be thought out better.

omoerbeek:
IMO with my latest commit we're getting close to this being usable and with a step by step migration path from old to new.

One thought: if we want to integrate tracing we need to be able to setup a logger separate form the main logger. THat logger should be associated with a particular **instance** of `SyncRes`. The tracing works in such a way that trace lines are collected, and then logged in one go, if certain conditions hold. In the future we also want to be able to send that trace information via other channels to the outside word, like `rec_control` or a separate logging channel.

A solution would be extend the "register a backend method" wit an object (reference/pointer), which gets passed to the backend each time it is called. In C it would be a `void *`, but since we're doing C++ there are better alternatives e.g. `boost::any`. That way we can make a logger backend specific to an instance of `SyncRes`. 

I hope I explained the issue and the potential solution clearly.

rgacogne:
> since we're doing C++ there are better alternatives e.g. `boost::any`

Perhaps dynamic dispatch might be good alternative as well? Of course we would need to be careful about the lifetime of such a reference.

pdns/rpzloader.cc
```diff
@@ -394,18 +399,18 @@ void RPZIXFRTracker(const std::vector<ComboAddress>& primaries, boost::optional<
...
-          dumpZoneToDisk(zoneName, newZone, dumpZoneFileName);
+          dumpZoneToDisk(logger, zoneName, newZone, dumpZoneFileName);
         }
 
         /* no need to try another primary */
         break;
       }
       catch(const std::exception& e) {
-        g_log<<Logger::Warning<<"Unable to load RPZ zone '"<<zoneName<<"' from '"<<primary<<"': '"<<e.what()<<"'. (Will try again in "<<refresh<<" seconds...)"<<endl;
+        logger->info(Logr::Warning, "Unable to load RPZ zone, will retry", "from", Logging::Loggable(primary), "exception", Logging::Loggable(e.what()), "refresh", Logging::Loggable(refresh));
```
chbruyand, line 409: What is the semantic difference between `logger->info(Logr::Warning,...)` and `logger->error(Logr::Warning,...` ?

omoerbeek: The latter has a required argument (either a `string err` or `int err`) that is supposed to hold the underlying OS-level error.

omoerbeek:
> > since we're doing C++ there are better alternatives e.g. `boost::any`
> 
> Perhaps dynamic dispatch might be good alternative as well? Of course we would need to be careful about the lifetime of such a reference.

We deided to postpone this feature until after this PR has been reviewed and merged.

pdns/recursordist/logging.cc
```diff
@@ -0,0 +1,179 @@
...
+  if (!enabled(p)) {
+    return;
+  }
+  Entry entry;
+  entry.level = _level;
+  entry.d_priority = p;
+  Utility::gettimeofday(&entry.d_timestamp);
+  entry.name = _name;
+  entry.message = msg;
+  entry.error = err;
```
rgacogne, line 66: Perhaps (but untested):
```suggestion
  entry.error = std::move(err);
```

pdns/recursordist/logr.hh
```diff
@@ -0,0 +1,167 @@
...
+
+namespace Logr
+{
+struct Loggable
+{
+  virtual std::string to_string() const = 0;
+};
+
+// In addition to level which specifies the amount of detail and is
+// structered so that a derived logger always has a higher level
```
rgacogne, line 38:
```suggestion
// structured so that a derived logger always has a higher level
```

pdns/recursordist/logr.hh
```diff
@@ -0,0 +1,167 @@
...
+// than its parent we also have a priority/urgency field that maps
+// to the same field of the old logger which in turns has a direct
+// mapping to syslog priority. This is done to make it easier to
+// move step by step to structured logging. We consider both level
+// and priorty to select which messages are logged, amnd a backend
```
rgacogne, line 43:
```suggestion
// and priority to select which messages are logged, and a backend
```

rgacogne (approved):

chbruyand:
@rgacogne fixed your comments, thanks for the review.

omoerbeek (approved):
LGTM!

PR #10161, opened 2021-03-10
"data" size is unreliable, start using RES, as it seem reliable and relavant

### LINKED ISSUE #7591: The memory statistics in 4.2.0 beta1 are wrong
- Program: Authoritative
 - Issue type: Bug

### Short description
There is a huge jump in memory usage between alpha1 and beta1 in metronome

image

Total memory usage on the system is lower than the reported PowerDNS usage.

```
[root@ns2 etc]# free -m
              total        used        free      shared  buff/cache   available
Mem:            991         195         605           6         190         630
Swap:           819           0         819
````

### Environment
 - Operating system: CentOS7
 - Software version:  4.2.0 beta1

omoerbeek:
According to https://www.kernel.org/doc/html/latest/filesystems/proc.html "data" aka "drs" is broken. 
In practice most of the time it shows a number somewhere between RES and VIRT, but it seems to be completely off in multi-threaded processed once in a while.

So move to the same number top(1) displays as "RES". This number is not marked broken and top(1) and the metrics agreed all the time in my tests.

Should fix #7591

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/misc.cc
+++ b/pdns/misc.cc
@@ -1270,7 +1270,9 @@ uint64_t getRealMemoryUsage(const std::string&)
   uint64_t size, resident, shared, text, lib, data;
   ifs >> size >> resident >> shared >> text >> lib >> data;
 
-  return data * getpagesize();
+  // We used to use "data" here, but it proves unreliable and even is marked "broken"
+  // in https://www.kernel.org/doc/html/latest/filesystems/proc.html 
+  return resident * getpagesize();
 #else
   struct rusage ru;
   if (getrusage(RUSAGE_SELF, &ru) != 0)


mnordhoff:
The color of the bikeshed concerns me! RES fluctuates based on swap usage, right?

That's potentially good (excluding bits of memory that are never used), but also potentially bad ("oh good, my memory usage is going down" vs. "the server is overloaded and PowerDNS is being swapped to death").

omoerbeek:
> The color of the bikeshed concerns me! RES fluctuates based on swap usage, right?
> 
> That's potentially good (excluding bits of memory that are never used), but also potentially bad ("oh good, my memory usage is going down" vs. "the server is overloaded and PowerDNS is being swapped to death").

Yes, RES is only one dimension of a complex equation. Making any evaluation and/or decision based on only such a one-dimension measure is unwise.

rgacogne (approved):

PR #10164, opened 2021-03-10
amazonlinux-2 packaging improvements

Habbie:
### Short description
Just to match their docs more closely. This does not switch the build to an actually different EPEL repo.

Sidenote: the auth build for amazonlinux-2 is currently broken, but only because of something between ixfrdist and yaml-cpp. I've decided to not investigate that right now.

### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/builder-support/dockerfiles/Dockerfile.target.amazon-2
+++ b/builder-support/dockerfiles/Dockerfile.target.amazon-2
@@ -5,7 +5,7 @@
 # Put only the bare minimum of common commands here, without dev tools
 FROM amazonlinux:2 as dist-base
 ARG BUILDER_CACHE_BUSTER=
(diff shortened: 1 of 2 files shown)


pieterlexis (approved):

pieterlexis:
> Sidenote: the auth build for amazonlinux-2 is currently broken, but only because of something between ixfrdist and yaml-cpp. I've decided to not investigate that right now.

fwiw: you could put a conditional in the specfile and not build ixfrdist on amazon linux

Habbie:
> fwiw: you could put a conditional in the specfile and not build ixfrdist on amazon linux

That makes sense - weakforced, which needs yaml-cpp more than we do, now buils yaml-cpp statically, but I don't want to go there.

Habbie:
> fwiw: you could put a conditional in the specfile and not build ixfrdist on amazon linux

done

pieterlexis (approved):

PR #10165, opened 2021-03-10
Install vim in docker-auth image

rytis:
### Short description

`pdnsutil edit-zone` requires `EDITOR` to be set. Currently there's no editor installed at all. Adding vim, so that it can be used as default editor by `pdnsutil edit-zone`.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/Dockerfile-auth
+++ b/Dockerfile-auth
@@ -71,8 +71,9 @@ FROM debian:10-slim
 # Reusable layer for base update - Should be cached from builder
 RUN apt-get update && apt-get -y dist-upgrade && apt-get clean
 
-# Ensure python3 and jinja2 is present (for startup script), and sqlite3 (for db schema), and tini (for signal management)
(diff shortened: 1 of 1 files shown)


Habbie:
Hello! This makes sense to me, but vim is 33 MB, bumping the image size by almost 20%. vim-tiny is 2MB which yields a much more reasonable 1% growth. If you change it to vim-tiny, I'm happy to merge this.

Habbie:
Also, can you check that it works reliably? If not, we may need to set EDITOR=vim in the environment.

rytis:
Hi! Good point, I haven't thought about that. :) I've updated with `vim-tiny` now. The build works fine, and `pdnsutil edit-zone` launches the editor as well.

Habbie (approved):

Habbie:
> `pdnsutil edit-zone` launches the editor as well.

Yep, works for me too!

PR #10169, opened 2021-03-11
check secpoll syntax

Habbie:
### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/.github/workflows/secpoll.yml
+++ b/.github/workflows/secpoll.yml
@@ -1,0 +1,26 @@
+---
+name: 'Verify secpoll zone syntax'
+
+on:
+  push:
+  pull_request:
+
+jobs:
+  build:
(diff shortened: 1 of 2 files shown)


.github/workflows/secpoll.yml
```diff
@@ -0,0 +1,25 @@
+---
+name: 'Verify secpoll zone syntax'
+
+on:
+  push:
+  pull_request:
+
+jobs:
```
pieterlexis, line 8: it would be nice to run this only on changes to the secpoll?

Habbie: It's only a 50 second job so I did not bother restricting it, but we could.

pieterlexis (approved):

PR #10171, opened 2021-03-11
lua: don't destroy keys during table iteration

Habbie:
via https://github.com/ahupowerdns/luawrapper/pull/38

### Short description
This avoids a dnsdist crash when code returns an array-shaped table (i.e. `return {1,2,3}`).

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/ext/luawrapper/include/LuaContext.hpp
+++ b/ext/luawrapper/include/LuaContext.hpp
@@ -2690,12 +2690,12 @@ struct LuaContext::Reader<std::string>
         size_t len;
         const auto val = lua_tolstring(state, -1, &len);
 
-        if (val != 0)
+        if (val != nullptr)
           result.assign(val, len);
 
         lua_pop(state, 1);
 
-        return val != 0 ? boost::optional<std::string>{ std::move(result) } : boost::none;
+        return val != nullptr ? boost::optional<std::string>{ std::move(result) } : boost::none;
     }
 };
 
--- a/ext/luawrapper/include/LuaContext.hpp
+++ b/ext/luawrapper/include/LuaContext.hpp
@@ -2681,11 +2681,21 @@ struct LuaContext::Reader<std::string>
     static auto read(lua_State* state, int index)
         -> boost::optional<std::string>
     {
+        std::string result;
+
+        // lua_tolstring might convert the variable that would confuse lua_next, so we
+        //   make a copy of the variable.
+        lua_pushvalue(state, index);
+
         size_t len;
-        const auto val = lua_tolstring(state, index, &len);
-        if (val == 0)
-            return boost::none;
-        return std::string(val, len);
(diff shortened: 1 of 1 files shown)


ext/luawrapper/include/LuaContext.hpp
```diff
@@ -2681,11 +2681,21 @@ struct LuaContext::Reader<std::string>
...
+        lua_pushvalue(state, index);
+
         size_t len;
-        const auto val = lua_tolstring(state, index, &len);
-        if (val == 0)
-            return boost::none;
-        return std::string(val, len);
+        const auto val = lua_tolstring(state, -1, &len);
+
+        if (val != 0)
```
cmouse, line 2693: `val != nulltpr`

ext/luawrapper/include/LuaContext.hpp
```diff
@@ -2681,11 +2681,21 @@ struct LuaContext::Reader<std::string>
...
+        // lua_tolstring might convert the variable that would confuse lua_next, so we
+        //   make a copy of the variable.
+        lua_pushvalue(state, index);
+
         size_t len;
-        const auto val = lua_tolstring(state, index, &len);
-        if (val == 0)
-            return boost::none;
-        return std::string(val, len);
+        const auto val = lua_tolstring(state, -1, &len);
```
cmouse, line 2691: `const auto& val`

ext/luawrapper/include/LuaContext.hpp
```diff
@@ -2681,11 +2681,21 @@ struct LuaContext::Reader<std::string>
...
+
+        if (val != 0)
+          result.assign(val, len);
+
+        lua_pop(state, 1);
+
+        return val != 0 ? boost::optional<std::string>{ std::move(result) } : boost::none;
```
cmouse, line 2698: `val != nullptr` here as well.

Habbie:
@cmouse thanks, your changes make sense - I'll upstream those first to go through the LuaWrapper testsuite.

ext/luawrapper/include/LuaContext.hpp
Habbie: done

ext/luawrapper/include/LuaContext.hpp
Habbie: skipped, it's a pointer already

omoerbeek (approved):
An alternative I like a little better: (it has two pop calls which isn't very nice either):
But approved either way.

```
if (val != nullptr) {
  std::string result;
  result.assign(val, len);
  lua_pop(state, 1);
  return boost::optional<std::string>{ std::move(result) };
}
else {
  lua_pop(state, 1);
  return boost::none;
}
```

PR #10178, opened 2021-03-16
rec: Don't pick up random root NS records from AUTHORITY sections

### LINKED ISSUE #10125: rec: do not take root NSset authority updates from random packets
Background:
* https://lists.dns-oarc.net/pipermail/dns-operations/2021-February/020924.html ('Broken A and J root responses')
* https://lists.dns-oarc.net/pipermail/dns-operations/2021-February/020927.html ('Incorrect NSEC responses from Verisign root server instances')

It looks like Recursor accepts the new root NSset from those bad responses. While this makes sense in terms of our whole implementation, and various words from RFC2181, I don't think it makes sense to accept 'authority updates' for the root NSset. Such updates would always fall into any of these 3 classes:
1. it's the root NSset we saw on the last prime
2. it's a new root NSset we will see on the next prime
3. it's garbage

Secondary to that, i wonder if we could/should be doing any DNSSEC validation on authority updates (in general, outside the root), and on our root priming.

omoerbeek:
### Short description

Fixes #10125 

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [X] documented the code

### DIFF
--- a/pdns/syncres.cc
+++ b/pdns/syncres.cc
@@ -2970,18 +2970,21 @@ void SyncRes::sanitizeRecords(const std::string& prefix, LWResult& lwr, const DN
     }
 
     if (rec->d_place == DNSResourceRecord::AUTHORITY && rec->d_type == QType::NS && (isNXDomain || isNXQType)) {
-      /* we don't want to pick up NS records in AUTHORITY and their ADDITIONAL sections of NXDomain answers
-         because they are somewhat easy to insert into a large, fragmented UDP response
-         for an off-path attacker by injecting spoofed UDP fragments.
-      */
+      /*
+       * We don't want to pick up NS records in AUTHORITY and their ADDITIONAL sections of NXDomain answers
+       * because they are somewhat easy to insert into a large, fragmented UDP response
+       * for an off-path attacker by injecting spoofed UDP fragments. So do not add these to allowedAdditionals.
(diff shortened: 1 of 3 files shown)


pdns/syncres.cc
```diff
@@ -2979,6 +2979,14 @@ void SyncRes::sanitizeRecords(const std::string& prefix, LWResult& lwr, const DN
       continue;
     }
 
+    if (rec->d_place == DNSResourceRecord::AUTHORITY && rec->d_type == QType::NS && !d_updatingRootNS && rec->d_name == g_rootdnsname) {
+      /* we don't want to pick up NS records in AUTHORITY and their ADDITIONALs sections of random queries
```
Habbie, line 2983: Do you want to put the word 'root' in this comment, like in the PR title?

Also, the comment says AUTHORITY+ADDITIONAL but the code only says AUTHORITY?

omoerbeek: The `if (...AUTHORITY...)` checks determine what gets handled in the `if (...  ADDITIONAL ...)` block by manipulating `allowedAdditionals`. So while the code only checks for AUTHORITY it does impact how ADDITIONAL records are handled. I will change the comment to make it a bit more clear.

rgacogne (approved):

PR #8639, opened 2019-12-16
rec: debian postinst / do not fail on user creation if it already exists

captainark:
### Short description

We manage the `pdns` user on our Debian systems manually, and we are having issues with the pdns-recursor postinst script failing because we use a non-system uid for it.

I simply copied the if statements [from the pdns-server postinst script](https://github.com/PowerDNS/pdns/blob/e963f425a360f6348102208308ffdcd24e8d5e09/builder-support/debian/authoritative/debian-buster/pdns-server.postinst) to the pdns-recursor ones so that the script doesn't try to create the user if it already exists.

### Checklist
I have:
- [x] read the [CONTRIBUTING.md](https://github.com/PowerDNS/pdns/blob/master/CONTRIBUTING.md) document
- [ ] compiled this code
- [x] tested this code
- [ ] included documentation (including possible behaviour changes)
- [ ] documented the code
- [ ] added or modified regression test(s)
- [ ] added or modified unit test(s)

PR #10180, opened 2021-03-16
auth: debian postinst / do not fail on user creation if it already exists

thibmac:
### Short description
Since pdns-server version 4.4.1, package configuration on Debian systems already provided with a system pdns user fails, which makes package upgrade impossible.
This PR provides a fix preventing package configuration from aborting at the groupe and user creation in case it already exists.

A similar PR can be found here on the recursor https://github.com/PowerDNS/pdns/pull/8639

### Checklist

I have:
- [ ] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/builder-support/debian/authoritative/debian-buster/pdns-server.postinst
+++ b/builder-support/debian/authoritative/debian-buster/pdns-server.postinst
@@ -8,8 +8,8 @@ fi
 
 case "$1" in
   configure)
-    addgroup --quiet --system pdns
-    adduser --quiet --system --home /var/spool/powerdns --shell /bin/false --ingroup pdns --disabled-password --disabled-login --gecos "PowerDNS" pdns
+    addgroup --quiet --system pdns || true
+    adduser --quiet --system --home /var/spool/powerdns --shell /bin/false --ingroup pdns --disabled-password --disabled-login --gecos "PowerDNS" pdns || true
     chown root:pdns /etc/powerdns/pdns.conf || true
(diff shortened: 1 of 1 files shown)


zeha:
The pdns scripts expect that the pdns user is a "system" user (i.e. lives in the system uid space). If this is the case, adduser continues without changing the user or failing.

Habbie:
@thibmac any comments on what @zeha said?

thibmac:
On our installation, we need to have the pdns user uid/gid to be predictible for audit purposes, this is why we need to set up the user and group before installing the package

I could propose replacing this :

```
    adduser --quiet --system --home /var/spool/powerdns --shell /bin/false --ingroup pdns --disabled-password --disabled-login --gecos "PowerDNS" pdns || true

```

by this :

```    
    if ! grep -q "^pdns:" /etc/passwd; then
      adduser --quiet --system --home /var/spool/powerdns --shell /bin/false --ingroup pdns --disabled-password --disabled-login --gecos "PowerDNS" pdns
    else
      [ `id -u pdns` -lt `grep -Po "^UID_MIN\s+\K\d+$" /etc/login.defs` ] || exit 2
    fi
```
Would that work for everyone?

Habbie:
I'm missing something - either your `pdns` UID is in the system space, and there never was a problem, or it's not, and you will trigger the `exit 2` on your systems?

Habbie:
Ah! That makes sense to me. Closing this then, thank you!

PR #10182, opened 2021-03-17
rec: Insert hints as non-auth into cache

### LINKED ISSUE #10177: rec: smarter priming
I think there is room for improvement wrt priming: currently both NS, A and AAAA records from the hints are inserted into the record cache with a TTL of 1000 hours in `primeHints()`. After that, `SyncRes::getRootNS()` is periodically called and resolves the root NS set (and will find it is in the cache most of the time), but no explicit queries are done for the A/AAAA's of the root servers and no cache entries for them are updated with facts learned from the net, at least not for the first 1000 hours.

I tested this with modifying `root-addreses.hh` to only contains a few correct IPs of NS and indeed, the wrong addresses were not updated in the cache.

It is probably better to move to an approach where the hints are used to learn the NS and their addresses from the net, and make sure those learned records are inserted into the cache. As a bonus I think it would be good to have the address records of the root servers not all have the same TTD as they have now, when refreshing them it will be possible to use the non-expired ones, and so no need to fall back to the hints most of the time.

 - Program: Recursor
 - Issue type: Feature request

omoerbeek:
Insert hints as non-auth into cache, so info received from the net is recorded in the cache.

Also make sure the root NS refresh happens more often if `max-cache-ttl` is low.
This is needed as the records no longer maintain the 1000 hours TTL.
In the existing setup, a reprime (with potential outdated info) was done at that
point in time since all root-server address records would expire at the same
time.

Lastly, fix a infinite (caught by depth check) recursion in getBestNSFromCache().

Fixes #10177.

As a bonus, reformat `reczones.cc` to make it easier on the eyes. (Ignore whitespace to make the diff easier).

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [X] documented the code

### DIFF
--- a/pdns/reczones.cc
+++ b/pdns/reczones.cc
@@ -75,13 +75,13 @@ bool primeHints(time_t ignored)
       aset.push_back(arr);
       /*
        * Originally the hint records were inserted with the auth flag set, with the consequence that data from AUTHORITY and
-       * ADDITIONAL sections (as seen in a ,. NS response) were not used. This (together with the long ttl) caused outdated
+       * ADDITIONAL sections (as seen in a . NS response) were not used. This (together with the long ttl) caused outdated
        * hint to be kept in cache. So insert as non-auth, and the extra sections in the . NS refreshing cause the cached
        * records to be updated with up-to-date information received from a real root server.
        *
        * Note that if a user query is done for one of the root-server.net names, it will be inserted into the cache with the
        * auth bit set. Further NS refreshes will not update that entry. If all root names are queried at the same time by a user,
-       * all root-server.net names will be marked auth and will expired at the same time. A re-prime is then triggered,
+       * all root-server.net names will be marked auth and will expire at the same time. A re-prime is then triggered,
        * as before, when the records were inserted with the auth bit set and the TTD comes.
        */
       g_recCache->replace(now, DNSName(templ), QType(QType::A), aset, vector<std::shared_ptr<RRSIGRecordContent>>(), vector<std::shared_ptr<DNSRecord>>(), false, g_rootdnsname, boost::none, boost::none, validationState, from); // auth, nuke it all
--- a/pdns/reczones.cc
+++ b/pdns/reczones.cc
@@ -34,7 +34,8 @@ extern char** g_argv;
 
 static thread_local set<DNSName> t_rootNSZones;
 
-static void insertIntoRootNSZones(const DNSName &name) {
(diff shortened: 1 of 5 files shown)


pdns/reczones.cc
```diff
@@ -50,34 +51,48 @@ bool primeHints(time_t ignored)
...
+      nsrr.d_content = std::make_shared<NSRecordContent>(DNSName(templ));
+      arr.d_content = std::make_shared<ARecordContent>(ComboAddress(rootIps4[c - 'a']));
       vector<DNSRecord> aset;
       aset.push_back(arr);
-      g_recCache->replace(time(0), DNSName(templ), QType(QType::A), aset, vector<std::shared_ptr<RRSIGRecordContent>>(), vector<std::shared_ptr<DNSRecord>>(), true, g_rootdnsname, boost::none, boost::none, validationState, from); // auth, nuke it all
-      if (rootIps6[c-'a'] != NULL) {
-        aaaarr.d_content=std::make_shared<AAAARecordContent>(ComboAddress(rootIps6[c-'a']));
+      /*
+       * Originally the hint records were inserted with the auth flag set, with the consequence that data from AUTHORITY and
+       * ADDITIONAL sections (as seen in a ,. NS response) were not used. This (together with the long ttl) caused outdated
```
rgacogne, line 78:
```suggestion
       * ADDITIONAL sections (as seen in a . NS response) were not used. This (together with the long ttl) caused outdated
```

pdns/reczones.cc
```diff
@@ -50,34 +51,48 @@ bool primeHints(time_t ignored)
...
+       * hint to be kept in cache. So insert as non-auth, and the extra sections in the . NS refreshing cause the cached
+       * records to be updated with up-to-date information received from a real root server.
+       *
+       * Note that if a user query is done for one of the root-server.net names, it will be inserted into the cache with the
+       * auth bit set. Further NS refreshes will not update that entry. If all root names are queried at the same time by a user,
+       * all root-server.net names will be marked auth and will expired at the same time. A re-prime is then triggered,
```
rgacogne, line 84:
```suggestion
       * all root-server.net names will be marked auth and will expire at the same time. A re-prime is then triggered,
```

rgacogne (approved):
Looks good! I wouldn't mind a unit test or two checking that we do actually refresh the `A` and `AAAA` now.

rgacogne (approved):

PR #10185, opened 2021-03-17
rec: make sure we take the right minimum for the PC TTL data.

omoerbeek:
Also add safety belt to the ageing code to not wrap TTLs.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdist-cache.cc
+++ b/pdns/dnsdist-cache.cc
@@ -285,7 +285,13 @@ bool DNSDistPacketCache::get(DNSQuestion& dq, uint16_t queryId, uint32_t* keyOut
   }
 
   if (!d_dontAge && !skipAging) {
-    ageDNSPacket(reinterpret_cast<char *>(&response[0]), response.size(), age);
+    if (!stale) {
+      ageDNSPacket(reinterpret_cast<char *>(&response[0]), response.size(), age);
+    }
+    else {
+      editDNSPacketTTL(reinterpret_cast<char *>(&response[0]), response.size(),
+        [staleTTL = d_staleTTL](uint8_t section, uint16_t class_, uint16_t type, uint32_t ttl) { return staleTTL; });
+    }
   }
 
   d_hits++;
--- a/pdns/dnsparser.hh
+++ b/pdns/dnsparser.hh
@@ -514,7 +514,11 @@ public:
     uint32_t tmp;
     memcpy(&tmp, (void*) p, sizeof(tmp));
     tmp = ntohl(tmp);
-    tmp-=decrease;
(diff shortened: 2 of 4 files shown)


Habbie (changes requested):
Taking just the patch from pdns_recursor.cc does not fix the issue for me, and I don't see why it would..

Habbie:
The inline TTL calculation was needlessly complex before, and this PR does not make that better; and probably because of that, it manages to miss a case.

Proposal: https://gist.github.com/Habbie/5e8a2ea5187cfe02ec74be8590db538e

Habbie:
Correction! The inline TTL calculation confused me to the point that I only took one line from the patch instead of two (thus stacking `min()` calls the wrong way). The PR does solve the issue.

However, the PR looks wrong for the case where packet cache TTL < packet cache servfail TTL. My proposed version covers that, and improves readability.

omoerbeek:
> Correction! The inline TTL calculation confused me to the point that I only took one line from the patch instead of two (thus stacking `min()` calls the wrong way). The PR does solve the issue.
> 
> However, the PR looks wrong for the case where packet cache TTL < packet cache servfail TTL. My proposed version covers that, and improves readability.

Thers is code to ensure `packetcacheservfailttl` <= `s_packetcachettl` (line 4603)

Habbie:
> There is code to ensure `packetcacheservfailttl` <= `s_packetcachettl` (line 4603)

Right! Then your version was, in fact, correct. We can take my readability improvements, and we could decide to leave what now appears to be an additional belt/suspender from me in, or take that out.

Habbie (approved):

Habbie:
> test does not like the new safety belt

I pushed a fix for the test; please check the comment (and the change); we need to squash things so we don't end up with a broken commit in our history.

rgacogne (approved):
Looks good to me once the last commits have been squashed!

Habbie:
Squashed & tested locally once more.

Habbie:
4 failing checks are unrelated (auth mysql)

Habbie:
More dnsdist failures.

Habbie:
```
FAIL: Cache: Cache entry, set backend down, get stale entry
  File "/opt/project/regression-tests.dnsdist/test_Caching.py", line 1086, in testCacheStale
    self.assertEquals(an.ttl, self._staleCacheTTL)
AssertionError: 0 != 60
```

Habbie:
@rgacogne I'm going to need another review from you :)

rgacogne (approved):

Habbie:
Thanks! Squashing again :)

PR #10196, opened 2021-03-19
auth: implement AXFR priority level queue code

### LINKED ISSUE #10190: auth: Have multiple priority levels for AXFR's
- Program: Authoritative
 - Issue type: Feature request

### Short description

We want API or NOTIFY initiated AXFR's to be executed before AXFR's triggered by the SOA refresh update cycle.

### Usecase
We are dealing with a large set of domains (1M+) and the weekly signature refresh occasionally causes long congestion in the AXFR queue of our slaves, which means the "normal" AXFR's to fetch updates are delayed.

We would like to be able to give a higher priority (IE do those first) to AXFR's that we initiate manually or via NOTIFY.

### Description

I think the simplest implementation would be to have a sort of priority system in the waiting AXFR's with the highest priority are executed before any of the lower priority AXFR's. Within each priority level the execution order should still be FIFO. Also it should be possible to insert a queue item with a higher priority if a lower priority item already exists. The lower priority item should then be removed.

RobinGeuze:
### Short description

This PR makes it possible to have AXFR requests of a different priority, where requests with the highest priority get handled first. closes #10190

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/docs/modes-of-operation.rst
+++ b/docs/modes-of-operation.rst
@@ -140,6 +140,12 @@ Slave operation can also be programmed using several
 command is especially useful as it triggers an immediate retrieval of
 the zone from the configured master.
 
+Zone transfers are added to a queue and processed according to priority
+and order of addition. Order levels are (from high to low): pdns control,
+api, notify, serial changed during refresh and signatures changed during
+refresh. High priority zone transfers are always processed first, in a 
+first in first out order.
+
 PowerDNS supports multiple masters. For the BIND backend, the native
 BIND configuration language suffices to specify multiple masters, for
 SQL based backends, list all master servers separated by commas in the
--- a/docs/modes-of-operation.rst
+++ b/docs/modes-of-operation.rst
@@ -140,7 +140,7 @@ Slave operation can also be programmed using several
 command is especially useful as it triggers an immediate retrieval of
 the zone from the configured master.
 
-Zone transfers are added to a queue and processed according to priority
+Since 4.5.0, zone transfers are added to a queue and processed according to priority
 and order of addition. Order levels are (from high to low): pdns control,
 api, notify, serial changed during refresh and signatures changed during
 refresh. High priority zone transfers are always processed first, in a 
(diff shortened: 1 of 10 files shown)


pdns/communicator.hh
```diff
@@ -46,22 +46,36 @@ struct SuckRequest
   DNSName domain;
   ComboAddress master;
   bool force;
+  uint8_t priority;
```
omoerbeek, line 49: YOu might want to make these a pair, so you don't have to define a custom compare function

RobinGeuze: Changed it to a pair :D

pdns/slavecommunicator.cc
```diff
@@ -48,20 +48,34 @@
...
+  sr.priorityAndOrder.second = d_sorthelper++;
   pair<UniQueue::iterator, bool>  res;
 
-  res=d_suckdomains.push_back(sr);
+  res=d_suckdomains.insert(sr);
   if(res.second) {
     d_suck_sem.post();
+  } else {
+    // Domain is already in there, we should check the priority and whether its forced
+    domains_by_name_t& nameindex=boost::multi_index::get<IDTag>(d_suckdomains);
```
rgacogne, line 67: Just a nit but if I'm not mistaken you could use `project()` (`d_suckdomains.project<IDTag>(res.first);`) to avoid doing a second lookup.

RobinGeuze: I realized I didn't even need the project, I could just call modify on the original iter directly.

docs/modes-of-operation.rst
```diff
@@ -140,6 +140,12 @@ Slave operation can also be programmed using several
 command is especially useful as it triggers an immediate retrieval of
 the zone from the configured master.
 
+Zone transfers are added to a queue and processed according to priority
```
rgacogne, line 143: The documentation covers more than the current version, so perhaps something like:
```suggestion
Since 4.5.0, zone transfers are added to a queue and processed according to priority
```

mind04 (approved):

omoerbeek (approved):

Habbie (approved):

PR #10206, opened 2021-03-22
API: return RRType when finding a comment

pieterlexis:
### Short description
Does what it says on the tin.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


Habbie (changes requested):
Looks good, but how about a test?

pieterlexis:
> Looks good, but how about a test?

added a test!

Habbie:
> added a test!

Nice! I don't think we had any tests for comments in search at all!

pieterlexis:
> > added a test!
> 
> Nice! I don't think we had any tests for comments in search at all!

correct!

Habbie (approved):

PR #10208, opened 2021-03-23
dnsdist: Fix the handling of DoH queries with a non-zero ID

rgacogne:
### Short description

rfc8484 states that clients "SHOULD use a DNS ID of 0 in every DNS request", not MUST, and it does indeed happen.
The issue was introduced in 341d2553b74c579df9d9843959f3ca6f5c3dc954 (between 1.5.x and 1.6.0-alpha1) when we moved to a safer PacketBuffer.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdistdist/doh.cc
+++ b/pdns/dnsdistdist/doh.cc
@@ -535,7 +535,7 @@ static int processDOHQuery(DOHUnit* du)
     ids->du = du;
 
     ids->cs = &cs;
-    ids->origID = queryId;
+    ids->origID = htons(queryId);
     setIDStateFromDNSQuestion(*ids, dq, std::move(qname));
 
     dq.getHeader()->id = idOffset;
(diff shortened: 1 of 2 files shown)


omoerbeek (approved):

jedisct1:
Hi!

Unfortunately, the query id issue may still be partially present in alpha 3.

A query to an uncached name seems to return a response with the wrong query id.

Identical queries sent afterwards (which I imagine are cached) get a correct query id.

jedisct1:
Scratch that. This is happening on Bortzmeyer's server, but I couldn't reproduce it locally.

rgacogne:
I know you figured that out in the meantime but in case anyone else stumbled upon that issue, that behaviour is the one from 1.6.0-alpha1 to alpha2 included, but it should be fixed in alpha3.

PR #10210, opened 2021-03-24
rec: Setup env properly to skip v6 test.

omoerbeek:
CicleCI docs indeed suggest the environment setting under docker do not apply
to the jobs steps.

So specify it explicitly for the test run itself.

### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1275,7 +1275,6 @@ jobs:
         environment:
           UBSAN_OPTIONS: 'print_stacktrace=1:halt_on_error=1'
           ASAN_OPTIONS: detect_leaks=0
-          SKIP_IPV6_TESTS: y
     steps:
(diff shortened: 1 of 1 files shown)


rgacogne:
I wonder whether `UBSAN_OPTIONS` and `ASAN_OPTIONS` are applied, then? On the other hand I think we would see errors if `ASAN_OPTIONS` was not actually applied?

omoerbeek:
> I wonder whether `UBSAN_OPTIONS` and `ASAN_OPTIONS` are applied, then? On the other hand I think we would see errors if `ASAN_OPTIONS` was not actually applied?

It's a mystery so far... the tests ran fine when the padding PR was still a PR... but once merged, it started to show trouble..

PR #10217, opened 2021-03-25
auth API cryptokeys: add cds array when configured to do so. Closes #10215

### LINKED ISSUE #10215: auth api: honor PUBLISH-CDS in response of /cryptokeys endpoint
- Program: Authoritative
 - Issue type: Feature request

### Short description
The `ds` JSON field of each key returned by the `/servers/{server_id}/zones/{zone_id}/cryptokeys` endpoint should honor the hash algorithms given by `PUBLISH-CDS` (or `default-publish-cds`).

### Usecase
Typically, the `ds` entries given in the JSON response are conveyed to the registry (for initial DNSSEC bootstrapping). As such, the returned hash algorithms can be expected to be the same as the ones in CDS records (for later key transitions).

We return the `ds` entries to our users, and don't want to return algorithm 1. We therefore have set `default-publish-cds=2,4` (which works for CDS), but we have to do manual filtering in our application when conveying the initial DS RRset.

### Description
See above. - I was unsure if this could be considered a bug, but I decided against file it as such.

Note that one could argue that `ds` should be renamed to `cds` if `PUBLISH-CDS` is honored. I would oppose that change, and instead take the algorithm list of `PUBLISH-CDS` to mean "list of hash algorithms that should get promoted into the DS RRset eventually". For the initial DS RRset (when bootstrapping the delegation), it would then be fair to honor this setting without renaming the field in the JSON schema.

Habbie:
### Short description
When CDS publication for a zone is configured, also add a `"cds"` entry to the auth API cryptokeys object.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/docs/http-api/swagger/authoritative-api-swagger.yaml
+++ b/docs/http-api/swagger/authoritative-api-swagger.yaml
@@ -1262,6 +1262,11 @@ definitions:
         items:
           type: string
         description: 'An array of DS records for this key'
+      cds:
+        type: array
+        items:
+          type: string
+        description: 'An array of DS records for this key, filtered by CDS publication settings'
       privatekey:
         type: string
         description: 'The private key in ISC format'
--- a/pdns/ws-auth.cc
+++ b/pdns/ws-auth.cc
@@ -1095,13 +1095,34 @@ static void apiZoneCryptokeysGET(const DNSName& zonename, int inquireKeyId, Http
         { "bits", value.first.getKey()->getBits() }
     };
 
+    string publishCDS;
+    dk->getPublishCDS(zonename, publishCDS);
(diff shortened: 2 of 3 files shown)


pieterlexis (changes requested):
No tests, no bueno

zeha:
+swagger I guess?

Habbie:
> +swagger I guess?

I did that, but in a copy generated during build, I see now. Thanks!

Habbie:
Test added, plus some calls to `clearMetaCache`.

pieterlexis (approved):
looks good and has a test, I like

PR #10224, opened 2021-03-26
rec: Backport 9883 to rec 4.4.x: Lookup DS entries before CNAME entries

omoerbeek:
### Short description

Backport of #9883 

This is a largish backport, so some extra careful review appreciated!

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/recursor_cache.cc
+++ b/pdns/recursor_cache.cc
@@ -111,7 +111,7 @@ static void updateDNSSECValidationStateFromCache(boost::optional<vState>& state,
   }
 }
 
-int32_t MemRecursorCache::handleHit(MapCombo& map, MemRecursorCache::OrderedTagIterator_t& entry, const DNSName& qname, vector<DNSRecord>* res, vector<std::shared_ptr<RRSIGRecordContent>>* signatures, std::vector<std::shared_ptr<DNSRecord>>* authorityRecs, bool* variable, boost::optional<vState>& state, bool* wasAuth)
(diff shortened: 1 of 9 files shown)


pdns/recursor_cache.cc
```diff
@@ -434,19 +436,25 @@ void MemRecursorCache::replace(time_t now, const DNSName &qname, const QType& qt
...
   }
 
+  ce.d_signatures = signatures;
+  ce.d_authorityRecs = authorityRecs;
   ce.d_records.clear();
   ce.d_records.reserve(content.size());
+  ce.d_authZone = authZone;
+  if (from) {
+    ce.d_from = *from;
+  }
```
rgacogne, line 455: I see we always pass `from` in practice, but if we ever do not we might leave the existing `from` value in place when replacing an entry. Perhaps we should clear it when `from` is `boost::none`?

pdns/reczones.cc
```diff
@@ -46,6 +46,7 @@ bool primeHints(void)
 {
   // prime root cache
   const vState validationState = vState::Insecure;
+  const ComboAddress from("255.255.255.255");
```
rgacogne, line 49: Looks like we don't use `from` here, while we do on master?

pdns/recursor_cache.cc
omoerbeek: Right, should be ported to master and 4.5x as well

rgacogne (approved):

PR #10238, opened 2021-03-29
rec: More fail-safe handling of NOD files

omoerbeek:
### Short description

Current code does not take into account file operation failures in a safe way.

This PR makes sure we write to a temp file and rename only after a succesful write, plus we do not crash on corrupt files, we remove them.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/nod.cc
+++ b/pdns/nod.cc
@@ -86,7 +86,9 @@ bool PersistentSBF::init(bool ignore_pid) {
             remove(newest_file);
           }
           catch (const std::runtime_error& e) {
-            g_log<<Logger::Warning<<"NODDB init: Cannot parse file: " << filename << endl;
+            infile.close();
+            remove(newest_file);
+            g_log<<Logger::Warning<<"NODDB init: Cannot parse file: " << filename << ": " << e.what() << "; removed" << endl;
           }
         }
       }
@@ -124,11 +126,13 @@ bool PersistentSBF::snapshotCurrent(std::thread::id tid)
     std::stringstream ss;
     ss << d_prefix << "_" << tid;
     f /= ss.str() + "_" + std::to_string(getpid()) + "." + bf_suffix;
+    path ftmp = f;
+    ftmp += ".tmp";
     if (exists(p) && is_directory(p)) {
       try {
         std::ofstream ofile;
         std::stringstream iss;
-        ofile.open(f.string(), std::ios::out | std::ios::binary);
+        ofile.open(ftmp.string(), std::ios::out | std::ios::binary);
         {
           // only lock while dumping to a stringstream
           std::lock_guard<std::mutex> lock(d_sbf_mutex);
@@ -137,8 +141,13 @@ bool PersistentSBF::snapshotCurrent(std::thread::id tid)
         // Now write it out to the file
         ofile << iss.str();
 
-        if (ofile.fail())
-          throw std::runtime_error("Failed to write to file:" + f.string());
+        if (ofile.fail()) {
+          ofile.close();
+          remove(ftmp);
+          throw std::runtime_error("Failed to write to file:" + ftmp.string());
+        }
+        ofile.close();
+        rename(ftmp, f);
         return true;
       }
       catch (const std::runtime_error& e) {
--- a/pdns/nod.cc
+++ b/pdns/nod.cc
@@ -28,8 +28,7 @@
 #include <ctime>
 #include <thread>
 #include "threadname.hh"
-#include <unistd.h>
-#include <boost/filesystem.hpp>
+#include <stdlib.h>
 #include "logger.hh"
 #include "misc.hh"
 
@@ -40,10 +39,9 @@ namespace filesystem = boost::filesystem;
 
 std::mutex PersistentSBF::d_cachedir_mutex;
 
-void PersistentSBF::remove_tmp_files()
+void PersistentSBF::remove_tmp_files(const filesystem::path& p, std::lock_guard<std::mutex>& lock)
 {
-  filesystem::path p(d_cachedir);
-  Regex file_regex(d_prefix + ".*\\." + bf_suffix + "\\.[[:xdigit:]]{8}$");
+  Regex file_regex(d_prefix + ".*\\." + bf_suffix + "\\..{8}$");
   for (filesystem::directory_iterator i(p); i != filesystem::directory_iterator(); ++i) {
     if (filesystem::is_regular_file(i->path()) && file_regex.match(i->path().filename().string())) {
       filesystem::remove(*i);
@@ -65,7 +63,7 @@ bool PersistentSBF::init(bool ignore_pid) {
     filesystem::path p(d_cachedir);
     try {
       if (filesystem::exists(p) && filesystem::is_directory(p)) {
-        remove_tmp_files();
+        remove_tmp_files(p, lock);
         filesystem::path newest_file;
         std::time_t newest_time=time(nullptr);
         Regex file_regex(d_prefix + ".*\\." + bf_suffix + "$");
@@ -138,7 +136,6 @@ bool PersistentSBF::snapshotCurrent(std::thread::id tid)
     std::stringstream ss;
     ss << d_prefix << "_" << tid;
     f /= ss.str() + "_" + std::to_string(getpid()) + "." + bf_suffix;
-    filesystem::path ftmp(filesystem::unique_path(f.string() + ".%%%%%%%%"));
     if (filesystem::exists(p) && filesystem::is_directory(p)) {
       try {
         std::ofstream ofile;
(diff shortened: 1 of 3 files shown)


pdns/recursordist/stable-bloom.hh
```diff
@@ -106,13 +106,28 @@ public:
...
+    }
     is.read((char*)&bitstr_len, sizeof(bitstr_len));
+    if (is.fail()) {
+      throw std::runtime_error("SBF: read failed (file too short?)");
+    }
     bitstr_len = ntohl(bitstr_len);
     char* bitcstr = new char[bitstr_len];
     is.read((char*)bitcstr, bitstr_len);
+    if (is.fail()) {
+      throw std::runtime_error("SBF: read failed (file too short?)");
```
rgacogne, line 129: There is a leak there, I'd advise the use of a smart pointer for `bitcstr`.

pdns/nod.cc
```diff
@@ -137,8 +141,13 @@ bool PersistentSBF::snapshotCurrent(std::thread::id tid)
...
 
-        if (ofile.fail())
-          throw std::runtime_error("Failed to write to file:" + f.string());
+        if (ofile.fail()) {
+          ofile.close();
+          remove(ftmp);
+          throw std::runtime_error("Failed to write to file:" + ftmp.string());
+        }
+        ofile.close();
+        rename(ftmp, f);
```
rgacogne, line 150: Perhaps we could check whether `rename` succeeded?

neilcook:
Might be worth also having some code to cleanup .tmp files on startup?

pdns/nod.cc
```diff
@@ -124,11 +126,13 @@ bool PersistentSBF::snapshotCurrent(std::thread::id tid)
     std::stringstream ss;
     ss << d_prefix << "_" << tid;
     f /= ss.str() + "_" + std::to_string(getpid()) + "." + bf_suffix;
+    path ftmp = f;
+    ftmp += ".tmp";
```
rgacogne, line 130: I wonder if we should use some variation of `mkstemp` here?

rgacogne (changes requested):

pdns/nod.cc
```diff
@@ -124,11 +138,12 @@ bool PersistentSBF::snapshotCurrent(std::thread::id tid)
     std::stringstream ss;
     ss << d_prefix << "_" << tid;
     f /= ss.str() + "_" + std::to_string(getpid()) + "." + bf_suffix;
+    path ftmp(unique_path(f.string() + ".%%%%%%%%"));
     if (exists(p) && is_directory(p)) {
       try {
         std::ofstream ofile;
         std::stringstream iss;
-        ofile.open(f.string(), std::ios::out | std::ios::binary);
+        ofile.open(ftmp.string(), std::ios::out | std::ios::binary);
```
Habbie, line 146: moving the `open` to after the `dump` shortens the window for leaving temp files around (even if we do clean them up)

pdns/nod.cc
```diff
@@ -137,8 +152,20 @@ bool PersistentSBF::snapshotCurrent(std::thread::id tid)
...
+        if (ofile.fail()) {
+          ofile.close();
+          remove(ftmp);
+          throw std::runtime_error("Failed to write to file:" + ftmp.string());
+        }
+        ofile.close();
+        try {
+          rename(ftmp, f);
+        }
+        catch (const std::runtime_error& e) {
```
Habbie, line 164: I don't think `rename` throws :)

pdns/nod.cc
```diff
@@ -137,8 +152,20 @@ bool PersistentSBF::snapshotCurrent(std::thread::id tid)
         // Now write it out to the file
         ofile << iss.str();
 
-        if (ofile.fail())
-          throw std::runtime_error("Failed to write to file:" + f.string());
+        if (ofile.fail()) {
+          ofile.close();
+          remove(ftmp);
```
Habbie, line 157: Shouldn't this be `unlink`? i have trouble even figuring what we're calling here.

pdns/nod.cc
omoerbeek: This is `boost::filesystem::rename`

pdns/nod.cc
Habbie: +1 to that, thanks

Habbie (approved):

pdns/nod.cc
```diff
@@ -34,12 +34,23 @@
...
 
 using namespace nod;
-using namespace boost::filesystem;
+namespace filesystem = boost::filesystem;
 
 // PersistentSBF Implementation 
 
 std::mutex PersistentSBF::d_cachedir_mutex;
 
+void PersistentSBF::remove_tmp_files()
```
rgacogne, line 43: Might be nice to add a comment stating that we need to hold the mutex while calling that function?

rgacogne: Or take a `std::lock_guard<std::mutex>&` parameter to actually enforce that? And perhaps a `const filesystem::path&` while we are at it?

pdns/nod.cc
```diff
@@ -119,26 +133,39 @@ void PersistentSBF::setCacheDir(const std::string& cachedir)
...
         std::ofstream ofile;
         std::stringstream iss;
-        ofile.open(f.string(), std::ios::out | std::ios::binary);
         {
           // only lock while dumping to a stringstream
           std::lock_guard<std::mutex> lock(d_sbf_mutex);
           d_sbf.dump(iss);
         }
         // Now write it out to the file
+        ofile.open(ftmp.string(), std::ios::out | std::ios::binary);
```
rgacogne, line 152: Do we care about the TOCTOU race in that case?

pdns/recursordist/stable-bloom.hh
```diff
@@ -106,15 +106,29 @@ public:
...
     num_cells = ntohl(num_cells);
     is.read((char*)&p, sizeof(p));
+    if (is.fail()) {
+      throw std::runtime_error("SBF: read failed (file too short?)");
+    }
     is.read((char*)&bitstr_len, sizeof(bitstr_len));
+    if (is.fail()) {
+      throw std::runtime_error("SBF: read failed (file too short?)");
+    }
     bitstr_len = ntohl(bitstr_len);
```
rgacogne, line 125: I wonder if it would make sense to do some sanity check on that value before allocating the buffer?

pdns/recursordist/stable-bloom.hh
```diff
@@ -106,15 +106,29 @@ public:
...
-    char* bitcstr = new char[bitstr_len];
-    is.read((char*)bitcstr, bitstr_len);
-    std::string bitstr(bitcstr, bitstr_len);
-    delete[] bitcstr;
+    unique_ptr<char[]> bitcstr = make_unique<char[]>(bitstr_len);
+    is.read(bitcstr.get(), bitstr_len);
+    if (is.fail()) {
+      throw std::runtime_error("SBF: read failed (file too short?)");
+    }
+    std::string bitstr(bitcstr.get(), bitstr_len);
```
rgacogne, line 131: Do we care about performance? If we do we could allocate and read into the string instead of using a temporary buffer.

pdns/nod.cc
Habbie: What is the race?

pdns/recursordist/stable-bloom.hh
Habbie: It's only at startup, it does take about a second per file on my system.

pdns/nod.cc
rgacogne: We generate a name, do some stuff and then open a file for writing with that name. It looks like it could have been created as a symlink to another file in the meantime, although I agree it's likely extremely difficult, depending on how truly random the result is.

Habbie: Right! The dirty secret is that there is no TOC here. `unique_path` just operates on a string.

Habbie: (So depending on how you look at it, there is no window, or the window is infinite already.)

rgacogne: Doesn't `boost::filesystem::unique_path` check whether the file exists? The documentation is a bit terse.

Habbie: It doesn't - here is the code, it's short enough to read in a minute https://github.com/boostorg/filesystem/blob/6b5e38134a336b6ea777cd8a69b6ae929819db7e/src/unique_path.cpp#L234-L270

omoerbeek: Ugh in my mind I thought it would be a safe function, But it cannot be if it does not create the file itself.

Habbie: more complaints here - https://stackoverflow.com/questions/43946120/how-does-boostfilesystemunique-path-resolve-the-need-in-mkstemp-analogue-i

suggests `O_EXCL | O_CREAT` on the `open`, but perhaps we want to do real `mkstemp` instead?

rgacogne: It looks like even C++17 doesn't offer anything better than `mkstemp`..

omoerbeek: There is no interface to ofstream to request `O_EXCL | O_CREAT` afaiks, so i'll go for `mkstemp`

pdns/recursordist/stable-bloom.hh
omoerbeek: I left this for now, if/when we change the on-disk format, we can make it fast.

rgacogne (approved):
One nit, looks good otherwise :+1:

PR #9552, opened 2020-09-29
auth: remove local-ipv6, query-local-address6

zeha:
### Short description
#5368 for auth.

### Checklist

I have:
- [x] read the [CONTRIBUTING.md](https://github.com/PowerDNS/pdns/blob/master/CONTRIBUTING.md) document
- [x] compiled this code
- [ ] tested this code
- [ ] included documentation (including possible behaviour changes)
- [ ] documented the code
- [ ] added or modified regression test(s)
- [ ] added or modified unit test(s)

PR #10251, opened 2021-03-31
auth: remove local-ipv6, query-local-address6 (redo of #9552). Closes #5368

### LINKED ISSUE #5368: Consistency IPv6 configure options
 - Issue type: Feature request 
### Short description
Consistency in IPv6 options

### Usecase
Have consistency in configs between Auth and Rec.

### Description
Authoritative has this option:
```
  --local-ipv6=...
	Local IP address to which we bind
```

where as Recursor accepts IPv6 in `--local-address`.

I think it would be clearer and more logical if both applications use the same options for the same features. 

```
12:24 < Evo--> just random interest question. Why does the authoritative server have the option 
               local-ipv6, but the recursor has ipv6 in local-address? 
12:24 < ahu> because we've been working on this stuff for 18 years now
12:24 < ahu> and we are sadly not always consistent
12:25 < Evo--> fair enough
12:26 < Evo--> I was just wondering :) 
12:27 < lieter> we could try to remove local-ipv6 in auth 4.1
12:27 < ahu> consistency between auth and recursor has never been an explicit goal
12:27 < lieter> sure, but is could point to some code duplication though :)
```

Habbie:
### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -775,15 +775,8 @@ Fail to start if one or more of the
 
 ``local-ipv6``
 --------------
-.. deprecated:: 4.3.0
-  This setting has been deprecated, use :ref:`setting-local-address`
-
--  IPv6 Addresses, separated by commas or whitespace
--  Default: ``::``
-
-Local IPv6 address to which we bind. It is highly advised to bind to
-specific interfaces and not use the default 'bind to any'. This causes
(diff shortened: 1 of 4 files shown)


pieterlexis (approved):
lgtm

PR #10255, opened 2021-04-01
auth postgres: escape identifiers

Habbie:
### Short description
In #3477 we stopped escaping identifiers on all distributions (not just on CentOS 6) because PG_VERSION_NUM was not set by the include we use. Now that we dropped support for CentOS 6, we can get rid of the #ifdef altogether.

(spotted by @mind04, thanks)

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/modules/gpgsqlbackend/spgsql.cc
+++ b/modules/gpgsqlbackend/spgsql.cc
@@ -126,16 +126,11 @@ public:
     // by stored procedures. you can return more than one
     // if you return SETOF refcursor.
     if (PQftype(d_res_set, 0) == 1790) { // REFCURSOR
-#if PG_VERSION_NUM > 90000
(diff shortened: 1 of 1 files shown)


zeha:
The docs for `PQescapeIdentifier` indicate that this code should not actually work ...:

> The return string will also be surrounded by double quotes.

But we add more double quotes around it!?

Habbie:
I was going to say 'but the tests pass' but we don't test this code!

Habbie:
I wrote a test. This code hasn't worked since 80760247ad05f914ad4d3a62f3dd06b39d815c99.

Habbie:
I.e. since 4.0.

Habbie:
Replaced by #10259

PR #10121, opened 2021-02-25
Implement RFC 6742 RR Types

pieterlexis:
### Short description
This commit implements the NID, L32, L64, and LP record types from RFC
6742. `dig` can properly parse the wiredata we send, and with the added
tests, I'm certain of the implementation.

The RFC has optional additional processing, this is not implemented.

Closes #10119

### Checklist

I have:
- [x] read the [CONTRIBUTING.md](https://github.com/PowerDNS/pdns/blob/master/CONTRIBUTING.md) document
- [x] compiled this code
- [x] tested this code
- [ ] included documentation (including possible behaviour changes)
- [ ] documented the code
- [ ] added or modified regression test(s)
- [x] added or modified unit test(s)

PR #10256, opened 2021-04-01
RFC6742 types: use correct size

Habbie:
### Short description
The implementation in #10121 threw compiler warnings. After investigation, I found that that code only works by accident, because pointers are 64 bits on the systems we compile for.

In this PR, the first and third(!) commits are two alternative fixes for this problem. Please review and let me know what you like.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/misc.hh
+++ b/pdns/misc.hh
@@ -630,4 +630,4 @@ size_t parseSVCBValueList(const std::string &in, vector<std::string> &val);
 std::string makeLuaString(const std::string& in);
 
 // Used in NID and L64 records
-typedef uint8_t NodeOrLocatorID[8];
+struct NodeOrLocatorID { uint8_t content[8]; };
--- a/pdns/dnsparser.cc
+++ b/pdns/dnsparser.cc
@@ -386,7 +386,7 @@ void PacketReader::xfrNodeOrLocatorID(NodeOrLocatorID& ret)
   if (d_pos + sizeof(ret) > d_content.size()) {
     throw std::out_of_range("Attempt to read 64 bit value outside of packet");
   }
-  memcpy(&ret, &d_content.at(d_pos), sizeof(ret));
+  memcpy(&ret.content, &d_content.at(d_pos), sizeof(ret.content));
   d_pos += sizeof(ret);
 }
 
--- a/pdns/dnswriter.cc
+++ b/pdns/dnswriter.cc
@@ -140,7 +140,7 @@ template <typename Container> void GenericDNSPacketWriter<Container>::xfr48BitIn
   d_content.insert(d_content.end(), bytes, bytes + sizeof(bytes));
 }
 
-template <typename Container> void GenericDNSPacketWriter<Container>::xfrNodeOrLocatorID(NodeOrLocatorID val)
+template <typename Container> void GenericDNSPacketWriter<Container>::xfrNodeOrLocatorID(const NodeOrLocatorID& val)
 {
   d_content.insert(d_content.end(), val.content, val.content + sizeof(val.content));
 }
--- a/pdns/dnswriter.cc
+++ b/pdns/dnswriter.cc
@@ -142,7 +142,7 @@ template <typename Container> void GenericDNSPacketWriter<Container>::xfr48BitIn
 
 template <typename Container> void GenericDNSPacketWriter<Container>::xfrNodeOrLocatorID(NodeOrLocatorID val)
 {
-  d_content.insert(d_content.end(), val, val + sizeof(val));
+  d_content.insert(d_content.end(), val.content, val.content + sizeof(val.content));
 }
 
 template <typename Container> void GenericDNSPacketWriter<Container>::xfr32BitInt(uint32_t val)
--- a/pdns/dnswriter.hh
+++ b/pdns/dnswriter.hh
@@ -86,7 +86,7 @@ public:
   void truncate();
 
   void xfr48BitInt(uint64_t val);
-  void xfrNodeOrLocatorID(NodeOrLocatorID val);
+  void xfrNodeOrLocatorID(const NodeOrLocatorID& val);
   void xfr32BitInt(uint32_t val);
   void xfr16BitInt(uint16_t val);
   void xfrType(uint16_t val)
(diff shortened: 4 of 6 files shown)


omoerbeek (approved):

Habbie:
> In this PR, the first and third(!) commits are two alternative fixes for this problem. Please review and let me know what you like.

Now reduced to a single commit with the final version.

rgacogne:
I'm not sure I understand the initial issue, since I don't see `sizeof` being applied to a pointer, and unless I'm mistaken it is supposed to do the right thing with references: `When applied to a reference type, the result is the size of the referenced type`?

Habbie:
```
dnswriter.cc: In instantiation of ‘void GenericDNSPacketWriter<Container>::xfrNodeOrLocatorID(uint8_t*) [with Container = std::vector<unsigned char>; uint8_t = unsigned char]’:
dnswriter.cc:514:16:   required from here
dnswriter.cc:145:48: warning: ‘sizeof’ on array function parameter ‘val’ will return size of ‘uint8_t*’ {aka ‘unsigned char*’} [-Wsizeof-array-argument]
   d_content.insert(d_content.end(), val, val + sizeof(val));
                                                ^~~~~~~~~~~
In file included from dnswriter.cc:29:
dnswriter.hh:89:43: note: declared here
   void xfrNodeOrLocatorID(NodeOrLocatorID val);
                           ~~~~~~~~~~~~~~~~^~~
dnswriter.cc: In instantiation of ‘void GenericDNSPacketWriter<Container>::xfrNodeOrLocatorID(uint8_t*) [with Container = std::vector<unsigned char, noinit_adaptor<std::allocator<unsigned char> > >; uint8_t = unsigned char]’:
dnswriter.cc:516:16:   required from here
dnswriter.cc:145:48: warning: ‘sizeof’ on array function parameter ‘val’ will return size of ‘uint8_t*’ {aka ‘unsigned char*’} [-Wsizeof-array-argument]
   d_content.insert(d_content.end(), val, val + sizeof(val));
                                                ^~~~~~~~~~~
dnswriter.cc:143:106: note: declared here
 template <typename Container> void GenericDNSPacketWriter<Container>::xfrNodeOrLocatorID(NodeOrLocatorID val)
                                                                                          ~~~~~~~~~~~~~~~~^~~
```

Indeed, the code that got a reference does not throw a warning.

Habbie:
So 2/3rds of the PR is unnecessary.

rgacogne:
Right, I missed the fact that `NodeOrLocatorID` is not an object but a `uint8_t[8]`..  Perhaps we should change that?

Habbie:
Before I squashed everything today, I had 3 variants sitting in this PR. One was:
```
-typedef uint8_t NodeOrLocatorID[8];
+typedef struct NodeOrLocatorID { uint8_t content[8]; } NodeOrLocatorID;
```

(plus related changes), but Otto and I thought fixing the sizeof calls would be simpler.

Habbie:
Here's another alternative for this whole PR:
```
diff --git a/pdns/dnswriter.cc b/pdns/dnswriter.cc
index 775793dcd..4842fd10e 100644
--- a/pdns/dnswriter.cc
+++ b/pdns/dnswriter.cc
@@ -140,7 +140,7 @@ template <typename Container> void GenericDNSPacketWriter<Container>::xfr48BitIn
   d_content.insert(d_content.end(), bytes, bytes + sizeof(bytes));
 }
 
-template <typename Container> void GenericDNSPacketWriter<Container>::xfrNodeOrLocatorID(NodeOrLocatorID val)
+template <typename Container> void GenericDNSPacketWriter<Container>::xfrNodeOrLocatorID(NodeOrLocatorID& val)
 {
   d_content.insert(d_content.end(), val, val + sizeof(val));
 }
diff --git a/pdns/dnswriter.hh b/pdns/dnswriter.hh
index 96e8d4f9a..f968856dc 100644
--- a/pdns/dnswriter.hh
+++ b/pdns/dnswriter.hh
@@ -86,7 +86,7 @@ public:
   void truncate();
 
   void xfr48BitInt(uint64_t val);
-  void xfrNodeOrLocatorID(NodeOrLocatorID val);
+  void xfrNodeOrLocatorID(NodeOrLocatorID& val);
   void xfr32BitInt(uint32_t val);
   void xfr16BitInt(uint16_t val);
   void xfrType(uint16_t val)
```

rgacogne:
My feeling is that it would be better to avoid hiding an array behind a `typedef` for that exact reason, pass-by-value not doing what you would expect and `sizeof` giving unexpected results. The current PR would solve the issue, but only until someone passes a `NodeOrLocatorID` by value again (or, well, expect so).

Habbie:
> My feeling is that it would be better to avoid hiding an array behind a `typedef` for that exact reason, pass-by-value not doing what you would expect and `sizeof` giving unexpected results. The current PR would solve the issue, but only until someone passes a `NodeOrLocatorID` by value again (or, well, expect so).

This does make sense to me. How do you like the struct version I pasted earlier?

rgacogne:
The struct looks good, except I would get rid of the useless `typedef` :)

Habbie:
@omoerbeek opinions?

omoerbeek:
Agree with @rgacogne, it is better to make sure another small mistake will not lead to trouble.

Habbie:
Ok, updated again! I'm unsure about `val` vs. `val.content`, I think they'd be equivalent, but this felt clearer to me.

pdns/misc.hh
```diff
@@ -630,4 +630,4 @@ size_t parseSVCBValueList(const std::string &in, vector<std::string> &val);
 std::string makeLuaString(const std::string& in);
 
 // Used in NID and L64 records
-typedef uint8_t NodeOrLocatorID[8];
+struct NodeOrLocatorID { uint8_t content[8]; }; // NodeOrLocatorID;
```
omoerbeek, line 633:
```suggestion
struct NodeOrLocatorID { uint8_t content[8]; };
```

omoerbeek (approved):
one nit

omoerbeek:
Formatting failure is fixed on master

Habbie:
squashed

Habbie:
I polished the yak some more. I promise I will stop soon. :)

rgacogne (approved):
:+1:

PR #2743, opened 2015-09-12
Support multiple results set in gMySQL and gPGSql

cmouse:
Add result set support for gmysql and gpgsql. Allows use of stored procedures in these backends. Goracle would require quite a lot of effort to do this (at least I could not find any sensible way forward).

PR #10259, opened 2021-04-01
auth gpgsql: add stored procedure test, drop refcursor support

Habbie:
### Short description
#2743 introduced support for 'multiple result sets' from Postgres (and MySQL). For Postgres, this needed the wrapping of even simple `SELECT` queries in transactions so that the 'refcursor' result sets would not get freed before we could get to them. That PR did not contain tests for the functionality.

(For reference, that would make the test in -this- PR work if the function looked more like:
```
CREATE FUNCTION basic_query(incoming_type varchar(10), incoming_name varchar(255))
RETURNS REFCURSOR AS $$
DECLARE
  ref refcursor;
BEGIN
  OPEN ref FOR SELECT content,ttl,prio,type,domain_id,disabled,name,auth FROM records WHERE disabled=false and type=incoming_type and name=incoming_name;
  RETURN ref;
END;
$$ LANGUAGE plpgsql;
```
)

Later, #3477 added an `#ifdef` because the code in #2743 relied on Postgres 9 functionality, while CentOS 6 was on a lower version.

Later, #5233 removed the transaction around all those `SELECT` queries because it was not obvious why they were there, and they added a bunch of roundtrips to each backend query, hurting performance.

Recently, @mind04 pointed out to me that the `#ifdef` in #3477 did nothing because `PG_VERSION_NUM` was not actually defined.

Today, I filed #10255 to undo #3477, because we no longer support CentOS 6. Then @zeha pointed out 'this code cannot work' (see reasons (quoting) in #10255). I added a test (part of this PR right here) and noticed that REFCURSOR support was broken for reasons other than quoting.

So, it turns out REFCURSOR support has been broken since Auth 4.0, and as far as I can tell, nobody complained.

This PR:
* adds a test (which can also serve as an example) for doing simple (non-refcursor) stored procedures with Postgres
* removes the broken REFCURSOR support
* serves as a reference to be used in an error message stating we do not support REFCURSORs

Dear user, if you made it through all that, and feel that we've done you wrong by not supporting REFCURSORs, please speak up so that we can figure out how to serve your usecase best.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/modules/gpgsqlbackend/spgsql.cc
+++ b/modules/gpgsqlbackend/spgsql.cc
@@ -116,33 +116,16 @@ public:
   void nextResult()
   {
     if (d_res_set == nullptr)
-      return; // no refcursor
+      return;
     if (d_cur_set >= PQntuples(d_res_set)) {
       PQclear(d_res_set);
       d_res_set = nullptr;
       return;
     }
-    // this code handles refcursors if they are returned
-    // by stored procedures. you can return more than one
-    // if you return SETOF refcursor.
     if (PQftype(d_res_set, 0) == 1790) { // REFCURSOR
-#if PG_VERSION_NUM > 90000
-      // PQescapeIdentifier was added to libpq in postgresql 9.0
-      char* val = PQgetvalue(d_res_set, d_cur_set++, 0);
-      char* portal = PQescapeIdentifier(d_db(), val, strlen(val));
-      string cmd = string("FETCH ALL FROM \"") + string(portal) + string("\"");
-      PQfreemem(portal);
-#else
-      string portal = string(PQgetvalue(d_res_set, d_cur_set++, 0));
-      string cmd = string("FETCH ALL FROM \"") + portal + string("\"");
-#endif
-      // execute FETCH
-      if (d_dolog) {
-        g_log << Logger::Warning << "Query: " << cmd << endl;
-      }
-      d_res = PQexec(d_db(), cmd.c_str());
-      d_resnum = PQntuples(d_res);
-      d_residx = 0;
+      g_log << Logger::Error << "Postgres query returned a REFCURSOR and we do not support those - see https://github.com/PowerDNS/pdns/pull/10259" << endl;
+      PQclear(d_res_set);
+      d_res_set = nullptr;
     }
     else {
       d_res = d_res_set;
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -881,6 +881,8 @@ jobs:
           context: gpgsql-nsec3-optout-both
       - auth-regress:
           context: gpgsql-nsec3-narrow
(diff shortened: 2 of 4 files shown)


modules/gpgsqlbackend/spgsql.cc
```diff
@@ -126,23 +126,7 @@ class SPgSQLStatement : public SSqlStatement
     // by stored procedures. you can return more than one
     // if you return SETOF refcursor.
```
zeha, line 127: I guess the comments here should go away

modules/gpgsqlbackend/spgsql.cc
```diff
@@ -126,23 +126,7 @@ class SPgSQLStatement : public SSqlStatement
...
-      string cmd = string("FETCH ALL FROM \"") + portal + string("\"");
-#endif
-      // execute FETCH
-      if (d_dolog) {
-        g_log << Logger::Warning << "Query: " << cmd << endl;
-      }
-      d_res = PQexec(d_db(), cmd.c_str());
-      d_resnum = PQntuples(d_res);
-      d_residx = 0;
+      g_log << Logger::Error << "Postgres query returned a REFCURSOR and we do not support those" << endl;
```
zeha, line 129: might need cleanup of d_res_set?

zeha (approved):
looks correct to me!

modules/gpgsqlbackend/spgsql.cc
```diff
@@ -116,33 +116,16 @@ class SPgSQLStatement : public SSqlStatement
...
-      string cmd = string("FETCH ALL FROM \"") + portal + string("\"");
-#endif
-      // execute FETCH
-      if (d_dolog) {
-        g_log << Logger::Warning << "Query: " << cmd << endl;
-      }
-      d_res = PQexec(d_db(), cmd.c_str());
-      d_resnum = PQntuples(d_res);
-      d_residx = 0;
+      g_log << Logger::Error << "Postgres query returned a REFCURSOR and we do not support those - see https://github.com/PowerDNS/pdns/pull/10259" << endl;
```
jsoref, line 126: If it's safe to say that pdns has never supported them (and I think if I'm reading the PR description this is correct), then you might want to say that here:
```suggestion
      g_log << Logger::Error << "Postgres query returned a REFCURSOR and we have never supported them - see https://github.com/PowerDNS/pdns/pull/10259" << endl;
```

I'm not certain on either side (neither my conclusion, nor whether changing the text to be explicit is useful).

Habbie: I haven't figured out if they worked before 4.0, and I'm not going to - so I think I'll stick with the current text which also does not know that. Thanks!

PR #10260, opened 2021-04-01
auth: lower max-nsec3-iterations to 100

mind04:
### Short description
A high number of iterations is a waste of resources and is not future proof .

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -1012,10 +1012,13 @@ means unlimited.
 ------------------------
 
 -  Integer
--  Default: 500
+-  Default: 100
 
 Limit the number of NSEC3 hash iterations for zone configurations.
 For more information see :ref:`dnssec-operational-nsec-modes-params`.
+
+.. note::
+  Pre 4.5.0 the default was 500.
 
 .. _setting-max-packet-cache-entries:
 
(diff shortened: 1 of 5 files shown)


zeha:
https://tools.ietf.org/html/draft-hardaker-dnsop-nsec3-guidance-02 appears to suggest 0?

mind04:
This is based on https://tools.ietf.org/html/draft-hardaker-dnsop-nsec3-guidance-02#page-5
```
Validating resolvers SHOULD  return a SERVFAIL when processing NSEC3 records with iterations  larger than 100.
```

omoerbeek:
Isn't it a bit too overzealous to already lower the limit when the proposal is still in draft state and without any idea on how many domains use iterations > 100?

Habbie:
It might be a bit early to merge this, yes. The PR does also contain a bug fix for presigned zones :)

omoerbeek:
As a side note, it would be nice if PRs like this (i.e. changing defaults or adding new settings) would include a upgrade guide snippet.

Habbie (approved):

PR #10268, opened 2021-04-06
rec: Log local IP in dnstap messages

### LINKED ISSUE #10265: Set the `query_address` field of Dnstap messages
- Program:  Recursor
 - Issue type: Feature request

### Short description

Set the `query_address` field of Dnstap log messages.

### Usecase
When receiving Dnstap log messages from more than one recursor, you can identify them by their IP addresses. If you have a direct TCP connection between recursor and log receiver, the recursor IP address is available as source address of the connection.

In our setup we have an intermediary system that we do not control that relays those message to us. Due to that setup, we cannot simply take the recursor IP addresses from the TCP connections since we just see the one from the intermediary system. That makes us unable to identify the different recursors since the Dnstap message itself does not contain such information and we rely on the IP address for identification.

### Description

The `query_address` field of a Dnstap log message identifies the questioning instance in form of its IP address and should be set in order to provide that information to the log receiver.

omoerbeek:
### Short description

Fixes #10265.

Also fix two cases where we should not send a dnstap message:

1) if we're chaining on an existing query; 
2) if the sending failed

Needs tests, therefore draft.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/lwres.cc
+++ b/pdns/lwres.cc
@@ -72,8 +72,7 @@ static bool isEnabledForQueries(const std::shared_ptr<std::vector<std::unique_pt
   return false;
 }
 
-static void logFstreamQuery(const std::shared_ptr<std::vector<std::unique_ptr<FrameStreamLogger>>>& fstreamLoggers, const struct timeval &queryTime, const ComboAddress& ip, bool doTCP,
-  boost::optional<const DNSName&> auth, const vector<uint8_t>& packet)
+static void logFstreamQuery(const std::shared_ptr<std::vector<std::unique_ptr<FrameStreamLogger>>>& fstreamLoggers, const struct timeval &queryTime, const ComboAddress& localip, const ComboAddress& ip, bool doTCP, boost::optional<const DNSName&> auth, const vector<uint8_t>& packet)
 {
   if (fstreamLoggers == nullptr)
     return;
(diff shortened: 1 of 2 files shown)


pdns/lwres.cc
```diff
@@ -389,7 +403,7 @@ LWResult::Result asyncresolve(const ComboAddress& ip, const DNSName& domain, int
 
 #ifdef HAVE_FSTRM
   if (isEnabledForResponses(fstrmLoggers)) {
-    logFstreamResponse(fstrmLoggers, ip, doTCP, context ? context->d_auth : boost::none, buf, queryTime, *now);
+    logFstreamResponse(fstrmLoggers, ip, doTCP, context ? context->d_auth : boost::none, buf, queryTime, *now, clientip);
```
rgacogne, line 406: For the UDP case it looks like we might not set the client IP if we don't log queries?

pdns/lwres.cc
```diff
@@ -308,16 +310,23 @@ LWResult::Result asyncresolve(const ComboAddress& ip, const DNSName& domain, int
     }
 
     ret = asendto((const char*)&*vpacket.begin(), vpacket.size(), 0, ip, qid,
-                  domain, type, &queryfd);
+                  domain, type, &queryfd, fstrmQEnabled ? &clientip : nullptr);
```
rgacogne, line 314: I wonder if it would not make more sense to call `getsockname()` on the returned `queryfd` from here instead of doing it in `asendto`. Or if we don't care about altering more code for that feature, perhaps skip that call completely by going all the way down to `makeClientSocket()`, but that does not seem very clean.

pdns/lwres.cc
omoerbeek: Yes, in that case we would log an all-zero clientip... Will look into that.

pdns/lwres.cc
omoerbeek: We're looking at remains of an earlier attempt to set the address in `makeClientSocket()`. I then decided to push the `getsockname()` call upward for cleanliness, but I see now that it can be pushed even higher.

rgacogne (approved):
Looks good, perhaps we could add a check in the regression tests?

omoerbeek:
on it!

PR #10271, opened 2021-04-07
fix read alignment in dnspcap

Habbie:
### Short description
This avoids:
```
dnspcap.cc:146:98: runtime error: member access within misaligned address 0x7ffc47a138d2 for type 'struct ip', which requires 4 byte alignment
0x7ffc47a138d2: note: pointer points here
 00 00  08 00 45 00 00 54 06 83  00 00 40 11 76 14 7f 00  00 01 7f 00 00 01 89 f3  14 b4 00 40 fe 53
              ^ 
    #0 0x564956a3f534 in PcapPacketReader::getUDPPacket() /home/peter/projects/powerdns/pdns/pdns/dnspcap.cc:146
    #1 0x564956c4ed99 in main /home/peter/projects/powerdns/pdns/pdns/dnsscope.cc:230
    #2 0x7f2f7c6bb09a in __libc_start_main ../csu/libc-start.c:308
    #3 0x5649569182d9 in _start (/home/peter/projects/powerdns/pdns/pdns/dnsscope+0xd3f2d9)

```

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnspcap.cc
+++ b/pdns/dnspcap.cc
@@ -59,6 +59,13 @@ PcapPacketReader::PcapPacketReader(const string& fname) : d_fname(fname)
   else throw runtime_error((boost::format("Unsupported link type %d") % d_pfh.linktype).str());
 
   d_runts = d_oversized = d_correctpackets = d_nonetheripudp = 0;
+
+  size_t alignmentCorrection = d_skipMediaHeader % alignof(struct ip);
+
+  d_buffer = d_readbuffer + alignmentCorrection;
+  d_bufsize = sizeof(d_readbuffer) - alignmentCorrection;
+
+  if (d_skipMediaHeader > d_bufsize) throw runtime_error("media header is too big");
 }
 
 void PcapPacketReader::checkedFreadSize(void* ptr, size_t size) 
@@ -85,9 +92,9 @@ try
       continue;
     }
 
-    if(d_pheader.caplen > sizeof(d_buffer)) {
+    if(d_pheader.caplen > d_bufsize) {
       d_oversized++;
-      throw runtime_error((boost::format("Can't handle a %d byte packet, have space for %d")  % d_pheader.caplen % sizeof(d_buffer)).str());
+      throw runtime_error((boost::format("Can't handle a %d byte packet, have space for %d")  % d_pheader.caplen % d_bufsize).str());
     }
 
     checkedFreadSize(d_buffer, d_pheader.caplen);
--- a/pdns/dnspcap.cc
+++ b/pdns/dnspcap.cc
@@ -94,7 +94,7 @@ try
 
     if(d_pheader.caplen > d_bufsize) {
       d_oversized++;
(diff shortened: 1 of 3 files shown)


pdns/dnspcap.hh
```diff
@@ -115,7 +115,9 @@ public:
 
   pdns_pcap_file_header d_pfh;
   unsigned int d_runts, d_oversized, d_correctpackets, d_nonetheripudp;
-  char d_buffer[32768];
+  alignas (struct ip) char d_readbuffer[32768];
```
omoerbeek, line 118: To compensate for the part being chopped by the `d_bufsize` calculation maybe this?
```suggestion
  alignas (struct ip) char d_readbuffer[32768 + alignof(struct ip)];
```
or is it already oversized?

pdns/dnspcap.cc
```diff
@@ -85,9 +92,9 @@ try
       continue;
     }
 
-    if(d_pheader.caplen > sizeof(d_buffer)) {
+    if(d_pheader.caplen > d_bufsize) {
       d_oversized++;
-      throw runtime_error((boost::format("Can't handle a %d byte packet, have space for %d")  % d_pheader.caplen % sizeof(d_buffer)).str());
+      throw runtime_error((boost::format("Can't handle a %d byte packet, have space for %d")  % d_pheader.caplen % d_bufsize).str());
```
omoerbeek, line 97: Wrong format specifier (unless boost is very smart)
```suggestion
      throw runtime_error((boost::format("Can't handle a %d byte packet, have space for %zu")  % d_pheader.caplen % d_bufsize).str());
```

omoerbeek (approved):
This reads good, with two nits; though it is tricky stuff. I'm assuming the various sanitizers are happy...

pdns/dnspcap.hh
Habbie: It is already way oversized, and I do reduce `d_bufsize` to avoid overruns, so I think we're good..

rgacogne:
This PR aligns the buffer for `struct ip`, but it looks like we also need to be aligned with `struct pdns_lcc_header`, `struct ether_header`, `struct ip6_hdr`, `struct tcphdr` and `struct udphdr`. Is that change enough to fix that as well (perhaps because they are similarly aligned)?

Habbie:
> `struct pdns_lcc_header`

is 16 bytes (a multiple of 4), sitting before `ip_hdr` or `ip6_hdr`, so automatically aligned

> `struct ether_header`

is 14 (not a multiple of 4), and is the whole reason for this PR, which means the Sanitizer actually checked my work on this one. Now that I look at it, I wonder why that one is not a problem. I suspect it's fine because we only read a 2 byte value from it, right before `ip_hdr`, which means that 2 byte value (in a packed struct!) is aligned at 2, which should be fine? 

> `struct ip6_hdr`

is aligned identically with ip_hdr

> `struct tcphdr`
> `struct udphdr`

These sit after the ip header (length always a multiple of 4) or ipv6 header (length 40, a multiple of 4).

rgacogne:
Thanks, that PR looks good to me then!

PR #10274, opened 2021-04-07
dnsdist: Fix some issues reported by Thread Sanitizer

rgacogne:
### Short description

None are important issues but should be fixed anyway:
- a possible race when reloading a DoH or DoT certificate, very unlikely on x86_64 but could be an issue on other platforms under huge load ;
- a race in QPSAction and QPSPoolAction could lead to the QPS rate being under-evaluated ;
- a race in TeeAction could lead to under-evaluated metrics.

It would be nice to add a TSAN-enabled build to our CI, using the included suppressions list and passing something like: `TSAN_OPTIONS=halt_on_error=1:suppressions=/path/to/dnsdist-tsan.supp`

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code


Habbie:
> It would be nice to add a TSAN-enabled build to our CI, using the included suppressions list and passing something like: `TSAN_OPTIONS=halt_on_error=1:suppressions=/path/to/dnsdist-tsan.supp`

I have not had a look at all - can we combine this with ASAN/UBSAN or would it need to be separate?

rgacogne:
> I have not had a look at all - can we combine this with ASAN/UBSAN or would it need to be separate?

Unfortunately TSAN can't be enabled along with ASAN or UBSAN. I pushed a first attempt at doing a TSAN-enabled dnsdist build and regression tests in CircleCI, but advices are welcome :)

.circleci/config.yml
```diff
@@ -1908,6 +1908,91 @@ jobs:
...
+            CXXFLAGS="-O1 -Werror=vla -Werror=shadow -Wformat=2 -Werror=format-security -Werror=string-plus-int -Wp,-D_GLIBCXX_ASSERTIONS" \
+            ./configure \
+            --enable-option-checking=fatal \
+            --enable-unit-tests \
+            --enable-dnstap \
+            --enable-dnscrypt \
+            --enable-dns-over-tls \
+            --enable-dns-over-https \
+            --enable-systemd \
+            --prefix=/opt/dnsdist \
```
Habbie, line 1970:
```suggestion
            --prefix=/opt/dnsdist-with-tsan \
```

.circleci/config.yml
```diff
@@ -1939,6 +2024,36 @@ jobs:
...
+            /etc/init.d/snmpd start
+      - run:
+          name: install prometheus tools
+          command: |
+            apt-get -qq --no-install-recommends install prometheus
+      - run:
+          name: Run dnsdist tests
+          workdir: ~/project/regression-tests.dnsdist
+          command: |
+            DNSDISTBIN="/opt/dnsdist/bin/dnsdist-with-tsan" \
```
Habbie, line 2054:
```suggestion
            DNSDISTBIN="/opt/dnsdist-with-tsan/bin/dnsdist" \
```

Habbie (changes requested):
two diffs suggested - unless I'm totally misunderstanding something!

.circleci/config.yml
```diff
@@ -1908,6 +1908,91 @@ jobs:
...
+          command: make check || (cat test-suite.log; false)
+          working_directory: ~/project/pdns/dnsdistdist
+      - run:
+          name: Install resulting binaries
+          command: make install
+          working_directory: ~/project/pdns/dnsdistdist
+      - persist_to_workspace:
+          root: /opt
+          paths:
+            - dnsdist
```
Habbie, line 1994: I believe this would overwrite the non-tsan build (or vice versa, i.e., this TSAN PR introduces a race condition ;) )

rgacogne: Ah, I was wondering about that! It's not very clear to me how `persist_to_workspace` works :-/

Habbie: The workspace is per workflow and both the tsan and non-tsan builds are in the `build-and-test-all` workflow. You could make a separate workflow (but I don't really see benefits to that), or take my previous suggestions.

Habbie: Ah, you already did that!

rgacogne: I think I'm starting to see it, thanks! It looks like I also need to change the `persist_to_workplace` path to `dnsdist-with-tsan`, right?

Habbie: yes!

Habbie (approved):
circleci changes look good now!

rgacogne:
I squashed some commits to clean up the history, will merge as soon as the tests are green again.

PR #10288, opened 2021-04-13
rec: Stop using potentially offensive names internally and warn about deprecated settings.

omoerbeek:
### Short description

ATM table s used by both rec and auth, that seems a bit wrong, but otoh simplifies matters...

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/pdns_recursor.cc
+++ b/pdns/pdns_recursor.cc
@@ -1308,7 +1308,7 @@ static bool checkFrameStreamExport(LocalStateHolder<LuaConfigItems>& luaconfsLoc
 static bool nodCheckNewDomain(const DNSName& dname)
 {
   bool ret = false;
-  // First check the (sub)domain isn't whitelisted for NOD purposes
+  // First check the (sub)domain isn't ignored for NOD purposes
   if (!g_nodDomainWL.check(dname)) {
     // Now check the NODDB (note this is probabilistic so can have FNs/FPs)
     if (t_nodDBp && t_nodDBp->isNewDomain(dname)) {
@@ -4499,7 +4499,7 @@ static void setupNODThread()
   }
 }
 
-static void parseNODWhitelist(const std::string& wlist)
+static void parseNODIgnorelist(const std::string& wlist)
 {
   vector<string> parts;
   stringtok(parts, wlist, ",; ");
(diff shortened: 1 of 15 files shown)


pdns/pdns_recursor.cc
```diff
@@ -4499,7 +4499,7 @@ static void setupNODThread()
   }
 }
 
-static void parseNODWhitelist(const std::string& wlist)
+static void parseNODIgnorelist(const std::string& wlist)
```
pieterlexis, line 4502: a whitelist and an ignore list seem to be eachother's opposites?

pdns/pdns_recursor.cc
```diff
@@ -5572,22 +5572,22 @@ int main(int argc, char **argv)
     ::arg().set("snmp-master-socket", "If set and snmp-agent is set, the socket to use to register to the SNMP daemon (deprecated)")="";
     ::arg().set("snmp-daemon-socket", "If set and snmp-agent is set, the socket to use to register to the SNMP daemon")="";
 
-    std::string defaultBlacklistedStats = "cache-bytes, packetcache-bytes, special-memory-usage";
+    std::string defeaultDisabledStats = "cache-bytes, packetcache-bytes, special-memory-usage";
```
pieterlexis, line 5575: s/defeault/default/

pdns/pdns_recursor.cc
omoerbeek: Original code used "whitelisted" to not report as newly observed. So this is consitent.

pdns/pdns_recursor.cc
```diff
@@ -5057,15 +5057,15 @@ static int serviceMain(int argc, char*argv[])
 
   g_useKernelTimestamp = ::arg().mustDo("protobuf-use-kernel-timestamp");
 
-  blacklistStats(StatComponent::API, ::arg()["stats-api-blacklist"]);
-  blacklistStats(StatComponent::Carbon, ::arg()["stats-carbon-blacklist"]);
-  blacklistStats(StatComponent::RecControl, ::arg()["stats-rec-control-blacklist"]);
-  blacklistStats(StatComponent::SNMP, ::arg()["stats-snmp-blacklist"]);
+  disableStats(StatComponent::API, ::arg()["stats-api-blacklist"]);
```
pieterlexis, line 5060: this still uses the old names of the configs. Perhaps we should merge their values into the new one?

omoerbeek: `disableStats` is both called for the old and new names, see below.

4.6.0 wil warn about them and 4.7 will remove the odl names.

pieterlexis: excellent

pieterlexis (approved):

PR #10289, opened 2021-04-13
Implement EDNS cookies in the auth

pieterlexis:
### Short description
This implements the siphash-based interoperable DNS COOKIES defined in
RFC 9018 for the authoritative server. The EDNSCookieOpt struct has been
expanded to accommodate this and can now has constructors and functions
to check and generate a server cookie.

Cookies will only be sent out if the client sent a cookie and the
`edns-cookie-secret` setting is configured. The auth will respond with
EDNS+FORMERR when the client cookie is malformed, BADCOOKIE when the
client sent a server cookie we can't decode or is invalid and a normal
response with a cookie (either new or sent by the client) when the
cookie can be validated.

There's no ratelimiting or policy based on the presence, absence, or validity of the cookie. This is something that will be implemented in dnsdist in a future PR.

I would like to sollicit comments on the behaviour of the EDNSCookie struct, as it might have some sharp edges that are created to prevent having to call the hash function more than once if the cookie has been validated.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code


pdns/test-packetcache_hh.cc
```diff
@@ -352,14 +349,8 @@ BOOST_AUTO_TEST_CASE(test_PacketCacheRecCollision) {
...
     ednsOptions.push_back(std::make_pair(EDNSOptionCode::ECS, makeEDNSSubnetOptsString(opt)));
-    EDNSCookiesOpt cookiesOpt;
-    cookiesOpt.client = string("deadbeef");
-    cookiesOpt.server = string("deadbeef");
-    cookiesOpt.server[4] = -20;
-    cookiesOpt.server[5] = -114;
-    cookiesOpt.server[6] = 0;
-    cookiesOpt.server[7] = 0;
-    ednsOptions.push_back(std::make_pair(EDNSOptionCode::COOKIE, makeEDNSCookiesOptString(cookiesOpt)));
+    EDNSCookiesOpt cookiesOpt(string("deadbeefdead\x11\xee\x00\x00").c_str(), 16);
```
rgacogne, line 352: I was surprised by the new values, wondering how the test could still pass. It turns out that we don't need to compute a hash collision for this since the content of cookies is skipped in the current packet cache code, so that doesn't really matter :-)

rgacogne (commented):
Code looks good :+1: I have a few questions below about design choices, nothing bad.

docs/settings.rst
```diff
@@ -617,6 +617,21 @@ ADDITIONAL section when sending a referral.
...
+``edns-cookie-secret``
+--------------------------
+
+.. versionadded:: 4.6.0
+
+-  String
+-  Default: (empty)
+
+When set, PowerDNS will respond with :rfc:`9018` EDNS Cookies to queries that have the EDNS0 Cookie option.
+PowerDNS will also respond with BADCOOKIE to clients that have no or a bad server cookie (section 5.2.3 and 5.2.4 of :rfc:`<7873>`).
```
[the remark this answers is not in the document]
mnordhoff: FWIW, Knot DNS uses BADCOOKIE; BIND evidently just accepts the query.

pdns/ednscookies.cc
```diff
@@ -20,32 +20,147 @@
...
+
+  string toHash = client + server.substr(0, 8) + source.toByteString();
+  string hashResult;
+  hashResult.resize(8);
+  crypto_shorthash(
+    reinterpret_cast<unsigned char*>(&hashResult[0]),
+    reinterpret_cast<const unsigned char*>(&toHash[0]),
+    toHash.length(),
+    reinterpret_cast<const unsigned char*>(&secret[0]));
+  valid = (server.substr(8) == hashResult);
```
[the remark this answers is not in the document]
pieterlexis: I had a look at moving this to misc.cc. This means that we suddenly need to link `dumresp` and `pdns_control` to libcrypto. perhaps we should move this to a file that already uses libcrypto anyway?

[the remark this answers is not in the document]
rgacogne: That would make sense! Or even to a new file since that function is not used in a lot of places.

docs/settings.rst
```diff
@@ -619,6 +619,21 @@ ADDITIONAL section when sending a referral.
...
+``edns-cookie-secret``
+--------------------------
+
+.. versionadded:: 4.6.0
+
+-  String
+-  Default: (empty)
+
+When set, PowerDNS will respond with :rfc:`9018` EDNS Cookies to queries that have the EDNS0 Cookie option.
+PowerDNS will also respond with BADCOOKIE to clients that have no or a bad server cookie (section 5.2.3 and 5.2.4 of :rfc:`7873`).
```
rgacogne, line 633: Perhaps:
```suggestion
PowerDNS will also respond with BADCOOKIE to clients that have sent only a client cookie, or a bad server cookie (section 5.2.3 and 5.2.4 of :rfc:`7873`).
```

pdns/Makefile.am
```diff
@@ -345,6 +347,7 @@ pdnsutil_SOURCES = \
 	dnssecsigner.cc \
 	dnswriter.cc dnswriter.hh \
 	dynlistener.cc \
+	ednscookies.cc \
```
rgacogne, line 350:
```suggestion
	ednscookies.cc ednscookies.hh \
```

pdns/common_startup.cc
```diff
@@ -161,6 +161,8 @@ void declareArguments()
   ::arg().setSwitch("any-to-tcp","Answer ANY queries with tc=1, shunting to TCP")="yes";
   ::arg().setSwitch("edns-subnet-processing","If we should act on EDNS Subnet options")="no";
 
+  ::arg().set("edns-cookie-secret", "When set, set a server cookie in a response to a query with a Client cookie (in hex)")="";
```
rgacogne, line 164:
```suggestion
  ::arg().set("edns-cookie-secret", "When set, set a server cookie when responding to a query with a Client cookie (in hex)")="";
```

pdns/common_startup.cc
```diff
@@ -563,6 +571,25 @@ void mainthread()
...
+       if (::arg()["edns-cookie-secret"].size() != EDNSCOOKIESECRETSIZE) {
+         throw std::range_error("wrong size (" + std::to_string(::arg()["edns-cookie-secret"].size()) + "), must be " + std::to_string(EDNSCOOKIESECRETSIZE));
+       }
+       DNSPacket::s_EDNSCookieKey = makeBytesFromHex(::arg()["edns-cookie-secret"]);
+     } catch(const std::range_error &e) {
+       g_log<<Logger::Error<<"edns-cookie-secret invalid: "<<e.what()<<endl;
+       exit(1);
+     }
+#else
+     g_log<<Logger::Error<<"Support for EDNS Cookies is not available because of missing cryptographic functions"<<endl;
```
rgacogne, line 588:
```suggestion
     g_log<<Logger::Error<<"Support for EDNS Cookies is not available because of missing cryptographic functions (libsodium support should be enabled, with the crypto_shorthash() function available)"<<endl;
```

pdns/ednscookies.hh
```diff
@@ -21,13 +21,58 @@
  */
 #pragma once
 #include "namespaces.hh"
+#include "iputils.hh"
+
+#define EDNSCOOKIESECRETSIZE 32
+#define EDNSCOOKIEOPTSIZE 24
```
rgacogne, line 27: Just a nit but I would prefer the use of static constants in the class below, so that these values are scoped. We could also perhaps add a `static_assert` checking that the secret size is `crypto_shorthash_KEYBYTES*2`, since it's hex-encoded?

pdns/ednscookies.cc
```diff
@@ -19,33 +19,155 @@
...
+
+  if (secret.length() != crypto_shorthash_KEYBYTES) {
+    return false;
+  }
+
+  server.clear();
+  server = "\x01"; // Version
+  server.resize(4, '\0'); // 3 reserved bytes
+  uint32_t now = htonl(static_cast<uint32_t>(time(nullptr)));
+  server += string(reinterpret_cast<const char*>(&now), 4);
```
rgacogne, line 157:
```suggestion
  server += string(reinterpret_cast<const char*>(&now), sizeof(now));
```

pdns/ednscookies.cc
```diff
@@ -19,33 +19,155 @@
...
+  if (isValid(secret, source) && !shouldRefresh()) {
+    return true;
+  }
+
+  if (secret.length() != crypto_shorthash_KEYBYTES) {
+    return false;
+  }
+
+  server.clear();
+  server = "\x01"; // Version
```
rgacogne, line 154:
```suggestion
  server.reserve(16);
  server = "\x01"; // Version
```

rgacogne (approved):
Great! I made a few suggestions but the code looks solid to me as it is.

klaus-nicat:
Just a question. If all servers use the same edns-cookie-secret, will a cookie generated by one Pdns-auth be accepted by another PDNS-auth (anycast scenario)?

pieterlexis:
> Just a question. If all servers use the same edns-cookie-secret, will a cookie generated by one Pdns-auth be accepted by another PDNS-auth (anycast scenario)?

Yes, provided it is still valid (cookies are valid for an hour).

pdns/ednscookies.cc
```diff
@@ -19,33 +19,158 @@
...
+  uint32_t now = static_cast<uint32_t>(time(nullptr));
+  // RFC 9018 section 4.3:
+  //    The DNS server
+  //    SHOULD allow cookies within a 1-hour period in the past and a
+  //    5-minute period into the future
+  if (rfc1982LessThan(now + 300, ts) && rfc1982LessThan(ts + 3600, now)) {
+    return false;
+  }
+  if (secret.length() != crypto_shorthash_KEYBYTES) {
+    // XXX should we throw std::range_error here?
```
Habbie, line 100: better to check it at startup, where you already have a catch for range_error anyway

pdns/dnspacket.hh
```diff
@@ -121,6 +122,9 @@ public:
   bool couldBeCached() const; //!< returns 0 if this query should bypass the packet cache
   bool hasEDNSSubnet() const;
   bool hasEDNS() const;
+  bool hasEDNSCookie() const;
+  bool hasWellFormedEDNSCookie() const;
+  bool hasValidEDNSCookie(); // Not const, some cookie params might be set
```
Habbie, line 127: I see nothing being set in any called code, so I put the `const` back (also in the .cc) and it compiles. I suspect the comment was true for a previous version :-)

Habbie (commented):
two nits, otherwise awesome!

pdns/ednscookies.cc
pieterlexis: we do, but this function could potentially also be called from other places to do cookie things

pdns/dnspacket.cc
```diff
@@ -325,6 +335,11 @@ void DNSPacket::wrapup()
         opts.push_back(make_pair(8, opt)); // 'EDNS SUBNET'
       }
 
+      if (d_haveednscookie && d_eco.isWellFormed()) {
+        d_eco.makeServerCookie(s_EDNSCookieKey, getRemote());
```
pieterlexis, line 339: @Habbie Should this become `getInnerRemote`?

pdns/ednscookies.cc
Habbie: We only check the length of the input string at startup - but I just re-read `makeBytesFromHex` and clearly it cannot generate an output string of the wrong length.

So, I guess the comment can just go.

pdns/dnspacket.cc
Habbie: after PROXY is merged, yes!

docs/settings.rst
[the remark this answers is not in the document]
Habbie: New cookie sounds good to me.

Habbie:
TODO, as far as I can tell:
* stop sending BADCOOKIE
* use `getInnerRemote()`
* fix conflicts

pieterlexis:
> stop sending BADCOOKIE

This technically defeats the idea of COOKIE, as a malicious client could just keep sending client only cookies

Habbie:
> This technically defeats the idea of COOKIE, as a malicious client could just keep sending client only cookies

As we don't give clients any special status based on their cookies of lack thereof, I'm not sure this is so important right now, but I'm okay either way.

Under https://datatracker.ietf.org/doc/html/rfc7873#section-5.2.3, BADCOOKIE is a legit choice for UDP, but not recommended for TCP. So let's change it for TCP at least.

PR #10292, opened 2021-04-14
rec: Fix TSAN complaints: max stacksize and response stats size counters

omoerbeek:
Start using new generic (atomic) histogram code and while there, fix a bug
concerning the rcode stats (which was not passed in by the caller by rec).

Note that this changes the histogram: instead of lumping all large counts into the last 2^16-1 bucket., there is an explicit bucket with bound MAX_UINT64

So the existing code has buckets
```
... 64800 65000 65535
``` 
with all values > 6500 end up in the last, we now have
```
... 64800 65000 MAX_UINT64
```

Draft since I have not tested auth yet.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/tcpreceiver.cc
+++ b/pdns/tcpreceiver.cc
@@ -23,6 +23,7 @@
 #include "config.h"
 #endif
 #include <boost/algorithm/string.hpp>
+#include <boost/scoped_array.hpp>
 #include "auth-packetcache.hh"
 #include "utility.hh"
 #include "threadname.hh"
--- a/.not-formatted
+++ b/.not-formatted
@@ -275,8 +275,6 @@
 ./pdns/resolver.cc
 ./pdns/resolver.hh
 ./pdns/responsestats-auth.cc
-./pdns/responsestats.cc
-./pdns/responsestats.hh
 ./pdns/rfc2136handler.cc
 ./pdns/root-addresses.hh
 ./pdns/root-dnssec.hh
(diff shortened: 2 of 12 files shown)


omoerbeek:
Investigating regress errors. Current hypothesis is that the new code uses <= for bucket boundaries, while the old code uses <.

omoerbeek:
Looks like the original histogram code is buggy:
```
RESPONSE LENGTH: 73
AFTER RESPONSE 73
20 0
40 0
60 1
80 0
...
```

omoerbeek:
The ` test-auth-regress-lmdb` failure looks like a fluke...

rgacogne:
Looks good, I made a few suggestions in a pull request (that was easier) at https://github.com/omoerbeek/pdns/pull/2

omoerbeek:
@Habbie can you take a look? This will change the response size bucket counts of auth (to be correct).

Habbie:
Fixing the bucket counts sounds good to me!

pdns/tcpreceiver.cc
```diff
@@ -23,6 +23,7 @@
 #include "config.h"
 #endif
 #include <boost/algorithm/string.hpp>
+#include <boost/scoped_array.hpp>
```
Habbie, line 26: This seems weird to me, why does tcpreceiver suddenly need this include?

omoerbeek: `tcpreceiver.cc` uses scoped arrays, previously it got the include (indirectly) via `responsestats.hh`, but that one is no longer including it, since it switched to using `std::array`s.

Habbie: thanks!

Habbie (approved):

PR #10299, opened 2021-04-15
rec: update setting for aggressive-nsec-cache-size

phonedph1:
### Short description
If `setSwitch` is used `--help` will offer you `yes` or `no` but these quickly produce an exception in `asNum(..)` so just use `set` instead?

Otherwise should it have a `--disable-XYZ` like the packet cache does?

### Checklist
I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


zeha:
setSwitch is wrong, yes.

omoerbeek:
= 0 means disabled, so no need for a separate "disabled" switch.

omoerbeek (approved):

rgacogne (approved):

PR #10302, opened 2021-04-16
Introduce LockGuarded, a lock-protected data

rgacogne:
### Short description

The general idea has been borrowed from Rust's locks: instead of defining two objects, the one to be protected, T, and the lock, we define a single `LockGuarded<T>` object which contains the object.
That provides two big advantages:
- it is immediately clear which data is protected by the lock
- that data simply can't be accessed without holding the lock.

This PR is a simple proof of concept, I moved a few types to demonstrate how it would work. If the general idea is approved, I'll do more tests including performance ones, and we can move more types when we feel like it.

Please start by looking at the new types introduced in `lock.hh` :-)

### Checklist

I have:
- [x] compiled this code
- [ ] tested this code
- [ ] documented the code


rgacogne:
- Rebased on master
- Added SharedLockGuarded, read-write lock equivalent to LockGuarded
- Converted almost all locks in dnsdist

Habbie (approved):
I love it! I only reviewed lock.hh.

pdns/lock.hh
```diff
@@ -147,3 +147,235 @@ private:
...
+  explicit SharedLockGuarded()
+  {
+  }
+
+  SharedLockGuardedTryHolder<T> try_lock()
+  {
+    return SharedLockGuardedTryHolder<T>(d_value, d_mutex);
+  }
+
+  SharedLockGuardedHolder<T> lock()
```
omoerbeek, line 363: Maybe rename to try_write_lock(), to remind the reader that is an exclusive lock?

omoerbeek:
I love it too. The `lock.hh` file could use some documentation and/or examples, and a more clear distinction between the helper (Holder) classes and the classes to be used by the developer explicitly.

omoerbeek:
I think it is time to move this to non-Draft and do a final review. AFAIK all products are in the start of a release cycle.

I'd like to see a separate reformat config for loch.hh, but that is certainly not a blocker.

rgacogne:
> I think it is time to move this to non-Draft and do a final review. AFAIK all products are in the start of a release cycle.

I would like that as well but the PR is currently not ready to be reviewed. There are changes in the recursor caches that I don't like, and the auth caches have not been converted. I have also not written the documentation we discussed in lock.hh.

> I'd like to see a separate reformat config for loch.hh, but that is certainly not a blocker.

reformat config?

rgacogne:
Alright, I believe this PR is finally ready for review :-)
It's a big PR, but I tried to do only one change by commit so we can drop some of them if needed, and review commit by commit as well.

rgacogne:
Rebased to fix a conflict in `pdns_recursor.cc`.

pdns/recursor_cache.hh
```diff
@@ -201,55 +201,60 @@ private:
...
+  Entries getEntries(MapCombo::LockedContent& content, const DNSName &qname, const QType qt, const OptTag& rtag);
+  cache_t::const_iterator getEntryUsingECSIndex(MapCombo::LockedContent& content, time_t now, const DNSName &qname, QType qtype, bool requireAuth, const ComboAddress& who);
 
-  time_t handleHit(MapCombo& map, OrderedTagIterator_t& entry, const DNSName& qname, uint32_t& origTTL, vector<DNSRecord>* res, vector<std::shared_ptr<RRSIGRecordContent>>* signatures, std::vector<std::shared_ptr<DNSRecord>>* authorityRecs, bool* variable, boost::optional<vState>& state, bool* wasAuth, DNSName* authZone);
+  time_t handleHit(MapCombo::LockedContent& content, OrderedTagIterator_t& entry, const DNSName& qname, uint32_t& origTTL, vector<DNSRecord>* res, vector<std::shared_ptr<RRSIGRecordContent>>* signatures, std::vector<std::shared_ptr<DNSRecord>>* authorityRecs, bool* variable, boost::optional<vState>& state, bool* wasAuth, DNSName* authZone);
 
 public:
-  static LockGuardedTryHolder<MapCombo> lock(LockGuarded<MapCombo>& map)
+  static LockGuardedTryHolder<MapCombo::LockedContent> lock(LockGuarded<MapCombo::LockedContent>& content)
```
omoerbeek, line 246: I think it would be nicer if `lock` would be a method of `LockedContent` instead of a `static` method of `MemRecursorCache`.

Same for negcache. I have a local version with those changed here if you agree.

omoerbeek (approved):
I reviewed `lock.hh` and the recursor parts. I suggest some changes. Butt approved either way.

pdns/recursor_cache.cc
```diff
@@ -19,23 +19,22 @@ MemRecursorCache::MemRecursorCache(size_t mapsCount) : d_maps(mapsCount)
 {
 }
 
-size_t MemRecursorCache::size()
+size_t MemRecursorCache::size() const
```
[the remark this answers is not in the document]
rgacogne: There is trade-off to be considered, since having a single atomic counter would increase the number of times that a cache line has to be cleared. Unless I'm mistaken we don't call `size()` too often, so perhaps the current behaviour is better?

pdns/nod.hh
```diff
@@ -42,31 +43,28 @@ namespace nod {
...
   private:
     void remove_tmp_files(const boost::filesystem::path&, std::lock_guard<std::mutex>&);
 
     bool d_init{false};
-    bf::stableBF d_sbf; // Stable Bloom Filter
+    LockGuarded<bf::stableBF> d_sbf; // Stable Bloom Filter
     std::string d_cachedir;
     std::string d_prefix = sbf_prefix;
-    std::mutex d_sbf_mutex; // Per-instance mutex for snapshots
     static std::mutex d_cachedir_mutex; // One mutex for all instances of this class
```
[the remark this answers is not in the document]
rgacogne: I left that one out because I don't understand it at all. The mutex is static, with a comment explaining that it's because there should only be one thread looking at at that dir at once:
```
// The mutex has to be static because we can't have multiple (i.e. per-thread)                                                                                                                                
// instances iterating and writing to the cache dir at the same time    
```
But I don't see anything preventing different instances from having different directories, and moreover we don't alway take the lock when we access the directory in `snapshotCurrent()` (but sometimes it is called from `init()` where we hold the lock).
If the mutex is supposed to protect all accesses to the directory, we need to make sure it is held in all places. If it's only supposed to protect calls to `init()`, perhaps we should rename it. If all instances use the same directory, we can perhaps make `d_cachedir` `static` and wrap it.
@neilcook ?

pdns/recursor_cache.hh
rgacogne: Good idea!

omoerbeek: Committed

pdns/nod.hh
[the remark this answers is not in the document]
omoerbeek: Looking a bit mode closer, the current code is indeed inconsistent.

rgacogne:
Rebased to fix conflicts.

rgacogne:
I'd really like to be able to use these in new code, so I think I'll split this pull request in two, one with the new feature and the changes for dnsdist and the rec, which are small, and a second one for the auth which is significantly bigger.

rgacogne:
Replaced by #10649 and #10653.

PR #10303, opened 2021-04-19
rec: Use the correct ECS address when proxy-protocol is enabled

rgacogne:
### Short description

The recursor was using the address of the proxy instead of the one of the proxied client to generate the EDNS Client Subnet value to send to authoritative servers.
Reported on the mailing-list: https://mailman.powerdns.com/pipermail/pdns-users/2021-April/027168.html

I reproduced the issue locally and tested the fix, but this PR could use a regression test.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/pdns_recursor.cc
+++ b/pdns/pdns_recursor.cc
@@ -1620,7 +1620,7 @@ static void startDoResolve(void *p)
 #ifdef HAVE_FSTRM
     sr.setFrameStreamServers(t_frameStreamServers);
 #endif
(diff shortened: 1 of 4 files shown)


rgacogne:
Regression test added.

omoerbeek:
This looks good (though the regression test is not happy yet). I also verified the address used to see if padding should be used is right, and I think it is: it's using the address of the proxy as intended and documented.

rgacogne:
I managed to botch the Proxy Protocol regression tests, I'll fix that.

omoerbeek (approved):

PR #10329, opened 2021-04-23
boost 1.76 containers: use standard exceptions

Habbie:
### Short description

https://www.boost.org/users/history/version_1_76_0.html

> Replaced default standard exception classes with Boost.Container own classes, reducing considerably the included files overhead. Example: in MSVC 19 boost/container/vector.hpp preprocessed file size reduces from 1,5MB to 930KB. If you still want to use standard exception classes, you can define BOOST_CONTAINER_USE_STD_EXCEPTIONS before using any Boost.Container class.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/Makefile.am
+++ b/pdns/Makefile.am
@@ -7,7 +7,8 @@ AM_CPPFLAGS += \
 	$(LIBCRYPTO_INCLUDES) \
 	$(SYSTEMD_CFLAGS) \
 	$(YAML_CFLAGS) \
-	-I$(top_srcdir)/ext/protozero/include
+	-I$(top_srcdir)/ext/protozero/include \
+	-DBOOST_CONTAINER_USE_STD_EXCEPTIONS
 
 AM_CXXFLAGS = \
 	-DSYSCONFDIR=\"$(sysconfdir)\" \
--- a/pdns/dnsdistdist/Makefile.am
+++ b/pdns/dnsdistdist/Makefile.am
@@ -7,7 +7,8 @@ AM_CPPFLAGS += $(SYSTEMD_CFLAGS) \
 	$(NET_SNMP_CFLAGS) \
 	$(LIBCAP_CFLAGS) \
 	-I$(top_srcdir)/ext/protozero/include \
-	-DSYSCONFDIR=\"${sysconfdir}\"
+	-DSYSCONFDIR=\"${sysconfdir}\" \
+	-DBOOST_CONTAINER_USE_STD_EXCEPTIONS
 
 ACLOCAL_AMFLAGS = -I m4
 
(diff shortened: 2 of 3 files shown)


Habbie:
@omoerbeek asked for your review because I touched one recursor file - although I don't think we'd ever see a container exception in there. Also, please judge the rec backport labels. Thanks!

omoerbeek:
Isn't it better to define this in the Makefile? A new use of a boost container in a random place won't get it if you do it like this.

Habbie:
I considered that but then I saw we only had a few occurrences. But, still makes sense. I'll change it.

omoerbeek (approved):
OK with me, though I'm wondering---considering the versions using `std::` exceptions are so big---if we're indeed better off using these container instead of "ordinary" `std::` ones.  I did not look into the actual use of the Boost containers to answer this question, though.

Habbie:
> OK with me, though I'm wondering---considering the versions using `std::` exceptions are so big---if we're indeed better off using these container instead of "ordinary" `std::` ones.

I had the impression it was just 'parsing lots of headers', not 'making the binary bigger', but I did not check.

omoerbeek:
Somwhat off-topic: IMO, if there is a good fit in the standard C++ library for a certain data structure, it's preferably to use it instead of a boost one. As an illustration, the `boost::container::flat_set<uint16_t>`  of UDP port numbers in `pdns_recursor.cc` could also be a `std::bitset<65535>`.

Boost has many nice things, but it is less stable, as the inclusion of the define in this PR shows.

rgacogne:
Do you know if that change is limited to `boost::container` classes, or if there is a risk that it somehow spreads to `boost::circular_buffer` or `boost::multi_index` as well?

rgacogne:
> if there is a good fit in the standard C++ library for a certain data structure, it's preferably to use it instead of a boost one

Agreed. I'd like to replace our use of `boost::shared_array`, `boost::scope_array`, `boost::tuple`, `boost::tie` as well whenever possible.

omoerbeek:
> Do you know if that change is limited to `boost::container` classes, or if there is a risk that it somehow spreads to `boost::circular_buffer` or `boost::multi_index` as well?

Grepped in the boost 1.76 tree, so far the `BOOST_CONTAINER_USE_STD_EXCEPTIONS` only occurs in the `boost/container` subdir. Both circular buffer and mult-index live in a separate dir. Release notes also say it's for `boost/containers` only.

rgacogne:
My worry is more about other parts of boost, especially non header-only ones, using `boost::container` objects, and thus throwing unexpected exceptions all of sudden. But it's hard to know for sure :-/

omoerbeek:
> My worry is more about other parts of boost, especially non header-only ones, using `boost::container` objects, and thus throwing unexpected exceptions all of sudden. But it's hard to know for sure :-/

Both multi-index and circular buffer are include-only. So those are safe, as for other parts, I hope the boost devs are aware of this potential issue...

rgacogne (approved):

PR #10336, opened 2021-04-26
Get rid of our string_view selection process, C++17 has all we need

rgacogne:
### Short description

Inspired by the discussion in https://github.com/PowerDNS/pdns/pull/10329.

I only tested that change on Arch, let's see whether our CI likes it.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


omoerbeek (approved):
The almost almost empty `views.hh` annoys me a bit, but it's a rather boring job to fix it... We should do that some day.

rgacogne:
> The almost almost empty views.hh annoys me a bit, but it's a rather boring job to fix it... We should do that some day.

I pondered that, getting rid of the alias would make backports very painful, but we could move the `using` line to a different header if we want.

omoerbeek:
Move it to `namespaces.hh` ?

rgacogne:
> Move it to `namespaces.hh` ?

Done!

PR #10338, opened 2021-04-28
dnsdist: Implement cross-protocol queries

rgacogne:
### Short description

This PR implements cross-protocol queries, allowing passing a query received over a given protocol to a backend over a different protocol. Until now, a query received over UDP was passed over UDP, TCP over TCP, DoT over TCP and DoH over UDP. This PRs implements:
- support for outgoing DoT (DoT between dnsdist and the backend)
- fallback to TCP in case of truncation for queries receiver over DoH
- passing queries received over UDP and DoH to TCP or DoT.

It lacks:
- documentation
- tests
- performance testing
- TLS session resumption between dnsdist and the backend
- a good cleanup.
- 
### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


rgacogne:
Rebased to fix conflicts.

rgacogne:
Rebased to fix conflicts.

rgacogne:
Rebased on master.

rgacogne:
Otto, I just requested a review from you but to be clear I don't expect you to look at the dnsdist bits, certainly not in depth. It would be nice if you could have a look at the parts that impact the recursor, like TCPIOHandler, though :)

omoerbeek:
On OpenBSD I'm hitting this static assert in `DOHUnit::handleUDPResponse()`:
```
static_assert(sizeof(*this) <= PIPE_BUF, "Writes up to PIPE_BUF are guaranteed not to be interleaved and to either fully succeed or fail")
```
Actually Im wondering if the `static_assert` is correct, since in the end a pointer is being written if I see things correctly.

rgacogne:
> Actually I'm wondering if the static_assert is correct, since in the end a pointer is being written if I see things correctly.

It's not, and it should not even be here since it calls `sendDoHUnitToTheMainThread` which has the proper check.

bjornfro:
Wanted to test support for outgoing DoT (DoT between dnsdist and the backend).  However, does that support having a hostname(dns-record) for the DoT server? In my scenario we cannot "hardcode" to an IP-address. If so, how is that configured? I get "Error creating new server with address.....Unable to convert presentation address". 

In our case the DNS record of the DoT server has two IP-addresses. 

Regards, Bjorn

Trying with: 
Version     : 1.7.0
Release     : 0.alpha0.master.594.gfc905e965.1pdns.el7

rgacogne:
No, that's not supported and not in target for 1.7.0, but perhaps the workarounds suggested in https://github.com/PowerDNS/pdns/issues/10413 might help.

PR #10347, opened 2021-05-03
dnsdist: Add ChangeLog and secpoll for 1.6.0-rc2

rgacogne:
### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code


omoerbeek (approved):
rec uses secpoll status 3 for old pre-releases, but that's debatable

rgacogne:
> rec uses secpoll status 3 for old pre-releases, but that's debatable

I'll think about it, as we can change that later :)

omoerbeek:
> > rec uses secpoll status 3 for old pre-releases, but that's debatable
> 
> I'll think about it, as we can change that later :)

My motivation is mostly to stimulate people to use the latest pre-release.

rgacogne:
Understood, but on the other hand I don't want to blur the line between 'unsupported' and 'known vulnerable'.

PR #10349, opened 2021-05-03
rec: move to tcpiohandler for outgoing TCP

omoerbeek:
This is needed to be able to act as a DoT client, but I first start with regular TCP53.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/pdns_recursor.cc
+++ b/pdns/pdns_recursor.cc
@@ -398,14 +398,14 @@ static bool isHandlerThread()
 
 static void handleTCPClientWritable(int fd, FDMultiplexer::funcparam_t& var);
 
-LWResult::Result asendtcp(const string& data, Socket* sock)
+LWResult::Result asendtcp(const PacketBuffer& data, Socket* sock)
 {
   PacketID pident;
   pident.tcpsock=sock->getHandle();
-  pident.outMSG=data;
+  pident.outMSG = data;
 
   t_fdm->addWriteFD(sock->getHandle(), handleTCPClientWritable, pident);
-  string packet;
+  PacketBuffer packet;
 
   int ret = MT->waitEvent(pident, &packet, g_networkTimeoutMsec);
   if (ret == 0) { //timeout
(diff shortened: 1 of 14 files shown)


pdns/syncres.hh
```diff
@@ -920,43 +921,47 @@ private:
 
 class Socket;
 /* external functions, opaque to us */
-LWResult::Result asendtcp(const string& data, Socket* sock);
-LWResult::Result arecvtcp(string& data, size_t len, Socket* sock, bool incompleteOkay);
+LWResult::Result asendtcp(const PacketBuffer& data, Socket* sock);
+LWResult::Result arecvtcp(PacketBuffer& data, size_t len, Socket* sock, bool incompleteOkay);
+LWResult::Result asendtcp(const PacketBuffer& data, TCPIOHandler&);
+LWResult::Result arecvtcp(PacketBuffer& data, size_t len, TCPIOHandler&, bool incompleteOkay);
```
rgacogne, line 927: I wonder whether it would make sense to use a `TCPIOHandler` in the webserver and the carbon exporter as well, so we can have a single version of that code?

pdns/syncres.hh
```diff
@@ -920,43 +921,47 @@ private:
...
-  string::size_type outPos;    // how far we are along in the outMSG
+  PacketBuffer inMSG; // they'll go here
+  PacketBuffer outMSG; // the outgoing message that needs to be sent
 
   typedef set<uint16_t > chain_t;
   mutable chain_t chain;
-  mutable uint32_t nearMisses; // number of near misses - host correct, id wrong
-  int fd;
-  mutable bool closed; // Processing already started, don't accept new chained ids
+  TCPIOHandler *tcphandler{nullptr};
```
rgacogne, line 944: While the current code seems fine, I wonder if it would make sense to make that a shared pointer so we never have a dangling pointer there? There is a small overhead, of course.

pdns/pdns_recursor.cc
```diff
@@ -4054,13 +4150,44 @@ static void handleTCPClientWritable(int fd, FDMultiplexer::funcparam_t& var)
...
+{
+  PacketID* pid = boost::any_cast<PacketID>(&var);
+  assert(pid->tcphandler != nullptr);
+  assert(fd == pid->tcphandler->getDescriptor());
+
+  try {
+    IOState state = pid->tcphandler->tryWrite(pid->outMSG, pid->outPos, pid->outMSG.size());
+    switch (state) {
+    case IOState::Done: {
+      PacketID tmp = *pid;
```
rgacogne, line 4168: I wonder if we could move `*pid` here and in the exception handling below, to prevent a full alloc+copy?

rgacogne (approved):
I made a few suggestions, but the code looks good for plain TCP. We will need to handle IO events a bit differently for TLS, though.

omoerbeek:
I will process the suggestions in a new PR that also will do outgoing TLS.

PR #10353, opened 2021-05-04
rec: Apply dns64 on RPZ hits generated after a gettag_ffi hit

rgacogne:
### Short description

We do special case the `qname RPZ` processing after a `gettag_ffi` hit, leading to `dns64` to not be applied in that case. This commit adds `dns64` handling to the special case.

<strike>Needs regression tests.</strike>

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/pdns_recursor.cc
+++ b/pdns/pdns_recursor.cc
@@ -1476,6 +1476,24 @@ int getFakePTRRecords(const DNSName& qname, vector<DNSRecord>& ret)
   return rcode;
 }
 
+static bool answerIsNOData(uint16_t requestedType, int rcode, const std::vector<DNSRecord>& records)
+{
+  if (rcode != RCode::NoError) {
+    return false;
+  }
+  for (const auto& rec : records) {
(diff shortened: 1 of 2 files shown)


pdns/pdns_recursor.cc
```diff
@@ -1742,6 +1760,10 @@ static void startDoResolve(void *p)
         else {
           auto policyResult = handlePolicyHit(appliedPolicy, dc, sr, res, ret, pw);
           if (policyResult == PolicyResult::HaveAnswer) {
+            if (dq.qtype == QType::AAAA && answerIsNOData(dc->d_mdp.d_qtype, res, ret) && g_dns64Prefix) {
```
omoerbeek, line 1763:
```suggestion
            if (g_dns64Prefix && dq.qtype == QType::AAAA && answerIsNOData(dc->d_mdp.d_qtype, res, ret)) {
```

omoerbeek (approved):
One small nit: test most simple condition first

PR #10364, opened 2021-05-06
auth: Don't choke on non-base64 values when importing zone keys

rgacogne:
### Short description

`DNSCryptoKeyEngine::makeFromISCFile()`, called by `pdnsutil import-zone-key` or the API, for example, would try to parse almost all values as a base64 string. Depending on the version of OpenSSL, it could have led to a weird `Error: BIO_read failed to read all data from memory buffer` error when the file contains a non-base64 value, like for example:
```
Flags: 257
```
Recent versions of OpenSSL seem to simply return that the value could not be parsed, but older ones (OpenSSL 1.0.2k from CentOS 7 for example) would report an incomplete read (`BIO_should_retry()` returning 1), triggering an exception that prevents the key from being loaded.
This commits keeps a longer list of known non-base64 values, but more importantly catch the base64 decoding exception and then store the initial value instead of aborting. Only failure to decode known base64 values prevents the key from being loaded.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnssecinfra.cc
+++ b/pdns/dnssecinfra.cc
@@ -130,7 +130,7 @@ std::unique_ptr<DNSCryptoKeyEngine> DNSCryptoKeyEngine::makeFromISCString(DNSKEY
           }
         }
         catch (const std::exception& e) {
-          throw std::runtime_error("Error while trying to pase the value of the '" + key + "' key from the ISC map: " + e.what());
+          throw std::runtime_error("Error while trying to parse the numeric value of the '" + key + "' key from the ISC map: " + e.what());
         }
       }
     }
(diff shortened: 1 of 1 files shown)


Habbie (approved):

pdns/dnssecinfra.cc
```diff
@@ -76,50 +76,87 @@ std::unique_ptr<DNSCryptoKeyEngine> DNSCryptoKeyEngine::makeFromISCFile(DNSKEYRe
...
+      else if (it->second == KeyTypes::numeric) {
+        try {
+          unsigned int num = pdns_stou(value);
+          stormap[key] = std::to_string(num);
+          if (key == "algorithm") {
+            algorithm = num;
+          }
+        }
+        catch (const std::exception& e) {
+          throw std::runtime_error("Error while trying to pase the value of the '" + key + "' key from the ISC map: " + e.what());
```
Habbie, line 133:
```suggestion
          throw std::runtime_error("Error while trying to parse the numeric value of the '" + key + "' key from the ISC map: " + e.what());
```

Habbie (changes requested):
one typo, otherwise approved

Habbie (approved):

PR #10373, opened 2021-05-10
auth pdnsutil add-zone-key: clarify zsk default

### LINKED ISSUE #10372: pdnsutil add-zone-key adds flag 256 CSK by default if no KSK|ZSK option is provided
- Program: Authoritative
 - Issue type: Bug report

### Short description
If the ZSK|KSK option is not provided on pdnsutil add-zone-key it will generate a flag 256 CSK which is confusing because most registrars default to flag 257 when you add DNSSEC public keys|DS on their (web) interfaces.

### Environment
 - Operating system: Debian buster/10
 - Software version: 4.4.1-1pdns.buster

### Steps to reproduce
1. pdnsutil create-zone example.com
2. pdnsutil add-zone-key example.com ECDSAP256SHA256
3. pdnsutil show-zone example.com

### Expected behaviour
Default behaviour of pdnsutil add-zone-key should match what secure-zone does, so adding a CSK with flag 257 instead of flag 256.

### Actual behaviour
```
root@shinjuku:~# pdnsutil create-zone example.com
Creating empty zone 'example.com'
root@shinjuku:~# pdnsutil show-zone example.com
This is a Native zone
Zone is not actively secured
Metadata items: None
No keys for zone 'example.com'.
root@shinjuku:~# pdnsutil show-zone example.com
This is a Native zone
Zone is not actively secured
Metadata items: None
Zone has NSEC semantics
keys: 
ID = 64 (CSK), flags = 256, tag = 62210, algo = 13, bits = 256  Inactive         Published  ( ECDSAP256SHA256 ) 
CSK DNSKEY = example.com. IN DNSKEY 256 3 13 vg0ZmSiHnVpKSG2dvYgbOKF/9AiR+iKfsHF77QQyS97RZwVvWj9uTU3YXwqGL9OL7SG8XAWaUVfa7R9bFRVEZQ== ; ( ECDSAP256SHA256 )
DS = example.com. IN DS 62210 13 1 27dcdb4192fb325b6b54e3cffc8c498892e97de9 ; ( SHA1 digest )
DS = example.com. IN DS 62210 13 2 14a224f00a41ee22d3114d4070f4774b81d18ce94aee3af545b3002240f28384 ; ( SHA256 digest )
DS = example.com. IN DS 62210 13 4 dd632c3df98249d3fb3e750f72871390577bc23aae36a1cb1e09a0ea9de51426a2634958db27f40a95a8877d70b422b2 ; ( SHA-384 digest )
```

vs.

```
root@shinjuku:~# pdnsutil remove-zone-key example.com 64
root@shinjuku:~# pdnsutil show-zone example.com
This is a Native zone
Zone is not actively secured
Metadata items: None
No keys for zone 'example.com'.
root@shinjuku:~# pdnsutil secure-zone example.com
Securing zone with default key size
Adding CSK (257) with algorithm ecdsa256
Zone example.com secured
Adding NSEC ordering information 
root@shinjuku:~# pdnsutil show-zone example.com
This is a Native zone
Metadata items: None
Zone has NSEC semantics
keys: 
ID = 65 (CSK), flags = 257, tag = 64921, algo = 13, bits = 256    Active         Published  ( ECDSAP256SHA256 ) 
CSK DNSKEY = example.com. IN DNSKEY 257 3 13 6UthEyCM1CWFFv41q2PUE3l494bO/1o77vRerD526vRjhWvpbI3ahhWw7aK2pQIxQkpVwAUlf3IEn1lM4gh3LA== ; ( ECDSAP256SHA256 )
DS = example.com. IN DS 64921 13 1 763c0a167e85220351cd6e5dc296b3c989fd2fb3 ; ( SHA1 digest )
DS = example.com. IN DS 64921 13 2 34ea98bf4d91ed13b6bdb57cfdc4c7304b3c882890d98bd9e28ad29feb5d77e5 ; ( SHA256 digest )
DS = example.com. IN DS 64921 13 4 1e4c67b44b3b670306c5d70c6dedb36fe6e60741b3210efcf867a773497b2f240b211b15c5b0ec1ed3bedc97fc809a75 ; ( SHA-384 digest )
```

Habbie:
### Short description
fixes #10372

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/docs/manpages/pdnsutil.1.rst
+++ b/docs/manpages/pdnsutil.1.rst
@@ -48,8 +48,8 @@ algorithms are supported:
 
 activate-zone-key *ZONE* *KEY-ID*
     Activate a key with id *KEY-ID* within a zone called *ZONE*.
-add-zone-key *ZONE* {**KSK**,\ **ZSK**} [**active**,\ **inactive**] [**published**,\ **unpublished**] *KEYBITS* *ALGORITHM*
-    Create a new key for zone *ZONE*, and make it a KSK or a ZSK, with
+add-zone-key *ZONE* [**KSK**,\ **ZSK**] [**active**,\ **inactive**] [**published**,\ **unpublished**] *KEYBITS* *ALGORITHM*
+    Create a new key for zone *ZONE*, and make it a KSK or a ZSK (default), with
     the specified algorithm. The key is inactive by default, set it to
     **active** to immediately use it to sign *ZONE*. The key is published
     in the zone by default, set it to **unpublished** to keep it from
--- a/pdns/pdnsutil.cc
+++ b/pdns/pdnsutil.cc
@@ -2673,7 +2673,7 @@ try
 
   else if (cmds.at(0) == "add-zone-key") {
     if(cmds.size() < 3 ) {
-      cerr << "Syntax: pdnsutil add-zone-key ZONE zsk|ksk [BITS] [active|inactive] [rsasha1|rsasha1-nsec3-sha1|rsasha256|rsasha512|ecdsa256|ecdsa384";
+      cerr << "Syntax: pdnsutil add-zone-key ZONE [zsk|ksk] [BITS] [active|inactive] [rsasha1|rsasha1-nsec3-sha1|rsasha256|rsasha512|ecdsa256|ecdsa384";
 #if defined(HAVE_LIBSODIUM) || defined(HAVE_LIBDECAF) || defined(HAVE_LIBCRYPTO_ED25519)
       cerr << "|ed25519";
 #endif
@@ -2681,6 +2681,8 @@ try
       cerr << "|ed448";
 #endif
       cerr << "]"<<endl;
+      cerr << endl;
+      cerr << "If zsk|ksk is omitted, add-zone-key makes a key with flags 256 (a 'ZSK')."<<endl;
       return 0;
     }
     DNSName zone(cmds.at(1));


mind04:
Changing this default, will definitely do more harm than good for existing dnssec implementations. I'm against merging this change.

Habbie:
> Changing this default, will definitely do more harm than good for existing dnssec implementations.

Why?

klaus-nicat:
What about enforcing a keytype <ZSK|KSK|CSK> to avoid ambiguity?

mind04:
For example, ZSK rollovers will become a mess when when you rely on the default. Suddenly you end up with a bunch of 257 keys instead of one 257 and one/or two 256 keys. After the last ZSK is gone the KSKs will magically become CSKs, breaking all the "grep KSK" in scripts interacting with the registry. A "you should have read the upgrade notes" is not good enough for an unnecessary and potential extremely breaking change like this. Properly document the current default will avoid all of this and will also fix #10372 since the real issue here is unexpected behavior.

mind04:
Better, but I'm still afraid this will only move the pain to a later date. The problem is that the waning will not be noticed in exiting implementations.

Habbie:
Right. The question then becomes - will people notice `pdnsutil` *failing* in 4.6, or not.

We can still add the warning, and then not change it into an error later. Then again, the tool already tells you what it did.

mind04:
Why is properly document the current default behavior not an option?
Especially since the default was like this since 3.0.1.

Habbie:
> Why is properly document the current default behavior not an option?

It is an option, but it does not help users who get a ZSK while they were expecting a KSK. But, to be honest, those are always real users, and not scripts, and there is only so much we can do for them (as the tool already reports what it does).



>  Especially since the default was like this since 3.0.1.

We could argue that it should have changed when we changed to the default 257 CSK.

But, I agree, we'll document this and then it's done.

Habbie:
PR changed to just clarify the ZSK behaviour.

mind04 (approved):

PR #10381, opened 2021-05-10
dnsdist: Reorganize the IDState and Rings fields

rgacogne:
### Short description

Reducing the space lost to padding and thus the memory usage. This change saves 1 MB of memory per downstream server in the default configuration, and around 8 bytes per entry in the ring buffer. It also adds a few words about memory consumption in the documentation.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/pdns/dnsdist-rings.hh
+++ b/pdns/dnsdist-rings.hh
@@ -36,23 +36,23 @@
 struct Rings {
   struct Query
   {
-    struct timespec when;
     ComboAddress requestor;
     DNSName name;
+    struct timespec when;
+    struct dnsheader dh;
     uint16_t size;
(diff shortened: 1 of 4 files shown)


omoerbeek (approved):
Love this! Do you use a tool to help you get the sizes and alignments? If so, I like to know how. I would use in gdb `ptype /o IDState`

rgacogne:
> Do you use a tool to help you get the sizes and alignments? If so, I like to know how. I would use in `gdb ptype /o IDState`

Unfortunately I don't. I was doing it by looking at the `struct`s in `gdb` but your exact way is much better! Thanks :)

PR #10383, opened 2021-05-11
rec: Backport 10303 to rec 4.4.x: Use the correct ECS address when proxy-protocol is enabled 

omoerbeek:
### Short description

Backport of #10303 

### Checklist

I have:
- [X] compiled this code
- [ ] tested this code
- [ ] documented the code


rgacogne (approved):

omoerbeek:
@rgacogne the extra regression test added in this PR (`testTooLargeToAddZeroScope`) fails since 4.4.x does not have the extra size check when adding ECS to the reply. That check was added in a separate PR (I think the padding work). So I'm going to remove this regression test from 4.4.x. OK?

rgacogne:
I'm OK with removing the test from this PR, but I wonder if we should backport that check (and the test) to 4.4.x?

omoerbeek:
> I'm OK with removing the test from this PR, but I wonder if we should backport that check (and the test) to 4.4.x?

Yes, we should look into that and then re-add the extra regression test. But it's not a straightforward backport, since the original is tied to the new padding code.

PR #10392, opened 2021-05-11
auth gpgsql: use SELECT .. RETURNING to get inserted row ID

### LINKED ISSUE #7175: Postgresql Backend: using lastval is dangerous to get the last sequence value
- Program: Authoritative
 - Issue type: Bug report

### Short description
I found this issue while trying "pdnsutil secure-zone". pdnsutil does:
1. Create and insert the new key: ''insert into cryptokeys (domain_id, flags, active, content) select ...''
2. uses ''select lastval()'' to retrieve the ''id'' of the currently inserted row.

The bug is, that PDNS assumes that the INSERT action only changes the sequence of the cryptokeys.id column. This may be wrong. There can be additional sequences be changed, eg. due to triggers on tables.

For example we use Slony for database replication. Slony is trigger based and uses plenty of sequences internally. Hence, lastval() returns a value of some Slony internal sequence instead of the cryptokey's id.

AFAIS, the above scenario is the only place where ''lastval'' is currently used. Hence, it could be changed to ''currval('cryptokeys_id_seq')''. Whereas I am not sure that the sequence has the same name on every installation. Or a config option to specify the sequence name? Or replaceing lastval with some other SELECT query to fetch the id?

Another workaround would be to document how a zone can be signed manually, instead of using pdnsutil. BTW: what is the preferred programatical way to enable DNSSEC? Ie. can it be done via the API?

Thanks
Klaus

Habbie:
### Short description
Fixes #7175

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/modules/gpgsqlbackend/gpgsqlbackend.cc
+++ b/modules/gpgsqlbackend/gpgsqlbackend.cc
@@ -141,8 +141,8 @@ public:
     declare(suffix, "delete-rrset-query", "", "delete from records where domain_id=$1 and name=$2 and type=$3");
     declare(suffix, "delete-names-query", "", "delete from records where domain_id=$1 and name=$2");
 
-    declare(suffix, "add-domain-key-query", "", "insert into cryptokeys (domain_id, flags, active, published, content) select id, $1, $2, $3, $4 from domains where name=$5");
-    declare(suffix, "get-last-inserted-key-id-query", "", "select lastval()");
+    declare(suffix, "add-domain-key-query", "", "insert into cryptokeys (domain_id, flags, active, published, content) select id, $1, $2, $3, $4 from domains where name=$5 returning id");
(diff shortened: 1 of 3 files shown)


Habbie:
And I should turn the lastval query into an error for PG.

Habbie:
I should still call reset when there is no next row.

Habbie:
> I should check if other databases can do this too.

SQLite can do it, but only since two months ago. MSSQL could probably do it with a procedure, but I'm not going to figure that out now. MySQL does not have this trick, but guarantees that `LAST_INSERT_ID()` is for the -first- row inserted by the last query, so any triggers that run after that, do not affect it.

Habbie:
Ready for review.

zeha:
Makes sense to me.

zeha (approved):

PR #10401, opened 2021-05-12
auth: change the consistent-backends default to 'yes'

mind04:
### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


zeha:
I've set the 4.5.0-alpha1 milestone so this change will be **considered**. Hope this is in your interest, @Habbie.

Habbie:
I think this is an excellent idea, if we also put it in `upgrading.rst` :)

Habbie (approved):

PR #10410, opened 2021-05-14
dnsdist: Fix EDNS bug when ECS is enabled for DNS NOTIFY (resolves #10409)

### LINKED ISSUE #10409: dnsdist: Error while parsing a NOTIFY query when useClientSubnet=true is set
- Program: dnsdist 
 - Issue type: Bug report

### Short description

Basically as stated in the title. I have a server that has `useClientSubnet=true` and whenever it receives a NOTIFY message it errors out and drops the query.

The exception message is this and it's similar for TCP/UDP:
```
Got an error while parsing a UDP query from 127.0.0.1:34963, id 19538: slowRewriteQueryWithExistingEDNS() should not be called for queries that have no EDNS
```

### Environment

 - Operating system: Ubuntu 20.04 Linux
 - Software version: commit 69eddfc8e07143912448b989da79231cbb5ad27e (Tue May 11 15:25:08 2021 +0200)

### Steps to reproduce

1. Run dnsdist with a server with the following configuration:
```
newServer({
  address="127.0.0.1:5300",
  useClientSubnet=true,
})
setLocal("0.0.0.0:5300")
setACL({"0.0.0.0/0", "::/0"})
```
2. Trigger a DNS server to send a NOTIFY message 

### Expected behaviour

dnsdist should parse NOTIFY message, append insert the ECS option and forward to the server

### Actual behaviour

drops the packet when a NOTIFY is received with the exception message provided above

### Other information

The issue comes from this check in the code
https://github.com/PowerDNS/pdns/blob/master/pdns/dnsdist-ecs.cc#L587

It seems that the code falsely think that the incoming packet has a EDNS header and fails in the `slowRewriteQueryWithExistingEDNS` function.

slowr:
### Short description

When ECS is enabled for a server we call some functions that are only working if EDNS is present which is not the case for all scenarios. (resolves #10409)

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/pdns/dnsdist-ecs.cc
+++ b/pdns/dnsdist-ecs.cc
@@ -583,27 +583,6 @@ bool handleEDNSClientSubnet(PacketBuffer& packet, const size_t maximumSize, cons
   assert(qnameWireLength <= packet.size());
 
   const struct dnsheader* dh = reinterpret_cast<const struct dnsheader*>(packet.data());
-
-  if (ntohs(dh->ancount) != 0 || ntohs(dh->nscount) != 0 || (ntohs(dh->arcount) != 0 && ntohs(dh->arcount) != 1)) {
-    PacketBuffer newContent;
-    newContent.reserve(packet.size());
-
-    if (!slowRewriteQueryWithExistingEDNS(packet, newContent, ednsAdded, ecsAdded, overrideExisting, newECSOption)) {
-      ednsAdded = false;
-      ecsAdded = false;
-      return false;
-    }
-
-    if (newContent.size() > maximumSize) {
-      ednsAdded = false;
-      ecsAdded = false;
-      return false;
-    }
-
-    packet = std::move(newContent);
-    return true;
-  }
-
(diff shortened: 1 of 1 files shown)


slowr:
Oh, I thought the fix would be straight-forward. Will take a look on the failed unit-tests.

rgacogne:
Hi! Thanks a lot for reporting this issue and for this pull request! I took a different approach in #10419, calling the "slow" function as soon as the query has a non-empty answer or authoritative section, as the "fast" path is only designed to deal with queries that have no answer, authority or additional records except an OPT one. I also changed the condition in the "slow" function so it accepts queries without EDNS and renamed it to make what it does clearer. 
Let me know if you disagree :)

slowr:
hey @rgacogne I had a working solution already but didn't have time to push to this PR. Either way we can go with your solution as it's more clean! I tested with NOTIFY as well and it works fine. I will close this PR in favor of yours. Thanks!

PR #10422, opened 2021-05-18
Rec: backport 10420 to 4.5.x: For the NOD lookup case, we don't want QName Minimization.

omoerbeek:
### Short description

Backport of #10420 

### Checklist

I have:
- [X] compiled this code
- [ ] tested this code
- [ ] documented the code


paddg:
We did a test and it looks good! NODs are now no longer truncated. And together with "dont-throttle-netmasks" for the receiver address, it now works uninterrupted. However, "all-outqueries" increases by 25%. I am not sure if nod-lookups should better be counted separately.

rgacogne (approved):

PR #10424, opened 2021-05-18
auth: replace domain-metadata-cache-ttl by zone-metadata-cache-ttl

mind04:
### Short description
and some domain -> zone in the settings descriptions and settings.rst

### Checklist
I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/pdnsutil.cc
+++ b/pdns/pdnsutil.cc
@@ -135,7 +135,8 @@ static void loadMainConfig(const std::string& configdir)
   ::arg().set("default-soa-content","Default SOA content")="a.misconfigured.dns.server.invalid hostmaster.@ 0 10800 3600 604800 3600";
   ::arg().set("chroot","Switch to this chroot jail")="";
   ::arg().set("dnssec-key-cache-ttl","Seconds to cache DNSSEC keys from the database")="30";
-  ::arg().set("domain-metadata-cache-ttl","Seconds to cache domain metadata from the database")="60";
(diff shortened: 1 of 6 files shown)


klaus-nicat:
Basically this makes sense. But if you go this way you also have to change the table name - TTL setting and table name IMO belong together. Or you will end up with lots renaming and have not renamed all (then better do not rename at all).

Habbie:
> Basically this makes sense. But if you go this way you also have to change the table name - TTL setting and table name IMO belong together. Or you will end up with lots renaming and have not renamed all (then better do not rename at all).

We cannot fix everything at once. The API, for example, has always called it zone metadata. We will get there eventually.

pdns/pdnsutil.cc
```diff
@@ -3673,7 +3675,8 @@ try
       std::map<std::string, std::vector<std::string> > meta;
       if (src->getAllDomainMetadata(di.zone, meta)) {
         for (const auto& i : meta) {
-          if (!tgt->setDomainMetadata(di.zone, i.first, i.second)) throw PDNSException("Failed to feed domain metadata");
+          if (!tgt->setDomainMetadata(di.zone, i.first, i.second))
+            throw PDNSException("Failed to feed zones metadata");
```
Habbie, line 3679:
```suggestion
            throw PDNSException("Failed to feed zone metadata");
```

Habbie (changes requested):
one change requested, otherwise good!

Habbie (approved):

PR #10426, opened 2021-05-19
rec: When refreshing, do not consider root almost expired 

omoerbeek:
When SyncREs is running in refresh-almost-expired mode, we should
not consider NS and their A/AAAA records to be almost expired, that
can cause refetch propagation all the way up to the root and marking
of nameservers as non-resolving.

Issue reported on IRC by ph1.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/recursordist/test-syncres_cc2.cc
+++ b/pdns/recursordist/test-syncres_cc2.cc
@@ -1845,7 +1845,7 @@ BOOST_AUTO_TEST_CASE(test_cache_almost_expired_ttl)
   BOOST_REQUIRE(ret[0].d_type == QType::NS);
   BOOST_CHECK_EQUAL(getRR<NSRecordContent>(ret[0])->getNS(), DNSName("pdns-public-ns1.powerdns.com."));
   ttl = ret[0].d_ttl - now;
-  BOOST_CHECK_EQUAL(ttl, 86400U);
+  BOOST_CHECK_EQUAL(ttl, std::min(SyncRes::s_maxcachettl, 172800U));
 
   // ATM we are not testing the almost expiry of root infra records, it would require quite some cache massage...
 }
--- a/pdns/recursordist/test-syncres_cc2.cc
+++ b/pdns/recursordist/test-syncres_cc2.cc
@@ -2,6 +2,7 @@
 #include <boost/test/unit_test.hpp>
 
 #include "test-syncres_cc.hh"
+#include "rec-taskqueue.hh"
 
 BOOST_AUTO_TEST_SUITE(syncres_cc2)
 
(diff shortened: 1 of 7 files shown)


rgacogne:
> that can cause refetch propagation all the way up to the root and marking of nameservers as non-resolving.

I'm not sure I understand exactly what is happening to cause some servers to be marked as non-resolving? During the review of the initial PR we stated that "both RC entries needed for resolving a name and the name itself end up in the queue (if considered almost expired). The queue is processed in order, so e.g. the NS records needed will be refreshed before the final name and won't be re-refreshed during the resolving of the final name: after they are refreshed that are not amost expired anymore". Is there cases where this does not hold? Or is this a different issue?

omoerbeek:
Will revisit this,

omoerbeek:
The way to reproduce is:
```
./pdns_recursor --threads=1 --dnssec=off --socket-dir=/tmp/rec --local-port=5301 --quiet=yes --refresh-on-ttl-perc=50 --disable-packetcache=yes --loglevel=9 --qname-minimization=yes --max-cache-ttl=60
``
and then run
watch `dig -p 5301 A ns1.google.com  ns2.google.com ns3.google.com ns4.google.com`

After 30s I'm seeing tasks being pushed for ns1...ns4.google.com. execution of these tasks fail (I added some debug code when a taks is submitted and print the result of the task):
```
Jun 02 15:06:49 TaskQueue: pushed ns1.google.com.|A
Jun 02 15:06:49 TaskQueue: resolving ns1.google.com.|A
Jun 02 15:06:49 TaskQueue: pushed ns2.google.com.|A
Jun 02 15:06:49 TaskQueue: resolving ns2.google.com.|A
Jun 02 15:06:49 TaskQueue: pushed ns3.google.com.|A
Jun 02 15:06:49 TaskQueue: resolving ns3.google.com.|A
Jun 02 15:06:49 TaskQueue: pushed ns4.google.com.|A
Jun 02 15:06:49 TaskQueue: resolving ns4.google.com.|A
Jun 02 15:06:49 TaskQueue: res = 2 ns1.google.com.|A
Jun 02 15:06:49 TaskQueue: no result ns1.google.com.|A
Jun 02 15:06:49 TaskQueue: res = 2 ns3.google.com.|A
Jun 02 15:06:49 TaskQueue: no result ns3.google.com.|A
Jun 02 15:06:49 TaskQueue: res = 2 ns2.google.com.|A
Jun 02 15:06:49 TaskQueue: no result ns2.google.com.|A
Jun 02 15:06:49 TaskQueue: res = 2 ns4.google.com.|A
Jun 02 15:06:49 TaskQueue: no result ns4.google.com.|A
```
The task resolve the given name in refetch mode: names that are "almost expired" are considered expired. 
So the NS set of google.com is considered expired but chain of delegations to it as well, because of the TTL limit of 60s and the 50% almost expired condition.

Now a question indeed is why the parent NS records where not submitted. Investigating further...

**Comment by omoerbeek** [comment]
So the NS lookup task is not triggered because an "almost expired" record is still considered valid (only as a side effect a new task is submitted). During the resolve of that task (in refetch mode) all infra records needed are considered expired and trigger an immediate external lookup, up to and including the root. Now testing a patch to never consider `.` records expired.

**Comment by rgacogne** [comment]
Thanks for the explanation! I understand and agree with the current fix. Perhaps a unit test would be nice?

I'm wondering if it would not make sense to not consider all "almost expired" records as expired when in refresh mode, but instead only the one we are currently refreshing? It would limit the "snowball effect" where refreshing one record leads to refreshing a lot of infra records needed to resolve it, while these are still valid. Of course these infra records would then not be added to the refresh queue so we would have to get them again the next time, so perhaps there is not much of a real gain.

**Review by rgacogne — APPROVED** [review]

**Comment by omoerbeek** [comment]
> Thanks for the explanation! I understand and agree with the current fix. Perhaps a unit test would be nice?
> 
> I'm wondering if it would not make sense to not consider all "almost expired" records as expired when in refresh mode, but instead only the one we are currently refreshing? It would limit the "snowball effect" where refreshing one record leads to refreshing a lot of infra records needed to resolve it, while these are still valid. Of course these infra records would then not be added to the refresh queue so we would have to get them again the next time, so perhaps there is not much of a real gain.

Yes, that is my assumption as well. I'll add a test and then merge.

File: pdns/recursordist/test-syncres_cc2.cc [hunk 1]
```diff
@@ -1769,4 +1770,84 @@ BOOST_AUTO_TEST_CASE(test_cache_expired_ttl)
...
+  BOOST_CHECK_EQUAL(ttl, 60U);
+
+  // Also check if NS record was updated
+  ret.clear();
+  BOOST_REQUIRE_GT(g_recCache->get(now, target, QType(QType::NS), false, &ret, ComboAddress()), 0);
+  BOOST_REQUIRE_EQUAL(ret.size(), 1U);
+  BOOST_REQUIRE(ret[0].d_type == QType::NS);
+  BOOST_CHECK_EQUAL(getRR<NSRecordContent>(ret[0])->getNS(), DNSName("pdns-public-ns1.powerdns.com."));
+  ttl = ret[0].d_ttl - now;
+  BOOST_CHECK_EQUAL(ttl, 86400U);
```
rgacogne, line 1848: Perhaps we could make it clear where the TTL comes from, something like: 
```suggestion
  BOOST_CHECK_EQUAL(ttl, std::min(SyncRes::s_maxcachettl, 172800U));
```

rgacogne (approved):
Looks good, one suggestion.

PR #10427, opened 2021-05-19
rec: Backport 10426 to rec-4.5.x: Be less aggressive forcing refresh of almost-expired NS and their A/AAAA records

omoerbeek:
Backport of #10426

Slightly different approach, since 4.5.x does not do explicit cache calls in `getAddr()`.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code


omoerbeek:
Obsolete, since #10426 now has a better fix.

PR #10428, opened 2021-05-19
Rec: cleanup of outgoing TCP code and DoT to auth or forwarders

omoerbeek:
### Short description

If a forwarder specifies port 853, DoT wil be done to the target (without cert validation atm). This mechanism can be swicthed off through a setting (`dot-to-port-853`).

<strike>There are some things in the TCP handlers that could be improved as they seem to be more complex than needed, but it seems to work.</strike>

Not all tasks below are done, but it would be nice to get a review at this point. In the meantime I can think which method to use to decide to do DoT to auth or not.

I decided to go for a fixed configurable list of names/suffixes.

This PR is already too big. Anything more fancy or advanced should be done in a separate PR.

TODO
- [x] logging should be parametrized
- [x] time-out unit -> milliseconds
- [X] when to TLS to forwarders? 
- [x] when to TLS to auth?
- [X] metrics
- [x] SNMP metrics plus Prometheus annotations
- [x] tests
- [x] docs

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [x] documented the code


pdns/pdns_recursor.cc
```diff
@@ -396,114 +396,90 @@ static bool isHandlerThread()
...
-    t_fdm->removeWriteFD(sock->getHandle());
-    return LWResult::Result::PermanentError;
-  }
-  else if (data.empty()) {// error, EOF or other
-    return LWResult::Result::PermanentError;
+  TCPLOG("arecvtcp called " << len << ' ' << data.size() << endl);
+  size_t pos = 0;
+  data.resize(len);
+  TCPLOG("calling tryRead() " << data.size() << ' ' << len << endl);
+  /* IOState state = */ handler->tryRead(data, pos, len);
```
rgacogne, line 448: `tryRead()` might throw, for example if the connection has been closed by the remote host, it doesn't look like we handle that?

rgacogne: How does this code handle the fact that we might need to write to be able to read, since it does not look at the `IOState`?

rgacogne (commented):
Just a few comments after a partial review, I'll do a full review later, hopefully today.

pdns/pdns_recursor.cc
omoerbeek: If tryRead does not return an answer TCPIOHandlerReadbale function. A needRead/needWrite return will be handled there. 

But as I say in the PR intro and also in the `"TCPIOHandlerReadable more? flip to write seems needed???..."` case, I'm not 100% happy they way this is structured, but I fail to see a better way atm.

Maybe the whole approach should be changed to better match the way tryRead/tryWrite work.

rgacogne: Doesn't `TCPIOHandlerReadable` get called when the underlying descriptor gets readable? And if so, what happens if it doesn't because we need to write first?

omoerbeek: All I know is that if I leave out the initial `tryRead()` call in `arecvtcp()` my TLS connection does not work. With (fixed) TCPLOG enabled:
```
...
arecvtcp 1 2 arecvtcp success
arecvtcp called 525 525
arecvtcp 0 0 timeout
```
Maybe what is happening is that the data is already in (because TLS does larger data chunks underneath)? So the fd does not become readable again, but the data is there and we just need to collect it? That is more or less what I'm doing with that initial `tryRead()` call.

rgacogne: Yes, it's likely that you need to read first, and then only wait for the descriptor to become readable if you could not read. That's indeed because the data has already been processed at the TLS layer.
But be careful that sometimes reading will require sending data, and 99.9% of the time the socket will be writable so you will not see it, but if the send buffer is full then `tryRead()` will return `IOState::NeedWrite` and you'll need to wait for the descriptor to become writable because it won't become readable otherwise (well, the connection will be closed eventually but that's not what you want).

omoerbeek: Ok, then what I need to add is a check on the state returnd by the initial `tryWrite()`, if it is `needWrite`, I need to do `addWriteFD` with TCPIOHandlerWritable.

rgacogne: Actually I think you need to call `addWriteFD` but with `TCPIOHandlerReadable` as the callback since you really want to read, from your point of view (but be careful about the need to update the mplexer later, though) :)

omoerbeek: Also, be aware that I only tested TLS on OpenBSD which uses the `pollmultiplexer` so far. There might be subtle differences in the way the multiplexers work...

omoerbeek: > Actually I think you need to call `addWriteFD` but with `TCPIOHandlerReadable` as the callback since you really want to read, from your point of view (but be careful about the need to update the mplexer later, though) :)

I'm a bit sad that we are forced to either check for either readable or writeable with the current API, while the underlying mechanisms *are* capable of doing both simultaneously...

omoerbeek: > Hmm, at least in dnsdist we don't really have cases where we want to be notified when the descriptor is either writable OR readable, but we could update the multiplexer code if the recursor needs that.

I don't know yet if we really need it, but atm I think it would make the code easier. But that is just a hunch, no hard data.

omoerbeek:
OK, processes some of the comments. I must say that I'm still not seeing things 100% clear about the mutliplexer. I'll try to explain:

What does NeedRead/NeedWrite mean:  I'm assuming you need to wait until the fd becomes readable/writeable and the call tryRead/tryWrite. 

But if I do `addWriteFD` but with `TCPIOHandlerReadable` as handler we will wait until the fd becomes writable and then call `tryRead` ???

From the OpenSSL SSL_read/write manpage:

     If the underlying BIO is non-blocking, a read function will also return
     when the underlying BIO could not satisfy the needs of the function to
     continue the operation.  In this case a call to SSL_get_error(3) with the
     return value of the read function will yield SSL_ERROR_WANT_READ or
     SSL_ERROR_WANT_WRITE.  As at any time a re-negotiation is possible, a
     read function may also cause write operations.  The calling process must
     then repeat the call after taking appropriate action to satisfy the needs
     of the read function.  The action depends on the underlying BIO.  When
     using a non-blocking socket, nothing is to be done, but select(2) can be
     used to check for the required condition.  When using a buffering BIO,
     like a BIO pair, data must be written into or retrieved out of the BIO
     before being able to continue.

The "When using a non-blocking socket, nothing is to be done..." part makes me ponder that using a R or W condition in the poll can be handy: poll for both conditions, and if any of the conditions arises, do the appropriate action.

I'm going to see if the code works with epoll (on Linux) now and I'll also study the dnsdist code, maybe that gives me a clearer picture.

rgacogne:
> What does `NeedRead`/`NeedWrite` mean: I'm assuming you need to wait until the fd becomes readable/writeable and the call `tryRead`/`tryWrite`.

Yes, `NeedRead` means "wait until the underlying descriptor becomes readable then retry the operation you were attempting to perform" (it could be `tryRead()` but it could also be `tryWrite()`), and `NeedWrite` means "wait until the underlying descriptor becomes writable then retry the operation you were attempting to perform".

> But if I do `addWriteFD` but with `TCPIOHandlerReadable` as handler we will wait until the fd becomes writable and then call `tryRead` ???

Well, yes. If `tryRead()` returns `NeedWrite` it means the TLS layer needs to write before being able to read (for example because the handshake has not been performed yet) and attempting to write returned `EAGAIN` or `EWOULDBLOCK`, likely because the underlying buffer is full.

> The "When using a non-blocking socket, nothing is to be done..." part makes me ponder that using a R or W condition in the poll can be handy: poll for both conditions, and if any of the conditions arises, do the appropriate action.

I don't interpret it that way, I think it means there is no need to do anything _except waiting for the socket to become readable for `SSL_ERROR_WANT_READ` or writable for `SSL_ERROR_WANT_WRITE`_. But I don't think there is a case where we want to be able to either read or write. Most of the time we will be waiting for the socket to become readable, and it will be pretty much always be writable.

omoerbeek:
A, that explains quite a bit of my confusion. I'm now in the process of rewriting the TCPIOHandler callbacks to be more like the dnsdist code.

omoerbeek:
> I'll add more documentation to the TCPIOHandler code, as you should not have had to figure that out by yourself.

Maybe we should rename the IOStates: they actually say something like `waitForFDReadable` and `waitForFDWritable` instead of `needRead` and `needWrite`. In my mind the latter named referr to a function, and not the state of the fd.

rgacogne:
That's true of the values returned by the `TCPIOHandler`, it's a bit less simple in other parts of dnsdist. It if really is confusing I might have to use a different type in these parts and do conversions between them.

omoerbeek:
@rgacogne Can you take a look at the current state? I think it is better now. I avoid copying PacketID objects by using a shared pointer and I'm using a single handler function.

pdns/lwres.cc
```diff
@@ -341,16 +341,27 @@ LWResult::Result asyncresolve(const ComboAddress& ip, const DNSName& domain, int
...
       s.setNonBlocking();
       localip = pdns::getQueryLocalAddress(ip.sin4.sin_family, 0);
       s.bind(localip);
 
       std::shared_ptr<TLSCtx> tlsCtx{nullptr};
-      TCPIOHandler handler("", s.releaseHandle(), timeout, tlsCtx, now->tv_sec);
-      IOState state = handler.tryConnect(SyncRes::s_tcp_fast_open_connect, ip);
+      if (SyncRes::s_dot_to_port_853 && ip.getPort() == 853) {
+        TLSContextParameters tlsParams;
+        tlsParams.d_provider = "openssl";
```
rgacogne, line 354: Just a note but if you ever want to support GnuTLS, I have fixes for client-side TLS in #10338.

pdns/pdns_recursor.cc
```diff
@@ -396,114 +396,113 @@ static bool isHandlerThread()
...
+    TCPLOG("arcvtcp tryRead() returned " << int(state) << ' ' << pos << '/' << len << endl);
+    switch (state) {
+    case IOState::Done:
+    case IOState::NeedRead:
+      if (pos == len || (incompleteOkay && pos > 0)) {
+        data.resize(pos);
+        TCPLOG("acecvtcp success A" << endl);
+        return LWResult::Result::Success;
+      }
+      break;
```
rgacogne, line 462: How do you manage a partial read without storing `pos`? Or do we need to move the partial result into `inMSG`?

pdns/pdns_recursor.cc
```diff
@@ -396,114 +396,113 @@ static bool isHandlerThread()
...
+  pident->highState = TCPAction::DoingWrite;
+  pident->lowState = IOState::NeedWrite;
 
-LWResult::Result asendtcp(const PacketBuffer& data, TCPIOHandler& handler)
-{
-  PacketID pident;
-  pident.tcphandler = &handler;
-  pident.tcpsock = handler.getDescriptor();
-  pident.outMSG = data;
+  t_fdm->addWriteFD(handler->getDescriptor(), TCPIOHandlerIO, pident);
```
rgacogne, line 418: I'm afraid writing might require a read, especially if the handshake has not been completed :-/

pdns/pdns_recursor.cc
```diff
@@ -396,114 +396,113 @@ static bool isHandlerThread()
...
+    t_fdm->addReadFD(handler->getDescriptor(), TCPIOHandlerIO, pident);
+  }
 
-  int ret = MT->waitEvent(pident, &data, g_networkTimeoutMsec);
+  int ret = MT->waitEvent(*pident, &data, g_networkTimeoutMsec);
+  TCPLOG("arecvtcp " << ret << ' ' << data.size() << ' ' );
   if (ret == 0) {
-    t_fdm->removeReadFD(handler.getDescriptor());
+    TCPLOG("timeout" << endl);
+    t_fdm->removeReadFD(handler->getDescriptor());
```
rgacogne, line 492: What if we were waiting for the ability to write?

pdns/pdns_recursor.cc
```diff
@@ -396,114 +396,113 @@ static bool isHandlerThread()
...
     return LWResult::Result::Timeout;
   }
   else if (ret == -1) {
-    t_fdm->removeWriteFD(handler.getDescriptor());
+    TCPLOG("PermanentError" << endl);
+    t_fdm->removeWriteFD(handler->getDescriptor());
```
rgacogne, line 497: And here if we were waiting for the ability to read?

pdns/pdns_recursor.cc
```diff
@@ -4078,125 +4077,130 @@ static void handleRCC(int fd, FDMultiplexer::funcparam_t& var)
...
+      switch (newstate) {
+      case IOState::Done:
+      case IOState::NeedRead:
+        TCPLOG("tryRead: Done or NeedRead " << int(newstate) << ' ' << pos << '/' << pid->inNeeded << endl);
+        pid->inMSG.resize(pos); // old content (if there) + new bytes read
+        pid->inNeeded -= pos;
+        TCPLOG("TCPIOHandlerIO " << pid->inNeeded << ' ' << pid->inIncompleteOkay << endl);
+        if (pid->inNeeded == 0 || pid->inIncompleteOkay) {
+          newstate = IOState::Done;
+          // removeReadFD seems to clobber PacketID, so take a copy
```
rgacogne, line 4106: That seems weird, shouldn't `PacketID` be opaque to the multiplexer?

pdns/pdns_recursor.cc
```diff
@@ -4078,125 +4077,130 @@ static void handleRCC(int fd, FDMultiplexer::funcparam_t& var)
...
+          PacketID tmp = *pid;
+          t_fdm->removeReadFD(fd);
```
rgacogne, line 4108: What if we were waiting for a writable fd before that operation?

pdns/pdns_recursor.cc
```diff
@@ -4078,125 +4077,130 @@ static void handleRCC(int fd, FDMultiplexer::funcparam_t& var)
...
+      TCPLOG("tryWrite: " << pid->outPos << '/' << pid->outMSG.size() << ' ' << pid << " -> ");
+      TCPLOG(int(pid->outMSG[0]) << ' ' << int(pid->outMSG[1]) << ' ');
+      newstate = pid->tcphandler->tryWrite(pid->outMSG, pid->outPos, pid->outMSG.size());
+      TCPLOG(pid->outPos << '/' << pid->outMSG.size() << endl);
+      switch (newstate) {
+      case IOState::Done: {
+        TCPLOG("tryWrite: Done" << endl);
+        // removeWriteFD seems to clobber PacketID, so take a copy
+        PacketID tmp = *pid;
+        t_fdm->removeWriteFD(fd);
```
rgacogne, line 4138: What if we were waiting for a readable fd before that operation?

pdns/pdns_recursor.cc
```diff
@@ -4078,125 +4077,130 @@ static void handleRCC(int fd, FDMultiplexer::funcparam_t& var)
...
+        break;
+      }
     }
-    else {
-      //      cerr<<"Still have "<<pident->inNeeded<<" left to go"<<endl;
+    catch (const std::exception& e) {
+      newstate = IOState::Done;
+      TCPLOG("read exception..." << e.what() << endl);
+      PacketID tmp = *pid;
+      t_fdm->removeReadFD(fd); // pid might now be invalid (it isn't, but still)
```
rgacogne, line 4120: same

pdns/pdns_recursor.cc
```diff
@@ -4078,125 +4077,130 @@ static void handleRCC(int fd, FDMultiplexer::funcparam_t& var)
...
-  if (ret > 0) {
-    pid->outPos += (ssize_t)ret;
-    if (pid->outPos == pid->outMSG.size()) {
-      PacketID tmp=*pid;
+    catch (const std::exception& e) {
+      newstate = IOState::Done;
+      TCPLOG("write exception..." << e.what() << endl);
+      // removeWriteFD seems to clobber PacketID, so take a copy
+      PacketID tmp = *pid;
       t_fdm->removeWriteFD(fd);
```
rgacogne, line 4155: same

pieterlexis:
jfyi: the auths for powerdns.com have dnsdist with TLS using valid certificates. You could the nicely against that as well.

pdns/pdns_recursor.cc
omoerbeek: I suppose this is a remainder of my earlier confusion. Let me clean up this and the other issues, it *should* lead to cleaner code.

pdns/pdns_recursor.cc
omoerbeek: The latter

pdns/pdns_recursor.cc
omoerbeek: Now I'm confused again... do we need to wait for a readable fd, or do an actual tryRead()? 

I'm asuming the former:

I suppose If the `tryWrite` call is done and returns `NeedRead`, the readable check will be taken care of.
The only case I can see go wrong is that the fd never becomes writeable... so the `tryWrite` is never done.

But if we set `lowState = NeedRead` and call `addReadFD` here, aren't we running the risk that the fd never becomes readable?

rgacogne: Yes, the former. If the `tryWrite` call is done and returns `NeedRead`, we need to wait for the fd to become readable then call `tryWrite`. At this point if the is not writable `tryWrite` will return `NeedWrite` and we will have to wait again, for the fd to become writable this time.
If the fd never becomes readable, something is wrong and the timeout will kick in, I think?

omoerbeek: So my understanding is I should do an explicit tryWrite (without waiting for a condition) to get the flow going, like a mirror of `arcvtcp` that does a `tryRead` first?

rgacogne: Yes! That might even save a round-trip when fast open works properly, for example :-)

omoerbeek:
I think we are slowly getting there... The upcoming squash is going to be epic :-)

I do worry a bit about the order of updating the fd state and calling sendEvent. Is it safe to first call `sendEvent` (telling the upper layer: we're done) and then remove the FD from the set of FDs that are watched?

pdns/pdns_recursor.cc
omoerbeek: Handled, including the other cases.

rgacogne:
> I do worry a bit about the order of updating the fd state and calling sendEvent. Is it safe to first call sendEvent (telling the upper layer: we're done) and then remove the FD from the set of FDs that are watched?

AFAIK the mtasker doesn't know about the multiplexer, so I think it only depends on what we do in the callback?

omoerbeek:
Looking at the test failures...

rgacogne:
> Looking at the test failures...

The dnsdist one seems like a spurious failure, FWIW.

omoerbeek:
```
14:20 < ottom> rgacogne: there does seem to be an order dependency, if sendEvent leads to a close, a later epoll set update fails
14:21 < ottom> so i need to do the fd bookkeeping before I call sendEvent
14:23 < ottom> this is fd bookkeeping followed by sendEvent order is also in the original code
```

omoerbeek:
Rebased

regression-tests.recursor-dnssec/test_SimpleForwardOverDoT.py
```diff
@@ -0,0 +1,51 @@
+import dns
+import os
+import subprocess
+from recursortests import RecursorTest
+
+class testSimpleForwardOverDoT(RecursorTest):
+    """
+    This forwarding to a DoT srever in avery basic way and is dependent on Quad9 working
```
rgacogne, line 8:
```suggestion
    This is forwarding to a DoT server in a very basic way and is dependent on Quad9 working
```

regression-tests.recursor-dnssec/test_SimpleForwardOverDoT.py
```diff
@@ -0,0 +1,51 @@
...
+    """
+
+    _confdir = 'SimpleForwardOverDoT'
+    _config_template = """
+dnssec=validate
+forward-zones-recurse=.=9.9.9.9:853
```
rgacogne, line 14: Would be nice to add a test for `dot-to-auth-names` as well :)

rgacogne (changes requested):

pdns/pdns_recursor.cc
```diff
@@ -4078,125 +4088,146 @@ static void handleRCC(int fd, FDMultiplexer::funcparam_t& var)
...
+      pid->inMSG.resize(pos + pid->inNeeded); // make room for what we'll read
+      newstate = pid->tcphandler->tryRead(pid->inMSG, pos, pid->inNeeded);
+      switch (newstate) {
+      case IOState::Done:
+      case IOState::NeedRead:
+        TCPLOG("tryRead: Done or NeedRead " << int(newstate) << ' ' << pos << '/' << pid->inNeeded << endl);
+        pid->inMSG.resize(pos); // old content (if there) + new bytes read
+        pid->inNeeded -= pos;
+        TCPLOG("TCPIOHandlerIO " << pid->inNeeded << ' ' << pid->inIncompleteOkay << endl);
+        if (pid->inNeeded == 0 || pid->inIncompleteOkay) {
```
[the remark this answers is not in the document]
omoerbeek: I think I botched this when I adapted the code to read immediately into inMSG instead of a local tmp buf.

pdns/pdns_recursor.cc
```diff
@@ -4078,125 +4088,144 @@ static void handleRCC(int fd, FDMultiplexer::funcparam_t& var)
...
+  // In the code below, we want to update the state of the fd before calling sendEvent
+  // a sendEvent might close the fd, and some poll multiplexers do not like to manipulate a closed fd
+
+  switch (pid->highState) {
+  case TCPAction::DoingRead:
+    TCPLOG("highState: Reading" << endl);
+    // In arecvtcp, the buffer was resized already so inWanted bytes will fit
+    // try reading
+    try {
+      newstate = pid->tcphandler->tryRead(pid->inMSG, pid->inPos, pid->inWanted - pid->inPos);
```
rgacogne, line 4167: `tryRead()` does `toRead` - `pos` already.
```suggestion
      newstate = pid->tcphandler->tryRead(pid->inMSG, pid->inPos, pid->inWanted);
```

pdns/pdns_recursor.cc
rgacogne: Are you sure we are not going to try to remove the FD from the "waiting for read" set of the multiplexer if a timeout occurs while we are waiting for the FD to become writable?

pdns/syncres.hh
```diff
@@ -941,8 +945,9 @@ struct PacketID
 
   typedef set<uint16_t > chain_t;
   mutable chain_t chain;
-  TCPIOHandler *tcphandler{nullptr};
-  size_t inNeeded{0}; // if this is set, we'll read until inNeeded bytes are read
+  shared_ptr<TCPIOHandler> tcphandler{nullptr};
+  string::size_type inPos{0};   // how for are we along in the inMSG
```
rgacogne, line 949:
```suggestion
  string::size_type inPos{0};   // how far are we along in the inMSG
```

pdns/pdns_recursor.cc
omoerbeek: Riught, I'm going to fix that by calling `TCPIOHandlerStateChange`

omoerbeek:
Squased (from 36 to 11 commits).

pdns/pdns_recursor.cc
```diff
@@ -396,114 +397,132 @@ static bool isHandlerThread()
...
 
-  return LWResult::Result::Success;
-}
-
-static void TCPIOHandlerReadable(int fd, FDMultiplexer::funcparam_t& var);
+  auto pident = std::make_shared<PacketID>();
+  pident->tcphandler = handler;
+  pident->tcpsock = handler->getDescriptor();
+  // We might have a partial result
+  pident->inMSG = data;
```
rgacogne, line 496: Since we are going to clear it below anyway:
```suggestion
  pident->inMSG = std::move(data);
```

rgacogne (approved):
LGTM!

PR #10430, opened 2021-05-19
auth: use primary and secondary for pdnsutil and pdns_control commands 

mind04:
### Short description
The old versions of the commands are still available for compatibility reasons. The tests are not modified and still use the old version of the commands, to make sure they still work as expected. We should change this at a later date.

Both tools now consistently use 'zone'. Before this  pull it was a mix of 'domain' and 'zone'.

### Checklist
I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dynhandler.cc
+++ b/pdns/dynhandler.cc
@@ -242,13 +242,13 @@ string DLNotifyRetrieveHandler(const vector<string>&parts, Utility::pid_t ppid)
   extern CommunicatorClass Communicator;
   ostringstream os;
   if(parts.size()!=2 && parts.size()!=3)
-    return "syntax: retrieve domain [ip]";
+    return "syntax: retrieve zone [ip]";
 
   DNSName domain;
   try {
     domain = DNSName(parts[1]);
   } catch (...) {
-    return "Failed to parse domain as valid DNS name";
+    return "Failed to parse zone as valid DNS name";
   }
 
   ComboAddress master_ip;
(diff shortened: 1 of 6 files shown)


pdns/dynhandler.cc
```diff
@@ -316,10 +316,10 @@ string DLNotifyHandler(const vector<string>&parts, Utility::pid_t ppid)
   extern CommunicatorClass Communicator;
   UeberBackend B;
   if(parts.size()!=2)
-    return "syntax: notify domain";
+    return "syntax: notify zone";
   if(!::arg().mustDo("primary") && !(::arg().mustDo("secondary") && ::arg().mustDo("secondary-do-renotify")))
-      return "PowerDNS not configured as primary (master), or secondary (slave) with re-notifications";
-  g_log<<Logger::Warning<<"Notification request for domain '"<<parts[1]<<"' received from operator"<<endl;
+    return "PowerDNS not configured as primary (primary), or secondary (secondary) with re-notifications";
```
Habbie, line 321: I left those words between () in on purpose, to fully remove later.

pdns/pdnsutil.cc
```diff
@@ -2526,7 +2528,7 @@ try
   {
     signingServer();
   }
-  else if(cmds.at(0) == "signing-slave")
+  else if(cmds.at(0) == "signing-secondary")
```
Habbie, line 2531: This is kind of wrong but this is dead code we should just remove anyway.

Habbie (changes requested):
Nice! Two notes.

pdns/dynhandler.cc
Habbie: But, let's remove them.

Habbie (approved):

PR #10431, opened 2021-05-19
auth lua records: new option to set the http status code to match in ifurlup function

dmachard:
### Short description

This pull request is a proposal to be more permissive on the default response code of the `ifurlup` lua function and to be able to match a specific http status code other than 200.   In some use case is can be useful to match status code other than 200 OK.
A new option `httpcode` has been introducted in `ifurlup`, the default response code is always 200.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code
- [ ] checked that this code was merged to master


pdns/minicurl.hh
```diff
@@ -35,13 +35,15 @@ public:
...
   static void init();
 
-  MiniCurl(const string& useragent="MiniCurl/0.0");
+  MiniCurl(const string& useragent="MiniCurl/0.0", bool failonerror=true);
   ~MiniCurl();
   MiniCurl& operator=(const MiniCurl&) = delete;
-  std::string getURL(const std::string& str, const ComboAddress* rem=nullptr, const ComboAddress* src=nullptr, int timeout = 2);
+  std::string getURL(const std::string& str, const ComboAddress* rem=nullptr, const ComboAddress* src=nullptr, int timeout = 2, int http_status = 200);
   std::string postURL(const std::string& str, const std::string& postdata, MiniCurlHeaders& headers, int timeout = 2, bool fastopen = false);
+  void unsetFailOnError();
```
Habbie, line 43: This appears to be unimplemented - and perhaps unnecessary?

pdns/lua-record.cc
```diff
@@ -97,15 +97,21 @@ class IsUpOracle
       if (cd.opts.count("useragent")) {
         useragent = cd.opts.at("useragent");
       }
-      MiniCurl mc(useragent);
+      int http_code = 200;
+      if (cd.opts.count("httpcode")) {
+        http_code = std::atoi(cd.opts.at("httpcode").c_str());
```
Habbie, line 102: I think you can use `pdns_stou` here.

Habbie (changes requested):
Two comments, looks good otherwise, thanks!

dmachard:
PR updated according to your review., thanks

pdns/lua-record.cc
```diff
@@ -97,15 +97,21 @@ class IsUpOracle
       if (cd.opts.count("useragent")) {
         useragent = cd.opts.at("useragent");
       }
-      MiniCurl mc(useragent);
+      int http_code = 200;
+      if (cd.opts.count("httpcode")) {
+        http_code = pdns_stou(cd.opts.at("httpcode").c_str());
```
Habbie, line 102: You can lose `c_str()` I believe - `pdns_stou` takes string. I see now that I did not mention that earlier, I just thought I did!

Habbie (changes requested):

pdns/lua-record.cc
dmachard: I made the modification, and also re-made some tests which are OK.

liewebagency:
Hello,

Im running 4 servers, bare metal Dedcated servers ALL are IDENTICAL setup, ive spet 4 months to make it work, even used all the money I had which I should have used to my medicine to prolong my life, I have a terminal liver disease and cancer and has 3 months left it seems. My daughter is going to take over the company but I need to get powerdns with recursors and OUR own resolvers which we have signed a deal with German BND and been approved to have our own.

I rented a pro, he says he was for 20.000 dollars to setup powerdns and powerdns admin. He tried postgresql that didnt work because the sync or replica stops working like with ALL db backend mariadbl, mysql. So I went from 12 months left to 3.

All I need is a way to setup 4 nameservers with 4 resolvers on each server and synched db so it ALWAYS works and a panel that ACTUALLY works with the Powerds, the Powerdns admin that all google shows dont work good, it makes wrong with records and serials and so on.

DO ANYONE have a GOOOOOD guide how to setup what I need ? For free, cause now I am on thin ice. We have 500.000 customers, but it won’t stay that way with 3 month down even with conoid because we changed server farm company from Leaseweb to Hetzner.

So I need is 4 nameservers on each server, the fastest, most secure db backend there is, a panel, dnssec OVER TLS and DOH. From there we are FREE! Not depended on cpanel which launched their trial with ubuntu that dot work at all and dont let you have own resolvers. And replicate db/synched. We run Hirsute on all servers 21.04. So. Do any one have a detailed good guide I can use and which DB which will work and not go bad like all the other times? I Saw MariaDB 16 came out with json support and much things. And a easy way guide to setup Powerds with recursors what I would need is I guess master and 3slaves but that dont work if the master goes down ? And a good monitoring tool over the db?

I would appreciate ALL HELP!

David
Lie Web Agency

Habbie:
> Hello,

Hello David,

a random pull request is not the place for a support question. Support questions are handled in our other online communities: IRC and our mailing lists. Please see https://www.powerdns.com/opensource.html for information about those.

I have removed your comment. Please do not comment further here.

miodvallat:
Sorry for the PR grave digging.

Is there still an interest in this feature? Do you need help to bring the PR up-to-date?

miodvallat:
@dmachard Could you sync your PowerDNS fork `master` branch with upstream?

miodvallat:
Superseded by #15127, sorry for the mess.

PR #10449, opened 2021-05-26
auth: enable zone cache by default

Habbie:
### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/docs/upgrading.rst
+++ b/docs/upgrading.rst
@@ -44,6 +44,10 @@ Changed defaults
 - The default value of the ``max-nsec3-iterations`` option has been changed from ``500`` to ``100``.
 - The default value of the ``timeout`` option for :ref:`ifportup` and :ref:`ifurlup` functions has been changed from ``1`` to ``2`` seconds.
 
+Starting with auth-4.5.0-alpha2:
+
+- The default value of the ``zone-cache-refresh-interval`` option has been changed from ``0`` to ``300``.
+
 4.3.x to 4.4.0
 --------------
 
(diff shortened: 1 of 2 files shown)


mind04 (approved):

franklouwers:
From current docs: https://docs.powerdns.com/authoritative/settings.html#zone-cache-refresh-interval

> If your backends do not respond to unknown or dynamically generated zones, it is suggested to enable consistent-backends and set this option to 60.

If the new default is going to be 300 minutes, am I correct a new zone won't show up for 5 minutes by default? If that's the case, then I would certainly highlight that and make that obvious to the users!

If 300 is the new default, then the text quoted above should be changed from 60 to 300 as well.

Habbie:
> If the new default is going to be 300 minutes, am I correct a new zone won't show up for 5 minutes by default? If that's the case, then I would certainly highlight that and make that obvious to the users!

That depends on how the zone is created. API and autosecondary do update the zone cache. I can add a few more words to the upgrade guide.

We have not decided the default setting for the final 4.5.0 release; we're enabling it now to get some feedback.

> If 300 is the new default, then the text quoted above should be changed from 60 to 300 as well.

Yep, agreed!

zeha:
> I can add a few more words to the upgrade guide.

Thats probably a good idea, also maybe in the docs for `pdnsutil create-zone`?

franklouwers:
@zeha yeah, as `pdnsutil create-zone` will probably bypass the notification system the API and auto-secondary uses

Habbie:
Replaced by #10454

PR #10453, opened 2021-05-26
gsqlbackend: throw when lookup of newly created zone fails

zeha:
### Short description
gsqlbackend returns false instead of throwing when a zone has been inserted, but the id lookup for it fails. That seems inconsistent with surrounding code.

### Checklist

I have:
- [x] compiled this code
- [ ] tested this code
- [ ] documented the code


Habbie:
Closing in favour of #10461.

PR #10454, opened 2021-05-27
auth: add/fix getAllDomains()

mind04:
### Short description
getAllDomains() was missing in some backend and was broken in others.

Closes #10449 (commit included in this pull)

After this pull the following backends are still incompatible with the zone cache:
- bindbackend (hybrid mode)
- ldapbackend
- randombackend

### Checklist
I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsbackend.cc
+++ b/pdns/dnsbackend.cc
@@ -29,7 +29,8 @@
 #include "logger.hh"
 
 #include <sys/types.h>
-#include "pdns/packetcache.hh"
+#include "packetcache.hh"
+#include "auth-zonecache.hh"
 #include "dnspacket.hh"
 #include "dns.hh"
 #include "statbag.hh"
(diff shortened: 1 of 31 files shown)


Habbie:
> After this pull the following backends are still incompatible with the zone cache:

We should document this - and perhaps make those error on startup if the zone cache is enabled.

pdns/dnsbackend.cc
```diff
@@ -297,6 +298,14 @@ bool DNSBackend::getBeforeAndAfterNames(uint32_t id, const DNSName& zonename, co
   return ret;
 }
 
+void DNSBackend::getAllDomains(vector<DomainInfo>* domains, bool include_disabled)
+{
+  if (g_zoneCache.isEnabled()) {
+    g_log << Logger::Error << "One of the backends does not support zone caching. Put a zone-cache-refresh-interval=0 in the config file to disable this cache." << endl;
```
Habbie, line 304:
```suggestion
    g_log << Logger::Error << "One of the backends does not support zone caching. Put zone-cache-refresh-interval=0 in the config file to disable this cache." << endl;
```

Habbie (approved):

zeha (approved):

PR #10461, opened 2021-05-31
auth: simplify createDomain()

mind04:
### Short description
Remove the zone id code from create(Slave)Domain(). This code was adding a lot of complexity to the backends and was also slightly wrong in the gsql backends.
- A zone is now added to the zone-cache after the create transaction is committed (api). 
- The auto secondary code is now verifying the creation of the zone before it is added to the zone-cache.

Closes: #10453

### Checklist
I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/modules/bindbackend/bindbackend2.cc
+++ b/modules/bindbackend/bindbackend2.cc
@@ -1388,7 +1388,7 @@ BB2DomainInfo Bind2Backend::createDomainEntry(const DNSName& domain, const strin
   return bbd;
 }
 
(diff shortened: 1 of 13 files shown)


Habbie (changes requested):
Looks great, one question: I think we should remove `returning id` from `add-domain-key-query` in gpgsqlbackend.cc now, otherwise that query leaves one unconsumed row. This likely is fine because we `reset()` but it's not clean.

Habbie:
> I think we should remove `returning id` from `add-domain-key-query` in gpgsqlbackend.cc now, otherwise that query leaves one unconsumed row.

I meant `insert-zone-query`

Habbie (approved):

PR #10463, opened 2021-06-01
auth LUA: add filterForward function, to limit the scope of createForward[6]

### LINKED ISSUE #9701: auth: Document that Lua's createForward should be restricted, make it easy to do so
- Program: Authoritative
 - Issue type: Feature request

### Short description

Nils Trampel reported (thanks a lot!) that Lua's `createForward()`, when used with wildcards without restricting it to specific subnets, can create a security issue (impersonation, cookie stealing, etc).
We should document that, and add an optional NetmaskGroup (NMG) object to it so it can easily be restricted.

Habbie:
### Short description
Fixes #9701

Ready for review.

### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/docs/lua-records/functions.rst
+++ b/docs/lua-records/functions.rst
@@ -185,11 +185,10 @@ Reverse DNS functions
 ~~~~~~~~~~~~~~~~~~~~~
 
 .. warning::
-  The reverse DNS functions are under active development. **They may**
-  **not be safe for production use.** The syntax of these functions may change at any
-  time.
+  For :func:`createForward` and :func:`createForward6`, we recommend filtering with :func:`filterForward`, to prevent PowerDNS from generating A/AAAA responses to addresses outside of your network.
+  Not limiting responses like this may, in some situations, help attackers with impersonation and attacks like such as cookie stealing.
 
-.. function:: createReverse(format)
+.. function:: createReverse(format, [exceptions])
 
   Used for generating default hostnames from IPv4 wildcard reverse DNS records, e.g. ``*.0.0.127.in-addr.arpa`` 
   
@@ -200,7 +199,8 @@ Reverse DNS functions
   Returns a formatted hostname based on the format string passed.
 
   :param format: A hostname string to format, for example ``%1%.%2%.%3%.%4%.static.example.com``.
-  
+  :param exceptions: An optional table of overrides. For example ``{['10.10.10.10'] = 'quad10.example.com.'}`` would, when generating a name for IP ``10.10.10.10``, return ``quad10.example.com`` instead of something like ``10.10.10.10.example.com``.
+
   **Formatting options:**
 
   - ``%1%`` to ``%4%`` are individual octets
@@ -251,7 +251,7 @@ Reverse DNS functions
     $ dig +short A 127.0.0.5.static.example.com @ns1.example.com
     127.0.0.5
   
-.. function:: createReverse6(format)
+.. function:: createReverse6(format[, exceptions])
 
   Used for generating default hostnames from IPv6 wildcard reverse DNS records, e.g. ``*.1.0.0.2.ip6.arpa``
   
@@ -265,7 +265,8 @@ Reverse DNS functions
   Returns a formatted hostname based on the format string passed.
 
   :param format: A hostname string to format, for example ``%33%.static6.example.com``.
-  
+  :param exceptions: An optional table of overrides. For example ``{['2001:db8::1'] = 'example.example.com.'}`` would, when generating a name for IP ``2001:db8::1``, return ``example.example.com`` instead of something like ``2001--db8.example.com``.
+
   Formatting options:
    
   - ``%1%`` to ``%32%`` are individual characters (nibbles)
@@ -316,6 +317,18 @@ Reverse DNS functions
   
     $ dig +short AAAA 2001-a-b--1.static6.example.com @ns1.example.com
     2001:a:b::1
+
+.. function:: filterForward(address, masks[, fallback])
+
+  Used for limiting the output of :func:`createForward` and :func:`createForward6` to a set of netmasks.
+
+  :param address: A string containing an address, usually taken directly from :func:`createForward: or :func:`createForward6`.
+  :param masks: A NetmaskGroup; any address not matching the NMG will be replaced by the fallback address.
+  :param fallback: A string containing the fallback address. Defaults to ``0.0.0.0`` or ``::``.
+
+  Example::
+
+    *.static4.example.com IN LUA A "filterForward(createForward(), newNMG():addMasks{'192.0.2.0/24', '10.0.0.0/8'})"
 
 Helper functions
 ~~~~~~~~~~~~~~~~
--- a/docs/lua-records/reference/netmask.rst
+++ b/docs/lua-records/reference/netmask.rst
@@ -86,7 +86,7 @@ They can be matched against netmasks objects:
   nmg = newNMG()
   nmg:addMask("127.0.0.0/8")
   nmg:addMasks({"213.244.168.0/24", "130.161.0.0/16"})
-  nmg:addMasks(dofile("bad.ips")) -- contains return {"ip1","ip2"..}
+  nmg:addMasks(dofile("bad-ips.lua")) -- contains return {"ip1","ip2"..}
 
   if nmg:match(dq.remoteaddr) then
     print("Intercepting query from ", dq.remoteaddr)
@@ -94,9 +94,15 @@ They can be matched against netmasks objects:
 
 Prefixing a mask with ``!`` excludes that mask from matching.
 
-.. function:: newNMG() -> NetMaskGroup
+.. function:: newNMG([masks]) -> NetMaskGroup
 
-  Returns a new, empty :class:`NetMaskGroup`.
+  .. versionchanged:: 4.5.0
+    Added the optional ``masks`` parameter.
+
+  Returns a new :class:`NetMaskGroup`.
+  If no masks are passed, the object is empty.
+
+  :param {str} masks: The masks to add.
 
 .. class:: NetMaskGroup
 
--- a/pdns/lua-base4.cc
+++ b/pdns/lua-base4.cc
@@ -171,7 +171,18 @@ void BaseLua4::prepareContext() {
   d_lw->registerToStringFunction(&Netmask::toString);
 
   // NetmaskGroup
-  d_lw->writeFunction("newNMG", []() { return NetmaskGroup(); });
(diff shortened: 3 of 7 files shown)


docs/lua-records/functions.rst
```diff
@@ -185,11 +185,9 @@ Reverse DNS functions
 ~~~~~~~~~~~~~~~~~~~~~
 
 .. warning::
-  The reverse DNS functions are under active development. **They may**
-  **not be safe for production use.** The syntax of these functions may change at any
-  time.
+  For :func:`createForward` and :func:`createForward6`, we recommend filtering with :func:`filterForward`.
```
pieterlexis, line 188: _why_ do we recommend this?

docs/lua-records/functions.rst
```diff
@@ -200,7 +198,8 @@ Reverse DNS functions
   Returns a formatted hostname based on the format string passed.
 
   :param format: A hostname string to format, for example ``%1%.%2%.%3%.%4%.static.example.com``.
-  
+  :param exceptions: An optional table of overrides, for example ``{['10.10.10.10'] = 'quad10.example.com.'}``
```
pieterlexis, line 201: this does not make clear what these overrides do

docs/lua-records/functions.rst
```diff
@@ -317,6 +317,17 @@ Reverse DNS functions
     $ dig +short AAAA 2001-a-b--1.static6.example.com @ns1.example.com
     2001:a:b::1
 
+.. function:: filterForward(address, masks)
+
+  Used for limiting the output of :func:`createForward` and :func:`createForward6` to a set of netmasks.
```
pieterlexis, line 322: are the addresses in the `masks` list allowed or not allowed?

pdns/lua-record.cc
```diff
@@ -729,6 +727,24 @@ static void setupLuaRecords()
       return std::string("unknown");
     });
 
+  lua.writeFunction("filterForward", [](string address, std::vector<std::pair<int, string>> masks) {
+      NetmaskGroup nmg;
+      ComboAddress ca(address);
+
+      for (const auto& mask : masks) {
```
pieterlexis, line 734: it could make sense to store an NMG per LUA record so there's no need to recreate is every lookup.

pdns/lua-record.cc
```diff
@@ -729,6 +727,24 @@ static void setupLuaRecords()
...
+        nmg.addMask(mask.second);
+      }
+      if (nmg.match(ComboAddress(address))) {
+        return address;
+      } else {
+        if (ca.isIPv4()) {
+          return string("0.0.0.0");
```
pieterlexis, line 741: Would it make sense to have an option to `filterForward` for setting the fallback addresses?

pieterlexis (changes requested):
Two comments on the code itself, several on the docs. Nothing serious.

pdns/lua-record.cc
Habbie: `filterForward` now takes an NMG instead of a table, which makes caching possible (with some extra typing on the user's end)

pdns/lua-record.cc
Habbie: pushed

docs/lua-records/functions.rst
Habbie: I updated the param text to `  :param masks: A NetmaskGroup; any address not matching the NMG will be replaced by the fallback address.`

docs/lua-records/functions.rst
Habbie: I added more words.

docs/lua-records/functions.rst
Habbie: added a sentence

pdns/lua-base4.cc
```diff
@@ -172,8 +172,8 @@ void BaseLua4::prepareContext() {
 
   // NetmaskGroup
   d_lw->writeFunction("newNMG", []() { return NetmaskGroup(); });
-  d_lw->registerFunction<void(NetmaskGroup::*)(const std::string&mask)>("addMask", [](NetmaskGroup&nmg, const std::string& mask) { nmg.addMask(mask); });
-  d_lw->registerFunction<void(NetmaskGroup::*)(const vector<pair<unsigned int, std::string>>&)>("addMasks", [](NetmaskGroup&nmg, const vector<pair<unsigned int, std::string>>& masks) { for(const auto& mask: masks) { nmg.addMask(mask.second); } });
+  d_lw->registerFunction<NetmaskGroup(NetmaskGroup::*)(const std::string&mask)>("addMask", [](NetmaskGroup&nmg, const std::string& mask) { nmg.addMask(mask); return nmg; });
```
pieterlexis, line 175: I think this also affects dnsdist and the recursor? So those docs need to be updated as well.

docs/lua-records/reference/netmask.rst
```diff
@@ -104,13 +104,13 @@ Prefixing a mask with ``!`` excludes that mask from matching.
...
 
-      Adds ``mask`` to the NetMaskGroup.
+      Adds ``mask`` to the NetMaskGroup and returns the NMG.
 
       :param str mask: The mask to add.
 
   .. method:: NetMaskGroup:addMasks(masks)
 
-      Adds ``masks`` to the NetMaskGroup.
+      Adds ``masks`` to the NetMaskGroup and returns the NMG.
```
pieterlexis, line 113:
``` rst
.. versionchanged: 4.5.0
  ``addMask`` returns the ``NetMaskGroup`` object
```

pieterlexis (approved):
Code looks good imho. two comments

pdns/lua-base4.cc
rgacogne: It does not affect dnsdist (it would be nice if it would, but that's clearly out of scope for this PR ;)).

rgacogne: Do we intend on returning the altered NMG or a copy of the altered NMG?

docs/lua-records/functions.rst
```diff
@@ -317,6 +318,18 @@ Reverse DNS functions
     $ dig +short AAAA 2001-a-b--1.static6.example.com @ns1.example.com
     2001:a:b::1
 
+.. function:: filterForward(address, masks[, fallback])
+
+  Used for limiting the output of :func:`createForward` and :func:`createForward6` to a set of netmasks.
+
+  :param address: A string containing an address, usually taken directly from :func:`createForward: or :func:`createForward6`.
+  :param masks: A NetmaskGroup; any address not matching the NMG will be replaced by the fallback address.
+  :param fallback: A string containing the fallback address. Defaults to ``0.0.0.0`` or ``::``.
```
rgacogne, line 327: I wonder if a different fallback for v6 would make sense?

pieterlexis: I now also wonder if we should produce a NODATA or NXDomain when no fallback is set.. But I think LUA records can't do that?

Habbie: I did not invent the fallbacks in this PR, to be clear - they were already there for names that could not be parsed into addresses. That said, I'm open to suggestions.

We can generate a malformed NODATA by returning an empty table, but DNSSEC hates it.

pdns/lua-base4.cc
Habbie: The intention is to keep the one object that newNMG made. I see I lost a `&` on the left hand side somewhere in the process, I'll check what is happening.

docs/lua-records/functions.rst
rgacogne: Right, so perhaps we would better leave it as it is right now?

Habbie: I'd like that for this PR, yes, unless we have an excellent idea for a better address :)

Habbie: > I now also wonder if we should produce a NODATA or NXDomain when no fallback is set.. But I think LUA records can't do that?

There's a CNAME chaining trick for NXDOMAIN, and I -think- `filterForward` could even be used in that, but I'd like to give that more thought outside of this PR.

pdns/lua-base4.cc
Habbie: Turns out we add the mask, and then we copy the NMG. I'll fix it.

Habbie: I couldn't make LuaWrapper pass the reference around, so I've undone my changes, and added an optional list of masks to `newNMG` as a parameter. I also synced the recursor docs.

docs/lua-records/reference/netmask.rst
Habbie: No longer true.

Habbie:
All comments resolved. I changed the NMG interface back, but added an optional initialiser param to `newNMG()`.

pdns/recursordist/docs/lua-scripting/netmask.rst
```diff
@@ -94,9 +94,12 @@ They can be matched against netmasks objects:
 
 Prefixing a mask with ``!`` excludes that mask from matching.
 
-.. function:: newNMG() -> NetMaskGroup
+.. function:: newNMG([masks]) -> NetMaskGroup
 
-  Returns a new, empty :class:`NetMaskGroup`.
+  Returns a new :class:`NetMaskGroup`.
```
pieterlexis, line 99: Please add a `versionchanged` that the masks are supported from 4.6 and up (as this is recursor).

docs/lua-records/reference/netmask.rst
```diff
@@ -86,31 +86,34 @@ They can be matched against netmasks objects:
...
+  nmg:addMasks(dofile("bad-ips.lua")) -- contains return {"ip1","ip2"..}
 
   if nmg:match(dq.remoteaddr) then
     print("Intercepting query from ", dq.remoteaddr)
   end
 
 Prefixing a mask with ``!`` excludes that mask from matching.
 
-.. function:: newNMG() -> NetMaskGroup
+.. function:: newNMG([masks]) -> NetMaskGroup
```
pieterlexis, line 97: Please add a `versionchanged` that masks are supported since auth 4.5.0

pieterlexis (approved):

Habbie:
Squashed.

pdns/lua-base4.cc
rgacogne: I remember having to switch to a raw pointer instead of a reference for that reason, a long time ago, for some of dnsdist bindings.
I like the optional list of masks even better, so that's great :-)

rgacogne (approved):

Habbie:
Rebased after #10470

PR #10464, opened 2021-06-02
pdnsutil: add run-lua-record-script command

Habbie:
### Short description
needs docs; comments on the interface welcom

### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/Makefile.am
+++ b/pdns/Makefile.am
@@ -290,11 +290,6 @@ pdns_server_LDADD = \
 	$(LIBCRYPTO_LIBS) \
 	$(SYSTEMD_LIBS)
 
-if HAVE_LUA_RECORDS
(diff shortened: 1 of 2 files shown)


Habbie:
I did not do ECS yet.

Perhaps there should be a reasonable set of defaults so most users get away with a much smaller setup script. Perhaps, with that set of defaults, some of the arguments could become positional.

Habbie:
DNSPacket lacks setters for ECS, but for most use cases, the ability to set `who` covers all needs.

I pushed an interface rework (see commit message). It's better but I'm not happy yet.

Habbie:
rebased, fixed usage message

Habbie:
Not sure we really should have `zoneId` in there.

PR #10483, opened 2021-06-08
Implement log methods for QType and QClass 

### LINKED ISSUE #7389: `QClass` needs a `toString` method
 - Issue type: Feature request

### Description
In order to log the QClass ('IN', 'CH', 'HS'), it would be nice to have a `toString()` function on the `QClass` object, as well as having an `operator<<(QClass)` for the `Logger` that uses this function.

Once those are there, replacing all the logging of the QClass to use this would be handy as well.

omoerbeek:
### Short description

Plus reformat logger.hh and .cc

Fixes #7389 fully at last...

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/.not-formatted
+++ b/.not-formatted
@@ -197,8 +197,6 @@
 ./pdns/libssl.cc
 ./pdns/libssl.hh
 ./pdns/lock.hh
-./pdns/logger.cc
-./pdns/logger.hh
 ./pdns/lua-auth4.cc
 ./pdns/lua-auth4.hh
 ./pdns/lua-base4.cc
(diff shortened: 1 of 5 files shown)


pieterlexis (approved):
I'm not a fan of the whitespace fixes in the first commit, but it is ok. Code lgtm. Now we need to replace all usage.

omoerbeek:
> I'm not a fan of the whitespace fixes in the first commit, but it is ok. Code lgtm. Now we need to replace all usage.

Yeah, that should have been a separate commit.

PR #10501, opened 2021-06-16
dnsdist: Add support for Lua per-thread FFI rules and actions

rgacogne:
### Short description

This allows Lua FFI rules and actions that don't need access to the global, shared Lua state to be lock-less, avoiding lock contention.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdist-lua-actions.cc
+++ b/pdns/dnsdist-lua-actions.cc
@@ -549,13 +549,13 @@ private:
     func_t d_func;
     bool d_initialized{false};
   };
-  static uint64_t s_functionsCounter;
+  static std::atomic<uint64_t> s_functionsCounter;
   static thread_local std::map<uint64_t, PerThreadState> t_perThreadStates;
   std::string d_functionCode;
   uint64_t d_functionID;
 };
 
-uint64_t LuaFFIPerThreadAction::s_functionsCounter = 0;
+std::atomic<uint64_t> LuaFFIPerThreadAction::s_functionsCounter = 0;
 thread_local std::map<uint64_t, LuaFFIPerThreadAction::PerThreadState> LuaFFIPerThreadAction::t_perThreadStates;
 
 class LuaFFIResponseAction: public DNSResponseAction
@@ -671,13 +671,13 @@ private:
     bool d_initialized{false};
   };
 
-  static uint64_t s_functionsCounter;
+  static std::atomic<uint64_t> s_functionsCounter;
   static thread_local std::map<uint64_t, PerThreadState> t_perThreadStates;
   std::string d_functionCode;
   uint64_t d_functionID;
 };
 
-uint64_t LuaFFIPerThreadResponseAction::s_functionsCounter = 0;
+std::atomic<uint64_t> LuaFFIPerThreadResponseAction::s_functionsCounter = 0;
 thread_local std::map<uint64_t, LuaFFIPerThreadResponseAction::PerThreadState> LuaFFIPerThreadResponseAction::t_perThreadStates;
 
 thread_local std::default_random_engine SpoofAction::t_randomEngine;
--- a/pdns/dnsdist-lua-actions.cc
+++ b/pdns/dnsdist-lua-actions.cc
@@ -504,8 +504,15 @@ public:
       auto& state = t_perThreadStates[d_functionID];
       if (!state.d_initialized) {
         setupLuaFFIPerThreadContext(state.d_luaContext);
-        state.d_func = state.d_luaContext.executeCode<func_t>(d_functionCode);
+        /* mark the state as initialized first so if there is a syntax error
+           we only try to execute the code once */
         state.d_initialized = true;
+        state.d_func = state.d_luaContext.executeCode<func_t>(d_functionCode);
+      }
+
+      if (!state.d_func) {
+        /* the function was not properly initialized */
+        return DNSAction::Action::None;
       }
 
       dnsdist_ffi_dnsquestion_t dqffi(dq);
@@ -520,9 +527,11 @@ public:
         }
       }
       return static_cast<DNSAction::Action>(ret);
-    } catch (const std::exception &e) {
+    }
+    catch (const std::exception &e) {
       warnlog("LuaFFIPerThreadAction failed inside Lua, returning ServFail: %s", e.what());
-    } catch (...) {
+    }
+    catch (...) {
       warnlog("LuaFFIPerthreadAction failed inside Lua, returning ServFail: [unknown exception]");
     }
     return DNSAction::Action::ServFail;
@@ -616,8 +625,15 @@ public:
       auto& state = t_perThreadStates[d_functionID];
       if (!state.d_initialized) {
         setupLuaFFIPerThreadContext(state.d_luaContext);
-        state.d_func = state.d_luaContext.executeCode<func_t>(d_functionCode);
+        /* mark the state as initialized first so if there is a syntax error
+           we only try to execute the code once */
         state.d_initialized = true;
+        state.d_func = state.d_luaContext.executeCode<func_t>(d_functionCode);
+      }
+
+      if (!state.d_func) {
+        /* the function was not properly initialized */
+        return DNSResponseAction::Action::None;
       }
 
       dnsdist_ffi_dnsquestion_t dqffi(dq);
@@ -632,9 +648,11 @@ public:
         }
       }
       return static_cast<DNSResponseAction::Action>(ret);
-    } catch (const std::exception &e) {
+    }
+    catch (const std::exception &e) {
       warnlog("LuaFFIPerThreadResponseAction failed inside Lua, returning ServFail: %s", e.what());
-    } catch (...) {
+    }
+    catch (...) {
       warnlog("LuaFFIPerthreadResponseAction failed inside Lua, returning ServFail: [unknown exception]");
     }
     return DNSResponseAction::Action::ServFail;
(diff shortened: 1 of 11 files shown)


pdns/dnsdistdist/docs/rules-actions.rst
```diff
@@ -498,6 +498,18 @@ These ``DNSRule``\ s be one of the following items:
...
+
+  .. versionadded:: 1.7.0
+
+  Invoke a Lua FFI function that accepts a pointer to a ``dnsdist_ffi_dnsquestion_t`` object, whose bindings are defined in ``dnsdist-lua-ffi.hh``.
+
+  The ``function`` should return true if the query matches, or false otherwise. If the Lua code fails, false is returned.
+
+  The function will be invoked in a per-thread Lua state, without access to the global Lua state.
+
+  :param string function: the name of a Lua function
```
wojas, line 511: The param must be a Lua script string returning a Lua function, not the name of a function.

rgacogne: Good catch, it looks like I c/p that from the description for `LuaFFIRule` and missed it!

rgacogne:
CI failure is a transient network error (couldn't reach quad9).

pdns/dnsdist-lua-actions.cc
```diff
@@ -489,6 +489,65 @@ class LuaFFIAction: public DNSAction
...
+  {
+  }
+
+  DNSAction::Action operator()(DNSQuestion* dq, std::string* ruleresult) const override
+  {
+    try {
+      auto& state = t_perThreadStates[d_functionID];
+      if (!state.d_initialized) {
+        setupLuaFFIPerThreadContext(state.d_luaContext);
+        state.d_func = state.d_luaContext.executeCode<func_t>(d_functionCode);
```
omoerbeek, line 507: if the `executeCode` fails, what's get assigned to `state.d_func` ?

pdns/dnsdist-lua-actions.cc
```diff
@@ -489,6 +489,65 @@ class LuaFFIAction: public DNSAction
...
+  }
+
+private:
+  struct PerThreadState
+  {
+    LuaContext d_luaContext;
+    func_t d_func;
+    bool d_initialized{false};
+  };
+  static uint64_t s_functionsCounter;
```
omoerbeek, line 543: Is this multi-thread safe? I do not know enough about the dnsdist runtime to verify that.

omoerbeek (commented):
Two questions so far, which also apply to `LuaFFIPerThreadResponseAction` and `LuaFFIPerThreadRule`

pdns/dnsdist-lua-actions.cc
rgacogne: It should be, the actions are created one by one during the initial parsing of the configuration, and later the console user needs to hold the `g_luamutex` mutex while inserting a new rule. Still there is no good reason not to make that counter atomic since we don't care about performance, so I'll do that.

pdns/dnsdist-lua-actions.cc
rgacogne: The default value is an empty function, which should yield `std::bad_function_call ` on later invocation. That should not happen since the state would not be marked as initialized so we would retry again and again. That's not very nice but I'm not sure we can do better, except perhaps by marking the state as initialized before calling `executeCode` and checking later that the function is not empty, so we only try to execute it once. Any thoughts?

omoerbeek: I'm torn. Multiple error messages is indeed ugly, I like a single error message better  But: having faulty Lua is a bad thing, and might event warrant a fatal error...

pdns/dnsdist-lua-actions.cc
wojas: Related question: what is the life cycle of the created Lua states? Do I understand correctly that one is created globally for every thread and never destroyed? The reason I am asking is that we may need a way to do cleanup if they get destroyed frequently.

rgacogne: A new state is created every time a new thread processes the rules, and is never destroyed. The number of threads processing rules is not bounded as one might add new frontends (add*Local) or backends (newServer) at runtime, in addition to the ones created at runtime. In most cases the number of threads is quite stable but I guess there might be cases where backends are constantly added and removed at runtime. I'm not sure what to do in that case, though..

wojas: Are threads ever removed? If they are, one good solution would be to push the Lua state to a pool for later reuse by a new thread.

My main concern is that every state may open a new LMDB env with a large mapsize.

rgacogne: Frontend threads are never removed, as we don't support removing a frontend. UDP backend threads do stop when the backend is removed, and thread-local objects should be destroyed automatically when that happens.

wojas: Assuming the Lua state is also destroyed, I wonder if this would also allow Lua to garbage collect all the objects, in which case there should be no issue.

rgacogne: I will test whether this is actually true and will let you know.

rgacogne: I confirmed that the Lua state is destroyed when a thread exits, by allocating a global object in the Lua code passed to `LuaFFIPerThreadRule` and making sure the destructor is called after removing a backend. Note that backend-related threads only get response rules and their associated Lua states, not (query) rules.

pdns/dnsdist-lua-actions.cc
rgacogne: I don't want a fatal error to happen at runtime, so I guess what I will do is log a warning only once.

pdns/dnsdist-lua-actions.cc
wojas: Great, then everything should be fine. Thanks!

rgacogne:
Test failures are unrelated (q9 unreachable from Circle + spell checking in a weird state).

pdns/dnsdistdist/dnsdist-rules.hh
```diff
@@ -1178,6 +1178,62 @@ private:
...
+  {
+    LuaContext d_luaContext;
+    func_t d_func;
+    bool d_initialized{false};
+  };
+
+  static std::atomic<uint64_t> s_functionsCounter;
+  static thread_local std::map<uint64_t, PerThreadState> t_perThreadStates;
+  std::string d_functionCode;
+  uint64_t d_functionID;
```
omoerbeek, line 1234:
```suggestion
  const std::string d_functionCode;
  const uint64_t d_functionID;
```

pdns/dnsdist-lua-actions.cc
```diff
@@ -537,6 +605,81 @@ class LuaFFIResponseAction: public DNSResponseAction
...
+  {
+    LuaContext d_luaContext;
+    func_t d_func;
+    bool d_initialized{false};
+  };
+
+  static std::atomic<uint64_t> s_functionsCounter;
+  static thread_local std::map<uint64_t, PerThreadState> t_perThreadStates;
+  std::string d_functionCode;
+  uint64_t d_functionID;
```
omoerbeek, line 677:
```suggestion
  const std::string d_functionCode;
  const uint64_t d_functionID;
```

pdns/dnsdist-lua-actions.cc
```diff
@@ -489,6 +489,74 @@ class LuaFFIAction: public DNSAction
...
+  struct PerThreadState
+  {
+    LuaContext d_luaContext;
+    func_t d_func;
+    bool d_initialized{false};
+  };
+  static std::atomic<uint64_t> s_functionsCounter;
+  static thread_local std::map<uint64_t, PerThreadState> t_perThreadStates;
+  std::string d_functionCode;
+  uint64_t d_functionID;
```
omoerbeek, line 555:
```suggestion
  const std::string d_functionCode;
  const uint64_t d_functionID;
```

pdns/dnsdistdist/docs/rules-actions.rst
```diff
@@ -498,6 +498,18 @@ These ``DNSRule``\ s be one of the following items:
   :param KeyValueStore kvs: The key value store to query
   :param KeyValueLookupKey lookupKey: The key to use for the lookup
```
omoerbeek, line 500: I was wondering if there are certain illegal combinations of global rules/action and per-thread rules/actions. If so, that should be documented and detected (if possible).

Thinking about it some more, I think any combinations is acceptable, but you get runtime (or silent?) errors if you try to access the global context from a per-thread rule or action. Some explanation or example describing what's in the global context (or more importantly what's not in the per-thread context) would be helpful:

Just an example: `DNSClass.IN`, is that a name I can use in a per-thread context?

omoerbeek (commented):
Looks good, though docs should be more explicit about global vs local context.

pdns/dnsdistdist/docs/rules-actions.rst
rgacogne: Pretty much everything is allowed, as long as you remember that you are in a separate Lua context and not in the global one, as documented below. All constants are available in the per-thread context, and all FFI functions. Objects and their bindings that are not usable in a FFI context (`DNSQuestion`, `DNSDistProtoBufMessage`, `PacketCache`, ...) are not available.
That seems quite obvious to me but I'll add a couple lines to the descriptions.

omoerbeek (approved):

PR #10505, opened 2021-06-18
Add a periodic zones-to-cache function.

omoerbeek:
Code is now completely rewritten using Lua config and a simpler approach to only fill the record cache. The requirement that the records should be sorted in a particular way is also not longer there. During a first scan of the reords, I collect the signature recors and delegation data, so I can use that data on a seconds pass that fills the cache.

Records can be retrieved via axf, http, https and file.

One to do item left:  tests.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [X] documented the code


pdns/syncres.cc
```diff
@@ -4459,3 +4461,161 @@ int SyncRes::getRootNS(struct timeval now, asyncresolve_t asyncCallback, unsigne
...
+          }
+          NegCache::NegCacheEntry ne;
+          ne.authoritySOA = soa;
+          ne.DNSSECRecords.records = v;
+          ne.DNSSECRecords.signatures = sigs;
+          ne.d_name = qname;
+          ne.d_auth = zone;
+          ne.d_ttd = now + last_ttl;
+          ne.d_qtype = qtype;
+          g_negCache->add(ne);
```
rgacogne, line 4571: I don't understand what we are negatively caching? At this point `qname` and `qtype` are still from the previous entry, and having a `NSEC` for a given `qname` is **not** a denial of that name. It's even more complicated for a `NSEC3`, of course.

pdns/syncres.cc
```diff
@@ -4459,3 +4461,161 @@ int SyncRes::getRootNS(struct timeval now, asyncresolve_t asyncCallback, unsigne
...
+            }
+          }
+#endif
+          break;
+        }
+        default:
+          if (!dnssec) {
+            sigsrr.clear();
+          }
+          bool auth = qname == zone;
```
rgacogne, line 4587: That doesn't sound right, but we would probably need to keep track of delegations to know which records are authoritative?

omoerbeek: This indeed a quick hack that works for (most of?) the records as they appear in the root zone. I should have commented on that.

pdns/syncres.cc
omoerbeek: I should have left out this part. It was a misguided attempt.

omoerbeek:
I confused myself, being under the impression the NSEC and NSEC3 records needed to go to the negcache. But that's not true I realized later, but the code remained. Now the code does not treat NSEC and NSEC3 records in a special way anywmorfe. But the auth flag should be revisited, as it is not correct for these.

pdns/syncres.cc
rgacogne: Right, but in that case perhaps we need to document that it only works with a root zone at the moment?

pdns/syncres.cc
```diff
@@ -4459,3 +4461,137 @@ int SyncRes::getRootNS(struct timeval now, asyncresolve_t asyncCallback, unsigne
...
+          }
+          // XXX The lines below are a quick hack that works (?) for the root zone as presented by
+          // https://www.internic.net/domain/root.zone
+          bool auth = qname == zone;
+          if (qtype == QType::DS) {
+            DNSName parent(qname);
+            parent.chopOff();
+            auth = parent.isPartOf(zone);
+          }
+          g_recCache->replace(now, qname, qtype, v, sigsrr,
```
rgacogne, line 4569: Note that we don't insert NSEC3 into the cache (but the root zone only has NSEC anyway) because we can't fetch out.

pdns/syncres.cc
rgacogne: And if we are indeed supporting the root zone only, it might make sense to treat these records as Secure and to insert the NSEC into the NSEC cache?

omoerbeek:
In the end it is th emabition to support arbitray zones and support ZONEMD. For the latter sorting is needed, so I decided to do not do that here. Also, AXFR and file source should be supported.

I see now that this PR is *not* marked *draft*. I really thought is was marked as such. Fixing.

klaus-nicat:
Hi! I would suggest to rename the feature. In 4.5 there is "zone-cache-refresh-interval" for the new zone-cache. Your feature "zones-to-cache" sounds similar but it totally different. Maybe "load-zones-from-url-into-packetcache".

Habbie:
> Hi! I would suggest to rename the feature. In 4.5 there is "zone-cache-refresh-interval" for the new zone-cache. Your feature "zones-to-cache" sounds similar but it totally different. Maybe "load-zones-from-url-into-packetcache".

Note that these are separate products, but it might indeed make sense to not have very similar naming. Thanks!

pdns/rec-lua-conf.cc
```diff
@@ -398,6 +398,55 @@ void loadRecursorLuaConfig(const std::string& fname, luaConfigDelayedThreads& de
...
+          conf.d_refreshPeriod = boost::get<uint32_t>(have["refreshPeriod"]);
+        }
+        if (have.count("retryOnErrorPeriod")) {
+          conf.d_retryOnError = boost::get<uint32_t>(have["retryOnErrorPeriod"]);
+        }
+      }
+
+      delayedThreads.ztcConfigs.push_back(conf);
+    }
+    catch (const std::exception&e ) {
```
rgacogne, line 445:
```suggestion
    catch (const std::exception& e) {
```

pdns/rec-lua-conf.cc
```diff
@@ -398,6 +398,55 @@ void loadRecursorLuaConfig(const std::string& fname, luaConfigDelayedThreads& de
...
+      g_log<<Logger::Error<<"Problem configuring zoneToCache: "<<e.what()<<endl;
```
rgacogne, line 446: Perhaps we could include the zone name?

pdns/recursordist/rec-zonetocache.cc
```diff
@@ -0,0 +1,288 @@
...
+#include <fstream>
+
+struct ZoneData
+{
+  ZoneData(shared_ptr<Logr::Logger>& log) :
+    d_log(log) {}
+
+  std::map<pair<DNSName, QType>, vector<DNSRecord>> d_all;
+  std::map<pair<DNSName, QType>, vector<shared_ptr<RRSIGRecordContent>>> d_sigs;
+  std::set<DNSName> d_delegations;
```
rgacogne, line 47: Perhaps a `SuffixMatchTree` would make more sense? It might actually be slower in some cases, though.

pdns/recursordist/rec-zonetocache.cc
```diff
@@ -0,0 +1,288 @@
...
+  time_t d_now;
+  DNSName d_zone;
+  shared_ptr<Logr::Logger>& d_log;
+
+  bool isRRSetAuth(const DNSName& qname, QType qtype);
```
rgacogne, line 52:
```suggestion
  bool isRRSetAuth(const DNSName& qname, QType qtype) const;
```

pdns/recursordist/rec-zonetocache.cc
```diff
@@ -0,0 +1,288 @@
...
+  void parseDRForCache(DNSRecord& dr);
+  void getByAXFR(const RecZoneToCache::Config&);
+  void ZoneToCache(const RecZoneToCache::Config& config);
+};
+
+bool ZoneData::isRRSetAuth(const DNSName& qname, QType qtype)
```
rgacogne, line 58:
```suggestion
bool ZoneData::isRRSetAuth(const DNSName& qname, QType qtype) const
```

pdns/recursordist/rec-zonetocache.cc
```diff
@@ -0,0 +1,288 @@
...
+
+#include <fstream>
+
+struct ZoneData
+{
+  ZoneData(shared_ptr<Logr::Logger>& log) :
+    d_log(log) {}
+
+  std::map<pair<DNSName, QType>, vector<DNSRecord>> d_all;
+  std::map<pair<DNSName, QType>, vector<shared_ptr<RRSIGRecordContent>>> d_sigs;
```
rgacogne, line 46: Would it make sense to merge these two maps into one?

pdns/rec-lua-conf.cc
```diff
@@ -398,6 +398,55 @@ void loadRecursorLuaConfig(const std::string& fname, luaConfigDelayedThreads& de
...
+      RecZoneToCache::Config conf;
+      conf.d_zone = zoneName;
+      conf.d_method = method;
+      if (srcs.type() == typeid(std::string)) {
+        conf.d_sources.push_back(boost::get<std::string>(srcs));
+      }
+      else {
+        for (const auto& src : boost::get<std::vector<std::pair<int, std::string>>>(srcs)) {
+          conf.d_sources.push_back(src.second);
+        }
```
rgacogne, line 414: It might be nice to check that the list of sources is not empty.

pdns/recursordist/rec-zonetocache.cc
```diff
@@ -0,0 +1,288 @@
...
+      d_log->info("Getting zone by URL");
+      lines = getURL(config);
+    }
+    else if (config.d_method == "file") {
+      d_log->info("Getting zone from file");
+      lines = getLinesFromFile(config);
+    }
+    DNSResourceRecord drr;
+    ZoneParserTNG zpt(lines, d_zone);
+    zpt.setMaxGenerateSteps(0);
```
rgacogne, line 213: It looks like `0` means unlimited, is that what we want?

pdns/recursordist/rec-zonetocache.cc
```diff
@@ -0,0 +1,288 @@
...
+      break;
+    case QType::RRSIG:
+      break;
+    default: {
+      vector<shared_ptr<RRSIGRecordContent>> sigsrr;
+      auto it = d_sigs.find(key);
+      if (it != d_sigs.end()) {
+        for (const auto& sig : it->second) {
+          sigsrr.push_back(sig);
+        }
```
rgacogne, line 237:
```suggestion
        sigsrr = it->second;
```

pdns/recursordist/rec-zonetocache.cc
```diff
@@ -0,0 +1,288 @@
...
+      break;
+    }
+    }
+  }
+}
+
+// Config must be a copy, so call by value!
+void RecZoneToCache::ZoneToCache(RecZoneToCache::Config config, uint64_t configGeneration)
+{
+  setThreadName("pdns-r/ztc/" + config.d_zone);
```
rgacogne, line 254: Not sure what happens if the thread name is too long?

pdns/recursordist/rec-zonetocache.cc
```diff
@@ -0,0 +1,288 @@
...
+    DNSResourceRecord drr;
+    ZoneParserTNG zpt(lines, d_zone);
+    zpt.setMaxGenerateSteps(0);
+
+    while (zpt.get(drr)) {
+      DNSRecord dr(drr);
+      parseDRForCache(dr);
+    }
+  }
+
```
rgacogne, line 220: I know it's a bit cumbersome but it might be nice to check whether the configuration has been updated at this point, since the transfer/loading might have taken quite some time, and a different thread with a more recent configuration might have updated the cache with more recent values in the meantime.

pdns/recursordist/rec-zonetocache.cc
```diff
@@ -0,0 +1,288 @@
...
+  time_t axfrStart = time(nullptr);
+  time_t axfrNow = time(nullptr);
+
+  while (axfr.getChunk(nop, &chunk, (axfrStart + axfrTimeout - axfrNow))) {
+    for (auto& dr : chunk) {
+      parseDRForCache(dr);
+    }
+    axfrNow = time(nullptr);
+    if (axfrNow < axfrStart || axfrNow - axfrStart > axfrTimeout) {
+      throw PDNSException("Total AXFR time for zoneToCache exceeded!");
```
rgacogne, line 144: Is there a particular reason you are using `PDNSException` and not an exception derived from `std::exception` instead? I would prefer less custom exceptions whenever possible :)

pdns/recursordist/rec-zonetocache.cc
```diff
@@ -0,0 +1,288 @@
...
+    }
+  }
+}
+
+static std::vector<std::string> getLinesFromFile(const RecZoneToCache::Config& config)
```
rgacogne, line 149: Feels a bit weird to pass the whole config and not just the file name here.

rgacogne (approved):
A few comments on the actual code but the whole functionality and design seem great! :+1:

pdns/recursordist/rec-zonetocache.cc
omoerbeek: fixed

pdns/recursordist/rec-zonetocache.cc
omoerbeek: right

pdns/recursordist/rec-zonetocache.cc
omoerbeek: It will generate a warning...

pdns/recursordist/rec-zonetocache.cc
omoerbeek: This is a copy-paste from rpzloader

pdns/recursordist/rec-zonetocache.cc
omoerbeek: It's not clear to me if that wild make the code easier to understand. I'll add a comment in the code that it *might* be an improvement.

pdns/recursordist/rec-zonetocache.cc
omoerbeek: It might be indeed. Will add a comment here as well.

omoerbeek:
Squashed and review comments processed

pdns/recursordist/rec-zonetocache.cc
rgacogne: Works for me! I was under the impression that we usually use the same key and access both maps, but I realize that this is not 100% true for RRSIGs.

Issue #10500, opened 2021-06-15
Dnsdist reports metrics about down upstream servers


hhoffstaette:
- Program: dnsdist
 - Issue type: Bug report
   (more an observation of unexpected behaviour; whether it qualifies as bug is a matter of perspective. :)

### Short description
I noticed that server metrics (e.g. for latencies) are still reported even when an ustream server
is in the "down" state. This is confusing to me, and IMHO the values other than server_status
should be dropped from the metrics as long as the server is down. I believe this would be a more
accurate representation of things.
A related followup problem is whether the statistics for the server in question should be
zeroed out when it goes down; otherwise metrics would continue to report the old stale latencies
when the server comes back (even though they would then adjust eventually).

### Environment

 - Operating system: Linux
 - Software version: dnsdist 1.6.0
 - Software source: self-compiled

### Steps to reproduce
1. set an active server to "down" after some use
2. get metrics, see stale latencies

### Expected behaviour
No latency metrics as long as the server is dead 

### Other information
My original problem - Grafana would show "valid" but stale latencies even for a dead server - could
be kind-of-solved with a hack: instead of simply displaying the latency, I multiply the latency value with
server_status (i.e. 0 or 1). This gives me either the value or a 0-measurement as long as the server is
dead (which can take arbitrarily long, of course).
Still, not reporting the stale metrics in the first place seems cleaner to me.
I briefly looked into the code and found the part in dnsdist_web.cc where I could add an
additional if-condition to verify the server state and then suppress the latency metrics - easy enough.
But the longer I thought about this, the more I asked myself what should happen when
the server comes back. Should the latency metrics continue where they left off? Start from 0?

I can gladly make a PR with the necessary check to drop metrics from dead servers, but first wanted to
hear what other people think.
Thanks for dnsdist :)

rgacogne:
> IMHO the values other than server_status should be dropped from the metrics as long as the server is down. I believe this would be a more accurate representation of things.

In theory I agree with you, although I wonder if all the metrics collector would be fine with the metric disappearing. Any idea?

> A related followup problem is whether the statistics for the server in question should be zeroed out when it goes down; otherwise metrics would continue to report the old stale latencies when the server comes back (even though they would then adjust eventually).

That sounds like a good idea. I'm a bit worried about someone using the latency in a custom load-balancing policy and suddenly sending all queries to a server that just came back up because its latency would be zero, but I guess it could be solved by handling zero as a special case, which would not be a bad idea anyway.

[...]

PR #10508, opened 2021-06-19
dnsdist: Do not report metrics about down upstream servers

hhoffstaette:
### Short description
This patch suppresses output of metrics of dead upstream servers, as reported and discussed
in bug #10500. Tested manually by curling /metrics, setting a server to down and verifying
again that the output no longer contains this server's metrics except for dnsdist_server_status.

### Checklist
I have:
- [x] compiled this code
- [x ] tested this code
- [ ] documented the code


hhoffstaette:
The seond part of the behaviour change discussed in #10500 - zeroing stats when a server goes down - will come
in a separate PR.

rgacogne (approved):
Looks good, thanks! I'm wondering if it makes sense to hide all values, or if hiding only the latency would not be better, because the other values might still be useful to investigate what is going on.

hhoffstaette:
> Looks good, thanks! I'm wondering if it makes sense to hide all values, or if hiding only the latency would not be better, because the other values might still be useful to investigate what is going on.

Hmm..not sure either, but if I start clearing stats on setDown() I should know how much to reset. :)
Personally I don't care, but if there's something you want to see preverved across state changes just let me know. Maybe others have some opinion on this too.

rgacogne:
I would prefer to only hide (and automatically clear) the latency for now, since I personally would find it a weird to reset the other metrics.
I would be OK with adding a `DownstreamState::resetMetrics()` binding (or something like that) to clear all the metrics for a backend, but I'd really prefer to keep it separate. I have seen servers going up and down from time to time in production and I would not want to have the metrics reset automatically in that case.

hhoffstaette:
We now only omit latency ofr a down server, as discussed.

hhoffstaette:
Second commit resets the latency on change of availability; tested manually, works fine (not exactly rocket surgery).
I hope hooking into setDown() is enough; I didn't really understand the exact interaction with setAuto().

hhoffstaette:
A pretty picture!

state-change

Violet is a bit slow, so disable it at 19:40 and bring it back up at 19:50. Latency was reset and is slowly ramping up (not too many requests going on so it takes a while), but the general behaviour should be clear.

hhoffstaette:
Looks like I'm missing something wrt, the state management. The latency is reset when I set a server down manually, but not when it fails the liveliness check. I just had a server fail by itself and when it came back the latency resumed with the previous measurement. Apparently there is a second path where a server is marked "down".

rgacogne:
>  found it, "upStatus" is a second state variable that is updated outside of setDown().

We currently have two different variables, `availability` which is either:
- `Up` if `setUp()` has been called, forcing the server in the `up` state, disabling health-checking ;
- `Down` if `setDown()` has been called, forcing the server in the `down` state, disabling health-checking as well ;
- `Auto`, enabling health-checking.

and `upStatus` which is updated by the health-checks and is `true` is the server is `up`, `false` otherwise. The health-checking code directly updates that value, but it might be cleaner to define a method to do that, allowing you to plug yourself into that new method.

hhoffstaette:
Updated second commit to also introduce a new `setUpStatus()` for all `upStatus` updates and also reset the latency there.
- Pulled the cable on my switch
- Waited for liveliness check to kick in: latency gets reset
- Plugged back in: latency ramp up again

image

rgacogne (approved):

PR #10511, opened 2021-06-21
SVCB: on parse error, throw instead of truncate

### LINKED ISSUE #10442: SVCB: malformed values are thrown away instead of causing an error
good:

```
$ pdnsutil --config-dir=. raw-lua-from-content SVCB '1 . alpn=foo'
"\000\001\000\000\001\000\004\003foo"
```

bad:

```
$ pdnsutil --config-dir=. raw-lua-from-content SVCB '1 . alpn=foo\\'
"\000\001\000\000\001\000\000"
```

Habbie:
### Short description
Fixes #10442

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnslabeltext.rl
+++ b/pdns/dnslabeltext.rl
@@ -293,6 +293,10 @@ size_t parseSVCBValueListFromParsedRFC1035CharString(const std::string &in, std:
   (void) dns_text_to_value_list_en_main;
   %% write exec;
 
+  if ( cs < dns_text_to_value_list_first_final ) {
+          throw runtime_error("Unable to parse DNS SVCB value list '"+in+"'");
+  }
+
   return counter;
 }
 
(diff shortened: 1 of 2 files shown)


rgacogne (approved):
Looks good! I wonder if we need that check in `parseRFC1035CharString` too? The other functions in that file all seem to have it.

Habbie:
> Looks good! I wonder if we need that check in `parseRFC1035CharString` too? The other functions in that file all seem to have it.

That's tricky, because that one is often expected to not reach the end of a string. If I add the check there, a lot of tests break.

I did a few manual tests and it looks like that one won't throw away partial data when it runs into something weird, so I believe we're good here.

rgacogne:
Alright, thanks a lot for testing! :+1:

PR #10518, opened 2021-06-23
rec: Make sure that we pass the SOA along the NSEC(3) proof for DS queries

rgacogne:
### Short description

If the client is requesting a DS that does not exist, we need to provide the SOA (+RRSIGs) along with the NSEC(3) proof (+RRSIGs) and we might not have it if we picked up the proof from a delegation, in which case we need to keep on to do the actual DS query.
It used to work before 4.5.0 because the zone cuts determination code was requesting the DS records before doing any resolution, so we would get the denial and the SOA at the very beginning and not replace it on a delegation because we knew the zone was Insecure at that point.
Note that we still want to use the "no SOA denial" for internal zone cuts computation since we don't care about the SOA at that point, and that saves quite some outgoing queries.

Still needs some tests (unit and perhaps regression ones).

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code


omoerbeek (approved):

omoerbeek:
test-recursor-regression failure is an intermittent network issue

PR #10520, opened 2021-06-23
dnsdist: Don't look up the LMDB dbi by name for every query

rgacogne:
### Short description

This yields a ~5% performance improvement for LMDB lookups.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


rgacogne:
Pushed a commit to open the LMDB database in read-only mode since we never update it, that seems to yield a further 10% gain!

PR #10523, opened 2021-06-24
Update slavecommunicator.cc

rudybroersma:
Very teeny tiny fix for a missing space in a log line:

Domain 'exsilia.net' is fresh, but RRSIGs differ on master2a01:1b0:7999:402::29, so DNSSEC is stale, serial is 2021051001
into:
Domain 'exsilia.net' is fresh, but RRSIGs differ on master 2a01:1b0:7999:402::29, so DNSSEC is stale, serial is 2021051001


pieterlexis (approved):
awesome, thanks!

PR #10525, opened 2021-06-24
dnsdist: Add support for range-based lookups into a Key-Value store 

rgacogne:
### Short description

This feature allows doing a range-based lookup (mostly useful for IP addresses), assuming that:
- there is a key for the last element of the range (2001:0db8:ffff:ffff:ffff:ffff:ffff:ffff for 2001:db8::/32)
which contains the first element of the range (2001:0db8:0000:0000:0000:0000:0000:0000) followed by any data in the value
- AND there is no overlapping ranges in the database !!

This requires that the underlying store supports ordered keys, which is true for LMDB but not for CDB, for example.

PS: this PR is based on the work done in #10520 and might require a rebase once it is merged. So you can skip the first two commits on that branch when reviewing :)

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code


pdns/dnsdistdist/dnsdist-kvs.cc
```diff
@@ -30,15 +30,24 @@ std::vector<std::string> KeyValueLookupKeySourceIP::getKeys(const ComboAddress&
...
   if (truncated.isIPv4()) {
     truncated.truncate(d_v4Mask);
-    result.emplace_back(reinterpret_cast<const char*>(&truncated.sin4.sin_addr.s_addr), sizeof(truncated.sin4.sin_addr.s_addr));
+    key.reserve(sizeof(truncated.sin4.sin_addr.s_addr) + (d_includePort ? sizeof(truncated.sin4.sin_port) : 0));
+    key.append(reinterpret_cast<const char*>(&truncated.sin4.sin_addr.s_addr), sizeof(truncated.sin4.sin_addr.s_addr));
   }
   else if (truncated.isIPv6()) {
     truncated.truncate(d_v6Mask);
-    result.emplace_back(reinterpret_cast<const char*>(&truncated.sin6.sin6_addr.s6_addr), sizeof(truncated.sin6.sin6_addr.s6_addr));
+    key.reserve(sizeof(truncated.sin6.sin6_addr.s6_addr) + (d_includePort ? sizeof(truncated.sin4.sin_port) : 0));
```
wojas, line 41: The ports are only useful for IPv4 CG-NAT setups, not for IPv6. And I guess `sin4.sin_port` would need to use `sin6`?

rgacogne: Right, I ponder adding it only for IPv4 but it seemed easier to have the same code path, even though it will probably never be used for IPv6. The structures are designed in such a way that accessing `sin4.sin_port` or `sin6.sin6_port` is equivalent.

pdns/dnsdistdist/dnsdist-kvs.cc
```diff
@@ -114,6 +121,51 @@ bool LMDBKVStore::keyExists(const std::string& key)
...
+    // to be stored with the last value of the range as key
+    // and the first value of the range as data, sometimes
+    // followed by any other content we don't care about
+
+    // retrieve the first key greater or equal to our key
+    int rc = cursor.lower_bound(MDBInVal(key), actualKey, result);
+
+    if (rc == 0) {
+      auto last = actualKey.get<std::string>();
+      if (last.size() != key.size() || key > last) {
```
omoerbeek, line 141: hmmm, do string compares work for abbreviated addresses? Likely not, so it should be documented that this only works for fully written out v6 addresses.

Likewise I do not think string compare works for IPv4 in their normal presentation form: take for example `200.1.1.1` which sorts before `3.1.1.1` or `1.1.1.200` which sort before `1.1.1.3`
We probably should document that v4 string should be represented with leading zeroes (though watch out: that could lead to octets being interpreted as octal in some contexts).

omoerbeek (changes requested):

pdns/dnsdist-lua-actions.cc
```diff
@@ -1461,6 +1461,44 @@ class KeyValueStoreLookupAction : public DNSAction
...
+  DNSAction::Action operator()(DNSQuestion* dq, std::string* ruleresult) const override
+  {
+    std::vector<std::string> keys = d_key->getKeys(*dq);
+    std::string result;
+    for (const auto& key : keys) {
+      if (d_kvs->getRangeValue(key, result) == true) {
+        break;
+      }
+    }
+
```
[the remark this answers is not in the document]
rgacogne: The existing lookup functions set the tag to an empty value when nothing was found, so I'll keep that behaviour and document it better.

pdns/dnsdistdist/dnsdist-kvs.cc
rgacogne: The logic assumes that addresses are stored in binary, in network order, as returned by `KeyValueLookupKeySourceIP`. I guess I need to document that more clearly.

omoerbeek: Right, I made a wrong assumption, as long as embedded NUL bytes are handled ok that should be fine indeed.

rgacogne: You should not have had to guess, so I documented that a bit better :)

pdns/dnsdist-lua-actions.cc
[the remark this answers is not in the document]
rgacogne: Documented now, I'll ponder changing that behaviour in the next major release but it's out of scope for this PR.

omoerbeek (approved):
This looks all good to me now. TSAN failure should be investigated.

rgacogne:
> This looks all good to me now.

Thanks, Otto!

> TSAN failure should be investigated.

I looked into it, and it seems to be a timing issue (we have a QPS limit that might not match if the test takes too long). It's not related to this PR anyway so I'm going to merge it.

rgacogne:
Ah, I need to merge 10520 first so I'll wait a bit :)

PR #10527, opened 2021-06-25
dnsdist: Implement 'reload()' to rotate Log(Response)Action's log file

### LINKED ISSUE #10502: dnsdist: Ponder adding the ability to rotate the LogAction file
- Program: dnsdist
 - Issue type: Feature request

### Short description

Some users are using `LogAction` for long-term logging of queries, keeping it enabled at all time instead of only for debugging purpose. For that use-case it would be nice to be able to ask for a rotation of the log file from time to time, via the console.

rgacogne:
### Short description

Fixes #10502.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdist-lua-actions.cc
+++ b/pdns/dnsdist-lua-actions.cc
@@ -817,6 +817,7 @@ private:
 
     if (!fp) {
       /* don't fall on our sword when reopening */
+      warnlog("Unable to open file '%s' for logging: %s", d_fname, stringerror());
       return;
     }
 
--- a/pdns/dnsdist-lua-actions.cc
+++ b/pdns/dnsdist-lua-actions.cc
@@ -737,8 +737,7 @@ public:
       return;
     }
 
-    reopenLogFile();
-    if (!d_fp) {
+    if (!reopenLogFile())  {
       throw std::runtime_error("Unable to open file '" + str + "' for logging: " + stringerror());
     }
   }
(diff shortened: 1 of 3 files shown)


pdns/dnsdist-lua-actions.cc
```diff
@@ -806,46 +797,68 @@ class LogAction : public DNSAction, public boost::noncopyable
...
+    std::shared_ptr<FILE> fp = nullptr;
+
+    if (d_append) {
+      fp = std::shared_ptr<FILE>(fopen(d_fname.c_str(), "a+"), fclose);
+    }
+    else {
+      fp = std::shared_ptr<FILE>(fopen(d_fname.c_str(), "w"), fclose);
+    }
+
+    if (!fp) {
```
pieterlexis, line 818: would it make sense to log that we were unable to re-open the file?

rgacogne: Yes! I have just done that :+1:

pandaedward:
Hi team, quick question, how do I call the reload method of the `LogAction` please? It is not very well documented.

At the moment I have this:
```
addAction(AllRule(), LogAction("/var/log/dnsdist.log", false, true, false))
```

How do I configure the conf file if I want `dnsdist.log` to be rotated please?

rgacogne:
You are asking a question on a closed pull request, which is likely to be unseen. I would suggest posting on the mailing-list or on IRC instead.
The short version is that `LogAction("/var/log/dnsdist.log", false, true, false)` returns on object on which you can call the `reload()` method:
```
mylogaction = LogAction("/var/log/dnsdist.log", false, true, false)
[...]
mylogaction:reload()
```

This can be done via the console: https://dnsdist.org/guides/console.html or in `maintenance`, for example: https://dnsdist.org/reference/config.html#maintenance

PR #10534, opened 2021-06-28
auth: very much WIP for removing the need to use ::get after a ::lookup call

RobinGeuze:
### Short description

This (attempts) to remove the need for calling ::get after a ::lookup call by returning all the records directly in a passed along vector.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


miodvallat:
Not sure this is something worth doing. Lookups can return a lot of records, which would cause a huge `vector` to be created with all the record data in memory at the same time.

Moreover, some code paths do not necessarily need to process all lookup results and can act with a subset of the results (these are all the code path which nowadays use `lookupEnd` to abort partial lookups).

It is probably better to keep `lookup` and `get` separate and let users of `lookup` decide how they want to process the lookup results.

Issue #10497, opened 2021-06-15
dnsdist: dnstap writes down wrong protocol


tomverrecktecheib:
- Program: dnsdist 
 - Issue type: Bug report

### Short description

Dnstap writes down wrong protocol (see https://mailman.powerdns.com/pipermail/dnsdist/2021-June/001090.html). When querying dnsdist with DoH (tcp), dnstap shows UDP.

### Environment

 - Operating system: ---
 - Software version: 1.5.2
 - Software source: self-compiled

### Steps to reproduce

1. Query dnsdist with DoH (dig @10.1.1.1 +https google.com)

### Expected behaviour
I expect to see TCP for the DoH query in the dnstap logfile.

### Actual behaviour
Dnstap writes down the wrong protocol (UDP instead of TCP).
```
$ dnstap-read -p 2021-06-13-dnstap.fstrm
...
...
14-Jun-2021 14:19:30.058 CQ x.x.x.x:44749 -> y.y.y.y:443 UDP 51b google.com/IN/A
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id:   1015
;; flags: rd ad; QUESTION: 1, ANSWER: 0, AUTHORITY: 0, ADDITIONAL: 1
;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 1232
; COOKIE: 1f5e39ad95e85cd8
;; QUESTION SECTION:
;google.com.            IN    A

14-Jun-2021 14:19:30.059 CR x.x.x.x:44749 <- y.y.y.y:443 UDP 83b google.com/IN/A
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id:   1015
;; flags: qr rd ra; QUESTION: 1, ANSWER: 1, AUTHORITY: 0, ADDITIONAL: 1
;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 1232
; COOKIE: 1f5e39ad95e85cd80100000060c74952f681b113fec0d503
;; QUESTION SECTION:
;google.com.            IN    A

;; ANSWER SECTION:
google.com.        266    IN    A    216.58.215.238
...
...
```

PR #10537, opened 2021-06-29
dnsdist: Carry the exact incoming protocol (Do53, DNSCrypt, DoT, DoH) in DQ

rgacogne:
### Short description

This turned out to be more difficult than I expected. Part of the work already existed in #10338 but was not enough to be able to pass the right protocol to dnstap, which I wanted to do since it now supports DoT and DoH (updated dnstap.proto is included in this PR).

Fixes https://github.com/PowerDNS/pdns/issues/10497.

I added the "backport to dnsdist-1.6.x" label because I think we need to fix the dnstap issue on that branch as well, but we clearly should not backport this PR. Perhaps a simple test on whether the request was received over DoH (dq.du is not `nullptr` would be enough for 1.6.x?

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/test-dnsdist_cc.cc
+++ b/pdns/test-dnsdist_cc.cc
@@ -62,7 +62,7 @@ static void validateECS(const PacketBuffer& packet, const ComboAddress& expected
   uint16_t qtype;
   uint16_t qclass;
   DNSName qname(reinterpret_cast<const char*>(packet.data()), packet.size(), sizeof(dnsheader), false, &qtype, &qclass, &consumed);
-  DNSQuestion dq(&qname, qtype, qclass, nullptr, &rem, const_cast<PacketBuffer&>(packet), false, nullptr);
+  DNSQuestion dq(&qname, qtype, qclass, nullptr, &rem, const_cast<PacketBuffer&>(packet), DNSQuestion::Protocol::DoUDP, nullptr);
   BOOST_CHECK(parseEDNSOptions(dq));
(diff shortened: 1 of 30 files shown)


pdns/test-dnsdistpacketcache_cc.cc
```diff
@@ -15,6 +15,8 @@
 
 BOOST_AUTO_TEST_SUITE(test_dnsdistpacketcache_cc)
 
+static bool receivedOverUDP = true;
```
omoerbeek, line 18: Looks like these tests only do the `false/DoUDP` case. It might be good to add more cases (this is an issue with the existing tests already)

pdns/test-dnsdist_cc.cc
```diff
@@ -62,7 +62,7 @@ static void validateECS(const PacketBuffer& packet, const ComboAddress& expected
   uint16_t qtype;
```
omoerbeek, line 62: Looks like these tests only do the false/DoUDP case. It might be good to add more cases (this is an issue with the existing tests already)

omoerbeek (approved):
Indeed much longer that you;d think. Anyway, looks good (with a few nits).

pdns/test-dnsdistpacketcache_cc.cc
rgacogne: I added one such unit test, but this is quite extensively tested in the regression tests.

PR #10543, opened 2021-06-30
auth: actually use the QueryCache for the UeberBackend unit tests

RobinGeuze:
### Short description

Due to an oversight the QueryCache wasn't actually enabled for the UeberBackend unit tests cases where it should.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code


RobinGeuze:
So enabling the query cache breaks certain test cases, will need to be fixed. They appear to be test bugs rather than code bugs so far.

Habbie (approved):

PR #10545, opened 2021-06-30
rec, nsec3dig: NSEC(3) debugging improvements

Habbie:
### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/pdns/nsec3dig.cc
+++ b/pdns/nsec3dig.cc
@@ -36,6 +36,7 @@ StatBag S;
 
 typedef std::pair<string,string> nsec3;
 typedef set<nsec3> nsec3set;
+typedef map<string, int> nsec3types;
 
 static string nsec3Hash(const DNSName &qname, const string &salt, unsigned int iters)
 {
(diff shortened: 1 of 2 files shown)


pdns/nsec3dig.cc
```diff
@@ -163,8 +163,9 @@ try
   set<DNSName> namesseen;
   set<DNSName> namestocheck;
   nsec3set nsec3s;
+  nsec3types nsec3t;
   string nsec3salt;
-  int nsec3iters = 0;
+  int nsec3iters = -1;
```
rgacogne, line 168: That sounds good but the implications of this change are not obvious to me :-)

rgacogne (approved):

pdns/nsec3dig.cc
Habbie: ah, I was trying to add detection of changed params but it was going to be ugly, I forgot to take this bit out. I'll fix that :)

rgacogne (approved):
Nice! I love the new comments in the aggressive cache, thanks!

omoerbeek (approved):

PR #10553, opened 2021-07-01
rec: assorted improvements

Habbie:
### Short description
a05df8bd4
   remove DNS_OVER_TLS guard around applying the dot-to-port-853 setting

   before this commit: pointing a forward at port 853, without DoT support,
   causes Recursor to attempt to do UDP over port 853 to the upstream. This
   rarely works.

   after this commit: much swifter failure with an error log message saying
   `45.55.10.200:853 requested but not available`

b42f9eb31
   emit error when dot-to-auth-names is set without DoT support

3f2abbbe0
   toTimestampStringMilli is only called from inside this file, make it static

e1cab5663153c17eca8f40574e6029be7bfb7dc9
    rec docs: fix warnings

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/pdns_recursor.cc
+++ b/pdns/pdns_recursor.cc
@@ -5599,7 +5599,7 @@ catch(...) {
 //static std::string s_timestampFormat = "%m-%dT%H:%M:%S";
 static std::string s_timestampFormat = "%s";
 
-const char* toTimestampStringMilli(const struct timeval& tv, char *buf, size_t sz)
+static const char* toTimestampStringMilli(const struct timeval& tv, char *buf, size_t sz)
 {
   struct tm tm;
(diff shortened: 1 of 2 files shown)


Habbie:
>  emit error when dot-to-auth-names is set without DoT support

Maybe we should `exit(1);` here.

omoerbeek:
This all looks good, though I'm a bit confused. I added ` --enable-dns-over-tls` to the `build-recursor` target in the circleCI yaml.

Habbie:
> I added ` --enable-dns-over-tls` to the `build-recursor` target in the circleCI yaml.

The UDP/853 problem is happening on Buildbot, which uses our built Debian packages. Pieter said he'd fix that soon.

On CircleCI we appear to be having a more intermittent problem.

omoerbeek (approved):

PR #10564, opened 2021-07-05
Backport 10555 to rec 4.5.x: Work around clueless servers sending AA=0 answers

omoerbeek:
### Short description

Backport of #10555 

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/recursordist/test-syncres_cc3.cc
+++ b/pdns/recursordist/test-syncres_cc3.cc
@@ -106,8 +106,8 @@ BOOST_AUTO_TEST_CASE(test_unauth_any)
 
   vector<DNSRecord> ret;
   int res = sr->beginResolve(target, QType(QType::ANY), QClass::IN, ret);
-  BOOST_CHECK_EQUAL(res, RCode::ServFail);
-  BOOST_CHECK_EQUAL(ret.size(), 0U);
+  BOOST_CHECK_EQUAL(res, RCode::NoError);
+  BOOST_CHECK_EQUAL(ret.size(), 1U);
 }
 
 static void test_no_data_f(bool qmin)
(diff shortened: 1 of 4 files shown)


liewebagency:
Hello,

Incredible WORK! I use Baremetal Dedicated servers. I dont like the containers no offence, I like the good old away to add a GUI Admin as Vhost on either Apache2 or Nginx. What do you have probrly have experienced with all the GUI admins for PowerDNS, and also in general for a safe and FAST, SECURE, what is the best for Webhosting for small ammount of Web Customers, its DNS and Email which is our priority. But still we want to make sure the hosting is FAST And SECURE.

One more thing, Hetzner which is my provider has never given me a straight answer on how to setup ipv6 dedicated ip in /etc/network/interfaces like with ip4 its simple:
For example:
auto eth0:22
allow-hotplug eth0:22
iface eth0:22 inet static
    address 138.the rest of ipv4
    netmask 255.255.255.224

But ive tried ipv6, and its something not right, they are routed trough the Main ip like ALL ipv4 is and connected the network cards Mac number so I know net mask or gateway isnt needed but I like to be sure.

The only one which seems to work of the /64 range I got is the default one with the Hetzner installing debin reacue system:
iface eth0 inet6 static
  address 2a01:4f8:173:2382::2
  netmask 64
  gateway fe80::1

But its linked to the same interface as the main ipv4. I want to setup my own ipv6 on own virtual interfaces. Any ideas ? I know this is abit out of this field, but I would like to start using ipv6 of everything since that’s the way it seems it goes in the future. Someone also said it is faster somehow.

One more think, FLOATED ip was something I was well, told to consider, to put 2 of the 4, split them up 2 and 2 on each vsitch. Or HaProxy. Any tips what is the best ?

I would apriciate any help.

//David Lie
***@***.***
Lie Web Agency
+474139910

rgacogne (approved):

PR #10565, opened 2021-07-05
rec: do not use DNSKEYs found below an apex for validation

Habbie:
### Short description
Test added.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


rgacogne (commented):
Don't we need the same logic in `SyncRes::doCacheCheck`, for on-demand validation?

Habbie:
> Don't we need the same logic in `SyncRes::doCacheCheck`, for on-demand validation?

You're right, a `+noad +cd` query followed by a normal one hits that path, and fails.

Habbie:
> Don't we need the same logic in `SyncRes::doCacheCheck`, for on-demand validation?

pushed, including two more tests

rgacogne (approved):

Habbie:
@omoerbeek I'll leave the merge timing and backport decisions to you.

Issue #7558, opened 2019-03-11
rec: optional detailed tracing info in protobuf


wojas:
- Program:  Recursor
- Issue type: Feature request 

### Short description

For performance analysis it would be very helpful to have detailed tracing information available in the protobuf. Right now we can tell that a query was slow, but we cannot tell which part of its execution was slow

This is related to #7420, but it takes it a step further with detailed timing information for just about everything we may find useful.

### Usecase

Enable detailed per-query analytics on recursor performance. Right now we can only do external black box measurements and have to deduce internal performance from aggregated statistics.

### Description

I propose extending the `PBDNSMessage` protobuf with new optional trace messages. These messages would be generated all over the code base with a relative nanosecond timestamp.

```protobuf
message PBDNSMessage {

  message Trace {
    // TODO: These events just illustrate the concept. The actual event types
    // would have to be determined during the implementation.
    enum EventType {
      Processing = 1;      // Start/end processing (can occur multiple times)
      KernelRecv = 2;      // Kernel received the packet
      RecRecv = 3;         // Recursor received the packet
      DistPipe = 4;        // Put in worker pipe by distributor or popped out (intVal: worker ID)
      PCacheCheck = 5;     // Packet cache check (boolVar indicates hit)
      QCacheCheck = 6;     // Query cache check (boolVar indicates hit)
      DNSOutQuery = 7;     // DNS outgoing query sent/received

      // Lua calls (at start and end using the 'start' flag)
      LuaGettag = 50;      // intVal: returned cache tag (also for gettag_ffi)
      LuaPreResolve = 51;
      LuaPostResolve = 52;
      LuaPostResolve = 53;
      LuaIPFilter = 54;
      LuaPreRPZ = 55;
      LuaNoData = 56;
      LuaNXDomain = 57;
      LuaPreOutQuery = 58;
    }

    // Timestamp in nanoseconds since first recorded event (up to ~4.3 sec)
    optional uint32 ts = 1;
    required EventType event = 2;
    optional bool start = 3;        // Indicates if this is a start or stop of given event
    
    // These are optional parameters with addition trace information, depending on EventType
    optional bool boolVal = 4;         // cache hit, etc
    optional int32 intVal = 5;         // worker ID, etc
    optional string stringVal = 6;
    optional binary binaryVal = 7;
  }
  repeated Trace trace = ...; // FIXME: next available field number here
}
```

[...]

PR #10567, opened 2021-07-06
Rec: eventtrace

omoerbeek:
### Short description

This is a stab at event traces, as suggested in #7558

Quite some rough edges: 
- always on, needs a (runtime configurable!) setting
- lots of more event types to be created and added to the code
- protobuf mesages are created too early, so the trace in the protobuf does not contains all info.
- types of timestamps and values have to be validated. ts are now ns and 64-bit to avoid having to think about wrapping. We might want to squeeze out a few bits. OTOH, I seem to remember protobuf uses variable length coding anyway, so leading zero bytes do not hurt really.
- docs
- tests

Example output (no PC hit):
```
        - RecRecv(241)
        - PCacheCheck(35979)
        - PCacheCheck(40177,False,done)
        - SyncRes(82968)
        - SyncRes(3803710778,0,done)
```

With a PC hit:
```
- Event Trace:
        - RecRecv(321)
        - PCacheCheck(47110)
        - PCacheCheck(64343,True,done)
```
Note that the `AnswerSent` events are missing, see the bullet point above

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsmessage.proto
+++ b/pdns/dnsmessage.proto
@@ -126,6 +126,34 @@ message PBDNSMessage {
     required MetaValue value = 2;
   }
   repeated Meta meta = 22;                      // Arbitrary meta-data - to be used in future rather than adding new fields all the time
+
+  enum EventType {
+    RecRecv = 1;
+    DistPipe = 2;
+    PCacheCheck = 3;
+    SyncRes = 4;
+    AnswerSent = 5;
+    LuaGetTag = 100;
+    LuaGetTagFFI = 101;
+    LuaIPFilter = 102;
+    LuaPreRPZ = 103;
+    LuaPreResolve = 104;
+    LuaPreOutQuery = 105;
+    LuaPostResolve = 106;
+    LuaNoData = 107;
+    LuaNXDomain = 108;
+  }
+  
+  message Event {
+    required uint64 ts = 1;
+    required EventType event = 2;
+    required bool start = 3;
+    optional bool boolVal = 4;
+    optional int64 intVal = 5;
+    optional string stringVal = 6;
+    optional bytes bytesVal = 7;
+  }
+  repeated Event trace = 23;
 }
 
 message PBDNSMessageList {
--- a/pdns/recursordist/rec-eventtrace.hh
+++ b/pdns/recursordist/rec-eventtrace.hh
@@ -22,8 +22,13 @@
 #pragma once
 
 #include "namespaces.hh"
+#include "misc.hh"
+#include "noinitvector.hh"
+
+#include <protozero/pbf_writer.hpp>
 #include <optional>
 #include <time.h>
+#include <unordered_map>
 #include <variant>
 
 class RecEventTrace
@@ -32,24 +37,23 @@ public:
   enum EventType : uint8_t
   {
     // Don't forget to add a new entry to the table in the .cc file!
-    Processing = 1,
-    RecRecv = 2,
-    DistPipe = 3,
-    PCacheCheck = 4,
-    SyncRes = 5,
-    AnswerSent = 6,
-    LuaGetTag = 50,
-    LuaGetTagFFI = 51,
-    LuaIPFilter = 52,
-    LuaPreRPZ = 53,
-    LuaPreResolve = 54,
-    LuaPreOutQuery = 55,
-    LuaPostResolve = 56,
-    LuaNoData = 57,
-    LuaNXDomain = 58
+    RecRecv = 1,
+    DistPipe = 2,
+    PCacheCheck = 3 ,
+    SyncRes = 4,
+    AnswerSent = 5,
+    LuaGetTag = 100,
+    LuaGetTagFFI = 101,
+    LuaIPFilter = 102,
+    LuaPreRPZ = 103,
+    LuaPreResolve = 104,
+    LuaPreOutQuery = 105,
+    LuaPostResolve = 106,
+    LuaNoData = 107,
+    LuaNXDomain = 108,
   };
 
-  static const std::map<EventType, std::string> s_eventNames;
+  static const std::unordered_map<EventType, std::string> s_eventNames;
 
   RecEventTrace()
   {
@@ -63,6 +67,7 @@ public:
   {
     old.d_status = Invalid;
   }
+
   RecEventTrace(RecEventTrace&& old) :
     d_events(std::move(old.d_events)),
     d_base(old.d_base),
@@ -70,6 +75,7 @@ public:
   {
     old.d_status = Invalid;
   }
+  
   RecEventTrace& operator=(const RecEventTrace& old) = delete;
   RecEventTrace& operator=(RecEventTrace&& old)
   {
@@ -80,7 +86,8 @@ public:
     return *this;
   }
 
-  typedef std::variant<std::nullopt_t, bool, int32_t, uint32_t, std::string> Value_t;
+  // We distinguisg beteen string and byres. Does not amtter in C++, but in Go, .Java etc it does
+  typedef std::variant<std::nullopt_t, bool, int64_t, std::string, PacketBuffer> Value_t;
 
   static std::string toString(const EventType v)
   {
@@ -95,21 +102,22 @@ public:
     else if (std::holds_alternative<bool>(v)) {
       return std::to_string(std::get<bool>(v));
     }
-    else if (std::holds_alternative<int32_t>(v)) {
-      return std::to_string(std::get<int32_t>(v));
-    }
-    else if (std::holds_alternative<uint32_t>(v)) {
-      return std::to_string(std::get<uint32_t>(v));
+    else if (std::holds_alternative<int64_t>(v)) {
+      return std::to_string(std::get<int64_t>(v));
     }
     else if (std::holds_alternative<std::string>(v)) {
       return std::get<std::string>(v);
+    }
+    else if (std::holds_alternative<PacketBuffer>(v)) {
+      const PacketBuffer& p = std::get<PacketBuffer>(v);
+      return makeHexDump(std::string(reinterpret_cast<const char*>(p.data()), p.size()));
     }
     return "?";
   }
 
   struct Entry
   {
-    Entry(Value_t& v, EventType e, bool start, uint32_t ts) :
+    Entry(Value_t& v, EventType e, bool start, uint64_t ts) :
       d_value(v), d_ts(ts), d_event(e), d_start(start)
     {
     }
@@ -197,6 +205,11 @@ public:
     return ret;
   }
 
+  const std::vector<Entry>& getEvents() const
+  {
+    return d_events;
+  }
+  
 private:
   std::vector<Entry> d_events;
   uint64_t d_base;
--- a/pdns/recursordist/rec-eventtrace.hh
+++ b/pdns/recursordist/rec-eventtrace.hh
@@ -1,0 +1,210 @@
+/*
+ * This file is part of PowerDNS or dnsdist.
+ * Copyright -- PowerDNS.COM B.V. and its contributors
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of version 2 of the GNU General Public License as
+ * published by the Free Software Foundation.
+ *
+ * In addition, for the avoidance of any doubt, permission is granted to
+ * link this program with OpenSSL and to (re)distribute the binaries
+ * produced as the result of such linking.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+#pragma once
+
+#include "namespaces.hh"
+#include <optional>
+#include <time.h>
+#include <variant>
+
+class RecEventTrace
+{
+public:
+  enum EventType : uint8_t
+  {
+    // Don't forget to add a new entry to the table in the .cc file!
+    Processing = 1,
+    RecRecv = 2,
+    DistPipe = 3,
+    PCacheCheck = 4,
+    SyncRes = 5,
+    AnswerSent = 6,
+    LuaGetTag = 50,
+    LuaGetTagFFI = 51,
+    LuaIPFilter = 52,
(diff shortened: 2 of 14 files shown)


omoerbeek:
@wojas I think this is ready , albeit limited in scope. Can you take a look if this is going in the right direction? 

The objective is to include this in 4.6.0 as an experimental feature and then do a round of improvements (preferably based on real-life experience with it)  in the 4.7 release cycle.

pdns/dnsmessage.proto
```diff
@@ -126,6 +126,34 @@ message PBDNSMessage {
     required MetaValue value = 2;
   }
   repeated Meta meta = 22;                      // Arbitrary meta-data - to be used in future rather than adding new fields all the time
+
+  enum EventType {
+    RecRecv = 1;
```
wojas, line 131: It would be useful to have comments here describing the various events.

pdns/dnsmessage.proto
```diff
@@ -126,6 +126,34 @@ message PBDNSMessage {
     required MetaValue value = 2;
   }
   repeated Meta meta = 22;                      // Arbitrary meta-data - to be used in future rather than adding new fields all the time
+
+  enum EventType {
```
wojas, line 130: It would be useful to have an event for when an mthread is active.

This was the idea behind the `Processing` in the original issue, but perhaps a more explicit name like `MThreadActive` would be better.

pdns/dnsmessage.proto
```diff
@@ -126,6 +126,34 @@ message PBDNSMessage {
...
+  message Event {
+    required uint64 ts = 1;
+    required EventType event = 2;
+    required bool start = 3;
+    optional bool boolVal = 4;
+    optional int64 intVal = 5;
+    optional string stringVal = 6;
+    optional bytes bytesVal = 7;
+  }
+  repeated Event trace = 23;
```
wojas, line 156: All of this will need to be added https://github.com/PowerDNS/dnsmessage before merge.

pdns/dnsmessage.proto
```diff
@@ -126,6 +126,34 @@ message PBDNSMessage {
...
+    LuaIPFilter = 102;
+    LuaPreRPZ = 103;
+    LuaPreResolve = 104;
+    LuaPreOutQuery = 105;
+    LuaPostResolve = 106;
+    LuaNoData = 107;
+    LuaNXDomain = 108;
+  }
+  
+  message Event {
```
wojas, line 147: In the future we may want to add a Lua API to add custom events. It may make sense to add a `name` field for such custom events. It would be useful to already allocate the field here, so that performance tooling can already depend on it being present.

The C++ events would not use this, because we know in advance what events are possible and it would just increase the size of the protobuf.

wojas (commented):
Very nice! I will try to create some tooling to process this data soon.

I'm curious how much processing overhead enabling this would add in the recursor.

wojas:
> types of timestamps and values have to be validated. ts are now ns and 64-bit to avoid having to think about wrapping. We might want to squeeze out a few bits. OTOH, I seem to remember protobuf uses variable length coding anyway, so leading zero bytes do not hurt really.

The protobuf uses variable length encoding indeed, so this only affects the representation on the C++ side. 4 seconds in uint32 is probably too tight, so uint64 makes sense. Perhaps make it int64 to calculate differences? This does not affect the wire format.

pdns/dnsmessage.proto
omoerbeek: I briefly looked into this, but atm it is not easy to do. The `eventTrace` data structure is not available in the spots where the context switching happens. I think I'll opt to postpone this until a later iteration.

omoerbeek:
> > types of timestamps and values have to be validated. ts are now ns and 64-bit to avoid having to think about wrapping. We might want to squeeze out a few bits. OTOH, I seem to remember protobuf uses variable length coding anyway, so leading zero bytes do not hurt really.
> 
> The protobuf uses variable length encoding indeed, so this only affects the representation on the C++ side. 4 seconds in uint32 is probably too tight, so uint64 makes sense. Perhaps make it int64 to calculate differences? This does not affect the wire format.

The protobuf value and associated methods already moved to `int64`, the comment in the PR is outdated.

pdns/dnsmessage.proto
omoerbeek: Either that or reserve a numeric range for "custom events"?

omoerbeek:
Rebased.

pdns/dnsmessage.proto
wojas: Then you would have to maintain and distribute a mapping to every component that needs it in order to interpret those custom events.  Not impossible, but the gains would need to be worth the pain.

pdns/dnsmessage.proto
```diff
@@ -126,6 +126,36 @@ message PBDNSMessage {
     required MetaValue value = 2;
   }
   repeated Meta meta = 22;                      // Arbitrary meta-data - to be used in future rather than adding new fields all the time
+
+  // The well known EventTrace event numbers
+  enum EventType {
+    CustomEvent = 0;                            // A custom event
+    RecRecv = 1;                                // A request was received by the recursor process
```
rgacogne, line 133: We might want to rename that event so that dnsdist and the auth can later use the same event? Something like `RequestReceived`?

pdns/dnsmessage.proto
```diff
@@ -126,6 +126,36 @@ message PBDNSMessage {
...
+    PCacheCheck = 2;                            // A packet cache check was initiated or completed; value: bool cacheHit
+    SyncRes = 3;                                // Syncres main function has started or completed; value: int rcode
```
rgacogne, line 135: I wonder if it would make sense to move events specific to the recursor to the 2xx range?

pdns/recursordist/rec-eventtrace.hh
```diff
@@ -0,0 +1,248 @@
...
+
+  static const std::unordered_map<EventType, std::string> s_eventNames;
+
+  RecEventTrace()
+  {
+    reset();
+  }
+
+  RecEventTrace(const RecEventTrace& old) :
+    d_events(std::move(old.d_events)),
```
rgacogne, line 64: I doubt we will be able to move the content of a const object, and I don't believe we should.

pdns/recursordist/rec-eventtrace.hh
```diff
@@ -0,0 +1,248 @@
...
+    d_base(old.d_base),
+    d_status(old.d_status)
+  {
+    old.d_status = Invalid;
```
rgacogne, line 68: That sounds fishy, do we really want this? Doing a copy of an object invalidating that object?

pdns/recursordist/rec-eventtrace.hh
```diff
@@ -0,0 +1,248 @@
...
+
+  const std::vector<Entry>& getEvents() const
+  {
+    return d_events;
+  }
+
+private:
+  std::vector<Entry> d_events;
+  int64_t d_base;
+  enum Status
```
rgacogne, line 241:
```suggestion
  enum Status : uint8_t
```

pdns/recursordist/rec-eventtrace.hh
```diff
@@ -0,0 +1,248 @@
...
+      const PacketBuffer& p = std::get<PacketBuffer>(v);
+      return makeHexDump(std::string(reinterpret_cast<const char*>(p.data()), p.size()));
+    }
+    return "?";
+  }
+
+  struct Entry
+  {
+    Entry(Value_t& v, EventType e, bool start, int64_t ts) :
+      d_value(v), d_ts(ts), d_event(e), d_start(start)
```
rgacogne, line 121:
```suggestion
    Entry(Value_t&& v, EventType e, bool start, int64_t ts) :
      d_value(std::move(v)), d_ts(ts), d_event(e), d_start(start)
```

pdns/recursordist/rec-eventtrace.hh
```diff
@@ -0,0 +1,248 @@
...
+  {
+    assert(d_status != Invalid);
+    if (d_status == Disabled) {
+      return;
+    }
+    struct timespec ts;
+    clock_gettime(CLOCK_MONOTONIC, &ts);
+    int64_t stamp = ts.tv_nsec + ts.tv_sec * 1000000000;
+    stamp -= d_base;
+    d_events.emplace_back(v, e, start, stamp);
```
rgacogne, line 171:
```suggestion
  void add(E e, Value_t&& v, bool start)
  {
    assert(d_status != Invalid);
    if (d_status == Disabled) {
      return;
    }
    struct timespec ts;
    clock_gettime(CLOCK_MONOTONIC, &ts);
    int64_t stamp = ts.tv_nsec + ts.tv_sec * 1000000000;
    stamp -= d_base;
    d_events.emplace_back(std::move(v), e, start, stamp);
```

pdns/recursordist/rec-eventtrace.hh
```diff
@@ -0,0 +1,248 @@
...
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+#pragma once
+
+#include "namespaces.hh"
+#include "misc.hh"
+#include "noinitvector.hh"
+
+#include <protozero/pbf_writer.hpp>
```
rgacogne, line 28:
```suggestion
```

pdns/recursordist/rec-eventtrace.hh
```diff
@@ -0,0 +1,248 @@
...
+    d_status = Disabled;
+  }
+
+  std::string toString() const
+  {
+    assert(d_status != Invalid);
+    if (d_status == Disabled) {
+      return "Disabled\n";
+    }
+    std::string ret = "eventTrace [";
```
rgacogne, line 218: I wonder if using an `ostringstream` would make this easier, but that does not really matter.

pdns/dnsmessage.proto
```diff
@@ -126,6 +126,36 @@ message PBDNSMessage {
...
+    LuaPreRPZ = 103;
+    LuaPreResolve = 104;
+    LuaPreOutQuery = 105;
+    LuaPostResolve = 106;
+    LuaNoData = 107;
+    LuaNXDomain = 108;
+  }
+
+  message Event {
+    required int64 ts = 1;                      // Timestamp in ns relative to time of creation of event trace data structure
```
rgacogne, line 149: I'm not sure I see a case where this can be negative? If we know some events happen before the creation of the event trace structure, perhaps it would make sense to set the base time of event trace structure to the first of these events when it is created, provided that it can be done?

rgacogne (commented):
I like that new feature very much! Did you get a chance to measure the overhead introduced by this tracing, by any chance?

pdns/dnsmessage.proto
omoerbeek: > I like that new feature very much! Did you get a chance to measure the overhead introduced by this tracing, by any chance?

not yet. My plan wast to get the semantics and code in shape and then do measurements.

omoerbeek: > We might want to rename that event so that dnsdist and the auth can later use the same event? Something like `RequestReceived`?

Yes, makes sense

pdns/dnsmessage.proto
omoerbeek: Yes. Will do so.

pdns/recursordist/rec-eventtrace.hh
omoerbeek: The idea was to invalidate the object to detect the case that we transferred ownership to a different owner and still try to modify the original.

pdns/dnsmessage.proto
omoerbeek: An example of events with a negative (relative) timestamps are kernel provided packet timestamps. By defintion they are eaelri than the eventTrace data strcuture creation (the current base ts). If we change the base beacuse of such an eaelier timestamp, all already added event timestamps should be adapted as well.

pdns/recursordist/rec-eventtrace.hh
omoerbeek: yes, I messed things up there.

pdns/dnsmessage.proto
rgacogne: Understood, ans I guess it's not easy to pass that timestamp at the creation of the eventTrace data structure. That means the receiver will either need to work with negative time offsets as well or scan the events an compute the relative timestamps, I guess?

omoerbeek: The timestamps are already stored relatively. Alternatively I could do a "normalize" step when exporting, so that the oldest timestamp becomes zero, but it will cost some effort. I'll think about it.

wojas: `int64` makes calculating differences safer. The same proto will also be used by consumers that will base their type on this definition.

Ideally we would avoid actually sending negative timestamp. If we do want to send those, perhaps use `sint64`, which uses a zig-zag encoding to prevent negative numbers from using the maximum size in the wire protocol.

pdns/recursordist/rec-eventtrace.hh
omoerbeek: Looked at this again. This was an attempt at avoiding the copy the event vector in `handleNewUDPQuestion`. There is a local obect there that is being passed (by value) to `distributeAsyncFunction`. The original object will be reused and so it does not make a lot of sense to keep the vector in the original. 

I will reinstate a regular copy constructor, but it would be nice to avoid the copy of the vector and do a move. But OTOH it is maybe overkill, it will contain very few events at this point in time.

pdns/dnsmessage.proto
omoerbeek: I decided to keep the int64_t. When a event is added before the current `d_base`, its timestamp wil become the new `d_base` and existing events will be adjusted to be relative to the new `d_base`.

The main reason I'm using relative timestamps is hat it will generate much smaller values, so that the protobuf encoding will be relatively small as well.

I also discovered that kernel packet timestamps use an unspecified time base, so it might be hard to fit them in. But that is for the future.

omoerbeek:
Initial measurements showsa qps drop of 10%, doing only queries in the packet cache (I figured that for that case the performance impact would be worst). This is a bit more than expected. Investigating.

omoerbeek:
gprof is showing that `pdns::ProtoZero::RecMessage::addEvents(RecEventTrace const&)` takes about 4% of CPU time. Some other `RecEventTrace` related methods a few percents as well.

omoerbeek:
Even though it takes a bit more CPU to add the data to the protobuf message than anticipated I'd say this is good enough to be included in rec-4.6.0. It *is* marked as experimental and we can spend some time making it faster later.

omoerbeek:
Rebased to fix conflict

PR #10570, opened 2021-07-06
rec: Make really sure we did not miss a cut on validation failure

rgacogne:
### Short description

Since 4.5.0 we do not pre-compute the zone cuts at the start of the resolution process, which saves a lot of queries but also works around broken DNS implementations that do not handle DS queries correctly. Unfortunately before this PR we were sometimes misled by signed, broken but insecure zones (no DS at the parent) and could return an erroneous Bogus status.
This PR keeps the global logic introduced in 4.5, following the hint provided by the signer field in signatures, since that works fine for 99% of the cases, but double-checks whether we missed a zone cut if the answer does not validate properly.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/pdns/syncres.cc
+++ b/pdns/syncres.cc
@@ -1714,7 +1714,7 @@ void SyncRes::computeNegCacheValidationStatus(const NegCache::NegCacheEntry& ne,
     vState neValidationState = ne.d_validationState;
     dState expectedState = res == RCode::NXDomain ? dState::NXDOMAIN : dState::NXQTYPE;
     dState denialState = getDenialValidationState(ne, expectedState, false);
-    updateDenialValidationState(neValidationState, ne.d_name, state, denialState, expectedState);
+    updateDenialValidationState(neValidationState, ne.d_name, state, denialState, expectedState, qtype == QType::DS, depth);
   }
   if (state != vState::Indeterminate) {
     /* validation succeeded, let's update the cache entry so we don't have to validate again */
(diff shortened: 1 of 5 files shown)


pdns/syncres.cc
```diff
@@ -2755,8 +2763,17 @@ vState SyncRes::validateDNSKeys(const DNSName& zone, const std::vector<DNSRecord
...
-    return state;
+    LOG(d_prefix<<": let's check whether we missed a zone cut before returning a Bogus state from "<<__func__<<"("<<zone<<")"<<endl);
+    /* try again to get the missed cuts, harder this time */
+    auto zState = getValidationStatus(zone, false, false, depth);
+    if (zState == vState::Secure) {
+      /* too bad */
+      LOG(d_prefix<<": after checking the zone cuts we are still in a Secure zone, returning Bogus state from "<<__func__<<"("<<zone<<")"<<endl);
+      return state;
+    }
+    else {
```
omoerbeek, line 2774: Maybe log something here ?

pdns/syncres.cc
```diff
@@ -2799,45 +2816,70 @@ vState SyncRes::getDNSKeys(const DNSName& signer, skeyset_t& keys, unsigned int
...
+    if (state != vState::Secure) {
+      /* try again to get the missed cuts, harder this time */
+      LOG(d_prefix<<"checking whether we missed a zone cut before returning a Bogus state"<<endl);
+      auto zState = getValidationStatus(name, false, type == QType::DS, depth);
+      if (zState == vState::Secure) {
+        LOG(d_prefix<<"we are still in a Secure zone, returning "<<vStateToString(state)<<endl);
+        /* too bad */
         return state;
       }
+      else {
```
omoerbeek, line 2879: Same comment about logging here

pdns/syncres.cc
```diff
@@ -2853,7 +2895,17 @@ vState SyncRes::validateRecordsWithSigs(unsigned int depth, const DNSName& qname
...
-  return state;
+  /* try again to get the missed cuts, harder this time */
+  auto zState = getValidationStatus(name, false, type == QType::DS, depth);
+  LOG(d_prefix<<"checking whether we missed a zone cut before returning a Bogus state"<<endl);
+  if (zState == vState::Secure) {
+    /* too bad */
+    LOG(d_prefix<<"we are still in a Secure zone, returning "<<vStateToString(state)<<endl);
+    return state;
+  }
+  else {
```
omoerbeek, line 2906: And here

pdns/syncres.cc
```diff
@@ -3402,8 +3456,16 @@ void SyncRes::updateDenialValidationState(vState& neValidationState, const DNSNa
       neValidationState = vState::Insecure;
     }
     else {
-      LOG(d_prefix<<"Invalid denial found for "<<neName<<", returning Bogus, res="<<denialState<<", expectedState="<<expectedState<<endl);
-      neValidationState = vState::BogusInvalidDenial;
+      LOG(d_prefix<<"Invalid denial found for "<<neName<<", res="<<denialState<<", expectedState="<<expectedState<<", checking whether we have missed a zone cut before returning a Bogus state"<<endl);
+      /* try again to get the missed cuts, harder this time */
+      auto zState = getValidationStatus(neName, false, isDS, depth);
+      if (zState != vState::Secure) {
```
omoerbeek, line 3462: And here

omoerbeek (commented):
Would it make sense to add logging in the mentioned places? Our log are already quite talkative, so only if it really adds something...

rgacogne:
Looks like this PR causes a loop in some cases, investigating..

rgacogne:
> Would it make sense to add logging in the mentioned places? Our log are already quite talkative, so only if it really adds something...

I forgot to answer, sorry about that. The code looking at the cuts is already pretty verbose, so I don't think it would add anything in these cases, no.

omoerbeek (approved):

PR #10587, opened 2021-07-13
rec: Ancestor NSEC3s can only deny the existence of a DS

rgacogne:
### Short description

Before that commit, the aggressive NSEC(3) cache could have mistakenly used NSEC3s from the parent zone to prove that a given name in the child zone did not exist, which is incorrect.
It happened because we did not properly detect that the NSEC3 for the closest encloser was an ancestor NSEC3 indicating a delegation, and then in the unlikely but possible case that we found a NSEC3 from the parent zone whose hashes covered the next closer we wrongly concluded that the name did not exist, returning a NXDomain with an invalid proof of denial.

This PR needs to be backported to rel/rec-4.5.x where the aggressive NSEC(3) cache was introduced.

It also explicitly sets the number of NSEC3 iterations in the unit tests that clearly depend on that (although I might have missed some) to prevent a breakage if we ever update the default value.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/pdns/recursordist/aggressive_nsec.cc
+++ b/pdns/recursordist/aggressive_nsec.cc
@@ -545,6 +545,8 @@ bool AggressiveNSECCache::getNSEC3Denial(time_t now, std::shared_ptr<AggressiveN
     }
 
     const DNSName signer = getSigner(exactNSEC3.d_signatures);
+    /* here we need to allow an ancestor NSEC3 proving that a DS does not exist as it is an
+       exact match for the name */
     if (type != QType::DS && isNSEC3AncestorDelegation(signer, exactNSEC3.d_owner, nsec3)) {
       /* RFC 6840 section 4.1 "Clarifications on Nonexistence Proofs":
          Ancestor delegation NSEC or NSEC3 RRs MUST NOT be used to assume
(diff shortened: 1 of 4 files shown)


pdns/recursordist/aggressive_nsec.cc
```diff
@@ -580,6 +580,18 @@ bool AggressiveNSECCache::getNSEC3Denial(time_t now, std::shared_ptr<AggressiveN
         break;
       }
 
+      const DNSName signer = getSigner(closestNSEC3.d_signatures);
+      if (type != QType::DS && isNSEC3AncestorDelegation(signer, closestNSEC3.d_owner, nsec3)) {
```
Habbie, line 584: Does this mean we do allow the NSEC3 from above the cut to deny a DS one level below the cut?

rgacogne: Nice catch! I pushed a fix, several comments, and amended the test to catch that as well.

rgacogne:
Rebased to fix a conflict.

pdns/recursordist/test-aggressive_nsec_cc.cc
```diff
@@ -1106,4 +1340,538 @@ BOOST_AUTO_TEST_CASE(test_aggressive_nsec3_rollover)
...
+
+    BOOST_CHECK_EQUAL(cache->getEntriesCount(), 1U);
+
+    /* the cache should now be able to deny other types (except the DS) */
+    int res;
+    std::vector<DNSRecord> results;
+    BOOST_CHECK_EQUAL(cache->getDenial(now, name, QType::AAAA, results, res, ComboAddress("192.0.2.1"), boost::none, true), true);
+    BOOST_CHECK_EQUAL(res, RCode::NoError);
+    BOOST_CHECK_EQUAL(results.size(), 3U);
+    /* but not any type that lives in the child zone */
```
omoerbeek, line 1393: You're testing the DS case, not the child zone case here I believe, so the comment does not seem to match

pdns/recursordist/test-aggressive_nsec_cc.cc
```diff
@@ -1106,4 +1340,538 @@ BOOST_AUTO_TEST_CASE(test_aggressive_nsec3_rollover)
...
```
omoerbeek, line 1584: Here the comment also does not seem to match the code

rgacogne:
Nicely spotted, thanks a lot! I pushed a commit fixing these erroneous comments.

omoerbeek (approved):

Issue #10367, opened 2021-05-06
SVCB support request for dnsdist 


sauceress:
- Program: dnsdist 
 - Issue type: Feature request

### Short description
We want enhanced SVCB support within dnsdist that supports multiple records with the hints returned in the 'additional' section  in addition to IPv4 and IPv6 hints (since the standard is not yet finalized). This request is from Quad9 DNS. 

### Usecase

Format reference: https://datatracker.ietf.org/doc/html/draft-schwartz-svcb-dns-03#section-8

_dns.resolver.arpa          SVCB 1  . ( alpn="dot" port="853" ipv4hint="9.9.9.9" ipv6hint="2620:fe::fe" )
_dns.resolver.arpa          SVCB 2  . ( alpn="h2" port="443" ipv4hint="9.9.9.9" ipv6hint="2620:fe::fe" dohpath="/dns-query{?dns}" )
         

### Description
We want SVCB records with entries in the 'additional' section in addition to IPv4 and IPv6 hints (since the standard is not yet finalized).

Per https://datatracker.ietf.org/doc/html/draft-schwartz-svcb-dns-03#section-10 the number for 'dohpath' is still undefined. We would like the ability to set/override this value at run time. 

We need multiple records definable each with their own set of characteristics. 
Example (per https://datatracker.ietf.org/doc/html/draft-schwartz-svcb-dns-03#section-8)

         _dns 7200 IN SVCB 1 @ alpn=dot,h2,h3 dohpath=/dns-query{?dns}
                               SVCB 2 @ alpn=dot port=8530
                               SVCB 3 fooexp port=5353 alpn=foo foo-info=...

Ultimately we want to be able to do this synthetically within dnsdist without requiring a backing zone, to support real time reflection of load balancing decisions.

An alternative is to chain A record look-ups for insertion into the additional section if IPv4 and IPv6 hints are also included. We'd prefer to do a single lookup versus more than one to be able to start using DNS. Our preference is for a single lookup and response. 

Perhaps this could be a LUA object represented as a table. 

spoofSVCB({
  index=NUM,  
  dohpath="STRING", 
  ipv4hints="STRING", (could also be combo address or a group of combo addresses)
  ipv6hints="STRING", (could also be combo address or a group of combo addresses)
  alpn="STRING",
  port=NUM
})

In order to synthesize the equivalent of:  

[...]

PR #10597, opened 2021-07-19
dnsdist: Implement SpoofSVCAction to return SVC responses

rgacogne:
### Short description

Preliminary implementation of #10367. It's very likely that some parts of the design will change, and it would be nice to extend the regression tests which are a bit limited due to missing support in the current versions of dnspython.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code


rgacogne:
Rebased to fix a conflict.

PR #10598, opened 2021-07-20
Rec: almost expired refactor and stats

omoerbeek:
### Short description

Reorganized things a bit to have a better separation between general and specific code for the task queue. Also added metrics that are "almost-expired" specific, since in the future we likely will have other task types. Note that I skip a SNMP id since there's is another PR pending that adds 114 (#10546). 

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [X] documented the code


pdns/recursordist/rec-taskqueue.cc
```diff
@@ -21,18 +21,62 @@
...
 #include "syncres.hh"
 
 static thread_local pdns::TaskQueue t_taskQueue;
+static pdns::stat_t s_almost_expired_tasks_pushed;
+static pdns::stat_t s_almost_expired_tasks_run;
+static pdns::stat_t s_almost_expired_tasks_exceptions;
+
+static void resolve(const struct timeval& now, bool logErrors, const pdns::ResolveTask& task)
+{
+  ++s_almost_expired_tasks_run;
```
rgacogne, line 35: I wonder if we should only increase that counter after the task was successfully completed? That way we would have `almost-expired-pushed` - `currently queued tasks` == `almost-expired-run` + `almost-expired-exceptions`. It does not matter much, to be fair.

pdns/recursordist/taskqueue.cc
```diff
@@ -64,33 +64,7 @@ bool TaskQueue::runOnce(bool logErrors)
...
-    }
-    catch (const PolicyHitException& e) {
-      if (logErrors) {
-        g_log << Logger::Notice << "Policy hit while running the background task queue" << endl;
-      }
-    }
-    catch (...) {
-      g_log << Logger::Error << "Exception while running the background task queue" << endl;
-    }
+    task.func(now, logErrors, task);
```
rgacogne, line 67: It might be nice to check that `task.func` is not `nullptr` here.

omoerbeek: Done

omoerbeek:
Rebased after #10598 and review comments processed.

rgacogne (approved):

PR #10601, opened 2021-07-20
Document setTag(Response)Action behaviour

steffann:
### Short description
It surprised me that setTagAction and setTagResponseAction didn't overwrite existing tag values. This adds a line to the docs so that future users aren't surprised by this.

### Checklist
I have:
- [ ] compiled this code
- [X] tested this code
- [X] documented the code


rgacogne (approved):
Thanks a lot! I wonder whether it makes sense to keep that behaviour, but in the meantime it's good to have it properly documented.

PR #10602, opened 2021-07-20
rec: Process policy and potential Drop action after Lua hooks

omoerbeek:
### Short description

If preresolve returns `true` and indicates the packet should be dropped, do it. 

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/pdns_recursor.cc
+++ b/pdns/pdns_recursor.cc
@@ -1918,6 +1918,13 @@ static void startDoResolve(void *p)
           if (answerIsNOData(dc->d_mdp.d_qtype, res, ret)) {
             if (t_pdl && t_pdl->nodata(dq, res)) {
               shouldNotValidate = true;
+              auto policyResult = handlePolicyHit(appliedPolicy, dc, sr, res, ret, pw);
+              if (policyResult == PolicyResult::HaveAnswer) {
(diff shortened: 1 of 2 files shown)


pdns/pdns_recursor.cc
```diff
@@ -1939,6 +1939,15 @@ static void startDoResolve(void *p)
         }
       }
     }
+    else if (t_pdl) {
```
omoerbeek, line 1942: Maybe it would be better to rearrange a bit so this block ends up close to the `preresolve()` call?

rgacogne (approved):
Looks good. A regression test would be nice to have. I wonder if we should call `handlePolicyHit` after the `nodata` and `nxdomain` hooks as well?

pdns/pdns_recursor.cc
omoerbeek: I tried, but it was too messy in the end.

omoerbeek:
> Looks good. A regression test would be nice to have. I wonder if we should call `handlePolicyHit` after the `nodata` and `nxdomain` hooks as well?

I'll take a look at the other cases and ad regress test(s).

rgacogne (approved):

PR #10613, opened 2021-07-26
auth: correct upper bounds on d_qtypecounters

Habbie:
### Short description
This fixes security advisory 2021-01.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/responsestats.cc
+++ b/pdns/responsestats.cc
@@ -27,11 +27,11 @@ static auto sizeBounds()
 ResponseStats::ResponseStats() :
   d_sizecounters("SizeCounters", sizeBounds())
 {
-  for (unsigned int n = 0; n < 65535; ++n) {
-    d_qtypecounters[n].value = 0;
+  for (auto& entry : d_qtypecounters) {
+    entry.value = 0;
   }
-  for (unsigned int n = 0; n < 256; ++n) {
-    d_rcodecounters[n].value = 0;
+  for (auto& entry : d_rcodecounters) {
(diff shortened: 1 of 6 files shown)


pdns/responsestats.cc
```diff
@@ -53,7 +53,7 @@ map<uint16_t, uint64_t> ResponseStats::getQTypeResponseCounts() const
 {
   map<uint16_t, uint64_t> ret;
   uint64_t count;
-  for (unsigned int i = 0; i < 65535; ++i) {
+  for (unsigned int i = 0; i < d_qtypecounters.size(); ++i) {
```
RobinGeuze, line 56: Why not use the `auto& entry : d_qtypecounters` syntax here as well?

pdns/responsestats.cc
```diff
@@ -77,7 +77,7 @@ map<uint8_t, uint64_t> ResponseStats::getRCodeResponseCounts() const
 {
   map<uint8_t, uint64_t> ret;
   uint64_t count;
-  for (unsigned int i = 0; i < 256; ++i) {
+  for (unsigned int i = 0; i < d_rcodecounters.size(); ++i) {
```
RobinGeuze, line 80: Same here.

pdns/responsestats.cc
Habbie: hmm, we need the index

RobinGeuze (approved):

PR #10617, opened 2021-07-28
Auth: add zone to zonecache on flush endpoint

zeha:
### Short description
Some people want to add domains to the zonecache as soon as possible. This is currently automatic when the zone is created using the API or autoprimary. For database-based provisioning systems this allows the provisioning system to add the zone by calling the /cache/flush endpoint.

### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code


pieterlexis (approved):
code lgtm. Small question: but would it make sense to have a separate endpoint for the zonecache specifically so the records are not purged from the caches?

klaus-nicat:
In our use case the flush would be a benefit:
1. Customer adds zone
2. Customer requests lets encrypt certificated
3. Lets encrypt performs a DNS lookup
4. The zone-list cache does not have the new zone in the cache -> REFUSED is returned and put in packet cache.
5. Lets encrypt fails. Customer has to wait until packet cache expires.

With "flush newzone" the zone would be added to the zone list cache and old cache entries could be purged.

How is the flush working? will it flush from query-cache and packet-cache?

zeha:
> code lgtm. Small question: but would it make sense to have a separate endpoint for the zonecache specifically so the records are not purged from the caches?

I think that would just lead to two requests in sequence - I don't see a valid usecase for -not- clearing the other caches really.

klaus-nicat:
Any chance to have this in 4.5.2? It can be seen as a missing feature of the new zone-list cache. Thanks

mind04:
I do see why you need this.  The api and autoprimary both add the zone to the zone cache. Adding a zone a second time will not help you when the packet cache is the actual problem. Flushing the packetcache is already possible and this should be enough.

Adding zones in the flush code will create unnecessary duplicates in the zone cache every time flush is called.

klaus-nicat:
The packet cache is not the actual problem. The problem is that zones added directly to the database (DB replication to public facing name servers) will not be seen immediately. Hence, I need a way to make the zone-list cache aware of the new zone (without completely reloading the list). I do not mind if this is done via the "flush" endpoint or a new endpoint. In my case I like the above solution as I would need to call "flush" anyways - but if necessary - i could also live with a second API call to add and flush.

zeha:
> I do not see why you need this. The api and autoprimary both add the zone to the zone cache.

As said in the original MR text: direct database inserts.

mind04:
I'm no fan of the "put it in flush" approach. Flush and add-zone have very different use cases.
A separate add-zone endpoint to support cases like this might be an option. But we must make it very clear in the docs you do not need this for zones which are created via the api.

zeha:
I can split it into a new endpoint, but:

- naming it is hard. Certainly needs the “most people should not call this” explanation. Recommendations on the name?
- people who want to call this will also end up calling flush, to clear any old entries in packetcache

Therefore I am not sure it’s worth splitting really.

klaus-nicat:
I have a question about the cache-flush: does it flush single domain names or all domain names belonging to a zone? For example, can I flush www.example.com? Or only example.com which will flush also www.example.com? In the first case, if someone flushes www.example.com I will not that www.example.com will be added to the zone-list cache.

Maybe an additional URL parameter to the flush endpoint can be used to add the zone to the zone-list cache.

zeha:
I believe that "removing the virtual negative cache entry from zonecache" shall be thought of as "flushing the (zone)cache". I am inclined to put the code into `purgeAuthCaches(Exact)` instead of the API code though.
Would appreciate someone to step into this discussion to un-stuck it.

pdns/ws-auth.cc
```diff
@@ -2255,6 +2255,14 @@ static void apiServerCacheFlush(HttpRequest* req, HttpResponse* resp) {
 
   DNSName canon = apiNameToDNSName(req->getvars["domain"]);
 
+  SOAData sd;
+  UeberBackend B;
+  if (g_zoneCache.isEnabled() && B.getSOAUncached(canon, sd)) {
+    // zone exists (uncached), add/update it in the zone cache.
```
Habbie, line 2261: this only checks that a SOA exists in `records`. It does not check that the domain is in `domains`. If it is not, itl'l fall out of the zone cache in ~30 seconds, which is confusing. Can you add that check?

pdns/ws-auth.cc
```diff
@@ -2255,6 +2255,14 @@ static void apiServerCacheFlush(HttpRequest* req, HttpResponse* resp) {
...
+    // Handle this first, to avoid concurrent queries re-populating the other caches.
+    g_zoneCache.add(canon, sd.domain_id);
+  }
+
   uint64_t count = purgeAuthCachesExact(canon);
```
Habbie, line 2266: the `Exact` purge is not actually what the 'new zone' situation wants, but I don't immediately know how to fix that cleanly. I wonder if the API should just grow a suffix purge?

Habbie (changes requested):

pdns/ws-auth.cc
zeha: For (some) consistency with `pdns_control`, maybe the API should append '$' and call `purgeCaches` instead (giving us a "subtree" purge). This is also what rfc2136handler.cc does.

Not sure if one should also clear DNSSECKeeper caches, as pdns_control does today.

Habbie: > For (some) consistency with `pdns_control`, maybe the API should append '$' and call `purgeCaches` instead (giving us a "subtree" purge). This is also what rfc2136handler.cc does.

This, or add a flag, not sure what is cleaner. Your proposal changes the behaviour - but we could argue it improves it.

> Not sure if one should also clear DNSSECKeeper caches, as pdns_control does today.

I'm thinking yes.

pdns/ws-auth.cc
zeha: Right, I shall switch this to use B.getAuth(... cachedOk=false) really!

Habbie: > Right, I shall switch this to use B.getAuth(... cachedOk=false) really!

Hmm, does that actually hit `domains`?

zeha: Hmm, no. getDomainInfo it is, then.

pdns/ws-auth.cc
zeha: I've tried to improve the cache cleaning bits so there's now a `purgeAuthCacheForZone` function that should do the right thing in all of these cases. I am not super convinced it actually covers everything, but thats what I can see from existing usage of separate cache cleaning functions. API now also uses `purgeAuthCacheForZone`, resulting in a zone cache, dnsseckeeper cache, query, packet cache flush.

pdns/ws-auth.cc
```diff
@@ -968,7 +968,7 @@ static void apiZoneMetadata(HttpRequest* req, HttpResponse *resp) {
       throw ApiException("Could not update metadata entries for domain '" +
         zonename.toString() + "'");
 
-    DNSSECKeeper::clearMetaCache(zonename);
+    purgeAuthCacheForZone(zonename);
```
Habbie, line 971: This flushes a lot more than it used to, is that necessary?

Habbie (commented):
Nice refactor. One question posted. Will do another review later.

pdns/auth-caches.cc
```diff
@@ -45,7 +47,31 @@ uint64_t purgeAuthCaches(const std::string& match)
...
+{
+  uint64_t ret = 0;
+
+  if (g_zoneCache.isEnabled()) {
+    DomainInfo di;
+    UeberBackend B;
+    if (B.getDomainInfo(name, di, false)) {
+      // zone exists, add/update it in the zone cache.
+      // Handle this first, to avoid concurrent queries re-populating the other caches.
+      g_zoneCache.add(name, di.id);
```
Habbie, line 61: this takes a lock even if the zone cache already had the entry. Perhaps `g_zoneCache.add` could be smarter about that?

pdns/auth-caches.cc
```diff
@@ -45,7 +47,31 @@ uint64_t purgeAuthCaches(const std::string& match)
...
-/* remove specific entries from all caches, no wildcard matching */
+/* remove zone from all caches */
+uint64_t purgeAuthCacheForZone(const DNSName& name)
+{
+  uint64_t ret = 0;
+
+  if (g_zoneCache.isEnabled()) {
+    DomainInfo di;
+    UeberBackend B;
+    if (B.getDomainInfo(name, di, false)) {
```
Habbie, line 58: This is not free, and almost always unnecessary. For example, the 2136 code that calls this even already has the di for you.

pdns/ws-auth.cc
zeha: "Maybe". Depends on what metadata setting is changed.

If the idea previously was to make DNSSECKeeper see new data, the outcome of that never made it into the packetcache. If its a SOA-EDIT-xxx metadata change, I think its expected that all caches are flushed and queries immediately see the new SOA (edits), etc?

pdns/auth-caches.hh
```diff
@@ -25,6 +25,7 @@
 
 #include "dnsname.hh"
 
-uint64_t purgeAuthCaches(); /* empty all caches */
-uint64_t purgeAuthCaches(const std::string& match); /* remove specific entries from all caches, can be $ terminated */
-uint64_t purgeAuthCachesExact(const DNSName& qname); /* remove specific entries from all caches, no wildcard matching */
+uint64_t purgeAuthCaches(); /* empty packet and query caches */
+uint64_t purgeAuthCaches(const std::string& match); /* remove specific entries from packet and query caches, can be $ terminated */
+uint64_t purgeAuthCacheForZone(const DNSName& name); /* remove zone from all caches */
+uint64_t purgeAuthCachesExact(const DNSName& qname); /* remove specific entries from packet and query caches, no wildcard matching */
```
jsoref, line 31: Does `no wildcard matching` imply that `purgeAuthCaches` and `purgeAuthCacheForZone` both support wildcards?

Habbie:
Postponed to 4.7, I asked @zeha to separately PR the original, smaller version of this, so that we can go into 4.6-beta1 without a lot of new changes.

pdns/auth-caches.hh
zeha: i think so!

zeha:
#11066 targets 4.6.0.

klaus-nicat:
I think that could be closed as ##11066 already solved the issue.

PR #10623, opened 2021-08-02
Rec: make all PacketIDs shared_ptr

omoerbeek:
### Short description

This PR moves to shared objects for the `PacketID` keys to the events index.

The TCP code already uses `shared_ptr<PacketID>`'s, but then passed the `*` object to the event code. Since the event code copies the keys (of type `PacketId`), updates to the state where not reflected after `waitEvent` returned. In the case of TCP timeouts than could lead to inconsistent state and FD actions.

This also moves everything (including the UDP code) to `shared_ptr<PacketID>` and fixes a bug where the regular `operator<` and the birthday compare functions did not agree.

Issues found while working on a "Keep TCP connections open" branch.
 
### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/syncres.hh
+++ b/pdns/syncres.hh
@@ -974,7 +974,7 @@ struct PacketID
       return false;
     }
 
-    return tie(fd, id, domain) < tie(b.fd, b.id, b.domain);
+    return tie(domain, fd, id) < tie(b.domain, b.fd, b.id);
   }
 };
 
--- a/pdns/syncres.hh
+++ b/pdns/syncres.hh
@@ -967,12 +967,10 @@ struct PacketID
 
   bool operator<(const PacketID& b) const
   {
-    int ourSock= tcpsock;
-    int bSock = b.tcpsock;
-    if (tie(remote, ourSock, type) < tie(b.remote, bSock, b.type)) {
+    if (tie(remote, tcpsock, type) < tie(b.remote, b.tcpsock, b.type)) {
       return true;
     }
-    if (tie(remote, ourSock, type) > tie(b.remote, bSock, b.type)) {
+    if (tie(remote, tcpsock, type) > tie(b.remote, b.tcpsock, b.type)) {
       return false;
     }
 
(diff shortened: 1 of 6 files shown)


omoerbeek:
Investigating the bulk test failure

rgacogne:
> This also moves everything (including the UDP code) to shared_ptr<PacketID> and fixes a bug where the regular operator< and the birthday compare functions did not agree.

Looks like I introduced that issue in ec3f366df73eee1b32016f7d3614ec74579cd721 :-/

omoerbeek:
> > This also moves everything (including the UDP code) to shared_ptr and fixes a bug where the regular operator< and the birthday compare functions did not agree.
> 
> Looks like I introduced that issue in [ec3f366](https://github.com/PowerDNS/pdns/commit/ec3f366df73eee1b32016f7d3614ec74579cd721) :-/

And I approved it (also :-/)

pdns/syncres.hh
```diff
@@ -967,38 +967,53 @@ struct PacketID
 
   bool operator<(const PacketID& b) const
   {
-    int ourSock= tcpsock;
-    int bSock = b.tcpsock;
-    if (tie(remote, ourSock, type) < tie(b.remote, bSock, b.type)) {
+    // We don't want explcit PacketID compare here, but always via predicate classes below
```
rgacogne, line 970:
```suggestion
    // We don't want explicit PacketID compare here, but always via predicate classes below
```

pdns/syncres.hh
```diff
@@ -967,38 +967,53 @@ struct PacketID
...
       return false;
     }
 
-    return tie(fd, id, domain) < tie(b.fd, b.id, b.domain);
+    return tie(a->domain, a->fd, a->id) < tie(b->domain, b->fd, b->id);
   }
 };
 
-struct PacketIDBirthdayCompare: public std::binary_function<PacketID, PacketID, bool>
+struct PacketIDBirthdayCompare
```
rgacogne, line 1001: I'm pondering whether adding some comment would be a good idea, something like:
"This comparison needs to be consistent (same ordering, so we can omit right-most keys but not change the order, see CompatibleCompare) with the one in `PacketIDCompare` above."

rgacogne (approved):

pdns/syncres.hh
rgacogne: Perhaps some unit tests would help as well :-)

omoerbeek: will do so.

PR #10630, opened 2021-08-03
Dev/no tcp

gothremote:
### Short description
Load key from ISCMap (pub and priv) based on changed openssl lib. 
Change size to fit bit size of falcon


gothremote:
Error merging in wrong repo

PR #10647, opened 2021-08-16
Fix a warning about catching a polymorphic exception type by value

rgacogne:
### Short description

```
decafsigners.cc: In member function ‘virtual bool DecafED25519DNSCryptoKeyEngine::verify(const string&, const string&) const’:
decafsigners.cc:140:11: warning: catching polymorphic type ‘class decaf::CryptoException’ by value [-Wcatch-value=]
  140 |   } catch(CryptoException) {
      |           ^~~~~~~~~~~~~~~
decafsigners.cc: In member function ‘virtual bool DecafED448DNSCryptoKeyEngine::verify(const string&, const string&) const’:
decafsigners.cc:276:11: warning: catching polymorphic type ‘class decaf::CryptoException’ by value [-Wcatch-value=]
  276 |   } catch(CryptoException) {
      |           ^~~~~~~~~~~~~~~
```

On the other hand, I wonder if we should not drop the decaf code altogether since:
- 15 is provided by `libsodium` and recent versions of `OpenSSL`
- 16 is provided by recent versions of `OpenSSL`
- but more importantly it looks like we don't really test that code..

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/decafsigners.cc
+++ b/pdns/decafsigners.cc
@@ -137,7 +137,7 @@ bool DecafED25519DNSCryptoKeyEngine::verify(const std::string& msg, const std::s
 
   try {
     pub.verify(sig, message);
-  } catch(CryptoException) {
+  } catch(const CryptoException& e) {
     return false;
   }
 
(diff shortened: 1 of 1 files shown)


omoerbeek (approved):
OK for the change. As for dropping libdecaf support, I agree.

PR #10648, opened 2021-08-17
Drop libdecaf support

rgacogne:
### Short description

Algo 15 is now provided by libsodium and recent versions of OpenSSL, 16 by recent versions of OpenSSL as well, but more importantly it looks like we don't really enable and test the libdecaf code.

### Checklist

I have:
- [x] compiled this code
- [ ] tested this code
- [ ] documented the code


mind04:
This is too soon. With this pull you are dropping ed448 support for distributions with a not recent enough openssl version, like centos7.

rgacogne:
> This is too soon. With this pull you are dropping ed448 support for distributions with a not recent enough openssl version, like centos7.

I don't mind keeping that support a bit longer, but is decaf support actually enabled on our CentOS 7 builds?

omoerbeek:
> > This is too soon. With this pull you are dropping ed448 support for distributions with a not recent enough openssl version, like centos7.
> 
> I don't mind keeping that support a bit longer, but is decaf support actually enabled on our CentOS 7 builds?

On our CentOS7 builds it is not:
```
checking whether we will be linking in libdecaf... no
...
configure: OpenSSL ecdsa: yes
configure: ed25519: yes
configure: ed448: no
```

rgacogne:
Is libdecaf available there? If so, shall we enable it?

omoerbeek:
> Is libdecaf available there? If so, shall we enable it?

https://pkgs.org/search/?q=decaf
Does not list a lot of platforms having it.

omoerbeek:
> > Is libdecaf available there? If so, shall we enable it?
> 
> https://pkgs.org/search/?q=decaf
> Does not list a lot of platforms having it.

I think the conclusion is: libdecaf is generally unsupported and we can get rid of the support

rgacogne:
Rebased to fix a conflict.

PR #10660, opened 2021-08-23
auth: incoming PROXY support

### LINKED ISSUE #8403: auth: Support proxy protocol for incoming queries
- Program: Authoritative
 - Issue type: Feature request

### Short description
Basically support the other side of #8401, to be able to provide the original source and destination addresses and ports, as well as any TLVs data, to the `Lua` hooks. We might also want to be able to rewrite the source and destination address passed to the rest of the code, including backends.

We should be careful to only accept these from whitelisted hosts.

Habbie:
### Short description
incoming PROXY support for:

* AXFR ACLs
* NOTIFY sources
* getting the remote address in LUA records
* 2136 ACLs
* query/response accounting
* log output
* the remote used by backends to take geo-style decisions

Fixes #8403.

Needs:
* [x] tests
* [x] docs

Check that backends use the right IP:
* [x] lua2backend
* [x] geoipbackend
* [x] remotebackend
* [x] tinydnsbackend
* [x] pipebackend

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -1253,7 +1253,7 @@ prevent-self-notification to "no".
 
 Turn on operating as a primary. See :ref:`primary-operation`.
 
-.. _setting-proxy-protocol-from
+.. _setting-proxy-protocol-from:
 
 ``proxy-protocol-from``
 -----------------------
@@ -1266,6 +1266,7 @@ Ranges that are required to send a Proxy Protocol version 2 header in front of U
 Queries that are not prefixed with such a header will not be accepted from clients in these ranges. Queries prefixed by headers from clients that are not listed in these ranges will be dropped.
 
 Note that once a Proxy Protocol header has been received, the source address from the proxy header instead of the address of the proxy will be checked against primary addresses sending NOTIFYs, and the ACLs for any client requesting AXFRs.
+When using this setting combined with :ref:`setting-trusted-notification-proxy`, please be aware that the trusted address will also be checked against the source address in the PROXY header.
 
 .. _setting-proxy-protocol-maximum-size:
 
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -1253,6 +1253,31 @@ prevent-self-notification to "no".
 
 Turn on operating as a primary. See :ref:`primary-operation`.
 
+.. _setting-proxy-protocol-from
+
+``proxy-protocol-from``
+-----------------------
+.. versionadded:: 4.6.0
+
+-  IP addresses or netmasks, separated by commas
+-  Default: empty
+
+Ranges that are required to send a Proxy Protocol version 2 header in front of UDP and TCP queries, to pass the original source and destination addresses and ports to the Authoritative.
+Queries that are not prefixed with such a header will not be accepted from clients in these ranges. Queries prefixed by headers from clients that are not listed in these ranges will be dropped.
+
+Note that once a Proxy Protocol header has been received, the source address from the proxy header instead of the address of the proxy will be checked against primary addresses sending NOTIFYs, and the ACLs for any client requesting AXFRs.
+
+.. _setting-proxy-protocol-maximum-size:
+
+``proxy-protocol-maximum-size``
+-------------------------------
+.. versionadded:: 4.6.0
+
+-  Integer
+-  Default: 512
+
+The maximum size, in bytes, of a Proxy Protocol payload (header, addresses and ports, and TLV values). Queries with a larger payload will be dropped.
+
 .. _setting-query-cache-ttl:
 
 ``query-cache-ttl``
--- a/pdns/common_startup.cc
+++ b/pdns/common_startup.cc
@@ -479,7 +479,7 @@ try
     S.ringAccount("queries", question.qdomain, question.qtype);
     S.ringAccount("remotes", question.d_remote);
     if(logDNSQueries) {
-      g_log << Logger::Notice<<"Remote "<< question.getRemoteString() <<" wants '" << question.qdomain<<"|"<<question.qtype.toString() <<
+      g_log << Logger::Notice<<"Remote "<< question.getRemoteString() <<" wants '" << question.qdomain<<"|"<<question.qtype <<
         "', do = " <<question.d_dnssecOk <<", bufsize = "<< question.getMaxReplyLen();
       if(question.d_ednsRawPacketSizeLimit > 0 && question.getMaxReplyLen() != (unsigned int)question.d_ednsRawPacketSizeLimit)
         g_log<<" ("<<question.d_ednsRawPacketSizeLimit<<")";
--- a/pdns/common_startup.cc
+++ b/pdns/common_startup.cc
@@ -95,7 +95,7 @@ void declareArguments()
   ::arg().setSwitch("dnsupdate","Enable/Disable DNS update (RFC2136) support. Default is no.")="no";
   ::arg().setSwitch("write-pid","Write a PID file")="yes";
   ::arg().set("allow-dnsupdate-from","A global setting to allow DNS updates from these IP ranges.")="127.0.0.0/8,::1";
-  ::arg().set("proxy-protocol-from","A Proxy Protocol header is only allowed from these subnets, and is mandatory then too.")="";
+  ::arg().set("proxy-protocol-from", "A Proxy Protocol header is only allowed from these subnets, and is mandatory then too.")="";
   ::arg().set("proxy-protocol-maximum-size", "The maximum size of a proxy protocol payload, including the TLV values")="512";
   ::arg().setSwitch("send-signed-notify", "Send TSIG secured NOTIFY if TSIG key is configured for a zone") = "yes";
   ::arg().set("allow-unsigned-notify", "Allow unsigned notifications for TSIG secured zones") = "yes"; //FIXME: change to 'no' later
--- a/pdns/common_startup.cc
+++ b/pdns/common_startup.cc
@@ -64,6 +64,8 @@ double avg_latency{0.0};
 unique_ptr<TCPNameserver> TN;
 static vector<DNSDistributor*> g_distributors;
 vector<std::shared_ptr<UDPNameserver> > g_udpReceivers;
+NetmaskGroup g_proxyProtocolACL;
+size_t g_proxyProtocolMaximumSize;
 
 ArgvMap &arg()
 {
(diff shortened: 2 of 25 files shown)


pdns/tcpreceiver.cc
```diff
@@ -240,9 +242,55 @@ void TCPNameserver::doConnection(int fd)
...
+      ssize_t used;
+
+      // this for-loop ends by throwing, or by having gathered a complete proxy header
+      for (;;) {
+        used = isProxyHeaderComplete(proxyData);
+        if (used < 0) {
+          ssize_t origsize = proxyData.size();
+          proxyData.resize(origsize + -used);
+          try {
+            readnWithTimeout(fd, &proxyData[origsize], -used, d_idleTimeout, true, remainingTime);
```
rgacogne, line 263: We might want to call `maxConnectionDurationReached()` here as well?

pdns/dnspacket.cc
```diff
@@ -423,7 +448,10 @@ int DNSPacket::noparse(const char *mesg, size_t length)
   d_rawpacket.assign(mesg,length); 
   if(length < 12) { 
     g_log << Logger::Debug << "Ignoring packet: too short ("<<length<<" < 12) from "
-      << d_remote.toStringWithPort()<< endl;
+      << d_remote.toStringWithPort();
+    if(d_inner_remote)
+      g_log << "("<<d_inner_remote->toStringWithPort()<<")";
+    g_log << endl;
```
rgacogne, line 454: Perhaps we could call `getRemoteString()`?

pdns/nameserver.cc
```diff
@@ -301,7 +303,26 @@ bool UDPNameserver::receive(DNSPacket& packet, std::string& buffer)
...
   else
-    packet.d_dt.set(); // timing    
+    packet.d_dt.set(); // timing
+
+  if (g_proxyProtocolACL.match(remote)) {
+    ComboAddress psource, pdestination;
+    bool proxyProto, tcp;
+    std::vector<ProxyProtocolValue> ppvalues;
+
+    ssize_t used = parseProxyHeader(std::string(buffer, 0, len), proxyProto, psource, pdestination, tcp, ppvalues);
```
rgacogne, line 313: I wonder if resizing `buffer` to `len` and then passing it directly wouldn't be easier. We will need to resize it to `DNSPacket::s_udpTruncationThreshold` before using it again anyway, because of the call to `buffer.erase(0, used)` below.

pdns/packethandler.cc
```diff
@@ -942,9 +942,14 @@ int PacketHandler::trySuperMaster(const DNSPacket& p, const DNSName& tsigkeyname
 int PacketHandler::trySuperMasterSynchronous(const DNSPacket& p, const DNSName& tsigkeyname)
 {
   ComboAddress remote = p.getRemote();
+  // this uses the outer (non-PROXY) remote on purpose
   if(p.hasEDNSSubnet() && pdns::isAddressTrustedNotificationProxy(remote)) {
     remote = p.getRealRemote().getNetwork();
   }
+  else {
+    // but we fall back to the inner (PROXY) remote if there is no ECS
```
rgacogne, line 950:
```suggestion
    // but we fall back to the inner (PROXY) remote if there is no ECS forwarded by a trusted proxy
```

pdns/dnspacket.hh
```diff
@@ -61,8 +61,9 @@ public:
   const string& getString(); //!< for serialization - just passes the whole packet
 
   // address & socket manipulation
-  void setRemote(const ComboAddress*);
+  void setRemote(const ComboAddress*, const ComboAddress* = nullptr);
```
rgacogne, line 64: Just a nit but perhaps void `setRemote(const ComboAddress&, std::optional<ComboAddress> = boost::none);` would convey the intent a bit better?

pdns/tcpreceiver.cc
```diff
@@ -240,9 +242,55 @@ void TCPNameserver::doConnection(int fd)
...
+          }
+          catch(NetworkError& ae) {
+            throw NetworkError("Error reading PROXYv2 header from TCP client "+remote.toString()+": "+ae.what());
+          }
+        }
+        else if (used == 0) {
+          throw NetworkError("Error reading PROXYv2 header from TCP client "+remote.toString()+": PROXYv2 header was invalid");
+        }
+        else if ((size_t)used > g_proxyProtocolMaximumSize) {
```
rgacogne, line 272:
```suggestion
        else if (static_cast<size_t>(used) > g_proxyProtocolMaximumSize) {
```

pdns/tcpreceiver.cc
```diff
@@ -240,9 +242,55 @@ void TCPNameserver::doConnection(int fd)
...
+      }
+      ComboAddress psource, pdestination;
+      bool proxyProto, tcp;
+      std::vector<ProxyProtocolValue> ppvalues;
+
+      used = parseProxyHeader(proxyData, proxyProto, psource, pdestination, tcp, ppvalues);
+      if (used <= 0) {
+        throw NetworkError("Error reading PROXYv2 header from TCP client "+remote.toString()+": PROXYv2 header was invalid");
+      }
+      if ((size_t) used > g_proxyProtocolMaximumSize) {
```
rgacogne, line 287:
```suggestion
      if (static_cast<size_t>(used) > g_proxyProtocolMaximumSize) {
```

pdns/tcpreceiver.cc
Habbie: Makes sense. I added that (locally) but I fail to trigger it directly. However, having it makes `readnWithTimeout` eventually fail, which is fine :)

pdns/dnspacket.cc
Habbie: In this case I felt it might be nice to keep the ports. I considered adding a `getRemoteStringWithPort` variant but this might be the only consumer. I'll ponder some more.

rgacogne: I totally missed the fact that this version has the ports! Keeping it as it is makes sense to me, then, although implementing `getRemoteStringWithPort()` to make it obvious to the next reader might be good as well :)

pdns/dnspacket.hh
Habbie: Yes! My compiler did not like `boost::none` but `std::nullopt` appears to work.

pdns/tcpreceiver.cc
Habbie: pushed

pdns/nameserver.cc
Habbie: good call on resize before use! pushed

pieterlexis:
Accepting only PROXY prefixed from the configured addresses might break the dnsdist healthcheck. Please check this :)

Habbie:
> Accepting only PROXY prefixed from the configured addresses might break the dnsdist healthcheck. Please check this :)

dnsdist adds a minimal PROXY header to the health check when `useProxyProtocol=true` is set on a babckend. I tested this and it works.

Habbie:
> I wonder if `getOuterRemote()` would make more sense

I cannot come up with a design for this that would make a lot of things very messy. Instead I'll do a thorough audit of all references to `d_remote` and `getRemote()`.

berthubert:
Hi! Can I interest you in this one line patch that adds 'localwho' to the Lua variables? This allows scripts to know on which port or address a query came in:
```
diff --git a/pdns/lua-record.cc b/pdns/lua-record.cc
index 682759432..d827d7e92 100644
--- a/pdns/lua-record.cc
+++ b/pdns/lua-record.cc
@@ -958,6 +958,7 @@ std::vector<shared_ptr<DNSRecordContent>> luaSynth(const std::string& code, cons
   lua.writeVariable("zone", zone);
   lua.writeVariable("zoneid", zoneid);
   lua.writeVariable("who", dnsp.getRemote());
+  lua.writeVariable("localwho", dnsp.getLocal());
   lua.writeVariable("dh", (dnsheader*)&dnsp.d);
   lua.writeVariable("dnssecOK", dnsp.d_dnssecOk);
   lua.writeVariable("tcp", dnsp.d_tcp);
```

pdns/common_startup.cc
```diff
@@ -93,6 +95,8 @@ void declareArguments()
   ::arg().setSwitch("dnsupdate","Enable/Disable DNS update (RFC2136) support. Default is no.")="no";
   ::arg().setSwitch("write-pid","Write a PID file")="yes";
   ::arg().set("allow-dnsupdate-from","A global setting to allow DNS updates from these IP ranges.")="127.0.0.0/8,::1";
+  ::arg().set("proxy-protocol-from","A Proxy Protocol header is only allowed from these subnets, and is mandatory then too.")="";
+  ::arg().set("proxy-protocol-maximum-size", "The maximum size of a proxy protocol payload, including the TLV values")="512";
```
jsoref, line 99: Occasional complaint about inconsistent whitespace between first and second arguments

Habbie: added a space :)

Habbie:
> Can I interest you in this one line patch that adds 'localwho' to the Lua variables?

On further inspection, it turns out that with this PR, it would no longer be a one line patch. Want to file a separate issue so we don't forget later?

pieterlexis (changes requested):
The name `inner_remote` confuses me greatly. Inner relative to what? Other than that, some small nits.

Habbie:
> Inner relative to what?

Relative to the physical IP packet header, which is 'outer'. I'll see if I can put some clarifying words in at least one comment.

pdns/common_startup.cc
```diff
@@ -469,12 +479,7 @@ try
     S.ringAccount("queries", question.qdomain, question.qtype);
     S.ringAccount("remotes", question.d_remote);
     if(logDNSQueries) {
-      string remote;
-      if(question.hasEDNSSubnet()) 
-        remote = question.getRemote().toString() + "<-" + question.getRealRemote().toString();
-      else
-        remote = question.getRemote().toString();
-      g_log << Logger::Notice<<"Remote "<< remote <<" wants '" << question.qdomain<<"|"<<question.qtype.toString() << 
+      g_log << Logger::Notice<<"Remote "<< question.getRemoteString() <<" wants '" << question.qdomain<<"|"<<question.qtype.toString() <<
```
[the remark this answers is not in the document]
Habbie: tested & confirmed & committed

regression-tests.auth-py/test_ProxyProtocol.py
```diff
@@ -0,0 +1,227 @@
...
+import socket
+import struct
+import threading
+import time
+import unittest
+
+from authtests import AuthTest
+from proxyprotocol import ProxyProtocol
+
+class TestProxyProtocolLuaRecords(AuthTest):
```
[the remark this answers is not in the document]
Habbie: pushed

docs/settings.rst
```diff
@@ -1253,6 +1253,31 @@ prevent-self-notification to "no".
 
 Turn on operating as a primary. See :ref:`primary-operation`.
 
+.. _setting-proxy-protocol-from
+
+``proxy-protocol-from``
```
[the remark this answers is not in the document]
Habbie: Added a note on the interaction.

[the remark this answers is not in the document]
Habbie: For guide/operational instructions, how about we add that to https://dnsdist.org/advanced/passing-source-address.html?highlight=proxy#proxy-protocol instead? The words would be pretty much identical for auth and rec anyway, and then dnsdist is the common factor (except when haproxy is the common factor, of course).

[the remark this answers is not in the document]
Habbie: Noted separately as #10731

Habbie:
> Relative to the physical IP packet header, which is 'outer'. I'll see if I can put some clarifying words in at least one comment.

Also did this.

Habbie:
With some documentation delegated to #10731, I believe I addressed all review comments.

PR #10669, opened 2021-08-25
Rec: tcp/dot connection pooling

omoerbeek:
### Short description

This maintains open TCP/DoT connections to auth or forwarders.  If a new connection is needed to a specific IP, one from the idle set is used if available. Connections are put in the idle set after a query is done with some basic success. If a connection comes from the idle set, it might be closed by the peer, so the code has to be prepared that it may not work and a new one has to be requested from the set (and if not available, created).

There might be some overlap with (upcoming) dnsdist code, we should look into merging at some point.
But for now I want to publish this so people can review and test.

ATM this mechanism is enabled with `tcp-out-max-idle-ms=10000`, `tcp-out-max-idle-per-auth=10`, `tcp-out-max-queries=0` (no maximum) and `"tcp-out-max-idle-per-thread=100`. See the docs for the semantics.

We might want to disable this by defaulting to `"tcp-out-max-idle-per-thread=0` on the final release.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [X] documented the code

### DIFF
--- a/pdns/lwres.cc
+++ b/pdns/lwres.cc
@@ -558,7 +558,7 @@ LWResult::Result asyncresolve(const ComboAddress& ip, const DNSName& domain, int
       ret = asyncresolve(ip, domain, type,doTCP, sendRDQuery, EDNS0Level, now, srcmask, context, outgoingLoggers, fstrmLoggers, exportTypes, lwr, chained, connection);
     } 
     if (connection.d_handler && lwr->d_validpacket) {
-      t_tcp_manager.store(ip, connection);
+      t_tcp_manager.store(*now, ip, std::move(connection));
     }
   }
   return ret;
--- a/pdns/lwres.cc
+++ b/pdns/lwres.cc
@@ -236,6 +236,7 @@ static bool tcpconnect(const struct timeval& now, const ComboAddress& ip, TCPOut
 {
   dnsOverTLS = SyncRes::s_dot_to_port_853 && ip.getPort() == 853;
 
+
   while (true) {
     connection = t_tcp_manager.get(ip);
     if (connection.d_handler) {
@@ -262,14 +263,50 @@ static bool tcpconnect(const struct timeval& now, const ComboAddress& ip, TCPOut
     }
     connection.d_handler = std::make_shared<TCPIOHandler>("", s.releaseHandle(), timeout, tlsCtx, now.tv_sec);
     // Returned state ignored
-    try {
-      connection.d_handler->tryConnect(SyncRes::s_tcp_fast_open_connect, ip);
-    }
-    catch (const std::runtime_error&) {
-      continue;
-    }
+    // This can throw an excepion, retry will need to happen at higher level
+    connection.d_handler->tryConnect(SyncRes::s_tcp_fast_open_connect, ip);
     return true;
   }
+}
+
+static LWResult::Result tcpsendrecv(const ComboAddress& ip, TCPOutConnectionManager::Connection& connection,
+                                    ComboAddress& localip, const vector<uint8_t>& vpacket, size_t& len, PacketBuffer& buf)
+{
+  socklen_t slen = ip.getSocklen();
+  uint16_t tlen = htons(vpacket.size());
+  const char *lenP = reinterpret_cast<const char*>(&tlen);
+  const char *msgP = reinterpret_cast<const char*>(&*vpacket.begin());
+
+  localip.sin4.sin_family = ip.sin4.sin_family;
+  getsockname(connection.d_handler->getDescriptor(), reinterpret_cast<sockaddr*>(&localip), &slen);
+
+  PacketBuffer packet;
+  packet.reserve(2 + vpacket.size());
+  packet.insert(packet.end(), lenP, lenP + 2);
+  packet.insert(packet.end(), msgP, msgP + vpacket.size());
+
+  LWResult::Result ret = asendtcp(packet, connection.d_handler);
+  if (ret != LWResult::Result::Success) {
+    return ret;
+  }
+
+  ret = arecvtcp(packet, 2, connection.d_handler, false);
+  if (ret != LWResult::Result::Success) {
+    return ret;
+  }
+
+  memcpy(&tlen, packet.data(), sizeof(tlen));
+  len = ntohs(tlen); // switch to the 'len' shared with the rest of the calling function
+
+  // XXX receive into buf directly?
+  packet.resize(len);
+  ret = arecvtcp(packet, len, connection.d_handler, false);
+  if (ret != LWResult::Result::Success) {
+    return ret;
+  }
+  buf.resize(len);
+  memcpy(buf.data(), packet.data(), len);
+  return LWResult::Result::Success;
 }
 
 /** lwr is only filled out in case 1 was returned, and even when returning 1 for 'success', lwr might contain DNS errors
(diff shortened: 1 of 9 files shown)


omoerbeek:
Build failure on centos7, investigating.

```
g++ -std=c++17 -DHAVE_CONFIG_H -I.  -I/usr/include/luajit-2.0   -I./ext/yahttp   -DNETSNMP_ENABLE_IPV6 -fno-strict-aliasing -O2 -g -pipe -Wall -Wp,-D_FORTIFY_SOURCE=2 -fexceptions -fstack-protector-strong --param=ssp-buffer-size=4 -grecord-gcc-switches -m64 -mtune=generic -D_RPM_4_4_COMPAT -Ulinux -Dlinux=linux -D_REENTRANT -D_GNU_SOURCE -fno-strict-aliasing -pipe -fstack-protector -I/usr/local/include -D_LARGEFILE_SOURCE -D_FILE_OFFSET_BITS=64 -I/usr/lib64/perl5/CORE -I. -I/usr/include   -O3 -Wall -pthread -DSYSCONFDIR=\"/etc/pdns-recursor\"  -I./ext/json11 -I./ext/protozero/include -I./ext/yahttp  -DBOOST_CONTAINER_USE_STD_EXCEPTIONS  -I/usr/include/luajit-2.0      -DSYSCONFDIR=\"/etc/pdns-recursor\" -DPKGLIBDIR=\"/usr/lib64/pdns-recursor\" -DLOCALSTATEDIR=\"/var/run\" -DNODCACHEDIR=\"/var/lib/pdns-recursor\" -fPIE -DPIE -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=2 --param ssp-buffer-size=4 -fstack-protector -Wall -Wextra -Wshadow -Wno-unused-parameter -Wmissing-declarations -Wredundant-decls -g -O2 -O2 -g -pipe -Wall -Wp,-D_FORTIFY_SOURCE=2 -fexceptions -fstack-protector-strong --param=ssp-buffer-size=4 -grecord-gcc-switches   -m64 -mtune=generic -c -o rec-tcpout.o rec-tcpout.cc
In file included from recursor_cache.hh:30,
                 from syncres.hh:41,
                 from rec-tcpout.hh:27,
                 from rec-tcpout.cc:23:
dnsrecords.hh: In member function 'virtual uint16_t CERTRecordContent::getType() const':
dnsrecords.hh:463:22: error: expected unqualified-id before 'char'
   includeboilerplate(CERT)
                      ^~~~
dnsrecords.hh:44:53: note: in definition of macro 'includeboilerplate'
   uint16_t getType() const override { return QType::RNAME; }                                   \
                                                     ^~~~~
dnsrecords.hh:463:22: error: expected ';' before 'char'
   includeboilerplate(CERT)
                      ^~~~
dnsrecords.hh:44:53: note: in definition of macro 'includeboilerplate'
   uint16_t getType() const override { return QType::RNAME; }                                   \
                                                     ^~~~~
dnsrecords.hh:463:22: error: declaration does not declare anything [-fpermissive]
   includeboilerplate(CERT)
                      ^~~~
dnsrecords.hh:44:53: note: in definition of macro 'includeboilerplate'
   uint16_t getType() const override { return QType::RNAME; }

File: pdns/recursordist/rec-tcpout.hh [hunk 1]
```diff
@@ -0,0 +1,73 @@
...
+      }
+      return "";
+    }
+
+    std::shared_ptr<TCPIOHandler> d_handler;
+    timeval d_last_used{0, 0};
+    size_t d_numqueries{0};
+  };
+
+  void store(const ComboAddress& ip, Connection& connection);
```
rgacogne, line 55: It might be better to take a rvalue ref for connection so that the caller can't hold to it? That would also prevent having to touch the atomic refcount.

pdns/pdns_recursor.cc
```diff
@@ -3734,6 +3735,7 @@ static void houseKeeping(void *)
       SyncRes::pruneThrottledServers();
       SyncRes::pruneNonResolving(now.tv_sec - SyncRes::s_nonresolvingnsthrottletime);
       Utility::gettimeofday(&last_prune, nullptr);
+      t_tcp_manager.cleanup();
```
rgacogne, line 3738: We could pass the time to the callee.

pdns/lwres.cc
```diff
@@ -343,70 +346,91 @@ LWResult::Result asyncresolve(const ComboAddress& ip, const DNSName& domain, int
...
+      while (true) {
+        // If we get a new (not re-used) TCP connection that does not
+        // work, we give up. For reused connections, we assume the
+        // peer has closed it on error, so we retry. At some point we
+        // *will* get a new connection, so this loop is not endless.
+        bool isNew = false;
+        connection = t_tcp_manager.get(ip);
+        if (!connection.d_handler) {
+          isNew = true;
+          const struct timeval timeout{ g_networkTimeoutMsec / 1000, static_cast<suseconds_t>(g_networkTimeoutMsec) % 1000 * 1000};
```
rgacogne, line 358: It might be good to move the code preparing a new connection to a separate function, for clarity?

pdns/lwres.cc
```diff
@@ -343,70 +346,91 @@ LWResult::Result asyncresolve(const ComboAddress& ip, const DNSName& domain, int
...
+            if (tlsCtx == nullptr) {
+              g_log << Logger::Error << "DoT to " << ip << " requested but not available" << endl;
+          }
+            else {
+              dnsOverTLS = true;
+            }
+          }
+          connection.d_handler = std::make_shared<TCPIOHandler>("", s.releaseHandle(), timeout, tlsCtx, now->tv_sec);
+          // Returned state ignored
+          connection.d_handler->tryConnect(SyncRes::s_tcp_fast_open_connect, ip);
```
rgacogne, line 380: Do we want to retry if that throws, because it looks like we don't?

pdns/lwres.cc
```diff
@@ -343,70 +346,91 @@ LWResult::Result asyncresolve(const ComboAddress& ip, const DNSName& domain, int
...
-      std::shared_ptr<TLSCtx> tlsCtx{nullptr};
-      if (SyncRes::s_dot_to_port_853 && ip.getPort() == 853) {
-        TLSContextParameters tlsParams;
-        tlsParams.d_provider = "openssl";
-        tlsParams.d_validateCertificates = false;
-        //tlsParams.d_caStore = caaStore;
-        tlsCtx = getTLSContext(tlsParams);
-        if (tlsCtx == nullptr) {
-          g_log << Logger::Error << "DoT to " << ip << " requested but not available" << endl;
+      while (true) {
```
rgacogne, line 349: I wonder if we could turn the body of that loop into a function, which would not care about whether the connection is new and the loop would handle the retry. Of course that means creating a new function handling the part common to the UDP and TCP path but it might not be a bad thing anyway?

pdns/recursordist/rec-tcpout.hh
```diff
@@ -0,0 +1,73 @@
...
+  {
+    return d_idle_connections.size();
+  }
+  uint64_t* getSize() const
+  {
+    return new uint64_t(size());
+  }
+
+private:
+  std::multimap<ComboAddress, Connection> d_idle_connections;
```
rgacogne, line 69: One nit but this will probably not work well if we later want to use SNI and have different connections to the same host using different names. That's fine for now, though.

omoerbeek: Yep, I'll make a not of that in a comment.

pdns/lwres.cc
omoerbeek: Latest commit does refactoring, but not as far as you suggest. That is for a next step I'd say.

pdns/recursordist/rec-tcpout.hh
```diff
@@ -0,0 +1,75 @@
...
+#pragma once
+
+#include "iputils.hh"
+#include "tcpiohandler.hh"
+
+class TCPOutConnectionManager
+{
+public:
+  // Max idle time for a connection, 0 is no timeout
+  static struct timeval maxIdleTime;
```
rgacogne, line 32: Perhaps we could prefix the static values with `s_`?
```suggestion
  static struct timeval s_maxIdleTime;
```

pdns/lwres.cc
```diff
@@ -229,10 +232,87 @@ static void logIncomingResponse(const std::shared_ptr<std::vector<std::unique_pt
   }
 }
 
+static bool tcpconnect(const struct timeval& now, const ComboAddress& ip, TCPOutConnectionManager::Connection& connection, bool& dnsOverTLS)
+{
+  dnsOverTLS = SyncRes::s_dot_to_port_853 && ip.getPort() == 853;
+
+
+  while (true) {
```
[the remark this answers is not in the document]
omoerbeek: Right, the loop is a remainder of an older version.

pdns/lwres.cc
```diff
@@ -524,3 +561,19 @@ LWResult::Result asyncresolve(const ComboAddress& ip, const DNSName& domain, int
...
 }
 
+LWResult::Result asyncresolve(const ComboAddress& ip, const DNSName& domain, int type, bool doTCP, bool sendRDQuery, int EDNS0Level, struct timeval* now, boost::optional<Netmask>& srcmask, boost::optional<const ResolveContext&> context, const std::shared_ptr<std::vector<std::unique_ptr<RemoteLogger>>>& outgoingLoggers, const std::shared_ptr<std::vector<std::unique_ptr<FrameStreamLogger>>>& fstrmLoggers, const std::set<uint16_t>& exportTypes, LWResult *lwr, bool* chained)
+{
+  TCPOutConnectionManager::Connection connection;
+  auto ret = asyncresolve(ip, domain, type, doTCP, sendRDQuery, EDNS0Level, now, srcmask, context, outgoingLoggers, fstrmLoggers, exportTypes, lwr, chained, connection);
+
+  if (doTCP) {
+    if (!lwr->d_validpacket) {
+      ret = asyncresolve(ip, domain, type, doTCP, sendRDQuery, EDNS0Level, now, srcmask, context, outgoingLoggers, fstrmLoggers, exportTypes, lwr, chained, connection);
```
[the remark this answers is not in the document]
omoerbeek: I'll revisit this as well to see if it still needed. It might also be a leftover of an old version.

omoerbeek:
I'll do a new round of extensive testing with these changes.

omoerbeek:
Did the testing and found no issues.

pdns/recursordist/rec-tcpout.cc
```diff
@@ -0,0 +1,87 @@
...
+
+  if (d_idle_connections.size() >= s_maxIdlePerThread) {
+    return;
+  }
+  if (d_idle_connections.count(ip) >= s_maxIdlePerAuth) {
+    return;
+  }
+
+  ++connection.d_numqueries;
+  if (s_maxQueries > 0 && connection.d_numqueries > s_maxQueries) {
```
rgacogne, line 68: Wouldn't it be better to move that check before the checks above? Because it looks like we can be expunging valid connections even though we will decide not to reuse that one.

pdns/recursordist/rec-tcpout.cc
```diff
@@ -0,0 +1,87 @@
...
+    return;
+  }
+  gettimeofday(&connection.d_last_used, nullptr);
+  d_idle_connections.emplace(ip, connection);
```
rgacogne, line 72:
```suggestion
  d_idle_connections.emplace(ip, std::move(connection));
```

rgacogne (approved):
Looks good, two nits that do not prevent this PR from being merged IMHO.

omoerbeek:
Rebased to fix unretaled regress test failure

omoerbeek:
Ignoring the dnsdist-tsan regerss failure.

PR #10692, opened 2021-09-06
auth: added self weighted lua function

n0tlu5:
### Short description

added a new Lua function that returns pickwhashed or pickrandom(if there is no available candidate) but the weight is fetched from candidates via URL. Modified some functions to be able to provide weight as a state.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/pdns/lua-record.cc
+++ b/pdns/lua-record.cc
@@ -233,9 +233,17 @@ private:
   }
   
   void setWeight(const CheckDesc& cd, string content){
-    ReadLock lock{&d_lock};
-    auto& state = d_statuses[cd];
-    state->weight = stoi(content);
+    auto statuses = d_statuses.write_lock();
+    auto& state = (*statuses)[cd];
+    try {
+      state->weight = stoi(content);
+    } catch (const PDNSException& e) {
+      // set weight to 0
+      state->weight = 0;
+    }
+    if (state->first) {
+      state->first = false;
+    }
   }
 
   void setDown(const ComboAddress& rem, const std::string& url=std::string(), const opts_t& opts = opts_t())
--- a/pdns/lua-record.cc
+++ b/pdns/lua-record.cc
@@ -71,6 +71,8 @@ private:
     CheckState(time_t _lastAccess): lastAccess(_lastAccess) {}
     /* current status */
     std::atomic<bool> status{false};
+    /* current weight */
+    std::atomic<int> weight{0};
     /* first check ? */
     std::atomic<bool> first{true};
     /* last time the status was accessed */
@@ -83,9 +85,9 @@ public:
     d_checkerThreadStarted.clear();
   }
   ~IsUpOracle() = default;
-  bool isUp(const ComboAddress& remote, const opts_t& opts);
-  bool isUp(const ComboAddress& remote, const std::string& url, const opts_t& opts);
-  bool isUp(const CheckDesc& cd);
+  int isUp(const ComboAddress& remote, const opts_t& opts);
+  int isUp(const ComboAddress& remote, const std::string& url, const opts_t& opts);
+  int isUp(const CheckDesc& cd);
 
 private:
   void checkURL(const CheckDesc& cd, const bool status, const bool first = false)
@@ -127,8 +129,10 @@ private:
       }
 
       if(!status) {
-        g_log<<Logger::Info<<"LUA record monitoring declaring "<<remstring<<" UP for URL "<<cd.url<<"!"<<endl;
+        g_log<<Logger::Info<<"LUA record monitoring declaring "<<remstring<<" UP for URL "<<cd.url<<"!"<<" with WEIGHT "<<content<<"!"<<endl;
       }
+      
+      setWeight(cd, content);
       setUp(cd);
     }
     catch(std::exception& ne) {
@@ -219,6 +223,12 @@ private:
       state->first = false;
     }
   }
+  
+  void setWeight(const CheckDesc& cd, string content){
+    ReadLock lock{&d_lock};
+    auto& state = d_statuses[cd];
+    state->weight = stoi(content);
+  }
 
   void setDown(const ComboAddress& rem, const std::string& url=std::string(), const opts_t& opts = opts_t())
   {
@@ -244,7 +254,7 @@ private:
   }
 };
 
-bool IsUpOracle::isUp(const CheckDesc& cd)
+int IsUpOracle::isUp(const CheckDesc& cd)
 {
   if (!d_checkerThreadStarted.test_and_set()) {
     d_checkerThread = std::make_unique<std::thread>([this] { return checkThread(); });
@@ -255,6 +265,9 @@ bool IsUpOracle::isUp(const CheckDesc& cd)
     auto iter = statuses->find(cd);
     if (iter != statuses->end()) {
       iter->second->lastAccess = now;
+      if (iter->second->weight > 0) {
+        return iter->second->weight;
+      }
       return iter->second->status;
     }
   }
@@ -272,13 +285,13 @@ bool IsUpOracle::isUp(const CheckDesc& cd)
   return false;
 }
 
-bool IsUpOracle::isUp(const ComboAddress& remote, const opts_t& opts)
+int IsUpOracle::isUp(const ComboAddress& remote, const opts_t& opts)
 {
   CheckDesc cd{remote, "", opts};
   return isUp(cd);
 }
 
-bool IsUpOracle::isUp(const ComboAddress& remote, const std::string& url, const opts_t& opts)
+int IsUpOracle::isUp(const ComboAddress& remote, const std::string& url, const opts_t& opts)
 {
   CheckDesc cd{remote, url, opts};
   return isUp(cd);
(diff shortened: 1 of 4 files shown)


Habbie:
`  lua-record.cc:229:20: error: use of undeclared identifier 'd_lock'` - are you sure you compiled the version you PRed here?

pdns/lua-record.cc
```diff
@@ -128,7 +130,10 @@ class IsUpOracle
 
       if(!status) {
         g_log<<Logger::Info<<"LUA record monitoring declaring "<<remstring<<" UP for URL "<<cd.url<<"!"<<endl;
+        g_log<<Logger::Info<<"LUA record monitoring declaring "<<remstring<<" WEIGHT is "<<content<<"!"<<endl;
```
Habbie, line 133: maybe the weight can go into the existing message

Habbie (changes requested):
this makes sense at a quick glance. It'll need a minor rework because we changed how locking works. Can you add docs?

I'm unsure about the name `selfweighted` but that's a small detail. Once you have docs it'll be easier to reason about what this is and thus what we should call it.

pdns/lua-record.cc
darfisultoni: sure

n0tlu5: merged the log, added a try catch just in case the content is not a weight(int)

n0tlu5:
compiled 5f3a65f, i can provide the docker image if needed

pdns/lua-record.cc
```diff
@@ -219,6 +231,20 @@ class IsUpOracle
       state->first = false;
     }
   }
+  
+  void setWeight(const CheckDesc& cd, string content){
```
Habbie, line 235: how about `  void setWeight(const CheckDesc& cd, int weight){` ? Then the exception handling only needs to live in the other spot.

n0tlu5: pushed the changes

pdns/lua-record.cc
```diff
@@ -1152,6 +1177,38 @@ static void setupLuaRecords(LuaContext& lua) // NOLINT(readability-function-cogn
       return pickRandom<string>(items);
     });
 
+  /*
+   * Based on the hash of `bestwho`, returns an IP address from the list
+   * supplied, weighted according to the results of isUp calls.
+   * @example selfweighted("{ "192.0.2.20", "203.0.113.4", "203.0.113.2" })
+   */
+  lua.writeFunction("selfweighted", [](const iplist_t& ips,
```
Habbie, line 1185: let's call it `pickselfweighted`

pdns/lua-record.cc
```diff
@@ -126,9 +128,20 @@ class IsUpOracle
         throw std::runtime_error(boost::str(boost::format("unable to match content with `%s`") % cd.opts.at("stringmatch")));
       }
 
-      if(!status) {
-        g_log<<Logger::Info<<"LUA record monitoring declaring "<<remstring<<" UP for URL "<<cd.url<<"!"<<endl;
+      int weight = 0;
+      try {
+        weight = stoi(content);
```
Habbie, line 133: so this tries to parse the content as a number -always- and then stores it in weight if it parses? interesting. seems harmless :)

Habbie (commented):
code looks decent. Still needs documentation and a test.

pdns/lua-record.cc
```diff
@@ -255,6 +277,9 @@ bool IsUpOracle::isUp(const CheckDesc& cd)
     auto iter = statuses->find(cd);
     if (iter != statuses->end()) {
       iter->second->lastAccess = now;
+      if (iter->second->weight > 0) {
```
miodvallat, line 280: This logic means that, if an ip is found up with a non-zero weight, and then later gets marked down (`status` = `false`), it would still get returned as up with its previous weight.

You need to either check for `status` first, or reset `weight` to `0` in `setDown`.

pdns/lua-record.cc
```diff
@@ -219,6 +232,15 @@ class IsUpOracle
...
     }
   }
+  
+  void setWeight(const CheckDesc& cd, int weight){
+    auto statuses = d_statuses.write_lock();
+    auto& state = (*statuses)[cd];
+    state->weight = weight;
+    if (state->first) {
+      state->first = false;
+    }
```
miodvallat, line 242: Since this duplicates work done in `setStatus` (and thus in `setUp` and `setDown`), and `setWeight` is always followed by `setUp`, there is no need to keep these lines.
```suggestion
```

Or you might consider keeping that and adding `state->status = true`, and remove the `setUp` call following the `setWeight` call, since a non-zero weight implies being up.

n0tlu5: updated the setweight to only sets weight and setUp will updates the `first` check flag

pdns/lua-record.cc
n0tlu5: added weight reset when a candidate is down

n0tlu5:
should i resolve those failing checks?
i do not understand the "CodeQL and clang-tidy" check

miodvallat:
> should i resolve those failing checks? i do not understand the "CodeQL and clang-tidy" check

Yes, please. You need to add the new function name to the extra words list of the spell checker. The clang-tidy results are indeed not always easy to find. Here they are, for the lines changed by your PR:

```
Warning: pdns/lua-record.cc:90: parameter name 'cd' is too short, expected at least 3 characters (readability-identifier-length)
Warning: pdns/lua-record.cc:245: parameter name 'cd' is too short, expected at least 3 characters (readability-identifier-length)
Warning: pdns/lua-record.cc:275: parameter name 'cd' is too short, expected at least 3 characters (readability-identifier-length)
Warning: pdns/lua-record.cc:1196: statement should be inside braces (readability-braces-around-statements)
Warning: pdns/lua-record.cc:1200: converting integer literal to bool, use bool literal instead (modernize-use-bool-literals)
Warning: pdns/lua-record.cc:1200: implicit conversion 'int' -> bool (readability-implicit-bool-conversion)
Warning: pdns/lua-record.cc:1207: converting integer literal to bool, use bool literal instead (modernize-use-bool-literals)
Warning: pdns/lua-record.cc:1207: implicit conversion 'int' -> bool (readability-implicit-bool-conversion)
```

They should be easy to address, or you can add `//NOLINT(readability-identifier-length)` to the first three cases, to avoid having to touch too many lines of code. The other points (braces for single-statement `if` and explicit casts) need to be addressed, though.

omoerbeek:
> should i resolve those failing checks? i do not understand the "CodeQL and clang-tidy" check

You can see the reported issues in context i the "Files changed" tab.

pdns/lua-record.cc
```diff
@@ -244,17 +266,21 @@ class IsUpOracle
...
-bool IsUpOracle::isUp(const CheckDesc& cd)
+int IsUpOracle::isUp(const CheckDesc& cd)
 {
   if (!d_checkerThreadStarted.test_and_set()) {
     d_checkerThread = std::make_unique<std::thread>([this] { return checkThread(); });
   }
   time_t now = time(nullptr);
   {
     auto statuses = d_statuses.read_lock();
+    //NOLINT(readability-identifier-length)
```
miodvallat, line 277: I'd rather put that annotation on line 269 where the parameter is seen first.

n0tlu5:
the statement inside braces check confuses me

n0tlu5:
the check still fails

miodvallat:
Please use `NOLINTNEXTLINE` instead of `NOLINT` when the comment concerns the following line. That ought to appease `clang-tidy`.

n0tlu5:
i ran `clang-tidy` before pushing the commit, it didn't show the warn at line 246, but there is a warn for line 238

```
~/D/p/pdns sulton/self-weighted• 28.1s | 1 1 ❱ docker run --rm --entrypoint bash -v $(pwd):/app pdns-clang-tidy -c "clang-tidy -checks=* pdns/lua-record.cc -export-fixes=clang-tidy-local.yml | grep 'is too short'"
Error while trying to load a compilation database:
Could not auto-detect compilation database for file "pdns/lua-record.cc"
No compilation database found in /app/pdns or any parent directory
fixed-compilation-database: Error while opening fixed database: No such file or directory
json-compilation-database: Error while opening JSON database: No such file or directory
Running without flags.
244847 warnings and 1 error generated.
Error while processing /app/pdns/lua-record.cc.
/app/pdns/lua-record.cc:94:34: warning: parameter name 'cd' is too short, expected at least 3 characters [readability-identifier-length]
/app/pdns/lua-record.cc:110:16: warning: variable name 'mc' is too short, expected at least 3 characters [readability-identifier-length]
/app/pdns/lua-record.cc:155:34: warning: parameter name 'cd' is too short, expected at least 3 characters [readability-identifier-length]
/app/pdns/lua-record.cc:161:14: warning: variable name 's' is too short, expected at least 3 characters [readability-identifier-length]
/app/pdns/lua-record.cc:228:35: warning: parameter name 'cd' is too short, expected at least 3 characters [readability-identifier-length]
/app/pdns/lua-record.cc:238:35: warning: parameter name 'cd' is too short, expected at least 3 characters [readability-identifier-length]
/app/pdns/lua-record.cc:253:15: warning: variable name 'cd' is too short, expected at least 3 characters [readability-identifier-length]
/app/pdns/lua-record.cc:258:33: warning: parameter name 'cd' is too short, expected at least 3 characters [readability-identifier-length]
/app/pdns/lua-record.cc:263:31: warning: parameter name 'cd' is too short, expected at least 3 characters [readability-identifier-length]
/app/pdns/lua-record.cc:303:13: warning: variable name 'cd' is too short, expected at least 3 characters [readability-identifier-length]
/app/pdns/lua-record.cc:309:13: warning: variable name 'cd' is too short, expected at least 3 characters [readability-identifier-length]
...
```

```
~/D/p/pdns sulton/self-weighted• 28.3s ❱ cat Dockerfile
FROM ubuntu:latest

# Install dependencies (same as before)
RUN apt-get update && apt-get install -y \
    clang \
    clang-tidy \
    python3 \
    git \
    build-essential \
    g++ libboost-all-dev libtool make pkg-config default-libmysqlclient-dev libssl-dev libluajit-5.1-dev python3-venv \
    autoconf automake ragel bison flex \
    && rm -rf /var/lib/apt/lists/*

RUN apt-get update && apt-get install -y \
    libcurl4-openssl-dev luajit lua-yaml-dev libyaml-cpp-dev libtolua-dev lua5.3 autoconf automake ragel bison flex g++ libboost-all-dev libtool make pkg-config libssl-dev lua-yaml-dev libyaml-cpp-dev libluajit-5.1-dev libcurl4 gawk libsqlite3-dev python3-venv \
    && rm -rf /var/lib/apt/lists/*

# Set working directory
WORKDIR /app

# Copy your project
COPY . /app

# Create build directory and run cmake
RUN autoreconf -vi && \
    ./configure --with-modules="bind lua2"
```

miodvallat:
I have taken the liberty of pushing a commit which ought to appease `clang-tidy` in order to be able to move forward. I hope you don't mind.

miodvallat (approved):

Habbie (approved):

PR #10703, opened 2021-09-08
dnsdist: Add debug messages to understand the "Advanced: includeDirectory()" failure on CI

rgacogne:
### Short description

Hopefully it will fail for this PR as well.

### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdist-console.cc
+++ b/pdns/dnsdist-console.cc
@@ -242,7 +242,7 @@ void doClient(ComboAddress server, const std::string& command)
     sendMessageToServer(fd, command, readingNonce, writingNonce, false);
 
     close(fd);
-    return; 
+    return;
   }
 
   string histfile = historyFile();
@@ -268,7 +268,7 @@ void doClient(ComboAddress server, const std::string& command)
     }
     lastline=line;
     free(sline);
-    
+
     if(line=="quit")
       break;
     if(line=="help" || line=="?")
@@ -310,7 +310,7 @@ void doConsole()
     }
     lastline=line;
     free(sline);
-    
+
     if(line=="quit")
       break;
     if(line=="help" || line=="?")
@@ -327,7 +327,7 @@ void doConsole()
         auto ret = lua->executeCode<
           boost::optional<
             boost::variant<
-              string, 
+              string,
               shared_ptr<DownstreamState>,
               ClientState*,
               std::unordered_map<string, double>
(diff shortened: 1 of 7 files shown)


Habbie:
> Hopefully it will fail for this PR as well.

It did!

omoerbeek:
With my latest changes (rebase that removed some debug code and only run advanced test):
```
  ERROR: Advanced: includeDirectory()
    File "/home/runner/work/pdns/pdns/regression-tests.dnsdist/test_Advanced.py", line 1038, in testAdvancedIncludeDirAllowed
      print(self.sendConsoleCommand('grepq("")'))
    File "/home/runner/work/pdns/pdns/regression-tests.dnsdist/dnsdisttests.py", line 698, in sendConsoleCommand
      theirNonce = sock.recv(len(ourNonce))
  socket.timeout: timed out
```
This looks like a console command now times out. Will investigate later. I might have broken something with the rebase.

omoerbeek:
I can't spot an error in the rebase. A possible explanation could be that dnsdist deadlocks somehow, but only if opendir and friends are used with tsan....????

rgacogne:
I can't spot any error either. dnsdist's logs show that the console connection was accepted but then the process seems stuck indeed. I just pushed a few debugging lines..

omoerbeek:
I have re-ran the test a many times yesterday and today and saw no failures yet. To be continued...

omoerbeek:
Another couple of runs without failure. With the debug terminal I/O the failure does not seem to happen.
I think we can conclude that the issues is with tsan and specific I/O in the GH environment.

The pragamatic solution would be: do not run test in the test with the current env and call it a day.

It would be nice if the test gets re-enabled automatically if the compiler or runtime env changes. What would be the best way to achieve that?

rgacogne:
> I think we can conclude that the issues is with tsan and specific I/O in the GH environment.

Agreed.

> The pragamatic solution would be: do not run test in the test with the current env and call it a day.

Yes, we have spent way too many man-days on this..

> It would be nice if the test gets re-enabled automatically if the compiler or runtime env changes. What would be the best way to achieve that?

Unsure, we can easily disable a test based on environment variables with unittest, but automatically detecting a change in the environment is not easy.
I'll open a PR to skip that test on GH actions for now.

rgacogne:
Closing as the test is now disabled on GH actions.

PR #10704, opened 2021-09-08
dnsdist: Stop raising the number of TCP workers to the number of TCP binds

rgacogne:
### Short description

And better document that change in the upgrade guide.

### Checklist

I have:
- [x] compiled this code
- [ ] tested this code
- [ ] documented the code


rgacogne:
At the end I did not include this change in 1.6.1, so I'm going to update this PR to say 1.7.0 instead of 1.6.1.

PR #10726, opened 2021-09-15
auth: remove unused variable last_modified from DNSResourceRecord

mind04:
### Checklist
I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


pdns/lua-base4.cc
```diff
@@ -96,7 +96,6 @@ void BaseLua4::prepareContext() {
   d_lw->registerFunction<DNSName(DNSResourceRecord::*)()>("qname", [](DNSResourceRecord& rec) { return rec.qname; });
   d_lw->registerFunction<DNSName(DNSResourceRecord::*)()>("wildcardName", [](DNSResourceRecord& rec) { return rec.wildcardname; });
   d_lw->registerFunction<string(DNSResourceRecord::*)()>("content", [](DNSResourceRecord& rec) { return rec.content; });
-  d_lw->registerFunction<time_t(DNSResourceRecord::*)()>("lastModified", [](DNSResourceRecord& rec) { return rec.last_modified; });
```
omoerbeek, line 99: This is a documented Lua API function. Code might break if we remove it. Maybe just return 0?

mind04: It is impossible to put a value in last_modified, so it is highly unlikely this function was ever used.

omoerbeek: If you say so... 
But at least the documentation should be adapted and an upgrade guide entry should be added.

PR #10727, opened 2021-09-15
SVCB: Accept known SVCParams in generic format

pieterlexis:
### Short description
This allows PowerDNS to accept SVC Params like `key4=\192\000\002\222` or `key4=abcd`. I've included tests, but am not truly happy with the implementation and would like some comments.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/pdns/rcpgenerator.cc
+++ b/pdns/rcpgenerator.cc
@@ -347,8 +347,9 @@ void RecordTextReader::xfrSvcParamKeyVals(set<SvcParam>& val)
     // We've reached a space or equals-sign or the end of the string (d_pos is at this char)
     string k = d_string.substr(pos, d_pos - pos);
     SvcParam::SvcParamKey key;
+    bool generic;
     try {
-      key = SvcParam::keyFromString(k);
+      key = SvcParam::keyFromString(k, generic);
     } catch (const std::invalid_argument &e) {
       throw RecordTextException(e.what());
     }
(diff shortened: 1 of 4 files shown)


pdns/rcpgenerator.cc
```diff
@@ -409,22 +453,33 @@ void RecordTextReader::xfrSvcParamKeyVals(set<SvcParam>& val)
     }
     case SvcParam::port: {
       uint16_t port;
-      xfr16BitInt(port);
+      if (generic) {
+        string v;
+        xfrRFC1035CharString(v);
+        port = (v.at(0) << 8);
```
rgacogne, line 459: Might be good to check the length so we can have a better error message if it's != 2.

pdns/rcpgenerator.cc
```diff
@@ -396,11 +410,41 @@ void RecordTextReader::xfrSvcParamKeyVals(set<SvcParam>& val)
     }
     case SvcParam::alpn: {
       vector<string> value;
-      xfrSVCBValueList(value);
+      if (generic) {
+        string v;
+        xfrRFC1035CharString(v);
+        size_t spos{0}, len;
+        while(spos < v.length()) {
```
rgacogne, line 417:
```suggestion
        while (spos < v.length()) {
```

rgacogne (approved):
The code looks good and the tests are great! I'm not sure there is a better way to handle the generic format :-/

PR #10734, opened 2021-09-17
dnsdist: Add experimental support for TLS asynchronous engines

rgacogne:
### Short description

Tested with the Intel QAT engine on a C3XXX, where it seems to reduce the CPU usage significantly, to a third of the CPU usage seen in normal mode.
At the moment this is only implemented for incoming DNS over TLS, but it should not be too hard to implement for outgoing DNS over TLS and outgoing DNS over HTTPS. Incoming DNS over HTTPS would be more tricky since it requires implementing the feature in libh2o as well.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code


rgacogne:
`SSL_waiting_for_async` and friends require OpenSSL >=  `1.1.0`, I'll need to handle that.

pdns/dnsdistdist/docs/advanced/tuning.rst
```diff
@@ -107,6 +107,10 @@ If the number of queued connections ("Queued" in :func:`showTCPStats`) reaches t
 
 A different possibility is that there is not enough threads accepting new connections and distributing them to worker threads. Looking at whether the ``listenOverflows`` metric in :func:`dumpStats` increase over time will tell if we are losing TCP connections because the queue is full. In that case, since a single :func:`addLocal` or :func:`addTLSLocal` directive results in only one acceptor thread, it might useful to add more of these.
 
+For incoming and outgoing DNS over TLS support, the choice of the TLS provider (OpenSSL and GnuTLS are both supported) might yield very different results depending on the exact architecture.
+
+Since 1.7.0, incoming DNS over TLS might also benefits from experimental support for TLS acceleration engines, like Intel QAT. See :func:`loadTLSEngine`, and the `tlsAsyncMode` parameter of :func:`addTLSLocal` for more information.
```
zeha, line 112:
```suggestion
Since 1.7.0, incoming DNS over TLS might also benefit from experimental support for TLS acceleration engines, like Intel QAT. See :func:`loadTLSEngine`, and the `tlsAsyncMode` parameter of :func:`addTLSLocal` for more information.
```

rgacogne:
Rebased to fix CI failures (missing promtool).

pdns/libssl.cc
```diff
@@ -702,12 +756,21 @@ std::unique_ptr<SSL_CTX, void(*)(SSL_CTX*)> libssl_init_server_context(const TLS
...
-    SSL_CTX_set_mode(ctx.get(), SSL_MODE_RELEASE_BUFFERS);
+    mode |= SSL_MODE_RELEASE_BUFFERS;
+  }
+#endif
+
+#ifdef SSL_MODE_ASYNC
+  if (config.d_asyncMode) {
+    mode |= SSL_MODE_ASYNC;
   }
 #endif
```
omoerbeek, line 770: Add an #else part to make setting `asynMode` an error/warning if not available?
Also a question: does it make sense to set `asyncMode` if no engine is set? If not, it would be nice to detect and warn about that.

pdns/dnsdistdist/docs/reference/config.rst
```diff
@@ -1564,6 +1567,17 @@ Other functions
   :param int numberOfDaysOfValidity: Number of days this OCSP response should be valid.
   :param int numberOfMinutesOfValidity: Number of minutes this OCSP response should be valid, in addition to the number of days.
 
+.. function:: loadTLSEngine(engineName [, defaultString])
+
+  .. versionadded:: 1.7.0
+
+  Load the OpenSSL engine named ``engineName``, setting the engine default string to ``defaultString`` if supplied. Engines can be used to accelerate cryptographic operations, like for example Intel QAT.
+  At the moment up to a maximum of 32 loaded engines are supported, and that support is experimental.
+  Some engines might not actually decrease, and sometimes increase, the CPU usage when the TLS asynchronous mode of OpenSSL is not enabled. To enable it see the ``tlsAsyncMode`` parameter on :func:`addTLSLocal`.
```
omoerbeek, line 1576: This double negation is confusing. How about:
```suggestion
Some engines might actually decrease performance unless the TLS asynchronous mode of OpenSSL is enabled. To enable it see the ``tlsAsyncMode`` parameter on :func:`addTLSLocal`. To enable it see the ``tlsAsyncMode`` parameter on :func:`addTLSLocal`.
```

omoerbeek (commented):
I reviewed this concentrating mostly on two questions:
- Does it build with libressl (it does)
- Does it not breakk the non-async way of doing things (it does not)

As for the async code itself: I never worked with that libssl mode, so it's new territory. I will try to look at that part in a next round of review later.

pdns/libssl.cc
```diff
@@ -115,6 +129,46 @@ void unregisterOpenSSLUser()
...
+  }
+
+  auto engines = s_engines.lock();
+  if (engines->count(engineName) > 0) {
+    return { false, "TLS engine already loaded" };
+  }
+
+  ENGINE* enginePtr = ENGINE_by_id(engineName.c_str());
+  if (enginePtr == nullptr) {
+    return { false, "unable to load TLS engine" };
```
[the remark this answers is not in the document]
rgacogne: My reasoning is that it would make more sense for the caller to do that, since it supplied the name to that function and will have to handle the return value anyway (not an exception which could bubble up).

pdns/tcpiohandler.hh
```diff
@@ -9,7 +9,8 @@
 #include "misc.hh"
 #include "noinitvector.hh"
 
-enum class IOState : uint8_t { Done, NeedRead, NeedWrite };
+/* Async is only returned for TLS connections, if OpenSSL's async mode has been enabled */
+enum class IOState : uint8_t { Done, NeedRead, NeedWrite, Async };
```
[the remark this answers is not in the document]
rgacogne: How would you like me to fix that, adding a new case to all switches, throwing an exception if the handler ever returns `IOState::Async`, since it will never happen unless someone writes the code to load an engine and enable the asynchronous mode?

pdns/libssl.cc
rgacogne: Sure, I will add the warning.
As for your second question, it does not make sense today, but I'm not sure that will not change in the future. I believe it would make sense for OpenSSL to offer more asynchronous processing mechanisms.

pdns/tcpiohandler.hh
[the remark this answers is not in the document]
omoerbeek: An exception for a "can't happen" situation is fine.

pdns/dnsdistdist/docs/reference/config.rst
jsoref: I'd suggest `degrade` instead of `decrease`

pdns/dnsdist-tcp.cc
```diff
@@ -376,10 +376,31 @@ void IncomingTCPConnectionState::terminateClientConnection()
...
+    d_handler.close();
+  }
+  else {
+    /* we might already be waiting, but we might also not because sometimes we have already been
+       notified via the descriptor, not received Async again, but the async job still exists.. */
+    auto state = shared_from_this();
+    for (const auto fd : afds) {
+      try {
+	state->d_threadData.mplexer->addReadFD(fd, handleAsyncReady, state);
+      }
```
jsoref, line 398: ditto

pdns/dnsdistdist/docs/reference/config.rst
```diff
@@ -1564,6 +1567,17 @@ Other functions
...
+
+  :param string engineName: The name of the engine to load.
+  :param string defaultString: The default string to pass to the engine. The exact value depends on the engine but represents the algorithms to register to this engine, as a list of  comma-separated keywords. For example "RSA,EC,DSA,DH,PKEY,PKEY_CRYPTO,PKEY_ASN1".
```
jsoref, line 1579:
```suggestion
  :param string defaultString: The default string to pass to the engine. The exact value depends on the engine but represents the algorithms to register with the engine, as a list of comma-separated keywords. For example "RSA,EC,DSA,DH,PKEY,PKEY_CRYPTO,PKEY_ASN1".
```

rgacogne:
I rebased that PR on master to fix the conflicts then applied the suggestions made during code review.

Habbie (approved):

rgacogne:
@omoerbeek Do you want to take a new look at this PR before I merge it? d10c3054e526d06a372bac9a8493c1460e0a2bb5 in particular affects the recursor.

omoerbeek (approved):

PR #10740, opened 2021-09-21
Fix "Large object passed by value" issues as spoted by CodeQL

omoerbeek:
Plus a sign issue in a test as detected by OpenBSD/clang.
I simplified the hostog.hh template since it is used in only one place (dnsscope)
and the (optinal) argument is not used there anyway.

### Checklist

I have:
- [X] compiled this code
- [ ] tested this code
- [ ] documented the code


omoerbeek:
cf #10741

pdns/rpzloader.cc
```diff
@@ -361,7 +361,7 @@ static bool dumpZoneToDisk(const shared_ptr<Logr::Logger>& plogger, const DNSNam
   return true;
 }
 
-void RPZIXFRTracker(const std::vector<ComboAddress>& primaries, boost::optional<DNSFilterEngine::Policy> defpol, bool defpolOverrideLocal, uint32_t maxTTL, size_t zoneIdx, const TSIGTriplet& tt, size_t maxReceivedBytes, const ComboAddress& localAddress, const uint16_t axfrTimeout, const uint32_t refreshFromConf, std::shared_ptr<SOARecordContent> sr, std::string dumpZoneFileName, uint64_t configGeneration)
+void RPZIXFRTracker(const std::vector<ComboAddress>& primaries, const boost::optional<DNSFilterEngine::Policy>& defpol, bool defpolOverrideLocal, uint32_t maxTTL, size_t zoneIdx, const TSIGTriplet& tt, size_t maxReceivedBytes, const ComboAddress& localAddress, const uint16_t axfrTimeout, const uint32_t refreshFromConf, std::shared_ptr<SOARecordContent> sr, const std::string& dumpZoneFileName, uint64_t configGeneration)
```
rgacogne, line 364: Did you check the lifetime of these parameters? I'm worried about the case where this function is called from `reload-lua-config`.

rgacogne: Oops, you mentioned that already.

pdns/rec-lua-conf.cc
```diff
@@ -626,6 +626,8 @@ void startLuaConfigDelayedThreads(const luaConfigDelayedThreads& delayedThreads,
 {
   for (const auto& rpzPrimary : delayedThreads.rpzPrimaryThreads) {
     try {
+      // The get calls all return a value object here. That is esential, since we want copies so that RPZIXFRTracker gets values
```
rgacogne, line 629:
```suggestion
      // The get calls all return a value object here. That is essential, since we want copies so that RPZIXFRTracker gets values
```

rgacogne (approved):

Issue #7014, opened 2018-10-02
rec: Wipe cache on NOTIFY


phonedph1:
- Program: Recursor
 - Issue type: Feature request

### Short description
Lots of users used to run with the auth w/ recursor world and have now converted to a standalone auth/recursor setup using ```forward-zones```. It is a common occurrence that they want updates to be reflected as fast as possible without needing to rely on short TTLs. 

I somewhat jokingly suggested that recursor should be ```NOTIFY``` aware, but this idea seems to have some real benefits.

### Usecase
Allow automatic clearing cached data, allowing quicker updates without needing to rely on short TTLs or other tricks. Allows use in multi-platform solutions that may be able to easily send an additional notify, but not perform an API call.

### Description
On receiving a ```NOTIFY``` from a white-listed source (similar to ```allow-from```) purge all cached data for the notified zone name (and sub-tree).

There are security and performance considerations to be thought about, but I suspect most users who would use this feature will be doing so on a local trusted network.

phonedph1:
As a POC this "seems" to work for basic UDP with no access restrictions.

```
diff --git a/pdns/pdns_recursor.cc b/pdns/pdns_recursor.cc
index 9444cdb2f..838b1ffb7 100644
--- a/pdns/pdns_recursor.cc
+++ b/pdns/pdns_recursor.cc
@@ -1989,7 +1989,7 @@ static string* doProcessUDPQuestion(const std::string& question, const ComboAddr
     */
 #endif

-    if(needECS || needXPF || (t_pdl && (t_pdl->d_gettag || t_pdl->d_gettag_ffi))) {
+    if(needECS || needXPF || (t_pdl && (t_pdl->d_gettag || t_pdl->d_gettag_ffi)) || dh->opcode == Opcode::Notify) {
       try {
         EDNSOptionViewMap ednsOptions;
         bool xpfFound = false;
@@ -2094,6 +2094,17 @@ static string* doProcessUDPQuestion(const std::string& question, const ComboAddr
       g_stats.avgLatencyOursUsec=(1-1.0/g_latencyStatSize)*g_stats.avgLatencyOursUsec + 0.0; // we assume 0 usec
       return 0;
     }
+
+    // not a cache hit
+    if (dh->opcode == Opcode::Notify) {
[...]

PR #10751, opened 2021-09-22
rec: Add support for NOTIFY queries to wipe cache entries

kpfleming:
### Short description
Add support for NOTIFY queries clearing cached entries in Recursor (issue #7014).

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code


kpfleming:
For what it's worth, I now have this deployed on my network, two recursors getting NOTIFY messages from one auth server, and it is working as planned.

kpfleming:
There is one 'issue' that I know of: since the NOTIFY message will probably be for an SOA RR, and the recursor does a resolution for that record after wiping the caches, and it receives the NOTIFY at the same time as the auth servers for the zone... it will likely resolve the SOA to the same version that was already in the cache and return it as the NOTIFY response. This is *not* placed into the packet cache, but will be placed into the normal record cache, so subsequent queries for that SOA won't return the updated value which triggered the wipe in the first place.

I doubt it's worth trying to address this.

hlindqvist:
From my user perspective, I don't think that it makes sense to introduce this setting with any networks allowed by default. 
It does seem like a feature that can be useful in some scenarios, but if an operator needs this, surely they know which hosts they want to trust with the power of wiping the cache?
It seems like this has the potential to be disruptive if abused, and potentially the whole user base could be in these default networks.

It's probably also worth pointing out in the documentation that the risk of abuse via spoofing has to be considered depending on where the allowed hosts/networks are in the network topology. Ideally one would only want to allow signed NOTIFY for this purpose, I think.

It may also be worth considering a `forward-zones`-style value for the allow list, ie specifying the names together with the allowed networks to limit the power of this (so that they for example cannot wipe `.` even if they are trusted with wiping `foo.example.com`).

pdns/recursordist/docs/settings.rst
```diff
@@ -58,6 +58,35 @@ Note that specifying an IP address without a netmask uses an implicit netmask of
...
 
+.. _setting-allow-notify-from:
+
+``allow-notify-from``
+---------------------
+-  IP addresses or netmasks, separated by commas
+-  Default: 127.0.0.0/8, 10.0.0.0/8, 100.64.0.0/10, 169.254.0.0/16, 192.168.0.0/16, 172.16.0.0/12, ::1/128, fc00::/7, fe80::/10
+
+Netmasks (both IPv4 and IPv6) that are allowed to issue NOTIFY queries to the server.
+The default allows access only from :rfc:`1918` private IP addresses.
```
hlindqvist, line 69: This statement is inconsistent with the default value listed a few lines above. 
("Only rfc1918" would be exactly "10.0.0.0/8, 172.16.0.0/12, 192.168.0.0/16", but the stated default value includes a number of other networks as well)

I think it'd be better to have an empty default acl, but if this is kept, I think this mismatch needs to be resolved.

kpfleming: It is incorrect, but was copied from the existing docs for `allow-from` (and also appears in other places in PowerDNS docs). However, I'm going to change the default to 'no addresses allowed' as you've suggested.

kpfleming:
I think it would make sense to have an `allow-notify-for` setting that limited the zones for which NOTIFY-wipes are allowed, and that could have a setting which allowed it to be populated from the `forward-zones` setting. Future work :-)

kpfleming:
After discussion in #powerdns today I've agreed to add `allow-notify-for` functionality, and link `forward-zones-file` to it so that the (likely) most common configuration will both simple and secure.

omoerbeek:
@kpfleming If this is ready for review please mark it. Otherwise I'm going to assume it's not ready yet.

kpfleming:
It is not ready yet, but I will be completing the `allow-notify-for` feature tomorrow and then adding it to this branch, as was briefly discussed in IRC. I'll mark this as ready for review once the complete set of code passes CI checks.

omoerbeek:
Thanks for the PS. I plan to do a review soon.

omoerbeek:
Can you fix the conflicts? Thanks!

omoerbeek:
> Thanks for the initial review; I'll dive into the code again on Friday and clean up the problems.
> 
> Regarding the need to issue a cache-wipe request 'up' to a distributor thread: there are thread-local caches that need to be wiped, and a worker thread can't distribute commands to other worker threads.

I will try to see if there is a better point in the incoming request path to do this.

omoerbeek:
OK, looking at the code again I think the spot where you send the request to the handler is OK. I also had doubts if calling SyncRes was needed, but it is since we need to answer with the right SOA record.

So if you could rebase this I can do the final review.

Once more remark: is it really needed to create a SyncRes object and process the notify (SOA) query through it?

pdns/pdns_recursor.cc
```diff
@@ -2897,50 +2929,99 @@ static void handleRunningTCPQuestion(int fd, FDMultiplexer::funcparam_t& var)
         ++g_stats.qcounter;
         ++g_stats.tcpqcounter;
 
+        if(dc->d_mdp.d_header.opcode == Opcode::Notify) {
+          if(t_allowNotifyFrom && !t_allowNotifyFrom->match(dc->d_source)) {
```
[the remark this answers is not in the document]
kpfleming: Correct, this was an oversight in the change from 'default open' to 'default closed.' Will fix in both locations.

pdns/pdns_recursor.cc
```diff
@@ -3369,10 +3485,10 @@ static void handleNewUDPQuestion(int fd, FDMultiplexer::funcparam_t& var)
             g_log<<Logger::Error<<"Ignoring answer from "<<fromaddr.toString()<<" on server socket!"<<endl;
           }
         }
-        else if(dh->opcode) {
+        else if(dh->opcode && dh->opcode != Opcode::Notify) {
```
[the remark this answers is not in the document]
kpfleming: Fixed.

pdns/pdns_recursor.cc
```diff
@@ -2876,10 +2908,10 @@ static void handleRunningTCPQuestion(int fd, FDMultiplexer::funcparam_t& var)
         terminateTCPConnection(fd);
         return;
       }
-      if (dc->d_mdp.d_header.opcode) {
+      if (dc->d_mdp.d_header.opcode && dc->d_mdp.d_header.opcode != Opcode::Notify) {
```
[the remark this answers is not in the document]
kpfleming: Fixed.

pdns/pdns_recursor.cc
```diff
@@ -201,9 +203,11 @@ enum class PaddingMode { Always, PaddedQueries };
 
 static listenSocketsAddresses_t g_listenSocketsAddresses; // is shared across all threads right now
 static set<int> g_fromtosockets; // listen sockets that use 'sendfromto()' mechanism (without actually using sendfromto())
-static std::atomic<uint32_t> counter;
```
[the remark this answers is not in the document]
kpfleming: Fixed.

pdns/pdns_recursor.cc
```diff
@@ -2897,50 +2929,99 @@ static void handleRunningTCPQuestion(int fd, FDMultiplexer::funcparam_t& var)
...
+        if(dc->d_mdp.d_header.opcode == Opcode::Query) {
+          /* It might seem like a good idea to skip the packet cache lookup if we know that the answer is not cacheable,
+             but it means that the hash would not be computed. If some script decides at a later time to mark back the answer
+             as cacheable we would cache it with a wrong tag, so better safe than sorry. */
+	  dc->d_eventTrace.add(RecEventTrace::PCacheCheck);
+          bool cacheHit = checkForCacheHit(qnameParsed, dc->d_tag, conn->data, qname, qtype, qclass, g_now, response, dc->d_qhash, pbData, true, dc->d_source);
+	  dc->d_eventTrace.add(RecEventTrace::PCacheCheck, cacheHit, false);
+
+          if (cacheHit) {
+            if (t_protobufServers && dc->d_logResponse && !(luaconfsLocal->protobufExportConfig.taggedOnly && pbData && !pbData->d_tagged)) {
```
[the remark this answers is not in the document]
kpfleming: Yes, that appears to be the case. Removed.

kpfleming:
> Once more remark: is it really needed to create a SyncRes object and process the notify (SOA) query through it?

I'm not sure I understand the question here; I don't think the new code is doing that, but I could easily be wrong :)

Another comment: the new code doesn't feed anything to protobuf servers or participate in the new eventTrace system.

omoerbeek:
> > Once more remark: is it really needed to create a SyncRes object and process the notify (SOA) query through it?
> 
> I'm not sure I understand the question here; I don't think the new code is doing that, but I could easily be wrong :)

Both the TCP and UDP case create an mthread executing `startDoResolve`. That function creates an `SyncRes` object. Likely after the `dq` object is initialized we can go for `sendit` like the code is already doing for other cases. Now the code calls `sr.beginResolve` to resolve the SOA record for the qname, as can also be seen in the trace. Will check in more detail tomorrow.

> 
> Another comment: the new code doesn't feed anything to protobuf servers or participate in the new eventTrace system.
I can live with that.

kpfleming:
Ahh, I didn't write that part; it's the fall-through path for all responses for normal queries. A NOTIFY operation is treated as if it was a normal SOA query after the cache-wipe has been requested.

pdns/reczones.cc
```diff
@@ -459,19 +468,27 @@ std::shared_ptr<SyncRes::domainmap_t> parseAuthAndForwards()
...
+        if (instructions.empty()) { // empty line
+          continue;
+        }
         throw PDNSException("Error parsing line " + std::to_string(linenum) + " of " + ::arg()["forward-zones-file"]);
       }
 
+      bool allowNotifyFor = false;
+
+      for (;; domain.erase(0, 1)) {
+        switch (domain[0]) {
```
omoerbeek, line 481: Is there a `domain[0]` after the erase? Often at() is preferred since it has bounds-checking.
Additionally, I can see various malformed lines go undetected:
```
+^
+^=
+^;=
```
likely there are more cases. The old code is more strict.

omoerbeek (changes requested):
In general this looks good! I do have a few remarks that should be look at though.

omoerbeek:
Also: as this PR is quite large I'd like another review.

pdns/pdns_recursor.cc
```diff
@@ -1687,7 +1712,7 @@ static void startDoResolve(void *p)
     checkFrameStreamExport(luaconfsLocal);
 #endif
 
-    DNSPacketWriter pw(packet, dc->d_mdp.d_qname, dc->d_mdp.d_qtype, dc->d_mdp.d_qclass);
+    DNSPacketWriter pw(packet, dc->d_mdp.d_qname, dc->d_mdp.d_qtype, dc->d_mdp.d_qclass, dc->d_mdp.d_header.opcode);
```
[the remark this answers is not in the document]
kpfleming: Well, the caches will have just been emptied for that zone, so I presume the SOA record won't be available without performing the lookup.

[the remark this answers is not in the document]
omoerbeek: A notify reply is supposed to be a copy of the request with the QR bit set. So no need to actually do a lookup. See https://datatracker.ietf.org/doc/html/rfc1996 section 4.7

pdns/reczones.cc
kpfleming: Good point.... I'll re-do this part.

kpfleming: Next update will have this fixed, although as best I can tell the old code would have accepted `+;=`, at least through the basic parsing stage.

pdns/pdns_recursor.cc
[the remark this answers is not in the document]
kpfleming: Applied.

pdns/recursordist/RECURSOR-MIB.txt
```diff
@@ -1032,6 +1032,21 @@ udp6InCsumErrors OBJECT-TYPE
         "Number of UDP6 in checksum errors (Linux only)"
     ::= { stats 123 }
 
+unauthorizedNotify OBJECT-TYPE
```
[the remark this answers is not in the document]
kpfleming: Done.

pdns/pdns_recursor.cc
```diff
@@ -1596,6 +1606,21 @@ static bool answerIsNOData(uint16_t requestedType, int rcode, const std::vector<
...
+{
+  if (t_allowNotifyFor->empty()) {
+    return false;
+  }
+
+  notifyset_t::const_iterator ret;
+  do {
+    ret = t_allowNotifyFor->find(qname);
+    if (ret != t_allowNotifyFor->end())
+      break;
```
pieterlexis, line 1619: nit: you could `return true` here and `return false` after `while`

pdns/pdns_recursor.cc
```diff
@@ -1773,7 +1798,7 @@ static void startDoResolve(void *p)
     dq.extendedErrorExtra = &dc->d_extendedErrorExtra;
     dq.meta = std::move(dc->d_meta);
 
-    if(ednsExtRCode != 0) {
+    if(ednsExtRCode != 0 || dc->d_mdp.d_header.opcode == Opcode::Notify) {
```
pieterlexis, line 1801: This feels a bit late. We've been allocating and initializing a bunch of things for nothing

pdns/pdns_recursor.cc
```diff
@@ -2876,10 +2908,10 @@ static void handleRunningTCPQuestion(int fd, FDMultiplexer::funcparam_t& var)
         terminateTCPConnection(fd);
         return;
       }
-      if (dc->d_mdp.d_header.opcode) {
+      if (dc->d_mdp.d_header.opcode != Opcode::Query && dc->d_mdp.d_header.opcode != Opcode::Notify) {
         g_stats.ignoredCount++;
         if (g_logCommonErrors) {
-          g_log<<Logger::Error<<"Ignoring non-query opcode from TCP client "<< dc->getRemote() <<" on server socket!"<<endl;
+          g_log<<Logger::Error<<"Ignoring unsupported opcode from TCP client "<< dc->getRemote() <<" on server socket!"<<endl;
```
pieterlexis, line 2914: It would be good log the actual opcode here

pdns/pdns_recursor.cc
```diff
@@ -2897,50 +2929,94 @@ static void handleRunningTCPQuestion(int fd, FDMultiplexer::funcparam_t& var)
...
-        dc->d_eventTrace.add(RecEventTrace::PCacheCheck);
-        bool cacheHit = checkForCacheHit(qnameParsed, dc->d_tag, conn->data, qname, qtype, qclass, g_now, response, dc->d_qhash, pbData, true, dc->d_source);
-        dc->d_eventTrace.add(RecEventTrace::PCacheCheck, cacheHit, false);
-
-        if (cacheHit) {
+        if(dc->d_mdp.d_header.opcode == Opcode::Query) {
+          /* It might seem like a good idea to skip the packet cache lookup if we know that the answer is not cacheable,
+             but it means that the hash would not be computed. If some script decides at a later time to mark back the answer
+             as cacheable we would cache it with a wrong tag, so better safe than sorry. */
+	  dc->d_eventTrace.add(RecEventTrace::PCacheCheck);
```
pieterlexis, line 2959: indent issue

pdns/pdns_recursor.cc
```diff
@@ -2897,50 +2929,94 @@ static void handleRunningTCPQuestion(int fd, FDMultiplexer::funcparam_t& var)
...
+          bool cacheHit = checkForCacheHit(qnameParsed, dc->d_tag, conn->data, qname, qtype, qclass, g_now, response, dc->d_qhash, pbData, true, dc->d_source);
+	  dc->d_eventTrace.add(RecEventTrace::PCacheCheck, cacheHit, false);
```
pieterlexis, line 2961: indent issue

pdns/pdns_recursor.cc
```diff
@@ -2897,50 +2929,94 @@ static void handleRunningTCPQuestion(int fd, FDMultiplexer::funcparam_t& var)
...
+              g_log<<Logger::Notice<<t_id<< " TCP question answered from packet cache tag="<<dc->d_tag<<" from "<<dc->d_source.toStringWithPort()<<(dc->d_source != dc->d_remote ? " (via "+dc->d_remote.toStringWithPort()+")" : "")<<endl;
+            }
 
+            bool hadError = sendResponseOverTCP(dc, response);
+            finishTCPReply(dc, hadError, false);
+            struct timeval now;
+            Utility::gettimeofday(&now, nullptr);
+            uint64_t spentUsec = uSec(now - start);
+            g_stats.cumulativeAnswers(spentUsec);
+	    dc->d_eventTrace.add(RecEventTrace::AnswerSent);
```
pieterlexis, line 2974: indent issue

pdns/pdns_recursor.cc
```diff
@@ -2897,50 +2929,94 @@ static void handleRunningTCPQuestion(int fd, FDMultiplexer::funcparam_t& var)
...
+	    if (dc->d_eventTrace.enabled() && SyncRes::s_event_trace_enabled & SyncRes::event_trace_to_log) {
+	      g_log << Logger::Info << dc->d_eventTrace.toString() << endl;
+	    }
+          } // cache hit
+        } // query opcode
+
+        if(dc->d_mdp.d_header.opcode == Opcode::Notify) {
           if (!g_quiet) {
-            g_log<<Logger::Notice<<t_id<< " TCP question answered from packet cache tag="<<dc->d_tag<<" from "<<dc->d_source.toStringWithPort()<<(dc->d_source != dc->d_remote ? " (via "+dc->d_remote.toStringWithPort()+")" : "")<<endl;
+            g_log<<Logger::Notice<<t_id<< " got NOTIFY for "<<qname.toLogString()<<" from "<<dc->d_source.toStringWithPort()<<(dc->d_source != dc->d_remote ? " (via "+dc->d_remote.toStringWithPort()+")" : "")<<endl;
```
pieterlexis, line 2989: I'm not sure if using `d_remote` is correct. @Habbie might chime in wrt `getInnerRemote` and `getRealRemote`.

pdns/pdns_recursor.cc
```diff
@@ -2897,50 +2929,94 @@ static void handleRunningTCPQuestion(int fd, FDMultiplexer::funcparam_t& var)
...
 
-          if (dc->d_eventTrace.enabled() && SyncRes::s_event_trace_enabled & SyncRes::event_trace_to_log) {
-            g_log << Logger::Info << dc->d_eventTrace.toString() << endl;
+            unixDie("write to thread pipe returned wrong size or error");
           }
+
+          // the operation will now be treated as a Query, generating
+          // a normal response, as the rest of the code does not
+          // check dh->opcode, but we need to ensure that the response
+          // to this request does not get put into the packet cache
```
pieterlexis, line 3006: Not a huge fan of the implication this comment makes. We might need to rethink the code structure (in a different PR)

pdns/pdns_recursor.cc
```diff
@@ -3205,13 +3284,45 @@ static string* doProcessUDPQuestion(const std::string& question, const ComboAddr
...
+      delete tmsg;
+
+      unixDie("write to thread pipe returned wrong size or error");
+    }
+
+    // the operation will now be treated as a Query, generating
+    // a normal response, as the rest of the code does not
+    // check dh->opcode, but we need to ensure that the response
+    // to this request does not get put into the packet cache
+    variable = true;
```
pieterlexis, line 3323: it might be good to have a function to this, as the TCP code is almost identical.

pdns/pdns_recursor.cc
```diff
@@ -3369,10 +3480,10 @@ static void handleNewUDPQuestion(int fd, FDMultiplexer::funcparam_t& var)
             g_log<<Logger::Error<<"Ignoring answer from "<<fromaddr.toString()<<" on server socket!"<<endl;
           }
         }
-        else if(dh->opcode) {
+        else if(dh->opcode != Opcode::Query && dh->opcode != Opcode::Notify) {
           g_stats.ignoredCount++;
           if(g_logCommonErrors) {
-            g_log<<Logger::Error<<"Ignoring non-query opcode "<<dh->opcode<<" from "<<fromaddr.toString()<<" on server socket!"<<endl;
+            g_log<<Logger::Error<<"Ignoring unsupported opcode "<<dh->opcode<<" from "<<fromaddr.toString()<<" on server socket!"<<endl;
```
pieterlexis, line 3486: Please log the opcode

pdns/pdns_recursor.cc
```diff
@@ -4601,15 +4722,72 @@ static void checkOrFixFDS()
...
+      pos=line.find('#');
+      if(pos!=string::npos)
+        line.resize(pos);
+      boost::trim(line);
+      if(line.empty())
+        continue;
+
+      result->addMask(line);
+    }
+    g_log<<Logger::Warning<<"Done parsing "<<result->size()<<" "<<aclSetting<<" ranges from file '"<<::arg()[aclFile]<<"' - overriding '"<<aclSetting<<"' setting"<<endl;
```
pieterlexis, line 4768: `Logger::Info` might be better?

pdns/pdns_recursor.cc
```diff
@@ -4633,61 +4816,45 @@ void parseACLs()
...
+  broadcastFunction([=]{ return pleaseSupplantAllowFrom(allowFrom); });
   oldAllowFrom = nullptr;
 
+  auto oldAllowNotifyFrom = t_allowNotifyFrom;
+
+  auto allowNotifyFrom = parseACL("allow-notify-from-file", "allow-notify-from");
+
+  g_initialAllowNotifyFrom = allowNotifyFrom;
+  broadcastFunction([=]{ return pleaseSupplantAllowNotifyFrom(allowNotifyFrom); });
+  oldAllowNotifyFrom = nullptr;
```
pieterlexis, line 4856: `oldAllowNotifyFrom` seems unused

pdns/recursordist/RECURSOR-MIB.txt
```diff
@@ -1032,6 +1035,21 @@ udp6InCsumErrors OBJECT-TYPE
...
+    DESCRIPTION
+        "Number of unauthorized NOTIFY operations"
+    ::= { stats 124 }
+
+disallowedNotify OBJECT-TYPE
+    SYNTAX Counter64
+    MAX-ACCESS read-only
+    STATUS current
+    DESCRIPTION
+        "Number of disallowed NOTIFY operations"
```
pieterlexis, line 1051: please mention if this about the domain or IP

pdns/recursordist/RECURSOR-MIB.txt
```diff
@@ -1032,6 +1035,21 @@ udp6InCsumErrors OBJECT-TYPE
         "Number of UDP6 in checksum errors (Linux only)"
     ::= { stats 123 }
 
+unauthorizedNotify OBJECT-TYPE
+    SYNTAX Counter64
+    MAX-ACCESS read-only
+    STATUS current
+    DESCRIPTION
+        "Number of unauthorized NOTIFY operations"
```
pieterlexis, line 1043: please mention if this about the domain or IP

pdns/recursordist/docs/settings.rst
```diff
@@ -58,6 +58,69 @@ Note that specifying an IP address without a netmask uses an implicit netmask of
...
+``allow-notify-for``
+---------------------
+.. versionadded:: 4.6.0
+
+-  Comma separated list of domain-names
+-  Default: (empty)
+
+Domain names specified in this list are used to permit incoming
+NOTIFY operations to wipe any cache entries that match the domain
+name.
```
pieterlexis, line 72: what does empty mean? Use only the IP ACL, or disallow all?

pdns/recursordist/docs/settings.rst
```diff
@@ -58,6 +58,69 @@ Note that specifying an IP address without a netmask uses an implicit netmask of
...
+
+Note that specifying an IP address without a netmask uses an implicit
+netmask of /32 or /128.
+
+NOTIFY operations received from a client listed in one of these netmasks
+will be accepted and used to wipe any cache entries whose zones match
+the zone specified in the NOTIFY operation, but only if that zone (or
+one of its parents) is included in `allow-notify-for`_,
+`allow-notify-for-file`_, or `forward-zones-file_` with a '^' prefix.
+
```
pieterlexis, line 111: please mention what empty does? Can no-one notify or do we accept notifies for names in the allow-list from any IP?

pieterlexis (changes requested):
Cool stuff, I have a bunch of comments.

I would also like some tests that actually check if the ACLs are working and counters are increased correctly.

pdns/pdns_recursor.cc
omoerbeek: yes, but doing it earlier will make the compiler complain about skipping of initializations of vars used later

pieterlexis: we need to kill the goto

omoerbeek: yes, this whole function needs a rewrite... but not as part of this PR

pdns/pdns_recursor.cc
kpfleming: This (and the other indent issues) display properly in Emacs and other tools on my system, but not in the GitHub viewer... and it's because the lines have mixed tabs/spaces for indentation. This is why I brought that topic up on the IRC channel yesterday... if we're OK with updating the .clang-format file to disable use of tabs for indentation, I'll reformat this branch.

pdns/pdns_recursor.cc
kpfleming: Nice! Done.

pdns/pdns_recursor.cc
kpfleming: Fixed.

pdns/pdns_recursor.cc
kpfleming: Fixed.

pdns/pdns_recursor.cc
kpfleming: Fixed.

pdns/pdns_recursor.cc
kpfleming: it was, but not as a string, but will be now :-)

pdns/pdns_recursor.cc
kpfleming: Agreed, it was this way in the original code but I wasn't sure that `Warning` was appropriate.

pdns/pdns_recursor.cc
kpfleming: Indeed, and this pattern was duplicated from the block above for `allow-from`. The only assumption I can make is that this ensures that the object's destructor is called in *this* thread and not in any of the worker threads who are replacing their copies of the `shared_ptr`. If that's not the reason, then the reason is unclear, or lost to the mists of time.

pdns/recursordist/RECURSOR-MIB.txt
kpfleming: Done, but this points out a potential issue. The two new configuration settings are `allow-notify-for` and `allow-notify-from`, but the metrics use `disallowed` and 'unauthorized` for those two cases. This is not consistent.

Possibly the metrics could be named "sourceDisallowedNotify" and "zoneDisallowedNotify".

pdns/recursordist/RECURSOR-MIB.txt
kpfleming: Done, pending discussion in comments about the sibling metric.

pdns/recursordist/docs/settings.rst
kpfleming: This is addressed in the first paragraph of this section. If that's not sufficient, let me know.

pdns/recursordist/docs/settings.rst
kpfleming: Done.

pdns/pdns_recursor.cc
kpfleming: Agreed, done.

pdns/recursordist/docs/settings.rst
omoerbeek: I'd say it's enough.

pdns/recursordist/RECURSOR-MIB.txt
omoerbeek: Yes, I would very much like metric names that reflect the settings names.

pdns/pdns_recursor.cc
omoerbeek: The original code did object allocation, replacement and delete by hand: https://github.com/PowerDNS/pdns/commit/f26bf5471ce15c1e826a8a3250954c2bc26e7370 . That code needed to save the old pointer to be able to delete the old object.

The extra shared pointer ref in the current code indeed causesthe object pointed to by  `t_allowFrom` is only deleted after the `nullptr` assignment in this thread. I don't think this is really needed anymore, but OTOH it does not hurt a lot.

omoerbeek:
@kpfleming you marked most review comments by @pieterlexis as resolved, but I do not see the corresponding code changes. Did you forget to push the latest changes or do I have lack of caffeine?

pdns/recursordist/RECURSOR-MIB.txt
kpfleming: Will do.

kpfleming:
I avoided pushing because it would be a force-push and I was afraid that GitHub would lose the open conversations. Now that the metric name discussion has been resolved I'll make those changes and push the results.

kpfleming:
Also to answer @pieterlexis other request for tests; I'll be happy to work on that in a followup PR, as I don't think there are examples of those types of tests already in the tree and so it will take some time to develop them.

pdns/pdns_recursor.cc
kpfleming: I've removed these.

kpfleming:
OK, the remaining two review comments appear to be about issues larger than this PR, so hopefully this is in good shape now.

omoerbeek (approved):
Only one small nit left.

omoerbeek:
Although this lacks good test coverage, I'm going to merge. We would very much like to have more tests of the functionality added.

omoerbeek (approved):

pieterlexis (approved):
let's do it, but ensure we have tests before a final release

PR #10753, opened 2021-09-22
Add our various random implementations to speedtest

omoerbeek:
### Short description

Fixes #6492 (or at least gives us some data).

### Checklist

I have:
- [X] read the [CONTRIBUTING.md](https://github.com/PowerDNS/pdns/blob/master/CONTRIBUTING.md) document
- [X] compiled this code
- [X] tested this code
- [ ] included documentation (including possible behaviour changes)
- [ ] documented the code
- [ ] added or modified regression test(s)
- [X] added or modified unit test(s)

On old bullseye laptop:
```
'Random test sodium' 0.38 seconds: 1847879.0 runs/s, 0.54 usec/run
'Random test openssl' 0.12 seconds: 580113.7 runs/s, 1.72 usec/run
'Random test getrandom' 0.34 seconds: 1883117.4 runs/s, 0.53 usec/run
'Random test urandom' 0.51 seconds: 1649074.8 runs/s, 0.61 usec/run
```
Random buster VM:
```
'Random test sodium' 0.48 seconds: 1252499.3 runs/s, 0.80 usec/run
'Random test openssl' 0.12 seconds: 332932.7 runs/s, 3.00 usec/run
'Random test getrandom' 0.54 seconds: 1280666.6 runs/s, 0.78 usec/run
'Random test urandom' 0.45 seconds: 1115434.9 runs/s, 0.90 usec/run
```

on OpenBSD (pretty beefy AMD):
```
'Random test sodium' 0.10 seconds: 35829910.5 runs/s, 0.03 usec/run
'Random test openssl' 0.10 seconds: 30367527.4 runs/s, 0.03 usec/run
'Random test arc4random' 0.10 seconds: 60572491.4 runs/s, 0.02 usec/run
'Random test urandom' 1.26 seconds: 1781943.5 runs/s, 0.56 usec/run
```

All in all I think the current auto order of getrandom, arc4random, libsodium, openssl, urandom is ok.

PR #10755, opened 2021-09-23
speedtest: Count the system time as well as the user time

rgacogne:
### Short description

Our tests are designed to run for 100 ms of CPU time, but this is currently measured with `ITIMER_VIRTUAL` which means only when the process is executing in userspace. Switching to `ITIMER_PROF` would also account for the time spent
when the system is running on behalf of the process, which seems closer to what we want.

It fixes the reported time often exceeding 100ms in the numbers reported for `sodium`, `getrandom` and `urandom` in #10753, as they spend most of their time in kernel space.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/speedtest.cc
+++ b/pdns/speedtest.cc
@@ -36,8 +36,8 @@ template<typename C> void doRun(const C& cmd, int mseconds=100)
   it.it_interval.tv_sec=0;
   it.it_interval.tv_usec=0;
 
-  signal(SIGVTALRM, alarmHandler);
-  setitimer(ITIMER_VIRTUAL, &it, 0);
+  signal(SIGPROF, alarmHandler);
+  setitimer(ITIMER_PROF, &it, 0);
 
   unsigned int runs=0;
   g_stop=false;


Habbie:
> POSIX.1-2008  marks  getitimer()  and       setitimer() obsolete, recommending the use of the POSIX timers API (timer_gettime(2), timer_settime(2),       etc.) instead.

I know this is not what you're changing here, but should we worry about this?

Habbie (approved):

rgacogne:
> I know this is not what you're changing here, but should we worry about this?

Definitely, although I'm quite sad that the recommended API seems much more complicated to use than the existing one..

rgacogne:
After reading the documentation I feel inclined to ignore the issue for now..

Habbie:
> After reading the documentation I feel inclined to ignore the issue for now..

Works for me. It's only speedtest anyway, so no big pain if a new distro breaks it.

omoerbeek:
> > After reading the documentation I feel inclined to ignore the issue for now..
> 
> Works for me. It's only speedtest anyway, so no big pain if a new distro breaks it.

Agreed.

omoerbeek (approved):

PR #10760, opened 2021-09-24
Don't create file with wide permissions 

omoerbeek:
As spotted by CodeQL. This file potentially holds confidential info, so be strict here.

### Checklist

I have:
- [X] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/libssl.cc
+++ b/pdns/libssl.cc
@@ -782,7 +782,7 @@ static void libssl_key_log_file_callback(const SSL* ssl, const char* line)
 std::unique_ptr<FILE, int(*)(FILE*)> libssl_set_key_log_file(std::unique_ptr<SSL_CTX, void(*)(SSL_CTX*)>& ctx, const std::string& logFile)
 {
 #ifdef HAVE_SSL_CTX_SET_KEYLOG_CALLBACK
-  int fd = open(logFile.c_str(),  O_WRONLY | O_CREAT, 0600);
+  int fd = open(logFile.c_str(),  O_WRONLY | O_CREAT | O_APPEND, 0600);
   if (fd == -1) {
     unixDie("Error opening TLS log file '" + logFile + "'");
   }
--- a/pdns/libssl.cc
+++ b/pdns/libssl.cc
@@ -15,6 +15,7 @@
 #include <openssl/ocsp.h>
 #include <openssl/rand.h>
 #include <openssl/ssl.h>
+#include <fcntl.h>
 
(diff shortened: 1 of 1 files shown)


pdns/libssl.cc
```diff
@@ -779,8 +780,13 @@ static void libssl_key_log_file_callback(const SSL* ssl, const char* line)
 std::unique_ptr<FILE, int(*)(FILE*)> libssl_set_key_log_file(std::unique_ptr<SSL_CTX, void(*)(SSL_CTX*)>& ctx, const std::string& logFile)
 {
 #ifdef HAVE_SSL_CTX_SET_KEYLOG_CALLBACK
-  auto fp = std::unique_ptr<FILE, int(*)(FILE*)>(fopen(logFile.c_str(), "a"), fclose);
+  int fd = open(logFile.c_str(),  O_WRONLY | O_CREAT, 0600);
+  if (fd == -1) {
+    throw std::runtime_error("Error opening TLS log file '" + logFile + "'");
```
Habbie, line 785: if you use `unixDie` from misc.h, you get a nice error string.

Habbie (approved):
looks fine, but I suggested one improvement

Habbie (approved):

pdns/libssl.cc
```diff
@@ -779,9 +782,15 @@ static void libssl_key_log_file_callback(const SSL* ssl, const char* line)
 std::unique_ptr<FILE, int(*)(FILE*)> libssl_set_key_log_file(std::unique_ptr<SSL_CTX, void(*)(SSL_CTX*)>& ctx, const std::string& logFile)
 {
 #ifdef HAVE_SSL_CTX_SET_KEYLOG_CALLBACK
-  auto fp = std::unique_ptr<FILE, int(*)(FILE*)>(fopen(logFile.c_str(), "a"), fclose);
+  int fd = open(logFile.c_str(),  O_WRONLY | O_CREAT, 0600);
```
rgacogne, line 785: I _think_ it's OK to call `open` without the `append` flag then call `fdopen` with `a` but it feel more logic that way to me:
```suggestion
  int fd = open(logFile.c_str(),  O_WRONLY | O_CREAT | O_APPEND, 0600);
```

rgacogne (approved):

pdns/libssl.cc
omoerbeek: Yes, the original is correct but indeed is a bit unclear.

PR #10768, opened 2021-09-28
rec: Use the correct RPZ policy name when loading via XFR

rgacogne:
### Short description

This commit fixes two issues:
- if the existing zone name is not empty we should use it, instead of the zone domain
- if the zone domain has to be used, it should not include a final dot

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


omoerbeek (changes requested):
This works ok to get the stats right, but would there be any possible surpsies when we loose the dot we had before? In that light I would be hesitant about backporting and we should add a note in the upgrade guide.

rgacogne:
`polName` seems to be exclusively used for the stats in that function so I'm not worried.

omoerbeek (approved):
Indeed. My worries were not needed.

PR #10770, opened 2021-09-28
auth: Gracefully handle uncaught exceptions in the UDP path

rgacogne:
### Short description

These exceptions should never bubble up to this point, but if it does it makes no sense to terminate the whole process because of it.
This commit logs a message at error level and moves on to the next query, like we do in the TCP path.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/common_startup.cc
+++ b/pdns/common_startup.cc
@@ -414,11 +414,16 @@ static void sendout(std::unique_ptr<DNSPacket>& a)
 {
   if(!a)
     return;
-  
-  N->send(*a);
 
-  int diff=a->d_dt.udiff();
-  avg_latency=0.999*avg_latency+0.001*diff;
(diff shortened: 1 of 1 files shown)


omoerbeek (approved):
A bit strange to see "uncaught exception" in a `catch` block. Maybe change the message into: "Caught unhandled exception..." ? OK either way.

Habbie:
> A bit strange to see "uncaught exception" in a `catch` block. Maybe change the message into: "Caught unhandled exception..." ? OK either way.

Fully agreed. 'Caught unhandled ...' would be nice. Otherwise, lgtm.

rgacogne:
Done!

Habbie (approved):

PR #10776, opened 2021-09-29
docs: generate requirements.txt with package hashes

Habbie:
### Short description
DRAFT. GitHub URLs should move to /PowerDNS/ and one of them can be replaced with a normal pip line now. GitHub URLs should probably point at commit hashes instead of branch heads.

### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code


docs/requirements.txt
```diff
@@ -1,7 +1,258 @@
...
+    --hash=sha256:c4fdf4019605b6e5423637e01bc9fe4daef873709a7973e195ceba0a62bbc844
+    # via requests
+zipp==3.5.1 \
+    --hash=sha256:1fc9641b26f3bd81069b7738b039f2819cab6e3fc3399a953e19d92cc81eff4d \
+    --hash=sha256:8dc6c4d5a809d659067cc713f76bcf42fae8ae641db12fddfa93694a15abc96b
+    # via importlib-metadata
+
+# WARNING: The following packages were not pinned, but pip requires them to be
+# pinned when the requirements file includes hashes. Consider using the --allow-unsafe flag.
+# setuptools
```
rgacogne, line 258: Is this something we need to worry about?

Habbie: It's a terribly confusing story. `--allow-unsafe` will become the default in the next version of `pip-tools` because ... that's the safe choice.

In this specific case, we never install setuptools (`python3 -mvenv` does that, from the distro version I believe, which I will check), so there indeed is nothing to worry about.

rgacogne: Good, thank you :)

Habbie: For context, this is what `--allow-unsafe` does:
```
-# WARNING: The following packages were not pinned, but pip requires them to be
-# pinned when the requirements file includes hashes. Consider using the --allow-unsafe flag.
-# setuptools
+# The following packages are considered to be unsafe in a requirements file:
+setuptools==58.1.0 \
+    --hash=sha256:5de67252090e08d25f240f07d80310f778a5a46cdcf9ea9855662630ac8547b2 \
+    --hash=sha256:7324fd4b66efa05cdfc9c89174573a4410acc7848f318cc0565c7fb659dfdc81
+    # via sphinx
```

so depending on whether we use the flag, setuptools is not pulled in, or it is pinned.

pieterlexis:
Instructions on how to update the pin-file would be nice to have in the requirements.in

Habbie:
> Instructions on how to update the pin-file would be nice to have in the requirements.in

That's a good point. I put the instructions in the .txt, but step 1 destroys those (and there are currently two steps).

Habbie:
I moved the instructions.

Habbie:
To check that nobody ran `pip install` separately, `pip-sync -n` can check things. For now, this reports:
```
Would uninstall:
  changelog
  sphinx-jsondomain
  sphinxcontrib-openapi
Would install:
  changelog @ https://github.com/pieterlexis/sphinx-changelog/archive/refs/heads/render-tags.zip
  sphinx-jsondomain @ https://github.com/pieterlexis/sphinx-jsondomain/archive/refs/heads/no-type-links.zip
  sphinxcontrib-openapi @ https://github.com/Habbie/sphinxcontrib-openapi/archive/refs/heads/use-jsondomain-pdns.zip
```

Need to figure out if we can help it be smarter. Alternatively, we might prefer upstreaming these things.

Additionally, for `pip-sync` to work, `pip-tools` needs to be installed into the venv (and thus into every venv).

Habbie:
@pieterlexis suspects that we can switch to upstream `changelog` at least.

.github/actions/spell-check/excludes.txt
```diff
@@ -8,7 +8,7 @@
 (?:^|/)m4/
 (?:^|/)package(?:-lock|)\.json$
 (?:^|/)package-lock\.json$
-(?:^|/)requirements\.txt$
+(?:^|/)requirements\.(txt|in)$
```
jsoref, line 11:
```suggestion
(?:^|/)requirements\.(?:txt|in)$
```

rgacogne (approved):

PR #10786, opened 2021-10-01
Tell the operator they are notifying a native zone

pieterlexis:
### Short description
When using `pdns_control notify [ZONE]`, PowerDNS now returns some extra
text if the zone being notified is a native zone. As notifying a native
zone makes little sense and the operator probably wants to change the
zone kind to master/primary.

### Checklist
I have:
- [x] compiled this code
- [ ] tested this code
- [ ] documented the code


mind04:
This is not helping, it is only adding more confusion. Notifying native zones is a valid use case for people who want to control the process themself.

pieterlexis:
> This is not helping, it is only adding more confusion. Notifying native zones is a valid use case for people who want to control the process themself.

This is why the text is written as a suggestion. Anyone doing notifications via `pdns_control` using other processes can safely ignore the message. There is no change in the behaviour of PowerDNS, we still queue the notify regardless of the zone-kind.

omoerbeek (approved):

pieterlexis:
I updated the text somewhat to quell the confusion.

Issue #10585, opened 2021-07-12
dnsdist: feature to skip arbitrary EDNS options in hashing 


phonedph1:
- Program: dnsdist
 - Issue type: Feature request

### Short description
Ignoring some other option codes beyond cookies can greatly improve packet cache hits.

### Usecase
We have some devices that send 65001/65074 but do not care about their contents. Ignoring these increases normal packet cache hits considerably. 

### Description
A more generic way to set which options you want to skip, possibly as options sent to the packet cache.

```
-      if (optionCode == EDNSOptionCode::COOKIE) {
+      if (optionCode == EDNSOptionCode::COOKIE || optionCode == 65001 || optionCode == 65074) {
```

rgacogne:
Makes sense to me, I guess `PacketCache::canHashPacket` and `PacketCache::hashAfterQname` could take a `const std::unordered_set<uint16_t>& optionsToSkip` parameter instead of the single boolean. A quick and dirty test suggests that it would not be detrimental to performance.

PR #10791, opened 2021-10-01
dnsdist: allow skipping arbitrary EDNS options when computing packet hash

chbruyand:
### Short description

This adds a new option to `newPacketCache()` to skip arbitrary EDNS options when hashing packets.
I was a bit unsure to keep the `cookieHashing` setting but this makes the lua configuration more readable.

It also affects recursor's packet cache so must be reviewed carrefully.

Fix https://github.com/PowerDNS/pdns/issues/10585

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdist-cache.hh
+++ b/pdns/dnsdist-cache.hh
@@ -28,6 +28,7 @@
 #include "lock.hh"
 #include "noinitvector.hh"
 #include "stat_t.hh"
+#include "ednsoptions.hh"
 
 struct DNSQuestion;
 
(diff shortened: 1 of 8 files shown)


pdns/dnsdist-cache.hh
```diff
@@ -144,5 +143,5 @@ private:
   bool d_deferrableInsertLock;
   bool d_parseECS;
   bool d_keepStaleData{false};
-  bool d_cookieHashing{false};
+  std::unordered_set<uint16_t> d_optionsToSkip{EDNSOptionCode::COOKIE};
```
rgacogne, line 146: Mostly my OCD acting up but I'd prefer moving this right after the `d_shards` one :) We don't really care about the size of that struct here but it might also prevent the need for padding.

pdns/dnsdistdist/dnsdist-lua-bindings-packetcache.cc
```diff
@@ -90,6 +93,11 @@ void setupLuaBindingsPacketCache(LuaContext& luaCtx, bool client)
         if (vars->count("cookieHashing")) {
           cookieHashing = boost::get<bool>((*vars)["cookieHashing"]);
```
rgacogne, line 94: I wonder if it wouldn't be cleaner to insert into `optionsToSkip` and get rid of `DNSDistPacketCache::setCookieHashing` instead? It does not seem to be used anywhere else, except in `fuzz_dnsdistcache.cc` where we will need to use the new options anyway.

rgacogne (approved):
Two small nits that would not prevent the merge of this PR anyway. Great work, thanks!

PR #10797, opened 2021-10-05
rec: Use packetcache-servfail-ttl for all packet cache entries considered an error reply

### LINKED ISSUE #9135: rec: Packet cache caches empty negative responses for packetcache-ttl
- Program: Recursor
 - Issue type: Bug report

### Short description
If a negative response has no `SOA` record -- because the authoritative server is buggy and didn't include one, or it included one with an inappropriate owner name and the Recursor filtered it -- the regular cache (what's it called? record cache? main cache?) won't cache it at all. However, the packet cache will cache it for `packetcache-ttl` seconds.

I assume this affects nodata responses, but I only checked NXDOMAIN.

### Environment
 - Software version: 4.4.0~alpha1+master.119.g5e71c4330-1pdns.xenial

### Steps to reproduce
1. `dig +nocookie -x 2604:a880:400:d0::8a0:a002`
2. Wait a little while
3. `dig +nocookie -x 2604:a880:400:d0::8a0:a002`

(The authoritative servers' response includes an `SOA` record for `4.0.6.2.ip6.arpa`, so it gets filtered.)

### Expected behaviour
Packet cached for <= 60 seconds.

### Actual behaviour
Packet cached for 3600 seconds.

### Other information
@rgacogne explained the details and said:

> ... we do cache an entry for as long as the lowest between the smallest TTL in the answer or `packetcache-ttl`. In that case we have no records in the answer so we use `packetcache-ttl`, which defaults to 3600s. I guess we could do something smarter here, from refusing to cache an answer with no records (which is what dnsdist does) to using a smaller TTL (perhaps `packetcache-servfail-ttl` would make sense).

I agree. :P 

See also #5040.

omoerbeek:
The "this is an error" test is: no answer records and rcode not equal to NoError. This needs to be validated.
Fixes #9135.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/pdns_recursor.cc
+++ b/pdns/pdns_recursor.cc
@@ -2299,8 +2299,11 @@ static void startDoResolve(void *p)
       g_stats.variableResponses++;
     }
     if (!SyncRes::s_nopacketcache && !variableAnswer && !sr.wasVariable()) {
-      minTTL = min(minTTL, pw.getHeader()->rcode == RCode::ServFail ? SyncRes::s_packetcacheservfailttl :
-                   SyncRes::s_packetcachettl);
+      const auto& hdr = pw.getHeader();
+      if (hdr->ancount == 0 && hdr->rcode != RCode::NoError) {
+        minTTL = min(minTTL, SyncRes::s_packetcacheservfailttl);
+      }
+      minTTL = min(minTTL, SyncRes::s_packetcachettl);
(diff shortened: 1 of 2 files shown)


rgacogne:
Do we want to apply the ServFail TTL to NXDomains?

omoerbeek:
> Do we want to apply the ServFail TTL to NXDomains?

That's  one case I'm not sure about. It will put some load on the record cache (and potentially aggressive NSEC cache)...

rgacogne:
I don't think we should, that could potentially be bad under PRSD attacks.

PR #10798, opened 2021-10-05
Recursor: update powerdns-example-script.lua

gmelikov:
`print()` won't output properly, use `pdnslog()`.

We already use `pdnslog()` on first line of this example.

In addition - clean up spaces/empty lines.

Maybe something was wrong on my side, but I didn't see output from `print()` anywhere even with loglevel=9.

### Checklist

I have:
- [ ] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/recursordist/contrib/powerdns-example-script.lua
+++ b/pdns/recursordist/contrib/powerdns-example-script.lua
@@ -18,37 +18,37 @@ magicMetric = getMetric("magic")
 -- return false to say you did not take over the question, but we'll still listen to 'variable'
 -- to selectively disable the cache
 function preresolve(dq)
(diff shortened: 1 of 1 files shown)


omoerbeek:
Running with `daemon=no` in a terminal does work with `print()`. I suspect that running under `systemd`  causes the stdout stream to end up in an unexpected place. So `pdnslog` is likely better indeed.

gmelikov:
@omoerbeek Thank you for info, indeed I did use default systemd service.

omoerbeek (changes requested):
There are more whitespace issues in this file. Can fix those as well? Preferably we do ws changes in a seperate commit.

gmelikov:
@omoerbeek no problem, done.

omoerbeek (approved):

PR #10814, opened 2021-10-06
dnsdist: add lua support for SetEDNSOptionAction

chbruyand:
### Short description

This adds support for lua `SetEDNSOptionAction`. This allows to set arbitrary EDNS option and content to the query.
Fix https://github.com/PowerDNS/pdns/issues/3688

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdist-lua-actions.cc
+++ b/pdns/dnsdist-lua-actions.cc
@@ -927,6 +927,8 @@ public:
     auto& data = dq->getMutableData();
     if (generateOptRR(optRData, data, dq->getMaximumSize(), g_EdnsUDPPayloadSize, 0, false)) {
       dq->getHeader()->arcount = htons(1);
+      // make sure that any EDNS sent by the backend is removed before forwarding the response to the client
+      dq->ednsAdded = true;
     }
 
     return Action::None;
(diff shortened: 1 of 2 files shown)


pdns/dnsdist-lua-actions.cc
```diff
@@ -939,6 +939,41 @@ class SetMacAddrAction : public DNSAction
...
+{
+public:
+  // this action does not stop the processing
+  SetEDNSOptionAction(uint16_t code, const std::string& data) : d_code(code), d_data(data)
+  {}
+
+  DNSAction::Action operator()(DNSQuestion* dq, std::string* ruleresult) const override
+  {
+    if (dq->getHeader()->arcount) {
+      return Action::None;
```
rgacogne, line 953: I wonder if we could easily refactor `slowRewriteQueryWithRecords` (`dnsdist-ecs.cc`) to handle that case as well?

rgacogne (changes requested):
Nice work, thanks! I think we need to fix the `dq.ednsAdded` part before merging, and perhaps add a test for that. The other suggestion might not be easy, and even if it is I'm of course okay with postponing that for a different PR.

PR #10815, opened 2021-10-06
dnsdist: Add range support for dynamic blocks

### LINKED ISSUE #4993: Count dynamic block exceed...() statistics by Netmask rather than individual IP address
- Program: dnsdist 
 - Issue type: Feature request

### Short description
The exceedXXX() functions used for triggering dynamic blocks return a set of ComboAddresses containing the IP addresses that matched the rule. However they only count based on individual IP addresses, (i.e. /32 for v4 and /128 for v6). It would be nice if they had the capability to count based on a netmask, e.g. exceedServFails(rate, seconds, v4mask, v6mask)

An example would be: exceedServFailsMask(20, 10, 32, 64)

Which would count v4 addresses by individual IP, but v6 addresses by /64 aggregation. 

The return value from these functions should probably be a set of Netmasks or a NetmaskGroup rather than a set of ComboAddress's, otherwise the set of ComboAddress's returned would be potentially huge.

### Usecase
Count groups of IP addresses rather than individual IPs addresses when creating dynamic blocks.

### Description
See above.

rgacogne:
### Short description

This PR makes it possible to group queries by subnets, for example taking a whole /64 IPv6 range into account instead of a   single address.
It also adds the ability to take the IPv4 UDP and TCP ports into account, for CGNAT deployments, by setting the number of bits of the port to consider.

Fixes #4993.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdist-lua-inspection.cc
+++ b/pdns/dnsdist-lua-inspection.cc
@@ -798,6 +798,15 @@ void setupLuaInspection(LuaContext& luaCtx)
     });
   luaCtx.registerFunction<void(std::shared_ptr<DynBlockRulesGroup>::*)(uint8_t, uint8_t, uint8_t)>("setMasks", [](std::shared_ptr<DynBlockRulesGroup>& group, uint8_t v4, uint8_t v6, uint8_t port) {
       if (group) {
+        if (v4 > 32) {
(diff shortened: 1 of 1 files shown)


rgacogne:
Aaaaaaaaand rebased again after merging #10870.

Issue #10829, opened 2021-10-08
auth: backend creation before chroot() is causing startup failures


swegener:
Commit d5385b3570b0d8c13626dcb4cd8fe5fc344db16d from issue #9464 creates backend instances before complete initialization like chroot()+chdir() and dropping privileges is done.
This causes backends that open files, like the gsqlite3 backend, to fail during startup.

This seems to be some unintended side-effect of the implementation.

Habbie:
I'll be honest, I thought we didn't even have `chroot` any more. Can you tell us why you are using it?

PR #10830, opened 2021-10-08
auth: Update zone cache after initialization

swegener:
The code in `receiver.cc` creates backends before complete initialization with chroot() is done, causing startup failures for backends that require files from inside of the chroot.

Closes: https://github.com/PowerDNS/pdns/issues/10829

### Checklist
I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/common_startup.cc
+++ b/pdns/common_startup.cc
@@ -704,6 +704,17 @@ void mainthread()
 
   pdns::parseTrustedNotificationProxy(::arg()["trusted-notification-proxy"]);
 
+  try {
+    UeberBackend B;
+    B.updateZoneCache();
+  }
+  catch(PDNSException &e) {
+    g_log<<Logger::Error<<"PDNSException while updating zone cache: "<<e.reason<<endl;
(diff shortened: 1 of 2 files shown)


pdns/common_startup.cc
```diff
@@ -697,6 +697,12 @@ void mainthread()
 
   pdns::parseTrustedNotificationProxy(::arg()["trusted-notification-proxy"]);
 
+  try {
+    UeberBackend B;
+    B.updateZoneCache();
+  }
+  catch(...) {}
```
Habbie, line 704: that's a big catchall

swegener: We can catch the same exceptions as in the loop and log them, but should we exit during the initial refresh?

Habbie:
As we are close to the 4.6 release, I'd rather not refactor startup order now. How about postponing -this- to 4.7, and teaching 4.5 and 4.6 to error out on chroot+zone-cache?

PR #9998, opened 2021-01-21
dnsdist: aligned atomics

rgacogne:
### Short description

Rebased version of #9455. I also squashed the commits a bit so that every commit compiles.

### Checklist

I have:
- [x] read the [CONTRIBUTING.md](https://github.com/PowerDNS/pdns/blob/master/CONTRIBUTING.md) document
- [x] compiled this code
- [x] tested this code
- [ ] included documentation (including possible behaviour changes)
- [ ] documented the code
- [ ] added or modified regression test(s)
- [ ] added or modified unit test(s)

PR #10836, opened 2021-10-11
rec: Prevent false sharing for the recursor's metrics

rgacogne:
### Short description

We did not get any negative feedback after implementing that in dnsdist in #9998, quite the contrary since several users reported a slight decrease in CPU usage after moving from 1.5.x to 1.6.0. Other improvements could explain that decrease but at the very least we did not experience any drawback from that change, so I think it would be worth considering it in the recursor as well.

### Checklist

I have:
- [x] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/stat_t.hh
+++ b/pdns/stat_t.hh
@@ -21,6 +21,8 @@
  */
 #pragma once
 
+#include <atomic>
+
 #define CPU_LEVEL1_DCACHE_LINESIZE 64 // Until we know better via configure/getconf
 
 namespace pdns {


omoerbeek:
This looks good, though there is more:
`AtomicHistogram`, ` std::atomic<double>`, `AtomicCounter` in `syncres.hh`.
Plus more in other files as well. But it's a good start :-)

omoerbeek:
> I'm not sure `AtomicCounter` still offers anything interesting compared to `std::atomic` (or `pdns::stat_t` for that matter), by the way?

Nope, I see this as an opportunity to get rid of it.

omoerbeek:
@rgacogne Do you want to merge this and do the remaining part in another PR or add more commits to this PR?

PR #10841, opened 2021-10-13
dnsdist: Use per-thread credentials for GnuTLS client connections

rgacogne:
### Short description

It looks like there is a race in some versions when the credentials are shared between several threads opening TLS client connections, so for now we make sure these objects are not shared.
Also make sure that the `SSL_CTX` object for OpenSSL client connections is alive as long as there is still a connection using it, even if the backend is removed.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


rgacogne:
For the record, the issue in GnuTLS has been fixed in 3.7.3, see https://gitlab.com/gnutls/gnutls/-/issues/1277

PR #10845, opened 2021-10-13
dnsdist: Keep watching idle DoH backend connections

rgacogne:
### Short description

So we can quickly detect a connection closed by the remote host. The `isTCPSocketUsable()` function is unfortunately not enough if the remote end properly sent a GO AWAY frame before closing the connection, as the socket then becomes readable and usable, instead of closed. We need to actually read that message to know that the remote end closed
the connection.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


omoerbeek:
Weird indeed...

PR #10849, opened 2021-10-15
rec: Fire and forget mode for directResolve()

### LINKED ISSUE #10425: rec: for the NOD lookup, ponder a "fire and forget" mode
NOD lookups ignore the result, so it does not make sense to wait for it: waiting wastes resources in rec.
Suggested by @paddg

omoerbeek:
### Short description

An (experimental) approach to fix #10425. Not too happy about it as it adds an extra argument (and logic) in way too many functions.

As I never really understood why a  full DNS lookup would be needed to signal an NOD...

It might be better to just change the way NOD lookup works:

- a config that supplies an IP. 
- on NOD, we sent a query for the domain (with suffix) to that IP and done. 

As I understood a typical config includes a forwarder for the NOD suffix domain anyway.

Or can somebody explain why a full DNS lookup is good for a NOD?

### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/lwres.cc
+++ b/pdns/lwres.cc
@@ -283,7 +283,8 @@ static bool tcpconnect(const struct timeval& now, const ComboAddress& ip, TCPOut
 }
 
 static LWResult::Result tcpsendrecv(const ComboAddress& ip, TCPOutConnectionManager::Connection& connection,
-                                    ComboAddress& localip, const vector<uint8_t>& vpacket, size_t& len, PacketBuffer& buf)
+                                    ComboAddress& localip, const vector<uint8_t>& vpacket, size_t& len, PacketBuffer& buf,
+                                    bool faf = false)
 {
   socklen_t slen = ip.getSocklen();
   uint16_t tlen = htons(vpacket.size());
@@ -298,10 +299,9 @@ static LWResult::Result tcpsendrecv(const ComboAddress& ip, TCPOutConnectionMana
   packet.insert(packet.end(), vpacket.begin(), vpacket.end());
 
   LWResult::Result ret = asendtcp(packet, connection.d_handler);
-  if (ret != LWResult::Result::Success) {
+  if (ret != LWResult::Result::Success || faf) {
     return ret;
   }
-
   ret = arecvtcp(packet, 2, connection.d_handler, false);
   if (ret != LWResult::Result::Success) {
     return ret;
@@ -409,7 +409,7 @@ static LWResult::Result asyncresolve(const ComboAddress& ip, const DNSName& doma
 
     ret = asendto((const char*)&*vpacket.begin(), vpacket.size(), 0, ip, qid, domain, type, &queryfd);
 
-    if (ret != LWResult::Result::Success) {
+    if (ret != LWResult::Result::Success || (context && context->d_fire_and_forget)) {
       return ret;
     }
 
--- a/pdns/lwres.cc
+++ b/pdns/lwres.cc
@@ -407,9 +407,10 @@ static LWResult::Result asyncresolve(const ComboAddress& ip, const DNSName& doma
       g_stats.ipv6queries++;
     }
 
-    ret = asendto((const char*)&*vpacket.begin(), vpacket.size(), 0, ip, qid, domain, type, &queryfd);
+    const bool faf = context && context->d_fire_and_forget;
+    ret = asendto((const char*)&*vpacket.begin(), vpacket.size(), 0, ip, qid, domain, type, &queryfd, faf);
 
-    if (ret != LWResult::Result::Success || (context && context->d_fire_and_forget)) {
+    if (ret != LWResult::Result::Success || faf) {
       return ret;
     }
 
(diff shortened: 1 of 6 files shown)


neilcook:
> Or can somebody explain why a full DNS lookup is good for a NOD?

It's a good question. I wanted to use something that would require minimal configuration, and just use all the existing DNS plumbing we had in recursor. I had hoped that we would already have a fire-and-forget mechanism for that, but we didn't.

In terms of the tradeoffs between a full DNS lookup vs just sending a query to an IP address, well I guess that depends on how the latter would be implemented. The benefits of the former are that we don't have to worry about additional config, which protocols are used, or encryption, (i.e. we get all that for free) we can leverage any forwarding setup in recursor to actually perform the lookup etc. The benefits of the latter are, by the sounds of it, simpler code.

omoerbeek:
We're now at a point where we should decide to do this or not (for rec-4.6.0).
I'm still in doubt. The code is maybe not *that* bad... but is this approach something we can accept or should we encourage NOD receivers to just answer like a proper DNS server?

paddg:
> ... but is this approach something we can accept or should we encourage NOD receivers to just answer like a proper DNS server?

A NOD receiver can be operated by third parties. However, the outage of a NOD receiver can have a negative impact on resource consumption in the Recursor if it has to wait for responses.

Habbie:
How about dnsdist, near the Recursor, returning quick responses to everything, while using `TeeAction` to throw things towards the other party's NOD resolver?

paddg:
> How about dnsdist, near the Recursor, returning quick responses to everything, while using `TeeAction` to throw things towards the other party's NOD resolver?

Yes, that would be a solution too. It avoids ugly code by solving the problem with a "creative" configuration. Is that DevOps then? ;-) And what is the lesser evil? But it's definitely the first choice if this feature can't be implemented :-)

Habbie:
> Yes, that would be a solution too. It avoids ugly code by solving the problem with a "creative" configuration.

And it can be deployed today :)

> Is that DevOps then? ;-)

Sometimes I don't even know what words mean any more ;)

> And what is the lesser evil? But it's definitely the first choice if this feature can't be implemented :-)

The lesser evil depends on who you ask ;)

omoerbeek:
Yes, a `TeeAction` sounds really nice. It starts a thread in dnsdist that listen for answers, just to keep stats on incoming rcodes. The amount of resources consumed does not depend on the receiver answering or not.

johnhtodd:
This is looking more and more like something that consumes dnstap or pdns protobuf data out of dnsdist.

omoerbeek:
> This is looking more and more like something that consumes dnstap or pdns protobuf data out of dnsdist.

Or directly out of the recursor.

omoerbeek:
I'm leaning towards the conclusion to do fire-and-forget with the help of dnsdist and investigate alternative ways of NOD signalling (dnstap) in the future.
That means that this PR will be closed, it did serve a purpose, if only to get ideas.

paddg:
Thanks for your effort @omoerbeek!

neilcook:
> > This is looking more and more like something that consumes dnstap or pdns protobuf data out of dnsdist.
> 
> Or directly out of the recursor.

Indeed, this is already supported, i.e. protobuf directly out of the recursor already contains NOD information.

omoerbeek:
> > > This is looking more and more like something that consumes dnstap or pdns protobuf data out of dnsdist.
> > 
> > 
> > Or directly out of the recursor.
> 
> Indeed, this is already supported, i.e. protobuf directly out of the recursor already contains NOD information.

I think it would be a nice feature to add the possibility to only send protobuf (or dnstap for that matter) messages on NOD events.

paddg:
> > > This is looking more and more like something that consumes dnstap or pdns protobuf data out of dnsdist.
> > 
> > 
> > Or directly out of the recursor.
> 
> Indeed, this is already supported, i.e. protobuf directly out of the recursor already contains NOD information.

Does https://github.com/PowerDNS/pdns/blob/master/pdns/dnsmessage.proto already contain a NOD flag?

neilcook:
> > > > This is looking more and more like something that consumes dnstap or pdns protobuf data out of dnsdist.
> > > 
> > > 
> > > Or directly out of the recursor.
> > 
> > 
> > Indeed, this is already supported, i.e. protobuf directly out of the recursor already contains NOD information.
> 
> Does https://github.com/PowerDNS/pdns/blob/master/pdns/dnsmessage.proto already contain a NOD flag?

Yes:

`optional bool  newlyObservedDomain = 18;  `

Also, the unique domain response information is available on a per RR basis:
`optional bool  udr = 6;  `

omoerbeek:
The dnsdist rule discussed above would look something like:
```
addAction(AllRule(), TeeAction('1.2.3.4:53'))
addAction(AllRule(), RCodeAction(DNSRCode.NOERROR))
```
(if you are using a dedicated dnsdist for this fire and forget task).

omoerbeek:
Closing this. #10849 (send out protobuf and/or dnstap on NOD events) will likely be done for 4.7.0.

PR #10868, opened 2021-10-19
get rid of make_pair

neheb:
With C++17, std::pair can completely replace make_pair


neheb:
Guessing that fuzzer is bugged.

PR #10869, opened 2021-10-19
replace boost::optional with std version

neheb:
It's the same.


omoerbeek:
Lots of compile errors...

I tried this before. There *are* some differences between `boost::optional` and `std::optional`. IIRC in combination with boost's multi indexes things went wrong.

So a more step by-step approach is preferred indeed.

PR #10908, opened 2021-10-27
rec: A SHA-384 DS should not trump a SHA-256 one, so only potentially zap SHA-1

omoerbeek:
### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [X] documented the code

### DIFF
--- a/pdns/recursordist/test-syncres_cc9.cc
+++ b/pdns/recursordist/test-syncres_cc9.cc
@@ -819,9 +819,9 @@ BOOST_AUTO_TEST_CASE(test_getDSRecords_multialgo_all_sha)
   dsmap_t ds;
   auto state = sr->getDSRecords(target, ds, false, 0, false);
   BOOST_CHECK_EQUAL(state, vState::Secure);
-  BOOST_REQUIRE_EQUAL(ds.size(), 1U);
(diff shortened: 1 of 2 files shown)


vcunat:
So GOST still zaps SHA-1?  🤔

(On a quick glance.)  Not that I'd expect it to matter/happen in practice.  Still, if that's so, it would be inconsistent with the comment you added.

vcunat:
OK, but I was reading code, not just comments: https://github.com/PowerDNS/pdns/blob/b995eb5/pdns/syncres.cc#L2616

Habbie:
Indeed, dead code, even if the compiler can't figure that out.

vcunat:
Could call it "g(h)ost" code.

vcunat:
Speaking of GOST DS, I'd expect [a new one](https://datatracker.ietf.org/doc/draft-ietf-dnsop-rfc5933-bis/) in the IANA registry soon (like 2022?).  But that does not imply you implementing it.  I expect that like us, any algorithm that you add in future will be considered significantly more secure than SHA1, so it makes sense to write the code like the diff in this PR.

For reference, my reason for looking into this: https://gitlab.nic.cz/knot/knot-resolver/-/merge_requests/1251

PR #10920, opened 2021-10-28
dnsdist: Remove unreachable code in HTTP/2 connections cleanup

rgacogne:
### Short description

Reported by Coverity (CID 373724).

It also immediately removes a HTTP/2 connection from the list if it is no longer usable (TCP socket is dead).

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdist-lua.cc
+++ b/pdns/dnsdist-lua.cc
@@ -1306,11 +1306,11 @@ static void setupLuaConfig(LuaContext& luaCtx, bool client, bool configCheck)
   });
 
   luaCtx.writeFunction("setMaxCachedTCPConnectionsPerDownstream", [](size_t max) {
-    DownstreamTCPConnectionsManager::setMaxCachedConnectionsPerDownstream(max);
+    DownstreamTCPConnectionsManager::setMaxIdleConnectionsPerDownstream(max);
   });
 
-  luaCtx.writeFunction("setMaxCachedDoHConnectionsPerDownstream", [](size_t max) {
-    setDoHDownstreamMaxConnectionsPerBackend(max);
+  luaCtx.writeFunction("setMaxIdleDoHConnectionsPerDownstream", [](size_t max) {
+    setDoHDownstreamMaxIdleConnectionsPerBackend(max);
   });
 
   luaCtx.writeFunction("setOutgoingDoHWorkerThreads", [](uint64_t workers) {
--- a/pdns/dnsdist-lua.cc
+++ b/pdns/dnsdist-lua.cc
@@ -1306,7 +1306,7 @@ static void setupLuaConfig(LuaContext& luaCtx, bool client, bool configCheck)
   });
 
   luaCtx.writeFunction("setMaxCachedTCPConnectionsPerDownstream", [](size_t max) {
-    DownstreamConnectionsManager::setMaxCachedConnectionsPerDownstream(max);
+    DownstreamTCPConnectionsManager::setMaxCachedConnectionsPerDownstream(max);
   });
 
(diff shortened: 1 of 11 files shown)


rgacogne:
Refactoring done, and unit tests added (fixing a bug in the process, as always). Also fixed a race in the AXFR regression tests.

omoerbeek (dismissed):

rgacogne:
I think I have spotted an issue in the way "idle" connections are expunged when the limit is reached, it looks like 1/ we count all connections, not only the idle ones 2/ we can expunge a non-idle connection.

omoerbeek:
> I think I have spotted an issue in the way "idle" connections are expunged when the limit is reached, it looks like 1/ we count all connections, not only the idle ones 2/ we can expunge a non-idle connection.

I missed that. I suppose you are talking about the last part of `getConnectionToDownstream()` in `dnsdist-tcp-downstream.hh`. As for the counting, shouldn't idle connection be dropped by the call to `cleanupClosedConnections()` before? In that case the max is real and no new connection should be allowed I think. Which means the check should be before the creation of `newConnection`.

rgacogne:
The problem comes from the fact that before 1.7.0, outgoing TCP connections were only added to the list when they were idle, since the same connection was never used for more than one incoming client. But in 1.7.0 both DoH and DoTCP53/DoT connections can be used for several incoming clients, and are therefore always added to that list. So the setting limiting the number of idle connections (I don't it would make sense to limit the number of active connections) should no longer apply to all the connections in there, only the idle ones.
Perhaps we should have two lists, then..

rgacogne:
I pushed a new commit splitting the list in two: one for active connections, and one for idle ones. That way it's easy to count the number of idle connections and to expunge one of them when the limit is reached. It also makes it easier to try the idle connections first when we need a new one, since we are more likely to be able to write to it right away, instead of having to wait for a pending one to be sent when reusing an active connection.

rgacogne:
Needs a serious review, though. I'll do more testing on Monday.

rgacogne:
Rebased due to a conflict. Added a new unit test as well.

omoerbeek:
Seeing this on OpenBSD about 1 in 3 runs:
```
[otto@lou:172]$ ./testrunner                                                                                         
Running 134 test cases...
test-dnsdisttcp_cc.cc(217): fatal error: in "test_dnsdisttcp_cc/test_IncomingConnection_BackendNoOOOR": critical check step.request == !d_client ? ExpectedStep::ExpectedRequest::closeClient : ExpectedStep::ExpectedRequest::closeBackend has failed [connect to the backend != close connection to backend]
libc++abi: terminating with uncaught exception of type boost::execution_aborted
unknown location(0): fatal error: in "test_dnsdisttcp_cc/test_IncomingConnection_BackendNoOOOR": signal: generated by kill() (or family); uid=0; pid=0
test-dnsdisttcp_cc.cc(217): last checkpoint

*** 2 failures are detected in the test module "unit"
```

rgacogne:
So both times we decide to close a backend connection while we expected to open one, apparently. Would you be able to re-run these tests after recompiling with the `#if 0` line in `tcpiohandler-mplexer.hh` set to `#if 1`, by any chance?

rgacogne:
Thanks! Still I would like to figure out what is happening, I'm running the tests in a loop on Linux and no error so far..
I'll prepare a PR adding more debug lines..

omoerbeek (commented):
So far only small stuff seen. Will continue review Monday.

omoerbeek (approved):
This is a really nice piece of code. Apart from some nits I did not spot an issue.

PR #10929, opened 2021-10-29
Carbon regression test for rec and auth.

omoerbeek:
Copied from dnsdist one. That test uses two difference instance names, but
rec and auth settings do not allow that.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [X] documented the code


regression-tests.auth-py/test_Carbon.py
```diff
@@ -0,0 +1,175 @@
...
+            self.assertTrue(parts[1].isdigit())
+            self.assertTrue(parts[2].isdigit())
+            self.assertTrue(int(parts[2]) <= int(after))
+
+        # make sure every carbon server has received at least one connection
+        for key in self._carbonCounters:
+            value = self._carbonCounters[key]
+            self.assertTrue(value >= 1)
+
+    def testCarbonServerUp(self):
```
rgacogne, line 116: It looks like that test does not really make sense here, it tries to test the output for dnsdist's backends, which is never going to match.

regression-tests.recursor-dnssec/test_Carbon.py
```diff
@@ -0,0 +1,191 @@
...
+            self.assertTrue(parts[1].isdigit())
+            self.assertTrue(parts[2].isdigit())
+            self.assertTrue(int(parts[2]) <= int(after))
+
+        # make sure every carbon server has received at least one connection
+        for key in self._carbonCounters:
+            value = self._carbonCounters[key]
+            self.assertTrue(value >= 1)
+
+    def testCarbonServerUp(self):
```
rgacogne, line 132: Same.

rgacogne (changes requested):

regression-tests.auth-py/test_Carbon.py
omoerbeek: Thanks for spotting that. Dunno how I overlooked that.

PR #10930, opened 2021-10-29
rec: Move to a stream based socket for the control channel

omoerbeek:
### Short description

Nice consequences:

- No more need to set buffer sizes or have a large read buffer. 
- No more need to create a temporary socket for replies, a session concept is built into SOCK_STREAM sockets and we can just write the reply into the stream.

Survived basic testing here.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/pdns_recursor.cc
+++ b/pdns/pdns_recursor.cc
@@ -4206,22 +4206,20 @@ template ThreadTimes broadcastAccFunction(const boost::function<ThreadTimes*()>&
 
 static void handleRCC(int fd, FDMultiplexer::funcparam_t& var)
 {
+  int clientfd;
   try {
-    string remote;
-    string msg = s_rcc.recv(&remote).d_str;
+    clientfd = accept(fd, nullptr, nullptr);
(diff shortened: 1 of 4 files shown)


omoerbeek:
Looks like this unveovered/introduced a bug in command line parsing/passing to the recursor
```
Oct 29 14:09:43 Received rec_control command 'dump-cacheX' via controlsocket
```

dmgeurts:
Is there a way to revert/re-enable the old control protocol for 3rd party packages that haven't been updated yet? I'm specifically interested in Telegraf reporting as above.

dmgeurts:
No idea, I'm just following their documentation of how to feed Influx with PowerDNS telemetry. Using the HTTP API would mean I don't need to deploy Telegraf to my PowerDNS servers.

dmgeurts:
Found this which I'll try instead of using the unix-socket
https://github.com/influxdata/telegraf/issues/812#issuecomment-374352733

PR #10950, opened 2021-11-05
dnsdist: Add a lot more of build-time options to select features

rgacogne:
### Short description

This PR adds a lot of build-time options to select which features are built into dnsdist, to reduce the attack surface, memory usage and disk usage.

These features can now be disabled via a configure option, thus not linking against the corresponding libraries:
- cdb
- ipcipher
- libedit

For that last one, the console will still work if not disabled (see below) but no completion, history or line editing will be available.

These features can now be disabled by defining a value in `CXXFLAGS`:
- carbon (`DISABLE_CARBON`)
- completion (`DISABLE_COMPLETION`)
- protobuf (`DISABLE_PROTOBUF`)
- legacy dynamic blocks not using the new `DynBlockRulesGroup` interface (`DISABLE_DEPRECATED_DYNBLOCK`)
- recvmmsg (`DISABLE_RECVMMSG`)
- security polling (`DISABLE_SECPOLL`)
- prometheus (`DISABLE_PROMETHEUS`)
- accessing the configuration via the web interface (`DISABLE_WEB_CONFIG`)
- built-in web page (`DISABLE_BUILTIN_HTML`)
- custom Lua web handlers (`DISABLE_LUA_WEB_HANDLERS`)

Additionally several Lua bindings can be removed when they are not needed:
- `DISABLE_NON_FFI_DQ_BINDINGS`
- `DISABLE_POLICIES_BINDINGS`
- `DISABLE_DOWNSTREAM_BINDINGS`
- `DISABLE_DNSHEADER_BINDINGS`
- `DISABLE_COMBO_ADDR_BINDINGS`
- `DISABLE_DNSNAME_BINDINGS`
- `DISABLE_SUFFIX_MATCH_BINDINGS`
- `DISABLE_NETMASK_BINDINGS`
- `DISABLE_QPS_LIMITER_BINDINGS`
- `DISABLE_PACKETCACHE_BINDINGS`

It also cleans up the usage of various integer types in our Lua bindings, to prevent silent overflow from happening. Cleaning the integer types used reduces the number of template instantiations and thus the final binary size.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdistdist/docs/install.rst
+++ b/pdns/dnsdistdist/docs/install.rst
@@ -107,3 +107,36 @@ OS Specific Instructions
 ~~~~~~~~~~~~~~~~~~~~~~~~
 
 None, really.
+
+Build options
+~~~~~~~~~~~~~
+
+Our ``configure`` script provides a fair number of options with regard to which features should be enabled, as well as which libraries should be used. In addition to these options, more features can be disabled at compile-time by defining the following symbols:
+
+* ``DISABLE_BUILTIN_HTML`` removes the built-in web pages
+* ``DISABLE_CARBON`` for carbon support
+* ``DISABLE_COMPLETION`` for completion support in the console
+* ``DISABLE_DEPRECATED_DYNBLOCK`` for legacy dynamic blocks not using the new ``DynBlockRulesGroup`` interface
+* ``DISABLE_ECS_ACTIONS`` to disable actions altering EDNS Client Subnet
+* ``DISABLE_LUA_WEB_HANDLERS`` for custom Lua web handlers support
+* ``DISABLE_PROMETHEUS`` for prometheus
+* ``DISABLE_PROTOBUF`` for protocol-buffer support, including dnstap
+* ``DISABLE_RECVMMSG`` for ``recvmmsg`` support
+* ``DISABLE_RULES_ALTERING_QUERIES`` to remove rules altering the content of queries
+* ``DISABLE_SECPOLL`` for security polling
+* ``DISABLE_WEB_CONFIG`` to disable accessing the configuration via the web interface
+
+Additionally several Lua bindings can be removed when they are not needed, as they increase the memory required during compilation and the size of the final binary:
+
+* ``DDISABLE_CLIENT_STATE_BINDINGS``
+* ``DISABLE_COMBO_ADDR_BINDINGS``
+* ``DISABLE_DNSHEADER_BINDINGS``
+* ``DISABLE_DNSNAME_BINDINGS``
+* ``DISABLE_DOWNSTREAM_BINDINGS``
+* ``DISABLE_NETMASK_BINDINGS``
+* ``DISABLE_NON_FFI_DQ_BINDINGS``
+* ``DISABLE_PACKETCACHE_BINDINGS``
+* ``DISABLE_POLICIES_BINDINGS``
+* ``DISABLE_QPS_LIMITER_BINDINGS``
+* ``DISABLE_SUFFIX_MATCH_BINDINGS``
+* ``DISABLE_TOP_N_BINDINGS``
--- a/pdns/dnsdistdist/docs/install.rst
+++ b/pdns/dnsdistdist/docs/install.rst
@@ -128,7 +128,7 @@ Our ``configure`` script provides a fair number of options with regard to which 
 
 Additionally several Lua bindings can be removed when they are not needed, as they increase the memory required during compilation and the size of the final binary:
 
-* ``DDISABLE_CLIENT_STATE_BINDINGS``
+* ``DISABLE_CLIENT_STATE_BINDINGS``
 * ``DISABLE_COMBO_ADDR_BINDINGS``
 * ``DISABLE_DNSHEADER_BINDINGS``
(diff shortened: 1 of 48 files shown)


rgacogne:
Rebased.

pieterlexis:
I thing we prefer positive names (so `ENABLE_XYZ`) compared to negative ones

rgacogne:
> I thing we prefer positive names (so `ENABLE_XYZ`) compared to negative ones

Let's put aside the fact that I would have to redo quite some work to change that now, it would also mean these names need to be defined by default and undefined when we want to disable a feature, which seems very cumbersome?

pieterlexis:
It might not be worth it indeed

pdns/dnsdistdist/m4/pdns_check_libedit.m4
```diff
@@ -1,3 +1,31 @@
...
+    AS_IF([test "x$with_libedit" = "xyes" -o "x$with_libedit" = "xauto"], [
+      PKG_CHECK_MODULES([LIBEDIT], [libedit], [
+        [HAVE_LIBEDIT=1]
+        AC_DEFINE([HAVE_LIBEDIT], [1], [Define to 1 if you have libedit])
+        save_CFLAGS=$CFLAGS
+        save_LIBS=$LIBS
+        CFLAGS="$LIBEDIT_CFLAGS $CFLAGS"
+        LIBS="$LIBEDIT_LIBS $LIBS"
+        CFLAGS=$save_CFLAGS
+        LIBS=$save_LIBS
```
pieterlexis, line 20: This code has no effect

rgacogne: Indeed! I cleaned that up, thanks!

chbruyand (approved):
Would be nice to document the `DISABLE_` feature and binding options ; maybe in a simple txt file ?
Or maybe/also have a CI minimum build target with everything disabled ?

pdns/dnsdistdist/test-dnsdistrules_cc.cc
```diff
@@ -7,6 +7,14 @@
 
 #include "dnsdist-rules.hh"
 
+void checkParameterBound(const std::string& parameter, uint64_t value, size_t max);
+void checkParameterBound(const std::string& parameter, uint64_t value, size_t max)
+{
+  if (value > std::numeric_limits<uint16_t>::max()) {
```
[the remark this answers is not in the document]
rgacogne: Ah, right, we fixed that in `dnsdist-lua.cc` where it actually matters but I should fix it here as well :+1:

rgacogne:
> Would be nice to document the `DISABLE_` feature and binding options ; maybe in a simple txt file ?

Indeed, I'll add a few lines to the documentation.

> Or maybe/also have a CI minimum build target with everything disabled ?

That's a good idea but we would also need to disable the corresponding tests in the regression tests which is a bit of work, so I'll do that later.

pdns/dnsdistdist/docs/install.rst
```diff
@@ -107,3 +107,36 @@ OS Specific Instructions
...
+* ``DISABLE_PROMETHEUS`` for prometheus
+* ``DISABLE_PROTOBUF`` for protocol-buffer support, including dnstap
+* ``DISABLE_RECVMMSG`` for ``recvmmsg`` support
+* ``DISABLE_RULES_ALTERING_QUERIES`` to remove rules altering the content of queries
+* ``DISABLE_SECPOLL`` for security polling
+* ``DISABLE_WEB_CONFIG`` to disable accessing the configuration via the web interface
+
+Additionally several Lua bindings can be removed when they are not needed, as they increase the memory required during compilation and the size of the final binary:
+
+* ``DDISABLE_CLIENT_STATE_BINDINGS``
```
pieterlexis, line 131:
```suggestion
* ``DISABLE_CLIENT_STATE_BINDINGS``
```

rgacogne:
The build with the least amount of features is failing because of #11105:
```
2021-12-16T09:08:29.6307081Z /usr/bin/ld: dnsdist.o: in function `healthChecksThread()':
2021-12-16T09:08:29.6310966Z /home/runner/work/pdns/pdns/pdns/dnsdistdist/dnsdist.cc:1899: undefined reference to `handleDOHTimeout(std::unique_ptr<DOHUnit, void (*)(DOHUnit*)>&&)'
2021-12-16T09:08:29.6321519Z /usr/bin/ld: dnsdist.o: in function `processUDPQuery(ClientState&, LocalHolders&, msghdr const*, ComboAddress const&, ComboAddress&, std::vector<unsigned char, noinit_adaptor<std::allocator<unsigned char> > >&, mmsghdr*, unsigned int*, iovec*, cmsgbuf_aligned*)':
2021-12-16T09:08:29.6324552Z /home/runner/work/pdns/pdns/pdns/dnsdistdist/dnsdist.cc:1554: undefined reference to `handleDOHTimeout(std::unique_ptr<DOHUnit, void (*)(DOHUnit*)>&&)'
```

rgacogne:
Rebased.

rgacogne:
```

PR #10955, opened 2021-11-05
If possible, use SuffixMatchNodeRule() instead of RegexRule()

omoerbeek:
### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code


rgacogne (approved):

omoerbeek:
About the unrelated dnsdist tsan faiikng test:

`........................................................SF` 
the test after the previously failing `includeDir` test that is now skipped fails now. This suggest a build up of recource usage. I do notice that the number of threads of the python process grows. I suspect the UDP and RCP responder threads are never cleaned up and each test created those two. Of course I do not know if this is the root cause of the tsan test troubles we're seeing, but it never hurts to fix this.

PR #10976, opened 2021-11-10
dnsdist: Add more debug in case of (re-)connection failures

rgacogne:
### Checklist

I have:
- [x] read the [CONTRIBUTING.md](https://github.com/PowerDNS/pdns/blob/master/CONTRIBUTING.md) document
- [x] compiled this code
- [ ] tested this code
- [ ] included documentation (including possible behaviour changes)
- [ ] documented the code
- [ ] added or modified regression test(s)
- [ ] added or modified unit test(s)

PR #10977, opened 2021-11-12
dnsdist: Disable the actual connect() in the test_dnsdisttcp_cc_c unit tests.

omoerbeek:
They are not needed and cause (at least on OpenBSD) firewall state table
clashes: they remain in a embryotic state because no actual activity
occurs on them due to the rest of the tests using mockup code.

tcpiohandler.cc is not linked into the tests, so define it locally in
test-dnsdisttcp_cc.cc as well.

Followup on the diagnosis made in #10976.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdistdist/test-dnsdisttcp_cc.cc
+++ b/pdns/dnsdistdist/test-dnsdisttcp_cc.cc
@@ -41,6 +41,8 @@ GlobalStateHolder<servers_t> g_dstates;
 
 QueryCount g_qcount;
 
+const bool TCPIOHandler::s_disableConnectForUnitTests = true;
+
 bool checkDNSCryptQuery(const ClientState& cs, PacketBuffer& query, std::unique_ptr<DNSCryptQuery>& dnsCryptQuery, time_t now, bool tcp)
 {
   return false;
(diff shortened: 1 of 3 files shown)


rgacogne (approved):
Looks great, thanks Otto!

omoerbeek:
Looking at the link errors. So far I am not able to reproduce...

Habbie:
Are you trying with `  --enable-unit-tests --enable-backend-unit-tests  --enable-fuzz-targets` ?

omoerbeek:
only `--enable-backend-unit-tests` was missing, retrying

Habbie:
Hmm, that one wasn't the focus of the question ;)

omoerbeek:
This feels like ccache is playing some game here... tcpiohandler.o should contain the new symbol and *is* linked into sdig..

Habbie:
It's not ccache. Because sdig has different build settings, it also gets a separate set of `.o` files, built with those different flags. This is happening since we added DoT and I haven't been able to re-unify that. I hope this at least explains what you are seeing.

omoerbeek:
Nope I still do not understand. `sdig-tcpiohandler.o` is lined into `sdig`

PR #10980, opened 2021-11-15
rec: Return the proper ede on validation failure

### LINKED ISSUE #10936: Recursor: Wrong EDE when DNSKEY signatures have expired
- Program: Recursor
 - Issue type: Bug report

### Short description
When resolving servfail.nl|AAAA, the recursor returns a SERVFAIL with EDE code 9 (DNSKEY missing) instead of code 7 (signature expired).

The EDE code definitions in RFC 8917:

7:

> The resolver attempted to perform DNSSEC validation, but no signatures are presently valid and some (often all) are expired.

9:

> A DS record existed at a parent, but no supported matching DNSKEY record could be found for the child.

9 is wrong because the DNSKEY belonging to the DS record in the parent exists. However, because all signatures in the zone are expired (including the DNSKEY ones), see [dnsviz](https://dnsviz.net/d/servfail.nl/YX_JZA/dnssec/).

### Environment
 - Software version: master at b1f97a4cc7d2d9197fbb7b2745a255dc65860023

### Steps to reproduce
1. `./pdns_recursor --local-port=5301  --socket-dir=. --extended-resolution-errors --dnssec=validate`
2. `dig @127.0.0.1 -p5301 servfail.nl +rec +dnssec +ad +endsopt=15`

### Expected behaviour
EDE option 6 in the response.

### Actual behaviour
```
;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags: do; udp: 512
; EDE: 9 (DNSKEY Missing)
```

### Other information
The trace shows we see the right condition:

```
Nov 01 16:05:31 Signature is expired (inception: 1586339579, inception skew: 60, expiration: 1586350379, now: 1635779131)
Nov 01 16:05:31 Validation did not succeed!
```

From some discussions:

* `checkSignatureWithKey` does not distinguish between the various cases
* It looks like `validateDNSKeysAgainstDS` falls into the return `vState::BogusNoValidDNSKEY` case
* it should take into consideration why `checkSignatureWithKey` decided a sig was invalid

omoerbeek:
### Short description

Fixes #10936

Should grow a test.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [X] documented the code

### DIFF
--- a/pdns/validate.cc
+++ b/pdns/validate.cc
@@ -937,7 +937,7 @@ static bool checkSignatureWithKey(time_t now, const shared_ptr<RRSIGRecordConten
       }
     }
     else {
-      ede = (sig->d_siginception - g_signatureInceptionSkew > now) ? vState::BogusSignatureNotYetValid : vState::BogusSignatureExpired;
+      ede = ((sig->d_siginception - g_signatureInceptionSkew) > now) ? vState::BogusSignatureNotYetValid : vState::BogusSignatureExpired;
       LOG("Signature is "<<(ede == vState::BogusSignatureNotYetValid ? "not yet valid" : "expired")<<" (inception: "<<sig->d_siginception<<", inception skew: "<<g_signatureInceptionSkew<<", expiration: "<<sig->d_sigexpire<<", now: "<<now<<")"<<endl);
      }
   }
--- a/pdns/validate.cc
+++ b/pdns/validate.cc
@@ -920,7 +920,7 @@ bool isRRSIGIncepted(const time_t now, const shared_ptr<RRSIGRecordContent>& sig
   return sig->d_siginception - g_signatureInceptionSkew <= now;
 }
 
-static bool checkSignatureWithKey(time_t now, const shared_ptr<RRSIGRecordContent> sig, const shared_ptr<DNSKEYRecordContent> key, const std::string& msg)
+static bool checkSignatureWithKey(time_t now, const shared_ptr<RRSIGRecordContent> sig, const shared_ptr<DNSKEYRecordContent> key, const std::string& msg, vState& ede)
 {
   bool result = false;
   try {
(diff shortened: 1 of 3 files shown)


pdns/validate.cc
```diff
@@ -932,13 +932,18 @@ static bool checkSignatureWithKey(time_t now, const shared_ptr<RRSIGRecordConten
...
       result = dke->verify(msg, sig->d_signature);
       LOG("signature by key with tag "<<sig->d_tag<<" and algorithm "<<DNSSECKeeper::algorithm2name(sig->d_algorithm)<<" was " << (result ? "" : "NOT ")<<"valid"<<endl);
+      if (!result) {
+        ede = vState::BogusNoValidRRSIG;
+      }
     }
     else {
-      LOG("Signature is "<<((sig->d_siginception - g_signatureInceptionSkew > now) ? "not yet valid" : "expired")<<" (inception: "<<sig->d_siginception<<", inception skew: "<<g_signatureInceptionSkew<<", expiration: "<<sig->d_sigexpire<<", now: "<<now<<")"<<endl);
-    }
+      ede = (sig->d_siginception - g_signatureInceptionSkew > now) ? vState::BogusSignatureNotYetValid : vState::BogusSignatureExpired;
```
rgacogne, line 940:
```suggestion
      ede = ((sig->d_siginception - g_signatureInceptionSkew) > now) ? vState::BogusSignatureNotYetValid : vState::BogusSignatureExpired;
```

rgacogne (approved):
Looks good! A test would be nice indeed ;-)

pieterlexis (approved):
Seems to work as intended!

regression-tests.recursor-dnssec/test_ExtendedErrors.py
```diff
@@ -119,6 +119,19 @@ def testExpired(self):
             self.assertEqual(res.options[0].otype, 15)
             self.assertEqual(res.options[0], extendederrors.ExtendedErrorOption(7, b''))
 
+    def testAllExpired(self):
+        qname = 'servfail.nl.'
```
pieterlexis, line 123: we really shouldn't rely on external domains in our tests

omoerbeek: I think some of the existign tests already do or didn't I look close enough?

omoerbeek: Not that I do not agree, but setting up something for these tests not relying on external servers is quite some work.

omoerbeek: #10983

omoerbeek:
I added a test base on the existing `test_Expired`. Since all sigs are expired this hits a different case than the "sig on DS is expired" from this issue, I kept it as it is nice to have anyway.

PR #10981, opened 2021-11-15
auth 2136: apply new TTL to whole RRset, not only to the added record

### LINKED ISSUE #10921: auth: 2136 does not enforce per-RRset TTL
- Program: Authoritative,
 - Issue type: Bug report

### Short description
using nsupdate it's possible to add two A records for the same DNS entry having a different TTL

### Environment

 - Operating system: CentOS 7
 - Software version: 4.4.1
 - it happens also in vesion 4.5.1

### Steps to reproduce

[root@localhost ~]# nsupdate
> server 127.0.0.1
> update add test.example.com 60 IN A 192.168.1.111
> send
> quit

[root@localhost ~]# nsupdate
> server 127.0.0.1
> update add test.example.com 3600 IN A 192.168.1.112
> send
> quit

[root@localhost ~]# pdnsutil list-zone example.com

test.example.com	60	IN	A	192.168.1.111
test.example.com	3600	IN	A	192.168.1.112

### Expected behaviour

nsupdate should return an error and not insert the two same A record with different TTL

### Actual behaviour

nsupdate applied the two update without any warning and doing a check with pdnsutil  it shows the error

[root@localhost ~]# pdnsutil check-all-zones |grep Err
[Error] TTL mismatch in rrset: 'test.example.com IN A 192.168.1.112' (60 != 3600)

Habbie:
### Short description
fixes #10921

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/rfc2136handler.cc
+++ b/pdns/rfc2136handler.cc
@@ -156,7 +156,7 @@ uint PacketHandler::performUpdate(const string &msgPrefix, const DNSRecord *rr, 
           di->backend->replaceRRSet(di->id, oldRec->qname, oldRec->qtype, rrset);
           *updatedSerial = true;
           changedRecords++;
-          g_log<<Logger::Notice<<msgPrefix<<"Replacing record "<<rr->d_name<<"|"<<rrType.toString()<<endl;
+          g_log<<Logger::Notice<<msgPrefix<<"Replacing SOA record "<<rr->d_name<<"|"<<rrType.toString()<<endl;
         } else {
           g_log<<Logger::Notice<<msgPrefix<<"Provided serial ("<<sdUpdate.serial<<") is older than the current serial ("<<sdOld.serial<<"), ignoring SOA update."<<endl;
         }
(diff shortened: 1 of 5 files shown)


Kian0815 (approved):
test the pull an it works

pdns/rfc2136handler.cc
```diff
@@ -204,10 +206,10 @@ uint PacketHandler::performUpdate(const string &msgPrefix, const DNSRecord *rr,
         }
         if (updateTTL > 0) {
           di->backend->replaceRRSet(di->id, rr->d_name, rrType, rrset);
-          g_log<<Logger::Notice<<msgPrefix<<"Replacing record "<<rr->d_name<<"|"<<rrType.toString()<<endl;
+          g_log<<Logger::Notice<<msgPrefix<<"Updating TTLs for "<<rr->d_name<<"|"<<rrType.toString()<<endl;
```
rgacogne, line 209: I'm a bit confused by that change, since it looks like we might be updating more than the TTLs in that case, or am I missing something?

Habbie: It is, painfully, a full DELETE+INSERT, but the net change really should be just the TTLs. If that's not true, I missed something.

rgacogne: You are not, I was! It took me several tries to understand exactly what `rr` and `rrset` are at this point, sorry.

rgacogne (approved):

Habbie:
I split this into two commits, same content.

PR #10984, opened 2021-11-15
docs: Add hint on how to find generic records in SQL backends

danel1:
### Short description
- Add a hint including an SQL-Query on how to identify records with an unknown record type (as discussed on IRC).

### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -1784,8 +1784,9 @@ Disable after record contents have been upgraded.
 This option is supported by the bind and Generic SQL backends. 
 
 .. note::
-  When using a generic SQL backend, records with an unknown record type (see :doc:`../appendices/types`) can be identified with the following SQL-Query:
-  SELECT * from records where type like 'TYPE%';
+  When using a generic SQL backend, records with an unknown record type (see :doc:`../appendices/types`) can be identified with the following SQL-Query::
+  
+      SELECT * from records where type like 'TYPE%';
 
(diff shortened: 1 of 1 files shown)


docs/settings.rst
```diff
@@ -1783,6 +1783,10 @@ Disable after record contents have been upgraded.
 
 This option is supported by the bind and Generic SQL backends. 
 
+.. note::
+  When using a generic SQL backend, records with an unknown record type (see :doc:`../appendices/types`) can be identified with the following SQL-Query:
```
pieterlexis, line 1787: I suggest this for correct rendering:

```suggestion
  When using a generic SQL backend, records with an unknown record type (see :doc:`../appendices/types`) can be identified with the following SQL-Query::
  
```

docs/settings.rst
```diff
@@ -1783,6 +1783,10 @@ Disable after record contents have been upgraded.
...
+  SELECT * from records where type like 'TYPE%';
```
pieterlexis, line 1788: To render this nicely, try this:

```suggestion
.. note::
  When using a generic SQL backend, records with an unknown record type (see :doc:`../appendices/types`) can be identified with the following SQL-Query::
  
      SELECT * from records where type like 'TYPE%';
```

pieterlexis (approved):

PR #10986, opened 2021-11-16
Check formatting on GH

omoerbeek:
### Short description

Should be marked as "required"

### Checklist

I have:
- [ ] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/.github/workflows/formatting.yml
+++ b/.github/workflows/formatting.yml
@@ -17,7 +17,7 @@ jobs:
           submodules: recursive
       - run: |
           sudo apt-get update
-          sudo apt-get -qq --no-install-recommends install git clang-12
+          sudo apt-get -qq --no-install-recommends install git clang-11
       - run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/formatting.yml
```diff
@@ -0,0 +1,33 @@
...
+    name: verify formatting and Makefile.am sort order
+    # on a ubuntu-20.04 VM
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v2.3.4
+        with:
+          fetch-depth: 5
+          submodules: recursive
+      - run: |
+          sudo apt-get update
```
Habbie, line 19: In the build log, this step now calls itself `sudo apt-get update`, which looks like the install bit is missing. Putting them on one line with `&&` works, making them two steps also works, and giving the step a `name:` is cleanest.

Habbie: (The next two steps also look confusing in the build log for the same reason.)

Habbie (changes requested):

Habbie (approved):

PR #10990, opened 2021-11-16
rec: Basic notify test

omoerbeek:
<strike>It turns out a notify increments cache-hits, that feel a bit strange and should be looked into.</strike>

Pinging @kpfleming 

### Checklist

I have:
- [ ] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/pdns_recursor.cc
+++ b/pdns/pdns_recursor.cc
@@ -2412,11 +2412,13 @@ static void startDoResolve(void *p)
       g_log<<endl;
     }
 
-    if (sr.d_outqueries || sr.d_authzonequeries) {
-      g_recCache->cacheMisses++;
-    }
-    else {
-      g_recCache->cacheHits++;
+    if (dc->d_mdp.d_header.opcode != Opcode::Notify) {
+      if (sr.d_outqueries || sr.d_authzonequeries) {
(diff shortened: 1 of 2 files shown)


pdns/pdns_recursor.cc
```diff
@@ -2412,11 +2412,13 @@ static void startDoResolve(void *p)
       g_log<<endl;
     }
 
-    if (sr.d_outqueries || sr.d_authzonequeries) {
-      g_recCache->cacheMisses++;
-    }
-    else {
-      g_recCache->cacheHits++;
+    if (dc->d_mdp.d_header.opcode != Opcode::Notify) {
```
pieterlexis, line 2415: I _think_ this makes more sense, we reject other OPCODEs earlier.

```suggestion
    if (dc->d_mdp.d_header.opcode == Opcode::Query) {
```

omoerbeek: It's equivalent, but it might read better indeed.

kpfleming:
Including them in the count wasn't intentional, just the easiest path to getting NOTIFY support implemented, so I have zero concerns about changing that behaviour.

pieterlexis (approved):

PR #11008, opened 2021-11-18
dnsdist: Add a function to set the UDP recv/snd buffer sizes

### LINKED ISSUE #10898: dnsdist: Ponder automatically raising SO_RCVBUF/SO_SNDBUF
- Program:dnsdist 
 - Issue type: Feature request

### Short description
On Linux, at least, dnsdist is well-known to perform much better on the UDP paths when `SO_RCVBUF` and `SO_SNDBUF` are raised on the incoming and outgoing sockets, because it provides much more room to deal with spikes.
It would be nice to try to raise these settings to higher values at startup, unless configured not to.
We usually recommend these settings:
```
net.core.rmem_max=33554432
net.core.wmem_max=33554432
net.core.rmem_default=16777216
net.core.wmem_default=16777216
```
so raising to 16777216 by default when possible would make sense.

Note that:
- `calidns` tries to raise both to `2000000`
-  the `recursor` tries to raise `SO_RECVBUF` to `250000` for incoming UDP sockets, and `SO_SNDBUF` to `65000` for incoming TCP ones.

Neither of them lowers these values if the system defaults are higher than that, and we shouldn't either.

rgacogne:
### Short description

And raise them to 16777216 by default. Although perhaps we should only do that on Linux?

This PR also refactors the functions duplicated in the recursor, calidns and dnsreplay and therefore impacts the auth and the rec.

Closes #10898.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdist.cc
+++ b/pdns/dnsdist.cc
@@ -141,6 +141,8 @@ bool g_servFailOnNoPolicy{false};
 bool g_truncateTC{false};
 bool g_fixupCase{false};
 bool g_dropEmptyQueries{false};
+uint32_t g_socketUDPSendBuffer{16777216};
+uint32_t g_socketUDPRecvBuffer{16777216};
 
 std::set<std::string> g_capabilitiesToRetain;
 
@@ -2070,6 +2072,26 @@ static void setUpLocalBind(std::unique_ptr<ClientState>& cs)
     }
     catch(const std::exception& e) {
       warnlog("Failed to set IP_MTU_DISCOVER on UDP server socket for local address '%s': %s", cs->local.toStringWithPort(), e.what());
+    }
+  }
+
+  if (!cs->tcp) {
+    if (g_socketUDPSendBuffer > 0) {
+      try {
+        setSocketSendBuffer(cs->udpFD, g_socketUDPSendBuffer);
+      }
+      catch (const std::exception& e) {
+        warnlog(e.what());
+      }
+    }
+
+    if (g_socketUDPRecvBuffer > 0) {
+      try {
+        setSocketReceiveBuffer(cs->udpFD, g_socketUDPRecvBuffer);
+      }
+      catch (const std::exception& e) {
+        warnlog(e.what());
+      }
     }
   }
 
--- a/pdns/dnsdist.cc
+++ b/pdns/dnsdist.cc
@@ -141,8 +141,8 @@ bool g_servFailOnNoPolicy{false};
 bool g_truncateTC{false};
 bool g_fixupCase{false};
 bool g_dropEmptyQueries{false};
-uint32_t g_socketUDPSendBuffer{16777216};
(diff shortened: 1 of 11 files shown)


pdns/pdns_recursor.cc
```diff
@@ -3744,7 +3722,12 @@ static void makeUDPServerSockets(deferredAdd_t& deferredAdds)
 
     setCloseOnExec(fd);
 
-    setSocketReceiveBuffer(fd, 250000);
+    try {
+      setSocketReceiveBuffer(fd, 25000);
```
phonedph1, line 3726: is this value supposed to change from before?

rgacogne: Not at all, thanks a lot!

pdns/dnsdist.cc
```diff
@@ -141,6 +141,8 @@ bool g_servFailOnNoPolicy{false};
 bool g_truncateTC{false};
 bool g_fixupCase{false};
 bool g_dropEmptyQueries{false};
+uint32_t g_socketUDPSendBuffer{16777216};
```
omoerbeek, line 144: This is quite greedy, on OpenBSD:
```
Unable to raise socket buffer size to 16777216: No buffer space available
Unable to raise socket buffer size to 16777216: No buffer space available
```

omoerbeek (commented):
A nit for the OpenBSD case.

omoerbeek:
The failing rtecursor regression tests seem to fail to start the auth... ??

pdns/dnsdist.cc
rgacogne: Right, I was afraid of something like that. I now see that the default on Linux is likely smaller than that on most distributions (212992 on x86_64, 163840 on 32 bits). So I see three options:
- a smaller default value, like the recursor does, perhaps 160k?
- default value of 0 so we keep the system default
- default value of 0 except for Linux, where we set something like 200k?

omoerbeek: 250000 works on OpenBSD. so a default of that (or smaller) for dnsdist would work for me

rgacogne: On all Linux distributions I checked, `rmem_default` == `rmem_max` and `wmem_default` == `wmem_max` by default, so we would not be able to raise the value. So I guess doing nothing by default makes more sense, and the administrator can set higher values if needed.

omoerbeek (approved):
Now wondering why rec only sets the receive buffer size and not the send buffer size. A peak in incoming request normally also leaks to a peak in replies I would say. But that's a different issue.

rgacogne:
> Now wondering why rec only sets the receive buffer size and not the send buffer size. A peak in incoming request normally also leaks to a peak in replies I would say.

Agreed. The good news is that raising `net.core.wmem_default` is enough to fix that, but it would be nice to do it by default.

PR #11017, opened 2021-11-22
dnsdist: Add Lua bindings to get the list of network interfaces, addresses

rgacogne:
### Short description
This PR adds Lua bindings to retrieve the list of network interfaces and associated addresses, provided that `getifaddrs` is supported.
I tested this code on Linux, the documentation suggests that it should work on FreeBSD and OpenBSD as well but I did not test that.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/iputils.cc
+++ b/pdns/iputils.cc
@@ -25,6 +25,10 @@
 #include "iputils.hh"
 #include <sys/socket.h> 
 
+#if HAVE_GETIFADDRS
+#include <ifaddrs.h>
+#endif
+
 /** these functions provide a very lightweight wrapper to the Berkeley sockets API. Errors -> exceptions! */
 
 static void RuntimeError(const boost::format& fmt)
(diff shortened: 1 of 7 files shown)


pdns/iputils.cc
```diff
@@ -543,3 +547,56 @@ void setSocketSendBuffer(int fd, uint32_t size)
...
+  std::vector<ComboAddress> result;
+#if HAVE_GETIFADDRS
+  struct ifaddrs *ifaddr;
+  if (getifaddrs(&ifaddr) == -1) {
+    return result;
+  }
+
+  for (struct ifaddrs *ifa = ifaddr; ifa != nullptr; ifa = ifa->ifa_next) {
+    if (ifa->ifa_name == nullptr || strcmp(ifa->ifa_name, itf.c_str()) != 0) {
+      continue;
```
Habbie, line 583: Couldn't find a better spot for this comment. If the user passes in a name that does not exist, the output is an empty table. Should it be an error? (This is really a question, I'm not suggesting either way is better):

```
> getListOfAddressesOfNetworkInterface('foo')
{}
```

rgacogne: Good question! I kind of like the current behaviour, but that's likely because I only care about interfaces that do have addresses assigned to them and I see how someone could have a different opinion. On the other hand raising an exception might be a bit too much, especially since on some setups interfaces come and go.

Habbie: That makes sense, let's leave it.

rgacogne:
Rebased to fix conflicts.

pdns/iputils.cc
```diff
@@ -543,3 +547,56 @@ void setSocketSendBuffer(int fd, uint32_t size)
...
+    }
+    ComboAddress addr;
+    try {
+      addr.setSockaddr(ifa->ifa_addr, ifa->ifa_addr->sa_family == AF_INET ? sizeof(struct sockaddr_in) : sizeof(struct sockaddr_in6));
+    }
+    catch (...) {
+      continue;
+    }
+
+    result.push_back(addr);
```
chbruyand, line 596: Would it make sense to push the new addr only if there was no exception ?

chbruyand (approved):
LGTM!

pdns/iputils.cc
rgacogne: Don't we `continue` to the next address when an exception is raised? Or do you mean that we should not continue but stop right away, perhaps?

chbruyand: Right, didn't see the `continue` (not really used to it)

PR #11018, opened 2021-11-22
rec: Fix logic botch introduced by notify handling

omoerbeek:
See #10751, some extra scrutiny review is needed to make sure no other similar issue remains.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/pdns_recursor.cc
+++ b/pdns/pdns_recursor.cc
@@ -2952,6 +2952,7 @@ static void handleRunningTCPQuestion(int fd, FDMultiplexer::funcparam_t& var)
             }
 
             g_stats.sourceDisallowedNotify++;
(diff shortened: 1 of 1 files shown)


rgacogne (approved):
That change looks good to me. Please also have a look at whether we should call `terminateTCPConnection` or perhaps `sendErrorOverTCP` in these two cases:
- `g_stats.zoneDisallowedNotify++`
- `g_stats.sourceDisallowedNotify++`

rgacogne (approved):

omoerbeek:
I think these cases are OK now, but we really should check `startDoResolve` for not doing TCP connection accounting, as it has `return` statements. Will open an issue for that.

Issue #11009, opened 2021-11-18
dnsdist: allow users to inject their own XDP program to handle dropped/redirected packets


MiniPierre:
- Program: dnsdist 
 - Issue type: Feature request

### Short description
Allow the XDP filter to perform logging action on dropped and redirected packets

### Usecase
After the XDP filter pull request https://github.com/PowerDNS/pdns/pull/10498, it appeared that we wanted to have the opportunity to log packets before performing DROP or TC action. As other people may have the same need but do not log the same way, logging facilities should be as modular as possible

### Description
I was thinking of using a BPF_MAP_TYPE_PROG_ARRAY inside the filter that allow user to inject their own XDP program at specific indexes (let say index 0 is for logging dropped packets, index 1 is for redirected packets). The XDP filter would check if any program is loaded at the given indexes and perform the tail call if something is found, or perform the default action if no program is found

Might be something like:
```
BPF_TABLE("prog", int, int, progsarray, 2);
...
//if packet is dropped
if progsarray.lookup(0) { 
  progsarray.call(ctx, 0);
}

return XDP_DROP
...
//if packet is redirected
if progsarray.lookup(1) {
  progsarray.call(ctx, 1);
}

return XDP_TX
```
I would be totally OK to work on such thing but I would like to know if it is something that would have any interest for dnsdist, or if it is something too specific to be implemented

rgacogne:
The feature seems very interesting! Are you considering a pinned program table, that could be updated from an external source without reloading the main XDP program, or a "local" table that will require the logging functions to be loaded by the main XDP program? Both options are fine by me, I'm just trying to better understand what you have in mind.

MiniPierre:
I was considering a pinned table, so that we can update the loaded programs at will from an external source. Adding logging function to dnsdist for such situation might make it quite heavy for a need that is quite specific to our organization.

rgacogne:
That makes complete sense, thanks! So yes, I would gladly merge such a feature.

MiniPierre:
For information, it seems that map lookup are not authorized in progs array. It is not a problem as it appears that if no program is loaded at the given index, the BPF program does continue its execution, performing the remaining actions as a "default" behavior

```If no eBPF program is found at the given index of the program array (because the map slot doesn't contain a valid program file descriptor, the specified lookup index/key is out of bounds, or the limit of 32 nested calls has been exceed), execution continues with the current eBPF program. This can be used as a fall-through for default cases. ```
[...]

PR #11020, opened 2021-11-22
dnsdist: Added XDP middleware for dropped/redirected queries logging

MiniPierre:
### Short description

According to this issue https://github.com/PowerDNS/pdns/issues/11009, this pull request add the possibility to load user-defined XDP function when performing TC or DROP action. This allow the user to log the queries even if they do not reach libpcap.
I added a custom XDP filter and a Python script to load it and output information about dropped/tc packets
 
### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/contrib/xdp-filter.ebpf.src
+++ b/contrib/xdp-filter.ebpf.src
@@ -117,6 +117,7 @@ struct map_value
 BPF_TABLE_PINNED("hash", uint32_t, struct map_value, v4filter, 1024, "/sys/fs/bpf/dnsdist/addr-v4");
 BPF_TABLE_PINNED("hash", struct in6_addr, struct map_value, v6filter, 1024, "/sys/fs/bpf/dnsdist/addr-v6");
 BPF_TABLE_PINNED("hash", struct dns_qname, struct map_value, qnamefilter, 1024, "/sys/fs/bpf/dnsdist/qnames");
+BPF_TABLE_PINNED("prog", int, int, progsarray, 2, "/sys/fs/bpf/dnsdist/progs");
 
 /*
  * bcc has added BPF_TABLE_PINNED7 to the latest commit of the master branch, but it has not yet been released.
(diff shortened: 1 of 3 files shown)


MiniPierre:
I saw that @rgacogne added a type to dns_action, actually it made the test script to fail so I removed it, no problem with having it back

rgacogne:
> I saw that @rgacogne added a type to dns_action, actually it made the test script fail so I removed it, no problem with having it back

I briefly tested adding it back and the program loaded fine, but perhaps it fails after the startup?

MiniPierre:
Just adapted the indentation, I was in a hurry and did not took the time to fix this before.

> > I saw that @rgacogne added a type to dns_action, actually it made the test script fail so I removed it, no problem with having it back
> 
> I briefly tested adding it back and the program loaded fine, but perhaps it fails after the startup?

It was failing at the python script startup, maybe it is a problem on our side

rgacogne (approved):

rgacogne:
Thanks for fixing the indentation!

> It was failing at the python script startup, maybe it is a problem on our side

Actually that's my fault, you can't specify the underlying type of an enum in C, I guess I'm too used to C++11 these days, and my compiler somehow allows that..
The problem with the current code is that the size of the `map_value` depends on the compiler, which might be a problem. But it's not related to your changes, so don't worry about it, I'll fix in a different PR.

MiniPierre:
Could you tell me how you converted your .ebpf.src files to .ebpf ? I am trying to switch from Python to C++ to load the  logging middleware as Python is not fast enough to pull the event perf buffer under high load, so events get dropped.

rgacogne:
The easiest way for you, since you are already using `bcc`, is to dump the eBPF code of a given function to a file like this:
```python
xdp = BPF(src_file="xdp-filter.ebpf.src")
newFile = open ("/tmp/ebpf.bin", "wb")
dump = xdp.dump_func("xdp_dns_filter")
newFile.write(dump)
newFile.close()
```

In dnsdist I did not want to commit a huge binary blob, especially since we have to update the file descriptors in that blob to match the one linked to eBPF maps, so I used this script (which doesn't cover all functions but would be easy to update) to get the eBPF code in a more easy to understand form: https://coredump.fr/static/powerdns/reverse.py

rgacogne:
If you try to use that script, `113` is the code for `BPF_func_skb_output` :)

rgacogne (approved):
Rebased on master to fix conflicts, looks good to me!

PR #11038, opened 2021-11-26
rec: Disable tsan regression runs for rec for now,

omoerbeek:
There is a failure mode that if it hits makes almost all remaining test fail.  Symptom
is that the auths do not start up properly.

### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/.github/workflows/build-and-test-all.yml
+++ b/.github/workflows/build-and-test-all.yml
@@ -303,7 +303,7 @@ jobs:
     runs-on: ubuntu-20.04
     strategy:
       matrix:
-        sanitizers: [ubsan+asan, tsan]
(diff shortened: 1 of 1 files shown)


rgacogne (approved):
> There is a failure mode that if it hits makes almost all remaining test fail. Symptom is that the auths do not start up properly.

Fine by me, there is nothing worse than tests that are regularly failing. Any clue to what is happening?

I'm pondering doing the same for the dnsdist TSAN tests, but these caught real issues in the past so I'm a bit reluctant..

omoerbeek:
I have no clue (otherwise I would spend time on it) but I intend to come back to this one day.

PR #11053, opened 2021-11-30
Numeric sort for numeric DNS name fields

p12n:
### Short description
Sort numeric DNS fields numerically. This is particularly noticeable in the `pdnsutil edit-zone` use case for IPv4 reverse zones, where many DNS name components are small integers. For example,

`99.1.168.192.in-addr.arpa`
`101.1.168.192.in-addr.arpa`

now sorts correctly. The previous lexical sort is retained for mixed alphanumeric fields, as it is not clear what users would expect in all cases. So, for example, `ns12` still sorts above `ns8`.

### Checklist
I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsname.hh
+++ b/pdns/dnsname.hh
@@ -214,6 +214,17 @@ inline bool DNSName::canonCompare(const DNSName& rhs) const
     uint8_t ourlen = *(ourstr - 1), rhslen = *(rhsstr - 1);
     bool res;
 
+    // If both names are integers of different lengths: longer name is greater
+    if (ourlen != rhslen) {
+      res = true;
+      for (int i = 0; res && i < ourlen; i++)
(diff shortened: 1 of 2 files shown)


Habbie:
Hello! This is cool, and people have been asking for this, but you cannot change `canonCompare`, it has to be exactly as it is, because otherwise DNSSEC breaks. If i put `10`, `15` and `100.example.com` in a zone, I get:

```
10.example.com.         60      IN      NSEC    15.example.com. A RRSIG NSEC
```

but it should point to 100.

So, can you make a separate comparator to use just for pdnsutil?

(I'll contribute a testcase that breaks your current version)

Habbie:
Hi! Did you want to finish this work?

miodvallat:
Superseded by #15999.

PR #11061, opened 2021-12-02
auth-4.6.0-beta1: secpoll&docs

Habbie:
~Unfinished, alpha2 might grow a few more PRs.~ Now finished, and renamed to beta.1

PR #11064, opened 2021-12-03
secpoll: for pre-releases, use status 2 when supserseded.

omoerbeek:
### Short description

A proposal:

Slight change in the secpoll policy: status 2 is also issued for non-security related upgrade recommendations, but for pre-releases only. This is to nudge people into not running a pre-release version if a newer one is available.

Also add the policy to `secpoll.zone` itself.

If this is deemed a good thing, I'll update the secpoll records of pre-releases as well.

### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code


Habbie (approved):
Looks good to me! Yesterday's #11061 was already compliant :D

rgacogne (approved):

PR #11066, opened 2021-12-03
Auth: add zone to zonecache on flush API endpoint

zeha:
### Short description
Some people want to add domains to the zonecache as soon as possible. This is currently automatic when the zone is created using the API or autoprimary. For database-based provisioning systems this allows the provisioning system to add the zone by calling the /cache/flush endpoint.

This is #10617 without any cleanup, and adding the zone works only on the HTTP APIs cache flush endpoint. Nothing happens on the command line, etc.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/ws-auth.cc
+++ b/pdns/ws-auth.cc
@@ -2270,6 +2270,16 @@ static void apiServerCacheFlush(HttpRequest* req, HttpResponse* resp) {
 
   DNSName canon = apiNameToDNSName(req->getvars["domain"]);
 
+  if (g_zoneCache.isEnabled()) {
+    DomainInfo di;
+    UeberBackend B;
+    if (B.getDomainInfo(canon, di, false)) {
+      // zone exists (uncached), add/update it in the zone cache.
+      // Handle this first, to avoid concurrent queries re-populating the other caches.
+      g_zoneCache.add(di.zone, di.id);
+    }
+  }
+
   // purge entire zone from cache, not just zone-level records.
   uint64_t count = purgeAuthCaches(canon.toString() + "$");
   resp->setJsonBody(Json::object {
--- a/pdns/ws-auth.cc
+++ b/pdns/ws-auth.cc
@@ -2270,7 +2270,8 @@ static void apiServerCacheFlush(HttpRequest* req, HttpResponse* resp) {
 
   DNSName canon = apiNameToDNSName(req->getvars["domain"]);
 
-  uint64_t count = purgeAuthCachesExact(canon);
+  // purge entire zone from cache, not just zone-level records.
+  uint64_t count = purgeAuthCaches(canon.toString() + "$");
   resp->setJsonBody(Json::object {
       { "count", (int) count },
       { "result", "Flushed cache." }


pdns/ws-auth.cc
```diff
@@ -2270,6 +2270,14 @@ static void apiServerCacheFlush(HttpRequest* req, HttpResponse* resp) {
 
   DNSName canon = apiNameToDNSName(req->getvars["domain"]);
 
+  SOAData sd;
+  UeberBackend B;
+  if (g_zoneCache.isEnabled() && B.getSOAUncached(canon, sd)) {
+    // zone exists (uncached), add/update it in the zone cache.
+    // Handle this first, to avoid concurrent queries re-populating the other caches.
+    g_zoneCache.add(canon, sd.domain_id);
+  }
```
mind04, line 2279:
```suggestion
  int zoneId{0};
  DomainInfo di;
  if (g_zoneCache.isEnabled() && !g_zoneCache.getEntry(canon, zoneId)) {
    UeberBackend B;
    if (B.getDomainInfo(canon, di, false)) {
      // zone exists (uncached), add/update it in the zone cache.
      // Handle this first, to avoid concurrent queries re-populating the other caches.
      g_zoneCache.add(canon, di.id);
    }
  }
```
Prevent some DB actions for zones already present in the zonecache and add zones without SOA records as well.

pdns/ws-auth.cc
```diff
@@ -2270,6 +2270,14 @@ static void apiServerCacheFlush(HttpRequest* req, HttpResponse* resp) {
...
+
   uint64_t count = purgeAuthCachesExact(canon);
```
mind04, line 2281:
```suggestion
  uint64_t count = purgeAuthCaches(canon.toString() + "$");
```
Bonus fix unrelated to the pull ;)

mind04 (changes requested):

pdns/ws-auth.cc
zeha: I like it, but I'll point out that changing the zoneId will not work with this code as is.

pdns/ws-auth.cc
zeha: added!

pdns/ws-auth.cc
zeha: added.

mind04: I think it is better to remove "&& !g_zoneCache.getEntry(canon, zoneId)". 
We nuke the entire query cache now, so one additional query will probably not hurt.

zeha: Agreed, done.

mind04 (approved):

zeha:
dnsdist tsan failure looks unrelated, i think?

Habbie:
> dnsdist tsan failure looks unrelated, i think?

Ongoing, spurious but often, not debugged yet.

discostur:
Is this just for the api or also for `pdns_control`? If i use the `pdns_control flush` command it will not add the zone to the cache, or?

zeha:
> Is this just for the api or also for `pdns_control`? If i use the `pdns_control flush` command it will not add the zone to the cache, or?

No, thats in #10617 as stated above.

PR #11070, opened 2021-12-03
rec: Regression tests: startup/cleanup fixes

omoerbeek:
### Short description

Trying to diagnose the recursor asan + ubsan failures, I spotted this.

### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/regression-tests.recursor-dnssec/recursortests.py
+++ b/regression-tests.recursor-dnssec/recursortests.py
@@ -556,14 +556,11 @@ distributor-threads={threads}""".format(confdir=confdir,
         cls.waitForTCPSocket(ipaddress, 53)
 
         if cls._auths[ipaddress].poll() is not None:
-            try:
-                cls._auths[ipaddress].kill()
-            except OSError as e:
-                if e.errno != errno.ESRCH:
-                    raise
-                with open(logFile, 'r') as fdLog:
-                    print(fdLog.read())
-            sys.exit(cls._auths[ipaddress].returncode)
+            print(f"\n*** startAuth log for {logFile} ***")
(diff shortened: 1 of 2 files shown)


omoerbeek:
While hunting regression test failures I noticed that the startup and teardown code is trying to kill processes
even when `proces.poll()` returns `not None`. That makes little sense, as that's an indication the process is not running:

```
Popen.poll()
    Check if child process has terminated. Set and return returncode attribute. Otherwise, returns None.
```
While there, fix a race on startup of the API test and print the log capture with clear markers.

regression-tests.recursor-dnssec/recursortests.py
```diff
@@ -674,6 +668,9 @@ def tearDownResponders(cls):
 
     @classmethod
     def killProcess(cls, p):
+        # Don't try to kill it if it's already dead
+        if p.poll() is not None:
```
rgacogne, line 672: I wonder it there are valid cases where the recursor is already gone at this point? (Note that this PR does not change anything in that regard since I believe we are already ignoring that case in the existing code).

rgacogne (approved):
Looks good!

regression-tests.recursor-dnssec/recursortests.py
omoerbeek: Yes, this is something to keep an eye on. If the startup succeeded, all processes we're trying to kill should be alive at the end. Any process not running is an indication of a crash or other issue.

But what I do not know is what happens if startup of one of the processes fails: does the rest of the startup continue? Will the teardown code be called?

PR #11074, opened 2021-12-07
rec: postresolve ffi

omoerbeek:
### Short description

Initial stab at providing an ffi interface for postresolve.

Although I think this is already useful, this PR is mainly to solicit feedback.
The API likely need some extra tweaks and methods.

Many simple cases can be handled by `pdns_postresolve_ffi_handle_set_appliedpolicy_kind()` (custom is not available for that case, since it is really RPZ oriented)
It is also possible to modify existing records (content only), remove all records and add records.
Current methods:
```
  typedef struct pdns_postresolve_ffi_handle pdns_postresolve_ffi_handle_t;

  const char* pdns_postresolve_ffi_handle_get_qname(pdns_postresolve_ffi_handle_t* ref) __attribute__((visibility("default")));
  void pdns_postresolve_ffi_handle_get_qname_raw(pdns_postresolve_ffi_handle_t* ref, const char** qname, size_t* qnameSize) __attribute__((visibility("default")));
  uint16_t pdns_postresolve_ffi_handle_get_qtype(const pdns_postresolve_ffi_handle_t* ref) __attribute__((visibility("default")));
  uint16_t pdns_postresolve_ffi_handle_get_rcode(const pdns_postresolve_ffi_handle_t* ref) __attribute__((visibility("default")));
  void pdns_postresolve_ffi_handle_set_rcode(const pdns_postresolve_ffi_handle_t* ref, uint16_t rcode) __attribute__((visibility("default")));
  pdns_policy_kind_t pdns_postresolve_ffi_handle_get_appliedpolicy_kind(const pdns_postresolve_ffi_handle_t* ref) __attribute__((visibility("default")));
  void pdns_postresolve_ffi_handle_set_appliedpolicy_kind(pdns_postresolve_ffi_handle_t* ref, pdns_policy_kind_t kind) __attribute__((visibility("default")));
  bool pdns_postresolve_ffi_handle_get_record(pdns_postresolve_ffi_handle_t* ref, unsigned int i, pdns_ffi_record_t* record, bool raw) __attribute__((visibility("default")));
  bool pdns_postresolve_ffi_handle_set_record(pdns_postresolve_ffi_handle_t* ref, unsigned int i, const char* content, size_t contentLen, bool raw) __attribute__((visibility("default")));
  void pdns_postresolve_ffi_handle_clear_records(pdns_postresolve_ffi_handle_t* ref) __attribute__((visibility("default")));
  bool pdns_postresolve_ffi_handle_add_record(pdns_postresolve_ffi_handle_t* ref, const char* name, uint16_t type, uint32_t ttl, const char* content, size_t contentLen, pdns_record_place_t place, bool raw) __attribute__((visibility("default")));
  const char* pdns_postresolve_ffi_handle_get_authip(pdns_postresolve_ffi_handle_t* ref) __attribute__((visibility("default")));
  void pdns_postresolve_ffi_handle_get_authip_raw(pdns_postresolve_ffi_handle_t* ref, const void** addr, size_t* addrSize) __attribute__((visibility("default")));
```

Refer to the regression test for some example use.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [X] documented the code

### DIFF
--- a/pdns/lua-recursor4-ffi.hh
+++ b/pdns/lua-recursor4-ffi.hh
@@ -59,6 +59,7 @@ extern "C"
   typedef struct pdns_ffi_record
   {
     const char* name;
+    size_t name_len;
     const char* content;
     size_t content_len;
     uint32_t ttl;
@@ -114,8 +115,10 @@ extern "C"
   typedef struct pdns_postresolve_ffi_handle pdns_postresolve_ffi_handle_t;
 
   const char* pdns_postresolve_ffi_handle_get_qname(pdns_postresolve_ffi_handle_t* ref) __attribute__((visibility("default")));
+  void pdns_postresolve_ffi_handle_get_qname_raw(pdns_postresolve_ffi_handle_t* ref, const char** qname, size_t* qnameSize) __attribute__((visibility("default")));
   uint16_t pdns_postresolve_ffi_handle_get_qtype(const pdns_postresolve_ffi_handle_t* ref) __attribute__((visibility("default")));
   uint16_t pdns_postresolve_ffi_handle_get_rcode(const pdns_postresolve_ffi_handle_t* ref) __attribute__((visibility("default")));
+  void pdns_postresolve_ffi_handle_set_rcode(const pdns_postresolve_ffi_handle_t* ref, uint16_t rcode) __attribute__((visibility("default")));
   pdns_policy_kind_t pdns_postresolve_ffi_handle_get_appliedpolicy_kind(const pdns_postresolve_ffi_handle_t* ref) __attribute__((visibility("default")));
   void pdns_postresolve_ffi_handle_set_appliedpolicy_kind(pdns_postresolve_ffi_handle_t* ref, pdns_policy_kind_t kind) __attribute__((visibility("default")));
   bool pdns_postresolve_ffi_handle_get_record(pdns_postresolve_ffi_handle_t* ref, unsigned int i, pdns_ffi_record_t* record, bool raw) __attribute__((visibility("default")));
@@ -125,3 +128,5 @@ extern "C"
   const char* pdns_postresolve_ffi_handle_get_authip(pdns_postresolve_ffi_handle_t* ref) __attribute__((visibility("default")));
   void pdns_postresolve_ffi_handle_get_authip_raw(pdns_postresolve_ffi_handle_t* ref, const void** addr, size_t* addrSize) __attribute__((visibility("default")));
 }
+
+#undef PDNS_VISIBILITY
--- a/pdns/lua-recursor4-ffi.hh
+++ b/pdns/lua-recursor4-ffi.hh
@@ -128,5 +128,3 @@ extern "C"
   const char* pdns_postresolve_ffi_handle_get_authip(pdns_postresolve_ffi_handle_t* ref) __attribute__((visibility("default")));
   void pdns_postresolve_ffi_handle_get_authip_raw(pdns_postresolve_ffi_handle_t* ref, const void** addr, size_t* addrSize) __attribute__((visibility("default")));
 }
-
-#undef PDNS_VISIBILITY
--- a/pdns/lua-recursor4-ffi.hh
+++ b/pdns/lua-recursor4-ffi.hh
@@ -56,7 +56,8 @@ extern "C"
     pdns_policy_kind_custom = 5
   } pdns_policy_kind_t;
 
-  typedef struct pdns_ffi_record {
+  typedef struct pdns_ffi_record
+  {
     const char* name;
     const char* content;
     size_t content_len;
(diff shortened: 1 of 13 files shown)


pdns/lua-recursor4.hh
```diff
@@ -58,6 +58,21 @@ struct LuaContext::Pusher<pdns_ffi_param*>
   }
 };
 
+// pdns_postresolve_ffi_param_t is a lightuserdata
```
rgacogne, line 61:
```suggestion
// pdns_postresolve_ffi_handle is a lightuserdata
```

rgacogne (commented):
Looks good at a quick glance. I'm a bit surprised by the "from auth IP" change, what is the use-case?

omoerbeek:
> Looks good at a quick glance. I'm a bit surprised by the "from auth IP" change, what is the use-case

When I discussed this with @neilcook and @wojas it came up that reaction on the IP (and maybe later name) of the NS tha provided the answer would be useful. Compare it to the "NSIP" trigger (.rpz-nsip) from RPZs. This one does not take into account intermediate nameservers used in resolving, but only the name server that provides the final answer. We do have this info in the case, so it was relatively easy to add.

pdns/lua-recursor4.cc
```diff
@@ -1033,3 +1034,150 @@ void pdns_ffi_param_add_meta_single_int64_kv(pdns_ffi_param_t* ref, const char*
...
+
+    auto ret = d_postresolve_ffi(&handle);
+    return ret;
+  }
+  return false;
+}
+
+const char* pdns_postresolve_ffi_handle_get_qname(pdns_postresolve_ffi_handle_t* ref)
+{
+  auto str = ref->insert(ref->handle.d_dq.qname.toStringNoDot());
```
wojas, line 1069: The `toStringNoDot()` call used to be expensive, not sure if it was optimized since. The gettag_ffi API also has a raw API that returns the string in DNS wire format.

But we can keep it as is, as I don't think we will actually use this function?

Would a raw version even make sense in a response where I suppose label compression could be used?

pdns/lua-recursor4.cc
```diff
@@ -1033,3 +1034,150 @@ void pdns_ffi_param_add_meta_single_int64_kv(pdns_ffi_param_t* ref, const char*
...
+}
+
+bool pdns_postresolve_ffi_handle_get_record(pdns_postresolve_ffi_handle_t* ref, unsigned int i, pdns_ffi_record_t* record, bool raw)
+{
+  if (i >= ref->handle.d_dq.currentRecords->size()) {
+    return false;
+  }
+  try {
+    DNSRecord& r = ref->handle.d_dq.currentRecords->at(i);
+    record->name = ref->insert(r.d_name.toStringNoDot())->c_str();
```
wojas, line 1100: See other comment. This is a field that we would never need when checking response IPs.

pdns/lua-recursor4.cc
```diff
@@ -1033,3 +1034,150 @@ void pdns_ffi_param_add_meta_single_int64_kv(pdns_ffi_param_t* ref, const char*
...
+    if (raw) {
+      auto content = ref->insert(r.d_content->serialize(r.d_name, true));
```
wojas, line 1102: Out of curiosity: why is `d_name` needed here?

pdns/lua-recursor4-ffi.hh
```diff
@@ -89,4 +110,18 @@ extern "C"
   void pdns_ffi_param_set_padding_disabled(pdns_ffi_param_t* ref, bool disabled) __attribute__((visibility("default")));
   void pdns_ffi_param_add_meta_single_string_kv(pdns_ffi_param_t* ref, const char* key, const char* val) __attribute__((visibility("default")));
   void pdns_ffi_param_add_meta_single_int64_kv(pdns_ffi_param_t* ref, const char* key, int64_t val) __attribute__((visibility("default")));
+
+  typedef struct pdns_postresolve_ffi_handle pdns_postresolve_ffi_handle_t;
+
+  const char* pdns_postresolve_ffi_handle_get_qname(pdns_postresolve_ffi_handle_t* ref) __attribute__((visibility("default")));
+  uint16_t pdns_postresolve_ffi_handle_get_qtype(const pdns_postresolve_ffi_handle_t* ref) __attribute__((visibility("default")));
+  uint16_t pdns_postresolve_ffi_handle_get_rcode(const pdns_postresolve_ffi_handle_t* ref) __attribute__((visibility("default")));
```
wojas, line 118: A `set_rcode` function would be useful too.

pdns/lua-recursor4.cc
omoerbeek: `ref->handle.d_dq.qname` is the original qname from the query, so we have the full raw name available.

pdns/lua-recursor4-ffi.hh
omoerbeek: Likely setting the policy kind is what you want in many cases. But if you are going to build an answer by hand, you would need to set the rcode, yes.

pdns/lua-recursor4.cc
omoerbeek: `serialize()` builds a complete record. I only want the content part. But at the moment of writing I could not find the proper API for that. This is certainly something that should be revisited.

pdns/lua-recursor4.cc
omoerbeek: We likely could provide  the raw name by default, and have a separate method to get the human readable form.

pdns/lua-recursor4.cc
rgacogne: I think we always have the full name available at this point anyway, since we are using `DNSName` objects and not some view into a DNS packet?

omoerbeek:
> Does that work fine when passed in Lua FFI definition? https://luajit.org/ext_ffi_api.html states that preprocessing is not done so I never dared.

What I have seen in existing Lua code: the visibilityb attribute is stripped off in cdef sections.

II see now that dnsdist has a more luxurious interface  that already includes everything needed when setting up Lua. Some day the recursor might do the same, so Ill revert the commit.

pdns/lua-recursor4.cc
omoerbeek: I checked things but the `DNSPacketWriter` code currently needs the existing context setup in `DNSRecordContent::serialize`to work properly. One day we should make it easier to just get the content, i.e. make a version of `DNSRecordContent::serialize()` that does as little as possible for this use case.

pdns/lua-recursor4.cc
wojas: I think that would make sense. I suppose providing the raw name would be almost free?

omoerbeek: > I think that would make sense. I suppose providing the raw name would be almost free?

Indeed, sonce we anot creating a tremp code there's also no need to store the string into the string pool. 

I changed the code to always return a name_len and fill a raw name or a human readable name depending on the raw flag, so both name and content are now either raw or human readable.

wojas: Can the raw name have back references in this context? If so, the raw value may not be useful without a separate function to convert them.

wojas: And if we are going that way, we could also just omit the name and have a separate function to fetch the name with an index parameter.

omoerbeek: No back refs here, we're dealing with objects at this stage. The recursor will create the answer packet later.

pdns/lua-recursor4-ffi.hh
wojas: We are currently not using those fields, but we could if they are useful. It would still be useful to be able to set any rcode we want, just as we have the ability to modify the rrsets anyway we want.

omoerbeek: Latest version has `pdns_postresolve_ffi_handle_set_rcode`

omoerbeek:
@wojas do you think this is good to be merged? After it is merged you cam start using it on master builds, and based on your feedback we can improve & extend.

wojas (approved):
Yes, LGTM!

rgacogne (approved):

PR #11075, opened 2021-12-07
dnsdist: Fix a double-free when a DoH cross-protocol response is dropped

rgacogne:
### Short description

Introduced in 1.7.0, and only when a cross-protocol response to a DoH query is dropped by a rule.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdist-idstate.hh
+++ b/pdns/dnsdist-idstate.hh
@@ -241,7 +241,7 @@ struct IDState
   std::unique_ptr<QTag> qTag{nullptr}; // 8
   boost::optional<uint32_t> tempFailureTTL; // 8
   const ClientState* cs{nullptr}; // 8
-  DOHUnit* du{nullptr}; // 8
+  DOHUnit* du{nullptr}; // 8 (not a unique_ptr because we currently need to be able to peek at it without knowing taking ownership until later)
   std::atomic<int64_t> usageIndicator{unusedIndicator}; // set to unusedIndicator to indicate this state is empty   // 8
   std::atomic<uint32_t> generation{0}; // increased every time a state is used, to be able to detect an ABA issue    // 4
   uint32_t cacheKey{0}; // 4
--- a/pdns/dnsdist-idstate.hh
+++ b/pdns/dnsdist-idstate.hh
@@ -241,7 +241,7 @@ struct IDState
   std::unique_ptr<QTag> qTag{nullptr}; // 8
   boost::optional<uint32_t> tempFailureTTL; // 8
   const ClientState* cs{nullptr}; // 8
-  DOHUnit* du{nullptr}; // 8 (not a unique_ptr because we currently need to be able to peek at it without knowing taking ownership until later)
+  DOHUnit* du{nullptr}; // 8 (not a unique_ptr because we currently need to be able to peek at it without taking ownership until later)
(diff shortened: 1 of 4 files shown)


omoerbeek (approved):
This change looks good.

I checked the places where `DoHTCPCrossQuerySender` and `DoHCrossProtocolQuery` objects are used, and noted one spot where an `DoHTCPCrossQuerySender` object is first created and then the `get()` is done: line 1328. I do not think it is wrong, but the pattern looked strange to me.

rgacogne:
Thanks Otto! I'll review all the places where we touch the reference counter to check that we are doing the right thing. I'll also check one more time if we cannot use a smart pointer instead but h2o's API make that hard, if not impossible.

rgacogne:
It was indeed leaking when a UDP response to a DoH query was truncated. I pushed three commits:
1. fixing that leak
2. making it easier to understand which function has ownership of the object
3. wrapping the object in a `unique_ptr` whenever possible, reducing the amount of manual book-keeping

rgacogne:
2171e7c7ef360c5646646c6504d29d83a74e18c2 does not play well with TSAN, investigating.

rgacogne:
Fixed for now, it looks like TSAN is confused by the fact that the answer from the backend might come back in a UDP responder thread before the `send()` syscall returned in the initial thread.

pdns/dnsdist-idstate.hh
```diff
@@ -241,7 +241,7 @@ struct IDState
   std::unique_ptr<QTag> qTag{nullptr}; // 8
   boost::optional<uint32_t> tempFailureTTL; // 8
   const ClientState* cs{nullptr}; // 8
-  DOHUnit* du{nullptr}; // 8
+  DOHUnit* du{nullptr}; // 8 (not a unique_ptr because we currently need to be able to peek at it without knowing taking ownership until later)
```
omoerbeek, line 244:
```suggestion
  DOHUnit* du{nullptr}; // 8 (not a unique_ptr because we currently need to be able to peek at it without taking ownership until later)
```

omoerbeek, line 244: Suggestion: use `duNaked` or something like that to make it very clear it's not a smart pointer when using it as other instances of `du` hold a `unique_pointer`.

omoerbeek:
In `doh.cc:` 

`/* you can't touch du after this line, because it might already have been freed */`

still `du` is accessed in the block below (this is not part of this PR but it caught my eye anyway). Either the comment should be adapted or something else is going on.

rgacogne:
The comment should say that `du` can't be touched if that call returns a non-negative value, right?

omoerbeek (approved):
Some small remarks. Ref counting is inherently a bit of a scary subject as it is easy to get wrong, but I do think this approach is better than before.

rgacogne:
Thanks, Otto! I pushed a few commits following your advice, I'm still pondering the `duNaked` one, I can't decide whether I like that idea.

omoerbeek:
> The comment should say that `du` can't be touched if that call returns a non-negative value, right?

yes, that is more clear

rgacogne:
I'm going to merge this without changing the `duNaked` stuff because I'd like to release rc1 quickly, and also because I hope to be able to use `unique_ptr`s everywhere in 1.8.0 anyway.

PR #11079, opened 2021-12-08
dnsdist: Check the size of the query when re-sending a DoH query

rgacogne:
### Short description

When the UDP response to a DoH query was truncated, we re-send the query via TCP, passing it to a TCP worker. We need to edit the ID to its original value before that, and while there is no reason that the query is smaller than a dnsheader, we need to check its size, not the size of the response.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdistdist/doh.cc
+++ b/pdns/dnsdistdist/doh.cc
@@ -1320,7 +1320,10 @@ static void on_dnsdist(h2o_socket_t *listener, const char *err)
       continue;
     }
 
-    if (!du->tcp && du->truncated && du->query.size() > sizeof(dnsheader)) {
+    if (!du->tcp &&
+        du->truncated &&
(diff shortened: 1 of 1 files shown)


pdns/dnsdistdist/doh.cc
```diff
@@ -1320,7 +1320,7 @@ static void on_dnsdist(h2o_socket_t *listener, const char *err)
       continue;
     }
 
-    if (!du->tcp && du->truncated && du->response.size() > sizeof(dnsheader)) {
+    if (!du->tcp && du->truncated && du->query.size() > sizeof(dnsheader)) {
```
Habbie, line 1323: Should this also take into account `du->proxyProtocolPayloadSize`?

rgacogne: It should!

PR #11081, opened 2021-12-08
Docker images: Remove capability requirements

nvaatstra:
### Short description
The current Docker images are built with a dependency on the NET_BIND_SERVICE capability which used to be required for binding to a privileged port. This capability included in the image causes problems when trying to deploy the image on a Kubernetes platform with restrictive security policies in place (which is becoming the norm).

This pull request removes those dependencies and adds instructions to:
- Avoid these privileged ports (by binding to non-privileged ports)
- Add the NET_BIND_SERVICE capability to containers

### Checklist
I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/Docker-README.md
+++ b/Docker-README.md
@@ -53,3 +53,5 @@ There are multiple ways of dealing with these restrictions if you encounter them
 * Use custom configuration files to bind to alternate ports outside of the privileged range. This can be done via the following configuration settings:
     * dnsdist: `setLocal()`
     * Auth & Recursor: `local-address` and/or `local-port`
+
+Note: Docker Engine 20.10.0 (released december 2020) removed the need to set the `NET_BIND_SERVICE` capability when attempting to bind to a privileged port. 
--- a/Docker-README.md
+++ b/Docker-README.md
@@ -42,3 +42,14 @@ In a plain Docker or Compose setup, this can be done by using the host PID names
 We have a Docker Compose example at https://github.com/PowerDNS/pdns/blob/master/docker-compose.yml .
 It brings up all three services, and exposes them to eachother by name (using Docker's internal DNS).
 In the dockerdata dir, you can find an example dnsdist Lua config (with Python helper to make DNS lookups non-blocking for dnsdist) for managing your auth/rec backends by name.
+
+# Privileged ports
+
+The default configurations included for dnsdist, Auth and Recursor attempt to bind to port 53, which may not be permitted by the platform on which you intend to use these images. Kubernetes clusters, for example, might have a restriction on binding to privileged ports unless the `NET_BIND_SERVICE` capability is explicitly added to the container's security context.
+
+There are multiple ways of dealing with these restrictions if you encounter them:
+
+* Grant the `NET_BIND_SERVICE` capability to the containers which utilize these images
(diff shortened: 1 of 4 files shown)


Habbie (approved):

Habbie:
@omoerbeek @rgacogne a backport to 4.6 and 1.7 would be very convenient. Any objections?

rgacogne:
I'm not sure I understand what happens by default after that PR is merged. The default ports are still privileged but the services no longer have the capability to bind to these ports, right? In that case can the services start without changing the configuration? Is running the container in privileged mode enough, even though we run under a non-privileged user?

Habbie:
> Is running the container in privileged mode enough, even though we run under a non-privileged user?

Nico has just extended the README to note that since Docker 20.10.0, binding <1024 simply is allowed by default. Specifically, 20.10.0 and up set `net.ipv4.ip_unprivileged_port_start = 0` inside containers.

rgacogne:
Ah, I missed that, thanks. That's nice. And I guess before that change we can't start even if the container runs in privileged mode?

Habbie:
(users with other runtimes may need to do work, as the readme says, so perhaps upgrade notes are in order, I can do that)

Habbie:
> Ah, I missed that, thanks. That's nice. And I guess before that change we can't start even if the container runs in privileged mode?

Before that, privileged mode works, or the `setcap` that this PR removes.

rgacogne:
So when running in privileged mode and with a non-root user, we still keep all capabilities?

Habbie:
> So when running in privileged mode and with a non-root user, we still keep all capabilities?

Uh, I think so yes.

rgacogne (approved):

Habbie:
> So when running in privileged mode and with a non-root user, we still keep all capabilities?

After more investigation: yes, Docker's `USER` is not great at dropping capabilities, but dnsdist is, so we're safe.

rgacogne:
Note that the recursor also calls `dropCapabilities` but it looks like the authoritative server doesn't.

Habbie:
Proposed text for upgrade notes for all 3 products:

Privileged port binding in Docker
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

In our Docker image, our binaries no longer get the ``net_bind_service`` capability, as this is unnecessary in many deployments.
For more information, see the section "Privileged ports" in [Docker-README](https://github.com/PowerDNS/pdns/blob/master/Docker-README.md#privileged-ports).

jsoref:
> Proposed text for upgrade notes for all 3 products:
> 
> ```
> Privileged port binding in Docker
> ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> 
> In our Docker image, our binaries no longer get the ``net_bind_service`` capability, as this is unnecessary in many deployments.

I'm not sure about the word `get`, do you mean `ask for`, `are assigned`, ...?

> For more information, see the section "Privileged ports" in [Docker-README](https://github.com/PowerDNS/pdns/blob/master/Docker-README.md#privileged-ports).
> ```

rgacogne:
Or perhaps `are granted`?

jsoref:
As `are no longer granted`

PR #11090, opened 2021-12-10
auth: show QType by name in log message

zeha:
### Short description

"Should not get here" log message previously showed the QType as the numeric code, instead of the more common name (ex: instead of A, it showed 1).

### Checklist

I have:
- [x] compiled this code
- [ ] tested this code
- [ ] documented the code


Habbie (approved):

zeha:
Set a milestone so this is not forgotten.

PR #11098, opened 2021-12-14
dnsdist: lua support to remove resource records from a response

chbruyand:
### Short description

This PR adds lua support to remove resource records from a response.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdist-lua-actions.cc
+++ b/pdns/dnsdist-lua-actions.cc
@@ -1734,7 +1734,7 @@ public:
     if (d_qtypes.size() > 0) {
       clearDNSPacketRecordTypes(dr->getMutableData(), d_qtypes);
     }
-    return DNSResponseAction::Action::HeaderModify;
+    return DNSResponseAction::Action::None;
   }
 
   std::string toString() const override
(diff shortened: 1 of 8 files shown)


pdns/dnsdistdist/docs/rules-actions.rst
```diff
@@ -838,6 +838,14 @@ The following actions exist.
 
   Let these packets go through.
 
+.. function::ClearRecordTypesResponseAction(types)
+
+  .. versionadded:: 1.8.0
+
+  Removes given type(s) records from the response.
```
pieterlexis, line 845: some extra text would be good that the operator has to be careful about accidentally turning an answer into a NODATA response without a SOA record in the additional section. This could be prevented by using a `NotRule{QTypeRule(XXX)}` where XXX are the record types being stripped in the response.

pdns/dnsdist-lua-actions.cc
```diff
@@ -1712,6 +1712,32 @@ class SetTagResponseAction : public DNSResponseAction
...
+  ClearRecordTypesResponseAction(const std::set<QType>& qtypes) : d_qtypes(qtypes)
+  {
+  }
+
+  DNSResponseAction::Action operator()(DNSResponse* dr, std::string* ruleresult) const override
+  {
+    if (d_qtypes.size() > 0) {
+      clearDNSPacketRecordTypes(dr->getMutableData(), d_qtypes);
+    }
+    return DNSResponseAction::Action::HeaderModify;
```
rgacogne, line 1729: I _think_ we might not want to stop the processing of subsequent rules after this one, what do you think? That would mean returning `DNSResponseAction::Action::None` instead., listing it in the list of actions that do not stop the processing in https://dnsdist.org/rules-actions.html#actions (I wonder why that list is broken, btw), and adding `Subsequent rules are processed after this action.` to its description.

rgacogne (approved):
Nice work!

pdns/dnsdist-lua-actions.cc
chbruyand: "Subsequent rules are *not* processed after this action" you mean ?

rgacogne: Well it depends :) If we keep returning `DNSResponseAction::Action::HeaderModify` then we do not have to change anything, but if we decide to return `DNSResponseAction::Action::None` instead then `Subsequent rules are processed after this action` would be correct, right?

chbruyand: My bad, thanks!

pdns/dnsdistdist/docs/rules-actions.rst
chbruyand: Alright, thanks! I added a word about that and an example code.

PR #11100, opened 2021-12-15
First part of ZONEMD support

omoerbeek:
### Short description

1) Define record type
2) Verification according to RFC, including many failure cases and RSIG handling. Support sha384 and sha512
3) Unit tests, including all examples from RFC and a few failure cases.
4) Command line for pdnsutil: zonemd-verify-file

Based on POC from @Habbie, with many changes.

Things to be done I can think of (details to be decided):
 
For the Recursor: 
- validation of incoming zones (potentially Zone to Cache, RPZ, Auth-Zones). Those cases need to consider DNSSEC validation of ZONEMD records as well per RFC. Also config of zone should indicate if ZONEMD verification is required. For Zone to Cache and RPZ we can do that from Lua, for Auth Zones it is not immediately clear how.

For the Authoritative Server:
- verify/insert/update/remove ZONEMD in DB 
- send ZONEMD out on AXFR. 
- verify incoming AXFR ZONEMD. 

Both should be handled in separate PRs imo.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [X] documented the code

### DIFF
--- a/pdns/pdnsutil.cc
+++ b/pdns/pdnsutil.cc
@@ -3200,7 +3200,7 @@ try
   }
   else if (cmds.at(0) == "unset-nsec3") {
     if(cmds.size() < 2) {
-      cerr<<"Syntax: pdnsutil unset-nsec3 ZON"<<endl;
+      cerr<<"Syntax: pdnsutil unset-nsec3 ZONE"<<endl;
       return 0;
     }
     if (!dk.unsetNSEC3PARAM(DNSName(cmds.at(1)))) {
--- a/pdns/pdnsutil.cc
+++ b/pdns/pdnsutil.cc
@@ -1390,7 +1390,64 @@ static int zonemdVerifyFile(const DNSName& zone, const string& fname) {
   std::map<std::pair<uint8_t, uint8_t>, std::shared_ptr<ZONEMDRecordContent>> zonemdRecords;
   std::shared_ptr<SOARecordContent> soarc;
 
-  while(zpt.get(rr)) {
+  class SHADigest
+  {
+  public:
+    SHADigest(int bits)
+    {
+      mdctx = EVP_MD_CTX_new();
+      if (mdctx == nullptr) {
+        throw std::runtime_error("VSHADigest: P_MD_CTX_new failed");
+      }
+      switch (bits) {
+      case 256:
+        md = EVP_sha256();
+        break;
+      case 384:
+        md = EVP_sha384();
+        break;
+      case 512:
+        md = EVP_sha512();
+        break;
+      default:
+        throw std::runtime_error("SHADigest: unsupported size");
+      }
+      if (EVP_DigestInit_ex(mdctx, md, NULL) == 0) {
+        throw std::runtime_error("SHADigest: init error");
+      }
+    }
+    ~SHADigest()
+    {
+      // No free of md needed afaik
+      if (mdctx != nullptr) {
+        EVP_MD_CTX_free(mdctx);
+      }
+    }
+    void process(const string& msg, size_t sz)
+    {
+      if (EVP_DigestUpdate(mdctx, msg.data(), msg.size()) == 0) {
+        throw std::runtime_error("SHADigest: update error");
+      }
+    }
+    std::string digest()
+    {
+      string md_value;
+      md_value.resize(EVP_MD_size(md));
+      unsigned int md_len;
+      if (EVP_DigestFinal_ex(mdctx, reinterpret_cast<unsigned char*>(md_value.data()), &md_len) == 0) {
+        throw std::runtime_error("SHADigest: finalize error");
+      }
+      if (md_len != md_value.size()) {
+        throw std::runtime_error("SHADigest: inconsisten size");
+      }
+      return md_value;
+    }
+  private:
+    EVP_MD_CTX *mdctx{nullptr};
+    const EVP_MD *md;
+  };
+
+  while (zpt.get(rr)) {
     if(!rr.qname.isPartOf(zone) && rr.qname!=zone) {
       continue;
       cerr<<"File contains record named '"<<rr.qname<<"' which is not part of zone '"<<zone<<"'"<<endl;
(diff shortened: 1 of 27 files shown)


pdns/sha.hh
```diff
@@ -51,3 +52,66 @@ inline std::string pdns_sha512sum(const std::string& input)
...
+        md = EVP_sha256();
+        break;
+      case 384:
+        md = EVP_sha384();
+        break;
+      case 512:
+        md = EVP_sha512();
+        break;
+      default:
+        throw std::runtime_error("SHADigest: unsupported size");
```
pieterlexis: maybe use `std::out_of_range`?

pdns/sha.hh
```diff
@@ -51,3 +52,67 @@ inline std::string pdns_sha512sum(const std::string& input)
   SHA512(reinterpret_cast<const unsigned char*>(input.c_str()), input.length(), result);
   return std::string(result, result + sizeof result);
 }
+
+namespace pdns
+{
+class SHADigest
+{
+public:
+  SHADigest(int bits)
```
rgacogne, line 61:
```suggestion
  SHADigest(unsigned int bits)
```

pdns/sha.hh
```diff
@@ -51,3 +52,67 @@ inline std::string pdns_sha512sum(const std::string& input)
...
+      md = EVP_sha256();
+      break;
+    case 384:
+      md = EVP_sha384();
+      break;
+    case 512:
+      md = EVP_sha512();
+      break;
+    default:
+      throw std::runtime_error("SHADigest: unsupported size");
```
rgacogne, line 78: This leaks `mdctx`. Perhaps we could wrap it in a `unique_ptr` right away?

pdns/sha.hh
```diff
@@ -51,3 +52,67 @@ inline std::string pdns_sha512sum(const std::string& input)
...
+
+  ~SHADigest()
+  {
+    // No free of md needed afaik
+    if (mdctx != nullptr) {
+      EVP_MD_CTX_free(mdctx);
+    }
+  }
+
+  void process(const std::string& msg, size_t sz)
```
rgacogne, line 93:
```suggestion
  void process(const std::string& msg)
```

pdns/sha.hh
```diff
@@ -51,3 +52,67 @@ inline std::string pdns_sha512sum(const std::string& input)
...
+  std::string digest()
+  {
+    std::string md_value;
+    md_value.resize(EVP_MD_size(md));
+    unsigned int md_len;
+    if (EVP_DigestFinal_ex(mdctx, reinterpret_cast<unsigned char*>(md_value.data()), &md_len) == 0) {
+      throw std::runtime_error("SHADigest: finalize error");
+    }
+    if (md_len != md_value.size()) {
+      throw std::runtime_error("SHADigest: inconsisten size");
```
rgacogne, line 109:
```suggestion
      throw std::runtime_error("SHADigest: inconsistent size");
```

pdns/zonemd.cc
```diff
@@ -0,0 +1,187 @@
...
+};
+
+typedef std::map<rrSetKey_t, rrVector_t, CanonrrSetKeyCompare> RRsetMap_t;
+
+void pdns::zonemdVerify(const DNSName& zone, ZoneParserTNG& zpt, bool& validationDone, bool& validationOK)
+{
+  validationDone = false;
+  validationOK = false;
+
+  // scheme,hasalgo -> duplicate,zonemdrecord
```
rgacogne, line 33:
```suggestion
  // scheme,hashalgo -> duplicate,zonemdrecord
```

pdns/pdnsutil.cc
```diff
@@ -3154,7 +3200,7 @@ try
   }
   else if (cmds.at(0) == "unset-nsec3") {
     if(cmds.size() < 2) {
-      cerr<<"Syntax: pdnsutil unset-nsec3 ZONE"<<endl;
+      cerr<<"Syntax: pdnsutil unset-nsec3 ZON"<<endl;
```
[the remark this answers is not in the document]
omoerbeek: Dunno what happened there

pdns/sha.hh
omoerbeek: `Defines a type of object to be thrown as exception. It reports errors that are consequence of attempt to access elements out of defined range.`
So close, but not really it.

pdns/sha.hh
```diff
@@ -51,3 +52,64 @@ inline std::string pdns_sha512sum(const std::string& input)
   SHA512(reinterpret_cast<const unsigned char*>(input.c_str()), input.length(), result);
   return std::string(result, result + sizeof result);
 }
+
+namespace pdns
+{
+class SHADigest
```
chbruyand, line 58: Maybe you could disable default constructor ?

chbruyand (approved):
Didn't dive too much into `zonemdVerify()` as I don't *know* ZONEMD and found it a bit hard to read (lots of abbreviations that are lowercased/uppercased), but LGTM as it's tested and the API is readable

omoerbeek:
> Didn't dive too much into `zonemdVerify()` as I don't _know_ ZONEMD and found it a bit hard to read (lots of abbreviations that are lowercased/uppercased), but LGTM as it's tested and the API is clear and speaks for itself

Good point, I'll rename some vars

pdns/sha.hh
omoerbeek: Default constructor is implicitly deleted if another one exists I think

pdns/sha.hh
```diff
@@ -51,3 +52,64 @@ inline std::string pdns_sha512sum(const std::string& input)
...
+      throw std::invalid_argument("SHADigest: unsupported size");
+    }
+    if (EVP_DigestInit_ex(mdctx.get(), md, NULL) == 0) {
+      throw std::runtime_error("SHADigest: init error");
+    }
+  }
+
+  ~SHADigest()
+  {
+    // No free of md needed afaik
```
Habbie, line 87: I agree - so let's remove the afaik and mention the unique_ptr?

pdns/zonemd.cc
```diff
@@ -0,0 +1,187 @@
...
+#include "zoneparser-tng.hh"
+
+typedef std::pair<DNSName, QType> RRSetKey_t;
+typedef std::vector<std::shared_ptr<DNSRecordContent>> RRVector_t;
+
+struct CanonRRSetKeyCompare : public std::binary_function<RRSetKey_t, RRSetKey_t, bool>
+{
+  bool operator()(const RRSetKey_t& a, const RRSetKey_t& b) const
+  {
+    // FIXME surely we can be smarter here
```
Habbie, line 15: I believe this is my FIXME, I'll have a go at it

pdns/zonemd.cc
```diff
@@ -0,0 +1,187 @@
...
+
+  DNSResourceRecord dnsResourceRecord;
+
+  // Get all records and remember RRSets and TTLs
+  while (zpt.get(dnsResourceRecord)) {
+    if (!dnsResourceRecord.qname.isPartOf(zone) && dnsResourceRecord.qname != zone) {
+      continue;
+    }
+    if (dnsResourceRecord.qtype == QType::SOA && soaRecordContent) {
+      // XXX skip extra SOA?
```
Habbie, line 54: We should handle or remove this comment. I'm leaning towards remove.

pdns/zonemd.cc
```diff
@@ -0,0 +1,187 @@
...
+      throw PDNSException(err);
+    }
+    if (dnsResourceRecord.qtype == QType::SOA && dnsResourceRecord.qname == zone) {
+      soaRecordContent = std::dynamic_pointer_cast<SOARecordContent>(drc);
+    }
+    if (dnsResourceRecord.qtype == QType::ZONEMD && dnsResourceRecord.qname == zone) {
+      auto zonemd = std::dynamic_pointer_cast<ZONEMDRecordContent>(drc);
+      auto inserted = zonemdRecords.insert({pair(zonemd->d_scheme, zonemd->d_hashalgo), {zonemd, false}});
+      if (!inserted.second) {
+        // Mark as duplicate;
```
Habbie, line 76: CodeQL thinks this is code; perhaps removing the ; will fix that :)

Habbie (approved):
Awesome work. I posted a few nits, but nothing major.

pdns/zonemd.cc
Habbie: Ah yes, can't use `std::tie` because of the custom comparison function. I'll leave it for now..

PR #11115, opened 2021-12-20
auth: lmdb: fix records removal in deleteDomain()

### LINKED ISSUE #11114: lmdb: deleting secondary zone does not delete records from lmdb
- Program: Authoritative
 - Issue type: Bug report

### Short description
Deleting a secondary zone does not delete associated records. After deleting a secondary zone and creating a new one, the new secondary zone is populated with the records of the deleted zone.

### Environment

 - Operating system: Docker
 - Software version:  4.6
 - Software source: tested with docker images provided by powerdns via docker hub (tested both powerdns/pdns-auth-45:latest and powerdns/pdns-auth-46:latest)

### Steps to reproduce

1. Create a new zone on the primary, set ALLOW-AXFR-FROM and add some records, raise serial
2. Create the same zone on the secondary: `pdnsutil create-secondary-zone domain ip`
3. Make sure the zone is transferred to the secondary
4. Remove the zone on the secondary: `pdnsutil delete-zone domain`
5. Create a new zone on the secondary: `pdnsutil create-secondary-zone otherdomain ip`
6. List the newly created zone on the secondary: `pdnsutil list-zone otherdomain`

### Expected behaviour
An empty zone should be returned on step 7

### Actual behaviour
The records of the previously deleted zone appear in the newly created zone.

A zone transfer might not fix it, as the serial might be lower than that of the newly created zone on the primary.

### Other information
This only works when you delete the most recently added domain on the secondary, so it looks like the records remain in the database, pointing to the id the deleted domain used to have.

Also, this only seems to work on secondary zones, couldn't reproduce this behaviour with for primary/native zones.

Used the following docker-compose.yml and minimal pdns configs:
https://github.com/cedricdubois/pdns-auth-test

mind04:
fixes #11114

### Checklist
I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/modules/lmdbbackend/lmdbbackend.cc
+++ b/modules/lmdbbackend/lmdbbackend.cc
@@ -648,16 +648,7 @@ bool LMDBBackend::deleteDomain(const DNSName& domain)
   }
 
   doms.del(id);
-  compoundOrdername co;
-  string match = co(id);
-
-  auto cursor = txn->txn->getCursor(txn->db->dbi);
-  MDBOutVal key, val;
-  if (!cursor.find(match, key, val)) {
-    do {
-      cursor.del();
-    } while (!cursor.next(key, val) && key.get<StringView>().rfind(match, 0) == 0);
-  }
+  deleteDomainRecords(*txn, id);
 
   if (needCommit)
     txn->txn->commit();


RobinGeuze:
Could potentially use a test case. Also might be good to build a pdnsutil function to clean up stuff from existing bad databases?

rgacogne (commented):
Am I correctly understanding that the fix is that we need to do a range-based lookup because we will not have an exact match since the key is a composite one, containing the domain ID, the record name and type?
If so that change looks good to me.

mind04:
> Am I correctly understanding that the fix is that we need to do a range-based lookup because we will not have an exact match since the key is a composite one, containing the domain ID, the record name and type?

You are right, this was the problem.

rgacogne (approved):

PR #11124, opened 2021-12-22
rec: enable recursor regression tests using tsan

omoerbeek:
### Short description

Some modifications were needed to make tsan work for rec regression test:
- Update `numberOfDistributedQueries` only from the thread itself
- Do not `use boost::format` (it calls into locale handling code that does not seem to be thread-safe)
- Stop printing periodic stats (they are racey, but we do not care)

Survived a couple of test runs here.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/recursordist/recursor-tsan.supp
+++ b/pdns/recursordist/recursor-tsan.supp
@@ -1,3 +1,2 @@
 # We don't care about stats for now
-race:doStats
 race:g_stats
--- a/pdns/recursordist/recursor-tsan.supp
+++ b/pdns/recursordist/recursor-tsan.supp
@@ -1,2 +1,3 @@
 # We don't care about stats for now
+race:doStats
 race:g_stats
--- a/pdns/syncres.cc
+++ b/pdns/syncres.cc
@@ -104,6 +104,19 @@ int SyncRes::s_event_trace_enabled;
 
 #define LOG(x) if(d_lm == Log) { g_log <<Logger::Warning << x; } else if(d_lm == Store) { d_trace << x; }
 
+// A helper function to print a double with specific precsision
+// Not using boost::format since it is not thread safe while calling into locale handling code according to tsan
+// This allocates a string, but that's nothing compared to what boost::format is doing
+static inline std::string fmtfloat(const char* fmt, double f)
+{
+  char buf[20];
+  int ret = snprintf(buf, sizeof(buf), fmt, f);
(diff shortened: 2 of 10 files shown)


pdns/syncres.cc
```diff
@@ -104,6 +104,21 @@ int SyncRes::s_event_trace_enabled;
 
 #define LOG(x) if(d_lm == Log) { g_log <<Logger::Warning << x; } else if(d_lm == Store) { d_trace << x; }
 
+// A helper function to print a double with specific printf format.
+// Not using boost::format since it is not thread safe while calling
+// into locale handling code according to tsan.
+// This allocates a string, but that's nothing compared to what
+// boost::format is doing and maybe even gets optimized away anyway.
```
jsoref, line 111:
```suggestion
// boost::format is doing and may even be optimized away anyway.
```

pdns/syncres.cc
```diff
@@ -104,6 +104,21 @@ int SyncRes::s_event_trace_enabled;
...
+// This allocates a string, but that's nothing compared to what
+// boost::format is doing and may even be optimized away anyway.
+static inline std::string fmtfloat(const char* fmt, double f)
+{
+  char buf[20];
+  int ret = snprintf(buf, sizeof(buf), fmt, f);
+  if (ret < 0 || ret >= static_cast<int>(sizeof(buf))) {
+    return "?";
+  }
+  return std::string(buf, strlen(buf));
```
rgacogne, line 119: Perhaps we could use `ret` instead of calling `strlen` ?

pdns/recursordist/recursor-tsan.supp
```diff
@@ -1,3 +1,2 @@
 # We don't care about stats for now
-race:doStats
```
rgacogne, line 2: I guess we don't need that anymore because we set `statistics-interval=0`? In that case I would prefer keeping the suppression though, as the suppression file might be used outside of the regression tests. I, for one, use it when I run with TSAN support enabled :)

rgacogne (approved):
Nice, looks good to me! Do you know if the code that is not run anymore because of `statistics-interval=0` is tested in any other way?

pdns/recursordist/recursor-tsan.supp
omoerbeek: Will do, but in some case TSAN was not able to produce the correct backtrace, so the suppression of `doStats` did not work (as it is dependent on a proper unwinding of the stack).

PR #11126, opened 2021-12-22
dnsdist: add newThread() function (experimental, subject to change)

Habbie:
### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/pdns/dnsdist-lua.cc
+++ b/pdns/dnsdist-lua.cc
@@ -264,6 +264,40 @@ void checkParameterBound(const std::string& parameter, uint64_t value, size_t ma
   }
 }
 
+static void LuaThread(const std::string code)
+{
+  LuaContext l;
+  // submitToMainThread is camelcased, threadmessage is not.
+  // This follows our tradition of hooks we call being lowercased but functions the user can call being camelcased.
+  l.writeFunction("submitToMainThread", [](std::string cmd, std::unordered_map<std::string, std::string> data) {
+    auto lua = g_lua.lock();
+    // maybe offer more than `void`
+    auto func = lua->readVariable<boost::optional<std::function<void(std::string cmd, std::unordered_map<std::string, std::string> data)>>>("threadmessage");
+    if (func) {
+      func.get()(cmd, data);
+    }
+    else {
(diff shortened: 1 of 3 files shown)


pdns/dnsdist-lua.cc
```diff
@@ -264,6 +264,40 @@ void checkParameterBound(const std::string& parameter, uint64_t value, size_t ma
...
+      errlog("luathread called submitToMainThread but no threadmessage receiver is defined");
+    }
+  });
+
+  // function threadmessage(data) print("got thread data:") for k,v in pairs(data) do print(k,v) end end
+
+  for (;;) {
+    try {
+      l.executeCode(code);
+      errlog("luaThread exited, restarting in 5 seconds");
```
rgacogne, line 289: I wonder if it would make sense to make that behaviour optional, so that a Lua thread can actually exit when it's done?

rgacogne (commented):
Looks good, could use some documentation :-)

pdns/dnsdist-lua.cc
Habbie: I also wondered that!

regression-tests.dnsdist/test_Lua.py
```diff
@@ -0,0 +1,42 @@
...
+      while true
+      do
+          socket.sleep(1)
+          submitToMainThread("setCounter", {i=i})
+          i = i + 1
+      end
+    ]==])
+
+    counter = 0
+    function threadmessage(cmd, data)
```
rgacogne, line 29: Just a nit but perhaps it would be better to define that function before calling `newThread()`? It does not matter here before the background thread will sleep for one second which should be enough for us to return, though.

rgacogne (approved):

regression-tests.dnsdist/test_Lua.py
Habbie: I'll fix that

Habbie: pushed

rgacogne:
CI failure is unrelated (usual `Basics: test if addAction accepts a DNSName` response getting lost in flight).

PR #11132, opened 2021-12-27
Test correct member in DynBlockRatioRule::warningRatioExceeded

### LINKED ISSUE #11131: dnsdist: setRCodeRatio Warning mechanism doesn't seem to work
- Program: dnsdist 
 - Issue type: Bug report

### Short description
The  Warning parameter in the `setRCodeRatio()` function, isn't triggered

### Environment
 - Operating system: Centos 8
 - Software version: 1.6.1
 - Software source:  PowerDNS repository
 - 
### Steps to reproduce
1. Configure dnsdist with:
```
dbr:setRCodeRatio(DNSRCode.NXDOMAIN, 0.9, 10, "Exceeded NXD ratio", 60, 1, DNSAction.Drop, 0.1)

function maintenance()
  dbr:apply()
end
```

2. Try to trigger the Warning by querying random names returning an NXDomain, alternated with entries that do return a value. Don't forget to disable your packetCache for the test.

### Expected behaviour
As long as more than 10% but less than 90% of the queries return an NXDomain, I would expect a Warning to be issued. This should be visible in the system logs, as in the `showDynBlocks()` output. The test above should generate 50% NXDomains so the Warning should trigger.

### Actual behaviour
No warning is triggered. By skipping the "correct" lookup and thus pushing the ratio above 90%, we do see the block happening.

### Other information
See info to be provided by @omoerbeek  and @dwfreed 

While looking at the issue, might clarify the docs as well (double vs int, example code?)

dwfreed:
### Short description

`d_warningRate` in this function will always be 0, as that's how it's initialized and it's never updated, so we need to check `d_warningRatio` instead to see if warnings are enabled.

This also adds some more tests of `d_enabled` as the correct short-circuit for when these are initialized with their zero-argument constructor.

Fixes #11131

### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdist-dynblocks.hh
+++ b/pdns/dnsdist-dynblocks.hh
@@ -108,6 +108,10 @@ private:
 
     bool warningRateExceeded(unsigned int count, const struct timespec& now) const
     {
+      if (!d_enabled) {
+        return false;
+      }
+
       if (d_warningRate == 0) {
         return false;
       }
@@ -177,7 +181,11 @@ private:
 
     bool warningRatioExceeded(unsigned int total, unsigned int count) const
     {
-      if (d_warningRate == 0) {
+      if (!d_enabled) {
+        return false;
+      }
+
+      if (d_warningRatio == 0.0) {
         return false;
       }
 


rgacogne (approved):

dwfreed:
Backport note: issue has existed since ratio blocking was introduced in dnsdist 1.5.0, so all versions since then are affected.

PR #11137, opened 2021-12-30
Try shorter thread names

jsoref:
https://man7.org/linux/man-pages/man3/pthread_setname_np.3.html

       ... The thread name is a
       meaningful C language string, whose length is restricted to 16
       characters, including the terminating null byte ('\0').

### Short description

User reports:
> Could not set thread name pdns-r/ztc/rpz.urlhaus.abuse.ch. for thread: Numerical result out of range

the `pdns-r/ztc` is a really expensive prefix for a 15 char threadname, I'd personally suggest using `ztc:` instead, that'd give 6 chars to the thread name that are currently wasted.

### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/threadname.cc
+++ b/pdns/threadname.cc
@@ -39,7 +39,7 @@
 
 #include "threadname.hh"
 
-void setThreadName(const std::string& threadName) {
+int trySetThreadName(const std::string& threadName) {
   int retval = 0;
 
 #ifdef HAVE_PTHREAD_SETNAME_NP_2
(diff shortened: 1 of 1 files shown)


omoerbeek:
Thanks! This was on my todo for the next week :-)

pdns/threadname.cc
```diff
@@ -39,7 +39,7 @@
 
 #include "threadname.hh"
 
-void setThreadName(const std::string& threadName) {
+int trySetThreadName(const std::string& threadName) {
```
omoerbeek, line 42:
```suggestion
static int trySetThreadName(const std::string& threadName) {
```

omoerbeek (approved):
The string length 15 limit is Linux specific, but there is no portable way to determine the max.
Many other systems truncate silently, so the `ERANGE` is likely to happen on Linux only.

PR #11140, opened 2022-01-03
Sort by content if name and type are equal

decryptus:
### Short description
Sort webservice records and comments by content if name and type are equal.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/ws-auth.cc
+++ b/pdns/ws-auth.cc
@@ -412,6 +412,9 @@ static void fillZone(UeberBackend& B, const DNSName& zonename, HttpResponse* res
               /* if you ever want to update this comparison function,
                  please be aware that you will also need to update the conditions in the code merging
                  the records and comments below */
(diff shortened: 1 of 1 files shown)


pdns/ws-auth.cc
```diff
@@ -430,6 +433,9 @@ static void fillZone(UeberBackend& B, const DNSName& zonename, HttpResponse* res
               /* if you ever want to update this comparison function,
                  please be aware that you will also need to update the conditions in the code merging
                  the records and comments below */
+              if (a.qname == b.qname && b.qtype == a.qtype) {
```
omoerbeek, line 436: To avoid repeating compares, I'd rather have a nested if else here

decryptus: Hi @omoerbeek 
I hope new changes will be good for you.

PR #11141, opened 2022-01-04
Stop using the deprecated boost::shared_array

rgacogne:
### Short description

It was deprecated in 1.65.0: `This facility is deprecated because a shared_ptr to T[] or T[N] is now available, and is superior in every regard`.

As far as I can tell we never actually used the reference counting 'shared' feature anyway.

I have only partially tested this change, in particular I did not test the `ports` and `kqueue` multiplexers.
 
### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


omoerbeek (approved):
Looks good and builds, tests and runs fine on OpenBSD, using kqueue.

PR #11147, opened 2022-01-06
doc: clarify tinydns wildcard divergence

jsoref:
### Short description

Clarify what this section does / doesn't cover

### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code


Habbie:
Both the old and the new text are wrong. I'll leave this open for a bit in case I have time to come up with better text.

docs/backends/tinydns.rst
zeha: link could be updated to https://jdebp.uk/FGA/djbdns-problems.html

jsoref: I think it's already there, it's just that this PR is so stale that the context has shifted

PR #11157, opened 2022-01-10
Auth: add zone removal to the zone cache

mind04:
### Short description
This pull fixes:
- Removal of a zone via the api did not remove the zone from the zonecache.
- New zones did not show up in the stats until the next full refresh.

The ueberBackand zonecache tests are broken (never worked). The test itself is fixed but cannot function without getAllDomans() in SimpleBackend. Commented out the test for now.

### Checklist
I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/docs/.gitignore
+++ b/docs/.gitignore
@@ -17,3 +17,4 @@
 /doctrees
 html-docs.tar.bz2
 /mans
+/mans.tmp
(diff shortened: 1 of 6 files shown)


RobinGeuze:
Don't you also want to hook this in on the places where a zone actually gets removed, rather than only purging it once you encounter an error or trigger cache purging manually?

mind04:
> Don't you also want to hook this in on the places where a zone actually gets removed, rather than only purging it once you encounter an error or trigger cache purging manually?

No, because that will mess up transactions

PR #11168, opened 2022-01-11
rec: servfail.nl changed setup

omoerbeek:
This was prompted by a change in one of the external tests we do.
servfail.nl/AAAA should maybe return edns error code 9 (DNSKEY Missing), but it is not clear 1) if that is correct and 2) how to do that in the current code

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/regression-tests.recursor-dnssec/test_ExtendedErrors.py
+++ b/regression-tests.recursor-dnssec/test_ExtendedErrors.py
@@ -119,7 +119,7 @@ extended-resolution-errors=yes
             self.assertEqual(res.edns, 0)
             self.assertEqual(len(res.options), 1)
             self.assertEqual(res.options[0].otype, 15)
-            self.assertEqual(res.options[0], extendederrors.ExtendedErrorOption(7, b''))
+            self.assertEqual(res.options[0], extendederrors.ExtendedErrorOption(6, b''))
 
     def testBogus(self):
         qname = 'bogussig.ok.bad-dnssec.wb.sidnlabs.nl.'


pdns/recursordist/test-syncres_cc4.cc
```diff
@@ -1022,7 +1022,7 @@ BOOST_AUTO_TEST_CASE(test_dnssec_bogus_rrsig_signed_with_unknown_dnskey)
   vector<DNSRecord> ret;
   int res = sr->beginResolve(target, QType(QType::NS), QClass::IN, ret);
   BOOST_CHECK_EQUAL(res, RCode::NoError);
-  BOOST_CHECK_EQUAL(sr->getValidationState(), vState::BogusNoValidRRSIG);
+  BOOST_CHECK_EQUAL(sr->getValidationState(), vState::BogusNoValidDNSKEY);
```
rgacogne, line 1025: That does not seem right to me, as in that case we have a valid DNSKEY corresponding to the DS but the RRSIG has been signed by a different key that is not present in the DNSKEY set.
rfc8914 defines the code 9 as "DNSKEY Missing: A DS record existed at a parent, but no supported matching DNSKEY record could be found for the child".

pdns/recursordist/test-syncres_cc4.cc
```diff
@@ -1478,7 +1478,7 @@ BOOST_AUTO_TEST_CASE(test_dnssec_bogus_bad_algo)
...
```
rgacogne, line 1481: It's even weirder here, we have a RRSIG signed with a wrong algo. Or do we consider code 9 to mean "we do not know this DNSKEY"?

pdns/recursordist/test-syncres_cc4.cc
omoerbeek: Yes, It seems validateWithKeySet nees to become a bit smarter.

omoerbeek:
I don't see an easy way yet to distinguish better between the various cases in validate.cc:`validateWithKeySet`, so fix the regression test only for now.

rgacogne (approved):

PR #11171, opened 2022-01-11
dnsdist: Defer the actual allocation of the ring buffer entries

rgacogne:
### Short description

It is a bit cumbersome to have to call `Rings::init()` when we are ready, but it prevents allocating the default number of entries, 10k, which is quite a lot for small setups. Of course the memory was released once the configuration had been parsed, but it might be too late in some cases, and we would end up with a bigger reported memory usage than our actual one since heap shrinkage seldom happens.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdist-rings.cc
+++ b/pdns/dnsdist-rings.cc
@@ -24,6 +24,49 @@
 
 #include "dnsdist-rings.hh"
 
+void Rings::setCapacity(size_t newCapacity, size_t numberOfShards)
+{
+  if (d_initialized) {
+    throw std::runtime_error("Rings::setCapacity() should not be called once the rings have been initialized");
+  }
+  d_capacity = newCapacity;
+  d_numberOfShards = numberOfShards;
+}
+
+void Rings::init()
+{
+  if (d_initialized) {
+    throw std::runtime_error("Rings::init() should only be called once");
+  }
+  d_initialized = true;
+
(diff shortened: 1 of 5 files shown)


pdns/dnsdist-rings.cc
```diff
@@ -24,6 +24,49 @@
...
+void Rings::setCapacity(size_t newCapacity, size_t numberOfShards)
+{
+  if (d_initialized) {
+    throw std::runtime_error("Rings::setCapacity() should not be called once the rings have been initialized");
+  }
+  d_capacity = newCapacity;
+  d_numberOfShards = numberOfShards;
+}
+
+void Rings::init()
```
omoerbeek, line 36: There's is a race if two threads calls `init()` on the same object, but I suppose init() is called when still in single thread mode?
The race is e.g. on `d_shars.resize(...)`

rgacogne: Yes, it should be called only once, in the main thread. I could use an `atomic_flag` if you think it makes things more clear.

omoerbeek: No hard preference, but it might avoid mistakes in the future when this code is used in another context.

omoerbeek (approved):

rgacogne:
Rebased to fix a conflict.

PR #11174, opened 2022-01-11
dnsdist: Add a parameter to PoolAction to keep processing rules

rgacogne:
### Short description

It seems silly not being able to apply other rules after that one.

Could use a regression test.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdist-lua-actions.cc
+++ b/pdns/dnsdist-lua-actions.cc
@@ -273,12 +273,20 @@ void TeeAction::worker()
 class PoolAction : public DNSAction
 {
 public:
-  PoolAction(const std::string& pool) : d_pool(pool) {}
+  PoolAction(const std::string& pool, bool stopProcessing) : d_pool(pool), d_stopProcessing(stopProcessing) {}
+
   DNSAction::Action operator()(DNSQuestion* dq, std::string* ruleresult) const override
   {
-    *ruleresult=d_pool;
-    return Action::Pool;
+    if (d_stopProcessing) {
+      *ruleresult = d_pool;
+      return Action::Pool;
+    }
+    else {
+      dq->poolname = d_pool;
+      return Action::None;
+    }
   }
+
   std::string toString() const override
   {
     return "to pool "+d_pool;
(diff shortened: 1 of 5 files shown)


pdns/dnsdist-lua-actions.cc
```diff
@@ -273,19 +273,28 @@ void TeeAction::worker()
...
 {
 public:
-  PoolAction(const std::string& pool) : d_pool(pool) {}
+  PoolAction(const std::string& pool, bool stopProcessing) : d_pool(pool), d_stopProcessing(stopProcessing) {}
+
   DNSAction::Action operator()(DNSQuestion* dq, std::string* ruleresult) const override
   {
-    *ruleresult=d_pool;
-    return Action::Pool;
+    if (d_stopProcessing) {
```
omoerbeek, line 280: Depending on the flag `dq.poolname` is set via `ruleresult` or directly. I find that a bit inconsistent.
Maybe always set it in the action and leave out the assignment in `applyRulesToQuery`? (Don't forget to fix `QPSPoolAction` as wel if you go that route.

omoerbeek (approved):
A nit, but OK either way.

pdns/dnsdist-lua-actions.cc
rgacogne: That would be much better but I can't leave the assignment out because a custom Lua action can return ``DNSAction.Spoof`, <value to spoof>`. I guess we will have to deprecate that syntax someday because it's not the first time it leads to pain.

omoerbeek: I guess you mean `Action::Pool`, but OK, a comment is in place then. Another question: shouldn't `QPSPoolAction` get the same optional arg?

rgacogne: Ugh, yes, `DNSAction.Pool`! You are right, I only looked at `PoolAction` because that was what I needed at the moment but I'll see if more actions could use the same option.

rgacogne:
Now with regression tests.

PR #11175, opened 2022-01-12
rec: Rename static, global and thread_local vars to follow naming conventions

omoerbeek:
The `thread_local` vars in `houseKeeping()` should be reviewed, as some of them are only relevant for the handler thread, so they do not need to be `thread_local`.

This PR only does mechanical renaming.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code


rgacogne (approved):
Looks good! I wonder if it would make sense to try to group all these global variables in structs. I have been meaning to try to do that in dnsdist for a while.

omoerbeek:
You read my mind :-)

PR #11178, opened 2022-01-13
dnsdist: Build with `-fvisibility=hidden` by default

rgacogne:
### Short description

Not marking our symbols visible by default allows the compiler to do better optimizations, especially with Link-Time Optimizations that several distributions are now enabling by default.
The most obvious optimization comes from dead code removal because the compiler now knows that it cannot be used by an external shared object, but better inlining might also occur.
Even without LTO, it reduces the final size of stripped binaries and might improve loading times a bit.

Note that we already mark the symbols that we want to export with the 'default' visibilty, for example for Lua FFI.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdistdist/configure.ac
+++ b/pdns/dnsdistdist/configure.ac
@@ -18,8 +18,8 @@ AC_DEFINE([DNSDIST], [1],
 LT_PREREQ([2.2.2])
 LT_INIT([disable-static])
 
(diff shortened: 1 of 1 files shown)


omoerbeek (approved):
I Will also do this for rec.

PR #11184, opened 2022-01-13
dnsdist: add support to store mac address in query rings

chbruyand:
### Short description
This PR adds support for storing mac address of the requestor in query rings (only when compiled with `-DDNSDIST_RINGS_WITH_MACADDRESS`).

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


rgacogne (approved):
Looks good, thanks! I agree it doesn't make sense to have this compiled in by default, wasting memory as it should be useful in some particular setups.

PR #11187, opened 2022-01-14
auth: build with -fvisibility=hidden

Habbie:
### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code


Habbie:
This appears to actually break ODBC.

rgacogne:
I wonder if it's related to the "Problems with C++ exceptions" described in https://gcc.gnu.org/wiki/Visibility It might make hiding our symbols by default problematic in the authoritative server because it loads plugins at runtime.

PR #11196, opened 2022-01-17
auth, dnsdist, rec: Fix build with OpenSSL 3.0.0

rgacogne:
### Short description

99% of that PR is based on Habbie's work, to be clear.

### Checklist

I have:
- [x] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/libssl.cc
+++ b/pdns/libssl.cc
@@ -17,9 +17,9 @@
 #endif
 #include <openssl/err.h>
 #include <openssl/ocsp.h>
+#include <openssl/pkcs12.h>
 #include <openssl/rand.h>
 #include <openssl/ssl.h>
-#include <openssl/pkcs12.h>
 #include <fcntl.h>
 
 #ifdef HAVE_LIBSODIUM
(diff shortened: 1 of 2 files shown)


Habbie (approved):

pdns/libssl.cc
```diff
@@ -812,7 +812,7 @@ std::unique_ptr<SSL_CTX, void(*)(SSL_CTX*)> libssl_init_server_context(const TLS
       }
       auto key = std::unique_ptr<EVP_PKEY, void(*)(EVP_PKEY*)>(keyptr, EVP_PKEY_free);
       auto cert = std::unique_ptr<X509, void(*)(X509*)>(certptr, X509_free);
-      auto ca = std::unique_ptr<STACK_OF(X509), void(*)(STACK_OF(X509)*)>(captr, sk_X509_free);
+      auto ca = std::unique_ptr<STACK_OF(X509), void(*)(STACK_OF(X509)*)>(captr, [](STACK_OF(X509)* st){ sk_X509_free(st); });
```
omoerbeek, line 815: OK, though I wonder if we missed an opportunity to use `decltype(&sk_X509_free)` for the second are here....

fredmorcos: I had planned to fix that during the openssl-3 port but it turns out that `sk_x509_free` is a macro.

PR #11225, opened 2022-01-21
rec: Reject non-apex NSEC(3)s that have both the NS and SOA bits set

rgacogne:
### Short description

Ancestor NSEC(3)s have the SOA bit clear (delegation), and the remaining non-apex ones should not have the NS set.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/pdns/validate.cc
+++ b/pdns/validate.cc
@@ -653,25 +653,15 @@ dState getDenial(const cspmap_t &validrrsets, const DNSName& qname, const uint16
           continue;
         }
 
-        const DNSName& owner = v.first.first;
+        const DNSName& hashedOwner = v.first.first;
         const DNSName signer = getSigner(v.second.signatures);
-        if (!owner.isPartOf(signer)) {
-          LOG("Owner "<<owner<<" is not part of the signer "<<signer<<", ignoring"<<endl);
+        if (!hashedOwner.isPartOf(signer)) {
+          LOG("Owner "<<hashedOwner<<" is not part of the signer "<<signer<<", ignoring"<<endl);
           continue;
         }
 
         if (qtype == QType::DS && !qname.isRoot() && signer == qname) {
           LOG("A NSEC3 RR from the child zone cannot deny the existence of a DS"<<endl);
-          continue;
-        }
-
-        /* The NSEC is either a delegation one, from the parent zone, and
-         * must have the NS bit set but not the SOA one, or a regular NSEC
-         * either at apex (signer == owner) or with the SOA or NS bits clear.
-         */
-        const bool notApex = signer.countLabels() < owner.countLabels();
-        if (notApex && nsec3->isSet(QType::NS) && nsec3->isSet(QType::SOA)) {
-          LOG("However, that NSEC3 is not at the apex and has both the NS and the SOA bits set!"<<endl);
           continue;
         }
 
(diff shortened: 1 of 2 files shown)


pdns/validate.cc
```diff
@@ -512,6 +512,16 @@ dState getDenial(const cspmap_t &validrrsets, const DNSName& qname, const uint16
            continue;
         }
 
+        /* The NSEC is either a delegation one, from the parent zone, and
+         * must have the NS bit set but not the SOA one, or a regular NSEC
+         * either at apex (signer == owner) or with the SOA or NS bits clear.
+         */
+        const bool notApex = signer.countLabels() < owner.countLabels();
```
omoerbeek, line 519: Isn't this test equivalent to `signer != owner` as we know `owner.isPartOf(signer)`, as established by the `if (... || !owner.isPartOf(signer)) continue` test above?

Same for NSEC3 case.

pdns/validate.cc
```diff
@@ -692,9 +715,11 @@ dState getDenial(const cspmap_t &validrrsets, const DNSName& qname, const uint16
            * set and that the DS bit is not set in the Type Bit Maps field of the
            * NSEC3 RR.
            */
-          if (referralToUnsigned && qtype == QType::DS && !nsec3->isSet(QType::NS)) {
-            LOG("However, no NS record exists at this level!"<<endl);
-            return dState::NODENIAL;
+          if (referralToUnsigned && qtype == QType::DS) {
```
omoerbeek, line 718: This new code is also equivalent to the old code, right?

pdns/validate.cc
```diff
@@ -661,18 +674,28 @@ dState getDenial(const cspmap_t &validrrsets, const DNSName& qname, const uint16
         nsec3Seen = true;
 
         LOG("\tquery hash: "<<toBase32Hex(h)<<endl);
-        string beginHash=fromBase32Hex(v.first.first.getRawLabels()[0]);
+        string beginHash = fromBase32Hex(hashedOwner.getRawLabels()[0]);
 
         // If the name exists, check if the qtype is denied
         if (beginHash == h) {
 
+          /* The NSEC is either a delegation one, from the parent zone, and
```
omoerbeek, line 682:
```suggestion
          /* The NSEC3 is either a delegation one, from the parent zone, and
```

pdns/validate.cc
```diff
@@ -661,18 +674,28 @@ dState getDenial(const cspmap_t &validrrsets, const DNSName& qname, const uint16
...
+           * must have the NS bit set but not the SOA one, or a regular NSEC
```
omoerbeek, line 683:
```suggestion
           * must have the NS bit set but not the SOA one, or a regular NSEC3
```

pdns/validate.cc
```diff
@@ -661,18 +674,28 @@ dState getDenial(const cspmap_t &validrrsets, const DNSName& qname, const uint16
...
+          }
+
           /* RFC 6840 section 4.1 "Clarifications on Nonexistence Proofs":
              Ancestor delegation NSEC or NSEC3 RRs MUST NOT be used to assume
              nonexistence of any RRs below that zone cut, which include all RRs at
              that (original) owner name other than DS RRs, and all RRs below that
              owner name regardless of type.
           */
-          if (qtype != QType::DS && isNSEC3AncestorDelegation(signer, v.first.first, nsec3)) {
+          if (qtype != QType::DS && isNSEC3AncestorDelegation(signer, qname, nsec3)) {
```
omoerbeek, line 698: In addition to the new check above this was also an already existing bug?
It worries me a b it that this test is similar, but not equal to the NSEC case. If possible it would be good to make them more alike.

omoerbeek (commented):
I's hard for me to follow when to use the `owner` vs `hashedOwner` value in the NSEC3 case, but this is mostly my problem. Feel free to use or ignore my remarks.

pdns/validate.cc
```diff
@@ -546,9 +556,11 @@ dState getDenial(const cspmap_t &validrrsets, const DNSName& qname, const uint16
            * attention.  Bits corresponding to the delegation NS RRset and any
            * RRsets for which the parent zone has authoritative data MUST be set
            */
-          if (referralToUnsigned && qtype == QType::DS && !nsec->isSet(QType::NS)) {
-            LOG("However, no NS record exists at this level!"<<endl);
-            return dState::NODENIAL;
+          if (referralToUnsigned && qtype == QType::DS) {
```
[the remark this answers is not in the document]
rgacogne: It is equivalent, I kept that version because it seemed a bit more clear to me at the moment but that's no longer obvious to me so I can revert it if you prefer to limit the number of changes.

pdns/validate.cc
rgacogne: It's more correct that way but I'm not sure there is an actual bug. If the record comes from the parent zone (DS) the hashed and unhashed names will have more labels than the signer name, by definition. If it comes from the child zone, the hashed and unhashed names will have at least as many labels than the signer name.

rgacogne: As for making that case more similar to the NSEC case, it's complicated as the depth of the zone is lost with NSEC3 until the `closest encloser` has been found.

omoerbeek (commented):
Looked at it again and I'm happy. We should consider this for backporting, but no rush as far as I am concerned.

omoerbeek (approved):

PR #11245, opened 2022-01-28
rec: a CNAME answer on DS query should abort DS retrieval

omoerbeek:
### Short description

For a /broken/ domain that returns a CNAME for a DS query, make sure we abort DS retrieval instead of going into a potential recursion that produces a ServFail.

Maybe the test can be simplified...

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code


rgacogne (approved):
Looks good! Do you know if we have a test checking that we correctly go Bogus for the same case (CNAME answer to a DS query) inside an actually secure zone?

omoerbeek:
> Looks good! Do you know if we have a test checking that we correctly go Bogus for the same case (CNAME answer to a DS query) inside an actually secure zone?

I don;t think so, but it would be good to add. I'll make an issue as I am currently already too many things simultaneously.

PR #11248, opened 2022-01-29
 Declare attribute type 'PdnsDomainId' as Integer with ordering match

### LINKED ISSUE #11247: Declare attribute type 'PdnsDomainId' as Integer
- Program: Authoritative
 - Issue type: Re-factoring

### Short description

The attribute type _PdnsDomainId_ is currently defined with LDAP syntax Numeric String.
And I wonder why this particular syntax was chosen.

Note:
  * with this syntax a filter `(PdnsDomainId=1)` does not match an attribute value '001'.
  * you cannot use MOD_INCREMENT which this syntax
  * attribute values may contain spaces!  (see [RFC 4517  section 3.3.23](https://datatracker.ietf.org/doc/html/rfc4517#section-3.3.23))

### Usecase

I'm currently thinking about automatically generating _PdnsDomainId_ values (within my web2ldap). The commonly used techniques for generating unique Integer(!) values are to use

  * a pool entry with MOD_INCREMENT on a sequence attribute and/or
  * unique constraints (covering pool entry and generated entries) and/or
  * searching the highest existing ID with server-side sorting.

All of the above is nearly impossible to get right with Numeric String syntax.

### Description

My recommendation:

Declare attribute type _PdnsDomainId_ as Integer (see [RFC 4517 section 3.3.16](https://datatracker.ietf.org/doc/html/rfc4517#section-3.3.16)).

Note: Depending on the specific LDAP server used such a schema change requires re-importing (or least re-indexing) the database.

stroeder:
Declare attribute type 'PdnsDomainId' as Integer with ordering match (closes #11247).

### Short description

LDAP schema change to declare attribute type 'PdnsDomainId' as Integer to enable commonly used techniques   for unique value assignment (see details in issue #11247).

### Checklist
I have:

### DIFF
--- a/modules/ldapbackend/pdns-domaininfo.schema
+++ b/modules/ldapbackend/pdns-domaininfo.schema
@@ -20,8 +20,9 @@
 
 attributetype ( 1.3.6.1.4.1.27080.2.1.1 NAME 'PdnsDomainId'
     DESC 'Domain identifier in the LDAP backend - mapped to DomainInfo::id'
-    EQUALITY numericStringMatch
-    SYNTAX 1.3.6.1.4.1.1466.115.121.1.36 SINGLE-VALUE )
+    EQUALITY integerMatch
+    ORDERING integerOrderingMatch
+    SYNTAX 1.3.6.1.4.1.1466.115.121.1.27 SINGLE-VALUE )
 
 attributetype ( 1.3.6.1.4.1.27080.2.1.2 NAME 'PdnsDomainNotifiedSerial'
     DESC 'Last serial number of this zone that slaves have seen - mapped to DomainInfo::notified_serial'


rgacogne:
If we do this, we might want to add a clear warning to the upgrade notes and instructions on how to re-import and/or re-index the database?

Habbie (changes requested):
Thanks, please add a note to `docs/upgrading.rst`

Habbie:
This still needs docs. If you're quick, we can target 4.9 :) but otherwise, "to master" is always valid, and then we can fix it when we release.

Habbie:
ping! (we missed 4.9, but "to master" is still fine for the upgrade notes)

PR #11252, opened 2022-01-31
Rec: thread refactoring

omoerbeek:
### Short description

1. Refactor recursor thread code, `RecThreadInfo` becomes a proper class with methods.
2. Introduce a new thread: `taskThead`. This is a `recursorThread` that can be used for async task execution. Currently one, but in theory there could be more.
3. Use the new `taskThread` to execute the `Zone to Cache`code and the `taskQueue`. The `taskQueue` now has become shared data instead of a thread local one.

The `broadcastAccFucntion` does call the distributor, worker and the new `taskThread`, but skips the handler. This needs to be revisited. One of the questions is; should all the `recursorThreads` be included in stats gathering? ATM, handler is skipped, but it is not clear to me why, as the handler thread also does resolving work, e.g. to prime the root info so constributes to stat counts. I did not change this right now, as there is a complication wit the Lua context, so changing this is better done in a separate PR.

This PR does have the consequence that e.g.`rec_channel ping` now reports one more `pong`, as we have gained an extra `recursorThread`. As the exact response is not documented, I took the liberty to reply with the thread name. Note that atm this function also skips the handler.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/recursordist/rec-main.hh
+++ b/pdns/recursordist/rec-main.hh
@@ -290,8 +290,6 @@ static bool sendResponseOverTCP(const std::unique_ptr<DNSComboWriter>& dc, const
   return hadError;
 }
 
-void* recursorThread();
-
 // For communicating with our threads effectively readonly after
 // startup.
 // First we have the handler thread, t_id == 0 (some other helper
--- a/pdns/recursordist/rec-main.hh
+++ b/pdns/recursordist/rec-main.hh
@@ -384,6 +384,11 @@ public:
     t_id = id;
   }
 
+  std::string getname() const
+  {
+    return name;
+  }
+
   static unsigned int numHandlers()
   {
     return 1;
@@ -466,6 +471,7 @@ private:
   bool taskThread{false};
 
   static thread_local unsigned int t_id;
+  std::string name;
   static std::vector<RecThreadInfo> s_threadInfos;
   static bool s_weDistributeQueries; // if true, 1 or more threads listen on the incoming query sockets and distribute them to workers
   static unsigned int s_numDistributorThreads;
--- a/pdns/recursordist/rec-main.hh
+++ b/pdns/recursordist/rec-main.hh
@@ -139,7 +139,7 @@ struct DNSComboWriter
   std::map<std::string, RecursorLua4::MetaValue> d_meta;
 };
 
-extern thread_local FDMultiplexer* t_fdm;
+extern thread_local unique_ptr<FDMultiplexer> t_fdm;
 extern uint16_t g_minUdpSourcePort;
 extern uint16_t g_maxUdpSourcePort;
 
--- a/pdns/recursordist/rec-main.hh
+++ b/pdns/recursordist/rec-main.hh
@@ -35,6 +35,7 @@
 #include "syncres.hh"
 #include "rec-snmp.hh"
 #include "rec_channel.hh"
+#include "threadname.hh"
 
 #ifdef NOD_ENABLED
 #include "nod.hh"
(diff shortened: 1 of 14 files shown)


rgacogne:
> The broadcastAccFucntion does call the distributor, worker and the new taskThread, but skips the handler. This needs to be revisited.

Doesn't the current code just execute the requested function directly in the handler thread?

omoerbeek:
> > The broadcastAccFucntion does call the distributor, worker and the new taskThread, but skips the handler. This needs to be revisited.
> 
> Doesn't the current code just execute the requested function directly in the handler thread?

Nope that is true for `broadcastFunction`, but not for `broadcastAccFunction`

rgacogne:
> One of the questions is; should all the recursorThreads be included in stats gathering? ATM, handler is skipped, but it is not clear to me why, as the handler thread also does resolving work, e.g. to prime the root info so constributes to stat counts. I did not change this right now, as there is a complication wit the Lua context, so changing this is better done in a separate PR.

Right, this was not a problem in the past because the handler was not really doing any meaningful resolving work (the cache was not shared), but I see why this might not make sense nowadays.

pdns/recursordist/rec-main.hh
```diff
@@ -313,26 +307,175 @@ struct RecThreadInfo
...
+  {
+    return t_id;
+  }
+
+  static void setThreadId(unsigned int id)
+  {
+    t_id = id;
+  }
+
+  std::string getname() const
```
rgacogne, line 387:
```suggestion
  std::string getName() const
```

pdns/recursordist/rec-main.hh
```diff
@@ -313,26 +307,175 @@ struct RecThreadInfo
...
+  bool handler{false};
+  // accept incoming queries (and distributes them to the workers if pdns-distributes-queries is set)
+  bool listener{false};
+  // process queries
+  bool worker{false};
+  // run async tasks: from TastQueue and ZoneToCache
+  bool taskThread{false};
+
+  static thread_local unsigned int t_id;
+  std::string name;
```
rgacogne, line 474: Perhaps move that field up, with the other non-static ones?

omoerbeek:
> > One of the questions is; should all the recursorThreads be included in stats gathering? ATM, handler is skipped, but it is not clear to me why, as the handler thread also does resolving work, e.g. to prime the root info so constributes to stat counts. I did not change this right now, as there is a complication wit the Lua context, so changing this is better done in a separate PR.
> 
> Right, this was not a problem in the past because the handler was not really doing any meaningful resolving work (the cache was not shared), but I see why this might not make sense nowadays.

I played with the idea to add an argument to `broadcastAccFunction` to signal if it should include the handler thread. That way, we can switch the callers after inspection one by one.

rgacogne (approved):
Looks great, one cosmetic comment or two but I love it.

kpfleming:
The new `taskThread` could be used to distribute the NOTIFY-triggered cache wipes too, I suspect.

pdns/recursordist/rec-taskqueue.cc
```diff
@@ -70,29 +71,29 @@ static void resolve(const struct timeval& now, bool logErrors, const pdns::Resol
 
 void runTaskOnce(bool logErrors)
 {
-  t_taskQueue.runOnce(logErrors);
+  s_taskQueue.lock()->runOnce(logErrors);
```
omoerbeek, line 74: Reviewing my own code, I think this should be done differently. The lock should not be help while running the task.
It should be seomthign like this:
```
auto task = s_taskQueue.lock()->pop();
if (!task.empty()) {
  task.run();
}
```
plus some code if ther is no task available.

omoerbeek:
Rebased to fix conflict

PR #11258, opened 2022-01-31
dnsdist: Do not execute 'newThread' in client or config check modes

rgacogne:
### Short description

New Lua threads will no longer get started in 'client' and 'config check' modes, where they make little sense.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdist-lua.cc
+++ b/pdns/dnsdist-lua.cc
@@ -2794,7 +2794,10 @@ static void setupLuaConfig(LuaContext& luaCtx, bool client, bool configCheck)
   });
 #endif /* HAVE_LIBSSL */
 
-  luaCtx.writeFunction("newThread", [](const std::string& code) {
+  luaCtx.writeFunction("newThread", [client, configCheck](const std::string& code) {
+    if (client || configCheck) {
+      return;
+    }
     std::thread newThread(LuaThread, code);
     newThread.detach();
   });
--- a/pdns/dnsdistdist/docs/reference/config.rst
+++ b/pdns/dnsdistdist/docs/reference/config.rst
@@ -1608,6 +1608,7 @@ Other functions
 
   Spawns a separate thread running the supplied code.
   Code is supplied as a string, not as a function object.
+  Note that this function does nothing in 'client' or 'config-check' modes.
 
 .. function:: submitToMainThread(cmd, dict)
 


rgacogne:
Rebased.

rgacogne:
CI failure unrelated (ALIAS).

Habbie (approved):
Approved - but I wonder if newthread should even do anything during a config check.

regression-tests.dnsdist/test_Lua.py
```diff
@@ -20,17 +20,19 @@ class TestLuaThread(DNSDistTest):
         counter = tonumber(data.i)
     end
 
-    newThread([==[
-      local socket = require'socket'
-      local i=1
-      while true
-      do
+    if inConfigCheck() == false then
```
omoerbeek, line 23: I'm not following (but this is likely lack of knowledge) how the thread is started. If the test fails, is this code re-executed at some later moment when the test is expected to succeed?

Also, what's the role of threadmessage?

Habbie: The testing framework (in dnsdisttests.py) runs `dnsdist --check-config` against the given config, before actually starting dnsdist. During check-config, we do not want this extra thread to generate messages, or run at all. `threadmessage` is called in the main Lua thread when this separate thread calls `submitToMainThread`.

omoerbeek: I looked up `inConfigCheck` and now its more clear. The question about `threadmessage` remains.

omoerbeek: Thanks. Searching for either `submitToMainThread` or `threadmessage` in the docs did not produce anything.

Habbie: This interface is (was?) likely to change before release, so we did not document it at the time. Now that it's in 1.7.0, I'll fix that.

rgacogne:
> Approved - but I wonder if newthread should even do anything during a config check.

That's a fair question. On one hand I do not think it should and now would be a good time to change that behaviour since the feature is still fresh, but on the other hand that means arbitrarily increasing the parts of the configuration that are not checked during a config check.

Habbie:
> arbitrarily increasing the parts of the configuration that are not checked during a config check.

Right, but I don't think we're really checking the code in newThread right now either.

rgacogne:
> Right, but I don't think we're really checking the code in newThread right now either.

It is executed, right? :)

Habbie:
> It is executed, right? :)

if it's fast enough! but failures do not bubble up to main

rgacogne:
Ah, right! I'll update that PR to disable `newThread` in config check and client modes then.

Habbie (approved):

PR #11269, opened 2022-02-01
rec: move two maps from thread_local to shared

omoerbeek:
### Short description

These maps are only used or modified when we're going out on the
net, so the performance impact of the locking should be relatively
low, while threads could benefit greatly from information
learned by other threads.

Also, pruning of these data structure is cheap, so holding the lock
while pruning should be a short period of time.

When dumping, we get a copy to not have to deal with the issue that I/O might stall while holding a lock.

Also includes fixed to metrics for these two tables. It's sad that one of the metrics is called `failed-hosts`, while in other places the table is called `failedservers`. Add to that that the existing implementation actually returns the size of the throttle table....

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code


pdns/recursordist/RECURSOR-MIB.txt
```diff
@@ -1050,6 +1053,15 @@ zoneDisallowedNotify OBJECT-TYPE
...
         "Number of NOTIFY operations not allowed by allow-notify-for"
     ::= { stats 125 }
+
+nonResolvingNameserverEntries OBJECT-TYPE
+    SYNTAX CounterBasedGauge64
+    MAX-ACCESS read-only
+    STATUS current
+    DESCRIPTION
+        "Number of entries in the non-resolving NS name cache"
+    ::= { stats 126 }
```
rgacogne, line 1063: The new entry (and the two previous ones, by the look of it) should be added to the "Objects conformance group" below.

pdns/recursordist/docs/metrics.rst
```diff
@@ -546,6 +546,10 @@ noerror-answers
 ^^^^^^^^^^^^^^^
 counts the number of times it answered NOERROR   since starting
 
+non-resolving-nameserver-entries
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
rgacogne, line 550:
```suggestion
non-resolving-nameserver-entries
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

rgacogne (approved):
Looks good, one nit in the SNMP MIB.

Issue #6138, opened 2018-01-03
rec: Query to a.root-servers.net validates as Bogus


rgacogne:
- Program: Recursor
 - Issue type: Bug report

### Short description
When we receive a query for `a.root-servers.net|A` with an empty cache, we start by determining the zone cuts and their status:
- `.`: we have a trusted anchor, so we know there is a cut and the zone is Secure ;
-  `net.`: we look for a `DS` or its secure denial by interrogating one of the `NS` for `.`, retrieve the `DS` and validate it ;
- `root-servers.net.`: we send a query for `root-servers.net|DS` to the most specific `NS` we have, which is still the `NS` for `.`, for example `b.root-servers.net`. Some of these return a referral to the `net.` nameservers (a c e h j k l) which is fine, but some of them return a `NODATA` (b d f g i m), because they are authoritative for the `root-servers.net` zone:

```
$ dig root-servers.net DS @m.root-servers.net +dnssec

; <<>> DiG 9.11.2 <<>> root-servers.net DS @m.root-servers.net +dnssec
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 24923
;; flags: qr aa rd; QUERY: 1, ANSWER: 0, AUTHORITY: 1, ADDITIONAL: 1
;; WARNING: recursion requested but not available

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags: do; udp: 4096
;; QUESTION SECTION:
;root-servers.net.              IN      DS

;; AUTHORITY SECTION:
root-servers.net.       3600000 IN      SOA     a.root-servers.net. nstld.verisign-grs.com. 2017111600 14400 7200 1209600 3600000

;; Query time: 23 msec
;; SERVER: 2001:dc3::35#53(2001:dc3::35)
;; WHEN: Wed Jan 03 15:39:20 CET 2018
;; MSG SIZE  rcvd: 105
```
The zone is not signed so this answer does not have any `RRSIG`, but since the last cut we know of (`net.`) was Secure, we consider this answer as Bogus.
This seldom happens because as soon as we learn which servers are authoritative for `net.` we query them directly for `root-servers.net|DS`. That happens very quickly, especially with our security polling mechanism. This still happens from time to time, for example with a `dnsdist` in front whose default health check query is `a.root-servers.net|A`.

### Environment
 - Software version: PowerDNS Recursor 4.1

### Expected behaviour

Answer should be Insecure.

[...]

PR #11283, opened 2022-02-02
rec: Allow disabling of processing the root hints

omoerbeek:
Disabling can be useful in special setups, e.g. if the root servers cannot be reached and you're forwarding all queries to another recursor. This PR then allows you to not have  to setup a fake root by setting `hint-file=no`.

This also make sure we use the right dnssec mode for processing hints
and changes a few log levels to Debug to be less verbose.

I do worry that people will disable root-hints processing when they should not and then come complaining that their recursor is broken... hence the draft status

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/recursordist/docs/settings.rst
+++ b/pdns/recursordist/docs/settings.rst
@@ -879,8 +879,11 @@ If set, EDNS options in incoming queries are extracted and passed to the :func:`
 ``hint-file``
 -------------
 -  Path
+-  Default: empty
 
-If set, the root-hints are read from this file. If unset, default root hints are used.
+If set, the root-hints are read from this file. If empty, the default built-in root hints are used.
+In some special cases, processing the root hints is not needed, for example when forwarding all queries to another recursor.
+For these special cases, it is possible to disable the processing of root hints by setting the value to ``no``.
 
 .. _setting-ignore-unknown-settings:
 
--- a/pdns/recursordist/docs/settings.rst
+++ b/pdns/recursordist/docs/settings.rst
@@ -881,7 +881,12 @@ If set, EDNS options in incoming queries are extracted and passed to the :func:`
 -  Path
 -  Default: empty
 
+.. versionchanged:: 4.7.0
+
+  Introduced the value ``no`` to disable root-hints processing.
+
 If set, the root-hints are read from this file. If empty, the default built-in root hints are used.
+
 In some special cases, processing the root hints is not needed, for example when forwarding all queries to another recursor.
 For these special cases, it is possible to disable the processing of root hints by setting the value to ``no``.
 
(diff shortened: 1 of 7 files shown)


pdns/reczones.cc
```diff
@@ -168,16 +173,14 @@ bool primeHints(time_t ignored)
...
   struct timeval now;
   gettimeofday(&now, 0);
   SyncRes sr(now);
 
-  if (dnssecmode) {
-    sr.setDoDNSSEC(true);
-    sr.setDNSSECValidationRequested(true);
-  }
+  sr.setDoDNSSEC(mode != DNSSECMode::Off);
+  sr.setDNSSECValidationRequested(mode != DNSSECMode::Off && mode != DNSSECMode::ProcessNoValidate);
```
rgacogne, line 183: Am I correctly understanding that we used to do DNSSEC validation in `process` mode and will no longer do so? If so that seems to make sense, and at a quick glance I don't think it will cause any issue.

pdns/recursordist/docs/settings.rst
```diff
@@ -879,8 +879,11 @@ If set, EDNS options in incoming queries are extracted and passed to the :func:`
 ``hint-file``
 -------------
 -  Path
+-  Default: empty
 
-If set, the root-hints are read from this file. If unset, default root hints are used.
+If set, the root-hints are read from this file. If empty, the default built-in root hints are used.
+In some special cases, processing the root hints is not needed, for example when forwarding all queries to another recursor.
+For these special cases, it is possible to disable the processing of root hints by setting the value to ``no``.
```
rgacogne, line 886: I wonder if it would make sense to check whether we have a forward-recurse target for the root when that option is enabled, and to issue a loud warning (or even refuse to start) otherwise?

rgacogne (approved):
Code looks good!

pdns/reczones.cc
omoerbeek: yes, I noticed that when working on this. In `process` mode we should default to on-demand validation.

pdns/recursordist/docs/settings.rst
omoerbeek: Wondering as wel, don;t know if forward-recurse of root is the only one, there might be more (strange) configs I did not think of yet.

omoerbeek:
TODO: note in upgrade guide

pdns/reczones.cc
omoerbeek: Looking at this again, it think it is more subtle: we were doing immediate validation in `ProcessNoValidate` mode. Which is wrong (but not harmful afaiks) . For `Process` it's debatable if we should call `setDNSSECValidationRequested`.

But the issue is also a bit mood, as the is the function that refreshes the domain information of the root servers names (`root-servers.net`) and not the update of the root NS records themselves and `root-servers.net` is a `Insecure` domain anyway.

This function was needed because of #6138. I now wonder if it still is relevant with the new approach to determining zone-cuts.

PR #11143, opened 2022-01-05
auth: Reintroduce GSS-TSIG processing

omoerbeek:
### Short description

Based on the code that was removed in #9385 with the following changes and additions:

- Even when compiled in, the accepting is only done if the new setting `enable-gss-tsig` is set to `yes`.
- We only try to create a `GssContext` if we succeeded in finding an accepting principal to avoid creating security contexts that are useless.
- The three maps used by `GssContext` are now thread safe.
- The three maps have an expiry mechanism
- If the update query fails, e.g. because we could not find a matching`TSIG-ALLOW-DNSUPDATE` metadata, we immediately clean up the associated security context after the reply has been sent.
- There are regression tests.

A slightly related change is that if the reading the `lua-dnsupdate-policy-script` fails, no Lua policy is activated.

Also contains improvements to the auth startup and teardown code in the `auth-py` regression test, copied from the rec regression tests.

### Checklist

I have:
- [X] read the [CONTRIBUTING.md](https://github.com/PowerDNS/pdns/blob/master/CONTRIBUTING.md) document
- [X] compiled this code
- [X] tested this code
- [X] included documentation (including possible behaviour changes)
- [X] documented the code
- [X] added or modified regression test(s)
- [ ] added or modified unit test(s)

PR #11285, opened 2022-02-02
auth dnsupdate: add option which aborts transaction on error

### LINKED ISSUE #8422: dns lua update policy sends wrong response to client when update is denied

jdreichmann:
### Short description
This PR adds a config option `lua-dnsupdate-policy-script-strict-mode`, which, when set to `yes`, makes PowerDNS abort a DNSUPDATE transaction, which in turn can resolve #8422 when turned on.

The PR is based on #8422 and aims to be a more configurable option, as changing the existing behaviour is not desirable

### Checklist

I have:
- [x] compiled this code
- [ ] tested this code
- [x] documented the code

### DIFF
--- a/docs/dnsupdate.rst
+++ b/docs/dnsupdate.rst
@@ -69,6 +69,15 @@ authorization methods, and you are expected to take care of everything
 yourself. See :ref:`dnsupdate-update-policy` for details and
 examples.
 
+.. _dnsupdate-lua-dnsupdate-policy-script-strict-mode:
+
+``lua-dnsupdate-policy-script-strict-mode``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+When turning this setting on, PowerDNS will handle DNS updates as atomic,
+rolling back an update when a single record in an update causes the
+configured lua script to return false, and returning `REFUSED` to the client.
+
 The semantics are that first a dynamic update has to be allowed either
 by the global :ref:`setting-allow-dnsupdate-from` setting, or by a per-zone
 ``ALLOW-DNSUPDATE-FROM`` metadata setting.
--- a/pdns/common_startup.cc
+++ b/pdns/common_startup.cc
@@ -222,6 +222,7 @@ void declareArguments()
 
   ::arg().set("lua-prequery-script", "Lua script with prequery handler (DO NOT USE)")="";
   ::arg().set("lua-dnsupdate-policy-script", "Lua script with DNS update policy handler")="";
+  ::arg().setSwitch("lua-dnsupdate-policy-script-strict-mode", "Enables strict adherence to RFC2136 Section 3.7 by failing the entire transaction when a single record does not pass the policy script")="no";
 
   ::arg().setSwitch("traceback-handler","Enable the traceback handler (Linux only)")="yes";
   ::arg().setSwitch("direct-dnskey","Fetch DNSKEY, CDS and CDNSKEY RRs from backend during DNSKEY or CDS/CDNSKEY synthesis")="no";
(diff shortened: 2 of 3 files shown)


Habbie:
Thanks! I notice that we do not test this feature at all currently, with or without this flag. Are you up for adding tests (with and without this setting)? `regression-tests.auth-py` is probably the easiest spot as you can easily spawn instances with different settings.

pdns/common_startup.cc
```diff
@@ -222,6 +222,7 @@ void declareArguments()
 
   ::arg().set("lua-prequery-script", "Lua script with prequery handler (DO NOT USE)")="";
   ::arg().set("lua-dnsupdate-policy-script", "Lua script with DNS update policy handler")="";
+  ::arg().setSwitch("lua-dnsupdate-policy-script-strict-mode", "Enables strict adherence to RFC2136 Section 3.7 by failing the entire transaction when a single record does not pass the policy script")="no";
```
Habbie, line 225: the RFC reference is confusing, I think the text would be better without it

docs/dnsupdate.rst
```diff
@@ -69,6 +69,15 @@ authorization methods, and you are expected to take care of everything
...
 
+.. _dnsupdate-lua-dnsupdate-policy-script-strict-mode:
+
+``lua-dnsupdate-policy-script-strict-mode``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+When turning this setting on, PowerDNS will handle DNS updates as atomic,
+rolling back an update when a single record in an update causes the
+configured lua script to return false, and returning `REFUSED` to the client.
+
```
Habbie, line 80: I think it might make more sense to move this block down, to just below `.. _dnsupdate-metadata:`

pdns/rfc2136handler.cc
```diff
@@ -869,9 +869,15 @@ int PacketHandler::processUpdate(DNSPacket& p) {
       if (rr->d_place == DNSResourceRecord::AUTHORITY) {
         /* see if it's permitted by policy */
         if (this->d_update_policy_lua != nullptr) {
+          const bool rfc2136_strict_mode = ::arg().mustDo("lua-dnsupdate-policy-script-strict-mode");
```
Habbie, line 872: I think this could even be `static`, to avoid the call every time.

Habbie:
> Are you up for adding tests (with and without this setting)? regression-tests.auth-py is probably the easiest spot as you can easily spawn instances with different settings.

@omoerbeek just pointed out to me that #11143 has a test that uses a policy script - perhaps you can steal from there :-)

docs/dnsupdate.rst
jdreichmann: My rst-knowledge is limited, but wouldn't the location then imply that this setting is related to per-zone metadata settings?
(from what i figured, it would be show up when browsing to https://docs.powerdns.com/authoritative/dnsupdate.html#dnsupdate-metadata as the first thing in the viewport)

pdns/rfc2136handler.cc
jdreichmann: https://github.com/PowerDNS/pdns/pull/11285/commits/2b6b2f0358617a7f089f3f50fa4afb2e435a6099 changed here to a `static bool`

jdreichmann:
I am currently working on DNSUpdate tests, though covering all the possible (existing and new) settings might take a while

PR #11288, opened 2022-02-04
Rec: document changes to policy.DROP better and warn on using the now unsupported way

### LINKED ISSUE #11287: Exception thrown by a callback function: Trying to cast a lua variable from "string" to "NSt15__exception_ptr13exception_ptrE" (pdns.DROP causing havoc)
> This bugreport is created to make this problem searchable. 

 - Program: Recursor
 - Issue type: bugreport

Upgrading recursor from 4.3.xx to 4.6.xx breaks lua script in a most unintuitive way. It is really hard to debug, really hard to guess at the problem and the documentation is lying. 

### What happens after upgrade
If you have a lua script you may get infinite amounts of:

```
STL Error( ... Exception thrown by a callback function: Trying to cast a lua variable from "string" to "NSt15__exception_ptr13exception_ptrE" ...)
```
With no visible reason or explanation, nor any line numbers or function indicators. "It is in the lua, any of them, somewhere."

### Cause
`dq.rcode = pdns.DROP` _seems to be_ depreciated, then transformed into a bug generator. This assignment actually generated the error message, and may or may not cause other problems.

The documentation suggests that _this is a proper solution_. 

By browsing issues and searching for wildly random terms it turns out that issue #10080 and pull #10111 actually took care of something, and suggested a documentation change which may or may not actually happen (but definitely have not changed the examples section in the lua titles of the recursor documentation), and in due time intentionally or not broke `pdns.DROP` with little visibility.

### Solution
Code like
```lua
function preresolve(dq)
  if blocklist:check(dq.qname) then
    dq.rcode=pdns.DROP
    return true
  end
end
```
shall be replaced by
```lua
function preresolve(dq)
  if blocklist:check(dq.qname) then
    dq.appliedPolicy.policyKind = pdns.policykinds.Drop
    return false
  end
end
```
and the error will go away.

### Suggestions
- fix the doc
- fix the error message (I guess `pdns.DROP` is `nil`now, or it's maybe a non-existing element in some collection?)
- warn the people, it hurts.
- feel free to close this issue, if you believe there's nothing more left to do.

omoerbeek:
### Short description

Address documentation issues with the old `pdns.DROP` method and warn if `rcode` is set to it in Lua hooks.

Also fix a few other cases of inconsistent style (and errors) in the example script.

Fixes #11287

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [X] documented the code

### DIFF
--- a/pdns/lua-base4.cc
+++ b/pdns/lua-base4.cc
@@ -244,7 +244,8 @@ void BaseLua4::prepareContext() {
                                        {"YXRRSET",  RCode::YXRRSet  },
                                        {"NXRRSET",  RCode::NXRRSet  },
                                        {"NOTAUTH",  RCode::NotAuth  },
-                                       {"NOTZONE",  RCode::NotZone  }};
+                                       {"NOTZONE",  RCode::NotZone  },
+                                       {"DROP",    -2               }}; // To give backport-incompatibilityy warning
   for(const auto& rcode : rcodes)
     d_pd.push_back({rcode.first, rcode.second});
 
--- a/pdns/lua-base4.cc
+++ b/pdns/lua-base4.cc
@@ -245,7 +245,7 @@ void BaseLua4::prepareContext() {
                                        {"NXRRSET",  RCode::NXRRSet  },
                                        {"NOTAUTH",  RCode::NotAuth  },
                                        {"NOTZONE",  RCode::NotZone  },
-                                       {"DROP",    -2               }}; // To give backport-incompatibilityy warning
+                                       {"DROP",    -2               }}; // To give backport-incompatibility warning
   for(const auto& rcode : rcodes)
     d_pd.push_back({rcode.first, rcode.second});
 
--- a/pdns/lua-recursor4.cc
+++ b/pdns/lua-recursor4.cc
@@ -500,7 +500,7 @@ void RecursorLua4::getFeatures(Features& features)
 static void warnDrop(const RecursorLua4::DNSQuestion& dq)
 {
   if (dq.rcode == -2) {
-    g_log << Logger::Error << "Returing -2 (pdns.DROP) is not supported anymore, see https://docs.powerdns.com/recursor/lua-scripting/hooks.html#hooksemantics" << endl;
+    g_log << Logger::Error << "Returning -2 (pdns.DROP) is not supported anymore, see https://docs.powerdns.com/recursor/lua-scripting/hooks.html#hooksemantics" << endl;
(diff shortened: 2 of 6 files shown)


pdns/lua-base4.cc
```diff
@@ -244,7 +244,8 @@ void BaseLua4::prepareContext() {
                                        {"YXRRSET",  RCode::YXRRSet  },
                                        {"NXRRSET",  RCode::NXRRSet  },
                                        {"NOTAUTH",  RCode::NotAuth  },
-                                       {"NOTZONE",  RCode::NotZone  }};
+                                       {"NOTZONE",  RCode::NotZone  },
+                                       {"DROP",    -2               }}; // To give backport-incompatibilityy warning
```
rgacogne, line 248:
```suggestion
                                       {"DROP",    -2               }}; // To give backport-incompatibility warning
```

pdns/lua-recursor4.cc
```diff
@@ -497,6 +497,15 @@ void RecursorLua4::getFeatures(Features& features)
   features.emplace_back("PR8001_devicename", true);
 }
 
+static void warnDrop(const RecursorLua4::DNSQuestion& dq)
+{
+  if (dq.rcode == -2) {
+    g_log << Logger::Error << "Returing -2 (pdns.DROP) is not supported anymore, see https://docs.powerdns.com/recursor/lua-scripting/hooks.html#hooksemantics" << endl;
```
rgacogne, line 503:
```suggestion
    g_log << Logger::Error << "Returning -2 (pdns.DROP) is not supported anymore, see https://docs.powerdns.com/recursor/lua-scripting/hooks.html#hooksemantics" << endl;
```

pdns/recursordist/contrib/powerdns-example-script.lua
```diff
@@ -3,140 +3,141 @@ pdnslog("pdns-recursor Lua script starting!", pdns.loglevels.Warning)
...
+  end
+
+  local a = dq:getEDNSOption(3)
+  if a
+  then
+    pdnslog("There is an EDNS option 3 present: "..a)
+  end
+
+  loc = newCA("127.0.0.1")
+  if dq.remoteaddr:equal(loc) then
```
rgacogne, line 46: If we really want to be consistent with the `then` being on a new line, which I am really not a fan of:
```suggestion
  if dq.remoteaddr:equal(loc)
  then
```

rgacogne (approved):

PR #11294, opened 2022-02-04
Rec: resolve AAAA for NS in an async task

omoerbeek:
### Short description

If we have to resolve a nameserver name, submit a AAAA query for the same name asynchronously if it's not in the negcache.

<strike>To make this work, I also had to add the negcaching of SOA-less NODATA responses. Otherwise this task would be inserted each time that nameserver is used as the result is not negcached at the moment and so would trigger a new task. The  protection below also helps to avoid this case. 

Needs more testing, especially for failing to resolve cases.
</strike>

Thinking about this a bit more, a general rate limit for resolve tasks for a specific qname/qtype combo is needed. That way, we can also seperate the the negcaching of SOA-less NODATA responses issue into a independent PR.

So I did that thing mentioned in the last paragraph. The negcache changes are no longer there.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [X] documented the code

### DIFF
--- a/pdns/recursordist/rec-taskqueue.cc
+++ b/pdns/recursordist/rec-taskqueue.cc
@@ -71,6 +71,11 @@ public:
     return inserted;
   }
 
+  void clear()
+  {
+    d_set.clear();
+  }
+
 private:
   struct Entry
   {
@@ -205,6 +210,18 @@ uint64_t getTaskExpired()
 uint64_t getTaskSize()
 {
   return s_taskQueue.lock()->queue.size();
+}
+
+void taskQueueClear()
+{
+  auto lock = s_taskQueue.lock();
+  lock->queue.clear();
+  lock->rateLimitSet.clear();
+}
+
+pdns::ResolveTask taskQueuePop()
+{
+  return s_taskQueue.lock()->queue.pop();
 }
 
 uint64_t getAlmostExpiredTasksPushed()
--- a/pdns/recursordist/rec-taskqueue.cc
+++ b/pdns/recursordist/rec-taskqueue.cc
@@ -167,22 +167,18 @@ void runTaskOnce(bool logErrors)
 void pushAlmostExpiredTask(const DNSName& qname, uint16_t qtype, time_t deadline)
 {
   pdns::ResolveTask task{qname, qtype, deadline, true, resolve};
-  auto lock = s_taskQueue.lock();
-  bool running = !lock->rateLimitSet.insert(time(nullptr), task);
-  if (!running) {
-    ++s_almost_expired_tasks.pushed;
-    lock->queue.push(std::move(task));
-  }
+  s_taskQueue.lock()->queue.push(std::move(task));
+  ++s_almost_expired_tasks.pushed;
 }
 
-void pushResolveTask(const DNSName& qname, uint16_t qtype, time_t deadline)
+void pushResolveTask(const DNSName& qname, uint16_t qtype, time_t now, time_t deadline)
 {
   pdns::ResolveTask task{qname, qtype, deadline, false, resolve};
   auto lock = s_taskQueue.lock();
-  bool running = !lock->rateLimitSet.insert(time(nullptr), task);
-  if (!running) {
-    ++s_resolve_tasks.pushed;
+  bool inserted = lock->rateLimitSet.insert(now, task);
+  if (inserted) {
     lock->queue.push(std::move(task));
+    ++s_resolve_tasks.pushed;
   }
 }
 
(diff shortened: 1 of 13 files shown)


pdns/recursordist/rec-taskqueue.cc
```diff
@@ -22,50 +22,129 @@
...
-#include "logger.hh"
+#include "logging.hh"
 #include "stat_t.hh"
 #include "syncres.hh"
 
-static LockGuarded<pdns::TaskQueue> s_taskQueue;
-static pdns::stat_t s_almost_expired_tasks_pushed;
-static pdns::stat_t s_almost_expired_tasks_run;
-static pdns::stat_t s_almost_expired_tasks_exceptions;
+// For rate lmiting, we maintain a set of tasks recently submitted.
```
rgacogne, line 29:
```suggestion
// For rate limiting, we maintain a set of tasks recently submitted.
```

pdns/recursordist/rec-taskqueue.cc
```diff
@@ -22,50 +22,129 @@
...
+  {
+  }
+  bool insert(time_t now, const pdns::ResolveTask& task)
+  {
+    time_t ttd = now + d_expiry_seconds;
+    bool inserted = d_set.emplace(task, ttd).second;
+    if (!inserted) {
+      // Instead of a periodic clean, we always do it on a hit
+      // the operation should be cheap as we just walk the ordered time_t index
+      // There is a slim chance if we never hit a rate limiting case we'll never clean... oh well
```
rgacogne, line 44: I'm a bit worried about the odds of that happening in some setups, because without collision the set will grow endlessly..

rgacogne (approved):
Looks good! A few unit and perhaps regression tests would be good to have.

omoerbeek:
Thanks, I'll address tests and the "never hitting purge" case before I'll merge.

mnordhoff:
Did you mean to push a commit regarding #11341 here?

omoerbeek:
> Did you mean to push a commit regarding #11341 here?

Thanks for spotting that, wrong base branch for that commit. Should be fixed now,

pdns/recursordist/rec-taskqueue.cc
```diff
@@ -22,50 +22,139 @@
...
+{
+public:
+  TimedSet(time_t t) :
+    d_expiry_seconds(t)
+  {
+  }
+
+  uint64_t purge(time_t now)
+  {
+    // This purge is relatively sleep, as we're walking an ordered index
```
Habbie, line 40: sleep?

Habbie: fast, I guess, as indeed we `break` the moment we hit things that are still new enough

pdns/recursordist/rec-taskqueue.cc
```diff
@@ -22,50 +22,139 @@
...
+  {
+    // We periodically purge
+    if (++d_count % 1024 == 0) {
+      purge(now);
+    }
+    time_t ttd = now + d_expiry_seconds;
+    bool inserted = d_set.emplace(task, ttd).second;
+    if (!inserted) {
+      uint64_t erased = purge(now);
+      // Try again if the purge deleted at least one entry
```
Habbie, line 66: Why does a deletion allow the insert to happen?

pdns/recursordist/test-syncres_cc.cc
```diff
@@ -558,3 +558,7 @@ void pushAlmostExpiredTask(const DNSName& qname, uint16_t qtype, time_t deadline
 {
   g_test_tasks.push({qname, qtype, deadline, true, nullptr});
 }
+
+void pushResolveTask(const DNSName& qname, uint16_t qtype, time_t now, time_t deadline)
```
Habbie, line 562:
```suggestion
static void pushResolveTask(const DNSName& qname, uint16_t qtype, time_t now, time_t deadline)
```

Habbie (approved):
Just a few nits

pdns/recursordist/rec-taskqueue.cc
omoerbeek: I meant to type cheap, but my fingers are autonomous beings

pdns/recursordist/rec-taskqueue.cc
omoerbeek: `d_set` can contain an equal task that is expired but not yet purged as we do only purge once every 1000 calls. So if the insert fails, we do an extra purge sweep to make sure we an expired task does not prevent the insert.

pdns/recursordist/test-syncres_cc.cc
omoerbeek: That wont fly as is declared non-static in `rec-taskqueue.hh` (and it needs to be non-static).

PR #11300, opened 2022-02-08
rec: Use the Lua context stored in SyncRes when calling hooks.

### LINKED ISSUE #11289: rec_control reload-lua-script causes STL errors
- Program: Recursor
 - Issue type: Bug report

### Short description
When reloading lua custom code (using rec_control reload-lua-script), it seems (some / all) in-flight queries generate an STL error. I use `dq` objects in Lua, including modifying them.

### Environment
 - Operating system:  Centos 7
 - Software version: 4.4.3.0

### Steps to reproduce
1. Configure pdns-recursor to use Lua hooks, use `dq` and `dq.data` in Lua
2. Have a decent amount of qps :)
2. `rec_control reload-lua-script`

### Expected behaviour
Not crash / Not generate a ServFail or other error, not generate STL errors in the logs

### Actual behaviour
Queries seem to get a ServFail (or other error), and log entries as in https://gist.github.com/franklouwers/28074f09e45dfdd9c2c29a15ea6f7bde appear. (IPs stripped in log)

omoerbeek:
After the Lua context is recreated by a reload script, in-flight queries should still use the old one and not the new.

Fixes #11289

The Lua context used by the followCNAMERecords(), getFakeAAAARecords() and
getFakePTRRecords() functions should be validated. I have marked the spots
with XXX in this commit

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/recursordist/rec-main.hh
+++ b/pdns/recursordist/rec-main.hh
@@ -48,8 +48,8 @@
 //! used to send information to a newborn mthread
 struct DNSComboWriter
 {
-  DNSComboWriter(const std::string& query, const struct timeval& now) :
-    d_mdp(true, query), d_now(now), d_query(query)
+  DNSComboWriter(const std::string& query, const struct timeval& now, shared_ptr<RecursorLua4> luaContext) :
+    d_mdp(true, query), d_now(now), d_query(query), d_luaContext(luaContext)
   {
   }
 
--- a/pdns/recursordist/rec-main.hh
+++ b/pdns/recursordist/rec-main.hh
@@ -53,8 +53,8 @@ struct DNSComboWriter
   {
   }
 
-  DNSComboWriter(const std::string& query, const struct timeval& now, std::unordered_set<std::string>&& policyTags, LuaContext::LuaObject&& data, std::vector<DNSRecord>&& records) :
-    d_mdp(true, query), d_now(now), d_query(query), d_policyTags(std::move(policyTags)), d_records(std::move(records)), d_data(std::move(data))
+  DNSComboWriter(const std::string& query, const struct timeval& now, std::unordered_set<std::string>&& policyTags, shared_ptr<RecursorLua4> luaContext, LuaContext::LuaObject&& data, std::vector<DNSRecord>&& records) :
+    d_mdp(true, query), d_now(now), d_query(query), d_policyTags(std::move(policyTags)), d_records(std::move(records)), d_luaContext(luaContext), d_data(std::move(data))
   {
   }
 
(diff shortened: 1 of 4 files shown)


rgacogne (commented):
Do you understand where the STL error comes from before that PR? At first sight it's not clear to me why we cannot use the new context, as it looks like we are using bindings and functions that are associated to that new context. I'm assuming we are mixing something that comes from the old one with the new one, but I'm not sure what it is exactly?

omoerbeek:
The test code in #11289 uses `data` which is stored in the `DNSQuestion` object stored in the `SyncRes` object.  Modifying `data` from a newly created Lua context causes the exception, I suppose because the Lua wrapper  code detects the `postresolve` function is defined in a different (created by the reload) context than the `data` object is accesses.  So calling `postresolve` should be done using the Lua context stored in the SyncRes` object.

With the diff I could not reproduce the exception, without it is was very easy.

rgacogne:
Ah, thanks, that makes a lot of sense. I couldn't figure what could be passed from the old context to the new one, but a `LuaContext::LuaObject` is tied to the Lua context from the moment it is constructed indeed. It means we need to be very careful not to yield between the moment we create that object and the moment we call `SyncRes::setLuaEngine`, though. That's not a problem with the current code, but perhaps it would be wiser to store a copy of the `RecursorLua4` shared pointer inside our `DNSComboWriter` to make the relationship clearer?

omoerbeek:
I see your point. I'll take a look if that is possible.
On a side note: I'm not a big fan of how `DNSQuestion` is a mix of values and refs or pointers to other objects.
Someday I would like investigate if it's possible to move it inside DNSCombowriter, change the types to be values  and do away with all the locals in `startDoResolve()`.

pdns/recursordist/rec-main.hh
```diff
@@ -119,6 +119,7 @@ struct DNSComboWriter
   std::unordered_set<std::string> d_policyTags;
   std::string d_routingTag;
   std::vector<DNSRecord> d_records;
+  shared_ptr<RecursorLua4> d_luaContext;
```
rgacogne, line 122:
```suggestion
  // d_data is tied to this LuaContext so we need to keep it alive and use it, not a newer one, as long as d_data exist
  shared_ptr<RecursorLua4> d_luaContext;
```

rgacogne (approved):
Looks good, thanks you! I suggested one comment, and I think we can get rid of the `// XXX correct to use t_pdl?` comments before merging.

rgacogne (approved):

omoerbeek:
Looking into the regress failures

PR #11304, opened 2022-02-08
Stop using the boost version of tie and tuple by default

rgacogne:
### Short description

Also try to pull a bit less of boost headers everywhere in our code base. Hopefully it will make compilation a bit faster but I mostly hope to avoid using the boost version instead of the `std` one by default without knowing it.

This PR needs a good review, as the changes should not break anything but it might some subtle differences between the boost and `std` versions that I missed.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/namespaces.hh
+++ b/pdns/namespaces.hh
@@ -21,7 +21,6 @@
  */
 #pragma once
 
-#include <boost/tuple/tuple.hpp>
 #include <boost/format.hpp>
 #include <boost/optional.hpp>
 #include <iostream>
@@ -30,12 +29,6 @@
 #include <set>
 #include <string>
 #include <vector>
-
-// We have a few paces where boost::tuple is used, and other places where an unscoped tuple is used
-// prefer the boost one for now. We might want to switch to std::tuple one day. Same for tie.
-using boost::make_tuple;
-using boost::tuple;
-using boost::tie;
 
 using std::cerr;
 using std::clog;
--- a/modules/bindbackend/bindbackend2.hh
+++ b/modules/bindbackend/bindbackend2.hh
@@ -29,8 +29,6 @@
 #include <mutex>
 #include <boost/utility.hpp>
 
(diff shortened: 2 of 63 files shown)


pdns/namespaces.hh
```diff
@@ -21,7 +21,6 @@
  */
 #pragma once
 
-#include <boost/tuple/tuple.hpp>
 #include <boost/format.hpp>
```
omoerbeek, line 24: It would be nice to get rid of the remaining boost includes as well here and move them to the specific places where they are needed. But that 's likely better done in a separate PR.

omoerbeek (changes requested):
Reads and builds fine on OpenBSD as well, but is not complete I think. 
`grep -lr 'tie(' *.[ch][c] | grep -v std::` gives me a few remaining hits in auth, rec and dnsdist, same for `tuple(`.

I'm as well a bit worried about potential subtle changes, but this is the right moment in the release cycles.

rgacogne:
> Reads and builds fine on OpenBSD as well, but is not complete I think.

It depends on what our goal is :) My main goal was to remove the `using` lines making the boost version the one used by default:
```
using boost::make_tuple;
using boost::tuple;
using boost::tie;
```
and then I fixed the places that were no longer compiling.
I would be very happy to do more than that, though :)

omoerbeek:
> > Reads and builds fine on OpenBSD as well, but is not complete I think.
> 
> It depends on what our goal is :) My main goal was to remove the `using` lines making the boost version the one used by default:
> 
> ```
> using boost::make_tuple;
> using boost::tuple;
> using boost::tie;
> ```
> 
> and then I fixed the places that were no longer compiling. I would be very happy to do more than that, though :)

My main issue is that it is unclear to which version "naked" `tie` and `tuple` refer to. Better be explicit.

omoerbeek (approved):
Added a small commit with three more cases

PR #11305, opened 2022-02-09
rec: Detect a malformed question early so we can drop it asap

omoerbeek:
it instead of letting a worker do that

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [X] documented the code

### DIFF
--- a/pdns/test-dnsname_cc.cc
+++ b/pdns/test-dnsname_cc.cc
@@ -391,39 +391,48 @@ BOOST_AUTO_TEST_CASE(test_QuestionHash) {
   vector<unsigned char> packet(sizeof(dnsheader));
   reportBasicTypes();
 
+  bool ok;
   // A return init case
-  BOOST_CHECK_EQUAL(hashQuestion(packet.data(), sizeof(dnsheader), 0xffU), 0xffU);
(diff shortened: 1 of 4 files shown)


pdns/test-dnsname_cc.cc
```diff
@@ -391,39 +391,48 @@ BOOST_AUTO_TEST_CASE(test_QuestionHash) {
...
+  auto hash2 = hashQuestion(&packet[0], packet.size() - 4, 0, ok);
   BOOST_CHECK_EQUAL(hash1, hash2);
+  BOOST_CHECK(ok);
 
   vector<uint32_t> counts(1500);
   for(unsigned int n = 0; n < 100000; ++n) {
     packet.clear();
     DNSPacketWriter dpw(packet, DNSName(std::to_string(n) + "." + std::to_string(n*2) + "."), QType::AAAA);
-    counts[hashQuestion(&packet[0], packet.size() - 4, 0) % counts.size()]++;
+    BOOST_CHECK(ok);
```
rgacogne, line 434: I would remove that one, or maybe assert on it if we really care, because `BOOST_CHECK()` is reaaaaally slow.

rgacogne (approved):

PR #11293, opened 2022-02-04
dnsdist: Dynamic discovery and upgrade of backends

rgacogne:
### Short description

Implement https://datatracker.ietf.org/doc/html/draft-ietf-add-ddr-04 to automatically discover if we can upgrade Do53 backends to DoT or DoH.

<strike>Needs regression tests.</strike>

### Checklist

I have:
- [x] read the [CONTRIBUTING.md](https://github.com/PowerDNS/pdns/blob/master/CONTRIBUTING.md) document
- [x] compiled this code
- [x] tested this code
- [x] included documentation (including possible behaviour changes)
- [ ] documented the code
- [x] added or modified regression test(s)
- [x] added or modified unit test(s)

PR #11307, opened 2022-02-09
rec: Add SNI to DoT if available

omoerbeek:
In the forwarding case, it is not available until we have a better config language)

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [X] documented the code

### DIFF
--- a/pdns/lwres.cc
+++ b/pdns/lwres.cc
@@ -248,7 +248,7 @@ static void logIncomingResponse(const std::shared_ptr<std::vector<std::unique_pt
   }
 }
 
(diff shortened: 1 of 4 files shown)


rgacogne (approved):
That looks good. Note that #11293 introduces better support for the validation of IP addresses in certificates presented by servers, so it might make sense to support passing the IP address when we do not have a name.

omoerbeek:
> That looks good. Note that #11293 introduces better support for the validation of IP addresses in certificates presented by servers, so it might make sense to support passing the IP address when we do not have a name.

Note that we do not do any validation at the moment. So there's more work in that area.
Once of the issues is that out config language does not allow good specification of validation parameters atm.

PR #11308, opened 2022-02-09
rec: Remember parent NS set, to be able to fallback to it if needed

### LINKED ISSUE #10594: Consider auth NS records we receive from parent domains for recursor NS cache
- Program: Recursor
 - Issue type: Feature request

### Short description
When resolving a name, powerdns-recursor should consider adding NS records it receives from "parent" (closer to .) name servers to the NS cache for a zone. 

### Usecase
Resolve domains which are misconfigured, but where we could know where and and how to get the proper answer.

#### Given:
Imaging example.tld having NS ns.foo.tld  at the .tld TLD. The zone (served by ns.foo.tld) has an A record for www.example.tld 10.10.10.10, and has NS records for example.tld set to NS ns.bar.tld. ns.bar.tld doesn't answer on queries to www.example.tld.

#### Observed behaviour:
The first time (empty cache) pdns-rec must resolve www.example.tld, it succeeds: pdns-rec will ask ns.foo.tld and accept the result. Hence, the client gets a result.

When processing the first result, PowerDNS will cache the NS records it receives from ns.foo.tld as the (only) NS records for example.tld, so it stores ns.bar.tld as the NS for example.tld.

This means that the second query for eg. mail.example.tld will use the cached NS set – consisting of only ns.bar.tld – which doesn't reply.

### Description
PowerDNS should add both NSs to the cache set. The first NS (ns.foo.tld) is received from a trusted "parent" which sends a record we trust (and use in the case of an empty cache). Why not add it to the cache?

I know that in the case above, the domain is misconfigured. And that we shouldn't encourage this behaviour. But when I look at the failures that are reported to me for people using pdns-resolver, this issue 

However, other resolvers (both closed and open source) are able to resolve www.example.tld perfectly fine in this case.

I considered reporting this as a bug instead of a feature request, as the PowerDNS behaviour depends on the state of the cache.

omoerbeek:
When a non-auth NS set is replaced by an auth one, remember the non-auth one if it contains NS not in the auth set.

Use that NS set later as a fallback if resolving fails.

TODO:
 
- [x] general validation of method
- [x] pruning the new table (based on TTL of child and parent NS set?)
- [x] tests
- [x] config switch? Ir can this be alwasy-on?
- [x] docs
- [x] move the implementation of the data structure out of syncres.hh, it's already way too big

Should fix #10594

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [x] documented the code

### DIFF
--- a/pdns/rec_channel_rec.cc
+++ b/pdns/rec_channel_rec.cc
@@ -377,6 +377,11 @@ static uint64_t* pleaseDumpFailedServers(int fd)
   return new uint64_t(SyncRes::doDumpFailedServers(fd));
 }
 
+static uint64_t* pleaseDumpSavedParentNSSets(int fd)
+{
+  return new uint64_t(SyncRes::doDumpSavedParentNSSets(fd));
+}
+
 static uint64_t* pleaseDumpNonResolvingNS(int fd)
 {
   return new uint64_t(SyncRes::doDumpNonResolvingNS(fd));
@@ -1904,6 +1909,8 @@ RecursorControlChannel::Answer RecursorControlParser::getAnswer(int s, const str
             "dump-failedservers <filename>    dump the failed servers to the named file\n"
             "dump-non-resolving <filename>    dump non-resolving nameservers addresses to the named file\n"
             "dump-nsspeeds <filename>         dump nsspeeds statistics to the named file\n"
+            "dump-saved-parent-ns-sets <filename>\n"
+            "                                 dump saved parent ns sets that were used successfully as fallback\n"
             "dump-rpz <zone name> <filename>  dump the content of a RPZ zone to the named file\n"
             "dump-throttlemap <filename>      dump the contents of the throttle map to the named file\n"
             "get [key1] [key2] ..             get specific statistics\n"
@@ -1979,6 +1986,9 @@ RecursorControlChannel::Answer RecursorControlParser::getAnswer(int s, const str
   }
   if (cmd == "dump-failedservers") {
     return doDumpToFile(s, pleaseDumpFailedServers, cmd, false);
+  }
+  if (cmd == "dump-saved-parent-ns-sets") {
+    return doDumpToFile(s, pleaseDumpSavedParentNSSets, cmd, false);
   }
   if (cmd == "dump-rpz") {
     return doDumpRPZ(s, begin, end);
--- a/pdns/rec_channel_rec.cc
+++ b/pdns/rec_channel_rec.cc
@@ -1910,7 +1910,7 @@ RecursorControlChannel::Answer RecursorControlParser::getAnswer(int s, const str
             "dump-non-resolving <filename>    dump non-resolving nameservers addresses to the named file\n"
             "dump-nsspeeds <filename>         dump nsspeeds statistics to the named file\n"
             "dump-saved-parent-ns-sets <filename>\n"
-            "                                 dump saved parent ns sets that were used successfully as fallback\n"
+            "                                 dump saved parent ns sets that were successfully used as fallback\n"
             "dump-rpz <zone name> <filename>  dump the content of a RPZ zone to the named file\n"
             "dump-throttlemap <filename>      dump the contents of the throttle map to the named file\n"
             "get [key1] [key2] ..             get specific statistics\n"
--- a/pdns/recursordist/docs/appendices/internals.rst
+++ b/pdns/recursordist/docs/appendices/internals.rst
@@ -445,6 +445,15 @@ new data should overwrite old data.
 Note that PowerDNS deviates from RFC 2181 (section 5.4.1) in this
 respect.
 
+Starting with version 4.7.0, there is a mechanism to save the
+parent NS set if it contains *more* names than the child NS set.
+This allows fallback to the saved parent NS set on resolution errors
+using the child specified NS set.
+As experience shows, this configuration error is encountered in the
+wild often enough to warrant this workaround.
+See :ref:`setting-save-parent-ns-set`.
+
+
  Some small things
 ------------------
 
--- a/pdns/recursordist/docs/appendices/internals.rst
+++ b/pdns/recursordist/docs/appendices/internals.rst
@@ -447,7 +447,7 @@ respect.
 
 Starting with version 4.7.0, there is a mechanism to save the
 parent NS set if it contains *more* names than the child NS set.
-This allows fallback to the saved parent NS set on resolution errors
+This allows falling back to the saved parent NS set on resolution errors
 using the child specified NS set.
 As experience shows, this configuration error is encountered in the
 wild often enough to warrant this workaround.
--- a/pdns/recursordist/docs/manpages/rec_control.1.rst
+++ b/pdns/recursordist/docs/manpages/rec_control.1.rst
@@ -123,9 +123,9 @@ dump-rpz *ZONE NAME* *FILE NAME*
 
 dump-saved-parent-ns-sets *FILE NAME*
     Dump the entries of the map containing saved parent NS sets
-    that were used successfully in resolving.
+    that were successfully used in resolving.
     The total number of entries is also printed in the header.
-    An entry is saved if the recursor sees that parent set includes
+    An entry is saved if the recursor sees that the parent set includes
     names not in the child set. This is an indication of a
     misconfigured domain.
 
--- a/pdns/recursordist/docs/manpages/rec_control.1.rst
+++ b/pdns/recursordist/docs/manpages/rec_control.1.rst
@@ -121,6 +121,14 @@ dump-rpz *ZONE NAME* *FILE NAME*
     overwrite it otherwise. While dumping, the recursor will not answer
     questions.
 
+dump-saved-parent-ns-sets *FILE NAME*
+    Dump the entries of the map containing saved parent NS sets
+    that were used successfully in resolving.
+    The total number of entries is also printed in the header.
+    An entry is saved if the recursor sees that parent set includes
+    names not in the child set. This is an indication of a
+    misconfigured domain.
+
 dump-throttlemap *FILENAME*
     Dump the contents of the throttle map to the *FILENAME* mentioned.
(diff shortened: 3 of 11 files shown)


franklouwers:
interesting

omoerbeek:
I think this is ready for a first round of review. If the method is considered sane and useful, I'll work on finishing this: documentation, tests, maybe a config switch.

pdns/syncres.cc
```diff
@@ -4628,8 +4739,13 @@ int SyncRes::doResolveAt(NsSet &nameservers, DNSName auth, bool flawedNSSet, con
         }
       }
       else {
-        /* if tns is empty, retrieveAddressesForNS() knows we have hardcoded servers (i.e. "forwards") */
-        remoteIPs = retrieveAddressesForNS(prefix, qname, tns, depth, beenthere, rnameservers, nameservers, sendRDQuery, pierceDontQuery, flawedNSSet, cacheOnly, addressQueriesForNS);
+        if (fallBack == nullptr) {
+          /* if tns is empty, retrieveAddressesForNS() knows we have hardcoded servers (i.e. "forwards") */
+          remoteIPs = retrieveAddressesForNS(prefix, qname, tns, depth, beenthere, rnameservers, nameservers, sendRDQuery, pierceDontQuery, flawedNSSet, cacheOnly, addressQueriesForNS);
+        } else {
+          // should be save, caller makes sure nameservers and fallback contain the same names
```
rgacogne, line 4746:
```suggestion
          // should be safe, caller makes sure nameservers and fallback contain the same names
```

pdns/syncres.cc
```diff
@@ -3401,6 +3458,55 @@ void SyncRes::sanitizeRecords(const std::string& prefix, LWResult& lwr, const DN
...
+    auto content = getRR<NSRecordContent>(ns);
+    authSet.insert(content->getNS());
+  }
+  // The glue IPs could also differ, but we're not checking that yet, we're only looking for child NS records not
+  // in the parent set
+  bool shouldSave = false;
+  for (const auto& ns : existing) {
+    auto content = getRR<NSRecordContent>(ns);
+    if (authSet.count(content->getNS()) == 0) {
+      LOG(d_prefix << domain << ": at least one child NS was not in the parent NS set, remembering parent NS set and cached IPs" << endl);
```
rgacogne, line 3490:
```suggestion
      LOG(d_prefix << domain << ": at least one parent-side NS was not in the child-side NS set, remembering parent NS set and cached IPs" << endl);
```

rgacogne (approved):
The general logic looks good. Out of curiosity, did you do a large run (top 1m or something like that to see how many entries end up in the table?

omoerbeek:
> The general logic looks good. Out of curiosity, did you do a large run (top 1m or something like that to see how many entries end up in the table?

Nope, I first wanted to collect some feedback on the general mechanism before proceeding. I'll spend some time on testing soon.

omoerbeek:
Did a run with umbrella top 1M.
```
$ ./rc dump-saved-parent-ns-sets -
; dump of saved parent nameserver sets succesfully used follows
; total entries: 4577
; domain        success ttd
fundacite-merida.gob.ve.        2       Tue Mar 15 19:43:02 2022
newamerica.org. 1       Wed Mar 16 14:43:45 2022
cnd.org.        1       Wed Mar 16 14:41:11 2022
tridenttech.org.        3       Wed Mar 16 14:38:51 2022
hackettstown.org.       1       Wed Mar 16 14:42:35 2022
cloudw.infinitecampus.org.      1       Wed Mar 16 14:34:31 2022
seed.bitcoin.jonasschnelli.ch.  1       Wed Mar 16 14:42:12 2022
follow.hk.      1       Tue Mar 15 22:44:30 2022
langdons.co.uk. 1       Wed Mar 16 14:47:23 2022
p.jibjab.com.   3       Tue Mar 15 15:04:30 2022
demo-gs-stakelogic.com. 1       Wed Mar 16 14:39:59 2022
nahdionline.com.        2       Wed Mar 16 14:38:13 2022
myscarlettmonroe.com.   1       Wed Mar 16 14:45:03 2022
weather-and-climate.com.        2       Wed Mar 16 14:36:02 2022
celebritycarsblog.com.  1       Wed Mar 16 14:43:08 2022
ezlifetech.com. 1       Wed Mar 16 14:38:26 2022
tritontek.com.  1       Wed Mar 16 14:41:52 2022
noplink.com.    1       Wed Mar 16 14:47:42 2022
evinternal.com. 1       Wed Mar 16 14:34:05 2022
achoon.com.     2       Wed Mar 16 14:36:34 2022
respawn.com.    3       Wed Mar 16 14:33:01 2022
thaismescenter.com.     1       Wed Mar 16 14:46:32 2022
bondageobserver.com.    1       Wed Mar 16 14:43:24 2022
yalovaunvertur.com.     1       Wed Mar 16 14:42:28 2022
cablemas.com.   1       Wed Mar 16 14:42:07 2022
blogtopsites.com.       1       Wed Mar 16 14:46:06 2022
fappenist.com.  1       Wed Mar 16 14:41:46 2022
pixibeauty.com. 1       Wed Mar 16 14:41:52 2022
youth.cn.       2       Wed Mar 16 14:35:12 2022
dcb.ln.cn.      1       Wed Mar 16 14:37:27 2022
axiscapital.co.in.      1       Wed Mar 16 14:44:01 2022
cun.edu.co.     1       Tue Mar 15 15:45:07 2022
interleads.net. 1       Wed Mar 16 14:49:04 2022
istole.it.      1       Tue Mar 15 17:33:16 2022
q1.lt.  3       Tue Mar 15 16:36:15 2022
southtexascollege.edu.  3       Wed Mar 16 14:41:23 2022
cdc.gov.        1       Wed Mar 16 14:32:43 2022
dump-saved-parent-ns-sets: dumped 37 records
```
nsspeeds-entries        84787
```

**Comment by omoerbeek** [comment]
I spot checked a few of the 4% cases. They often boil down to the child set being smaller or using different names, but in the end having the same (subset of) IPs.

An issue here is that the code might not know the IPs at the point it needs to decide to save the parent set or not, so it cannot base the decision on the actual IPs in the parent vs child set, it has to do it using names.

So in the end we are saving to a bit to many parent sets.  I *could* consider actively deleting the parent set if I see the child set is used successfully, but I'm not sure that's worth the added complexity, plus the saved set might still be useful to resolve *some* names in the case the child set contains some broken name servers.

**Comment by rgacogne** [comment]
I wonder if saving that many sets is really an issue. I don't expect the memory difference to be noticeable in most cases, so that would leave the overhead of restarting with the parent set when we failed to resolve with the child one. Do you know how often that actually happens?

**Comment by omoerbeek** [comment]
In my test of 1m domains 53 domains were fixed. Some domains had multiple cases, see report above.

**Comment by rgacogne** [comment]
> In my test of 1m domains 53 domains were fixed. Some domains had multiple cases, see report above.

I missed that line, sorry! That seems quite acceptable to me, then?

File: pdns/rec_channel_rec.cc [hunk 3]
```diff
@@ -1904,6 +1909,8 @@ RecursorControlChannel::Answer RecursorControlParser::getAnswer(int s, const str
             "dump-failedservers <filename>    dump the failed servers to the named file\n"
             "dump-non-resolving <filename>    dump non-resolving nameservers addresses to the named file\n"
             "dump-nsspeeds <filename>         dump nsspeeds statistics to the named file\n"
+            "dump-saved-parent-ns-sets <filename>\n"
+            "                                 dump saved parent ns sets that were used successfully as fallback\n"
```
jsoref, line 1913:
```suggestion
            "                                 dump saved parent ns sets that were successfully used as fallback\n"
```

pdns/recursordist/docs/appendices/internals.rst
```diff
@@ -445,6 +445,15 @@ new data should overwrite old data.
 Note that PowerDNS deviates from RFC 2181 (section 5.4.1) in this
 respect.
 
+Starting with version 4.7.0, there is a mechanism to save the
+parent NS set if it contains *more* names than the child NS set.
+This allows fallback to the saved parent NS set on resolution errors
```
jsoref, line 450:
```suggestion
This allows falling back to the saved parent NS set on resolution errors
```
fallback would be a noun, and I don't think this wants a noun.

pdns/recursordist/docs/manpages/rec_control.1.rst
```diff
@@ -121,6 +121,14 @@ dump-rpz *ZONE NAME* *FILE NAME*
     overwrite it otherwise. While dumping, the recursor will not answer
     questions.
 
+dump-saved-parent-ns-sets *FILE NAME*
+    Dump the entries of the map containing saved parent NS sets
+    that were used successfully in resolving.
```
jsoref, line 126:
```suggestion
    that were successfully used in resolving.
```
not a fan of `in`, maybe `while` or `during` or `when`?

pdns/recursordist/docs/manpages/rec_control.1.rst
```diff
@@ -121,6 +121,14 @@ dump-rpz *ZONE NAME* *FILE NAME*
...
+    The total number of entries is also printed in the header.
+    An entry is saved if the recursor sees that parent set includes
```
jsoref, line 128:
```suggestion
    An entry is saved if the recursor sees that the parent set includes
```

pdns/recursordist/docs/upgrade.rst
```diff
@@ -18,6 +18,10 @@ If IPv6 is enabled for outgoing queries using :ref:`setting-query-local-address`
 These addresses will then be used for future queries to authoritative nameservers.
 This has the consequence that authoritative nameservers will be contacted over IPv6 in more case than before.
 
+New settings
+^^^^^^^^^^^^
+- The :ref:`settings-save-parent-ns-set` setting has been introduced, enabling fall-back cases if the parent ``NS`` set contains names not in the child ``NS`` set.
```
jsoref, line 23: note that this uses `fall-back` as opposed to the `fallback` I flagged above.

PR #11310, opened 2022-02-09
Remove include boost/format.hpp from namespaces.hh 

omoerbeek:
Only include it where used

### Checklist

I have:
- [X] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/iputils.cc
+++ b/pdns/iputils.cc
@@ -22,8 +22,11 @@
 #ifdef HAVE_CONFIG_H
 #include "config.h"
 #endif
+
 #include "iputils.hh"
-#include <sys/socket.h> 
+
+#include <sys/socket.h>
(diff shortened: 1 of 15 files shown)


pdns/iputils.cc
```diff
@@ -22,8 +22,11 @@
 #ifdef HAVE_CONFIG_H
 #include "config.h"
 #endif
+
 #include "iputils.hh"
-#include <sys/socket.h> 
+
+#include <sys/socket.h>
+#include <boost/format.hpp>
```
rgacogne, line 29: I'm actually removing that usage in #11293 so we will need to rebase one of these PRs, likely #11293, but that's fine.

rgacogne (approved):

PR #11312, opened 2022-02-11
Rec: packet cache improvements

omoerbeek:
### Short description

This PR changes the eviction policy for the packet cache. It requires less operations in the LRU list and spreads the cost of pruning the full cache by removing an item when an item is added to a full cache.

This makes the order of the PC cache now strictly LRU, with the reasoning that if a cache item is found to be expired, we are going to update it very soon as we only look for entry we are looking up for a client. If it is expired, a resolve will follow immediately. (no recursion desired is really an exception). So the existing method of moving it to the front will be followed by an move to the back very soon. We do keep moving it to the back of it is not expired.

So what this PR does is:

- Tell the PC its size, so that it can evict an item when an item is added to a full cache
- If we get a PC miss, do not move the item to the front the of "to be deleted" list. We will *very likely* update the item soon and move it to the back of the "to be deleted queue". So leave it.
- only create and prune the packetcache for threads that need it
- If the packet cache is disabled, warn that setting the size using rec_control is a no-op

Test setup: a recursor with a `pre_resolve` Lua function that sets an answer and returns true. This way mostly the PC cache performance is measured.

Wit master and using `dnsperf` with a large test file I see the memory usage and PC entries grow beyond the limit. Only when the cleanup action is done, the number of PC cache entries it reduced again.

With this PR, the PC entries never grow beyond the limit and the RES size is lower, as expected.

The QPS I'm seeing vary, but in general the PR code is a bit faster.

If I use a smaller test file that fits in the PC, I'm seeing a speedup as well since the cache lookup has a little bit less work to do.

Additionally fix the PC hit ratio and qps printed by stats.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/recpacketcache.cc
+++ b/pdns/recpacketcache.cc
@@ -92,8 +92,8 @@ bool RecursorPacketCache::checkResponseMatches(std::pair<packetCache_t::index<Ha
       return true;
     }
     else {
-      // We used to move the item to the fron ot the to be deleted sequence,
-      // but we're very likely will update the entry very soon, so leave it
+      // We used to move the item to the front of "the to be deleted" sequence,
+      // but we very likely will update the entry very soon, so leave it
       d_misses++;
       break;
     }
--- a/pdns/recpacketcache.cc
+++ b/pdns/recpacketcache.cc
@@ -10,11 +10,6 @@
 #include "namespaces.hh"
 #include "rec-taskqueue.hh"
 
-RecursorPacketCache::RecursorPacketCache()
-{
-  d_hits = d_misses = 0;
-}
-
 unsigned int RecursorPacketCache::s_refresh_ttlperc{0};
 
 int RecursorPacketCache::doWipePacketCache(const DNSName& name, uint16_t qtype, bool subtree)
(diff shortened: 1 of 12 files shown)


pdns/recpacketcache.cc
```diff
@@ -195,15 +191,15 @@ void RecursorPacketCache::insertResponsePacket(unsigned int tag, uint32_t qhash,
...
 
     d_packetCache.insert(e);
-  }
-}
 
-uint64_t RecursorPacketCache::size()
-{
-  return d_packetCache.size();
+    if (d_packetCache.size() > d_maxSize) {
+      auto it = d_packetCache.begin();
```
rgacogne, line 196: I would expect this to take the first item of the first index of `d_packetCache`, so the `HashTag` one, and not the ` SequencedTag`? In any case I would prefer explicitly using the `SequencedTag` index so our intention is clear.

omoerbeek: Ugh, thanks for spotting that!

rgacogne (approved):

omoerbeek:
Rebased to fix conflict

PR #11315, opened 2022-02-11
el7 builds: switch to boost 1.69

Habbie:
### Checklist

I have:
- [x] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/builder-support/specs/dnsdist.spec
+++ b/builder-support/specs/dnsdist.spec
@@ -11,14 +11,19 @@ BuildRequires: libedit-devel
 BuildRequires: openssl-devel
 
 %if 0%{?suse_version}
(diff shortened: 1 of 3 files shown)


Habbie:
I have tested the resulting binaries (dnsdist forwarding a few queries, Recursor correctly setting/not setting AD for a few queries, and auth responding from an sqlite3 database).

I did find one crash in dnsdist (in libedit, I think) when I ^Z it, but this crash also occurs on the previous boost version.

omoerbeek:
I see the console ^Z crash on OpenBSD/master as well.

omoerbeek (approved):
As far as I understand the specfile syntax this looks good.

PR #11319, opened 2022-02-14
Read the base Lua definitions into the Lua context for reading the Lua config

omoerbeek:
This allows symbolic names for e.g. QTypes. Use that in protobufServer() and
outgoingProtobufServer(). Will also be used for the upcoming additional records code.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/rec-lua-conf.cc
+++ b/pdns/rec-lua-conf.cc
@@ -5,6 +5,7 @@
 #include <thread>
 #include "namespaces.hh"
 #include "logger.hh"
+#include "lua-base4.hh"
 #include "rec-lua-conf.hh"
 #include "sortlist.hh"
 #include "filterpo.hh"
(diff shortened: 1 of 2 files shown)


pdns/rec-lua-conf.cc
```diff
@@ -322,11 +322,33 @@ static void rpzPrimary(LuaConfigItems& lci, luaConfigDelayedThreads& delayedThre
   delayedThreads.rpzPrimaryThreads.push_back(std::make_tuple(primaries, defpol, defpolOverrideLocal, maxTTL, zoneIdx, tt, maxReceivedXFRMBytes, localAddress, axfrTimeout, refresh, sr, dumpFile));
 }
 
+// A wrapper clas that load the snadrad Lua defintions into the conext, so that we can use things
```
rgacogne, line 325:
```suggestion
// A wrapper class that loads the standard Lua definitions into the context, so that we can use things
```

pdns/rec-lua-conf.cc
```diff
@@ -322,11 +322,33 @@ static void rpzPrimary(LuaConfigItems& lci, luaConfigDelayedThreads& delayedThre
...
+// like pdns.A
+class RecLuaConfigContext : public BaseLua4
+{
+public:
+  RecLuaConfigContext()
+  {
+    prepareContext();
+  }
+  virtual void postPrepareContext() override
```
rgacogne, line 334:
```suggestion
  void postPrepareContext() override
```

pdns/rec-lua-conf.cc
```diff
@@ -322,11 +322,33 @@ static void rpzPrimary(LuaConfigItems& lci, luaConfigDelayedThreads& delayedThre
...
+  {
+  }
+  void postLoad() override
+  {
+  }
+  std::unique_ptr<LuaContext>& operator()()
```
rgacogne, line 340: Perhaps overriding the `->` operator and returning a `LuaContext&` would have been easier to use?

omoerbeek:
@rgacogne overloading `operator->` won't work in this case I think, as there is a restriction:

`The overload of operator -> must either return a raw pointer, or return an object (by reference or by value) for which operator -> is in turn overloaded. `

`LuaContext` itsef has no `operator->` . So I went for the raw pointer case.

rgacogne (approved):

PR #11346, opened 2022-02-22
auth: Basic abstraction for handling colored terminal output, respecting isatty(), --no-colors and NO_COLOR

### LINKED ISSUE #11341: pdnsutil color handling
- Program: Authoritative 
 - Issue type: Both a bit of Feature request and Bug report

There are a few issues with `pdnsutil` sending ANSI escape sequences unconditionally

1) Literal escape codes inside C++ code is ugly
2) They are sent unconditionally, no support for `NO_COLOR` or `--no-color`
3) They also potentialy end up in log files when scripting `pdnsutil` as there is no check for `isatty()`
4) No support for non-ANSI using termcap or other terminal capability lib.

I can live with 4), but the other points should be addressed, potentially by defining a helper class that uses `isatty()`, en respects `NO_COLOR` and/or `--no-color` and that can be used to insert the escape sequences and that is a no-op when not applicable :

```
pdns::TerminalColor color(...);
cout << color.red() << .... << color.reset() << endl
```

omoerbeek:
A pet peeve from me...

Fixes #11341

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/pdnsutil.cc
+++ b/pdns/pdnsutil.cc
@@ -1136,7 +1136,34 @@ static int clearZone(DNSSECKeeper& dk, const DNSName &zone) {
   return EXIT_SUCCESS;
 }
 
-static int editZone(const DNSName &zone) {
+class PDNSColors
+{
+public:
+  PDNSColors(bool nocolors)
+    : d_colors(!nocolors && isatty(STDOUT_FILENO) && getenv("NO_COLORS") == nullptr)
+  {
+  }
+  [[nodiscard]] string red() const
+  {
+    return d_colors ? "\x1b[31m" : "";
+  }
+  [[nodiscard]] string green() const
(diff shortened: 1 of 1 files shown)


pdns/pdnsutil.cc
```diff
@@ -84,7 +84,7 @@ static void loadMainConfig(const std::string& configdir)
   if(::arg().mustDo("help")) {
     cout<<"syntax:"<<endl<<endl;
     cout<<::arg().helpstring(::arg()["help"])<<endl;
-    exit(0);
+   exit(0);
```
jsoref, line 87: revert stray damage
```suggestion
    exit(0);
```

Habbie (approved):

PR #11349, opened 2022-02-23
Rec: do not tread qtype ALIAS and LUA special

omoerbeek:
### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code


Habbie (approved):
Looks good. Alternatively, "if ! REC" could also be "if AUTH", but the difference is somewhat academic in our codebase.

omoerbeek:
> Looks good. Alternatively, "if ! REC" could also be "if AUTH", but the difference is somewhat academic in our codebase.

ATM there is no AUTH define AFAIKS, only RECURSOR

Habbie:
> ATM there is no AUTH define AFAIKS, only RECURSOR

Even more academic than I thought, then ;)

omoerbeek:
We do have some test failures
```
Feb 23 09:41:51 STL Exception: Parsing zone content on line 20193 of file '../regression-tests/zones/example.com': 'ALIAS' doesn't look like a qtype, stopping loop
```

PR #11351, opened 2022-02-24
Update README.md for c++17 support

omoerbeek:
### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ This will bring up a USAGE-page which will explain how to build the different re
 COMPILING Authoritative Server
 ------------------------------
 The PowerDNS Authoritative Server depends on Boost, OpenSSL and Lua, and requires a
-compiler with C++-2011 support.
+compiler with C++-2017 support.
 
 On Debian 9, the following is useful:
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -144,10 +144,10 @@ FreeBSD Notes
 -------------
 You need to compile using gmake - regular make only appears to work, but doesn't in fact. Use gmake, not make.
 
-The clang compiler installed through FreeBSD's package manager does not expose all of the C++11 features needed under `std=gnuc++11`. Force the compiler to use `std=c++11` mode instead.
+The clang compiler installed through FreeBSD's package manager does not expose all of the C++17 features needed under `std=gnuc++11`. Force the compiler to use `std=c++17` mode instead.
```
rgacogne, line 147: If that's still true:
```suggestion
The clang compiler installed through FreeBSD's package manager does not expose all of the C++17 features needed under `std=gnuc++17`. Force the compiler to use `std=c++17` mode instead.
```

omoerbeek: Default on FreeBSD is `gnuc++14`, see https://www.freebsd.org/cgi/man.cgi?query=cc&apropos=0&sektion=0&manpath=FreeBSD+13.0-RELEASE+and+Ports&arch=default&format=html

So it should become:

...C++17 features  needed under the default `std=gnuc++14`

omoerbeek, line 147:
```suggestion
The clang compiler installed through FreeBSD's package manager does not expose all of the C++17 features needed under the default `std=gnuc++14`. Force the compiler to use `std=c++17` mode instead.
```

PR #11364, opened 2022-02-25
rec: Introduce dnsheader_aligned: a helper class to access dnsheader data

omoerbeek:
As the string holding the packet may be potentially unaligned wrt to int access.

I'm using a copy of the header, but the copy is only made if the passed in pointer is pointing to unaligned data.

Some day we should switch to `PacketBuffer` instead of `std::string` for packet data. The first is naturally aligned (it allocates using `new()` instead of the more complex way a `std::string` works) and has the nice feature of not initializing its buffer when (re)sizing.

I am not doing that now as it has tentacles all over the place including the packet cache code that is shared with auth. Also `PacketBuffer` is missing a few methods that the current code uses (e.g. `compare()`. This is all solvable, but now for now.

This was caught by using `-fsanitize=undefined -fsanitize-minimal-runtime` that was included in OpenBSD recently. I don't know why the regular ubsan we use on CI does not catch this. On common archs this is likely not an actual problem, as they allow unaligned access, but it might be on some ARM, MIPS and certainly Sparc64 archs.

This only handles rec. The other products should also be checked for this.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dns.hh
+++ b/pdns/dns.hh
@@ -207,6 +207,30 @@ struct dnsheader {
 
 static_assert(sizeof(dnsheader) == 12, "dnsheader size must be 12");
 
+class dnsheader_aligned
+{
+public:
+  dnsheader_aligned(const void* mem)
+  {
+    if (reinterpret_cast<uintptr_t>(mem) % sizeof(uint32_t) == 0) {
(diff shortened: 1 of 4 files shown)


rgacogne (approved):
:+1:

PR #11368, opened 2022-02-26
Additional geo methods for Lua records

rage4:
### Short description
Extended Lua records with:
- region method, similar functionality as country and continent but for regions
- continentCode, contryCode, regionCode - returns two letter codes, or '--' if unknown

### Checklist

I have:
- [ x] read the [CONTRIBUTING.md](https://github.com/PowerDNS/pdns/blob/master/CONTRIBUTING.md) document
- [ x] compiled this code
- [ x] tested this code
- [ x] included documentation (including possible behaviour changes)
- [ x] documented the code
- [ x] added or modified regression test(s)
- [ x] added or modified unit test(s)

### DIFF
--- a/pdns/dnsdist-lua.cc
+++ b/pdns/dnsdist-lua.cc
@@ -1091,6 +1091,10 @@ static void setupLuaConfig(LuaContext& luaCtx, bool client, bool configCheck)
       setWebserverStatsRequireAuthentication(boost::get<bool>(vars->at("statsRequireAuthentication")));
     }
 
+    if (vars->count("apiRequiresAuthentication")) {
+      setWebserverAPIRequiresAuthentication(boost::get<bool>(vars->at("apiRequiresAuthentication")));
+    }
+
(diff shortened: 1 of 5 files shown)


pdns/lua-record.cc
```diff
@@ -945,6 +952,27 @@ static void setupLuaRecords()
         });
 
     });
+  lua.writeFunction("countryCode", []() {
+      string res = getGeo(s_lua_record_ctx->bestwho.toString(), GeoIPInterface::Country2);
+      if ( res == "unknown" ) {
+       res = "--";
```
setharnold, line 958: Are *all* these strings stored in the text segment?

Or, another way, is this the only string that doesn't need to be deallocated?

rage4: The string handling is similar to the other, existing, methods ("asnum" for example). 
The "unknown" bit is a default response from getGeo in the case when there is no matching country, since it is all two-letter ISO I decided to use "--" instead.

PR #11373, opened 2022-02-28
rec: Add setting to prefer a matching forward over a NS entry found in the cache.

omoerbeek:
This (conditionally) reverts to the pre-4.4 behavior for forwarded domains.

The change to prefer cached NS records was introduced in #9351 to allow e.g. a local root auth zone, but we have reports that in specific (complex) setups this is not desired.

The setting is called `prefer-forward-over-cached-ns` Default remains `no`: prefer cached NS records.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code


rgacogne (approved):
One typo, the PR looks good! I would not mind a unit and/or regression test, though :)

omoerbeek:
> One typo, the PR looks good! I would not mind a unit and/or regression test, though :)

Will do!

omoerbeek:
Closing this. The case in question was solved by using `forward-zones-recurse`.

PR #11381, opened 2022-03-01
rec: Adaptive root refresh interval, normally at 80% of max-cache-ttl

omoerbeek:
and shortening he interval on failure.

The original  code has a fixed 2 hours interval. That was changed to be `max-cache-ttl/12`. But now that we have an easy way to do an adaptive interval that is much preferable.
 
### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/recursordist/docs/appendices/FAQ.rst
+++ b/pdns/recursordist/docs/appendices/FAQ.rst
@@ -56,3 +56,26 @@ Or, in a diagram::
                  |
       client bufsize (stub => recursor)
      bufsize reported to client (recursor => stub [always 512])
+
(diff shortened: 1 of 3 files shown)


rgacogne (approved):
The code looks good, we likely need to document that change? The period will converge to 10s quite quickly in case of repeated errors, but I guess that's the point?

omoerbeek:
Added text, wondering if it is in the right place.

rgacogne (approved):
Looks good to me, great doc! Perhaps we could add a ref from setting-hint-file to that new section?

PR #11388, opened 2022-03-03
dnsdist: add --log-timestamps flag

Habbie:
### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dolog.hh
+++ b/pdns/dolog.hh
@@ -103,7 +103,9 @@ void genlog(int level, const char* s, Args... args)
     time_t t;
     time(&t);
     localtime_r(&t, &tm);
-    strftime(buffer, sizeof(buffer), "%b %d %H:%M:%S ", &tm);
+    if (strftime(buffer, sizeof(buffer), "%b %d %H:%M:%S ", &tm) == 0) {
+      buffer[0] = '\0';
+    }
     std::cout<<buffer;
   }
 #endif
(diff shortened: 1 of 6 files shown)


pdns/dolog.hh
```diff
@@ -88,8 +92,22 @@ void genlog(int level, const char* s, Args... args)
...
     syslog(level, "%s", str.str().c_str());
+
+#ifdef DNSDIST
+  if (g_logtimestamps) {
+    char buffer[50] = "";
+    struct tm tm;
+    time_t t;
+    time(&t);
+    localtime_r(&t, &tm);
+    strftime(buffer, sizeof(buffer), "%b %d %H:%M:%S ", &tm);
```
rgacogne, line 106: We might want to check the return value of `strftime` before reading `buffer` since POSIX states that: `If the total number of resulting bytes including the terminating null byte is not more than maxsize, [...]. Otherwise, 0 is returned and the contents of the array are indeterminate`.

rgacogne (changes requested):

pdns/dolog.hh
Habbie: Good idea! I added a commit fixing that for auth, rec and dnsdist logging (the code above is copied from auth/rec). We have a few other unchecked `strftime` uses but I think that's for a separate PR.

rgacogne (approved):

PR #11389, opened 2022-03-03
auth API: fetch individual rrsets

Habbie:
### Short description
~~needs docs, tests~~

### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/docs/http-api/swagger/authoritative-api-swagger.yaml
+++ b/docs/http-api/swagger/authoritative-api-swagger.yaml
@@ -188,6 +188,14 @@ paths:
           description: '“true” (default) or “false”, whether to include the “rrsets” in the response Zone object.'
           type: boolean
           default: true
+        - name: rrset_name
+          in: query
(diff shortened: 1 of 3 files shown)


neilcook (approved):
LGTM. The only thing I would say is that as discussed previously this changes the behaviour with certain backends such as geoip where specifying the rrset_name will actually return results, because it uses lookup rather than list. I don't know if that's worth documenting somewhere.

Habbie:
> LGTM. The only thing I would say is that as discussed previously this changes the behaviour with certain backends such as geoip where specifying the rrset_name will actually return results, because it uses lookup rather than list. I don't know if that's worth documenting somewhere.

It follows from "Master: Yes" at https://doc.powerdns.com/authoritative/backends/geoip.html but only if you know a lot about how PowerDNS works internally. I'm not sure it's worth documenting more though.

PR #11401, opened 2022-03-08
Checked conversion from string to signed/unsigned integer

fredmorcos:
### Short description
This refactors `pdns_stou` into `Pdns::checked_conv` and implements `Pdns::checked_stoi` (signed) and `Pdns::checked_stou` (unsigned) in terms of that. This is a small step into a good direction of avoiding narrowing casts since those silence compiler warnings and provide no runtime warnings.

Unfortunately, due to the signed vs. unsigned split, each `checked_stoi` call-site needs to be checked individually and the target data-type corrected (if needed) because, previously, all these calls were to `pdns_stou` which only returned unsigned values.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code


codedocs/doxygen.conf
```diff
@@ -213,7 +213,7 @@ SUBGROUPING            = YES
 # Private class members and static file members will be hidden unless 
 # the EXTRACT_PRIVATE and EXTRACT_STATIC tags are set to YES
 
-EXTRACT_ALL            = NO
+EXTRACT_ALL            = YES
```
fredmorcos, line 216: Is this acceptable to everyone? It was the only way I could find to generate documentation for the `pdns::checked_*` functions.

modules/bindbackend/binddnssec.cc
```diff
@@ -323,8 +323,8 @@ bool Bind2Backend::getDomainKeys(const DNSName& name, std::vector<KeyData>& keys
     SSqlStatement::row_t row;
     while (d_getDomainKeysQuery_stmt->hasNextRow()) {
       d_getDomainKeysQuery_stmt->nextRow(row);
-      kd.id = pdns_stou(row[0]);
-      kd.flags = pdns_stou(row[1]);
+      kd.id = pdns::checked_stou<decltype(kd.id)>(row[0]);
```
rgacogne, line 326: I think we can merge `pdns::checked_stou` and `pdns::checked_stoi` based on the type, if we want to:
```c++
template <typename T>
auto checked(const std::string& str, size_t* idx = nullptr, int base = 10) -> T
{
  static_assert(std::numeric_limits<T>::is_integer, "checked: The `T` type must be an integer");

  if (str.empty()) {
    return 0; // compatibility                                                                                                                                                        
  }

  if constexpr (std::is_unsigned_v<T>) {
    return pdns::checked_conv<T>(std::stoull(str, idx, base));
  }
  else {
    return pdns::checked_conv<T>(std::stoll(str, idx, base));
  }
}
```

Did you consider writing a version of `pdns::checked_*` taking the destination as a parameter instead of returning it, so that we can avoid the duplication of the target name? Something like: `pdns::checked_stou_into(row[0], kd.id);`.
A macro would likely also work:
```c++
#define pdns_checked_stoi_into(str, dest) \
  dest = pdns::checked_stoi<decltype(dest)>(str)

#define pdns_checked_stou_into(str, dest) \
  dest = pdns::checked_stou<decltype(dest)>(str)
```

rgacogne (commented):
Nice work, I love the idea of making these conversions more strict! I'm wondering if we could not get rid of some of the duplication, see the comments below.

modules/bindbackend/binddnssec.cc
fredmorcos: I'm all for merging `pdns::checked_stou` into `pdns::checked_stoi` with the use of `constexpr` and already made the change. Thank you!

As for `checked_stoi_into`, that's also fine (I prefer it to be a function over a macro). However, I must mention that its usage breaks the flow of readability - it's very sad that the compiler is unable to infer the target type of `checked_stoi`. Example:

```c++
r.qname = DNSName(parts[1 + extraFields]);
r.qtype = parts[3 + extraFields];
pdns::checked_stoi_into(parts[4 + extraFields], r.ttl);
pdns::checked_stoi_into(parts[5 + extraFields], r.domain_id);
```

fredmorcos: Note: The first parameter of `checked_stoi_into` is the out parameter (the target).

rgacogne (approved):
Looks good, a couple suggestion but nothing that would prevent merging this PR!

pdns/backends/gsql/gsqlbackend.cc
```diff
@@ -315,7 +315,7 @@ bool GSQLBackend::getDomainInfo(const DNSName &domain, DomainInfo &di, bool getS
 
   ASSERT_ROW_COLUMNS("info-zone-query", d_result[0], 7);
 
-  di.id=pdns_stou(d_result[0][0]);
+  pdns::checked_stoi_into(di.id, d_result[0][0]);
```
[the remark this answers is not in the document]
fredmorcos: Could you explain more what you mean here? `di.id` is unsigned so I'm a bit confused.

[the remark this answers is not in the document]
rgacogne: You are not confused, I was! `domain_id` is signed in some places in the authoritative server and I thought that was the case here as well, but that's not true, so you can forget all about this comment. Sorry!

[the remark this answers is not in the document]
Habbie: It does mean that `di.id` really wants to stay in "signed positive" range, but when it doesn't, we're already in a ton of trouble.

PR #11402, opened 2022-03-09
WIP: docker: support cross compiling in Docker (rec)

wojas:
### Short description

Support cross compiling the recursor with native build host binaries in Docker.

This is not a full clean cross compilation environment. The build host must
have QEMU installed and handlers registered through binfmt_misc.

The build overwrites the target arch compiler and some other binaries with native
versions to speed the build up by avoiding emulation of the most CPU
intensive tools.

Target archs tested with amd64 build host:

- amd64 (same as build host): OK
- 386: error: size of time_t is 4, which is not large enough to fix the y2k38 bug (expected)
- arm/v7: error: size of time_t is 4, which is not large enough to fix the y2k38 bug (expected)
- arm/v6: not tested, expecting same time_t error
- arm64: OK
- mips64le: still building
- ppc64le: OK
- riscv64: no Debian 10 image
- s390x: still building

Progress:

- [x] Update rec Dockerfile
- [ ] Update dnsdist Dockerfile
- [ ] Update auth Dockerfile
- [ ] Update package builder dockerfiles
    - [ ] Figure out how to integrate this in package builds
    - [ ] This affects fakeroot, because it relies on `LD_PRELOAD`. How to fix this? Specify multiple libs? 
- [ ] Can the same trick work for 32 bit RPi?
- [ ] Check if the Dockerfiles still work with older docker versions without buildkit, and decide if we care if not
- [ ] Docker Hub build integration?

### Checklist
I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code


wojas:
This does not work without buildkit.

Build with `DOCKER_BUILDKIT=0` set:

```
Step 1/52 : FROM --platform=$BUILDPLATFORM debian:10-slim AS native-tools
failed to parse platform : "" is an invalid component of "": platform specifier component must match "^[A-Za-z0-9_-]+$": invalid argument
```

I would say that it's acceptable to require a recent Docker for builds. If this is not available, users can edit the Dockerfile to remove the `--platform` and pass `--build-arg NO_NATIVE=1` to the build.

Some more info on multiarch builds: https://www.docker.com/blog/faster-multi-platform-builds-dockerfile-cross-compilation-guide/

wojas:
Integration with the package builders requires more work:

- The current approach only works for Debian and probably Ubuntu.
- For RHEL based distributions we would need to figure out if there are equivalent packages and the locations of all the native binaries and libs to bundle.
- During package builds packages are installed. We may need to unpack the bundle of native binaries in the middle of the package build process to overwrite newly installed non-native binaries.
- We need to pass the extra `--host` argument to configure from an env var.
- The Dockerfiles are currently using per-arch Docker Hub repos instead of the newer multiarch manifests. Perhaps we need to add cross compilation support to the pdns-builder instead.
- The build infrastructure currently appears to assume we always build on amd64.
- fakeroot will not work, because it uses LD_PRELOAD. This is probably fine if we build as root.

rgacogne:
Is this PR still relevant?

romeroalx:
> Is this PR still relevant?

When we added `arm64` images, we discarded cross-compiling as was defined in this PR instead used Docker multi-platform builds without actually modifying the Dockerfiles: https://github.com/PowerDNS/pdns/pull/14171

Later,  we moved to building the images directly on a runner of the specific architecture and then creating the multi-platform image by joining them https://github.com/PowerDNS/pdns/pull/15996

This only applies to `amd64` and `arm64` but not to the other architectures listed in this PR.

Habbie:
> This only applies to `amd64` and `arm64` but not to the other architectures listed in this PR.

We don't ship the other architectures. I think we can close this PR.

PR #11409, opened 2022-03-11
auth-4.6.x API: fetch individual rrsets

Habbie:
### Short description
backport of #11389

### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/ws-auth.cc
+++ b/pdns/ws-auth.cc
@@ -348,7 +348,7 @@ static bool shouldDoRRSets(HttpRequest* req) {
   throw ApiException("'rrsets' request parameter value '"+req->getvars["rrsets"]+"' is not supported");
 }
 
-static void fillZone(UeberBackend& B, const DNSName& zonename, HttpResponse* resp, bool doRRSets) {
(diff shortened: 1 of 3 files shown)


pdns/ws-auth.cc
```diff
@@ -395,14 +395,24 @@ static void fillZone(UeberBackend& B, const DNSName& zonename, HttpResponse* res
...
+      if (req->getvars.count("rrset_name") == 0) {
+        di.backend->list(zonename, di.id, true); // incl. disabled
+      } else {
+        QType qt;
+        if (req->getvars.count("rrset_type") == 0) {
+          qt = QType::ANY;
+        } else {
+          qt = req->getvars["rrset_type"];
+        }
+        di.backend->lookup(qt, DNSName(req->getvars["rrset_name"]), di.id);
```
zeha, line 414: SQL backends will not return disabled RRs here, I think? Thats probably surprising for users.

Habbie: you're right. Filed as #11473

PR #11410, opened 2022-03-11
dnsdist: Define types for Lua arrays and associative arrays

rgacogne:
### Short description

That PR defines types for the mapping of Lua arrays and associative arrays to C++, improving readability and consistency:
```c++
template <class T> using LuaArray = std::vector<std::pair<int, T>>;
template <class T> using LuaAssociativeTable = std::unordered_map<std::string, T>;
template <class T> using LuaTypeOrArrayOf = boost::variant<T, LuaArray<T>>;
```

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdist-lua.hh
+++ b/pdns/dnsdist-lua.hh
@@ -129,13 +129,17 @@ private:
   uint32_t d_max{std::numeric_limits<uint32_t>::max()};
 };
 
-typedef boost::variant<string, vector<pair<int, string>>, std::shared_ptr<DNSRule>, DNSName, vector<pair<int, DNSName> > > luadnsrule_t;
+template <class T> using LuaArray = std::vector<std::pair<int, T>>;
+template <class T> using LuaAssociativeTable = std::unordered_map<std::string, T>;
+template <class T> using LuaTypeOrArrayOf = boost::variant<T, LuaArray<T>>;
+
+using luadnsrule_t = boost::variant<string, LuaArray<std::string>, std::shared_ptr<DNSRule>, DNSName, LuaArray<DNSName>>;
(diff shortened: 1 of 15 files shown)


pdns/dnsdist-lua.hh
```diff
@@ -129,14 +129,18 @@ private:
   uint32_t d_max{std::numeric_limits<uint32_t>::max()};
 };
 
-typedef boost::variant<string, vector<pair<int, string>>, std::shared_ptr<DNSRule>, DNSName, vector<pair<int, DNSName> > > luadnsrule_t;
+template <class T> using LuaTable = std::vector<std::pair<int, T>>;
+template <class T> using LuaAssociativeTable = std::unordered_map<std::string, T>;
+template <class T> using LuaTypeOrListOf = boost::variant<T, LuaTable<T>>;
```
Habbie, line 134: it might be clearer to rename `LuaTable` to `LuaList` or `LuaArray` - I like `LuaArray` (or vector?). Then, `LuaTypeOrListOf` should use the same word instead of `List`.

Habbie (changes requested):
One comment - let me know what you think before I review the rest for correctness :)

pdns/dnsdist-lua.hh
rgacogne: `LuaArray` sounds better indeed, and then `LuaTypeOrListOf` becomes `LuaTypeOrArrayOf`. I'll change that!

rgacogne: Done!

Habbie (approved):
Awesome work. Noticed some unrelated cleanups too, those also look good.

PR #11414, opened 2022-03-14
Import and export of ECDSA keys from/to PEM

fredmorcos:
### Short description
Add support for importing and exporting ECDSA keys from/to PEM

### Checklist

I have:

- [X] compiled this code
- [X] tested this code
- [X] documented the code

### DIFF
--- a/pdns/dnssecinfra.cc
+++ b/pdns/dnssecinfra.cc
@@ -168,6 +168,23 @@ std::unique_ptr<DNSCryptoKeyEngine> DNSCryptoKeyEngine::makeFromISCString(DNSKEY
   return dpk;
 }
 
+std::unique_ptr<DNSCryptoKeyEngine> DNSCryptoKeyEngine::makeFromPEMString(DNSKEYRecordContent& drc, const std::string& raw)
+{
+  for (const makers_t::value_type& maker : getMakers()) {
+    std::unique_ptr<DNSCryptoKeyEngine> ret = nullptr;
+
+    try {
+      ret = maker.second(maker.first);
+      ret->fromPEMString(drc, raw);
+      return ret;
+    }
+    catch (...) {
+    }
+  }
+
+  return nullptr;
+}
+
 std::string DNSCryptoKeyEngine::convertToISC() const
 {
   storvector_t storvector = this->convertToISCVector();
@@ -367,23 +384,6 @@ std::unique_ptr<DNSCryptoKeyEngine> DNSCryptoKeyEngine::makeFromPublicKeyString(
   auto dpk = make(algorithm);
   dpk->fromPublicKeyString(content);
   return dpk;
-}
-
-std::unique_ptr<DNSCryptoKeyEngine> DNSCryptoKeyEngine::makeFromPEMString(DNSKEYRecordContent& drc, const std::string& raw)
-{
-  for (const makers_t::value_type& val : getMakers()) {
-    std::unique_ptr<DNSCryptoKeyEngine> ret = nullptr;
-
-    try {
-      ret = val.second(val.first);
-      ret->fromPEMString(drc, raw);
-      return ret;
-    }
-    catch (...) {
-    }
-  }
-
-  return nullptr;
 }
 
 /**
--- a/pdns/dnssecinfra.cc
+++ b/pdns/dnssecinfra.cc
@@ -168,21 +168,11 @@ std::unique_ptr<DNSCryptoKeyEngine> DNSCryptoKeyEngine::makeFromISCString(DNSKEY
   return dpk;
 }
 
-std::unique_ptr<DNSCryptoKeyEngine> DNSCryptoKeyEngine::makeFromPEMString(DNSKEYRecordContent& drc, const std::string& raw)
+std::unique_ptr<DNSCryptoKeyEngine> DNSCryptoKeyEngine::makeFromPEMFile(DNSKEYRecordContent& drc, const std::string& filename, std::FILE& fp, const uint8_t algorithm)
 {
-  for (const makers_t::value_type& maker : getMakers()) {
-    std::unique_ptr<DNSCryptoKeyEngine> ret = nullptr;
-
-    try {
-      ret = maker.second(maker.first);
-      ret->fromPEMString(drc, raw);
-      return ret;
-    }
-    catch (...) {
-    }
-  }
-
-  return nullptr;
+  auto maker = DNSCryptoKeyEngine::make(algorithm);
+  maker->createFromPEMFile(drc, filename, fp);
+  return maker;
 }
 
 std::string DNSCryptoKeyEngine::convertToISC() const
--- a/pdns/dnssecinfra.cc
+++ b/pdns/dnssecinfra.cc
@@ -188,10 +188,11 @@ std::unique_ptr<DNSCryptoKeyEngine> DNSCryptoKeyEngine::make(unsigned int algo)
 {
   const makers_t& makers = getMakers();
   makers_t::const_iterator iter = makers.find(algo);
-  if (iter != makers.cend())
+  if (iter != makers.cend()) {
     return (iter->second)(algo);
+  }
   else {
-    throw runtime_error("Request to create key object for unknown algorithm number "+std::to_string(algo));
+    throw runtime_error("Request to create key object for unknown algorithm number " + std::to_string(algo));
   }
 }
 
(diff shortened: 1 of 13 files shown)


pdns/dnssecinfra.cc
```diff
@@ -168,13 +168,20 @@ std::unique_ptr<DNSCryptoKeyEngine> DNSCryptoKeyEngine::makeFromISCString(DNSKEY
   return dpk;
 }
 
+auto DNSCryptoKeyEngine::makeFromPEMFile(DNSKEYRecordContent& drc, const std::string& filename, std::FILE* fp, const uint8_t algorithm) -> std::unique_ptr<DNSCryptoKeyEngine>
```
rgacogne, line 171: Is there any benefit to the form you are using with regard to the more "usual" one?
```suggestion
std::unique_ptr<DNSCryptoKeyEngine> DNSCryptoKeyEngine::makeFromPEMFile(DNSKEYRecordContent& drc, const std::string& filename, std::FILE* fp, const uint8_t algorithm)
```

pdns/dnssecinfra.hh
```diff
@@ -37,55 +37,81 @@ class DNSCryptoKeyEngine
     explicit DNSCryptoKeyEngine(unsigned int algorithm) : d_algorithm(algorithm) {}
     virtual ~DNSCryptoKeyEngine() {};
     virtual string getName() const = 0;
-    
+
     typedef std::map<std::string, std::string> stormap_t;
     typedef std::vector<std::pair<std::string, std::string > > storvector_t;
     virtual void create(unsigned int bits)=0;
+    virtual auto createFromPEMFile(DNSKEYRecordContent& drc, const std::string& filename, std::FILE* fp) -> void
```
rgacogne, line 44: Perhaps a reference to a `std::FILE` would make more sense here, in `makeFromPEMFile` and in `convertToPEM``, since it cannot be `nullptr`?
```suggestion
    virtual auto createFromPEMFile(DNSKEYRecordContent& drc, const std::string& filename, std::FILE& fp) -> void
```

Habbie (changes requested):
Thanks! One inline comment, plus a request: can you add the new pdnsutil functions to `docs/manpages/pdnsutil.1.rst` and `contrib/pdnsutil.bash_completion.d`? You can then mark the PR with 'closes: #11321'.

Other than that, awesome work!

pdns/misc.cc
```diff
@@ -198,6 +198,30 @@ size_t writen2WithTimeout(int fd, const void * buffer, size_t len, const struct
   return len;
 }
 
+auto pdns::getMessageFromErrno(const int errnum) -> std::string
```
rgacogne, line 201: Perhaps ponder using this function in `stringerror`?

pdns/opensslsigners.cc
```diff
@@ -627,6 +660,47 @@ void OpenSSLECDSADNSCryptoKeyEngine::create(unsigned int bits)
...
+    throw runtime_error(getName() + " computing public key from private failed");
+  }
+
+  ret = EC_KEY_set_public_key(d_eckey.get(), pub_key.get());
+  if (ret != 1) {
+    ERR_print_errors_fp(stderr);
+    throw runtime_error(getName() + " setting public key failed");
+  }
+
+  EC_KEY_set_asn1_flag(d_eckey.get(), OPENSSL_EC_NAMED_CURVE);
```
rgacogne, line 694: Do you know when we actually need this? After reading https://wiki.openssl.org/index.php/Elliptic_Curve_Diffie_Hellman it's not clear to me if we need this only when we generated the key ourselves, when exporting it, or when loading it..

pdns/test-signers.cc
```diff
@@ -159,56 +209,73 @@ static void checkRR(const signerParams& signer)
...
+BOOST_AUTO_TEST_CASE(test_generic_signers)
+{
+  for (const auto& signer : signers) {
+    DNSKEYRecordContent drc;
+    auto dcke = std::shared_ptr<DNSCryptoKeyEngine>(DNSCryptoKeyEngine::makeFromISCString(drc, signer.iscMap));
+    test_generic_signer(dcke, drc, signer);
+
+    if (signer.pem.has_value()) {
+      unique_ptr<std::FILE, decltype(&std::fclose)> fp{fmemopen((void*)signer.pem->c_str(), signer.pem->length(), "r"), &std::fclose};
+      BOOST_CHECK(fp.get() != nullptr);
```
rgacogne, line 271:
```suggestion
      BOOST_REQUIRE(fp.get() != nullptr);
```

rgacogne (commented):
Nice work!

pdns/dynloader.cc
```diff
@@ -105,23 +106,22 @@ int main(int argc, char **argv)
...
-      D=shared_ptr<DynMessenger>(new DynMessenger(socketname));
+    if (::arg()["remote-address"].empty()) {
+      D = std::make_shared<DynMessenger>(socketname);
+    }
     else {
-      uint16_t port;
       try {
-        port = static_cast<uint16_t>(pdns_stou(::arg()["remote-port"]));
+        auto port = static_cast<uint16_t>(pdns_stou(::arg()["remote-port"]));
+        D = std::make_shared<DynMessenger>(ComboAddress(::arg()["remote-address"], port), ::arg()["secret"]);
```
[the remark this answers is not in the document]
fredmorcos: Right. I'll just delete this particular change altogether.

pdns/dnssecinfra.cc
fredmorcos: It's modern C++ at this point, and suggested by clang-tidy. IIRC the reason for it is due to an ambiguity with the old-style when combined with templates in certain cases.

I think we should just systematically move to the new-style, what do you think?

pdns/pdnsutil.cc
```diff
@@ -3344,7 +3343,7 @@ try
       cerr << std::to_string(id) << "Key was added, but there was a failure while returning the key id" << endl;
     }
     else {
-      cout << std::to_string(id) << endl;
+      cout << "Added key with ID " << std::to_string(id) << endl;
```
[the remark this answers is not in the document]
fredmorcos: I'd just delete this commit altogether.

pdns/misc.cc
```diff
@@ -198,6 +198,30 @@ size_t writen2WithTimeout(int fd, const void * buffer, size_t len, const struct
...
+{
+  const size_t errLen = 2048;
+  std::string errMsgData{};
+  errMsgData.reserve(errLen);
```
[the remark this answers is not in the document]
fredmorcos: Ah. `resize()` and `size()` are the correct methods to use here, I don't know why I went for `reserve()` which would leave other string metadata (like the size/length) incorrect.

pdns/misc.cc
fredmorcos: I would leave this sort of change up to a cleanup round at some point if that's OK?
We'd replace the use of `stringerror` with `getMessageFromErrno`.

pdns/opensslsigners.cc
fredmorcos: I don't actually know except that "the key isn't going to be very useful without it". And having the named OID is nice, it helped us notice an inconsistency when testing with `openssl asn1parse`.

Is it harmful here in any way?

pdns/misc.cc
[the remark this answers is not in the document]
fredmorcos: However, regarding avoiding the copy, I don't think that's possible. From `strerror_r(3)`:

> The GNU-specific strerror_r() returns a pointer to a string containing the error message.  This may be either a pointer to a string that the function stores in buf, or a  pointer  to  some
>        (immutable)  static  string (in which case buf is unused).  If the function stores a string in buf, then at most buflen bytes are stored (the string may be truncated if buflen is too small
>        and errnum is unknown).  The string always includes a terminating null byte ('\0').

pdns/dnssecinfra.cc
rgacogne: Personally I'm not convinced by the trailing return type syntax, except of course when it is actually needed. I'm also a bit worried that moving everything to the new style would require a lot of changes and make backporting fixes more painful. That being said I'm not strongly opposed to it, especially if there is some benefits that I'm not seeing yet :)

pdns/misc.cc
rgacogne: Sure, a separate PR for this at a later time works for me!

pdns/misc.cc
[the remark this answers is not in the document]
rgacogne: Right, I always forget that the GNU version tries to do the optimization for you.. I guess in the GNU case we could compare `errMsg` to `errMsgData.data()`, though, since we have already allocated a buffer, and `assign()` the content pointed to by `errMsg` if needed, but I'm not sure it is worth it.
It looks like the XSI case is not going to work if `res != 0`, though?

pdns/opensslsigners.cc
rgacogne: No, I'm not advocating to remove it here. I was instead wondering if we need to add it in other places, now that we know it exists.

pdns/misc.cc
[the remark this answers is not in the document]
fredmorcos: >  It looks like the XSI case is not going to work if res != 0, though?

Ah, I think you mean if `res == 0`?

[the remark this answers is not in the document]
rgacogne: Uh, yes, sorry!

pdns/dnssecinfra.cc
fredmorcos: Had a short discussion with @Habbie over lunch about this and we'll stick with the old-style for non-template functions. I'll also see if I can disable that particular diagnostic from clang-tidy.

rgacogne (approved):
Looks very good to me, great work!

PR #11417, opened 2022-03-15
Use RSA-2048 instead of RSA-1024 for crypto benchmark

nils-wisiol:
### Short description
This changes the crypto algorithm benchmark used by `pdnsutil test-algorithm` and `pdnsutil test-algorithms` to use RSA keys of 2048 bits instead of 1024 bits.

### Checklist
I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnssecinfra.cc
+++ b/pdns/dnssecinfra.cc
@@ -279,7 +279,7 @@ void DNSCryptoKeyEngine::testMakers(unsigned int algo, maker_t* creator, maker_t
   cout<<"Testing algorithm "<<algo<<"("<<DNSSECKeeper::algorithm2name(algo)<<"): '"<<dckeCreate->getName()<<"' ->'"<<dckeSign->getName()<<"' -> '"<<dckeVerify->getName()<<"' ";
   unsigned int bits;
   if(algo <= 10)
-    bits=1024;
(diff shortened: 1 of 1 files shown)


nils-wisiol:
One of the recursor tests fails with `AssertionError: False is not true : No AD flag found in the message for dns.google.`. I think that's a random Internet hickup and unrelated to the changes in this PR.

rgacogne:
> One of the recursor tests fails with `AssertionError: False is not true : No AD flag found in the message for dns.google.`. I think that's a random Internet hickup and unrelated to the changes in this PR.

Indeed:
```
  Mar 15 08:57:49 [1] dns.google: Trying IP 9.9.9.9:853, asking 'dns.google|A'
  Mar 15 08:57:49 [1] dns.google: using DoT with 9.9.9.9:853
  Mar 15 08:57:49 [1] dns.google: error resolving from 9.9.9.9 over TCP, possible error: Transport endpoint is not connected
  Mar 15 08:57:49 [1] dns.google: Failed to resolve via any of the 1 offered NS at level '.'
  Mar 15 08:57:49 [1] dns.google: failed (res=-1)
  ```

omoerbeek (approved):

PR #11376, opened 2022-02-28
rec: Be more careful using refresh mode only for the record asked.

omoerbeek:
Otherwise we get bad interaction with QM, as newly discovered
delegation points are stored in the cache, but not seen the QM
algorithm. Might/should fix #11371.

### Checklist

I have:
- [X] read the [CONTRIBUTING.md](https://github.com/PowerDNS/pdns/blob/master/CONTRIBUTING.md) document
- [X] compiled this code
- [X] tested this code
- [ ] included documentation (including possible behaviour changes)
- [ ] documented the code
- [ ] added or modified regression test(s)
- [ ] added or modified unit test(s)

PR #11419, opened 2022-03-15
rec: Backport 11376 to rec 4.5.x: Be more careful using refresh mode only for the record asked

omoerbeek:
### Short description

Backport of #11376 

### Checklist

I have:
- [X] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/syncres.cc
+++ b/pdns/syncres.cc
@@ -1086,6 +1086,7 @@ vector<ComboAddress> SyncRes::getAddrs(const DNSName &qname, unsigned int depth,
   d_DNSSECValidationRequested = false;
   d_followCNAME = true;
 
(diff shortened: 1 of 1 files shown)


rgacogne (commented):
I wonder if we need to save the value of `d_refresh` and set it to false before calling `doResolve()` in `SyncRes::getAddrs()`, then restore it after that?

omoerbeek:
> I wonder if we need to save the value of `d_refresh` and set it to false before calling `doResolve()` in `SyncRes::getAddrs()`, then restore it after that?

I think you are right. For the 4.6.x branch, we know the antries are not in the cache as we tested that earlier.

rgacogne (approved):

PR #11427, opened 2022-03-17
dnsdist: Implement Channels for communication between threads

rgacogne:
### Short description

Internally this uses the same mechanism as before, passing pointers over a pipe, but the new classes make that easier to use:
- clear separation between sender and receiver
- clear ownership of the descriptor
- less code duplication
- 
### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code


rgacogne:
Interestingly TSAN now thinks there is a race in the dnsdist code handling TC=1 responses received over UDP, for incoming queries received over DoH:
- the query is received in the DoH thread, sent to the backend over UDP (TSAN READ)
- the UDP responder thread picks up the response, notices it's truncated in `handleUDPResponseForDoH`, passes it to the DoH thread, which passes it to a TCP worker thread
- the worker thread prepend the size of the query in the query buffer, causing TSAN to detect a race (TSAN WRITE)

I don't think there is a race, unless I'm missing something, since the query has to have been sent by the time the corresponding response is received. It might take some annotation magic to convince TSAN, though.

rgacogne:
Sounds a lot like https://github.com/google/sanitizers/issues/1498
And indeed the actual range check is done after `send()` has returned: https://github.com/llvm/llvm-project/blob/main/compiler-rt/lib/sanitizer_common/sanitizer_common_interceptors.inc#L6982

rgacogne:
Disabling the `send` wrappers via `intercept_send=0` works around that issue. This is a bit sad but better than the alternatives, I guess.

rgacogne:
Marking this as "ready for review" to get feedback.

pdns/channel.hh
```diff
@@ -0,0 +1,329 @@
...
+  {
+    return receive(std::default_delete<T>());
+  }
+
+  template <typename T, typename D>
+  std::optional<std::unique_ptr<T, D>> Receiver<T, D>::receive(D deleter)
+  {
+    while (true) {
+      std::optional<std::unique_ptr<T, D>> result;
+      T* obj{nullptr};
```
omoerbeek, line 271:
```suggestion
      T* objPtr{nullptr};
```

pdns/channel.hh
```diff
@@ -0,0 +1,329 @@
...
+      else if (got == -1) {
+        if (errno == EINTR) {
+          continue;
+        }
+        if (errno == EAGAIN || errno == EWOULDBLOCK) {
+          return result;
+        }
+        throw std::runtime_error("Error while reading from Channel receiver: " + stringerror());
+      }
+      else {
```
omoerbeek, line 295: Does it make sense to test for write returning != writesize on the write side as well? It *should* never happen, but still...

omoerbeek (commented):
Two small comments, will do a full review soon.

pdns/channel.hh
rgacogne: I think we do, don't we?

omoerbeek: Ah right, just the zero case is not handled as a separate case. errno is not set if write returns zero, so to be pedantic the zero case should be handled separately,.

rgacogne: Agreed, I'll do that!

pdns/dnsdistdist/dnsdist-tcp-upstream.hh
```diff
@@ -3,6 +3,8 @@
 #include "dolog.hh"
 #include "dnsdist-tcp.hh"
 
+class TCPCrossProtocolResponse;
```
omoerbeek, line 6:
```suggestion
struct TCPCrossProtocolResponse;
```

pdns/channel.hh
```diff
@@ -0,0 +1,329 @@
...
+  template <typename T, typename D>
+  void Sender<T, D>::close()
+  {
+    d_fd.reset();
+  }
+
+  template <typename T, typename D>
+  std::optional<std::unique_ptr<T, D>> Receiver<T, D>::receive()
+  {
+    return receive(std::default_delete<T>());
```
omoerbeek, line 263:
```suggestion
    return receive(D());
```

omoerbeek (changes requested):

rgacogne:
Comments applied, thanks!

pdns/channel.hh
```diff
@@ -0,0 +1,332 @@
...
+
+  /**
+   * \brief Create a channel to pass objects between threads, accepting multiple senders and receivers.
+   *
+   * \return A pair of Sender and Receiver objects.
+   *
+   * \throw runtime_error if the channel creation failed.
+   */
+  template <typename T, typename D = std::default_delete<T>>
+  std::pair<Sender<T, D>, Receiver<T, D>> createObjectQueue(bool sendNonBlocking = true, bool writeNonBlocking = true, size_t pipeBufferSize = 0, bool throwOnEOF = true);
```
omoerbeek, line 141:
```suggestion
  std::pair<Sender<T, D>, Receiver<T, D>> createObjectQueue(bool sendNonBlocking = true, bool receiveNonBlocking = true, size_t pipeBufferSize = 0, bool throwOnEOF = true);
```

omoerbeek (approved):
Looks really nice, much more easier on the eye than the current code. Looking forward to using it in the recursor a well.

rgacogne:
Added unit tests.

rgacogne:
[d5eaeb3](https://github.com/PowerDNS/pdns/pull/11427/commits/d5eaeb3337d0e98db6e1548a4a9ab607082c15fd) broke the auth build, so I will have to squash the corresponding fixup commit [e8dca71](https://github.com/PowerDNS/pdns/pull/11427/commits/e8dca71d79009aaf972264b07ef290547aa8b850) before merging this PR.

rgacogne:
Rebased to fix conflicts.

rgacogne:
Otto, Fred, how would you feel about merging this PR?

fredmorcos:
> Otto, Fred, how would you feel about merging this PR?

It's been on my todo for a while. I need to go through it still. I'll do that first thing on Monday.

fredmorcos (approved):
Just a few nitpicks, nothing important. Generally I really like this and I guess you know why :-D. It did clean up quite a bit.

Also please wait for a review from @omoerbeek as well.

pdns/doh.hh
```diff
@@ -247,6 +250,7 @@ struct DOHUnit
   st_h2o_req_t* req{nullptr};
   DOHUnit** self{nullptr};
   DOHServerConfig* dsc{nullptr};
+  pdns::channel::Sender<DOHUnit, void(*)(DOHUnit*)>* responseSender{nullptr};
```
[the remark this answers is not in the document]
rgacogne: Unfortunately it wouldn't, because we have a non-static `release()` member function and a static `release(DOHUnit*)` one, so the compiler cannot decide which one we are referring to.

pdns/channel.cc
```diff
@@ -0,0 +1,115 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+
+#include "channel.hh"
+
+namespace pdns::channel
```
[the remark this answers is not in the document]
rgacogne: That's weird because the CI's `clang-tidy` actually complained about the initial `namespace pdns { namespace channel`, leading me to switch to the new form in https://github.com/PowerDNS/pdns/pull/11427/commits/35b27ac8d18b43702a63e4078b68776c66f561ab : https://clang.llvm.org/extra/clang-tidy/checks/modernize/concat-nested-namespaces.html

omoerbeek (changes requested):
This is really nice! 
A thing: clang-tidy does report some issues (as shown in the build logs), but these are not detected in the summary task. Wonder why....

Some of these issues might indicate a real problem like `d_fd` not being initialised in all cases, reported in https://github.com/PowerDNS/pdns/actions/runs/5253429184/jobs/9490749878 dnsdist build log.

pdns/channel.cc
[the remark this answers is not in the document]
fredmorcos: Ah. I think I figured out why, creating a git worktree with magit - which I usually do to review PRs - doesn't copy over ignored files. So my clang-tidy was complaining using the default set of configuration, in this particular case saying that nested namespace declarations are a C++17 extension.

pdns/channel.cc
```diff
@@ -0,0 +1,115 @@
...
+{
+
+Notifier::Notifier(FDWrapper&& fd) :
```
fredmorcos, line 28: clang-tidy should complain about the identifier being too short here...

rgacogne:
> A thing: clang-tidy does report some issues (as shown in the build logs), but these are not detected in the summary task. Wonder why....
> 
> Some of these issues might indicate a real problem like `d_fd` not being initialised in all cases, reported in https://github.com/PowerDNS/pdns/actions/runs/5253429184/jobs/9490749878 dnsdist build log.

I'll have a second look tomorrow but at the moment I cannot make sense of this warning, as far as I can tell `d_fd` cannot be uninitialized.

pdns/channel.cc
rgacogne: Fixed!

rgacogne:
> > A thing: clang-tidy does report some issues (as shown in the build logs), but these are not detected in the summary task. Wonder why....
> > Some of these issues might indicate a real problem like `d_fd` not being initialised in all cases, reported in https://github.com/PowerDNS/pdns/actions/runs/5253429184/jobs/9490749878 dnsdist build log.
> 
> I'll have a second look tomorrow but at the moment I cannot make sense of this warning, as far as I can tell `d_fd` cannot be uninitialized.

I still don't see it :-/ Otto, Fred, did you look at the reported issue or just at the report?

omoerbeek:
> I still don't see it :-/ Otto, Fred, did you look at the reported issue or just at the report

I did not look at the details of the report, doing that now.

omoerbeek:
> I did not look at the details of the report, doing that now.

I cannot spot the issue either. It seems to think that the default ct of `Receiver` does not init `d_fd`, but `d_fd` is of type `FDWrapper`, which itself has a default  value for it's only field `d_fd` of type `int`.

omoerbeek (approved):

PR #11432, opened 2022-03-18
Enable named curves on ECDSA key generation and import from ISC

fredmorcos:
### Short description
Enable named curves on ECDSA key generation and import from ISC

### Checklist

I have:

- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/opensslsigners.cc
+++ b/pdns/opensslsigners.cc
@@ -658,6 +658,8 @@ void OpenSSLECDSADNSCryptoKeyEngine::create(unsigned int bits)
   if (res == 0) {
     throw runtime_error(getName()+" key generation failed");
   }
+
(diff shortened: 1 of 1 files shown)


Habbie (approved):

rgacogne:
I was not fast enough but the documentation states that:
> the OPENSSL_EC_NAMED_CURVE value was added in OpenSSL 1.1.0; previous versions should use 0 instead.

So we might need to check whether that value is properly defined before using it :-/ Well unless we do not need to care about OpenSSL < 1.1.0 which would be good, but 1.0.2 is still supported by OpenSSL for paying customers if I'm not mistaken.

Habbie:
EL7, which this fix is for, uses `1.0.2k`, where this fix does in fact help..

rgacogne:
The documentation is lying, `OPENSSL_EC_NAMED_CURVE` is already there in 0.9.8. Carry on.

PR #11438, opened 2022-03-21
Checked int conversions: Avoid checks when boundary values are equal

fredmorcos:
### Short description
We can avoid extraneous checks when the min or max values of different int types are equal.

### Checklist

I have:

- [X] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/misc.hh
+++ b/pdns/misc.hh
@@ -656,8 +656,20 @@ auto checked_conv(F from) -> T
   static_assert((std::numeric_limits<F>::is_signed && std::numeric_limits<T>::is_signed) || (!std::numeric_limits<F>::is_signed && !std::numeric_limits<T>::is_signed),
                 "checked_conv: The `T` and `F` types must either both be signed or unsigned");
 
-  if (from < std::numeric_limits<T>::min() || from > std::numeric_limits<T>::max()) {
-    throw std::out_of_range("checked_conv: conversion from value that is out of range for target type");
+  constexpr auto tMin = std::numeric_limits<T>::min();
+  if constexpr (std::numeric_limits<F>::min() != tMin) {
+    if (from < tMin) {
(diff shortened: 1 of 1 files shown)


pdns/misc.hh
```diff
@@ -656,8 +657,20 @@ auto checked_conv(F from) -> T
   static_assert((std::numeric_limits<F>::is_signed && std::numeric_limits<T>::is_signed) || (!std::numeric_limits<F>::is_signed && !std::numeric_limits<T>::is_signed),
                 "checked_conv: The `T` and `F` types must either both be signed or unsigned");
 
-  if (from < std::numeric_limits<T>::min() || from > std::numeric_limits<T>::max()) {
-    throw std::out_of_range("checked_conv: conversion from value that is out of range for target type");
+  constexpr auto tMin = std::numeric_limits<T>::min();
+  if constexpr (std::numeric_limits<F>::min() != tMin) {
+    if (from < tMin) {
+      string s = boost::str(boost::format("checked_conv: source value %1% is smaller than target's minimum possible value %2%") % from % tMin);
```
rgacogne, line 663: I would prefer not using `boost::format`here since we have had trouble with it calling `std:::ctype::narrow` which turned up not being thread-safe on some implementations, if I recall correctly.

fredmorcos: I switched the use to `std::to_string`. I force-pushed because it's small enough and can be merged immediately if you agree with it.

pdns/misc.hh
```diff
@@ -20,6 +20,7 @@
  * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
  */
 #pragma once
+#include <boost/format.hpp>
```
rgacogne, line 23: I believe this can go now :)

rgacogne (approved):
Looks good to me, thanks! One minor nit that will not invalidate this review :+1:

PR #11444, opened 2022-03-23
rec: Use nice format for timestamp printing

omoerbeek:
### Short description

This changes to timestamp printing to only print relevant fields: year and day of week are not useful in this context. Also make the  sortable if needed using e.g. `... - | sort -k3`

Compare:
```
; edns from thread follows
;
192.5.6.30      1       Wed Mar 23 13:34:38 2022
192.31.80.30    1       Wed Mar 23 13:34:38 2022
192.42.93.30    1       Wed Mar 23 13:34:39 2022
192.52.178.30   1       Wed Mar 23 13:34:39 2022
```
to
```
; edns from thread follows
;
192.5.6.30      1       03-23T13:33:36
192.55.83.30    1       03-23T13:33:36
192.42.93.30    1       03-23T13:33:36
192.41.162.30   1       03-23T13:33:37
```
### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/syncres.cc
+++ b/pdns/syncres.cc
@@ -713,6 +713,47 @@ bool SyncRes::isForwardOrAuth(const DNSName &qname) const {
   return iter != t_sstorage.domainmap->end() && (iter->second.isAuth() || !iter->second.shouldRecurse());
 }
 
+# if 0
+// Will be needed in the future
+static const char* timestamp(const struct timeval& tv, char* buf, size_t sz)
+{
+  const std::string s_timestampFormat = "%Y-%m-%dT%T";
+  struct tm tm;
+  size_t len = strftime(buf, sz, s_timestampFormat.c_str(), localtime_r(&tv.tv_sec, &tm));
+  if (len == 0) {
+    int ret = snprintf(buf, sz, "%lld", static_cast<long long>(tv.tv_sec));
+    if (ret < 0 || static_cast<size_t>(ret) >= sz) {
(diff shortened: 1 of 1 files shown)


rgacogne:
I wonder if it would make sense to keep the year even though it's not really useful in that context, just because people (and tools?) are used to seeing it?

omoerbeek:
> I wonder if it would make sense to keep the year even though it's not really useful in that context, just because people (and tools?) are used to seeing it?

Fair enough, I'll push a fix to include the year so it becomes ISO 8601

pdns/syncres.cc
```diff
@@ -713,6 +713,35 @@ bool SyncRes::isForwardOrAuth(const DNSName &qname) const {
...
+}
+#endif
+
+static const char* timestamp(time_t t, char* buf, size_t sz)
+{
+  const std::string s_timestampFormat = "%Y-%m-%dT%T";
+  struct tm tm;
+  size_t len = strftime(buf, sz, s_timestampFormat.c_str(), localtime_r(&t, &tm));
+  if (len == 0) {
+    snprintf(buf, sz, "%lld", static_cast<long long>(t));
```
rgacogne, line 740: `snprintf` should never fail as long as long as a long long fit in 25 characters, but how about actually checking the return value and setting `buf[0] = '\0';` if it does just to be sure?

rgacogne (approved):

PR #11447, opened 2022-03-25
Add linguist-language tag for dns zones

jsoref:
### Short description

On the right-hand side of:
https://github.com/powerdns/pdns/
you can see a chart like this:
Language statistics listing: DIGITAL Command Language 5.1%

This can be tuned with:
https://github.com/github/linguist/blob/master/docs/overrides.md#using-gitattributes

With this change, the matching files will be treated as `DNS Zone`, by default the are not [detectable](https://github.com/github/linguist/blob/master/docs/overrides.md#detectable) and thus will fall out of the Languages chart:

This change is currently live on https://github.com/jsoref/pdns:
image

### Checklist

I have:
- [ ] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+# https://github.com/github/linguist/blob/master/docs/overrides.md
+*/zones/* linguist-language=DNS-Zone


Habbie:
Any idea why your version does not mention DNS Zones in the breakdown?

jsoref:
Yes, it's considered data and not a programming language and thus excluded by default. 

I could add `linguist-detectable` to the line and then it'd appear in the view.

Habbie:
Ah, understood. That's not necessary. Thanks!

Habbie (approved):

PR #11466, opened 2022-03-25
ixfr: Fix a case where an incomplete read caused by network error might result in a truncated zone

omoerbeek:
As we might break from the loop early, we need to check if the end SOA was seen after the loop.
Also make sure we detect end conditions for both AXFR and IXFR style properly, to avoid processing
data after the end marker.

This is PowerDNS Security Advisory 2022-01

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code


Habbie:
rebased to make CI run

Habbie:
The tests were missing, I put them back in.

rgacogne (approved):

PR #11470, opened 2022-03-28
Rec: scanbuild fixes

omoerbeek:
### Short description

No real bug uncovered, but some dead code eliminated, an `assert` added and a declaration made `const`. All sensible things to do. Two cases that are false positives (AFAIKS) remain.

### Checklist

I have:
- [X] compiled this code
- [ ] tested this code
- [ ] documented the code


rgacogne (approved):

omoerbeek:
Note to self: the mem leak disappears if I switch from `boost::function` to `std::function` for `broadcastAccfunction`

PR #11476, opened 2022-03-29
Switch from boost::function to std::function everywhere

omoerbeek:
### Short description

One more step in the general rule: if a facility is available in `std`, stop using the `boost` equivalent.
The only remaining ref to `boost::function` is now in `yahttp`, but that is conditional on C++11, so we're clean.

As a side effect this fixes a memory leak reported by scanbuild when using a function object in `rec-main.cc`.
The memory leak itself seems to be deep inside the boost code.

This touches `mtasker`, so it is a bit scary. I did not test the `ucontext` case (and I wonder if we should keep on supporting it).

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/mtasker.hh
+++ b/pdns/mtasker.hh
@@ -32,7 +32,7 @@
 #include "misc.hh"
 #include "mtasker_context.hh"
 #include <memory>
-#include <boost/function.hpp>
+
 using namespace ::boost::multi_index;
 
 // #define MTASKERTIMING 1
@@ -56,7 +56,7 @@ private:
   struct ThreadInfo
   {
 	std::shared_ptr<pdns_ucontext_t> context;
-	boost::function<void(void)> start;
+	std::function<void(void)> start;
 	char* startOfStack;
 	char* highestStackSeen;
 #ifdef MTASKERTIMING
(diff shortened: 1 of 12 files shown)


rgacogne (approved):
> As a side effect this fixes a memory leak reported by scanbuild when using a function object in rec-main.cc.
> The memory leak itself seems to be deep inside the boost code.

That's likely a false positive, right? Or did you manage to see that leak happening, in which case it would be nice to report it.

> This touches mtasker, so it is a bit scary. I did not test the ucontext case (and I wonder if we should keep on supporting it).

The changes do not look scary to me. I actually tested the ucontext case and saw no issues.
I think it makes sense to keep supporting the ucontext case because we are still used on architectures were boost::context is not supported and the maintenance cost does not feel that high, but you know better than I do :)

omoerbeek:
I'll check with valgrind if it sees a leak.

omoerbeek:
A test run calling wipe-cache:
```
==13633== HEAP SUMMARY:
==13633==     in use at exit: 0 bytes in 0 blocks
==13633==   total heap usage: 49,781 allocs, 49,781 frees, 18,907,037 bytes allocated
==13633== 
==13633== All heap blocks were freed -- no leaks are possible
```

PR #11485, opened 2022-03-30
Add plumbing for the upcoming code formatting

fredmorcos:
### Short description
This adds a few things (and more are yet to come) for the upcoming code formatting cleanup we are going to do.

1. A `clang-tidy` config file to be able to autofix missing block braces.
2. A new `clang-format` config file.
3. A new `.not-formatted` file which exlcudes vendored dependencies.
4. A script to find the least heavily modified files that are unformatted.

For now this PR is just a place to iron out the discussion on how to move forward with our code-formatting.

### Checklist

I have:

- [ ] compiled this code
- [X] tested this code
- [ ] documented the code


.not-formatted
```diff
@@ -310,11 +320,14 @@
...
 ./pdns/test-lua_auth4_cc.cc
 ./pdns/test-misc_hh.cc
+./pdns/test-mplexer.cc
 ./pdns/test-nameserver_cc.cc
 ./pdns/test-packetcache_cc.cc
 ./pdns/test-packetcache_hh.cc
 ./pdns/test-proxy_protocol_cc.cc
 ./pdns/test-rcpgenerator_cc.cc
+./pdns/test-recpacketcache_cc.cc
+./pdns/testrunner.cc
```
jsoref, line 330: This almost looked sorted until this line.

.not-formatted
```diff
@@ -1,8 +1,11 @@
-./ext/lmdb-safe/lmdb-safe.cc
```
omoerbeek, line 1: `make format-code` tries to format these vendor files if you remove these `ext` lines...

omoerbeek:
> Why are you adding files to `.not-formatted`? Files that are not in there should be formatted already. e.g. the files in pdns/recursordist. The unwritten rule is that all the files in that dir are formatted.

I see now that newer versions of `clang-format` indeed find these file to be not-formatted. I still think it's not a good idea to add names to `.not-formatted` as the files not in there were/are formatted, even if newer clang-format does not 100% agree.

.clang-format.full
```diff
@@ -0,0 +1,115 @@
...
+  AfterControlStatement: false
+  AfterEnum: true
+  AfterExternBlock: true
+  AfterFunction: true
+  AfterNamespace: true
+  AfterObjCDeclaration: true
+  AfterStruct: true
+  AfterUnion: true
+  BeforeCatch: false
+  BeforeElse: false
```
rgacogne, line 50: Most of our current code conforms to these, I believe:
```suggestion
  BeforeCatch: true
  BeforeElse: true
```

.clang-format.full
```diff
@@ -0,0 +1,115 @@
+---
+BasedOnStyle: LLVM
+Language: Cpp
+Standard: Auto
+
+IndentWidth: 2
+UseTab: Never
+ColumnLimit: 100
```
rgacogne, line 8: I'm not a huge fan of hard limits on line sizes, but I could probably live with it.

.clang-format.full
```diff
@@ -0,0 +1,115 @@
...
+IndentCaseLabels: true
+IndentPPDirectives: None
+IndentWrappedFunctionNames: false
+KeepEmptyLinesAtTheStartOfBlocks: false
+NamespaceIndentation: All
+SpaceAfterCStyleCast: true
+SpaceAfterLogicalNot: false
+SpaceAfterTemplateKeyword: true
+SpaceBeforeAssignmentOperators: true
+SpaceBeforeCpp11BracedList: true
```
rgacogne, line 102: I have a slight preference for `false`

.clang-format.full
rgacogne: And I would not mind a `BeforeWhile: true`

rgacogne (commented):
I have formatted a few random files with the .clang-format.full, and apart from a few settings for which I suggested changes I was quite happy with the results!
I think we need to ensure that everyone is onboard with the settings, and I think I agree with Otto that I would prefer not regressing regarding which files are formatted, but overall that's a very nice job, thanks a lot!

.not-formatted
fredmorcos: I see. I keep forgetting to look at `make format-code` because I don't use it at all.

However, do vendored files really fall under our responsibility? Doesn't formatting them just make it more work everytime we want to update the dependencies?

I guess my question is about whether we should disable the formatting of vendored files in `make format-code`.

fredmorcos:
> > Why are you adding files to `.not-formatted`? Files that are not in there should be formatted already. e.g. the files in pdns/recursordist. The unwritten rule is that all the files in that dir are formatted.
> 
> I see now that newer versions of `clang-format` indeed find these file to be not-formatted. I still think it's not a good idea to add names to `.not-formatted` as the files not in there were/are formatted, even if newer clang-format does not 100% agree.

I'm not sure I like this approach because it means that we will have to start ignoring clang-format (in CI as well), so it would be a step back. Or am I misunderstanding something?

.clang-format.full
fredmorcos: Right. I did notice that and I changed it on purpose. I think `else` and `catch` should belong to the corresponding `if` and `try` in cases of multiple consecutive blocks. It's a very subjective thing though.

Keeping it as it is of course simplifies our formatting passes and the issues that will come because of it.

omoerbeek:
My idea is: file not in `.clang-formatted` are formatted. Maybe not perfectly according to the newest clang-format version and settings, but good enough.
My worry is that if we add files to already-formatted, they will tend to become badly formatted over time.
So a run to format them according to the (to be introduced) formatting settings is in order instead of adding them to `.not-formatted`.

.clang-format.full
fredmorcos: I prefer it to be able to tile and see multiple files next to each other without the horrors of line-wrapping.

fredmorcos:
> So a run to format them according to the (to be introduced) formatting settings is in order instead of adding them to `.not-formatted`.

Alright, that's the part I was missing. I think at some point it was mentioned that huge formatting commits are not something we want and that's why I worked with the assumption that doing this would be out of the question.

omoerbeek:
> > So a run to format them according to the (to be introduced) formatting settings is in order instead of adding them to `.not-formatted`.
> 
> Alright, that's the part I was missing. I think at some point it was mentioned that huge formatting commits are not something we want and that's why I worked with the assumption that doing this would be out of the question.

The assumption is that they only need minor changes, as they are already formatted,

PR #11487, opened 2022-03-30
Rec: probe auth for DoT support

omoerbeek:
### Short description

Creating this PR so other can play wit this. I don't think this is production ready!

This implements an experimental approach to probe nameservers for DoT support and start using DoT to those if it works. Probing is done by a separate task. Until that task comes with a conclusion, the IP will be contacted in the regular way.

Enable by setting `max-busy-dot-probes` to a value > 0. Default is 0.

The `max-busy-dot-probes` number says how many async tasks can be scheduled doing DoT probes concurrently.
The tasks have to compete with other tasks to be done, so it may take a while for it to be executed. Until then the map entry is in state `Busy`. If the task sees it supports DoT, the entry becomes `Good`, otherwise `Bad`.
If there are too may Busy tasks, no new task will be scheduled. This is a crude mechanism, but it should have the effect that the most popular IPs will have a higher chance of getting a probe scheduled. After a while the status table will be filled and a new probe will only be scheduled if a unknown IP is encountered.

Check status by doing `rec-control dump-dot-probe-map - ` e.g.:
```
$ rec-control dump-dot-probe-map - | grep Good
188.166.104.87  powerdns.com.   Good    2022-04-02T15:48:00
185.89.218.12   instagram.com.  Good    2022-04-02T15:48:30
66.111.48.12    whatsapp.com.   Good    2022-04-02T15:51:20
66.111.49.12    whatsapp.com.   Good    2022-04-02T15:52:14
129.134.30.12   instagram.com.  Good    2022-04-02T15:53:00
185.89.219.12   facebook.com.   Good    2022-04-02T15:53:01
```
Note that key is IP and the name printed is the auth domain that triggered the probe. The IP can be authoritative for multiple (other) auth domains. The timestamp is the ttd (time to die) of the entry. Good entries will be kept for three days after last successful use. Bad entries live for 1 day.

This uses some ideas and timeout values from https://datatracker.ietf.org/doc/draft-dkgjsal-dprive-unilateral-probing/ but also leaves many things out. e.g. persistence of the knowledge gained.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/rec_control.cc
+++ b/pdns/rec_control.cc
@@ -96,7 +96,8 @@ int main(int argc, char** argv)
     "dump-rpz",
     "dump-throttlemap",
     "dump-non-resolving",
-    "dump-saved-parent-ns-sets"};
+    "dump-saved-parent-ns-sets",
+    "dump-dot-probe-map"};
   try {
     initArguments(argc, argv);
     string sockname = "pdns_recursor";
--- a/pdns/rec_control.cc
+++ b/pdns/rec_control.cc
@@ -97,7 +97,8 @@ int main(int argc, char** argv)
     "dump-throttlemap",
     "dump-non-resolving",
     "dump-saved-parent-ns-sets",
-    "dump-dot-probe-map"};
+    "dump-dot-probe-map",
+  };
   try {
     initArguments(argc, argv);
     string sockname = "pdns_recursor";
--- a/pdns/recursordist/docs/manpages/rec_control.1.rst
+++ b/pdns/recursordist/docs/manpages/rec_control.1.rst
@@ -92,6 +92,9 @@ dump-cache *FILENAME*
     also dumped to the same file. The per-thread positive and negative cache
     dumps are separated with an appropriate comment.
 
+dump-dot-probe-map *FILENAME*
+    Dump the contents of the DoT probe map to the *FILENAME* mentioned.
+
 dump-edns *FILENAME*
     Dumps the EDNS status to the filename mentioned. This file should not exist
     already, PowerDNS will refuse to overwrite it. While dumping, the recursor
--- a/pdns/recursordist/docs/settings.rst
+++ b/pdns/recursordist/docs/settings.rst
@@ -1066,6 +1066,30 @@ Path to a lua file to manipulate the Recursor's answers. See :doc:`lua-scripting
 The interval between calls to the Lua user defined `maintenance()` function in seconds.
 See :ref:`hooks-maintenance-callback`
 
+.. _setting-max-busy-dot-probes:
+
+``max-busy-dot-probes``
+-----------------------
+.. versionadded:: 4.7.0
+
+- Integer
+- Default: 0
+
+Limit the maxmium number of simultaneous DoT probes the Recursor will schedule.
+The default value 0 means no DoT probes are scheduled.
+
+DoT probes are used to check if an authoritative server's IP address supports DoT.
+If the probe determines an IP address supports DoT, the Recursor will use DoT to contact it for subsequent queries.
+The results of probes are remembered and can be viewed by the ``rec_control dump-dot-probe-map`` command.
+If the maximum number of pending probes is reached, no probe wil be scheduled, even if no DoT status is known for an address.
+If the result of a probe is not yet available, the Recursor will contact the authoritative server in the regular way,
+unless an authoritative server is configured to be contacted over DoT always using :ref:`setting-dot-to-auth-names`.
+In that case no probe will be scheduled.
+
+Note::
+  DoT probing is an experimental feature.
+  Please test thoroughly if it is suitable in your specific production environment before enabling.
+
 .. _setting-max-cache-bogus-ttl:
 
 ``max-cache-bogus-ttl``
--- a/pdns/recursordist/docs/settings.rst
+++ b/pdns/recursordist/docs/settings.rst
@@ -1081,7 +1081,7 @@ The default value 0 means no DoT probes are scheduled.
 DoT probes are used to check if an authoritative server's IP address supports DoT.
 If the probe determines an IP address supports DoT, the Recursor will use DoT to contact it for subsequent queries.
 The results of probes are remembered and can be viewed by the ``rec_control dump-dot-probe-map`` command.
-If the maximum number of pending probes is reached, no probe wil be scheduled, even if no DoT status is known for an address.
+If the maximum number of pending probes is reached, no probes will be scheduled, even if no DoT status is known for an address.
 If the result of a probe is not yet available, the Recursor will contact the authoritative server in the regular way,
 unless an authoritative server is configured to be contacted over DoT always using :ref:`setting-dot-to-auth-names`.
 In that case no probe will be scheduled.
--- a/pdns/recursordist/docs/settings.rst
+++ b/pdns/recursordist/docs/settings.rst
@@ -1089,7 +1089,7 @@ In that case no probe will be scheduled.
 
 Note::
   DoT probing is an experimental feature.
-  Please test thoroughly if it is suitable in your specific production environment before enabling.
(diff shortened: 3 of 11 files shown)


pdns/rec_control.cc
```diff
@@ -96,7 +96,8 @@ int main(int argc, char** argv)
     "dump-rpz",
     "dump-throttlemap",
     "dump-non-resolving",
-    "dump-saved-parent-ns-sets"};
+    "dump-saved-parent-ns-sets",
+    "dump-dot-probe-map"};
```
jsoref, line 100: Don't suppose there would be a way to have a trailing comma (maybe a "" entry?) so that blame wouldn't need to keep shifting.

pdns/recursordist/docs/settings.rst
```diff
@@ -1066,6 +1066,30 @@ Path to a lua file to manipulate the Recursor's answers. See :doc:`lua-scripting
...
+- Integer
+- Default: 0
+
+Limit the maximum number of simultaneous DoT probes the Recursor will schedule.
+The default value 0 means no DoT probes are scheduled.
+
+DoT probes are used to check if an authoritative server's IP address supports DoT.
+If the probe determines an IP address supports DoT, the Recursor will use DoT to contact it for subsequent queries.
+The results of probes are remembered and can be viewed by the ``rec_control dump-dot-probe-map`` command.
+If the maximum number of pending probes is reached, no probe wil be scheduled, even if no DoT status is known for an address.
```
jsoref, line 1084:
```suggestion
If the maximum number of pending probes is reached, no probes will be scheduled, even if no DoT status is known for an address.
```

pdns/recursordist/rec-taskqueue.cc
```diff
@@ -163,6 +163,65 @@ static void resolve(const struct timeval& now, bool logErrors, const pdns::Resol
...
+      log->error(Logr::Error, msg, e.reason);
+    }
+  }
+  catch (const PolicyHitException& e) {
+    if (logErrors) {
+      log->error(Logr::Notice, msg, "PolicyHit");
+    }
+  }
+  catch (...) {
+    log->error(Logr::Error, msg, "Unexpectec exception");
```
jsoref, line 197:
```suggestion
    log->error(Logr::Error, msg, "Unexpected exception");
```

pdns/recursordist/taskqueue.hh
```diff
@@ -43,15 +50,23 @@ struct ResolveTask
 {
   DNSName d_qname;
   uint16_t d_qtype;
+  // Deadline is not part of index and <
```
jsoref, line 53: `and <`??

omoerbeek: `operator<`(), used when comparing `ResolveTask`s

pdns/syncres.cc
```diff
@@ -4611,6 +4698,85 @@ bool SyncRes::processRecords(const std::string& prefix, const DNSName& qname, co
...
+
+bool SyncRes::tryDoT(const DNSName& qname, const QType qtype, const DNSName& nsName, ComboAddress address, time_t now)
+{
+  LWResult lwr;
+  bool truncated;
+  bool spoofed;
+  boost::optional<Netmask> nm;
+  address.setPort(853);
+  // We use the fact that qname equals auth
+  bool ok = doResolveAtThisIP("", qname, qtype, lwr, nm, qname, false, false, nsName, address, true, true, truncated, spoofed);
```
rgacogne, line 4773: It looks like we will not update the status of that entry if `doResolveAtThisIP` throws, and that would mean that we keep an entry in a `Busy` state, still counted in `d_numBusy` as it should, but we do not have a task for it anymore?

pdns/recursordist/docs/settings.rst
```diff
@@ -1066,6 +1066,29 @@ Path to a lua file to manipulate the Recursor's answers. See :doc:`lua-scripting
...
+.. versionadded:: 4.7.0
+
+- Integer
+- Default: 0
+
+Limit the maximum number of simultaneous DoT probes the Recursor will schedule.
+The default value 0 means no DoT probes are scheduled.
+
+DoT probes are used to check if an authoritative server's IP address supports DoT.
+If the probe determines an IP address supports DoT, the Recursor will use DoT to contact it for subsequent queries.
```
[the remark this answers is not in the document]
omoerbeek: right, we will fall back to Do53 and when the record expires (1 day)  try DoT again. This is one part that could/should be improved.

pdns/recursordist/rec-taskqueue.cc
```diff
@@ -163,6 +163,65 @@ static void resolve(const struct timeval& now, bool logErrors, const pdns::Resol
...
+  }
+  else {
+    ++s_resolve_tasks.run;
+  }
+}
+
+void runTasks(size_t max, bool logErrors)
+{
+  for (size_t count = 0; count < max; count++) {
+    pdns::ResolveTask task;
```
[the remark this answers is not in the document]
omoerbeek: yes, it does require `runTaskOnce` to return if a task was found though. Wil do.

omoerbeek:
I also see an unwanted interference with the throttle map. Will need to fix that.

pdns/recursordist/docs/manpages/rec_control.1.rst
```diff
@@ -92,6 +92,9 @@ dump-cache *FILENAME*
     also dumped to the same file. The per-thread positive and negative cache
     dumps are separated with an appropriate comment.
 
+dump-dot-probe-map *FILENAME*
+    Dump the contents of the DoT probe map to the *FILENAME* mentioned.
+
 dump-edns *FILENAME*
     Dumps the EDNS status to the filename mentioned. This file should not exist
```
jsoref, line 99: Not your code, but it probably should use `*FILENAME*` as you did.

I'm not a fan of _mentioned_, I'd use `specified FILENAME `, but it appears to be house style.

pdns/recursordist/docs/settings.rst
```diff
@@ -1066,6 +1066,31 @@ Path to a lua file to manipulate the Recursor's answers. See :doc:`lua-scripting
...
+
+- Integer
+- Default: 0
+
+Limit the maximum number of simultaneous DoT probes the Recursor will schedule.
+The default value 0 means no DoT probes are scheduled.
+
+DoT probes are used to check if an authoritative server's IP address supports DoT.
+If the probe determines an IP address supports DoT, the Recursor will use DoT to contact it for subsequent queries until a failure occurs.
+After a failure, the Recursor will stop using DoT for that specific IP address for a while.
```
jsoref, line 1083: As mentioned by someone somewhere, _a while_ isn't very helpful.

pdns/recursordist/docs/settings.rst
```diff
@@ -1066,6 +1066,31 @@ Path to a lua file to manipulate the Recursor's answers. See :doc:`lua-scripting
...
+The results of probes are remembered and can be viewed by the ``rec_control dump-dot-probe-map`` command.
+If the maximum number of pending probes is reached, no probes will be scheduled, even if no DoT status is known for an address.
+If the result of a probe is not yet available, the Recursor will contact the authoritative server in the regular way, unless an authoritative server is configured to be contacted over DoT always using :ref:`setting-dot-to-auth-names`.
+In that case no probe will be scheduled.
+
+
+Note::
+  DoT probing is an experimental feature.
+  Please test thoroughly if it is suitable in your specific production environment before enabling.
```
jsoref, line 1092:
```suggestion
  Please test thoroughly to determine if it is suitable in your specific production environment before enabling.
```

pdns/recursordist/docs/settings.rst
omoerbeek: This is under-documented on purpose, as these things will change once the code becomes less experimental.

PR #11492, opened 2022-04-01
Rec: add deferred mode for additional records

omoerbeek:
### Short description

This was left unimplemented in the initial PR.

If a task is pushed, I mark the triggering question as variable, so the result is not inserted into the packet cache. Once the additional records are retrieved (or absence is determined) no tasks are pushed and answer will be inserted into the PC. 

This mechanism has to be evaluated critically, as a additional records with low TTLs now often prevent caching the main answer in the PC as those low TTL record cause new tasks to be pushed often. So I'm actually not sure if this mechanism is effective, especially given the trend of lower and lower TTLs (e.g. check microsoft.com or any domain that hosts their mail with MS)

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/syncres.cc
+++ b/pdns/syncres.cc
@@ -254,7 +254,7 @@ SyncRes::SyncRes(const struct timeval& now) :  d_authzonequeries(0), d_outquerie
 
 static void allowAdditionalEntry(std::unordered_set<DNSName>& allowedAdditionals, const DNSRecord& rec);
 
-void SyncRes::resolveAdditionals(const DNSName& qname, QType qtype, AdditionalMode mode, std::vector<DNSRecord>& additionals, unsigned int depth, bool& taskPushed)
+void SyncRes::resolveAdditionals(const DNSName& qname, QType qtype, AdditionalMode mode, std::vector<DNSRecord>& additionals, unsigned int depth, bool& additionalsNotInCache)
 {
   vector<DNSRecord> addRecords;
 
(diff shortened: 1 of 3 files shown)


omoerbeek:
I discussed this briefly with @neilcook. He thinks the general approach is good to go forward. So I'm marking this as ready for review.

pdns/syncres.cc
```diff
@@ -254,7 +254,7 @@ SyncRes::SyncRes(const struct timeval& now) :  d_authzonequeries(0), d_outquerie
 
 static void allowAdditionalEntry(std::unordered_set<DNSName>& allowedAdditionals, const DNSRecord& rec);
 
-void SyncRes::resolveAdditionals(const DNSName& qname, QType qtype, AdditionalMode mode, std::vector<DNSRecord>& additionals, unsigned int depth)
+void SyncRes::resolveAdditionals(const DNSName& qname, QType qtype, AdditionalMode mode, std::vector<DNSRecord>& additionals, unsigned int depth, bool& taskPushed)
```
rgacogne, line 257: I was wondering if it would not be clearer to rename `taskPushed` to something like `missingAdditionals`, or `incompleteAnswer`? Because it might not be obvious why `taskPushed` leads the answer to not being inserted into the packet cache.

rgacogne (approved):
Code looks good!

> This mechanism has to be evaluated critically, as a additional records with low TTLs now often prevent caching the main answer in the PC as those low TTL record cause new tasks to be pushed often. So I'm actually not sure if this mechanism is effective, especially given the trend of lower and lower TTLs (e.g. check microsoft.com or any domain that hosts their mail with MS)

This might be something worth mentioning in the documentation? My feeling is that enabling this option means that you want these records in the answer bad enough to pay the price.

rgacogne (approved):

PR #11495, opened 2022-04-01
Coverity: 1419402 Uncaught exception in LuaWrapper

omoerbeek:
Catch any exception that might be thrown when (copy) constructing an object.
As the push method is declared `noexcept` we cannot have that.

If this is acceptable we should upstream it I suppose.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code


rgacogne (approved):
I did a quick by simulating an exception being raised:
- before:
```
terminate called without an active exception
Aborted (core dumped)
```
- after:
```
terminate called after throwing an instance of 'std::runtime_error'
  what():  test
Aborted (core dumped)
```

PR #11496, opened 2022-04-02
Prevent segfault with empty allow-from-file and allow-from options

swegener:
### Short description
parseACL() returns a nullptr when both options are set empty.

### Checklist
I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/recursordist/rec-main.cc
+++ b/pdns/recursordist/rec-main.cc
@@ -958,8 +958,6 @@ static std::shared_ptr<NetmaskGroup> parseACL(const std::string& aclFile, const 
       result->addMask(line);
     }
     g_log << Logger::Info << "Done parsing " << result->size() << " " << aclSetting << " ranges from file '" << ::arg()[aclFile] << "' - overriding '" << aclSetting << "' setting" << endl;
-
-    return result;
   }
   else if (!::arg()[aclSetting].empty()) {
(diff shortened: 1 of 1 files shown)


rgacogne:
Thanks! I wonder if it wouldn't be better to return an empty `std::shared<NetmaskGroup>` from `parseACL` instead of `nullptr` when both directives are empty, so basically to replace `return nullptr` with `return result`. Let's see what Otto prefers :)

swegener:
I've changed the logic to correctly print the warning message, if the options are set empty. It was just avoiding the segfault before.

omoerbeek:
> Thanks! I wonder if it wouldn't be better to return an empty `std::shared<NetmaskGroup>` from `parseACL` instead of `nullptr` when both directives are empty, so basically to replace `return nullptr` with `return result`. Let's see what Otto prefers :)

I like a single `return result` at the end of `parseACL` better indeed.

kpfleming:
Oops! Sorry about overlooking this when I redid that code :-)

swegener:
Updated to `return result`

omoerbeek:
I'm a bit sad none of CodeQL, coverity or scanbuild has seen this bug...

omoerbeek (approved):

Issue #9994, opened 2021-01-20
dnsdist: Ponder supporting TCP_FASTOPEN_KEY when available


rgacogne:
- Program: Authoritative, Recursor, dnsdist 
 - Issue type: Feature request

### Short description
Setting the TCP Fast Open key via `TCP_FASTOPEN_KEY` would allow accepting the same TFO cookie for all machines in a cluster (think anycast).

Y7n05h:
I will try to resolve the dnsdist part of this issue. This may take some time.

Y7n05h:
We need to set `TCP_FASTOPEN_KEY` via setsockopt . But how to generate a key? I guess this key needs a random number that is cryptographically secure. So we probably can't use `rand()` and `random_engine` from c or c++. Perhaps `/dev/urandom` could be used, I found code using them. I also noticed that there are some random number generation implementations in the current code, but I'm not sure if they can be used in scenarios where cryptographic security is required.

rgacogne:
Since the goal is to be able to use the same key on multiple dnsdist instances, likely deployed in an anycast cluster, I think it makes sense to assume that the key will be provided by the administrator in dnsdist's configuration, a bit like [setWHashedPertubation](https://dnsdist.org/guides/serverselection.html?highlight=setwhashedpertubation#setWHashedPertubation). So I wouldn't worry too much about generating the key, as it will very likely be generated outside of dnsdist.

rgacogne:
Implemented in https://github.com/PowerDNS/pdns/pull/11497.

PR #11497, opened 2022-04-03
dnsdist: Add setTCPFastOpenKey()

Y7n05h:
Signed-off-by: Y7n05h <Y7n05h@protonmail.com>

### Short description

Partially implemented #9994

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/pdns/dnsdist-lua.cc
+++ b/pdns/dnsdist-lua.cc
@@ -1993,7 +1993,7 @@ static void setupLuaConfig(LuaContext& luaCtx, bool client, bool configCheck)
   luaCtx.writeFunction("setTCPFastOpenKey", [](const std::string& keyString) {
     setLuaSideEffect();
     uint32_t key[4] = {};
-    auto ret = sscanf(keyString.c_str(), "%" SCNx32 "%" SCNx32 "%" SCNx32 "%" SCNx32, &key[0], &key[1], &key[2], &key[3]);
+    auto ret = sscanf(keyString.c_str(), "%" SCNx32 "-%" SCNx32 "-%" SCNx32 "-%" SCNx32, &key[0], &key[1], &key[2], &key[3]);
     if (ret != 4) {
       g_outputBuffer = "Invalid value passed to setTCPFastOpenKey()!\n";
       return;
--- a/pdns/dnsdist-lua.cc
+++ b/pdns/dnsdist-lua.cc
@@ -20,8 +20,10 @@
  * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
  */
 
+#include <cstdint>
 #include <dirent.h>
 #include <fstream>
+#include <cinttypes>
 
 // for OpenBSD, sys/socket.h needs to come before net/if.h
 #include <sys/socket.h>
@@ -1990,13 +1992,13 @@ static void setupLuaConfig(LuaContext& luaCtx, bool client, bool configCheck)
   luaCtx.writeFunction("setTCPInternalPipeBufferSize", [](uint64_t size) { g_tcpInternalPipeBufferSize = size; });
   luaCtx.writeFunction("setTCPFastOpenKey", [](const std::string& keyString) {
     setLuaSideEffect();
-    unsigned int key[4] = {};
-    auto ret = sscanf(keyString.c_str(), "%x-%x-%x-%x", &key[0], &key[1], &key[2], &key[3]);
+    uint32_t key[4] = {};
+    auto ret = sscanf(keyString.c_str(), "%" SCNx32 "%" SCNx32 "%" SCNx32 "%" SCNx32, &key[0], &key[1], &key[2], &key[3]);
     if (ret != 4) {
       g_outputBuffer = "Invalid value passed to setTCPFastOpenKey()!\n";
       return;
     }
-    extern vector<unsigned int> g_TCPFastOpenKey;
+    extern vector<uint32_t> g_TCPFastOpenKey;
     for (const auto i : key) {
       g_TCPFastOpenKey.push_back(i);
     }
--- a/pdns/dnsdist-lua.cc
+++ b/pdns/dnsdist-lua.cc
@@ -30,6 +30,7 @@
 #include <sys/types.h>
 #include <sys/stat.h>
 #include <thread>
+#include <vector>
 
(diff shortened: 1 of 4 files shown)


pdns/dnsdist-lua.cc
```diff
@@ -1987,6 +1988,19 @@ static void setupLuaConfig(LuaContext& luaCtx, bool client, bool configCheck)
   });
 
   luaCtx.writeFunction("setTCPInternalPipeBufferSize", [](uint64_t size) { g_tcpInternalPipeBufferSize = size; });
+  luaCtx.writeFunction("setTCPFastOpenKey", [](const std::string& keyString) {
+    setLuaSideEffect();
+    unsigned int key[4] = {};
+    auto ret = sscanf(keyString.c_str(), "%x-%x-%x-%x", &key[0], &key[1], &key[2], &key[3]);
```
omoerbeek, line 1994: This way of representing and reading the key is endian dependent. I'd go for a Base64 encoding.

omoerbeek: I see now this is the repsentation Linux uses for their `sysctl. So using this reperesentation has it's advantages. It should be tested though if setting via `dnsdist` and setting via the `sysctl` produces the same fast open cookies.

pdns/dnsdist.cc
```diff
@@ -1983,6 +1983,11 @@ static void setUpLocalBind(std::unique_ptr<ClientState>& cs)
     if (cs->fastOpenQueueSize > 0) {
 #ifdef TCP_FASTOPEN
```
omoerbeek, line 1984: This should also be if `defined(TCP_FASTOPEN_KEY)`

pdns/dnsdist.cc
```diff
@@ -1983,6 +1983,11 @@ static void setUpLocalBind(std::unique_ptr<ClientState>& cs)
...
       SSetsockopt(fd, IPPROTO_TCP, TCP_FASTOPEN, cs->fastOpenQueueSize);
+      if (!g_TCPFastOpenKey.empty()) {
```
omoerbeek, line 1986: This should be `ifdef TCP_FASTOPEN_KEY`

pdns/dnsdist-lua.cc
Y7n05h: Yes, this method relies on endianness. I also considered using `Base64` to solve it. But I wrote this because of the [`self test`](https://github.com/torvalds/linux/blob/1329e40ebd18f1171f89f9e61b7a4aed9a1de406/tools/testing/selftests/net/tcp_fastopen_backup_key.c#L62) of the linux kernel.

pdns/dnsdist.cc
Y7n05h: Oh yes

pdns/dnsdist-lua.cc
Y7n05h: Hi, I wrote a simple program to test if setting `TCP_FASTOPEN_KEY` like this is consistent with setting it using `sysctl`. I open two ports, one of which uses the method in the pull request to modify `TCP_FASTOPEN_KEY` and the other does not. 

By comparing the FastOpen Cookie changes before and after `sysctl`, I think my modification will be consistent with `sysctl` (at least in `x86_64` environment).

test environment:
> ArchLinux 5.17.1-arch1-1 x86_64

Test Program and Test Result: https://gist.github.com/Y7n05h/8bc5873b8e80951dd545de86fbd9f388

rgacogne, line 1994: It might be better to use `uint32_t` instead of `unsigned int` since it seems to be what the kernel expect, and that might prevent some surprise on uncommon architectures.
```suggestion
    uint32 key[4] = {};
    auto ret = sscanf(keyString.c_str(), "%" SCNx32 "-%" SCNx32 "-%" SCNx32 "-%" SCNx32, &key[0], &key[1], &key[2], &key[3]);
```
And then we also need to change the type of `g_TCPFastOpenKey`.

Note that we likely need to add `#include <cinttypes>` at the top of that file for the `SCNx32` macro to be defined.

rgacogne (changes requested):
Thanks a lot! The code looks good, I suggested a small change but nothing big. Do you think you could add a few lines of documentation in `docs/reference/config.rst`, likely at the end of the `Other functions` section?

pdns/dnsdist-lua.cc
Y7n05h: Okay, it's all done.

pdns/dnsdist-lua.cc
```diff
@@ -1987,6 +1990,19 @@ static void setupLuaConfig(LuaContext& luaCtx, bool client, bool configCheck)
   });
 
   luaCtx.writeFunction("setTCPInternalPipeBufferSize", [](uint64_t size) { g_tcpInternalPipeBufferSize = size; });
+  luaCtx.writeFunction("setTCPFastOpenKey", [](const std::string& keyString) {
+    setLuaSideEffect();
+    uint32_t key[4] = {};
+    auto ret = sscanf(keyString.c_str(), "%" SCNx32 "%" SCNx32 "%" SCNx32 "%" SCNx32, &key[0], &key[1], &key[2], &key[3]);
```
rgacogne, line 1996:
```suggestion
    auto ret = sscanf(keyString.c_str(), "%" SCNx32 "-%" SCNx32 "-%" SCNx32 "-%" SCNx32, &key[0], &key[1], &key[2], &key[3]);
```

pdns/dnsdistdist/docs/reference/config.rst
```diff
@@ -1651,6 +1651,14 @@ Other functions
 
   :param int size: The maximum size in bytes (default is 512)
 
+.. function:: setTCPFastOpenKey(key)
+
+  .. versionadded:: 1.7.1
+
+  Set a specified same ``TCP Fast Open`` key for all frontends.
```
rgacogne, line 1658:
```suggestion
  .. versionadded:: 1.8.0

  Set the supplied ``TCP Fast Open`` key on all frontends. This can for example be used to allow all dnsdist instances in an anycast cluster to use the same ``TCP Fast Open`` key, reducing round-trips.
```

rgacogne (approved):
Looks very good to me, thanks a lot!

PR #11396, opened 2022-03-07
Rec: proxy by table

omoerbeek:
### Short description

A way to internally used a mapped adres as the source address for a query.

### Checklist

I have:
- [X] read the [CONTRIBUTING.md](https://github.com/PowerDNS/pdns/blob/master/CONTRIBUTING.md) document
- [X] compiled this code
- [X] tested this code
- [X] included documentation (including possible behaviour changes)
- [X] documented the code
- [X] added or modified regression test(s)
- [ ] added or modified unit test(s)

PR #11507, opened 2022-04-06
Rec: proxy by table per domain queried

omoerbeek:
### Short description

Followup on #11396 

It turns out there needs to be another condition on which to let the mapping depend: the name queried.

As a bonus, fix some small issues in the `DNS Suffix Match Group` docs.

ping @neilcook 

### Checklist

I have:
- [X ] compiled this code
- [X] tested this code
- [X] documented the code

### DIFF
--- a/pdns/rec-lua-conf.cc
+++ b/pdns/rec-lua-conf.cc
@@ -725,11 +725,18 @@ void loadRecursorLuaConfig(const std::string& fname, luaConfigDelayedThreads& de
     lci.allowAdditionalQTypes.insert_or_assign(qtype, pair(targets, mode));
   });
 
-  Lua->writeFunction("addProxyMapping", [&proxyMapping](const string& netmaskArg, const string& addressArg) {
+  Lua->writeFunction("addProxyMapping", [&proxyMapping](const string& netmaskArg, const string& addressArg, boost::optional<std::vector<pair<int,std::string>>> smnStrings) {
     try {
       Netmask netmask(netmaskArg);
       ComboAddress address(addressArg);
-      proxyMapping.insert_or_assign(netmask, address);
+      boost::optional<SuffixMatchNode> smn;
+      if (smnStrings) {
+        smn = boost::make_optional(SuffixMatchNode{});
+        for (const auto& el : *smnStrings) {
+          smn->add(el.second);
+        }
+      }
+      proxyMapping.insert_or_assign(netmask, {address, smn});
     }
     catch (std::exception& e) {
       g_log << Logger::Error << "Error processing addProxyMapping: " << e.what() << endl;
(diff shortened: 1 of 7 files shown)


pdns/recursordist/docs/lua-config/proxymapping.rst
```diff
@@ -39,13 +40,17 @@ If the subnets specified in multiple :func:`addProxyMapping` calls overlap, the
...
 
   .. versionadded:: 4.7.0
 
   Specify a table based mapping for a subnet.
 
   :param string subnet: a subnet to match
   :param string ip: the IP address or IPaddress port combination to match the subnet to.
+  :param array domains: An array of strings used to fill a :ref:`dns-suffix-match-group`.
 
+If this ``domains`` argument is given to this function, only queries for names matching the :ref:`dns-suffix-match-group` will use the value ``M`` to determine the outgoing ECS, other queries will use the value ``S``.
```
neilcook, line 53:
```suggestion
If the optional ``domains`` argument is given to this function, only queries for names matching the :ref:`dns-suffix-match-group` will use the value ``M`` to determine the outgoing ECS; other queries will use the value ``S``.
```

pdns/rec-lua-conf.cc
```diff
@@ -725,11 +725,18 @@ void loadRecursorLuaConfig(const std::string& fname, luaConfigDelayedThreads& de
     lci.allowAdditionalQTypes.insert_or_assign(qtype, pair(targets, mode));
   });
 
-  Lua->writeFunction("addProxyMapping", [&proxyMapping](const string& netmaskArg, const string& addressArg) {
+  Lua->writeFunction("addProxyMapping", [&proxyMapping](const string& netmaskArg, const string& addressArg, boost::optional<std::vector<pair<int,std::string>>> smnStrings) {
     try {
       Netmask netmask(netmaskArg);
       ComboAddress address(addressArg);
-      proxyMapping.insert_or_assign(netmask, address);
+      boost::optional<SuffixMatchNode> smn;
```
Habbie, line 732: Maybe this could be `std::optional`? Not very important, but it'll be one fewer cleanup, some day.

Habbie (approved):
Few nits, please read them - otherwise, looks good.

pdns/rec-lua-conf.cc
omoerbeek: This ends up in a class handled by `LuaWrapper`, which has no `std::optional` support afaik.

Habbie: Ah, I didn't think that class would also hit `LuaWrapper`. In that case, indeed, carry on.

pdns/recursordist/docs/lua-config/proxymapping.rst
```diff
@@ -39,13 +40,17 @@ If the subnets specified in multiple :func:`addProxyMapping` calls overlap, the
...
 
   .. versionadded:: 4.7.0
 
   Specify a table based mapping for a subnet.
 
   :param string subnet: a subnet to match
   :param string ip: the IP address or IPaddress port combination to match the subnet to.
+  :param array domains: An array of strings used to fill a :ref:`dns-suffix-match-group`.
 
+If the optional ``domains`` argument is given to this function, only queries for names matching the :ref:`dns-suffix-match-group` will use the value ``M`` to determine the outgoing ECS; other queries will use the value ``S``.
```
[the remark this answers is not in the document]
Habbie: You're right!

Issue #10447, opened 2021-05-26
auth: slave zone soa check does not use tcp if udp answer was truncated


qvr:
- Program: Authoritative
 - Issue type: Bug report

### Short description
auth does not use tcp if slave zone soa check's udp answer was truncated

### Environment
 - Operating system: RHEL
 - Software version: 4.2.3 (but checking the code the issue seems to be in master as well)
 - Software source: pdns repo
 
### Steps to reproduce
1. Have a slave zone for which the SOA udp query gets truncated
2. auth reports `While checking domain freshness: Query to '10.0.0.1' for SOA of 'slavetest.com' did not return a SOA`
3. Zone does not get transferred

### Expected behaviour
auth should use TCP if UDP response was truncated

### Actual behaviour
TCP is not used, slave zone does not get AXFR'd

### Other information
-

klaus-nicat:
We just get hit by the same problem. Customer primary uses DNSSEC with plenty of RRSIG. Hence, the slave checks receives "truncated". AFAIK this could be worked around by not setting the DO bit for slave checks, as then the RRSIGs would not be included. Is there a config option to not set DO bit for slave check? Or can I simply patch the DO bit away until there is a proper fix/workaround? Thanks

Habbie:
The check uses DO on purpose, so that it can also notice signature refreshes that did not bump the serial. Right now the only way to drop the DO is to disable dnssec in your backend, which most likely is not what you want.

klaus-nicat:
Indeed. I thought of adding an option to skip the checks for signature freshness. (Signature Freshness Checks are only needed as PowerDNS as this stateless serial bumping). If the primaries behave correct and increase the serial on RRSIG changes then this aditional checks can be disabled and it is not necessary to set the DO flag. Unless adding truncate support is on the way ...

Habbie:
I think adding a flag to skip DO makes sense at this point.

klaus-nicat:
Just for the records. We now use "check-signature-freshness=no" to avoid the truncated-issue. But know we have other problems. It seems several of our customers are using PDNS as master/signer and do not increase the serial (no INCEPTION configured). Hence, with these zones we now server stale RRSIGs. Hence, the problem was not solved with our workaround, but only moved to other customers.

So, PowerDNS really should support fallback to TCP on truncated SOA-Checks to prevent this issue.

ttyS4:
fwiw. I was bitten by this for one of my personal domains.
[...]

PR #11508, opened 2022-04-06
New setting compare-signatures-on-zone-freshness-check to disable DO flag for SOA checks

klaus-nicat:
This is a workaround for #10447 until SOA Checks over TCP are supported.

Turning this off will disable the DO flag for SOA queries during zone freshness checks of secondary zones
to workaround truncated SOA responses. It will also disable signature comparison which are used to detect
signature changes even when the serial was not increased. Hence, disable this setting only if the Primary
name server always increases the serial on signature changes.
Default: yes (= old behavior)

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/docs/modes-of-operation.rst
+++ b/docs/modes-of-operation.rst
@@ -108,12 +108,12 @@ This is useful if the Primary is also PowerDNS as the serial may not be
 increased although signatures are updated. To compare also the RRSIGs,
 PowerDNS sets the DO flag when querying the SOA on the Primary. Setting
 the DO flag may trigger truncated responses and the SOA check should
-fall-back to TCP. As this fall-back is currently not supported in
+fall back to TCP. As this fallback is not currently supported in
 PowerDNS, freshness checks may fail. If it is known that the Primary
 always increases the serial on signature changes, signature comparison
 can be turned off by disabling
 :ref:`setting-compare-signatures-on-zone-freshness-check`. This will disable
-the DO flag and should work around the truncate issue.
+the DO flag and should work around the issue with truncate.
 
 When the freshness of a domain cannot be checked, e.g. because the
 master is offline, PowerDNS will retry the domain after
--- a/docs/modes-of-operation.rst
+++ b/docs/modes-of-operation.rst
@@ -109,7 +109,7 @@ increased although signatures are updated. To compare also the RRSIGs,
 PowerDNS sets the DO flag when querying the SOA on the Primary. Setting
 the DO flag may trigger truncated responses and the SOA check should
 fall-back to TCP. As this fall-back is currently not supported in
-PowerDNS, freshnes checks may fail. If it is known that the Primary
+PowerDNS, freshness checks may fail. If it is known that the Primary
 always increases the serial on signature changes, signature comparison
 can be turned off by disabling
 :ref:`setting-compare-signatures-on-zone-freshness-check`. This will disable
--- a/docs/modes-of-operation.rst
+++ b/docs/modes-of-operation.rst
@@ -95,25 +95,24 @@ Secondary operation
 On launch, PowerDNS requests from all backends a list of domains that
 have not been checked recently for changes. This should happen every
 '**refresh**' seconds, as specified in the SOA record. All domains that
-are unfresh are then checked for changes over at their master. If the
+are unfresh are then checked for changes over at their primary server. If the
 :ref:`types-SOA` serial number there is higher, the domain is
 retrieved and inserted into the database. In any case, after the check,
 the domain is declared 'fresh', and will only be checked again after
 '**refresh**' seconds have passed.
 
-If the serial on the Primary is equal to the serial on the Secondary,
-but the zone is presigned, the Secondary will also compare the RRSIG
-of the SOA and queue a zone transfer if the signatures are different.
-This is useful if the Primary is also PowerDNS as the serial may not be
-increased although signatures are updated. To compare also the RRSIGs,
-PowerDNS sets the DO flag when querying the SOA on the Primary. Setting
-the DO flag may trigger truncated responses and the SOA check should
-fall back to TCP. As this fallback is not currently supported in
-PowerDNS, freshness checks may fail. If it is known that the Primary
-always increases the serial on signature changes, signature comparison
-can be turned off by disabling
-:ref:`setting-compare-signatures-on-zone-freshness-check`. This will disable
-the DO flag and should work around the issue with truncate.
+If the serial is equal, PowerDNS as a secondary with a presigned zone
+will also compare the SOA RRSIG (signature). If the signatures are
+different, the zone is also queued for a zone transfer.
+This is useful when the primary server updates DNSSEC signatures without
+changing the zone serial. In some configurations, a PowerDNS primary can
+exhibit this behaviour.
+To allow for this check, the DO flag is set on the SOA query towards
+the primary server. In some conditions, some primary servers answer with
+a truncated SOA response (indicating TCP is required), and the freshness
+check will fail. As a workaround, the signature check and DO flag can be
+turned off by disabling
+:ref:`setting-compare-signatures-on-zone-freshness-check`.
 
 When the freshness of a domain cannot be checked, e.g. because the
 master is offline, PowerDNS will retry the domain after
--- a/docs/modes-of-operation.rst
+++ b/docs/modes-of-operation.rst
@@ -112,7 +112,7 @@ the primary server. In some conditions, some primary servers answer with
 a truncated SOA response (indicating TCP is required), and the freshness
 check will fail. As a workaround, the signature check and DO flag can be
 turned off by disabling
-:ref:`setting-compare-signatures-on-zone-freshness-check`.
+:ref:`setting-secondary-check-signature-freshness`.
 
 When the freshness of a domain cannot be checked, e.g. because the
 master is offline, PowerDNS will retry the domain after
--- a/docs/modes-of-operation.rst
+++ b/docs/modes-of-operation.rst
@@ -101,6 +101,20 @@ retrieved and inserted into the database. In any case, after the check,
 the domain is declared 'fresh', and will only be checked again after
 '**refresh**' seconds have passed.
 
+If the serial on the Primary is equal to the serial on the Secondary,
+but the zone is presigned, the Secondary will also compare the RRSIG
+of the SOA and queue a zone transfer if the signatures are different.
+This is useful if the Primary is also PowerDNS as the serial may not be
+increased although signatures are updated. To compare also the RRSIGs,
+PowerDNS sets the DO flag when querying the SOA on the Primary. Setting
+the DO flag may trigger truncated responses and the SOA check should
+fall-back to TCP. As this fall-back is currently not supported in
+PowerDNS, freshnes checks may fail. If it is known that the Primary
+always increases the serial on signature changes, signature comparison
+can be turned off by disabling
+:ref:`setting-compare-signatures-on-zone-freshness-check`. This will disable
+the DO flag and should work around the truncate issue.
+
 When the freshness of a domain cannot be checked, e.g. because the
 master is offline, PowerDNS will retry the domain after
 :ref:`setting-xfr-cycle-interval` seconds.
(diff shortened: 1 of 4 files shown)


docs/modes-of-operation.rst
```diff
@@ -101,6 +101,20 @@ retrieved and inserted into the database. In any case, after the check,
...
 '**refresh**' seconds have passed.
 
+If the serial on the Primary is equal to the serial on the Secondary,
+but the zone is presigned, the Secondary will also compare the RRSIG
+of the SOA and queue a zone transfer if the signatures are different.
+This is useful if the Primary is also PowerDNS as the serial may not be
+increased although signatures are updated. To compare also the RRSIGs,
+PowerDNS sets the DO flag when querying the SOA on the Primary. Setting
+the DO flag may trigger truncated responses and the SOA check should
+fall-back to TCP. As this fall-back is currently not supported in
```
jsoref, line 111:
```suggestion
fall back to TCP. As this fallback is not currently supported in
```

docs/modes-of-operation.rst
```diff
@@ -101,6 +101,20 @@ retrieved and inserted into the database. In any case, after the check,
...
+PowerDNS, freshness checks may fail. If it is known that the Primary
+always increases the serial on signature changes, signature comparison
+can be turned off by disabling
+:ref:`setting-compare-signatures-on-zone-freshness-check`. This will disable
+the DO flag and should work around the truncate issue.
```
jsoref, line 116:
```suggestion
the DO flag and should work around the issue with truncate.
```

docs/settings.rst
```diff
@@ -269,6 +269,21 @@ Either don't ``chroot`` on these systems or set the 'Type' of the
...
+``compare-signatures-on-zone-freshness-check``
+----------------------------------------------
+
+.. versionadded:: 4.7.0
+
+-  Boolean
+-  Default: yes
+
+Turning this off will disable the DO flag for SOA queries during zone freshness checks of secondary zones
+to workaround truncated SOA responses. It will also disable signature comparison which are used to detect
```
jsoref, line 283:
```suggestion
to work around truncated SOA responses. It will also disable signature comparison which is used to detect
```

docs/modes-of-operation.rst
```diff
@@ -101,6 +101,20 @@ retrieved and inserted into the database. In any case, after the check,
 the domain is declared 'fresh', and will only be checked again after
 '**refresh**' seconds have passed.
 
+If the serial on the Primary is equal to the serial on the Secondary,
+but the zone is presigned, the Secondary will also compare the RRSIG
+of the SOA and queue a zone transfer if the signatures are different.
+This is useful if the Primary is also PowerDNS as the serial may not be
+increased although signatures are updated. To compare also the RRSIGs,
```
jsoref, line 108:
```suggestion
This is useful if the Primary is also PowerDNS as it may not update
the serial when updating signatures. To also check the RRSIGs,
```

docs/modes-of-operation.rst
```diff
@@ -101,6 +101,20 @@ retrieved and inserted into the database. In any case, after the check,
...
+PowerDNS sets the DO flag when querying the SOA on the Primary. Setting
+the DO flag may trigger truncated responses and the SOA check should
+fall back to TCP. As this fallback is not currently supported in
+PowerDNS, freshness checks may fail. If it is known that the Primary
+always increases the serial on signature changes, signature comparison
+can be turned off by disabling
+:ref:`setting-compare-signatures-on-zone-freshness-check`. This will disable
+the DO flag and should work around the issue with truncate.
```
jsoref, line 116:
```suggestion
the DO flag and should work around the truncation issue.
```

docs/modes-of-operation.rst
```diff
@@ -101,6 +101,20 @@ retrieved and inserted into the database. In any case, after the check,
...
+of the SOA and queue a zone transfer if the signatures are different.
+This is useful if the Primary is also PowerDNS as the serial may not be
+increased although signatures are updated. To compare also the RRSIGs,
+PowerDNS sets the DO flag when querying the SOA on the Primary. Setting
+the DO flag may trigger truncated responses and the SOA check should
+fall back to TCP. As this fallback is not currently supported in
+PowerDNS, freshness checks may fail. If it is known that the Primary
+always increases the serial on signature changes, signature comparison
+can be turned off by disabling
+:ref:`setting-compare-signatures-on-zone-freshness-check`. This will disable
```
jsoref, line 115:
```suggestion
:ref:`setting-compare-signatures-on-zone-freshness-check`. This will skip
```

docs/settings.rst
```diff
@@ -269,6 +269,21 @@ Either don't ``chroot`` on these systems or set the 'Type' of the
...
+
+``compare-signatures-on-zone-freshness-check``
+----------------------------------------------
+
+.. versionadded:: 4.7.0
+
+-  Boolean
+-  Default: yes
+
+Turning this off will disable the DO flag for SOA queries during zone freshness checks of secondary zones
```
jsoref, line 282:
```suggestion
Turning this off will avoid sending the DO flag for SOA queries during zone freshness checks of secondary zones
```

docs/settings.rst
```diff
@@ -269,6 +269,21 @@ Either don't ``chroot`` on these systems or set the 'Type' of the
...
+to work around truncated SOA responses. It will also disable signature comparison which is used to detect
+signature changes even when the serial was not increased. Hence, disable this setting only if the Primary
```
jsoref, line 284:
```suggestion
to work around truncated SOA responses. It will also disable signature comparison which is used to detect
signature changes without changes to the serial. Hence, disable this setting only if the Primary
```

docs/settings.rst
```diff
@@ -269,6 +269,21 @@ Either don't ``chroot`` on these systems or set the 'Type' of the
...
+name server always increases the serial on signature changes.
```
jsoref, line 285:
```suggestion
name server always increases the serial when it updates signatures.
```

jsoref (commented):
This is more of a grammatical read.

Note that I'm not sure about column width rules for these files, and someone else should read for correctness and PoV/perspective.

pdns/slavecommunicator.cc
```diff
@@ -992,7 +996,7 @@ void CommunicatorClass::slaveRefresh(PacketHandler *P)
     }
     else if(hasSOA && theirserial == ourserial) {
       uint32_t maxExpire=0, maxInception=0;
-      if(dk.isPresigned(di.zone)) {
+      if(dk.isPresigned(di.zone) && ::arg().mustDo("compare-signatures-on-zone-freshness-check")) {
```
zeha, line 999: I think it would be cleaner if the "dnssecOk flag" came out of `val` (= `Inflighter<vector<DomainNotificationInfo>, SlaveSenderReceiver>` return data), instead of rereading the arg().

pdns/common_startup.cc
```diff
@@ -147,6 +147,7 @@ void declareArguments()
   ::arg().set("allow-notify-from","Allow AXFR NOTIFY from these IP ranges. If empty, drop all incoming notifies.")="0.0.0.0/0,::/0";
   ::arg().set("slave-cycle-interval","Schedule slave freshness checks once every .. seconds")="60";
   ::arg().set("xfr-cycle-interval","Schedule primary/secondary SOA freshness checks once every .. seconds")="60";
+  ::arg().set("compare-signatures-on-zone-freshness-check", "Set DO flag on SOA queries to receive signatures for signature comparison") = "yes";
```
zeha, line 150: personally i'd like to see a different name, but maybe its good enough.

maybe: `secondary-check-dnssecok`?

zeha:
generally lgtm, modulu naming nit

klaus-nicat:
Hi! What can I do to have this PR accepted? I would happily rename it if you can agree on a config option (or will you implement fallback to TCP for 4.7?)

zeha:
I've rebased this on top of current master and reworded the texts. Maybe @jsoref wants to give them a re-read?

Could this land in 4.7.0?

docs/settings.rst
```diff
@@ -273,6 +273,24 @@ Either don't ``chroot`` on these systems or set the 'Type' of the
...
+----------------------------------------------
+
+.. versionadded:: 4.7.0
+
+-  Boolean
+-  Default: yes
+
+Enabled by default, freshness checks for secondary zones will set the DO flag on SOA queries. PowerDNS
+uses the DNSSEC signatures in the SOA response to detect (signature) changes on the primary server, when
+the serial number was not increased.
```
jsoref, line 288:
```suggestion
can detect (signature) changes on the primary server without serial number bumps using the DNSSEC signatures in the SOA response.
```

pdns/slavecommunicator.cc
```diff
@@ -1182,7 +1182,11 @@ void CommunicatorClass::slaveRefresh(PacketHandler *P)
 
       DomainNotificationInfo dni;
       dni.di=di;
-      dni.dnssecOk = dk.doesDNSSEC();
+      if (::arg().mustDo("compare-signatures-on-zone-freshness-check")) {
```
mind04, line 1185: Checking this over and over again inside the loop is a waste of resources since the value will never change.

pdns/slavecommunicator.cc
```diff
@@ -1327,7 +1331,7 @@ void CommunicatorClass::slaveRefresh(PacketHandler *P)
     }
     else if(hasSOA && theirserial == ourserial) {
       uint32_t maxExpire=0, maxInception=0;
-      if(dk.isPresigned(di.zone)) {
+      if(dk.isPresigned(di.zone) && ::arg().mustDo("compare-signatures-on-zone-freshness-check")) {
```
mind04, line 1334:
```suggestion
      if(::arg().mustDo("compare-signatures-on-zone-freshness-check") && dk.isPresigned(di.zone)) {
```
will save a huge amount of unnecessary db queries when the check is disabled.

mind04 (changes requested):
I'm no fan of the option name 'compare-signatures-on-zone-freshness-check'. It is over descriptive but it is still unclear what it is actually doing. I think i prefer something like 'disable-signatures-freshness-check' since no one in his right mind will enable an option like this by accident, and 99.999% of the users don't need this.

zeha:
> I'm no fan of the option name 'compare-signatures-on-zone-freshness-check'. It is over descriptive but it is still unclear what it is actually doing. I think i prefer something like 'disable-signatures-freshness-check' since no one in his right mind will enable an option like this by accident, and 99.999% of the users don't need this.

Yeah, I dont like the name either. Earlier I thought maybe `secondary` should be part of the name, but `secondary-disable-signature-freshness-check` is ... loooooong.

zeha:
Renamed the setting `secondary-check-signature-freshness` (not a fan of inverted option names), applied text feedback, added some words to discourage disabling the setting. Setting is now only checked once in slaveRefresh, and isPresigned call is skipped. Thanks for the review!

pdns/slavecommunicator.cc
zeha: indeed. addressed, thanks.

pdns/slavecommunicator.cc
zeha: thanks, addressed

docs/settings.rst
zeha: thanks, committed!

docs/settings.rst
```diff
@@ -273,24 +273,26 @@ Either don't ``chroot`` on these systems or set the 'Type' of the
...
-uses the DNSSEC signatures in the SOA response to detect (signature) changes on the primary server, when
-the serial number was not increased.
+can detect (signature) changes on the primary server without serial number bumps using the DNSSEC
+signatures in the SOA response.
 
-In some scenarios, primary servers send truncated SOA responses. As a workaround, this setting can be
-turned off, and the DO flag as well as the signature checking will be disabled. To avoid additional
+In some problematic scenarios, primary servers send truncated SOA responses. As a workaround, this setting
+can be turned off, and the DO flag as well as the signature checking will be disabled. To avoid additional
 drift, primary servers then must always increase the zone serial on signature changes.
```
jsoref, line 292:
```suggestion
drift, primary servers must then always increase the zone serial on signature changes.
```

PR #11509, opened 2022-04-06
Rec: three more coverity cases: 1487743, 1419403 and 1419401

omoerbeek:
### Short description

With these three we no longer have outstanding coverity issues.

Check the commit message of the 2nd commit for some background info.

Once this is in, I plan to
- see if adding a model would reduce the amount of false positives. In particular, coverity has some issues understanding passing pointer values through pipes and sees those as leaks.
- review closed reports to see if we might have dismissed reports wrongly

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/iputils.hh
+++ b/pdns/iputils.hh
@@ -884,17 +884,20 @@ private:
     }
   }
 
-  void copyTree(const NetmaskTree& rhs)
+  void copyTree(const NetmaskTree& rhs) noexcept
   {
-    TreeNode *node;
-
-    node = rhs.d_root.get();
-    if (node != nullptr)
-      node = node->traverse_l();
-    while (node != nullptr) {
-      if (node->assigned)
-        insert(node->node.first).second = node->node.second;
-      node = node->traverse_lnr();
+    try {
+      TreeNode *node = rhs.d_root.get();
+      if (node != nullptr)
+        node = node->traverse_l();
+      while (node != nullptr) {
+        if (node->assigned)
+          insert(node->node.first).second = node->node.second;
+        node = node->traverse_lnr();
+      }
+    }
+    catch (...) {
+      abort();
     }
   }
 
@@ -967,11 +970,11 @@ public:
   NetmaskTree() noexcept: d_root(new TreeNode()), d_left(nullptr), d_size(0) {
   }
 
-  NetmaskTree(const NetmaskTree& rhs): d_root(new TreeNode()), d_left(nullptr), d_size(0) {
+  NetmaskTree(const NetmaskTree& rhs) noexcept: d_root(new TreeNode()), d_left(nullptr), d_size(0) {
     copyTree(rhs);
   }
 
-  NetmaskTree& operator=(const NetmaskTree& rhs) {
+  NetmaskTree& operator=(const NetmaskTree& rhs) noexcept {
     clear();
     copyTree(rhs);
     return *this;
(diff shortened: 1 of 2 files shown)


rgacogne:
> see if adding a model would reduce the amount of false positives. In particular, coverity has some issues understanding passing pointer values through pipes and sees those as leaks.

I believe #11427 might help with that, at least by reducing the number of places a pointer is written to a pipe to a single one. It still requires a fair amount of work, though.

pdns/iputils.hh
```diff
@@ -884,17 +884,20 @@ private:
...
+      if (node != nullptr)
+        node = node->traverse_l();
+      while (node != nullptr) {
+        if (node->assigned)
+          insert(node->node.first).second = node->node.second;
+        node = node->traverse_lnr();
+      }
+    }
+    catch (...) {
+      abort();
```
rgacogne, line 900: I agree that copying an existing tree is probably safe in most cases, but calling `abort()` if we run out of memory while trying to copy a large NetmaskTree seems wrong?

omoerbeek: I  am not very happy with that either, reading up on the exact semantics of `noexcept' to see if we can only catch `NetmaskException` here.

omoerbeek: If I understand correctly, `noexcept` fucntions that actually throw cause a runtime kill.
So `noexcept` cannot be used if we cacth `NetmaskException` only, as allocation failures might still throw. Calling code would have a chance to react on OOM situations. Let's see if Coverity likes that enough.

pdns/iputils.hh
```diff
@@ -964,14 +967,14 @@ public:
...
-  NetmaskTree() noexcept: d_root(new TreeNode()), d_left(nullptr), d_size(0) {
+  NetmaskTree(): d_root(new TreeNode()), d_left(nullptr), d_size(0) {
   }
 
   NetmaskTree(const NetmaskTree& rhs): d_root(new TreeNode()), d_left(nullptr), d_size(0) {
     copyTree(rhs);
   }
 
-  NetmaskTree& operator=(const NetmaskTree& rhs) {
+  NetmaskTree& operator=(const NetmaskTree& rhs) noexcept {
```
rgacogne, line 977:
```suggestion
  NetmaskTree& operator=(const NetmaskTree& rhs) {
```

rgacogne (changes requested):

rgacogne (approved):
:+1:

rgacogne:
> yes, in the meantime, I'm wondering why an dynamic allocation is done while the write and read seem to be blocking . We could just allocate on the stack, pass the pointer and worry no more...
This is from looking at the code, did not test it yet, might overlook something.

In most cases I can think of, especially in dnsdist, that would not work because we are often going to queue several objects into the pipe before the other end even start reading. Even if the other cases that sounds very brittle, what happens if there is  a timeout and the sender has exited the function when the receiver reads from the pipe, for example?

omoerbeek:
> > yes, in the meantime, I'm wondering why an dynamic allocation is done while the write and read seem to be blocking . We could just allocate on the stack, pass the pointer and worry no more...
> > This is from looking at the code, did not test it yet, might overlook something.
> 
> In most cases I can think of, especially in dnsdist, that would not work because we are often going to queue several objects into the pipe before the other end even start reading. Even if the other cases that sounds very brittle, what happens if there is a timeout and the sender has exited the function when the receiver reads from the pipe, for example?

Agreed, this was a bad idea as I already found out myself as well.

PR #11510, opened 2022-04-06
new option 'ignore-errors' for setting 'outgoing-axfr-expand-alias'

### LINKED ISSUE #11503: AXFR fails if outgoing-axfr-expand-alias fails
- Program: Authoritative,
 - Issue type: Bug report - maybe

### Short description
If an ALIAS target becomes unresolveable, the whole zone transfer fails, and possible adds problem to other RRs which will not be updated anymore.

### Environment
 - Software version: 4.6.1-1pdns.bionic

### Steps to reproduce
1. In pdns.conf set outgoing-axfr-expand-alias=yes
2. Add an ALIAS RR whose target is not resolveable
3. AXFR the zone from PowerDNS

### Expected behaviour
Difficult to say. PowerDNS could be liberal and just hand out the ALIAS RR itself if the RR is unresolveable instead of breaking the AXFR. On the other hand I would also accept a "Won't fix. Fix the zone" as answer, although it is much wokr to reach out to all my customers.

### Actual behaviour
```
AXFR-out zone 'test1.klaus', client '83.136.34.27', error resolving for ALIAS fdasfdsafdsafsa.com., aborting AXFR
```

klaus-nicat:
Closes #11503

If the ALIAS target can not be resolved during AXFR the AXFR
will fail. To allow outgoing AXFR also if the ALIAS targets are broken
set this setting to 'ignore-errors', but be warned, this will lead to
inconsistent zones between Primary and Secondary name server.

### Checklist
I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/tcpreceiver.cc
+++ b/pdns/tcpreceiver.cc
@@ -582,7 +582,7 @@ namespace {
 
 
 /** do the actual zone transfer. Return 0 in case of error, 1 in case of success */
-int TCPNameserver::doAXFR(const DNSName &target, std::unique_ptr<DNSPacket>& q, int outsock)
+int TCPNameserver::doAXFR(const DNSName &target, std::unique_ptr<DNSPacket>& q, int outsock)  // NOLINT(readability-function-cognitive-complexity)
 {
   string logPrefix="AXFR-out zone '"+target.toLogString()+"', client '"+q->getRemoteString()+"', ";
 


mind04:
Not only this will lead to inconsistencies between primary and secondaries. It may also result in inconsistencies between secondaries. Such inconsistencies are problematic because they interact badly with aggressive negative caching.

Sorry, but I'm no fan of this outgoing-axfr-expand-alias is "maybe" mode. It will only move the pain around (from the authoritative to the resolver operator). 

Ps. If it is safe to ignore errors while expanding alias records, why are those alias records there in the first place? 

#11503

klaus-nicat:
> Ps. If it is safe to ignore errors while expanding alias records, why are those alias records there in the first place?

Use case: Customer (PowerDNS with ALIAS RR on Pimary Nameserver) and a DNS-Provider (we) providing Secondary services. The customer puts in an ALIAS and the ALIAS is transferred as ALIAS to our PowerDNS. For whatever reason the ALIAS target dissapears - but this is not an issue if the ALIAS is transferred as ALIAS.

We (the secondary DNS provider) use public facing PowerDNS instances and also Knot instances. Knot fetches the zones from PowerDNS by AXFR. So, PowerDNS happily serves all the records of the zone except the ALIAS as it is not resolvable. When Knot transfers the zone, the transfer fails as the ALIAS can not be resolved. Hence - without this patch our secondaries are inconsisten. The PowerDNS Secondaries serve the zone except the unresolveable ALIAS whereas the Knot Secondaries do not serve the zone at all. With this patch, Knot can transfer the zone (without the unresolveable ALIAS) and Knot can also serve the zone, except the ALIAS. (in fact PowerDNS will answer SERVFAIL for the unresolveable RRs whereas Knot will answer NXDOMAIN or NODATA or the wildcard if exists). So, yes this patch is not perfect but it solves a problem which really exists.

Another option, if ALIAS expanding fails on AXFR, would be to put the ALIAS itself in the AXFR. Hence, if expanding is not possible, the put the unexpanded RR into the AXFR. That would also solve our problem and maybe is more "consistent" as it does not silently drops an RR. Would you prefer this second approach?

zeha:
Given the default for `outgoing-axfr-expand-alias` is already `no`, I think this does absolutely no harm. I.e. using ALIAS with AXFR already gives you inconsistent zones by default.

docs/settings.rst
```diff
@@ -1236,9 +1236,12 @@ To notify all IP addresses apart from the 192.168.0.0/24 subnet use the followin
 ``outgoing-axfr-expand-alias``
 ------------------------------
 
--  Boolean
+-  One of ``no``, ``yes``, or ``ignore-errors``, String
 -  Default: no
 
+.. versionchanged:: 4.7.0
+  Option 'ignore-errors' added.
```
zeha, line 1243:
```suggestion
  Option `ignore-errors` added.
```

klaus-nicat:
Hi! Is there something I can do to have this PR accepted? Thanks

zeha (approved):
I've rebased this on top of current master, and think this is a good thing to have as long as its default off.

klaus-nicat:
Any chance to get this PR merged? The 'ignore-errors' is not default, and the docs clearly mention that this feature may cause inconsistent zones. Except for the broken ALIAS, zones will be working. Omitting an unresolvable ALIAS does not have any real-world consequences (NXDOMAIN/NOERROR instead of SERVFAIL)

zeha:
Rebased to:
- fix conflict in tcpreceiver.cc
- update docs for "one sentence per line" style
- 4.7 -> 4.9

docs/guides/alias.rst
```diff
@@ -42,6 +44,11 @@ default), ALIAS records are sent as-is (RRType 65401 and a DNSName in
 the RDATA) in the AXFR. When set to 'yes', PowerDNS will lookup the A
 and AAAA records of the name in the ALIAS-record and send the results in
 the AXFR.
+If the ALIAS target can not be resolved during AXFR the AXFR will fail.
+To allow outgoing AXFR also if the ALIAS targets are broken you can set
+:ref:`setting-outgoing-axfr-expand-alias` to 'ignore-errors', but
+be warned, this will lead to inconsistent zones between the Primary and
+Secondary name servers.
```
Habbie, line 51: please add some formatting so this warning is *very* clear

pdns/tcpreceiver.cc
```diff
@@ -823,15 +823,24 @@ int TCPNameserver::doAXFR(const DNSName &target, std::unique_ptr<DNSPacket>& q,
     }
     zrr.dr.d_name.makeUsLowerCase();
     if(zrr.dr.d_name.isPartOf(target)) {
-      if (zrr.dr.d_type == QType::ALIAS && ::arg().mustDo("outgoing-axfr-expand-alias")) {
+      if (zrr.dr.d_type == QType::ALIAS && (::arg().mustDo("outgoing-axfr-expand-alias") || ::arg()["outgoing-axfr-expand-alias"] == "ignore-errors")) {
```
Habbie, line 826: `mustDo` considers `ignore-errors` as a positive answer already. However, if we ever change that, it would be good if this code survives it, so you can leave it in.

pdns/tcpreceiver.cc
```diff
@@ -823,15 +823,24 @@ int TCPNameserver::doAXFR(const DNSName &target, std::unique_ptr<DNSPacket>& q,
...
-          return 0;
+        if (ret1 != RCode::NoError || ret2 != RCode::NoError) {
+          if (::arg()["outgoing-axfr-expand-alias"] != "ignore-errors") {
+            g_log<<Logger::Warning<<logPrefix<<zrr.dr.d_name.toLogString()<<": error resolving for ALIAS "<<zrr.dr.getContent()->getZoneRepresentation()<<", aborting AXFR"<<endl;
+            outpacket->setRcode(RCode::ServFail);
+            sendPacket(outpacket,outsock);
+            return 0;
+          } else {
+            if (ret1 != RCode::NoError) {
+              g_log<<Logger::Warning<<logPrefix<<zrr.dr.d_name.toLogString()<<": error resolving A record for ALIAS target "<<zrr.dr.getContent()->getZoneRepresentation()<<", continuing AXFR"<<endl;
```
Habbie, line 838: `Logger::Error`?

Habbie: thinking some more about logging, it would also be good if `ignore-errors` was logged at Error during startup.

Habbie (changes requested):

pdns/tcpreceiver.cc
zeha: Can't say I expected mustDo to consider this a positive answer in the first place :)

Habbie: `notNoOrOff` would be a more accurate name ;)

zeha: right :)

docs/guides/alias.rst
Habbie: oh, speaking of reformatting, we're trying to transition to one sentence per line, which makes diff cleaners. Can you do that for the bits you're adding/changing?

zeha: ah yes, done

zeha:
green ❤️

docs/guides/alias.rst
```diff
@@ -34,19 +34,29 @@ When the authoritative server receives a query for the A-record for
...
+.. _alias_axfr:
+
+AXFR Zone transfers
+-------------------
+
+When a zone containing ALIAS records is transferred over AXFR, the :ref:`setting-outgoing-axfr-expand-alias` setting controls the behaviour of ALIAS records.
+
+When set to 'no' (the default), ALIAS records are sent as-is (RRType 65401 and a DNSName in the RDATA) in the AXFR.
+
+When set to 'yes', PowerDNS will lookup the A and AAAA records of the name in the ALIAS-record and send the results in the AXFR.
```
jsoref, line 48:
```suggestion
When set to 'yes', PowerDNS will look up the A and AAAA records of the name in the ALIAS-record and send the results in the AXFR.
```

docs/guides/alias.rst
```diff
@@ -34,19 +34,29 @@ When the authoritative server receives a query for the A-record for
 ``example.net``, it will resolve the A record for
 ``mywebapp.paas-provider.net`` and serve an answer for ``example.net``
 with that A record.
+If the ALIAS target can not be resolved (SERVFAIL) or does not exist (NXDOMAIN) the authoritative server will answer SERVFAIL.
```
jsoref, line 37:
```suggestion
If the ALIAS target cannot be resolved (SERVFAIL) or does not exist (NXDOMAIN) the authoritative server will answer SERVFAIL.
```

docs/guides/alias.rst
```diff
@@ -34,19 +34,29 @@ When the authoritative server receives a query for the A-record for
...
+This is useful when your secondary servers do not understand ALIAS, or should not look up the addresses themselves.
+Note that secondaries will not automatically follow changes in those A/AAAA records unless you AXFR regularly.
+
+If the ALIAS target can not be resolved, the AXFR will fail.
+When set to 'ignore-errors', an unresolvable ALIAS target will cause the ALIAS record to be ignored in the outgoing transfer instead.
+
+.. warning::
+  Setting ``setting-outgoing-axfr-expand-alias`` to 'ignore-errors', will allow an outgoing AXFR with a broken ALIAS target to complete, but the secondary server will receive an incomplete zone.
+  There is no standard mechanism for automatic re-transfer for zones broken in this way.
+  You should make sure this behaviour is acceptable in your usecase, provide custom integration tooling to monitor such problems, and possibly fix them automatically.
```
jsoref, line 58:
```suggestion
  You should make sure this behaviour is acceptable in your use case, provide custom integration tooling to monitor such problems, and possibly fix them automatically.
```

docs/settings.rst
```diff
@@ -1289,6 +1292,12 @@ If this is disabled (the default), ALIAS records are sent verbatim
 during outgoing AXFR. Note that if your slaves do not support ALIAS,
 they will return NODATA for A/AAAA queries for such names.
 
+If the ALIAS target can not be resolved during AXFR the AXFR will fail.
+To allow outgoing AXFR also if the ALIAS targets are broken set this
```
jsoref, line 1296: the `also` here is odd, do you mean `even`?

docs/guides/alias.rst
zeha: thx, done

docs/settings.rst
```diff
@@ -1289,6 +1292,12 @@ If this is disabled (the default), ALIAS records are sent verbatim
 during outgoing AXFR. Note that if your slaves do not support ALIAS,
 they will return NODATA for A/AAAA queries for such names.
 
+If the ALIAS target can not be resolved during AXFR the AXFR will fail.
```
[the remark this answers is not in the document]
zeha: thx, done

docs/guides/alias.rst
zeha: thx, done

docs/guides/alias.rst
```diff
@@ -34,19 +34,29 @@ When the authoritative server receives a query for the A-record for
...
+
+When a zone containing ALIAS records is transferred over AXFR, the :ref:`setting-outgoing-axfr-expand-alias` setting controls the behaviour of ALIAS records.
+
+When set to 'no' (the default), ALIAS records are sent as-is (RRType 65401 and a DNSName in the RDATA) in the AXFR.
+
+When set to 'yes', PowerDNS will lookup the A and AAAA records of the name in the ALIAS-record and send the results in the AXFR.
+This is useful when your secondary servers do not understand ALIAS, or should not look up the addresses themselves.
+Note that secondaries will not automatically follow changes in those A/AAAA records unless you AXFR regularly.
+
+If the ALIAS target can not be resolved, the AXFR will fail.
```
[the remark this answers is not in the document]
zeha: done

docs/guides/alias.rst
```diff
@@ -34,19 +34,29 @@ When the authoritative server receives a query for the A-record for
...
+When set to 'ignore-errors', an unresolvable ALIAS target will cause the ALIAS record to be ignored in the outgoing transfer instead.
```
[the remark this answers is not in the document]
zeha: yeah. i reworded the sentence to:

When set to 'ignore-errors', an unresolvable ALIAS target will be omitted from the outgoing transfer instead.

docs/guides/alias.rst
zeha: thx

docs/guides/alias.rst
```diff
@@ -34,19 +34,29 @@ When the authoritative server receives a query for the A-record for
...
+When a zone containing ALIAS records is transferred over AXFR, the :ref:`setting-outgoing-axfr-expand-alias` setting controls the behaviour of ALIAS records.
+
+When set to 'no' (the default), ALIAS records are sent as-is (RRType 65401 and a DNSName in the RDATA) in the AXFR.
+
+When set to 'yes', PowerDNS will look up the A and AAAA records of the name in the ALIAS-record and send the results in the AXFR.
+This is useful when your secondary servers do not understand ALIAS, or should not look up the addresses themselves.
+Note that secondaries will not automatically follow changes in those A/AAAA records unless you AXFR regularly.
+
+If the ALIAS target cannot be resolved, the AXFR will fail.
+When set to 'ignore-errors', an unresolvable ALIAS target will be omitted from the outgoing transfer instead.
```
jsoref, line 53:
```suggestion
When set to 'ignore-errors', an unresolvable ALIAS target will be omitted from the outgoing transfer.
```

I don't think `instead` adds value...

zeha: indeed. dropped it

pdns/tcpreceiver.cc
zeha: Are you okay with `Warning` in the startup code? I felt Error was too much (given the existing warnings there).

Habbie: given the footgun impact of this option, I'd really like `Error` in startup logging.

PR #11511, opened 2022-04-07
auth: Small performance improvement for the LMDB backend

rgacogne:
### Short description

This two commits improve the performance of the LMDB backend a bit (I remember seeing up to 10% reduction in CPU usage in an unrealistic benchmark):
- avoid the allocation of a string in `LMDBBackend::get()`
- reduce allocations and copies in `DNSRecordContent::deserialize()`: the move from a vector to a PacketBuffer prevents the memory from being zeroed while we are going to overwrite it right away. The change from MOADNSParser to PacketReader avoids the need to copy the header and the whole packet (allocation + copy), parse the qname again, as well as the allocation of a vector of records while we know we have only one of these.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/modules/lmdbbackend/lmdbbackend.cc
+++ b/modules/lmdbbackend/lmdbbackend.cc
@@ -853,7 +853,7 @@ bool LMDBBackend::get(DNSZoneRecord& zr)
         continue;
       }
 
-      serFromString(d_currentVal.get<string>(), d_currentrrset);
+      serFromString(d_currentVal.get<string_view>(), d_currentrrset);
       d_currentrrsetpos = 0;
     }
     else {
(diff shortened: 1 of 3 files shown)


rgacogne:
> I do note that in my testing I've seen that `DNSRecordContent::masterMake` is actually also very slow.

It can be, yes, and I hate that it yields a `shared_ptr` while most of the time we only need a `unique_ptr`. I have some ideas about that but it's going to have to wait a bit :)
But note that `MOADNSParser` called `DNSRecordContent::masterMake` already, with this PR we just do less work around that call.

Habbie (approved):

PR #11512, opened 2022-04-07
dnsdist: Bunch of small cleanups

rgacogne:
### Short description

This PR contains a bunch of small cleanups for dnsdist, most of these straightforward ones.
The only controversial one being the decision to revert the `time_t` check for dnsdist, which I believe makes sense because:
- dnsdist does not deal with signatures, and therefore is less sensitive to time issues than the auth or the rec
- it is more and more often deployed on embedded devices were this will still be an issue for some time, until a 64-bit `time_t` becomes available, and it seems silly to just prevent using dnsdist there. In fact we are seeing downstream maintainers removing that check, which is probably a good indication that it is more a hindrance than anything else. 

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/pdns/dnsdist.cc
+++ b/pdns/dnsdist.cc
@@ -1063,6 +1063,8 @@ static bool isUDPQueryAcceptable(ClientState& cs, LocalHolders& holders, const s
        the address is set to 0.0.0.0:0 which makes our sendfromto() use
        the wrong address. In that case it's better to let the kernel
        do the work by itself and use sendto() instead.
+       This is indicated by setting the family to 0 which is acted upon
+       in sendUDPResponse() and DelayedPacket::().
     */
     const ComboAddress bogusV4("0.0.0.0:0");
     const ComboAddress bogusV6("[::]:0");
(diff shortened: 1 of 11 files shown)


pdns/dnsdist.cc
```diff
@@ -1058,8 +1059,23 @@ static bool isUDPQueryAcceptable(ClientState& cs, LocalHolders& holders, const s
   }
 
   if (HarvestDestinationAddress(msgh, &dest)) {
-    /* we don't get the port, only the address */
-    dest.sin4.sin_port = cs.local.sin4.sin_port;
+    /* so it turns out that sometimes the kernel lies to us:
+       the address is set to 0.0.0.0:0 which makes our sendfromto() use
+       the wrong address. In that case it's better to let the kernel
+       do the work by itself and use sendto() instead.
```
omoerbeek, line 1065:
```suggestion
       do the work by itself and use sendto() instead. This is indicated by setting the family to 0 which is acted upon in `DelayedPacket::()`
```

pdns/dnsdistdist/configure.ac
```diff
@@ -9,8 +9,6 @@ AC_PROG_CC
 AC_PROG_CXX
 AC_LANG([C++])
 
-PDNS_CHECK_TIME_T
```
omoerbeek, line 12: I agree that this can be left out for dnsdist.

omoerbeek (approved):

PR #11517, opened 2022-04-08
Multiplexer: Take the maximum number of events as a hint

rgacogne:
### Short description

This allows indicating the maximum number of events we want to process in a single run, which is usually bounded by the number of file descriptors we are planning on watching.
The default is still 1024 events, but this change makes it possible to allocate a smaller vector of events to reduce the memory usage when we know we are going to need to process so many events in a single run.

As usual for code that touches to our multiplexer, I unfortunately do not have the platforms needed to test /dev/poll, kqueue and ports at hand, so I would appreciate any feedback, as well as a good review :)
 
### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/pdns/devpollmplexer.cc
+++ b/pdns/devpollmplexer.cc
@@ -62,7 +62,7 @@ private:
   int d_devpollfd;
 };
 
-static FDMultiplexer* makeDevPoll()
+static FDMultiplexer* makeDevPoll(unsigned int)
 {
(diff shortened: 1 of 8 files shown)


omoerbeek:
It builds and seems to run fone on OpenBSD (using kqueue). Will do a review now.

omoerbeek (approved):

Issue #9690, opened 2020-11-09
dnsdist: Consider using BPF_MAP_TYPE_LPM_TRIE for range matching


rgacogne:
- Program: dnsdist
 - Issue type: Feature request

### Short description
Linux supports the `BPF_MAP_TYPE_LPM_TRIE` type since 4.11, it would be nice to take advantage of that (when available) to offer eBPF range blacklisting in dnsdist.

### Usecase

Block a network range instead of a single address.

### Description

```
bpf = newBPFFilter(1024, 1024, 1024)
setDefaultBPFFilter(bpf)
bpf:blockRange("192.0.2.1/24")
```

Y7n05h:
I will try to finish this future request next.

rgacogne:
Implemented in #11526.

PR #11526, opened 2022-04-13
use BPF_MAP_TYPE_LPM_TRIE for range matching

Y7n05h:
### Short description
#9690 

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/contrib/xdp-filter.ebpf.src
+++ b/contrib/xdp-filter.ebpf.src
@@ -332,7 +332,7 @@ static inline enum dns_action udp_dns_reply_v4(struct cursor *c, struct CIDR4 *k
   }
 
   key->cidr = 32;
-  key->addr=bpf_htonl(key->addr);
+  key->addr = bpf_htonl(key->addr);
   value = cidr4filter.lookup(key);
   if (value) {
     __sync_fetch_and_add(&value->counter, 1);
@@ -404,7 +404,7 @@ static inline enum dns_action udp_dns_reply_v6(struct cursor *c, struct CIDR6* k
   return action;
 }
 
-int xdp_dns_filter(struct xdp_md *ctx)
+int xdp_dns_filter(struct xdp_md* ctx)
 {
   // store variables
   struct cursor   c;
@@ -419,50 +419,50 @@ int xdp_dns_filter(struct xdp_md *ctx)
 
   // pass the packet if it is not an ethernet one
   if ((eth = parse_eth(&c, &eth_proto))) {
-  	// IPv4 packets
-  	if (eth_proto == bpf_htons(ETH_P_IP))
-  	{
-  		if (!(ipv4 = parse_iphdr(&c)) || bpf_htons(ipv4->protocol != IPPROTO_UDP)) {
-  			return XDP_PASS;
-  		}
+    // IPv4 packets
+    if (eth_proto == bpf_htons(ETH_P_IP))
+    {
+      if (!(ipv4 = parse_iphdr(&c)) || bpf_htons(ipv4->protocol != IPPROTO_UDP)) {
+        return XDP_PASS;
+      }
+
       struct CIDR4 key;
       key.addr = bpf_htonl(ipv4->saddr);
       // if TC bit must not be set, apply the action
-  		if ((r = udp_dns_reply_v4(&c, &key)) != TC) {
-  	    return r == DROP ? XDP_DROP : XDP_PASS;
-  		}
-  
+      if ((r = udp_dns_reply_v4(&c, &key)) != TC) {
+        return r == DROP ? XDP_DROP : XDP_PASS;
+      }
+
       // swap src/dest IP addresses
-  		uint32_t swap_ipv4 = ipv4->daddr;
-  		ipv4->daddr = ipv4->saddr;
-  		ipv4->saddr = swap_ipv4;
-  	}
-  	// IPv6 packets
-  	else if (eth_proto == bpf_htons(ETH_P_IPV6))
-  	{
-  		;
-  		if (!(ipv6 = parse_ipv6hdr(&c)) || bpf_htons(ipv6->nexthdr != IPPROTO_UDP)) {
-  			return XDP_PASS;
-  		} 
+      uint32_t swap_ipv4 = ipv4->daddr;
+      ipv4->daddr = ipv4->saddr;
+      ipv4->saddr = swap_ipv4;
+    }
+    // IPv6 packets
+    else if (eth_proto == bpf_htons(ETH_P_IPV6)) {
+      if (!(ipv6 = parse_ipv6hdr(&c)) || bpf_htons(ipv6->nexthdr != IPPROTO_UDP)) {
+        return XDP_PASS;
+      }
       struct CIDR6 key;
       key.addr = ipv6->saddr;
-      // if TC bit must not be set, apply the action
-  		if ((r = udp_dns_reply_v6(&c, &key)) != TC) {
-  			return r == DROP ? XDP_DROP : XDP_PASS;
-  		}
-  
-      // swap src/dest IP addresses
-  		struct in6_addr swap_ipv6 = ipv6->daddr;
-  		ipv6->daddr = ipv6->saddr;
-  		ipv6->saddr = swap_ipv6;
 
-  	}
-  	// pass all non-IP packets
-  	else {
-  		return XDP_PASS;
-  	}
-  } else {
-  	return XDP_PASS;
+      // if TC bit must not be set, apply the action
+      if ((r = udp_dns_reply_v6(&c, &key)) != TC) {
+        return r == DROP ? XDP_DROP : XDP_PASS;
+      }
+
+      // swap src/dest IP addresses
+      struct in6_addr swap_ipv6 = ipv6->daddr;
+      ipv6->daddr = ipv6->saddr;
+      ipv6->saddr = swap_ipv6;
+    }
+    // pass all non-IP packets
+    else {
+      return XDP_PASS;
+    }
+  }
+  else {
+    return XDP_PASS;
   }
 
   // swap MAC addresses
--- a/contrib/xdp-filter.ebpf.src
+++ b/contrib/xdp-filter.ebpf.src
@@ -94,6 +94,17 @@ enum dns_action : uint8_t {
   TC = 2
 };
 
+struct CIDR4
+{
+  uint32_t cidr;
+  uint32_t addr;
+};
+struct CIDR6
+{
+  uint32_t cidr;
+  struct in6_addr addr;
+};
+
 /*
  * Store the matching counter and the associated action for a blocked element
  */
@@ -106,6 +117,18 @@ struct map_value
 BPF_TABLE_PINNED("hash", uint32_t, struct map_value, v4filter, 1024, "/sys/fs/bpf/dnsdist/addr-v4");
 BPF_TABLE_PINNED("hash", struct in6_addr, struct map_value, v6filter, 1024, "/sys/fs/bpf/dnsdist/addr-v6");
 BPF_TABLE_PINNED("hash", struct dns_qname, struct map_value, qnamefilter, 1024, "/sys/fs/bpf/dnsdist/qnames");
+
+/*
+ * bcc has added BPF_TABLE_PINNED7 to the latest commit of the master branch, but it has not yet been released.
+ * https://github.com/iovisor/bcc/commit/fff25a8d4d445c6156b65aa8a4016ce0d78ab7fb
+ */
+#ifndef BPF_TABLE_PINNED7
+#define BPF_TABLE_PINNED7(_table_type, _key_type, _leaf_type, _name, _max_entries, _pinned, _flags) \
+  BPF_F_TABLE(_table_type ":" _pinned, _key_type, _leaf_type, _name, _max_entries, _flags)
+#endif
+
+BPF_TABLE_PINNED7("lpm_trie", struct CIDR4, struct map_value, cidr4filter, 1024, "/sys/fs/bpf/dnsdist/cidr4", BPF_F_NO_PREALLOC);
+BPF_TABLE_PINNED7("lpm_trie", struct CIDR6, struct map_value, cidr6filter, 1024, "/sys/fs/bpf/dnsdist/cidr6", BPF_F_NO_PREALLOC);
 
 /*
  * Initializer of a cursor pointer
@@ -282,7 +305,7 @@ static inline enum dns_action check_qname(struct cursor *c)
  *         TC if (modified) message needs to be replied
  *         DROP if message needs to be blocked
  */
-static inline enum dns_action udp_dns_reply_v4(struct cursor *c, uint32_t key)
+static inline enum dns_action udp_dns_reply_v4(struct cursor *c, struct CIDR4 *key)
 {
   struct udphdr  *udp;
   struct dnshdr  *dns;
@@ -297,7 +320,7 @@ static inline enum dns_action udp_dns_reply_v4(struct cursor *c, uint32_t key)
   }	
 
   // if the address is blocked, perform the corresponding action
-  struct map_value* value = v4filter.lookup(&key);
+  struct map_value* value = v4filter.lookup(&key->addr);
 
   if (value) {
     __sync_fetch_and_add(&value->counter, 1);
@@ -306,16 +329,26 @@ static inline enum dns_action udp_dns_reply_v4(struct cursor *c, uint32_t key)
     } else {
       return value->action;
     }
-  } else {
-    enum dns_action action = check_qname(c);
-  	if (action == TC) {
-  		return set_tc_bit(udp, dns);
-  	} else {
-      return action;
+  }
+
+  key->cidr = 32;
+  key->addr=bpf_htonl(key->addr);
+  value = cidr4filter.lookup(key);
+  if (value) {
+    __sync_fetch_and_add(&value->counter, 1);
+    if (value->action == TC) {
+      return set_tc_bit(udp, dns);
+    }
+    else {
+      return value->action;
     }
   }
 
-  return PASS;
+  enum dns_action action = check_qname(c);
+  if (action == TC) {
+    return set_tc_bit(udp, dns);
+  }
+  return action;
 }
 
 /*
@@ -324,7 +357,7 @@ static inline enum dns_action udp_dns_reply_v4(struct cursor *c, uint32_t key)
  *         TC if (modified) message needs to be replied
  *         DROP if message needs to be blocked
  */
-static inline enum dns_action udp_dns_reply_v6(struct cursor *c, struct in6_addr key)
+static inline enum dns_action udp_dns_reply_v6(struct cursor *c, struct CIDR6* key)
 {
    struct udphdr  *udp;
    struct dnshdr  *dns;
@@ -341,7 +374,7 @@ static inline enum dns_action udp_dns_reply_v6(struct cursor *c, struct in6_addr
   }
 
   // if the address is blocked, perform the corresponding action
-  struct map_value* value = v6filter.lookup(&key);
+  struct map_value* value = v6filter.lookup(&key->addr);
 
   if (value) {
     __sync_fetch_and_add(&value->counter, 1);
@@ -350,16 +383,25 @@ static inline enum dns_action udp_dns_reply_v6(struct cursor *c, struct in6_addr
     } else {
       return value->action;
     }
-  } else {
-    enum dns_action action = check_qname(c);
-  	if (action == TC) {
-  		return set_tc_bit(udp, dns);
-  	} else {
-      return action;
+  }
+
+  key->cidr = 128;
+  value = cidr6filter.lookup(key);
+  if (value) {
+    __sync_fetch_and_add(&value->counter, 1);
+    if (value->action == TC) {
+      return set_tc_bit(udp, dns);
+    }
+    else {
+      return value->action;
     }
   }
 
-  return PASS;
+  enum dns_action action = check_qname(c);
+  if (action == TC) {
+    return set_tc_bit(udp, dns);
+  }
+  return action;
 }
 
 int xdp_dns_filter(struct xdp_md *ctx)
@@ -383,8 +425,10 @@ int xdp_dns_filter(struct xdp_md *ctx)
   		if (!(ipv4 = parse_iphdr(&c)) || bpf_htons(ipv4->protocol != IPPROTO_UDP)) {
   			return XDP_PASS;
   		}
+      struct CIDR4 key;
+      key.addr = bpf_htonl(ipv4->saddr);
       // if TC bit must not be set, apply the action
-  		if ((r = udp_dns_reply_v4(&c, bpf_htonl(ipv4->saddr))) != TC) {
+  		if ((r = udp_dns_reply_v4(&c, &key)) != TC) {
   	    return r == DROP ? XDP_DROP : XDP_PASS;
   		}
   
@@ -400,8 +444,10 @@ int xdp_dns_filter(struct xdp_md *ctx)
   		if (!(ipv6 = parse_ipv6hdr(&c)) || bpf_htons(ipv6->nexthdr != IPPROTO_UDP)) {
   			return XDP_PASS;
   		} 
+      struct CIDR6 key;
+      key.addr = ipv6->saddr;
       // if TC bit must not be set, apply the action
-  		if ((r = udp_dns_reply_v6(&c, ipv6->saddr)) != TC) {
+  		if ((r = udp_dns_reply_v6(&c, &key)) != TC) {
   			return r == DROP ? XDP_DROP : XDP_PASS;
   		}
   
--- a/pdns/bpf-filter.cc
+++ b/pdns/bpf-filter.cc
@@ -705,13 +705,6 @@ std::vector<std::pair<ComboAddress, uint64_t> > BPFFilter::getAddrStats()
   static_assert(sizeof(v6Addr.sin6_addr.s6_addr) == sizeof(v6Key), "POSIX mandates s6_addr to be an array of 16 uint8_t");
   memset(&v6Key, 0, sizeof(v6Key));
 
-
-  CIDR4 cidr4[2];
-  memset(cidr4, 0, sizeof(cidr4));
-  CIDR6 cidr6[2];
-  memset(cidr6, 0, sizeof(cidr6));
-  
-
   auto maps = d_maps.lock();
 
   {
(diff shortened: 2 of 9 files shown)


Y7n05h:
Like the size of the three maps specified in `newBPFFilter`, the size of the two newly added maps in `blockRange` also needs to be specified.
It may not be good practice to add two extra parameters to newBPFFilter to specify the size of the two maps.

rgacogne:
> It may not be good practice to add two extra parameters to newBPFFilter to specify the size of the two maps.

It might be time to use a table-based approach for these parameters instead?

Y7n05h:
> It might be time to use a table-based approach for these parameters instead?

Can you elaborate? I didn't understand what you meant.

rgacogne:
> Can you elaborate? I didn't understand what you meant.

Yes, my bad, I should have included an example right away! If you take a look at, for example, `setWebserverConfig` https://dnsdist.org/reference/config.html#setWebserverConfig you will see that it takes a Lua associative table instead of a fixed number of parameters. That way we can be much more flexible in what we accept, and add parameters without breaking compatibility with existing configurations.
So I think we have two options for `newBPFFilter`, we can either:
- add an additional, optional associative table after the existing parameters: `newBPFFilter(v4Parameters, v6Parameters, qnamesParameters [, optionalTable])`
- or replace all parameters by a single associative table: `newBPFFilter(table)`

The second one is a lot cleaner but would break compatibility with existing configurations. I'm fine with doing that in 1.8.0 if needed.

Y7n05h:
> > Can you elaborate? I didn't understand what you meant.
> 
> Yes, my bad, I should have included an example right away! If you take a look at, for example, `setWebserverConfig` https://dnsdist.org/reference/config.html#setWebserverConfig you will see that it takes a Lua associative table instead of a fixed number of parameters. That way we can be much more flexible in what we accept, and add parameters without breaking compatibility with existing configurations. So I think we have two options for `newBPFFilter`, we can either:
> 
> * add an additional, optional associative table after the existing parameters: `newBPFFilter(v4Parameters, v6Parameters, qnamesParameters [, optionalTable])`
> * or replace all parameters by a single associative table: `newBPFFilter(table)`
> 
> The second one is a lot cleaner but would break compatibility with existing configurations. I'm fine with doing that in 1.8.0 if needed.

If you add an optional table for newBPFFilter then newBPFFilter will become:

```
newBPFFilter(v4Parameters, v6Parameters, qnamesParameters [,external] [, optionalTable])
```

I prefer the second one. Compatibility is necessary, but `newBPFFilter` has extended the 4th parameter in a similar way. It might not be nice to add new optional parameters after optional parameters. Or we could merge the 4th parameter with the 5th parameter, but this is also not consistent with the previous program. In this case, let's put all the parameters into the table.

Y7n05h:
Why do we need use different types of value depending on `format == MapFormat::Legacy`. Why doesn't `MapFormat::Legacy` use `CounterAndActionValue` as value?

https://github.com/PowerDNS/pdns/blob/053a020a8c8a23c9c7471a9d6e1161606738b6a6/pdns/bpf-filter.cc#L211-L246

rgacogne:
> Why do we need use different types of value depending on `format == MapFormat::Legacy`. Why doesn't `MapFormat::Legacy` use `CounterAndActionValue` as value?

It's quite frustrating! It comes from the limitations on eBPF socket filter programs in older kernels, where the number of eBPF instructions was very limited, so I couldn't get a composite key based qname + qtype to work and had to use that weird construct where the key is the qname only. We could get rid of that on newer kernels, but that would mean that all of a sudden eBPF filtering would not work on older kernels, which I would like to avoid for a bit longer, at least until major distributions no longer support these kernel versions.
The non-legacy format does not care about that, in part because XDP requires a more recent kernel anyway but also because I'm OK with requiring a recent kernel for new features, as opposed to breaking existing features.
By the way, it means that we will need to be able to build on kernel where LPM trie support does not exist, but of course that feature will not be enabled then.

pdns/bpf-filter.cc
```diff
@@ -470,6 +506,101 @@ void BPFFilter::unblock(const ComboAddress& addr)
...
+  if (addr.isIPv4()) {
+    CIDR4 key(addr);
+    auto maps = d_maps.lock();
+    auto& map = maps->d_cidr4;
+    if (map.d_count >= map.d_config.d_maxItems) {
+      throw std::runtime_error("Table full when trying to block " + addr.toString());
+    }
+
+    res = bpf_lookup_elem(map.d_fd.getHandle(), &key, &value);
+    if (res != -1 && value.action == action) {
```
rgacogne, line 523: Correct me if I'm wrong because I did not test, but I think we will get a match if we are trying to insert a more specific subnet? Like, if we already have "127.0.0.0/24" in the map and try to insert "127.0.0.0/25", it looks like it will throw?
I think it might be better to only refuse to insert if we have an exact match, so we would likely need to check that the key is exactly the same if `bpf_lookup_elem()` return a match. That way we can have one action for a subnet and a different action for a more specific subnet, for example.
Does that make sense to you?

pdns/bpf-filter.cc
```diff
@@ -470,6 +506,101 @@ void BPFFilter::unblock(const ComboAddress& addr)
...
+  if (addr.isIPv4()) {
+    CIDR4 key(addr);
+    auto maps = d_maps.lock();
+    auto& map = maps->d_cidr4;
+    res = bpf_delete_elem(map.d_fd.getHandle(), &key);
+    if (res == 0) {
+      --map.d_count;
+    }
+    else {
+      res = bpf_update_elem(map.d_fd.getHandle(), &key, &value, BPF_NOEXIST);
```
rgacogne, line 578: I'm likely missing something here, why are we inserting if the deletion failed? Or is that intended as a way to allow-list a more specific entry, hence the `MatchAction::Pass` action? But then we would unblock() again to remove the allow-listed entry, which seems a bit counter-intuitive?

pdns/bpf-filter.cc
```diff
@@ -602,6 +740,32 @@ std::vector<std::pair<ComboAddress, uint64_t> > BPFFilter::getAddrStats()
     }
   }
 
+  {
```
rgacogne, line 743: There is a small issue here because we cannot return the mask, so perhaps it would be better to implement a new `std::vector<std::pair<Netmask, uint64_t> > BPFFilter::getRangeStats()` (or something like that) method instead?

contrib/xdp-filter.ebpf.src
```diff
@@ -383,8 +425,10 @@ int xdp_dns_filter(struct xdp_md *ctx)
   		if (!(ipv4 = parse_iphdr(&c)) || bpf_htons(ipv4->protocol != IPPROTO_UDP)) {
   			return XDP_PASS;
   		}
+      struct CIDR4 key;
+      key.addr = bpf_htonl(ipv4->saddr);
       // if TC bit must not be set, apply the action
```
rgacogne, line 430: Just a nit but the indentation is off here, as the comment was already misindented.

contrib/xdp-filter.ebpf.src
```diff
@@ -400,8 +444,10 @@ int xdp_dns_filter(struct xdp_md *ctx)
   		if (!(ipv6 = parse_ipv6hdr(&c)) || bpf_htons(ipv6->nexthdr != IPPROTO_UDP)) {
   			return XDP_PASS;
   		} 
+      struct CIDR6 key;
```
rgacogne, line 447: Indentation is off here as well.

pdns/bpf-filter.cc
```diff
@@ -602,6 +740,32 @@ std::vector<std::pair<ComboAddress, uint64_t> > BPFFilter::getAddrStats()
...
+    auto& map  =maps->d_cidr4;
+    int res=bpf_get_next_key(map.d_fd.getHandle(), &cidr4[0], &cidr4[1]);
```
rgacogne, line 745:
```suggestion
    auto& map = maps->d_cidr4;
    int res = bpf_get_next_key(map.d_fd.getHandle(), &cidr4[0], &cidr4[1]);
```

pdns/bpf-filter.cc
```diff
@@ -602,6 +740,32 @@ std::vector<std::pair<ComboAddress, uint64_t> > BPFFilter::getAddrStats()
...
+        result.emplace_back(ComboAddress(&v4Addr), value.counter);
+      }
+
+      res = bpf_get_next_key(map.d_fd.getHandle(), &cidr4[0], &cidr4[1]);
+    }
+  }
+
+  {
+    auto& map  =maps->d_cidr6;
+    int res=bpf_get_next_key(map.d_fd.getHandle(), &cidr6[0], &cidr6[1]);
```
rgacogne, line 758:
```suggestion
    auto& map = maps->d_cidr6;
    int res = bpf_get_next_key(map.d_fd.getHandle(), &cidr6[0], &cidr6[1]);
```

pdns/bpf-filter.cc
```diff
@@ -602,6 +740,32 @@ std::vector<std::pair<ComboAddress, uint64_t> > BPFFilter::getAddrStats()
...
+    while (res == 0) {
+      if (bpf_lookup_elem(map.d_fd.getHandle(), &cidr6[0], &value) == 0) {
+        v6Addr.sin6_addr=cidr6[0].addr;
```
rgacogne, line 761:
```suggestion
        v6Addr.sin6_addr = cidr6[0].addr;
```

pdns/dnsdistdist/docs/reference/ebpf.rst
```diff
@@ -77,6 +81,16 @@ These are all the functions, objects and methods related to the :doc:`../advance
 
     :param ComboAddress address: The address to block
 
+  .. method:: BPFFilter:blockRange(Netmask)
+
+  .. versionchanged:: 1.8.0
+
+    Block ip in this range.
```
rgacogne, line 88:
```suggestion
    Block all IP addresses in this range. 
```

pdns/dnsdistdist/docs/reference/ebpf.rst
```diff
@@ -77,6 +81,16 @@ These are all the functions, objects and methods related to the :doc:`../advance
 
     :param ComboAddress address: The address to block
 
+  .. method:: BPFFilter:blockRange(Netmask)
+
+  .. versionchanged:: 1.8.0
+
+    Block ip in this range. 
+
+    DNSDist check block range after check block ip before check block DNSName.
```
rgacogne, line 90:
```suggestion
    DNSDist eBPF code first checks if an exact IP match is found, then if a range matches, and finally if a DNSName does.
```

pdns/dnsdistdist/docs/reference/ebpf.rst
```diff
@@ -94,6 +108,18 @@ These are all the functions, objects and methods related to the :doc:`../advance
 
     :param ComboAddress address: The address to unblock
 
+  .. method:: BPFFilter:unblockRange(Netmask)
+
+  .. versionchanged:: 1.8.0
+
+    Unblock ip in this range.
```
rgacogne, line 115:
```suggestion
    Unblock this range. 
```

pdns/dnsdistdist/docs/reference/ebpf.rst
```diff
@@ -94,6 +108,18 @@ These are all the functions, objects and methods related to the :doc:`../advance
...
 
+  .. method:: BPFFilter:unblockRange(Netmask)
+
+  .. versionchanged:: 1.8.0
+
+    Unblock ip in this range. 
+
+    DNSDist check unblock range after check block ip before check block DNSName. 
+    If ip is unblocked by this method but blocked by DNSName, the packet will be accept.
+    If ip is unblocked by this method but blocked by ``block`` , the packet will be blocked.
```
rgacogne, line 119: I'm afraid I don't really understand the last sentence?
```suggestion
    DNSDist eBPF code first checks if an exact IP match is found, then if a range matches, and finally if a DNSName does.
    If a query coming from an IP or range marked as allowed, but is for a qname for which a DNSName block exists, the packet will be accepted as the first match wins.
```

rgacogne (changes requested):
This is really great work, thank you! I have made a few comments and I think there are a couple points we might have to change, but I'm very happy with that pull request :+1:

pdns/bpf-filter.cc
Y7n05h: I think throw a error is ok because of `value.action == action`. This means the same processing behavior as previously existing rules.
`blockRange("127.0.0.0/24")` and `blockRange("127.0.0.0/25")` will get a error.
But `blockRange("127.0.0.0/24")` and `unblockRange("127.0.0.0/25")` is ok because of the difficult actions.

pdns/bpf-filter.cc
Y7n05h: If we want to block an ip range, but also give an exception.
We might do like this:

```
bpf:blockRange("192.68.1.0/24")
bpf:unblockRange("192.168.1.0/28")`
```

When we unblock `192.168.1.0/28`, `bpf_delete_elem` could not find a item with key `192.168.1.0/28`. So,we need add a item with key `192.168.1.0/28` and value `MatchAction::Pass`.

I admit it's a bit counter-intuitive. But I can't think of a better way to do it.

contrib/xdp-filter.ebpf.src
Y7n05h: I think it might be caused by this code mixing tabs and spaces. This code is indented correctly when it shows tabs as two spaces in my IDE.

pdns/bpf-filter.cc
rgacogne: You are right, different actions will indeed work. It does mean that we cannot have an entry for a more specific subnet with the same action. That might be fine because it does not really make sense to have the same action for a general subnet and a more specific one, but I wonder if we might have a problem with that later, especially if the block rules are inserted by a program.. I guess we can cross that bridge when we get to it.

pdns/bpf-filter.cc
rgacogne: Yes, that's clever. Perhaps we could rename `unblockRange` to `allowRange`, or something like that, and document that it will either remove any existing entry blocking that range, or add a new entry explicitly allowing that range otherwise?

pdns/bpf-filter.cc
Y7n05h: I may not understand what you mean. Do we need to modify this code to allow adding entries for the subnet with the same action?

pdns/bpf-filter.cc
Y7n05h: Oh yeah, renaming `unblockRange` seems like a good way to describe this feature more clearly.

pdns/bpf-filter.cc
Y7n05h: You are right. But I still don't think it's a good idea to always allow users to add item with the same action for subnet. 
XDP programs run in the kernel and are relatively sensitive to runtime. The more items in the map, the longer the XDP program will run when it search in the map. Sometimes users may just forget to block the subnet under a larger netmask.
Maybe we can add an optional bool parameter called `force` to blockRange.
`blockRange(Netmask [, force])`
Insert a extra item of the same action for the subnet will be accepted if and only if `force==true`.
What do you think of this idea?

rgacogne: That sounds like a very good idea!

rgacogne (changes requested):
On issue with the current code is that we can no longer use the legacy code, as it tries to create CIDR maps that are not supported. I'll be attaching a suggestion in the next comment as gH refuses to accept it here for some reason.
Then I'm also having trouble getting the XDP program to start because it is refused by the eBPF verifier, but I'm having the same issue on master so it's not your code. On which environment are you testing, as the easiest way might be for me to use the same environment?

rgacogne:
```patch
diff --git pdns/bpf-filter.cc pdns/bpf-filter.cc
index 7c41faa44..21fb200ff 100644
--- pdns/bpf-filter.cc
+++ pdns/bpf-filter.cc
@@ -227,7 +227,7 @@ BPFFilter::Map::Map(const BPFFilter::MapConfiguration& config, BPFFilter::MapFor
         valueSize = sizeof(QNameValue);
         break;
       default:
-        throw std::runtime_error("Unsupported eBPF map type: " + std::to_string(static_cast<uint8_t>(d_config.d_type)));
+        throw std::runtime_error("Unsupported eBPF map type: " + std::to_string(static_cast<uint8_t>(d_config.d_type)) + " for legacy eBPF, perhaps you are trying to use an external program instead?");
       }
     }
     else {
@@ -364,8 +364,12 @@ BPFFilter::BPFFilter(std::unordered_map<std::string, MapConfiguration>& configs,
   maps->d_v4 = BPFFilter::Map(configs["ipv4"], d_mapFormat);
   maps->d_v6 = BPFFilter::Map(configs["ipv6"], d_mapFormat);
   maps->d_qnames = BPFFilter::Map(configs["qnames"], d_mapFormat);
-  maps->d_cidr4 = BPFFilter::Map(configs["cidr4"], d_mapFormat);
-  maps->d_cidr6 = BPFFilter::Map(configs["cidr6"], d_mapFormat);
+
+  if (d_mapFormat != BPFFilter::MapFormat::Legacy) {
+    maps->d_cidr4 = BPFFilter::Map(configs["cidr4"], d_mapFormat);
+    maps->d_cidr6 = BPFFilter::Map(configs["cidr6"], d_mapFormat);
+  }
+
   if (!external) {
     BPFFilter::MapConfiguration filters;
     filters.d_maxItems = 1;
@@ -520,6 +524,9 @@ void BPFFilter::block(const Netmask& addr, bool force, BPFFilter::MatchAction ac
     CIDR4 key(addr);
     auto maps = d_maps.lock();
     auto& map = maps->d_cidr4;
+    if (map.d_fd.getHandle() == -1) {
+      throw std::runtime_error("Trying to use an unsupported map type, likely adding a range to a legacy eBPF program");
+    }
     if (map.d_count >= map.d_config.d_maxItems) {
       throw std::runtime_error("Table full when trying to block " + addr.toString());
     }
@@ -542,6 +549,9 @@ void BPFFilter::block(const Netmask& addr, bool force, BPFFilter::MatchAction ac
 
     auto maps = d_maps.lock();
     auto& map = maps->d_cidr6;
+    if (map.d_fd.getHandle() == -1) {
+      throw std::runtime_error("Trying to use an unsupported map type, likely adding a range to a legacy eBPF program");
+    }
     if (map.d_count >= map.d_config.d_maxItems) {
       throw std::runtime_error("Table full when trying to block " + addr.toString());
     }
@@ -575,6 +585,9 @@ void BPFFilter::allow(const Netmask& addr)
     CIDR4 key(addr);
     auto maps = d_maps.lock();
     auto& map = maps->d_cidr4;
+    if (map.d_fd.getHandle() == -1) {
+      throw std::runtime_error("Trying to use an unsupported map type, likely adding a range to a legacy eBPF program");
+    }
     res = bpf_delete_elem(map.d_fd.getHandle(), &key);
     if (res == 0) {
       --map.d_count;
@@ -590,6 +603,9 @@ void BPFFilter::allow(const Netmask& addr)
 
     auto maps = d_maps.lock();
     auto& map = maps->d_cidr6;
+    if (map.d_fd.getHandle() == -1) {
+      throw std::runtime_error("Trying to use an unsupported map type, likely adding a range to a legacy eBPF program");
+    }
     res = bpf_delete_elem(map.d_fd.getHandle(), &key);
     if (res == 0) {
       --map.d_count;

```

Y7n05h:
> On issue with the current code is that we can no longer use the legacy code, as it tries to create CIDR maps that are not supported. I'll be attaching a suggestion in the next comment as gH refuses to accept it here for some reason. Then I'm also having trouble getting the XDP program to start because it is refused by the eBPF verifier, but I'm having the same issue on master so it's not your code. On which environment are you testing, as the easiest way might be for me to use the same environment?

I use [`linux-zen`](https://archlinux.org/packages/extra/x86_64/linux-zen/) in ArchLinux.

```
Kernel version: 5.17.5-zen1-1-zen
```

I test d3d121d39d0018cf1dfb6d2d47f10026c8f97878 and b1419e5e2e7799fafbf057595113c5de5d23289a by executing `sudo python xdp.py` . They work fine for me.

rgacogne:
> I use [`linux-zen`](https://archlinux.org/packages/extra/x86_64/linux-zen/) in ArchLinux.
> 
> ```
> Kernel version: 5.17.5-zen1-1-zen
> ```
> 
> I test [d3d121d](https://github.com/PowerDNS/pdns/commit/d3d121d39d0018cf1dfb6d2d47f10026c8f97878) and [b1419e5](https://github.com/PowerDNS/pdns/commit/b1419e5e2e7799fafbf057595113c5de5d23289a) by executing `sudo python xdp.py` . They work fine for me.

Thanks, that's very helpful! I'm using Arch's hardened kernel, I'll try with -zen.

rgacogne:
I did not manage to make it work last time I tried, but I will take the time to test this properly this week.

Y7n05h:
I just tested `xdp.py` on the `linux-zen 5.18.1.zen1-1` and `linux 5.18.1.arch1-1` kernels of ArchLinux, respectively, and he works fine. But when testing on `linux-hardened 5.17.12.gardened2-1`, I encountered this error:

```
unknown func bpf_probe_read#4
```

I think this is a `linux-hardened` specific problem, probably related to the kernel configuration used by `linux-hardened`.

pdns/bpf-filter.cc
```diff
@@ -470,6 +515,113 @@ void BPFFilter::unblock(const ComboAddress& addr)
...
+  value.counter = 0;
+  value.action = MatchAction::Pass;
+  if (addr.isIPv4()) {
+    CIDR4 key(addr);
+    auto maps = d_maps.lock();
+    auto& map = maps->d_cidr4;
+    if (map.d_fd.getHandle() == -1) {
+      throw std::runtime_error("Trying to use an unsupported map type, likely adding a range to a legacy eBPF program");
+    }
+    res = bpf_delete_elem(map.d_fd.getHandle(), &key);
```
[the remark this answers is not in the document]
Y7n05h: I agree with you. This does kind of suck. The semantics of `allowRange`, `blockRange` are affected by the semantics of allow packets by default. Since this shows the `add` and `remove` semantics might be better. I plan to change the api to 

```
addRangeRule(subnet:str,type:str,[force:bool])
type: allow or block

rmRangeRule(subnet)
lsRangeRule()
```

pdns/bpf-filter.hh
```diff
@@ -47,6 +51,18 @@ public:
...
   };
+  static std::string toString(MatchAction s) noexcept
+  {
+    switch (s) {
+    case MatchAction::Pass:
+      return "Pass";
+    case MatchAction::Drop:
+      return "Drop";
+    case MatchAction::Truncate:
+      return "Truncat";
```
rgacogne, line 62:
```suggestion
      return "Truncate";
```

pdns/bpf-filter.hh
```diff
@@ -47,6 +51,18 @@ public:
...
+    }
+    return "Unknow";
```
rgacogne, line 64:
```suggestion
    return "Unknown";
```

pdns/dnsdistdist/docs/reference/ebpf.rst
```diff
@@ -81,15 +81,16 @@ These are all the functions, objects and methods related to the :doc:`../advance
...
 
   .. versionchanged:: 1.8.0
 
     Block all IP addresses in this range. 
 
     DNSDist eBPF code first checks if an exact IP match is found, then if a range matches, and finally if a DNSName does.
 
-    :param string Netmask: The ip range to block
+    :param string Netmask: The ip range to block or unblock
+    :param int action: set ``action``  to ``0`` to unblock a range, set ``action`` to ``1`` to block a range.
```
rgacogne, line 93: I wonder if something like "set ``action`` to ``0`` to allow a range, set ``action`` to ``1`` to block a range." would not be clearer?
Shouldn't we document the "2"/"truncate" option there?

rgacogne (changes requested):
That looks very good, I made a few comments but my feeling is that is pull request is almost done, nice job!

rgacogne:
I missed it but there is a build error when we build without `HAVE_EBPF`:
```
bpf-filter.cc:922:32: error: use of undeclared identifier 'CounterAndActionValue'
  std::vector<std::pair<Netmask, CounterAndActionValue>> BPFFilter::getRangeRule(){
  depbase=`echo dnsdist-console.o | sed 's|[^/]*$|.deps/&|;s|\.o$||'`;\
                                 ^
  bpf-filter.cc:924:10: error: no viable conversion from returned value of type 'std::vector<std::pair<Netmask, CounterAndActionValue>>' to function return type 'int'
    return result;
           ^~~~~~
  2 errors generated.
  ```

File: pdns/dnsdistdist/docs/reference/ebpf.rst (hunk 16 above)
Comment (Y7n05h) [reply pdns/dnsdistdist/docs/reference/ebpf.rst:93]: Oh, yeah, I guess that makes sense.

**Comment by Y7n05h** [comment]
> I missed it but there is a build error when we build without HAVE_EBPF:
Can you post how you built it? I did not successfully reproduce it. The build I just tested was.

```
git clean -fXd
autoreconf -i
./configure --prefix=/usr --sysconfdir=/etc --localstatedir=/var --with-gnutls --with-libsodium --with-libssl --with-nghttp2 --with-re2 --enable-dnstap --enable-dns-over-tls --enable-dns-over-https --enable-dnscrypt
make -j16
```

File: pdns/dnsdistdist/docs/reference/ebpf.rst [hunk 17]
```diff
@@ -77,6 +81,18 @@ These are all the functions, objects and methods related to the :doc:`../advance
 
     :param ComboAddress address: The address to block
 
+  .. method:: BPFFilter:addRangeRule(Netmask , action [, force=false])
+
+  .. versionchanged:: 1.8.0
```
rgacogne, line 86:
```suggestion
    .. versionadded:: 1.8.0
```

pdns/dnsdistdist/docs/reference/ebpf.rst
```diff
@@ -94,6 +110,18 @@ These are all the functions, objects and methods related to the :doc:`../advance
...
 
+  .. method:: BPFFilter:rmRangeRule(Netmask)
+
+  .. versionchanged:: 1.8.0
+
+    :param Netmask string: The rule you want to remove
+
+  .. method:: BPFFilter:lsRangeRule()
+
+  .. versionchanged:: 1.8.0
```
rgacogne, line 121:
```suggestion
    .. versionadded:: 1.8.0
```

rgacogne:
The latest CI failure is unrelated to this PR (Google resolvers were not in a good mood: `AXFR-out zone 'example.com', client '127.0.0.1', error resolving for ALIAS google-public-dns-a.google.com., aborting AXFR`).

rgacogne (approved):

PR #11536, opened 2022-04-14
Fix Coverity 1487923 Out-of-bounds read (wrong use of sizeof)

omoerbeek:
### Short description

Rookie mistake :-(

### Checklist

I have:
- [X] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/syncres.cc
+++ b/pdns/syncres.cc
@@ -300,9 +300,9 @@ struct DoTStatus
   mutable Status d_status{Unknown};
   std::string toString() const
   {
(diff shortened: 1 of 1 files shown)


rgacogne (approved):
Fix is good, although that issue kind of makes the case by itself for using `std::array` instead of C-style arrays :)

omoerbeek:
> Fix is good, although that issue kind of makes the case by itself for using `std::array` instead of C-style arrays :)

Yo are right of course. Fixed.

rgacogne (approved):
:+1:

PR #11537, opened 2022-04-14
rec: responsestats on packet cache hits

### LINKED ISSUE #11534: response stats do not reflect packet cache hits
- Program: Recursor
 - Issue type: Bug report

### Short description

image

Statistics seem to throw bogus numbers. I can't prove it, but I have some clue:
1. you can see in the Grafana panel that NXDOMAINs are more than A. 
2. right today I migrated from Unbound to PowerDNS and the A records went down from 20 thousand per minute to 200. 

Either Unbound or PowerDNS, are throwing the wrong numbers, but since point 1 is also true, I tend to believe that PowerDNS is throwing the wrong numbers. 

### Environment
 - Software version:  PowerDNS recursor 4.5.9
 - Software source: `deb [arch=amd64] http://repo.powerdns.com/ubuntu focal-rec-45 main`
### Steps to reproduce

1. Since your tools provides only cumulative statistics, I am using a custom script, which does some arithmetic operations to throw statistics per minute
2. from the json object I am grabbing: `nxdomain-answers`, `A`, `AAAA`, `SRV` ...

omoerbeek:
### Short description

This fixes #11534 

This has to be evaluated for performance impact. If the impact is relatively high, we could change the size
histogram to be raw counts (like qtype) and do the transformation to buckets on reading the data. This would avoid the finding of the right bucket in the histogram code for each packet processed (although the implementation does use a binary search, so it should be quick)

If the performance hit is still too high after that, we should  at least document that packet count hits are not accounted in the response histograms.

If we do merge this, we should make it very clear (both in upgrade guide and other docs) the semantics have changed. Also, we should probably add these stats to `/metrics` in a Prometheus friendly way (as Prometheus histograms).

Note this also impacts auth, but only in a minor way.

### Checklist

I have:
- [X] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dynhandler.cc
+++ b/pdns/dynhandler.cc
@@ -40,8 +40,6 @@
 #include "ueberbackend.hh"
 #include "common_startup.hh"
 
-extern ResponseStats g_rs;
-
 static bool s_pleasequit;
 static string d_status;
 
@@ -182,13 +180,13 @@ string DLCCHandler(const vector<string>&parts, Utility::pid_t ppid)
 
 string DLQTypesHandler(const vector<string>&parts, Utility::pid_t ppid)
 {
-  return g_rs.getQTypeReport();
+  return g_responseStats.getQTypeReport();
 }
 
 string DLRSizesHandler(const vector<string>&parts, Utility::pid_t ppid)
 {
   typedef map<uint16_t, uint64_t> respsizes_t;
-  respsizes_t respsizes = g_rs.getSizeResponseCounts();
+  respsizes_t respsizes = g_responseStats.getSizeResponseCounts();
   ostringstream os;
   boost::format fmt("%d\t%d\n");
   for(const respsizes_t::value_type& val :  respsizes) {
--- a/pdns/nameserver.cc
+++ b/pdns/nameserver.cc
@@ -217,7 +217,7 @@ UDPNameserver::UDPNameserver( bool additional_socket )
 void UDPNameserver::send(DNSPacket& p)
 {
   const string& buffer=p.getString();
-  g_rs.submitResponse(p, true);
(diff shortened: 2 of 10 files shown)


omoerbeek:
In my tests the change is completely lost in noise. In a case where almost all answers come out of the PC, checkForCacheHit takes 0.41% of the time in both cases according to the flamegraphs I collected. Also not seeing significant difference in qps (around 185kqps). Again, with 100% PC hit rate.
Now this *could* be different with many threads or other cache architecture, so I'll also do some test on another CPU arch.

omoerbeek:
My previous tests were using flamegraphs, measuring both user and system time, there the difference is not relevant. When I look at user time only using `gprof`, the difference is noticeable in queries with a 100% packet cache hit ratio. I'll look into optimizing the code.

omoerbeek:
I think I  prefer a more fundamental approach: move all stats that are modified in the PC-cache hot (PC cache hit) path to thread local, and compute aggregate values only when needed. 

So while `SyncRes` related maps move to shared, we prefer thread local for values being modified in the PC hot path.

Another case where there is room for improvement is `burtle`, in the 100% cache hit test about 9% of the user time is spent in it.

On a side note, it seem the aging code is less than optimal: `DNSPacketMangler::moveOffset` takes2.5% , likely it is not unlined because of the throw. We should investigate a `noexcept` variant, but it will make life of the callers more painful.

fredmorcos (approved):
It seems fine to me.

I'm unable to build the recursor on my machine, even after a rebase on master. A linker issue with `arguments.o` not being of a recognized format. I don't think it's related but another pair of eyes on this PR would be good.

omoerbeek:
Superseded by TCounters work

PR #11539, opened 2022-04-14
rec: DNSSEC counters track responses sent, not actual validations performed

rgacogne:
### Short description

Since 4.1 these counters are updated for every response sent, even if the DNSSEC status was fetched from the records cache and did not involve any actual validation.

### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code


omoerbeek:
The Prometheus code in `ws-recursor.cc` also has strings describing the metrics. We really need to fix that some day, having only once source of description per metric.

rgacogne:
> The Prometheus code in `ws-recursor.cc` also has strings describing the metrics. We really need to fix that some day, having only once source of description per metric.

Good catch, thanks!

omoerbeek (approved):

PR #11544, opened 2022-04-15
dnsdist: Properly use eBPF when the DynBlock is not set

rgacogne:
### Short description

When the DynBlock rule does not set a specific action we use the default one, set with `setDynBlocksAction()`, so we should follow the same logic when determining whether to insert an eBPF block.

Fixes #11504.
### Checklist

I have:
- [x] read the [CONTRIBUTING.md](https://github.com/PowerDNS/pdns/blob/master/CONTRIBUTING.md) document
- [x] compiled this code
- [x] tested this code
- [ ] included documentation (including possible behaviour changes)
- [x] documented the code
- [ ] added or modified regression test(s)
- [ ] added or modified unit test(s)

PR #11550, opened 2022-04-15
dnsdist-1.7.x: Properly use eBPF when the DynBlock is not set

rgacogne:
### Short description

When the DynBlock rule does not set a specific action we use the default one, set with `setDynBlocksAction()`, so we should follow the same logic when determining whether to insert an eBPF block.

(cherry picked from commit 8742251acdcce8cd27d859595e5e30cecfa4e4ab)

Backport of #11544 to rel/dnsdist-1.7.x.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdist.cc
+++ b/pdns/dnsdist.cc
@@ -132,11 +132,7 @@ Rings g_rings;
 QueryCount g_qcount;
 
 GlobalStateHolder<servers_t> g_dstates;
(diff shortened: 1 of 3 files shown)


rgacogne:
Rebased to fix the jinja2 failure.

chbruyand (approved):

PR #11567, opened 2022-04-21
dnsdist: Add setVerbose() to switch the verbose mode at runtime

rgacogne:
### Short description

And also from the configuration, without having to edit the arguments passed to the program.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdist-lua.cc
+++ b/pdns/dnsdist-lua.cc
@@ -1713,7 +1713,9 @@ static void setupLuaConfig(LuaContext& luaCtx, bool client, bool configCheck)
     return pool;
   });
 
+  luaCtx.writeFunction("setVerbose", [](bool verbose) { g_verbose = verbose; });
   luaCtx.writeFunction("setVerboseHealthChecks", [](bool verbose) { g_verboseHealthChecks = verbose; });
+
   luaCtx.writeFunction("setStaleCacheEntriesTTL", [](uint64_t ttl) {
     checkParameterBound("setStaleCacheEntriesTTL", ttl, std::numeric_limits<uint32_t>::max());
     g_staleCacheEntriesTTL = ttl;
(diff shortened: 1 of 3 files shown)


pdns/dnsdist-lua.cc
```diff
@@ -1713,7 +1713,9 @@ static void setupLuaConfig(LuaContext& luaCtx, bool client, bool configCheck)
     return pool;
   });
 
+  luaCtx.writeFunction("setVerbose", [](bool verbose) { g_verbose = verbose; });
```
Habbie, line 1716: should this also run in `dnsdist -c`?

rgacogne: I don't see any drawback right away from being able to switch the verbose mode from the configuration when in client mode. Of course it will not apply to the current process but to the one we connect to if the command is ussed at runtime.

Habbie: :+1:

rgacogne:
CI failure is unrelated:
```
   FAIL: testA (test_SimpleForwardOverDoT.testSimpleForwardOverDoT)
    File "/home/runner/work/pdns/pdns/regression-tests.recursor-dnssec/test_SimpleForwardOverDoT.py", line 36, in testA
      self.assertMessageIsAuthenticated(res)
    File "/home/runner/work/pdns/pdns/regression-tests.recursor-dnssec/recursortests.py", line 856, in assertMessageIsAuthenticated
      self.assertTrue('AD' in msgFlags, "No AD flag found in the message for %s" % msg.question[0].name)
  AssertionError: False is not true : No AD flag found in the message for dns.google.
  
  ```

File: pdns/dnsdistdist/docs/reference/config.rst [hunk 2]
```diff
@@ -1055,6 +1055,14 @@ Status, Statistics and More
   :param {str} selectors: A lua table of selectors. Only queries matching all selectors are shown
   :param int num: Show a maximum of ``num`` recent queries+responses, default is 10.
 
+.. function:: setVerboseHealthChecks(verbose)
```
chbruyand, line 1058:
```suggestion
.. function:: setVerbose(verbose)
```

rgacogne: Duh! Thanks!

chbruyand (approved):

PR #11594, opened 2022-04-29
Update statistics.rst

ClaudioRifo:
Notes that help understand some UDP counters.
Note that help understand that counters from /proc/net/ are global and O.S. related.

### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdistdist/docs/statistics.rst
+++ b/pdns/dnsdistdist/docs/statistics.rst
@@ -1,7 +1,7 @@
 Statistics
 ==========
 
-dnsdist keeps statistics on the queries is receives and send out. They can be accessed in different ways:
+:program:`dnsdist` keeps statistics on the queries is receives and send out. They can be accessed in different ways:
 
 - via the console (see :ref:`Console`), using :func:`dumpStats` for the general ones,
   :func:`showServers()` for the ones related to the backends, :func:`showBinds()` for the frontends,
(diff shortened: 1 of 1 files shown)


omoerbeek:
Wouldn't it be better to refer to the authoritative documentation of these counters (if that exists)?

ClaudioRifo:
> Wouldn't it be better to refer to the authoritative documentation of these counters (if that exists)?

perhaps.. But dnsdist and powerdns do not have the same counters. (some seem to be the same but with different name)
I checked some of the counters I modified, and they seem better documented on PowerDNS (but I'm not quite sure how accurate the notes are, take a look at udp-noport-errors) . But they do not clarify that this are OS Counters.

I actually think it would be even better if there is a single page for counters (stating what product uses them).
but this is my first commit, so I would not even know how to do that, (If someone from PowerDNS creates the page I'm willing to create a table that reflects this)

Finally:
I also think it would be useful to have a separated repo for documentation. Too many considerations/validations required for a documentation change (checklist).

omoerbeek:
I actually meant "refer to the Linux docs".

omoerbeek:
https://www.kernel.org/doc/html/latest/networking/snmp_counter.html  is incomplete. But https://datatracker.ietf.org/doc/html/rfc1213 has the descriptions. Linux does change some of the names a bit though. e.g.  what Linux calls `InErrors` is called `ifInErrors` according to the RFC.

pdns/dnsdistdist/docs/statistics.rst
```diff
@@ -1,7 +1,7 @@
 Statistics
 ==========
 
-dnsdist keeps statistics on the queries is receives and send out. They can be accessed in different ways:
+:program:`dnsdist` keeps statistics on the queries is receives and send out. They can be accessed in different ways:
```
rgacogne, line 4:
```suggestion
:program:`dnsdist` keeps statistics on the queries it receives and send out. They can be accessed in different ways:
```

rgacogne (approved):
One nit (not introduced by this PR, sorry!), rest looks good!

rgacogne (approved):

PR #11598, opened 2022-05-02
Rec: shared throttle map

omoerbeek:
### Short description

Move throttle map from a thread local to shared. This map is only consulted when going auth on the net. Information collected by a thread is useful to other threads to avoid double work finding out an auth needs to be throttled.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [X] documented the code

### DIFF
--- a/pdns/syncres.cc
+++ b/pdns/syncres.cc
@@ -5060,7 +5060,7 @@ bool SyncRes::doResolveAtThisIP(const std::string& prefix, const DNSName& qname,
       if (s_serverdownmaxfails > 0 && (auth != g_rootdnsname) && s_fails.lock()->incr(remoteIP, d_now) >= s_serverdownmaxfails) {
         LOG(prefix<<qname<<": Max fails reached resolving on "<< remoteIP.toString() <<". Going full throttle for "<< s_serverdownthrottletime <<" seconds" <<endl);
         // mark server as down
-        doThrottle(d_now.tv_sec, remoteIP, g_rootdnsname, 0, s_serverdownthrottletime, 10000);
+        doThrottle(d_now.tv_sec, remoteIP, s_serverdownthrottletime, 10000);
       }
       else if (resolveret == LWResult::Result::PermanentError) {
         // unreachable, 1 minute or 100 queries
--- a/pdns/syncres.cc
+++ b/pdns/syncres.cc
@@ -294,8 +294,7 @@ public:
     d_cont.clear();
   }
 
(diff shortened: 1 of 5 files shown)


pdns/syncres.cc
```diff
@@ -3126,12 +3231,12 @@ vector<ComboAddress> SyncRes::retrieveAddressesForNS(const std::string& prefix,
 
 bool SyncRes::throttledOrBlocked(const std::string& prefix, const ComboAddress& remoteIP, const DNSName& qname, const QType qtype, bool pierceDontQuery)
 {
-  if(t_sstorage.throttle.shouldThrottle(d_now.tv_sec, std::make_tuple(remoteIP, g_rootdnsname, 0))) {
+  if (s_throttle.lock()->shouldThrottle(d_now.tv_sec, std::make_tuple(remoteIP, g_rootdnsname, 0))) {
```
rgacogne, line 3234: I wonder if it would make sense to use the wrappers from the SyncRes class instead?

rgacogne (approved):
Code looks good! We probably need to document that change, though?

pdns/syncres.cc
omoerbeek: yep, also did it for the other cases, which makes the code a little bit easier to read.

pdns/syncres.cc
```diff
@@ -4950,15 +5060,15 @@ bool SyncRes::doResolveAtThisIP(const std::string& prefix, const DNSName& qname,
       if (s_serverdownmaxfails > 0 && (auth != g_rootdnsname) && s_fails.lock()->incr(remoteIP, d_now) >= s_serverdownmaxfails) {
         LOG(prefix<<qname<<": Max fails reached resolving on "<< remoteIP.toString() <<". Going full throttle for "<< s_serverdownthrottletime <<" seconds" <<endl);
         // mark server as down
-        t_sstorage.throttle.throttle(d_now.tv_sec, std::make_tuple(remoteIP, g_rootdnsname, 0), s_serverdownthrottletime, 10000);
+        doThrottle(d_now.tv_sec, remoteIP, g_rootdnsname, 0, s_serverdownthrottletime, 10000);
```
rgacogne, line 5063:
```suggestion
        doThrottle(d_now.tv_sec, remoteIP, s_serverdownthrottletime, 10000);
```

rgacogne (approved):
Looks good, one cosmetic suggestion.

PR #11604, opened 2022-05-04
dnsdist: Fix invalid proxy protocol payload on a DoH TC to TCP retry

rgacogne:
### Short description

dnsdist forwards incoming DoH queries to its backend over UDP, and retry over TCP if the response is truncated (TC=1).
When the proxy protocol is used between dnsdist and its backend, the second query, over TCP, needs to take into account that the proxy protocol payload has already been handled. This was not properly done in that exact case because the proxy protocol payload length was not propagated to the code handling the TCP communication, leading to the query ID being edited at the wrong offset in the packet and thus to an invalid proxy protocol payload.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdistdist/doh.cc
+++ b/pdns/dnsdistdist/doh.cc
@@ -699,7 +699,7 @@ static void processDOHQuery(DOHUnitUniquePtr&& du)
 
     if (du->downstream->d_config.useProxyProtocol) {
       size_t payloadSize = 0;
-      if (addProxyProtocol(dq)) {
+      if (addProxyProtocol(dq, &payloadSize)) {
         du->proxyProtocolPayloadSize = payloadSize;
       }
     }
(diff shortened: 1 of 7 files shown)


pdns/dnsdistdist/doh.cc
```diff
@@ -699,7 +699,7 @@ static void processDOHQuery(DOHUnitUniquePtr&& du)
 
     if (du->downstream->d_config.useProxyProtocol) {
       size_t payloadSize = 0;
-      if (addProxyProtocol(dq)) {
+      if (addProxyProtocol(dq, &payloadSize)) {
```
omoerbeek, line 702: There is one `addProxyProtocol()` call left without the size pointer argument in `dnsdist.cc`. It's hard to decide for me if it needs the 2nd argument there. Maybe make it a required argument to force the developer to think if the size should be handled?

rgacogne: It's not needed for the call in `dnsdist.cc`, as the query is sent over UDP. We only need to keep the size of the payload around in the other cases because we have to prepend the DNS payload with its size over TCP, and thus we need to know where to insert it (after the proxy protocol one).
Making the size a required argument (so probably a reference) would in my mind convey a different intent: that you _have_ to use the size, which is not true.

omoerbeek: Fair enough.

omoerbeek (approved):

PR #11608, opened 2022-05-06
rec: Smarter embed of web data files, reduces size of pdns_recursor

omoerbeek:
- remove unused files d3.js underscore.js underscore-min.js
- move non-minimized files to seperate dir (for reference) and skip it when packing
- gzip -9 files
- gunzip in web server
- rename map so that switching between branches without regenerating produces a link error

### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code


rgacogne:
I'm not very comfortable with the idea of inflating the content for every HTTP query. Perhaps we could cache the inflated content, so we do that operation at most once, trading RAM for CPU?

omoerbeek:
> I'm not very comfortable with the idea of inflating the content for every HTTP query. Perhaps we could cache the inflated content, so we do that operation at most once, trading RAM for CPU?

These resources *should* be cached by any decent browser, but to be sure I'll do the caching.

zeha:
if we assume modern browsers, you could also assume they can decompress gzip...

omoerbeek:
> if we assume modern browsers, you could also assume they can decompress gzip...

To do that properly, I would need to interpret the `Accept-Encoding` request header. At the time of writing the PR, I did not feel like that.

Initially, my main motivation was to reduce the executable size and compile time of the data. Just including only the needed (minimized) files already does that for quite a bit. I might make a new PR with just that part.

omoerbeek:
Mostly obsoleted by #11859

Issue #10316, opened 2021-04-20
dns update (rfc2136) support for LMDB backend


franklouwers:
- Program: Authoritative
 - Issue type: Feature request

### Short description
https://doc.powerdns.com/authoritative/dnsupdate.html lists a few backends as being compatible with DNS Update. LMDB is not one of them.

### Usecase
As LMDB became a lot more interesting in recent updates (performance, efficiency, ...), it would be handy if the LMDB backend supported dns update.

### Description
After discussions on IRC, is seems that LMDB supports most of the required features for rfc2136, except for the listSubZone() function.

PR #11611, opened 2022-05-08
WIP LMDB dnsupdate (rfc2136) support

hlindqvist:
### Short description
WIP rfc2136 dnsupdate support for the LMDB backend #10316

- listSubZone implementation in lmdbbackend
- replaceRRSet support for replace ANY (delete) in lmdbbackend
- Split "dyndns" test suite by which tests are Mysql specific and which can be run also for other backends
- LMDB test suite changes to make it run "dyndns" test suite

Known remaining issues:
- Running the dyndns tests surfaced what appears to be a NSEC bug (test `1dyndns-update-in-between` fails), but the same problem seems to happen also via pdnsutil edit (and also on master) - #11612
- The "dyndns" test suite has limited scope (24 out of 35) for lmdb as some of the existing tests are mysql-oriented

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/modules/lmdbbackend/lmdbbackend.cc
+++ b/modules/lmdbbackend/lmdbbackend.cc
@@ -521,15 +521,27 @@ bool LMDBBackend::replaceRRSet(uint32_t domain_id, const DNSName& qname, const Q
   auto cursor = txn->txn->getCursor(txn->db->dbi);
   MDBOutVal key, val;
   string match;
-  if (qt.getCode() == QType::ANY)
+  if (rrset.empty())
     match = co(domain_id, qname.makeRelative(di.zone));
   else
     match = co(domain_id, qname.makeRelative(di.zone), qt.getCode());
 
+  bool somethingLeft = false;
+  bool foundNsec3Helper = false;
+  LMDBResourceRecord nsecLrr;
   if (!cursor.lower_bound(match, key, val)) {
     while (key.get<StringView>().rfind(match, 0) == 0) {
       if (qt.getCode() == QType::ANY || co.getQType(key.get<StringView>()) == qt)
         cursor.del();
+      else {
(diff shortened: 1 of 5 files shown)


Habbie:
@kpfleming reports he is using DNS UPDATE with LMDB today. Given that we don't have `listSubZone` (outside of this PR), I wonder how that even works.

kpfleming:
That is indeed a quandary. I can confirm that I recently added an A record for a name which already had an AAAA record, and I did it by modifying the data used by an Ansible role, and that role ran on the machine that the records point to (so definitely *not* the hidden primary), and it used an Ansible module which used `dnspython` to send a TSIG-signed UPDATE query to the hidden primary, and the record is present :-)

Since I can repeat this at will and at essentially zero cost, if there's any verbose/debug logging I can provide from a transaction I'm happy to do that.

Habbie:
> I can confirm that I recently added an A record for a name which already had an AAAA record

This specific case avoids all `listSubZone` calls :-)

kpfleming:
Lucky me! I'll test some other scenarios soon.

kpfleming:
I have since also added A and AAAA records for brand-new names (no other records existed for those names). It continues to work smoothly.

miodvallat:
I've started to look at this. The `replaceRRSet` fix (third commit) should go in independently, and I'll open a PR for it very shortly.

I am however a bit worried with the logic in the fourth commit; invoking `replaceRRSet` with no new record but a `QType` different from `ANY` should IMHO not cause the removal of all records (including NSEC helpers), as there might be several records of different `QType` for a given name; therefore the NSEC helpers should probably only get removed if there are no `A` and `AAAA` records remaining.

PR #11613, opened 2022-05-09
hashHeaderAndQName burtle optimization

omoerbeek:
### Short description

Inline the `burtle` and `burtleCI` functions and don't hash the packet one byte at the time, but compute the length and then call `burtleCI` once. Header is still done separately, as it calls `burtle` and not `burtleCI`. We could choose to hash only the relevant bits in the header some day.

Small but noticeable gproff difference (burle + burtleCI drop from 1.86s to 1.23s) in a rec run of 60s with 100% PC cache hits. So roughly 1% CPU time less in this particular test.

Unit test had to be adapted since the collisions happen with different values now.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [X] documented the code


pdns/packetcache.hh
```diff
@@ -105,27 +105,23 @@ public:
...
-
-      for (size_t idx = 0; idx < labelLen && pos < packetSize; ++idx, ++pos) {
-        const unsigned char l = dns_tolower(packet.at(pos));
-        currentHash = burtle(&l, 1, currentHash);
-      }
+      pos = std::min(pos + labelLen, packetSize);
     }
-
-    return currentHash;
+    return burtleCI(reinterpret_cast<const unsigned char*>(&packet.at(sizeof(dnsheader))), pos - sizeof(dnsheader), currentHash);
```
rgacogne, line 124: I think we could skip this call when `packetSize == sizeof(dnsheader)` but it probably doesn't matter much.
With that change we are going to call `dns_tolower()` on the label lengths as well. This should not be an issue for well-formed query names because the label lengths have to be < 64, but it means that we can create collision for invalid label lengths. I don't see any issue with that at the moment, though.

rgacogne (approved):

pdns/packetcache.hh
omoerbeek: Yes, I thought about that potential issue as well when doing this PR, and I did not see a real problem.

PR #11615, opened 2022-05-09
Max body size for Lua records

RGanor:
### Short description
Lua records can be used to monitor webservers and change the result of DNS queries according to the web server availability.
One issue with the current Lua functions is that it is not possible to limit the downloaded body size, and those risking the PowerDNS server availability by monitoring large file.

This PR limit the default body size to 1mb and allow to change it on record basis.
`recrord     IN      LUA     A       "ifurlup('http://record.example.com/file', {'ip', 'ip'} ,{ useragent='test', bodysize='15728640'  })"`
Note, the body size is in bytes.

### Checklist

I have:
- [x] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/lua-record.cc
+++ b/pdns/lua-record.cc
@@ -97,6 +97,11 @@ private:
       if (cd.opts.count("useragent")) {
         useragent = cd.opts.at("useragent");
       }
+      long maxbody_size = 1048576; // Default to 1 MegaBytes
+       if(cd.opts.count("bodySize")) {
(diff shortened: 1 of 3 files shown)


omoerbeek:
IMO it would be more useful to limit to downloaded body size without regarding an incomplete download as an error. Then potentially scan for a string in the (partial) download and return results based on that.

RGanor:
@omoerbeek Thanks for the feedback!
Your feedback make sense, to check on the already downloaded body, but I can't quite figure out how to do so. Do you have any lead on how to implement this?

PR #11624, opened 2022-05-16
SuffixMatchTree: Improve lookup performance

rgacogne:
### Short description

This pull request introduces a new zero-allocation, zero-copy way of visiting DNSName raw labels, as opposed to the existing code allocating a vector of strings for that purpose. It then combines that with the fact that C++14 introduced heterogeneous lookups in ordered containers to avoid cloning the labels during lookups, significantly reducing the cost of these.

The raw labels visitor might be useful in other places, as well, but I did not look into that.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code


rgacogne:
CI failure on Circle smells like https://gcc.gnu.org/bugzilla/show_bug.cgi?id=85965

fredmorcos (approved):
This looks fine to me. I did gcc/clang builds, ran tests, and it seems OK. I don't have sufficiently old compilers to try to reproduce the CI issue with. I'd still ask to wait for a second review before merging.

rgacogne:
Let's wait for the new CI run but it looks like fixing the compilation issue on older libc++s was easier than I expected..

I did an additional benchmark in the process and I'm consistently getting very promising results from this pull request, on my test VM at DO the number of lookups per second more than doubles with that PR applied, compared to master.

pdns/dnsname.cc
```diff
@@ -491,3 +491,61 @@ bool DNSName::has8bitBytes() const
 
   return false;
 }
+
+DNSName::RawLabelsVisitor::RawLabelsVisitor(const DNSName::string_t& storage): d_storage(storage)
+{
+  size_t position = 0;
+  while (position < storage.size()) {
+    auto labelLength = static_cast<unsigned char>(storage.at(position));
```
omoerbeek, line 499: For consistency (and pedantic correctness, as char might be multiple bytes):
```suggestion
    auto labelLength = static_cast<uint8_t>(storage.at(position));
```

pdns/dnsname.hh
```diff
@@ -273,16 +301,48 @@ struct SuffixMatchTree
     }
     return *this;
   }
+  bool operator<(const SuffixMatchTree& rhs) const
+  {
+    return strcasecmp(d_name.c_str(), rhs.d_name.c_str()) < 0;
```
omoerbeek, line 306: `strcasecmp()` is locale dependent, which in theory could lead to surprises. A method in the style of DNSName::canonCompare() is better I think. Time to factor that out?

omoerbeek (changes requested):

pdns/dnsname.cc
fredmorcos: I sort of tripped on this one too. `std::string` is a `std::basic_string<char>` so `unsigned char` is more correct. I'm not sure about the boost variant.

omoerbeek: IMO the target (a position) is more relevant than the source value. The array holding positions is `std::array<uint8_t,128>`

fredmorcos: Right.. yes that makes sense

rgacogne: Right, I'll make that change. Just for the record, though, POSIX mandates `char` to be exactly one byte :)

pdns/dnsname.hh
rgacogne: In theory that makes sense to me, yes. In practice we need to be careful because `DNSName::canonCompare()` uses `DNSName::getRawLabels()`, so probably something else is needed.
It also feels a bit out of scope for this PR to me :)

omoerbeek: I'm worried this code suddenly becomes slow if you do not use the C locale. I'm OK with doing it separately.

rgacogne: Yeah, this is a valid concern. I'll open an issue so we can track it.

omoerbeek (approved):

PR #11630, opened 2022-05-20
auth: forward EDNS Client Subnet option during ALIAS processing

### LINKED ISSUE #5469: PowerDNS Auth does not forward EDNS Client Subnet for ALIAS records

klyoungblood:
Forward EDNS subnet to authoritative nameservers when resolving ALIAS records. Fixes #5469


Habbie:
Did you test this code?

Habbie:
> Did you test this code?

Oh, perhaps the failure is just in `stubquery`

klyoungblood:
> > Did you test this code?
> 
> Oh, perhaps the failure is just in `stubquery`

Seems to have hit some complier specific issues when built with clang vs GCC, pushed a hopeful fix for that.

Habbie:
> Seems to have hit some complier specific issues when built with clang vs GCC, pushed a hopeful fix for that.

Ah, those are always fun ;)

klyoungblood:
> > Seems to have hit some complier specific issues when built with clang vs GCC, pushed a hopeful fix for that.
> 
> Ah, those are always fun ;)

I was wrong. Not compiler-specific, I just wasn't build-testing tools properly before commit. I've committed the needed Makefile changes for those, and addressed a small bug related to discard of edns scope.

jmazzi:
Checking in again to see if there is any ETA on when this might be merged.

Habbie:
I finally (sorry about that) had a look, and I believe this code generally looks good. The PR will need a rebase (just a small Makefile conflict).

Do you think you can add a test for this functionality to `regression-tests.auth-py/test_ALIAS`? With the confidence of a test backing us, we can do a deeper review.

Habbie:
(with that, I believe we should have no trouble getting this into auth 4.8)

harko12:
> I finally (sorry about that) had a look, and I believe this code generally looks good. The PR will need a rebase (just a small Makefile conflict).
> 
> Do you think you can add a test for this functionality to `regression-tests.auth-py/test_ALIAS`? With the confidence of a test backing us, we can do a deeper review.

Looking over that test, I'm not sure it would exercise the changes we made.  I may be speaking from a place of unfamiliarity with the code base, but it looks like the part we changed, which is in completePacket, is being mocked away in that test with the AliasUDPResponder() class.  

would a test more like what is in https://github.com/PowerDNS/pdns/tree/master/regression-tests/tests/alias-mx be more useful?

Habbie:
completePacket calls stubDoResolve which talks to the Python responder, so it should be the right place

pdns/dnsproxy.cc
```diff
@@ -91,14 +93,25 @@ void DNSProxy::go()
 //! look up qname target with r->qtype, plonk it in the answer section of 'r' with name aname
 bool DNSProxy::completePacket(std::unique_ptr<DNSPacket>& r, const DNSName& target,const DNSName& aname, const uint8_t scopeMask)
 {
+  bool pass_edns = false;
```
Habbie, line 96:
```suggestion
  bool pass_ecs = false;
```

pdns/dnsproxy.cc
```diff
@@ -91,14 +93,25 @@ void DNSProxy::go()
...
+  string ECSOptionStr = "";
+
+  DLOG(g_log<<"dnsproxy::completePacket: Parsed edns source: "<<r->d_eso.source.toString()<<", scope: "<<r->d_eso.scope.toString()<<", family = "<<r->d_eso.scope.getNetwork().sin4.sin_family<<endl);
+  if (r->hasEDNSSubnet())
+  {
+    ECSOptionStr = makeEDNSSubnetOptsString(r->d_eso);
+    DLOG(g_log<<"from dnsproxy::completePacket: adding edns options "<<ECSOptionStr<<endl);
```
Habbie, line 103:
```suggestion
    DLOG(g_log<<"from dnsproxy::completePacket: adding edns options "<<makeHexDump(ECSOptionStr)<<endl);
```

pdns/dnsproxy.cc
```diff
@@ -151,6 +164,15 @@ bool DNSProxy::completePacket(std::unique_ptr<DNSPacket>& r, const DNSName& targ
   DNSPacketWriter pw(packet, target, qtype);
   pw.getHeader()->rd=true;
   pw.getHeader()->id=id ^ d_xor;
+  // Add EDNS Subnet if the client sent one - issue #5469
+  if (pass_edns)
+  {
+    DLOG(g_log<<"from dnsproxy::completePacket: adding ECS option "<<ECSOptionStr<<endl);
```
Habbie, line 170:
```suggestion
    DLOG(g_log<<"from dnsproxy::completePacket: adding ECS option "<<makeHexDump(ECSOptionStr)<<endl);
```

Habbie (changes requested):
A few suggestions :)

pdns/stubresolver.hh
```diff
@@ -22,8 +22,9 @@
...
 #include "namespaces.hh"
 #include "dnsparser.hh"
+#include "ednssubnet.hh"
 
 void stubParseResolveConf();
 bool resolversDefined();
-int stubDoResolve(const DNSName& qname, uint16_t qtype, vector<DNSZoneRecord>& ret);
-int stubDoResolve(const DNSName& qname, uint16_t qtype, vector<DNSRecord>& ret);
+int stubDoResolve(const DNSName& qname, uint16_t qtype, vector<DNSZoneRecord>& ret, EDNSSubnetOpts* d_eso);
+int stubDoResolve(const DNSName& qname, uint16_t qtype, vector<DNSRecord>& ret, EDNSSubnetOpts* d_eso);
```
Habbie, line 30:
```suggestion
int stubDoResolve(const DNSName& qname, uint16_t qtype, vector<DNSZoneRecord>& ret, EDNSSubnetOpts* d_eso = nullptr);
int stubDoResolve(const DNSName& qname, uint16_t qtype, vector<DNSRecord>& ret, EDNSSubnetOpts* d_eso = nullptr);
```

would avoid having to touch all other users of the function

Habbie (changes requested):

pdns/stubresolver.hh
klyoungblood: Implemented.

pdns/dnsproxy.cc
klyoungblood: Implemented.

pdns/dnsproxy.cc
klyoungblood: Implemented.

pdns/dnsproxy.cc
klyoungblood: Implemented.

klyoungblood:
> I tested this a bit, and it indeed forwards the client's ECS on to the backend resolver, but the resolver's ECS data does not make it back to the client. It looks like the client just gets a copy of their ECS question option back.

Our use case for this, at least, is for a backend resolver to return an A record of a CDN endpoint nearest the reported EDNS subnet.

Habbie:
> Our use case for this, at least, is for a backend resolver to return an A record of a CDN endpoint nearest the reported EDNS subnet.

Understood, but without the scope coming back, an ECS-aware resolver (which is what is talking to your auth in this case) cannot cache the response. In fact, with scope zero, I think it might reuse the response for all its clients, regardless of subnet.

Habbie:
Sorry, missed your comment earlier. Please see below. Let me know if there are questions.

config, (stripped from things that are irrelevant, unless I made a mistake!):

```
expand-alias
edns-subnet-processing
resolver=9.9.9.11
launch=gsqlite3
gsqlite3-database=./pdns.sqlite3
local-port=5300
```

below, dig output + tcpdump of traffic between auth and 9.9.9.11

# master:

```
$ dig a google.example.com @127.0.0.1 -p 5300 +subnet=194.109.6.0/24

; <<>> DiG 9.16.33-Debian <<>> a google.example.com @127.0.0.1 -p 5300 +subnet=194.109.6.0/24
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 30406
;; flags: qr aa rd; QUERY: 1, ANSWER: 1, AUTHORITY: 0, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 1232
; CLIENT-SUBNET: 194.109.6.0/24/0
;; QUESTION SECTION:
;google.example.com.		IN	A

;; ANSWER SECTION:
google.example.com.	300	IN	A	142.251.39.100

;; Query time: 16 msec
;; WHEN: Mon Jan 30 09:40:40 CET 2023
;; MSG SIZE  rcvd: 74
```

```
09:41:27.274860 IP (tos 0x0, ttl 64, id 19862, offset 0, flags [DF], proto UDP (17), length 60)
    10.91.39.50.52799 > 9.9.9.11.53: [bad udp cksum 0x43da -> 0x4061!] 7808+ A? www.google.com. (32)
09:41:27.282656 IP (tos 0x0, ttl 59, id 30150, offset 0, flags [none], proto UDP (17), length 76)
    9.9.9.11.53 > 10.91.39.50.52799: [udp sum ok] 7808 q: A? www.google.com. 1/0/0 www.google.com. A 142.251.39.100 (48)
```

no ECS in callout to 9.9.9.11, mirrored ECS option in response from auth.

# this PR:

```
$ dig a google.example.com @127.0.0.1 -p 5300 +subnet=194.109.6.0/24

; <<>> DiG 9.16.33-Debian <<>> a google.example.com @127.0.0.1 -p 5300 +subnet=194.109.6.0/24
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 8121
;; flags: qr aa rd; QUERY: 1, ANSWER: 1, AUTHORITY: 0, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 1232
; CLIENT-SUBNET: 194.109.6.0/24/0
;; QUESTION SECTION:
;google.example.com.		IN	A

;; ANSWER SECTION:
google.example.com.	300	IN	A	142.251.36.4

;; Query time: 12 msec
;; WHEN: Mon Jan 30 09:44:29 CET 2023
;; MSG SIZE  rcvd: 74
```

```
09:44:29.367234 IP (tos 0x0, ttl 64, id 30208, offset 0, flags [DF], proto UDP (17), length 82)
    10.91.39.50.52102 > 9.9.9.11.53: [bad udp cksum 0x43f0 -> 0x89eb!] 9631+ [1au] A? www.google.com. ar: . OPT UDPsize=512 [ECS 194.109.6.0/24/0] (54)
09:44:29.376677 IP (tos 0x0, ttl 59, id 8792, offset 0, flags [none], proto UDP (17), length 98)
    9.9.9.11.53 > 10.91.39.50.52102: [udp sum ok] 9631 q: A? www.google.com. 1/0/1 www.google.com. A 142.251.36.4 ar: . OPT UDPsize=1232 [ECS 194.109.6.0/24/19] (70)
```

ECS in callout to 9.9.9.11, but ECS from client still mirrored in response from auth. Note the identical CLIENT-SUBNET lines in both dig outputs.

9.9.9.11 gave us a /19 scope for this answer. If the auth would return that to the client, the client can cache this response for that whole /19. This would show as `; CLIENT-SUBNET:O 194.109.6.0/24/19` in the dig output.

pdns/dnsproxy.cc
```diff
@@ -260,7 +260,7 @@ void DNSProxy::mainloop()
 
         DNSPacket p(false),q(false);
         p.parse(buffer,(size_t)len);
-        q.parse(buffer,(size_t)len);
+        //q.parse(buffer,(size_t)len); // removed.  never used?
```
harko12, line 263: @Habbie is this ok to remove?  It didn't seem to be used anywhere.

pdns/iputils.hh
```diff
@@ -261,6 +261,16 @@ union ComboAddress {
     return ret;
   }
 
+  bool isValid() const
```
harko12, line 264: We added this isValid methods so we could easily test if the scope was currently set to something or not.

pdns/stubresolver.cc
```diff
@@ -122,6 +124,18 @@
   DNSPacketWriter pw(packet, qname, qtype);
   pw.getHeader()->id=dns_random_uint16();
   pw.getHeader()->rd=1;
+  
+  if(d_eso != nullptr)
+  {
+    // EDNSSubnetOpts d_eso2;
+    // d_eso2.source = d_eso->source;
+    // pass along EDNS subnet from client if given - issue #5469
```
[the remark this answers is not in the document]
klyoungblood: Cleaned that up.

Habbie:
I wrote a test for this PR. You can find it in the last 3 commits (please cherry-pick them individually from oldest to newest, so from bottom to top) in https://github.com/Habbie/pdns/commits/pr11630-test

Failing run, when the test sits on top of an older commit in this PR: https://github.com/Habbie/pdns/actions/runs/4114903872/jobs/7103860762#step:9:433

Passing run, when the test sits on top of a later commit, after you added the copying of scope from response to response: https://github.com/Habbie/pdns/actions/runs/4114976582/jobs/7103921619#step:9:424

klaus-nicat:
@klyoungblood Just a note if you plan to use ALIAS records heavily: They do not support DNSSEC yet. Maybe you want to fix that too ;-)

bghira:
as someone who looks forward to ALIAS support, scope creep for DNSSEC support seems like it should be a follow-up PR

pdns/dnsproxy.cc
Habbie: yes

pdns/iputils.hh
```diff
@@ -261,6 +261,16 @@ union ComboAddress {
...
+  {
+    char host[1024];
+    int retval = 0;
+    if(sin4.sin_family && !(retval = getnameinfo(reinterpret_cast<const struct sockaddr*>(this), getSocklen(), host, sizeof(host),0, 0, NI_NUMERICHOST)))
```
Habbie, line 268: what exactly does this check?

Habbie (changes requested):
One response, one question.

Please leave all cherrypicks from my test branch as separate commits.

Will do anoter full review once the above is addressed

pdns/iputils.hh
harko12: We ran into a situation with the logging where if there was no edns info passed along, it still tried to print out the edns info, and it came out as 'Invalid ' because the toString method could not render it.  I wanted a way to check if there was something in the scope and subnet at all, so I used the same check that toString did, but had it let me know if there was or was not something printable there.  If there isn't, then we don't try to log it out.

harko12:
> One response, one question.
> 
> Please leave all cherrypicks from my test branch as separate commits.
> 
> Will do anoter full review once the above is addressed

Re-did the test as cherry picks from your branch.

jmazzi:
Thanks for all the help on this so far. Could we please request another review? We believe we've implemented everything required.

jmazzi:
@Habbie did you have any questions for us?

pdns/Makefile.am
```diff
@@ -790,7 +792,9 @@ ixplore_SOURCES = \
 	dnsrecords.cc \
 	dnssecinfra.cc \
 	dnswriter.cc dnswriter.hh \
-        gss_context.cc gss_context.hh  \
+	ednsoptions.cc ednsoptions.hh \
+	ednssubnet.cc ednssubnet.hh \
+  gss_context.cc gss_context.hh  \
```
Habbie, line 797: this is misaligned

pdns/dnspacket.cc
```diff
@@ -184,7 +184,7 @@
     }
     d_dedup.insert(hash);
   }
-
+  //cerr<<"adding zone record with name "<<rr.dr.d_name<<" and scopmask "<<std::to_string(rr.scopeMask)<<endl;
```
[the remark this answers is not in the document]
Habbie: please remove these or switch them to DLOG

pdns/dnspacket.hh
```diff
@@ -172,6 +172,7 @@ public:
   static bool s_doEDNSSubnetProcessing;
   static bool s_doEDNSCookieProcessing;
   static string s_EDNSCookieKey;
+  EDNSSubnetOpts d_eso; //moved to pass along to stubDoResolve
```
Habbie, line 175: "moved" will not tell the next reader much that is useful

Habbie (changes requested):
I see a few more nits, and there was that bit of you code for which you asked if you could remove it - I see it's still there.

Mostly this looks to be in great shape. I will do a full review once the diff looks somewhat cleaner (mostly around commented cerr lines, unnecessary white space changes, etc.)

Habbie:
I'll pick up the remaining work :)

Habbie:
Can you turn on "Allow edits and access to secrets by maintainers"? It should be in the right sidebar on this page.

bghira:
after discussion with OP, it seems he's not working at the company, or, on this project anymore. if you close this PR out, i can open a new one on their behalf. otherwise, it might work best for you to open one, so that you have complete control over whatever changes are required @Habbie

Habbie:
thank you for letting me know. I'll open a new one soon :)

Habbie:
replaced by #13441

PR #11631, opened 2022-05-23
rec: use structured logging for startup and config reporting

omoerbeek:
### Short description

An interesting part is `arguments.[ch]`, where common code with auth is touched. I have chosen to do that using an `#fdef RECURSOR` for now. Still needs a few changes in auth to make it compile. An alternative would be to eliminate *all* logging form the common code, but I figured that would be more intrusive as all callers wold have to be modified.

There's also the issue of protecting `strftime` calls, as it can does `settz` internally, which is not thread safe and I observed actual cases with tsan. I have solved that for the only structured logging backend we have, other cases of `strftime` will be done in a separate PR.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code


chbruyand (approved):
Just a few messages that could be improved IMO (nothing blocking), but looks good otherwise

Habbie (approved):
auth bits approved

PR #11652, opened 2022-05-28
dnsdist: add AF_XDP support for UDP

Y7n05h:
Signed-off-by: Y7n05h <Y7n05h@protonmail.com>

### Short description

Added partial `AF_XDP Socket` (aka `xsk`) support for dnsdist's UDP client

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

---

In order to maximize the advantages of `xsk`, I have tried to avoid the copying of `frame` to `PacketBuffer`. But this attempt was unsuccessful, because `PacketBuffer` is required by so much code that it is very difficult to change the implementation of `PacketBuffer`.


Y7n05h:
This pr is currently just used to show the idea of my implementation. Adding `xsk` support to dnsdist would be quite complicated, I hope anyone can give me some good advice, if possible.

pdns/dnsdist.hh
```diff
@@ -50,6 +50,7 @@
 #include "uuid-utils.hh"
 #include "proxy-protocol.hh"
 #include "stat_t.hh"
+#include "xsk.hh"
```
rgacogne, line 53: It might make sense not to include this file in dnsdist.hh, and to instead do a forward declaration of class XskSocket; before the `ClientState` definition.

pdns/dnsdistdist/xsk.cc
```diff
@@ -0,0 +1,418 @@
+/*
```
rgacogne, line 1: We need to add xsk.cc and xsk.hh to `dnsdist_SOURCES` in Makefile.am. We will worry later about what happens if XDP/libxdp/... is not available.

pdns/dnsdist.cc
```diff
@@ -1670,7 +1828,24 @@ static void MultipleMessagesUDPClientThread(ClientState* cs, LocalHolders& holde
 }
 #endif /* defined(HAVE_RECVMMSG) && defined(HAVE_SENDMMSG) && defined(MSG_WAITFORONE) */
 #endif /* DISABLE_RECVMMSG */
-
+static void xskClientThread(ClientState* cs)
```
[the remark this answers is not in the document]
Y7n05h: Yes, but not exactly right. My current idea is to have `xskClientThread` replace `udpClientThread` and `tcpClientThread` when `XDP` is enabled. The work on the `tcpClientThrea` part here has not started yet. This is just a current idea, and may change in the future.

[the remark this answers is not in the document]
rgacogne: Understood! I wonder if that makes sense to support XDP for TCP in our case, do you know if the performance improvements are significant?

pdns/dnsdist.cc
```diff
@@ -1584,6 +1598,150 @@ static void processUDPQuery(ClientState& cs, LocalHolders& holders, const struct
...
+  const ComboAddress& dest = packet.getToAddr();
+  ComboAddress proxiedRemote = remote;
+  ComboAddress proxiedDestination = dest;
+
+  try {
+    bool expectProxyProtocol = false;
+    if (!isXskQueryAcceptable(packet, cs, holders, expectProxyProtocol)) {
+      return;
+    }
+    auto query = packet.clonetoPacketBuffer();
```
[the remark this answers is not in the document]
Y7n05h: I don't really agree. It seems to me that an AF_XDP Socket would correspond to a queue of NIC devices. Sending packets for different protocols from the same NIC queue to a different AF_XDP Socket, while feasible, doesn't make much sense. So our `xskClientThread` needs to support all types of network protocols. So refactoring `processUDPQuery()` is probably not a good practice.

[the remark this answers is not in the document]
rgacogne: Right, I missed that but it makes sense. Would that mean that we cannot get DoT or DoH traffic on that NIC queue?

pdns/dnsdist.cc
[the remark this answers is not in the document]
rgacogne: No worry, especially since I now understand that handling UDP but not TCP would likely be hard, so it looks like we will have to handle XDP for TCP anyway.

pdns/dnsdist.cc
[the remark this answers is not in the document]
rgacogne: Sounds good, thanks!

Y7n05h:
Can someone tell me what `cmsghdr` is used for? In my understanding, `cmsghdr` can only be used for `unix domain sockets`, not for the Internet. But in dnsdist, it appears in multiple scenarios related to `tcp` and `udp`. Am I understanding something wrong?

https://github.com/PowerDNS/pdns/blob/5744e93f8726155dc477871521afcfcf998df212/pdns/dnsdist.cc#L1008-L1039

https://github.com/PowerDNS/pdns/blob/5744e93f8726155dc477871521afcfcf998df212/pdns/dnsdist.cc#L1434

Y7n05h:
> Yes, `sendmsg` and `recvmsg` can be used with `tcp` and `udp`. One purpose is to supply an `iovec` array to send or receive using multiple (non-contiguous) buffers. Another use is to set or get various associated parameters, like the source address used for outgoing UDP requests. That is what happening in this case. See e.g. http://man.openbsd.org/ip.4 for some other uses.

Thanks for the quick reply. 
I fully understand being able to use `iovec` in `tcp` `udp` via `sendmsg` and `recvmsg` and being able to get source address information with them, but what I'm wondering is whether `msghdr.msg_control` has any practical meaning? Because they don't appear in the Ethernet payload.

rgacogne:
> what I'm wondering is whether msghdr.msg_control has any practical meaning? Because they don't appear in the Ethernet payload.

We use the header to communicate with the kernel, mostly for the purpose of getting or setting the source address, as you have already found out. So that information doesn't end up in the Ethernet payload directly, but it does influence what the kernel is doing with our datagram.

Y7n05h:
> We use the header to communicate with the kernel, mostly for the purpose of getting or setting the source address, as you have already found out. So that information doesn't end up in the Ethernet payload directly, but it does influence what the kernel is doing with our datagram.

Okay, I understand. Thank you very much.

Y7n05h:
One thing I might want to clarify: I used and partially modified some of the code from https://github.com/google/packetdrill/blob/master/gtests/net/packetdrill/packet_checksum.c . 
https://github.com/PowerDNS/pdns/blob/a8a5943a748e70abb96830b1c3a92ef773adf752/pdns/dnsdistdist/xsk.cc#L492-L614
They are under GPLv2 or later. Dnsdist is GPLv3 or later.So they are compatible.

rgacogne:
> I've now got xsk support for UDP mostly done, but all the code has not been tested yet and I will finish testing it in the next few days.

That's awesome! Let me know when you think this is ready for a review and I'll try to do it quickly. It does not have to be 100% finished at this point, to be clear.

Y7n05h:
> That's awesome! Let me know when you think this is ready for a review and I'll try to do it quickly. It does not have to be 100% finished at this point, to be clear.

This pr has a little bit of work left before it gets reviewed, and I'll finish it before I start tomorrow. But before that, could you please review #11526 first? The changes I made to the BPF program are based on #11526 .

pdns/dnsdistdist/xsk.cc
Y7n05h: I have implemented the compile argument `-with-xsk` to decide whether to use `xsk` or not. But it didn't complete the decision to link `libxdp` based on the presence or absence of `-with-xsk`.

rgacogne:
Can you show us the Lua code you used to get that result? The `LuaWrapper` class we use is not always easy to understand, but we should be able to help you with that.

Y7n05h:
> Can you show us the Lua code you used to get that result? The `LuaWrapper` class we use is not always easy to understand, but we should be able to help you with that.

I have not changed any Lua files. All the files I used for testing (except for the dnsdist config file) I have committed. I just `print` the type information for `(*vars)["xskSocket"]` and found them to be inconsistent with what I expected.
https://github.com/PowerDNS/pdns/blob/9cc6fe80f0ad6fb88bb69fb146fd35f3eedfa6c7/pdns/dnsdist-lua.cc#L146 


My config file is similar to below.

```
addCapabilitiesToRetain({"CAP_NET_BIND_SERVICE","CAP_BPF","CAP_NET_RAW","CAP_SYS_ADMIN"})
xsk=newXsk({NIC_queue_id=0,frameNums=65536,ifName="wlan0"})
newServer({address="1.1.1.1",MACAddr="aa:aa:aa:aa:aa:aa",xskSocket=xsk})
newServer({address="1.0.0.1",MACAddr="aa:aa:aa:aa:aa:aa",xskSocket=xsk})
addACL("192.168.30.0/24")
setServerPolicy(firstAvailable)
```

I want to pass a `std::shared_ptr<XskSocket>` via `xskSocket=xsk`.

rgacogne:
> My config file is similar to below.

Thanks, that's what I meant since our configuration is written in Lua :) I'll do some tests in a couple hours.

Y7n05h:
By the way, what is the preferred way to handle TCP packets? 

- One option is to use a userspace network stack, e.g. [mTCP](https://github.com/mtcp-stack/mtcp), [f-stack](https://github.com/F-Stack/f-stack), something like that. After bypassing the kernel network stack, using a userspace network stack seems to be a flow on approach among DPDK users. 
- Or we don't use the userspace network stack and just manually parse TCP packets and selectively support ***a subset*** of TCP. (`knot` uses this method. )

Implementing the full TCP protocol manually is very difficult, and testing the correctness of the implementation may be more difficult than implementing it.

Oh yeah, and the two options above are not just for TCP, the above discussion also applies to ip options and ipv6 Extension headers, which I don't support in the current code.

Y7n05h:
> Do you have any rough idea of how much work each option represents? I don't mind adding another dependency for `AF_XDP` support, provided that the license is compatible with our GPLv2, so if it gives us better TCP support and is easier that's fine by me. On the other hand manually handling a subset of TCP is also fine, and we can add better support via a library later if that makes sense.

I'm not quite sure yet. I am trying both scenarios separately. I may be a few days away from getting the answer. But maybe don't worry about the license issue, both mTCP and f-stack use the BSD license. (I remember BSD is compatible with GPLv2. But I'm not very sure about this, because there are too many variants of BSD)

rgacogne:
> > My config file is similar to below.
> 
> Thanks, that's what I meant since our configuration is written in Lua :) I'll do some tests in a couple hours.

I can reproduce the issue. With the config you provided it happens on newServer(), and we get a `DownstreamState::checkfunc_t` instead of a `std::shared_ptr<XskSocket>` which seems to be a LuaWrapper bug (or limitation). I'll keep digging.

rgacogne:
So the bad news is that I don't understand why Lua thinks that our custom `shared_ptr<XskSocket>` is a function (`lua_isfunction()` return true), and it messes up the type deduction. 

The good news is that you can just work around that by defining your type before the function type, so that the type deduction tries it first:
```diff
-  typedef LuaAssociativeTable<boost::variant<bool, std::string, LuaArray<std::string>, DownstreamState::checkfunc_t,std::shared_ptr<XskSocket>>> newserver_t;
+  typedef LuaAssociativeTable<boost::variant<bool, std::string, LuaArray<std::string>, std::shared_ptr<XskSocket>, DownstreamState::checkfunc_t>> newserver_t;
```
and it should work well enough for now.

Y7n05h:
I have tested dnsdist on my LAN and it works fine so far. Here is the config file I used in my testing :

```lua
addCapabilitiesToRetain({"CAP_NET_BIND_SERVICE","CAP_BPF","CAP_NET_RAW","CAP_SYS_ADMIN"})
xsk=newXsk({NIC_queue_id=0,frameNums=65536,ifName="wlan0",xskMapPath="/sys/fs/bpf/dnsdist/xskmap"})
newServer({address="192.168.30.170",MACAddr="e0:d4:e8:50:bd:0b",xskSocket=xsk,source="192.168.30.211:5301"})
setLocal("192.168.30.211:5300",{xskSocket=xsk})
addACL("192.168.30.211/24")
setServerPolicy(firstAvailable)
```

I will write documentation after my design is accepted. So I may now need to explain the contents of this config file to make it easier to review my code. In order to use AF_XDP, you must fill in `source` in newServer, and `source` cannot be `0.0.0.0`. Since it's cumbersome to handle arp requests manually, I didn't implement it. So `MACAddr` is also a required field. Its value is usually the MAC address of the gateway of the machine where you are running dnsdist.

Don't forget to add all `ports` used in `AF_XDP` in `xdp-filter.ebpf.src`, such as:

```
#define IN_DNS_PORT_SET(x) (COMPARE_PORT(x, 5300)||COMPARE_PORT(x, 5301)||COMPARE_PORT(x, 5302)||COMPARE_PORT(x, 5303))
```

Please remember not to use a port already used in AF_XDP in another application, that will cause that application to not work properly.

Y7n05h:
> I guess it will be the MAC address of the server if it is on the local network?
Yes.

> I wonder if we could use an eBPF map to make that more dynamic?

Yes we could do it, but I'm not sure it's worth it. In my opinion there won't be a lot of ports, and it might be better performance to write them in the code. Of course, if you think it's necessary, I'd love to do it.

rgacogne:
> Yes we could do it, but I'm not sure it's worth it. In my opinion there won't be a lot of ports, and it might be better performance to write them in the code. Of course, if you think it's necessary, I'd love to do it.

Understood! I'm not sure it's worth it either, so let's keep it as it is for now.

Y7n05h:
> > Do you have any rough idea of how much work each option represents? I don't mind adding another dependency for `AF_XDP` support, provided that the license is compatible with our GPLv2, so if it gives us better TCP support and is easier that's fine by me. On the other hand manually handling a subset of TCP is also fine, and we can add better support via a library later if that makes sense.
> 
> I'm not quite sure yet. I am trying both scenarios separately. I may be a few days away from getting the answer. But maybe don't worry about the license issue, both mTCP and f-stack use the BSD license. (I remember BSD is compatible with GPLv2. But I'm not very sure about this, because there are too many variants of BSD)

These days, I've tried both scenarios separately. Unfortunately `AF_XDP` is currently not easy to combine with userspace network stacks like `f-stack` or `mTCP`. They support `DPDK` well, but there may be a lot of work to be done to get `f-stack` or `mTCP` to work with `AF_XDP`. So I prefer to implement a subset of TCP manually. `knot` is implemented like this. At least I can get some experience and inspiration from `knot`.

rgacogne:
> These days, I've tried both scenarios separately. Unfortunately `AF_XDP` is currently not easy to combine with userspace network stacks like `f-stack` or `mTCP`. They support `DPDK` well, but there may be a lot of work to be done to get `f-stack` or `mTCP` to work with `AF_XDP`. So I prefer to implement a subset of TCP manually. `knot` is implemented like this. At least I can get some experience and inspiration from `knot`.

That seems like the right thing to do :+1:

pdns/dnsdist-idstate.hh
```diff
@@ -271,4 +272,5 @@ struct IDState
   bool destHarvested{false}; // if true, origDest holds the original dest addr, otherwise the listening addr
   bool dnssecOK{false};
   bool useZeroScope{false};
+  std::unique_ptr<PacketBuffer> xskPacketHeader;
```
rgacogne, line 275: Just a nit but this should go with the other unique_ptr to prevent losing space due to padding.

pdns/dnsdist.cc
```diff
@@ -577,146 +578,111 @@ void handleResponseSent(const IDState& ids, double udiff, const ComboAddress& cl
...
+      auto xskInfo = dss->xskInfo;
+      const auto xskFd = xskInfo->workerWaker;
+      while (!dss->isStopped()) {
+        while (!xskInfo->cq.read_available()) {
+          xskInfo->waitForXskSocket();
         }
+        xskInfo->cq.consume_all([&](XskPacket& packet) {
+          if (packet.dataLen() < sizeof(dnsheader)) {
+            xskInfo->sq.push(XskDelayPacketInfo(packet, xskInfo->frameOffset(packet)));
+            return;
```
rgacogne, line 601: return would terminate the thread, that does not seem right?

pdns/dnsdist.cc
```diff
@@ -577,146 +578,111 @@ void handleResponseSent(const IDState& ids, double udiff, const ComboAddress& cl
...
-             - we already got a response for this query and this one is a duplicate.
-             Either way, we don't touch it.
-          */
-          continue;
+    setThreadName("dnsdist/respond");
+    auto localRespRuleActions = g_respruleactions.getLocal();
+
+#ifdef HAVE_XSK
+    if (dss->xskInfo) {
+      dss->setUp();
```
rgacogne, line 591: Why is `setUp()` called here? Are the regular health-check mechanisms not working in that case?

pdns/dnsdist.cc
```diff
@@ -2661,7 +2901,23 @@ int main(int argc, char** argv)
     }
     handleQueuedHealthChecks(*mplexer, true);
 
+    for (auto i : g_xsk) {
```
rgacogne, line 2904: Likely needs to be guarded by `#ifdef HAVE_XSK` as well?

pdns/dnsdistdist/xsk.cc
```diff
@@ -0,0 +1,747 @@
...
+#include <sys/socket.h>
+#include <vector>
+
+extern "C"
+{
+#include <xdp/libxdp.h>
+}
+
+#ifdef HAVE_XSK
+constexpr bool XskSocket::is_pow_of_two(uint32_t value) noexcept
```
rgacogne, line 55: Is that actually used?

rgacogne (commented):
I have not reviewed everything yet, and I will continue tomorrow. So far I have not seen any issue, I have commented a few minor things (and I believe we can get rid of some of the code duplication in the incoming UDP path but we can deal with that later), and the general design looks good. Nice work!

pdns/dnsdist.cc
Y7n05h: Oh yes.  Currently AF_XDP is wrong to work with health checks.  Because AF_XDP will intercept all packets and make the health check timeout.  I called setup() to avoid the trouble of health checks when debugging.  I forgot that this place wasn't handled properly.  I will fix it today.

pdns/dnsdist.cc
Y7n05h: The return here is just returning from the lambda function.

rgacogne: Oh, right, I don't know how I missed that, sorry!

pdns/dnsdist-idstate.hh
Y7n05h: Done.

pdns/dnsdist.cc
Y7n05h: Now xsk works properly with healthcheck.

pdns/dnsdist.cc
Y7n05h: Done.

pdns/dnsdistdist/xsk.cc
Y7n05h: The call to this function was missed earlier. Fixed now.

rgacogne:
Oh, that sounds very nice! I guess we would need to start the "regular" threads in addition to the AD_XDP ones, right? If so, that's perfectly fine.

Y7n05h:
> Oh, that sounds very nice! I guess we would need to start the "regular" threads in addition to the AD_XDP ones, right? If so, that's perfectly fine.



> Oh, that sounds very nice! I guess we would need to start the "regular" threads in addition to the AD_XDP ones, right? If so, that's perfectly fine.

Yes. I think we just need to implement fallback for the client. Because it is uncertain which features of tcp the client uses, but the server was manually added by the administrator and enabled the AF_XDP feature, so if the server cannot communicate with dnsdist using xsk, it is an error, and fallback should not be used.

pdns/dnsdist.cc
```diff
@@ -1602,6 +1748,151 @@ static void processUDPQuery(ClientState& cs, LocalHolders& holders, const struct
...
+  const auto& dest = packet.getToAddr();
+  ComboAddress proxiedRemote = remote;
+  ComboAddress proxiedDestination = dest;
+
+  try {
+    bool expectProxyProtocol = false;
+    if (!isXskQueryAcceptable(packet, cs, holders, expectProxyProtocol)) {
+      return;
+    }
+    auto query = packet.clonePacketBuffer();
```
rgacogne, line 1765: I think we could avoid a fair amount of the code duplication below that line, but that would require a bit of refactoring. I can do that later.

pdns/dnsdist.cc
```diff
@@ -1868,13 +2177,12 @@ static void healthChecksThread()
...
-      dss->prev.queries.store(dss->queries.load());
-      dss->prev.reuseds.store(dss->reuseds.load());
-
-      dss->handleTimeouts();
+#ifdef HAVE_XSK
+      if (dss->xskInfo) {
+        continue;
+      }
+#endif /* HAVE_XSK */
+      dss->updateStatisticsInfo();
```
rgacogne, line 2185: I think it might make sense to keep `handleTimeouts()` separated from `updateStatisticsInfo()`. Or at least it would need a better name, because it was not clear to me that the timeouts would be handled in that function.

pdns/dnsdistdist/xsk.cc
```diff
@@ -0,0 +1,806 @@
...
+
+std::shared_ptr<XskExtraInfo> XskExtraInfo::create()
+{
+  return std::make_shared<XskExtraInfo>();
+}
+void XskSocket::addWorker(std::shared_ptr<XskExtraInfo> s, uint16_t port, bool isTCP)
+{
+  extern std::atomic<bool> g_configurationDone;
+  if (g_configurationDone) {
+    throw runtime_error("Add a server with xsk at runtime is not supported");
```
rgacogne, line 697:
```suggestion
    throw runtime_error("Adding a server with xsk at runtime is not supported");
```

pdns/dnsdistdist/xsk.hh
```diff
@@ -0,0 +1,237 @@
...
+  friend XskSocket;
+  friend XskExtraInfo;
+  friend XskDelayPacketInfo;
+
+  enum Flags : uint32_t
+  {
+    TCP = 1 << 0,
+    UPDATE = 1 << 1,
+    DELAY = 1 << 3,
+    REWIRTE = 1 << 4
```
rgacogne, line 149:
```suggestion
    REWRITE = 1 << 4
```

pdns/dnsdist.cc
```diff
@@ -2763,3 +3096,59 @@ int main(int argc, char** argv)
 #endif
   }
 }
```
rgacogne, line 3098: It took me a while to understand what exactly the router does, and how it interacts with the other components, so perhaps a few comments about the general XSK design/architecture would help?

rgacogne:
I see that this is now conflicted because we merged other changes in the meantime. Do not worry about rebasing it, I can do that later so you can focus on this PR.

pdns/dnsdist.hh
```diff
@@ -752,6 +754,8 @@ struct DownstreamState: public std::enable_shared_from_this<DownstreamState>
     bool d_tcpCheck{false};
     bool d_tcpOnly{false};
     bool d_addXForwardedHeaders{false}; // for DoH backends
+    MACAddr sourceMACAddr;
+    MACAddr destMACAddr;
```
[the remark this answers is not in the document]
Y7n05h: Are you perhaps misreading the content? There are no shared pointers in `DownstreamState::Config`. you probably meant to say `string` ?

pdns/dnsdist-lua-bindings.cc
```diff
@@ -688,7 +689,49 @@ void setupLuaBindings(LuaContext& luaCtx, bool client)
       }
     });
 #endif /* HAVE_EBPF */
-
+#ifdef HAVE_XSK
+  using xskopt_t = LuaAssociativeTable<boost::variant<bool, uint32_t, std::string>>;
```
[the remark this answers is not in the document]
Y7n05h: Oh, `bool` is not needed for now.

pdns/dnsdist-lua-bindings.cc
```diff
@@ -688,7 +689,49 @@ void setupLuaBindings(LuaContext& luaCtx, bool client)
...
+      throw std::runtime_error("ifName field is required!");
+    }
+    if (opts.count("xskMapPath") == 1) {
+      path = boost::get<std::string>(opts.at("xskMapPath"));
+    }
+    else {
+      throw std::runtime_error("bpfProgPath field is required!");
+    }
+    extern std::vector<std::shared_ptr<XskSocket>> g_xsk;
+    auto socket = std::make_shared<XskSocket>(frameNums, 2048, ifName, queue_id, path);
```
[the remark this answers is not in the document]
Y7n05h: You are right. I was initially concerned that there might be a need to create multiple AF_XDP sockets with different frame sizes in dnsdist, and it's easier to scale when such a need arises. But it does seem a bit over-engineered.

pdns/dnsdist.cc
Y7n05h: Ok, no problem. I added some comments.
The design here is indeed a bit complicated.
Since `boost::lockfree::spsc_queue` does not accept `unique_ptr<XskPacket>` as a parameter (`unique_ptr` cannot be copy constructed), I made `XskPacket` as the parameter of `boost::lockfree::spsc_queue`. But I am worried about the copying overhead of XskPacket, so XskFrameInfo and XskDelayPacketInfo are constructed instead of XskPacket in many cases to reduce the copying overhead.
The presence of `XskFrameInfo` and `XskDelayPacketInfo` adds a lot of complexity here.
I now use `XskPacket*` as an element in boost::lockfree::spsc_queue in my refactoring. In this way, the overhead of copying is eliminated, and `XskFrameInfo` and `XskDelayPacketInfo` are also eliminated. I also changed the name of `XskExtraInfo` to `XskWorker`, which may be more appropriate.
The code here is probably better understood now.
I also added some comments, but I'm not sure if my comments are enough. I can add more comments if needed.

rgacogne:
> Used instead of specifying the MAC address in the configuration file. But this function only works properly when the DNS Server and the machine where dnsdist is deployed are in the same LAN, otherwise it cannot get the MAC address of the gateway as I expected. So I don't use it.

Right, that function is only meant to provide MAC for devices on the local network indeed. We should clearly document that, I'll do that later.

pdns/dnsdist.cc
rgacogne: Thanks a lot, I'll have a look and do a new review tomorrow!

pdns/dnsdistdist/xsk.hh
```diff
@@ -0,0 +1,232 @@
...
+// We use an XskSocket to manage an AF_XDP Socket corresponding to a NIC queue.
+// The XDP program running in the kernel redirects the data to the XskSocket in userspace.
+// XskSocket routes packets to multiple worker threads registered on XskSocket via XskSocket::addWorker based on the destination port number of the packet.
+// The kernel and the worker thread holding XskWorker will wake up the XskSocket through XskFd and the Eventfd corresponding to each worker thread, respectively.
+class XskSocket
+{
+  struct XskRouteInfo
+  {
+    __be16 port;
+    int xskSocketWaker;
```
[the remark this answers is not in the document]
Y7n05h: No, we can't do that.
`XskRouteInfo` does not take ownership of `xskSocketWaker`.
`XskRouteInfo::xskSocketWaker` is just a copy of `XskRouteInfo::worker` pointing to `XskWorker::xskSocketWaker` so that `multi_index_container` can use it as an index.
`XskWorker` takes ownership of `XskWorker::xskSocketWaker`.
I'm sure `XskRouteInfo::xskSocketWaker` always works because of `XskRouteInfo::worker` . Closing `XskWorker::xskSocketWaker` is the responsibility of `XskWorker`.

However, I should have used `FDWrapper` instead of `int` in `XskWorker`, I will correct now.

pdns/dnsdistdist/xsk.cc
```diff
@@ -0,0 +1,819 @@
...
+  memset(&rx, 0, sizeof(rx));
+  xsk_umem_config umemCfg;
+  umemCfg.fill_size = fqCapacity;
+  umemCfg.comp_size = cqCapacity;
+  umemCfg.frame_size = frameSize;
+  umemCfg.frame_headroom = XSK_UMEM__DEFAULT_FRAME_HEADROOM;
+  umemCfg.flags = 0;
+  auto ret = xsk_umem__create(&umem, bufBase, frameSize * frameNum, &fq, &cq, &umemCfg);
+  if (ret != 0) {
+    throw std::runtime_error("Error creating a umem of size" + std::to_string(frameSize * frameNum) + stringerror(ret));
```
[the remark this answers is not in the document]
Y7n05h: Yes, it was necessary. It has now been fixed.

Y7n05h:
@omoerbeek, thanks for your suggestion. Since there are many suggestions, I need some time to reply to each item separately.

> Another question that popped up while reading PR: I don't understand why the XSK code would need its own health-check function. I would assume the health-checking is done as before, and the result of the health-checks are used to manipulate data used by the XDP program.

There is no abstraction of file descriptors for using `AF_XDP` to accomplish network IO compared to other methods. So `read`/`write`/`sendmsg`/`recv`/`epoll` these system calls will not work on `AF_XDP`. AF_XDP-based network IO is far more complex than socket-based, so it may not be a good choice to combine these two very different IO methods.

The most important thing is that the data of `AF_XDP` cannot be read directly, only `XskSocket` can read the data in it, and `XskRouter` distributes data packets to multiple `XskWorker` according to the destination IP address and port. The responses of health check and non-health check have the same ip and port, so `XskRouter` can't distinguish between health check and normal response, so it can't distribute them differently.

Y7n05h:
> * It is not clear to me what parts of dnsdists' backend selection decisions are done by XDP, and which parts are still done by userland.

It took me some time to confirm that the backend selection for XDP was incorrect.
I reused the code from the previous udp backend selection without making any changes here. But this is not correct.
 
`XskClient` should only choose to use `DownstreamState` that uses the same `XskSocket`. It may also be necessary to allow a `DownstreamState` or `ClientState` to use more than one `XskSocket` at the same time. The modification here may not be difficult, but the `AF_XDP` related code may be more complicated.
I didn't notice this problem before because I only tested all `DownstreamState` with AF_XDP.

Y7n05h:
> If all packets are routed to userland I wonder what the role of the XDP program is compared to the already existing bpf filter functionality.

The original XDP-based filter still works. Not all packets are redirected to userspace. Only packets whose destination port is a DNSDist-related port and are not blocked by the rules in `BPFFilter` will be redirected to userspace.

AF_XDP does not bring more functionality to dnsdist. AF_XDP, a brand new asynchronous network interface, is used for performance reasons. AF_XDP is similar to `DPDK`, bypassing the kernel's network protocol stack and avoiding the overhead of system calls. There is a high probability that the performance of dnsdist can be improved.

Y7n05h:
I drew two diagrams to explain the design idea of the XDP part. Please let me know if you need more additional information.


xdp

Due to the size limitation of the first picture, I cannot fully show the relationship between XskSocket and XskWorker in the first picture, so I put them in the second picture.

XskWorker

Y7n05h:
> This is very useful, thank you so much! Do you think you could look into integrating these into the documentation? It's in the pdns/dnsdistdist/docs directory.

I would love to do this. 
But the code review for this pr is not finished yet. I'm concerned that anyone's comments on pr lead to changes to the code. In order to reduce useless documentation modification work, I think it is better to add documentation at the end of the code review for this pr.

Y7n05h:
I did some performance tests on this pr of mine. Some bugs were found during performance testing and have now been fixed in the 4 commits above.
In performance testing I found that AF_XDP effectively improves DNSDist throughput.

Y7n05h:
The results and process of performance testing I wrote on my blog.
https://blog.y7n05h.dev/speedUpDnsdistWithAFXDP-en/

rgacogne:
There are great news, thanks! I'll do a more thorough review in the next coming days! @omoerbeek might be interested as well :)

rgacogne:
The number of queries per second (less than 2000 qps) reached in your test environment seems very low, do you have any idea why? On hardware such as PC1 I usually get around 80k qps per core, assuming that the network can keep up (which might be an issue with WiFi). The number of queries lost is also worrying, do you have any idea what is happening? Looking at the metrics via `dumpStats()` (in the foreground or via the console), the prometheus endpoint or the REST API might give us some clues.

Y7n05h:
> The number of queries per second (less than 2000 qps) reached in your test environment seems very low, do you have any idea why? On hardware such as PC1 I usually get around 80k qps per core, assuming that the network can keep up (which might be an issue with WiFi). The number of queries lost is also worrying, do you have any idea what is happening? Looking at the metrics via `dumpStats()` (in the foreground or via the console), the prometheus endpoint or the REST API might give us some clues.

I'm guessing the error may be caused by the network being used by another devices or my test method is not correct.
I'm currently not at the location where I did the test 2 days ago, and there isn't enough devices for me to repeat the test, so I may not be able to give more information.

Y7n05h:
Why can't I reproduce the error in CI after 215a31c and b308246 changes. I don't understand why this happens in CI.

```
In file included from dnsdist-cache.cc:24:
./dnsdist.hh:53:10: fatal error: 'xsk.hh' file not found
#include "xsk.hh"
         ^~~~~~~~
1 error generated.
make[3]: *** [Makefile:3381: dnsdist-cache.o] Error 1
  CXX      dnsdist-ecs.o
  CXX      dnsdist-protocols.o
  CXX      fuzz_dnsdistcache.o
In file included from dnsdist-ecs.cc:23:
./dnsdist.hh:53:10: fatal error: 'xsk.hh' file not found
#include "xsk.hh"
         ^~~~~~~~
```

rgacogne:
The error in CI happens when building the fuzzing targets for dnsdist, which is done in the parent directory, so you probably need to add `xsk.cc` `xsk.hh` to the `fuzz_target_dnsdistcache_SOURCES` list in `pdns/Makefile.am`. I remember I initially set up the fuzzing targets in the parent directory because of the way OSS-Fuzz was working at the time, but I'll see if it can be improved.

Y7n05h:
> The error in CI happens when building the fuzzing targets for dnsdist, which is done in the parent directory, so you probably need to add `xsk.cc` `xsk.hh` to the `fuzz_target_dnsdistcache_SOURCES` list in `pdns/Makefile.am`. I remember I initially set up the fuzzing targets in the parent directory because of the way OSS-Fuzz was working at the time, but I'll see if it can be improved.

Seems to be caused by another problem.
In my local environment, I executed these instructions in sequence in the project root directory in the debain container, but I did not encounter the "cannot find `xsk.hh`" problem.
Maybe the problem is caused by ccache not clearing the cache in CI?

rgacogne:
I don't think there is any kind of caching at that step. I would usually suggest rebasing on master to be sure that this is not a disconnect between the PR branch and master, but in your case I'm afraid you would have to fix a conflict first, which might be painful. I'll see if I can reproduce the issue later today.

Y7n05h:
Great, I've modified it many times and now it's completely passed the CI.

rgacogne:
I finally found the time to look into this a bit more. Based on what I have read I believe we can write our kernel-side eBPF program in such a way that it does not handle TCP packets and let the regular kernel network stack handle them, by returning `XDP_PASS` instead of `XDP_REDIRECT`. I'll do some tests first, to see if it indeed works.

Y7n05h:
> we can write our kernel-side eBPF program in such a way that it does not handle TCP packets and let the regular kernel network stack handle them, by returning `XDP_PASS` instead of `XDP_REDIRECT`.

Yes, I'm sure it's fine, it works. The XDP program in pr returns `XDP_REDIRECT` for TCP packet only because of code left over from the original attempt to implement TCP parsing based on `AF_XDP`. It's not necessary for UDP, so we can remove it and just return `XDP_PASS`.

rgacogne:
I have now done a fair amount of testing on this new functionality, and first I have to say that you have done a really nice job, so thanks again!
I have experienced a few crashes, most of them caused by mis-configuration on my side but it would still be nice to handle these cases better. More on that later.
I see the following points that need to be addressed before we can merge this, and I will likely find a few more in the process:
- we need to rebase this pull request on the current master
- punt all TCP packets to the normal stack from the eBPF program. This will be painful, so I'll do it myself.
- handle large responses: we can easily get response payloads larger than the current frame size (`2048`), so we need to be able to forward these to the client, likely by falling back to regular `sendto()` calls
- it looks like we pass a copy of the DNS payload to the query/response processing functions, which might modify that copy, but we then discard the changes, unless I'm mistaken
- I'm currently worried we might be calling `xsk_ring_cons__release` too early, as we are still accessing data stored in the `umem` after that. It might be that I'm confused about the lifetime of that data, though.
- be able to handle packets received via `AF_XDP` but forwarded over "normal" UDP to the backend, as I don't think we can expect to be able to use `AF_XDP` toward all backends, especially since it limits the number of usable source ports to communicate with the backend to one
- I'm wondering if we could not avoid the complicated health-check logic by using a different source port for health-checks, and thus they could be punted to the "regular" stack
- the `#ifdef HAVE_XSK` guard should come a lot sooner in the `xsk.cc` and `xsk.hh` files, otherwise the compilation will always fail on non-Linux targets

I'll see how it goes but since we are almost ready to issue a beta of 1.8.0, I'm unfortunately considering delaying this PR to the 1.9 milestone. I expect 1.9 to be a lot smoother than 1.8, though, so it should not delay that feature by long.

Y7n05h:
Thanks for your review.
I tried to do the rebase, but since there are so many conflicts, it may take me some time to complete.

rgacogne:
>     * we need to rebase this pull request on the current master

It was indeed painful but I have a working branch. Would you mind if I force-push on your branch?
 
>     * punt all TCP packets to the normal stack from the eBPF program.

Done as well.

>     * it looks like we pass a copy of the DNS payload to the query/response processing functions, which might modify that copy, but we then discard the changes, unless I'm mistaken

I was mistaken, we do commit the changes.

>     * I'm currently worried we might be calling `xsk_ring_cons__release` too early, as we are still accessing data stored in the `umem` after that. It might be that I'm confused about the lifetime of that data, though.

I was indeed confused, we own the umem until we put it back to the fill queue, so it's fine.

>     * be able to handle packets received via `AF_XDP` but forwarded over "normal" UDP to the backend, as I don't think we can expect to be able to use `AF_XDP` toward all backends, especially since it limits the number of usable source ports to communicate with the backend to one

Done.

>     * the `#ifdef HAVE_XSK` guard should come a lot sooner in the `xsk.cc` and `xsk.hh` files, otherwise the compilation will always fail on non-Linux targets

Done.

>     * handle large responses: we can easily get response payloads larger than the current frame size (`2048`), so we need to be able to forward these to the client, likely by falling back to regular `sendto()` calls

Needs to be done.

>     * I'm wondering if we could not avoid the complicated health-check logic by using a different source port for health-checks, and thus they could be punted to the "regular" stack

I still need to look into that.

Y7n05h:
> It was indeed painful but I have a working branch. Would you mind if I force-push on your branch?

No problem at all.
I am sorry that I have been busy with some very important personal matters recently, which prevented me from immediately following up on your previous review comments. I expect to pick up work here by the end of next week.

rgacogne:
> * handle large responses: we can easily get response payloads larger than the current frame size (`2048`), so we need to be able to forward these to the client, likely by falling back to regular `sendto()` calls

Now done.

> No problem at all.
> I am sorry that I have been busy with some very important personal matters recently, which prevented me from immediately following up on your previous review comments. I expect to pick up work here by the end of next week.

Don't worry, I'm the one that took a very long time to review your work! I need to do a few more tests but I'll push my branch here on Monday, and the few comments I have. Thanks again!

Y7n05h:
> handle large responses: we can easily get response payloads larger than the current frame size (`2048`), so we need to be able to forward these to the client, likely by falling back to regular `sendto()` calls

I added a parameter to allow users to customize the size of frame when creating `XskSocket`.

rgacogne:
> > handle large responses: we can easily get response payloads larger than the current frame size (`2048`), so we need to be able to forward these to the client, likely by falling back to regular `sendto()` calls
> 
> I added a parameter to allow users to customize the size of frame when creating `XskSocket`.

Nice, thanks! This will be helpful to be able to receive large queries, and to be able to process bigger responses. The current code should also properly fallback to sending the responses without using `XSK` if it is larger than the available space (in theory responses can be up to 65535 bytes, but over UDP 4096 bytes is a good bet).

rgacogne:
@Y7n05h How do you feel about the idea to use a map to let the `XDP` program know which IP address/port tuple we want to pass to the `XSK socket? The reason I suggest that is that it would allow using any port to contact the backend via `AF_XDP`, which is good to avoid spoofing attempts and would also allow the health-checks to be done via a regular, non `XSK` socket.

Y7n05h:
> How do you feel about the idea to use a map to let the `XDP` program know which IP address/port tuple we want to pass to the `XSK socket? 

Do you mean that we only pass the request where the source ip and source port are in the map to xsk?  

> The reason I suggest that is that it would allow using any port to contact the backend via `AF_XDP`, which is good to avoid spoofing attempts and would also allow the health-checks to be done via a regular, non `XSK` socket.

Oh yes, this approach does have some benefits.  And it is not difficult to implement this feature.  My concern is that the number of entries in the map is limited and we cannot store a large number of tuples.

rgacogne:
> > How do you feel about the idea to use a map to let the `XDP` program know which IP address/port tuple we want to pass to the `XSK socket?
> 
> Do you mean that we only pass the request where the source ip and source port are in the map to xsk?

Not exactly, I ad in mind to pass the request to xsk if the destination IP and destination port are in the map. Since we use a different port the health-checks, the responses would be passed to the non-xsk flow which would keep the health-check code much simpler.

rgacogne:
> Not exactly, I had in mind to pass the request to xsk if the destination IP and destination port are in the map. Since we use a different port the health-checks, the responses would be passed to the non-xsk flow which would keep the health-check code much simpler.

Would you mind if I pushed these changes to your branch? I'd like to finally get this merged for 1.9 :)

rgacogne:
I just pushed a rebased version of this branch, plus a fair number of commits to address some issues I found during testing. Once CI is green, I think this PR is ready for a review (which I'm no longer qualified to provide, given I have done significant work on this PR).
I plan to get this merged before 1.9.0-rc1.

rgacogne:
Test run of packages building at https://github.com/rgacogne/pdns/actions/runs/7653905496 As expected `AF_XDP`/`XSK` support is enabled for el-8, el-9 and Debian Bookworm.

rgacogne:
Reminder to self: I checked and all the commits on this branch compile!

rgacogne:
I pushed a commit adding a new fuzzing target for the XSK code.

rgacogne:
Ah, the "is there a binary present in the pull request" check is working, but this one is in the `fuzzing/corpus` directory so it's fine, I guess I'll have to improve the check :)

rgacogne:
> Ah, the "is there a binary present in the pull request" check is working, but this one is in the `fuzzing/corpus` directory so it's fine, I guess I'll have to improve the check :)

https://github.com/PowerDNS/pdns/pull/13749

pdns/dnsdistdist/docs/advanced/xsk.rst
```diff
@@ -0,0 +1,112 @@
...
+In addition to these, an ``eBPF`` ``XDP`` program needs to be loaded to decide which packets to distribute via the ``AF_XDP`` socket (and to which, as there are usually more than one). This program uses a ``BPF`` map of type ``XSKMAP`` (located at ``/sys/fs/bpf/dnsdist/xskmap`` by default) that is populated by :program:``dnsdist` at startup to locate the ``AF_XDP`` socket to use. :program:`dnsdist` also sets up two additional ``BPF`` maps (located at ``/sys/fs/bpf/dnsdist/xsk-destinations-v4`` and ``/sys/fs/bpf/dnsdist/xsk-destinations-v6``) to let the ``XDP`` program know which IP destinations are to be routed to the ``AF_XDP`` sockets and which are to be passed to the regular network stack (health-checks queries and responses, for example). A ready-to-use `XDP program <https://github.com/PowerDNS/pdns/blob/master/contrib/xdp.py>`_ can be found in the ``contrib`` directory of the PowerDNS Git repository. The name of the network interface to use might have to be updated, though. Once it has been updated, the ``XDP`` program can be started::
+
+  $ python xdp.py
```
chbruyand: Maybe it would be nice to update the default interface name in `xdp.py` to reflect the examples bellow (`enp1s0` instead of `eth1`).

chbruyand (approved):
lgtm, nothing jumped out at me.

pdns/xsk.hh
```diff
@@ -45,9 +38,14 @@
...
+#include <linux/types.h>
 
-#ifdef HAVE_XSK
 #include <xdp/xsk.h>
+
+#include "iputils.hh"
+#include "lock.hh"
+#include "misc.hh"
+#include "noinitvector.hh"
 #endif /* HAVE_XSK */
```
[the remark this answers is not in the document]
rgacogne: Done!

PR #11655, opened 2022-05-31
Change dns_tolower() and dns_toupper() to use a table

omoerbeek:
### Short description

I figured this micro-optimization should lead to a speedup as there are less conditionals, smaller code and the page containing the tables will be cached almost certainly. And indeed:

New speedtest:
```
    Before:
    'BurtleHashCI' 0.10 seconds: 66975796.8 runs/s, 0.01 usec/run
    
    After:
    'BurtleHashCI' 0.10 seconds: 96758136.6 runs/s, 0.01 usec/run
```
I'm also folding `ascii.hh` into `dnsname.{cc,hh}

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code


rgacogne (approved):
:+1:

PR #11668, opened 2022-06-01
dnsdist: Optionally send 'verbose' messages to a file, and log them at 'DEBUG' level otherwise

rgacogne:
### Short description

This pull request:
- sends 'verbose' message at syslog's 'debug' level instead of 'info'
- adds a new configuration option, `setVerboseLogDestination(dest)` which, when used at configuration time, redirects these messages to a file instead of sending to the standard output and/or syslog.

This is useful to be able to log these messages separately when investigating an issue, without flooding the system logs.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


chbruyand (approved):

Habbie (changes requested):
one nit, otherwise approved

Habbie (approved):

PR #11675, opened 2022-06-02
dnsdist: skip DoT/DoH frontend when a tls configuration error occurs

chbruyand:
### Short description
Currently, `dnsdist` exits when encountering an error in the DoT/DoH configuration (such as files for keys are not found or readable).

This PR is a proposal to skip the DoT/DoH frontend configuration and to continue launching anyway.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdist-lua.cc
+++ b/pdns/dnsdist-lua.cc
@@ -2501,6 +2501,19 @@ static void setupLuaConfig(LuaContext& luaCtx, bool client, bool configCheck)
       }
 
       parseTLSConfig(frontend->d_tlsConfig, "addDOHLocal", vars);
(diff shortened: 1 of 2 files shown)


rgacogne (commented):
The code looks good, thanks! I'm not sure we actually want this, though, as personally I don't expect dnsdist to start if something as important is messed up, at least not by default.
Then if we decide that we want that (optionally, maybe?), I'm worried that the frontend might still be stored in a Lua variable (the shared pointer returned `getBind()`), and I believe we also have a `TLSFrontend` in a weird state in `g_tlslocals` by then.

Issue #11393, opened 2022-03-04
Recursor must allow hostnames in forward-zone to be Docker / Docker Swarm compatible


jnehlmeier:
- Program: Authoritative, Recursor
 - Issue type: Feature request

### Description

When having a docker swarm cluster with several hosts you use an overlay network to let containers communicate across hosts. In combination with docker swarm ingress port publishing you can easily choose multiple docker swarm cluster hosts to act as nameserver and requests will always end up in the PowerDNS Recursor container, regardless of the hosts it is executed on.

So for example you can have `/etc/resolv.conf` with 
```
nameserver 10.1.1.1:53 # docker swarm host 1 of 5
nameserver 10.1.1.2:53 # docker swarm host 2 of 5
```
and have a recursor running once (executed on any of the 5 hosts).

When also having private DNS zones and deploying an authoritative server to docker swarm, the recursor needs an entry in its `forward-zone` configuration to point to that authoritative server for the private zones. In that configuration option only IP addresses are allowed.

Given that you do not know where the authoritative server will be executed by docker you need to use the docker service name of the authoritative server instead of an IP. That service name will point to a virtual IP provided by docker and let docker route the requests to the appropriate host/container.

So in short: Running recursor and authoritative inside docker swarm does effectively not work since you can not forward to the authoritative server as you do not know the IP. The only solution is to allow hostnames in addition to IPs in `forward-zone` which the docker embedded DNS server will resolve to the correct target IP.

jnehlmeier:
An additional solution would be to provide a docker image which runs recursor and authoritative server in the same container so that both can talk to each other via localhost.

Habbie:
Using DNS to run your DNS is a recipe for confusion and hard to debug problems. We recommend writing some tooling that generates/updates your forward-zones.

Habbie:
> Using DNS to run your DNS is a recipe for confusion and hard to debug problems. We recommend writing some tooling that generates/updates your forward-zones.

`dnsdist-resolver.*` in https://github.com/PowerDNS/pdns/tree/master/dockerdata might provide some inspiration - although the Recursor does not provide the same convenient Lua hooks.

omoerbeek:
One possible approach is to set forward zones using the REST API.

jnehlmeier:
> 

[...]

PR #11679, opened 2022-06-04
Add support for domains in also-notify

maxaudron:
### Short description
I am running a powerdns setup in Kubernetes with dynamical provisioned
secondaries which I want to notify. Kubernetes provides through it's
services domains, which can be used to discover all the available
secondaries.

This PR adds support to lookup and resolve all the IPs of a domain
using the already existing `ns.lookup()` function for also-notify.

I've been running this change in my setup for roughly half a
year now, mainly because I totally forgot about it...
I've not expired any problems thus far.

I'm pretty inexperienced with C++ so any feedback would be appreciated.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


Habbie:
hello there! apologies for leaving your PR lying around for so long, my first reaction was "well people shouldn't do that, but clearly some want to" and then we never got around to reviewing the actual code.

Meanwhile this question pops up semi-frequently. We won't have time to review/merge this for 4.9 (coming out next month I hope) but I hope we can take a good look for the next release.

If you have a version of your patch that applies to recent pdns versions, updating the PR would be great!

Habbie:
https://github.com/PowerDNS/pdns/issues/11393#issuecomment-1907829911 has some collected thoughts about similar work for the Recursor (where there are many more pitfalls than in Auth)

maxaudron:
I've rebased the PR to master, but i'm not able to test it anymore as i've migrated away from my previous setup.

PR #11682, opened 2022-06-07
Recursor: Load IPv6 entries from `etc-hosts` file

### LINKED ISSUE #2248: Recursor - Not loading IPv6 entries from etc-hosts-file.
The use case is simple.  I need to inject a couple of RRs internally to a zone that I am hosting externally. 

I created a new etc-hosts-file.  Added two entries in standard format.  One for v4, one for v6.  Enabled the export and restarted.  "host host.example.com" only returns the v4 address and "dig aaaa host.example.com" returns no v6 address.

Expected behavior is that the v6 address is returned on query as well.

fredmorcos:
### Short description
This adds the missing parts to load IPv6 entries from a provided `etc-hosts` file:
* Adds support for reversing IPv4 and IPv6 addresses to `ComboAddress`. Also adds tests for that.
* Replaces the usage of strings for IP addresses with `ComboAddress`.
* Moves a few things around for testability (in a separate commit).
* Also breaks down some parts into smaller functions (also in a separate commit).
* Fixes a pre-existing issue where `PTR` entries were being added for aliases (i.e. "non canonical hostname(s)").
* Adds tests for the pre-existing IPv4-only functionality in `test-reczones-helpers.cc`.
* Also adds tests for loading IPv6 entries as well.
* Adds the ability to print domain maps and their corresponding record entries to a `std::stringstream`.
* The usual misc cleanups & clang-tidy fixes.

Closes #2248 

PS: To run the tests with print output use `./testrunner --run_test=... --log_level=message`

### Checklist
I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/ext/lmdb-safe/lmdb-safe.hh
+++ b/ext/lmdb-safe/lmdb-safe.hh
@@ -1,4 +1,5 @@
 #pragma once
+#include <string_view>
 #include <lmdb.h>
 #include <iostream>
 #include <fstream>
@@ -17,14 +18,14 @@ using std::string_view;
 /* open issues:
  *
  * - missing convenience functions (string_view, string)
- */ 
+ */
 
 /*
 The error strategy. Anything that "should never happen" turns into an exception. But things like 'duplicate entry' or 'no such key' are for you to deal with.
  */
 
 /*
-  Thread safety: we are as safe as lmdb. You can talk to MDBEnv from as many threads as you want 
+  Thread safety: we are as safe as lmdb. You can talk to MDBEnv from as many threads as you want
 */
 
 /** MDBDbi is our only 'value type' object, as 1) a dbi is actually an integer
@@ -35,13 +36,13 @@ public:
   MDBDbi(): d_dbi(-1)
   {
   }
-  explicit MDBDbi(MDB_env* env, MDB_txn* txn, string_view dbname, int flags);  
+  explicit MDBDbi(MDB_env* env, MDB_txn* txn, string_view dbname, int flags);
 
   operator const MDB_dbi&() const
   {
     return d_dbi;
   }
-  
+
   MDB_dbi d_dbi;
 };
 
@@ -64,7 +65,7 @@ public:
   }
 
   MDBDbi openDB(const string_view dbname, int flags);
-  
+
   MDBRWTransaction getRWTransaction();
   MDBROTransaction getROTransaction();
 
@@ -106,7 +107,7 @@ struct MDBOutVal
     T ret;
     if(d_mdbval.mv_size != sizeof(T))
       throw std::runtime_error("MDB data has wrong length for type");
-    
+
     memcpy(&ret, d_mdbval.mv_data, sizeof(T));
     return ret;
   }
@@ -121,7 +122,7 @@ struct MDBOutVal
     T ret;
     if(d_mdbval.mv_size != sizeof(T))
       throw std::runtime_error("MDB data has wrong length for type");
-    
+
     memcpy(&ret, d_mdbval.mv_data, sizeof(T));
     return ret;
   }
@@ -131,11 +132,11 @@ struct MDBOutVal
   {
     if(d_mdbval.mv_size != sizeof(T))
       throw std::runtime_error("MDB data has wrong length for type");
-    
+
     return reinterpret_cast<const T*>(d_mdbval.mv_data);
   }
-  
-  
+
+
   MDB_val d_mdbval;
 };
 
@@ -159,7 +160,7 @@ public:
   template <class T,
             typename std::enable_if<std::is_arithmetic<T>::value,
                                     T>::type* = nullptr>
-  MDBInVal(T i) 
+  MDBInVal(T i)
   {
     memcpy(&d_memory[0], &i, sizeof(i));
     d_mdbval.mv_size = sizeof(T);
@@ -171,20 +172,20 @@ public:
     d_mdbval.mv_size = strlen(s);
     d_mdbval.mv_data = (void*)s;
   }
-  
-  MDBInVal(const string_view& v) 
+
+  MDBInVal(const string_view& v)
   {
     d_mdbval.mv_size = v.size();
     d_mdbval.mv_data = (void*)&v[0];
   }
 
-  MDBInVal(const std::string& v) 
+  MDBInVal(const std::string& v)
   {
     d_mdbval.mv_size = v.size();
     d_mdbval.mv_data = (void*)&v[0];
   }
 
-  
+
   template<typename T>
   static MDBInVal fromStruct(const T& t)
   {
@@ -193,7 +194,7 @@ public:
     ret.d_mdbval.mv_data = (void*)&t;
     return ret;
   }
-  
+
   operator MDB_val&()
   {
     return d_mdbval;
@@ -250,7 +251,7 @@ public:
                      const_cast<MDB_val*>(&val.d_mdbval));
     if(rc && rc != MDB_NOTFOUND)
       throw std::runtime_error("getting data: " + std::string(mdb_strerror(rc)));
-    
+
     return rc;
   }
 
@@ -263,7 +264,7 @@ public:
     return rc;
   }
 
-  
+
   // this is something you can do, readonly
   MDBDbi openDB(string_view dbname, int flags)
   {
@@ -272,7 +273,7 @@ public:
 
   MDBROCursor getCursor(const MDBDbi&);
   MDBROCursor getROCursor(const MDBDbi&);
-    
+
   operator MDB_txn*()
   {
     return d_txn;
@@ -288,8 +289,8 @@ public:
   }
 };
 
-/* 
-   A cursor in a read-only transaction must be closed explicitly, before or after its transaction ends. It can be reused with mdb_cursor_renew() before finally closing it. 
+/*
+   A cursor in a read-only transaction must be closed explicitly, before or after its transaction ends. It can be reused with mdb_cursor_renew() before finally closing it.
 
    "If the parent transaction commits, the cursor must not be used again."
 */
@@ -379,7 +380,7 @@ public:
        throw std::runtime_error("Unable to find from cursor: " + std::string(mdb_strerror(rc)));
     return rc;
   }
-  
+
   int lower_bound(const MDBInVal& in, MDBOutVal& key, MDBOutVal& data)
   {
     key.d_mdbval = in.d_mdbval;
@@ -390,7 +391,7 @@ public:
     return rc;
   }
 
-  
+
   int nextprev(MDBOutVal& key, MDBOutVal& data, MDB_cursor_op op)
   {
     int rc = mdb_cursor_get(d_cursor, const_cast<MDB_val*>(&key.d_mdbval), &data.d_mdbval, op);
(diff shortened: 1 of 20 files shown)


pdns/reczones.cc
```diff
@@ -197,7 +197,7 @@ void primeRootNSZones(DNSSECMode mode, unsigned int depth)
   }
 }
 
-static void makeNameToIPZone(std::shared_ptr<SyncRes::domainmap_t> newMap, const DNSName& hostname, const string& ip, Logr::log_t log)
+static void makeNameToIPZone(const std::shared_ptr<SyncRes::domainmap_t>& newMap, const DNSName& hostname, const string& ip, Logr::log_t log)
```
cmouse: is this some lint that is in CI pipeline too?

pdns/reczones.cc
```diff
@@ -237,15 +237,15 @@ static void makeNameToIPZone(const std::shared_ptr<SyncRes::domainmap_t>& newMap
...
+  vector<string> ipParts;
+  stringtok(ipParts, address, ".");
 
   SyncRes::AuthDomain ad;
   ad.d_rdForward = false;
 
   DNSRecord dr;
-  for (int n = ipparts.size() - 1; n >= 0; --n) {
-    dr.d_name.appendRawLabel(ipparts[n]);
+  for (auto part = ipParts.rbegin(); part != ipParts.rend(); ++part) {
```
cmouse: `const auto& part`

ext/lmdb-safe/lmdb-safe.hh
```diff
@@ -1,4 +1,5 @@
 #pragma once
+#include <string_view>
```
cmouse, line 2: this commit seems totally unrelated here.

pdns/dnsparser.hh
```diff
@@ -304,6 +305,18 @@ struct DNSRecord
   uint16_t d_clen{};
   DNSResourceRecord::Place d_place;
 
+  [[nodiscard]] std::string toString(const std::string& indent = "") const
```
cmouse, line 308: this does not look like toString for a record, more like some special represetantion.

pdns/reczones-helpers.cc
```diff
@@ -27,115 +27,129 @@
...
                                        Logr::log_t log)
 {
-  for (unsigned int n = 1; n < parts.size(); ++n) {
-    if (searchSuffix.empty() || parts[n].find('.') != string::npos) {
-      makeNameToIPZone(newMap, DNSName(parts[n]), parts[0], log);
+  const ComboAddress address{parts[0]};
+
+  // Go over the hostname and aliases (parts[1], parts[2], etc...) and add entries
+  // for forward lookups.
+  for (auto name = parts.cbegin() + 1; name != parts.cend(); ++name) {
```
cmouse, line 136: `const auto& name`

cmouse (changes requested):

pdns/reczones.cc
fredmorcos: Not yet, but I want to get it integrated.

ext/lmdb-safe/lmdb-safe.hh
fredmorcos: It was to fix an error message I was getting in my IDE, so it's part of "cleanup". Is it OK if I keep it?

pdns/iputils.hh
```diff
@@ -283,6 +283,32 @@ union ComboAddress {
...
+      auto c = (ip >> 16) & 0xFF;
+      auto d = (ip >> 24) & 0xFF;
+      return std::to_string(a) + "." + std::to_string(b) + "." + std::to_string(c) + "." + std::to_string(d);
+    }
+    else {
+      const auto* addr = &sin6.sin6_addr;
+      std::stringstream res{};
+      res << std::hex;
+      for (int i = 15; i >= 0; i--) {
+        auto byte = (int)addr->s6_addr[i];
```
cmouse, line 301: why this (int) cast here?

pdns/reczones-helpers.cc
fredmorcos: I must be missing something here. Doesn't the `const` in your suggestion apply to the iterator object here? How would the iterator advance if it cannot be mutated?

pdns/reczones.cc
fredmorcos: This was removed later in favor of `ComboAddress`.

pdns/reczones-helpers.cc
cmouse: oh right it's iterator... my bad.

ext/lmdb-safe/lmdb-safe.hh
Habbie: this can stay

pdns/iputils.hh
fredmorcos: Probably a left-over.

pdns/dnsparser.hh
fredmorcos: You mean it should be renamed to something else?

cmouse: yes.

omoerbeek:
I did not look at the code yet, but did a few tests.

Looking at at an example `/etc/hosts`  containing
```
194.109.6.2     p
194.109.6.3     p
2001:888:0:7::2 p
2001:888:0:7::2 q
```
I see that `dig -t A p` succeeds but has one record. `dig -t AAAA p` gives me an empty answer section. `dig -t AAAA q` works and gives me a single record.

In an ideal world the forward mapping would work for all cases: two answer records for `dig -t A p`, one record for `dig -t AAAA p` and one record for `dig -t AAAA q`.

Having both an IPv4 and an IPv6 address for the same name *is* very common. So it would be nice if we would handle at least that case.

I don't know if this is feasible with the current structure, I guess it would mean merging domains in some cases. 

The reverse (PTR) mapping should give only the first name encountered for each address. This seems to work as expected.

omoerbeek:
To  be clear, the behavior above is already present in master, so it is not a regression.

pdns/dnsparser.hh
```diff
@@ -289,17 +290,33 @@ protected:
...
+  DNSRecord(const std::string& name, std::shared_ptr<DNSRecordContent> content, const uint16_t type, const uint16_t qclass = QClass::IN, const uint32_t ttl = 86400, const uint16_t clen = 0, const DNSResourceRecord::Place place = DNSResourceRecord::ANSWER) :
+    d_name(DNSName(name)), d_content(std::move(content)), d_type(type), d_class(qclass), d_ttl(ttl), d_clen(clen), d_place(place) {}
+
   DNSName d_name;
   std::shared_ptr<DNSRecordContent> d_content;
   uint16_t d_type;
-  uint16_t d_class;
-  uint32_t d_ttl;
-  uint16_t d_clen;
+  uint16_t d_class{};
```
omoerbeek, line 303: Why this empty initialization? Isn't it the default? Or are you trying to make that explicit? In that case it make sense.

pdns/syncres.hh
```diff
@@ -114,6 +114,38 @@ public:
...
+    bool operator==(const AuthDomain& rhs) const
+    {
+      return
+        d_records == rhs.d_records &&
+        d_servers == rhs.d_servers &&
+        d_name == rhs.d_name &&
+        d_rdForward == rhs.d_rdForward;
+    }
+
+    [[nodiscard]] std::string toString(const std::string& indent = "", const std::string& indentLevel = "  ") const
```
omoerbeek, line 126: I think this is also not a generic toString(), more of a debug aid, so a better fitting name is needed.

pdns/reczones.cc
```diff
@@ -197,100 +200,14 @@ void primeRootNSZones(DNSSECMode mode, unsigned int depth)
...
-
 static void convertServersForAD(const std::string& zone, const std::string& input, SyncRes::AuthDomain& ad, const char* sepa, Logr::log_t log, bool verbose = true)
 {
   vector<string> servers;
   stringtok(servers, input, sepa);
   ad.d_servers.clear();
 
   vector<string> addresses;
-  for (vector<string>::const_iterator iter = servers.begin(); iter != servers.end(); ++iter) {
+  for (auto iter = servers.begin(); iter != servers.end(); ++iter) {
```
omoerbeek, line 210:
```suggestion
  for (const auto& server : servers) {
```

pdns/reczones.cc
```diff
@@ -197,100 +200,14 @@ void primeRootNSZones(DNSSECMode mode, unsigned int depth)
...
     ComboAddress addr = parseIPAndPort(*iter, 53);
```
omoerbeek, line 211:
```suggestion
    ComboAddress addr = parseIPAndPort(server, 53);
```

pdns/dnsparser.hh
fredmorcos: It is to be explicit, and then there's no need to initialize them in the constructors. I believe without assignment, members don't get "zero values" and instead just get pre-existing memory garbage.

omoerbeek:
> @omoerbeek I've opened an issue regarding the problem you mention: #11690 I think we could get this PR merged since it doesn't introduce a regression, and I could work on the issue as a separate PR. I hope that is alright with you.

Great! I intend to finish the code review soon.

pdns/dnsparser.hh
```diff
@@ -291,19 +291,19 @@ protected:
 struct DNSRecord
 {
   DNSRecord() :
-    d_type(0), d_class(QClass::IN), d_place(DNSResourceRecord::ANSWER)
```
omoerbeek: In the original, type was always initialized (either to zero or to the argument to the ct). The default constructor now only inits `d_class` (explicitly) and `d_place` (implicitly), leaving `d_type` uninitialized. I'm worried there is code that depends on the zero value.
The original and the PR leave `d_ttl` and `d_clen` uninitialized in the default ct, but we can be almost be sure nothing depends on that. For `d_type` that is not clear.

pdns/syncres.hh
```diff
@@ -114,6 +114,38 @@ public:
     DNSName d_name;
     bool d_rdForward{false};
 
+    bool operator==(const AuthDomain& rhs) const
```
omoerbeek, line 117: IMO both these functions do not need to be inline. The code is only used for debugging and this header file is already too big...

pdns/dnsparser.hh
```diff
@@ -289,16 +290,44 @@ protected:
...
-  uint32_t d_ttl;
-  uint16_t d_clen;
-  DNSResourceRecord::Place d_place;
+  uint16_t d_type{};
+  uint16_t d_class{};
+  uint32_t d_ttl{};
+  uint16_t d_clen{};
+  DNSResourceRecord::Place d_place = DNSResourceRecord::ANSWER;
+
+  [[nodiscard]] std::string print(const std::string& indent = "") const
```
omoerbeek, line 320: No need for inlining here

omoerbeek (changes requested):

pdns/dnsparser.hh
fredmorcos: `d_type` is being initialized to 0 because of `uint16_t d_type{};` and the PR does not leave `d_ttl` and `d_clen` uninitialized because of:
```
uint32_t d_ttl{};
uint16_t d_clen{};
```

Am I missing something?

omoerbeek: Nope, I was missing something, I read the `{}` as default init, while it is actually zero-init.

fredmorcos: Umm.. what is the difference?

omoerbeek: For integral fields the default (absent) initializer means not initialized at all, i.e. garbage. For some reason in equated `{}` as the default initialized in my initial comment, while it actually does init to zero.

fredmorcos: Right. Thanks for the clarification.

omoerbeek (approved):
Should be good to merge after conflict is resolved.

PR #11684, opened 2022-06-07
dnsdist: Fix 'dumpStats()' with custom metrics

rgacogne:
### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


rgacogne:
CI failure is unrelated.

chbruyand (approved):
Sorry I missed that!

rgacogne:
> Sorry I missed that!

Don't worry about it, I also missed it during my review!
I did not want to introduce too many changes in that PR but I believe the existing behaviour of a condition-less, final `else` is error-prone, it would be better to test that the value is actually what we expect and do nothing otherwise. I'll check later today if it can easily be done.

PR #11689, opened 2022-06-10
dnsdist: Avoid allocating memory in LB policies for small number of servers

rgacogne:
### Short description

The gist of this commit is that we do not need to dynamically allocate memory in our load-balancing policies when we have a few number of servers to consider: we can very well use a stack-based array instead.
Could use a good review :)

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdistdist/dnsdist-lbpolicies.cc
+++ b/pdns/dnsdistdist/dnsdist-lbpolicies.cc
@@ -29,6 +29,26 @@
 GlobalStateHolder<ServerPolicy> g_policy;
 bool g_roundrobinFailOnNoServer{false};
 
+template <class T> static std::shared_ptr<DownstreamState> getLeastOutstanding(const ServerPolicy::NumberedServerVector& servers, T& poss)
+{
+  /* so you might wonder, why do we go through this trouble? The data on which we sort could change during the sort,
+     which would suck royally and could even lead to crashes. So first we snapshot on what we sort, and then we sort */
+  size_t position = 0;
+  for (const auto& d : servers) {
+    if (d.second->isUp()) {
+      poss[position] = std::make_pair(std::make_tuple(d.second->outstanding.load(), d.second->d_config.order, d.second->latencyUsec), position);
+    }
+    ++position;
+  }
+
+  if (position == 0) {
+    return shared_ptr<DownstreamState>();
+  }
+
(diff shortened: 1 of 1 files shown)


omoerbeek (commented):
A few comments, will finish review tomorrow.

pdns/dnsdistdist/dnsdist-lbpolicies.cc
```diff
@@ -101,24 +94,56 @@ static shared_ptr<DownstreamState> valrandom(unsigned int val, const ServerPolic
         sum += d.second->d_config.d_weight;
       }
 
-      poss.emplace_back(sum, d.first);
+      poss[usableServers]  = std::make_pair(sum, d.first);
+      usableServers++;
     }
   }
 
   // Catch poss & sum are empty to avoid SIGFPE
```
omoerbeek, line 102:
```suggestion
  // Catch usableServers & sum are empty to avoid SIGFPE
```

pdns/dnsdistdist/dnsdist-lbpolicies.cc
```diff
@@ -101,24 +94,56 @@ static shared_ptr<DownstreamState> valrandom(unsigned int val, const ServerPolic
...
-  if (poss.empty() || sum == 0) {
+  if (usableServers == 0 || sum == 0) {
     return shared_ptr<DownstreamState>();
   }
 
   int r = val % sum;
-  auto p = upper_bound(poss.begin(), poss.end(),r, [](int r_, const decltype(poss)::value_type& a) { return  r_ < a.first;});
+  auto p = upper_bound(poss.begin(), poss.begin() + usableServers, r, [](int r_, const typename T::value_type& a) { return  r_ < a.first;});
   if (p == poss.end()) {
```
omoerbeek, line 109: Hmm, this does not feel right. I believe `std::upper_bound` will return `last` of nothing was found. And `last` does not necessarily equals `poss.end()`

omoerbeek (changes requested):
A few nits and one potential real issue (wrt `std::upper_bound`) found.

pdns/dnsdistdist/dnsdist-lbpolicies.cc
rgacogne: Ouch, that test is indeed no longer correct, good catch!

omoerbeek (approved):

PR #11692, opened 2022-06-10
rec: Run tasks from houskeeping thread in a proper way.

omoerbeek:
Previously, this was only done if log-common-errors was true, due
to argument reversal.  In general task *would* be executed, as they
are also run after each query processed by SyncRes (so not after
packet cache hits).

Thanks to @jelu!

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code


pdns/recursordist/rec-main.cc
```diff
@@ -2084,8 +2084,9 @@ static void houseKeeping(void*)
     // Below are the thread specific tasks for the handler and the taskThread
     // Likley a few handler tasks could be moved to the taskThread
     if (info.isTaskThread()) {
+      cerr << time(nullptr) << " T" << endl;
```
rgacogne, line 2087: I'm pretty sure you did not mean to commit this :)

omoerbeek: Oops, it really is Friday afternoon

rgacogne (approved):

PR #11696, opened 2022-06-14
dnsdist: Bind to the requested src interface without a src address

rgacogne:
### Short description

In some cases we want to bind to a specific interface when contacting a backend without specifying an exact source address, leaving the kernel select one for us.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdistdist/dnsdist-backend.cc
+++ b/pdns/dnsdistdist/dnsdist-backend.cc
@@ -58,17 +58,18 @@ bool DownstreamState::reconnect()
     }
     if (!IsAnyAddress(d_config.remote)) {
       fd = SSocket(d_config.remote.sin4.sin_family, SOCK_DGRAM, 0);
+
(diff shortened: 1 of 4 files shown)


rgacogne:
CI failure is unrelated (DoT to powerdns.com failing?)

omoerbeek (commented):
I'm wondering why the `IP_BIND_ADDRESS_NO_PORT` call is missing for `dnsdist-backend.cc:DownstreamState::reconnect()` (this is already the case in master). If it is intended, a comment is in place.

Rest looks good.

rgacogne:
> I'm wondering why the `IP_BIND_ADDRESS_NO_PORT` call is missing for `dnsdist-backend.cc:DownstreamState::reconnect()` (this is already the case in master). If it is intended, a comment is in place.

I don't expect a big impact given that we only bind these outgoing UDP sockets once, but it's a lot better to be consistent. Pushed!

Issue #3841, opened 2016-05-10
recursor: validation fails when query for validation fails


pieterlexis:
This does not need to be the case: e.g. when a NSEC proof of DS-non-existence in a parent zone exists, it does not matter that the childs-zone nameserver returns a non-NOERROR.

this is e.g. the case with `swivelpole-com.mail.protection.outlook.com.`, see log:
```
May 10 14:16:48 1 [1/1] question for 'swivelpole-com.mail.protection.outlook.com.|A' from 127.0.0.1
May 10 14:16:48 [1] swivelpole-com.mail.protection.outlook.com.: Wants DNSSEC processing in query for A
May 10 14:16:48 [1] swivelpole-com.mail.protection.outlook.com.: Looking for CNAME cache hit of 'swivelpole-com.mail.protection.outlook.com.|CNAME'
May 10 14:16:48 [1] swivelpole-com.mail.protection.outlook.com.: No CNAME cache hit of 'swivelpole-com.mail.protection.outlook.com.|CNAME' found
May 10 14:16:48 [1] swivelpole-com.mail.protection.outlook.com.: No cache hit for 'swivelpole-com.mail.protection.outlook.com.|A', trying to find an appropriate NS record
May 10 14:16:48 [1] swivelpole-com.mail.protection.outlook.com.: Checking if we have NS in cache for 'swivelpole-com.mail.protection.outlook.com.'
May 10 14:16:48 [1] swivelpole-com.mail.protection.outlook.com.: no valid/useful NS in cache for 'swivelpole-com.mail.protection.outlook.com.'
May 10 14:16:48 [1] swivelpole-com.mail.protection.outlook.com.: Checking if we have NS in cache for 'mail.protection.outlook.com.'
May 10 14:16:48 [1] swivelpole-com.mail.protection.outlook.com.: no valid/useful NS in cache for 'mail.protection.outlook.com.'
May 10 14:16:48 [1] swivelpole-com.mail.protection.outlook.com.: Checking if we have NS in cache for 'protection.outlook.com.'
May 10 14:16:48 [1] swivelpole-com.mail.protection.outlook.com.: no valid/useful NS in cache for 'protection.outlook.com.'
May 10 14:16:48 [1] swivelpole-com.mail.protection.outlook.com.: Checking if we have NS in cache for 'outlook.com.'
May 10 14:16:48 [1] swivelpole-com.mail.protection.outlook.com.: no valid/useful NS in cache for 'outlook.com.'
[...]

PR #11704, opened 2022-06-17
Rec: make edns table a sparse table

omoerbeek:
### Short description

Only store exceptional status: `Ignorant` or `No`. Including cleanup of timestamp handling.
An `OK` entry now does not expire, as it is not stored. An `Ignorant` or `Bad` status will expire after two hours. After that the IP will be probed again. To avoid unnessesary churn, timestamps will only be set on initial status update or change. After two hours the status will be reset to OK by removing the entry, and potentially a new state will be created after that.

If an `Ignorant` entry turns out to be `OK` it will be erased.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/recursordist/rec-main.cc
+++ b/pdns/recursordist/rec-main.cc
@@ -2117,7 +2117,7 @@ static void houseKeeping(void*)
 
       static PeriodicTask pruneEDNSTask{"pruneEDNSTask", 60};
       pruneEDNSTask.runIfDue(now, [now]() {
-        SyncRes::pruneEDNSStatuses(now.tv_sec - 2 * 3600);
+        SyncRes::pruneEDNSStatuses(now.tv_sec);
       });
 
       if (SyncRes::s_max_busy_dot_probes > 0) {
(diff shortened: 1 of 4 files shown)


omoerbeek:
Investigating regress failure

omoerbeek:
So mandatory does not really mean mandatory. This is a bit confusing, but luckily we have a test: `test_Interop.py:testInterop.testFORMERR`.
Which basically says: even if EDNS is mandatory, `NOEDNS` overrides. Fix upcoming.
`

omoerbeek:
Setting to draft, since the reasoning saying `ednsMANDATORY` has becomes an obsolete argument has to be double checked.

rgacogne:
> Setting to draft, since the reasoning saying `ednsMANDATORY` has becomes an obsolete argument has to be double checked.

Hmm, it would mean that we accept to contact a nameserver without sending an OPT record even though we know that we need DNSSEC records? That does not sound good..

omoerbeek:
> > Setting to draft, since the reasoning saying `ednsMANDATORY` has becomes an obsolete argument has to be double checked.
> 
> Hmm, it would mean that we accept to contact a nameserver without sending an OPT record even though we know that we need DNSSEC records? That does not sound good..

I found that also surprising, but master does no send `OPT` if `NOEDNS`, even when mandatory is set. See also the mentioned test and #3841.

rgacogne:
I guess it depends on the goal: if we want to make sure we get DNSSEC records and hope that a different nameserver will support EDNS, it might make sense to not send a query without EDNS, but I guess it's too late in the process to know if we still have other nameservers to try, and if they do support EDNS.
On the other hand sending the query without an OPT record means we will get an answer if this is the best server we will get, and we can serve it if CD=1.. So, yeah, not an easy call to make :)

rgacogne (approved):

omoerbeek:
It's likely best to keep the existing logic wrt mandatory, even if is a bit twisted. Later we can revisit.

PR #11712, opened 2022-06-21
dnsdist: fill ringbuffers with responses served from the cache

### LINKED ISSUE #11585: dnsdist: Add queries/responses handled by packetcache into ringbuffers?
- Program: dnsdist
 - Issue type: Bug report / Feature Request

### Short description
Queries handled by the packet cache don't seem to end up in the ringbuffers. This makes dynamic query counting, e.g. via exceedServFails(), useless for the usual repeating packet floods by misbehaving clients on similar FQDNs.
While it certainly has some performance impact I wonder if it would make sense to adjust the ring buffers so that they can also be used for proper counting of queries hitting the caches, potentially in a configurable way. Or look into a different solution for being able to handle excessive clients/queries without causing excessive performance impact.
If kept as is I think the current behavior might be missing some documentation at ringbuffers / dynamic blocks.
(As discussed)

chbruyand:
### Short description
This tries to make sure we call `handleResponseSent()` when returning directly a self generated answer (like when serving from the packet cache).

`grepq()` display is now:

```
-2.3    127.0.0.1:60976                                 DoUDP                     11991 www.example.com.          A               RD    Question
-2.2    127.0.0.1:60976                                 DoUDP        9.9.9.9:53   11991 www.example.com.          A     102.7     RD    No Error. 1 answers
-1.7    127.0.0.1:52311                                 DoUDP                     25359 www.example.com.          A               RD    Question
-1.7    127.0.0.1:52311                                 -            Cache        25359 www.example.com.          A     0.0       RD    No Error. 1 answers
```

I was a bit unsure about what to do with latencies as it seems  to be initialized or not depending on the context (UDP/TCP/Do{TH}). Maybe zero it everywhere ? Also I may simply have missed something ?

Fix #11585 

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdist-tcp.cc
+++ b/pdns/dnsdist-tcp.cc
@@ -257,6 +257,10 @@ static void handleResponseSent(std::shared_ptr<IncomingTCPConnectionState>& stat
       backendProtocol = dnsdist::Protocol::DoTCP;
     }
     ::handleResponseSent(ids, udiff, state->d_ci.remote, ds->d_config.remote, static_cast<unsigned int>(currentResponse.d_buffer.size()), currentResponse.d_cleartextDH, backendProtocol);
+  } else {
+    const auto& ids = currentResponse.d_idstate;
+    double udiff = ids.sentTime.udiff();
+    ::handleResponseSent(ids, udiff, state->d_ci.remote, ComboAddress(), static_cast<unsigned int>(currentResponse.d_buffer.size()), currentResponse.d_cleartextDH, ids.protocol);
   }
 }
 
@@ -743,6 +747,13 @@ static void handleQuery(std::shared_ptr<IncomingTCPConnectionState>& state, cons
     TCPResponse response;
     response.d_selfGenerated = true;
     response.d_buffer = std::move(state->d_buffer);
+    setIDStateFromDNSQuestion(response.d_idstate, dq, std::move(qname));
+    response.d_idstate.origID = dh->id;
+    response.d_idstate.cs = state->d_ci.cs;
+
+    DNSResponse dr = makeDNSResponseFromIDState(response.d_idstate, response.d_buffer);
+    memcpy(&response.d_cleartextDH, dr.getHeader(), sizeof(response.d_cleartextDH));
+
     state->d_state = IncomingTCPConnectionState::State::idle;
     ++state->d_currentQueriesCount;
     state->queueResponse(state, now, std::move(response));
--- a/pdns/dnsdist-tcp.cc
+++ b/pdns/dnsdist-tcp.cc
@@ -259,8 +259,7 @@ static void handleResponseSent(std::shared_ptr<IncomingTCPConnectionState>& stat
     ::handleResponseSent(ids, udiff, state->d_ci.remote, ds->d_config.remote, static_cast<unsigned int>(currentResponse.d_buffer.size()), currentResponse.d_cleartextDH, backendProtocol);
   } else {
     const auto& ids = currentResponse.d_idstate;
-    double udiff = ids.sentTime.udiff();
-    ::handleResponseSent(ids, udiff, state->d_ci.remote, ComboAddress(), static_cast<unsigned int>(currentResponse.d_buffer.size()), currentResponse.d_cleartextDH, ids.protocol);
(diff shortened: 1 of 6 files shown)


pdns/dnsdist-tcp.cc
```diff
@@ -750,6 +754,13 @@ static void handleQuery(std::shared_ptr<IncomingTCPConnectionState>& state, cons
     TCPResponse response;
     response.d_selfGenerated = true;
     response.d_buffer = std::move(state->d_buffer);
+    setIDStateFromDNSQuestion(response.d_idstate, dq, std::move(qname));
+    response.d_idstate.origID = dh->id;
```
rgacogne, line 758: The pointer might have been invalidated by the move above, so we should get the ID before the moves.

pdns/dnsdist-tcp.cc
```diff
@@ -750,6 +754,13 @@ static void handleQuery(std::shared_ptr<IncomingTCPConnectionState>& state, cons
...
+    response.d_idstate.cs = state->d_ci.cs;
+
+    DNSResponse dr = makeDNSResponseFromIDState(response.d_idstate, response.d_buffer);
```
rgacogne, line 761: That sound a bit wasteful, don't we already have a pointer to the header above (before the moves)?

pdns/dnsdistdist/doh.cc
```diff
@@ -620,6 +620,9 @@ static void processDOHQuery(DOHUnitUniquePtr&& du)
       if (du->response.empty()) {
         du->response = std::move(du->query);
       }
+      auto dh = const_cast<struct dnsheader*>(reinterpret_cast<const struct dnsheader*>(du->response.data()));
+
+      handleResponseSent(qname, QType(qtype), 0., du->downstream->d_config.remote, ComboAddress(), du->response.size(), *dh, du->downstream->getProtocol());
```
rgacogne, line 625:
```suggestion
      handleResponseSent(qname, QType(qtype), 0., du->ids.origDest, ComboAddress(), du->response.size(), *dh, dnsdist::Protocol::DoUDP);
```

rgacogne (changes requested):
> I was a bit unsure about what to do with latencies as it seems to be initialized or not depending on the context (UDP/TCP/Do{TH}). Maybe zero it everywhere ?

Right, it's indeed not yet initialized at that point in some cases. I suggest we zero it everywhere for now, as I don't think we really care about the latency for self-generated answers anyway.
We also have a problem with the protocol which is usually the backend protocol, but we have not yet selected a backend at this point. How about always setting it to the incoming protocol for self-generated answers?

pdns/dnsdistdist/doh.cc
rgacogne: Right, I wrote that comment before settling on the incoming protocol, sorry!

chbruyand:
Thanks for the review.

pdns/dnsdist-tcp.cc
rgacogne: I'm afraid this is still a bit too late if `dh` is pointing to the data that was in `state->d_buffer`?

rgacogne (changes requested):

pdns/dnsdistdist/doh.cc
```diff
@@ -620,6 +620,9 @@ static void processDOHQuery(DOHUnitUniquePtr&& du)
       if (du->response.empty()) {
         du->response = std::move(du->query);
       }
+      auto dh = const_cast<struct dnsheader*>(reinterpret_cast<const struct dnsheader*>(du->response.data()));
+
+      handleResponseSent(qname, QType(qtype), 0., du->ids.origDest, ComboAddress(), du->response.size(), *dh, dnsdist::Protocol::DoH);
```
rgacogne, line 625: The dnsdist CI failure is a real issue. In doh.cc, we can get `ProcessQueryResult::SendAnswer` with an answer that has been generated from Lua, for example, and such an answer does not have to be a valid DNS packet (it could be any kind of content that a HTTP client can handle, like HTML). This should be enough to handle almost all cases, and make sure that we do not read past our buffer in any case:
```suggestion
      if (du->response.size() > sizeof(dnsheader) && du->contentType.empty()) {
        auto dh = reinterpret_cast<const struct dnsheader*>(du->response.data());

        handleResponseSent(qname, QType(qtype), 0., du->ids.origDest, ComboAddress(), du->response.size(), *dh, dnsdist::Protocol::DoH);
      }
```

rgacogne (approved):
:+1:

PR #11713, opened 2022-06-21
dnsdist: Implement a limit of concurrent connections to a backend

rgacogne:
### Short description

This PR adds a new option to limit the number of TCP connections to a backend. When that limit is reached, no new connection to that backend will be opened, and queries routed to that backend will fail, until the number of connections goes back below that limit.
It also introduces a new metric for that value, and a counter indicating how often it was reached.
Finally it slightly improves the way we are reusing TCP connections by detaching them from the incoming connection a bit earlier in the process, and immediately releasing idle connections when no longer need them.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code


rgacogne:
Rebased to fix a conflict.

rgacogne:
Rebased to fix a conflict.

Habbie (approved):

PR #11719, opened 2022-06-23
Fix warnings found by compiler on MacOS, where char is unsigned

omoerbeek:
### Short description

The test in `dnslabeltext.rl` should be exactly equivalent, but not cause a compiler warning on platforms where `char` is unsigned.

The test in `dnsname.cc` gets more strict, it will now detect illegal label lengths on all platforms, not just > 127 only on platforms where `char` is signed, as it does before this PR.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnslabeltext.rl
+++ b/pdns/dnslabeltext.rl
@@ -108,11 +108,11 @@ DNSName::string_t segmentDNSNameRaw(const char* realinput, size_t inputlen)
         const char* eof = pe;
         int cs;
         char val = 0;
-        char labellen=0;
+        unsigned char labellen=0;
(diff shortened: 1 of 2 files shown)


rgacogne (approved):

rgacogne:
> The test in `dnsname.cc` gets more strict, it will now detect illegal label lengths on all platforms, not just > 127 only on platforms where `char` is signed, as it does before this PR.

That test should in theory not be needed because we are dealing with already validated data from `DNSName` objects, if I'm not mistaken, so the existing behaviour should not be an issue?

PR #11723, opened 2022-06-24
dnsdist: Use getrandom() if available

rgacogne:
### Short description

It turns out to use much less CPU than RANDOM_bytes() from libcrypto.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


rgacogne:
> This looks good though I'd love to see a single implementation shared by all products one day.

Fully agreed, but then `dns_random` needs to stop pulling in the arguments parser and logger :-)

PR #11738, opened 2022-07-01
auth 2136: match autosplit TXT more usefully

Habbie:
### Short description
Overly long TXT records, autosplit by PDNS during packet serialization, were impossible to delete via Updates except by deleting the whole RRset. This PR fixes that.

I also added a few tests for things that already worked.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


Habbie:
I think this, or something like this, should in fact be done for all types. I'll PR that to master after 4.7 is branched so that bigger change can soak for a while.

rgacogne:
Out of curiosity, do you have an example of what the representation looks like before and after your fix?

pdns/rfc2136handler.cc
```diff
@@ -398,7 +399,17 @@ uint PacketHandler::performUpdate(const string &msgPrefix, const DNSRecord *rr,
           recordsToDelete.push_back(rec);
       }
       if (rr->d_class == QClass::NONE) { // 3.4.2.4
-        if (rrType == rec.qtype && rec.getZoneRepresentation() == rr->d_content->getZoneRepresentation())
+        auto repr = rec.getZoneRepresentation();
+        if (rec.qtype == QType::TXT) {
```
rgacogne, line 403: Just a performance nit, but would it make sense to skip that dance when `rrType != rec.qtype` (although I'm not sure that actually happens)?

rgacogne (approved):

pdns/rfc2136handler.cc
Habbie: can't happen on this path, indeed; however, it's a very good idea for the bigger followup PR I want to land in 4.8 because it'll touch more code paths

PR #11761, opened 2022-07-05
dnsdist: Keep retained capabilities even when switching user/group

rgacogne:
### Short description

On Linux, we support retaining some capabilities if we are running as root (eeew) or as an unprivileged user with ambient capabilities, but we did not yet support keeping these if we were started as root but then switched to a different user ID and/or group ID.
This commit uses `PR_SET_KEEPCAPS`, when available, to do just that, to be able to retain the capabilities we need without running as a fully privileged users even when we cannot easily use ambient capabilities.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/pdns/capabilities.cc
+++ b/pdns/capabilities.cc
@@ -27,12 +27,13 @@
 
 #ifdef HAVE_LIBCAP
 #include <sys/capability.h>
+#include <sys/prctl.h>
 #endif
 
 #include "capabilities.hh"
 #include "misc.hh"
 
-void dropCapabilities(std::set<std::string> capabilitiesToKeep)
+bool dropCapabilities(std::set<std::string> capabilitiesToKeep)
 {
 #ifdef HAVE_LIBCAP
    cap_t caps = cap_get_proc();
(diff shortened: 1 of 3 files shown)


pdns/capabilities.cc
```diff
@@ -27,12 +27,13 @@
...
 #ifdef HAVE_LIBCAP
 #include <sys/capability.h>
+#include <sys/prctl.h>
 #endif
 
 #include "capabilities.hh"
 #include "misc.hh"
 
-void dropCapabilities(std::set<std::string> capabilitiesToKeep)
+bool dropCapabilities(std::set<std::string> capabilitiesToKeep)
```
Habbie, line 36: we don't consume this bool anywhere, unless I'm missing something

pdns/dnsdist.cc
```diff
@@ -2639,6 +2647,10 @@ int main(int argc, char** argv)
       dropUserPrivs(newuid);
     }
 
+    if (retainedCapabilities) {
+      dropCapabilitiesAfterSwitchingIDs();
```
Habbie, line 2651: what would be the harm in calling this in the `!retainedCapabilities` case?

pdns/capabilities.cc
rgacogne: True, I wanted to stop dnsdist from warning about not being able to drop our capabilities when we don't have any.
But I still wanted to provide the caller with the information of whether something was actually done, even though I don't really care in dnsdist.

Habbie: Oh, right, the `throw` is masked by `return false` for the EPERM case. Ok!

pdns/dnsdist.cc
rgacogne: I don't see any from the top of my mind, but I prefer not to do anything at all when possible :)

Habbie: :+1:

Habbie (approved):

PR #11763, opened 2022-07-06
Add option to NegativeAndSOAAction to place SOA in the authority section

hlindqvist:
### Short description

This allow generating a full NODATA/NXDOMAIN cachable negative response, rather than an RPZ-style reponse with only an informational SOA record in the additional section.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/pdns/dnsdistdist/docs/rules-actions.rst
+++ b/pdns/dnsdistdist/docs/rules-actions.rst
@@ -1169,6 +1169,9 @@ The following actions exist.
 
   .. versionadded:: 1.6.0
 
+  .. versionchanged:: 1.8.0
+    Added the ``soaInAuthoritySection`` option.
+
   Turn a question into a response, either a NXDOMAIN or a NODATA one based on ''nxd'', setting the QR bit to 1 and adding a SOA record in the additional section.
   Note that this function was called :func:`SetNegativeAndSOAAction` before 1.6.0.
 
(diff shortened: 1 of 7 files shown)


pdns/dnsdistdist/docs/rules-actions.rst
```diff
@@ -1189,6 +1189,7 @@ The following actions exist.
   * ``aa``: bool - Set the AA bit to this value (true means the bit is set, false means it's cleared). Default is to clear it.
   * ``ad``: bool - Set the AD bit to this value (true means the bit is set, false means it's cleared). Default is to clear it.
   * ``ra``: bool - Set the RA bit to this value (true means the bit is set, false means it's cleared). Default is to copy the value of the RD bit from the incoming query.
+  * ``soaInAuthoritySection``: bool - Place the SOA record in the authority section for a complete NXDOMAIN/NODATA reponse that works as a cachable negative response, rather than the RPZ-style response with a purely informational SOA in the additional section.
```
rgacogne, line 1192: A note about when the change was introduced would be nice, something like:
```
.. versionchanged:: 1.8.0
  Added the ``soaInAuthoritySection`` option.
```
below the function definition.

rgacogne: It would also be nice to state that the default is `false`.

rgacogne (commented):
Great work, thank you! I would happily merge it as it is once the couple nits in the documentation have been fixed :-)

rgacogne (approved):
Great work, thanks a lot!

PR #11768, opened 2022-07-07
Enable include-dir by default in RPM builds, to be in line with DEB builds

### LINKED ISSUE #11766: Include support for "recursor.d/*conf" in RH based distro's
- Program: Recursor (but might want to verify Authoritative as well)
 - Issue type: Feature request

### Short description
Where the DEB packages include a statement to include configs from resursor.d/*, the RPM packages do not.

### Usecase
If one wants to include their own customised version of the recursor settings in a separate .rpm, there's currently no good way to do that. Including support for recursor.d/*.conf in the default recursor.conf, would allow that.

Would also increase consistency cross distributions, as the DEB package do ship with that in the default configfile.

### Description
Add a line to the default recursor.conf in the RPM package to support dropping config files in /etc/pdns-recursor/recursor.d/myconfig.conf

franklouwers:
Closes #11766

### DIFF
--- a/builder-support/specs/pdns-recursor.spec
+++ b/builder-support/specs/pdns-recursor.spec
@@ -75,6 +75,7 @@ make %{?_smp_mflags} check || (cat test-suite.log && false)
 make install DESTDIR=%{buildroot}
 
 %{__mv} %{buildroot}%{_sysconfdir}/%{name}/recursor.conf{-dist,}
+%{__mkdir} %{buildroot}%{_sysconfdir}/%{name}/recursor.d
 
 # change user and group to pdns-recursor and add default include-dir
 sed -i \
--- a/builder-support/specs/pdns-recursor.spec
+++ b/builder-support/specs/pdns-recursor.spec
@@ -44,7 +44,7 @@ package if you need a dns cache for your network.
 
 
 %prep
-%autosetup -p1 -n %{name}-%{getenv:BUILDER_VERSION} 
+%autosetup -p1 -n %{name}-%{getenv:BUILDER_VERSION}
 
 %build
 %if 0%{?rhel} < 8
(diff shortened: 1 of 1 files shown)


builder-support/specs/pdns-recursor.spec
```diff
@@ -75,11 +75,13 @@ make %{?_smp_mflags} check || (cat test-suite.log && false)
...
 
 %{__mv} %{buildroot}%{_sysconfdir}/%{name}/recursor.conf{-dist,}
+%{__mkdir} %{buildroot}%{_sysconfdir}/%{name}/recursor.d
 
-# change user and group to pdns-recursor
+# change user and group to pdns-recursor and add default include-dir
 sed -i \
     -e 's/# setuid=/setuid=pdns-recursor/' \
     -e 's/# setgid=/setgid=pdns-recursor/' \
+    -e 's!# include-dir=.*!&\ninclude-dir=%{_sysconfdir}/%{name}/recursor.d!' \
```
fredmorcos, line 84: I would prefer not using `&\n` here: `'s!# include-dir=.*!include-dir=%{_sysconfdir}/%{name}/recursor.d!'`

fredmorcos (changes requested):

builder-support/specs/pdns-recursor.spec
webidsecurityrepo: Everywhere i see i see Powrdns Auth and one of the db as backend, mysql crash, postgresql crash they all crash in repicas and stand alones in 1 day, 1 week or 1 month, why ? Ive tried mysql, postgres, sqlite which is the only one which fits all the owners criterias. But not Mariadb , redis , but most of all CockroachDB which is twice or 10 times faster, more stable and self healing than postgresql. And Mongodb is never mention.

But i was wondering since all the db we have tried as backends, and almost 80% of all guides say use db as backed, but why cant we use PowerDns as master and 2 or 3 slaves replicating ? That would be substainable or ? Why does it has to be auth with db as backend ? Has anyone ran a master + 2 or 3 powerdns recursors as slave ? And dnsdist is that then needed ? Has powerdns released any panel like knot and nsd?  A Noregian web browsing company has made a GUI for powerdns but it would be nice if they gave us what they use.

omoerbeek:
@franklouwers and @fredmorcos I'd like to merge this. Can you two decide what the appropriate `sed` command is? Thanks!

fredmorcos (approved):

omoerbeek:
@Habbie ie can you check if you want this in auth as well?

PR #11772, opened 2022-07-11
Implement catalog zones in the authoritative server

### LINKED ISSUE #9824: auth: implement catalog zones
https://zones.cat/

For PowerDNS, we have the experimental [PowerCATZ](https://github.com/PowerDNS/powercatz/) implementation that can consume catalog zones (without support for the `SERIAL` property in draft-ietf-dnsop-dns-catalog-zones-01). PowerCATZ uses the `account` field on a domain to keep track of what it owns.

This ticket serves as a place to collect ideas, thoughts and insights for our real implementation.

mind04:
### Short description
This pull implements catalog zones https://datatracker.ietf.org/doc/html/draft-ietf-dnsop-dns-catalog-zones-06 in the gsql and lmdb backends (producer and consumer).

Implementation status of optional properties:
- producer: 
  - [x] coo
  - [x] group
  - [ ] ext
- consumer: 
  - [x] coo
  - [x] group (partial, stored but no templates)
  - [ ] ext

TODO:
- [x] better tooling, 
- [x] docs
- [x] test at scale (750K zones)

Closes #9824 

### Checklist
I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


cmouse:
Why not use 'domainmetadata' table for options?

pdns/slavecommunicator.cc
```diff
@@ -86,6 +86,311 @@ struct ZoneStatus
...
+            }
+
+            vector<string> primaries;
+            for (const auto& primary : di.masters) {
+              primaries.push_back(primary.toStringWithPortExcept(53));
+            }
+            g_log << Logger::Warning << logPrefix << "update primaries for zone '" << ciXFR.d_zone << "' to '" << boost::join(primaries, ", ") << "'" << endl;
+            di.backend->setMasters(ciXFR.d_zone, di.masters);
+
+            retreive.emplace_back(ciXFR);
```
cmouse, line 160: `retreive` => `retrieve`

cmouse (changes requested):

mind04:
> Why not use 'domainmetadata' table for options?

An early version of this pull was using metadata but it did not scale well for a larger numbers of member zones. Composing a catalog from domains and metadata was a time consuming process. Storing the data in the domains table resulted in a massive performance boost.
Metadata is also parsed during normal dns lookups so you want to keep the number of items per zone as low as possible.
And since catalog zones is using a lot of them the alternate storage location in domains was also an improvement for normal lookups.

modules/godbcbackend/godbcbackend.cc
```diff
@@ -83,7 +83,7 @@ class gODBCFactory : public BackendFactory
 
     declare(suffix, "info-zone-query", "", "select id,name,master,last_check,notified_serial,type,options,catalog,account from domains where name=?");
 
-    declare(suffix, "info-all-slaves-query", "", "select id,name,master,last_check from domains where type='SLAVE'");
+    declare(suffix, "info-all-slaves-query", "", "select domains.id, domains.name, domains.type, domains.master, domains.last_check, records.content from domains LEFT JOIN records ON records.domain_id=domains.id AND records.type='SOA' AND records.name=domains.name where domains.type in ('SLAVE', 'CONSUMER')");
```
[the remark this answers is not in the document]
mind04: Yes, otherwise zones without a soa record will not show up in the result

modules/gmysqlbackend/schema.mysql.sql
```diff
@@ -3,13 +3,16 @@ CREATE TABLE domains (
   name                  VARCHAR(255) NOT NULL,
   master                VARCHAR(128) DEFAULT NULL,
   last_check            INT DEFAULT NULL,
-  type                  VARCHAR(6) NOT NULL,
+  type                  VARCHAR(8) NOT NULL,
   notified_serial       INT UNSIGNED DEFAULT NULL,
+  options               VARCHAR(64000) DEFAULT NULL,
```
RobinGeuze, line 8: Is there a specific reason you are using VARCHAR here rather than TEXT? This way rows can potentially get very big, which can hamper InnoDB performance.

pdns/tcpreceiver.cc
```diff
@@ -727,18 +734,46 @@ int TCPNameserver::doAXFR(const DNSName &target, std::unique_ptr<DNSPacket>& q,
...
+    // Ignore all records except NS at apex
+    sd.db->lookup(QType::NS, target, di.id);
+    while (sd.db->get(zrr)) {
+      zrrs.emplace_back(zrr);
+    }
+    if (zrrs.empty()) {
+      zrr.dr.d_name = target;
+      zrr.dr.d_ttl = 0;
+      zrr.dr.d_type = QType::NS;
+      zrr.dr.d_content = std::make_shared<TXTRecordContent>("invalid.");
```
Habbie, line 751: TXT? in an NS?

modules/lmdbbackend/lmdbbackend.cc
```diff
@@ -49,7 +49,7 @@
 
 #include "lmdbbackend.hh"
 
-#define SCHEMAVERSION 3
+#define SCHEMAVERSION 4
```
Habbie, line 52: do we need a migration? or are we good because you appended to the end of structs?

docs/catalog.rst
```diff
@@ -0,0 +1,125 @@
...
+
+Starting with the PowerDNS Authoritative Server 4.7.0, catalog zone support is available.
+
+Supported catalog versions
+--------------------------
+
++-----------------+----------+----------+
+| Catalog version | Producer | Consumer |
++=================+==========+==========+
+| 1 (ICS)         | No       | Yes      |
```
Habbie, line 12:
```suggestion
| 1 (ISC)         | No       | Yes      |
```

docs/catalog.rst
```diff
@@ -0,0 +1,125 @@
...
++-----------------+----------+----------+
+| 2 (RFC TBD)     | Yes      | Yes      |
++-----------------+----------+----------+
+
+All all the important features of catalog zone version "2" are supported. There are however a few properties where support is limited:
```
Habbie, line 17:
```suggestion
All the important features of catalog zones version "2" are supported. There are however a few properties where support is limited:
```

docs/catalog.rst
```diff
@@ -0,0 +1,125 @@
...
+
+Per zone settings
+-----------------
+
+It is highly recommended to protect catalog zones with :doc:`TSIG <../tsig>`
+
+CATALOG-HASH
+~~~~~~~~~~~~
+
+Producer zones are storing the member state as a hash in this metadata setting. This setting is managed by the authoritative server. Modify or deleting this value will result in a serial increase of the producer zone and the update or recreation of this value.
```
Habbie, line 47:
```suggestion
Producer zones store the member state as a hash in this metadata setting. This setting is managed by the authoritative server. Modifying or deleting this value will result in a serial increase of the producer zone and the update or recreation of this value.
```

docs/catalog.rst
```diff
@@ -0,0 +1,125 @@
...
+
+Setup Catalog Zones
+-------------------
```
Habbie, line 50:
```suggestion
Setting up catalog zones
------------------------
```

docs/catalog.rst
```diff
@@ -0,0 +1,125 @@
...
+
+.. note::
+  Catalog zone specification and operation is described in `DNS Catalog Zones <https://datatracker.ietf.org/doc/draft-ietf-dnsop-dns-catalog-zones)>`__.
+
+Setting up a producer zone
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Setting up a producer zone is not very different from a regular primary zone. A producer zone is a minimal zone of type CONSUMER with only SOA and NS records at apex. All the records in a producer zone are ignored while generating a catalog.
```
Habbie, line 58: please put a newline after ever sentence

docs/catalog.rst
```diff
@@ -0,0 +1,125 @@
...
+
+  pdnsutil load-zone catalog.invalid zones/catalog.invalid ZONEFILE
+  pdnsutil set-kind catalog.invalid producer
+
+Creating producer zones is supported in the :doc:`API <http-api/zone>`.
+
+Assigning members to a producer zone
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+After the producer zone is created it is necessary to assign member zone to it. In the example below example.com is the member and catalog.invalid is the catalog.
```
Habbie, line 89:
```suggestion
After the producer zone is created it is necessary to assign member zones to it. In the example below ``example.com`` is the member and ``catalog.invalid`` is the catalog.
```

docs/catalog.rst
```diff
@@ -0,0 +1,125 @@
...
+
+- coo - A single DNSName
+- group - Multiple string values for group are allowed
+
+.. code-block:: shell
+
+  pdnsutil set-option example.com producer coo other-catalog.invalid
+  pdnsutil set-option example.com producer group pdns-group-x pdns-group-y
+
+There is also an option to set a specific <unique-N> value for a zone this is done by setting a the `unique` value. This is used to signal a state reset to the consumer. The value u unique is a single dns label.
```
Habbie, line 107:
```suggestion
There is also an option to set a specific <unique-N> value for a zone. This is done by setting a the ``unique`` value. This is used to signal a state reset to the consumer. The value for ``unique`` is a single DNS label.
```

Habbie (changes requested):

docs/catalog.rst
klaus-nicat, line 58: Why is a "producer" zone of type "CONSUMER"? That sounds wrong.

modules/lmdbbackend/lmdbbackend.cc
mind04: The BOOST_CLASS_VERSION for DomainInfo is also increased. The migration is done on the fly based on the class version present in the database. https://github.com/PowerDNS/pdns/pull/11772/files#diff-c5a93f9a620612f89117271584c30d3b541e82f86cef733211d6f201c4d4a69cR209

modules/gpgsqlbackend/4.3.0_to_4.7.0_schema.pgsql.sql
```diff
@@ -0,0 +1,13 @@
+BEGIN;
+  ALTER TABLE domains ADD COLUMN options VARCHAR(65535) DEFAULT NULL;
+  ALTER TABLE domains ADD COLUMN catalog VARCHAR(255) DEFAULT NULL;
+
+  ALTER TABLE domains ADD COLUMN account_new VARCHAR(40) DEFAULT NULL;
```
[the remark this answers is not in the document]
mind04: Temporary column to move account back to the end of the row. This is done to keep an upgraded schema identical to the provided full schema.

Habbie (approved):

Habbie:
Klaus, thank you for your comments. #2 was documented. #1 and #3 seem like nice discussions for after we all gain some operational experience with this implementation. Perhaps you can post them as separate tickets at some point.

PR #11776, opened 2022-07-11
Rec: serve stale

omoerbeek:
### Short description

Showing this so others can take a look and play with it. General description in `recursor_cache.cc`.
Missing: docs, proper handling of caches records with ecs (they now expire too soon)I have to think how this can be tested in regression tests, basic unit test are there.

I have been testing this by blackholing outging dns requests using my firewall.
Enable by setting `serve-stale-extensions`, unit is 30s. So setting it to 100 will serve a stale record for 3000 seconds after it became stale.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [X] documented the code


pdns/recursor_cache.cc
```diff
@@ -187,6 +218,7 @@ MemRecursorCache::cache_t::const_iterator MemRecursorCache::getEntryUsingECSInde
       else {
         /* this netmask-specific entry has expired */
         moveCacheItemToFront<SequencedTag>(map.d_map, entry);
+        // XXX when serving stale, it should be kept, but we don't want a match wth lookupBestMatch()...
```
rgacogne, line 221: It might simply be OK to not serve stale ECS-specific entries.

pdns/recursor_cache.cc
```diff
@@ -200,6 +232,12 @@ MemRecursorCache::cache_t::const_iterator MemRecursorCache::getEntryUsingECSInde
   auto key = std::make_tuple(qname, qtype, boost::none, Netmask());
   auto entry = map.d_map.find(key);
   if (entry != map.d_map.end()) {
+    // If we are serving this record stale (or *should*) and the
+    // ttd has passed increase ttd to the future and remember that
+    // we did. Also push a refresh task.
+    if ((serveStale || entry->d_servedStale > 0) && entry->d_ttd <= now && entry->d_servedStale < s_maxServedStaleExtensions) {
```
rgacogne, line 238: This test could probably be moved to a function

pdns/recursor_cache.cc
```diff
@@ -152,7 +154,31 @@ time_t MemRecursorCache::handleHit(MapCombo::LockedContent& content, MemRecursor
...
+
+void MemRecursorCache::updateStaleEntry(time_t now, MemRecursorCache::OrderedTagIterator_t& entry)
+{
+  // We need to take care a infrequently access stale item cannot be extended past
+  // s_maxServedStaleExtension * s_serveStaleExtensionPeriod
+  // We we look how old the entry is, and increase d_servedStale accordingly, taking care not to overflow
+  const time_t howlong = std::max(static_cast<time_t>(1), now - entry->d_ttd);
+  const uint32_t extension = std::max(1U, std::min(entry->d_orig_ttl, s_serveStaleExtensionPeriod));
+  entry->d_servedStale = std::min(entry->d_servedStale + 1 + howlong / extension, static_cast<time_t>(s_maxServedStaleExtensions));
+  entry->d_ttd = now + std::min(entry->d_orig_ttl, s_serveStaleExtensionPeriod);
```
rgacogne, line 176: I wonder if we should use `extension` (which is at least 1)?
```suggestion
  entry->d_ttd = now + extension;
```

pdns/recursor_cache.cc
```diff
@@ -152,7 +154,31 @@ time_t MemRecursorCache::handleHit(MapCombo::LockedContent& content, MemRecursor
...
+}
+
+void MemRecursorCache::updateStaleEntry(time_t now, MemRecursorCache::OrderedTagIterator_t& entry)
+{
+  // We need to take care a infrequently access stale item cannot be extended past
+  // s_maxServedStaleExtension * s_serveStaleExtensionPeriod
+  // We we look how old the entry is, and increase d_servedStale accordingly, taking care not to overflow
+  const time_t howlong = std::max(static_cast<time_t>(1), now - entry->d_ttd);
+  const uint32_t extension = std::max(1U, std::min(entry->d_orig_ttl, s_serveStaleExtensionPeriod));
+  entry->d_servedStale = std::min(entry->d_servedStale + 1 + howlong / extension, static_cast<time_t>(s_maxServedStaleExtensions));
```
rgacogne, line 175: I'm afraid I don't really understand the `howlong / extension` part?

pdns/recursor_cache.cc
```diff
@@ -335,7 +377,8 @@ time_t MemRecursorCache::get(time_t now, const DNSName& qname, const QType qt, b
       for (auto i = entries.first; i != entries.second; ++i) {
         firstIndexIterator = map->d_map.project<OrderedTag>(i);
 
-        if (i->d_ttd <= now) {
+        // When serving stale, we consider expired records
+        if (i->d_ttd <= now && !serveStale && i->d_servedStale == 0) {
```
rgacogne, line 381: It might be good to create a `isEntryUsable(const CacheEntry& entry, time_t now, bool serveStale)` function, or something like that

pdns/recursor_cache.cc
```diff
@@ -15,6 +15,8 @@
 #include "cachecleaner.hh"
 #include "rec-taskqueue.hh"
 
+uint16_t MemRecursorCache::s_maxServedStaleExtensions;
```
rgacogne, line 18: Hmm, so this file is no longer a symlink, but we now have two versions?

rgacogne: Or is it a GH bug?

pdns/recursor_cache.cc
rgacogne: Ah, it might be because the explanation is in the other version of that file..

rgacogne (commented):
So, when serve-stale is enabled and we fail to get a response, we do increase the TTD by up to 1440s, and during that time we will only try to get a new value once, as a background task, possibly very quickly after the first failed attempt? I'm a bit worried about using a stale record long after the auth has recovered, I wonder if we could retry more often?

The GH UI doesn't allow me to comment on the cursed  pdns/recursordist/recursor_cache.cc version so I'm commenting out of context, sorry: I wonder if a percentage would be easier to understand than a period for `s_serveStaleExtensionPeriod`?

pdns/recursor_cache.cc
omoerbeek: Lemme heck that, I might have used patch(1) locally, which has the habbit of changing a symlink to a file.

omoerbeek:
> 



> So, when serve-stale is enabled and we fail to get a response, we do increase the TTD by up to 1440s, and during that time we will only try to get a new value once, as a background task, possibly very quickly after the first failed attempt? I'm a bit worried about using a stale record long after the auth has recovered, I wonder if we could retry more often?
> 
> The GH UI doesn't allow me to comment on the cursed pdns/recursordist/recursor_cache.cc version so I'm commenting out of context, sorry: I wonder if a percentage would be easier to understand than a period for `s_serveStaleExtensionPeriod`?

The extension is 30s (or lower if the original TTL was lower) each time.  THe numbe rof times an extension occurs ia 1404 max, by default. 

> So, when serve-stale is enabled and we fail to get a response, we do increase the TTD by up to 1440s, and during that time we will only try to get a new value once, as a background task, possibly very quickly after the first failed attempt? I'm a bit worried about using a stale record long after the auth has recovered, I wonder if we could retry more often?

The extra _time_ per extension is 30s, or lower if the orignal TTL was lower. The max _number_ of extensions is 1440. So I don't think your worry is warranted. There shoud be a retry scheduled every 30s, if the record keeps gettinn quried. If that is not happening there is a bug....

rgacogne:
OK, then I did not understand it correctly! I think you should discard my comments until we are sure I'm looking at the correct version of recursor_cache.cc, otherwise you might be wasting your time.

omoerbeek:
Right. Will rebase soon as well.

omoerbeek:
`recursor_cache.cc` botch fixed and rebased. I plan to continue to work on this the coming time. Will be doing more tests and plus review of the  approach and code.

pdns/recursor_cache.cc
```diff
@@ -15,6 +15,45 @@
...
+ *
+ * With s_serveStaleExtensionPeriod of 30 seconds, setting s_maxServedStaleExtensions to 1440 will
+ * cause a record to be served stale a maximum of 30s * 1440 = 12 hours. If the original TTL is
+ * smaller than 30, this period wil be shorter. If there was a long time between serve-stale
+ * extensions, the value of d_servedStale will be incremented by more than one to account for the
+ * longer period.
+ *
+ * If serve-stale is enabled, the resolving process first will try to resolve a record in the
+ * ordinary way, with the difference that a timeout will not lead to an ImmediateServFailException
+ * being passed to the caller, but the resolving wil be tried again with a flag to allow marking
```
rgacogne, line 42:
```suggestion
 * being passed to the caller, but the resolving will be tried again with a flag to allow marking
```

pdns/recursor_cache.cc
```diff
@@ -152,7 +191,31 @@ time_t MemRecursorCache::handleHit(MapCombo::LockedContent& content, MemRecursor
...
+  else {
+    pushAlmostExpiredTask(qname, qtype, deadline, netmask);
+  }
+}
+
+void MemRecursorCache::updateStaleEntry(time_t now, MemRecursorCache::OrderedTagIterator_t& entry)
+{
+  // We need to take care a infrequently access stale item cannot be extended past
+  // s_maxServedStaleExtension * s_serveStaleExtensionPeriod
+  // We we look how old the entry is, and increase d_servedStale accordingly, taking care not to overflow
```
rgacogne, line 209:
```suggestion
  // We look how old the entry is, and increase d_servedStale accordingly, taking care not to overflow
```

pdns/recursor_cache.cc
```diff
@@ -152,7 +191,31 @@ time_t MemRecursorCache::handleHit(MapCombo::LockedContent& content, MemRecursor
...
+  const time_t howlong = std::max(static_cast<time_t>(1), now - entry->d_ttd);
+  const uint32_t extension = std::max(1U, std::min(entry->d_orig_ttl, s_serveStaleExtensionPeriod));
+  entry->d_servedStale = std::min(entry->d_servedStale + 1 + howlong / extension, static_cast<time_t>(s_maxServedStaleExtensions));
```
rgacogne, line 212: So this is a bit more clever than `how many times the serve-stale status was extended`, since it takes into account for how long the entry was expired when we decide to use it as stale. Which we have to do not to exceed `s_maxServedStaleExtension * s_serveStaleExtensionPeriod`, given that we don't keep the original TTD around.
I see we check that `entry->d_servedStale < s_maxServedStaleExtensions` before entering this function, but since we can increment `d_servedStale` by more than one it looks like we could exceed `s_maxServedStaleExtensions` once?

pdns/recursordist/test-syncres_cc10.cc
```diff
@@ -1059,4 +1061,261 @@ BOOST_AUTO_TEST_CASE(test_PacketIDCompare)
...
+  // record is expired, so serve stale should kick in
+  ret.clear();
+  res = sr->beginResolve(target, QType(QType::A), QClass::IN, ret);
+  BOOST_CHECK_EQUAL(res, RCode::NoError);
+  BOOST_REQUIRE_EQUAL(ret.size(), 1U);
+  BOOST_CHECK(ret[0].d_type == QType::A);
+  BOOST_CHECK_EQUAL(ret[0].d_name, target);
+  BOOST_CHECK_EQUAL(downCount, 4U);
+  BOOST_CHECK_EQUAL(lookupCount, 1U);
+
```
rgacogne, line 1140: I'm confused, shouldn't we have pushed a refresh task by now?

pdns/recursordist/test-syncres_cc10.cc
```diff
@@ -1059,4 +1061,261 @@ BOOST_AUTO_TEST_CASE(test_PacketIDCompare)
...
+      addRecordToLW(res, "powerdns.com.", QType::NS, "pdns-public-ns2.powerdns.com.", DNSResourceRecord::AUTHORITY, 5);
+      addRecordToLW(res, "pdns-public-ns1.powerdns.com.", QType::A, "192.0.2.2", DNSResourceRecord::ADDITIONAL, 5);
+      addRecordToLW(res, "pdns-public-ns1.powerdns.com.", QType::AAAA, "2001:DB8::2", DNSResourceRecord::ADDITIONAL, 5);
+      addRecordToLW(res, "pdns-public-ns2.powerdns.com.", QType::A, "192.0.2.3", DNSResourceRecord::ADDITIONAL, 5);
+      addRecordToLW(res, "pdns-public-ns2.powerdns.com.", QType::AAAA, "2001:DB8::3", DNSResourceRecord::ADDITIONAL, 5);
+      return LWResult::Result::Success;
+    }
+    else if (ip == ComboAddress("192.0.2.2:53") || ip == ComboAddress("192.0.2.3:53") || ip == ComboAddress("[2001:DB8::2]:53") || ip == ComboAddress("[2001:DB8::3]:53")) {
+      setLWResult(res, 0, true, false, true);
+      addRecordToLW(res, target, QType::A, "192.0.2.4", DNSResourceRecord::ANSWER, 5);
```
rgacogne, line 1104: May I suggest using a constant for the TTL, so that it's clear where the 6 (TTL+1), 12 and 18 later values come from?

rgacogne (commented):
The logic seems quite good, nice work! I had a few questions and comments, but I don't see any major issue.

pdns/recursor_cache.cc
omoerbeek: Line 202 assign `std::min(..., s_maxServedStaleExtension)` to `entry->d_servedStale` , so if the computed value is larger than `s_maxServedStaleExtension` the new value of `entry->d_servedStale` will be `s_maxServedStaleExtension`.

pdns/recursordist/test-syncres_cc10.cc
omoerbeek: Yes, we do. I am just not checking that. The second push still would keep the count to 1, uniqueness constraints prevent the same task to be pushed twice.

pdns/recursor_cache.cc
rgacogne: Uh, right.

pdns/recursor_cache.cc
```diff
@@ -15,6 +15,45 @@
...
+ * SERVE-STALE: the general approach
+ *
+ * The general switch to enable serve-stale is s_maxServedStaleExtensions. If this value is zero, no
+ * serve-stale is done. If it is positive, it determines how many times the serve-stale status of a
+ * record can be extended.
+ *
+ * Each record in the cache has a field d_servedStale. If this value is zero, no special handling is
+ * done. If it is positive, the record is being served stale. The value determines how many times
+ * the serve-stale status was extended. Each time an extension happens, the value is incremented and
+ * a task to see is the record resolves will be pushed. When the served-stale status is extended,
```
Habbie, line 28:
```suggestion
 * a task to see if the record resolves will be pushed. When the served-stale status is extended,
```

pdns/recursor_cache.cc
Habbie: this conversation was resolved but I still see the typo

Habbie (approved):
I read the approach as documented in `recursor_cache.cc`, and this sounds excellent to me. I did spot two typos (one of which Rémi spotted before).

pdns/recursor_cache.cc
omoerbeek: I think I changed another instance of the same typo. To compensate, I found another one and fixed those two for good.

omoerbeek:
Rebased to master to solve conflict

PR #11780, opened 2022-07-12
rec: Save the last nsspeed recorded plus output it in rec_control dump-nsspeeds

### LINKED ISSUE #11736: Add actual last-seen latency to dump-nsspeeds output in recursor
- Program: Recursor
 - Issue type: Feature request

### Short description
In the results of "rec_control dump-nsspeeds /path/to/filename" it would be useful to have the actual last sampled latency as an additional field on each IP address output in the command. Currently, the "latency" is actually a decaying, computed figure instead of the actual latency as seen by the recursor to the distant authoritative server - this is confusing and not very useful for  obvious external automated analysis.

### Usecase
It is interesting (and sometimes required) to understand the latency figures for remote authoritative servers. When network changes are performed, it would be useful in bulk fashion to see what happens before and after a change is implemented when looking at authoritative answers.  Trivial parsing of the text output of dump-nsspeeds can be performed quickly, if that data existed. Currently there is no automated way to do this at scale with data obtained from the resolver (without extensive dnstap, protobuf, or pcap data evaluation) but the resolver is very close to having that data but doesn't quite store what is required for the analysis.  Doing packet captures or log analysis with external tool add-ons are heavy and requires significant code implementation; putting the real figure in the dump-nsspeeds result seems to be much more elegant and automatic, and this work is already being done by the recursor.  Using a running average of results over X seconds (or Y ringbuffer results?) is also an option that would be even better, but requires then at least two data points for each authoritative server - the average, and number of samples collected so that the next result can be averaged into the average. This averaged method may be more complex than necessary, though it would smooth outliers, which may or may not be a desirable result.  If an average is used, then you might as well go "all the way" and do min, max, average, and standard deviation for the sample interval.

### Description
For the purposes of this request, the decaying latency figure can remain the same (and no code change seems to be required there.)

One possible output model would be to add actual latency with a "/" following the computed/decaying latency figure. The exact method by which this is reported in the output is not strongly preferred. Perhaps an entirely different "rec_control" command is warranted. Example using the inclusion-by-/ model:

anyns.pch.net -> 204.61.216.4/14.109665/15.993292 2001:500:14:6004:ad::1/520.199158/533.821582

omoerbeek:
Fixes #11736 in the most basic way. There are some plans to unify
the various NS state tables into a single elaborate data structure,
that would be the moment to store a moving average or similar.

Also modify the output a bit: show times in ms instead of us and
without all the decimals.
Example output:
```
; nsspeed dump follows
; nsname        timestamp       [ip/decaying-ms/last-ms...]
a6-192.akamaiedge.net   2022-07-12T13:59:29.585 23.211.133.192/4.440/8.837
usw6.akam.net   2022-07-12T13:59:25.472 23.61.199.64/423.307/1000.000
ns6-194.akamaiedge.net  2022-07-12T13:59:29.585 95.100.168.194/5.835/14.186
ns5-194.akamaiedge.net  2022-07-12T13:59:29.585 184.85.248.194/11.788/28.661
a11-192.akamaiedge.net  2022-07-12T13:59:29.585 84.53.139.192/6.843/10.938      2600:1480:1::c0/14.585/18.466
a12-192.akamaiedge.net  2022-07-12T13:59:29.585 184.26.160.192/47.739/116.072
...
```
### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/syncres.cc
+++ b/pdns/syncres.cc
@@ -123,13 +123,13 @@ private:
     void submit(int arg, const struct timeval& last, const struct timeval& now)
     {
       d_last = arg;
-      float val = static_cast<float>(arg);
+      auto val = static_cast<float>(arg);
       if (d_val == 0) {
         d_val = val;
       }
       else {
-        float diff = makeFloat(last - now);
-        float factor = expf(diff) / 2.0f; // might be '0.5', or 0.0001
+        auto diff = makeFloat(last - now);
+        auto factor = expf(diff) / 2.0f; // might be '0.5', or 0.0001
         d_val = (1.0f - factor) * val + factor * d_val;
       }
     }
--- a/pdns/syncres.cc
+++ b/pdns/syncres.cc
@@ -120,15 +120,17 @@ private:
   struct DecayingEwma
   {
   public:
-    void submit(int val, const struct timeval& last, const struct timeval& now)
+    void submit(int arg, const struct timeval& last, const struct timeval& now)
     {
+      d_last = arg;
+      float val = static_cast<float>(arg);
       if (d_val == 0) {
-        d_val = static_cast<float>(val);
+        d_val = val;
       }
       else {
         float diff = makeFloat(last - now);
         float factor = expf(diff) / 2.0f; // might be '0.5', or 0.0001
-        d_val = (1.0f - factor) * static_cast<float>(val) + factor * d_val;
+        d_val = (1.0f - factor) * val + factor * d_val;
       }
     }
 
@@ -142,7 +144,13 @@ private:
       return d_val;
     }
 
+    int last(void) const
+    {
+      return d_last;
+    }
+
     float d_val{0};
+    int d_last{0};
   };
 
 public:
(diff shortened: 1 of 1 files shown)


mnordhoff:
If it isn't already explained somewhere, it might be good to document that "1000.000" almost always means a timeout.

omoerbeek:
> If it isn't already explained somewhere, it might be good to document that "1000.000" almost always means a timeout.

It is mentioned in https://docs.powerdns.com/recursor/appendices/internals.html#some-of-the-things-we-glossed-over

pdns/syncres.cc
```diff
@@ -120,15 +120,17 @@ class DecayingEwmaCollection
   struct DecayingEwma
   {
   public:
-    void submit(int val, const struct timeval& last, const struct timeval& now)
+    void submit(int arg, const struct timeval& last, const struct timeval& now)
     {
+      d_last = arg;
+      float val = static_cast<float>(arg);
```
fredmorcos, line 126: `auto val = ...`

fredmorcos (approved):
Small nitpick but that's it

PR #11797, opened 2022-07-19
Add support for EL9 to repo test script

aerique:
### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/build-scripts/docker/repo-test/generate-repo-files.py
+++ b/build-scripts/docker/repo-test/generate-repo-files.py
@@ -24,7 +24,7 @@ from jinja2 import Environment, FileSystemLoader
 
 # Globals
 
-g_version = '1.0.0'
+g_version = '1.0.1'
 
 g_verbose = False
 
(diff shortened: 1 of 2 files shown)


build-scripts/docker/repo-test/generate-repo-files.py
```diff
@@ -172,6 +172,12 @@ def write_release_files (release):
         write_dockerfile('ubuntu', 'jammy', release)
         write_list_file('ubuntu', 'jammy', release)
 
+    if release in ['auth-master',
+                   'rec-master',
+                   'dnsdist-master']:
+        write_pkg_pin_file(release)
```
Habbie, line 178: only Debian/Ubuntu need the pin file, which, now that I look at it, also means it's missing from some other targets, and depending on build order, a dnsdist Dockerfile might get a pdns pin file.

So, there's a bug there, but for this PR, the line can just go :-)

(Usually, the repo package version will be higher than what's in the distro as stock, and then the pin file doesn't matter, which probably explains that we never noticed.)

Habbie (changes requested):

build-scripts/docker/repo-test/generate-repo-files.py
aerique: I was just cargo culting. Moved it to the proper sections now.

Habbie (approved):
Approved, although we still put the wrong pin file in a docker image sometimes. Perhaps the pin file should just list `pdns*` and `dnsdist` always? That can be another PR though.

rgacogne:
I'm afraid this branch needs to be rebased!

aerique:
Old PR and has been superseded by newer versions of the repo test script.

PR #11800, opened 2022-07-21
Log socket directory

fredmorcos:
### Short description
Log socket directory

### Checklist
I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/recursordist/rec-main.cc
+++ b/pdns/recursordist/rec-main.cc
@@ -698,14 +698,14 @@ static void writePid(Logr::log_t log)
 
 static void checkSocketDir(Logr::log_t log)
 {
-  struct stat st;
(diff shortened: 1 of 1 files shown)


omoerbeek:
Nice, but it would be even better if it also would print the right dir if chroot is used or running under systemd (maybe it does already for systemd, I need to check that...).

omoerbeek:
I' included to merge this, even with the regression failure, as it is unrelated and fixed in master.

fredmorcos:
> I'm inclined to merge this, even with the regression failure, as it is unrelated and fixed in master.

Alright, could you approve and merge it?

omoerbeek (approved):

PR #11802, opened 2022-07-22
Upgrade check-spelling to v0.0.20

jsoref:
### Short description
This upgrades check-spelling to [v0.0.20](https://github.com/check-spelling/check-spelling/releases/tag/v0.0.20)

### Checklist

I have:
- [ ] compiled this code
- [x] tested this code
- [ ] documented the code

### Testing
https://github.com/check-spelling/pdns/pull/1

### Notable changes
* comments are collapsed as new runs are performed. A passing run should even collapse the last comment leaving none expanded.
* you can forbid patterns, e.g. `Github`
* duplicate word detection, e.g. `the the` -- note that there can be false positives, it's quite likely that I shouldn't have changed `really really` to `really` -- this can be masked by adjusting `patterns.txt`
* advice refresh -- small tweaks based on feedback
* permissions -- the v0.0.19 workflow relies on default permissions. It's possible for an organization or a repository to set read-only default permissions which results in the action not being able to post comments (and a really ugly error message in v0.0.19) -- the workflow now only requests the permissions it needs (instead of either getting nearly all possible write permissions, or getting only readonly), and it now runs in two phases -- the first phase where it's considering untrusted user input doesn't have write permissions, the second phase where it produces comments has write permissions (to post a comment to a PR or to a commit)
* updating pull request head branches by talking to the bot is enabled for forks (but not for this repository)

### DIFF
--- a/.github/actions/spell-check/line_forbidden.patterns
+++ b/.github/actions/spell-check/line_forbidden.patterns
@@ -1,0 +1,39 @@
+# reject `m_data` as there's a certain OS which has evil defines that break things if it's used elsewhere
+# \bm_data\b
+
+# s.b. GitHub
+\bGithub\b
+
+# s.b. GitLab
+\bGitlab\b
+
+# s.b. JavaScript
+\bJavascript\b
+
+# s.b. Microsoft
+\bMicroSoft\b
+
+# s.b. another
+\ban[- ]other\b
+
+# s.b. greater than
+\bgreater then\b
+
+# s.b. less than
+\bless then\b
+
+# s.b. otherwise
+\bother[- ]wise\b
+
+# s.b. nonexistent
+\bnon existing\b
+\b[Nn]o[nt][- ]existent\b
+
+# s.b. preexisting
+[Pp]re-existing
+
+# s.b. preemptively
+[Pp]re-emptively
+
+# Reject duplicate words
+\s([A-Z]{3,}|[A-Z][a-z]{2,}|[a-z]{3,})\s\g{-1}\s
--- a/.github/actions/spell-check/patterns.txt
+++ b/.github/actions/spell-check/patterns.txt
@@ -26,6 +26,14 @@ addNSECRecordToLW.*DNSName.*powerdnt.com.*QType::NS.*res->d_records
 data:[a-zA-Z=;,/0-9+]+
 BOOST_CHECK_EQUAL\(b64, "[a-zA-Z=;,/0-9+]+"
 
+# acceptable duplicates
+# ls directory listings
+[-bcdlpsw](?:[-r][-w][-sx]){3}\s+\d+\s+(\S+)\s+\g{-1}\s+\d+\s+
+# C types
+\s(long|LONG) \g{-1}\s
+# javadoc / .net
+(?:[\\@](?:groupname|param)|(?:public|private)(?:\s+static|\s+readonly)*)\s+(\w+)\s+\g{-1}\s
+
 # ignore long runs of a single character:
 \b([A-Za-z])\g{-1}{3,}\b
 
--- a/.github/workflows/spelling2.yml
+++ b/.github/workflows/spelling2.yml
@@ -16,17 +16,41 @@ on:
 jobs:
   spelling:
     name: Spell checking
+    permissions:
+      contents: read
+      pull-requests: read
+      actions: read
+    outputs:
+      followup: ${{ steps.spelling.outputs.followup }}
     runs-on: ubuntu-latest
+    if: "contains(github.event_name, 'pull_request') || github.event_name == 'push'"
+    concurrency:
+      group: spelling-${{ github.event.pull_request.number || github.ref }}
+      # note: If you use only_check_changed_files, you do not want cancel-in-progress
+      cancel-in-progress: true
     steps:
-    - name: checkout-merge
-      if: "contains(github.event_name, 'pull_request')"
-      uses: actions/checkout@v2
-      with:
-        ref: refs/pull/${{github.event.pull_request.number}}/merge
-    - name: checkout
-      if: ${{ github.event_name == 'push' }}
-      uses: actions/checkout@v2
-    - uses: check-spelling/check-spelling@v0.0.19
+    - name: check-spelling
       id: spelling
+      uses: check-spelling/check-spelling@v0.0.20
       with:
         config: .github/actions/spell-check
+        suppress_push_for_open_pull_request: 1
+        checkout: true
+        post_comment: 0
+        check_extra_dictionaries: ''
+
+  comment:
+    name: Report
+    runs-on: ubuntu-latest
+    needs: spelling
+    permissions:
+      contents: write
+      pull-requests: write
+    if: (success() || failure()) && needs.spelling.outputs.followup
+    steps:
+    - name: comment
+      uses: check-spelling/check-spelling@v0.0.20
+      with:
+        config: .github/actions/spell-check
+        checkout: true
+        task: ${{ needs.spelling.outputs.followup }}
--- a/docs/changelog/4.1.rst
+++ b/docs/changelog/4.1.rst
@@ -671,7 +671,7 @@ Changelogs for 4.1.x
     :tags: Internals, Bug Fixes
     :pullreq: 5917
 
-    Use ``_exit()`` when we really really want to exit, for example
+    Use ``_exit()`` when we *really* want to exit, for example
     after a fatal error. This stops us dying while we die. A call to
     ``exit()`` will trigger destructors, which may paradoxically stop
     the process from exiting, taking down only one thread, but harming
--- a/docs/changelog/pre-4.0.rst
+++ b/docs/changelog/pre-4.0.rst
@@ -5246,7 +5246,7 @@ Changes
 ^^^^^^^
 
 -  The monitor command **set** no longer allows the changing of
-   non-existent variables.
+   nonexistent variables.
 -  IBM Universal Database DB2 backend now included in source
    distribution (untested!)
 -  Oracle backend now included in source distribution (slightly tested!)
--- a/docs/changelog/pre-4.0.rst
+++ b/docs/changelog/pre-4.0.rst
@@ -5916,7 +5916,7 @@ Bugs fixed
 -  PostgreSQL backend was case sensitive and returned only answers in
    case an exact match was found. The Generic PostgreSQL backend is now
    officially all lower case and zone2sql in PostgreSQL mode enforces
-   this. Documentation has been been updated to reflect the case change.
+   this. Documentation has been updated to reflect the case change.
    Thanks to Maikel Verheijen of Ladot for spotting this!
 -  Documentation bug - postgresql create/index statements created a
    duplicate index. If you've previously copy pasted the commands and
--- a/docs/changelog/pre-4.0.rst
+++ b/docs/changelog/pre-4.0.rst
@@ -3288,7 +3288,7 @@ New features
 -  Added support for DHCID, IPSECKEY and KX records, thanks Norbert
    Sendetzky for the hint. Implemented in `commit
    1144 <http://wiki.powerdns.com/projects/trac/changeset/1144>`__.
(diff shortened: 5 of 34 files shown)


.github/actions/spell-check/line_forbidden.patterns
```diff
@@ -0,0 +1,39 @@
+# reject `m_data` as there's a certain OS which has evil defines that break things if it's used elsewhere
+# \bm_data\b
```
jsoref, line 2: This is purely a sample and can be removed.

It (and a lot of hte other content) comes from
https://github.com/check-spelling/spell-check-this/

.github/actions/spell-check/patterns.txt
```diff
@@ -26,6 +26,14 @@ addNSECRecordToLW.*DNSName.*powerdnt.com.*QType::NS.*res->d_records
 data:[a-zA-Z=;,/0-9+]+
 BOOST_CHECK_EQUAL\(b64, "[a-zA-Z=;,/0-9+]+"
 
+# acceptable duplicates
+# ls directory listings
+[-bcdlpsw](?:[-r][-w][-sx]){3}\s+\d+\s+(\S+)\s+\g{-1}\s+\d+\s+
+# C types
+\s(long|LONG) \g{-1}\s
```
jsoref, line 33: If `really really` is something one wanted to keep, this is the change to prevent the bot from whining about it:
```suggestion
\s(long|LONG|really) \g{-1}\s
```

docs/changelog/4.1.rst
```diff
@@ -671,7 +671,7 @@ Changelogs for 4.1.x
     :tags: Internals, Bug Fixes
     :pullreq: 5917
 
-    Use ``_exit()`` when we really really want to exit, for example
+    Use ``_exit()`` when we really want to exit, for example
```
jsoref, line 674: It's possible that `really really` is an intentional stress (and that I wasn't awake enough when I made these commits):
```suggestion
    Use ``_exit()`` when we really really want to exit, for example
```

.github/workflows/spelling2.yml
```diff
@@ -12,21 +12,83 @@ on:
...
+      with:
+        config: .github/actions/spell-check
+        experimental_apply_changes_via_bot: ${{ github.repository_owner != 'powerdns' && 1 || 0 }}
+        suppress_push_for_open_pull_request: 1
+        checkout: true
+        post_comment: 0
+        extra_dictionaries:
+          cspell:aws/aws.txt
+          cspell:filetypes/filetypes.txt
+          cspell:django/django.txt
```
jsoref, line 46: I can't really explain what django is, but it seems to be a very popular word list (in that it has a lot of words used by a lot of projects, and often covers more words than other dictionaries, even dictionaries that should directly read on a given project).

.github/workflows/spelling2.yml
```diff
@@ -12,21 +12,83 @@ on:
...
+  issue_comment:
+    types: [created]
 
 jobs:
   spelling:
     name: Spell checking
+    permissions:
+      contents: read
+      pull-requests: read
+      actions: read
```
jsoref, line 24: This is the read restriction I mentioned in the PR.

.github/workflows/spelling2.yml
```diff
@@ -12,21 +12,83 @@ on:
...
+          cspell:css/css.txt
+        check_extra_dictionaries: ''
+
+  comment:
+    name: Report
+    runs-on: ubuntu-latest
+    needs: spelling
+    permissions:
+      contents: write
+      pull-requests: write
```
jsoref, line 61: This enables commenting on a commit or PR respectively.

.github/workflows/spelling2.yml
```diff
@@ -12,21 +12,83 @@ on:
...
+        config: .github/actions/spell-check
+        checkout: true
+        task: ${{ needs.spelling.outputs.followup }}
+        experimental_apply_changes_via_bot: ${{ github.repository_owner != 'powerdns' && 1 || 0 }}
+
+  update:
+    name: Update PR
+    permissions:
+      contents: write
+      pull-requests: write
```
jsoref, line 76: This could create commits on branches (as well as interacting w/ comments on PRs or making comments when things go wrong).

.github/workflows/spelling2.yml
```diff
@@ -12,21 +12,83 @@ on:
...
+    runs-on: ubuntu-latest
+    if: ${{
+        github.repository_owner != 'powerdns' &&
```
jsoref, line 79: This constraint means that the workflow won't run in this organization.

-- I tested this by adding an additional exclusion for check-spelling. It works.

.github/workflows/spelling2.yml
```diff
@@ -12,21 +12,83 @@ on:
...
+      pull-requests: write
+    if: (success() || failure()) && needs.spelling.outputs.followup
+    steps:
+    - name: comment
+      uses: check-spelling/check-spelling@v0.0.20
+      with:
+        config: .github/actions/spell-check
+        checkout: true
+        task: ${{ needs.spelling.outputs.followup }}
+        experimental_apply_changes_via_bot: ${{ github.repository_owner != 'powerdns' && 1 || 0 }}
```
jsoref, line 70: These guards prevent the bot from suggesting something that won't work when running from the powerdns repository.

.github/workflows/spelling2.yml
```diff
@@ -12,21 +12,83 @@ on:
...
+      group: spelling-update-${{ github.event.issue.number }}
+      cancel-in-progress: false
+    steps:
+    - name: apply spelling updates
+      uses: check-spelling/check-spelling@v0.0.20
       with:
         config: .github/actions/spell-check
+        experimental_apply_changes_via_bot: ${{ github.repository_owner != 'powerdns' && 1 || 0 }}
+        checkout: true
+        ssh_key: "${{ secrets.CHECK_SPELLING }}"
```
jsoref, line 94: Secrets that don't exist resolve to empty strings. If there's no ssh_key, then the `GITHUB_TOKEN` is used. When that happens, workflows won't run. The bot will provide instructions for how to set up this token.

docs/changelog/pre-4.0.rst
```diff
@@ -5246,7 +5246,7 @@ Changes
 ^^^^^^^
 
 -  The monitor command **set** no longer allows the changing of
-   non-existent variables.
+   nonexistent variables.
```
jsoref, line 5249: If a given thing isn't desired, it can be removed from the forbidden file.

docs/domainmetadata.rst
```diff
@@ -73,7 +73,7 @@ This metadata item controls whether or not a zone is fully rectified on changes
 to the contents of a zone made through the :doc:`API <http-api/index>`.
 
 When the ``API-RECTIFY`` value is "1", the zone will be rectified on changes.
-Any other other value means that it will not be rectified. If this is not set
+Any other value means that it will not be rectified. If this is not set
```
jsoref, line 76: English is a funny language, sometimes doublings have real meaning.

[Buffalo buffalo Buffalo buffalo buffalo buffalo Buffalo buffalo](https://en.wikipedia.org/wiki/Buffalo_buffalo_Buffalo_buffalo_buffalo_buffalo_Buffalo_buffalo).

docs/manpages/pdnsutil.1.rst
```diff
@@ -99,7 +99,7 @@ remove-zone-key *ZONE* *KEY-ID*
     Remove a key with id *KEY-ID* from a zone called *ZONE*.
 set-nsec3 *ZONE* ['*HASH-ALGORITHM* *FLAGS* *ITERATIONS* *SALT*'] [**narrow**]
     Sets NSEC3 parameters for this zone. The quoted parameters are 4
-    values that are used for the the NSEC3PARAM record and decide how
+    values that are used for the NSEC3PARAM record and decide how
```
jsoref, line 102: Often doubled words are just mistakes.

pdns/recursordist/docs/changelog/4.1.rst
```diff
@@ -832,7 +832,7 @@ See :doc:`EOL Statements <../appendices/EOL>`.
     :tags: Internals, Bug Fixes
     :pullreq: 5917
 
-    Use ``_exit()`` when we really really want to exit, for example
+    Use ``_exit()`` when we really want to exit, for example
```
jsoref, line 835:
```suggestion
    Use ``_exit()`` when we really really want to exit, for example
```

jsoref:
I'm not changing `expect.txt` here, once this lands, one needs to push a small change to update it, see:
https://github.com/jsoref/pdns/commit/90195a06aec35951df0589f014314b71ad139f32#commitcomment-79144916

It's really hard to make two different versions of the engine agree about the state of the world.

.github/actions/spell-check/patterns.txt
jsoref: Spoke w/ @Habbie and he's ok w/ only one really, so this change isn't necessary.

docs/changelog/4.1.rst
jsoref: We'll go w/ some stress instead:
```suggestion
    Use ``_exit()`` when we *really* want to exit, for example
```

.github/workflows/spelling2.yml
jsoref: I'm going to split this change out into a separate PR.

.github/workflows/spelling2.yml
jsoref: I'm going to split this into a separate PR.

.github/workflows/spelling2.yml
Habbie: `actions: read` does not provide read access to secrets?

.github/workflows/spelling2.yml
Habbie: does it also allow any other kind of write? to other PRs perhaps?

Habbie: what does `contents: write` enable?

Habbie (commented):
All text changes look good. Some remaining questions about permissions.

.github/workflows/spelling2.yml
jsoref: Correct, technically it provides the items listed in this section:
https://docs.github.com/en/rest/overview/permissions-required-for-github-apps#permission-on-actions

(Of course, if a workflow is authorized and it explicitly has `${{ secrets.TOKEN }}` it will get that token when it runs.)

Secrets are guarded by a reasonably well named permission:
https://docs.github.com/en/rest/overview/permissions-required-for-github-apps#permission-on-secrets

.github/workflows/spelling2.yml
jsoref: Sadly way too much (it includes the ability to push to the repository). It's here to enable a comment to be posted on `push`.
https://docs.github.com/en/rest/overview/permissions-required-for-github-apps#permission-on-contents

It's really unfortunate that github doesn't have a `comment` permission and instead lumps the ability to `comment` into either `contents: write` (for commenting on  commits) or `pull-requests: write` (for commenting on pull requests).

If this is a concern, this could be split into two different jobs.

One job could run for forks and include `contents: write`, and a different job could run for the primary repository and not include it.

The privilege isolation means that the code that's running should only be posting a comment based on the data generated by the job which takes untrusted input and checks its spelling. -- It should be possible for you to review the commenting codepath if you're concerned (and I'd welcome that review).

I'm going to investigate using [`GITHUB_STEP_SUMMARY`](
https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#adding-a-job-summary) instead of comments (in addition to trying to get sarif to work). I think it might be possible.

Also note that instead of using `@v0.0.20` I could change the commit to use the equivalent SHA if you're concerned that the tag could move -- I have no intention of ever moving a tag, and w/ this version if I need to disable a release, I can do it via [secpoll](https://github.com/check-spelling/check-spelling/wiki/Feature:-Security-Polling).

Habbie:
Not merging this - holding out for a version that needs even fewer permissions.

PR #11823, opened 2022-08-03
Handle Lua script loading errors

fredmorcos:
### Short description
Handle errors when loading Lua scripts.

### Checklist
I have:
- [X] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/lua-base4.cc
+++ b/pdns/lua-base4.cc
@@ -17,12 +17,13 @@
 BaseLua4::BaseLua4() {
 }
 
-int BaseLua4::loadFile(const std::string &fname) {
+int BaseLua4::loadFile(const std::string& fname)
+{
   int ret = 0;
   std::ifstream ifs(fname);
   if (!ifs) {
     ret = errno;
-    SLOG(g_log<<Logger::Error<<"Unable to read configuration file from '"<<fname<<"': "<<stringerror(ret)<<endl,
+    SLOG(g_log << Logger::Error << "Unable to read configuration file from '" << fname << "': " << stringerror(ret) << endl,
          g_slog->withName("lua")->error(Logr::Error, ret, "Unable to read configuration file", "file", Logging::Loggable(fname)));
     return ret;
   }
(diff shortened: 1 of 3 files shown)


omoerbeek (commented):
This looks good to me, but @Habbie needs to review too, as this is a change in behaviour wrt the lua backend: previously only an error was logged, and now it throws if the Lua file cannot be found or is not accessible.

pdns/lua-base4.cc
```diff
@@ -17,18 +18,18 @@
...
+void BaseLua4::loadFile(const std::string& fname)
 {
-  int ret = 0;
   std::ifstream ifs(fname);
   if (!ifs) {
-    ret = errno;
-    SLOG(g_log << Logger::Error << "Unable to read configuration file from '" << fname << "': " << stringerror(ret) << endl,
-         g_slog->withName("lua")->error(Logr::Error, ret, "Unable to read configuration file", "file", Logging::Loggable(fname)));
-    return ret;
+    assert(errno != 0);
```
Habbie, line 25: does this assert have a risk of stopping the whole process?

Habbie: or is the point that a failing ifstream that leaves errno zero is a very clear bug, in which case exiting is fine?

Habbie (commented):
One question, looks good otherwise!

pdns/lua-base4.cc
fredmorcos: It's the latter, yes.

omoerbeek: That's what I thought as well. Though on second thought: while system calls are guaranteed to set `errno` on error, in general lib functions are not. So I prefer the assert to be removed.

fredmorcos: Alright, done. I fixed-up and force-pushed, I hope that's OK since the change is simple enough.

Habbie: :+1:

Habbie (approved):

PR #11826, opened 2022-08-03
auth: somewhat cleanup global symbols

zeha:
### Short description
- Make many things in auth-main.cc static
- Rename some of the single/two-letter globals to longer names, but mostly those local to auth-main.cc
- Move SOAData, DNSBackend to auth-only include files

I've deliberately ignored the symbols used across many files (QC, PC, ...), in the hope this could still go into 4.7.0.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


pdns/auth-main.cc
```diff
@@ -117,24 +117,25 @@ AuthPacketCache PC; //!< This is the main PacketCache, shared across all threads
 AuthQueryCache QC;
 AuthZoneCache g_zoneCache;
 std::unique_ptr<DNSProxy> DP{nullptr};
-std::unique_ptr<DynListener> dl{nullptr};
+static std::unique_ptr<DynListener> g_DynListener{nullptr};
```
rgacogne, line 120: Just a nit but now that the variable is local to this file, the `g_` prefix might no longer be appropriate :) `s_DynListener` ?
Same for `g_TcpNameserver`, `g_udpNameserver` and `g_udpReceivers`.

pdns/auth-main.hh
```diff
@@ -35,22 +35,13 @@
 #include "tcpreceiver.hh"
 #include "dnsseckeeper.hh"
 
+extern time_t s_starttime;
```
rgacogne, line 38:
```suggestion
extern time_t g_starttime;
```

rgacogne (approved):
Looks good, a couple nits about the prefixes but nothing that really matters.

pdns/auth-main.cc
zeha: Right, wasn't sure if they should be `s_` then. I'll change that, and maybe move some into mainthread() and out of the global scope.

pdns/auth-main.hh
zeha: done, thanks!

pdns/auth-main.cc
zeha: all done.

pdns/auth-main.cc
```diff
@@ -127,14 +127,15 @@ static vector<std::shared_ptr<UDPNameserver>> g_udpReceivers;
 NetmaskGroup g_proxyProtocolACL;
 size_t g_proxyProtocolMaximumSize;
 
-void mainthread();
+// auth-carbon.cc
+void carbonDumpThread();
```
Habbie, line 131: this avoids having `auth-carbon.hh`? maybe add a comment explaining that

Habbie (commented):
one small question, otherwise approved. Happy to get this into 4.7 too.

pdns/auth-main.cc
zeha: yes. added a comment

zeha:
Rebased to fix conflict.

PR #11830, opened 2022-08-04
Also reconnect on ENETUNREACH.

asgeirn:
Ref. #4155

### Short description
Also reconnect on ENETUNREACH.

On Ubuntu 22.04/dnsdist 1.6.1 I get `ENETUNREACH` errno for backend servers only accessible through VPN.

I hid these backends in a `PoolAvalilableRule`, and this works, but when VPN is up again, all DNS packets to the backends are dropped.

After adding the check for `ENETUNREACH` dnsdist is able to resume using these backends.

I've added a `strerror()` wrap around the errno for good measure, as I found the numeric errno less than helpful.

### Checklist
I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdist.cc
+++ b/pdns/dnsdist.cc
@@ -1091,7 +1091,7 @@ ssize_t udpClientSendRequestToBackend(const std::shared_ptr<DownstreamState>& ss
 
   if (result == -1) {
     int savederrno = errno;
-    vinfolog("Error sending request to backend %s: %s", ss->d_config.remote.toStringWithPort(), strerror(savederrno));
+    vinfolog("Error sending request to backend %s: %s", ss->d_config.remote.toStringWithPort(), stringerror(savederrno));
 
     /* This might sound silly, but on Linux send() might fail with EINVAL
        if the interface the socket was bound to doesn't exist anymore.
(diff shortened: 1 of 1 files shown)


pdns/dnsdist.cc
```diff
@@ -1091,14 +1091,14 @@ ssize_t udpClientSendRequestToBackend(const std::shared_ptr<DownstreamState>& ss
 
   if (result == -1) {
     int savederrno = errno;
-    vinfolog("Error sending request to backend %s: %d", ss->d_config.remote.toStringWithPort(), savederrno);
+    vinfolog("Error sending request to backend %s: %s", ss->d_config.remote.toStringWithPort(), strerror(savederrno));
```
rgacogne, line 1094: We try to use our own wrapper, `stringerror`, instead of `strerror`, although we might still have some remaining calls to `strerror`:
```suggestion
    vinfolog("Error sending request to backend %s: %s", ss->d_config.remote.toStringWithPort(), stringerror(savederrno));
```

rgacogne (changes requested):
Thanks, one small suggestion but the rest looks good!

rgacogne (approved):

PR #11843, opened 2022-08-10
Clear the caches *after* loading authzones.

omoerbeek:
### Short description

Fix a race between (re)loading an authzone and wiping the caches.
### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [X] documented the code

### DIFF
--- a/pdns/reczones.cc
+++ b/pdns/reczones.cc
@@ -311,15 +311,18 @@ string reloadZoneConfiguration()
       }
     }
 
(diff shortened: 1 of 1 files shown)


rgacogne (approved):

paddg:
Your solution works for us, the problem has gone. Thanks!! We would be grateful if you could backport this to 4.7.x if possible.

omoerbeek:
Thanks for testing! This will be backported to 4.7.x.

PR #11844, opened 2022-08-10
dnsdist: Do not keep the mplexer created for the initial health-check around

rgacogne:
### Short description

We no longer need it after that initial health-check round so it is only wasting (a very small amount of) memory and one file descriptor at that point.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


Habbie:
CI failure is unrelated

Habbie (approved):

PR #11849, opened 2022-08-11
Rec: consider dns64 processing in more cases than Rcode == NoError

omoerbeek:
### Short description

RFC 6147 section 5.1 says: all RCodes except NXDomain should be candidate for dns64.
The current code only considers dns64 if RCode == NoError

We also restructure things a bit to get rid of most of the repeated lines handling `policyResult`.

To be added: tests

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/pdns_recursor.cc
+++ b/pdns/pdns_recursor.cc
@@ -1178,26 +1178,31 @@ void startDoResolve(void* p)
 
       if (dc->d_luaContext) {
         PolicyResult policyResult = PolicyResult::NoAction;
-        if (res == RCode::NoError) {
-          if (answerIsNOData(dc->d_mdp.d_qtype, res, ret)) {
-            if (dc->d_luaContext->nodata(dq, res, sr.d_eventTrace)) {
-              shouldNotValidate = true;
-              policyResult = handlePolicyHit(appliedPolicy, dc, sr, res, ret, pw, tcpGuard);
-            }
+        if (answerIsNOData(dc->d_mdp.d_qtype, res, ret)) {
+          if (dc->d_luaContext->nodata(dq, res, sr.d_eventTrace)) {
+            shouldNotValidate = true;
+            policyResult = handlePolicyHit(appliedPolicy, dc, sr, res, ret, pw, tcpGuard);
           }
         }
         else if (res == RCode::NXDomain && dc->d_luaContext->nxdomain(dq, res, sr.d_eventTrace)) {
           shouldNotValidate = true;
           policyResult = handlePolicyHit(appliedPolicy, dc, sr, res, ret, pw, tcpGuard);
         }
-
         if (policyResult == PolicyResult::HaveAnswer) {
           goto haveAnswer;
         }
         else if (policyResult == PolicyResult::Drop) {
           return;
         }
+      } // dc->d_luaContext
 
+      if (g_dns64Prefix && dc->d_mdp.d_qtype == QType::AAAA && !vStateIsBogus(dq.validationState) && dns64Candidate(dc->d_mdp.d_qtype, res, ret)) {
+        res = getFakeAAAARecords(dq.qname, *g_dns64Prefix, ret);
+        shouldNotValidate = true;
+      }
+
+      if (dc->d_luaContext) {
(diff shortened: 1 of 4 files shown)


pdns/pdns_recursor.cc
```diff
@@ -1166,60 +1176,53 @@ void startDoResolve(void* p)
...
+        }
+        if (policyResult == PolicyResult::HaveAnswer) {
+          goto haveAnswer;
+        }
+        else if (policyResult == PolicyResult::Drop) {
+          return;
+        }
+      } // dc->d_luaContext
+
+      if (g_dns64Prefix && dc->d_mdp.d_qtype == QType::AAAA && !vStateIsBogus(dq.validationState) && dns64Candidate(dc->d_mdp.d_qtype, res, ret)) {
```
rgacogne, line 1223: If I'm not mistaken after that change `postresolve` and `postresolve_ffi` will no longer see the result of the DNS64 processing, is that intended?

omoerbeek: Hmm, will look into this. I might have gone too far.

rgacogne (commented):
So our dns64 code will now be executed even if the `nodata` hook returned `true` (and perhaps even if `nxdomain` returned `true` after changing the `rcode`), and I don't think it was the case before? I could not find a description of the expected behaviour in the documentation, perhaps we need to think about it and update the docs?

omoerbeek:
Yes, the docs do not say anything about the intercation between Lua hooks and native dns64. We should decide on the desired behaviour and document that (and likely fix the PR again).

omoerbeek:
I thin this is what we want to document:

1. Native DNS64 processing is done after potential nodata()/nxdomain() Lua hooks, but only if they did not return true.
2. Native DNS64 processing is only done: if 
 - prefix is set up
 - original query was an AAAA query
 - The rcode is not NxDomain and no result of the AAAA query was obtained (either no answer records or other error).
 - if DNSSEC processing is active a NoError result should not be Bogus. (BTW: Should we consider the CD flag, this is not done ATM?)

rgacogne:
That sounds good to me! 

> Should we consider the CD flag and dnssec mode, this is not done ATM....?

Right, checking whether validation is actually enabled would make sense to me indeed.

omoerbeek:
I have to fix the added regerssion test, as it cannot set the RCode from Lua anymore.

pdns/pdns_recursor.cc
```diff
@@ -1166,61 +1176,57 @@ void startDoResolve(void* p)
...
-              shouldNotValidate = true;
-              auto policyResult = handlePolicyHit(appliedPolicy, dc, sr, res, ret, pw, tcpGuard);
-              // haveAnswer case redundant
-              if (policyResult == PolicyResult::Drop) {
-                return;
-              }
-            }
-          }
-          else if (dc->d_luaContext->postresolve(dq, res, sr.d_eventTrace)) {
+      if (!luaHookHandled && g_dns64Prefix && dc->d_mdp.d_qtype == QType::AAAA && (!shouldNotValidate || !sr.isDNSSECValidationRequested() || !vStateIsBogus(dq.validationState)) && dns64Candidate(dc->d_mdp.d_qtype, res, ret)) {
```
rgacogne, line 1202:
```suggestion
      if (!luaHookHandled && g_dns64Prefix && dc->d_mdp.d_qtype == QType::AAAA && (shouldNotValidate || !sr.isDNSSECValidationRequested() || !vStateIsBogus(dq.validationState)) && dns64Candidate(dc->d_mdp.d_qtype, res, ret)) {
```

rgacogne: If I'm not mistaken the condition is wrong, we don't have to consider the validation state when `shouldNotValidate` is set?

rgacogne (changes requested):

pdns/pdns_recursor.cc
omoerbeek: Right, thanks for spotting that!

rgacogne (approved):

PR #11850, opened 2022-08-11
Fix recursor not responsive after Lua config reload

### LINKED ISSUE #11795: Recursor stops responding after `rec_control reload-lua-config`
- Program: Recursor
 - Issue type: Bug report

### Short description
Recursor stops responding after `rec_control  reload-lua-config`

### Environment
 - Operating system:  SLES15-SP4
 - Software version: PowerDNS Recursor 4.7.1
 - Software source: PowerDNS repository, [self compiled](https://github.com/PowerDNS/pdns/files/9181220/powerdns-recursor_4.7.1_build.log) (build-log)

### Steps to reproduce
```
mkdir -p tmp; cd tmp
> recursor.conf

echo 'dnstapFrameStreamServer( "10.20.30.40:1234" )' > lua.conf

pdns_recursor --daemon=no --config-dir=. --socket-dir=. --local-address=127.0.0.53 --lua-config-file=./lua.conf

dig +tries=1 +time=1 @127.0.0.53 localhost  ## WORKS

rec_control --socket-dir=. reload-lua-config

dig +tries=1 +time=1 @127.0.0.53 localhost  ## FAILS (timeout)


```

### Expected behaviour
Recursor should never stop responding after `reload-lua-config`

### Actual behaviour
Recursor does not work anymore after `reload-lua-config` on circumstances.

### Other information
To trigger the error, the FrameStreamServer IP address must *not* be a local address of the server itself, but a external one. It doesn't matter if the address answers or not.

There is nothing in the journal what indicates an error.

fredmorcos:
### Short description
- Avoid reloading Lua config if it hasn't changed
- Asynchronously destroys dnstap connections from the old config in case of reload

Fixes #11795

### Checklist
I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/rec-lua-conf.cc
+++ b/pdns/rec-lua-conf.cc
@@ -43,6 +43,27 @@ LuaConfigItems::LuaConfigItems()
 
 /* DID YOU READ THE STORY ABOVE? */
 
+bool operator==(const FrameStreamExportConfig& configA, const FrameStreamExportConfig& configB)
+{
+  // clang-format off
+  return configA.enabled              == configB.enabled              &&
+         configA.logQueries           == configB.logQueries           &&
+         configA.logResponses         == configB.logResponses         &&
+         configA.bufferHint           == configB.bufferHint           &&
+         configA.flushTimeout         == configB.flushTimeout         &&
+         configA.inputQueueSize       == configB.inputQueueSize       &&
+         configA.outputQueueSize      == configB.outputQueueSize      &&
+         configA.queueNotifyThreshold == configB.queueNotifyThreshold &&
+         configA.reopenInterval       == configB.reopenInterval       &&
+         configA.servers              == configB.servers;
+  // clang-format on
+}
+
+bool operator!=(const FrameStreamExportConfig& configA, const FrameStreamExportConfig& configB)
+{
+  return !(configA == configB);
+}
+
 template <typename C>
 typename C::value_type::second_type constGet(const C& c, const std::string& name)
 {
--- a/pdns/recursordist/rec-main.cc
+++ b/pdns/recursordist/rec-main.cc
@@ -58,8 +58,7 @@ static thread_local uint64_t t_protobufServersGeneration;
 static thread_local uint64_t t_outgoingProtobufServersGeneration;
 
 #ifdef HAVE_FSTRM
-thread_local std::shared_ptr<std::vector<std::unique_ptr<FrameStreamLogger>>> t_frameStreamServers{nullptr};
-thread_local uint64_t t_frameStreamServersGeneration;
+thread_local FrameStreamServersInfo t_frameStreamServersInfo;
 #endif /* HAVE_FSTRM */
 
 string g_programname = "pdns_recursor";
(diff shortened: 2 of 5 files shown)


pdns/rec-lua-conf.cc
```diff
@@ -43,6 +43,28 @@ LuaConfigItems::LuaConfigItems()
 
 /* DID YOU READ THE STORY ABOVE? */
 
+bool operator==(const FrameStreamExportConfig& configA, const FrameStreamExportConfig& configB)
+{
+  // clang-format off
+  return configA.servers.size()       == configB.servers.size()       &&
```
omoerbeek, line 49: The size check is redundant, part of the vector `operator==`

pdns/recursordist/rec-main.cc
```diff
@@ -626,31 +625,41 @@ static std::shared_ptr<std::vector<std::unique_ptr<FrameStreamLogger>>> startFra
...
+}
+
 bool checkFrameStreamExport(LocalStateHolder<LuaConfigItems>& luaconfsLocal)
 {
   if (!luaconfsLocal->frameStreamExportConfig.enabled) {
-    if (t_frameStreamServers) {
+    if (t_frameStreamServersInfo.servers) {
       // dt's take care of cleanup
-      t_frameStreamServers.reset();
+      asyncFrameStreamLoggersCleanup(std::move(t_frameStreamServersInfo.servers));
```
omoerbeek, line 641:
```suggestion
      asyncFrameStreamLoggersCleanup(std::move(t_frameStreamServersInfo.servers));
      t_frameStreamServersInfo.config = luaconfsLocal->frameStreamExportConfig;
```

pdns/recursordist/rec-main.cc
```diff
@@ -626,31 +625,41 @@ static std::shared_ptr<std::vector<std::unique_ptr<FrameStreamLogger>>> startFra
...
 
-    auto log = g_slog->withName("dnstap");
-    t_frameStreamServers = startFrameStreamServers(luaconfsLocal->frameStreamExportConfig, log);
-    t_frameStreamServersGeneration = luaconfsLocal->generation;
+    auto dnsTapLog = g_slog->withName("dnstap");
+    t_frameStreamServersInfo.servers = startFrameStreamServers(luaconfsLocal->frameStreamExportConfig, dnsTapLog);
+    t_frameStreamServersInfo.config = luaconfsLocal->frameStreamExportConfig;
   }
 
+  t_frameStreamServersInfo.generation = luaconfsLocal->generation;
```
omoerbeek, line 661: This statement can be moved into the `if` block I believe

omoerbeek (changes requested):
As this is also supposed to compile without  HAVE_FSTRM ...
`FrameStreamExportConfig` and its methods can be moved into `rec-main.cc`
`struct FrameStreamServersInfo` can be moved into `rec-main.hh` inside the existing `HAVE_FSTRM` block.

pdns/recursordist/rec-main.cc
fredmorcos: Is that change because the config was previously enabled and now it isn't and so we need to update it?

fredmorcos: Right. Thanks!

omoerbeek (approved):

PR #11851, opened 2022-08-11
dnsdist: Only call getsockname() once per incoming DoH connection

rgacogne:
### Short description

The current code is calling `h2o_socket_getpeername()` and `h2o_socket_getsockname() `once per DoH _query_, and while the former is cheap because h2o caches the result for us, the latter is actually expensive, so this code caches both values so that we only retrieve
them once per DoH connection.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


Habbie (approved):
Very nice, I like the defer until the first query, doesn't get much more efficient than that!

PR #11852, opened 2022-08-12
dnsdist: Slightly reduce contention around a pool's servers

rgacogne:
### Short description

We only need to take the lock to get the shared pointer, as the actual content is guaranteed not to change, so we do not need to hold the lock while we iterate over the servers list to check whether they are up, or what their current outstanding count is.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


omoerbeek (approved):
You might consider start using `std::shared_ptr<const ServerPolicy::NumberedServerVector>` From a quick  look, this will need a few changes here and there, but not *too* much. The big benefit is it will make clear that the vector  does not change. It might even prevent mistakes...
But OK anyway.

rgacogne:
Good idea, thanks, I'll give it a try!

rgacogne:
Done!

omoerbeek (approved):

PR #11860, opened 2022-08-17
Auth: fix libcrypto handling in automake files

omoerbeek:
### Short description

libcrypto related files are not available in the system include dir on MacOS, so make sure we pickup the proper dir.
All backends compile tested, except ldap (gives other unrelated errors) and tinydb (not available in brew).

For a few other cases  adding  LD flags is also required.

### Checklist

I have:
- [X] compiled this code
- [ ] tested this code
- [ ] documented the code


Habbie (approved):
tested on Debian 11, including ldap and tinydns

PR #11862, opened 2022-08-17
libssl: Properly load ciphers and digests with OpenSSL 3.0

### LINKED ISSUE #11853: DNSDist: Failed to load TLS server certificate
I've been using DNSDist without issue for over a year now, but I just updated it and now it's having an issue with the Let's Encrypt/certbot fullchain certificate and fails to start. The error message provided is as follows:

40D074D4A37F0000:error:0A00018E:SSL routines:SSL_CTX_use_certificate:ca md too weak:../ssl/ssl_rsa.c:221:
Fatal error: Error setting up TLS context for DoH listener on '66.70.208.168:443': An error occurred while trying to load the TLS server certificate file: /etc/dnsdist/fullchain.pem

From some research I did, I saw a similar issue existed for OpenVPN with MD5 certificates, and required certificate to be SHA256, but I confirmed my certificate has a Signature Algorithm of sha256WithRSAEncryption. Can someone assist me with this?

And yes, the files have the correct permissions:
-rw-r--r-- 1 root _dnsdist 6254 Aug 13 17:06 fullchain.pem
-rw-r----- 1 root _dnsdist 3272 Aug 13 17:06 privkey.pem

Thank you

rgacogne:
### Short description

Since https://github.com/openssl/openssl/commit/661595ca0933fe631faeadd14a189acd5d4185e0
we can no longer rely on the ciphers and digests required for TLS to be loaded by `OPENSSL_init_ssl()`, so let's give up and load everything.
I hope you have a lot of RAM.

Hopefully fixes #11853 (I don't have the exact same certificate to be 100% sure).

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/pdns/libssl.cc
+++ b/pdns/libssl.cc
@@ -83,11 +83,19 @@ void registerOpenSSLUser()
 #ifdef HAVE_OPENSSL_INIT_CRYPTO
     /* load the default configuration file (or one specified via OPENSSL_CONF),
        which can then be used to load engines.
-       Do not load all ciphers and digests, we only need a few of them and these
+    */
+#if defined(OPENSSL_VERSION_MAJOR) && OPENSSL_VERSION_MAJOR >= 3
+    /* Since 661595ca0933fe631faeadd14a189acd5d4185e0 we can no longer rely on the ciphers and digests
+       required for TLS to be loaded by OPENSSL_init_ssl(), so let's give up and load everything */
+    OPENSSL_init_crypto(OPENSSL_INIT_LOAD_CONFIG, nullptr);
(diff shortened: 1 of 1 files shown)


rgacogne:
I reverted the logic since `OPENSSL_VERSION_MAJOR` was not defined before OpenSSL 3

omoerbeek (approved):
Tested this with recursor using DoT for both OpenSSL 1.1 and 3.0 on MacOS. Both work fine. Also tested libressl on OpenBSD current.

It is sad though we have to do this but I see n other way.

PR #11866, opened 2022-08-19
Rec: proxymapping stats

### LINKED ISSUE #11648: Implement metrics for ProxyMapping
- Program: Recursor
- Issue type: Feature request

### Short description
Add metrics for ProxyMapping

### Usecase
Visual inspection of ProxyMapping usage.

### Description
Expose metrics via Prometheus and rec_control about queries that were processes by ProxyMapping thing.

omoerbeek:
### Short description

This implements two counters per ProxyMapping subnet. First counter (netmaskMatches) counts the number of hits based on netmask. The second counter counts the hits of domain (if relevant for the mapping).

I did not want to introduce another atomic count, as one day we will be moving to an mcounter like approach.

Actually this PR contains two approaches avoiding atomic counters:
 - The first part (commit 1-3)). Creates a per-thread map of netmask->counts.
This has the drawback that for each ProxyMapping match, we also need to do a lookup in the counts map to find the counters to increment.
- The second approach (last commit) changes the ProxyMapping config to be per-thread (in master it is a shared pointer per thread, all these pointers point to a common config object). That way we can keep the counts in the config object and we do not need the extra lookup. Does need a bit of changing types. We do make sure old counts are not lost on reload.

Example Prometheus output:
```
# TYPE pdns_recursor_proxy_mapping_total counter
# HELP pdns_recursor_proxy_mapping_total Number of queries matching proxyMappings
pdns_recursor_proxy_mapping_total{netmask="127.0.0.0/24",count="netmaskmatches"} 1
pdns_recursor_proxy_mapping_total{netmask="192.168.178.0/24",count="netmaskmatches"} 0
pdns_recursor_proxy_mapping_total{netmask="192.168.178.6/32",count="netmaskmatches"} 0
pdns_recursor_proxy_mapping_total{netmask="192.168.178.99/32",count="netmaskmatches"} 0
pdns_recursor_proxy_mapping_total{netmask="127.0.0.0/24",count="suffixmatches"} 1
pdns_recursor_proxy_mapping_total{netmask="192.168.178.0/24",count="suffixmatches"} 0
pdns_recursor_proxy_mapping_total{netmask="192.168.178.6/32",count="suffixmatches"} 0
pdns_recursor_proxy_mapping_total{netmask="192.168.178.99/32",count="suffixmatches"} 0
```
There is also
```
$ ./rec_control get-proxymapping-stats
127.0.0.0/24	1	1
192.168.178.0/24	0	0
192.168.178.6/32	0	0
192.168.178.99/32	0	0
```
<strike>One potential issue is that the suffixmatches are only incremented on non-PC hits. This has to do with where the suffix the matching is done: after the PC check. I have to see if that is fixable in an easy way.</strike>

Fixes  #11648

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/pdns_recursor.cc
+++ b/pdns/pdns_recursor.cc
@@ -1843,10 +1843,10 @@ bool checkForCacheHit(bool qnameParsed, unsigned int tag, const string& data,
 
     // This is only to get the proxyMapping suffixMatch stats right i the case of a PC hit
     if (t_proxyMapping && source != mappedSource) {
-      if (auto it = t_proxyMapping->lookup(source)) {
-        if (it->second.suffixMatchNode) {
-          if (it->second.suffixMatchNode->check(qname)) {
-            ++it->second.stats.suffixMatches;
+      if (const auto* found = t_proxyMapping->lookup(source)) {
+        if (found->second.suffixMatchNode) {
+          if (found->second.suffixMatchNode->check(qname)) {
+            ++found->second.stats.suffixMatches;
           }
         }
       }
--- a/pdns/pdns_recursor.cc
+++ b/pdns/pdns_recursor.cc
@@ -963,6 +963,9 @@ void startDoResolve(void* p)
             // No match in domains, use original source
             useMapped = false;
           }
+          else {
+            ++it->second.stats.suffixMatches;
+          }
         }
         // No suffix match node defined, use mapped address
       }
@@ -1812,7 +1815,7 @@ bool checkForCacheHit(bool qnameParsed, unsigned int tag, const string& data,
                       DNSName& qname, uint16_t& qtype, uint16_t& qclass,
                       const struct timeval& now,
                       string& response, uint32_t& qhash,
-                      RecursorPacketCache::OptPBData& pbData, bool tcp, const ComboAddress& source)
+                      RecursorPacketCache::OptPBData& pbData, bool tcp, const ComboAddress& source, const ComboAddress& mappedSource)
 {
   if (!t_packetCache) {
     return false;
@@ -1835,6 +1838,17 @@ bool checkForCacheHit(bool qnameParsed, unsigned int tag, const string& data,
       }
       if (t_bogusqueryring) {
         t_bogusqueryring->push_back({qname, qtype});
+      }
+    }
+
+    // This is only to get the proxyMapping suffixMatch stats right i the case of a PC hit
+    if (t_proxyMapping && source != mappedSource) {
+      if (auto it = t_proxyMapping->lookup(source)) {
+        if (it->second.suffixMatchNode) {
+          if (it->second.suffixMatchNode->check(qname)) {
+            ++it->second.stats.suffixMatches;
+          }
+        }
       }
     }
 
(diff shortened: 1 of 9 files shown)


omoerbeek:
After some more thinking I like the second approach much better. So I mark this as ready for review. I'll squash before merge. Note that the addition of a hash function to `Netmask` is not needed for the second approach, as I'm not using an `unordered_map` with key `Netmask` there.

pdns/rec_channel_rec.cc
```diff
@@ -928,6 +928,27 @@ static uint64_t doGetThreadCPUMsec(int n)
   return tt.times.at(n);
 }
 
+static ProxyMappingStats_t* pleaseGetProxyMappingStats()
+{
+  auto ret = new ProxyMappingStats_t;
```
fredmorcos, line 933: Why does this function return a raw pointer instead of returning, say, an owned object or a unique_ptr?

pdns/rec_channel_rec.cc
```diff
@@ -928,6 +928,27 @@ static uint64_t doGetThreadCPUMsec(int n)
...
+      ret->emplace(std::make_pair(key, ProxyMappingCounts{entry.stats.netmaskMatches, entry.stats.suffixMatches}));
+    }
+  }
+  return ret;
+}
+
+static string doGetProxyMappingStats()
+{
+  ostringstream ret;
+  auto m = broadcastAccFunction<ProxyMappingStats_t>(pleaseGetProxyMappingStats);
```
fredmorcos, line 945: Just a nitpick, but perhaps renaming `m` to `proxyMappingStats` would be better.

pdns/rec_channel_rec.cc
```diff
@@ -1138,6 +1159,26 @@ static StatsMap toRPZStatsMap(const string& name, LockGuarded<std::unordered_map
   return entries;
 }
 
+static StatsMap toProxyMappingStatsMap(const string& name)
+{
+  const string pbasename = getPrometheusName(name);
+  StatsMap entries;
+
+  auto m = broadcastAccFunction<ProxyMappingStats_t>(pleaseGetProxyMappingStats);
```
fredmorcos, line 1167: Same nitpick here regarding `m`.

pdns/rec_channel_rec.cc
```diff
@@ -1864,9 +1908,25 @@ static string setEventTracing(T begin, T end)
...
+    t_proxyMapping = nullptr;
+  }
+  else {
+    // Copy the existing stats values, for the new config items also present in the old
+    auto newmapping = make_unique<ProxyMapping>();
+    for (const auto& [nm, entry] : pm) {
+      auto& newentry = newmapping->insert(nm);
+      newentry.second = entry;
+      if (t_proxyMapping) {
+        if (auto existing = t_proxyMapping->lookup(nm); existing != nullptr) {
```
fredmorcos, line 1923: Clang-tidy complains about this line, `auto existing` can be changed to `const auto *existing`.

pdns/pdns_recursor.cc
```diff
@@ -1838,6 +1841,17 @@ bool checkForCacheHit(bool qnameParsed, unsigned int tag, const string& data,
       }
     }
 
+    // This is only to get the proxyMapping suffixMatch stats right i the case of a PC hit
+    if (t_proxyMapping && source != mappedSource) {
+      if (auto it = t_proxyMapping->lookup(source)) {
```
fredmorcos, line 1846: Same here with `const auto *it`.

fredmorcos (approved):
Just a couple of minor things. I did not test the PR though.

pdns/rec_channel_rec.cc
omoerbeek: This is old code, passing (pointers to) objects between threads. One day we will fix that, but now is not the right time. Remi already made a start with a better approach, but it is not finished yet, see #11427.

PR #11869, opened 2022-08-22
rec: Keep time and count metrics when maintenance is called.

### LINKED ISSUE #6981: rec: worker maintenance duration stats
- Program: Recursor
 - Issue type: Feature request

### Short description

Recursor worker threads perform regular maintenance that includes the Lua `maintenance()` callback and several housekeeping tasks, like clearing caches. This can introduce jitter, but currently no statistic measures the time this maintenance takes.

### Usecase

Measuring the time a round of maintenance takes can give valuable information about jitter introduced due to maintenance.

### Description

I suggest adding the following metrics:

- `maintenance-usec`: total number of microseconds spent performing maintenance across all threads
- `maintenance-count`: total number of times maintenance was run across all threads

These metrics together allow measuring the average time a maintenance cycle took over a time window, giving an estimate for introduced jitter.

I picked microseconds as the unit under the assumption that the counter is an integer. If it is a double, seconds should be used instead.

omoerbeek:
Fixes #6981

If we think it will provide useful numbers, we could split the metrics in a general maintenance and Lua maintenance.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/recursordist/RECURSOR-MIB.txt
+++ b/pdns/recursordist/RECURSOR-MIB.txt
@@ -1066,7 +1066,7 @@ nonResolvingNameserverEntries OBJECT-TYPE
     ::= { stats 126 }
 
 maintenanceUSec OBJECT-TYPE
-    SYNTAX CounterBasedGauge64
+    SYNTAX Counter64
     MAX-ACCESS read-only
     STATUS current
     DESCRIPTION
@@ -1074,7 +1074,7 @@ maintenanceUSec OBJECT-TYPE
     ::= { stats 127 }
 
 maintenanceCount OBJECT-TYPE
-    SYNTAX CounterBasedGauge64
+    SYNTAX Counter64
     MAX-ACCESS read-only
     STATUS current
     DESCRIPTION
--- a/pdns/recursordist/RECURSOR-MIB.txt
+++ b/pdns/recursordist/RECURSOR-MIB.txt
@@ -48,6 +48,9 @@ rec MODULE-IDENTITY
     REVISION "202201310000Z"
     DESCRIPTION "Added non-resolving NS name metric."
 
+    REVISION "202208220000Z"
+    DESCRIPTION "Added internal maintenance metrics."
+
     ::= { powerdns 2 }
 
 powerdns		OBJECT IDENTIFIER ::= { enterprises 43315 }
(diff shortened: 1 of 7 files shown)


pdns/recursordist/RECURSOR-MIB.txt
```diff
@@ -1235,7 +1238,9 @@ recGroup OBJECT-GROUP
         udp6InCsumErrors,
         sourceDisallowedNotify,
         zoneDisallowedNotify,
-        nonResolvingNameserverEntries
+        nonResolvingNameserverEntries,
+        maintenanceUSec,
```
rgacogne, line 1242: The new objects have to be defined first.

rgacogne (changes requested):

pdns/recursordist/RECURSOR-MIB.txt
omoerbeek: Right, tx. Some day we *will* generate all this.

pdns/recursordist/RECURSOR-MIB.txt
```diff
@@ -1062,6 +1065,22 @@ nonResolvingNameserverEntries OBJECT-TYPE
         "Number of entries in the non-resolving NS name cache"
     ::= { stats 126 }
 
+maintenanceUSec OBJECT-TYPE
+    SYNTAX CounterBasedGauge64
```
rgacogne, line 1069:
```suggestion
    SYNTAX Counter64
```

pdns/recursordist/RECURSOR-MIB.txt
```diff
@@ -1062,6 +1065,22 @@ nonResolvingNameserverEntries OBJECT-TYPE
...
+    MAX-ACCESS read-only
+    STATUS current
+    DESCRIPTION
+        "Time spent doing internal maintenance, including Lua maintenance"
+    ::= { stats 127 }
+
+maintenanceCount OBJECT-TYPE
+    SYNTAX CounterBasedGauge64
```
rgacogne, line 1077:
```suggestion
    SYNTAX Counter64
```

rgacogne (approved):
Looks good!

wojas (approved):
Very useful and look good!

The 64 bit part is important for `maintenance-usec`, an overflow would make it useless.

PR #11881, opened 2022-08-24
Reorganize protobuf queueData() with respect to logging

omoerbeek:
Let `queueData()` return a status and log that via a program supplied helper.
This way, the program specific (`recursor`, `dnsdist`) logging isn't polluting the common code.
I'm not 100% happy with this helper, but I lack inspiration to come up with something better atm.

There are a few other cases that need to be dealt with some day.

`dnsdist` log levels should be reviewed (I copied the existing), they might be too verbose.

One observable change in `dnsdist`:
```
"FrameStreamLogger: queue full, dropping."
```
now turned into
```
"dnstap: queue full, dropping."
```
If that is considered a problem, the name of the subsystem should not come the class via a virtual method, but from the caller of the helper.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdist-lua-actions.cc
+++ b/pdns/dnsdist-lua-actions.cc
@@ -1433,6 +1433,26 @@ static DnstapMessage::ProtocolType ProtocolToDNSTap(dnsdist::Protocol protocol)
   throw std::runtime_error("Unhandled protocol for dnstap: " + protocol.toPrettyString());
 }
 
+void remoteLoggerQueueData(RemoteLoggerInterface& r, const std::string& data)
+{
+  auto ret = r.queueData(data);
+
+  switch (ret) {
+  case RemoteLoggerInterface::Result::Queued:
+    break;
+  case RemoteLoggerInterface::Result::PipeFull: {
+    vinfolog("%s: queue full, dropping.", r.name().c_str());
+    break;
+  }
+  case RemoteLoggerInterface::Result::TooLarge: {
+    warnlog("%s: Not sending too large protobuf message", r.name().c_str());
+    break;
+  }
+  case RemoteLoggerInterface::Result::OtherError:
+    warnlog("%s: submitting to queue failed", r.name().c_str());
+  }
+}
+
 class DnstapLogAction : public DNSAction, public boost::noncopyable
 {
 public:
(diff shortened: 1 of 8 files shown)


pdns/dnsdist-lua-actions.cc
```diff
@@ -1433,6 +1433,26 @@ static DnstapMessage::ProtocolType ProtocolToDNSTap(dnsdist::Protocol protocol)
...
 
+void remoteLoggerQueueData(RemoteLoggerInterface& r, const std::string& data)
+{
+  auto ret = r.queueData(data);
+
+  switch (ret) {
+  case RemoteLoggerInterface::Result::Queued:
+    break;
+  case RemoteLoggerInterface::Result::PipeFull: {
+    vinfolog("%s: queue full, dropping.", r.name().c_str());
```
rgacogne, line 1444: The `c_str()` calls should not be needed, unless I'm missing something?

pdns/lwres.cc
```diff
@@ -64,6 +64,37 @@ std::shared_ptr<Logr::Logger> g_slogout;
...
+void remoteLoggerQueueData(RemoteLoggerInterface& r, const std::string& data)
+{
+  auto ret = r.queueData(data);
+
+  switch (ret) {
+  case RemoteLoggerInterface::Result::Queued:
+    break;
+  case RemoteLoggerInterface::Result::PipeFull: {
+    const auto name = r.name();
+    const auto msg = "queue full, dropping";
```
rgacogne, line 76: Perhaps we could have a method returning a `const string&` from a `RemoteLoggerInterface::Result` instead of duplicating these messages (dnsdist and rec)?

pdns/remote_logger.hh
```diff
@@ -118,3 +125,6 @@ private:
   LockGuarded<RuntimeData> d_runtime;
   std::thread d_thread;
 };
+
+// Helper to be defined by main program: queue data and log based on return value of queueData()
+void remoteLoggerQueueData(RemoteLoggerInterface&, const std::string&);
```
rgacogne, line 130: I'm not sure I like this helper being defined here, I would prefer to have it static in dnsdist-lua.cc, for example. It does not matter much, but it might make unit testing that class harder, etc.

rgacogne (approved):
I made a few nits but I like the refactoring very much, nice work!

rgacogne (approved):

Issue #11886, opened 2022-08-29
Feature: Dynamic library backend


darnuria:
- Program: Authoritative
- Issue type: Upstreaming a Feature + design discussion
- Related to: 
   - PR <later today/tomorrow> it's an update on pdns master of #5160
   -  #11863 Some improvement possible to avoid copies
   -  <PR maybe later> Optional Need a way to avoid string copies with `ComboAdress`

Hi, I am working on updating our powerdns setup and we found time to revive upstreaming the work initiated by @baloo (thanks for the commits and the idea) about making a generic dynamic library based backend to powerDNS auth.

We use it daily at Gandi, it's a design production tested. In the PR I will list the methods who may need more design reflection.

I open a issue to address design discussion of the feature.

### Description

Allow pdns to dynamically link to a 3rd party dynamic library compiled in an arbitrary language able to make a C-Foreign Function Interface (like Rust, C, Cpp, Go, Ocaml, etc.) to support more database without the cost of context switch (IE: in the same address space as PowerDNS) and parsing with **remote backend**.

In the rest of the issue I will call this design `DlsoBackend` and call 3rd party dynlib the code dynamically linked.

### Use-case

Use (any) database back-end implemented in an (any) other language and/or without the process context switch and parsing price.

### General design

The main goal of this backend is to dynamically link (`dllopen`) against a 3rd party dynlib to support more databases.

The 3rd party API (calls, structure and callbacks): see [modules/dlsobackend/dlsobackend_api.h](https://github.com/PowerDNS/pdns/pull/5160/files#diff-11cb7e7594d440e102b51e53c595a805d739bd33ca2c37b015c84e49dc9ab1e2)

#### Calling the 3rd party dynlib

DlsoBackend work with a versioned call interface.

Possible calls are hold with a `struct lib_so_api` holding `Nullable` functions pointers, these functions pointers represent the API exposed by the 3rd party dynlib. `NULL` in this context means error in the initialization of the 3rd party dynlib or the function is not supported by the 3rd party dynlib.

It's not mandatory for a 3rd party dynlib, to implement all the API.

#### Use result from the 3rd party dynlib

Like in #5160, the DlsoBackend get back the result with function callbacks `fill_*_cb_t`. These callbacks are provided on function calls and shall be called by the 3rd party dynlib to get back the result.

This callbacks works with opaque data passed in api functions by the DlsoBackend and are expected to write or read data from or for the 3rd party dynlib.

[...]

PR #11887, opened 2022-08-30
Auth: Feature dynamic library backend

darnuria:
- Design/Feature issue: https://github.com/PowerDNS/pdns/issues/11886
- Revival of https://github.com/PowerDNS/pdns/pull/5160

Code heavily used on pdns 4.3.x, updated on master to upstream it.

Citing original @baloo pr because it's a rebase on master + fix to make it compiling/testing again.

I have co-edited some patch to avoid awful commit due to clang-format rules on powerdns code base, that were not enforced at the time of the first-PR.

### Short description

Powerdns will `dlopen` a lib to act as a backend.

## Missing:

- [x]Api versioning. I intend to add an api version at the beginning of the `struct lib_so_api`.
- [x] Packing. A bunch of the structs needs packing.
- [x] Malloc responsibility. `dlso_register_t` should be the one doing the `malloc` of `struct lib_so_api`, not dlso its struct.
- [ ] Documentation
- [ ] Debugability: adds some logging to debug a lib.
- [ ] Maybe rephrasing+rollup my own commits

## Checklist

I have:

- [x] Read the [CONTRIBUTING.md](https://github.com/PowerDNS/pdns/blob/master/CONTRIBUTING.md) document
- [x] `clang-format`ed the code
- [x] Compiled and tested this code: part of the DLSO backend is heavily used at work.
- [ ] Included documentation (including possible behavior changes)
- [ ] Documented the code
- [ ] Added or modified regression test(s) (may have some issue in actual regression test of this PR some may have drifted (investigating))
- [ ] Added or modified unit test(s)

I have not squashed yet 3cb7241cb6f3ea8ab9bb9bae02c5779d543c602c because before methods were not marked `override` and the original PR suffered some patch drift.


darnuria:
Summary: Little bit of noise on this PR due to high pertinant warning with clang-tidy both in test and implementation. Will work on it today and try to fix CI.

darnuria:
CI failing: Some CI fail due to `Unable to load module '../regression-tests/modules/libgsqlite3backend.so': ../regression-tests/modules/libgsqlite3backend.so: cannot open shared object file: No such file or directory`

darnuria:
Did a first pass of clang-tidying in regression test of the backend.

Next task a path in the implementation and header of the backend.

modules/dlsobackend/dlsobackend.cc
```diff
@@ -0,0 +1,722 @@
...
+  rr->qtype = record->qtype;
+  string qname = string(record->qname, record->qname_len);
+  rr->qname = DNSName(qname);
+  rr->qclass = QClass::IN;
+  rr->content = string(record->content, record->content_len);
+  rr->ttl = record->ttl;
+  rr->auth = record->auth;
+  rr->scopeMask = record->scope_mask;
+  rr->domain_id = record->domain_id;
+}
```
darnuria, line 161: I am not sure if this`const` invalidation is  pertinant.

modules/dlsobackend/dlsobackend.cc
```diff
@@ -0,0 +1,722 @@
...
+    in_query = false;
+    return false;
+  }
+
+  return true;
+}
+
+void fill_meta_cb(const void* ptr, uint8_t value_len, const struct dns_value* values)
+{
+  std::vector<std::string>* meta = (std::vector<std::string>*)ptr;
```
darnuria, line 177: Same here.

modules/dlsobackend/dlsobackend.cc
```diff
@@ -0,0 +1,722 @@
...
+    const struct dns_value* value = &values[i];
+    string value_s = string(value->value, value->value_len);
+
+    (*meta).push_back(value_s);
+  }
+}
+
+void fill_metas_cb(const void* ptr, uint8_t meta_len, const struct dns_meta* c_metas)
+{
+  std::map<std::string, std::vector<std::string>>* metas = (std::map<std::string, std::vector<std::string>>*)ptr;
```
darnuria, line 188: Same `const` invalidation is it pertinent?

modules/dlsobackend/dlsobackend.cc
```diff
@@ -0,0 +1,722 @@
...
+
+  string qname = name.toString();
+  bool status = api->set_meta(api->handle, qname.size(), qname.c_str(), kind.size(), kind.c_str(), meta.size(), values);
+  free(values);
+  return status;
+}
+
+void fill_key_cb(const void* ptr, const struct dnskey* dnskey)
+{
+  std::vector<DNSBackend::KeyData>* keys = (std::vector<DNSBackend::KeyData>*)ptr;
```
darnuria, line 243: Also the const thing. the root of the problem is in the API `lib_so_api` that pass these opaque ptr as `const * void` it's an open question.

The intend is to clearly said to 3rd party lib implementer using this backend to avoid at all cost writing by themself to this pr.

modules/dlsobackend/dlsobackend.hh
```diff
@@ -0,0 +1,98 @@
...
+  bool abortTransaction() override;
+
+  void getUnfreshSlaveInfos(vector<DomainInfo>* unfreshDomains) override;
+  void setNotified(uint32_t domain_id, uint32_t serial) override;
+  void setFresh(uint32_t domain_id) override;
+
+  bool replaceRRSet(uint32_t domain_id, const DNSName& qname, const QType& qt, const vector<DNSResourceRecord>& rrset) override;
+  bool feedRecord(const DNSResourceRecord& rr, const DNSName& ordername, bool ordernameIsNSEC3 = false) override;
+  bool feedEnts(int domain_id, map<DNSName, bool>& nonterm) override;
+  bool feedEnts3(int domain_id, const DNSName& domain, map<DNSName, bool>& nonterm, const NSEC3PARAMRecordContent& ns3prc, bool narrow) override;
```
darnuria, line 87: Open ended question: is it voluntary that `domain_id` is not unified to a `uint32_t` across DNSBackend` API?

uint32_t to int is implementation defined got this clang-tidy:
```cpp
/home/axel/gandi/pdns/regression-tests/modules/dlso-gsql/dlsogsql.cc:529:35: warning: narrowing conversion from 'uint32_t' (aka 'unsigned int') to signed type 'int' is implementation-defined [bugprone-narrowing-conversions,cppcoreguidelines-narrowing-conversions]
  return handle->module->feedEnts(domain_id, nonterm);
                                  ^
```

regression-tests/modules/dlso-gsql/dlsogsql.cc
```diff
@@ -0,0 +1,661 @@
...
+
+std::mutex g_configuration_mutex;
+
+extern "C" bool pdns_dlso_register(uint32_t abi_version, struct lib_so_api** ptr, bool dnssec, const char* args)
+{
+  auto* gsql = static_cast<struct dlso_gsql*>(malloc(sizeof(struct dlso_gsql)));
+  if (gsql == nullptr) {
+    return false;
+  }
+  auto* api = static_cast<struct lib_so_api*>(malloc(sizeof(struct lib_so_api)));
```
darnuria, line 562: I am quite rusty (sorry for the joke) with cpp, if there is a better way to write "I want owned memory and manage by myself the releasing" for this backend I would be happy to hear it.

Tried a little bit messing around with `unique_ptr` but wasn't exactly what I wanted since the release will occur on `api->release` call.

Also checking `abi_version` in test is pertinant will fix it tomorrow.

regression-tests/modules/dlso-gsql/dlsogsql.cc
```diff
@@ -0,0 +1,661 @@
...
+  if (api == nullptr) {
+    free(gsql);
+    return false;
+  }
+
+  memset(api, 0, sizeof(*api));
+  *ptr = api;
```
darnuria, line 569: It's a test but since we may fail after we are leaking a pointer to a potentially free-memory here.

modules/dlsobackend/dlsobackend.cc
```diff
@@ -0,0 +1,722 @@
...
+    records[i].scope_mask = rr.scopeMask;
+    records[i].domain_id = rr.domain_id;
+
+    i++;
+  }
+
+  string qname_ = qname.toString();
+
+  try {
+    bool status = api->replace_record(api->handle, domain_id, qname_.size(), qname_.c_str(), qt.getCode(), rrset.size(), records);
```
darnuria, line 608: I suppose using a `vector` ctor+reserve, and give access to the 3rd party lib to ptr with `vector::data()` + size https://en.cppreference.com/w/cpp/container/vector/data may be fine?

modules/dlsobackend/dlsobackend.cc
```diff
@@ -0,0 +1,722 @@
...
+  if (api->set_meta == nullptr)
+    return false;
+
+  struct dns_value* values = (struct dns_value*)calloc(meta.size(), sizeof(struct dns_value));
+  if (values == nullptr)
+    return false;
+  for (uint i = 0; i < meta.size(); i++) {
+    values[i].value_len = meta[i].size();
+    values[i].value = meta[i].c_str();
+  }
```
darnuria, line 233: Same maybe it's a good spot for a vector and `vector::data()`, but same concern with placement (is the data directly constructed on heap).

regression-tests/modules/dlso-gsql/dlsogsql.cc
```diff
@@ -0,0 +1,661 @@
...
+  // declareArguments() the map could do double-free. This is bad and this
+  // shouldn't happen.
+  // Powerdns normally does declare argument and parse configuration
+  // in its very beginning in a single-thread. This is not ArgvMap
+  // responsability to protect from current access. This test backend
+  // being a wrapper Our test backend here may be executed multiple times
+  // by different threads (signing thread, packet receiver)
+  // For this reason, I chose to have a simple mutex, to not allow two
+  // threads to register simultinaeously.
+  std::lock_guard<std::mutex> guard(g_configuration_mutex);
```
rgacogne, line 582: I would suggest wrapping the data you want to protect in a LockGuarded object, see `lock.hh`.

regression-tests/modules/dlso-gsql/dlsogsql.cc
```diff
@@ -0,0 +1,661 @@
...
+
+  // First load the sqlite3 backend, and declare arguments
+  gSQLite3Factory* factory = nullptr;
```
rgacogne, line 585: It might make sense to use a unique_ptr here.

regression-tests/modules/dlso-gsql/dlsogsql.cc
rgacogne: How about wrapping `api` in a `unique_ptr` so you don't need to worry about freeing it in case of error, and if you want to transfer the ownership to `ptr`, and `ptr` cannot be a `unique_ptr` because of your ABI, do `*ptr = api.release()` just before returning?

rgacogne: Same for `sql`, btw.

modules/dlsobackend/dlsobackend.cc
rgacogne: Ideally you would never use a C-style cast in C++, so you would likely need a `reinterpret_cast` and a `const_cast`.

modules/dlsobackend/dlsobackend.cc
rgacogne: Using a `vector`, either passing the final size to the constructor or using `resize()` (`reserve()` does not guarantee anything), would be the right way to go, yes.

modules/dlsobackend/dlsobackend.cc
rgacogne: Please use a `vector` :)

modules/dlsobackend/dlsobackend.cc
rgacogne: I would personally get rid of the `const` qualifier, as I find it misleading because the content is actually modified. Perhaps using opaque types would be better than a `void*`, but I have not reviewed the code to be sure it is possible to do so.

rgacogne (commented):
I added a few comments, but please be aware that we might decide not to merge a new backend, depending on the interest from the community, and the estimated burden to maintain it.
I think it might help if you can commit to maintain it for a reasonable time, for example.

Habbie:
Echoing what Rémi said (we might not merge this), and not having looked at the code, my first question was: uh, but what is this for. Then I remembered #5160 had a few comments explaining this - the problem is not dlopening backends, loading .so, etc. - PowerDNS already does all that! The problem is C++ name mangling. Perhaps the backend could get a better name that clarifies that.

darnuria:
Thanks for your time and the early review. Sorry for the late reply.

About the general feedback about merging or not, the name etc:

From @rgacogne 
> I added a few comments, but please be aware that we might decide not to merge a new backend, depending on the interest from the community, and the estimated burden to maintain it.

From @Habbie 
> Echoing what Rémi said (we might not merge this), 

I totally understands, and people in my team also. It's opensource and free-software rules, at the end of the day it must be maintained, and must have some interest in the community.

> I think it might help if you can commit to maintain it for a reasonable time, for example.

This is something we are considering in my team, cannot state before a team decision.

> and not having looked at the code, my first question was: uh, but what is this for.

Opened an issue for discussing the design (it can change) and the intend and the possibles alternatives: #11886 (may change the name there too).

> Then I remembered #5160 had a few comments explaining this - the problem is not dlopening backends, loading .so, etc. - PowerDNS already does all that! The problem is C++ name mangling. Perhaps the backend could get a better name that clarifies that.

I agree that the current name is maybe misleading, maybe `FFIBackend`, `CFFIBackend`, or `CABIBackend`? I am not great at picking names.

For the rest of the review I will take some time to process it. I might also document it, it's not only for up-streaming it, it already need docs as in.

modules/dlsobackend/dlsobackend.cc
darnuria: The opaque type as in a C definition may be better, may explore it.

For the `const` qualifier get rid of it found it also confusing since we modify the data, may need some doc to express that this pointer need to remain untouched in all the callback-api.

darnuria:
Update: it's quite stalled since I think until "safe closing and ressource release" of powerdns is implemented (5.0? I understand) it's dangerous to open the door to "arbitrary backend" imho.

Would lead to incorrect databases states if D-tor doesn't get called.

Habbie:
Hmm, I'm not actually sure what problem you are referring to - but perhaps I forgot?

darnuria:
Like in case of Signal received by Powerdns historically I had some issue with this backend. But I may have to catch-up with master if it's already fixed! :)

modules/dlsobackend/dlsobackend.cc
```diff
@@ -0,0 +1,821 @@
...
+    return false;
+  }
+
+  return true;
+}
+
+extern "C" void fill_meta_cb(void* ptr, uint8_t value_len, const struct dns_value* values)
+{
+  auto* meta = static_cast<std::vector<std::string>*>(ptr);
+  meta->resize(meta->size() + value_len);
```
AureliaDolo, line 205:
```suggestion
  meta->reserve(meta->size() + value_len);
```

Hello, I've been working on setting things to enable pdns to accept notify host request from pdns control and send corresponding signed notifies and I stumbled onto this. It does not seem like this part was used somewhere else in our setup for a long time. 

With this change, pdns is able to found the needed key and send the notify.

However I'm not fluent in C++, so there may be a better solution.

Thanks !

darnuria:
Hello, I no longer work for Gandi so I cannot commit time to this. Thanks for the review time and the help. :)

pieterlexis:
As there is no work being done on this PR and we have a ticket (#11886), I'm closing this PR.

PR #11889, opened 2022-08-31
dnsdist: Add local ComboAddress parameter for SBind() at TeeAction()

FredericDT:
### Short description
In my prod env, I need to Tee queries from specific ip address, since there are multiple ip addresses on a single interface and the receiver has source address based ACLs and Views.

So I proposed an optional parameter in TeeAction to set the socket bind address.

Usage:
    `addAction(AllRule(), TeeAction("192.0.2.54", false, "192.0.2.53"))`

    In which case, "192.0.2.54" is the ComboAddress of receiver, "192.0.2.53" is the ComboAddress of sender.

### Checklist
I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdist-lua-actions.cc
+++ b/pdns/dnsdist-lua-actions.cc
@@ -135,7 +135,6 @@ public:
 
 private:
   ComboAddress d_remote;
-  boost::optional<ComboAddress> d_local;
   std::thread d_worker;
   void worker();
 
@@ -157,12 +156,12 @@ private:
 };
 
 TeeAction::TeeAction(const ComboAddress& rca, const boost::optional<ComboAddress>& lca, bool addECS) 
-  : d_remote(rca), d_local(lca), d_addECS(addECS)
+  : d_remote(rca), d_addECS(addECS)
 {
   d_fd=SSocket(d_remote.sin4.sin_family, SOCK_DGRAM, 0);
   try {
-    if (d_local) {
-      SBind(d_fd, *d_local);
+    if (lca) {
+      SBind(d_fd, *lca);
     }
     SConnect(d_fd, d_remote);
     setNonBlocking(d_fd);
--- a/pdns/dnsdist-lua-actions.cc
+++ b/pdns/dnsdist-lua-actions.cc
@@ -127,7 +127,7 @@ class TeeAction : public DNSAction
 {
 public:
   // this action does not stop the processing
-  TeeAction(const ComboAddress& ca, bool addECS=false);
+  TeeAction(const ComboAddress& rca, const ComboAddress& lca, bool setLocalBindAddress=false, bool addECS=false);
   ~TeeAction() override;
   DNSAction::Action operator()(DNSQuestion* dq, std::string* ruleresult) const override;
   std::string toString() const override;
@@ -135,6 +135,7 @@ public:
 
 private:
   ComboAddress d_remote;
+  ComboAddress d_local;
   std::thread d_worker;
   void worker();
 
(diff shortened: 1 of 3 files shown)


omoerbeek:
I think there is need to have an extra boolean flag, a `boost::optional<X>` already has that flag in itself.

pdns/dnsdistdist/docs/rules-actions.rst
```diff
@@ -1658,10 +1658,11 @@ The following actions exist.
   Before 1.7.0 this action was performed even when the query had been received over TCP, which required the use of :func:`TCPRule` to
   prevent the TC bit from being set over TCP transports.
 
-.. function:: TeeAction(remote[, addECS])
+.. function:: TeeAction(remote[, addECS[, local]])
```
rgacogne, line 1662:
```suggestion
.. versionchanged:: 1.8.0
  Added the optional parameter ``local``.
```

pdns/dnsdist-lua-actions.cc
```diff
@@ -2416,8 +2422,8 @@ void setupLuaActions(LuaContext& luaCtx)
     });
 #endif /* DISABLE_PROTOBUF */
 
-  luaCtx.writeFunction("TeeAction", [](const std::string& remote, boost::optional<bool> addECS) {
-      return std::shared_ptr<DNSAction>(new TeeAction(ComboAddress(remote, 53), addECS ? *addECS : false));
+  luaCtx.writeFunction("TeeAction", [](const std::string& remote, boost::optional<bool> addECS, boost::optional<std::string> local) {
+      return std::shared_ptr<DNSAction>(new TeeAction(ComboAddress(remote, 53), ComboAddress(local ? *local : "0.0.0.0", 0), local ? true : false, addECS ? *addECS : false));
```
rgacogne, line 2426: I agree with Otto, we could pass a `boost::optional<ComboAddress>` to the TeeAction constructor, something like:
```c++
boost::optional<ComboAddress> localAddr{boost::none};
if (local) {
  localAddr = ComboAddress(*local);
}
return std::shared_ptr<DNSAction>(new TeeAction(ComboAddress(remote, 53), localAddr, addECS ? *addECS : false));
```

pdns/dnsdist-lua-actions.cc
```diff
@@ -127,14 +127,15 @@ class TeeAction : public DNSAction
...
-  TeeAction(const ComboAddress& ca, bool addECS=false);
+  TeeAction(const ComboAddress& rca, const ComboAddress& lca, bool setLocalBindAddress=false, bool addECS=false);
   ~TeeAction() override;
   DNSAction::Action operator()(DNSQuestion* dq, std::string* ruleresult) const override;
   std::string toString() const override;
   std::map<std::string, double> getStats() const override;
 
 private:
   ComboAddress d_remote;
+  ComboAddress d_local;
```
rgacogne, line 138: Since we only use the local address in the constructor, I don't think we need to keep it around?

FredericDT:
> I think there is need to have an extra boolean flag, a `boost::optional<X>` already has that flag in itself.

Sounds good,

I think there are two options:
1. Pass that `boost::optional<std:string>` to `TeeAction` class, build `ComboAddress` in the constructor.
2. Pass `boost::optional<ComboAddress>` as parameter `local`.

rgacogne (changes requested):
Thanks for that pull request! I have made a few suggestions but nothing big, the code looks good.

pdns/dnsdist-lua-actions.cc
FredericDT: Possibly using that `d_local` in statistical function?

pdns/dnsdist-lua-actions.cc
FredericDT: Agree, I will make a commit.

pdns/dnsdistdist/docs/rules-actions.rst
FredericDT: Commit in https://github.com/PowerDNS/pdns/pull/11889/commits/76fd1dd4505ff168c621d6543f27b9c54bb25222

pdns/dnsdist-lua-actions.cc
FredericDT: Commit in https://github.com/PowerDNS/pdns/pull/11889/commits/8dbb2c9b9da9d4cb1864158f79f59cd2f50efa89

pdns/dnsdist-lua-actions.cc
rgacogne: I would prefer not keeping it for now, we can always add it back later when we actually decide to do something with it :)

FredericDT: OK, I will make a commit

FredericDT: Commit in https://github.com/PowerDNS/pdns/pull/11889/commits/f7bf9ae833f4de1be3a6fe9374f13880678e9ffb

rgacogne (approved):
Looks good, thanks!

PR #11890, opened 2022-08-31
rec: Failure to retrieve DNSKEYs of an Insecure zone should not be fatal.

omoerbeek:
_As with any validation related PR, this should be reviewed with extra care!_

This issue happens if a record set is signed even though the zone
itself is Insecure. Syncres then tries to retrieve DNSKEYs and a
timeout on that would lead to an ImmediateServFailException.

So do not throw an exception.  Further validation with the potentially
empty set of DNSKEYs should do the right thing. Including tests for
both Secure and Insecure zones.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [X] documented the code

### DIFF
--- a/pdns/recursordist/test-syncres_cc5.cc
+++ b/pdns/recursordist/test-syncres_cc5.cc
@@ -2179,12 +2179,12 @@ BOOST_AUTO_TEST_CASE(test_dnssec_secure_servfail_ds)
   }
 }
 
-BOOST_AUTO_TEST_CASE(test_dnssec_secure_servfail_dnskey)
+static void dnssec_secure_servfail_dnskey(DNSSECMode mode, vState expectedValidationResult)
 {
   std::unique_ptr<SyncRes> sr;
   initSR(sr, true);
 
-  setDNSSECValidation(sr, DNSSECMode::ValidateAll);
+  setDNSSECValidation(sr, mode);
 
   primeHints();
   const DNSName target("powerdns.com.");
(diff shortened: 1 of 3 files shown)


rgacogne:
This might cause a regression to the behaviour we had before https://github.com/PowerDNS/pdns/pull/9292, though :-/

omoerbeek:
> 



> This might cause a regression to the behaviour we had before #9292, though :-/

yes, that might be an issue. Will revisit soon.

omoerbeek:
With this change I see the failed DNSKEY retrieval leading to:
```
Sep 05 08:55:23 [1]  QM incometax.gov.in.|DNSKEY child=incometax.gov.in: Step3 Final resolve: Server Failure/0
Sep 05 08:55:23 [1] Returning Bogus state from getDNSKeys(incometax.gov.in)
Sep 05 08:55:23 [1] checking whether we missed a zone cut for incometax.gov.in before returning a Bogus state for www.incometax.gov.in|CNAME
Sep 05 08:55:23 [1] : no or invalid signature/proof for incometax.gov.in, we likely missed a cut between gov.in and incometax.gov.in, looking for it
Sep 05 08:55:23 [1] : - Looking for a DS at incometax.gov.in
Sep 05 08:55:23 [1] : no TA found for 'incometax.gov.in' among 1
Sep 05 08:55:23 [1] : no TA found for 'incometax.gov.in' among 1
Sep 05 08:55:23 [1] : no TA found for 'gov.in' among 1
Sep 05 08:55:23 [1] : no TA found for 'in' among 1
Sep 05 08:55:23 [1] : got TA for '.'
Sep 05 08:55:23 [1]  QM incometax.gov.in.|DS child=(empty): doResolve
Sep 05 08:55:23 [1]  incometax.gov.in: Wants DNSSEC processing, auth data in query for DS
Sep 05 08:55:23 [1]  incometax.gov.in: Recursion not requested for 'incometax.gov.in|DS', peeking at auth/forward zones
Sep 05 08:55:23 [1]  incometax.gov.in: DS is negatively cached via 'in' for another 1797 seconds
Sep 05 08:55:23 [1]  incometax.gov.in: updating validation state with negative cache content for incometax.gov.in to Insecure
Sep 05 08:55:23 [1]  QM incometax.gov.in.|DS child=(empty): Step0 Found in cache
Sep 05 08:55:23 [1] : - Found cut at incometax.gov.in
Sep 05 08:55:23 [1] : New state for incometax.gov.in is Insecure
Sep 05 08:55:23 [1] Validation result is Insecure, current state is Secure
Sep 05 08:55:23 [1] validation state was Secure, state update is Insecure, validation state is now Insecure
```
So the `Bogus` state is "corrected" by the parent saying the zone is `Insecure`. The cache also does not have any `Bogus` results. So that's good.

The new unit tests (with an `Insecure` zone) also do not leave RRSets in `Bogus` state in the cache.
But the (slighly modified) existing unit tests (with a `Secure` zone) _do leave_ records marked `Bogus`.

In a way that is correct, as we where not able to verify the RRSet since the DNSKEY wasn't available and the parent said the zone was `Secure`. On the other hand, in master the records would not be entered in the cache and  new request would retry to get the DNSKEYs, potentially succeeding (not in this particular case as the auth insists on net replying over TCP).

So the question boils down to: do we want to have this (potentially transient) validation failure cached? It might improve performance since new request will be based on the cache instead of doing the DNSKEY retrieval again. OTOH a transient network failure could cause a lot of havoc.

All in all, I think approach indeed has too many risk in caching transient network induced validation errors. Thinking if there is another viable alternative approach.

rgacogne:
> So the question boils down to: do we want to have this (potentially transient) validation failure cached? It might improve performance since new request will be based on the cache instead of doing the DNSKEY retrieval again. OTOH a transient network failure could cause a lot of havoc.

I would be completely fine with that failure being cached as a Server Failure, for a limited amount of time. Caching it as a validation failure seems wrong to me, though :-/

omoerbeek:
Alternative approach: introduce a separate `Bogus` state (`BogusServFailOnGetDNSKEYs`), used for this `ServFail` case. Make sure this validation state does not end up being cached in the record cache. I am a bit worried though that `Bogus` states for other records derived from this special `Bogus` state could still end up being cached. In unit test case this does not happen (the record previous marked `Bogus` stays `Indeterminate`), but it's hard for me to see the full consequences of this.

rgacogne:
I wonder if we could deal with the new state in `SyncRes::validateRecordsWithSigs` instead. `getDNSKeys()` would keep returning a `Bogus` state, we would still be trying to find a missed cut, but if we cannot find one and are still in a Secure zone with a `BogusServFailOnGetDNSKEYs` state we would then, and only then, throw an `ImmediateServFailException`. Do you think it would work?

omoerbeek:
That sound like  plan indeed.

pdns/validate.cc
```diff
@@ -1425,7 +1425,7 @@ DNSName getSigner(const std::vector<std::shared_ptr<RRSIGRecordContent> >& signa
...
 {
-  static const std::vector<std::string> vStates = {"Indeterminate", "Insecure", "Secure", "NTA", "TA", "Bogus - No valid DNSKEY", "Bogus - Invalid denial", "Bogus - Unable to get DSs", "Bogus - Unable to get DNSKEYs", "Bogus - Self Signed DS", "Bogus - No RRSIG", "Bogus - No valid RRSIG", "Bogus - Missing negative indication", "Bogus - Signature not yet valid", "Bogus - Signature expired", "Bogus - Unsupported DNSKEY algorithm", "Bogus - Unsupported DS digest type", "Bogus - No zone key bit set", "Bogus - Revoked DNSKEY", "Bogus - Invalid DNSKEY Protocol" };
+  static const std::vector<std::string> vStates = {"Indeterminate", "Insecure", "Secure", "NTA", "TA", "Bogus - No valid DNSKEY", "Bogus - Invalid denial", "Bogus - Unable to get DSs", "Bogus - Unable to get DNSKEYs", "Bogus - Self Signed DS", "Bogus - No RRSIG", "Bogus - No valid RRSIG", "Bogus - Missing negative indication", "Bogus - Signature not yet valid", "Bogus - Signature expired", "Bogus - Unsupported DNSKEY algorithm", "Bogus - Unsupported DS digest type", "Bogus - No zone key bit set", "Bogus - Revoked DNSKEY", "Bogus - Invalid DNSKEY Protocol", "Bogus - ServFail on geting DNSKEYs" };
```
rgacogne, line 1428:
```suggestion
  static const std::vector<std::string> vStates = {"Indeterminate", "Insecure", "Secure", "NTA", "TA", "Bogus - No valid DNSKEY", "Bogus - Invalid denial", "Bogus - Unable to get DSs", "Bogus - Unable to get DNSKEYs", "Bogus - Self Signed DS", "Bogus - No RRSIG", "Bogus - No valid RRSIG", "Bogus - Missing negative indication", "Bogus - Signature not yet valid", "Bogus - Signature expired", "Bogus - Unsupported DNSKEY algorithm", "Bogus - Unsupported DS digest type", "Bogus - No zone key bit set", "Bogus - Revoked DNSKEY", "Bogus - Invalid DNSKEY Protocol", "Bogus - ServFail on getting DNSKEYs" };
```

omoerbeek: I was pondering if maybe we could skip introducing an extra `vState`, and have `getDNSKeys` set a flag instead which says: `ServFail` while getting keys.

rgacogne: That would make sense to me, yes :+1:

pdns/recursordist/test-syncres_cc5.cc
```diff
@@ -2289,4 +2289,121 @@ BOOST_AUTO_TEST_CASE(test_dnssec_secure_servfail_dnskey)
...
+  std::unique_ptr<SyncRes> sr;
+  initSR(sr, true);
+
+  setDNSSECValidation(sr, mode);
+
+  primeHints();
+  const DNSName target("powerdns.com.");
+  const ComboAddress targetAddr("192.0.2.42");
+
+  // We use two sets of keys, as powerdns.com is Inscure according to parent but returns signed results,
```
rgacogne, line 2311:
```suggestion
  // We use two sets of keys, as powerdns.com is Insecure according to parent but returns signed results,
```

rgacogne (approved):
One nit but looks good to me otherwise!

omoerbeek:
Comment processed and squashed.

rgacogne (approved):

z00z00z00:
Is there a link between this issue and the Non-Responsive Delegation Attack affecting BIND and Unbound ?

omoerbeek:
No.

PR #11891, opened 2022-08-31
builder: add --no-install-recommends to apt-get install

Habbie:
### Short description
This was triggered by Ubuntu Kinetic pulling in a version of 
systemd-resolved that breaks inside Docker. systemd-resolved is an indirect
(via Recommends somewhere) dependency of devscripts, which we need.

However, if we were relying on Recommends, that was a bug, so I'm applying
the flag to all distributions.

### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code


jsoref (approved):
Looks good to me

PR #11903, opened 2022-09-06
rec: protobuf and dnstap metrics

### LINKED ISSUE #11841: Feature request: counters/telemetry on telemetry
 - Issue type: Feature request

### Short description
Create logging metrics that show status and counters on telemetry consumers

### Usecase
We send a large number of dnstap messages to nearby consumers. We don't know when those consumers are overwhelmed, or if they are overwhelmed, how many messages we've lost. Also, sometimes we see flapping behaviors due to restarts or faults on the consumer side. It would be useful to know what is happening on logging consumers from the perspective of the logging origin, and this would apply to all packages: rec, auth, and dnsdist (though we are most interested in dnsdist)
  If there was some way to subtract successful messages from possible transmitted messages, then we could see when we are hitting faults if the socket was open and not able to consume fast enough.  If we knew the status of the socket, then we could tell the difference between times when the that delta was increasing due to simple network failure versus if the delta was increasing due to inability to consume the data fast enough.

### Description
Counters that collect and make available the following information on each set of IP:port consumers (maybe make those tags in the Prometheus format of the output in /metrics ?)

  - number of messages sent to each telemetry consumer. This counter would increment even if the consumer was unavailable, not connected, not able to ACK previous messages fast enough, etc. - it is a counter of "possible" messages that the consumer could have received.
  -  number of messages actually transmitted to the consumer. This would count how many messages were successfully (at least, from the TCP definition of "success") transmitted to the consumer. If the consumer is not fast enough to accept messages, and messages are dropped, those dropped messages would not increment this counter.
  -  number of times the telemetry socket had been opened.  Every time the session achieves a successful three-way handshake, this would increment.
  -  current status of TCP session - 0 for not fully established, 1 for fully established (I don't think it's necessary to track half-opened sockets here, as that seems to be extra work that is not really helpful)
 
https://drive.google.com/file/d/1YgplffRgtaiksqi-Gg2Gxw0gTR9XkDGs/view?usp=sharing

omoerbeek:
### Short description

Fixes #11841 (at least most of it).

Includes a change to the remote logger counting, as they are always accessed when holding a lock, so they do not need to be atomic.

Example output of `rec_control get-remotelogger-stats`
```
Queued	Pipe-	Too-	Other-	Address	Type
	Full	Large	error
2	0	0	0	127.0.0.1:9997	protobuf
42	0	0	0	127.0.0.1:9996	outgoingProtobuf
42	0	0	0	/tmp/dnstap.sock	dnstapFrameStream
```
example Prometheus output:
```
# TYPE pdns_recursor_remote_logger counter
# HELP pdns_recursor_remote_logger Number of remote logging events
pdns_recursor_remote_logger{address="127.0.0.1:9997",type="protobuf",count="otherError"} 0
pdns_recursor_remote_logger{address="127.0.0.1:9996",type="outgoingProtobuf",count="otherError"} 0
pdns_recursor_remote_logger{address="/tmp/dnstap.sock",type="dnstapFrameStream",count="otherError"} 0
pdns_recursor_remote_logger{address="127.0.0.1:9997",type="protobuf",count="pipeFull"} 0
pdns_recursor_remote_logger{address="127.0.0.1:9996",type="outgoingProtobuf",count="pipeFull"} 0
pdns_recursor_remote_logger{address="/tmp/dnstap.sock",type="dnstapFrameStream",count="pipeFull"} 0
pdns_recursor_remote_logger{address="127.0.0.1:9997",type="protobuf",count="queued"} 2
pdns_recursor_remote_logger{address="127.0.0.1:9996",type="outgoingProtobuf",count="queued"} 42
pdns_recursor_remote_logger{address="/tmp/dnstap.sock",type="dnstapFrameStream",count="queued"} 42
pdns_recursor_remote_logger{address="127.0.0.1:9997",type="protobuf",count="tooLarge"} 0
pdns_recursor_remote_logger{address="127.0.0.1:9996",type="outgoingProtobuf",count="tooLarge"} 0
pdns_recursor_remote_logger{address="/tmp/dnstap.sock",type="dnstapFrameStream",count="tooLarge"} 0
```

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code


omoerbeek:
Rebased to fix conflicts

wojas:
Nit: perhaps name it `pdns_recursor_remote_logger_count`. It more clearly describes the metric and makes it easier to add additional `pdns_recursor_remote_logger_*` metrics (e.g. histograms) in the future without name conflicts.

omoerbeek:
> Nit: perhaps name it `pdns_recursor_remote_logger_count`. It more clearly describes the metric and makes it easier to add additional `pdns_recursor_remote_logger_*` metrics (e.g. histograms) in the future without name conflicts.

Good point. Will do.

pdns/remote_logger.hh
```diff
@@ -113,17 +146,16 @@ private:
...
   ComboAddress d_remote;
-  std::atomic<uint64_t> d_drops{0};
-  std::atomic<uint64_t> d_processed{0};
   uint16_t d_timeout;
   uint8_t d_reconnectWaitTime;
   std::atomic<bool> d_exiting{false};
   bool d_asyncConnect{false};
 
-  LockGuarded<RuntimeData> d_runtime;
+  mutable LockGuarded<RuntimeData> d_runtime;
```
rgacogne, line 158: I'm worried about making this mutable, this is fine now since we are only using that to acquire the lock to read metrics, but it means we could later alter something more critical in a `const` method and not notice. I wonder how painful it would be to make `getStats()` and `toString()` not `const` instead..

rgacogne (approved):
Looks good. I'm not very happy about making the whole runtime data mutable, but perhaps it's the least bad option.

pdns/remote_logger.hh
omoerbeek: Fixed as suggested.

pdns/fstrm_logger.hh
```diff
@@ -39,15 +39,30 @@ public:
   FrameStreamLogger(int family, const std::string& address, bool connect, const std::unordered_map<string,unsigned>& options = std::unordered_map<string,unsigned>());
   ~FrameStreamLogger();
   [[nodiscard]] RemoteLoggerInterface::Result queueData(const std::string& data) override;
+
+  std::string address() const override
```
fredmorcos, line 43: `[[nodiscard]]`

pdns/fstrm_logger.hh
```diff
@@ -39,15 +39,30 @@ public:
...
+  {
+    return d_address;
+  }
+
   const std::string name() const override
```
fredmorcos, line 48: Unrelated to this PR, but if you add the `[[nodiscard]]` above then also add it here, and remove the `const` (I guess they're going to have to change in the base class as well).

pdns/rec_channel_rec.cc
```diff
@@ -939,6 +939,18 @@ static ProxyMappingStats_t* pleaseGetProxyMappingStats()
   return ret;
 }
 
+static RemoteLoggerStats_t* pleaseGetRemoteLoggerStats()
+{
+  auto ret = make_unique<RemoteLoggerStats_t>();
+
+  if (t_protobufServers) {
+    for (const auto& s : *t_protobufServers) {
```
fredmorcos, line 947: clang-tidy complains about a too short name for `s`. Perhaps `server` would be better here.

pdns/rec_channel_rec.cc
```diff
@@ -950,6 +962,58 @@ static string doGetProxyMappingStats()
   return ret.str();
 }
 
+static RemoteLoggerStats_t* pleaseGetOutgoingRemoteLoggerStats()
+{
+  auto ret = make_unique<RemoteLoggerStats_t>();
+
+  if (t_outgoingProtobufServers) {
+    for (const auto& s : *t_outgoingProtobufServers) {
```
fredmorcos, line 970: Same here.

pdns/rec_channel_rec.cc
```diff
@@ -950,6 +962,58 @@ static string doGetProxyMappingStats()
...
+  return ret.release();
+}
+
+#ifdef HAVE_FSTRM
+static RemoteLoggerStats_t* pleaseGetFramestreamLoggerStats()
+{
+  auto ret = make_unique<RemoteLoggerStats_t>();
+
+  if (t_frameStreamServersInfo.servers) {
+    for (const auto& s : *t_frameStreamServersInfo.servers) {
```
fredmorcos, line 983: And here.

pdns/rec_channel_rec.cc
```diff
@@ -950,6 +962,58 @@ static string doGetProxyMappingStats()
...
+      ret->emplace(std::make_pair(s->address(), s->getStats()));
+    }
+  }
+  return ret.release();
+}
+#endif
+
+static void remoteLoggerStats(const string& type, const RemoteLoggerStats_t& stats, ostringstream& os)
```
fredmorcos, line 991: Too short name for `os` here, perhaps `outStringStream` would be better.

pdns/rec_channel_rec.cc
```diff
@@ -950,6 +962,58 @@ static string doGetProxyMappingStats()
...
+    return;
+  }
+  for (const auto& [key, entry] : stats) {
+    os << entry.d_queued << '\t' << entry.d_pipeFull << '\t' << entry.d_tooLarge << '\t' << entry.d_otherError << '\t' << key << '\t' << type << endl;
+  }
+}
+
+static string getRemoteLoggerStats()
+{
+  ostringstream os;
```
fredmorcos, line 1003: Same here.

pdns/rec_channel_rec.cc
```diff
@@ -1195,6 +1259,36 @@ static StatsMap toProxyMappingStatsMap(const string& name)
...
+  const auto pbasename = getPrometheusName(name);
+  StatsMap entries;
+
+  auto stats1 = broadcastAccFunction<RemoteLoggerStats_t>(pleaseGetRemoteLoggerStats);
+  auto stats2 = broadcastAccFunction<RemoteLoggerStats_t>(pleaseGetOutgoingRemoteLoggerStats);
+  auto stats3 = broadcastAccFunction<RemoteLoggerStats_t>(pleaseGetFramestreamLoggerStats);
+  uint64_t count = 0;
+  for (const auto& [stats, type] : {make_pair(stats1, "protobuf"), make_pair(stats2, "outgoingProtobuf"), make_pair(stats3, "dnstapFrameStream")}) {
+    for (const auto& [key, entry] : stats) {
+      auto keyname = pbasename + "{address=\"" + key + "\",type=\"" + type + "\",count=\"";
```
fredmorcos, line 1273: clang-tidy says `+` is inefficient and might incur unnecessary allocations. If it's important here, use `+=` or `append()` instead.

pdns/recursordist/rec-main.cc
```diff
@@ -1294,6 +1294,14 @@ static ProxyMappingStats_t& operator+=(ProxyMappingStats_t& a, const ProxyMappin
   return a;
 }
 
+static RemoteLoggerStats_t& operator+=(RemoteLoggerStats_t& a, const RemoteLoggerStats_t& b)
+{
+  for (const auto& [key, entry] : b) {
```
fredmorcos, line 1299: I did not know this kind of pattern matching worked! 🥳

rgacogne (approved):

pdns/remote_logger.hh
```diff
@@ -91,15 +112,27 @@ public:
...
     return "protobuf";
   }
-  std::string toString() const override
+  std::string toString() override
+  {
+    auto runtime = d_runtime.lock();
+    return d_remote.toStringWithPort() + " (" + std::to_string(runtime->d_stats.d_queued) + " processed, " + std::to_string(runtime->d_stats.d_pipeFull + runtime->d_stats.d_tooLarge + runtime->d_stats.d_otherError) + " dropped)";
+  }
+
+  virtual RemoteLoggerInterface::Stats getStats() override
```
fredmorcos, line 131: `virtual` isn't needed here because it's already `override`.

fredmorcos (approved):
Just small nitpicks and clang-tidy stuff.

pdns/recursordist/rec-main.cc
omoerbeek: A lot more is possible, see https://en.cppreference.com/w/cpp/language/structured_binding

PR #11904, opened 2022-09-06
rec: For zones having many NS records take a sample

omoerbeek:
### Short description

so tak…
### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/recursordist/docs/settings.rst
+++ b/pdns/recursordist/docs/settings.rst
@@ -1224,6 +1224,19 @@ it by the number of NS records found above the
 `max-ns-address-qperq`_ value. The limit wil not be reduced to a
 number lower than 5.
 
+.. _setting-max-ns-per-resolve:
+
+``max-ns-per-resolve``
+----------------------
+.. versionadded:: 4.8.0
+
+-  Integer
+-  Default: 13
+
+The maximum number of NS records that will be considered to select a nameserver to contact to resolve a name.
+If a zone has more than `max-ns-per-resolve`_ NS records, a random sample of this size will be used.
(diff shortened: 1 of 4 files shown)


pdns/recursordist/docs/settings.rst
```diff
@@ -1224,6 +1224,19 @@ it by the number of NS records found above the
 `max-ns-address-qperq`_ value. The limit wil not be reduced to a
 number lower than 5.
 
+.. _setting-max-nsperq:
```
rgacogne, line 1227: A small nit, I find the `perq` suffix confusing for this setting, as it suggests (to me, at least) that it is the number of NS we are willing to consider per query, while it's actually the number of NS we are considering per zone?

rgacogne (approved):
Code looks good, I have a small dislike for the setting/internal variable naming but I could live with it.

pdns/recursordist/docs/settings.rst
omoerbeek: It's wrong indeed. It's per name, I'd say. Once client query might need to resolve several names, each such resolve is subject to the restriction, while max-qperq is _per client query_.

Would the name `max-ns-per-resolve` suit you better? Cannot think of a better name atm.

rgacogne: Yes, it would be much better. Thanks!

.github/actions/spell-check/expect.txt
```diff
@@ -1130,6 +1130,7 @@ nsid
 nsis
 nsl
 nsname
+nsperq
```
rgacogne, line 1133: This might need an update?

PR #11906, opened 2022-09-06
rec: Implement padding of (DoT) messages to auth

omoerbeek:
### Short description

See https://www.rfc-editor.org/rfc/rfc8467 and https://www.rfc-editor.org/rfc/rfc7830

A few questions remaining:

1. Do we want to make this configurable? Currently this PR pads outgoing messages if and only if we're doing DoT. It could be we expect our outgoing traffic to be encapsulated in an encrypted channel. In that case padding could also be an option for un-encrypted outgoing traffic, but I find that a bit far fetched and not something you would do typically. 
3. (Related to 1.) Does it make sense to not want padding for DoT?
4. This PR uses a padding size of 128, as we are (kind of) client. Is that OK?

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/lwres.cc
+++ b/pdns/lwres.cc
@@ -57,6 +57,7 @@
 
 thread_local TCPOutConnectionManager t_tcp_manager;
 std::shared_ptr<Logr::Logger> g_slogout;
+bool g_paddingOutgoing;
 
 void remoteLoggerQueueData(RemoteLoggerInterface& r, const std::string& data)
 {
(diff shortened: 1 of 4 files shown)


rgacogne (approved):
Great!

> Do we want to make this configurable?

I think it would be good to have a boolean setting to enable/disable padding for outgoing DoT queries, yes. That might be useful as a work-around if it turns out that padding queries breaks some authoritative servers (given that they support DoT I would be surprised, but DNS tends to do that).
I don't think we need to support padding for non-DoT traffic, no.

> This PR uses a padding size of 128, as we are (kind of) client. Is that OK?

That seems good to me, yes.

rgacogne (approved):
:+1:

PR #11907, opened 2022-09-06
Recursor: Add `--config[=check|=diff|=default]`

fredmorcos:
### Short description
Add `--config[=check|=diff|=default]` to the recursor

### Checklist
I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/recursordist/docs/manpages/pdns_recursor.1.rst
+++ b/pdns/recursordist/docs/manpages/pdns_recursor.1.rst
@@ -46,6 +46,11 @@ at `<https://doc.powerdns.com/>`
     chroot the process to *directory*.
 --client-tcp-timeout=<num>
     Timeout in seconds when talking to TCP clients.
+--config
+    Show the current configuration. There are three optional values:
+    ``--config=default`` to show the default configuration.
+    ``--config=diff``    show modified options in the current configuration.
+    ``--config=check``   to check the current configuration for errors.
 --config-dir=<directory>
     Location of configuration directory (recursor.conf), the default
     depends on the SYSCONFDIR option at build-time, which is usually
--- a/pdns/recursordist/rec-main.cc
+++ b/pdns/recursordist/rec-main.cc
@@ -1157,37 +1157,43 @@ void parseACLs()
   static bool l_initialized;
 
   if (l_initialized) { // only reload configuration file on second call
-    string configname = ::arg()["config-dir"] + "/recursor.conf";
-    if (::arg()["config-name"] != "") {
-      configname = ::arg()["config-dir"] + "/recursor-" + ::arg()["config-name"] + ".conf";
+    string configName = ::arg()["config-dir"] + "/recursor.conf";
+    if (!::arg()["config-name"].empty()) {
+      configName = ::arg()["config-dir"] + "/recursor-" + ::arg()["config-name"] + ".conf";
     }
-    cleanSlashes(configname);
+    cleanSlashes(configName);
 
-    if (!::arg().preParseFile(configname.c_str(), "allow-from-file"))
-      throw runtime_error("Unable to re-parse configuration file '" + configname + "'");
-    ::arg().preParseFile(configname.c_str(), "allow-from", LOCAL_NETS);
+    if (!::arg().preParseFile(configName.c_str(), "allow-from-file")) {
+      throw runtime_error("Unable to re-parse configuration file '" + configName + "'");
+    }
+    ::arg().preParseFile(configName.c_str(), "allow-from", LOCAL_NETS);
 
-    if (!::arg().preParseFile(configname.c_str(), "allow-notify-from-file"))
-      throw runtime_error("Unable to re-parse configuration file '" + configname + "'");
-    ::arg().preParseFile(configname.c_str(), "allow-notify-from");
+    if (!::arg().preParseFile(configName.c_str(), "allow-notify-from-file")) {
+      throw runtime_error("Unable to re-parse configuration file '" + configName + "'");
+    }
+    ::arg().preParseFile(configName.c_str(), "allow-notify-from");
 
-    ::arg().preParseFile(configname.c_str(), "include-dir");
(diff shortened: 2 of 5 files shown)


pdns/recursordist/rec-main.cc
```diff
@@ -2893,6 +2894,35 @@ int main(int argc, char** argv)
...
+          exit(0);
+        }
+        catch (const ArgException& argException) {
+          SLOG(g_log << Logger::Warning << "Unable to parse configuration file '" << configname << "'" << endl,
+               startupLog->error("Cannot parse configuration", "Unable to parse configuration file", "config_file", Logging::Loggable(configname)));
+          exit(1);
+        }
+      }
+      else {
+        cout << ::arg().configstring(true, true);
```
rgacogne, line 2921: Is there an actual difference between `--config` and `--config=default` at this point?

fredmorcos: Right... Probably should add a call to `::arg().laxFile(...)` before the block to have `--config` dump a loaded configuration file (with warnings about unknown settings). I'm not sure if it's the right way to do it though...

pdns/recursordist/rec-main.cc
```diff
@@ -2893,6 +2894,35 @@ int main(int argc, char** argv)
...
+                 startupLog->error("No such file", "Unable to open configuration file", "config_file", Logging::Loggable(configname)));
+            exit(1);
+          }
+
+          ::arg().parse(argc, argv);
+          exit(0);
+        }
+        catch (const ArgException& argException) {
+          SLOG(g_log << Logger::Warning << "Unable to parse configuration file '" << configname << "'" << endl,
+               startupLog->error("Cannot parse configuration", "Unable to parse configuration file", "config_file", Logging::Loggable(configname)));
```
omoerbeek, line 2916: With a wrong config file I now see:
```
Sep 07 09:33:17 msg="Unable to parse configuration file" error="Cannot parse configuration" subsystem="config" level="0" tid="0" ts="1662535997.193" config_file="./recursor.conf"
```
The ectual underlying `argException.reason` is not included in the error message.

pdns/recursordist/rec-main.cc
omoerbeek: Did you take a look how auth does it?

fredmorcos:
When approving, please don't merge this just yet, I need to run a fixup of the commits.

pdns/recursordist/rec-main.cc
```diff
@@ -2893,9 +2894,43 @@ int main(int argc, char** argv)
...
+          ::arg().parse(argc, argv);
+          exit(0);
+        }
+        catch (const ArgException& argException) {
+          SLOG(g_log << Logger::Warning << "Unable to parse configuration file '" << configname << "': " << argException.reason << endl,
+               startupLog->error("Cannot parse configuration", "Unable to parse configuration file", "config_file", Logging::Loggable(configname), "reason", Logging::Loggable(argException.reason)));
+          exit(1);
+        }
+      }
+      else {
```
omoerbeek, line 2919: This is for the `--confg=xyz` (where `xyz` may be empty) case. The man page change does not describe this.
While looking how auth describes this I see it also has `diff`, which would be really nice to have in rec as well 

So can you take the man pages lines from auth and see if `diff` can be added as well?

omoerbeek (changes requested):

pdns/recursordist/rec-main.cc
fredmorcos: Alright

pdns/recursordist/docs/manpages/pdns_recursor.1.rst
```diff
@@ -46,6 +46,11 @@ at `<https://doc.powerdns.com/>`
     chroot the process to *directory*.
 --client-tcp-timeout=<num>
     Timeout in seconds when talking to TCP clients.
+--config
+    Show the current configuration. There are two optional values:
```
omoerbeek, line 50:
```suggestion
    Show the current configuration. There are three optional values:
```

pdns/recursordist/rec-main.cc
fredmorcos: Is it better now?

pdns/recursordist/rec-main.cc
```diff
@@ -2893,9 +2894,49 @@ int main(int argc, char** argv)
...
+        if (!::arg().laxFile(configname.c_str())) {
+          SLOG(g_log << Logger::Warning << "Unable to open configuration file '" << configname << "'" << endl,
+               startupLog->error("No such file", "Unable to open configuration file", "config_file", Logging::Loggable(configname)));
+          exit(1);
+        }
+        ::arg().laxParse(argc, argv);
+
+        if (config == "diff") {
```
omoerbeek, line 2927: We might want to also check before producing a diff. That would mean the test above becomes "check" or "diff".

omoerbeek (approved):
Two small nits

pdns/recursordist/rec-main.cc
```diff
@@ -2893,9 +2894,49 @@ int main(int argc, char** argv)
...
+          ::arg().parse(argc, argv);
+          exit(0);
+        }
+        catch (const ArgException& argException) {
+          SLOG(g_log << Logger::Warning << "Unable to parse configuration file '" << configname << "': " << argException.reason << endl,
+               startupLog->error("Cannot parse configuration", "Unable to parse configuration file", "config_file", Logging::Loggable(configname), "reason", Logging::Loggable(argException.reason)));
+          exit(1);
+        }
+      }
+      else if (config == "test" || config == "diff") {
```
omoerbeek, line 2919: My suggestion for check before diff was not right, sorry. You will need to restructure the `if ... else` blocks a bit to check check before diff working..

omoerbeek (changes requested):

omoerbeek (approved):

PR #11836, opened 2022-08-08
auth catalog zones: simplify migration schemas by appending columns

Habbie:
### Checklist

I have:
- [x] read the [CONTRIBUTING.md](https://github.com/PowerDNS/pdns/blob/master/CONTRIBUTING.md) document
- [x] compiled this code
- [x] tested this code
- [ ] included documentation (including possible behaviour changes)
- [ ] documented the code
- [ ] added or modified regression test(s)
- [ ] added or modified unit test(s)
- [ ]  checked that this code was merged to master

PR #11913, opened 2022-09-08
auth-4.7.x catalog zones: simplify migration schemas by appending columns

Habbie:
### Short description
backport of #11836

### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code


cwchristerw:
Column length too big for column 'options' (max = 16383); use BLOB or TEXT instead

kpfleming:
This isn't really a good place to post problem reports, but you should at least indicate which database backend (and database, and version) you are using.

cwchristerw:
I'm using MariaDB 10.10.2

cwchristerw:
I fixed by using suggested TEXT

kpfleming:
Have you added other columns to the `domains` table? The schema as designed fits within the 65,536 byte limit for MySQL/MariaDB. The error message you show above indicates that your table definition must have other columns consuming around 47,000 bytes of the maximum row size.

Habbie:
Or a charset issue causing characters to consume (way) more than 1 byte each - we have seen this before.

PR #11919, opened 2022-09-08
auth: fix gcc warning

mind04:
### Short description
```
auth-main.cc:133:6: warning: redundant redeclaration of ‘void mainthread()’ in same scope [-Wredundant-decls]
  133 | void mainthread();
      |      ^~~~~~~~~~
In file included from auth-main.cc:55:
auth-main.hh:51:13: note: previous declaration of ‘void mainthread()’
   51 | extern void mainthread();
      |             ^~~~~~~~~~
```

### Checklist
I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/auth-main.hh
+++ b/pdns/auth-main.hh
@@ -48,7 +48,7 @@ extern double avg_latency;
 extern std::unique_ptr<TCPNameserver> TN;
 extern void declareArguments();
 extern void declareStats();
-extern void mainthread();
+void mainthread();
 extern int isGuarded(char**);
 void carbonDumpThread();
 extern bool g_anyToTcp;
--- a/pdns/auth-main.cc
+++ b/pdns/auth-main.cc
@@ -130,8 +130,6 @@ vector<std::shared_ptr<UDPNameserver>> g_udpReceivers;
 NetmaskGroup g_proxyProtocolACL;
 size_t g_proxyProtocolMaximumSize;
 
-void mainthread();
-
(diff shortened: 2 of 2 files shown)


kpfleming:
Just a nit, but the commit message and PR title would be better if they focused on the cause (the redundant declaration) and not the symptom (GCC warning) :-) Future readers of the repository history will learn more from the commit message that way.

pdns/auth-main.hh
```diff
@@ -48,7 +48,7 @@ extern double avg_latency;
 extern std::unique_ptr<TCPNameserver> TN;
 extern void declareArguments();
 extern void declareStats();
-extern void mainthread();
+void mainthread();
```
Habbie, line 51: looks like this line can go entirely

Habbie (changes requested):

pdns/auth-main.hh
Habbie: .. in which case `mainthread` in `auth-main.cc` can even be declared `static`

omoerbeek: `extern` for function prototypes is very old-school anyway

Habbie: Right, I was starting to suspect a bigger cleanup is in order here. I'll merge this one because it quiets the compiler, and then we can do better later.

Habbie (approved):

PR #11951, opened 2022-09-13
make builds pass with -Werror on _my_ system

Habbie:
### Short description
While I would not recommend running with `-Werror` always, I find it convenient to catch silly mistakes during development. The patches in this PR make `-Werror` work on my system. After discussion, perhaps we can upstream some things to the autoconf-archive.

### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/m4/ax_check_sign.m4
+++ b/m4/ax_check_sign.m4
@@ -43,7 +43,7 @@ AC_DEFUN([AX_CHECK_SIGN], [
  AC_CACHE_CHECK([whether $1 is signed], ax_cv_decl_${typename}_signed, [
    AC_COMPILE_IFELSE(
      [AC_LANG_PROGRAM([[$4]],
-       [[ int foo @<:@ 1 - 2 * !((($1) -1) < 0) @:>@ ]])],
+       [[ [[maybe_unused]] int foo @<:@ 1 - 2 * !((($1) -1) < 0) @:>@ ]])],
      [ eval "ax_cv_decl_${typename}_signed=\"yes\"" ],
      [ eval "ax_cv_decl_${typename}_signed=\"no\"" ])
  ])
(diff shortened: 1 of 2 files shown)


omoerbeek:
With this my configure runs fails on MacOS:
```
configure: error: time_t is unsigned, PowerDNS code relies on it being signed```
```
`config.log` snippet:
```
configure:7283: ccache c++ -c -g -O2 -Wall -Wextra -Wshadow -Wno-unused-parameter -Wmissing-declarations -Wredundant-decls -Wno-unused-command-line-argument -Wno-deprecated-copy-with-user-provided-copy -Wno-deprecated-copy -I/opt/homebrew/opt/openssl@1.1/include conftest.cpp >&5
conftest.cpp:46:2: error: expected expression
 [[maybe_unused]] int foo [ 1 - 2 * !(((time_t) -1) < 0) ]
 ^
1 error generated.
```

Habbie:
Interesting, what compiler?

omoerbeek:
> Interesting, what compiler?

```
mbp:pdns otto$ c++ --version
Apple clang version 14.0.0 (clang-1400.0.29.102)
Target: arm64-apple-darwin21.6.0
Thread model: posix
InstalledDir: /Library/Developer/CommandLineTools/usr/bin
```

omoerbeek:
Will push a potential solution. For some reason, the `m4/ax_cxx_compile_stdcxx.m4` one is not hit here...

Habbie:
> For some reason, the `m4/ax_cxx_compile_stdcxx.m4` one is not hit here...

Probably because the compiler is actually instructed to be ++17 in that case?

omoerbeek (approved):
I also upstreamed this: https://github.com/autoconf-archive/autoconf-archive/pull/268

Issue #11804, opened 2022-07-25
AXFR Fails when many SMIMEA Records are Present


nils-wisiol:
- Program: Authoritative
 - Issue type: Bug report

### Short description
AXFR Fails when many SMIMEA Records are Present

### Environment
 - Operating system: Ubuntu Bionic
 - Software version: 4.6
 - Software source: Debian packets at http://repo.powerdns.com/ubuntu bionic-auth-46 main

### Steps to reproduce
Please see minimal example at https://github.com/desec-io/pdns-smimea-axfr-failure

### Expected behaviour
- Full AXFR with all records is sent.
- If AXFR cannot be fully produced, do not send at all
- On receiving side: do not accept truncated AXFR for zone updates

### Actual behaviour
- Sending side: Truncated AXFR is sent, containing only the SOA record.
- Receiving side: commits zone update using truncated AXFR, discarding missing records (impacting availability of services!)

### Other information
1. Minimal example uses lmdb, but we also observed this issue using gmysql backend.
2. We observed that more than just the SOA record is present in the AXFR (but not everything).
3. When removing any RRset from the minimal example, things seem to work. The transfer size is then suspiciously close to 64kB

cc @peterthomassen

peterthomassen:
@Habbie Do you think this'll make it into 4.7.0?

(Background: Zone truncation is pretty serious so we can't risk it, but we're still looking to quickly get rid of the current "workaround" of simply disabling SMIMEA. Also, we don't know whether this can happen with other rdtypes, so we're a little "uneasy" generally.)

Habbie:
I see I was overzealous when I moved this to 4.8.0. I'll have a look soon.

Habbie:
This is two bugs. 

1. as a primary, we may overfill packets, and the way the code flow works, there is little we can do but TC a chunk, which is terrible.
2. as a secondary, we treat a TC chunk as a valid but empty chunk, and then merrily go on to treating the next chunk (containing the closing SOA) as a valid terminator to the stream

#11953 fixes (2), reducing this problem from "random data loss" to "transfers now always fail", which is a good start.

Habbie:
> Do you think this'll make it into 4.7.0?

So - the two changes mentioned in my previous comment will make it into 4.7.0 (unless it turns out they break things, etc.). A full fix (allowing your example zone to actually transfer) is more involved and will likely have to wait.

peterthomassen:
[...]

PR #11954, opened 2022-09-14
auth AXFR server: abort on chunk with TC set

Habbie:
### Short description
This makes the server break the whole transfer if it was about to send a truncated chunk. Part of mitigation for #11804 without yet solving it completely.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnspacket.hh
+++ b/pdns/dnspacket.hh
@@ -58,7 +58,7 @@ public:
 
   int noparse(const char *mesg, size_t len); //!< just suck the data inward
   int parse(const char *mesg, size_t len); //!< parse a raw UDP or TCP packet and suck the data inward
-  const string& getString(); //!< for serialization - just passes the whole packet
+  const string& getString(bool throwsOnTruncation=false); //!< for serialization - just passes the whole packet. If throwsOnTruncation is set, an exception will be raised if the records are too large to fit inside a single DNS payload, instead of setting the TC bit
 
   // address & socket manipulation
   void setRemote(const ComboAddress*, std::optional<ComboAddress> = std::nullopt);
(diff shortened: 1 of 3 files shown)


Habbie:
After discussion with @mind04, the exception can bubble all the way up, but it would be nice to catch it (where I do now) and rethrow it with a zone name.

Habbie:
> After discussion with @mind04, the exception can bubble all the way up, but it would be nice to catch it (where I do now) and rethrow it with a zone name.

Amended.

Habbie:
`Sep 15 10:09:48 TCP nameserver had error, cycling backend: during axfr-out of example.com., this happened: attempt to write an oversized chunk`

pdns/dnspacket.hh
```diff
@@ -58,7 +58,7 @@ public:
 
   int noparse(const char *mesg, size_t len); //!< just suck the data inward
   int parse(const char *mesg, size_t len); //!< parse a raw UDP or TCP packet and suck the data inward
-  const string& getString(); //!< for serialization - just passes the whole packet
+  const string& getString(bool noTruncate=false); //!< for serialization - just passes the whole packet
```
rgacogne, line 61:
```suggestion
  const string& getString(bool throwsOnTruncation=false); //!< for serialization - just passes the whole packet. If throwsOnTruncation is set, an exception will be raised if the records are too large to fit inside a single DNS payload, instead of setting the TC bit
```

pdns/dnspacket.hh
```diff
@@ -105,7 +105,7 @@ public:
   void setQuestion(int op, const DNSName &qdomain, int qtype);  // wipes 'd', sets a random id, creates start of packet (domain, type, class etc)
 
   DTime d_dt; //!< the time this packet was created. replyPacket() copies this in for you, so d_dt becomes the time spent processing the question+answer
-  void wrapup();  // writes out queued rrs, and generates the binary packet. also shuffles. also rectifies dnsheader 'd', and copies it to the stringbuffer
+  void wrapup(bool noTruncate=false);  // writes out queued rrs, and generates the binary packet. also shuffles. also rectifies dnsheader 'd', and copies it to the stringbuffer
```
rgacogne, line 108:
```suggestion
  void wrapup(bool throwsOnTruncation=false);  // writes out queued rrs, and generates the binary packet. also shuffles. also rectifies dnsheader 'd', and copies it to the stringbuffer. If throwsOnTruncation is set, an exception will be raised if the records are too large to fit inside a single DNS payload, instead of setting the TC bit
```

rgacogne (approved):
The code looks good, I'm a bit confused by the name of the boolean variable but nothing that would prevent a merge.

PR #11955, opened 2022-09-14
Clarify return codes for the LUA hooks in the Recursor

franklouwers:
### Short description
Clarifies documentation: not only `ipfilter` and `preresolve` should return `true` when they influence/modify the normal query flow.

### Checklist
I have:
- [ ] compiled this code
- [ ] tested this code
- [x] documented the code


omoerbeek:
Thanks. A few Related things *are* mentioned under `Writing Lua PowerDNS Recursor scripts`, but that text also leaves a few things open to interpretation. The remark is certainly true for the case of changing rcode or records. But other `dq` vars and methods follow different rules. 
I have to double check a few things and then likely come up with some tweaks.

omoerbeek:
@franklouwers I added a few tweaks. Can you check it? Still not happy about the whole page, but this *is* an improvement IMO.

franklouwers:
@omoerbeek Yes, better!

omoerbeek (approved):

PR #11957, opened 2022-09-14
[clang-tidy] Use nullptr in getNSEC3PARAM + init bool at callsite

darnuria:
### Short description

Small cpp modernization found with `clang-tidy`:

- Initialize out parameter to `false` at call site of `getNSEC3PARAM`
- Use `nullptr` instead of `0`
- Avoid bool auto conv check for `p != nullptr` instead

## Alternative

I didn't digged the `std::optional` way, ~due to the absence of `optional` in pdns sources and~ mostly because i am inexperienced with. ~No hard feeling about it.~ It's used in pdns source see for example: <https://github.com/PowerDNS/pdns/blob/9e912e583ac8954fa26c4f1b93782515741caeda/pdns/dnstap.hh> (redacted to add an example).
Why redacted? Made a typo while searching. :facepalm:.

https://en.cppreference.com/w/cpp/utility/optional

### Checklist

I have:
- [x] compiled this code
- [x] tested this code runned unit-test / reg test (partially)
- [ ] documented the code

### DIFF
--- a/modules/bindbackend/bindbackend2.cc
+++ b/modules/bindbackend/bindbackend2.cc
@@ -491,7 +491,7 @@ void Bind2Backend::alsoNotifies(const DNSName& domain, set<string>* ips)
 void Bind2Backend::parseZoneFile(BB2DomainInfo* bbd)
 {
   NSEC3PARAMRecordContent ns3pr;
-  bool nsec3zone;
+  bool nsec3zone = false;
   if (d_hybrid) {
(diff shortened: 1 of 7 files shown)


darnuria:
The title was messed up rewrote it to something more human friendly.

rgacogne (approved):

darnuria:
Little whopsie clicked twice to the button explain the "delete/restore".

PR #11958, opened 2022-09-15
Rec: lock record cache entries

omoerbeek:
### Short description

While we already are very careful to check if a received answer is really the answer to a question we asked an authoritative server this provides an extra layer of protection against spoofing. 

To quote from the docs:

This adds an extra layer of protection---as it limits the window of time cache updates are accepted---at the cost of a less efficient record cache.

The default value of 0 means no extra locking of entries in the record cache occurs. When non-zero, record sets received (e.g. in the Additional Section) will not replace existing record sets in the record cache until the given percentage of the original TTL has expired. So 80 means that an entry with an original TTL of 3600 will only be replaced if its older than  0.8 * 3600 = 2880 seconds. A value of 100 means only expired record sets will be replaced.

There are a few cases where record sets will be replaced anyway:

- Record sets that are expired will always be replaced.
- If the new record set passed DNSSEC validation it will replace an existing entry.
- Record sets produced by refresh-on-ttl-perc tasks will also replace existing record sets.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [X] documented the code

### DIFF
--- a/pdns/recursordist/docs/settings.rst
+++ b/pdns/recursordist/docs/settings.rst
@@ -1684,6 +1684,29 @@ Disabled by default, which also disables outgoing IPv6 support.
 
 Don't log queries.
 
+.. _setting-record-cache-locked-ttl-perc:
+
+``record-cache-locked-ttl-perc``
+--------------------------------
+.. versionadded:: 4.8.0
+
+- Integer
+- Default: 0
+
+Replace record sets in the record cache only after this percentage of the original TTL has passed.
+The PowerDNS Recursor already has several mechanisms to protect against spoofing attempts.
+This adds an extra layer of protection---as it limits the window of time cache updates are accepted---at the cost of a less efficient record cache.
+
+The default value of 0 means no extra locking occurs.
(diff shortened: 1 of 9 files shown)


pdns/recursordist/docs/settings.rst
```diff
@@ -1684,6 +1684,29 @@ Disabled by default, which also disables outgoing IPv6 support.
...
+This adds an extra layer of protection---as it limits the window of time cache updates are accepted---at the cost of a less efficient record cache.
+
+The default value of 0 means no extra locking occurs.
+When non-zero, record sets received (e.g. in the Additional Section) will not replace existing record sets in the record cache until the given percentage of the original TTL has expired.
+A value of 100 means only expired record sets will be replaced.
+
+There are a few cases where records will be replaced anyway:
+
+- Record sets that are expired will always be replaced.
+- If the new record set passed DNSSEC validation it will replace an existing entry.
```
rgacogne, line 1707:
```suggestion
- If the new record set belongs to a DNSSEC-secure zone and successfully passed validation it will replace an existing entry.
```

PR #11989, opened 2022-09-21
Improve error message when invalid values for `local-address` are provided in recursor config file

fredmorcos:
### Short description
Improve error message when invalid values for `local-address` are provided in recursor config file. The recursor used to depend on `parseService` subroutines to parse addresses. This subroutine didn't do much except "scan until colon or EOF is reached".

This PR replaces the use of `parseService` with `ComboAddress` which offers better parsing and error reporting for address values.

This PR also cleans up and modernizes `getHostname`.

### Checklist
I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/misc.cc
+++ b/pdns/misc.cc
@@ -310,51 +310,6 @@ bool stripDomainSuffix(string *qname, const string &domain)
   return true;
 }
 
-static void parseService4(const string& descr, ServiceTuple& st)
-{
-  vector<string> parts;
-  stringtok(parts, descr, ":");
-  if (parts.empty()) {
-    throw PDNSException("Unable to parse '" + descr + "' as a service");
-  }
-  st.host = parts[0];
-  if (parts.size() > 1) {
-    pdns::checked_stoi_into(st.port, parts[1]);
-  }
-}
-
-static void parseService6(const string& descr, ServiceTuple& st)
-{
-  string::size_type pos = descr.find(']');
-  if (pos == string::npos) {
-    throw PDNSException("Unable to parse '" + descr + "' as an IPv6 service");
-  }
-
-  st.host = descr.substr(1, pos - 1);
-  if (pos + 2 < descr.length()) {
-    pdns::checked_stoi_into(st.port, descr.substr(pos + 2));
-  }
-}
-
-void parseService(const string &descr, ServiceTuple &st)
-{
-  if(descr.empty())
-    throw PDNSException("Unable to parse '"+descr+"' as a service");
-
-  vector<string> parts;
-  stringtok(parts, descr, ":");
-
-  if(descr[0]=='[') {
-    parseService6(descr, st);
-  }
-  else if(descr[0]==':' || parts.size() > 2 || descr.find("::") != string::npos) {
-    st.host=descr;
-  }
-  else {
-    parseService4(descr, st);
-  }
-}
-
 // returns -1 in case if error, 0 if no data is available, 1 if there is. In the first two cases, errno is set
 int waitForData(int fd, int seconds, int useconds)
 {
--- a/pdns/misc.cc
+++ b/pdns/misc.cc
@@ -510,17 +510,46 @@ string urlEncode(const string &text)
   return ret;
 }
 
-string getHostname()
+static size_t getMaxHostNameSize()
 {
-#ifndef MAXHOSTNAMELEN
-#define MAXHOSTNAMELEN 255
+#if defined(HOST_NAME_MAX)
+  return HOST_NAME_MAX;
 #endif
 
-  char tmp[MAXHOSTNAMELEN];
-  if(gethostname(tmp, MAXHOSTNAMELEN))
-    return "UNKNOWN";
+#if defined(_SC_HOST_NAME_MAX)
+  auto tmp = sysconf(_SC_HOST_NAME_MAX);
+  if (tmp != -1) {
+    return tmp;
+  }
+#endif
 
-  return string(tmp);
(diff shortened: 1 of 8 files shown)


pdns/misc.cc
```diff
@@ -510,17 +510,14 @@ string urlEncode(const string &text)
...
+std::optional<string> getHostname()
 {
-#ifndef MAXHOSTNAMELEN
-#define MAXHOSTNAMELEN 255
-#endif
-
-  char tmp[MAXHOSTNAMELEN];
-  if(gethostname(tmp, MAXHOSTNAMELEN))
-    return "UNKNOWN";
+  std::array<char, HOST_NAME_MAX> tmp{};
```
omoerbeek, line 470: From http://man.openbsd.org/gethostname:
Note that the corresponding POSIX definition HOST_NAME_MAX in <limits.h> does not
include the terminating NUL character.

So both occurences of `HOST_NAME_MAX` should be `HOST_NAME_MAX + 1`

omoerbeek: Or rather use getMaxHostNameSize()
See also https://docs.powerdns.com/recursor/security-advisories/powerdns-advisory-2020-03.html

fredmorcos: I'm not sure `getMaxHostNameSize()` can be used in a const context. It also - I think - isn't adding space for the NUL byte. I changed the code to use `MAXHOSTNAMELEN` and define it to 256 if it is not defined in accordance with the OpenBSD manpage: `Host names are limited to MAXHOSTNAMELEN (from <sys/param.h>) characters, currently 256. This includes the terminating NUL character.`

omoerbeek: > I'm not sure `getMaxHostNameSize()` can be used in a const context.

It cannot, but you could use a vector instead of an array.

>It also - I think - isn't adding space for the NUL byte.

You need to add 1 indeed.

>I changed the code to use `MAXHOSTNAMELEN` and define it to 256 if it is not defined in accordance with the OpenBSD manpage: `Host names are limited to MAXHOSTNAMELEN (from <sys/param.h>) characters, currently 256. This includes the terminating NUL character.`

The problem with `MAXHOSTNAMELEN` is that it is a non-posix define. So the chances of that name being available are even less than `HOST_NAME_MAX`. I really would use `getMaxHostNameSize()` as it is proven to work on many (even exotic) systems.

omoerbeek:
I see this now (and CI as well):
```
mbp:recursordist otto$ ./testrunner 
Running 463 test cases...
test-misc_hh.cc:217: error: in "test_misc_hh/test_getCarbonHostName": check my_hostname == hostname has failed [mbp_local != mbp_local]
test-misc_hh.cc:218: error: in "test_misc_hh/test_getCarbonHostName": check my_hostname.size() == hostname.size() has failed [9 != 256]

*** 2 failures are detected in the test module "Master Test Suite"
```

fredmorcos:
> I see this now (and CI as well):
> 
> ```
> mbp:recursordist otto$ ./testrunner 
> Running 463 test cases...
> test-misc_hh.cc:217: error: in "test_misc_hh/test_getCarbonHostName": check my_hostname == hostname has failed [mbp_local != mbp_local]
> test-misc_hh.cc:218: error: in "test_misc_hh/test_getCarbonHostName": check my_hostname.size() == hostname.size() has failed [9 != 256]
> 
> *** 2 failures are detected in the test module "Master Test Suite"
> ```
> 
> The issue is that `shrink_to_fit` reduces `capacity`, but we want to actually shrink `size`.

Ah damn. This should shrink back the string from maxHostNameBufSize. Which means it should be fixed in `getHostname()`. Also, I'm not sure whether `std::string` handles the terminating NULL internally. Does `.resize(32)` allocate 32 or 33 bytes internally? From running a small example program it seems to not handle the NULL terminating byte.

omoerbeek:
iirc: `resize()` sets the string _length_, so excluding the terminating null in the C-string representation. The terminating null is not guaranteed to be in the internal buffer. But since C++11 if you use `data()` (or `c_str()`) you _do_ get a null terminated string (which might contain embedded nulls).

fredmorcos:
> But since C++11 if you use `data()` (or `c_str()`) you _do_ get a null terminated string (which might contain embedded nulls).

Does that mean that these methods could potentially incur an allocation in the worst case?

> And `size()`, `resize()` and `capacity()` are consistent with each other in the sense that they _do not_ count the terminating null.

Strictly speaking, I would have expected `capacity()` to deviate, to account for the null byte.

Anyways, I switched the line back to using `resize(strlen(...))`.

rgacogne:
> > But since C++11 if you use `data()` (or `c_str()`) you _do_ get a null terminated string (which might contain embedded nulls).
> 
> Does that mean that these methods could potentially incur an allocation in the worst case?

C++11 specifies that these functions are `const` so it should not.

rgacogne:
> C++11 specifies that these functions are `const` so it should not.

And `noexcept`, by the way!

omoerbeek:
Right, thanks for the clarification. So the terminating null should be there form the start (but is not counted).

omoerbeek (approved):
Looks good. I'd like a review from @Habbie as well, since this touches auth.

Habbie (approved):
nice cleanup, thanks!

fredmorcos:
> nice cleanup, thanks!

Thanks. Ideally, `getCarbonHostName()` should return some sort of `result<string, ErrorMsg>` (or even an `optional<string>` like `getHostname()` now does) instead of throwing an exception, but that would have touched too many other places so I left it as is for now.

PR #12001, opened 2022-09-21
dnsdist: Implement a small MAC address cache

rgacogne:
### Short description

When the MAC address feature is enabled, dnsdist usually keeps seeing the same devices, so it makes sense to keep a tiny (10 entries), short-lived (60s) cache of MAC addresses to speed things up.
We might want to make the size configurable at a later point if this proves useful.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdistdist/dnsdist-mac-address.hh
+++ b/pdns/dnsdistdist/dnsdist-mac-address.hh
@@ -1,0 +1,48 @@
+/*
+ * This file is part of PowerDNS or dnsdist.
+ * Copyright -- PowerDNS.COM B.V. and its contributors
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of version 2 of the GNU General Public License as
+ * published by the Free Software Foundation.
+ *
+ * In addition, for the avoidance of any doubt, permission is granted to
+ * link this program with OpenSSL and to (re)distribute the binaries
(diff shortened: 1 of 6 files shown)


pdns/dnsdistdist/dnsdist-mac-address.hh
```diff
@@ -0,0 +1,48 @@
...
+class MacAddressesCache
+{
+public:
+  static int get(const ComboAddress& ca, char* dest, size_t len);
+
+private:
+  struct Entry
+  {
+    ComboAddress ca;
+    char mac[6];
```
omoerbeek, line 39: Move this field down, so that alignment of the time_t field is better.
Also: why char[6] and not array<uint8_t, 6>? clang-tidy does not like "C stye" arrays at all...

omoerbeek (approved):
Small nits, you might want to define a MacAddress type.

pdns/dnsdistdist/dnsdist-mac-address.hh
rgacogne: That's a very good point, I'll look into it!

rgacogne: Done:

omoerbeek (approved):

PR #12003, opened 2022-09-22
dnsdist: Add a 'single acceptor thread' build option, reducing the number of threads

rgacogne:
### Short description

On some environments that have very limited resources and deal with a fairly low amount of QPS, it does not make sense to have several UDP, TCP, DoT and DoH handler threads to deal with the various network interfaces, so this PR adds a build option where there is a single UDP thread to deal with all incoming queries, a TCP/DoT worker thread can deal with several/all listening IP addresses (merging the `accept()` thread into the worker) and a single DoH worker thread can deal with several/all listening IP addresses as well.
Reducing the number of threads that way yields a lower memory footprint and lower the cost of context switches.
Note that this change is not enabled by default, as it would not make sense for most environments.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code


chbruyand (approved):
Nothing jumped out at me

PR #12011, opened 2022-09-23
rec: Enable SNMP for debian and ubuntu builds.

### LINKED ISSUE #11999: rec: enable snmp on debian/ubuntu builds

omoerbeek:
Fixes #11999

### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code


omoerbeek:
Tested this by doing a build for all debian derived versions currently supported. All show `configure: SNMP: yes` for the actual build part. The `sdist` part shows `configure: SNMP: no` but I do not think that matters.

Habbie:
> The `sdist` part shows `configure: SNMP: no` but I do not think that matters.

Indeed, it does not (or, that is to say, it should not - we've had bugs in the past, but if the package builds are happy, all is good).

omoerbeek:
Mopped the test commit up and force pushed

omoerbeek:
While builds went ok, we should be very careful, on some targets `libsnmp-dev` adds extra flags we do not want

omoerbeek:
So pre-jammy version of ubuntu do not have a pkgconfig for netsnmp. This works and avoids the introduction of extra compiler flags:
```
NET_SNMP_LIBS="-lnetsnmpmibs -lnetsnmpagent -lnetsnmp" ./configure ...

But I do not know yet how to get that into the build in a proper way.

Habbie:
This looks like it might break if the headers live in a weird path. Then again, meanwhile this improves compilation on every other platform (which is more common). I suggest merging this, and, if necessary, finding this comment later so we can say "see, we told ourselves so!"

Habbie (approved):

PR #12017, opened 2022-09-26
Include (with precautions) offending data in messages about too long a label

### LINKED ISSUE #12014: pdns_recursor: Improve logging for "label too long to append" errors to include the domain name.
- Program: Recursor (versions 4.6 and 4.7+)
 - Issue type: Feature request

First of all, thanks for PowerDNS and the great work you guys do. We are using it on our network very successfully handling billions of requests per day.

The problem we have (very minimal) is related to this error message that shows up whenever we reload it:

Sep 25 20:04:01 server pdns_recursor-2[30871]: label too long to append
Sep 25 20:04:01 server pdns_recursor-2[30871]: label too long to append
Sep 25 21:04:04 server syslogd: last message repeated 7 times

Since the logging doesn't tell what domain is causing it, we can't fix it. Ideally, it would be great if it showed the domain name that is causing the problem. Like:

Sep 25 20:04:01 server pdns_recursor-2[30871]: label too long to append on domain name: example.com

That allow us to troubleshoot and fix the problem. I looked at the code base and I think that by modifying dnsname.cc (line 69) from:

```throw std::range_error("label too long to append");```

to:

```throw std::range_error("label too long to append on domain name: " + string(p));```

Should solve it. White at it, it might be good to do the same thing for:

```throw std::range_error("name too long to append: " + string(p));```

Thanks!

### PR DESCRIPTION
**Pull request by omoerbeek**
Fixes #12014

Since we do not know the callers of this, we need to be extra careful not to print random strings. Especially call errors like `DNSName(p, -1)` should not be disastrous.

Note: the `throw std::runtime_error("Found . in wrong position in DNSName "+string(p));` case becomes a `range_error`. If that is not acceptable I should add a `throwSafeRuntimeError` function.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [X] documented the code

### DIFF
--- a/pdns/dnsname.cc
+++ b/pdns/dnsname.cc
@@ -41,6 +41,19 @@ std::ostream & operator<<(std::ostream &os, const DNSName& d)
   return os <<d.toLogString();
 }
 
+void DNSName::throwSafeRangeError(const std::string& msg, const char* buf, size_t length)
+{
+  std::string dots;
+  const size_t maxDNSNameLength = 255; // Maybe this should be a symbol in DNSName itself
+  if (length > maxDNSNameLength) {
+    length = maxDNSNameLength;
+    dots = "...";
+  }
+  std::string label;
+  DNSName::appendEscapedLabel(label, buf, length);
+  throw std::range_error(msg + label + dots);
+}
+
 DNSName::DNSName(const char* p, size_t length)
 {
   if(p[0]==0 || (p[0]=='.' && p[1]==0)) {
@@ -55,7 +68,7 @@ DNSName::DNSName(const char* p, size_t length)
       for(auto iter = pbegin; iter != pend; ) {
         lenpos = d_storage.size();
         if(*iter=='.')
-          throw std::runtime_error("Found . in wrong position in DNSName "+string(p));
+          throwSafeRangeError("Found . in wrong position in DNSName: ", p, length);
         d_storage.append(1, (char)0);
         labellen=0;
         auto begiter=iter;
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: pdns/dnsname.hh [hunk 1]
```diff
@@ -149,7 +149,7 @@ public:
   DNSName& operator+=(const DNSName& rhs)
   {
     if(d_storage.size() + rhs.d_storage.size() > 256) // one extra byte for the second root label
-      throw std::range_error("name too long");
+      throwSafeRangeError("name too long", rhs.d_storage.data(), rhs.d_storage.size());
```
rgacogne, line 152: This one could be a bit confusing since `rhs.d_storage` might not be too long by itself, but getting that information is better than none at all.

pdns/dnsname.cc
```diff
@@ -41,6 +41,19 @@ std::ostream & operator<<(std::ostream &os, const DNSName& d)
   return os <<d.toLogString();
 }
 
+void DNSName::throwSafeRangeError(const std::string& msg, const char* buf, size_t length)
+{
+  std::string dots;
+  const size_t maxDNSNameLength = 255; // Maybe this should be a symbol in DNSName itself
```
rgacogne, line 47: Making that a constant in the `DNSName` class would make sense to me indeed.

rgacogne (approved):

pdns/dnsname.hh
omoerbeek: Maybe "resulting name too long"?

rgacogne: Sounds good!

pdns/dnsname.hh
```diff
@@ -77,6 +77,8 @@ inline unsigned char dns_tolower(unsigned char c)
 class DNSName
 {
 public:
+  static const size_t maxDNSNameLength = 255;
```
rgacogne, line 80:
```suggestion
  static const size_t s_maxDNSNameLength = 255;
```

PR #12025, opened 2022-09-27
Proxy Protocol: Only allocate once when generating the payload

rgacogne:
### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/proxy-protocol.cc
+++ b/pdns/proxy-protocol.cc
@@ -27,12 +27,12 @@
 #define PROXYMAGIC "\x0D\x0A\x0D\x0A\x00\x0D\x0A\x51\x55\x49\x54\x0A"
 #define PROXYMAGICLEN sizeof(PROXYMAGIC)-1
 
-static string proxymagic(PROXYMAGIC, PROXYMAGICLEN);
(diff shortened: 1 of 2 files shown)


pdns/proxy-protocol.cc
```diff
@@ -82,8 +82,7 @@ std::string makeProxyHeader(bool tcp, const ComboAddress& source, const ComboAdd
 
   const uint16_t contentlen = htons(static_cast<uint16_t>(total));
   std::string ret;
-  ret.reserve(total);
-  makeSimpleHeader(command, protocol, contentlen, ret);
+  makeSimpleHeader(command, protocol, contentlen, total, ret);
```
chbruyand, line 85: Maybe just a wording issue, but it looks like you are passing the `total` here, and the allocation in `makeSimpleHeader()`  is expecting an additional count

rgacogne: You are right, something does not look correct at all. I'll re-check everything, thanks!

rgacogne: I just pushed a fix, thanks for spotting this!

chbruyand (approved):

PR #12027, opened 2022-09-27
rec: If new data is auth and existing data is not, replace even if cache locking is active

omoerbeek:
Followup to #11958

Current code gives the right answer (the authoritative records) to the client but does not replace the unauthoritative record set  in the cache by the authoritative one. That's a bit too strict interpretation of the locking feature.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [X] documented the code

### DIFF
--- a/pdns/recursordist/test-syncres_cc10.cc
+++ b/pdns/recursordist/test-syncres_cc10.cc
@@ -1512,8 +1512,8 @@ BOOST_AUTO_TEST_CASE(test_glued_referral_additional_no_update_because_locked)
 
 BOOST_AUTO_TEST_CASE(test_locked_nonauth_update_to_auth)
 {
-  // Test that additional records do not update the cache
-  // We use two zones that share NS and their addresses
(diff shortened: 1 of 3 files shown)


rgacogne (approved):

pdns/recursordist/test-syncres_cc10.cc
```diff
@@ -1510,4 +1510,78 @@ BOOST_AUTO_TEST_CASE(test_glued_referral_additional_no_update_because_locked)
   BOOST_CHECK_NE(firstTTL, secondTTL);
 }
 
+BOOST_AUTO_TEST_CASE(test_locked_nonauth_update_to_auth)
+{
+  // Test that additional records do not update the cache
+  // We use two zones that share NS and their addresses
```
rgacogne, line 1516:
```suggestion
  // Test that a non-bogus authoritative record replaces a non-authoritative one
  // even if the cache is locked
```

rgacogne (changes requested):

rgacogne (approved):

PR #12039, opened 2022-09-28
Dnsdist: coverity 20220928

omoerbeek:
### Short description

Two sets of coverity reports. Should make dnsdist clean again.

Two last two in `dnscryp.cc` look strange as we're modifying through a `const &`. But that happens via the lock guard `operator->` which returns a `T*`, not a `const T*`. 

### Checklist

I have:
- [X] compiled this code
- [ ] tested this code
- [ ] documented the code


rgacogne (approved):
Thanks!

PR #12063, opened 2022-10-05
rec: Compare old and new config for protobuf logger config objects on Lua config reload

omoerbeek:
We already did it for dnstap. This also does the protobuf loggers.

I'm not doing the async part now, as TSAN reports there would be a race. We do a synchronous cleanup of the old config when we see the config changed only.

The case occurring the most: no changes is now handled quickly and that is the main thing.

I'm also adding/removing a few places where the config checking is done:

- do it where we need to know if a certain logger is enabled in both the UDP and TCP input path
- do it in startDoResolve, so any thread handling a non-PC hit gets the new config.
- do it periodically since not al threads actually handle incoming packets or call startDoResolve that often.

I spotted the latter case by changing config and the looking at `rec_control get-remotelogger-stats` and seeing the olde config never being cleaned for some threads.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [X] documented the code

### DIFF
--- a/pdns/pdns_recursor.cc
+++ b/pdns/pdns_recursor.cc
@@ -43,8 +43,8 @@
 
 thread_local std::shared_ptr<RecursorLua4> t_pdl;
 thread_local std::shared_ptr<Regex> t_traceRegex;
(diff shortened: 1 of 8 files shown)


chbruyand (approved):
LGTM!

PR #12066, opened 2022-10-06
auth: Detect invalid bytes in `makeBytesFromHex()`

rgacogne:
### Short description

Also:
- only allocate the required number of bytes in `makeBytesFromHex()`, not twice that.
- switch to a `std::array` in `makeHexDump()`

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/misc.cc
+++ b/pdns/misc.cc
@@ -583,13 +583,13 @@ string U32ToIP(uint32_t val)
 
 string makeHexDump(const string& str)
 {
-  char tmp[5];
+  std::array<char, 5> tmp;
   string ret;
-  ret.reserve((int)(str.size()*2.2));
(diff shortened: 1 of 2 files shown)


pdns/misc.cc
```diff
@@ -599,14 +599,18 @@ string makeBytesFromHex(const string &in) {
     throw std::range_error("odd number of bytes in hex string");
   }
   string ret;
-  ret.reserve(in.size());
+  ret.reserve(in.size() / 2);
+
   unsigned int num;
```
omoerbeek, line 604: decl can go inside loop

omoerbeek (approved):
one nit

Issue #12075, opened 2022-10-07
dnsdist: expungeByName overwrites the output when called multiple times


christofchen:
- Program: dnsdist 
 - Issue type: Bug report

### Short description
When calling expungeByName multiple times on the same line, there is only one line of output (the last one)  

### Environment

 - Operating system: Linux
 - Software version: 1.7.x master
 - Software source: github, self compiled

### Steps to reproduce

preload the cache with entries for www.google.com and www.amazon.com, and expunge the records from the cache in one line:

> getPool(""):getCache():expungeByName("www.google.com") getPool(""):getCache():expungeByName("www.amazon.com")
Expunged 1 records

### Expected behaviour
Expunged 1 records 
Expunged 1 records

### Actual behaviour
Expunged 1 records   (single line)

### Other information
The output should be appended, not overwritten here:
```
diff --git a/pdns/dnsdistdist/dnsdist-lua-bindings-packetcache.cc b/pdns/dnsdistdist/dnsdist-lua-bindings-packetcache.cc
index fb560c568..184ff33c2 100644
--- a/pdns/dnsdistdist/dnsdist-lua-bindings-packetcache.cc
+++ b/pdns/dnsdistdist/dnsdist-lua-bindings-packetcache.cc
@@ -166,7 +166,7 @@ void setupLuaBindingsPacketCache(LuaContext& luaCtx, bool client)
                   qname = DNSName(boost::get<string>(dname));
                 }
                 if (cache) {
-                  g_outputBuffer="Expunged " + std::to_string(cache->expungeByName(qname, qtype ? *qtype : QType(QType::ANY).getCode(), suffixMatch ? *suffixMatch : false)) + " records\n";
+                  g_outputBuffer+="Expunged " + std::to_string(cache->expungeByName(qname, qtype ? *qtype : QType(QType::ANY).getCode(), suffixMatch ? *suffixMatch : false)) + " records\n";
                 }
     });
   luaCtx.registerFunction<void(std::shared_ptr<DNSDistPacketCache>::*)()const>("printStats", [](const std::shared_ptr<DNSDistPacketCache>& cache) {

```

omoerbeek:
Merged

PR #12077, opened 2022-10-07
retain output when expunging from multiple caches

christofchen:
### Short description
append to the output buffer in expungeByName (instead of overwriting)

### Checklist
I have:
- [ X] compiled this code
- [ X] tested this code
- [ ] documented the code
- [X ]  checked that this code was merged to master

This fix addresses #12075


rgacogne (approved):
I was a bit worried about `g_outputBuffer` potentially still having some left-over data from the previous command, but it looks like we are careful to clear it between lines, so it should be OK.

Issue #8717, opened 2020-01-20
Atomic replacement of zone contents using the API


jonathaneen:
- Program: Authoritative
 - Issue type: Feature request

### Short description
In the current functionality of the API of PowerDNS, it seems only possible to modify the contents of a zone by deleting or replacing RRSets in the zone. This way of editing a zone requires knowledge about the current state of the zone in PowerDNS, so a modification in the zone would always require two consecutive requests to the API. 
Please add functionality to replace the contents of a zone in a single, atomic request to the API. The functionality would be similar to 'load-zone' of pdnsutil.

### Usecase
The most important value this feature would add for us, is that a complete zone change could be done in one atomic request. This minimizes the chance of race conditions and it emulates the behaviour of a database transaction in which an update is performed. 

### Description
I'm requesting the addition of a new endpoint to the API of PowerDNS, at which a complete zone with records and comments can be pushed in JSON format. The endpoint should then clear the current zone contents and replace the contents with the new zone contents (preferably in one database transaction). In case of a validation error or any other error, the zone modifications should be rolled back to the original state and the (validation) errors should be returned in the response. 
Changes in zone metadata are beyond the scope of this issue, it only focuses on changes in RRsets, Records and Comments.

zeha:
Would suggest doing this within the existing PUT code; but refactoring all the rrset update code to be reusable might be ... annoying :)

saiachyuth5:
Any updates on this ?

Habbie:
> Any updates on this ?

No, sorry. Plenty of people that want this but none that have decided to type it in so far (which I get, it's quite a task).

dearco:
wait for updating

franklouwers:
For future ref and SEO foo magic: We'd like the equivalent of `pdnsutile load-zone` via the API

klaus-nicat:
I think this can closed, now with #12086 beeing merged.

PR #12086, opened 2022-10-12
API: Auth: replace zone contents et al

zeha:
### Short description
Implementation of #8717.

For emptying out secondary zones, sending an empty rrset array should work. 

Also:
- forbid negative TTL in RRsets
- forbid changing RRsets in Consumer zones (deleting them all is still okay)
- reduce code directly in updateDomainSettingsFromDocument, to make clang-tidy happy

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/ws-auth.cc
+++ b/pdns/ws-auth.cc
@@ -657,6 +657,23 @@ static void extractDomainInfoFromDocument(const Json& document, boost::optional<
   }
 }
 
+/*
+ * Build vector of TSIG Key ids from domain update document.
+ * jsonArray: JSON array element to extract TSIG key ids from.
+ * metadata: returned list of domain key ids for setDomainMetadata
+*/
+static void extractJsonTSIGKeyIds(UeberBackend& B, const Json& jsonArray, vector<string>& metadata) {
+  for(const auto& value : jsonArray.array_items()) {
+    auto keyname(apiZoneIdToName(value.string_value()));
+    DNSName keyAlgo;
+    string keyContent;
+    if (!B.getTSIGKey(keyname, keyAlgo, keyContent)) {
+      throw ApiException("A TSIG key with the name '"+keyname.toLogString()+"' does not exist");
+    }
+    metadata.push_back(keyname.toString());
+  }
+}
+
 // Must be called within backend transaction.
 static void updateDomainSettingsFromDocument(UeberBackend& B, DomainInfo& di, const DNSName& zonename, const Json& document, bool zoneWasModified) {
   boost::optional<DomainInfo::DomainKind> kind;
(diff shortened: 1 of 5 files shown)


pdns/ws-auth.cc
```diff
@@ -1896,10 +1896,76 @@ static void apiServerZoneDetail(HttpRequest* req, HttpResponse* resp) {
...
+      di.backend->getDomainMetadataOne(zonename, "SOA-EDIT", soaEditKind);
+
+      vector<DNSResourceRecord> new_records;
+      vector<Comment> new_comments;
+
+      for (const auto& rrset : rrsets.array_items()) {
+        DNSName qname = apiNameToDNSName(stringFromJson(rrset, "name"));
+        apiCheckQNameAllowedCharacters(qname.toString());
+        QType qtype;
+        qtype = stringFromJson(rrset, "type");
```
cmouse, line 1920: why not `QType qtype = stringFromJson(rrset, "type");`

pdns/ws-auth.cc
```diff
@@ -1896,10 +1896,76 @@ static void apiServerZoneDetail(HttpRequest* req, HttpResponse* resp) {
...
+        if (qtype.getCode() == 0) {
+          throw ApiException("RRset "+qname.toString()+" IN "+stringFromJson(rrset, "type")+": unknown type given");
+        }
+        if (rrset["records"].is_array()) {
+          int ttl = intFromJson(rrset, "ttl");
```
cmouse, line 1925: ttl could be validated to be positive, or maybe it could be done in checkNewRecords?

cmouse: should 0 be default?

regression-tests.api/test_Zones.py
```diff
@@ -2304,11 +2301,68 @@ def test_put_slave_tsig_key_ids_non_existent(self):
...
             'slave_tsig_key_ids': [keyname]
         }
-        r = self.session.put(self.url('/api/v1/servers/localhost/zones/' + name),
-                             data=json.dumps(payload),
-                             headers={'content-type': 'application/json'})
-        self.assertEqual(r.status_code, 422)
-        self.assertIn('A TSIG key with the name', r.json()['error'])
+        self.put_zone(name, payload, expect_error='A TSIG key with the name')
+
+    def test_zone_replace_rrsets_basic(self):
```
cmouse, line 2306: could add test for invalid ttl.

pdns/ws-auth.cc
zeha: QType cannot initialize from string:
```
ws-auth.cc:2142:13: error: no viable conversion from 'std::string' (aka 'basic_string<char, char_traits<char>, allocator<char>>') to 'QType'
      QType qtype = stringFromJson(rrset, "type");
```

pdns/ws-auth.cc
zeha: defaulting the ttl seems wrong to me, nobody has ever asked for it in the other endpoints.

I've reworked *FromJson to add uintFromJson and have that raise.

regression-tests.api/test_Zones.py
zeha: added

pdns/ws-auth.cc
cmouse: right.

zeha:
Rebased to fix conflict with 580f1f76bc6a537cdc83642af4fb7bd20a84262d

zeha:
This includes the single commit from #12757 to make the api tests not hang.

zeha:
green tests :heart:

Habbie:
12757 was merged

zeha:
This PR grew a few cleanups prompted by clang-tidy. Rebasing them is inconvenient, so I'd appreciate it if this won't linger around for too long.

mind04:
It would be nice to also add a test to ensure that a broken zone (e.g. CNAME and other records exists) is rejected.

zeha:
Rebased and added a (multi-case) test as suggested by @mind04. If there are addtl. test suggestions for invalid combinations it should be easy to add them.

mind04:
Requiring a SOA to clear out a secondary zone is harmful. This will result in a secondary denying everything in the zone until the updated zone is re transferred. Without a SOA queries for the zone are answered with a more harmless "refused" until the zone is re transferred.

Apart from cleaning the caches for secondary zones a call to setStale() is required and will speedup the retransfer.
Primary zones require "unsetting" notified serial or a forced notify.

zeha:
Added a commit to allow sending empty rrset for secondary zones. This also ignores a `false` return value from dk.rectifyZone, which seems like the right thing to do (even though it could be communicated better from rectifyZone).

mind04:
Clearing consumer zones is fine, loading them makes no sense and should not be possible, unless you want to do the processing as well.

zeha:
> Clearing consumer zones is fine, loading them makes no sense and should not be possible, unless you want to do the processing as well.

It is possible today to create a consumer zone and fill it using the other API operations. This PR does not change this.

If consumer zones should be restricted in API, that should go into another PR. Is there a ticket describing the valid and invalid usecases?

zeha:
> Allowing records in a consumer zones in in the rest of the api is a bug.
> 
> You should not allow replacing consumer zones unless you want to do the full processing of the new catalog (please don't ;) ). 

Well, I can add some code to forbid changing consumer zones here in this new API. For the existing API surface, thats really out of scope.

> The same applies btw to presigned zones since rectifyZone() is incompatible with presigned zones. rectifyZone() is rectifying according to the PowerDNS rules. A presigned zone does not necessarily follow those rules.  This is why the presigned check is present in rectyfyZone() so please do not ignore the return value.

There's already a check before `dk.rectifyZone` to skip presigned zones. Don't know what the plan is for API-RECTIFY=1 + presigned. Thats probably a misconfiguration.

> Clearing consumer and presigned zones is the only valid option.

It would be really good to have a doc somewhere describing what can be done to what zone type and what is supported. Right now these rules do not seem to be enforced consistently and in some places not at all.

kpfleming:
> It would be really good to have a doc somewhere describing what can be done to what zone type and what is supported. Right now these rules do not seem to be enforced consistently and in some places not at all.

Next week I'm going to be working a docs PR to improve the "Modes of Operation" page, and as part of that I'm also planning on a "Zone Kinds" page, since there is a lot of interaction between zone kinds and operation modes.

zeha:
I've added additional restrictions for Consumer zones on the existing API surface.

Habbie (approved):
Wonderful cleanup in the process, thanks. I see nothing obviously wrong, and I get the impression that concerns from @mind04 were addressed.

@mind04 any comment?

mind04 (approved):

PR #12100, opened 2022-10-19
dnsdist: Properly handle single-SOA XFR responses

### LINKED ISSUE #12099: Dnsdist "stuck" after a few empty IXFRs if config has setMaxTCPClientThreads
- Program: dnsdist 
 - Issue type: Bug report

### Short description

Dnsdist gets "stuck" after completing a number of "empty IXFRs" (ie IXFR for current SOA.SERIAL) if the config has eg `setMaxTCPClientThreads(20)`.
IXFRs that has changes (ie IXFR for past SOA.SERIAL) or AXFR does not seem to trigger this problem.

### Environment

 - Software version: 1.7.2

### Steps to reproduce

0. Prereq: ensure you have an IXFR-capable nameserver. I validated the problem primarily with ixfrdist, but pdns-auth likely also works for the actual breaking case (IXFR for current SOA.SERIAL), although I think it will "unstick" dnsdist by closing connections after a while.

1.  Install dnsdist, configure it like so:
```
addLocal("127.0.0.1")

newServer({address='127.0.3.1:53', pool='ixfrdist', checkName="test.example", checkType="SOA" })
addAction(AllRule(), PoolAction('ixfrdist'))

setMaxTCPClientThreads(20)
```

2. Run something like this (example for current SOA.SERIAL = 2):
```
for f in `seq 1 100`; do echo $f ; dig @127.0.0.1 test.example IXFR=2 ; sleep 0.1 ; done
```

### Expected behaviour

All rounds should complete

### Actual behaviour

Every single time the loop gets stuck after some number of rounds as dnsdist stops responding to TCP queries (in my tests it seemed to be after round 10, but idk if that is reliably so).

(Also compare what happens with `IXFR=1` (previous serial), `AXFR`, or without `setMaxTCPClientThreads` in dnsdist.conf; none of those variations of the above scenario results in the same breakage.)

### Other information

At the point when dnsdist is "stuck" it has a bunch of connections remaining open to the backend (as observable in eg `netstat -an`). 

Possibly the issue is something along the lines that dnsdist doesn't close connections after completing the empty IXFR, and then is stuck after having exhausted all TCP connections to the backend.

For me personally, the *immediate* problem has been solved by simply removing `setMaxTCPClientThreads` from dnsdist.conf; the meaning of that directive has changed quite radically since the time I added it to the config anyway, and it isn't really of interest to me anymore after having read documentation for current dnsdist.
However, the above behavior seems to indicate that something is wrong in the IXFR handling, which may still manifest in some other way, or at the very least for others who do want to use `setMaxTCPClientThreads`.

rgacogne:
### Short description

From rfc1995 section 2 "Brief Description of the Protocol":

"If an IXFR query with the same or newer version number than that of the server is received, it is replied to with a single SOA record of the server's current version, just as in AXFR."

Until now we considered such a message to be an unfinished response to the pending {A,I}XFR, waiting for more DNS messages to come up and keeping the connection open for as long as the remote host was willing to accept that. This causes an issue for servers keeping the connection open for a very long time, like ixfrdist.

Fixes #12099 

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/pdns/dnsdistdist/dnsdist-tcp-downstream.cc
+++ b/pdns/dnsdistdist/dnsdist-tcp-downstream.cc
@@ -803,7 +803,7 @@ bool TCPConnectionToBackend::isXFRFinished(const TCPResponse& response, TCPQuery
         if (query.d_xfrMasterSerial == 0) {
           // store the first SOA in our client's connection metadata
           query.d_xfrMasterSerial = serial;
-          if (query.d_xfrMasterSerial == query.d_xfrQuerySerial) {
+          if (query.d_xfrMasterSerial <= query.d_xfrQuerySerial) {
             /* This is the first message with a master SOA:
                RFC 1995 Section 2:
                  If an IXFR query with the same or newer version number
--- a/pdns/dnsdistdist/dnsdist-tcp-downstream.cc
+++ b/pdns/dnsdistdist/dnsdist-tcp-downstream.cc
@@ -803,7 +803,7 @@ bool TCPConnectionToBackend::isXFRFinished(const TCPResponse& response, TCPQuery
         if (query.d_xfrMasterSerial == 0) {
           // store the first SOA in our client's connection metadata
           query.d_xfrMasterSerial = serial;
-          if (query.d_idstate.qtype == QType::IXFR && query.d_xfrMasterSerial <= query.d_ixfrQuerySerial) {
+          if (query.d_idstate.qtype == QType::IXFR && (query.d_xfrMasterSerial == query.d_ixfrQuerySerial || rfc1982LessThan(query.d_xfrMasterSerial, query.d_ixfrQuerySerial))) {
             /* This is the first message with a master SOA:
                RFC 1995 Section 2:
                  If an IXFR query with the same or newer version number
--- a/pdns/dnsdistdist/dnsdist-tcp-downstream.cc
+++ b/pdns/dnsdistdist/dnsdist-tcp-downstream.cc
@@ -149,6 +149,62 @@ void TCPConnectionToBackend::release(){
   }
 }
 
+static uint32_t getSerialFromRawSOAContent(const std::vector<uint8_t>& raw)
+{
+  /* minimal size for a SOA record, as defined by rfc1035:
+     MNAME (root): 1
+     RNAME (root): 1
+     SERIAL: 4
+     REFRESH: 4
+     RETRY: 4
+     EXPIRE: 4
+     MINIMUM: 4
+     = 22 bytes
+  */
+  if (raw.size() < 22) {
+    throw std::runtime_error("Invalid content of size " + std::to_string(raw.size()) + " for a SOA record");
+  }
+  /* As rfc1025 states that "all domain names in the RDATA section of these RRs may be compressed",
+     and we don't want to parse these names, start at the end */
+  uint32_t serial = 0;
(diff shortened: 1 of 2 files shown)


pdns/dnsdistdist/dnsdist-tcp-downstream.cc
```diff
@@ -760,22 +800,34 @@ bool TCPConnectionToBackend::isXFRFinished(const TCPResponse& response, TCPQuery
         }
         auto raw = unknownContent->getRawContent();
         auto serial = getSerialFromRawSOAContent(raw);
-        ++query.d_xfrSerialCount;
         if (query.d_xfrMasterSerial == 0) {
           // store the first SOA in our client's connection metadata
-          ++query.d_xfrMasterSerialCount;
           query.d_xfrMasterSerial = serial;
+          if (query.d_xfrMasterSerial == query.d_xfrQuerySerial) {
```
hlindqvist, line 806: Should this be `query.d_xfrMasterSerial <= query.d_xfrQuerySerial` as per the RFC citation below? (If not, maybe comment why?)

rgacogne: Yes, it should, many thanks for spotting that!

pdns/dnsdistdist/dnsdist-tcp-downstream.cc
```diff
@@ -190,6 +246,10 @@ static void prepareQueryForSending(TCPQuery& query, uint16_t id, QueryState quer
       query.d_proxyProtocolPayloadAddedSize = 0;
     }
   }
+  if (query.d_idstate.qclass == QClass::IN && (query.d_idstate.qtype == QType::AXFR || query.d_idstate.qtype == QType::IXFR)) {
```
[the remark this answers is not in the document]
rgacogne: Ugh, I was wondering about that but was indeed confused about that sentence. Thanks again, I'll fix it.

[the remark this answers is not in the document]
hlindqvist: Great, thanks!

Fwiw, the errata for RFC1995 has a confirmation that the "just as in AXFR" mention is in error: https://www.rfc-editor.org/errata/eid3196

pdns/dnsdistdist/dnsdist-tcp-downstream.cc
```diff
@@ -760,22 +800,34 @@ bool TCPConnectionToBackend::isXFRFinished(const TCPResponse& response, TCPQuery
         }
         auto raw = unknownContent->getRawContent();
         auto serial = getSerialFromRawSOAContent(raw);
-        ++query.d_xfrSerialCount;
         if (query.d_xfrMasterSerial == 0) {
           // store the first SOA in our client's connection metadata
-          ++query.d_xfrMasterSerialCount;
           query.d_xfrMasterSerial = serial;
+          if (query.d_idstate.qtype == QType::IXFR && query.d_xfrMasterSerial <= query.d_ixfrQuerySerial) {
```
Habbie, line 806: I think you want `rfc1982LessThan` from `misc.hh` here

rgacogne: Oh, right!

Habbie (approved):

PR #12114, opened 2022-10-21
dnsdist: Fix 'comparison of integer expressions of different signedness' warning

rgacogne:
### Short description

We are making sure that our computation will not overflow the maximum value of a `time_t`, and we know that this maximum value is positive, so we can use a `size_t` to do the comparison.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdistdist/dnsdist-backend.cc
+++ b/pdns/dnsdistdist/dnsdist-backend.cc
@@ -587,7 +587,7 @@ void DownstreamState::updateNextLazyHealthCheck(LazyHealthCheckStats& stats)
     else {
       time_t backOff = d_config.d_lazyHealthCheckMaxBackOff;
       const uint16_t failedTests = currentCheckFailures;
-      double backOffCoeffTmp = std::pow(2U, failedTests);
+      double backOffCoeffTmp = std::pow(2.0, failedTests);
       if (backOffCoeffTmp != HUGE_VAL && backOffCoeffTmp <= std::numeric_limits<time_t>::max()) {
         time_t backOffCoeff = static_cast<time_t>(backOffCoeffTmp);
         if ((std::numeric_limits<time_t>::max() / d_config.d_lazyHealthCheckFailedInterval) >= backOffCoeff) {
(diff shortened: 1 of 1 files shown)


pdns/dnsdistdist/dnsdist-backend.cc
```diff
@@ -588,7 +588,7 @@ void DownstreamState::updateNextLazyHealthCheck(LazyHealthCheckStats& stats)
       const uint16_t failedTests = currentCheckFailures;
       size_t backOffCoeff = std::pow(2U, failedTests);
       time_t backOff = d_config.d_lazyHealthCheckMaxBackOff;
-      if ((std::numeric_limits<time_t>::max() / d_config.d_lazyHealthCheckFailedInterval) >= backOffCoeff) {
+      if ((static_cast<size_t>(std::numeric_limits<time_t>::max()) / d_config.d_lazyHealthCheckFailedInterval) >= backOffCoeff) {
```
omoerbeek, line 591: I think this does not work on platforms with 32-bit `size_t` and 64-bit `time_t` (e.g all OpenBSD 32-bit platforms). Maybe make `backOffCoeff` a `time_t` instead?
```
time_t backOffCoeff = std::pow(2, failedTests);
```

omoerbeek (changes requested):

pdns/dnsdistdist/dnsdist-backend.cc
rgacogne: I pushed a new version which should be better, or at least I hope so :)

pdns/dnsdistdist/dnsdist-backend.cc
```diff
@@ -585,13 +585,16 @@ void DownstreamState::updateNextLazyHealthCheck(LazyHealthCheckStats& stats)
...
     else {
-      const uint16_t failedTests = currentCheckFailures;
-      size_t backOffCoeff = std::pow(2U, failedTests);
       time_t backOff = d_config.d_lazyHealthCheckMaxBackOff;
-      if ((std::numeric_limits<time_t>::max() / d_config.d_lazyHealthCheckFailedInterval) >= backOffCoeff) {
-        backOff = d_config.d_lazyHealthCheckFailedInterval * backOffCoeff;
-        if (backOff > d_config.d_lazyHealthCheckMaxBackOff || (std::numeric_limits<time_t>::max() - now) <= backOff) {
-          backOff = d_config.d_lazyHealthCheckMaxBackOff;
+      const uint16_t failedTests = currentCheckFailures;
+      double backOffCoeffTmp = std::pow(2U, failedTests);
```
omoerbeek, line 590:
```suggestion
      double backOffCoeffTmp = std::pow(2.0, failedTests);
```

omoerbeek (approved):
One nit.

PR #12125, opened 2022-10-24
Timout handling for ixfrs as a client.

omoerbeek:
One complicating factor is that this is shared code, but auth and rec do not agree on the definiton of the timeout value: auth states it is a maximum idle time, while rec state it is the total xfr time. While both apporaches make sense and in the end we would like to enforce both, we now go for a more simple solution that respects auth or rec behaviour based on a flag.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/ixfr.cc
+++ b/pdns/ixfr.cc
@@ -123,9 +123,13 @@ vector<pair<vector<DNSRecord>, vector<DNSRecord> > > processIXFRRecords(const Co
 }
 
 // Returns pairs of "remove & add" vectors. If you get an empty remove, it means you got an AXFR!
-vector<pair<vector<DNSRecord>, vector<DNSRecord> > > getIXFRDeltas(const ComboAddress& primary, const DNSName& zone, const DNSRecord& oursr, 
-                                                                   const TSIGTriplet& tt, const ComboAddress* laddr, size_t maxReceivedBytes)
+vector<pair<vector<DNSRecord>, vector<DNSRecord>>> getIXFRDeltas(const ComboAddress& primary, const DNSName& zone, const DNSRecord& oursr,
+                                                                 uint16_t xfrTimeout, bool totalTimeout,
+                                                                 const TSIGTriplet& tt, const ComboAddress* laddr, size_t maxReceivedBytes)
 {
+  // Auth documents xfrTimeout to be a max idle time
+  // Rec documents it do be a total XFR time
+  //
   vector<pair<vector<DNSRecord>, vector<DNSRecord> > >  ret;
   vector<uint8_t> packet;
   DNSPacketWriter pw(packet, zone, QType::IXFR);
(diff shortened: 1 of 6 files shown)


omoerbeek:
Did some extra testing with an instrumented `ixfdist`. The timeout handling seems to work as intended. Marking ready for review.

pdns/ixfr.cc
```diff
@@ -123,9 +123,13 @@ vector<pair<vector<DNSRecord>, vector<DNSRecord> > > processIXFRRecords(const Co
...
 
 // Returns pairs of "remove & add" vectors. If you get an empty remove, it means you got an AXFR!
-vector<pair<vector<DNSRecord>, vector<DNSRecord> > > getIXFRDeltas(const ComboAddress& primary, const DNSName& zone, const DNSRecord& oursr, 
-                                                                   const TSIGTriplet& tt, const ComboAddress* laddr, size_t maxReceivedBytes)
+vector<pair<vector<DNSRecord>, vector<DNSRecord>>> getIXFRDeltas(const ComboAddress& primary, const DNSName& zone, const DNSRecord& oursr,
+                                                                 uint16_t xfrTimeout, bool totalTimeout,
+                                                                 const TSIGTriplet& tt, const ComboAddress* laddr, size_t maxReceivedBytes)
 {
+  // Auth documents xfrTimeout to be a max idle time
+  // Rec documents it do be a total XFR time
```
Habbie, line 131:
```suggestion
  // Auth documents xfrTimeout to be a max idle time (sets totalTimeout=false)
  // Rec documents it to be a total XFR time (sets totalTimeout=true)
```

pdns/ixfr.cc
```diff
@@ -157,20 +161,38 @@ vector<pair<vector<DNSRecord>, vector<DNSRecord> > > getIXFRDeltas(const ComboAd
...
-  //  cout<<"Connected"<<endl;
-  s.writen(msg);
+  }
+  s.setNonBlocking();
+
+  const time_t xfrStart = time(nullptr);
+
+  // Helper function: if we have a total timeout, check it and set elapsed to the total time taken sofar,
+  // otherwise set elapsed to 0, making the total time limit ineffective
+  const auto timeoutChecker = [=] () -> time_t {
```
Habbie, line 173: perhaps explicit capture of `xfrStart` would be clearer

Habbie (approved):
functionally good, minor nits

pdns/ixfr.cc
omoerbeek: It would become: ` [totalTimeout, xfrStart, xfrTimeout, &zone, &primary]`

Habbie: oh, that's more than I realised. Nevermind :)

Issue #6880, opened 2018-08-24
assorted ixfrdist issues


Habbie:
Updating this while dogfooding. We can split into multiple tickets later.

* serial checks (`getSerialFromMaster()`) does not have a timeout
* ports in domains->masters do not default to 53 (and I suspect, neither to ports on listen addresses)
* figuring out why ^C 'hangs on shutdown' is only possible with --debug

pieterlexis:
> figuring out why ^C 'hangs on shutdown' is only possible with --debug

```
 △ src/PowerDNS/pdns ./pdns/ixfrdist --config pdns/ixfrdist.yml
[NOTICE] IXFR distributor version 0.0.ixfrdistfixes.g8f0ca614a.dirty starting up!
[NOTICE] ACL set to 127.0.0.0/8, ::1/128.
[NOTICE] Update Thread started
[WARNING] Unable to get SOA serial update for 'example.com' from master 127.0.0.1: Reading from a socket: Connection refused
^C[NOTICE] Got Interrupt signal, stopping
[NOTICE] Shutting down!
[NOTICE] UpdateThread stopped
[NOTICE] IXFR distributor stopped
```
Should this report "stopping all threads"?

Habbie:
> Should this report "stopping all threads"?

I don't follow, the best response I have is 'perhaps the ^C response can mention what threads are supposed to die so the user can at least see which threads have not in fact died yet'.

pieterlexis:
There is no signalling between threads, other than a globar var telling them to stop. I can elaborate on the message telling the user "will quit when current transfers are complete", additionally, I could ensure that a second ^C will kill regardless

Habbie:
> I could ensure that a second ^C will kill regardless

This appears to work already, to be clear!

Habbie:
* `[ERROR] Unable to read 'listen' value: yaml-cpp: error at line 0, column 0: bad conversion` when ':53' (without IP) is in the `listen` array. Presumably the line/column could be better.

pieterlexis:
>     [ERROR] Unable to read 'listen' value: yaml-cpp: error at line 0, column 0: bad conversion when ':53' (without IP) is in the listen array. Presumably the line/column could be better.

this is yaml-cpp not knowing that info :(

Habbie:
* `12345.partial` files may stick around
* when I had ~15 non-partial files for a zone, and changed `keep` from 20 to 2, suddenly I had ~15 'partial' files

[...]

PR #12127, opened 2022-10-24
Fix multiple-version IXFR request handling in ixfrdist

hlindqvist:
### Short description

Ixfrdist used to generate invalid IXFR responses if it has multiple zone versions between the requested serial and the latest serial. See https://github.com/PowerDNS/pdns/issues/6880#issuecomment-420980817

This essentially moves the adding of the enclosing "latest serial" SOAs out of the loop which adds the changes of all the zone versions.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/ixfrdist.cc
+++ b/pdns/ixfrdist.cc
@@ -717,23 +717,33 @@ static bool handleIXFR(int fd, const ComboAddress& destination, const MOADNSPars
     return handleAXFR(fd, mdp);
   }
 
-  std::vector<std::vector<uint8_t>> packets;
-  for (const auto& diff : toSend) {
-    /* An IXFR packet's ANSWER section looks as follows:
-     * SOA new_serial
-     * SOA old_serial
-     * ... removed records ...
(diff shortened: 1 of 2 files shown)


hlindqvist:
My suspicion regarding the test failure seen here, is that it actually indicates that the test support function `waitUntilCorrectSerialIsLoaded` is not truly waiting for the zone with the correct serial to be loaded.
For me these tests have been passing consistently, and as some kind of stab at validating the above hypothesis I added `checkFullZone` calls for the new tests, and that also fails in the Github-triggered test run (ie AXFR also does not serve the expected zone version).

I'll take a closer look at `waitUntilCorrectSerialIsLoaded`, I suppose it could (should?) query ixfrdist for SOA until it serves a record with the correct serial instead of waiting for the test-XFR-server to signal it has finished sending the AXFR (this seems racy, as it doesn't reflect ixfrdist's state).

Habbie (approved):

PR #11709, opened 2022-06-20
Rec: systemd-journal structured logging followup

omoerbeek:
### Short description

As @mnordhoff noted, the new systemd-journal backend isn't nice for human consumption. So switch back to the row of `k=v` pairs output by default and introduce a setting to use the systemd-journal backend. This setting also allows for other backends to be used (in the future).

Also: respect loglevel. The traditional output was filtered, but for structured logging, that was not done. While there, shortcut the traditional way as well, to avoid unnecessary churn.

Should fix #11705 and #11706

### Checklist

I have:
- [X] read the [CONTRIBUTING.md](https://github.com/PowerDNS/pdns/blob/master/CONTRIBUTING.md) document
- [X] compiled this code
- [X] tested this code
- [X] included documentation (including possible behaviour changes)
- [ ] documented the code
- [ ] added or modified regression test(s)
- [ ] added or modified unit test(s)

Issue #11706, opened 2022-06-17
Recursor startup logging missing details


mnordhoff:
- Program: Recursor
 - Issue type: Bug report

### Short description
I upgraded Recursor master and the startup logging lost all detail. For example it repeatedly logs "Listening for queries" but not what IPs or ports it's listening on.

See also #11705.

### Environment
 - Operating system: Ubuntu 20.04 (amd64)
 - Software version: 4.8.0~alpha0+master.335.g86727d60b-1pdns.focal
 - Software source: PowerDNS repository

I was previously using 4.8.0~alpha0+master.255.gf049f691b-1pdns.focal.

### Steps to reproduce
1. Install?
2. Maybe enable `log-common-errors`?

### Expected behaviour
Detailed logging similar to or the same as what it was before

### Actual behaviour
`syslog` shows:
```
Jun 17 14:21:24 ultimate pdns-recursor[2734553]: PowerDNS Recursor 4.8.0-alpha0.335.master.g86727d60b (C) 2001-2022 PowerDNS.COM BV
Jun 17 14:21:24 ultimate pdns-recursor[2734553]: Using 64-bits mode. Built using gcc 9.4.0 on Jun 17 2022 11:58:38 by root@01cab996a69e.
Jun 17 14:21:24 ultimate pdns-recursor[2734553]: PowerDNS comes with ABSOLUTELY NO WARRANTY. This is free software, and you are welcome to redistribute it according to the terms of the GPL version 2.
Jun 17 14:21:24 ultimate pdns_recursor[2734553]: If using IPv6, please raise sysctl net.ipv6.route.max_size to a size >= 16384
Jun 17 14:21:24 ultimate pdns_recursor[2734553]: Enabling IPv4 transport for outgoing queries
Jun 17 14:21:24 ultimate pdns_recursor[2734553]: Enabling IPv6 transport for outgoing queries
Jun 17 14:21:24 ultimate pdns_recursor[2734553]: Read Trust Anchors from file, no changes detected
Jun 17 14:21:24 ultimate pdns_recursor[2734553]: Setting access control
Jun 17 14:21:24 ultimate pdns_recursor[2734553]: Will not send queries to
Jun 17 14:21:24 ultimate pdns_recursor[2734553]: Redirecting queries
Jun 17 14:21:24 ultimate pdns_recursor[2734553]: message repeated 4 times: [ Redirecting queries]
[...]

Issue #11705, opened 2022-06-17
rec: Task queue(?) logging "resolving" and "done"


mnordhoff:
- Program: Recursor
 - Issue type: Bug report

### Short description
I upgraded Recursor master and now it's frequently logging "`resolving`" and "`done`". I'm guessing it's coming from the task queue system (fetching nameserver `AAAA` records) but it's hard to be sure. (I can't just grep the codebase for "done"...)

I only have `log-common-errors` and otherwise default logging, so this is surprising. And without further detail, it's borderline useless.

### Environment
 - Operating system: Ubuntu 20.04 (amd64)
 - Software version: 4.8.0~alpha0+master.335.g86727d60b-1pdns.focal
 - Software source: PowerDNS repository

I was previously using 4.8.0~alpha0+master.255.gf049f691b-1pdns.focal.

### Steps to reproduce
1. Install?
2. Maybe enable `log-common-errors`?
3. `dig mattnordhoff.net` or something that will result in `AAAA` tasks

### Expected behaviour
No logging

### Actual behaviour
```
Jun 17 14:22:55 ultimate pdns_recursor[2734553]: resolving
Jun 17 14:22:55 ultimate pdns_recursor[2734553]: resolving
Jun 17 14:22:55 ultimate pdns_recursor[2734553]: done
Jun 17 14:22:55 ultimate pdns_recursor[2734553]: resolving
Jun 17 14:22:55 ultimate pdns_recursor[2734553]: done
Jun 17 14:22:55 ultimate pdns_recursor[2734553]: resolving
Jun 17 14:22:55 ultimate pdns_recursor[2734553]: done
Jun 17 14:22:55 ultimate pdns_recursor[2734553]: resolving
Jun 17 14:22:55 ultimate pdns_recursor[2734553]: done
Jun 17 14:22:55 ultimate pdns_recursor[2734553]: resolving
Jun 17 14:22:55 ultimate pdns_recursor[2734553]: done
Jun 17 14:22:56 ultimate pdns_recursor[2734553]: message repeated 3 times: [ done]
```

### Other information
```
$ sudo rg -vN '^(#|$)' /etc/powerdns/recursor.conf
aggressive-nsec-cache-size=10000
[...]

PR #12128, opened 2022-10-25
enable structured logging by default in systemd

SjonHortensius:
when running as a systemd service - we want to enable structured logging, also allows removing some now-outdated settings

### Checklist

I have:
- [ ] compiled this code
- [X] tested this code
- [ ] documented the code

would you want me to propagate this change on other places as well?

### DIFF
--- a/docs/running.rst
+++ b/docs/running.rst
@@ -28,18 +28,6 @@ the inner process and terminate it, after which it terminates itself.
 Requests that require data from the actual nameserver are passed to the
 inner process as well.
 
-Logging to syslog on systemd-based operating systems
-----------------------------------------------------
-
-By default, logging to syslog is disabled in the the systemd unit file
-to prevent the service logging twice, as the systemd journal picks up
-the output from the process itself.
-
-Removing the ``--disable-syslog`` option from the ``ExecStart`` line
-using ``systemctl edit --full pdns`` enables logging to syslog.
-
-.. _logging-to-syslog:
-
 Logging to syslog
 -----------------
 This chapter assumes familiarity with syslog, the unix logging device.
--- a/pdns/recursordist/pdns-recursor.service.in
+++ b/pdns/recursordist/pdns-recursor.service.in
@@ -7,7 +7,7 @@ Before=nss-lookup.target
 After=network-online.target
 
 [Service]
-ExecStart=@sbindir@/pdns_recursor --daemon=no --write-pid=no --disable-syslog --log-timestamp=no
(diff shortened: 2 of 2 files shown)


omoerbeek:
Well, that is a question of taste/opinion. I you prefer the journal stream it is easy to enable it, I do not want to enable it by default at this moment.

SjonHortensius:
I'm not sure I follow. I modified the `.service`, which will only be used when running under systemd implying the journal stream is used anyway. The current output is already in this "structured" output but without the journal parsing it properly - how is this a matter of taste?
You literally say *The structured logging info in journal mode is pretty hard to read for humans* so why would you want to output it unparseable by default?

omoerbeek:
I see your confusion. In both case structured logging infomration is stored.
The default is now to output all structured logging information in a text line containing a sequence of key=value pairs. I find that easy, as all information is there without the need to dig deeper with `journalctl`. This is also the same format as is logged to `syslogd` on other platforms or when not using `systemd`.

If we change this to journal format, only the `msg` part is shown by default by `journalctl`. It might be easier to read, but it also contains less information as the details are only shown if you specify extra `journalctl` arguments.

SjonHortensius:
I applaud the structured logging and unification across logging backends but I'm not sure why you prefer replacing easy to read journal output with hard to read output, while all it takes is a setting to make the journal parse the field, allowing the user to output it in any way they like (for example, you could use `journalctl --output json` to have 'all information there without the need to dig deeper')

I'd be interested in other people's opinions on this. Your comments seem pretty conflicting

> structured logging info in journal mode is pretty hard to read for humans

> I find that easy, as all information is there

omoerbeek:
You might want to read https://blog.powerdns.com/2022/10/03/structured-logging-in-powerdns-recursor/
It gives some background info.
When I introduced journal mode, it was the default. Within a few days I received complaints from a power user running the master branch, so I decided the keep the text line format as default. See #11705, #11706 and #11709

SjonHortensius:
Okay, I see your predicament. I did notice the issue you mention with a few other messages as well, where `msg` is no longer unique and thus less useful without the details. Personally I'd prefer the structured logging to provide extra information (when necessary) which is why I think you should only log `msg` but ensure important details aren't moved to 'invisible' fields. 

I think the current solution goes against the very idea it's supposed to implement but I'm fine with changing this flag myself. Maybe in the future the default can be changed, hopefully after the `msg` has a better balance of being informational on its own.

Thanks for the detailed explanation!

omoerbeek:
In general agreed; it will take some time to arrive there though.

PR #12175, opened 2022-11-07
rec: Get DS records with QM switched on.

### LINKED ISSUE #12160: rec: startup race can happen that provokes a.root-servers.net to go Bogus on statup
This is a return of #6138
I can provoke it of I only only use the `a` and `j` `root-servers` as these return `aa` answers for `DS`queries:

```
$ dig root-servers.net DS @j.root-servers.net +dnssec

; <<>> DiG 9.10.6 <<>> root-servers.net DS @j.root-servers.net +dnssec
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 51491
;; flags: qr aa rd; QUERY: 1, ANSWER: 0, AUTHORITY: 1, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags: do; udp: 1472
;; QUESTION SECTION:
;root-servers.net.		IN	DS

;; AUTHORITY SECTION:
root-servers.net.	3600000	IN	SOA	a.root-servers.net. nstld.verisign-grs.com. 2022102600 14400 7200 1209600 3600000

;; Query time: 173 msec
;; SERVER: 192.58.128.30#53(192.58.128.30)
;; WHEN: Thu Nov 03 10:40:17 CET 2022
;; MSG SIZE  rcvd: 105
```
`a.root-servers.net` sometimes does. So the set of NS that show the `aa` response is different from  three years ago.

During thread startup, the threads that handle queries start already doing so while the priming of the root in the handler thread is still underway.

Also, the initial priming does not enforce the the `net. NS` records to be retrieved by calling `primeRootNSZones`, as the housekeeping does.

These things together can cause an early `a.root-servers.net|A` query to find no `.net|A` records in the cache, hitting the #6138 scenario.

To provoke: use `dnssec=validate` and `qname-minimization=no`, restart a few times with dnsdist healthcheck to the recursor  active.

omoerbeek:
This avoid `a.root-servers.net` going `Bogus`, which can happen if the `.net NS` are not cached and we miss the cut.

Fixes #12160 and avoids the need to prime the `.net NS` records explicitly.

Draft for now.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/recursordist/rec-main.cc
+++ b/pdns/recursordist/rec-main.cc
@@ -2229,30 +2229,6 @@ static void houseKeeping(void*)
         if (res == 0) {
           // Success, go back to the defaut period
           rootUpdateTask.setPeriod(std::max(SyncRes::s_maxcachettl * 8 / 10, minRootRefreshInterval));
-          const string msg = "Exception while priming the root NS zones";
-          try {
-            primeRootNSZones(g_dnssecmode, 0);
-          }
-          catch (const std::exception& e) {
-            SLOG(g_log << Logger::Error << "Exception while priming the root NS zones: " << e.what() << endl,
-                 log->error(Logr::Error, e.what(), msg, "exception", Logging::Loggable("std::exception")));
-          }
-          catch (const PDNSException& e) {
-            SLOG(g_log << Logger::Error << "Exception while priming the root NS zones: " << e.reason << endl,
-                 log->error(Logr::Error, e.reason, msg, "exception", Logging::Loggable("PDNSException")));
-          }
-          catch (const ImmediateServFailException& e) {
-            SLOG(g_log << Logger::Error << "Exception while priming the root NS zones: " << e.reason << endl,
-                 log->error(Logr::Error, e.reason, msg, "exception", Logging::Loggable("ImmediateServFailException")));
-          }
-          catch (const PolicyHitException& e) {
-            SLOG(g_log << Logger::Error << "Policy hit while priming the root NS zones" << endl,
-                 log->info(Logr::Error, msg, "exception", Logging::Loggable("PolicyHitException")));
-          }
-          catch (...) {
-            SLOG(g_log << Logger::Error << "Exception while priming the root NS zones" << endl,
-                 log->info(Logr::Error, "Exception while priming the root NS zones"));
-          }
         }
         else {
           // On failure, go to the middle of the remaining period (initially 80% / 8 = 10%) and shorten the interval on each
--- a/pdns/recursordist/test-syncres_cc7.cc
+++ b/pdns/recursordist/test-syncres_cc7.cc
@@ -1393,10 +1393,19 @@ BOOST_AUTO_TEST_CASE(test_dnssec_bogus_nodata)
     if (type == QType::DS || type == QType::DNSKEY) {
       return genericDSAndDNSKEYHandler(res, domain, domain, type, keys);
     }
-    else {
-
-      setLWResult(res, 0, true, false, true);
-      return LWResult::Result::Success;
+    else if (type == QType::A) {
+      if (domain == DNSName("com.")) {
+        setLWResult(res, 0, true, false, true);
+        addRecordToLW(res, DNSName("com"), QType::SOA, "whatever.com. blah.com. 2017032301 10800 3600 604800 3600", DNSResourceRecord::AUTHORITY, 3600);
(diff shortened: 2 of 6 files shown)


omoerbeek:
```
  FAIL: test_Bogus (test_AggressiveNSECCache.AggressiveNSECCacheNSEC)
    File "/home/runner/work/pdns/pdns/regression-tests.recursor-dnssec/test_AggressiveNSECCache.py", line 171, in test_Bogus
      self.assertEqual(entries + 1, self.getMetric('aggressive-nsec-cache-entries'))
  AssertionError: 1 != 0
```
Now there's something fishy going on, as on master the `test_AggressiveNSECCache` only succeeds if all tests in it are run.
If I run only test_Bogus (which is normally run last) I see:

```
$ ./runtests test_AggressiveNSECCache:AggressiveNSECCacheNSEC.test_Bogus
...
test_Bogus (test_AggressiveNSECCache.AggressiveNSECCacheNSEC) ... FAIL

FAIL: test_Bogus (test_AggressiveNSECCache.AggressiveNSECCacheNSEC)
  File "/home/otto/pdns/regression-tests.recursor-dnssec/test_AggressiveNSECCache.py", line 171, in test_Bogus
    self.assertEqual(entries + 1, self.getMetric('aggressive-nsec-cache-entries'))
AssertionError: 2 != 1
```

omoerbeek:
On master, it's a race condition between the wiping of the cache in `test_AggressiveNSECCache::setUp` and the priming of the root. With that fixed i get consistent results.

rgacogne (commented):
I see why always using QM when looking for DS records solves the problem we have in #12160, and the code looks good. But I'm wondering if it makes sense to solve the problem that way, and what the cost is when QM is switched off for other queries: are we sending and caching a fair amount of queries that would not have been generated otherwise?.
I'm also a bit worried that there might other cases where we can get a child-size reply to a DS query and not notice (perhaps because some other issue made us disable QM?).

omoerbeek:
> I see why always using QM when looking for DS records solves the problem we have in #12160, and the code looks good. But I'm wondering if it makes sense to solve the problem that way, and what the cost is when QM is switched off for other queries: are we sending and caching a fair amount of queries that would not have been generated otherwise?. I'm also a bit worried that there might other cases where we can get a child-size reply to a DS query and not notice (perhaps because some other issue made us disable QM?).

As for the number of DS queries: if we have seen a delegation we have typically seen the corresponding DS already.
The QM step just makes sure we have seen the whole delegation chain. In a way the has similarity to the old way we computed the zone cuts. I do understand your worry this PR might degrade resolving the DS to that way.

I also agree that detecting the "this DS is coming from a child" is better, but is there actually a proper way to decide that?

rgacogne:
> As for the number of DS queries: if we have seen a delegation we have typically seen the corresponding DS already.
The QM step just makes sure we have seen the whole delegation chain. In a way the has similarity to the old way we computed the zone cuts. I do understand your worry this PR might degrade resolving the DS to that way.

Ah, was I mistaken in assuming that we would be using `A` queries instead of `DS` queries when QM is enabled?

> I also agree that detecting the "this DS is coming from a child" is better, but is there actually a proper way to decide that?

I think we could do that by looking at the SOA in the response, but I have not actually written that code and tried so.. :)

omoerbeek:
> Ah, was I mistaken in assuming that we would be using `A` queries instead of `DS` queries when QM is enabled?

No, we do use A queries when doing QM. But a DS record (and corresponding RRSIG) is typically seen when the answer to that A query is a delegation.

rgacogne:
Oh, you are completely right, we only do send these `A` queries if the answer is not already in cache, and most of the time it will be. OK, so we should not have a real performance impact. Given that we do not know of any actual case where we might have a problem with child-side DS answers, let's not make the code more complex until we actually see one.
So I'm OK with merging this PR once the tests are updated, unless you see a problem with the changes I proposed?

omoerbeek:
for some reason I cannot cherry pick 994d2e09f088f2908a887e2539f8abbd1e36d003

 Will do a diff and patch

rgacogne:
It's on https://github.com/rgacogne/pdns/tree/rec-getdsrecords-qm if that helps.

rgacogne (approved):

Issue #12142, opened 2022-10-31
dnsdist 1.7.2 and master do not build on amazonlinux 2


cmueller9999:
- Program: dnsdist 
 - Issue type: Bug report

### Short description

compilation error

### Environment

 - Operating system: Amazon Linux release 2 (Karoo)
 - Software version: 1.7.2
 - Software source: compiled yourself

### Steps to reproduce

./configure
make
1. 
2. 
3. 

### Actual behaviour

  CXX      dnsdist-lua-bindings.o
dnsdist-lua-bindings.cc: In lambda function:
dnsdist-lua-bindings.cc:438:92: error: ‘class boost::optional<bool>’ has no member named ‘value_or’; did you mean ‘get_value_or’?
       return std::make_shared<BPFFilter>(v4Config, v6Config, qnameConfig, format, external.value_or(false));
                                                                                            ^~~~~~~~
                                                                                            get_value_or
dnsdist-lua-bindings.cc: In lambda function:
dnsdist-lua-bindings.cc:470:72: error: ‘class boost::optional<short unsigned int>’ has no member named ‘value_or’; did you mean ‘get_value_or’?
           return bpf->block(qname, BPFFilter::MatchAction::Drop, qtype.value_or(255));
                                                                        ^~~~~~~~
                                                                        get_value_or
dnsdist-lua-bindings.cc:488:49: error: ‘class boost::optional<short unsigned int>’ has no member named ‘value_or’; did you mean ‘get_value_or’?
           return bpf->block(qname, match, qtype.value_or(255));
                                                 ^~~~~~~~
                                                 get_value_or
make[2]: *** [dnsdist-lua-bindings.o] Error 1
make[2]: Leaving directory `/root/dnsdist-1.7.2'
make[1]: *** [all-recursive] Error 1
make[1]: Leaving directory `/root/dnsdist-1.7.2'
make: *** [all] Error 2

phonedph1:
I got past these by using amazon linux 2 `gcc10` and building the latest version of `boost` (amazon linux 2 version is quite old). My build isn't done yet but I suspect you'd have success doing that.

Relevant: #9089

Habbie:
Our own amazon-2 package build (via `builder/build.sh`) is broken because of this since 053a020a8c8a23c9c7471a9d6e1161606738b6a6

Habbie:
[...]

PR #12177, opened 2022-11-08
dnsdist: Fix building with boost < 1.56

rgacogne:
### Short description

`boost::optional::value_or()` has been introduced in 1.56 and we only require 1.53, so stop using it.

Hopefully fixes the `master` part of #12142.

### Checklist

I have:
- [x] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdist-lua-bindings.cc
+++ b/pdns/dnsdist-lua-bindings.cc
@@ -538,12 +538,12 @@ void setupLuaBindings(LuaContext& luaCtx, bool client)
     default:
       throw std::runtime_error("Unsupported action for BPFFilter::block");
     }
-    return bpf->addRangeRule(Netmask(range), force.value_or(false), match);
+    return bpf->addRangeRule(Netmask(range), force ? *force : false, match);
   });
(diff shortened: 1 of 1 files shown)


Habbie:
I see your portable construct, and I like it, but I also note we use `get_value_or` in other places. However, that one has been deprecated in Boost, so we shouldn't switch to that.

#12183 backports the part of this PR that applies, and adds another fix in a place that master apparently does not have any more.

Habbie:
> Hopefully fixes the `master` part of #12142.

There are at least two other problems on `master` for amazon-2 - one being the specfile now pulling in Boost 1.69 which doesn't work there, and one being the noncopyable include that somebody else also noticed. I hope to fix those two things soon.

Habbie (approved):

rgacogne:
> I see your portable construct, and I like it, but I also note we use `get_value_or` in other places. However, that one has been deprecated in Boost, so we shouldn't switch to that.

I went for the minimal fix in this PR but I agree we should also stop using `get_value_or` as it might get removed one day or another. I'll open an issue for that.

PR #12193, opened 2022-11-15
Introducing TCounters

omoerbeek:
This is a mostly lockless (and not using atomics) way to keep track of counters and other metrics.

Atomic value are more expensive than you would think (especially if your platform has no native atomic support for your data type), and using locking all the time for often updated counters is very expensive as well.

The idea for `TCounters` is based on

https://github.com/ahupowerdns/mcounter

But addresses the issues raised in
https://github.com/ahupowerdns/mcounter/issues/3

Templates are used, the application has to provide a specific class to hold the values and enums to index these values.  The application specific class also has to provide a `merge()` method to merge two instances of the application specific data.  For counters that is simple: just add them. Averages (or histograms) requires a bit more work. This is demonstrated in `rec-tcounters.{cc,hh}`

From `tcounters.hh`:
```
// We keep three sets of related counters:
//
// 1. The current counters (thread-local, updated individually by thread code very often)
// 2. The snapshot counters (thread local, updated by thread code in one single mutex protected copy)
// 3. The history counters (global) to keep track of the counters of deleted threads

// We have two main clasess: one that holds the thread local counters
// (both current and snapshot ones) and one that aggregates the
// values for all threads adn keeps the history counters.

// The thread local current counters are the ones updated by
// performance critical code.  Every once in a while, all values in the
// current counters are copied to the snapshot thread local copies in
// a thread safe way.

// The snapshot counters are aggregated by the GlobalCounters
// class, as these can be accessed safely from multiple threads.

// Make sure to call the thread local tlocal.updatesAtomics() once
// in a while. This will fill the snapshot values for that thread if some
// time has passed since the last snap update.

// To fet aggregate value call globals.sum(counter1) or
// globals.avg(counter2), or any aggreggation function.  If multiple
// counters need to be collected in a consistent way:
// auto data = globals.aggregatedSnap();
//
// Note that the aggregate values can mix somewhat older thread-local
// with newer thread-local info from another thread. So it is possible
// to see the following:
//
// If thread T1 increments "received" and then passes the packet to
// thread T2 that increments "processed", it may happen that the value
// of "processed" observed by sum() is higher than "received", as T1
// might not have called updateSnap() yet while T2 did.  To avoid this
// inconsistentcy, be careful to update related counters in a single
// thread only.
```
The snapshot approach was suggested by @wojas.

This PR de demonstrates `TCounters` for a most Recursor metrics: simple counters and double typed average values. For the latter weights are kept, so that the average of averages can be computed in a proper way. Also hostograms in two variaties are done.

To work around timing specific issue in the regression tests (which in some cases read metrics) a flag has been introduced that forces snaps and avoids the root refresh to run. The latter increase outgoing queries that break some of the test that inspect the related outgoing metrics.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [X] documented the code

### DIFF
--- a/pdns/histogram.hh
+++ b/pdns/histogram.hh
@@ -38,9 +38,18 @@ struct Bucket
 {
   Bucket(std::string name, uint64_t boundary, uint64_t val) :
     d_name(std::move(name)), d_boundary(boundary), d_count(val) {}
-  std::string d_name;
-  uint64_t d_boundary{0};
+  const std::string d_name;
+  const uint64_t d_boundary;
   mutable uint64_t d_count{0};
+
+  Bucket(const Bucket&) = default;
+  Bucket& operator=(const Bucket& rhs)
+  {
+    assert(d_name == rhs.d_name);
+    assert(d_boundary == rhs.d_boundary);
+    d_count = rhs.d_count;
+    return *this;
+  }
 };
 
 struct AtomicBucket
--- a/pdns/histogram.hh
+++ b/pdns/histogram.hh
@@ -21,6 +21,7 @@
  */
 #pragma once
 
+#include <cassert>
 #include <algorithm>
 #include <limits>
 #include <stdexcept>
@@ -35,8 +36,10 @@ namespace pdns
 // By convention, we are using microsecond units
 struct Bucket
 {
-  const std::string d_name;
-  const uint64_t d_boundary{0};
+  Bucket(std::string name, uint64_t boundary, uint64_t val) :
+    d_name(std::move(name)), d_boundary(boundary), d_count(val) {}
+  std::string d_name;
+  uint64_t d_boundary{0};
   mutable uint64_t d_count{0};
 };
 
@@ -147,9 +150,22 @@ public:
     d_sum += d;
   }
 
+  BaseHistogram& operator+=(const BaseHistogram& rhs)
+  {
+    assert(d_name == rhs.d_name);
+    assert(d_buckets.size() == rhs.d_buckets.size());
+    for (size_t bucket = 0; bucket < d_buckets.size(); ++bucket) {
+      assert(d_buckets[bucket].d_name == rhs.d_buckets[bucket].d_name);
+      assert(d_buckets[bucket].d_boundary == rhs.d_buckets[bucket].d_boundary);
+      d_buckets[bucket].d_count += rhs.d_buckets[bucket].d_count;
+    }
+    d_sum += rhs.d_sum;
+    return *this;
+  }
+
 private:
   std::vector<B> d_buckets;
-  const std::string d_name;
+  std::string d_name;
   mutable SumType d_sum{0};
 
   std::vector<uint64_t> to125(uint64_t start, int num)
--- a/pdns/recursordist/rec-main.cc
+++ b/pdns/recursordist/rec-main.cc
@@ -62,6 +62,7 @@ thread_local FrameStreamServersInfo t_nodFrameStreamServersInfo;
 string g_programname = "pdns_recursor";
 string g_pidfname;
 RecursorControlChannel g_rcc; // only active in the handler thread
+bool g_regressionTestMode;
 
 #ifdef NOD_ENABLED
 bool g_nodEnabled;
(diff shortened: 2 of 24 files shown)


pdns/tcounters.hh
```diff
@@ -0,0 +1,228 @@
...
+{
+
+// We have two main clasess: one that holds the thread local counters
+// (both current and snapshot ones) and one that aggregates the
+// values for all threads.
+
+// The thread local current counters are the ones updates by
+// performance critcal code.  Every one in a while, the values in the
+// current counters are copied to the sanpshots thread local copies in
+// a thread safe way.
```
rgacogne, line 42:
```suggestion
// The thread local current counters are the ones updated by
// performance critical code.  Every once in a while, the values in the
// current counters are copied to the snapshot thread local copies in
// a thread safe way.
```

pdns/tcounters.hh
```diff
@@ -0,0 +1,228 @@
...
+
+// The snapshot counters are aggregated by the GlobalCounters
+// class, as these can be accessed safely from multiple threads.
+
+// Make sure to call the thread specific tlocal.updatesAtomics() once
+// in a while.  This will fil the snapshot values for that thread.
```
rgacogne, line 48:
```suggestion
// Make sure to call the thread specific tlocal.updatesAtomics() once
// in a while. This will fill the snapshot values for that thread.
```

pdns/tcounters.hh
```diff
@@ -0,0 +1,228 @@
...
+{
+public:
+  // Register a thread local set of values
+  void subscribe(TLocalCounters<Counters>* ptr)
+  {
+    auto lock = d_guarded.lock();
+    lock->d_instances.emplace(ptr);
+  }
+
+  // Unregisterm, typically done when a thread exits
```
rgacogne, line 78:
```suggestion
  // Unregister, typically done when a thread exits
```

pdns/tcounters.hh
```diff
@@ -0,0 +1,228 @@
...
+  void unsubscribe(TLocalCounters<Counters>* ptr, const Counters& data)
+  {
+    auto lock = d_guarded.lock();
+    lock->d_instances.erase(ptr);
+    lock->d_history.merge(data);
+  }
+
+  // Two ways of computing aggregated values for a specific counter: simple additions of all thread data, or taking weighted avarages into account
```
rgacogne, line 86:
```suggestion
  // Two ways of computing aggregated values for a specific counter: simple additions of all thread data, or taking weighted averages into account
```

pdns/tcounters.hh
```diff
@@ -0,0 +1,228 @@
...
+  template <typename Enum>
+  auto sum(Enum index);
+  template <typename Enum>
+  auto avg(Enum index);
+
+  // Aggragte all coounter data for all threads
```
rgacogne, line 92:
```suggestion
  // Aggregate all counter data for all threads
```

rgacogne (commented):
Looks good at first glance, nice work! I'll do a better review later.
Out of curiosity, any idea what the performance impact is?

omoerbeek:
Using the test code I just committed:
MacOS M1 Max:
```
mbp:recursordist otto$ ./a.out 8 100000000
Starting 8 threads doing 100000000 iterations using atomics
Sum is 800000000 elapsed is 34.8775
Now doing the same with tcounters
Sum is 800000000 elapsed is 0.367856
```
Random Debian amd64 machine:
```
otto@pepper:~/pdns/pdns/recursordist$ ./a.out 8 100000000
Starting 8 threads doing 100000000 iterations using atomics
Sum is 800000000 elapsed is 24.1347
Now doing the same with tcounters
Sum is 800000000 elapsed is 0.393679
```
Some ubuntu vm:
```
otto@ubuntu-s-2vcpu-4gb-amd-ams3-01:~/pdns/pdns/recursordist$ ./a.out 8 100000000
Starting 8 threads doing 100000000 iterations using atomics
Sum is 800000000 elapsed is 14.141
Now doing the same with tcounters
Sum is 800000000 elapsed is 0.988841
```
OpenBSD Apple M1:
```
[otto@macmini:7]$ ./a.out 8 100000000
Starting 8 threads doing 100000000 iterations using atomics
Sum is 800000000 elapsed is 19.7271
Now doing the same with tcounters
Sum is 800000000 elapsed is 3.15709
```
OpenBSD amd64:
```
[otto@lou:2]$ ./a.out 8 100000000
Starting 8 threads doing 100000000 iterations using atomics
Sum is 800000000 elapsed is 6.65176
Now doing the same with tcounters
Sum is 800000000 elapsed is 3.1043
```

Conclusion: unless my measurement is wrong, tcounters are very good. OpenBSD is slow but tcounters are still faster on it.

omoerbeek:
Found an issue. I'm not always reading from the snaps when I need to.

omoerbeek:
Rebased to fix conflicts

pdns/recursordist/rec-tcounters.hh
```diff
@@ -0,0 +1,219 @@
...
+    for (auto& elem : auth.rcodeCounters) {
+      elem = 0;
+    }
+  }
+
+  // Merge a set of counters into an existing set of counters. For simple counters, that will be additions
+  // for averages, we should take the weights into account. Histograms need to sum all individual counts.
+  Counters& merge(const Counters& data);
+
+  // The following accessors select the rightcounter type based on the index type
```
rgacogne, line 190:
```suggestion
  // The following accessors select the right counter type based on the index type
```

pdns/recursordist/rec-main.cc
```diff
@@ -2587,6 +2601,8 @@ int main(int argc, char** argv)
 #else
     ::arg().set("stack-size", "stack size per mthread") = "200000";
 #endif
+    // This mode forces metrics snap updates and dsiable root-refresh, to get consistent counters
```
rgacogne, line 2604:
```suggestion
    // This mode forces metrics snap updates and disable root-refresh, to get consistent counters
```

pdns/histogram.hh
```diff
@@ -35,8 +36,10 @@ namespace pdns
 // By convention, we are using microsecond units
 struct Bucket
 {
-  const std::string d_name;
-  const uint64_t d_boundary{0};
+  Bucket(std::string name, uint64_t boundary, uint64_t val) :
+    d_name(std::move(name)), d_boundary(boundary), d_count(val) {}
+  std::string d_name;
+  uint64_t d_boundary{0};
```
rgacogne, line 42: Looking at the diff, it seems that `d_name` and `d_boundary` could still be `const`?

omoerbeek: I had some issues there with an earlier version, will revisit.

Issue #12205, opened 2022-11-17
auth: 'Consumer' zone created via API gets synthesized SOA record


kpfleming:
- Program: Authoritative
 - Issue type: Bug report

### Short description
Creation of a 'Consumer' catalog zone via the API results in an AXFR failure because the zone receives an synthesized SOA RR which will very likely have a serial higher than the primary's SOA serial.

### Environment
 - Operating system: Debian Bookworm
 - Software version: 4.7.2
 - Software source: self-compiled using PowerDNS builder

### Steps to reproduce
1. Use API to create a 'Consumer' zone, without specifying any SOA details (as none should be necessary).
2. Run 'pdnsutil list-zone <zone>'.
3. See SOA record in output which does not match SOA record from primary (also note that an error is emitted in the log/journal because the local SOA serial is higher than the primary SOA serial).

### Expected behaviour
SOA record should not be synthesised, as a 'Consumer' zone is similar in behavior to a 'Slave' zone.

### Actual behaviour
Zone AXFR fails due to SOA serial being incorrect.

### Other information
I've found the code in `ws-auth.cc` which is responsible for this; it only checks the zone kind for 'Slave' and does not check for 'Consumer'.

I'll open two PRs to address this:

* For `rel/auth-4.7.x` I'll just patch ws-auth.cc to have the correct zone kind tests.
* For `master` I'll lift `DomainKind` and its conversion/classification functions out of `DomainInfo` into a class-wrapped-enum, and then update all of the users to use the new class. This should help avoid this type of problem in the future as direct comparisons against the zone 'kind' would only be needed in very specific circumstances.

Habbie:
Fixed in #12285 (and #12291)

PR #12212, opened 2022-11-18
Convert DomainInfo::DomainKind to ZoneKind class-wrapped-enum

kpfleming:
### Short description
This patch 'lifts' DomainKind out of DomainInfo (so that it can be used when no DomainInfo object is available), renames it to ZoneKind, and moves its conversion and classification functions into the new class as member functions.

This was inspired by #12205, where code that did not have a DomainInfo object was unable to make use of the isSecondary() classification function, which resulted in a logic error for zones of type 'Consumer'.

### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code


mind04:
I see no advantage in in moving DomainKind outside DomainInfo. There is zero use of kind unrelated to DomainInfo so why is this exactly better. And is it worth making backports complicated for a long time?

kpfleming:
`ws-auth.cc` uses DomainKind outside of DomainInfo objects, which is what prompted me to consider doing this.

zeha:
General approach seems worthwhile to me. LMDB not explicitly defining the encoding of ZoneKind for its data files seems like a bug to me.

(ws-auth.cc is conflicted)

PR #12213, opened 2022-11-18
Ensure that 'Consumer' zones do not get API-synthesized SOA records

### LINKED ISSUE #12205: auth: 'Consumer' zone created via API gets synthesized SOA record

kpfleming:
### Short description
When a 'Consumer' zone is created via the API, it should get the same treatment as a 'Slave' zone, since it is also a 'secondary' zone kind.

Closes #12205.

### Checklist

I have:
- [x] compiled this code
- [ ] tested this code
- [ ] documented the code
- [ ]  checked that this code was merged to master (#12212 includes this fix for master)

### DIFF
--- a/pdns/ws-auth.cc
+++ b/pdns/ws-auth.cc
@@ -1764,7 +1764,7 @@ static void apiServerZones(HttpRequest* req, HttpResponse* resp) {
     autorr.auth = true;
     autorr.ttl = ::arg().asNum("default-ttl");
 
-    if (!have_soa && zonekind != DomainInfo::Slave) {
+    if (!have_soa && zonekind != DomainInfo::Slave && zonekind != DomainInfo::Consumer) {
       // synthesize a SOA record so the zone "really" exists
       string soa = ::arg()["default-soa-content"];
       boost::replace_all(soa, "@", zonename.toStringNoDot());


Habbie:
replaced by #12291

PR #12237, opened 2022-11-25
dnsdist: Ignore unclean TLS session shutdown

### LINKED ISSUE #12236: dnsdist: TestOutgoingTLSOpenSSL broken with OpenSSL 3.0.x
- Program: dnsdist 
 - Issue type: Bug report

### Short description

Our TestOutgoingTLSOpenSSL regression test check whether we correctly resumed the TLS connection with the server. Unfortunately OpenSSL 3.0.x invalidates the ticket when the remote end closed the connection without sending a `close_notify` first, and thus our test is currently broken.
Grab some popcorn and head to https://github.com/openssl/openssl/issues/11378 for the details.

rgacogne:
### Short description

OpenSSL 3.0 "helpfully" treats an unclean TLS session shutdown as an error, flooding our logs and killing TLS session resumption. We do not care about a possible "truncation attack" since we already know how many bytes we are supposed to get, so we can ignore this.
Fixes #12236.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


rgacogne:
For what is is worth, other projects have applied the same fix:
- https://github.com/php/php-src/pull/8558/files
- https://github.com/python/cpython/commit/6f37ebc61e9e0d13bcb1a2ddb7fc9723c04b6372
- https://hg.nginx.org/nginx/rev/1a03af395f44

omoerbeek (approved):

PR #12248, opened 2022-11-29
systemd: Add "After" dependency on time-sync.target

### LINKED ISSUE #11153: Consider adding 'After=time-sync.target' to systemd service unit files
 - Issue type: Feature request

### Short description
Ensure that the system's clock is synchronized before starting PowerDNS services.

### Usecase
Allow PowerDNS services to operate correctly on machines where the hardware clock is faulty (or non-existent). Failure to have a proper real-time clock on the machine can result in hard-to-diagnose DNS failures when DNSSEC is in use.

### Description
See https://www.freedesktop.org/software/systemd/man/systemd.special.html#time-sync.target for a description of this target.

Adding 'After=time-sync.target' to the service unit files will not have any immediate effect, unless the system administrator has *also* enabled some sort of service which has a 'Before' dependency on that target; there are none of those services in a default configuration. systemd itself adds such an 'After' dependency to all timer units with `OnCalendar` settings, and to some other units.

However, in a system which makes use of `systemd-timesyncd`, the admin can enable `systemd-time-wait-sync.service`, which has such a dependency. The result would be that none of the PowerDNS services on the system would be started until after `systemd-timesyncd` is certain that the system's clock is synchronized with some external source.

A similar configuration can be achieved on systems which use `ntpd`, `chronyd`, or other time synchronization tools.

kpfleming:
### Short description
Adds 'After=time-sync.target' to the service unit files; this will not have any immediate effect, unless the system administrator has also enabled a unit which has a 'Before' dependency on that target; there are none of those units enabled in a default configuration in common distributions. systemd itself adds such an 'After' dependency to all timer units with OnCalendar settings, and to some other units.

However, in a system which makes use of systemd-timesyncd, the admin can enable systemd-time-wait-sync.service, which has such a dependency. The result would be that none of the PowerDNS services on the system would be started until after systemd-timesyncd is certain that the system's clock is synchronized with the configured external source.

A similar configuration can be achieved on systems which use ntpd, chronyd, or other time synchronization tools.

Closes #11153.

### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdistdist/dnsdist.service.in
+++ b/pdns/dnsdistdist/dnsdist.service.in
@@ -3,7 +3,7 @@ Description=DNS Loadbalancer
 Documentation=man:dnsdist(1)
 Documentation=https://dnsdist.org
 Wants=network-online.target
-After=network-online.target
+After=network-online.target time-sync.target
 
 [Service]
 ExecStartPre=@bindir@/dnsdist --check-config
--- a/pdns/ixfrdist.service.in
+++ b/pdns/ixfrdist.service.in
@@ -4,7 +4,7 @@ Documentation=man:ixfrdist(1)
 Documentation=man:ixfrdist.yml(5)
 Documentation=https://doc.powerdns.com
 Wants=network-online.target
-After=network-online.target
+After=network-online.target time-sync.target
 
 [Service]
 Type=simple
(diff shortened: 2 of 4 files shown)


omoerbeek (approved):
This looks fine and safe to me.

rgacogne (approved):

rgacogne:
CI failures are unrelated to this PR.

PR #12249, opened 2022-11-29
Upgrade check-spelling to v0.0.21

jsoref:
### Short description

Upgrades to v0.0.21

https://github.com/check-spelling/check-spelling/security/advisories/GHSA-p8r9-69g4-jwqq

* Permanently stubs `spelling2.yml` (as `spelling.yml` is stubbed)
* Switches to using sarif reporting using security-events (the model for this is frustrating and almost argues for using a distinct task just as the comment-push job is distinct) for PRs
* Introduces step summaries (which are longer than comments, but not quite as easy to see)
* Allows users to talk to the bot in their own forks (it will walk them through the process)
* `candidate.patterns` automatically suggest entries for `patterns.txt`
* `line_forbidden.patterns` flags patterns that shouldn't be used
* [`act`](https://github.com/nektos/act) should usually work

* For more details, see the release notes for [v0.0.20](https://github.com/check-spelling/check-spelling/releases/tag/v0.0.20) and [v0.0.21](https://github.com/check-spelling/check-spelling/releases/tag/v0.0.21)

### Checklist

I have:
- [ ] compiled this code
- [x] tested this code
- [ ] documented the code


jsoref:
The failed commit https://github.com/PowerDNS/pdns/pull/12249/commits/48424f864b1072d8748d18fc021c232ab787c56a has two complaints:
> Warning: .github/workflows/spelling3.yml:83:9 ... 21, Warning - use_sarif needs security-events: write - unsupported configuration (unsupported-configuration)
Error: https://github.com/PowerDNS/pdns/pull/12249#:13:30 ... 36, Error - `nektos` is not a recognized word. (unrecognized-spelling)

### `nektos`

As of https://github.com/PowerDNS/pdns/pull/12249/commits/48424f864b1072d8748d18fc021c232ab787c56a, check-spelling is configured to check PR titles and descriptions. This can be turned off using:
```diff
diff --git a/.github/workflows/spelling3.yml b/.github/workflows/spelling3.yml
index 8a9c6d118..f7c133212 100644
--- a/.github/workflows/spelling3.yml
+++ b/.github/workflows/spelling3.yml
@@ -83,3 +83,2 @@ jobs:
         use_sarif: 1
-        check_commit_messages: title description
         extra_dictionaries:
```

Or, I could avoid including unrecognized words in the PR title / description. I'm leaning to the former.

### `sarif`

sarif isn't working because it's configured as a `pull_request` and the rules for PRs from forks: https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token

To limit sarif to the same repository, we could do:
```diff
diff --git a/.github/workflows/spelling3.yml b/.github/workflows/spelling3.yml
index f7c133212..e8f0f9a7c 100644
--- a/.github/workflows/spelling3.yml
+++ b/.github/workflows/spelling3.yml
@@ -82,3 +82,3 @@ jobs:
         experimental_apply_changes_via_bot: ${{ github.repository_owner != 'powerdns' && 1 || 0 }}
-        use_sarif: 1
+        use_sarif: ${{ github.event.pull_request.head.repo.full_name == github.repository && 1 }}
         extra_dictionaries:
```

Note that as-is, it isn't fatal (based on a change I made in my sprint before the release). But, it also isn't particularly helpful.

Alternatively, we could go back to `pull_request_target` which I believe leads us back to having write permissions.

Note that either way, job step summaries mean users can see a pretty report: https://github.com/PowerDNS/pdns/actions/runs/3574678549#summary-9771582537 -- it just isn't particularly easy to get to it...

.github/workflows/spelling3.yml
```diff
@@ -0,0 +1,140 @@
...
+        experimental_apply_changes_via_bot: ${{ github.repository_owner != 'powerdns' && 1 || 0 }}
+        checkout: true
+        spell_check_this: check-spelling/spell-check-this@prerelease
+        task: ${{ needs.spelling.outputs.followup }}
+
+  update:
+    name: Update PR
+    permissions:
+      contents: write
+      pull-requests: write
```
Habbie, line 121: I'd rather not have this.

.github/workflows/spelling3.yml
```diff
@@ -0,0 +1,140 @@
...
+      - 'opened'
+      - 'reopened'
+      - 'synchronize'
+  issue_comment:
+    types: [created]
+
+jobs:
+  spelling:
+    name: Spell checking
+    permissions:
```
Habbie, line 59: please add comments explaining why each permission is needed

Habbie (changes requested):

.github/workflows/spelling3.yml
```diff
@@ -0,0 +1,140 @@
...
+      contents: read
```
jsoref, line 60: This is to allow checking out the code

.github/workflows/spelling3.yml
```diff
@@ -0,0 +1,140 @@
...
+      pull-requests: read
```
jsoref, line 61: This is to check to see if there's a corresponding open pull request, in which case, `suppress_push_for_open_pull_request: 1` in the `push` event will cause the workflow to try to stop.

Actually, this pair could be dropped because the workflow as configured here will only comment on the PR and not the push, and thus there won't be a doubled comment, so there's nothing significant to suppress.

.github/workflows/spelling3.yml
```diff
@@ -0,0 +1,140 @@
...
+      actions: read
```
jsoref, line 62: If the workflow had `actions: write` it could properly kill itself (see the previous line).

With `actions: read`, it can slightly more easily find its own filename when it wants to report about problems in the running workflow (github does *not* make it easy to identify the running workflow -- there's a theoretical variable [`GITHUB_WORKFLOW`](https://docs.github.com/en/actions/learn-github-actions/environment-variables), but it only rarely has a path and mostly has a pretty name...) Otherwise, it has to search through each file in `.github/workflows/*.{yml,yaml}` trying to find itself and then try to guess which job in the workflow is running (it isn't particularly good at that).

.github/workflows/spelling3.yml
```diff
@@ -0,0 +1,140 @@
...
+      security-events: write
```
jsoref, line 63: This is for `use_sarif: 1`

.github/workflows/spelling3.yml
```diff
@@ -0,0 +1,140 @@
...
+      id: spelling
+      uses: check-spelling/check-spelling@v0.0.21
+      with:
+        config: .github/actions/spell-check
+        suppress_push_for_open_pull_request: 1
+        checkout: true
+        spell_check_this: check-spelling/spell-check-this@prerelease
+        post_comment: 0
+        experimental_apply_changes_via_bot: ${{ github.repository_owner != 'powerdns' && 1 || 0 }}
+        use_sarif: 1
```
jsoref, line 83: As noted in https://github.com/PowerDNS/pdns/pull/12249#issuecomment-1330766532, I _think_ we could set it to check to see if the request is from the same repo, but, otherwise `use_sarif` won't work w/o `pull_request_target` (which as we know is dangerous in general).

https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflows-in-forked-repositories
is trying to say that pull requests from forks get at most read permissions (the text is improperly worded and can't mean what it says, I'll file a bug/PR).

https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token
Is clearer about `Maximum access by forked repos`. There's a missing asterisk here (see below).

https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository#enabling-workflows-for-private-repository-forks
This is the asterisk: for *private* repositories (and this repository isn't private, and thus it isn't relevant), a repository / org (or presumably enterprise?) can allow passing write tokens to PRs from forks.

.github/workflows/spelling3.yml
```diff
@@ -0,0 +1,140 @@
...
+      contents: write
+      pull-requests: write
+      actions: read
+    runs-on: ubuntu-latest
+    if: ${{
+        github.repository_owner != 'powerdns' &&
+        github.event_name == 'issue_comment' &&
+        github.event.issue.pull_request &&
+        contains(github.event.comment.body, '@check-spelling-bot apply')
+      }}
```
jsoref, line 129: This job only runs in forks. I suppose it could be split into a distinct workflow so it could be disabled individually.

I'll just remove it.

.github/workflows/spelling3.yml
jsoref: See the related if.

.github/workflows/spelling3.yml
jsoref: In short, this is working around https://github.com/actions/runner/issues/853

.github/workflows/spelling3.yml
jsoref: I think it's probably only necessary for private repositories (we use check-spelling in quite a few of those, where it definitely matters)

.github/workflows/spelling3.yml
jsoref: similarly, I suspect this isn't strictly necessary for public repositories, although at worst it'd be a no-op otherwise.

jsoref (commented):
I could add comments, or we could try w/o. I'd kinda rather leave them present since it means that the workflow is more likely to work if someone decides to copy+paste it into a private repo (e.g. if there's a security release and someone wants to use CI to make sure things don't break).

.github/workflows/spelling3.yml
Habbie: thank you for the explanations. Now, can you please add comments to the yaml? :)

Habbie (changes requested):

.github/actions/spell-check/patterns.txt
```diff
@@ -22,6 +23,9 @@ msdn\.microsoft\.com/(?:[^/]+/|)library/\S*\.aspx?
 \b[0-9A-F]{8,}\b
 \bcommit [0-9a-f]+\b
 
+# HMAC
+hmac-sha512(:\w+|\.|)\s+[-a-zA-Z=;:/0-9+]*
```
jsoref, line 27: I just added this as some new items in `master` trip on this pattern

.github/workflows/spelling3.yml
```diff
@@ -0,0 +1,97 @@
...
+    steps:
+    - name: check-spelling
+      id: spelling
+      uses: check-spelling/check-spelling@v0.0.21
+      with:
+        config: .github/actions/spell-check
+        checkout: true
+        spell_check_this: check-spelling/spell-check-this@prerelease
+        post_comment: 0
+        use_sarif: ${{ (!github.event.pull_request || (github.event.pull_request.head.repo.full_name == github.repository)) && 1 }}
```
jsoref, line 84: This line is a bit more complicated than before.

There are three cases:

1. It's a push and not a pull request in which case using sarif is fine (this is new)
2. It's a pull request and it's from the same repo in which case using sarif is fine
3. It's a pull request and it's from another repo in which case using sarif won't work

jsoref: I'm using the SARIF flag to suppress the warning to annotation reporting that GitHub normally does – because GitHub will collate the SARIF report to the PR and add annotations using it.

If the workflow doesn't have access to use the SARIF reporting service, then if it did this, you'd end up with no annotations at all.

In prerelease, I've added a details section in the step summary which provides links to individual hits, which isn't precisely an alternative to the annotation, but at least makes it easy to get to a line...

I'm still tinkering with these behaviors, I may decouple the SARIF report from suppressing the other annotations.

.github/workflows/spelling3.yml
```diff
@@ -0,0 +1,97 @@
+# spelling.yml is blocked per https://github.com/check-spelling/check-spelling/security/advisories/GHSA-g86g-chm8-7r2p
+# spelling2.yml is blocked per https://github.com/check-spelling/check-spelling/security/advisories/GHSA-p8r9-69g4-jwqq
+name: Spell checking
+
+# Comment management is handled through a secondary job, for details see:
+# https://github.com/check-spelling/check-spelling/wiki/Feature%3A-Restricted-Permissions
+#
+# `jobs.comment-push` runs when a push is made to a repository and the `jobs.spelling` job needs to make a comment
```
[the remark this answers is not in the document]
jsoref: I can drop as much as you like. It's from my template to explain things you can do, but, we've done things.

PR #12260, opened 2022-12-02
auth: Properly encode json strings containing binary data

### LINKED ISSUE #9349: Bad json for binary log messages in statistics API response
- Program: Authoritative
 - Issue type: Bug report

### Short description
If there are binary log messages (e.g. from attempted hacking of the webserver) then the response from the statistics API endpoint can return invalid json.

### Environment

 - Operating system: Debian GNU/Linux 10 (buster)
 - Software version: 4.3.0-1pdns.buster
 - Software source: PowerDNS repository buster-auth-43

### Steps to reproduce
I'm not sure exactly how to generate a bad log message but here's an example of one as sent in the API response where the illegal UTF-8 bytes have been replaced by <0xHH>

```
{
      "value" : "1",
      "name" : "[webserver] 8ef848a1-570b-4b1f-aa25-eb7f59f4c63f 192.241.237.178:52962 \"\u0016\u0003\u0001\u0000<0x8b>\u0001\u0000\u0000<0x87>\u0003\u0003<0xe2>Q<0x8c><0x94>[<0xf9><0xa8>K?^<0xe8>\u0010Y<0xc7><0xbe>S\u0016RK<0xc5>D^\u0004_<0x9f>\u00056\u001b<0xb1><0xe7><0xa3><0x9f>\u0000\u0000\u001a<0xc0>/<0xc0>+<0xc0>\u0011<0xc0>\u0007<0xc0>\u0013<0xc0>  HTTP/0.9\" 400 143"
}
```

### Expected behaviour
The API response should be valid JSON (i.e. valid UTF-8)

This would probably involve some kind of escaping of binary blobs.

### Actual behaviour
The code calling the statistics API json (specifically [PowerDNS-Admin](https://github.com/ngoduykhanh/PowerDNS-Admin/blob/94da9198c0aae382854a96d8537aec78fe94ee7b/powerdnsadmin/models/server.py#L51-L57)) dies while trying to decode the response.

omoerbeek:
Mostly auth, as it has code to retrieve log lines via REST API. But other products could be affected as well.

The existing json11 code assumes the strings are already valid UTF8.

Fixes #9349

Also url encode `path` in log lines, as it trips pytest:
```
Running tests...

$ 'pytest' '--junitxml=pytest.xml' '-v'
==STDOUT===

==STDERRR===
  File "/home/otto/pdns/regression-tests.api/runtests.py", line 304, in <module>
    print(serverproc.stderr.read())
  File "/usr/lib/python3.9/codecs.py", line 322, in decode
    (result, consumed) = self._buffer_decode(data, self.errors, final)
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xeb in position 4304: invalid continuation byte
```
There might be more places where this is needed.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [X] documented the code

### DIFF
--- a/ext/json11/json11.cpp
+++ b/ext/json11/json11.cpp
@@ -93,18 +93,10 @@ static void dump(const string &value, string &out) {
             out += "\\r";
         } else if (ch == '\t') {
             out += "\\t";
-        } else if (static_cast<uint8_t>(ch) <= 0x1f) {
+        } else if (static_cast<uint8_t>(ch) <= 0x1f || static_cast<uint8_t>(ch) >= 0x7f) {
             char buf[8];
             snprintf(buf, sizeof buf, "\\u%04x", ch);
             out += buf;
-        } else if (static_cast<uint8_t>(ch) == 0xe2 && static_cast<uint8_t>(value[i+1]) == 0x80
-                   && static_cast<uint8_t>(value[i+2]) == 0xa8) {
-            out += "\\u2028";
-            i += 2;
-        } else if (static_cast<uint8_t>(ch) == 0xe2 && static_cast<uint8_t>(value[i+1]) == 0x80
-                   && static_cast<uint8_t>(value[i+2]) == 0xa9) {
-            out += "\\u2029";
-            i += 2;
         } else {
             out += ch;
         }
--- a/pdns/webserver.cc
+++ b/pdns/webserver.cc
@@ -535,7 +535,7 @@ void WebServer::serveConnection(const std::shared_ptr<Socket>& client) const {
   }
 
   if (d_loglevel >= WebServer::LogLevel::Normal) {
-    SLOG(g_log<<Logger::Notice<<logprefix<<remote<<" \""<<req.method<<" "<<req.url.path<<" HTTP/"<<req.versionStr(req.version)<<"\" "<<resp.status<<" "<<reply.size()<<endl,
+    SLOG(g_log<<Logger::Notice<<logprefix<<remote<<" \""<<req.method<<" "<<YaHTTP::Utility::encodeURL(req.url.path)<<" HTTP/"<<req.versionStr(req.version)<<"\" "<<resp.status<<" "<<reply.size()<<endl,
          d_slog->info(Logr::Info, "Request", "remote", Logging::Loggable(remote), "method", Logging::Loggable(req.method),
                       "urlpath", Logging::Loggable(req.url.path), "HTTPVersion", Logging::Loggable(req.versionStr(req.version)),
                       "status", Logging::Loggable(resp.status), "respsize",  Logging::Loggable(reply.size())));
--- a/regression-tests.api/test_Servers.py
+++ b/regression-tests.api/test_Servers.py
@@ -2,6 +2,7 @@ import json
 import operator
 import requests
 import unittest
+import socket
 from test_helper import ApiTestCase, is_auth, is_recursor, is_auth_lmdb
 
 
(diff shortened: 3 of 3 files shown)


zeha:
> Also url encode path in log lines, as it trips pytest.xml:

It trips `runtest.py`, which prints stdout/stderr, assuming everything is encoded 'just fine'. But if we are here because of encoding issues, maybe runtest.py should not assume this either...

omoerbeek:
> > Also url encode path in log lines, as it trips pytest.xml:
> 
> It trips `runtest.py`, which prints stdout/stderr, assuming everything is encoded 'just fine'. But if we are here because of encoding issues, maybe runtest.py should not assume this either...

`pytest.py` calls `pytest`. I did not fined where that is coming from yet.
But as there are other programs reading log lines that might have the same assumptions, I prefer to fix the generated output to be valid UTF8 and use pytest` as a way to detect log lines containing invalid UTF8.

ext/json11/json11.cpp
```diff
@@ -93,18 +93,10 @@ static void dump(const string &value, string &out) {
             out += "\\r";
         } else if (ch == '\t') {
             out += "\\t";
-        } else if (static_cast<uint8_t>(ch) <= 0x1f) {
+        } else if (static_cast<uint8_t>(ch) <= 0x1f || static_cast<uint8_t>(ch) >= 0x7f) {
```
cmouse, line 96: you should not emit `\u0000` that is not allowed in json strings. You may not emit anything that looks like surrogate, if you aren't using it as one,  and also any unicode > 9999 has to be encoded with surrogate pair. See https://en.wikipedia.org/wiki/UTF-16#Code_points_from_U+010000_to_U+10FFFF

omoerbeek: After some discussion on irc iit turned out that \x0000 is not an issue.

The other issue is that not everything is encoded in a form that can be translated back. 
For the moment we feel that that is not a big issue. The most important part is to produce valid UTF8 to not confuse clients.

zeha:
> > > Also url encode path in log lines, as it trips pytest.xml:
> > 
> > 
> > It trips `runtest.py`, which prints stdout/stderr, assuming everything is encoded 'just fine'. But if we are here because of encoding issues, maybe runtest.py should not assume this either...
> 
> `pytest.py` calls `pytest`. I did not fined where that is coming from yet. But as there are other programs reading log lines that might have the same assumptions, I prefer to fix the generated output to be valid UTF8 and use pytest` as a way to detect log lines containing invalid UTF8.

the traceback in your initial comment shows quite clearly where its coming from, or i dont understand what you're saying?

omoerbeek:
> > > > Also url encode path in log lines, as it trips pytest.xml:
> > > 
> > > 
> > > It trips `runtest.py`, which prints stdout/stderr, assuming everything is encoded 'just fine'. But if we are here because of encoding issues, maybe runtest.py should not assume this either...
> > 
> > 
> > `pytest.py` calls `pytest`. I did not fined where that is coming from yet. But as there are other programs reading log lines that might have the same assumptions, I prefer to fix the generated output to be valid UTF8 and use pytest` as a way to detect log lines containing invalid UTF8.
> 
> the traceback in your initial comment shows quite clearly where its coming from, or i dont understand what you're saying?

The traceback shows the stderr of `pytest` contains invalid UTF8. From the looks of it `pytest` prints the log. I just could not find the implementation of `pytest` itself. Anyway, if we would make this robust in the presence of invalid UTF8, we would loose the ability to detect invalid UTF8. That's the point I"m trying to make.

omoerbeek:
Rebased to pickup CI fix

rgacogne (approved):
I'm afraid I don't know much about Unicode encoding, but with that in mind:
- your reasoning makes sense to me
- the actual changes to the code also make sense to me

Habbie (approved):

zeha (approved):
ok from auth webserver perspective, imo

PR #12269, opened 2022-12-06
auth: fix invalid catalog zone sql query for gpgsqlbackend

frei-style:
### Short description

I tried to setup a catalog zone but it didn't work. I use PostgreSQL as a backend. The following error was logged repeatedly. PostgreSQL doesn't convert integers (0 in this case) to a boolean. I replaced 0 with false to get a valid SQL statement. I tested the SQL statement itself via psql and it worked, but I did not compile PowerDNS.

```
Oct 23 14:19:17 ns1.example.com pdns_server[924]: AXFR-out zone 'catalog.example.com', client '2001:abcd:99:ef::2a0:1', transfer initiated
Oct 23 14:19:17 ns1.example.com pdns_server[924]: TCP server is without backend connections in doAXFR, launching
Oct 23 14:19:17 ns1.example.com pdns_server[924]: TCP nameserver had error, cycling backend: virtual bool GSQLBackend::getCatalogMembers(const DNSName&, std::vector<CatalogInfo>&, CatalogInfo::CatalogType) unable to retrieve list of member zones: Fatal error during prePQpreparepare: select domains.id, domains.name, domains.options from records join domains on records.domain_id=domains.id and records.name=domains.name where domains.type='MASTER' and domains.catalog=$1 and records.type='SOA' and records.disabled=0: ERROR:  operator does not exist: boolean = integer
Oct 23 14:19:17 ns1.example.com pdns_server[924]: LINE 1: ...ins.catalog=$1 and records.type='SOA' and records.disabled=0
Oct 23 14:19:17 ns1.example.com pdns_server[924]:                                                                      ^
Oct 23 14:19:17 ns1.example.com pdns_server[924]: HINT:  No operator matches the given name and argument types. You might need to add explicit type casts.
```

### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/modules/gpgsqlbackend/gpgsqlbackend.cc
+++ b/modules/gpgsqlbackend/gpgsqlbackend.cc
@@ -140,7 +140,7 @@ public:
     declare(suffix, "update-serial-query", "", "update domains set notified_serial=$1 where id=$2");
     declare(suffix, "update-lastcheck-query", "", "update domains set last_check=$1 where id=$2");
     declare(suffix, "info-all-master-query", "", "select domains.id, domains.name, domains.type, domains.notified_serial, domains.options, domains.catalog, records.content from records join domains on records.domain_id=domains.id and records.name=domains.name where records.type='SOA' and records.disabled=false and domains.type in ('MASTER', 'PRODUCER')");
-    declare(suffix, "info-producer-members-query", "", "select domains.id, domains.name, domains.options from records join domains on records.domain_id=domains.id and records.name=domains.name where domains.type='MASTER' and domains.catalog=$1 and records.type='SOA' and records.disabled=0");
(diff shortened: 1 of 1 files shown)


mnordhoff:
FWIW, you don't have to recompile to test SQL query changes. They're all regular settings. You can set `gpgsql-info-producer-members-query` in `pdns.conf` (or `gpgsql.conf` or whatever).

frei-style:
> FWIW, you don't have to recompile to test SQL query changes. They're all regular settings. You can set `gpgsql-info-producer-members-query` in `pdns.conf` (or `gpgsql.conf` or whatever).

Thank you very much @mnordhoff, I didn't know that. I added the following line to `pdns.conf` and my catalog zone is working now.

```
gpgsql-info-producer-members-query=select domains.id, domains.name, domains.options from records join domains on records.domain_id=domains.id and records.name=domains.name where domains.type='MASTER' and domains.catalog=$1 and records.type='SOA' and records.disabled=false
```

zeha (approved):

PR #12278, opened 2022-12-06
dnsdist: Small improvements

rgacogne:
### Short description

- [Only allocate a ConnectionInfo object when actually needed](https://github.com/PowerDNS/pdns/commit/d4a188180a3b8e690bd215f04a5bdca6abcf53e5)
- [dnsdist: Prevent a copy when inserting a tag](https://github.com/PowerDNS/pdns/commit/7c878adc15091d0b0ed36d85f6ed4baa4c2f0edd)
- [dnsdist: Keep the local copy of the downstream servers around](https://github.com/PowerDNS/pdns/commit/9a0f37a7f99b22f133f3f1c333964dd793eec337)

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdist.cc
+++ b/pdns/dnsdist.cc
@@ -1942,13 +1942,12 @@ static void healthChecksThread()
   setThreadName("dnsdist/healthC");
 
   constexpr int interval = 1;
+  auto states = g_dstates.getLocal(); // this points to the actual shared_ptrs!
 
   for (;;) {
     sleep(interval);
 
     std::unique_ptr<FDMultiplexer> mplexer{nullptr};
-
-    auto states = g_dstates.getLocal(); // this points to the actual shared_ptrs!
     for (auto& dss : *states) {
       auto delta = dss->sw.udiffAndSet()/1000000.0;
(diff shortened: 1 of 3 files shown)


pdns/dnsdist.cc
```diff
@@ -1942,13 +1942,12 @@ static void healthChecksThread()
   setThreadName("dnsdist/healthC");
 
   constexpr int interval = 1;
+  auto states = g_dstates.getLocal(); // this points to the actual shared_ptrs!
```
omoerbeek, line 1945: I am wondering how this works when the set of downstreams changes.

rgacogne: We get a `LocalStateHolder` which holds a reference-counted snapshot of the last data, and checks whether there is an updated version every time we access the data, so it should be fine?

omoerbeek: Right. Somehow I thought the local state was only checked on `getLocal()`.

omoerbeek (approved):

PR #12281, opened 2022-12-06
dnsdist: Get rid of TCPCrossProtocolQuerySender

rgacogne:
### Short description

We need this construct to deal with cross-protocol queries, like queries received over TCP or DoT, but forwarded over DoH, because the thread dealing with the client and the one dealing with the backend will not be the same in that case, and we do not want to have different threads touching the same TCP connections. So we pass the query and response to the correct thread via pipes. Until now we were allocating an additional object, `TCPCrossProtocolQuerySender`, to deal with that case, but I noticed that the existing `IncomingTCPConnectionState` object already does everything we need, except that it needs to know that the response is a cross-protocol one in order to pass it via the pipe instead of treating it in a different way. This can be done by looking if the current thread ID differs from the one that created this object: if it does, we are dealing with a cross-protocol response and should pass it via the pipe, and if it does not we can deal with it directly.
This change saves the need to allocate a new object wrapped in a shared pointer for each cross-protocol query, which is quite nice.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdist-tcp.cc
+++ b/pdns/dnsdist-tcp.cc
@@ -491,7 +491,7 @@ void IncomingTCPConnectionState::updateIO(std::shared_ptr<IncomingTCPConnectionS
 /* called from the backend code when a new response has been received */
 void IncomingTCPConnectionState::handleResponse(const struct timeval& now, TCPResponse&& response)
 {
-  if (std::this_thread::get_id() != d_mainThreadID) {
+  if (std::this_thread::get_id() != d_creatorThreadID) {
     handleCrossProtocolResponse(now, std::move(response));
     return;
   }
--- a/pdns/dnsdist-tcp.cc
+++ b/pdns/dnsdist-tcp.cc
@@ -491,6 +491,11 @@ void IncomingTCPConnectionState::updateIO(std::shared_ptr<IncomingTCPConnectionS
 /* called from the backend code when a new response has been received */
 void IncomingTCPConnectionState::handleResponse(const struct timeval& now, TCPResponse&& response)
 {
+  if (std::this_thread::get_id() != d_mainThreadID) {
+    handleCrossProtocolResponse(now, std::move(response));
+    return;
+  }
+
   std::shared_ptr<IncomingTCPConnectionState> state = shared_from_this();
 
   if (response.d_connection && response.d_connection->getDS() && response.d_connection->getDS()->d_config.useProxyProtocol) {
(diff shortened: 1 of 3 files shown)


pdns/dnsdistdist/dnsdist-tcp-upstream.hh
```diff
@@ -170,6 +172,7 @@ static void handleTimeout(std::shared_ptr<IncomingTCPConnectionState>& state, bo
   size_t d_proxyProtocolNeed{0};
   size_t d_queriesCount{0};
   size_t d_currentQueriesCount{0};
+  std::thread::id d_mainThreadID;
```
omoerbeek, line 175: I find this name a bit confusing, as "main thread" is commonly used for the first thread of a process. Suggestions: `d_creatorThreadID`?

pdns/dnsdist-tcp.cc
```diff
@@ -566,66 +571,11 @@ struct TCPCrossProtocolResponse
...
-
-private:
-  std::shared_ptr<IncomingTCPConnectionState> d_state;
-};
-
 class TCPCrossProtocolQuery : public CrossProtocolQuery
 {
 public:
-  TCPCrossProtocolQuery(PacketBuffer&& buffer, IDState&& ids, std::shared_ptr<DownstreamState>& ds, std::shared_ptr<TCPCrossProtocolQuerySender>& sender): d_sender(sender)
+  TCPCrossProtocolQuery(PacketBuffer&& buffer, IDState&& ids, std::shared_ptr<DownstreamState> ds, std::shared_ptr<IncomingTCPConnectionState> sender): CrossProtocolQuery(InternalQuery(std::move(buffer), std::move(ids)), ds), d_sender(std::move(sender))
```
omoerbeek, line 577: `sender` is moved while it is not a moveable object

omoerbeek (approved):
Two nits

pdns/dnsdist-tcp.cc
rgacogne: I believe it is a movable object, we are moving the copy of the `shared_ptr` to avoid creating a new one (thus increasing the atomic reference counter before decreasing it again when exiting the constructor).

omoerbeek: Ah right. Wondering if the compiler would be smart enough without the expilcit `move`.

PR #12289, opened 2022-12-07
rec: refactor unsuppored qtype code and make sure we ServFail on all unsupported qtypes

### LINKED ISSUE #12251: rec: behaviour wrt a QTYPE0 (and a few others) changed: drop instead of ServFail
Previously we would anser `ServFail` on unsupported qtypes, but the shortcut introduced in #11338 changed that to a drop, as this code in `SyncRes::beginResolve` is no longer hit:
```
  auto qtypeCode = qtype.getCode();
  /* rfc6895 section 3.1 */
  if (qtypeCode == 0 || (qtypeCode >= 128 && qtypeCode <= 254) || qtypeCode == QType::RRSIG || qtypeCode == QType::NSEC3 || qtypeCode == QType::OPT || qtypeCode == 65535) {
    return -1;
  }
```

We have seen reports of clients asking TYPE0 questions, so this happens in the wild. The changed behaviour might confuse operators as they see dropped query counters increasing.

We need to re-evaluate if the shortcut is wise.

omoerbeek:
This fixes #12251

Also I'd like to know why we ServFail on NSEC3 but not on NSEC: we should either fix that or add a comment explaining this.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/qtype.hh
+++ b/pdns/qtype.hh
@@ -135,7 +135,7 @@ public:
 
   const static uint16_t rfc6895MetaLowerBound = 128;
   const static uint16_t rfc6895MetaUpperBound = 254; // Note 255: ANY is not included
-  const static uint16_t rfc6896Reserved = 65535;
+  const static uint16_t rfc6895Reserved = 65535;
 
   const static map<const string, uint16_t> names;
   const static map<uint16_t, const string> numbers;
--- a/pdns/qtype.hh
+++ b/pdns/qtype.hh
@@ -133,6 +133,10 @@ public:
 #endif
   };
 
+  const static uint16_t rfc6895MetaLowerBound = 128;
+  const static uint16_t rfc6895MetaUpperBound = 254; // Note 255: ANY is not included
+  const static uint16_t rfc6896Reserved = 65535;
+
   const static map<const string, uint16_t> names;
   const static map<uint16_t, const string> numbers;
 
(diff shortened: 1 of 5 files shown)


pdns/qtype.hh
```diff
@@ -133,6 +133,10 @@ public:
 #endif
   };
 
+  const static uint16_t rfc6895MetaLowerBound = 128;
+  const static uint16_t rfc6895MetaUpperBound = 254; // Note 255: ANY is not included
+  const static uint16_t rfc6896Reserved = 65535;
```
Habbie, line 138:
```suggestion
  const static uint16_t rfc6895Reserved = 65535;
```

pdns/syncres.hh
```diff
@@ -436,10 +436,21 @@ public:
...
+    if (qcode >= QType::rfc6895MetaLowerBound && qcode <= QType::rfc6895MetaUpperBound) {
+      return true;
+    }
+    switch (qcode) {
       // Internal types
-    case QType::ENT:
+    case QType::ENT: // aka TYPE0
     case QType::ADDR:
+      // RFC
+    case QType::rfc6896Reserved:
```
Habbie, line 449:
```suggestion
    case QType::rfc6895Reserved:
```

Habbie (changes requested):
good refactor, minor nits

Habbie (approved):

PR #12292, opened 2022-12-07
auth, rec: Wrap the CURL raw pointers in smart pointers

rgacogne:
### Short description

Having naked pointers around make my skin crawl.. Unfortunately this is harder that it should be because cURL used to alias `CURL` to `void` (why!?) and now cannot get rid of that because it would break too many projects, so we have to set `CURL_STRICTER` when it is available (7.50+) and fallback to naked pointers otherwise.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/minicurl.cc
+++ b/pdns/minicurl.cc
@@ -23,9 +23,16 @@
  */
 
 #include "minicurl.hh"
-#include <curl/curl.h>
 #include <stdexcept>
 #include <boost/format.hpp>
+
+#ifdef CURL_STRICTER
+#define getCURLPtr(x) \
+  x.get()
+#else
+#define getCURLPtr(x) \
+  x
+#endif
 
 void MiniCurl::init()
 {
@@ -42,17 +49,24 @@ void MiniCurl::init()
 
 MiniCurl::MiniCurl(const string& useragent)
 {
+#ifdef CURL_STRICTER
+  d_curl = std::unique_ptr<CURL, decltype(&curl_easy_cleanup)>(curl_easy_init(), curl_easy_cleanup);
+#else
   d_curl = curl_easy_init();
+#endif
   if (d_curl == nullptr) {
     throw std::runtime_error("Error creating a MiniCurl session");
   }
-  curl_easy_setopt(d_curl, CURLOPT_USERAGENT, useragent.c_str());
+  curl_easy_setopt(getCURLPtr(d_curl), CURLOPT_USERAGENT, useragent.c_str());
 }
 
 MiniCurl::~MiniCurl()
 {
-  // NEEDS TO CLEAN HOSTLIST
+  clearHeaders();
+  clearHostsList();
+#ifndef CURL_STRICTER
   curl_easy_cleanup(d_curl);
+#endif
 }
 
 size_t MiniCurl::write_callback(char *ptr, size_t size, size_t nmemb, void *userdata)
@@ -104,7 +118,7 @@ static string extractHostFromURL(const std::string& url)
 
 void MiniCurl::setupURL(const std::string& str, const ComboAddress* rem, const ComboAddress* src, int timeout, size_t byteslimit, bool fastopen, bool verify)
 {
-  if(rem) {
+  if (rem) {
     struct curl_slist *hostlist = nullptr; // THIS SHOULD BE FREED
 
     // url = http://hostname.enzo/url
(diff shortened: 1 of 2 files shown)


rgacogne:
The (unrelated) SNMP failure is weird:
```
2022-12-08T08:52:34.5756576Z === configs/dnsdist_TestSNMP.log ===
2022-12-08T08:52:34.5766682Z Read configuration from 'configs/dnsdist_TestSNMP.conf'
2022-12-08T08:52:34.5766932Z Added downstream server 127.0.0.1:5350
2022-12-08T08:52:34.5767150Z Adding server to default pool
2022-12-08T08:52:34.5767306Z unknown snmp version 193
2022-12-08T08:52:34.5767575Z NET-SNMP version 5.8 AgentX subagent connected
2022-12-08T08:52:34.5767744Z Listening on 127.0.0.1:5340
2022-12-08T08:52:34.5768268Z dnsdist 0.0.0-git1 comes with ABSOLUTELY NO WARRANTY. This is free software, and you are welcome to redistribute it according to the terms of the GPL version 2
2022-12-08T08:52:34.5768463Z ACL allowing queries from: 127.0.0.1/32
2022-12-08T08:52:34.5768688Z Console ACL allowing connections from: 127.0.0.0/8, ::1/128
2022-12-08T08:52:34.5768861Z Adding TCP Client thread
2022-12-08T08:52:34.5769034Z Adding TCP Client thread
2022-12-08T08:52:34.5769196Z Adding TCP Client thread
2022-12-08T08:52:34.5769358Z Adding TCP Client thread
2022-12-08T08:52:34.5769521Z Adding TCP Client thread
2022-12-08T08:52:34.5769683Z Adding TCP Client thread
2022-12-08T08:52:34.5769843Z Adding TCP Client thread
2022-12-08T08:52:34.5770005Z Adding TCP Client thread
2022-12-08T08:52:34.5770167Z Adding TCP Client thread
2022-12-08T08:52:34.5770333Z Adding TCP Client thread
2022-12-08T08:52:34.5770494Z Adding DoH Client thread
2022-12-08T08:52:34.5770780Z Marking downstream servername (127.0.0.1:5350) as 'up'
2022-12-08T08:52:34.5770948Z Got TCP connection from 127.0.0.1:33132
2022-12-08T08:52:34.5771214Z Got an exception while handling (writing) TCP query from 127.0.0.1:33132: EOF while reading message
2022-12-08T08:52:34.5771501Z Got query for simplea.snmpv2c.tests.powerdns.com|A from 127.0.0.1:57754, relayed to servername (127.0.0.1:5350)
2022-12-08T08:52:34.5771734Z Got answer from 127.0.0.1:5350, relayed to 127.0.0.1:57754, took 1037.71 usec
2022-12-08T08:52:34.5771981Z Timeout while waiting for the health check response from backend servername (127.0.0.1:5350)
2022-12-08T08:52:34.5772268Z Marking downstream servername (127.0.0.1:5350) as 'down'
2022-12-08T08:52:34.5772578Z Dropped query for simplea.snmpv2.tests.powerdns.com|A from 127.0.0.1:57754, no downstream server available
2022-12-08T08:52:34.5772861Z Marking downstream servername (127.0.0.1:5350) as 'up'
2022-12-08T08:52:34.5773028Z Exiting on user request
```
The initial health-check query succeeds, the first UDP query as well, then the second health-check fails and later a third one succeeds. I don't understand why this is happening, and why this did not happen before.

rgacogne:
I wonder if we would not be better off investigating https://github.com/libcpr/cpr instead.

pdns/minicurl.cc
```diff
@@ -104,7 +118,7 @@ static string extractHostFromURL(const std::string& url)
 
 void MiniCurl::setupURL(const std::string& str, const ComboAddress* rem, const ComboAddress* src, int timeout, size_t byteslimit, bool fastopen, bool verify)
 {
```
chbruyand, line 120: I *think* we should `clearHostsList()` first

chbruyand (commented):
Nice work (especially the host list mem leak)!
Just one nit as I don't think we want to keep appending stuff to that list while reusing the same MiniCurl instance ?

pdns/minicurl.cc
rgacogne: That's a nice one! It looks like we are good if the call is passing a non-null pointer for `rem`, as we would then completely replace the list, starting with an empty one, but we would indeed keep the list around if `rem` which doesn't seem good. I'll fix it, thanks!

rgacogne: Digging a bit more, I don't think we can currently reuse a `MiniCurl` object for more requests, as we are not calling `curl_easy_reset`. Technically it should be fine as long as we use the same parameters, but we might be leaking memory.
Are you aware of any place where we reuse these objects?

rgacogne: I'm going to push an attempt at making it safe to reuse a `MiniCurl` object, as this is clearly something that might happen

chbruyand (approved):

PR #12277, opened 2022-12-06
auth: add byteslimit support to lua ifurlup() function

chbruyand:
### Short description

This PR adds a `byteslimit` option to lua `ifurlup()` function that allows limiting the size read while querying the given URL.

Note that we may actually read a bit more than `byteslimit` depending on the underlying curl buffering.

Fix #12240 

### Checklist

I have:
- [x] read the [CONTRIBUTING.md](https://github.com/PowerDNS/pdns/blob/master/CONTRIBUTING.md) document
- [x] compiled this code
- [x] tested this code
- [x] included documentation (including possible behaviour changes)
- [ ] documented the code
- [ ] added or modified regression test(s)
- [ ] added or modified unit test(s)

PR #12297, opened 2022-12-07
minicurl: fix missing CURLOPT_XFERINFOFUNCTION on old curl versions

chbruyand:
### Short description
```
minicurl.cc:134:30: error: 'CURLOPT_XFERINFOFUNCTION' was not declared in this scope; did you mean 'CURLOPT_WRITEFUNCTION'?
```
 Fix missing `CURLOPT_XFERINFOFUNCTION` for libcurl versions before 7.32.0

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/minicurl.cc
+++ b/pdns/minicurl.cc
@@ -65,6 +65,7 @@ size_t MiniCurl::write_callback(char *ptr, size_t size, size_t nmemb, void *user
   return 0;
 }
 
+#if defined(LIBCURL_VERSION_NUM) && LIBCURL_VERSION_NUM >= 0x072000 // 7.32.0
 size_t MiniCurl::progress_callback(void *clientp, curl_off_t dltotal, curl_off_t dlnow, curl_off_t ultotal, curl_off_t ulnow)
 {
   if (clientp != nullptr) {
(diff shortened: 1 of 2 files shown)


Habbie:
Followup to #12277

Habbie:
Did you do a compile test on el7 by any chance?

chbruyand:
el7 build is green https://github.com/chbruyand/pdns/actions/runs/3639954288

pdns/minicurl.cc
```diff
@@ -130,9 +143,17 @@ void MiniCurl::setupURL(const std::string& str, const ComboAddress* rem, const C
   d_byteslimit = byteslimit;
   if (d_byteslimit > 0) {
     /* enable progress meter */
+#if defined(LIBCURL_VERSION_NUM)
+// CURL_AT_LEAST_VERSION has been introduced AFTER 7.32.0
```
Habbie, line 147: what's the relation between the `#if` and the comment?

Habbie (commented):
one question, otherwise good!

pdns/minicurl.cc
chbruyand: CURL_AT_LEAST_VERSION is easier to read than LIBCURL_VERSION_NUM

Habbie, line 147:
```suggestion
// CURL_AT_LEAST_VERSION has been introduced AFTER 7.32.0, so we have to use LIBCURL_VERSION_NUM
#if defined(LIBCURL_VERSION_NUM)
```

Habbie (changes requested):

pdns/minicurl.cc
Habbie: Right! Do we even need to check if LIBCURL_VERSION_NUM is defined? (if not, you can ignore my other suggestion)

chbruyand: I don't think so because we expect curl functions to be defined. But it is also checked at various other places so I kept the same pattern.

Habbie: It's confusing that `CURLOPT_NOPROGRESS` depends on whether `LIBCURL_VERSION_NUM` is defined. Maybe move the defined() check to go with the version check, like in the header?

omoerbeek: When you are testing a preprocessor symbol for greater than zero a `defined` guard is not _needed_, as undefined symbols evaluate to 0. Still I think it is worthwhile to have an explicit `defined` guard, as testing for other conditions (e.g. `== 0`) need it.

rgacogne (approved):

Habbie (approved):

PR #12309, opened 2022-12-09
Test ANY bind UDP handling for the v4 case

### LINKED ISSUE #3965: add tests for 0.0.0.0/:: responses
All our products support replying, on UDP, from the correct IP address even when bound to a wildcard (on OSes that support it). However, we do not test this. I touched this code today (in #3962) and now I need to remember to test this. Writing tests for this should not be a terribly huge challenge.

No product label as it applies to all - even though the code is shared.

omoerbeek:
v6 is more work, as the test client code assumes v4

Tested by neutering IsAnyAddress handling of auth and rec and verifying the tests fail.

Fixes #3965 (at least th v4 part).

### Checklist

I have:
- [ ] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/regression-tests.auth-py/test_AnyBind.py
+++ b/regression-tests.auth-py/test_AnyBind.py
@@ -1,0 +1,51 @@
+#!/usr/bin/env python
+import dns
+import os
+import socket
+
+from authtests import AuthTest
+
+
+class TestBindAny(AuthTest):
+    _config_template = """
+launch=bind
+"""
+
+    _zones = {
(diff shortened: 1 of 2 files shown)


Habbie (approved):

Habbie:
I wonder if there are kernels that will pick `127.0.0.2` as the socket source when we connect to `127.0.0.2`. Mine doesn't, and clearly yours doesn't, which gives me confidence this test will remain valid.

PR #12313, opened 2022-12-11
Clang-tidy: one more time

darnuria:
### Short description

One more pass of clang-tidy across the sources, someday bit by bit clangd will not burn my CPU while running. :rofl: 

### Checklist

I have:

- [x] compiled this code
- [ ] tested this code (doing unit+reg test)
- [x] clang tidy
- [ ] I have rebased after review is complete if neccessary

### DIFF
--- a/pdns/auth-catalogzone.hh
+++ b/pdns/auth-catalogzone.hh
@@ -41,10 +41,10 @@ public:
     Consumer
   };
 
-  static constexpr const char* getTypeString(enum CatalogType type)
+  static const string& getTypeString(enum CatalogType type)
   {
-    const char* types[] = {"none", "producer", "consumer"};
-    return types[type];
+    static const std::array<const string, 3> types = {"none", "producer", "consumer"};
+    return types.at(type);
   }
 
   CatalogInfo() :
--- a/pdns/dnsparser.cc
+++ b/pdns/dnsparser.cc
@@ -114,7 +114,7 @@ shared_ptr<DNSRecordContent> DNSRecordContent::deserialize(const DNSName& qname,
     (void) pos;
   }
 
-  MOADNSParser mdp(false, (char*)&*packet.begin(), (unsigned int)packet.size());
+  MOADNSParser mdp(false, reinterpret_cast<const char *>(packet.data()), packet.size());
   shared_ptr<DNSRecordContent> ret= mdp.d_answers.begin()->first.d_content;
   return ret;
 }
--- a/pdns/dnsparser.cc
+++ b/pdns/dnsparser.cc
@@ -96,8 +96,8 @@ shared_ptr<DNSRecordContent> DNSRecordContent::deserialize(const DNSName& qname,
 
   memcpy(&packet[0], &dnsheader, sizeof(dnsheader)); pos+=sizeof(dnsheader);
 
-  char tmp[6]="\x0" "\x0\x1" "\x0\x1"; // root question for ns_t_a
-  memcpy(&packet[pos], &tmp, 5); pos+=5;
+  constexpr std::array<uint8_t, 5> tmp= {'\x0', '\x0', '\x1', '\x0', '\x1' }; // root question for ns_t_a
+  memcpy(&packet[pos], tmp.data(), tmp.size()); pos += tmp.size();
 
   memcpy(&packet[pos], encoded.c_str(), encoded.size()); pos+=(uint16_t)encoded.size();
 
--- a/pdns/dnsparser.cc
+++ b/pdns/dnsparser.cc
@@ -108,7 +108,7 @@ shared_ptr<DNSRecordContent> DNSRecordContent::deserialize(const DNSName& qname,
   drh.d_clen=htons(serialized.size());
 
   memcpy(&packet[pos], &drh, sizeof(drh)); pos+=sizeof(drh);
-  if (serialized.size() > 0) {
+  if (!serialized.empty()) {
     memcpy(&packet[pos], serialized.c_str(), serialized.size());
     pos += (uint16_t) serialized.size();
     (void) pos;
--- a/pdns/pdnsutil.cc
+++ b/pdns/pdnsutil.cc
@@ -3854,7 +3854,7 @@ try
     }
     DNSName zone(cmds.at(1));
     string kind = cmds.at(2);
-    static vector<string> multiMetaWhitelist = {"ALLOW-AXFR-FROM", "ALLOW-DNSUPDATE-FROM",
+    const static std::array<string, 7> multiMetaWhitelist = {"ALLOW-AXFR-FROM", "ALLOW-DNSUPDATE-FROM",
       "ALSO-NOTIFY", "TSIG-ALLOW-AXFR", "TSIG-ALLOW-DNSUPDATE", "GSS-ALLOW-AXFR-PRINCIPAL",
       "PUBLISH-CDS"};
     bool clobber = true;
--- a/pdns/zoneparser-tng.cc
+++ b/pdns/zoneparser-tng.cc
@@ -66,7 +66,8 @@ void ZoneParserTNG::stackFile(const std::string& fname)
     std::error_code ec (err, std::generic_category());
     throw std::system_error(ec, "Unable to open file '" + fname + "': " + stringerror(err));
   }
-  struct stat st;
+
+  struct stat st = {};
   if (fstat(fd, &st) == -1) {
     int err = errno;
     close(fd);
@@ -132,7 +133,7 @@ unsigned int ZoneParserTNG::makeTTLFromZone(const string& str)
   if(str.empty())
     return 0;
 
-  unsigned int val;
+  unsigned int val = 0;
   try {
     pdns::checked_stoi_into(val, str);
   }
@@ -278,9 +279,11 @@ static void chopComment(string& line)
 {
   if(line.find(';')==string::npos)
     return;
-  string::size_type pos, len = line.length();
-  bool inQuote=false;
-  for(pos = 0 ; pos < len; ++pos) {
+
+  string::size_type pos = 0;
+  auto len = line.length();
+  bool inQuote = false;
+  for(; pos < len; ++pos) {
     if(line[pos]=='\\')
       pos++;
     else if(line[pos]=='"')
--- a/pdns/zoneparser-tng.cc
+++ b/pdns/zoneparser-tng.cc
@@ -311,7 +311,7 @@ DNSName ZoneParserTNG::getZoneName()
 
 string ZoneParserTNG::getLineOfFile()
 {
-  if (d_zonedata.size() > 0)
+  if (!d_zonedata.empty())
     return "on line "+std::to_string(std::distance(d_zonedata.begin(), d_zonedataline))+" of given string";
 
   if (d_filestates.empty())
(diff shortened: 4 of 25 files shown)


pdns/auth-catalogzone.hh
```diff
@@ -43,8 +43,8 @@ public:
 
   static constexpr const char* getTypeString(enum CatalogType type)
   {
-    const char* types[] = {"none", "producer", "consumer"};
-    return types[type];
+    const std::array<const char*, 3> types = {"none", "producer", "consumer"};
+    return types.at(type);
   }
```
darnuria, line 48: If it makes sense I can move this change in an other PR. Hesitated to change to string_view should I do it?

pdns/dnsparser.cc
```diff
@@ -96,8 +96,8 @@ shared_ptr<DNSRecordContent> DNSRecordContent::deserialize(const DNSName& qname,
 
   memcpy(&packet[0], &dnsheader, sizeof(dnsheader)); pos+=sizeof(dnsheader);
 
-  char tmp[6]="\x0" "\x0\x1" "\x0\x1"; // root question for ns_t_a
-  memcpy(&packet[pos], &tmp, 5); pos+=5;
+  const std::array<uint8_t, 6> tmp= {'\x0', '\x0', '\x1', '\x0', '\x1', '\0' }; // root question for ns_t_a
+  memcpy(&packet[pos], tmp.data(), 5); pos+=5;
```
darnuria, line 100: Same on this one I can make an other PR, asked myself if a `string_view` is not better.

pdns/pdnsutil.cc
```diff
@@ -3854,7 +3854,7 @@ try
     }
     DNSName zone(cmds.at(1));
     string kind = cmds.at(2);
-    static vector<string> multiMetaWhitelist = {"ALLOW-AXFR-FROM", "ALLOW-DNSUPDATE-FROM",
+    const static std::array<string, 7> multiMetaWhitelist = {"ALLOW-AXFR-FROM", "ALLOW-DNSUPDATE-FROM",
       "ALSO-NOTIFY", "TSIG-ALLOW-AXFR", "TSIG-ALLOW-DNSUPDATE", "GSS-ALLOW-AXFR-PRINCIPAL",
       "PUBLISH-CDS"};
```
darnuria, line 3859: Same maybe an Array of `string_view` is better but preferred to leave a question can move into an other PR!

darnuria (commented):
Leaved some comment about possible usage of `string_view`. it's possible to split this out into multiple PR if you prefers! Tell me. :)

darnuria:
hmm I have an itoa issue, will do an other PR cannot build mysel since I rebased onto `origin/master` !

pdns/auth-catalogzone.hh
omoerbeek: what's the reason you suggest `string_view` instead of just `string`?

pdns/dnsparser.cc
omoerbeek: Some question here: what's the reason you suggest string_view instead of just string?

pdns/recursor_cache.cc
```diff
@@ -145,11 +144,11 @@ time_t MemRecursorCache::handleHit(MapCombo::LockedContent& content, MemRecursor
   time_t ttd = entry->d_ttd;
   origTTL = entry->d_orig_ttl;
 
-  if (variable && (!entry->d_netmask.empty() || entry->d_rtag)) {
+  if (variable != nullptr && (!entry->d_netmask.empty() || entry->d_rtag)) {
     *variable = true;
   }
 
-  if (res) {
+  if (!res->empty()) {
```
omoerbeek, line 151: This is wrong, `res` can be a `nullptr`

pdns/recursor_cache.hh
```diff
@@ -146,7 +146,7 @@ private:
 
     Netmask lookupBestMatch(const ComboAddress& addr) const
     {
-      const auto best = d_nmt.lookup(addr);
+      const auto * best = d_nmt.lookup(addr);
```
omoerbeek, line 149: Wondering what the point is here, did clang-tidy suggest this?

pdns/zoneparser-tng.cc
```diff
@@ -107,14 +107,19 @@ static string makeString(const string& line, const pair<string::size_type, strin
...
 static bool isTimeSpec(const string& nextpart)
 {
-  if(nextpart.empty())
+  if (nextpart.empty()) {
     return false;
-  for(string::const_iterator iter = nextpart.begin(); iter != nextpart.end(); ++iter) {
-    if(isdigit(*iter))
+  }
+  for (auto iter = nextpart.begin(); iter != nextpart.end(); ++iter) {
+    if (isdigit(*iter) != 0) {
```
omoerbeek, line 114: to be safe, `*iter` should be cast to `unsiged char`. See https://en.cppreference.com/w/cpp/string/byte/tolower

pdns/zoneparser-tng.cc
```diff
@@ -107,14 +107,19 @@ static string makeString(const string& line, const pair<string::size_type, strin
...
       continue;
-    if(iter+1 != nextpart.end())
+    }
+
+    if (iter + 1 != nextpart.end()) {
       return false;
-    char c=tolower(*iter);
+    }
+
+    char c = static_cast<char>(tolower(*iter));
```
omoerbeek, line 122: Same here

omoerbeek (changes requested):
A few questions and a few improvements cspotted.

pdns/pdnsutil.cc
omoerbeek: Some question here: what's the reason you suggest string_view instead of just string?

pdns/auth-catalogzone.hh
darnuria: hmm it's my limit with C++ memory model, I don't know now (exactly) where are stored `static` (program lived) `strings` in cpp. At least I know that a `char *` is not heap allocated in this context.

Context: I mostly do rust, `String` are heap beast where `&'static str` are in memory image of the program.

pdns/recursor_cache.cc
darnuria: Oh good spot, totally true. With the related issue I close about .yy files and me not forgetting a make I missed this one Should have put this PR in Draft.

pdns/recursor_cache.hh
darnuria: > readability-qualified-auto: `const auto best` can be declared auto * const best`. 

And also one of my error, that I always do in cpp misplaced the `const` in my fix. :facepalm:

pdns/zoneparser-tng.cc
darnuria: Thanks will finish processing review later this morning it's really interresting thanks for your time. I have somes holes with Cpp rust and C are OK.

Will take some time to check what exactly clang-tidy suggested hope it's my error and not the suggestion

pdns/recursor_cache.hh
darnuria: Pushed something to follow exactly clang suggestion

pdns/auth-catalogzone.hh
darnuria: In this case it would be a pain would need change in json11 maybe not today. I suppose.

```cpp
auth-catalogzone.cc: In member function ‘void CatalogInfo::fromJson(const std::string&, CatalogType)’:
auth-catalogzone.cc:41:15: error: no match for ‘operator[]’ (operand types are ‘json11::Json’ and ‘std::string_view’ {aka ‘std::basic_string_view<char>’})
   41 |     if (!d_doc[getTypeString(d_type)].is_null()) {
      |               ^
```

pdns/dnsparser.cc
darnuria: Would have done something like:
```cpp
std::string_view tmp = "\x0" "\x0\x1" "\x0\x1\0"; // root question for ns_t_a
memcpy(&packet[pos], tmp.data(), tmp.length() - 1); pos += 5;
```
Or 
```cpp
std::string_view tmp = "\x0" "\x0\x1" "\x0\x1"; // root question for ns_t_a
memcpy(&packet[pos], tmp.data(), tmp.length()); pos += 5;
```

darnuria:
Push force'ed to fix conflics

pdns/auth-catalogzone.hh
omoerbeek: > hmm it's my limit with C++ memory model, I don't know now (exactly) where are stored `static` (program lived) `strings` in cpp. At least I know that a `char *` is not heap allocated in this context.
> 
> Context: I mostly do rust, `String` are heap beast where `&'static str` are in memory image of the program.

Often `string` implementations store the data for small string inside the object itself, and only use the heap for larger strings. For `const string`s there might be more smarter things going on.

But does it really matter how strings are stored exactly?  The main thing is to avoid naked pointers. 

As your json11 example shows, many functions expect `string` and not `string_view`. Another reason to go for `string`.

darnuria: It's more clear, will just goes for `string` , thanks for the explanation.

pdns/dnsparser.cc
darnuria: Why would I do that: I mostly just want a contiguous view on constant Read only memory as `src` for `memcpy` but if `string` is fine I can just go `const string`. But what drived me to `string_view` is that `string` is not usable in `constexp` and that I just want a view of memory like a `char *` not a heap allocated thing. :)

omoerbeek: I think a `std::array` is fine in this case, but please fix the 5 vs 6 inconsistency.

darnuria: That's a good point: I prefers to spell it out, I added did a size of 6 to explicitly include a `\0` like the `char *` did but I suppose it's just better do "do not add this `\0` not used. and go for 5.

To avoid inconsistency: keep 5 and do not include `\0` since we don't read it, right?

pdns/auth-catalogzone.hh
```diff
@@ -41,10 +41,10 @@ public:
     Consumer
   };
 
-  static constexpr const char* getTypeString(enum CatalogType type)
+  static const string& getTypeString(enum CatalogType type)
   {
-    const char* types[] = {"none", "producer", "consumer"};
-    return types[type];
+    static const std::array<const string, 3> types = {"none", "producer", "consumer"};
+    return types.at(type);
```
darnuria, line 47: I remember: The main problem of removing the `char*` is loosing the `constexpr` It was one of the reason for trying out `string_view` instead of `string`

pdns/dnsparser.cc
```diff
@@ -96,8 +96,8 @@ shared_ptr<DNSRecordContent> DNSRecordContent::deserialize(const DNSName& qname,
 
   memcpy(&packet[0], &dnsheader, sizeof(dnsheader)); pos+=sizeof(dnsheader);
 
-  char tmp[6]="\x0" "\x0\x1" "\x0\x1"; // root question for ns_t_a
-  memcpy(&packet[pos], &tmp, 5); pos+=5;
+  constexpr std::array<uint8_t, 5> tmp= {'\x0', '\x0', '\x1', '\x0', '\x1' }; // root question for ns_t_a
```
omoerbeek, line 99:
```suggestion
  constexpr std::array<uint8_t, 5> tmp = {'\x0', '\x0', '\x1', '\x0', '\x1' }; // root question for ns_t_a
```

omoerbeek (approved):

darnuria:
Rebased against `master` and incorporated review into original commits if any rebase/squashing is neccessary let me know!

omoerbeek:
Thanks! Still looks good to me after latest changes. I would not mind an extra review though.

darnuria:
@omoerbeek no problem, I am available for any change needed or rebase.

Also happy new year! :) :tada:

rgacogne (approved):
LGTM, thanks!

darnuria:
Thanks for the review time it was instructive. :)

PR #12314, opened 2022-12-11
to_string: Fix compilation with gcc12: itoa not declared in this scope

darnuria:
```
bindparser.yy: In function ‘void yyerror(const char*)’:
bindparser.yy:39:93: error: ‘itoa’ was not declared in this scope
   39 | PDNSException("Error in bind configuration '"+string(current_filename)+"' on line "+std::to_string(linenumber)+": "+str);
      |                                                                                     ^~~~
```
### Checklist

I have:
- [x] compiled this code
- [x] tested this code

### DIFF
--- a/pdns/bindparser.yy
+++ b/pdns/bindparser.yy
@@ -1,6 +1,7 @@
 %{
 
 #include <stdio.h>
+#include <string>
 #include <string.h>
 #include <stdlib.h>
 #include <string>


omoerbeek (dismissed):
I was confused by the PR title, I think it should read `to_string` not declared

jsoref:
the `.cc` file probably has [`#line`](https://learn.microsoft.com/en-us/cpp/preprocessor/hash-line-directive-c-cpp?view=msvc-170) referencing a `.yy` file.

Habbie:
it does :)

darnuria:
Oh willl check if my configure/clean has an issue or if i am just missing something.

darnuria:
Yep definitely  it's forgetting to do a `make clean` + configure after a rebase onto upstream master my bad. The error was really misleading and followed blindly gcc advise!

PR #12320, opened 2022-12-13
dnsdist: Fix loading PKCS12-encrypted (RC2 CBC) certs with OpenSSL 3.x

rgacogne:
### Short description

Note that the `openssl pkcs12` command uses `AES-256-CBC with PBKDF2` since OpenSSL 3.x, meaning PKCS12 files generated from now on do not have this issue, but existing ones do. 

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/pdns/libssl.cc
+++ b/pdns/libssl.cc
@@ -20,6 +20,9 @@
 #include <openssl/ocsp.h>
 #endif /* DISABLE_OCSP_STAPLING */
 #include <openssl/pkcs12.h>
+#if defined(OPENSSL_VERSION_MAJOR) && OPENSSL_VERSION_MAJOR >= 3
+#include <openssl/provider.h>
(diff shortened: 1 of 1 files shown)


chbruyand (approved):
lgtm!
Out of curiosity, what would be the cost of trying to load the legacy provider anyway if running OpenSSL 3.x (which would simplify a bit the code IMO) ?

rgacogne:
> Out of curiosity, what would be the cost of trying to load the legacy provider anyway if running OpenSSL 3.x (which would simplify a bit the code IMO) ?

I think the cost would be negligible, but it would mean enabling unsafe ciphers, digests and perhaps protocols (I'm unsure about that last one) so I would like to avoid that if possible: https://www.openssl.org/docs/man3.0/man7/OSSL_PROVIDER-legacy.html

PR #12329, opened 2022-12-14
dnsdist: Turn DNSQuestion and DNSResponse into internal state overlays

rgacogne:
### Short description

We have had several issues related to some of the values being passed back and forth between the `DNSQuestion` /`DNSResponse` objects and the `IDState` object that keeps the information while the query is in flight.
This PR changes the `DNSQuestion` and `DNSResponse` objects work by making them hold a reference to the a new underlying `InternalQueryState` object which becomes the de-facto source of truth. This `InternalQueryState` is very similar to the existing `IDState` but without the parts that are specific to the UDP path, as the rest of the code doesn't need these.
This also makes it possible to clean up a few fields that were duplicated, or whose meaning was unclear.
That change should be completely transparent to Lua, so neither the configuration or Lua scripts are impacted.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code


chbruyand (approved):
LGTM!

PR #12334, opened 2022-12-16
rec: Generate EDE in more cases, specifically on unreachable auths or sythesized results.

omoerbeek:
Examples:
```
; EDE: 0 (Other): (Result synthesized from aggressive NSEC cache (RFC8198))
; EDE: 0 (Other): (Result synthesized by root-nx-trust)
; EDE: 22 (No Reachable Authority): (delegation timeout4.com)
```

As there is no specific EDE for synthesized, use noError with a text.

We have to be careful here: a single client query can lead to multiple beginResolve calls. Some of these are done after the main result has been looked up, for example to validate the result. These subsequent calls can generate EDE's but we do not want to copy the EDE to the main result in those cases. A typical example would be an absent DS for an Insecure domain. Nothing wrong with these but we do not want the potential absent DS EDE (which could be synthesized) to be returned with the main query,

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/recursordist/syncres.cc
+++ b/pdns/recursordist/syncres.cc
@@ -2822,13 +2822,15 @@ bool SyncRes::doCacheCheck(const DNSName& qname, const DNSName& authname, bool w
         if (ne.d_qtype.getCode()) {
           LOG(prefix << qname << ": " << qtype << " is negatively cached via '" << ne.d_auth << "' for another " << sttl << " seconds" << endl);
           res = RCode::NoError;
+          if (s_addExtendedResolutionDNSErrors) {
+            context.extendedError = EDNSExtendedError{0, "Result from negative cache"};
+          }
         }
         else {
           LOG(prefix << qname << ": Entire name '" << qname << "' is negatively cached via '" << ne.d_auth << "' for another " << sttl << " seconds" << endl);
-        }
-        if (s_addExtendedResolutionDNSErrors) {
-          // XXX Do we want this?
-          context.extendedError = EDNSExtendedError{0, "Result from negative cache"};
+          if (s_addExtendedResolutionDNSErrors) {
+            context.extendedError = EDNSExtendedError{0, "Result from negative cache for entire name"};
+          }
         }
       }
     }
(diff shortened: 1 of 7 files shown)


rgacogne:
> I am not terribly happy with the extra argument. Maybe we should move to an object holding the parameters and result status of the nested or subsequent calls. This would also avoid some of the saveX, setX, beginResolve, restore X sequences.

My rule of thumb is, when I'm pondering adding an object to group values it's pretty much always the right call :)

pdns/recursordist/syncres.cc
```diff
@@ -2822,6 +2826,10 @@ bool SyncRes::doCacheCheck(const DNSName& qname, const DNSName& authname, bool w
         else {
           LOG(prefix << qname << ": Entire name '" << qname << "' is negatively cached via '" << ne.d_auth << "' for another " << sttl << " seconds" << endl);
         }
+        if (s_addExtendedResolutionDNSErrors) {
+          // XXX Do we want this?
+          context.extendedError = EDNSExtendedError{0, "Result from negative cache"};
```
rgacogne, line 2831: That would make sense to me, at least when the entire name is negatively cached (and it would be nice to convey that in our EDE message).

pdns/recursordist/syncres.cc
```diff
@@ -2992,7 +3003,7 @@ bool SyncRes::doCacheCheck(const DNSName& qname, const DNSName& authname, bool w
       if (!giveNegative)
         res = 0;
       LOG(prefix << qname << ": updating validation state with cache content for " << qname << " to " << cachedState << endl);
```
rgacogne, line 3005: Perhaps we might want to add an EDE status on positive hits as well? But that might be a bit too much?

rgacogne (approved):
Looks good to me!

pdns/recursordist/syncres.cc
omoerbeek: A push a commit making the distinction

pdns/recursordist/syncres.cc
omoerbeek: IMO that would be a bit too much

PR #12337, opened 2022-12-19
rec: tcounter test tweaks

omoerbeek:
Some system habe a low resoltion nanosleep(), which sleeps for at least a few ms. Compensate for that by running fewer loops with longer sleeps.

Also use dns_random and make sure it is initlized properly for all tests.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/recursordist/testrunner.cc
+++ b/pdns/recursordist/testrunner.cc
@@ -31,6 +31,8 @@
 #include <iomanip>
 #include "logger.hh"
 #include "logging.hh"
+#include "arguments.hh"
+#include "dns_random.hh"
 
 static std::string s_timestampFormat = "%s";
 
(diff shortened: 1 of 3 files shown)


pdns/recursordist/testrunner.cc
```diff
@@ -76,6 +78,10 @@ static void loggerBackend(const Logging::Entry& entry)
 
 static bool init_unit_test()
 {
+  ::arg().set("rng") = "auto";
+  ::arg().set("entropy-source") = "/dev/urandom";
+  // Force init while w're still unthreaded
```
rgacogne, line 83:
```suggestion
  // Force init while we are still unthreaded
```

rgacogne (approved):

PR #12352, opened 2022-12-21
Refresh of negcache stale entry might use wrong qtype

zjs604381586:
### Short description

When searching in negcache, when the stale function is triggered and the type of the cache is type0, the asynchronous expired task request resolution type is A record, and the actual client request is AAAA type request, so the request type of the asynchronous expired task should be changed Change to the type requested by the user
### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


omoerbeek:
Please wait with a backport to 4.8.x until this is merged. I see some things I want to change in this PR. Some patience required, thanks.

rgacogne (approved):

Issue #12358, opened 2022-12-22
LUA dblookup for records other than QType other than LUA


pfak:
- Program: Authoritative
 - Issue type: Feature request

### Short description
Presently `dblookup`  in an `IN LUA` record only supports LUA lookups. It would be helpful if it could support looking up for example `A`, `AAAA` QType so that the IN LUA record can reference other records instead of IP addresses.

### Usecase
To be able to use DNS records instead of IP addresses in `ifurlup` in `LUA` QType.

### Description
Record example:
```
IN LUA A ifurlup("luarecord.example.com", {dblookup("host1.example.com.", "A"), dblookup("host2.example.com", "A")})
```

Habbie:
> Presently `dblookup` in an `IN LUA` record only supports LUA lookups.

it's `include`, which looks up -and executes- LUA records.

pfak:
The build that @Habbie provided me works as I envisioned for this feature request. 

Examples of what I did:

1. Expose online PowerDNS-Admin instance:

```
dns IN LUA A "ifurlup('https://dns.example.com/powerdns-admin/login', {dblookup('ns1.example.com', 'A'), dblookup('ns2.example.com', 'A'), dblookup('ns3.example.com', 'A')}, {stringmatch='PowerDNS-Admin', selector='hashed'})"
dns IN LUA AAAA "ifurlup('https://dns.example.com/powerdns-admin/login', {dblookup('ns1.example.com', 'AAAA'), dblookup('ns2.example.com', 'AAAA'), dblookup('ns3.example.com', 'AAAA')}, {stringmatch='PowerDNS-Admin', selector='hashed'})"
```

2. Return IP address to connect to my home network for WireGuard listening on OpenWrt (I did not use `ifportup` because it does not support ordering, and I needed to return the first result if it is up):

```
IN A "ifurlup('http://remote.example.com/', {dblookup('isp1.remote.example.com', 'A'), dblookup('isp2.remote.example.com', 'A')}, {stringmatch='LuCI - Lua Configuration Interface', selector='hashed'})"
```

Habbie:
> The build that @Habbie provided me works as I envisioned for this feature request.

that's #12359

Habbie:
> that's #12359

which was merged!

peterthomassen:
[...]

PR #12359, opened 2022-12-23
auth LUA: add dblookup function

Habbie:
### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/lua-record.cc
+++ b/pdns/lua-record.cc
@@ -4,6 +4,7 @@
 #include <utility>
 #include <algorithm>
 #include <random>
+#include "qtype.hh"
 #include "version.hh"
 #include "ext/luawrapper/include/LuaContext.hpp"
 #include "lock.hh"
@@ -501,6 +502,17 @@ static std::vector<DNSZoneRecord> lookup(const DNSName& name, uint16_t qtype, in
     }
   }
   return ret;
+}
+
+static bool getAuth(const DNSName& name, uint16_t qtype, SOAData* soaData)
+{
+  static LockGuarded<UeberBackend> s_ub;
+
+  vector<DNSZoneRecord> ret;
+  {
+    auto ueback = s_ub.lock();
+    return ueback->getAuth(name, qtype, soaData);
+  }
 }
 
 static std::string getOptionValue(const boost::optional<std::unordered_map<string, string>>& options, const std::string &name, const std::string &defaultValue)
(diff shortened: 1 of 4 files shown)


Habbie:
zeha: depending on usage context, the user may want to put `[1]` after dblookup to get one item instead of a list.

dwfreed: "Some things, like ifurlup, don't like empty tables, so be careful not to accidentally look up a name that does not have any records of that type"

Habbie:
after squash, commit message should mention that this fixes #12358

pdns/lua-record.cc
```diff
@@ -1103,6 +1115,36 @@ static void setupLuaRecords(LuaContext& lua)
       return result;
     });
 
+  lua.writeFunction("dblookup", [](const string& record, const string& type) {
+    DNSName rec;
+    QType qtype;
+    vector<string> ret;
+    try {
+      rec = DNSName(record);
+      qtype = type;
```
omoerbeek, line 1124: `QType &operator=(const string &)` does not throw an exception on error, but assigns zero.

pdns/lua-record.cc
```diff
@@ -1103,6 +1115,36 @@ static void setupLuaRecords(LuaContext& lua)
...
+    }
+    catch (const std::exception& e) {
+      g_log << Logger::Error << "DB lookup cannot be performed, the name (" << record << ") or type (" << type << ") is malformed: " << e.what() << endl;
```
omoerbeek, line 1127:
```suggestion
      g_log << Logger::Error << "DB lookup cannot be performed, the name (" << record << ") or type (" << qtype << ") is malformed: " << e.what() << endl;
```

pdns/lua-record.cc
```diff
@@ -1103,6 +1115,36 @@ static void setupLuaRecords(LuaContext& lua)
...
+        return ret;
+      }
+
+      vector<DNSZoneRecord> drs = lookup(rec, qtype, soaData.domain_id);
+      for (const auto& drec : drs) {
+        ret.push_back(drec.dr.getContent()->getZoneRepresentation());
+      }
+    }
+    catch (std::exception& e) {
+      g_log << Logger::Error << "Failed to do DB lookup for " << rec << "/" << type << ": " << e.what() << endl;
```
omoerbeek, line 1143:
```suggestion
      g_log << Logger::Error << "Failed to do DB lookup for " << rec << "/" << qtype << ": " << e.what() << endl;
```

pdns/lua-record.cc
```diff
@@ -503,6 +504,17 @@ static std::vector<DNSZoneRecord> lookup(const DNSName& name, uint16_t qtype, in
   return ret;
 }
 
+static bool getAuth(const DNSName& name, uint16_t qtype, SOAData* soaData)
+{
+  static LockGuarded<UeberBackend> s_ub;
+
+  vector<DNSZoneRecord> ret;
```
omoerbeek, line 511: Looks unused

pdns/lua-record.cc
Habbie: good catch, will handle

pdns/lua-record.cc
Habbie: hmm, if we reach this line, we know for a fact qtype has not been set, right?

omoerbeek: Hmm, right i mixed the types up.

pdns/lua-record.cc
Habbie:
```suggestion
      qtype = type;
      if (qtype.getCode() == 0) {
        throw std::invalid_argument("unknown type");
      }
```

plus leave `type` in the next log line?

omoerbeek: yes

pdns/lua-record.cc
Habbie: i did apply this one

Habbie:
thank you, should all be resolved now

PR #12374, opened 2023-01-03
Rec: make cache cleaning more fair when under pressure

omoerbeek:
### Short description

Based on an initial diff by @rgacogne 

Take shard size and number of remaining shards into account when cleaning in pressure mode, so that big shards get pruned more and if we are behind due to rounding etc we get more eager.

This also adds code to print a few details on shards.
The negcache size and number of shards is now also made consistent: negcach is 1/8th of the size of the record cache and also has 1/8 of the shards if the record cache.

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code


pdns/cachecleaner.hh
```diff
@@ -167,43 +176,51 @@ uint64_t pruneMutexCollectionsVector(C& container, std::vector<T>& maps, uint64_
...
+    size_t removed = 0;
+    for (auto i = sidx.begin(); i != sidx.end() && removed < toTrimForThisShard; removed++) {
+      container.preRemoval(*shard, *i);
+      i = sidx.erase(i);
+      --content.d_entriesCount;
+      ++totErased;
+      if (--toTrim == 0) {
+        return totErased;
       }
+      --cacheSize;
```
rgacogne, line 221: It looks unused at this point?

pdns/cachecleaner.hh
```diff
@@ -167,43 +176,51 @@ uint64_t pruneMutexCollectionsVector(C& container, std::vector<T>& maps, uint64_
...
-        }
+  /* it was not enough, so we need to remove entries that are not expired,
+     still using the LRU */
+
+  double remainingShards = numberOfShards;
+  const auto factor = static_cast<double>(numberOfShards) / static_cast<double>(cacheSize);
+  for (auto& content : maps) {
+    auto shard = content.lock();
+    const auto shardSize = shard->d_map.size();
+    const uint64_t toTrimForThisShard = std::round(factor * toTrim * shardSize / remainingShards);
```
rgacogne, line 205: I'm not sure the computation is still correct for the last shards, as the number of entries (`cacheSize`) from which the ratio is computed is not updated. I wonder if taking the value `toTrim` before that loop times `cacheSize` before the loop divided by the shard size would not be more accurate?

rgacogne (changes requested):

pdns/cachecleaner.hh
omoerbeek: Hmm I think I messed up when cleaning the code before commit. Will revisit.

pdns/cachecleaner.hh
```diff
@@ -131,34 +131,43 @@ uint64_t pruneLockedCollectionsVector(std::vector<T>& maps)
...
-    auto mc = content.lock();
-    mc->invalidate();
-    auto& sidx = boost::multi_index::get<S>(mc->d_map);
-    uint64_t erased = 0, lookedAt = 0;
+    auto shard = content.lock();
+    const auto shardSize = shard->d_map.size();
+    const uint64_t toScanForThisShard = std::ceil(lookAt * ((1.0 * shardSize) / cacheSize));
+    const uint64_t toTrimForThisShard = toTrim > 0 ? std::ceil(toTrim * ((1.0 * shardSize) / cacheSize)) : 0;
+    if (toTrimForThisShard == 0) {
+      continue;
```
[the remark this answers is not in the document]
omoerbeek: Yes you are right, we do want to evict items even if `toTrim` is 0.

pdns/cachecleaner.hh
omoerbeek: Dividing by shard size is not correct in any case, as we want ro remove more from a shard if it is big

omoerbeek: So the issue with `const uint64_t toTrimForThisShard = std::ceil(1.0 * toTrim * shardSize / cacheSize);`
is that it overshoots (cleans too much) because of the `ceil`. A `round` undershoots in some cases. I was trying to compensate for that by updating `toTrim` and `cacheSize` in the loop but obviously messed up.

I need to think about if overshooting (cleaning too much) is bad (it is limited by the number of shards anyway).

omoerbeek:
Pushed a fixed version, with more explanation on the strategy used. One point to consider: in the first loop, we might ignore `toTrim` and just hunt for expired entries, using the `toScan` bound.
That means that under pressure, we prefer to evict more expired entries and will fall back to the second loop in less cases.
It might increase imbalance, but that *should* not be an issue: we are removing stale items only.

Another thing to ponder:

Even while "data" record sets of a zone are retrieved and placed in the back of the expiry queue, the infra (NS + glue) record sets for the zone are not, as they are not touched when just checking the known "data" record sets of a zone. So infra record sets tend to travel to the front of the expiry queue, even when their data records are very popular. Only an expired data record set will trigger a retrieval of the infra  record set from the cache so that it will be placed in the back of the expiry queue. So when the cache is under pressure it might clean more infra records than you would think.

This might be a case of "this is just the way it is", but IMO this is another reason we want to consider a separate cache fo infra records, with a different eviction policy.

pdns/cachecleaner.hh
```diff
@@ -167,15 +172,15 @@ uint64_t pruneMutexCollectionsVector(C& container, std::vector<T>& maps, uint64_
         ++i;
       }
 
-      if (toTrim && erased >= toTrim / maps_size)
+      if (toTrim > 0 && erased >= toTrimForThisShard) {
```
rgacogne, line 175: I wonder if it is smart to actually stop scanning for expired entries after we have cleaned up enough entries from this shard to meet our target. If we are under pressure (cache is full), we will likely end up cleaning just enough entries to get just below the maximum cache size, but not less, which means that the cache will very likely get full again before the next scan, and so on and so forth.
Perhaps we should keep scanning for expired entries even after meeting our target?

pdns/cachecleaner.hh
```diff
@@ -184,26 +189,53 @@ uint64_t pruneMutexCollectionsVector(C& container, std::vector<T>& maps, uint64_
...
-        --content.d_entriesCount;
-        totErased++;
-        toTrim--;
-        if (toTrim == 0) {
-          return totErased;
-        }
+  // It was not enough, so we need to remove entries that are not
+  // expired, still using the LRU index.
+
+  // From here on cacheSize is the total size of the shards that still
```
rgacogne, line 195:
```suggestion
  // From here on cacheSize is the total number of entries in the shards that still
```

pdns/cachecleaner.hh
```diff
@@ -184,26 +189,53 @@ uint64_t pruneMutexCollectionsVector(C& container, std::vector<T>& maps, uint64_
...
+  // need to be cleaned. When a shard is processed, we subtract its
+  // original size from cacheSize as it used to compute the
```
rgacogne, line 197:
```suggestion
  // original size from cacheSize as we use this value to compute the
```

pdns/cachecleaner.hh
```diff
@@ -184,26 +189,53 @@ uint64_t pruneMutexCollectionsVector(C& container, std::vector<T>& maps, uint64_
...
+  //
+  // Suppose we have 10 perfectly balanced shards, each filled with
+  // 100 entries. So cacheSize is 1000. When cleaning 10%, after shard
+  // 0 we still need to processs 900 entries, spread out of 9
+  // shards. So cacheSize becomes 900, and toTrim 90, since we cleaned
+  // 10 items from shard 0. Our fraction remains 10%. For the last
+  // shard, we would end up with cacheSize 100, and to clean 10.
+  //
+  // When the balance is not perfect, e.g. shard 0 has 54 entries, we
+  // would clean 5 entries due to rounding, and for the remaning
```
rgacogne, line 209:
```suggestion
  // would clean 5 entries due to rounding, and for the remaining
```

pdns/cachecleaner.hh
omoerbeek: See my comment above

rgacogne: > Pushed a fixed version, with more explanation on the strategy used. One point to consider: in the first loop, we might ignore `toTrim` and just hunt for expired entries, using the `toScan` bound. That means that under pressure, we prefer to evict more expired entries and will fall back to the second loop in less cases. It might increase imbalance, but that _should_ not be an issue: we are removing stale items only.

I don't think it will increase imbalance, as we will scan the exact same number of entries for expired ones in every shard, since we do not take into account how many entries were removed so far for `toScan`. And if we do not remove enough entries that way, the second loop will nuke entries based on what is remaining at that point, so the balance should still be correct.

> Another thing to ponder:
> 
> Even while "data" record sets of a zone are retrieved and placed in the back of the expiry queue, the infra (NS + glue) record sets for the zone are not, as they are not touched when just checking the known "data" record sets of a zone. So infra record sets tend to travel to the front of the expiry queue, even when their data records are very popular. Only an expired data record set will trigger a retrieval of the infra record set from the cache so that it will be placed in the back of the expiry queue. So when the cache is under pressure it might clean more infra records than you would think.
> 
> This might be a case of "this is just the way it is", but IMO this is another reason we want to consider a separate cache fo infra records, with a different eviction policy.

My feeling is that it is working as expected: these infra records move to the front of the queue because we do not need them often, and thus we prefer to keep other records in the cache instead.

omoerbeek: Yes, but the potential impact of evicting infra records is higher compared to data records, as a cache miss of a data record that also has missing infra records induces more effort compared to just a missing data record.

omoerbeek: > I don't think it will increase imbalance, as we will scan the exact same number of entries for expired ones in every shard, since we do not take into account how many entries were removed so far for `toScan`. And if we do not remove enough entries that way, the second loop will nuke entries based on what is remaining at that point, so the balance should still be correct.

Right, as we take into account the size if the shard and we scan a fraction of t, not a fixed number as we used to.

I'll push a commit to that effect.

rgacogne: > Yes, but the potential impact of evicting infra records is higher compared to data records, as a cache miss of a data record that also has missing infra records induces more effort compared to just a missing data record.

That's certainly true in some cases, but less so in others, for example because we do not DNSSEC-validate infra records unless asked to by a direct query. I'm currently not convinced a complex logic to keep the useful infra records in the cache would be worth it, but I might be wrong.

omoerbeek:
`test_RecursorCache_ExpungingExpiredEntries` is not happy with the more eager evicting (it assumes things no longer true), looking into it.

rgacogne (approved):
Code looks good to me, nice work!

PR #12378, opened 2023-01-04
Update the pgp keys listed in docs

omoerbeek:
### Short description

Remove Bert and Pieter, add Remi and Otto
Same order as mentioned key file.

### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code


rgacogne:
Following the principle of least privileges, I don't think my key needs to be allowed to sign tarballs for the auth and the rec at this point.
For Bert and Pieter, perhaps we should add a paragraph explaining which past releases were signed with their keys, so it's still possible to validate old releases?

omoerbeek:
> Following the principle of least privileges, I don't think my key needs to be allowed to sign tarballs for the auth and the rec at this point. For Bert and Pieter, perhaps we should add a paragraph explaining which past releases were signed with their keys, so it's still possible to validate old releases?

I was listing the keys in the current published in https://www.powerdns.com/powerdns-keyblock.asc, which includes your key. I can leave your key out, but the keyblock file will still contain it.

I will push an update that lists Bert's and Pieter's keys as only valid for old releases.

rgacogne:
Ah, right, I'll have to check if my key has a good reason to be in there, then :)

omoerbeek:
> Ah, right, I'll have to check if my key has a good reason to be in there, then :)

One scenario: an emergency release of rec or auth and nobody of the other three signers available. 
dnsdist has its own list and keyblock file which both need updates as well.

rgacogne:
Alright, let's keep it that way then!

omoerbeek:
> Alright, let's keep it that way then!

Also because of: https://github.com/PowerDNS/pdns/security/policy

rgacogne (approved):

PR #12354, opened 2022-12-21
dnsdist: Move DNSQuestion's EDNS options to a unique pointer

rgacogne:
### Checklist

I have:
- [x] read the [CONTRIBUTING.md](https://github.com/PowerDNS/pdns/blob/master/CONTRIBUTING.md) document
- [x] compiled this code
- [x] tested this code
- [ ] included documentation (including possible behaviour changes)
- [ ] documented the code
- [ ] added or modified regression test(s)
- [ ] added or modified unit test(s)

PR #12383, opened 2023-01-06
dnsdist: Stronger guarantees against data race in the UDP path 

rgacogne:
### Short description

This pull request replaces the existing mechanism to ensure that UDP states can be safely accessed by several threads with a simpler, more robust one. This is possible now that we introduced the `InternalQueryState`, splitting the data from `IDState`.

<strike>Note that this pull request builds on https://github.com/PowerDNS/pdns/pull/12354 and will have to be rebased.</strike>

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/pdns/dnsdist-idstate.hh
+++ b/pdns/dnsdist-idstate.hh
@@ -160,7 +160,7 @@ struct IDState
   }
 
   IDState(const IDState& orig) = delete;
-  IDState(IDState&& rhs) noexcept:
+  IDState(IDState&& rhs) noexcept :
     internal(std::move(rhs.internal))
   {
     inUse.store(rhs.inUse.load());
--- a/pdns/dnsdist-idstate.hh
+++ b/pdns/dnsdist-idstate.hh
@@ -160,14 +160,13 @@ struct IDState
   }
 
   IDState(const IDState& orig) = delete;
-  IDState(IDState&& rhs)
+  IDState(IDState&& rhs) noexcept: internal(std::move(rhs.internal))
   {
     inUse.store(rhs.inUse.load());
     age.store(rhs.age.load());
-    internal = std::move(rhs.internal);
   }
 
-  IDState& operator=(IDState&& rhs)
+  IDState& operator=(IDState&& rhs) noexcept
   {
     inUse.store(rhs.inUse.load());
     age.store(rhs.age.load());
@@ -177,7 +176,7 @@ struct IDState
 
   bool isInUse() const
   {
-    return inUse == true;
+    return inUse;
   }
 
   /* For performance reasons we don't want to use a lock here, but that means
--- a/pdns/dnsdist-idstate.hh
+++ b/pdns/dnsdist-idstate.hh
@@ -162,73 +162,25 @@ struct IDState
   IDState(const IDState& orig) = delete;
   IDState(IDState&& rhs)
   {
-    if (rhs.isInUse()) {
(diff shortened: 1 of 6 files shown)


pdns/dnsdist-idstate.hh
```diff
@@ -157,73 +157,25 @@ struct IDState
...
-      throw std::runtime_error("Trying to move an in-use IDState");
-    }
-#ifdef __SANITIZE_THREAD__
+    inUse.store(rhs.inUse.load());
     age.store(rhs.age.load());
-#else
-    age = rhs.age;
-#endif
-
     internal = std::move(rhs.internal);
```
omoerbeek, line 169: clang-tidy prefers to move this assignment to a member init and to mark this op noexcept

pdns/dnsdist-idstate.hh
```diff
@@ -157,73 +157,25 @@ struct IDState
...
-    auto currentGeneration = generation++;
-    return markAsUsed(currentGeneration);
-  }
-
-  /* mark as used no matter what, return true if the state was in use before */
-  bool markAsUsed(int64_t currentGeneration)
-  {
-    int64_t oldUsage = usageIndicator.exchange(currentGeneration);
-    return oldUsage != unusedIndicator;
+    return inUse == true;
```
omoerbeek, line 175:
```suggestion
    return inUse;
```

pdns/dnsdist-idstate.hh
```diff
@@ -241,26 +193,52 @@ struct IDState
...
+  {
+    bool expected = false;
+    if (locked.compare_exchange_strong(expected, true)) {
+      return std::optional<StateGuard>(*this);
+    }
+    return std::nullopt;
+  }
+
+  void release()
+  {
```
omoerbeek, line 236: It is not clear to me (yet) if a `release` could happen for the non-locked case. If that would indicate an error, it might be good to catch that.

pdns/dnsdistdist/dnsdist-backend.cc
```diff
@@ -386,110 +383,170 @@ void DownstreamState::handleUDPTimeouts()
...
+    while (true);
   }
-  else {
-    if (stateId >= idStates.size()) {
-      return nullptr;
+
+  do {
+    IDState* ids = nullptr;
+    uint16_t selectedID = (idOffset++) % idStates.size();
+    ids = &idStates[selectedID];
```
omoerbeek, line 452: Avoid pointer
```suggestion
    IDState& ids = idStates[selectedID];
```

omoerbeek (approved):
While I agree that this is more simple than the original, it still is quite complex. Given that, I could not spot an error. I have some nits though.

It would also be nice to add some comments that when `randomizeIDs` is true, the locking mechanism of individual IDState object is not used (if I'm understanding things correctly). Maybe even a few consistency checks are in place with respect to that.

rgacogne:
Thanks Otto! I applied your suggestions and rebased on master to (hopefully) fix the CI.
I agree this is unfortunately still quite complex, mostly because we need to deal with two very different cases:
- dnsdist as a reverse proxy with a trusted path to the backend, allowing us to optimize for performance
- dnsdist as a proxy, or a reverse proxy with an untrusted path to the backend, where we need to do full randomization

We could always use the second option, but it has a very noticeable performance impact, so at least until we find a better solution I have kept the two options alive despite the added complexity.

PR #12386, opened 2023-01-06
dnsdist: Merge the 'main' and 'client' DoH threads in single acceptor mode

rgacogne:
### Short description

When we are in "single acceptor thread" mode, merge the 'main' and 'client' DoH threads into a single one. We use separate threads to reduce the separate the handling of the HTTP/2 traffic from the DNS handling, to reduce latency, but that does not really make sense on small devices with a single, limited CPU core. On these we prefer using as few threads as possible to reduce the context switches and the memory usage.

<strike>Note that this pull request builds on https://github.com/PowerDNS/pdns/pull/12383 and will have to be rebased.</strike>

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/pdns/dnsdistdist/doh.cc
+++ b/pdns/dnsdistdist/doh.cc
@@ -172,6 +172,7 @@ struct DOHServerConfig
   DOHServerConfig(uint32_t idleTimeout, uint32_t internalPipeBufferSize): accept_ctx(std::make_shared<DOHAcceptContext>())
   {
     int fd[2];
+#ifndef USE_SINGLE_ACCEPTOR_THREAD
     if (pipe(fd) < 0) {
       unixDie("Creating a pipe for DNS over HTTPS");
     }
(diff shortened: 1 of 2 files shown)


chbruyand (approved):

pdns/dnsdistdist/doh.cc
```diff
@@ -567,9 +590,27 @@ static void processDOHQuery(DOHUnitUniquePtr&& unit)
...
       du->status_code = 500;
-      sendDoHUnitToTheMainThread(std::move(du), "DoH killed in flight");
+      handleImmediateResponse(std::move(du), "DoH killed in flight");
       return;
     }
+
+    {
+      // if there was no EDNS, we add it with a large buffer size
+      // so we can use UDP to talk to the backend.
+      auto dh = const_cast<struct dnsheader*>(reinterpret_cast<const struct dnsheader*>(du->query.data()));
```
chbruyand, line 600: Just to be sure, do we already know that `data()` is at least the size of a `dnsheader` at this point (otherwise we would not hit that point) ?

rgacogne: Yes, otherwise it would have been discarded at the beginning of `doh_dispatch_query`.

PR #12387, opened 2023-01-06
dnsdist: Handle out-of-memory exceptions in the UDP receiver thread

rgacogne:
### Short description

Most exceptions are handled by `handleOnePacket()`, but we might be out of memory (`std::bad_alloc`) in which case we DO NOT want to log (as it would trigger another memory allocation attempt that might throw as well) but wait a bit (one millisecond) and then try to recover.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/pdns/dnsdist.cc
+++ b/pdns/dnsdist.cc
@@ -1791,10 +1791,11 @@ static void udpClientThread(std::vector<ClientState*> states)
 
           processUDPQuery(*param.cs, holders, &msgh, remote, dest, packet, nullptr, nullptr, nullptr, nullptr);
         }
-        catch (const std::exception& e) {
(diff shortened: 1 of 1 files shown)


omoerbeek (commented):
I was a bit puzzled by the while loop calling `handleOnePacket` not having a try-catch block and the (dynamic) nesting of try-catch blocks in the else branch.
Is it an idea to not have a `try/catch` block in `handleOnePacket`, but have it in the two call spots? IMO it would make the structure more clear as the two spots calling `handleOnePacket` would have the same structure.

rgacogne:
Done!

omoerbeek (approved):

Issue #10690, opened 2021-09-04
Allow falling back to using authoritatives if forwarders respond with servfail 


ctheune:
- Program: Recursor
 - Issue type: Feature request

### Short description

I'd like to be able to instruct pdns_resolver to do a regular authoritative lookup "dance" if none of the (recursive) forwarders respond. 

### Usecase

In our setups (data center for VMs as well as our own home networks) we typically deploy resolver configuration by running a local recursor and trying to delegate upstream to services like Quad9.

In the past we have been burned by either operators of authoritative servers accidentally blocking our networks due to typos and not having NOCs reachable around the clock, but we have also been burned by caching services like Quad9 having issues.

After a bit of back and forth we are currently running bind servers that forward with the "forward first" strategy: if the forwarder responds with a servfail the local resolver will fall back trying to talk to the authoritative directly.

### Description

I'm currently debugging some issues in my home network where I'm trying to get a similar configuration running with pdns_recursor but I think the knob for "when I run out of forwarders, just do the regular dance to talk to the authoritative" is missing. I _think_ I could cobble this together by running a second pdns_recursor as a second forwarder, but this seems dirty. 

Looking at the archive I found very little discussion about this except a short exchange back in 2013 and the issue tracker was similarly quiet.

I might be completely missing something obvious, but maybe not ... :)

ctheune:
Here's an example of a forward that ended up in a SERVFAIL for the client because the recursive forwarder gave us a servfail:

```
Sep 04 07:27:34 cecil pdns_recursor[153731]: 3 [18/1] question for 'ligabank.de|A' from 192.168.2.228:53395
Sep 04 07:27:35 cecil pdns_recursor[153731]: : no TA found for 'ligabank.de' among 1
Sep 04 07:27:35 cecil pdns_recursor[153731]: : no TA found for 'de' among 1
Sep 04 07:27:35 cecil pdns_recursor[153731]: : got TA for '.'
Sep 04 07:27:35 cecil pdns_recursor[153731]: ligabank.de: Wants DNSSEC processing, auth data in query for A
Sep 04 07:27:35 cecil pdns_recursor[153731]: ligabank.de: No cache hit for 'ligabank.de|A', trying to find an appropriate NS record
Sep 04 07:27:35 cecil pdns_recursor[153731]: ligabank.de: Cache consultations done, have 1 NS to contact
[...]

PR #12388, opened 2023-01-06
dnsdist: Implement async processing of queries and responses

rgacogne:
### Short description

This PR implements the ability to process queries and responses in an asynchronous way, suspending them to continue processing other queries and responses, while we are waiting for an external event to occur.
It also implements the ability to restart a query based on the content of the response, for example to re-try sending the initial query to a different pool of servers if the initial attempts yields a `ServFail` or `Refused` response code.

<strike>The documentation needs to be updated before merging this PR.</strike>

<strike>Note that this pull request builds on https://github.com/PowerDNS/pdns/pull/12383 and will have to be rebased.</strike>

Closes https://github.com/PowerDNS/pdns/issues/10690

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code


phonedph1:
I might have missed it, but are `dq:setRestartable()` and `dq:restart()` in the docs somewhere? The test case for this looks like it will do some pretty nice stuff - I almost feel like this test case should be copied into the docs somewhere for how simplistic to configure but powerful it is in use.

rgacogne:
> I might have missed it, but are `dq:setRestartable()` and `dq:restart()` in the docs somewhere? The test case for this looks like it will do some pretty nice stuff - I almost feel like this test case should be copied into the docs somewhere for how simplistic to configure but powerful it is in use.

You did not miss it, I did not add that part of the documentation yet but I will do that soon!

rgacogne:
> You did not miss it, I did not add that part of the documentation yet but I will do that soon!

Done so now, thanks :-)

rgacogne:
Rebased to fix conflicts.

rgacogne:
Rebased to fix (new) conflicts.

pdns/dnsdistdist/docs/reference/dq.rst
```diff
@@ -350,6 +384,35 @@ DNSResponse object
...
+          -- discard the received response and
+          -- restart the processing of the query
+          dr:restart()
+        end
+        return DNSResponseAction.None
+      end
+      addAction(AllRule(), LuaAction(makeQueryRestartable))
+      addResponseAction(AllRule(), LuaResponseAction(restartOnServFail))
+
+    For this function to be usable, the query should have been made restartable first, via :func:`DNSQuestion:setRestartable`.
```
omoerbeek, line 414: Move this line up, after "Discard..."

pdns/dnsdistdist/docs/reference/dq.rst
```diff
@@ -217,6 +223,14 @@ This state can be modified from the various hooks.
 
     :param string reason: An optional string describing the reason why this trap was sent
 
+  .. method:: DNSQuestion:setContent(data)
+
+    .. versionadded:: 1.8.0
+
+    Replace the DNS payload of the query with the supplied data, and adjusts the ID in the DNS header to the existing query.
```
omoerbeek, line 230: Example would be nice, especially since it is not 100% clear if the payload should include the header.

pdns/dnsdistdist/docs/reference/dq.rst
```diff
@@ -438,3 +507,56 @@ EDNSOptionView object
...
+  .. method:: AsynchronousObject:getDR() -> DNSResponse
+
+    Return a DNSResponse object for the suspended object.
+
+  .. method:: AsynchronousObject:resume() -> bool
+
+    Resume the processing of the suspended object.
+    For a question, it means first checking whether it was turned into a response,
+    and sending the response out it it was. Otherwise do a cache-lookup, select a
+    backend and send the query to the backend.
```
omoerbeek, line 540: Explains what happens on cache-miss. What happens on cache-hit?

pdns/dnsdistdist/docs/reference/dq.rst
```diff
@@ -303,6 +325,18 @@ This state can be modified from the various hooks.
     :param string raw: The raw string to be spoofed, e.g. `"\\192\\000\\002\\001"`.
     :param table raws: The raw strings to be spoofed, e.g. `{ "\\192\\000\\002\\001", "\\192\\000\\002\\002" }`.
 
+  .. method:: DNSQuestion:suspend(asyncID, queryID, timeoutMS) -> bool
+
+    .. versionadded:: 1.8.0
+
+    Suspend the processing for the current query, making it asynchronous. The query is then placed into memory, in a map called the Asynchronous Holder, until it is either resumed or the supplied timeout kicks in. The object is stored under a key composed of the tuple (`asyncID`, `queryID`) which is needed to retrieve it later, which can be done via :func:`getAsynchronousObject`.
```
omoerbeek, line 332: Definition of legal range of both IDs is missing.

Also I woulld have expect the query to be identified by an ID or IDs created by the Asynchronous Holder. If I understand correctly, the burden of creating a unique asyncID, queryID pair is now upon the caller of `suspend`?

omoerbeek (commented):
Started with reading the docs, notes a few things. 

To be continued.

pdns/dnsdistdist/dnsdist-lua-ffi.cc
```diff
@@ -75,6 +84,15 @@ size_t dnsdist_ffi_dnsquestion_get_mac_addr(const dnsdist_ffi_dnsquestion_t* dq,
   return 6;
 }
 
+uint64_t dnsdist_ffi_dnsquestion_get_elapsed_us(const dnsdist_ffi_dnsquestion_t* dq)
+{
+  if (dq == nullptr) {
+    return 0;
+  }
+
+  return dq->dq->ids.queryRealTime.udiff();
```
omoerbeek, line 93: you might want to call round() instead of the implicit truncate

pdns/dnsdist.hh
```diff
@@ -69,6 +73,7 @@ struct DNSQuestion
   DNSQuestion(const DNSQuestion&) = delete;
   DNSQuestion& operator=(const DNSQuestion&) = delete;
   DNSQuestion(DNSQuestion&&) = default;
+  virtual ~DNSQuestion() = default;
```
omoerbeek, line 76: Is this virtual dt needed? Only once class derives from `DNSQuestion` and it does not override the dt (and adds no fields either). Though I see it will make `dnsdist_ffi_dnsresponse_set_async` more complicated, as it wants to distinguish the two one way or another.

pdns/dnsdistdist/test-dnsdistasync.cc
```diff
@@ -0,0 +1,165 @@
...
+
+BOOST_AUTO_TEST_CASE(test_AddingExpiredEvent)
+{
+  auto holder = std::make_unique<dnsdist::AsynchronousHolder>(false);
+  uint16_t asyncID = 1;
+  uint16_t queryID = 42;
+  struct timeval ttd;
+  gettimeofday(&ttd, nullptr);
+  // timeout was 10 ms ago, for some reason (long processing time, CPU starvation...)
+  ttd.tv_usec -= 10000;
```
omoerbeek, line 145:
```suggestion
  const timeval sub{0, 10000};
  timersub(&ttd, &sub, &ttd);
```

omoerbeek (commented):
Some local nits noted. Still have to try to grasp the whole thing.

pdns/dnsdistdist/docs/reference/dq.rst
rgacogne: I'll clear up the range of both IDs, nice catch.
As for making the burden of creating unique pairs on the caller, I believe this is the most flexible way, as the Asynchronous holder has no clue about the logic behind the objects that are stored, while the caller does.

rgacogne:
I pushed a few commits addressing your remarks, thanks!

pdns/dnsdistdist/test-dnsdistasync.cc
```diff
@@ -84,7 +84,8 @@ BOOST_AUTO_TEST_CASE(test_Basic)
     struct timeval ttd;
     gettimeofday(&ttd, nullptr);
     // timeout in 100 ms
-    ttd.tv_usec += 100000;
+    const timeval sub{0, 100000};
+    timersub(&ttd, &sub, &ttd);
```
omoerbeek, line 88: A `+=` changed into a `timersub` here

omoerbeek: There are a few other instances of timeval computations left.

omoerbeek (changes requested):

pdns/dnsdistdist/test-dnsdistasync.cc
rgacogne: :+1:

omoerbeek:
There is still something fishy going on on OpenBSD `testrunner` (both amd64 and arm64):
```
 16536 testrunner CALL  nanosleep(0x7f7ffffcad18,0)
 16536 testrunner CALL  kevent(5,0,0,0x488ede80e00,1,0x488c5f69100)
 16536 testrunner STRU  struct timespec { 0.200000000 }
 16536 testrunner STRU  struct timespec { 0.-01000000 }
 16536 testrunner RET   kevent -1 errno 22 Invalid argument
```

pdns/lock.hh
```diff
@@ -288,7 +288,7 @@ public:
     return LockGuardedHolder<T>(d_value, d_mutex);
   }
 
-  LockGuardedHolder<const T> read_only_lock() const
```
omoerbeek, line 291: Would it make sense to have both a `const` and non-`const` variant here? e.g.
```
  LockGuardedHolder<const T> read_only_lock() const
  {
    return LockGuardedHolder<const T>(d_value, d_mutex);
  }

  LockGuardedHolder<T> read_only_lock()
  {
    return LockGuardedHolder<T>(d_value, d_mutex);
  }
```

omoerbeek (changes requested):
One nit and and a bug, described in comments.

rgacogne:
> Looks like the default value -1 of timeout in `mainThread` is used. running with a default value of 1 works better.

When we have a bigger problem than this PR, because mplexer.hh documents -1 as 'will wait forever', and this is the behaviour I want, not being woken up every second for no reason.. I guess it has always been broken with `kqueue`, I'll investigate.

pdns/lock.hh
rgacogne: Hmm, why would we want a writable version of the data with a read-only lock?

omoerbeek:
> > Looks like the default value -1 of timeout in `mainThread` is used. running with a default value of 1 works better.
> 
> When we have a bigger problem than this PR, because mplexer.hh documents -1 as 'will wait forever', and this is the behaviour I want, not being woken up every second for no reason.. I guess it has always been broken with `kqueue`, I'll investigate.

`kqueue` expects NULL ts for infinite timeout

rgacogne:
> BTW earlier I suggested to use `timeradd` and `timersub`. But now I see we also have
> 
> ```
> void normalizeTV(struct timeval& tv)
> const struct timeval operator+(const struct timeval& lhs, const struct timeval& rhs)
> const struct timeval operator-(const struct timeval& lhs, const struct timeval& rhs)
> ```
> 
> in misc.cc (with a funny `const` return values). e.g. `suspendQuery` and `suspendResponse` can use normalizeTV. And the tests can use the operators. Sorry for not spotting these methods earlier.

Nice, thank you! I pushed that change now.

pdns/lock.hh
omoerbeek: Hmm, ok, I was confused. Of course the method should not be `const`, and it does correcly return a `const T` holder.

omoerbeek:
Tests OK now on both macOS and OpenBSD.

omoerbeek (approved):
I cannot say I checked every detail, but this reads good now and after the latest changes I've run tests on macOS, OpenBSD and debian bullseye successfully

PR #12395, opened 2023-01-08
When the stale function is triggered, wrong data can be returned from negcache and record cache

zjs604381586:
### Short description

1. When the parsing request triggers the stale function, first of all, when doing a cache check, if there is record information in the negcache, it will return the nxdomain data in the negcache, but there is a parsing result set in the recursor_cache, such as A records. Therefore, it is necessary to clear the record of the domain name in the negcache when the authority returns the resolution record and writes it to the recursor_cache
2. When a domain name has a cname, the dns request is resolved once, and then the cname of the domain name is deleted from the authority, and then the second dns request is resolved and returned to nxdomain. When performing subsequent dns resolution on the domain name, if the stale function is triggered, the cname will be returned, and the result of the cname will be returned instead of the latest nxdomain data. Looking at the code, it is found that in the doCNAMECacheCheck function, only the recursor_cache is searched, and no check is made from the negcache, so the old cname data will be returned. Therefore, it is necessary to return to nxdomain in the authority, do the type verification of cname when adding negcache, and do the cleaning action in recursor_cache
### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/recursordist/negcache.cc
+++ b/pdns/recursordist/negcache.cc
@@ -44,6 +44,25 @@ size_t NegCache::size() const
   return count;
 }
 
+void NegCache::purge(const DNSName& qname, QType qtype)
+{
+  auto& mc = getMap(qname);
+  auto content = mc.lock();
+  auto& idx = content->d_map.get<NegCacheEntry>();
+
+  auto range = idx.equal_range(qname);
+  auto ni = range.first;
+
+  while (ni != range.second) {
+    // We have an entry
+    if ((ni->d_qtype == QType::ENT) || ni->d_qtype == qtype) {
+        ni = idx.erase(ni);
+    }
+    else
+        ++ni;
+  }
+}
+
 /*!
  * Set ne to the NegCacheEntry for the last label in qname and return true if there
  * was one.
--- a/pdns/recursordist/negcache.cc
+++ b/pdns/recursordist/negcache.cc
@@ -44,25 +44,6 @@ size_t NegCache::size() const
   return count;
 }
 
-void NegCache::purge(const DNSName& qname, QType qtype)
-{
-  auto& mc = getMap(qname);
-  auto content = mc.lock();
-  auto& idx = content->d_map.get<NegCacheEntry>();
-
-  auto range = idx.equal_range(qname);
-  auto ni = range.first;
-
-  while (ni != range.second) {
-    // We have an entry
-    if ((ni->d_qtype == QType::ENT) || ni->d_qtype == qtype) {
-        ni = idx.erase(ni);
-    }
-    else
-        ++ni;
-  }
-}
-
 /*!
  * Set ne to the NegCacheEntry for the last label in qname and return true if there
  * was one.
@@ -257,6 +238,26 @@ size_t NegCache::wipe(const DNSName& name, bool subtree)
     i = content->d_map.erase(i);
     ret++;
     --map.d_entriesCount;
+  }
+  return ret;
+}
+
+size_t NegCache::wipe(const DNSName& qname, QType qtype)
+{
+  size_t ret = 0;
+  auto& map = getMap(qname);
+  auto content = map.lock();
+  auto range = content->d_map.equal_range(std::tie(qname));
+  auto i = range.first;
+  while (i != range.second) {
+    if (i->d_qtype == QType::ENT || i->d_qtype == qtype) {
+      i = content->d_map.erase(i);
+      ++ret;
+      --map.d_entriesCount;
+    }
+    else {
+      ++i;
+    }
   }
   return ret;
 }
--- a/pdns/recursordist/recursor_cache.cc
+++ b/pdns/recursordist/recursor_cache.cc
@@ -316,6 +316,20 @@ bool MemRecursorCache::entryMatches(MemRecursorCache::OrderedTagIterator_t& entr
   return match;
 }
 
+void MemRecursorCache::purge(const DNSName& qname, QType qt)
+{
+   if(qt == QType::CNAME)
+   {
+        auto& mc = getMap(qname);
+        auto map = mc.lock();
+        auto key = std::make_tuple(qname, qt, boost::none, Netmask());
+        auto entry = map->d_map.find(key);
+        if (entry != map->d_map.end()) {
+            map->d_map.erase(entry);
+        }
+   }  
+}
+
 // Fake a cache miss if more than refreshTTLPerc of the original TTL has passed
 time_t MemRecursorCache::fakeTTD(MemRecursorCache::OrderedTagIterator_t& entry, const DNSName& qname, QType qtype, time_t ret, time_t now, uint32_t origTTL, bool refresh)
 {
--- a/pdns/recursordist/recursor_cache.cc
+++ b/pdns/recursordist/recursor_cache.cc
@@ -316,20 +316,6 @@ bool MemRecursorCache::entryMatches(MemRecursorCache::OrderedTagIterator_t& entr
   return match;
 }
 
-void MemRecursorCache::purge(const DNSName& qname, QType qt)
-{
-   if(qt == QType::CNAME)
-   {
-        auto& mc = getMap(qname);
-        auto map = mc.lock();
-        auto key = std::make_tuple(qname, qt, boost::none, Netmask());
-        auto entry = map->d_map.find(key);
-        if (entry != map->d_map.end()) {
-            map->d_map.erase(entry);
-        }
-   }  
-}
-
 // Fake a cache miss if more than refreshTTLPerc of the original TTL has passed
 time_t MemRecursorCache::fakeTTD(MemRecursorCache::OrderedTagIterator_t& entry, const DNSName& qname, QType qtype, time_t ret, time_t now, uint32_t origTTL, bool refresh)
 {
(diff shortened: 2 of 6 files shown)


omoerbeek:
Thanks you for the PR!

First a general comment: we prefer rebasing to merge commits in PRs.

As for the code: the NegCache case is clear to me. The  CNAME case I need to look into to fully understand.
I do see a few problems with your implementation of the purge functions. I comment on the code involved.

I'll try to push an update to the PR soon.

pdns/recursordist/negcache.cc
```diff
@@ -86,6 +86,25 @@ bool NegCache::getRootNXTrust(const DNSName& qname, const struct timeval& now, N
...
+  auto& mc = getMap(qname);
+  auto content = mc.lock();
+  auto& idx = content->d_map.get<NegCacheEntry>();
+
+  auto range = idx.equal_range(qname);
+  auto ni = range.first;
+
+  while (ni != range.second) {
+    // We have an entry
+    if ((ni->d_qtype == QType::ENT) || ni->d_qtype == qtype) {
```
omoerbeek, line 100:
```suggestion
    if (ni->d_qtype == QType::ENT || ni->d_qtype == qtype) {
```

pdns/recursordist/negcache.cc
```diff
@@ -86,6 +86,25 @@ bool NegCache::getRootNXTrust(const DNSName& qname, const struct timeval& now, N
...
+        ni = idx.erase(ni);
+    }
```
omoerbeek, line 102: Add `--map.d_entriesCount;` otherwise the entries count will be off.

Alternatively, we may want to extend the existing `NegCache::wipe` function to deal with `qtype == ENT`

pdns/recursordist/recursor_cache.cc
```diff
@@ -316,6 +316,20 @@ bool MemRecursorCache::entryMatches(MemRecursorCache::OrderedTagIterator_t& entr
   return match;
 }
 
+void MemRecursorCache::purge(const DNSName& qname, QType qt)
```
omoerbeek, line 319: This new method is not needed, see `MemRecursorCache::doWipeCache`
(it also neglects to wipe the ECS data)

pdns/recursordist/syncres.cc
```diff
@@ -4803,6 +4805,9 @@ bool SyncRes::processRecords(const std::string& prefix, const DNSName& qname, co
       */
       if (newtarget.empty() && putInNegCache) {
         g_negCache->add(ne);
+        if(qtype == QType::CNAME){
+            g_recCache->purge(qname, qtype);
```
omoerbeek, line 4809:
```suggestion
            g_recCache->doWipeCache(qname, false, qtype);
```

omoerbeek (changes requested):

omoerbeek:
I force pushed a rebased version and then committed my suggested changes. I still have to look into the CNAME case to  understand it.

zjs604381586:
Case: When there is a cname for a domain name in the pdns-recursor cache, then delete the cname of the domain name from the authoritative dns server.

Step 1: When the TTL expires and then initiate a dns request for the domain name, nxdomain will be returned.

Step 2: If you make another request and the stale function is triggered at this time, the cname result will still be returned, which is not as expected.

Cause of the problem:
Because the code is in the doCNAMECacheCheck function, there is no check negcache, but just get recursor_cache.

omoerbeek:
Wondering if a negcache check in  `doCNAMECacheCheck` would be better.

zjs604381586:
At present, what I think is to delete the **recursor-cache** when cname nxdomain returns. If you check negcache in **doCNAMECacheCheck** every time, will it affect performance?

omoerbeek:
Sorry for the delay. I well continue working on this very soon.

pdns/recursordist/syncres.cc
```diff
@@ -4803,6 +4809,11 @@ bool SyncRes::processRecords(const std::string& prefix, const DNSName& qname, co
       */
       if (newtarget.empty() && putInNegCache) {
         g_negCache->add(ne);
+        // doCNAMECacheCheck() checks record cache and does not look into negcache. That means that and old record might be found if
```
mnordhoff, line 4812:
```suggestion
        // doCNAMECacheCheck() checks record cache and does not look into negcache. That means that an old record might be found if
```
Not important, but "and" -> "an" typo

pdns/recursordist/syncres.cc
```diff
@@ -4638,6 +4638,12 @@ RCode::rcodes_ SyncRes::updateCacheFromRecords(unsigned int depth, LWResult& lwr
         }
         g_recCache->replace(d_now.tv_sec, i->first.name, i->first.type, i->second.records, i->second.signatures, authorityRecs, i->first.type == QType::DS ? true : isAA, auth, i->first.place == DNSResourceRecord::ANSWER ? ednsmask : boost::none, d_routingTag, recordState, remoteIP, d_refresh);
 
+        // Delete potential negcache entry. When a record recovers with serve-stale the negcache entry can cause the wrong entry to
+        // served, as negcache entries are checked before record cache entries
```
rgacogne, line 4642:
```suggestion
        // be served, as negcache entries are checked before record cache entries
```

rgacogne (approved):
Code looks good, a few tests would be nice to have.

omoerbeek:
> Code looks good, a few tests would be nice to have.

Will do.

pdns/recursordist/test-syncres_cc10.cc
```diff
@@ -1331,6 +1331,149 @@ BOOST_AUTO_TEST_CASE(test_servestale_neg)
...
+      addRecordToLW(res, "pdns-public-ns1.powerdns.com.", QType::A, "192.0.2.2", DNSResourceRecord::ADDITIONAL, theTTL);
+      addRecordToLW(res, "pdns-public-ns1.powerdns.com.", QType::AAAA, "2001:DB8::2", DNSResourceRecord::ADDITIONAL, theTTL);
+      addRecordToLW(res, "pdns-public-ns2.powerdns.com.", QType::A, "192.0.2.3", DNSResourceRecord::ADDITIONAL, theTTL);
+      addRecordToLW(res, "pdns-public-ns2.powerdns.com.", QType::AAAA, "2001:DB8::3", DNSResourceRecord::ADDITIONAL, theTTL);
+      return LWResult::Result::Success;
+    }
+    else if (ip == ComboAddress("192.0.2.2:53") || ip == ComboAddress("192.0.2.3:53") || ip == ComboAddress("[2001:DB8::2]:53") || ip == ComboAddress("[2001:DB8::3]:53")) {
+      if (negLookup) {
+        setLWResult(res, 0, true, false, true);
+        addRecordToLW(res, "powerdns.com.", QType::SOA, "pdns-public-ns1.powerdns.com. pieter\\.lexis.powerdns.com. 2017032301 10800 3600 604800 60", DNSResourceRecord::AUTHORITY);
```
rgacogne, line 1379: It might be nicer to use `negTTL` explicitly here?

rgacogne (approved):

pdns/recursordist/test-syncres_cc10.cc
```diff
@@ -1474,6 +1473,161 @@ BOOST_AUTO_TEST_CASE(test_servestale_neg_to_available)
...
+
+  BOOST_REQUIRE_EQUAL(getTaskSize(), 2U);
+  auto task = taskQueuePop();
+  BOOST_CHECK(task.d_qname == target);
+  BOOST_CHECK_EQUAL(task.d_qtype, QType::CNAME);
+  task = taskQueuePop();
+  BOOST_CHECK(task.d_qname == DNSName("cname.powerdns.com"));
+  BOOST_CHECK_EQUAL(task.d_qtype, QType::A);
+
+  // Now simulate a succeeding task execution and NxDomain on expliit CNAME result becomes available
```
rgacogne, line 1604:
```suggestion
  // Now simulate a succeeding task execution and NxDomain on explicit CNAME result becomes available
```

pdns/recursordist/test-syncres_cc10.cc
```diff
@@ -1474,6 +1473,161 @@ BOOST_AUTO_TEST_CASE(test_servestale_neg_to_available)
...
+  cnameOK = false;
+  sr->setNow(timeval{now + 3 * (theTTL + 1), 0});
+  downServers.clear();
+  sr->setRefreshAlmostExpired(true);
+
+  ret.clear();
+  res = sr->beginResolve(target, QType(QType::CNAME), QClass::IN, ret);
```
rgacogne, line 1611: Did you really intend to directly ask for a `CNAME`, and no `A`?

omoerbeek: Yes, the refresh task does as well. I have made a note of that as it might be wrong/superfluous.

omoerbeek: Thinking about this a bit more, the CNAME case really needs more attention indeed. I'll wait with merging this.

omoerbeek: It's easy to show with `refresh-on-ttl-perc`: if you have a CNAME with a short TTL, a query that hits it will use the cached qtype of the record, so an explicit CNAME query will be pushed.

omoerbeek: RFC 1035 states that asking for a CNAME explicitly is not wrong (3.6.1 and .2) and has a well defined meaning. So my immediate concern is solved.

rgacogne: That works for me, I just wanted to be sure that you really meant to test that case!

zjs604381586:
nice change， thanks

Issue #10600, opened 2021-07-20
“refresh almost expired” feature for dnsdist


paddg:
- Program: dnsdist 
 - Issue type: Feature request

### Short description

It would be nice if one can profit from Recursor's “refresh almost expired” feature even if there is a caching dnsdist in front of the Recursor.

### Usecase

We have dnsdist processes in front of our Recursor processes, which is great! Unfortunately the dnsdist cache makes the “refresh almost expired” feature of the Recursor completely ineffective.

### Description

A solution, I could imagine is, to give the dnsdist packet cache the same feature as  the Recursor have: "refresh-on-ttl-perc".

But you guys may have  a better solution.

rasgard:
Hi @rgacogne 
Regarding the function added with https://github.com/PowerDNS/pdns/pull/12400, this may not be very helpful if you have more than one resolver backend. There should be a real 'refresh almost expired' feature.

paddg:
@rasgard, you could use `setServerPolicy(whashed)`, to ensure that the "correct" backend is addressed.

rgacogne:
I'm not sure I follow your reasoning. I'm guessing, but it would be better if you could elaborate, that you mean we will end up requesting the almost expired record from only one resolver, and it might not be the one that will be queried for the next query coming in for that record, which would result in a miss. How do you imagine a "real" refresh almost expired feature in dnsdist would fix this?

rasgard:
yes I expect that some of the querys will not be answerd by the same backend and result in longer responses but i have to check setServerPolicy(whashed) thanks

rasgard:
Sorry, I should have read the manual again :)  I have observed this problem since 2021....
thank you @rgacogne @paddg

rgacogne:
You might also want to have a look at `chashed`, which delivers consistent hashing even when servers are added or removed: https://dnsdist.org/guides/serverselection.html#chashed

rgacogne:
I'm closing this feature request for now, as I don't think we will get a better solution than the one implemented in #12400. I'm of course happy to re-open if someone disagrees and suggests a better solution, especially if it comes with a pull request ;-)

PR #12400, opened 2023-01-09
dnsdist: Add SetReducedTTLResponseAction

rgacogne:
### Short description

Reduce the TTL of records in a response to a percentage of the original TTL. For example, passing 50 means that the original TTL will be cut in half.
One use-case for this action is when dnsdist is in front of a recursor with the "refresh almost expired" feature enabled: dnsdist will then cache the packet for shorter duration than the recursor, allowing the recursor to re-fetch before the record has expired.

Related to https://github.com/PowerDNS/pdns/issues/10600.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/pdns/dnsdist-lua-actions.cc
+++ b/pdns/dnsdist-lua-actions.cc
@@ -2001,6 +2001,32 @@ private:
   uint8_t d_type;
 };
 
+class SetReducedTTLResponseAction : public DNSResponseAction, public boost::noncopyable
+{
+public:
+  // this action does not stop the processing
+  SetReducedTTLResponseAction(uint8_t percentage) : d_ratio(percentage / 100.0)
+  {
+  }
+
+  DNSResponseAction::Action operator()(DNSResponse* dr, std::string* ruleresult) const override
+  {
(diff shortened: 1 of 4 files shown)


pdns/dnsdist-lua-actions.cc
```diff
@@ -2001,6 +2001,32 @@ class SetAdditionalProxyProtocolValueAction : public DNSAction
...
+    auto visitor = [&](uint8_t section, uint16_t qclass, uint16_t qtype, uint32_t ttl) {
+      return ttl * d_ratio;
+    };
+    editDNSPacketTTL(reinterpret_cast<char *>(dr->getMutableData().data()), dr->getData().size(), visitor);
+    return DNSResponseAction::Action::None;
+  }
+
+  std::string toString() const override
+  {
+    return "reduce ttl to " + std::to_string(d_ratio * 100) + " of its value";
```
chbruyand, line 2023:
```suggestion
    return "reduce ttl to " + std::to_string(d_ratio * 100) + " percent of its value";
```

chbruyand (approved):

omoerbeek (approved):

PR #12401, opened 2023-01-09
Openssl 3 prep work

fredmorcos:
### Short description
- Relax some `clang-tidy` warnings
- Format some files (and remove them from `.not-formatted`)
- Cleanup and fix `clang-tidy` lints in various files
- Add `pdns::OpenSSL::error(...)` for throwing openssl errors with as much info as possible
- Adds an extra check for `getPubKeyHash()` in `test-signers`

### Checklist
I have:
- [X] compiled this code
- [X] tested this code
- [X] documented the code


pdns/misc.cc
```diff
@@ -225,6 +227,39 @@ auto pdns::getMessageFromErrno(const int errnum) -> std::string
...
+auto pdns::OpenSSL::error(const std::string& errorMessage) -> std::runtime_error
+{
+  unsigned long errorCode = 0;
+  auto fullErrorMessage{errorMessage};
+#if OPENSSL_VERSION_MAJOR >= 3
+  const char* filename = nullptr;
+  const char* functionName = nullptr;
+  int lineNumber = 0;
+  while ((errorCode = ERR_get_error_all(&filename, &lineNumber, &functionName, nullptr, nullptr)) != 0) {
+    fullErrorMessage += std::string(": ") + ERR_error_string(errorCode, nullptr);
```
omoerbeek, line 236:
```
    ERR_error_string() generates a human-readable string representing the
     error code e and places it in buf.  buf must be at least 256 bytes long.
     If buf is NULL, the error string is placed in a static buffer.  Note that
     this function is not thread-safe and does no checks on the size of the
     buffer; use ERR_error_string_n() instead.
```

omoerbeek: I prefer a thread safe solution.

pdns/ipcipher.cc
```diff
@@ -11,92 +11,102 @@ int PKCS5_PBKDF2_HMAC_SHA1(const char *pass, int passlen,
...
+  ipcrypt_decrypt((unsigned char*)&ret.sin4.sin_addr.s_addr,
+                  (const unsigned char*)&ca.sin4.sin_addr.s_addr,
+                  (const unsigned char*)key.c_str());
+
   return ret;
 }
 
-
-static ComboAddress encryptCA6(const ComboAddress& ca, const std::string &key)
+static ComboAddress encryptCA6(const ComboAddress& comboAddress, const std::string& key)
```
mind04, line 54: Please use a less confusing variable name here and in a few places below. ComboAddres vs comboAdress is not improving code readability.

fredmorcos: Agreed. Is it worse than `ca`, though? Would `address` be okay?

pdns/misc.cc
fredmorcos: I opted for the following because `ERR_error_string_n()` doesn't really tell me how to find out the required buffer length.

```
    const auto* lib = ERR_lib_error_string(errorCode);
    const auto* reason = ERR_reason_error_string(errorCode);
    fullErrorMessage += std::string(": ") + std::to_string(errorCode) + ":" + lib + "::" + reason;
```

fredmorcos:
This should be ready now.

pdns/misc.cc
omoerbeek: The !OpenSSL 3 code still uses `ERR_error_string`

fredmorcos: Ah... the editor shows it as comments and it's not easy to notice.

pdns/misc.cc
```diff
@@ -225,6 +229,45 @@ auto pdns::getMessageFromErrno(const int errnum) -> std::string
...
+    }
+    if (functionName != nullptr) {
+      fullErrorMessage += std::string(" - ") + functionName;
+    }
+  }
+#else
+  while ((errorCode = ERR_get_error()) != 0) {
+    const auto* lib = ERR_lib_error_string(errorCode);
+    const auto* reason = ERR_reason_error_string(errorCode);
+    fullErrorMessage += std::string(": ") + std::to_string(errorCode) + ":" + lib + "::" + reason;
```
rgacogne, line 259: Two questions:
- do we need to be careful about `lib` or `reason` being `nullptr`? The documentation in 1.1.1l states that it might happen: `ERR_lib_error_string(), ERR_func_error_string() and ERR_reason_error_string() return the strings, and B<NULL> if none is registered for the error code.`
- do we want `ERR_func_error_string()` as well, since it was still OK to use before 3.0.0 and might provide valuable information?

fredmorcos: The answer is yes to both. I've added the necessary code.

rgacogne: Awesome, thanks!

rgacogne (approved):
:+1:

omoerbeek:
Compiler version does not here. My clang says:
`Apple clang version 14.0.0 (clang-1400.0.29.202)`
So I see:
```
configure: CC: ccache cc version
configure: CXX: ccache c++ -std=c++17 version
```

fredmorcos:
> Probably easiest to leave the awk part out.

Done.

omoerbeek (approved):

pdns/opensslsigners.cc
```diff
@@ -225,10 +224,10 @@ class OpenSSLRSADNSCryptoKeyEngine : public DNSCryptoKeyEngine
   bool verify(const std::string& hash, const std::string& signature) const override;
   std::string getPubKeyHash() const override;
   std::string getPublicKeyString() const override;
-  std::unique_ptr<BIGNUM, void(*)(BIGNUM*)>parse(std::map<std::string, std::string>& stormap, const std::string& key) const;
+  std::unique_ptr<BIGNUM, void (*)(BIGNUM*)> parse(std::map<std::string, std::string>& stormap, const std::string& key) const;
   void fromISCMap(DNSKEYRecordContent& drc, std::map<std::string, std::string>& stormap) override;
   void fromPublicKeyString(const std::string& content) override;
-  bool checkKey(vector<string> *errorMessages) const override;
+  bool checkKey(vector<string>* errorMessages) const override;
```
cmouse, line 230: would be maybe opportune moment to not use vector<T>* here?

pdns/opensslsigners.cc
```diff
@@ -238,10 +237,9 @@ class OpenSSLRSADNSCryptoKeyEngine : public DNSCryptoKeyEngine
 private:
   static int hashSizeToKind(size_t hashSize);
 
-  std::unique_ptr<RSA, void(*)(RSA*)> d_key;
+  std::unique_ptr<RSA, void (*)(RSA*)> d_key;
```
cmouse, line 240: RAS_new, RSA_free?

fredmorcos: Do you mean to use `decltype(&RSA_free)` instead of the raw function pointer type? If yes, this is done in an upcoming PR.

pdns/opensslsigners.cc
fredmorcos: Agreed, `vector<string>&` is probably better here.

pdns/opensslsigners.cc
cmouse: yeah

pdns/opensslsigners.cc
fredmorcos: I'll add the change to an upcoming PR.

Issue #12398, opened 2023-01-09
dnsdist: Unexpected behavior of cache misses counter for DoH queries


M4t7e:
- Program: dnsdist 
 - Issue type: Bug report

### Short description

For DoH queries the prometheus metric `dnsdist_pool_cache_misses` is calculated differently compared to Do53 (UDP & TCP) or DoT queries. It seems that for DoH queries the `dnsdist_pool_cache_misses` counter is increased while having a cache hit according to the `dnsdist_pool_cache_hits` metric.

@omoerbeek and @rgacogne mentioned in IRC that this could be related to the way how cache lookups are performed for DoH queries according to #11901. From the perspective of the cache it seems to be technically the correct number of hits and misses but from an operator perspective, who would like to see how good the cache performs and wants to calculate a cache hit rate, this behavior is likely unexpected. Since there is no other way do get a per pool/packet-cache metric with a logical cache misses counter, it would be good to either make this metric to a logical cache misses counter or provide an alternative to this metric.

`timeless` proposed in IRC that the behavior should be documented to make it clear for customers what this metric really does. A HELP text might be helpful in the documentation: https://dnsdist.org/guides/webserver.html#get--metrics
Here just a few metrics (including `dnsdist_pool_cache_misses`) don't have a HELP text and this could be helpful for understanding the metrics.

### Environment

 - Operating system: Ubuntu 22.04
 - Software version: dnsdist 1.7.3
 - Software source: PowerDNS repository

### Expected behaviour
A cache misses counter that logically counts the cache misses instead of technically from the perspective of the packet cache.

### Actual behaviour
Only technical cache misses are counted and can't be used to observe how the cache performs or calculate proper cache hit rates.

### Other information
The global metrics `dnsdist_cache_misses` and `dnsdist_cache_hits` are calculated logically but globally only. If one just needs global metrics, these metrics can be used.

Issue #12045, opened 2022-09-29
dnsdist: API stats response value in frontend object excludes cache hits


Vascko:
Hey folks,

### Category
 - Program: dnsdist
 - Issue type: Bug report

### Short description
When leveraging a packet cache, it appears that the `frontend response` stats from `api/v1/servers/localhost`, don't reflect the responses that are served from the packet cache while the `frontend queries` do:
```
"frontends": [
   {
      <... truncated ...>
      "queries": 39793301,
      "responses": 18505168,
      <... truncated ...>
   }
],
"pools": [
   {
      <... truncated ...>
      "cacheHits": 21288003,
      "cacheMisses": 18505110,
      <... truncated ...>
   }
]
```
This problem also exists for the prometheus endpoint:
```
dnsdist_responses 18539638
dnsdist_queries 39866885
dnsdist_frontend_queries{frontend="<redacted>",proto="UDP",thread="0"} 39866885
dnsdist_frontend_responses{frontend="<redacted>",proto="UDP",thread="0"} 18539638
dnsdist_pool_cache_hits{pool="_default_"} 21327108
dnsdist_pool_cache_misses{pool="_default_"} 18539589
```
### Environment

 - Operating system: RHEL 7
 - Software version: v1.7.0
 - Software source: PowerDNS repository

### Steps to reproduce
Configure a lister + pool with packet cache and query the HTTP API for the stats

### Expected behaviour
Summarized frontend responses should either include the number of responses served from the cache or should be broken out in `responsesFromCache` and `responsesFromDownstream`

### Actual behaviour
Responses don't reflect the actual number of responses to clients

PR #12405, opened 2023-01-09
dnsdist: Proper accounting of response and cache hits

rgacogne:
### Short description

- Properly record cache-hits as responses: For a very long time we have not been adding cache-hits to the responses counter, which is wrong. Let's fix it now. Closes https://github.com/PowerDNS/pdns/issues/12045.
- Only record one hit or miss per query in the cache metrics: The scope-zero feature and the DoH paths can actually do more than one lookup per query, and until now this led to an increase of the per-cache metric for every lookup, while the global `cache-hits` and `cache-misses` metrics were only updated once per query. This has led to several questions and misunderstandings, so we now only update the per-cache metrics once per query as well. Closes https://github.com/PowerDNS/pdns/issues/12398.
- Update the prometheus sample in the documentation: It was very old and not up-to-date.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/.github/actions/spell-check/expect.txt
+++ b/.github/actions/spell-check/expect.txt
@@ -1114,6 +1114,8 @@ noerrors
 NOLOCK
 nometasync
 Nominet
+noncompliantqueries
+noncompliantresponses
 nonexist
 Nonnekes
 noout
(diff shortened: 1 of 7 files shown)


rgacogne:
Rebased after squashing, hopefully the CI output will be better :)

chbruyand (approved):
lgtm!

PR #12411, opened 2023-01-11
Support OpenSSL 3.0 for ipcipher CA6 encryption/decryption

fredmorcos:
### Short description
Support OpenSSL 3.0 for ipcipher CA6 encryption/decryption. Nothing special here, use OpenSSL 3.0's `EVP_CIPHER_*` instead of the deprecated `AES_*` API but continue to support the deprecated OpenSSL 1.1.1 API.

### Checklist
I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/ipcipher.cc
+++ b/pdns/ipcipher.cc
@@ -1,5 +1,6 @@
 #include "ipcipher.hh"
 #include "ext/ipcrypt/ipcrypt.h"
+#include <cassert>
 #include <openssl/aes.h>
 #include <openssl/evp.h>
 
@@ -59,10 +60,47 @@ static ComboAddress encryptCA6(const ComboAddress& address, const std::string& k
 
   ComboAddress ret = address;
 
+#if OPENSSL_VERSION_MAJOR >= 3
+  auto ctx = std::unique_ptr<EVP_CIPHER_CTX, decltype(&EVP_CIPHER_CTX_free)>(EVP_CIPHER_CTX_new(), &EVP_CIPHER_CTX_free);
+  if (ctx == nullptr) {
+    throw pdns::OpenSSL::error("encryptCA6: Could not initialize cipher context");
+  }
+
+  auto aes128cbc = std::unique_ptr<EVP_CIPHER, decltype(&EVP_CIPHER_free)>(EVP_CIPHER_fetch(nullptr, "AES-128-CBC", nullptr), &EVP_CIPHER_free);
+
+  // NOLINTNEXTLINE(*-cast): Using OpenSSL C APIs.
+  if (EVP_EncryptInit(ctx.get(), aes128cbc.get(), reinterpret_cast<const unsigned char*>(key.c_str()), nullptr) == 0) {
+    throw pdns::OpenSSL::error("encryptCA6: Could not initialize encryption algorithm");
+  }
+
+  // Disable padding
+  const auto inSize = sizeof(address.sin6.sin6_addr.s6_addr);
+  static_assert(inSize == 16, "We disable padding and so we must assume a data size of 16 bytes");
+  const auto blockSize = EVP_CIPHER_get_block_size(aes128cbc.get());
+  assert(blockSize == 16);
+  EVP_CIPHER_CTX_set_padding(ctx.get(), 0);
+
+  int updateLen = 0;
+  // NOLINTNEXTLINE(*-cast): Using OpenSSL C APIs.
+  const auto* input = reinterpret_cast<const unsigned char*>(&address.sin6.sin6_addr.s6_addr);
+  // NOLINTNEXTLINE(*-cast): Using OpenSSL C APIs.
+  auto* output = reinterpret_cast<unsigned char*>(&ret.sin6.sin6_addr.s6_addr);
+  if (EVP_EncryptUpdate(ctx.get(), output, &updateLen, input, static_cast<int>(inSize)) == 0) {
+    throw pdns::OpenSSL::error("encryptCA6: Could not encrypt address");
+  }
+
+  int finalLen = 0;
+  if (EVP_EncryptFinal_ex(ctx.get(), output + updateLen, &finalLen) == 0) {
+    throw pdns::OpenSSL::error("encryptCA6: Could not finalize address encryption");
+  }
+
+  assert(updateLen + finalLen == inSize);
+#else
   AES_KEY wctx;
   AES_set_encrypt_key((const unsigned char*)key.c_str(), 128, &wctx);
   AES_encrypt((const unsigned char*)&address.sin6.sin6_addr.s6_addr,
               (unsigned char*)&ret.sin6.sin6_addr.s6_addr, &wctx);
+#endif
 
   return ret;
 }
--- a/pdns/ipcipher.cc
+++ b/pdns/ipcipher.cc
@@ -113,10 +113,47 @@ static ComboAddress decryptCA6(const ComboAddress& address, const std::string& k
 
   ComboAddress ret = address;
 
+#if OPENSSL_VERSION_MAJOR >= 3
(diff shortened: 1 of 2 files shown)


pdns/ipcipher.cc
```diff
@@ -59,10 +59,40 @@ static ComboAddress encryptCA6(const ComboAddress& address, const std::string& k
 
   ComboAddress ret = address;
 
+#if OPENSSL_VERSION_MAJOR >= 3
+  auto ctx = std::unique_ptr<EVP_CIPHER_CTX, decltype(&EVP_CIPHER_CTX_free)>(EVP_CIPHER_CTX_new(), &EVP_CIPHER_CTX_free);
+  if (ctx == nullptr) {
+    throw pdns::OpenSSL::error("encryptCA6: Could not initialize cipher context");
+  }
+
+  EVP_CIPHER_CTX_set_padding(ctx.get(), 0);
```
rgacogne, line 68: Actually this does not work for me (see `test-ipcrypt_cc.cc`), I need to call `EVP_CIPHER_CTX_set_padding(ctx.get(), 0);` *after* `EVP_EncryptInit` (and *after* `EVP_DecryptInit`) otherwise OpenSSL adds padding, messing things up.

pdns/ipcipher.cc
```diff
@@ -59,10 +59,40 @@ static ComboAddress encryptCA6(const ComboAddress& address, const std::string& k
...
+  // NOLINTNEXTLINE(*-cast): Using OpenSSL C APIs.
+  auto* output = reinterpret_cast<unsigned char*>(&ret.sin6.sin6_addr.s6_addr);
+  auto inSize = sizeof(address.sin6.sin6_addr.s6_addr);
+  if (EVP_EncryptUpdate(ctx.get(), output, &len, input, static_cast<int>(inSize)) == 0) {
+    throw pdns::OpenSSL::error("encryptCA6: Could not encrypt address");
+  }
+
+  if (EVP_EncryptFinal_ex(ctx.get(), output + len, &len) == 0) {
+    throw pdns::OpenSSL::error("encryptCA6: Could not finalize address encryption");
+  }
```
rgacogne, line 89: Shouldn't be summing up the length returned by `EVP_EncryptUpdate` and the one returned by `EVP_EncryptFinal_ex`? And same for the decryption? In practice it might work because we only have one block, but that does seem brittle.

rgacogne (changes requested):

rgacogne:
That's absolutely not on you but it looks like the new code is three times slower than the existing one when doing one round of encryption/decryption. I guess that's likely because the new API requires at least two allocations per round while the deprecated one did no allocation at all..

pdns/ipcipher.cc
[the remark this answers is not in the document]
fredmorcos: I'm not sure how to add a static assertion here. The only way I can find is to either use the `OSSL_CIPHER_PARAM_BLOCK_SIZE` parameter on the EVP context object or call `EVP_CIPHER_get_block_size`. However the documentation mentions the following:

```
"blocksize" (OSSL_CIPHER_PARAM_BLOCK_SIZE) <unsigned integer>
           Gets the block size for the associated cipher algorithm cipher.  The block size should be 1 for stream ciphers.  Note that the block size for a cipher may be different to the block
           size for the underlying encryption/decryption primitive.  For example AES in CTR mode has a block size of 1 (because it operates like a stream cipher), even though AES has a block
           size of 16.  Use EVP_CIPHER_get_block_size() to retreive the cached value.
```

pdns/ipcipher.cc
```diff
@@ -59,10 +59,40 @@ static ComboAddress encryptCA6(const ComboAddress& address, const std::string& k
...
+    throw pdns::OpenSSL::error("encryptCA6: Could not initialize encryption algorithm");
+  }
+
+  int len = 0;
+  // NOLINTNEXTLINE(*-cast): Using OpenSSL C APIs.
+  const auto* input = reinterpret_cast<const unsigned char*>(&address.sin6.sin6_addr.s6_addr);
+  // NOLINTNEXTLINE(*-cast): Using OpenSSL C APIs.
+  auto* output = reinterpret_cast<unsigned char*>(&ret.sin6.sin6_addr.s6_addr);
+  auto inSize = sizeof(address.sin6.sin6_addr.s6_addr);
+  if (EVP_EncryptUpdate(ctx.get(), output, &len, input, static_cast<int>(inSize)) == 0) {
```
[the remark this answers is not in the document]
fredmorcos: What is a type system 🙃 🤷🏽‍♂️

pdns/ipcipher.cc
fredmorcos: That is strange. I was testing using `testrunner --run_test=test_ipcrypt_hh` (which is in `test-ipcrypt_cc.cc`), that's how I figured out the need for explicitly setting the padding to 0. For some reason it also breaks for me now.

fredmorcos:
> That's absolutely not on you but it looks like the new code is three times slower than the existing one when doing one round of encryption/decryption. I guess that's likely because the new API requires at least two allocations per round while the deprecated one did no allocation at all..

Perhaps we can reuse context objects. But I would attempt to do that after all the openssl3 updates so that we at least have a basis of "correctness" to base that change on.

pdns/ipcipher.cc
fredmorcos: So I'm not entirely sure what you mean. The "final" piece of encryped/decrypted data needs to be written starting at byte `output + len` (where `len` is coming from the previous call to an encryption/decryption routine, in this case `EVP_EncryptUpdate`).

I'm definitely missing something, you know better, but Isn't that also independent of how many blocks there are?

Additionally, the man page mentions the following:

```
If padding is disabled then EVP_EncryptFinal_ex() will not encrypt any more data and it will return an error if any data remains in a partial block: that is if the total data length
           is not a multiple of the block size.
```

pdns/ipcipher.cc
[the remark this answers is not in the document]
rgacogne: Sorry, I should have been more clear. I had in mind a static assertion checking that the size of the data we are going to encrypt is 16 bytes, with a message explaining that this is mandatory given that padding is disabled.
What I'm worried about is someone reusing that code later with some data that is not a multiple of the block size, or a different algorithm, without understanding that limitation.

pdns/ipcipher.cc
rgacogne: What I have in mind is that `EVP_EncryptUpdate` might be writing `N` bytes to `output`, and then `EVP_EncryptFinal_ex` might be adding `F` bytes to that. The current code correctly account for the `N` bytes by passing `output + len` to `EVP_EncryptFinal_ex` but then overwrites `len`, so we never know exactly how many bytes were written.
You are right that this works today, because:
- our data size is a multiple of the block size, with no padding, so  `EVP_EncryptFinal_ex` doesn't write anything
- we do not care about the final length

What I'm worried about is that someone later does a copy/paste of that code, or reuse it somewhere else. So I would prefer that we use a different variable to get the number of bytes written by  `EVP_EncryptFinal_ex` , and check that the sum of `N`+`F` is equal to `inSize`. It will always be true for this exact case, but might not if we reuse the code later for something else.

rgacogne:
> Perhaps we can reuse context objects. But I would attempt to do that after all the openssl3 updates so that we at least have a basis of "correctness" to base that change on.

Agreed, and even then I would not worry too much about it until we measure an actual regression. I don't know how many people actually use that ipcipher code in production nowadays.

pdns/ipcipher.cc
fredmorcos: Reasonable. I added it as an assertion.

rgacogne (approved):
Looks good, nice work!

PR #12412, opened 2023-01-11
OpenSSL 3.0: TLS MAC handling

fredmorcos:
### Short description
OpenSSL 3.0: TLS MAC handling. Use `EVP_MAC_CTX` instead of `HMAC_CTX` on OpenSSL 3.0 but continue to support `HMAC_CTX` for OpenSSL < 3.0.

### Checklist
I have:
- [X] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/libssl.cc
+++ b/pdns/libssl.cc
@@ -216,7 +216,7 @@ void libssl_set_ticket_key_callback_data(SSL_CTX* ctx, void* data)
 
 int libssl_ticket_key_callback(SSL* s, OpenSSLTLSTicketKeysRing& keyring, unsigned char keyName[TLS_TICKETS_KEY_NAME_SIZE], unsigned char* iv, EVP_CIPHER_CTX* ectx, HMAC_CTX* hctx, int enc)
 {
-  if (enc) {
+  if (enc != 0) {
     const auto key = keyring.getEncryptionKey();
     if (key == nullptr) {
       return -1;
@@ -233,7 +233,7 @@ int libssl_ticket_key_callback(SSL* s, OpenSSLTLSTicketKeysRing& keyring, unsign
     return 0;
   }
 
-  if (key->decrypt(iv, ectx, hctx) == false) {
+  if (!key->decrypt(iv, ectx, hctx)) {
     return -1;
   }
 
--- a/pdns/libssl.cc
+++ b/pdns/libssl.cc
@@ -214,7 +214,7 @@ void libssl_set_ticket_key_callback_data(SSL_CTX* ctx, void* data)
   SSL_CTX_set_ex_data(ctx, s_ticketsKeyIndex, data);
 }
 
-int libssl_ticket_key_callback(SSL *s, OpenSSLTLSTicketKeysRing& keyring, unsigned char keyName[TLS_TICKETS_KEY_NAME_SIZE], unsigned char *iv, EVP_CIPHER_CTX *ectx, HMAC_CTX *hctx, int enc)
+int libssl_ticket_key_callback(SSL* s, OpenSSLTLSTicketKeysRing& keyring, unsigned char keyName[TLS_TICKETS_KEY_NAME_SIZE], unsigned char* iv, EVP_CIPHER_CTX* ectx, HMAC_CTX* hctx, int enc)
 {
   if (enc) {
     const auto key = keyring.getEncryptionKey();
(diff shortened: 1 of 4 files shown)


fredmorcos:
@rgacogne Please double-check the build guards.

pdns/libssl.cc
```diff
@@ -707,18 +721,74 @@ int OpenSSLTLSTicketKey::encrypt(unsigned char keyName[TLS_TICKETS_KEY_NAME_SIZE
...
 
+#if OPENSSL_VERSION_MAJOR >= 3
+  using ParamsBuilder = std::unique_ptr<OSSL_PARAM_BLD, decltype(&OSSL_PARAM_BLD_free)>;
+
+  auto params_build = ParamsBuilder(OSSL_PARAM_BLD_new(), OSSL_PARAM_BLD_free);
+  if (params_build == nullptr) {
+    return -1;
+  }
+
+  if (OSSL_PARAM_BLD_push_utf8_string(params_build.get(), OSSL_MAC_PARAM_DIGEST, "sha256", 7) == 0) {
```
rgacogne, line 732: It might be nice to move "sha256" to a constant, but more importantly, what is `7`?

omoerbeek: "The length of the string bsize should not include the terminating NUL byte." So it dos not look like the length of "sha256"

omoerbeek: You can also use 0 to let the function compute the length.

fredmorcos: I used 7 as the length of "sha256" and missed the part about not counting the terminating NUL byte. I also didn't want the function the re-compute something we already know the length of.

Does the following look OK?

```
constexpr static const char* sha256KeyName{"sha256"};
constexpr static const size_t sha256KeyNameLength = 6;
```

rgacogne: Understood, I thought I had missed something! :-) Yes, the constants look good, thank you!

fredmorcos:
Reminder to force-merge this please.

pdns/libssl.hh
```diff
@@ -91,8 +91,17 @@ public:
   ~OpenSSLTLSTicketKey();
 
   bool nameMatches(const unsigned char name[TLS_TICKETS_KEY_NAME_SIZE]) const;
-  int encrypt(unsigned char keyName[TLS_TICKETS_KEY_NAME_SIZE], unsigned char *iv, EVP_CIPHER_CTX *ectx, HMAC_CTX *hctx) const;
-  bool decrypt(const unsigned char* iv, EVP_CIPHER_CTX *ectx, HMAC_CTX *hctx) const;
+
+#if OPENSSL_VERSION_MAJOR >= 3
+  constexpr static const char* sha256KeyName{"sha256"};
+  constexpr static const size_t sha256KeyNameLength{6};
```
rgacogne, line 97: Would you mind moving these to `libssl.cc` instead?

rgacogne (changes requested):
This currently does not build for me because we need to update the `ticket_key_callback` function and co in `doh.cc` as well.

pdns/libssl.hh
cmouse: why not just use std::string? you can then use c_str()

omoerbeek: > why not just use std::string? you can then use c_str()

and `.size()`.

fredmorcos: I don't think it's possible to use `std::string` with `constexpr`?
I suspect the compiler's constant folding mechanism should kick in, but I'm never sure in C++ so I figured it's just better to be explicit about what I want.

omoerbeek: `constexpr` strings are in C++20 I believe. But I prefer `const` over `constexpr` if it allows to enhance readability.

fredmorcos:
> This currently does not build for me because we need to update the `ticket_key_callback` function and co in `doh.cc` as well.

Fixed.

rgacogne (dismissed):
:+1:

rgacogne (changes requested):
We need that change as well:
```diff
diff --git pdns/dnsdistdist/doh.cc pdns/dnsdistdist/doh.cc
index ed2a165cf..443b15008 100644
--- pdns/dnsdistdist/doh.cc
+++ pdns/dnsdistdist/doh.cc
@@ -1390,7 +1390,11 @@ static void setupTLSContext(DOHAcceptContext& acceptCtx,

   if (tlsConfig.d_enableTickets && tlsConfig.d_numberOfTicketsKeys > 0) {
     acceptCtx.d_ticketKeys = std::make_unique<OpenSSLTLSTicketKeysRing>(tlsConfig.d_numberOfTicketsKeys);
+#if OPENSSL_VERSION_MAJOR >= 3
+    SSL_CTX_set_tlsext_ticket_key_evp_cb(ctx.get(), &ticket_key_callback);
+#else
     SSL_CTX_set_tlsext_ticket_key_cb(ctx.get(), &ticket_key_callback);
+#endif
     libssl_set_ticket_key_callback_data(ctx.get(), &acceptCtx);
   }
```

fredmorcos:
Hopefully done now :-D

rgacogne (approved):

PR #12422, opened 2023-01-16
dnsdist: Document interactions between XFR, NOTIFY and the cache

### LINKED ISSUE #11135: dnsdist: Caching of invalid NOTIFY responses might prevent valid NOTIFY
- Program: dnsdist
 - Issue type: Bug report

### Short description
dnsdist in front of an Auth server seems to put refused notify responses in its packet cache based on `temporaryFailureTTL` regardless of source which prevents a following valid notify from reaching the Auth server.

### Environment
 - Software version: 1.6.1-1pdns.focal
 - Software source: PowerDNS repo

### Steps to reproduce
1. Put pdns behind dnsdist with PROXY protocol and configure packet cache
2. Send DNS notify from non-primary IP
3. Send DNS notify from primary IP

### Expected behaviour
2nd notify from proper primary IP reaching the Auth server and not being handled by packet cache.

### Actual behaviour
2nd notify handled by packet cache and resulting in immediate REFUSED response.

### Other information
This might be a bit of a not very common edge case, especially as it needs the PROXY protocol (or another way to pass the real client IP to Auth) - but it still blocks valid NOTIFY messages for the duration of `temporaryFailureTTL`. On the other hand this behavior makes some sense when considering how a packet cache works..
I could think of some configuration workarounds of not caching SOA messages that are notifies or using a lower `temporaryFailureTTL` for these. I ended up using the following action:
```
 addAction(AndRule({OpcodeRule(DNSOpcode.Notify), QTypeRule(DNSQType.SOA)}), SetSkipCacheAction())
```
Given that valid notifies are (of course) not cached I guess the proper solution might be not to cache "refused" replies for incorrect ones - or cache them based on source IP however that might be a more complicated topic.. especially as all other Auth replies (in my use case) are not source-dependent so the whole fancy ECS handling is not in use.
The current behavior might have certain side-effects on timely secondary NS updates when faced with a continuous stream of invalid notifies.
I wonder if this needs a fix or if we are fine with adding some documentation about that behavior, e.g. to [`pdns/pdns/dnsdistdist/docs/advanced/axfr.rst`](https://github.com/PowerDNS/pdns/blob/master/pdns/dnsdistdist/docs/advanced/axfr.rst) - but that might be missed.
(I just noticed that this file might need an update to mention the PROXY protocol or other ways of passing client information)

rgacogne:
### Short description

Closes #11135.

### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdistdist/docs/advanced/axfr.rst
+++ b/pdns/dnsdistdist/docs/advanced/axfr.rst
@@ -39,7 +39,7 @@ to do that are described in :doc:`Passing the source address to the backend <pas
   .. code-block:: lua
 
     -- this rule will not stop the processing, but disable caching for AXFR and IXFR responses
-    addAction(ORRule({QTypeRule(DNSQType.AXFR), QTypeRule(DNSQType.IXFR)}), SetSkipCacheAction())
+    addAction(OrRule({QTypeRule(DNSQType.AXFR), QTypeRule(DNSQType.IXFR)}), SetSkipCacheAction())
     -- this rule will route SOA, AXFR and IXFR queries to a specific pool of servers
     addAction(OrRule({QTypeRule(DNSQType.SOA), QTypeRule(DNSQType.AXFR), QTypeRule(DNSQType.IXFR)}), PoolAction("primary"))
 
--- a/pdns/dnsdistdist/docs/advanced/axfr.rst
+++ b/pdns/dnsdistdist/docs/advanced/axfr.rst
@@ -43,6 +43,9 @@ to do that are described in :doc:`Passing the source address to the backend <pas
     -- this rule will route SOA, AXFR and IXFR queries to a specific pool of servers
     addAction(OrRule({QTypeRule(DNSQType.SOA), QTypeRule(DNSQType.AXFR), QTypeRule(DNSQType.IXFR)}), PoolAction("primary"))
 
+.. versionchanged:: 1.8.0
+  Since 1.8.0, dnsdist will no longer cache responses to AXFR and IXFR queries.
+
 In front of secondaries
 -----------------------
 
--- a/pdns/dnsdistdist/docs/advanced/axfr.rst
+++ b/pdns/dnsdistdist/docs/advanced/axfr.rst
@@ -65,5 +65,5 @@ check to :program:`dnsdist`'s side::
 
   .. code-block:: lua
 
-    -- this rule will disable caching for NOTIFY responses
-    addAction(AndRule({OpcodeRule(DNSOpcode.Notify), QTypeRule(DNSQType.SOA)}), SetSkipCacheAction())
+    -- this rule will disable the caching of responses for NOTIFY queries
+    addAction(OpcodeRule(DNSOpcode.Notify), SetSkipCacheAction())
--- a/pdns/dnsdistdist/docs/advanced/axfr.rst
+++ b/pdns/dnsdistdist/docs/advanced/axfr.rst
@@ -1,6 +1,9 @@
 AXFR, IXFR and NOTIFY
 =====================
 
+In front of primaries
+---------------------
+
 When :program:`dnsdist` is deployed in front of a primary authoritative server, it might receive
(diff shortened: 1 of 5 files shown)


pdns/dnsdistdist/docs/advanced/axfr.rst
```diff
@@ -24,6 +27,25 @@ and moving the source address check to :program:`dnsdist`'s side::
...
+.. warning::
+
+  Be wary of dnsdist caching the responses to AXFR and IXFR queries and sending these to the wrong clients.
+  This is mitigated by default when the source IP of the client is passed using EDNS Client Subnet, but
+  not when the proxy protocol is used, so disabling caching for these kinds of queries is advised:
+
+  .. code-block:: lua
+
+    -- this rule will not stop the processing, but disable caching for AXFR and IXFR responses
+    addAction(ORRule({QTypeRule(DNSQType.AXFR), QTypeRule(DNSQType.IXFR)}), SetSkipCacheAction())
```
Habbie, line 42: what would dnsdist be caching? not a whole zone transfer I hope? but perhaps a short REFUSED answer?

pdns/dnsdistdist/docs/advanced/axfr.rst
```diff
@@ -34,3 +56,14 @@ check to :program:`dnsdist`'s side::
...
+
+.. warning::
+
+  Be wary of dnsdist caching the responses to NOTIFY queries and sending these to the wrong clients.
+  This is mitigated by default when the source IP of the client is passed using EDNS Client Subnet, but
+  not when the proxy protocol is used, so disabling caching for these kinds of queries is advised:
+
+  .. code-block:: lua
+
+    -- this rule will disable caching for NOTIFY responses
```
Habbie, line 68:
```suggestion
    -- this rule will disable caching for NOTIFY and SOA responses
```

(assuming that was intentional)

pdns/dnsdistdist/docs/advanced/axfr.rst
Habbie:
```suggestion
    addAction(OrRule({QTypeRule(DNSQType.AXFR), QTypeRule(DNSQType.IXFR)}), SetSkipCacheAction())
```

Habbie (changes requested):

pdns/dnsdistdist/docs/advanced/axfr.rst
rgacogne: dnsdist only caches answers smaller or equal to 4096 bytes, so hopefully not the full AXFR response, although in theory it would be possible. Perhaps we should explicitly skip that?

pdns/dnsdistdist/docs/advanced/axfr.rst
rgacogne: Would `this rule will disable the caching of responses for NOTIFY queries with a SOA qtype` be more clear, perhaps? As far as I know there is no other supported `qtype` for `NOTIFY` queries, but I might have missed one RFC or two or twenty ;-).

pdns/dnsdistdist/docs/advanced/axfr.rst
Habbie: Yes, I think hardcoding that makes sense!

pdns/dnsdistdist/docs/advanced/axfr.rst
Habbie: oh sorry, it's `AndRule`! I think you can just leave it. We could also play it extra safe and ignore the qtype?

Habbie: I'm less sure about hardcoding it for NOTIFY, although I can't argue why very well.

rgacogne: Alright, let's not, then, it's easy to add a rule if needed!

Habbie: :+1:

Habbie (approved):

PR #12423, opened 2023-01-16
OpenSSL 3.0: Offer TLS providers as an alternative to TLS engines in DNSdist

fredmorcos:
### Short description
OpenSSL 3.0 replaced engines with providers. This change offers `loadTLSProvider` as an alternative to `loadTLSEngine`. This is currently used for e.g. HW accelerated crypto: https://github.com/intel/QAT_Engine/blob/master/docs/qat_common.md#openssl-30-provider-support

Providers are disabled by default. The configure flag `--enable-tls-providers` is available to force-enable them over engines. It is marked as experimental though.

### Checklist
I have:
- [X] compiled this code
- [ ] tested this code
- [ ] documented the code


fredmorcos:
# Regarding build guards (i.e. #ifdefs)

- [ ] Add an experimental `--enable-tls-provider` build flag
- [ ] Ensure that `loadTLSEngine` and related functions are only built when `#if defined(HAVE_LIBSSL) && !defined(ENABLE_TLS_PROVIDER)`
- [ ] Ensure that `loadTLSProvider` and related functions are only built when `#if defined(HAVE_LIBSSL) && OPENSSL_VERSION_MAJOR >= 3 && defined(ENABLE_TLS_PROVIDER)`: [This is due to the Intel QAT OpenSSL provider still being experimental](https://github.com/intel/QAT_Engine/blob/master/docs/qat_common.md#openssl-30-provider-support).

# Documentation
- [ ] Documentation needs to be double-checked.
- [ ] Which DNSdist [version](https://github.com/PowerDNS/pdns/pull/12423/files#diff-abd4de1691f1c874111fbd923e679855f845526ea13bca77f6084be3d187f7b1R1860) will this be introduced in?

# Tests
- [ ] Run tests for `loadTLSEngine` with those changes.
- [ ] Add tests for `loadTLSProvider`.

rgacogne:
>     * [ ]  Add an experimental `--enable-tls-provider` build flag

I was pondering doing it the other way around: have a `--enable-deprecated-tls-engine` that would use the TLS engine interface when building against OpenSSL 3.0.x. It's officially deprecated but it might be the only way to support some functionalities for which a TLS provider is not available.
 
>     * [ ]  Which DNSdist [version](https://github.com/PowerDNS/pdns/pull/12423/files#diff-abd4de1691f1c874111fbd923e679855f845526ea13bca77f6084be3d187f7b1R1860) will this be introduced in?

I'm aiming for 1.8.0.

pdns/dnsdistdist/docs/reference/config.rst
```diff
@@ -1855,6 +1855,16 @@ Other functions
...
   :param string defaultString: The default string to pass to the engine. The exact value depends on the engine but represents the algorithms to register with the engine, as a list of  comma-separated keywords. For example "RSA,EC,DSA,DH,PKEY,PKEY_CRYPTO,PKEY_ASN1".
 
+.. function:: loadTLSProvider(providerName)
+
+  .. versionadded:: ???
+
+  Load the OpenSSL provider named ``providerName``. Providers can be used to accelerate cryptographic operations, like for example Intel QAT.
+  At the moment up to a maximum of 32 loaded providers are supported, and that support is experimental.
+  Some providers might actually degrade performance unless the TLS asynchronous mode of OpenSSL is enabled. To enable it see the ``tlsAsyncMode`` parameter on :func:`addTLSLocal`.
+
```
rgacogne, line 1865: We might want to state that this is only available when built against OpenSSL >= 3.0.0

pdns/libssl.cc
```diff
@@ -165,6 +174,32 @@ void unregisterOpenSSLUser()
...
+  if (s_users.load() == 0) {
+    /* We need to make sure that OpenSSL has been properly initialized before loading an engine.
+       This messes up our accounting a bit, so some memory might not be properly released when
+       the program exits when engines are in use. */
+    registerOpenSSLUser();
+  }
+
+  auto providers = s_providers.lock();
+  if (providers->count(providerName) > 0) {
+    return { false, "TLS engine already loaded" };
```
rgacogne, line 189:
```suggestion
    return { false, "TLS provider already loaded" };
```

pdns/libssl.cc
```diff
@@ -165,6 +174,32 @@ void unregisterOpenSSLUser()
...
+  }
+
+  auto provider = std::unique_ptr<OSSL_PROVIDER, decltype(&OSSL_PROVIDER_unload)>(OSSL_PROVIDER_load(nullptr, providerName.c_str()), OSSL_PROVIDER_unload);
+  if (provider == nullptr) {
+    return { false, "unable to load TLS engine '" + providerName + "'" };
```
rgacogne, line 194:
```suggestion
    return { false, "unable to load TLS provider '" + providerName + "'" };
```

rgacogne:
> I'll try with QAT later today.

So.. Technically I managed to load the QAT provider, and to handle a few TLS connections with it. However it crashed pretty quickly in the QAT provider itself, but I don't think this is on us, the QAT provider is very experimental and from what I read not really tested.
I manually edited the guard to restore the ability to load an OpenSSL engine even with OpenSSL >= 3.0 and it works as expected. So I believe we really need to make it easy to have the ability to use OpenSSL engines even when compiled against OpenSSL >= 3.0, as providers might not be production-ready for a while. How about adding `--enable-tls-providers` and `--enable-tls-engines`? I'm not even sure we need to make them exclusive, to be honest, so we might be able to just implement `--enable-tls-engines`, disabled by default, and only enable providers when OpenSSL >= 3.0?

fredmorcos:
> so we might be able to just implement `--enable-tls-engines`, disabled by default, and only enable providers when OpenSSL >= 3.0?

I'm not sure this is a good idea given the sorry state of the Intel QAT provider (which is the only use-case I know of so far). I would prefer `--enable-tls-providers` which only works with OpenSSL >= 3.0. Is that OK?

rgacogne:
> > so we might be able to just implement `--enable-tls-engines`, disabled by default, and only enable providers when OpenSSL >= 3.0?
> 
> I'm not sure this is a good idea given the sorry state of the Intel QAT provider (which is the only use-case I know of so far). I would prefer having engines enabled by default and `--enable-tls-providers` which only works with OpenSSL >= 3.0. Is that OK?

Yes, I think you are right that this is a better default. I'm a bit sad because it means we will have some remaining warnings by default when building with OpenSSL >= 3.0, because of the deprecated engines, but in practice this is likely what most users are going to want for now.

pdns/dnsdistdist/configure.ac
```diff
@@ -69,6 +69,17 @@ AM_CONDITIONAL([HAVE_CDB], [false])
...
 PDNS_CHECK_LIBCRYPTO
 
+DNSDIST_ENABLE_TLS_PROVIDERS
+
+AS_IF([test "x$enable_tls_providers" != "xno"], [
+  AS_IF([test "x$HAVE_LIBSSL_3_PLUS" != "x1"], [
+    AC_MSG_ERROR([TLS providers support requires OpenSSL >= 3.0])
+  ])
+
+  PDNS_WITH_LIBSSL
```
rgacogne, line 79: Do we really need to call `PDNS_WITH_LIBSSL` here?

fredmorcos: It sets up all the necessary paths for openssl so I would say yes it's required.

rgacogne: Isn't it getting called on line 76 already?

fredmorcos: Did you perhaps mean line 72?
`DNSDIST_ENABLE_TLS_PROVIDERS` only checks the openssl version and doesn't concern itself with paths.

rgacogne: That's 76 _before_ the PR, sorry, so 87 now.

fredmorcos: Right.. and since provider support isn't very useful without TLS in our case, we can skip the call here.

fredmorcos: I'll make the changes, shall I squash after that?

rgacogne: Yes, please!

Issue #11216, opened 2022-01-19
dnsdist: carbon - inaccurate timestamps when using a short reporting interval


rygl:
- Program: dnsdist
 - Issue type: Bug report

### Short description
When using a carbon reporting with a reporting interval of about 5s the timestamps in the carbon protocol are not accurate. This causes a problem on the carbon server - the data are not accepted when using a resolution of 5s as they do not match 5s pattern. It causes empty datapoints and spoils graphs.

### Environment

 - Operating system: Debian 11.2
 - Software version: 1.7.0, 1.6.1
 - Software source: PowerDNS repository

### Steps to reproduce

1. Configure a dnsdist instance with rather a complex configuration
2. Configure a carbon server with reporting interval of 5s
3. Watch the data stored in whisper db files or capture the carbon traffic

### Expected behaviour
The timestamps are strictly in 5s interval

### Actual behaviour
A sample decoded from the TCP stream:
dnsdist.rzt-dns-lb1.main.responses 54052979 1642623170 ~ Wed 19 Jan 2022 **08:54:04** PM CET
dnsdist.rzt-dns-lb1.main.responses 54030274 1642623159 ~Wed 19 Jan 2022 **09:12:39** PM CET

A restart does not help here.

### Other information
What is interesting is the fact that this happens in an interval of about 6min - every 6 minutes there is a shifted timestamp. It looks like that with reporting interval of 10s it does not occur. And not all dnsdist instances we use are affected by this. Is is likely related to the complexity of the configuration (number of listeners etc.) than the load (instances nearly without traffic are affected the same way as instances with thousands of qps)

Both dnsdist and carbon receiver are hooked to GPS NTP.

rgacogne:
I think the root cause is that we sleep for the configured interval between two runs, regardless of how long a run took. We already know [1] that it causes issues when a server is unreachable but it might also becomes one if a single run is taking too long.

[1]: https://github.com/PowerDNS/pdns/issues/10517

rgacogne:
Fixed by #12424.

PR #12424, opened 2023-01-16
dnsdist: Better handling of multiple carbon servers

### LINKED ISSUE #10517: dnsdist: Carbon intervals are slightly off when exporting to multiple servers and one is unreachable
- Program: dnsdist
 - Issue type: Bug report

### Short description
When multiple Carbon servers are configured in dnsdist (with multiple "carbonServer" statements) and one is unreachable the reporting frequency to the remaining server(s) becomes unreliable, for example 32 seconds instead of 30 at random intervals, which causes some issues for reports that rely on exact data update intervals.

### Environment

 - Software version: 1.5.2
 - Software source: PowerDNS repo (1.5.2-1pdns.bionic)

### Steps to reproduce

1. Configure at least two carbonServer's
2. Make one of them unreachable (timeout)
3. Validate reporting frequency to the remaining reachable carbon server

### Expected behaviour
Strict reporting interval

### Actual behaviour
Random reporting interval +/- a few seconds each..

### Other information
"As discussed."
Thanks. :wink:

rgacogne:
### Short description

This PR implements a refactoring of the Carbon handling in dnsdist. Each remote endpoint is now handled by its own thread, as to prevent a slow/unreachable server from impacting others.
It also implements an exponential back-off when a server cannot be reached properly, as to not overload a suffering endpoint.

Hopefully closes #10517 and #11216.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/pdns/dnsdist-carbon.cc
+++ b/pdns/dnsdist-carbon.cc
@@ -264,13 +264,13 @@ static bool doOneCarbonExport(const Carbon::Endpoint& endpoint)
   return true;
 }
 
-static void carbonHandler(Carbon::Endpoint endpoint)
+static void carbonHandler(Carbon::Endpoint&& endpoint)
 {
   setThreadName("dnsdist/carbon");
   const auto intervalUSec = endpoint.interval * 1000 * 1000;
 
   try {
-    size_t consecutiveFailures = 0;
+    uint8_t consecutiveFailures = 0;
     do {
       DTime dt;
       dt.set();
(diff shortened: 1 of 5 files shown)


pdns/dnsdist-carbon.cc
```diff
@@ -31,255 +31,321 @@
...
+  }
+  catch (const std::exception& e) {
+    warnlog("Problem sending carbon data to %s: %s", server.toStringWithPort(), e.what());
+    return false;
+  }
+
+  return true;
+}
+
+static void carbonHandler(Carbon::Endpoint endpoint)
```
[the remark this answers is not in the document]
omoerbeek: A comment that this argument copy is needed would be nice, as not doing a call by value would break things. Potentially a call by ref and a local var copy would silence the warning and make intentions more clear.

pdns/dnsdist-carbon.cc
```diff
@@ -31,255 +31,321 @@
...
         }
-        catch(const std::exception& e) {
-          warnlog("Problem sending carbon data: %s", e.what());
+        consecutiveFailures = 0;
+      }
+      else {
+        /* maximum interval between two attempts is 10 minutes */
+        const time_t maxBackOff = 10 * 60;
+        time_t backOff = 1;
+        double backOffTmp = std::pow(2.0, consecutiveFailures);
```
omoerbeek, line 292: likely some static analysis tool will complain about the implicit `size_t` -> `double` conversion.

pdns/dnsdist-carbon.cc
```diff
@@ -31,255 +31,321 @@
...
+  if (config->d_running) {
+    // we already started the threads, let's just spawn a new one
+    std::thread newHandler(carbonHandler, endpoint);
+    newHandler.detach();
   }
-  catch(const PDNSException& e)
-  {
-    errlog("Carbon thread died, PDNSException: %s", e.reason);
+  else {
+    config->d_endpoints.push_back(std::move(endpoint));
```
omoerbeek, line 331: Should the endpoint only be pushed if not running? In that case you might as well clear the endpoint vector in `run()` as it is only used for the defered startup.

omoerbeek (approved):
A few nits

pdns/dnsdist-carbon.cc
rgacogne: Done!

rgacogne:
I pushed a new commit:
- We now explicitly convert to double, making sure that we will not overflow by restricting the value of the counter
- Clear the endpoints list when the carbon threads are started, to make clear we do not need them anymore
- Move the endpoints passed to the carbon threads, to make static analysis tools happy.

PR #12425, opened 2023-01-17
rec: Change the way RD=0 forwarded queries are handled.

omoerbeek:
Note: I'm marking this as "Draft" since I _might_ have overlooked a case where the forward to a forwarder in the RD=0 case is actually needed or useful. I cannot think of one and the unit tests and regression tests agree, but I do need independent confirmation on this. That said:

Since forever, there has been special case code for forwarded queries in the RD=0 case.  This special case code does a hardcoded RD=0 query to the specified forwarder.  This code has two consequences:

1. Even if the forwarder is marked recursive it gets a RD=0 query
2. The cache is not consulted at all

The corresponding unit tests actually test this behaviour, but after historic digging with help from @rgacogne it turns out the the unit test do not reflect the desired functionality, but the current state of affairs to help with a refactoring PR.  That is good, since refactoring should not change functionality.

But now the time has come to change the code to do the desired thing:

1. If an RD=0 query is received, do a cache only-lookup in all cases.
2. Never send a RD=0 query to a recursive forwarder

I already did a similar thing when I wrote the QName Minimization code, introducing a conditional that only gets set for that case, to avoid changing unrelated (to QM) functionality.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/recursordist/syncres.cc
+++ b/pdns/recursordist/syncres.cc
@@ -1655,8 +1655,7 @@ int SyncRes::doResolve(const DNSName& qname, const QType qtype, vector<DNSRecord
   bool fromCache = false;
   // For cache peeking, we tell doResolveNoQNameMinimization not to consider the (non-recursive) forward case.
   // Otherwise all queries in a forward domain will be forwarded, while we want to consult the cache.
-  // The out-of-band cases for doResolveNoQNameMinimization() should be reconsidered and redone some day.
-  int res = doResolveNoQNameMinimization(qname, qtype, retq, depth, beenthere, context, &fromCache, nullptr, false);
+  int res = doResolveNoQNameMinimization(qname, qtype, retq, depth, beenthere, context, &fromCache, nullptr);
   setCacheOnly(old);
   if (fromCache) {
     QLOG("Step0 Found in cache");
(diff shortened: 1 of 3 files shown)


rgacogne (approved):
The changes look good to me, nice work!

>  I might have overlooked a case where the forward to a forwarder in the RD=0 case is actually needed or useful. I cannot think of one and the unit tests and regression tests agree, but I do need independent confirmation on this.

I don't see it either. There has even been quite some debate on whether `RD=0` queries to recursor should be discarded right away, and apart from debugging I don't remember anyone providing a valid use-case, so the impact of these changes should be low anyway.

paddg:
> There has even been quite some debate on whether `RD=0` queries to recursor should be discarded right away, ... I don't remember anyone providing a valid use-case

We use `SetNoRecurseAction()`  in dnsdist on a `MaxQPSIPRule` to make the rate-limit softer. Answers from the Recursor cache are cheap.

rgacogne:
Ah, interesting! I guess it would be fine not to go to forwarded-to servers for these queries, right? It would probably be even better.

paddg:
> I guess it would be fine not to go to forwarded-to servers for these queries, right?

Yes :)

PR #12427, opened 2023-01-17
Add a new configure option to initialize automatic variables

rgacogne:
### Short description

The new option, `--enable-auto-var-init`, when enabled, sets the `-ftrivial-auto-var-init` flag when supported by the compiler (GCC 12+, clang 16+) to either:
- `zero`: zero-initializes all automatic variables, and is enabled with `--enable-auto-var-init=yes` or `--enable-auto-var-init=zero`. This can be used as a hardening measure in production, reducing information leakage issues.
- `pattern`: initialize all automatic variables to a pattern that is likely to be detected, like 0xAA, and is enabled via `--enable-auto-var-init=pattern`. This is useful in tests, especially when the cost of sanitizers is too high.

I have not done any performance testing, but the zero option is generally considered to have a less than 5% performance cost.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


fredmorcos (approved):

fredmorcos:
This is awesome, thank you!

omoerbeek (approved):

PR #12434, opened 2023-01-18
rec: unify trace logging 

omoerbeek:
### Short description

1. Plumbing to let `validate.cc` and `aggressive-nsec.cc` use the recursor trace and trec-regex logging

Until now the logging of the DNSSEC validation and aggressive cache code used it's own logging, which has the consequence that prefixed, tracing and indenting all worked differently compared to the regular (trace) logging in rec.

So provide a construct to unify them without exposing syncres internals to the validation code.

2. Consistently use prefix for both trace logging as trace-regex logging.

Some log lines do not prefix the current name and/or some lines could be more clear/compact: e.g.
```
Jan 18 14:20:08 [1] : no TA found for 'www.drijf.net' among 1
Jan 18 14:20:08 [1] : no TA found for 'drijf.net' among 1
Jan 18 14:20:08 [1] : no TA found for 'net' among 1
Jan 18 14:20:08 [1] : got TA for '.'
```
But that's for another PR.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [X] documented the code


pdns/validate.hh
```diff
@@ -73,18 +73,18 @@ struct sharedDNSKeyRecordContentCompare
...
 
-vState validateWithKeySet(time_t now, const DNSName& name, const sortedRecords_t& records, const vector<shared_ptr<RRSIGRecordContent> >& signatures, const skeyset_t& keys, bool validateAllSigs=true);
+vState validateWithKeySet(time_t now, const DNSName& name, const sortedRecords_t& records, const vector<shared_ptr<RRSIGRecordContent> >& signatures, const skeyset_t& keys, const OptLog& log, bool validateAllSigs=true);
 bool isCoveredByNSEC(const DNSName& name, const DNSName& begin, const DNSName& next);
 bool isCoveredByNSEC3Hash(const std::string& h, const std::string& beginHash, const std::string& nextHash);
 bool isCoveredByNSEC3Hash(const DNSName& h, const DNSName& beginHash, const DNSName& nextHash);
-void validateWithKeySet(const cspmap_t& rrsets, cspmap_t& validated, const skeyset_t& keys);
+void validateWithKeySet(const cspmap_t& rrsets, cspmap_t& validated, const skeyset_t& keys, const OptLog& log);
 cspmap_t harvestCSPFromRecs(const vector<DNSRecord>& recs);
 vState getKeysFor(DNSRecordOracle& dro, const DNSName& zone, skeyset_t& keyset);
```
rgacogne, line 82: This definition is now outdated, but since `toysdig` is gone the function can be made `static`.

rgacogne (changes requested):
The changes look good, and I have only minor nits, so I'm OK with merging this PR once the nits have been addressed.

Still it really feels like we could use some `Context` object containing:
- const references to the current qname, qtype, auth
- the current prefix, depth
- mutable references to the current rcode (res) and records (ret), perhaps beenthere as well?

If pretty much all functions received a reference of such an object, we could add/change fields without changing the function signatures, and it might be easier to read. We would need a new copy of the object when we update `depth`, at least, which would impact `prefix` as well. So I guess most functions could take a const reference to the context object, and only `doResolve`, `doResolveNoQNameMinimization` and `beginResolve` would need a new copy?

omoerbeek:
> Still it really feels like we could use some `Context` object containing:

Agreed, this has been on my list for too long already. Once the tracing stuff is done (I have followup stuff brewing) I intend to work on a better `Context` object.

pdns/validate.hh
rgacogne: Oh, sorry, the comment was about `getKeysFor` but I see now that the GH UI made that very unclear!

omoerbeek: Hmm, my change did compile locally..???

omoerbeek: Improvement by accident :-)

rgacogne (approved):

pdns/validate.hh
rgacogne: Oh, so `getKeysFor` can even be removed, nice :D

rgacogne (approved):

PR #12435, opened 2023-01-18
dnsdist: Gracefully handle a failure to create a TLS server context

rgacogne:
### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/libssl.cc
+++ b/pdns/libssl.cc
@@ -804,11 +804,10 @@ bool OpenSSLTLSTicketKey::decrypt(const unsigned char* iv, EVP_CIPHER_CTX* ectx,
 std::unique_ptr<SSL_CTX, void(*)(SSL_CTX*)> libssl_init_server_context(const TLSConfig& config,
                                                                        std::map<int, std::string>& ocspResponses)
 {
-  auto ctx = std::unique_ptr<SSL_CTX, void(*)(SSL_CTX*)>(SSL_CTX_new(SSLv23_server_method()), SSL_CTX_free);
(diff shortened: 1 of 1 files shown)


pdns/libssl.cc
```diff
@@ -806,6 +806,11 @@ std::unique_ptr<SSL_CTX, void(*)(SSL_CTX*)> libssl_init_server_context(const TLS
 {
   auto ctx = std::unique_ptr<SSL_CTX, void(*)(SSL_CTX*)>(SSL_CTX_new(SSLv23_server_method()), SSL_CTX_free);
```
fredmorcos, line 807: While we're at it, is it OK to use `decltype(&SSL_CTX_free)` here?

pdns/libssl.cc
```diff
@@ -806,6 +806,11 @@ std::unique_ptr<SSL_CTX, void(*)(SSL_CTX*)> libssl_init_server_context(const TLS
...
 
+  if (!ctx) {
+    ERR_print_errors_fp(stderr);
+    throw std::runtime_error("Error creating an OpenSSL server context");
```
fredmorcos, line 811: Is it important to have the errors printed here? If not, then why not use `throw pdns::OpenSSL::error(...)`?

fredmorcos (approved):

pdns/libssl.cc
rgacogne: That would be better indeed, I just have to get used to it :)

pdns/libssl.cc
rgacogne: Done!

PR #12441, opened 2023-01-20
Add a configure option to enable LTO

rgacogne:
### Short description

The supported options are:
- `thin`: this is the recommended option for clang, as it is more scalable. It automatically determines the number of CPU cores to use. It requires support in the linker as well.
- `auto`: this is the recommended option for GCC, and a valid fallback for clang when `thin` is not supported. It automatically determines the number of CPU cores to use as well.
- `yes`: this translates to `-flto`, which will use a single CPU core in most cases, but is a valid fallback.
- `no`: LTO is disabled.

For now the default is `no`.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


fredmorcos (approved):
So it falls back to `auto` if `thin` is specified but isn't supported. IIUC this is to accommodate `gcc`?

rgacogne:
> So it falls back to `auto` if `thin` is specified but isn't supported. IIUC this is to accommodate `gcc`?

`gcc` but also perhaps other compilers, yes. My idea was to be able to use `enable-lto=thin` without having to worry about compiler support (`auto` would have been clearer but it was already taken in this context :)).

PR #12451, opened 2023-01-23
CI: Initialize automatic variables to a pattern in our CI 

rgacogne:
### Short description

This PR enables initialization of automatic variables to a pattern in our CI, in order to better catch uninitialized variables.
Enabling this option has two theoretical drawbacks:
- runtime is slightly slower
- it does hide the use of uninitialized variables from memory sanitizers like clang's Memory Sanitizer and Valgrind.

The first point should not matter much for our CI since we are already running ASAN, TSAN and/or UBSAN anyway, and the second does not apply either since we do not enable these sanitizers either (Valgrind is too slow, and memory sanitizer is hard to get right because some cases require instrumenting shared libraries as well). Note that this does not prevent from compiler from reporting the use of uninitialized variables via `-Wuninitialized` or `wmaybe-uninitialized` at compilation time.

### Checklist

I have:
- [ ] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -331,6 +331,7 @@ def ci_auth_configure(c):
                       --prefix=/opt/pdns-auth \
                       --enable-ixfrdist \
                       --enable-fortify-source=auto \
+                      --enable-auto-var-init=pattern \
                       --enable-asan \
                       --enable-ubsan''', warn=True)
     if res.exited != 0:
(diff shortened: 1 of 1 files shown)


omoerbeek:
Are our compilers in CI new enough? Not that it hurts if they are not, but maybe they need an upgrade to make use of this feature.

rgacogne:
clang 12 is fine for `pattern`, and would need a nudge [1] for `zero`.

[1]: `-enable-trivial-auto-var-init-zero-knowing-it-will-be-removed-from-clang` which was removed in clang 16+

omoerbeek (approved):

PR #12462, opened 2023-01-23
OpenSSL3.0: RSA signer

fredmorcos:
### Short description
Support OpenSSL 3.0 in the RSA signer.

### Checklist
I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/opensslsigners.cc
+++ b/pdns/opensslsigners.cc
@@ -33,6 +33,12 @@
 #include <openssl/sha.h>
 #include <openssl/rand.h>
 #include <openssl/rsa.h>
+#if OPENSSL_VERSION_MAJOR >= 3
+#include <openssl/types.h>
+#include <openssl/core_names.h>
+#include <openssl/param_build.h>
+#include <openssl/params.h>
+#endif
 #include <openssl/opensslv.h>
 #include <openssl/err.h>
 #include <openssl/pem.h>
@@ -175,25 +181,29 @@ void openssl_seed()
   RAND_seed((const unsigned char*)entropy.c_str(), 1024);
 }
 
+using BigNum = unique_ptr<BIGNUM, decltype(&BN_clear_free)>;
+
+static auto mapToBN(const std::string& componentName, const std::map<std::string, std::string>& stormap, const std::string& key) -> BigNum
+{
+  const std::string& value = stormap.at(key);
+
+  // NOLINTNEXTLINE(*-cast): Using OpenSSL C APIs.
+  const auto* valueCStr = reinterpret_cast<const unsigned char*>(value.c_str());
+  auto number = BigNum{BN_bin2bn(valueCStr, static_cast<int>(value.length()), nullptr), BN_clear_free};
+  if (number == nullptr) {
+    throw pdns::OpenSSL::error(componentName, "Failed to parse key `" + key + "`");
+  }
+
+  return number;
+}
+
 class OpenSSLRSADNSCryptoKeyEngine : public DNSCryptoKeyEngine
 {
 public:
-  explicit OpenSSLRSADNSCryptoKeyEngine(unsigned int algo) :
-    DNSCryptoKeyEngine(algo), d_key(std::unique_ptr<RSA, decltype(&RSA_free)>(nullptr, RSA_free))
-  {
-    int ret = RAND_status();
-    if (ret != 1) {
-      throw runtime_error(getName() + " insufficient entropy");
-    }
-  }
-
-  ~OpenSSLRSADNSCryptoKeyEngine()
-  {
-  }
+  explicit OpenSSLRSADNSCryptoKeyEngine(unsigned int algo);
 
   [[nodiscard]] string getName() const override { return "OpenSSL RSA"; }
-  [[nodiscard]] int getBits() const override { return RSA_size(d_key.get()) << 3; }
-
+  [[nodiscard]] int getBits() const override;
   void create(unsigned int bits) override;
 
   /**
@@ -226,11 +236,13 @@ public:
   void convertToPEM(std::FILE& outputFile) const override;
 
   [[nodiscard]] storvector_t convertToISCVector() const override;
+
+  // TODO Fred: hash() can probably be completely removed. See #12464.
   [[nodiscard]] std::string hash(const std::string& message) const override;
   [[nodiscard]] std::string sign(const std::string& message) const override;
   [[nodiscard]] bool verify(const std::string& message, const std::string& signature) const override;
   [[nodiscard]] std::string getPublicKeyString() const override;
-  std::unique_ptr<BIGNUM, decltype(&BN_clear_free)> parse(std::map<std::string, std::string>& stormap, const std::string& key) const;
+
   void fromISCMap(DNSKEYRecordContent& drc, std::map<std::string, std::string>& stormap) override;
   void fromPublicKeyString(const std::string& content) override;
   bool checkKey(vector<string>* errorMessages) const override;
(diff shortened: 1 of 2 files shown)


pdns/opensslsigners.cc
```diff
@@ -226,11 +235,13 @@ class OpenSSLRSADNSCryptoKeyEngine : public DNSCryptoKeyEngine
   void convertToPEM(std::FILE& outputFile) const override;
 
   [[nodiscard]] storvector_t convertToISCVector() const override;
+
+  // TODO Fred: ::hash() can probably be completely removed.
```
rgacogne, line 239: Isn't `hash()` used in `makeDSFromDNSKey()`, at least?

pdns/opensslsigners.cc
```diff
@@ -274,7 +332,24 @@ void OpenSSLRSADNSCryptoKeyEngine::create(unsigned int bits)
...
+  auto ctx = KeyContext(EVP_PKEY_CTX_new_from_name(nullptr, "RSA", nullptr), EVP_PKEY_CTX_free);
+  if (ctx == nullptr) {
+    throw pdns::OpenSSL::error(getName(), "Could not initialize context");
+  }
+
+  if (EVP_PKEY_keygen_init(ctx.get()) != 1) {
+    throw pdns::OpenSSL::error(getName(), "Could not initialize keygen");
+  }
+
+  EVP_PKEY* key = nullptr;
```
rgacogne, line 345: Don't we need to call `EVP_PKEY_CTX_set_rsa_keygen_bits(ctx.get(), bits)` and `EVP_PKEY_CTX_set1_rsa_keygen_pubexp(ctx.get(), exponent.get())` first? The documentation states that the default exponent is 65537 anyway, but it seems safer to set it explicitly, especially since we already computed it above.

pdns/opensslsigners.cc
```diff
@@ -285,30 +360,182 @@ void OpenSSLRSADNSCryptoKeyEngine::create(unsigned int bits)
...
+
+#if OPENSSL_VERSION_MAJOR >= 3
+void OpenSSLRSADNSCryptoKeyEngine::getKeyParams(BIGNUM** modulus, BIGNUM** publicExponent, BIGNUM** privateExponent, BIGNUM** prime1, BIGNUM** prime2, BIGNUM** dmp1, BIGNUM** dmq1, BIGNUM** iqmp) const
+{
+  if ((modulus != nullptr) && EVP_PKEY_get_bn_param(d_key.get(), OSSL_PKEY_PARAM_RSA_N, modulus) == 0) {
+    throw pdns::OpenSSL::error(getName(), "Could not get key's modulus (n) parameter");
+  }
+
+  if ((publicExponent != nullptr) && EVP_PKEY_get_bn_param(d_key.get(), OSSL_PKEY_PARAM_RSA_E, publicExponent) == 0) {
+    throw pdns::OpenSSL::error(getName(), "Could not get key's public exponent (e) parameter");
```
rgacogne, line 407: Aren't we leaking `*modulus` here?

pdns/opensslsigners.cc
```diff
@@ -285,30 +360,182 @@ void OpenSSLRSADNSCryptoKeyEngine::create(unsigned int bits)
...
+
+#if OPENSSL_VERSION_MAJOR >= 3
+  BIGNUM* modulus = nullptr;
+  BIGNUM* publicExponent = nullptr;
+  BIGNUM* privateExponent = nullptr;
+  BIGNUM* prime1 = nullptr;
+  BIGNUM* prime2 = nullptr;
+  BIGNUM* dmp1 = nullptr;
+  BIGNUM* dmq1 = nullptr;
+  BIGNUM* iqmp = nullptr;
```
rgacogne, line 536: We are leaking the `BIGNUM*`s in this function, as confirmed with Valgrind.

pdns/opensslsigners.cc
```diff
@@ -406,42 +677,100 @@ int OpenSSLRSADNSCryptoKeyEngine::hashSizeToKind(const size_t hashSize)
...
+    throw pdns::OpenSSL::error(getName(), "Could not create context for signing");
+  }
+
+  if (EVP_DigestSignInit(ctx.get(), nullptr, hasher(), nullptr, d_key.get()) == 0) {
+    throw pdns::OpenSSL::error(getName(), "Could not initialize context for signing");
+  }
+
+  std::size_t signatureLen = 0;
+  // NOLINTNEXTLINE(*-cast): Using OpenSSL C APIs.
+  auto* messageCStr = const_cast<unsigned char*>(reinterpret_cast<const unsigned char*>(message.c_str()));
```
rgacogne, line 694: Using `message.data()` instead of `message.c_str()` might save us a `const_cast`.

pdns/opensslsigners.cc
```diff
@@ -406,42 +677,100 @@ int OpenSSLRSADNSCryptoKeyEngine::hashSizeToKind(const size_t hashSize)
...
+  if (EVP_DigestSign(ctx.get(), nullptr, &signatureLen, messageCStr, message.size()) == 0) {
+    throw pdns::OpenSSL::error(getName(), "Could not get message signature length");
+  }
+
+  signature.resize(signatureLen);
+
+  // NOLINTNEXTLINE(*-cast): Using OpenSSL C APIs.
+  auto* signatureCStr = const_cast<unsigned char*>(reinterpret_cast<const unsigned char*>(signature.c_str()));
```
rgacogne, line 702: Same with `signature.data()` instead of `signature.c_str()`.

pdns/opensslsigners.cc
```diff
@@ -406,42 +677,100 @@ int OpenSSLRSADNSCryptoKeyEngine::hashSizeToKind(const size_t hashSize)
...
+  }
+
+  if (EVP_DigestVerifyInit(ctx.get(), nullptr, hasher(), nullptr, d_key.get()) == 0) {
+    throw pdns::OpenSSL::error(getName(), "Failed to initialize context for verifying signature");
+  }
+
+  // NOLINTNEXTLINE(*-cast): Using OpenSSL C APIs.
+  const int ret = EVP_DigestVerify(ctx.get(), reinterpret_cast<const unsigned char*>(signature.data()), signature.size(), reinterpret_cast<const unsigned char*>(message.data()), message.size());
+  if (ret == 0) {
+    throw pdns::OpenSSL::error(getName(), "Signature did not verify successfully: Either the given message does not match the original message, or the signature has an invalid form");
```
rgacogne, line 739: Should we really throw on failure to validate the signature? Would returning `false` not be enough?

pdns/test-signers.cc
```diff
@@ -85,6 +86,18 @@ static const SignerParams rsaSha256SignerParams = SignerParams{
   .algorithm = DNSSECKeeper::RSASHA256,
   .isDeterministic = true,
 
+#if OPENSSL_VERSION_MAJOR >= 3
```
rgacogne, line 89: Some comment about OpenSSL >= 3.0.0 using a generic key interface and thus storing the key PKCS#8-encoded might be nice?

pdns/opensslsigners.cc
```diff
@@ -285,30 +360,182 @@ void OpenSSLRSADNSCryptoKeyEngine::create(unsigned int bits)
...
+
+  if ((dmq1 != nullptr) && OSSL_PARAM_BLD_push_BN(params_build.get(), OSSL_PKEY_PARAM_RSA_EXPONENT2, dmq1) == 0) {
+    throw pdns::OpenSSL::error(getName(), "Could not create key's second exponent parameter");
+  }
+
+  if ((iqmp != nullptr) && OSSL_PARAM_BLD_push_BN(params_build.get(), OSSL_PKEY_PARAM_RSA_COEFFICIENT1, iqmp) == 0) {
+    throw pdns::OpenSSL::error(getName(), "Could not create key's first coefficient parameter");
+  }
+
+  auto* params = OSSL_PARAM_BLD_to_param(params_build.get());
```
rgacogne, line 513: Valgrind suggests that we are leaking the memory allocated into `params` here.

rgacogne (changes requested):
There are a few things we need to fix before merging this, if I'm not mistaken, but you did an impressive work of working out the new APIs introduced in OpenSSL 3.0.x!

pdns/opensslsigners.cc
fredmorcos: Yes, but it's really a hacky way to use DCKEs for hashing.
So when comes the time to remove support for OpenSSL1.1, we can re-implement `makeDSFromDNSKey()` properly. I created #12464 to track some of the `TODO`s.

rgacogne: That works for me, although I'm unfortunately pretty sure it will take years before we can remove OpenSSL 1.1 support :-(

fredmorcos:
I've made the changes as fixup commits to make them easy to review. Let me know before merging so that I squash them.

rgacogne (changes requested):
Looks great! I have still one small comment but it looks pretty much done.
Out of curiosity I did a few tests with `pdnsutil test-algorithm 10` and it looks like signing/verifying` is a bit slower (~15%) than it was before, which does not surprise me given how the new interface works. I suggest we ignore that for this PR, but later looks at whether this actually matters in production setups. If it does, we might have to cache per-thread `EVP_MD_CTX` objects to amortize the cost.

pdns/opensslsigners.cc
```diff
@@ -558,27 +935,49 @@ void OpenSSLRSADNSCryptoKeyEngine::fromPublicKeyString(const std::string& conten
...
-  if (!n) {
+  // NOLINTNEXTLINE(*-cast): Using OpenSSL C APIs.
+  auto modulusBN = BigNum(BN_bin2bn(reinterpret_cast<unsigned char*>(const_cast<char*>(modulus.c_str())), static_cast<int>(modulus.length()), nullptr), BN_clear_free);
+  if (!modulusBN) {
     throw runtime_error(getName() + " error loading modulus (n) value of public key");
   }
 
-  RSA_set0_key(key.get(), n.release(), e.release(), nullptr);
+#if OPENSSL_VERSION_MAJOR >= 3
+  auto params = makeKeyParams(modulusBN.release(), publicExponentBN.release(), nullptr, nullptr, nullptr, nullptr, nullptr, nullptr);
```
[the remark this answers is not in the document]
fredmorcos: Correct. I'll make the change. Shall I also squash the commits?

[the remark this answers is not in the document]
rgacogne: Yes, please!

fredmorcos:
Reminder to force-merge.

PR #12467, opened 2023-01-24
rec: do not use "message" as key, it has a special meaning to systemd-journal

### LINKED ISSUE #12466: structured-logging-backend=systemd-journal - Some entries erroneously contain multiple "message" fields
- Program: PowerDNS Recursor 
 - Issue type: Bug report

### Short description
After enabling structured-logging, some messages contain multiple "message" fields

### Environment
 - Operating system:  archlinux
 - Software version: 4.8.0
 - Software source: repository

### Steps to reproduce
1. Start powerdns-recursor with structured-logging-backend=systemd-journal
2. pdns will check if an upgrade is needed
3. (with an outdated version) it will log "Upgrade now, see ...." to the journal
4. the above message is (already) truncated, use `journalctl --output json-pretty` will reveal there are 2 MESSAGE fields:
```
{
        "STATUS" : "3",
        "QUERY" : "recursor-4.8.0.security-status.secpoll.powerdns.com",
        "_SYSTEMD_SLICE" : "system.slice",
        "_SYSTEMD_CGROUP" : "/system.slice/pdns-recursor.service",
        "SUBSYSTEM" : "housekeeping",
        "MESSAGE" : [
                "PowerDNS Security Update Mandatory",
                "Upgrade now, see https://doc.powerdns.com/recursor/security-advisories/powerdns-advisory-2023-01.html" 
        ],
        "_COMM" : "pdns_recursor",
```

### Expected behaviour
Please use newlines instead of multiple MESSAGE fields 

### Actual behaviour
An array of MESSAGE fields is logged, which most tooling doesn't support, leading to truncated info

### Other information
https://github.com/systemd/systemd/issues/26175

omoerbeek:
Fixes #12466

A more fundamental approach would be to make sure to not use keys that have special meaning to systemd-journal in the structured logging backend code.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/recursordist/secpoll-recursor.cc
+++ b/pdns/recursordist/secpoll-recursor.cc
@@ -85,7 +85,7 @@ void doSecPoll(time_t* last_secpoll, Logr::log_t log)
 
   g_security_message = security_message;
 
-  auto rlog = vlog->withValues("message", Logging::Loggable(g_security_message), "status", Logging::Loggable(security_status));
+  auto rlog = vlog->withValues("securitymessage", Logging::Loggable(g_security_message), "status", Logging::Loggable(security_status));
   if (g_security_status != 1 && security_status == 1) {
     SLOG(g_log << Logger::Warning << "Polled security status of version " << pkgv << ", no known issues reported: " << g_security_message << endl,
          rlog->info(Logr::Notice, "Polled security status of version, no known issues reported"));


rgacogne (approved):
Looks good to me! A quick search did not reveal a list of keys with special meaning to `journald`, so I guess we will have to do that empirically if we want to.

omoerbeek:
> Looks good to me! A quick search did not reveal a list of keys with special meaning to `journald`, so I guess we will have to do that empirically if we want to.

There is https://www.man7.org/linux/man-pages/man7/systemd.journal-fields.7.html

PR #12477, opened 2023-01-26
rec: less bulk test variations and enable TSAN for bulk tests

omoerbeek:
### Short description

It seems that putting a sleep between the `dnsbulktest` runs in `regression-tests/recursor-test` circumvents the UDP packet loss issues we have been seeing without the sleep.

Still have to find out how and why.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/.github/workflows/build-and-test-all.yml
+++ b/.github/workflows/build-and-test-all.yml
@@ -345,7 +345,7 @@ jobs:
     runs-on: ubuntu-20.04
     strategy:
       matrix:
-        sanitizers: [ubsan+asan]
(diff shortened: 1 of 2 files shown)


omoerbeek:
I have done several runs, all succeeding.

rgacogne (approved):
Looks good to me! I wonder if the sleeps simply prevents some part of the network stack from melting due to the amount of UDP datagrams?

Habbie (approved):

PR #12479, opened 2023-01-26
Add coverity scans to our daily GitHub actions flow, remove them from Circle CI

rgacogne:
### Checklist

I have:
- [ ] compiled this code
- [x] tested this code
- [ ] documented the code


rgacogne:
Rebased to fix a conflict.

.github/workflows/build-and-test-all.yml
```diff
@@ -15,8 +15,11 @@ jobs:
     name: build auth
     runs-on: ubuntu-20.04
     env:
-      UBSAN_OPTIONS: "print_stacktrace=1:halt_on_error=1:suppressions=${{ github.workspace }}/build-scripts/UBSan.supp"
       ASAN_OPTIONS: detect_leaks=0
+      FUZZING_TARGETS: yes
+      SANITIZERS: asan+ubsan
+      UBSAN_OPTIONS: "print_stacktrace=1:halt_on_error=1:suppressions=${{ github.workspace }}/build-scripts/UBSan.supp"
+      UNIT_TESTS: yes
```
fredmorcos, line 22: I'm not following where exactly those get used

fredmorcos (approved):
In general, how it's implemented is fine by me. I don't understand where the added `env` variables in `build-and-test-all` are being used. And, can we take a look at a sample run of `Various daily checks`?

rgacogne:
> I don't understand where the added `env` variables in `build-and-test-all` are being used.

They are used in `tasks.py`, in  `ci_auth_configure`, `ci_rec_configure` and `ci_dnsdist_configure`. The general idea is that I did not want to duplicate the `configure` lines, but Otto rightly remarked that we do not want to submit the unit tests fuzzing targets and backend unit tests to the coverity scan, as it adds a lot of useless noise, so I need to make these optional. We also do not want to build with sanitizers enabled to avoid wasting CPU cycles. Now, we could have passed parameters to the function instead of using the environment, but that's how the sanitizers options are currently handled and I did not want to refactor it at this time, so I used the same mechanism to make the unit tests and fuzzing targets optional.

> And, can we take a look at a sample run of `Various daily checks`?

You can see a test run on my own fork, where I temporarily enabled the coverity bits as part of the normal "push" flow to test: https://github.com/rgacogne/pdns/actions/runs/4015333935

fredmorcos:
Awesome! 👍🏽 Thanks!

Habbie (approved):

rgacogne:
I pushed a commit passing the coverity token via the environment instead of the command-line, and setting `hide=True` on the curl commands. I don't think we care very much about leaking the e-mail address if we mess up so it's still passed on the command-line.

Habbie (approved):

PR #12493, opened 2023-01-31
Only store NSEC3 records in aggressive cache if we expect them to be effective.

omoerbeek:
This PR explores the following idea:

The aggressive cache is not very effective for large NSEC3 domains: each NSEC3 record only covers a relatively small amount of random names.

We only want to put NSEC3 records in the aggressive cache if there is a decent chance a hash will hit it. Hashes are random wrt the corresponding names.

So look at the bitwise common prefix of the owner and next, and do not store the record if the common prefix is long.

We also might want to introduce a switch to completely forget the aggressive cache for the NSEC3 case.

Missing: tests, settings code.

BTW: this also fixes/circumvents a bug in `isMinimallyCoveringNSEC3` as the master one confuses the representations of owner and next.

The common prefix loop can abort if it exceeds the threshold. I'll write that if the general approach is considered good.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [X] documented the code

### DIFF
--- a/pdns/recursordist/aggressive_nsec.cc
+++ b/pdns/recursordist/aggressive_nsec.cc
@@ -244,7 +244,7 @@ static bool commonPrefixIsLong(const string& one, const string& two, size_t boun
       }
       continue;
     }
-    // bytes differ, lets look at the bits
+    // bytes differ, let's look at the bits
     for (ssize_t j = CHAR_BIT - 1; j >= 0; j--) {
       const auto bit1 = byte1 & (1 << j);
       const auto bit2 = byte2 & (1 << j);
--- a/pdns/recursordist/aggressive_nsec.cc
+++ b/pdns/recursordist/aggressive_nsec.cc
@@ -20,6 +20,7 @@
  * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
  */
 #include <cinttypes>
+#include <climits>
 
 #include "aggressive_nsec.hh"
 #include "cachecleaner.hh"
@@ -28,6 +29,7 @@
 #include "validate.hh"
 
 std::unique_ptr<AggressiveNSECCache> g_aggressiveNSECCache{nullptr};
+uint8_t AggressiveNSECCache::s_maxNSEC3CommonPrefix = AggressiveNSECCache::s_default_maxNSEC3CommonPrefix;
 
 /* this is defined in syncres.hh and we are not importing that here */
 extern std::unique_ptr<MemRecursorCache> g_recCache;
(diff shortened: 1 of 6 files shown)


rgacogne (approved):
Looks good to me! I made a couple suggestions but nothing that would prevent merging this PR as it is now.

pdns/recursordist/aggressive_nsec.cc
```diff
@@ -226,21 +228,39 @@ static bool isMinimallyCoveringNSEC(const DNSName& owner, const std::shared_ptr<
...
 }
 
-// This function name is somewhat misleading. It only returns true if the nextHash is ownerHash+2, as is common
-// in minimally covering NXDOMAINs (i.e. the NSEC3 covers hash[deniedname]-1 .. hash[deniedname]+2.
-// Minimally covering NSEC3s for NODATA tend to be ownerHash+1, because they need to prove the name, so they
-// can tell us what types are in the bitmap for that name. For those names, this function returns false.
-// This is on purpose because NODATA denials actually do contain useful information we can reuse later -
-// specifically, the type bitmap for a name that does exist.
-static bool isMinimallyCoveringNSEC3(const DNSName& owner, const std::shared_ptr<NSEC3RecordContent>& nsec)
+static size_t computeCommonPrefix(const string& one, const string& two)
```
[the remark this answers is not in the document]
omoerbeek: See the last line of the PR description :-)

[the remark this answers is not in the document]
rgacogne: Oh! :)

omoerbeek:
I changed the name of the option and tried (with some struggle) to describe it in a way that can be understood. Feedback welcome!

omoerbeek:
One thing I might consider adding: special handling handling of wildcard and closest encounter NSEC3 records

pdns/recursordist/aggressive_nsec.cc
```diff
@@ -226,25 +228,51 @@ static bool isMinimallyCoveringNSEC(const DNSName& owner, const std::shared_ptr<
...
+    const auto byte2 = two.at(i);
+    // shortcut
+    if (byte1 == byte2) {
+      length += CHAR_BIT;
+      if (length > bound) {
+        return true;
+      }
+      continue;
+    }
+    // bytes differ, lets look at the bits
```
rgacogne, line 247:
```suggestion
    // bytes differ, let's look at the bits
```

pdns/recursordist/rec-main.cc
```diff
@@ -2828,6 +2832,7 @@ int main(int argc, char** argv)
     ::arg().setSwitch("extended-resolution-errors", "If set, send an EDNS Extended Error extension on resolution failures, like DNSSEC validation errors") = "no";
 
     ::arg().set("aggressive-nsec-cache-size", "The number of records to cache in the aggressive cache. If set to a value greater than 0, and DNSSEC processing or validation is enabled, the recursor will cache NSEC and NSEC3 records to generate negative answers, as defined in rfc8198") = "100000";
+    ::arg().set("aggressive-cache-max-nsec3-zone-size", "The maximum estimated size of a zone to store NSEC3 records into the aggressive cache") = "2000";
```
rgacogne, line 2835: Doesn't the `2000` value differs from the `1000` in the documentation?

pdns/recursordist/docs/settings.rst
```diff
@@ -42,6 +42,21 @@ In this case the address ``128.66.1.2`` is excluded from the addresses allowed a
...
+``aggressive-cache-max-nsec3-zone-size``
+----------------------------------------
+
+.. versionadded: 4.9.0
+
+- Integer
+- Default: 1000
+
+The maximum (estimated) zone size (number of names) for which to put NSEC3 entries into the aggressive NSEC cache.
+For large zones the effectiveness of the NSEC3 cache is reduced since the names are replaced by hashes, which are random by nature.
```
rgacogne, line 56:
```suggestion
For large zones the effectiveness of the NSEC3 cache is reduced since the names are replaced by hashes, which are randomly distributed by nature.
```

omoerbeek:
I think I pushed the wrong version this morning. This has the latest.

pdns/recursordist/docs/settings.rst
```diff
@@ -42,6 +42,24 @@ In this case the address ``128.66.1.2`` is excluded from the addresses allowed a
...
+----------------------------------------
+
+.. versionadded: 4.9.0
+
+- Integer
+- Default: 2000
+
+The limit for which to put NSEC3 records into the aggressive cache.
+A value of ``n`` means that an NSEC3 record is only put into the aggressive cache if the estimated probability of a random name hitting the NSEC3 record is higher than ``1/n``.
+A higher ``n`` will cause more records to be put into the aggressive cache, e.g. a value of 4000 will cause records to be put in the aggresisve cache even if the estimatyed probability of hiting them is twice as low a the case for ``n=2000``.
```
rgacogne, line 57:
```suggestion
A higher ``n`` will cause more records to be put into the aggressive cache, e.g. a value of 4000 will cause records to be put in the aggressive cache even if the estimated probability of hitting them is twice as low as it would be the case for ``n=2000``.
```

pdns/recursordist/docs/settings.rst
```diff
@@ -42,6 +42,24 @@ In this case the address ``128.66.1.2`` is excluded from the addresses allowed a
...
+A value of 0 means no NSEC3 records will be put into the aggressive cache.
+
+For large zones the effectiveness of the NSEC3 cache is reduced since each NSEC3 record only covers a randomly distributed subset of all possible hashes.
+This setting avoids doing unneccesary work for such large zones.
```
rgacogne, line 61:
```suggestion
For large zones the effectiveness of the NSEC3 cache is reduced since each NSEC3 record only covers a randomly distributed subset of all possible hashes.
This setting avoids doing unnecessary work for such large zones.
```

pdns/recursordist/docs/settings.rst
```diff
@@ -42,6 +42,24 @@ In this case the address ``128.66.1.2`` is excluded from the addresses allowed a
 The number of records to cache in the aggressive cache. If set to a value greater than 0, the recursor will cache NSEC and NSEC3 records to generate negative answers, as defined in :rfc:`8198`.
 To use this, DNSSEC processing or validation must be enabled by setting `dnssec`_ to ``process``, ``log-fail`` or ``validate``.
 
+.. _setting-aggressive-cache-max-nsec3-hit-chance:
```
rgacogne, line 45:
```suggestion
.. _setting-aggressive-cache-max-nsec3-hit-ratio:
```

pdns/recursordist/rec-main.cc
```diff
@@ -2828,6 +2832,7 @@ int main(int argc, char** argv)
     ::arg().setSwitch("extended-resolution-errors", "If set, send an EDNS Extended Error extension on resolution failures, like DNSSEC validation errors") = "no";
 
     ::arg().set("aggressive-nsec-cache-size", "The number of records to cache in the aggressive cache. If set to a value greater than 0, and DNSSEC processing or validation is enabled, the recursor will cache NSEC and NSEC3 records to generate negative answers, as defined in rfc8198") = "100000";
+    ::arg().set("aggressive-cache-min-nsec3-hit-ratio", "The minimun expected hit ratio to store NSEC3 records into the aggressive cache") = "2000";
```
rgacogne, line 2835:
```suggestion
    ::arg().set("aggressive-cache-min-nsec3-hit-ratio", "The minimum expected hit ratio to store NSEC3 records into the aggressive cache") = "2000";
```

rgacogne (approved):
Sorry, I stumbled upon two new typos, but I think these are the last ones!

rgacogne (approved):

Issue #12498, opened 2023-02-02
GitHub action workflows with on: schedule should honor repository variables to allow owners to disable them


jsoref:
- Program: Authoritative, Recursor, dnsdist 
 - Issue type: Bug report

### Short description

* There are 5 daily crons, 4 of which are not cheap
* Because it runs on all forks with actions enabled and there aren't many ways for a repository owner to provide input
* Add an opt-in per file variable to control whether a workflow should do work when run from `on: schedule`

### Environment

 - Operating system: 
 - Software version: 
 - Software source: GitHub PowerDNS repository (or a fork, linked to the fork network or not)

### Steps to reproduce

1. Fork pdns
2. Enable workflow actions 
3. Wait a day
4. Check the actions log

### Expected behaviour
Each scheduled workflow should say "skipped" or similar and use no significant CPU time

### Actual behaviour

https://github.com/jsoref/pdns/actions/runs/4069181815/jobs/7008749997#step:7:6294
(A lot of CPU time was spent and I got a failure notification)

### Other information
* https://docs.github.com/en/actions/learn-github-actions/variables
* `if: github.event.schedule` && (variable)

rgacogne:
Would using `if: github.repository == 'PowerDNS/pdns'` in the steps make sense, instead of a variable? GitHub's documentation [1] suggests it should work fine.

[1]: https://docs.github.com/en/actions/using-jobs/using-conditions-to-control-job-execution

Habbie:
> Would using `if: github.repository == 'PowerDNS/pdns'` in the steps make sense, instead of a variable? GitHub's documentation [1](https://docs.github.com/en/actions/using-jobs/using-conditions-to-control-job-execution) suggests it should work fine.

It would, but the ticket came out of a discussion where we considered that a suboptimal solution. The variable would allow fork owners to opt in if they wanted to.

Habbie:
That said, the current situation is worse, and adding `if: github.repo....` would be a decent temporary solution.

rgacogne:
Alright, I'll add conditions with a variable then.

PR #12519, opened 2023-02-08
CI: Only run scheduled workflows if RUN_SCHEDULED_FLOWS is set to 'yes'

rgacogne:
### Short description

I have tested that the 'normal' flows are still executed on a push, but not that the scheduled ones are still working properly.

Hopefully closes https://github.com/PowerDNS/pdns/issues/12498

Comments welcome!

### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/.github/workflows/build-and-test-all.yml
+++ b/.github/workflows/build-and-test-all.yml
@@ -13,6 +13,7 @@ permissions: # least privileges, see https://docs.github.com/en/actions/using-wo
 jobs:
   build-auth:
     name: build auth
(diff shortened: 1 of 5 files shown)


rgacogne:
> Specifically, my plan (based on talking w/ @Habbie) was that each workflow got its own variable.

My reasoning was based on the assumption that I don't expect anyone to ever wanted to run these workflows outside of the PowerDNS/pdns repository, but I'm OK with your solution (or any solution as long as we stop executing these flows in forks where they are not needed, really) :-).

rgacogne:
Superseded by https://github.com/PowerDNS/pdns/pull/12522

Issue #9335, opened 2020-07-14
Add User-Agent: logging data to protobuf


SrX:
- Program: dnsdist
 - Issue type: Feature request

### Short description
Export the User-Agent for DoH connection via protobuf

### Usecase
Have visibility about the end-client types that making queries to the service.

SrX:
This issue is related to #9102

PR #12520, opened 2023-02-08
dnsdist: Add support for metadata in protobuf messages

rgacogne:
### Short description

Closes https://github.com/PowerDNS/pdns/issues/9335.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/pdns/dnsdist-protobuf.cc
+++ b/pdns/dnsdist-protobuf.cc
@@ -22,6 +22,7 @@
 #include "config.h"
 
 #ifndef DISABLE_PROTOBUF
+#include "base64.hh"
 #include "dnsdist.hh"
 #include "dnsdist-protobuf.hh"
 #include "protozero.hh"
@@ -101,6 +102,11 @@ void DNSDistProtoBufMessage::setEDNSSubnet(const Netmask& nm)
 void DNSDistProtoBufMessage::addTag(const std::string& strValue)
 {
   d_additionalTags.push_back(strValue);
+}
+
+void DNSDistProtoBufMessage::addMeta(const std::string& key, std::string&& value)
+{
+  d_metaTags.push_back({key, std::move(value)});
 }
 
 void DNSDistProtoBufMessage::addRR(DNSName&& qname, uint16_t uType, uint16_t uClass, uint32_t uTTL, const std::string& strBlob)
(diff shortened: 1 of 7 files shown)


pdns/dnsdist-protobuf.cc
```diff
@@ -186,6 +192,134 @@ void DNSDistProtoBufMessage::serialize(std::string& data) const
   }
 
   m.commitResponse();
+
+  for (const auto& [key, value] : d_metaTags) {
```
chbruyand, line 196: Add metas *before* commiting ?

rgacogne: So that's what I initially did, but that doesn't work because the meta tags are not in the response, so we would be trying to add fields after an incomplete field (the message) :-)

pdns/dnsdistdist/docs/rules-actions.rst
```diff
@@ -1316,39 +1316,64 @@ The following actions exist.
...
+  * ``doh-path``: the HTTP path for DoH queries, empty otherwise
+  * ``doh-query-string``: the HTTP query string for DoH queries, empty otherwise
+  * ``doh-scheme``: the HTTP scheme for DoH queries, empty otherwise
+  * ``pool``: the currently selected pool of servers
+  * ``proxy-protocol-value:<TYPE>``: the content of the proxy protocol value of type ``<TYPE>``, if any
+  * ``proxy-protocol-values``: the content of all proxy protocol values as a "<type1>:<value1>,...,<typeN>:<valueN>" string
+  * ``b64-content``: the base64-encoded DNS payload of the current query
+  * ``sni``: the Server Name Indication value for queries received over DoT or DoH. Empty otherwise.
+  * ``tag:<TAG>``: the content of the corresponding ``<TAG>`` if any
+  * ``tags``: the list of all tags, and their values, as a "<key1>:<value1>,...,<keyN>:<valueN>" string
```
wojas, line 1343: We also have a different kinds of 'tags` in the protobuf DNSResponse message. I suggest renaming this metadata field to avoid further tag confusion, perhaps to something like `dnsdist-tag` or some other more descriptive name.

wojas: Also note that a `MetaValue` supports `repeated string stringVal` which can be used instead of concatenating them with a comma. This may make it easier for other components to match on a specific value.

rgacogne: > We also have a different kinds of 'tags' in the protobuf DNSResponse message. I suggest renaming this metadata field to avoid further tag confusion, perhaps to something like `dnsdist-tag` or some other more descriptive name.

I'm not sure that confusion exists in dnsdist, as we only have one kind of tags, and if we start calling these `dnsdist-tag` in this part of the documentation we will likely have to edit the other parts where we are dealing with tags to be consistent.

> Also note that a `MetaValue` supports `repeated string stringVal` which can be used instead of concatenating them with a comma. This may make it easier for other components to match on a specific value.

Right, it will introduce a bit of overhead for most of the types that only need a single value, but it makes sense. I'll change it.

rgacogne:
I pushed a new commit using repeated strings instead of concatenating values into a single strings, and rebased on the current master to (hopefully) have a working CI.

rgacogne:
I pushed two new commits:
- no longer add a traiing `:` after a tag key when the value is empty, which is nicer to read
- new option to export dnsdist tags in the Protocol Buffer `tags` field as well, in addition to the `Meta` field

chbruyand (approved):
last changes lgtm

PR #12524, opened 2023-02-08
OpenSSL 3.0: Stay compatible with OpenSSL 1.0 to Fix CentOS7 builds

fredmorcos:
### Short description
Fix some incorrect usages of OpenSSL 1.1/3.0 functions outside of build=guarded blocks.

### Checklist
I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/opensslsigners.cc
+++ b/pdns/opensslsigners.cc
@@ -287,7 +287,6 @@ private:
   using MessageDigest = std::unique_ptr<EVP_MD, decltype(&EVP_MD_free)>;
 #else
   using Key = std::unique_ptr<RSA, decltype(&RSA_free)>;
-  using MessageDigest = std::unique_ptr<EVP_MD, decltype(&EVP_MD_meth_free)>;
 #endif
 
   Key d_key;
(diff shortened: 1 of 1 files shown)


rgacogne (approved):

Habbie:
How did the test build go?

cmouse:
It could make sense to concentrate all those OpenSSL using things to a separate header file.

rgacogne:
> It could make sense to concentrate all those OpenSSL using things to a separate header file.

I quite like the way it is currently done:
- crypto operations are in `opensslsigners.cc` and `decafsigners.cc`
- TLS ones are in `libssl.cc` and `tcpiohandler.cc`
- random numbers generation in `dns_random.cc`
- etc

It means that our products only loads what they need, which would not be the case if everything went into a single openssl-related header. dnsdist in particular does not care about many of these things.

fredmorcos:
> How did the test build go?

Looks like it worked: https://github.com/fredmorcos/pdns/actions/runs/4126583365

Habbie:
> Looks like it worked: https://github.com/fredmorcos/pdns/actions/runs/4126583365

:tada:

PR #12526, opened 2023-02-09
GH actions: added auth odbc{sqlitle3, mssql} tests. Removed from CircleCI

romeroalx:
### Short description
Added auth regression tests for odbc-sqlitle3 and odbc-mssql to GitHub Actions. Removed from CircleCI.

### Checklist

I have:
- [ ] compiled this code
- [x] tested this code
- [ ] documented the code


tasks.py
```diff
@@ -556,28 +558,92 @@ def test_api(c, product, backend=''):
...
+    godbc_sqlite3 = [
+        'godbc_sqlite3-nodnssec'
+    ],
+    godbc_mssql = [
+        'godbc_mssql',
+        'godbc_mssql-nodnssec',
+        'godbc_mssql-nsec3',
+        'godbc_mssql-nsec3-optout',
+        'godbc_mssql-nsec3-narrow'
+    ],
```
fredmorcos, line 570: Feel free to align these with the table above them.

tasks.py
```diff
@@ -556,28 +558,92 @@ def test_api(c, product, backend=''):
...
+Port=1433
+Database=pdns
+TDS_Version=7.1
+[pdns-mssql-docker-nodb]
+Driver=FreeTDS
+Trace=No
+Server=127.0.0.1
+Port=1433
+TDS_Version=7.1
+__EOF__''')
```
fredmorcos, line 593: I would prefer to use python for something like this. As an example putting the file contents in a variable called `odbc_config` and doing something like the following:

```python
with open("~/.odbc.ini", "w") as f:
    print(odbc_config, file=f)
```

tasks.py
```diff
@@ -556,28 +558,92 @@ def test_api(c, product, backend=''):
...
+
+def setup_godbc_sqlite3(c):
+    c.run('''cat >> ~/.odbc.ini <<-__EOF__
+[pdns-sqlite3-1]
+Driver = SQLite3
+Database = ${PWD}/regression-tests/pdns.sqlite3
+[pdns-sqlite3-2]
+Driver = SQLite3
+Database = ${PWD}/regression-tests/pdns.sqlite32
+__EOF__''')
```
fredmorcos, line 608: Same here.

fredmorcos (changes requested):

tasks.py
fredmorcos: Or of course using `f.write(...)`

romeroalx: Done

tasks.py
romeroalx: Done

tasks.py
```diff
@@ -517,67 +519,147 @@ def test_api(c, product, backend=''):
...
+    "Port=1433\n"
+    "TDS_Version=7.1\n"
+    "\n"
+    "[pdns-sqlite3-1]\n"
+    "Driver = SQLite3\n"
+    "Database = pdns.sqlite3\n"
+    "\n"
+    "[pdns-sqlite3-2]\n"
+    "Driver = SQLite3\n"
+    "Database = pdns.sqlite32\n"
```
fredmorcos, line 610: Perhaps use a [triple quoted string](https://docs.python.org/3/library/stdtypes.html?highlight=triple%20quoted%20string#text-sequence-type-str) here?

fredmorcos (approved):
Other than this one comment everything looks fine to me. However I would still recommend waiting for another review from @Habbie

tasks.py
```diff
@@ -517,67 +519,147 @@ def test_api(c, product, backend=''):
...
     if backend == 'remote':
         ci_auth_install_remotebackend_test_deps(c)
 
     if backend == 'authpy':
         with c.cd('regression-tests.auth-py'):
-            c.run(f'PDNS=/opt/pdns-auth/sbin/pdns_server PDNS2=/opt/pdns-auth/sbin/pdns_server SDIG=/opt/pdns-auth/bin/sdig NOTIFY=/opt/pdns-auth/bin/pdns_notify NSEC3DIG=/opt/pdns-auth/bin/nsec3dig SAXFR=/opt/pdns-auth/bin/saxfr ZONE2SQL=/opt/pdns-auth/bin/zone2sql ZONE2LDAP=/opt/pdns-auth/bin/zone2ldap ZONE2JSON=/opt/pdns-auth/bin/zone2json PDNSUTIL=/opt/pdns-auth/bin/pdnsutil PDNSCONTROL=/opt/pdns-auth/bin/pdns_control PDNSSERVER=/opt/pdns-auth/sbin/pdns_server SDIG=/opt/pdns-auth/bin/sdig GMYSQLHOST=127.0.0.1 GMYSQL2HOST=127.0.0.1 MYSQL_HOST="127.0.0.1" PGHOST="127.0.0.1" PGPORT="5432" WITHKERBEROS=YES ./runtests')
+            c.run(f'{pdns_auth_env_vars} WITHKERBEROS=YES ./runtests')
+        return
+
+    # to avoid sudo cd error
```
Habbie, line 639: I'm not sure this comment makes sense here any more

Habbie (changes requested):
noted one possibly spurious comment. Fred's comment makes sense. Otherwise, great work!

tasks.py
romeroalx: You are correct! I will clean that up now.

tasks.py
Habbie: how about this one?

romeroalx: Yes, I noticed that one an hour ago. Running tests against my fork. I will update this PR once confirmed I do not break anything

Habbie (approved):

PR #12540, opened 2023-02-13
Signers cleanup

fredmorcos:
### Short description
- Fixes some confusing argument ordering and method naming.
- Instead of PEM-related functions only being able to handle files, introduce helpers to handle (import/export from) strings as well.
- Use the new helpers to simplify the tests a little.
- Some documentation formatting.

### Checklist
I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnssecinfra.hh
+++ b/pdns/dnssecinfra.hh
@@ -86,24 +86,22 @@ class DNSCryptoKeyEngine
     /**
      * \brief Creates a key engine from a PEM file.
      *
-     * Receives an open file handle with PEM contents and creates a key
-     * engine corresponding to the algorithm requested.
+     * Receives an open file handle with PEM contents and creates a key engine
+     * corresponding to the algorithm requested.
      *
      * \param[in] drc Key record contents to be populated.
-     *
-     * \param[in] filename Only used for providing filename information
-     * in error messages.
-     *
-     * \param[in] fp An open file handle to a file containing PEM
-     * contents.
      *
      * \param[in] algorithm Which algorithm to use. See
      * https://www.iana.org/assignments/dns-sec-alg-numbers/dns-sec-alg-numbers.xhtml
      *
-     * \return A key engine corresponding to the requested algorithm and
-     * populated with the contents of the PEM file.
+     * \param[in] fp An open file handle to a file containing PEM contents.
+     *
+     * \param[in] filename Only used for providing filename information in error messages.
+     *
+     * \return A key engine corresponding to the requested algorithm and populated with
+     * the contents of the PEM file.
      */
-    static std::unique_ptr<DNSCryptoKeyEngine> makeFromPEMFile(DNSKEYRecordContent& drc, const std::string& filename, std::FILE& inputFile, uint8_t algorithm);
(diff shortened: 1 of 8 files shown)


rgacogne:
>     * Instead of PEM-related functions only being able to handle files, introduce helpers to handle (import/export from) strings as well.

Is this useful outside of `test-signers.cc`? If it isn't, I'm not sure I'm sold on increasing (slightly, granted) the complexity in the signers to make our unit tests simpler.

rgacogne:
Right, if you think it might be useful in the future, and is not just a helper for these tests, that's fine by me!

pdns/dnssecinfra.hh
```diff
@@ -42,19 +42,59 @@ class DNSCryptoKeyEngine
...
+     *
+     * \param[in] drc Key record contents to be populated.
+     *
+     * \param[in] contents The PEM string contents.
+     *
+     * \return A key engine populated with the contents of the PEM string.
+     */
+    void createFromPEMString(DNSKEYRecordContent& drc, const std::string& contents)
+    {
+      unique_ptr<std::FILE, decltype(&std::fclose)> inputFile{fmemopen((void*)contents.data(), contents.length(), "r"), &std::fclose};
```
rgacogne, line 67:
```suggestion
      unique_ptr<std::FILE, decltype(&std::fclose)> inputFile{fmemopen(const_cast<char*>(contents.data()), contents.length(), "r"), &std::fclose};
```

pdns/dnssecinfra.hh
```diff
@@ -42,19 +42,59 @@ class DNSCryptoKeyEngine
...
+     *
+     * \return A string containing the key's PEM contents.
+     */
+    [[nodiscard]] auto convertToPEMString() const -> std::string
+    {
+      const size_t buflen = 4096;
+
+      std::string output{};
+      output.resize(buflen);
+      unique_ptr<std::FILE, decltype(&std::fclose)> outputFile{fmemopen(static_cast<void*>(output.data()), output.length() - 1, "w"), &std::fclose};
```
rgacogne, line 90:
```suggestion
      unique_ptr<std::FILE, decltype(&std::fclose)> outputFile{fmemopen(output.data(), output.length() - 1, "w"), &std::fclose};
```

rgacogne (approved):
One nit but it looks good to me!

rgacogne:
I would prefer not merging this PR until the auth tests in our CI are green, which will likely require rebasing on top of master.

pdns/dnssecinfra.hh
fredmorcos: That needs to be `const_cast<char*>(static_cast<const char*>(contents.data()))`. Fixed.

rgacogne: I don't understand why the `static_cast` is needed?

fredmorcos: Ah damn.. nevermind. For some reason I read your suggested change as a `static_cast` and didn't really question it. Fixed.

rgacogne (approved):

fredmorcos:
Merge reminder :-)

PR #12542, opened 2023-02-13
dnsdist: Fix warnings from cppcheck

rgacogne:
### Short description

Most of these are 'performance' warnings in parts of code where performance does not really matter (or the compiler should already do the right thing without our help), but there is no harm in being more consistent. It also makes it easier to spot more meaningful warnings in the future. The remaining changes are false-positives where `cppcheck` does not detect that null-pointer dereferences cannot actually happen, so I edited the code to make it clearer (removing redundant checks in some cases, actually adding some in other cases).

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


chbruyand (approved):

rgacogne:
Rebased for CI.

PR #12533, opened 2023-02-10
auth lmdb: stop leaving replaced/deleted metadata items behind in the database

Habbie:
### Checklist

I have:
- [x] read the [CONTRIBUTING.md](https://github.com/PowerDNS/pdns/blob/master/CONTRIBUTING.md) document
- [x] compiled this code
- [x] tested this code
- [ ] included documentation (including possible behaviour changes)
- [ ] documented the code
- [x] added or modified regression test(s)
- [ ] added or modified unit test(s)
- [ ]  checked that this code was merged to master

PR #12546, opened 2023-02-14
auth: LMDB schema version 5

Habbie:
### Short description
~~rebase after #12533~~ closed it in favour of this one entirely

TODO (some of these might go into a future PR instead):

* [x] write deleted=1 flag instead of doing deletes
* [x] make "deleted=1 instead of actual deletes" optional
* [x] treat deleted=1 records as absent
* [x] actually put timestamps in the LS header
* [x] introduce version-suffixed subDBs
* [x] migration from v3/v4
* [x] LS header endianness
* [x] skip extra bytes on reading
* [x] document
* [x] unbreak `lmdb-random-ids=no` after overly optimistic removal of `MDB_INTEGERKEY` (specialise MDBInVal/Outval for unsigned int to go to/from network order?)
* [x] unbreak dnsdist usage of LMDB (even if this works locally for me, I suspect I point at `gettime.hh` wrong, should ifdef that out)
* [x] check, beyond just the tests, that `lmdb-safe.*` are basically unchanged when `DNSDIST` is defined (maybe test with `unifdef`?)
* [x] be clean under `-Wall -Wextra -Werror=vla -Werror=shadow -Wformat=2 -Werror=format-security -Werror=string-plus-int` because #12549 avoids touching LMDB so it doesn't conflict with this PR (ALMOST done - unused variable warnings are gone for LMDB, but there are three new warnings that will need some refactoring later)
* [x] rework from idvec to internal dupsort hack (append key+value as key, value is only LSHeader)
* [x] dupsort hack causes inefficient range scans. The regression tests (with `dnssec-parent.com` that has various childs) can, depending on random ID generation, report `scanned=5` (see commented `cerr` in `lmdb-typed.hh`). Suggested improvement: prepend DNSName keyConvs with a length byte or a labelcount byte so the range will only match the correct entry
* [x] my editor automatically adds headers. Many of these are wrong.
* [x] unbreak `lmdb-schema-version` pdns.conf setting (that can prevent startup if the user does not want to upgrade)
* [x] add pdnsutil command to see -expected- schema version

postponed:

* [ ] add `prune` command
* [ ] migration from v1/v2 (might not do this)

### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code


ext/lmdb-safe/lmdb-safe.cc
```diff
@@ -184,6 +185,11 @@ MDB_txn *MDBRWTransactionImpl::openRWTransaction(MDBEnv *env, MDB_txn *parent, i
 MDBRWTransactionImpl::MDBRWTransactionImpl(MDBEnv* parent, int flags):
   MDBRWTransactionImpl(parent, openRWTransaction(parent, nullptr, flags))
 {
+  struct timeval tv;
+
+  gettimeofday(&tv, NULL);
+
+  d_txtime = tv.tv_sec * 1E9 + tv.tv_usec * 1E3;
```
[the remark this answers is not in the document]
Habbie: `pdns/gettime.cc` already solves this, will switch to it

ext/lmdb-safe/lmdb-safe.hh
```diff
@@ -12,9 +14,49 @@
...
+#include <bit>
+#include <byteswap.h>
+#include "../../pdns/misc.hh"
 
 using std::string_view;
 
+#if __BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__
+#define _LMDB_SAFE_BSWAP64MAYBE(x) bswap_64(x)
+#else
+#define _LMDB_SAFE_BSWAP64MAYBE(x) ()
```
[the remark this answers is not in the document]
Habbie: absolutely! fixed :)

ext/lmdb-safe/lmdb-safe.hh
```diff
@@ -1,4 +1,7 @@
 #pragma once
+#include <bits/stdint-uintn.h>
```
wojas, line 2: on macOS:

```
../../ext/lmdb-safe/lmdb-safe.hh:2:10: fatal error: 'bits/stdint-uintn.h' file not found
```

according to https://stackoverflow.com/a/65983825/297793:

`#include <stdint.h>` is the correct way to declare these types. Any other approach is undefined and/or non-portable.

ext/lmdb-safe/lmdb-safe.hh
```diff
@@ -12,9 +15,95 @@
 #include <mutex>
 #include <vector>
 #include <algorithm>
+#include <bit>
+#include <byteswap.h>
```
wojas, line 19: on macOS:

```
../../ext/lmdb-safe/lmdb-safe.hh:19:10: fatal error: 'byteswap.h' file not found
```

wojas: As a workaround I commented the include out and added this from the header file (renamed):

```cpp
#define bswap_64(x)			\
  ((((x) & 0xff00000000000000ull) >> 56)	\
   | (((x) & 0x00ff000000000000ull) >> 40)	\
   | (((x) & 0x0000ff0000000000ull) >> 24)	\
   | (((x) & 0x000000ff00000000ull) >> 8)	\
   | (((x) & 0x00000000ff000000ull) << 8)	\
   | (((x) & 0x0000000000ff0000ull) << 24)	\
   | (((x) & 0x000000000000ff00ull) << 40)	\
   | (((x) & 0x00000000000000ffull) << 56))
```

wojas:
Probably unrelated to this PR, but I get this on macOS:

```
In file included from unixconnector.cc:28:
In file included from ./remotebackend.hh:29:
In file included from ../../pdns/dnsbackend.hh:45:
../../pdns/sha.hh:25:10: fatal error: 'openssl/sha.h' file not found
```

Some configure output related to openssl:

```
checking whether compiling and linking against OpenSSL's libcrypto works... yes
checking for RAND_bytes... yes
checking for RAND_pseudo_bytes... yes
checking for CRYPTO_memcmp... yes
checking for OPENSSL_init_crypto... yes
checking for EVP_MD_CTX_new... yes
checking for EVP_MD_CTX_free... yes
checking for RSA_get0_key... yes
checking whether EVP_PKEY_CTX_set1_scrypt_salt is declared... yes
checking for /usr/local/Cellar/openssl@3/3.0.8/include/openssl/ecdsa.h... yes
...
configure: Configured with:  '--with-modules=bind lmdb pipe remote gsqlite3' '--with-lmdb=/usr/local/Cellar/lmdb/0.9.30' 'CFLAGS=' 'LDFLAGS=' 'PKG_CONFIG_PATH=/usr/local/opt/openssl/lib/pkgconfig'
...
configure: OpenSSL ecdsa: yes
```

wojas:
> unbreak lmdb-random-ids=no after overly optimistic removal of MDB_INTEGERKEY

Is there any point in still supporting sequential IDs?

ext/lmdb-safe/lmdb-safe.hh
Habbie: I switched it to `__builtin_bswap64` which is available in GCC and Clang.

Habbie:
> Is there any point in still supporting sequential IDs?

Excellent question. Maybe around auth 4.8 alpha2 or beta1 we can flip random IDs on by default and see what people say :)

ext/lmdb-safe/lmdb-safe.hh
Habbie: silly, clang tooling invented this #include for me :(

adjusted.

ext/lmdb-safe/lmdb-safe.hh
```diff
@@ -12,9 +15,95 @@
...
+    uint32_t d_reserved;
+    uint16_t d_numextra;
+
+    LSheader(uint64_t timestamp, uint64_t txnid, uint8_t flags=0, uint8_t version=0, uint8_t numextra=0):
+      d_timestamp(_LMDB_SAFE_BSWAP64MAYBE(timestamp)),
+      d_txnid(_LMDB_SAFE_BSWAP64MAYBE(txnid)),
+      d_version(version),
+      d_flags(flags),
+      d_reserved(0),
+      d_numextra(numextra)
```
wojas, line 48: `numextra` also need to be stored in network order.

Passing it as an `uint8_t` does not save you here, because it is stored as `uint16_t` in the wrong byte order.

ext/lmdb-safe/lmdb-safe.hh
```diff
@@ -12,9 +15,95 @@
...
+// maybe put these functions in a namespace with a name?
+
+  // this also returns a pointer to the string's data. Do not hold on to it too long!
+  LSheader* LSassertFixedHeaderSize(std::string &val) {
+    // cerr<<"val.size()="<<val.size()<<endl;
+    if (val.size() < LS_HEADER_SIZE) {
+      throw std::runtime_error("LSheader too short");
+    }
+
+    return (LSheader*)val.data();
```
wojas, line 75: Since the data typically points somewhere into an LMDB page, there is no guarantee that it is 8-byte aligned.

It may be faster to manually assemble all the individual bytes on access, than to use a clever bswap64 only to incur unaligned load penalties on access.

Do other parts of the LMDB backend code assume that the application value is aligned?

ext/lmdb-safe/lmdb-safe.hh
Habbie: ah, yes. I did have `ntohs` on the consuming side already.

ext/lmdb-safe/lmdb-safe.hh
Habbie: > It may be faster to manually assemble all the individual bytes on access, than to use a clever bswap64 only to incur unaligned load penalties on access.

probably true. This was simpler for now :)

> Do other parts of the LMDB backend code assume that the application value is aligned?

I don't think so, but I'm not 100% sure.

wojas: I guess the only fields that are currently accessed are the byte `flags` and `version` fields, or which case it does not matter, and the `numExtra` field for which it does.

ext/lmdb-safe/lmdb-safe.cc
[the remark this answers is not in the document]
Habbie: changed, don't see much difference though

Habbie:
> Is there any point in still supporting sequential IDs?

Turns out development is way more fun with sequential IDs ;)

pdns/pdnsutil.cc
```diff
@@ -2607,6 +2609,20 @@ try
...
   loadMainConfig(g_vm["config-dir"].as<string>());
 
+  if (cmds.at(0) == "lmdb-get-schema-version") {
+    // FIXME: add command to help
+    if(cmds.size() != 2) {
+      cerr << "Syntax: pdnsutil lmdb-get-schema-version /path/to/pdns.lmdb"<<endl;
+      return 0;
+    }
+
+    auto res = LMDBBackend::getSchemaVersionAndShards(cmds.at(1));
```
wojas, line 2619: I expected this command to just print the current schema version of this LMDB backend (constant value) instead of opening an LMDB and reading the schema version of that LMDB.

modules/lmdbbackend/lmdbbackend.cc
```diff
@@ -59,6 +66,570 @@ static bool s_first = true;
...
+  MDB_env* env = nullptr;
+  MDB_dbi dbi;
+  MDB_val key, data;
+  MDB_txn* txn = nullptr;
+
+  if ((rc = mdb_env_create(&env)) != 0) {
+    throw std::runtime_error("mdb_env_create failed");
+  }
+
+  if ((rc = mdb_env_set_maxdbs(env, 20)) != 0) {
```
wojas, line 85: This limit is a bit low when the LMDB can also contain other DBIs.

modules/lmdbbackend/lmdbbackend.cc
```diff
@@ -59,6 +66,570 @@ static bool s_first = true;
...
+    mdb_env_close(env);
+    throw std::runtime_error("mdb_env_set_maxdbs failed");
+  }
+
+  if ((rc = mdb_env_open(env, filename.c_str(), MDB_NOSUBDIR, 0600)) != 0) {
```
wojas, line 90: I suspect that this will also create the LMDB if it does not exist yet?

As the mapsize has not been explicitly set with `mdb_env_set_mapsize`, it will default to 10 MB. A value of 0 would probably prevent this. The mapsize is always automatically adjusted up to a bigger mapsize set in the LMDB.

modules/lmdbbackend/lmdbbackend.cc
```diff
@@ -59,6 +66,570 @@ static bool s_first = true;
...
+
+  if (data.mv_size == 4) {
+    // schemaversion is < 5 and is stored in 32 bits, in host order
+
+    memcpy(&schemaversion, data.mv_data, data.mv_size);
+  }
+  else if (data.mv_size == LS_HEADER_SIZE + sizeof(schemaversion)) {
+    // schemaversion presumably is 5, stored in 32 bits, network order, after the 24 byte LS header
+
+    memcpy(&schemaversion, (char*)data.mv_data + LS_HEADER_SIZE, sizeof(schemaversion));
```
wojas, line 136: Fixed header size assumption

modules/lmdbbackend/lmdbbackend.cc
Habbie: does that matter if pdns is not opening them?

wojas (commented):
I just realised I still had these pending comments waiting for submission.

modules/lmdbbackend/lmdbbackend.cc
```diff
@@ -90,31 +661,46 @@ LMDBBackend::LMDBBackend(const std::string& suffix)
...
-      if (!txn->get(pdnsdbi, "schemaversion", _schemaversion)) {
-        schemaversion = _schemaversion.get<uint32_t>();
+      auto currentSchemaVersionAndShards = getSchemaVersionAndShards(filename);
+      uint32_t currentSchemaVersion = currentSchemaVersionAndShards.first;
+      // std::cerr<<"current schema version: "<<currentSchemaVersion<<", shards="<<currentSchemaVersionAndShards.second<<std::endl;
+
+// FIXME: I accidentally took out the code that checks pdns.conf lmdb-schema-version
+
+      if (currentSchemaVersion < 4) {
+        throw std::runtime_error("this version of the lmdbbackend can only upgrade from schema v4 to v5. Upgrading from older schemas is not yet supported.");
```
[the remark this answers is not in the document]
Habbie: it's v3/v4 to v5 now. I might leave migration from v1/v2 until after alpha1, which indeed is a TODO then.

modules/lmdbbackend/lmdbbackend.cc
```diff
@@ -59,6 +66,570 @@ static bool s_first = true;
...
+    bool done = false;
+
+    if ((rc = mdb_cursor_open(txn, sdbi, &cur)) !=0) {
+      throw std::runtime_error("mdb_cursur_open failed");
+    }
+
+    rc = mdb_cursor_get(cur, &key, &data, MDB_FIRST);
+
+    while (rc == 0) {
+      std::string skey((char*) key.mv_data, key.mv_size);
```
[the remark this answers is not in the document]
Habbie: yes

modules/lmdbbackend/lmdbbackend.cc
```diff
@@ -59,6 +66,570 @@ static bool s_first = true;
...
+      id = htonl(id);
+
+      uint16_t len = htons(skey.size());
+      memcpy((void*) lenprefix.data(), &len, sizeof(len));
+      std::string stkey = lenprefix + skey + std::string((char*)&id, sizeof(uint32_t));
+
+      MDB_val tkey;
+      MDB_val tdata;
+
+      tkey.mv_data = (char*) stkey.c_str();
```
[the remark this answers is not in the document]
Habbie: I never realised that `c_str` might copy. So, yes.

modules/lmdbbackend/lmdbbackend.cc
Habbie: resolved now except in the raw bits

modules/lmdbbackend/lmdbbackend.cc
Habbie: I'm missing something - a shard migration needs 2 dbis, a main db migration needs 9, both numbers are way smaller than 20. Multiple migrations might not fit, depending on what future migrations look like.

pdns/pdnsutil.cc
Habbie: there is `pdns/pdns_server --no-config --config --launch=lmdb | grep schema` but I'll add a separate command that emits a single number

modules/lmdbbackend/lmdbbackend.cc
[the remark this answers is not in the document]
Habbie: Turns out `data` and `c_str` are identical since C++11, so there definitely is no copy. I might still change to `data` to clarify intentions.

pdns/pdnsutil.cc
Habbie: it would in fact be very convenient for me if I can drop the command that inspects an existing LMDB, would that be acceptable? The command that returns the supported version is easy.

Habbie: @nvaatstra

modules/lmdbbackend/lmdbbackend.cc
wojas: If I understand correctly, a migration from v4 to v5 opens 17 DBIs:

- pdns
- 8 unversioned v4 names
- 8 versioned v5 names

A future migration would likely also use 8 new names, so of you would need to migrate from v4 to v6 in two steps in the same process, you would likely need 25 DBIs, unless you are careful to close the old ones after each migration.

Habbie: it's 1+4+4, not 1+8+8. Future migrations will likely be entirely separate functions, especially now that this migration is done in terms of the C API, and thus easy to keep standalone. And if not, we can up the number then :)

ext/lmdb-safe/lmdb-safe.hh
```diff
@@ -511,9 +603,29 @@ public:
...
+    // std::cerr<<"mdb_put txid="<<txid<<std::endl;
+    // txid = bswap_64(txid);
+
+    if (d_txtime == 0) { throw std::runtime_error("got zero txtime"); }
+
+    // MDBInVal inv;
+    std::string ins =
+      // std::string((const char*)&txid, sizeof(txid)) +
+      LSheader(d_txtime, txid).toString()+
+      std::string((const char*)val.d_mdbval.mv_data, val.d_mdbval.mv_size);
```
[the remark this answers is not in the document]
Habbie: we could avoid a few allocs that way, yes. Will have to carefully manage the buffer size - can't really predict maximum value size. I think I'll postpone this until alpha2.

ext/lmdb-safe/lmdb-safe.hh
Habbie: Agreed, should not be hard, but let's do it in alpha2.

modules/lmdbbackend/lmdbbackend.cc
wojas: I think you are forgetting the `_0` DBIs here?

Habbie: you are right! 17 it is. I'll add a comment.

ext/lmdb-safe/lmdb-safe.hh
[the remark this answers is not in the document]
Habbie: indeed, I had similar thoughts

ext/lmdb-safe/lmdb-safe.hh
```diff
@@ -99,54 +166,97 @@ struct MDBOutVal
...
 
   template<class T>
   const T* get_struct_ptr() const
   {
-    if(d_mdbval.mv_size != sizeof(T))
-      throw std::runtime_error("MDB data has wrong length for type");
+    size_t offset = LMDBLS::LScheckHeaderAndGetSize(this, sizeof(T));
 
-    return reinterpret_cast<const T*>(d_mdbval.mv_data);
+    return reinterpret_cast<const T*>((char *)d_mdbval.mv_data+offset);
```
rgacogne, line 228: I'm worried about alignment here. Do we actually use this?

ext/lmdb-safe/lmdb-safe.cc
```diff
@@ -16,6 +21,56 @@ static string MDBError(int rc)
...
+  size_t LScheckHeaderAndGetSize(std::string_view val, size_t datasize) {
+    LSheader* lsh = LSassertFixedHeaderSize(val);
+
+    if (lsh->d_version != 0) {
+      throw std::runtime_error("LSheader has wrong version (not zero)");
+    }
+
+    size_t headersize = LS_MIN_HEADER_SIZE;
+
+    headersize += ntohs(lsh->d_numextra) * LS_BLOCK_SIZE;
```
omoerbeek, line 46: I suppose we trust the LMDB file? Otherwise this need an overflow guard

ext/lmdb-safe/lmdb-safe.hh
```diff
@@ -99,54 +166,97 @@ struct MDBOutVal
...
   T get_struct() const
   {
     T ret;
-    if(d_mdbval.mv_size != sizeof(T))
-      throw std::runtime_error("MDB data has wrong length for type");
 
-    memcpy(&ret, d_mdbval.mv_data, sizeof(T));
+    size_t offset = LMDBLS::LScheckHeaderAndGetSize(this, sizeof(T));
+
+    memcpy(&ret, (char *)d_mdbval.mv_data+offset, sizeof(T));
```
rgacogne, line 219:
```suggestion
    memcpy(&ret, reinterpret_cast<const char *>(d_mdbval.mv_data)+offset, sizeof(T));
```

ext/lmdb-safe/lmdb-safe.hh
```diff
@@ -157,15 +267,19 @@ public:
...
 
+#ifndef DNSDIST
   template <class T,
-            typename std::enable_if<std::is_arithmetic<T>::value,
+            typename std::enable_if<std::is_integral<T>::value,
                                     T>::type* = nullptr>
   MDBInVal(T i)
   {
-    memcpy(&d_memory[0], &i, sizeof(i));
+    auto j = htonl(i);    // all actual usage in our codebase is 32 bits. If that ever changes, this will break the build and avoid runtime surprises
```
rgacogne, line 276: I would really prefer a `static_assert(sizeof(T) == 4, "...")` if we want to rely on that.

rgacogne: I'm really wondering if this is the right place to do that conversion, though, because it makes an assumption about the data that is read that might be application-dependent.

ext/lmdb-safe/lmdb-typed.hh
```diff
@@ -92,6 +94,49 @@ inline std::string keyConv(const T& t)
...
 
 
+namespace {
+  MDBOutVal getKeyFromCombinedKey(MDBInVal combined) {
+    if (combined.d_mdbval.mv_size < sizeof(uint32_t)) {
+      throw std::runtime_error("combined key too short to get ID from");
+    }
+
+    MDBOutVal ret;
+    ret.d_mdbval.mv_data = (char*) combined.d_mdbval.mv_data;
```
rgacogne, line 104:
```suggestion
    ret.d_mdbval.mv_data = combined.d_mdbval.mv_data;
```

ext/lmdb-safe/lmdb-typed.hh
```diff
@@ -92,6 +94,49 @@ inline std::string keyConv(const T& t)
...
+  }
+
+  std::string makeCombinedKey(MDBInVal key, MDBInVal val)
+  {
+    std::string lenprefix(sizeof(uint16_t), '\0');
+    std::string skey((char*) key.d_mdbval.mv_data, key.d_mdbval.mv_size);
+    std::string sval((char*) val.d_mdbval.mv_data, val.d_mdbval.mv_size);
+
+    uint16_t len = htons(skey.size());
+    memcpy((void*) lenprefix.data(), &len, sizeof(len));
```
rgacogne, line 129:
```suggestion
    memcpy(lenprefix.data(), &len, sizeof(len));
```

ext/lmdb-safe/lmdb-typed.hh
```diff
@@ -92,6 +94,49 @@ inline std::string keyConv(const T& t)
...
+    }
+
+    MDBOutVal ret;
+    ret.d_mdbval.mv_data = (char*) combined.d_mdbval.mv_data + combined.d_mdbval.mv_size - sizeof(uint32_t);
+    ret.d_mdbval.mv_size = sizeof(uint32_t);
+
+    return ret;
+  }
+
+  std::string makeCombinedKey(MDBInVal key, MDBInVal val)
```
rgacogne, line 122: If I understand correctly, the combined key is first a variable-length key then an unsigned 32-bits ID? We might want to enforce the size of `val` to be `sizeof(uint32_t)` then?

modules/lmdbbackend/lmdbbackend.cc
```diff
@@ -59,6 +59,568 @@ static bool s_first = true;
...
 
+std::pair<uint32_t, uint32_t> LMDBBackend::getSchemaVersionAndShards(std::string& filename)
+{
+  // cerr << "getting schema version for path " << filename << endl;
+
+  uint32_t schemaversion;
+
+  int rc;
+  MDB_env* env = nullptr;
+  MDB_dbi dbi;
```
omoerbeek, line 72: I prefer to declare vars "as late as possible" so they can either be inited at the spot or in the very next statement

modules/lmdbbackend/lmdbbackend.cc
```diff
@@ -693,21 +1286,24 @@ bool LMDBBackend::deleteDomain(const DNSName& domain)
...
 
   { // Remove cryptokeys
     auto txn = d_tkdb->getRWTransaction();
-    auto range = txn.equal_range<0>(domain);
+    LMDBIDvec ids;
+    txn.get_multi<0>(domain, ids);
 
-    for (auto& iter = range.first; iter != range.second; ++iter) {
-      iter.del();
+    for (auto _id : ids) {
```
omoerbeek, line 1309: This might benefit from a `&`

ext/lmdb-safe/lmdb-typed.hh
```diff
@@ -474,7 +573,7 @@ public:
     {
       typename Parent::cursor_t cursor = (*d_parent.d_txn)->getCursor(std::get<N>(d_parent.d_parent->d_tuple).d_idx);
 
-      std::string keystr = keyConv(key);
+      std::string keystr = makeCombinedKey(keyConv(key), MDBInVal(""));
```
rgacogne, line 576: Ah, so the ID part of the combined key might be empty? How does that work?

rgacogne (commented):
It feels like pretty much all the changes made to files in `ext/lmdb-safe` are lightning-stream specific, while that code was so far quite agnostic. I know it's going to be painful at this point, but I wonder if it would be possible to define new classes that would derive from the ones in `ext/lmdb-safe` instead. Granted, it's a significant amount of work.

Unrelated question, how are entries that are marked as "deleted" ever expunged from the DB?

ext/lmdb-safe/lmdb-safe.hh
```diff
@@ -99,54 +166,97 @@ struct MDBOutVal
...
+
+    size_t offset = LMDBLS::LScheckHeaderAndGetSize(this, sizeof(T));
+
+    memcpy(&ret, (char *)d_mdbval.mv_data+offset, sizeof(T));
+    ret = ntohl(ret);
+    return ret;
+  }
+
+  template <class T,
+          typename std::enable_if<std::is_arithmetic<T>::value,
```
omoerbeek, line 185:
```suggestion
          typename std::enable_if<std::is_integral<T>::value,
```

ext/lmdb-safe/lmdb-safe.hh
```diff
@@ -99,54 +166,97 @@ struct MDBOutVal
...
+                                  T>::type* = nullptr> const
+  T getNoStripHeader()
   {
     T ret;
     if(d_mdbval.mv_size != sizeof(T))
       throw std::runtime_error("MDB data has wrong length for type");
 
     memcpy(&ret, d_mdbval.mv_data, sizeof(T));
+    ret = ntohl(ret);
```
omoerbeek, line 194: This assumes ret has a specific size

ext/lmdb-safe/lmdb-safe.hh
omoerbeek, line 228: potential alignment issue

Habbie:
> It feels like pretty much all the changes made to files in ext/lmdb-safe are lightning-stream specific, while that code was so far quite agnostic. I know it's going to be painful at this point, but I wonder if it would be possible to define new classes that would derive from the ones in ext/lmdb-safe instead. Granted, it's a significant amount of work.

I desperately want to do this, but did not have the time. I hope to do it after the alpha1 release. I got lucky because dnsdist only uses a few small bits.

> Unrelated question, how are entries that are marked as "deleted" ever expunged from the DB?

Presumably, only Lightningstream users would enable that flag, and LS then handles expunging. After alpha1 I'll also add a prune command for users that accidentally turned the flag on.

Habbie:
> I see that formatting is not consistent and also not following our guidelines everywhere: e.g. blocks without braces, missing newlines and missing spaces. Should be fixed some day.

lmdb-safe was not originally "ours", so we couldn't format it. I have add braces in many places. Given how we forked lmdb-safe (honestly, already before this work) a formatting run likely is in order. I'll do that in a separate PR.

> I do worry about alignment issues that are hiding in the code.

Konrad also mentioned a few, I'll have another go at them.

> Also did a test compile on macOS

thanks!

rgacogne:
> I desperately want to do this, but did not have the time. I hope to do it after the alpha1 release. I got lucky because dnsdist only uses a few small bits.

Understood, let's see how we can make that happen after alpha1, then :)

> Presumably, only Lightningstream users would enable that flag, and LS then handles expunging. After alpha1 I'll also add a prune command for users that accidentally turned the flag on.

Alright, that makes sense to me. Thanks!

ext/lmdb-safe/lmdb-safe.cc
```diff
@@ -16,6 +21,56 @@ static string MDBError(int rc)
...
+
+namespace LMDBLS {
+  // this also returns a pointer to the string's data. Do not hold on to it too long!
+  LSheader* LSassertFixedHeaderSize(std::string_view val) {
+    // cerr<<"val.size()="<<val.size()<<endl;
+    if (val.size() < LS_MIN_HEADER_SIZE) {
+      throw std::runtime_error("LSheader too short");
+    }
+
+    return (LSheader*)val.data();
```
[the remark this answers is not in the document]
Habbie: Konrad mentioned the same thing. All accesses into the struct are for single bytes, except in one place where we read a uint16_t. I'll change that one.

pdns/pdnsutil.cc
```diff
@@ -2649,6 +2651,20 @@ try
...
```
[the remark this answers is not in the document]
Habbie: I removed this while you were reviewing, for exactly that reason :)

omoerbeek:
> Konrad also mentioned a few, I'll have another go at them.

or Remi? :-)

Habbie:
> or Remi? :-)

Konrad, then Rémi :)

pdns/pdnsutil.cc
nvaatstra: Yep, no need for that command from my end

pdns/pdnsutil.cc
```diff
@@ -2649,6 +2651,20 @@ try
...
+    cerr << "schemaversion: "<<res.first<<endl;
```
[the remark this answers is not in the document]
Habbie: it was, so that anything reading stdout just got a number. The code is gone now though.

ext/lmdb-safe/lmdb-safe.cc
Habbie: we do (the existing code will also crash if there is garbage in the file), but over time I would indeed like to be more robust against these cases

Habbie: but, in this case there's an `if` right below, is it missing a case?

ext/lmdb-safe/lmdb-safe.hh
Habbie: added the assert.

I agree, it's not the best place, but in the context of our code base, it is currently the simplest place.. (that said, if this really was uint32_t, most applications -would- prefer storage be in network order - and of the two current consumers of this copy of lmdb-safe, one (auth) has conversion in this version, and the other (dnsdist) currently does not read numbers this way)

ext/lmdb-safe/lmdb-typed.hh
Habbie: this is for a range search. Given a key "X", the following keys might exist:

* X|20
* X|634
* X|1231231

this keystr is the prefix that finds all those.

ext/lmdb-safe/lmdb-typed.hh
Habbie: you're right - the "split" functions are explicit about this, but this code is unnecessarily generic

modules/lmdbbackend/lmdbbackend.cc
Habbie: makes sense!

ext/lmdb-safe/lmdb-safe.hh
Habbie: we don't, and we don't even use the `get_struct` variant. (If we did, I'd be surprised at this point, because v5 tries to use network order everywhere so that databases become portable between systems).

I'll remove both methods.

modules/lmdbbackend/lmdbbackend.cc
Habbie: is that actually beneficial for `uint32_t`?

omoerbeek: Likely not indeed. But in the general case I like the compiler to optimize that, as type defs might change.

ext/lmdb-safe/lmdb-safe.hh
Habbie: entire method removed

ext/lmdb-safe/lmdb-safe.cc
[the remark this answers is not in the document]
Habbie: The unaligned 16 bit access is gone.

ext/lmdb-safe/lmdb-safe.cc
```diff
@@ -42,7 +42,9 @@ namespace LMDBLS {
 
     size_t headersize = LS_MIN_HEADER_SIZE;
 
-    headersize += ntohs(lsh->d_numextra) * LS_BLOCK_SIZE;
+    uint16_t numextra = (val[LS_NUMEXTRA_OFFSET] << 8) + val[LS_NUMEXTRA_OFFSET+1];
```
omoerbeek, line 45: As `string_view` holds to chars, the potential signedness can destroy your values:
```
        char a = 0xff;
        char b = 0xff;
        unsigned short x = (a << 8) + b;
        printf("%hx\n", x);
```
Will print `0xffef` if `char` is signed. 
USe
`uint16_t numextra = ((uint16_t)val[LS_NUMEXTRA_OFFSET] << 8) + (uint16_t)val[LS_NUMEXTRA_OFFSET+1];`

omoerbeek, line 45: As `char` can be signed and the expression would go wrong. Use something like
```
unsigned char xxx* = (unsigned char*)val.data();
uint16_t numextra = (xxx[LS_NUMEXTRA_OFFSET] << 8) + xxx[LS_NUMEXTRA_OFFSET+1];
```

omoerbeek (changes requested):

modules/lmdbbackend/lmdbbackend.cc
```diff
@@ -59,6 +61,577 @@ static bool s_first = true;
...
+    throw std::runtime_error("mdb_cursur_open failed");
+  }
+
+  MDB_val key, data;
+
+  rc = mdb_cursor_get(cur, &key, &data, MDB_FIRST);
+
+  while (rc == 0) {
+    std::string skey((char*)key.mv_data, key.mv_size);
+    std::string sdata((char*)data.mv_data, data.mv_size);
```
rgacogne, line 209:
```suggestion
    std::string skey(reinterpret_cast<const char*>(key.mv_data), key.mv_size);
    std::string sdata(reinterpret_cast<const har*>(data.mv_data), data.mv_size);
```

modules/lmdbbackend/lmdbbackend.cc
```diff
@@ -59,6 +61,577 @@ static bool s_first = true;
...
+
+    std::string stdata = header + sdata;
+
+    // cerr<<"got key="<<makeHexDump(skey)<<", data="<<makeHexDump(sdata)<<", sdata="<<makeHexDump(stdata)<<endl;
+
+    MDB_val tkey;
+    MDB_val tdata;
+
+    tkey.mv_data = (char*)skey.c_str();
```
rgacogne, line 218:
```suggestion
    tkey.mv_data = const_cast<char*>(skey.c_str());
```

modules/lmdbbackend/lmdbbackend.cc
```diff
@@ -59,6 +61,577 @@ static bool s_first = true;
...
+    tkey.mv_size = skey.size();
+    tdata.mv_data = (char*)stdata.c_str();
```
rgacogne, line 220:
```suggestion
    tdata.mv_data = const_cast<char*>(stdata.c_str());
```

modules/lmdbbackend/lmdbbackend.cc
```diff
@@ -59,6 +61,577 @@ static bool s_first = true;
...
+    id = htonl(id);
+
+    // cerr<<"got key="<<makeHexDump(skey)<<", data="<<makeHexDump(sdata)<<", sdata="<<makeHexDump(stdata)<<endl;
+
+    MDB_val tkey;
+    MDB_val tdata;
+
+    tkey.mv_data = (char*)&id;
+    tkey.mv_size = sizeof(uint32_t);
+    tdata.mv_data = (char*)stdata.c_str();
```
rgacogne, line 278:
```suggestion
    tdata.mv_data = reinterpret_cast<char*>(stdata.c_str());
```

modules/lmdbbackend/lmdbbackend.cc
```diff
@@ -59,6 +61,577 @@ static bool s_first = true;
...
+    }
+
+    cerr << "shard mbd_drop=" << mdb_drop(shtxn, shdbi, 1) << endl;
+    mdb_txn_commit(shtxn);
+    mdb_dbi_close(shenv, shdbi2);
+    mdb_env_close(shenv);
+  }
+
+  MDB_dbi fromtypeddbi[4];
+  MDB_dbi totypeddbi[4];
```
rgacogne, line 461:
```suggestion
  std::array<MDB_dbi,4> fromtypeddbi;
  std::array<MDB_dbi, 4> totypeddbi;
```

modules/lmdbbackend/lmdbbackend.cc
```diff
@@ -59,6 +61,577 @@ static bool s_first = true;
...
+
+    // mdb_dbi_close(env, dbi2);
+    // mdb_dbi_close(env, dbi);
+    std::cerr << "migrated " << dbname << std::endl;
+
+    index++;
+  }
+
+  MDB_dbi fromindexdbi[4];
+  MDB_dbi toindexdbi[4];
```
rgacogne, line 501:
```suggestion
  std::array<MDB_dbi,4> fromindexdbi;
  std::array<MDB_dbi, 4> toindexdbi;
```

modules/lmdbbackend/lmdbbackend.cc
```diff
@@ -59,6 +61,577 @@ static bool s_first = true;
...
+    catch (std::exception& e) {
+      mdb_dbi_close(shenv, shdbi2);
+      mdb_dbi_close(shenv, shdbi);
+      mdb_txn_abort(shtxn);
+      mdb_env_close(shenv);
+      throw std::runtime_error("copyDBIAndAddLSHeader failed");
+    }
+
+    cerr << "shard mbd_drop=" << mdb_drop(shtxn, shdbi, 1) << endl;
+    mdb_txn_commit(shtxn);
```
rgacogne, line 455: I'm a bit unclear about how how LMDB transactions work. What happens if we commit that transaction but fail later, like while calling `copyTypedDBI`? What's the state of the DB?

modules/lmdbbackend/lmdbbackend.cc
```diff
@@ -59,6 +61,577 @@ static bool s_first = true;
...
+    std::cerr << "migrated " << dbname << std::endl;
+
+    index++;
+  }
+
+  MDB_dbi dbi;
+
+  // finally, migrate the pdns db
+  if ((rc = mdb_dbi_open(txn, "pdns", 0, &dbi)) != 0) {
+    mdb_txn_abort(txn);
```
rgacogne, line 545: Do we need to close the `totypeddbi` and `toindexdbi` DBIs?

modules/lmdbbackend/lmdbbackend.cc
```diff
@@ -59,6 +61,577 @@ static bool s_first = true;
...
+  std::string header(LMDBLS::LS_MIN_HEADER_SIZE, '\0');
+
+  for (const std::string& keyname : {"schemaversion", "shards"}) {
+    cerr << "migrating pdns." << keyname << endl;
+
+    key.mv_data = (char*)keyname.c_str();
+    key.mv_size = keyname.size();
+
+    if ((rc = mdb_get(txn, dbi, &key, &data))) {
+      throw std::runtime_error("mdb_get pdns.shards failed");
```
rgacogne, line 561: Do we need to abort the transaction and close the env?

modules/lmdbbackend/lmdbbackend.cc
```diff
@@ -59,6 +61,577 @@ static bool s_first = true;
...
+    }
+
+    uint32_t value;
+
+    if (data.mv_size != sizeof(uint32_t)) {
+      throw std::runtime_error("got non-uint32_t key");
```
rgacogne, line 567: Same

modules/lmdbbackend/lmdbbackend.cc
```diff
@@ -59,6 +61,577 @@ static bool s_first = true;
...
+  }
+
+  for (const std::string& keyname : {"uuid"}) {
+    cerr << "migrating pdns." << keyname << endl;
+
+    key.mv_data = (char*)keyname.c_str();
+    key.mv_size = keyname.size();
+
+    if ((rc = mdb_get(txn, dbi, &key, &data))) {
+      throw std::runtime_error("mdb_get pdns.shards failed");
```
rgacogne, line 599: Same (sorry!)

modules/lmdbbackend/lmdbbackend.cc
```diff
@@ -59,6 +61,577 @@ static bool s_first = true;
...
+
+    std::string stdata = header + sdata;
+
+    MDB_val tdata;
+
+    tdata.mv_data = (char*)stdata.c_str();
+    tdata.mv_size = stdata.size();
+
+    if ((rc = mdb_put(txn, dbi, &key, &tdata, 0)) != 0) {
+      throw std::runtime_error("mdb_put failed");
```
rgacogne, line 612: And last one :)

rgacogne (commented):
I still have a few questions, mostly because I don't know much about LMDB, sorry about that!
I also made a few suggestions that you should feel completely free to ignore at this point :)

docs/backends/lmdb.rst
```diff
@@ -110,6 +111,14 @@ Size, in megabytes, of each LMDB database.
 This number can be increased later, but never decreased.
 Defaults to 100 on 32 bit systems, and 16000 on 64 bit systems.
 
+.. _settings-lmdb-flag-deleted:
+``lmdb-flag-deleted``
+^^^^^^^^^^^^^^^^^^^^^
+
+  .. versionadded:: 4.8.0
+
+Instead of deleting items from the database, flag them as deleted in the item's [LightningStream](FIXME add link) header.
```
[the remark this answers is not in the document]
Habbie: thanks! the target page does not exist yet :)

modules/lmdbbackend/lmdbbackend.cc
Habbie: This commit really is a full final commit. If we crash later, this at least stays committed.

modules/lmdbbackend/lmdbbackend.cc
Habbie: abort closes DBIs (which are just integer indexes anyway, no interesting associated memory for us to manage)

modules/lmdbbackend/lmdbbackend.cc
Habbie: Yes. This code, and a few other throws, currently rely on the process exiting when this happens (which both pdnsutil and pdns_server do). After alpha1, I want to untangle LS from lmdb-safe a bit, so we can actually use lmdb-safe and lmdb-typed to do the migrations, which will make everything much cleaner.

ext/lmdb-safe/lmdb-safe.hh
```diff
@@ -363,13 +476,101 @@ public:
...
+    }
+
+    // when we get here
+    // * mdb_cursor_get has been called at least once
+    // * it found an entry, as far as LMDB is concerned, so key+data contain something
+    // * but that might be a LS deleted=1 entry
+    // * we know the cursor op that got us here
+
+    while (true) {
+      std::string sval = data.getNoStripHeader<std::string>();
```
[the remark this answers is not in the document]
Habbie: yes!

ext/lmdb-safe/lmdb-safe.hh
```diff
@@ -544,17 +782,19 @@ public:
...
-      throw std::runtime_error("getting data: " + std::string(mdb_strerror(rc)));
-    return rc;
-  }
+    if(rc && rc != MDB_NOTFOUND) {
+          throw std::runtime_error("getting data: " + std::string(mdb_strerror(rc)));
+    }
+
+#ifndef DNSDIST
+    if(rc != MDB_NOTFOUND) {  // key was found, value was retrieved
+      std::string sval = val.getNoStripHeader<std::string>();
```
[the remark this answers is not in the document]
Habbie: indeed!

docs/backends/lmdb.rst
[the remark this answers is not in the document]
Habbie: (now I see what you did.. this remains an unresolved conversation so I don't forget! thanks :D )

ext/lmdb-safe/lmdb-safe.hh
Habbie: did it here anyway

Habbie:
noted all (I hope!) post-merge issues in #12663

PR #12555, opened 2023-02-16
GH Actions: added ldap and geoip-mmdb tests. Removed from CircleCI.

romeroalx:
### Short description
Added LDAP and geoip-mmdb tests to GitHub Actions. Removed from CircleCI. Simplified collector job in build-and-test-all.yml.

### Checklist

I have:
- [ ] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/.github/workflows/build-and-test-all.yml
+++ b/.github/workflows/build-and-test-all.yml
@@ -482,19 +482,22 @@ jobs:
       - test-recursor-api
       - test-recursor-regression
       - test-recursor-bulk
+    if: success() || failure()
     runs-on: ubuntu-20.04
     steps:
+      - name: Install jq and yq
+        run: "sudo snap install jq yq"
+      - name: Fail job if any of the previous jobs failed
+        run: "for i in `echo '${{ toJSON(needs) }}' | jq '.[].result' | tr -d '\"'`; do if [[ $i == 'failure' ]]; then echo '${{ toJSON(needs) }}'; exit 1; fi; done;"
       - uses: actions/checkout@v3.1.0
         with:
           fetch-depth: 5
           submodules: recursive
-      - name: Install yq
-        run: sudo wget https://github.com/mikefarah/yq/releases/download/v4.9.6/yq_linux_amd64 -O /usr/bin/yq && sudo chmod +x /usr/bin/yq
-      - name: Get full list of jobs for this workflow
-        run: yq e '.jobs | keys' .github/workflows/build-and-test-all.yml | grep -v '^- collect' | sort | tee /tmp/workflow-jobs-list.yml
-      - name: Get list of jobs the collect job depends on
-        run: yq e '.jobs.collect.needs | ... comments=""' .github/workflows/build-and-test-all.yml | sort | tee /tmp/workflow-collect-dependencies.yml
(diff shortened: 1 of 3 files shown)


fredmorcos (approved):
It seems fine to me. But as with all things CI please wait for a review from @Habbie

.github/workflows/build-and-test-all.yml
```diff
@@ -482,19 +482,12 @@ jobs:
...
-      - name: Get full list of jobs for this workflow
-        run: yq e '.jobs | keys' .github/workflows/build-and-test-all.yml | grep -v '^- collect' | sort | tee /tmp/workflow-jobs-list.yml
-      - name: Get list of jobs the collect job depends on
-        run: yq e '.jobs.collect.needs | ... comments=""' .github/workflows/build-and-test-all.yml | sort | tee /tmp/workflow-collect-dependencies.yml
-      - name: Diff them
-        run: diff -u /tmp/workflow-jobs-list.yml /tmp/workflow-collect-dependencies.yml
+      - name: Install jq
+        run: "sudo DEBIAN_FRONTEND=noninteractive apt-get install -qq -y jq"
+      - name: Fail job if any of the previous jobs failed
+        run: "for i in `echo '${{ toJSON(needs) }}' | jq '.[].result' | tr -d '\"'`; do if [[ $i == 'failure' ]]; then echo '${{ toJSON(needs) }}'; exit 1; fi; done;"
```
Habbie, line 491: If we ever add a job, and forget to add it to `needs`, you're likely to hit that job as "not yet succeeded" here I think. So I think the diff (or something like it) might still be necessary?

.github/workflows/build-and-test-all.yml
```diff
@@ -216,49 +216,89 @@ jobs:
...
-          POSTGRES_USER: runner
-          POSTGRES_HOST_AUTH_METHOD: trust
-          MYSQL_ALLOW_EMPTY_PASSWORD: 1
-          ACCEPT_EULA: Y
-          SA_PASSWORD: 'SAsa12%%'
-        ports:
-          - 3306:3306
-          - 5432:5432
-          - 1433:1433
+        env: ${{ matrix.env }}
```
Habbie, line 300: nice cleanup!

Habbie (changes requested):
ldap+geoip tests look great. left one comment on the `collect` thing.

.github/workflows/build-and-test-all.yml
romeroalx: Sure, we could add a third step for checking other jobs not included in the needs section. However, bear in mind that if that job has not been included in the needs section, it could (potentially) still be running when the check for other jobs is executed.

Habbie: Right, I see my first comment was incorrect. Your check looks at `needs` so it would never run into the value. Then indeed we need something that checks whether `needs` is still complete.

romeroalx: Added check to fail job also if the list of prerequired jobs is not complete (new jobs added and not reflected there).

Habbie (approved):

PR #12558, opened 2023-02-16
auth: new default param for ifurlup/ifportup lua functions to control…

chbruyand:
… default answer (prior to any check)

### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/lua-record.cc
+++ b/pdns/lua-record.cc
@@ -67,6 +67,7 @@ private:
   struct CheckState
   {
     CheckState(time_t _lastAccess): lastAccess(_lastAccess) {}
+    CheckState(time_t _lastAccess, bool _status): status(_status), lastAccess(_lastAccess) {}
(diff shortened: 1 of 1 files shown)


Habbie:
Want to write docs and a test?

chbruyand:
Closing as this doesn't provide anything useful. First call to `if*up()` function will result in all tests being negative, so we will fallback to use the desired selector on ALL ips.

thestinger:
It's not the same as the case where there are multiple sets though. It uses the combination of all the sets for the fallback instead of the first set. Filed https://github.com/PowerDNS/pdns/issues/12789. I think the behavior should just be changed to make all IPs being up and all IPs being down return the same result (selector on first set) without needing an option to control it. I can't think of a use case for the current behavior over that.

Issue #12560, opened 2023-02-16
dnsdist: no more protobuf messages sent with metadata in some conditions


dmachard:
- Program: dnsdist 1.8.0 (alpha) 
 - Issue type: Bug report

### Description

I am  testing the PR [Add support for metadata in protobuf messages](https://github.com/PowerDNS/pdns/pull/12520) from alpha version of dnsdist with the following configuration

```
setLocal('0.0.0.0:5553')
addDOHLocal('0.0.0.0:5554',   '/etc/dnsdist/conf.d/server.crt',  '/etc/dnsdist/conf.d/server.key', 
            { '/dns-query' },  { keepIncomingHeaders=true })

rl = newRemoteLogger("127.0.0.1:6001")

local mytags = { pool='pool', agent='doh-header:user-agent'   }

addAction(AllRule(),RemoteLogAction(rl, nil, {serverID="dnsdist"}, mytags))
addResponseAction(AllRule(),RemoteLogResponseAction(rl, nil, true, {serverID="dnsdist"}, mytags))
addCacheHitResponseAction(AllRule(), RemoteLogResponseAction(rl, nil, true, {serverID="dnsdist"}, mytags))

newServer({address='1.1.1.1:53', pool='test'})
addAction(AllRule(),PoolAction("test"))
```
I observe something weird, DOH messages are correctly sent to my remote protobuf logger but no more DNS UDP/TCP messages.
If I remove the `doh-header` key from my tags, all DNS messages are correctly sent 

It's the expected behavior ?

### Environment
 - Software version: docker image - dnsdist 1.8.0-alpha0.2664.master.gdb96144a0  (from master branch)

rgacogne:
Hi Denis,
This is completely unexpected and thus clearly a bug! Thank you for testing and reporting that quickly, I'm on it!

rgacogne:
I have not been able to reproduce so far, with your configuration (addresses/ports changed) I'm correctly getting protobuf messages for UDP and TCP queries (output from the Protocol Buffer logger in `contrib/ProtobufLogger.py`:
```
[2023-02-17 10:36:14.249850] Query of size 48: 127.0.0.1:51452 (N/A) -> 127.0.0.1:8053 (UDP) id: 59765 uuid: b'b86a3ea67c114ce0acaa9130ad2cc25b' requestorid: N/A deviceid: N/A devicename: N/A serverid: b'dnsdist' nod: 0
- agent ->
- pool ->
[...]

PR #12561, opened 2023-02-17
dnsdist: Ensure we have at least one protobuf MetaValue

rgacogne:
### Short description

The `MetaValue` is required to exist, which means that we need to have at least one value.

The DoH-related values should be empty non-DoH protocols, but the protocol buffer messages should still be sent, with the expected content.
Fixes https://github.com/PowerDNS/pdns/issues/12560

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code


rgacogne:
Rebased on master to fix CI issues.

fredmorcos (approved):
Looks OK to me. But I'd prefer waiting for a second review.

omoerbeek (approved):

PR #12576, opened 2023-02-23
dnsdist: The security status for 1.8.0-rc1 should be 2, not 3

rgacogne:
### Short description

Which means Update recommended instead of mandatory. Neither are nice, so perhaps we need a different level for pre-releases?

### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code


aerique (approved):
LGTM

Issue #12581, opened 2023-02-23
dnsdist: 1.8.0-rc1 (and current git) confused by any binds?


phonedph1:
- Program: dnsdist
 - Issue type: Bug report

### Environment
 - Operating system: Debian 
 - Software version:  git current
 - Software source: compiled myself

### Steps to reproduce
```
controlSocket("127.0.0.1:5900")
addLocal("0.0.0.0:53")
addLocal("[::]:53")
newServer({address="9.9.9.9"})
```

```
dig foo @127.0.0.2
```

### Expected behaviour
Response from 127.0.0.2 without error in dig.

### Actual behaviour
```
$ dig foo @127.0.0.2
;; reply from unexpected source: 127.0.0.1#53, expected 127.0.0.2#53
```
```
13:06:34.970669 IP 127.0.0.1.50968 > 127.0.0.2.53: 34634+ [1au] A? foo. (44)
13:06:35.243499 IP 127.0.0.1.53 > 127.0.0.1.50968: 34634$ 0/1/1 (130)
```

### Other information
1.8.0-rc1 on RHEL 7 exhibits the same issue. We've also got responses from v6 addresses (hosted on lo) instead of ::1, and from addresses on eth0 where the query was sent to an address on lo.

phonedph1:
This works at 3372321249830832072f28df23cd75e0cc4d0138, trying to work my way forward to find where it breaks

phonedph1:
Actually even this gets responses from the wrong place,

```
--addLocal("0.0.0.0:53")
--addLocal("[::]:53")
addLocal("127.0.0.1:53")
addLocal("127.0.0.2:53")
```

phonedph1:
Maybe something in #12003?

rgacogne:
I suspect either a problem when we harvest the destination address, or that address not being properly passed around during the processing. I'm on it!

rgacogne:
I have reproduction, and it looks like it doesn't happen on self-generated answers (spoof, cache-hits) so it looks like the address is not properly passed around.

PR #12586, opened 2023-02-24
dnsdist: Fix the harvesting of destination addresses

rgacogne:
### Short description

The original destination was not properly updated: `hopLocal` contains the destination address of the packet we received, and matches `origDest` unless the proxy protocol is used, in which case `origDest` will be updated by the 'real' destination address as seen by the client and the first hop.
Reported by @phonedph1 (many thanks!).
Fixes https://github.com/PowerDNS/pdns/issues/12581.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdist.cc
+++ b/pdns/dnsdist.cc
@@ -1165,6 +1165,10 @@ static bool isUDPQueryAcceptable(ClientState& cs, LocalHolders& holders, const s
     dest.sin4.sin_family = 0;
   }
 
+  if (dest.sin4.sin_family == 0) {
+    dest = cs.local;
+  }
+
   ++cs.queries;
   ++g_stats.queries;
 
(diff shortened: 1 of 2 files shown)


pdns/dnsdist.cc
```diff
@@ -1165,6 +1165,10 @@ static bool isUDPQueryAcceptable(ClientState& cs, LocalHolders& holders, const s
     dest.sin4.sin_family = 0;
   }
```
omoerbeek, line 1167: I don't follow this part. In the code above you say you want to set port to 0 in some cases, but here it gets re-assigned to a nonzero value in those cases.

rgacogne: I think you are right: we want to set a fallback for `origDest`, because it might confuse the users, but `hopLocal` should never be set to that fallback, and I changed that by moving that block earlier. I'll fix it and a comment explaining what is going on. Thanks!

phonedph1:
This seems to fix it for me, thanks so much.

omoerbeek (approved):

PR #12587, opened 2023-02-24
dnsdist: Skip signal-unsafe logging when we are about to exit, with TSAN

rgacogne:
### Short description

TSAN is rightfully unhappy about this:
```
WARNING: ThreadSanitizer: signal-unsafe call inside of a signal
```

This is not a real problem for us, as the worst case is that we crash trying to exit, but let's try to avoid the warnings in our tests.
### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdist-idstate.hh
+++ b/pdns/dnsdist-idstate.hh
@@ -89,15 +89,6 @@ private:
   bool d_needRealTime;
 };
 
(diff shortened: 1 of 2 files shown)


omoerbeek (approved):
It would be better to set a flag and the check for that flag, but this is OK for now.

PR #12589, opened 2023-02-24
YaHTTP: Better detection of whether C++11 features are available

rgacogne:
### Short description

The previous version relied on having `HAVE_CXX11` defined, which is not true when you are compiling with C++17, for example, even though the C++11 features are available (`HAVE_CXX17` is defined but that does not help).

I'll open a pull request in the YaHTTP repo as well.

### Checklist

I have:
- [x] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/ext/yahttp/yahttp/reqresp.hpp
+++ b/ext/yahttp/yahttp/reqresp.hpp
@@ -1,4 +1,4 @@
-#ifdef HAVE_CXX11
+#if __cplusplus >= 201103L
 #include <functional>
 #define HAVE_CPP_FUNC_PTR
 namespace funcptr = std;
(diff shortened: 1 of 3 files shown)


rgacogne:
https://github.com/cmouse/yahttp/pull/27

omoerbeek:
There's also https://github.com/cmouse/yahttp/pull/26 , which provides a more fundamental approach.

rgacogne:
That would indeed be better, but I would prefer not diverging too much from upstream, so unless we feel 26 might be merged quickly I would prefer merging this one in the meantime, especially since it was already merged upstream.

omoerbeek:
Yes, agreed.

omoerbeek (approved):

PR #12590, opened 2023-02-24
dnsdist: fix backend servfail metric

phonedph1:
### Short description
Right now servfail-responses (from backends) and frontend-servfail (to clients) are the same metric. Adjust this to only increment servfail-responses if the answer came from something non-self-generated. 

We could also compare the backend server or backend port, similar to how the Lua inspection stuff does, but it seems to me all callers send udiff==0 on cache hit.

### Checklist
I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdist.cc
+++ b/pdns/dnsdist.cc
@@ -639,7 +639,8 @@ void handleResponseSent(const DNSName& qname, const QType& qtype, double udiff, 
     ++g_stats.frontendNXDomain;
     break;
   case RCode::ServFail:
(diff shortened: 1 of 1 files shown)


phonedph1:
image

Before is 1.7.2, then 1.8.0-rc1, then 1.8.0-rc1+this

rgacogne:
Thanks a lot for this pull request! I went with a slightly different approach in #12592 in the end, but your fix would have been fine as well :+1:

Issue #11032, opened 2021-11-24
Ponder compiling our packages with LTO on modern distributions


rgacogne:
- Program: Authoritative, Recursor, dnsdist 
 - Issue type: Feature request

### Short description
Enabling Link-Time Optimizations with a modern compiler is as easy as passing `-flto` (or `-flto=thin` for clang if we want to reduce the amount of memory needed) and allows the compiler to generate more accurate warnings, more optimized code and smaller binaries.
I measured a reduction of at least 15% in binary size for dnsdist by adding `-fvisibility=hidden -flto`. I did not measure the performance impact (although running the unit tests takes slightly less time, but very close to the error margin).

Unless the cost in term of memory usage during the compilation is too high, I'd like to enable LTO in our packages for Bullseye, Focal and perhaps CentOS 8.

omoerbeek:
I did a quick test on OpenBSD and it seems to work and reduce executable size, at the cost of increased link times. Compile times feel quicker. I'd have to check a full build to see how that adds up. I wonder if this is something we want for development work.

rgacogne:
With g++ the linking time is really much longer, so I personally would not want LTO enabled during development. I was only considering our final packages, but perhaps it would make sense in CI as well?

omoerbeek:
Having production executables that differ quite a bit from development ones worries me a bit.

CI executables used should as be close to production as possible.

rgacogne:
> CI executables used should as be close to production as possible.

I fully agree but note that we already have very different binaries since we do CI with sanitizers enabled, for example.

> Having production executables that differ quite a bit from development ones worries me a bit.

I understand your point, and I would like to keep them as close as possible, but we do build for distributions we do not really do development on anyway, so the compiler, kernels and libraries differ already. I don't think we should enable LTO in our packages without testing it in CI, but I'm not too worried about not enabling it for development.

omoerbeek:
> > CI executables used should as be close to production as possible.
> 
> I fully agree but note that we already have very different binaries since we do CI with sanitizers enabled, for example.

I think we also shold do runs without sanitizers. It' would not be the first time that debug aids make things work differently.

zeha:
Debian plans to enable LTO for everything. It's possible Fedora and Ubuntu already did that. I'd be in favour of having CI builds with LTO.

rgacogne:
Note that, for what it's worth, Arch already ships both the authoritative and recursor with LTO enabled.

[...]

PR #12593, opened 2023-02-27
m4: Clean up the fortify and LTO m4 by not directly editing flags

rgacogne:
### Short description

`gl_COMPILER_OPTION_IF` already does the needful, so we do not need to tinker with the flags.
This might help with the issue reported in https://github.com/PowerDNS/pdns/issues/11032 but I'm unfortunately unable to reproduce it, so I'm not sure.

### Checklist

I have:
- [ ] compiled this code
- [x] tested this code
- [ ] documented the code


omoerbeek (approved):
This reads and works good.

PR #12610, opened 2023-03-03
rec: Fix serve-stale loop logic

omoerbeek:
There are several issues here: we do not want to retry on an exception ever as an exception indicates a final failure. e.g. a resolution timeout for the whole query. Recursing in that state will generate an exception storm. Individual timeouts of contacting a nameserver do not generate an exception.  Also, we do not want to recurse if the cache lookup for stale records did not produce anything in the second iteration, we know it's probably going to be fatal and it requires (portentially) a lot of work to find out.

This solves CPU pegging seen.

Check diff ignoring white-space, otherwise it's hard to read.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/recursordist/test-syncres_cc10.cc
+++ b/pdns/recursordist/test-syncres_cc10.cc
@@ -1563,7 +1563,7 @@ BOOST_AUTO_TEST_CASE(test_servestale_cname_to_nxdomain)
   BOOST_CHECK(ret[1].d_type == QType::A);
   BOOST_CHECK_EQUAL(ret[0].d_name, target);
   BOOST_CHECK_EQUAL(ret[1].d_name, DNSName("cname.powerdns.com"));
-  BOOST_CHECK_EQUAL(downCount, 4U);
+  BOOST_CHECK_EQUAL(downCount, 8U);
   BOOST_CHECK_EQUAL(lookupCount, 2U);
 
   // Again, no lookup as the record is marked stale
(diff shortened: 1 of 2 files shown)


pdns/recursordist/test-syncres_cc10.cc
```diff
@@ -1563,7 +1563,7 @@ BOOST_AUTO_TEST_CASE(test_servestale_cname_to_nxdomain)
   BOOST_CHECK(ret[1].d_type == QType::A);
   BOOST_CHECK_EQUAL(ret[0].d_name, target);
   BOOST_CHECK_EQUAL(ret[1].d_name, DNSName("cname.powerdns.com"));
-  BOOST_CHECK_EQUAL(downCount, 4U);
+  BOOST_CHECK_EQUAL(downCount, 8U);
```
rgacogne, line 1566: I don't understand why we now seem to try to reach these servers twice?

omoerbeek: Oops, forgot about that, need to figure out the reason indeed.

omoerbeek: With this PR, we are seeing 8 timeouts. 4 for the `www.powerdns.com|A` query and 4 for the `cname.powerdns.com|A` query. 4 because we have two addresses (A and AAAA) and 2 NS.

With master, I do not see a query `cname.powerdns.com|A`, because the cache consultation stays in "serve-stale" mode (`d_serveStale` remains true for deeper recursions as well), so the cname record is immediately served from the cache with server-stale processing, while with the PR a resolve attempt is done before deciding to serve-stale. I think the second resolve is correct, as we should first try to get a fresh record before falling back to serving stale.

rgacogne: Agreed, this makes more sense!

omoerbeek:
> A unit test highlighting the issue would be nice to have, if possible, but we can do that in a follow-up PR.

Agreed.

PR #12611, opened 2023-03-03
rec: Fix serve-stale logic in negcache by following the record cache case more closely

omoerbeek:
Use same `isEntryUsable()` logic as record cache.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/recursordist/negcache.cc
+++ b/pdns/recursordist/negcache.cc
@@ -141,7 +141,8 @@ bool NegCache::get(const DNSName& qname, QType qtype, const struct timeval& now,
         // Not expired
         ne = *ni;
         moveCacheItemToBack<SequenceTag>(content->d_map, firstIndexIterator);
-        return !refresh; // when refreshing, we consider everything outdated
+        // when refreshing, we consider served-stale entries outdated
+        return !(refresh && ni->d_servedStale > 0);
       }
     }
   }
--- a/pdns/recursordist/negcache.cc
+++ b/pdns/recursordist/negcache.cc
@@ -119,27 +119,31 @@ bool NegCache::get(const DNSName& qname, QType qtype, const struct timeval& now,
 
   const auto& idx = content->d_map.get<NegCacheEntry>();
   auto range = idx.equal_range(qname);
-  auto ni = range.first;
 
-  while (ni != range.second) {
+  for (auto ni = range.first; ni != range.second; ++ni) {
     // We have an entry
     if ((!typeMustMatch && ni->d_qtype == QType::ENT) || ni->d_qtype == qtype) {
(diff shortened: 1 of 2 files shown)


pdns/recursordist/negcache.cc
```diff
@@ -119,27 +119,31 @@ bool NegCache::get(const DNSName& qname, QType qtype, const struct timeval& now,
...
+      if ((serveStale || ni->d_servedStale > 0) && ni->d_ttd <= now.tv_sec && ni->d_servedStale < s_maxServedStaleExtensions) {
         updateStaleEntry(now.tv_sec, firstIndexIterator, qtype);
       }
-      if (now.tv_sec < ni->d_ttd && !(refresh && ni->d_servedStale > 0)) {
+      if (now.tv_sec < ni->d_ttd) {
         // Not expired
         ne = *ni;
         moveCacheItemToBack<SequenceTag>(content->d_map, firstIndexIterator);
-        return true;
+        return !refresh; // when refreshing, we consider everything outdated
```
rgacogne, line 144: Is there a reason we are considering every entry outdated, and not just the ones that are close to expiry as we do in the positive cache?

pdns/recursordist/negcache.cc
```diff
@@ -119,27 +119,31 @@ bool NegCache::get(const DNSName& qname, QType qtype, const struct timeval& now,
...
+      }
+      // If we are serving this record stale (or *should*) and the ttd has passed increase ttd to
+      // the future and remember that we did. Also push a refresh task.
+      if ((serveStale || ni->d_servedStale > 0) && ni->d_ttd <= now.tv_sec && ni->d_servedStale < s_maxServedStaleExtensions) {
         updateStaleEntry(now.tv_sec, firstIndexIterator, qtype);
       }
-      if (now.tv_sec < ni->d_ttd && !(refresh && ni->d_servedStale > 0)) {
+      if (now.tv_sec < ni->d_ttd) {
         // Not expired
         ne = *ni;
```
rgacogne, line 142: Small nit, we might want to skip the copy when `refresh` is `true`.

omoerbeek: I was thinking that as well, but the record cache is doing the same and I wanted to keep the logic similar. Should fix that for both in a future PR.

pdns/recursordist/negcache.cc
omoerbeek: I do remember some complications with that, as in: a non-expired negcache would prevent the whole refresh attempt. In other words: I'm afraid it would break things. The original code has a `ni->d_servedStale` condition though that I think should be reinstated.

rgacogne (approved):

PR #12615, opened 2023-03-03
dnsdist: Add Lua bindings for PB requestorID, deviceName and deviceID

rgacogne:
### Short description

I forgot to add these before, and they are already used by the recursor.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdistdist/dnsdist-lua-ffi.cc
+++ b/pdns/dnsdistdist/dnsdist-lua-ffi.cc
@@ -512,6 +512,39 @@ void dnsdist_ffi_dnsquestion_set_tag_raw(dnsdist_ffi_dnsquestion_t* dq, const ch
   dq->dq->setTag(label, std::string(value, valueSize));
 }
 
+void dnsdist_ffi_dnsquestion_set_requestor_id(dnsdist_ffi_dnsquestion_t* dq, const char* value, size_t valueSize)
+{
+  if (!dq || !dq->dq || !value) {
+    return;
+  }
+  if (!dq->dq->ids.d_protoBufData) {
+    dq->dq->ids.d_protoBufData = std::make_unique<InternalQueryState::ProtoBufData>();
+  }
+  dq->dq->ids.d_protoBufData->d_requestorID = std::string(value, valueSize);
+}
+
+void dnsdist_ffi_dnsquestion_set_device_id(dnsdist_ffi_dnsquestion_t* dq, const char* value, size_t valueSize)
+{
+  if (!dq || !dq->dq || !value) {
+    return;
+  }
+  if (!dq->dq->ids.d_protoBufData) {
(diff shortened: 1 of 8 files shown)


pdns/dnsdistdist/docs/reference/dq.rst
```diff
@@ -59,6 +71,12 @@ This state can be modified from the various hooks.
 
     :ref:`ComboAddress` of the remote client.
 
+  .. attribute:: DNSQuestion.requestorID
+
+    .. versionadded:: 1.8.0
+
+    The identifier of the requestor, which will be exported via ProtoBuf if set.
```
omoerbeek, line 78: in the `from` field.

pdns/dnsdistdist/dnsdist-lua-ffi.cc
```diff
@@ -512,6 +512,39 @@ void dnsdist_ffi_dnsquestion_set_tag_raw(dnsdist_ffi_dnsquestion_t* dq, const ch
   dq->dq->setTag(label, std::string(value, valueSize));
 }
 
+void dnsdist_ffi_dnsquestion_set_requestor_id(dnsdist_ffi_dnsquestion_t* dq, const char* value, size_t valueSize)
+{
+  if (!dq || !dq->dq || !value) {
+    return;
```
omoerbeek, line 518: I do not like this "do nothing silently" much. Wouldn't be better to indicate failure to set with a bool return value? I see that this void return value for setters is common, so something for the future to consider. Another thing to consider: the testing for `dq` and `dq->value` is not done in all setters.

omoerbeek (approved):
Two remarks, reads good.

pdns/dnsdistdist/docs/reference/dq.rst
rgacogne: Isn't that in the `requestorId` field?

pdns/dnsdistdist/dnsdist-lua-ffi.cc
rgacogne: I agree this isn't great. We have added these checks after noticing that crashing dnsdist because of a typo in the Lua code was terrible (a `nil` value becomes a `nullptr`).
Yielding an exception in the FFI API doesn't work well at the moment, so we could indeed return a boolean instead, but it would make handling the potential errors in the Lua code quite cumbersome.

pdns/dnsdistdist/docs/reference/dq.rst
omoerbeek: Hmm, right. I mixed that field up with `d_requestor` used in the `m.setRequest` call. These names are too similar :-(

rgacogne: You are right, `requestorId` did not exist initially but now it's confusing indeed.

Issue #12645, opened 2023-03-13
DNSDIST restarts when DoH query made


allahshukur-ahmadzada:
When running DoH query agains dnsdist which kdig for first query I get result but  right after it if I do second query I get error and when I check dnsdist logs I see its restartted with below logs

```
Mar 13 14:21:49 aws-nameserver dnsdist[2026984]: Got an error in DOH question thread while parsing a query from 162.219.4.28:52452, id 0: Attempt to print an unset dnsname
Mar 13 14:21:49 aws-nameserver dnsdist[2026984]: dnsdist: /build/h2o-tgUKZx/h2o-2.2.5+dfsg2/lib/http2/stream.c:339: finalostream_send: Assertion `stream->_data.size == 0' failed.
Mar 13 14:21:51 aws-nameserver dnsdist[2027066]: Configuration '/etc/dnsdist/dnsdist.conf' OK!
Mar 13 14:21:51 aws-nameserver dnsdist[2027066]: Configuration '/etc/dnsdist/dnsdist.conf' OK!
Mar 13 14:21:51 aws-nameserver dnsdist[2027067]: Read configuration from '/etc/dnsdist/dnsdist.conf'
Mar 13 14:21:51 aws-nameserver dnsdist[2027067]: Loading default TLS provider openssl
```
kdig query:

> kdig -d @IP +https=hostname example.com

dnsdist version:
```
dnsdist 1.7.3 (Lua 5.1.4 [LuaJIT 2.1.0-beta3])
Enabled features: cdb dns-over-tls(gnutls openssl) dns-over-https(DOH) dnscrypt ebpf fstrm ipcipher libsodium lmdb outgoing-dns-over-https(nghttp2) protobuf re2 recvmmsg/sendmmsg snmp systemd
```

dnsdist config which forwards qeuries to PowerDNS recursor:
```
addTLSLocal('172.31.4.54', '/etc/dnsdist/certs/cert.pem', '/etc/dnsdist/certs/key.pem')
addDOHLocal('172.31.4.54', '/etc/dnsdist/certs/cert.pem', '/etc/dnsdist/certs/key.pem')
addACL("0.0.0.0/0")
newServer({
 address="172.31.4.54:53"
})
```
OS:

> Ubuntu 20.04.5 LTS

Habbie:
Looking at the code, the last message is a `vinfolog`, so I asked to try without `-v` - the crash is gone.

rgacogne:
[...]

PR #12646, opened 2023-03-14
dnsdist: Use toLogString for the qname to prevent an error if it's empty

rgacogne:
### Short description

This will likely fix https://github.com/PowerDNS/pdns/issues/12645, even though it's not clear to me how we end up with an unset qname.
Note that master already use `toLogString()` since that code has been refactored in the meantime.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdistdist/doh.cc
+++ b/pdns/dnsdistdist/doh.cc
@@ -771,7 +771,7 @@ static void processDOHQuery(DOHUnitUniquePtr&& du)
       throw;
     }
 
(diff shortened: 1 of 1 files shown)


Habbie (approved):

rgacogne:
> This will likely fix #12645, even though it's not clear to me how we end up with an unset qname. 

One possibility is that we have already received the response from the backend when we process that log message. While unlikely this might happen if we are running on a box with a single CPU core, for example, so threads are effectively serialized. 1.8.0 handles that much better anyway, by logging _before_ the query has been sent to the backend.

PR #12648, opened 2023-03-14
dnsdist: Report the TCP latency for TCP-only Do53, DoT and DoH backends

rgacogne:
### Short description

Since 1.8.0 we separately record the latency for queries forwarded over UDP and TCP, to get a better picture of what is going on. It means we have to be careful to pick the relevant one when looking at the latency.
This is a follow-up on https://github.com/PowerDNS/pdns/pull/12638.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


rgacogne (commented):
I can't approve my own PR but thanks for your changes, Charles-Henri!

rgacogne:
Ah, `test_BackendDiscovery.py` is broken because it parses the output of `showServers` which is modified in this PR. I'll fix it a bit later.

chbruyand:
> Ah, `test_BackendDiscovery.py` is broken because it parses the output of `showServers` which is modified in this PR. I'll fix it a bit later.

Sorry I didn't catch this one!

chbruyand (approved):

PR #12649, opened 2023-03-14
Run `clang-tidy` on auth, rec and dnsdist when opening PRs

fredmorcos:
### Short description
Run `clang-tidy` on auth, rec and dnsdist when opening PRs.

This adds a new Github Actions (CI) workflow with 3 jobs that configure, build and lint (using `clang-tidy`) the auth, rec and dnsdist. The lint messages are then posted as Github annotations which [show up on the summary page of the action's run](https://github.com/fredmorcos/pdns/actions/runs/4415458994), and also [show up as code comments on the PR's diff](https://github.com/fredmorcos/pdns/pull/1/files). Next to the annotations, "user-friendly" messages are also output in the console in case that's more convenient, here are examples for [auth](https://github.com/fredmorcos/pdns/actions/runs/4415458994/jobs/7738451560), [rec](https://github.com/fredmorcos/pdns/actions/runs/4415458994/jobs/7738451725) and [dnsdist](https://github.com/fredmorcos/pdns/actions/runs/4415458994/jobs/7738451309).

The annotations are produced by a python script which reads the messages from the `clang-tidy` YAML output and exports them to the Github format. The compilation database for each product is generated using `bear` through `pyinvoke` tasks.

This also cleans up the auth and rec pyinvoke configure tasks.

### Checklist
I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/.github/workflows/clang-tidy.yml
+++ b/.github/workflows/clang-tidy.yml
@@ -108,3 +108,55 @@ jobs:
           if [ -f clang-tidy-results/rec-fixes.yml ]; then
             python .github/scripts/clang-tidy.py --fixes-file clang-tidy-results/rec-fixes.yml
           fi
+
+  clang-tidy-dnsdist:
+    name: dnsdist clang-tidy
+    runs-on: ubuntu-20.04
+    env:
+      UNIT_TESTS: yes
+      SANITIZERS:
+    steps:
+      - uses: PowerDNS/pdns/set-ubuntu-mirror@meta
+      - uses: actions/checkout@v3.1.0
+        with:
+          fetch-depth: 2
+      - name: get timestamp for cache
+        id: get-stamp
+        shell: bash
+        run: |
+          echo "stamp=$(/bin/date +%s)" >> "$GITHUB_OUTPUT"
+      - name: let GitHub cache our ccache data
+        uses: actions/cache@v3.2.5
+        with:
+          path: ~/.ccache
+          key: dnsdist-full-ccache-${{ steps.get-stamp.outputs.stamp }}
+          restore-keys: dnsdist-full-ccache-
+      - run: build-scripts/gh-actions-setup-inv  # this runs apt update+upgrade
+      - run: inv apt-fresh
+      - run: inv install-clang
+      - run: inv install-clang-tidy-tools
+      - run: inv install-dnsdist-build-deps
+      - run: inv ci-autoconf
+        working-directory: pdns/dnsdistdist/
+      - run: inv ci-dnsdist-configure full
+        working-directory: pdns/dnsdistdist/
+      - run: inv ci-dnsdist-make-bear
+        working-directory: pdns/dnsdistdist/
+      - run: ccache -s
+      - run: mkdir clang-tidy-results
+      - run: ln -s .clang-tidy.full .clang-tidy
+      - name: Run clang-tidy
+        working-directory: pdns/dnsdistdist/
+        run: git diff -U0 HEAD^ | python /usr/bin/clang-tidy-diff-12.py -p3 -export-fixes ../../clang-tidy-results/dnsdist-fixes.yml
+      - name: Print clang-tidy fixes YAML
+        shell: bash
+        run: |
+          if [ -f clang-tidy-results/dnsdist-fixes.yml ]; then
+            cat clang-tidy-results/dnsdist-fixes.yml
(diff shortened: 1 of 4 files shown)


.github/workflows/clang-tidy.yml
```diff
@@ -0,0 +1,151 @@
...
+        id: get-stamp
+        shell: bash
+        run: |
+          echo "stamp=$(/bin/date +%s)" >> "$GITHUB_OUTPUT"
+      - name: let GitHub cache our ccache data
+        uses: actions/cache@v3.2.5
+        with:
+          path: ~/.ccache
+          key: auth-ccache-${{ steps.get-stamp.outputs.stamp }}
+          restore-keys: auth-ccache-
```
rgacogne, line 33: I wonder if we need a different cache key than the one we used in other workflows, given than we build with different options (no sanitizers)?

.github/workflows/clang-tidy.yml
```diff
@@ -0,0 +1,151 @@
...
+jobs:
+  clang-tidy-auth:
+    name: auth clang-tidy
+    runs-on: ubuntu-20.04
+    env:
+      UNIT_TESTS: yes
+      SANITIZERS:
+    steps:
+      - uses: PowerDNS/pdns/set-ubuntu-mirror@meta
+      - uses: actions/checkout@v3.1.0
```
rgacogne, line 20:
```suggestion
      - uses: actions/checkout@v3
```

.github/workflows/clang-tidy.yml
```diff
@@ -0,0 +1,151 @@
...
+        with:
+          fetch-depth: 2
+      - name: get timestamp for cache
+        id: get-stamp
+        shell: bash
+        run: |
+          echo "stamp=$(/bin/date +%s)" >> "$GITHUB_OUTPUT"
+      - name: let GitHub cache our ccache data
+        uses: actions/cache@v3.2.5
```
rgacogne, line 29:
```suggestion
        uses: actions/cache@v3
```

.github/workflows/clang-tidy.yml
```diff
@@ -0,0 +1,151 @@
...
+
+  clang-tidy-rec:
+    name: rec clang-tidy
+    runs-on: ubuntu-20.04
+    env:
+      UNIT_TESTS: yes
+      SANITIZERS:
+    steps:
+      - uses: PowerDNS/pdns/set-ubuntu-mirror@meta
+      - uses: actions/checkout@v3.1.0
```
rgacogne, line 64:
```suggestion
      - uses: actions/checkout@v3
```

.github/workflows/clang-tidy.yml
fredmorcos: Isn't this already a difference cache key? The ones from `build-and-test-all.yml` have the sanitizer name in them which doesn't exist here.

rgacogne: They do for dnsdist and the recursor, but not for the authoritative, unless I'm mistaken?

fredmorcos:
> * do we actually need a separate workflow? I'm wondering if we might be able to run `bear`, generate the clang-tidy fixes `YAML`, the annotations and set some output as an additional step, then process the output in a different job, all of that in `build-and-test-all.py`? It might not make sense at all, just tell me so in that case :)

Makes sense. I'll give it a shot.

.github/workflows/clang-tidy.yml
fredmorcos: Ah, correct.

fredmorcos:
> > * do we actually need a separate workflow? I'm wondering if we might be able to run `bear`, generate the clang-tidy fixes `YAML`, the annotations and set some output as an additional step, then process the output in a different job, all of that in `build-and-test-all.py`? It might not make sense at all, just tell me so in that case :)
> 
> Makes sense. I'll give it a shot.

Actually processing the output is not time consuming at all. The most time consuming part of the clang-tidy specific process is running clang-tidy itself. What we can do is store the produced compilation database and run clang-tidy in a different job. I'm not sure it's worth the hassle though. @Habbie?

rgacogne:
Right, I was mostly worried about compiling everything twice, and I only mentioned processing the output in a different job so that clang-tidy errors would get a separate status in the checklist, but that might not be worth it.

PR #12656, opened 2023-03-16
 auth: Prevent a race during the processing of SVC auto-hints

rgacogne:
### Short description

When `svc-autohints` is enabled, the content of SVCB and HTTPS records is modified in `PacketHandler::doAdditionalProcessing()` to expand the IPv4 and IPv6 with their actual values.
This causes an issue because the content of these records might be shared between threads, via the record cache, and one thread could be trying to read from the internal `std::set` while a second thread is altering it, leading to a data race and possibly to memory corruption and a crash.
This is correctly detected by TSAN:
```
WARNING: ThreadSanitizer: data race (pid=102795)
  Write of size 8 at 0x7b3400010350 by thread T33:
    #0 operator delete(void*) <null> (pdns_server+0x211b7c) (BuildId: 384adc19a67695435bd5e89d0a77f562561f4593)
    #1 std::__new_allocator<std::_Rb_tree_node<SvcParam>>::deallocate(std::_Rb_tree_node<SvcParam>*, unsigned long) /usr/bin/../lib64/gcc/x86_64-pc-linux-gnu/12.2.1/../../../../include/c++/12.2.1/bits/new_allocator.h:158:2 (pdns_server+0x33fc78) (BuildId: 384adc19a67695435bd5e89d0a77f562561f4593)
    #2 std::allocator_traits<std::allocator<std::_Rb_tree_node<SvcParam>>>::deallocate(std::allocator<std::_Rb_tree_node<SvcParam>>&, std::_Rb_tree_node<SvcParam>*, unsigned long) /usr/bin/../lib64/gcc/x86_64-pc-linux-gnu/12.2.1/../../../../include/c++/12.2.1/bits/alloc_traits.h:496:13 (pdns_server+0x33fc78)
    #3 std::_Rb_tree<SvcParam, SvcParam, std::_Identity<SvcParam>, std::less<SvcParam>, std::allocator<SvcParam>>::_M_put_node(std::_Rb_tree_node<SvcParam>*) /usr/bin/../lib64/gcc/x86_64-pc-linux-gnu/12.2.1/../../../../include/c++/12.2.1/bits/stl_tree.h:565:9 (pdns_server+0x33fc78)
[...]

PR #12664, opened 2023-03-17
DNSRecord: Ensure that the content can be read or replaced, not edited

rgacogne:
### Short description

This is a huge PR, sorry about that, but the individual changes are actually very small.
Mostly, this makes `DNSRecord::d_content` private and the content of the actual pointer `const`, and provides just enough accessors to get a read-only version or replace the existing content with a new one, thus preventing actually editing the content itself, as it might be shared with other threads (see #12659).
It also addresses the longstanding issue that `DNSRecordContent::serialize()` was not `const`, which led to quite ugly work-arounds. 
It also reverts the fix from #12659 to use the "copy" approach from #12656 since the main issue, editing the content, now requires explicitly working around the new mechanism which should not happen by mistake.
It will also make it a bit easier to try to replace these `std::shared_ptr` with `std::unique_ptr` later, should we decide to give it a shot. 

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/modules/lmdbbackend/lmdbbackend.cc
+++ b/modules/lmdbbackend/lmdbbackend.cc
@@ -868,7 +868,7 @@ bool LMDBBackend::get(DNSZoneRecord& zr)
         zr.domain_id = compoundOrdername::getDomainID(key);
         zr.dr.d_type = compoundOrdername::getQType(key).getCode();
         zr.dr.d_ttl = lrr.ttl;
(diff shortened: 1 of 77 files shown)


omoerbeek (approved):
Like it very much!

rgacogne:
Merging this now since auth-4.8.0-alpha1 has been released, rec is also in a good place for such a change and dnsdist is pretty much not concerned.

PR #12673, opened 2023-03-22
rec: Sanitize d_orig_ttl stored in record cache

omoerbeek:
The computed `d_orig_ttl` can be wrong if `d_now.tv_sec` is more recent than the `d_now.tv_sec` the TTD was computed from. This can happen if we went out to get e.g. keys and that took time. `d_orig_ttl` even potentially wraps around in that case if the TTL value was smaller than that delay. Work around that and make sure `d_orig_ttl` is within the legal range.

An alternative approach would be to store `d_orig_ttl` the moment we change the `d_ttl` to be a TTD. I looked into that but we have no convenient place to store it. There are also a few places the TTD is replaced yet again with another value, so things started to get quite complicated. But maybe I missed an approach to have the real original TTL stored somewhere. I'll keep looking into that, as that is a better approach in the end. Hence the draft status.

Note: 2nd commit takes the alternate approach. I jave left the original approach in between `#if 0/#endif` markers. Those blocks will be removed if the 2nd approach is deemed OK.

### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/.clang-tidy.full
+++ b/.clang-tidy.full
@@ -113,7 +113,7 @@ CheckOptions:
   - key:             readability-function-size.LineThreshold
     value:           '4294967295'
   - key:             bugprone-easily-swappable-parameters.MinimumLength
-    value:           '2'
+    value:           '4'
   - key:             portability-simd-intrinsics.Suggest
     value:           'false'
   - key:             cppcoreguidelines-pro-bounds-constant-array-index.GslHeader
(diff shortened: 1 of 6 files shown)


rgacogne (approved):
I agree with Peter, both options look fine. The second one is more accurate but also more complicated, I wonder if it would not be easier to record the time when we computed the TTD instead of the original TTL, so we would in theory not need to "compensate" the value later?

omoerbeek:
> I agree with Peter, both options look fine. The second one is more accurate but also more complicated, I wonder if it would not be easier to record the time when we computed the TTD instead of the original TTL, so we would in theory not need to "compensate" the value later?

That looks like a viable thing to investigate, will do so.

PR #12677, opened 2023-03-23
auth API: rectify after bumping serial, so the SOA record also gets rectified

Habbie:
### Short description
I think the old behaviour is harmless, as other records (like the apex NSset) will fill out the NSEC3 chain nicely, but it's better to be entirely correct.

Needs a test.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/ws-auth.cc
+++ b/pdns/ws-auth.cc
@@ -792,22 +792,6 @@ static void updateDomainSettingsFromDocument(UeberBackend& B, const DomainInfo& 
   }
 
   if (shouldRectify && !isPresigned) {
-    // Rectify
-    string api_rectify;
-    di.backend->getDomainMetadataOne(zonename, "API-RECTIFY", api_rectify);
-    if (api_rectify.empty()) {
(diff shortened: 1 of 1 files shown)


pdns/ws-auth.cc
```diff
@@ -826,6 +810,22 @@ static void updateDomainSettingsFromDocument(UeberBackend& B, const DomainInfo&
         }
       }
     }
+
+    // Rectify
```
zeha, line 814:
```suggestion
    // Rectify after ALL RRset modifications are done (incl. SOA)
```

mind04: The proposed comment is confusing. Rectify is also triggered by non rrset changes here.

Habbie: good point, in fact my test case was changing nsec3param!

zeha:
```suggestion
    // Rectify. This needs to happen after all (RRset and other) modification are done, incl. automatic RRs.
```

PR #12678, opened 2023-03-23
dnsdist: Add support for incoming DoH via nghttp2

rgacogne:
### Short description

This needs much more testing, do not put in production!

This PR adds support for incoming DNS over HTTPS using the `nghttp2` library instead of `h2o`. `h2o` is a very nice library but the development is now pretty much done for Fastly needs only, and there has been no release for several years and none planned. Distributions are stating to drop it, or have never packaged it to begin with requiring a special build step for our packages. It's sad that the DNS ecosystem is losing diversity, but it's time to move on.
The new code should be compatible with existing configurations without any changes, with the caveat that we no longer support HTTP/1, which was never intended for DoH anyway.
The expected benefits are:
- up-to-date library, properly packaged
- incoming proxy protocol support
- `GnuTLS` should now be supported for incoming DoH along `OpenSSL`
- hardware-accelerated TLS should now work for incoming DoH, including Linux's `kTLS` and Intel's QAT

<strike>TODO:
- more tests, especially memory leaks, load testing, GnuTLS
- write documentation
- ensure that we can still build without `nghttp2`, with or without `h2o`
- ensure that we can build without `h2o` but with `nghttp2` (probably broken right now)
- squash the commits on this branch :)</strike>

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/pdns/dnsdistdist/dnsdist-nghttp2-in.cc
+++ b/pdns/dnsdistdist/dnsdist-nghttp2-in.cc
@@ -387,32 +387,32 @@ ssize_t IncomingHTTP2Connection::send_callback(nghttp2_session* session, const u
   return length;
 }
 
-static const std::unordered_map<std::string, std::string> s_constants{
-  {"200-value", "200"},
-  {"method-name", ":method"},
-  {"method-value", "POST"},
-  {"scheme-name", ":scheme"},
-  {"scheme-value", "https"},
-  {"authority-name", ":authority"},
-  {"x-forwarded-for-name", "x-forwarded-for"},
-  {"path-name", ":path"},
-  {"content-length-name", "content-length"},
-  {"status-name", ":status"},
-  {"location-name", "location"},
-  {"accept-name", "accept"},
-  {"accept-value", "application/dns-message"},
-  {"cache-control-name", "cache-control"},
-  {"content-type-name", "content-type"},
-  {"content-type-value", "application/dns-message"},
-  {"user-agent-name", "user-agent"},
-  {"user-agent-value", "nghttp2-" NGHTTP2_VERSION "/dnsdist"},
-  {"x-forwarded-port-name", "x-forwarded-port"},
-  {"x-forwarded-proto-name", "x-forwarded-proto"},
-  {"x-forwarded-proto-value-dns-over-udp", "dns-over-udp"},
-  {"x-forwarded-proto-value-dns-over-tcp", "dns-over-tcp"},
-  {"x-forwarded-proto-value-dns-over-tls", "dns-over-tls"},
-  {"x-forwarded-proto-value-dns-over-http", "dns-over-http"},
-  {"x-forwarded-proto-value-dns-over-https", "dns-over-https"},
+static const std::array<const std::string, static_cast<size_t>(NGHTTP2Headers::HeaderConstantIndexes::COUNT)> s_headerConstants{
+  "200",
+  ":method",
+  "POST",
+  ":scheme",
+  "https",
+  ":authority",
+  "x-forwarded-for",
+  ":path",
+  "content-length",
+  ":status",
+  "location",
+  "accept",
+  "application/dns-message",
+  "cache-control",
+  "content-type",
+  "application/dns-message",
+  "user-agent",
+  "nghttp2-" NGHTTP2_VERSION "/dnsdist",
+  "x-forwarded-port",
+  "x-forwarded-proto",
+  "dns-over-udp",
+  "dns-over-tcp",
+  "dns-over-tls",
+  "dns-over-http",
+  "dns-over-https"
 };
 
 static const std::string s_authorityHeaderName(":authority");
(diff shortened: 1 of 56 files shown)


rgacogne:
I can't wrap my head around the formatting suggestion, what the heck am I missing?

omoerbeek:
> I can't wrap my head around the formatting suggestion, what the heck am I missing?

Either a bug in `clang-format` or a very obscure error in the code, but I do not see it, and emacs agrees.

rgacogne:

zeha:
IIRC there was some talk about dropping h2o in Debian trixie. Would be great to have this sooner than later :-)

rgacogne:
> IIRC there was some talk about dropping h2o in Debian trixie. Would be great to have this sooner than later :-)

I'm doing my best ;-) I expect this PR to be ready for review in the next couple weeks, and I would really like to have an 19.0 alpha including this work for the very beginning of September.

rgacogne:
Rebased on master, splitting the first commit in several smaller ones. There is still a big commit adding the actual code to handle incoming HTTP/2 queries with nghttp2, but it should be a bit easier to review now.

pdns/dnsdistdist/docs/guides/dns-over-https.rst
```diff
@@ -54,6 +54,16 @@ To let dnsdist listen for DoH queries over HTTP on localhost at port 8053 add on
...
   addDOHLocal("127.0.0.1:8053", nil, nil, "/", { reusePort=true })
 
+HTTP/1 support
+^^^^^^^^^^^^^^
+
+dnsdist initially relied on the ``h2o`` library to support incoming DNS over HTTPS. Since 1.9.0, ``h2o`` has been deprecated and ``nghttp2`` is the
+preferred library for incoming DoH support, because ``h2o`` has unfortunately really never been maintained in a way that is suitable for use as a library
+(see https://github.com/h2o/h2o/issues/3230). While we took great care to make the migration as painless as possible, ``h2o`` supported HTTP/1 while ``nghttp2``
+does not. This is not an issue for actual DNS over HTTPS clients that support HTTP/2, but might be one in setups running dnsdist behind a reverse-proxy that
+does not support HTTP/1, like nginx. We do not plan on implementing HTTP/1, and recommend using HTTP/2 between the reverse-proxy and dnsdist for performance reasons.
```
franklouwers, line 64: Shouldn't this be HTTP/2? If nginx would not support HTTP/1, the comms between nginx and dnsdist would be HTTP/2 (or other), so what would the issue be?

rgacogne: Good catch, fixed, thanks!

pdns/dnsdistdist/doh.cc
```diff
@@ -402,6 +452,12 @@ static void handleResponse(DOHFrontend& df, st_h2o_req_t* req, uint16_t statusCo
   }
 }
 
+static std::unique_ptr<DOHUnit> getDUFromIDS(InternalQueryState& ids)
```
chbruyand, line 465: Why can't we just move the `unique_ptr` ?

chbruyand (approved):
Thanks for splitting your PR as much as possible into individual understandable commits.
Nothing caught my eyes although as the PR is quiet big, another pair of eyes would be nice on it.
LGTM!

pdns/dnsdistdist/dnsdist-nghttp2-in.cc
```diff
@@ -0,0 +1,1214 @@
...
+        break;
+      }
+
+      if (!responseBody.empty()) {
+        NGHTTP2Headers::addDynamicHeader(headers, "content-type-name", "text/plain; charset=utf-8");
+      }
+      else {
+        static const std::string invalid{"invalid DNS query"};
+        static const std::string notAllowed{"dns query not allowed"};
+        static const std::string noDownstream{"no downstream server available"};
```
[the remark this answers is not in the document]
rgacogne: I would agree but I'm keeping the same messages than we have today in the h2o implementation, in the hope of easing the transition.

rgacogne:
I pushed a commit applying the suggestions from Charles-Henri's review, and rebased on top of master to fix a conflict.

pdns/dnsdistdist/doh.cc
```diff
@@ -2,6 +2,7 @@
 #include "doh.hh"
 
 #ifdef HAVE_DNS_OVER_HTTPS
+#ifdef HAVE_LIBH2OEVLOOP
```
fredmorcos, line 5: This entire file is now unused when libh2o is being used. Perhaps gating it in the build system instead of using the preprocessor would be a better option?

fredmorcos (approved):

pdns/tcpiohandler.cc
```diff
@@ -1817,46 +1817,57 @@ bool setupDoTProtocolNegotiation(std::shared_ptr<TLSCtx>& ctx)
...
+  if (!newCtx) {
 #ifdef HAVE_LIBSSL
-  newCtx = std::make_shared<OpenSSLTLSIOCtx>(*this);
+    newCtx = std::make_shared<OpenSSLTLSIOCtx>(*this);
 #else /* HAVE_LIBSSL */
 #ifdef HAVE_GNUTLS
-  newCtx = std::make_shared<GnuTLSIOCtx>(*this);
+    newCtx = std::make_shared<GnuTLSIOCtx>(*this);
 #endif /* HAVE_GNUTLS */
 #endif /* HAVE_LIBSSL */
```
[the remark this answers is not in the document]
rgacogne: >  I don't remember seeing anything in the auth configure.ac and accompanying scripts that stops from enabling both libssl and gnuTLS. Also running dnsdist's configure script with --with-libssl and --with-gnutls shows that it's possible to enable both.

Correct, we do want to be able to enable both of them at compile at time, but also at runtime! This code is only used if the user did not explicitly select one in the configuration.

> I would then prefer the use of `#elif` over `#endif`+`#ifdef` and the nested `#else`+`#ifdef` since it makes it more clear which library we prioritize and the respective library calls are always in the same order (e.g. libssl-related code will always come before GnuTLS-related code).

It makes a lot of sense, yes. I need to check whether this code is actually expected to compile when neither libssl nor GnuTLS are available, because if it does the final `#else` should go, but the rest looks good at first glance. Thanks!

pdns/dnsdistdist/doh.cc
```diff
@@ -1661,55 +1618,117 @@ void dohThread(ClientState* cs)
...
+    }
+    catch (const std::runtime_error& e) {
+      throw std::runtime_error("Error setting up TLS context for DoH listener on '" + d_tlsContext.d_addr.toStringWithPort() + "': " + e.what());
+    }
+  }
 }
 
-#endif /* HAVE_DNS_OVER_HTTPS */
+#endif /* HAVE_LIBH2OEVLOOP */
+#endif /* HAVE_LIBH2OEVLOOP */
```
rgacogne, line 1734:
```suggestion
#endif /* HAVE_DNS_OVER_HTTPS */
```

pdns/dnsdistdist/doh.cc
rgacogne: Makes sense to me, I'll have a look!

rgacogne: Done!

pdns/tcpiohandler.cc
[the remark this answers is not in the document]
rgacogne: Done now whenever possible.

PR #12688, opened 2023-03-27
rec: re-establish "recursion depth is always increasing" invariant

omoerbeek:
Now that we have getQMFallbackMode(), we can go back to always increase depth and never decrease it and adapt the upper bound check if needed.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/recursordist/syncres.cc
+++ b/pdns/recursordist/syncres.cc
@@ -1785,6 +1785,18 @@ int SyncRes::doResolve(const DNSName& qname, const QType qtype, vector<DNSRecord
   return RCode::ServFail;
 }
 
+unsigned int SyncRes::getAdjustedRecursionBound() const
+{
+  auto bound = s_maxdepth; // 40 is default value of s_maxdepth
+  if (getQMFallbackMode()) {
+    // We might have hit a depth level check, but we still want to allow some recursion levels in the fallback
+    // no-qname-minimization case. This has the effect that a qname minimization fallback case might reach 150% of
+    // maxdepth, taking care to not repeatedly increase the bound.
+    bound += s_maxdepth / 2;
+  }
+  return bound;
+}
+
 /*! This function will check the cache and go out to the internet if the answer is not in cache
  *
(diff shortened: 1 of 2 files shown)


rgacogne (approved):
The change looks good to me, it certainly seems less brittle.
I just noticed that in two places we are actually checking `depth` against hard-coded values that are not related to `s_maxdepth`, so the behaviour after this PR will be slightly different in these cases:
- `SyncRes::doCNAMECacheCheck`: `if ((depth > 9 && d_outqueries > 10 && d_throttledqueries > 5) || depth > 15) {`
- `SyncRes::handleNewTarget`: `if (depth > 10) {`

Technically these happen when we are following a `CNAME`, so QM might matter less.

omoerbeek:
I never liked these conditions with all the arbitrary numbers.... I'll take a look to see if these can be fixed in this PR.

omoerbeek:
> I never liked these conditions with all the arbitrary numbers.... I'll take a look to see if these can be fixed in this PR.

This is a bit of a can of worms, especially the check in `doCNAMECacheCheck`, as it is called in many cases after the initial depth check in `doResolveNoQNameMinimization()`. That means that the hardcoded bounds are always in effect if it they lower than a `s_maxdepth` (derived) one.

So I'm leaving that for another time.

omoerbeek:
I really do wonder if the depth check in `doCNAMECacheCeck()` is in the right place.

rgacogne:
> I really do wonder if the depth check in `doCNAMECacheCeck()` is in the right place.

I had the same thought, and I think the answer is likely 'no'. I suggest we revisit these checks soon in a different issue/PR, if that's fine by you?

omoerbeek:
> > I really do wonder if the depth check in `doCNAMECacheCeck()` is in the right place.
> 
> I had the same thought, and I think the answer is likely 'no'. I suggest we revisit these checks soon in a different issue/PR, if that's fine by you?

Yep, that's the plan.

PR #12721, opened 2023-04-05
auth: calm down the communicator loop

mind04:
### Short description
Kicking the communicator loop for notifies is counter productive since the delay caused by the extra slaveRefresh() and masterUpdateCheck() calls is fare bigger than the time we gain on average. Since #7790 doNotifications() is already called about once every second.

The increased freshness check frequentie is also resulting in a significant load increase. Especially in larger setups when catalog zones are in use.

### Checklist
I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


pdns/auth-main.cc
```diff
@@ -1241,6 +1241,8 @@ int main(int argc, char** argv)
       ::arg().set("allow-unsigned-autoprimary") = "yes";
     if (!::arg().isEmpty("domain-metadata-cache-ttl"))
       ::arg().set("zone-metadata-cache-ttl") = ::arg()["domain-metadata-cache-ttl"];
+    if (!::arg().isEmpty("slave-cycle-interval"))
```
Habbie, line 1244: perhaps default `slave-cycle-interval` to empty in that case, like we did for `domain-metadata-cache-ttl`?

Habbie (changes requested):
one comment, otherwise good

Habbie (approved):

PR #12726, opened 2023-04-05
dnsdist: Properly handle reconnection failure for backend UDP sockets

### LINKED ISSUE #12711: dnsdist: high CPU usage (150-200%) when gateway/routes disappear since 1.8.0
- Program: dnsdist
 - Issue type: Bug report

### Short description
Since dnsdist-1.8.0 there seems to be a change in behaviour in case of a router disconnect/IPv6 prefix change. I run dnsdist as proxy/cache for external DNS servers (at my ISP and others), with a mix of IPv4 and IPv6 servers. Unlike the v4 servers, the v6 servers are broken when my DSL line disconnects & I receive a new prefix (typically in the middle of the night), so an elaborate monitoring script detects this and properly restarts v6-related services when the new prefix is available. This has been working fine for a while, and still does.
However, I noticed that 1.8.0 seems to **really** dislike the change in routing, and at least two threads (probably UDP request sender & availability checker?) start spinning like crazy - see the attached image. This definitely wasn't the case with 1.7.x.

### Environment
 - Operating system: Linux, kernel 6.2.x
 - Software version: 1.8.0
 - Software source: self-compiled package

### Steps to reproduce
1. run dnsdist
2. remove route to upstream server / power-cycle router

### Expected behaviour
Changed IP addresses shouldn't make threads go crazy.

### Actual behaviour
dnsdist-cpu
As you can see the DSL line went down at ~3:26 and dnsdist went into overdrive. DSL reconnection takes a while and my IPv6-prefix-detector only runs every 5 minutes, but eventually it restarts dnsdist and things are back to normal.

### Other information
I understand that detecting topology changes was discussed in the past (e.g. via netlink) and that it would be quite a bit of work; however threads burning CPU seems like an avoidable problem.

rgacogne:
### Short description

We try to reconnect our UDP sockets toward backends on some kind of network errors that indicate a topology change, but we need to be careful to handle the case where we actually fail to reconnect, as we end up with no remaining sockets to use.
This commit properly deals with this case by pausing the thread handling UDP responses from the backend, instead of having it enter a busy loop, and by attempting to reconnect if we get a `bad file number` error when trying to send a UDP datagram to the backend.

Hopefully fixes #12711.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code


hhoffstaette:
Cool! I'll give it a try and will report back - give it a week or so.

hhoffstaette:
I triggered 2 reconnects and did not observe any CPU hogging. Cool!
:+1:

zeha:
This needs a review, right?

rgacogne:
> This needs a review, right?

Yes

omoerbeek:
I tried to reproduce the busy looping on master by manipulating the v6 default route, but failed to do so on OpenBSD, macOS and debian. I wonder what extra condition is needed to trigger the busy loop.

zeha:
The last busy loop I saw had very little useful log output, but maybe someone can extrapolate:

```
Apr 30 06:06:15 las-sh01 dnsdist[547]: Error connecting to new server with address 45.90.28.62:53: connecting socket to 45.90.28.62:53: Network is unreachable
```

I'm not sure what happened at that time, but it looks like an automated update/restart of `systemd-networkd`:
```
Apr 30 06:06:14 las-sh01 systemd[1]: Stopping Network Service...
Apr 30 06:06:14 las-sh01 systemd-networkd[479]: inetupc: DHCPv6 lease lost
Apr 30 06:06:14 las-sh01 systemd-networkd[479]: ppp0: DHCPv6 lease lost
Apr 30 06:06:14 las-sh01 systemd[1]: systemd-networkd.service: Succeeded.
Apr 30 06:06:14 las-sh01 systemd[1]: Stopped Network Service.
Apr 30 06:06:14 las-sh01 systemd[1]: systemd-networkd.service: Consumed 2min 57.626s CPU time.
Apr 30 06:06:14 las-sh01 systemd[1]: Starting Network Service...
Apr 30 06:06:14 las-sh01 systemd-networkd[620845]: /etc/systemd/network/inetfonira.network:23: Unknown key name 'SendRelease' in section 'DHCP', ignoring.
Apr 30 06:06:14 las-sh01 systemd-networkd[620845]: /etc/systemd/network/inetupc.network:16: Unknown key name 'SendRelease' in section 'DHCP', ignoring.
Apr 30 06:06:14 las-sh01 systemd-networkd[620845]: /etc/systemd/network/wired.network: vlan20 NetDev could not be found, ignoring assignment.
Apr 30 06:06:14 las-sh01 systemd-networkd[620845]: wg0: netdev ready
Apr 30 06:06:14 las-sh01 systemd-networkd[620845]: ppp0: Gained IPv6LL
Apr 30 06:06:14 las-sh01 systemd-networkd[620845]: tailscale0: Gained IPv6LL
Apr 30 06:06:14 las-sh01 systemd-networkd[620845]: iot: Gained IPv6LL
Apr 30 06:06:14 las-sh01 systemd-networkd[620845]: inetupc: Gained IPv6LL
Apr 30 06:06:14 las-sh01 systemd-networkd[620845]: enp6s18: Gained IPv6LL
Apr 30 06:06:14 las-sh01 systemd-networkd[620845]: Enumeration completed
Apr 30 06:06:14 las-sh01 systemd[1]: Started Network Service.
Apr 30 06:06:14 las-sh01 systemd-networkd[620845]: wg0: netdev exists, using existing without changing its parameters
Apr 30 06:06:14 las-sh01 systemd-networkd[620845]: ppp0: found matching network '/etc/systemd/network/ppp0.network', based on potentially unpredictable ifname
Apr 30 06:06:15 las-sh01 systemd-networkd[620845]: inetupc: DHCPv4 address 84.114.19.241/24 via 84.114.19.1
Apr 30 06:06:18 las-sh01 systemd-networkd[620845]: inetfonira: DHCPv4 address 100.64.216.24/18 via 100.64.192.1
```

I *guess* this means there was no IPv4 and also no IPv6 default route for about a second or two.


Startup log from dnsdist 10 days earlier:
```
Apr 20 06:01:41 las-sh01 dnsdist[547]: dnsdist 1.9.0-alpha0.68.master.gda45a962d comes with ABSOLUTELY NO WARRANTY. This is free software, and you are welcome to redistribute it according to the terms of the GPL version 2
Apr 20 06:01:41 las-sh01 dnsdist[547]: Passing a plain-text password via the 'password' parameter to 'setWebserverConfig()' is not advised, please consider generating a hashed one using 'hashPassword()' instead.
Apr 20 06:01:41 las-sh01 dnsdist[547]: Passing a plain-text API key via the 'apiKey' parameter to 'setWebserverConfig()' is not advised, please consider generating a hashed one using 'hashPassword()' instead.
Apr 20 06:01:41 las-sh01 dnsdist[547]: Added downstream server 127.0.0.1:5300
Apr 20 06:01:41 las-sh01 dnsdist[547]: Error connecting to new server with address [2a07:a8c0::3f:2ba1]:53: connecting socket to [2a07:a8c0::3f:2ba1]:53: Network is unreachable
Apr 20 06:01:41 las-sh01 dnsdist[547]: Added downstream server [2a07:a8c0::3f:2ba1]:53
Apr 20 06:01:41 las-sh01 dnsdist[547]: Error connecting to new server with address [2a07:a8c1::3f:2ba1]:53: connecting socket to [2a07:a8c1::3f:2ba1]:53: Network is unreachable
Apr 20 06:01:41 las-sh01 dnsdist[547]: Added downstream server [2a07:a8c1::3f:2ba1]:53
Apr 20 06:01:41 las-sh01 dnsdist[547]: Added downstream server 45.90.28.62:53
Apr 20 06:01:41 las-sh01 dnsdist[547]: Added downstream server 45.90.30.62:53
Apr 20 06:01:41 las-sh01 dnsdist[547]: Listening on 172.16.172.4:53
Apr 20 06:01:41 las-sh01 dnsdist[547]: Listening on 172.16.172.1:53
Apr 20 06:01:41 las-sh01 dnsdist[547]: Listening on 127.0.0.1:53
Apr 20 06:01:41 las-sh01 dnsdist[547]: Listening on 172.16.20.1:53
Apr 20 06:01:41 las-sh01 dnsdist[547]: Listening on [fe80::20d:b9ff:fe41:41b0]:53
Apr 20 06:01:41 las-sh01 dnsdist[547]: Listening on [2a02:1748:fad4:5021::1]:53
Apr 20 06:01:41 las-sh01 dnsdist[547]: Listening on [2a02:1748:fad4:5022::1]:53
Apr 20 06:01:41 las-sh01 dnsdist[547]: ACL allowing queries from: 10.0.0.0/8, 100.64.0.0/10, 127.0.0.0/8, 169.254.0.0/16, 172.16.0.0/12, 192.168.0.0/16, ::1/128, 2a02:1748:fad4:5021::/64, 2a02:1748:fad4:5022::/64, fc00::/7, fe80::/10
Apr 20 06:01:41 las-sh01 dnsdist[547]: Console ACL allowing connections from: 127.0.0.0/8, ::1/128
Apr 20 06:01:41 las-sh01 dnsdist[547]: Marking downstream [2a07:a8c0::3f:2ba1]:53 as 'down'
Apr 20 06:01:41 las-sh01 dnsdist[547]: Marking downstream [2a07:a8c1::3f:2ba1]:53 as 'down'
Apr 20 06:01:41 las-sh01 dnsdist[547]: Webserver launched on 172.16.172.1:8083
Apr 20 06:01:41 las-sh01 dnsdist[547]: Accepting control connections on 127.0.0.1:5199
Apr 20 06:01:41 las-sh01 dnsdist[547]: Marking downstream 127.0.0.1:5300 as 'up'
Apr 20 06:01:42 las-sh01 dnsdist[547]: Marking downstream 45.90.28.62:53 as 'down'
Apr 20 06:01:42 las-sh01 dnsdist[547]: Marking downstream 45.90.30.62:53 as 'down'
Apr 20 06:27:13 las-sh01 dnsdist[547]: Marking downstream 45.90.30.62:53 as 'up'
Apr 20 06:27:13 las-sh01 dnsdist[547]: Marking downstream 45.90.28.62:53 as 'up'
```

omoerbeek:
Playing some more on debian I managed to reproduce on master:

1. Configure a v6 backend with no healthcheck  `newServer({address='2620:fe::9', source='enp3s0', healthCheckMode='up'})`
2. Start dnsdist
3. Check things work by doing a query
4. Drop default v6 route: `ip route del ::/0 `
5. Do a query and hear fans spin up
6. Adding the default route back (`ip -6 route add default via <your gw's local link addres> dev enp3s0`) does not improve things, dnsdist keeps looping

With the PR applied things are fine: no looping and whn I add the default route back thinsg work again.  Two observations:

1. If I start dndist with no default v6 route available and add it later dnsdist keeps failing, it does not reach a working state.
2. After dropping the default route after starting dnsdist with it present I see a message each time an attempt is done to connect to the backend:
`Error connecting to new server with address [2620:fe::9]:53: connecting socket to [2620:fe::9]:53: Network is unreachable` that may be too much.

I'll review the code next.

omoerbeek (commented):
Code looks good, but please read my comments I added after testing.

rgacogne:
>     1. If I start dndist with no default v6 route available and add it later dnsdist keeps failing, it does not reach a working state.

That's unexpected and I would like to fix that!

>     2. After dropping the default route after starting dnsdist with it present I see a message each time an attempt is done to connect to the backend:
>        `Error connecting to new server with address [2620:fe::9]:53: connecting socket to [2620:fe::9]:53: Network is unreachable` that may lead to too much log spam.

Oh, yeah, that's not nice. Perhaps we should log at verbose level only when called from `udpClientSendRequestToBackend` (so not a human action)?

rgacogne:
I pushed two changes:
- failed re-connection attempts are only logged at info level once, and then at verbose level only
- check that the UDP responder thread is running after a successful re-connection, and start it otherwise

omoerbeek:
This works for me.

I do see a data race when running with TSAN after droppingthe v6 default route and doing a request:
```
WARNING: ThreadSanitizer: data race (pid=83799)
  Write of size 8 at 0x7ba000000090 by thread T16 (mutexes: write M196):
    #0 close ../../../../src/libsanitizer/tsan/tsan_interceptors_posix.cpp:1675 (libtsan.so.0+0x335a8)
    #1 DownstreamState::reconnect(bool) /home/otto/pdns/pdns/dnsdistdist/dnsdist-backend.cc:60 (dnsdist+0x241394)
    #2 udpClientSendRequestToBackend(std::shared_ptr<DownstreamState> const&, int, std::vector<unsigned char, noinit_adaptor<std::allocator<unsigned char> > > const&, bool) /home/otto/pdns/pdns/dnsdistdist/dnsdist.cc:1137 (dnsdist+0x88611e)
...
  Previous read of size 8 at 0x7ba000000090 by thread T15:
    #0 recv ../../../../src/libsanitizer/sanitizer_common/sanitizer_common_interceptors.inc:6394 (libtsan.so.0+0x4d656)
    #1 recv /usr/include/x86_64-linux-gnu/bits/socket2.h:44 (dnsdist+0x888bf8)
    #2 responderThread(std::shared_ptr<DownstreamState>) /home/otto/pdns/pdns/dnsdistdist/dnsdist.cc:750 (dnsdist+0x888bf8)
...
  Location is file descriptor 9 created by main thread at:
    #0 connect ../../../../src/libsanitizer/tsan/tsan_interceptors_posix.cpp:1650 (libtsan.so.0+0x41bf2)
    #1 SConnect(int, ComboAddress const&) /home/otto/pdns/pdns/dnsdistdist/iputils.cc:58 (dnsdist+0x8f8a9b)
```
Plus a bunch of
```
WARNING: ThreadSanitizer: signal-unsafe call inside of a signal (pid=83799)
    #0 calloc ../../../../src/libsanitizer/tsan/tsan_interceptors_posix.cpp:668 (libtsan.so.0+0x2ffe8)
    #1 <null> <null> (libedit.so.2+0x140df)
    #2 EpollFDMultiplexer::getAvailableFDs(std::vector<int, std::allocator<int> >&, int) /home/otto/pdns/pdns/dnsdistdist/epollmplexer.cc:156 (dnsdist+0x9431ba)
```

rgacogne:
Not much we can do about:
```
WARNING: ThreadSanitizer: signal-unsafe call inside of a signal (pid=83799)
    #0 calloc ../../../../src/libsanitizer/tsan/tsan_interceptors_posix.cpp:668 (libtsan.so.0+0x2ffe8)
    #1 <null> <null> (libedit.so.2+0x140df)
    #2 EpollFDMultiplexer::getAvailableFDs(std::vector<int, std::allocator<int> >&, int) /home/otto/pdns/pdns/dnsdistdist/epollmplexer.cc:156 (dnsdist+0x9431ba)
```
For the other one:
```
WARNING: ThreadSanitizer: data race (pid=83799)
  Write of size 8 at 0x7ba000000090 by thread T16 (mutexes: write M196):
    #0 close ../../../../src/libsanitizer/tsan/tsan_interceptors_posix.cpp:1675 (libtsan.so.0+0x335a8)
    #1 DownstreamState::reconnect(bool) /home/otto/pdns/pdns/dnsdistdist/dnsdist-backend.cc:60 (dnsdist+0x241394)
    #2 udpClientSendRequestToBackend(std::shared_ptr<DownstreamState> const&, int, std::vector<unsigned char, noinit_adaptor<std::allocator<unsigned char> > > const&, bool) /home/otto/pdns/pdns/dnsdistdist/dnsdist.cc:1137 (dnsdist+0x88611e)
...
  Previous read of size 8 at 0x7ba000000090 by thread T15:
    #0 recv ../../../../src/libsanitizer/sanitizer_common/sanitizer_common_interceptors.inc:6394 (libtsan.so.0+0x4d656)
    #1 recv /usr/include/x86_64-linux-gnu/bits/socket2.h:44 (dnsdist+0x888bf8)
    #2 responderThread(std::shared_ptr<DownstreamState>) /home/otto/pdns/pdns/dnsdistdist/dnsdist.cc:750 (dnsdist+0x888bf8)
...
  Location is file descriptor 9 created by main thread at:
    #0 connect ../../../../src/libsanitizer/tsan/tsan_interceptors_posix.cpp:1650 (libtsan.so.0+0x41bf2)
    #1 SConnect(int, ComboAddress const&) /home/otto/pdns/pdns/dnsdistdist/iputils.cc:58 (dnsdist+0x8f8a9b)
```
We do protect the re-connection attempt with a mutex, so only thread will write to the map, but there is indeed a slight race where the responder thread could still be looking (trying to receive) at the old file descriptors during the re-connection before noticing what is happening. Taking the mutex here would not be great for performance, and re-connection should pretty much never happen, so I would be inclined to just leave it like that.

rgacogne:
Build is failing because of:
```
Warning: function 'reconnect' has cognitive complexity of 45 (threshold 25) (readability-function-cognitive-complexity - Level=Warning)
Warning: /home/runner/work/pdns/pdns/pdns/dnsdistdist/dnsdist-backend.cc:40: function 'reconnect' has cognitive complexity of 45 (threshold 25) (readability-function-cognitive-complexity)
Warning: do not access members of unions; use (boost::)variant instead (cppcoreguidelines-pro-type-union-access - Level=Warning)
Warning: /home/runner/work/pdns/pdns/pdns/dnsdistdist/dnsdist-backend.cc:93: do not access members of unions; use (boost::)variant instead (cppcoreguidelines-pro-type-union-access)
```

omoerbeek:
> so I would be inclined to just leave it like that

Agreed. A entry in the suprression file would be nice though.

rgacogne:
@fredmorcos How do you feel about raising the cognitive complexity threshold? I'm afraid the current one is really too low, as `DownstreamState::reconnect()` is not a complicated function at all and it is already at 45 which is way over the 25 limit.

fredmorcos:
> @fredmorcos How do you feel about raising the cognitive complexity threshold? I'm afraid the current one is really too low, as `DownstreamState::reconnect()` is not a complicated function at all and it is already at 45 which is way over the 25 limit.

I'm okay with raising it. As I said I believe the current value is just a clang tidy default. We can change it but we shouldn't go overboard. In other words we shouldn't use this value to virtually disable the lint. e.g. serviceMain really needs to be broken down into multiple functions.

rgacogne:
> > @fredmorcos How do you feel about raising the cognitive complexity threshold? I'm afraid the current one is really too low, as `DownstreamState::reconnect()` is not a complicated function at all and it is already at 45 which is way over the 25 limit.
> 
> I'm okay with raising it. As I said I believe the current value is just a clang tidy default. We can change it but we shouldn't go overboard. In other words we shouldn't use this value to virtually disable the lint. e.g. serviceMain really needs to be broken down into multiple functions.

Agreed! Do you have any clue as to what would be a sane default? Perhaps we could raise to 100 for now, and see what happens?

Unrelated question, I'm wondering if our clang-tidy is looking at all the files changed in a pull request or only the ones changed in the last commit? We use `git diff -U0 HEAD^..HEAD` to look at the changes, right?

fredmorcos:
> ```
> Warning: do not access members of unions; use (boost::)variant instead (cppcoreguidelines-pro-type-union-access - Level=Warning)
> Warning: /home/runner/work/pdns/pdns/pdns/dnsdistdist/dnsdist-backend.cc:93: do not access members of unions; use (boost::)variant instead (cppcoreguidelines-pro-type-union-access)
> ```
> 
> Well, unless I'm mistaken this completely prevents the use of unions, which makes our `ComboAddress` object unusable.

Yep. It's something I've come across several times and I don't know what to do about it. We need to discuss this. Here is what I'm thinking about, from best to worst (regardless of this PR):

- Move unions to std::variant. There are probably performance implications. I don't know if C++ compilers are able to optimize std::variant back into (tagged) unions and I doubt it but I'm not an expert. So we pretty much need to port ComboAddress and see what perf issues we end up with.

- If it is possible, disable the lint by marking ComboAddress itself instead of marking the usages of it. I don't know how or if it's possible because the warnings show up at usage sites.

- Disable the lint altogether in the config file.

- Disable the lint on each usage site.

Of course any of those disabling a lint must come with a GitHub issue.

fredmorcos:
> Agreed! Do you have any clue as to what would be a sane default? Perhaps we could raise to 100 for now, and see what happens?

I don't really know. Let's start with 50 or some value that allows `DownstreamState::reconnect()` to pass, then we will see how it goes.

> Unrelated question, I'm wondering if our clang-tidy is looking at all the files changed in a pull request or only the ones changed in the last commit? We use `git diff -U0 HEAD^..HEAD` to look at the changes, right?

Correct. I think this would then be the merge commit that Github Actions creates, no?

zeha:
> Agreed! Do you have any clue as to what would be a sane default? Perhaps we could raise to 100 for now, and see what happens?

I think this limit will be a real problem: what should be done about pre-existing code that crosses this limit?
My specific example is doAXFR, which has 450 or so. But I'm not going to do a major refactor in a contributing MR that touches this function...

rgacogne:
> I think this limit will be a real problem: what should be done about pre-existing code that crosses this limit? My specific example is doAXFR, which has 450 or so. But I'm not going to do a major refactor in a contributing MR that touches this function...

This was (briefly) discussed in https://github.com/PowerDNS/pdns/pull/12765: we can mark code that we do not want to refactor with a `NOLINT` comment, while keeping a list of the functions that we will eventually need to refactor, likely in a ticket.
I agree with you that this might not work, but I think it's worth a try. We can also decide to manually ignore these warnings when needed.

rgacogne:
> I don't really know. Let's start with 50 or some value that allows `DownstreamState::reconnect()` to pass, then we will see how it goes.

Alright, I'll open a PR raising that value.

> > Unrelated question, I'm wondering if our clang-tidy is looking at all the files changed in a pull request or only the ones changed in the last commit? We use `git diff -U0 HEAD^..HEAD` to look at the changes, right?
> 
> Correct. I think this would then be the merge commit that Github Actions creates, no?

I actually have no idea. I know that GH checkout merge against the current master before running, but I'm a bit lost to what the git hierarchy looks like after that. If you think this is OK, that works for me.

rgacogne:
> > Well, unless I'm mistaken this completely prevents the use of unions, which makes our `ComboAddress` object unusable.
> 
> Yep. It's something I've come across several times and I don't know what to do about it. We need to discuss this. Here is what I'm thinking about, from best to worst (regardless of this PR):
> 
>     * Move unions to std::variant. There are probably performance implications. I don't know if C++ compilers are able to optimize std::variant back into (tagged) unions and I doubt it but I'm not an expert. So we pretty much need to port ComboAddress and see what perf issues we end up with.

In most cases `std::variant` might be good enough for us as far as performance goes. The problem here is that this particular `ComboAddress` is used everywhere in the code base and has been exposing its internal fields to everyone, so the refactoring is going to be very painful, and conflict with a lot of PRs. So while I would welcome that change, I'm not convinced this is the right solution at the moment.

>     * If it is possible, disable the lint by marking ComboAddress itself instead of marking the usages of it. I don't know how or if it's possible because the warnings show up at usage sites.

If possible, that might be our best option for now.

>     * Disable the lint altogether in the config file.

I did a quick check and we use very few unions, but it would be nice to be able to enforce that. Let's keep that option in mind if the previous one does not work.

>     * Disable the lint on each usage site.

I'm afraid there is going to be a huge list of these, but I might be wrong.

I'll see if I can make option "disable the lint by marking ComboAddress itself" work.

fredmorcos:
> This was (briefly) discussed in #12765: we can mark code that we do not want to refactor with a `NOLINT` comment, while keeping a list of the functions that we will eventually need to refactor, likely in a ticket.

I would like to point out that this is a last resort. We really should be cleaning up functions that we touch if/when possible. Otherwise none of this matters. Something like `doAXFR` is the technical debt we accumulated that we now have to (or should) pay.

rgacogne:
> I would like to point out that this is a last resort. We really should be cleaning up functions that we touch if/when possible. Otherwise none of this matters. Something like `doAXFR` is the technical debt we accumulated that we now have to (or should) pay.

I do agree, but we cannot ask contributors to clean up our mess. If someone is kind enough to contribute a fix / implement a feature and are told they need to clean up something like `doAXFR`, we are doing it wrong.

fredmorcos:
> I do agree, but we cannot ask contributors to clean up our mess. If someone is kind enough to contribute a fix / implement a feature and are told they need to clean up something like `doAXFR`, we are doing it wrong.

Agreed.

rgacogne:
> > > Unrelated question, I'm wondering if our clang-tidy is looking at all the files changed in a pull request or only the ones changed in the last commit? We use `git diff -U0 HEAD^..HEAD` to look at the changes, right?
> > 
> > 
> > Correct. I think this would then be the merge commit that Github Actions creates, no?
> 
> I actually have no idea. I know that GH checkout merge against the current master before running, but I'm a bit lost to what the git hierarchy looks like after that. If you think this is OK, that works for me.

I did a quick test by pushing a few commits to a branch and adding a `git diff -U0 HEAD^..HEAD` step and I only see the content of my last commit, so I don't think this is working as expected. @fredmorcos, would you be able to look into it?

rgacogne:
> > > > Unrelated question, I'm wondering if our clang-tidy is looking at all the files changed in a pull request or only the ones changed in the last commit? We use `git diff -U0 HEAD^..HEAD` to look at the changes, right?
> > > 
> > > 
> > > Correct. I think this would then be the merge commit that Github Actions creates, no?
> > 
> > 
> > I actually have no idea. I know that GH checkout merge against the current master before running, but I'm a bit lost to what the git hierarchy looks like after that. If you think this is OK, that works for me.
> 
> I did a quick test by pushing a few commits to a branch and adding a `git diff -U0 HEAD^..HEAD` step and I only see the content of my last commit, so I don't think this is working as expected. @fredmorcos, would you be able to look into it?

It turns out I was wrong, as Fred kindly explained :) `git diff -U0 HEAD^..HEAD`  works as expected on pull request because GH actions is doing a merge under the hood. It was not working on my `push` tests because the mechanism is different, but that's fine.

rgacogne:
It looks like my last commit adding a suppression for `ComboAddress::toString()` worked. We will likely need more of these, but at the very least we might not need to add one for every callers.

pdns/iputils.hh
```diff
@@ -265,6 +265,7 @@ union ComboAddress {
   {
     char host[1024];
     int retval = 0;
+    // NOLINTNEXTLINE(cppcoreguidelines-pro-type-union-access)
```
fredmorcos, line 268: Could you add a comment, and also an issue if you think this is something we will change/fix in the future.

I added this bit of documentation for when it comes to silencing lint warnings: https://github.com/PowerDNS/pdns/blob/master/CONTRIBUTING.md#clang-tidy-and-ci

fredmorcos (changes requested):

pdns/dnsdistdist/dnsdist-backend.cc
```diff
@@ -89,7 +89,9 @@ bool DownstreamState::reconnect()
       connected = true;
     }
     catch (const std::runtime_error& error) {
-      infolog("Error connecting to new server with address %s: %s", d_config.remote.toStringWithPort(), error.what());
+      if (initialAttempt || g_verbose) {
+        infolog("Error connecting to new server with address %s: %s", d_config.remote.toStringWithPort(), error.what());
```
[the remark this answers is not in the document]
fredmorcos: Ah, you need to rebase the branch.

pdns/iputils.hh
fredmorcos: Awesome, thank you!

rgacogne:
> It looks like my last commit adding a suppression for `ComboAddress::toString()` worked. We will likely need more of these, but at the very least we might not need to add one for every callers.

After a rebase it turns out it did not work at all:
> /home/runner/work/pdns/pdns/pdns/dnsdistdist/dnsdist-backend.cc:93: do not access members of unions; use (boost::)variant instead (cppcoreguidelines-pro-type-union-access)

We might have to disable that warning globally for now, unfortunately, as I don't think adding an annotation to every callers is going to work.

fredmorcos:
> We might have to disable that warning globally for now, unfortunately, as I don't think adding an annotation to every callers is going to work.

Yes, this was my suspicion.

rgacogne:
Done in https://github.com/PowerDNS/pdns/pull/12812

omoerbeek (approved):

omoerbeek:
Shouldn't this be marked for backporting to 1.8.x ?

rgacogne:
It should!

PR #12729, opened 2023-04-06
auth LMDB: handle duplicate domain existence consistently

Habbie:
### Short description
When two LightningStream writers add the same zone, they most likely will get different zone IDs, and LightningStream will then synchronise both copies to all receivers. With this PR, auth handles that situation consistently across a cluster.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/ext/lmdb-safe/lmdb-typed.hh
+++ b/ext/lmdb-safe/lmdb-typed.hh
@@ -310,6 +310,7 @@ public:
       // auto range = prefix_range<N>(key);
       LMDBIDvec ids;
 
+      // because we know we only want one item, pass onlyOldest=true to consistently get the same one out of a set of duplicates
       get_multi<N>(key, ids, true);
 
       if (ids.size() == 0) {
--- a/ext/lmdb-safe/lmdb-typed.hh
+++ b/ext/lmdb-safe/lmdb-typed.hh
@@ -310,7 +310,7 @@ public:
       // auto range = prefix_range<N>(key);
       LMDBIDvec ids;
 
-      get_multi<N>(key, ids);
+      get_multi<N>(key, ids, true);
 
       if (ids.size() == 0) {
         return 0;
@@ -646,7 +646,7 @@ public:
     };
 
     template<int N>
-    void get_multi(const typename std::tuple_element<N, tuple_t>::type::type& key, LMDBIDvec& ids)
+    void get_multi(const typename std::tuple_element<N, tuple_t>::type::type& key, LMDBIDvec& ids, bool onlyOldest=false)
     {
       // std::cerr<<"in get_multi"<<std::endl;
       typename Parent::cursor_t cursor = (*d_parent.d_txn)->getCursor(std::get<N>(d_parent.d_parent->d_tuple).d_idx);
@@ -657,6 +657,9 @@ public:
       out.d_mdbval = in.d_mdbval;
 
       int rc = cursor.get(out, id,  MDB_SET_RANGE);
+
+      uint64_t oldestts = UINT64_MAX;
+      uint32_t oldestid = 0;
 
       while (rc == 0) {
         auto sout = out.getNoStripHeader<std::string>(); // FIXME: this (and many others) could probably be string_view
@@ -670,7 +673,19 @@ public:
 
         if (sthiskey == keyString) {
           auto _id = getIDFromCombinedKey(out);
-          ids.push_back(_id.getNoStripHeader<uint32_t>());
+          uint64_t ts = LMDBLS::LSgetTimestamp(id.getNoStripHeader<string_view>());
+          uint32_t __id = _id.getNoStripHeader<uint32_t>();
+
+          if (ts < oldestts) {
+            oldestts = ts;
+            oldestid = __id;
+          }
+          ids.push_back(__id);
+        }
+
+        if (onlyOldest && ids.size() > 1) {
+          ids.clear();
+          ids.push_back(oldestid);
         }
 
         rc = cursor.get(out, id, MDB_NEXT);
--- a/ext/lmdb-safe/lmdb-typed.hh
+++ b/ext/lmdb-safe/lmdb-typed.hh
@@ -676,16 +676,17 @@ public:
           uint64_t ts = LMDBLS::LSgetTimestamp(id.getNoStripHeader<string_view>());
           uint32_t __id = _id.getNoStripHeader<uint32_t>();
 
-          if (ts < oldestts) {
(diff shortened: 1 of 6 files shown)


Habbie:
still need to handle cleanup (or at least make delete do the right thing), and check other methods that go over all domains (primary/secondary checks, catalog)

modules/lmdbbackend/lmdbbackend.cc
```diff
@@ -1665,7 +1669,27 @@ void LMDBBackend::getAllDomains(vector<DomainInfo>* domains, bool /* doSerial */
...
+    }
+    else {
+      zonemap[di.zone] = di;
+    }
+  }
+
+  for (auto zone : dups) {
+    DomainInfo di;
+
+    if (!(di.id = txn.get<0>(zone, di))) {
```
wojas, line 1683: Not sure if if I understand this correctly, but does this not just check if a reverse index entry is present, instead of checking if this is the active one?

Habbie: `.get` calls `.get_multi` passing `true` to the new `onlyOldest` argument

modules/lmdbbackend/lmdbbackend.cc
```diff
@@ -1651,22 +1659,69 @@ bool LMDBBackend::createDomain(const DNSName& domain, const DomainInfo::DomainKi
...
+
+    map<DNSName, DomainInfo> zonemap;
+    set<DNSName> dups;
+
+    for (auto iter = txn.begin(); iter != txn.end(); ++iter) {
+      DomainInfo di = *iter;
+      di.id = iter.getID();
+      di.backend = this;
+
+      if (zonemap.count(di.zone) == 1) {
```
omoerbeek, line 1675: Just insert and if the insert fails add it to dups?

modules/lmdbbackend/lmdbbackend.cc
```diff
@@ -1651,22 +1659,69 @@ bool LMDBBackend::createDomain(const DNSName& domain, const DomainInfo::DomainKi
...
+    }
+
+    for (auto [k, v] : zonemap) {
+      if (allow(v)) {
+        domains->push_back(v);
+      }
+    }
+  }
+  else {
+    auto txn = d_tdomains->getROTransaction();
```
omoerbeek, line 1701: The `getROTransaction` line can move outside the if/else block, it's common to both.

modules/lmdbbackend/lmdbbackend.cc
```diff
@@ -1651,22 +1659,69 @@ bool LMDBBackend::createDomain(const DNSName& domain, const DomainInfo::DomainKi
...
+    for (auto iter = txn.begin(); iter != txn.end(); ++iter) {
```
omoerbeek, line 1702: maybe range loop?

ext/lmdb-safe/lmdb-typed.hh
```diff
@@ -665,7 +673,19 @@ public:
...
+          uint32_t __id = _id.getNoStripHeader<uint32_t>();
+
+          if (ts < oldestts) {
+            oldestts = ts;
+            oldestid = __id;
+          }
+          ids.push_back(__id);
+        }
+
+        if (onlyOldest && ids.size() > 1) {
```
rgacogne, line 686: I wonder if that test would make more sense inside the `sthiskey == keyString` block, since we only insert in this case? And then, perhaps, when `onlyOldest` is `true`, only clear/insert when `ts < oldestts`?
The end result would be the same, to be clear.

modules/lmdbbackend/lmdbbackend.cc
```diff
@@ -1651,22 +1659,69 @@ bool LMDBBackend::createDomain(const DNSName& domain, const DomainInfo::DomainKi
...
+      }
+      else {
+        zonemap[di.zone] = di;
+      }
+    }
+
+    for (const auto& zone : dups) {
+      DomainInfo di;
+
+      if (!(di.id = txn.get<0>(zone, di))) {
```
rgacogne, line 1686: I'm always wary of setting a value in a condition, I find it easy to misunderstand.

rgacogne: Do I correctly understand that this works because `get` will return the most recent entry? If so this might need a comment.

modules/lmdbbackend/lmdbbackend.cc
```diff
@@ -1762,26 +1807,29 @@ void LMDBBackend::setNotified(uint32_t domain_id, uint32_t serial)
...
+    ci.d_id = di.id;
+    ci.d_zone = di.zone;
+    ci.d_primaries = di.masters;
     try {
-      ci.fromJson(iter->options, type);
+      ci.fromJson(di.options, type);
     }
     catch (const std::runtime_error& e) {
-      g_log << Logger::Warning << __PRETTY_FUNCTION__ << " options '" << iter->options << "' for zone '" << iter->zone << "' is no valid JSON: " << e.what() << endl;
+      g_log << Logger::Warning << __PRETTY_FUNCTION__ << " options '" << di.options << "' for zone '" << di.zone << "' is no valid JSON: " << e.what() << endl;
```
rgacogne, line 1825: If I'm not mistaken there is a slight change in behaviour here, an exception will now only ignore all previous entries but keep the ones coming after this one and return `true`, instead of returning `false`.

modules/lmdbbackend/lmdbbackend.cc
Habbie: I don't see how to do that here - specifically, I don't see how to `getID` when I do.

omoerbeek: Right, I missed that.

ext/lmdb-safe/lmdb-typed.hh
Habbie: nice, that would entirely avoid the slower path for non-LS users. Will do.

modules/lmdbbackend/lmdbbackend.cc
Habbie: the least recent - I'll add a comment!

Habbie: also moved get outside of if

modules/lmdbbackend/lmdbbackend.cc
Habbie: this now has comments

PR #12739, opened 2023-04-12
rec: Fix uninited var causing wrong `bytes` field reported in protobuf log in rare cases

omoerbeek:
Reported by gcc build:
```
wres.cc: In function 'asyncresolve':
lwres.cc:631:26: warning: 'len' may be used uninitialized in this function [-Wmaybe-uninitialized]
  631 |       logIncomingResponse(outgoingLoggers, context ? context->d_initialRequestId : boost::none, uuid, ip, domain, type, qid, doTCP, dnsOverTLS, srcmask, len, lwr->d_rcode, lwr->d_records, queryTime, exportTypes);
      |                          ^
lwres.cc:390:10: note: 'len' was declared here
  390 |   size_t len;
      |          ^
```
While there, replace some mandatory pointers by refs

### Checklist

I have:
- [X] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/recursordist/pdns_recursor.cc
+++ b/pdns/recursordist/pdns_recursor.cc
@@ -328,12 +328,12 @@ LWResult::Result arecvfrom(PacketBuffer& packet, int /* flags */, const ComboAdd
   pident->remote = fromaddr;
 
   int ret = MT->waitEvent(pident, &packet, g_networkTimeoutMsec, &now);
+  len = 0;
 
   /* -1 means error, 0 means timeout, 1 means a result from handleUDPServerResponse() which might still be an error */
   if (ret > 0) {
     /* handleUDPServerResponse() will close the socket for us no matter what */
     if (packet.empty()) { // means "error"
-      len = 0;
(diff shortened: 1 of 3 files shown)


pdns/recursordist/pdns_recursor.cc
```diff
@@ -327,16 +327,17 @@ LWResult::Result arecvfrom(PacketBuffer& packet, int /* flags */, const ComboAdd
   pident->type = qtype;
   pident->remote = fromaddr;
 
-  int ret = MT->waitEvent(pident, &packet, g_networkTimeoutMsec, now);
+  int ret = MT->waitEvent(pident, &packet, g_networkTimeoutMsec, &now);
```
rgacogne, line 330: It might make sense to set `len = 0;` here if we want to be consistent, as we are now setting it to `0` in the `LWResult::Result::PermanentError` case but not in the `LWResult::Result::Timeout` one, if I'm not mistaken.

rgacogne (approved):

PR #12750, opened 2023-04-14
rec: Log if the answer was marked variable by SyncRes and if it was stored into  the packet cache (if !quiet)

omoerbeek:
Ideally we would log the received scope info as well, but we do not have access ot that information in the spot where the log object is created.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code


Habbie (approved):

rgacogne (approved):
Looks good! I wonder if we want to log these in the non-structured log case as well?

PR #12765, opened 2023-04-21
Run `clang-tidy` on auth, rec and dnsdist when opening PRs

fredmorcos:
### Short description
Run `clang-tidy` on auth, rec and dnsdist when opening PRs.

This adds some steps to the `build-auth`, `build-rec` and `build-dnsdist` jobs in the `build-and-test-all` workflow to lint (using `clang-tidy`) the auth, rec and dnsdist. The lint messages are then posted as Github annotations which [show up on the summary page of the action's run](https://github.com/fredmorcos/pdns/actions/runs/4763862838), and also [show up as code comments on the PR's diff](https://github.com/fredmorcos/pdns/pull/3/files). Next to the annotations, "user-friendly" messages are also output in the job's console output in case that's more convenient, here are examples for [auth](https://github.com/fredmorcos/pdns/actions/runs/4763862838/jobs/8467862341), [rec](https://github.com/fredmorcos/pdns/actions/runs/4763862838/jobs/8467863130) and [dnsdist](https://github.com/fredmorcos/pdns/actions/runs/4763862838/jobs/8467862891).

The compilation database for each product is generated during the build step using `bear` through pyinvoke tasks.

Two scripts are added as part of this PR to support the new checks: `git-filter.py` and `clang-tidy.py`. The pipeline for getting from a PR to Github annotations is as follows: `git diff | git-filter.py (ours) | clang-tidy-diff | clang-tidy.py (ours)`:

- `git-filter.py` will rewrite the diff passed to exclude any files that are not in the current product's compilation database produced by the build step.
- `clang-tidy.py` will read out the lints from `clang-tidy`'s output YAML and produce annotations in Github-compatible output.

`clang-tidy-diff` is a script provided by clang which can read a diff and produce lints only for the contents of the diff (the range of files and lines present in the diff only).

The `clang-tidy` configuration file used is the `clang-tidy.full` one. Any lints are considered a failure in the build step, this is to ensure that we really do make progress on modernizing our codebase.

It turns out that `clang-tidy` looks for its configuration file in the current directory and keeps walking up the directory tree until root (`/`) if it cannot find one. So the configuration file is symlinked in the current directory.

### Checklist
I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code


omoerbeek:
In general I like this. A question on the development process if this is merged:

What we should do with the lint warning we cannot be fixed in an easy way or short-term. For example, fixing the cognitive complexity of `serviceMain()` in recursor is not something you can do easily, it is a project in itself. Should we annotate these? Or maybe we we should start with a smaller `.clang-tidy` file? I'm a bit worried we end up in a situation where either progress is stalled or we start to ignore the linting. It's a delicate balancing act.

fredmorcos:
Personally, I think each function with lints that is being touched should be cleaned up first. This is e.g. what I do with my changes: first a commit to reformat the function, another commit to clean it up, then a commit to implement the actual functionality or fix.

I am against the idea to have a smaller `.clang-tidy` file or to ignore/bypass certain lints, it is already pretty relaxed and the lints are only applied to what is being touched. Sometimes someone unlucky will run into something like the cognitive complexity of `serviceMain()` and will have to work on that. But I guess this is going to be rare. It is also the technical debt that we have to pay for letting something like `serviceMain()` reach this point.

However, if we keep relaxing or bypassing the rules whenever they cause an inconvenience then we will never get anywhere.

omoerbeek:
In general I agree.My point is that we do not always have the opportunity to follow the royal road. If we hit the complex case, we can choose to invest time to fix the thing, or add an annotation (`// NOLINT(...)`) if no opportunity is available at that moment.  We then should create an issue to not forget to fix the code at a later moment so we can remove the annotation.

fredmorcos:
I agree that the high road isn't always possible. But also adding `// NOLINT(...)` means it will get forgotten in an issue which will get assigned to me during one of our triaging sessions, which is not fair :-D

fredmorcos:
> Creating those issues with an assignee != you from the start should prevent that...

Okay. I'm willing to accept the following: Any added `NOLINT` directive needs to have an issue title and issue number and link next to it as a comment along with the name/nickname of the person that wrote it and the issue must have an assignee and a description of what needs to be done.

fredmorcos:
> to avoid some unnecessary cognitive complexity increase induced by `SLOG` and friends.

Done.

fredmorcos:
In case there aren't any more comments for this I'd like to get it merged.

rgacogne (approved):
That's very impressive, Fred, thanks!
We might have to tune the settings a bit but that does not prevent merging this PR as far as I'm concerned. The cognitive complexity setting might be too low, for example, because if `DynBlockRulesGroup::addOrRefreshBlock()` is already too complex I'm worried we might have to rewrite the entire code base :)

fredmorcos:
BTW I believe most of these settings are the defaults. IIRC I dumped them from a clang-tidy run.

PR #12772, opened 2023-04-27
auth: Also transfer custom properties of catalog zones

rgacogne:
### Short description

Custom properties are implementation-dependent, so we might not understand them, but that should not prevent us from providing them to secondaries.
BIND defines custom properties for ACL, primaries and TSIG settings, for example: https://bind9.readthedocs.io/en/stable/chapter6.html#catalog-zone-custom-properties

My use-case is simple: I want to provision the TSIG keys and ACL to use on the secondaries for the zones learned on the consumer via the catalog zone itself.

Feedback welcome! @mind04 in particular will likely have an opinion on this :)

I have a PoC implementing consumer-side handling of some of these options to set primaries, and metadata (`AXFR-MASTER-TSIG` and `ALLOW-AXFR-FROM`) on secondaries, but this is a bit more involved so let's see how this one is received first :)

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/tcpreceiver.cc
+++ b/pdns/tcpreceiver.cc
@@ -773,12 +773,36 @@ int TCPNameserver::doAXFR(const DNSName &target, std::unique_ptr<DNSPacket>& q, 
 
   // Catalog zone start
   if (di.kind == DomainInfo::Producer) {
-    // Ignore all records except NS at apex
-    sd.db->lookup(QType::NS, target, di.id);
+    const DNSName extLabel = DNSName("ext");
+    const DNSName customPropertiesSuffix = extLabel + target;
+    const DNSName zonesPropertiesSuffix = DNSName("zones") + target;
+
+    bool seenNS = false;
+    sd.db->list(target, di.id);
     while (sd.db->get(zrr)) {
-      zrrs.emplace_back(zrr);
+      // Ignore all records except:
+      // 1. NS at apex
+      if (zrr.dr.d_name == target && zrr.dr.d_type == QType::NS) {
+        zrrs.emplace_back(zrr);
+        seenNS = true;
+      }
+      // 2. Global custom properties
+      else if (zrr.dr.d_name.isPartOf(customPropertiesSuffix)) {
+        zrrs.emplace_back(zrr);
+      }
+      // 3. Member zone custom properties
+      else if (zrr.dr.d_name.isPartOf(zonesPropertiesSuffix) && zrr.dr.d_name.countLabels() > 4) {
+        // get labels after making the name relative to zones.TARGET
+        auto labels = zrr.dr.d_name.makeRelative(zonesPropertiesSuffix).getRawLabels();
+        // check the last but one label, the last one being the unique ID of the zone
+        if (DNSName(labels.at(labels.size() - 2)) == extLabel) {
+          zrrs.emplace_back(zrr);
+        }
+      }
     }
-    if (zrrs.empty()) {
+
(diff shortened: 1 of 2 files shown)


pdns/tcpreceiver.cc
```diff
@@ -773,21 +773,33 @@ int TCPNameserver::doAXFR(const DNSName &target, std::unique_ptr<DNSPacket>& q,
...
+      }
+      // 2. Custom properties
+      else if (zrr.dr.d_name.isPartOf(customPropertiesSuffix)) {
+        zrrs.emplace_back(zrr);
+      }
     }
-    if (zrrs.empty()) {
+
+    /* if we only have the version record, synthesize a NS record */
+    if (zrrs.size() == 1) {
```
mind04, line 795: Detect ns records in the loop above and only add the default one when there are no NS records present in the zone.

rgacogne:
> I have mixed feelings about this way of implementing global custom properties. As it will make a future "real" implementation a lot harder. The current approach will work for the producer zone but it does not scale to member zones as well. 

I'm not sure that's true, wouldn't it be possible to insert entries for member zones as well using the `<property-prefix>.ext.<unique-N>.zones.$CATZ` syntax? It requires knowing the unique identifier of the zone, of course.

> The idea is to implement custom properties as a bunch of rrsets in the domain options field (formatted like rrsets in the api) in the domains table. This will work for producer and member zones and will result in reusable code and free property change detection.

I see. The thing that pains me a bit with this idea is that it does not allow using the usual way of updating of the content of zone (API, pdnsutil edit-zone) to set custom properties, and that's actually what I like the much about catalog zones: they are regular DNS zones, with a bit of automation on top, and I can use existing tools to deal with them. 

> Ps. With this pull updated global custom properties will not result in a catalog zone update.

It seems to be OK as long as the serial is incremented in my tests, but I might be missing something.

> An other approach could be to lossen it even more and add a config switch to enable zone content merging for producer zones. With this switch enabled ALL records in the db for the producer are added to the catalog (not just ns records). Enabling of this switch and adding custom records to the catalog is however unsupported/experimental.
> 
> This will give some level of custom properties and group support to those who are in desperate need but it has no negative impact on future catalog zone development.

If I understand the current catalog zones draft correctly, a catalog zone should not contain anything besides a SOA, NS records, PTR/TXT for member zones and custom properties. Given that we generate PTR/TXT records automatically for member zones, I'm not convinced there is an actual need to include anything else, and that might indeed make it more confusing.

rgacogne:
> I'm not sure that's true, wouldn't it be possible to insert entries for member zones as well using the `<property-prefix>.ext.<unique-N>.zones.$CATZ` syntax? It requires knowing the unique identifier of the zone, of course.

I see now that you are right, this PR does not support member zones indeed. I did write the code to handle them but it looks like I messed up while rebasing, or something, I'll have a look.

rgacogne:
```
Warning: function 'doAXFR' has cognitive complexity of 415 (threshold 25) (readability-function-cognitive-complexity - Level=Warning)
Warning: /home/runner/work/pdns/pdns/pdns/tcpreceiver.cc:585: function 'doAXFR' has cognitive complexity of 415 (threshold 25) (readability-function-cognitive-complexity)
```
Yeah, well..

rgacogne:
I'm closing this PR since there was no interest in it. I opened https://github.com/PowerDNS/pdns/issues/13682 to track the feature request.

PR #12776, opened 2023-04-29
Make DNSQType.TSIG available in dnsdist

jacobbunk:
### Short description
Make DNSQType.TSIG available fot matching in dnsdist.

The TSIG line was commented in 914353c, I suspect this change was committed by mistake, but it prevents a user of dnsdist from matching incoming queries with QTypeRule(DNSQType.TSIG)

### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/backends/gsql/gsqlbackend.cc
+++ b/pdns/backends/gsql/gsqlbackend.cc
@@ -2156,7 +2156,7 @@ void GSQLBackend::extractRecord(SSqlStatement::row_t& row, DNSResourceRecord& r)
 
   r.qtype=row[3];
 
-  if (d_upgradeContent && DNSRecordContent::isUnknownType(row[3]) && r.qtype.isSupportedType()) {
+  if (d_upgradeContent && DNSRecordContent::isUnknownType(row[3]) && r.qtype.isSupportedType() && !r.qtype.isMetadataType() && DNSRecordContent::isRegisteredType(r.qtype, r.qclass) && r.qtype != QType::NSEC && r.qtype != QType::NSEC3 && r.qtype != QType::ANY) {
     r.content = DNSRecordContent::upgradeContent(r.qname, r.qtype, row[0]);
   }
   else if (r.qtype==QType::MX || r.qtype==QType::SRV) {
--- a/pdns/backends/gsql/gsqlbackend.cc
+++ b/pdns/backends/gsql/gsqlbackend.cc
@@ -2156,7 +2156,7 @@ void GSQLBackend::extractRecord(SSqlStatement::row_t& row, DNSResourceRecord& r)
 
   r.qtype=row[3];
 
-  if (d_upgradeContent && DNSRecordContent::isUnknownType(row[3]) && r.qtype.isSupportedType() && !r.qtype.isMetadataType() && DNSRecordContent::isRegisteredType(r.qtype, r.qclass) && r.qtype != QType::NSEC && r.qtype != QType::NSEC3 && r.qtype != QType::ANY) {
+  if (d_upgradeContent && DNSRecordContent::isUnknownType(row[3]) && DNSRecordContent::isRegisteredType(r.qtype, r.qclass)) {
     r.content = DNSRecordContent::upgradeContent(r.qname, r.qtype, row[0]);
   }
   else if (r.qtype==QType::MX || r.qtype==QType::SRV) {
--- a/pdns/dnsparser.cc
+++ b/pdns/dnsparser.cc
@@ -195,6 +195,11 @@ DNSRecordContent::zmakermap_t& DNSRecordContent::getZmakermap()
   return zmakermap;
 }
 
+bool DNSRecordContent::isRegisteredType(uint16_t rtype, uint16_t rclass)
+{
+  return getTypemap().count(pair(rclass, rtype)) != 0;
+}
+
 DNSRecord::DNSRecord(const DNSResourceRecord& rr): d_name(rr.qname)
 {
   d_type = rr.qtype.getCode();
--- a/pdns/dnsparser.hh
+++ b/pdns/dnsparser.hh
@@ -268,7 +268,7 @@ public:
     throw runtime_error("Unknown DNS type '"+name+"'");
   }
 
-  static const string NumberToType(uint16_t num, uint16_t classnum=1)
+  static const string NumberToType(uint16_t num, uint16_t classnum = QClass::IN)
   {
     auto iter = getT2Namemap().find(pair(classnum, num));
     if(iter == getT2Namemap().end())
@@ -276,6 +276,11 @@ public:
       //      throw runtime_error("Unknown DNS type with numerical id "+std::to_string(num));
     return iter->second;
   }
+
+  /**
+   * \brief Return whether we have implemented a content representation for this type
+   */
+  static bool isRegisteredType(uint16_t rtype, uint16_t rclass = QClass::IN);
 
   virtual uint16_t getType() const = 0;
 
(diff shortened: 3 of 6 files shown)


rgacogne (approved):
Looks good to me, I don't see any reason not to have TSIG in that list. Thanks!

mind04:
> Looks good to me, I don't see any reason not to have TSIG in that list. Thanks!

I think this pull will make TSIG the only ANY class type in the list. Are you sure this is harmless in the entire code base?

Ps. This is the commit where it was disabled first https://github.com/PowerDNS/pdns/commit/b7c7d87250b6db4f8311e94d11dfd7078391c9de Doesn't smell like an accidental change.

rgacogne:
Good question:
- `QType::names` is directly used in `/pdns/dnsdist-lua-vars.cc`, `./pdns/recursordist/lua-recursor4.cc` and `./pdns/lua-base4.cc` to export the QType constants to Lua, which should be fine
- It is also used to initialize `QType::numbers`, which is never directly referenced outside of the `QType` class but used to implement `QType::toString()` and `QType::isSupportedType()`. The former should be OK. I'm a bit unsure about the semantics of the latter, which is called in `GSQLBackend::extractRecord()` to determine whether it's OK to upgrade content, and in `PacketHandler::checkUpdatePrescan()`. For these two I guess it would be wrong to allow `TSIG` records, so perhaps `QType::isSupportedType()` should be amended (and documented!) to explicitly deny `TSIG`?

omoerbeek:
`QType` is a shared type by the various products, so`QType::isSupportedType()` is a bit of a design accident. Knowing the number/name is a different thing compared to supporting something. Also: what is supported by whom? 

Note that recurorsor has its own method: `Syncres::isUnsupported`.

Habbie:
I think those two mostly want the same thing:

* a type we know about
* that actually has a content representation (this rules out OPT and hopefully TSIG)
* that is not a meta type
* that is not NSEC or NSEC3

for `extractRecord`, we want to refuse ANY. For `checkUpdatePrescan`, ANY should be allowed because it's handled a few lines down.

for 2136, we could consider expanding this to unsupported types, with TYPExx, but that's not for this PR.

rgacogne:
> * that actually has a content representation (this rules out OPT and hopefully TSIG)

Do you mean that we have implemented a `DNSRecordContent` class for that type? Or that the type has a presentation format defined in the RFC?
I have tried to implement my current understanding in https://github.com/rgacogne/pdns/commit/3762e2ec337c7415dde8b00bfc2f246b10e14e4d but I'm pretty sure I'm getting this part wrong.

Habbie:
> Do you mean that we have implemented a `DNSRecordContent` class for that type? Or that the type has a presentation format defined in the RFC?

The second. I do believe this is as simple as blocking OPT and TSIG (and the meta types).

Habbie:
Ah, of course, for upgradeContent we do need to have an implementation of the presentation format.

Habbie:
> I have tried to implement my current understanding in [rgacogne@3762e2e](https://github.com/rgacogne/pdns/commit/3762e2ec337c7415dde8b00bfc2f246b10e14e4d) but I'm pretty sure I'm getting this part wrong.

at a quick glance that seems fine!

rgacogne:
> Ah, of course, for upgradeContent we do need to have an implementation of the presentation format.

We do not need that for `PacketHandler::checkUpdatePrescan`, I gather?

Habbie:
> We do not need that for `PacketHandler::checkUpdatePrescan`, I gather?

agreed

rgacogne:
I have pushed an updated commit: https://github.com/PowerDNS/pdns/commit/b6591fa590e43fcace9a05fe34424b8a59cac10c which:

- adds a new method, `DNSRecordContent::isRegisteredType()`, that can be used to check whether we have implemented a `DNSRecordContent` class for a specific type ;
-  updates `QType::isMetadataType()` to match what is defined in `rfc6895 section 3.1`. This adds `QType::OPT` and `QType::TKEY`  to the list of types for which this function will return true ;
- updates `GSQLBackend::extractRecord()` to use the new function, checking that a type is not a metadata one, that we have an implementation for it, and that it is neither a `NSEC`, `NSEC3` or `ANY` ;
- updates a few formatting nits in `PacketHandler::checkUpdatePrescan()` but that's it, since we already check if the type is a metadata one, handle `ANY`, and later deal with `NSEC` and `NSEC3` as well.

rgacogne:
Requesting a review from either Peter or Otto, because I cannot in good conscience review my own changes :-)

Habbie:
Your changes do not appear to be in this PR

rgacogne:
They are not, which made sense to me at the time but makes no sense for a review, I will push them!

rgacogne:
Done

Habbie (approved):

pdns/backends/gsql/gsqlbackend.cc
```diff
@@ -2156,7 +2156,7 @@ void GSQLBackend::extractRecord(SSqlStatement::row_t& row, DNSResourceRecord& r)
 
   r.qtype=row[3];
 
-  if (d_upgradeContent && DNSRecordContent::isUnknownType(row[3]) && r.qtype.isSupportedType()) {
+  if (d_upgradeContent && DNSRecordContent::isUnknownType(row[3]) && r.qtype.isSupportedType() && !r.qtype.isMetadataType() && DNSRecordContent::isRegisteredType(r.qtype, r.qclass) && r.qtype != QType::NSEC && r.qtype != QType::NSEC3 && r.qtype != QType::ANY) {
```
Habbie, line 2159: @mind04 pointed out to me that, while NSEC/NSEC3 should never be in the records table, if they are, we probably do still want to convert them.

Habbie (changes requested):

pdns/backends/gsql/gsqlbackend.cc
Habbie: how about: `d_upgradeContent && DNSRecordContent::isUnknownType(row[3]) && r.qtype.isSupportedType() && !r.qtype.isMetadataType() && r.qclass == QClass::IN && r.qtype != QType::ANY`

Habbie: because I now realise we don't really care whether the type and class are valid together; our database is strictly class IN

Habbie: I missed that `registered` actually checks that we can convert it from/to text. So, perhaps it can be as short as: `d_upgradeContent && DNSRecordContent::isUnknownType(row[3]) && DNSRecordContent::isRegisteredType(QClass::r.qtype, r.qclass)`

Habbie: I was typing a comment and then forgot to post it. Looking at the code, `regist` skips the typemap registration because of the `nullptrs`, which indeed matches @mind04's observation.

rgacogne: Alright, thanks! I just applied the change.

Habbie: I don't see it

rgacogne: Ugh, I had pushed it to my own fork *again*, sorry! Sleep really is underrated.

Habbie (approved):

pdns/backends/gsql/gsqlbackend.cc
jacobbunk: Please feel free to do this PR against your own fork if that makes your life easier, @rgacogne

Habbie: thanks Jacob, I believe we're ready to merge :)

PR #12780, opened 2023-05-02
auth: add setting workaround-11804-max-chunk-records

zeha:
### Short description

Workaround for https://github.com/PowerDNS/pdns/issues/11804. Defaults to previously hard-coded value of 100.

I will add docs.

### Checklist

I have:
- [ ] read the [CONTRIBUTING.md](https://github.com/PowerDNS/pdns/blob/master/CONTRIBUTING.md) document
- [x] compiled this code
- [ ] tested this code
- [ ] included documentation (including possible behaviour changes)
- [ ] documented the code
- [ ] added or modified regression test(s)
- [ ] added or modified unit test(s)

PR #12784, opened 2023-05-03
DRAFT: auth: avoid overflowing AXFR reply packet

zeha:
### Short description

Discussion basis for a potential fix for #11804.

This **works** with a test zone with very large TXT RRs. It is obviously not a fix for the other sendPacket cases in doAXFR, but if this is mostly acceptable, I will try to dedup the sendPacket callers and handle the "rechunking" in a single place.

### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code


pdns/dnspacket.hh
```diff
@@ -58,7 +58,7 @@ public:
 
   int noparse(const char *mesg, size_t len); //!< just suck the data inward
   int parse(const char *mesg, size_t len); //!< parse a raw UDP or TCP packet and suck the data inward
-  const string& getString(bool throwsOnTruncation=false); //!< for serialization - just passes the whole packet. If throwsOnTruncation is set, an exception will be raised if the records are too large to fit inside a single DNS payload, instead of setting the TC bit
+  const string& getString(bool leaveOverflow=false); //!< for serialization - just passes the whole packet. If leaveOverflow is set, records not fitting into a single DNS packet will be left in d_leftRRs and TC will not be set.
```
rgacogne, line 61: We might need both options:
- throw if the records do not fit into a TCP packet at all, for non-XFR case
- keep non-fitting records around, for XFR cases
?

rgacogne (commented):
The general idea looks OK to me. I made a few comments even though I know this is just a draft/PoC :)

Habbie:
I milestoned this for auth 4.9, as something like #12780 seems more fitting for the scope of 4.8

PR #12788, opened 2023-05-04
clang-tidy: Increase the code complexity threshold to 50 for now

rgacogne:
### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code


fredmorcos (approved):
I personally think 100 is too big of a jump from 25, especially that `DownstreamState::reconnect()` is at about 45. I would prefer to increase it by small amounts until we find something that works for us.

Ultimately I would leave that decision in your hands, so I'm approving the PR.

rgacogne:
OK, let's try 50.

fredmorcos (approved):

PR #12808, opened 2023-05-12
Reformat and delint arguments.cc and arguments.hh

omoerbeek:
### Checklist

I have:
- [X] compiled this code
- [ ] tested this code
- [ ] documented the code


pdns/arguments.cc
```diff
@@ -35,126 +35,124 @@
...
 
   stringtok(parts, param->second, ", \t");
-  for (const auto& part: parts) {
-    if (part == val) {
-      return true;
-    }
-  }
-
-  return false;
+  return std::any_of(parts.begin(), parts.end(), [&](const std::string& str) { return str == val; });
```
fredmorcos, line 119: This was available as an LSP/eglot push-button action, right?

fredmorcos (approved):
I'm fine with the nolints here. The `readdir` rabbit hole is annoying though.

pdns/arguments.cc
omoerbeek: yes, iirc

pdns/arguments.cc
```diff
@@ -190,30 +192,34 @@ string ArgvMap::configstring(bool running, bool full)
...
   help += formatOne(running, full, "ignore-unknown-settings", helpmap["ignore-unknown-settings"], defaultmap["ignore-unknown-settings"], d_params["ignore-unknown-settings"]);
 
-  for (const auto& i : helpmap) {
-    if (d_typeMap[i.first] == "Command")
+  for (const auto& helpietm : helpmap) {
+    if (d_typeMap[helpietm.first] == "Command") {
       continue;
-    if (i.first == "ignore-unknown-settings")
+    }
+    if (helpietm.first == "ignore-unknown-settings") {
```
Habbie, line 209: maybe fix this typo while you're here :)

Habbie (changes requested):
generally approved. One nit that was there already but while you're here, maybe fix it.

I think you can drop the last two commits, together, with no change to PR outcome?

Habbie (approved):

PR #12810, opened 2023-05-12
Run build-and-test-all.yml on a custom Debian 11 container instead of an Ubuntu Runner VM

romeroalx:
### Short description
Run build-and-test-all.yml actions on a custom Debian 11 (bullseye) container. Uses a pre-generated Debian 11 docker image with pre-loaded packages. Some relevant changes:

- IPv6 tests skipped. IPv6 is not available for containers.
- Clang upgraded to version 13
- For PDNS backend tests, to allow communication between Docker containers, the default docker bridge IPv4 address is used instead of localhost.

### Checklist

I have:
- [ ] compiled this code
- [x] tested this code
- [ ] documented the code


rgacogne:
> IPv6 tests skipped. IPv6 is not available for containers.

Ouch, that's sad (but completely not your fault, I know!).

tasks.py
```diff
@@ -5,6 +5,10 @@
 import sys
 import time
 
+auth_backend_ip_addr = os.environ.get('AUTH_BACKEND_IP_ADDR') if os.environ.get('AUTH_BACKEND_IP_ADDR') else "127.0.0.1"
+
+clang_version = os.environ.get('CLANG_VERSION') if os.environ.get('CLANG_VERSION') else "12"
```
fredmorcos, line 10: Can we default to "13" here?

.github/workflows/build-and-test-all.yml
```diff
@@ -34,23 +44,21 @@ jobs:
...
       - name: let GitHub cache our ccache data
         uses: actions/cache@v3
         with:
           path: ~/.ccache
           key: auth-ccache-${{ steps.get-stamp.outputs.stamp }}
           restore-keys: auth-ccache-
-      - run: build-scripts/gh-actions-setup-inv  # this runs apt update+upgrade
-      - run: inv install-clang
       - run: inv install-clang-tidy-tools
-      - run: inv install-auth-build-deps
```
fredmorcos, line 46: Where did those go? I'm guessing into the base image? If yes, can we also just install the clang-tidy-tools in the base image instead of installing them for each product?

fredmorcos (changes requested):

fredmorcos:
I know it's still a draft and I hope I'm not getting ahead of things :-)

tasks.py
romeroalx: Ok! I will update this to "13"

.github/workflows/build-and-test-all.yml
romeroalx: Yes, the idea is that all those packages were already installed while building the base image. I will add that step also there and remove it from here

tasks.py
```diff
@@ -5,6 +5,10 @@
 import sys
 import time
 
+auth_backend_ip_addr = os.environ.get('AUTH_BACKEND_IP_ADDR') if os.environ.get('AUTH_BACKEND_IP_ADDR') else "127.0.0.1"
+
+clang_version = os.environ.get('CLANG_VERSION') if os.environ.get('CLANG_VERSION') else "13"
```
Habbie, line 10: you can just do `os.environ.get('CLANG_VERSION', 13)` because get is defined as `get(key, default=None)`

tasks.py
```diff
@@ -509,13 +513,13 @@ def ci_dnsdist_configure(c, features):
     sanitizers = ' '.join('--enable-'+x for x in os.getenv('SANITIZERS').split('+')) if os.getenv('SANITIZERS') != '' else ''
```
Habbie, line 513: I was going to suggest `os.environ.get` but it turns out `os.getenv` also takes a second `default=` argument, so either would work

Habbie (commented):
just one nit (repeated 3 times), otherwise, great! if you fix those few things, we can merge this.

I think we should make a ticket reminding us to figure out IPv6 later.

tasks.py
romeroalx: Done

PR #12793, opened 2023-05-09
Another set of delinting

omoerbeek:
A set of random files made clean. During this process .clang-tidy.full was also amended a bit.

### Checklist

I have:
- [X] read the [CONTRIBUTING.md](https://github.com/PowerDNS/pdns/blob/master/CONTRIBUTING.md) document
- [X] compiled this code
- [ ] tested this code
- [ ] included documentation (including possible behaviour changes)
- [ ] documented the code
- [ ] added or modified regression test(s)
- [ ] added or modified unit test(s)
- [ ]  checked that this code was merged to master

PR #12813, opened 2023-05-12
clang-tidy: Disable cppcoreguidelines-avoid-non-const-global-variables

rgacogne:
### Short description

While I agree that we should try to avoid non-const global vars as much as possible, this rule kills the boost testing framework so we cannot keep it.

### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code


fredmorcos:
There's got to be a way to make use of the boost test suite without triggering those warnings. So we can disable it for now, I'm fine with that, but we need to have an issue open to re-enable it if there's a path forward.

omoerbeek:
A problem wit this is that this warning is also triggered for private static class variables. IMO that is going too far. See #12793. So I approve.

omoerbeek (approved):

rgacogne:
Opened an issue so we can reconsider later: https://github.com/PowerDNS/pdns/issues/12816

PR #12858, opened 2023-05-27
auth API: Allow '*' in record names (rfc4592).

anp135:
### Short description

Allow to use wildcard records like `_acme-challenge.*.host` according rfc4592.

This is needed for provide dns challenge LetsEncrypt for wildcard zones.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/ws-api.cc
+++ b/pdns/ws-api.cc
@@ -391,7 +391,7 @@ string apiZoneNameToId(const DNSName& dname) {
 }
 
 void apiCheckNameAllowedCharacters(const string& name) {
-  if (name.find_first_not_of("abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ01234567890_/.-") != std::string::npos)
+  if (name.find_first_not_of("abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ01234567890_/.-*") != std::string::npos)
     throw ApiException("Name '"+name+"' contains unsupported characters");
 }
 


kpfleming:
If you don't mind, please update the summary to something like "auth API: Allow '*' in record names." The current summary is not very specific and implies that wildcard records are not currently supported at all :-)

anp135:
> If you don't mind, please update the summary to something like "auth API: Allow '*' in record names." The current summary is not very specific and implies that wildcard records are not currently supported at all :-)

Is it ok now?

kpfleming:
thanks!

mind04:
> Allow to use wildcard records like `_acme-challenge.*.host` according rfc4592.
This is not a wildcard record, as explained in section 2.2.1 and 2.2.2 of rfc4592

anp135:
> > Allow to use wildcard records like `_acme-challenge.*.host` according rfc4592.
> 
> This is not a wildcard record, as explained in section 2.1.2 and 2.2.1 of rfc4592

Perhaps I wrong select RFC for explain. The goal in to use `*` in records. The real case - DNS challenge LE.
I already did that in 4.6.1 for dcape (we are use pdns auth in https://github.com/dopos/powerdns-alpine) and forget make PR.

Habbie:
> This is needed for provide dns challenge LetsEncrypt for wildcard zones.

can you provide a reference for this?

anp135:
> > This is needed for provide dns challenge LetsEncrypt for wildcard zones.
> 
> can you provide a reference for this?

https://letsencrypt.org/docs/challenge-types/

So in case of willdcard using in [dcape](https://github.com/dopos/dcape/blob/v2/apps/traefik/traefik.acme.yml#L25-L29)
when we did `host A` and `*.host CNAME host` we need `_acme-challenge.host` and `_acme-challenge.*.host` records for DNS challenges for that. For example you can dig `s1.racedb.pro`, '*.s1.racedb.pro`, `_acme-challenge.s1.racedb.pro`,`_acme-challenge.*.s1.racedb.pro`.

As we use powerdns auth in our project, we need to manage records `_acme-challenge.*.host`. We did this in https://github.com/dopos/powerdns-alpine with [patch](https://github.com/dopos/powerdns-alpine/blob/master/allow_wildcard.patch).

mnordhoff:
Let's Encrypt will not send a DNS query for the literal name `_acme-challenge.*.example.com`.

To validate `foo.example.com`, they query for `_acme-challenge.foo.example.com`.

For `*.example.com`, they query for `_acme-challenge.example.com`.

anp135:
> Let's Encrypt will not send a DNS query for the literal name `_acme-challenge.*.example.com`.
> 
> To validate `foo.example.com`, they query for `_acme-challenge.foo.example.com`.
> 
> For `*.example.com`, they query for `_acme-challenge.example.com`.

Cert for *`.example.com` with DNS challenge will be done only after successfull txt query with `_acme-challenge.*.example.com`. Receive cert for wildcard (`*.example.com`) with http challenge impossible -)

Habbie:
> Cert for *`.example.com` with DNS challenge will be done only after successfull txt query with `_acme-challenge.*.example.com`.

I believe this is not true. Can you show documentation that states this clearly? Alternatively, can you provide traffic dumps that clearly show LE requesting that TXT?

anp135:
> > Cert for *`.example.com` with DNS challenge will be done only after successfull txt query with `_acme-challenge.*.example.com`.
> 
> I believe this is not true. Can you show documentation that states this clearly? Alternatively, can you provide traffic dumps that clearly show LE requesting that TXT?

Yes, you right.
Right now not needed so no more cases with `_acme-challenge.*.example.com`

Habbie:
Thank you for reporting back! I closed this now.

PR #12859, opened 2023-05-28
pdnsutil: if user pushes unknown key in response to "problem with zone" prompt, do not throw away their changes

Habbie:
### Short description
Before this patch, we clear the delta if the user does not select "apply", including when the user hits an unknown key. With this patch, we only clear on the specific choices that want that behaviour.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


mind04 (approved):

omoerbeek (approved):
Ugly code... but OK.

Habbie:
> Ugly code... but OK.

yes.. it's no surprise we find a bug like this once or twice a year.

PR #12865, opened 2023-05-30
rec: Support for populating forward-zones using catalog zones

kpfleming:
### Short description
Adds support for catalog zones in recursor; their contents are used to populate the `forward-zones` list.

### Checklist

I have:
- [x] compiled this code
- [ ] tested this code
- [x] documented the code

### DIFF
--- a/pdns/recursordist/docs/settings.rst
+++ b/pdns/recursordist/docs/settings.rst
@@ -110,7 +110,7 @@ Like `allow-notify-for`_, except reading from file. To use this
 feature, supply one domain name per line, with optional comments
 preceded by a "#".
 
-NOTIFY-allowed zones can also be specified using `forward-zones-file`_.
+NOTIFY-allowed zones can also be specified using `forward-zones-file`_ and `forward-zones-catalog`_.
 
 .. _setting-allow-notify-from:
 
@@ -136,7 +136,7 @@ NOTIFY operations received from a client listed in one of these netmasks
 will be accepted and used to wipe any cache entries whose zones match
 the zone specified in the NOTIFY operation, but only if that zone (or
 one of its parents) is included in `allow-notify-for`_,
-`allow-notify-for-file`_, or `forward-zones-file`_ with a '^' prefix.
+`allow-notify-for-file`_, `forward-zones-file`_ with a '^' prefix, or `forward-zones-catalog`_.
 
 .. _setting-allow-notify-from-file:
 
@@ -574,7 +574,7 @@ This can have odd effects, depending on your network, and may even be a security
 Therefore, the PowerDNS Recursor by default does not query private space IP addresses.
 This setting can be used to expand or reduce the limitations.
 
-Queries for names in forward zones and to addresses as configured in any of the settings `forward-zones`_, `forward-zones-file`_ or `forward-zones-recurse`_ are performed regardless of these limitations.
+Queries for names in forward zones and to addresses as configured in any of the settings `forward-zones`_, `forward-zones-file`_, `forward-zones-recurse`_, or `forward-zones-catalog`_ are performed regardless of these limitations.
 
 .. _setting-ecs-add-for:
 
@@ -883,6 +883,43 @@ This allows e.g. a forward to a local authoritative server holding a copy of the
 To prevent this, add a Negative Trust Anchor (NTA) for this zone in the `lua-config-file`_ with ``addNTA("your.zone", "A comment")``.
 If this forwarded zone is signed, instead of adding NTA, add the DS record to the `lua-config-file`_.
 See the :doc:`dnssec` information.
+
+.. _setting-forward-zones-catalog
+
+``forward-zones-catalog``
+-------------------------
+.. versionadded:: 4.10.0
+-  'zonename=IP[:port]'
+
+The contents of the specified ``catalog zone`` will be retrieved from
+the server at the specified IP (or IP:port).  Each entry in the
+catalog zone will be added to the `forward-zones`_ list.
+
+.. code-block:: none
+
+    forward-zones-catalog=catalog1.example.org=203.0.113.210
+    forward-zones-catalog=catalog2.example.org=[2001:DB8::1:3]:5300
+
+The catalog zone itself will be added to the `allow-notify-for`_ list,
+but when NOTIFY queries are received the recursor will not perform any
+cache-clearing operations; instead it will retrieve the SOA record for
+the zone, and if the serial number is higher than the serial number
+last seen, the zone contents will be retrieved. Any zones which have
+been removed from the catalog zone will also be removed from the
+`forward-zones`_ list; any zones which have been added to the catalog
+zone will be added to the `forward-zones`_ list.
+
+Forwarded queries for zones obtained from a catalog zone will be sent
+to the IP/port used to retrieve the contents of the catalog zone.
+
+Forwarded queries for zones obtained from a catalog zone will have the
+``recursion desired (RD)`` bit set to ``0``. See the `forward-zones`_
+setting for more details.
+
+Zones added to the `forward-zones`_ list due to inclusion in the
+catalog zone will also be added to the `allow-notify-for`_ list; the
+recursor's behavior when a NOTIFY query is received for such a zone
+will be the same as described in that setting.
 
 .. _setting-forward-zones-file:
 
--- a/pdns/recursordist/docs/settings.rst
+++ b/pdns/recursordist/docs/settings.rst
@@ -110,7 +110,7 @@ Like `allow-notify-for`_, except reading from file. To use this
 feature, supply one domain name per line, with optional comments
 preceded by a "#".
 
-NOTIFY-allowed zones can also be specified using `forward-zones-file`_ and `forward-zones-catalog`_.
+NOTIFY-allowed zones can also be specified using `forward-zones-file`_ and `forward-zones-catalogs`_.
 
 .. _setting-allow-notify-from:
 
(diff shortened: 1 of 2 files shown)


pdns/recursordist/docs/settings.rst
```diff
@@ -865,6 +865,43 @@ To prevent this, add a Negative Trust Anchor (NTA) for this zone in the `lua-con
...
+the server at the specified IP (or IP:port).  Each entry in the
+catalog zone will be added to the `forward-zones`_ list.
+
+.. code-block:: none
+
+    forward-zones-catalog=catalog1.example.org=203.0.113.210
+    forward-zones-catalog=catalog2.example.org=[2001:DB8::1:3]:5300
+
+The catalog zone itself will be added to the `allow-notify-for`_ list,
+but when NOTIFY queries are received the recursor will not perform any
```
omoerbeek, line 885:
```suggestion
but when NOTIFY queries are received for the catalog zone the recursor will not perform any
```

pdns/recursordist/docs/settings.rst
```diff
@@ -865,6 +865,43 @@ To prevent this, add a Negative Trust Anchor (NTA) for this zone in the `lua-con
...
+cache-clearing operations; instead it will retrieve the SOA record for
+the zone, and if the serial number is higher than the serial number
```
omoerbeek, line 887:
```suggestion
the catalog zone, and if the serial number is higher than the serial number
```

pdns/recursordist/docs/settings.rst
```diff
@@ -865,6 +865,43 @@ To prevent this, add a Negative Trust Anchor (NTA) for this zone in the `lua-con
...
+last seen, the zone contents will be retrieved. Any zones which have
```
omoerbeek, line 888:
```suggestion
last seen, the catalog zone contents will be retrieved. Any zones which have
```

pdns/recursordist/docs/settings.rst
```diff
@@ -865,6 +865,43 @@ To prevent this, add a Negative Trust Anchor (NTA) for this zone in the `lua-con
 If this forwarded zone is signed, instead of adding NTA, add the DS record to the `lua-config-file`_.
 See the :doc:`dnssec` information.
 
+.. _setting-forward-zones-catalog
+
+``forward-zones-catalog``
+-------------------------
+.. versionadded:: 4.10.0
+-  'zonename=IP[:port]'
```
omoerbeek, line 873:
```suggestion
-  List of 'zonename=IP[:port]', separared by semicolons
```

pdns/recursordist/docs/settings.rst
```diff
@@ -865,6 +865,43 @@ To prevent this, add a Negative Trust Anchor (NTA) for this zone in the `lua-con
...
+
+The contents of the specified ``catalog zone`` will be retrieved from
+the server at the specified IP (or IP:port).  Each entry in the
+catalog zone will be added to the `forward-zones`_ list.
+
+.. code-block:: none
+
+    forward-zones-catalog=catalog1.example.org=203.0.113.210
+    forward-zones-catalog=catalog2.example.org=[2001:DB8::1:3]:5300
```
omoerbeek, line 882:
```suggestion
    forward-zones-catalog=catalog1.example.org=203.0.113.210;
    forward-zones-catalog+=catalog2.example.org=[2001:DB8::1:3]:5300;
```

omoerbeek (commented):
I like the general idea! I added some text clarification, assuming you want to be able to specify more than one catalog zone. The settings syntax needs to take that into account.

pdns/recursordist/docs/settings.rst
```diff
@@ -865,6 +865,43 @@ To prevent this, add a Negative Trust Anchor (NTA) for this zone in the `lua-con
...
 See the :doc:`dnssec` information.
 
+.. _setting-forward-zones-catalog
+
+``forward-zones-catalog``
+-------------------------
+.. versionadded:: 4.10.0
+-  'zonename=IP[:port]'
+
+The contents of the specified ``catalog zone`` will be retrieved from
```
omoerbeek, line 875:
```suggestion
The contents of the specified ``catalog zones`` will be retrieved from
```

pdns/recursordist/docs/settings.rst
```diff
@@ -865,6 +865,43 @@ To prevent this, add a Negative Trust Anchor (NTA) for this zone in the `lua-con
...
+the server at the specified IP (or IP:port).  Each entry in the
+catalog zone will be added to the `forward-zones`_ list.
```
omoerbeek, line 877:
```suggestion
catalog zones will be added to the `forward-zones`_ list.
```

kpfleming:
Indeed it needs to be possible to specify more that one catalog zone, which means this config option needs to have a plural name instead of singular (`forward-zones-catalog` -> `forward-zones-catalogs`). The double plural is a bit odd, but it's correct (one or more catalog zones which each provide one or more forward zones).

pdns/recursordist/docs/settings.rst
kpfleming: The config parser supplies commas between items if `+=` is used, so the text will need to indicate that (as it does for other options already).

kpfleming:
All of the suggestions have been incorporated while rewording some of the content to better reflect the ability to specify more than one catalog zone. Thanks!

pdns/recursordist/docs/settings.rst
```diff
@@ -884,6 +884,45 @@ To prevent this, add a Negative Trust Anchor (NTA) for this zone in the `lua-con
...
+not perform any cache-clearing operations; instead it will retrieve
+the SOA record for the catalog zone in the NOTIFY query, and if the
+serial number is higher than the serial number last seen, the zone
+contents will be retrieved. Any zones in the `forward-zones`_ list
+which were learned from the previous version of the catalog zone, and
+which are no longer present in the catalog zone, will be removed from
+the `forward-zones`_ list; any zones which have been added to the
+catalog zone will be added to the `forward-zones`_ list.
+
+Forwarded queries for zones obtained from a catalog zone will be sent
```
franklouwers, line 914: Could we allow for specifying a different server/ip to forward to?

kpfleming: It's certainly possible, now that the recursor supports config in YAML, more complex configurations are practical to express.

SrX: I'm not sure if recursor (consumer) will be able to read catalog zone from another consumer (secondary auth), but this would the clear use case.
Usually in complex architectures, producer of catalog usually not gets the queries, consumer receive the user traffic.

Having the option to forward to different set of IPs, would be great.

omoerbeek:
I intend to work on this very soon. I'm not taking the suggested documentation though, as this feature requires a more complex configuration. What I am proposing is something like this (this is thinking in progress and subject to change!)

YAML sketch

```YAML
forward_catalog_zones:
    - name: name of catalog zone
      addresses: [] Sequence of SocketAddress
      zoneSizeHint: number
      tsig:
          name: string
          algo: string
          secret: base64string
      refresh: number
      maxReceivedMBytes: number
      localAddress: IP address
      axfrTimeout: number
      group_property: forward_rec_powerdns_com
      custom_property: forward.rec.powerdns.com
      forwarders: [] Seqence of SocketAddress
      recurse: boolean
      allow_notify: boolean
    - ...
``` 
THe first fields define all the parameters needed to retrieve a zone from an autoritatieve server using (A/I)XFR, compare RPZ configuration.

Then come two properties (`group_property` and `custom_property`) which define how to interpret certain records in the catalog zone.

The last three fields define the defaults for members in the catalog zone. A catalog zone itself
can override these properties by using group properties or global custom properties. The names of the
group or private properties are defined by the `group_property` and `custom_property` YAML fields. The
values are TXT record containing an object that defines the values for the three YAML fields
`forwarders`, `recurse` and `allow_notify`.

In the case of a group property, the key should be the string defined as group_property and the
value should be a YAML object containing the values to be overwritten.

In the case of a custom property, the TXT value should be a YAML object containing the values to be
overwritten.

An example of a group property for a specific member:

    group.nfwxa33.zones.catalog.invalid 0 TXT "forward_rec_powerdns_com: { forwarders: ['127.0.0.:5301'], recurse: true}"

A global custom property

    forward.rec.powerdns.com.ext.catalog.invalid 0 TXT "{ forwarders: ['127.0.0.:5301'], recurse: true}"

The values are processed in a hierarchical way. Values in the YAML settings are overwritten by global
custom properties which are overwritten by group properties (all if applicable).

Maybe we also want a way to select only specifc members, based on a group property. E.g. we could define in the YAML settings:

``` YAML  
groups: [GroupA, GroupB]
```
to only select members that have a group property starting with `GroupA` or `GroupB`.

omoerbeek:
> This seems like a good configuration model to me, although I'm still a fan of putting the 'zone retrieval' parts in a top-level `zone` definition, the same as they would be for an RPZ or any other zone held in the recursor.

I adapted the proposed YAML.

Habbie:
> Then come two properties (`group_property` and `custom_property`) which define how to interpret certain records in the catalog zone.

Proposal: don't do `custom_property` until it turns out somebody really needs it. `group_property` doesn't look right. The name of the group property in a catalog zone is `group`. The context is free form, but I suggest making it a user configurable key, like `poolA`, `poolB`, etc., which then each get a YAML part that looks just like `defaults`.

So:

```yaml
forwarding_catalog_zones:
    - name: name of catalog zone
      allow_notify: boolean, default false
      xfr:
         addresses: [] Sequence of SocketAddress
         zoneSizeHint: number
         tsig:
           name: string
           algo: string
           secret: base64string
         refresh: number
         maxReceivedMBytes: number
         localAddress: IP address
         axfrTimeout: number
      defaults:
        forwarders: [] Sequence of SocketAddress
        recurse: boolean, default false
        allow_notify: boolean default false
      groups:  # this entire section is optional
        poolA:
          forwarders: [] Sequence of SocketAddress
          recurse: boolean, default false
          allow_notify: boolean default false
        customer835:
          forwarders: ...
          ...
    - ...
```

omoerbeek:
Superseded by #14759

PR #12866, opened 2023-05-30
auth lmdb: delete duplicate domain entries in deleteDomain

Habbie:
### Short description
also enforce shards=1 in LS mode

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


modules/lmdbbackend/lmdbbackend.cc
```diff
@@ -673,6 +673,10 @@ LMDBBackend::LMDBBackend(const std::string& suffix)
     d_random_ids = true;
     d_handle_dups = true;
     LMDBLS::s_flag_deleted = true;
+
+    if (atoi(getArg("shards").c_str()) != 1) {
```
omoerbeek, line 677: Not a fan of atoi due to its lack of error handling

modules/lmdbbackend/lmdbbackend.cc
```diff
@@ -1297,56 +1306,62 @@ bool LMDBBackend::deleteDomain(const DNSName& domain)
...
     auto txn = d_tdomains->getROTransaction();
 
     DomainInfo di;
-    id = txn.get<0>(domain, di);
+    idvec.push_back(txn.get<0>(domain, di));
   }
+  else {
+    auto txn = d_tdomains->getROTransaction();
+
+    txn.get_multi<0>(domain, idvec);
```
omoerbeek, line 1323: This part I do not follow. AFAIKS, it throws always

omoerbeek (commented):
Two comments. Second one might be important. I do not know the proper way to do transactions with lmdb, so I cannot tell if that part is correct.

modules/lmdbbackend/lmdbbackend.cc
Habbie: agreed, but currently this mirrors usage elsewhere in the file

modules/lmdbbackend/lmdbbackend.cc
Habbie: get_multi pushes into idvec, does that help?

omoerbeek: I get that, but the throw below is unconditional...

modules/lmdbbackend/lmdbbackend.cc
```diff
@@ -1297,56 +1306,62 @@ bool LMDBBackend::deleteDomain(const DNSName& domain)
...
+    throw std::runtime_error("in LMDBBackend::deleteDomain, domain was not found");
+  }
+
+  startTransaction(domain, idvec[0]);
+  for (auto id : idvec) {
```
Habbie, line 1329: test

modules/lmdbbackend/lmdbbackend.cc
Habbie: there was an if there, looks like i lost a commit during a rebase..

modules/lmdbbackend/lmdbbackend.cc
```diff
@@ -1297,56 +1306,62 @@ bool LMDBBackend::deleteDomain(const DNSName& domain)
...
+      }
 
-    for (auto _id : ids) {
-      txn.del(_id);
+      txn.commit();
     }
 
+    // Remove records
+    commitTransaction();
+    startTransaction(domain, id);
```
mind04, line 1360: Keep using transactionDomain, transactionDomainId here

mind04 (changes requested):

modules/lmdbbackend/lmdbbackend.cc
```diff
@@ -1297,56 +1306,65 @@ bool LMDBBackend::deleteDomain(const DNSName& domain)
...
     DomainInfo di;
-    id = txn.get<0>(domain, di);
+    idvec.push_back(txn.get<0>(domain, di));
+  }
+  else {
+    auto txn = d_tdomains->getROTransaction();
+
+    txn.get_multi<0>(domain, idvec);
+
+    if (idvec.size() == 0) {
```
omoerbeek, line 1325:
```suggestion
    if (idvec.empty()) {
```

modules/lmdbbackend/lmdbbackend.cc
Habbie: and also right before this `for` loop then, I assume

Habbie: pushed

mind04 (approved):

PR #12884, opened 2023-06-02
rec: reduce memory needed to manage RPZs

### LINKED ISSUE #12870: rec: optimize RPZ memory usage
Currently a single RPZ line translates into `Policy` object that takes more than 100 bytes. A big RPZ (one million records) takes more than 100 megabytes.

There are a couple of things that could be optimized:

1. `d_custom` could move to a unique pointer, only allocated if actual custom records are present. Current size of the field is 24, would become 8.
2. `d_trigger` is only used if a hit has occurred. No need to store these in the in memory RPZ Policy. Ponder an `AppliedPolicy` struct, that includes or extends the existing `Policy` without the `d_trigger` field and ads `d_trigger` itself. Saving: 24 bytes per `Policy` object. `Syncres` would use `ApliedPolicy` but the in-memory objects would be `Policy`.
3. Same for `d_hit`: also saves 24 bytes per `Policy`.

Total saving: 16 + 24 + 24 is 64 bytes per `Policy object`. So this cuts the memory usage for RPZ roughly in half.

omoerbeek:
### Short description

Should fix #12870

In one test I could see the required memory drop from 1.7G to 1.0G. This need more tests and eyes, in particular validation of the copy/move constructors and assignment ops. So I'm marking it as draft. 

Inlcudes delinting of related files.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/recursordist/filterpo.hh
+++ b/pdns/recursordist/filterpo.hh
@@ -28,6 +28,7 @@
 #include <map>
 #include <unordered_map>
 #include <limits>
+#include <utility>
 
 /* This class implements a filtering policy that is able to fully implement RPZ, but is not bound to it.
    In other words, it is generic enough to support RPZ, but could get its data from other places.
@@ -84,7 +85,7 @@ public:
     NSDName,
     NSIP
   };
-  typedef uint16_t Priority;
+  using Priority = uint16_t;
   static const Priority maximumPriority = std::numeric_limits<Priority>::max();
 
   static std::string getKindToString(PolicyKind kind);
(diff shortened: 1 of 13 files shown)


pdns/recursordist/filterpo.hh
```diff
@@ -160,35 +198,65 @@ public:
...
-    std::vector<DNSRecord> getRecords(const DNSName& qname) const;
+    [[nodiscard]] std::vector<DNSRecord> getCustomRecords(const DNSName& qname, uint16_t qtype) const;
+    [[nodiscard]] std::vector<DNSRecord> getRecords(const DNSName& qname) const;
+
+    void allocateCustomRecords(size_t res)
+    {
+      if (d_custom == nullptr) {
+        d_custom = std::make_unique<std::vector<std::shared_ptr<const DNSRecordContent>>>();
+      }
+      d_custom->reserve(res);
```
rgacogne, line 222:
```suggestion
      d_custom->reserve(d_custom.size() + res);
```

omoerbeek:
Mostly obsoleted by #13462

PR #12892, opened 2023-06-06
YaHTTP: Prevent integer overflow on very large chunks

rgacogne:
### Short description

If the chunk_size is very close to the maximum value of an integer, we trigger an integer overflow when checking if we have a trailing newline after the payload.
Reported by OSS-Fuzz as:
* https://oss-fuzz.com/testcase-detail/6439610474692608
* https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=56804

Upstream request at https://github.com/cmouse/yahttp/pull/28

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/ext/yahttp/yahttp/reqresp.cpp
+++ b/ext/yahttp/yahttp/reqresp.cpp
@@ -1,5 +1,7 @@
 #include "yahttp.hpp"
 
+#include <limits>
+
 namespace YaHTTP {
 
   template class AsyncLoader<Request>;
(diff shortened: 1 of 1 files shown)


ext/yahttp/yahttp/reqresp.cpp
```diff
@@ -177,6 +179,9 @@ namespace YaHTTP {
             throw ParseError("Unable to parse chunk size");
           }
           if (chunk_size == 0) { state = 3; break; } // last chunk
+          if (chunk_size > (std::numeric_limits<decltype(chunk_size)>::max() - 2)) {
```
fredmorcos, line 182: `- 2` for a potential `\r\n`? Perhaps a comment would help here.

fredmorcos (approved):

PR #12893, opened 2023-06-07
rec: add feature to switch off unsupported DNSSEC algos

### LINKED ISSUE #12890: auth, rec: RSA-SHA1 broken on EL9
 - Issue type: Bug report
- Affects: auth 4.8, rec 4.9
- 
### Short description
Red Hat has decided that RSASHA1 validation (and signing?) is, by default, [forbidden in EL9, even for DNSSEC](https://bugzilla.redhat.com/show_bug.cgi?id=2073068).

This breaks auth if a user has algo 5/7 keys:
```
# pdnsutil test-algorithms
Testing algorithm 5(RSASHA1): 'OpenSSL RSA' ->'OpenSSL RSA' -> 'OpenSSL RSA' (2048 bits) OpenSSL RSA: Could not initialize context for signing: 50331800:digital envelope routines::invalid digest - crypto/evp/m_sigver.c:343 - do_sigver_init
Testing algorithm 7(RSASHA1-NSEC3-SHA1): 'OpenSSL RSA' ->'OpenSSL RSA' -> 'OpenSSL RSA' (2048 bits) OpenSSL RSA: Could not initialize context for signing: 50331800:digital envelope routines::invalid digest - crypto/evp/m_sigver.c:343 - do_sigver_init
```

This breaks recursor - `icann.org`, which currently is signed with algo 7, is declared BOGUS.

### Workaround
Running `update-crypto-policies --set DEFAULT:SHA1` (and restarting affected daemons) makes the problem go away. `LEGACY` also works, but likely allows more things that users may not want. (you may need to `dnf install crypto-policies-scripts`)

### Handling in other software
For BIND, `update-crypto-policies` generates a snippet (included from `named.conf`) that explicitly disables the affected algorithms, which changes the outcome from Bogus to Insecure.

Unbound handles the failure case explicitly: https://github.com/nlnetLabs/unbound/commit/6cfcf214516b93622602684be5c3036bc1e3df1e

cznic libdnssec (gnutls, not openssl): https://gitlab.nic.cz/knot/knot-dns/-/commit/b0c6f0709aeab#3b7edfb68fbd436c6b46dcbb2c7a8c0cc6d5fd92_264_266

### Fixing this

Some proposals:

#### configurable DNSSEC policy

if we add support for a feature like `disable-algorithms=5,7` to Recursor, users at least have a quick workaround to go insecure for such keys without changing the systemwide policy. Configuration of that feature could maybe be integrated into the systemwide policy like RHEL has done for BIND.

#### detect the situation/handle the error

Either during validation handle the specific error (like Unbound does), or do a check at startup and then automatically disable the algorithms.

#### document it

We could also decide to just document that everything will be broken until the user updates the systemwide policy so OpenSSL allows RSA-SHA1 validation.

omoerbeek:
### Short description

introduce a way to disable DNSSEC algos. Default is to determine automatically (for algos 5 and 7 only) . You can also set it manually by setting `"dnssec-disabled-algorithms` to a list of algorithm numbers.
Fixes #12890

On a Rocky Linux 9.2 system with default config:
`Jun 07 09:20:07 msg="Disabled DNSSEC algorithms" subsystem="config" level="0" prio="Notice" tid="0" ts="1686129607.324" algorithms="5 7"`
and [icann.org](http://icann.org/) is Insecure
And after setting the crypto policy to --set DEFAULT:SHA1 it does not print the log line and starts validating [icann.org](http://icann.org/)

You can get the status wit a new `rec_control command: list-dnssec-algos`. Example output:
`RSASHA1(disabled) RSASHA1-NSEC3-SHA1(disabled) RSASHA256 RSASHA512 ECDSAP256SHA256 ECDSAP384SHA384 ED25519 ED448`

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnssecinfra.cc
+++ b/pdns/dnssecinfra.cc
@@ -291,6 +291,103 @@ bool DNSCryptoKeyEngine::testOne(int algo)
   return ret;
 }
 
+static map<string, string> ICSStringToMap(const string& argStr)
+{
+  unsigned int algorithm = 0;
+  string sline;
+  string key;
+  string value;
+  string raw;
+  std::istringstream str(argStr);
+  map<string, string> stormap;
+
+  while(std::getline(str, sline)) {
+    std::tie(key,value)=splitField(sline, ':');
+    boost::trim(value);
+    if(pdns_iequals(key,"algorithm")) {
+      pdns::checked_stoi_into(algorithm, value);
+      stormap["algorithm"] = std::to_string(algorithm);
+      continue;
+    }
+    if (pdns_iequals(key,"pin")) {
+      stormap["pin"] = value;
+      continue;
+    }
+    if (pdns_iequals(key,"engine")) {
+      stormap["engine"] = value;
+      continue;
+    }
+    if (pdns_iequals(key,"slot")) {
+      int slot = std::stoi(value);
+      stormap["slot"]=std::to_string(slot);
+      continue;
+    }
+    if (pdns_iequals(key,"label")) {
+      stormap["label"] = value;
+      continue;
+    }
+    if(pdns_iequals(key, "Private-key-format")) {
+      continue;
+    }
+    raw.clear();
+    B64Decode(value, raw);
+    stormap[toLower(key)] = raw;
+  }
+  return stormap;
+}
+
+void DNSCryptoKeyEngine::testVerify(unsigned int algo, maker_t* verifier)
+{
+  const string message("Hi! How is life?");
+  const string pubkey5 = "AwEAAe2srzo8UfPx5WwoRXTRdo0H8U4iYW6qneronwKlRtXrpOqgZWPtYGVZl1Q7JXqbxxH9aVK5iK6aYOVfxbwwGHejaY0NraqrxL60F5FhHGHg+zox1en8kEX2TcQHxoZaiK1iUgPkMrHJlX5yI5+p2V4qap5VPQsR/WfeFVudNsBEF/XRvg0Exh65fPI/e8sYNgAiflzdN9/5RM644r6viBdieuwUNwEV2HPizCBMssYzx2F29CqNseToqCKQlj1tghuGAsiiSKeosfDLlRPDe/uxtij0wqe0FNybj1oL3OG8Lq3xp8yXIG4CF59xmRDKdnGDmVycKzUWkVOZpesCsUU=";
+  const string sig5 = "nMnMakbQiiCKIYsEiv4R75+8wvjQav2LPGIKucbqUZUz5sy1ovc2Pp7JVcOuyVyzQu5XH+CetDnTlqiEJWFHNU1jqEwwFK83GVOLABtvXSOvgmGwZGnHOouAchkrzgSSBoEh3+UUN3OsFZA21q6TZVRJBNBm7Ch/PxqSBkFS46ko/qLAUJ1p7/ymzwGNhuOfguHO3dAJ+LgcrNGLZQFDJ1aqT3kZ7LtXX2CQdd7EXgUs6VkE4Z3JN1RmPTk8kAJdZ4JLUR6lgu1dRlSPLGzqv+5d1yI7+h+B0LFNuDdQblDlBstO3LEs1KSaQld+TqVExpjj87oEg6wL/G/XOGabmQ==";
+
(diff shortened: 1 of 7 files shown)


pdns/dnssecinfra.cc
```diff
@@ -291,6 +316,103 @@ bool DNSCryptoKeyEngine::testOne(int algo)
   return ret;
 }
 
+static map<string, string> ICSStringToMap(const string& argStr)
```
Habbie, line 319: `ICS` should be `ISC`.

I know you did not introduce this duplication, but it looks like `DNSCryptoKeyEngine::makeFromISCString` might be reusable here. Not necessary for this PR.

pdns/dnssecinfra.cc
```diff
@@ -364,6 +456,15 @@ void DNSCryptoKeyEngine::testMakers(unsigned int algo, maker_t* creator, maker_t
...
   unsigned int udiffSign= dt.udiff()/100, udiffVerify;
 
+#if 0
+  if (algo == 7) {
+    auto pubkey = Base64Encode(dckeSign->getPublicKeyString());
+    auto sig = Base64Encode(signature);
+    cerr << "PubKey: " << pubkey << endl;
+    cerr << "Signature: " << sig << endl;
+  }
+#endif
```
Habbie, line 466: remove this?

pdns/recursordist/docs/settings.rst
```diff
@@ -533,6 +533,23 @@ Set the mode for DNSSEC processing, as detailed in :doc:`dnssec`.
...
+- Comma separated list of DNSSEC algorithm numbers
+- Default: (none)
+
+A list of DNSSEC algorithm numbers that should be considered disabled.
+These algorithms will not be used to validate DNSSEC signatures.
+Zones (only) signed with these algorithms will be considered ``Insecure``.
+
+If this setting is empty (the default), :program:`Recursor` will determine which algorithms to disable automatically.
+This is important on systems that have a default strict crypto policy, like RHEL9 derived systems.
+On such systems not disabling some algorithms (or changing the security policy) will make affected zones to be considered ``Bogus`` as using these algorithms fails.
```
Habbie, line 551: maybe mention which algorithms we do the check for

Habbie (changes requested):
just a few things

pdns/dnssecinfra.cc
```diff
@@ -291,6 +316,103 @@ bool DNSCryptoKeyEngine::testOne(int algo)
...
+  string pubkey;
+  string sig;
+  B64Decode(b64pubkey, pubkey);
+  B64Decode(b64sig, sig);
+  auto dckeVerify = verifier(algo);
+  dckeVerify->fromPublicKeyString(pubkey);
+
+  auto ret = dckeVerify->verify(message, sig);
+  if (!ret) {
+    throw runtime_error("Verification of verifier "+dckeVerify->getName() + " failed");
```
rgacogne, line 397: Would it make sense to return a boolean instead?

pdns/dnssecinfra.hh
```diff
@@ -189,6 +194,7 @@ class DNSCryptoKeyEngine
       static allmakers_t s_allmakers;
       return s_allmakers;
     }
+    static std::unordered_set<unsigned int> s_switchedOff;
```
rgacogne, line 197: Perhaps we could add a comment stating that this can only be altered at configuration time, when no other thread can touch it? Perhaps one day we could even automatically enforce that with a wrapper class that refuses to provide write access once the configuration phase is over?

rgacogne (approved):
A couple suggestions, but nothing that would prevent merging as it is.

pdns/dnssecinfra.cc
omoerbeek: done

pdns/dnssecinfra.hh
omoerbeek: done

pdns/recursordist/docs/settings.rst
```diff
@@ -546,7 +546,10 @@ A list of DNSSEC algorithm numbers that should be considered disabled.
 These algorithms will not be used to validate DNSSEC signatures.
 Zones (only) signed with these algorithms will be considered ``Insecure``.
 
-If this setting is empty (the default), :program:`Recursor` will determine which algorithms to disable automatically.
+If this setting is empty (the default), :program:`Recursor` will determine which algorithms to
```
Habbie, line 549: you accidentally introduced a newline

pdns/dnssecinfra.cc
```diff
@@ -393,21 +393,19 @@ void DNSCryptoKeyEngine::testVerify(unsigned int algo, maker_t* verifier)
...
 
 bool DNSCryptoKeyEngine::verifyOne(unsigned int algo)
 {
-  bool ret=true;
+  bool ret = false;
 
   for (auto* verifier : getAllMakers()[algo]) {
     try {
-      testVerify(algo, verifier);
+      ret = testVerify(algo, verifier);
```
rgacogne, line 405: We might want to `break` when `ret` is `false`.

omoerbeek:
I'll squash before merge

rgacogne (approved):
:+1:

Habbie (approved):

PR #12900, opened 2023-06-09
rec: Prevent duplicate C/DNAMEs being included when doing serve-stale

omoerbeek:
This can happen if the CNAME record itself was found, but its target not

Reported by @phonedph1  on IRC.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [X] documented the code

Needs test

### DIFF
--- a/pdns/recursordist/syncres.cc
+++ b/pdns/recursordist/syncres.cc
@@ -1733,7 +1733,7 @@ int SyncRes::doResolve(const DNSName& qname, const QType qtype, vector<DNSRecord
       if (child == qname) {
         LOG(prefix << qname << ": Step3 Going to do final resolve" << endl);
         res = doResolveNoQNameMinimization(qname, qtype, ret, depth, beenthere, context);
-        LOG(prefix << qname << ": Step3 Final resolve: " << RCode::to_s(res) << "/" << ret.size() << endl);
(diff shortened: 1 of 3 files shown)


pdns/recursordist/syncres.cc
```diff
@@ -1733,7 +1733,7 @@ int SyncRes::doResolve(const DNSName& qname, const QType qtype, vector<DNSRecord
       if (child == qname) {
         LOG(prefix << qname << ": Step3 Going to do final resolve" << endl);
         res = doResolveNoQNameMinimization(qname, qtype, ret, depth, beenthere, context);
-        LOG(prefix << qname << ": Step3 Final resolve: " << RCode::to_s(res) << "/" << ret.size() << endl);
+        LOG(prefix << qname << ": Step3 Final resolve: " << (res >= 0 ? RCode::to_s(res) : std::to_string(res)) << "/" << ret.size() << endl);
```
Habbie, line 1736: why this change?

Habbie (approved):
approved with one unimportant question

pdns/recursordist/syncres.cc
omoerbeek: SyncRes overloads the return value, it can be an rcode or negative for special cases.
RCode::to_s(res) prints `"ErrOutOfRange"` in that case.

Habbie: ok!

PR #12913, opened 2023-06-13
dns random: add method to get full 32-bits of randomness

omoerbeek:
### Short description

Currently `dns_random(uint32_t upper_bound)` has a drawbacks: the exclusive upper bound is mandatory. So the simple case: get full 32-bit randomness is not covered and the code to avoid modulo-bias is always executed. This is wasteful.

So add ``dns_random_uint32()`.

Start with using it in `pdns::dns_random_engine` and ` dns_random_uint16()`.

Also: reformat, cleanup, delint and refactor and *enable* tests.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dns_random.cc
+++ b/pdns/dns_random.cc
@@ -256,6 +256,7 @@ uint32_t dns_random_uint32()
       if (got != sizeof(num)) {
         throw std::runtime_error("getrandom() failed: " + stringerror());
       }
+      break;
     } while (true);
     return num;
 #else
@@ -289,6 +290,7 @@ uint32_t dns_random_uint32()
         attempts--;
         continue;
       }
+      break;
     } while (true);
     return num;
   }
--- a/pdns/dns_random.cc
+++ b/pdns/dns_random.cc
@@ -157,7 +157,7 @@ static void dns_random_setup(bool force = false)
 
 #if defined(HAVE_GETRANDOM)
   if (chosen_rng == RNG_GETRANDOM) {
-    char buf;
+    char buf = 0;
     // some systems define getrandom but it does not really work, e.g. because it's
     // not present in kernel.
     if (getrandom(&buf, sizeof(buf), 0) == -1 && errno != EINTR) {
--- a/pdns/dns_random.cc
+++ b/pdns/dns_random.cc
@@ -31,7 +31,6 @@
 #include "dns_random.hh"
 #include "arguments.hh"
 #include "logger.hh"
-#include "boost/lexical_cast.hpp"
 
 #if defined(HAVE_RANDOMBYTES_STIR)
 #include <sodium.h>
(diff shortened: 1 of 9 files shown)


pdns/dns_random.cc
```diff
@@ -201,114 +214,197 @@ void dns_random_init(const string& data __attribute__((unused)), bool force) {
...
+    uint32_t num = 0;
+    size_t attempts = 5;
+    ssize_t got = read(urandom_fd, &num, sizeof(num));
+    do {
+      if (got < 0) {
+        if (errno == EINTR) {
+          continue;
+        }
+
+        (void)close(urandom_fd);
```
rgacogne, line 282: This is not something you introduced in this pull request but I'm not fond of the behaviour when `read()` fails for a non `EINTR` `errno`: we are closing the static file descriptor but not setting it to `-1`, which means we will end up doing a use-after-free the next time this function is called, depending on how the exception is handled.

rgacogne: Actually this is worse than that: the file descriptor is used by all threads after the initial setup, which means that we can end up closing it from thread A while thread B is using it. I really don't think we should close this file descriptor.

omoerbeek: Thanks, I'll fix that soon.

I must say I do not like `dns_random.cc` very much, and not only because the awkward API until this PR. The `#ifdef` maze makes the code hard to read and I do really wonder if we really *need* all the variety in methods.

mnordhoff:
I assume you're not going through the entire codebase in this PR, but can I suggest two places to use the new function?

https://github.com/PowerDNS/pdns/blob/47097c4f403a3dd2ce4d8c131ecf91fd3e0d911c/pdns/opensslsigners.cc#L183

https://github.com/PowerDNS/pdns/blob/47097c4f403a3dd2ce4d8c131ecf91fd3e0d911c/pdns/tsigutils.cc#L54

pdns/dns_random.cc
zeha: > The #ifdef maze makes the code hard to read and I do really wonder if we really need all the variety in methods

I've never liked this either, and I think fewer options, best no options, make for tested, known-working, secure code.

If one implementation doesn't cut it, maybe it can be reduced to one per platform?

rgacogne: I like the idea very much! We should perhaps check the performance, although I'm not too worried. I'm curious how big the code would be when we need to provide our own implementation, though? The link you provided looks small enough, but I guess we need to pull in a ChaCha implementation as well?

omoerbeek: ChaCha is pretty fast, on OpenBSD/aarch64, the arc4random method is the quickest:
```
'Random test arc4random' 0.10 seconds: 7725808.0 runs/s, 0.13 us/run
'Random test sodium' 0.10 seconds: 6378117.4 runs/s, 0.16 us/run
'Random test openssl' 0.10 seconds: 5360387.9 runs/s, 0.19 us/run
'Random test urandom' 0.10 seconds: 181305.4 runs/s, 5.52 us/run
```
The linked OpenSSH portability library provides implementations of ChaCha both based on libcrypto and a standalone one.

Code size is small.

rgacogne: It's also true on my Linux (Arch):
```
'Random test getrandom' 0.11 seconds: 3926998.4 runs/s, 0.25 us/run
'Random test arc4random' 0.11 seconds: 3927979.4 runs/s, 0.25 us/run
'Random test sodium' 0.11 seconds: 3820573.9 runs/s, 0.26 us/run
'Random test openssl' 0.11 seconds: 2028661.7 runs/s, 0.49 us/run
'Random test urandom' 0.11 seconds: 2733101.3 runs/s, 0.37 us/run
```

pdns/dns_random.cc
```diff
@@ -98,99 +99,112 @@ static void dns_random_setup(bool force=false)
 #else
   rng = ::arg()["rng"];
   rdev = ::arg()["entropy-source"];
-  if (rng == "auto") {
-# if defined(HAVE_GETRANDOM)
+  if (rng == "auto") { // NOLINT: I see no way to avoid repeating blocks reported by clang-tidy
```
jsoref, line 102: Would it complain about:
```c
#ifdef FOO_1
#define FOO FOO_1
#elif defined FOO_2
#define FOO FOO_2
...
#endif
x = FOO;
```

pdns/dns_random.cc
omoerbeek: > one thing to consider: if you import the compat code from openssh-portable, you'll have to keep it updated as well. I -think- there is no library version of it?

True, though I don't think the burden is very high.

PR #12923, opened 2023-06-16
dnsdist: Automatically load Lua FFI inspection functions

rgacogne:
### Short description

We used to do that for all Lua FFI functions except the ones used for inspecting StatNode objects.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


pdns/dnsdistdist/dnsdist-lua-inspection-ffi.h
```diff
@@ -0,0 +1,46 @@
+/*
```
omoerbeek, line 1: Maybe an include guard?

omoerbeek (approved):
One nit.

pdns/dnsdistdist/dnsdist-lua-inspection-ffi.h
rgacogne: The reason I have not been doing that is that LuaJIT documentation states:
> C declarations are not passed through a C pre-processor, yet. No pre-processor tokens are allowed, except for #pragma pack. Replace #define in existing C header files with enum, static const or typedef and/or pass the files through an external C pre-processor (once). Be careful not to include unneeded or redundant declarations from unrelated header files. 

So I'm not sure it would work, and if it does that it would work with all versions of LuaJIT.

PR #12931, opened 2023-06-20
Use arc4random only for random values

omoerbeek:
### Short description

As discussed in #12913 

This starts using a single method to generate random numbers, taken from the arc4random() code from openssh-portable. This is a ChaCha bases stream computed in userland, which gets stirred on startup and once in a while by using either `getrandom()`, `getentropy()` or reading from `/dev/urandom`.

There is one place where `random()` is still used: the fallback code of dnsdist-random.cc. I do think that code could also move to `dns_random_uint32()` (which is just a wrapper to `arc4random()`.

We do get to maintain a few extra source files, but I do not think the burden is very big. We do get to drop `dns_random.cc` and `dns_random_urandom.cc`.

So fat tested by me on macOS, OpenBSD, Debian bullseye.

### Checklist

I have:
- [X] compiled this code
- [ ] tested this code
- [ ] documented the code


omoerbeek:
Marking thsi as draft, there are some dependency issues wrt testing and fuzzing that I cannot solve right now.

omoerbeek:
So this is how the fuzzing target knows it must build yahttp: https://github.com/google/oss-fuzz/blob/cd86bd01df1710aed5fe303bb7ddea7c7e37a226/projects/powerdns/build.sh#LL37C1-L37C22

For this PR we need to add a line for arc4random as well. This took me a while to realise....

rgacogne:
Yes, this is terrible. You also have to sign the developer agreement or what not, feel free to send me a patch and I'll take care of it.

omoerbeek:
> Yes, this is terrible. You also have to sign the developer agreement or what not, feel free to send me a patch and I'll take care of it.

Thanks, it is a bit too early for that I suppose...

rgacogne (approved):
Looks good to me! Are you still planning some changes?

omoerbeek:
> Looks good to me! Are you still planning some changes?

I have to go through it one more time and then it can be merged. Once this is in, we have to fix the fuzzing build.

PR #12943, opened 2023-06-22
dnsdist: Apply performance suggestions from Coverity

rgacogne:
### Short description

I don't expect these changes to actually make a difference in practice, because pretty much all of them impact configuration processing, where performance is not that critical. Still they should not hurt and a few of them (mplexer, OpenSSLTLSConnection, MOADNSParser) might actually improve some edge cases.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdist-dynblocks.hh
+++ b/pdns/dnsdist-dynblocks.hh
@@ -264,13 +264,13 @@ public:
   void setSuffixMatchRule(unsigned int seconds, const std::string& reason, unsigned int blockDuration, DNSAction::Action action, smtVisitor_t visitor)
   {
     d_suffixMatchRule = DynBlockRule(reason, blockDuration, 0, 0, seconds, action);
-    d_smtVisitor = visitor;
+    d_smtVisitor = std::move(visitor);
   }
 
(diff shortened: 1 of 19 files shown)


fredmorcos (approved):
It's fine by me but I never expected `std::move` to offer performance improvements...
Is it perhaps allowing the compiler to assume that the source object's space can be reused? Or is it more about explicitly avoiding copy construction? (The latter would be a big question mark from me)

rgacogne:
It's complicated. `std::move` offers huge performance improvements when used in combination with move constructors and move assignment operator, of course, by allowing the resources held by an object to be reused instead of duplicated. In the specific case of the "issues" reported by Coverity in this PR, pretty all of these are indeed about explicitly avoiding copy construction, and I would be really surprised if the compiler was not smart enough to do it already. Other tools like `cppcheck` also report the same kind of "optimization", so I guess it might be, or used to be, a thing in some cases.

omoerbeek (approved):
OK, though when adding future changes, we need to be extra careful to not use a moved object in new code. I do not know if compilers (or static analysers) are always smart enough to detect that use.

PR #12945, opened 2023-06-22
Fix issue with clang-tidy and module source files

fredmorcos:
### Short description
The problem was that clang-tidy-diff needs to truncate paths for clang-tidy to find the files relative to the current directory and where the compilation database is, so files under directories like `ext` and `modules` didn't end up being found because they were expected to be under `pdns`. What this change does is hacky, but it's all I can think of for now (short of a rework and/or changing the project structure).

### Checklist
I have:
- [X] compiled this code
- [X] tested this code
- [X] documented the code

### DIFF
--- a/.github/scripts/git-filter.py
+++ b/.github/scripts/git-filter.py
@@ -19,6 +19,8 @@ def main():
     compdb = helpers.load_compdb("compile_commands.json")
     compdb = helpers.index_compdb(compdb)
 
+    pdns_path = os.path.join("pdns", "")
+    cwd = os.getcwd()
     root = helpers.get_repo_root()
 
     diff = sys.stdin.read()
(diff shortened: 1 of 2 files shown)


.github/scripts/git-filter.py
```diff
@@ -19,14 +19,28 @@ def main():
...
     diff = sys.stdin.read()
     patch_set = unidiff.PatchSet(diff)
     for patch in patch_set:
         path = os.path.join(root, patch.path)
         if path in compdb:
+            if not patch.path.startswith(pdns_path):
+                # If the file being diffed is not under the pdns/ directory, we
+                # need to reconstruct its filename in the patch adding extra
+                # paths that clang-tidy-diff will get rid of: this way
+                # clang-tidy can work with the correct file path.
```
rgacogne, line 35: It might be nice to have a small comment with an example of the content of `patch.path` and the subsequent values of `path` and `relpath`. I _think_ I get it but I'm not 100% sure.

fredmorcos: Done!

rgacogne: Thanks a lot! :)

rgacogne (approved):

PR #12948, opened 2023-06-23
dnsdist: Implement the AsynchronousHolder with the new channel feature

rgacogne:
### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdistdist/dnsdist-async.cc
+++ b/pdns/dnsdistdist/dnsdist-async.cc
@@ -31,7 +31,9 @@ AsynchronousHolder::Data::Data(bool failOpen) :
   d_failOpen(failOpen)
 {
   auto [notifier, waiter] = pdns::channel::createNotificationQueue(true);
+  // NOLINTNEXTLINE(cppcoreguidelines-prefer-member-initializer): how I am supposed to do that?
   d_waiter = std::move(waiter);
+  // NOLINTNEXTLINE(cppcoreguidelines-prefer-member-initializer): how I am supposed to do that?
   d_notifier = std::move(notifier);
 }
 
@@ -60,7 +62,7 @@ bool AsynchronousHolder::wait(AsynchronousHolder::Data& data, FDMultiplexer& mpl
 {
   readyFDs.clear();
   mplexer.getAvailableFDs(readyFDs, atMostMs);
-  if (readyFDs.size() == 0) {
+  if (readyFDs.empty()) {
     /* timeout */
     return true;
   }
@@ -79,10 +81,13 @@ void AsynchronousHolder::stop()
   notify();
 }
 
+// NOLINTNEXTLINE(performance-unnecessary-value-param): this is a long-lived thread, and we want to make sure the reference count of the shared pointer has been increased
 void AsynchronousHolder::mainThread(std::shared_ptr<Data> data)
 {
   setThreadName("dnsdist/async");
-  struct timeval now;
+  struct timeval now
+  {
+  };
   std::list<std::pair<uint16_t, std::unique_ptr<CrossProtocolQuery>>> expiredEvents;
 
   auto mplexer = std::unique_ptr<FDMultiplexer>(FDMultiplexer::getMultiplexerSilent(1));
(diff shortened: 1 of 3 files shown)


pdns/dnsdistdist/dnsdist-async.cc
```diff
@@ -31,7 +31,9 @@ AsynchronousHolder::Data::Data(bool failOpen) :
   d_failOpen(failOpen)
 {
   auto [notifier, waiter] = pdns::channel::createNotificationQueue(true);
+  // NOLINTNEXTLINE(cppcoreguidelines-prefer-member-initializer): how I am supposed to do that?
   d_waiter = std::move(waiter);
+  // NOLINTNEXTLINE(cppcoreguidelines-prefer-member-initializer): how I am supposed to do that?
   d_notifier = std::move(notifier);
```
fredmorcos: I missed those. What is the problem here? i.e. what is clang-tidy complaining about?

pdns/dnsdistdist/dnsdist-async.cc
```diff
@@ -79,10 +81,13 @@ void AsynchronousHolder::stop()
   notify();
 }
 
+// NOLINTNEXTLINE(performance-unnecessary-value-param): this is a long-lived thread, and we want to make sure the reference count of the shared pointer has been increased
```
fredmorcos, line 84: I find it strange that clang-tidy complains about passing a shared pointer object by value...

rgacogne: That's because to do that you need to increment the reference counter, which is not cheap, while taking a reference to the same `shared_pointer` is free.

pdns/dnsdistdist/dnsdist-async.cc
fredmorcos: So I take it `std::move` is not usable in an initializer list?

rgacogne: `clang-tidy` tells us to initialize `d_waiter` and `d_notifier` in the member initializer list, which as far as I know it's not possible.

rgacogne: `std::move` is, turning a pair into two member variables, I don't think so?

pdns/dnsdistdist/dnsdist-async.cc
fredmorcos: Sure. But that's the whole point of using a `shared_ptr`. When I pass it by value I really mean it, so I find it stupid that `clang-tidy` is complaining about this and suggesting a reference instead.

pdns/dnsdistdist/dnsdist-async.cc
fredmorcos: > `std::move` is, turning a pair into two member variables, I don't think so?

Right, I would also guess it's not possible.

pdns/dnsdistdist/dnsdist-async.cc
rgacogne: I'm a bit torn, to be honest, because in my experience passing a `shared_pointer` by value is actually very often an oversight. On the other hand it's annoying to have to justify yourself every time you do so on purpose.

fredmorcos: Is the ref counter incremented by passing a `shared_ptr` as a reference?

rgacogne: No, it's not, you get the same `shared_ptr`, and not a different `shared_ptr` pointing to the same data and metadata.

fredmorcos: Well in that case I would disagree that passing it by value would be an oversight. If a reference needs to be passed then it needs to be a reference to the internal object rather than the shared pointer itself, this way it's clear that the caller has to outlive the callee.

rgacogne: For maximum safety, yes, for maximum performance no :)

fredmorcos: I think getting a reference to the internal object from the shared pointer object is very cheap... no?

rgacogne: You are right, of course! I read your previous comment too quickly, sorry! I agree it's better to pass `const T&` than `const std::shared_ptr<const T>&`, clearly!

PR #12963, opened 2023-06-29
rec: fix qname length getting out-of-sync with qname-minimization iteration count (approach 2)

### LINKED ISSUE #12956: rec: QM sometimes runs out of iterations before it's done
- Program: recursor
 - Issue type: Bug report

### Short description
Qnames with a long amount of labels but missing or expired glue at a much lower level seems to exhaust QM. This appears it could be a regression due to the step4 skip feature.

### Environment
 - Operating system: Debian
 - Software version:  4.9.0-rc1 & master
 - Software source:  compiled

### Steps to reproduce
```
./pdns_recursor --local-port=5353 --socket-dir=/var/tmp --trace --dnssec=off
dig ns rocket-cdn.com. @127.0.0.1 -p 5353
./rec_control --socket-dir=/var/tmp wipe-cache ns3.rocket-cdn.com
./rec_control --socket-dir=/var/tmp wipe-cache ns2.rocket-cdn.com
./rec_control --socket-dir=/var/tmp wipe-cache ns1.rocket-cdn.com
```

`dig pull-flv-f10-gcp01.tiktokcdn.com.rocket-cdn.com. @127.0.0.1 -p 5353` will fail

### Expected behaviour
Should hopefully work

### Actual behaviour
`QM pull-flv-f10-gcp01.tiktokcdn.com.rocket-cdn.com: Max iterations reached, return ServFail`

Tace log: https://paste.pfak.org/uOfgme

### Other information
```
23:42 < dwfreed> qmStepLen is only adding 1 label every time, but i becomes desynced with child label count
23:43 < dwfreed> because step1 keeps bringing child back to 'com', but i is incremented in every step4 skip
23:44 < dwfreed> in the inner loop, I think if you init i to child label count, you'll thoroughly bandaid it
23:47 < dwfreed> but yeah, it'll eventually happen whenever step1 brings child back to before its value when step4 hits a              delegation
23:48 < dwfreed> but it needs a label count high enough and a step1 regression far enough back that the sequence 0 + 1 + 2 + 3 +  ... will exceed qnamelen
```

Indeed a `i--;` in the step 4 skip part makes it work (or + 1 in the loop condition), but I am unsure if that breaks anything else.

omoerbeek:
Approach two: fall back to non-QM mode if loop detected Fixes #12956

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/recursordist/syncres.cc
+++ b/pdns/recursordist/syncres.cc
@@ -1599,7 +1599,12 @@ static unsigned int qmStepLen(unsigned int labels, unsigned int qnamelen, unsign
   return targetlen;
 }
 
-int SyncRes::doResolve(const DNSName& qname, const QType qtype, vector<DNSRecord>& ret, unsigned int depth, set<GetBestNSAnswer>& beenthere, Context& context)
+static string resToString(int res)
+{
+  return res >= 0 ? RCode::to_s(res) : std::to_string(res);
+}
+
+int SyncRes::doResolve(const DNSName& qname, const QType qtype, vector<DNSRecord>& ret, unsigned int depth, set<GetBestNSAnswer>& beenthere, Context& context) // NOLINT(readability-function-cognitive-complexity)
 {
   auto prefix = getPrefix(depth);
   auto luaconfsLocal = g_luaconfs.getLocal();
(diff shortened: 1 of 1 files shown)


dwfreed:
I think this is the better approach, because I think that the other approach is really going to boil down to doing this anyway, and this gets it done a lot faster.  Step 4's stop on delegation would never result in step 1 producing a new name with more labels, so it'll just loop until eventually the name to query is the original name.

Habbie:
what dwfreed said!

Habbie (approved):

PR #12979, opened 2023-07-03
builder-dispatch: Explicitly grant permissions to the build package workflow

rgacogne:
### Short description

So that we can use it from forks:
> For reusable workflows outside your enterprise or organization, the permissions setting for id-token should be explicitly set to write at the caller workflow level or in the specific job that calls the reusable workflow.

### Checklist

I have:
- [ ] compiled this code
- [x] tested this code
- [ ] documented the code


Habbie:
I'm currently running a build with this fix, and I can confirm it fixes the permission error.

PR #12985, opened 2023-07-04
auth lmdb: add backend commands for checking & refreshing indexes

Habbie:
### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/modules/lmdbbackend/lmdbbackend.cc
+++ b/modules/lmdbbackend/lmdbbackend.cc
@@ -2620,17 +2620,107 @@ bool LMDBBackend::getTSIGKeys(std::vector<struct TSIGKey>& keys)
 
 string LMDBBackend::directBackendCmd(const string& query)
 {
-  if (query == "info") {
-    ostringstream ret;
+  ostringstream ret, usage;
 
+  usage << "info                               show some information about the database" << endl;
+  usage << "index check domains                check zone<>ID indexes" << endl;
+  usage << "index refresh domains <ID>         refresh index for zone with this ID" << endl;
+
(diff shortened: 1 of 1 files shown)


modules/lmdbbackend/lmdbbackend.cc
```diff
@@ -2620,17 +2620,100 @@ bool LMDBBackend::getTSIGKeys(std::vector<struct TSIGKey>& keys)
...
+      // index refresh domains 12345
+      if (argv.size() < 4) {
+        return "usage: index refresh domains <ID>\n";
+      }
+
+      if (argv[2] != "domains") {
+        return "can only refresh in the domains index\n";
+      }
+
+      uint32_t id = atoi(argv[3].c_str());
```
rgacogne, line 2704:
```suggestion
      auto id = checked_stoi<uint32_t>(argv[3]);
```

rgacogne (approved):

modules/lmdbbackend/lmdbbackend.cc
Habbie: I didn't do that because it says `stoi`, but you're saying it should just work?

rgacogne: Yes, it should convert the string to the corresponding `uint32_t` value, or throw an exception if the string cannot be parsed as an unsigned integer value. Unless I'm missing something and you need a different behaviour?

Habbie: that sounds perfect. Thanks!

PR #12989, opened 2023-07-05
auth, rec: update option text and docs to clarify logging is to stderr, not stdout

Habbie:
~~PRing now so I don't forget, but this is not done:~~ done:

* [x] update docs
* [x] check what dnsdist does

### DIFF
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -601,7 +601,7 @@ regression testing.
 -  Boolean
 -  Default: no
 
-Do not log to syslog, only to stdout. Use this setting when running
+Do not log to syslog, only to stderr. Use this setting when running
 inside a supervisor that handles logging (like systemd).
 
 .. warning::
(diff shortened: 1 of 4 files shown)


Habbie:
> check what dnsdist does

dnsdist logs to stdout and says this correctly in the help text for `--disable-syslog` and `--log-timestamps`

pdns/dolog.hh
```diff
@@ -231,7 +231,7 @@ void verboselog(const char* formatStr, const Args&... args)
 #endif /* DNSDIST */
     genlog(std::cout, LOG_DEBUG, false, formatStr, args...);
 #ifdef DNSDIST
-  }
+}
```
jsoref, line 234: This change appears errant 😦 
```suggestion
  }
```

omoerbeek (approved):

PR #13001, opened 2023-07-07
Fix build part2: include chacha_private.h and explicit_bzero.c

omoerbeek:
### Short description

The latter is needed on e.g. el-7

Followup to #12999 

This also fixes the redundant declaration warnings

### Checklist

I have:
- [X] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -154,7 +154,7 @@ dnl Checks for library functions.
 dnl the *_r functions are in posix so we can use them unconditionally, but the ext/yahttp code is
 dnl using the defines.
 AC_CHECK_FUNCS_ONCE([strcasestr localtime_r gmtime_r recvmmsg sched_setscheduler])
-AC_CHECK_FUNCS_ONCE([getrandom getentropy arc4random arc4random_uniform arc4random_buf])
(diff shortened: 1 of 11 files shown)


configure.ac
```diff
@@ -154,7 +154,7 @@ dnl Checks for library functions.
 dnl the *_r functions are in posix so we can use them unconditionally, but the ext/yahttp code is
 dnl using the defines.
 AC_CHECK_FUNCS_ONCE([strcasestr localtime_r gmtime_r recvmmsg sched_setscheduler])
-AC_CHECK_FUNCS_ONCE([getrandom getentropy arc4random arc4random_uniform arc4random_buf])
```
rgacogne, line 157: May I suggest reusing `PDNS_CHECK_SECURE_MEMSET`, which is defined in `m4/pdns_check_secure_memset.m4` and used by at least dnsdist? It already checks `explicit_bzero` and `explicit_memset` (which seems forgotten here?) but not yet `memset_s`

rgacogne:
I think you might be missing a symbolic link for `pdns_check_secure_memset.m4` in `recursordist/m4/`?

rgacogne (approved):

Issue #12846, opened 2023-05-19
dnsdist: metrics to count failures for health-check domain for the backend server


SrX:
- Program: dnsdist
 - Issue type: Feature request

### Short description
Shows now many failures for health-check domain happed for each of the backed

### Usecase
Be able to monitor if domains used to perform a health-check in a backend reported as failed. That should help to detect network loss or health-check target eventual failure before the backed is marked as down.

### Description
Get a metrics that shows now many failures for health-check domain happed for each of the backend server.

rgacogne:
We need to be careful that the set of domains used for the health-checks would then needs to be bounded, otherwise we have an issue, for example if someone uses a random prefix for each health-check query.

SrX:
I would base the metrics on healcheck per backed, no adding the domain to prometheus metric. Once failures where detected by health check, we do print to logs the domain name.

PR #13009, opened 2023-07-10
dnsdist: Add metrics for health-check failures

rgacogne:
### Short description

Closes https://github.com/PowerDNS/pdns/issues/12846.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdistdist/dnsdist-healthchecks.cc
+++ b/pdns/dnsdistdist/dnsdist-healthchecks.cc
@@ -60,6 +60,7 @@ static bool handleResponse(std::shared_ptr<HealthCheckData>& data)
   auto& ds = data->d_ds;
   try {
     if (data->d_buffer.size() < sizeof(dnsheader)) {
+      ++data->d_ds->d_healthCheckMetrics.d_parseErrors;
       if (g_verboseHealthChecks) {
         infolog("Invalid health check response of size %d from backend %s, expecting at least %d", data->d_buffer.size(), ds->getNameWithAddr(), sizeof(dnsheader));
       }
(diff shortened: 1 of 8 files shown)


pdns/dnsdistdist/dnsdist-healthchecks.cc
```diff
@@ -151,6 +158,7 @@ class HealthCheckQuerySender : public TCPQuerySender
 
   void notifyIOError(InternalQueryState&& query, const struct timeval& now) override
   {
+    ++d_data->d_ds->d_healthCheckMetrics.d_networkErrors;
```
omoerbeek, line 161: Several cases of `d_networkErrors` are lumped together. Does it make sense to separate those?

pdns/dnsdistdist/dnsdist-healthchecks.cc
```diff
@@ -471,6 +483,7 @@ void handleQueuedHealthChecks(FDMultiplexer& mplexer, bool initial)
           infolog("Timeout while waiting for the health check response (ID %d) from backend %s", data->d_queryID, data->d_ds->getNameWithAddr());
         }
 
+        ++data->d_ds->d_healthCheckMetrics.d_timeOuts;
         data->d_ds->submitHealthCheckResult(initial, false);
       }
       catch (const std::exception& e) {
```
omoerbeek, line 489: No increment of any counter here, some a few other cases in `handleQueuedHealthChecks()`. I'm not  sure if these cases are already handled by the callback functions or not. At least a comment would be nice.

omoerbeek (commented):
I have two questions.

pdns/dnsdistdist/dnsdist-healthchecks.cc
rgacogne: I don't think so, as my current plan is to get a better understanding of the generic kind of failures. If we want more details we can turn the "verbose healthchecks" option to get detailed errors in the logs.

pdns/dnsdistdist/dnsdist-healthchecks.cc
rgacogne: I added a couple comments

omoerbeek (approved):

PR #13011, opened 2023-07-10
Rel/rec 4.9.x The reverse-solution request of dns64 records needs to add TTL to v4 cname

zjs604381586:
### Short description

For the reverse resolution of **dns64** records, the TTL of the v4 record after the cname is 0. Since the authority has not configured the reverse resolution of v4, it will cause some irrelevant resolution requests to reach the pdns-recursor service. You can set the v4 cname in **getFakePTRRecords()** Add ttl to improve cache hit rate
### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code


omoerbeek:
I'd rather not introduce yet another setting. Maybe we could borrow the TTL form the PTR records that's get added... have to think about it.

omoerbeek:
> I'd rather not introduce yet another setting. Maybe we could borrow the TTL form the PTR records that's get added... have to think about it.

#13043 shows the alternative approach.

zjs604381586:
> > I'd rather not introduce yet another setting. Maybe we could borrow the TTL form the PTR records that's get added... have to think about it.
> 
> #13043 shows the alternative approach.

Great, this solution looks simpler and more reasonable

omoerbeek:
#13043 was merged.

PR #13012, opened 2023-07-10
dnsdist: Better handling of spurious wake-ups, interrupted calls

rgacogne:
### Short description

If we were woken up by the multiplexer but have actually nothing to read, or the call to `recv()` is interrupted, we do not want to stop listening for the health-check response event. It is also not useful to log about it, even at "verbose health-check" level.
Note that we would have logged previously, so this kind of event would not have gone unnoticed anyway.

### Checklist

I have:
- [x] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/.not-formatted
+++ b/.not-formatted
@@ -74,8 +74,6 @@
 ./pdns/dnsdistdist/connection-management.hh
 ./pdns/dnsdistdist/dnsdist-backend.cc
 ./pdns/dnsdistdist/dnsdist-dynblocks.cc
-./pdns/dnsdistdist/dnsdist-healthchecks.cc
-./pdns/dnsdistdist/dnsdist-healthchecks.hh
 ./pdns/dnsdistdist/dnsdist-kvs.cc
 ./pdns/dnsdistdist/dnsdist-kvs.hh
 ./pdns/dnsdistdist/dnsdist-lbpolicies.cc
--- a/pdns/dnsdistdist/dnsdist-healthchecks.cc
+++ b/pdns/dnsdistdist/dnsdist-healthchecks.cc
@@ -190,11 +190,11 @@ static void healthCheckUDPCallback(int descriptor, FDMultiplexer::funcparam_t& p
     data->d_buffer.resize(512);
 
     // NOLINTNEXTLINE(cppcoreguidelines-pro-type-reinterpret-cast)
-    got = recvfrom(data->d_udpSocket.getHandle(), &data->d_buffer.at(0), data->d_buffer.size(), 0, reinterpret_cast<sockaddr*>(&from), &fromlen);
+    got = recvfrom(data->d_udpSocket.getHandle(), data->d_buffer.data(), data->d_buffer.size(), 0, reinterpret_cast<sockaddr*>(&from), &fromlen);
     if (got < 0) {
       int savederrno = errno;
       if (savederrno == EINTR) {
-        /* interrupted, let's try again */
+        /* interrupted before any data was available, let's try again */
         continue;
       }
       if (savederrno == EWOULDBLOCK || savederrno == EAGAIN) {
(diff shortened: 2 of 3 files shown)


rgacogne:

rgacogne:
I don't understand the `narrowing conversion from 'unsigned int' to signed type 'int' is implementation-defined (bugprone-narrowing-conversions - Level=Warning)` clang-tidy warnings on this PR, and my local version of clang-tidy does not generate them.. I _think_ `clang-tidy` is drunk on that one..

omoerbeek:
It is not value preserving indeed, as I mentioned, but in a two's-complement world it's no implementation defined, I believe.

kpfleming:
Well, in a pedantic sense it is, but all implementations available define it the same way... and the C/C++ standards have moved to declaring two's-complement as the only supported implementation too IIRC.

rgacogne:
But have you looked at the actual location of the warnings? Both sides seem `unsigned` to me..

omoerbeek:
bitfields, they are special.... When reading the C99 (6.3.1.1 clause 2):
```
2 The following may be used in an expression wherever an int or unsigned int may be used:
— An object or expression with an integer type whose integer conversion rank is less than the rank of int and unsigned int.
— A bit-field of type _Bool, int, signed int, or unsigned int.
If an int can represent all values of the original type, the value is converted to an int; otherwise, it is converted to an unsigned int. These are called the integer promotions.45) All other types are unchanged by the integer promotions.
```
So it looks like there is an intermediate promotion to `int`, even if the field itself is `unsigned int: size`.

rgacogne:
Ouch. I wonder once more why we could not replace these `unsigned :16` by a `uint16_t`..

omoerbeek:
> 



> Ouch. I wonder once more why we could not replace these `unsigned :16` by a `uint16_t`..

Looks like all `unsigned :16` are on a 16 bit boundary, so that might work indeed.

rgacogne:
Rebased on top of master now that #13026 has been merged.

.not-formatted
```diff
@@ -74,8 +74,6 @@
 ./pdns/dnsdistdist/connection-management.hh
 ./pdns/dnsdistdist/dnsdist-backend.cc
 ./pdns/dnsdistdist/dnsdist-dynblocks.cc
-./pdns/dnsdistdist/dnsdist-healthchecks.cc
-./pdns/dnsdistdist/dnsdist-healthchecks.hh
```
fredmorcos, line 78: we're chipping through little by little 😃

pdns/dnsdistdist/dnsdist-healthchecks.cc
```diff
@@ -169,22 +169,37 @@ class HealthCheckQuerySender : public TCPQuerySender
...
+  do {
+    from.sin4.sin_family = data->d_ds->d_config.remote.sin4.sin_family;
+    auto fromlen = from.getSocklen();
+    data->d_buffer.resize(512);
+
+    got = recvfrom(data->d_udpSocket.getHandle(), &data->d_buffer.at(0), data->d_buffer.size(), 0, reinterpret_cast<sockaddr *>(&from), &fromlen);
+    if (got < 0) {
+      int savederrno = errno;
+      if (savederrno == EINTR) {
+        /* interrupted, let's try again */
```
fredmorcos, line 197: I know it's mentioned on the manual page, but perhaps saying the following would improve readability (I had to go check the manual page to make sure):

> interrupted before any data was available

fredmorcos (approved):

pdns/dnsdistdist/dnsdist-healthchecks.cc
```diff
@@ -221,7 +249,7 @@ static void healthCheckTCPCallback(int fd, FDMultiplexer::funcparam_t& param)
       ioState = data->d_tcpHandler->tryRead(data->d_buffer, data->d_bufferPos, data->d_buffer.size());
       if (ioState == IOState::Done) {
         data->d_bufferPos = 0;
-        uint16_t responseSize;
+        uint16_t responseSize{0};
         memcpy(&responseSize, &data->d_buffer.at(0), sizeof(responseSize));
```
omoerbeek, line 253:
```suggestion
        memcpy(&responseSize, data->d_buffer.data(), sizeof(responseSize));
```

omoerbeek (approved):
A few nits.

pdns/dnsdistdist/dnsdist-healthchecks.cc
rgacogne: Will do!

PR #13013, opened 2023-07-11
dnsdist: Add Lua bindings to access selector and action

### LINKED ISSUE #13007: dnsdist: Ponder adding a better accessor for rules
- Program: dnsdist 
 - Issue type: Feature request

### Short description
Currently, if one wants to get the number of matches of a given rule from Lua, the only option is to use `getAction(<rule number>):getStats()['']` which requires finding the position of the rule first.
It would be better if:
- `addAction` returned a reference to the created rule
- `getAction` accepted a UUID and/or a rule name

A bit unrelated but `getTopRules` and friends seems to return a `vector` instead of a Lua array, which makes them useless.

rgacogne:
### Short description

This PR adds Lua binding to make it possible to retrieve the selector and action of a rule, and to get the number of match of a given selector:
```
> addAction(AllRule(), AllowAction(), {name='myname', uuid='78589624-ebd0-44a5-994c-448ece5e188c'})
> getRule('78589624-ebd0-44a5-994c-448ece5e188c'):getSelector():getMatches()
0
> getRule('myname'):getSelector():getMatches()
0
> getRule(0):getSelector():getMatches()
0
```

Closes #13007.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


pdns/dnsdist-lua-rules.cc
```diff
@@ -224,12 +224,65 @@ static std::vector<T> getTopRules(const std::vector<T>& rules, unsigned int top)
...
+        }
+      }
+    }
+    catch (const std::exception& e) {
+      /* a name, then */
+      for (const auto& rule : rules) {
+        if (rule.d_name == *str) {
+          return rule;
+        }
+        }
```
omoerbeek, line 261:
```suggestion
      }
```

pdns/dnsdist-lua-rules.cc
```diff
@@ -224,12 +224,65 @@ static std::vector<T> getTopRules(const std::vector<T>& rules, unsigned int top)
...
+    }
+  }
+  else if (auto pos = boost::get<int>(&selector)) {
```
omoerbeek, line 264: might be out-of-bounds?

pdns/dnsdistdist/docs/rules-actions.rst
```diff
@@ -169,12 +169,62 @@ For Rules related to the incoming query:
...
+
+  .. versionadded:: 1.9.0
+
+  Return the rule corresponding to the selector, if any.
+  The selector can be the position of the rule in the list, as an integer,
+  its name as a string or its UUID as a string as well.
+
+  :param int or str selector: The position in the list, name or UUID of the rule to return.
+
+.. function:: getSelfAnsweredResponseRule(selector) -> DNSDistResponseRuleAction
```
omoerbeek, line 218: This name is a bit confusing to me. "SelfAnswered" can be more than just cache-hits, right? i.e. answers synthesized by dnsdist itself. Do they count or not?

rgacogne: I agree the name is not great, but it is already used by several other functions (`addSelfAnsweredResponseAction`, `mvSelfAnsweredResponseRule`, `mvSelfAnsweredResponseRuleToTop`, `mvSelfAnsweredResponseRuleToTop`, `showSelfAnsweredResponseRules`, ...). The rules are applied to all responses that comes from dnsdist and not from a backend, so cache hits and answers synthesized from rules do indeed count.

pdns/dnsdist-lua-rules.cc
rgacogne: Yes, and then I think `at()` throwing an exception in that case is fine?

omoerbeek: A comment indicating that would be nice.

pdns/dnsdistdist/docs/rules-actions.rst
omoerbeek: OK, then the explanation of the count should mention that it's more than just cache-hits.

pdns/dnsdist-lua-rules.cc
rgacogne: Done

pdns/dnsdistdist/docs/rules-actions.rst
rgacogne: Oh, I see now that there was a copy/paste mistake, fixed!

omoerbeek (approved):

Issue #13015, opened 2023-07-12
pdns-tools: calidns with ECS-range of 0.0.0.0/0 doesn't work as expected


aj-gh:
- Program: calidns
 - Issue type: Bug report

### Short description
Using an ECS range of 0.0.0.0/0 really seems to use an ECS of 0.0.0.0/0 instead of the expected randomization; using 0.0.0.0/1 works but obviously isn't /0 :wink: 

### Environment

 - Operating system: Ubuntu 20.04 LTS
 - Software version: pdns-tools 4.8.1-1pdns.focal
 - Software source: PowerDNS repository

### Steps to reproduce
1. Specify `--ecs 0.0.0.0/0`
2. Check traffic capture
3. Specify `--ecs 0.0.0.0/1`
4. Check traffic capture

### Expected behaviour
Randomized ECS on each query.

### Actual behaviour
0.0.0.0/0 ECS when specifying 0.0.0.0/0, randomized within 0.0.0.0 - 127.255.255.255 when specifying 0.0.0.0/1

### Other information
I guess the 0.0.0.0/0 comes from [here](https://github.com/PowerDNS/pdns/blob/29d9168874867c712da624bc4173b62f07cb79bf/pdns/calidns.cc#L370) and should be replaced [here](https://github.com/PowerDNS/pdns/blob/29d9168874867c712da624bc4173b62f07cb79bf/pdns/calidns.cc#L168) but might be considered as empty() ?

Running it with 0.0.0.0/0:
```
calidns --ecs 0.0.0.0/0 --minimum-success-rate 98 xy 1.2.3.4 200 0
Adding ECS option to outgoing queries with random addresses from the 0.0.0.0/0 range

            Option: CSUBNET - Client subnet
                Option Code: CSUBNET - Client subnet (8)
                Option Length: 8
                Option Data: 0001200000000000
                Family: IPv4 (1)
                Source Netmask: 32
                Scope Netmask: 0
                Client Subnet: 0.0.0.0
```

Running it with 0.0.0.0/1:
```
calidns --ecs 0.0.0.0/1 --minimum-success-rate 98 xy 1.2.3.4 200 0
Adding ECS option to outgoing queries with random addresses from the 0.0.0.0/1 range

            Option: CSUBNET - Client subnet
                Option Code: CSUBNET - Client subnet (8)
                Option Length: 8
                Option Data: 000120003adf5ffc
                Family: IPv4 (1)
                Source Netmask: 32
[...]

PR #13018, opened 2023-07-12
calidns: Fix setting an ECS source of 0

rgacogne:
### Short description

Fixes https://github.com/PowerDNS/pdns/issues/13015.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/calidns.cc
+++ b/pdns/calidns.cc
@@ -121,7 +121,7 @@ static ComboAddress getRandomAddressFromRange(const Netmask& ecsRange)
 {
   ComboAddress result = ecsRange.getMaskedNetwork();
   uint8_t bits = ecsRange.getBits();
-  uint32_t mod = 1 << (32 - bits);
+  uint32_t mod = (bits > 0) ? (1 << (32 - bits)) : std::numeric_limits<uint32_t>::max();
   result.sin4.sin_addr.s_addr = result.sin4.sin_addr.s_addr + ntohl(dns_random(mod));
   return result;
 }
--- a/pdns/calidns.cc
+++ b/pdns/calidns.cc
@@ -121,8 +121,14 @@ static ComboAddress getRandomAddressFromRange(const Netmask& ecsRange)
 {
   ComboAddress result = ecsRange.getMaskedNetwork();
   uint8_t bits = ecsRange.getBits();
-  uint32_t mod = (bits > 0) ? (1 << (32 - bits)) : std::numeric_limits<uint32_t>::max();
-  result.sin4.sin_addr.s_addr = result.sin4.sin_addr.s_addr + ntohl(dns_random(mod));
+  if (bits > 0) {
+    uint32_t mod = 1 << (32 - bits);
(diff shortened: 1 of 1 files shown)


pdns/calidns.cc
```diff
@@ -121,7 +121,7 @@ static ComboAddress getRandomAddressFromRange(const Netmask& ecsRange)
 {
   ComboAddress result = ecsRange.getMaskedNetwork();
   uint8_t bits = ecsRange.getBits();
-  uint32_t mod = 1 << (32 - bits);
+  uint32_t mod = (bits > 0) ? (1 << (32 - bits)) : std::numeric_limits<uint32_t>::max();
   result.sin4.sin_addr.s_addr = result.sin4.sin_addr.s_addr + ntohl(dns_random(mod));
```
omoerbeek, line 125: Pedantic mode: if `mod` is  `std::numeric_limits<uint32_t>::max()` dns_random_uint32() should be used, otherwise 0xffffffff will never be chosen.

cmouse: also since there are dns helpers that are actually verified to return somewhat uniform result, it's no good to reinvent here.

rgacogne: > Pedantic mode: if `mod` is `std::numeric_limits<uint32_t>::max()` dns_random_uint32() should be used, otherwise 0xffffffff will never be chosen.

Done!

pdns/calidns.cc
```diff
@@ -121,8 +121,14 @@ static ComboAddress getRandomAddressFromRange(const Netmask& ecsRange)
 {
   ComboAddress result = ecsRange.getMaskedNetwork();
   uint8_t bits = ecsRange.getBits();
-  uint32_t mod = 1 << (32 - bits);
-  result.sin4.sin_addr.s_addr = result.sin4.sin_addr.s_addr + ntohl(dns_random(mod));
+  if (bits > 0) {
+    uint32_t mod = 1 << (32 - bits);
+    result.sin4.sin_addr.s_addr = result.sin4.sin_addr.s_addr + ntohl(dns_random(mod));
```
Habbie, line 126: although it makes no functional difference, this should be `htonl`

Habbie (approved):
your change is approved but I noticed a pre-existing mistake

pdns/calidns.cc
rgacogne: Nice catch, I'll fix it!

rgacogne:
Interesting, I pushed a commit to that branch yesterday that does show up on my fork (https://github.com/rgacogne/pdns/tree/calidns-fix-ecs-0), which is also where clicking on the "from [rgacogne:calidns-fix-ecs-0](https://github.com/rgacogne/pdns/tree/calidns-fix-ecs-0)" link on the top this page gets me, but the commit does not show up on this page..

rgacogne:
I'm assuming some GH trigger misfired or something. Force pushing fixed it.

omoerbeek (approved):

PR #13019, opened 2023-07-12
calidns: Prevent a crash on an empty domains file

rgacogne:
### Short description

calidns could have crashed with a null pointer derefence if the main thread exits before the receiver thread has finished, because the vector of sockets no longer existed.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/calidns.cc
+++ b/pdns/calidns.cc
@@ -55,6 +55,7 @@ static po::variables_map g_vm;
 
 static bool g_quiet;
 
+//NOLINTNEXTLINE(performance-unnecessary-value-param): we do want a copy to increase the reference count, thank you very much
 static void recvThread(const std::shared_ptr<std::vector<std::unique_ptr<Socket>>> sockets)
 {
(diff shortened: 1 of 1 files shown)


pdns/calidns.cc
```diff
@@ -55,10 +55,14 @@ static po::variables_map g_vm;
 
 static bool g_quiet;
 
-static void* recvThread(const vector<std::unique_ptr<Socket>>* sockets)
+//NOLINTNEXTLINE(performance-unnecessary-value-param): we do want a copy to increase the reference count, thank you very much
+static void recvThread(const std::shared_ptr<std::vector<std::unique_ptr<Socket>>> sockets)
```
fredmorcos, line 59: Still boggles my mind that the tooling doesn't understand the language's standard library.

fredmorcos (approved):

Habbie (approved):

PR #13021, opened 2023-07-12
Rec: fix setting of policy tags

omoerbeek:
### Short description

As `gettag` can add protobuf policy tags, we do not want to cache the protobuf policy tags stored (and therefore fixed) in the packet cache.
Also, set `SocketFamily` always (was missing on a cache hit).

This is a Draft PR, as the current state of this PR changes behaviour if policy tags are set in an interception function (e.g. `postResolve`) later in the resolving process. In that case the policy tags should en up in the packet cache. I have to think a bit how to arrange that, and what to do if `gettag` also add tags for a case that *does* have policy tags stored already in the packet cache.

Reported by @wojas

Follow-up: we now keep track of the tags set by `gettag(_ffi)` separately. This allows for them to not end up in the PC.
Do pass the tags set by `gettag(_ffi)` to other other Lua functions, they might expect to see them and use them for conditional processing.
Before constructing the partial protobuf message that gets stored into the PC, we strip the tags set by `gettag(_ffi)` from the set to be stored into the PC.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/recursordist/pdns_recursor.cc
+++ b/pdns/recursordist/pdns_recursor.cc
@@ -906,6 +906,18 @@ static uint32_t capPacketCacheTTL(const struct dnsheader& hdr, uint32_t ttl, boo
   return ttl;
 }
 
+static void addPolicyTagsToPBMessageIfNeeded(DNSComboWriter& comboWriter, pdns::ProtoZero::RecMessage& pbMessage)
+{
+  if (!comboWriter.d_gettagPolicyTags.empty()) {
+    for (const auto& tag : comboWriter.d_gettagPolicyTags) {
+      comboWriter.d_policyTags.erase(tag);
+    }
+  }
+  if (!comboWriter.d_policyTags.empty()) {
+    pbMessage.addPolicyTags(comboWriter.d_policyTags);
+  }
+}
+
 void startDoResolve(void* arg) // NOLINT(readability-function-cognitive-complexity): https://github.com/PowerDNS/pdns/issues/12791
 {
   auto comboWriter = std::unique_ptr<DNSComboWriter>(static_cast<DNSComboWriter*>(arg));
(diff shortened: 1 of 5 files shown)


pdns/recursordist/pdns_recursor.cc
```diff
@@ -1679,9 +1691,10 @@ void startDoResolve(void* arg) // NOLINT(readability-function-cognitive-complexi
         pbMessage.setAppliedPolicyHit(appliedPolicy.d_hit);
         pbMessage.setAppliedPolicyKind(appliedPolicy.d_kind);
       }
-      pbMessage.addPolicyTags(comboWriter->d_policyTags);
       pbMessage.setInBytes(packet.size());
       pbMessage.setValidationState(resolver.getValidationState());
+      // See if we want to store the policyTags into th PC
```
rgacogne, line 1696:
```suggestion
      // See if we want to store the policyTags into the PC
```

pdns/recursordist/pdns_recursor.cc
```diff
@@ -906,6 +906,18 @@ static uint32_t capPacketCacheTTL(const struct dnsheader& hdr, uint32_t ttl, boo
   return ttl;
 }
 
+static void addPolicyTagsToPBMessageIfNeeded(DNSComboWriter& comboWriter, pdns::ProtoZero::RecMessage& pbMessage)
+{
+  if (!comboWriter.d_gettagPolicyTags.empty()) {
```
rgacogne, line 911:
```suggestion
  /* we do _not_ want to store policy tags set by the gettag hook into the packet cache,
     since the call to gettag for subsequent queries could yield the same PC tag but different policy tags */
  if (!comboWriter.d_gettagPolicyTags.empty()) {
```

rgacogne (approved):

rgacogne (approved):

PR #13043, opened 2023-07-17
rec: set TTL in getFakePTRRecords

omoerbeek:
Alternative solution to #13011

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/recursordist/pdns_recursor.cc
+++ b/pdns/recursordist/pdns_recursor.cc
@@ -768,14 +768,19 @@ int getFakePTRRecords(const DNSName& qname, vector<DNSRecord>& ret)
   }
   newquery += "in-addr.arpa.";
 
+  auto log = g_slog->withName("dns64")->withValues("method", Logging::Loggable("getPTR"));
+  vector<DNSRecord> answers;
+  int rcode = directResolve(DNSName(newquery), QType::PTR, QClass::IN, answers, t_pdl, log);
+
   DNSRecord record;
   record.d_name = qname;
   record.d_type = QType::CNAME;
   record.setContent(std::make_shared<CNAMERecordContent>(newquery));
+  // Copy the TTL of the synhtesized CNAME from the actual answer
+  record.d_ttl = (rcode == RCode::NoError && !answers.empty()) ? answers.at(0).d_ttl : SyncRes::s_minimumTTL;
   ret.push_back(record);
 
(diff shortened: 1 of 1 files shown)


pdns/recursordist/pdns_recursor.cc
```diff
@@ -768,14 +768,19 @@ int getFakePTRRecords(const DNSName& qname, vector<DNSRecord>& ret)
...
 
+  auto log = g_slog->withName("dns64")->withValues("method", Logging::Loggable("getPTR"));
+  vector<DNSRecord> answers;
+  int rcode = directResolve(DNSName(newquery), QType::PTR, QClass::IN, answers, t_pdl, log);
+
   DNSRecord record;
   record.d_name = qname;
   record.d_type = QType::CNAME;
   record.setContent(std::make_shared<CNAMERecordContent>(newquery));
+  // Copy the TTL of the synhtesized CNAME from the actual answer
```
Habbie, line 779:
```suggestion
  // Copy the TTL of the synthesized CNAME from the actual answer
```

pdns/recursordist/pdns_recursor.cc
```diff
@@ -768,14 +768,19 @@ int getFakePTRRecords(const DNSName& qname, vector<DNSRecord>& ret)
...
+  auto log = g_slog->withName("dns64")->withValues("method", Logging::Loggable("getPTR"));
+  vector<DNSRecord> answers;
+  int rcode = directResolve(DNSName(newquery), QType::PTR, QClass::IN, answers, t_pdl, log);
+
   DNSRecord record;
   record.d_name = qname;
   record.d_type = QType::CNAME;
   record.setContent(std::make_shared<CNAMERecordContent>(newquery));
+  // Copy the TTL of the synthesized CNAME from the actual answer
+  record.d_ttl = (rcode == RCode::NoError && !answers.empty()) ? answers.at(0).d_ttl : SyncRes::s_minimumTTL;
```
rgacogne, line 780: Do we care to check if the first answer record is actually the `PTR` one? I'm not sure we do, to be clear.

rgacogne (approved):

pdns/recursordist/pdns_recursor.cc
omoerbeek: I do not care until I see a case where it would make a difference.

rgacogne (approved):

PR #13062, opened 2023-07-21
auth: add loglevel-show setting

### LINKED ISSUE #12988: Add a configuration option to prefix log messages sent to stdout with the log level
 - Issue type: Feature request

### Short description

Currently auth (well recursor as well, but it doesn't matter so much with recursor since the new structured logging now adds log level) does not add level level to log messages which are sent to stdout. This matters particularly for docker/kubernetes deployments, where all logging is via stdout and thus there is no log level indication in the logs by default.

### Usecase

See log levels in log messages sent to stdout.

### Description

It would be good if there was a configuration option to add a log level prefix to the log messages sent to stdout. It appears that logger.h/cc already has this facility, it's just not being used by auth currently.

Habbie:
### Short description
When enabled, log messages are formatted like structured logs, including their log level/priority: 

`msg="Unable to launch, no backends configured for querying" prio="Error"`

fixes #12988 

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -974,6 +974,18 @@ Corresponds to "syslog" level values (e.g. 0 = emergency, 1 = alert, 2 = critica
 Each level includes itself plus the lower levels before it.
 Not recommended to set this below 3.
 
+.. _setting-loglevel-show:
+
+``loglevel-show``
+-------------------
+
+-  Bool
+-  Default: no
+
+.. versionadded:: 4.9.0
+
+When enabled, log messages are formatted like structured logs, including their log level/priority: ``msg="Unable to launch, no backends configured for querying" prio="Error"``
+
 .. _setting-lua-axfr-script:
 
 ``lua-axfr-script``
--- a/pdns/logger.cc
+++ b/pdns/logger.cc
@@ -57,13 +57,14 @@ void Logger::log(const string& msg, Urgency u) noexcept
   bool mustAccount(false);
 #endif
   if (u <= consoleUrgency) {
-    char buffer[50] = "";
+    std::array<char, 50> buffer{};
+    buffer[0] = '\0';
     if (d_timestamps) {
       struct tm tm;
       time_t t;
       time(&t);
       localtime_r(&t, &tm);
-      if (strftime(buffer, sizeof(buffer), "%b %d %H:%M:%S ", &tm) == 0) {
+      if (strftime(buffer.data(), buffer.size(), "%b %d %H:%M:%S ", &tm) == 0) {
         buffer[0] = '\0';
       }
     }
@@ -101,14 +102,20 @@ void Logger::log(const string& msg, Urgency u) noexcept
       }
     }
 
-    static std::mutex m;
-    std::lock_guard<std::mutex> l(m); // the C++-2011 spec says we need this, and OSX actually does
+    static std::mutex mutex;
+    std::lock_guard<std::mutex> lock(mutex); // the C++-2011 spec says we need this, and OSX actually does
+
+    // To avoid issuing multiple syscalls, we write the complete line to clog with a single << call.
+    // For that we need a buffer allocated, we might want to use writev(2) one day to avoid that.
+    ostringstream line;
+    line << buffer.data();
     if (d_prefixed) {
-      clog << string(static_cast<char*>(&buffer[0])) << "msg=" << std::quoted(msg) << " prio=" << std::quoted(severity) << endl;
+      line << "msg=" << std::quoted(msg) << " prio=" << std::quoted(severity) << endl;
     }
     else {
-      clog << string(static_cast<char*>(&buffer[0])) << msg << endl;
+      line << msg << endl;
     }
+    clog << line.str();
 #ifndef RECURSOR
     mustAccount = true;
 #endif
--- a/pdns/logger.cc
+++ b/pdns/logger.cc
@@ -19,6 +19,7 @@
  * along with this program; if not, write to the Free Software
  * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
  */
+#include <ostream>
 #ifdef HAVE_CONFIG_H
 #include "config.h"
 #endif
@@ -115,7 +116,7 @@ void Logger::log(const string& msg, Urgency u) noexcept
     else {
       line << msg << endl;
     }
-    clog << line.str();
+    clog << line.str() << std::flush;
 #ifndef RECURSOR
     mustAccount = true;
 #endif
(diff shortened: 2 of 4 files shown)


docs/settings.rst
```diff
@@ -962,6 +962,16 @@ Corresponds to "syslog" level values (e.g. 0 = emergency, 1 = alert, 2 = critica
 Each level includes itself plus the lower levels before it.
 Not recommended to set this below 3.
 
+.. _setting-loglevel-prefix:
```
omoerbeek, line 965: Please add `.. versionadded:: ...`

omoerbeek (approved):
Approved with one nit.

docs/settings.rst
Habbie: ah yes thanks! I amended the commit

Habbie:
@neilcook is this an output format that works for you?

neilcook:
> @neilcook is this an output format that works for you?

Yes, perfect

danielslab:
It would be nice if the format could be configurable as well.
The containers of OX-Cloudcontrol have the following format:
`time="2023-07-03T09:53:25Z" level=info msg="LS readiness changed to: true" app=auth-agent component=readiness`

for log processing it would be better if all components of CloudControl had the same log format.

Habbie:
That's a great point. I'll adjust this to emit `level=info msg="...."` (with the right level substituted, of course)

Habbie:
reworked to match Recursor's log output format, which sadly is slightly different from CloudControl's

danielslab:
@neilcook Maybe it is an idea for a future product improvement for cloudcontrol that all used containers have the same log format?

neilcook:
> @neilcook Maybe it is an idea for a future product improvement for cloudcontrol that all used containers have the same log format?

Yes, it's a topic we're aligning on internally. Logging in the recursor/auth/dnsdist has historically been a completely separate topic from logging in cloudcontrol, but alignment is definitely a goal.

pdns/logger.cc
```diff
@@ -67,42 +68,47 @@ void Logger::log(const string& msg, Urgency u) noexcept
...
+        severity = "None";
         break;
       }
     }
 
     static std::mutex m;
     std::lock_guard<std::mutex> l(m); // the C++-2011 spec says we need this, and OSX actually does
-    clog << string(buffer) + prefix + msg << endl;
+    if (d_prefixed) {
+      clog << string(static_cast<char*>(&buffer[0])) << "msg=" << std::quoted(msg) << " prio=" << std::quoted(severity) << endl;
```
omoerbeek, line 107: Probably more nice to change buffer to:
```
    std::array<char, 50> buffer{};
    buffer[0] = '\0';
```
and then use `buffer.size()` and `buffer.data()` in appropriate places

omoerbeek (commented):
Ok, since recursor does not use `setPrefixed()` itself.

pdns/logger.cc
omoerbeek: Also, `msg` is glued to prefix.

omoerbeek: I am also wondering if we should keep using a single string, so that the actual write happens in one syscall.

omoerbeek: I'm doing a few tests to see if the number of syscalls per log line is changing

omoerbeek:
```
99113 pdns_recursor CALL  write(2,0x1f972cf9780,0x90)
 99113 pdns_recursor GIO   fd 2 wrote 144 bytes
       "Aug 01 09:21:59 msg="Listening for HTTP requests" subsystem="webserver\
        " level="0" prio="Info" tid="0" ts="1690874519.656" address="0.0.0.0:8\
        083""
 99113 pdns_recursor RET   write 144/0x90
 99113 pdns_recursor CALL  write(2,0x1f8adb46b8f,0x1)
 99113 pdns_recursor GIO   fd 2 wrote 1 bytes
       "
       "
 99113 pdns_recursor RET   write 1
```

With the PR:
```2629 pdns_recursor CALL  write(2,0xa365443aa51,0x10)
  2629 pdns_recursor GIO   fd 2 wrote 16 bytes
       "Aug 01 09:24:04 "
  2629 pdns_recursor RET   write 16/0x10
  2629 pdns_recursor CALL  write(2,0xa3623d3f0c0,0x80)
  2629 pdns_recursor CALL  kbind(0x79387a645d78,24,0xc7ca0038866b7d4a)
  2629 pdns_recursor CALL  kbind(0xa36c599a6e8,24,0xc7ca0038866b7d4a)
  2629 pdns_recursor GIO   fd 2 wrote 128 bytes
       "msg="Listening for HTTP requests" subsystem="webserver" level="0" prio\
        ="Info" tid="0" ts="1690874644.702" address="0.0.0.0:8083""
  2629 pdns_recursor RET   kbind 0
  2629 pdns_recursor RET   write 128/0x80
  2629 pdns_recursor CALL  kbind(0x79387a645d78,24,0xc7ca0038866b7d4a)
  2629 pdns_recursor RET   kbind 0
  2629 pdns_recursor CALL  write(2,0xa365443a93f,0x1)
  2629 pdns_recursor GIO   fd 2 wrote 1 bytes
       "
       "
  2629 pdns_recursor RET   write 1
 ```

So the line itself is split with this PR into a separate write for the timestamp and the msg. The end-of-line handling did not change (but we might want to change that so it's all in a single write)

Comment (Habbie) [reply pdns/logger.cc]: thanks for pushing this change!

**Comment by rgacogne** [comment]
Are we moving forward with the format as it currently is in this PR? I'm asking because I would like https://github.com/PowerDNS/pdns/pull/13156 to be aligned with what the auth and rec are doing :)

**Comment by Habbie** [comment]
This should be fully aligned with rec (except I don't have as much metadata about things like components). I suggest you use rec as your example. I'll check again whether this PR is aligned with rec.

**Comment by Habbie** [comment]
auth (with this PR):

```
Nov 02 17:05:42 msg="Creating backend connection for TCP" prio="Error"
Nov 02 17:05:42 msg="About to create 3 backend threads for UDP" prio="Warning"
Nov 02 17:05:42 msg="Done launching threads, ready to distribute questions" prio="Warning"
```

rec (master):
```
Nov 02 16:55:47 msg="Inserting rfc 1918 private space zones" subsystem="config" level="0" prio="Notice" tid="0" ts="1698940547.587"
Nov 02 16:55:47 msg="Fatal error" error="Resolver binding to server socket on 127.0.0.1:53: Permission denied" subsystem="config" level="0" prio="Critical" tid="0" ts="1698940547.590" exception="PDNSException"
```

so `msg` and `prio` are identical; all other fields are missing from auth. `ts` would be easy to add if necessary.

omoerbeek (approved):

PR #13067, opened 2023-07-24
Use backticks in rec_control(1)

jsoref:
### Short description

Use backticks in [rec_control(1)](https://doc.powerdns.com/recursor/manpages/rec_control.1.html)

> wipe-cache DOMAIN [DOMAIN] […]
Wipe entries for DOMAIN (exact name match) from the cache. This is useful if, for example, an important server has a new IP address, but the TTL has not yet expired. Multiple domain names can be passed. DOMAIN can be suffixed with a ‘$’. to delete the whole tree from the cache. i.e. ‘powerdns.com$’ will remove all cached entries under and including the powerdns.com name.

### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/recursordist/docs/manpages/rec_control.1.rst
+++ b/pdns/recursordist/docs/manpages/rec_control.1.rst
@@ -69,14 +69,14 @@ current-queries
     Shows the currently active queries.
 
 clear-dont-throttle-names NAME [NAME...]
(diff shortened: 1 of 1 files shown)


omoerbeek:
This renders as html without any quotes. Was that intended? I'd also have to check what the man page rendering would look like.

PR #13071, opened 2023-07-25
Fix code producing json

### LINKED ISSUE #13050: Unicode characters mangled in RRset comments retrieved through the API
- Program: Authoritative
 - Issue type: Bug report

### Short description
RRset comments containing Unicode (UTF-8) characters are mangled on retrieval. They are correctly stored in the database.

### Environment

 - Operating system: RHEL 8.8
 - Software version: pdns-4.7.4-1pdns.el8.x86_64
 - Software source: https://repo.powerdns.com/repo-files/el-auth-47.repo

### Steps to reproduce
1. Insert a RRset comment, e.g. by using this API call:
```
curl -X PATCH --data '{"rrsets": [ {"comments": [ { "account": "uedvt359", "content": "Motörhead" } ], "name": "motorhead.example.org.", "type": "A", "changetype": "REPLACE", "records": [ {"content": "8.8.8.8", "disabled": false } ], "ttl": 900 } ] }' -H 'X-API-Key: ...' http://localhost:8081/api/v1/servers/localhost/zones/example.org.
```
2. verify that the comment ended up in the Database correctly:
```
# psql -Upostgres -dpowerdns
powerdns=# select * from comments where name like 'motorhead.%';
  id   | domain_id |         name          | type | modified_at | account  | comment  
-------+-----------+-----------------------+------+-------------+----------+----------
 81337 |       564 | motorhead.example.org | A    |  1689757634 | uedvt359 | Motörhead
```
3. retrive the RRset again from the API
```
curl http://localhost:8081/api/v1/servers/localhost/zones/example.org. -H "X-API-Key: ..."|jq .
// ...
  "rrsets": [
    {
      "comments": [
        {
          "account": "uedvt359",
          "content": "Mot￿f￿frhead",
          "modified_at": 1689758495
        }
      ],
// ...
```

### Expected behaviour
The comment says `Motörhead`.

### Actual behaviour
The comment says `Mot￿f￿frhead`. The `ö` is replaced by `\u{ffff}f\u{ffff}f` (the invalid noncharacter, followed by a literal lowercase f; twice).

omoerbeek:
### Short description

Should fix #13050, followup to #12260.

This reverts #12260 and adds a function to test if a URL passed in is valid.

There remains one potential issue: through the API, it is possible to produce data that is not properly UTF8, e.g. by retrieving log lines which themselves can contains arbitrary data (I'm assuming this to be on the safe side as I have no evidence otherwise). Currently we pass that data through, so it ends up in the json output. At the moment we do not have any function to validate if the source data we pass through via json is valid UTF8, and so it's hard to filter invalid data out. There's code to be found on the internet to validate if data is valid UTF8 (e.g. http://bjoern.hoehrmann.de/utf-8/decoder/dfa/), we could use that, but there still remains to decide what to do if we encounter invalid UTF8: skip, encode in some form, fix the source, ...

Draft as this needs more tests.

(BTW, I noticed the API does not handle % encoded, but valid URLs, e.g. `http://localhost:8083/metric%73` (`s` is ASCII `0x73`) I'm wondering if this a defect in YaHTTP, or in the code that matches requests to handlers).

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/test-webserver_cc.cc
+++ b/pdns/test-webserver_cc.cc
@@ -18,10 +18,16 @@ BOOST_AUTO_TEST_CASE(test_validURL)
     {"http://www.powerdns.com/\x7f?foo=123", false},
     {"http://www.powerdns.com/\x80?foo=123", false},
     {"http://www.powerdns.com/?\xff", false},
+    {"/?foo=123&bar", true},
+    {"/?foo=%ff&bar", true},
+    {"/?\x01foo=123", false},
+    {"/?foo=123\x01", false},
+    {"/\x7f?foo=123", false},
+    {"/\x80?foo=123", false},
+    {"/?\xff", false},
   };
 
   for (const auto& testcase : urls) {
-    cerr << testcase.first << endl;
     BOOST_CHECK_EQUAL(WebServer::validURL(testcase.first), testcase.second);
   }
 }
--- a/pdns/test-webserver_cc.cc
+++ b/pdns/test-webserver_cc.cc
@@ -1,0 +1,29 @@
+#define BOOST_TEST_DYN_LINK
+#define BOOST_TEST_NO_MAIN
+
+#include <boost/test/unit_test.hpp>
+#include "webserver.hh"
+
+BOOST_AUTO_TEST_SUITE(test_webserver_cc)
+
+BOOST_AUTO_TEST_CASE(test_validURL)
+{
+  // We cannot test\x00 as embedded NULs are not handled by YaHTTP other than stopping the parsing
+  const std::vector<std::pair<string, bool>> urls = {
+    {"http://www.powerdns.com/?foo=123", true},
+    {"http://ww.powerdns.com/?foo=%ff", true},
+    {"http://\x01ww.powerdns.com/?foo=123", false},
+    {"http://\xffwww.powerdns.com/?foo=123", false},
+    {"http://www.powerdns.com/?foo=123\x01", false},
+    {"http://www.powerdns.com/\x7f?foo=123", false},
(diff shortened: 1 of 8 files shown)


pdns/test-webserver_cc.cc
```diff
@@ -0,0 +1,29 @@
...
+  const std::vector<std::pair<string, bool>> urls = {
+    {"http://www.powerdns.com/?foo=123", true},
+    {"http://ww.powerdns.com/?foo=%ff", true},
+    {"http://\x01ww.powerdns.com/?foo=123", false},
+    {"http://\xffwww.powerdns.com/?foo=123", false},
+    {"http://www.powerdns.com/?foo=123\x01", false},
+    {"http://www.powerdns.com/\x7f?foo=123", false},
+    {"http://www.powerdns.com/\x80?foo=123", false},
+    {"http://www.powerdns.com/?\xff", false},
+  };
```
uedvt359, line 21: I think it might be good to add some URIs without host (`/...`) to check that validURLChars doesn't croak when some `url` members are unset. (i haven't looked how yahttp initializes them). While HTTP requests might contain full URLs, mostly we will receive just paths + query strings.

rgacogne (approved):
This looks OK to me. I would prefer to have the validation code to a different file that can be used by dnsdist's webserver (`webserver.hh` pulls in `logging.hh` so it's not an option at the moment) but I can do that in a follow-up PR.

zeha:
>  I'm wondering if this a defect in YaHTTP, or in the code that matches requests to handlers

registerApiHandler et al delegate to YaHTTP::Router, so i'd look there.

Habbie (approved):

PR #13084, opened 2023-07-31
dnsdist: Fix the group of the dnsdist.conf file when installed via RPM

rgacogne:
### Short description

I _hope_ the `%attr` directive does the right thing, based on http://ftp.rpm.org/max-rpm/s1-rpm-anywhere-specifying-file-attributes.html
I'm using this directive instead of calling `chgrp` in `%post` because I'm told using `chgrp` would make `verify` complain about the ownership later.

This replaces https://github.com/PowerDNS/pdns/pull/13037 which was not working properly.

### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/builder-support/specs/dnsdist.spec
+++ b/builder-support/specs/dnsdist.spec
@@ -113,7 +113,6 @@ make %{?_smp_mflags} check || (cat test-suite.log && false)
 %make_install
 install -d %{buildroot}/%{_sysconfdir}/dnsdist
 %{__mv} %{buildroot}%{_sysconfdir}/dnsdist/dnsdist.conf-dist %{buildroot}%{_sysconfdir}/dnsdist/dnsdist.conf
-chgrp dnsdist %{buildroot}/%{_sysconfdir}/dnsdist/dnsdist.conf
 chmod 0640 %{buildroot}/%{_sysconfdir}/dnsdist/dnsdist.conf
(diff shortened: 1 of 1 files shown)


Habbie:
I'm testing this.

Habbie:
`-rw-r----- 1 root dnsdist 4478 Aug  1 07:29 /etc/dnsdist/dnsdist.conf`

builder-support/specs/dnsdist.spec
```diff
@@ -62,9 +62,6 @@ dnsdist is a high-performance DNS loadbalancer that is scriptable in Lua.
 %prep
 %autosetup -p1 -n %{name}-%{getenv:BUILDER_VERSION}
 
-# run as dnsdist user
-sed -i '/^ExecStart/ s/dnsdist/dnsdist -u dnsdist -g dnsdist/' dnsdist.service.in
```
Habbie, line 66: commit message doesn't mention this change, while it seems important enough to get into the changelog

Habbie (changes requested):
code good, but a nit about the commit message/PR title

builder-support/specs/dnsdist.spec
rgacogne: Indeed, and it should even have been a separate commit. I'll make it happen, thanks!

rgacogne: I removed this part of the commit, and moved it to a separate PR: https://github.com/PowerDNS/pdns/pull/13088

Habbie (approved):

PR #13091, opened 2023-08-01
auth: Fixes a few issues reported by Coverity

rgacogne:
### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/pdnsutil.cc
+++ b/pdns/pdnsutil.cc
@@ -32,6 +32,7 @@
 #include <fstream>
 #include <utility>
 #include <cerrno>
+#include <sys/stat.h>
 #include <termios.h>            //termios, TCSANOW, ECHO, ICANON
 #include "opensslsigners.hh"
 #ifdef HAVE_LIBSODIUM
@@ -1171,6 +1172,13 @@ static int editZone(const DNSName &zone, const PDNSColors& col) {
     cerr << "Zone '" << zone << "' not found!" << endl;
     return EXIT_FAILURE;
   }
+
+  /* ensure that the temporary file will only
+     be accessible by the current user, not even
+     by other users in the same group, and certainly
+     not by other users.
+  */
+  umask(S_IRGRP|S_IWGRP|S_IROTH|S_IWOTH);
   vector<DNSRecord> pre, post;
   char tmpnam[]="/tmp/pdnsutil-XXXXXX";
   int tmpfd=mkstemp(tmpnam);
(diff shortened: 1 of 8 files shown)


Habbie (approved):
thanks!

pdns/pdnsutil.cc
```diff
@@ -1171,6 +1172,13 @@ static int editZone(const DNSName &zone, const PDNSColors& col) {
     cerr << "Zone '" << zone << "' not found!" << endl;
     return EXIT_FAILURE;
   }
+
+  /* ensure that the temporary file will only
```
omoerbeek, line 1176: Posix defines `mstemp` to create a file with `open(pathname, O_RDWR|O_CREAT|O_EXCL, S_IRUSR|S_IWUSR)`, but Coverity seems to insist, oh well.

omoerbeek (approved):

pdns/pdnsutil.cc
rgacogne: Are you sure? The Linux man page states:
```
In glibc versions 2.06 and earlier, the file is created with permissions  0666,  that
is,  read  and  write for all users.  This old behavior may be a security risk, espe‐
cially since other UNIX flavors use 0600, and somebody  might  overlook  this  detail
when porting programs.  POSIX.1-2008 adds a requirement that the file be created with
mode 0600.

More generally, the POSIX specification of mkstemp() does not say anything about file
modes, so the application should make sure its file mode creation mask (see umask(2))
is set appropriately before calling mkstemp() (and mkostemp()).
```
and I don't see any mention of the permissions in https://pubs.opengroup.org/onlinepubs/009604499/functions/mkstemp.html.

rgacogne: Ah, I need to learn how to read, apparently :) `POSIX.1-2008 adds a requirement that the file be created with
mode 0600.
`

rgacogne: And the updated version has it, of course: https://pubs.opengroup.org/onlinepubs/9699919799/

omoerbeek: Issue 7 does define the mode quoted above: https://pubs.opengroup.org/onlinepubs/9699919799/functions/mkdtemp.html

rgacogne:
clang-tidy is unhappy about a few things in `ixfrdist`, I'll ignore these warnings for now but will open a new pull request addressing them in a few minutes.

PR #13094, opened 2023-08-01
ixfrdist: Reduce the complexity of the main function

rgacogne:
### Short description

Very lightly tested as I am missing a complex enough setup for ixfrdist. Needs a good review!

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/ixfrdist.cc
+++ b/pdns/ixfrdist.cc
@@ -1179,12 +1179,26 @@ static bool parseAndCheckConfig(const string& configpath, YAML::Node& config) {
   return retval;
 }
 
-int main(int argc, char** argv) {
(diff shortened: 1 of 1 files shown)


Habbie:
I added a commit, but I wonder if `done` is the best word here. My first assumption was that `done` meant "configuration completely parsed" but you're using it in almost the opposite way. How about `exit0`?

rgacogne:
I agree `done` is a terrible choice. `shouldExit`, perhaps?

Habbie:
sold. I'll handle it, thanks :)

Habbie (approved):

PR #13095, opened 2023-08-01
Add a coding guidelines document to the public repository

rgacogne:
### Short description

We have had an internal version of this document for a while now, but there is no reason not to make it available to external contributors as well.

### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/CODING_GUIDELINES.md
+++ b/CODING_GUIDELINES.md
@@ -661,7 +661,7 @@ For example, the main connection handling function of a server can catch `std::e
 
 ### Catching Exceptions
 
-Catching exceptions should always be done by const reference:
+Catching exceptions should always be done by `const`-reference:
 
 ```c+++
 try {
--- a/CODING_GUIDELINES.md
+++ b/CODING_GUIDELINES.md
@@ -105,7 +105,7 @@ private:
 The use of smart pointers can be a solution to most resource leakage problems, but otherwise the only way is to be careful about exceptions in constructors:
 
 ```C++
-BadFileDescriptorWrapper()
+GoodFileDescriptorWrapper()
 {
   d_fd = open(...);
   if (something) {
--- a/CODING_GUIDELINES.md
+++ b/CODING_GUIDELINES.md
@@ -15,28 +15,28 @@ It does assume that you have already read the contributing document at [CONTRIBU
 # Memory Handling
 
 The memory model in C++, inherited from the C era, is very powerful but also very error-prone.
-Several features are available in modern (C++11 and up) C++ to make it possible to avoid most of the pitfalls, while conserving the same level of performance.
+Several features are available in modern C++ (11 and up) to make it possible to avoid most of the pitfalls, while conserving the same level of performance.
 
-Most of the issues related to memory allocation (memory leaks, use-after-free) can be solved by using standard containers, or taking advantage of RAII and smart pointers, which take care of destructing the object when it is not used anymore.
+Most of the issues related to memory allocation (memory leaks, use-after-free) can be solved by using standard containers, or taking advantage of RAII and smart pointers, which take care of destroying objects when it is not used anymore.
 
 ## Stack-based Memory Allocation
 
 Default allocations, when declaring a variable local to a function for example, are done on the stack instead of doing a dynamic allocation on the heap.
-Allocating objects on the stack is faster, especially in threaded programs, and provides the benefit that objects are automatically destroyed when the function is exited.
+Allocating objects on the stack is faster, especially in threaded programs, and provides the benefit that objects are automatically destroyed when the function exits.
 
-One caveat is that the programmer needs to be wary of is the size of the object in order not to exceed the space available on the stack, which would corrupt other objects in memory and could lead to a crash, or even execution of arbitrary code.
-This is especially true in the Recursor which uses a custom stack-switching in user-space mechanism and thus has a reduced stack size.
+One caveat that the programmer needs to be aware of is the size of the object in order to not exceed the space available on the stack, which would corrupt other objects in memory and could lead to a crash, or even execution of arbitrary code.
+This is especially true in the Recursor which uses a custom mechanism for stack-switching in user-space and thus has a reduced stack size.
 
 ### Variable-Length Arrays (VLAs)
 
 In order to avoid smashing the stack, special care should be taken to limit the depth of function calls that, for example, can grow quickly with recursion.
-A second common source of smash stacking is the use of Variable-Length Arrays, whose size is determined at runtime and is therefore very hard to predict.
-The C++ language does not support VLAs but a lot of compilers inherit such a support from C99, so it is possible to use them by accident.
+A second common source of stack smashing is the use of Variable-Length Arrays (VLAs), whose size is determined at runtime and is therefore very hard to predict.
+The C++ language does not support VLAs but a lot of compilers inherit such support from C99, so it is possible to use them by accident.
 PowerDNS strictly forbids the use of VLAs, as does the Linux kernel, and enforces that with the `-Werror=vla` compiler flag.
 
 ### C-style Arrays
 
-While you might still find some in the existing code base, we are actively trying to get rid of C-style arrays like this one:
+While you might still find some uses of C-style arrays in the existing code base, we are actively trying to get rid of them. One example is as follows:
 
 ```C++
 somestruct buffer[12];
@@ -44,9 +44,9 @@ auto bufferSize = sizeof(buffer) / sizeof(*buffer);
 auto& firstElement = buffer[0];
 ```
 
-It is immediately obvious that computing the actual number of elements is error-prone, as `sizeof()` does not return the number of elements but the total memory space used by the array.
+It is immediately obvious that computing the actual number of elements is error-prone, because `sizeof()` does not return the number of elements but the total memory space used by the array.
 Another obvious issue is that accesses to the array are not bound-checked.
-These are not the only drawbacks of these arrays, but are bad enough already to justify getting rid of them.
+These are not the only drawbacks of C-style arrays, but are bad enough already to justify getting rid of them.
 
 The modern C++ way is to use `std::array`s:
 
@@ -58,19 +58,19 @@ auto& firstElement = buffer.at(0);
 
 ### `alloca`
 
-The use of `alloca()` is forbidden in the code base as it is much too easy to smash the stack.
+The use of `alloca()` is forbidden in the code base because it is too easy to smash the stack.
 
 ## RAII
 
 Resource acquisition is initialization ([RAII](https://en.cppreference.com/w/cpp/language/raii)) is one of the fundamental concepts in C++.
 Resources are allocated during the construction of an object and destroyed when the object is itself destructed.
-It means that if an object is correctly designed, the resources associated with it can not survive its lifetime.
-Since objects that are allocated on the stack (local variables in a function, for example) are automatically destroyed when a function is exited, be it by reaching the last line, calling return or throwing an exception, it makes it possible to ensure that resources are always properly destroyed by wrapping them into an object.
+It means that if an object is correctly designed, the resources associated with it cannot survive its lifetime. In other words, the resources associated with a correctly designed object are owned by the object and cannot outlive it.
+Since stack-allocated objects, like local variables in a function, are automatically destroyed when a function exits, be it by reaching the last line, calling return or throwing an exception, it makes it possible to ensure that resources are always properly destroyed by wrapping them in an object.
 
 We describe the use of smart pointers, containers and other wrappers for that purpose below, but first a few words of caution.
-Resources stored in a object are only tied to this object if the constructor finished properly.
-If an exception is raised in the constructor body, the object is not created and therefore the destructor will not get called.
-This means that if the object has non-object members holding resources, like naked file descriptors or naked pointers, they need to be explicitly released before raising the exception, otherwise they are lost.
+Resources stored in a object are only tied to this object if the constructor executes fully and completes properly.
+If an exception is raised in the constructor's body, the object is not created and therefore the destructor will not be called.
+This means that if the object has non-object members holding resources, like raw file descriptors or raw C-style pointers, they need to be explicitly released before raising the exception, otherwise they are lost or leaked.
 
 ```C++
 class BadFileDescriptorWrapper
@@ -99,7 +99,7 @@ private:
 };
 ```
 
-The use of smart pointers can be a solution to most resources leakage, but otherwise the only way is to be careful about exceptions in constructors:
+The use of smart pointers can be a solution to most resource leakage problems, but otherwise the only way is to be careful about exceptions in constructors:
 
 ```C++
 BadFileDescriptorWrapper()
@@ -115,35 +115,35 @@ BadFileDescriptorWrapper()
 
 ## Smart Pointers
 
-There is almost no good reason not to use a smart pointer when doing dynamic memory allocation.
+There is almost no good reason to not use a smart pointer when doing dynamic memory allocation.
 Smart pointers will keep track of whether the dynamically allocated object is still used, and destroy it when the last user goes away.
 
-Using naked pointers quickly results in security issues, ranging from memory leaks to arbitrary code execution.
+Using raw pointers quickly results in security issues, ranging from memory leaks to arbitrary code execution.
 Examples of such issues can be found in the following PowerDNS security advisories:
 
 * [2017-07: Memory leak in DNSSEC parsing](https://docs.powerdns.com/recursor/security-advisories/powerdns-advisory-2017-07.html)
 * [2018-04: Crafted answer can cause a denial of service](https://docs.powerdns.com/recursor/security-advisories/powerdns-advisory-2018-04.html)
 
 Most allocations should be wrapped in a `std::unique_ptr`, using `make_unique`.
-There can only be one owner at a given time, as opposed to shared pointers, but the ownership can be passed along using `std::move()` if needed.
+There can only be one owner at any given time, as opposed to shared pointers, but the ownership can be passed along using `std::move()` if needed.
 
 If the dynamically allocated object needs to be referenced in several places, the use of a `std::shared_ptr` is advised instead, via `std::make_shared`.
 
-The use of the `make_*` methods has three advantages:
+The use of `make_*` methods has three advantages:
 
 * they result in a single allocation for `shared_ptr`s, instead of two otherwise ;
 * they avoid duplicating the type name twice ;
 * they prevent a possible issue if an exception is raised with temporaries.
 
-They also make is easier to spot naked pointers by looking for "new" and "delete" throughout the code :)
+They also make is easier to spot raw pointers by searching or `grep`ping for "new" and "delete" throughout the code :)
 
-Please note however that while unique pointers are as cheap as naked pointers, shared pointers are much more expensive.
+Please note, however, that while unique pointers are as cheap as raw pointers, shared pointers are much more expensive.
 That is because they need to use atomic operations to update their internal counters, so making a copy of a shared pointer is expensive.
 Passing one by reference is cheap, however.
 
 ### Shared Pointers
 
-An important thing to be aware of with shared pointers is that taking a new copy of a shared pointer or releasing, thus updating its internal reference counter, is atomic and therefore thread-safe.
+An important thing to be aware of with shared pointers is that making a new copy or releasing a shared pointer, thus updating its internal reference counter, is atomic and therefore thread-safe.
 Altering the content of the object pointed to is not, though, and is subject to the usual locking methods.
 The often misunderstood part is that updating the target of the shared pointer is not thread-safe.
 Basically, you can copy the shared pointer from multiple threads at once, and then each thread can assign a new target to its own copy safely, like that:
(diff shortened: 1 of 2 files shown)


Habbie:
I was about to post 40 nits as review comments, but it makes more sense for me to just push. Will do that later today.

CODING_GUIDELINES.md
```diff
@@ -0,0 +1,690 @@
...
+## const-ness
+
+* Mark parameters and variables that should not be modified as `const`.
+  This is especially important for references and pointers that comes from outside the function, but it also makes sense to do it for local variables or parameters passed by value because it might help detect a logic error later.
+* Mark const methods as such (and make them thread-safe)
+* Prefer using `at()` on containers so that no insertion can take place by mistake, and to get bounds checking.
+
+## static
+
+Functions that are only used inside a single file should be marked as `static`, so that:
```
Habbie, line 625: I think we prefer `namespace { ... }` these days?

CODING_GUIDELINES.md
```diff
@@ -0,0 +1,690 @@
...
+* [2018-04: Crafted answer can cause a denial of service](https://docs.powerdns.com/recursor/security-advisories/powerdns-advisory-2018-04.html)
+
+Most allocations should be wrapped in a `std::unique_ptr`, using `make_unique`.
+There can only be one owner at a given time, as opposed to shared pointers, but the ownership can be passed along using `std::move()` if needed.
+
+If the dynamically allocated object needs to be referenced in several places, the use of a `std::shared_ptr` is advised instead, via `std::make_shared`.
+
+The use of the `make_*` methods has three advantages:
+
+* they result in a single allocation for `shared_ptr`s, instead of two otherwise ;
```
Habbie, line 134: I see ` ;` after a lot of list items, but not after all of them. Is this some leftover from the conversion from another format? Whatever we do, we should do it consistently.

Habbie (changes requested):
This is excellent work, and re-reading it made me remember a couple of things I had meanwhile forgotten!

I pushed some changes before this review. I have requested a few changes/pointed out a few small problems in this review.

CODING_GUIDELINES.md
```diff
@@ -0,0 +1,690 @@
...
+C-style casts should be avoided, as the compiler does almost no check on the validity of the operation.
+They are also very hard to spot in a code.
+C++-style casts can easily be spotted in a code, which makes it easy to review them.
+
+* `const_cast` can be used to remove the const qualifier on a variable.
+  It's usually a bad sign, but sometimes it is needed to call a function that will not modify the variable but lacks the const qualifier, for example.
+* `dynamic_cast` can be used to cast a pointer to a derived class or to a base class, while checking that the operation is valid.
+  If the casted object is not valid for the intended type, a nullptr value will be returned (or a bad_cast exception for references) so the result of the operation should be checked!
+  Note that the RTTI check needed to verify that the casted object is valid has a non-negligible CPU cost.
+  Not checking the return value might lead to remote denial of service by nullptr dereference, as happened with the issue described in this advisory: https://docs.powerdns.com/recursor/security-advisories/powerdns-advisory-2017-08.html
```
Habbie, line 676: forgot this one in my earlier pushes:

```suggestion
  Not checking the return value might lead to remote denial of service by nullptr dereference, as happened with the issue described in this advisory: [2017-08: Crafted CNAME answer can cause a denial of service](https://docs.powerdns.com/recursor/security-advisories/powerdns-advisory-2017-08.html)
```

CODING_GUIDELINES.md
```diff
@@ -0,0 +1,690 @@
...
+  }
+private:
+  int d_fd{-1}; 
+};
+```
+
+The use of smart pointers can be a solution to most resources leakage, but otherwise the only way is to be careful about exceptions in constructors:
+
+```C++
+BadFileDescriptorWrapper()
```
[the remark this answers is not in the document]
rgacogne: Fixed by Fred :)

CODING_GUIDELINES.md
```diff
@@ -0,0 +1,690 @@
...
+
+It is very common to use pointer arithmetic to calculate a position in a buffer, or to test whether a given offset is outside of a given buffer.
+Unfortunately it is quite easy to trigger undefined behaviour when doing so, as the C++ standard does not allow pointer arithmetic pointing inside an object, except for arrays where it is also permitted to point one element past the end.
+Still that undefined behaviour is mostly harmless, but it might lead to real issue on some platforms.
+
+One such example occurred in dnsdist: [2017-01: Crafted backend responses can cause a denial of service](https://dnsdist.org/security-advisories/powerdns-advisory-for-dnsdist-2017-01.html)
+
+In that case, a pointer was set to the start of a buffer plus a given length, to see whether the result would go past another pointer that was set to the end of the buffer.
+Unfortunately, if the start of the buffer is at a very high virtual address, the result of the addition might overflow and wrap around, causing the check to become true and leading to either a crash or the reading of unrelated memory.
+While very unlikely on a 64 bits platform, it could happen on 16 or 32 bits platform.
```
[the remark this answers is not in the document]
rgacogne: Will do!

CODING_GUIDELINES.md
```diff
@@ -0,0 +1,690 @@
...
+The cost of using `at()` is negligible for most use cases, and can be avoided by using the `[]` operator in the rare case when the cost cannot be afforded.
+Note that several Linux distributions now build with `-Wp,-D_GLIBCXX_ASSERTIONS` enabled by default, which turns on cheap range checks for C++ arrays, vectors, and strings anyway.
+
+Regarding performance, it is advised to `reserve()` the needed size in advance when a rough estimate is known to avoid reallocations and copies.
+Resizing in advance is not advised, though, as it makes it harder to know what is exactly in the container in case of early returns or exceptions.
+
+In C++11, move operators make it possible to cheaply get the contents of a container into a different variable if needed.
+
+The need to pass a subset of a container without copying it often leads to passing a pointer to an array of chars along with a size.
+Introduced in C++14 but already available in PowerDNS via boost (see views.hh), views provides a nice way to borrow the content of a container to pass it to a function, without any copy or dynamic memory allocation.
```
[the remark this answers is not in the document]
rgacogne: Good catch! We already got rid of `views.hh` in https://github.com/PowerDNS/pdns/pull/10336 and I'm updating this line.

CODING_GUIDELINES.md
```diff
@@ -0,0 +1,690 @@
...
+
+Some of these tips are actually enforced by `clang-tidy` nowadays, but it is still useful to keep them in mind.
+
+## auto
+
+C++11 introduced automatic type deduction, using the `auto` keyword.
+In addition to saving the typing of a few more letters, using automatic type deduction prevents nasty surprises if the variable is initialized from another one, or from a function, and the other type is changed to a different one.
+The code might still compile while now involving a copy or worse.
+
+## Boolean expressions
```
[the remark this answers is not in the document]
rgacogne: The section is now gone, I moved the content earlier but forgot to remove the section altogether.

CODING_GUIDELINES.md
rgacogne: I'll mention both options since I don't really have an opinion on whether one is better than the other, but perhaps you or Fred do?

CODING_GUIDELINES.md
rgacogne: Applied!

CODING_GUIDELINES.md
Habbie: I have a small preference for `namespace { ... }`

CODING_GUIDELINES.md
```diff
@@ -0,0 +1,701 @@
...
+This means that if the object has non-object members holding resources, like raw file descriptors or raw C-style pointers, they need to be explicitly released before raising the exception, otherwise they are lost or leaked.
+
+```C++
+class BadFileDescriptorWrapper
+{
+  BadFileDescriptorWrapper()
+  {
+    d_fd = open(...);
+    if (something) {
+      throw std::runtime_error(...);
```
zeha, line 82:
```suggestion
      throw std::runtime_error(...);  // WRONG, DO NOT DO THIS
```

CODING_GUIDELINES.md
fredmorcos: Same here.

rgacogne: Alright!

CODING_GUIDELINES.md
```diff
@@ -0,0 +1,701 @@
...
+* It provides safe (and fast) operations like comparisons, iterators, etc..
+
+One issue that could have been prevented by the use of a container can be found in the following advisory: [2018-09: Crafted query can cause a denial of service](https://docs.powerdns.com/recursor/security-advisories/powerdns-advisory-2018-09.html)
+
+The use of a container and its corresponding `at()` operator would have prevented an out-of-bounds read since calling `at()` on an invalid offset results in an exception being raised.
+The cost of using `at()` is negligible for most use cases, and can be avoided by using the `[]` operator in the rare case when the cost cannot be afforded.
+Note that several Linux distributions now build with `-Wp,-D_GLIBCXX_ASSERTIONS` enabled by default, which turns on cheap range checks for C++ arrays, vectors, and strings.
+
+Regarding performance, it is advised to `reserve()` the needed size in advance when a rough estimate is known to avoid reallocations and copies.
+Resizing in advance is not advised, though, as it makes it harder to exactly know what is in the container in case of early returns or exceptions.
```
zeha, line 231: this conflicts with the sentence before, at least without further qualification

CODING_GUIDELINES.md
rgacogne: Will do!

CODING_GUIDELINES.md
rgacogne: Applied!

CODING_GUIDELINES.md
```diff
@@ -0,0 +1,701 @@
...
+For example getting the `size()` of a container might not be thread-safe if a different thread might be writing to the container.
+Some functions might also not be thread-safe, for example if they have a static non-const variable.
+
+We currently use three solutions, depending on the use-case.
+The first one is used when we only need to share some kind of counter or gauge, and involves the use of `std::atomic` which allows atomic operations to be performed from different threads without locking. More on that later.
+The second one is the "share nothing" approach, where each thread has its own data (using `thread_local`, for example), avoiding the need for data synchronization.
+When a thread needs to communicate with another one, it might use a `pdns::channel` to pass a pointer to that second thread.
+That works quite well but sometimes sharing data is much more efficient than the alternative.
+
+For these cases, we use the classic locking approach, using either a simple mutex or read-write lock, depending on the use case.
```
zeha, line 255:
```suggestion
For other cases, we use the classic locking approach, using either a simple mutex or read-write lock, depending on the use case.
```

?

at least it doesn't feel consistent with the "three solutions" lead in

rgacogne: I'll rewrite it.

CODING_GUIDELINES.md
```diff
@@ -0,0 +1,701 @@
...
+This is well-known as the birthday paradox, the fact that the probability of having two entries colliding is a lot higher than the probability of finding a collision for a specific entry.
+This means that it is important to verify that the entries are actually identical, and just not that they hash to the same value.
+
+This is especially important when hashing attacker-controlled values, as they can be specially crafted to trigger collisions to cause:
+
+* Cache pollution (see [2018-06: Packet cache pollution via crafted query](https://docs.powerdns.com/recursor/security-advisories/powerdns-advisory-2018-06.html)) ;
+* Denial of service via hash table flooding (in a map, all entries that hash to the same value are often placed into a linked-list, making it possible to cause a linear scan of entries by making all of them hash to that same value).
+
+The first issue can be prevented by comparing the entries and not just the value they hash to.
+The second one can be used by using some sort of secret when computing the hash so that the result cannot be guessed by the attacker.
```
zeha, line 590: used? -> avoided?

CODING_GUIDELINES.md
```diff
@@ -0,0 +1,701 @@
...
+
+# Readability Tips
+
+Some of these tips are actually enforced by `clang-tidy` nowadays, but it is still useful to keep them in mind.
+
+## `auto`
+
+C++11 introduced automatic type deduction, using the `auto` keyword.
+Using automatic type deduction prevents nasty surprises if the variable is initialized from another one, or from a function, and the other type is changed to a different one.
+The code might still compile while now involving a copy or worse.
```
zeha, line 601: could be more explicit, ex:
```suggestion
Without `auto`, code might still compile but trigger a copy or worse.
```

CODING_GUIDELINES.md
```diff
@@ -0,0 +1,701 @@
...
+
+## Boolean Expressions
```
zeha: missing section text?

CODING_GUIDELINES.md
```diff
@@ -0,0 +1,719 @@
...
+
+## `nullptr`
+
+When representing a pointer, using `nullptr` makes it immediately obvious that we are dealing with a pointer, as opposed to the use of `0`.
+It also cannot be silently taken as an integer, which can happens with `0` but also with `NULL`.
+
+## `const`-ness
+
+* Mark parameters and variables that should not be modified as `const`.
+  This is especially important for references and pointers that comes from outside the function, but it also makes sense to do it for local variables or parameters passed by value because it might help detect a logic error later ;
```
zeha, line 628:
```suggestion
  This is especially important for references and pointers that come from outside the function, but it also makes sense to do it for local variables or parameters passed by value because it might help detect a logic error later ;
```

CODING_GUIDELINES.md
```diff
@@ -0,0 +1,719 @@
...
+```
+
+but the unnamed namespace form is now preferred.
+
+For the same reason, global variables that are only accessed from a single file should be marked static as well.
+
+## Variables
+
+Try to declare variables in the innermost scope possible and avoid uninitialized variables as much as possible.
+Declare and initialize them when the values needed to initialize them are available.
```
zeha, line 664: double `them` in this sentence

CODING_GUIDELINES.md
```diff
@@ -0,0 +1,719 @@
...
+
+## Exceptions
+
+Should be reserved for unexpected events (corrupted data, timeouts, ...) and should not be triggered in normal processing.
+
+Do not be afraid of using them, though, as the cost of an exception that is not thrown is usually very small, thanks to the zero-cost exception model.
+It might still force the compiler to refrain from some optimizations, so it might make sense to avoid them in some very performance-sensitive, narrow code paths.
```
zeha, line 671: this section could use more examples or a definite list of what is considered an unexpected event. I'd think for people used to do network programming, read timeouts wouldn't be unexpected, f.e.

CODING_GUIDELINES.md
```diff
@@ -0,0 +1,719 @@
...
+
+### Custom Exceptions
+
+Exceptions defined by the standards should be used whenever possible, as they already cover a lot of use cases.
+
+If custom exceptions are necessary, to be able to catch them explicitly, they should still derive from `std::exception`, directly or indirectly, so that they can still be caught in a more generic way to prevent the program from terminating.
```
zeha, line 677: two `still` in this sentence, neither of which might really add sth

CODING_GUIDELINES.md
```diff
@@ -0,0 +1,719 @@
...
+catch (const std::exception& e) {
+  std::cerr << e.what() <<endl;
+}
+```
+
+Not using a reference would result in the exception object being sliced, meaning that a custom exception derived from `std::exception` would not see its overriding `what()` method called but the one from the base class instead.
+
+## Casts
+
+C-style casts should be avoided, as the compiler does almost no check on the validity of the operation.
```
zeha, line 696:
```suggestion
C-style casts should be avoided, as the compiler does almost no checking on the validity of the operation.
```

CODING_GUIDELINES.md
rgacogne: I'll rewrite that paragraph :+1:

CODING_GUIDELINES.md
fredmorcos: It does no conflict, reserving and resizing are two different operations.

zeha: Yeah I read it three times and only now I see the difference :)

rgacogne: Done!

CODING_GUIDELINES.md
```diff
@@ -0,0 +1,723 @@
...
+This means that it is important to verify that the entries are actually identical, and just not that they hash to the same value.
+
+This is especially important when hashing attacker-controlled values, as they can be specially crafted to trigger collisions to cause:
+
+* Cache pollution (see [2018-06: Packet cache pollution via crafted query](https://docs.powerdns.com/recursor/security-advisories/powerdns-advisory-2018-06.html)) ;
+* Denial of service via hash table flooding (in a map, all entries that hash to the same value are often placed into a linked-list, making it possible to cause a linear scan of entries by making all of them hash to that same value).
+
+The first issue can be prevented by comparing the entries and not just the value they hash to.
+The second one can be avoided by using some sort of secret when computing the hash so that the result cannot be guessed by the attacker.
+That can be achieved by using an unpredictable seed for certain hash algorithms, or a secret for some other like `SipHash`.
```
chbruyand, line 591: Maybe add a reference link for `SipHash`

Habbie (approved):

CODING_GUIDELINES.md
```diff
@@ -0,0 +1,723 @@
...
+
+C-style casts should be avoided, as the compiler does almost no checking on the validity of the operation.
+They are also very hard to spot in a code.
+C++-style casts can easily be spotted in a code, which makes it easy to review them.
+
+* `const_cast` can be used to remove the `const` qualifier on a variable.
+  It's usually a bad sign, but is sometimes needed to call a function that will not modify the variable but lacks the `const` qualifier ;
+* `dynamic_cast` can be used to cast a pointer to a derived class or to a base class, while checking that the operation is valid.
+  If the cast object is not valid for the intended type, a `nullptr` value will be returned (or a `bad_cast` exception for references) so the result of the operation should be checked!
+  Note that the RTTI check needed to verify that the cast object is valid has a non-negligible CPU cost.
```
chbruyand, line 708:
```suggestion
  Note that the RunTime Type Information (RTTI) check needed to verify that the cast object is valid has a non-negligible CPU cost.
```

chbruyand (commented):
My comments. Please excuse me and ignore them if they interfere/contradict with past comments

CODING_GUIDELINES.md
```diff
@@ -0,0 +1,723 @@
...
+
+```c++
+static bool thisOneAsWell()
+{
+}
+```
+
+but the unnamed namespace form is now preferred.
+
+For the same reason, global variables that are only accessed from a single file should be marked static as well.
```
omoerbeek, line 659: Or be put in a unnamed namespace?

CODING_GUIDELINES.md
```diff
@@ -0,0 +1,723 @@
...
+  Note that the RTTI check needed to verify that the cast object is valid has a non-negligible CPU cost.
+  Not checking the return value might lead to remote denial of service by `nullptr`-dereference, as happened with the issue described in this advisory: [2017-08: Crafted CNAME answer can cause a denial of service](https://docs.powerdns.com/recursor/security-advisories/powerdns-advisory-2017-08.html) ;
+* `static_cast` can perform downcast in place of `dynamic_cast`, with none of the cost associated to the check, but can only be done if the cast is known to be valid.
+  It can also do implicit conversion between types (from `ssize_t` to `size_t`, **after** checking that the value is greater or equal to zero) ;
+* `reinterpret_cast` is quite dangerous, since it can be used to turn a type into a different one.
+  It cannot be be used to remove a `const` qualifier.
+  When used to reinterpret the content of a buffer it can quickly lead to alignment issues, as described in the [Alignment Issues] section.
+
+## `errno`
```
omoerbeek, line 716: The `errno` block should probably be moved to after  the TOCTU section

CODING_GUIDELINES.md
```diff
@@ -0,0 +1,723 @@
...
+
+The use of smart pointers can be a solution to most resource leakage problems, but otherwise the only way is to be careful about exceptions in constructors:
+
+```C++
+GoodFileDescriptorWrapper()
+{
+  d_fd = open(...);
+  if (something) {
+    close(d_fd);
+    throw std::runtime_error(...);
```
omoerbeek, line 113:
```suggestion
    close(d_fd);
    d_fd = -1;
    throw std::runtime_error(...);
```

CODING_GUIDELINES.md
```diff
@@ -0,0 +1,723 @@
...
+
+Regarding performance, it is advised to [`reserve()`](https://en.cppreference.com/w/cpp/container/vector/reserve) the needed size in advance when a rough estimate is known to avoid reallocations and copies. It usually triggers the allocation of enough memory to hold the requested number of items but does not increase the size of the container as reported by `size()`.
+Calling [`resize()`](https://en.cppreference.com/w/cpp/container/vector/resize) in advance is not advised, though, as it makes it harder to exactly know what is in the container in case of early returns or exceptions.
+
+In C++11, move operators make it possible to cheaply get the contents of a container into a different variable if needed.
+
+The need to pass a subset of a container without copying it often leads to passing a pointer to an array of chars along with a size.
+Introduced in C++14, `views` provide a nice way to borrow the content of a container to pass it to a function, without any copying or dynamic memory allocation.
+
+The basic `string_view` class provides that feature for a container of chars, but the same feature exists for other types, like `uint8_t`.
```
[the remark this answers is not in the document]
rgacogne: It is confusing indeed, I'll remove it.

chbruyand (approved):

Habbie (approved):

omoerbeek (approved):

PR #13096, opened 2023-08-02
auth lmdb: when broadcasting indexes, -do- rewrite them even if they are unchanged

Habbie:
### Short description
before this patch, there was a chance of desync:

assume two LS+auth nodes are in sync

1. node A deletes zone example.com (this deletes the domain object, an index reference to it, and a few other things)
2. node B sends out notify for example.com which is a primary zone, and updates the domain object (last_notified)
3. LS syncs. For the domain object, the notified one is newer than the deleted one, so auth1 gets the domain object back. The index entry, however, had an older timestamp, and remains deleted

With this patch, the index entry gets a fresh timestamp too, so conflict resolution stops breaking the index.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/ext/lmdb-safe/lmdb-typed.hh
+++ b/ext/lmdb-safe/lmdb-typed.hh
@@ -164,10 +164,8 @@ struct LMDBIndexOps
 
     MDBOutVal currentvalue;
 
-    // check if the entry already exists, so we don't uselessly bump the timestamp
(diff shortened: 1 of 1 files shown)


wojas (approved):
LGTM

PR #13101, opened 2023-08-03
GH actions. build-and-test-all.yml: add make distdir as an additional step for build jobs

romeroalx:
### Short description
Adding `make distdir` as an additional step during the build of all components in `build-and-test-all.yml`

### Checklist

I have:
- [ ] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/.github/workflows/build-and-test-all.yml
+++ b/.github/workflows/build-and-test-all.yml
@@ -14,6 +14,7 @@ env:
   CLANG_VERSION: '13'
   # github.workspace variable points to the Runner home folder. Container home folder defined below.
   REPO_HOME: '/__w/pdns/pdns'
+  BUILDER_VERSION: '0.0.0-git1'
 
 jobs:
   build-auth:
@@ -29,6 +30,9 @@ jobs:
         UBSAN_OPTIONS: "print_stacktrace=1:halt_on_error=1:suppressions=${{ env.REPO_HOME }}/build-scripts/UBSan.supp"
         UNIT_TESTS: yes
       options: --sysctl net.ipv6.conf.all.disable_ipv6=0
+    defaults:
+      run:
+        working-directory: ./pdns-${{ env.BUILDER_VERSION }}
     outputs:
       clang-tidy-failed: ${{ steps.clang-tidy-annotations.outputs.failed }}
     steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/build-and-test-all.yml
```diff
@@ -50,6 +50,7 @@ jobs:
           restore-keys: auth-ccache-
       - run: inv ci-autoconf
       - run: inv ci-auth-configure
+      - run: inv ci-make-distdir
       - run: inv ci-auth-make-bear  # This runs under pdns/
```
Habbie, line 54: Can you run this build from inside the freshly generated distdir? That way we can detect missing files in dist.

Habbie (changes requested):

.github/workflows/build-and-test-all.yml
Habbie: (which means running configure again, but not autoconf)

romeroalx: Done

romeroalx: The folder name comes from the BUILDER_VERSION variable that has been fixed in `tasks.py`:

```
def ci_autoconf(c):
    c.run('BUILDER_VERSION=0.0.0-git1 autoreconf -vfi')
```

If the goal is to configure and make from inside that directory, we might need to do it with `FUZZING_TARGETS=no` or do something in addition. I will have a look.

Thank you for the review!

Habbie (approved):

.github/workflows/build-and-test-all.yml
```diff
@@ -50,13 +51,18 @@ jobs:
           restore-keys: auth-ccache-
       - run: inv ci-autoconf
       - run: inv ci-auth-configure
-      - run: inv ci-auth-make-bear  # This runs under pdns/
-      - run: ln -s .clang-tidy.full .clang-tidy
+      - run: inv ci-make-distdir
+      - run: inv ci-auth-configure
+        working-directory: pdns-${{ env.BUILDER_VERSION }}
```
rgacogne, line 56: Would it make sense to make that directory the default working directory for this job?

rgacogne (approved):
Looks good to me, nice work! I'm suggesting defining default directories but I'm fine with merging the PR as it is. It will make my life a bit harder with regard to generating code coverage but I'll deal with it :)

.github/workflows/build-and-test-all.yml
romeroalx: Makes sense. I just updated the PR to include this change.

Thank you :)

rgacogne (approved):

PR #13112, opened 2023-08-10
Modify the cache cleaning part in the reload-zones logic to make reload-zones execute faster

zjs604381586:
### Short description

When there are many **forward-zones** domain names, such as tens of thousands. The execution of the **rec_control reload-zones** command will be very slow. The more domain names in **forward-zones**, the longer the **reload-zones** will take, which increases linearly.
### Mainly modify the cache cleaning logic of **reload-zones**
- Optimize the forwarding domain name list by copying each domain name object in turn into a smart pointer for dispatching a batch of domain name lists
- Asynchronize the cache clearing logic of recursor cache & negcache
- For clearing the cache logic reload does not get the number of entries cleared this time, eliminating the blocking read of the pipeline
#### 48,000 forwarding domain names：
- **time cost**：
     - before：**4m40s**
     - After optimization and modification： **90ms**
     
#### 95,000 forwarding domain names：
- **time cost**：
     - before：**7m24s**
     - After optimization and modification： **260ms**

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


omoerbeek:
Thanks for your PR. I can see the current approach having performance issues with that many forwarding zones. I do wonder about the use-case for that many forwarding zones, and if recursor is the right spot to make the forwarding decisions in that use-case.

As for the PR itself, I did not look in detail yet, but the wiping work still has to be done. That means that some other thread is doing it, and potentially you are just moving a bottleneck from one thread to another thread.
Also note that as of 4.9.0, the packet cache is also a shared data structure, so that part of the PR would need to be rewritten for the master branch.

I am thinking about an alternate approach, which involves comparing the old and new lists of forwards, and then only wiping the zones that have a changed, are no longer forwarded or are new forwards. Implementing this is quite complex, but would avoid wiping all unchanged forwarding zones.

zjs604381586:
> Thanks for your PR. I can see the current approach having performance issues with that many forwarding zones. I do wonder about the use-case for that many forwarding zones, and if recursor is the right spot to make the forwarding decisions in that use-case.
> 
> As for the PR itself, I did not look in detail yet, but the wiping work still has to be done. That means that some other thread is doing it, and potentially you are just moving a bottleneck from one thread to another thread. Also note that as of 4.9.0, the packet cache is also a shared data structure, so that part of the PR would need to be rewritten for the master branch.
> 
> I am thinking about an alternate approach, which involves comparing the old and new lists of forwards, and then only wiping the zones that have a changed, are no longer forwarded or are new forwards. Implementing this is a bit complex, but would avoid wiping all unchanged forwarding zones.

- Our case：
     We built an iterative pdns in front of the authoritative server, which requires a gray-scale strategy for authoritative domain name resolution. Iterative pdns is needed to control which domain names are forwarded to which authoritative server. Therefore, forward-zones will configure many forwarded domain names, and it may be possible in the future For some time, the number of this configuration item has been growing


- PR illustrate：
     - This pr was modified on rec 4.8.x, mainly to encapsulate the domain name list into a **smart pointer**, which saves a lot of copying time consumption.
     - The second modification is to clear the cache asynchronously, and put the asynchronous task in the next scheduling of the thread pool to execute instead of executing it in the reload command. And there is no need to obtain the number of cleared entries (when there are many caches, the real clearing logic will also increase the time-consuming), so there is no need to perform blocking reading of the thread communication **pipe readToThread**, and the pipe can return after the successful write, reducing the time-consuming of reloading
     -  The previous version of **rec 4.8.x** was mainly time-consuming in the clearing logic of the packetcache, which required copying the domain name. I have seen the adjustment of **rec 4.9.x**, and changed the packetcache into a global share. Then can assemble the clearing of **recursor cache, negcache, and packetcache** into an asynchronous task, put it in the handle thread, and let the next handle thread schedule to execute
 - new thinking
     - I have also considered that reload only clears the domain names that have changed before and after (addition and deletion). However, if the forward-zones are deleted in full or added in a large amount, there will be time-consuming scenarios, so I did not use this solution

PR #13145, opened 2023-08-16
Fix building our fuzzing targets from a dist tarball

rgacogne:
### Short description

Until now all of our fuzzing targets where built with the authoritative server, even though one of them is specific to dnsdist.
This made it easy to build all of them at once, especially for OSS-Fuzz and CI-Fuzz, but had the unfortunate drawback of pulling several dnsdist-specific files into the main `pdns/` directory for no good reason. It also prevented building the fuzzing targets from a dist tarball/directory.
This commit moves the dnsdist-specific fuzzing target to the dnsdist build process, and ensure that the `standalone_fuzz_target_runner.cc` file is part of the dist tarball, making it possible to build the fuzzing targets from the dist.
It does not move the dnsdist-specific files to the `pdns/dnsdistdist/` directory yet because this would conflict with existing PRs.
This will require a follow-up pull request for OSS-Fuzz and CI-Fuzz to pick up the dnsdist fuzzing target once this PR has been merged.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


romeroalx (approved):

fuzzing/README.md
```diff
@@ -5,21 +5,24 @@ This repository contains several fuzzing targets that can be used with generic
 fuzzing engines like AFL and libFuzzer.
 
 These targets are built by passing the --enable-fuzz-targets option to the
```
Habbie, line 7: could use backticks around `--enable..`, same for `make fuzz_targets` below etc.

Habbie (approved):
one nit (that really is older than this PR), approved otherwise. Feel free to just merge :)

fuzzing/README.md
rgacogne: Indeed, thanks! Done that now!

rgacogne:
I'm going to merge this pull request, which means that the dnsdist fuzzing will no longer get built either OSS-fuzz or CI-fuzz. It should not cause any error, and I'll open a pull request against the OSS-fuzz repository right away.

PR #13147, opened 2023-08-16
dnsdist: DOH - make query channel receiving part blocking

chbruyand:
### Short description
When not in single acceptor mode, the DOH thread processing incoming queries after they have been parsed is using 100% as the incoming channel is not blocking. This PR makes the receiving part to block.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


rgacogne:
I introduced that bug in the channel PR, so after 1.8.0, fortunately..

rgacogne:
I think it's time to replace the booleans by `enum class` objects so it's clearer what they mean, I'll do that in a follow-up PR.

PR #13153, opened 2023-08-18
auth: Move method checking to Router

cmouse:
### Short description
This change moves the HTTP verb checking into YaHTTP router. 

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/recursordist/ws-recursor.cc
+++ b/pdns/recursordist/ws-recursor.cc
@@ -515,7 +515,7 @@ static void apiServerRPZStats(HttpRequest* /* req */, HttpResponse* resp)
   resp->setJsonBody(ret);
 }
 
-static void prometheusMetrics(HttpRequest* req, HttpResponse* resp)
+static void prometheusMetrics(HttpRequest* /* req */, HttpResponse* resp)
 {
   static MetricDefinitionStorage s_metricDefinitions;
 
--- a/pdns/recursordist/ws-recursor.cc
+++ b/pdns/recursordist/ws-recursor.cc
@@ -389,40 +389,44 @@ static void apiServerZonesGET(HttpRequest* /* req */, HttpResponse* resp)
   resp->setJsonBody(doc);
 }
 
-static void apiServerZoneDetail(HttpRequest* req, HttpResponse* resp)
+static inline DNSName findZoneById(HttpRequest* req)
 {
-  DNSName zonename = apiZoneIdToName(req->parameters["id"]);
-
-  auto iter = SyncRes::t_sstorage.domainmap->find(zonename);
-  if (iter == SyncRes::t_sstorage.domainmap->end()) {
+  auto zonename = apiZoneIdToName(req->parameters["id"]);
+  if (SyncRes::t_sstorage.domainmap->find(zonename) == SyncRes::t_sstorage.domainmap->end()) {
     throw ApiException("Could not find domain '" + zonename.toLogString() + "'");
   }
+  return zonename;
+}
 
-  if (req->method == "PUT") {
-    Json document = req->json();
+static void apiServerZoneDetailPUT(HttpRequest* req, HttpResponse* resp)
+{
+  auto zonename = findZoneById(req);
+  const auto& document = req->json();
 
-    doDeleteZone(zonename);
-    doCreateZone(document);
-    reloadZoneConfiguration(g_yamlSettings);
-    resp->body = "";
-    resp->status = 204; // No Content, but indicate success
-  }
-  else if (req->method == "DELETE") {
-    if (!doDeleteZone(zonename)) {
-      throw ApiException("Deleting domain failed");
-    }
+  doDeleteZone(zonename);
+  doCreateZone(document);
+  reloadZoneConfiguration(g_yamlSettings);
+  resp->body = "";
+  resp->status = 204; // No Content, but indicate success
+}
 
-    reloadZoneConfiguration(g_yamlSettings);
-    // empty body on success
-    resp->body = "";
-    resp->status = 204; // No Content: declare that the zone is gone now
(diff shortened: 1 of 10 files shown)


cmouse:
Upstream PR https://github.com/cmouse/yahttp/pull/30

pdns/ws-auth.cc
```diff
@@ -1530,327 +1524,327 @@
...
+  fillZone(B, zonename, resp, req);
+  resp->status = 201;
+}
 
+static void apiServerZonesGET(HttpRequest* req, HttpResponse* resp) {
+  UeberBackend B;
+  DNSSECKeeper dk(&B);
   vector<DomainInfo> domains;
 
+  // FIXME: Shouldn't this fail with 404 if not found?
```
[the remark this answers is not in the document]
cmouse: @Habbie I'm wondering if the API should indeed fail with 404 when a zone is not found? Now it just returns empty 200 OK result.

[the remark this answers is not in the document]
cmouse: ping, this is last issue left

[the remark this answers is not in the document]
Habbie: empty 200 matches the [documentation](https://doc.powerdns.com/authoritative/http-api/zone.html):

> When set to the name of a zone, only this zone is returned. If no zone with that name exists, the response is an empty array. 

so we can keep that

Habbie:
This has >50 clang-tidy warnings

Habbie (commented):
Looks good to me! Will need a review from Otto as well. There's a conflict in ws-auth.cc currently. If Otto agrees, we can finally merge this (after that conflict)

pdns/recursordist/ws-recursor.cc
```diff
@@ -486,10 +485,6 @@ static void apiServerCacheFlush(HttpRequest* req, HttpResponse* resp)
 
 static void apiServerRPZStats(HttpRequest* req, HttpResponse* resp)
```
omoerbeek, line 486:
```suggestion
static void apiServerRPZStats(HttpRequest* /* req */, HttpResponse* resp)
```

omoerbeek: Same for `prometheusMetrics`

pdns/ws-auth.cc
```diff
@@ -928,14 +929,27 @@ static bool isValidMetadataKind(const string& kind, bool readonly) {
   return found;
 }
 
+// this is easier as macro since we need UeberBackend instance in most places
```
omoerbeek, line 932: Not terribly happy with this macro, but I do not see a way around it..

pdns/recursordist/ws-recursor.cc
```diff
@@ -86,91 +86,100 @@ static void apiWriteConfigFile(const string& filebasename, const string& content
   ofconf.close();
 }
 
-static void apiServerConfigACL(const std::string& aclType, HttpRequest* req, HttpResponse* resp)
+static void apiServerConfigACLGET(const std::string& aclType, HttpRequest* req __attribute__((unused)), HttpResponse* resp)
```
omoerbeek, line 89:
```suggestion
static void apiServerConfigACLGET(const std::string& aclType, HttpRequest* /* req */, HttpResponse* resp)
```

omoerbeek (changes requested):
A round of review done, will also do some basic tests

pdns/ws-auth.cc
Habbie: yes, I had the same feeling

pdns/ws-auth.cc
```diff
@@ -928,14 +929,27 @@ static bool isValidMetadataKind(const string& kind, bool readonly) {
...
+  DNSName zonename = apiZoneIdToName((req)->parameters["id"]); \
+  UeberBackend B; \
+  DNSSECKeeper dk(&B); \
+  DomainInfo di; \
+  try { \
+    if (!B.getDomainInfo(zonename, di)) { \
+      throw HttpNotFoundException(); \
+    } \
+  } catch(const PDNSException &e) { \
+    throw HttpInternalServerErrorException("Could not retrieve Domain Info: " + e.reason); \
```
zeha, line 944: what are we gaining from this throw?

omoerbeek: The handling of `HttpInternalServerErrorException` (which is derived from in `HttpException`) in `WebServer::handleRequest` is different, though the only difference I see is in what's being logged.

omoerbeek (commented):
Did a (basic) review with some nits. I also did some tests for rec and they seem to work as expected. As the regression tests also run fine, I am happy (especially as rec is in that start of a release cycle). Once the nits and the open question are answered, I'm fine with this.

pdns/recursordist/ws-recursor.cc
```diff
@@ -384,48 +389,46 @@ static void apiServerZones(HttpRequest* req, HttpResponse* resp)
...
 }
 
-static void apiServerZoneDetail(HttpRequest* req, HttpResponse* resp)
-{
-  DNSName zonename = apiZoneIdToName(req->parameters["id"]);
-
-  auto iter = SyncRes::t_sstorage.domainmap->find(zonename);
-  if (iter == SyncRes::t_sstorage.domainmap->end()) {
-    throw ApiException("Could not find domain '" + zonename.toLogString() + "'");
+#define findZoneById(req)                                                         \
```
omoerbeek, line 392: In the case of rec, this can  be a function returning a `DNSName`. The callers then become:
```
auto zonename = findZoneById(req);
```

omoerbeek (changes requested):

omoerbeek (approved):

PR #13154, opened 2023-08-18
dnsdist: Switch to pytest for our regression tests, run with in parallel

rgacogne:
### Short description

This PR switches our regression tests from `nose`, which is deprecated, to `pytest`.
It also uses the `xdist` `pytest` plugin to run our tests in parallel when more than one CPU core is available. To avoid concurrency issues in tests that were not designed to be run in parallel, we group tests by their containing file (`--dist loadfile`), and the tests have been updated to use dynamic UDP and TCP ports, based on the worker they are running on. This roughly cuts the time needed to run the regression tests in half on GH actions, where we have two cores. We could in theory do better with more cores but at least one file, the `DynBlock` one, is taking a fair amount of the total time so the gain is not immediate. It might be good to split that file in several smaller ones in the future.

### Checklist

I have:
- [ ] compiled this code
- [x] tested this code
- [ ] documented the code


regression-tests.dnsdist/test_Async.py
```diff
@@ -5,7 +5,7 @@
 import threading
 import unittest
 import dns
```
chbruyand, line 7:
```suggestion
import dns
import sys
```

chbruyand (approved):
Great work, this really speeds up the test suite and avoids using old python versions :)
Juste one nit as I hit a `sys.exit()` locally and this produces some warning.

regression-tests.dnsdist/test_Async.py
rgacogne: Nice one, thanks!

rgacogne:
`[test-auth-backend (authpy, coscale/docker-sleep)](https://github.com/PowerDNS/pdns/actions/runs/5998751172/job/16268739176#logs)` failure is unrelated to these changes, I'll ignore it.

PR #13156, opened 2023-08-18
dnsdist: Add a structured log-like logging option

### LINKED ISSUE #12987: dnsdist: Add log level to log messages sent to stdout
Currently, only log messages sent to syslog include the log level, otherwise the log message is simply output as per the following code:
https://github.com/PowerDNS/pdns/blob/4dabc73e392626c26ce90239bf8282fe8e708349/pdns/dolog.hh#L119C10-L119C10

It would be super-useful if this line prefixed the log level to the log message. This would be particularly useful in environments such as k8s. To avoid changing the default logging behaviour, this behaviour could be enabled with a configuration option.

rgacogne:
### Short description

<strike>Still a draft until we decide on an unified "format" for auth and rec as well.</strike>
Matches the format described in https://github.com/PowerDNS/pdns/pull/13062#issuecomment-1791031178

Closes #12987.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/pdns/dnsdistdist/docs/reference/config.rst
+++ b/pdns/dnsdistdist/docs/reference/config.rst
@@ -1203,6 +1203,27 @@ Status, Statistics and More
 
   * ``outputFile=path``: string - Write the output of the command to the supplied file, instead of the standard output.
 
+.. function:: setStructuredLogging(enable[, options])
+
+  .. versionadded:: 1.9.0
+
+  Set whether log messages should be in a structured-logging-like format. This is turned off by default.
+  The resulting format looks like this (when timestamps are enabled via ``--log-timestamps`` and with ``levelPrefix="prio"`` and ``timeFormat="ISO8601"``)::
+
+  tm="2023-11-06T12:04:58+0100" prio="info" msg="Added downstream server 127.0.0.1:53"
+
+  And with ``levelPrefix="level"`` and ``timeFormat="numeric"``)::
+
+  tm="1699268815.133" level="info" msg="Added downstream server 127.0.0.1:53"
+
+  :param bool enable: Set to true if you want to enable structured logging
+  :param table options: A table with key: value pairs with options described below.
+
+  Options:
+
+  * ``levelPrefix=prefix``: string - Set the prefix for the log level. Default is ``prio``.
+  * ``timeFormat=format``: string - Set the time format. Supported values are ``ISO8601`` and ``numeric``. Default is ``numeric``.
+
 .. function:: setVerbose(verbose)
 
   .. versionadded:: 1.8.0
--- a/pdns/dnsdistdist/docs/reference/config.rst
+++ b/pdns/dnsdistdist/docs/reference/config.rst
@@ -1210,11 +1210,11 @@ Status, Statistics and More
   Set whether log messages should be in a structured-logging-like format. This is turned off by default.
   The resulting format looks like this (when timestamps are enabled via ``--log-timestamps`` and with ``levelPrefix="prio"`` and ``timeFormat="ISO8601"``)::
 
-  tm="2023-11-06T12:04:58+0100" prio="info" msg="Added downstream server 127.0.0.1:53"
+  ts="2023-11-06T12:04:58+0100" prio="Info" msg="Added downstream server 127.0.0.1:53"
 
   And with ``levelPrefix="level"`` and ``timeFormat="numeric"``)::
 
-  tm="1699268815.133" level="info" msg="Added downstream server 127.0.0.1:53"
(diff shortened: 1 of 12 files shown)


rgacogne:
Note to self: ponder adding `vwarnlog` and `verrlog`, or something that looks a bit more pretty than these names, so that we can log at different levels in verbose mode for things that are actually not informational but that we don't want to log for every query by default either.

pdns/dnsdistdist/docs/reference/config.rst
```diff
@@ -1196,6 +1196,27 @@ Status, Statistics and More
...
   * ``outputFile=path``: string - Write the output of the command to the supplied file, instead of the standard output.
 
+.. function:: setStructuredLogging(enable[, options])
+
+  .. versionadded:: 1.9.0
+
+  Set whether log messages should be in a structured-logging-like format. This is turned off by default.
+  The resulting format looks like this (when timestamps are enabled via ``--log-timestamps`` and with ``levelPrefix="prio"`` and ``timeFormat="ISO8601"``)::
+
+  tm="2023-11-06T12:04:58+0100" prio="info" msg="Added downstream server 127.0.0.1:53"
```
omoerbeek, line 1206: rec uses "ts", see https://docs.powerdns.com/recursor/appendices/structuredlogging.html
I 'd prefer if the keywords would match at least.
Also, the prio value are capitalized in rec. But this is debatable. The idea was to conform to other products, but we have not decided on the common terms. So this is good as a start in an alpha release, but it might need some changes in the future.

pdns/dnsdistdist/dolog.cc
```diff
@@ -0,0 +1,96 @@
...
+void logTime(std::ostream& stream)
+{
+  std::array<char, 50> buffer{""};
+
+  if (LoggingConfiguration::getStructuredLogging() && LoggingConfiguration::getStructuredLoggingTimeFormat() == LoggingConfiguration::TimeFormat::Numeric) {
+    struct timeval now
+    {
+    };
+    gettimeofday(&now, nullptr);
+    snprintf(buffer.data(), buffer.size(), "%lld.%03ld", static_cast<long long>(now.tv_sec), now.tv_usec / 1000);
```
omoerbeek, line 62: On some systems (ie..macOS `suseconds_t` is an `int32_t`...
```suggestion
    snprintf(buffer.data(), buffer.size(), "%lld.%03ld", static_cast<long long>(now.tv_sec), static_cast<long>(now.tv_usec / 1000));
```

pdns/dnsdistdist/dolog.cc
```diff
@@ -0,0 +1,96 @@
...
+      auto lock = std::lock_guard(mutex);
+
+      if (strftime(buffer.data(), buffer.size(), timeFormat, &localNow) == 0) {
+        buffer[0] = '\0';
+      }
+    }
+  }
+
+  if (dnsdist::logging::LoggingConfiguration::getStructuredLogging()) {
+    stream << "tm=\"" << buffer.data() << "\" ";
```
omoerbeek, line 89: better use `std::quoted()` as `buffer.data()` might contain quotes itself (not atm, but maybe someday)

pdns/dnsdistdist/docs/reference/config.rst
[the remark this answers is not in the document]
rgacogne: Agreed, but we will tackle moving to an actual structured format in 1.10 at best, not 1.9.

rgacogne: > rec uses "ts", see https://docs.powerdns.com/recursor/appendices/structuredlogging.html I 'd prefer if the keywords would match at least.

It was my intent but I somehow confused myself, good catch!

> Also, the prio value are capitalized in rec. But this is debatable. 

Only for the `systemd-journal` backend, unless I'm mistaken? Do you think it would make sense for dnsdist to make it configurable as well?

> The idea was to conform to other products, but we have not decided on the common terms. So this is good as a start in an alpha release, but it might need some changes in the future.

Understood. I used https://github.com/PowerDNS/pdns/pull/13062 to determine the current common terms, which is still a work in progress. It would be nice to get together and decide on the common terms in the near future :)

omoerbeek: I mean initial cap. So "Info", not "info" or "INFO" in rec.

rgacogne: My bad, I read `key` not `value`. I don't really have an opinion so I'll align on what the rec is doing.

rgacogne:
Rebased on master to fix a conflict, and added a few commits addressing Otto's comments (many thanks!).

omoerbeek (approved):

PR #13166, opened 2023-08-23
rec/buildbit: Avoid running SNMP tests on the current buildbot setup

omoerbeek:
Since the change from nose to pytest, the SNMP exclusion via the environment var NOSE_EXLUDE does no longer work.

### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code


build-scripts/test-recursor
```diff
@@ -59,7 +59,13 @@ sleep 3
 ./clean.sh
 
 cd ../regression-tests.recursor-dnssec
-./runtests $@ || EXIT=1
+
+EXTRA_ARG=""
+if [ $PWD = /srv/buildbot-worker/test-rec-debian-buster/build ]; then
```
Habbie, line 64: I wonder, given the `cd` above, this is still the right `$PWD`

omoerbeek: You wondered rightfully...

Habbie (approved):

PR #13181, opened 2023-08-28
Fix provenance generation for our packages

rgacogne:
### Short description

- [Fix the generation of packages JSON list for provenance](https://github.com/PowerDNS/pdns/commit/99c370a2c30ae7152791d852fb7b37280ac35b22): the conditions were broken because our templating system supports a very limited syntax
- [build-packages: Fix 'check if hashes were created for all requested targets](https://github.com/PowerDNS/pdns/commit/00feba47c530579440085ba16248adcf4c2cc71f): If the 'inputs.os' list ends with a newline the check fails because the `for` loop syntax is wrong.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/.github/workflows/build-packages.yml
+++ b/.github/workflows/build-packages.yml
@@ -14,7 +14,7 @@ on:
         type: string
         # please remember to update the pkghashes below when you
         # update this list, as well as the one in builder-dispatch.yml
-        default: >
+        default: >-
           el-7
           el-8
           el-9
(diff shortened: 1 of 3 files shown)


romeroalx:
Looks good :)

My only suggestion is to add also the modification for stripping the new line character at the end (`>-`), here: <https://github.com/PowerDNS/pdns/blob/master/.github/workflows/builder-dispatch.yml#L18>

rgacogne:
> My only suggestion is to add also the modification for stripping the new line character at the end (`>-`), here: https://github.com/PowerDNS/pdns/blob/master/.github/workflows/builder-dispatch.yml#L18

Good one, thanks!

PR #13191, opened 2023-08-29
Wrap `DIR*` objects in unique pointers to prevent memory leaks

rgacogne:
### Short description

As far as I can tell there is no actual issue with the current code, but let's make sure that an exception cannot cause a memory leak, for example.

### Checklist

I have:
- [x] compiled this code
- [ ] tested this code
- [ ] documented the code


modules/geoipbackend/geoipbackend.cc
```diff
@@ -85,12 +85,12 @@ GeoIPBackend::GeoIPBackend(const string& suffix)
...
   if (!getArg("dnssec-keydir").empty()) {
-    DIR* dir = opendir(getArg("dnssec-keydir").c_str());
-    if (dir == nullptr) {
+    auto dirHandle = std::unique_ptr<DIR, decltype(&closedir)>(opendir(getArg("dnssec-keydir").c_str()), closedir);
+    if (!dirHandle) {
       throw PDNSException("dnssec-keydir " + getArg("dnssec-keydir") + " does not exist");
     }
     d_dnssec = true;
-    closedir(dir);
+    dirHandle.reset();
```
fredmorcos, line 93: Isn't `dirHandle` going out of scope here? Why do you need the call to `reset()`?

pdns/dnsdist-lua.cc
```diff
@@ -1861,16 +1861,18 @@ static void setupLuaConfig(LuaContext& luaCtx, bool client, bool configCheck)
       return;
     }
 
-    DIR* dirp;
-    struct dirent* ent;
+    auto dirp = std::unique_ptr<DIR, decltype(&closedir)>(opendir(dirname.c_str()), closedir);
```
fredmorcos, line 1864: `dirHandle` would be a better name here.

pdns/dnsdist-lua.cc
```diff
@@ -1887,7 +1889,7 @@ static void setupLuaConfig(LuaContext& luaCtx, bool client, bool configCheck)
       }
     }
 
-    closedir(dirp);
+    dirp.reset();
```
fredmorcos, line 1892: I'm fine with the call to `reset()` here, but I have a slight preference for adding a dummy scope instead, or separating this part into a function altogether.

pdns/ixfrdist.cc
```diff
@@ -202,20 +202,22 @@ static bool sortSOA(uint32_t i, uint32_t j) {
...
+    // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-array-to-pointer-decay): this is what dirent is
+    if (strcmp(entry->d_name, ".") == 0 || strcmp(entry->d_name, "..") == 0) {
       continue;
     }
-    zoneVersions.push_back(std::stoi(d->d_name));
+    //  NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-array-to-pointer-decay): this is what dirent is
+    zoneVersions.push_back(std::stoi(entry->d_name));
   }
-  closedir(dp);
+  dirHandle.reset();
```
fredmorcos, line 220: Same here.

I feel this routine to open a directory and do something with the entries inside could be moved to the `pdns::` namespace in `misc.cc/hh`, and it could accept a lambda for the condition and/or a closure for taking some action on entries.

pdns/misc.cc
```diff
@@ -1377,23 +1377,27 @@ DNSName getTSIGAlgoName(TSIGHashEnum& algoEnum)
 uint64_t getOpenFileDescriptors(const std::string&)
 {
 #ifdef __linux__
-  DIR* dirhdl=opendir(("/proc/"+std::to_string(getpid())+"/fd/").c_str());
-  if(!dirhdl)
+  auto dirhdl = std::unique_ptr<DIR, decltype(&closedir)>(opendir(("/proc/"+std::to_string(getpid())+"/fd/").c_str()), closedir);
```
fredmorcos, line 1380: `dirHandle`.

We have enough disk space and auto-completion 🙃

fredmorcos (approved):
Nothing blocking, but if you decide to do that refactoring into `misc.cc/hh` that I mention in one of the comments, please re-request a review.

pdns/misc.cc
rgacogne: Fair enough :smile: :+1:

pdns/ixfrdist.cc
rgacogne: Sounds like a good idea, I'll give it a try later today. If it leads to a big commit I might merge this PR first and open a new one.

modules/geoipbackend/geoipbackend.cc
rgacogne: It's a good question: I tried to keep my changes restricted to the smart pointer move as much as possible to avoid any side-effects, but I agree with you in this case it is probably useless. I'll take a second look, thanks!

rgacogne:
I pushed a fix resolving most of the comments, except for the one suggesting a generic function to visit a directory: I'll have a look at that later today in a separate PR.

pdns/dnsdist-lua.cc
```diff
@@ -1861,33 +1861,36 @@ static void setupLuaConfig(LuaContext& luaCtx, bool client, bool configCheck)
...
+        if (ent->d_name[0] == '.') {
           continue;
         }
 
-        files.push_back(namebuf.str());
+        if (boost::ends_with(ent->d_name, ".conf")) {
+          std::ostringstream namebuf;
+          namebuf << dirname << "/" << ent->d_name;
+
+          if (stat(namebuf.str().c_str(), &st) || !S_ISREG(st.st_mode)) {
```
fredmorcos, line 1885: `stat(3)` returns an `int`.

PR #13195, opened 2023-08-30
rec: move tcp-in processing to dedicated thread(s)

### LINKED ISSUE #8394: rec: Separate TCP acceptor thread
- Program: Recursor
 - Issue type: Feature request

### Short description
I believe we should ponder moving the code responsible for accepting new TCP connections, and handling them, to a separate thread. 
The current situation means that a new TCP connections might have to wait until `max-udp-queries-per-round` queries have been processed before it can be accepted, which is fine as long as you don't get too many TCP queries, and they are not too latency-sensitive, but this might change.
The obvious issue with doing that is that this new thread would no longer share the query cache (the packet cache is currently irrelevant for TCP, which is also something that we might consider changing at sme point). So perhaps we need to wait until some sort of shared query cache has been implemented before considering this.

omoerbeek:
### Short description

The new TCP worker thread(s) are listeners and workers for incoming TCP only. As all the caches are shared these days there is no reason to handle them in regular distributor threads or workers.

To simplify matters a bit, always run at least one TCP Listener/worker thread. This has the consequence that regular worker(s) do no longer do TCP (if pdns-distributes-queries is false) and neither do the distributor thread(s) (if pdns-distributes-qeuries is true).

Basic test done on macOS, with all the settings combinations (which are plenty!). Needs:

- [ ] validation on a loaded machine on various platforms and various settings
- [ ] A config settings

and docs. Hence draft status.

Fixes #8394

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [X] documented the code

### DIFF
--- a/pdns/recursordist/docs/performance.rst
+++ b/pdns/recursordist/docs/performance.rst
@@ -80,10 +80,14 @@ MTasker and MThreads
 PowerDNS Recursor uses a cooperative multitasking in userspace called ``MTasker``, based either on ``boost::context`` if available, or on ``System V ucontexts`` otherwise. For maximum performance, please make sure that your system supports ``boost::context``, as the alternative has been known to be quite slower.
 
 The maximum number of simultaneous MTasker threads, called ``MThreads``, can be tuned via :ref:`setting-max-mthreads`, as the default value of 2048 might not be enough for large-scale installations.
+This number limits the number of mthreads *per physical (Posix) thread*.
+The threads that create mthreads are the distributor and worker threads.
 
 When a ``MThread`` is started, a new stack is dynamically allocated for it on the heap. The size of that stack can be configured via the :ref:`setting-stack-size` parameter, whose default value is 200 kB which should be enough in most cases.
 
-To reduce the cost of allocating a new stack for every query, the recursor can cache a small amount of stacks to make sure that the allocation stays cheap. This can be configured via the :ref:`setting-stack-cache-size` setting. The only trade-off of enabling this cache is a slightly increased memory consumption, at worst equals to the number of stacks specified by :ref:`setting-stack-cache-size` multiplied by the size of one stack, itself specified via :ref:`setting-stack-size`.
+To reduce the cost of allocating a new stack for every query, the recursor can cache a small amount of stacks to make sure that the allocation stays cheap. This can be configured via the :ref:`setting-stack-cache-size` setting.
+This limit is per physcial (Posix) thread.
+The only trade-off of enabling this cache is a slightly increased memory consumption, at worst equals to the number of stacks specified by :ref:`setting-stack-cache-size` multiplied by the size of one stack, itself specified via :ref:`setting-stack-size`.
 
 Performance tips
 ----------------
@@ -177,8 +181,13 @@ Each of the queries processed will consume an mthread until processing is done.
 A response to a query is sent immediately when it becomes available; the response can be sent before other responses to queries that were received earlier by the Recursor.
 This is the Out-of-Order feature which greatly enhances performance, as a single slow query does not prevent other queries to be processed.
 
+Before version 5.0.0, TCP queries are processed by either the distributer thread(s) if :ref:`setting-pdns-distributes-queries` is true, or by worker threads if :ref:`setting-pdns-distributes-queries` is false.
+Starting with version 5.0.0, :program:`Recursor` has dedicated thread(s) processing TCP queries.
+
 The maximum number of mthreads consumed by TCP queries is :ref:`setting-max-tcp-clients` times :ref:`setting-max-concurrent-requests-per-tcp-connection`.
-This number should be (much) lower than :ref:`setting-max-mthreads`, to also allow UDP queries to be handled as these also consume mthreads.
+If :ref:`setting-pdns-distributes-queries` is true, this number should be (much) lower than :ref:`setting-max-mthreads`, to also allow UDP queries to be handled as these also consume mthrea ds.
+Note that :ref:`setting-max-mthreads` is a per Posix thread setting.
+This means that the global maximum number of mthreads  is (#distributor threads + #worker threads) * max-mthreads.
 
 If you expect few clients, you can increase :ref:`setting-max-concurrent-requests-per-tcp-connection`, to allow more concurrency per TCP connection.
 If you expect many clients and you have increased :ref:`setting-max-tcp-clients`, reduce :ref:`setting-max-concurrent-requests-per-tcp-connection` number to prevent mthread starvation or increase the maximum number of mthreads.
@@ -188,6 +197,7 @@ To see the current number of mthreads in use consult the :ref:`stat-concurrent-q
 If a query could not be handled due to mthread shortage, the :ref:`stat-over-capacity-drops` metric is increased.
 
 As an example, if you have typically 200 TCP clients, and the default maximum number of mthreads of 2048, a good number of concurrent requests per TCP connection would be 5. Assuming a worst case packet cache hit ratio, if all 200 TCP clients fill their connections with queries, about half (5 * 200) of the mthreads would be used by incoming TCP queries, leaving the other half for incoming UDP queries.
+Note that starting with versino 5.0.0, TCP queries are processed by dedicated TCP thread(s), so the sharing of mthreads between UDP and TCP queries no longer applies.
 
 The total number of incoming TCP connections is limited by :ref:`setting-max-tcp-clients`.
 There is also a per client address limit: :ref:`setting-max-tcp-per-client` to limit the impact of a single client.
--- a/pdns/recursordist/docs/performance.rst
+++ b/pdns/recursordist/docs/performance.rst
@@ -80,13 +80,13 @@ MTasker and MThreads
 PowerDNS Recursor uses a cooperative multitasking in userspace called ``MTasker``, based either on ``boost::context`` if available, or on ``System V ucontexts`` otherwise. For maximum performance, please make sure that your system supports ``boost::context``, as the alternative has been known to be quite slower.
 
 The maximum number of simultaneous MTasker threads, called ``MThreads``, can be tuned via :ref:`setting-max-mthreads`, as the default value of 2048 might not be enough for large-scale installations.
-This number limits the number of mthreads *per physical (Posix) thread*.
+This setting limits the number of mthreads *per physical (Posix) thread*.
 The threads that create mthreads are the distributor and worker threads.
 
 When a ``MThread`` is started, a new stack is dynamically allocated for it on the heap. The size of that stack can be configured via the :ref:`setting-stack-size` parameter, whose default value is 200 kB which should be enough in most cases.
 
 To reduce the cost of allocating a new stack for every query, the recursor can cache a small amount of stacks to make sure that the allocation stays cheap. This can be configured via the :ref:`setting-stack-cache-size` setting.
-This limit is per physcial (Posix) thread.
+This limit is per physical (Posix) thread.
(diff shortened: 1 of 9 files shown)


omoerbeek:
The bulk tests looks broken for the single thread case. Locally (on macOS) it does seem to work, investigating....

omoerbeek:
Bulk test fixed, and in the meantime did a few dnsperf tests on Debian0-bookworm.

pdns/recursordist/docs/performance.rst
```diff
@@ -80,10 +80,14 @@ MTasker and MThreads
 PowerDNS Recursor uses a cooperative multitasking in userspace called ``MTasker``, based either on ``boost::context`` if available, or on ``System V ucontexts`` otherwise. For maximum performance, please make sure that your system supports ``boost::context``, as the alternative has been known to be quite slower.
 
 The maximum number of simultaneous MTasker threads, called ``MThreads``, can be tuned via :ref:`setting-max-mthreads`, as the default value of 2048 might not be enough for large-scale installations.
+This number limits the number of mthreads *per physical (Posix) thread*.
```
rgacogne, line 83:
```suggestion
This setting limits the number of mthreads *per physical (Posix) thread*.
```

pdns/recursordist/docs/performance.rst
```diff
@@ -80,10 +80,14 @@ MTasker and MThreads
...
+The threads that create mthreads are the distributor and worker threads.
 
 When a ``MThread`` is started, a new stack is dynamically allocated for it on the heap. The size of that stack can be configured via the :ref:`setting-stack-size` parameter, whose default value is 200 kB which should be enough in most cases.
 
-To reduce the cost of allocating a new stack for every query, the recursor can cache a small amount of stacks to make sure that the allocation stays cheap. This can be configured via the :ref:`setting-stack-cache-size` setting. The only trade-off of enabling this cache is a slightly increased memory consumption, at worst equals to the number of stacks specified by :ref:`setting-stack-cache-size` multiplied by the size of one stack, itself specified via :ref:`setting-stack-size`.
+To reduce the cost of allocating a new stack for every query, the recursor can cache a small amount of stacks to make sure that the allocation stays cheap. This can be configured via the :ref:`setting-stack-cache-size` setting.
+This limit is per physcial (Posix) thread.
```
rgacogne, line 89:
```suggestion
This limit is per physical (Posix) thread.
```

pdns/recursordist/docs/performance.rst
```diff
@@ -177,8 +181,13 @@ Each of the queries processed will consume an mthread until processing is done.
 A response to a query is sent immediately when it becomes available; the response can be sent before other responses to queries that were received earlier by the Recursor.
 This is the Out-of-Order feature which greatly enhances performance, as a single slow query does not prevent other queries to be processed.
 
+Before version 5.0.0, TCP queries are processed by either the distributer thread(s) if :ref:`setting-pdns-distributes-queries` is true, or by worker threads if :ref:`setting-pdns-distributes-queries` is false.
+Starting with version 5.0.0, :program:`Recursor` has dedicated thread(s) processing TCP queries.
+
 The maximum number of mthreads consumed by TCP queries is :ref:`setting-max-tcp-clients` times :ref:`setting-max-concurrent-requests-per-tcp-connection`.
-This number should be (much) lower than :ref:`setting-max-mthreads`, to also allow UDP queries to be handled as these also consume mthreads.
+If :ref:`setting-pdns-distributes-queries` is true, this number should be (much) lower than :ref:`setting-max-mthreads`, to also allow UDP queries to be handled as these also consume mthrea ds.
```
rgacogne, line 188:
```suggestion
If :ref:`setting-pdns-distributes-queries` is true, this number should be (much) lower than :ref:`setting-max-mthreads`, to also allow UDP queries to be handled as these also consume mthreads.
```

pdns/recursordist/rec-main.cc
```diff
@@ -104,7 +104,8 @@ std::shared_ptr<Logr::Logger> g_slogudpin;
 std::shared_ptr<Logr::Logger> g_slogudpout;
 
 /* without reuseport, all listeners share the same sockets */
-deferredAdd_t g_deferredAdds;
+static deferredAdd_t g_deferredAdds;
+static deferredAdd_t g_deferredTCPAdds;
```
rgacogne, line 108: Just a nit but the `g_` prefix usually indicates globals, so we might want to rename these.
```suggestion
static deferredAdd_t s_deferredAdds;
static deferredAdd_t s_deferredTCPAdds;
```

pdns/recursordist/rec-main.cc
```diff
@@ -1687,50 +1721,45 @@ static void initDistribution(Logr::log_t log)
...
-        auto& deferredAdds = info.deferredAdds;
-        auto& tcpSockets = info.tcpSockets;
+      for (unsigned int i = 0; i < RecThreadInfo::numWorkers(); i++, threadNum++) {
+        auto& info = RecThreadInfo::info(threadNum);
+        auto& deferredAdds = info.getDeferredAdds();
         makeUDPServerSockets(deferredAdds, log);
-        makeTCPServerSockets(deferredAdds, tcpSockets, log);
       }
     }
+    threadNum = 1 + RecThreadInfo::numDistributors() + RecThreadInfo::numWorkers();
```
rgacogne, line 1744: Isn't that already the case?

pdns/recursordist/rec-main.hh
```diff
@@ -436,6 +444,11 @@ public:
     return s_numWorkerThreads;
   }
 
+  static unsigned int numTCPWorkers()
```
rgacogne, line 447: Would it make sense to rename `numWorkers()` to `numUDPWorkers()?`

pdns/recursordist/rec-main.hh
```diff
@@ -456,24 +469,70 @@ public:
...
+  {
+    tcpSockets = socks;
+  }
+
+  deferredAdd_t& getDeferredAdds()
+  {
+    return deferredAdds;
+  }
+
+  ThreadPipeSet& getPipes()
```
rgacogne, line 510:
```suggestion
  const ThreadPipeSet& getPipes() const
```

pdns/recursordist/rec-tcp.cc
```diff
@@ -27,6 +27,36 @@
 #include "mplexer.hh"
 #include "uuid-utils.hh"
 
+// When pdns_distributes query is false with reuseport true (the default since 4.9.0), TCP queries
```
rgacogne, line 30:
```suggestion
// When pdns-distributes-queries is false with reuseport true (the default since 4.9.0), TCP queries
```

pdns/recursordist/rec-tcp.cc
```diff
@@ -27,6 +27,36 @@
...
+// hit. If that does not produce a hit, the query is passed to an mthread in the same way as with
+// UDP queries, but do note that the mthread processing is serviced by the distributor thread. The
+// final answer will be sent by the same distributor thread that originally picked up the query.
+//
+// Changing this, and having incoming TCP queries handled by worker threads is somewhat more complex
+// than UDP, as the socket must remain avaiable in the distributor thread (for reading more
+// queries), but the TCP socket must also be passed to a worker thread so it can write its
+// answer. The in-flight bookkeeping also has to be aware of how a query is handled to do the
+// accounting properly. I am not sure if changing the current setup is worth all this trouble,
+// especilly since the default is now to not use pdns-distributes-queries, which works well in many
```
rgacogne, line 50:
```suggestion
// especially since the default is now to not use pdns-distributes-queries, which works well in many
```

pdns/recursordist/rec-main.cc
```diff
@@ -1303,7 +1337,7 @@ void broadcastFunction(const pipefunc_t& func)
   }
 
   unsigned int thread = 0;
-  for (const auto& threadInfo : RecThreadInfo::infos()) {
+  for (auto& threadInfo : RecThreadInfo::infos()) {
```
rgacogne, line 1340: If we make `getPipes()` `const`:
```suggestion
  for (const auto& threadInfo : RecThreadInfo::infos()) {
```

pdns/recursordist/docs/performance.rst
omoerbeek: Note to self: double check if this `true` should actually be `false`, as currently (before this PR) TCP queries are processes by distributor thread(s) if there are any, and only if `setting-pdns-distributes-queries` is false the workers will process both TCP and UDP queries.

pdns/recursordist/rec-main.cc
```diff
@@ -2074,6 +2103,12 @@ static int serviceMain(Logr::log_t log)
          log->info(Logr::Warning, "Asked to run with 0 threads, raising to 1 instead"));
     RecThreadInfo::setNumWorkerThreads(1);
   }
+  RecThreadInfo::setNumTCPWorkerThreads(1); // XXX
```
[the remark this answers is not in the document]
omoerbeek: I am for the new settings code to get merged to avoid causing a conflict introducing a TCP threads setting.

pdns/recursordist/rec-main.cc
```diff
@@ -218,80 +220,110 @@ void RecThreadInfo::start(unsigned int tid, const string& tname, const std::map<
 int RecThreadInfo::runThreads(Logr::log_t log)
 {
   int ret = EXIT_SUCCESS;
-  unsigned int currentThreadId = 1;
   const auto cpusMap = parseCPUMap(log);
 
-  if (RecThreadInfo::numDistributors() + RecThreadInfo::numWorkers() == 1) {
+  if (RecThreadInfo::numDistributors() + RecThreadInfo::numUDPWorkers() == 1) {
     SLOG(g_log << Logger::Warning << "Operating with single distributor/worker thread" << endl,
```
rgacogne, line 226: Perhaps this should be `Operating with single UDP distributor/worker thread` now?

pdns/recursordist/rec-tcp.cc
```diff
@@ -27,6 +27,35 @@
...
+// In the case of pdns-distributes-queries true and reuseport false the queries are read and
+// initially processed by the distributor thread(s).
+//
+// Initial processing consist of parsing, calling gettag and checking if we have a packet cache
+// hit. If that does not produce a hit, the query is passed to an mthread in the same way as with
+// UDP queries, but do note that the mthread processing is serviced by the distributor thread. The
+// final answer will be sent by the same distributor thread that originally picked up the query.
+//
+// Changing this, and having incoming TCP queries handled by worker threads is somewhat more complex
+// than UDP, as the socket must remain avaiable in the distributor thread (for reading more
```
rgacogne, line 46:
```suggestion
// than UDP, as the socket must remain available in the distributor thread (for reading more
```

pdns/recursordist/rec-tcp.cc
```diff
@@ -27,6 +27,35 @@
...
+// answer. The in-flight bookkeeping also has to be aware of how a query is handled to do the
+// accounting properly. I am not sure if changing the current setup is worth all this trouble,
+// especially since the default is now to not use pdns-distributes-queries, which works well in many
+// cases.
+//
+// The drawback mentioned in https://github.com/PowerDNS/pdns/issues/8394 are not longer true, so an
+// alternative approach would be to introduce dedicated TCP worker thread(s).
+//
+// And this approach was implemented in https://github.com/PowerDNS/pdns/pull/13195. The distributor
+// and worker thread(s) now no longe process TCP queries.
```
rgacogne, line 57:
```suggestion
// and worker thread(s) now no longer process TCP queries.
```

rgacogne (approved):
A few nits but nothing that would prevent merging this PR. Nice work!

PR #13202, opened 2023-09-01
asan: Compile with -fsanitize-address-use-after-return=never if ASAN is enabled

omoerbeek:
When compiling with ASAN, starting with clang version 15 the default on linux for
`ASAN_OPTIONS=detect_stack_use_after_return` is 1. This does not work well with the mthread code. At
this point in time, it is unknown if that is an issue with our usage of
`__sanitizer_start_switch_fiber()` and `__sanitizer_finish_switch_fiber()` or something else. But as
these `__sanitizer` functions are barely documented, it's hard find out what exactly is going on. So
as a workaround, completely disable `detect_stack_use_after_return` by compiling with
`-fsanitize-address-use-after-return=never` if ASAN is enabled.

It might be better to have this option only for the recursor build, but I did not see a quick way to do that.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [X] documented the code

### DIFF
--- a/m4/pdns_enable_sanitizers.m4
+++ b/m4/pdns_enable_sanitizers.m4
@@ -26,6 +26,14 @@ AC_DEFUN([PDNS_ENABLE_SANITIZERS], [
   ])
 ])
 
+# When compiling with ASAN, starting with clang version 15 the default on linux for
+# ASAN_OPTIONS=detect_stack_use_after_return is 1. This does not work well with the mthread code. At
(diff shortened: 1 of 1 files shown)


omoerbeek:
Thinking about this some more, setting a specific ASAN_OPTION is better than using the big hammer approach. I'm dropping this PR. `detect_stack_use_after_return` should be set to 0 for runtime environments where the default changed until we find out what is going on with our `__sanitizer_finish_switch_fiber()` use.

PR #13205, opened 2023-09-02
auth: catalog, include groups in hash calculation

### LINKED ISSUE #13204: Catalog group editing does not increment the SOA serial in the catalog zone
- Program: Authoritative 
 - Issue type: Bug report

### Short description
If a catalog group is modified in a member zone, the SOA serial is not incremented in the corresponding catalog zone and slaves are not notified of that change.

### Environment

 - Operating system: Ubuntu 20.04.6
 - Software version: 4.7.4 (PostgreSQL backend is used)

### Steps to reproduce

1. Set up an environment with a catalg zone and a member zone
2. Let slave servers configure member zone (optional)
3. Change the group in the member zone, e.g. via pdnsutil
    ```
    pdnsutil set-option example.com producer group testgroup
    ```

### Expected behaviour
The SOA serial in the catalog zone should be incremented in the same way as it is when listing or delisting a member zone.

### Actual behaviour
SOA serial is not incremented and these changes are not notified to slave servers.

mind04:
### Short description
fixes #13204

### Checklist
I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/auth-catalogzone.cc
+++ b/pdns/auth-catalogzone.cc
@@ -107,7 +107,7 @@ std::string CatalogInfo::toJson() const
 
 void CatalogInfo::updateHash(CatalogHashMap& hashes, const DomainInfo& di) const
 {
-  hashes[di.catalog].process(std::to_string(di.id) + di.zone.toLogString() + "\0" + d_coo.toLogString() + "\0" + d_unique.toLogString());
+  hashes[di.catalog].process(std::to_string(di.id) + di.zone.toLogString() + string("\0", 1) + d_coo.toLogString() + string("\0", 1) + d_unique.toLogString());
   for (const auto& group : d_group) {
     hashes[di.catalog].process(std::to_string(group.length()) + group);
   }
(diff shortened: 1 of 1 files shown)


pdns/auth-catalogzone.cc
```diff
@@ -108,6 +108,9 @@ std::string CatalogInfo::toJson() const
 void CatalogInfo::updateHash(CatalogHashMap& hashes, const DomainInfo& di) const
 {
   hashes[di.catalog].process(static_cast<char>(di.id) + di.zone.toLogString() + "\0" + d_coo.toLogString() + "\0" + d_unique.toLogString());
+  for (const auto& group : d_group) {
+    hashes[di.catalog].process(static_cast<char>(group.length()) + group);
```
omoerbeek, line 112: I find this a somewhat obscure construct. First I had to realize is `char + string` and not `char + char *`, which wold not work at all as expected. Why not use `std::to_string(group(length())`? Also, I do wonder why the length is explicitly hashed.
Similar issue with `static_cast<char>(di.id)` above. If `di.id` does not fit into a char you loose significant bits.

omoerbeek: Last comment: is it wise to create all the tmp strings? You could hash the individual parts separately, working with the raw presentation: `process(di.zone.getStorage())` etc...

Habbie (approved):

Habbie:
rebased

PR #13208, opened 2023-09-04
remove make_tuple and make_pair

neheb:
Not needed

### Short description
This is C++17. No need for these.

### Checklist

I have:
- [ x] compiled this code
- [ ] tested this code
- [ ] documented the code


zeha:
two places where clang-tidy complains:
* avoid repeating the return type from the declaration; use a braced initializer list instead (modernize-return-braced-init-list - Level=Warning)

neheb:
still failing.

omoerbeek:
> still failing.

should be fixed

neheb:
fixed indeed. one "improvement" would be to replace std::pair with {}. Stylistic difference. The clang-tidy checks is only for returns.

rgacogne (approved):

PR #13209, opened 2023-09-05
rec: make aggressive cache pruning more effective and more fair

### LINKED ISSUE #13109: rec: address the aggresive cache pruning issues mentioned in #13106
In some cases) pruning is unfair (it scans the zones always in the same order and stops clearing when it has reached the goal) and/or not very active (when the recursor is lightly loaded). this has the consequence that old expired records can remain in the cache that prevent new data to be recorded and used (unless #13106 is merged).

On a lightly loaded rec I see this with `rec_control dump-cache -`:
```
; Zone .
. -36987 IN NSEC aaa. NS SOA RRSIG NSEC DNSKEY
```
Note that the root zone is last in the list, so entries in it are most likely to be skipped when pruning without pressure.

The unfairness should be fixed, possibly using recent changes in ``cachecleaner.hh`` as a guide.

There's also another issue with pruning, I see on another rec in the aggressive cache:
```
...
; Zone nl.
; Zone org.
; Zone pro.
; Zone pub.
; Zone tv.
; Zone co.uk.
; Zone video.
; Zone world.
; Zone .
...
```
from looking at the pruning code these empty zones should have been cleaned up. Maybe they are created empty?

omoerbeek:
### Short description

There are several issues with the current pruning approach. See #13109.

A consequence of the issues mentioned there is that the aggressive cache size can remain "too big" after a pruning run, causing the next pruning run also to scan many items and still do nothing. 

The hypothesis is that we're seeing bad consequences if the ineffective pruning since #13106, as a typical aggressive cache now contains way less expired items than before, causing it to hit the "bad case" more often. We have to confirm this, however. If we get confirmation, this must be backported to the 4.9.x branch.

Should fix #13109.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [X] documented the code

### DIFF
--- a/pdns/recursordist/test-aggressive_nsec_cc.cc
+++ b/pdns/recursordist/test-aggressive_nsec_cc.cc
@@ -1076,6 +1076,52 @@ BOOST_AUTO_TEST_CASE(test_aggressive_nsec3_wildcard_synthesis)
   BOOST_CHECK_EQUAL(queriesCount, 5U);
 }
 
+BOOST_AUTO_TEST_CASE(test_aggressive_nsec_replace)
+{
+  const size_t testSize = 10000;
+  auto cache = make_unique<AggressiveNSECCache>(testSize);
+
+  struct timeval now{};
+  Utility::gettimeofday(&now, nullptr);
+
+  vector<DNSName> names;
+  names.reserve(testSize);
+  for (size_t i = 0; i < testSize; i++) {
+    names.emplace_back(std::to_string(i) + "powerdns.com");
+  }
+
+  DTime time;
+  time.set();
+
+  for (const auto& name : names) {
+    DNSRecord rec;
+    rec.d_name = name;
+    rec.d_type = QType::NSEC3;
+    rec.d_ttl = now.tv_sec + 10;
+    rec.setContent(getRecordContent(QType::NSEC3, "1 0 500 ab HASG==== A RRSIG NSEC3"));
+    auto rrsig = std::make_shared<RRSIGRecordContent>("NSEC3 5 3 10 20370101000000 20370101000000 24567 dummy. data");
+    cache->insertNSEC(DNSName("powerdns.com"), rec.d_name, rec, {rrsig}, true);
+  }
+  auto diff1 = time.udiff(true);
+
+  BOOST_CHECK_EQUAL(cache->getEntriesCount(), testSize);
+  for (const auto& name : names) {
+    DNSRecord rec;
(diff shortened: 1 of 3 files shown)


pdns/recursordist/test-aggressive_nsec_cc.cc
```diff
@@ -1152,26 +1200,21 @@ BOOST_AUTO_TEST_CASE(test_aggressive_nsec_pruning)
...
+  rec.d_ttl = now.tv_sec + 20;
   rec.setContent(getRecordContent(QType::NSEC3, "1 0 500 ab HASG==== A RRSIG NSEC3"));
   rrsig = std::make_shared<RRSIGRecordContent>("NSEC3 5 3 10 20370101000000 20370101000000 24567 dummy. data");
   cache->insertNSEC(DNSName("powerdns.org"), rec.d_name, rec, {rrsig}, true);
 
   BOOST_CHECK_EQUAL(cache->getEntriesCount(), 3U);
 
   /* we have set a upper bound to 2 entries, so we are above,
-     and all entries are actually expired, so we will prune one entry
+     and one entry are actually expired, so we will prune one entry
```
rgacogne, line 1211:
```suggestion
     and one entry is actually expired, so we will prune one entry
```

pdns/recursordist/test-aggressive_nsec_cc.cc
```diff
@@ -1152,26 +1200,21 @@ BOOST_AUTO_TEST_CASE(test_aggressive_nsec_pruning)
...
 
-  /* now we are at the limit, so we will scan 1/5th of the entries,
-     and prune the expired ones, which mean we will also remove only one */
-  cache->prune(now.tv_sec + 600);
-  BOOST_CHECK_EQUAL(cache->getEntriesCount(), 1U);
-
-  /* now we are below the limit, so we will scan 1/5th of the entries again,
-     and prune the expired ones, which mean we will remove the last one */
+  /* now we are at the limit, so we will scan 1/5th of all zones entries, rounded up,
+     and prune the expired ones, which mean we will also remoing twoe */
```
rgacogne, line 1217:
```suggestion
  /* now we are at the limit, so we will scan 1/10th of all zones entries, rounded up,
     and prune the expired ones, which mean we will also be removing the remaining two */
```

rgacogne (approved):
A few typos but the main change looks good to me. I was a bit surprised to see the sneaky addition of the selective cache dump, but that's OK :)

PR #13226, opened 2023-09-08
Cleanup `ueberbackend.cc` a bit

fredmorcos:
### Short description
While I was fixing an issue with meson on MacOS related to the loading of backend shared modules when they are dynamically loaded, I formatted and cleaned up `ueberbackend.cc` a bit.

### Checklist
I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/pdnsutil.cc
+++ b/pdns/pdnsutil.cc
@@ -1,4 +1,5 @@
 
+#include <boost/smart_ptr/make_shared_array.hpp>
 #ifdef HAVE_CONFIG_H
 #include "config.h"
 #endif
@@ -2361,7 +2362,7 @@ static int testSchema(DNSSECKeeper& dk, const DNSName& zone)
   cout<<"Constructing UeberBackend"<<endl;
   UeberBackend B("default");
   cout<<"Picking first backend - if this is not what you want, edit launch line!"<<endl;
-  DNSBackend *db = B.backends[0];
+  DNSBackend *db = B.backends[0].get();
   cout << "Creating secondary zone " << zone << endl;
   db->createSlaveDomain("127.0.0.1", zone, "", "_testschema");
   cout << "Secondary zone created" << endl;
@@ -2505,6 +2506,7 @@ static int addOrSetMeta(const DNSName& zone, const string& kind, const vector<st
   return 0;
 }
 
+// NOLINTNEXTLINE(readability-function-cognitive-complexity): TODO Clean this function up.
 int main(int argc, char** argv)
 try
 {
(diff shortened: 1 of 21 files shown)


pdns/ueberbackend.cc
```diff
@@ -131,65 +140,71 @@ bool UeberBackend::createDomain(const DNSName& domain, const DomainInfo::DomainK
...
+  for (auto* backend : backends) {
+    if (backend->doesDNSSEC()) {
       return true;
+    }
   }
   return false;
 }
 
-bool UeberBackend::addDomainKey(const DNSName& name, const DNSBackend::KeyData& key, int64_t& id)
+bool UeberBackend::addDomainKey(const DNSName& name, const DNSBackend::KeyData& key, int64_t& domainID)
```
mind04, line 151: This is not a domain id and there are a few others like this later on.

mind04 (changes requested):

pdns/ueberbackend.hh
```diff
@@ -106,8 +106,8 @@ public:
 
   void getUnfreshSlaveInfos(vector<DomainInfo>* domains);
   void getUpdatedMasters(vector<DomainInfo>& domains, std::unordered_set<DNSName>& catalogs, CatalogHashMap& catalogHashes);
-  bool getDomainInfo(const DNSName& domain, DomainInfo& di, bool getSerial = true);
-  bool createDomain(const DNSName& domain, const DomainInfo::DomainKind kind, const vector<ComboAddress>& masters, const string& account);
+  bool getDomainInfo(const DNSName& domain, DomainInfo& domainInfo, bool getSerial = true);
+  bool createDomain(const DNSName& domain, DomainInfo::DomainKind kind, const vector<ComboAddress>& masters, const string& account);
```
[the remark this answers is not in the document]
fredmorcos: `DomainKind` is an enum and is passed by value, `const` is pointless in the declaration and only really useful in the definition. And the caller doesn't need to care whether it's `const` or not because they're passing a copy anyway.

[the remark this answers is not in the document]
Habbie: I hoped it was something like that, thanks!

pdns/ueberbackend.cc
```diff
@@ -553,164 +584,168 @@ void UeberBackend::cleanup()
...
   }
 
-  if(d_cached) {
-    while(d_cachehandleiter != d_answers.end()) {
-      rr=*d_cachehandleiter++;;
-      if((d_qtype == QType::ANY || rr.dr.d_type == d_qtype)) {
+  if (d_cached) {
+    while (d_cachehandleiter != d_answers.end()) {
+      rr = *d_cachehandleiter++;
+      ;
```
Habbie, line 731: what's this `;` for?

Habbie (commented):
one question, otherwise approved!

pdns/ueberbackend.cc
fredmorcos: It was there before on the line above `rr = *d_cachehandleiter++;;` and `clang-format` didn't remove it.

Habbie: oh right. perhaps you can remove it then :)

Habbie (approved):

fredmorcos:
Hopefully the last round of review for this.

pdns/ueberbackend.cc
```diff
@@ -374,173 +453,157 @@
...
-            j->second = *sd;
-            if(sd->qname == shorter) {
-              break;
-            }
-          } else {
-            DLOG(g_log<<Logger::Error<<"no match for: "<<shorter<<endl);
-          }
-        }
-      }
+      auto [backend, bestMatch] = findBestMatch(backends, matches, shorter, soaData);
```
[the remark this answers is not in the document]
Habbie: looks like another opportunity for cleanup

pdns/pdnsutil.cc
```diff
@@ -4084,14 +4086,18 @@ try
...
-        tgt = b;
+    for (auto& backend : BackendMakers().all()) {
+      auto* backendPtr = backend.release();
+
+      if (backendPtr->getPrefix() == cmds.at(1)) {
+        src = std::make_shared<DNSBackend>(backendPtr);
+      }
+      if (backendPtr->getPrefix() == cmds.at(2)) {
+        tgt = std::make_shared<DNSBackend>(backendPtr);
+      }
```
fredmorcos, line 4100: @rgacogne Is this any better?

fredmorcos: Since we can abort when OLD == NEW then using `unique_ptr`s for `src` and `tgt` and moving them out should be fine.

rgacogne: Nope, you would potentially have two smart pointers managing the same memory which is a terrible idea :)

fredmorcos: Ah correct. So it's `backendPtr` which should be the `shared_ptr` here.
Anyway, I switched them to `unique_ptr`.

rgacogne: I'm not sure a shared pointer is the best solution. Peter confirmed that we can stop right away if the `OLD` and `NEW` backend names are the same, so if we add that early check we can just keep using unique pointers, unless I'm missing something.

rgacogne (approved):
Great work! The cleanup of `UeberBackend::getAuth()` has been on my list for a long time and it was a daunting task, thanks a lot for making it happen!

Habbie (approved):

PR #13227, opened 2023-09-08
dnsdist: Fix unit tests for incoming DoH w/ nghttp2

rgacogne:
### Short description

These tests were failing on EL with LTO enabled, and passing everywhere else. It turns out that we did not properly reset the `s_processQuery` hack that we used in these tests to simulate the policy decision (rules and actions), and thus inherited what the last test set it to instead of the default (dropping queries), which was very unexpected.
This is a draft for now, as I only tested it on a development VM. I'll trigger the "Test package building for specific distributions" workflow in a few minutes to make sure that it indeed fixes it.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdistdist/test-dnsdistnghttp2-in_cc.cc
+++ b/pdns/dnsdistdist/test-dnsdistnghttp2-in_cc.cc
@@ -500,6 +500,10 @@ BOOST_FIXTURE_TEST_CASE(test_IncomingConnection_SelfAnswered, TestFixture)
   pwR.xfr32BitInt(0x01020304);
   pwR.commit();
 
+  /* we _NEED_ to set this function to empty otherwise we might get what was set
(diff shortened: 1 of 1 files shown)


Habbie (approved):

omoerbeek:
Pondering introducing randomization of test order to catch these type of bugs, same for regression tests.

Habbie:
> Pondering introducing randomization of test order to catch these type of bugs, same for regression tests.

That sounds good!

That said, I'd love to merge this PR soon, as we are currently not building dnsdist master for EL distributions.

rgacogne:
> That said, I'd love to merge this PR soon, as we are currently not building dnsdist master for EL distributions.

Will do once the checks are green :)

PR #13244, opened 2023-09-12
Rec settings minor cleanups

fredmorcos:
### Short description
- I added the special setup for Emacs/LSP in a `.dir-locals.el` file so it makes sense to git ignore it.
- Changes the use of references like `&Vec<T>` to `&[T]`.

### Checklist
I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/recursordist/Makefile.am
+++ b/pdns/recursordist/Makefile.am
@@ -46,7 +46,7 @@ CLEANFILES = htmlfiles.h \
 htmlfiles.h: incfiles html/* html/js/*
 	./incfiles > $@
 
-# We explicilt build settings in two steps, as settings modifies files in the settings/rust subdir
+# We explicitly build settings in two steps, as settings modifies files in the settings/rust subdir
 SUBDIRS=ext settings settings/rust
 
 if LUA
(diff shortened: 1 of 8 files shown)


rgacogne (approved):

omoerbeek:
The vector changes break the C++ callers.

fredmorcos:
> The vector changes break the C++ callers.

Right, I had to clean and rebuild to see the errors. I fixed them now but since you might not be happy about creating a `rust::Slice` object for each call, feel free to reject the PR.

omoerbeek:
I'm not terribly fond of the extra `Slice` objects on the C++ side. It's a case of either having the extra object on the Rust side or on the C++ side. OTOH for internal calls from Rust to Rust using a slice might be better indeed.

fredmorcos:
> I'm not terribly fond of the extra `Slice` objects on the C++ side. It's a case of either having the extra object on the Rust side or on the C++ side. OTOH for internal calls from Rust to Rust using a slice might be better indeed.

Alright. I reverted the commit in case we want to look back at it.

I added attributes for clippy to not warn about those instances.

omoerbeek (approved):

PR #13251, opened 2023-09-13
rec: When an exception is thrown from an mthread register stack switch (in the ASAN case)

omoerbeek:
This allows the test to be reintroduced on macOS plus the mtasker tests now start to work with ASA and randomization on: after the exception test is run other mtasker test would fail before.

I *do* wonder if the `d_runQueue.pop();` call should be done in the "exception is thrown" case. But as it is skipped in the current state of affairs, I am not pulling it iside the `catch` block for now`.

Also silence logging in the zoneToCache tests, previously an earlier test wold take care of that.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [X] documented the code

### DIFF
--- a/pdns/mtasker.cc
+++ b/pdns/mtasker.cc
@@ -380,7 +380,13 @@ template<class Key, class Val, class Cmp>bool MTasker<Key,Val,Cmp>::schedule(con
         ttdindex.erase(i++);                  // removes the waitpoint
 
         notifyStackSwitch(d_threads[d_tid].startOfStack, d_stacksize);
-        pdns_swapcontext(d_kernel, *uc); // swaps back to the above point 'A'
+        try {
+          pdns_swapcontext(d_kernel, *uc); // swaps back to the above point 'A'
+        }
+        catch (...) {
+          notifyStackSwitchDone();
+          throw;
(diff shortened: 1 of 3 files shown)


pdns/mtasker.cc
```diff
@@ -325,7 +325,14 @@ template<class Key, class Val, class Cmp>bool MTasker<Key,Val,Cmp>::schedule(con
     d_threads[d_tid].dt.start();
 #endif
     notifyStackSwitch(d_threads[d_tid].startOfStack, d_stacksize);
-    pdns_swapcontext(d_kernel, *d_threads[d_tid].context);
+    try {
+      pdns_swapcontext(d_kernel, *d_threads[d_tid].context);
```
rgacogne, line 329: Don't we need the same construct a few lines below, and in `::sendEvent()` as well?

omoerbeek: yes

rgacogne: It looks like the new commit you pushed only does the `sendEvent()` case, and not the one on line 370.

rgacogne (approved):

PR #13252, opened 2023-09-13
rec: Add make -C rust clean when generate.py is run

omoerbeek:
See comment for motivation. While working on the settings project, I had this line, but I thouhgt it was no longer needed so I removed it. It turns out it *is* needed. Found out when adding the recursor.tcp_threads field.

To reproduce the issue this PR fixes: rename a field in `table.py` and run make in `recursordist`. You'll see that `settings/rust/src/lib.rs.h` will not contain the name change.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code


Habbie (approved):
if it looks stupid and it works, it's not stupid. approved :)

PR #13262, opened 2023-09-15
Pass the requested qtype to lookup method

### LINKED ISSUE #13140: consistent-backend property for remote HTTP backends doesn't work as expected
- Program: Authoritative
 - Issue type: Bug report

### Short description
Hi, I'm using remote HTTP backend for lookups and I would like to receive lookup requests with specific qtype rather than qtype=ANY.
I found the following option - https://doc.powerdns.com/authoritative/settings.html#consistent-backends and set it to `no` in pdns.conf, but it doesn't work.
Power DNS still sends ANY requests to my backend.

### Environment
 - Operating system: macOS aarch64
 - Software version: master
 - Software source: compiled from master branch

### Steps to reproduce
1. Config:
```
launch=remote
remote-connection-string=http:url=http://host.docker.internal:8082/api/v1,timeout=2000
consistent-backends=no
receiver-threads=10
log-dns-queries=yes
log-dns-details=yes
loglevel=6
query-cache-ttl=0
local-address=0.0.0.0
local-port=53
zone-cache-refresh-interval=0
api=yes
webserver=true
webserver-address=0.0.0.0
webserver-port=8090
webserver-allow-from=0.0.0.0/0
```
2. Run `dig -p 53 @127.0.0.1 example.com A`

### Expected behaviour
PowerDNS makes `/api/v1/lookup/example.com/A` request to our HTTP backend

### Actual behaviour
PowerDNS makes `/api/v1/lookup/example.com/ANY` request to our HTTP backend

### Other information
I found the following code in packethandler.cc, which is invoked before lookup:
```B.lookup(QType(QType::ANY), target, d_sd.domain_id, &p);```
https://github.com/PowerDNS/pdns/blob/master/pdns/packethandler.cc#L1577

So, `d_handle.qtype=s_doANYLookupsOnly ? QType::ANY : qtype;` inside of lookup function makes no sense
https://github.com/PowerDNS/pdns/blob/master/pdns/ueberbackend.cc#L639

glebkin:
Closes #13140

### Short description
I'm using PowerDNS with remote HTTP backend and PowerDNS is always using ANY qtype on lookups regardless of consistent-backend property.
So I've changed this behaviour to pass initial QType to Lookup method.

### Checklist
I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/packethandler.cc
+++ b/pdns/packethandler.cc
@@ -1574,7 +1574,7 @@ std::unique_ptr<DNSPacket> PacketHandler::doQuestion(DNSPacket& p)
 #endif
 
     // see what we get..
-    B.lookup(QType(QType::ANY), target, d_sd.domain_id, &p);
+    B.lookup(p.qtype, target, d_sd.domain_id, &p);
     rrset.clear();
     haveAlias.clear();
     aliasScopeMask = 0;


rgacogne:
We are not going to merge this, as it would actually degrade performance for everyone else just to allow you not to have to deal with `ANY` queries.

Habbie:
This change would also break NXDOMAIN handling and wildcard support.

glebkin:
@rgacogne PowerDNS has consistent-backends property defaulted to yes, which will cause PDNS to use ANY type in lookup - 
https://github.com/PowerDNS/pdns/blob/master/pdns/ueberbackend.cc#L639

What is the reason of this property then if Lookup method is always getting ANY type as input param?

Habbie:
Hello,

our GitHub issue tracker is for bug reports and feature requests. Your question looks like a support question. Support questions are handled in our other online communities: [Github Discussions](https://github.com/PowerDNS/pdns/discussions), IRC and our mailing lists. Please see https://www.powerdns.com/opensource.html for information about those.

PR #13275, opened 2023-09-18
dnsdist: add lua binding to downstream address

### LINKED ISSUE #13201: dnsdist: add selected backend IP in DNSResponse
- Program: dnsdist
 - Issue type: Feature request

### Short description

It would be useful to get the selected backend IP in the DNSResponse for DNStap/Protobuf logging purpose.

### Usecase

I want to be able to filter replies per backend IP from DNS logs, to facilitate traffic debugging

### Description

Since dnsdist 1.8, the [pool](https://dnsdist.org/reference/dq.html?highlight=dnsquestion#DNSQuestion.pool) attribute is available and it can be added in DNStap or Protobuf messages (with extra field) to get more details regarding pools routing in DNS logs. In the same way, it will be useful to also have the selected backend IP. 

I took a look to the code, I can make a pull request for that but I prefer to check with you if you are OK with this idea.

dmachard:
### Short description

First shot to add a "selectedBackend" LUA binding to get the `ComboAddress` of the selected backend from `DNSResponse::d_downstream`. This pull request will fix #13201.

Successful testing with the following config

```lua
function alterDnstapResponse(dr, tap)
        tap:setExtra(dr.selectedBackend:toString() )
end

rl = newFrameStreamTcpLogger("10.0.0.100:6000")
addResponseAction(AllRule(), DnstapLogResponseAction("dnsdist1", rl, alterDnstapResponse))
```

### Checklist
I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdist-lua-bindings-dnsquestion.cc
+++ b/pdns/dnsdist-lua-bindings-dnsquestion.cc
@@ -501,15 +501,8 @@ private:
     return dnsdist::queueQueryResumptionEvent(std::move(query));
   });
 
-  luaCtx.registerMember<const ComboAddress (DNSResponse::*)>("selectedBackend", 
-    []( const DNSResponse& dr) -> const ComboAddress { 
-      if (dr.d_downstream == nullptr) {
-        return ComboAddress();
-      } else {
-        return dr.d_downstream->d_config.remote;
-      }
-    }, 
-    [](DNSResponse& dr, const ComboAddress newSelectedBackend) { (void) newSelectedBackend; }
-  );
+  luaCtx.registerFunction<std::shared_ptr<DownstreamState>(DNSResponse::*)(void)const>("getSelectedBackend", [](const DNSResponse& dr) {
+    return dr.d_downstream;
+  });
 #endif /* DISABLE_NON_FFI_DQ_BINDINGS */
 }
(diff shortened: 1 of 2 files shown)


rgacogne (changes requested):
Thanks a lot, much appreciated! I think it might be better to expose the `DownstreamState` object directly, instead of only its `ComboAddress`, as it would open more options in the future. We already have the bindings to get the `ComboAddress` from a `DownstreamState` so it shouldn't be too hard.
Since it doesn't make sense to be able to set the selected backend at this point, I think it might also be better to expose it via a method, like `DNSResponse:getSelectedBackend()` instead of an attribute. It would also make it possible to return a `boost::optional<DownstreamState>`, which would be `boost::none` when `dr.d_downstream` is `nullptr`, and would neatly translate into `nil` in `Lua`.
Do you agree? If you do, would you mind doing these changes?

dmachard:
Thanks for your help on IRC :) PR updated according to your requested changes.
Otherwise tested with success to get the selected IP and put it in extra field. 

```lua
function alterDnstapResponse(dr, tap)
        local backend = dr:getSelectedBackend()
	tap:setExtra(backend:getNameWithAddr())
end
```

rgacogne (approved):
Looks wonderful, thank you!

PR #13278, opened 2023-09-19
rec: packetcache submit task fixes

### LINKED ISSUE #13266: rec: packetcache submits tasks with unsupported qtypes
Happens if `refresh-on-ttl-prec != 0` and a repeated query for e.g. qtype `TYPE65535` comes in in the refresh window of time. PC should be more restrictive on `qtypes` pushed as refresh tasks. Maybe the `rcode` should also play a role, e.g. only push tasks triggered by cached queries having `rcode` 0. 

No real harm is done, expect for spurious log entries, btw.

omoerbeek:
### Short description

Fixes #13266.

ALso, only submit tasks that produce rcode 0 result. While it is not _wrong_ to submit tasks for queries that previously produced rcode != 0, these cache entries or governed by different (smaller) TTLs, so we are likely to submit more tasks than makes sense.
When these rcode != 0 entries with small TTLs are expired, on repeated client queries we'll trigger regular query processing (using the record cache), which itself wil trigger tasks if relevant.

Also include some general delinting of common cache code.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/recursordist/test-recpacketcache_cc.cc
+++ b/pdns/recursordist/test-recpacketcache_cc.cc
@@ -10,6 +10,8 @@
 #include "dns_random.hh"
 #include "iputils.hh"
 #include "recpacketcache.hh"
+#include "taskqueue.hh"
+#include "rec-taskqueue.hh"
 #include <utility>
 
 BOOST_AUTO_TEST_SUITE(test_recpacketcache_cc)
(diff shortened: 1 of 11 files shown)


pdns/recursordist/test-recpacketcache_cc.cc
```diff
@@ -80,6 +82,66 @@ BOOST_AUTO_TEST_CASE(test_recPacketCacheSimple)
...
+  BOOST_CHECK_EQUAL(rpc.size(), 1U);
+  uint32_t qhash2 = 0;
+  bool found = rpc.getResponsePacket(tag, qpacket, qname, QType::A, QClass::IN, now, &fpacket, &age, &qhash2);
+  BOOST_CHECK_EQUAL(found, true);
+  BOOST_CHECK_EQUAL(qhash, qhash2);
+  BOOST_CHECK_EQUAL(fpacket, rpacket);
+
+  BOOST_REQUIRE_EQUAL(getTaskSize(), 0U);
+
+  // Half of time has passed, no task should have been submitted
```
rgacogne, line 124: I'm not sure I agree with this:
- `RecursorPacketCache::s_refresh_ttlperc` is set to `70` which, looking at the code, is equivalent to setting `refresh-on-ttl-perc` to `70`
- the documentation states `Whenever a record is fetched from the packet or record cache and only refresh-on-ttl-perc percent or less of its original TTL is left [...]`
- at this point, only 50% of the original TTL is left, which seems less than 70% to me?

omoerbeek: Oops, I might have confused myself.... will revisit.

omoerbeek: Unconfused myself. The original deadline computation is OK. Removed that commit and fixed the test, force pushed to now have the wrong commit in history.

rgacogne (approved):
Looks good!

PR #13289, opened 2023-09-22
rec: implement probabilistic unthrottle

omoerbeek:
### Short description

This mechanism selects servers marked down to be used anyway with a given chance. Inspired by the OARC talk
https://indico.dns-oarc.net/event/47/contributions/1018/ by  Fenglu Zhang (Tsinghua University) .

If picked a chance of 1/25, but I do not have data yet to determine the "right" value.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [X] documented the code

### DIFF
--- a/pdns/recursordist/settings/table.py
+++ b/pdns/recursordist/settings/table.py
@@ -2155,15 +2155,15 @@ Throttle a server that has failed to respond :ref:`setting-server-down-max-fails
  ''',
     },
     {
-        'name' : 'server_down_use_probability',
+        'name' : 'bypass_server_throttling_probability',
         'section' : 'recursor',
         'type' : LType.Uint64,
         'default' : '25',
         'help' : 'Determines the probability of a server marked down to be used anyway',
         'doc' : '''
 This setting determines the probability of a server marked down to be used anyway.
-A value of ``n`` means that the chance of a server marked down being used after it wins speed selection is is ``1/n``.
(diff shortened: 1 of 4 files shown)


pdns/recursordist/settings/table.py
```diff
@@ -2154,6 +2154,19 @@
 Throttle a server that has failed to respond :ref:`setting-server-down-max-fails` times for this many seconds.
  ''',
     },
+    {
+        'name' : 'server_down_use_probability',
```
rgacogne, line 2158: I wonder if it would make sense to use the word `throttled` here? Perhaps something like `bypass_server_throttling_probability`?

pdns/recursordist/settings/table.py
```diff
@@ -2154,6 +2154,19 @@
...
+        'section' : 'recursor',
+        'type' : LType.Uint64,
+        'default' : '25',
+        'help' : 'Determines the probability of a server marked down to be used anyway',
+        'doc' : '''
+This setting determines the probability of a server marked down to be used anyway.
+A value of ``n`` means that the chance of a server marked down being used after it wins speed selection is is ``1/n``.
+If this setting is zero this mechanism is not active.
```
rgacogne, line 2166: It's not clear to me after reading this sentence whether `0` means that the throttling is never bypassed or that the throttling is never enforced.

rgacogne (approved):

PR #13293, opened 2023-09-22
sdig: add rudimentary EDE output

Habbie:
### Short description
It would be nice to map the numbers to strings, but we don't have that map in our codebase yet.

### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code


Habbie:
> It would be nice to map the numbers to strings, but we don't have that map in our codebase yet.

and if we don't add that map now and decide to merge this with numbers, I'm not sure about the exact formatting yet

rgacogne (approved):

Habbie:
@omoerbeek as this touches recursor files, I added you as reviewer too

omoerbeek (approved):

Habbie:
Thanks! Then we'll leave "text instead of number" for later.

PR #13296, opened 2023-09-25
rec: Make QName Minimization parameters from RFC 9156 settable

omoerbeek:
Also fix a counting ommission

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/recursordist/rec-main.cc
+++ b/pdns/recursordist/rec-main.cc
@@ -1687,6 +1687,8 @@ static int initSyncRes(Logr::log_t log)
   SyncRes::s_ecscachelimitttl = ::arg().asNum("ecs-cache-limit-ttl");
 
   SyncRes::s_qnameminimization = ::arg().mustDo("qname-minimization");
+  SyncRes::s_minimize_one_lab = ::arg().asNum("qname-minimize-one-label");
+  SyncRes::s_max_minimize_count = ::arg().asNum("qname-max-minimize-count");
 
   SyncRes::s_hardenNXD = SyncRes::HardenNXD::DNSSEC;
   string value = ::arg()["nothing-below-nxdomain"];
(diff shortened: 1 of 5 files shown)


pdns/recursordist/rec-main.cc
```diff
@@ -1687,6 +1687,8 @@ static int initSyncRes(Logr::log_t log)
   SyncRes::s_ecscachelimitttl = ::arg().asNum("ecs-cache-limit-ttl");
 
   SyncRes::s_qnameminimization = ::arg().mustDo("qname-minimization");
+  SyncRes::s_minimize_one_lab = ::arg().asNum("qname-minimize-one-label");
```
rgacogne, line 1690: It's not obvious to me without the context that `lab` in `minimize_one_lab`  means `label`, perhaps:
```suggestion
  SyncRes::s_minimize_one_label = ::arg().asNum("qname-minimize-one-label");
```

pdns/recursordist/syncres.cc
```diff
@@ -1788,6 +1788,7 @@ int SyncRes::doResolve(const DNSName& qname, const QType qtype, vector<DNSRecord
       LOG(prefix << qname << ": Step4 Resolve " << child << "|A result is " << RCode::to_s(res) << "/" << retq.size() << "/" << stopAtDelegation << endl);
       if (stopAtDelegation == Stopped) {
         LOG(prefix << qname << ": Delegation seen, continue at step 1" << endl);
+        i++;
```
rgacogne, line 1791: I wonder if it would be clearer to increase `i` on line `1698` instead?
```c++
for (unsigned int i = 0; i <= qnamelen; i++) {
```

pdns/recursordist/rec-main.cc
omoerbeek: Right. The name comes from the RFC, but full `label`` is better.

rgacogne (approved):

PR #13304, opened 2023-09-28
dnsdist: Fix a race in the Async unit tests

rgacogne:
### Short description

We used to set the `errorRaised` variable from the `AsynchronousHolder` thread then check its value from the main thread, which is correctly reported as TSAN as a data race. We do know that we have waited enough, and that otherwise it's fine to fail, but TSAN cannot know that. Switching to a `std::atomic<bool>` fixes it.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


rgacogne:
CI failures are unrelated (Recursor DoT test failing).

PR #13305, opened 2023-09-28
dnsdist: Set proper levels when logging messages

rgacogne:
### Short description

This PR re-evaluates the log level of all messages in the DNSdist code base. The rules are simple:
- 'Error' should only be used when a critical service will not function properly
- 'Warning' should be used when a non-essential feature may not be working as expected
- 'Info' is for everything that should not go unnoticed but also cannot be triggered by an outside attacker, otherwise we might be flooding our logs
- 'Verbose' can be used for everything else.

There are a few exceptions, something we log at the 'info' level for events that might be happening a lot, but only when told to do so, like for example with verbose health-check reporting.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdistdist/dnsdist-kvs.cc
+++ b/pdns/dnsdistdist/dnsdist-kvs.cc
@@ -96,8 +96,8 @@ bool LMDBKVStore::getValue(const std::string& key, std::string& value)
       return false;
     }
   }
-  catch(const std::exception& e) {
-    warnlog("Error while looking up key '%s' from LMDB file '%s', database '%s': %s", key, d_fname, d_dbName, e.what());
+  catch (const std::exception& e) {
+    vinfolog("Error while looking up key '%s' from LMDB file '%s', database '%s': %s", key, d_fname, d_dbName, e.what());
   }
   return false;
 }
@@ -115,8 +115,8 @@ bool LMDBKVStore::keyExists(const std::string& key)
       return false;
     }
   }
-  catch(const std::exception& e) {
-    warnlog("Error while looking up key '%s' from LMDB file '%s', database '%s': %s", key, d_fname, d_dbName, e.what());
+  catch (const std::exception& e) {
+    vinfolog("Error while looking up key '%s' from LMDB file '%s', database '%s': %s", key, d_fname, d_dbName, e.what());
   }
   return false;
 }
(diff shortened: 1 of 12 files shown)


phonedph1:
+1 on the nghttp2 TLS exceptions changing, I had actually patched this locally as the exceptions at infolog were way too much.

pdns/dnsdistdist/dnsdist-kvs.cc
```diff
@@ -96,8 +96,8 @@ bool LMDBKVStore::getValue(const std::string& key, std::string& value)
       return false;
     }
   }
-  catch(const std::exception& e) {
-    warnlog("Error while looking up key '%s' from LMDB file '%s', database '%s': %s", key, d_fname, d_dbName, e.what());
+  catch (const std::exception& e) {
+    vinfolog("Error while looking up key '%s' from LMDB file '%s', database '%s': %s", key, d_fname, d_dbName, e.what());
```
Habbie, line 100: I wonder if "info" is too low here, as "not found" is not an error condition

Habbie (approved):
one question, but fine to merge if you feel that one is right

pdns/dnsdistdist/dnsdist-kvs.cc
rgacogne: I agree `info` is a bit low, but this is not something we might want to be logging for every query either if something goes wrong, which is why I have gated it behind the `verbose` flag. But perhaps it means that we need `verrlog` and `vwarnlog` (this looks ugly, but you get the idea) so that we can log at a meaningful level in verbose mode?

Habbie: Agreed, don't want this repeated 1000 times per second. New `v*` calls make sense to me!

Habbie: (merging this as info for now also makes sense to me)

rgacogne: I added a comment in https://github.com/PowerDNS/pdns/pull/13156#issuecomment-1740585643 since I'm refactoring a fair amount of the logging code in that PR anyway.

Habbie: ah, makes sense!

Issue #12876, opened 2023-06-01
auth: g++ 13.1 warnings in the LMDB backend


rgacogne:
- Program: Authoritative
 - Issue type: Bug report

### Short description

Compiling auth 4.8.0 on Arch with g++ 13.1 yields:
```
lmdbbackend.cc: In static member function ‘static bool LMDBBackend::upgradeToSchemav5(std::string&)’:
lmdbbackend.cc:467:27: warning: loop variable ‘dbname’ of type ‘const std::string&’ {aka ‘const std::__cxx11::basic_string<char>&’} binds to a temporary constructed from type ‘const char* const’ [-Wrange-loop-construct]
  467 |   for (const std::string& dbname : {"domains", "keydata", "tsig", "metadata"}) {
      |                           ^~~~~~
lmdbbackend.cc:467:27: note: use non-reference type ‘const std::string’ {aka ‘const std::__cxx11::basic_string<char>’} to make the copy explicit or ‘const char* const&’ to prevent copying
lmdbbackend.cc:507:27: warning: loop variable ‘dbname’ of type ‘const std::string&’ {aka ‘const std::__cxx11::basic_string<char>&’} binds to a temporary constructed from type ‘const char* const’ [-Wrange-loop-construct]
  507 |   for (const std::string& dbname : {"domains", "keydata", "tsig", "metadata"}) {
      |                           ^~~~~~
lmdbbackend.cc:507:27: note: use non-reference type ‘const std::string’ {aka ‘const std::__cxx11::basic_string<char>’} to make the copy explicit or ‘const char* const&’ to prevent copying
lmdbbackend.cc:556:27: warning: loop variable ‘keyname’ of type ‘const std::string&’ {aka ‘const std::__cxx11::basic_string<char>&’} binds to a temporary constructed from type ‘const char* const’ [-Wrange-loop-construct]
  556 |   for (const std::string& keyname : {"schemaversion", "shards"}) {
      |                           ^~~~~~~
lmdbbackend.cc:556:27: note: use non-reference type ‘const std::string’ {aka ‘const std::__cxx11::basic_string<char>’} to make the copy explicit or ‘const char* const&’ to prevent copying
lmdbbackend.cc:594:27: warning: loop variable ‘keyname’ of type ‘const std::string&’ {aka ‘const std::__cxx11::basic_string<char>&’} binds to a temporary constructed from type ‘const char* const’ [-Wrange-loop-construct]
[...]

PR #13321, opened 2023-10-03
Minor cleanups to geoip and lmdb

fredmorcos:
### Short description
Cleanup some compilation warnings in geoip and lmdb.

### Checklist
I have:
- [X] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/modules/geoipbackend/geoipbackend.cc
+++ b/modules/geoipbackend/geoipbackend.cc
@@ -38,8 +38,8 @@ ReadWriteLock GeoIPBackend::s_state_lock;
 
 struct GeoIPDNSResourceRecord : DNSResourceRecord
 {
-  int weight;
-  bool has_weight;
+  int weight{};
+  bool has_weight{};
 };
 
 struct GeoIPService
@@ -53,7 +53,7 @@ struct GeoIPDomain
 {
   int id;
   DNSName domain;
-  int ttl;
+  int ttl{};
   map<DNSName, GeoIPService> services;
   map<DNSName, vector<GeoIPDNSResourceRecord>> records;
   vector<string> mapping_lookup_formats;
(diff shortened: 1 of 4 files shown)


modules/geoipbackend/geoipbackend.cc
```diff
@@ -121,7 +121,7 @@ static bool validateMappingLookupFormats(const vector<string>& formats)
   return true;
 }
 
-bool GeoIPBackend::loadDomain(const YAML::Node& domain, unsigned int id, GeoIPDomain& dom)
+bool GeoIPBackend::loadDomain(const YAML::Node& domain, unsigned int /* id */, GeoIPDomain& dom)
```
rgacogne, line 124: If we don't use `id` it should probably go?

rgacogne (changes requested):

modules/geoipbackend/geoipbackend.cc
fredmorcos: Correct

Habbie (dismissed):

rgacogne:
I'm sorry, Fred, but I now realize that the change I asked you to do is wrong. The problem is that we need to `dom.id` to a different value for all domains, and it looks like we were no longer doing that since https://github.com/PowerDNS/pdns/commit/4287cb762b1d3603047d60e3d0918e7cf09ddd00 So I think the correct fix is to revert the last change and to set `dom.id = id` as we used to do before the aforementioned commit.

Habbie (approved):

rgacogne (approved):
:+1:

modules/geoipbackend/geoipbackend.cc
```diff
@@ -51,9 +52,9 @@ struct GeoIPService
 
 struct GeoIPDomain
 {
-  int id;
+  std::size_t id{};
```
fredmorcos, line 55: Is this OK?

Habbie: it needs to fit in 31 bits, so uint32_t might make more sense (that's what dnsbackend.hh uses)

fredmorcos:
This needs another review.

rgacogne (approved):

Habbie (approved):

Habbie:
I believe this fixes #12876 !

fredmorcos:
> I believe this fixes #12876 !

Yes it does.

Issue #6994, opened 2018-09-27
Ixfrdist Notify Support


oliken:
- Program: ixfrdist
 - Issue type: Feature request

### Short description
It would be nice if IXFRDist support NOTIFY, Catch DNS NOTIFY from master and update AFXR/Send DNS NOTIFY to some hosts.

### Usecase
When ixfrdist is use with auth server, this feature permit to initiate AXFR only when necessary.

### Description
ixfrdist can catch NOTIFY from master/auth, and can initiate AXFR transfer zone and SOA check only when there is modification on master server vs check SOA expiration check only.
This feature may permit to replace the limitation of PowerDNS to send IXFR.

Woet:
Seems like you skipped two sections for the feature request template.

```
<!--
Use the part below to file a feature request and delete the bug report part above.
-->
### Usecase
<!-- Tell what you're trying to achieve, without describing _what_ the requested feature should do -->

### Description
<!-- Describe as extensively as possible what you want the software to do -->
```

isnuryusuf:
### Usecase


### Description
- currently ixfrdist doesnt support automatic zone update after get notify from master, we need to restart ixfrdist service to get complete up-to-date zone record from master, cause every we start ixfrdist, it will check serial on master and compare to local-copy. so every time serial has update, we need to restart the ixfrdist service

- after ixfrdist have same zone record and serial from master, they cannot notify slave member, at the slave side, we need to restart the pdns-slave/other dns service, so slave can get update from ixfrdist

if we have scenario like this
- pdns master with mysql backend and milion of record
- ixfrdist outsite pdns master server to distribute record using IXFR
- lot of pnds slave connecting to ixfrdist server

> lot of pdns slave direct connecting to pdns master where handle milion of record and pdns currently doenst support IXFR transfer, it will consume lot of Bandwidth and High SQL query (at the database backend)
- the main goal is to make PDNS master to support IXFR, but ixfrdist is nice when can translate from AXFR to IXFR to lot of PDNS slave

Habbie:
handling of incoming notify is done in #13322

PR #13322, opened 2023-10-03
ixfrdist: add NOTIFY receive support

Habbie:
### Short description
part of #6994

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/ixfrdist-stats.hh
+++ b/pdns/ixfrdist-stats.hh
@@ -71,9 +71,9 @@ class ixfrdistStats {
       progStats.unknownDomainInQueries += 1;
     }
 
-    void incrementNotImplemented(const string& opcode)
+    void incrementNotImplemented(uint8_t opcode)
     {
-      notimpStats[opcode] ++;
+      notimpStats.at(opcode) ++;
     }
 
   private:
@@ -99,7 +99,7 @@ class ixfrdistStats {
     };
 
     std::map<DNSName, perDomainStat> domainStats;
-    std::map<string, std::atomic<uint64_t>> notimpStats;
+    std::array<std::atomic<uint64_t>, 16> notimpStats;
     programStats progStats;
 
     std::map<DNSName, perDomainStat>::iterator getRegisteredDomain(const DNSName& d) {
--- a/pdns/ixfrdist-stats.hh
+++ b/pdns/ixfrdist-stats.hh
@@ -24,6 +24,7 @@
 #include <map>
 #include <string>
 
+#include "dns.hh"
 #include "dnsname.hh"
 #include "pdnsexception.hh"
 
@@ -70,6 +71,11 @@ class ixfrdistStats {
       progStats.unknownDomainInQueries += 1;
     }
 
+    void incrementNotImplemented(const string& opcode)
+    {
+      notimpStats[opcode] ++;
+    }
+
   private:
     class perDomainStat {
       public:
@@ -93,6 +99,7 @@ class ixfrdistStats {
     };
 
     std::map<DNSName, perDomainStat> domainStats;
+    std::map<string, std::atomic<uint64_t>> notimpStats;
     programStats progStats;
 
     std::map<DNSName, perDomainStat>::iterator getRegisteredDomain(const DNSName& d) {
--- a/pdns/ixfrdist.cc
+++ b/pdns/ixfrdist.cc
@@ -278,7 +278,7 @@ static void updateCurrentZoneInfo(const DNSName& domain, std::shared_ptr<ixfrinf
   // FIXME: also report zone size?
 }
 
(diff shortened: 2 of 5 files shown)


pdns/ixfrdist-stats.hh
```diff
@@ -70,6 +71,11 @@ class ixfrdistStats {
       progStats.unknownDomainInQueries += 1;
     }
 
+    void incrementNotImplemented(const string& opcode)
+    {
+      notimpStats[opcode] ++;
```
rgacogne, line 76: It looks like the map can be accessed by several threads at at time, which would be an issue because one thread can be inserting a new entry while another one is reading the map.

Habbie: good catch! making the values atomic is not enough

Habbie: given opcode is only 4 bits, an array might be simplest here?

rgacogne: A `std::array` would work, sure! Given that we are not supposed to increment this value very often a mutex might work as well.

Habbie: done

Habbie:
made clang-tidy happy

pdns/ixfrdist.cc
```diff
@@ -474,33 +483,94 @@ static void updateThread(const string& workdir, const uint16_t& keep, const uint
...
+  bool masterFound = false;
+
+  auto saddrPort0 = saddr;
+  saddrPort0.setPort(0);
+
+  for(const auto &master : masters) {
+    auto masterPort0 = master;
+    masterPort0.setPort(0);
+    cerr<<masterPort0.toStringWithPort()<<"=="<<saddrPort0.toStringWithPort()<<endl;
+    if (masterPort0 == saddrPort0) {
```
rgacogne, line 518: `` 
```suggestion
    if (ComboAddress::addressOnlyEqual()(saddr, master)) {
```

pdns/ixfrdist.cc
```diff
@@ -806,14 +922,24 @@ static void handleUDPRequest(int fd, boost::any&) {
     return;
   }
 
-  if (!allowedByACL(saddr)) {
+  MOADNSParser mdp(true, string(&buf[0], static_cast<size_t>(res)));
```
rgacogne, line 925: I'm not thrilled by the parsing happening before checking the ACL, but I can live with it.

rgacogne (commented):
Looks good, a few remaining error messages, one suggestion.

pdns/ixfrdist.cc
Habbie: ooh

pdns/ixfrdist.cc
Habbie: hmm, I actually have two NMGs that, together, cover all possible legitimate sources, I'll improve this

Habbie: in fact the TCP path already does that!

Habbie:
Thanks Rémi, I added two commits

pdns/ixfrdist.cc
```diff
@@ -857,7 +988,9 @@ static void handleTCPRequest(int fd, boost::any&) {
     return;
   }
 
-  if (!allowedByACL(saddr)) {
+  // we allow the connection if this is a legit client or a legit NOTIFY source
+  // need to check per-operation later
+  if (!allowedByACL(saddr) && !allowedByACL(saddr, true)) {
     g_log<<Logger::Warning<<"TCP query from "<<saddr.toString()<<" is not allowed, dropping"<<endl;
     close(cfd);
```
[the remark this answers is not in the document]
Habbie: noted in https://github.com/PowerDNS/pdns/issues/13403, thanks

pdns/ixfrdist.cc
```diff
@@ -827,9 +949,14 @@ static void handleUDPRequest(int fd, boost::any&) {
      */
     g_stats.incrementSOAinQueries(mdp.d_qname); // FIXME: this also counts IXFR queries (but the response is the same as to a SOA query)
     makeSOAPacket(mdp, packet);
-  } else {
+  } else if (respt == ResponseType::EmptyNoError) {
+    makeEmptyNoErrorPacket(mdp, packet);
```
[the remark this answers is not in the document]
Habbie: `SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /usr/bin/../lib/gcc/x86_64-linux-gnu/12/../../../../include/c++/12/bits/stl_vector.h:1124:9 in `

yep!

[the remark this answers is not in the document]
Habbie: I removed the `return false` path, it was a bogus check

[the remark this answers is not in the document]
rgacogne: Good!

rgacogne (approved):

PR #13346, opened 2023-10-09
rec:  Fix a spuriously failing recursorcache_cc test: reset globals (indirectly) used

omoerbeek:
### Short description

For SyncRes tests we have a general mechanism setting the globals
before the test are run, but the non-syncres tests do not have that, while
they still use some globals.  In this particular case, the test would
fail if the last SyncRes test run before was setting
SyncRes::s_locked_ttlperc.

While there, pass the time to the prune functions. This avoids
potential timing issues for some tests.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code
- [X added or modified unit test(s)

### DIFF
--- a/pdns/recursordist/test-recursorcache_cc.cc
+++ b/pdns/recursordist/test-recursorcache_cc.cc
@@ -13,6 +13,7 @@ BOOST_AUTO_TEST_SUITE(recursorcache_cc)
 
 static void simple(time_t now)
 {
+  MemRecursorCache::resetStaticsForTests();
   MemRecursorCache MRC;
 
   std::vector<DNSRecord> records;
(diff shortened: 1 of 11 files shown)


pdns/recursordist/test-recursorcache_cc.cc
```diff
@@ -387,6 +388,7 @@ BOOST_AUTO_TEST_CASE(test_RecursorCacheSimpleDistantFuture)
 
 BOOST_AUTO_TEST_CASE(test_RecursorCacheGhost)
```
rgacogne, line 389: Note that you could use a fixture [1] to do that, although in this particular case that might not be worth the trouble.

[1]: https://www.boost.org/doc/libs/1_83_0/libs/test/doc/html/boost_test/tests_organization/fixtures/case.html

rgacogne (approved):

PR #13348, opened 2023-10-11
dnsdist: Switch to our fork of h2o to mitigate http2 rapid reset

rgacogne:
### Short description

Testing in progress at https://github.com/rgacogne/pdns/actions/runs/6480831098

### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code


rgacogne:
I just realized I probably need to bump the Debian revision, looking into how this is done.

Habbie:
> I just realized I probably need to bump the Debian revision, looking into how this is done.

Just bumping the normal version should be more than enough I think?

mnordhoff:
I know the `curl` commands were already there, but it might be good to add the [`-f`](https://curl.se/docs/manpage.html#-f) option?

> (HTTP) Fail fast with no output at all on server errors. This is useful to enable scripts and users to better deal with failed attempts. In normal cases when an HTTP server fails to deliver a document, it returns an HTML document stating so (which often also describes why and more). This flag prevents curl from outputting that and return error 22.
>
> This method is not fail-safe and there are occasions where non-successful response codes slip through, especially when authentication is involved (response codes 401 and 407).

rgacogne:
> I know the `curl` commands were already there, but it might be good to add the [`-f`](https://curl.se/docs/manpage.html#-f) option?
> 
> > (HTTP) Fail fast with no output at all on server errors. This is useful to enable scripts and users to better deal with failed attempts. In normal cases when an HTTP server fails to deliver a document, it returns an HTML document stating so (which often also describes why and more). This flag prevents curl from outputting that and return error 22.
> > This method is not fail-safe and there are occasions where non-successful response codes slip through, especially when authentication is involved (response codes 401 and 407).

Sounds like a good idea, thanks!

Habbie (approved):

PR #13350, opened 2023-10-11
dnsdist: Prepare ChangeLog and secpoll for DNSdist 1.8.2 and 1.7.5

rgacogne:
### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code


docs/secpoll.zone
```diff
@@ -517,4 +517,5 @@ dnsdist-1.8.0-rc2.security-status                          60 IN TXT "2 Unsuppor
 dnsdist-1.8.0-rc3.security-status                          60 IN TXT "2 Unsupported pre-release (no known vulnerabilities)"
 dnsdist-1.8.0.security-status                              60 IN TXT "1 OK"
 dnsdist-1.8.1.security-status                              60 IN TXT "1 OK"
```
Habbie, line 519: 1.8.0 and 1.8.1 are no longer OK

rgacogne: That's not really true, is it? They are OK when linked against a patched h2o library.

rgacogne: So technically our packages are not OK, but I don't think we can make this distinction in secpoll, unfortunately.

Habbie: That is true, but the 1.8.1 package from repo.powerdns.com will never be linked against a patched library, because it's already built and released. secpoll is not used much outside of our packages, as far as I know

rgacogne: Done

docs/secpoll.zone
```diff
@@ -1,4 +1,4 @@
-@       86400   IN  SOA pdns-public-ns1.powerdns.com. peter\.van\.dijk.powerdns.com. 2023100501 10800 3600 604800 10800
+@       86400   IN  SOA pdns-public-ns1.powerdns.com. peter\.van\.dijk.powerdns.com. 2023101103 10800 3600 604800 10800
```
Habbie, line 1: I think this could be `01`, not `03`, but I don't care much :)

Habbie (approved):

docs/secpoll.zone
rgacogne: Right, I did not check and emacs helpfully auto-updated it for me :shrug:

PR #13360, opened 2023-10-12
Update alias.rst

stasic:
If a zone is operated with different nameserver software than PowerDNS and ALIAS is used it could result in an out-of-sync sencondary.

### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/docs/guides/alias.rst
+++ b/docs/guides/alias.rst
@@ -49,6 +49,8 @@ When set to 'yes', PowerDNS will look up the A and AAAA records of the name in t
 This is useful when your secondary servers do not understand ALIAS, or should not look up the addresses themselves.
 Note that secondaries will not automatically follow changes in those A/AAAA records unless you AXFR regularly.
 
+If secondaries are operated with nameserver software which is not capable of synthesising ALIAS records the zone will probably go out of sync. Because updated ALIAS target records will not result in an increment of the Serial, which means that Secondaries won't pull the Zone.
+
 If the ALIAS target cannot be resolved, the AXFR will fail.
 When set to 'ignore-errors', an unresolvable ALIAS target will be omitted from the outgoing transfer.
 


klaus-nicat:
I would add the sentence:
"To keep secondary nameservers up2date with the ALIAS target, you have to periodically retransfer the zone from PowerDNS, to force ALIAS-resolving during AXFR. This can be done either "dumb", for example every minute, or implement some "out of band" logic to retransfer zones only if the ALIAS target actually changed."

docs/guides/alias.rst
```diff
@@ -49,6 +49,8 @@ When set to 'yes', PowerDNS will look up the A and AAAA records of the name in t
 This is useful when your secondary servers do not understand ALIAS, or should not look up the addresses themselves.
 Note that secondaries will not automatically follow changes in those A/AAAA records unless you AXFR regularly.
 
+If secondaries are operated with nameserver software which is not capable of synthesising ALIAS records the zone will probably go out of sync. Because updated ALIAS target records will not result in an increment of the Serial, which means that Secondaries won't pull the Zone.
```
Habbie, line 52:
```suggestion
If secondaries are operated with name server software which is not capable of synthesizing ALIAS records the zone will probably go out of sync. Because updated ALIAS target records will not result in an increment of the Serial, which means that Secondaries won't pull the Zone.
```

Habbie (changes requested):
and please look at Klaus' comment, thanks!

pieterlexis:
@Habbie I addressed the review comments myself and updated the PR.

docs/guides/alias.rst
```diff
@@ -60,6 +60,11 @@ When set to 'yes', PowerDNS will look up the A and AAAA records of the name in t
 This is useful when your secondary servers do not understand ALIAS, or should not look up the addresses themselves.
 Note that secondaries will not automatically follow changes in those A/AAAA records unless you AXFR regularly.
 
+If secondaries are operated with name server software which is not capable of synthesizing ALIAS records the zone will probably go out of sync.
```
miodvallat:
```suggestion
If secondaries are operated with name server software which is not capable of synthesizing ALIAS records, the zone will probably go out of sync.
```

docs/guides/alias.rst
```diff
@@ -60,6 +60,11 @@ When set to 'yes', PowerDNS will look up the A and AAAA records of the name in t
...
+Because updated ALIAS target records will not result in an increment of the Serial, which means that Secondaries won't pull the Zone.
+To keep secondary name servers up to date with the ALIAS target, you'll have to periodically retransfer the zone from PowerDNS to force ALIAS-resolving during AXFR.
+This can be done either "dumb", for example every minute, or implement some "out of band" logic to retransfer zones only if the ALIAS target actually changed.
```
miodvallat:
```suggestion
This can be done either in a dumb-but-reliable way, for example every minute, or using some "out of band" logic to retransfer zones only when the ALIAS target actually changes.
```

docs/guides/alias.rst
```diff
@@ -60,6 +60,11 @@ When set to 'yes', PowerDNS will look up the A and AAAA records of the name in t
 This is useful when your secondary servers do not understand ALIAS, or should not look up the addresses themselves.
 Note that secondaries will not automatically follow changes in those A/AAAA records unless you AXFR regularly.
 
+If secondaries are operated with name server software which is not capable of synthesizing ALIAS records the zone will probably go out of sync.
+Because updated ALIAS target records will not result in an increment of the Serial, which means that Secondaries won't pull the Zone.
```
miodvallat:
```suggestion
Because updated ALIAS target records will not result in an increment of the zone serial number, secondaries won't consider the zone as in need of an update, and won't pull its contents.
```

miodvallat (dismissed):
Don't forget to add a signed-off line to your commit (maybe squash?)

pieterlexis:
> Don't forget to add a signed-off line to your commit (maybe squash?)

This is because the original submitter did not sign off (in the first commit), as that was not a requirement at the time. I could squash all commits and remove the submitter as the author, but that feels wrong.

miodvallat:
But you can sign-off for the original author.

PR #13361, opened 2023-10-12
dnsdist: Fix the documentation of `dq:restart` to prevent an endless loop

phonedph1:
### Short description
If the re-try is also a servfail, dnsdist will restart it endlessly. Give some tips to not do that.

### Checklist
I have:
- [ ] compiled this code
- [x] tested this code
- [ ] documented the code


rgacogne (approved):
Good catch, thanks!

PR #13362, opened 2023-10-13
rec: convert API config from old style to YAML if YAML settings are active

### LINKED ISSUE #12679: Don't require api-config-dir and include-dir to be the same when allowing updates [security]
- Program:  Recursor
 - Issue type: Feature request

### Short description
As the [docs](https://doc.powerdns.com/recursor/settings.html#api-config-dir) indicate, the `api-config-dir` and `include-dir` settings must be set to the same dir and that dir must be writable by the PowerDNS proces for API  updates to work. This creates multiple security risks.

### Usecase
Adhere to best security practices, while still allowing API usage to update certain settings. Also work nicely with `systemd`.

### Description
The only configuration settings which can be set (per the docs) today are:
* dns resolving interference: `forward-zones`, `forward-zones-recurse`, `auth-zones`
* `allow-from` and `allow-notify-from`

Implementation-wise, this might be a good approach:
* include both dirs on load. 
* require only the `api-config-dir` to be writable?
* filter settings from the the `api-config-dir`-loader so that only the allowed settings are read from that dir

On the systemd remark: you currently have to choose between disabling the systemd write protection on `/etc`, or putting your configuration in e.g. `/var`, which doesn't make real sense if you use `include-dir` to split the main pdns config file into multiple files.

### LINKED ISSUE #13233: rec: add automated conversion to YAML of recursor-maintained config files
The rec managed config files should automatically be converted to YAML. Also, we should allow (or maybe even enforce?) that `include-dir` and `api-dir` are separate dirs. See #12679

omoerbeek:
### Short description

Fixes #13233
Also fixes #12679 by requiring* incldue dir and api-dir are not the same (but only if YAML settings are active).

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [X] documented the code


omoerbeek:
Hmm, please wait a bit with reviewing, I might have introduced some last minute errors.

omoerbeek:
> Issues are fixed now, though our coverage cod barfs:

This is fixed now by https://github.com/PowerDNS/pdns/pull/13362/commits/5a3e2bc9da9f5076eb5a31c922ae6c8352aa2b2c

clang-tidy isn't 100% happy it cnanot find the generated file `cxx.h`, but that's a different issue which is going to be fixed in a separate PR.

pdns/recursordist/docs/http-api/zone.rst
```diff
@@ -32,7 +32,8 @@ be true:
   command line. Setting these options on the command line will
   override what has been set in the dynamically generated
   configuration files.
-* ``include-dir`` must refer to the same directory as
-  ``api-config-dir`` for the dynamic reloading to work.
+
+* For configuration changes to work :ref:`setting-include-dir` and :ref:`setting-api-config-dir` should have the same value for   old-style settings.
```
rgacogne, line 36:
```suggestion
* For configuration changes to work :ref:`setting-include-dir` and :ref:`setting-api-config-dir` should have the same value for old-style settings.
```

pdns/recursordist/settings/cxxsupport.cc
```diff
@@ -119,18 +122,172 @@ void pdns::settings::rec::oldStyleAllowFileToBridgeStruct(const std::string& fil
...
+  ofstream ofconf(tmpfilename);
+  if (!ofconf) {
+    int err = errno;
+    log->error(Logr::Error, err, "Cannot open for file for writing YAML", "to", Logging::Loggable(tmpfilename));
+    throw runtime_error("YAML Conversion");
+  }
+  ofconf << "# Generated by pdns-recursor REST API, DO NOT EDIT" << endl;
+  ofconf << yaml << endl;
+  ofconf.close();
+  if (rename(path.c_str(), (path + ".converted").c_str()) != 0) {
```
rgacogne, line 188: We might want to look at `rdstate()` (or `bad()`) after calling `close()` and before renaming?

pdns/recursordist/settings/cxxsupport.cc
```diff
@@ -119,18 +122,172 @@ void pdns::settings::rec::oldStyleAllowFileToBridgeStruct(const std::string& fil
...
+      throw runtime_error("YAML Conversion");
+    }
+  }
+  ofconf.close();
+  ifconf.close();
+}
+static void possiblyConvertForwardsandAuths(const std::string& includeDir, const std::string& apiDir, Logr::log_t log)
+{
+  std::vector<std::string> forwAndAuthFiles{};
+  ::arg().gatherIncludes(includeDir, "..conf", forwAndAuthFiles);
```
rgacogne, line 239: Unless I'm mistaken, the suffix is wrong:
```suggestion
  ::arg().gatherIncludes(includeDir, ".conf", forwAndAuthFiles);
```

rgacogne (changes requested):

pdns/recursordist/settings/cxxsupport.cc
omoerbeek: Nope, the recursor generated file names in the api dir for API managed forward and auth file are using double dots.

rgacogne: Well, today I learned.. :)

rgacogne (approved):

PR #13364, opened 2023-10-13
rec: If we miss glue, but not for all NS records, try to resolve the missing address records

omoerbeek:
There are several cases where we can have missing glue in cache for in-bailiwick NS:

- The auth did not sent us ADDITIONAL address records for all NS records
- The A/AAAA records are in cache, but expired
- The A/AAAA records were in cache, and not expired, but thy were evicted anyway because of cache cleaning under pressure

In these cases the trace contains `but needs glue` and we skip these NS records. Bu if we have at least one NS with associated address(es), we can try to resolve these missing names to addresses, so the next time we have more complete glue.

If *all* NS records lack glue, there's no chance the resolves wil succeed, so don't push tasks in that case, and let the existing logic do its work.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [X] documented the code


pdns/recursordist/syncres.cc
```diff
@@ -2288,6 +2292,7 @@ void SyncRes::getBestNSFromCache(const DNSName& qname, const QType qtype, vector
           else {
```
rgacogne, line 2292:
```suggestion
          else if (nrr != nullptr) {
```

rgacogne (changes requested):
I noticed one issue, but more generally I wonder if we should limit the number of tasks we are willing to yield for the NSs of a given domain. I'm worried this might be used for amplification otherwise.

rgacogne (approved):

PR #13353, opened 2023-10-11
rec: If serving stale, wipe CNAME records from cache when we get a NODATA negative response for them

omoerbeek:
PR #12395 already did that for the NXDOMAIN case.

### Checklist

I have:
- [X] read the [CONTRIBUTING.md](https://github.com/PowerDNS/pdns/blob/master/CONTRIBUTING.md) document
- [X] compiled this code
- [X] tested this code
- [ ] included documentation (including possible behaviour changes)
- [X] documented the code
- [ ] added or modified regression test(s)
- [X] added or modified unit test(s)

PR #13378, opened 2023-10-17
Followup to #13353, missed the suggested change; add recursor.yml to .gitignore

omoerbeek:
I accidentally merged the wrong version for  #13353

### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/recursordist/.gitignore
+++ b/pdns/recursordist/.gitignore
@@ -57,3 +57,4 @@ PowerDNS-Recursor.pdf
 /*.pb.h
 /.cache
 /.clang-tidy
+/recursor.yml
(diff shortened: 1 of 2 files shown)


rgacogne (approved):

omoerbeek:
clang-tidy trips because of the annotation removal...

fredmorcos:
What I don't see is what was changed in that function signature that caused `clang-tidy` on the CI to complain about it?

omoerbeek:
What happened is that I added the annotation because emacs flagged it. The add was not really needed, and now the drop of the annotation causes flagging in our CI, which should disappear in the next build.

fredmorcos:
I see. In general, unless the CI complains about it, please don't `// NOLINT` it. This just silences a warning that we should be fixing instead, and it makes it harder to spot in the future.

omoerbeek:
yes, noted.

PR #13382, opened 2023-10-17
builder-support: Enable missing dnsdist features on Amazon 2023, unbreak dnsdist package on EL8/9

rgacogne:
### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/builder-support/specs/dnsdist.spec
+++ b/builder-support/specs/dnsdist.spec
@@ -101,7 +101,7 @@ export RANLIB=gcc-ranlib
   --with-libsodium \
   --with-net-snmp \
 %endif
-%if 0%{?rhel} >= 8
+%if 0%{?rhel} >= 8 || 0%{?amzn} == 2023
   --enable-dns-over-quic \
   --with-quiche \
 %endif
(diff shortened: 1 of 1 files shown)


Habbie:
with

```
diff --git a/builder-support/specs/dnsdist.spec b/builder-support/specs/dnsdist.spec
index f03301fcc..df9d6b7c4 100644
--- a/builder-support/specs/dnsdist.spec
+++ b/builder-support/specs/dnsdist.spec
@@ -101,7 +101,7 @@ export RANLIB=gcc-ranlib
   --with-libsodium \
   --with-net-snmp \
 %endif
-%if 0%{?rhel} >= 8
+%if 0%{?rhel} >= 8 || 0%{?amzn} == 2023
   --enable-dns-over-quic \
   --with-quiche \
 %endif
```

I get `configure: DNS over QUIC (DoQ): yes`. Build is running, will report back.

Habbie:
> I get `configure: DNS over QUIC (DoQ): yes`. Build is running, will report back.

Patch is incomplete. Will fix and then push to this PR :)

Habbie:
```
dnsdist 1.9.0-alpha2.1.ddistfixamz2023.g5f51ec9e1.dirty (Lua 5.1.4 [LuaJIT 2.1.0-beta3])
Enabled features: cdb dns-over-quic dns-over-tls(gnutls openssl) dns-over-https(h2o nghttp2) ebpf fstrm ipcipher libedit lmdb protobuf re2 recvmmsg/sendmmsg systemd
```

(push coming)

Habbie:
I rebased everything and added a commit. 

The last commit is quite important.

Besides making DoQ be included in the build on amzn2023, it prevents the RPM (on any DoQ-supporting platform) from generating a libdnsdist-quiche dependency in package metadata, while not providing it. To get out of that situation, I added `%define __requires_exclude libdnsdist-quiche\\.so` to remove the dep.

@phonedph1 mentioned https://github.com/cloudflare/quiche/pull/599/files to me, which would solve it differently - the dep would still be generated, but the Provides would also be. However, I don't think we want our package to say 'provides libdnsdist-quiche', so the filter I committed makes the most sense to me.

Habbie (approved):

rgacogne:
> which is a great improvement, but I think we want DoQ to say 'yes'?

We do! :facepalm:

rgacogne (commented):
Thanks!

PR #13383, opened 2023-10-17
The refresh verification should be added to the negcache getRootNXTrust() function.

zjs604381586:
### Short description

When the qname of the domain name resolution request is an illegal and non-existent top-level domain, the root domain "." will return the response result of nxdomain. When the TTL of 3600s expires, if the resolution request is made again, and this resolution request triggers For the stale function, the asynchronous task generated at this time will not go out of the network and be requested again, because the stale function adds a TTL of 30 seconds and directly hits the cache in the **getRootNXTrust()** function and returns it. The expectation is that asynchronous tasks can be parsed out of the network. Therefore, the judgment logic of refresh and stale should be added when searching the cache of **negcache**.

         // We have something
          if (now.tv_sec < ni->d_ttd && !(refresh && ni->d_servedStale > 0)) {
            ne = *ni;
            moveCacheItemToBack<SequenceTag>(content->d_map, ni);
            return true;
          }

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


omoerbeek:
I wonder if it's possible to call `get(...)` in `getRootNXTrust()`, so we don't have to duplicate the refresh and serve-stale logic.

zjs604381586:
> I wonder if it's possible to call `get(...)` in `getRootNXTrust()`, so we don't have to duplicate the refresh and serve-stale logic.

Calling the **get** function in the **getRootNXTrust** function can reduce some code redundancy, but it will add some additional judgment logic, which will make it complicated for people to understand, and it feels like the single responsibility principle of the function will be lost.

omoerbeek:
I think I disagree, the diff I uploaded looks simpler that the alternative.
[x.diff.txt](https://github.com/PowerDNS/pdns/files/12991333/x.diff.txt)

zjs604381586:
> I think I disagree, the diff I uploaded looks simpler that the alternative. [x.diff.txt](https://github.com/PowerDNS/pdns/files/12991333/x.diff.txt)

I checked the logic before and after the modification. The modification looks good and the code is very concise.

omoerbeek:
OK, I'll make a PR from the attached diff soon.

omoerbeek:
Superseded by #13409

PR #13385, opened 2023-10-17
Assorted delinting, mark do while loops as ok

omoerbeek:
### Checklist

I have:
- [X] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/.clang-tidy.full
+++ b/.clang-tidy.full
@@ -1,5 +1,5 @@
 ---
-Checks:          'clang-diagnostic-*,clang-analyzer-*,cppcoreguidelines-*,bugprone-*,concurrency-*,modernize-*,performance-*,portability-*,readability-*,-modernize-use-trailing-return-type,-readability-magic-numbers,-cppcoreguidelines-avoid-magic-numbers,-cppcoreguidelines-pro-type-union-access,-cppcoreguidelines-avoid-non-const-global-variables,-cppcoreguidelines-pro-type-vararg'
+Checks:          'clang-diagnostic-*,clang-analyzer-*,cppcoreguidelines-*,bugprone-*,concurrency-*,modernize-*,performance-*,portability-*,readability-*,-modernize-use-trailing-return-type,-readability-magic-numbers,-cppcoreguidelines-avoid-magic-numbers,-cppcoreguidelines-pro-type-union-access,-cppcoreguidelines-avoid-non-const-global-variables,-cppcoreguidelines-pro-type-vararg,-cppcoreguidelines-avoid-do-while'
 WarningsAsErrors: ''
 HeaderFilterRegex: ''
 AnalyzeTemporaryDtors: false
--- a/pdns/dnsdistdist/test-dnsdist-dnsparser.cc
+++ b/pdns/dnsdistdist/test-dnsdist-dnsparser.cc
@@ -451,7 +451,7 @@ BOOST_AUTO_TEST_CASE(test_Overlay)
 
       BOOST_CHECK_EQUAL(record.d_name, target);
       BOOST_CHECK_EQUAL(record.d_class, QClass::IN);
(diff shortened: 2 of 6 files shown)


.clang-tidy.full
```diff
@@ -1,5 +1,5 @@
 ---
-Checks:          'clang-diagnostic-*,clang-analyzer-*,cppcoreguidelines-*,bugprone-*,concurrency-*,modernize-*,performance-*,portability-*,readability-*,-modernize-use-trailing-return-type,-readability-magic-numbers,-cppcoreguidelines-avoid-magic-numbers,-cppcoreguidelines-pro-type-union-access,-cppcoreguidelines-avoid-non-const-global-variables,-cppcoreguidelines-pro-type-vararg'
+Checks:          'clang-diagnostic-*,clang-analyzer-*,cppcoreguidelines-*,bugprone-*,concurrency-*,modernize-*,performance-*,portability-*,readability-*,-modernize-use-trailing-return-type,-readability-magic-numbers,-cppcoreguidelines-avoid-magic-numbers,-cppcoreguidelines-pro-type-union-access,-cppcoreguidelines-avoid-non-const-global-variables,-cppcoreguidelines-pro-type-vararg,-cppcoreguidelines-avoid-do-while'
```
fredmorcos, line 2: I disagree with a project-wide disabling of this.
The rest looks fine to me.

fredmorcos (approved):

.clang-tidy.full
rgacogne: Would you mind explaining why? I genuinely do not see the issue with do while loops. Bjarne's explanation has never made any sense to me :-/

fredmorcos: Not specific to `do-while` loops, but more that they might be a symptom of something that needs to be cleaned up properly. Example is the one I came across when cleaning up `UeberBackend`, it was a 3 page long `do-while` block with multiple `goto`s (granted they were only forward `goto`s), it was easy to lose sight of where I was. Disabling the warning project-wide makes the urgency of cleaning stuff like that up even less visible.

fredmorcos: And since we probably also want to avoid `goto`s, then something like:
```C++
while (true) {
    ...
    if (something) {
        break;
    }
    ...
    if (something-else) {
        break;
    }
    ...
    if (loop-condition) {
        break;
    }
}
```
should be fine and more coherent anyway.

omoerbeek: I suppose clang-tidy will still warn about the high complexity and the `goto`'s in that case.
IMO `do-while` loops can help in structuring code. They are just not loved very much because they are rare.
The syntactic "surprise" can be avoided by writing it in the recommended way with the while. after the closing brace:
```
do {
  ...
} while (condition);
```

rgacogne: While I 100% agree with `goto` being a huge red flag, I don't really agree that `do-while` is a sign that a clean-up is needed. I'm reviewing the PR you are mentioning right now and while you are doing a wonderful job of cleaning up that code, I don't see how:
```c+++
  bool first = true;
  while (first || CONDITON) {
    first = false;
    ...
  }
```
is better than:
```c+++
  do {
    ...
  }
  while (CONDITION);
```

fredmorcos: If we merge this PR first, I can revert `UeberBackend::getAuth` to using a do-while loop.

rgacogne: Thanks, I don't think reverting is needed, it was just an example to explain why I'm not convinced that the alternative to `do-while` is less error-prone :) I'm completely fine with being proven wrong about `do-while` loops, it's just that so far I have not seen any conclusive example of why they are more error-prone, so I'm wondering if this rule comes from cargo-cult or it is really useful.

omoerbeek: So I think the conclusion is I can merge as-is?

PR #13391, opened 2023-10-18
Add Lua function to pick records via name hash

devicenull:
### Short description
This adds a Lua function to return a record based on a weighted hash of the DNS record name.  One use case here is to consistently return the same IP address for a particular cache server based on what subdomain is requesting the data.

This is pretty much a copy and paste of pickwhashed

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/docs/lua-records/functions.rst
+++ b/docs/lua-records/functions.rst
@@ -237,7 +237,8 @@ Record creation functions
   :param values: table of weight, string (such as IPv4 or IPv6 address).
 
   This allows basic persistent load balancing across a number of backends.  It means that
-  test.example.com will always resolve to the same IP, but test2.example.com may go elsewhere.
+  test.mydomain.example.com will always resolve to the same IP, but test2.mydomain.example.com
+  may go elsewhere.  This function is only useful for wildcard records.
 
   This works similar to round-robin load balancing, but has the advantage of making traffic
   for the same domain always end up on the same server which can help cache hit rates.
@@ -246,7 +247,7 @@ Record creation functions
 
   An example::
 
-    mydomain.example.com    IN    LUA    A ("picknamehashed({                             "
+    *.mydomain.example.com    IN    LUA    A ("picknamehashed({                        "
                                             "        {15,  "192.0.2.1"},               "
                                             "        {100, "198.51.100.5"}             "
                                             "})                                        ")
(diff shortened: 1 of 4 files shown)


docs/lua-records/functions.rst
```diff
@@ -226,6 +226,29 @@ Record creation functions
...
+  test.example.com will always resolve to the same IP, but test2.example.com may go elsewhere.
+
+  This works similar to round-robin load balancing, but has the advantage of making traffic
+  for the same domain always end up on the same server which can help cache hit rates.
+
+  This function also works for CNAME or TXT records.
+
+  An example::
+
+    mydomain.example.com    IN    LUA    A ("picknamehashed({                             "
```
Habbie, line 247:
```suggestion
    mydomain.example.com    IN    LUA    A ("picknamehashed({                          "
```

regression-tests.auth-py/test_LuaRecords.py
```diff
@@ -790,6 +792,33 @@ def testWHashed(self):
...
+        Basic picknamehashed() test with a set of A records
+        As the name is hashed, we should always get the same IP back for the same record name.
+        """
+
+        queries = [
+            {
+                'expected': dns.rrset.from_text('test.namehashed.example.org.', 0,
+                                       dns.rdataclass.IN, 'A',
+                                       '1.2.3.4'),
+                'query': dns.message.make_query('test.namehashed.example.org', 'A')
```
Habbie, line 807: can you put `query` above `expected` in both dicts?

docs/lua-records/functions.rst
Habbie: also, perhaps the example makes more sense as a wildcard

Habbie (changes requested):
great work! I posted a couple of nits

devicenull:
@Habbie What are your thoughts on if you *want* this or not?  For our uses I was able to reimplement this entirely in Lua, and we'll probably go with that instead.

I'm perfectly willing to just close this MR unless you think this is going to be a useful feature for other people going forward.

I'm also willing to spend the time to get this across the line.  I don't have any real strong preferences either way.

Habbie:
> What are your thoughts on if you _want_ this or not?

_want_ is a very strong word. I think it's a neat feature, it comes with tests, which means it will be easy to maintain/keep, so I'd like to merge this after minor cleanup.

devicenull:
Code updated :)

Habbie (approved):

Habbie:
rebased to fix a conflict

Habbie:
> rebased to fix a conflict

I did not rebase correctly there. Should be fixed now.

PR #13392, opened 2023-10-18
rec: Describe the keys and values produced by structured logging in more detail

omoerbeek:
This describes the current state of affairs.

In the future, we should strive to unify the structured logging formats of the various products. We can either adapt the default backend if we want to (and accept it might be disruptive to existing scanning of structured logs), or we can define a new backend that conforms to the yet to be set standard.

### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/recursordist/docs/appendices/structuredlogging.rst
+++ b/pdns/recursordist/docs/appendices/structuredlogging.rst
@@ -37,7 +37,7 @@ The following keys are always present:
 |             |                  |                                      |entry together with the value of       |
 |             |                  |                                      |``msg``.                               |
 +-------------+------------------+--------------------------------------+---------------------------------------+
-| ``level``   |``number``        | ``"0"``                              |The detail level of the log entry, do  |
+| ``level``   |``number``        |``"0"``                               |The detail level of the log entry, do  |
 |             |                  |                                      |not confuse with                       |
 |             |                  |                                      |:ref:`setting-loglevel`. Not actively  |
 |             |                  |                                      |used currently.                        |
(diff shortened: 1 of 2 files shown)


pdns/recursordist/docs/appendices/structuredlogging.rst
```diff
@@ -0,0 +1,103 @@
...
+|``msg``      |``string``        | ``"Launching distributor threads"``  |Value is the same for all instances of |
+|             |                  |                                      |this log entry, together with          |
+|             |                  |                                      |``subsystem`` it uniquely identifies   |
+|             |                  |                                      |the log message.                       |
++-------------+------------------+--------------------------------------+---------------------------------------+
+|``subsystem``|``string``        |``"incoming"``                        |Uniquely identifies the log            |
+|             |                  |                                      |entry together with the value of       |
+|             |                  |                                      |``msg``.                               |
++-------------+------------------+--------------------------------------+---------------------------------------+
+| ``level``   |``number``        | ``"0"``                              |The detail level of the log entry, do  |
```
rgacogne, line 40: Some entries in these two tables have a leading white space, other do not. Do we care?

rgacogne (approved):

pdns/recursordist/docs/appendices/structuredlogging.rst
omoerbeek: I don't think it matters for the rendering, but I'll fix it anyway before merge.

PR #13408, opened 2023-10-22
Handle stack memory on NetBSD as on OpenBSD

tih:
### Short description

On NetBSD, memory used as stack should be R/W, and marked MAP_STACK, as on OpenBSD.

After a fresh build of the git HEAD of pdns on NetBSD/amd64-current, both updated to 2023-10-21 state,
the pdns_recursor would not run: the worker threads it created immediately died, throwing std::bad_alloc.

Otto Moerbeek suggested that the adjustments to the "lazy allocator" done for OpenBSD might be relevant.
A ktrace of the crashing recursor showed mprotect returning an error (with errno 13, EACCES) to each of
the threads just before they crashed, so I made the accompanying changes, which turned out to resolve
the situation.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/recursordist/lazy_allocator.hh
+++ b/pdns/recursordist/lazy_allocator.hh
@@ -27,8 +27,8 @@
 #include <sys/mman.h>
 #include <unistd.h>
 
-// On OpenBSD mem used as stack should be marked MAP_STACK
-#ifdef __OpenBSD__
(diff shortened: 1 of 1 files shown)


omoerbeek (approved):
Thanks!

PR #13413, opened 2023-10-24
dnsdist: Allow building of dnsdist against boringssl on OpenBSD

omoerbeek:
DISABLE_OCSP_STAPLING has to be defined as well, since boring does not seem to have it.

### Short description

Including a copy-paste and an include fix.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdistdist/m4/dnsdist_enable_doq.m4
+++ b/pdns/dnsdistdist/m4/dnsdist_enable_doq.m4
@@ -1,5 +1,5 @@
 AC_DEFUN([DNSDIST_ENABLE_DNS_OVER_QUIC], [
-  AC_MSG_CHECKING([whether to enable incoming DNS over QUIC (DoH) support])
+  AC_MSG_CHECKING([whether to enable incoming DNS over QUIC (DoQ) support])
   AC_ARG_ENABLE([dns-over-quic],
     AS_HELP_STRING([--enable-dns-over-quic], [enable incoming DNS over QUIC (DoQ) support (requires quiche) @<:@default=no@:>@]),
     [enable_dns_over_quic=$enableval],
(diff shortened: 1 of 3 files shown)


pdns/dnsdistdist/m4/dnsdist_enable_doq.m4
```diff
@@ -1,5 +1,5 @@
 AC_DEFUN([DNSDIST_ENABLE_DNS_OVER_QUIC], [
-  AC_MSG_CHECKING([whether to enable incoming DNS over QUIC (DoH) support])
+  AC_MSG_CHECKING([whether to enable incoming DNS over QUIC (DoQ) support])
```
rgacogne, line 2: Oh, nice one! :sweat_smile:

pdns/libssl.cc
```diff
@@ -288,7 +290,11 @@ int libssl_ticket_key_callback(SSL* /* s */, OpenSSLTLSTicketKeysRing& keyring,
   return 1;
 }
 
+#ifdef BORINGSSL_API_VERSION
+static int libssl_server_name_callback(SSL* ssl, int* al, void* arg)
+#else
```
rgacogne, line 295: `int` is also the correct return value with OpenSSL, apparently. The compiler doesn't complain because the OpenSSL header helpfully casts the callback function to `(void (*)(void))cb` :facepalm: 
I think you can just change it from `long` to `int`.

rgacogne (changes requested):
Thanks! I requested one change to fix a bug you did not introduce :)

rgacogne (approved):
Thanks!

rgacogne (approved):

PR #13414, opened 2023-10-24
CI: Move clang-tidy checks to the CodeQL workflow

rgacogne:
### Short description

Moving the clang-tidy checks from the build-and-test-all workflow to the CodeQL one, to avoid the shenanigans involved with building from a `dist` directory. This also has the nice properties of grouping the static analysis tools together. 
I also modified the `git-filter.py` script to to make sure that all modified files are processed in the context of the relevant products. This is important because even though the source file is the same, the compilation flags might differ causing some code to be left-out in some product and included in others. This is a bit tricky because we have several cases, for example:
- in ext/: ext/lmdb-safe/lmdb-safe.cc
- in modules/: modules/lmdbbackend/lmdbbackend.cc
- files that live in the dnsdist or rec dir only: pdns/dnsdistdist/dnsdist-dnsparser.cc or pdns/recursordist/rec-tcp.cc
- files that live in pdns/ and are used by several products (but possibly not with the same compilation flags, so it is actually important that they are processed for all products: pdns/misc.cc

Finally I modified the `clang-tidy.py` script which converts the YAML output to GitHub annotations: GitHub only supports 10 of these per job, the rest are not displayed, so we are now also outputting GitHub markdown step summaries which have no such limits.

<strike>The last commit of this PR is only there to trigger some warnings in various files, and should be removed before merging this PR, which is why this PR is only a draft for now.</strike>

### Checklist

I have:
- [ ] compiled this code
- [x] tested this code
- [x] documented the code


omoerbeek:
This looks really good. A bit sad  the static analysis run takes quite a bit of time, but that is mostly CodeQL's fault.
Read trough the changes and saw nothing that really caught my attention.

tasks.py
```diff
@@ -366,22 +368,36 @@ def ci_docs_add_ssh(c, ssh_key, host_key):
...
     if sanitizers != '':
         sanitizers = sanitizers.split('+')
         sanitizers = ['--enable-' + sanitizer for sanitizer in sanitizers]
         sanitizers = ' '.join(sanitizers)
     return sanitizers
 
+def get_c_compiler():
+    if os.getenv('COMPILER', 'clang') == 'clang':
+        return f'clang-{clang_version}'
+    return 'gcc'
```
fredmorcos, line 381:
```suggestion
def get_c_compiler():
    compiler = os.getenv('COMPILER', 'clang')
    return f'clang-{clang_version}' if compiler == 'clang' else 'gcc'
```

tasks.py
```diff
@@ -366,22 +368,36 @@ def ci_docs_add_ssh(c, ssh_key, host_key):
...
 def get_cflags():
     return " ".join([
-        "-O1",
+        get_optimizations(),
         "-Werror=vla",
         "-Werror=shadow",
         "-Wformat=2",
         "-Werror=format-security",
-        "-Werror=string-plus-int",
+        "-Werror=string-plus-int" if os.getenv('COMPILER', 'clang') == 'clang' else '',
```
fredmorcos, line 400: perhaps use `get_c_compiler()` here?

fredmorcos (approved):
I'm glad we've come full circle to what I originally wanted to do 🙃

rgacogne:
I removed the commit introducing some warnings, so this PR should now be in good shape to be merged once the CI is green.

tasks.py
```diff
@@ -366,22 +368,35 @@ def ci_docs_add_ssh(c, ssh_key, host_key):
...
         sanitizers = sanitizers.split('+')
         sanitizers = ['--enable-' + sanitizer for sanitizer in sanitizers]
         sanitizers = ' '.join(sanitizers)
     return sanitizers
 
+def get_c_compiler():
+    compiler = os.getenv('COMPILER', 'clang')
+    return f'clang-{clang_version}' if compiler == 'clang' else 'gcc'
+
+def get_cxx_compiler():
```
fredmorcos, line 382: Same here.

tasks.py
```diff
@@ -366,22 +368,35 @@ def ci_docs_add_ssh(c, ssh_key, host_key):
...
+    if os.getenv('COMPILER', 'clang') == 'clang':
+        return f'clang++-{clang_version}'
+    return 'g++'
+
+def get_optimizations():
```
fredmorcos, line 386: And here.

tasks.py
fredmorcos: Right. Something like `get_c_compiler().startswith('clang')`.

fredmorcos (approved):

tasks.py
```diff
@@ -366,22 +368,34 @@ def ci_docs_add_ssh(c, ssh_key, host_key):
...
+    return f'clang-{clang_version}' if compiler == 'clang' else 'gcc'
+
+def get_cxx_compiler():
+    compiler = os.getenv('COMPILER', 'clang')
+    return f'clang++-{clang_version}' if compiler == 'clang' else 'g++'
+
+def get_optimizations():
+    if os.getenv('OPTIMIZATIONS', 'yes') == 'yes':
+        return "-O1"
+    return "-O0"
```
fredmorcos, line 389:
```suggestion
def get_optimizations():
    optimizations = os.getenv('OPTIMIZATIONS', 'yes')
    return '-O1' if optimizations == 'yes' else '-O0'
```

fredmorcos (approved):

tasks.py
rgacogne: Applied as well

tasks.py
rgacogne: That does not look better to me than the current version, as it would introduce a new abstraction and immediately make an assumption about the value returned by that abstraction.

jsoref:
You could switch from github annotations to an additional sarif file, as check-spelling does: https://github.com/check-spelling/check-spelling/blob/ed14e107b2404ceec4db32161bc3cc353b123b46/action.yml#L574-L580

rgacogne:
> You could switch from github annotations to an additional sarif file, as check-spelling does: https://github.com/check-spelling/check-spelling/blob/ed14e107b2404ceec4db32161bc3cc353b123b46/action.yml#L574-L580

That sounds nice, but I suggest we do that in a follow up PR because right now our clang-tidy checks are broken, which means we are potentially missing important warnings.
@fredmorcos, @omoerbeek, do you want to take a look at the commits added since your last review?

fredmorcos (approved):
LGTM

PR #13423, opened 2023-10-26
rec: Update performance.rst

phonedph1:
### Short description

We should suggest people only stop tracking on client DNS queries, instead of more blanket rules.

The other rules allow source port 53 traffic in without any filtering.

Even on larger recursor instances out queries are generally small enough to be handled by stateful tracking defaults.

### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/recursordist/docs/performance.rst
+++ b/pdns/recursordist/docs/performance.rst
@@ -110,56 +110,36 @@ Connection tracking and firewalls
 A Recursor under high load puts a severe stress on any stateful (connection tracking) firewall, so much so that the firewall may fail.
 
 Specifically, many Linux distributions run with a connection tracking firewall configured.
-For high load operation (thousands of queries/second), It is advised to either turn off iptables completely, or use the ``NOTRACK`` feature to make sure DNS traffic bypasses the connection tracking.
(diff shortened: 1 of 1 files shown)


omoerbeek:
I agree with the intent, but I do not speak iptables. Can somebody knowledgable in this area check? Thanks!

ukleinek:
The dropped rules only affect incoming packets with srcport = 53 and outgoing packets. I would have expected lines 120 and 131 (both with `-t OUTPUT`) to be dropped, too.

phonedph1:
> The dropped rules only affect incoming packets with srcport = 53 and outgoing packets. I would have expected lines 120 and 131 (both with `-I OUTPUT`) to be dropped, too.

You need these to prevent state from being created from your responses back to the client. If you do not disable tracking on the outgoing responses you end up with no gains from doing this.

rgacogne (approved):
The removed lines make sense to me if we want to prevent tracking for incoming queries / outgoing responses only. I'm not 100% I agree that
> Even on larger recursor instances out queries are generally small enough to be handled by stateful tracking defaults.

On the other hand people running very large instances hopefully know what to do, or can infer it from the remaining instructions.

PR #13430, opened 2023-10-30
rec: Add endbr64 instructions in the right spots for OpenBSD/amd64

omoerbeek:
OpenBSD has IBT (aka control flow enforcement) on amd64, for processor models that support it. But it turns out that the existing endbr64 instructions in the boost assembly for context switching are not enough: there needs to be endbr64 instructions right after the jump_fcontext() call, as indirect branches end up there.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/recursordist/mtasker_fcontext.cc
+++ b/pdns/recursordist/mtasker_fcontext.cc
@@ -32,6 +32,15 @@ using boost::context::make_fcontext;
 using boost::context::detail::make_fcontext;
 #endif /* BOOST_VERSION < 106100 */
 
+// __CET__ is set by the compiler if relevant, so far only relevant/tested for amd64 on OpenBSD
+#if defined(__amd64__)
+#if __CET__ & 0x1
+#define CET_ENDBR __asm("endbr64")
+#else
+#define CET_ENDBR
+#endif
+#endif
+
 #ifdef PDNS_USE_VALGRIND
 #include <valgrind/valgrind.h>
 #endif /* PDNS_USE_VALGRIND */
@@ -132,6 +141,7 @@ extern "C"
                   static_cast<fcontext_t>(args->prev_ctx), 0);
 #else
     transfer_t res = jump_fcontext(t.fctx, 0);
+    CET_ENDBR;
     /* we got switched back from pdns_swapcontext() */
     if (res.data) {
       /* if res.data is not a nullptr, it holds a pointer to the context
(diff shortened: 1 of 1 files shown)


pdns/recursordist/mtasker_fcontext.cc
```diff
@@ -32,6 +32,15 @@ using boost::context::make_fcontext;
 using boost::context::detail::make_fcontext;
 #endif /* BOOST_VERSION < 106100 */
 
+// __CET__ is set by the compiler if relevant, so far only relevant/tested for amd64 on OpenBSD
+#if defined(__amd64__)
+#if __CET__ & 0x1
+#define CET_ENDBR __asm("endbr64")
```
rgacogne, line 38: Would it perhaps make sense to include `<cet.h>` and use `_CET_ENDBR`, which is set to `endbr64`, `endbr32` or nothing (when `__CET__ & 0x1` is `false`)?

pdns/recursordist/mtasker_fcontext.cc
```diff
@@ -132,6 +141,7 @@ extern "C"
                   static_cast<fcontext_t>(args->prev_ctx), 0);
 #else
     transfer_t res = jump_fcontext(t.fctx, 0);
+    CET_ENDBR;
```
rgacogne, line 144: Do you think it would be possible to contribute this upstream, or does it have to live in the code resumed from `jump_fcontext`?

omoerbeek: I think the latter.

pdns/recursordist/mtasker_fcontext.cc
rgacogne: Ah, too bad!

pdns/recursordist/mtasker_fcontext.cc
rgacogne: Alright :+1:

pdns/recursordist/mtasker_fcontext.cc
rgacogne: Do we care about the 32-bit case for the recursor?

rgacogne (approved):
With the caveat that I don't the right hardware to test this, the changes look OK to me. Nice work!

pdns/recursordist/mtasker_fcontext.cc
omoerbeek: > Do we care about the 32-bit case for the recursor?

AFAIK OpenBSD/i386 does not do enforcement. I might add the 32-bit case, but I hardly think it matters at this moment, and likely never will.

rgacogne: Makes sense to me, let's leave it out for now then.

PR #13434, opened 2023-10-31
rec: tidy syncres

omoerbeek:
### Short description

This is a pretty big beast. A few complaints remain. I marked a few functions with the "high-complexity" annotation. Not because I think we should not refactor them, but because I think that should be done in a separate PR. I did handle the more simple cases, though, as they were not very intrusive.
A changed conditions using the clang-tidy suggested substitutions, but given the recent experience with those, beware and double check (I did that as well, but it's easy to oversee something).

It also includes a few cases outside syncres.?? that were either needed (type changes) or just happened to catch my attention.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/recursordist/syncres.cc
+++ b/pdns/recursordist/syncres.cc
@@ -5158,7 +5158,7 @@ static void updateDoTStatus(ComboAddress address, DoTStatus::Status status, time
   auto iter = lock->d_map.find(address);
   if (iter != lock->d_map.end()) {
     iter->d_status = status;
-    lock->d_map.modify(iter, [=](DoTStatus& status) { status.d_ttd = time; });
+    lock->d_map.modify(iter, [=](DoTStatus& statusToModify) { statusToModify.d_ttd = time; });
     if (updateBusy) {
       --lock->d_numBusy;
     }
--- a/pdns/recursordist/syncres.cc
+++ b/pdns/recursordist/syncres.cc
@@ -58,9 +58,9 @@ public:
   };
 
   using cont_t = multi_index_container<value_t,
-                                indexed_by<
-                                  ordered_unique<tag<T>, member<value_t, T, &value_t::key>>,
-                                  ordered_non_unique<tag<time_t>, member<value_t, time_t, &value_t::last>>>>;
+                                       indexed_by<
+                                         ordered_unique<tag<T>, member<value_t, T, &value_t::key>>,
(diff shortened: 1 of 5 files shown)


omoerbeek:
Rebased to fix conflicts

pdns/recursordist/syncres.cc
```diff
@@ -909,25 +922,25 @@ bool SyncRes::AuthDomain::operator==(const AuthDomain& rhs) const
 [[nodiscard]] std::string SyncRes::AuthDomain::print(const std::string& indent,
                                                      const std::string& indentLevel) const
 {
-  std::stringstream s;
-  s << indent << "DNSName = " << d_name << std::endl;
-  s << indent << "rdForward = " << d_rdForward << std::endl;
-  s << indent << "Records {" << std::endl;
+  std::stringstream str;
```
fredmorcos, line 925: `domainInfoString` or `outputString` would be better here.

fredmorcos (approved):
Just one nitpick but looks fine. Thank you!

pdns/recursordist/syncres.cc
```diff
@@ -1608,15 +1621,15 @@ unsigned int SyncRes::s_max_minimize_count; // default is 10
 /* number of iterations that should only have one label appended */
 unsigned int SyncRes::s_minimize_one_label; // default is 4
 
-static unsigned int qmStepLen(unsigned int labels, unsigned int qnamelen, unsigned int i)
+static unsigned int qmStepLen(unsigned int labels, unsigned int qnamelen, unsigned int iter)
```
chbruyand, line 1624: `iter` sounds like an iterator to me. Maybe `index` or `idx` ?

pdns/recursordist/syncres.cc
```diff
@@ -4512,10 +4505,11 @@ RCode::rcodes_ SyncRes::updateCacheFromRecords(unsigned int depth, const string&
     }
   }
 
-  for (tcache_t::iterator i = tcache.begin(); i != tcache.end(); ++i) {
+  for (auto i = tcache.begin(); i != tcache.end(); ++i) {
```
chbruyand, line 4508: maybe rename `i` ?

chbruyand (approved):
Just a nit. LGTM otherwise.

omoerbeek:
Rebased to fix conflicts

PR #13441, opened 2023-11-01
auth: forward EDNS Client Subnet option during ALIAS processing

Habbie:
### Short description
replaces #11630 

### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsproxy.cc
+++ b/pdns/dnsproxy.cc
@@ -270,6 +270,7 @@ void DNSProxy::mainloop()
         MOADNSParser mdp(false, p.getString());
         if (p.d_eso.scope.isValid()){
           // update the EDNS options with info from the resolver - issue #5469
+          // note that this relies on the ECS string encoder to use the source network, and only take the prefix length from scope
           i->second.complete->d_eso.scope = p.d_eso.scope;
           DLOG(g_log<<"from dnsproxy::mainLoop: updated EDNS options from resolver EDNS source: "<<i->second.complete->d_eso.source.toString()<<" EDNS scope: "<<i->second.complete->d_eso.scope.toString()<<endl);
         }
--- a/pdns/dnsproxy.cc
+++ b/pdns/dnsproxy.cc
@@ -268,12 +268,10 @@ void DNSProxy::mainloop()
         memset(&msgh, 0, sizeof(struct msghdr));
         string reply; // needs to be alive at time of sendmsg!
         MOADNSParser mdp(false, p.getString());
-        if (p.d_eso.scope.isValid()){
-          // update the EDNS options with info from the resolver - issue #5469
-          // note that this relies on the ECS string encoder to use the source network, and only take the prefix length from scope
-          i->second.complete->d_eso.scope = p.d_eso.scope;
-          DLOG(g_log<<"from dnsproxy::mainLoop: updated EDNS options from resolver EDNS source: "<<i->second.complete->d_eso.source.toString()<<" EDNS scope: "<<i->second.complete->d_eso.scope.toString()<<endl);
-        }
+        // update the EDNS options with info from the resolver - issue #5469
+        // note that this relies on the ECS string encoder to use the source network, and only take the prefix length from scope
+        i->second.complete->d_eso.scope = p.d_eso.scope;
+        DLOG(g_log<<"from dnsproxy::mainLoop: updated EDNS options from resolver EDNS source: "<<i->second.complete->d_eso.source.toString()<<" EDNS scope: "<<i->second.complete->d_eso.scope.toString()<<endl);
 
         if (mdp.d_header.rcode == RCode::NoError) {
           for (const auto & answer : mdp.d_answers) {        
--- a/pdns/dnsproxy.cc
+++ b/pdns/dnsproxy.cc
@@ -42,98 +42,98 @@
 
 extern StatBag S;
 
-DNSProxy::DNSProxy(const string &remote): d_xor(dns_random_uint16())
+DNSProxy::DNSProxy(const string& remote) :
+  d_xor(dns_random_uint16())
 {
-  d_resanswers=S.getPointer("recursing-answers");
-  d_resquestions=S.getPointer("recursing-questions");
-  d_udpanswers=S.getPointer("udp-answers");
+  d_resanswers = S.getPointer("recursing-answers");
+  d_resquestions = S.getPointer("recursing-questions");
+  d_udpanswers = S.getPointer("udp-answers");
 
   vector<string> addresses;
   stringtok(addresses, remote, " ,\t");
   d_remote = ComboAddress(addresses[0], 53);
 
-  if((d_sock=socket(d_remote.sin4.sin_family, SOCK_DGRAM,0))<0) {
-    throw PDNSException(string("socket: ")+stringerror());
+  if ((d_sock = socket(d_remote.sin4.sin_family, SOCK_DGRAM, 0)) < 0) {
+    throw PDNSException(string("socket: ") + stringerror());
   }
 
   ComboAddress local;
-  if(d_remote.sin4.sin_family==AF_INET) {
+  if (d_remote.sin4.sin_family == AF_INET) {
     local = ComboAddress("0.0.0.0");
   }
   else {
     local = ComboAddress("::");
   }
-    
-  unsigned int n=0;
-  for(;n<10;n++) {
-    local.sin4.sin_port = htons(10000+dns_random(50000));
-    
-    if(::bind(d_sock, (struct sockaddr *)&local, local.getSocklen()) >= 0) 
+
+  unsigned int n = 0;
+  for (; n < 10; n++) {
+    local.sin4.sin_port = htons(10000 + dns_random(50000));
+
+    if (::bind(d_sock, (struct sockaddr*)&local, local.getSocklen()) >= 0)
       break;
   }
-  if(n==10) {
+  if (n == 10) {
     closesocket(d_sock);
-    d_sock=-1;
-    throw PDNSException(string("binding dnsproxy socket: ")+stringerror());
+    d_sock = -1;
+    throw PDNSException(string("binding dnsproxy socket: ") + stringerror());
   }
 
-  if(connect(d_sock, (sockaddr *)&d_remote, d_remote.getSocklen())<0) {
-    throw PDNSException("Unable to UDP connect to remote nameserver "+d_remote.toStringWithPort()+": "+stringerror());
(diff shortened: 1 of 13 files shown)


Habbie:
@jmazzi @bghira this PR replaces #11630

pdns/dnsproxy.cc
```diff
@@ -250,10 +272,14 @@ void DNSProxy::mainloop()
         memset(&msgh, 0, sizeof(struct msghdr));
         string reply; // needs to be alive at time of sendmsg!
         MOADNSParser mdp(false, p.getString());
-        //	  cerr<<"Got completion, "<<mdp.d_answers.size()<<" answers, rcode: "<<mdp.d_header.rcode<<endl;
+        if (p.d_eso.scope.isValid()){
```
rgacogne, line 275: I believe we need to check that the received source matches the one we sent, otherwise we might reply with garbage. Thinking about it, perhaps it would be better to only provide read-only access to `DNSPacket::d_eso` and to provide a `DNSPacket` method to request a cap on the ECS scope instead?

pdns/iputils.hh
```diff
@@ -261,6 +261,16 @@ union ComboAddress {
     return ret;
   }
 
+  bool isValid() const
+  {
+    char host[1024];
+    int retval = 0;
+    if(sin4.sin_family && !(retval = getnameinfo(reinterpret_cast<const struct sockaddr*>(this), getSocklen(), host, sizeof(host),0, 0, NI_NUMERICHOST)))
```
rgacogne, line 268: I'm not sure this check is actually useful?

pdns/stubresolver.cc
```diff
@@ -103,7 +105,7 @@ void stubParseResolveConf()
 }
 
 // s_resolversForStub contains the ComboAddresses that are used to resolve the
-int stubDoResolve(const DNSName& qname, uint16_t qtype, vector<DNSZoneRecord>& ret)
+int stubDoResolve(const DNSName& qname, uint16_t qtype, vector<DNSZoneRecord>& ret, EDNSSubnetOpts* d_eso)
```
rgacogne, line 108:
```suggestion
int stubDoResolve(const DNSName& qname, uint16_t qtype, vector<DNSZoneRecord>& ret, const EDNSSubnetOpts* d_eso)
```

pdns/stubresolver.cc
```diff
@@ -171,9 +183,9 @@ int stubDoResolve(const DNSName& qname, uint16_t qtype, vector<DNSZoneRecord>& r
   return RCode::ServFail;
 }
 
-int stubDoResolve(const DNSName& qname, uint16_t qtype, vector<DNSRecord>& ret) {
+int stubDoResolve(const DNSName& qname, uint16_t qtype, vector<DNSRecord>& ret, EDNSSubnetOpts* d_eso) {
```
rgacogne, line 186:
```suggestion
int stubDoResolve(const DNSName& qname, uint16_t qtype, vector<DNSRecord>& ret, const EDNSSubnetOpts* d_eso) {
```

pdns/stubresolver.hh
```diff
@@ -22,8 +22,9 @@
...
 #include "namespaces.hh"
 #include "dnsparser.hh"
+#include "ednssubnet.hh"
 
 void stubParseResolveConf();
 bool resolversDefined();
-int stubDoResolve(const DNSName& qname, uint16_t qtype, vector<DNSZoneRecord>& ret);
-int stubDoResolve(const DNSName& qname, uint16_t qtype, vector<DNSRecord>& ret);
+int stubDoResolve(const DNSName& qname, uint16_t qtype, vector<DNSZoneRecord>& ret, EDNSSubnetOpts* d_eso = nullptr);
+int stubDoResolve(const DNSName& qname, uint16_t qtype, vector<DNSRecord>& ret, EDNSSubnetOpts* d_eso = nullptr);
```
rgacogne, line 30:
```suggestion
int stubDoResolve(const DNSName& qname, uint16_t qtype, vector<DNSZoneRecord>& ret, const EDNSSubnetOpts* d_eso = nullptr);
int stubDoResolve(const DNSName& qname, uint16_t qtype, vector<DNSRecord>& ret, const EDNSSubnetOpts* d_eso = nullptr);
```

rgacogne (changes requested):

pdns/iputils.hh
```diff
@@ -261,6 +261,16 @@ union ComboAddress {
...
+      return true;
+    else
+      return false;
```
chbruyand, line 271:
```suggestion
     if (sin4.sin_family && getnameinfo(reinterpret_cast<const struct sockaddr*>(this), getSocklen(), host, sizeof(host), 0, 0, NI_NUMERICHOST) == 0) {                                                                                                                                                                                              
       return true;                                                                                                                                                                                                                                                                                                                                  
     }                                                                                                                                                                                                                                                                                                                                               
     return false;
```

pdns/dnsproxy.cc
```diff
@@ -250,10 +272,14 @@ void DNSProxy::mainloop()
...
+          // update the EDNS options with info from the resolver - issue #5469
+          i->second.complete->d_eso = p.d_eso;
+          DLOG(g_log<<"from dnsproxy::mainLoop: updated EDNS options from resolver EDNS source: "<<i->second.complete->d_eso.source.toString()<<" EDNS scope: "<<i->second.complete->d_eso.scope.toString()<<endl);
+        }
+
         if (mdp.d_header.rcode == RCode::NoError) {
-          for (auto& answer : mdp.d_answers) {
-            //	    cerr<<"comp: "<<(int)j->first.d_place-1<<" "<<j->first.d_label<<" " << DNSRecordContent::NumberToType(j->first.d_type)<<" "<<j->first.d_content->getZoneRepresentation()<<endl;
+          for(const auto & answer : mdp.d_answers) {
```
chbruyand, line 282:
```suggestion
          for (const auto & answer : mdp.d_answers) {        
```

pdns/dnsproxy.cc
```diff
@@ -91,14 +93,25 @@ void DNSProxy::go()
 //! look up qname target with r->qtype, plonk it in the answer section of 'r' with name aname
 bool DNSProxy::completePacket(std::unique_ptr<DNSPacket>& r, const DNSName& target,const DNSName& aname, const uint8_t scopeMask)
 {
+  bool pass_ecs = false;
+  string ECSOptionStr;
+
+  if (r->hasEDNSSubnet())
+  {
+    DLOG(g_log<<"dnsproxy::completePacket: Parsed edns source: "<<r->d_eso.source.toString()<<", scope: "<<r->d_eso.scope.toString()<<", family = "<<r->d_eso.scope.getNetwork().sin4.sin_family<<endl);
```
chbruyand, line 101: All the `DLOG()` calls could be formated a bit nicer (with spaces around `<<` for instance)

pdns/dnsproxy.cc
```diff
@@ -238,7 +260,7 @@ void DNSProxy::mainloop()
 
         DNSPacket p(false),q(false);
         p.parse(buffer,(size_t)len);
-        q.parse(buffer,(size_t)len);
+        //q.parse(buffer,(size_t)len); // removed.  never used?
```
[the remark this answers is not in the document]
Habbie: indeed, in fact it was gone already, will remove this line

pdns/dnsproxy.cc
Habbie: Indeed, the current code allows forwarding a response that does not match what was asked. I'll push a (failing) test for that.

When there is a mismatch, what do you think we should do? Drop ECS from the response? Just apply the scope?

pdns/iputils.hh
Habbie: I also wondered. This question stayed open in the original review, which I totally failed to reread before asking for your review

pdns/dnsproxy.cc
Habbie: (just apply the scope)

rgacogne: Agreed, applying the received scope to the initial ECS value should be fine.

pdns/dnsproxy.cc
```diff
@@ -91,14 +93,25 @@ void DNSProxy::go()
 //! look up qname target with r->qtype, plonk it in the answer section of 'r' with name aname
 bool DNSProxy::completePacket(std::unique_ptr<DNSPacket>& r, const DNSName& target,const DNSName& aname, const uint8_t scopeMask)
 {
+  bool pass_ecs = false;
+  string ECSOptionStr;
```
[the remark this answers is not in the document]
Habbie: good idea

pdns/iputils.hh
Habbie: I removed isValid entirely

pdns/dnsproxy.cc
```diff
@@ -91,14 +93,23 @@ void DNSProxy::go()
...
+  }
+
   if(r->d_tcp) {
     vector<DNSZoneRecord> ips;
     int ret1 = 0, ret2 = 0;
-
+    // rip out edns info here, pass it to the stubDoResolve
     if(r->qtype == QType::A || r->qtype == QType::ANY)
-      ret1 = stubDoResolve(target, QType::A, ips);
+      ret1 = stubDoResolve(target, QType::A, ips, &r->d_eso);
```
rgacogne, line 110: Is this going to send a `0.0.0.0/0` EDNS Client Subnet to the resolver if the initial query did not have ECS? If so, perhaps this would be better:
```suggestion
      ret1 = stubDoResolve(target, QType::A, ips, r->hasEDNSSubnet() ? &r->d_eso : nullptr);
```

pdns/dnsproxy.cc
```diff
@@ -91,14 +93,23 @@ void DNSProxy::go()
...
     if(r->qtype == QType::AAAA || r->qtype == QType::ANY)
-      ret2 = stubDoResolve(target, QType::AAAA, ips);
+      ret2 = stubDoResolve(target, QType::AAAA, ips, &r->d_eso);
```
rgacogne, line 112: Same question, of course :)

pdns/dnsproxy.cc
Habbie: Oh, good one. I did verify that clients (to the auth) that don't send ECS don't get ECS (see comment in `test_ALIAS.py` currently on line 76). I'll check this.

Habbie: Indeed, on TCP that happens. Will fix :)

Habbie: pushed!

rgacogne (approved):
Nice work!

PR #13442, opened 2023-11-01
Kill a bunch of warnings emitted by clang

omoerbeek:
A few are coming from boost, I disabled those by adding #pragma's. Ugly but I see no other way. It appears boost is not interested in fixing them, as they are present already for a very long time: https://github.com/boostorg/accumulators/issues/32

Als go back to the standard `do {... } while (0)` idiom for the `SLOG` macro, which exposed a few calls without trailing semicolon.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code


rgacogne (approved):

modules/geoipbackend/geoipbackend.cc
```diff
@@ -34,7 +34,10 @@
 #include <fstream>
 #include <filesystem>
 #include <utility>
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wshadow"
 #include <yaml-cpp/yaml.h>
+#pragma clang diagnostic pop
```
fredmorcos, line 40: GCC warns about these.

PR #13445, opened 2023-11-02
dnsdist: Fix Coverity CID 1523748: Performance inefficiencies in dolog.hh

rgacogne:
Coverity reports:
```
Performance inefficiencies (COPY_INSTEAD_OF_MOVE): "args" is
passed-by-value as parameter to "dolog" when it could be moved instead.
```
This PR implements perfect forwarding for the variadic template parameters of DNSdist's logging methods to address it, even though the performance of these functions should not matter much.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dolog.hh
+++ b/pdns/dolog.hh
@@ -58,7 +58,7 @@ inline void dolog(std::ostream& os, const char*s)
 }
 
 template<typename T, typename... Args>
-void dolog(std::ostream& os, const char* s, T value, Args&&... args)
+void dolog(std::ostream& os, const char* s, T value, const Args&... args)
 {
   while (*s) {
     if (*s == '%') {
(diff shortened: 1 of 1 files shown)


omoerbeek:
I'm a bit worried this approach would introduce "use after move" cases for logged value. Maybe a const ref type for `args` is a better safe approach?

rgacogne:
> I'm a bit worried this approach would introduce "use after move" cases for logged value. Maybe a const ref type for `args` is a better safe approach?

It shouldn't unless we start moving values passed to the log functions, unless I'm mistaken? But I agree that this doesn't really make sense for `args` to be mutable, so I'll try the approach you are suggesting!

rgacogne:
Done

omoerbeek:
> > I'm a bit worried this approach would introduce "use after move" cases for logged value. Maybe a const ref type for `args` is a better safe approach?
> 
> It shouldn't unless we start moving values passed to the log functions, unless I'm mistaken? But I agree that this doesn't really make sense for `args` to be mutable, so I'll try the approach you are suggesting!

Right, if the callers takes care it should not be a problem. But it's an easy to make mistakes (and compiler/linters are not always warning agains tjis in all cases.

omoerbeek (approved):

PR #13446, opened 2023-11-03
rec: for structured logging always log addresses including port

omoerbeek:
As documented. The way this is done is by defined a specific log function if SL is used. Could potentially be used for other cases.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/logging.hh
+++ b/pdns/logging.hh
@@ -69,7 +69,7 @@ struct is_to_string_available<T, std::void_t<decltype(std::to_string(std::declva
 {
 };
 
-// Same mechanism for t.toLogString()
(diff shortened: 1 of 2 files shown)


pdns/logging.hh
```diff
@@ -69,7 +69,7 @@ struct is_to_string_available<T, std::void_t<decltype(std::to_string(std::declva
 {
 };
 
-// Same mechanism for t.toLogString()
+// Same mechanism for t.toLogString() and t.toStucturedLogString()
```
rgacogne, line 72:
```suggestion
// Same mechanism for t.toLogString() and t.toStructuredLogString()
```

rgacogne (approved):

PR #13468, opened 2023-11-08
rec: fix ubsan error: using a value of 80 for bool

omoerbeek:
This is due to an unititalizded value, since the CXX code does not generate constructors for structs.

`fix settings/cxxsupport.cc:525:40: runtime error: load of value 80, which is not a valid value for type 'bool'`

Mostly harmless

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code


rgacogne (approved):
Looks good to me. I see the same construct on line 621 which might need the same fix.

rgacogne (approved):

PR #13483, opened 2023-11-13
When all NS IPs are throttled, there is a problem with the recursor's cachehit calculation.

zjs604381586:
### Short description

When all NS IPs are throttled, there is a problem with the cachehit calculation of recursorcache. When res is RCode::ServFail, the hit statistics of the recursor cache should also be cachemiss, not cachehit.
### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


pdns/recursordist/pdns_recursor.cc
```diff
@@ -1846,7 +1846,7 @@ void startDoResolve(void* arg) // NOLINT(readability-function-cognitive-complexi
     }
 
     if (comboWriter->d_mdp.d_header.opcode == static_cast<unsigned>(Opcode::Query)) {
-      if (resolver.d_outqueries != 0 || resolver.d_authzonequeries != 0) {
+      if (res == RCode::ServFail || resolver.d_outqueries != 0 || resolver.d_authzonequeries != 0) {
```
rgacogne, line 1849: While I agree with that the current behaviour is not right, we cannot simply rely on the rcode. We do serve `ServFail` answers from the cache, for example when a DNSSEC proof is broken, so a `ServFail` is not always a cache miss.

omoerbeek: Indeed, will look into a better condition. 

Also a nte to @zjs604381586: we prefer PRs that do not contain merge commits Start with a clean, up-to-date master, branch and only add commits relevant to the fix. Then use the branch to submit the PR.

zjs604381586: > Indeed, will look into a better condition.
> 
> Also a nte to @zjs604381586: we prefer PRs that do not contain merge commits Start with a clean, up-to-date master, branch and only add commits relevant to the fix. Then use the branch to submit the PR.

I see, thank you

omoerbeek:
Superseded by #13497 which has (I think) the proper fix and no spurious commits.

zjs604381586:
> Superseded by #13497 which has (I think) the proper fix and no spurious commits.

Great, this modification looks good

zjs604381586:
@omoerbeek There is another question. If we only determine the number of **d_throttledqueries**, when the **query blocked by 'dont-query'** is triggered in **SyncRes::throttledOrBlocked()** function, do we need to perform the operation of **d_throttledqueries++**?

PR #13485, opened 2023-11-13
Add `pdns::visit_directory()`, wrapping opendir/readdir/closedir

rgacogne:
### Short description

Follow-up to #13191 

### Checklist

I have:
- [x] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsbackend.cc
+++ b/pdns/dnsbackend.cc
@@ -112,7 +112,6 @@ vector<string> BackendMakerClass::getModules()
 
 void BackendMakerClass::load_all()
 {
-  // TODO: Implement this?
   auto directoryError = pdns::visit_directory(arg()["module-dir"], [this]([[maybe_unused]] ino_t inodeNumber, const std::string_view& name) {
     if (boost::starts_with(name, "lib") &&
         name.size() > 13 &&
--- a/pdns/dnsbackend.cc
+++ b/pdns/dnsbackend.cc
@@ -113,21 +113,16 @@ vector<string> BackendMakerClass::getModules()
 void BackendMakerClass::load_all()
 {
   // TODO: Implement this?
-  auto dir = std::unique_ptr<DIR, decltype(&closedir)>(opendir(arg()["module-dir"].c_str()), closedir);
-  if (!dir) {
-    g_log<<Logger::Error<<"Unable to open module directory '"<<arg()["module-dir"]<<"'"<<endl;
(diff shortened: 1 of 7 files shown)


pdns/dnsbackend.cc
```diff
@@ -113,21 +113,16 @@ vector<string> BackendMakerClass::getModules()
 void BackendMakerClass::load_all()
 {
   // TODO: Implement this?
```
fredmorcos, line 115: Is this a leftover that can now be removed?

pdns/misc.cc
```diff
@@ -1745,3 +1740,26 @@ bool constantTimeStringEquals(const std::string& a, const std::string& b)
...
 }
+
+namespace pdns
+{
+std::optional<std::string> visit_directory(const std::string& directory, const std::function<bool(ino_t inodeNumber, const std::string_view& name)>& visitor)
+{
+  auto dirHandle = std::unique_ptr<DIR, decltype(&closedir)>(opendir(directory.c_str()), closedir);
+  if (!dirHandle) {
+    auto err = errno;
+    return std::string("Error opening directory '" + directory + "': " + stringerror(err));
```
fredmorcos, line 1751: I would prefer to throw an exception with an object that contains the error message extracted from `errno` using `pdns::getMessageFromErrno()` directly.

pdns/dnsbackend.cc
rgacogne: I have no idea :-/ @Habbie, do you have any clue for us?

pdns/misc.cc
rgacogne: I find it hard to make the right choice between returning a potential error or throwing an exception. In that case I opted for the former because most callers seem to expect the directory not to exist, or not to be readable, so I felt that returning the error was better than forcing the caller to wrap the call in a `try .. catch` block, but I think it really is a matter of taste.

pdns/dnsbackend.cc
Habbie: There used to be `#ifndef WIN32` below this comment. My best guess is that the comment refers to implementing it for Windows.

Even if I'm wrong, clearly "this" is implemented, and the comment can go :-)

rgacogne: Great, thanks a lot! I just pushed a commit removing the comment :)

PR #13497, opened 2023-11-16
rec: take into account throttled queries when determining if we had a cachehit

omoerbeek:
Problem spotted by @zjs604381586 in #13483, though proposed fix was not correct.

### Checklist

I have:
- [X] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/recursordist/pdns_recursor.cc
+++ b/pdns/recursordist/pdns_recursor.cc
@@ -1846,7 +1846,7 @@ void startDoResolve(void* arg) // NOLINT(readability-function-cognitive-complexi
     }
 
     if (comboWriter->d_mdp.d_header.opcode == static_cast<unsigned>(Opcode::Query)) {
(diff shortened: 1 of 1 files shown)


rgacogne (approved):

zjs604381586:
@omoerbeek There is another question. If we only determine the number of **d_throttledqueries**, when the query **blocked by 'dont-query'** is triggered in **SyncRes::throttledOrBlocked()** function, do we need to perform the operation of **d_throttledqueries++**?

omoerbeek:
> @omoerbeek There is another question. If we only determine the number of **d_throttledqueries**, when the query **blocked by 'dont-query'** is triggered in **SyncRes::throttledOrBlocked()** function, do we need to perform the operation of **d_throttledqueries++**?

We do not record that into the SyncRes context, so there's no way to check that at this moment. I will make a note to look into the blocked case later.

Issue #10158, opened 2021-03-09
dnsdist: magic argument to addAction/makeRule is confusing


Habbie:
This came up because of a user who did `addAction('duckduckgo.com', ..)` and was surprised it also matched subdomains.

Then I dug a bit further.

1. `addAction` does not document that the first argument is magical (exactly like `makeRule`).
2. There's a word missing in 'Make a `NetmaskGroupRule()` or a `SuffixMatchNodeRule()`, depending on it is called. ' at https://dnsdist.org/rules-actions.html?highlight=addaction#makeRule
3. That same bit of docs does not clarify that both netmasks and domain suffixes can go into tables.
4. If the table contains both netmasks and domain suffixes, we match the netmasks and throw away the domain suffixes.

So, this at least wants more documentation, but I wonder if it's time to remove the magic from `addAction` and perhaps even drop `makeRule`.

PR #13500, opened 2023-11-16
dnsdist: Improve `NetmaskGroupRule`/`SuffixMatchNodeRule`, deprecate `makeRule`

rgacogne:
### Short description

- `NetmaskGroupRule` now accepts a string or list of strings, instead of requiring the use of `newNMG`
- `SuffixMatchNodeRule` also accepts a string or list of strings, instead of requiring the use of `newSuffixMatchNode`
- Passing a string or list of strings to `addAction` is now deprecated
- `makeRule` is now deprecated

Fixes https://github.com/PowerDNS/pdns/issues/10158

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdist-lua-actions.cc
+++ b/pdns/dnsdist-lua-actions.cc
@@ -2305,7 +2305,7 @@ static void addAction(GlobalStateHolder<vector<T> > *someRuleActions, const luad
   parseRuleParams(params, uuid, name, creationOrder);
   checkAllParametersConsumed("addAction", params);
 
-  auto rule = makeRule(var);
+  auto rule = makeRule(var, "addAction");
   someRuleActions->modify([&rule, &action, &uuid, creationOrder, &name](vector<T>& ruleactions){
     ruleactions.push_back({std::move(rule), std::move(action), std::move(name), std::move(uuid), creationOrder});
     });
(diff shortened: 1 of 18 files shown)


phonedph1:
I think the example config shipped might have some of these deprecated forms now.

```
-- addAction({"abuse.example.org.", "xxx."}, PoolAction("abuse"))
-- addAction("192.0.2.0/24", PoolAction("abuse"))
```

rgacogne:
> I think the example config shipped might have some of these deprecated forms now.
> 
> ```
> -- addAction({"abuse.example.org.", "xxx."}, PoolAction("abuse"))
> -- addAction("192.0.2.0/24", PoolAction("abuse"))
> ```

Good catch, thanks!

omoerbeek (changes requested):
Some comments on the docs, plus I'd like to see a few testcases for `makeRule` remain, e.g. in a `test_Deprecated.py` file. It would become too easy to break if not tested.

rgacogne:
> I'd like to see a few testcases for `makeRule` remain, e.g. in a `test_Deprecated.py` file. It would become too easy to break if not tested.

Good idea, done!

omoerbeek (approved):

PR #13501, opened 2023-11-16
rec: introduce a setting to allow RPZ duplicates, including a dup handling fix

### LINKED ISSUE #12842: Duplicate entries in RPZ files should be ignored
- Program: Recursor
 - Issue type: Feature request

### Short description
Loading a RPZ file with duplicate entries should not prevent loading that file

### Usecase
I use the ThreatFox RPZ zone file from abuse.ch to block C&C domains. The file sometimes contains duplicate entries. 

### Description
Loading a RPZ file containing duplicate entries should not fail with `msg="Exception while loading RPZ zone from file" error="Adding a QName-based filter policy of kind . but a policy of kind . already exists for the following name: soul-kissed.org" subsystem="luaconfig" level="0" prio="Error" tid="0" ts="1684447502.277" file="/etc/powerdns/threatfox.rpz`
It should instead ignore the duplicate entries, write a warning and finish loading the file

omoerbeek:
First the fix: the 2nd dup check in both `addNameTrigger` and `addNetmaskTrigger` was lacking a `!` I believe (unless I am misunderstanding things, so double check please!). While there, also fix the exception strings that were mixing up the new and existing policies.

Add a regress checks for that case, and also for the newly introduced `ignoreDuplicates` flag. Plus an example file from the draft for good measure.

Fixes #12842

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/recursordist/docs/lua-config/rpz.rst
+++ b/pdns/recursordist/docs/lua-config/rpz.rst
@@ -156,7 +156,14 @@ includeSOA
 .. versionadded:: 4.9.0
 
 Include the RPZ's SOA record to the reply's additional section if modified by a policy hit.
-Defaults to ``no``.
+Defaults to ``false``.
+
+ignoreDuplicates
+^^^^^^^^^^^^^^^^
+.. versionadded:: 5.0.0
+
+When loading an RPZ, ignore duplicate entries.
+Defaults to ``false``, duplicate entries will cause failure to load the zone.
 
 maxTTL
 ^^^^^^
--- a/pdns/recursordist/docs/lua-config/rpz.rst
+++ b/pdns/recursordist/docs/lua-config/rpz.rst
@@ -162,7 +162,7 @@ ignoreDuplicates
 ^^^^^^^^^^^^^^^^
 .. versionadded:: 5.0.0
 
-When loading an RPZ, ignore duplicate entries.
+When loading an RPZ, ignore duplicate entries, keeping only the first one present in the zone.
 Defaults to ``false``, duplicate entries will cause failure to load the zone.
 
 maxTTL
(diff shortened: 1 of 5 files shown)


pdns/recursordist/test-rpzloader_cc.cc
```diff
@@ -37,4 +40,135 @@ BOOST_AUTO_TEST_CASE(test_rpz_loader)
...
+                       "25.128.2.0.192.rpz-ip       MX      20 mx2.example.com\n"
+                       "25.128.2.0.192.rpz-ip       TXT     \"Contact Central Services\"\n"
+                       "25.128.2.0.192.rpz-ip       TXT     \"Your system is infected.\"\n";
+
+  auto rpz = makeFile(lines);
+
+  ::arg().set("max-generate-steps") = "1";
+  ::arg().set("max-include-depth") = "20";
+  auto zone = std::make_shared<DNSFilterEngine::Zone>();
+  cerr << "rpz " << rpz << endl;
```
rgacogne, line 111: This should go.

pdns/recursordist/docs/lua-config/rpz.rst
```diff
@@ -156,7 +156,14 @@ includeSOA
...
 
 Include the RPZ's SOA record to the reply's additional section if modified by a policy hit.
-Defaults to ``no``.
+Defaults to ``false``.
+
+ignoreDuplicates
+^^^^^^^^^^^^^^^^
+.. versionadded:: 5.0.0
+
+When loading an RPZ, ignore duplicate entries.
```
rgacogne, line 165:
```suggestion
When loading an RPZ, ignore duplicate entries, keeping only the first one present in the zone.
```

rgacogne (approved):
One left-over debug line that should be removed before merging, and a documentation suggestion, looks good to me otherwise.

PR #13502, opened 2023-11-16
systemd service: restart less agressively but still quicker

ukleinek:
If a service fails to start because it's not properly configured (e.g. directly after installation) the existing settings make the daemons restart in a loop. Each start logs several lines to the journal which quickly results in older but probably still relevant logs of other serives being rotated out.

Dropping RestartSec makes systemd use the default of 100ms. So a restart happens quicker. With dropping StartLimitInterval=0 systemd applies some ratelimiting and (by default) only starts the service 5 times per 10s.

Replacing Restart=on-failure by Restart=on-abnormal makes systemd not restart the service if the daemon process exits with a non-zero exit code. It's still restarted when the process exits by a signal (other than SIGHUP, SIGINT, SIGTERM or SIGPIPE).

Debian-Bug: https://bugs.debian.org/1055529

### Short description
I'm well aware that these settings are very much depending on personal preferences. But IMHO this is another hint that pdns should stick to system defaults and let the administrator fine tune the settings if needed. The motivation to change these settings came from installing the Debian package which in my case resulted in an annoying restart loop. See the linked Debian bug report. The Debian maintainer asked me to forward this issue upstream; here it is.

### Checklist
I have:
- [x] tested this code

### DIFF
--- a/pdns/dnsdistdist/dnsdist.service.in
+++ b/pdns/dnsdistdist/dnsdist.service.in
@@ -13,10 +13,8 @@ User=@service_user@
 Group=@service_group@
 SyslogIdentifier=dnsdist
 Type=notify
-Restart=on-failure
-RestartSec=2
+Restart=on-abnormal
 TimeoutStopSec=5
-StartLimitInterval=0
 
 # Tuning
 TasksMax=8192
--- a/pdns/ixfrdist.service.in
+++ b/pdns/ixfrdist.service.in
@@ -9,9 +9,7 @@ After=network-online.target time-sync.target
 [Service]
 Type=simple
 ExecStart=@bindir@/ixfrdist
-Restart=on-failure
-RestartSec=1
-StartLimitInterval=0
+Restart=on-abnormal
 
 # Sandboxing
 CapabilityBoundingSet=CAP_NET_BIND_SERVICE CAP_SETGID CAP_SETUID
--- a/pdns/pdns.service.in
+++ b/pdns/pdns.service.in
@@ -11,9 +11,7 @@ SyslogIdentifier=pdns_server
 User=@service_user@
 Group=@service_group@
 Type=notify
-Restart=on-failure
-RestartSec=1
-StartLimitInterval=0
+Restart=on-abnormal
 RuntimeDirectory=pdns
 
 # Sandboxing
--- a/pdns/recursordist/pdns-recursor.service.in
+++ b/pdns/recursordist/pdns-recursor.service.in
@@ -10,8 +10,7 @@ ExecStart=@sbindir@/pdns_recursor --daemon=no --write-pid=no --disable-syslog --
 User=@service_user@
 Group=@service_group@
 Type=notify
-Restart=on-failure
-StartLimitInterval=0
+Restart=on-abnormal
 RuntimeDirectory=pdns-recursor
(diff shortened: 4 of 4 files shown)


omoerbeek:
The reasoning appears solid, but this needs review by somebody that knows systemd (and that is not me).

zeha:
Seems like a bad time to make this change when systemd upstream seems to lean towards changing to a more "stay up" default configuration. Maybe wait for that to happen?
(Sorry don't have links at hand.)

pieterlexis:
> Seems like a bad time to make this change when systemd upstream seems to lean towards changing to a more "stay up" default configuration. Maybe wait for that to happen?

Should you have links nearby, that would be appreciated.

I think this PR makes sense in principle. But we don't know about *every* failure mode, the service might not come up if we catch an exception and call `exit(1)`.

zeha:
Sorry, I forgot what was happening there last year.
Looking at it now, I think a reasonable thing would be to restart automatically + aggressively if the service reports a successful startup to systemd, and if not, then not.

pieterlexis (approved):
re-reading this diff and the comments, I think this PR would be good. It should restart the service when it aborts/segfaults/OOM-killed, but keep it in a Failure state when it exits with a non-zero exit-code.

rgacogne (approved):
Agreed, that makes sense to me. Let's not backport this, though :)

mind04:
I'm no fan of this. This doesn't feel right for a service like DNS. Restarting is annoying, but a DNS outage is a major problem. Pick one...

rgacogne:
Do you see a case where a DNS service would start properly then exit with a non-zero exit code, rather than being terminated by a signal? If so then I agree this is a bad idea.

mind04:
I don't have any direct examples, but the "Exiting because communicator thread died" has a long history. There is a lot of exit(1) in the authoritative server, and every one of them could be fatal after this.

zeha:
at least in the past mysql being unavailable (briefly?) was a problem like this. but indeed if the burden for a service that should stay up shall be shifted to admins instead of the default install, then this is the way.

rgacogne:
> I don't have any direct examples, but the "Exiting because communicator thread died" has a long history. There is a lot of exit(1) in the authoritative server, and every one of them could be fatal after this.

That's a fair point, so I think we might want to investigate whether there is a way to tell systemd not to restart us if we failed to start initially (perhaps via `sd_notify`?) but to always restart us after that.

rgacogne:
There is `RestartPreventExitStatus` that could be useful, but we would need to be very consistent in our return code if we fail during the initialization phase.

pieterlexis:
> There is `RestartPreventExitStatus` that could be useful, but we would need to be very consistent in our return code if we fail during the initialization phase.

I like that idea, perhaps we can implement this in dnsdist 2.2 and see if it works well before implementing this in the other products.

mind04:
> at least in the past mysql being unavailable (briefly?) was a problem like this. 

That one is still present
```
Jun 02 10:29:58 Exiting because communicator thread died with error: virtual void GSQLBackend::getUnfreshSecondaryInfos(vector<DomainInfo> *) unable to retrieve list of secondary domains: Unable to connect to database: ERROR 2002 (HY000): Can't connect to local server through socket '/var/lib/mysql/mysql.sock' (2)
```

PR #13504, opened 2023-11-17
rec: introduce command to set aggressive NSEC cache size

### LINKED ISSUE #13265: set-aggressive-nsec-cache-size rec_control command
- Program: Recursor
 - Issue type: Feature request

### Short description

`rec_control` supports changing the cache size at runtime without a restart (`set-max-cache-entries`)
but there is no such command for the [NSEC cache](https://doc.powerdns.com/recursor/settings.html#aggressive-nsec-cache-size).

### Usecase

The use case is limited to debugging for us. 

We only file this feature request because Otto asked us to submit it :)

omoerbeek:
To avoid race conditions creating and destroying the cache, we do not allow setting the cache size if it is not enabled from startup. Also, setting the cache size to 0 with this command does not disable the cache, it just makes it mostly ineffective (entries can still appear in it and get used until the purge task is run). 

Fixes #13265.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/recursordist/aggressive_nsec.hh
+++ b/pdns/recursordist/aggressive_nsec.hh
@@ -48,6 +48,11 @@ public:
   {
   }
 
+  void setMaxEntries(uint64_t number)
+  {
+    d_maxEntries = number;
+  }
+
   static bool nsec3Disabled()
   {
     return s_maxNSEC3CommonPrefix == 0;
--- a/pdns/recursordist/aggressive_nsec.hh
+++ b/pdns/recursordist/aggressive_nsec.hh
@@ -21,6 +21,7 @@
  */
 #pragma once
 
+#include <atomic>
 #include <boost/utility.hpp>
 #include <boost/multi_index_container.hpp>
 #include <boost/multi_index/ordered_index.hpp>
@@ -162,7 +163,7 @@ private:
   pdns::stat_t d_nsecWildcardHits{0};
   pdns::stat_t d_nsec3WildcardHits{0};
   pdns::stat_t d_entriesCount{0};
-  uint64_t d_maxEntries{0};
+  std::atomic<uint64_t> d_maxEntries{0};
 };
 
 extern std::unique_ptr<AggressiveNSECCache> g_aggressiveNSECCache;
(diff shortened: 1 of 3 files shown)


pdns/recursordist/docs/manpages/rec_control.1.rst
```diff
@@ -227,6 +227,12 @@ set-dnssec-log-bogus *SETTING*
 set-ecs-minimum-ttl *NUM*
     Set ecs-minimum-ttl-override to *NUM*.
 
+set-max-aggr-nsec-cache-size *NUM*
+    Change the maximum number of entries in the NSEC aggressive cache. If the
+    cache is disabled by setting its size to 0 in the config, the cache size
+    cannot be set by this command. Setting the size to 0 by this command still
+    keeps the cache, but makes it mostly ineffective as it emptied periodically.
```
rgacogne, line 234:
```suggestion
    keeps the cache, but makes it mostly ineffective as it is emptied periodically.
```

pdns/recursordist/rec_channel_rec.cc
```diff
@@ -2085,7 +2104,8 @@ static RecursorControlChannel::Answer help()
           "reload-lua-config [filename]     (re)load Lua configuration file\n"
           "reload-zones                     reload all auth and forward zones\n"
           "set-ecs-minimum-ttl value        set ecs-minimum-ttl-override\n"
-          "set-max-cache-entries value      set new maximum cache size\n"
+          "set-max-aggr-nsec-cache-size     set new maximum aggressive NSEC cache size\n"
```
rgacogne, line 2107:
```suggestion
          "set-max-aggr-nsec-cache-size value set new maximum aggressive NSEC cache size\n"
```
Although it means re-aligning the existing entries, I'm afraid :-/

pdns/recursordist/aggressive_nsec.hh
```diff
@@ -48,6 +48,11 @@ public:
   {
   }
 
+  void setMaxEntries(uint64_t number)
+  {
+    d_maxEntries = number;
```
rgacogne, line 53: Would it make sense to turn `d_maxEntries` into a `std::atomic<uint64_t>`, if only to make Thread Sanitizer happy?

rgacogne (approved):

Issue #9875, opened 2020-12-16
dnsdist: provide async gethostbyname Lua function


wojas:
- Program: dnsdist
 - Issue type: Feature request

### Short description

Provide an async `gethostbyname` (or better name) function in dnsdist Lua to perform a DNS lookup using the system resolver.

This is primarily useful for cloud deployments where endpoints are only known by domain name, and for certain HA or dynamic setups.

### Usecase

Let's consider a simple deployment of dnsdist and the recursor in plain Docker for this example.

In this case we want to spin up two contains: one with dnsdist, one with the recursor. The IP address of the recursor will be dynamically assigned, but dnsdist needs to know this IP address in order to route queries to it.

One possible way of implementing the dnsdist container is:

- Create an entrypoint script that waits until the container with name 'recursor' is available
- Once it's available, create a dnsdist config from a template, filling in the IP address.
- Only then start dnsdist

This is of course quite painful and fragile, but it works...

..kind of. Consider what happens if the recursor container is recreated with a new IP address. The dnsdist container will no longer be sending traffic to the right IP address. An administrator need to know to also restart or recreate the dnsdist service in order to make the system work again, even though the dnsdist service did not itself require any changes.

The first situation (only perform the lookup on startup) could be handled by creating a blocking `gethostbyname` function, for example with FFI. In the second situation we need to watch for changes and blocking function would be undesirable, as they would block the entire Lua state for the duration of the request.

PR #9093 added official Docker images with a workaround that dynamically updates the upstreams when the IP address changes using an external helper script. While clever and effective, I think few people would call this solution elegant, and it could block on filesystem I/O.

### Description

This can all be solved by providing a way to perform an async `gethostbyname` from Lua.

On the Lua side this means that the function would take a callback parameter:

```lua
// Only on startup
gethostbyname("recursor.backend.example", function (ips)
    for _, ip in ipairs(ips) do
        newServer(ip)
    end
end)
```

The C++ code would work as follows:

- Create a new thread and pass it the query and callback reference
- The new thread would perform a blocking `gethostbyname`
- When it receives the result, it grabs a lock on the Lua state and calls the callback function with a list of results.
- The thread is terminated.

The same mechanism could also be used to dynamically update the remotes every fixed interval by adding a `maintenance` callback, performing an async `gethostbyname` in there, comparing the result to the previous result, and calling `newServer` and `rmServer` as appropriate, all without blocking dnsdist.

[...]

PR #13505, opened 2023-11-17
dnsdist: Add `getAddressInfo()` for asynchronous DNS resolution

rgacogne:
### Short description

Closes https://github.com/PowerDNS/pdns/issues/9875

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdistdist/dnsdist-resolver.cc
+++ b/pdns/dnsdistdist/dnsdist-resolver.cc
@@ -1,0 +1,52 @@
+/*
+ * This file is part of PowerDNS or dnsdist.
+ * Copyright -- PowerDNS.COM B.V. and its contributors
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of version 2 of the GNU General Public License as
+ * published by the Free Software Foundation.
+ *
+ * In addition, for the avoidance of any doubt, permission is granted to
+ * link this program with OpenSSL and to (re)distribute the binaries
+ * produced as the result of such linking.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+#include <vector>
+
+#include "dnsdist-resolver.hh"
+#include "iputils.hh"
+
+namespace dnsdist::resolver
+{
+void asynchronousResolver(const std::string& hostname, const std::function<void(const std::string& hostname, std::vector<ComboAddress>& ips)>& callback)
+{
+  addrinfo hints{};
+  hints.ai_family = AF_UNSPEC;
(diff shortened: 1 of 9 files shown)


pdns/dnsdistdist/docs/reference/config.rst
```diff
@@ -1812,6 +1812,25 @@ These values can be set at configuration time via:
 Other functions
 ---------------
 
+.. function:: getAddressInfo(hostname, callback)
+
+  .. versionadded:: 1.9.0
+
+  Asynchronously resolve, via the system resolver (using ``getaddrinfo()``), the supplied ``hostname`` to IPv4 and IPv6 before invoking the supplied ``callback`` function with the ``hostname`` and a list of IPv4 and IPv6 addresses as :class:`ComboAddress`.
+  For example, to get the addresses of Quad9's and dynamically add them as backends:
```
Habbie, line 1820:
```suggestion
  For example, to get the addresses of Quad9's resolver and dynamically add them as backends:
```

rgacogne: Oops, thanks!

Habbie:
After we merge this, we should update `docker/dnsdist*` to use it, and drop the python file.

pdns/dnsdistdist/dnsdist-resolver.cc
```diff
@@ -0,0 +1,52 @@
...
+#include "iputils.hh"
+
+namespace dnsdist::resolver
+{
+void asynchronousResolver(const std::string& hostname, const std::function<void(const std::string& hostname, std::vector<ComboAddress>& ips)>& callback)
+{
+  addrinfo hints{};
+  hints.ai_family = AF_UNSPEC;
+  hints.ai_flags = AI_V4MAPPED | AI_ADDRCONFIG;
+  hints.ai_socktype = SOCK_DGRAM;
```
omoerbeek, line 34: you should also set `ai_protocol`
`AI_V4MAPPED` might not be generally available

rgacogne: Good point about `AI_V4MAPPED`, in fact I don't think we actually want to set it (it's the default on Linux but not everywhere indeed).
I'm genuinely curious as to what is the point of setting `ai_protocol`, though?

rgacogne:

pdns/dnsdistdist/docs/reference/config.rst
```diff
@@ -1915,6 +1915,25 @@ These values can be set at configuration time via:
 Other functions
 ---------------
 
+.. function:: getAddressInfo(hostname, callback)
+
+  .. versionadded:: 1.9.0
+
+  Asynchronously resolve, via the system resolver (using ``getaddrinfo()``), the supplied ``hostname`` to IPv4 and IPv6 before invoking the supplied ``callback`` function with the ``hostname`` and a list of IPv4 and IPv6 addresses as :class:`ComboAddress`.
```
omoerbeek, line 1922: We use `AI_ADDRCONFIG`, so we only get A if v4 is configured and AAAA if v6. 
```suggestion
  Asynchronously resolve, via the system resolver (using ``getaddrinfo()``), the supplied ``hostname`` to IPv4 and IPv6 addresses (if configured on the host) before invoking the supplied ``callback`` function with the ``hostname`` and a list of IPv4 and IPv6 addresses as :class:`ComboAddress`.
```

omoerbeek (approved):
One small nit

bondar-aleksandr:
team, how can I use this feature to dynamically refresh list of real-servers?

PR #13507, opened 2023-11-17
rec: disallow (by answering Refused) RD=0 by default

### LINKED ISSUE #13386: rec: ponder introducing an ACL for RD=0 queries
Currently we handle RD=0 queries for all client allowed to send queries. We might restrict those queries, to make cache snooping harder.

Options are

- always disallow RD=0 queries
- only allow RD=0 queries for client addresses in `allow-rd0-from` ACL
- make `allow-rd0-from` dynamically settable via the REST API, like the existing `allow-from` and `allow-notify-from`.
- action on disallow: either drop or answer REFUSED
IMO, a settable ACL is nice (default empty), but I see little need for the REST API part. 

This is what others do:

- Knot has global switch with answer REFUSED, default: RD=0 is not allowed.
- Unbound has it tied to an ACL, action (refuse/drop)depends on other ACLs
- BInd has an ACL for this (allow-query-cache, exact semantics depend on if it is configured for recursion only, auth only or mixed), from reading it I think REFUSED is answered for non allowed RD=0 queries.

omoerbeek:
Fixes #13386 using the option 2: send `Refused` before actual resolving is done. Subsequent queries for the same name with RD=0 wil be answered by the packet cache.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code


rgacogne (approved):
PR looks good to me, and the new feature makes sense as well. A regression test checking that RD=0 queries are actually refused might be nice?

omoerbeek:
Yes, will add test before merge.

PR #13513, opened 2023-11-20
Fix unnecessary copies reported by Coverity

rgacogne:
### Short description

Almost all of these are in places where performance does not matter much, but it is still nice to clean it up when it doesn't make the code more complicated/brittle.

While I was at it I refactored `cleanSlashes()` to avoid the looking back to the previous value, which might be error-prone, so this change in particular might need a good review.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/test-misc_hh.cc
+++ b/pdns/test-misc_hh.cc
@@ -403,6 +403,9 @@ BOOST_AUTO_TEST_CASE(test_CleanSlashes) {
     cleanSlashes(fullStr);
     return fullStr;
   };
+  BOOST_CHECK_EQUAL(cleanSlashesWrapper(""), "");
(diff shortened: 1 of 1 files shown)


fredmorcos (approved):

omoerbeek (approved):
One nit: test more edge cases

rgacogne:
> One nit: test more edge cases

Added, thanks!

Habbie (approved):

rgacogne:
I'm going to merge this PR with some warnings from clang-tidy, since it has three approvals, but I'll open a follow-up PR fixing these warnings in a few minutes.

Issue #12922, opened 2023-06-16
dnsdist: incoming PROXY protocol as option to addLocal


wojas:
- Program: dnsdist
 - Issue type: Feature request

### Short description

Support enabling incoming PROXY protocol support on a per-port basis using a new `addLocal` option.

### Usecase

- Debugging filtering in a containerized environment
- Debugging filtering without losing the ability to do normal DNS queries without PROXY protocol from that address
- Support a mixed environment where the IPs of a load balancer with PROXY support are not known in advance, and normal DNS queries also need to be supported.

### Description

Currently `setProxyProtocolACL` can be used to enable PROXY protocol for incoming queries for certain addresses. When enabled, the PROXY protocol is also required from these configured sources.

This complicates debugging in development and production deployments that use containers. Overriding the source IP with the PROXY protocol can be very useful to test filtering when a query comes from specific source IPs (e.g. using `kdig +proxy=0.1.2.3-0.0.0.0`), but if you enable it for all IPs in a container, you can no longer perform normal queries without PROXY header against it.

If it were possible to enable the PROXY protocol on a per-port basis, every deployment could expose an additional port with the PROXY protocol enabled for debugging, that is only accessible from trusted IPs.

Other servers like nginx make this configurable on a per-port basis: `listen 12345 proxy_protocol;`.

Example:

```lua
addLocal('0.0.0.0:53')
addLocal('192.0.2.1:5300', {proxyProtocol=true})
```

### Alternative

Alternatively, dnsdist could allow ordinary queries without the PROXY protocol headers on `setProxyProtocolACL` IPs.

This would solve debugging issues for development environment, but have reduced utility in production environments and result in a higher risk of security issues due to misconfiguration.

franklouwers:
Thanks Remi. Any chance we’d see this for rec/auth one day?On 8 Dec 2023, at 17:14, Remi Gacogne ***@***.***> wrote:﻿
Closed #12922 as completed via #13517.

omoerbeek:
> Thanks Remi. Any chance we’d see this for rec/auth one day?On 8 Dec 2023, at 17:14, Remi Gacogne ***@***.***> wrote:﻿ Closed #12922 as completed via #13517. —Reply to this email directly, view it on GitHub, or unsubscribe.You are receiving this because you commented.Message ID: ***@***.***>

Testing PowerGit's handling of funny chars

rgacogne:
[...]

PR #13517, opened 2023-11-20
dnsdist: Allow enabling incoming PROXY protocol on a per-bind basis

rgacogne:
### Short description

The per-bind option defaults to `true` as to not break existing configuration, but setting `allowProxyProtocol=false` on a `add*Local()` directive disables proxy-protocol handling for this specific bind.

Closes https://github.com/PowerDNS/pdns/issues/12922

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdistdist/docs/reference/config.rst
+++ b/pdns/dnsdistdist/docs/reference/config.rst
@@ -84,7 +84,7 @@ Listen Sockets
     Added ``maxInFlight`` and ``maxConcurrentTCPConnections`` parameters.
 
   .. versionchanged:: 1.9.0
-    Added ``allowProxyProtocol`` parameter, which was always ``true`` before 1.9.0.
+    Added ``enableProxyProtocol`` parameter, which was always ``true`` before 1.9.0.
 
   Add to the list of listen addresses. Note that for IPv6 link-local addresses, it might be necessary to specify the interface to use: ``fe80::1%eth0``. On recent Linux versions specifying the interface via the ``interface`` parameter should work as well.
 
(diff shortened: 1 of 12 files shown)


rgacogne:

chbruyand (approved):

pdns/dnsdistdist/docs/reference/config.rst
```diff
@@ -99,6 +102,7 @@ Listen Sockets
   * ``tcpListenQueueSize=SOMAXCONN``: int - Set the size of the listen queue. Default is ``SOMAXCONN``.
   * ``maxInFlight=0``: int - Maximum number of in-flight queries. The default is 0, which disables out-of-order processing.
   * ``maxConcurrentTCPConnections=0``: int - Maximum number of concurrent incoming TCP connections. The default is 0 which means unlimited.
+  * ``allowProxyProtocol=true``: str - Whether to allow a proxy protocol v2 header in front of incoming queries. Both this option and :func:`setProxyProtocolACL` needs to agree for a payload to be accepted. Default is ``true``, meaning that queries are allowed if they come from an address present in the :func:`setProxyProtocolACL` ACL.
```
Habbie, line 105: is it really "allow"? does it still accept packets without a proxy header?

rgacogne: It's a very good point! No, it doesn't accept packets without a proxy protocol header, so the name is wrong.

Habbie: I just checked the auth for inspiration, but it uses `proxy-protocol-from`, which is not a form that helps you. Maybe something as simple as `enableProxyProtocol`?

rgacogne: `enableProxyProtocol` sounds good, thanks.

rgacogne: Done!

PR #13518, opened 2023-11-20
Fix a few warnings from clang-tidy that were not fixed in #13513

rgacogne:
### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/rcpgenerator.cc
+++ b/pdns/rcpgenerator.cc
@@ -255,16 +255,19 @@ void RecordTextReader::xfrName(DNSName& val, bool, bool)
   skipSpaces();
   DNSName sval;
 
-  const char* strptr=d_string.c_str();
   string::size_type begin_pos = d_pos;
   while (d_pos < d_end) {
-    if (strptr[d_pos]!='\r' && dns_isspace(strptr[d_pos])) {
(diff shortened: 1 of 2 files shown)


pdns/rcpgenerator.cc
```diff
@@ -255,16 +255,19 @@ void RecordTextReader::xfrName(DNSName& val, bool, bool)
...
+    if (d_string[d_pos]!='\r' && dns_isspace(d_string[d_pos])) {
       break;
     }
 
     d_pos++;
   }
-  sval = DNSName(std::string(strptr+begin_pos, strptr+d_pos));
+
+  {
+    std::string_view view(d_string);
```
omoerbeek, line 268: What's the reason to create a view here? Is `d_string.substr(begin_pos, d_pos - begin_pos)` not good?

rgacogne: It works, but it creates an unnecessary `std::string`, so potentially one allocation and one copy, while we only need a view/slice.

omoerbeek: Ah, right, of course ` std::string_view substr()` returns view itself

omoerbeek (approved):

PR #13522, opened 2023-11-21
Fix unnecessary copies reported by Coverity (round 2)

rgacogne:
### Checklist

I have:
- [x] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/tcpreceiver.cc
+++ b/pdns/tcpreceiver.cc
@@ -197,10 +197,10 @@ static bool maxConnectionDurationReached(unsigned int maxConnectionDuration, tim
 {
   if (maxConnectionDuration) {
     time_t elapsed = time(nullptr) - start;
-    if (elapsed >= maxConnectionDuration) {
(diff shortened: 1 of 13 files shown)


pdns/tcpreceiver.cc
```diff
@@ -197,10 +197,10 @@ static bool maxConnectionDurationReached(unsigned int maxConnectionDuration, tim
 {
   if (maxConnectionDuration) {
     time_t elapsed = time(nullptr) - start;
-    if (elapsed >= maxConnectionDuration) {
+    if (elapsed > 0 && elapsed >= maxConnectionDuration) {
```
omoerbeek, line 200: why the `elapsed > 0` test here?

rgacogne: My initial intent was to avoid treating a "clock has gone backward" event as a timeout, but it does not look necessary. I'll remove it.

PR #13527, opened 2023-11-23
build-and-test-all: Restrict Coveralls code coverage to the main repo

rgacogne:
### Short description

We are not interested in the coverage of forks, and it is confusing since relative coverage for forks does not really make sense.

### Checklist

I have:
- [ ] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/.github/workflows/build-and-test-all.yml
+++ b/.github/workflows/build-and-test-all.yml
@@ -68,6 +68,7 @@ jobs:
       - run: inv ci-auth-install-remotebackend-test-deps
       - run: inv ci-auth-run-unit-tests
       - run: inv generate-coverage-info ./testrunner $GITHUB_WORKSPACE
+        if: ${{ env.COVERAGE == 'yes' }}
         working-directory: ./pdns-${{ env.BUILDER_VERSION }}/pdns
       - name: Coveralls Parallel auth unit
         if: ${{ env.COVERAGE == 'yes' }}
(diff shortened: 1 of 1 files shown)


.github/workflows/build-and-test-all.yml
```diff
@@ -70,6 +70,7 @@ jobs:
       - run: inv generate-coverage-info ./testrunner $GITHUB_WORKSPACE
```
romeroalx, line 70: For `recursor` and `dnsdist` jobs, this step is also skipped. Should we also add here `if: ${{ env.COVERAGE == 'yes' }}`?

.github/workflows/build-and-test-all.yml
```diff
@@ -278,6 +279,7 @@ jobs:
       - run: inv test-api auth -b ${{ matrix.backend }}
       - run: inv generate-coverage-info /opt/pdns-auth/sbin/pdns_server $GITHUB_WORKSPACE
```
romeroalx, line 280: For `recursor` and `dnsdist` jobs, this step is also skipped. Should we also add here `if: ${{ env.COVERAGE == 'yes' }}`?

.github/workflows/build-and-test-all.yml
```diff
@@ -398,6 +400,7 @@ jobs:
       - run: inv test-auth-backend -b ${{ matrix.backend }}
       - run: inv generate-coverage-info /opt/pdns-auth/sbin/pdns_server $GITHUB_WORKSPACE
```
romeroalx, line 401: For `recursor` and `dnsdist` jobs, this step is also skipped. Should we also add here `if: ${{ env.COVERAGE == 'yes' }}`?

romeroalx (approved):

PR #13530, opened 2023-11-23
GH Actions: build-and-test-all workflow on Debian Bookworm

romeroalx:
### Short description
Run jobs for the workflow `build-and-test-all` on Debian Bookworm containers. Upgrade Runner's OS to `ubuntu-22.04`.

### Checklist

I have:
- [ ] compiled this code
- [x] tested this code
- [ ] documented the code


rgacogne (approved):

rgacogne:
@omoerbeek @Habbie Do you want to have a look before we merge this PR? The change looks good to me, but I want to make sure it's fine by you to upgrade the base distribution of our CI.

omoerbeek:
I'm OK with merging this, though it would be nice to have build-and-test-all runs using openssl 1.1 as well. Maybe not for each PR, but once every day or so.

rgacogne:
Agreed, but I'm not sure how easy it would be to make the distribution used in the testing workflow configurable.

romeroalx:
Actually, it seems that we only require `libssl-dev`. In this [run](https://github.com/romeroalx/pdns/actions/runs/7007598053) I did remove the dependency with `libssl1.1` and the workflow succeded.

For `libssl-dev` we are not pinning a version but we are installing `1.1` in Debian 11 and v3 in Debian 12

rgacogne:
Right, `libssl-dev` contains the files needed for development (headers, static libraries and pkg-conf files) but depends on the package that contains the dynamic libraries, so installing it will get us the correct version indeed.

Habbie (approved):
LGTM! Occasional testing on openssl 1.1 would indeed be nice, but I do not immediately have a good idea for how to do that cleanly, without duplicating the workflow.

romeroalx:
I will add running the workflow `build-and-test-all` also with `openssl1.1` to my to-do list (@romeroalx)

I just rebased from master this PR and will merge it now.

Thank you for your review.

Issue #11563, opened 2022-04-20
dnsdist: Cache doesn't store records beyond a certain size


Vascko:
Hi Folks,

 - Program: dnsdist
 - Issue type: Bug report

### Short description
I'm using a dnsdist pool with a cache and it appears that for records beyond a certain size, the cache doesn't commit the result. And there is no error message nor have I found a limitation in the docs that would explain this limit/behavior.
In this case I'm working with SRV records of varying sizes. An example record that makes it to the cache successfully is 590 bytes long and a record that fails to make it into the cache is 35982 bytes long.
It does not matter if I let the client fail from truncated UDP to TCP by itself or if I force TCP from the get go - same result.

### Environment
 - Operating system: RHEL 7
 - Software version: 1.7.0
 - Software source: PowerDNS repository

### Steps to reproduce
Simple pool configuration with a cache:
```
newPacketCache(100000, {numberOfShards=20, maxTTL=86400, minTTL=60, temporaryFailureTTL=60, staleTTL=60, maxNegativeTTL=300, keepStaleData=true, dontAge=false})
setStaleCacheEntriesTTL(3600)
```
Query SRV records with increasing size and observe cache dump and hit ratio.

### Expected behaviour
All records are committed to the cache

### Actual behaviour
Only records up to a certain size make it into the cache

### Other information
I don't have an adhoc way to simulate granular record sizes right now to provide more data but could set that up if there is not an obvious answer/solution to the problem I described.

rgacogne:
It's a design choice, the packet cache will not cache packets bigger than 4096 bytes: https://github.com/PowerDNS/pdns/blob/ca171679ced245e7001aa42a6f3cfc5e5f4f30be/pdns/dnsdist.hh#L1120
I agree it should be properly documented, though!

Vascko:
Thank you, @rgacogne! That makes a lot of sense.

Is there some context behind that design choice? Memory growth limitations? What are the chances that this might become a configurable variable in the future?

rgacogne:
> Is there some context behind that design choice? Memory growth limitations?

Yes! Since we limit the number of entries instead of the exact memory used by the cache (which would be costly to compute accurately), having a cap on the maximum size of entries makes it possible to predict the maximum amount of memory that will be used. It also has the nice benefit that after a while the probability of having to re-allocate when replacing an entry is very low, keeping both memory allocations and memory fragmentation at a low-level.

> What are the chances that this might become a configurable variable in the future?

[...]

PR #13537, opened 2023-11-24
dnsdist: Make the max size of entries in the packet cache configurable

rgacogne:
### Short description

It used to be set to 4096 bytes, which is also a hard limit for UDP responses anyway, because of the internal buffer size, but the limit can now be raised for responses received over TCP (including DoT and DoH).

Since the existing value was used to determine the initial size of TCP buffers, these are also updated:
- the default buffer size for incoming TCP connections is now 2 bytes: no need to waste a lot of memory for incoming connections until we got at least one query
- the default buffer size for outgoing connection is now 512 bytes: it should be enough for a fair percentage of responses

This PR is based on top of https://github.com/PowerDNS/pdns/pull/13536 and will need to be rebased, so the first commit can be ignored for now.

Closes https://github.com/PowerDNS/pdns/issues/11563.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code


chbruyand (approved):
LGTM!

PR #13538, opened 2023-11-27
Cleanup `ssql` and `ssqlite3`

fredmorcos:
### Short description
I was looking into a build issue with meson and came across those files which are easily cleaned up.

### Checklist
I have:
- [X] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/ssqlite3.cc
+++ b/pdns/ssqlite3.cc
@@ -290,24 +290,24 @@ SSQLite3::SSQLite3(const std::string& database, const std::string& journalmode, 
 {
   if (access(database.c_str(), F_OK) == -1) {
     if (!creat) {
-      throw sPerrorException("SQLite database '" + database + "' does not exist yet");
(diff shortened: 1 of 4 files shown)


pdns/ssqlite3.cc
```diff
@@ -38,104 +39,188 @@
...
+    };
+    return this;
+  }
+
+  SSqlStatement* execute() override
+  {
     prepareStatement();
     if (d_dolog) {
-      g_log<<Logger::Warning<< "Query "<<((long)(void*)this)<<": " << d_query << endl;
+      g_log << Logger::Warning << "Query " << this << ": " << d_query << endl;
```
rgacogne, line 161: It looks like the existing code really wants to print the address of the object, probably to correlate the log line, and I _think_ this will no longer be possible with this change? I would argue this calls for a unique ID or something when logging is enabled, but this might be out of scope of this PR.

rgacogne (approved):

PR #13543, opened 2023-11-29
rec: a single NSEC3 record covering everything is a special case

### LINKED ISSUE #13542: rec: NSEC3 aggressive cache does not work (for empty zones?)
- [x] This is not a support question, I have read [about opensource](https://www.powerdns.com/opensource.html) and will send support questions to the IRC channel, [Github Discussions](https://github.com/PowerDNS/pdns/discussions/) or the mailing list.
- [x] I have read and understood the ['out in the open' support policy](https://blog.powerdns.com/2016/01/18/open-source-support-out-in-the-open/)

 - Program: Recursor
 - Issue type: Bug report

### Short description
NSEC3 aggressive cache does not work for zones which have only apex and are otherwise empty.

### Environment
 - Operating system: Arch Linux, x86_64
 - Software version: powerdns-recursor 4.9.1-3

### Steps to reproduce
1. Recursor config:
```
dnssec=validate
dont-query=
local-address=127.0.0.1,::1
security-poll-suffix=
socket-dir=/tmp
```
2. Test auth config:
- [local.testiscorg.ch.zone.signed.txt](https://github.com/PowerDNS/pdns/files/13501376/local.testiscorg.ch.zone.signed.txt)
- BIND auth server: 
[auth.conf.txt](https://github.com/PowerDNS/pdns/files/13501379/auth.conf.txt)

3. Add test IP address: `sudo ip addr add 10.53.0.2 dev lo`
4. Run auth: `named -g -c auth.conf`
5. Observe traffic to the test auth: `tcpdump -n -i lo 'host 10.53.0.2 and port 53'`
6. Query for random names in the zone: `dig @127.0.0.1 $RANDOM$RANDOM.local.testiscorg.ch.`

### Expected behaviour
Queries stop going out as soon as NSEC3 chain is received. (I.e. first NSEC3 answer because there is just **one** NSEC3 RR in the whole zone.)

### Actual behaviour
Individual queries hit the auth.

### Other information
Insane config like this does not help:
```
aggressive-nsec-cache-size=1G
aggressive-cache-min-nsec3-hit-ratio=100000
```

omoerbeek:
Fixes #13542

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code


rgacogne (approved):
Change looks OK to me. It feels a bit weird to do this test in `commonPrefixIsLong`, I wonder if it wouldn't make more sense to have it in `AggressiveNSECCache::isSmallCoveringNSEC3` instead, before calling `commonPrefixIsLong`, but this makes no functional difference.

omoerbeek:
> Change looks OK to me. It feels a bit weird to do this test in `commonPrefixIsLong`, I wonder if it wouldn't make more sense to have it in `AggressiveNSECCache::isSmallCoveringNSEC3` instead, before calling `commonPrefixIsLong`, but this makes no functional difference.

Right, I had the same thought this morning. Will change.

pspacek:
I confirm it works in my tests case with an empty zone.

rgacogne (approved):

PR #13555, opened 2023-12-04
clang-tidy cleanups

neheb:
small for now.

### DIFF
--- a/modules/ldapbackend/ldapbackend.hh
+++ b/modules/ldapbackend/ldapbackend.hh
@@ -167,7 +167,7 @@ class LdapBackend : public DNSBackend
 
 public:
   LdapBackend(const string& suffix = "");
-  ~LdapBackend();
+  ~LdapBackend() override;
 
   // Native backend
   bool list(const DNSName& target, int domain_id, bool include_disabled = false) override;
--- a/pdns/saxfr.cc
+++ b/pdns/saxfr.cc
@@ -129,7 +129,7 @@ try
         throw PDNSException("tcp read failed");
 
       len=ntohs(len);
-      std::unique_ptr<char[]> creply(new char[len]);
+      auto creply = std::make_unique<char[]>(len);
       int n=0;
       int numread;
       while(n<len) {
--- a/modules/bindbackend/bindbackend2.hh
+++ b/modules/bindbackend/bindbackend2.hh
@@ -258,6 +258,9 @@ private:
 
     handle();
 
+    handle(const handle&) = delete;
+    handle& operator=(const handle&) = delete; // don't go copying this
+
     shared_ptr<const recordstorage_t> d_records;
     recordstorage_t::index<UnorderedNameTag>::type::const_iterator d_iter, d_end_iter;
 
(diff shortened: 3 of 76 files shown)


modules/ldapbackend/ldapbackend.hh
```diff
@@ -123,16 +123,11 @@ class LdapBackend : public DNSBackend
   {
     QType qtype;
     DNSName qname;
-    uint32_t ttl;
-    time_t lastmod;
+    uint32_t ttl{0};
+    time_t lastmod{0};
```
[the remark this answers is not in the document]
neheb: What type change?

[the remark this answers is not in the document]
omoerbeek: Oops read the diff wrong, sorry

pdns/dns.hh
```diff
@@ -72,18 +72,18 @@ public:
...
   QType qtype; //!< qtype of this record, ie A, CNAME, MX etc
-  uint16_t qclass; //!< class of this record
+  uint16_t qclass{1}; //!< class of this record
 
-  uint8_t scopeMask;
-  bool auth;
-  bool disabled;
+  uint8_t scopeMask{0};
+  bool auth{1};
+  bool disabled{0};
```
rgacogne, line 86:
```suggestion
  bool auth{true};
  bool disabled{false};
```

rgacogne (approved):
One nit that would not prevent this PR from being merged, especially since it was already like that before.
clang-tidy is reporting several things that were not introduced by this PR, but we need to decide whether we want to fix them now or open an issue to fix them later.

neheb:
I looked into the clang-tidy errors. Those are a lot of errors to fix...

pdns/dns.hh
neheb: Fixed.

rgacogne:
> I looked into the clang-tidy errors. Those are a lot of errors to fix...

Sorry, I should have been more clear about the fact that we do not expect _you_ to fix them :) If we decide to fix them now, do you mind if we push a few commits to this branch?

neheb:
It’s best to handle later then. I need to remove a few commits. Something broke.

neheb:
Fixed.

neheb:
ping @rgacogne

pdns/saxfr.cc
```diff
@@ -129,7 +129,7 @@ try
         throw PDNSException("tcp read failed");
 
       len=ntohs(len);
-      std::unique_ptr<char[]> creply(new char[len]);
+      auto creply = std::make_unique<char[]>(len);
```
rgacogne, line 132: Not your doing but switching to a `std::vector` or a `PacketBuffer` would be nice, perhaps in a follow-up PR.

rgacogne (approved):

rgacogne (approved):

neheb:
anything else needed to merge this?

rgacogne:
I was hoping for a second review, but I'll just do a second pass myself.

PR #13564, opened 2023-12-05
dnsdist: Add `PayloadSizeRule` and `TCResponseAction`

rgacogne:
### Short description

`PayloadSizeRule` can be used to match queries and responses based on their sizes, and `TCResponseAction` to truncate an already existing answer. Together they can be used to enforce a stricter truncation limit than what the backend is doing.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdistdist/dnsdist-rules.hh
+++ b/pdns/dnsdistdist/dnsdist-rules.hh
@@ -1350,3 +1350,66 @@ private:
   boost::optional<std::string> d_value;
   uint8_t d_type;
 };
+
+class PayloadSizeRule : public DNSRule
+{
+  enum class Comparisons : uint8_t { equal, greater, greaterOrEqual, smaller, smallerOrEqual };
+public:
+  PayloadSizeRule(const std::string& comparison, uint16_t size): d_size(size)
+  {
+    if (comparison == "equal") {
+      d_comparison = Comparisons::equal;
+    }
+    else if (comparison == "greater") {
+      d_comparison = Comparisons::greater;
+    }
+    else if (comparison == "greaterOrEqual") {
+      d_comparison = Comparisons::greaterOrEqual;
+    }
+    else if (comparison == "smaller") {
+      d_comparison = Comparisons::smaller;
+    }
+    else if (comparison == "smallerOrEqual") {
+      d_comparison = Comparisons::smallerOrEqual;
+    }
+    else {
+      throw std::runtime_error("Unsupported comparison '" + comparison + "'");
+    }
+  }
+
+  bool matches(const DNSQuestion* dq) const override
+  {
+    const auto size = dq->getData().size();
+
+    switch (d_comparison) {
+    case Comparisons::equal:
+      return size == d_size;
+    case Comparisons::greater:
+      return size > d_size;
+    case Comparisons::greaterOrEqual:
+      return size >= d_size;
+    case Comparisons::smaller:
+      return size < d_size;
+    case Comparisons::smallerOrEqual:
+      return size <= d_size;
+    default:
+      return false;
+    }
+  }
+
+  string toString() const override
+  {
+    static const std::array<const std::string, 5> comparisonStr{
+      "equal to" ,
+      "greater than",
+      "equal to or greater than",
+      "smaller than",
+      "equal to or smaller than"
+    };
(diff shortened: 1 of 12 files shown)


pdns/dnsdistdist/dnsdist-rules.hh
```diff
@@ -1350,3 +1350,66 @@ private:
...
+public:
+  PayloadSizeRule(const std::string& comparison, uint16_t size): d_size(size)
+  {
+    if (comparison == "equal") {
+      d_comparison = Comparisons::equal;
+    }
+    else if (comparison == "greater") {
+      d_comparison = Comparisons::greater;
+    }
+    else if (comparison == "greaterOrEqual") {
```
omoerbeek, line 1366: Wondering if `notequal` would be nice to have. Though you could create that with `NotRule`.  Come to think of it, you could maybe reduce the number of primitives and let the user use the Boolean combination rules to make the missing ones.

rgacogne: Yeah, my reasoning was that `not equal` is really not something I expect to be useful, and it can be done via `NotRule` if really needed indeed. The other ones seemed more useful.

pdns/dnsdistdist/docs/reference/constants.rst
```diff
@@ -154,6 +154,9 @@ All named `QTypes <https://www.iana.org/assignments/dns-parameters/dns-parameter
 DNSResponseAction
 -----------------
 
+.. versionchanged:: 1.8.0
```
Habbie, line 157: 1.9.0?

pdns/dnsdistdist/docs/rules-actions.rst
```diff
@@ -726,6 +726,15 @@ These ``DNSRule``\ s be one of the following items:
 
   :param int code: The opcode to match
 
+.. function:: PayloadSizeRule(comparison, size)
+
+  .. versionadded:: 1.9.0
+
+  Matches queries or responses whose DNS payload size is equal, greater, greater or equal, smaller or smaller or equal to the specified size.
```
Habbie, line 733: the list inside this sentence is a bit awkward

```suggestion
  Matches queries or responses whose DNS payload size fits the given comparison.
```

perhaps? the actual operator list is just below

pdns/dnsdistdist/docs/rules-actions.rst
```diff
@@ -726,6 +726,15 @@ These ``DNSRule``\ s be one of the following items:
...
+
+  :param str comparison: The comparison operator to use. Supported values are ``equal``, ``greater``, ``greaterOrEqual``, ``smaller`` and ``smallerOrEqual``.
+  :param int size: The size to compare to.w
```
Habbie, line 736:
```suggestion
  :param int size: The size to compare to.
```

pdns/dnsdistdist/docs/rules-actions.rst
```diff
@@ -1887,6 +1896,12 @@ The following actions exist.
   Before 1.7.0 this action was performed even when the query had been received over TCP, which required the use of :func:`TCPRule` to
   prevent the TC bit from being set over TCP transports.
 
+.. function:: TCResponseAction()
+
+  .. versionadded:: 1.9.0
+
+  Truncate an existing answer, to force the client to TCP.
```
Habbie, line 1903:
```suggestion
  Truncate an existing answer, to force the client to TCP.
  Only applied to UDP answers.
```

Habbie (commented):
code and tests look good, minor doc nits

pdns/dnsdistdist/docs/rules-actions.rst
omoerbeek: Does it actually truncate or does it set the TC bit _without_ modifying the data pat of the packet?

rgacogne: It does actually truncate.

rgacogne:
I pushed a commit addressing Peter's findings. Many thanks!

Habbie (approved):

pdns/dnsdistdist/docs/rules-actions.rst
rgacogne: Alright, I'll add it to the documentation.

PR #13584, opened 2023-12-07
dnsdist: Fix a small race in the NetworkListener

rgacogne:
### Short description

The main thread needs to be able to access the data even if the NetworkListener object has been destroyed first, which usually only happens when DNSdist is exiting, but could also happen earlier if the Lua handle is garbage collected (see #13583).

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


chbruyand (approved):
lgtm!

PR #13593, opened 2023-12-11
dnsdist: Add regression tests for DNS header set/get via Lua FFI

rgacogne:
### Checklist

I have:
- [ ] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/regression-tests.dnsdist/test_LuaFFI.py
+++ b/regression-tests.dnsdist/test_LuaFFI.py
@@ -322,24 +322,33 @@ class TestLuaFFIHeader(DNSDistTest):
     local bit = require("bit")
     local ffi = require("ffi")
 
-    function setAAResponseAction(dr)
+    -- check that the AA bit is clear, set the rcode to REFUSED otherwise
+    function checkAAResponseAction(dr)
       local header_void = ffi.C.dnsdist_ffi_dnsquestion_get_header(dr)
       local header = ffi.cast("unsigned char *", header_void)
       -- get AA
       local aa = bit.band(header[2], bit.lshift(1, 2)) ~= 0
       if aa then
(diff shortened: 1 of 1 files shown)


omoerbeek (commented):
I find it hard to follow the logic of these tests.

The `testLuaFFISetAAHeader` claims to test the `AA=0` case, but REFUSED is returned for the `aa` case in `setAAResponseAction`, if I'm not mistaken.

Also but maybe that's related to me not knowing the API very well): `DNSResponseAction.HeaderModify` is returned while the header is not modified and DNSResponseAction.None` is returned when the header *is* modified.

rgacogne:
> I find it hard to follow the logic of these tests.
> 
> The `testLuaFFISetAAHeader` claims to test the `AA=0` case, but REFUSED is returned for the `aa` case in `setAAResponseAction`, if I'm not mistaken.

Both tests use the same `setAAResponseAction` Lua function which checks whether the AA bit is clear in the initial response, and:
- set it if it was clear, as tested by `testLuaFFISetAAHeader`
- change the rcode to `REFUSED` if it was set, at tested by `testLuaFFIgetAAHeader`

> Also but maybe that's related to me not knowing the API very well): `DNSResponseAction.HeaderModify` is returned while the header is not modified and `DNSResponseAction.None` is returned when the header _is_ modified.

You are right that the two return values are not very consistent. It does not really matter for these cases because the modification has already been applied in both cases and there is no other rules, so `DNSResponseAction.None` would work, but the intent of returning `DNSResponseAction.HeaderModify`  is to prevent subsequent rules from being executed once we have decided to return a `REFUSED` response.

I'll refactor the tests and add some comments.

rgacogne:
I updated the tests, splitting the function in two and adding comments.

omoerbeek (approved):

PR #13594, opened 2023-12-11
dnsdist: Send a HTTP 400 response to HTTP/1.1 clients

rgacogne:
### Short description

Explaining that DNSdist with nghttp2 only supports DNS over HTTP2, as a best effort.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdistdist/dnsdist-nghttp2-in.cc
+++ b/pdns/dnsdistdist/dnsdist-nghttp2-in.cc
@@ -272,7 +272,7 @@ bool IncomingHTTP2Connection::checkALPN()
     return true;
   }
 
-  const std::string data("HTTP/1.1 400 Bad Request\r\nConnection: Close\r\n\r\n<html><body>This server implements RFC 8484 - DNS Queries over HTTP, and requires HTTP/2 in accordance to section 5.2 of the RFC.</body></html>\r\n");
+  const std::string data("HTTP/1.1 400 Bad Request\r\nConnection: Close\r\n\r\n<html><body>This server implements RFC 8484 - DNS Queries over HTTP, and requires HTTP/2 in accordance with section 5.2 of the RFC.</body></html>\r\n");
   d_out.insert(d_out.end(), data.begin(), data.end());
   writeToSocket(false);
 
--- a/pdns/dnsdistdist/dnsdist-nghttp2-in.cc
+++ b/pdns/dnsdistdist/dnsdist-nghttp2-in.cc
@@ -271,6 +271,11 @@ bool IncomingHTTP2Connection::checkALPN()
   if (protocols.size() == h2ALPN.size() && memcmp(protocols.data(), h2ALPN.data(), h2ALPN.size()) == 0) {
     return true;
   }
+
+  const std::string data("HTTP/1.1 400 Bad Request\r\nConnection: Close\r\n\r\n<html><body>This server implements RFC 8484 - DNS Queries over HTTP, and requires HTTP/2 in accordance to section 5.2 of the RFC.</body></html>\r\n");
+  d_out.insert(d_out.end(), data.begin(), data.end());
+  writeToSocket(false);
+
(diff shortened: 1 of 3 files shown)


regression-tests.dnsdist/test_DOH.py
```diff
@@ -377,6 +377,31 @@ def testDOHQueryInvalidALPN(self):
         except:
             pass
 
+    def testDOHHTTP1(self):
+        """
+        DOH: HTTP/1.1
+        """
+        if self._dohLibrary == 'h2o':
```
Habbie, line 384: I guess `@unittest.skipIf` won't work here because `_dohLibrary` is an object property?

regression-tests.dnsdist/test_DOH.py
```diff
@@ -377,6 +377,31 @@ def testDOHQueryInvalidALPN(self):
...
+                                         "Accept: application/dns-message"])
+        conn.setopt(pycurl.URL, url)
+        conn.setopt(pycurl.RESOLVE, ["%s:%d:127.0.0.1" % (self._serverName, self._dohServerPort)])
+        conn.setopt(pycurl.SSL_VERIFYPEER, 1)
+        conn.setopt(pycurl.SSL_VERIFYHOST, 2)
+        conn.setopt(pycurl.CAINFO, self._caCert)
+        data = conn.perform_rb()
+        rcode = conn.getinfo(pycurl.RESPONSE_CODE)
+        self.assertEqual(rcode, 400)
+        self.assertEqual(data, b'<html><body>This server implements RFC 8484 - DNS Queries over HTTP, and requires HTTP/2 in accordance to section 5.2 of the RFC.</body></html>\r\n')
```
Habbie, line 403:
```suggestion
        self.assertEqual(data, b'<html><body>This server implements RFC 8484 - DNS Queries over HTTP, and requires HTTP/2 in accordance with section 5.2 of the RFC.</body></html>\r\n')
```

pdns/dnsdistdist/dnsdist-nghttp2-in.cc
```diff
@@ -271,6 +271,11 @@ bool IncomingHTTP2Connection::checkALPN()
   if (protocols.size() == h2ALPN.size() && memcmp(protocols.data(), h2ALPN.data(), h2ALPN.size()) == 0) {
     return true;
   }
+
+  const std::string data("HTTP/1.1 400 Bad Request\r\nConnection: Close\r\n\r\n<html><body>This server implements RFC 8484 - DNS Queries over HTTP, and requires HTTP/2 in accordance to section 5.2 of the RFC.</body></html>\r\n");
```
Habbie, line 275:
```suggestion
  const std::string data("HTTP/1.1 400 Bad Request\r\nConnection: Close\r\n\r\n<html><body>This server implements RFC 8484 - DNS Queries over HTTP, and requires HTTP/2 in accordance with section 5.2 of the RFC.</body></html>\r\n");
```

Habbie (commented):
LGTM, one language nit, one small question

pdns/tcpiohandler.cc
```diff
@@ -1816,8 +1816,9 @@ bool setupDoHProtocolNegotiation(std::shared_ptr<TLSCtx>& ctx)
     return false;
   }
   /* we want to set the ALPN to doh */
-  const std::vector<std::vector<uint8_t>> dohAlpns = {{'h', '2'}};
+  const std::vector<std::vector<uint8_t>> dohAlpns{{'h', '2'},{'h', 't', 't', 'p', '/', '1', '.', '1'}};
```
omoerbeek, line 1819: Why advertise 1.1 if we're not supporting it? Or is that *only* to be able to send the new error message back? A comment explaining that would help.

regression-tests.dnsdist/test_DOH.py
rgacogne: I _think_ this is a reason, but I remember not being able to make it work.

rgacogne:
I pushed a commit fixing the issue spotted by Peter and another adding a comment about why HTTP/1.1 is added to the ALPN values. Thanks!

Habbie (approved):

PR #13596, opened 2023-12-11
configure.ac fixup: do not require bash

eli-schwartz:
- configure: remove broken bashism

  In a configure check that was carefully written for pre-unix-wars versions of the bourne shell, some code which was only valid using GNU bash was included.
  
  The `==` operator is a bash-specific alias for `=`. It behaves exactly the same, except more confusing. It contains no added functionality, other than making an otherwise /bin/sh compatible script only work when /bin/sh is a symlink to /bin/bash.

- configure: remove useless multi-step check for cdb

  cdb doesn't provide a pkg-config file. Debian patches one in, but theirs is broken and provides no functionality which isn't already covered by checking if -lcdb works.
  
  The compile test is already needed in the general case, provides more accurate results even on Debian, and avoids running commands that cannot work on non-Debian systems.

### DIFF
--- a/m4/pdns_d_fortify_source.m4
+++ b/m4/pdns_d_fortify_source.m4
@@ -28,13 +28,13 @@ AC_DEFUN([AC_CC_D_FORTIFY_SOURCE],[
   AS_IF([test "x$enable_fortify_source" != "xno"], [
 
     dnl Auto means the highest version we support, which is currently 3
-    AS_IF([test "x$enable_fortify_source" == "xauto"],
+    AS_IF([test "$enable_fortify_source" = "auto"],
       [enable_fortify_source=3],
       []
     )
 
     dnl If 3 is not supported, we try to fallback to 2
-    AS_IF([test "x$enable_fortify_source" == "x3"], [
+    AS_IF([test "$enable_fortify_source" = "3"], [
       gl_COMPILER_OPTION_IF([-D_FORTIFY_SOURCE=3], [
         CFLAGS="-U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=3 $CFLAGS"
         CXXFLAGS="-U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=3 $CXXFLAGS"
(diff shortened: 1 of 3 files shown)


zeha:
>configure: remove useless multi-step check for cdb

ah yes, thanks for making my life harder

eli-schwartz:
Can you clarify what's harder about it?

zeha:
As already explained in #debian-devel, dropping a file into PKG_CONFIG_PATH is a lot easier than manually fiddling with all CFLAGS/CPPFLAGS/CXXFLAGS/LDFLAGS vars, especially on non-Linux.

eli-schwartz:
> As already explained in #debian-devel, dropping a file into PKG_CONFIG_PATH is a lot easier than manually fiddling with all CFLAGS/CPPFLAGS/CXXFLAGS/LDFLAGS vars, especially on non-Linux.

As I say in my commit message: "Debian patches one in, but theirs is broken and provides no functionality which isn't already covered by checking if -lcdb works."

I'm one of the greatest proponents of pkg-config around. It makes me very sad to say that this time, using a pkg-config file is slightly (not a lot) worse than not using one and just using AC_CHECK_LIB.

zeha:
> Debian patches one in, but theirs is broken

This misses out the fact that:
1) libcdb.pc in Debian stable works.
2) Fedora also adds the .pc file.
3) tinycdb upstream (=mjt) might add the .pc file, given he already has it in the debian package and I've asked him today about it.

eli-schwartz:
> libcdb.pc in Debian stable works.

It is missing all the important parts that make pkg-config files actually work. AC_CHECK_LIBS provides *more* information than the pkg-config file does.

> Fedora also adds the .pc file.

Aha, theirs appears to be well formed and actually functional.

> tinycdb upstream (=mjt) might add the .pc file, given he already has it in the debian package and I've asked him today about it.

Okay. If he adds one upstream then relying on it downstream seems like a good idea.

Habbie:
if you do just the bashism changes, we can merge this, thanks!

zeha:
> Okay. If he adds one upstream then relying on it downstream seems like a good idea.

This was done, IIRC.

dwfreed:
Fedora uses the one that was in 0.77, which includes -L/-I in Libs/Cflags fields.  The version in 0.81 upstream is the same as the one Debian had in 0.78, which is missing these flags.  So I doubt @eli-schwartz's complaint with it has been resolved by upstream.

eli-schwartz:
```
$ wget http://www.corpit.ru/mjt/tinycdb/tinycdb-0.81.tar.gz && bsdtar -xf tinycdb-0.81.tar.gz && cd tinycdb-0.81 
$ make install DESTDIR=$PWD/pkg
[...]
installing libcdb.pc to /tmp/tinycdb-0.81/pkg/usr/local/lib/pkgconfig/libcdb.pc

$ cat pkg/usr/local/lib/pkgconfig/libcdb.pc
libdir=/usr/local/lib
includedir=/usr/local/include
Name: libcdb
Description: tinycdb - Constant Data Base library
Version: 0.81
Libs: -L${libdir} -lcdb
Cflags: -I${includedir}
```

It looks fine to me. Based on the contents of http://deb.debian.org/debian/pool/main/t/tinycdb/tinycdb_0.81-1.debian.tar.xz there's still the old debian/libcdb.pc.in file which has different contents.

http://ftp.us.debian.org/debian/pool/main/t/tinycdb/libcdb-dev_0.81-1_amd64.deb -> data.tar.xz -> usr/lib/x86_64-linux-gnu/pkgconfig/libcdb.pc

contains the pkgconfig file installed by the Makefile, not the one in the debian/ directory. It should be fine.

rgacogne (approved):

PR #13221, opened 2023-09-07
dnsdist: Remove the fuzz_target_dnsdistcache binary

rgacogne:
### Short description

I somehow managed to add a 10 MB ELF binary to the repository in #13145 (thanks GitHub for sneakily displaying `Binary file not shown` during the code review, I feel it should be a bit more hard to ignore!), sorry!
I wonder if it's possible -and worth our time- to check this on PRs?

### Checklist

I have:
- [x] read the [CONTRIBUTING.md](https://github.com/PowerDNS/pdns/blob/master/CONTRIBUTING.md) document
- [ ] compiled this code
- [ ] tested this code
- [ ] included documentation (including possible behaviour changes)
- [ ] documented the code
- [ ] added or modified regression test(s)
- [ ] added or modified unit test(s)

PR #13608, opened 2023-12-12
GH Actions: force CI failure if there are binaries present in a Pull Request

romeroalx:
### Short description

Force CI failure if there are binaries present in a Pull Request.

The original request comes from this [pull request](https://github.com/PowerDNS/pdns/pull/13221)

This is an example of a failure in the CodeQL workflow when a binary file is added to a PR: [here](https://github.com/romeroalx/pdns/actions/runs/7182275016/job/19558412138)

### Checklist

I have:
- [ ] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -250,7 +250,7 @@ jobs:
 
   check-clang-tidy:
     needs: analyze
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     name: Check whether clang-tidy succeeded
     steps:
       - run: |
@@ -266,3 +266,14 @@ jobs:
             echo "::error::Rec clang-tidy failed"
             exit 1
           fi
+
+  check-for-binaries:
+    runs-on: ubuntu-22.04
+    name: Force failure in case there are binaries present in a pull request
+    if: ${{ github.event_name == 'pull_request' }}
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v4
+      with:
+        fetch-depth: 2
+    - run: if [[ "$(file -i --dereference $(git diff --name-only HEAD^..HEAD) | grep binary | grep -v 'image/' | grep -v 'inode/x-empty')" != "" ]]; then exit 1; fi


.github/workflows/codeql-analysis.yml
```diff
@@ -266,3 +266,14 @@ jobs:
             echo "::error::Rec clang-tidy failed"
             exit 1
           fi
+
+  check-for-binaries:
+    runs-on: ubuntu-22.04
+    name: Force failure in case there are binaries present in a pull request
+    if: ${{ github.event_name == 'pull_request' }}
```
rgacogne, line 273: Would it make sense to enable this test for push events as well?

rgacogne (approved):
Looks good, nice work!

.github/workflows/codeql-analysis.yml
romeroalx: I thought about it but I did not find a clean way to achieve it.

If my understanding is correct, in case of a push event, this command `git diff HEAD^..HEAD` will only compare the latest commit with the previous one (for a PR, GitHub creates a special commit including all changes). In case of pushing more than one commit at the same time, that diff no longer works as expected. This is already happening during the `analyze` step.

Habbie: I guess that would look like a very different check - scan the whole tree for binaries, check them against an allowlist. Does not seem worth the work I think

rgacogne: So when we push to the branch used in a pull request we don't get a new "special" merge commit? That's awful, it explains some things I have seen in the analyze step indeed, but it also means we cannot easily get the changes introduced by a PR which seems like a very bad design choice to me. I'll look into it.

Habbie: On the PR side, you do get a new special merge commit, covering the entire PR.

rgacogne: > On the PR side, you do get a new special merge commit, covering the entire PR.

Even on a push to the branch after the PR has been opened?

Habbie: yes

rgacogne: My tests seem to agree, nice! So then I would expect `git diff HEAD^..HEAD` to always provide the difference between the current HEAD of target branch (master, mostly) and the HEAD of the PR branch, which would include all commits on the PR branch. Still I'm quite sure I have seen what Alexis described earlier, where on a push to a branch for which a PR already exists, `git diff HEAD^..HEAD` only return the changes of the last commit. Although, I'm now wondering if I somehow got confused and looked at the CI run on my repository and not the one for the PR on PowerDNS/pdns.

Habbie: Looking at your repo would definitely explain that!

Habbie: It wouldn't hurt, but it would not be too useful - if I push a binary to a branch, it gets detected; then I push another thing (without adding another binary), and the branch status for this check would go green. Meanwhile the PR check would stay red as it checks the entire PR.

rgacogne: Ah, I think I might be confused about something else then! With the `github.event_name == 'pull_request'` condition, is this check executed if I push a commit to a branch corresponding to an existing PR?

Habbie: yes, it will!

Habbie: (event really is the wrong word there)

rgacogne: > For example, if no activity types are specified, the workflow runs when a pull request is opened or reopened or when the head branch of the pull request is updated.

Riiiiiiiiiiiiiiight. Thank you, now I wonder how I managed to misunderstand that for so long :facepalm:

.github/workflows/codeql-analysis.yml
```diff
@@ -266,3 +266,14 @@ jobs:
...
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v4
+      with:
+        fetch-depth: 2
```
jsoref, line 278: This scares me...

Habbie: why?

jsoref: fetch-depth 2 in theory only gives you two commits of history, if i make a PR series with 10 commits, then I don't expect you to be able to get good answers from git when you ask it about that.

romeroalx: It was a little bit confusing for me too. 

For a pull request, the checkout action creates a "merge commit" which aggregates all commits in the PR, where `HEAD` will be pointing to (unless we specify otherwise, see [here](https://github.com/actions/checkout#Checkout-pull-request-HEAD-commit-instead-of-merge-commit) and [here](https://github.com/actions/checkout/issues/426)). Then we compare this commit with its parent, `HEAD^`.

It is the same syntax we are already using in the `analyze` job of the `codeql` workflow, [line 64](https://github.com/PowerDNS/pdns/blob/master/.github/workflows/codeql-analysis.yml#L64) to later do a similar `git diff` command for `clang-tidy`, in [line 128](https://github.com/PowerDNS/pdns/blob/master/.github/workflows/codeql-analysis.yml#L128).

Habbie (approved):

PR #13613, opened 2023-12-13
Remove the `extern`ed `StatBag` from `ws-auth`

fredmorcos:
### Short description
This removes the `extern StatBag` object from `ws-auth` and instead passes it around as a function/method argument. We should do this, in general, as part of a wider cleanup.

This also cleans up and formats some files automatically using `clang-format`.

Better reviewed on a per-commit basis.

### Checklist
I have:
- [X] read the [CONTRIBUTING.md](https://github.com/PowerDNS/pdns/blob/master/CONTRIBUTING.md) document
- [X] compiled this code
- [X] tested this code
- [ ] included documentation (including possible behaviour changes)
- [ ] documented the code
- [ ] added or modified regression test(s)
- [ ] added or modified unit test(s)

PR #13629, opened 2023-12-14
auth: Fix a construction ordering issue between `StatBag` and `Logger`

rgacogne:
### Short description

By turning the global `StatBag` into a singleton.

Reported by Coverity as CID 1524949: Initialization or destruction ordering is unspecified.

This is only a draft because turning `StatBag` into a singleton might not be the way we want to tackle this. It's the only way I found that did not require an awful lot of refactoring in various places, but hopefully someone will get a better idea :)

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/modules/remotebackend/test-remotebackend-http.cc
+++ b/modules/remotebackend/test-remotebackend-http.cc
@@ -32,12 +32,10 @@
 #include "pdns/logger.hh"
 #include "pdns/arguments.hh"
 #include "pdns/json.hh"
-#include "pdns/statbag.hh"
 #include "pdns/auth-packetcache.hh"
 #include "pdns/auth-querycache.hh"
 #include "pdns/auth-zonecache.hh"
 
-StatBag S;
 AuthPacketCache PC;
 AuthQueryCache QC;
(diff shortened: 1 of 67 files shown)


Habbie:
> hopefully someone will get a better idea

like #13613? Or do we need both these PRs? If we're explicitly passing it always (in 13613) maybe that gives us enough control over when to construct it.

rgacogne:
> > hopefully someone will get a better idea
> 
> like #13613? Or do we need both these PRs? If we're explicitly passing it always (in 13613) maybe that gives us enough control over when to construct it.

#13613 does not fix this issue, no. If we managed to always pass the `StatBag` along we could make it non-global which would solve the issue, but I really don't see that happening given how widespread its use is. I'd very happy to be proven wrong.

Habbie:
> we managed to always pass the `StatBag` along

right, 13613 only does one file. Thank you for explaining!

rgacogne:
Closing this PR since it is now conflicted and we seem to be aiming for a different approach.

PR #13630, opened 2023-12-14
dnsdist: Fix Coverity warnings

rgacogne:
### Short description

One real issue that impact the TCP connection termination metrics, and some insignificant "performance" warnings in places where we do not care about performance.

The TCP metric issue was introduced in 448d66d4bb6fc1a6abff2329bd01cb65cf45005a so we will need to backport the fix to 1.8.x and perhaps 1.7.x.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


chbruyand (approved):
LGTM!

PR #13638, opened 2023-12-15
dnsdist: Handle congested DoQ streams

### LINKED ISSUE #13631: dnsdist closes congested DNS-over-QUIC streams with INTERNAL_ERROR
- [x] This is not a support question, I have read [about opensource](https://www.powerdns.com/opensource.html) and will send support questions to the IRC channel, [Github Discussions](https://github.com/PowerDNS/pdns/discussions/) or the mailing list.
- [x] I have read and understood the ['out in the open' support policy](https://blog.powerdns.com/2016/01/18/open-source-support-out-in-the-open/)

 - Program: dnsdist 
 - Issue type: Bug report

### Short description

While testing [my implementation of DNS-over-QUIC](https://gitlab.nic.cz/knot/shotgun/-/merge_requests/75) in [DNS Shotgun](https://dns-shotgun.readthedocs.io) against `dnsdist`, I have found that it closes a large number of streams with an `INTERNAL_ERROR` app code.

### Environment

 - Operating system: OpenSUSE Tumbleweed
 - Software version: `master` - 5012710e403dbf6c93451b6fc159bc83b0fbaf79

### Steps to reproduce

1. Send high-volume DNS-over-QUIC traffic to dnsdist, e.g. using DNS Shotgun.

### Expected behaviour
Pretty much 100% of the queries should be answered in a localhost setting.

### Actual behaviour
A large number of the queries are lost because dnsdist closes the streams with an `INTERNAL_ERROR` app code.

### Other information

After inspecting the code, the culprit seems to be that the `quiche_conn_stream_send` function returns the value `-1` (`QUICHE_ERR_DONE`), which is because the stream is congested, [as per the Quiche documentation](https://github.com/cloudflare/quiche/blob/8e64c0670835d2005dfb10c743382c9fd52f56d6/quiche/src/lib.rs#L4550). This causes `dnsdist` to lose a lot of legitimate queries in a high-volume traffic scenario.

When `quiche_conn_stream_send` returns `QUICHE_ERR_DONE`, `dnsdist` should defer sending the reply to a later time at which it gets notified that the stream is writable again.

rgacogne:
### Short description

If the stream has no capacity left, Quiche will refuse to queue more data and return `QUICHE_ERR_DONE`. We then have to wait until the stream becomes writable again to retry sending our response.

Hopefully closes #13631. I managed to hit the reported issue and this PR fixes it for me, but I'm not 100% sure I am sending the same kind of traffic, so it might not fix all issues.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdistdist/doh3.cc
+++ b/pdns/dnsdistdist/doh3.cc
@@ -67,6 +67,7 @@ public:
   QuicheConnection d_conn;
   QuicheHTTP3Connection d_http3{nullptr, quiche_h3_conn_free};
   std::unordered_map<uint64_t, PacketBuffer> d_streamBuffers;
+  std::unordered_map<uint64_t, PacketBuffer> d_streamOutBuffers;
 };
 
 static void sendBackDOH3Unit(DOH3UnitUniquePtr&& unit, const char* description);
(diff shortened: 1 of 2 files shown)


chbruyand (approved):
LGTM!
Thanks for the fix

PR #13650, opened 2023-12-18
dnsdist: Fall back to libcrypto for authenticated encryption

rgacogne:
### Short description

We used to fall back to plain-text for console communications when libsodium was not available, which was not great. Now that we are also using the authenticated encryption module to secure our QUIC tokens, let's fall back to OpenSSL's Chacha20 Poly 1305 implementation instead.
Note that, unfortunately, both implementations are not compatible so the console communication format will be different depending on whether libsodium is available. I believe this is still better than plain-text :)

I'm renaming `sodcrypto.cc` and `sodcrypto.hh` in the process and GitHub is unfortunately not handling that very well, so it might be easier to review the diff in a different tool, sorry about that.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/.github/actions/spell-check/allow.txt
+++ b/.github/actions/spell-check/allow.txt
@@ -3335,7 +3335,6 @@ sockname
 sockowner
 socktype
 sodbc
-sodcrypto
 sodiumsigners
 sokolov
 somedata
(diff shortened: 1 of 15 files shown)


rgacogne:
By the way, Neil has been doing something similar in Weakforced for a while: https://github.com/PowerDNS/weakforced/pull/399

PR #13652, opened 2023-12-19
rec: Mtasker cleanup and move to recursordist

omoerbeek:
### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code


rgacogne (approved):
The PR looks good, with one thing that bothers me a bit: the first commit will not compile if we ever want to do a bisect, right?

omoerbeek:
> The PR looks good, with one thing that bothers me a bit: the first commit will not compile if we ever want to do a bisect, right?

That is correct, but see #12318

rgacogne:
> > The PR looks good, with one thing that bothers me a bit: the first commit will not compile if we ever want to do a bisect, right?
> 
> That is correct, but see #12318

Ack, this is not ideal but I don't have a better idea so let's do it.

PR #13654, opened 2023-12-19
rec: allow out-of-tree builds

zeha:
### Short description
Allow building recursor with autoconf in a separate build dir.

The codegen for the rust parts is not 100% clean: it writes two files into the srcdir. But it does build.

### Checklist

I have:
- [x] compiled this code
- [ ] tested this code
- [ ] documented the code


omoerbeek:
I wanted to test this, but it is not clear for me how to define the build dir. `./configure --help` does not give me a clue.

omoerbeek (approved):
This looks good and works for me. The writing into the src tree by the generate python script is not easily avoided, as then cargo gets confused. We also need to keep an eye on the openwrt package build, as it is slightly different.

Habbie (approved):

PR #13664, opened 2023-12-22
dnsdist: increase receive and send buffers to max

chbruyand:
### Short description
This tries to increase read and send buffers to the max value as read in `/proc/sys/net/core/{r,w}mem_max` for UDP sockets.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


pdns/dnsdist.cc
```diff
@@ -2390,6 +2390,32 @@ static void setupLocalSocket(ClientState& clientState, const ComboAddress& addr,
...
+#endif /* SO_RCVBUF */
+#ifdef SO_SNDBUF
+    {
+      std::ifstream ifs("/proc/sys/net/core/wmem_max");
+      if (ifs) {
+        std::string line;
+        if (getline(ifs, line)) {
+          auto maxWmem = std::stoi(line);
+          SSetsockopt(socket, SOL_SOCKET, SO_SNDBUF, maxWmem);
+        }
```
rgacogne, line 2414: Perhaps we could turn this into a function, passing the name of the /proc file and the socket option as parameters?

pdns/dnsdist.cc
```diff
@@ -2390,6 +2390,32 @@ static void setupLocalSocket(ClientState& clientState, const ComboAddress& addr,
...
+      }
+    }
+#endif /* SO_RCVBUF */
+#ifdef SO_SNDBUF
+    {
+      std::ifstream ifs("/proc/sys/net/core/wmem_max");
+      if (ifs) {
+        std::string line;
+        if (getline(ifs, line)) {
+          auto maxWmem = std::stoi(line);
```
rgacogne, line 2412: `pdns::checked_stoi` might be a better fit than `std::stoi`, and we might to catch any exception, logging at warning level.

pdns/dnsdist.cc
```diff
@@ -2390,6 +2390,32 @@ static void setupLocalSocket(ClientState& clientState, const ComboAddress& addr,
         warnlog(e.what());
       }
     }
+#ifdef __linux__
```
rgacogne, line 2393: I don't think we want to raise automatically if either `g_socketUDPSendBuffer` or `g_socketUDPRecvBuffer` is set.

pdns/dnsdist.cc
rgacogne, line 2414: Perhaps log at info level after raising the buffer size, advertising the new value?

rgacogne (changes requested):
Nice! A few comments on the implementation but the general idea is great :+1:

phonedph1:
I don't disagree with any of this but I will point out this actually makes it twice the value of max as opposed to if you got them via rmem_default/wmem_default... this probably doesn't make a big difference for most people. 

SO_RCVBUF
              Sets or gets the maximum socket receive buffer in bytes.
              The kernel doubles this value (to allow space for
              bookkeeping overhead) when it is set using [setsockopt(2)](https://man7.org/linux/man-pages/man2/setsockopt.2.html),
              and this doubled value is returned by [getsockopt(2)](https://man7.org/linux/man-pages/man2/getsockopt.2.html).

rgacogne (approved):

PR #13666, opened 2023-12-22
dnsdist: Optimize the DoQ packet handling path

rgacogne:
### Short description

By reading as many UDP datagrams as possible from the socket when it becomes readable, processing as many DoQ events as possible in one go, and preventing a few allocations and copies.

I do get very nice results with this pull request and increasing the rmem and wmem buffers (which quic-go appears to do by default):

latency

And even better on a longer-run, as the cost of establishing the QUIC connections is amortized (60 seconds run vs 20 seconds for the previous one above):

latency

Might help with https://github.com/PowerDNS/pdns/issues/13631 even though I believe https://github.com/PowerDNS/pdns/pull/13664 might very well be as big a win.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


chbruyand (approved):
LGTM!

PR #13670, opened 2023-12-28
dnsdist: doq,doh3 make sure we enforce any ACL

chbruyand:
### Short description
This makes sure we enforce ACL in the DOQ and DOH3 paths.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdistdist/doh3.cc
+++ b/pdns/dnsdistdist/doh3.cc
@@ -471,6 +471,16 @@ static void processDOH3Query(DOH3UnitUniquePtr&& doh3Unit)
     auto& holders = dsc->holders;
     ClientState& clientState = *dsc->clientState;
 
+    if (!holders.acl->match(remote)) {
+      vinfolog("Query from %s (DoH3) dropped because of ACL", remote.toStringWithPort());
+      ++dnsdist::metrics::g_stats.aclDrops;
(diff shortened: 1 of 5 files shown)


rgacogne (commented):
Looks good! I wonder if it would make sense to drop the DoQ connection earlier in the process, before spending CPU time negotiating the key materials?

chbruyand:
> Looks good! I wonder if it would make sense to drop the DoQ connection earlier in the process, before spending CPU time negotiating the key materials?

That would be nice, but I think that if we do so, we cannot properly terminate the connection, but only discard received data ?

rgacogne:
> > Looks good! I wonder if it would make sense to drop the DoQ connection earlier in the process, before spending CPU time negotiating the key materials?
> 
> That would be nice, but I think that if we do so, we cannot properly terminate the connection, but only discard received data ? Anyway we can indeed do that earlier and avoid some unnecessary processing.

You are right! Perhaps we could make it optional? I think proper termination is great most of the time, but when you are under attack you might want to just drop DoQ packets not allowed by the ACL.

rgacogne:
I'm going to merge this because not enforcing the ACL is a clear issue, and we can look for improvements later.

PR #13688, opened 2024-01-09
build-packages: Fix the handling of provenance artifacts

rgacogne:
### Short description

The current version of the SLSA framework uses upload-artifact v3 which is not compatible with download-artifact v4 (don't ask me).

### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [x] documented the code

### DIFF
--- a/.github/workflows/build-packages.yml
+++ b/.github/workflows/build-packages.yml
@@ -187,12 +187,12 @@ jobs:
     steps:
       - name: Download source tarball provenance for ${{ inputs.product }} (${{ inputs.ref }})
         id: download-src-provenance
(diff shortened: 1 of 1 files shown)


rgacogne:
> Maybe rollback also the upload-artifact action to v3 in line 93 of the same file?

Unsure, on one side it would make sense to use the same version everywhere in that file, but on the other side we migrated to v4 everywhere so it might make sense to only rollback to v3 what we really have to.

rgacogne:
It turns out we should not need to backport this because the workflows calling this one explicitly call the version living on master!

PR #13701, opened 2024-01-10
rec: RPZ from primary refactor and allow notifies for RPZs

### LINKED ISSUE #12777: Allow clients to subscribe to receive NOTIFY messages for zone updates
- Program: Authoritative
 - Issue type: Feature request

### Short description
Provide an additional mechanism for an auth server to learn about clients which should receive NOTIFY messages when a zone's SOA is updated.

### Usecase
It would be useful to be able to stand up/shut down 'hidden' secondary auth servers (not listed in the NS RRset for a zone) without having to modify the configuration of the primary auth server, or even adjust metadata for the zones. It would also be useful to be able to do this with recursive resolvers which obtain RPZs from the primary auth server. Finally, it would also be useful for recursive resolvers which obtain catalog zones from the primary auth server (should that be implemented as discussed in #12463).

### Description
The idea is to allow DNS clients to send some sort of request to be added to the 'NOTIFY target' list for a specified zone. This subscription could expire, and thus need to be refreshed by the client, in order to avoid the auth server sending unwanted NOTIFY messages to an address where no client is present.

Some possible mechanisms for the 'subscribe' operation:

1. Clients could send a TSIG-signed UPDATE query to add a TXT record into the zone; the data portion of the record would include the address/port to receive the NOTIFY, and a timestamp for expiration of the subscription. UPDATE would be used to 'refresh' the subscription, or to delete the subscription.
2. The auth server could be configured with a list of TSIG key names which, when used by a client to sign an AXFR, would automatically subscribe that client for NOTIFY messages for the zone. The server could store the list of subscribers in the ALSO-NOTIFY metadata for the zone, or in a newly-created metadata item. This mechanism would not provide any easy way to specify an alternate address/port for the NOTIFY message, or to specify an expiration timestamp.

Additionally, if a client subscribed to NOTIFY messages for a `Producer` zone, they would also be subscribed to such messages for all zones in the catalog, and for zones added to the catalog.

omoerbeek:
### Short description

This refactors the RPZ load via IXFR thread code and allows processing of notifies, causing the RPZ code to immediately initiate an IXFR.

Fixes #12777 (for the RPZ part)

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [X] documented the code

### DIFF
--- a/pdns/recursordist/docs/lua-config/rpz.rst
+++ b/pdns/recursordist/docs/lua-config/rpz.rst
@@ -214,7 +214,8 @@ Base64 encoded TSIG secret
 refresh
 ^^^^^^^
 An integer describing the interval between checks for updates.
-By default, the RPZ zone's default is used
+By default, the RPZ zone's default is used.
+If allowed by :ref:`setting-allow-notify-for` and :ref:`setting-allow-notify-from`, a ``notify`` for an RPZ zone will initiate an freshness check.
 
 maxReceivedMBytes
 ^^^^^^^^^^^^^^^^^
--- a/pdns/recursordist/docs/lua-config/rpz.rst
+++ b/pdns/recursordist/docs/lua-config/rpz.rst
@@ -215,7 +215,7 @@ refresh
 ^^^^^^^
 An integer describing the interval between checks for updates.
 By default, the RPZ zone's default is used.
-If allowed by :ref:`setting-allow-notify-for` and :ref:`setting-allow-notify-from`, a ``notify`` for an RPZ zone will initiate an freshness check.
+If allowed by :ref:`setting-allow-notify-for` and :ref:`setting-allow-notify-from`, a ``notify`` for an RPZ zone will initiate a freshness check.
 
 maxReceivedMBytes
 ^^^^^^^^^^^^^^^^^
--- a/pdns/recursordist/rpzloader.cc
+++ b/pdns/recursordist/rpzloader.cc
@@ -386,47 +386,27 @@ static bool dumpZoneToDisk(Logr::log_t logger, const DNSName& zoneName, const st
   return true;
 }
 
-void RPZIXFRTracker(const std::vector<ComboAddress>& primaries, const boost::optional<DNSFilterEngine::Policy>& defpol, bool defpolOverrideLocal, uint32_t maxTTL, size_t zoneIdx, const TSIGTriplet& tt, size_t maxReceivedBytes, const ComboAddress& localAddress, const uint16_t xfrTimeout, const uint32_t refreshFromConf, std::shared_ptr<const SOARecordContent> sr, const std::string& dumpZoneFileName, uint64_t configGeneration)
+static void preloadRPZFIle(RPZTrackerParams& params, const DNSName& zoneName, std::shared_ptr<DNSFilterEngine::Zone>& oldZone, uint32_t& refresh, const string& polName, Logr::log_t logger)
 {
-  setThreadName("rec/rpzixfr");
-  bool isPreloaded = sr != nullptr;
-  auto logger = g_slog->withName("rpz");
-
-  /* we can _never_ modify this zone directly, we need to do a full copy then replace the existing zone */
-  std::shared_ptr<DNSFilterEngine::Zone> oldZone = g_luaconfs.getLocal()->dfe.getZone(zoneIdx);
-  if (!oldZone) {
-    SLOG(g_log << Logger::Error << "Unable to retrieve RPZ zone with index " << zoneIdx << " from the configuration, exiting" << endl,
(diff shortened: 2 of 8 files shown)


pdns/recursordist/docs/lua-config/rpz.rst
```diff
@@ -214,7 +214,8 @@ Base64 encoded TSIG secret
 refresh
 ^^^^^^^
 An integer describing the interval between checks for updates.
-By default, the RPZ zone's default is used
+By default, the RPZ zone's default is used.
+If allowed by :ref:`setting-allow-notify-for` and :ref:`setting-allow-notify-from`, a ``notify`` for an RPZ zone will initiate an freshness check.
```
rgacogne, line 218:
```suggestion
If allowed by :ref:`setting-allow-notify-for` and :ref:`setting-allow-notify-from`, a ``notify`` for an RPZ zone will initiate a freshness check.
```

pdns/recursordist/rpzloader.cc
```diff
@@ -445,176 +459,252 @@ void RPZIXFRTracker(const std::vector<ComboAddress>& primaries, const boost::opt
...
+  }
+  return true;
+}
+
+// As there can be multiple threads doing updates (due to config reloads), we use a multimap.
+// The value contains the actual thread id that owns the struct.
+
+static LockGuarded<std::multimap<DNSName, RPZWaiter&>> condVars;
+
+// Notify all threads trakcing the RPZ name
```
rgacogne, line 647:
```suggestion
// Notify all threads tracking the RPZ name
```

pdns/recursordist/rpzloader.cc
```diff
@@ -445,176 +459,252 @@ void RPZIXFRTracker(const std::vector<ComboAddress>& primaries, const boost::opt
...
+  while (RPZTrackerIteration(params, zoneName, oldZone, refresh, polName, skipRefreshDelay, configGeneration, waiter, logger)) {
+    // empty
+  }
+
+  // Zap our (and only our) RPZWaiter struct out of the multimap
+  auto lock = condVars.lock();
+  auto [start, end] = lock->equal_range(zoneName);
+  while (start != end) {
+    if (start->second.id == std::this_thread::get_id()) {
+      start = lock->erase(start);
```
rgacogne, line 704:
```suggestion
      start = lock->erase(start);
      break;
```

rgacogne (commented):
Nice work! I know the sources allowed to send us a notify are supposed to be trusted, but since this is UDP, do you think it would make sense to add a sanity check to prevent doing a freshness check too often? Like, only trigger a freshness check if the last one was more than `n` second(s) ago?

pdns/recursordist/rpzloader.cc
omoerbeek: There can be multiple thread handling the same rpz (e.g. old ones that are still sleeping and will conclude it's time to stop when woken). Notifying all of them has the consequence that the old thread cleans itself up immediately and that a newer thread for the same zone gets refreshed as well.

omoerbeek:
> Nice work! I know the sources allowed to send us a notify are supposed to be trusted, but since this is UDP, do you think it would make sense to add a sanity check to prevent doing a freshness check too often? Like, only trigger a freshness check if the last one was more than `n` second(s) ago?

Yes, that makes sense, will do.

pdns/recursordist/rpzloader.cc
rgacogne: But there is only one entry with our thread ID, right?

omoerbeek: Right. But the key into the `multimap` is a `DNSName`

rgacogne: I'm sorry to be so thick but I I'm afraid I still don't get it :-) This loop is iterating over the multimap to remove the entry corresponding to our thread, so I was assuming that once we have done that there is no need to iterate over the remaining entries, but perhaps I'm wrong about that?

omoerbeek: Consider the scenario with an Lua reload. That wil create a new thread serving RPZ zone `X`. The old thread will also doing that and will not quit until it awakens. So after the reload there are two entries serving `X` in the multimap. 

When a notify comes in and we iterate over those two entries, sending then both a `notify_one`. The first thread will wake up, see its config is outdated by the generation mechanism and remove itself from the multimap and quit, the second will wake up and trigger a refresh without removing itself.

omoerbeek: Oops, yes, sorry about that, was looking at the wrong loop in the the editor :-(. Your suggestion is correct.

rgacogne (approved):

omoerbeek:
The new rate limiting code makes the test fail, will look into it soon.

PR #13713, opened 2024-01-12
dnsdist: DoH3 add content-type header information in responses

### LINKED ISSUE #13690: dnsdist 1.9.0-pre name resolution via DoH fails in Chrome when h3 is active
- [X] This is not a support question, I have read [about opensource](https://www.powerdns.com/opensource.html) and will send support questions to the IRC channel, [Github Discussions](https://github.com/PowerDNS/pdns/discussions/) or the mailing list.
- [X] I have read and understood the ['out in the open' support policy](https://blog.powerdns.com/2016/01/18/open-source-support-out-in-the-open/)

 - Program: dnsdist 
 - Issue type: Bug report

### Short description

Chrome does not like the responses from the DoH3 endpoint of dnsdist 1.9.0-pre.

This is a follow-up to #13687 to specifically address the Chrome/Chromium problem that was mentioned there in passing.

### Environment

 - Operating system: Debian
 - Software version: 1.9.0-pre / master (dnsdist-1.9.0-alpha4.160.master.g6c8e140a3)
 - Software source: compiled myself

### Steps to reproduce

1. Run dnsdist with DoH2 and DoH3 endpoints. Something to the effect of:
```
v = "127.0.0.1"
addDOHLocal(v, '/etc/dnsdist/cert/fullchain.pem', '/etc/dnsdist/cert/privkey.pem', '/dns-query',{ customResponseHeaders={ ["alt-svc"]= "h3=\":443\"; ma=3600" } })
addDOH3Local(v, '/etc/dnsdist/cert/fullchain.pem', '/etc/dnsdist/cert/privkey.pem')
newServer("9.9.9.9")
```
2. Configure Chrome with a doh url of `https://dns.example/dns-query` (maybe restart to be sure)
3. Browse

### Expected behaviour

Chrome should be able to resolve names and work normally

### Actual behaviour

Chrome just says `ERR_NAME_NOT_RESOLVED` to the user

In the `chrome://net-export/` log, I also found:
```
t=109 [st=1]       -HOST_RESOLVER_MANAGER_REQUEST
                    --> net_error = -800 (ERR_DNS_MALFORMED_RESPONSE)
t=109 [st=1]     -TRANSPORT_CONNECT_JOB_CONNECT
                  --> net_error = -105 (ERR_NAME_NOT_RESOLVED)
```
There are probably more details to be found through the logging mechanism if necessary, but the problem appears to be entirely consistent (for me at least) and the amount of log data immediately becomes quite huge. I'm sure I can add a log if that is actually helpful, but for the moment I'm assuming it is easier to observe what happens locally.

### Other information

The same dnsdist setup with only h2 works fine, there seems to be something specific to the h3 endpoint that Chrome does not like.

chbruyand:
This adds `Content-Type: application/dns-message` header in HTTP/3 responses which seems to make chromium happier...

This also unify `vinfolog()` content in the QUIC and HTTP/3 response paths so that we display the response size and always use `DoH3` just like in the query log (and not `http/3`).

Fixes #13690 on my setup.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdistdist/doh3.cc
+++ b/pdns/dnsdistdist/doh3.cc
@@ -319,7 +319,7 @@ static void h3_send_response(H3Connection& conn, const uint64_t streamID, uint16
   auto returnValue = quiche_h3_send_response(conn.d_http3.get(), conn.d_conn.get(),
                                              streamID, headers.data(),
                                              // do not include content-type header info if there is no content
-                                             (len > 0 ? headers.size() : headers.size() - 1),
+                                             (len > 0 && statusCode == 200U ? headers.size() : headers.size() - 1),
                                              len == 0);
   if (returnValue != 0) {
     /* in theory it could be QUICHE_H3_ERR_STREAM_BLOCKED if the stream is not writable / congested, but we are not going to handle this case */
--- a/pdns/dnsdistdist/doh3.cc
+++ b/pdns/dnsdistdist/doh3.cc
@@ -159,7 +159,7 @@ public:
 
     if (!unit->ids.selfGenerated) {
       double udiff = unit->ids.queryRealTime.udiff();
-      vinfolog("Got answer from %s, relayed to %s (http/3), took %f us", unit->downstream->d_config.remote.toStringWithPort(), unit->ids.origRemote.toStringWithPort(), udiff);
+      vinfolog("Got answer from %s, relayed to %s (DoH3, %d bytes), took %f us", unit->downstream->d_config.remote.toStringWithPort(), unit->ids.origRemote.toStringWithPort(), unit->response.size(), udiff);
 
       auto backendProtocol = unit->downstream->getProtocol();
       if (backendProtocol == dnsdist::Protocol::DoUDP && unit->tcp) {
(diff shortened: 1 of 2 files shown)


hlindqvist:
Fixing the content-type problem does indeed seem to make Chrome happy.

However, maybe this should only be set for DNS responses, leaving http-level error responses as text/plain? (As those are just plain text, not dns data)

rgacogne:
>However, maybe this should only be set for DNS responses, leaving http-level error responses as text/plain? (As those are just plain text, not dns data)

Agreed, it should be mostly the case with the latest version of this PR, I think? We could n theory have non-DNS responses with a 200 status but I don't think it can happen with current code.

hlindqvist:
> However, maybe this should only be set for DNS responses, leaving http-level error responses as text/plain? (As those are just plain text, not dns data)
> Agreed, it should be mostly the case with the latest version of this PR, I think? We could n theory have non-DNS responses with a 200 status but I don't think it can happen with current code.

Ah sorry, I built a6a6615cc earlier and did not notice the update. The current version seems good enough until there is responsemap support / similar also for DoH3 👍

Issue #13672, opened 2023-12-29
dnsdist: No responses when DNStap is enabled with DoH3 and DoQ


dmachard:
- Program: dnsdist
 - Issue type: Bug report

### Short description

When DNStap logging is enabled in config, DNS resolutions failed with DoQ and DoH3.

### Environment
 - Software source: dnsdist 1.9.0 alpha4

### Steps to reproduce

Configuration used

```
setLocal('0.0.0.0:53', { reusePort=true })
addDOQLocal('0.0.0.0:853', '/etc/dnsdist/cert.pem', '/etc/dnsdist/key.pem')
addDOH3Local('0.0.0.0:443', '/etc/dnsdist/cert.pem', '/etc/dnsdist/key.pem')

fstl = newFrameStreamTcpLogger("192.168.1.17:6000")
addAction(AllRule(), DnstapLogAction("dnsdist-tcp", fstl))
```

The following appears 

```
Got an error while parsing a TCP query from 192.168.1.210:48434, id 30750: Unhandled protocol for dnstap: DNS over HTTP/3
Got an error while parsing a TCP query from 192.168.1.210:48435, id 30751: Unhandled protocol for dnstap: DNS over QUIC

```
### Other information

For DoQ I did the minimal changes in code with success

in dnsdist-lua-actions.cc 

```
else if (protocol == dnsdist::Protocol::DoQ) {
	return DnstapMessage::ProtocolType::DoQ;
}
```
	  
in dnstap.hh , in the enum class ProtocolType I add  `DoQ = 7`
For DoH3,  what are you planned, reuse the DoH type or just UDP or a new one ?

PS: I can do a PR if you want.

hlindqvist:
> For DoH3, what are you planned, reuse the DoH type or just UDP or a new one ?

I think "UDP" would be very misleading, and while there is indeed a type for DoQ (as per https://github.com/dnstap/dnstap.pb/blob/master/dnstap.proto#L67 ) there is no specific type for "DoH3", which presumably should just be recorded as DoH.

dmachard:
Issue opened https://github.com/dnstap/dnstap.pb/issues/20

edmonds:
[...]

Issue #13673, opened 2023-12-29
dnsdist:  DoQ and Doh3 are sent with UDP protocol type in protobuf messages


dmachard:
- Program: dnsdist
 - Issue type: Bug report

### Short description

Perhaps it's not a issue but currently DoQ and Doh3 queries are sent through  protobuf messages with `UDP` protocol type. It's not wrong but it does not allow them to be follow them.

### Environment
 - Software version: dnsdist 1.9.0-alpha4

### Steps to reproduce

Configuration used

```
setLocal('0.0.0.0:53', { reusePort=true })
addDOQLocal('0.0.0.0:853', '/etc/dnsdist/cert.pem', '/etc/dnsdist/key.pem')
addDOH3Local('0.0.0.0:443', '/etc/dnsdist/cert.pem', '/etc/dnsdist/key.pem')

rl = newRemoteLogger("192.168.1.17:6001")
addAction(AllRule(),RemoteLogAction(rl, nil, {serverID="dnsdist"}))
```

### Expected behaviour

Specific protocol types for each one

hlindqvist:
> It's not wrong but it does not allow them to be follow them.

I think it's very clear that it's wrong, as it is not DNS over UDP.

Probably relevant:
https://github.com/dnstap/dnstap.pb/blob/master/dnstap.proto#L67  DoQ has its own type, presumably "DoH3" is still just DoH.

rgacogne:
Yep, it's a bug. Contrary to dnstap for which we cannot decide by ourselves, I think we want a separate value for DoH3 in our own protobuf format.

PR #13716, opened 2024-01-15
dnsdist: properly set protocol when logging

chbruyand:
### Short description
This PR properly sets protocol when logging via dnstap or protobuf when using DNS over QUIC and DNS over HTTP/3.

For protobuf, we add a `httpVersion` field so we can distinguish between DoH and DoHTTP/3 (https://github.com/PowerDNS/dnsmessage/pull/8).
We should eventually do something similar for dnstap if there is a possibility to do so (https://github.com/dnstap/dnstap.pb/issues/20).

Fix https://github.com/PowerDNS/pdns/issues/13672 and https://github.com/PowerDNS/pdns/issues/13673

Todo:
- [x] add regression tests
- [x] should I include the formatting commit ?

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdist-lua-actions.cc
+++ b/pdns/dnsdist-lua-actions.cc
@@ -1491,7 +1491,7 @@ static DnstapMessage::ProtocolType ProtocolToDNSTap(dnsdist::Protocol protocol)
   else if (protocol == dnsdist::Protocol::DoT) {
     return DnstapMessage::ProtocolType::DoT;
   }
-  else if (protocol == dnsdist::Protocol::DoH) {
+  else if (protocol == dnsdist::Protocol::DoH || protocol == dnsdist::Protocol::DoH3) {
     return DnstapMessage::ProtocolType::DoH;
   }
   else if (protocol == dnsdist::Protocol::DNSCryptUDP) {
@@ -1499,6 +1499,9 @@ static DnstapMessage::ProtocolType ProtocolToDNSTap(dnsdist::Protocol protocol)
   }
   else if (protocol == dnsdist::Protocol::DNSCryptTCP) {
     return DnstapMessage::ProtocolType::DNSCryptTCP;
+  }
+  else if (protocol == dnsdist::Protocol::DoQ) {
+    return DnstapMessage::ProtocolType::DoQ;
   }
   throw std::runtime_error("Unhandled protocol for dnstap: " + protocol.toPrettyString());
 }
--- a/pdns/dnsdist-lua-actions.cc
+++ b/pdns/dnsdist-lua-actions.cc
@@ -327,8 +327,10 @@ public:
   }
 
 private:
-  std::string d_pool;
-  bool d_stopProcessing;
+  // NOLINTNEXTLINE(cppcoreguidelines-avoid-const-or-ref-data-members)
+  const std::string d_pool;
+  // NOLINTNEXTLINE(cppcoreguidelines-avoid-const-or-ref-data-members)
+  const bool d_stopProcessing;
 };
 
 class QPSPoolAction : public DNSAction
@@ -355,8 +357,10 @@ public:
 
 private:
   mutable LockGuarded<QPSLimiter> d_qps;
-  std::string d_pool;
-  bool d_stopProcessing;
+  // NOLINTNEXTLINE(cppcoreguidelines-avoid-const-or-ref-data-members)
+  const std::string d_pool;
+  // NOLINTNEXTLINE(cppcoreguidelines-avoid-const-or-ref-data-members)
+  const bool d_stopProcessing;
 };
 
 class RCodeAction : public DNSAction
@@ -747,8 +751,10 @@ private:
   };
   static std::atomic<uint64_t> s_functionsCounter;
   static thread_local std::map<uint64_t, PerThreadState> t_perThreadStates;
-  std::string d_functionCode;
-  uint64_t d_functionID;
+  // NOLINTNEXTLINE(cppcoreguidelines-avoid-const-or-ref-data-members)
+  const std::string d_functionCode;
+  // NOLINTNEXTLINE(cppcoreguidelines-avoid-const-or-ref-data-members)
+  const uint64_t d_functionID;
 };
 
 std::atomic<uint64_t> LuaFFIPerThreadAction::s_functionsCounter = 0;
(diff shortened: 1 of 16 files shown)


pdns/dnsdist-protobuf.cc
```diff
@@ -149,13 +149,21 @@ void DNSDistProtoBufMessage::serialize(std::string& data) const
   }
   else if (distProto == dnsdist::Protocol::DoH) {
     protocol = pdns::ProtoZero::Message::TransportProtocol::DoH;
+    m.setHTTPVersion(pdns::ProtoZero::Message::HTTPVersion::HTTP2);
```
rgacogne, line 154: Technically it could be HTTP/1.1, when DoH is provided by h2o, but we unfortunately don't expose the HTTP version at the moment. I vote for ignoring this since we will be deprecating h2o support soon anyway.

rgacogne (approved):
Nice!

rgacogne:
>     * [ ]  should I include the formatting commit ?

If you can find the motivation to fix the clang-tidy warnings resulting from the reformat then yes, please :)

pdns/dnsdist-protobuf.cc
chbruyand: Wouldn't something like this do the trick ? also the `HAVE_LIBH2OEVLOOP` define would make it easy to find when removing the legacy code.

```suggestion
+#ifdef HAVE_DNS_OVER_HTTPS                                                                                                                                                                                                                                          #ifdef HAVE_LIBH2OEVLOOP
+    m.setHTTPVersion(pdns::ProtoZero::Message::HTTPVersion::HTTP1);
+#else
     m.setHTTPVersion(pdns::ProtoZero::Message::HTTPVersion::HTTP2);
+#endif
+#endif

Comment (rgacogne) [reply pdns/dnsdist-protobuf.cc:154]: It's not that easy, I'm afraid. When `HAVE_LIBH2OEVLOOP` is not enabled it is indeed always HTTP/2, but otherwise it might be HTTP/1 or HTTP/2, and right now we don't pass this information to the protobuf/dnstap layer.

Comment (rgacogne) [reply pdns/dnsdist-protobuf.cc:154]: We could add a method to `DNSQuestion` that would query a method in `DOHUnit` looking at `req->version`, but I'm not sure it's worth the complexity.

File: pdns/dnsdist-lua-actions.cc [hunk 2]
```diff
@@ -864,18 +869,18 @@ class LuaFFIPerThreadResponseAction : public DNSResponseAction
 
   static std::atomic<uint64_t> s_functionsCounter;
   static thread_local std::map<uint64_t, PerThreadState> t_perThreadStates;
-  const std::string d_functionCode;
-  const uint64_t d_functionID;
+  std::string d_functionCode;
+  uint64_t d_functionID;
```
rgacogne, line 873: I'm a bit surprised by this change?

pdns/dnsdist-lua-actions.cc
```diff
@@ -742,8 +747,8 @@ class LuaFFIPerThreadAction : public DNSAction
   };
   static std::atomic<uint64_t> s_functionsCounter;
   static thread_local std::map<uint64_t, PerThreadState> t_perThreadStates;
-  const std::string d_functionCode;
-  const uint64_t d_functionID;
+  std::string d_functionCode;
+  uint64_t d_functionID;
```
rgacogne, line 751: I'm a bit surprised by this change?

pdns/dnsdist-protobuf.hh
```diff
@@ -83,6 +87,8 @@
   };
   std::unordered_map<std::string, MetaValue> d_metaTags;
 
+  // FIXME wondering if the cost of moving to a shared_ptr would be that bad
```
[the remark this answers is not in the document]
rgacogne: For `DNSQuestion` and `DNSResponse` it would mean always allocating them as a `shared_ptr` since we cannot just turn them into `shared_ptr` just-in-time, so the cost is likely significant.

pdns/dnsdist-protobuf.hh
```diff
@@ -123,9 +127,9 @@ class ProtoBufMetaKey
 
   struct KeyTypeDescription
   {
-    const std::string d_name;
-    const Type d_type;
-    const std::function<std::vector<std::string>(const DNSQuestion&, const std::string&, uint8_t)> d_func;
+    std::string d_name;
+    Type d_type;
+    std::function<std::vector<std::string>(const DNSQuestion&, const std::string&, uint8_t)> d_func;
```
rgacogne, line 132: OK, now I'm getting it, you are removing `const` from members because of clang-tidy: https://clang.llvm.org/extra/clang-tidy/checks/cppcoreguidelines/avoid-const-or-ref-data-members.html
This is a very opinionated rule, and I'm not really convinced so I guess we need to discuss it with the rest of the team.

pdns/dnsname.hh
```diff
@@ -100,7 +100,7 @@ public:
   DNSName(DNSName&& a) = default;
 
   explicit DNSName(std::string_view sw); //!< Constructs from a human formatted, escaped presentation
-  DNSName(const char* p, int len, int offset, bool uncompress, uint16_t* qtype=nullptr, uint16_t* qclass=nullptr, unsigned int* consumed=nullptr, uint16_t minOffset=0); //!< Construct from a DNS Packet, taking the first question if offset=12. If supplied, consumed is set to the number of bytes consumed from the packet, which will not be equal to the wire length of the resulting name in case of compression.
+  DNSName(const char* pos, size_t len, size_t offset, bool uncompress, uint16_t* qtype = nullptr, uint16_t* qclass = nullptr, unsigned int* consumed = nullptr, uint16_t minOffset = 0); //!< Construct from a DNS Packet, taking the first question if offset=12. If supplied, consumed is set to the number of bytes consumed from the packet, which will not be equal to the wire length of the resulting name in case of compression.
```
rgacogne, line 103: This is a very scary change. It might be right, to be clear, but I'm not sure I can see all the implications. Why are we doing it in this PR?

rgacogne (commented):
Thanks a lot for dealing with the massive pain of delinting these files! I think we need to discuss one clang-tidy rule with the team, so I'll bring it up on the internal chat.

chbruyand:
> Thanks a lot for dealing with the massive pain of delinting these files! I think we need to discuss one clang-tidy rule with the team, so I'll bring it up on the internal chat.

I reverted the const member related changes and the `dnsname.{cc,hh}` files.

Sorry I've gone a bit wild with clang-tidy.

chbruyand:
There is still one clang-tidy warning left:
```
pdns/dnsdist-lua-actions.cc:945 narrowing conversion from 'std::vector::size_type' (aka 'unsigned long') to signed type 'int' is implementation-defined (bugprone-narrowing-conversions)
```

This should be fixed with the modification made in:
https://github.com/PowerDNS/pdns/pull/13723

The purposed fix might have some collateral damages, so I purpose to ignore that warning for the scope of this PR.

rgacogne (approved):

PR #13722, opened 2024-01-16
auth LUA createForward: allow non-hex word prefix

Habbie:
### Short description
best reviewed per commit

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/lua-record.cc
+++ b/pdns/lua-record.cc
@@ -771,7 +771,7 @@ static void setupLuaRecords(LuaContext& lua) // NOLINT(readability-function-cogn
         if(input.length() >= 8) {
           auto last8 = input.substr(input.length()-8);
           if(sscanf(last8.c_str(), "%02x%02x%02x%02x", &x1, &x2, &x3, &x4)==4) {
-            return std::to_string(x1)+"."+std::to_string(x2)+"."+std::to_string(x3)+"."+std::to_string(x4);
+            return std::to_string(x1) + "." + std::to_string(x2) + "." + std::to_string(x3) + "." + std::to_string(x4);
           }
         }
       }
--- a/pdns/lua-record.cc
+++ b/pdns/lua-record.cc
@@ -751,12 +751,12 @@ static void setupLuaRecords(LuaContext& lua)
         if (ip_parts.size() >= 4) {
           // 1-2-3-4 with any prefix (e.g. ip-foo-bar-1-2-3-4)
           string ret;
-          for (size_t n=4; n > 0; n--) {
-            auto octet = ip_parts[ip_parts.size() - n];
+          for (size_t index=4; index > 0; index--) {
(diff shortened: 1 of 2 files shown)


pdns/lua-record.cc
```diff
@@ -738,6 +738,13 @@ static void setupLuaRecords(LuaContext& lua)
       } else if (parts.size() >= 1) {
         // either hex string, or 12-13-14-15
         vector<string> ip_parts;
+
+        // allow a word without - in front, as long as it does not contain anything that could be a number
+        size_t nonhexprefix = strcspn(parts[0].c_str(), "0123456789abcdef");
```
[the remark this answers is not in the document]
Habbie: it should not be necessary, but it doesn't hurt, and feels safer. will do

pdns/lua-record.cc
```diff
@@ -738,14 +738,15 @@ static void setupLuaRecords(LuaContext& lua)
       } else if (parts.size() >= 1) {
         // either hex string, or 12-13-14-15
         vector<string> ip_parts;
+        auto input = parts[0];
```
chbruyand, line 741: I *think* you want `auto& input` otherwise you get a copy ?

chbruyand: Also even if we check the size above, `.at()` would be preferable on a vector.

pdns/lua-record.cc
```diff
@@ -766,8 +766,12 @@ static void setupLuaRecords(LuaContext& lua)
...
           ret.resize(ret.size() - 1); // remove trailing dot after last octet
           return ret;
-        } else if(input.length() >= 8 && sscanf(input.c_str()+(input.length()-8), "%02x%02x%02x%02x", &x1, &x2, &x3, &x4)==4) {
-          return std::to_string(x1)+"."+std::to_string(x2)+"."+std::to_string(x3)+"."+std::to_string(x4);
+        }
+        if(input.length() >= 8) {
+          auto last8 = input.substr(input.length()-8);
+          if(sscanf(last8.c_str(), "%02x%02x%02x%02x", &x1, &x2, &x3, &x4)==4) {
+            return std::to_string(x1)+"."+std::to_string(x2)+"."+std::to_string(x3)+"."+std::to_string(x4);
+          }
```
chbruyand, line 774: I know the file isn't formatted, but it would be easier to read with a bit of spacing

pdns/lua-record.cc
Habbie: > I _think_ you want `auto& input` otherwise you get a copy ?

right, in an earlier version I wanted to leave `parts[0]` intact, but that's no longer true.

Habbie:
all review comments processed

chbruyand (approved):

PR #13723, opened 2024-01-18
DNSName: correct len and offset types

chbruyand:
### Short description
This PR ~format `dnsname.{hh,cc}` files and~ corrects the type of `len` and `offset` arguments from `int` to `size_t`.

### Checklist

I have:
- [x] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsname.cc
+++ b/pdns/dnsname.cc
@@ -137,8 +137,8 @@ void DNSName::packetParser(const char* qpos, size_t len, size_t offset, bool unc
       labellen &= (~0xc0);
       size_t newpos = (labellen << 8) + *(const unsigned char*)pos;
 
-      if(newpos < offset) {
-        if(newpos < static_cast<size_t>(minOffset)) {
+      if (newpos < offset) {
+        if (newpos < minOffset) {
           throw std::range_error("Invalid label position during decompression ("+std::to_string(newpos)+ " < "+std::to_string(minOffset)+")");
         }
         if (++depth > 100) {
--- a/pdns/dnsname.cc
+++ b/pdns/dnsname.cc
@@ -120,10 +120,12 @@ void DNSName::packetParser(const char* qpos, size_t len, size_t offset, bool unc
   unsigned char labellen;
   const unsigned char *opos = pos;
 
-  if (offset >= len)
+  if (offset >= len) {
     throw std::range_error("Trying to read past the end of the buffer ("+std::to_string(offset)+ " >= "+std::to_string(len)+")");
-  if (offset < static_cast<size_t>(minOffset))
+  }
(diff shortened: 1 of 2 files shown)


rgacogne:
I'm wondering if it would be better to move the formatting of the files to a different pull request, so that we can focus on the (hopefully) non-impact of the type changes in this PR? In theory it should not matter but we get clang-tidy warnings for the whole PR, not per commit, so it feels like this PR is changing a lot of things when it isn't.

chbruyand:
> I'm wondering if it would be better to move the formatting of the files to a different pull request, so that we can focus on the (hopefully) non-impact of the type changes in this PR? In theory it should not matter but we get clang-tidy warnings for the whole PR, not per commit, so it feels like this PR is changing a lot of things when it isn't.

Agreed. I'll also stick to not format/clean anything not *required*.

rgacogne (approved):
The change looks good to me, there is no valid reason for the length or the offset to be negative. Of course there is a small risk than an existing piece of code relies on the existing behaviour, but then I think it's a bug that needs to be fixed. 
I just pushed a commit to a throw-away branch (https://github.com/rgacogne/pdns/tree/dnsname-test-negative) that aborts if the length or the offset is negative, let's see if it passes our CI.

rgacogne:
> Of course there is a small risk than an existing piece of code relies on the existing behaviour, but then I think it's a bug that needs to be fixed. I just pushed a commit to a throw-away branch (https://github.com/rgacogne/pdns/tree/dnsname-test-negative) that aborts if the length or the offset is negative, let's see if it passes our CI.

The CI run was successful!

I would not mind at least one other review before merging this PR, though :-)

Habbie (approved):

pdns/dnsname.cc
```diff
@@ -134,16 +135,19 @@ void DNSName::packetParser(const char* qpos, int len, int offset, bool uncompres
         throw std::range_error("Found compressed label, instructed not to follow");
 
       labellen &= (~0xc0);
-      int newpos = (labellen << 8) + *(const unsigned char*)pos;
+      size_t newpos = (labellen << 8) + *(const unsigned char*)pos;
 
       if(newpos < offset) {
-        if(newpos < (int) minOffset)
+        if(newpos < static_cast<size_t>(minOffset)) {
```
omoerbeek, line 141: I think this cast is not needed, as `newpos` and  and `minOffser` are both unsigned types.

omoerbeek (approved):
One nit

pdns/dnsname.cc
chbruyand: Done, thanks.

PR #13728, opened 2024-01-19
STDOUT instead of STDERR. Probably a typo.

paddg:
### Short description

It's not an error message. So I'd go for STDOUT in this case.

### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code


paddg:
But I'm not sure if we should even print this message every time

omoerbeek:
> But I'm not sure if we should even print this message every time

IMO it is good while YAML support is still "new".

omoerbeek (dismissed):

omoerbeek (approved):

PR #13749, opened 2024-01-29
ci: Exclude the fuzzing/corpus dir from the 'no binary files' check

rgacogne:
### Short description

The fuzzing/corpus directory contains payloads that are used to speed-up the process of finding interesting code paths, and some of these are binary files.

### Checklist

I have:
- [ ] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -276,4 +276,4 @@ jobs:
       uses: actions/checkout@v4
       with:
         fetch-depth: 2
-    - run: if [[ "$(file -i --dereference $(git diff --name-only HEAD^..HEAD) | grep binary | grep -v 'image/' | grep -v 'inode/x-empty' | grep -v 'inode/directory')" != "" ]]; then exit 1; fi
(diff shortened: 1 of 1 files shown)


romeroalx (approved):

romeroalx:
@rgacogne, the changes look good to me. I am just curious why whitelisting the whole folder instead of specific files (besides being a tedious task :))

rgacogne:
It's a good question :-) Technically we could allow-list all the files that are binary ones, or the directories that contain only binary files, for example. I choose to instead allow the whole corpus directory for simplicity, because most of the files in that directory are binary ones, except for the ones in txt-records/ and zones/ which are kind of an exception because most of the parsers we want to fuzz are likely dealing with raw data. So it means we will likely have to edit the allow-list every time we add a new file to the corpus.
But If you think it's better to list specific directories or even files, I'll do so!

romeroalx:
Both options are fine by me, but I agree with you, I will leave this PR as is.

The main purpose of the workflow job was to identify pushes of binaries made by accident, mainly from builds. I think it already did his job with this folder.

rgacogne:
Thanks! Agreed, the check works exactly as expected, nice work!
I'm going to merge this PR as it is, allow-listing binary files in fuzzing/corpus, but we can revisit later if needed.

PR #13752, opened 2024-01-31
auth: lua records, pickchashed function

chbruyand:
### Short description
Adds the `pickchashed()` function for lua records.

This function works almost like :func:`pickwhashed` while bringing the following properties:
- reordering the list of entries won't affect the distribution
- updating the weight of an entry will only affect a part of the distribution
- because of the previous properties, the CPU and memory cost is a bit higher than :func:`pickwhashed`

### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code


pdns/lua-record.cc
```diff
@@ -616,6 +618,104 @@ typedef struct AuthLuaRecordContext
...
+    }
+    std::sort(locked->begin(), locked->end());
+  }
+};
+
+static std::map<
+  std::tuple<int, std::string, std::string>, // zoneid qname entry
+  std::shared_ptr<EntryHashesHolder> // entry w/ corresponding hashes
+  >
+s_zone_hashes;
```
Habbie, line 655: I guess we need some cleanup for this?

Habbie (commented):
Did a first short review. Concept makes sense. I posted a comment - I fear the cache may grow without bounds if people keep adding/removing records that call this function.

Did you test in practice that adding/removing items does not make every client move? A test for that would be cool but I see how it's a bunch of work.

chbruyand:
> Did a first short review. Concept makes sense. I posted a comment - I fear the cache may grow without bounds if people keep adding/removing records that call this function (unless I missed something).

Yes, that's a potential issue. It's kind of the same design than with the `if*up` tool. Computing the hashes is quiet expensive. We could have a time field updated each time an entry is used and have a maintenance utility cleaning old stuff. Question would be how old ?

> Did you test in practice that adding/removing items does not make every client move? A test for that would be cool but I see how it's a bunch of work.

Hashes are pre computed and hashed based on `"{entry name}-{weight}"` such as increasing or decreasing the weight will only add or remove entries, but `"{entry name}-1", "{entry name}-2` result hashes won't change.

Habbie:
> Question would be how old ?

Configurable, defaulting to 24 hours?

chbruyand:
Recursor test failure is unrelated to this PR

Habbie (approved):

Issue #11581, opened 2022-04-26
dnsdist: our eBPF filtering could use regression tests


rgacogne:
- Program: dnsdist 
 - Issue type: Feature request

### Short description
We broke a fair amount of eBPF features in 1.7.0, I would like to avoid doing that again :)

PR #13761, opened 2024-02-01
dnsdist: Add regression tests for eBPF blocks (static / dynamic)

rgacogne:
### Short description

Fixes https://github.com/PowerDNS/pdns/issues/11581

I have tested this locally but I'm pretty sure I forgot something that will fail on our CI, let's see :)

### Checklist

I have:
- [ ] compiled this code
- [x] tested this code
- [ ] documented the code


rgacogne:
OK, so we cannot use/test eBPF from inside the container that we use for the "Build and test everything" workflow. I need to figure out how to run this test, then.

PR #13768, opened 2024-02-06
dnsdist: Add a Lua maintenance hook

rgacogne:
### Short description

Having a hook that can accept multiple callbacks makes it a lot easier to manage the execution of several, independant tasks that needs to be called regularly.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdistdist/dnsdist-lua-hooks.cc
+++ b/pdns/dnsdistdist/dnsdist-lua-hooks.cc
@@ -1,0 +1,32 @@
+
+#include "dnsdist-lua-hooks.hh"
+#include "dnsdist-lua.hh"
+#include "lock.hh"
+
+namespace dnsdist::lua::hooks
+{
+static LockGuarded<std::vector<MaintenanceCallback>> s_maintenanceHook;
+
+void runMaintenanceHook(const LuaContext& context)
+{
+  (void)context;
+  for (const auto& callback : *(s_maintenanceHook.lock())) {
+    callback();
+  }
+}
+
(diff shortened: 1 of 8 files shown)


pdns/dnsdistdist/dnsdist-lua-hooks.cc
```diff
@@ -0,0 +1,32 @@
+
+#include "dnsdist-lua-hooks.hh"
+#include "dnsdist-lua.hh"
+#include "lock.hh"
+
+namespace dnsdist::lua::hooks
+{
+static LockGuarded<std::vector<MaintenanceCallback>> s_maintenanceHook;
+
+void runMaintenanceHook(const LuaContext& context)
```
chbruyand, line 10:
```suggestion
void runMaintenanceHooks(const LuaContext& context)
```
Maybe ?

chbruyand (approved):
Just a small nit that is a matter of taste, LGTM otherwise

pdns/dnsdistdist/dnsdist-lua-hooks.cc
rgacogne: Done!

rgacogne:
The recursor regression failure can be ignored (the test uses an external DNS zone which is currently broken).
As for `clang-tidy` warning about a potential leak in LuaWrapper's code, which was not modified by this pull request, it seems like a false positive to me: 
```
ext/luawrapper/include/LuaContext.hpp:2888 Potential memory leak (clang-analyzer-cplusplus.NewDeleteLeaks)
```

chbruyand (approved):

PR #13770, opened 2024-02-07
auth: Add setting to make TSIG required for DNS updates

### LINKED ISSUE #12234: auth: Cannot enforce TSIG for DNS updates
- Program: Authoritative
 - Issue type: Bug report

### Short description
If you want to allow TSIG verififed DNS updates, but cannot check or don't want to check sender IP, you configure PowerDNS to allow DNS updates from any source, and use per-zone TSIG keys to verify the updates.

This has a problem though that when you provision new zone, or have zones that you would prefer not to be updated, it is easy to forget to prevent DNS updates for these zones.

Few ways to fix this:

 - Add new dnsupdate-require-tsig option
 - Add new metadata to prevent updates for zone (or metadata to allow updates in the first place)
 - Remove global allow-dnsupdates-from option, and require providing it as per-zone metadata

### Environment

 - Operating system: GNU/Linux
 - Software version: 4.7

### Steps to reproduce
1. Set allow-dnsupdate-from = 0.0.0.0/0 ::/0
2. Configure two zones
3. Add tsig key to one of the zones
4. You can now update the other zone too

### Expected behaviour
I suppose the expected behaviour might be that the other zone wouldn't go wide open by accident.

### Actual behaviour
You can update any zone with no TSIG key attached.

### Other information
N/A

Assumeru:
### Short description
Closes #12234 by adding the proposed `dnsupdate-require-tsig` setting.

Not ideal perhaps, but backwards compatible, which is nice. Adding to `test_GSSTSIG.py` is a bit iffy considering it's not GSS related but it is convenient.

### Checklist
I have:
- [x] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/docs/dnsupdate.rst
+++ b/docs/dnsupdate.rst
@@ -46,6 +46,13 @@ combination with the ``ALLOW-DNSUPDATE-FROM`` :doc:`domain metadata <domainmetad
 zone. Setting a range here and in ``ALLOW-DNSUPDATE-FROM`` enables updates
 from either address range.
 
+``dnsupdate-require-tsig``
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+A setting to require DNS updates to be signed by a valid TSIG signature.
+The default is no, which means zones without TSIG keys can be updated by
+unauthenticated agents operating from an allowed address range.
+
 ``forward-dnsupdate``
 ~~~~~~~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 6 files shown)


Assumeru:
Should I rename variables in `PacketHandler::processUpdate` to appease clang-tidy?

Habbie:
I'll handle that, thanks!

Spudz76:
The (documentation)[https://doc.powerdns.com/authoritative/settings.html#dnsupdate-require-tsig] lists `dnsupdate-require-tsig` as if it always existed (no "New in version x.y.z" note such as on `consistent-backends`).  I just upgraded to 4.9.x hoping I would gain this option and it's still not there... is it even in 5.0.x?

Habbie:
> is it even in 5.0.x?

it is currently on `master` so it will be in 5.0.x. I'll update the docs.

PR #13776, opened 2024-02-10
clang-tidy fixes

neheb:
Some other stuff as well.

### DIFF
--- a/pdns/pdnsutil.cc
+++ b/pdns/pdnsutil.cc
@@ -924,10 +924,11 @@ static int increaseSerial(const DNSName& zone, DNSSECKeeper &dk)
 
   sd.db->startTransaction(zone, -1);
 
-  if (!sd.db->replaceRRSet(sd.domain_id, zone, rr.qtype, {rr})) {
-   sd.db->abortTransaction();
-   cerr<<"Backend did not replace SOA record. Backend might not support this operation."<<endl;
-   return -1;
+  auto rrs = vector<DNSResourceRecord>{rr};
+  if (!sd.db->replaceRRSet(sd.domain_id, zone, rr.qtype, rrs)) {
+    sd.db->abortTransaction();
+    cerr << "Backend did not replace SOA record. Backend might not support this operation." << endl;
+    return -1;
   }
 
   if (sd.db->doesDNSSEC()) {
--- a/modules/geoipbackend/geoipbackend.cc
+++ b/modules/geoipbackend/geoipbackend.cc
@@ -895,7 +895,7 @@ bool GeoIPBackend::getDomainInfo(const DNSName& domain, DomainInfo& di, bool /* 
 {
   ReadLock rl(&s_state_lock);
 
-  for (GeoIPDomain dom : s_domains) {
+  for (const GeoIPDomain& dom : s_domains) {
     if (dom.domain == domain) {
       SOAData sd;
       this->getSOA(domain, sd);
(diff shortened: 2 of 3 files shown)


pdns/pdnsutil.cc
```diff
@@ -914,10 +913,11 @@ static int increaseSerial(const DNSName& zone, DNSSECKeeper &dk)
 
   sd.db->startTransaction(zone, -1);
 
-  if (!sd.db->replaceRRSet(sd.domain_id, zone, rr.qtype, {rr})) {
-   sd.db->abortTransaction();
-   cerr << "Backend did not replace SOA record. Backend might not support this operation." << '\n';
-   return -1;
+  vector<DNSResourceRecord> rrs;
+  if (!sd.db->replaceRRSet(sd.domain_id, zone, rr.qtype, rrs)) {
```
omoerbeek, line 917: Not using `rr` anymore?

omoerbeek (changes requested):
I *think* the endl -> '\n' change will change flushing behaviour, which is unwanted.

pdns/pdnsutil.cc
neheb: Fixed.

neheb:
> The endl -> '\n' change will change flushing behaviour, which is unwanted. We want the implicit flush.

Removed.

pdns/sdig.cc
```diff
@@ -422,7 +422,7 @@ try {
     Socket sock(dest.sin4.sin_family, SOCK_STREAM);
     sock.setNonBlocking();
     setTCPNoDelay(sock.getHandle()); // disable NAGLE, which does not play nicely with delayed ACKs
-    TCPIOHandler handler(subjectName, false, sock.releaseHandle(), timeout, std::move(tlsCtx));
+    TCPIOHandler handler(subjectName, false, sock.releaseHandle(), timeout, tlsCtx);
```
omoerbeek, line 425: Coverity does not seem to agree, see previous commit: https://github.com/PowerDNS/pdns/commit/3b45a4347c8af1b28e6871597221eb906c143a29

rgacogne: Coverity was right at the time, but https://github.com/PowerDNS/pdns/commit/224085cccd7d7f52ceac3bd1f0ec8e4cd262948d changed the constructor.

pdns/pdnsutil.cc
```diff
@@ -913,10 +912,11 @@ static int increaseSerial(const DNSName& zone, DNSSECKeeper &dk)
 
   sd.db->startTransaction(zone, -1);
 
-  if (!sd.db->replaceRRSet(sd.domain_id, zone, rr.qtype, {rr})) {
-   sd.db->abortTransaction();
-   cerr<<"Backend did not replace SOA record. Backend might not support this operation."<<endl;
-   return -1;
+  vector<DNSResourceRecord> rrs(rr);
```
rgacogne, line 915:
```suggestion
  vector<DNSResourceRecord> rrs = {rr};
```

neheb: why =?

rgacogne: The initial syntax is wrong, there is no constructor in the `std::vector` class that accepts a single element. Now if your question is why ` = {rr}` and not `({rr})`, I don't really have an opinion since performance is completely irrelevant here, so whatever compiles would work for me.

neheb:
CI passes now.

pdns/pdnsutil.cc
neheb: This is fixed.

pdns/sdig.cc
neheb: Fixed.

neheb:
ping

rgacogne (approved):

Habbie (approved):

PR #13814, opened 2024-02-19
Update remote.rst to reflect that getAllDomains might be mandatory

wwijkander:
### Short description
Update remote backend documentation to reflect the fact that getAllDomains is mandatory unless you change the Zone Cache default

### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/docs/backends/remote.rst
+++ b/docs/backends/remote.rst
@@ -1521,7 +1521,7 @@ Response:
 
 Get DomainInfo records for all domains in your backend.
 
--  Mandatory: no
+-  Mandatory: unless the zone cache has been disabled by setting :ref:`setting-zone-cache-refresh-interval` to ``0`` (not recommended for performance reasons)(since 4.5.0)
 -  Parameters: include_disabled
 -  Reply: array of DomainInfo
 


docs/backends/remote.rst
```diff
@@ -1521,7 +1521,7 @@ Response:
 
 Get DomainInfo records for all domains in your backend.
 
--  Mandatory: no
+-  Mandatory: unless the zone cache has been disabled by setting zone-cache-refresh-interval to 0(not recommended for performance reasons)
```
rgacogne, line 1524:
```suggestion
-  Mandatory: unless the zone cache has been disabled by setting :ref:`zone-cache-refresh-interval` to ``0`` (recommended for performance reasons)
```

rgacogne: Uh, no, sorry, I somehow managed to get rid of the `not` while I just wanted to add a space :man_facepalming:

omoerbeek: I thin the :ref: should be :ref:`setting-zone-cache-refresh-interval`

Habbie: you can't, but you can make a new comment with an updated suggestion. The push works too, of course :-)

Habbie (approved):

PR #13820, opened 2024-02-22
Update overload-queue-length and overload-drops documentation

devicenull:
It wasn't clear to be what overload-queue-length did until I looked at the code.  This is a documentation only update

### DIFF
--- a/docs/performance.rst
+++ b/docs/performance.rst
@@ -236,7 +236,7 @@ Number of currently open TCP connections
 
 overload-drops
 ^^^^^^^^^^^^^^
-Number of questions dropped because backends overloaded
+Number of questions dropped because backends overloaded (backends are overloaded if they have more queries outstanding then the value of :ref:`setting-overload-queue-length`)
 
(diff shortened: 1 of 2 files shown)


docs/performance.rst
```diff
@@ -236,7 +236,7 @@ Number of currently open TCP connections
 
 overload-drops
 ^^^^^^^^^^^^^^
-Number of questions dropped because backends overloaded
+Number of questions dropped because backends overloaded (backends are overloaded if they have more queries outstanding then the value of :ref:`setting-overload-queue-length`)
```
rgacogne, line 239:
```suggestion
Number of questions dropped because backends overloaded (backends are overloaded if they have more outstanding queries than the value of :ref:`setting-overload-queue-length`)
```

rgacogne (changes requested):
Thanks a lot! One nit but other than that the changes look good to me.

devicenull:
Change seems fine to me

rgacogne (approved):

PR #13824, opened 2024-02-23
rec: log if a dnssec related limit was hit (if log_bogus is set)

omoerbeek:
### Short description

Note: only structured logging case is done, as non-structured logging is deprecated.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/recursordist/pdns_recursor.cc
+++ b/pdns/recursordist/pdns_recursor.cc
@@ -569,10 +569,10 @@ static PolicyResult handlePolicyHit(const DNSFilterEngine::Policy& appliedPolicy
           break;
         }
         catch (const pdns::validation::TooManySEC3IterationsException& e) {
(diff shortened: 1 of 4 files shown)


rgacogne (approved):

rgacogne:
Not for this PR but I'm wondering if we could use `getrusage(RUSAGE_THREAD, ..)` (Linux-only, unfortunately) to have actual CPU usage limit per query.

omoerbeek:
> Not for this PR but I'm wondering if we could use `getrusage(RUSAGE_THREAD, ..)` (Linux-only, unfortunately) to have actual CPU usage limit per query.

That is something worth investigating indeed. BTW, OpenBSD also has `getrusage(RUSAGE_THREAD, ..)`.

PR #13834, opened 2024-02-26
dnsdist: Properly handle a failure of the first lazy health-check

rgacogne:
### Short description

Needs more testing!

Fixes https://github.com/PowerDNS/pdns/issues/13837
### Checklist

I have:
- [x] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdistdist/dnsdist-backend.cc
+++ b/pdns/dnsdistdist/dnsdist-backend.cc
@@ -692,6 +692,7 @@ bool DownstreamState::healthCheckRequired(std::optional<time_t> currentTime)
         lastResults.clear();
         vinfolog("Backend %s reached the lazy health-check threshold (%f%% out of %f%%, looking at sample of %d items with %d failures), moving to Potential Failure state", getNameWithAddr(), current, maxFailureRate, totalCount, failures);
         stats->d_status = LazyHealthCheckStats::LazyStatus::PotentialFailure;
+        consecutiveSuccessfulChecks = 0;
         /* we update the next check time here because the check might time out,
            and we do not want to send a second check during that time unless
            the timer is actually very short */
@@ -751,7 +752,7 @@ void DownstreamState::updateNextLazyHealthCheck(LazyHealthCheckStats& stats, boo
 
       time_t backOff = d_config.d_lazyHealthCheckMaxBackOff;
       const ExponentialBackOffTimer backOffTimer(d_config.d_lazyHealthCheckMaxBackOff);
-      auto backOffCoeffTmp = backOffTimer.get(failedTests);
+      auto backOffCoeffTmp = backOffTimer.get(failedTests - 1);
       /* backOffCoeffTmp cannot be higher than d_config.d_lazyHealthCheckMaxBackOff */
       const auto backOffCoeff = static_cast<time_t>(backOffCoeffTmp);
       if ((std::numeric_limits<time_t>::max() / d_config.d_lazyHealthCheckFailedInterval) >= backOffCoeff) {
@@ -800,12 +801,12 @@ void DownstreamState::submitHealthCheckResult(bool initial, bool newResult)
   if (newResult) {
     /* check succeeded */
     currentCheckFailures = 0;
+    consecutiveSuccessfulChecks++;
 
     if (!upStatus) {
       /* we were previously marked as "down" and had a successful health-check,
          let's see if this is enough to move to the "up" state or if we need
          more successful health-checks for that */
-      consecutiveSuccessfulChecks++;
       if (consecutiveSuccessfulChecks < d_config.minRiseSuccesses) {
         /* we need more than one successful check to rise
            and we didn't reach the threshold yet, let's stay down */
@@ -846,7 +847,7 @@ void DownstreamState::submitHealthCheckResult(bool initial, bool newResult)
         auto stats = d_lazyHealthCheckStats.lock();
         vinfolog("Backend %s failed its health-check, moving from Potential failure to Failed", getNameWithAddr());
         stats->d_status = LazyHealthCheckStats::LazyStatus::Failed;
-        currentCheckFailures = 0;
+        currentCheckFailures = 1;
         updateNextLazyHealthCheck(*stats, false);
       }
     }
--- a/pdns/dnsdistdist/dnsdist.cc
+++ b/pdns/dnsdistdist/dnsdist.cc
@@ -3405,6 +3405,7 @@ int main(int argc, char** argv)
           }
 
           if (!queueHealthCheck(mplexer, dss, true)) {
+            dss->submitHealthCheckResult(true, false);
             dss->setUpStatus(false);
             warnlog("Marking downstream %s as 'down'", dss->getNameWithAddr());
           }
--- a/pdns/dnsdistdist/dnsdist.cc
+++ b/pdns/dnsdistdist/dnsdist.cc
@@ -1260,8 +1260,11 @@ ssize_t udpClientSendRequestToBackend(const std::shared_ptr<DownstreamState>& ba
        We don't want to reconnect the real socket if the healthcheck failed,
        because it's not using the same socket.
     */
-    if (!healthCheck && (savederrno == EINVAL || savederrno == ENODEV || savederrno == ENETUNREACH || savederrno == EBADF)) {
-      backend->reconnect();
+    if (!healthCheck) {
+      if (savederrno == EINVAL || savederrno == ENODEV || savederrno == ENETUNREACH || savederrno == EBADF) {
+        backend->reconnect();
+      }
+      backend->reportTimeoutOrError();
     }
   }
 
--- a/pdns/dnsdistdist/test-dnsdistbackend_cc.cc
+++ b/pdns/dnsdistdist/test-dnsdistbackend_cc.cc
@@ -264,8 +264,8 @@ BOOST_AUTO_TEST_CASE(test_LazyExponentialBackOff)
   BOOST_CHECK_EQUAL(ds.getStatus(), "down");
   BOOST_CHECK_EQUAL(ds.healthCheckRequired(currentTime), false);
   /* and the wait time between two checks will double every time a failure occurs */
-  BOOST_CHECK_EQUAL(ds.getNextLazyHealthCheck(), (currentTime + (config.d_lazyHealthCheckFailedInterval * std::pow(2U, ds.currentCheckFailures))));
-  BOOST_CHECK_EQUAL(ds.currentCheckFailures, 0U);
+  BOOST_CHECK_EQUAL(ds.getNextLazyHealthCheck(), (currentTime + (config.d_lazyHealthCheckFailedInterval * std::pow(2U, ds.currentCheckFailures - 1))));
+  BOOST_CHECK_EQUAL(ds.currentCheckFailures, 1U);
 
   /* so after 5 failures */
   const size_t nbFailures = 5;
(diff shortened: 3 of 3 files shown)


omoerbeek:
I did a test with the PR. It does not sem to cover all cases. If i run this on my laptop:

```
setStructuredLogging(true, {timeFormat="ISO8601"})
controlSocket("127.0.0.1:5300")
setKey("redacted")
setLocal('127.0.0.1:5302')
newServer({address='2620:fe::fe', healthCheckMode="lazy"})
setVerbose(true)
setVerboseHealthChecks(true)

```
 and disable WiFi after starting dnsdist, I see failures querying dnsdist as expected:
```
prio="Debug" msg="Got query for www.apple.com|A from 127.0.0.1:62240, relayed to [2620:fe::fe]:53"
prio="Debug" msg="Error sending request to backend [2620:fe::fe]:53: No route to host"
```
but the lazy health checks never kicks in again, even after trying hundreds of queries and the backend remains in "up" status.

rgacogne:
Right, it's a different case (error from `sendmsg` as compared to timeout/servfail/...) that it usually not handled by the health-check but by the re-connection mechanism. Unfortunately it looks like we are handling `ENETUNREACH` but not ` EHOSTUNREACH`, I guess we should?

omoerbeek:
Adding 
```
diff --git a/pdns/dnsdist.cc b/pdns/dnsdist.cc
index 3c32201ab..e6755b913 100644
--- a/pdns/dnsdist.cc
+++ b/pdns/dnsdist.cc
@@ -1262,7 +1262,7 @@ ssize_t udpClientSendRequestToBackend(const std::shared_ptr<DownstreamState>& ss
        We don't want to reconnect the real socket if the healthcheck failed,
        because it's not using the same socket.
     */
-    if (!healthCheck && (savederrno == EINVAL || savederrno == ENODEV || savederrno == ENETUNREACH || savederrno == EBADF)) {
+    if (!healthCheck && (savederrno == EINVAL || savederrno == ENODEV || savederrno == ENETUNREACH || savederrno == EHOSTUNREACH || savederrno == EBADF)) {
       ss->reconnect();
     }
   }

```
is not enough, and no sign of the heathcheck, and even after I reconnect the WifI I still get:
```
prio="Debug" msg="Got query for www.apple.com|A from 127.0.0.1:53996, relayed to [2620:fe::fe]:53"
prio="Debug" msg="Error sending request to backend [2620:fe::fe]:53: Bad file descriptor"
prio="Info" msg="Error connecting to new server with address [2620:fe::fe]:53: connecting socket to [2620:fe::fe]:53: No route to host"
```
Note that some macOS specific issue may be at play here.

rgacogne:
Hmm, so the service is reachable on `[2620:fe::fe]:53` but dnsdist cannot connect to it? I do agree that this particular seems like a good fit for scheduling health-check instead of attempting to reconnect, but that's a bigger change than the one I intended for this PR. I'll think about it.

omoerbeek:
Sorry, my test was wrong, after I connected the Wifi again (but now properly) things start to work again. But the fact remains the lazy healthcheck does not get activated.

rgacogne:
> Sorry, my test was wrong, after I connected the Wifi again (but now properly) things start to work again. But the fact remains the lazy healthcheck does not get activated after I disable Wifi and do a lot of queries.

I tried to reproduce this and I don't understand why it does not activate the health-checks in your case. In my case, there is a UDP timeout detected every time `sendmsg` fails, which does trigger the lazy health-check mechanism:

```
Error sending request to backend 192.0.2.1:9999: Network is unreachable
Had a downstream timeout from 192.0.2.1:9999 () for query for sigsegv.fr|A from 127.0.0.1:48129
Backend 192.0.2.1:9999 reached the lazy health-check threshold (100% out of 20%, looking at sample of 1 items with 1 failures), moving to Potential Failure state
Backend 192.0.2.1:9999 is in potential failure state, next check in 1 seconds
Sending health-check query for 192.0.2.1:9999 which is still in the Potential Failure state
Got query for sigsegv.fr|A from 127.0.0.1:32915, relayed to 192.0.2.1:9999
Error sending request to backend 192.0.2.1:9999: Network is unreachable
Backend 192.0.2.1:9999 failed its health-check, moving from Potential failure to Failed
Backend 192.0.2.1:9999 is in failed state and has failed 1 consecutive checks, next check in 5 seconds
Marking downstream 192.0.2.1:9999 as 'down'
```

rgacogne:
Rebased on master to fix a conflict (`dnsdist.cc` has been moved to `pdns/dnsdistdist/`).

omoerbeek:
Redid the test with the same result. One difference is that after I disable WiFi, the next query immediately gets an error on `sendmsg`, while in your log it shows a timeout which results in the lazy healthcheck to be activated.

```
rio="Warning" msg="You can increase this value by using ulimit."
prio="Info" msg="Marking downstream [2620:fe::fe]:53 as 'up'"
> prio="Warning" msg="Response code 'Non-Existent domain' received from the secpoll stub resolver 192.168.178.6 for 'dnsdist-0.0.28112.0.ddistmovefiles3.g7678ac23d9.security-status.secpoll.powerdns.com.'"
prio="Info" msg="Error while retrieving the security update for version dnsdist-0.0.28112.0.ddistmovefiles3.g7678ac23d9: Unable to get a valid Security Status update"
prio="Info" msg="Not validating response for security status update, this is a non-release version."
prio="Debug" msg="Got query for www.apple.com|A from 127.0.0.1:53108, relayed to [2620:fe::fe]:53"
prio="Debug" msg="Got answer from [2620:fe::fe]:53, relayed to 127.0.0.1:53108 (UDP), took 27364 us"

here I disable WiFi

prio="Debug" msg="Got query for www.apple.com|A from 127.0.0.1:52003, relayed to [2620:fe::fe]:53"
prio="Debug" msg="Error sending request to backend [2620:fe::fe]:53: No route to host"
prio="Info" msg="Error connecting to new server with address [2620:fe::fe]:53: connecting socket to [2620:fe::fe]:53: No route to host"
```
I still have the patch for the EHOSTUNRAECH case in my tree.

rgacogne:
I thought about it and I'm not convinced we should reconnect when we get `EHOSTUNREACH`, which indicates that we still know how to reach the subnet of the destination but that we cannot reach this specific host. It's a bit different for `ENETUNREACH` where the network topology has clearly changed.

omoerbeek:
An update: I was not able to reproduce on Linux, but I am able to on OpenBSD and macOS.

rgacogne:
Looking at the code we are supposed to clean the UDP state if the query could not be properly sent, so the Linux behaviour might actually be the wrong one. I'll have a look.

rgacogne:
According to my tests the discrepancy comes from the bug fixed by https://github.com/PowerDNS/pdns/pull/13914, so the UDP timeout was triggered because we failed to properly detect that the query could not be sent (because an exception was raised), and thus never cleared the corresponding state.
I'll push a fix so that the failure to send a query is correctly accounted toward the health-check trigger.

rgacogne:
> I'll push a fix so that the failure to send a query is correctly accounted toward the health-check trigger.

Done!

omoerbeek:
Much better, the health checks gets activated. There remains a failure case., related to EHOSTUNREACH
```
Start with all OK:

prio="Debug" msg="Got query for www.apple.com|A from 127.0.0.1:61119, relayed to [2620:fe::fe]:53"
prio="Debug" msg="Got answer from [2620:fe::fe]:53, relayed to 127.0.0.1:61119 (UDP), took 14572 us"

After disabling Wifi

prio="Debug" msg="Got query for www.apple.com|A from 127.0.0.1:60208, relayed to [2620:fe::fe]:53"
prio="Debug" msg="Error sending request to backend [2620:fe::fe]:53: No route to host"
prio="Debug" msg="Backend [2620:fe::fe]:53 reached the lazy health-check threshold (50% out of 20%, looking at sample of 2 items with 1 failures), moving to Potential Failure state"
prio="Debug" msg="Backend [2620:fe::fe]:53 is in potential failure state, next check in 30 seconds"
prio="Info" msg="Error checking the health of backend [2620:fe::fe]:53: connecting to [2620:fe::fe]:53: No route to host"
prio="Debug" msg="Backend [2620:fe::fe]:53 failed its health-check, moving from Potential failure to Failed"
prio="Debug" msg="Backend [2620:fe::fe]:53 is in failed state, next check in 30 seconds"
prio="Info" msg="Marking downstream [2620:fe::fe]:53 as 'down'"

Try a few queries

prio="Debug" msg="Dropped query for www.apple.com|A from 127.0.0.1:61543, no downstream server available"
prio="Debug" msg="Dropped query for www.apple.com|A from 127.0.0.1:60612, no downstream server available"
prio="Debug" msg="Dropped query for www.apple.com|A from 127.0.0.1:62639, no downstream server available"
prio="Debug" msg="Dropped query for www.apple.com|A from 127.0.0.1:57717, no downstream server available"
prio="Debug" msg="Dropped query for www.apple.com|A from 127.0.0.1:58328, no downstream server available"
prio="Debug" msg="Dropped query for www.apple.com|A from 127.0.0.1:62021, no downstream server available"
prio="Debug" msg="Dropped query for www.apple.com|A from 127.0.0.1:57084, no downstream server available"
prio="Debug" msg="Dropped query for www.apple.com|A from 127.0.0.1:63761, no downstream server available"
prio="Debug" msg="Sending health-check query for [2620:fe::fe]:53 which is still in the Failed state"
prio="Debug" msg="Backend [2620:fe::fe]:53 is in failed state, next check in 30 seconds"

enabling Wifi again, v6 works using `ping6 2620:fe::fe`

prio="Debug" msg="Backend [2620:fe::fe]:53 had 1 successful checks, moving to Healthy"
prio="Info" msg="Marking downstream [2620:fe::fe]:53 as 'up'"
prio="Debug" msg="Got query for www.apple.com|A from 127.0.0.1:50818, relayed to [2620:fe::fe]:53"

But still `EHOSTUNREACH` and no reconnect, while the health check succeeds

prio="Debug" msg="Error sending request to backend [2620:fe::fe]:53: No route to host"
prio="Debug" msg="Backend [2620:fe::fe]:53 reached the lazy health-check threshold (100% out of 20%, looking at sample of 1 items with 1 failures), moving to Potential Failure state"
prio="Debug" msg="Backend [2620:fe::fe]:53 is in potential failure state, next check in 30 seconds"
prio="Debug" msg="Backend [2620:fe::fe]:53 had 1 successful checks, moving to Healthy"
prio="Debug" msg="Got query for www.apple.com|A from 127.0.0.1:49601, relayed to [2620:fe::fe]:53"
prio="Debug" msg="Error sending request to backend [2620:fe::fe]:53: No route to host"
prio="Debug" msg="Backend [2620:fe::fe]:53 reached the lazy health-check threshold (100% out of 20%, looking at sample of 1 items with 1 failures), moving to Potential Failure state"
prio="Debug" msg="Backend [2620:fe::fe]:53 is in potential failure state, next check in 30 seconds"
prio="Debug" msg="Backend [2620:fe::fe]:53 had 1 successful checks, moving to Healthy"
prio="Debug" msg="Got query for www.apple.com|A from 127.0.0.1:49672, relayed to [2620:fe::fe]:53"
prio="Debug" msg="Error sending request to backend [2620:fe::fe]:53: No route to host"
prio="Debug" msg="Backend [2620:fe::fe]:53 reached the lazy health-check threshold (100% out of 20%, looking at sample of 1 items with 1 failures), moving to Potential Failure state"
prio="Debug" msg="Backend [2620:fe::fe]:53 is in potential failure state, next check in 30 seconds"
prio="Debug" msg="Backend [2620:fe::fe]:53 had 1 successful checks, moving to Healthy"
```

So at least for macOS (and I think OpenBSD as well)  `EHOSTUNREACH` can indicate a need to reconnect I think.

rgacogne:
Alright, POSIX is a bit unclear on what `EHOSTUNREACH` actually means: 

> Host is unreachable. The destination host cannot be reached (probably because the host is down or a remote router cannot reach it).

But I can see how it might warrant a re-connection indeed. Would you mind submitting it as a separate PR?

omoerbeek:
Will do

omoerbeek (approved):

PR #13846, opened 2024-03-01
rec: Backport 13830 to rec 5.0.x: reduce RPZ memory usage

omoerbeek:
### Short description

Partial backport of #13830 

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/recursordist/rec-main.cc
+++ b/pdns/recursordist/rec-main.cc
@@ -2750,8 +2750,7 @@ static void recursorThread()
       }
     }
 
-    unsigned int ringsize = ::arg().asNum("stats-ringbuffer-entries") / RecThreadInfo::numUDPWorkers();
-    if (ringsize != 0) {
+    if (unsigned int ringsize = ::arg().asNum("stats-ringbuffer-entries") / RecThreadInfo::numUDPWorkers(); ringsize != 0) {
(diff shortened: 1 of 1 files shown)


pdns/recursordist/rec-main.cc
```diff
@@ -2750,8 +2750,7 @@ static void recursorThread()
       }
     }
 
-    unsigned int ringsize = ::arg().asNum("stats-ringbuffer-entries") / RecThreadInfo::numUDPWorkers();
-    if (ringsize != 0) {
+    if (unsigned int ringsize = ::arg().asNum("stats-ringbuffer-entries") / RecThreadInfo::numUDPWorkers(); ringsize != 0) {
```
Habbie, line 2753: I don't like this syntax, but given this is a backport, I'm too late :)

Habbie (approved):

pdns/recursordist/rec-main.cc
omoerbeek: Oh, I like the combination of restricted score and condition very much.

Habbie: oh, I like that too. I just don't like how it looks. I do see how another layer of `{}` is not pretty either, but this form makes me think too hard ;)

Issue #13817, opened 2024-02-21
rec: when recursive forwarding, records in AUTHORITY section are missing in some cases


omoerbeek:
- [X This is not a support question, I have read [about opensource](https://www.powerdns.com/opensource.html) and will send support questions to the IRC channel, [Github Discussions](https://github.com/PowerDNS/pdns/discussions/) or the mailing list.
- [X] I have read and understood the ['out in the open' support policy](https://blog.powerdns.com/2016/01/18/open-source-support-out-in-the-open/)

 - Program: Recursor
 - Issue type: Bug report

### Short description

When doing recursive forwarding, a CNAME target that matches wildcard does not cause NSEC records and their RRSIGS to be added to the AUTHORITY SECTION.

Reproducer currently at `test.drijf.net`. Query using quad1:
```
$ dig @1.1.1.1 test.drijf.net +dnssec 

; <<>> DiG 9.18.24 <<>> @1.1.1.1 test.drijf.net +dnssec
; (1 server found)
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 27488
;; flags: qr rd ra ad; QUERY: 1, ANSWER: 4, AUTHORITY: 2, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags: do; udp: 1232
;; QUESTION SECTION:
;test.drijf.net.                        IN      A

;; ANSWER SECTION:
test.drijf.net.         60      IN      CNAME   r.sub.m6k.nl.
test.drijf.net.         60      IN      RRSIG   CNAME 13 3 60 20240229000000 20240208000000 39531 drijf.net. QVMKGQFKRswf9KLHzRnPEtYkv/7F9M/IvtukVJMY2BteO/YseCrfTLH6 DrsmOZBUeN0KFPF/czCKrpVA1tbsgw==
r.sub.m6k.nl.           60      IN      A       1.2.3.4
r.sub.m6k.nl.           60      IN      RRSIG   A 13 3 60 20240229000000 20240208000000 53536 m6k.nl. fp7jqKSVbRDTkm/IXXGwlBkdMS7YWUCphe6VCvGKQp66AwkoL4cL9fB1 t5RFJ+KAhKTg6uZfduvL17Nwd8eA3g==

;; AUTHORITY SECTION:
*.sub.m6k.nl.           60      IN      NSEC    test.m6k.nl. A RRSIG NSEC
[...]

PR #13818, opened 2024-02-21
rec: add AUTHORITY section records received for the recursive forwarding case

omoerbeek:
Tentative fix of #13817

This needs test and a thorough review.

### Checklist

I have:
- [X] read the [CONTRIBUTING.md](https://github.com/PowerDNS/pdns/blob/master/CONTRIBUTING.md) document
- [X] compiled this code
- [X] tested this code
- [ ] included documentation (including possible behaviour changes)
- [ ] documented the code
- [ ] added or modified regression test(s)
- [ ] added or modified unit test(s)
- [ ]  checked that this code was merged to master

With this PR the testcase of the Issue becomes:

```
$ mydig test.drijf.net +dnssec

; <<>> DiG 9.18.24 <<>> +retry -p 5301 @127.0.0.1 test.drijf.net +dnssec
; (1 server found)
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 42215
;; flags: qr rd ra ad; QUERY: 1, ANSWER: 4, AUTHORITY: 2, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags: do; udp: 512
;; QUESTION SECTION:
;test.drijf.net.                        IN      A

;; ANSWER SECTION:
test.drijf.net.         60      IN      CNAME   r.sub.m6k.nl.
r.sub.m6k.nl.           60      IN      A       1.2.3.4
test.drijf.net.         60      IN      RRSIG   CNAME 13 3 60 20240229000000 20240208000000 39531 drijf.net. QVMKGQFKRswf9KLHzRnPEtYkv/7F9M/IvtukVJMY2BteO/YseCrfTLH6 DrsmOZBUeN0KFPF/czCKrpVA1tbsgw==
r.sub.m6k.nl.           60      IN      RRSIG   A 13 3 60 20240229000000 20240208000000 53536 m6k.nl. fp7jqKSVbRDTkm/IXXGwlBkdMS7YWUCphe6VCvGKQp66AwkoL4cL9fB1 t5RFJ+KAhKTg6uZfduvL17Nwd8eA3g==

;; AUTHORITY SECTION:
*.sub.m6k.nl.           60      IN      NSEC    test.m6k.nl. A RRSIG NSEC
[...]

PR #13847, opened 2024-03-01
rec: Fix gathering of denial of existence proof for wildcard-expanded names

rgacogne:
### Short description

When the recursor is forwarding to a resolver, we accept the names composing the CNAME chain starting at the queried name. This means we also need to gather the denial of existence proof for CNAMEs that were expanded from a wildcard, otherwise the response sent to the client cannot be DNSSEC-validated.

Fixes https://github.com/PowerDNS/pdns/issues/13817 and supersedes https://github.com/PowerDNS/pdns/pull/13818

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/recursordist/syncres.cc
+++ b/pdns/recursordist/syncres.cc
@@ -4353,10 +4353,36 @@ RCode::rcodes_ SyncRes::updateCacheFromRecords(unsigned int depth, const string&
   sanitizeRecords(prefix, lwr, qname, qtype, auth, wasForwarded, rdQuery);
 
   std::vector<std::shared_ptr<DNSRecord>> authorityRecs;
-  const unsigned int labelCount = qname.countLabels();
   bool isCNAMEAnswer = false;
   bool isDNAMEAnswer = false;
   DNSName seenAuth;
+
+  // names that might be expanded from a wildcard, and thus require denial of existence proof
+  // this is the queried name and any part of the CNAME chain from the queried name
+  // the key is the name itself, the value is initially false and is set to true once we have
+  // confirmed it was actually expanded from a wildcard
+  std::map<DNSName, bool> wildcardCandidates{{qname, false}};
+
+  if (rdQuery) {
+    std::unordered_map<DNSName, DNSName> cnames;
+    for (const auto& rec : lwr.d_records) {
+      if (rec.d_type != QType::CNAME || rec.d_class != QClass::IN) {
+        continue;
+      }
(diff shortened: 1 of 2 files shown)


pdns/recursordist/syncres.cc
```diff
@@ -4377,14 +4403,16 @@ RCode::rcodes_ SyncRes::updateCacheFromRecords(unsigned int depth, const string&
...
     if (rec.d_type == QType::RRSIG) {
       auto rrsig = getRR<RRSIGRecordContent>(rec);
       if (rrsig) {
         /* As illustrated in rfc4035's Appendix B.6, the RRSIG label
            count can be lower than the name's label count if it was
            synthesized from the wildcard. Note that the difference might
            be > 1. */
-        if (rec.d_name == qname && isWildcardExpanded(labelCount, *rrsig)) {
+        if (wildcardCandidates.count(rec.d_name) > 0 && isWildcardExpanded(labelCount, *rrsig)) {
+          wildcardCandidates[rec.d_name] = true;
```
omoerbeek, line 4415: Avoid a double lookup:
```suggestion
        if (auto wcIt = wildcardCandidates.find(rec.d_name); wcIt != wildcardCandidates.end() && isWildcardExpanded(labelCount, *rrsig)) {
          wcIt->second = true;
```

pdns/recursordist/syncres.cc
```diff
@@ -4705,6 +4740,13 @@ RCode::rcodes_ SyncRes::updateCacheFromRecords(unsigned int depth, const string&
     }
   }
 
+  if (gatherWildcardProof) {
+    if (auto wcIt = wildcardCandidates.find(qname); wcIt != wildcardCandidates.end() && !wcIt->second) {
+      // the queried name was not expended from a wildcard, a record in the CNAME chain was, so we don't need to gather wildcard proof now: we will do that when looking up the CNAME chain
```
omoerbeek, line 4745:
```suggestion
      // the queried name was not expanded from a wildcard, a record in the CNAME chain was, so we don't need to gather wildcard proof now: we will do that when looking up the CNAME chain
```

pdns/recursordist/syncres.cc
rgacogne: Done, thanks!

pdns/recursordist/syncres.cc
rgacogne: Applied, thanks!

omoerbeek (approved):

Issue #13838, opened 2024-02-27
Authserver, Recursor and dnsdist containers logging credentials on startup


maresac:
- [x] This is not a support question, I have read [about opensource](https://www.powerdns.com/opensource.html) and will send support questions to the IRC channel, [Github Discussions](https://github.com/PowerDNS/pdns/discussions/) or the mailing list.
- [x] I have read and understood the ['out in the open' support policy](https://blog.powerdns.com/2016/01/18/open-source-support-out-in-the-open/)

 - Program: Authoritative, Recursor, dnsdist 
 - Issue type: Bug report

### Short description

Each of the components (using the official Docker images) print their whole configuration on startup, including plain text credentials such as the built-in webserver API key and password or the auth backend db password. Anyone with read-access to container logs can use these credentials to access the backend database and/or use the components' APIs exposed through the built-in webservers. 

### Environment

 This affects container images for the Auth Server, Recursor and dnsdist from https://hub.docker.com/u/powerdns

### Steps to reproduce

Run components with official Docker images, providing API keys and other credentials through environment variables.

### Expected behaviour

In production environments, credentials should not be logged unmasked.

### Actual behaviour

Plain text credentials are accessible in container logs. The whole configs are printed by the Python script used on startup:
* https://github.com/PowerDNS/pdns/blob/master/dockerdata/startup.py#L54
* https://github.com/PowerDNS/pdns/blob/master/dockerdata/startup.py#L66

### Other information

I would propose to either redact the credentials when printing the config or remove the print statements altogether. I can submit a PR if you let me know what you prefer :)

Habbie:
Thanks for the report, that's an excellent point. I'd love a PR to only do the print if some debugging env var is set.

Habbie:
(if you feel like doing a PR that can do redaction, that would be even better, but it's a harder job, because that script doesn't know about all secret variables, just some of them)

maresac:
Great, I'll see what I can come up with. PR will follow soon.

Habbie:
fixed in #13849, thanks!

PR #13849, opened 2024-03-01
Docker: Only print config if debug flag is set

maresac:
### Short description

This changes the default behavior of Auth, Recursor and dnsdist containers to only print their configuration on startup if the environment variable `DEBUG_CONFIG` is set to `'yes'`. Per default, the env variable is not set.

I also added an explanation to the Docker README and additionally adjusted parts that seemed to be outdated.

Fix for #13838

### Checklist

I have:
- [ ] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/Docker-README.md
+++ b/Docker-README.md
@@ -19,15 +19,16 @@ Other data involved in the Docker build process can be found at https://github.c
 
 The images are ready to run with limited functionality.
 At container startup, the startup.py wrapper (from the dockerdata directory linked above) checks for `PDNS_RECURSOR_API_KEY` / `PDNS_AUTH_API_KEY` / `DNSDIST_API_KEY` environment variables for the product you are running.
-If such a variable is found, `/etc/powerdns-api.conf` or `/etc/dnsdist-api.conf` is written, enabling the webserver in all products, and the dnsdist console.
+If such a variable is found, `/etc/powerdns/recursor.d/_api.conf` / `/etc/powerdns/pdns.d/_api.conf` / `/etc/dnsdist/conf.d/_api.conf` is written, enabling the webserver in all products, and the dnsdist console.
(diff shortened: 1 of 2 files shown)


Docker-README.md
```diff
@@ -19,15 +19,16 @@ Other data involved in the Docker build process can be found at https://github.c
...
 
 The default configs shipped in the image (see dockerdata above) parse all files in `/etc/powerdns/pdns.d` / `/etc/powerdns/recursor.d` / `/etc/dnsdist/conf.d`.
-The image also ships a symlink to the API config file inside those `.d` dirs.
 For Auth and Recursor, extra configuration can be passed on the command line, or via a volume mount into `/etc/powerdns` or the `.d` dir.
 For dnsdist, only the volume mount is applicable.
 
-If you want to volume mount a config, but also take the keys from the environment, please take care to include the same `X-api.conf` symlink in your `.d` directory.
+If you want to volume mount a config, but also take the keys from the environment, please take care to include the same `_api.conf` file in your `.d` directory.
+
+If you want to read the configuration for debugging purposes, you can run the containers with the `DEBUG_CONFIG` environment variable set to `'yes'`). This will print the full config on startup. Please keep in mind that this also includes credentials, therefore this setting should never be used in production environments.
```
Habbie, line 31:
```suggestion
If you want to read the configuration for debugging purposes, you can run the containers with the `DEBUG_CONFIG` environment variable set to `'yes'`.
This will print the full config on startup. Please keep in mind that this also includes credentials, therefore this setting should never be used in production environments.
```

(removed `)`, put a newline between sentences, removed trailing white space)

Habbie (approved):
Thanks!

omoerbeek (approved):

PR #13863, opened 2024-03-06
auth: Informative logging from error/warn to notice/info

nvaatstra:
### Short description
Auth 4.9 parameter `loglevel-show` exposes some informative messages are logged as `Warning` or `Error`. This PR lowers the level of several informative messages in an attempt to keep `Warning` and `Error` levels for events that users should act upon.

Note: This is not an exhaustive audit of all logging statements, only a specific few which I ran into as a user.
Note: The default `loglevel` is `Warning`, so in a deployment with default logging these messages would be discarded as a result of this PR

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/auth-main.cc
+++ b/pdns/auth-main.cc
@@ -1067,7 +1067,7 @@ static int guardian(int argc, char** argv)
       }
       newargv[n] = nullptr;
 
-      g_log << Logger::Error << "Guardian is launching an instance" << endl;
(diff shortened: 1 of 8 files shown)


mind04:
Can we please not do this without rebalancing all log messages. This pull will force people to over verbose logging for daily operation.

nvaatstra:
Closing, choice made to not do this incrementally.

PR #13866, opened 2024-03-06
rec: Don't enter wildcard qnames into the cache in the ZoneToCache function

omoerbeek:
In the record cache, a wildcard has no meaning, it will only match a literal '*'. No need for backport AFAIKS, as it does not prevent queries matching a wildcard in the auth to be processed by rec (either the query is sent to the auth if no match was found in the record cache, or it is matched by the record cache against a previously received record from the auth).

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/recursordist/rec-zonetocache.cc
+++ b/pdns/recursordist/rec-zonetocache.cc
@@ -403,6 +403,9 @@ void ZoneData::ZoneToCache(const RecZoneToCache::Config& config)
   d_now = time(nullptr);
   for (const auto& [key, v] : d_all) {
     const auto& [qname, qtype] = key;
+    if (qname.isWildcard()) {
+      continue;
+    }
     switch (qtype) {
     case QType::NSEC:
     case QType::NSEC3:


rgacogne:
If I'm not mistaken the wildcard entries are useful for the aggressive NSEC cache, though?

omoerbeek:
> If I'm not mistaken the wildcard entries are useful for the aggressive NSEC cache, though?

Yes, but the zoneToCache function does not enter anything into the aggressive cache atm. One day it might, and then wildcards records play a role indeed.

rgacogne:
> > If I'm not mistaken the wildcard entries are useful for the aggressive NSEC cache, though?
> 
> Yes, but the zoneToCache function does not enter anything into the aggressive cache atm. One day it might, and then wildcards records play a role indeed.

Ah, okay, I see your point: the aggressive cache can pull wildcard entries from the records cache but only if the corresponding NSEC(3) records are already in the aggressive cache.

PR #13879, opened 2024-03-08
auth LUA: support returning empty set in filterForward

### LINKED ISSUE #12436: Lua Records: Support of empty answers on filterForward()
- Program: Authoritative
 - Issue type: Feature request

### Short description

Return empty answers from  [`filterForward()`](https://doc.powerdns.com/authoritative/lua-records/functions.html#filterForward) for addresses in the `NetmaskGroup` argument, instead of `0.0.0.0` or `::0`.

### Usecase

We'd like to use [`createForward()`](https://doc.powerdns.com/authoritative/lua-records/functions.html#createForward) in conjunction with [`filterForward()`](https://doc.powerdns.com/authoritative/lua-records/functions.html#filterForward). Up to now we have answered queries for addresses outside our networks with an empty answer. [`filterForward()`](https://doc.powerdns.com/authoritative/lua-records/functions.html#filterForward) responds with `0.0.0.0` or `::0` by default in this case. Since we do not want to change the usual behavior of our DNS servers, it would be nice if we could continue do so. 

### Description

In addition, [`filterForward()`](https://doc.powerdns.com/authoritative/lua-records/functions.html#filterForward) should accept `''` (empty string) as the `fallback` argument and it should respond with a empty answer in this case.

Habbie:
### Short description
fixes #12436

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/pdns/lua-record.cc
+++ b/pdns/lua-record.cc
@@ -1057,24 +1057,24 @@ static void setupLuaRecords(LuaContext& lua) // NOLINT(readability-function-cogn
       return std::string("unknown");
     });
 
-  lua.writeFunction("filterForward", [](string address, NetmaskGroup& nmg, boost::optional<string> fallback) {
+  lua.writeFunction("filterForward", [](const string& address, NetmaskGroup& nmg, boost::optional<string> fallback) -> vector<string> {
       ComboAddress ca(address);
 
       if (nmg.match(ComboAddress(address))) {
-        return vector<string>{address};
+        return {address};
       } else {
         if (fallback) {
           if (fallback->empty()) {
             // if fallback is an empty string, return an empty array
-            return vector<string>{};
+            return {};
           }
-          return vector<string>{*fallback};
+          return {*fallback};
         }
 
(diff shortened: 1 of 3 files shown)


Habbie:
amended for tidy

pdns/lua-record.cc
```diff
@@ -1061,16 +1061,20 @@ static void setupLuaRecords(LuaContext& lua) // NOLINT(readability-function-cogn
       ComboAddress ca(address);
```
omoerbeek, line 1061: If you add an explicit return type (` -> vector<string>`, you can get rid off all the `vector<string>` in the body, i.e. use `return {address}` or `return {}`.

omoerbeek (approved):
Looks good, one suggestion you can take or leave.

pdns/lua-record.cc
Habbie: oh i love it

Habbie: done

PR #13921, opened 2024-03-15
Rec: add feature to allow names (resolved by system resolver) in forwarding config

### LINKED ISSUE #11393: Recursor must allow hostnames in forward-zone to be Docker / Docker Swarm compatible

omoerbeek:
### Short description

Fixes #11393

- Must be enabed by setting the `system resolver ttl` setting.
- Can only be used for forwarding atm.
- This settings control the period of the thread checking if names did change, a reasonable value is 60. If that happens forwarding config is reloaded.
- More details in the comments in `rec-system-resolve.hh`

Both feedback on the general design as the specific implementation welcome!

After the regression test has been reviewed (especially the sudo parts!) we should add ENABLE_SUDO_TESTS to the test run in `tasks.py`, like `dnsdist` does.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [X] documented the code


pdns/recursordist/rec-system-resolve.cc
```diff
@@ -0,0 +1,292 @@
...
+    const time_t startTime = time(nullptr);
+    time_t wakeTime = startTime;
+    while (wakeTime - startTime < d_interval) {
+      std::unique_lock lock(mutex);
+      time_t remaining = d_interval - (wakeTime - startTime);
+      if (remaining <= 0) {
+        break;
+      }
+      condVar.wait_for(lock, std::chrono::seconds(remaining),
+                       [&wakeup = wakeup] { return wakeup.load(); });
```
fredmorcos, line 249:
```suggestion
                       [&doWakeup = wakeup] { return doWakeup.load(); });
```

pdns/recursordist/rec-system-resolve.cc
```diff
@@ -0,0 +1,292 @@
...
+      return txt->d_text;
+    }
+  }
+  return {};
+}
+} // anonymous namespace
+
+std::string pdns::RecResolve::s_serverID;
+time_t pdns::RecResolve::s_ttl{0};
+std::function<void()> pdns::RecResolve::s_callback;
```
fredmorcos, line 97:
```suggestion
// RecResolve class members.
std::string pdns::RecResolve::s_serverID;
time_t pdns::RecResolve::s_ttl{0};
std::function<void()> pdns::RecResolve::s_callback;
```

pdns/recursordist/rec-system-resolve.cc
```diff
@@ -0,0 +1,292 @@
...
+
+void pdns::RecResolve::Refresher::refreshLoop()
+{
+  setThreadName("rec/sysres");
+  time_t lastSelfCheck = 0;
+
+  while (!stop) {
+    const time_t startTime = time(nullptr);
+    time_t wakeTime = startTime;
+    while (wakeTime - startTime < d_interval) {
```
fredmorcos, line 242: This looks strange to me. Wouldn't `wakeTime - startTime` be zero?

pdns/recursordist/settings/rust/Cargo.toml
```diff
@@ -13,6 +13,7 @@ serde = { version = "1.0", features = ["derive"] }
 serde_yaml = "0.9"
 ipnet = "2.8"
 once_cell = "1.18.0"
+hostname-validator = "1.1.1" # This is temporary. PR 13819 has the infra to call C++ from, so we can arrange for DNSName::ishostname() to be called instead of importing another crate aftyer that one is merged.
```
fredmorcos, line 16:
```suggestion
hostname-validator = "1.1.1" # This is temporary. PR 13819 has the infra to call C++ from, so we can arrange for DNSName::ishostname() to be called instead of importing another crate after that one is merged.
```

fredmorcos (changes requested):

pdns/recursordist/rec-system-resolve.cc
```diff
@@ -0,0 +1,292 @@
...
+ComboAddress resolve(const std::string& name)
+{
+  struct addrinfo hints = {};
+  hints.ai_flags = AI_ADDRCONFIG;
+  hints.ai_family = 0;
+
+  struct addrinfo* res = nullptr;
+  auto ret = getaddrinfo(name.c_str(), nullptr, &hints, &res);
+  // We pick the first address returned for now.
+  // XXX This might trigger unwanted "changes detected "if the sort order varies
```
rgacogne, line 47: We could convert all returned addresses to `ComboAddress` and sort them, I guess?

pdns/recursordist/rec-system-resolve.cc
```diff
@@ -0,0 +1,292 @@
...
+  if (ret == 0) {
+    auto address = ComboAddress{res->ai_addr, res->ai_addrlen};
```
rgacogne, line 49: It looks like we will leak `res` if `ComboAddress`'s ctor throws (unlikely, granted).

pdns/recursordist/rec-system-resolve.hh
```diff
@@ -0,0 +1,150 @@
...
+  static RecResolve& getInstance();
+
+  RecResolve(time_t ttl = 60, const std::function<void()>& callback = nullptr);
+  ~RecResolve();
+  // Lookup a name and register it in the names to be checked if not already there
+  ComboAddress lookupAndRegister(const std::string& name, time_t now);
+  // Lookup a name which must be already registered
+  ComboAddress lookup(const std::string& name);
+
+  // When an instance is created, it will runn a refresh thread, stop it wit this method
```
rgacogne, line 90:
```suggestion
  // When an instance is created, it will run a refresh thread, stop it with this method
```

pdns/recursordist/rec-system-resolve.hh
```diff
@@ -0,0 +1,150 @@
...
+  void stopRefresher();
+  // And restart it again
+  void startRefresher();
+  // Wipe one or all names
+  void wipe(const std::string& name = "");
+  // Did we se a cahnage? Calling this functino will reset the flag.
```
rgacogne, line 96:
```suggestion
  // Did we see a change? Calling this function will reset the flag.
```

pdns/recursordist/reczones.cc
```diff
@@ -62,6 +63,30 @@ bool primeHints(time_t now)
...
+{
+  try {
+    ComboAddress addr = parseIPAndPort(str, defPort);
+    return addr;
+  }
+  catch (const PDNSException&) {
+    uint16_t port = defPort;
+    string::size_type pos = str.rfind(':');
+    if (pos != string::npos) {
+      cerr << str.substr(pos) << endl;
```
rgacogne, line 76: I guess this line should go?

pdns/recursordist/rec-system-resolve.cc
omoerbeek: yes it would initially be zero, the expression `wakeTime - startTime`  is equal to the time slept. I'll add a comment

pdns/recursordist/settings/docs-new-preamble-in.rst
```diff
@@ -168,6 +168,9 @@ An example of a ``forward_zones`` entry, which consists of a sequence of forward
     recurse: true
     notify_allowed: true
 
+Starting with version 5.1.0, names can be used if
+:ref:`setting-yaml-recursor.system_resolver_ttl` is set.
+The names will be resolved using the system resolver and an automatic refresh of the forwarding zones will happend if a name starts resolving to a new address.
```
zeha, line 173:
```suggestion
The names will be resolved using the system resolver and an automatic refresh of the forwarding zones will happen if a name starts resolving to a new address.
```

regression-tests.recursor-dnssec/test_NamedForward.py
```diff
@@ -0,0 +1,80 @@
...
+        query.flags |= dns.flags.AD
+
+        res = self.sendUDPQuery(query)
+
+        self.assertMessageIsAuthenticated(res)
+        self.assertRRsetInAnswer(res, expected)
+        self.assertMatchingRRSIGInAnswer(res, expected)
+
+@unittest.skipUnless('ENABLE_SUDO_TESTS' in os.environ, "sudo is not available")
+class testNamedForwardWithChange(RecursorTest):
```
zeha, line 44: a test changing getaddrinfo behaviour might also make sense?

iirc setting ipv6.disable=1 might do such a thing on linux

omoerbeek: I'll think about this. The just committed code sorts v4 addresses before v6, so if both are available a v4 address is picked. Disabling v6 does not make a difference then.

pdns/recursordist/rec-system-resolve.hh
```diff
@@ -0,0 +1,150 @@
...
+  ComboAddress lookup(const std::string& name);
+
+  // When an instance is created, it will run a refresh thread, stop it with this method
+  void stopRefresher();
+  // And restart it again
+  void startRefresher();
+  // Wipe one or all names
+  void wipe(const std::string& name = "");
+  // Did we see a change? Calling this function will reset the flag.
+  bool changeDetected();
```
rgacogne, line 97: Seems unused?

rgacogne (changes requested):
The new feature looks great and works very well in my tests! I found a few minor nits. The only thing I did not expect at first (but it is clearly stated in the documentation!) is that a change in one of the addresses for a forward zone triggers a reload of the whole forward configuration, re-reading it from the files on disk. I believe it's fine, but perhaps we could emphasize this behaviour in the documentation?

pdns/recursordist/rec-system-resolve.cc
```diff
@@ -0,0 +1,308 @@
...
+}
+
+pdns::RecResolve& pdns::RecResolve::getInstance()
+{
+  static unique_ptr<RecResolve> res = make_unique<pdns::RecResolve>(s_ttl, s_callback);
+  return *res;
+}
+
+pdns::RecResolve::RecResolve(time_t ttl, const std::function<void()>& callback) :
+  d_ttl(ttl), d_refresher(ttl / 6, callback, *this)
```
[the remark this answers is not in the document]
omoerbeek: I just made something up. Will revisit.

pdns/recursordist/settings/docs-new-preamble-in.rst
```diff
@@ -168,6 +168,11 @@ An example of a ``forward_zones`` entry, which consists of a sequence of forward
     recurse: true
     notify_allowed: true
 
+Starting with version 5.1.0, names can be used if
+:ref:`setting-yaml-recursor.system_resolver_ttl` is set.
+The names will be resolved using the system resolver and an automatic refresh of the forwarding zones will happen if a name starts resolving to a new address.
+Te refresh is done by performig the equivalent of ``rec_control reload-zones``.
```
rgacogne, line 174:
```suggestion
The refresh is done by performing the equivalent of ``rec_control reload-zones``.
```

rgacogne (approved):
One nit in the documentation, looks good to me otherwise!

fredmorcos (approved):

Issue #5760, opened 2017-10-02
dnsdist: wishlist for an AllRule() that matches after the packetcache


giganteous:
- Program: dnsdist
 - Issue type: Wishlist

### Short description

While testing recursors with a TeeAction(), I'm getting traffic on the Teed recursor (called B) that is cached by the packetcache for the recursor that lives in the pool.

### Usecase
I'd like to be able to compare the recursor stats of recursor A and recursor B.

### Description
I think TeeAction should be able to hook into the stream somewhat later, after the pool selection process, although I can imagine the current code might not be suited to this?

Anyway, put it on a wishlist for 1.3, but it's up to you :)

Issue #13750, opened 2024-01-30
dnsdist: Cache-Only lookup then custom rcode as dynamic block action


rgacogne:
- Program: dnsdist 
 - Issue type: Feature request

### Short description

It would be nice to be able to set a "cache-lookup" only action for dynamic blocks, where a blocked suffix or client IP still triggers a normal cache lookup, but on a cache miss the query doesn't get forwarded to a backend and is immediately dropped/answered by a self-generated rcode (REFUSED/SERVFAIL/NODATA?) instead.

### Usecase

It would reduce the impact of PRSD for legitimate clients, and reduce the impact of a false positive block as well.

rgacogne:
With the 'set tag from dynamic block' change suggested in #13374 plus a new "cache miss" rules chain, this should be easy to implement. The "cache miss" rule chain might also fix https://github.com/PowerDNS/pdns/issues/5760.

PR #13922, opened 2024-03-15
dnsdist: Add a new query rules chain triggered after a cache miss

rgacogne:
### Short description

This new chain of rules allows postponing the decision of what to do with the query to after a cache-lookup has been done. This is particularly useful when dealing with abuse: we might want to allow cache hits to be processed normally since they are cheap while dropping/refusing/routing to a different pool queries that result in a cache miss.

Closes https://github.com/PowerDNS/pdns/issues/5760 and is the first step to implement https://github.com/PowerDNS/pdns/issues/13750

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code


chbruyand (approved):
lgtm!

Issue #13859, opened 2024-03-04
dnsdist: AXFR Client responses not logged by DnstapLogResponseAction


danielslab:
- [x] This is not a support question, I have read [about opensource](https://www.powerdns.com/opensource.html) and will send support questions to the IRC channel, [Github Discussions](https://github.com/PowerDNS/pdns/discussions/) or the mailing list.
- [x] I have read and understood the ['out in the open' support policy](https://blog.powerdns.com/2016/01/18/open-source-support-out-in-the-open/)

 - Program: dnsdist 
 - Issue type: Bug report

### Short description

AXFR-Client-Response is not logged via DnstapLogResponseAction

### Environment

 - Operating system: debian 12.4
 - Software version: dnsdist 1.9.0
 - Software source: dockerhub powerdns/dnsdist-19

### Steps to reproduce

1. dnsdist.conf:
```
    setLocal('0.0.0.0:10053')
    addLocal('[::]:10053')

    newServer({address='10.10.0.200:53', pool='auth'})

    authdomains_internal = newSuffixMatchNode()
    authdomains_internal:add(newDNSName("box."))

    fsul = newFrameStreamTcpLogger("10.10.0.210:6000")
    tapRule = AllRule()
    addAction(tapRule, DnstapLogAction("query", fsul))
    addResponseAction(tapRule, DnstapLogResponseAction("response", fsul))
    addCacheHitResponseAction(tapRule, DnstapLogResponseAction("cache-response", fsul))

    addAction(SuffixMatchNodeRule(authdomains_internal), PoolAction('auth'))
```
2. Initiate AXFR
`dig  @172.16.0.1 box AXFR`

### Expected behaviour
AXFR-Client-Query and AXFR-Client-Response are logged via Dnstap

### Actual behaviour
Only CLIENT_QUERY are sent via  dnstap

```
2024-03-04T19:07:55.139128265Z query CLIENT_QUERY NOERROR 192.168.178.50 53251 IPv4 TCP 44b box AXFR 0.000000
```

rgacogne:
It happens because {A,I}XFR responses are currently not handled in a special way, bypassing the response rules. I do see how it's problematic for this use-case, but I'm not sure we will be able to change that soon.

PR #13923, opened 2024-03-15
dnsdist: Add a new response chain for XFR responses

rgacogne:
### Short description

Until now it wasn't possible to look at response packets for AXFR or IXFR queries: the fact that there usually is more than one response packet means that DNSdist is handling these responses as a special case, bypassing the normal response rules chains.

Closes https://github.com/PowerDNS/pdns/issues/13859

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code


chbruyand (approved):
lgtm

PR #13926, opened 2024-03-18
rec: fix trace=fail regression

omoerbeek:
### Short description

This was introduced with the trace rewrite. Add a regression test to detect future failure.
While there, modify the ZTC test to not start auths, they are not needed.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/recursordist/pdns_recursor.cc
+++ b/pdns/recursordist/pdns_recursor.cc
@@ -857,6 +857,15 @@ static void dumpTrace(const string& trace, const timeval& timev)
   if (trace.empty()) {
     return;
   }
+  if (t_tracefd < 0) {
+    std::istringstream buf(trace);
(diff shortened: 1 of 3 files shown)


rgacogne (approved):

gertvdijk:
Wow, I just ran into this after upgrading 4.8.x to 5.0.3 and I wasn't able to find any useful clues after hitting this:

```
msg="Could not dup trace file" error="Bad file descriptor" subsystem="trace" level="0" prio="Error" tid="2" ts="1711471117.728"
```

Came here after noticing this was fixed on master, thanks for having this handled already!
Would appreciate a backport to stable; it's quite useful to debug lookup failures.

omoerbeek:
It's marked for backporting, so you can assume the next minor releases will have it. 

BTW, running with `trace=fail` has a rather high impact on performance, so you might not want to do that always.

Issue #13925, opened 2024-03-16
Multiple compiler and linker warnings building dnsdist 1.9.1


pieterlexis:
- [x] This is not a support question, I have read [about opensource](https://www.powerdns.com/opensource.html) and will send support questions to the IRC channel, [Github Discussions](https://github.com/PowerDNS/pdns/discussions/) or the mailing list.
- [x] I have read and understood the ['out in the open' support policy](https://blog.powerdns.com/2016/01/18/open-source-support-out-in-the-open/)

 - Program: dnsdist
 - Issue type: Bug report

### Short description
Building the AUR package for dnsdist 1.9.1 yields several compiler warnings. This does not prevent dnsdist to build.

### Environment
 - Operating system: Arch Linux
 - Software version: 1.9.1
 - Software source: AUR

```
configure: Configuration summary
configure: =====================
configure:
configure: dnsdist configured with:  '--prefix=/usr' '--sysconfdir=/etc' '--localstatedir=/var' '--with-ebpf' '--with-gnutls' '--with-libsodium' '--with-libssl' '--with-net-snmp' '--with-nghttp2' '--with-quiche' '--with-re2' '--with-xsk' '--enable-dnstap' '--enable-dns-over-http3' '--enable-dns-over-https' '--enable-dns-over-tls' '--enable-dns-over-quic' '--enable-dnscrypt' '--enable-systemd' '--enable-lto=thin' '--enable-auto-var-init=zero' '--with-service-user=dnsdist' '--with-service-group=dnsdist' 'CFLAGS=-march=x86-64 -mtune=generic -O2 -pipe -fno-plt' 'LDFLAGS=-Wl,-O1,--sort-common,--as-needed,-z,relro,-z,now' 'CPPFLAGS=-D_FORTIFY_SOURCE=2' 'CXXFLAGS=-march=x86-64 -mtune=generic -O2 -pipe -fno-plt'
configure:
configure: CC: gcc (gcc (GCC) 13.2.1 20230801)
configure: CXX: g++ -std=c++17 (g++ (GCC) 13.2.1 20230801)
configure: LD: /usr/bin/ld -m elf_x86_64
[...]

PR #13927, opened 2024-03-18
Fix annoying compiler warnings by introducing and switching to `pdns::UniqueFilePtr`

rgacogne:
### Short description

Using a deleter instead of `decltype(&fclose)` has two big advantages:
- the deleter is included in the type and does not have to be passed when creating a new object (easier to use, less memory usage, in theory better inlining)
- we avoid the annoying `ignoring attributes on template argument ‘int (*)(FILE*)’` warning from the compiler, which is there because `fclose` is tagged as `__nonnull((1))`

See https://github.com/PowerDNS/pdns/issues/13925.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/pdns/dnspcap2protobuf.cc
+++ b/pdns/dnspcap2protobuf.cc
@@ -63,9 +63,11 @@ try {
 
   PcapPacketReader pr(argv[1]);
 
+  // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-pointer-arithmetic): it's argv..
   auto filePtr = pdns::openFileForWriting(argv[2], 0600, true, false);
   if (!filePtr) {
     auto error = errno;
+    // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-pointer-arithmetic): it's argv..
     cerr<<"Error opening output file "<<argv[2]<<": "<<stringerror(error)<<endl;
     exit(EXIT_FAILURE);
   }
--- a/pdns/dnspcap2protobuf.cc
+++ b/pdns/dnspcap2protobuf.cc
@@ -63,8 +63,8 @@ try {
 
   PcapPacketReader pr(argv[1]);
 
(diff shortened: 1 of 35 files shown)


pdns/misc.hh
```diff
@@ -836,4 +836,20 @@ private:
...
+struct FilePtrDeleter
+{
+  /* using a deleter instead of decltype(&fclose) has two big advantages:
+     - the deleter is included in the type and does not have to be passed
+       when creating a new object (easier to use, less memory usage, in theory
+       better inlining)
+     - we avoid the annoying "ignoring attributes on template argument ‘int (*)(FILE*)’"
+       warning from the compiler, which is there because fclose is tagged as __nonnull((1))
+  */
+  void operator()(FILE* filePtr) {
```
omoerbeek, line 849: const correctness police: this method can be marked `const`

omoerbeek (changes requested):
Looks good, one nit and a few CodeQL complaints should be fixed.

pdns/misc.hh
rgacogne: Good one! `noexcept` as well, I guess, I'll look into it!

pdns/dnspcap2protobuf.cc
```diff
@@ -63,7 +63,7 @@
 
   PcapPacketReader pr(argv[1]);
 
-  auto fp = std::unique_ptr<FILE, int(*)(FILE*)>(fopen(argv[2], "w"), fclose);
+  auto fp = pdns::UniqueFilePtr(fopen(argv[2], "w"));
```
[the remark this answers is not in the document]
rgacogne: Work as designed: this is a command-line tool without elevated privileges. If the user has the privileges to access these files they can do so without this tool.

omoerbeek (approved):

PR #13935, opened 2024-03-18
rec: on rhel and debian install recursor.yml default file, but only for new installs

omoerbeek:
So only if no existing recursor.conf and no recursor.yml was found. This allows:

- new installes to have a default .yml file.
- existing .conf install to be untouched
- existing .yml installs to be left untouched

This needs careful review from somebody familiar with packaging do's and don't s.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/builder-support/debian/recursor/debian-buster/pdns-recursor.postinst
+++ b/builder-support/debian/recursor/debian-buster/pdns-recursor.postinst
@@ -5,6 +5,9 @@ case "$1" in
   configure)
     addgroup --system pdns
     adduser --system --home /var/spool/powerdns --shell /bin/false --ingroup pdns --disabled-password --disabled-login --gecos "PowerDNS" pdns
+    if [ ! -e /etc/powerdns/recursor.conf -a ! -e /etc/powerdns/recursor.yml ]; then
+      cp /etc/powerdns/recursor.yml-dist /etc/powerdns/recursor.yml
+    fi
   ;;
 
   *)
--- a/builder-support/specs/pdns-recursor.spec
+++ b/builder-support/specs/pdns-recursor.spec
@@ -80,7 +80,7 @@ make %{?_smp_mflags} check || (cat test-suite.log && false)
 %install
 make install DESTDIR=%{buildroot}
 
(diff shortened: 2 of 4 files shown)


builder-support/specs/pdns-recursor.spec
```diff
@@ -106,6 +106,9 @@ getent passwd pdns-recursor > /dev/null || \
 exit 0
 
 %post
+if [ ! -e %{_sysconfdir}/%{name}/recursor.conf -a ! -e %{_sysconfdir}/%{name}/recursor.yml ]; then
+    cp %{_sysconfdir}/%{name}/recursor.yml-dist %{_sysconfdir}/%{name}/recursor.yml
```
omoerbeek, line 110: The `cp` should probably be an install to make sure the perms and owner/group are right.

builder-support/debian/recursor/debian-buster/pdns-recursor.postinst
```diff
@@ -5,6 +5,9 @@ case "$1" in
   configure)
     addgroup --system pdns
     adduser --system --home /var/spool/powerdns --shell /bin/false --ingroup pdns --disabled-password --disabled-login --gecos "PowerDNS" pdns
+    if [ ! -e /etc/powerdns/recursor.conf -a ! -e /etc/powerdns/recursor.yml ]; then
+      cp /etc/powerdns/recursor.yml-dist /etc/powerdns/recursor.yml
```
zeha, line 9: 1) this leaves recursor.yml-dist in /etc/powerdns with possible user confusion
2) this doesn't register recursor.yml as a conffile with dpkg (but recursor.yml-dist would be!)

zeha: Looking at https://manpages.debian.org/testing/dpkg-dev/deb-conffiles.5.en.html I think we're SOL with this approach and instead must install recursor.yml (at build time). 
But that also means migrating old installs!

omoerbeek:
It is clear that this is not the right approach. I have an idea though which might work, will investigate that soon.

PR #13951, opened 2024-03-19
auth: deprecate output of SHA1, GOST hashes

### LINKED ISSUE #13950: Remove SHA1 DS-Record from /cryptokeys API
- Program: Authoritative
 - Issue type: Feature request

### Short description
If you query the HTTP API /cryptokeys, you will currently also receive a SHA1 hash in the list of DS records.

### Description
SHA1 hashes are no longer accepted in most cases and should no longer be delivered with the query.

Habbie:
### Short description

https://datatracker.ietf.org/doc/html/rfc8624#section-3.3

```
   | Number | Mnemonics       | DNSSEC Delegation | DNSSEC Validation |
   +--------+-----------------+-------------------+-------------------+
   | 0      | NULL (CDS only) | MUST NOT [*]      | MUST NOT [*]      |
   | 1      | SHA-1           | MUST NOT          | MUST              |
   | 2      | SHA-256         | MUST              | MUST              |
   | 3      | GOST R 34.11-94 | MUST NOT          | MAY               |
   | 4      | SHA-384         | MAY               | RECOMMENDED       |
```

closes #13950

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/pdnsutil.cc
+++ b/pdns/pdnsutil.cc
@@ -2031,7 +2031,7 @@ static int setZoneKind(const DNSName& zone, const DomainInfo::DomainKind kind)
   return EXIT_SUCCESS;
 }
 
-static bool showZone(DNSSECKeeper& dk, const DNSName& zone, bool exportDS = false)
+static bool showZone(DNSSECKeeper& dk, const DNSName& zone, bool exportDS = false) // NOLINT(readability-function-cognitive-complexity)
 {
   UeberBackend B("default");
   DomainInfo di;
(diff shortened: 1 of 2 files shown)


rgacogne:
It looks like `GOST is also a `MUST NOT`, shall we remove it as well?

Habbie:
> It looks like `GOST is also a `MUST NOT`, shall we remove it as well?

oh yes, agreed. I didn't think about it because we don't ship an implementation at all, but I'll do that. I'll move it under `-v` in pdnsutil (like SHA1 already is)

rgacogne (approved):

PR #13959, opened 2024-03-20
Add Meson as a build system for auth

fredmorcos:
### Short description
Adds Meson as a build system for the authoritative. This has been brewing for quite a while and, IMHO, is still not ideal. But honestly there's too much to clean up until we reach a point where our build system can be considered clean and simple. This is a step in that direction.

This currently supports building the auth, with its backends, with build-time configuration and feature/dependency detection. It is also able to build and run `pdns-auth-testrunner` and the remotebackend tests reliably.

This PR is an invitation to review it, play with it, break it, report bugs, missing stuff and create a list of items that need to be worked on for this to eventually become the default.

I think we can merge this since it can build the auth and at least gives us a development environment that is easier to use. I won't list the advantages of Meson here but things like a compilation database that is correctly updated automatically, and having several build directories with different configuration flags is a huge plus.

Most of this are new files, if not all of it, so the review and merging process should be quite easy.

A lot is still missing and once this PR is merged I hope that we can all start contributing to making this the default for auth. Here's a rough list:
- [ ] Man pages can be generated by a python script but there's no reasonable way to create them from meson yet. Sphinx is annoying to work with because it wants to handle a full directory at a time instead of individual files. Current man pages are defined as a run target but that's not idea because 1) they don't get built with the souces, and 2) they don't get installed.
- [ ] Installation
- [ ] Dist (e.g. `meson dist`) - this might end up working without any involvement
- [ ] Regression tests are not yet integrated
- [ ] I haven't tried randomly enabling and disabling various build options
- [ ] CI integration - there's already been 1 attempt but things have changed since then, I attached a diff in case it's useful as a starting point [meson-pyinvoke-ci.patch.gz](https://github.com/PowerDNS/pdns/files/14693191/meson-pyinvoke-ci.patch.gz)
- [ ] Documentation updates
- [ ] Take a look at meson sourcesets if it could simplify all the conditionals this uses
- [ ] Muon fmt, and use it in CI to check the formatting of `meson.build` files too
- [ ] Muon static analyzer, probably also make use of this in CI
- [ ] See if we can avoid rebuilding everything when a backend is added/removed (because of `PDNS_MODULES` in `config.h`), or when tests are enabled/disabled
- [ ] Generated files like `bindparser.cc` should go into the `build` directory. As long as we're still using autotools this will not be possible to achieve.

### Checklist
I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code


rgacogne:
I have only skimmed through it for now but I really like what I have seen so far. Wonderful job, Fred!

fredmorcos:
> I have only skimmed through it for now but I really like what I have seen so far. Wonderful job, Fred!

Thank you Remi

Habbie (commented):
This is amazing and impressive. I did not review everything yet, I'd rather see this built and tested in CI first, but then I will.

Local testing revealed a few small issues, taking those to chat with Fred.

Habbie:
a TODO for later (I presume we'll open one or more separate issues when we merge this): muon fmt and perhaps the muon linter that Fred mentioned to me once

fredmorcos:
> a TODO for later (I presume we'll open one or more separate issues when we merge this): muon fmt and perhaps the muon linter that Fred mentioned to me once

Adding those to the TODO list

fredmorcos:
> This is amazing and impressive. I did not review everything yet, I'd rather see this built and tested in CI first, but then I will.
> 
> Local testing revealed a few small issues, taking those to chat with Fred.

Thank you Peter.

I am hoping to get this merged before it's integrated with CI. Then we can all start chiming in with completing TODOs and working on CI integration.

Habbie:
> I am hoping to get this merged before it's integrated with CI.

Given that it works (which I can now confirm), that works for me too :)

Habbie (approved):
This generates working binaries for me, and touches almost nothing that already existed, so it is safe to merge!

omoerbeek (approved):

PR #13969, opened 2024-03-22
rec: use shared NOD (and/or UDR) DB, to avoid multiple copies in memory and on disk

### LINKED ISSUE #13677: rec: ponder NOD/UDR data to be shared betwen threads
Right now each worker thread maintains a full NOD/UDR DB. It's probably better to share that data (though this statement needs to be validated in case there are many workers to not cause lock contention).

Related: each 10 minutes all threads write the full DB to storage. This leads to many GBs of data written each interval, wearing out storage devices. Moving to a longer (tunable?) period will alleviate this (in addition to using 1 DB instead of N).

omoerbeek:
### Short description

This modified the threading model of the NOD/UDR subsystem. We now use a single DB, which has the big advantage that there's way less memory and disk storage used. Also, multiple threads will not report the same name, as they base their decision on the same data.

This PR also cleans up the code and introduces a setting to set the interval for the writing snapshots to disk (0 means switch the disk based snapshot mechanism off).

As a bonus the code to select the newest snap actually does what's promised. Existing code picked the oldest snap, unless I'm very confused, so please do a review for that commit with good attention to detail.

Ideally it should be confirmed that lock contention is not  an issue here. I do not expect given the number we have seen for the record cache (packet cache hits do not call into the NOD code). 

There might be a way to reduce potential lock contention even more: only call the NOD/UDR code if the answer was produced using outgoing queries (i.e. a non-record cache hit). If the answer is coming straight out of the record cache, it will not contain new data. This may not be true however if the cache refresh mechanism is used, so beware!

Tagging @neilcook, for some reason I cannot add him as reviewer.

Fixes #13677

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [X] documented the code

### DIFF
--- a/pdns/recursordist/nod.cc
+++ b/pdns/recursordist/nod.cc
@@ -145,18 +145,17 @@ bool PersistentSBF::snapshotCurrent(std::thread::id tid)
         }
         // Now write it out to the file
         std::string ftmp = file.string() + ".XXXXXXXX";
-        int fileDesc = mkstemp(ftmp.data());
+        auto fileDesc = FDWrapper(mkstemp(ftmp.data()));
         if (fileDesc == -1) {
           throw std::runtime_error("Cannot create temp file: " + stringerror());
         }
         const std::string str = oss.str(); // XXX creates a copy, with c++20 we can use view()
         ssize_t len = write(fileDesc, str.data(), str.length());
         if (len != static_cast<ssize_t>(str.length())) {
-          close(fileDesc);
           filesystem::remove(ftmp.c_str());
           throw std::runtime_error("Failed to write to file:" + ftmp);
         }
-        if (close(fileDesc) != 0) {
+        if (fileDesc.reset() != 0) {
           filesystem::remove(ftmp);
           throw std::runtime_error("Failed to write to file:" + ftmp);
         }
--- a/pdns/recursordist/nod.cc
+++ b/pdns/recursordist/nod.cc
@@ -43,9 +43,9 @@ std::mutex PersistentSBF::d_cachedir_mutex;
 void PersistentSBF::remove_tmp_files(const filesystem::path& path, std::lock_guard<std::mutex>& /* lock */)
 {
   Regex file_regex(d_prefix + ".*\\." + bf_suffix + "\\..{8}$");
-  for (filesystem::directory_iterator i(path); i != filesystem::directory_iterator(); ++i) {
-    if (filesystem::is_regular_file(i->path()) && file_regex.match(i->path().filename().string())) {
-      filesystem::remove(*i);
+  for (const auto& file : filesystem::directory_iterator (path)) {
+    if (filesystem::is_regular_file(file.path()) && file_regex.match(file.path().filename().string())) {
+      filesystem::remove(file);
     }
   }
 }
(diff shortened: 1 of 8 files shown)


pdns/recursordist/nod.cc
```diff
@@ -59,27 +59,27 @@ bool PersistentSBF::init(bool ignore_pid)
...
-          if (filesystem::is_regular_file(i->path()) && file_regex.match(i->path().filename().string())) {
-            if (ignore_pid || (i->path().filename().string().find(std::to_string(getpid())) == std::string::npos)) {
+        for (const auto& file : filesystem::directory_iterator(path)) {
+          if (filesystem::is_regular_file(file.path()) && file_regex.match(file.path().filename().string())) {
+            if (ignore_pid || (file.path().filename().string().find(std::to_string(getpid())) == std::string::npos)) {
               // look for the newest file matching the regex
-              if ((last_write_time(i->path()) < newest_time) || newest_file.empty()) {
-                newest_time = last_write_time(i->path());
-                newest_file = i->path();
+              if (last_write_time(file.path()) > newest_time) {
```
neilcook, line 74: This new logic is definitely correct :)

neilcook (commented):
There are comments in the PersistentSBFFilter add(), and testAndAdd() methods that state `// The only time this should block is when snapshotting`. However that's no longer true, so I recommend changing or removing these comments.

pdns/recursordist/nod.cc
neilcook: BTW the comment for this function could probably do with updating, it currently states:
```
// This looks for an old (per-thread) snapshot. The first one it finds,
// it restores from that.
```
I think it should say something like:
```
// This looks for the newest (per-thread) snapshot it can find,
// and restores from that.
```

pdns/recursordist/nod.cc
```diff
@@ -130,39 +130,38 @@ void PersistentSBF::setCacheDir(const std::string& cachedir)
...
           // only lock while dumping to a stringstream
-          d_sbf.lock()->dump(iss);
+          d_sbf.lock()->dump(oss);
         }
         // Now write it out to the file
-        std::string ftmp = f.string() + ".XXXXXXXX";
-        int fd = mkstemp(&ftmp.at(0));
-        if (fd == -1) {
+        std::string ftmp = file.string() + ".XXXXXXXX";
+        int fileDesc = mkstemp(ftmp.data());
```
rgacogne, line 148: `FDWrapper` might be nice?

rgacogne: Nice!

pdns/recursordist/nod.cc
```diff
@@ -40,46 +40,46 @@ namespace filesystem = boost::filesystem;
...
-// instances iterating and writing to the cache dir at the same time
+// This looks for the newest (per-thread) snapshot it can find and it restores from that. Then
+// immediately snapshots with the current thread id, before removing the old snapshot.
+// In this way, we can have per-thread SBFs, but still snapshot and restore.  The mutex has to be
+// static because we can't have multiple (i.e. per-thread) instances iterating and writing to the
+// cache dir at the same time
 bool PersistentSBF::init(bool ignore_pid)
 {
   auto log = g_slog->withName("nod");
   std::lock_guard<std::mutex> lock(d_cachedir_mutex);
```
rgacogne, line 61: Unrelated to this PR but it's unclear to me what `d_cachedir_mutex` protects. If it's there to prevent two threads from touching the cache directory at the same time, shouldn't `snapshotCurrent()` hold the mutex?

rgacogne (approved):
One question, but the PR looks good to me!

pdns/recursordist/nod.cc
omoerbeek: A added a comment to explain this.

paddg:
I would have thought that this would mean that there would now only be one file on the disk. But there are still two. Did I misunderstand?

```
# pdns_recursor --version 2>&1 | head -1
Nov 04 13:25:59 PowerDNS Recursor 5.1.2 (C) PowerDNS.COM BV

# ls -l /run/pdns-recursor-nod-1/*
-rw------- 1 pdns-recursor pdns-recursor 67108874  4. Nov 13:05 /run/pdns-recursor-nod-1/nod_140629661593728_12237.bf
-rw------- 1 pdns-recursor pdns-recursor 67108874  4. Nov 13:05 /run/pdns-recursor-nod-1/udr_140629661593728_12237.bf
```

omoerbeek:
The NOD and UDR functions each use a separate DB (file). Previously, there would be a two DBs (files) _per thread_.

paddg:
Ah! Got it now. Previously there were 5 each :-) Thanks!

PR #13980, opened 2024-03-25
Do shuffle TCP responses except *XFRs

### LINKED ISSUE #13003: auth: records aren't shuffled in the responses for TCP queries
- Program: Authoritative
 - Issue type: Bug report

### Short description

Records aren't shuffled in the responses for TCP queries but rather they're always returned in the same order. This will have an impact on load balancing if the resolver querying the server always uses the first response or does round-robin instead of randomizing the order.

### Environment
 - Software version: 4.8.1-1

### Steps to reproduce

I think this will occur with any PowerDNS setup, but this is ours:

https://github.com/GrapheneOS/ns1.grapheneos.org

We use the GeoIP backend and Lua records, but this occurs for regular records too.

You can easily reproduce this querying anything with multiple answers via TCP. For example:

```bash
for i in {1..10}; do drill -t releases.grapheneos.org @ns2.grapheneos.org; done
```

Alternatively:

```bash
for i in {1..10}; do dig +tcp releases.grapheneos.org @ns2.grapheneos.org; done
```

### Expected behaviour

Responses via TCP should be randomized like responses via UDP.

### Actual behaviour

The order of records is the same every time when queried via TCP.

### Other information

This isn't impacted by the packet cache, query cache or `consistent-backends`. I have not yet tried other backends, but I don't see a way it could be specifically caused by the GeoIP backend.

This change appears to have been a side effect of avoiding shuffling in AXFR from all the way back in 2005:

e9dd48f9018ef291849c881d88faa66baab5c592

```diff
-  if(!arg().mustDo("no-shuffle")) {
+  if(!d_tcp && !arg().mustDo("no-shuffle")) {
```

karelbilek:
### Short description
Currently, TCP responses are not shuffled.

It seems the heuristics way back in 2005 was "when it's TCP, it's probably a zone transfer", which might have made sense back then, but not in 2024.

Fixes #13003 in the way it was suggested in the github discussion.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [x] documented the code

### DIFF
--- a/pdns/dnspacket.cc
+++ b/pdns/dnspacket.cc
@@ -279,7 +279,7 @@ void DNSPacket::wrapup(bool throwsOnTruncation)
     });
   static bool mustNotShuffle = ::arg().mustDo("no-shuffle");
 
-  if(!d_tcp && !mustNotShuffle) {
+  if(!d_xfr && !mustNotShuffle) {
     pdns::shuffle(d_rrs);
   }
   d_wrapped=true;
@@ -453,6 +453,7 @@ std::unique_ptr<DNSPacket> DNSPacket::replyPacket() const
   r->d_haveednscookie = d_haveednscookie;
   r->d_ednsversion = 0;
   r->d_ednsrcode = 0;
+  r->d_xfr = d_xfr;
 
   if(d_tsigkeyname.countLabels()) {
     r->d_tsigkeyname = d_tsigkeyname;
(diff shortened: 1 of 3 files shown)


Habbie (approved):

urbanchef:
Any plans to release pdns auth with this fix?

Habbie:
absolutely, but no date is set yet

PR #13999, opened 2024-03-28
dnsdist: Fix DNS over HTTP connections/queries counters with `nghttp2` 

rgacogne:
### Short description

This PR:
- increases the HTTP/1.1 query counter when a DoH connection is initiated with a `http/1.1` ALPN and the `nghttp2` provider is used, so we can detect these. We will not actually process the DoH payload (if any) anyway ;
- fix the HTTP connections counter when the `nghttp2` provider is used, as it was not updated ;
- add a regression test for these two values.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


chbruyand (approved):
LGTM!

PR #14006, opened 2024-03-29
FDWrapper: Do not try to close negative file descriptors

rgacogne:
### Short description

It turns out that some of the BPF helper functions return a negative `errno` value in case of failure, and since we wrap the return value into a `FDWrapper` right away this led to a warning from Valgrind about trying to close an invalid file descriptor.

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/misc.hh
+++ b/pdns/misc.hh
@@ -803,7 +803,7 @@ struct FDWrapper
 
   FDWrapper& operator=(FDWrapper&& rhs) noexcept
   {
-    if (d_fd != -1) {
+    if (d_fd >= 0) {
       close(d_fd);
     }
     d_fd = rhs.d_fd;
@@ -824,7 +824,7 @@ struct FDWrapper
   int reset()
   {
     int ret = 0;
-    if (d_fd != -1) {
+    if (d_fd >= 0) {
       ret = close(d_fd);
       d_fd = -1;
     }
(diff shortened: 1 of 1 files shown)


omoerbeek (approved):

pdns/misc.hh
```diff
@@ -803,7 +803,7 @@ struct FDWrapper
 
   FDWrapper& operator=(FDWrapper&& rhs) noexcept
   {
-    if (d_fd != -1) {
+    if (d_fd >= 0) {
```
cmouse, line 806: should there be check here that fd != -1?

pdns/misc.hh
```diff
@@ -824,7 +824,7 @@ struct FDWrapper
   int reset()
   {
     int ret = 0;
-    if (d_fd != -1) {
+    if (d_fd >= 0) {
       ret = close(d_fd);
       d_fd = -1;
```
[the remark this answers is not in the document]
rgacogne: Will do!

pdns/misc.hh
rgacogne: I'm not sure I get what you have in mind?

cmouse: i was just wondering if it is acceptable, or not, to pass around closed fd handle?

rgacogne: Ah, right! I believe it is fine, yes, so I don't think we need an additional check.

cmouse (approved):

PR #14018, opened 2024-04-02
Rec: add setting to exclude specific listen socket addresses from requiring proxy protocol

### LINKED ISSUE #13948: rec: allow more flexible specification of proxy_protocol_from
The `proxy-protocol-from` setting is a list of subnets, defining the source IP ranges that must use the proxy protocol. Port does not play a role in this currently.

Often a recursor listens on 127.0.0.1, getting requests form dnsdist using the proxy protocol. For debugging purposes it would be nice to be able to connect to (and from) 127.0.0.1 using a tool that does not know the proxy protocol. 
A possible way to do this is to specify a port in the config that does not require the proxy protocol. 

Exact syntax and semantics to be defined.

omoerbeek:
### Short description

Fixes #13948

I had to move the code that retrieves the local address up to be able to check the exception for both UDP and TCP. This part of the change should be reviewed extra carefully to not have bad side-effects.
Potentially an extra syscall is done before dropping a connection/packet. 

You can use this as follows:

```yaml
incoming:
  listen:
    - 0.0.0.0:5301
    - 0.0.0.0:5302
    - '[::]:5301'
  proxy_protocol_from: [127.0.0.1]
  proxy_protocol_exceptions: [127.0.0.1:5302]
```
Has the effect that queries _to_ `127.0.0.1:5302` will not require proxy protocol, while queries _to_ `127.0.0.1:5301` will. Note that `proxy_protocol_from` talks about remote addresses/subnets, while `proxy_protocol_exceptions` talks about local addresses. I hope I made that clear in the docs.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/recursordist/settings/table.py
+++ b/pdns/recursordist/settings/table.py
@@ -2072,7 +2072,7 @@ The dnsdist docs have `more information about the PROXY protocol <https://dnsdis
         'help' : 'A Proxy Protocol header should not be used for these listen addresses.',
         'doc' : '''
 If set, clients sending from an address in :ref:`setting-proxy-protocol-from` to a address:port listed here are excluded from using the Proxy Protocol.
-If not port is specified, port 53 is assumed.
+If no port is specified, port 53 is assumed.
 This is typically used to provide an easy to use address and port to send debug queries to.
  ''',
         'versionadded' : '5.1.0',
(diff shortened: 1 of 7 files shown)


pdns/recursordist/settings/table.py
```diff
@@ -2064,6 +2064,19 @@
...
     },
+    {
+        'name' : 'proxy_protocol_exceptions',
+        'section' : 'incoming',
+        'type' : LType.ListSocketAddresses,
+        'default' : '',
+        'help' : 'A Proxy Protocol header should not be used for these listen addresses.',
+        'doc' : '''
+If set, clients sending from an address in :ref:`setting-proxy-protocol-from` to a address:port listed here are excluded from using the Proxy Protocol.
+If not port is specified, port 53 is assumed.
```
rgacogne, line 2075:
```suggestion
If no port is specified, port 53 is assumed.
```

PR #14021, opened 2024-04-02
auth LUA: make whitespace insertion on chunk combine optional

Habbie:
### Short description
draft. Needs docs. clang-tidy will likely complain about short variable names.

On backport to 4.9, make sure to flip the default for this setting to 'yes'

### Checklist

I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


pdns/dnsrecords.cc
```diff
@@ -19,6 +19,7 @@
  * along with this program; if not, write to the Free Software
  * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
  */
+#include "auth-main.hh"
```
Habbie, line 22: this needs to go

neilcook:
I tested this and indeed it works as expected at execution time - a long Lua script that previous failed due to the whitespace issue now runs just fine and returns a result.
However I found that when querying the API for the record, the whitespace still appeared, which I found quite confusing, particularly as I found this before I tried executing the script, so I initially assumed it wasn't working.

Habbie:
> However I found that when querying the API for the record, the whitespace still appeared, which I found quite confusing, particularly as I found this before I tried executing the script, so I initially assumed it wasn't working.

good one. Same in `pdnsutil edit-zone`. I wonder if we can do the join there too, but we don't want to confuse people who did manual chunking at the same time.

Habbie:
> when querying the API for the record, the whitespace still appeared

After merging this PR, I'll reopen the original issue to cover that part of the question. We'll backport only this PR (with the default set to whitespace insertion) to 4.9 and might do API and list-zone/edit-zone for 5.0.

rgacogne (approved):
Looks good!

PR #14023, opened 2024-04-03
rec: add interface (not subject to proxy protocol substitutions) addresses in Lua DNSQuestion and corresponding FFI

### LINKED ISSUE #13730: REC: When using PROXY, provide the "real" localaddr somewhere
- Program: Recursor
 - Issue type: Feature request

### Short description
When using the Proxy protocol, `dq.localaddr` and `dq.remoteaddr` contain the address the client connected to, and the client's ip.

While this makes ACLs etc real easy, sometimes we need the "local" localaddr and "local" remoteaddr.

### Usecase
Imagine a setup with one or more dnsdist proxies and a number of pdns-rec backends, using the PROXY protocol. Each backend-server could have multiple backend IP addresses.

For debugging/logging/metric-gathering purposes, we'd need to log not only the client's ip address (`dq.remoteaddr`), the ip the client connected to (`dq.localaddr`), but also the local ip on the backend which processed the query. 

### Description
Note that as the backend might have multiple IPs accepting queries, just the "primary" ip or hostname might not be enough in the logs.

Suggestion would be to add e.g. `real_localaddr` and `real_remoteaddr` to the `dq` object when removing the PROXY headers. That removes a bit of the PROXY transparancy, but I feel this would be justified in this case.

omoerbeek:
### Short description

Fixes #13730

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/recursordist/docs/lua-scripting/dq.rst
+++ b/pdns/recursordist/docs/lua-scripting/dq.rst
@@ -57,13 +57,13 @@ The DNSQuestion object contains at least the following fields:
       :class:`ComboAddress` where this query was received on.
       If the proxy protocol is used, this will contain the destination address from the proxy protocol header.
 
-  .. attribute:: DNSQuestion.phys_remoteaddr
+  .. attribute:: DNSQuestion.interface_remoteaddr
 
-      :class:`ComboAddress` of the physical requestor, that is, the physical network source address of the request.
+      :class:`ComboAddress` of the interface the requestor used, that is, the network source address of the request.
 
-  .. attribute:: DNSQuestion.phys_localaddr
+  .. attribute:: DNSQuestion.interface_localaddr
 
-      The physical :class:`ComboAddress` where this query was received on, which is one of the listening addresses of the recursor.
+      The :class:`ComboAddress` of the interface where this query was received on, which is one of the listening addresses of the recursor.
 
   .. attribute:: DNSQuestion.variable
 
--- a/pdns/recursordist/docs/lua-scripting/dq.rst
+++ b/pdns/recursordist/docs/lua-scripting/dq.rst
@@ -59,11 +59,11 @@ The DNSQuestion object contains at least the following fields:
 
   .. attribute:: DNSQuestion.interface_remoteaddr
 
-      :class:`ComboAddress` of the interface the requestor used, that is, the network source address of the request.
+      Source :class:`ComboAddress` of the packet received by the recursor. If the proxy protocol is not used, the value will match ``remoteaddr``.
 
   .. attribute:: DNSQuestion.interface_localaddr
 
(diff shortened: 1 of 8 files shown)


rgacogne (commented):
Code looks good but I'm not fond of `physical` which is a bit confusing to me, as I would expect an ethernet address. DNSdist uses `hopRemote` and `hopLocal` internally which is not great either, to be honest.

omoerbeek:
Te name is confusing indeed. Pondering original/real/interface...

omoerbeek:
I'm leaning towards `interface_remoteaddr` and `interface_localaddr`. @rgacogne  @Habbie how about that?

Habbie:
sounds good to me

franklouwers:
Keeping Physical would be consistent with https://doc.powerdns.com/recursor/lua-config/proxymapping.html (note that I don't like the naming there either, if we change it here, we might consider changing it there as well to remain consistent)

omoerbeek:
> Keeping Physical would be consistent with https://doc.powerdns.com/recursor/lua-config/proxymapping.html (note that I don't like the naming there either, if we change it here, we might consider changing it there as well to remain consistent)

The term "physical" is only used in the docs there, it is not something that appears in the current Lua scripting/config API. I'll add a commit to change the terminology there as well..

pdns/recursordist/docs/lua-scripting/dq.rst
```diff
@@ -50,10 +50,20 @@ The DNSQuestion object contains at least the following fields:
...
+      If the proxy protocol is used, this will contain the source address from the proxy protocol header.
 
   .. attribute:: DNSQuestion.localaddr
 
       :class:`ComboAddress` where this query was received on.
+      If the proxy protocol is used, this will contain the destination address from the proxy protocol header.
+
+  .. attribute:: DNSQuestion.interface_remoteaddr
+
+      :class:`ComboAddress` of the interface the requestor used, that is, the network source address of the request.
```
rgacogne, line 62:
```suggestion
      Source :class:`ComboAddress` of the packet received by the recursor. If the proxy protocol is not used, the value will match ``remoteaddr``.
```

pdns/recursordist/docs/lua-scripting/dq.rst
```diff
@@ -50,10 +50,20 @@ The DNSQuestion object contains at least the following fields:
...
+
+  .. attribute:: DNSQuestion.interface_localaddr
+
+      The :class:`ComboAddress` of the interface where this query was received on, which is one of the listening addresses of the recursor.
```
rgacogne, line 66:
```suggestion
      Destination :class:`ComboAddress` of the packet received by the recursor. If the proxy protocol is not used, the value will match ``localaddr``.
```

rgacogne (approved):
Looks good, one suggestion for the documentation.

PR #13960, opened 2024-03-20
Convert remotebackend unit tests to use python

cmouse:
### Short description
Change remotebackend to use python unit tests.
Also fixes few minor issues.

### Checklist

I have:
- [x] read the [CONTRIBUTING.md](https://github.com/PowerDNS/pdns/blob/master/CONTRIBUTING.md) document
- [x] compiled this code
- [x] tested this code
- [ ] included documentation (including possible behaviour changes)
- [ ] documented the code
- [x] added or modified regression test(s)
- [x] added or modified unit test(s)

PR #14030, opened 2024-04-04
ci: Enable LeakSanitizer during dnsdist and recursor unit tests

rgacogne:
### Short description

We need to fix some one-time allocations in the authoritative server that are reported as leaked memory before we can enable it for the authoritative server as well. See:
- https://github.com/PowerDNS/pdns/pull/14028
- https://github.com/PowerDNS/pdns/pull/14029

<strike>There is also a leak in the remotebackend unit tests that I will investigate after https://github.com/PowerDNS/pdns/pull/13960 has been merged. Fixed by https://github.com/PowerDNS/pdns/pull/14033</strike>

### Checklist

I have:
- [ ] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/.github/workflows/build-and-test-all.yml
+++ b/.github/workflows/build-and-test-all.yml
@@ -122,7 +122,6 @@ jobs:
     container:
       image: ghcr.io/powerdns/base-pdns-ci-image/debian-12-pdns-base:master
       env:
-        ASAN_OPTIONS: detect_leaks=0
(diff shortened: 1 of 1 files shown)


romeroalx:
But it is already enabled for auth, right @rgacogne ?

rgacogne:
I'm merging this one now and will do a follow-up PR for the authoritative server later.

PR #14034, opened 2024-04-04
dnsdist: Document how to generate a console key without dnsdist

rgacogne:
### Checklist

I have:
- [ ] compiled this code
- [ ] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/dnsdistdist/docs/guides/console.rst
+++ b/pdns/dnsdistdist/docs/guides/console.rst
@@ -20,6 +20,10 @@ Once you have a console-enabled dnsdist, the first step to enable encryption is 
   > makeKey()
   setKey("ENCODED KEY")
 
+The key does not have a specific format, so base-64 encoding 32 random bytes works as well::
+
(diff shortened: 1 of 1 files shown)


Habbie (approved):
I also had a macOS user test the line, works there too.

`openssl rand -base64 32` would work on many systems as well. I'm pondering which version is better, or whether we should just mention both? Approved in any case.

rgacogne:
Good one, I'll add `openssl rand -base64 32` as well!

phonedph1:
If we add the `-C /dev/null` trick to the dnsdist command example wouldn't that solve most issues people have had with making a key too?

rgacogne:
> If we add the `-C /dev/null` trick to the dnsdist command example wouldn't that solve most issues people have had with making a key too?

Yes and no, I understand some people are generating the configuration on a machine where dnsdist is not installed. But I'll add it as well!

rgacogne:
Done.

Habbie (approved):

Issue #13987, opened 2024-03-26
Auth Meson Tracking Issue


fredmorcos:
This is a top-level issue to track the progress of getting auth fully built, tested and released using Meson:

[ ] Man pages can be generated by a python script but there's no reasonable way to create them from meson yet. Sphinx is annoying to work with because it wants to handle a full directory at a time instead of individual files. Current man pages are defined as a run target but that's not idea because 1) they don't get built with the souces, and 2) they don't get installed.
[ ] Installation.
[ ] Dist (e.g. meson dist) - this might end up working without any involvement.
[ ] Regression tests are not yet integrated.
[ ] I haven't tried randomly enabling and disabling various build options.
[ ] CI integration - there's already been 1 attempt but things have changed since then, I attached a diff in case it's useful as a starting point [meson-pyinvoke-ci.patch.gz](https://github.com/PowerDNS/pdns/files/14693191/meson-pyinvoke-ci.patch.gz)
[ ] Documentation and README updates.
[ ] Take a look at meson sourcesets if it could simplify all the conditionals we use.
[ ] Muon fmt, and use it in CI to check the formatting of `meson.build` files too.
[ ] Muon static analyzer, probably also make use of this in CI.
[ ] See if we can avoid rebuilding everything when a backend is added/removed (because of `PDNS_MODULES` in `config.h`, currently only used in `version.cc`), or when tests are enabled/disabled.
[ ] Generated files like bindparser.cc should go into the build directory. As long as we're still using autotools this will not be possible to achieve.
[ ] The MacOS linker is producing warnings about duplicate link flags, see if this can be avoided:
```
[116/121] Linking target pdns-zone2sql
ld: warning: ignoring duplicate libraries: '-lcurl'
[117/121] Linking target pdns-auth-control
ld: warning: ignoring duplicate libraries: '-lcurl'
[118/121] Linking target pdns-zone2json
ld: warning: ignoring duplicate libraries: '-lcurl'
[119/121] Linking target pdns-auth
ld: warning: ignoring duplicate libraries: '-lcurl'
[121/121] Linking target pdns-auth-util
ld: warning: ignoring duplicate libraries: '-lcurl'
```

miodvallat:
See also: #15048

RvdE:
I just downloaded the 5.0.1 tarball and noticed there are 2 directories with regard to meson: meson.build and meson_options.txt.
Both contain a file with the same name as the directory, but they should not be in a directory in the first place.

[...]

PR #14036, opened 2024-04-05
GH Actions - Meson: Build auth using meson in `build-and-test-all`.

romeroalx:
### Short description
Initial work as part of the tasks included in #13987 for building `auth` using Meson inside the `build-and-test-all` workflow.

Output binaries from the meson build are not installed but only generated (current `ci-make-install step`).

### Checklist

I have:
- [ ] compiled this code
- [x] tested this code
- [ ] documented the code

### DIFF
--- a/.github/workflows/build-and-test-all.yml
+++ b/.github/workflows/build-and-test-all.yml
@@ -38,10 +38,18 @@ jobs:
       env:
         ASAN_OPTIONS: detect_leaks=0
         FUZZING_TARGETS: yes
-        SANITIZERS: asan+ubsan
         UBSAN_OPTIONS: "print_stacktrace=1:halt_on_error=1:suppressions=${{ env.REPO_HOME }}/build-scripts/UBSan.supp"
         UNIT_TESTS: yes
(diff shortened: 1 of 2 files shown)


fredmorcos:
I had a short call with @romeroalx and we discussed moving more logic into `tasks.py` in a way that completely abstract build system internals from the user and from Github workflows.

fredmorcos (approved):
Awesome! This is fine by me to get merged for now.
Once we merge the `--enable-systemd` flag for meson you can fix that TODO and merge this PR.

One thing that irks me (but we shouldn't do anything about it just yet) is the use of environment variables. I think in the long term we want to move to being very explicit about everything we pass, so using invoke parameters would be better, and we can get rid of weird stuff like `echo SANITIZERS=... >> $GITHUB_ENV`.

fredmorcos:
The systemd configure flag is added here: #14055

fredmorcos:
> The systemd configure flag is added here: #14055

That was merged. So this PR can now be merged after the systemd configure flag is enabled on meson.

Habbie:
This branch has conflicts that must be resolved

PR #14049, opened 2024-04-08
rec: do not count RRSIGs using unsupported algorithms towards RRSIGs limit

omoerbeek:
I don't think we should count RRSIGs with unsupported algos towards the RRSIG limit, just like expired RRSIGs are skipped.

I found two places where the RRSIG's limit is increased if we're encountering unsupported algo's.

In `validateWithKeySet` the `checkSignatureWithKey` call currently returns `false` because `DNSCryptoKeyEngine::makeFromPublicKeyString` fails for unsupported algorithms. We must check earlier, as otherwise the counter is already increased. It does not look lik `ede` set is passed to the caller here. 

In `validateDNSKeysAgainstDS` we have a similar case. The `ede` does not seem to get used if an unsupported algo is encountered.

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/pdns/validate.cc
+++ b/pdns/validate.cc
@@ -1022,6 +1022,9 @@ vState validateWithKeySet(time_t now, const DNSName& name, const sortedRecords_t
     }
 
     vState ede = vState::Indeterminate;
+    if (!DNSCryptoKeyEngine::isAlgorithmSupported(signature->d_algorithm)) {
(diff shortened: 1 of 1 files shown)


rgacogne (approved):
Looks good to me! A test would be nice to have :)

PR #14057, opened 2024-04-10
Auth: fix a crash and some cleanup in the auth-catalogzone.cc

mind04:
### Short description
An incomplete SOA record like ```ns.example.com mail.example.com 1``` in a producer zone resulted in termination of the communicator thread.

### Checklist
I have:
- [x] compiled this code
- [x] tested this code
- [ ] documented the code


modules/lmdbbackend/lmdbbackend.cc
```diff
@@ -1819,7 +1819,7 @@ void LMDBBackend::getUpdatedPrimaries(vector<DomainInfo>& updatedDomains, std::u
 
     if (di.kind == DomainInfo::Producer) {
       catalogs.insert(di.zone);
-      catalogHashes[di.zone].process("\0");
+      catalogHashes[di.zone].process("");
```
Habbie, line 1822: I know this change changes nothing, but what is the purpose of processing zero bytes?

Habbie (commented):
one question; will do a full review later

modules/lmdbbackend/lmdbbackend.cc
mind04: This will create catalogHashes[di.zone] for a catalog with no members. Without this there is no serial increase when you delete the last member from a catalog.

docs/lua-records/index.rst
```diff
@@ -177,9 +177,9 @@ outside of Europe will hit 198.51.100.1 as long as it is available, and the
 
 Advanced topics
 ---------------
-By default, LUA records are executed with ``return `` prefixed to them. This saves
+By default, LUA records are executed with ``return`` prefixed to them. This saves
 a lot of typing for common cases. To run actual Lua scripts, start a record with a ``;``
-which indicates no ``return `` should be prepended.
+which indicates no ``return`` should be prepended.
```
Habbie, line 182: this appears to have snuck into the wrong PR. It's also actually wrong :)

pdns/auth-catalogzone.cc
```diff
@@ -83,41 +85,48 @@ std::string CatalogInfo::toJson() const
...
   if (!d_coo.empty()) {
     object["coo"] = d_coo.toString();
   }
+
+  // unique property
   if (!d_unique.empty()) {
-    if (d_unique.countLabels() > 1) {
-      throw std::out_of_range("Multiple labels in a unique value are not allowed");
+    if (d_unique.countLabels() != 1) {
+      throw std::out_of_range("Invallid number of labels in unique value");
```
Habbie, line 97:
```suggestion
      throw std::out_of_range("Invalid number of labels in unique value");
```

Habbie (changes requested):
nice cleanup :) I left a few comments. Also, clang-tidy is unhappy about your short `ci` and `di` variable names (you can see this in the Files view or https://github.com/PowerDNS/pdns/actions/runs/8660761587)

pdns/auth-catalogzone.cc
```diff
@@ -25,56 +25,58 @@
...
-        if (items["group"].is_array()) {
-          for (const auto& value : items["group"].array_items()) {
-            d_group.insert(value.string_value());
-          }
-        }
-        else {
-          throw std::out_of_range("Key 'group' is not an array");
-        }
+      for (const auto& value : items["group"].array_items()) {
+        d_group.insert(value.string_value());
```
miodvallat, line 76: Out of paranoia, this probably needs a `value.is_string()` guard before invoking `string_value()`.

pdns/test-dnsname_cc.cc
```diff
@@ -1015,6 +1015,27 @@ BOOST_AUTO_TEST_CASE(test_getlastlabel) {
   BOOST_CHECK_EQUAL(ans, DNSName("com"));
 }
 
+BOOST_AUTO_TEST_CASE(test_getfirtslabel)
```
miodvallat:
```suggestion
BOOST_AUTO_TEST_CASE(test_getfirstlabel)
```

pdns/auth-catalogzone.cc
mind04: https://github.com/PowerDNS/pdns/blob/master/ext/json11/json11.hpp#L143

According to this string_value() is always returning a string. An empty group value is harmless.

miodvallat: Fine, then.

Please add a `Signed-off-by` line to your commits so that this can get merged.

mind04:
> What do you plan to use the last commit (`getFirstLabel`) for?

You'll find out, but not today ;)

miodvallat (approved):

PR #14031, opened 2024-04-04
Meson fix for `libdecaf` detection

fredmorcos:
### Short description
This fixes `libdecaf` detection on Github runners. The library is installed in `/usr/local` and, when given custom directories to search in, meson doesn't add the `/lib` or `/include` postfix so we need to add those ourselves.

When given `/usr/local`, strace shows that meson searches for the library files in the given directories verbatim:
```
newfstatat(AT_FDCWD, "/usr/local/libdecaf.so", 0x7ffd8d377be0, 0) = -1 ENOENT (No such file or directory)
newfstatat(AT_FDCWD, "/usr/local/decaf.so", 0x7ffd8d377be0, 0) = -1 ENOENT (No such file or directory)
newfstatat(AT_FDCWD, "/usr/local/libdecaf.a", 0x7ffd8d377be0, 0) = -1 ENOENT (No such file or directory)
```

There's a tiny cleanup for the postgres backend.

*Do not merge until we've tested it again on the runner*.

### Checklist
I have:
- [X] read the [CONTRIBUTING.md](https://github.com/PowerDNS/pdns/blob/master/CONTRIBUTING.md) document
- [X] compiled this code
- [X] tested this code
- [ ] included documentation (including possible behaviour changes)
- [ ] documented the code
- [ ] added or modified regression test(s)
- [ ] added or modified unit test(s)

PR #14068, opened 2024-04-15
meson: don't assume libdecaf is present when declaring dependency

omoerbeek:
### Short description

Fixes the build in case libdecaf is not available at all for me (tested on macOS and OpenBSD)

### Checklist

I have:
- [X] compiled this code
- [X] tested this code
- [ ] documented the code

### DIFF
--- a/meson/libdecaf/meson.build
+++ b/meson/libdecaf/meson.build
@@ -48,10 +48,12 @@ if not opt_libdecaf.disabled()
       header_include_directories: include_dirs,
     )
 
-    dep_libdecaf = declare_dependency(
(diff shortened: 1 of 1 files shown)


zeha:
unrelated to this fix, but on which platform is '/usr' a valid libdir?

fredmorcos:
> unrelated to this fix, but on which platform is '/usr' a valid libdir?

https://github.com/PowerDNS/pdns/pull/14031#discussion_r1551556290

tldr: I can sleep better at night.

fredmorcos (approved):
It's good. This is the correct fix.

